PR #5711, opened 2019-06-24
Show the list view paging component even when there are less than 10 pages

kjac:
### Prerequisites

### Description

Looks like d52f6b58 introduced a slight bug; list views with less than 10 result pages no longer have the paging component displayed 😄 

image

This PR fixes it; when applied the list view paging works as you'd expect for both large and small result sets:

image


emmaburstow:
Nice spot @kjac! Thanks for the work, as ever. We'll poke you if we need anything. Probably not a real poke. Virtual poke.

Em

nul800sebastiaan:
D'oh! That was silly.. All good on this one, thanks very much!

Issue #4759, opened 2019-02-26
Change format of LocalizedText file to reduce page weight


PeteDuncanson:
The back office gets a JSON object of all the translations a user needs for the labels etc. The format of this works but could be compressed somewhat to same some bytes.

Currently its all done like this:

```
...
actions_toInTheTreeStructureBelow: "In the tree structure below"
actions_translate: "Translate"
actions_unlock: "Unlock"
actions_unpublish: "Unpublish"
actions_update: "Update"
assignDomain_addNew: "Add new Domain"
assignDomain_domain: "Domain"
assignDomain_domainCreated: "New domain '%0%' has been created"
assignDomain_domainDeleted: "Domain '%0%' is deleted"
assignDomain_domainExists: "Domain '%0%' has already been assigned"
...

```

However if we nested translations by area in a JSON object then I think we would save quite a bit, something like this:

```
...
actions: {
	toInTheTreeStructureBelow: "In the tree structure below"
	translate: "Translate"
	unlock: "Unlock"
	unpublish: "Unpublish"
	update: "Update"
},
assignDomain {
	addNew: "Add new Domain"
	domain: "Domain"
	domainCreated: "New domain '%0%' has been created"
	domainDeleted: "Domain '%0%' is deleted"
	domainExists: "Domain '%0%' has already been assigned"
}
...
```

Doing a quick bit of cut and paste testing in a text editor I think I can get my 130kb file down to about 115kb file. Not a massive saving but every little helps and should be quite easy to do as only two points to touch to do it.

First you need to change the "shape" of the data we generate here:

https://github.com/umbraco/Umbraco-CMS/blob/91c52cffc8b7c70dc956f6c6610460be2d1adc51/src/Umbraco.Web/Editors/BackOfficeController.cs#L177

And then either change where/how we get it here:

https://github.com/umbraco/Umbraco-CMS/blob/dev-v8/src/Umbraco.Web.UI.Client/src/common/services/localization.service.js#L39

So that we take the new shape into account.

Or change the shape of the data back into a dictionary (ie flatten the object) so the existing code "just works". To do that you would need to loop over the objects after we've loaded the file here:

https://github.com/umbraco/Umbraco-CMS/blob/dev-v8/src/Umbraco.Web.UI.Client/src/common/services/localization.service.js#L85

nul800sebastiaan:
[...]

PR #5712, opened 2019-06-24
Reduce size of LocalizedText output.

mortenbock:
This fixes https://github.com/umbraco/Umbraco-CMS/issues/4759

### Description
Refactored the json format returned in order to save some bytes, and adapted the frontend code that did not use the correct method for labels.

### Possible breaking changes
There is no more `localizationService.dictionary`. The data has changed, and I don't think it was supposed to be available to begin with.

The data in `eventsService.emit("localizationService.updated", response.data);` has changed. I cannot find any consumers of this event. I left it because some might want to be notified, but maybe the data should just not be included since the format is internal?

Does there need to be any cachebusting for when people upgrade, in case they have cached the localizations? Or does the upgrade already bust those?

I added a parameter to the `localize` method. Will the docs automatically update? https://our.umbraco.com/apidocs/v8/ui/#/api/umbraco.services.localizationService

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/localization.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/localization.service.js
@@ -29,35 +29,48 @@ angular.module('umbraco.services')
     var resourceFileLoadStatus = "none";
     var resourceLoadingPromise = [];
 
-    function _lookup(value, tokens, dictionary) {
+    // array to hold the localized resource string entries
+    var innerDictionary = [];
+
+    function _lookup(alias, tokens, dictionary, fallbackValue) {
 
         //strip the key identifier if its there
-        if (value && value[0] === "@") {
-            value = value.substring(1);
+        if (alias && alias[0] === "@") {
+            alias = alias.substring(1);
         }
 
+        var underscoreIndex = alias.indexOf("_");
         //if no area specified, add general_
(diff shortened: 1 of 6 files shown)


emmaburstow:
Hi @mortenbock 

Thanks for the work here! We'll take a look and shout if we need anything from you. Speak soon!

Emma

nul800sebastiaan:
> There is no more localizationService.dictionary. The data has changed, and I don't think it was supposed to be available to begin with.

Correct. I'll mark this PR as breaking change, even though I don't think anybody would be using the changed items.

> The data in eventsService.emit("localizationService.updated", response.data); has changed. I cannot find any consumers of this event. I left it because some might want to be notified, but maybe the data should just not be included since the format is internal?

Good call, I found a little bit of example code online that uses it, so I think it's fine to leave it.

> Does there need to be any cachebusting for when people upgrade, in case they have cached the localizations? Or does the upgrade already bust those?

The CDF version update during upgrade and that will do all the cachebusting

> I added a parameter to the localize method. Will the docs automatically update? our.umbraco.com/apidocs/v8/ui/#/api/umbraco.services.localizationService

Yep.

nul800sebastiaan:
Thanks very much for the help @mortenbock - code looks good and seems to function well! 👍

PR #5713, opened 2019-06-24
V7: Reintroduce the bypass security option in pickers

bergmania:
Fixes: https://umbraco.visualstudio.com/D-Team%20Tracker/_workitems/edit/1479


src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -1,1219 +1,1227 @@
...
+        {
+            if (contentId <= 0) throw new HttpResponseException(Request.CreateResponse(HttpStatusCode.NotFound));
+            var content = Services.ContentService.GetById(contentId);
+            if (content == null) throw new HttpResponseException(Request.CreateResponse(HttpStatusCode.NotFound));
+
+            //TODO: Should non-admins be able to see detailed permissions?
+
+            var allUserGroups = Services.UserService.GetAllUserGroups();
+
+
```
Shazwazza: Shouldn't need to modify content controller at all, all changes will come from the EntityController

Shazwazza: Is this change even required?

src/Umbraco.Web/Editors/EntityController.cs
```diff
@@ -1,957 +1,1024 @@
...
+        /// <summary>
+        /// Searches for results based on the entity type
+        /// </summary>
+        /// <param name="query"></param>
+        /// <param name="type"></param>
+        /// <param name="searchFrom">
+        /// A starting point for the search, generally a node id, but for members this is a member type alias
+        /// </param>
+        /// <returns></returns>
+        [Obsolete("This method is obsolete, use the overload with dataTypeId instead", false)]
```
Shazwazza: You could still make this backwards compatible by just adding an optional `Guid? dataTypeId = null` parameter as an additional param

src/Umbraco.Web/Editors/EntityController.cs
```diff
@@ -1,957 +1,1024 @@
...
+        {
+            var x = GetResultForId(id, UmbracoEntityTypes.Document);
+
+            var url = Umbraco.Url(id);
+            var anchorValues = Services.ContentService.GetAnchorValuesFromRTEs(id);
+            return new UrlAndAnchors(url, anchorValues);
+        }
+
+        [HttpPost]
+        public IList<string> GetAnchors(string rteContent)
```
Shazwazza: Should be IEnumerable, IList is mutable which we don't want

src/Umbraco.Web/Editors/EntityController.cs
```diff
@@ -1,957 +1,1024 @@
...
+                case UmbracoEntityTypes.Domain:
+                case UmbracoEntityTypes.Language:
+                case UmbracoEntityTypes.User:
+                case UmbracoEntityTypes.Macro:
+                default:
+                    throw new NotSupportedException("The " + typeof(EntityController) + " does not currently support data for the type " + type);
+            }
+        }
+
+        [Obsolete("This method is obsolete, use the overload with dataTypeId instead", false)]
```
Shazwazza: Same as above, this can be backwards compat by just adding `Guid? dataTypeId = null` as the last parameter

src/Umbraco.Web/Editors/EntityController.cs
```diff
@@ -1,957 +1,1024 @@
...
+
+                    int[] aids = null;
+                    switch (type)
+                    {
+                        case UmbracoEntityTypes.Document:
+                            aids = Security.CurrentUser.CalculateContentStartNodeIds(Services.EntityService);
+                            break;
+                        case UmbracoEntityTypes.Media:
+                            aids = Security.CurrentUser.CalculateMediaStartNodeIds(Services.EntityService);
+
```
Shazwazza: I think we should just make `Guid? dataTypeId = null` as an optional parameter - webapi might have issues selecting either of these two actions (not sure, but can easily just be one)

src/Umbraco.Web/Editors/MediaController.cs
```diff
@@ -1,1010 +1,1017 @@
...
+        /// </summary>
+        [FilterAllowedOutgoingMedia(typeof(IEnumerable<ContentItemBasic<ContentPropertyBasic, IMedia>>), "Items")]
+        public PagedResult<ContentItemBasic<ContentPropertyBasic, IMedia>> GetChildren(int id,
+            int pageNumber = 0,
+            int pageSize = 0,
+            string orderBy = "SortOrder",
+            Direction orderDirection = Direction.Ascending,
+            bool orderBySystemField = true,
+            string filter = "",
+            Guid? dataTypeId = null)
```
Shazwazza: We shouldn't need to modify MediaController at all, all changes go through the EntityController

src/Umbraco.Web/Models/ContentEditing/UrlAndAnchors.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            Url = url;
+            AnchorValues = anchorValues;
+        }
+
+        [DataMember(Name = "url")]
+        public string Url { get; }
+
+        [DataMember(Name = "anchorValues")]
+        public IList<string> AnchorValues { get;  }
```
Shazwazza: Should be IEnumerable, IList is mutable which we don't want

src/Umbraco.Web/Models/ContentEditing/UrlAndAnchors.cs
```diff
@@ -0,0 +1,21 @@
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+namespace Umbraco.Web.Models.ContentEditing
+{
+    [DataContract(Name = "urlAndAnchors", Namespace = "")]
+    public class UrlAndAnchors
+    {
+        public UrlAndAnchors(string url, IList<string> anchorValues)
```
Shazwazza: Should be IEnumerable, IList is mutable which we don't want

src/Umbraco.Web/Models/Mapping/ContentPropertyBasicConverter.cs
```diff
@@ -1,75 +1,78 @@
...
+            {
+                LogHelper.Error<ContentPropertyBasicConverter<T>>(
+                    "No property editor found, converting to a Label",
+                    new NullReferenceException("The property editor with alias " +
+                                               property.PropertyType.PropertyEditorAlias + " does not exist"));
+
+                editor = PropertyEditorResolver.Current.GetByAlias(Constants.PropertyEditors.NoEditAlias);
+            }
+
+            var dataTypeDefinition = DataTypeService.GetDataTypeDefinitionById(property.PropertyType.DataTypeDefinitionId);
```
Shazwazza: This is kinda ugly since it causes an N+1 DB lookup - though these are cached so might not be such an issue.... just mentioning cuz it could be nicer to do in v8

Alternatively which will be quicker - is to use the EntityService.GetKeyForId method since that's all it's doing instead of resolving the full object

src/Umbraco.Web/Models/Mapping/PreValueDisplayResolver.cs
```diff
@@ -1,84 +1,94 @@
...
+            // Set up the defaults
+            var dataTypeService = _dataTypeService;
+            var preVals = dataTypeService.GetPreValuesCollectionByDataTypeId(source.Id);
+            IDictionary<string, object> dictionaryVals = preVals.FormatAsDictionary().ToDictionary(x => x.Key, x => (object)x.Value);
+            var result = Enumerable.Empty<PreValueFieldDisplay>();
+
+            // If we have a prop editor, then format the pre-values based on it and create it's fields
+            if (propEd != null)
+            {
+                result = propEd.PreValueEditor.Fields.Select(Mapper.Map<PreValueFieldDisplay>).AsEnumerable();
```
Shazwazza: don't need `AsEnumerable()` ... it is already IEnumerable based on a Select

src/Umbraco.Web/Search/UmbracoTreeSearcher.cs
```diff
@@ -28,12 +29,37 @@ internal class UmbracoTreeSearcher
         /// <param name="pageSize"></param>
         /// <param name="pageIndex"></param>
         /// <returns></returns>
+        [Obsolete("This method is obsolete, use the overload with ignoreUserStartNodes instead", false)]
```
Shazwazza: why not just add `ignoreUserStartNodes = false` to as another param? If we are worried about people having to rebuild their projects, then we will need to change a lot more in this PR! otherwise no need for an extra overload.

src/Umbraco.Web/WebApi/Filters/EnsureUserPermissionForContentAttribute.cs
```diff
@@ -1,143 +1,163 @@
...
+
+        public override void OnActionExecuting(HttpActionContext actionContext)
+        {
+            if (UmbracoContext.Current.Security.CurrentUser == null)
+            {
+                //not logged in
+                throw new HttpResponseException(System.Net.HttpStatusCode.Unauthorized);
+            }
+
+            var ignoreUserStartNodes = false;
```
Shazwazza: There should be no changes to this file at all, there's no need since no changes go through the content controller (this is also dangerous to do)

src/Umbraco.Web/WebApi/Filters/FilterAllowedOutgoingMediaAttribute.cs
```diff
@@ -77,9 +78,20 @@ public override void OnActionExecuted(HttpActionExecutedContext actionExecutedCo
 
         protected virtual void FilterItems(IUser user, IList items)
         {
-            FilterBasedOnStartNode(items, user);
+
+            Guid? dataTypeId = Guid.TryParse(HttpContext.Current.Request.QueryString.Get(TreeQueryStringParameters.DataTypeId), out var temp) ? (Guid?)temp : null;
```
Shazwazza: There should be no changes to this file at all, there's no need since no changes go through the media controller (this is also dangerous to do)

src/Umbraco.Web.UI.Client/src/common/resources/entity.resource.js
```diff
@@ -1,555 +1,624 @@
...
+                        "entityApiBaseUrl",
+                        "GetUrlAndAnchors",
+                        [{ id: id }])),
+                'Failed to retrieve url and anchors data for id ' + id);
+        },
+
+
+        getAnchors: function (rteContent) {
+
+            if (rteContent == null || rteContent.length === 0) {
```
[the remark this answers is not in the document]
bergmania: is `rteContent == null` with only `==` not doing the same?

Issue #5700, opened 2019-06-20
Macro Rendered in Richtext Editor when not configured to do so.


bkclerke:
I have a blank v8.0.2 site that i setup locally with a sql ce database and created a test macro with a content picker parameter called page. The macro is configured to "use in rich text editor and the grid" but it is not configured to "render in rich text editor and the grid" however in my rich text editor in the content the macro is rendering the content in the editor instead of the default macro name in the orange dotted box in the richtext editor.

Macro configuration:
![Screen Shot 2019-06-20 at 2 38 30 PM](https://user-images.githubusercontent.com/4968254/59883342-95778d00-9369-11e9-967e-83e81d45c399.png)

Rendering my macro content in editor:
![Screen Shot 2019-06-20 at 2 42 19 PM](https://user-images.githubusercontent.com/4968254/59883358-a58f6c80-9369-11e9-8429-c6587a2534d9.png)

This macro is super basic for testing purposes, just rendering the name of a content picker page that I have selected via the parameter "page". I think this is what is giving me an error in another v8 site that I have setup. the backoffice is trying to render the content in the editor (when it isn't configured to do so) so I am getting an error in the backoffice of umbraco when I load that content (however my macro works just fine when rendered on the site). 

The screenshot below is of initial error that I'm receiving which lead me to test why the macro is trying to render in the backoffice content on my grid editor (it is not set to render content as well).
![Screen Shot 2019-06-20 at 2 45 33 PM](https://user-images.githubusercontent.com/4968254/59883557-57c73400-936a-11e9-9961-270c76171904.png)

This is a guess as to why I'm receiving on another site but I think the initial issue of rendering the content in the editor if not configured to do so, should stop it from rendering in the editor (and may possibly fix the other issue i'm seeing). Unless the functionality has changed for v8, I assume that is how it is supposed to work, the macro defaults and will only render the content inside of the editor if it is configured to do so. Please let me know if I am mistaken. I ran this test on a new blank site so I think it is relatively easy to test. Thank you!

nul800sebastiaan:
I can confirm this is a big, should be fixed in https://github.com/umbraco/Umbraco-CMS/pull/5719

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/5719

PR #5719, opened 2019-06-25
Fixes Macro Rendered in Richtext Editor when not configured to do so

nul800sebastiaan:
See #5700 for bug report. We were just using the wrong variable. 

You can compare it to the v7 version where the correct variable was used: 

https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Web/Editors/MacroController.cs#L105

To test, disable/enable `Render in rich text editor and the grid` and verify that it renders / doesn't render.

### DIFF
--- a/src/Umbraco.Web/Editors/MacroRenderingController.cs
+++ b/src/Umbraco.Web/Editors/MacroRenderingController.cs
@@ -108,7 +108,7 @@ namespace Umbraco.Web.Editors
 
             //if it isn't supposed to be rendered in the editor then return an empty string
             //currently we cannot render a macro if the page doesn't yet exist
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Note: the first build failed randomly, a subsequent build of this commit succeeded: https://umbraco.visualstudio.com/Umbraco%20Cms/_build/results?buildId=22391

warrenbuckley:
One-liner code change reads fine, off to test this out

warrenbuckley:
Have tested before & after the PR applied and can confirm this is all good :)

bkclerke:
thank you both for fixing this! :)

PR #5732, opened 2019-06-27
Hide grid reorder when scrolling to end of grid content

### LINKED ISSUE #5147: Hide grid reorder when scrolling to end of grid content
By default the grid reorder is positioned static, but get sticky (fixed) when its position reach the bottom of the header. However you can have several properties after the grid content and when scrolling futher down the page the reorder bar remain sticky. Instead it should be hidden when the bottom of the grid content reach the bottom of header.

image

image

Here is an example where the sticky element (in this case a sidebar) moves up, when it reach the footer. https://abouolia.github.io/sticky-sidebar/

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #5147 

### Description

Changes here fix the overly-persistent umbStickyBar directive - currently the directive does all manner of cloning and positioning and scroll sniffing to attach the element top of the container, but does not release the stuck element when the container has scrolled off screen (or out of view). Screenshots in @bjarnef 's issue show this.

Attaching/detaching the umbStickyBar  element is trivial using position:sticky. It sticks when it hits the top, and detaches appropriately too. Issue with sticky is that there isn't an event (or state) to indicate that the element has been stuck.

Enter IntersectionObserver. 👍 

Using IntersectionObserver we can watch for when an element enters or exits another (typically a parent or ancestor). To make that work here, I've rewritten the umbStickyBar directive to add two observed elements as siblings of the umbStickyBar. These 'sentinels' are positioned at the top and bottom of the umbStickyBar's parent element so we can monitor its relative position as the page scrolls.

When the sentinels enter/exit, we toggle a class on the umbStickyBar element, to indicate [in]active state.

All that to add a box-shadow to the umbStickyBar when it is stuck! Sounds like a lot, but compared to the old directive, it's much, much cleaner and should be easier to maintain.

To verify, open a page with a grid editor and scroll. The grid header (with the reorder button) is the umbStickyBar element, and will unstick once the grid has scroll out of view. I've verified this on a grid in a standard editor, on a grid in an infinite editor, and on a list view in an infinite editor (from the homepage node in the starterkit).


nathanwoulfe:
@poornimanayar I destroyed the previous PR - rebased my branch but ended up with merge commits fouling up the history. Easier to just delete that branch and start fresh. Means a few duplicate commits in this PR.

Looking for the console error, I can't reproduce, but have found an issue with the --active class being added on page load, which I'm sure wasn't happening before. Will fix that.

nul800sebastiaan:
Thanks @nathanwoulfe - I need some help from people at HQ on this one as I'm not too familiar with this whole sticky thing :-)

Will get back to you!

poornimanayar:
@nathanwoulfe sorry I had to put you through this. I rebased the latest to your old PR and that's when I started seeing the console error. But I can see the functionality works well :-)

nathanwoulfe:
@nul800sebastiaan all good my friend. The sticky part essentially works out of the box, the javascript exists to add the box shadow on the header (which it does by watching for when the element gets stuck)

@poornimanayar don't apologise, was my fault for messy git work! I see Seb merged latest from dev into this, so should be up to date. Do you remember the console error?

poornimanayar:
hey @nathanwoulfe ,

I had a quick wander yesterday after the update to the PR and I cannot see the error anymore. The error I saw earlier was on the lines of "el not defined". And the minute I switched to some other PRs and rebuilt it was gone. Hence my message to you :-)

Poornima

Shazwazza:
@nathanwoulfe any chance you would know what will happen if someone does try to use a browser that either doesn't support position sticky or IntersectionObserver?

Shazwazza:
@nathanwoulfe also ... on a similar note, have you seen the code that pins the tinymce menu when it's inside of the grid and you need to scroll through the grid? Might be useful there as well (different issue of course)

nathanwoulfe:
@Shazwazza sticky and observer are supported everywhere other than IE11, which isn't officially V8-friendly, so should be fine. Sticky in IE11 will not do anything, it's an unknown CSS property so parser will ignore it. Observer though I should probably wrap in a check, otherwise it's going to throw as undefined (I'd imagine). Will check it out and update if needed.

nathanwoulfe:
IE11 has bigger problems than Intersection Observer - usages of .closest() being the first thing to throw when opening the backoffice...

Shazwazza:
@nathanwoulfe haha, all good then :) I'll get to testing this

Shazwazza:
@nathanwoulfe this is brilliant :) I didn't know about these APIs and it works great!

Shazwazza:
_not sure why this PR isn't auto-closing, i merged it via git cmd line, oh well, it's merged with your commits, will close now_

nathanwoulfe:
@Shazwazza it's pretty slick hey. Takes a bit of wiring up, but way more efficient than watching scroll, and much easier to apply to nested elements.

I reckon the auto-closing is due to me massacring an earlier PR, and then deleting my fork before this was merged...

bjarnef:
@nathanwoulfe it seems this have affects the layout in media section a bit.

v8.1.0 release:

image

image


Latest changes in `v8/dev` branch:

image

image

bjarnef:
When opening developer window and navigating to e.g. content section dashboards, it seems `rootBounds` in some cases is `null`.

image

nathanwoulfe:
Should be able to resolve the error by null-checking rootBounds, but will need to look at why it's null too - guessing it will be related to the dimensions of the parent element...

nathanwoulfe:
@bjarnef => #5834 fixes these issues.

Issue #5727, opened 2019-06-26
Accidental click outside of a dialog closes the dialog, and changes are lost


skttl:
When you are deep into a dialog (like the public access dialog), and by accident, clicks outside of it, it just closes and forgets all about what you where doing.

A warning or a confirm dialog would be really nice here :)

https://i.imgur.com/8DO3gZ8.gifv

nul800sebastiaan:
Absolutely!

kjac:
I did something similar to another dialog not so long ago. I will have a look at applying that to public access as well. A generic solution across all dialogs is probably going to clash miserably with the current solution for content.

kjac:
In the latest codebase I can't reproduce click-outside-to-close:

![public-access-click-outside](https://user-images.githubusercontent.com/7405322/60254543-72e7f580-98ce-11e9-9a9b-0165d7f065e9.gif)

However the point of this issue is still valid if you accidentally hit "close" after spending time setting it all up. PR #5733 addresses this.

kjac:
Disregard my last comment. It seems to be a bug in the latest codebase... I'm on it.

kjac:
New PR in #5740

PR #5690, opened 2019-06-19
V8: Fix confirmation for permissions dialog

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description

The content permissions dialog has an "unsaved changes" confirmation feature, if you navigate away without saving the configured permissions. It's a bit unorthodox for dialogs, but I guess it makes sense as there is (potentially) a lot of configuration involved.

However... the confirmation looks really really funny 🤣 

set-permission-confirmation-before

This PR fixes the confirmation dialog. When applied it looks like this:

set-permission-confirmation-after

PR #5733, opened 2019-06-27
V8: Prompt to discard changes when you close public access without saving

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5727

### Description

As described in #5727 you may risk loosing a bunch of work when setting up public access, if you accidentally hit "close" instead of "save" in the public access dialog.

This PR adds a confirmation prompt in case you have made changes and hit "close":

public-access-prompt-to-save

#### Future considerations

This fix has pretty much duplicated code from #5690. If both changes are merged, we should consider refactoring both and introduce a central service for this kind of prompting.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/content.protect.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/content.protect.controller.js
@@ -1,7 +1,7 @@
 (function () {
     "use strict";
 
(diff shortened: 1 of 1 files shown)


kjac:
Disregard this PR, it's founded upon something that seems to be a bug in the current codebase. I'll get back with another solution 😄

PR #5734, opened 2019-06-27
V8: Fix broken on-outside-click for context menus and dialogs

kjac:
### Prerequisites

### Description

Looks like fd5b5b67 broke the ability to close context menus and dialogs by clicking outside them:

outside-click-before

As is stands, the only way to close them is to pick another node or hit escape.

Not sure if the change to `events.directive.js` was founded in another issue or just good ol' housekeeping - @Shazwazza might have input here?

Anyway... this PR reverts the change, and this makes the context menus and dialogs close as expected when clicking outside them:

outside-click-after


nul800sebastiaan:
I was just about to ping you if you remember touching this part! 

Thanks for this, that was a frustrating few minutes to try and get out of those dialogs 🙈

Issue #5737, opened 2019-06-27
GetCropUrl returns null with valid cropAlias param V8


cleversolutions:
### Bug summary
ImageCropperTemplateExtensions.GetCropUrl returns null when provided a valid cropAlias parameter unless you save each image in your media library after adding a new crop.

### Specifics
This occurs in 8.0.2 and v8/dev using Chrome (but should affect all browsers)

### Steps to reproduce
- V8 with SK installed
- Update Image Media Type -> Update umbracoFile Image Cropper DataType settings to include a crop type (ie square)
- Update Products.cshtml / template with `<img src='@product.Photos.GetCropUrl("square")' />`
- View page on frontend & confirm image renders (they don't)
- Goto Media Section & Choose a Product Image
- Set manual crops for the image
- View page on frontend & confirm image renders (they do)

### Expected result
The images should be rendered. See reproduction for more details.

### Actual result
The images are not. See reproduction for more details.

nul800sebastiaan:
Yes, unfortunately this will be true for any updates you will do for datatypes. We do not update existing data when a datatype updates simply because it is an expensive operation that could take a very long time and much processing power. 

It's up to you to do some defensive querying for this or possibly provide a migration tool for when you know things will need to change. Just be aware that you'll need to be careful with that migration script in case you have a large number of nodes (see previous paragraph).

nul800sebastiaan:
Hmm, I just saw your related PR, and it's not what I thought you wanted to get solved. ;-)

I'll have a look at it soon, this might be an okay solution, not sure yet.

kjac:
@cleversolutions I can't reproduce this issue on the latest codebase. `GetCropUrl(string alias)` returns URLs for newly added crops on the image cropper datatype, even if the image in question hasn't been saved with the newly added crops.

Could I ask you to try this out again on the latest codebase?

cleversolutions:
Thanks @kjac I’ll retest and confirm tomorrow.

cleversolutions:
Hi @kjac I reproduced it this morning from a fresh clone of v8/dev. 
I created a new site with the starter kit.
Added a 100x100pixel crop to ImageCropper data type called Test.
![cropper](https://user-images.githubusercontent.com/23444011/71902384-fad97280-312f-11ea-9616-7f4778a9b306.JPG)

Then on line 35 of Products.cshtml changed the line to use the crop
```
<div class="product-grid">
            @if (Model.FeaturedProducts != null)
            {
[...]

PR #5739, opened 2019-06-27
Add ImageCrops from DataType if empty

cleversolutions:
### Prerequisites

This fixes issue #5737 

### Description
When you add a crop alias to Image Cropper after media already exists in your Media library it doesn't pick up any of the crops. This PR updates ImageCropperTemplateExtensions to check for this condition and configure crops from the DataType.Configuration. 

### Steps to reproduce
As per issue #5737 
- V8 with SK installed
- Update Image Media Type -> Update umbracoFile Image Cropper DataType settings to include a crop type (ie square)
- Update Products.cshtml / template with 
- View page on frontend & confirm images render (they don't before PR)
- Goto Media Section & Choose a Product Image
- Set manual crops for an image
- View page on frontend & confirm all images render (only the updated image does before PR)

With this PR, I've also tested another scenario
- Update Image Media Type -> Update umbracoFile Image Cropper DataType settings to include another crop type (ie 1x1)
- Update Products.cshtml / template with ``
- View page on frontend & confirm all images render including the updated one. 

### Caveat
There may be a better way of solving this, I'm new to the Umbraco code base. The fact that the image which was edited has 2 crops in imageCropperValue.Crops the second time makes me wonder if the real issue lies somewhere. Someone with a better understanding of the caching/converters may have a better idea.


poornimanayar:
Hey @cleversolutions ,

Welcome here and  congratulations on your first PR! We will review the work and get back to you if we need anything more. Watch out this space :-)

Regards,
Poornima, [Pull Request Team](https://umbraco.com/blog/introducing-the-new-cpr-team/)

cleversolutions:
I added a comment to the change, I felt it needed an explanation.

nul800sebastiaan:
Really cool @cleversolutions - I didn't think of dynamically adding these crops if they don't yet exist. I will need to talk to some people at HQ to see if this is a workable solution. Will get back to you!

cleversolutions:
Hey @nul800sebastiaan is there anything you need from me for this PR?

nul800sebastiaan:
Thanks for the reminder @cleversolutions - I am not sure if we should do this, it seems like the performance of the `GetCropUrl` would suffer quite severely if we have to go to the database for each call until someone actually saves the crops for all the affected media items. I'll have to ask around.

Shazwazza:
Hi, this should instead be done in the `ImageCropperValueConverter` instead of in this ext method which is responsible for populating the `ImageCropperValue` model. You can apply the same/similar technique there and it should be fine. However, you can probably use the method `DataType.ConfigurationAs<ImageCropperConfiguration>()` instead of doing all that try/cast. 

Since this data is coming from the front-end cache, the way access to a DataType works is that it does hit the DB once but not over and over again so the way you are accessing the DataType is fine.

cleversolutions:
Thanks @shazwazza, I’ll work it in like that instead.

cleversolutions:
I couldn't figure out how to fix the mess I made of this so I re-created the branch and have submitted a new pull request #6950

Issue #3355, opened 2018-10-19
Add SourceLink metadata


tcmorris:
SourceLink allows for source code debugging of .NET assemblies from NuGet. Developers who download the package and have SourceLink enabled in Visual Studio can step into its source code.

More info:
https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink

Happy to pick this up, but thought I'd raise it as an idea first and see if there are any potential issues.

nul800sebastiaan:
Wow, that looks super cool! Go for it! 

Not sure how to test it but.. we'll get to that ;-)

kristianschneider:
I think it is only for dotnet core: 
[https://github.com/dotnet/sourcelink#prerequisites](https://github.com/dotnet/sourcelink#prerequisites)

dampee:
I've been testing with sourceLink for some time now.  The only thing which you need to do is add the (pre-release) nuget package:

    <ItemGroup>
      <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-*" PrivateAssets="All"/>
    </ItemGroup>

warrenbuckley:
@dampee hmmm 🤔 I did try this & had no luck when I added the NuGet reference to that package. The Powershell build was not creating the nuget artifacts with the XML metadata for the commit SHA & the GitHub repo URL that SourceLink should be creating.

Hence from my understanding of how SourceLink works was that it needs to work with the newer SDK CSProj files. Once I had that done I was going to come back to try adding SourceLink in order to get this developer debugging story done.

warrenbuckley:
A PR for this is here @dampee @kristianschneider & @tcmorris feel free to test & prod it too if you like - https://github.com/umbraco/Umbraco-CMS/pull/5359

tcmorris:
Nice one @warrenbuckley not sure I'll have much time to test this, but wanted to say cheers for looking into it and the relevant changes to make it work. #H5YR

PR #5743, opened 2019-06-28
V8: Adds in SourceLink Debugging Feature

warrenbuckley:
Adds in debugging functionality called SourceLink allowing users to F11 directly into Umbraco source code from files hosted on GitHub itself

Fixes: #3355


warrenbuckley:
OK latest test with the commit 23a7654 seems to be unsuccessful. There needs to be a nicer/easier way to test & verify this without the rinse & repeat pattern.

warrenbuckley:
## Test Notes

- Use Visual Studio 2019
- Enable SourceLink (See image)
image

PR #5745, opened 2019-06-29
V8: Make the markdown editor use the Umbraco link picker

### LINKED ISSUE #5714: Markdown editor link picker
Umbraco v8.0.2:

The Markdown editor link picker throws a console error when clicked.

Reproduction
------------

### Steps to reproduce

- Create a new doc type
- Add a markdown editor property
- Edit the property value on a content node
- Click the link picker

### Expected result
- Link modal / dialogue will show, allowing the user to pick the link location

### Actual result
console error is throw.

```
$(dialog).modal()
"TypeError: $(...).modal is not a function"
```
/umbraco/lib/markdown/markdown.editor.js line 1133

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #5714

### Description

As described in #5714, the "insert link" feature of the markdown editor is currently broken. This PR fixes it. 

However... in V7, "insert link" spawned a very un-Umbraco-ish dialog like this:

image

I have taken the liberty to change that, so the V8 version utilized the built-in linkpicker. When this PR is applied, it all works like this:

markdown-editor-linkpicker


poornimanayar:
Hey @kjac ,

Thanks for the PR. You know the drill :-)

Poornima

skttl:
Is the URL going to be hardcoded? I looked at the markdown valueconverter, and it seems like it parses internal links. So could be useful to add internal links in the same way as the rte does?

kjac:
I thought about that and yes, it would make sense. But the media picker inserts media in the same fashion, so I figured making both of them internally managed would fit better in a separate fix.

kjac:
PR updated with latest v8/dev

nul800sebastiaan:
@kjac Hmm, I understand the idea here and I quite like it, I'm just bothered by a few things that will not work, for example opening in a new tab is not going to work in markdown does the "link title" field actually insert a link title?

kjac:
@nul800sebastiaan I will test the link title, I'm not sure if we can push it to the markdown editor. Maybe the link picker could be made configurable, so it could hide "title" and "opens in a new window", but that also seems like overkill to solve a nice-have for a markdown editor that's probably not used all that much.

kjac:
@nul800sebastiaan OK. I have updated the PR so the markdown editor uses the link picker title for the link text by default, _unless other text is preselected for the link_. Here's a GIF that demonstrates it:

markdown-link-title

Markdown doesn't support link targets ("open in new window"), but it seems a bit overkill to introduce an option to hide the "open in new window" checkbox from the linkpicker, no?

nul800sebastiaan:
You're right, it would be very nice to have but.. let's see if there's a serious enough need for opening in new window first.

This title functionality looks most excellent! 👍

mjbarlow:
Hey, is this due to be merged in?, this would massively improve on the current broken "insert link" feature.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/markdowneditor/markdowneditor.controller.js
```diff
@@ -27,6 +27,19 @@ function MarkdownEditorController($scope, $element, assetsService, editorService
         editorService.mediaPicker(mediaPicker);
     }
 
+    function openLinkPicker(callback) {
+        var linkPicker = {
```
abjerner: @kjac While not documented, the link picker editor already has a `hideTarget` property - I'm using it in my redirects package here:

https://github.com/skybrud/Skybrud.Umbraco.Redirects/blob/7f8680784de6808fa60eb707e2c1385d5e18a22d/src/Skybrud.Umbraco.Redirects/App_Plugins/Skybrud.Umbraco.Redirects/Scripts/Controllers/Editors/Destination.js#L6

kjac: @abjerner nice! I'll get that fixed asap.

kjac: Works brilliantly.. all done!

kjac:
As per @abjerner's request the target selector is now hidden, making the link picker dialog look like this:

image

Issue #5744, opened 2019-06-28
MNTP (v8) is not filtering child nodes where list view is enabled


spudstuff:
The MNTP isn't filtering out child nodes of doctypes that are enabled for list view. The doctype filter is only working on non-listview nodes.

![MNTP in v8.0.2](https://user-images.githubusercontent.com/675943/60345495-872a1080-99fc-11e9-8526-3d7e869fa0d7.gif)

Reproduction
------------
* Install Starter Kit in 8.0.2.
* Create a MNTP datatype with allow items of type set to "content page".
* Add that datatype to the doctype, eg. the Home doctype.
* Edit the home node and use the picker to browse the tree. Note all nodes that aren't listview enabled are greyed out except those of doctype "Content Page" (About Us, etc.)
* Expand "People" which is enabled for list view. Note you can now select people. This is the bug - people should not be selectable.
* If you disable the People doctype for listview, it will be correctly filtered in the tree (because it's not being shown in an umb-mini-list-view anymore).

### Bug summary
Looks like the umb-mini-list-view isn't doing the filtering at all. It does have code to try to filter on doctype along (but not advanced filtering), but does not seem to be working:

> Umbraco-CMS-release-8.0.2\src\Umbraco.Web.UI.Client\src\common\directives\components\umbminilistview.directive.js

```
// filter items if there is a filter and it's not advanced
// ** ignores advanced filter at the moment
if (scope.entityTypeFilter && !scope.entityTypeFilter.filterAdvanced) {
    var a = scope.entityTypeFilter.filter.toLowerCase().replace(/\s/g, '').split(',');
    var found = a.indexOf(c.metaData.ContentTypeAlias.toLowerCase()) >= 0;
    
    if (!scope.entityTypeFilter.filterExclude && !found || scope.entityTypeFilter.filterExclude && found) {
        c.allowed = false;
    }
}
```

I put a `console.log(scope.entityTypeFilter)` just above that if statement and it was `undefined`.

After a bit of a play and thought I almost had it solved with these changes:

`treepicker.controller.js` line 60 define vm.filter:
```
vm.filter = {};
```

`treepicker.controller.js` line 162 within the if($scope.model.filter) check, populate vm.filter:
```
vm.filter.filter = $scope.model.filter;
vm.filter.filterAdvanced = $scope.model.filterAdvanced;
vm.filter.filterExclude = $scope.model.filterExclude;
```

`treepicker.html`, line 77 pass in vm.filter instead:
[...]

PR #5746, opened 2019-06-30
V8: Fixes filter problem with minilistview when no filter is applied (#5744)

spudstuff:
Leading on from @Shazwazza's fix in https://github.com/umbraco/Umbraco-CMS/commit/5c2b0647f3feaf23311bc26876a183bbb348bad5 I noticed two things:

1. The `umbminilistview.directive.js` has the incorrect case on `hasChildren` on line 74 though.

```
c.hasChildren = c.metaData.HasChildren;
```

2. When no doctype filter has been configured on the MNTP, the umbminilistview is still trying to use `scope.entityTypeFilter.filter` but it's undefined, resulting in this exception:

```
angular.js?cdv=1:15536 TypeError: Cannot read property 'toLowerCase' of null
    at umbraco.directives.js?cdv=1:13025
    at Function.h.each.h.forEach (underscore-min.js?cdv=1:5)
    at umbraco.directives.js?cdv=1:13008
    at processQueue (angular.js?cdv=1:17914)
    at angular.js?cdv=1:17962
    at Scope.$digest (angular.js?cdv=1:19075)
    at Scope.$apply (angular.js?cdv=1:19463)
    at done (angular.js?cdv=1:13312)
    at completeRequest (angular.js?cdv=1:13569)
    at XMLHttpRequest.requestLoaded (angular.js?cdv=1:13474) "Possibly unhandled rejection: {}"
```

The fix is to just check `scope.entityTypeFilter.filter` has been defined before filtering.

### To reproduce
* Install Starter Kit in 8.0.2.
* Create a MNTP datatype with no allowed items (everything is allowed).
* Add that datatype to the doctype, eg. the Home doctype.
* Edit the home node and use the picker to browse the tree. Note all regular nodes are selectable.
* Expand "People" which is enabled for list view. Exception seen in console.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbminilistview.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbminilistview.directive.js
@@ -71,7 +71,7 @@
                             }
                             // set published state for content
                             if (c.metaData) {
-                                c.hasChildren = c.metaData.HasChildren;
+                                c.hasChildren = c.metaData.hasChildren;
                                 if(scope.entityType === "Document") {
                                     c.published = c.metaData.IsPublished;
                                 }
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Excellent, thanks very much for the help @spudstuff !

PR #5748, opened 2019-07-01
Prepare MB future

zpqrtbnk:
Prepare for MB future by making it possible to unplug Umbraco.ModelsBuilder in a nice way and replace it.


src/Umbraco.Core/Manifest/IManifestFilter.cs
```diff
@@ -0,0 +1,19 @@
...
+    public interface IManifestFilter
+    {
+        /// <summary>
+        /// Filters package manifests.
+        /// </summary>
+        /// <param name="manifests">The package manifests.</param>
+        /// <remarks>
+        /// <para>It is possible to remove, change, or add manifests.</para>
+        /// </remarks>
+        void Filter(List<PackageManifest> manifests);
```
Shazwazza: Why is `List<>` being used here?

zpqrtbnk: Because we want the filters to be able to add, remove, change... etc the list of manifests. A `List` was simple. Each filter can alter the list. But we can switch to an `IEnumerable` if you'd rather have it an `IEnumerable` - makes each filter a bit more convoluted but would work.

zpqrtbnk:
Description of changes:

**Introduce a `IManifestFilter`** - after manifests are read from the file system, but before they are merged into "the manifest", filters can tweak the list of manifests that have been read - thus allowing to programatically remove manifests, alter manifests, or inject virtual manifests.

_Why? Because in some cases we may want to entirely disable some stuff that are introduced by a package manifest, and the APIs don't fully allow us to do so._

**Make `IEditorValidator` public** - and all related stuff.

_Why? Because otherwise it is impossible to implement another models builder that can validate property aliases._

**Cleanup `GridEditorConfig`** - use the injected manifest parser.

_Why? Because that was a TODO._

src/Umbraco.Core/Manifest/IManifestFilter.cs
Shazwazza: Would it make more sense to have the method return the result of the filter ? so like `IEnumerable<PackageManifest> Filter(IEnumerable<PackageManifest> manifests)` ?

zpqrtbnk: Can do, does not make a lot of difference I believe. Want me to?

Shazwazza: no it's fine, just had to go look at it's usage.

PR #5751, opened 2019-07-01
#5750 - make actions like create folder work again

### LINKED ISSUE #5750: V8.1 - create folder not working in several places

In several places in the settings section several actions from the create dialogue are not working anymore. If you click on the action the dialogoue will be close.

This is on the latest v8.1 source.

folder create

I discovered following actions that are not working : 

### Settings section

**Documenttypes** 

- Documenttype collection
- Folder

**Media types**
- Folder

**Datatypes**
- Folder

**Partial views**
- New Partial view from snippet
- Folder

**Partial view macro files**
- New Partial view macro from snippet
- Folder

** Stylesheets ** 
- Folder

** Javacript ** 
- Folder

There are no errors logged in the console when this happens. Probably all these are cause by the same issue. There maybe other places in the backend where this occurs.

Dave

dawoe:
### Prerequisites

If there's an existing issue for this PR then this fixes #5750 

### Description
Before this fix it was not possible to create folders for documenttypes, datatypes, etc... see related issue.

This PR fixes this. Maybe @kjac can do a review because it was caused by a PR by him last week. As far as I can see the things fixed in his PR are still working

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/events/events.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/events/events.directive.js
@@ -131,6 +131,12 @@ angular.module('umbraco.directives')
                     return;
                 }
 
+                // ignore clicks on dialog actions
+                var actions = $(event.target).parents(".umb-action");
+                if (actions.length === 1) {
+                    return;
+                }
+
                 //ignore clicks inside this element
                 if ($(element).has($(event.target)).length > 0) {
                     return;


kjac:
Great 😄 This works as expected 👍 

The only drawback is that you can now have the actions and the context menu open at the same time. This is to be expected with the way we have to fix these things - same thing applies to the save and publish button "pop-up" menu:

image

Ideally we should get rid of all the custom handling in `function oneTimeClick(event)`, but that's definitively for another time.

nul800sebastiaan:
Thanks very much @dawoe - just noticed this bug while testing before you reported it as well!

PR #5753, opened 2019-07-01
Typo in umb-grid-selector

leekelleher:
### Prerequisites

It's a typo correction, doesn't really need steps to test it.

### Description

The `umb-grid-selector` view template says "Akk" instead of "All".

The localization would automatically replace it.
But for languages where there is no direct translation, it would display incorrectly.


nul800sebastiaan:
ACK! 🤦‍♂  

How did I miss this one? Thanks @leekelleher 👍

PR #5756, opened 2019-07-01
DocType Editor sorting - vertical direction

leekelleher:
### Prerequisites

### Description

In the DocType Editor, when sorting the groups and properties, the draggable direction is freeform (horizontal, vertical or even diagonal).

We can restrict the direction, by setting the `axis` property in the sort options.

---

To test, in the DocType Editor, press "Reorder", then drag the groups and/or properties, the direction will be restricted to the vertical axis.


emmaburstow:
Hiiiiiiii Leeeeee,

Thanks for your PR. We'll take a look and prod you if we need anything. 

Em

nul800sebastiaan:
Oops, looks like I forgot to say thank you last night, thanks Lee! Makes total sense! 👍

PR #5757, opened 2019-07-01
Added constants for Property Editor Groups

leekelleher:
### Prerequisites

### Description

I've added a constants class for Property Editor groups. Mostly for consistency - and also as a package developer, I can know which group names are used for Umbraco core editors.

I've marked this pull request as a draft, as I got carried away with re-formatting the `DataEditor` attributes - breaking each parameter onto a newline, for readability purposes.

If you don't like this style, I can rework the pull request to put them all back on a single line.

---

In terms of testing, it'd be a case of going into the DocType Editor and checking that the various Data Types appear in the correct Property Editor group.

### DIFF
--- a/src/Umbraco.Core/Constants-PropertyEditors.cs
+++ b/src/Umbraco.Core/Constants-PropertyEditors.cs
@@ -186,6 +186,24 @@ namespace Umbraco.Core
                 /// <remarks>Must be a valid <see cref="ValueTypes"/> value.</remarks>
                 public const string DataValueType = "umbracoDataValueType";
             }
+
+            /// <summary>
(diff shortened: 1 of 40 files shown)


emmaburstow:
Hey Lee!

Thanks for teh PR. I can see it's currently a draft but when you're done, we'll take a look at the work and let you know if we need anything from you before we merge. 

all the best,

Em

leekelleher:
Thanks @emmaburstow! I've removed the draft status now.

nul800sebastiaan:
Aah thanks so much for the consistency @leekelleher - the difference in casing between groups was definitely an eyesore!

Issue #5554, opened 2019-05-31
v8: Unpublished multilingual page has PublishedState of Published


PerplexDaniel:
### Bug summary

A page that has "Allow varying by culture" set to true and is unpublished for all cultures has a `PublishedState` of Published on its IContent object.

Is this intended, and are we then supposed to check the `PublishedCultures` property instead in this case? 

It is really confusing that this is different between document types based on the vary by culture checkmark. That is, if the checkmark is not set the page will properly have a `PublishedState` of Unpublished when it is unpublished.

### Steps to reproduce

1. Create a Document Type with Permissions > "Allow varying by culture" checked
2. Create a new page based on that document type.
3. Publish it for 1 language.
4. Unpublish it.
5. Check the `PublishedState` of the IContent:
```csharp
// Inject
IContentService cs;

// Check pageId in Umbraco
int pageId = 1234;

// Check state -- should be false but is true.
var publishedState = cs.GetById(pageId).PublishedState;
```

### Expected result

The `PublishedState` is `false`.

### Actual result

The `PublishedState` is `true`.

### Version
Umbraco 8.0.2

---
_This item has been added to our backlog [AB#1718](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/1718)_

nul800sebastiaan:
I think it's intentional yeah and you'd need to add a check for `PublishedCultures` to get the full picture of what the state of the node is. I'll have a chat at HQ and get some more feedback on this.

nul800sebastiaan:
We had a chat about this at HQ and it turns out I misunderstood what was going on, this seems like a bug that we need to fix.

tom-ingeniuus:
Just to add that, (tested in v8.1) the resulting node is in the ExternalIndex until that index is rebuilt, then it is removed from the index.

Shazwazza:
I created a unit test to show the problem in rev 5d92c8ac6ba8e5991791d313b50bebffebf21e35

[...]

PR #5695, opened 2019-06-19
Enable removal of unpublished content from ExternalIndex

linuselander:
DeferedDeleteIndex.Execute should delete entries in indexes with PublishedValuesOnly

### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes #5218 

### Description
As described in the issue; Content isn't removed from the ExternalIndex when you Unpublish content. You have to Rebuild the ExternalIndex to remove the Unpublished content.
A change to the logic in DeferedDeleteIndex.Execute that determines which Indexes should be affected fixes the issue.
To test: See #5218.

PR #5758, opened 2019-07-01
Unpublished multilingual page has PublishedState of Published

### LINKED ISSUE #5218: V8.0.1: Node remains in the index after being unpublished
I encountered an issue where a content node will remain as a document in my ExternalIndex after being unpublished, which means I will get it in my ISearchResults if I query for its NodeTypeAlias. All its values will be null if I convert it to an IPublishedContent, obviously. Since it is unpublished.

Reproduction
------------

1. Create a new document type and allow it as root.
2. Create a content node with some name
3. Check your ExternalIndex - the node is there, which is expected.
4. Unpublish the content node and check your index again.. you will notice that it is still there.

image

The node will be gone as a document if you rebuild your index.. but you don't wanna do that everytime you unpublish. ;-) 

This item has been added to our backlog [AB#1533](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/1533)

nul800sebastiaan:
This fixes: #5554

The `UnpublishCulture` method would set `PublishedState.Publishing` instead of `PublishedState.Unpublishing`, I assume this was copy/pasted from the `PublishCulture` method.

This fixes 5554 already but I saw further weird behavior in the `ContentService` where the variables `culturesPublishing` and `culturesUnpublishing` would only be populated when jumping into the `if(publishing)` method. In other words: in the `if(unpublishing)` path the line `if (culturesUnpublishing != null)` would always be `false`(!).

I was hoping that would fix #5218 but unfortunately that's still an issue. I don't know how items in the ExternalIndex get updated, so I'm not sure what the problem is here.

In any case, the `culturesPublishing` and `culturesUnpublishing` variables definitely need to be populated outside of the `if(publishing)` statement.

### DIFF
--- a/src/Umbraco.Core/Models/ContentRepositoryExtensions.cs
+++ b/src/Umbraco.Core/Models/ContentRepositoryExtensions.cs
@@ -239,7 +239,7 @@ namespace Umbraco.Core.Models
             foreach (var property in content.Properties)
                 property.UnpublishValues(culture);
 
-            content.PublishedState = PublishedState.Publishing;
+            content.PublishedState = PublishedState.Unpublishing;
         }
 
         public static void ClearPublishInfos(this IContent content)
--- a/src/Umbraco.Core/Services/Implement/ContentService.cs
+++ b/src/Umbraco.Core/Services/Implement/ContentService.cs
@@ -1076,8 +1076,11 @@ namespace Umbraco.Core.Services.Implement
             var variesByCulture = content.ContentType.VariesByCulture();
 
             //track cultures that are being published, changed, unpublished
-            IReadOnlyList<string> culturesPublishing = null;
-            IReadOnlyList<string> culturesUnpublishing = null;
+            //determine cultures publishing which will be based on previous calls to content.PublishCulture and ClearPublishInfo
+            var culturesUnpublishing = content.GetCulturesUnpublishing();
+            var culturesPublishing = variesByCulture
+                ? content.PublishCultureInfos.Values.Where(x => x.IsDirty()).Select(x => x.Culture).ToList()
+                : null;
             IReadOnlyList<string> culturesChanging = variesByCulture
                 ? content.CultureInfos.Values.Where(x => x.IsDirty()).Select(x => x.Culture).ToList()
                 : null;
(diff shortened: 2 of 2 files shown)


zpqrtbnk:
Would seriously recommend a lot of testing of that one + testing that events, caches, everything works correctly. The item is *not* unpublished, it is republished with different cultures. This is why we use `PublishedState.Publishing` and not `PublishedState.Unpublishing`. I understand it may be confusing, and maybe this fixes it, but again, test a *lot*.

And then, it may be for #5218 that Examine does not remove the item, because it sees the node as being republished, not unpublished. And again, maybe this PR fixes it all, but test, test and test again, including the front-end cache.

Shazwazza:
@nul800sebastiaan I agree with @zpqrtbnk here - the `Publishing` and `Unpublishing` actions combined with variants isn't black and white. Unpublishing a culture doesn't mean the content item is being unpublished. In fact there's different terminology internally for cultures - they can be "available" or "not available" (IIRC). It is definitely possible to have all cultures "available" and then unpublish the entire document. All this is doing is setting a flag and then none of those cultures are routable or will exist in the cache. Publishing the document again means that all cultures that were previously available are available again. So Unpublishing a document is not the same as 'unpublishing' - or making unavailable a culture. 

Currently - even if you unpublish all cultures, the document will remain in a Published state, this is true and possibly this functionality could be changed. 

We need to review the logic here and make things more clear as to what is actually happening.

tom-ingeniuus:
Without looking at the code at all, if I unpublish a node and carry out an Examine search, the node remains in the External Index, however if I then rebuild the External Index then that node is removed from the External Index so there seems to be a mismatch here, or perhaps the index rebuild process is different from the update process when a node is unpublished. I have left time for the index to be update so I don't think it's a timing issue. I have 2 cultures and only creating content for one of them (english), then unpublishing it.

Shazwazza:
@tom-ingeniuus isn't that this bug https://github.com/umbraco/Umbraco-CMS/pull/5695?

tom-ingeniuus:
> @tom-ingeniuus isn't that this bug #5695?

yes, I hadn't spotted that one

nul800sebastiaan:
I'll close this as the problem has been fixed in https://github.com/umbraco/Umbraco-CMS/pull/5695

PR #5585, opened 2019-06-04
V8: Accessibility improvements for page name entry

RachBreeze:
This PR aims to fix 43, 69, 75, 81 from the from the A & AA accessibility issue list #5277

**Creating a content item**
When creating a new content type there is screen reader only text which says:
"Create" + content type name

The concept is that it create wors as per Microsoft Teams, here when a user creates a folder they see:
image

**Editing a content item**
When editing an existing content item there is screen reader only text which says
"Edit" + content name

**Name label**
The "name" text box also now now has a hidden label which is "content type name" + Name

**Placeholder**
The placeholder in the name textbox remains the same as prior to this change, as placeholder text provides guidance for many users, placeholder text is not a replacement for labels.

**Class amends**
This PR contains a sr-only class for displaying content for screen readers only.  However there is nothing to stop the hidden text being displayed on the screen if HQ deem fit (subject to improved styling)

**Language amends*
The language files now have accessibility specific place holders these have been added to en.xml and en_us.xml only

Issue #5277, opened 2019-04-16
V8: Fix A & AA accessibility issues throughout the back-office


MMasey:
# IMPORTANT UPDATE
We have an updated trello board with all the tickets from the latest Accessibility Audit.  
Trello Board for Umbraco Accessibility Issues (8.4>) - https://trello.com/b/7M8RosrE/umbraco-accessiblity-issues-84

Note - I've left the link for the old trello board for reference. We may close this ticket soon and open a new one to tidy things up a little bit.

---

There are currently a large number of accessibility and in tern usability issues throughout the back-office. One of the wonderful members of the PR team Danny Lancaster has created an accessibility audit, detailing each of the bugs found. 

There are currently **151** known issues that need to be fixed. Rather than create an issue for each one, the idea is to do the following.

1. Have a look through the table below and pick an issue that you'd like to fix
2. Fix the issue 😄
3. Submit a PR
4. Leave a comment with a link to your PR here and we will update the trello board (see comment below).

I've added some links to various articles and tools to help you on your way below. https://github.com/umbraco/Umbraco-CMS/issues/5277#issuecomment-483609542

If you have any questions you can reach out either here or on the Umbraco Community #accessibility Slack channel. You can join the community slack channel by going to [umbracians.chat](http://umbracians.chat/).

**We have a Trello board with a list of resources and goals available at [trello.com/b/MwD8xuz3/umbraco-accessibility](https://trello.com/b/MwD8xuz3/umbraco-accessibility)**

MMasey:
To make things easier I've moved the spreadsheet over to here 😄

![image](https://user-images.githubusercontent.com/6573613/56197140-19db2600-6030-11e9-9f91-e2e7f770b1f3.png) |  | Umbraco V8 Accessibility Audit  |   |  
-- | -- | -- | -- | --
Issue | Level | Description | Area | PR Link
1 | Level AA : 2.4.7 - Unclear focus area | Tab states (focus) is insufficient or missing | General |  
2 | Level A : 2.4.1 - Skip links | No "skip navigation" links present | General |  
3 | Level AA : 1.4.4 - Resizing text | Fonts do not resize using the browser font size settings | General |  
4 | Level AA : 2.4.7 - Unclear focus area | Bottom publishing options ('Preview', 'Save', 'Save and publish') have insufficient focus (tab) state | General |  
[...]

PR #5760, opened 2019-07-01
V8: Accessibility improvements for page name entry

RachBreeze:
This PR is a resubmission of #5585

It aims to fix 43, 69, 75, 81 from the from the A & AA accessibility issue list #5277

**Creating a content item**
When creating a new content type there is screen reader only text which says:
"Create" + content type name

The concept is that it wkors as per Microsoft Teams, here when a user creates a folder they see:

image

**Editing a content item**
When editing an existing content item there is screen reader only text which says
"Edit" + content name

**Name label**
The "name" text box also now now has a hidden label which is "content type name" + Name

**Placeholder**
The placeholder in the name textbox remains the same as prior to this change, as placeholder text provides guidance for many users, placeholder text is not a replacement for labels.

_Language amends_
The language files now have accessibility specific place holders these have been added to en.xml and en_us.xml only


emmaburstow:
Hi Rachel,

Thanks ever so much for the work. We'll take a look and let you know if we need anything from you.

Emma

nul800sebastiaan:
Hey @RachBreeze! Thanks very much for the help here, I'll test it soon.

Just as a heads up: I have reverted most of your changes in the language files since it was impossible to figure out what actually changed. In the future, try to see in your git client, before you commit changes if all the whitespace has changed. It would be good to only commit the few lines that have actually changed so that the diff is readable. 👍

nul800sebastiaan:
Works great! 🔉 "Edit Home" and 🔉 "Create ContentPage" - cool stuff!

Thanks very much @RachBreeze - well done! 👍👍

nul800sebastiaan:
Cherry picked into 8.1.3 (9cab7894256507ec30a43f4d48fafb72cab5bddb)

PR #5741, opened 2019-06-28
Merging 7.15 - 8.1

Shazwazza:
A lot of manual merging with new features.

### Testing

* Make sure all tests pass (and JS ones)
* Make sure all pickers works correctly and adhere to the security settings of a user that has custom start nodes for both content and media. Create data types for each of these and do not select the option to bypass start nodes and test:
   * Media picker + Mini List View + Picker Search
   * Content picker + Mini List View + Picker Search
   * Multi url picker + Mini List View + Picker Search
   * RTE + Media picker + Mini List View + Picker Search
   * RTE + URL picker + Mini List View + Picker Search
   * Grid + RTE + Media picker + Mini List View + Picker Search
   * Grid + RTE + URL picker + Mini List View + Picker Search
* Then for each of the above data types, enable the bypass start nodes and test
* Check over the code for anything obvious

PR #5764, opened 2019-07-02
V8: Bugfixes regarding the Bypass Security feature after merge from v7

bergmania:
Related to: https://github.com/umbraco/Umbraco-CMS/pull/5741

Fixes:
- Use serverside logic to get anchors
- Fix when changing the linkpicker from content into media.
  - https://github.com/umbraco/Umbraco-CMS/pull/5764/files#diff-0e587028a6a3c1bc9d2693f8e90fcb29R186
- Fixes issue with Multinode Tree picker (for members)
  - Could not create the datatype, due unable to deserialize `-1` into a `Udi`
    - https://github.com/umbraco/Umbraco-CMS/pull/5764/files#diff-bb36e75fe7781ff23dd7166d22b12b84R1119
  - Null reference exception when navigating the "All members" tree
    - https://github.com/umbraco/Umbraco-CMS/pull/5764/files#diff-36b347bdd0bb79cfc8b8e4732df94a15R624


Shazwazza:
@bergmania comments below

> Fix when changing the linkpicker from content into media.
https://github.com/umbraco/Umbraco-CMS/pull/5764/files#diff-0e587028a6a3c1bc9d2693f8e90fcb29R186

awesome, tested and works

> Could not create the datatype, due unable to deserialize -1 into a Udi
https://github.com/umbraco/Umbraco-CMS/pull/5764/files#diff-bb36e75fe7781ff23dd7166d22b12b84R1119

Have adjusted this fix, all that was needed is to set the $scope.model.value.id to 'null'. The serialization error that occurs is because on the server side it is trying to deserialize to `Umbraco.Web.PropertyEditors.MultiNodePickerConfigurationTreeSource` expecting a UDI instead of an INT for ID. This of course means that the member picker for MNTP was never working in 8.x until now. IIRC the reason for resetting these values on `formSubmitting` is if the developer was selecting media/content and start nodes and then changed to member, i think this is just sanitizing the value before being sent to the server as a 'just in case' type of thing so a media/content start node id isn't sent.

tested and is working

> Null reference exception when navigating the "All members" tree
https://github.com/umbraco/Umbraco-CMS/pull/5764/files#diff-36b347bdd0bb79cfc8b8e4732df94a15R624

yup great

Issue #5761, opened 2019-07-02
V8: Mandatory colour picker can't be saved


AndyButland:
To replicate:

- Create a data type based on the "Color Picker"
- Add a few colours
- Create a property on a document type using the new data type
- Make it mandatory
- Try entering content based on the document type
- Even if a colour value is selected, the property is marked as invalid

![image](https://user-images.githubusercontent.com/1993459/60488195-a1d0e380-9ca1-11e9-927f-441284a6a1ab.png)

Why?  Not quite figured it out, but having added a few `console.log()`s it seems that the `val-property-validator` referenced in `colorpicker.html` is firing only when the content is loaded, and not re-firing when the value is changed or the save & publish button it clicked.  So doesn't seem that this is working as intended.

kjac:
By the looks of it the color picker needs a bit of cleaning and redo. I can replicate and I _think_ I have a solution.

kjac:
PR in #5765

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/5765

Cherry picked for 8.1.5 32941370065fcbea6857ad9c68a16a93c839812a

PR #5765, opened 2019-07-02
V8: Make the color picker mandatory validation work

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5761

### Description

This PR makes the color picker mandatory validation work again - see #5761 for more details.

While I was at it I also did a bit of cleaning up on the color picker controller.

With this PR applied, the color picker behaves like this when set to mandatory:

mandatory-color-picker

There are no changes in behavior when the color picker is _not_ set to mandatory.


emmaburstow:
Great work @kjac!

We'll take a look at this and let you know if we need anything.

Em

kjac:
PR updated with latest v8/dev

nul800sebastiaan:
Sweet! Thanks @kjac !

Issue #5671, opened 2019-06-16
Both view and content models are PureLive, with different versions error when making change to doctype


bkclerke:
[Forum thread for reference](https://our.umbraco.com/forum/umbraco-8/97731-error-causing-the-application-to-restart-after-a-doctype-change).

_I've tried this on v8.0.1 and v8.0.2 and am able to replicate the issue. When making changes to a doctype and refreshing the front end of the website I am receiving the following error:_

Cannot bind source content type Umbraco.Web.PublishedModels.PcGridLayout to model type Umbraco.Web.PublishedModels.PcGridLayout. Both view and content models are PureLive, with different versions. The application is in an unstable state and is going to be restarted. The application is restarting now.

**Description:** An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code.   

**Exception Details:** Umbraco.Web.Mvc.ModelBindingException: Cannot bind source content type Umbraco.Web.PublishedModels.PcGridLayout to model type Umbraco.Web.PublishedModels.PcGridLayout. Both view and content models are PureLive, with different versions. The application is in an unstable state and is going to be restarted. The application is restarting now.

_I've been able to fix the error by waiting for the application to restart or touching the web.config before I load the front end of the site._

Also, I'm curious if this might be related to [fix this issue too](https://our.umbraco.com/forum/umbraco-8/97235-possible-caching-error-umbracowebpublishedmodelsimage#comment-308507)?

nul800sebastiaan:
Thanks @bkclerke ! We have done work in the past to make this error at least detailed and a bit more friendly and we thought it should not happen any more but apparently there's still a race condition where this gets triggered. 
Is there a way you can easily reproduce this error?

bkclerke:
Yes! I just setup a local instance of v8.0.2. It has no starterkit and using SQL CE to just get it up quickly. I created a homepage doctype with a template. Put 2 test properties on it. No problems with the front end of the site loading when I make changes with only 1 doctype in there. When I created a second doctype and saved it (1 test property), i refreshed the front end of the site and I see the error:

_Cannot bind source content type Umbraco.Web.PublishedModels.Home to model type Umbraco.Web.PublishedModels.Home. Both view and content models are PureLive, with different versions. The application is in an unstable state and is going to be restarted. The application is restarting now._

[...]

PR #5767, opened 2019-07-02
V8: Bugfix for Modelsbuilder models with different versions error

bergmania:
Fixes: https://github.com/umbraco/Umbraco-CMS/issues/5671

If modelsbuilder is in live mode, we need to clear cache for all document types when at least one is cleared. This is due to modelsbuilder updating all the models to different versions.

Hopefully we can find a better solution, but this atleast resolves the issue.


zpqrtbnk:
so... is the issue in the end... that you end up having models of different versions in the same cache? then the issue is in the cache itself, which would not refresh all models... need to look at the code... but this fix... would just be working around the issue?

src/Umbraco.Web/PublishedCache/NuCache/ContentStore.cs
```diff
@@ -352,7 +359,7 @@ public void UpdateContentTypes(IEnumerable<int> removedIds, IEnumerable<IPublish
                     var node = link.Value;
                     if (node == null) continue;
                     var contentTypeId = node.ContentType.Id;
-                    if (removedIdsA.Contains(contentTypeId)) removedContentTypeNodes.Add(node.Id);
+                    if (removedIds.Contains(contentTypeId)) removedContentTypeNodes.Add(node.Id);
```
bergmania: Possible null reference exception
```suggestion
                    if (removedIdsA.Contains(contentTypeId)) removedContentTypeNodes.Add(node.Id);
```

PR #5783, opened 2019-07-03
V8: Hide the tree picker search results automatically when removing the search term

kjac:
### Prerequisites

### Description

If I search for something in a tree picker and I don't find what I'm looking for ([sorry couldn't help myself](https://www.youtube.com/watch?v=e3-5YC_oHjE)), I would expect the "search mode" to end when I delete the search term. As it is right now, I explicitly have to click the back button in the search field to exit "search mode":

tree-search-autohide-before

With this PR, the picker automatically exits "search mode" when the search term is deleted:

tree-search-autohide-after


emmaburstow:
Kenn, kenn, kenn,

Thanks for the work and the meticulous documentation as usual. We'll look at the work and let you know if we need you,

Em

kjac:
PR updated with latest v8/dev

nul800sebastiaan:
Thanks @kjac ! 👍

PR #5784, opened 2019-07-04
V8: Make log naming and colors consistent between Log Search and Log Overview

### LINKED ISSUE #5768: U8: Log Viewer pie chart has the wrong verbiage.
Not a big deal but ...
Settings -> Log Viewer -> Right side shows Number of errors and a pie chart of error counts.
There is a label called "Critical" this is incorrect. It should be labeled "Fatal". 

You can see this when you click "All Logs". Also ILogger doesn't have the ability to log "critical"

UmbracoMislabel

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #5768

### Description

I've fixed up the verbiage between the chart and the search view to use Fatal in both locations, including in the saved searches.
 
Also tweaked the color palette a little as it was unclear using the same color for both Error and Fatal, since they mean different things. Palette now looks like this: 

log-levels-chart

log-levels-dropdown

Have stuck with the existing color options for badges, in ascending order of importance, hence green for info as everything logged to information should NOT be a warning, error or fatality.

Made some changes to the search controller to use the existing `logLevels` variable to map colors onto the logItems, where we previously had a switch statement - consolidating that means there's now only one place in the controller to define the log colors so should make future changes easier.

Addendum: The datepicker was doing strange things when setting a range. Rather than formatting the displayed date `2019-07-03 to 2019-07-04` as it does on page load, it was `2019-07-03,2019-07-04`. Thought it might have been related to unbinding the element in an ng-if, but nope. The call to `logViewerResource.canViewLogs` in the preflight check function is somehow responsible, so resetting the date value in the preflight callback fixes it - I've done that by passing the datepicker instance into `preflightCheck` when that function is called in the datepicker change function, then conditionally resetting the date only when the instance exists. Seems to do the trick.

---
_This item has been added to our backlog [AB#1689](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/1689)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js
@@ -43,14 +43,20 @@
         vm.searchLogQuery = searchLogQuery;
         vm.findMessageTemplate = findMessageTemplate;
         
-        function preFlightCheck(){
-            vm.loading = true;
+        function preFlightCheck(instance){
+            vm.loading = true;          
+
+            
             //Do our pre-flight check (to see if we can view logs)
             //IE the log file is NOT too big such as 1GB & crash the site
             logViewerResource.canViewLogs(vm.startDate, vm.endDate).then(function(result){
                 vm.loading = false;
                 vm.canLoadLogs = result;
 
+                if (instance) {
+                    instance.setDate(vm.period);
+                }
+                
                 if(result){
                     //Can view logs - so initalise
                     init();
@@ -117,7 +123,7 @@
             $q.all([savedSearches, numOfErrors, logCounts, commonMsgs]).then(function() {
                 vm.loading = false;
             });
-
+            
             $timeout(function () {
                 navigationService.syncTree({ tree: "logViewer", path: "-1" });
             });
@@ -146,9 +152,9 @@
             mode: "range",
             maxDate: "today",
             conjunction: " to "
-        };
-        
-        vm.dateRangeChange = function(selectedDates) {
+        };        
+            
+        vm.dateRangeChange = function(selectedDates, dateStr, instance) {
             
             if(selectedDates.length > 0){
                 vm.startDate = selectedDates[0].toIsoDateString();
@@ -158,9 +164,9 @@
                     vm.period = [vm.startDate];
                 }else{
                     vm.period = [vm.startDate, vm.endDate];
-                }
-                
-                preFlightCheck();
+                }                
+                                
+                preFlightCheck(instance);
             }
 
         }
--- a/src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js
@@ -112,7 +112,7 @@
 
                 for (let [key, value] of Object.entries(data)) {
                     const index = vm.logTypeLabels.findIndex(x => key.startsWith(x));
-                    if (index > -1 && index < vm.logTypeData.length) {
+                    if (index > -1) {
                         vm.logTypeData[index] = value;
                     }                
                 }
--- a/src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js
@@ -12,9 +12,9 @@
         vm.commonLogMessagesCount = 10;
 
         // ChartJS Options - for count/overview of log distribution
-        vm.logTypeLabels = ["Info", "Debug", "Warning", "Error", "Critical"];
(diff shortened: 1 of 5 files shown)


nathanwoulfe:
Likely overlap with #5788

nathanwoulfe:
I've merged Andy's changes from #5788 - these are the latest commit on this branch and resolve the following:

 - Fixes issue #5768 (wrong label for "Fatal" errors on the pie chart).
 - Made the title variable so it either says "Log Overview For Today", "Log Overview For Selected Date" or "Log Overview For Selected Time Period" as appropriate (previously it said the former even if you'd adjusted the dates you were looking at.
 - Fixed up the selected date period display so the conjunction "to" is used between the dates when the selected time period is changed, and the query-string is updated to match the current search.
 - Made some minor code tidying updates.

#5788 can be closed since everything is now in the one place. 

FYI @AndyButland @nul800sebastiaan

AndyButland:
Thanks @nathanwoulfe for sorting out combining these.  I've closed the other PR now.

nathanwoulfe:
No worries @AndyButland, I'm just pleased I managed to git stuff and not set the world on fire...

src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js
```diff
@@ -74,7 +81,7 @@
                         "query": "Not(@Level='Verbose') and Not(@Level='Debug')"
                     },
                     {
-                        "name": "Find all logs that has an exception property (Warning, Error & Critical with Exceptions)",
+                        "name": "Find all logs that has an exception property (Warning, Error & Fatal with Exceptions)",
```
warrenbuckley: @nathanwoulfe or @AndyButland 
Suggest that the saved JS config file on disk of daved searches is updated as well please `src/Umbraco.Web.UI/config/logviewer.searches.config.js`

src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js
```diff
@@ -10,11 +10,12 @@
         vm.numberOfErrors = 0;
         vm.commonLogMessages = [];
         vm.commonLogMessagesCount = 10;
+        vm.dateRangeLabel = "";
 
         // ChartJS Options - for count/overview of log distribution
-        vm.logTypeLabels = ["Info", "Debug", "Warning", "Error", "Critical"];
+        vm.logTypeLabels = ["Debug", "Info", "Warning", "Error", "Fatal"];
```
warrenbuckley: If you really want all log levels to be displayed there is a `Verbose` level too in Serilog however I am not sure how useful or not that is (as doubt many really set the log level to that)

So it's
* Verbose
* Debug
* Information
* Warning
* Error
* Fatal

warrenbuckley:
At first glance the code all reads fine to me @nathanwoulfe & @AndyButland  I just need to test it out

Thumbs Up

src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js
nathanwoulfe: I've left verbose out, I agree it's not super useful.

src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js
nathanwoulfe: Done in latest commits

nathanwoulfe:
@warrenbuckley how bout this - I've updated to map the returned values onto the logTypeData array, so that the order of the log types won't matter - however they are ordered in the declared variable is how they'll be mapped.

Could go a step further and set the labels based on the values returned from the server, but we'd always need the array of colours, so it's probably not worth it.

warrenbuckley:
@nathanwoulfe code looks fine - will give it another re-test to be certain then it will get merged in :)

nathanwoulfe:
@warrenbuckley that's a bit strange - no console errors or anything, just an empty array? I'll re-check at my end, not sure why it wouldn't be working...

nathanwoulfe:
Issue was mine:
```javascript
const index = vm.logTypeLabels.findIndex(x => key.startsWith(x));
if (index > -1 && index < vm.logTypeData.length) { // err. no.
    vm.logTypeData[index] = value;
}    
```

vm.logTypeData is an empty array, so the condition was never met, and the value never added. Checking index > -1 is sufficient, as we know the array returned from the server server is the same length as the array of type labels, so no need to check the upper bound of the array.

I missed this yesterday due to having made the change to add the check, but browser was still hanging onto the previous, working version. All good now though.

warrenbuckley:
Yep works that time - ace stuff @nathanwoulfe

PR #4871, opened 2019-03-05
Added gulp-angular-embed-template

PeteDuncanson:
Fixes #4845 by embedded the angularjs templates for directives in the directives, cuts down on network traffic and increases speed to render

To test. Assuming you have already got a built version of the code up and running go take a look at the /src/umbraco.web.ui/umbraco/js/umbraco.directives.js file size, copy the file and rename it to something you'll remember (I just add "_pete" to the end of the file extension). Ensure the back office all runs as it should.

Change directory in a command prompt to "/src/umbraco.web.ui/" folder. 

Run `npm install` to get the new goodies in your build chain.

Run `gulp js` or `gulp dev` with the new config in this PR which will regenerate the javascript using the new bit of magic. Now compare the file size of the umbraco.directives.js file, it should have increased.

Hit up the back office again and ensure everything is working. Check that the number of network requests between the to checks has reduced as per the screen shots in the original issue. 

https://github.com/umbraco/Umbraco-CMS/issues/4845

PR #5785, opened 2019-07-04
V8: Add the AngularJS directive template url pathname into the template

### LINKED ISSUE #5047: V8: Add the AngularJS directive template url pathname into the template itself for easier finding
I had this as an issue on the old tracker. Currently if you are hacking around in the UI and need to find where a template lives so you can edit it its very difficult to work out where to go find it. I'd like it if in the HTML itself we could write out the filename/path to the template file in the source. That way you can very quickly find where the template lives and you can shorten the development time and focus on the actual problem.

We could inject it as a data-path attribute on the top most element in the template or a HTML comment but I'm not sure if that will be rendered out or not in dev tools. Something to check.

Two ways to do it. Manually go through and add them all, quickish but likely to break over time. Better would be to have gulp task that can just run through all the templates and add the pathname in automatically during the build, this is my preferred option I think and something to look into. If you do the gulp route then ensure it will run on gulp v4 as we are looking to upgrade to that (https://github.com/umbraco/Umbraco-CMS/issues/4846).

### LINKED ISSUE #6003: Remove AngularJS directive's template files from the compiled build
Now that directive templates are stored inline as part of the build (see #4845) we don't need to include the original template files in the zipped/nuget builds.

Ideally these should/could be removed to save some bytes in the output zip file.

Kicked off by this chat on twitter: https://twitter.com/nathanwoulfe/status/1154711340323201025

nathanwoulfe:
… view as part of the build

### Prerequisites

If there's an existing issue for this PR then this fixes #5047 and fixes #6003 

### Description

Changes here add a step to the `views` gulp task in which the view being copied is modified to add a `data-filepath` attribute to the root node (or an HTML comment if the view happens to be empty, which some are).

Changes can be verified by inspecting the backoffice and searching for 'filepath', and doing the same in directives.js.

filepath

These changes don't introduce any functional changes, they just add the new attribute which will be really really helpful for backoffice devs. Changes do introduce a couple of new NPM packages to the build chain, being gulp-modify-file and jsdom, both as dev dependencies.

The way the current build is run, tasks are run in parallel where possible, but that needed to change as the `js` task is now dependent on the output of `views` - since the directive templates are inlined (#4871), in the `js` task, we need to reference the files output by `views`, from their final location.

Changing `views`  to run in series doesn't add any noticeable time to the build, but not entirely sure that referencing the files in a different project is the best approach - open to suggestions. Could also chop a bit more off the file path too.

Updates to include a solution for #6003 are explained [down further](https://github.com/umbraco/Umbraco-CMS/pull/5785#issuecomment-517539524)


emmaburstow:
Hiiiiiya Nathan!

Thanks for the work - we'll get a look at it and ask if we've any questions.

Em

nathanwoulfe:
Picking up on @PeteDuncanson's suggestion in #6003, I've updated this to clean out the physical template files when the template content has been inlined into the respective directives. Why? Avoids shipping a stack of html files that aren't doing anything - it's not a huge saving size-wise, but removes 110 files and 22 directories from /umbraco/views (on build).

How does it work? It extends on (and depends on) the previous changes I made to add the file path as an attribute in the template. With that in place, I've added a new 'clean' task to the Gulp config, which runs at the end of the build process. First, it grabs all the compile JS files, and extracts the value from any data-filepath attributes in the file content (which will only appear in an inline template string, which has been embedded as part of the JS processing tasks). Armed with a collection of path segments, I'm then deleting the files from the output folder, because if their template string exists in the JS, the physical file isn't needed.

Next, a separate task recursively deletes empty directories from /umbraco/views, since empty directories are empty.

The clean up tasks only run on `gulp build` - for `dev` and `fastdev` there's no benefit in removing the files, so left them in place. Also makes sense that dev environment would make more use of the data-filepath attribute, so having the files at that location is sensible.

👍

PR #5788, opened 2019-07-04
V8: Minor log viewer improvements

AndyButland:
This PR contains a few minor improvements to the log viewer:

- Fixes issue #5768 (wrong label for "Fatal" errors on the pie chart).
 - Made the title variable so it either says "Log Overview For Today", "Log Overview For Selected Date" or "Log Overview For Selected Time Period" as appropriate (previously it said the former even if you'd adjusted the dates you were looking at.
- Fixed up the selected date period display so the conjunction "to" is used between the dates when the selected time period is changed, and the query-string is updated to match the current search.
- Made some minor code tidying updates.


nathanwoulfe:
I've merged this across into #5784 so we have no more issue with duplicate PRs

AndyButland:
No loner required as work now combined into PR #5784.

Issue #5791, opened 2019-07-04
v8: Rollback clear image cropper and grid


bjarnef:
In Umbraco v8 with latest changes in `v8/dev` branch, when I select rollback and select an older version, but without clicking submit button (rollback), but just close the dialog and go back to the content tab the image cropper and grid no longer shows any values.

But when I refresh the page the values image cropper image and grid data are shown.

![2019-07-04_22-18-07](https://user-images.githubusercontent.com/2919859/60686591-18403200-9eaa-11e9-9043-1162e29c0a4e.gif)

PerplexDaniel:
Yes, excellent @bjarnef! I was about to make a new issue and PR with essentially the same but thought "let's see if someone else already did this" ;-)

This applies to all property editors that have their value stored as JSON, so it's quite a big deal.
It it still present in 8.6.1 and [the latest `v8/contrib`](https://github.com/umbraco/Umbraco-CMS/commit/b2a9e17c90ab5c9d2bfa7c7b6d2e738a152073e0).

@nul800sebastiaan Can we please merge the linked PR into v8 soon-ish? Mainly it's about [this commit](https://github.com/umbraco/Umbraco-CMS/pull/5792/commits/b20737a9af76e92a511417d7166c903deee4a191) which does an `angular.copy` of the original property. There are 4 other commits in there with useful stuff but the core bug is fixed by just doing an `angular.copy`, which therefore could also be cherry-picked instead.

Full commit of @bjarnef with the fix is below (just to show it's tiny):

```javascript
// copy existing properties, so it doesn't manipulate existing properties on page
oldProperty = angular.copy(oldProperty);
property = angular.copy(property);
```
Considering #7950 we would just have to change that to `Utilities.copy` instead and we're good to go.

bjarnef:
Oh, it has been a while since I submitted that PR.. almost forgot about it 🙈 

It seems `Utilities.copy` still use `angular.copy()` under the hood, but sure it would be great to use the utility function here instead.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/utilities.js#L21

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/5792

nul800sebastiaan:
Ah forgot to update to Utilities.copy - done in 9be4775acfe63d53a8a4bc1557ea7a5268301138

PR #5792, opened 2019-07-04
v8: Fix disappearing image cropper and grid content when using rollback

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5791

### Description
It seems the rollback manipulated to original tab properties in background, which broke the image cropper and grid content. Now it create a copy of the property and old property before using it further in code.

It also shows the loading indicator while fetching a different rollback version.

A few notes to this:
- It seems the TinyMCE in the grid content sometimes collapse after changing rollback version via dropdown (but not submitting) or an issue when changing tab in the header.
- When doing a rollback, should the page then refresh or can we show the updated properties? At the moment it shows the current version (not the version roll back to), but the properties are updated after refreshing the page.


emmaburstow:
Hey @bjarnef 

Also, thanks for this one! We'll let you know if we need you,

Em

nul800sebastiaan:
Well, that was almost a year delayed 🙈 but excellent fix @bjarnef - thanks very much! 👍

Issue #5771, opened 2019-07-02
MNTP: Icons change when searching 


dnwhte:
### Bug summary

Doc type icons are changed when searching in MNTP.

### Specifics

Umbraco 7.14

### Steps to reproduce

Use the search feature in a MNTP data type. Icons will change.

### Expected result

Doc types icons should remain the same when searching.

### Actual result

When searching doc types icons are changed to an image type.

---
![6](https://user-images.githubusercontent.com/636470/60542187-2e8e9800-9cc8-11e9-91a0-503ae48e0ef7.PNG)

![5](https://user-images.githubusercontent.com/636470/60542192-31898880-9cc8-11e9-969e-ea4edd64ab92.PNG)

nul800sebastiaan:
Yep, I can reproduce this one, even on the latest v7.15 commits.

umbrabot:
Hi @dnwhte,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

kjac:
Yeah... I can reproduce this as well. Found the bug, will post PR soon.

kjac:
PR in #5794 (and V8 PR in #5795)

@dnwhte you can work around this by re-assigning the media type icon (or another icon).

Myster:
Note: This bug also occurs when searching in the sidebar, not just in the MNTP

Myster:
The workaround suggested by @kjac works for me.
I only needed to change the icon for the 'File' media Type, and then I needed to rebuild the internal index in the examine management dashboard.

Myster:
Still an issue in 7.15.1

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/5794 and https://github.com/umbraco/Umbraco-CMS/pull/5795

Cherry picked for 7.15.3 d152cf076fa8138b706367a14f9dd58c9c33bea6
Cherry picked for 8.1.4 18c02987204f0d14c9a93029ef3a96e34369dbc4

PR #5794, opened 2019-07-06
V7: Fix media file icon in search results

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5771

### Description

See #5771 for steps to reproduce/test. 

The issue happens because the tree searcher assumes that the icon "icon-document" in a media search result means a bad mapping from the search result; it's the default icon used by the mapper. However, "icon-document" is also used for the default file media type, so it doesn't have to be a mapping error.

With this PR applied, the search feature in the tree picker works as expected for file results:

mntp-search-file-icon-changes


emmaburstow:
Hi @kjac 

Lovely work here, by the looks of it. We'll holler if we need anything more from you.

Em

nul800sebastiaan:
Lovely, thanks @kjac ! 👍

PR #5795, opened 2019-07-06
V8: Ensure correct icon for file types in search results

kjac:
### Prerequisites

### Description

_This is the V8-ification of #5794_

When using the global search, media of the default "File" type are displayed with the wrong icon:

image

image

This issue happens because the tree searcher assumes that the icon "icon-document" in a media search result means a bad mapping from the search result; it's the default icon used by the mapper. However, "icon-document" is also used for the default file media type, so it doesn't have to be a mapping error.

With this PR applied, the search result icons behave as expected:
 
image


emmaburstow:
Oh hey @kjac 

Fancy seeing you here ;) Thanks for this work. Good spot too! 

Em

kjac:
PR updated with latest v8/dev

nul800sebastiaan:
And thanks for this one too @kjac !

PR #5796, opened 2019-07-06
V8: Ensure search result icons for all searchable trees

kjac:
### Prerequisites

### Description

For some odd reason there are no icons displayed in the search results for content types, media types, data types or templates:

search-result-icons-before

While the icons will always be the same for these trees, I still think it makes sense to show them, because:

1. It looks prettier 😄 and more consistent.
2. The icons give context to large result sets when you've scrolled past the result group separator.

So! This PR adds icons to search results for content types, media types, data types and templates:

search-result-icons-after

As you can tell I have done a bit of refactoring, since all of these trees basically had the same chunk of code in all implementations.


emmaburstow:
Ahhh thanks for fixing this @kjac 👍 We'll ask if we need anything,

Em

kjac:
PR updated with latest v8/dev

nathanwoulfe:
Just had a quick look at this, all works as @kjac described - did have a bit of a dig to see where the specific icons are coming from, but can't find em. Regardless, I think the defaults looks better than having none.

PR #5797, opened 2019-07-06
V8: Make sure search doesn't close when clicking "Clear"

kjac:
### Prerequisites

### Description

The search dialog disappears when you click "Clear":

search-result-clear-before

This happens because the `on-outside-click` directive can't find the "Clear"-button, simply because the "Clear"-button is no longer part of the DOM when the directive is executed, due to the use of `ng-if`.

By swapping `ng-if` with `ng-show` we get the correct behavior:

search-result-clear-after


emmaburstow:
Hey @kjac 

Excellent fix! We'll look at the work, as usual, and let you know if we need anything.

Em

kjac:
PR updated with latest v8/dev

nul800sebastiaan:
That works better, great stuff! 👍

PR #5801, opened 2019-07-08
Remove last casting to LuceneIndex

bielu:
### Description
It is really simple PR, which will allow to fully switch out all indexes from examine to any kind of provider.
As we discussed with @Shazwazza I switch to IIndex and remove CommitCount, because of not all type of indexer will use that.


Shazwazza:
Awesome! Thanks a ton :)

nul800sebastiaan:
This was unfortunately not cherry picked to the 8.1.1 release, it will be available in 8.1.5 instead: (cherry pick: 255bcfbc77e31db7bc49c87cd2f34581c4f77bbd )

PR #5821, opened 2019-07-09
v8: Refactor checkered background

bjarnef:
### Prerequisites

### Description
This PR remove the older checked background bitmap image and replace this with vector SVG, which will look sharper on retina screens on mobile devices or high resolution desktop screens.
This was inspired from this https://codepen.io/pascalvgaal/pen/jPXPNP but the principle is the same as these examples of repeatable background using SVG: https://www.heropatterns.com 

SVG background images has been supported for a while now https://caniuse.com/#feat=svg-css

image

image

image

image

image

This also make it easier to change color, contrast and opacity of the checkered background using the mixin `checkeredBackground`. When no parameter is specified it just use the default values.

**Example 1**
image

```
.checkeredBackground(#eee, #f5c1bc, 1);
```

**Example 2**
image

```
.checkeredBackground(#eee, #1b264f, 0.5);
```

**Example 3**
image
```
.checkeredBackground(#eee, #1b264f, 0.5);
```

Note in these examples the media grid card itself has an opacity on 50%.

Example 3 without opacity.

image


bjarnef:
Note that in IE11 it doesn't work well, but Umbraco 8 has also dropped support for IE11 and `object-fit` which is used in some part of the styling isn't supported in IE11  anyway.
https://caniuse.com/#search=object-fit

bjarnef:
The difference can also be reviewed by zooming in a lot in the browser. 🤣 

**Before**
image

**After**
image

bjarnef:
What do you think @nul800sebastiaan ? :)

bjarnef:
@nul800sebastiaan any thoughts on this? :)

nul800sebastiaan:
Super sweet! Thanks @bjarnef and sorry for the long wait!

Issue #5808, opened 2019-07-09
v8: Change password from user dialog doesn't work


bjarnef:
I just tried the latest release Umbraco v8.1.0 and it seems the "change password" from user dialog doesn't work.

Reproduction
------------

### Bug summary
Open the user dialog/overlay via the avatar and click the "change password" button.
The dialog close and the change password form as previous is not shown.

### Specifics

Umbraco v8.1.0
Google Chrome Version 75.0.3770.100

### Expected result
Clicking the "change password" button in the user dialog/overlay should show the change password form.

### Actual result
This dialog is just closing.

![image](https://user-images.githubusercontent.com/2919859/60875942-35bf2400-a23b-11e9-9c85-7a4c02d986f9.png)

Shazwazza:
The problem with several all of these types of bugs was with this PR https://github.com/umbraco/Umbraco-CMS/pull/5609

I have reverted these changes https://github.com/umbraco/Umbraco-CMS/pull/5609#issuecomment-510709482

So all of these things work again.

PR #5609, opened 2019-06-07
V8: Don't show multiple open menus (take two)

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description

This is a redo of #5563, this time taking into account that the dropdown toggle buttons (i.e. the little "popup-arrow" next to "Save and publish") should still work (ref. https://github.com/umbraco/Umbraco-CMS/pull/5563#issuecomment-499740834).

The functionality of this PR remains the same as #5563, only with the addition that the dropdown toggle buttons are allowed to leave other menus hanging; it looks something like this:

multiple-open-menus-take-two

/cc @Shazwazza

PR #5826, opened 2019-07-09
v8: Fix toggle change password in user dialog

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5808

### Description
When clicking the change password button this parent `.umb-overlay` wasn't found, but it was found when clicking the whitespace in the dialog outside the button.
https://github.com/umbraco/Umbraco-CMS/blob/1a437c1c6abe77ef5edd337bbbdb169edde5f41e/src/Umbraco.Web.UI.Client/src/common/directives/components/events/events.directive.js#L111

I guess it in some way conflict with the `umbButton` event.
https://github.com/umbraco/Umbraco-CMS/blob/040ebfe9dd7d7053ccf4393f8a83c11fc3866b69/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbutton.directive.js#L193

To fix this issue and without changing too much in `umbButton` component, I have stopped event bubbling when clicking the change password button, which toggle the change password fields.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/events/events.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/events/events.directive.js
@@ -95,7 +95,7 @@ angular.module('umbraco.directives')
         };
     })
 
-    .directive('onOutsideClick', function ($timeout, angularHelper) {
+    .directive('onOutsideClick', function ($timeout) {
         return function (scope, element, attrs) {
 
             var eventBindings = [];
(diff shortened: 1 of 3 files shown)


Shazwazza:
The problem with all of these bugs was with this PR https://github.com/umbraco/Umbraco-CMS/pull/5609

I have reverted these changes https://github.com/umbraco/Umbraco-CMS/pull/5609#issuecomment-510709482

So all of these things work again. I will close this PR but we might need to revisit some things once the originally intended issue from the #5609 PR is again resolved.

PR #5834, opened 2019-07-10
5732 hide grid reorder when scrolling - fix issues in listviews

nathanwoulfe:
### Prerequisites

### Description

This fixes a couple of issues in list views introduced by a couple of my previous CSS changes, mainly due to setting position:relative on .umb-property-editor. Doing that messed with the positioning of the umb-empty-state element as it's absolutely positioned, so the .umb-property-editor became the first positioned ancestor, and short story, busted it. Also had issues in the media section where the sticky bar was immediately stickied due to being at the top of its container, which wasn't good.

My changes here modify how the sticky sentinel elements are added, to remove the need for absolute positioning, which in turn removes the need for adding pos:rel on .umb-property-editor. Listviews/dashboards use a different wrapper to the grid editor, so have added a selector to ensure these are observed correctly too.

Have also added some additional null checks just to be safe.

To verify these changes, check:
- the 'Your recycle bin is empty' text is centered correctly
- the media section list view doesn't overlap the dropzone
- grid works as before


nathanwoulfe:
Don't know why this has found the previous commits - I pulled upstream into my fork, merged local v8/dev into my feature branch, from where most of those commits originated. Thought they'd have all been squashed. I love git.

Shazwazza:
Did you push to the same branch as your last PR? You should create new branches for each PR. It also probably didn't detect the merge because your last PR was odd - and it was submitted as a `patch` which i had to pull in differently than most.

nathanwoulfe:
@Shazwazza that's probably it - this would have been pushed to the same branch. I can kill this PR and cherry pick the commits into a new branch if that's preferred?

Shazwazza:
yep i think that will be best :) Please @ mention me in the new one.

nathanwoulfe:
Killing this mess, changes are in #5918, with an appropriate number of commits... FYI @Shazwazza

Issue #5840, opened 2019-07-11
V8: RelatedLinks error - migration from V7.15 - V8.1


dinc5150:
When I upgraded from V7.15.0 to V8.1.0 I got a null reference exception error during the migration process. 

Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

### Bug summary

I upgraded a website from V7.15.0 to V8.1.0 with existing content. Some of the pages had no related links picked in the RelatedLinks Property

### Steps to reproduce
Create content with a related Links Picker and dont pick any links. Then Upgrade to V8+

### Expected result

Migrated Content!

### Actual result

Error during Install Step:
```
The database failed to upgrade. ERROR: The database configuration failed with the following message: Object reference not set to an instance of an object. Please check log file for additional information (can be found in '/App_Data/Logs/')
```

## Note:
I have a pull request on its way

Shazwazza:
Duplicate https://github.com/umbraco/Umbraco-CMS/issues/5903#issuecomment-513113360

PR #5841, opened 2019-07-11
#5840 - RelatedLinks Migration check for null Value before .ToSting()

dinc5150:
### Prerequisites

If there's an existing issue for this PR then this fixes: #5840 

During the migration from V7 to V8, the relatedLinks picker gets migrated, but fails if the relatedLinks value is null (No selected links). To reproduce, create content with a related links picker and dont pick any links. Then upgrade to V8.

### Problem:
When calling .Tostring() on the RelatedLinks picker during migration, it doesn't check for a null value before calling .Tostring(). The following line checks string.IsNullOrWhiteSpace(), so handles the migration correctly if we parse the value as null.

I have tested this on the website database which caused the original issue.


emmaburstow:
Hey @dinc5150 

Thanks for the work here. One of us will take a look and shout if we need anything from you before we merge. 

Emma

nul800sebastiaan:
Hi there @dinc5150 - So sorry for the delay on this one! This is a great fix, thanks very much! ⭐️

PR #5862, opened 2019-07-12
Fixes #5789 - PublishedContentQuery Search always searches on culture

### LINKED ISSUE #5789: V8 Umbraco.ContentQuery.Search returns no results
### Bug summary
Using SK, executing Umbraco.ContentQuery.Search("Products", "ExternalIndex") returns no results, while using the ExaminManagement search in the backoffice returns 4 results. The simple ContentQuery.Search should return the same results as the backoffice search.

### Specifics
- Starter Kit (or any site, but the SK is consistent for everybody)
- V8/dev (also 8.0.2)
- Chrome (but this isn't a browser bug)
- Didn't check it in other browsers
- No error message, this is a delta between what is expected and what occurs

### Steps to reproduce

1.  Create new V8/dev site with SK
2.  Add to the Home.cshtml view the following:
```
    @{
        var easySearch = Umbraco.ContentQuery.Search("Products", "ExternalIndex");
        <div>Found @easySearch.Count() results the easy way.</div>
    }
```
3. Run and check out the site, it lists 0 results the easy way,

### Expected result
Search results are rendered when you perform the easy search using the simple one line method. I think a lot of devs will expect the simple 1 liner way of searching to just work as it did in V7.

### Actual result
No results are found

### Work around
Use the following code in your view for a simple Examine search instead:
```
        ExamineManager.Instance.TryGetIndex("ExternalIndex", out var index);
        var searcher = index.GetSearcher();
        var results = searcher.Search("Products", 100);
        <div>Found @results.Count() results the longer way</div>
```

### A little further digging
In PublishedContentQuery stick a break point on `if (culture.IsNullOrWhiteSpace())` and force execution of line `results = searcher.Search(term, count);` and you get 4 results (as expected). Normal execution runs down the else block which produces a radically different Lucene search. 
It seems that `culture = culture ?? _variationContextAccessor.VariationContext.Culture;` always returns a default culture (in my case en-US), so execution always flows into the else block.

---
_This item has been added to our backlog [AB#1697](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/1697)_

Shazwazza:
See #5789 

To test follow the steps provided in the issue and make sure the expected results work.

Have a look at the code changes - the reason why this problem occurred is because passing in 'null' for a culture should mean it only searches invariant but this was defaulting to `_variationContextAccessor.VariationContext.Culture;` which always returns a culture. Further to that, the method's optional parameter value was incorrect. A "*" is what should be the default indicating that all cultures will be searched (this is consistent with our other culture methods in Umbraco APIs). Specifying an explicit null will be an invariant search or specifying a culture will only search that culture.

### DIFF
--- a/src/Umbraco.Web/PublishedContentQuery.cs
+++ b/src/Umbraco.Web/PublishedContentQuery.cs
@@ -216,7 +216,7 @@ namespace Umbraco.Web
 
                 //get all index fields suffixed with the culture name supplied
                 var cultureFields = umbIndex.GetCultureAndInvariantFields(culture).ToArray();
-                var qry = searcher.CreateQuery().Field(UmbracoContentIndex.VariesByCultureFieldName, "y"); //must vary by culture
+                var qry = searcher.CreateQuery();
                 qry = qry.And().ManagedQuery(term, cultureFields);
                 results = qry.Execute(count);
             }
--- a/src/Umbraco.Web/PublishedContentQuery.cs
+++ b/src/Umbraco.Web/PublishedContentQuery.cs
@@ -215,7 +215,7 @@ namespace Umbraco.Web
                 //search only the specified culture
 
                 //get all index fields suffixed with the culture name supplied
-                var cultureFields = umbIndex.GetCultureFields(culture).ToArray();
+                var cultureFields = umbIndex.GetCultureAndInvariantFields(culture).ToArray();
                 var qry = searcher.CreateQuery().Field(UmbracoContentIndex.VariesByCultureFieldName, "y"); //must vary by culture
                 qry = qry.And().ManagedQuery(term, cultureFields);
                 results = qry.Execute(count);
--- a/src/Umbraco.Web/PublishedContentQuery.cs
+++ b/src/Umbraco.Web/PublishedContentQuery.cs
@@ -217,7 +217,7 @@ namespace Umbraco.Web
                 //get all index fields suffixed with the culture name supplied
                 var cultureFields = umbIndex.GetCultureAndInvariantFields(culture).ToArray();
                 var qry = searcher.CreateQuery();
-                qry = qry.And().ManagedQuery(term, cultureFields);
+                qry = qry.ManagedQuery(term, cultureFields);
                 results = qry.Execute(count);
             }
 
--- a/src/Umbraco.Web/PublishedContentQuery.cs
+++ b/src/Umbraco.Web/PublishedContentQuery.cs
@@ -20,14 +20,22 @@ namespace Umbraco.Web
     {
         private readonly IPublishedSnapshot _publishedSnapshot;
         private readonly IVariationContextAccessor _variationContextAccessor;
+        private readonly IExamineManager _examineManager;
+
+        [Obsolete("Use the constructor with all parameters instead")]
+        public PublishedContentQuery(IPublishedSnapshot publishedSnapshot, IVariationContextAccessor variationContextAccessor)
+            : this (publishedSnapshot, variationContextAccessor, ExamineManager.Instance)
+        {
+        }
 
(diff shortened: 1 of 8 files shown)


bergmania:
Hi @Shazwazza 

I have tested,
The `"*"` case works as expected, but I don't think it works as expected when I search for culture `en-US`.

I would expect when I search for culture `en-US` that the result had all the invariant items that matched also.

Shazwazza:
@bergmania i think i know what is going on, something overlooked for invariant props - was broken before too ;) Will fix up.

src/Umbraco.Web/PublishedContentQuery.cs
```diff
@@ -195,20 +195,29 @@ public IEnumerable<PublishedSearchResult> Search(string term, int skip, int take
...
+                results = qry.Execute(count);
+            }
             else
             {
+                //search only the specified culture
+
                 //get all index fields suffixed with the culture name supplied
-                var cultureFields = umbIndex.GetCultureFields(culture);
+                var cultureFields = umbIndex.GetCultureAndInvariantFields(culture).ToArray();
                 var qry = searcher.CreateQuery().Field(UmbracoContentIndex.VariesByCultureFieldName, "y"); //must vary by culture
```
bergmania:
```suggestion
                var qry = searcher.CreateQuery();
```

Sorry github do not support multiline suggestions.. But I think the issue is because we require the `Field(UmbracoContentIndex.VariesByCultureFieldName, "y")`

src/Umbraco.Web/PublishedContentQuery.cs
```diff
@@ -195,20 +195,29 @@ public IEnumerable<PublishedSearchResult> Search(string term, int skip, int take
...
-                qry = qry.And().ManagedQuery(term, cultureFields.ToArray());
+                qry = qry.And().ManagedQuery(term, cultureFields);
```
bergmania:
```suggestion
                qry = qry.ManagedQuery(term, cultureFields);
```

bergmania:
Preconditions:
- Starterkit installed
- Danish added as language
- Change home and product doctype to vary by culture

#### The results before the suggested changes:
Search for "Products" and culture `en-US`
- 'Home'

Search for "Products" and culture `da`
- ''

Search for "Products" and culture `*`
- 'Products'
- 'Todo list for the Starter Kit'

Search for "Products" and culture `null`
- 'Products'
- 'Todo list for the Starter Kit'

#### After changes:
Search for "Products" and culture `en-US`
- 'Home'
- 'Products'
- 'Todo list for the Starter Kit'

Search for "Products" and culture `da`
- ''
- 'Products'
- 'Todo list for the Starter Kit'

Search for "Products" and culture `*`
- 'Products'
- 'Todo list for the Starter Kit'

Search for "Products" and culture `null`
- 'Products'
- 'Todo list for the Starter Kit'

src/Umbraco.Web/PublishedContentQuery.cs
Shazwazza: @bergmania hrm 🤔 If we are searching for a culture surely we don't want to return invariant documents, only variant documents?

bergmania: I understand from a programmer perspective, but not from a user perspective. 
From a user perspective, a search for a given culture, is the same as navigating a tree in a given culture. And when I navigate the tree, I see both the invariant and variant items?

Shazwazza: That's really a fundamental change to how things have been designed though. In that sense, that would really come down to where you start searching from in the tree and what cultures you have assigned there which is not taken into account here at all.

Issue #5848, opened 2019-07-11
Media tree doesn't show folders in 7.15.0 any more for folders created before 7.8.0


nul800sebastiaan:
See for original report: https://our.umbraco.com/forum/using-umbraco-and-getting-started/98092-umbraco-715-media-folders-not-appearing-in-the-tree

In 7.8.0 we introduced a new table `cmsMedia` in which all media items get populated when they have a `mediaPath` - sadly, media items of type `Folder` do not have a path so they never got into the `cmsMedia` table during the upgrade.

This was no problem until we introduced this part of the query: 

https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Core/Persistence/Repositories/EntityRepository.cs#L373

An `INNER JOIN` on `cmsMedia` will exclude the old Folders from being return since they do not exist in `cmsMedia`. When changing this query to a `LEFT JOIN`, the media tree returns to normal.

---
_This item has been added to our backlog [AB#1714](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/1714)_

YodasMyDad:
Any idea when this will make it into 7.15.1 and when it will be released? Currently have a live site that no one can add new content on because cannot pick images for the content.

nul800sebastiaan:
A 7.15.1 release will have to wait a little longer.

If you need to fix it now, make sure to roll back to either 7.14.0, use the build output from the related PR.

MichaelNielsenDK:
I feel that this is a serious issue that needs to be fixed in a patch update or hotfix as soon as possible.

benjimedia:
This is a reasonably large problem for users. Perhaps HQ needs to highlight this issue on the 7.15.0 release page on Our ?

nul800sebastiaan:
Sure,  there you go: 😊

![image](https://user-images.githubusercontent.com/304656/61112332-e459a380-a48b-11e9-85d5-6396748760eb.png)

nul800sebastiaan:
@MichaelNielsenDK Obviously this is being fixed ASAP.

MichaelNielsenDK:
@nul800sebastiaan Awesome... 🤘🏻

nul800sebastiaan:
As a workaround, we can populate the `cmsMedia` table with some SQL. Note, this is mostly untested, make backups and copy the output of this script to make sure you can revert the INSERT statement if needed later

```
[...]

PR #5865, opened 2019-07-12
5848: changes innerjoin to left join, to allow folders in sql query

skttl:
#5848 

This one changes the InnerJoin to LeftJoin as advised in the issue.

I have tried this on my 7.15 solution and folders now show up again.


bergmania:
Tested and it fixes the issue as expected :)

Issue #5823, opened 2019-07-09
U8.x: umbracoUrlAlias comma separated list is taken literally.


bmargevicius:
Not sure if this is a bug. Feels like a display bug. 
When having a property umbracoUrlAlias as a textstring. Having comma separated routes are displayed and clicked through as literal and not as individual routes. This happens in v8.02 and v8.1. 

![UmbracoUrlAlias](https://user-images.githubusercontent.com/24920023/60913263-01f0f800-a255-11e9-9b16-39de0cdc5465.gif)

I expect it not to 404 and to create two separate lines in the info tab. manually typing /gold or /golden does work. So it's not the routing that's failing. It's the display of multiple routes on the info tab that's buggy.

nul800sebastiaan:
Hey @bmargevicius - `umbracoUrlAlias` supports a single alias only, so this is expected behavior. I would look into IIS rewrite rules if you need to redirect multiple URLs to a single content item.

bmargevicius:
Not according to documentation... 
**umbracoUrlAlias**
This property when created as a text string lets you provide a comma separated list of alternate full URL paths for the node. For example, if your URL was /some-category/some-page/content-node, by adding an umbracoUrlAlias of "flowers", a user can navigate to the node by simply going to /flowers. The URL alias remains in the browser address bar as a 'mask' over the real URL. You can also specify paths like "flowers/roses/red".
ref: https://our.umbraco.com/documentation/Reference/Routing/routing-properties

nul800sebastiaan:
You'll note that this documentation requires an update for v8 though. 😅

nul800sebastiaan:
Ah, we're for some reason not showing the banner on this page that this is not up-to-date documentation. That's unfortunate.

bmargevicius:
So it's not supported anymore? This is weird. Because the comma separated routes actually work.

ReVoid:
@nul800sebastiaan Really? It was a pretty useful feature to add many aliases to one page. Is it really necessary to delete the feature. 

I hoped alias would be improved in v8, like make it repeatable textstrings to be more obvious to user. 

:(

nul800sebastiaan:
Well, I'll go be ashamed in a corner. I would've sworn this feature was removed. But nope, all good! It's just the URLs that are displayed wrong on the info tab. My bad, forget I said anything! 🙈

We'd love some help getting the URLs to display correctly if you can. 

[...]

PR #5875, opened 2019-07-12
V8: Support multiple URL aliases in umbracoUrlAlias

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5823

### Description

Apparently it's possible to enter multiple URL aliases in `umbracoUrlAlias` by separating them with comma (*not* comma+space, though!). However, as #5823 points out, the backoffice doesn't really honor this scenario.

Turns out the `AliasUrlProvider` is the culprit. This PR updates it, so it can handle multiple URL aliases:

multiple-url-aliases


emmaburstow:
Hey Kenn,

Thanks for this. We'll shout if we need anything from you.

Em

ronaldbarendse:
@kjac What about the case when no domains are configured?

https://github.com/umbraco/Umbraco-CMS/blob/cc940f719dd92236fbe0e2d0d450fe53a37bd94b/src/Umbraco.Web/Routing/AliasUrlProvider.cs#L76-L93

kjac:
@ronaldbarendse 👍 good catch. I'll amend the PR.

kjac:
Done 👍

kjac:
PR updated with latest v8/dev

bmargevicius:
Did this PR get merged in? I really would like to have this feature restored..

ronaldbarendse:
@bmargevicius The PR is still open, so no...

@kjac It would be nice if the multiple textstring editor could also be used for this property, so you can see the URLs in a list and easily remove/add items. What happens when we retrieve the value using `node.Value<string[]>()`?

bmargevicius:
Thanks for the reply, I hope this gets pushed soon. I have a project in development sort of depending on this feature. I have a work around, but I'd rather not include my own route maker. 🎆Happy Friday the 13th everyone 🎆

nul800sebastiaan:
Works well, thanks very much @kjac and @ronaldbarendse for the double-check! 👍

nijdamb:
Hi there,

I'm sorry, as this is an old case but has there been any changes to possibly include the "repeatable text string" editor for this field?
With a comma separated list it's not very user friendly when you have more than 2 aliasses in the input.. especially if those aliasses aren't single words.
Having the "repeatable text string" editor work too would give a lot better overview on what Aliasses are set on a content node.

If I've missed any more recent cases for this I'm sorry, but this came up in Google :)

Issue #5867, opened 2019-07-12
8.1 Can't tab to property alias when defining properties in a doctype


skttl:
When defining a property in a doctype you used to be able to tab from the property label field to the alias field. Now it skips the alias field and goes to the description.

To reproduce:
Add a property to a doctype, set the label, then tab to the next field.

Expected result:
I expected the alias field to gain focus, as it did before.

Actual result:
The description field got focus, and I had to use my mouse (OH NO! :)) to get to the alias field.

nul800sebastiaan:
Uh oh, that seems like an oversight! 

I've marked this as "Up for grabs" so that you or someone else coming along could create a pull request for it.

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/5876

PR #5876, opened 2019-07-12
fix for #5867 - Enables tabbing to field-lock-toggler, and adds tab-focus styling

skttl:
This is a fix for #5867 

I changed the toggler to a button to enable tabbing, and keyboard clicking (for some reason I couldn't click the `<a>` by pressing space).

I also added the `.tabbing-active` style to the toggler.

I added some reset styling to the toggler to prevent it from looking like a button. Not sure if there already exists a utility class for that.

gif


emmaburstow:
Hey @skttl 

Thanks for the work and the lovely documenting of it too. We'll let you know if we have any questions or feedback.

Em

skttl:
Would love to have this one merged in :)

nul800sebastiaan:
Excellent, thanks @skttl ! I've added `aria-hidden="true"` to the icons as well, all good now!

Issue #5849, opened 2019-07-11
RTE link picker throw error after upgrade to 7.15.0


bjarnef:
After upgrading af site on Umbraco Cloud from 7.12.4 to 7.15.0 to RTE link picker seems to throw an `417 (Missing token)` error.

- Select some text in RTE
- Select a node in content tree, e.g. `My Page`. Previously the link picker then showed `/my-page` under `Link`, but now it shows something like this `/{localLink:umb://document/afcc3d4df40d4d1eb8f2c43fdc592b58}`
- Submit with the selected node
- When you then set the cursor at the link in the RTE and click link picker (edit), then it show the following error. `/umbraco/backoffice/UmbracoApi/Entity/GetUrlAndAnchors?id=umb://document/afcc3d4df40d4d1eb8f2c43fdc592b58` - status code: 417 Missing token

The link does however seems to work on frontend.

![image](https://user-images.githubusercontent.com/2919859/61053650-3e0b9080-a3ee-11e9-999f-73c1014ac353.png)

bjarnef:
I guess this method should have an overload for `Udi` and maybe `Guid` as well?
https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Web/Editors/EntityController.cs#L286-L291

or in Angular it need to convert it to an `int` id?

bjarnef:
It seems in only hit this method first time where it send a request like this: `/umbraco/backoffice/UmbracoApi/Entity/GetUrlAndAnchors?id=1129`

When your afterwards edit the link it send this request:
`/umbraco/backoffice/UmbracoApi/Entity/GetUrlAndAnchors?id=umb://document/afcc3d4df40d4d1eb8f2c43fdc592b58` which return the 417 (Missing token) error.

https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Web/Editors/EntityController.cs#L285-L291

I guess it is also an issue in v8:
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web/Editors/EntityController.cs#L290-L296

bjarnef:
In media picker there is something similar to use `udi` or `id`.
https://github.com/umbraco/Umbraco-CMS/blob/4733256ca1bfb49d4fe946110e734db1351f81d9/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js#L97

[...]

PR #5881, opened 2019-07-13
v7: Fix link picker url and anchors

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5849

### Description
This PR fixes the issue, where it returned `417 Missing token` in console, when edit an existing link in RTE.
I noticed when inserting a link first time it was using `id`, but on edit it use the existing `udi`. However it didn't seems to hit the method with the signature `int id`.

I had a look at the other method, e.g. `GetPath` and it seems the params need to be an array. After I made this change in `entityResource.getUrlAndAnchors` it hit the method with signature `Udi id` when editing the link.
https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Web.UI.Client/src/common/resources/entity.resource.js#L85
This might has something to do about these special `ParameterSwapControllerActionSelector` https://github.com/umbraco/Umbraco-CMS/blob/de9241bcf592ca9994810d43232545844fc06848/src/Umbraco.Web/Editors/EntityController.cs#L57-L61 , but now it is similar with get e.g. the `GetPath` method.

We could also have an overload method for `Guid id`, but it doesn't seem we need this at the moment. I have however added two overload method the `UmbracoHelper` to get Url from `Guid` as well.

Also note when edit the link the "Link" property field in the dialog no longer shows `localLink`, but the actual link to the page.

devenv_2019-07-13_14-45-52

devenv_2019-07-13_14-46-18

We should probably also merge this to v8 - the only difference is that v8 also have a parameter with culture in this method. https://github.com/umbraco/Umbraco-CMS/blob/4733256ca1bfb49d4fe946110e734db1351f81d9/src/Umbraco.Web/Editors/EntityController.cs#L290-L296
Let me know if I should submit a separate PR for v8 or you can merge in to v8 and adjust it a bit.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/resources/entity.resource.js
+++ b/src/Umbraco.Web.UI.Client/src/common/resources/entity.resource.js
@@ -172,7 +172,7 @@ function entityResource($q, $http, umbRequestHelper) {
                     umbRequestHelper.getApiUrl(
                         "entityApiBaseUrl",
                         "GetUrlAndAnchors",
-                        { id: id })),
+                        [{ id: id }])),
                 'Failed to retrieve url and anchors data for id ' + id);
         },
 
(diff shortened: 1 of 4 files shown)


bjarnef:
This is the current behaviour in v7.15.0 where it return the console error `417 Missing token` and note also the shown `locallink` in the Link disabled property field - not very useful to editors.

Shazwazza:
Great stuff @bjarnef  and good investigation :)

Issue #5857, opened 2019-07-11
v8: Media picker doesn't show preview of svg


bjarnef:
It seems media picker no longer show preview of selected svg as it did previous. Tested on Umbraco v8.1.0

![image](https://user-images.githubusercontent.com/2919859/61078391-cdca3280-a420-11e9-95c5-b5b814839ca1.png)

![image](https://user-images.githubusercontent.com/2919859/61078402-d4f14080-a420-11e9-8035-82ca63743f3b.png)

![image](https://user-images.githubusercontent.com/2919859/61078442-e9353d80-a420-11e9-9f70-0976037ac82c.png)

bjarnef:
For some reason the media url is rendered in the alt attribute for SVG files.

![image](https://user-images.githubusercontent.com/2919859/61078555-2994bb80-a421-11e9-8611-4695713ab8ec.png)

bjarnef:
On the media edit page, where it show the media properties incl. the fileupload, it has the following json for the `vm.files`:
```
[ { "fileName": "/media/rs4hln03/cherry-svgrepo-com.svg", "isImage": false, "extension": "svg", "fileSrc": null } ]
```

However it doesn't render the image because `fileSrc` is `null`.
https://github.com/umbraco/Umbraco-CMS/blob/d328a53590ec4ad4f532ce056795e16be132a86f/src/Umbraco.Web.UI.Client/src/views/components/upload/umb-property-file-upload.html#L21

bjarnef:
It seems `media` from this part in media picker property editor no longer returns meta data for extension as previous.

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.controller.js#L29

![image](https://user-images.githubusercontent.com/2919859/61089096-5d300f80-a43a-11e9-8e44-97745d8ad712.png)

bjarnef:
@Shazwazza @nul800sebastiaan do you know if there is a difference in the `metaData`. returned if the media item is file vs image?

It would also be great if these meta data properties was returned as camel case for consistency (but it would require where used in angular to be updated and probably a breaking change if people already use these properties in their own code.

bjarnef:
I also noticed when selecting e.g. a `txt` file the media picker preview was blank.

drbldgtl:
Also seeing this in 7.15.2

[...]

PR #5883, opened 2019-07-13
v8: Fix SVG image preview

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5857

### Description
This PR fixes some issues, where SVG files no longer was showing previews in media section on the page for the individual files and in media picker preview. I noticed the entities no longer returns meta data extension as previous. I guess it is because the entities have been refactored and e.g. it only set an additional property for `MediaPath` here. https://github.com/umbraco/Umbraco-CMS/blob/580ba79619cee347512131eff28e53d96af4958b/src/Umbraco.Web/Models/Mapping/EntityMapDefinition.cs#L52
Besides this meta data it also has `ContentTypeAlias` and `IsContainer` properties. Maybe media items should also have an `Extension` property and an `Size` property (umbracoBytes)? The `umbracoBytes` might require some additional lookup, but the extension can be determined from the media path.

This can however be determined from the media path as here:
https://github.com/umbraco/Umbraco-CMS/blob/ade9bb73246caf25a7073f2b9e5262641a201863/src/Umbraco.Web.UI.Client/src/common/services/mediahelper.service.js#L378-L387
https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web.UI.Client/src/common/directives/components/umbmediagrid.directive.js#L166

If there is no file path I think this function should return `null` instead of `false`.
https://github.com/umbraco/Umbraco-CMS/blob/ade9bb73246caf25a7073f2b9e5262641a201863/src/Umbraco.Web.UI.Client/src/common/services/mediahelper.service.js#L381

image


bjarnef:
An example of the SVG icon shown in both media picker, listview grid in media section and at the fileupload.



Maybe we should consider to add a component, e.g. `umb-file-icon` or similar with can be re-used the show the file icon and specify the extension and maybe a size as well.
For example `<umb-file-icon extension="txt" size="s" />`.

At the moment the file in both media picker and the icon at the fileupload has the following structure:

```
<div class="umb-icon-holder">
       <span class="file-icon">
              <i class="icon {{media.icon}}"></i>
                  <span>.{{media.extension}}</span>
              </span>
       <small>{{media.name}}</small>
</div>
```

Only in media grid layout the markup is a bit different, but it would be great if we just could re-use this component and adjust the styling from a component less file.

bjarnef:
@nul800sebastiaan it would be great if we could have these svg previews working again :)

nul800sebastiaan:
Thank you kindly good sir and apologies for the delay! 🙈

bjarnef:
Yeay! 😁 I wonder if it would make sense to create an `umb-file-icon` directive/component to wrap this markup, which can be re-used?

PR #5884, opened 2019-07-13
Accessibility: "Save and publish" has a drop down that can't be accessed via keyboard

### LINKED ISSUE #5882: Improve accessibility of the "Save and publish" dialog

This issue is related to issue no. 5 in the list of accessibility issues here https://github.com/umbraco/Umbraco-CMS/issues/5277

While it appears that the dropdown is now keyboard accessible there still are a few other concerns that needs to be addressed in order to improve screen reader accessibility.

1. For semantics the `<a>` tags should be changed to `<button>` tags
2. The aria-expanded and aria-controls attributes should be added to the caret communicating the state of the "popup" to screen readers
3. The aria-hidden attribute should be added to the button group communicating if the options are visible or not to screen readers
4. A visually hidden text of "More publishing options" or similar should be added to the "caret" element providing screen readers some proper context about what the button does
5. The ellipsis should be wrapped in a `<span>` element with the aria-hidden attribute set to true since the ellipsis does not make sense in a non-visual context. Maybe it does not in a visual context either? But that's not to be addressed in this issue

I'll create a draft PR for this issue.

BatJan:
### Prerequisites

This fixes issue 5 from #5277

### Description
Visually there is nothing changed with this PR - It all looks like it did before my changes. However a few things have happened under the hood.

1. `<a>` tags have been replaced with `<button>` tags in general
  1.1 Due to this some styling changes have been made so that the `button` tag is also being targeted just like the `<a>` tags currently is in the CSS.
2. The role attributes "menu" and "menuitem" have been added - Since Umbraco should be considered an app and not a website, this is the proper pattern for exposing the menu properly to screen readers. Heydon Pickering writes about this in the article "[Building Accessible Menu Systems](https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/)" - Jump to the headline that reads "True menus", since this is what's relevant for the changes that has been implemented here 😃 
3. aria-haspopup and aria-expanded have been added to the button that triggers the popup of the publishing options - Again communicating state to screen readers
4. The ellipsis part has been hidden for screen readers by wrapping the dots in a span with aria-hidden="true" for the publishing options that appear when clicking the caret sine they don't make sense in a non-visual context
5: A hidden text has been added inside the <span class="caret"> to provide screen readers with some context on what they can expect to find when clicking - This means the directive has been extended with a label and a label-key, which are optional since there are other places that will need to have such a text passed too at a later stage and making them optional so we don't risk breaking anything 😅 

I think this pretty much sums up the changes that have been made - And as mentioned nothing should have changed visually but hopefully things have improved for screen readers.

Huge thank you to Danny Lancaster for a good discussion about the ins and out's of this PR 🙌


emmaburstow:
Hi @BatJan 

Lots of wonderful work here, thanks. We'll be in touch if we need anything from you.

Emma

nul800sebastiaan:
Yay, finally!! 🎉

PR #5885, opened 2019-07-14
Change PropertyType.Alias to virtual

hifi-phil:
When unit testing PropertyType.Alias needs to be virtual to make it mockable. The underlying code on the set triggers IOC logic and makes it hard to test anything that uses Property Types.

This may have some security implications so I'll understand if you leave this as is....

### DIFF
--- a/src/Umbraco.Core/Models/PropertyType.cs
+++ b/src/Umbraco.Core/Models/PropertyType.cs
@@ -114,7 +114,7 @@ namespace Umbraco.Core.Models
         /// Gets of sets the alias of the property type.
         /// </summary>
         [DataMember]
-        public string Alias
+        public virtual string Alias
         {
             get => _alias;
             set => SetPropertyValueAndDetectChanges(SanitizeAlias(value), ref _alias, nameof(Alias));


Shazwazza:
it is get/set though so can you not just set the alias to what you want?

hifi-phil:
yes, I could, but does that not overly complicate testing? it spoils the flow of the developer doing the testing, giving them errors that are not expected and forcing them to use creative coding to get around a problem that can be easily fixed by just making it easier to mock

Shazwazza:
sure, just asking the questions that need to be asked ;) since none of the other properties are virtual though, does it mean inevitably they will all need to be virtual? Ideally it's an interface but that will be a breaking change so can't do that now. I'm fine with making the alias virtual but what about everything else with regards to testing? Making everything virtual starts to get a little ugly.

hifi-phil:
I asked the same questions myself when I was thinking about submitting the PR. I agree it would be much better as an Interface but as you said that would be a breaking change. Also setting all properties as virtual would get a little ugly, but I don't think that will be the case unless the underlying functionality is changed significantly and that might mean that it can be changed to an interface? I tried to make a surgical change to fix the immediate problem and make the developer experience better...

Shazwazza:
All good, if you are happy with just the alias being virtual, i'll merge it in. I'd assume that's the main thing needed for testing in most cases.

hifi-phil:
Thanks, Shannon. Yes, this is the property used most in testing, it will make a big difference...

PR #5888, opened 2019-07-14
V8: Accessibility Changes For Title On Login Page

RachBreeze:
This PR aims to give the user more context as to what the page is displaying by setting the `<title>` element

I have made this a draft PR because whilst is solves the issue on the login page, it isn't generic enough.

Ideally a `SetTitle` method based on this `init.js` routine should be created:

`$rootScope.$on('$routeChangeSuccess', function (event, current, previous) {..}`

Currently it relies  on a `.params` property which doesn't exist in the context of the login page.

A refactored "`SetTitle`" routine could then be reused once logged into Umbraco to provide more context as to what action the user is performing (currently on the section the user is in is displayed in the title, not what area within that section/ or if the user is creating/ editing something).  More information would help:

-  screen reader uses have context as to what the page is for
-  all users who in 2 or more browsers with the same Umbraco site open

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/application/umblogin.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/application/umblogin.directive.js
@@ -13,7 +13,7 @@
             }
         });
 
-    function UmbLoginController($scope, $location, currentUserResource, formHelper, mediaHelper, umbRequestHelper, Upload, localizationService, userService, externalLoginInfo, resetPasswordCodeInfo, $timeout, authResource, $q) {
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hi Rachel,

Thanks for this work. I can see it's still got Draft as its status. We'll take a look once it's complete.

Emma

RachBreeze:
Hey so have spent a while looking at making a generic SetTitle routine and failed.  Feel it would be good to get this as a PR and revisit the generic routine again

nul800sebastiaan:
Lovely @RachBreeze - thank you very much!

Issue #5810, opened 2019-07-09
v8: Log viewer refresh when no change in time period


bjarnef:
Recently a time period datepicker was added to log viewer.

I have tested this with latest v8.1.0 and when clicking the time period and afterwards clicking outside to close the datepicker it refresh the log overview although there is no change in datetime period.
I think it only some refresh when actually the (date)time period change.

![image](https://user-images.githubusercontent.com/2919859/60877213-acf5b780-a23d-11e9-92c6-f2aae0cfcebd.png)

Furthermore when only selecting a startdate and then clicking outside, it seems to reset the datetime period.

I think this should only change when both a start and end datetime has been chosen.
![image](https://user-images.githubusercontent.com/2919859/60877291-d3b3ee00-a23d-11e9-8b14-47b34fec0bd0.png)

Maybe the format of the daterange should also be updated from `2019-07-01,2019-07-09` to `2019-07-01 - 2019-07-09` or `2019-07-01 to 2019-07-09`

It seems there is a minor issue with styling of the selected daterange, e.g. selecting July 1st to July 4th.

When viewing month July:
![image](https://user-images.githubusercontent.com/2919859/60878156-7b7deb80-a23f-11e9-9fe5-4ab54d4d1607.png)

When viewing month June:
![image](https://user-images.githubusercontent.com/2919859/60878167-820c6300-a23f-11e9-88c4-b64f5bcd6eb5.png)

bjarnef:
It seems the value of the variable is not compiled correct, when used here:
https://github.com/umbraco/Umbraco-CMS/blob/a1722f268cb7b7f5b42eb3263ae28dda5924e570/src/Umbraco.Web.UI.Client/src/less/variables.less#L148

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web.UI.Client/src/less/components/umb-flatpickr.less#L20

The output in the browser, which in invalid css.

```
span.flatpickr-day.selected:hover {
    background-color: ligthen(#f2ebe7,10);
}
```
![image](https://user-images.githubusercontent.com/2919859/61186019-75817380-a660-11e9-8b4e-a76936c2e6f4.png)

nul800sebastiaan:
Fixed in: https://github.com/umbraco/Umbraco-CMS/pull/5890

PR #5890, opened 2019-07-14
v8: Log viewer datepicker

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5810

### Description
The logs are now only reloaded, when there is an actually change in the date range.

The date picker configuration has `conjunction: " to "`:
https://github.com/umbraco/Umbraco-CMS/blob/fd8ca35e63ef3f7159d923b3d0e40f64b3a2d45b/src/Umbraco.Web.UI.Client/src/views/logviewer/overview.controller.js#L153

This seems to work on init..

image

... but when changing the date range, it change to `startDate,endDate` instead of `startDate To endDate`. Not sure if this is a bug in the current version or something in the Umbraco implementation that cause this issue. It seems to work correct in the examples https://flatpickr.js.org/examples/#customizing-the-conjunction

image


emmaburstow:
Hiya Bjarne,

Lots of fab work here. we'll take a good look and let you know if we have any questions.

Em

bjarnef:
It seems this PR has been a bit forgotten. I have updated the branch with latest changes .

nul800sebastiaan:
Wow, a year later! I am so sorry for the delay @bjarnef - better late than never I guess 🙈  This is a wonderful update and great to have it finally merged, thanks very much!! ⭐

PR #5891, opened 2019-07-15
Adds additional security check for member controller routes

Shazwazza:
Adds additional security check for member controllers so they cannot be accessed by direct route. Adds additional security check in UmbracoAntiForgeryAdditionalDataProvider to validate current routed controller.

### DIFF
--- a/src/Umbraco.Web/Controllers/UmbProfileController.cs
+++ b/src/Umbraco.Web/Controllers/UmbProfileController.cs
@@ -11,6 +11,7 @@ using Umbraco.Core;
 
 namespace Umbraco.Web.Controllers
 {
+    [UmbracoProxyRouteOnly]
     [MemberAuthorize]
     public class UmbProfileController : SurfaceController
     {
--- a/src/Umbraco.Web/Controllers/UmbRegisterController.cs
+++ b/src/Umbraco.Web/Controllers/UmbRegisterController.cs
@@ -7,6 +7,7 @@ using Umbraco.Web.Mvc;
 
 namespace Umbraco.Web.Controllers
 {
+    [UmbracoProxyRouteOnly]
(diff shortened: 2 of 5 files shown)


liamlaverty:
Hey Shazwazza, 

I think there's an unintended negative consequence to returning a 4xx error at the line below. It effectively will show everyone which sites are and aren't patched to this update. If you get the current 5xx error, then you know it's unpatched and "vulnerable" to an exploits, whereas if you get the 4xx error it's patched. 

It'd maybe be worth implementing a 5xx error for one minor patch, and then changing it to 4xx in a future major patch. That way getting a 5xx error won't be a quick check for unpatched sites (for which a bot could be written)

https://github.com/umbraco/Umbraco-CMS/pull/5891/commits/14acb1421da2ee641e866dca96eb6e92259cfa1c#diff-43ecb7de644dbae2ea44c7baabeec89cR49

ronaldbarendse:
@liamlaverty There's probably lots of other ways to check the installed version (and determine whether it's patched) and if you use the workaround (removing the routes), you'll also get the 404 error.

I've already submitted an alternative that throws exceptions (just like with missing/invalid anti-forgery tokens): https://github.com/umbraco/Umbraco-CMS/pull/5895. I think that's the better approach, as it doesn't affect other MVC forms/anti-forgery tokens and gives better validation details.

Shazwazza:
@liamlaverty the opposite affect can be at play as well, in #5895 the exception is detailed which can provide more information than necessary. I've left @ronaldbarendse a msg about this too. But like @ronaldbarendse mentions applying the work around you'll get a 404 anyways. A 404 IMO is the correct response when trying to access an endpoint that doesn't really exist. I will close this PR anyways, and will look at @ronaldbarendse's for now and make whatever ammends necessary.

Issue #5894, opened 2019-07-15
U8 (8.1.0 Cloud) Backoffice throws an error when you add property to the media type


Baracuda011:
Dear all,

I have tried to add one more property to IMAGE media type. I created new property called **umbracoUrl** and used **Multi URL Picker** as the data type. When I saved, the backoffice threw the following error:

Received an error from the server
An error occured
Object reference not set to an instance of an object.

Exception Details
System.NullReferenceException: Object reference not set to an instance of an object.
Stacktrace
at Umbraco.Web.PublishedCache.NuCache.PublishedSnapshotService.RefreshMediaTypesLocked(List`1 removedIds, List`1 refreshedIds, List`1 otherIds, List`1 newIds)
   at Umbraco.Web.PublishedCache.NuCache.PublishedSnapshotService.Notify[T](ContentStore store, JsonPayload[] payloads, Action`4 action)
   at Umbraco.Web.PublishedCache.NuCache.PublishedSnapshotService.Notify(JsonPayload[] payloads)
   at Umbraco.Core.PublishedModelFactoryExtensions.WithSafeLiveFactory(IPublishedModelFactory factory, Action action)
   at Umbraco.Web.Cache.ContentTypeCacheRefresher.Refresh(JsonPayload[] payloads)
   at Umbraco.Core.Sync.ServerMessengerBase.Deliver[TPayload](ICacheRefresher refresher, TPayload[] payload)
   at Umbraco.Web.Cache.DistributedCache.RefreshByPayload[TPayload](Guid refresherGuid, IEnumerable`1 payloads)
   at Umbraco.Web.Cache.DistributedCacheExtensions.RefreshContentTypeCache(DistributedCache dc, ContentTypeChange`1[] changes)
   at Umbraco.Core.Events.TypedEventHandler`2.Invoke(TSender sender, TEventArgs e)
   at Umbraco.Core.Events.EventDefinition`2.RaiseEvent()
   at Umbraco.Core.Events.QueuingEventDispatcher.ScopeExitCompleted()
   at Umbraco.Core.Events.QueuingEventDispatcherBase.ScopeExit(Boolean completed)
   at Umbraco.Core.Scoping.Scope.<>c__DisplayClass72_0.<RobustExit>b__1()
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions)
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions)
[...]

PR #5897, opened 2019-07-15
V8: Fix YSOD when editing media types

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5894

### Description

As described in #5894, a server error is displayed when you try editing a media type:

image

This PR fixes it:

add-media-type-property

#### Testing this PR

1. Verify that you can add a new property (of any type) to an existing media type.
2. Verify that you can remote the property again.

### DIFF
--- a/src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
@@ -1057,7 +1057,7 @@ namespace Umbraco.Web.PublishedCache.NuCache
 
                 _mediaStore.UpdateContentTypes(removedIds, typesA, kits);
                 _mediaStore.UpdateContentTypes(CreateContentTypes(PublishedItemType.Media, otherIds.ToArray()).ToArray());
-                if(newIds != null && newIds.Any())
+                if(newIds.IsCollectionEmpty() == false)
                 {
                     _mediaStore.NewContentTypes(CreateContentTypes(PublishedItemType.Media, newIds.ToArray()).ToArray());
                 }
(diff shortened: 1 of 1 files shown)


src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
```diff
@@ -1057,7 +1057,10 @@ private void RefreshMediaTypesLocked(List<int> removedIds, List<int> refreshedId
 
                 _mediaStore.UpdateContentTypes(removedIds, typesA, kits);
                 _mediaStore.UpdateContentTypes(CreateContentTypes(PublishedItemType.Media, otherIds.ToArray()).ToArray());
-                _mediaStore.NewContentTypes(CreateContentTypes(PublishedItemType.Media, newIds.ToArray()).ToArray());
+                if(newIds != null && newIds.Any())
```
Shazwazza: `Any()` shouldn't be used here, we are already dealing with a `List<T>` so you can use `Length`. `Any` uses Linq which means it's first going to create an iterator and use that to see if there is anything in there but we already know that with our concrete object.

Also you can see above that we have an extension method to check exactly this which we are already doing: `newIds.IsCollectionEmpty()` so this ext method should be used instead.

kjac: Makes perfect sense, thanks. It's done now 👍

Shazwazza:
All looks good, we actually have these checks in the RefreshContentTypesLocked but not in RefreshMediaTypesLocked. This PR is good, but there's another check missing if you look at RefreshContentTypesLocked . I'll merge and cherry pick to 8.1 and add the additional check. Thanks!

vmachacek:
Hello, can someone please tell me when this will be released?

Shazwazza:
in 8.1.1, most likely shipped in one week.

PR #5905, opened 2019-07-16
Accessibility: No alt text on any icons

### LINKED ISSUE #5904: No alt text on any icons (Issue 122 from the accessibility list)
The icons should be hidden for screen readers using the aria-hidden="true" attribute and the <a> tag should be refactored to a <button> and the tabindex="-1" should be removed so the buttons are possible to navigate using a keyboard.

List of issues: https://github.com/umbraco/Umbraco-CMS/issues/5277

I'll do a PR shortly.

BatJan:
### Prerequisites

This fixes issue 29, 108 and 122 from #5277 

### Description
This PR fixes some accessibility issues with the button's in the top right corner of the users section "Users" and "Groups".

The `<a>` tag has been converted to a `<button>` and the `tabindex="-1"` has been removed, so it's possible to reach the options using the tab key. Furthermore the icon has been hidden from screen readers using the _aria-hidden_ attribute.

Also I've adjusted the styling to cater for the button now being used and so that it's possible to visually see that one has tabbed to one of the options.

This is a general change in the `<umb-editor-navigation-item>` directive/component so it will also improve the accessibility wherever this is used like in the content section (content apps) and package section for instance 😃 

**Edit 17/07/2019**: Turns out issue 29 from the accessibility issue is somewhat related to this issue as well so I have fixed the issue mentioned in this issue as well.

This means the dropdown has got the following updates

- The href attribute on the <a> tag has been updated to reference the group that it makes you jump to for semantic reasons since it's really an anchor, which means the "umb-tabbed-content" file has been modified too in order to have an id attribute that matches the anchor
- Furthermore the styling has been changed so the groups can be tabbed through like illustrated in the gif's below using the :focus-within selector
- I propose creating a new area in the language files called "visuallyHiddenTexts" as I foresee we're going to be needing many of those texts and if we don't put them in the same group of texts it can become hard to figure out where and why they're being used - Should also make it easier to keep track of for people wanting to translate the keys. So I have done that adding english and danish keys meaning screen readers will get "Jump to Content group" read out but visually one will only see "Content" in the dropdown

Looking forward to receive some feedback one this 👍 

**Tabbing before**
umb-editor-navigation-item-before

**Tabbing after**
umb-editor-navigation-item-after

**Tabbing where a tab has a dropdown before**
umb-editor-navigation-item-withdropdown-before

**Tabbing where a tab has a dropdown after**
umb-editor-navigation-item-withdropdown-after


emmaburstow:
Wonderful work here, Jan. Thanks ever so much. We'll get in touch if we need you at all.

Em

nul800sebastiaan:
Love it! Corrected the `visuallyhidden` class to `sr-only`! :-)

PR #5910, opened 2019-07-16
V8: Make user group sections mandatory

kjac:
### Prerequisites

### Description

I believe it makes no sense to have a user group without access to any sections, but this might not be entirely apparent for newcomers. So this PR makes it mandatory to grant user groups access to at least one section.

When applied it works like this:

group-sections-mandatory


Shazwazza:
User groups are the same as 'roles' in both the old membership provider land and in ASP.NET identity. It is common for controllers to have Authorization attributes that authorize specific roles (groups). It is certainly possible for people to use roles (groups) for any custom security they want in the back office so i'm not sure this is black and white.

Issue #5641, opened 2019-06-11
[U8] Can't User New User Groups


0Neji:
I'm trying to set up a new user group which has restricted permissions and access content nodes, however when I try to login as this user I'm presented with the Umbraco Back Office but with no menu, no content tree. I do see my user icon but if I click it, I'm immediately returned back to the login screen.

Reproduction
------------

Tested on an existing Umbraco 8 project and a fresh install. To test on fresh install:

1. Create new project/install of Umbraco 8 via NuGet
1. Login and go to `Users`
1. Go to `Groups` (top right)
1. Click `Create group`
1. For testing purposes, I used the following details
     Name: Test
     Sections: Content
     Content start node: Content root
     Default permissions (checked if listed, unchecked if not): Browse Node, Delete, Create, Publish, Unpublish, Update, Copy, Move, Sort
1. Save
1. Create new user (replicated with both invite and manual creation)
1. Assign new user the `Test` user group
1. In a private browser window, try to login as the user.

### Specifics

Umbraco Version: 8.0.2
Browser: Tested on latest version of Chrome, Brave, Firefox (all on Windows 10)

### Expected result

Expected successful login with access to only the selected nodes/section.

### Actual result

When I login I see:

![image](https://user-images.githubusercontent.com/4700480/59268212-1fc24180-8c44-11e9-9d0c-46f129a7c601.png)

And the console shows the following errors:

![image](https://user-images.githubusercontent.com/4700480/59268227-29e44000-8c44-11e9-857a-49bfae6ed305.png)

It looks like for whatever reason, the user doesn't have access to the JS files it needs to load the CMS fully. I've also tried the same user works with the 'Administrator' user group and login is allowed, works as normal.

nul800sebastiaan:
Looks like something is requiring a media start node to be set, even if you have no access to the media section. From your errors, it seems like it might have something to do with the tours.

I've marked this as "Up for grabs" so that you or someone else coming along could create a pull request for it.

AndyButland:
[...]

PR #5911, opened 2019-07-16
V8: Make user group start nodes mandatory

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5641

### Description

As @AndyButland points out on #5641, user groups with no content/media start nodes configured cause a bit of havoc when users assigned to these groups log into the back office. The users are actually logged in successfully, but due to the [back office identity claim types](https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Core/Security/UmbracoBackOfficeIdentity.cs#L118), any subsequent calls to back office resources end up as 401 unauthorized.

I do believe it makes sense to make both content and media start nodes mandatory for user groups for a few reasons:

1. Even if the user group does not allow access to content and/or media, the users may implicitly encounter those sections via pickers etc. - having been forced to consider the start nodes for these users would be beneficial.
2. It would of course solve #5641 😄 

With this PR applied, group editing behaves like this:

group-start-nodes-mandatory

See also #5910

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/users/group.html
+++ b/src/Umbraco.Web.UI.Client/src/views/users/group.html
@@ -48,7 +48,7 @@
                                         </a>
                                     </umb-control-group>
 
-                                    <umb-control-group style="margin-bottom: 20px;" label="@user_startnode" description="@user_startnodehelp">
+                                    <umb-control-group style="margin-bottom: 20px;" label="@user_startnode" description="@user_startnodehelp" required="true">
                                         <umb-node-preview
                                             ng-if="vm.userGroup.contentStartNode.id"
                                             style="max-width: 100%;"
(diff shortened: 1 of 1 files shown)


Shazwazza:
Same thing as https://github.com/umbraco/Umbraco-CMS/pull/5910 ... i'm not sure this is black and white. Having user groups just defined as groups that people use for their own security reasons should be possible without a group defining start nodes or sections.

kjac:
@Shazwazza I get your point. I still believe there's an ease of learning curve by making these things mandatory; as #5641 shows, it's too easy to skip over them and then be frustrated with the product afterwards.

Shazwazza:
@kjac but with #5641 surely we can fix that experience directly instead of trying to resolve the issue indirectly? I'm all for other ideas though but we can't prohibit people from creating user groups without start nodes or sections applied.

kjac:
@Shazwazza roger. I'll close the PRs 😄

Shazwazza:
@kjac i can try to follow up tomorrow/thur if you don't beat me to it ;) I'm sure we can find a solution to the underlying issue.

kjac:
@Shazwazza I'm having a look at it later today, see if we can't fix it during login with appropriate logging of errors.

marcemarc:
> Same thing as #5910 ... i'm not sure this is black and white. Having user groups just defined as groups that people use for their own security reasons should be possible without a group defining start nodes or sections.

+1 yes on large sites, we have users in multiple groups, some groups set the start node for the territory tree for the users in that group and then further groups set granular permissions on nodes within those territories, that not all editors are in - also we use userGroups to gather together users  for custom purposes not just tree start node eg 'SuperTerittoryAdminApprover'. All the User Groups combine to define what the editor can see and do... so not all UserGroups have start nodes!

kjac:
New PR in #5917

PR #5912, opened 2019-07-16
V8: Can't delete user groups

kjac:
### Prerequisites

### Description

It's not possible to delete user groups - the server simply returns an error:

image

It doesn't matter if you attempt to delete one or multiple user groups, the result is the same.

With this PR applied, user group deletion works as expected:

delete-user-groups


kjac:
Reported as an issue in #5958

emmaburstow:
Hey Kenn, Thanks for doing this. And thanks for linking it to the issue. We'll be in touch if we need you,

Em

kjac:
PR updated with latest v8/dev

nul800sebastiaan:
Lovely fix, thanks a lot @kjac !

PR #5918, opened 2019-07-17
5732 grid reorder on scroll listview bugfix

### LINKED ISSUE #5878: Sticky header shows in "scrolled down" state before scrolling
A lot of the time I see the sticky header in its "scrolled down" state even if I haven't started scrolling yet:

image

It happens on load. If I scroll down and up again, the "scrolled down" state disappears.

### LINKED ISSUE #5879: Sticky header doesn't stick to the top in the media section
The sticky header usually slides up and sticks to the top of the view when scrolling begins. This doesn't seem to be the case in the media section no more:

image

The result is that the scrolled content is visible behind the sticky header, as shown above.

### LINKED ISSUE #5880: Sticky header throws a JS console error
The following error logs to the JS console pretty often when browsing through folders in the media section:

image

Might be related to #5879

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #5878 and fixes #5879 and fixes #5880 and anything else that @kjac might have found... :)

This also includes the changes I made in PR #5834, which got real messy when I did a bad merge and ended up with 1372 commits instead of 4.

### Description

I've revisited my original solution for this one, realised it was overly complex, and reworked it to not be overly complex - previously I was adding two sentinel elements to track the top and bottom of the parent container, when really only need to know where the top of the sticky bar is, in relation to the scrolling element. I'm also disconnecting the observer when the directive is destroyed, not sure that it's necessary but probably doesn't hurt.

So, when the sentinel (now only one, at the top) leaves the top of the scrollable element, we know the sticky bar will be stuck, so can have the class applied. When the sentinel is inside the scrollable element, we know the stick bar is not at the top, so the active class can be removed. 

This is a good thing - less code, less complexity, less broken.

To verify, jump between the recycle bin, media section, users section and a content node with a grid editor. All sticky bars should be unstuck on load, and stick on scroll only. If this isn't the case, let me know so I can jump out a window.


kjac:
@nathanwoulfe you're a rockstar 🎸 seems to work like a charm, 'cept for the sticky bar disappearing behind the list view content:

sticky-header-list-view

From what I can tell, the culprit is the removal of `z-index: 99;` [here](https://github.com/umbraco/Umbraco-CMS/pull/5918/files#diff-c4783cfffd16db743bb788803974751fL40). Adding it again makes the listview _and_ dashboard headers play nice. I haven't tested the grid headers with that though.

nathanwoulfe:
Yup, you're right - I'd removed the z-indexes in a few places to try track down the original cause. Forgot to put that one back...

nul800sebastiaan:
What a *sticky* issue! 😁 

I won't pretend to understand much of the code but... It works soooo much better! 👍 🎉 

Thank you so much @nathanwoulfe! 💪

Issue #5902, opened 2019-07-15
v8.1 DB upgrade not mapping several property editor aliases


Zweben:
Several Umbraco 7 property editor aliases are not being mapped to their new aliases during v8.1's database upgrade process.

Reproduction
------------

### Bug summary

Upon trying to upgrade an Umbraco 7.14 database to 8.1 via Umbraco 8.1's upgrade process, the upgrade fails, and the log indicates that the CMS could not find property editors with certain aliases.

### Specifics
I receive many instances of the following message in the log, for the property editor aliases listed below:
> Skipping validation of configuration for data type {NodeId} : {NewEditorAlias} (was: {EditorAlias}) because no property editor with that alias was found. Please ensure that the configuration is valid. The site may fail to start and / or load data types and run.

- Umbraco.MultiNodeTreePicker2
- Umbraco.Textbox
- Umbraco.Date
- Umbraco.TextboxMultiple
- Umbraco.RelatedLinks2
- Terratype
- Spectrum.Color.Picker

In all cases, the string listed for NewEditorAlias is the same as the one listed for EditorAlias.

For the following property aliases, I receive a slightly different error message:

> Skipping validation of configuration for data type {NodeId} : {EditorAlias}. Please ensure that the configuration is valid. The site may fail to start and / or load data types and run.

- Umbraco.ContentPicker2
- Umbraco.RadioButtonList

### Steps to reproduce

1. I made a local copy of my Umbraco 8.0.2 Cloud project, upgraded it to 8.1.0 via NuGet, and removed all database files, media, UDA files, temp files, scripts, CSS, and templates.
2. I dragged the .MDF and .LDF files from my v7.14 project over to the v8.1 project and updated the Umbraco.Core.ConfigurationStatus version to 7.14.
3. I ran the site and started the upgrade

### Expected result

Aliases of built-in property editors are mapped automatically when trying to validate data type configuration, and the user is given a way to manually map the property aliases of 3rd-party property editors.

### Actual result

Upgrade fails to map property editor aliases during data type validation, causing a cascading failure that prevents the database from upgrading at all.

---
_This item has been added to our backlog [AB#1796](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/1796)_

stevemegson:
[...]

PR #5923, opened 2019-07-17
V8: Migration should not warn about build in data types that is handled correct but renamed

bergmania:
https://github.com/umbraco/Umbraco-CMS/issues/5902
- Fixed issue with migration of Umbraco.Date if it was never saved in v7, the default format was added the time-part.
- Fixed issue with warnings of migrations of build in types, that was renamed between v7 and v8. These are handled correct.
- Added constants for some of the legacy property editor aliases.


warrenbuckley:
Code reads fine @bergmania 
I am off to test this out :)

warrenbuckley:
Tested and the built-in DateTime datatype will not cause problems for a 7.15 -> 8.1.x content/db migration now

PR #5930, opened 2019-07-17
Accessibility: Open Package - Identifies back button as "Left hand arrow back button"

### LINKED ISSUE #5929: Open Package - Identifies back button as "Left hand arrow back button" (Issue 104 from the accessibility list)
This is issue 104 on the accessibility list: https://github.com/umbraco/Umbraco-CMS/issues/5277

Screen readers are reading out the ← character, which makes no sense. Therefore is should be wrapped in a span using an aria-hidden="true" attribute.

I'll create a PR shortly.

BatJan:
### Prerequisites

This fixes issue 104 and 117 from #5277 

### Description
This PR wraps every found instance of the unicode character &larr; in a aria-hidden="true" element so screen reader don't read it out loud, since it does not make much sense.

Also all of the places where the back button was used are refactored from `<a>` to `<button>` and styling has been fixed accordingly. Once this PR is approved I'll have a look at making some kind of button component since I noticed the exact same look and feel being implemented differently in the examine dashboard 😃

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-packages.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-packages.less
@@ -264,12 +264,16 @@
     flex-flow: row wrap;
 }
 
-a.umb-package-details__back-link {
+.umb-package-details__back-action {
     font-weight: bold;
     color: @black;
+    padding: 0;
+    border: 0;
+    background-color: transparent;
 }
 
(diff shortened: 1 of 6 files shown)


emmaburstow:
Yay! Another win for accessibility. Thanks Jan,

Em

nul800sebastiaan:
Sweet! Thanks @BatJan! ⭐️

PR #5938, opened 2019-07-18
Fixed #5907 by reloading content on bulk publish / unpublish

### LINKED ISSUE #5907: Listview not updated after bulk action, e.g. (un)publish
After upgrading to 7.14 we found that the listview does not get updated values and style after a bulk action. In our case we tried to publish and unpublish one or more nodes. 

### Bug summary
    * Write a short summary of the bug
Configure a listview. Select one or more rows in the listview and click "unpublish".  The nodes are unpublished according to the DB and logs, but the listview itself is not updated. The "clear selection"-button and the "2 of 3 selected"-label stay visible and the nodes stay selected. The "GetChildren"-endpoint is not called according to chrome dev console.

    * Try to pinpoint it as much as possible
This happens in multiple projects with version 7.13.0 (assembly: 1.0.6942.23076) or above (7.14+7.15). 
Using version 7.12.4 or 8.0.1 everything works as expected. 

    * Try to state the _actual problem_, and not just what you _think_ the solution might be.
Listview does not get updated with style and values after a publish- or unpublish action.

### Specifics
    * What version of Umbraco are you using (down to the very last digit!)
Umbraco version 7.14.0 assembly: 1.0.7009.26199
    * What browser and version you are using
    * Please mention if you've checked it in other browsers as well
Chrome + Edge (both latest version) on windows 10
    * Please include *full error messages* and *screenshots* if possible
Please find a screenrecording attached

### Steps to reproduce
Configure a listview. Select one or more rows in the listview and click "unpublish". 

### Expected result
Would expect to see the nodes get unchecked/unselected and, rows to be grayed out and the values in the "Is Published"-column change to "false". This should happen after the green notification ribbon is shown.

### Actual result
Listview is showing the same values and style as it appeared before clicking "Unpublish". It does not get updated in both its style and values. 

I also noticed that in the console the "PostUnPublish"-endpoint is called, but the "GetChildren" is not. 

Listview-issue

mattbrailsford:
Fixes #5907

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/listview/listview.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/listview/listview.controller.js
@@ -405,7 +405,7 @@ function listViewController($rootScope, $scope, $routeParams, $injector, $cookie
     };
 
    $scope.publish = function () {
-        applySelected(
+       var attempt = applySelected(
                 function (selected, index) { return contentResource.publishById(getIdCallback(selected[index])); },
                 function (count, total) {
                     var key = (total === 1 ? "bulk_publishedItemOfItem" : "bulk_publishedItemOfItems");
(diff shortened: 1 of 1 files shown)


emmaburstow:
Aaaand again. Thanks for getting this done. We'll look at it and let you know if we have any questions.

Em

nul800sebastiaan:
Works like a charm, thanks very much @mattbrailsford ! 👍

dsbaiz:
Hiya, I have the same issue's on v8 for every event, can this fix also be deployed for v8?

PR #5940, opened 2019-07-18
Remove unused .visuallyhidden class from main.less

### LINKED ISSUE #5939: Remove unused .visuallyhidden class from main.less
When I was having a look at some of the accessibility issues open and fixed I noticed that the .sr-only class has been used and that there are no longer any occurrences of the .visuallyhidden class being used no more. Both classes are identical and serve the same purpose. However the .sr-only class is being used and has some focus options attached to it, which the visuallyhidden class does not.

Therefore I'm going to do a PR removing the .visuallyhidden class shortly.

BatJan:
### Prerequisites

If there's an existing issue for this PR then this fixes #5939 

### Description
This PR removes the unused .visuallyhidden class from the main.less file 😃

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/main.less
+++ b/src/Umbraco.Web.UI.Client/src/less/main.less
@@ -592,13 +592,3 @@ input[type=checkbox]:checked + .input-label--small {
   background-color: @green-l3;
   text-decoration: none;
 }
-
-.visuallyhidden{
-    position: absolute !important;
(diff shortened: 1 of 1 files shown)


emmaburstow:
A lovely bit of tidy up here so thanks Jan. 

Em

nul800sebastiaan:
Lovely, thanks @BatJan ! 👍

PR #5942, opened 2019-07-18
Accessibility: Ensure "cancel" is clickable from move/restore dialogs

### LINKED ISSUE #5941: Accessibility: Ensure "cancel" is clickable from move/restore dialogs
Going through some of the issues from the accessibility list https://github.com/umbraco/Umbraco-CMS/issues/5277 I noticed some issues that does not seem to be reported regarding keyboard accessibility when using the "Move" and "Restore" dialogs, where it's currently not possible to tab through to the "cancel" button.

I'll do a PR for this shortly.

BatJan:
### Prerequisites

If there's an existing issue for this PR then this fixes #5941

### Description
The "cancel" button was not possible to reach in restore/move dialogs due to the anchor tag being used not containing a `href` attribute. Therefore it acts like a link container, which does not receive focus events.

However the fix for this refactors it to a `<button>` since it's not a link anyway.

The following GIF's illustrates the difference before and after and this affects content, developer and media sections.

**Before**
restore-move-before

**After**
restore-move-after


emmaburstow:
You have been busy @BatJan :)

We'll take a look and let you know if we have any questions. Nice giffing, as usual!

Em

src/Umbraco.Web.UI.Client/src/views/content/move.html
```diff
@@ -76,9 +76,9 @@
 
 
     <div class="umb-dialog-footer btn-toolbar umb-btn-toolbar" ng-hide="success">
-        <a class="btn btn-link" ng-click="close()" ng-show="!busy">
+        <button type="" class="btn btn-link" ng-click="close()" ng-show="!busy">
```
nul800sebastiaan: Any reason this is not `<button type="button"` @BatJan ?

nul800sebastiaan: Any reason this is not `<button type="button"` @BatJan ?

BatJan: Looks like an oversight because I had a an editor plugin that kinda worked and helped autocomplete stuff...except from when it did not 🤦‍♂ 😄 - I will add the types when I find a litle amount of time 👍

nul800sebastiaan: @BatJan No problem, it was just this line that was missing it, I was just curious if it was on purpose. I can add it as well when I check out the PR!

nul800sebastiaan:
This is all lovely @BatJan - thanks for the help! I corrected the missing `type="button"` I mentioned, all good now!

PR #5944, opened 2019-07-18
Accessibility: Media uploader widget can't be accessed via keyboard

BatJan:
### Prerequisites

This fixes issue 88 from #5277

### Description
In this PR I have changed the <div> to a <button> so it can be reached by keyboard and fixed the styling accordingly.

Apart from that I have added a text above the image to provide some more context about what is expected of the user here - I felt that the "- or click here..." was not providing enough context. Especially for editors as well.

However I'm open for suggestions for the text I have added since I'm not sure if my current suggestion is good enough. But I think it's better than having nothing. I'm not sure adding a text like "Drag and drop your files onto the image/dropzone below" is properly accessible - But neither is my current suggestion. But as mentioned, I think it's better than nothing 😃 

Below we have the before and after GIF's.

**Before**
media-widget-before

**After**
media-widget-after


emmaburstow:
Lovely work, Jan. Another win for accessibility. We'll shout if we need you,

Em

BatJan:
Thank you @emmaburstow 😃  However I think I ended up changing the added paragraph to say "area" instead of "Canvas" - I'm open for textual suggestions here 😄

nul800sebastiaan:
Awesome, thanks @BatJan ! 👍

Issue #5947, opened 2019-07-19
Custom Exception classes don't conform to best-practices


ronaldbarendse:
As discussed with @Shazwazza in PR https://github.com/umbraco/Umbraco-CMS/pull/5895, custom `Exception` classes in Umbraco currently don't all conform to the best-practices. See also:
- https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229064(v=vs.100)
- https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.iserializable
- https://stackoverflow.com/questions/94488/what-is-the-correct-way-to-make-a-custom-net-exception-serializable

This might not directly cause bugs or issues, but it's a small effort to just get it right 👍

ronaldbarendse:
@Shazwazza This should probably also get the `category/breaking` label, right?

Shazwazza:
Marked as breaking due to different exceptions being thrown in some scenarios, the mapping is:

* `DuplicateKeyException` -> `ArgumentException`
* `ArgumentNullOrEmptyException` -> both `ArgumentNullException` and `ArgumentException`
* `FileSecurityException ` -> `UnauthorizedAccessException`
* `WontImplementException` -> `InvalidOperationException`

PR #5950, opened 2019-07-19
Fixed custom Exception classes

ronaldbarendse:
### Prerequisites
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5947

### Description
All custom `Exception` classes shoud be updated with:
- `[Serializable]` attribute (as any class that might be serialized must be marked with this attribute)
- Constructor with the signature `(SerializationInfo information, StreamingContext context)` (as required for all classes that implement `ISerializable`)
- If custom properties are added, restore the data in the above constructor and override `GetObjectData(SerializationInfo info, StreamingContext context)` to save the data


src/Umbraco.Core/Exceptions/ArgumentNullOrEmptyException.cs
```diff
@@ -1,30 +1,54 @@
...
 namespace Umbraco.Core.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when a null reference, or an empty argument,
-    /// is passed to a method that does not accept it as a valid argument.
+    /// The exception that is thrown when a null reference, or an empty argument, is passed to a method that does not accept it as a valid argument.
     /// </summary>
+    /// <seealso cref="System.ArgumentNullException" />
+    [Serializable]
     public class ArgumentNullOrEmptyException : ArgumentNullException
```
ronaldbarendse: I see this exception is used quite a lot, but it's semantically not correct. If this is thrown for empty values, the base class is `ArgumentNullException`, but the argument isn't `null`... And as the current value isn't available nor the message specific enough, debugging whether it was `null` or empty will be unnecessarily hard.

Yes, the following code is quite cumbersome to write, but it's correct and gives good debug information:
```c#
if (param == null) throw new ArgumentNullException(nameof(param));
if (string.IsNullOrWhiteSpace(param)) throw new ArgumentException("Value can't be empty or consist only of white-space characters.", nameof(param));
```

See also:
- https://docs.microsoft.com/en-us/dotnet/api/system.argumentexception (explains the differences between the exceptions)
- https://docs.microsoft.com/en-us/dotnet/framework/debug-trace-profile/code-contracts (might be a better way to do these validations)

src/Umbraco.Core/Collections/ObservableDictionary.cs
```diff
@@ -235,15 +236,76 @@ public bool TryGetValue(TKey key, out TValue val)
...
+            /// <summary>
+            /// Initializes a new instance of the <see cref="DuplicateKeyException" /> class.
+            /// </summary>
+            /// <param name="key">The key.</param>
+            /// <param name="innerException">The exception that is the cause of the current exception, or a null reference (<see langword="Nothing" /> in Visual Basic) if no inner exception is specified.</param>
+            public DuplicateKeyException(string key, Exception innerException)
+                : base("Attempted to insert duplicate key \"" + key + "\" in collection.", innerException)
             {
-                Key = key;
+                this.Key = key;
```
Shazwazza: We don't really use `this` coding pattern in this codebase, it's fine but i will remove it.

Shazwazza:
Ah, i see this is still a draft, all good, have left a review note

src/Umbraco.Core/Exceptions/ArgumentNullOrEmptyException.cs
Shazwazza: I wasn't aware of this exception or why it exists. Both Visual Studio and Resharper offer a very easy way to check strings for both null and empty by just pressing ctrl + enter. I personally don't see a reason for this exception but i don't see a reason to throw both an ArgumentNullException and an ArgumentException.

Shazwazza: Visual Studio and Resharper both generate this single check for null/empty strings:

src/Umbraco.Core/Collections/ObservableDictionary.cs
ronaldbarendse: I'll update this in the next commits 👍

src/Umbraco.Core/Exceptions/ArgumentNullOrEmptyException.cs
ronaldbarendse: Well, the linked documentation from Microsoft clearly states:

> The primary derived classes of `ArgumentException` are `ArgumentNullException` and `ArgumentOutOfRangeException`. These derived classes should be used instead of `ArgumentException`, except in situations where neither of the derived classes is acceptable. For example, exceptions should be thrown by:
> 
> * `ArgumentNullException` whenever `null` is passed to a method that does not accept it as a valid argument.
> 
> * `ArgumentOutOfRangeException` when the value of an argument is outside the range of acceptable values; for example, when the value "46" is passed as the month argument during the creation of a `DateTime`.

This PR is all about best-practices, consistency and code-style, so I would recommend throwing the different exceptions for `null` and empty/white-space values. This also makes debugging a lot easier: just check what exception type was thrown to see what went wrong...

Shazwazza: Ok that's fine but we'll have to log a bug with MS themselves since they are generating code that doesn't adhere to their own standards ;) I think R# can do both exceptions but would need to check. Point is, that these null checks are generally performed by the tooling and that will continue to happen. There's are hundreds of these IsNullOrWhitespace checks with ArgumentException thrown because the tooling provides that. Perhaps the reason for this custom ArgumentNullOrEmptyException was to try to make this check clearer, i dunno.

nul800sebastiaan:
@ronaldbarendse is this still in draft or good to go now? :)

src/Umbraco.Core/Exceptions/ArgumentNullOrEmptyException.cs
ronaldbarendse: Visual Studio 2017 (15.9.15) generates the correct checks, so it's probably something in ReSharper or customized settings:
ArgumentNullException

ArgumentException

src/Umbraco.Core/Exceptions/DataOperationException.cs
```diff
@@ -1,21 +1,94 @@
...
+    [Serializable]
     internal class DataOperationException<T> : Exception
     {
+        /// <summary>
+        /// Gets the operation.
+        /// </summary>
+        /// <value>
+        /// The operation.
+        /// </value>
         public T Operation { get; private set; }
```
ronaldbarendse: As `T` could be any type, serialization can fail if the passed in type isn't serializable.

src/Umbraco.Core/Exceptions/WontImplementException.cs
```diff
@@ -1,27 +1,52 @@
...
-    /// <remarks>The <see cref="NotImplementedException"/> is to be used when some code is not implemented,
+    /// <remarks>
+    /// The <see cref="NotImplementedException" /> is to be used when some code is not implemented,
     /// but should eventually be implemented (i.e. work in progress) and is reported by tools such as ReSharper.
     /// This exception is to be used when some code is not implemented, and is not meant to be, for whatever
-    /// reason.</remarks>
+    /// reason.
+    /// </remarks>
+    /// <seealso cref="System.NotImplementedException" />
+    [Obsolete("If a method or operation is not, and will not be, implemented, it is invalid, so we should throw an InvalidOperationException instead.")]
```
ronaldbarendse: I've marked `WontImplementException` obsolete, as you should just throw `InvalidOperationException` instead.

src/Umbraco.Core/IO/FileSecurityException.cs
```diff
@@ -1,20 +1,46 @@
...
-using System.Text;
+using System.Runtime.Serialization;
 
 namespace Umbraco.Core.IO
 {
+    /// <summary>
+    /// The exception that is thrown when the caller does not have the required permission to access a file.
+    /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Obsolete("Throw an UnauthorizedAccessException instead.")]
```
ronaldbarendse: I've marked `FileSecurityException` obsolete, as it's only used once and just indicates unauthorized access (https://docs.microsoft.com/en-us/dotnet/api/system.unauthorizedaccessexception):
https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Core/IO/PhysicalFileSystem.cs#L317

src/Umbraco.Core/Persistence/EntityNotFoundException.cs
```diff
@@ -1,38 +1,95 @@
...
     // TODO: Would be good to use this exception type anytime we cannot find an entity
 
     /// <summary>
-    /// An exception used to indicate that an umbraco entity could not be found
+    /// An exception used to indicate that an Umbraco entity could not be found.
     /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Obsolete("This exception is only used in MediaController and could be replaced with an HTTP 404 error instead.")]
+    [Serializable]
     public class EntityNotFoundException : Exception
```
ronaldbarendse: I've marked `EntityNotFoundException` obsolete, as it's only used once in an API controller and could be replaced with a generic HTTP 404 error:
https://github.com/umbraco/Umbraco-CMS/blob/56545ad7f1fdee3c2b27e725c89378ccb947f65a/src/Umbraco.Web/Editors/MediaController.cs#L818

src/Umbraco.Core/Persistence/EntityNotFoundException.cs
```diff
@@ -1,38 +1,95 @@
...
     {
+        /// <summary>
+        /// Gets the identifier.
+        /// </summary>
+        /// <value>
+        /// The identifier.
+        /// </value>
         public object Id { get; private set; }
```
ronaldbarendse: Same as in `DataOperationException`, this property can contain a type that isn't serializable.

src/Umbraco.Core/Persistence/FaultHandling/RetryLimitExceededException.cs
```diff
@@ -4,59 +4,51 @@
...
+    /// <seealso cref="System.Exception" />
     [Serializable]
     public sealed class RetryLimitExceededException : Exception
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="RetryLimitExceededException"/> class with a default error message.
+        /// Initializes a new instance of the <see cref="RetryLimitExceededException" /> class with a default error message.
         /// </summary>
         public RetryLimitExceededException()
             : this("RetryLimitExceeded")
```
ronaldbarendse: This message could probably be formatted better.

src/Umbraco.Web/Install/InstallException.cs
```diff
@@ -1,44 +1,105 @@
...
+        /// </value>
         public string View { get; private set; }
+
+        /// <summary>
+        /// Gets the view model.
+        /// </summary>
+        /// <value>
+        /// The view model.
+        /// </value>
         public object ViewModel { get; private set; }
```
ronaldbarendse: This property can also contain a non-serializable type.

ronaldbarendse:
@Shazwazza Just to summarize the comments:
- Some exceptions can contain property types that may not be serializable (e.g. using `object` or generics): how should we handle this?
- Should we change `MediaController` to `throw new HttpResponseException(HttpStatusCode.NotFound)` instead of the custom exception?

Removing usage of obsoleted exceptions should probably be done in a seperate PR, as that might cause breaking changes, right?

Shazwazza:
Hi @ronaldbarendse 

> Some exceptions can contain property types that may not be serializable (e.g. using object or generics): how should we handle this?

I'm not really sure, if there's no clear cut way to handle this then should we just not allow serialization? It hasn't been a problem to date afaik so would it make sense to just not support the serialization for those types?

> Should we change MediaController to throw new HttpResponseException(HttpStatusCode.NotFound) instead of the custom exception?

yep i think so!

> Removing usage of obsoleted exceptions should probably be done in a seperate PR, as that might cause breaking changes, right?

I think this depends on where the usages are. If they are in our back office controllers, we are allowed those types of changes, same with at the service/repo level and some other places too.

src/Umbraco.Core/Collections/ObservableDictionary.cs
```diff
@@ -235,15 +236,76 @@ public bool TryGetValue(TKey key, out TValue val)
...
         #endregion
 
+        /// <summary>
+        /// The exception that is thrown when a duplicate key inserted.
+        /// </summary>
+        /// <seealso cref="System.Collections.ObjectModel.ObservableCollection{TValue}" />
+        /// <seealso cref="System.Collections.Generic.IReadOnlyDictionary{TKey, TValue}" />
+        /// <seealso cref="System.Collections.Generic.IDictionary{TKey, TValue}" />
+        [Serializable]
         internal class DuplicateKeyException : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

ronaldbarendse: This exception also seems redundant, as it's only used in its parent class (`ObservableDictionary<TKey, TValue>`) and can be replaced with a `ArgumentException` (like https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.add).

ronaldbarendse: The `InvalidOperationException` thrown by `ObservableDictionary<TKey, TValue>` when the key isn't found should also be changed to a `KeyNotFoundException` (like https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.item).

src/Umbraco.Core/Composing/TypeLoader.cs
```diff
@@ -813,11 +814,44 @@ public void Add(Type type)
         }
 
         /// <summary>
-        /// Represents the error that occurs when a type was not found in the cache type
-        /// list with the specified TypeResolutionKind.
+        /// Represents the error that occurs when a type was not found in the cache type list with the specified TypeResolutionKind.
         /// </summary>
+        /// <seealso cref="System.Exception" />
+        [Serializable]
         internal class CachedTypeNotFoundInFileException : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

src/Umbraco.Core/Exceptions/DataOperationException.cs
```diff
@@ -1,21 +1,94 @@
...
 
 namespace Umbraco.Core.Exceptions
 {
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <seealso cref="System.Exception" />
+    [Serializable]
     internal class DataOperationException<T> : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

src/Umbraco.Core/Exceptions/DataOperationException.cs
ronaldbarendse: As the exception is currently internal and the only type used for `T` is `MoveOperationStatusType`, we might just remove the generics, right?

src/Umbraco.Core/Migrations/DataLossException.cs
```diff
@@ -1,22 +1,45 @@
...
 
 namespace Umbraco.Core.Migrations
 {
     /// <summary>
-    /// Used if a migration has executed but the whole process has failed and cannot be rolled back
+    /// The exception that is thrown if a migration has executed, but the whole process has failed and cannot be rolled back.
     /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Serializable]
     internal class DataLossException : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

src/Umbraco.Web/Media/Exif/ExifExceptions.cs
```diff
@@ -1,42 +1,87 @@
...
 namespace Umbraco.Web.Media.Exif
 {
     /// <summary>
-    /// The exception that is thrown when the format of the JPEG/Exif file
-    /// could not be understood.
+    /// The exception that is thrown when the format of the JPEG/EXIF file could not be understood.
     /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Serializable]
     internal class NotValidExifFileException : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

src/Umbraco.Web/Media/Exif/ExifExceptions.cs
```diff
@@ -1,42 +1,87 @@
...
     }
 
     /// <summary>
-    /// The exception that is thrown when an invalid enum type is given to an
-    /// ExifEnumProperty.
+    /// The exception that is thrown when an invalid enum type is given to an ExifEnumProperty.
     /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Serializable]
     internal class UnknownEnumTypeException : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

src/Umbraco.Web/Media/Exif/JPEGExceptions.cs
```diff
@@ -1,99 +1,211 @@
...
 {
 
     /// <summary>
-    /// The exception that is thrown when the format of the image file
-    /// could not be understood.
+    /// The exception that is thrown when the format of the image file could not be understood.
     /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Serializable]
     internal class NotValidImageFileException : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

src/Umbraco.Web/Media/Exif/JPEGExceptions.cs
```diff
@@ -1,99 +1,211 @@
...
     }
 
     /// <summary>
-    /// The exception that is thrown when the format of the JPEG file
-    /// could not be understood.
+    /// The exception that is thrown when the format of the JPEG file could not be understood.
     /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Serializable]
     internal class NotValidJPEGFileException : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

src/Umbraco.Web/Media/Exif/JPEGExceptions.cs
```diff
@@ -1,99 +1,211 @@
...
     }
 
     /// <summary>
-    /// The exception that is thrown when the format of the TIFF file
-    /// could not be understood.
+    /// The exception that is thrown when the format of the TIFF file could not be understood.
     /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Serializable]
     internal class NotValidTIFFileException : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

src/Umbraco.Web/Media/Exif/JPEGExceptions.cs
```diff
@@ -1,99 +1,211 @@
...
     }
 
     /// <summary>
-    /// The exception that is thrown when the format of the TIFF header
-    /// could not be understood.
+    /// The exception that is thrown when the format of the TIFF header could not be understood.
     /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Serializable]
     internal class NotValidTIFFHeader : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

src/Umbraco.Web/Media/Exif/JPEGExceptions.cs
```diff
@@ -1,99 +1,211 @@
...
+            : base(info, context)
+        { }
     }
 
     /// <summary>
     /// The exception that is thrown when the length of a section exceeds 64 kB.
     /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Serializable]
     internal class SectionExceeds64KBException : Exception
```
ronaldbarendse: All exceptions should be `public`, as consumers can't catch it otherwise!

ronaldbarendse:
I've done some more cleaning up, so the following exceptions are now removed:
- `EntityNotFoundException`: used once in `MediaController` and replaced with an `HttpResponseException`
- `FileSecurityException`: used once in `PhysicalFileSystem` and replaced with an `UnauthorizedAccessException`
- `WontImplementException`: replaced with an `InvalidOperationException` including a descriptive message

A lot of the exceptions are also `internal` and thus not able to get catched by consumers (e.g. `try {} catch (DataOperationException) {}` is not possible). This might not be a problem it the exception is _always_ cached by an internal method (and not rethrown), but ensuring that might be quite hard...

ronaldbarendse:
I'll redo some of the commits, so it only obsoletes custom unneeded/incompatible (non-serializable) exceptions and removes all usages of them...

Returning different exceptions might cause breaking changes, especially if the base class is different, but I don't think that would be a real issue (might be good to merge this into the next minor version though).

The obsoleted `internal` exceptions _can_ get removed, as those can't be catched anyway. All other should be changed to `public`: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1064-exceptions-should-be-public.

I'll make sure the changes are done in separate commits, so you can always cherry-pick what gets merged 👍

ronaldbarendse:
I've rewritten this PR, so it doesn't remove any of the custom public exceptions (causing a binary breaking change). We should continue the conversation on PR https://github.com/umbraco/Umbraco-CMS/pull/6599.

Issue #5909, opened 2019-07-16
Cannot save and losing data in nested content after 8.1 upgrade (from 8.0.2)


miackers:
- When trying to save data, the data gets lost
- When editing nodes with nested content (from previous install), some of them lose their data.

Reproduction
------------

### Bug summary

When saving/editing nodes with nested content, some of them lose their data

### Specifics

This seems to happen when the nested content element is set to "allow varying by culture", but none/some of the element's properties is not set to vary.

### Steps to reproduce

1. Create doctype set to vary by culture, and make it an element type
2.1 Add property to the doctype, but not set it to vary.
2.2 (Optional: Add another but set it to vary - to see the difference)
3. Create nested content data type referencing above doc type (easiest to set min and max to 1, but not required)
4. Add nested content to your home node (for example). 
5. Ensure your site have more than one language (works if only 1 language - I think)
6. Try save information into nested content 

### Expected result

Content to save (and not lose data)

### Actual result

save loses data (and previous content from v8.0.2 gets deleted)

miackers:
I am still doing some investigations around this, so will update as I go along - but wanted to raise to see if anyone else is having the same issue/can be reproduced on another instance

miackers:
I've now managed to reproduce this in a clean Umbraco 8.1 installation, and updated the steps to reproduce this problem

kjac:
This might be a duplicate of #5817. 

@miackers let me see if I understand this correctly. You have defined the element type (used in Nested Content) as culture variant, and added a mix of culture variant _and_ invariant fields to the element type?

miackers:
Yes and no. Its not a requirement to have a mix - ie, all the fields can be invariant fields, and the issue will happen (I know in that scenario it doesnt make sense for the element type to be set as to have culture variants) 

One might ask why enable the nested element to have a culture variant - since currently (correct me if I am wrong) you can't make use of that functionality - ie, the UI doesnt allow you to inherit some fields from different languages anyway (I assumed this would work for 1 max/1 min, but not even for that?!)

So, currently for me the "fix" is to remove the "allow varying by culture" from my element types.

miackers:
updated steps to produce to make clear its an element type

kjac:
I've found the problem - will submit a fix shortly.

kjac:
PR In #5952

mikeallen222:
Thanks @kjac. I ran into the same bug after trying to make a nested content type with variant culture in 8.1.

[...]

Issue #5817, opened 2019-07-09
v8.1 Avoid data loss when using culture variant elements in Nested Content


thomashdk:
I just updated to version **Umbraco 8.1** on cloud.

I have notest that Nested Content inside Nested Content are not saved. 

![image](https://user-images.githubusercontent.com/6445636/60889431-48e0ec80-a259-11e9-9356-331f30f88985.png)

![image](https://user-images.githubusercontent.com/6445636/60889458-54341800-a259-11e9-907f-d39fba0a79fb.png)

When publishing every thing looks fine, but if you reload page you can see that the content are not saved.
I have looked in the database, and it's not saved there.
![image](https://user-images.githubusercontent.com/6445636/60889482-61510700-a259-11e9-9b0d-da7d1d2f34ce.png)

Level 1. Nested are saved, but not level 2. Nested Content 
![image](https://user-images.githubusercontent.com/6445636/60889500-6d3cc900-a259-11e9-94de-e74614ffd74f.png)

thomashdk:
After deleting the datatype and create a new one it worked.

thomashdk:
I opened this again because i'm still getting a similar error. 

Sometimes the content are not saved to database... ex. here I have en content picker en selectes a local node and after I have save and published and reload the page, the contetn are gone.

![image](https://user-images.githubusercontent.com/6445636/60963310-b865e300-a310-11e9-88a5-fde9ba70991c.png)

The same happens with the form componenet I have. All the fields are empty after save/publish and relaod page. 
The DB don't have the content saved
![image](https://user-images.githubusercontent.com/6445636/60963408-f9f68e00-a310-11e9-9d64-101dc8651e81.png)

thomashdk:
After some debug I notes that some of my other components worked. 

If the fields did'nt had the "Allow varying by culture" active, it wouldn't save the content.
The doctype with the fields were set to allow carying by culture"
![image](https://user-images.githubusercontent.com/6445636/60964126-d3d1ed80-a312-11e9-8f6b-f88ff0834663.png)

kjsguld:
I've just seen the same behavior on 8.1 as reported. 
Triggered the same behavior and resolution on a node on a list view, that had nested content (where the content ie. images and drop-downs can't be saved). 
Thanks for spreading the word on this one @thomashdk 🐛

kjac:
[...]

PR #5952, opened 2019-07-19
V8: Support mixed element variance in Nested Content

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5909 and https://github.com/umbraco/Umbraco-CMS/issues/5817

### Description

Nested Content doesn't like mixed variance on the element types used. Here's a brief summary of the problem, see #5909 and #5817 for further details.

#### Step 1: Create the element type

Create an element type configured as culture variant and configure culture *in*variance on at least one the properties - like this:

image

#### Step 2: The rest of the setup

Create a Nested Content data type that uses the above created element type. Use a culture invariant property as label template:

image

Create a culture variant doctype and add the Nested Content data type as a property.

image

#### Step 3: Test!

Create content based on the doctype and add one or more items to the Nested Content property. Observe how the label template doesn't work. Now save the content and reload the page. The data in the culture invariant Nested Content properties is gone.

Here's how it could play out:

nc-variant-invariant-before

#### Verifying this PR

Using the same setup and approach, here's how the same scenario plays out with this PR applied.

nc-variant-invariant-after


emmaburstow:
Kenn! Thanks for the steps to test here! Have I ever told you you're a ⭐️ ? Probably.

We'll ask if we have any questions. 

Em

kjac:
PR updated with latest v8/dev

nul800sebastiaan:
Wonderful, that works like a charm! Really good find as well, that was a tough one to reproduce. 😮

PR #5955, opened 2019-07-19
Fixed GetCurrentLoginStatus() summary

FransdeJong:
Small change but it saves some debugging.

In the summary of MembershipHelper.GetCurrentLoginStatus() it states that null will be returned when no member is logged in but that isn't the case (anymore?).


nul800sebastiaan:
Thanks Frans! It hasn't returned `null` for at least 5 years if I look at the git blame, so I doubt it ever did. ;-)

Issue #5913, opened 2019-07-16
NC keys: When copying page nested content does not update properly


deMD:
This issue description has been updated with a propsed solution and has been marked as up for grabs. Please read below about the original issue and the proposed solution. You can also scroll way down to read the entire discussion and history.

### Proposed solution

Taking some inspiration from the media tracking branch and looking at what we already have in core:

* Nested content (and any other property editor) can handle the `IDataValueEditor.FromEditor` to re-populate any missing keys. Nested content already overrides this method so it would be just adding the logic to check for empty keys. Empty keys may be sent up if the client side clears them when copying them with the clipboard service. Nested content calls FromEditor recursively so if there's an NC inside an NC this will still work.
* Create a new interface: `IDataValueCopying` which can be implemented by any `IDataValueEditor` (like like `IDataValueReference` for media tracking). It could have one method like: `object Copy(object input);`
* We create an external component to handle `ContentService.Copying`, the event handler will use the current `IContent` along with the `PropertyEditorCollection` to figure out if any of the `IDataValueEditor`s being used for any of the properties implement `IDataValueCopying` and if so, will invoke the `Copy` method and set the new property value to the result.
* the `NestedContentPropertyValueEditor` then implements this interface and it will have to deal with having nested nested contents but that shouldn't be too difficult.

## Original submitted issue

When we copy pages with Nested Content and edit a published element the copied page still shows values from the original content. For example an image block where we changed the image still gives us the original image when checking the strongly typed model. The source value of the property does update properly, as demonstrated in the image below:

![image](https://user-images.githubusercontent.com/22070292/61291409-be573a80-a7ce-11e9-843e-442e62c2598b.png)

Reproduction
------------
- Create a page with nested content
- copy the page
- change a property of a block in the nested content
- check the return values.

This bug appears in Umbraco version 8.0.2

---
_This item has been added to our backlog [AB#6684](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6684)_

kjac:
@deMD I'm not sure I follow. Would you expect the elements on the copy to update automatically when you update elements on the original?

deMD:
No the other way around.

Here is a screenshot of the current environment:
![image](https://user-images.githubusercontent.com/22070292/61291970-425df200-a7d0-11e9-9a17-12603cbc92ea.png)

[...]

PR #5706, opened 2019-06-22
V8: Copy on property level

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes #5551

### Description

For background info on this PR see #5551 and #5599.

This PR introduces the option to define a property editor as copyable. It's done by setting `CanCopy = true` in the `DataEditor` attribute. This results in a copy icon being displayed next to the property label:

image

It is of course up to the specific property editor implementation to handle the copy action. The backoffice simply broadcasts a `propertyCopy` event within the editor scope for the property editor controller to handle - something like this:

```js
$scope.$on("propertyCopy", function (ev, args) {
  // implement copy handling here
});
```

This PR also implements copy handling for Nested Content. It works like this:

nc-copy

#### Limitations

Obviously this feature doesn't work if the property label is hidden. As discussed with @nielslyngsoe on #5599, we need to supply a means to change the property label placement instead of hiding it, so property editors can make use of the full screen width and still have their label. This will be a separate feature/PR. 

#### Future

If this is merged in, the obvious next step would be to implement similar handling for Grid (and possibly other property editors that can be time-consuming to work with).

#### Testing this PR

1. Ensure that all items on a Nested Content property can be copied to another Nested Content property on content of the same type.
2. For language variant content where the Nested Content property is allowed to vary by language, ensure that all items can be copied between languages.
3. For language variant content where the Nested Content property is *not* allowed to vary by language, verify that the property copy icon is *only* present on the main language variant.
4. Verify that you can't copy items from one Nested Content to another if one or more of the copied item types aren't allowed on the target property configuration.
5. Ensure that single items can be copied between Nested Content properties (that this PR didn't break the single item copy feature, in other words).

PR #5961, opened 2019-07-19
V8: Force cachebust Nested Content items to handle content copying

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5913

### Description

See #5913 for a detailed description. 

The gist is: When copying a Nested Content property, the element keys embedded in the property data is obviously copied along with it. Because the element cache is based on the element keys, this becomes a problem if you want to render both the original and the copy within the same request.

#### Steps to reproduce

1. Create a Nested Content datatype using an element type with a property of type Textstring named  `Text`.
2. Add the datatype to a doctype and name it `List`.
3. Create a piece of content based on this doctype and add a few items to `List`.
4. Create a few copies of the content. Edit `Text` of all items in each copy so they're unique.
5. Render the original content using the template below. Replace the values of `contentIds` with the IDs of the original and the copies.
6. Notice how the rendered output is the same for each content, despite all values being unique.

Here's how it looks:

nc-copy-cache-before

...and here's the template for testing:

```cshtml
@{
	Layout = null;
	// TODO: replace 1111, 2222, 3333, 4444 with the IDs of the original and all copies 
	var contentIds = new [] {Model.Id}.Union(new [] {1111, 2222, 3333, 4444}).Distinct().ToArray();
}

<html>
    <body>
        @foreach(var contentId in contentIds) {
            var content = Umbraco.Content(contentId);
            var list = content.Value<IEnumerable<IPublishedElement>>("list");
            <h3>@content.Name (@contentId)</h3>
            <ul>
            @foreach(var element in list) {
                <li>@element.Value("text")</li>        
            }
            </ul>
        }
    </body>
</html>
```

#### The fix

This PR fixes the caching issue by assigning a new `Guid` to each element every time the cache is populated, instead of using the `key` property assigned to the element (the one that's copied along with the Nested Content data).

When applied, the copies render as you'd expect:

nc-copy-cache-after

#### But it's a hack?!

Yes. It is a hack. But right now I don't see any other (feasible) way around it.

I have tried changing the value of Nested Content's `PropertyCacheLevel`. As you'd expect that didn't work, but it was worth a shot.

I tried updating the `key` value of each element on save in [this method](https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs#L201). It works. But it is just as much a hack, and I'm not terribly fond of modifying the keys created on the client as they're actively used in the clientside controller for element identification. It also is an irreversible change to the editor data, whereas the proposed solution can be rolled back later on in favor of a non-hacky solution.

#### Is there a penalty?

I have not been able to identify any penalties to this solution, performance or caching wise. The elements are still cached on first hit, and subsequent renderings of the content or other content based on the same type and using the same template do not incur any extra rendering or caching of these elements.

#### Suggested approach (future)

One less hacky solution would be to propagate `EntityBase.ResetIdentity()` to the individual property implementations and let them modify the property data accordingly. 

This however would entail a performance impact, as each property would have to wire up their implementations, apply the property data and execute the identity reset. It is also likely to be a breaking change.


emmaburstow:
Brilliant, brilliant stuff. I love a good hack! We'll probably have to have a good chat about this one. Thanks for putting so much detail here though Kenn. It gives us plenty to work with.

Speak soon,

Em

ronaldbarendse:
@kjac Why would you generate random keys to avoid the caching if you can also just change/disable the caching itself on the value converter? This just creates unwanted side-effects, for example when a site uses the (static) key to identify the NC item on that page (sections, anchors, selects/dropdowns, etc.).

The `NestedContentValueConverterBase` doesn't override the `GetPropertyCacheLevel` method, but both `NestedContentSingleValueConverter` and `NestedContentManyValueConverter` do and set this to `PropertyCacheLevel.Element`. This can be easily changed to disable caching (although it's probably better to merge this by moving it to the base class):

```c#
public override PropertyCacheLevel GetPropertyCacheLevel(IPublishedPropertyType propertyType) => PropertyCacheLevel.None;
```

What strikes me as odd, is the comment for `PropertyCacheLevel.Element`:
https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Web/PublishedCache/PublishedElementPropertyBase.cs#L131-L132

This would make you believe the current cache level would be correct and cache the NC item per content/page, which it clearly isn't!

kjac:
@ronaldbarendse thank you for investigating. Unfortunately (as I wrote in the initial description) I did try changing the property cache level on the value converter implementations to no avail. The elements are cached by key in the current request. Hence this solution.

ronaldbarendse:
@kjac Just to be sure: did you change it in both the `NestedContentSingleValueConverter` and `NestedContentManyValueConverter`, as these override any cache level set in `NestedContentValueConverterBase`? And what cache levels are passed into the `ConvertIntermediateToObject` and therefore `ConvertToElement` methods?

If disabling the value converter cache still caches the element, that's probably done in the `IPublishedModelFactory` or the `IPublishedElement` itself...

Anyway: adding this hack would only introduce other unwanted side-effects, while the issue only occurs if NC items with the same key (caused by copying content) are retrieved _in the same request_.

kjac:
@ronaldbarendse I did change it in both converters, yes 😄 

Disabling the cache entirely is not an option if you ask me. The ideal solution to this problem would be to regenerate the keys on copy, as mentioned above. However I don't think that's feasible short term.

stevemegson:
Although the nested content itself is set to PropertyCacheLevel.Element, each PublishedElementPropertyBase within the element tries to take into account the cache level requested by its own value converter and choose the "shorter-lived" cache. If the "inner" value converter wants PropertyCacheLevel.Snapshot (as TextStringValueConverter does) then Snapshot will be used for that property, leading to the clash between different elements with the same key.

The simplest solution may be to have PublishedElementPropertyBase.GetCacheLevels never return Snapshot, and instead return None when ReferenceCacheLevel is Element and PropertyType.CacheLevel is Snapshot.

There'd be some performance impact if you're repeatedly accessing the same (non-copied) element in the same request, but hopefully that's minimal. If we really want to avoid that then we could have PublishedElementPropertyBase use a random Guid in its ValuesCacheKey rather than using Element.Key.

ronaldbarendse:
So the NC element itself isn't globally cached (not `Elements` or `Snapshot`), but the properties within that item might be and would then use the element key in the cache key, which isn't unique enough (as it can be copied).

I would refrain from generating unique keys (besides the already noted side-effects, this could fill up the cache with duplicate items on different cache keys!) or re-generating the key on copy/save, as that would require implementations for every property editor that uses elements!

@Shazwazza @nul800sebastiaan This should be fixed properly, probably by including more info in the property cache key:
https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Web/PublishedCache/PublishedElementPropertyBase.cs#L65-L66
https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Web/PublishedCache/PublishedElementPropertyBase.cs#L77-L80

Adding the `IPublishedContent`s key would fix this, but only the `IPublishedElement` is available in `PublishedElementPropertyBase` and we should also make sure this works correctly for deeper nested elements, e.g. Nested Content within Nested Content...

ronaldbarendse:
@Shazwazza @nul800sebastiaan Do you have any feedback on my last comment? Fixing this would require some major changes in the way the cache keys can be generated or to ensure the IDs can never be the same...

Besides copying a single NC item, PR https://github.com/umbraco/Umbraco-CMS/pull/5706 is already merged for 8.2.0, so duplicate element IDs may become even more common...

lcichanowicz:
@emmaburstow, it seems this is still an issue. Might there be a related issue with an update?

Shazwazza:
Will close this PR as we'll need to implement something a little different to maintain unique/consistent keys when copying.

enkelmedia:
I have the same problem on 8.5.5, @Shazwazza is the "something" that you are referring to implemented? Or is there any known work around?

nul800sebastiaan:
@enkelmedia Some exciting reading awaits you in issue https://github.com/umbraco/Umbraco-CMS/issues/5913 😉

In short, this is not fixed yet and the issue is up for grabs with a description on how we'd like to see it fixed.

PR #5966, opened 2019-07-21
Handle CSV values in TagsPropertyEditor (#5855)

### LINKED ISSUE #5855: V8/8.1: Tags erased when you "save and publish" an existing page
### Bug summary

I have a composition with a tags field. When I unpublish a page that uses the composition, the tags remain intact. When I "Save and Publish" the page the tags field is not saved. Re-entering the tags and re-saving the page works as expected.

### Specifics

* Happens in Umbraco 8.1
* Testing in Brave and Google Chrome on macOS Mojave
* Rebuilding the page cache does not help
* Editing the page after the tags have been cleared allows me to save new tags, so it appears to only happen when publishing a page with existing tags

### Steps to reproduce

* Edit an unpublished page with tags, click the "Save and Publish" green button; unpublish/publish in a list view works fine

### Expected result

* Tags should remain intact during publish/unpublish

### Actual result

* Tags do not get saved on re-publish unless the tags field is empty.

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #5855
### Description

Handles CSV values being returned from the tags editor, which happens when the data type is configured for CSV storage and the page is saved without modifying the tags,

To test:
- Create a Tags property with storage type set to CSV
- Create and save a document with some tags populated
- Reopen the document and save it without making any changes to the tags
- Reopen the document again and confirm that the tags are not blank

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/TagsPropertyEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/TagsPropertyEditor.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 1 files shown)


Shazwazza:
@stevemegson Any chance during your investigation you figured out when/how this stopped working since of course this all works in v7 - so i want to make sure we are changing the right things in the right places.

stevemegson:
It seems to have been a change to avoid triggering 'unsaved changes' messages when the only change was the editor converting the CSV value to JSON. I'm not sure why that wasn't a problem in v7, though.

Presumably that could also be fixed on the client, but it seemed reasonable to expect that an implementation of `DataValueEditor.FromEditor` should handle any value that `ToEditor` produced without assuming that the editor will have made changes.

nul800sebastiaan:
Thanks @stevemegson - these changes are great. I made them a little more robust here https://github.com/umbraco/Umbraco-CMS/commit/ca75a258029bf076ed1c00bbeb7e7456da2f10cd - I don't think it was necessary, but just making sure to check everything for null etc.

PR #5967, opened 2019-07-21
Content migration - typed labels based on alias rather than ID

### LINKED ISSUE #5908: Upgrade from 7.15 to 8.1 fails: "System.FormatException: The string was not recognized as a valid DateTime. There is an unknown word starting at index 0"
We have a database that was upgraded from Umbraco 6.x to 7.15 over time and now we are trying to upgrade to 8.1.

This migration error happens when Umbraco is trying to parse the values of certain hardcoded property types that it assumes to be DateTime properties in [AddTypedLabels.cs](https://github.com/umbraco/Umbraco-CMS/blob/release-8.1.0/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/AddTypedLabels.cs#L76):

image

But in our database these IDs correspond to these property types:

image

So it inevitably encounters some strings that cannot be parsed as DateTime.

---
_This item has been added to our backlog [AB#1757](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/1757)_

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #5908

### Description
During content migration, AddTypedLabels currently assumes that the properties to update exist with the standard IDs used in a fresh install. If the site has been upgraded from an old version then those IDs may be used by other properties which should not be updated. If the user has created their own media or member types with the standard property aliases, those properties will not be updated.

Testing:
- In 7.15..0, copy the default Image media type and create an item using the new type.
- Edit the original Image media type. Change the alias of the `umbracoWidth` property to `test` and switch it to a Textbox data type. Create a media item with a non-numeric value for that property.
- Upgrade to 8.1.0.
- Confirm that the width, height and size properties of the copied media type are updated to use the new typed labels
- Confirm that the `test` property of Image is unchanged.


stevemegson:
I hadn't intended to include that CheckBoxList commit on the same branch. It's unrelated to typed labels, but simply updates ValueListPreValueMigrator to handle the prevalues for CheckBoxList in the same way as it handles RadioButtonList.

ronaldbarendse:
@stevemegson You can copy/move/[cherry pick](https://git-scm.com/docs/git-cherry-pick) commit https://github.com/umbraco/Umbraco-CMS/pull/5967/commits/968463912ab548808ff87bc58760b2b2bc7bb45e to a different branch and update this branch to only include the specified fix (commit https://github.com/umbraco/Umbraco-CMS/pull/5967/commits/afb61de156adb5ed84036c90e1749baaa0821bc5). That way, the fix can be tested in isolation and will get it's own release note...

stevemegson:
@ronaldbarendse I think that's done the right thing...

elit0451:
Hi @stevemegson 

I have followed your test notes and the end result for me was that the `width`, the `height` and `size` properties of the copied Image media type haven't changed - same aliases (`umbracoWidth`, `umbracoHeight`, `umbracoBytes` ), same editors (Label). While all the media items created from the changed default Image media type have their **Width** property with alias `test`.

Was this the intended behaviour you were looking for?

stevemegson:
Yeah, I could have made it more clear what I meant by "updated" and "unchanged", couldn't I?

Everything should still be using the `Umbraco.Label` editor, but the properties on the copied media type should have been changed to use the `Label (integer)` and `Label (bigint)` data types. The renamed `test` property should be left using the `Label (string)` data type, where the previous behaviour would try to convert it to `Label (integer)`.

elit0451:
That gives me a much better idea of what I need to be looking for 😊 Thanks!

nul800sebastiaan:
@stevemegson The `test` property should still be a textbox, right, not a label?

stevemegson:
@nul800sebastiaan ah yes, I forgot that these test notes changed it to a text box to let you enter the non-numeric value. I was editing the database directly in my tests, so it's still a label in my test site.

elit0451:
Tested and it works! Merging 👍

PR #5992, opened 2019-07-24
V8/bugfix/5554 unpublish state

Shazwazza:
Fixing https://github.com/umbraco/Umbraco-CMS/issues/5554

__Important: merge this PR first https://github.com/umbraco/Umbraco-CMS/pull/6020__

Few issues:

* we weren't unpublishing the document when the last culture was unpublished
* we weren't returning the correct status when the default culture got unpublished
* we weren't saving the document in `SaveAndPublish(IContent content, string[] cultures` overload when there were no cultures - the document should still be saved, just no cultures published. We were just exiting with the status FailedPublishNothingToPublish without saving, it needs to save and then return the status FailedPublishNothingToPublish
* adds new status SuccessUnpublishLastCulture when we are unpublishing the last culture and thus also the document
* adds a few unit tests to support all of this 

Other changes in this PR:

* ContentService was always re-looking up languages and sure the language repo caches things but it's silly to rely on that, instead where needed we pass around a reference to all of the langs already looked up
* We were previously looking up tagEditors in ContentRepositoryBase to pass along to GetPropertyCollections but these were never used, so i've removed that code since it was extra overhead
* Added some notes for future refs

### DIFF
--- a/src/Umbraco.Core/Exceptions/PanicException.cs
+++ b/src/Umbraco.Core/Exceptions/PanicException.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace Umbraco.Core.Exceptions
+{
+    /// <summary>
+    /// Internal exception that in theory should never ben thrown, it is only thrown in circumstances that should never happen
+    /// </summary>
+    [Serializable]
+    internal class PanicException : Exception
+    {
+        public PanicException()
+        {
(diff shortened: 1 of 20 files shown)


bergmania:
#6020 merged..

bergmania:
Changes looks good.. I have tested and stuff seems to work as expected. 
Good work with the integration tests :)

PR #5998, opened 2019-07-25
V8: Set the form as dirty when adding/removing colors to/from Umbraco.ColorPicker

kjac:
### Prerequisites

### Description

When configuring Umbraco.ColorPicker, there's no "discard changes" warning if you accidentally leave the config without saving:

approved-color-setdirty-before

I ran into this issue yesterday and lost my config. So here's a PR that fixes it 😄 

_Update: This also applies to sorting of colors. As it turns out there was a bit of deprecated code going on in the `update` function for color sorting, so I removed it while I was at it._

#### Testing this PR

1. Verify that Umbraco.ColorPicker prompts you to discard changes if you add or remove colors and then leave the config without saving.
2. Verify that Umbraco.ColorPicker prompts you to discard changes if you reorder the colors and then leave the config without saving.
3. Verify that Umbraco.ColorPicker saves reordered colors correctly.


bjarnef:
@kjac maybe the form should also be set as dirty, when sorting the colors?
https://github.com/umbraco/Umbraco-CMS/blob/f64b8c2ed1279e8fc3b8ec172744672006808cf4/src/Umbraco.Web.UI.Client/src/views/propertyeditors/colorpicker/multicolorpicker.controller.js#L141-L154

It happens in some part of the UI, but not sure if it is consistent everwhere.
https://github.com/umbraco/Umbraco-CMS/blob/4733256ca1bfb49d4fe946110e734db1351f81d9/src/Umbraco.Web.UI.Client/src/views/propertyeditors/multiurlpicker/multiurlpicker.controller.js#L27

kjac:
@bjarnef you're right. I have updated the PR - thanks 👍 

Turns out all the code in the `update` function was deprecated, so I did a bit of housekeeping.

bjarnef:
@kjac didn't it use the function `getElementIndexByPrevalueText` here?
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/propertyeditors/colorpicker/multicolorpicker.controller.js#L144

I think I added the `ng-model` here some time ago, because it is required by `sortable-ui`.
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/propertyeditors/colorpicker/colorpicker.prevalues.html#L12

Not sure it this part in update function still is relevant:
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/propertyeditors/colorpicker/multicolorpicker.controller.js#L140-L152

kjac:
@bjarnef so it was you 😄 

Adding `ng-model` to the sortable was the right thing to do. It also made all the code in `update` obsolete, as that code was basically trying to do the same thing that adding `ng-model` to sortable does. So I have removed all the old code in `update` and in turn `getElementIndexByPrevalueText` (which was only used by `update`).

kjac:
PR updated with latest v8/dev

nul800sebastiaan:
Thanks, works like a charm! 👍

Issue #5995, opened 2019-07-24
Unable to remove start nodes


Luksor:
Reproduction
------------

### Bug summary

![image](https://user-images.githubusercontent.com/5837521/61802410-491be300-ae28-11e9-9c6a-607af71fe8ef.png)

There is no way of removing a content / media start node from a user.

### Specifics

* Umbraco version: 8.1.0

### Steps to reproduce

* Assign a start node to a user.
* The option to unassign the node is missing.

### Expected result

There should be a way of removing a node assignment.

### Actual result

There is no option allowing removal of node assignment.

kjac:
Oh nice one! Thanks @Luksor, looks like it was half way implemented but never completed. I'll submit a PR shortly 👍

kjac:
PR in #5999

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/5999 and cherry picked for 8.1.2 in https://github.com/umbraco/Umbraco-CMS/commit/a9a5d00916d2ce1129a2963619167eb48eba2a8a

PR #5999, opened 2019-07-25
V8: Allow removing user start nodes

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5995

### Description

See #5995 for details.

As it turns out, the user editing implementation already supports removal of user start nodes - it's just not been enabled in the view. So... this PR just does that.

When this PR is applied, user start nodes can be removed for both content and media:

remove-user-start-nodes


emmaburstow:
Hiya @kjac 

Thanks for the bug fix! And of course the documentation. 

Em

nul800sebastiaan:
Thanks @kjac - that was silly to forget hehe! 😅

PR #6001, opened 2019-07-25
V8: Date pickers - parsing non-standard formats

### LINKED ISSUE #5968: Updating from 8.0.2 to 8.1, moment.min.js reporting deprecation warning
Date picker in back office has stopped working after upgrading from 8.0.2 to 8.1.

### Specifics

This warning appears in the console when selecting a date in the backoffice. Save & Publish clears the field and trying to access field in view returns DateTime.MinValue. 

`moment.min.js?cdv=126457885:formatted:134 Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
Arguments: 
[0] _isAMomentObject: true, _isUTC: false, _useUTC: false, _l: undefined, _i: 21-07-2019, _f: undefined, _strict: undefined, _locale: [object Object]
Error
    at Function.createFromInputFallback (http://localhost:3000/umbraco/lib/moment/moment.min.js?cdv=126457885:1:3368)
    at Yt (http://localhost:3000/umbraco/lib/moment/moment.min.js?cdv=126457885:1:21353)
    at Ot (http://localhost:3000/umbraco/lib/moment/moment.min.js?cdv=126457885:1:22064)
    at Tt (http://localhost:3000/umbraco/lib/moment/moment.min.js?cdv=126457885:1:22146)
    at c (http://localhost:3000/umbraco/lib/moment/moment.min.js?cdv=126457885:1:209)
    at setDate (http://localhost:3000/umbraco/js/umbraco.controllers.js?cdv=126457885:15515:30)
    at ChildScope.$scope.inputChanged (http://localhost:3000/umbraco/js/umbraco.controllers.js?cdv=126457885:15497:13)
    at fn (eval at compile (http://localhost:3000/umbraco/lib/angular/angular.js?cdv=126457885:16387:15), <anonymous>:4:156)
    at callback (http://localhost:3000/umbraco/lib/angular/angular.js?cdv=126457885:28815:13)
    at ChildScope.$eval (http://localhost:3000/umbraco/lib/angular/angular.js?cdv=126457885:19356:28)`

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #5968

### Description
Date pickers cause issues if a non-default format like DD-MM-YYYY is configured. Moment.js is asked to parse the formatted value, which at best causes a javascript warning and at worst leaves the field back or with the wrong date. In a future version, Moment.js will remove support for the fallback parsing which currently triggers the warning, causing further problems.

This change passes the configured format string to Moment.js so that the value can be parsed properly. Since the user could still type a date in another format, it falls back to the browser's date parsing in a way which keeps Moment.js happy.

To test:
- Configure the date picker with a non-default format, such as DD-MM-YYYY.
- Open the picker by clicking in the text box, rather than on the button next to it.
- Pick 25 July 2019 (to cause a problem if 25-07-2019 is parsed as MM-DD-YYYY) and move focus to another property.
- There should be no javascript warning from moment, and the date should appear as expected.
- Type "1 Jan 2000" in the date picker's text box, and move focus to another field.
- The value should change to "01-01-2000", again with no javascript warning.
- Open the date picker again, and 1 January 2000 should be pre-selected.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
@@ -78,9 +78,20 @@ function dateTimePickerController($scope, notificationsService, assetsService, a
         setDatePickerVal();
     };
 
-    $scope.inputChanged = function() {
-        setDate($scope.model.datetimePickerValue);
-        setDatePickerVal();
+    $scope.inputChanged = function () {        
+        if ($scope.model.datetimePickerValue == "" && $scope.hasDatetimePickerValue) {
+                $scope.clearDate();
+        } else if ($scope.model.datetimePickerValue) {
+            var momentDate = moment($scope.model.datetimePickerValue, $scope.model.config.format, true);
+            if (!momentDate || !momentDate.isValid()) {
+                momentDate = moment(new Date($scope.model.datetimePickerValue));
+            }
+            if (momentDate && momentDate.isValid()) {
+                setDate(momentDate.format("YYYY-MM-DD HH:mm:ss"));
+            }
+            setDatePickerVal();
+            flatPickr.setDate($scope.model.value, false);
+        }
     }
     
     //here we declare a special method which will be called whenever the value has changed from the server
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
@@ -74,20 +74,23 @@ function dateTimePickerController($scope, notificationsService, assetsService, a
     };
 
     $scope.datePickerChange = function(date) {
-        setDate(date);
+        const momentDate = moment(date);
+        setDate(momentDate);
         setDatePickerVal();
     };
 
     $scope.inputChanged = function () {        
-        if ($scope.model.datetimePickerValue == "" && $scope.hasDatetimePickerValue) {
-                $scope.clearDate();
+        if ($scope.model.datetimePickerValue === "" && $scope.hasDatetimePickerValue) {
+            // $scope.hasDatetimePickerValue indicates that we had a value before the input was changed,
+            // but now the input is empty.
+            $scope.clearDate();
         } else if ($scope.model.datetimePickerValue) {
             var momentDate = moment($scope.model.datetimePickerValue, $scope.model.config.format, true);
             if (!momentDate || !momentDate.isValid()) {
                 momentDate = moment(new Date($scope.model.datetimePickerValue));
             }
             if (momentDate && momentDate.isValid()) {
-                setDate(momentDate.format("YYYY-MM-DD HH:mm:ss"));
+                setDate(momentDate);
             }
             setDatePickerVal();
             flatPickr.setDate($scope.model.value, false);
@@ -103,15 +106,14 @@ function dateTimePickerController($scope, notificationsService, assetsService, a
             var newDate = moment(newVal);
 
             if (newDate.isAfter(minDate)) {
-                setDate(newVal);
+                setDate(newDate);
             } else {
                 $scope.clearDate();
             }
         }
     };
 
-    function setDate(date) {
-        const momentDate = moment(date);
+    function setDate(momentDate) {        
         angularHelper.safeApply($scope, function() {
             // when a date is changed, update the model
             if (momentDate && momentDate.isValid()) {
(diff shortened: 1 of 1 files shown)


poornimanayar:
Hi @stevemegson ,

Many thanks for your PR. We shall test and get back to you in the next 14 days. Watch the space :-)

Regards,
Poornima, [Pull Request team](https://umbraco.com/blog/introducing-the-new-cpr-team/)

src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
```diff
@@ -78,9 +78,20 @@ function dateTimePickerController($scope, notificationsService, assetsService, a
         setDatePickerVal();
     };
 
-    $scope.inputChanged = function() {
-        setDate($scope.model.datetimePickerValue);
-        setDatePickerVal();
+    $scope.inputChanged = function () {        
+        if ($scope.model.datetimePickerValue == "" && $scope.hasDatetimePickerValue) {
```
nul800sebastiaan: Can you explain this one? If the value on the model is empty AND `hasDatetimePickerValue` is true? When would this be the case?

Also makes sure to use a triple equals there: `===`

src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
```diff
@@ -78,9 +78,20 @@ function dateTimePickerController($scope, notificationsService, assetsService, a
...
+                $scope.clearDate();
+        } else if ($scope.model.datetimePickerValue) {
+            var momentDate = moment($scope.model.datetimePickerValue, $scope.model.config.format, true);
+            if (!momentDate || !momentDate.isValid()) {
+                momentDate = moment(new Date($scope.model.datetimePickerValue));
```
nul800sebastiaan: For my understanding: this happens when the first time parsing the date with the current configuration fails, correct? When would we get `!momentDate` and when `!momentDate.isValid()`?

src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
```diff
@@ -78,9 +78,20 @@ function dateTimePickerController($scope, notificationsService, assetsService, a
...
+            }
+            if (momentDate && momentDate.isValid()) {
+                setDate(momentDate.format("YYYY-MM-DD HH:mm:ss"));
```
nul800sebastiaan: Why do we hardcode a format here instead of using the one from `$scope.model.config.format`?

src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
```diff
@@ -78,9 +78,20 @@ function dateTimePickerController($scope, notificationsService, assetsService, a
...
+            }
+            setDatePickerVal();
+            flatPickr.setDate($scope.model.value, false);
```
nul800sebastiaan: Why do we need to explicitly call `flatPickr.setDate` here? I don't see it being used elsewhere in the code.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
stevemegson: Yeah, in hindsight it could do with at least a comment to explain what it's checking for. `$scope.hasDatetimePickerValue` hasn't been updated yet, so it tells us whether we had a value before the text box changed. So the test makes sure that we only call `$scope.clearDate` if there's a previous date to clear. This matters because clearDate will mark the form as dirty. If you start with a blank date and then trigger inputChanged with an empty value, we don't want that to trigger "unsaved changes" messages.

stevemegson:
At the risk of confusing the comments above, I've pushed another commit to get rid of the redundant conversions between strings and moments around `setDate` and `updateModelValue`.

rbottema:
This issue has been biting us. Is there currently a timeline which we can tell our customer when this PR will be merged?

taraldr:
Anyone able to complete this PR?

nul800sebastiaan:
After some feedback from other people, I've looked at this code again, it looks good and works as expected. We'll release the fix with v8.6.0.

PR #6008, opened 2019-07-27
V8: Always open the selected content template in infinite editing

### LINKED ISSUE #6006: Changing document type template makes the document type unusable.
Whenever I change a default template in a document type Umbraco is no longer able to find it and render the content properly.

image

image

Reproduction
------------

### Bug summary

Changing the default template produces 404 when rendering content and causes the backoffice to throw a 404 error when trying to open the template from the Content Info tab.

### Specifics

* Tested on Umbraco 8.1.0

### Steps to reproduce

* Create a new document type
* Create a content item using this document type
* Create a new template
* Change the default template in the document type to the newly created one and remove the old one
* Try accessing the page / Opening the template from the content info page.

### Expected result

The page should render using the new template.

### Actual result

The page is returning a 404 error.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #6006 (in part)

### Description

When opening the content template from the content "info" section, the currently assigned template is always opened no matter if you selected another one (assuming you have multiple templates allowed):

open-template-before

The same issue is the root cause of a slightly more annoying (and rather less likely to happen) error, which can be produced by the following steps:

1. Create a content type without a template.
2. Create a content item based on this content type.
3. Assign a template to the content type.
4. Open the content item and switch to the "info" section. The template will list as being already assigned to the content item because it will be once the content item is saved.
5. Open the template in infinite editing. 404 error happens:

image

This PR solves both of these issues by ensuring we always open the currently selected template (by template alias) in infinite editing. 

Here's how things work with the PR applied:

open-template-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbcontentnodeinfo.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbcontentnodeinfo.directive.js
@@ -1,7 +1,7 @@
 (function () {
     'use strict';
 
-    function ContentNodeInfoDirective($timeout, logResource, eventsService, userService, localizationService, dateHelper, editorService, redirectUrlsResource, overlayService) {
+    function ContentNodeInfoDirective($timeout, logResource, eventsService, userService, localizationService, dateHelper, editorService, redirectUrlsResource, overlayService, entityResource) {
 
         function link(scope) {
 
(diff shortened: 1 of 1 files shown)


poornimanayar:
Hey @kjac ,

Thanks once again for the PR. You know how it goes from here :-)

Poornima

nul800sebastiaan:
Sweet fix @kjac ! 😁

Issue #5987, opened 2019-07-23
Macro in Editor returns Cached data in 8.1.0


philippjenni:
If you use a PartialViewMacroPage with a Macro inside a TinyMCE Editor, the Page returns a cached result of the PartialViewMacroPage.

Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

### Bug summary

Page returns cached macro output for each request in combination of rich text editor.

### Specifics

Problem in Umbraco 8.1.0

### Steps to reproduce

- Create a Partial Macro with an @DateTime.Now.Ticks output
- Create a Marco which uses the Partial Maco and Use in richt text editor and the grid enabled, Cache Period 0, Cache by Page Disable, Cache personalized disabled.
- Create a Page with an TinyMCE Editor Field an use the Macro inside the Editor.
- Plublish and show the page. The ticks will only changed, if you publish again.

### Expected result

Cache Settings should be applied.

### Actual result

Cache Settings are ignored

JannesMeyer:
I think this could be the same issue as #5984 (which was reported yesterday)

stevemegson:
The `RteMacroRenderingValueConverter` sets its cache level to Snapshot, but that isn't actually doing what we want it to. That cache level is used when caching the ObjectValue and XPathValue, but the macro rendering is happening in `ConvertSourceToIntermediate`. The intermediate value is effectively always cached at the Element level.

philippjenni:
This could be a potential security issue. I the macro Display user specific informations, it's possible you can see Information from an other user.

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6010
Cherry picked to 8.1.4 703092c12c979a8902afe7676c18beade4005f99

nul800sebastiaan:
Attached I've added a dll with a workaround for this issue. It is a PropertyValueConvertor that overrides the default on in Core. I had to name it `Umbraco.Tests.dll` because I needed to access some internal methods in Umbraco.

If you need this, please make sure to drop this dll in the bin folder and do not forget to remove it when you upgrade to 8.1.4 (coming out soon!).

[Umbraco.Tests.dll.zip](https://github.com/umbraco/Umbraco-CMS/files/3550418/Umbraco.Tests.dll.zip)

JannesMeyer:
Thanks very much @nul800sebastiaan for fixing this!

[...]

PR #6010, opened 2019-07-28
V8: Fix caching of macro results in RTE

### LINKED ISSUE #5984: Umbraco 8 caching CurrentUmbracoPage
Hi,
We have recently upgraded from Umbraco 6 to 8. One of the big changes in Umbraco is caching. There is an issue with displaying validation errors on some of our pages.

Reproduction
------------

### Bug summary

When having a form in a macro and with required fields, it is expected to see the required fields highlighted if they wee left out blank when submitting a form. However, probably because of caching, I am seeing the same page with no validation feedback when submitting an incomplete form.

### Steps to reproduce

1) Create a controller with `[HttpPost]` method that receives form that follows the standard pattern:

```
if (ModelState.IsValid == false)
{
      return CurrentUmbracoPage();
}
```

2) Create a macroPartial form that will have form fields with validation, e.g.:

```
<div class="input-group">
        @Html.LabelFor(m => model.Name)
        @Html.EditorFor(m => model.Name)
        @Html.ValidationMessageFor(m => model.Name)
    </div>
```

3) Create a corresponding model with `Required` fields.

4) Run the example, try to partially fill in the form and submit it.

### Expected result

Umbraco displaying the same form and indicating which fields have not been filled\validated.

### Actual result

Same form, with no validation feedback.

### Additional context

I believe it is strongly connected to caching, because once I Reload Memory Cache in Umbraco backoffice, and refresh the page, it highlights the fields that have not been filled.

There is a link, which might be related:

https://our.umbraco.com/forum/umbraco-8/92359-how-will-the-nucache-influence-model-alteration-in-the-pipeline

It states that, " I am talking about retrieving a model from cache, and adding stuff to it - that will not work anymore."; which might be the case in this situation.

Thanks!

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #5984, #5987, #6141

### Description
Macro results are being cached as part of RTE values even if the macro has all caching disabled.

`RteMacroRenderingValueConverter` sets its cache level to Snapshot, but that isn't actually doing what we want it to. That cache level is used when caching the ObjectValue and XPathValue, but the macro rendering is happening in `ConvertSourceToIntermediate`. The intermediate value is effectively always cached at the Element level.

This PR moves the processing into `ConvertIntermediateToObject` and `ConvertIntermediateToXPath`.

Testing:
- Create a Partial Macro with an @DateTime.Now.Ticks output
- Create a Macro which uses the Partial Macro and Use in rich text editor and the grid enabled, Cache Period 0, Cache by Page Disable, Cache personalized disabled.
- Create a Page with an TinyMCE Editor Field and use the Macro inside the Editor.
- Publish and show the page. Refresh and check that the time updates.

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/ValueConverters/RteMacroRenderingValueConverter.cs
+++ b/src/Umbraco.Web/PropertyEditors/ValueConverters/RteMacroRenderingValueConverter.cs
@@ -11,6 +11,7 @@ using Umbraco.Core.Cache;
 using Umbraco.Core.Services;
 using Umbraco.Web.Composing;
 using Umbraco.Web.Macros;
+using System.Web;
 
 namespace Umbraco.Web.PropertyEditors.ValueConverters
 {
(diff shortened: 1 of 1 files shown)


emmaburstow:
Fantastic work here @stevemegson - thanks for the PR and the description with testing steps. It makes our job much easier :)

We'll take a look and let you know if we need anything from you.

Emma

stevemegson:
In response to #6141, I've changed this only render macros in ObjectValue, not in XPathValue. This matches the behaviour of v7.

It seems unlikely that many people will be using xpath queries which depend on RTE values, so even if it didn't cause problems we'd just be wasting time rendering macros in order to throw the values away.

nul800sebastiaan:
Thanks very much @stevemegson - the change makes a lot of sense! Good detective work there!

PR #6012, opened 2019-07-29
V8: Clear the content type "common" cache when deleting templates

kjac:
### Prerequisites

_Original source for this issue here: https://github.com/umbraco/Umbraco-CMS/issues/6006#issuecomment-515677337_ 

### Description

When you delete a template that's assigned to a content type, the content type remains in the cache with the template assigned:

delete-template-reload-content-type-cache-before

Turns out the template cache refresher doesn't clear the "common" cache for content types, only the isolated content type cache. This PR amends that, and thus deleting a template also forces a content type cache refresh:

delete-template-reload-content-type-cache-after

#### Steps to test

1. Create a content type with a template assigned.
2. Delete the template.
3. Verify that the content type does not list the (now deleted) template in its "Templates" section.

### DIFF
--- a/src/Umbraco.Web/Cache/TemplateCacheRefresher.cs
+++ b/src/Umbraco.Web/Cache/TemplateCacheRefresher.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using Umbraco.Core.Cache;
 using Umbraco.Core.Models;
+using Umbraco.Core.Persistence.Repositories;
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hullo @kjac! This is an excellent fix! You're so very good at spotting these issues. We'll shout if we need you.

Emma

nul800sebastiaan:
Nice one, thanks @kjac !

Issue #5906, opened 2019-07-16
DefaultMediaUrlProvider throws error when using FileSystemProvider.Azure


deMD:
In an Umbraco 8.0.2 Website we use the FileSystemProvider.Azure package to load media from azure storage, but since upgrading to Umbraco 8.1 all old media files stopped working. Trying to resolve image.url produces the following exception in our api:

![image](https://user-images.githubusercontent.com/22070292/61276210-dd929f80-a7af-11e9-9dce-1091fac19af6.png)

This happens in our mapper function:
```
public IDtoBase Map(IPublishedElement publishedElement, string culture = null)
        {
            if (publishedElement is Image image)
            {
                return new ImageDto
                {
                    Type = Type,
                    Id = image.Key,
                    Path = image.Url(culture, UrlMode.Absolute),
                    WidthInPixels = image.Value(prop => prop.UmbracoWidth, culture),
                    HeightInPixels = image.Value(prop => prop.UmbracoHeight, culture)
                };
            }
            return null;
        }
```

When using the property variant of Image.Url we get the following message:
![image](https://user-images.githubusercontent.com/22070292/61276517-7d502d80-a7b0-11e9-9d84-27348b804676.png)

Reproduction
------------

- Create an Umbraco 8.0.2 website
- Add FileSystemProvider.Azure
- Upload image(s)
- Upgrade to Umbraco 8.1
- Image.Url() won't work anymore

rpeltola:
Seems to be related to these changes:

- #4942 (Create media Url provider)
- #5282 (Adds media url provider)

Quick fix is to implement similar override like CdnMediaUrlProvider example on the PR #5282 to always provide correct full URL for the images.

nul800sebastiaan:
I've been talking to @Jeavon and he's been doing some tests on this but he can't reproduce this scenario. @rasmusjp suggests it could be that you've got full URLs stored in the database for those previously created media items. Can you check if that's the case?

deMD:
That certainly is the case Sebastiaan, we have not been using a virtual path provider for media but instead serve them straight from the storage account due to the headless nature of our project.

rasmusjp:
If the stored URL is already absolute we should just return that value

PR: https://github.com/umbraco/Umbraco-CMS/pull/6016

nul800sebastiaan:
PR #6016 should fix that, I've cherrypicked it into the 8.1 branch for release with 8.1.2.

PR #6016, opened 2019-07-29
Fix exception when stored media url is already absolute

rasmusjp:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/5906


nul800sebastiaan:
👍 thanks!

PR #6019, opened 2019-07-30
v8 Changing a property type from invariant to variant and back again causes issues with reporting that a document is edited or not

Shazwazza:
v8 Changing a property type from invariant to variant and back again causes issues with reporting that a document is edited or not.

When the problem occurs, a document will always say there is pending changes, but there are not pending changes.

The problem is due to no re-normalizing the `edited` column values in the tables: `umbracoDocument` and `umbracoDocumentCultureVariation`. I've written unit tests for 2 examples, but there are other cases too which are now covered by the existing unit tests with further assertions.

### Test 1

_This test is written and is now passing_

* 2 langs - `en` (default) + `fr`, 1 variant content type, 1 variant property
* Create + Publish with all values filled in = zero cultures edited, document not edited
* Update value on property for both langs and Save = 2 cultures edited, document is edited
* Change property to invariant = 1 culture edited (`en` since its the default lang), document is edited (__In 8.1.1 this fails__)
* Update property value and Publish = zero cultures edited, document not edited
* Change property to variant = 1 culture edited (`fr` since it had pending changes before variance changes), document is edited (__In 8.1.1 this fails__)

### Test 2

Do the opposite of the test above (i.e. start with invariant)

### More tests

All of these tests are in the `ContentTypeServiceVariantsTests` ( i have moved many tests into this class from the `ContentTypeServiceTests` since all of the ones in `ContentTypeServiceVariantsTests` are related to changing variations). I've cleaned these tests up a little and also added a few more along with lots more assertions since those were actually failing before.

Reproduction
------------

Based on the unit test, you should be able to replicate the problem in the back office prior to this change by doing this:

* Have a couple languages installed
* Created variant doc type
* Create an invariant property type
* Create a document of this type, fill in the values and publish all cultures
* Change the invariant property type to variant
* Fill in all culture values and re-publish all cultures
* Change the invariant property back to invariant
* Re-publish all cultures 
* You will probably see the document says there are pending changes when there isn't

---
_This item has been added to our backlog [AB#2095](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2095)_


bergmania:
Hi @Shazwazza..

Can you see what i'm doing wrong when I try to reproduce in the UI?

I look for the green dot for pending changes. Do I have to look elsewhere?

6019-before

bergmania:
I finally reproduced and the changes fixes the issue. 

Code changes looks fine, and nice with the extra tests :)

Issue #5516, opened 2019-05-24
V8 - Configurable editorService width


jerpenol:
Hi,

I have a custom grid and my gridcontrols make use of infinite-editing. It works perfectly except for one thing: I can only open new overlays "full-width" or "small" and nothing in between. 

I would expect to be able to configure the width of a new overlay myself. I see multiple ways to solve this problem (neither of which are currently possible):
1. Add a 'medium' CSS class (or even more variants like 'large'), instead of only small
2. Use what the umb-editor directive is doing and allow us to set the ng-style. 
3. Allow us to inject CSS classes

The pros and cons are:
- **For solution 1**: The pros are that this is relatively easy to achieve and it creates a standardized set of sizes (large, medium and small). The con is that there is no "one size fits all", everyone expects these sizes to be different.
- **For solution 2**: The pros are that this is relatively easy aswell (only a small angular change in the directive template) and very flexible. The con is that it breaks the separation of styling and code/html.
- **For solution 3**: The pro in this case is that it gives us alot of flexibility in styling (even better so than _solution 2_).  There's not really a con here, but there are multiple ways of implementing this solution, for example: single string V.S. string array.

I would love a fix for this. My opinion is that a combination of all would be the solution to this problem, so it's suitable for every usecase. 

I'd like to get everyone's opinion on this, so we can create a PR for this.

nul800sebastiaan:
Hey @jerpenol - thanks for the feature request, I understand where it's coming from but I believe this was a conscious choice we made so as not to make things jump around too randomly. I've not personally been involved in those talks though but I expect the current behavior to stay the way it is.

However, I will have a chat with the team at HQ and see what they think and get you some more feedback on this issue. I'll update when I know more! 👍

jerpenol:
Hi @nul800sebastiaan.
Supertak! I'll await their feedback.

nul800sebastiaan:
Hi again @jerpenol - we had a little chat about this and we need some extra detail on what the problem is that this would solve for you. There's a few solutions in your feature request but we're not sure what problems they are solving?

jerpenol:
Hi @nul800sebastiaan , 

My problem is that I want to have overlays for my custom gridcontrol, but they don't need to be full-width all the time. 

[...]

PR #6028, opened 2019-07-31
V8: Use infinite editor for list view settings (5021)

### LINKED ISSUE #5021: UX: List view configuration should open as an infinite editor.
I would like the List View-configuration (in DocType-configuration) to open as an infinite editor.

Today it looks like:
image

The reason for this is to ensure a better scope for the current editing phase. Today the "Save & publish"-button is placed in a way where it invites the user to confirm their changes by clicking it. But the save-feature of the list-view-configuration is an inline button further down the page. So we need to create a focus so the user doesn't see the "Save & Publish"-button.

By implementing this as infinite editing we will ensure that the users focus is only on the configuration of the list view and we will be able to make the "save"-button stay visible in bottom right corner always.

When implemented I would like Listview configuration to be visible in the "Permissions"-tab. Which I then want renamed to "Configurations".

All question and suggestions are very welcome.

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #5021 

### Description
Changes here update the existing list view settings user experience to use an infinite editor rather than the collapsing panel. I've tidied up the directive a little, and updated the property styles in the overlay to make everything more consistent and v8-flavoured.

Gif for viewing pleasure:

list-view

Have verified everything works as before.


emmaburstow:
Oh lovely stuff, @nathanwoulfe! Excellent fix. We'll take a look at the work and shout if we need anything from you, as ever. 

Emma

nathanwoulfe:
Cheers legend! 😁

src/Umbraco.Web.UI.Client/src/common/directives/components/umblistviewsettings.directive.js
```diff
@@ -1,161 +1,143 @@
...
+            } 
 
-          });
+            /* ----------- LIST VIEW SETTINGS --------- */
+            const showSettingsOverlay = () => {
+                const overlay = {
+                    view: 'views/components/umb-list-view-settings-overlay.html',
+                    hideDescription: true,
+                    hideIcon: true,
+                    size: 'small',
```
skttl: I think it should be in the large infinite editor, that would give the settings some space to breathe.

nathanwoulfe: Hi @skttl, missed your comment - is there an inbetween size? The default (per infinite editor for a document type) might be too wide, something inbetween would be goldilocks-just-right. Can add an umb-editor--medium option easily enough

skttl: No, only large and small is supported. But maybe a PR with a medium setting would be approved? :)

nul800sebastiaan: Looks like I never replied to this one : https://github.com/umbraco/Umbraco-CMS/issues/5516 :-o

But yes, we'd be good with adding a medium!

nathanwoulfe: One medium coming up. Will keep it separate from this PR though.

If I update my local branch from upstream/v8/dev, will I end up with a heap of additional git history/commits? This branch was 8.1, so it's stale.

nul800sebastiaan: You will, but they'll be in your local v8/dev branch. Shouldn't be a problem! And you could even merge v8/dev into this PR branch as well.

nul800sebastiaan: I've accepted the `medium` PR and updated this PR. Also merged with v8/dev which went well.

nul800sebastiaan:
Sorry for the delay @nathanwoulfe - I am loving the update and the implementation is a joy to work with. No more "double save button" that I (and everybody else) keep forgetting to press! 🎉

Thanks again!! ⭐️

nielslyngsoe:
Thanks thanks thanks! 🤩

nathanwoulfe:
@nul800sebastiaan @nielslyngsoe my pleasure!

PR #6029, opened 2019-07-31
WIP - Better ux under long operations

bergmania:
Working on making ux better under long running operations, by caching the lock timeout exceptions and lookup what is using the lock.

Still WIP


bergmania:
Before:
Long-running-before-ux-improvements

After:
Long-running-after-ux-improvements

src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
```diff
@@ -237,6 +241,44 @@ public override bool TryGetDefaultConstraint(IDatabase db, string tableName, str
             return !constraintName.IsNullOrWhiteSpace();
         }
 
+        public override void WriteLock(IDatabase db, int? writeLockReasonId, params int[] lockIds)
+        {
+            // soon as we get Database, a transaction is started
+
+            if (db.Transaction.IsolationLevel < IsolationLevel.ReadCommitted)
+                throw new InvalidOperationException("A transaction with minimum RepeatableRead isolation level is required.");
```
Shazwazza: Shouldn't this say ReadCommitted since that is what it is checking against?

src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
```diff
@@ -237,6 +241,44 @@ public override bool TryGetDefaultConstraint(IDatabase db, string tableName, str
...
+        {
+            return exception is SqlException sqlException && sqlException.Number == 1222;
+        }
+
+        public override void ReadLock(IDatabase db, params int[] lockIds)
+        {
+            // soon as we get Database, a transaction is started
+
+            if (db.Transaction.IsolationLevel < IsolationLevel.ReadCommitted)
+                throw new InvalidOperationException("A transaction with minimum RepeatableRead isolation level is required.");
```
Shazwazza: Shouldn't this say ReadCommitted since that is what it is checking against?

src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
```diff
@@ -173,6 +174,9 @@ int GetInt32(IDataReader reader, int ordinal, int defaultValue)
             return items.Select(x => new Tuple<string, string, string, string>(x.TableName, x.ColumnName, x.Name, x.Definition));
         }
 
+        public override bool IsReadUncommittedSupported => true;
+        public override System.Data.IsolationLevel DefaultIsolationLevel => System.Data.IsolationLevel.RepeatableRead;
```
Shazwazza: Shouldn't this be ReadCommitted?

src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
```diff
@@ -237,6 +241,44 @@ public override bool TryGetDefaultConstraint(IDatabase db, string tableName, str
...
+                throw new InvalidOperationException("A transaction with minimum RepeatableRead isolation level is required.");
+
+
+            // *not* using a unique 'WHERE IN' query here because the *order* of lockIds is important to avoid deadlocks
+            foreach (var lockId in lockIds)
+            {
+                db.Execute(@"SET LOCK_TIMEOUT 1800;");
+                var i = db.Execute(@"UPDATE umbracoLock WITH (REPEATABLEREAD) SET value = value*-1, writeLockReasonId = @writeLockReasonId WHERE id=@id", new { id = lockId, writeLockReasonId = writeLockReasonId });
+                if (i == 0) // ensure we are actually locking!
+                    throw new Exception($"LockObject with id={lockId} does not exist.");
```
Shazwazza: better if we use a non base class `Exception`

src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
```diff
@@ -237,6 +241,44 @@ public override bool TryGetDefaultConstraint(IDatabase db, string tableName, str
...
+            // soon as we get Database, a transaction is started
+
+            if (db.Transaction.IsolationLevel < IsolationLevel.ReadCommitted)
+                throw new InvalidOperationException("A transaction with minimum RepeatableRead isolation level is required.");
+
+
+            // *not* using a unique 'WHERE IN' query here because the *order* of lockIds is important to avoid deadlocks
+            foreach (var lockId in lockIds)
+            {
+                db.Execute(@"SET LOCK_TIMEOUT 1800;");
```
Shazwazza: Need constants here

src/Umbraco.Core/Scoping/Scope.cs
```diff
@@ -235,6 +237,7 @@ public IUmbracoDatabase Database
                 try
                 {
                     _database.BeginTransaction(IsolationLevel);
+                    _database.Execute("SET LOCK_TIMEOUT 1800;");
```
Shazwazza: Perhaps should be part of SqlSyntax

src/Umbraco.Core/Services/Implement/EntityService.cs
```diff
@@ -257,20 +258,38 @@ public virtual IEntitySlim GetParent(int id, UmbracoObjectTypes objectType)
...
+            using (var scope = ScopeProvider.CreateScope(autoComplete: true))
             {
-                var query = Query<IUmbracoEntity>().Where(x => x.ParentId == parentId);
-                return _entityRepository.GetByQuery(query);
+                try
+                {
+                    var query = Query<IUmbracoEntity>().Where(x => x.ParentId == parentId);
+                    return _entityRepository.GetByQuery(query);
+                }
+                catch (LockTimeoutException e)
```
Shazwazza: I guess this means that we are effectively going to have to wrap every call to any repository query with this?

src/Umbraco.Web.UI.Client/src/common/directives/components/tree/umbtree.directive.js
```diff
@@ -243,33 +245,52 @@ function umbTreeDirective($q, $rootScope, treeService, notificationsService, use
...
+                        $scope.activeTree = $scope.tree.root;
+
+                        emitEvent("treeLoaded", { tree: $scope.tree });
+                        emitEvent("treeNodeExpanded", { tree: $scope.tree, node: $scope.tree.root, children: $scope.tree.root.children });
+
+                        $scope.isLoading = false;
+                        $scope.isLoadingReason = "";
+                        return $q.when(data);
+                    }, function (reason) {
+                        if(reason.status === 409){
```
Shazwazza: We can put this type of logic into angular interceptors, else we're going to have to duplicate this every single place in the application. We'll need an interceptor and then some custom service to display these messages. Will prob have to be some sort of new UX thing.

src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
bergmania: For the "1800" or what do you want as constant?

src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
bergmania: Sure, that part was not changed, but we can clearly improve..

src/Umbraco.Core/Services/Implement/EntityService.cs
bergmania: If we want a custom reason then yes, otherwize if falls back to the default

src/Umbraco.Web.UI.Client/src/common/directives/components/tree/umbtree.directive.js
bergmania: Cool.. Sounds like a good solution

src/Umbraco.Core/Scoping/Scope.cs
Shazwazza: I 'think', this doesn't actually need to be here since the lock timeout call is made when actually creating a write lock when in sql server mode.

Issue #5669, opened 2019-06-16
Infinite editing editorService.contentEditor unable to create new content item


pgregorynz:
I am currently working on a back office extension to allow users to create new content items without leaving the context of the current item using the infinite editing features.

The documentation for the contentEditor hints that this should be possible by setting the create flag on the editor options to true.  Although the docs mention Media and not Content?

create | Boolean | Create new content item
------ | ------- | -----------------------

See: https://our.umbraco.com/apidocs/v8/ui/#/api/umbraco.services.editorService

However there is nowhere to set the alias of the type of content you would like to create, or where in the tree it should create the new item.

Reproduction
------------

If I run the following:
```
function create() {
      var options = {
        create: true,
        submit: function (model) {
          editorService.close();
        },
        close: function () {
          editorService.close();
        }
      }
      editorService.contentEditor(options);
}
```
you end up with the following error.

`GET http://localhost:2427/umbraco/backoffice/UmbracoApi/Content/GetEmpty?contentTypeAlias=undefined&parentId=1068 404 (Not Found)`

1068 is the ID of the page I am currently on.

### Bug summary

There should be a way to create a new piece of content using the editorService.contentEditor I would expect the editor object to have attributes for:

* documentTypeAlias
* parentId

### Specifics

* Umbraco 8.0.2

### Steps to reproduce

As mentioned above

### Expected result

To be able to initiate the creation of new content from the editorService.contentEditor with a call like the following

```
function create() {
      var options = {
        create: true,
        parentId: 1069,
        documentTypeAlias: 'contentPage',
        submit: function (model) {
          editorService.close();
        },
        close: function () {
          editorService.close();
        }
      }
      editorService.contentEditor(options);
}
```

### Actual result

you can't.  And when you set the create flag I get the following error:

GET http://localhost:2427/umbraco/backoffice/UmbracoApi/Content/GetEmpty?contentTypeAlias=undefined&parentId=1068 404 (Not Found)

nul800sebastiaan:
Seems like an oversight that we don't use those parameters. I'll discuss with @nielslyngsoe and see if this actually supposed to work.

As a bit of background, can you tell us a bit more about the use case? Why do you need to create content items from a content item in infinite editing mode? Seems a bit dangerous as well, you'd then be able to create content inside the infinite editor and then in that one and.. well infinitely ;-)

pgregorynz:
[...]

PR #6031, opened 2019-07-31
V8: Implementation of Create method for infinite editing edtiorService.contentEditor

pgregorynz:
This PR is to fix the following issue.

https://github.com/umbraco/Umbraco-CMS/issues/5669

### Description

The editorService.contentEditor has a documented create flag that has no implementation. As mentioned in the linked issue, we assumed that the create flag would allow for a property editor to open a new infinite editor in a create state, however there were a number of other requirements missing.

This pull request adds the missing options that would be required to initiate the creation of a piece of content which are are `parentId`, `documentTypeAlias`, `allowSaveAndClose`, `allowPublishAndClose`

The call to open an infinite editor in creation mode looks like this:

```
      var options = {
        create: true,
        parentId: 1069,
        contentTypeAlias: 'contentPage',
        allowSaveAndClose: true,
        allowPublishAndClose: true,
        submit: function (model) {
          editorService.close();
        },
        close: function () {
          editorService.close();
        }
      }
      editorService.contentEditor(options);
```

- Adds a new scaffolding method for creating a new empty content ite, under infinite editing.

As an example implementation of this, we have created an editor that allows for the creation and linking of a new piece of content from a page without leaving the page it is needed on.  This is useful in situations that require the editor to maintain editing focus and flow which is why infinite editing was created in the first place.

image


emmaburstow:
Lovely fix here @pgregorynz - this fix is going to really improve the ux of creating new content items. Thanks also for the screenshot and the description. Enough info for us to get stuck in to testing functionality. 

We'll take a look at the code too and if we have any questions we will, of course, shout.

Emma

hartvig:
Woohooo - reviewing in progress #h5yr

hartvig:
Works perfectly - great functionality!

Only issue is the sample provided where it should be documentTypeAlias instead of contentTypeAlias.

There's a lot of options by default for saving. We should consider some better defaults and then make options for specifying different buttons.
Screenshot_2

We could consider only having
Save | Publish and Close | Close

And then you could specify "Save and Publish" and "Save and Close" as extra buttons.

hartvig:
Doh - can even see there's already options for buttons in config 🤦‍♂

pgregorynz:
The buttons are just ones that are already in the core, just they were not wired up fully.  I do tend to agree that the number of buttons or the layout of the buttons (which are configurable) could be improved.  The main reason we left the Save and Save and Publish alone is that it allows you to work with the content, save and save and publish without the infinite pane closing.

cpevans:
Would love to see this released - very keen to do a real-world test of @pgregorynz 's new block editor that requires this fix to work properly.

nul800sebastiaan:
@pgregorynz The buttons are a bit confusing though, but maybe that's because I'm not used to them. It's strange to have both "Save & Publish" and "Publish and Close". 

After reading this I realized - aha: you make a new content item and keep editing ("Save & Publish") or you make a new one and the editor closes so you're back to your previous item.

I guess this makes sense and it even makes sense leaving both buttons there so the editor has a choice. 

Tested using [@hartvig's code](https://github.com/umbraco/Umbraco-CMS/files/3495490/InfiniteCreator.zip) and it seems to work well. Great stuff and a cool new feature coming in 8.2.0.

A few tips for future contributions:

You made your changes straight in your clone of the `dev/v8` branch. This means that you can only have 1 PR open. After cloning the source, always make sure you create a new branch from there so you could switch back to `dev/v8`, update from the upstream (the Umbraco repository), merge changes into your branch, etc. Additionally, since I'm git-squashing this PR, you will not be able to update `dev/v8` from the main repo any more, because you're out of sync (easiest way to resolve now: make sure to delete your clone locally and your fork on github and make a new fork & clone if you intend to work on additional contributions).

There was some commented out code left in one of the files, no need to leave that in there, just delete.

Other than that, really cool work! Took me a while to wrap my head around what this was doing but now we can go To Infinity And Beyond. 😉

Thanks very much! 🏅

pgregorynz:
YAHOOO!!  Thanks for merging this!  We are looking forward to the 8.2 release :)  For now we will just keep our version of Umbraco up to date 💃

src/umbraco.sln
```diff
@@ -123,15 +123,13 @@ Global
 		{31785BC3-256C-4613-B2F5-A1B0BDDED8C1}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{31785BC3-256C-4613-B2F5-A1B0BDDED8C1}.Release|Any CPU.Build.0 = Release|Any CPU
 		{5D3B8245-ADA6-453F-A008-50ED04BFE770}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{5D3B8245-ADA6-453F-A008-50ED04BFE770}.Debug|Any CPU.Build.0 = Debug|Any CPU
```
Shazwazza: It's not great these changes were committed... I will put this back in my current branch for means for all devs currently the unit test and benchmarks projects won't build which means people won't be getting build errors for those projects :/

hartvig: Duh! What changes had that affect?

Shazwazza: It's just these changes in this .sln file, these 2x projects were excluded from the Debug Build (normally done using the 'Configuration Manager' in VS). Was prob done for some reason but the change was overlooked with this merge.

Issue #6035, opened 2019-07-31
Multiple Media Picker property editor does not have sortable items


whatwasmyusername:
I'm running Umbraco 8.1.1 and properties with the grid editor alias of Umbraco.MediaPicker are no longer sortable.

Reproduction
------------

### Bug summary

Sorting of the media in a media picker property are not sortable even when the media picker is setup to allow multiple items.

### Specifics
* Running on Umbraco 8.1.1
* $scope.sortableOptions is initialized before the setupViewModel() call where the property $scope.isMultiPicker is initialized, so $scope.sortableOptions.disabled is always **true**.
* Source file: https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.controller.js

### Steps to reproduce

* Create a media picker property with the "Pick multiple items" enabled
* Attempt to edit a content item with that field
* Images are not sortable

### Expected result
Media should be able to be dragged and dropped into desired order.

### Actual result
Media items are not sortable.

kjac:
I can reproduce this. The sorting is also a bit buggy when it's forced enabled, so I'll have a look at both.

kjac:
PR in #6037

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6037

Cherry picked for 8.1.5 089e450e09bd5d9e40ead8af2edc88658671e534

PR #6037, opened 2019-08-01
V8: Make sure media picker items are sortable

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6035

### Description

Items picked in a media picker with "Pick multiple items" enabled should be sortable, but they are not. See #6035 for more details.

This PR fixes the sortable implementation of the media picker. When it is applied, the picked items can be sorted:

media-picker-sortable-after

Note that I have removed the `$scope.$watch` as part of this PR as it didn't seem to have any positive effect. I removed it because it caused a reload of all images when sorting the media picker items, thus executing a bunch of unnecessary requests.

#### Testing this PR

1. Verify that media picker items _cannot_ be sorted when "Pick multiple items" is _disabled_ on the media picker config.
2. Verify that media picker items _can_ be sorted when "Pick multiple items" is _enabled_ on the media picker config.
3. Verify that the updated item sort order is persisted when saving the content.


emmaburstow:
Oh lovely. Yes, they should be sortable. Thanks for seeing to it that they are. We'll let you know if we need you.

Em

kjac:
Merge conflicts resolved

nul800sebastiaan:
Hmmm,I think some recent merges might have caused it to do some weird stuff!



I've also observed picked images going missing but it's now impossible to reproduce, so not sure what that is all about 🙈

kjac:
@nul800sebastiaan I can't reproduce your errors 😲 Works on My Machine (tm)

nul800sebastiaan:
Hmmmm.. well, my browser likes this change today, so I'm happy! Must've done something weird last time. 😁 

THANKS!

ceee:
I thought this PR was cherry picked for 8.1.5 release (and should also be there in 8.2)? 
I can't get it to work after updating to 8.2, anyone can confirm?

kjac:
@ceee while I haven't tested this in 8.2 it seems to work just fine in vLatest:

media-picker-sortable

PR #6038, opened 2019-08-01
Fixes how examine starts up to make it easier for custom index developers

Shazwazza:
There's an issue currently for developers creating custom lucene based indexes that most times requires that the index be manually unlocked on startup in their custom IComponent. This was never the intention and the reason is because Umbraco unlocks all of the indexes at startup but unfortunately this was happening before the developers component was executing.

This solves a couple of problems:

* Ensures the indexes are unlocked and the rebuild logic for empty indexes is queued AFTER all user components
* Creates a base class for custom index developers `ExamineUserComponent` so that developers can inherit from this instead of just `IComponent` and this will also ensure that no initialization is done if MainDom was not acquired. This is very important for Lucene and by inheriting from this means that devs don't need to worry about this check

### DIFF
--- a/src/Umbraco.Web/Search/ExamineFinalComposer.cs
+++ b/src/Umbraco.Web/Search/ExamineFinalComposer.cs
@@ -1,0 +1,11 @@
+﻿using Umbraco.Core.Composing;
+
+namespace Umbraco.Web.Search
+{
+    // examine's final composer composes after all user composers
+    // and *also* after ICoreComposer (in case IUserComposer is disabled)
+    [ComposeAfter(typeof(IUserComposer))]
+    [ComposeAfter(typeof(ICoreComposer))]
+    public class ExamineFinalComposer : ComponentComposer<ExamineFinalComponent>
+    { }
+}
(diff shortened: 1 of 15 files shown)


bergmania:
Code looks fine, because we can't make breaking changes..

Test:
- Can't start a new empty project 
image

bergmania:
Seems like a series of minor issues when we start a new project

- The indexes is null (The error above)
- ExamineFinalComponent don't initialize logger, .. (missing constructor)
...

src/Umbraco.Web/Search/ExamineFinalComposer.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Umbraco.Core.Composing;
+
+namespace Umbraco.Web.Search
+{
+    // examine's final composer composes after all user composers
+    // and *also* after ICoreComposer (in case IUserComposer is disabled)
+    [ComposeAfter(typeof(IUserComposer))]
+    [ComposeAfter(typeof(ICoreComposer))]
```
bergmania:
```suggestion
    [ComposeAfter(typeof(ICoreComposer))]
    [RuntimeLevel(MinLevel = RuntimeLevel.Run)]
```

bergmania:
Tested and works (after I added min run-time level to `ExamineFinalComposer`).

Issue #5993, opened 2019-07-24
8.1: "Allow varying by culture" value not included when creating package


limpa123:
Document types with "Allow varying by culture" set to true does not keep this value after installed from package. Might be the same problem with document types properties

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6046

PR #6046, opened 2019-08-02
V8: Import and export of document type and property variations setting in packages (5993)

AndyButland:
This PR writes and reads XML elements for the "allow varying by culture" field into the package definition, so this information can be saved and restored by the creation and installation of packages.

Reported in issue #5993.


emmaburstow:
Hey @AndyButland 

This is a great fix. Thank you! We'll take a look and let you know if we need anything further from you. 

Emma

nul800sebastiaan:
This is lovely @AndyButland - thanks very much! Good to see IsElementType now works too! 👍

PR #6049, opened 2019-08-02
v8: Fix for #6047 to Ensure a media type correctly inherits it's selected parent

### LINKED ISSUE #6047: Creating a new MediaType beneath a parent, doesn't set the parent as a composition
When creating a new MediaType beneath an existing MediaType, the composition is not set correctly, and any properties from the parent MediaType are not inherited by the child MediaType.

**Actual**
image

Confirmed in latest v7/dev and v8/dev branches

**Expected**
image

PRs incoming :)

tompipe:
When creating a new MediaType beneath an existing MediaType, the composition is not set correctly, and any properties from the parent MediaType are not inherited by the child MediaType.

This PR fixes #6047

### DIFF
--- a/src/Umbraco.Web/Editors/MediaTypeController.cs
+++ b/src/Umbraco.Web/Editors/MediaTypeController.cs
@@ -19,6 +19,7 @@ using Umbraco.Core.Dictionary;
 using Umbraco.Core.Logging;
 using Umbraco.Core.Persistence;
 using Umbraco.Web.Composing;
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hiya Tom! 

Hope you're well :) 

Thanks for the fix for media composition inheritance here. We'll take a look and let you know if we need anything. 

Emma

nul800sebastiaan:
Thanks again @tompipe ! 👐

Issue #5484, opened 2019-05-17
V8 : Back office  freeze when opening multiple content nodes


stherm:
When opening multiple node in a tree view in the backOffice At some point the back office stop. 

Reproduction
I have Contents that are kind of a tree, so multiples root,multiple childs and the child have multiple childs and so on. 

If I start to open all childs, at some point the back office will just stop rendering and freeze. The only way to resolve it is by refreshing the page. 

I've teste with SQLCE and SQL Server , happen with both. 

I get the bug between 26 and 40 nodes open.

![image](https://user-images.githubusercontent.com/1386693/57952373-fc151100-78ba-11e9-9b11-21886717a433.png)

stherm:
For more information, here is the javascript profiler 
![image](https://user-images.githubusercontent.com/1386693/58100130-08df8080-7bab-11e9-80d2-83b994d9000a.png)

I've check and the problem doesn't weems to be DB related

nul800sebastiaan:
@stherm I am not sure what you mean with the nodes being "open"? Do you mean that when you expand 26+ tree items the backoffice stops responding in your browser?

stherm:
Basically yes. I'm just expanding the tree and at some points, the browser freeze. It's not always at the same number of node/document. but I eventually get it to completely freeze in under 50 nodes

stherm:
I have some more details on the bug : 
There's some tests I've done 

All node where name 1,2,3,4, 1.1, 1.2 , 1.3 for test purpose

Not linked with tree depth - 27 levels ok (Although I can't create a 28th node due to sql error)

Test 2

Banch 1->27 node | Branch 2->27 node (pretty slow to render)

Banch 1->27 nodes | Branch 2->27 nodes  | branch 3-> 27 node (really slow to render)

Branche 4-> 27 node  | Branch 2 (flat structure) -> 15 nodes | branch 3 -> 4 nodes Branche 5 -> 1/27 node (FAIL and Freeze)

emmaburstow:
Hi everyone,

I can't recreate this on my latest install. I've created 5 parent nodes, each with 28 child nodes successfully, all open at the same time without freezing. 

![issue5484](https://user-images.githubusercontent.com/37150989/58865806-22a2bc80-86af-11e9-850e-352b9253579c.gif)

Emma

stherm:
Hi I've take a screen shot of the two bug I've descibe before. The first one is an SQL error related to the length of the url. That one I get. 
[...]

PR #6051, opened 2019-08-02
Fix for #5484 - V8 : Back office freeze when opening multiple content nodes

bdsoko:
This PR fixes the issue where the expandedPaths array grows out of control (adding duplicate paths)   [#5484 ](https://github.com/umbraco/Umbraco-CMS/issues/5484)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/tree.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/tree.service.js
@@ -69,12 +69,14 @@ function treeService($q, treeResource, iconHelper, notificationsService, eventsS
                 if (childPath.startsWith(p + ",")) {
                     //this means that the node's path supercedes this path stored so we can remove the current 'p' and replace it with node.path
                     expandedPaths.splice(expandedPaths.indexOf(p), 1); //remove it
-                    expandedPaths.push(childPath); //replace it
+                    if (expandedPaths.indexOf(childPath) === -1) {
+                        expandedPaths.push(childPath); //replace it
+                    }
                 }
                 else if (p.startsWith(childPath + ",")) {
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hey @bdsoko 

Thanks for the work here an dcongrats on your first PR to this repo! We'll take a look at the work and if it's all good, we'll merge. When we do, you'll get a contributor badge on your Our.Umbraco.Com profile :)

We'll be in touch if we have any questions.

Emma, [Pull Request team](https://umbraco.com/blog/introducing-the-new-pr-team/)

stherm:
Hi, 
   will this fix be available in a upcoming release?

danimalx:
When will this fix be implemented please? It's very frustrating for our content editors that the tree freezes every 10 minutes or so when making multiple edits.

bdsoko:
@emmaburstow - Just following up on this pull request as I submitted it close to 2 months ago.  We've been using this fix for awhile now and it continues to work with the latest 8.1.5 release without any visible side effects.  It would be nice if this can make it into the next release.  Any feedback is greatly appreciated.

Shazwazza:
I also still want to look into what the heck `_trackExpandedPaths` is even for and if its actually necessary for whatever it's doing

Shazwazza:
Right, so this is used for language switching predominently. This is for both the main tree and tree pickers so when you have expanded paths to a certain depth and then change language, it re-syncs to that depth so you don't have to do it again.

This fix is a good start, however, it does mean there is a persistent 'cache' - or leak - where the main tree is always tracking all open paths. This isn't great since this will occur for ALL trees it seems when this feature is only used for the content tree since that is the only one with languages. 

I'll merge this in and thanks a bunch for you patience. I'll look at updating this logic to see if we can make this a little better for not tracking things unecessarily.

Issue #6000, opened 2019-07-25
8.1: GetPagedCurrentUserLog returns NodeId which is just the database ID for the Log table not related to any nodes.


jmayntzhusen:
Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

### Bug summary
This issue comes from using the Angular `logResource.getPagedUserLog` which returns a `nodeId`.

However for the same node that value seemed to auto increment, I then found that the logResource calls the `GetPagedCurrentUserLog` method which maps the result to this `nodeId`. 

It gets it from the database as just dto.id though, which is not really useable in the backoffice for nodes.

### Specifics
Testing on v8.1, found this based on a v7 article, which leads me to believe this has worked differently in the past, but I haven't confirmed.

### Steps to reproduce

Make a call to `GetPagedCurrentUserLog` - check the returned `nodeId`, it will correspond to ids for entries in the [umbracoLog] table, not nodes.

### Expected result
Expected a nodeId that would correspond to some kind of content or media node.

### Actual result
Got the id of the row in the audit table from the database.

marcemarc:
@jmayntzhusen we had a similar fix applied in V7 https://github.com/umbraco/Umbraco-CMS/pull/3759

jmayntzhusen:
Thanks @marcemarc 

This made me realize it is not the Audit table rather the Log table it is getting the info from 😁

Then I guess the same fix should do it for v8 as well!

nul800sebastiaan:
Fixed in: https://github.com/umbraco/Umbraco-CMS/pull/6052

Cherry picked for 8.1.4 cec4376d199e617546292d726829c0a456c75655

PR #3759, opened 2018-11-26
Fix for #3758 - Map NodeId of Audited Item instead of Log entry PK

marcemarc:
The Entity NodeIds aren't being passed through from the database Audit table... the Log Entry Primary Key is used incorrectly instead, this commit fixes it so the NodeId is passed!

### Prerequisites

- [ X    ] (X placed to annoy @nul800sebastiaan) I have added steps to test this contribution in the description below

Full explaination is on the issue, here: https://github.com/umbraco/Umbraco-CMS/issues/3758

### Description

This PR, fixes the issue described in full here: https://github.com/umbraco/Umbraco-CMS/issues/3758

Essentially we just pass the nodeId of the entity instead of passing the audit tables primary key, in the AuditRepository's GetPagedResultsByQuery method

How to test:

In an angularJS controller, perhaps on a dashboard, make a request for the user log

var userLogoptions = {
    pageSize: 10,
    pageNumber: 1,
    orderDirection: "Descending",
    sinceDate: new Date(2018, 0, 1)
};
logResource.getPagedUserLog(userLogoptions)
.then(function (response) {
console.log(response);
}

the logged items should contain NodeIds of actual nodes that the entry relates to

PR #6052, opened 2019-08-02
V8 - Update GetPagedResultsByQuery to return id of the node the audit item…

marcemarc:
… relates to instead of table PK

This was fixed previously in v7 - see https://github.com/umbraco/Umbraco-CMS/pull/3759

Issue raised here: https://github.com/umbraco/Umbraco-CMS/issues/6000

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/AuditRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/AuditRepository.cs
@@ -174,7 +174,7 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
             totalRecords = page.TotalItems;
 
             var items = page.Items.Select(
-                dto => new AuditItem(dto.Id, Enum<AuditType>.ParseOrNull(dto.Header) ?? AuditType.Custom, dto.UserId ?? Constants.Security.UnknownUserId, dto.EntityType, dto.Comment, dto.Parameters)).ToList();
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hey @marcemarc 

Thanks for the pr. Are you working rather than watching Stranger Things too? Anyways, plenty to work with here in the issue report. We'll shout if we need you. Or perhaps we'll message as the sound probably wouldn't carry.

Emma

marcemarc:
@emmaburstow I hear ya...

it's something we fixed in V7 that must have got lost again along the road to V8 - the method is called on the 'info' tab to list out audit history, but you are already on the node at that point, and so it matters not that the correct nodeid isn't returned in each audit record... but it does matter in the dashboard tutorial, where we advise how to list what the logged in user has done recently using the audit trail, and here we need the nodeid of the umbraco item the audit record relates to in order to make the handy link back to the edited item...

I'm not sure about Stranger Things, I like the font... and the homage to the late seventies early eighties era of my youth - and I watched the first series, it is good, but I can't help feeling like I've somehow already seen it?, I know that might seem like sacrilege to some, but haven't been compelled to watch any more.

emmaburstow:
Yes, this all makes plenty of sense. I'll have a try and see how I get on. The code all looks fab so just now to check I can get it to do what it says it'll do. 

I felt much the same way about the movie Super 8. Atmospheric, sure, but in terms of content, nothing much to see. Many would disagree but I can live with that. 

Em

nul800sebastiaan:
Nice one @marcemarc - shame it got lost, but hey, another PR in the pocket for you, thanks for the help! 👍

marcemarc:
@nul800sebastiaan wouldn't be an Umbraco 'proper release' if we didn't somehow break the dashboard tutorial... :-P

PR #6059, opened 2019-08-06
V8: Display current log level in log viewer sidebar

### LINKED ISSUE #6055: The v8 Log Viewer should display the current log level
As far as I can tell there is no way the check the current log level without looking at the serilog.config file.

I would be handy if the current log level was shown on the home screen of the log viewer.

applies to v8.1.1

I think it would also be useful if the log level was added to the Health Check so that a minimum level of Warning is recommended for live sites (I will create a separate issue for this, but mention it here to save someone else adding this.)

t

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #6055 

### Description
Changes here display the current log level in the sidebar of the log viewer dashboard. Will look like this:

log-level

What's changed?

- updated LogViewerSourceBase.cs to add a method for fetching the current Serilog level.
- added endpoint to LogViewerController.cs to call into that new method
- updated logViewerResource.js to add a client-side function. Also tidied up to remove a heap of duplication (refactors call to umbRequestHelper into a separate function)
- updated overview.controller.js to request the log level, and uses the returned value to set the display color for the level to match the pie chart colors
- updates UI to display the new bit - I've added a variation for the `umb-box` directive to display these inline, by wrapping a set of boxes in an element with class `.umb-box-row`.

Verified by changing the log level in serilog.config, UI updates accordingly.

### DIFF
--- a/src/Umbraco.Core/Logging/Viewer/LogViewerSourceBase.cs
+++ b/src/Umbraco.Core/Logging/Viewer/LogViewerSourceBase.cs
@@ -1,26 +1,35 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Xml;
 using Newtonsoft.Json;
 using Serilog.Events;
 using Umbraco.Core.IO;
 using Umbraco.Core.Models;
 using Umbraco.Core.Persistence.DatabaseModelDefinitions;
+using Formatting = Newtonsoft.Json.Formatting;
 
 namespace Umbraco.Core.Logging.Viewer
 {
     public abstract class LogViewerSourceBase : ILogViewer
     {
-        protected LogViewerSourceBase(string pathToSearches = "")
+        protected LogViewerSourceBase(string pathToSearches = "", string pathToSerilog = "")
         {
             if (string.IsNullOrEmpty(pathToSearches))
                 // ReSharper disable once StringLiteralTypo
                 pathToSearches = IOHelper.MapPath("~/Config/logviewer.searches.config.js");
 
             _searchesConfigPath = pathToSearches;
+
+            if (string.IsNullOrEmpty(pathToSerilog))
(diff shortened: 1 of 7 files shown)


src/Umbraco.Core/Logging/Viewer/LogViewerSourceBase.cs
```diff
@@ -1,26 +1,35 @@
...
         {
             if (string.IsNullOrEmpty(pathToSearches))
                 // ReSharper disable once StringLiteralTypo
                 pathToSearches = IOHelper.MapPath("~/Config/logviewer.searches.config.js");
 
             _searchesConfigPath = pathToSearches;
+
+            if (string.IsNullOrEmpty(pathToSerilog))
+                // ReSharper disable once StringLiteralTypo
+                pathToSerilog = IOHelper.MapPath("~/Config/serilog.config");
```
nul800sebastiaan: I think instead, you should use `Umbraco.Core.Logging.Serilog.LoggerConfigExtensions.ReadFromConfigFile()` - this also helps not changing this method signature.

I'm not sure who would even need to give us the path of the serilog config? I see it's consistent with pathToSearches - but also not entirely sure why that's configurable.

nul800sebastiaan: Ooh, @warrenbuckley suggested a possibly better way than reading and parsing config files, just ask Serilog if any of it's levels are enabled: https://stackoverflow.com/questions/49226533/how-to-get-current-serilog-minimumlevel

nathanwoulfe: That's much better - will update

nul800sebastiaan:
Lovely work @nathanwoulfe - thanks very much!

Issue #6044, opened 2019-08-02
DI Lifetime.Request works per thread.


ekovandbor:
Brief explanation of issue on forum:
https://our.umbraco.com/forum/umbraco-8//98389-iusercomposer-with-lifetimerequest-works-as-lifetimetransient

In general: 
All would have been ok with DI, if i didnt need custom EF db context, not related to umbraco content - my own db for a lot of businesslogic data structure,  for webapi and custom mvc.

So issue in general: EntityContext.Instance() gets fired 35 times during request. I did static counter, and it is seen that it was created 35 times during request. Why Lifetime.Request doesnt create 1 instance per request? (all server side use async awaited Tasks)

Apart from that Lifetime.Request dependencies gets disposed at the end of first thread, so in controller constructor i have ok context, but in controller method it is already disposed.
Also every single dependency on EntityContext has its own instance, instead 1 instance per request. So my db context and transactions are disposed before being injected. And all dependencies have different db contexts - checked by static counter with instance saved value.

Entire concept of UnitOfWork with services breaks, as i cant handle all db updates at once which is important for business logic.  

I tried to create own Container and resolve DI with DependencyResolver, but entire backoffice stop to work - i can't even login (so umbraco uses DependencyResolver), also umbraco dependent instances wasnt loaded.

Is it a known issue?  i had 8.0.2 version, and then updated to 8.1 - issue still exist.
The only specific thing is that i extended UmbracoDefaultOwinStartup, as it was needed for webapi. rest shouldn't do anything to default installation.

```
public partial class Startup : UmbracoDefaultOwinStartup
    {
        public override void Configuration(IAppBuilder app)
        {
            base.Configuration(app);

            ServicesConfig.RegisterServices(app);          
            FluentValidationModelValidatorProvider.Configure(x => x.AddImplicitRequiredValidator = true);
            AuthConfiguration.ConfigureAuth(app);
        }

        protected override void ConfigureServices(IAppBuilder app, ServiceContext services)
        {
            base.ConfigureServices(app, services);
            ServicesConfig.RegisterServices(app, services);            
        }

        protected override void ConfigureMiddleware(IAppBuilder app)
        {
            base.ConfigureMiddleware(app);            
        }        
    }
```

Would be cool to get some answer or suggestions on forum :)

[...]

PR #6069, opened 2019-08-08
Fixes DI Lifetime.Request works per thread. #6044

Shazwazza:
Fixes DI Lifetime.Request works per thread. #6044

There's a long discussion on #6044 as well as on the LightInject repo https://github.com/seesharper/LightInject/issues/494 about this but it comes down to the fact that Lifetime.Request maps to LightInjects PerRequestLifeTime which is incorrect since Lifetime.Request means "one per request" and that's not what LI does for PerRequestLifeTime - which is just transient per request and disposing at the end of the request. 

This changes the mapping so that Lifetime.Request maps to PerScopeLifetime which in LI is "one per request".

However, there's a catch. LightInject wants to have controllers registered as PerRequestLifeTime, so in the `Umbraco.Web.Composing.LightInject.LightInjectContainer` we need to check for that.

There was discussion on #6044 to obsolete `Lifetime.Scope` however i don't think we should do this anymore since every container natively supports this and there will be some cases where a user actually wants to use Lifetime.Scope over Lifetime.Request for some scenarios (i.e. custom scopes).

### DIFF
--- a/src/Umbraco.Core/Composing/Lifetime.cs
+++ b/src/Umbraco.Core/Composing/Lifetime.cs
@@ -1,4 +1,6 @@
-﻿namespace Umbraco.Core.Composing
+﻿using System;
+
+namespace Umbraco.Core.Composing
 {
     /// <summary>
     /// Specifies the lifetime of a registered instance.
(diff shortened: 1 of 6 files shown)


Shazwazza:
I will close this task since nobody seems to be affected by this and the netcore MSDI switch will means its not necessary.

Issue #6043, opened 2019-08-02
Backoffice nodetree saved in cache even after logout


liamlaverty:
A user's content node tree is saved in their browser cache. When that user signs out, the tree isn't cleared from cache. If another user with a different account signs in on the same computer,  the first user's content tree is shown to the 2nd user - even if the 2nd user doesn't have privileges to view those nodes. 

### Bug summary

It looks as if this line isn't checking if the contents of the cache belong to the user trying to access it. 

https://github.com/umbraco/Umbraco-CMS/blob/acbe82f4c891bc8ae4a7b5568bc949a54e141bde/src/Umbraco.Web.UI.Client/src/common/services/tree.service.js#L566

There's also probably a bug in the sign out code, that's not clearing away the cache.

### Specifics

* Checked in Umbraco v8.0.2
* Chrome/FF/Edge latest versions as of 2nd Aug 2019
* No errors in the console

### Steps to reproduce

1. Create a few content items in an Umbraco site (make sure to add some child nodes for best reproduction results)
2. Create two users. 
    - User 1 (Admin) has full admin privileges over the entire content tree
    - User 2 (Editor) has restricted editing privileges, and a "content start node" of one of the children
3. As the admin user, open a node that the Editor doesn't have access to
4. Sign out
5. Sign in as the editor
6. View the node tree of the admin
7. Hard refresh to view the tree of the editor
8. Sign out
9. Sign in as the admin
10. View the node tree of the editor

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

### Expected result

Expected the node tree for the editor to contain the contents they're authorised to view

### Actual result

The editor was able to view the contents of the node tree they're not authorised to view.

liamlaverty:
https://github.com/liamlaverty/Umbraco-Node-Tree-Bug

You can use this repo with users pre-setup to reproduce. The error occurs in both versions. In v8.0.2 the tree never refreshes, so you always see the previous user's tree. In v8.1.1 the same error occurs, though after about half a second the current user's tree appears.

Credentials for both sites:
**admin**
admin@example.com
AdministrationCustomPass

**editor** 
editor@example.com
EditorCustomPass

kjac:
@liamlaverty thanks for reporting 👍 

I can reproduce this. But after debugging a bit, I don't think the tree cache is the culprit here. It looks more as if the tree component isn't destroyed on logout, thus retaining its state for a split second until the tree is reloaded after login.

[...]

PR #6071, opened 2019-08-08
V8: Fix tree flickering when switching between users with different start nodes

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6043

### Description

#6043 describes how the content tree momentarily displays the nodes visible to the previous user when a new user logs in. Steps to reproduce are given in the original issue description.

Long story short: It's not a caching issue. The built-in cache layers in Umbraco are flushed correctly when a new user logs in, and this causes the tree to reload correctly. 

The problem is that the previously loaded tree is still in the DOM when the new user logs in; the `umbTree` directive keeps the last loaded tree nodes in memory, and these remain rendered until the cache is cleared and the tree is re-rendered.

I can't think of any other solution to this issue than to explicitly unload the previously loaded tree upon logout, which is what this PR is all about.

There is no significant performance downside to this; the cache remains intact if the same user logs in and out (or times out), thus the tree isn't reloaded from the server and the previously known tree state (open nodes etc) isn't lost.

Here's how things work with this PR applied:

login-flicker

#### To do or not to do?

This PR certainly introduces a bit of extra complexity to fix an issue that is likely to be a rare occurrence. One could also argue that the issue isn't that grave, since the tree *is* reloaded correctly after log-in - although it may not be as quick to reload in a real life scenario.

#### Testing this PR

1. Make sure the flickering is gone when switching between different users in the same browser window.
2. Make sure the tree state remains the same between logins with the same user.
3. Make sure that the user is redirected back to the content section when an explicit logout is performed from any other section.
4. Make sure both the tree state and the current section remains the same when the user times out (hint: set the `Umbraco.Core.TimeOutInMinutes` app setting value to 1 for testing).


poornimanayar:
Hi @kjac ,

Thank you for the PR. Can you resolve the conflict when you get a min?

Poornima

kjac:
Sure thing, I'll get on it today.

kjac:
@poornimanayar all done 👍

PR #6072, opened 2019-08-08
WIP - block editor prototype

Shazwazza:
Based on the work done at the retreat and community activity afterwords. 

This PR is here so that I can review the code more easily and will eventually be merged into the block editor working branch.


Shazwazza:
Hi, @nathanwoulfe , @callumbwhyte , @agiraud 

Sorry to spam you here but i'm trying to get this up and running and also seeing what code has changed for this prototype of the block editor.

First question - why has the `Thumbnail` been made readonly and then not using a value in the database? From what I can tell ... this is removing some old legacy code where you could have file based thumbnails for content types instead of class names like we currently have, yet it is also still enabling some legacy way of still supporting file based thumbnails?

callumbwhyte:
Ah yeah, `Thumbnail` is a tricky one. We were working on 2 things simultaneously...

1. Block editor prototype

2. Giving better previews for any editor, via the `Thumbnail` property - this would appear when picking an editor for the grid, nested content, block editor, etc

I guess the thumbnail changes should perhaps be split into a separate branch?

nathanwoulfe:
Yup, thumbnail was rabbit hole I started down, to enable convention-based thumbnails for grid components - when picking a new component to add, could show a thumbnail as an example of how the component renders. 

Can't remember what the issue was, but I started hitting a wall

Shazwazza:
Ok, i am going to revert the thumbnail changes. 

We also had prototypes in the Grid 2.0 branch for a similar thing since - which is basically a readonly 'view' of an editor, not just a thumbnail. This can then also be used in the doc type editor when showing a 'preview' of the property editor. I think this is along the same lines of what you are trying to achieve here?

callumbwhyte:
I think the intention was for it to be a static example of what the editor could look like - e.g. an image - and is CMS wide, not just for grid / blocks.

Our idea for editor "views", like in grid v2, was the default view for any block can be overridden - what you implement in that is your choice. Could be a static example, could be an inline editor.

nathanwoulfe:
Was mainly piggybacking on a ridiculously slick piece of work @JeffreyPerplex demoed at the retreat. Super slick.

I think there's still space for doing something, but definitely separate to this work

Shazwazza:
Ok, for now I'm going to revert this thumbnail thing, try to make more sense out of the code/POC and get it up and running. Would be interested in seeing this 'slick piece of work' ! :)

Shazwazza:
The first commit of these thumbnail changes actually says

> Welcome to hell. 

So... not sure it has the best intentions in this PR 😂

Shazwazza:
Hi, @nathanwoulfe , @callumbwhyte , @agiraud I've got the POC running now (haven't got too far yet) but was wondering on the off chance if anyone took some notes about known issues, remaining work, etc... ? I can see loads of todos and can see missing things but just wanted to see if any of you had any list already made. Cheers!

Shazwazza:
There's also a bunch of changes to other things in this POC like small fixes here and there but i don't fully understand what they are since there's no context around them, the changes for nested content display the 'add' button for a content type seems ok - but that should be a separate PR/fix, there's a change to the grid.html for the same type of fix but then there's also an addition of `size="editorOverlay.size"` which i'm unsure what that's fixing. 
I think we'll need to pull this POC apart and move anything that i'm unsure about into a separate branch for you guys to review and make some sense out of ;)

nathanwoulfe:
Hell might have been a bit of overstatement...

Makes sense to branch off anything not specifically related to block editor - I know there are changes around standardising presentation of grid and nc which should stand alone. 

Been a while since I've looked at this, will try to find some time today

Shazwazza:
I have cleaned up quite a lot in here a and simplified it all. There was directives that shouldn't be directives, lots of usages of ng-include and have removed the thumbnail feature thing for now. It's mostly readable and reviewable now. 

I will clean it up more with more components, etc... but first I'm going to merge this into the v8/feature/block-editor branch and work directly from there instead of in callum's fork.

JeffreyPerplex:
@Shazwazza Here you go with your 'slick piece of work': [https://vimeo.com/336065742/dcddcd690c](https://vimeo.com/336065742/dcddcd690c).

Haven't updated the video for a while and will try to share the version we're currently using soon (but have to find some time to prioritize the right stuff after my vacation :))

PR #6074, opened 2019-08-08
Move CreateRoutes from WebFinalComponent to WebInitialComponent

### LINKED ISSUE #6045: Custom routes break umbraco backoffice
Forum topic: https://our.umbraco.com/forum/umbraco-8//98388-custom-routes-break-umbraco-backoffice

Umb version is 8.1, the same was in 8.0.2
In general: it looks that simple as described on forum. routes work just fine, but...
lot of functionallity in backoffice start using my routes and get 404. (in doc types the most, network tab totally red). So to use backoffice i have to comment out route lines... and add them back for site. Kinda frustrated thing. Any suggestions?

Would be cool to get some comments on forum topic.

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #6045

### Description
This implements Solution 1 discussed in #6045, moving the registration of Umbraco's routes into WebInitialComponent so that custom routes can more easily be placed after Umbraco's routes.

To test:
- Create an IUserComposer and IComponent registering a very general custom route, such as the standard `{controller}/{action}`.
- Check that the route doesn't take priority over more specific routes needed by Umbraco.

### DIFF
--- a/src/Umbraco.Tests/Routing/RenderRouteHandlerTests.cs
+++ b/src/Umbraco.Tests/Routing/RenderRouteHandlerTests.cs
@@ -40,7 +40,7 @@ namespace Umbraco.Tests.Routing
         {
             base.SetUp();
 
-            WebFinalComponent.CreateRoutes(
+            WebInitialComponent.CreateRoutes(
                 new TestUmbracoContextAccessor(),
                 TestObjects.GetGlobalSettings(),
                 new SurfaceControllerTypeCollection(Enumerable.Empty<Type>()),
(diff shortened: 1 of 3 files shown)


poornimanayar:
Hi @stevemegson ,

Thank you for your PR. We shall review it and get back to you. Do you think you can add any notes on how we can test this?

Regards,
Poornima

nul800sebastiaan:
It seems that this works as expected and I can't really see a scenario where this would be breaking for people either (as explained in the related issue).

Many thanks @stevemegson to take the time to move this code around! 👍

PR #6078, opened 2019-08-08
V8: Sorting of images in multiple media picker

### LINKED ISSUE #5504: Manual sorting of the images in Multiple Media Picker doesn't work in Umbraco 8
I want to sort the images in a Multiple Media Picker but it doesn't work.

The version of the application is Umbraco 8.0.2

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #5504

### Description
The order of execution when setting up a media picker has changed in v8, so that `$scope.isMultiPicker` is read before it has been initialized. This leads to sorting being disabled, because sortableOptions is always set up as though it's a single media picker.

Testing:
 - Add some items to a multiple media picker property editor
- Check that you can sort them by dragging

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.controller.js
@@ -229,7 +229,7 @@ angular.module('umbraco').controller("Umbraco.PropertyEditors.MediaPickerControl
 
 
         $scope.sortableOptions = {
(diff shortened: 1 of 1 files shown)


FransdeJong:
@nul800sebastiaan Is it possible to include this in 8.1.2? Or has that ship sailed already?

nul800sebastiaan:
Well that was silly easy! 🤦‍♂️

Thanks so much @stevemegson 🎉 great find!

I'll get it into 8.1.2, seems like a fantastic little much needed improvement. :)

Issue #6075, opened 2019-08-08
v7.15.1 Property Editior - Media Picker - Upload - button is disabled


elliottjb7:
When creating/editing content the media picker's upload button is no longer active. If I have a need to upload a new media file, I can no longer do so from within the media picker, I now need to manually go into the media location and upload there. After that I can return to the media picker and choose the recently uploaded media. Not Ideal.

On inspection (Chrome/F12) I can see that the button is set to disabled. Please fix

---
_This item has been added to our backlog [AB#2213](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2213)_

nul800sebastiaan:
Thanks for reporting @elliottjb7 - this seems to be a little wonky indeed! We'll get it looked at, thanks for reporting!

kjac:
@elliottjb7 looks like there's a temporary workaround for you: If you browse the node structure in the picker (open a folder and go back to where you came from), the upload button starts working.

I'm digging into the root cause now.

kjac:
Upon further investigation... this only happens when Umbraco can remember your last opened media node (it's in localstorage). I found the culprit, PR incoming.

kjac:
PR in #6081

kjac:
Turns out this is an issue in V8 as well; PR for V8 in #6082

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6082

nul800sebastiaan:
Cherry picked to 7.15.2: 79c4d7797cc67ddc70f195e62961addd7613ef2b
And 8.1.3: c87a22f3475bc14f51a42690dc91afa0fabcb63b

PR #6081, opened 2019-08-08
V7: Fix the disabled upload button in media picker

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6075

### Description

If Umbraco can figure out which media folder you last visited, the "Upload" button in the media picker is disabled. If you browse around within the media picker, the "Upload" button immediately becomes enabled, even for the folder where it was initially disabled:

media-picker-upload-before

The issue was introduced with 1dd2c1cb. Basically it's due to a missing `path` property in the starting node of the media picker, whichever node that might be (except the root).

This PR fixes the issue; when it's applied the "Upload" button works as expected:

media-picker-upload-after


nul800sebastiaan:
And this one too, cheers! 🍻

PR #6082, opened 2019-08-08
V8: Fix the disabled upload button in media picker

kjac:
### Prerequisites

### Description

_This is the V8-ification of #6081_

If Umbraco can figure out which media folder you last visited, the "Upload" button in the media picker is disabled. If you browse around within the media picker, the "Upload" button immediately becomes enabled, even for the folder where it was initially disabled:

media-picker-upload-before

This PR fixes the issue; when it's applied the "Upload" button works as expected:

media-picker-upload-after


nul800sebastiaan:
Super, thanks very much @kjac !

Issue #6079, opened 2019-08-08
Icons for protected pages are not visible in the tree for list views (containers)


mzajkowski:
I've noticed that there is no indicator of protected pages (using single-user or role-based built in protection) when those pages are of the type of listview. In simple words: icons added for listview are hiding (overriding) the styles for the icons of protected ("cancel") pages.

Because of that it's not possible to indicate by looking on the tree which pages are protected and which are not.

**I've seen it's fixed in v8.**

The issue is raised for the v7.x.

![Image of HTML code for the tree](https://dzwonsemrish7.cloudfront.net/items/1F042u3R40231M2w0G10/%5Bb22de07ddaef3fd8f0fce335f1ecbf99%5D_Image%25202019-08-08%2520at%252012.59.31%2520PM.png?v=9d835eb8)

---

Desired result:

![Image of the desired result](https://dzwonsemrish7.cloudfront.net/items/300l1H013x3K0j3L1w43/Image%202019-08-08%20at%201.01.10%20PM.png?v=ee56732b)

nul800sebastiaan:
Good to hear this has been fixed in v8, that means it might be easier to find a way to fix it in v7 too. 

Care to have a look at it and see if you can create a PR for this?

umbrabot:
Hi @mzajkowski,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

kjac:
Pretty sure I fixed this for V8 at some point. Lemme have a quick look.

kjac:
Yep, same fix... PR in #6083

mzajkowski:
Jeeezzz.. love it! And you're a machine @kjac 💥 Big #h5yr and I love to contribute and collaborate like that!

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6083

Cherry picked for 7.15.2 9865596dcce3735ad2573d9fc596ff16ed4f9a90

PR #6083, opened 2019-08-08
V7: Change the priority of the tree node annotations for containers

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6079

### Description

As described in #6079 you can't tell if a node is protected if it's a container (list view) node. As it turns out, the same thing applies to nodes with pending changes:

image

This PR changes the priority of the tree node annotations to be:

1. Is protected
2. Has pending changes
3. Is container (list view)

When the PR is applied, the same nodes will look like this:

image


nul800sebastiaan:
SuperTak @kjac ! 🍾

PR #6085, opened 2019-08-08
V8: Set media picker dirty when the value changes

kjac:
### Prerequisites

### Description

For some odd reason the media picker doesn't flag the current form as dirty when its value changes:

media-picker-setdirty-before

This applies to:

- Picking new media.
- Removing existing media.
- Sorting (which may not be testable until #6037 is merged).

When this PR is applied, the dirty flag is set on the current form as soon as the media picker value changes:

media-picker-setdirty-after


poornimanayar:
Hello @kjac ,

Thanks again! We shall review it :-)

Poornima

nul800sebastiaan:
Nice one, cheers! 🥂

nul800sebastiaan:
Cherry picked for 8.1.4 65b2d08997a44b6d10cff62b710e4257bdd92e3b

jonroberts24:
This is also happening in Version 7.15.3 - I presume this isnt going to be fixed in 7?

PR #6086, opened 2019-08-09
Fix the gulp build to allow directives/components to exist where they need to be allowed

Shazwazza:
A big part of the problem with our angular solution setup comes down to how the gulp build is configured which hasn't really changed from the beginning of v7 with the grunt build. The build only lets certain file types and naming configurations to exist in certain places which adds the confusion about where everything needs to go.

A great example: Building a property editor.

A property editor's JS controller files and it's view files exists in: `views/propertyeditors/[myeditorname]/*`

which is fine, BUT when you start trying to mix directives/components in with your property editor you immediately run into a problem because of the way the build system works and you cannot just create components alongside your property editor, you have to stored these files in odd places which makes things very hard to find and work with.

This boils down to how we are combining files, which is like this:

```
{
    preview: { files: ["./src/preview/**/*.js"], out: "umbraco.preview.js" },
    installer: { files: ["./src/installer/**/*.js"], out: "umbraco.installer.js" },
    controllers: { files: ["./src/{views,controllers}/**/*.controller.js"], out: "umbraco.controllers.js" },
    directives: { files: ["./src/common/directives/**/*.js"], out: "umbraco.directives.js" },
    filters: { files: ["./src/common/filters/**/*.js"], out: "umbraco.filters.js" },
    resources: { files: ["./src/common/resources/**/*.js"], out: "umbraco.resources.js" },
    services: { files: ["./src/common/services/**/*.js"], out: "umbraco.services.js" },
    security: { files: ["./src/common/interceptors/**/*.js"], out: "umbraco.interceptors.js" }
}
```

Which leaves very little room for flexibility. This is also why we have odd folders like `src/common/directives/components` which actually map views to `src/views/common/components` ... because "directives" and components must exist in this directives folder.

Ideally directives/component each have a folder that contains both their script and html files just like we do for property editors. This would make things so much easier - like WAY easier to work with. 

We should also allow for controllers, components, directives to exists wherever they need to be.

I also don't see a huge reason why we have things separated into this particular files but if we want to just keep them as-is for now, we should just update these definitions like:

* Leave preview, installer, filters, resources, services, interceptors, controllers as they are
* Change directives/components to include anything in the directive/components folders as well as anything with the `.directive.js` or `.component.js` suffix if it doesn't exist in: installer, filters, resources, services, interceptors, directive/component folders.

```
{
	preview: { files: ["./src/preview/**/*.js"], out: "umbraco.preview.js" },
	installer: { files: ["./src/installer/**/*.js"], out: "umbraco.installer.js" },            
	filters: { files: ["./src/common/filters/**/*.js"], out: "umbraco.filters.js" },
	resources: { files: ["./src/common/resources/**/*.js"], out: "umbraco.resources.js" },
	services: { files: ["./src/common/services/**/*.js"], out: "umbraco.services.js" },
	security: { files: ["./src/common/interceptors/**/*.js"], out: "umbraco.interceptors.js" },

	//the controllers for views
	controllers: { files: ["./src/{views,controllers}/**/*.controller.js"], out: "umbraco.controllers.js" },

	//directives/components 
	// - any JS file found in common / directives or common/ components
	// - any JS file found inside views that has the suffix .directive.js or .component.js
	directives: {
		files: [
			"./src/common/directives/_module.js",
			"./src/{common/directives,common/components}/**/*.js",
			"./src/{views}/**/*.{directive,component}.js"
		],
		out: "umbraco.directives.js"
	},
}
```

## End result

This change makes things SOOOOOO much better. We could easily make this even better than what this PR gives us, but .... baby steps (i.e. why do we have a folder called `common` at all???)

* No more strange `/controller` folder, this was an odd one since we don't have any other controller folders, these 2 main controller files can just exist in the root
* We can now move the /common/directives/components to just be /common/components - if we wanted to
* We can now just store the *.html files alongside the JS files in /common/directives ... that alone is a massive improvement

We don't have to move everything at once, and nor should we since that will probably break a bunch of PRs, but people should/could start submitting PRs to move small chunks of stuff around to make working with these files much easier.


Shazwazza:
I don't know why it didn't detect renames for moving the 2x controllers, but all they are is moved, no changes.

kjac:
This is awesome 🎉 being able to have controllers and views next to each other is a huuuuge win, both for the overall code structure and not least for newcomers. The current structure isn't exactly newcomer friendly.

As for migrating the current structure to the new one... it will most likely always clash with one or more PRs and it will likely also be a pain when dealing with hotfix merges between current versions and vNext. It can't be helped. 

Perhaps the most feasible approach would be to let HQ do the move bit by bit and then ask the authors of any affected PRs to do a redo after the move?

Shazwazza:
Yep we can move stuff bit by bit, it's no rush really but would make things a lot nicer if we can put views next to their corresponding directive JS files and be able to put specific components/directives wherever we need them. 

The build structure change won't affect anything the way it currently is but offers up this flexibility. Plus i need to use this for the block editor prototype since we have some components in there that are specific to the property editor that should exist in the property editor folder and without this change i can't do that ;)

I'm going to merge this in since there's full compatibility with this change and the current structure.

Cheers!

Issue #5959, opened 2019-07-19
LocalizationService.GetLanguageByIsoCode my return language different to the one requested


KevinJump:
Calling the LocalizationService GetLanguageByIsoCode may return a 'parent' language to the one requested if it can't find it. 

e.g GetLanguageByIsoCode("en-US") will return the "en" language if the en-US language is missing. 

this may or may not be a bug but the method should be at least annotated to tell you that you might not get what you asked for - maybe there should be a fallback option exposed? 

ultimately it happens in the repo: https://github.com/umbraco/Umbraco-CMS/blob/e5956e4955474694c12ed6da7a4d5600acd0af8c/src/Umbraco.Core/Persistence/Repositories/Implement/LanguageRepository.cs#L240

kjac:
For reference: The reason for doing so is described in #4519 

The new option to localize the backoffice using the language files (#5358) may also weigh in on any decision to undo #4519 😄

ronaldbarendse:
Ouch, that's not a change you would add to a repository, especially without opt-in! 😮

And PR https://github.com/umbraco/Umbraco-CMS/pull/4519 doesn't correctly get the parent culture for cultures that contain multiple dashes or just different names (e.g. `zh-Hans-HK`, which parents are `zh-CHS` > `zh-Hans` > `zh`).

kjac:
I'll pick this one up 👍

kjac:
OK so... we can't solve this with an opt-in/overload on the affected method(s) without causing a breaking change on `ILanguageRepository`. 

Given how #5358 now allows us to localize the backoffice independently of the dictionary and the active languages, perhaps the best approach here is to simply revert #4519 and no longer (attempt to) support fallbacks to culture invariant translations in the dictionary? Unless there is some Cloud implication I am unaware of here - can you have user defined language overrides on Cloud?

nul800sebastiaan:
Thanks @kjac - I know you slept on it and had some more clever ideas, look forward to seeing those!

kjac:
@nul800sebastiaan you got it 😆 new PR in #6090 reverts the changes from #4519 and adds the same feature by other means.

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6090

Cherry picked to 8.1.3 in 87845ee907c65a5ba09c62803b98c34649fde6ba

PR #4519, opened 2019-02-11
V8: Can't localize dynamic items (e.g. content types) using the dictionary

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description

If you have a region-less language enabled (e.g. "da" instead of "da-DK"), you can't translate dynamic items (e.g. content types) using the dictionary. This is because `LanguageRepository` can't resolve the reagon-less language from the current thread culture.

This PR lets `LanguageRepository` fall back to the region-less language if it can't find the specific language.

### Steps to reproduce

1. Add "Danish" to the active languages (*not* "Danish (Denmark)"):
image
2. Create a content type named "#MyPage" and allow it at root. 
3. Add a dictionary item named "MyPage" and enter values for English and Danish.
image
4. With `en-US` set as editor language, the translation works as expected:
image
5. With `da` set as editor language, the translation doesn't work - the content type simply displays as "MyPage" (without the "#"):
image

PR #6090, opened 2019-08-10
V8: Allow localization of the backoffice using parent cultures to the editor culture

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5959

### Description

This is a new take on the issue outlined in #4519. The PR reverts the changes from #4519 and handles dictionary based localization via region-less languages in a different way.

#### Setup for testing this PR

First and foremost make sure you have created both a region specific and a region-less language - e.g. add "da" if you still have the default "en-US":

image

Create a doctype that localizes both its name and a property label via the dictionary (using the "#" notation):

image

Add two dictionary entries, one for the name of the doctype and one for the name property label:

image

image

#### Testing this PR

Set your user language to match the region-less language you created above - e.g. use "da-DK" as your user language if you created "da". 

Verify that the localization works by testing the following:

1. The doctype name in create dialog:
image

2. The property label in the doctype editor:
image

3. The doctype name in the "Info" section:
image

Now change your user language to match the region specific language you created above and redo the verification of localization for this language.

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/LanguageRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/LanguageRepository.cs
@@ -250,7 +250,6 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
             lock (_codeIdMap)
             {
                 if (_codeIdMap.TryGetValue(isoCode, out var id)) return id;
-                if (isoCode.Contains('-') && _codeIdMap.TryGetValue(isoCode.Split('-').First(), out var invariantId)) return invariantId;
             }
             if (throwOnNotFound)
                 throw new ArgumentException($"Code {isoCode} does not correspond to an existing language.", nameof(isoCode));
--- a/src/Umbraco.Tests/Persistence/Repositories/LanguageRepositoryTest.cs
+++ b/src/Umbraco.Tests/Persistence/Repositories/LanguageRepositoryTest.cs
@@ -80,39 +80,6 @@ namespace Umbraco.Tests.Persistence.Repositories
         }
 
         [Test]
-        public void Can_Perform_Get_By_Invariant_Code_On_LanguageRepository()
-        {
-            var provider = TestObjects.GetScopeProvider(Logger);
-            using (var scope = provider.CreateScope())
-            {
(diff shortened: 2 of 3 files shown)


src/Umbraco.Core/Persistence/Repositories/Implement/LanguageRepository.cs
```diff
@@ -250,7 +250,6 @@ public ILanguage GetByIsoCode(string isoCode)
             lock (_codeIdMap)
             {
                 if (_codeIdMap.TryGetValue(isoCode, out var id)) return id;
-                if (isoCode.Contains('-') && _codeIdMap.TryGetValue(isoCode.Split('-').First(), out var invariantId)) return invariantId;
```
Shazwazza: @kjac are you sure about this change?? IIRC there's a reason for this check and the unit test that has been removed. Because we have odd language files that are only 2 chars but we have one that is properly named `en-us`

Shazwazza: Ah, maybe this was a change you made at some point - i guess my only concern is that there is some odd code in here to deal with the poorly named localization files that are both 2char and the normal iso codes

kjac: @Shazwazza yah this PR reverts the changes made by yours truly in #4519 and re-implements the same feature in another way, as #4519 introduced issues for uSync (and was likely to do so for others in the future as well). 

All of this only concerns the dictionary, not the language files. Here's what I'm trying to achieve:

You're able to create region-less languages for frontend rendering (e.g. "da") and thus you have a region-less dictionary items. However you can't have a region-less user locale - it's always region specific (e.g. "da-DK"). So these changes are trying to bridge that gap, allowing for localizing the backoffice for the region specific user locales using region-less dictionary items.

nul800sebastiaan:
I checked all the usages of this now a couple of times and this is a much better fix in the first place. Good find of that other method, it seems to be exactly made for this and we don't have to break the other one either (or at least.. it's being "unbroken" now!). 

Thanks very much! 👍

PR #6092, opened 2019-08-11
Ability to distinguish document types with same name in Nested Content config

ceee:
### Description

This pull request updates the **Nested Content** property editor so you can distinguish document types which have the same name by appending the alias.

It also disables the "Add" button in case no available document types are left.

### Before

image

### After

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
@@ -15,6 +15,10 @@
             });
         }
 
+        $scope.canAdd = function () {
+            return !$scope.model.docTypes || !$scope.model.value || $scope.model.value.length < $scope.model.docTypes.length;
+        }
+
         $scope.remove = function (index) {
             $scope.model.value.splice(index, 1);
         }
@@ -60,7 +64,6 @@
 
             // Count doctype name occurrences
             var docTypeNameOccurrences = _.countBy(docTypes, 'name');
-            console.info(docTypeNameOccurrences);
             
             // Populate document type tab dictionary
             // And append alias to name if multiple doctypes have the same name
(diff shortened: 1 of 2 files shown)


ronaldbarendse:
I've come across this multiple times too, having to pick the right one by trial and error 😖

I'm not sure changing the names is the right solution though, especially only for items with duplicate names: this makes the naming unpredictable, adds additional logic and might conflict with existing names (e.g. if you also have `Image (picture)`, which would still result in 2 items with the same name). Maybe we should just show both by default or just the alias?

Other UI solutions to address this issue might be:
* Adding `optgroup`s with the folder hierarchy (as most likely the element types with the same name are in different document type folders)
* Showing the alias of the selected item next to/below the select

ceee:
@ronaldbarendse That may be an option too, yes. 
Initially I have tried to suffix duplicate items to reduce visual noise. For sure it would be possible to add the alias to all items – wouldn't be a problem to do and reduce complexity.
Let's wait for an opinion from the HQ.

skttl:
I like this, and also the `optgroup` suggestion for the folder hierarchy.

I think this UI needs a bigger overhaul than this, but let's see what the new block based editors end up with. I'm thinking an infinite editor for each of the doctype settings, so you can get more space for the template (and maybe even a helpful "paste snippet" function to get started).

The doctype should IMO be picked in a dialog, where we can present the available doctypes in a better way, showing their icon and description too (you can't do that in a select box).

ceee:
> I think this UI needs a bigger overhaul than this, ...

Definitely, yes! I guess (or hope) a lot of issues are going to be addressed with the new block editor.
This PR should just fix what I believe is a bug, cause it this problem occurred in a recent real project.

nul800sebastiaan:
I think this is a great update for now, and yes, it would be nicer to have a modal doctype picker instead in the future (hint, hint 😉).

ronaldbarendse:
I've been playing with this and just adding `ng-attr-title="{{config.ncAlias}}"` to the select (to show the alias of the selected element type as a tooltip) would probably be enough (and IMHO also better then the current fix). This is in-line with how aliases of properties are currently shown in the back-office.

Or to make it more visible, just show it as a label in a new column:
Element alias

Creating a hierarchy using `optgroup`s could be shown like this:
Element type hierarchy

These options could even be combined to always show the alias and have the hierarchy context. @nul800sebastiaan What's your view on this (before I put more work into this 😉)?

skttl:
+1 for the optgroups :)

nul800sebastiaan:
Yep. I totally forgot that optgroups existed, looks great!

PR #6093, opened 2019-08-11
v8: Propagate default Searchers

bielu:
### Prerequisites
I notice Umbraco v8 doesn't propagate basic searchers for Indexes, and from my perspective, it is not correct behavior. As some API could use a list of searchers and they will be not able to access default index searchers.

### Description
Actual behavior, when you will go to back-office Umbraco will display empty section of searchers as default searchers for Indexes are not propagate:
image
After applied my PR:
image
As you could see now on the list of Searchers is containing default searchers for indexes. 
@Shazwazza it was any reason behind not propagating default searchers?

### DIFF
--- a/src/Umbraco.Web/Search/ExamineComponent.cs
+++ b/src/Umbraco.Web/Search/ExamineComponent.cs
@@ -101,8 +101,12 @@ namespace Umbraco.Web.Search
             }
 
             //create the indexes and register them with the manager
-            foreach(var index in _indexCreator.Create())
+            foreach (var index in _indexCreator.Create())
+            {
                 _examineManager.AddIndex(index);
+                _examineManager.AddSearcher(index.GetSearcher());
+            }
+
 
             _logger.Debug<ExamineComponent>("Examine shutdown registered with MainDom");
 


Shazwazza:
The reason being is because a standalone searcher is like a "Multi-index searcher"... a searcher that is not resolved directly from the indexers. The only reason the searcher collection exists is to do with this case.

I'm unsure the consequences of doing what you propose here because now the searchers will be double disposed and this concept is not something I've tested.

bielu:
Actually I was looking for way of get list of all indexes in umbraco, but in v8 it is not really direct way of doing that. I could search for all implementation of IIndex but I don't feel it is correct of doing that. And just from perspective of creating easy search why so need resolve index and later try to resolve searcher?

Shazwazza:
A searcher is created by an index because you need to have an index to search one and that's how the APIs work. Before in Examine this wasn't the case but that meant that the abstractions were totally broken because. A searcher is dependent on the index it comes from, thus the searcher in most cases needs to be created from the index.

The term "Configured Searchers" was to indicate that it's searchers that are explicitly configured. 

You can get a list of *All* searchers by combining the searchers from indexes and the explicitly configured searchers. There's no need to check for `IIndex` or anything. All indexers return a searcher.

An extension method could be created to return *all* searchers quite easily. I just don't think we want to add searchers coming from the configured indexes to the explicitly registered standalone searchers since that was never the intention.

Shazwazza:
Sorry, end of the day for me, will have a look at code tomorrow - you basically just want to get a list of all searchers, whether they are from indexes or based on stand along searchers right?

bielu:
Shawazza yes :)

Shazwazza:
can't you just do this?

```cs
public static class ExamineManagerExtensions
{
    public static IEnumerable<ISearcher> AllSearchers(IExamineManager mgr) 
        => mgr.Indexes.Select(x => x.GetSearcher()).Concat(mgr.RegisteredSearchers);
}
```

bielu:
@shazwazza like you could notice in last message I am doing exactly the same but no in extension, I more afraid about second part (of my last message) when I am trying access to searcher by name and have to first check with trygesearcher and later via trygetindex and later request .getsearcher()
62864874-cdb1a100-bd04-11e9-9cec-a85e956030a3

Shazwazza:
doh! sorry i didn't see your screenshot

I'm just not understanding the use-case here for the name check. Is someone typing in arbitrary name in to find either an index or a searcher?

bielu:
@Shazwazza I am working on recreate nuPickers actually and we have Lucene pickers, on which you could choose index, but that index could be removed/added/renamed and I prefer to be sure that index exists before it will be quering that

Shazwazza:
I guess I would suggest storing whether they chose to use an index or a searcher and then ask for the appropriate one otherwise I'd say try to find it in the list of indexes and then if not found, check the searchers - explicit searchers are going to be much more rare, it's really just for multi-index searching or if you somehow wish to have a searcher for a readonly index of some sort.

bielu:
I am closing that as yes it is solving my problem but still have feeling maybe that extension should be part of Umbraco, as I am sure not only nuPickers are using the list of searchers :)

Issue #6095, opened 2019-08-11
Rename component for flatpickr


bjarnef:
At the moment Umbraco v8 has a directive/component named `umbFlatpickr` for a daterange picker which is using the Flatpickr library: https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/common/directives/components/umbflatpickr.directive.js

The name is very much bound to the name of the library and a bit inconsistent with naming of other components/directives in core. For example we have a slider named `umbRangeSlider` using noUiSlider library, but we don't call this component `umbNoUiSlider` 😅

Furthermore we might also replace these libraries in the future because of various reasons: The library is no longer maintained, there is a better or more lightweight library with same or more features or for a third reason.

There has been some confusions, because in v7 we have `umbDateTimePicker`, and in v8 it is a very different name for this component.
https://twitter.com/m1ketayl0r/status/1156955466884419586
https://twitter.com/mattbrailsford/status/1156965431019757568
https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Web.UI.Client/src/common/directives/components/umbdatetimepicker.directive.js

Previous in v7 we could also use `umbEditor`, but later we could use a more specific name `umbPropertyEditor`.
https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyeditor.directive.js#L40-L42

We could use the same approach here, so we can use a better naming of this component, e.g. `umbDatetimePicker` or something similar, but keep `umbFlatpickr` for backwards compatibility.

nul800sebastiaan:
Sure thing, if we can do it in a backwards-compatible way then that's great!

umbrabot:
Hi @bjarnef,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

kjac:
Shoot sorry, accidentally mentioned this issue on #6118 - it has no relevance to #6118 whatsoever.

PR #6110, opened 2019-08-12
v8: Rename datetime picker component

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6095

### Description
The changes in this PR allow us to re-use the datetime picker component as `umb-date-time-picker` in views as it was in v7 https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Web.UI.Client/src/common/directives/components/umbdatetimepicker.directive.js

I considered `umbDatetimePicker` (umb-datetime-picker) and `umbDateTimePicker` (umb-date-time-picker), but even the latter one is a char longer in views, it kept the same naming as in v7.

It should still be possible to use the name `umb-flatpickr`, but I have renamed the instances in core to `umb-date-time-picker` to encourage developers to use this name of the component as we might remove the `umb-flatpickr` in the future.


poornimanayar:
Hi @bjarnef ,

Looks like a really useful one. Is it possible you can add some notes on how we can test this efectively?

Poornima

bjarnef:
@poornimanayar

- Test that datetime property editor works
- Test that datetime picker in e.g. schedule publishing works
- Test that using `umb-flatpickr` works in e.g. a dashboard.

Issue #5206, opened 2019-04-09
Umbraco exposes all tags publicly by default


ronaldbarendse:
Umbraco provides a `TagsController` (an `UmbracoApiController`) by default that allows to publicly query all tags within the site with a simple GET-request, e.g. `https://umbraco.com/umbraco/api/tags/GetAllTags`.

The controller is available in both Umbraco 7 & 8, although it is moved in V8:
- https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Web/WebServices/TagsController.cs
- https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web/Controllers/TagsController.cs

The following remarks are added in the source, so I don't think this is a BIG security issue (that's why I'm creating this as an issue):

> This controller does not contain methods to query for content, media or members based on tags, those methods would require authentication and should not be exposed publicly.

As tags could still expose information that shoudn't be public (e.g. classifications on media or members) and I don't see why Umbraco would include this controller by default, I would recommend removing this controller.

Shazwazza:
Tags are used for published content and as such are available publicly but that said i do understand that in some rare situations it isn't ideal to have this data public. Removing this controller would result in breaking changes but it is possible for you to remove this controller from the resolvers/containers on startup so it is not auto routed. This controller has been around since 2013 and although i made it, i can't recall the exact reason but I think it was during a trial of allowing easier 'headless' style data access.

So the questions is - how do we remove this? I think we will need to have an opt-in config flag that disables this controller from being routed. In new releases we will ship with this config flag so it's not routed which means for upgrades people will not be affected but they can opt-in with the config switch.

ronaldbarendse:
I would argue removing the controller, but marking it as a breaking change would be the best option. If the controller is used in a website (which is unlikely, as this isn't documented), copying the controller to the website project is enough to restore it...

If you could provide an example of how to remove this controller on startup, a temporary 'fix' could be implemented for those that are also concerned about exposing all tags 👍

Shazwazza:
I wrote a quick blog post about how to remove routes during startup so this can be achieved: https://shazwazza.com/post/need-to-remove-an-auto-routed-controller-in-umbraco/ 

Will chat to the team about this controller soon and get back to you

Cheers!

ronaldbarendse:
Thanks @Shazwazza!

[...]

PR #6114, opened 2019-08-13
Disable Mini Profiler

elit0451:
Mini Profiler will be disabled when a solution is not in a debug mode, therefore Mini Profiler Startup Profiler won't be accessible.

Fixes: [AB#2178](https://umbraco.visualstudio.com/D-Team%20Tracker/_workitems/edit/2178)

### DIFF
--- a/src/Umbraco.Web/Profiling/WebProfilerProvider.cs
+++ b/src/Umbraco.Web/Profiling/WebProfilerProvider.cs
@@ -1,7 +1,10 @@
 using System;
+using System.Linq;
 using System.Threading;
 using System.Web;
+using System.Web.Routing;
 using StackExchange.Profiling;
+using Umbraco.Core.Configuration;
 
 namespace Umbraco.Web.Profiling
 {
@@ -24,6 +27,19 @@ namespace Umbraco.Web.Profiling
         {
             // booting...
             _bootPhase = BootPhase.Boot;
+
+            // Remove Mini Profiler routes when not in debug mode
+            if (GlobalSettings.DebugMode == false)
+            {
+                var prefix = MiniProfiler.Settings.RouteBasePath.Replace("~/", string.Empty);
+
+                using (RouteTable.Routes.GetWriteLock())
+                {
+                    var routes = RouteTable.Routes.Where(x => x is Route r && r.Url.StartsWith(prefix)).ToList();
+                    foreach(var r in routes)
+                        RouteTable.Routes.Remove(r);
+                }
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Shazwazza:
@elit0451 great work! 

@nul800sebastiaan I have merged and cherry picked to dev/7.15 branch ready to go for the next 7.15.3 patch.

ronaldbarendse:
This looks like a really ugly hack, as it removes ASP.NET MVC routes based on URL... Why not make sure the MiniProfiler doesn't add the routes to begin with (https://stackoverflow.com/questions/37681166/is-it-possible-to-disable-mini-profiler-handler)?

ronaldbarendse:
Or at least remove the routes based on the name: https://github.com/umbraco/Umbraco-CMS/issues/5206#issuecomment-482024222

Shazwazza:
@ronaldbarendse because the old mini profiler isn't that flexible. The code that creates the routes in the old profiler adds the routes on a static constructor so as soon as that class is referenced for any reason the routes are created. So yes it's a hack, but it's the only way to prevent this. 

In 8 we don't need to do this and the new mini profiler is done in a nicer way without running logic on a static ctor.

ronaldbarendse:
I see, but removing the route based on the name would be better then on some configured URL prefix (that could therefore match more than the MiniProfiler route).

The route name is based on the prefix, so it's a quick rewrite: https://github.com/MiniProfiler/dotnet/blob/04f8ac465349a85a77d9fa4c6528fb75e63e157c/src/MiniProfiler/MiniProfilerHandler.cs#L59-L63.

Shazwazza:
Sure, if you have time to send a PR that would be superb

Shazwazza:
One thing to note is that code is referencing the newer version of MiniProfiler, you'll have to check back in history to when this was done in the static ctor and check what the route name might be there. It might not have changed but would need to be sure.

ronaldbarendse:
Just checked the refactored code from MiniProfiler 2.1.0 and the route is added the same way, but as it uses `Insert`, it's not using a named route :grimacing:

The current fix is still too greedy, as it could still find the route based on the exact URL (instead of 'starts with'). The route URL is constructed as follows:

```c#
string prefix = MiniProfiler.Settings.RouteBasePath.Replace("~/", string.Empty).EnsureTrailingSlash();
string routeUrl = prefix + "{filename}";
```

The `EnsureTrailingSlash()` extension method should probably be changed to use the Umbraco built-in `EnsureEndsWith()`, but otherwise it's a quite simple change (I currently don't have the time the create a PR for it though).

Issue #6099, opened 2019-08-12
Member password reset checkbox doesn't hide "new" and "confirm" password fields.


tomhicklingnc:
On an Umbraco 8 site, editing the password of a member presents you with a checkbox entitled "Reset password" and 2 password boxes. On previous versions of umbraco, ticking the "Reset password" checkbox would hide the two password boxes, and upon saving the changes would show you the new random password that was generated for the member. In 8, the checkbox doesn't seem to do anything, and saving the page with it checked presents validation errors if you didn't also put values into the "new" and "confirm" fields.

---
_This item has been added to our backlog [AB#2266](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2266)_

nul800sebastiaan:
Hey there @tomhicklingnc - thanks, I can reproduce in 8.1.2! 

I hear that this might be a merge issue (merging changes from v7 to v8). I'm sorry that this one slipped through, we'll work on it soon!

kjac:
PR in #6119

Shazwazza:
I'm fixing this up but it turns out that member saving has other issues and the changed values are not re-bound to the UI after saving so am fixing this too.

Shazwazza:
New PR: https://github.com/umbraco/Umbraco-CMS/pull/6120

nul800sebastiaan:
It

nul800sebastiaan:
This seemed fix but running a quick test just now shows it wasn't completely fixed, see PR #6120 
I'm moving this to the next release, unfortunately we won't have time to fix it for 8.1.4 which is due out tomorrow morning.

nul800sebastiaan:
Well, that is weird! After retesting in several browsers, this works as expected. I'll have to do a thorough cache clean in my Chrome. 8.1.4 is still good then! :)

PR #6119, opened 2019-08-13
V8: Make "reset password" for members work again

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6099

### Description

See #6099 for details.

The "reset password" checkbox is supposed to let admins reset a member's password without supplying old and new passwords. In 8.1.2 the checkbox does nothing:

reset-password-before

This PR fixes it; when applied, here's how the checkbox works:

reset-password-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/users/changepassword.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/users/changepassword.directive.js
@@ -17,8 +17,6 @@
       }
       */
 
-      $scope.showReset = false;
-
       //set defaults if they are not available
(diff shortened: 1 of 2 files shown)


Shazwazza:
That's not quite it ... the showReset is a configuration value so it's not entirely based on the checkbox.

I'm fixing this locally now. The issue is because of the change from v7 (old angular ) to v8 newer angular and the references to $parent $scope and boolean (non-reference value types). The fix is converting this to a real component with controllerAs `vm` which means that non-ref values exist on a base object not just on their own.

PR #6120, opened 2019-08-14
Fixes: Member password reset checkbox doesn't hide "new" and "confirm" password fields. #6099

Shazwazza:
Fixes: Member password reset checkbox doesn't hide "new" and "confirm" password fields. #6099

There was 2 problems here:

1. With the upgrade from v7 -> v8 and the the angularjs upgrade, the change-password directive had to be changed because the new angular version didn't support what the old one was doing with $scope's since the old one was using a lot of old $parent.$scope type of syntax. But changing it to use properties based on the normal $scope meant that non-reference (value) types such as boolean don't get properly watched by angular because it's not attached to a reference type. This could have been fixed in a couple of ways but the best way was to convert this old directive to a proper angular component with controllerAs syntax so that everything that is bound in the html file is based on a reference type (i.e. `vm.`) so then all watching of non-reference types like boolean values all work.
1. `contentEditingHelper.reBindChangedProperties` was broken for media and members to resync their values after saving. This has probably caused other odd issues with editing media/members that people may not have been aware of. This is now fixed so when you reset the password the re-bound generated password value is shown.

Testing: Make sure that member password reset all works correctly like it does in v7.


warrenbuckley:
Code changes seem fine to me with the change from `$scope` to `vm.`
Off to test & ensure this works like V7

warrenbuckley:
All good - merged in 🎉

warrenbuckley:
TO make @nul800sebastiaan happy - the `v8/8.1` branch has been merged into `v8/dev`

nul800sebastiaan:
This doesn't seem to work.

bergmania:
@nul800sebastiaan, It works for me on `v8/8.1` branch, if I build the frontend code. (Note this is hard currently. https://github.com/umbraco/Umbraco-CMS/pull/6169#issuecomment-527117568)

PR #6125, opened 2019-08-14
Removing legacyAjaxCalls.ValidateUser method

elit0451:
Removes an old web service endpoint which that can be used to bypass account lockout when validating credentials.

## Workaround

A work around can be put in place to prevent access to this endpoint using an IIS Rewrite rule.

In the Web.config, inside `<rewrite><rules>` tags:

```xml
<rule name="Fail bad requests" patternSyntax="ECMAScript">
  <match url="umbraco/webservices/legacyajaxcalls\.asmx"/>
    <conditions>
      <add input="{QUERY_STRING}" pattern="op=ValidateUser" />
    </conditions>
    <action type="CustomResponse" statusCode="403" statusReason="Forbidden: Access is denied." statusDescription="You do not have permission to view this directory or page." />
</rule>
```

And with the CustomResponse you can put response of your choice.

Any doubts about IIS Rewrite Rules setup -> https://our.umbraco.com/documentation/reference/Routing/IISRewriteRules/


Shazwazza:
_Cherry picked to 7.15.3_

PR #6129, opened 2019-08-14
V8: Don't reuse previously used content templates when creating new content

kjac:
### Prerequisites

### Description

Slightly odd scenario here:

1. Start creating content based on a content template. Don't save.
2. Now create content of another type that does *not* use a content template.
3. The editor still attempts to create content of the other type based on the previously selected content template.

content-template-reuse-before

Similar (and slightly less odd) scenario:

1. Start creating content based on a content template. Don't save.
2. Now create content of the same content type but _not_ based on a content template.
3. The previously selected content template data still lingers in the editor.

content-template-reuse-before-2

This PR fixes both of these scenarios - here's how it works when the PR is applied:

content-template-reuse-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/content.create.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/content.create.controller.js
@@ -51,7 +51,10 @@ function contentCreateController($scope,
             .search("create", "true")
             /* when we create a new node we want to make sure it uses the same 
             language as what is selected in the tree */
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Wow, nice catch @kjac

nul800sebastiaan:
Cherry picked for 8.1.4 455395154a569acc7a4c0c1f04d2f8849eae8a67

Issue #5925, opened 2019-07-17
[8.1] NuCache taking a long time to reload memory cache on large websites


DMCK96:
**Description
------------**

Reloading the memory cache on large sites takes an extremely long time when reaching large numbers. This issue does not occur for the content store, only the media store.

Content Store (119K items) took approximately 3-5 minutes to reload.
Media Store (189K items) took approximately 5-6 hours to reload. 

We've noticed the media store is falling over quite frequently and returning media items as null and the only fix we've found is to reload the memory cache although this takes a very long time to complete. This is causing 5 hours of daily downtime to our site.

We narrowed down the issue to the 'AddNodeLocked' method in the NuCache ContentStore. The items cached per minute progressively gets slower. We measured about 5-7K items being cached per minute up to around 70K items at which it had been reduced to about 10% of this speed and continued to slow down as the size of the cache got larger and larger.

_Note for the above: This only occurs on the media store._ 

**Reproduction
------------**

1. Start site with 119K content nodes and 189K media items.
2. Have media items all within a single parent folder
3. Trigger a reload of the memory cache from within 'Settings -> Published Status'

**Specifics
------------**

This issue only began occurring in 8.1, in 8.0 whilst the NuCache was still falling over frequently we could complete a reload of both content and media store in ~5-10 minutes.

No errors are produced in the log during this time. We ran a profiler and followed the source code to determine the cause of the issue. 

**Expected Result
------------**

Full NuCache reload should complete in less than 10 minutes.

**Actual Result
------------**

Full NuCache reload takes over 5 hours to complete.

---
_This item has been added to our backlog [AB#1844](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/1844)_

zpqrtbnk:
Discussed with OP, and all media items are under root. Meaning root has all media items as children. From 8.0 to 8.1, children went from being managed as arrays of ints, to linked lists, to reduce memory usage. Linked lists are fast when sorted, but slow when inserting and sorting at the same time. Inserting 190K ordered items is slow. Does not happen with content items because they have a tree structure, not all at the same level.

OK to keep linked list for snapshotted lists of children, but need an extra balanced tree for only the live snapshot, for fast inserting. Plus when reloading all, no need to sort during insert as things are coming already sorted.

opr:
[...]

PR #6131, opened 2019-08-15
Adds POC for fast building of in-memory nucache from persisted files

Shazwazza:
Fixes: NuCache taking a long time to reload memory cache on large websites #5925


Shazwazza:
The changes here are:

* On startup we use a new method to build the linked lists: `SetAllFastSorted`. Then we make sure the data is sorted by level + parent id + sort order (either from the db or from the cache files) and then we can just read forward to piece together the graph in a fast way without any additional lookups or looping. This method also bypasses all cloning and gen checks since it's ONLY used on startup.
* The linked lists that make the cache graph are updated to be double linked lists which improves performance when the site is running and items need to be added/removed/inserted into the graph. Previously this was only a single linked list which mean we'd have to iterate all siblings to determine where to update an item, now this is no longer necessary.
* There's a small optimization when looking up a parent since we were doing this 2x for each lookup to check first if it existed and then to find it, now it's just once
* Uses IProfilingLogger to log timing instead of manual stopwatch

You can see testing notes from the community here https://github.com/umbraco/Umbraco-CMS/issues/5925#issuecomment-522827498

Issue #4863, opened 2019-03-05
Element types should not inherit from non-element types


zpqrtbnk:
Apart from making no sense, an element type inheriting from a non-element type breaks Models Builder. See Models Builder [issue 188](https://github.com/zpqrtbnk/Zbu.ModelsBuilder/issues/188).

It should not be possible for an element type to inherit from a non-element type. Which means that, when picking compositions for an element type, it should only be possible to pick element types - and then, it should not be possible to flip an element type back to non-element if it is involved in a composition.

nul800sebastiaan:
While https://github.com/umbraco/Umbraco-CMS/pull/6136 doesn't fully fix the issue, I think we should have new issues created for the remaining parts. Closing this for now.

PR #6136, opened 2019-08-16
V8: Don't allow selecting non-element types when composing element types

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/4863 (in part)

### Description

If you pick a non-element type when composing an element type, ModelsBuilder throws up because the resulting content type cannot be expressed in code:

element-composition-before

This PR filters out any non-element types in the compositions picker when you compose element types:

element-composition-after

Kudos to @callumbwhyte for raising awareness on this issue (https://twitter.com/callumbwhyte/status/1161956628759007232)

#### Not a complete solution to the problem

This fix is by no means a complete solution to the issue described in #4863. This is only a first step (albeit an important one).

Later on the following should be implemented to improve on things:

1. It should not be possible to turn a non-element type into an element type if it is composed by or inherits non-element types.
2. It should not be possible to turn an element type into a non-element type if it is used as a composition for another element type, or if another element type inherits from it.

### DIFF
--- a/src/Umbraco.Core/Services/ContentTypeServiceExtensions.cs
+++ b/src/Umbraco.Core/Services/ContentTypeServiceExtensions.cs
@@ -22,12 +22,14 @@ namespace Umbraco.Core.Services
         /// This is required because in the case of creating/modifying a content type because new property types being added to it are not yet persisted so cannot
         /// be looked up via the db, they need to be passed in.
         /// </param>
+        /// <param name="isElement">Wether the composite content types should be applicable for an element type</param>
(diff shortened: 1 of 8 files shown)


poornimanayar:
Hi @kjac ,

Good one! Kudos to @callumbwhyte, I remember this one from UFD'19 hackathon :-) We shall test and get back to you!

Poornima

nul800sebastiaan:
Thanks @kjac - this is great for now, good protections in this one!

PR #6137, opened 2019-08-16
Secure svg media files preview

elit0451:
Prevent malicious JS content from .svg media files to be running inside of the back office.

Fixes: [AB#2180](https://umbraco.visualstudio.com/D-Team%20Tracker/_workitems/edit/2180)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/media/umb-media-node-info.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/media/umb-media-node-info.html
@@ -12,9 +12,15 @@
 
                 <ul ng-if="nodeUrl" class="nav nav-stacked" style="margin-bottom: 0;">
                     <li>
-                        <a href="{{nodeUrl}}" target="_blank">
+                        <ng-conainer ng-if="node.extension != 'svg'">
+                            <a href="{{nodeUrl}}" target="_blank">
+                                <i class="icon icon-window-popin"></i>
+                                <span>{{nodeUrl}}</span>
+                            </a>
+                        </ng-conainer>
+                        <a ng-if="node.extension == 'svg'">
                             <i class="icon icon-window-popin"></i>
-                            <span>{{nodeUrl}}</span>
+                            <span ng-click="openSVG()">{{nodeUrl}}</span>
                         </a>
                     </li>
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web.UI.Client/src/views/components/media/umb-media-node-info.html
```diff
@@ -12,9 +12,15 @@
 
                 <ul ng-if="nodeUrl" class="nav nav-stacked" style="margin-bottom: 0;">
                     <li>
-                        <a href="{{nodeUrl}}" target="_blank">
+                        <ng-conainer ng-if="node.extension != 'svg'">
+                            <a href="{{nodeUrl}}" target="_blank">
+                                <i class="icon icon-window-popin"></i>
+                                <span>{{nodeUrl}}</span>
+                            </a>
+                        </ng-conainer>
```
bjarnef: What is `ng-conainer`? I guess it should be `ng-container`?
https://alligator.io/angular/ng-container-element/

elit0451: @bjarnef you are absolutely right! Nice spot 👍

warrenbuckley:
@elit0451 this fixes the problem if you open the image from the link/info tab. However, if you click the SVG with the preview for the Upload File property it also opens the SVG with the XSS problem.

warrenbuckley:
Perhaps ignore me @elit0451 as I can now not reproduce it again 🤔 
Merged in :)

ronaldbarendse:
The `openSVG()` click event is only added to the `nodeUrl` text and as the `a` element doesn't have an empty `href` attribute, the cursor probably won't change to a pointer either.

The duplicated HTML can also be removed when using conditional AngularJS attributes:

```
<a ng-attr-href="{{node.extension != 'svg' ? nodeUrl : undefined}}" ng-click="node.extension == 'svg' && openSVG()"  target="_blank">
```

nul800sebastiaan:
Agreed, updated 851a2a3de8dc94d99b5a1b3d1d232b70a722aa96

Though `!=` should be `!==` and `==` should be `===`

Issue #6102, opened 2019-08-12
V8: Cannot edit dates in custom (en-GB) formats with date picker (flatpickr)


amitch17:
If a data type is set to use the date picker (Flatpickr) with a custom date format, it is not handled correctly when editing content. 

Dates may be parsed as invalid and blanked, or dates and months get switched around.

### Specifics

This occurs when editing content in the back office.
Umbraco 8.1.1 (cloud)
Observed in Chrome.

### Steps to reproduce

* Create a data type using the Date Picker property editor.
* Set the Date format to be "DD-MM-YYYY" This is valid according to momentjs rules.

![DateDataType](https://user-images.githubusercontent.com/18399155/62875763-7ec33600-bd1b-11e9-862c-94ffea4f6ccf.PNG)

* Save data type
* Create a document type, and a field that uses this Date type
* Create an instance of this document in the content tree
* edit the date field, select a date, e.g. 10th August, the date picker closes and displays 10-08-2019
* click elsewhere on the document (to blur the control)
* the date value switches over the digits to become 08-10-2019.  (it tries to parse the date in MM-DD-YYYY format?)

![datepicker-toggling digits](https://user-images.githubusercontent.com/18399155/62876869-c9de4880-bd1d-11e9-80c4-1a23953fcf66.gif)

Alternative:
* If a date is selected that is not valid in the original format, it is cleared.  e.g. Select 20th August, blur the control, and the value is cleared because 20/08/2019 is not valid in MM-DD-YYYY format.

![datepicker-invalid-date](https://user-images.githubusercontent.com/18399155/62876882-ce0a6600-bd1d-11e9-96b8-bd565ff877c4.gif)

It looks like the date is still selected in the dropdown calendar but there is no option to clear the value (the cross) in this case.

### Expected result

* When blurring the date control, the value in the date field should be unchanged.

Same problem occurs with date time picker.

nul800sebastiaan:
I think this is a duplicate of https://github.com/umbraco/Umbraco-CMS/pull/6001? Can you confirm please @amitch17?

amitch17:
Yes, this is a duplicate of that issue, well spotted mr @nul800sebastiaan  

I did search for flatpickr but didn't find this report. Sorry for the noise.

Haven't tested the fix myself but it looks to be doing the right things - can the PR by @stevemegson 
 get some attention please?

PR #6138, opened 2019-08-16
V7: Fixes #6132,  dates being blanked out on save/publish when user locale (momen…

### LINKED ISSUE #6132: The date picker is empty after save/publish for users with locales not matching the angular locale files
The date picker is empty after save/publish for users with locales not matching the angular locale files. (Probably affects more than the date picker)

Reproduction
------------

(done with Umbraco 7.13.1)

- Create a data type using the property editor Umbraco.DateTime.
- Keep the default format DD.MM.YYYY HH:mm (dunno if it affects anything)
- Keep offset on (dunno if it affects anything)
- Use the data type for a property

- Create a user with Norwegian locale
- Log into the backoffice
- Create doc. with the date picker
- Select a date
- Save & Publish

The date should now be blank, although stored. (save / publish now clears the date)
The date is back if the backoffice and/or content is reloaded.

### Bug summary

The Norwegian user locale is nb-NO. The angular-locale_nb-NO.js file is attempted loaded, but does not exist, since only angular-locale_no.js is bundled with angular 1.1.5.
This leads to the behavior described above.

### Expected result

Correct locale file loaded and dates shown after save/publish.

### Actual result

No locale file loaded and date blank after save/publish.

---

### Workaround

I work around it for now by copying the file on build:

    <Target Name="EnsureNorwegianLocaleFile" AfterTargets="CopyUmbracoFilesToWebRoot">
        <Copy SourceFiles=".\umbraco\lib\angular\1.1.5\i18n\angular-locale_no.js" DestinationFiles=".\umbraco\lib\angular\1.1.5\i18n\angular-locale_nb-NO.js" Condition="!Exists('.\umbraco\lib\i18n\angular-locale_nb-NO.js')"/>
    </Target>

### Suggested fix

Copy the file manually in Umbraco.Web.UI.Client since angular 1.1.5 won't be upgraded nor probably ever contain the file angular-locale_nb-NO.js. :)

lars-erik:
This should fix #6132 
Not much to say to expand on that. :)

### DIFF
--- a/src/Umbraco.Core/DateTimeExtensions.cs
+++ b/src/Umbraco.Core/DateTimeExtensions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text;
 
@@ -15,7 +16,7 @@ namespace Umbraco.Core
         /// <returns></returns>
         public static string ToIsoString(this DateTime dt)
         {
-            return dt.ToString("yyyy-MM-dd HH:mm:ss");
+            return dt.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture);
         }
 
         public static DateTime TruncateTo(this DateTime dt, DateTruncate truncateTo)
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
@@ -44,10 +44,10 @@ function dateTimePickerController($scope, notificationsService, assetsService, a
         if (newVal != oldVal) {
             //check for c# System.DateTime.MinValue being passed as the clear indicator
             var minDate = moment('0001-01-01');
-            var newDate = moment(newVal);
+            var newDate = moment(newVal, 'YYYY-MM-DD HH:mm:ss');
 
             if (newDate.isAfter(minDate)) {
-                applyDate({ date: moment(newVal) });
+                applyDate({ date: newDate });
             } else {
                 $scope.clearDate();
             }
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
@@ -44,10 +44,10 @@ function dateTimePickerController($scope, notificationsService, assetsService, a
         if (newVal != oldVal) {
             //check for c# System.DateTime.MinValue being passed as the clear indicator
             var minDate = moment('0001-01-01');
-            var newDate = moment(newVal, 'YYYY-MM-DD HH:mm:ss');
+            var newDate = moment(newVal);
 
             if (newDate.isAfter(minDate)) {
-                applyDate({ date: newDate });
+                applyDate({ date: moment(newVal) });
             } else {
                 $scope.clearDate();
             }


nul800sebastiaan:
Knowing nothing about moment, could you spend a few words on why this fixed the issue with a period in your datetime please? Does it just convert your localized Norwegian date/time to the hardcoded format?

And...... if you're hardcoding the format here, does it overrule the format you have set in the datype configuration, therefore.. is this the correct solution or should we use the format set in the datatype? (why/why not?)

lars-erik:
As far as I've understood the date is _always_ stored as `YYYY-MM-DD HH:mm:ss.fff` in the database.  
In the current (7.15.1) datepicker controller, when content is loaded, the date is parsed with moment and a hardcoded format to match the above:

`datepicker.controller.js (167):`

    var dateVal = $scope.model.value ? moment($scope.model.value, 'YYYY-MM-DD HH:mm:ss') : moment();

When the date is returned from the server it actually varies from the one in `$scope.model.value` because the latter has colons in it. (Probably due to moments locale settings)

The current then does `moment(newVal)` without a format on the date that comes from the server. As opposed to how it is parsed on load.

`datepicker.controller.js (47):`

    var newDate = moment(newVal);

This change merely makes the "on changed due to colon/dot transform when save/publish" do the same as on loaded, which apparently works fine.

I've tested this both with default format on the datatype and a "Norwegian" format on the datatype. (Blank vs. DD.MM.YYYY HH:mm)

stevemegson:
Although this works, I'm not sure that it really should. We're forcing moment to parse the date as `YYYY-MM-DD HH:mm:ss` even though we know it's really `YYYY-MM-DD HH.mm.ss`, and moment happens to handle that better than when we give it the same value with no format specified.

The real issue seems to be how we're formatting the date rather than how we're parsing it - the server should always be sending `HH:mm:ss` and not `HH.mm.ss`.

The code looks straightforward:

```c#
public static string ToIsoString(this DateTime dt)
{
    return dt.ToString("yyyy-MM-dd HH:mm:ss");
}
```

but in the format string, `:` means "the time separator in the current culture". When the current user's locale specifies `.` as the time separator, we generate incorrect "ISO" dates.

lars-erik:
Right, so rather change that line to `"yyyy-MM-dd HH\:mm\:ss"`, then?

stevemegson:
I guess we should pass `CultureInfo.InvariantCulture` into `ToString`, just in case any other locales have other surprises in store for us. There's bound to somewhere that formats years as roman numerals or something.

lars-erik:
Maybe, _but_ wouldn't this possibly break a lot of other things vs. nothing breaking from the change in the property editor controller?

stevemegson:
This seems to be the only place that actually calls `ToIsoString()`, so there shouldn't be any side effects.

lars-erik:
I'm thinking peeps might have used the value in views etc.  
But if so, we're probably helping "fix" or unobtrusively obsolete existing workarounds instead of breaking stuff.  
I'm making a debug build to test now.

lars-erik:
Looks good. :)

lars-erik:
As of now the fix is running in production for a customer until 7.15.3 ;)

nul800sebastiaan:
This last change certainly seems to make the most sense. 

>so there shouldn't be any side effects.

Famous last words 🙈 

Thanks for the help @lars-erik and @stevemegson - great detective work!

lars-erik:
\o/ 

#h5yr for great nudging in the right direction, @stevemegson!

crick3t:
Currently we have the same issue in Umbraco version 8.2.1, as the one on #6102 the second animation.
We have set the date picker to show the date in UK format (**DD/MM/YYYY**). After selecting the date if the focus is on the text box, the date is cleared out when it loses the focus. 
As a work around, we have set it to the default format YYYY-MM-DD and with that one it is working.
_(I am only commenting here because the other tasks are referencing this as a fix.)_

PR #6140, opened 2019-08-18
Fixed incorrect casing of filenames, both by renaming files where req…

jaddie:
Building on Linux is instantly failing due to the poor casing of file names in references and suchlike. This is the first set of changes related to this which will hopefully lead to a clean build in the future. 

This PR only contains small changes at the moment such as renaming files to have the expected casing or changing references to match the already existing files with uppercase names etc. Eventual intention being to make Umbraco work properly on Linux.


nul800sebastiaan:
Hey @jaddie - I'll have a look at this, but need to discuss at HQ. I remember the mess when we last tried to rename everything and I don't think it's feasible to do this. 

Please don't spend any more time on this as I fear we'll have to reject future PRs with these types of casing changes.

I am not sure what makes you think compilation on Linux will succeed? We know Umbraco can not run on Linux at the moment because we have hard dependencies on Windows-specific APIs. 

Note: we will work on a .NET Core version in the future (https://github.com/umbraco/rfcs/pull/6) - the goal not immediately being able to run on Linux but that would be a possibility in the future.

jaddie:
Hey @nul800sebastiaan 

Fair enough, let me know what the result is from it, I've spoken to @Shazwazza in the past primarily about supporting at the time, mono and now switching that to .net core instead etc so am already aware of it.

I've previously looking into compiling the code on Linux & from my vague memory of it, conversion wasn't actually as involved as it seemed. Yes there will be a lot of tweaks required but they won't be heavy just little changes etc.

The main thing is the implementations in the code which are unfortunately very hard coupled to Windows as you say which is bad anyway and should be re-written. An example would be replacing the AD backoffice code with LDAP etc.

I haven't renamed a lot, only the items which indeed are mismatched this mis-match would apply to Windows as well & isn't a problem currently only because Windows is run on a case-insensitive file system in practice.

Appreciate the link, I am keeping in eye on those discussions already as it is something I already would like to be got on with as its taken a lot longer than it should have to reach this point etc, also another main point is that I would very much so like to do anything to help that effort to make sure Umbraco beats Sitecore to being on .Net Core as they've already been transitioning for a while as it could set Umbraco apart & give it a much needed boost :D

nul800sebastiaan:
Thanks @jaddie - the conclusion is: we tried this and it was absolutely disastrous because of existing PR and checking out / in files with mis-matched casing. It just moved things into unpredictable places on GitHub, it messed with both local (windows) clones and with the build server and it is.. sadly something we don't want to repeat.

Let's focus on the future where we'll be working on .NET Core and there we get a chance to be consistent. Thanks for the offer, but for now we can't assign enough time to it to make this work correctly!

PR #6150, opened 2019-08-19
Fix #6148 - added data element to NestedContent - V7

### LINKED ISSUE #6148: NestedContent - Missing data-attribute on property editors.
### Bug summary
***
If someone has a better idea for a fix then I'm all ears.
***

NestedContent items are missing a data-attribute (data-element) which is added by default when not using NestedContent.

This is stopping a package I'm using from working properly when I'm using it with NestedContent since it relies on that data-attribute.

### Specifics
Currently using version 7.15.1 - I'm fairly certain that it's like this ever since NestedContent was incorporated into the core.

### Steps to reproduce
Create a NestedContent setup, create a property editor and inspect it. Create a normal tab, add a property editor and inspect it. Compare them (attached image)

### Expected result
data-element="property-alias"

### Actual result
Nothing.

nc-missing-attributes

bgunnarsson:
Related issue: [#6148 ](https://github.com/umbraco/Umbraco-CMS/issues/6148)

NestedContent items are missing a data-attribute that is added to property editors by default.

I added that data-attribute (data-property), tested it and everything works just the way it should work.

Here's a screenshot of the NestedContent item with the data-attribute added to it:

nc-not-missing-attribute

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.editor.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.editor.html
@@ -1,7 +1,7 @@
 ﻿<div class="umb-pane">
   <div ng-repeat="property in tab.properties" style="position: relative;">
 
-    <umb-property property="property" ng-class="{'umb-nested-content--not-supported': property.notSupported}" data-element="{{property.alias}}">
+    <umb-property property="property" ng-class="{'umb-nested-content--not-supported': property.notSupported}" data-element="property-{{property.alias}}">
       <umb-editor model="property"></umb-editor>
     </umb-property>
 
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.editor.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.editor.html
@@ -1,7 +1,7 @@
 ﻿<div class="umb-pane">
   <div ng-repeat="property in tab.properties" style="position: relative;">
 
-    <umb-property property="property" ng-class="{'umb-nested-content--not-supported': property.notSupported}">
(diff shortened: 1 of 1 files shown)


bgunnarsson:
I'm pretty baffled by that this is marked as a feature since it's clearly an oversight - The normal property editors have this and not the NestedContent ones which in turn breaks this package and other packages that rely on that data-attribute. A clear bug in my books since it's stopping me from utilizing NestedContent to it's fullest. 

This means that this bug is still present in v8 and anyone trying to utilize that data-attribute will run into the same issue as I am now.

Can anyone elaborate on why this is considered a feature and not a bug?

Thanks.

nul800sebastiaan:
Baffled, even! 😮

After some investigating, it was clear that this attribute has never existed in either the core implementation of NC nor the package before that.

So this new feature of a data element would be great to add in a future version. 😊 

From https://umbraco.com/blog/the-umbraco-issue-tracker-process/ : 

> - A bug is typically a defect; something that is not working and giving unexpected results.
> - A feature, on the other hand, would be defined as “not a bug”, most of the time adding something new to the product.

bgunnarsson:
Alright, fair enough.

I'll be happy if this is implemented into v8, I can handle v7 by simply editing the file myself and including that into my package.

nul800sebastiaan:
Aah I see what your fear is now! We advertised that we're not adding new features to v7 any more. 😊

Fair enough! I think we can bend the rules on this one a little bit, adding an html attribute is not exactly earth-shattering. Please note that I closed the v8 PR since I intended to merge this one and then merge that up to v8.

bgunnarsson:
You just warmed my heart @nul800sebastiaan <3 #H5YR :)

nul800sebastiaan:
Awesome thanks very much for the addition @bgunnarsson ! 

And congrats on your first contribution to Umbraco CMS, well done! ⭐️ 

You now have that shiny new Contributor badge on your profile on Our: https://our.umbraco.com/members/id:140828/ 🏅 🎉

PR #6151, opened 2019-08-19
V8: Restored indication of unpublished content in content picker display

### LINKED ISSUE #6143: Notify the editor if an unpublished node is picked in a content picker
Currently an editor will see a message at a node's content picker if the picked node is unavailable (deleted/recycle bin) - which is sweet!

However, there's no feedback if the node has been unpublished. That means editor A can open node X and think that everything is OK, even though editor B has unpublished node Y, which is picked on node X.

contentpicker

Providing a warning message when a picked node is unpublished would be awesome. :-)

AndyButland:
Fixes #6143.

Without this additional field, the greyed out display you see below on the second item isn't seen, so you can't distinguish published and unpublished items.

image


poornimanayar:
Hi @AndyButland ,

That looks like a really handy fix! We shall test this and get back to you.

Poornima

nul800sebastiaan:
Awesome! Thanks @AndyButland !

PR #6152, opened 2019-08-19
Fix #6148 - Added data-property attribute to NestedContent Items for v8

### LINKED ISSUE #6148: NestedContent - Missing data-attribute on property editors.

bgunnarsson:
Related issue - [#6148](https://github.com/umbraco/Umbraco-CMS/issues/6148) - That is for v7, this is for v8.

Same as [PR #6150](https://github.com/umbraco/Umbraco-CMS/pull/6150) except this is for v8.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.editor.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.editor.html
@@ -1,7 +1,7 @@
 ﻿<div class="umb-pane">
   <div ng-repeat="property in tab.properties" class="umb-nested-content-property-container">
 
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks, if we merge the v7 version we will already merge it up to v8 so a separate PR is not necessary. 👍

Issue #6097, opened 2019-08-12
Allow lang-files placed relative to package.manifest


soreng:
A brief description of your feature request goes here.

Umbraco 8 allows for manifest files to be placed "where ever" under app_plugins like App_Plugins/Plugin/Feature1/f1.manifest and App_Plugins/Plugin/Feature2/f2.manifest and so forth.

You could then build a plugin that can have multiple features.

Lang-files on the other hand can only live in one place -- App_Plugins/Plugin/Lang/*.xml

From our own part of the woods, we a building a lot of internal stuff that we reuse, and instead of having 20-30 Company.This.That and Company.Some.Thing folders, we would like to have:

`App_Plugins/Company/This/That` and `App_Plugins/Company/Some/Thing` - and have lang-files in subfolders, together with the respective manifest files

How can you help?
-------------------------------

https://github.com/umbraco/Umbraco-CMS/pull/4454 expanded the name of the lang files resolved. This could easily be changed to look for xml files in lang folder :

    appPlugins
        .GetDirectories("lang", SearchOption.AllDirectories)
        .SelectMany(x => x.GetFiles("*.xml", SearchOption.TopDirectoryOnly))

Another aspect of this could be to allow a "lang"-setting in the package manifest.

    {
        "lang": ["~/App_Plugins/Company/This/That/lang"],
        "propertyEditors": [],
        "gridEditors": [],
        "parameterEditors": [],
        "javascript": ["~/App_Plugins/Company/This/That/js/controllers.js"],
        "css": []
    }

What would people prefer? Expanding the searching to include subfolders, or implement lang-property in manifest?

nul800sebastiaan:
You can update files in `~/Config/Lang` - isn't that what you need? 

So the hierarchy is:

- `~/Umbraco/config/Lang`
- `~/Config/Lang` (overrules `~/Umbraco/config/Lang`)
- `~/App_Plugins/{myPlugin}/Lang` (overrules `~/Umbraco/config/Lang` and  `~/Config/Lang`)

I do understand the requirement you have here, but it seems like a bit of an edge case and technically you can already handle this in `~/Config/Lang` files.

soreng:
@nul800sebastiaan it's all about only shipping the needed files :)

I can easily create new lang/da-dk.xml files, but I can't easily edit the files in the places you list -- other than the "only root-folders" under App_Plugins

[...]

PR #6153, opened 2019-08-20
V8: Allow plugin "Lang" folders to reside in nested directories

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6097

### Description

This PR allows plugins to place their localization files anywhere in the plugin folder sub structure, as long as they reside within a folder called `Lang`.

To test this PR:

1. Unzip the attached plugin in `App_Plugins`
2. Log in with a user in `en-US` backoffice locale
3. Verify that the "Language test dashboard" is localized correctly. It'll use the localization files in `~/App_Plugins/LanguageTest/Some/Another/Lang/`, and it should look something like this:
image

The test plugin is here: 
[LanguageTest.zip](https://github.com/umbraco/Umbraco-CMS/files/3519659/LanguageTest.zip)

### DIFF
--- a/src/Umbraco.Core/Composing/CompositionExtensions/Services.cs
+++ b/src/Umbraco.Core/Composing/CompositionExtensions/Services.cs
@@ -96,7 +96,7 @@ namespace Umbraco.Core.Composing.CompositionExtensions
             var pluginLangFolders = appPlugins.Exists == false
                 ? Enumerable.Empty<LocalizedTextServiceSupplementaryFileSource>()
                 : appPlugins.GetDirectories()
-                    .SelectMany(x => x.GetDirectories("Lang"))
+                    .SelectMany(x => x.GetDirectories("Lang", SearchOption.AllDirectories))
                     .SelectMany(x => x.GetFiles("*.xml", SearchOption.TopDirectoryOnly))
                     .Select(x => new LocalizedTextServiceSupplementaryFileSource(x, false));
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
What happens if I have:

`~/App_Plugins/LanguageTest/Lang/en-US.xml` and `~/App_Plugins/LanguageTest/Some/Lang/en-US.xml` and `~/App_Plugins/LanguageTest/Some/Another/Lang/en-US.xml`? 

Will it take the first, will it take the last? How about overlapping entries?

kjac:
@nul800sebastiaan it'll grab all of them 😆 if for some reason you have overlapping localization entries across the individual lang files, it works as it always have: Last in wins.

nul800sebastiaan:
Sweet, sounds good then. And I assume that `SearchOption.AllDirectories` just walks the tree downwards, so the deepest you hide the lang files, that's the last one that comes in.

kjac:
That would be a fair assumption but I can't say for certain. Should be relatively easy to test though, if the need arises.

soreng:
That would be up to the implementer of the file system. To make sure we could just sort the folder-path

Livrem og seler :-)

nul800sebastiaan:
Thanks for the update and the feedback @kjac and @soreng - I think this makes sense! 👍

PR #6154, opened 2019-08-20
Improve error messages

jmayntzhusen:
### Prerequisites

### Description

When debugging deployments I ran into this error message, and I had to attach the Umbraco Core code to figure out which node had an issue being saved. 

Would be awesome to have more info on which specific content item it is that is failing for when you are looping through this call.

Not entirely sure how the site got into an error state where the content service couldn't save a node, so not sure on the steps to reproduce to get this error message. All this PR does is improve a bit on info in errors though.


nul800sebastiaan:
Thanks @jmayntzhusen !

nul800sebastiaan:
Cherry picked to 8.1.4 650663de0ca37dfad7baa22d3c6c2d9e18e9e8cb

PR #6167, opened 2019-08-21
Remove comparison in ApplicationUrlHelper.GetApplicationUrlFromCurrentRequest that is always false

smcl:
### Prerequisites

**Note**: I've left the above unchecked, I do not believe a new test is appropriate for this change

### Description

When deciding whether to use SSL, ApplicationUrlHelper.GetApplicationUrlFromCurrentRequest() has an expression `port == "443"` which is always `false`. The variable `port` is set to to either `""` or `":" + request.ServerVariables["SERVER_PORT"]`, so it might sometimes be ":443" but it will _never_ be just "443". 

This code has been this way for at least a year, and I don't know if the situation it tried to handle makes any sense (using port 443 over plain HTTP, for some reason??) so I've just removed it.

### DIFF
--- a/src/Umbraco.Core/Sync/ApplicationUrlHelper.cs
+++ b/src/Umbraco.Core/Sync/ApplicationUrlHelper.cs
@@ -97,8 +97,7 @@ namespace Umbraco.Core.Sync
                 ? ":" + request.ServerVariables["SERVER_PORT"]
                 : "";
 
-            var useSsl = globalSettings.UseHttps || port == "443";
-            var ssl = useSsl ? "s" : ""; // force, whatever the first request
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks @smcl - indeed, that makes no sense and it doesn't really make sense to  test for port number anyway, this is great! 👍 

And congrats on your first contribution to Umbraco CMS, well done! 🎉

I would love to give you a contributor badge on Our Umbraco if you have a profile there? Just link it here and I'll get it sorted! 🏅

nul800sebastiaan:
Cherry picked for 8.1.4 0dc840a4ed8668a8c8df17aa1a157f795e003d8f

Issue #6155, opened 2019-08-20
Duplicate Content Templates causes list to disappear.


NikRimington:
**Not tested in v8, but believe to be an issue there as well**

In Umbraco v7, by chance, multiple content templates were created with the same name (unsure how this occurred). However, the consequence of this was that a `Duplicates in repeater are not allowed` error is thrown. Looking at the v8 code this is still going to occur.

**To fix:**

Add ` track by $index` to ng-repeats here:
https://github.com/umbraco/Umbraco-CMS/blob/e4f19f71926ff1c0db427d6604798795fbe88f8c/src/Umbraco.Web.UI.Client/src/views/propertyeditors/listview/listview.html#L30

and here:
https://github.com/umbraco/Umbraco-CMS/blob/57e58c6cf75112c738ca08e9398dcc77915b64fb/src/Umbraco.Web.UI.Client/src/views/content/create.html#L45

kjac:
In V8 the UI protects us from creating content templates with duplicate names - not sure that's the case in V7?

![image](https://user-images.githubusercontent.com/7405322/63373478-4c06e680-c388-11e9-8824-b28d941b87cf.png)

![image](https://user-images.githubusercontent.com/7405322/63373538-68a31e80-c388-11e9-8394-c4f756c6f5f0.png)

nul800sebastiaan:
It might do in the interface, but I'm pretty sure, with a bit of forceful C# coding you could create duplicates anyway. I would imagine, for example, this could happen in an edge case with uSync.

The fix proposed is good @NikRimington - care to make it "official"? 😊

umbrabot:
Hi @NikRimington,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

NikRimington:
PR's submitted, one for v7 and one for v8 😃

nul800sebastiaan:
Cherry picked for 7.15.3 55d0ca25571f819e51caee1242483d8f9d511551
Cherry picked for 8.1.4 b396cc340cfdf0b1ea90ab3729a35758f435642c

PR #6170, opened 2019-08-22
Added track by $index to ng-repeats for listing blueprints

NikRimington:
Fix for https://github.com/umbraco/Umbraco-CMS/issues/6155 
Adds tracking by index to ng-repeat when listing available blueprints.


nul800sebastiaan:
No need to provide a PR for both versions, we always merge up from v7 to v8. I'll close this one.

Issue #5983, opened 2019-07-23
MNTP does not allow setting media type, or member type when you change to a matching node type of Media or Member in v8.1


tom-ingeniuus:
Using version 8.1 in the latest Chrome browser

I want to create a MNTP picker to select only a particular media type which I have created. When you change the node type to media and then try to select the allowed media types, the selector still shows document types. It should show media types. The same issue occurs if you select a member type nodes, you still get a list of document types, not member types.

Interestingly, when selecting the root node for media, it does display media folders. The root node correctly gets hidden when you select a member type node.

nul800sebastiaan:
Ah shoot.. we forgot to test this with media / member types @kjac ! The cause is in this PR: https://github.com/umbraco/Umbraco-CMS/pull/5506

kjac:
Shoot. I'll have a look.

kjac:
So I've been digging into this, and it's interesting to say the least 🤔 

Obviously it's an error that you get to select content types when you're configuring an MNTP for members or media. That needs to be fixed one way or another; I'll get back to that later. First to the interesting parts.

#### Media picker

The new V8 media picker implementation does not support the filtering function required by MNTP to dim non-allowed types. This used to work in V7 with the V7 style media picker. See the [media picker controller](https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js) - there is no usage of or reference to `$scope.model.filter`, which is where the media type filter is defined (it is however passed into the model as expected).

#### Member picker

The member picker doesn't seem to have been supporting filtering for a long, long time. Here's a GIF from V7.15; the "Member picker" property is an MNTP configured to select only members of type "Other":

![member-picker-with-filter](https://user-images.githubusercontent.com/7405322/61742225-e9bfc380-ad92-11e9-98ec-8a1222c27e98.gif)

I have tracked the lack of filtering down to [this bit of code](https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/common/directives/components/umbminilistview.directive.js#L81). The member picker filtering is "advanced" (meaning it's a custom function by the contentpicker controller [here](https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/propertyeditors/contentpicker/contentpicker.controller.js#L186)).

[...]

PR #5506, opened 2019-05-22
V8: Use a picker to select allowed types for MNTP

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description

The current configuration of allowed types for the MNTP isn't really very nice, nor very much in line with the rest of the config:

image

This PR swaps the comma separated list of aliases for a picker style configuration:

mntp-pick-allowed-types

The underlying data model remains the same. In the future we should probably consider swapping it for a list of UDI's, but that's beyond the scope of this PR.

#### Testing this PR

1. Create an MNTP and verify that you can configure its filter using the new picker.
2. Add the MNTP to a content type and verify that the filter works when using the MNTP.

PR #6177, opened 2019-08-22
V8: Allow MNTP to pick "allowed types" for media and member pickers

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5983

### Description

This PR amends #5506 so it becomes possible to pick "allowed types" for media and member MNTP pickers. It works like this:

mntp-allowed-types-media-member

### Testing this PR

1. Make sure you can pick "allowed types" of the selected node type.
2. When changing node type, verify that any previously selected "allowed types" are removed.

### ...but the resulting datatype is broken?

As it turns out, the MNTP property editor doesn't support "allowed types" for media and member ([more info here](https://github.com/umbraco/Umbraco-CMS/issues/5983#issuecomment-514360199)). If this PR is merged in, two new issues should be created to ensure that the MNTP property editor actually works with the configured "allowed types" for media and member pickers.


nul800sebastiaan:
Thanks very much @kjac! 

Only thing to fix now is member and media type restrictions but so far so good! 👍

Issue #6172, opened 2019-08-22
v8 - Link Picker errors upon re-opening when only a Query String / Anchor is provided.


NikRimington:
When only adding a Query String or Anchor value to the Multi-Url Picker, if you go back in to edit the value it will throw an exception and not load the editor. This only seems to occur if you've not saved the page you are editing, however saving it seems to allow the editor to work.

I've been unable to identify where the error is being thrown but it relates to the URL being empty.

Umbraco version: 8.1.2

nul800sebastiaan:
Hmmm, pinging @kjac - he did fix this but I guess not when you haven't saved the page? I guess it can't find the anchor anywhere because the page does not exist, so we need to just ignore that we can't find it.. or something? 🤔

kjac:
I have been pinged 😂 will dive into it.

kjac:
Fixed in #6178

nul800sebastiaan:
Cherry picked to 8.1.4 a2899b21b66b1bca9be193a03644596d6c94dc90

PR #6178, opened 2019-08-22
V8: Don't crash the Multi URL Picker if only an anchor has been entered

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6172

### Description

If you add a link with only an anchor or a query string value in a Multi URL Picker property and then go to edit the link again *without saving the content first*, the Multi URL Picker editor crashes with a JS error:

url-picker-anchor-only-before

This PR fixes it:

url-picker-anchor-only-after


nul800sebastiaan:
Awesome mate! That was easier than I expected! Thanks a bunch for looking into it! 👍

Issue #6176, opened 2019-08-22
Search state not persisted when using the back link


bourpie:
When searching in list view, and then you go to a document, the search state is not persisted when returning to list view using the back link.

### Specifics

Umbraco version 8.1.1

### Steps to reproduce
1. Initial state
![search1](https://user-images.githubusercontent.com/4865432/63528463-fbe76b80-c4d0-11e9-983b-df9ebddacdd1.gif)
2. Search and go to document
![search2](https://user-images.githubusercontent.com/4865432/63528502-128dc280-c4d1-11e9-81f7-9c1f61ecf4f8.gif)
3. Go back to list view
![search3](https://user-images.githubusercontent.com/4865432/63528578-32bd8180-c4d1-11e9-81a0-b398fe2b86f1.gif)
4. Search state is not persisted (query string, search result and pagination)
![search4](https://user-images.githubusercontent.com/4865432/63528682-5c76a880-c4d1-11e9-8642-f6d1640819d8.gif)

### Expected result

Search state should persiste (query string, search result and pagination) after returning to list view when using back link

umbrabot:
Hi @bourpie,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

kjac:
I can reproduce this and #6175. Took a bit of detective work, but I think I found it... Will report back later.

kjac:
PR in #6189

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6189

bourpie:
You guys rock!!!

PR #6189, opened 2019-08-24
V8: Include sort and filter in list view back link

### LINKED ISSUE #6175: Sort state not persisted when using the back link
When a listview is sorted, and you go to a document, the back link doesn't persist the sorted state

### Specifics

Umbraco version 8.1.1

### Steps to reproduce

sort1
sort2
sort3
sort4

### Expected result

Sort state should be persisted after back link

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #6175 and #6176 

### Description

Sort and filter (search) state is not applied when going back to a list view from a list view item. See #6175 and #6176 for details on the issue.

When this PR is applied, sorting and filtering will be re-applied when the editor goes back to the list view:

list-view-sort-filter-after

While fixing this I uncovered another issue which this PR also implicitly handles. When you navigate to a child of a list view, and said child also is list view enabled, the pagination of the parent list view is applied to the child list view:

list-view-pagination-inheritance-before

list-view-pagination-inheritance-after

#### Testing notes

When testing multiple levels of list views, remember that the "Back to list view" functionality only remembers the state of the parent. If you work your way downwards in a structure using list views, the current parent list view state will always overwrite the previous one. This may be a bug but it's definitively beyond the scope of this PR to fix it 😄

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
@@ -62,11 +62,13 @@
         function init() {
 
             var content = $scope.content;
-
             if (content.id && content.isChildOfListView && content.trashed === false) {
-                $scope.page.listViewPath = ($routeParams.page) ?
-                    "/content/content/edit/" + content.parentId + "?page=" + $routeParams.page :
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Sweet work @kjac ! 👍 😮

PR #6190, opened 2019-08-25
V8: Make sure the empty search result is visible for list views

kjac:
### Prerequisites

### Description

If no results are found when searching for items in list views, the label that should tell you so is rather hidden behind the sticky header bar:

image

With this PR applied, the label displays as expected:

image

The offending line of CSS was added in d17022d8 to solve something sticky header bar related. However, I haven't been able to find any sticky header bar related issues when removing the CSS, neither in the grid nor in list views.


nul800sebastiaan:
Thanks @kjac !

Cherry picked for 8.1.4 529792227cdfee52a4b092b301ccadb18d339422

PR #6192, opened 2019-08-25
V8 - missing media folders after migration

### LINKED ISSUE #6158: V8 Migrated media folders not showing up in Media Picker or on the left side of the Media tab
On all our sites after we migrate to 8 the media folders don't show up in the Media Picker or the left side of the media tab. Example below. On some of our smaller sites I have just recreated the folders and copied the media to the new folders but on some of our bigger sites that would take hours.
umbraco media

stevemegson:
### Prerequisites
If there's an existing issue for this PR then this fixes #6158

### Description
The upgrade to 7.8.0 would not create rows in `cmsMedia` for folders (or any media type without an `umbracoFile` property). From 7.15 the media tree used an inner join to `cmsMedia`, so folders created before the upgrade to 7.8.0 disappeared.

The problem was fixed for 7.15 in PR #5865., but v8 has a similar issue with those media folders (except that the table is now `umbracoMediaVersion`). This PR applies the equivalent fix for v8.

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/EntityRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/EntityRepository.cs
@@ -404,7 +404,7 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
             if (isMedia)
             {
                 sql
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Nice one @stevemegson ! 👍

Issue #5847, opened 2019-07-11
Media thumbnails not shown after 7.14.0 -> 7.15.0 upgrade


simonhawes:
It appears in v7.15.0 the ImageProcessor now uses cmsMedia.mediaPath to display thumbnails.  In many cases this field is NULL when images have been imported into Umbraco in a previous version.

This is characterized by angular reporting "Cannot resolve the file url from the mediaEntity, it does not contain the required metaData" in the developer tools console window.

In order to make the thumbnails appear correctly I had to write some code to update the mediaPath column from the cmsPropertyData table.  The data exists in either the dataNvarchar or within a json string in the dataNtext.

This should be taken care of as part of the upgrade process ?

---
_This item has been added to our backlog [AB#7181](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7181)_

nul800sebastiaan:
Can you show what you had to update? And where do you see errors/problems? Not sure why cmsMedia has nulls unless it's a media item without a filled in `umbracoFile` property.

Shazwazza:
I think the key to understanding this is this:

> In many cases this field is NULL when images have been imported into Umbraco in a previous version.

Which sounds like potentially very old versions of umbraco before there was a cmsMedia table that tracked media paths?

simonhawes:
Here is the SQL I used (I had to code it up in C# in order to process the json variants, but this is effectively what I did)

UPDATE [cmsMedia] SET mediaPath = cmsPropertyData.dataNvarchar
FROM
(
              select contentNodeId, dataNvarchar from cmsPropertyData
)cmsPropertyData

WHERE mediaPath IS NULL AND cmsMedia.nodeId = cmsPropertyData.contentNodeId

The site was new at 7.14.0 so it is not that it was an old version.  If you edit the thumbnail and save the node, the mediaPath is then added to the database correctly, but with thousands of images obviously this is not practical.

The document type contains a gallery property of type Umbraco.MediaPicker2 with “Pick Multiple Items” selected.

[...]

PR #6195, opened 2019-08-26
V7 - populate mediaPath when property alias is not umbracoFile

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes: #5847

### Description
In 7.15, the `mediaPath` property provides the thumbnail for images and determines whether an item is an image or a folder. It is populated assuming that the media is stored in an `umbracoFile` property and the path starts with `/media/`.

This PR implements the behaviour from 7.14, where any upload or image cropper property could provide the thumbnail. It includes a migration to deal with any existing NULL values.

Testing:
- In 7.15, create a media type with an upload or image cropper property. Set the alias to anything but `umbracoFile`.
- Create an item of this type, and confirm that the thumbnail is displayed correctly in a media picker.


stevemegson:
Updated to handle items which have an upload or image cropper property but have no file selected, as seen in #6441.

PR #6201, opened 2019-08-26
V8: Indexing of grid values containing JSON

### LINKED ISSUE #6164: If Grid Module contains object in value, then Grid Layout property is not in Examine Index
We using a lot of custom Grid Modules in Umbravo v8.1.2. But the Property of "Grid Layout" is not every time in the examine index.

It seems, that the grid layout property is always not indexing, if $scope.control.value contains an object instead a string.

At first I add a custom Grid module with an object in value:
screen1

In Examine Dashboard the grid property is missing:
screen2

At second, I add a custom Grid module with a simple string in value:
screen3

Now, the complete Grid Layout property is in Examine Index:
screen4

This is the Grid Module "Anchor" with a string in value:
screen5

This is the declaration in the "Headline/Text" Controller with an object in value:
screen6

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #6164

### Description
If a grid control stores a JSON value rather than a string, the value is not indexed by Examine. This PR indexes any string values found within the JSON.

Testing:
- Add a macro control to a grid, with a unique value in a macro parameter.
- Search for the unique value in the Examine dashboard, the document should be found.

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/GridPropertyIndexValueFactory.cs
+++ b/src/Umbraco.Web/PropertyEditors/GridPropertyIndexValueFactory.cs
@@ -41,9 +41,7 @@ namespace Umbraco.Web.PropertyEditors
                         {
                             var controlVal = control.Value;
 
-                            // TODO: If it's not a string, then it's a json formatted value -
-                            // we cannot really index this in a smart way since it could be 'anything'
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks very much @stevemegson - this makes a lot of sense to me!

Issue #6023, opened 2019-07-30
Can't highlight username on login screen


PeteDuncanson:
Type in a username on the login screen and then try to select the text with the mouse. You simply can't. Only way to clear the text box is put the cursor in the box at the end and delete. Very odd!

Frustrating when trying to login with different credentials to the ones Chrome has remembered. Seems to only be a Chrome issue, tried it in Edge v44 and IE11.

Doesn't appear to be a CSS thing (I tried disabling some styles in dev tools). I think there is an event that is firing which is stopping it. If you wiggle the mouse up and down while going side ways sort of in a zigzag then it does end up highlighting it just but its painful!

![image](https://user-images.githubusercontent.com/1674590/62137597-2123f800-b2de-11e9-8154-7d0c3366f123.png)

![image](https://user-images.githubusercontent.com/1674590/62137850-8a0b7000-b2de-11e9-9339-33bbd2af131d.png)

Can't see this getting to the top of anyones hit list of "hot fixies in v8.2!" but if someone likes a challenge then this might be a fun one to check out.

nul800sebastiaan:
Hmm, I'm on Chrome 77 already so maybe it's a Chrome bug but I can't repro. Have you checked in incognito mode too without any add-ons activated?

![6023](https://user-images.githubusercontent.com/304656/62347497-370e0480-b4fa-11e9-9c05-17f259393a31.gif)

stevemegson:
I can reproduce it in v7, but not v8. It looks like ng-swipe-right is grabbing mousemove events on the form.

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6204

PeteDuncanson:
Woot! Good debug skillz @stevemegson :)

PR #6204, opened 2019-08-26
V7: Selecting text in context menu closes the dialog

### LINKED ISSUE #4934: Selecting text in context menu closes the dialog
When you select text in the context menu (do a swipe left with your mouse) the dialog closes. It is especially annoying when using Umbraco Cloud, and you get a deploy error and want to copy the error message. The dialog then closes, and the error message is gone, and you have to wait for another deploy/restore.

The "functionality" is caused by the ng-swipe-left directive on the dialog. I think the annoyance from the feature overweighs the benefits of being able to swipe left on mobile to close the dialog and the directive should just be removed.

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #4934, #6023

### Description
V7 version of PR #6203.

Additionally, stops swipe events triggering from mouse events. This can also trigger unexpected dialog closes, and can make it impossible to select text when mousemove events are intercepted.

Testing:
- As for PR #6203
- Also, check that you can select text on the login screen.

### DIFF
--- a/src/Umbraco.Web.UI.Client/lib/angular/1.1.5/angular-mobile.js
+++ b/src/Umbraco.Web.UI.Client/lib/angular/1.1.5/angular-mobile.js
@@ -293,7 +293,7 @@ ngMobile.directive('ngClick', ['$parse', '$timeout', '$rootElement',
  * A leftward swipe is a quick, right-to-left slide of the finger.
  * Though ngSwipeLeft is designed for touch-based devices, it will work with a mouse click and drag too.
  *
- *                      [well, it did until we disabled that feature to avoid problems with selecting text]
+ *                  [well, it did until we disabled that feature to avoid problems with selecting text]
  *
  * @element ANY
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Brilliant @stevemegson! This is going to fix sooo many annoyances. The one thing it unfortunately doesn't seem to fix for me is when you drag outside of a dialog and release the mouse button, then it still closes the dialog. But all the other text selection events in many areas are now behaving properly. Great stuff!

marcemarc:
This is absolutely the bain of my life... thanks @stevemegson

nul800sebastiaan:
Note: I had to revert this commit: https://github.com/umbraco/Umbraco-CMS/pull/6204/commits/08480d61038cc59c87e79bc640b94c9b7f5c5b57

It would prevent the dialogs from opening from the RTE, so the embed and insert media buttons would only sort-of work if you handily double clicked them.

Issue #6202, opened 2019-08-26
MediaFactory - MediaPathPattern hardcoded to media


codyb11989:
I changed the umbracoMediaPath config variable and the mediaPath column in cmsMedia table is always NULL.

Reproduction
------------

I changed the default media folder. Adding new media gives me NULL value in cmsMedia table column mediaPath

#### MediaFactory.cs - Regex MediaPathPattern
 - Doesn't take the config variable into account for the Regex

### Specifics

v.7.15.1

`private static readonly Regex MediaPathPattern = new Regex(@"(/media/.+?)(?:['""]|$)", RegexOptions.Compiled);`

![image](https://user-images.githubusercontent.com/20603773/63700059-0450e680-c7f0-11e9-9b01-427b08fd1136.png)

### Steps to reproduce

Change default media path to anything else

### Expected result

mediaPath added to cmsMedia table

### Actual result

NULL mediaPath in cmsMedia table

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6205
Cherry picked for 7.15.3 f5a445101d88c632965c17d1e8b0b68e181696ed

PR #6205, opened 2019-08-26
use SystemDirectories.Media instead of hardcoded string media

codyb11989:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6202

### Description

Replaced hard-coded media string and used the SystemDirectories.Media path.

### DIFF
--- a/src/Umbraco.Core/Persistence/Factories/MediaFactory.cs
+++ b/src/Umbraco.Core/Persistence/Factories/MediaFactory.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Globalization;
 using System.Text.RegularExpressions;
+using Umbraco.Core.IO;
 using Umbraco.Core.Models;
 using Umbraco.Core.Models.Rdbms;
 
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Super! Thanks very much @codyb11989 👍 

And congrats on your first contribution to Umbraco CMS, well done! 🎉 

You are now the owner of a shiny new contributor badge on Our Umbraco, congrats! https://our.umbraco.com/members/id:278931/ 🥇

LouisHol4net:
@nul800sebastiaan  I'm facing this problem in the latest v8 version, will this be patched anytime soon  in 8?

skidmow:
FYI this has introduced a breaking change. By introducing a `$` to the beginning of the Regex, the media path will no longer respect virtual directories. For example if Umbraco runs as a virtual application in IIS, rather than as a website, the media will no longer be found and will 404 error.

**Replication scenario:**
The FileSystemProviders.config has this entry:
<add key="virtualRoot" value="~/media/" />

The website runs in a Virtual Directory named `cms`
Therefore all Umbraco pages and media are served with a `/cms/` prefix.

Since this code change, media paths are no longer being served from `/cms/media/`, but instead `/media/`, which doesn't exist and will therefore 404 error.

We cannot resolve this issue by changing to the value below, as that results in an error stating that media cannot live outside of the application root.
<add key="virtualRoot" value="~/cms/media/" />

I will raise this as a new issue now.

PR #6208, opened 2019-08-27
Support for Segments in ContentTypeRepositoryBase

### LINKED ISSUE #6174: Saving ContentType with Segment variation throws NotSupportedException
### Bug summary
When saving a ContentType with Variations set to ContentVariation.Segment, a NotSupportException is thrown. This also happens for PropertyType Variations.

### Steps to reproduce
Execute the following code in the Starter Kit. 

```csharp
var services = global::Umbraco.Core.Composing.Current.Services;
// Or inject...
IContentService contentService = services.ContentService;
IContentTypeService contentTypeService = services.ContentTypeService;

{
    // Homepage of Starter Kit
    IContent homepage = contentService.GetById(1104);

    IContentType contentType = contentTypeService.Get(homepage.ContentTypeId);

    // Set contentType as varied by segment
    contentType.Variations = ContentVariation.Segment;

    foreach (var propertyType in contentType.PropertyTypes)
    {
        // Set propertyType as varied by segment
        propertyType.Variations = ContentVariation.Segment;
    }

    // Throws NotSupportedException
    contentTypeService.Save(contentType);

    scope.Complete();
}
```
### Expected result
ContentType is saved successfully and the database column "variations" is updated correctly in the cmsContentType table. Similarly the "variations" columns of the relevant cmsPropertyType rows are updated.

### Actual result
A NotSupportException is thrown.

Note: setting the variations columns of cmsContentType and cmsPropertyType can be done manually with the following SQL statements:

```sql
-- Enable Segment variations for Home content type and heroHeader property in Starter Kit.

UPDATE cmsContentType
SET variations = 2
WHERE pk = 541

UPDATE cmsPropertyType
SET variations = 2
WHERE alias = 'heroHeader'
```

These exceptions originate from these source files:
* https://github.com/umbraco/Umbraco-CMS/blob/release/8.1.3/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs#L767
* https://github.com/umbraco/Umbraco-CMS/blob/release/8.1.3/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs#L684

PerplexDaniel:
### Prerequisites

Work in progress to fix #6174 

This PR was made after a discussion with HQ (Jacob + Claus) about the state of segments.

### Description
Adding `ContentVariation.Segment` to a ContentType / PropertyType currently throws an exception, which is fixed in this PR. 
The code below (mostly the same as in #6174) can now be run without the exception, and the appropriate "variations"-column in the database tables cmsContentType / cmsPropertyType is updated.

As mentioned, this is a work in progress PR mainly submitted already so you guys can also have a quick look and perhaps run a few tests if this breaks anything.

```csharp
// Or inject...
IContentService contentService = Current.Services.ContentService;
IContentTypeService contentTypeService = Current.Services.ContentTypeService;

{
    // Homepage of Starter Kit
    IContent homepage = contentService.GetById(1104);

    IContentType contentType = contentTypeService.Get(homepage.ContentTypeId);

    // Add Segment variation type
    contentType.Variations |= ContentVariation.Segment;

    foreach (var propertyType in contentType.PropertyTypes)
    {
        // Add Segment variation type
        propertyType.Variations |= ContentVariation.Segment;
    }

    contentTypeService.Save(contentType);

    scope.Complete();
}
```


src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
```diff
@@ -527,20 +505,18 @@ protected void PersistUpdatedBaseContentType(IContentTypeComposition entity)
...
         {
-            //if the entity does not vary at all, then the property cannot have a variance value greater than it
-            if (entity.Variations == ContentVariation.Nothing)
+            foreach (var prop in entity.PropertyTypes)
             {
-                foreach (var prop in entity.PropertyTypes)
-                {
-                    if (prop.IsPropertyDirty(nameof(prop.Variations)) && prop.Variations > entity.Variations)
-                        throw new InvalidOperationException($"The property {prop.Alias} cannot have variations of {prop.Variations} with the content type variations of {entity.Variations}");
-                }
```
PerplexDaniel: This validate seemed incomplete due to the `ContentVariation.Nothing` and the `prop.IsPropertyDirty()` check, it did not match with the description of the method. For example, it would allow a property type to have ContentVariation.Culture when the content type was set to ContentVariation.Nothing, as long as the property type was not dirty. This does not seem to make much sense, it should not be allowed at all regardless of the dirty state.

It also allowed various other invalid combinations, like content type of Culture and property type of Segment.

src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
```diff
@@ -693,78 +669,72 @@ private void MoveContentTypeVariantData(IContentTypeComposition contentType, Con
...
+                sqlInsert = Sql($"INSERT INTO {DocumentCultureVariationDto.TableName} ({cols})").Append(sqlSelect);
+
+                Database.Execute(sqlInsert);
+            }
+            else
+            {
+                //we don't need to move the names! this is because we always keep the invariant names with the name of the default language.
+
+                //however, if we were to move names, we could do this: BUT this doesn't work with SQLCE, for that we'd have to update row by row :(
+                // if we want these SQL statements back, look into GIT history
```
PerplexDaniel: Put back original comments

src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepository.cs
```diff
@@ -287,6 +287,16 @@ protected override void PersistUpdatedItem(IContentType entity)
...
+            // Update property variations based on the content type variation
+            foreach (var propertyType in entity.PropertyTypes)
+            {
+                // Determine variation for the property type.
+                // The property is only considered culture variant when the base content type is also culture variant.
+                // The property is only considered segment variant when the base content type is also segment variant.
+                // Example: Culture variant content type with a Culture+Segment variant property type will become ContentVariation.Culture
+                propertyType.Variations = entity.Variations & propertyType.Variations;
+            }
+
```
PerplexDaniel: This code auto corrects property type variations based on their content type variations. It was originally done *after* calling the validation method which we corrected [here](https://github.com/umbraco/Umbraco-CMS/pull/6208/files#diff-d367ba2cc01d9cb6029082d5bae70f1fR511-R517). However, unit tests were now failing. Apparently there is a hidden "feature" that property type variations were auto corrected to a valid variation based on their content type. The auto correction was done after the validation (??) which caused the unit tests to now fail. In order to fix the unit tests we had to perform the auto correct before the validation. 

It might seem like it's a big chance but the functionality should be preserved and now the validation method now covers all inputs.

PerplexDaniel:
This is all quite complex, perhaps we should arrange another call about this to discuss, it might be easier. Below I try to describe what we have done.

We've (me and @PerplexWouter) started adding test cases for content type service calls. Looking at the existing tests, we noticed many test methods did not systematically cover all possible permutations of inputs and each test method seemed to test a bunch of different things at the same time. For example, the test `Change_Content_Type_From_Invariant_Variant()` was testing:
- Node name (`doc.GetCultureName() / doc.SetCultureName()`)
- Property value (`doc.GetValue(...) / doc.SetValue(...)`)
- Edited state (`doc.Edited`)
- Culture edited state (`doc.IsCultureEdited()`)

The test was testing all that while changing the content type variations back and forth between ContentVariation.Nothing and ContentVariation.Culture. The test seemed very specific for this scenario but we would prefer some more focused tests that can easily tested for all input combinations. 

A method that we could fairly easily extend was `Change_Content_Type_Variation_Clears_Redirects`, which simply tested one thing: does changing the content type variation clear all redirects. We made a few small modifications and then used the `[TestCase]` attribute to simply let it validate all possible inputs.

We think the unit tests should probably be split up into focused assertions, testing a single piece of functionality. We have added a few methods like that (such as `Verify_If_Property_Type_Variation_Is_Correctly_Corrected_When_Content_Type_Is_Updated` and `Preserve_Content_Name_After_Content_Type_Variation_Change`), but splitting up all the existing tests will take too much time for us, especially because we are not aware of all the business rules that apply to changing content types. One such rule appears to be the "auto correct" of property type variations that we discussed [here](https://github.com/umbraco/Umbraco-CMS/pull/6208#discussion_r331488572). We could not find an explicit test for that behavior either but we happened to run into it. 

We like to hear your thoughts about the changes.

src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepository.cs
Shazwazza: Interesting. I suppose however that this auto-correct code needs to exist on the `ContentTypeRepositoryBase` since one day we will probably have media types allowing variance as well. If we are auto-correcting, is there any point in validating? This would seem to always be true in that case. 

It turns out I added that validation check as part of this PR https://github.com/umbraco/Umbraco-CMS/pull/6019 and in this commit: https://github.com/umbraco/Umbraco-CMS/commit/9538981730d71698e90a00aba92fea700b587eef

... but i honestly don't recall all of the details of why but it was probably more of a safety check since this was quite a difficult bug to fix. It seems to me though, if you are changing the variance of a doc type, the of course all property types will need to be auto-adjusted and there doesn't seem to be a need for validation?

src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
Shazwazza: Yep, see my last comment, maybe this is actually unecessary anyways if we are auto adjusting?

Shazwazza:
Hi @PerplexDaniel have left a couple more comments. Great that you have added some more tests and test cases. I agree that the current tests do test some specific scenarios, i mentioned this one already but a few tests were updated to support the fixes for this https://github.com/umbraco/Umbraco-CMS/pull/6019/files 

I 'think' that these tests could be updated to use test cases:

* Change_Property_Type_From_Invariant_Variant
* Change_Property_Type_From_Variant_Invariant
* Change_Property_Type_From_Variant_Invariant_On_A_Composition
* Change_Content_Type_From_Variant_Invariant_On_A_Composition
* Change_Property_Variations_From_Variant_To_Invariant_And_Ensure_Edited_Values_Are_Renormalized
* Change_Property_Variations_From_Invariant_To_Variant_And_Ensure_Edited_Values_Are_Renormalized

These are effectively integration tests and while I agree that having many more focused tests proving smaller tests cases, it would result in a huge amount of tests when most things can be tested in a single integration test with more assertions (which i agree there should be more of). I don't think any of us have time to write out every single test case that there could be and then actually write those tests, for now i think it makes more sense to assert what we have make sure it all works for the different variations and that it works for your guys. In the future we'll hopefully add some test coverage results too.

src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
```diff
@@ -219,6 +219,7 @@ protected void PersistNewBaseContentType(IContentTypeComposition entity)
 
         protected void PersistUpdatedBaseContentType(IContentTypeComposition entity)
         {
+            CorrectPropertyTypeVariations(entity);
```
PerplexDaniel: As requested I moved this behavior back to ContentTypeRepositoryBase.
It is done before validation because the validation method essentially checks what is done in the correction method. As you mentioned @Shazwazza this makes the validation method essentially obsolete but does not really hurt either.

Note this "auto correct" behavior is not done (yet) in `PersistNewBaseContentType()` but only in the update method. This is how it was before any of our changes so I didn't add the "auto correct" code there, but we could add it there as well.

PerplexDaniel:
I have updated the tests you described to now include the `ContentVariation.Segment` flag as well in all tests. That is, where before "Invariant to Variant" would mean only `Nothing` -> `Culture` it now means the following:
* `Nothing` -> `Culture`
* `Nothing` -> `CultureAndSegment`
* `Segment` -> `Culture`
* `Segment` -> `CultureAndSegment`

And vice versa for "Variant to Invariant".

This is achieved by adding parameters to the original test method and using `[TestCase]` attributes to run through the 4 combinations.

Tests still pass so seems OK. I didn't actually set any segment values on the properties and asserted their values later, all of that could probably be added as well.

src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
Shazwazza: All good (for now), maybe it could be added to the `PersistNewBaseContentType` but lets keep changes to a minimum for now.

Issue #6180, opened 2019-08-23
Content properties are empty when current segment is null on segmented content type


PerplexDaniel:
### Summary
When a content type is set to ContentVariation.Segment, property values of the IPublishedContent are all empty when the current segment is null. In this case it is possible to read segmented property values, using `@Model.Value("heroHeader", segment: "s1")`. It appears when ContentVariation.Segment is set, you can *only* read segment values, and not the unsegmented values, i.e. 
`@Model.Value("heroHeader", segment: null)`.

We would have expected it to be possible to read both non-segmented data as well as the segmented data. It appears this is not currently supported by Umbraco, segmented data and non-segmented data appear to be mutually exclusive by design. 

### Steps to reproduce
Using the Starter Kit, enable segment variation for the Home content type and the "heroHeader" property type (see #6179). 

Refresh content cache  (Settings > Published Status > Rebuild Database Cache + Reload Memory Cache).

In a the Views/home.cshtml file, put this code:
```cshtml
@Model.Value("heroHeader", segment: null)
```

### Expected result
The value "Umbraco Demo" will be shown.

### Actual result
Nothing is shown, the property value is empty.

Shazwazza:
Fixed with https://github.com/umbraco/Umbraco-CMS/pull/6209

PR #6209, opened 2019-08-27
Fixes writing / reading segment data combined with segment null

### LINKED ISSUE #6179: SetValue without segment on segmented content type fails
### Bug summary
Trying to call SetValue on an IContent based on a content type with its "Variations"-property set to `ContentVariation.Segment` without providing a value for the segment parameter results in Umbraco not saving the property value and throwing a NotSupportedException:
`Variation "<null>,<null>" is not supported by the property type`. 

When Variations is set to `ContentVariation.CultureAndSegment`, a similar exception occurs when trying to set a value for a culture (e.g. "en-US") but leaving the segment as null:
`Variation "en-US,<null>" is not supported by the property type.`

However, when working with segments we would like to also save the "original" property values which do not belong to any segment. In our use case, segments would be used as alternative property values, to be used in specific cases. For example, we could create some segment that is used only for certain visitors or only at a certain time of the day. All other cases still use the non-segmented / invariant values which therefore should be allowed to be used.

It appears currently the implementation allows only segmented values or only non-segmented values, but not both at the same time. For culture this makes sense, but for segments it seems like most use cases would use both. A "solution" here would be to store the original property values as some special segment, but preferably this would not be necessary and we can just use the values with the segment set to null.

### Steps to reproduce
In the Starter Kit, first execute these SQL statements to enable segments for the Home content type and heroHeader property type. This has to be done by hand as the API throws an exception at the moment (see #6174).

```sql
UPDATE cmsContentType
SET variations = 2
WHERE pk = 541

UPDATE cmsPropertyType
SET variations = 2
WHERE alias = 'heroHeader'
```

Then, run this code:

```csharp
var services = global::Umbraco.Core.Composing.Current.Services;
// Or inject...
IContentService contentService = services.ContentService;
IContentTypeService contentTypeService = services.ContentTypeService;

{
    // Homepage of Starter Kit
    IContent homepage = contentService.GetById(1104);

    // These work with ContentVariation.Segment.
    // For ContentVariation.CultureAndSegment, 
    // also specify a culture like "en-US".
    homepage.SetValue("heroHeader", "s1 value", segment: "s1");
    homepage.SetValue("heroHeader", "s2 value", segment: "s2");
    homepage.SetValue("heroHeader", "s3 value", segment: "s3");

    // This throws an exception
    homepage.SetValue("heroHeader", "Invariant", segment: null);

    contentService.SaveAndPublish(homepage);

    scope.Complete();
}
```

### Expected result
Property values are saved correctly

### Actual result
An exception is thrown, disallowing a property value without an associated segment.

PerplexDaniel:
### Prerequisites

Work in progress to fix #6179 and #6180 

This PR was made after a discussion with HQ (Jacob + Claus) about the state of segments.

### Description
This modifies the validation process of a Variation. It now allows to pass in a NULL value for the segment when the Variation is varied by segment. Currently in Umbraco this is not allowed, making it impossible to read/write both the segmented value _and_ the original, non-segmented property value.
However, we believe the general use case would be to render them both at the same time. For example, an A/B test could create a segmented value for the page title, but use the original property values elsewhere on the page. 

In order to test this, first enable the Segment variation on the ContentType of Home and PropertyType of "heroHeader" in the Starter Kit. See the code and/or PR of #6174 on how to enable it (or merge the PR added to #6174 into this branch). When the ContentType + PropertyType are set to ContentVariation.Segment, run the following code (also present in #6179):

```csharp
// Or inject
IContentService contentService = Current.Services.ContentService;
{
    // Homepage of Starter Kit
    IContent homepage = contentService.GetById(1104);

    // These work with ContentVariation.Segment.
    // For ContentVariation.CultureAndSegment, 
    // also specify a culture like "en-US".
    homepage.SetValue("heroHeader", "s1 value", segment: "s1");
    homepage.SetValue("heroHeader", "s2 value", segment: "s2");
    homepage.SetValue("heroHeader", "s3 value", segment: "s3");

    homepage.SetValue("heroHeader", "Invariant", segment: null);

    contentService.SaveAndPublish(homepage);

    scope.Complete();
}
```

This now completes without exceptions. To verify that the property values have been saved correctly, check the `umbracoPropertyData` table:
```sql
SELECT *
FROM [umbracoPropertyData]
WHERE segment IS NOT NULL
```

This should yield rows with the values for segments s1/s2/s3. Umbraco seems to also add these values to the last content version, so likely you will see 6 rows rather than the expected 3. I do not believe we caused this.

To render these values in the front-end, this code can be put in `home.cshtml`:
Note that to get the unsegmented value explicitly, we pass in the empty string. Passing in null would go to the "current segment" which can be different based on the currently active `IVariationContextAccessor`. 
```cshtml
<h2>Unsegmented: @Model.Value("heroHeader", segment: "")</h2>
<h2>s1: @Model.Value("heroHeader", segment: "s1")</h2>
<h2>s2: @Model.Value("heroHeader", segment: "s2")</h2>
<h2>s3: @Model.Value("heroHeader", segment: "s3")</h2>
```
The view should now render the unsegmented property value (default is "Umbraco Demo") together with all the segmented property values.

This PR is submitted so you guys can have a quick look at the changes and whether they make any sense or not.


zpqrtbnk:
Commenting on the PR and the two isses:

You write in #6179:
>A "solution" here would be to store the original property values as some special segment, but preferably this would not be necessary and we can just use the values with the segment set to null.

Yes, that was the original design idea: use a "default" segment if needed. But, as you point out, maybe it makes less sense for segments, and segments could both support an unsegmented value, and segmented values, at the same time. In other words, that "default" segment could very well be the unsegmented one. Makes sense.

And then in #6180 you try to use:

    @Model.Value("heroHeader", segment: null)

Please be aware that this does not return the value for the unsegmented value, but the value for *whatever segment is the current segment* as indicated by the `VariationContext`. See `VariationContextAccessorExtensions` to see how it works. Internally, the unsegmented value is represented by an empty string, not `null`.

The changes you propose to `ValidateVariation` seem OK. Basically, a `null` segment will always be validated, even when `VariesBySegment()`. At `IContent` level, you should be able to set unsegmented (`null`) or segmented (eg `"mobile"`) values, and they will be saved to DB.

Note, however, that at `IPublishedContent` level, `null` means "use the current value" and  `@Model.Value("heroHeader", segment: null)` will return the "current segment" value, and if you explicitely want the unsegmented value, and the current segment is, say, "foo", you'd have to get `@Model.Value("heroHeader", segment: string.Empty)`.

PerplexDaniel:
@zpqrtbnk 
Thanks for your reply Stephan, great that you took some time to look over this.

We were a bit confused at first about what to pass in as segment to actually get the "null"-segment values when the current segment is something else. 
At the time of posting #6180 we indeed thought we would have to use `@Model.Value("heroHeader", segment: null)`, but when working on the PRs we saw that would default back to the "current" value obtained from the VariationContextAccessor. We figured out the correct way then, so in this PR's description we did call it correctly: `Unsegmented: @Model.Value("heroHeader", segment: "")`.

PerplexDaniel:
@Shazwazza We have added unit tests for all the inputs of the `ValidateVariation` method. We have also added some missing test combinations and included the empty string in all test cases. Judging from the source it seems empty string and null should behave the same, so we believe this case should be tested as well.

Shazwazza:
@PerplexDaniel great stuff! I am reviewing now. I will need to refactor the test case variable names because as they are, they are meaningless :P 

Maybe you've already figured out what they stand for? Else i'll try to figure it out, but things like this:

`void Assert4A(ContentVariation v, string c, string s, bool xx)`

`void Assert4B(ContentVariation v, string c, string s, bool ew, bool nn, bool en, bool nw)`

any clue what these abbreviations actually mean? else i'll good and I'll research it.

Shazwazza:
Ahh, figured it out, all good, will update method/param names and add some code comments

PerplexDaniel:
@Shazwazza Yeah we were a bit puzzled at first as well by these curious names. We also expected multiple test methods with descriptive names, but alas. In the end we did figure it out:
```
c: Culture (e.g. "en-US")
s: Segment (anything you want as segment alias)
xx: Value passed in for both "exact" and "wildcards" parameters of ContentVariation.ValidateVariation
```

As for `ew/nn/en/nw`: those are indeed funky. They stand for the expected boolean result for the given culture/segment, in all 4 combinations of "exact" and "wilcards" parameters (true/true, false/false, true/false, false/true). The first letter is "e" or "n" and refers to exact match (true or false, respectively). The second letter is "n" or "w" and refers to "wildcards" (true or false, respectively).

Shazwazza:
Great! everything works, tested, code seems fine and I've updated the parameter names to make a little more sense.

Issue #6193, opened 2019-08-26
Core Multi Url Picker doesn't work inside Macro


viksohal:
Hi @nul800sebastiaan  , There is still an issue with the link picker when used in a macro, "ignoreUserStatusNodes" comes as null, I did a null check for this and that has fixed it but that js file will get overwritten with a clean build. This is the error we are getting.

TypeError: Cannot read property 'ignoreUserStartNodes' of null
at Object.$scope.openLinkPicker (umbraco.controllers.js?cdv=20190810:16193)
at angular.min.js?cdv=20190810:74
at angular-mobile.js?cdv=20190810:270
at Object.$eval (angular.min.js?cdv=20190810:92)
at Object.$apply (angular.min.js?cdv=20190810:92)
at HTMLAnchorElement. (angular-mobile.js?cdv=20190810:269)
at HTMLAnchorElement.dispatch (jquery.min.js?cdv=20190810:3)
at HTMLAnchorElement.r.handle (jquery.min.js?cdv=20190810:3)

For temp fix I have updated the umbraco.controller.js line 16192 to
dataTypeId: $scope.model && $scope.model.dataTypeId ? $scope.model.dataTypeId : null,
ignoreUserStartNodes: ($scope.model.config && $scope.model.config.ignoreUserStartNodes) ? $scope.model.config.ignoreUserStartNodes : "1",

nul800sebastiaan:
Sure, sounds great to add a null check there, we'd be happy to look at a fix if you could send us one? :)

umbrabot:
Hi @viksohal,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

viksohal:
Hi @nul800sebastiaan ,

I am attaching the umbraco.controllers.js file with the fix on line 16192
[umbraco.controllers.txt](https://github.com/umbraco/Umbraco-CMS/files/3543433/umbraco.controllers.txt)

Thanks

nul800sebastiaan:
[...]

PR #6211, opened 2019-08-27
V7: Core Multi Url Picker doesn't work inside Macro

stevemegson:
Shamelessly taking credit for @viksohal's fix for #6193 by putting his change into a PR...


nul800sebastiaan:
Lovely, thanks @stevemegson and @viksohal ! 👍

nul800sebastiaan:
Although.. hmm, the default value seem to be to set `ignoreUserStartNodes` to true if we can't determine it from the config, I think that is wrong, I'll give it another test.

nul800sebastiaan:
While it doesn't seem to make a difference if the default is 0 or 1, I've set it to 0 just to make sure, `ignoreUserStartNodes` should always be false unless you specifically configure it to be `true`.

PR #6212, opened 2019-08-27
Handle null/empty values in RefactorMacroColumns migration

### LINKED ISSUE #5827: v7 to 8.1 content migration fails due to missing data on a macro
The error is:

```
The database failed to upgrade. ERROR: The database configuration failed with the following message: Cannot insert the value NULL into column 'macroSource', table 'C:\GIT\SWAT-AUTO-UMBRACO8\APP_DATA\UMBRACO.MDF.dbo.cmsMacro'; column does not allow nulls. UPDATE fails.
 The statement has been terminated.
```

The cause seems to be that the partial view field for this macro was blank, deleting the macro helped to move the migration on.

---
_This item has been added to our backlog [AB#2236](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2236)_

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #5827 

### Description
The RefactorMacroColumns migration assumes that unused columns in `cmsMacro` will be NULL. This is true when a macro is first created while adding a partial view file, but they'll be empty strings if the macro has then been edited. It also assumes that at least one column will be non-NULL (which should be true, but apparently not always).

Testing:
- In 7.15, create a macro but don't save it. The row in `cmsMacro` should have NULL in `macroPython`, `macroScriptType`, `macroScriptAssembly`, `macroXSLT`.
- Create an XSLT macro and save it. The row in `cmsMacro` should have empty strings in `macroPython`, `macroScriptType`, `macroScriptAssembly`.
- Upgrade to 8.1.
- Both macros should be `macroType` 4. The ex-XSLT macro should have the XSLT filename in macroSource, not an empty string.

### DIFF
--- a/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/RefactorMacroColumns.cs
+++ b/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/RefactorMacroColumns.cs
@@ -15,21 +15,17 @@ namespace Umbraco.Core.Migrations.Upgrade.V_8_0_0
             {
                 //special trick to add the column without constraints and return the sql to add them later
                 AddColumn<MacroDto>("macroType", out var sqls1);
-                //now we need to update the new column with some values because this column doesn't allow NULL values
-                Update.Table(Constants.DatabaseSchema.Tables.Macro).Set(new { macroType = (int)MacroTypes.Unknown}).AllRows().Do();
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Ah yes, this is rather obscure but makes sense, thanks very much @stevemegson !

PR #6213, opened 2019-08-27
V8: Use a standard confirm dialog when removing content type compositions

kjac:
### Prerequisites

### Description

When you remove an existing composition from a content type, the confirmation dialog that pops up is rather unlike any other place in the backoffice. It is also kinda broken, as you can scroll it out of view:

remove-composition-before

This PR moves the confirmation to a default dialog (overlay), which is first and foremost in line with the other dialogs, and as an added bonus you can't scroll it out of view 😆 

Here's how things work with this PR applied:

remove-composition-after


nul800sebastiaan:
Lovely, thanks!

nul800sebastiaan:
Cherry picked for 8.1.4 87a0350d44cd7e76d9534e0f84ab800bd27e1ecc

Issue #6217, opened 2019-08-28
Preview not working with remove trailing slash


jveer:
Preview not working when removing trailing slashes via web.config:
```
<rule name="Remove trailing slash" stopProcessing="true">
	<match url="(.*)/$"/>
	<conditions>
		<add input="{REQUEST_FILENAME}" matchType="IsFile" negate="true"/>
		<add input="{REQUEST_FILENAME}" matchType="IsDirectory" negate="true"/>
	</conditions>
	<action type="Redirect" redirectType="Permanent" url="{R:1}"/>
</rule>
```

PR incoming

nul800sebastiaan:
Thanks for the PR, I left a comment. Also, you do know about the `<addTrailingSlash>true</addTrailingSlash>` setting in umbracoSettings.config? If you change the default to false, Umbraco shouldn't add any trailing slashes by default.

kjac:
If this is V8 related, I beg to differ. If you remove trailing slashes by means of Umbraco config, preview works just fine. At least that's my experience 🙂

patrickdemooij9:
@kjac But shouldn't the preview functionality work with the iis rule aswell? I assume not everyone knows about the trailing slash option in the umbraco config.

kjac:
@patrickdemooij9 oh by all means, if you can fix it so it works in all permutations; that is

- forced *removal* of trailing slash via umbraco config
- forced *addition* of trailing slash via umbraco config
- forced *removal* of trailing slash via IIS rewrite rules
- forced *addition* of trailing slash via IIS rewrite rules

patrickdemooij9:
@kjac I didn't mean to be offensive with my message. So sorry if it was. I was just trying to share my opinion on the matter :). 

I was just saying that I, as an user that doesn't know about addTrailingSlash, would expect the preview to work if I use the IIS rewrite rule to remove the trailing slash. The rest of the Umbraco CMS seems to work fine with it. It might be nice to avoid the confusion for the user by either a warning or by making the preview work in this scenario. I don't know how much work this would take as I don't claim to be a professional with the Umbraco code, but I think it would be a nice addition to the Umbraco CMS.

jveer:
@kjac I've just checked for all options you've listed, however:
- Forced _removal_ and _addition_ through the umbraco config doesn't have any effect on the backoffice, so works in either scenario.
- Forced _removal_ via IIS rewrite rules works as expected (after change)
[...]

PR #6218, opened 2019-08-28
temp-6217 | Preview not working with remove trailing slash

jveer:
Issue #6217

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
@@ -848,14 +848,14 @@
             // Chromes popup blocker will kick in if a window is opened
             // without the initial scoped request. This trick will fix that.
             //
-            var previewWindow = $window.open('preview/?init=true', 'umbpreview');
+            var previewWindow = $window.open('preview?init=true', 'umbpreview');
 
             // Build the correct path so both /#/ and #/ work.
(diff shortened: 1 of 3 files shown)


jveer:
@nul800sebastiaan Thanks for the feedback, actually I just noticed that the preview was hard-coded with a ending slash in one of the controllers. So I removed that hard-coded slash to make the preview work even when removing the slash via web.config instead of the `umbracoSettings`. PR is updated

nul800sebastiaan:
@jveer Have you tested this without your rewrite rules as well? 

There's a few seemingly random changes in here, there must be a reason we added the `../` at some point for example and I am afraid that we're just working around your particular problem right now instead of finding a structural solution.

jveer:
@nul800sebastiaan Yes, I've tested this locally with both the rewrite rule and without, same for the setting in `umbracoSettings`.

However, I think the reason `../` was added, was because the preview page had an trailing slash. The only location in the code where the `lazyload.min.js` is loaded using `../` is for the preview index, for all other locations it's directly loading `lib/lazyload-js/lazyload.min.js`.

Therefore I've changed all locations where the preview page is called to exclude the trailing slash, so the view will always work with the new accessing of the preview page.

nul800sebastiaan:
Thanks again @jveer - I've closed the related issue with a reason here: https://github.com/umbraco/Umbraco-CMS/issues/6217#issuecomment-553180844

PR #6220, opened 2019-08-28
V8: Fix preview with multiple cultures + highlight the current culture

kjac:
### Prerequisites

### Description

The preview sidebar doesn't handle multiple cultures very gracefully:

1. You can't scroll the sidebar downwards, leaving you unable to see all cultures (let alone close the preview).
2. The currently active culture isn't highlighted like the currently active device is.

Here, have a GIF:

preview-cultures-before

This PR fixes both of these issues; here's how the sidebar works when the PR is applied:

preview-cultures-after

...and for a site with only a few cultures:

preview-cultures-after-2

### Not a perfect solution

This isn't necessarily a perfect solution to the scrolling issue, since hiding the scrollbar is pretty bad practice. But I can't think of any better way to handle it for the time being... at least not a way that handles both multiple devices (if you expand beyond the default ones) and multiple cultures.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/canvas-designer.less
+++ b/src/Umbraco.Web.UI.Client/src/less/canvas-designer.less
@@ -266,7 +266,6 @@ a, a:hover{
 ul.sections {
     display: block;
     background: #1b264f;
-    height: 100%;
     position:absolute;
     top: 90px;
     width: 80px;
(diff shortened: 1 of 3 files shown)


emmaburstow:
Lovely work here, Kenn. As usual. Thanks for picking this up. We'll take a look and let you know if it works as a solution. Perhaps @nul800sebastiaan might want to look at alternative approaches here? 

Emma

nul800sebastiaan:
I was just about to reply! This solution is interesting but very hard to tell you can scroll. We talked about this one at HQ recently and think it's better to have 1 globe icon there that will allow you to switch between the available languages on click. Not sure what that should look like exactly, but scrolling this way is not very intuitive. That said, it's better than what we have now and I love incremental improvements, so maybe it's good to merge this one now and then we can take it from there.

nul800sebastiaan:
Lovely, thanks very much @kjac !

PR #6221, opened 2019-08-28
V8 - populate mediaPath when property alias is not umbracoFile

### LINKED ISSUE #5847: Media thumbnails not shown after 7.14.0 -> 7.15.0 upgrade

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #5847, applied to V8

### Description
Allows a `DataValueEditor`  to implement a `ToMediaPath()` method indicating that properties of this type can provide a value for `umbracoMediaVersion.path`. We then look for the first property providing a non-null path, rather than hardcoding a check of `umbracoFile`. This also removes the assumption that the path with start with "/media".

Also includes a migration which populates missing path values. For speed, this is hardcoded to look at Upload and Image Cropper properties rather than using ToMediaPath().

Testing:
- In 8.1, create a media type with an upload or image cropper property. Set the alias to anything but `umbracoFile`.
- Create an item of this type, and confirm that the thumbnail is displayed correctly in a media picker.


emmaburstow:
Hi Steve,

Thanks for the work here along with the testing steps. We'll take a look and get back to you if we need anything.

Thanks,

Emma

stevemegson:
I've rebased this to deal with the merge conflicts. I've kept only the changes to populate missing paths during an upgrade. The remaining changes to handle populating the path on new media were already handled in a slightly different way by #7075.

Issue #5724, opened 2019-06-26
Unable to add multiple property groups in membertype/doctype editor


skttl:
I was building out a membertype, and had added all my needed properties to the first group I created. I then proceeded to add the needed groups, so I could sort the properties in the end.

But when you create a new group, the "Add new group" option disappears until you have added properties to that group. And only if you add new properties, the option will reappear. I tried moving existing properties into it, but that didn't give me the option.

I made a gif.
https://i.imgur.com/fozi3CK.gifv

nul800sebastiaan:
Yeah, it's a deliberate choice to not save empty groups as they don't "do" anything. 

I'll need to discuss this way of working at HQ though, I understand what you're trying to do and how we're slowing down your productivity this way.

Will get back to you.

nul800sebastiaan:
Hey @skttl - sorry for the delay! I've had a chat about this and we feel like it would indeed be a productivity gain if this limitation wasn't in place.

umbrabot:
Hi @skttl,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6222

Cherry picked for 8.1.5 c7ea93199f59f3ef64201f17fd347e5ba8434b6f

PR #6222, opened 2019-08-28
5724: Unable to add multiple property groups

patrickdemooij9:
Issue: #5724 (Unable to add multiple property groups in membertype/doctype editor)

Moved initGroup to the addGroup function so that it'll create a new initial group every time a new group is added.

Also changed the deleteProperty function so that it doesn't remove the init group when the last property is removed, because there should always be an init group.


patrickdemooij9:
Just found some bugs regarding properties, will fix those first

patrickdemooij9:
The bugs regarding the properties weren't caused by this PR. I have created another issue for the groups/property problems as I think it's outside the scope of this PR: #6223

nul800sebastiaan:
Lovely, thanks very much @patrickdemooij9 👍

Issue #4253, opened 2019-01-25
V8: Dashboard Styling (Models Builder & PublishedContent)


warrenbuckley:
The following dashboards in the Settings section look a bit inconsistent next to the other dashboards in the Settings section.

* Models Builder `src\Umbraco.Web.UI\App_Plugins\ModelsBuilder\modelsbuilder.htm`
* Published Content `src\Umbraco.Web.UI.Client\src\views\dashboard\settings\publishedstatus.html`

Maybe you can pick this up please @nielslyngsoe

warrenbuckley:
## Models Builder
![image](https://user-images.githubusercontent.com/1389894/51751263-5855ea00-20ac-11e9-8253-f03e5f2416e6.png)

## Published Status
![image](https://user-images.githubusercontent.com/1389894/51751311-79b6d600-20ac-11e9-9bdb-527cc37475db.png)

bergmania:
My UI eyes can only see that the ModelBuilders dashboard looks inconsistent to the rest.

@warrenbuckley , What do you think is inconsistent in the Published Content ?

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6225

PR #6225, opened 2019-08-29
Cleanup to the published status dashboard

poornimanayar:
### Prerequisites

If there's an existing issue for this PR then this fixes <https://github.com/umbraco/Umbraco-CMS/issues/4253>

The above issue points out some inconsistencies to the modelsbuilder and the published status dashboards. While the modelsbuilder dashboard looks consistent now, the published status dashboard could do with some tweaks. I have got rid of the massive textarea and changed the orange colour buttons to red(similar to modelsbuilder dashboard). I was wondering whether it should be made to look like the Examine Management dashboard, with each title having the blue background, with the text and button within but decided against it as the styling seems to be used for panels. This also raises another question in my head - Should the rebuild indexes button on Examine Management dashboard be red - in line with Modelsbuilder dashboard? Dont know whether I am overthinking it here.

Any changes let me know. :-)

**Before**
image

After
image

Poornima

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/dashboard/settings/nucache.html
+++ b/src/Umbraco.Web.UI.Client/src/views/dashboard/settings/nucache.html
@@ -4,13 +4,13 @@
         <umb-load-indicator></umb-load-indicator>
     </div>
 
-    <p style="min-height: 120px; background: #f8f8f8; padding: 10px; margin-bottom: 4px;">
+    <p>
         <span ng-show="vm.working">(wait)</span>
         <span ng-show="!vm.working">{{vm.status}}</span>
     </p>
 
     <div>
-        <button type="button" ng-click="vm.verify($event)" class="btn btn-warning">Refresh status</button>
+        <button type="button" ng-click="vm.verify($event)" class="btn btn-danger">Refresh status</button>
     </div>
 
     <h4 class="mt4">Memory Cache</h4>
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks! Looks fine for now 😁

ronaldbarendse:
@poornimanayar Changing the `Rebuild index` buttons below Examine Management to red is a good idea, as that's quite a disruptive action (especially on larger sites - although you first get a nice confirmation dialog first).

And I also think it should be split with tabs, e.g. Status and Caches. And the descriptions below the titles can be changed to the small grey styling, something like this:

Published Status

Maybe even put the actions/buttons in a grey container, like on the Examine Management dashboard?

ronaldbarendse:
The big titles with the same name as the dashboard/tab also seem redundant... Although when you only have 1 dashboard (for the active section), the tabs aren't displayed:

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web.UI.Client/src/views/common/dashboard.html#L9

Should we have a dashboard setting to force always displaying a tab? So if the dashboard layout/content relies on the tab title, it can specify that somehow. The above check would then have to check whether we have more than 1 dashboard/tab or one of them (e.g. the only one) has this setting enabled (and still render nothing with zero dashbaords).

poornimanayar:
Hi @ronaldbarendse ,

Hmm, good idea. I will do that PR up.

Poornima

ronaldbarendse:
Great! I've created the mock-up by copying most HTML/styles from the Examine Dashboard, so make sure to do the same 😉

IMHO the status text should have a grey background, so it's clear this is some sort of output (opposed to a description).

PR #6228, opened 2019-08-29
Fix member properties [#6227]

### LINKED ISSUE #6227: Some published member properties are empty
Some published member properties are empty, ie they are not populated with their correct value. For instance, `Email`.

Repro: fetch a member eg

    var member = umbracoContext.PublishedSnapshot.Members.GetByEmail("foo@bar.com");

And try to get its email with `member.Value("Email")` - it should return the `"foo@bar.com"` email, but the value is empty.

Cause by `PublishedContentType` which around line 98 adds the builtin properties as user properties. `IPublishedContentTypeFactory` should be extended with a `CreateCorePropertyType` method which would be used to ensure that these properties are added as builtin (non-user) - and then they would properly receive a value.

zpqrtbnk:
Fixes #6227.


nul800sebastiaan:
Thanks @zpqrtbnk 👍

Issue #5007, opened 2019-03-18
Putting in a content title > 255 characters causes "String or binary data would be truncated. The statement has been terminated." exception.


kgiszewski:
## The Issue
Whilst not a "bug" per se, editors were alarmed when they copy\pasted a long name for a node into the content name field (see images).

## To Replicate
To replicate, type something greater than 255 characters into any node field title. Save\Publish.

## Version
v7.14.0

## Proposed Solution
Proposed solution is to limit the title via the UI to 255 so that the user cannot type any further. Maybe some wording that appears when the limit is encountered.

## Images
![image](https://user-images.githubusercontent.com/4653907/54531395-8e507580-495b-11e9-9017-ce910fbbf850.png)

![image](https://user-images.githubusercontent.com/4653907/54531449-b3dd7f00-495b-11e9-9ad5-f8ba58a56be9.png)

Shazwazza:
Hi sorry for the delay but i think the referenced PR didn't reference this correctly, it is here and is merged https://github.com/umbraco/Umbraco-CMS/pull/5033 so i believe this is fixed and due out in 8.1

Shazwazza:
Doh! actually sorry my mistake, it's this PR (the other one was closed) https://github.com/umbraco/Umbraco-CMS/pull/5034

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/5374

PR #6231, opened 2019-08-29
Better exception message when media name exceeds 255 characters

poornimanayar:
This is a very similar issue to [5007](https://github.com/umbraco/Umbraco-CMS/issues/5007). When you try and save a media (image/file) with name exceeding 255 characters it throws an exception. I have added a check to throw an exception with an error message.

I havent added a maxlength attribute to the input field as that would be a sweeping change. I will try and address this issue for doc types, data types, media types and so on before the maxlength can be introduced on the name fields for all the entities.

Poornima


nul800sebastiaan:
👍 Thanks!

Cherry picked for 8.1.5 d044dd07c3ab92b3606d924ff7e6534866406c6d

Issue #4633, opened 2019-02-18
v8 Too long document type name gives error


rhealefever:
An extreme case, but should be noted anyway.

I don't know why anyone would, but giving the name of a document type something with more than 255 characters gives this error:

![image](https://user-images.githubusercontent.com/47108275/52952091-d93e9380-3383-11e9-972a-4f98aee07467.png)

This error is seen everywhere (names, descriptions, etc) in the back office anytime too many characters are used.

Trying this in v7 Cloud gives a different kind of error:
![image](https://user-images.githubusercontent.com/47108275/52952111-e8254600-3383-11e9-95fe-18a171093615.png)

Perhaps this should just be replaced by a simple "You've used too many characters" error.

Error:
```
at System.Data.SqlServerCe.SqlCeCommand.ProcessResults(Int32 hr)
   at System.Data.SqlServerCe.SqlCeCommand.ExecuteCommandText(IntPtr& pCursor, Boolean& isBaseTableCursor)
   at System.Data.SqlServerCe.SqlCeCommand.ExecuteCommand(CommandBehavior behavior, String method, ResultSetOptions options)
   at System.Data.SqlServerCe.SqlCeCommand.ExecuteNonQuery()
   at StackExchange.Profiling.Data.ProfiledDbCommand.ExecuteNonQuery() in C:\projects\dotnet\src\MiniProfiler.Shared\Data\ProfiledDbCommand.cs:line 272
   at Umbraco.Core.Persistence.FaultHandling.FaultHandlingDbCommand.<ExecuteNonQuery>b__31_0()
   at Umbraco.Core.Persistence.FaultHandling.FaultHandlingDbCommand.<>c__DisplayClass33_0`1.<Execute>b__0()
   at Umbraco.Core.Persistence.FaultHandling.RetryPolicy.ExecuteAction[TResult](Func`1 func)
   at Umbraco.Core.Persistence.FaultHandling.FaultHandlingDbCommand.Execute[T](Func`1 f)
   at Umbraco.Core.Persistence.FaultHandling.FaultHandlingDbCommand.ExecuteNonQuery()
   at NPoco.Database.ExecuteNonQueryHelper(DbCommand cmd)
   at NPoco.Database.NPoco.IDatabaseHelpers.ExecuteNonQueryHelper(DbCommand cmd)
   at NPoco.Database.Execute(String sql, CommandType commandType, Object[] args)
   at NPoco.Database.Execute(Sql Sql)
[...]

PR #6232, opened 2019-08-29
Friendly exception message when datatype with name more than 255 char…

poornimanayar:
The issue [4633](https://github.com/umbraco/Umbraco-CMS/issues/4633) mentions how exception occurs on various parts of the CMS when the name exceeds 255 characters. I have added better exception on the DataType Service. I havent added maxlength attribute to the input field as that would be a sweeping change. It can be introduced once all the entities have this check in place

Before
image

After
image

Let me know if anything more is needed
Poornima


nul800sebastiaan:
Thank you! 👍

PR #6234, opened 2019-08-29
better exception message on content type save

poornimanayar:
The issue [4633 ](https://github.com/umbraco/Umbraco-CMS/issues/4633) mentions how an exception occurs on various parts of the CMS when the name exceeds 255 characters. I have added better exception upon saving on content types. A check has been put in place before the content type is saved as well as template creation. I havent added maxlength attribute to the input field as that would be a sweeping change. It can be introduced once all the entities have this check in place. This should also fix the same issue on MemberType save

Before
image

After
image

Regards,
Poornima


nul800sebastiaan:
woohoo, one more! 😁

Issue #6226, opened 2019-08-29
ReplaceLineBreaksForHtml does not HTML encode the text


ronaldbarendse:
Both Umbraco 7 and 8 have methods to replace line breaks with `<br />`. This can come in handy when rendering `Umbraco.Textarea` properties, but the current implementations contain several awkward implementation details regarding naming and (not) correctly HTML encoding the input (causing a potential security issue).

### Umbraco 7
First off, the V7 version returns a `string`:
https://github.com/umbraco/Umbraco-CMS/blob/54a2aa00a78caa4e6fe7b3b3cb9ff418fd1f408d/src/Umbraco.Web/HtmlStringUtilities.cs#L21-L29

https://github.com/umbraco/Umbraco-CMS/blob/4fa523db434b1832dc8022ff6b3788f141c82ae4/src/Umbraco.Web/UmbracoHelper.cs#L1411-L1419

Both have parameters named `text` and as the method name/description contains `HTML`, you would assume the text is correctly encoded, so you can use the return value as plain HTML - **wrong**:

```cshtml
@Html.Raw(Html.ReplaceLineBreaksForHtml("This is the first line\r\nThe second\r\n<script>document.write('And the third!');</script>"))
@* Becomes (note the script tag isn't HTML encoded): *@
This is the first line<br />The second<br /><script>document.write('And the third!');</script>
```

As you're explicitly using `@Html.Raw()`, you could argue correctly encoding the input is your own responsibility. So to correctly use this method, you would need to write the following in your views:

```cshtml
@Html.Raw(Html.ReplaceLineBreaksForHtml(Html.Encode("This is the first line\r\nThe second\r\n<script>document.write('And the third!');</script>")))
```

Not a very useful/easy to use method if you ask me...

### Umbraco 8
So V8 tried to make it easier to work with this method by returning `IHtmlString`:

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web/HtmlStringUtilities.cs#L18-L26

https://github.com/umbraco/Umbraco-CMS/blob/06b136fdf66ae685c8985af90ac075535e635e74/src/Umbraco.Web/HtmlHelperRenderExtensions.cs#L858-L867

[...]

PR #6235, opened 2019-08-29
Correctly html encode text

sniffdk:
Before returning the html string, referencing issue #6226

### DIFF
--- a/src/Umbraco.Web/HtmlStringUtilities.cs
+++ b/src/Umbraco.Web/HtmlStringUtilities.cs
@@ -27,7 +27,7 @@ namespace Umbraco.Web
         }
         
         /// <summary>
-        /// Html encodes the text and replaces text line breaks with HTML line breaks.
+        /// HTML encodes the text and replaces text line breaks with HTML line breaks.
         /// </summary>
         /// <param name="text">The text.</param>
         /// <returns>The HTML encoded text with text line breaks replaced with HTML line breaks (<c>&lt;br /&gt;</c>).</returns>
--- a/src/Umbraco.Web/HtmlStringUtilities.cs
+++ b/src/Umbraco.Web/HtmlStringUtilities.cs
@@ -33,7 +33,7 @@ namespace Umbraco.Web
         /// <returns>The html encoded text with text line breaks replaced with HTML line breaks (<br/>)</returns>
         public HtmlString ReplaceLineBreaks(string text)
         {
-            var value = HttpUtility.HtmlEncode(text ?? string.Empty)
+            var value = HttpUtility.HtmlEncode(text)?
                 .Replace("\r\n", "<br />")
                 .Replace("\r", "<br />")
                 .Replace("\n", "<br />");
--- a/src/Umbraco.Web/HtmlStringUtilities.cs
+++ b/src/Umbraco.Web/HtmlStringUtilities.cs
@@ -19,7 +19,7 @@ namespace Umbraco.Web
         /// Replaces text line breaks with HTML line breaks
         /// </summary>
         /// <param name="text">The text.</param>
-        /// <returns>The text with text line breaks replaced with HTML line breaks (<br/>)</returns>
+        /// <returns>The text with text line breaks replaced with HTML line breaks (<c>&lt;br /&gt;</c>).
         [Obsolete("This method doesn't HTML encode the text. Use ReplaceLineBreaks instead.")]
         public HtmlString ReplaceLineBreaksForHtml(string text)
         {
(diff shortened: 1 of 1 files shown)


sniffdk:
Sorry, that went a bit too fast, I've reverted and added a new method to replace the old one.

src/Umbraco.Web/HtmlStringUtilities.cs
```diff
@@ -20,9 +20,25 @@ public sealed class HtmlStringUtilities
         /// </summary>
         /// <param name="text">The text.</param>
         /// <returns>The text with text line breaks replaced with HTML line breaks (<br/>)</returns>
+        [Obsolete("Doesn't html encode the text. Use ReplaceLineBreaks instead")]
```
ronaldbarendse:
```suggestion
        [Obsolete("This method doesn't HTML encode the text. Use ReplaceLineBreaks instead.")]
```

src/Umbraco.Web/HtmlStringUtilities.cs
```diff
@@ -20,9 +20,25 @@ public sealed class HtmlStringUtilities
...
+        }
+        
+        /// <summary>
+        /// Html encodes the text and replaces text line breaks with HTML line breaks.
+        /// </summary>
+        /// <param name="text">The text.</param>
+        /// <returns>The html encoded text with text line breaks replaced with HTML line breaks (<br/>)</returns>
+        public HtmlString ReplaceLineBreaks(string text)
+        {
+            var value = HttpUtility.HtmlEncode(text ?? string.Empty)
```
ronaldbarendse: `HttpUtility.HtmlEncode()` and `HtmlString` happily accept `null` values, so no need to create empty strings. This also ensures the `Replace` calls are skipped when the text is `null` (by using the null-conditional operator):

```suggestion
            var value = HttpUtility.HtmlEncode(text)?
```

sniffdk: Sure, makes sense 👍

src/Umbraco.Web/HtmlStringUtilities.cs
```diff
@@ -20,9 +20,25 @@ public sealed class HtmlStringUtilities
         /// </summary>
         /// <param name="text">The text.</param>
         /// <returns>The text with text line breaks replaced with HTML line breaks (<br/>)</returns>
```
ronaldbarendse:
```suggestion
        /// <returns>The text with text line breaks replaced with HTML line breaks (<c>&lt;br /&gt;</c>).
```

src/Umbraco.Web/HtmlStringUtilities.cs
```diff
@@ -20,9 +20,25 @@ public sealed class HtmlStringUtilities
...
+        [Obsolete("This method doesn't HTML encode the text. Use ReplaceLineBreaks instead.")]
         public HtmlString ReplaceLineBreaksForHtml(string text)
         {
-            return new HtmlString(text.Replace("\r\n", @"<br />").Replace("\n", @"<br />").Replace("\r", @"<br />"));
+            return new HtmlString(text.Replace("\r\n", @"<br />").Replace("\n", @"<br />").Replace("\r", @"<br />"));            
+        }
+        
+        /// <summary>
+        /// Html encodes the text and replaces text line breaks with HTML line breaks.
```
ronaldbarendse:
```suggestion
        /// HTML encodes the text and replaces text line breaks with HTML line breaks.
```

src/Umbraco.Web/HtmlStringUtilities.cs
```diff
@@ -20,9 +20,25 @@ public sealed class HtmlStringUtilities
...
+        /// </summary>
+        /// <param name="text">The text.</param>
+        /// <returns>The html encoded text with text line breaks replaced with HTML line breaks (<br/>)</returns>
```
ronaldbarendse:
```suggestion
        /// <returns>The HTML encoded text with text line breaks replaced with HTML line breaks (<c>&lt;br /&gt;</c>).</returns>
```

src/Umbraco.Web/HtmlStringUtilities.cs
```diff
@@ -20,9 +20,25 @@ public sealed class HtmlStringUtilities
...
+        public HtmlString ReplaceLineBreaks(string text)
```
ronaldbarendse:
```suggestion
        public IHtmlString ReplaceLineBreaks(string text)
```

sniffdk: Yeah, we might as well, as that is what Umbraco uses in property value converters etc.

nul800sebastiaan:
Thanks @sniffdk for updating!

I don't really like that we're repeating the `Replace` code now but I guess it's fine since we wouldn't ever want to change it anyway because that would be a breaking change.

I guess the `Truncate` method suffers from the same problem and we should probably audit all methods that return `IHtmlString` / `HtmlString`.

sniffdk:
Also thanks to @ronaldbarendse 👍 
Not sure about the need to modify Truncate as I would assume that it expects html in the first place, considering the amount of work it does.

PR #6236, opened 2019-08-29
V8: Accessibility Changes For Title On Content Section

RachBreeze:
This PR sets the page title when the user is in the content section and builds on the work in #5888 

The aim is to help screen reader users and also content editors which have multiple tabs open determine what they are doing in the content section.

The logic for the title being set is in umbeditorcontentheader.directive.js as opposed to init.js to prevent multiple hits on the database to retrieve the content name/ content type being edited.


nul800sebastiaan:
Lovely, thanks a lot!

PR #6239, opened 2019-08-29
Maxlength attribute on macro name

poornimanayar:
The issue [4633 ](https://github.com/umbraco/Umbraco-CMS/issues/4633) mentions how an exception occurs on various parts of the CMS when the name exceeds 255 characters. I have added better exception upon saving on macros. I havent added maxlength attribute to the input field on the right hand panel as that would be a sweeping change. It can be introduced once all the entities have this check in place. I have not added the validation to the `EntityBasic `class either as that would be a sweeping change as well

**Before**
image

image

**After**
image

You should not be able to create a macro with more than 255 characters in the name to begin with. Even if you manually edit the name in the right hand panel name input field, it should give a friendly error message rather than the generic error message.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/macros/create.html
+++ b/src/Umbraco.Web.UI.Client/src/views/macros/create.html
@@ -19,7 +19,8 @@
                 </div>
 
                 <umb-control-group label="@macro_enterMacroName" hide-label="false">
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Sweet, thank you kindly @poornimanayar

PR #6246, opened 2019-08-30
AB 2439 -  Create API controller for uploading images for RTE

warrenbuckley:
## Testing Notes
* Add RTE to a doctype
* Add Grid with a RTE editor allowed in a cell
* Drag a single image from your disk into a RTE
* Verify image uploaded into media section with correct name (matching filename)
* Verify the source code contents of the RTE does it have a `/media/` URL and a `data-udi` attribute set

* Repeat with two or more images dragged into editor
* Copy an image that is on your machine into your clipboard and paste it in such as word document or Google Doc etc

* Test with something thats NOT an image
* Update allowed image file extensions in Umbraco.config & try uploading

### Word & Google Docs
Currently cannot copy Word/Google Doc content with images in clipboard into RTE. You must select the image individually and copy/paste the image in.

## Considerations
Using native `localStorage` for this is this OK as think more widely supported now or do we need to use the AngularJS localStorage which I think may fallback to cookies or similar - thoughts?

## Quick Demo
rte-media-upload

Fixes [AB#2439](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2439)


bjarnef:
@warrenbuckley I wonder if we can use the `ngFileUpload` API for handling the upload which also allow to restrict file/image types, which is already used in media section?
https://github.com/danialfarid/ng-file-upload/blob/master/README.md

warrenbuckley:
> @warrenbuckley I wonder if we can use the `ngFileUpload` API for handling the upload which also allow to restrict file/image types, which is already used in media section?
> https://github.com/danialfarid/ng-file-upload/blob/master/README.md

Hi @bjarnef I am not 100% sure we can do that, as we are using TinyMCE's APIs & functions to deal with uploading & pasting in images and not a unique UI component like the drag & drop of media in the media section

bjarnef:
I think the `Upload` service can be used inside `uploadImageHandler` instead of the traditional `XMLHttpRequest`.

Does the current PR also respect allow filetypes in Umbraco settings config?

warrenbuckley:
> Does the current PR also respect allow filetypes in Umbraco settings config?
Yep it does on the server, we could do some checks clientside before we send it up - but we would need to determine the UX - do we just remove the image automatically, show a prompt or...?

https://github.com/umbraco/Umbraco-CMS/pull/6246/files#diff-cf4c4738fa916b01523c1478966e576dR77-R81

>I think the Upload service can be used inside uploadImageHandler instead of the traditional XMLHttpRequest.

I am not 100% sure what the ng-file-upload directive & it's `Upload.upload()` method will give us as a benefit over `XMLHttpRequest`


Thanks for the feedback & engaging on this PR 👍

bjarnef:
There are some details about the `Upload` service here. https://github.com/danialfarid/ng-file-upload/blob/master/README.md#upload-service
Not sure if we can benefits from something specific and these options, but I thought it was more consistent with the dropzone upload in media section.

But now with ES6 it might be easier to use native javascript (and since we have Babel to compile it for "older" browsers).

bjarnef:
This project also looks cool to allow showing the preview, before uploading the file although we might not want to use this library, but maybe use it as inspiration. https://github.com/pqina/filepond/blob/master/README.md

warrenbuckley:
@bjarnef that does look neat but I think you maybe confused as to what this PR is doing - see video at top of PR description

bjarnef:
Okay, I didn't see the video before. Is it possible to highlight the droparea in dragover like in dropzone in media section?

I guess the media only is created on save or save and publish?

Although not supported yet in this PR, if copying content from Google Docs, Word etc. should the editor allow to select each of the images or all that should be included?

bjarnef:
I guess it also works with svg images, but shown in an `img` element?

warrenbuckley:
@bjarnef they auto get published/saved as they are dragged in, before you even save your content.
I have not tried with dragging in SVG images in yet - one my peer reviewer can test & let me know

bjarnef:
Okay, so og the editor undo the changes in TinyMCE (ctrl+z) the media still exists.
I fear it could end up with a lot of images in media section as editors paste in images, undo, try some other images.

warrenbuckley:
Good point @bjarnef any thoughts on this product owner @hemraker ?

hemraker:
That is a really good point @bjarnef, thanks for raising it! Would be great to have images as temp files until the content is saved or published. I think that would make sense from a workflow perspective. 
We'll need to discuss this one but if you have any other suggestions please do share.

src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
```diff
@@ -156,6 +156,85 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
...
+
+            success(json.location);
+        };
+
+        formData = new FormData();
+        formData.append('file', blobInfo.blob(), blobInfo.blob().name);
+
+        //TODO: Send Media Parent ID from config
+        //TODO: How will each RTE pass in this value to this function?!
+        formData.append('mediaParent', -1);
```
bergmania: Maybe we can just use `Umbraco.Sys.ServerVariables.umbracoSettings.mediaPath` like you use other settings, directly from here?

src/Umbraco.Web/Editors/TinyMceController.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            _mediaService.Save(f, Security.CurrentUser.Id);
+
+            // Need to get URL to the media item and its UDI
+            var udi = f.GetUdi();
+
+
+            // TODO: Check this is the BEST way to get the URL
+            // Ensuring if they use some CDN & blob storage?!
+            var mediaTyped = Umbraco.Media(f.Id);
```
bergmania: You should inject `UmbracoHelper` in the constructor and use that instead of the static one.

bergmania:
## Test notes:

- Upload svg:
  - Does not upload, but change the url of the browser and shows the svg.
- Upload exe:
  - Does not upload, but start a download of the exe.
- Upload gif:
  - Uploads to the root.
- Upload gif with user not allowed to see media root:
  - Uploads to the root.
- Copy/paste text and image from word:
  - Paste only text.
- Copy/paste text and image from google docs:
  - Paste only text.
- Copy/paste image from word:
  - Uploads the image as png no matter the original format.
- Copy/paste image from google docs:
  - Does nothing.


## Conclusions 
I think we need to avoid the default browser behavior, when dropping a non-supported file format into the drop zone..
Also I think we need to style the drop-zone such that it is easier to see you can drop images.

I think we need to upload to the users first media root. I don't think it is good that you can upload but not delete.

Can't upload the image individually from google docs?

Regarding localstorage, I think we should use the angular instead of native, to align with the rest of the code base.

I also agree with @hemraker and @bjarnef about the actual save in media section should not happen before the document is saved/published.

hemraker:
Good notes @bergmania 🙌

_Also I think we need to style the drop-zone such that it is easier to see you can drop images._
Having the marker indicate where the image is created would suffice, and might even be more precise. If that is not possible I agree there should be an indication that it is possible to drop the image here. 

_I don't think it is good that you can upload but not delete._
If the upload happens on Save/Publish and we add a nice notification stating where the media item is created, there is no need for deleting anything.

src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
warrenbuckley: @bergmania this will be set on each on Property Editor config and not global to all RTEs
It was just a noisy comment whilst I was working, so feel free to ignore it

src/Umbraco.Web/Editors/TinyMceController.cs
warrenbuckley: Yep will fix that up @bergmania

warrenbuckley: @bergmania `Umbraco` is coming from the base class - `UmbracoApiControllerBase` so do you still want me to inject this into the constructor?

warrenbuckley:
This is ready for a re-review @bergmania or @Shazwazza

src/Umbraco.Web/PropertyEditors/RichTextPropertyEditor.cs
```diff
@@ -99,6 +112,11 @@ public override object FromEditor(Core.Models.Editors.ContentPropertyData editor
 
                 var editorValueWithMediaUrlsRemoved = TemplateUtilities.RemoveMediaUrlsFromTextString(editorValue.Value.ToString());
                 var parsed = MacroTagParser.FormatRichTextContentForPersistence(editorValueWithMediaUrlsRemoved);
+
+                var userId = Current.UmbracoContext.Security.CurrentUser.Id;
+
+                // TODO: In future task(get the parent folder from this config) to save the media into
+                parsed = TemplateUtilities.FindAndPersistPastedTempImages(parsed, -1, userId, _mediaService, _contentTypeBaseServiceProvider);
```
bergmania:
```suggestion
                parsed = TemplateUtilities.FindAndPersistPastedTempImages(parsed, Constants.System.Root, userId, _mediaService, _contentTypeBaseServiceProvider);
```

Don't more introduce magic integers.

src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
```diff
@@ -22,25 +24,74 @@ namespace Umbraco.Web.PropertyEditors
...
+
+                foreach(var rte in rtes)
+                {
+                    // Parse the HTML
+                    var html = rte.Value?.ToString();
+                    
+                    var userId = Current.UmbracoContext.Security.CurrentUser.Id;
+
+                    // TODO: In future task(get the parent folder from this config) to save the media into
+                    var parsedHtml = TemplateUtilities.FindAndPersistPastedTempImages(html, -1, userId, _mediaService, _contentTypeBaseServiceProvider);
```
bergmania:
```suggestion
                    var parsedHtml = TemplateUtilities.FindAndPersistPastedTempImages(html, Constants.System.Root, userId, _mediaService, _contentTypeBaseServiceProvider);
```

Don't introduce more magic integers

src/Umbraco.Web/Templates/TemplateUtilities.cs
```diff
@@ -183,5 +186,57 @@ public static string ResolveMediaFromTextString(string text)
         internal static string RemoveMediaUrlsFromTextString(string text)
             // see comment in ResolveMediaFromTextString for group reference
             => ResolveImgPattern.Replace(text, "$1$3$4$5");
+
+        internal static string FindAndPersistPastedTempImages(string html, int mediaParentFolder, int userId, IMediaService mediaService, IContentTypeBaseServiceProvider contentTypeBaseServiceProvider)
```
bergmania: I think this method is unnecessarily hard to read, due to very short variable names
`f` => `file`
`fs` => `fileStream`
`absTmpImgPath` => `absoluteTemporaryImagePath`
....

src/Umbraco.Web/Templates/TemplateUtilities.cs
```diff
@@ -183,5 +186,57 @@ public static string ResolveMediaFromTextString(string text)
...
+            htmlDoc.LoadHtml(html);
+
+            var tmpImages = htmlDoc.DocumentNode.SelectNodes("//img[@data-tmpimg]");
+            if (tmpImages == null || tmpImages.Count == 0)
+                return html;
+            
+            foreach (var img in tmpImages)
+            {
+                // The data attribute contains the path to the tmp img to persist as a media item
+                var tmpImgPath = img.GetAttributeValue("data-tmpimg", string.Empty);
```
bergmania: `"data-tmpimg"` could be saved in a local constant.. It is used 3 times in this method. If we wan't to change the name, we should update all 3 locations anyway.

src/Umbraco.Web/PropertyEditors/RichTextPropertyEditor.cs
```diff
@@ -71,10 +85,9 @@ public override object Configuration
             /// Format the data for the editor
             /// </summary>
             /// <param name="property"></param>
-            /// <param name="dataTypeService"></param>
+            /// <param name="dataTypeService"></param>Rte
```
bergmania: Seems like an error this is added here

src/Umbraco.Web/Editors/TinyMceController.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            if (Request.Content.IsMimeMultipartContent() == false)
+            {
+                throw new HttpResponseException(HttpStatusCode.UnsupportedMediaType);
+            }
+
+            // Backoffice user ID (needed for unique folder path) to help with concurrent users
+            // to avoid filename clash along with UTC current time
+            var userId = Security.CurrentUser.Id;
+            var imageTempPath = IOHelper.MapPath(SystemDirectories.TempImageUploads + "/" + userId + "_" + DateTimeOffset.UtcNow.ToUnixTimeSeconds());
```
bergmania: Why not just use a guid ? Can't see you need to user ID, and en theory in a load balanced setup the time could be a little off. I think a guid is simpler and safer.

bergmania:
Have sometimes an issue about the `data-tmpimg` are not added to the image..

6246-missing-temp-location

warrenbuckley:
Super nice catch @bergmania hopefully after that call & debug/hack session we can mark this as merged once Azure has happy & built it all

src/Umbraco.Web/Editors/TinyMceController.cs
```diff
@@ -0,0 +1,101 @@
...
+        private IContentTypeBaseServiceProvider _contentTypeBaseServiceProvider;
+
+        public TinyMceController(IMediaService mediaService, IContentTypeBaseServiceProvider contentTypeBaseServiceProvider)
+        {
+            _mediaService = mediaService;
+            _contentTypeBaseServiceProvider = contentTypeBaseServiceProvider;
+        }
+
+        [HttpPost]
+        public async Task<HttpResponseMessage> UploadImage()
```
Shazwazza: @warrenbuckley there is no authorization assigned to this controller - this means that all back office users regardless of what they have access to can upload files. We probably want to limit the access of this controller to users that have access to content, media or member trees since those are the only places where the rte will exist?

src/Umbraco.Web/Editors/TinyMceController.cs
```diff
@@ -0,0 +1,101 @@
...
+                // Move the file from bodypart to a real filename
+                // This is what we return from this API so RTE updates img src path
+                // Until we fully persist & save the media item when persisting
+                // If we find <img data-temp-img /> data attribute
+                File.Move(currentFile, newFilePath);
+            }
+            catch (Exception ex)
+            {
+                // Could be a file permission ex
+                throw;
```
Shazwazza: Is there a reason to catch and then throw without doing anything? was this left over from testing?

src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
```diff
@@ -22,25 +24,74 @@ namespace Umbraco.Web.PropertyEditors
...
+                // Find all controls that use the RTE editor
+                var controls = grid.Sections.SelectMany(x => x.Rows.SelectMany(r => r.Areas).SelectMany(a => a.Controls));
+                var rtes = controls.Where(x => x.Editor.Alias.ToLowerInvariant() == "rte");
+
+                foreach(var rte in rtes)
+                {
+                    // Parse the HTML
+                    var html = rte.Value?.ToString();
+                    
+                    var userId = Current.UmbracoContext.Security.CurrentUser.Id;
```
Shazwazza: There is no guarantee this is being run inside of a request and therefore no guarantee of a user. This is why we don't like to use singletons because you can instantly run into problems. At the least you'll need a null check on the context, like `Current.UmbracoContext?.Security.CurrentUser.Id` ... but perhaps there's a nicer way to achieve all of this? not sure, will keep reading.

src/Umbraco.Web/PropertyEditors/RichTextPropertyEditor.cs
```diff
@@ -99,6 +112,11 @@ public override object FromEditor(Core.Models.Editors.ContentPropertyData editor
 
                 var editorValueWithMediaUrlsRemoved = TemplateUtilities.RemoveMediaUrlsFromTextString(editorValue.Value.ToString());
                 var parsed = MacroTagParser.FormatRichTextContentForPersistence(editorValueWithMediaUrlsRemoved);
+
+                var userId = Current.UmbracoContext.Security.CurrentUser.Id;
```
Shazwazza: Same issue as above, you need to at least null check UmbracoContext here since there's no guarantee this is running in a request.

src/Umbraco.Web/Editors/TinyMceController.cs
warrenbuckley: Will add a log message to help with debugging if anyone does run into a file exception & will change the exception type to be less generic catch all type

PR #6247, opened 2019-08-30
Loginpage: Added an id to input fields and a for attr on the labels so they are connected

fALKENdk:
Labels and input fields on the login page are not connected to each other, this is a accessibility problem.

So I have added an id to all the input fields and a for attr on the labels.


nul800sebastiaan:
Thanks!

Cherry picked for 8.1.5 da0ace18e69879471a55472e391aa7b98c96cb4d

PR #5635, opened 2019-06-10
v8: Extend the umb-checkbox component to handle localization

BatJan:
### Prerequisites

- [ ] I have added steps to test this contribution in the description below

This PR is currently work in progress - It's related to this issue: #5633

### Description

Description and images will follow once the PR is ready for rewview 😃

PR #6258, opened 2019-09-01
v8: Use umbCheckbox in packager

bjarnef:
### Prerequisites

### Description
This PR replaces the traditional checkboxes in the packager with `umb-checkbox` component used in many other parts of backoffice UI, e.g. in configuration of TinyMCE.
For some of the properties we might later re-consider if e.g. a tree picker is a better options for examples with many data types, templates etc... at the moment it could potentially be a very long list of checkboxes or we could consider if these should be collapsible?

For now I have re-placed these for `umb-checkbox` to be more consistent for the UI.

**Before**

**After**

Furthermore it seems `umb-checkbox` has been updated with scope properties, so e.g. disabled property is binded to `vm.disabled` while the ng-class evaluated `disabled`, so at the moment when setting `disabled="true"` on `umb-checkbox` it doesn't set the class `umb-form-check--disabled`.

This was broken in this commit https://github.com/umbraco/Umbraco-CMS/commit/320d2ae95356954e85ea671090833ba0544156a2 where ´umb-checkbox` it changed to a component and using `vm`, but it is missing to use `vm` as well in the ng-class expression for to set the disabled class.

Furthermore I think the radiobutton also should be changed to a component as very, as ´umb-radiobutton` and `umb-checkbox` component, should be as identical as possible.

I have fixes these issues in this PR as well.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
@@ -29,7 +29,7 @@
 @param {string} serverValidationField Set the <code>val-server-field</code> of the checkbox.
 @param {boolean} disabled Set the checkbox to be disabled.
 @param {boolean} required Set the checkbox to be required.
-@param {string} onChange Callback when the value of the checkbox change by interaction.
+@param {callback} onChange Callback when the value of the checkbox change by interaction.
 
 **/
 
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
@@ -49,7 +49,6 @@
         
     }
     
-    
     var component = {
         templateUrl: 'views/components/forms/umb-checkbox.html',
         controller: UmbCheckboxController,
@@ -63,7 +62,7 @@
             serverValidationField: "@",
             disabled: "<",
             required: "<",
-            onChange: "&"
+            onChange: "&?"
         }
     };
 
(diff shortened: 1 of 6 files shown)


bjarnef:
@nul800sebastiaan it would be great if this could be included in next release and fix the disabled styling on `umb-checkbox` and make `umb-radiobutton` consistent with this.

nul800sebastiaan:
Thanks @bjarnef - this will go into 8.2.0! 👍 💪

bjarnef:
@nul800sebastiaan maybe you can sherry pick a few of these commits to 8.1.5?

https://github.com/umbraco/Umbraco-CMS/commit/4b406155020e8838d04d9cf00b80d37a3aaf9dbc

nul800sebastiaan:
@bjarnef I may have some time next week! If you want to beat me to it then you can target a PR at the `v8/8.1` branch? 🙂

bjarnef:
@nul800sebastiaan I have submitted a PR https://github.com/umbraco/Umbraco-CMS/pull/6302 with cherry picked commits from this PR.

Furthermore I would like if we could make `umb-radiobutton` and `umb-checkbox` as identical as possible and the property editors using these components.

e.g. 

why are checkboxlist PE using one-way binding, while radiobuttons PE has two-way binding?
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/propertyeditors/checkboxlist/checkboxlist.html
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/propertyeditors/radiobuttons/radiobuttons.html

nul800sebastiaan:
I have no clue, maybe @BatJan can weigh in since he did the initial PRs for them.

BatJan:
Since I made the changes others have been working on the components as well it seems. When I find some spare time I will try to look into this.

However I also do have an open PR #5635 for them as well where it might make sense to address some of the concerns raised above and with the open PR I also address some usage patterns that are not desirable - Currently some places there are double labels etc. (Sorry, that's was a bit off topic but might be relevant in terms of how the implementation is made here? Have not read the commits yet).

The intention is to of course make them as reusable and flexible as possible... But not sure how much has been changed since the initial refactor. Need to get an overview.

PR #6262, opened 2019-09-02
Added GetAllElementTypes extension method to ContentTypeService

leekelleher:
### Prerequisites

### Description

Added an extension method for `ContentTypeService` to get all the Element Types.

> This is something I've developed for my own package. Thought it would be useful to share.
> https://github.com/leekelleher/umbraco-contentment/blob/6938fba139b45fab6d353027a6e2e3964cb12690/src/Umbraco.Community.Contentment/DataEditors/_/ContentTypeServiceExtensions.cs#L16-L19

I've done this as an extension method, as I wasn't sure whether adding to the concrete implementation would be considered a breaking-change or not?

This PR also updates NestedContent's controller for getting the Element Types.

I haven't added any unit-tests, etc.
_(To be honest, I haven't spent that much time on this - was more of a "I've coded this already, it could be useful in core?" - I could add unit-tests, etc - if it's deemed worthy enough 😄)_

### DIFF
--- a/src/Umbraco.Core/Services/ContentTypeServiceExtensions.cs
+++ b/src/Umbraco.Core/Services/ContentTypeServiceExtensions.cs
@@ -8,6 +8,21 @@ namespace Umbraco.Core.Services
     public static class ContentTypeServiceExtensions
     {
         /// <summary>
+        /// Gets all of the element types (e.g. content types that have been marked as an element type).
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Cool, I think it's indeed best as an extension method! Thanks Lee!

PR #6266, opened 2019-09-02
Update positioning of umb-overlay on resize/content-changes

ceee:
### Description

This PR fixes the positioning of overlays which are intialized with the `position: "target"` property.

The problem with the previous code was that once the overlay is displayed it uses the current dimensions for the calculation of the position. 
Sometimes the content has not been rendered yet when the position is already calculated:

image

Now it recalculates the position in case:

1. The contents of the overlay changes, so it's dimensions change.
2. The content-wrapper area is resized (due to a window resize or a tree resize)

It is using the `ResizeObserver` API which is implemented in (super) modern browsers ([stats](https://caniuse.com/#feat=resizeobserver)) and falls back to the legacy method in case it is not available.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/overlays/umboverlay.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/overlays/umboverlay.directive.js
@@ -208,8 +208,9 @@ Opens an overlay to show a custom YSOD. </br>
          var numberOfOverlays = 0;
          var isRegistered = false;
 
-          var modelCopy = {};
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hey @ceee 

Thanks for the work here and the fantastic documentation to go with it. It sure makes reviewing much easier. We'll take a look at the work and let you know if we need anything.

Emma

nul800sebastiaan:
Thanks @ceee - sorry for the delay on this one but this looks good and the code check out. 

Great job! 👍

Issue #6215, opened 2019-08-27
Replace Ace with Monaco Editor (VS Code)


rasmuseeg:
Hi i would like to suggest the change, to replace Ace with Monaco Editor instead.

I was able to get a sample running as a umbraco property in less than an hour.

Live running sample: https://microsoft.github.io/monaco-editor/

Samples: https://github.com/microsoft/monaco-editor-samples

Their api documentation is great: https://microsoft.github.io/monaco-editor/api/index.html

There is support for razor language.

We could over time add intellisense for models, methods etc.

Feature Comparision

IDE | Ace | Monaco
-- | -- | --
Autocomplete | Yes | Yes
Loader | Yes (AMD   loader) | Default   Loader.js is an AMD loader. Can use require.js
Themes | 20+ | 3
Code-folding | Yes | Yes
Multiple Cursur | Yes | Yes
**Diff Editor/Tool** | No | Yes *(Inline or split-view)*
Context Menu | *None*  | Yes
Command Palette | *None*  | Yes
Syntax Highligther | Yes | Yes
Browser Support | *Unable to find information* | IE 9/10/11, Edge, Chrome, Firefox, Safari and Opera *(Does not support mobile browsers. Scrolling does not work, writing jump etc.)*
Accessibility | No | Yes
Intellisense Support | Yes | Yes
Out-of-Box Intellisense | No | Yes
Live Syntax Checker | JavaScript/CoffeeScript/CSS/Xquery | TypeScript,   JavaScript, CSS, LESS, SCSS, JSON, HTML
**Syntax Colorization** | There   should be support for 110 modes. | XML, PHP, C#, C++,   Razor, Markdown, Diff, Java, VB,   CoffeeScript, Handlebars, Batch, Pug, F#, Lua, Powershell, Python, Ruby,   SASS, R, Objective-C, *(Maybe all 56)*
Languages | 110 | 56
**Rich IntelliSense, Validation**<br /> *(Syntax colorization plus errors, warnings, IntelliSense, formatting, outlining)* | None | **TypeScript, JavaScript, CSS, LESS, SCSS, JSON, HTML**
Github Stars | 20,300 | 16,792
Github Forks | 4,454 | 1,489
Github Used by | Unknown | 3,052
Github Watch | 640 | 467
Created At | Unkown | Unknown

nul800sebastiaan:
Could be lovely, especially with real C# support and possible future code completion.

[...]

PR #6267, opened 2019-09-02
#6215 Added basic monaco editor for stylesheet

rasmuseeg:
Providing a sample on how to integrate Monaco Editor. Keep open for discussion.

### DIFF
--- a/src/Umbraco.Web.UI.Client/gulp/tasks/dependencies.js
+++ b/src/Umbraco.Web.UI.Client/gulp/tasks/dependencies.js
@@ -19,23 +19,11 @@ gulp.task('dependencies', function () {
     // so we don't just ship with a lot of files that aren't needed
     const nodeModules = [
         {
-            "name": "ace-builds",
-            "src":  [
-                "./node_modules/ace-builds/src-min-noconflict/ace.js",
-                "./node_modules/ace-builds/src-min-noconflict/ext-language_tools.js",
-                "./node_modules/ace-builds/src-min-noconflict/ext-searchbox.js",
-                "./node_modules/ace-builds/src-min-noconflict/ext-settings_menu.js",
-                "./node_modules/ace-builds/src-min-noconflict/snippets/text.js",
-                "./node_modules/ace-builds/src-min-noconflict/snippets/javascript.js",
-                "./node_modules/ace-builds/src-min-noconflict/snippets/css.js",
-                "./node_modules/ace-builds/src-min-noconflict/theme-chrome.js",
-                "./node_modules/ace-builds/src-min-noconflict/mode-razor.js",
-                "./node_modules/ace-builds/src-min-noconflict/mode-javascript.js",
-                "./node_modules/ace-builds/src-min-noconflict/mode-css.js",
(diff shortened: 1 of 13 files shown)


rasmuseeg:
#6215

bjarnef:
Maybe we should rename the code editor to something independent on the library e.g. `umb-code-editor` similar to we now have `umb-range-slider` (noUiSlider) and `umb-date-time-picker` (Flatpickr).

bjarnef:
We should probably also change this to an Angular component and instead of use naming as "aceOption", it should be something more general like "options", "config" or similar.

nul800sebastiaan:
@rasmuseeg I'm really happy for you to keep hacking away on this, but we still need to decide if it worth it for HQ to switch the editor. Which is why we asked for a list of pros/cons on the related issue. 

Just letting you know that we haven't decided anything yet and I don't want to waste your time with building stuff that may not make it into Umbraco.

rasmuseeg:
@nul800sebastiaan Yes, i remember. I was making the draft to see how much needs to change, because i was not familiar with how it was structured.
Was also trying to get more familiar with the Monaco Editor and it's features, in order to later come up with the feature list for comparison.

@bjarnef Yes, i agree to some extend. By naming it monaco editor, we can reference to the monaco editor api. 
But if it is named umbraco-code-editor, we have to implement events, options, as methods or events inside the directive. I rather not build an API wrapper around the already great and well structured api.

bjarnef:
@rasmuseeg that is how other components are structured in Umbraco and it allow package developers the re-use this stuff, e.g. in a custom property editor, custom section or in a custom dashboard.

Furthermore I would prefer not naming it after the library, but based on its behaviour, e.g. `umb-code-editor`.

rasmuseeg:
@bjarnef Okay, well then we can add that to the pro/con list.
If we are to implement an API wrapper, we need to be picky about which options and events we want to implement.

bjarnef:
@rasmuseeg you already have the `umbMonacoEditor` directive, which is wrapped the monaco editor. This should probably be a angular component instead and renamed to something like `umbCodeEditor`. This component can implement the callback functions monaco editor has and can then be used in the different parts of backoffice in core or by (package) developers similar to other components in core.

rasmuseeg:
Yes, we can agree on that.
The callbacks already there, but the options and the editor is directly exposed through.
We can't just not wrap options and callbacks, and stille have the editor exposed through params.

bjarnef:
Instead of `vm.monacoEditorOptions` it can be more general like `vm.editorOptions` or similar.

zersiax:
One pro I can add is that, given this is not disabled, Monaco's accessibility mode is very mature due to it being battle-tested in Visual Studio Code. I myself use that as my daily driver and can confirm this works very well for a web-based editor.  TinyMCE is certainly usable, but for code I would definitely much prefer Monaco here.

bjarnef:
@zersiax I guess you meant Ace Editor not TinyMCE as code editor 😁

zersiax:
Hah :) I thought code editor was also tinyMCE, very much an Umbraco newbie still. Same goes for ACE though, Monaco would be a huge step up :)

Issue #4678, opened 2019-02-20
v8: Old password is not highlighted in validation


bjarnef:
When opening change password in user overlay and hit "change password" with blank fields, the validation is shown for each field. However for "old password" this isn't hightlighted.

![image](https://user-images.githubusercontent.com/2919859/53124574-f8901900-355b-11e9-840a-95a0213b319a.png)

The `control-group` seems to be missing the `error` class, which would make the validation looks like this:

![image](https://user-images.githubusercontent.com/2919859/53124646-31c88900-355c-11e9-8ab6-7f8e3d2d4b10.png)

nul800sebastiaan:
I've marked this as "Up for grabs" so that you or someone else coming along could create a pull request for it.

SoerenDeger:
It seems there is already fixed in dev-v8 branch?

![image](https://user-images.githubusercontent.com/7769446/54983044-52955b80-4fac-11e9-99d0-14bb08582943.png)

Shazwazza:
@bjarnef can you confirm?

bjarnef:
@Shazwazza yes, it seems this has been fixed in latest changes in `dev-v8` branch.

![image](https://user-images.githubusercontent.com/2919859/54998411-5b972480-4fce-11e9-92aa-17b415438ef5.png)

It is also fixes in Umbraco v8.0.1 release.

bjarnef:
However entering a value in "old password" doesn't remove the error validation on the label, but it is removed from the field.

![image](https://user-images.githubusercontent.com/2919859/54998566-bfb9e880-4fce-11e9-88f0-1ab9b4884f57.png)

bjarnef:
Oh, if I open the user dialog and hit "change password" with empty fields it show the validation error.

![image](https://user-images.githubusercontent.com/2919859/54998940-99e11380-4fcf-11e9-8c85-7140c509452f.png)

if I then close the dialog using the link "close" or using esc key and open the user dialog again, the validation of the form has been reset, but when now hitting "change password", it shows these validation errors.

![image](https://user-images.githubusercontent.com/2919859/54999169-f6443300-4fcf-11e9-9140-2da9aa783c17.png)

bjarnef:
I guess it has something to do for `formValid()` function which doesn't set the error class on subsequent validations of the form.

```
<div class="control-group umb-control-group" ng-class="{error: !formValid(), hidelabel:hideLabel=='true'}">
```

[...]

PR #6274, opened 2019-09-03
v8: Highlight of old password in change password form

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/4678

### Description
This PR fixes an issue where the old password field in change password form was not highlighted, because of the name of the form which returned `vm.passwordForm` and not the interpolated value `{{vm.passwordForm}}`, but `vm.passwordForm` is an object anyway. It seems this was changed here https://github.com/umbraco/Umbraco-CMS/blame/d2be30a228af13e34967e89fe11010749b1bad1b/src/Umbraco.Web.UI.Client/src/views/components/users/change-password.html#L19

I have changed this name to `changePasswordForm` since the parent regular form has the name `passwordForm` here: https://github.com/umbraco/Umbraco-CMS/blob/db6714d3b8f8e04dceb607c3a25117331f6b525d/src/Umbraco.Web.UI.Client/src/views/common/overlays/user/user.html#L98-L107

It seems the oldPassword field sometimes in highlighted and other times not. In user section on a user it seems always to be highlighted when invalid.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/overlays/user/user.html
+++ b/src/Umbraco.Web.UI.Client/src/views/common/overlays/user/user.html
@@ -24,10 +24,9 @@
             alias="changePassword"
             type="button"
             action="togglePasswordFields()"
-            button-style="action"
+            button-style="success"
             label="Change password"
-            label-key="general_changePassword"
-            button-style="success">
+            label-key="general_changePassword">
        </umb-button>
 
         <umb-button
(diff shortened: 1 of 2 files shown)


bjarnef:
I have fixes a few issue and removed some duplicate code.

The highlight of the old password field seems to work first time, but on close (esc or close link) and then re-open the overlay and submit the empty change password form, old password field doesn't seem to be highlighted.

If refreshing page it seems to work again.

nul800sebastiaan:
Thanks @bjarnef - it's better than before! Would be cool if you could find the cause of the re-validation problem.

Issue #6270, opened 2019-09-03
V8: Mandatory Multinode treepicker, when save & publish with no data, old data is removed


jsheard02:
Say we have a multinode treepicker, that is set to mandatory. Now, on a node using the mandatory multinode treepicker, we selected some number of items to populate that field. The Save&Publish event for that field is successful, and the content appears on the site. 

If we now go back to the node, and remove all items from the field, the Save&Publish event will fail with message "Value cannot be empty". Upon refreshing the page, the content previously selected in the multinode treepicker is not visible in the CMS, but the old content is still present on the site.

The problem is that there is a disparity between what is shown to the user in the CMS, and what is visible on the site.

Reproduction
------------

### Specifics

- Umbraco version 8.1.2
- Using Firefox 68.0.2, also confirmed behavior on Chrome Version 76.0.3809.132
- No error message is reported in the logs

### Steps to reproduce

- Add Multinode treepicker field on a node set to mandatory.
- Add content to the field and save and publish
- View output correct on frontend site
- Remove content from field and save and publish
- View error message
- Refresh page, and view content is not visible on field
- View output consistent with old data on frontend site

### Expected result

- Save & Publish event fails
- Upon refresh of the CMS, the field still contains the old data

### Actual result

- Save & Publish event fails
- Upon refresh of the CMS, the field contains no data

nul800sebastiaan:
It seems that the client-side validation doesn't kick in for the multinode treepicker when set to mandatory. 

You can work around this by configuring MNTP to require at least 1 selected item.

![image](https://user-images.githubusercontent.com/304656/64170230-e6b6f900-ce4f-11e9-834e-c89d390d2baf.png)

> the content previously selected in the multi treepicker is not visible in the CMS, but the old content is still present on the site.

This is correct, but also note that the content item is now in the "Saved" state instead of in the "Published" state, so this is consistent with saving changes without publishing them.

That said, it's a bug that there's no client-side checks for the "mandatory" configuration of an MNTP which should be fixed. If you can help then that would be great.

umbrabot:
Hi @jsheard02,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

[...]

PR #6282, opened 2019-09-04
V8: Fix mandatory validation clientside for Content Picker and MNTP

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6270

### Description

See #6270 for a description of the issue. Note that this applies to both MNTP and Content Picker properties.

With this PR applied, both property types expect at least one selected item before clientside validation passes:

mntp-content-picker-mandatory


nul800sebastiaan:
That's a lovely fix you got there @kjac - thank you kindly! 👍

PR #6290, opened 2019-09-06
V8: Make validation work properly with multiple invalid properties

kjac:
### Prerequisites

### Description

If you have multiple invalid fields in an editor in "submitted" state, the valid state of the fields doesn't *entirely* kick in until all validation errors of that specific type are corrected (e.g. all mandatory validations must be fulfilled before all fields appear completely valid).

It gets a lot worse when you add property editors that utilize `$setValidity` to do custom validation - for example MNTP with min/max count bounds.

Here's a GIF that shows all of the above:

validation-multiple-invalid-before

This PR ensures that the validation watcher aggregates all concrete errors and not just the different types of errors (e.g. mandatory, min-count). When it's applied, things work a tad more intuitively:

validation-multiple-invalid-after


nul800sebastiaan:
Nice one! Thanks @kjac!

PR #6292, opened 2019-09-06
Changes ManifestParser to be extendable

sitereactor:
This changes extends the `ManifestParser` class by making it `abstract` and providing a basic implementation, which allows us to setup dependency injection of the existing class. This means that we don't change any of the existing usages of the  `ManifestParser` class, but provide a way to replace the existing usages with a custom implementation.

In order to extend the `ManifestParser` you can provide your own implementation like this: 
`composition.RegisterUnique<ManifestParser, CustomManifestParser>();`.

This will allow you to load manifest files from a different source - ie. a database.
Ultimate goal here is to have everything from within the `App_Plugins` folder live in blob storage.

### DIFF
--- a/src/Umbraco.Core/Manifest/ManifestParser.cs
+++ b/src/Umbraco.Core/Manifest/ManifestParser.cs
@@ -12,24 +12,32 @@ using Umbraco.Core.PropertyEditors;
 
 namespace Umbraco.Core.Manifest
 {
+    internal class ManifestParserImpl : ManifestParser
+    {
+        public ManifestParserImpl(AppCaches appCaches, ManifestValueValidatorCollection validators, ManifestFilterCollection filters, ILogger logger)
+            : base(appCaches, validators, filters, logger)
+        {
+        }
+    }
+
     /// <summary>
     /// Parses the Main.js file and replaces all tokens accordingly.
     /// </summary>
-    public class ManifestParser
+    public abstract class ManifestParser
     {
         private static readonly string Utf8Preamble = Encoding.UTF8.GetString(Encoding.UTF8.GetPreamble());
 
-        private readonly IAppPolicyCache _cache;
-        private readonly ILogger _logger;
-        private readonly ManifestValueValidatorCollection _validators;
-        private readonly ManifestFilterCollection _filters;
+        protected readonly IAppPolicyCache Cache;
+        protected readonly ILogger Logger;
+        protected readonly ManifestValueValidatorCollection Validators;
+        protected readonly ManifestFilterCollection Filters;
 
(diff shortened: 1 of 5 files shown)


sitereactor:
Here is an example of a custom implementation of the ManifestParser, which loads the manifest files from a database through a service

```
public class CustomManifestParser : ManifestParser
{
    private readonly IVirtualAppPluginsService _virtualAppPluginsService;
    private readonly string _containerName;

    public CustomManifestParser(AppCaches appCaches, ManifestValueValidatorCollection validators, 
        ManifestFilterCollection filters, ILogger logger, IVirtualAppPluginsService virtualAppPluginsService) : base(appCaches, validators, filters, logger)
    {
        _virtualAppPluginsService = virtualAppPluginsService;
        _containerName = ConfigurationManager.AppSettings["Umbraco.Cloud.VirtualAppPlugins.BlobContainer"];
    }

    /// <summary>
    /// Gets all manifests, merged into a single manifest object.
    /// </summary>
    /// <returns></returns>
    public override PackageManifest Manifest
        => Cache.GetCacheItem<PackageManifest>("Umbraco.Core.Manifest.ManifestParser::Manifests", () =>
        {
            var manifests = GetManifests();
            return MergeManifests(manifests);
        }, new TimeSpan(0, 4, 0));

    /// <summary>
    /// Gets all manifests.
    /// </summary>
    protected override IEnumerable<PackageManifest> GetManifests()
    {
        var manifests = new List<PackageManifest>();
        var manifestsEntities = _virtualAppPluginsService.GetAsync().GetAwaiter().GetResult();
        foreach (var entity in manifestsEntities)
        {
            var manifest = ParseManifest(entity.Manifest);
            manifest.Source = entity.Name;
            manifests.Add(manifest);
        }

        Filters.Filter(manifests);

        return manifests;
    }
}
```

Shazwazza:
Hi @sitereactor there's an issue with this however:

> Ultimate goal here is to have everything from within the App_Plugins folder live in blob storage.

In many cases there will be cshtml files in these folders which *cannot* exist in blob storage, so that's something to consider. If you want to store cshtml in some virtual storage, you have to jump through a zillion hoops to make that work and it will be flimsy.

sitereactor:
Views would be picked up by a virtual path provider.
In my test setup that is already the case for the bits that Umbraco loads as Views.

Shazwazza:
@sitereactor ok but lets make sure we test perf of this because when loading in views not from the file system it will mean a lot of the caching mechanisms used by .net may be bypassed and they may be recompiled all the time or every time the app restarts (even a shallow restart).

PR #6295, opened 2019-09-06
V8: Make sure the validation error help is visible for Multi URL Picker

kjac:
### Prerequisites

### Description

The Multi URL Picker should show a red help text when its bounds (min/max) fail to validate, just like MNTP does. But it doesn't:

multi-url-picker-validation-helptext-before

Looks like it's a bit of a copy/paste error from MNTP, because the code is there to show the help text; it just doesn't work. But this PR fixes it all up:

multi-url-picker-validation-helptext-after


nul800sebastiaan:
Nice one, cheers!

PR #6296, opened 2019-09-06
V8: Make Multi URL Picker mandatory validation work clientside

kjac:
### Prerequisites

### Description

*This is the Multi URL Picker equivalent of #6282*

Multi URL Picker does not perform its mandatory validation on the clientside like it does its min/max bounds validation. The validation is performed serverside, but it would be more consistent to have it work clientside as well:

multi-url-picker-mandatory-before

This PR fixes the validation in exactly the same way as MNTP handles mandatory validation clientside:

multi-url-picker-mandatory-after

#### Testing this PR

1. Create a doctype with a mandatory Multi URL Picker property.
2. Ensure you can't create content of this type without picking at least one URL.
3. Update the Multi URL Picker so it requires at least two or more URLs picked.
4. Verify that the new minimum number of URLs is enforced when editing the content.


nul800sebastiaan:
Sweet, thanks again!! 👍

PR #6302, opened 2019-09-07
v8: umbCheckbox and umbRadiobutton fixes

bjarnef:
### Prerequisites

### Description
This PR cherry picks a few commits merged to `v8/dev` in https://github.com/umbraco/Umbraco-CMS/pull/6258

We should probably also change this comment in docs from `<code>id</code>` to `<code>input-id</code>`.

It also seems the documentation needs to be updated since none of these components are shown here https://our.umbraco.com/apidocs/v8/ui/#/api

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
@@ -29,7 +29,7 @@
 @param {string} serverValidationField Set the <code>val-server-field</code> of the checkbox.
 @param {boolean} disabled Set the checkbox to be disabled.
 @param {boolean} required Set the checkbox to be required.
-@param {string} onChange Callback when the value of the checkbox change by interaction.
+@param {callback} onChange Callback when the value of the checkbox change by interaction.
 
 **/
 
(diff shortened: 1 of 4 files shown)


nul800sebastiaan:
Thanks @bjarnef, you did beat me to it! 😁

I don't think the docs need an update, they are correct: the `id` attribute will be set, there is no `input-id` attribute. 🙂

nul800sebastiaan:
_Oh, and yes: we need to deploy the new docs.. it's on the list_

nul800sebastiaan:
Hmm, something is missing from both this and the #6258 PR. Currently, changes to checkboxes are not being saved at all. Radiobuttons work fine. Can I get you to have a look at that please @bjarnef ?

bjarnef:
Any specific places where it doesn't save checkbox changes? In packager it should save the changed selections when hitting save button?

bjarnef:
@nul800sebastiaan the packager seems to save the selected checkboxes.



But it seems there is an issue in checkboxlist property editor.

bjarnef:
@nul800sebastiaan the model passed in to `onChange` function is no longer an boolean, but an object.

So this condition is not equal true: https://github.com/umbraco/Umbraco-CMS/blob/b1674235e936401e9884364fb97eea3931518344/src/Umbraco.Web.UI.Client/src/views/propertyeditors/checkboxlist/checkboxlist.controller.js#L81

```
 function changed(model, value) {
     console.log("changed", model, value);
     ...
}
```

image

bjarnef:
@nul800sebastiaan it seems it is fixed now in this PR https://github.com/umbraco/Umbraco-CMS/pull/6345

I guess Angular was a bit confused by the `vm.onChange()` in the view and the callback with same name.

PR #6303, opened 2019-09-07
"Actions" is identified as a button, but behaves like a drop down

BatJan:
### Prerequisites

This fixes issue no. **8** and **44** from #5277 

### Description
Much of the groundwork for this was done in #5884 so the following things have been fixed
- Addition of `aria-haspopup` and `aria-expanded` attributes on the button that triggers the action dropdown
- The menu items have been changed from `<a>` to `<button>` elements
- the `aria-hidden="true"` attribute has been added to `<i>` and `<span>` elements that are used to display icons (This goes for all the possible outcomes in the umb-button.html view)
- The umb-botton directive has been extended so it's possible to add the above mentioned aria attributes by passing values for "isExpanded" and "hasPopup"
- The deep-blur directive has been added so the dropdown disappears when it's being tabbed out of using keyboard navigation (See the gif below)

**Before**
actions-dropdown-before

**After**
actions-dropdown-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbutton.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbutton.directive.js
@@ -67,6 +67,9 @@ Use this directive to render an umbraco button. The directive can be used to gen
 @param {boolean=} disabled Set to <code>true</code> to disable the button.
 @param {string=} addEllipsis Adds an ellipsis character (…) to the button label which means the button will open a dialog or prompt the user for more information.
 @param {string=} showCaret Shows a caret on the right side of the button label
+@param {string=} autoFocus add autoFocus to the button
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Thanks again @BatJan ! I know you still want to work on the `aria-haspopup` stuff but it's fine to do in a separate PR! 👍

BatJan:
@nul800sebastiaan Yes, but not for this PR. In this context everything is perfect 😃 

Since this is making a context menu visible the usage of aria-haspopup="true" is correct since the screen reader will announce something like "menu button".

But for #6304 and #6305 I was trying to be forward thinking and use a little progressive enhancement using the value of "dialog", which has been introduced with the ARIA 1.1 specification but, which screen readers are still implementing support. So for now NVDA will for instance announce it something like "menu button", which is not correct in this context since we're opening a dialog 😃 - PR to fix this is coming up shortly.

PR #6314, opened 2019-09-09
V8: Don't show Nested Content delete confirmation if an item can't be deleted

kjac:
### Prerequisites

### Description

The delete action isn't disabled when a Nested Content property has a minimum number of items allowed (but nothing happens after confirming the deletion):

nc-delete-bounds-before

Of course the delete confirmation dialog shouldn't be displayed to begin with, nor should the trash icon give any indication that it's active on hover. This PR fixes that; here's that same operation with this PR applied:

nc-delete-bounds-after

To be honest I would rather have hidden the delete action altogether when it's not applicable, but I'm not sure that's the best course of action here.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-nested-content.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-nested-content.less
@@ -154,6 +154,11 @@
 
 .umb-nested-content__icon--disabled {
     opacity: 0.3;
+    cursor: default !important;
+
+    &:hover {
+        color: @ui-option-type;
+    }
 }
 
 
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Hmm, I didn't know this was working this way. This seems like a pretty bad user experience, how am I supposed to know I can't delete items because there's a minimum required?

To me it would make more sense that you're able to delete whatever you want but when you try to save / publish you get a proper validation error.

Anyway, I see that it now works by forcing a minimum number of items in the view, so this solution seems good for now, seems that's how it was supposed to work.

Thanks Kenn!

nul800sebastiaan:
Cherry picked for 8.1.5 b0e788fdaa746ab0017d0fb180d61bed0c3d4a56

Issue #4646, opened 2019-02-19
v8: Fix up GlobalSettingsExtensions.IsReservedPathOrUrl


clausjensen:
Without caching, this does the full MVC route which means that we are double (triple, or more) running through the MVC routing engine on each request.

`//TODO: This value should be cached! Else this is doing double routing in MVC every request!`

https://github.com/umbraco/Umbraco-CMS/commit/02f8c4a106e9cdc19da38f90f6aba4578e9aca1a#diff-5424736481b632a8c9a23f1225a4dc47R131

PR #6317, opened 2019-09-10
Rebuilds PureLive models and notifies nucache of changes on a background thread when changing schema

Shazwazza:
The primary purpose of this change is to change the perceived performance of saving a document type (or other schema items like a Data Type) to be nearly 'instant'. 

The concern behind this is that in v8 when you save schema items and you are running in PureLive mode, the models must be rebuilt in order to update the nucache structure with the new models and currently all of this is done in the request thread that was used during the saving of a document type. This means that in some cases admins might be waiting for a much longer period than they anticipated when saving the document type.

In v7, the perceived performance doesn't affect saving a document type because in v7, the models and cache updates occur when a front-end request is executed after a schema item has changed. So the performance 'hit' occurs on a front-end request, not a back office request.

The changes in this PR move the model rebuild to a background thread so that the user that is saving a doc type doesn't need to wait for the rebuild. Then we use a lock for front-end requests to ensure that the requests are blocked until the rebuilding is done. This means that the performance 'hit' is no longer on a back office request and isn't specifically on a front-end request either - only that a front-end request may need to wait until the background operation is completed.

## How its done

* When either the ContentTypeCacheRefresher or the DataTypeCacheRefresher are executed, we no longer block the request by rebuilding models and notifying the cache in the request thread, instead we delegate this work to a new object `BackgroundPublishedSnapshotNotifier` which makes the normal calls to `_publishedModelFactory.WithSafeLiveFactory` and notifies the cache ... that part remains the same
* ModelsBuilder already takes care of ensuring new models with the above calls and ensuring that razor views are updated correctly, so there's no changes there
* A primary change however is that any front-end requests need to wait until the background operation is completed before they create PublishedSnapshot instances. If they don't wait, the created PublishedSnapshot will be created with out of date models whereas the razor views are compiled with the new models which means we'll get a `ModelBindingException`. To perform this wait, a new internal event is added to the UmbracoContext which is raised just before the lazily created PublishedSnapshot is created. In that event we block if there's a wait required and if it's a front-end request. This is also an important note - we don't block on back office requests because generally speaking there is no razor execution in the back office and the back office will happily run on a previous snapshot without any harm. This means that the back office requests like resyncing the tree after saving a doc type aren't blocked and execute as per normal.

## Other Notable changes

* Moves all reserved path detection and determining if it's a front-end request to a new class `RoutableDocumentFilter` ... Let me know if you hate this name. This is for a couple reasons: 
   * This fixes https://github.com/umbraco/Umbraco-CMS/issues/4646
   * the reserved path logic used to be in static methods and it was all ugly, 
   * now all of this logic is in one place, in a class without ugly statics
   * The way that route detection worked was totally broken if new routes were added dynamically at runtime
* We need to cache the result of `IsDocumentRequest` within the request since it may be called more than one time during the request so this is now an internal method call on the `UmbracoContext` which uses the `RoutableDocumentFilter` to do the check
* Mistakenly on startup with PureLive enabled we were forcing models to always regenerate from scratch which is bad for startup performance. MB already uses a caching system to use cached models on startup if they existed. This was simply an oversight, and a simple code change on startup from `publishedModelFactory.WithSafeLiveFactory(LoadCachesOnStartup);` to `LoadCachesOnStartup();` means we can use MBs model caches on startup if they exist. My tests show at least a 3 second improvement in startup time with a shallow boot (i.e. bumping web.config) when models already exist in PureLive mode.

## Testing

__Note:__ I have tested the changes made to the RoutableDocumentFilter.ReservedPaths (since this was moved from UmbracoModule.ReservedPaths). This collection deals with Identity providers and I have tested using a Google identity provider for the back office and have verified that this collection is populated correctly and is correctly used in the changes in RoutableDocumentFilter.IsReservedPathOrUrl

The test of the testing is:

* Code review
* Running in PureLive mode, go save a document type, and then refresh a front-end page probably within a couple seconds. You should see that the document type saves quite quickly AND that the tree is refreshed quite quickly (this proves that the back office requests are not blocked) and you'll see that the front-end page will take some time to render (most likely on a few seconds but sometimes up to 30 sec in some cases). You can also try saving a document type and then refreshing front-end pages on multiple tabs, you should see that all of these tabs will be waiting for the background operation to complete. - If you enable Debug logging, you will also see a log output for whenever a front-end page is waiting on the background operation: `Umbraco.Web.UmbracoModule` 
 "Request was suspended while waiting for background cache notifications to complete" 
* Run this test again but have the debugger attached and put a breakpoint on `ContentModelBinder.ThrowModelBindingException` ... when you refresh the front-end pages, we must ensure that this exception is not thrown. Without the debugger attached, if this is thrown we will essentially 'hide' it and automatically refresh the page, so this a rudimentary way to check if the exception is throwing or not.
* Run the same test when not using PureLive, the front-end requests will not block and should be more or less instant.

### DIFF
--- a/src/Umbraco.Web/UmbracoInjectedModule.cs
+++ b/src/Umbraco.Web/UmbracoInjectedModule.cs
@@ -13,6 +13,7 @@ using Umbraco.Core.Exceptions;
 using Umbraco.Core.Security;
 using Umbraco.Web.Composing;
 using Umbraco.Web.Cache;
+using Umbraco.Core.Models.PublishedContent;
 
 namespace Umbraco.Web
 {
@@ -38,6 +39,7 @@ namespace Umbraco.Web
         private readonly ILogger _logger;
         private readonly IPublishedRouter _publishedRouter;
         private readonly IUmbracoContextFactory _umbracoContextFactory;
+        private readonly IPublishedModelFactory _publishedModelFactory;
         private readonly BackgroundPublishedSnapshotNotifier _backgroundNotifier;
         private readonly RoutableDocumentFilter _routableDocumentLookup;
 
@@ -47,6 +49,7 @@ namespace Umbraco.Web
             ILogger logger,
             IPublishedRouter publishedRouter,
             IUmbracoContextFactory umbracoContextFactory,
+            IPublishedModelFactory publishedModelFactory,
             BackgroundPublishedSnapshotNotifier backgroundNotifier,
             RoutableDocumentFilter routableDocumentLookup)
         {
@@ -55,6 +58,7 @@ namespace Umbraco.Web
             _logger = logger;
             _publishedRouter = publishedRouter;
             _umbracoContextFactory = umbracoContextFactory;
+            _publishedModelFactory = publishedModelFactory;
             _backgroundNotifier = backgroundNotifier;
             _routableDocumentLookup = routableDocumentLookup;
         }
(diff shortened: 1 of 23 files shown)


src/Umbraco.Web/UmbracoInjectedModule.cs
```diff
@@ -99,9 +83,30 @@ private void BeginRequest(HttpContextBase httpContext)
             // ensure there's an UmbracoContext registered for the current request
             // registers the context reference so its disposed at end of request
             var umbracoContextReference = _umbracoContextFactory.EnsureUmbracoContext(httpContext);
+            umbracoContextReference.UmbracoContext.CreatingPublishedSnapshot += UmbracoContext_CreatingPublishedSnapshot;
```
zpqrtbnk: do we need to do all this even when not purelive?

Shazwazza: Not necessarily, when not in pure live we don't need to wait and any request that might execute while notifying is happening on a background thread will just serve the previous generation which is OK. I'll think about how we can encapsulate this stuff without having to have PureLive checks everywhere.

Shazwazza: Have pushed a change for this, though it rudimentary (i.e. not the prettiest) but it will do,

Shazwazza:
Closing this in favor of a new PR https://github.com/umbraco/Umbraco-CMS/pull/6347

PR #5487, opened 2019-05-18
v8: Add confirm overlay when deleting user account

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5464

### Description
This PR a confirm overlay to re-use and optionally specify a warning message and warning style.
Furthermore this doesn't add a new service, but just use the existing overlayService.

image

image

PR #6319, opened 2019-09-10
V8: Use an Umbraco confirmation dialog when deleting users

kjac:
### Prerequisites

### Description

When deleting a user we currently display a default browser confirmation instead of the custom Umbraco confirmation dialog:

delete-user-confirmation-before

And once the deletion has happened, the duplicate notification seems a bit excessive 😆 

This PR addresses both of the issues above; here's the same operation with this PR applied:

delete-user-confirmation-after

#### Testing note

You can't delete a user once it has been logged in... so in order to test this you have to create users to delete 😛

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/users/user.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/users/user.controller.js
@@ -1,7 +1,7 @@
 (function () {
     "use strict";
 
-    function UserEditController($scope, eventsService, $q, $timeout, $location, $routeParams, formHelper, usersResource, userService, contentEditingHelper, localizationService, notificationsService, mediaHelper, Upload, umbRequestHelper, usersHelper, authResource, dateHelper, editorService) {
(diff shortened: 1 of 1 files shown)


bjarnef:
@kjac yea, this has been open for a while now, but it would indeed be great to have this consistent confirm overlay 😁 https://github.com/umbraco/Umbraco-CMS/pull/5487

I wonder if it would be worth to have a general `confirm` overlay view with the option to specify a warning message as in my PR - the default overlay view is very simple. https://github.com/umbraco/Umbraco-CMS/blob/c4ba0454fc04bae4e388125c09833fc7f25cee6b/src/Umbraco.Web.UI.Client/src/views/common/overlays/default/default.html

emmaburstow:
Hey @kjac 

Thanks for the work here. This really improves the consistency of the UX. Much appreaciated. We'll let you know if it needs anything.

Em

nul800sebastiaan:
Thanks @kjac and @bjarnef - I do agree, that it would be much nicer to have a centralized confirm overlay that can be reused, so I'll take Bjarne's one over this one, so it can be built upon.

Issue #6162, opened 2019-08-21
Nested content copy choice label is just one letter


runegronkjaer:
When copying and then inserting content in the nested content data type, the label with the content name is just one letter and some dots. E.g. "C...". See screenshot from Chrome.

Umbraco v 8.1.3

Reproduction
------------
1. Make a nested content data type and add it to a documenttype
2. Create some content
3. Copy the content
4. Click "Add Content" again and see how the content is named

### Expected result
The label should be a text describing the item I have copied

### Actual result
The label shows one letter and some dots. See screenshot
![2019-08-21 10_01_05-Window](https://user-images.githubusercontent.com/3272628/63413779-b35d7e80-c3fa-11e9-986b-9696f21951ba.png)

---
_This item has been added to our backlog [AB#2347](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2347)_

nul800sebastiaan:
That seems to be a problem in your particular site, I can't reproduce this problem in a fresh 8.1.3 install.

Let us know if when you have more info to reproduce!

![image](https://user-images.githubusercontent.com/304656/63416813-6f6d7800-c400-11e9-81d7-31db4b3b7429.png)

runegronkjaer:
Are you sure? Your screenshot is not of the copy feature.
This is where the issue is:
![copy label](https://user-images.githubusercontent.com/3272628/63418066-c1170200-c402-11e9-8939-9fd50ea933ad.png)

nul800sebastiaan:
My screenshot was supposed to be an animated gif, oops!

![6162](https://user-images.githubusercontent.com/304656/63418173-eefc4680-c402-11e9-98c2-32658ee0fdb3.gif)

runegronkjaer:
Ok. I have got news on this. After many tests I have found out that the problem disappears if I delete the "UmbracoForms" folder in \App_Plugins. So I guess I'll have to move the issue to Umbraco Forms instead? It's most likely overwriting a truncate filter or something like that.

nul800sebastiaan:
Yikes!

nul800sebastiaan:
No, no need to move it.. this just means that we need to anticipate other people doing things that might affect core as well.. Great find! Really unfortunate.. we'll have to have a look how to avoid this problem.

nielslyngsoe:
PR: https://github.com/umbraco/Umbraco-CMS/pull/6338

PR #6338, opened 2019-09-13
Fixes v8: Correct label filter is applied when copying NC

elit0451:
Fixes: #6162 

The reason for the issue was that once Forms are installed `truncate` filter on Forms was overwriting the one on the CMS which required different parameters. So, after talking with @nielslyngsoe, we agreed on changing the name of the filter on the CMS, and thus resolving the problem.

Testing:
- Create a nested content data type or use an existing one in the following step
- Add it to a document type
- Create some content
- Copy the content
- Click "Add Content" again and see how the content is named

Result:
- You should be able to see the full name of the copied item, or part of it and then "..." if it is too long

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/filters/truncate.filter.js
+++ b/src/Umbraco.Web.UI.Client/src/common/filters/truncate.filter.js
@@ -4,28 +4,52 @@
  * @namespace truncateFilter
  * 
  * param {any} wordwise if true, the string will be cut after last fully displayed word.
- * param {any} max max length of the outputtet string
- * param {any} tail option tail, defaults to: ' ...'
- *
+ * param {any} max where to cut the string.
+ * param {any} tail option tail, defaults to: '…'
+ * 
+ * Legacy version:
+ * parameter noOfChars(wordwise) Where to cut the string.
+ * parameter appendDots(max) If true dots will be appended in the end.
+ * 
  * @description
  * Limits the length of a string, if a cut happens only the string will be appended with three dots to indicate that more is available.
  */
 angular.module("umbraco.filters").filter('truncate', 
     function () {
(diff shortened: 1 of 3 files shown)


bjarnef:
I think the `truncate` filter in Umbraco Forms should be renamed instead then if is isn't possible to use the existing `truncate` filter in Umbraco.

Developers might be using the `truncate` filter in Umbraco in property editors, dashboard, etc.

nielslyngsoe:
Hi @bjarnef 

Thanks for sharing your concerns regarding this one.

I agree, and it was also my first impression, the thing is that we do not limit people from using an old version of Forms with newer versions of the CMS, or the other way around.

Your comment did though make us rethink our options, cause we would like not to break the experience for anyone, meaning that we will be making the CMS truncate-filter able to work with the arguments of Umbraco-Forms as well as its current arguments. Updated code will follow.

src/Umbraco.Web.UI.Client/src/common/filters/truncate.filter.js
```diff
@@ -4,28 +4,52 @@
...
  * @description
  * Limits the length of a string, if a cut happens only the string will be appended with three dots to indicate that more is available.
  */
 angular.module("umbraco.filters").filter('truncate', 
     function () {
         return function (value, wordwise, max, tail) {
+            
             if (!value) return '';
+            
+            // Overload-fix to support Forms Legacy Version:
```
Shazwazza: IMO I think we need more code comments in here to indicate exactly what the heck is going on here. If another developer finds this code they will have no idea what this means and what it is doing. For any hacks like this we need to ensure we have a bunch of code comments in there to make it clear to other developers what is going on. You can include links to these issues too.

Shazwazza:
tested and works

PR #6342, opened 2019-09-15
Fix some accessibility misunderstandings

BatJan:
### Prerequisites

### Description
This PR fixes some accessibility misunderstandings on my part for some of PR's that has gotten accepted into the core already.

1. The usage of `aria-haspopup="dialog"` will not make screen readers announce that a dialog will be visible currenlty - NVDA for instances will announce that it's a menu. So for now it's better to use the ellipsis convention. Therefore I have added a `<span class="sr-only">...</span>` after having a good chat with Florian Beijers about this. He suggested adding the text "(Opens dialog)" since the ellipsis approach is not used much on the web but mostly in software - But then we agreed that while "(Opens dialog)" would be nice it was less work to use the ellipsis since it does not require translation
2. Which leads me to the next fix addressed in this PR - I removed the `aria-hidden="true"` attribute from the ellipsis since I realized that it's of course very relevant to have them announced letting screen reader users know that it will popup some kind of dialog - HQ got this correct from the start. #h5is 😄

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/buttons/umb-button-group.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/buttons/umb-button-group.html
@@ -46,7 +46,7 @@
             hotkey-when-hidden="{{subButton.hotKeyWhenHidden}}"
             prevent-default>
                <localize key="{{subButton.labelKey}}">{{subButton.labelKey}}</localize>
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Super, thanks @BatJan ! ⭐️

Issue #6224, opened 2019-08-29
Error upgrading due to SecurityStamp being null


arknu:
I was upgrading a 7.2.8 site to 7.15.2 and had the usual problem of not being able to log in. So I started to investigate. It turns out the problem is this line:
```
if (HasClaim(x => x.Type == Microsoft.AspNet.Identity.Constants.DefaultSecurityStampClaimType) == false)
    AddClaim(new Claim(Microsoft.AspNet.Identity.Constants.DefaultSecurityStampClaimType, SecurityStamp, ClaimValueTypes.String, Issuer, Issuer, this));
````

On upgrading the SecurityStamp is null, which causes the Claim constructor to throw an exception. And in turn this causes the login to fail.

I fixed this by modifying the above code this this:
````
if (HasClaim(x => x.Type == Microsoft.AspNet.Identity.Constants.DefaultSecurityStampClaimType) == false)
{
    if (ApplicationContext.Current.IsUpgrading && SecurityStamp == null)
    {
        AddClaim(new Claim(Microsoft.AspNet.Identity.Constants.DefaultSecurityStampClaimType, Guid.Empty.ToString(), ClaimValueTypes.String, Issuer, Issuer, this));
    }
    else
    {
        AddClaim(new Claim(Microsoft.AspNet.Identity.Constants.DefaultSecurityStampClaimType, SecurityStamp, ClaimValueTypes.String, Issuer, Issuer, this));
    }
}
````

What this does is just use an empty GUID in the case of this being an upgrade and the SecurityStamp being null. This allowed the upgrade to proceed normally.

Is this an acceptable solution. If so, I can convert it into a pull request.

Reproduction
------------
Take an old Umbraco site (in this case 7.2.8) and try upgrading to 7.15.2. You cannot log in to the upgrade installer to start the upgrade.

### Expected result
Log in and proceed normally with upgrade

### Actual result
Login fails and you cannot proceed with upgrade

nul800sebastiaan:
Seems acceptable to me if you want to make a PR for it! 👍

arknu:
@nul800sebastiaan Thanks, will do.

One other thing. After running the upgrade, I noticed that the SecurityStamp column in the umbracoUser table was blank. Isn't the upgrade supposed to fill these in? I was unable to log in until I manually added a GUID in this column. I don't know if these two issues are related somehow...

nul800sebastiaan:
Unfortunately, I have absolutely no idea. 😅 

[...]

PR #6343, opened 2019-09-15
V7: Error upgrading due to SecurityStamp being null

stevemegson:
Implements the two suggested fixes for #6224 to enable upgrades from pre-7.3.0 to 7.15.x:

- Allow login without a SecurityStamp to authorise upgrade
- Fill in any NULL security stamp values with a migration

### DIFF
--- a/src/Umbraco.Core/Persistence/Migrations/Upgrades/TargetVersionSevenFifteenFour/PopulateMissingSecurityStamps.cs
+++ b/src/Umbraco.Core/Persistence/Migrations/Upgrades/TargetVersionSevenFifteenFour/PopulateMissingSecurityStamps.cs
@@ -1,0 +1,23 @@
+﻿using Umbraco.Core.Logging;
+using Umbraco.Core.Persistence.SqlSyntax;
+
+namespace Umbraco.Core.Persistence.Migrations.Upgrades.TargetVersionSevenFifteenFour
+{
+    [Migration("7.15.4", 1, Constants.System.UmbracoMigrationName)]
+    public class PopulateMissingSecurityStamps : MigrationBase
+    {
+        public PopulateMissingSecurityStamps(ISqlSyntaxProvider sqlSyntax, ILogger logger) : base(sqlSyntax, logger)
+        {
+        }
+
+        public override void Up()
+        {
+            // A user with a NULL securityStampToken can't log in after v7.8.0
+            Execute.Sql(@"UPDATE umbracoUser SET securityStampToken = NEWID() WHERE securityStampToken IS NULL");
+        }
+
+        public override void Down()
+        {
+        }
+    }
+}
--- a/src/Umbraco.Core/Persistence/Migrations/Upgrades/TargetVersionSevenFifteenFour/PopulateMissingSecurityStamps.cs
+++ b/src/Umbraco.Core/Persistence/Migrations/Upgrades/TargetVersionSevenFifteenFour/PopulateMissingSecurityStamps.cs
@@ -1,4 +1,5 @@
-﻿using Umbraco.Core.Logging;
+﻿using System;
+using Umbraco.Core.Logging;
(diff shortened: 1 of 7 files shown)


emmaburstow:
Hey Steve, 

You've been busy! Thanks for the work. We'll shout if we need you.

Emma

nul800sebastiaan:
I fixed the unit tests, we need an `UmbracoContext` to be initialized before we can ask for `pplicationContext.Current`.

nul800sebastiaan:
Ah, this is so awesome but there's just ONE little hitch here! Once the upgrade is done, we try to redirect the user to the backoffice, but the user's data is at that point not being reloaded from the database and `AddUserDataClaims()` still fails ([here](https://github.com/umbraco/Umbraco-CMS/pull/6343/files#diff-8ba1426c8f760ba7e8733e11fcaef667R238)) because `SecurityStamp` is null. When you then manually go to `/umbraco` it shows the login screen and as soon as you enter your credentials, you CAN login (🎉). 

Any change you can figure out how to reload the user's data after `PopulateMissingSecurityStamps` has executed @stevemegson?

Great work by the way, soon people will no longer get stuck on that 7.8.0 upgrade "hump". 👍👍👍

src/Umbraco.Core/Persistence/Migrations/Upgrades/TargetVersionSevenFifteenFour/PopulateMissingSecurityStamps.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class PopulateMissingSecurityStamps : MigrationBase
+    {
+        public PopulateMissingSecurityStamps(ISqlSyntaxProvider sqlSyntax, ILogger logger) : base(sqlSyntax, logger)
+        {
+        }
+
+        public override void Up()
+        {
+            // A user with a NULL securityStampToken can't log in after v7.8.0
+            Execute.Sql(@"UPDATE umbracoUser SET securityStampToken = NEWID() WHERE securityStampToken IS NULL");
```
Shazwazza: v7 still supports mysql right? I'd suggest using a c# Guid.NewGuid() here instead of using SQL to create the guid

nul800sebastiaan: Ah, good catch! 👍

stevemegson: Good point. Do we need to do individual updates to give each user a different GUID, or is it safe to let everyone start with the same Guid.NewGuid()?

stevemegson:
Ah, the null value is being stored in the auth cookie, and after the upgrade completes we still have to be able to read that cookie. I've moved the null check one step earlier, so we populate SecurityStamp with a GUID rather than just handling the null value.

src/Umbraco.Core/Persistence/Migrations/Upgrades/TargetVersionSevenFifteenFour/PopulateMissingSecurityStamps.cs
Shazwazza: It is safe to have everyone use the same one, it is not used as a public check for anything, it is just a value to indicate the current representation of the user, so that when it ever changes, the system knows that it needs to clear all authn/authz cookies/etc..

nul800sebastiaan:
I bumped the version to 7.15.4 to make it easier to test this PR. Should be done at some point anyway, so it's fine to do it in the PR :-)

nul800sebastiaan:
Thanks @stevemegson - this will make a lot of people happy 😁

elit0451:
The upgrade works :heavy_check_mark: therefore merging!

PR #6344, opened 2019-09-15
V8: Redirect tracking on move

### LINKED ISSUE #5050: Moving node doesn't create redirect.
When you move the node to another level, URL redirect is not created. 
How to reproduce? 

1. Install standard Starter Kit 
2. Move About Us under People
3. Check Redirects and you will see they haven't been created. 

Tested this on latest CMS - 8.0.1

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #5050

### Description
RedirectTrackingComponent has to go to some effort to work properly with the order of events in V7. 

The order has changed in V8, and redirects are no longer created when a page is moved (because the Publishing event is no longer raised during a move).

The new event order allows the logic to be considerably simplified, recording the old routes in `Publishing` and `Moving`, then creating the redirects in `Published` and `Moved`.

Testing:
- Move a published node and confirm that redirect tracking records the change
- Rename a published node and confirm that redirect tracking records the change (just to check that it still works)

### DIFF
--- a/src/Umbraco.Web/Routing/RedirectTrackingComponent.cs
+++ b/src/Umbraco.Web/Routing/RedirectTrackingComponent.cs
@@ -1,17 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using Umbraco.Core.Cache;
 using Umbraco.Core.Composing;
 using Umbraco.Core.Configuration.UmbracoSettings;
 using Umbraco.Core.Events;
 using Umbraco.Core.Models;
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hiya Steve,

As ever, we'll take a look and let you know if we need anything more from you. Thanks for the testing steps. This really makes life easier for us :)

Emma

nul800sebastiaan:
Excellent! Thanks very much @stevemegson1 ⭐️

PR #6350, opened 2019-09-16
V8: Macro container rendering and caching

stevemegson:
The macro container property editor always renders as `Umbraco.Web.Macros.MacroContent` because it calls `ToString()` rather than `GetAsText()`. It also has the same caching issues that were fixed for the RTE in #6010.

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/ValueConverters/MacroContainerValueConverter.cs
+++ b/src/Umbraco.Web/PropertyEditors/ValueConverters/MacroContainerValueConverter.cs
@@ -57,16 +57,16 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
                         macroAlias,
                         umbracoContext.PublishedRequest?.PublishedContent,
                         //needs to be explicitly casted to Dictionary<string, object>
-                        macroAttributes.ConvertTo(x => (string)x, x => x)).ToString()));
+                        macroAttributes.ConvertTo(x => (string)x, x => x)).GetAsText()));
 
                 return sb.ToString();
             }
          }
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hey Steve,

Thanks also for this. Once again, we'll take a look and let you know if we need anything.

Thanks,

Emma

nul800sebastiaan:
Hmm, how do we get a Macro Container? Is it by upgrading from an old v7 where it was still available? 🙈

stevemegson:
@nul800sebastiaan does `showDeprecatedPropertyEditors` still work in V8 to let you create one? The flag is still around in the code, I assume it's still wired up to the configuration file and not just hardcoded false.

nul800sebastiaan:
It (horrifyingly..) does still work. The MacroContainer, however, no longer works 

image

Even when I do drop in the old view, most macros don't work because the macro parameters don't resolve any more. 

Looks like we had every intention of completely killing this datatype. I did try this one out after fixing most problems and it KINDA works, but it's a pretty bad experience. I would highly advise you re-implement this as a custom datatype if you absolutely must use it.

The alternative is of course to drop macros in a grid or in an RTE. 

As we don't want to have this datatype in our codebase any more, I will close this PR and not accept the changes (sorry!). I'll create a new issue for properly deleting the old code.

stevemegson:
Totally on board with removing the datatype entirely instead. I only discovered that it still existed while working on the RTE caching, and thought I'd better apply the same fix. Not sure how I didn't notice that the editor is dead, I must have just been testing rendering with content migrated from V7.

PR #6352, opened 2019-09-16
V8: Reload content after rollback

### LINKED ISSUE #6349: Rollback - Umbraco 8
Hi everyone, 

Rollback function in Umbraco version 8.1.2 doesn't work properly.

Reproduction
------------
When I choose version to which I want to return back current document, content of the page isn't updated with data from selected version - still old content are presented. Even I tried to use "Preview", "Save" or "Save and publish..." buttons, content remains the same like it was before (current-old version).

But, if I choose an version to which I want to rollback and just refresh the page in browser or click on any other node in navigation content tree from the left side, and click back again on current document node, content will be filled-in with correct data from chosen rollback version.

### Bug summary

I found the similar issue for an older version of Umbraco:
https://issues.umbraco.org/issue/U4-7261

### Specifics

Browser: Version 76.0.3809.132 (Official Build) (64-bit) normal and incognito mode

### Steps to reproduce

* Click on node for which you want to rollback the content
* Go to "Info" tab
* Click on "Rollback..." button
* Choose "Rollback to" which version you want to rollback
* Click on "Rollback" green button at the bottom-right corner.
* Go to "Content" tab
* Take a look on content of the document or
* Click on "Preview" button.

### Expected result

When I choose rollback version of current document and switch to content tab, I expect that content is filled-in with data from chosen version. As confirmation of this, I would see with "Preview" option updated content and don't need to refresh the page or switch to another node and come back to actual one.

### Actual result

When I choose rollback version of current document and switch to content tab, content data remains the same (from current - previous version).
In preview I also can see the same (old) version, not from chosen rollback version.

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #6349

### Description
The rollback dialog raises an event when it has made changes, but the content editor did not bind to that event.

Testing:
- Roll back a document to a previous version. The change should be immediately reflected in the editor, breadcrumbs, and tree.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
@@ -171,6 +171,14 @@
                 }
             }));
 
+            evts.push(eventsService.on("editors.content.reload", function (name, args) {                
+                if (args && args.node && $scope.content.id === args.node.id) {
+                    reload();
+                    loadBreadcrumb();
+                    syncTreeNode($scope.content, $scope.content.path);
+                }
+            }));
+
         }
 
         /**


emmaburstow:
Hiya Steve,

Thanks for the contribution here. We'll take a look and get back to you if we have any questions about your work.

Thanks,

Emma

nul800sebastiaan:
Haha it took me forever to figure out why I couldn't roll back from the right-click menu any more 🙈 

image

Right, this works great, thanks very much @stevemegson ! ⭐️

stevemegson:
> it took me forever to figure out why I couldn't roll back from the right-click menu any more

@nul800sebastiaan that makes two of us 😃

Emetico:
I was told this was released in 8.2.0, but it still persist in 8.3.0

stevemegson:
@Emetico I can't reproduce this. Could you have a pre-8.2.0 version of the javascript cached somewhere and still being used after the upgrade? Can you reproduce the problem in a fresh install of 8.3.0?

Emetico:
You are right, i somehow had a cached version.

narinkaew:
I got the same problem, the page doesn't reload after click rollback. It has to refresh manually (F5).

My umbraco version is 8.1.0. But there is no structure files same as your file changed
<code>src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js</code>

MY folder structure as below like this, same as others version. 
image

Could you please advise me how to solve this issue with my current structure?

Thanks in advance
narinkaew

PR #6356, opened 2019-09-17
V8: Tours - fix scrolling highlighted element into view

### LINKED ISSUE #5954: V8: Tours, popup hidden with no scrolling
Tour, `Data structure/Document types` step 12 does not scroll. Can not proceed unless you zoom out.

### Bug summary

When taking the Tour for Umbraco CMS 8, I found myself semi-stuck when doing `Data structure/Document types` step 12. The popup did not auto scroll down and did not allow scrolling manually. The only way I could get around was to go from `zoom 100%` to `zoom ~50%`.

### Specifics

Umbraco version 8.1.0
Tested in Microsoft Edge 44.18362.1.0 and Chrome Version 75.0.3770.142
`NB: No adblocker or script blocking addons are installed.`

#### Zoom 100% in Edge
zoom 100%

#### Zoom 50% in Edge
zoom 50%

### Steps to reproduce

1. Use standard Umbraco nuget packaged and use default template when asked in the setup process.

2. Follow the Tour for `Data structure/Document types` and proceed to step 12.

### Expected result

Should scroll down to the focused dialog.

### Actual result

Stuck at the top of the page unable to manually scroll down.

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #5954

### Description
Tours include code to scroll the highlighted element into view, but it incorrectly calculates the position of the highlighted item when the element's offsetParent is inside its scrollParent.

This happens in step 12 of the document types tour because property groups are position:relative.

Testing:
- View the Data structure/Document types tour
- On step 12, the page should scroll down to make the highlighted 'Add Property' button visible.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/application/umbtour.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/application/umbtour.directive.js
@@ -327,10 +327,21 @@ In the following example you see how to run some custom logic before a step goes
                     }
 
                     var scrollParent = element.scrollParent();
-                    var scrollToCenterOfContainer = element[0].offsetTop - (scrollParent[0].clientHeight / 2 ) + (element[0].clientHeight / 2);
+                    var el = element;
+                    var offsetTop = 0;
+                    if (scrollParent[0] === document) {
+                        offsetTop = el[0].offsetTop;
+                    } else {
+                        while ($.contains(scrollParent[0], el[0])) {
+                            offsetTop += el[0].offsetTop;
+                            el = el.offsetParent();
+                        }
+                    }
+                    
+                    var scrollToCenterOfContainer = offsetTop - (scrollParent[0].clientHeight / 2) + (element[0].clientHeight / 2);
 
                     // Detect if scroll is needed
-                    if (element[0].offsetTop > scrollParent[0].clientHeight) {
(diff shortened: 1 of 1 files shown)


sebkirller:
Hello @stevemegson,
I am the one who submitted the issue. Your solution work nicely on my laptop. However, a slight improvement could be made at step 24, where it scrolls to far down resulting in unreadable info text.

image

Maybe a solution could be to check if the container is bigger then the screen and then scroll to the top instead of the center of the container?

stevemegson:
@sebkirller it looks like my window was just tall enough to avoid triggering scrolling for that step. I've fixed that, and also changed the logic for alignment of the popover - I was seeing it align to the bottom when the bottom of the highlighted element is out of view.

sebkirller:
> @sebkirller it looks like my window was just tall enough to avoid triggering scrolling for that step. I've fixed that, and also changed the logic for alignment of the popover - I was seeing it align to the bottom when the bottom of the highlighted element is out of view.

@stevemegson This fixes the immediate issue on my laptop as far as I can tell :+1:.

nul800sebastiaan:
Looking good! We'll review soon, thanks so far! 👍

nul800sebastiaan:
Yay, super nice fix, thanks @stevemegson !

PR #6365, opened 2019-09-18
Content items must prevent saving while image upload is in progress

warrenbuckley:
* Emit JS event when an upload of an image starts & stops
* Listen for events in content edit JS controller to change the state of the buttons to be 'busy' to avoid page/node being saved whilst a large image is uploaded & or using a slow connection

## Test Notes
* Open Chrome DevTools & set the network connection to **Slow 3G**
* Upload an image in the RTE (drag & drop from your computer in)
* Verify that you are shown a notification about saving being disabled
* Verify that the buttons change to the busy state when uploading
* Verify the buttons change to the success tick once the image finished uploading
* Try with one or more images

Fixes [AB#2727](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2727)


warrenbuckley:
@Shazwazza or @nielslyngsoe any pointers on to help me fix up the JS unit text I broke please?!

src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
```diff
@@ -171,6 +171,18 @@
                 }
             }));
 
+            evts.push(eventsService.on("rte.file.uploading", function(){
+                notificationsService.info("Saving Disabled", "The node cannot be saved until the image/s has finished uploading")
```
Shazwazza: This is actually not necessary and i think might get annoying, turns out that TinyMCE has it's own image upload progress, but this can only be seen when running on a slower connection, if you throttle to slow 3g, you'll see it, example:

image

src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
```diff
@@ -167,6 +167,14 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
         xhr = new XMLHttpRequest();
         xhr.open('POST', Umbraco.Sys.ServerVariables.umbracoUrls.tinyMceApiBaseUrl + 'UploadImage');
 
+        xhr.onloadstart = function(e) {
+            eventsService.emit("rte.file.uploading");
```
Shazwazza: these need to be wrapped in a digest else the angular app doens't actually respond since these events are triggered outside of an angular digest. Safest thing to do is wrap this in an `angularHelper.safeApply`. If you don't, you'll notice that the state of the buttons doesn't change when you first start uploading until you move the mouse around, or similar.

warrenbuckley:
@Shazwazza done the fixes suggested and will move your deletion bug to a new issue/PR as I don't think it is related to these changes at all

Issue #6367, opened 2019-09-18
v8 Migration throws error on old related link


benjaminc:
If you have a related link that is still using an ID to reference a node in the database, the v8 migration has code to handle that, but the code throws an exception because it is missing a FROM clause.

Reproduction
------------
### Bug summary
When migrating a v7 site to v8, and the v7 site has a related links picker referencing data using the old ID method, the migration throws an exception in the ConvertRelatedLinksToMultiUrlPicker migration.

### Specifics

This occurs because the SELECT query on line 73 of the ConvertRelatedLinksToMultiUrlPicker.cs file doesn't include a FROM clause.

### Steps to reproduce

Create a v7 site, and add a Related Links picker.  You can either use an old v7 version site and upgrade it to the latest v7 after picking content, or use a new v7 version site and just manually change the database to have an ID in the link picker's content in the cmsPropertyData table. 
 Once you have the ID in the related picker link field, run the v8 migration.

### Expected result

The upgrade should succeed.

### Actual result

The upgrade fails, and the following error is shown in the logs and on the upgrade screen.

System.Data.SqlClient.SqlException (0x80131904): The multi-part identifier "umbracoNode.id" could not be bound.
The multi-part identifier "umbracoNode.id" could not be bound.
The multi-part identifier "umbracoNode.uniqueId" could not be bound.
The multi-part identifier "umbracoNode.parentId" could not be bound.
The multi-part identifier "umbracoNode.level" could not be bound.
The multi-part identifier "umbracoNode.path" could not be bound.
The multi-part identifier "umbracoNode.sortOrder" could not be bound.
The multi-part identifier "umbracoNode.trashed" could not be bound.
The multi-part identifier "umbracoNode.nodeUser" could not be bound.
The multi-part identifier "umbracoNode.text" could not be bound.
The multi-part identifier "umbracoNode.nodeObjectType" could not be bound.
The multi-part identifier "umbracoNode.createDate" could not be bound.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
[...]

PR #6368, opened 2019-09-18
Add missing FROM clause

benjaminc:
### Description

These changes fix the issue identified in #6367.  To reproduce that problem, you need to have a v7 site with a Related Links picker with data that still references a content node by its ID, rather than a UDI.  This can often happen if you upgrade a previous v7 site with a Related Links picker (and picked data on a content node) to the latest v7 version.  Attempting to migrate the v7 site to v8 results in an exception in the ConvertRelatedLinksToMultiUrlPicker migration.

The problem was that a SQL query in the class which attempts to resolve ID references did not include a FROM clause.  The fix was to add the FROM clause.


nul800sebastiaan:
Aha! That's quite a crucial part of the query to include yeah! 😉

Thanks for the fix! 👍

PR #6372, opened 2019-09-19
Improve content pasted from MS Word

nielslyngsoe:
This PR improves the state of content pasted from MS Word. This PR cleans the HTML, so you end up having a more clean HTML-code when pasting content from Word.


src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
```diff
@@ -332,7 +347,35 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
                     images_replace_blob_uris: false,
                     init_instance_callback: initEvents
                 };
-
+                
+                // We keep spans here, cause removing spans here also removes b-tags inside of them, instead we strip them out later.
+                var validElements = "-strong/b,-em/i,-u,-span,-p,-ol,-ul,-li,-p/div,-a[href|name],sub,sup,strike,br,del,table[width],tr,td[colspan|rowspan|width],th[colspan|rowspan|width],thead,tfoot,tbody,img[src|alt|width|height],ul,ol,li,hr,pre,dl,dt,figure,figcaption,wbr"
```
warrenbuckley: Minor thing to avoid confusion as in `config/tinyMceConfig.config` XML file there is ValidElements that get merged in with the following code. I suggest the variable name is more explicit to be something like `validWordElements` if we need to maintain a different list from the already existing one.

```js
 //These are absolutely required in order for the macros to render inline
//we put these as extended elements because they get merged on top of the normal allowed elements by tiny mce
var extendedValidElements = "@[id|class|style],-div[id|dir|class|align|style],ins[datetime|cite],-ul[class|style],-li[class|style],-h1[id|dir|class|align|style],-h2[id|dir|class|align|style],-h3[id|dir|class|align|style],-h4[id|dir|class|align|style],-h5[id|dir|class|align|style],-h6[id|style|dir|class|align],span[id|class|style]";
```

src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
```diff
@@ -332,7 +347,35 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
...
+                  
+                // add valid elements from styleFormats.
```
warrenbuckley: Add a more detailed explanation of what you are trying to do here. Only allow a H3 to be pasted across from Word if you have it setup/configured as a user-defined rule in the CSS

src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
```diff
@@ -332,7 +347,35 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
...
+                for(; i < styles.styleFormats.length; i++) {
+                    style = styles.styleFormats[i];
+                    if(style.block) {
+                        validElements += "," + style.block;
+                    }
+                }
+                
+                var pasteConfig = {
+                    
+                    paste_remove_styles: true,
```
warrenbuckley: Find out what these paste_ properties do and give an inline comment, so we understand whats going on in 6months time when we come back to this code.

warrenbuckley:
All tested & works fine
The failing JS unit test is in `v8/dev` and not caused by this PR

PR #6373, opened 2019-09-19
v7/bug/6366 - Update rte.controller.js

martinjgriffiths:
Fix issue where an embedded macro prevents RTE editor from returning to pristine/valid after a publish and causes the save changes dialog to appear.

### Prerequisites

Tested on the starter kit. v7.15.3

### Description

There seems to be a significant TODO comment in this file discussing issues with event handling of the RTE. So this is an interim fix prior to anyone working on a more permanent solution.

During debugging I noticed after a save and publish the editor always returned to valid/pristine despite the editor.on('SetContent') event being called each time immediately after, even if no changes had occurred inside the RTE itself. So it was a simple matter of adding an isDirty check to this event to prevent the syncContent method from being called if the editor had become valid/pristine.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/rte/rte.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/rte/rte.controller.js
@@ -257,7 +257,10 @@ angular.module("umbraco")
                     // Update model on change, i.e. copy/pasted text, plugins altering content
                     editor.on('SetContent', function (e) {
                         if (!e.initial) {
-                            syncContent(editor);
+                              // sync content if editor is dirty
+                              if (!editor.isNotDirty) {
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Wonderful @martinjgriffiths that works like a charm!

Congratulations on your first merged PR for Umbraco-CMS - well done! I've added a shiny new contributor badge to your Our Umbraco profile - https://our.umbraco.com/members/id:5674/🏅

Shazwazza:
Hi @nul800sebastiaan + @martinjgriffiths

This change will need to be manually applied to v8 since this does not merge upwards. It may not even be an issue in v8 but might be worth checking.

Thanks!

nul800sebastiaan:
All good, I can not reproduce this in v8, so no need to take further action.

PR #6388, opened 2019-09-19
Adds udi to the content+media indexes

### LINKED ISSUE #4019: Add GUIDs to internal index for debugging
When debugging deployment errors on Umbraco Cloud you sometimes get an error message with a content nodes GUID, this is currently not in the internal index, so you have to either search through all content nodes or search in the database. An error can look like this:

```
The source environment has thrown a Umbraco.Deploy.Exceptions.RemoteApiException with message: The remote API has thrown an exception. It might have been caused by an inner System.NotSupportedException with message: Invalid prefix ‘D’.

The following artifacts might be involved:

umb://document/ee804d7a5b774e15ae741e7fb686225c
```

It would be great if we could simply take that guid and search either in the examine index through the developer section, or simply search for it in the search field from the content page.

skttl:
Fixes #4019 

Adds a UDI field to both content and media indexes, and makes the TreeSearcher use this.

billede

billede

I don't know if it is intended, or a bug, but you need to escape the `:` in the examine dashboard for searching the Udi.

### DIFF
--- a/src/Umbraco.Examine/ContentValueSetBuilder.cs
+++ b/src/Umbraco.Examine/ContentValueSetBuilder.cs
@@ -45,6 +45,7 @@ namespace Umbraco.Examine
                     {"icon", c.ContentType.Icon?.Yield() ?? Enumerable.Empty<string>()},
                     {UmbracoExamineIndex.PublishedFieldName, new object[] {c.Published ? "y" : "n"}},   //Always add invariant published value
                     {"id", new object[] {c.Id}},
+                    {"udi", new object[] { c.GetUdi() } },
                     {UmbracoExamineIndex.NodeKeyFieldName, new object[] {c.Key}},
                     {"parentID", new object[] {c.Level > 1 ? c.ParentId : -1}},
                     {"level", new object[] {c.Level}},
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Thanks @skttl - I just need a nod from @Shazwazza that this is okay! :)

jmayntzhusen:
Just tested this out, and was thinking maybe instead of adding fx: `umb://document/1b652919eff9458585dbe0060b9edb1a` it would make more sense to add `1b652919eff9458585dbe0060b9edb1a`. You may not always get the whole thing. 

Also not sure if making this a wildcard search would break something else, but otherwise that would be awesome so you could do either..

skttl:
But then it's not the Udi? What about changing the key field to exclude `-`'s, or have both?

Shazwazza:
We want to try not to put duplicate data in the index. Index's become large quickly so would rather not start putting duplicate data in there purely based on searching.

We already have the GUID in the index since it's the 'key', in my opinion we don't need the whole UDI since we already know what type it is since we store this in the index (i.e. `content` or `media`). So you can already search on the GUID (`__Key`), It's probable that this value is hyphenated, so what we should do is create a custom GUID lucene analyzer that strips out all hyphens which means any formatted GUID passed in will end up without hyphens which also means when the field is searched on, the hyphens are automatically stripped out too.

nul800sebastiaan:
Alright, it looks like the conclusion is: this data already exists. The problem is: we can't query on it in the backoffice currently (or maybe we can but I don't know how). Let's retry this PR with Shannon's input in mind, I'll close this one for now as it seems it's the incorrect solution :-)

PR #6396, opened 2019-09-20
Bug: RTE Image deletion error for images that has not been saved as media items

warrenbuckley:
## Reported Error
When an image is added to a RTE and is uploaded to the TEMP directory but the content node is not saved, when the image is deleted/removed it would throw an error in the console about the missing blob image.

See attached video from @Shazwazza 
https://www.screencast.com/t/CCdOkqpvkEe

## Test Notes
* Have browser devtools open
* Add an image to the editor - wait for it to be uploaded
* Do not save/publish the content node
* Highlight & remove the image
* Verify that no error is no longer thrown in the browser devtools console

### More Notes & Testing

This was fixed by removing the following line
`tinymce.activeEditor.editorUpload.blobCache.removeByUri(imgSrc);`

However, this was originally added with @bergmania as if the same image is added to the RTE more than once it would not add the **data-tempimg** attribute and thus this would not persist correctly into the database. This PR also now fixes this bug at the same time.

* Drag an image into the RTE
* Drag it in again (several times if you wish)
* View the HTML source code using TinyMCE button
* Verify that all the images have the exact same data-tempimg attribute
* Save/Publish the node and verify that only one image is created in the media library

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
@@ -4,7 +4,7 @@
     function ContentEditController($rootScope, $scope, $routeParams, $q, $window,
         appState, contentResource, entityResource, navigationService, notificationsService,
         serverValidationManager, contentEditingHelper, localizationService, formHelper, umbRequestHelper,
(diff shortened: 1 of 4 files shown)


warrenbuckley:
Seems OK for me @bergmania 
image

Can we look at the sizing thing in a new issue/PR if we need to please?

PR #6401, opened 2019-09-22
Fixes #6269: Can't delete language because of tags

### LINKED ISSUE #6269: Can't delete language because of tags
When trying to delete a language, after the tags editor is used to create tags for that language, the cms throws the following error:
image

Reproduction
------------
- Install Umbraco
- Add a language
- Add a page with culture variants and a tag editor
- Add content in the added language
- Try to delete the language
* bonus: deleting the language without tags editor works fine and will properly delete the content as well

### Specifics

Umbraco v8.1.3

AndyButland:
Issue here seems to be that the delete operations weren't in the correct order with respect to tags, and a related table wasn't having it's content deleted.

Have fixed by re-ordering and updating the language delete clauses in language repositoy to ensure any tags for the language are deleted before the language itself is deleted.

Fixes #6269.

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/LanguageRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/LanguageRepository.cs
@@ -102,17 +102,17 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
 
         protected override IEnumerable<string> GetDeleteClauses()
         {
-
             var list = new List<string>
                            {
                                //NOTE: There is no constraint between the Language and cmsDictionary/cmsLanguageText tables (?)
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks @AndyButland! I'm not sure the `where in` query would work in SQL CE, did you test that one?

AndyButland:
Only checked on SQL Server, but wasn't aware of that CE limitation and seems there's quite a lot of use of `WHERE IN` via `NPocoSqlExtensions.WhereIn` in the code base already, so believe it should be OK.

nul800sebastiaan:
Yep, works great! Thanks @AndyButland! ⭐️

PR #6403, opened 2019-09-22
Resolves #6374: Hide settings and delete option when re-ordering doc type properties

### LINKED ISSUE #6374: hide settings when reordering DocType
We should hide the settings-icon when re-ordering a DocumentType.

Because it makes the re-order elements too high.

image

AndyButland:
As noted in #6374, having the settings icon visible when re-ordering document type properties makes the property elements taller than they need to be.

I've resolved that by hiding this when sorting is in operation, and also the delete button (which seems reasonable to remove when sorting, and when in place it has some UI issues as the confirm overlaps the sort number field).

Now looks like this:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/umb-groups-builder.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/umb-groups-builder.html
@@ -248,7 +248,7 @@
                             </div>
 
                             <!-- row tools -->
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Yep was going to suggest removing the recycle bin icon as well. we'll have a look soon!

nul800sebastiaan:
Lovely, thanks again @AndyButland ! 👍

PR #6404, opened 2019-09-22
Fixes #6400: Content Templates don't set name correctly when created from existing content

### LINKED ISSUE #6400: Content Templates don't set name correctly when created from existing content.
When I create a new content template from an existing page, the content template uses the name of the page instead of the name set during the creation process. Do note that this can be worked around by editing the name of the content template directly in the settings section after creating it, so this issue is just a minor nuisance.

### Steps to reproduce

1. Create a content template from a content node.
2. Enter a name that is different from the name of the content node.
3. Go to the content template folder in the settings section and check the name of the new content template.

### Expected result

A content template with the name set during the creation process.

### Actual result

The content template uses the name of the node it was created from instead of the entered name.

### Other Notes
**Umbraco Version:** 8.1.3

This is also another piece of the backoffice that would benefit from a loading indicator while the content template is being created by the api call. The submit button doesn't currently change when clicked, so it's unclear that umbraco is doing something until it finishes the api call.

AndyButland:
As reported in #6400, when content templates are created from content, the name provided isn't saved, and the content template takes the name of the selected content item.

This was happening as although the main name of the content item was set correctly, the variant name for the default language wasn't.  So this PR sets that too.

Note for reviewer: worth checking how I've used the "scope" here.  Without it, I get an error calling `_languageRepository.GetDefaultIsoCode()` due to the absence of an "ambient scope".  But no save operation is happening here, so I'm not calling `scope.Complete()`.


nul800sebastiaan:
Good find! We'll have a look soon! 👍

nul800sebastiaan:
Thanks very much @AndyButland - I've completed the scope anyway since it does do a check already to see if there's a value. I'm not entirely sure if it's absolutely necessary, but I figure it can't hurt.

PR #6412, opened 2019-09-23
Drag/Drop & Pasted images in RTE aren't resized to configured max image width setting

warrenbuckley:
When we choose an image using the media picker in the RTE, when the image is inserted it is resized to a configured width/height, but when you drag/drop an image, it retains the original (probably hi-res) size which isn't very usable. The drag/drop functionality should behave the same as the media picker and resize to the configured width/height.

## Test Notes
* Verify code changes
* Test with a RTE & ensure you change/set a desired max width for images in the datatype config
* Test with nested content
* Test in the Grid

Fixes [AB#2737](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2737)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
@@ -241,7 +241,12 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
                         // To create a media item & delete this tmp one etc
                         tinymce.activeEditor.$(img).attr({ "data-tmpimg": tmpLocation });
 
-                        // We need to remove the image from the cache, otherwise we can't handle if we upload the exactly 
+                        // Resize the image to the max size configured
+                        // NOTE: no imagesrc passed into func as the src is blob://...
+                        // We will append ImageResizing Querystrings on perist to DB with node save
+                        sizeImageInEditor(editor, img);
+
+                        // We need to remove the image from the cache, otherwise we can't handle if we upload the exactly
                         // same image twice
                         tinymce.activeEditor.editorUpload.blobCache.removeByUri(imgSrc);
                     });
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
```diff
@@ -241,27 +241,55 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
...
+
+    function sizeImageInEditor(editor, imageDomElement, imgUrl) {
+
+        var size = editor.dom.getSize(imageDomElement);
+        console.log('size', size);
+
+        if (editor.settings.maxImageSize && editor.settings.maxImageSize !== 0) {
+            var newSize = imageHelper.scaleToMaxSize(editor.settings.maxImageSize, size.w, size.h);
+
+            console.log('new size', newSize);
```
bergmania: Please remove the console logging

```suggestion

```

src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
```diff
@@ -241,27 +241,55 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
...
+
+            editor.dom.setAttrib(imageDomElement, 'width', newSize.width);
+            editor.dom.setAttrib(imageDomElement, 'height', newSize.height);
+
+            // Images inserted via Media Picker will have a URL we can use for ImageResizer QueryStrings
+            // Images pasted/dragged in are not persisted to media until saved & thus will need to be added
+            if(imgUrl){
+                console.log('got an img src use QS');
```
bergmania: Remove console logging
```suggestion

```

src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
```diff
@@ -241,27 +241,55 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
...
-        
-        
+
+
+    }
+
+    function sizeImageInEditor(editor, imageDomElement, imgUrl) {
+
+        var size = editor.dom.getSize(imageDomElement);
+        console.log('size', size);
```
bergmania:
```suggestion

```

bergmania:
Test is okay.

But when I upload an image of size `1285 x 723` on the default settings
It tries to save the image in size `500 x 281.3229571984436`

When this is executed in ImageProcessor, it ends up as `499 x 281`, but the html img is size `500 x 281`

Can we do something about that?

warrenbuckley:
@bergmania I haven't touched any of the logic as you can see but its probably doing some rounding stuff but the browser resizing up/down by a pixel is not going to make a huge difference IMO.

Check to see if this problem existed before my PR (imagine it does)

PR #6002, opened 2019-07-26
Content migration - nested content

stevemegson:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes #5889

This PR creates `NestedContentPropertyEditorsMigration`, but adding it to the migration plan is left as an exercise for the reader.

### Description
Property values within nested content are currently untouched during content migration. This PR applies the conversions currently applied to top-level properties:

- DropDownFlexible (or other dropdowns which have been converted)
- RadioButtonList
- CheckBoxList
- MultiUrlPicker (converted from RelatedLinks)
- further levels of nested content

Also, any content types used within nested content must currently be manually changed to element types. This PR makes the change automatically in simple cases. It will set a content type as an element type if it:

- is used in a nested item
- is not used by any existing content
- uses no compositions

Warnings are generated for any content type which is used in nested content but was not set as an element type.

This step could be made more thorough, but I think this handles enough common cases to be useful. Where a content type uses compositions, we could in some cases still safely set all the necessary types as element types. We could also consider content types which are selected in nested content configuration but are not currently used anywhere.

Testing:
- In 7.15.0, create nested content containing dropdowns, related links, and other nested content.
- Migrate to 8.1.0 and check that nested elements are visible and have the expected values.

---
_This item has been added to our backlog [AB#4935](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4935)_

PR #6413, opened 2019-09-24
V8: Migration of obsolete picker properties

### LINKED ISSUE #6241: Validate data types earlier in upgrade process
I have been working on upgrading a very large site to Umbraco 8.

The upgrade is taking a very long time to run (5+ hours). This is fine. However, then I get errors like this:

> The database failed to upgrade. ERROR: The database configuration failed with the following message: Cannot rename datatype alias "Umbraco.MemberPicker2" to "Umbraco.MemberPicker" because the target alias is already used.This is generally because when upgrading from a v7 to v8 site, the v7 site contains Data Types that reference old and already Obsolete Property Editors. Before upgrading to v8, any Data Types using property editors that are named with the prefix '(Obsolete)' must be migrated to the non-obsolete v7 property editors of the same type. Please check log file for additional information (can be found in '/App_Data/Logs/')

This is also fine, however it would be much nicer if these errors could be detected much sooner in the install process.

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #6241

### Description
Content migration currently fails if the database includes any properties of the pre-7.6 picker types, which store integer IDs rather than UDIs.

This PR adds a migration which converts these properties to the UDI-based versions.

Testing:
- In 7.15, enable obsolete property editors and create a content picker, media picker, and multi-node tree picker using the obsolete types. Add some test content.
- Upgrade to 8.1
- Check that the properties have been converted to the up-to-date versions, and the selected values are still visible in the editor.

### DIFF
--- a/src/Umbraco.Core/Constants-PropertyEditors.cs
+++ b/src/Umbraco.Core/Constants-PropertyEditors.cs
@@ -20,6 +20,7 @@ namespace Umbraco.Core
                 {
                     public const string Textbox = "Umbraco.Textbox";
                     public const string Date = "Umbraco.Date";
+                    public const string ContentPicker = "Umbraco.ContentPickerAlias";
                     public const string ContentPicker2 = "Umbraco.ContentPicker2";
                     public const string MediaPicker2 = "Umbraco.MediaPicker2";
                     public const string MemberPicker2 = "Umbraco.MemberPicker2";
--- a/src/Umbraco.Core/Migrations/Upgrade/UmbracoPlan.cs
+++ b/src/Umbraco.Core/Migrations/Upgrade/UmbracoPlan.cs
@@ -129,6 +129,7 @@ namespace Umbraco.Core.Migrations.Upgrade
             To<DataTypeMigration>("{8640C9E4-A1C0-4C59-99BB-609B4E604981}");
             To<TagsMigration>("{DD1B99AF-8106-4E00-BAC7-A43003EA07F8}");
             To<SuperZero>("{9DF05B77-11D1-475C-A00A-B656AF7E0908}");
+            To<LegacyPickersPropertyEditorsMigration>("{258CE811-4201-4004-95EF-402086B38B77}");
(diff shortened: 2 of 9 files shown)


nul800sebastiaan:
This sounds cool @stevemegson - we'll give it a look, thanks so far!

genckastrati:
Would love to have this in 8 soon :D thank you for building it @stevemegson

guruprasad83:
We are also stuck in this step where content pickers values are not showing in the UI after we change from obsolete to normal content pickers as this is the important step before, in migrating to V8.
Could you please tell me how sooner we can expect this fix into main umbraco.

nul800sebastiaan:
@guruprasad83 We don't have any news on an update for this at the moment, you're welcome to try out the steps I've detailed in this comment though, that helps us gather important input on whether this fix is indeed sufficient: https://github.com/umbraco/Umbraco-CMS/pull/6002#issuecomment-597288007

genckastrati:
@guruprasad83 @nul800sebastiaan I went ahead and wrote my own script to migrate from old obsolete fields to new ones.

nul800sebastiaan:
@genckastrati If there's anything you can share for that then that would be helpful as well! 👍

genckastrati:
Tried to find the code but for some reason I must have not saved it on Git, very strange. 

Anyway the process was simple:

Add the new data type that will replace the old one. For example I had an obsolete media picker "Thumbnail", and I added a new media picker "Image Thumbnail". 

Create an asxp page that you can execute on the server, and that you can access. You need to edit the web.config so it lets you run it.

Get the children of the specific node where you want these changes to happen.

Go through each and extract the old Thumbnail media item (which was stored as an ID), get the IPublishedContent object and store it in the new one. Save and publish that child item.

You might need to do some parsing because obsolete media pickers stored things as separated by commas, so you need to handle that. 

This is it really. I will try to find the code but am really confused as to why it isnt in git.

teajaysee:
This still never became part of Umbraco?

PR #6416, opened 2019-09-24
Don't store JSON property data as indented

### LINKED ISSUE #2996: Umbraco shouldn't store JSON property data as indented
### Bug summary

When saving a content or media item in Umbraco backoffice, the data of properties that store JSON will be saved as indented, which takes up more space in the database. Ideally JSON data shouldn't be indented when saved.

If saved via the content service, it instead depends on the specific implementation whether JSON data is indented or not, making it a bit inconsistent.

### Steps to reproduce

For instance, when saving a page with a grid property, the JSON value saved inthe database is something like:

```json
{
  "name": "1 column layout",
  "sections": [
    {
      "grid": 12,
      "rows": [
        {
          "label": "Section",
          "name": "Section",
          "areas": [
            {
              "grid": 12,
              "hasConfig": false,
              "controls": [
                {
                  "value": "Here is a headline",
                  "editor": {
                    "alias": "headline"
                  },
                  "active": false
                }
              ],
              "active": false
            }
          ],
          "hasConfig": false,
          "id": "e96ffdff-bfd2-8238-0909-33f1717cf7f3",
          "active": false
        },
        {
          "label": "Section",
          "name": "Section",
          "areas": [
            {
              "grid": 12,
              "hasConfig": false,
              "controls": [
                {
                  "value": {
                    "id": 1058,
                    "udi": "umb://media/31e5d65860174179b4052d54eca304b5",
                    "image": "/media/1031/photo-1460901419481-3fc27f0cdae9.jpg",
                    "caption": "Here is a bird"
                  },
                  "editor": {
                    "alias": "media"
                  },
                  "active": true
                }
              ],
              "active": true,
              "hasActiveChild": true
            }
          ],
          "hasConfig": false,
          "id": "1e4639a7-e8e2-00f7-c92a-7c9e27ef1140",
          "hasActiveChild": true,
          "active": true
        }
      ]
    }
  ]
}
```

```json
{"name":"1 column layout","sections":[{"grid":12,"rows":[{"label":"Section","name":"Section","areas":[{"grid":12,"hasConfig":false,"controls":[{"value":"Here is a headline","editor":{"alias":"headline"},"active":false}],"active":false}],"hasConfig":false,"id":"e96ffdff-bfd2-8238-0909-33f1717cf7f3","active":false},{"label":"Section","name":"Section","areas":[{"grid":12,"hasConfig":false,"controls":[{"value":{"id":1058,"udi":"umb://media/31e5d65860174179b4052d54eca304b5","image":"/media/1031/photo-1460901419481-3fc27f0cdae9.jpg","caption":"Here is a bird"},"editor":{"alias":"media"},"active":true}],"active":true,"hasActiveChild":true}],"hasConfig":false,"id":"1e4639a7-e8e2-00f7-c92a-7c9e27ef1140","hasActiveChild":true,"active":true}]}]}
```

The indended JSON string has a size of 1668 bytes, while the non-intended JSON string ha a size of 744 bytes. That is a reduction of roughly **44%**. It may be even more for larger JSON strings.

For a site with thousands of pages, which each may have lots of versions, this will be a lot of extra and unnecessary white space that is being saved to the database.

Another example is when uploading a new image in the media section, the crops of the image is automatically generated and saved (probably directly from C#). The crops are stored something like:

```json
{"src": "/media/1031/photo-1460901419481-3fc27f0cdae9.jpg", "crops": []}
```
Although the result here still has a few spaces here and there, there are no line break or indentation. However if an editor later saves the image through the backoffice, the JSON will instead be stored as indented:

```json
{
  "src": "/media/1031/photo-1460901419481-3fc27f0cdae9.jpg",
  "focalPoint": {
    "left": 0.51166666666666671,
    "top": 0.3575
  }
}
```

### Fix?

I haven't yet found the part of the code where the property values are being serialized to JSON. With a few hints in the right direction, I'd be happy to create a PR for this ;)

benjaminhowarth1:
### Prerequisites

This fixes #2996.

### Description

In areas where Formatting.Indented was previously used, this now adds a check on GlobalSettings.DebugMode to determine whether to use JSON indentation or not (in release mode, this should substantially reduce data storage size). Should not require a unit test, as it merely affects formatting.

### DIFF
--- a/src/Umbraco.Core/Serialization/JsonNetSerializer.cs
+++ b/src/Umbraco.Core/Serialization/JsonNetSerializer.cs
@@ -4,6 +4,7 @@ using System.Reflection;
 using System.Text;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
+using Umbraco.Core.Configuration;
 
 namespace Umbraco.Core.Serialization
 {
@@ -60,7 +61,8 @@ namespace Umbraco.Core.Serialization
         /// <returns></returns>
         public IStreamedResult ToStream(object input)
         {
-            string s = JsonConvert.SerializeObject(input, Formatting.Indented, _settings);
+            var formatting = GlobalSettings.DebugMode ? Formatting.Indented : Formatting.None;
+            string s = JsonConvert.SerializeObject(input, formatting, _settings);
             byte[] bytes = Encoding.UTF8.GetBytes(s);
             MemoryStream ms = new MemoryStream(bytes);
 
--- a/src/Umbraco.Core/Serialization/JsonNetSerializer.cs
+++ b/src/Umbraco.Core/Serialization/JsonNetSerializer.cs
@@ -61,8 +61,7 @@ namespace Umbraco.Core.Serialization
         /// <returns></returns>
         public IStreamedResult ToStream(object input)
         {
-            var formatting = GlobalSettings.DebugMode ? Formatting.Indented : Formatting.None;
-            string s = JsonConvert.SerializeObject(input, formatting, _settings);
+            string s = JsonConvert.SerializeObject(input, Formatting.None, _settings);
             byte[] bytes = Encoding.UTF8.GetBytes(s);
             MemoryStream ms = new MemoryStream(bytes);
 
--- a/src/Umbraco.Web/Editors/BackOfficeController.cs
+++ b/src/Umbraco.Web/Editors/BackOfficeController.cs
@@ -177,8 +177,7 @@ namespace Umbraco.Web.Editors
                 //the dictionary returned is fine but the delimiter between an 'area' and a 'value' is a '/' but the javascript
                 // in the back office requres the delimiter to be a '_' so we'll just replace it
                 .ToDictionary(key => key.Key.Replace("/", "_"), val => val.Value);
-            var formatting = GlobalSettings.DebugMode ? Formatting.Indented : Formatting.None;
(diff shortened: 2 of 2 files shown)


nul800sebastiaan:
We'll have to have a look @benjaminhowarth1 - thanks so far. 

I'm sure this also impacts how we do compares in Deploy on Cloud, so this change might not be feasible, or at least has a larger impact, not sure yet.

benjaminhowarth1:
@benjaminhowarth1 my pleasure 🙂 if I'm not mistaken, cause it's just a formatting thing on Serialize, It shouldn't affect existing data with Deserialize (cause Deserialize doesn't care about whitespace).

nul800sebastiaan:
Deploy team said it should be all good! 👍 

Now, I KNOW this is not your first contribution to Umbraco, but I had to double check on Our, you already had the contrib badge on your profile as well! 💪

I have no idea how well this will merge up to v8, I'll have to investigate 😅

clausjensen:
Hey @benjaminhowarth1 

We're good with the change on the `BackOfficeController.cs` .. that should reduce the size of the response and Chrome will format it properly anyway if you need to debug.

I however don't really see this PR changing anything in regards to how data is actually stored in the database... so if that was still the intention then I think it needs some rework.

So... unless I missed something - I think it would be good if we changed the title/description of this PR to actually indicate what the change is, since this is added to release notes and whatnot (@nul800sebastiaan ?)

PR #6422, opened 2019-09-25
Fixes #6419 regression issue in nucache causing a panic exception

### LINKED ISSUE #6419: V8: "panic: no more children" on Save and Publish
In the latest v8 code, I see a "panic: no more children" exception when publishing a document which is a "middle child" of its parent.

The problem appears to be the handling of generations in `ContentStore.AddTreeNodeLocked`. The code steps through the linked list of the siblings looking for the right place in the sort order to insert `content`. However, in the generation we're looking at this linked list contains a version of `content` which has had its `PreviousSiblingContentId` and `NextSiblingContentId` set to -1. This makes the loop end before finding the right place to insert `content`, leading to the "we should never get here" situation.

Rather than using the current generation of the parent node, I think we need to use the current generation of `content`, so that we get the generation in which `content` has been removed from the tree. Can we always assume that some current generation of `content` exists, or do we need to check and possibly to fallback to `parentLink.Gen`?

Although they're not directly causing this exception, while investigating I found a couple of other bits which I think need to change:
- In `ContentStore.Set`, we set `kit.Node.NextSiblingContentId` but not `kit.Node.PreviousSiblingContentId`.
- In `RemoveTreeNodeLocked`, I'm not sure that we should be resetting `content.NextSiblingContentId` and `content.PreviousSiblingContentId` to -1. We're modifying the existing generation here, not setting up the new version.

---
_This item has been added to our backlog [AB#2848](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2848)_

Shazwazza:
Fixes #6419 regression issue in nucache causing a panic exception

The cause of this was a recent fix for nucache when it was getting a linked node but it wasn't requesting by the current generation, see rev: 7fdbb84c04e0837761a173b209cc6c7501be3d8a 

The regression was a change to AddNodeLocked which we don't want to use the current gen, we want to get the latest gen (using `null`). It was causing an error because the change was requesting nodes by the current parent gen which is incorrect because the parent gen doesn't get incremented when changing a branch of children causing the current node to be returned instead of not returned since when we call `RefreshBranch` for a given node, we actually 'Clear' that node and we are expecting it to no longer exist in the next gen when we call AddNodeLocked. It's a little confusing how it works and is due to an explicit order of operations that must occur. 

The fix was relatively easy and I've added unit tests to show that it now works. The good news is that we now have some unit tests to test how generations in nucache work for both types of single item refreshing that we use: RefreshBranch + Remove

## Testing

* Ensure all unit tests pass
* Review code
* run Umbraco and publish nodes: First, middle and last item within a branch/level and make sure it works

No need to assign the bug report to a release version since this bug was introduced in the 8.2 work in progress.

### DIFF
--- a/src/Umbraco.Tests/PublishedContent/NuCacheChildrenTests.cs
+++ b/src/Umbraco.Tests/PublishedContent/NuCacheChildrenTests.cs
@@ -108,8 +108,8 @@ namespace Umbraco.Tests.PublishedContent
             );
 
             // create a scope provider
-            var scopeProvider = Mock.Of<IScopeProvider>();
-            Mock.Get(scopeProvider)
+            var scopeProvider = new Mock<IScopeProvider>();
+            scopeProvider
                 .Setup(x => x.CreateScope(
                     It.IsAny<IsolationLevel>(),
                     It.IsAny<RepositoryCacheMode>(),
@@ -143,7 +143,7 @@ namespace Umbraco.Tests.PublishedContent
                 _snapshotAccessor,
                 _variationAccesor,
                 Mock.Of<IProfilingLogger>(),
-                scopeProvider,
+                scopeProvider.Object,
                 Mock.Of<IDocumentRepository>(),
                 Mock.Of<IMediaRepository>(),
                 Mock.Of<IMemberRepository>(),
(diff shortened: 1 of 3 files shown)


bergmania:
Tested and works as expected :)

PR #6428, opened 2019-09-25
v8: Change button style for "create package" and "show more" (in the log viewer)

bjarnef:
### Prerequisites

### Description
This PR replace the `success` button style for "Create package" with `action` button style (purple) which is now used in most places like users, languages, etc.

It also change button style for "Show more" in log viewer and change this button to use `umb-button` component and I have cleaned up some odd empty class attributes in logviewer. The logviewer also need to be cleaned up for inline styles, but we can do that in another PR :)

image

image


emmaburstow:
Morning @bjarnef !

Thanks for the contribution. Loving the UX improvements. We'll take a look and let you know if we need anything.

Emma

nul800sebastiaan:
Thanks @bjarnef ! 👍

PR #6434, opened 2019-09-26
RTE initialization fix

nielslyngsoe:
We have seen an issue sometimes, causing problem initialising the TinyMCE in some cases, this fix applies a delay to help the browser append elements to document before looking for the placeholder element for TinyMCE.


bergmania:
The fix works on my machine.. But I would like if you could investigate if there is an event to use instead of a timeout of 150 ms.

bergmania:
The styling issue is almost fixed, but some sizes is still giving trouble when resizing the window:

image
image


Also if is has been small and is increased in size, the size panel is not resizing:
image

Only seems to be a problem to RTEs that was not there on page load.

PR #6438, opened 2019-09-26
Fixes selecting Macro Content in RTE to be non editable once node is saved

warrenbuckley:
Adds contenteditable=false attribute as for whatever reason the CSS class of .mceNonEditable is not working when a macro is saved & loaded in a content node

## Test Notes
* Add a macro into the RTE
* Before saving the page/node (ensure you can not select inside it & can drag/move it around)
* Save the node & retry moving it (Previous to this PR it would not work)

Fixes [AB#2877](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2877)


bergmania:
Works as expected :)

PR #6449, opened 2019-09-26
Health Check - Text alternatives for icons missing

BatJan:
Co-authored-by: @webdokkeren <thormadsenholm@gmail.com>

### Prerequisites

This fixes issue **no. 152** from #5277 

### Description
In this PR we have added a few next texts for "passed", "warning", "failed" and "suggestion" in the language files.

In the health check view these texts have been added inside a `<span class="sr-only">`, which means a screen reader will announce for instance "Configuration 1 passed, 2 warning" depending on what's exposed to the DOM.

We realise that pluralized texts need to be handled but that is something that needs to be done in another task as it's something that's not available out of the box currently and will affect other scenarios as well. This is a lot better than having no texts at all 😃

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/dashboard/settings/healthcheck.html
+++ b/src/Umbraco.Web.UI.Client/src/views/dashboard/settings/healthcheck.html
@@ -32,26 +32,41 @@
                             <umb-load-indicator></umb-load-indicator>
                         </div>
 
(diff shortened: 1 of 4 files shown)


nul800sebastiaan:
Woohoo! Thanks for the PR, well done gentlemen and congrats on entering Hacktoberfest a liiiiittle early this year! 🎉

Oh, and I hear it's @webdokkeren's very first Umbraco PR! So you now have that shiny new contributor badge on Our Umbraco, congrats! 🥇

image

PR #6453, opened 2019-09-26
Health Check - Text alternatives for icons missing (In Groups)

BatJan:
### Prerequisites

This fixes issue **no. 161** from the Trello board https://trello.com/c/dpVcuk4x/165-161-health-check-text-alternatives-for-icons-missing-in-groups

### Description
There was an `<a>`, which was converted to a button. Furthermore some icons have been hidden using `.sr-only` and some text alternatives have been provided instead.

Currently a screen reader might not pick up the texts at all but this is something that needs to be addressed in another task but the groundwork for making it work has been done 👍


nul800sebastiaan:
Thanks @BatJan !! ⭐️

PR #6454, opened 2019-09-26
v8: Use umb-radiobutton component in relation types

bjarnef:
### Prerequisites

### Description
This PR replace the traditional radiobuttons in relation type view with `umb-radiobutton` component.

The `value` property is defined as one-way binding using `@` which also convert the value as a string, while `<` has ability to pass an object with its original data type to component: https://stackoverflow.com/a/36667226/1693918

I tried the latter one, but it seemed to cause some other issues in checkbox list and radibutton list property editor where it use interpolation like this `value="{{item.value}}"`.

When the current behaviour in `umb-radiobutton` and `umb-checkbox` the initial model value passed in to the component could be a boolean true/false, where it didn't select the radiobutton when model and value was equal because model was a boolean, but value is converted to a boolean and when making a change in radiobutton options the original value e.g. a json string changed from boolean to a string value.

There is a great article about this behaviour here: https://medium.com/front-end-weekly/angularjs-component-binding-types-my-experiences-10f354d4660

For now I have converted the original value to a string, which seems to work well.

Example where the original value of `isBidirectional` property not is converted to a string.

Example where the original value of `isBidirectional` property is converted to a string.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
@@ -46,7 +46,7 @@
                     vm.onChange({ model: vm.model, value: vm.value });
                 }, 0);
             }
-        }     
+        }    
     }
 
     var component = {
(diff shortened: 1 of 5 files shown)


nul800sebastiaan:
Interesting! Thanks Bjarne, we'll have a look :)

nul800sebastiaan:
Thanks @bjarnef - all good! 👍

PR #6459, opened 2019-09-26
v8: Package repo categories selection updates

bjarnef:
### Prerequisites

### Description
This PR add a few changes, so the categories are not hidden (while loading) on selection and also set the active class faster instead of it might wait to update the active state here https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/packages/views/repo.controller.js#L128

Furthermore it doesn't show just Popular heading when selecting "Umbraco Pro" with no results.


nul800sebastiaan:
Thanks @bjarnef - we'll have a look soon!

nul800sebastiaan:
Awesome, thanks very much @bjarnef ! 👍

PR #6460, opened 2019-09-27
enables infinite editing for partial views

skttl:
I am trying to enable infinite editors for Diplo God Mode, and noticed that partial views didn't support this. So I added that, in this PR 🙂


nul800sebastiaan:
Sweet! We'll have a look soon!

nul800sebastiaan:
Thanks @skttl ! 👍

PR #6462, opened 2019-09-27
Fixes #2997 - ContentService returns outdated result

### LINKED ISSUE #2997: ContentService returns outdated result
When getting content through ContentService.GetById(guid) the returned result will be a cached version that is the first version from that app restart.
Using the normal/older ContentService.GetById(int) returns the up-to-date result

This can be reproduced by comparing the two results after performing a change on a node, which should have resulted in a new version being created.

Using the Guid version:
image

Using the int version:
image

This was tested on Umbraco v. 7.12.2

Shazwazza:
Fixes #2997 - ContentService returns outdated result

This fixes the issue when using ContentService.GetById(Guid) and MediaService.GetById(Guid), the cache for these methods is not refreshed with the cache refreshers since they are only evicting the cache by INT. 

This will need to be backported to 7.x

Testing:

Create a controller:

```cs
public class TestController : SurfaceController
{
    public ActionResult GetContentByKey(System.Guid key)
    {
        var c1 = Services.ContentService.GetById(key);
        var c2 = c1 == null ? null : Services.ContentService.GetById(c1.Id);
        return Content(c1?.Name + " - " + c2?.Name);
    }

    public ActionResult GetMediaByKey(System.Guid key)
    {
        var m1 = Services.MediaService.GetById(key);
        var m2 = m1 == null ? null : Services.MediaService.GetById(m1.Id);
        return Content(m1?.Name + " - " + m2?.Name);
    }
}
```

Then go find a content's GUID and a media item's GUID.

Open 2x tabs in your browser (of course use your own GUIDs accordingly):
* /umbraco/surface/test/GetContentBykey?key=df1eb830-411b-4d41-a343-3917b76d533c 
* /umbraco/surface/test/GetMediaBykey?key=64478ab3-b117-4891-aa30-2f635de1bd5d

You will see the name by INT and by GUID

Then go change the name of the content and media item and refresh these tabs - the names should be the same since the cache has refreshed for both the INT and the GUID. Previously only the INT name would be refreshed.

### DIFF
--- a/src/Umbraco.Tests/Cache/RefresherTests.cs
+++ b/src/Umbraco.Tests/Cache/RefresherTests.cs
@@ -12,20 +12,22 @@ namespace Umbraco.Tests.Cache
         [Test]
         public void MediaCacheRefresherCanDeserializeJsonPayload()
         {
-            var source = new[] { new MediaCacheRefresher.JsonPayload(1234, TreeChangeTypes.None) };
+            var source = new[] { new MediaCacheRefresher.JsonPayload(1234, Guid.NewGuid(), TreeChangeTypes.None) };
             var json = JsonConvert.SerializeObject(source);
             var payload = JsonConvert.DeserializeObject<MediaCacheRefresher.JsonPayload[]>(json);
             Assert.AreEqual(source[0].Id, payload[0].Id);
+            Assert.AreEqual(source[0].Key, payload[0].Key);
             Assert.AreEqual(source[0].ChangeTypes, payload[0].ChangeTypes);
         }
 
         [Test]
         public void ContentCacheRefresherCanDeserializeJsonPayload()
         {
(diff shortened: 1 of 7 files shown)


warrenbuckley:
Code changes seem simple enough - off to test :)

warrenbuckley:
All works good & merged in 😄

warrenbuckley:
@Shazwazza will you be backporting this?

Alkaersig:
This explains our issues on a v7 platform where we share content between instances by key. Sometimes deleted Content is resolved when using GetByKey. It's a pretty ugly one, so a backport would be appreciated :)

Alkaersig:
Any updates for v7 backport? This is a major issue in one of your solutions where we push content to multiple umbraco instances and thus have to use the key as the id is not unique across multiple instances.

PR #6465, opened 2019-09-27
fixed xml comment for GridConfigurationEditor

gunnim:
previously had the same comment as the GridConfiguration. Now matches comment for other configurationEditors


nul800sebastiaan:
Great, thanks @gunnim ! 👍

PR #6473, opened 2019-09-30
#2996 resubmitting ContentExtensions and ObjectExtensions fixes

benjaminhowarth1:
### Prerequisites

Re-submitted PR as per @nul800sebastiaan's request.

### Description

Object.TryConvertTo<T> now supports a check for JObject and calls SerializeObject with Formatting.None. This ensure JSON is minified when storing to DB & caches.


nul800sebastiaan:
Thanks @benjaminhowarth1 - I don't know why the tags converter was still included, that one already stored it's content without whitespace. You also added a few files that had nothing to do with this change. 

I've reverted the irrelevant bits now and tested the update, this looks good, everything is now nice and flat! 

image

As a tip for your next contribution: make sure to not just chuck in all the changed files, but only required changes, that will make it much easier  to test and merge! 👍

PR #6480, opened 2019-09-30
V8/fix user group selection

dawoe:
### Prerequisites

### Description

When creating or editing a user the selection made in the group selection dialog is immediatly visible in the edit screen. If the user clicks close it reverts back to the original state. See gif for behavior

user group selection before

This was because the "actual" current selection was passed to the dialog, instead of a copy triggering the updates immediatly. After this PR the behavior is fixed

user group selection after

I have seen this in more places in the backend. So expect some more PR's

Dave

PS : Should i create PR targeted at V7 for this as well ?


nul800sebastiaan:
Nice one, thanks @dawoe !

PR #6485, opened 2019-09-30
Add umb-checkbox for publish unpublished content items with no variants

bjarnef:
### Prerequisites

### Description
I noticed that the "Save and publish > Publish with descendants" overlay was using `umb-checkbox` when using variants.

image

But not without any variants as now:

image

This is how it looks at the moment using a traditional checkbox, while it does use `umb-checkbox` when the document type vary by culture.

image


nul800sebastiaan:
Very nice, cheers @bjarnef !

PR #6488, opened 2019-09-30
Use umb-checkbox for loglevel selection

bjarnef:
### Prerequisites

### Description
This PR change so `umb-checkbox` is used in loglevel selection in the dropdown.

At the moment the dropdown opens behind the search input. First thing I think we need to change here is that the view only use a single `<umb-editor-sub-header>` element in logviewer search.
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/logviewer/search.html#L20

as in this example https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web.UI.Client/src/common/directives/components/editor/subheader/umbeditorsubheadersection.directive.js#L15-L41

but we can look at this in another PR.


nul800sebastiaan:
Love the pretty checkboxes! 👍`

Works great, cheers!

PR #6489, opened 2019-09-30
Added punctuation

abjerner:
The four descriptions were missing a period at the end. This PR adds that both for `en.xml` and `en_us.xml`:

image


nul800sebastiaan:
Great! 
Period. 😁

PR #6493, opened 2019-10-01
Fixes null reference caused in GridPropertyValueEditor when the value is empty

Shazwazza:
Fixes null reference caused in GridPropertyValueEditor when the value is empty.

This can occur when the value stored for the grid is an empty string, in which case we were not checking for that scenario and a null ref exception was thrown.

---
_This item has been added to our backlog [AB#2972](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/2972)_

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
@@ -83,6 +83,9 @@ namespace Umbraco.Web.PropertyEditors
 
                 // editorValue.Value is a JSON string of the grid
                 var rawJson = editorValue.Value.ToString();
+                if (rawJson.IsNullOrWhiteSpace())
+                    return null;
+
                 var grid = JsonConvert.DeserializeObject<GridValue>(rawJson);
 
                 // Find all controls that use the RTE editor


src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
```diff
@@ -83,6 +83,9 @@ public override object FromEditor(ContentPropertyData editorValue, object curren
 
                 // editorValue.Value is a JSON string of the grid
                 var rawJson = editorValue.Value.ToString();
+                if (rawJson.IsNullOrWhiteSpace())
+                    return null;
+
```
bjarnef: @Shazwazza does it also need null check on `Value`?

```
var rawJson = editorValue.Value?.ToString();
```

Shazwazza: We already check in the first line if editorValue.Value == null

bjarnef: Ahh, didn't check the full source code.

Shouldn't this check be moved above these three lines?
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs#L81-L83

Shazwazza: it could be moved yes, not a huge deal but if you want to submit a PR should be easy to merge.

PR #6498, opened 2019-10-01
V8/bugfix/0000 tiny mce css icon issue

nielslyngsoe:
Correcting CSS for custom icons in TinyMCE, so they also can be used in PreValue Editor


warrenbuckley:
All good :)

PR #6505, opened 2019-10-01
Idea: only show property-actions on hover or when focus-within

nielslyngsoe:
We want to minimize the amount of visual elements on the screen, to underline a simple, calm and friendly visual environment. To ensure this in the future when more data-types will be supporting the property-copy-action or other future property-actions, it would be nice to already now hide the property-copy action, to avoid having it constantly shown.

image

The concept is to only show the property-actions on hover, when there is focus-within (thought that might have it difficulties...) and when the user have tabbed to on of the actions.

This PR, does the first two, meaning it's still missing the ability to tab to the actions.

If you have any thought or concerns regarding this PR, please share them in this thread, thanks.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/property-editors.less
+++ b/src/Umbraco.Web.UI.Client/src/less/property-editors.less
@@ -61,12 +61,23 @@
     font-size: 16px;
     padding: 4px;
     line-height: 10px;
-
+    opacity: 0;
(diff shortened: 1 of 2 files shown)


Shazwazza:
THis has been merged into the new working branch: https://github.com/umbraco/Umbraco-CMS/tree/v8/feature/AB3001-copy-paste-property

PR #6519, opened 2019-10-01
V8; Support allowed types for MNTP media picker

kjac:
### Prerequisites

### Description

As-is you can pick the allowed types when configuring MNTP as a media picker. However, the picker doesn't respect the allowed types filtering like it does when configured as a content picker. This was uncovered while working on #5983 (see also [this comment](https://github.com/umbraco/Umbraco-CMS/pull/6177#issuecomment-531805007) on the PR fixing #5983)

This PR implements the filtering; when it's applied here's how the MNTP works in media picker mode:

mntp-media-allowed-types

...and it also works when you browse around the media library in the picker 😄 

mntp-media-allowed-types-2

#### Testing this PR

To test this:

1. Create a new media type (easiest to clone the default "media" one)
2. Create media of multiple types
3. Configure MNTP as a media picker and only allow one of the media types
4. Verify that the MNTP only lets you pick media of the allowed type
5. Do the same for multiple allowed media types
6. Remove the constraints on allowed media types and verify that the MNTP indeed allows you to pick any type

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbmediagrid.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbmediagrid.directive.js
@@ -193,6 +193,9 @@ Use this directive to generate a thumbnail grid of media items.
             * Returns wether a item should be selectable or not.
             */
             function getSelectableState(item) {
+                if (item.filtered) {
+                    return false;
+                }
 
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Oooh 😁

nul800sebastiaan:
Lovely, works as advertised! 👍

PR #6524, opened 2019-10-01
V8: Support allowed types for MNTP member picker

kjac:
### Prerequisites

### Description

As-is you can pick the allowed types when configuring MNTP as a member picker. However, the picker doesn't respect the allowed types filtering like it does when configured as a content picker. This was uncovered while working on #5983 (see also [this comment](https://github.com/umbraco/Umbraco-CMS/pull/6177#issuecomment-531805007) on the PR fixing #5983)

This PR implements the filtering; when it's applied here's how the MNTP works in member picker mode:

mntp-member-allowed-types

#### Testing this PR

To test this PR, first create multiple different member types and members of said types. 

Configure an MNTP as a member picker which allows only a subset of these types to be picked. Verify that the MNTP only lets you pick these member types, even when searching for members (note: disallowed types are apparently explicitly filtered out when searching - that's not caused by this PR).

Also verify that you can still pick any member type when there is no constraint on allowed member types.


nul800sebastiaan:
We'll have a look soon, thanks!

nul800sebastiaan:
Wonderful update @kjac - thanks very much!!

PR #6530, opened 2019-10-01
Fixes Search an element from its GUID with the backoffice #6185

### LINKED ISSUE #6185: Search an element from its GUID with the backoffice
When an error occurs during the restore caused by a mismatch structure, or the status of a document, the message indicates always the GUID et not always the ID.
Therefore, we must use the database to find quickly the item causing the problem...

It could be very cool to use the main search engine to find a document, a media, a content type, a form etc... from the GUID... (with or without "-")

Thank you !

Shazwazza:
Fixes #6185

You can now search by all searchable entity types by GUID in the back office and in both formats of GUIDs: with and without hyphens.

## Testing

Test that you can search by GUID in the top right search box for:

* Content
* Media
* Members
* Templates
* Content Types
* Media Types
* Data Types


src/Umbraco.Web/Trees/ContentTreeControllerBase.cs
```diff
@@ -18,11 +18,23 @@
...
     {
+
+        protected ContentTreeControllerBase(IGlobalSettings globalSettings, IUmbracoContextAccessor umbracoContextAccessor, ISqlContext sqlContext, ServiceContext services, AppCaches appCaches, IProfilingLogger logger, IRuntimeState runtimeState, UmbracoHelper umbracoHelper) : base(globalSettings, umbracoContextAccessor, sqlContext, services, appCaches, logger, runtimeState, umbracoHelper)
+        {
+        }
+
+        protected ContentTreeControllerBase()
+        {
+        }
+
```
ronaldbarendse: What's the reason for this empty constructor? And shouldn't it otherwise call the base class?

src/Umbraco.Web/Trees/ContentTreeController.cs
```diff
@@ -36,17 +40,18 @@ public class ContentTreeController : ContentTreeControllerBase, ISearchableTree
...
         protected override bool RecycleBinSmells => Services.ContentService.RecycleBinSmells();
 
         private int[] _userStartNodes;
+
+        public ContentTreeController(UmbracoTreeSearcher treeSearcher, ActionCollection actions, IGlobalSettings globalSettings, IUmbracoContextAccessor umbracoContextAccessor, ISqlContext sqlContext, ServiceContext services, AppCaches appCaches, IProfilingLogger logger, IRuntimeState runtimeState, UmbracoHelper umbracoHelper) : base(globalSettings, umbracoContextAccessor, sqlContext, services, appCaches, logger, runtimeState, umbracoHelper)
+        {
+            _treeSearcher = treeSearcher;
+            _actions = actions;
+        }
+
```
ronaldbarendse: This constructor should be moved down a little more, so it's below the `UserStartNodes` property (keeping the order of fields, properties, constructors, methods).

src/Umbraco.Web/Trees/ContentTreeControllerBase.cs
Shazwazza: preventing breaking changes in case people are inheriting from this controller.

src/Umbraco.Web/Trees/ContentTreeController.cs
Shazwazza: That's not really a policy we enforce, I'm also not a huge fan of putting constructors below properties.

ronaldbarendse: Keeping the fields/properties together should at least be preferred, even without policies or guidelines 😉 So currently `_userStartNodes` should go up, just below `_treeSearcher` and `_actions` and the property `UserStartNodes` just above the constructor... Just to keep things tidy 😅

Shazwazza: aha, sorry missed that _rushing to get a PR done for hacktoberfest_ ;)

PR #6531, opened 2019-10-01
Use umb-checkbox in template section

bjarnef:
### Prerequisites

### Description
This PR replace the traditional checkbox in template insert field and section mandatory option with `umb-checkbox` and adjust the UI a bit to make it more consistent with the rest of the backoffice UI.


src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/templatesections/templatesections.controller.js
```diff
@@ -1,26 +1,36 @@
...
             } else {
                 $scope.model.insertType = 'renderBody';
             }
+
+            var labelKeys = [
+                "template_sectionMandatory"
+            ];
+
+            localizationService.localizeMany(labelKeys).then(function (data) {
+                vm.labels.sectionMandotory = data[0];
```
abjerner: Hi @bjarnef. I haven't tested the code yet, but spotted `sectionMandotory`. Could you change the occurences to `sectionMandatory`?

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/templatesections/templatesections.html
```diff
@@ -44,10 +44,10 @@
                             </div>
             
                             <div>
-                                <label>
-                                    <input type="checkbox" ng-model="model.mandatoryRenderSection" /> <localize key="template_sectionMandatory" />
-                                </label>
-            
+                                <div class="flex">
+                                    <umb-checkbox model="model.mandatoryRenderSection" text="{{vm.labels.sectionMandotory}}" />
```
abjerner: Typo here as well.

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/templatesections/templatesections.controller.js
bjarnef: Thanks @abjerner ... I have now fixed these typos ;)

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/insertfield/insertfield.html
```diff
@@ -33,10 +33,9 @@
 
                     <!-- Default value -->
                     <div>
-                        <div style="margin-bottom: 20px;">
-                            <i class="icon icon-add blue" ng-hide="vm.showDefaultValue"></i>
-                            <a href="" ng-click="vm.showDefaultValue=true" ng-hide="vm.showDefaultValue"><localize key="templateEditor_addDefaultValue">Add default value</localize></a>
-                        </div>
+                        <a prevent-default href="" class="umb-node-preview-add mb3" ng-click="vm.showDefaultValue=true" ng-hide="vm.showDefaultValue">
```
nul800sebastiaan: Should this be a `<button>` instead, since it is not used as an actual anchor link?

bjarnef: It probably could be.. it has an empty href and this is how most of add for picker with `umb-node-preview-add` works.
https://github.com/umbraco/Umbraco-CMS/search?q=umb-node-preview-add&unscoped_q=umb-node-preview-add

If we should change this, we should probably change it many other places as well :)

nul800sebastiaan: @bjarnef - quite a few nice accessibility PRs you could put in then in the future! 

Make sure to update this one for now, our visually impaired users will thank you for it! 👍

nul800sebastiaan: @bjarnef I think you can also just give it `role="button"`, that should be enough for now, see (for example): https://github.com/umbraco/Umbraco-CMS/pull/6555/files

bjarnef: @nul800sebastiaan it seems updating these `add` buttons to use `button` element is more correct. Maybe @BatJan want to submit a PR to update this in the different parts of the backoffice UI?

emmaburstow:
@nul800sebastiaan - is this good to merge?

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/insertfield/insertfield.html
```diff
@@ -52,11 +52,11 @@
                     <div class="control-group umb-control-group">
                         <div class="umb-el-wrap">
                             <div class="controls">
-                                <label class="control-label" >
+                                <label class="control-label">
                                     <localize key="templateEditor_recursive">Recursive</localize>
                                 </label>
+                                <umb-checkbox input-id="recursive" name="recursive" model="vm.recursive" disabled="!vm.field" />
```
BatJan: There is no need for the input-id and name attributes nor the label referencing it - It can be done like below where the label-key is also passed to the directive so it will be translated. Currently there will be 2 labels doing the same thing 😃

```
                                <umb-checkbox
                                    model="vm.recursive"
                                    text="Yes, make it recursive"
                                    label-key="templateEditor_recursiveDescr">
                                </umb-checkbox>
```

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/insertfield/insertfield.html
nul800sebastiaan: @bjarnef I was wrong and I've been told it needs to be a proper button! Could you update that so that @BatJan doesn't have to go in and correct it again? :-)

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/insertfield/insertfield.html
bjarnef: @BatJan at the time of submitting this PR the `label-key` didn't exist.
Sometime you might want to use `input-id` to make other than the wrapping label toggle the checkbox.

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/insertfield/insertfield.html
bjarnef: @nul800sebastiaan I can do that, but shouldn't we submit a different PR to update this in all another parts of the UI where it also use this format? E.g. here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/content/protect.html#L50

Just do a search for `umb-node-preview-add` 😁

nul800sebastiaan: @bjarnef I love your enthusiasm to fix all the things, but if we do it in small steps then it's much easier to handle! So I would like to keep this small and then move on to the next area. Make sense?

bjarnef: @nul800sebastiaan sure, I can update this anchor to a `button` element, but it wasn't really was this PR is about :)
And it just use the same markup as in this rest of the code. So once this was merged, it would be easier to submit a new PR to change the `<a prevent-default href="" class="umb-node-preview-add">` to `<button class="umb-node-preview-add">`

bjarnef: @nul800sebastiaan all fixed now 👍 🌮

nul800sebastiaan:
Thanks very much @bjarnef, a bit of a journey there but we got there! And thanks @abjerner and @BatJan for the feedback!

PR #6533, opened 2019-10-01
Split events.directive into separate event directives

BatJan:
### Prerequisites

### Description
I noticed that the events.directive.js is one big file chaining some of the event directives being used in the backoffice. In this PR I have refactored this so each event has it's own directive, which aligns with the IIFE approach used elsewhere in the backoffice and also makes it easier to write some proper documentation for each of the events I can imagine.

Also makes it easier to add new event directives in the future should the need arise.

Looking forward to receive any feedback you might have for this suggestion 😃


nul800sebastiaan:
Thanks @BatJan , we'll have a look soon! Bit of a bigger looking change, so not sure how quickly we can get to it. :-)

nul800sebastiaan:
Great, thanks @BatJan ! 👍

PR #6535, opened 2019-10-01
V8: Change the sort order of the default media list view

kjac:
### Prerequisites

### Description

The default media list view sort order is the opposite of the tree sort order:

media-default-listview-sort-before

This PR changes the sort order of the default media list view to match that of the tree:

media-default-listview-sort-after

The change is implemented in the installer, so current installs will not be affected by this change.


nul800sebastiaan:
Hmm, interesting, I actually have no idea what happens with a list view in the content area.. the last updated item is also shown first by default, no?

nul800sebastiaan:
I don't think this is the desired behavior @kjac - if you already have a 100 items here and update a few more, they will disappear immediately after upload on the last page of paging. This overview was meant to give you very quick access to the items you just uploaded so you could start cropping them etc. 

I am going to close this for now, but let me know if there's any other good arguments to update the sorting.

PR #6536, opened 2019-10-01
Too much scope

rasmusjp:
Noticed that I've added `$scope` in my last PR where it should't be, this fixes it.


nul800sebastiaan:
Descope all the things! 👍

PR #6538, opened 2019-10-01
Refactor language picker to improve accessibility

BatJan:
### Prerequisites

### Description
In this PR I have done the following
- Changed `<div>` and `<a>` to `<button>` to fix keyboard navigation and semantics
- Adjusted the styling so the buttons are styled correctly
- Added the `aria-haspopup` and `aria-expanded` attributes
- Added visual hidden texts for providing better non-visual context so a screen reader announces "Current language: English (United States)" instead of just "English (United States)" for instance
- Added the deep-blur directive in order to have the dropdown disappear when it looses tab-focus

The visual changes from before and after can be seen below

**Before**
language-switcher-before

**After**
language-switcher-after

I look forward to receive any feedback you might have for this 👍

### DIFF
--- a/src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
+++ b/src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
@@ -1636,7 +1636,7 @@ Mange hilsner fra Umbraco robotten
     <key alias="openBackofficeSearch">Åben backoffice søgning</key>
     <key alias="openCloseBackofficeHelp">Åben/Luk backoffice hjælp</key>
     <key alias="openCloseBackofficeProfileOptions">Åben/Luk dine profil indstillinger</key>
-    <key alias="currentLanguage">Aktivt sprog:</key>
-    <key alias="switchLanguage">Skift sprog til:</key>
+    <key alias="currentLanguage">Aktivt sprog</key>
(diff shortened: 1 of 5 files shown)


src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
```diff
@@ -1636,5 +1636,7 @@ Mange hilsner fra Umbraco robotten
     <key alias="openBackofficeSearch">Åben backoffice søgning</key>
     <key alias="openCloseBackofficeHelp">Åben/Luk backoffice hjælp</key>
     <key alias="openCloseBackofficeProfileOptions">Åben/Luk dine profil indstillinger</key>
+    <key alias="currentLanguage">Aktivt sprog:</key>
+    <key alias="switchLanguage">Skift sprog til:</key>
```
bjarnef: @BatJan I would probably just add the colons in the view/controller instead. It would make it easier to reuse these keys in other areas.

BatJan: Yes, good point - Done now 👍

nul800sebastiaan:
Thanks @BatJan - we'll have a look soon!

nul800sebastiaan:
Awesome, thanks very much @BatJan !

PR #6544, opened 2019-10-01
Replace checkbox with umb-checkbox in listview publish and unpublish overlay

bjarnef:
### Prerequisites

### Description
I noticed when a document type vary by culture and there is multiple languages configurated, the listview publish and unpublish used traditional checkboxes, while the general unpublish and publish with descendants overlays use `umb-checkbox` component.

Furthermore the unpublish button in listview overlay is `warning` style as the general unpublish overlay.

image

image

I am not sure if it is a bug that it has a condition `ng-if="vm.languages.length > 1"` in "listviewpublish" while it has `ng-if="vm.languages.length > 0` in "listviewunpublish"?

https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/propertyeditors/listview/overlays/listviewpublish.html#L7

https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/propertyeditors/listview/overlays/listviewunpublish.html#L9


bjarnef:
With only a single language there is this difference between "listviewpublish" and "listviewunpublish" regarding the `vm.languages.length > 0` vs `vm.languages.length > 1`:

image

image

nul800sebastiaan:
Looks nice, thanks Bjarne, we'll have a look soon!

nul800sebastiaan:
Good point, it was always supposed to only show the checkboxes if there are multiple choices, so I updated `listviewunpublish.html` to check for `vm.languages.length > 1` instead of 0 👍

nul800sebastiaan:
Great update, thanks @bjarnef! 👍

Issue #6275, opened 2019-09-03
Ensure all helper methods properly HTML encode text


ronaldbarendse:
After reporting the `ReplaceLineBreakForHtml` method didn't properly HTML encode the text (https://github.com/umbraco/Umbraco-CMS/issues/6226), making it hard to use and therefore easily prone to XSS attacks, @sniffdk noticed this might also be the case with the `Truncate` method (https://github.com/umbraco/Umbraco-CMS/pull/6235#issuecomment-527029498).

Having looked at it, the `Truncate`, `TruncateByWords` and `StripHtml` methods all expect HTML input (the parameter name is `html` and they have overloads accepting `IHtmlString`), so these might only need some clarification not to pass in unencoded text.

A quick search reveals a lot of other methods also accept text input as `string` and return it **without HTML encoding it** as `HtmlString`/`IHtmlString`:
- `HtmlHelperRenderExtensions`
  - `If`
  - `ReplaceLineBreaksForHtml` (should be replaced by `ReplaceLineBreaks`, as done on `StringUtilities` in https://github.com/umbraco/Umbraco-CMS/pull/6235)
- `PublishedContentExtensions`
  - `IsEqual`
  - `IsNotEqual`
  - `IsDescendant`
  - `IsDescendantOrSelf`
  - `IsAncestor`
  - `IsAncestorOrSelf`
- `IndexedArrayItem`
  - `IsFirst`
  - `IsNotFirst`
  - `IsIndex`
  - `IsModZero`
  - `IsNotModZero`
  - `IsNotIndex`
  - `IsLast`
  - `IsNotLast`
  - `IsEven`
  - `IsOdd`
- `HtmlTagWrapperTextNode` (used on `Html.Wrap()`)

The documentation also doesn't mention how to correctly use these methods: by making sure you manually HTML encode the input before passing it into the method. And even if it did, this can easily be forgotten.

nul800sebastiaan:
Honestly, I can't see a way to do this properly in a non-breaking way, without having ugly overloads of all of these methods. 

I also can't see that we would be willing to introduce a breaking change by fixing all of these methods up in one go. 

While I do recognize the possibility of a XSS problem occurring, this also means that you have a malicious content editor providing this content or someone specifically targeting an editor to get the script in the CMS somehow. The scenario is rather unlikely. 

I'm not sure what to do here so I'll take it to the team and see what their recommendation is.

sniffdk:
[...]

PR #6545, opened 2019-10-01
Properly HTML encode text in helper methods

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6275

### Description
This PR makes sure most helper methods accepting `string` parameters and returning `IHtmlString` properly HTML encode the text. As described by @Shazwazza in https://github.com/umbraco/Umbraco-CMS/issues/6275#issuecomment-532075299, it also adds documentation comments to let people know this is done (so you don't double-encode the text).


ronaldbarendse:
A lot of the helper methods on `IndexedArrayItem` are redundant, e.g. all the `IsNot*()` methods (just negate the return value or switch the parameters `valueIfTrue` and `valueIfFalse`) and the overloads containing only the `valueIfTrue` parameter (these could be removed by adding a default value to `valueIfFalse`).

Just having the `IsFirst()`, `IsLast()`, `IsIndex()`, `IsModZero()`, `IsEven()` and maybe even `IsOdd()` with the boolean return value and another with both optional parameters `valueIfTrue` and `valueIfFalse` set to `null` would be enough. And if these changes would be implemented, the return type could be changed from `HtmlString` (which should be `IHtmlString` to begin with) to `string` and therefore not require HTML encoding the values anymore.

For now I've kept all the existing method signatures in this PR, so we don't introduce binary breaking changes, only functional ones (as described in the already linked [comment](https://github.com/umbraco/Umbraco-CMS/issues/6275#issuecomment-532075299)).

nul800sebastiaan:
I am still not sure if we can just switch this up on people without causing some serious pain. Maybe @Shazwazza will have some more input on this one.

ronaldbarendse:
@nul800sebastiaan The linked comment already contains @Shazwazza's input and the worst thing that could happen is double-encoding the text... This is not so serious compared to the potential XXS issues (if you're not manually encoding the text before passing it into these methods) and fixing would mean only removing your own call to `HttpUtility.HtmlEncode()`.

ronaldbarendse:
Just found out `@Html.Wrap()` also doesn't encode the `innerText` value!

With this one it was a little harder to see why, as it first creates an `HtmlTagWrapperTextNode`:

https://github.com/umbraco/Umbraco-CMS/blob/06b136fdf66ae685c8985af90ac075535e635e74/src/Umbraco.Web/HtmlHelperRenderExtensions.cs#L807-L823

And this instance is later written to an output stream without HTML encoding the text:

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web/Mvc/HtmlTagWrapperTextNode.cs#L5-L17

nul800sebastiaan:
@ronaldbarendse What does your last comment mean for this PR? I am not sure how it's related?

ronaldbarendse:
It means I've found another helper method (`Html.Wrap()`) that doesn't properly HTML encode text, so that should also be fixed (either in this PR or as a new one).

nul800sebastiaan:
Thanks! I'll have to ask the team if we want to make all these breaking changes.

One thing we can't do is mark methods as obsolete without having an alternative for them. Maybe to you they don't make sense and I know we're probably violating the c# programming bible, but these methods are there, they're convenience methods we like to provide. 

In any case, I need to get someone at HQ to figure out if we want these changes and when and how we can make it clear that everybody will need to start cleaning up their views after upgrade. For example, on Umbraco Cloud with automatic upgrades we'll just straight-up be breaking people's views if we roll this out, so I am not very hopeful we can do them in this way.

nul800sebastiaan:
Ah, I just realized that we don't auto-upgrade minor releases on Cloud, so it's opt-in and they need to test on a separate environment, that makes it a lot lower in impact.

ronaldbarendse:
As @Shazwazza commented in the linked issue https://github.com/umbraco/Umbraco-CMS/issues/6275#issuecomment-532075299 (referenced in this PR), fixing potential XSS issues is more important then double-encoding text (if you use something like `Model.IsDescendant(child, HttpUtility.HtmlEncode(someValue))`) or encoding HTML (if you use these methods like `Html.Raw()`, e.g. expecting the text to be rendered as HTML).

This should be released as breaking change in a new minor, with the advise to check custom code/views that uses the methods listed in the issue.

Regarding the obsoleted methods: you can still use them and the alternative is to use simple conditionals, e.g. instead of `Html.If(Model.SomePropertyToCheck, "Render this text when the property is true")`, write:
```
@if (Model.SomePropertyToCheck) {
   <text>Render this text when the property is true</text>
}
```
These convenience methods aren't documented well and are completely custom to Umbraco, so I don't see any real advantage to having them compared to the standard, generic, platform independent Razor code.

nul800sebastiaan:
Alright, I've discussed this some more and it's good to get this encoding issue out of the way for once and for all. 

In the future, I'd like to ask you not to move all the methods around like you did with `BeginUmbracoForm`. Not only is it impossible for me to see what is changed, it's also not necesary.(yes, I know it looks more pleasing if they're in the right order 🙈). After spending time re-ordering I could finally see that you changed the value for the `htmlAttributes` parameter from `new Dictionary<string, object>()` to `null`. However, those method calls eventually propagate to this code: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/HtmlHelperRenderExtensions.cs#L746 and there is no null check here, so it will fail. I reverted all of those changes for now.

I also turned all the `Obsolete`s into `TODO`s. For a bit of background, those helper methods were once made to help people inexperienced with MVC / Razor to move from XSLT to Razor more easily. They're definitely legacy now and should be moved or removed. They're Friendly, in other words, and yes, they're not very nice where they are right now. The joys of legacy! 
We never want to `Obsolete` something without having an alternative available, that's just confusing and not helpful. So I've turned them into `TODO`s, I don't know if we want to keep there around for vNext but if we do it might be good to move their logic to `HtmlHelper` and then they can be `Obsolete`d.

Sorry it took quite a while to get this reviewed, but we're finally there! Thanks for all the work on this!

I will mark this as a breaking change since people might have been relying on the previous behavior of not encoding the HTML.

As @Shazwazza tweeted (I believe related to this), XKCD made an excellent cartoon for these types of breaking changes.. 😅 https://twitter.com/Shazwazza/status/1271005573060849665

PR #6550, opened 2019-10-02
V8: allow node preview content to wrap

### LINKED ISSUE #5916: Nested Url Picker Spacing issue
STR: 

Log in to CMS
Add nested content item
Add another nested content item within above
add a url picker
ER: 

Component spacing should remain consistent 
AR: 

see screenshot 
image

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #5916 

### Description
On small screens, the content in the umb-node-preview directive gets all squished because the parent container is a flex element, but doesn't allow wrapping. This change adds flex-wrap:wrap, and updates the view to keep the icon and node name grouped, while allowing the actions (open, edit, remove) to wrap to a new line.

Large-ish screen is unchanged:
chrome-capture

Small screen wraps like so (this is a super-small screen example):
chrome-capture (1)

Have verified this in nested content and the content picker editor.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-node-preview.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-node-preview.less
@@ -3,6 +3,7 @@
     display: flex;
     box-sizing: border-box;
     border-bottom: 1px solid @gray-9;
+    flex-wrap: wrap;
 }
 
 .umb-editor-wrapper .umb-node-preview {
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
That looks nice! Thanks, we'll have a look soon!

nul800sebastiaan:
Good stuff, thanks @nathanwoulfe! ⭐️

PR #6554, opened 2019-10-02
Fixes the media refs within the RTE in the Grid

Shazwazza:
This is sort of a regression from 8.1 where when we updated the media references within the RTE, the values would be persisted with UDI references and without the actual media paths.

Unfortunately the code for the Grid's RTE's was not updated so this never took place for the Grid.

The images stored in the DB should look something like this within the html:

```html
<img src="?width=500&amp;height=232.00514138817482" alt="" width="500" height="232.00514138817482" data-udi="umb://media/b6891d9fc71c40ffa00931fb64c2ae4d" />
```

But for the grid rte it would still maintain the path:

```html
<img src="/media/m2njyrqd/image.png?width=500&amp;height=302.08333333333337" alt="" width="500" height="302.08333333333337" data-udi="umb://media/469c1a6112034b8a81767e486dd2ed76" />
```

When the RTE is re-rendered in the back office, the `src` attribute is then re-updated with the correct media path. This was also not happening in the Grid's RTE.

This PR fixes this issue so now the RTE and the Grid RTE behave the same way.

 You can test this by:

Inserting an image into a normal RTE and save/publish, go check your DB:

```sql
SELECT TOP (1000) [id]
      ,[versionId]
      ,[propertyTypeId]
      ,[languageId]
      ,[segment]
      ,[intValue]
      ,[decimalValue]
      ,[dateValue]
      ,[varcharValue]
      ,[textValue]
  FROM [v80].[dbo].[umbracoPropertyData]
  ORDER BY id DESC
```

You'll be able to find your value near the top of this list in the `textValue` and it should have a `src` attribute without the media path

And of course in the back office the image should still render just fine.

Then perform the same test with the Grid's RTE. The behavior should be the same.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
@@ -1188,6 +1188,7 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
             var tinyMceRect = editor.editorContainer.getBoundingClientRect();
             var tinyMceTop = tinyMceRect.top;
             var tinyMceBottom = tinyMceRect.bottom;
+            var tinyMceWidth = tinyMceRect.width;
 
             var tinyMceEditArea = tinyMce.find(".mce-edit-area");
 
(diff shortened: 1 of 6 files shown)


bergmania:
Works as expected now :)

PR #6555, opened 2019-10-02
Labeled controls in Belle

zersiax:
### Prerequisites

### Description

Throughoutthe Umbraco backoffice, there are a lot of places where iconography is the only mechanism to explain what a control does. Screen readers cannot infer any information from this way of labeling controls.

This PR:

* Labels the media picker icons that make the mediapicker appear, as well as the icon to add a new folder 

* Labels the edit, delete and lock controls for the property editors in the document type screen.


nul800sebastiaan:
Thanks very much @zersiax - we'll give this a spin soon, but it seems good!

soreng:
Can/should these labels be localized @zersiax ?

zersiax:
they probably can be. If they should be is up to the team I think ...I've seen labels in the wild that weren't and others that were, so I'm not sure what the official stance is on accessibility labels and localization is currently

nul800sebastiaan:
Translating them is optional IMO, it's going to be a lot more work to do as the translations will need to  be added to the viewmodel and sent into the view. For now, a first round of improvement here is good enough and we're happy to look at localization later! 

Thanks again @zersiax - two PRs merged in a day, keep up the good work! 👍

Issue #6557, opened 2019-10-02
v7 ExpireDate in ShowMessageForPublishStatus transformed to LocalTime


PiotrKlys:
### Bug summary
Hi,

I've noticed an issue connected with the message after publishing: 

![image](https://user-images.githubusercontent.com/4973998/66045958-04e24880-e525-11e9-860f-eb55b8aaafdb.png)
 
ExpireDate used in ShowMessageForPublishStatus from ContentController is not converted to LocalTime so there is a possibility to have different time in the message and on the local machine.

### Specifics

* V7.15.1.2
* Browsers invariant

### Steps to reproduce

* Try to set ExpireDate in UTC

### Expected result

* ExpireDate should be shown in LocalTime

### Actual result

* ExpireDate is not converted to the LocalTime

stevemegson:
This problem is also discussed in #6357. It's difficult to fix because the server only knows its own time zone, so it can't convert to the client's local time. We'd need to give the client some way to recognise that the message contains a date and time, and adjust it to local time there.

nul800sebastiaan:
Yep, this is a duplicate of #6357 - I'll close this one. 👍

Issue #6357, opened 2019-09-17
v7.13 Content saved with release date notification message displays server time not client time


samtruedig:
Content saved with scheduled publish date - notification message displays server time not client time.

Reproduction
------------

### Specifics
v7.14

### Steps to reproduce
Server time zone must be different to client.
Save content with a future publish date. 
Notification will popup "Content saved: Changes will be published on DD Mon YYY at hh:MM" 

### Expected result
Expect time to be localized to client

### Actual result
Time is localized to server

introduced in commit https://github.com/umbraco/Umbraco-CMS/commit/fb094f46befe99433cebed345c931734de6d1778

nul800sebastiaan:
Indeed, I can reproduce this:

![image](https://user-images.githubusercontent.com/304656/65034921-44673d00-d948-11e9-8a04-ad0ec6282c57.png)

We'd be happy for some help to fix this. The problem, however, is not exactly caused by fb094f4 - that's only where it gets displayed for the first time. 

I am afraid that we can't do the conversion on the server, since it doesn't know what the user's timezone is. This might get complicated, the only thing I can see is to have some magic string in the notification (something like `Changes will be published on %0% at ###%1%###`) and then parsing the text in between the `###` to use `dateHelper.getLocalDate` on that (which we use elsewhere to convert to a local timezone. Of course this gets problematic if both the date and the time need to be updated (because it's after midnight on the local user's machine for example).

Since the datepicker in the info tab will show the time in the local timezone we can alternatively show a slight less specific text. So we could change the text to `Changes will be published on the scheduled date and time` / `and visible on the website until the scheduled date and time`.

umbrabot:
Hi @samtruedig,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

umbrabot:
Hiya @samtruedig,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

[...]

PR #6558, opened 2019-10-02
v7 ExpireDate in ShowMessageForPublishStatus transformed to LocalTime…

PiotrKlys:
… (#6557)

### Description
I've added ToLocalTime methods to ExpireDate to ensure that this time will be presented in LocalTime


nul800sebastiaan:
Thanks! I didn't see this one related to #6557 but let's relate it to #6357 instead then! :-)

We'll have a look soon!

nul800sebastiaan:
Thanks again @PiotrKlys - unfortunately this just won't work, it will show the local time of the server, which is what we're already showing. Somehow we need to combine the server / client time offset with the time that is returned from the server.

PR #6559, opened 2019-10-02
add html in css

tonzsm:
### Prerequisites


nul800sebastiaan:
Hi @tonzsm - what does this fix and why please? I'm not sure what the purpose is..

tonzsm:
i kind this html attr can use in css it  work

nul800sebastiaan:
Hi again @tonzsm - I am not sure why you would add this, can you please explain in detail what this fixes?

I'm going to close this for now, but we can re-open it if we see this actually fixes something real. 👍

PR #6461, opened 2019-09-27
enables infinite editing for members

skttl:
I am trying to enable infinite editors for Diplo God Mode, and noticed that members didn't support this. So I added that, in this PR 🙂

PR #6560, opened 2019-10-02
V8: Enable editing the search results directly from Examine Management

kjac:
### Prerequisites

### Description

This PR enables direct access to the search results in Examine Management simply by clicking the result name. It works like this:

examine-search-open-results

With the upcoming ability to search for GUIDs, this might also come in really handy for Cloud support handling.

It only works for well known entities - for now that's content, media and members.

#### But wait! Isn't this a security concern?

Well.. no. Anyone having access to Examine Management can see the (raw) values of whatever is found when searching, and they can easily construct the URLs for editing the results by means of these values. 

#### Testing this PR

When testing this PR, verify that you can open content, media and member search results by clicking their names. 

Also test that you can open search results in a new tab by Ctrl-clicking their names.

Lastly someone should probably test this using a "configured searcher". Apparently there are none of those by default in vLatest and I have no idea how to add them 🤦‍♂ 

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/editor.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/editor.service.js
@@ -898,6 +898,27 @@ When building a custom infinite editor view you can use the same components as a
             open(editor);
         }
 
+        /**
+         * @ngdoc method
+         * @name umbraco.services.editorService#memberEditor
+         * @methodOf umbraco.services.editorService
+         *
+         * @description
+         * Opens a member editor in infinite editing, the submit callback returns the updated member
+         * @param {Object} editor rendering options
+         * @param {String} editor.id The id (GUID) of the member
+         * @param {Boolean} editor.create Create new member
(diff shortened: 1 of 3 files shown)


skttl:
What a great addition - but shouldn't it be opening in an infinite editor?

nul800sebastiaan:
That's pretty cool, I assume the green link still opens the Examine result?

I don't think it should necessarily open an infinite editor, but that could be the next step.

We'll have a look soon!

kjac:
@skttl I really like the idea of opening in infinite editing (unless you ctrl-click the link). But I don't think it'd work for members, far as I know there's no infinite editing of those?

skttl:
@kjac #6461 😉😜

kjac:
@skttl would you look at that! I'll dig into it.

kjac:
@skttl totally worth it! Here's how it works now:

examine-management-open-results

skttl:
So much better ❤

nul800sebastiaan:
So cool!! 🎉 ⭐️ 

Great and useful update! 👍

PR #6456, opened 2019-09-26
In addition to #6236 - now this now longer keeps appending the conten…

nul800sebastiaan:
…t name on each change of the language dropdown

image

PR #6563, opened 2019-10-02
v8: Added a test for if an empty titlePrefix is passed into the $changeTitle

RachBreeze:
In addition to #6456 if an `changeTitle` is now called with an empty value for `titlePrefix` the value in `originalTitle` removing the "-" prefix on  the title element


nul800sebastiaan:
Nice one @RachBreeze - we'll give it a spin soon! :)

PR #6566, opened 2019-10-03
Respect UseHttps setting in MembershipHelper

### LINKED ISSUE #6553: Umbraco.Core.UseHttps: true setting not honoured for HTTPS site.
The above setting is not honoured on my https site on Azure.
When you login using Chrome and look at the cookies, you will see that the default authentication cookie "yourAuthCookie" is only mar
[Umbraco does not honour Umbraco.Core.UseHttps true.pdf](https://github.com/umbraco/Umbraco-CMS/files/3680884/Umbraco.does.not.honour.Umbraco.Core.UseHttps.true.pdf)
ked as HttpOnly - but the Secure flag is not checked.
I display the value of this setting on one of my page so I am sure it it set to true - see screenshot attached.

stevemegson:
If there's an existing issue for this PR then this fixes #6553

### Description
MembershipHelper doesn't respect the Umbraco.Core.UseHttps setting, since it relies on FormsAuthentication.SetAuthCookie to set the cookie.

This PR uses FormsAuthentication.GetAuthCookie instead, so we can set the Secure property before adding the cookie to the response.

### DIFF
--- a/src/Umbraco.Tests/Routing/RenderRouteHandlerTests.cs
+++ b/src/Umbraco.Tests/Routing/RenderRouteHandlerTests.cs
@@ -146,7 +146,7 @@ namespace Umbraco.Tests.Routing
             var handler = new RenderRouteHandler(umbracoContext, new TestControllerFactory(umbracoContextAccessor, Mock.Of<ILogger>(), context =>
             {
                 var membershipHelper = new MembershipHelper(
-                    umbracoContext.HttpContext, Mock.Of<IPublishedMemberCache>(), Mock.Of<MembershipProvider>(), Mock.Of<RoleProvider>(), Mock.Of<IMemberService>(), Mock.Of<IMemberTypeService>(), Mock.Of<IUserService>(), Mock.Of<IPublicAccessService>(), Mock.Of<AppCaches>(), Mock.Of<ILogger>());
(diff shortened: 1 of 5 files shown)


John-Blair:
Thanks - it does the job, but the way you have coded it now makes the  Umbraco.Core.UseHttps flag redundant as you always make it secure for a HTTPS connection. suggest you remove this setting from web.config and always make the other cookies secure automatically if HTTPs is detected.

stevemegson:
The Umbraco.Core.UseHttps setting isn't redundant because it also enforces the use of HTTPS in other places. However, I think you're right that there's actually no need to check the setting for the cookies. I was copying the logic that's used for the backoffice cookie, but in practice we only need to check whether the current request is secure.

nul800sebastiaan:
Thanks, we'll have a look soon!

nul800sebastiaan:
Aha, I remember now.. we should not do any of this, you need to set this up in your web.config to set frontend cookies to secure: https://docs.microsoft.com/en-us/dotnet/api/system.web.security.formsauthentication.requiressl?view=netframework-4.7.2

If `requireSSL="true"` then `FormsAuthentication.SetAuthCookie` will set the secure flag.

PR #6567, opened 2019-10-03
V8: Don't allow deleting property groups with locked properties

kjac:
### Prerequisites

### Description

Oh my. It's possible to delete property groups that contain locked properties, despite not being able to delete the individual properties:

group-locked-1

It used to be so you couldn't delete the property group unless it was empty, but I guess somewhere along the line this was changed (which makes sense).

Anyway. If you go ahead and delete the locked properties on a member type, members of this type start failing miserably (as one might expect):

group-locked-2

So... this PR removes the ability to delete a property group if it contains one or more locked properties:

group-locked-3

#### Testing this PR

1. Verify that you're not able to delete property groups with locked properties (i.e. on member types).
2. Verify that you're still able to delete property groups *without* locked properties (i.e. on document types).

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbgroupsbuilder.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbgroupsbuilder.directive.js
@@ -418,6 +418,10 @@
 
       };
 
+      scope.canRemoveGroup = function(group){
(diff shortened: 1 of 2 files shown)


emmaburstow:
Hey @kjac - this looks ace. We'll shout if we need a thing. Em

nul800sebastiaan:
Nice! Thanks @kjac! 👍

PR #6572, opened 2019-10-03
V8: Remove "Generic Properties" group from default member type

kjac:
### Prerequisites

### Description

The default member type created on install still has the good ol' "Generic Properties" group, but only for a few properties:

image

When you create a new member type, all properties go to the "Membership" group:

image

This PR fixes it up so all membership properties are created in the "Membership" group during install:

image

Note: This only works for new installs, I doubt it'd be a great idea to create a migration for this 😆


nul800sebastiaan:
Nice one! We'll have a look soon! 👍

Shazwazza:
@nul800sebastiaan please make sure you don't merge things that are failing the build, this has failing unit tests. I fear there are others too since we have a lot of random tests failing in dev/v8 at the moment :/

nul800sebastiaan:
@Shazwazza Ah crap, I thought this was submitted when all the builds failed because of something else, so I didn't notice.

PR #6573, opened 2019-10-03
V8: Help users create their first content type on blank installs

kjac:
### Prerequisites

### Description

On a brand new install with no document types created, Umbraco doesn't exactly help you a whole lot:

create-1

...unlike when you have a document type with no allowed child document types:

create-2

This PR adds a similar functionality to the content root (of course provided you're allowed to access the Settings section). I think this is more friendly and helpful - specially for newcomers:

create-at-root-after

While I was at it I also added a loading indicator while the "Create" dialog loads, as it is a bit erratic as-is while things are loading.


nul800sebastiaan:
I was going to ask: "but only if they have access to the Settings section" .. and then I looked at the code.

Looks good, making Umbraco simpler! We'll give it a spin soon! 👍

nul800sebastiaan:
What a great update, cheers @kjac! ⭐️

Issue #6477, opened 2019-09-30
Unable to sort by last edited date


AndyFelton:
Unable to sort members list by the last edited date.

Reproduction
------------

Clicking on the label on the table heading for all fields except the Last Edited label causes the listing to be ordered. For the Last Edited label it does nothing.

### Specifics
Reproduced on Chrome 77.0.3865.90 using Umbraco version 7.10.5 and 7.13.2

### Steps to reproduce
1. Click on the label for Last Edited on the Members Listing

### Expected result
I would expect the arrow to be displayed and the list reordered as the other columns do

### Actual result
No change in the order of the list

nul800sebastiaan:
Hmmm, I am not sure if we're supposed to be able to do that, I'll ask around!

AndyFelton:
If it shouldn't be clickable then maybe the hover cursor is incorrect forgot to mention it was a pointer! Let us know HQ's thoughts on this one.

AndyButland:
It seems to be expected to be non-clickable, there's some code in `listview.controller.js` (see comment beginning `Special case for members...`, that means clicking has no effect.

The PR above I've update the cursor to make it more obvious that the header isn't clickable.

Might also take a look at seeing if this header _could_ be sortable, but the above PR would be useful anyway for other list views.

AndyButland:
Seems like it can be without too much effort, so have implemented that in the second PR above.

RachBreeze:
Thank you, one of our clients would appreciate this as they use last edited as part of their member maintenance work flow (with all members being maintained in the back office)

AndyFelton:
Thanks that'll make my client happy if it does get merged.

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6575

AndyFelton:
@nul800sebastiaan This issue was raised against V7 any chance up it being included in a future update on V7?

AndyFelton:
Fixed for V7 in [#7088](https://github.com/umbraco/Umbraco-CMS/pull/7088)

PR #6575, opened 2019-10-03
Allows members to be ordered by additional system fields

AndyButland:
It was raised in #6477 that the last edited field on the default members list view doesn't allow sorting.

This PR adds that ability, along with two other system fields that developers are able to add to the list view - the created date, and the content type alias.

Also fixes a minor issue that when adding the content type alias as a header on the member list view, it was displaying the localised text of "media type" instead of "member type".


nul800sebastiaan:
Nice one! We'll give it a spin soon! 👍

nul800sebastiaan:
Awesome, thanks very much for the help @AndyButland!

PR #6579, opened 2019-10-04
Add ability for the main Umbraco search to include media file names

Jeavon:
The PR adds a new field to the MediaValueSet called `"__umbracoFile"`, `"__"` prefixed as this is to be used internally only.  
The new field is populated either by extracting the Src property from the Image Cropper Json or directly from the property in the case of files such as PDFs.
We have also updated the UmbracoTreeSearcher to include the new field and modified the mapper to include the filename in the search results in braces.


bjarnef:
@Jeavon should we add this as a constant here?
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Examine/UmbracoExamineIndex.cs

and then use it like this:
```
{UmbracoExamineIndex.UmbracoFileFieldName, new object[] {umbracoFile}}
```

Jeavon:
@bjarnef yes we should, will add it

src/Umbraco.Examine/MediaValueSetBuilder.cs
```diff
@@ -29,6 +34,28 @@ public override IEnumerable<ValueSet> GetValueSets(params IMedia[] media)
...
+                    {
+                        umbracoFilePath = cropper.Src;
+                    }
+                }
+                else
+                {
+                    umbracoFilePath = umbracoFileSource;
+                }
+
+                var uri = new Uri(_runtimeState.ApplicationUrl.GetLeftPart(UriPartial.Authority) + umbracoFilePath);
```
Shazwazza: Null or empty checks are needed for umbracoFilePath before using it. And then below when adding to the index, it should use the same notation as the other fields and use the Yield ext method else insert an empty array not an array with an empty element

src/Umbraco.Web/Search/UmbracoTreeSearcher.cs
```diff
@@ -85,6 +85,7 @@ public class UmbracoTreeSearcher
                     break;
                 case UmbracoEntityTypes.Media:
                     type = "media";
+                    fields = new[] { UmbracoExamineIndex.UmbracoFileFieldName };
```
Shazwazza: Is this now replacing all of the previous fields that it also must search on?

Jeavon: Well, these are in addition to nodename but that does mean I've removed searching by id (I didn't know you could do that), so I'll put the id fields back!

Jeavon: Interestingly you can't search by key, I'll see if I can add that also!

Jeavon: Also now trying to figure out how to allow searching when there is a hyphen in the file name, e.g. "jan-copy.jpg" as it fails when the hyphen is reached....

src/Umbraco.Examine/MediaValueSetBuilder.cs
Jeavon: Add null check and Yield

Jeavon:
@Shazwazza trying to come up with a Lucene query that will allow for trailing wildcard but where symbols (hyphens, fullstops, undersores) are allowed, any ideas.....
For example, a file called "jancaps-copy.jpg" Then this query doesn't work `+(umbracoFileSrc:jancaps\-copy* )`

Shazwazza:
@Jeavon don't worry about searching on Key, i've added that in a different PR, see https://github.com/umbraco/Umbraco-CMS/pull/6530

As for your question regarding escaping chars and allowing wildcards, can you make anything work in Luke? Does the wildcard query work when you aren't escaping chars? Is it because the hyphens gets stripped out by the standard analyzer and for this field it should use whitespace analyzer?

Jeavon:
@Shazwazza I spent quite a while in Luke, the only thing I could make work was to replace the hyphens with spaces `+(umbracoFileSrc:jancaps copy.jpg* )` this then works perfectly. 
I believe the internal index is using the CultureInvariantWhitespaceAnalyzer so I don't think that's an issue...

Jeavon:
I've added a workaround to replace the hyphens to spaces when searching this field. It now works as expected and shown in this gif however maybe there is a better way that I have not found...

src/Umbraco.Tests/UmbracoExamine/IndexInitializer.cs
```diff
@@ -44,11 +45,21 @@ public static ContentIndexPopulator GetContentIndexRebuilder(PropertyEditorColle
...
+            var mediaValueSetBuilder = new MediaValueSetBuilder(propertyEditors, new UrlSegmentProviderCollection(new[] { new DefaultUrlSegmentProvider() }), GetMockUserService(), MockRuntimeState(RuntimeLevel.Run));
             var mediaIndexDataSource = new MediaIndexPopulator(null, mediaService, mediaValueSetBuilder);
             return mediaIndexDataSource;
         }
 
+        public static IRuntimeState MockRuntimeState(RuntimeLevel level)
+        {
+            var runtimeState = Mock.Of<IRuntimeState>();
+            Mock.Get(runtimeState).Setup(x => x.Level).Returns(level);
+            Mock.Get(runtimeState).SetupGet(m => m.ApplicationUrl).Returns(new Uri("https://LocalHost/umbraco"));
```
Shazwazza: nit pick. `LocalHost` should be all lowercase: `localhost`

src/Umbraco.Web/Search/UmbracoTreeSearcher.cs
```diff
@@ -85,6 +85,7 @@ public class UmbracoTreeSearcher
                     break;
                 case UmbracoEntityTypes.Media:
                     type = "media";
+                    fields = new[] { "id", "__NodeId", UmbracoExamineIndex.UmbracoFileFieldName};
```
Shazwazza: Should we just make the above `fields` definition a List, so that you can just add fields you want to it instead of re-creating it, just means less duplicated code (same goes for the member stuff above)

src/Umbraco.Examine/MediaValueSetBuilder.cs
```diff
@@ -29,6 +34,31 @@ public override IEnumerable<ValueSet> GetValueSets(params IMedia[] media)
...
+                    }
+                }
+                else
+                {
+                    umbracoFilePath = umbracoFileSource;
+                }
+
+                if (!string.IsNullOrEmpty(umbracoFilePath))
+                {
+                var uri = new Uri(_runtimeState.ApplicationUrl.GetLeftPart(UriPartial.Authority) + umbracoFilePath);
```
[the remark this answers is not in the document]
Jeavon: I could have a dummy URI but that felt somehow wrong (I could also split the string), but can do that, what would you suggest for a dummy URI (is there a precedent)?

src/Umbraco.Web/Search/UmbracoTreeSearcher.cs
Jeavon: Yes, would love to do that

Jeavon:
I like the ? idea, I'll try that also

Jeavon:
@Shazwazza 
I switched to using a List as suggested.  
I have left you a comment about the Url here https://github.com/umbraco/Umbraco-CMS/pull/6579/files#r336117246 
I tried the ? suggestion for the wildcard search but it didn't work (in Luke also), space seems to work perfectly so leaving as that....?

src/Umbraco.Examine/MediaValueSetBuilder.cs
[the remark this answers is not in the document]
Shazwazza: I think a dummy URL would be better and then we don't need to inject the IRuntimeState, at the end of the day it won't make any difference and we don't need to change as much

src/Umbraco.Web/Search/UmbracoTreeSearcher.cs
```diff
@@ -234,11 +235,26 @@ private bool BuildQuery(StringBuilder sb, string query, string searchFrom, strin
 
                     foreach (var f in fields)
                     {
+                        var queryWordsReplaced = new string[querywords.Length];
+
+                        // when searching file names containing hyphens we need to replace the hyphens with spaces
```
Shazwazza: Perhaps fine for now but I think ideally we'd have a custom analyzer for this field?

Jeavon: Agreed, I think this is fine for now, but introducing a method for custom analyzers per field is something to think about for the future

Shazwazza: Cool, we'll just see how this goes in the wild and do it properly later if necessary.

src/Umbraco.Examine/MediaValueSetBuilder.cs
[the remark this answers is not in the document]
Shazwazza: @Jeavon  this will be the last thing we need to cleanup i think and then we can merge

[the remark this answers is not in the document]
Jeavon: @Shazwazza updated, thanks!

src/Umbraco.Examine/MediaValueSetBuilder.cs
```diff
@@ -29,6 +32,32 @@ public override IEnumerable<ValueSet> GetValueSets(params IMedia[] media)
...
                 var urlValue = m.GetUrlSegment(_urlSegmentProviders);
+
+                var umbracoFilePath = string.Empty;
+                var umbracoFile = string.Empty;
+
+                var umbracoFileSource  = m.GetValue<string>(Constants.Conventions.Media.File);
+
+                if (umbracoFileSource.DetectIsJson())
+                {
+                    var cropper = JsonConvert.DeserializeObject<ImageCropperValue>(m.GetValue<string>(Constants.Conventions.Media.File));
```
Shazwazza: We'll need a try/catch here just in case since if this fails then all indexing fails

src/Umbraco.Web/Models/Mapping/EntityMapDefinition.cs
```diff
@@ -177,6 +177,15 @@ private static void Map(ISearchResult source, SearchResultEntity target, MapperC
 
             target.Name = source.Values.ContainsKey("nodeName") ? source.Values["nodeName"] : "[no name]";
 
+            if (source.Values.ContainsKey(UmbracoExamineIndex.UmbracoFileFieldName))
```
Shazwazza: This would be better and more efficient if changed to `if (source.Values.TryGetValue(UmbracoExamineIndex.UmbracoFileFieldName, out var umbracoFile))...`

src/Umbraco.Examine/MediaValueSetBuilder.cs
Jeavon: Done, to log the exception I needed to inject the logger which of course meant I had to mock the logger for the tests (please check)

src/Umbraco.Web/Models/Mapping/EntityMapDefinition.cs
Jeavon: Done

Shazwazza:
Awesome stuff @Jeavon all merged in 🎉

PR #4418, opened 2019-02-05
V8: Allow all content types at root if none are explicitly allowed

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes: https://github.com/umbraco/Umbraco-CMS/issues/4414

### Description

#4414 calls for allowing all content types at the root if none are explicitly marked as "Allow as root". This PR fixes that.

### Testing this PR
This PR affects both content and media and the test steps below should be repeated for both content and media. 

1. Make sure that nothing is broken 😄 in other words test that if any content types are marked as "Allow as root" you can't create, move or copy any other content types to root.
2. Disable "Allow as root" on all content types.
3. Verify that you can create all content types at root.
4. Verify that you can move and copy all content types to root - also from the recycle bin.

Issue #4414, opened 2019-02-04
v8: Allow creation of all types of content in root if unset


clausjensen:
If no content types have been defined with the 'Allow in root' setting - all content types should be allowed when creating root content.

It makes very little sense for new developers that they cannot create content of the content type they just created if they did not see and set the 'Allow in root' setting while creating it.

As soon as this setting has been used on at least one document type, it should however be enforced to only allow selected types as root content.

kjac:
I've got this. I'm assuming this should also be extended to moving/copying/restoring content.

Fun fact: This is already the case for media (although it doesn't work for move/copy).

kjac:
PR in #4418

clausjensen:
@kjac Thanks for the PR!! 😄 haha I somehow just _knew_ you would pick this up .. actually thought about pinging you, but figured you would snatch it up anyways 😃

kjac:
Hahaha! I'm all in it for the editor experience 😆

warrenbuckley:
All tested & fixed. Thanks @kjac

PR #6580, opened 2019-10-04
V8: It should be possible to disallow all types at content root 

kjac:
### Prerequisites

### Description

*This PR partially reverts #4418*

As it stands we can't disallow all types at root in content and media; if no types are markes as allowed at root, all types explicitly are made available at root. #4414 called for this feature to help newcomers, and however sympathetic that is, it's kinda bad not being able to lock down the editors at root level, specially for content.

Here's how it currently works:

disallow-at-root-before

With this PR, you explicitly need to allow content types at root before they're made available (for this PR I have opted to keep the behavior as-is for media):

disallow-at-root-after

This also applies to move and restore operations:

disallow-at-root-after-move

As for aiding newcomers, #6573 is specifically aimed at this and, and even though it's primarily centered around blank installs it will also work for this scenario.

#### Testing this PR

1. For content, make sure you can disallow every all content types at root level.
2. Once disallowed, make sure you can't move any content types to the root level.
3. For media, verify that you can still create any media type at root level if no media types are allowed at root level.


nul800sebastiaan:
In #4414 the scenario is: a developer new to Umbraco just created a doctype and it's not showing up. #6573 then tells them: hey, you need to create a document type here in order to add something.

So I would suggest that #6573 looks for existing document types and suggests updating the permissions. Maybe an extra button that takes you to the settings section?

kjac:
@nul800sebastiaan while this PR and #6573 are designed to compliment each other, I didn't want to be presumptuous by expecting both to be accepted 😄 

If they are indeed both accepted, I would suggest the following changes to #6573 

1. Update the wording a bit, so it hints at the option to edit the permissions of existing doctypes in instead of creating a new one.
2. Look into making "Allow at root" default true for the first doctype created, as this is likely to be the case anyway.

I wouldn't add an extra button though, the top menu does a top job at taking people to the settings section.

nul800sebastiaan:
@kjac In order to help with number 1 I've added a few more things to this PR: https://github.com/umbraco/Umbraco-CMS/pull/6580/commits/fca68e149c4ea2e3eb8b69c80fe22853b6a1a023

This looks like this when there are existing doctypes but none are allowed at root

image

What do you think?

I think this is good for now, I am not sure about number 2 (first doctype allowed at root by default) but that's a different issue.

kjac:
That's awesome @nul800sebastiaan 😀

PR #6582, opened 2019-10-04
V8: Split IndexCreator and Index config methods

bielu:
I notice IndexCreator actually has 2 responsibilities, which in my opinion is not the best way of doing that and I think it should be separate as in that case it looks weird: 
I want override index config I have to override whole umbracoindexcreator.

And here I have an additional case which could be useful:
I am preparing the package which creates elasticsearch indexes and second developer @callumbwhyte  have a package which wants override config in examine and actually we are hitting here wall if we want install both packages as both cases overriding umbracoindexcreator. 

In a way what I am proposing we could have both packages without stress about the ordering of components and handling differently situation depends on installed packages.

### DIFF
--- a/src/Umbraco.Examine/IUmbracoIndexConfig.cs
+++ b/src/Umbraco.Examine/IUmbracoIndexConfig.cs
@@ -1,0 +1,12 @@
+using Examine;
+
+namespace Umbraco.Examine
+{
+    public abstract class IUmbracoIndexConfig
+    {
+        public abstract IContentValueSetValidator GetContentValueSetValidator();
+        public abstract IContentValueSetValidator GetPublishedContentValueSetValidator();
+        public abstract IValueSetValidator GetMemberValueSetValidator();
+
+    }
+}
--- a/src/Umbraco.Examine/IUmbracoIndexConfig.cs
+++ b/src/Umbraco.Examine/IUmbracoIndexConfig.cs
@@ -2,11 +2,11 @@ using Examine;
 
 namespace Umbraco.Examine
 {
-    public abstract class IUmbracoIndexConfig
+    public interface IUmbracoIndexConfig
     {
-        public abstract IContentValueSetValidator GetContentValueSetValidator();
-        public abstract IContentValueSetValidator GetPublishedContentValueSetValidator();
-        public abstract IValueSetValidator GetMemberValueSetValidator();
+        IContentValueSetValidator GetContentValueSetValidator();
+        IContentValueSetValidator GetPublishedContentValueSetValidator();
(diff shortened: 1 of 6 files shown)


Shazwazza:
Yep that's great! I haven't had a chance to test but a peek at the code looks good.... But we can't have breaking changes so you can't remove those methods, your gonna have to obsolete them and wrap their response.

bielu:
@Shazwazza ok I change that to have them and actually, I didn't remove them and move them but ok made them Obsolete and wrap answer from IUmbracoIndexConfig as they are just in a different location and I think now is not anymore breaking change :)

src/Umbraco.Examine/IUmbracoIndexConfig.cs
```diff
@@ -0,0 +1,12 @@
+using Examine;
+
+namespace Umbraco.Examine
+{
+    public abstract class IUmbracoIndexConfig
```
Shazwazza: A class shouldn't not be prefixed with "I", that is reserved for interfaces. Seems like this should be an interface instead?

src/Umbraco.Examine/UmbracoUmbracoIndexConfig.cs
```diff
@@ -0,0 +1,34 @@
+using Examine;
+using Umbraco.Core.Services;
+using Umbraco.Core.Services.Implement;
+
+namespace Umbraco.Examine
+{
+    public class UmbracoUmbracoIndexConfig : IUmbracoIndexConfig
```
Shazwazza: This name has a double prefix of "Umbraco", it should only have one

src/Umbraco.Web/Search/UmbracoIndexesCreator.cs
```diff
@@ -80,28 +84,31 @@ private IIndex CreateMemberIndex()
                 CreateFileSystemLuceneDirectory(Constants.UmbracoIndexes.MembersIndexPath),
                 new CultureInvariantWhitespaceAnalyzer(),
                 ProfilingLogger,
-                GetMemberValueSetValidator());
+                GetMemberValueSetValidator()
+                );
             return index;
         }
-        
+        [Obsolete]
```
Shazwazza: We need to have a description of why this is obsoleted in the ctor overload of this attribute, same goes for the other obsoleted ones.

bielu:
@Shazwazza I changed code based on your comments :)

Shazwazza:
Looks good, thanks! will merge.

nul800sebastiaan:
Hi there @bielu,

Kind regards,

PR #6584, opened 2019-10-04
Copy to clipboard in query builder

crgrieve:
Hello,

I thought it would be cool (and hopefully useful) to have "copy to clipboard" functionality to the Query Builder in template editor.

See screenshot:
image

Thanks,
Carole


bjarnef:
@crgrieve it seems like a great Idea, but I think it would be useful to create e.g. a component like `umb-code-snippet` which include this button as part of the code snippet. This can be re-used in other part of the backoffice.

Maybe this button as part of the code area, e.g. in top right corner.
https://i.stack.imgur.com/NtG3c.gif

nul800sebastiaan:
Thanks @crgrieve, we'll have a look soon!

@bjarnef adding a highlighter is nice but seems like a bit of overkill for one small piece of code, we would need a broader use-case for loading even more javascript libraries into people's browsers,

bjarnef:
@nul800sebastiaan okay, but adding this as a component might be useful anyway, e.g. in the custom view after installing a package :)

nul800sebastiaan:
We usually try to work with the 80/20 rule: will adding something like this benefit 80% of the users of the CMS? In this case, I don't see this benefiting very many people so the answer is: not right now. :-)

nul800sebastiaan:
Thanks very much @crgrieve - nice work! 👍

bjarnef:
@nul800sebastiaan an alternative to highlight.js could be [Prism](https://prismjs.com), but yes depends on if we need it somewhere else in backoffice.

bjarnef:
@nul800sebastiaan a benefit with Prism is that TinyMCE support this to allow inserting code snippet into the richtext editor.
https://www.tiny.cloud/docs/plugins/codesample/

image

PR #6585, opened 2019-10-05
V8 style preview badge

skttl:
### Description
The current preview badge looks a bit dated.

firefox_2019-10-05_15-23-21

So I changed it to feel more like v8 in terms of look and colors, and also added a hover/focus style :)

firefox_2019-10-05_15-29-25
firefox_2019-10-05_15-29-50


nul800sebastiaan:
Thanks @skttl. It's nice to update this one, but it shouldn't have the Umbraco logo in it, it doesn't necessarily mean anything to people editing content and it's just "one more" thing agencies will need to replace when customizing/white-labeling the umbraco backoffice to be more specific to the client.

skttl:
Makes sense, I've updated the PR :)

nielslyngsoe:
Hi @skttl, very good point and looking good. :)

I do fear that the icon might be interpreted as "hide this badget". So if we use an Icon, we should make sure it provides a clear meaning. So we might need to look for something communicating "turn off", "terminate", "end this", ...

skttl:
A few alternatives
billede
billede

or a simple text link `click to end` as in the original

nielslyngsoe:
Yep, what about the "turn off" icon you have there in the end? Let's try with that for now.
image

skttl:
Updated :)

nul800sebastiaan:
Looks great @skttl .. at least: in your screenshot! ;-)

I am having trouble getting the icon to show up though, what am I missing? I've checked all the usual suspects, browser cache, etc.

image

skttl:
Hi @nul800sebastiaan 

Maybe a flexbug glitch? I added `flex-shrink:0` on the end-element, can you see if that helped you?

elit0451:
Hi @skttl 
I am having the same issue as @nul800sebastiaan. I am not able to see the icon either :woman_shrugging:

nul800sebastiaan:
@skttl you should be able to replicate by downloading the build (artifacts: https://umbraco.visualstudio.com/Umbraco%20Cms/_build/results?buildId=27916) of your latest commit, unzip it, run it in VS Code and IISExpress and check out the badge. Maybe something is different on your machine in the PR branch that is somehow not committed to git? 🤔

skttl:
Works on my machine 🤨

https://i.imgur.com/7Lg2Wiq.png

rasmusjp:
@skttl it works in Firefox 
image
but in Chromium (and I guess Chrome) and Safari the icon isn't showing 
image

Looks like the problem is that the svg is rendered as 0x0, so maybe specifying width and height will fix it 
image

skttl:
I added a width to the svg to make Chrome behave :)

elit0451:
Wohoo! :tada: Now it appears - tested in different browsers.
Things work as expected! Good job @skttl and @rasmusjp :star2:
Merging 👍

ReVoid:
@skttl Thanks for brand new look, looks great :) 

I have an another related idea by the way!
Sometimes user watches a page in preview mode, but without Umbraco sidebar. It means he can't just go to the backoffice screen to edit page, he forced to login into Umbraco and then search the page.

May be should add an another button as link to backoffice screen of the page? Right before close button. 

What do you think?

skttl:
Hi @ReVoid 

Yes that might be useful. I think you should create an issue with the feature request.

ReVoid:
@skttl Sure #7177 :)

Issue #4318, opened 2019-01-29
v8: Macro editor property picker


bjarnef:
In macro parameters the editor is a traditional dropdown. It would be great to change this to a property picker similar to when selecting data type on document types or selecting field types in Umbraco Forms.

![image](https://user-images.githubusercontent.com/2919859/51942910-7c9d2800-2418-11e9-90ec-848f0243d623.png)

nul800sebastiaan:
To me it seems like a good idea, but let me get some feedback first. I'll let you know after our next regular meeting.

nul800sebastiaan:
We had a chat about this recently and @nielslyngsoe indicated he would get back at you with some feedback! 👍

bjarnef:
@nul800sebastiaan I tried a quick prototype for this and keep the UI consistent as much as possible with selecting datatypes on document types.

The top part in the overlay is the new stuff (consistent with datatypes on document types) and the bottom part is the current stuff (should be removed if we go with this stuff).

![2019-10-02_22-50-22](https://user-images.githubusercontent.com/2919859/66080834-295f1480-e567-11e9-9a16-dbd807537281.gif)

nul800sebastiaan:
Looks great @bjarnef! I assume this will not show all editor, but only the ones that are available as a macro parameter editor?

nielslyngsoe:
@bjarnef looks good, looking forward to the PR, and as Sebastian is asking; I guess the options are the same as before, its just the UI you will be changing, right?

Looks very good to me!

bjarnef:
@nul800sebastiaan it only get the same parameter editors as listed in the current dropdown.
I have a branch for it here https://github.com/bjarnef/Umbraco-CMS/tree/dev-v8-temp-macro-parameter-editor

but I can create a PR for this in a few days.

The existing dropdown has a method `GetParameterEditors` and I have added a new method `GetGroupedParameterEditors`.
https://github.com/bjarnef/Umbraco-CMS/commit/c302d59067b0f42a22d383871f8f69f72998fd1a

Maybe these should return `IDictionary<string, IEnumerable<T>>` similar to `GetGroupedDataTypes` instead of `HttpResponseMessage`, but would can always change this later.

https://github.com/umbraco/Umbraco-CMS/blob/9c06ef83e217affbce51c1d6682f10914be2b400/src/Umbraco.Web/Editors/DataTypeController.cs#L311-L335

[...]

PR #6586, opened 2019-10-05
Pick macro parameters in an infinite editor like content type properties

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/4318

### Description
This PR adjust to macro parameters to be more consistent with manage properties and their datatypes on document types, media types and member types.

It also don't request parameter editors, before you want to edit or add a parameter.

At the moment the parameters doesn't have info about the "datatype name", so it make a request here. Not sure if it is easy to extend the `MacroParameterDisplay` with a `DataTypeName` or similar property here?
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web/Editors/MacrosController.cs#L119-L125

It would be great if the macro parameters also supported description as well, so you can add a description like on properties for document, media and member types.

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/resources/macro.resource.js
+++ b/src/Umbraco.Web.UI.Client/src/common/resources/macro.resource.js
@@ -112,6 +112,13 @@ function macroResource($q, $http, umbRequestHelper) {
             );
         },
 
+        getParameterEditorByAlias: function(alias) {
+            return umbRequestHelper.resourcePromise(
+                $http.get(umbRequestHelper.getApiUrl("macroApiBaseUrl", "GetParameterEditorByAlias", { "alias": alias }),
+                    "Failed to get parameter editor")
+            );
+        },
+
         getById: function(id) {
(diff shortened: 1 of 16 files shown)


nul800sebastiaan:
Thanks @bjarnef, we'll have a look! 

Adding descriptions and group names to the `MacroParameterDisplay` class will be a breaking change and requires some database updates as well. I am not sure if that is going to be a priority for us, will ask around.

bjarnef:
@nul800sebastiaan yes, the property editors also use a `DataTypeBasic` model used in list of datatypes (or selecting property editors to configurate). It doesn't seem we have something similar for macro parameter editors.

I think we can start we this and enhance it later. Furthermore I don't think we need to request macro partial views we loading a macro, but er can change this in another PR.

nul800sebastiaan:
This is lovely work @bjarnef! Many thanks, it all works wonderfully! ⭐️ 💪

PR #6587, opened 2019-10-05
Adds missing danish dictionary keys, and changes a few others

skttl:
I found some missing dictionary keys, and added them.

Also changed the one for "Last locked out date", as I think the wording could be better.


nul800sebastiaan:
Thanks @skttl !

abjerner:
Most of these looks good, but could we please not call an *exception* for a *undtagelse*?

Although it's the correct literal translation, it's IMO one of those technical terms that we shouldn't try to translate to Danish. Similar to how a *firewall* shouldn't be translated to a *brandmur*.

skttl:
I had my doubts on that one too, it's fine for me to keep it as exception.

PR #6589, opened 2019-10-06
V8: Reload node children after publishing with descendants

kjac:
### Prerequisites

### Description

When publishing includes unpublished descendants, the content tree doesn't reflect the changes in published status for the descendants:

reload-after-publish-before

This PR ensures that the children of the current node are reloaded in the content tree when publishing with unpublished descendants:

reload-after-publish-after


nul800sebastiaan:
Thanks @kjac! A testing note: check what happens if you Publish with descendants on a listview node.

nul800sebastiaan:
Love it! Thanks @kjac!

PR #6592, opened 2019-10-06
Correct URL decoding of macro partial view names

### LINKED ISSUE #6590: V8: Creating a Macro Partial View in the backoffice with spaces in the name can't be used with a Macro.
Using Umbraco 8.1 + 8.2RC

Create a Macro Partial View via the backoffice - right click 'Partial View Macro Files' and choose new partial view macro (without macro).

This is a common scenario, where you create the 'Macro' first, but then realise you need to now create the Partial View Macro File... (whereas if you'd created the Partial View Macro File first, you'd have the option to generate the Macro at the same time... possibly the UI for this journey could be tweaked, with the 'Macros' above the 'Partial View Macro Files' it's much more likely you will create the Macro first... anyway I digress.)

When you create the Partial View Macro File you are asked to add the 'Name' the convention for this is that it appears in the Umbraco Tree as this name... but it also sets the underlying filename of the partial view on disk...

... therefore it's quite likely you are thinking 'name' and not 'file name' and you create the name with spaces...

MacroNameWithSpaces

The Macro File Name + Macro File is created on disk with the spaces:

partialcreatedwithspaces

(should at this point the spaces have been stripped in the filename? eg it's ok for filenames to have spaces but it's only going to cause problems...)

Now switch back to the Macro... and pick the Macro File you just created... it appears in the list, but the spaces have been Url Encoded into + characters:

viewname gets plusses instead of spaces

But you can pick it and save the Macro ok... however when you try to insert the macro, the associated Partial View cannot be found... 

error occurred

... because it's looking for a file on disk with plusses in!!

Anyway so there are three different places where this 'could be fixed'... 

1) filename not having spaces
2) spaces not being url encoded as plusses when the partial is picked
or
3) turning the encoded plusses back into spaces when looking for the file...

So depends on if we feel we need to support having Macro Partial File Names with spaces?

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #6590

### Description
When picking a partial view in a macro's settings, the picker returns the filename URL-encoded. Any spaces are left encoded as +, and we store `~/Views/MacroPartials/Name+with+spaces.cshtml`. Then when trying to render the macro, the file can't be found.

Testing:
- Create a macro partial view with spaces in the name.
- Create a macro and pick that partial view.
- The filename should be displayed with spaces, not +.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/macros/views/macro.settings.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/macros/views/macro.settings.controller.js
@@ -35,7 +35,7 @@ function MacrosSettingsController($scope, editorService, localizationService) {
             },
             filterCssClass: "not-allowed",
             select: function (node) {
-                const id = unescape(node.id.replace(/\+/g, " "));
+                const id = decodeURIComponent(node.id.replace(/\+/g, " "));
 
                 //vm.macro.view = id;
                 $scope.model.macro.view = "~/Views/MacroPartials/" + id;
--- a/src/Umbraco.Web.UI.Client/src/views/macros/views/macro.settings.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/macros/views/macro.settings.controller.js
@@ -35,7 +35,7 @@ function MacrosSettingsController($scope, editorService, localizationService) {
             },
             filterCssClass: "not-allowed",
             select: function (node) {
-                const id = unescape(node.id);
+                const id = unescape(node.id.replace(/\+/g, " "));
 
                 //vm.macro.view = id;
                 $scope.model.macro.view = "~/Views/MacroPartials/" + id;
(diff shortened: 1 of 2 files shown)


bjarnef:
Maybe it would be better to replace this `unescape` function instead since it is deprecated.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape

stevemegson:
Yeah, we should be using `decodeURIComponent`. It doesn't help with decoding `+`, but it handles Unicode properly.

nul800sebastiaan:
Cool, looks good, we'll have a look soon, thanks!

bjarnef:
Maybe it is worth to wrap this in a function in one of the services?
https://stackoverflow.com/a/20247435

At the moment we are using this a.few places.

nul800sebastiaan:
Works like a charm!

PR #5599, opened 2019-06-06
V8: Copy all items from one NC to another

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes #5551

### Description

This is a PoC for copying all items from one Nested Content property to another. At the time of writing it works like this:

nc-copy-all-items

I would love some input on the feature, and also... could someone figure out how to place the "copy all items" button?

PR #6593, opened 2019-10-07
V8: Allow copying and pasting all Nested Content items in one go

### LINKED ISSUE #5551: V8: Copy content between languages while editing side-by-side in split view
Hi @nielslyngsoe,

I've got a PoC going for cloning content when working in split view, so the editors can quickly replicate the structure of one language in another:

split-view-duplicate-wip

Is this meaningful? Obviously there are a few things that need sorting out still, like confirmation dialog, icon placement and of course testing a whole lot.

As for icon placement, the placement in the gif works really badly when there's a scrollbar on either side of the split view. I played around with adding the clone button to the header instead, but it also looks a bit off:

image

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #5551

### Description

This is a redo of #5706, this time non-breaking (clientside only).

When applied you can copy and paste all items from one Nested Content property to another (or the same, if you really want) in one go. It works like this:

nc-copy-all

#### Testing this PR

1. Ensure that all items on a Nested Content property can be copied to another Nested Content property on content of the same type.
2. For language variant content where the Nested Content property is allowed to vary by language, ensure that all items can be copied between languages.
3. Verify that you can't copy items from one Nested Content to another if one or more of the copied item types aren't allowed on the target property configuration.
4. Ensure that single items can be copied between Nested Content properties (that this PR didn't break the single item copy feature, in other words).

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
@@ -108,6 +108,8 @@ angular.module("umbraco").controller("Umbraco.PropertyEditors.NestedContent.Prop
     "clipboardService",
     "eventsService",
     "overlayService",
+    "$routeParams",
+    "editorState",
     
     function ($scope, $interpolate, $filter, $timeout, contentResource, localizationService, iconHelper, clipboardService, eventsService, overlayService, $routeParams, editorState) {
         
@@ -145,6 +147,9 @@ angular.module("umbraco").controller("Umbraco.PropertyEditors.NestedContent.Prop
             $scope.labels.grid_addElement = data[0];
             $scope.labels.content_createEmpty = data[1];
         });
+        localizationService.localize("content_copyPropertyToClipboard", [$scope.model.label]).then(function (data) {
+            $scope.labels.copyProperty = data;
+        });
 
         // helper to force the current form into the dirty state
         $scope.setDirty = function () {
@@ -175,7 +180,14 @@ angular.module("umbraco").controller("Umbraco.PropertyEditors.NestedContent.Prop
                 view: "itempicker",
                 event: $event,
                 clickPasteItem: function(item) {
-                    $scope.pasteFromClipboard(item.data);
+                    if (item.alias === "nc_pasteAllItems") {
+                        _.each(item.data, function (node) {
+                            delete node.$$hashKey;
+                            $scope.pasteFromClipboard(node, false);
+                        });
+                    } else {
+                        $scope.pasteFromClipboard(item.data, true);
+                    }
                     $scope.overlayMenu.show = false;
                     $scope.overlayMenu = null;
                 },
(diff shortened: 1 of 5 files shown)


nul800sebastiaan:
Thanks @kjac! That sounds good without the breaking changes. One thing that's puzzling me is the UI here has changed to being a button under all the items instead of having a copy button next to the property name. If I recall correctly, this was an older proposal and you agreed with @nielslyngsoe to go with the button like it was implemented in the previous PR?

kjac:
@nul800sebastiaan aye, and that was the reason for the breaking change. The property type interface had to change in order to manipulate the label area, as that's not a part of the property clientside implementation.

ronaldbarendse:
As I commented earlier (https://github.com/umbraco/Umbraco-CMS/pull/5599#issuecomment-529805022), if we're going to add property actions, it should be made generic (and probably as a dropdown, so it doens't clutter the UI).

@kjac I've seen a mockup somewhere that had the 'Copy all content' as button after the 'Add content' that's a lot cleaner IMHO. Or the 'Add content' button could get a dropdown (like the publish button), so you can have more options and display better labels?

kjac:
I'm all for figuring out an approach that can be applied in a generic fashion to any property editor. I believe I have mentioned that earlier. I think HQ needs to get involved here.

This PR is mostly an attempt to get the ball rolling. Also I'd hate to see this stall because we can't figure out a generic approach... I'd rather have the feature fast and then refine it.

ronaldbarendse:
@nielslyngsoe I don't think any of those designs will work (at least for me 😉), because:
- Not all properties/editors might expose actions (no need to copy/paste a checkbox value, right?), so only some properties will get the dropdown (which might be visually confusing);
- The actions should be vertically aligned, so they're at a consistent place, but aligning them to the right looks weird, especially on smaller screens (where the label is full width);
- Showing the actions after the label might also be problematic for multi-line label texts;
- Showing the actions below the label pushes down the description (visually disconnecting the two), putting it below the description might not make it obvious it's related to that property;
- It's not compatible with the 'Hide label' option, which is a great option when using a single Nested Content editor within a group:
Nested Content with hidden label

nielslyngsoe:
Hi @ronaldbarendse thanks for sharing your thoughts on the matter.
As of what we know so far I do not think we will be respecting the "hide-label" option, especially since its a legacy feature. In that regards we do though want to enable people to use full width. Meaning in 'near' future the option to put the label on top would be a solution for those scenarios.

That said the problem is not becoming easier with that in mind.

If you have any idea on how you would like to solve it feel free to share. Rough sketchs are very welcome. Please keep in mind that we should avoid a custom solution for each data type. Would be nice if people can visually decode the interface very fast and know if and where the copy-feature is, based on experience from other properties.

Again thanks for your inputs :-)

ronaldbarendse:
IMHO you should really reconsider your view on not respecting the 'Hide label' option in the future, as there are a lot of contexts and editors that really don't require showing the label, e.g.:
- If it's the only property in the group and the label has the same name (as the group)
- If it has an insignificant/meaningless name (example above), also partially because the group name already contains the name
- Editors that show notes/inline help messages (that can't be put in a Content App, outside of the current view/context)
------------
Regarding the UI problem for property actions: I don't think a generic solution should be added near the labels. I've thought about adding an 'Actions' drop-down button on the group header (like 'Rollback' in the info tab), but that would need some way to group the actions per property (e.g. a long list with headers, a fly-out menu or something along the lines of the dialog list view).

Other (better) options might be to just add the actions to the right or make a toggle (like 'Reorder' on Document Types) to switch the context from entering content to taking actions on it?

But having the AngularJS services available for editors to tap into (as done in this PR) and creating an example implementation (on Nested Content and other core editors) might also be enough for this feature, at least for now...

@kjac I've made an example mock-up of how the 'Copy all' button could be added:

Add or copy content

Yes, there's no descriptive 'Copy all content of...' text, but that's currently also the case for copying single elements. This could be added as tooltip/`aria-describedby` though!

nielslyngsoe:
Thanks for inputs and concerns, we are aware about the importance of this feature and are working on a solution at HQ to ensure such a critical central feature will be implemented with focus on our values — a Friendly, Simple and Open solution, focusing on the bigger picture, as well making room for features of tomorrow.
You can follow the progress here: https://github.com/umbraco/Umbraco-CMS/issues/6816

PR #6599, opened 2019-10-07
Fixed and cleaned up custom exception classes

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/5947 (this is a rewrite of PR https://github.com/umbraco/Umbraco-CMS/pull/5950).

### Description
All custom `Exception` classes shoud be updated with:
- `[Serializable]` attribute (as any class that might be serialized must be marked with this attribute)
- Constructor with the signature `(SerializationInfo information, StreamingContext context)` (as required for all classes that implement `ISerializable`)
- If custom properties are added, restore the data in the above constructor and override `GetObjectData(SerializationInfo info, StreamingContext context)` to save the data

While going through all custom exceptions, a lot of them could be cleaned up (by throwing framework exceptions instead). I've kept the exception classes to keep backwards-compatibility, at least at a binary-level. This could be marked as a functional breaking-change, as implementing code could depend on catching a specific exception.

---
_This item has been added to our backlog [AB#3410](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3410)_


src/Umbraco.Core/Collections/ObservableDictionary.cs
```diff
@@ -235,15 +236,77 @@ public bool TryGetValue(TKey key, out TValue val)
...
 
+        /// <summary>
+        /// The exception that is thrown when a duplicate key inserted.
+        /// </summary>
+        /// <seealso cref="System.Collections.ObjectModel.ObservableCollection{TValue}" />
+        /// <seealso cref="System.Collections.Generic.IReadOnlyDictionary{TKey, TValue}" />
+        /// <seealso cref="System.Collections.Generic.IDictionary{TKey, TValue}" />
+        [Obsolete("Throw an ArgumentException when trying to add a duplicate key instead.")]
+        [Serializable]
         internal class DuplicateKeyException : Exception
```
ronaldbarendse: As this exception is internal and now unused, I think we could remove it completely...

src/Umbraco.Core/Exceptions/DataOperationException.cs
```diff
@@ -1,21 +1,98 @@
...
 namespace Umbraco.Core.Exceptions
 {
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <seealso cref="System.Exception" />
+    [Obsolete("Refactor the generic type to a concrete serializable type.")]
+    [Serializable]
     internal class DataOperationException<T> : Exception
```
ronaldbarendse: This class is internal and the only type used is `MoveOperationStatusType`, so I think this could be refactored without breaking changes.

src/Umbraco.Core/Persistence/Repositories/Implement/EntityContainerRepository.cs
```diff
@@ -162,9 +162,12 @@ protected override void PersistDeletedItem(EntityContainer entity)
 
         protected override void PersistNewItem(EntityContainer entity)
         {
+            // TODO Ensure correct exceptions are thrown (entity.Name is not an argument and NullReferenceException shouldn't be thrown by user code)
+            if (entity == null) throw new ArgumentNullException(nameof(entity));
             EnsureContainerType(entity);
 
-            if (string.IsNullOrWhiteSpace(entity.Name)) throw new ArgumentNullOrEmptyException("entity.Name");
+            if (entity.Name == null) throw new ArgumentNullException(nameof(entity.Name));
+            if (string.IsNullOrWhiteSpace(entity.Name)) throw new ArgumentException("Value can't be empty or consist only of white-space characters.", nameof(entity.Name));
```
ronaldbarendse: This is throwing an `ArgumentException` for a value that isn't an agument (same as in `PersistUpdatedItem`): should we throw an `InvalidOperationException` instead?

src/Umbraco.Core/Xml/XmlHelper.cs
```diff
@@ -229,8 +228,9 @@ public static XmlDocument OpenAsXmlDocument(string filePath)
         /// <returns>a XmlAttribute</returns>
         public static XmlAttribute AddAttribute(XmlDocument xd, string name, string value)
         {
-            if (xd == null) throw new ArgumentNullException("xd");
-            if (string.IsNullOrEmpty(name)) throw new ArgumentNullOrEmptyException(nameof(name));
+            if (xd == null) throw new ArgumentNullException(nameof(xd));
+            if (name == null) throw new ArgumentNullException(nameof(name));
+            if (string.IsNullOrEmpty(name)) throw new ArgumentException("Value can't be empty.", nameof(name));
```
ronaldbarendse: `SetAttribute` uses `string.IsNullOrWhiteSpace()`: should we change this to keep it the same?

src/Umbraco.Web/GridTemplateExtensions.cs
```diff
@@ -26,18 +23,20 @@ public static MvcHtmlString GetGridHtml(this HtmlHelper html, IPublishedContent
...
         public static MvcHtmlString GetGridHtml(this HtmlHelper html, IPublishedContent contentItem, string propertyAlias, string framework)
         {
-            if (string.IsNullOrWhiteSpace(propertyAlias)) throw new ArgumentNullOrEmptyException(nameof(propertyAlias));
+            if (propertyAlias == null) throw new ArgumentNullException(nameof(propertyAlias));
+            if (string.IsNullOrWhiteSpace(propertyAlias)) throw new ArgumentException("Value can't be empty or consist only of white-space characters.", nameof(propertyAlias));
 
             var view = "Grid/" + framework;
             var prop = contentItem.GetProperty(propertyAlias);
-            if (prop == null) throw new NullReferenceException("No property type found with alias " + propertyAlias);
+            if (prop == null) throw new NullReferenceException("No property type found with alias " + propertyAlias); // TODO NullReferenceException should not be thrown by user code
```
ronaldbarendse: We should probably throw an `InvalidOperationException` instead.

src/Umbraco.Web/GridTemplateExtensions.cs
```diff
@@ -60,30 +59,25 @@ public static MvcHtmlString GetGridHtml(this IPublishedContent contentItem, Html
...
         public static MvcHtmlString GetGridHtml(this IPublishedContent contentItem, HtmlHelper html, string propertyAlias, string framework)
         {
-            if (string.IsNullOrWhiteSpace(propertyAlias)) throw new ArgumentNullOrEmptyException(nameof(propertyAlias));
+            if (propertyAlias == null) throw new ArgumentNullException(nameof(propertyAlias));
+            if (string.IsNullOrWhiteSpace(propertyAlias)) throw new ArgumentException("Value can't be empty or consist only of white-space characters.", nameof(propertyAlias));
 
             var view = "Grid/" + framework;
             var prop = contentItem.GetProperty(propertyAlias);
-            if (prop == null) throw new NullReferenceException("No property type found with alias " + propertyAlias);
+            if (prop == null) throw new NullReferenceException("No property type found with alias " + propertyAlias); // TODO NullReferenceException should not be thrown by user code
```
ronaldbarendse: Same here: we should probably throw an `InvalidOperationException` instead.

src/Umbraco.Web/HtmlHelperRenderExtensions.cs
```diff
@@ -243,9 +240,9 @@ protected override void Dispose(bool disposing)
...
                     || _controllerName == "UmbLoginStatus"
                     || _controllerName == "UmbLogin")
-			    {
+                {
                     _viewContext.Writer.Write(AntiForgery.GetHtml().ToString());
-			    }
+                }
 
                 //write out the hidden surface form routes
                 _viewContext.Writer.Write("<input name='ufprt' type='hidden' value='" + _encryptedString + "' />");
```
ronaldbarendse: Just a side note: the single quotes should probably be changed to double quotes...

src/Umbraco.Web/HtmlHelperRenderExtensions.cs
```diff
@@ -355,8 +352,10 @@ public static MvcForm BeginUmbracoForm(this HtmlHelper html, string action, stri
                                                IDictionary<string, object> htmlAttributes,
                                                FormMethod method)
         {
-            if (string.IsNullOrWhiteSpace(action)) throw new ArgumentNullOrEmptyException(nameof(action));
-            if (string.IsNullOrWhiteSpace(controllerName)) throw new ArgumentNullOrEmptyException(nameof(controllerName));
+            if (action == null) throw new ArgumentNullException(nameof(action));
+            if (string.IsNullOrWhiteSpace(action)) throw new ArgumentException("Value can't be empty or consist only of white-space characters.", nameof(action));
```
ronaldbarendse: Should `action` be renamed to `actionName` to keep it the same as in other methods and `controllerName`?

src/Umbraco.Web/HtmlHelperRenderExtensions.cs
```diff
@@ -355,8 +352,10 @@ public static MvcForm BeginUmbracoForm(this HtmlHelper html, string action, stri
...
+            if (controllerName == null) throw new ArgumentNullException(nameof(controllerName));
+            if (string.IsNullOrEmpty(controllerName)) throw new ArgumentException("Value can't be empty.", nameof(controllerName));
```
ronaldbarendse: This should be changed to `string.IsNullOrWhiteSpace()` 😇 

```suggestion
            if (string.IsNullOrWhiteSpace(controllerName)) throw new ArgumentException("Value can't be empty or consist only of white-space characters.", nameof(controllerName));
```

src/Umbraco.Web/HtmlHelperRenderExtensions.cs
```diff
@@ -687,8 +690,10 @@ public static MvcForm BeginUmbracoForm(this HtmlHelper html, string action, stri
                                                IDictionary<string, object> htmlAttributes,
                                                FormMethod method)
         {
-            if (string.IsNullOrEmpty(action)) throw new ArgumentNullOrEmptyException(nameof(action));
-            if (string.IsNullOrEmpty(controllerName)) throw new ArgumentNullOrEmptyException(nameof(controllerName));
+            if (action == null) throw new ArgumentNullException(nameof(action));
+            if (string.IsNullOrEmpty(action)) throw new ArgumentException("Value can't be empty.", nameof(action));
+            if (controllerName == null) throw new ArgumentNullException(nameof(controllerName));
+            if (string.IsNullOrEmpty(controllerName)) throw new ArgumentException("Value can't be empty.", nameof(controllerName));
```
ronaldbarendse: Should this be changed to `string.IsNullOrWhiteSpace()` to keep it the same as in other methods?

src/Umbraco.Web/Search/SearchableTreeAttribute.cs
```diff
@@ -8,42 +7,47 @@ public sealed class SearchableTreeAttribute : Attribute
...
+        public SearchableTreeAttribute(string serviceName, string methodName, int sortOrder)
         {
-            if (string.IsNullOrWhiteSpace(serviceName)) throw new ArgumentNullOrEmptyException(nameof(serviceName));
-            MethodName = "";
+            if (serviceName == null) throw new ArgumentNullException(nameof(serviceName));
+            if (string.IsNullOrWhiteSpace(serviceName)) throw new ArgumentException("Value can't be empty or consist only of white-space characters.", nameof(serviceName));
+
             ServiceName = serviceName;
-            SortOrder = DefaultSortOrder;
+            MethodName = methodName ?? throw new ArgumentNullException(nameof(methodName));
```
ronaldbarendse: We should allow an empty `methodName`, as the constructor without this argument sets it to an empty string anyway.

src/Umbraco.Web/Security/AuthenticationOptionsExtensions.cs
```diff
@@ -75,7 +75,9 @@ public static ExternalSignInAutoLinkOptions GetExternalAuthenticationOptions(thi
         /// </param>
         public static void ForUmbracoBackOffice(this AuthenticationOptions options, string style, string icon, string callbackPath = null)
         {
-            if (string.IsNullOrEmpty(options.AuthenticationType)) throw new ArgumentNullOrEmptyException("options.AuthenticationType");
+            // TODO Change exceptions to InvalidOperationException, as the value isn't an argument
+            if (options.AuthenticationType == null) throw new ArgumentNullException(nameof(options.AuthenticationType));
+            if (string.IsNullOrEmpty(options.AuthenticationType)) throw new ArgumentException("Value can't be empty.", nameof(options.AuthenticationType));
```
ronaldbarendse: Should this also be changed to `InvalidOperationException` and should we null-check `options` first?

nul800sebastiaan:
Thanks, we'll give this another look soon 👍

src/Umbraco.Core/Exceptions/DataOperationException.cs
Shazwazza: It could be but until then we shouldn't be marking this as obsolete since it's still in use.

src/Umbraco.Core/Exceptions/WontImplementException.cs
```diff
@@ -1,27 +1,52 @@
...
+    /// <remarks>
+    /// The <see cref="NotImplementedException" /> is to be used when some code is not implemented,
     /// but should eventually be implemented (i.e. work in progress) and is reported by tools such as ReSharper.
     /// This exception is to be used when some code is not implemented, and is not meant to be, for whatever
-    /// reason.</remarks>
+    /// reason.
+    /// </remarks>
+    /// <seealso cref="System.NotImplementedException" />
+    [Serializable]
+    [Obsolete("If a method or operation is not, and will not be, implemented, it is invalid, so we should throw an InvalidOperationException instead.")]
```
Shazwazza: I'm unsure about this one - if it's not implemented and doesn't intend to be, shouldn't it throw a `NotSupportedException`?

src/Umbraco.Core/Models/ContentRepositoryExtensions.cs
```diff
@@ -103,11 +102,11 @@ public static void CopyFrom(this IContent content, IContent other, string cultur
 
         public static void SetPublishInfo(this IContent content, string culture, string name, DateTime date)
         {
-            if (string.IsNullOrWhiteSpace(name))
-                throw new ArgumentNullOrEmptyException(nameof(name));
+            if (name == null) throw new ArgumentNullException(nameof(name));
```
Shazwazza: These null and then null/whitespace checks I think are unecessary. Both resharper and VS when using the tools will just do a single check and throw an `ArgumentException` which i think is fine. It's ok to leave these in here now that they are here but since the vast majority of us will be using VS or resharper tools most times checking for null/empty parameters will just be throwing a single ArgumentException.

src/Umbraco.Core/Persistence/Repositories/Implement/EntityContainerRepository.cs
Shazwazza: I guess the "argument" (:P) is that it's a property of an argument/parameter, but yeah i'm fine with InvalidOperationException

src/Umbraco.Core/Xml/XmlHelper.cs
Shazwazza: Yep IsNullOrWhiteSpace is better

src/Umbraco.Web/GridTemplateExtensions.cs
Shazwazza: Yep I agree

src/Umbraco.Web/GridTemplateExtensions.cs
Shazwazza: indeed

src/Umbraco.Web/HtmlHelperRenderExtensions.cs
Shazwazza: Happy for that change if you have time

src/Umbraco.Web/HtmlHelperRenderExtensions.cs
Shazwazza: Don't think we should change the public param name at this stage, it's a small thing but don't want to introduce some odd breaking change if someone is relying on that name (at least for now and this PR)

src/Umbraco.Web/HtmlHelperRenderExtensions.cs
Shazwazza: indeed

src/Umbraco.Web/HtmlHelperRenderExtensions.cs
Shazwazza: yep

src/Umbraco.Web/Install/InstallException.cs
```diff
@@ -1,44 +1,109 @@
...
+using System.Runtime.Serialization;
 
 namespace Umbraco.Web.Install
 {
     /// <summary>
-    /// Used for steps to be able to return a json structure back to the UI
+    /// Used for steps to be able to return a JSON structure back to the UI.
     /// </summary>
+    /// <seealso cref="System.Exception" />
+    [Obsolete("Refactor the view model into a concrete serializable type.")]
```
Shazwazza: This shouldn't be obsoleted since it's still in use. I understand that it might not conform to an ideal exception type but until we refactor to something different this can't be obsoleted. We can refactor this later (baby steps... would rather get this larger PR merged in sooner than later)

src/Umbraco.Web/Search/SearchableTreeAttribute.cs
Shazwazza: but when using this overload, shouldn't we assume it's being used because the methodName should be supplied?

src/Umbraco.Web/Security/AuthenticationOptionsExtensions.cs
Shazwazza: yep

src/Umbraco.Web/HtmlHelperRenderExtensions.cs
ronaldbarendse: Yes, it's minor, but indeed still a breaking change... Will leave this alone for now 👍

src/Umbraco.Core/Exceptions/WontImplementException.cs
ronaldbarendse: Both `NotSupportedException` and `InvalidOperationException` are a lot better compared to the custom exception based on `NotImplementedException` and deciding which one to throw probably depends on the context, see:
https://docs.microsoft.com/en-us/dotnet/api/system.notsupportedexception and https://docs.microsoft.com/en-us/dotnet/api/system.invalidoperationexception.

src/Umbraco.Web/Install/InstallException.cs
ronaldbarendse: I've removed the serialization of the (possibly complex and unsupported) view model, so it at least keeps the view name (so it's better then before) 👍

src/Umbraco.Core/Exceptions/DataOperationException.cs
ronaldbarendse: I've kept the generics, but added a generic type constraint, so `T` must always be an enum 👍

src/Umbraco.Web/Search/SearchableTreeAttribute.cs
ronaldbarendse: Hmm, because it's an attribute, the arguments are quite static (besides dynamically emitting/adding attributes 🤡), so the overload shouldn't be used with an empty method name indeed.

But why add the additional check if the value is valid anyway and exceptions thrown in attributes should be avoided in general (as it gets thrown when getting the attribute)?

src/Umbraco.Core/Models/ContentRepositoryExtensions.cs
ronaldbarendse: Only ReSharper does that: VS correctly suggests both `ArgumentNullException` and `ArgumentException` 😉

ronaldbarendse: As per comment here: https://github.com/umbraco/Umbraco-CMS/pull/5950#discussion_r317768334

ronaldbarendse:
I've processed your feedback and the only thing that should be done is making all custom exceptions public: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1064.

Shazwazza:
Gonna merge this beast! Thanks for the mammoth amount of work here 🎉we can keep making things better in baby steps, i don't want to make this PR any bigger or postpone any longer.

Shazwazza:
... but before I merge, the build service says it cannot build this https://umbraco.visualstudio.com/Umbraco%20Cms/_build/results?buildId=29626

Shazwazza:
@ronaldbarendse the errors are:

```
##[error]src\Umbraco.Core\Services\Implement\ContentService.cs(1252,39): Error CS0246: The type or namespace name 'PanicException' could not be found (are you missing a using directive or an assembly reference?)
Services\Implement\ContentService.cs(1252,39): error CS0246: The type or namespace name 'PanicException' could not be found (are you missing a using directive or an assembly reference?) [d:\a\1\s\src\Umbraco.Core\Umbraco.Core.csproj]
##[error]src\Umbraco.Core\Services\Implement\ContentService.cs(1337,27): Error CS0246: The type or namespace name 'PanicException' could not be found (are you missing a using directive or an assembly reference?)
Services\Implement\ContentService.cs(1337,27): error CS0246: The type or namespace name 'PanicException' could not be found (are you missing a using directive or an assembly reference?) [d:\a\1\s\src\Umbraco.Core\Umbraco.Core.csproj]
##[error]src\Umbraco.Core\Models\PublishedContent\ModelType.cs(79,27): Error CS0246: The type or namespace name 'PanicException' could not be found (are you missing a using directive or an assembly reference?)
Models\PublishedContent\ModelType.cs(79,27): error CS0246: The type or namespace name 'PanicException' could not be found (are you missing a using directive or an assembly reference?) [d:\a\1\s\src\Umbraco.Core\Umbraco.Core.csproj]
##[error]src\Umbraco.Core\Models\PublishedContent\ModelType.cs(118,27): Error CS0246: The type or namespace name 'PanicException' could not be found (are you missing a using directive or an assembly reference?)
Models\PublishedContent\ModelType.cs(118,27): error CS0246: The type or namespace name 'PanicException' could not be found (are you missing a using directive or an assembly reference?) [d:\a\1\s\src\Umbraco.Core\Umbraco.Core.csproj]
```

does it build on your machine? Maybe you can merge the latest v8/dev back into your branch and push

ronaldbarendse:
This should fix the build issue, as the `PanicException` was added after creating this PR 👍

The only minor thing is making the exceptions public, so they can be catched within user code...

ronaldbarendse:
I've checked the internal exceptions and the following are exposed publicly, so to make them catch-able in user code, I've made them public:
- `PanicException`
- `InstallException`
- `NotValidExifFileException`
- `NotValidJPEGFileException`
- `NotValidTIFFileException`
- `SectionExceeds64KBException`

ronaldbarendse:
The following 2 internal exceptions are still exposed publicly, but I think they should be removed instead of making them public:
- `ConnectionException`, used once to wrap another exception:
  https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Core/Services/Implement/PackagingService.cs#L58-L61
- `UnknownEnumTypeException`, used once in a property getter (this should be avoided according to the [Microsoft Framework Design Guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property)):
  https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web/Media/Exif/ExifExtendedProperty.cs#L67

I think this should be it for this PR, as cleaning up more internal exceptions might not be so straightforward. Internal types are exposed to quite a lot of assemblies using the `InternalsVisibleTo` attribute, so they might still be used. I've put my further findings below though 😉 

-----------

The following are never thrown and could be cleaned up:
- `DataLossException`
- `NotValidImageFileException`
- `CachedTypeNotFoundInFileException`, although needlessly checked once:
   https://github.com/umbraco/Umbraco-CMS/blob/9f817bf032fd5799b1c1dbec6093f88deaea493d/src/Umbraco.Core/Composing/TypeLoader.cs#L713

The internal `DataOperationException<T>` exception is always catched, although some methods could be rewritten to avoid throwing/catching the exception:
- https://github.com/umbraco/Umbraco-CMS/blob/ed6fee4fa4a1e29181fcb7f0cefccdf378af91e3/src/Umbraco.Core/Services/Implement/ContentTypeServiceBaseOfTRepositoryTItemTService.cs#L650
- https://github.com/umbraco/Umbraco-CMS/blob/ed6fee4fa4a1e29181fcb7f0cefccdf378af91e3/src/Umbraco.Core/Services/Implement/ContentTypeServiceBaseOfTRepositoryTItemTService.cs#L712
- https://github.com/umbraco/Umbraco-CMS/blob/ed6fee4fa4a1e29181fcb7f0cefccdf378af91e3/src/Umbraco.Core/Services/Implement/DataTypeService.cs#L310

Shazwazza:
Lets avoid removing more exceptions, i'll need to look through other projects to make sure they are not used. Feel free to obsolete the unused exceptions. I would rather get this merged in than start refactoring methods, etc... more cleanup can be done in smaller chunks.

Shazwazza:
@ronaldbarendse i'm just re-reviewing and will merge in and then make some adjustments for what you've written above

Shazwazza:
* ConnectionException = removed
* DataLossException = removed (used to be used for mysql)
* NotValidImageFileException (part of exif parsing) = removed
* UnknownEnumTypeException (part of exif parsing) = removed

see rev f42715548089e9fe294ba3c2ea6525b22fbd21db

Leaving 

* CachedTypeNotFoundInFileException - since i weary about removing this right now since i don't remember it's history and don't have time to investigate at the moment
* DataOperationException<T> - since i don't have time for refactoring at the moment

 ... happy to review a PR for these ones if you think it's important.

Thanks!!

ronaldbarendse:
Yes, it's merged! 🎉

The following can be done in separate PRs, indeed:
- Remove or change to public exception: `ConnectionException` and `UnknownEnumTypeException`
- Investigate usage of `CachedTypeNotFoundInFileException` and remove if unused (e.g. not thrown)
- Refactor usage of `DataOperationException<T>`, since it's sometimes thrown and catched in the same method

PR #6600, opened 2019-10-08
V8: Add irreversible warning when deleting items from the recycle bin

kjac:
### Prerequisites

### Description

When deleting content from the recycle bin, the warning message is the same as when you move content to the recycle bin - and it makes no sense in the context of deleting content from the recycle bin:

image

With this PR, the warning makes a little more sense when deleting content:

image

...and the same applies to media:

image

And as an added bonus, the confirmation button is now red when moving media to the recycle bin (it is currently green):

image


nul800sebastiaan:
Such warning! 😄  We'll get back to you Kenn, cheers!

nul800sebastiaan:
Sweet, thanks!

PR #6512, opened 2019-10-01
Boolean editors - adds ID attribute to umbToggle

leekelleher:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description

The Boolean editor (aka. TrueFalse, aka. Checkbox 🤔) using the `<umb-toggle>` directive (for both prevalue-editor and property-editor), isn't associated with the property's `<label>`.

ECBnKKPXsAExgUM

Meaning that if a user clicks on the property label, nothing happens.  Previous with the old checkbox (e.g. native HTML input), this would check/uncheck the input.

By adding an `id` attribute, with the property's alias (e.g. `id="{{model.alias}}"`), it will be associated with the toggle control.

In terms of testing, it'd be a case of clicking the label, and see the toggle state change.

PR #6608, opened 2019-10-08
Set focus to Umb-Tags when clicking property label

### LINKED ISSUE #6605: Set focus to Umb-Tags when clicking property label
Clicking the property-label for a umb-tags property should set focus to the "write new tag"-input.

image

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #6605

### Description
Allows the Tags property editor to pass the property alias into umb-tags-editor, where it can be used as the ID of the text box to match the label's `for` attribute.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/tags/umbtagseditor.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/tags/umbtagseditor.directive.js
@@ -18,7 +18,7 @@
                 config: "<",
                 validation: "<",
                 culture: "<?",
-                htmlId: "<?",
+                inputId: "@?",
                 onValueChanged: "&"
             }
         });
@@ -50,7 +50,7 @@
         vm.viewModel = [];
 
         function onInit() {
-            vm.htmlId = vm.htmlId || "t" + String.CreateGuid();
+            vm.inputId = vm.inputId || "t" + String.CreateGuid();
 
             assetsService.loadJs("lib/typeahead.js/typeahead.bundle.min.js").then(function () {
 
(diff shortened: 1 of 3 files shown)


bjarnef:
It would be great to name the property `input-id` (inputId) to make it consistent with setting `id` attribute in e.g. `umb-checkbox` and `umb-radiobutton`.. and also `umb-toggle`.

https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/components/tags/umb-tags-editor.html#L24

nul800sebastiaan:
Thanks @stevemegson - I just want to make sure you've seen the fix in https://github.com/umbraco/Umbraco-CMS/pull/6512/files - it deviated a little from this one. 

Specifically, I can never remember what the difference between `htmlId: "<?",` and `htmlId: "@",` would be?

bjarnef:
@nul800sebastiaan `@` means it is an one-way binding, but converted to a string value.  `<` is one-way binding, but the type is preserved, `=` is two-way binding.
`?` means it is optional :)

stevemegson:
Both excellent points, I hadn't seen that PR. I'll update this one to match.

nul800sebastiaan:
Hi @stevemegson - sorry for the delay! Unfortunately, this seems to have generated an unfortunate side-effect, now when you click on the tags label and try to navigate away to a different node, the current node has been marked as dirty and you get a "discard changes" notification. Any chance you can look into that one please?

nul800sebastiaan:
I've just re-tested this and the dirty marker doesn't get set this time, might have been some weird caching issue.. ! Thanks very much @stevemegson! 👍

PR #6609, opened 2019-10-08
Set focus to RTE when clicking property label 

### LINKED ISSUE #6602: Set focus to RTE when clicking property label
Clicking the label of a Rich Text Editor Property, should set focus to the edit-area of the RTE, like other properties.

image

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #6602

### Description

It feels like there should be a better solution for this, but this seems to work. 

Adds an invisible `input` before the main div to give something for the label to set focus on:
```html
<input type="text" id="{{model.alias}}" ng-focus="focus()" style="position:absolute;top:0;width:0;height:0;" />
```

In the controller, sets focus to the RTE whenever that `input` receives focus:
```javascript
$scope.focus = function () {
    tinyMceEditor.focus();
}
```

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
@@ -676,7 +676,7 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
 
                 //get all macro divs and load their content
                 $(editor.dom.select(".umb-macro-holder.mceNonEditable")).each(function () {
-                    self.loadMacroContent($(this), null);
+                    self.loadMacroContent($(this), null, editor);
                 });
 
             });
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
This sure is an "interesting" approach! I'll have to send it to my colleagues and see if it makes sense. We'll get back to you!

stevemegson:
You seem to have misspelled "horrifying".

nul800sebastiaan:
Hmm, maybe it's because I'm merging v8/dev in first.

stevemegson:
I can reproduce that now, it looks like it happens when an RTE contains rendered macros. When we load the macro result and update the editor, TinyMCE sees that as a change that marks the editor dirty. But it doesn't check for changes until the editor has received focus, whether by clicking on the label or clicking directly in the editor.

Hopefully the new commit fixes this. If the editor wasn't already marked as dirty before we load the macro result, it resets TinyMCE's change tracking after inserting the result.

nul800sebastiaan:
Works wonderfully now, thanks very much! 👍

Issue #6591, opened 2019-10-06
Sort hostnames in Culture and hostnames dialog


skttl:
https://issues.umbraco.org/issue/U4-11558 - #3210
Last year I did a PR for this in v7 (#3295) and I would like to have this feature in v8. But it would be a breaking change, so before doing the work I would like to hear HQs opinion on this :)

nul800sebastiaan:
I would say if you can do it without making it a breaking change then that would be great, You could remove the existing domain names and re-insert them in the correct order. Make sure to do the whole delete+insert in a transaction.

I do not know what the impact will be for Deploy though, but I'm guessing people do this manually now anyway (deleting them and then re-adding in the correct order) so it should hopefully not be a problem.

skttl:
Sounds great, I'll see what I can do. My v7 PR had breaking changes though :)

skttl:
By trying to fix this, I noticed that the urls are ordered before being returned to the ContentController here: https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Web/Routing/DomainUtilities.cs#L282

Can that ordering be removed and have the list ordered by its original order?

zpqrtbnk:
Quick note, ordering of the urls is important *in some places* so that if you define both "example.com/en" and "example.com" as hostnames, we try the longer one first when matching the current request - not sure it's relevant here, but just be careful.

skttl:
I don't think that would be a problem, as the sorting is used to generate the url. All the hostnames should be able to match anyway.

nul800sebastiaan:
As noted in PR https://github.com/umbraco/Umbraco-CMS/pull/6611#issuecomment-573586582 - we'll put this one on the list of ideas to implement in the future, we unfortunately can't do it now.

PR #6611, opened 2019-10-08
Sort hostnames (help needed)

skttl:
#6591

This is what I got so far.

I compare the order of hostnames between the submitted list, and the persisted list, and deletes the existing domains, if the order doesn't match.

As mentioned in https://github.com/umbraco/Umbraco-CMS/issues/6591#issuecomment-539188057 there is some ordering of the domains overriding the database order. I tried removing that, but that didn't help on the ordering in the links group on the info tab. I'll keep digging.

I'm not sure how to wrap the deletion/re-adding of domains in a transaction, it just uses the domainservice for updating, like before. Any pointers is welcome :)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/assigndomain.html
+++ b/src/Umbraco.Web.UI.Client/src/views/content/assigndomain.html
@@ -28,9 +28,10 @@
                     <localize key="assignDomain_domainHelpWithVariants"></localize>
                 </small>
                 <div class="umb-el-wrap  hidelabel">
-                    <table class="table table-condensed table-bordered domains" style="margin-bottom: 10px;" ng-if="vm.domains.length > 0">
+                    <table class="table table-condensed domains" style="margin-bottom: 10px;" ng-if="vm.domains.length > 0">
                         <thead>
                             <tr>
+                                <th></th>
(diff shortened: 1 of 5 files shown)


nul800sebastiaan:
Aha, I am sure the Service will take care of all that transaction stuff for you, but we'll give it a look as soon as this is out of draft! 👍

skttl:
I added the UI, and removed the ordering from various places. But they still don't follow the db order when going through `UrlProviderExtensions.GetContentUrls` (https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web/Routing/UrlProviderExtensions.cs#L11)

I stepped through the method, and the yields come in the right order, but somehow the end result is out of order. I need some help from this point...

nul800sebastiaan:
Hi @skttl - sorry for the delay on this one, it's been on the todo list for a while.

Unfortunately, we won't be able to consider a PR for this feature if it's not possible to do it without a breaking change. The feedback I got from the team is that you won't be able to reliably get the order from the DB. 

So unfortunately, I will have to close the PR and will put the issue on the ideas list for future consideration when we are allowed to make breaking changes.

PR #6612, opened 2019-10-08
Use umb-radiobutton in public access overlay

bjarnef:
### Prerequisites

### Description
This PR replace the traditional radiobuttons in the public access overlay using `umb-radiobutton` component. I have also adjusted the layout a bit and the `umbRadiobutton` component has the `inputId` property similar to `umbCheckbox`.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
@@ -22,6 +22,7 @@
 </pre>
 
 @param {boolean} model Set to <code>true</code> or <code>false</code> to set the radiobutton to checked or unchecked.
+@param {string} inputId Set the <code>id</code> of the radiobutton.
 @param {string} value Set the value of the radiobutton.
 @param {string} name Set the name of the radiobutton.
 @param {string} text Set the text for the radiobutton label.
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
@@ -55,6 +55,7 @@
         controllerAs: 'vm',
         bindings: {
             model: "=",
+            inputId: "@",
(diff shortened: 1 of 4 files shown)


nielslyngsoe:
@bjarnef  I love it, Thanks!

bjarnef:
@nielslyngsoe it would also be great to move this styling out of "hacks".
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/less/hacks.less#L91-L178

Not sure if we should put specific "overlay" styling in a specific folder?

nielslyngsoe:
> @nielslyngsoe it would also be great to move this styling out of "hacks".
> [/dev/src/Umbraco.Web.UI.Client/src/less/hacks.less@`v8`#L91-L178](https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/less/hacks.less#L91-L178) [](/umbraco/Umbraco-CMS/blob/HEAD@{2019-10-09T13:29:19Z}/dev/src/Umbraco.Web.UI.Client/src/less/hacks.less#L91-L178)
> 
> Not sure if we should put specific "overlay" styling in a specific folder?

Yeah, defiantly. I would say the purpose of "hacks.less" is to contain CSS-Framework overwrites, so the Public Access overlay stuff should not belong there.

So I'm all in for moving it, I would suggest it should go into its own new file.

bjarnef:
I also noticed these classes, but can't see they are being used?
https://github.com/umbraco/Umbraco-CMS/tree/v8/dev/src/Umbraco.Web.UI.Client/src/less/hacks.less#L11-L13

We already have many utilty classes here:
https://github.com/umbraco/Umbraco-CMS/tree/v8/dev/src/Umbraco.Web.UI.Client/src/less/utilities

nielslyngsoe:
@bjarnef yeah, but the problem is that we can't remove them cause it could be breaking a package if one is using this. But would be very nice to get it cleaned out. I have to investigate what our clean-up strategy is.

bjarnef:
@nielslyngsoe yes, but packages really shouldn't use core css classes - only core components and their own custom styling :)

PR #6613, opened 2019-10-08
V8: Do not warn about unpublishing multiple languages for invariant content

kjac:
### Prerequisites

### Description

If you have multiple languages, the warning in the delete content dialog is the same for variant *and* invariant content. As the warning mentions unpublishing specific languages, it seems a bit misplaced for culture invariant content:

image

Also I'm not really a big fan of the wording; "...go and unpublish it instead" comes off a bit rude to me (maybe it's just me?).

With this PR applied, here's how deleting variant content looks (just a change of wording, really):

image

...and here's now deleting *invariant* content looks:

image


emmaburstow:
Kenn! Thanks for the PR and the stunning-as-usual documentation here. We have loads to look at but as soon as we look at this, we'll let you know.

Em

Issue #6614, opened 2019-10-08
Member group tree nodes are not highlighted


bjarnef:
It seems when selecting member group tree nodes they are highlighted as selection of other tree nodes.

Reproduction
------------

### Bug summary

When selecting member group nodes they are not highlighted on selection.

### Specifics

- Umbraco v8.2.0 - latest changes in branch `v8/dev`
- Google Chrome v77.0.3865.90

### Steps to reproduce

- Create a few member groups
- Browse between the member group nodes and note they are not highlighted on selection.

### Expected result

When selecting member group tree nodes, they should be highlighted as other tree nodes to be consistent with rest of the UI.

### Actual result

The member group tree nodes are not highlighted on selection.

It does sync tree by using `syncTree` in `navigationService` and `path`.
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI.Client/src/views/membergroups/edit.controller.js#L52

However when inspecting the request `GetById` in `MemberGroup` api the path property (string) returns e.g. `-11159` where it is expected to return `-1,1159` where `-1` is root level and `1159` is the member group id.

![devenv_2019-10-08_21-34-44](https://user-images.githubusercontent.com/2919859/66428829-1214a100-ea17-11e9-9a68-979abcdea9ba.png)

![2019-10-08_22-08-59](https://user-images.githubusercontent.com/2919859/66429480-65d3ba00-ea18-11e9-88e7-6376524d4c61.gif)

It seems to be a mapping issue here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web/Editors/MemberGroupController.cs#L34

@Shazwazza @zpqrtbnk do you know why the path is not returned as `"-1,1159"` instead of `"-11159"`?

bjarnef:
PR submitted: https://github.com/umbraco/Umbraco-CMS/pull/6615

PR #6615, opened 2019-10-08
Fix incorrect path in member group mapping

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6614

### Description
Member group nodes was not highlighted because an incorrect path in member group mapping.
The path was missing a comma, so instead of `syncTree` in `navigationService` was using `path: "-1,1159"` it was using  `path: "-11159"`.


nul800sebastiaan:
Yikes! That is clearly an oversight, thanks @bjarnef !

Issue #6616, opened 2019-10-08
Exception in Terminate method of single component stops all other components


ronaldbarendse:
When adding a custom `IComponent`, the default scaffolding generates:

```c#
public class TestComponent : IComponent
{
	public void Initialize()
	{
		throw new NotImplementedException();
	}

	public void Terminate()
	{
		throw new NotImplementedException();
	}
}
```

If this component is [composed/added](https://our.umbraco.com/documentation/Implementation/Composing/#componentcomposert) to the `ComponentCollectionBuilder`, starting up Umbraco will result in a `BootFailedException`: this is great, as it fails fast and displays a nice descriptive exception 👍 

Next you implement the `Initialize` method, but leave the `Terminate` method unchanged, all seems fine: Umbraco starts up and everything works. However, when terminating, the `NotImplementedException` is thrown and this actually stops terminating all other components! 😢

Reproduction
------------
### Specifics
Umbraco 8.1.5

The documentation contained 5 example components that had this exact issue (that are fixed in https://github.com/umbraco/UmbracoDocs/pull/1956 and https://github.com/umbraco/UmbracoDocs/pull/1957). So some people might have used these examples, always causing an exception to be thrown while terminating.

Other times, something just goes wrong while terminating a component: this shouldn't result in other components not (correctly) terminating.

### Steps to reproduce
Add the following components:

```c#
public class ComponentsComposer : IUserComposer
{
	public void Compose(Composition composition)
	{
		composition.Components().Append<TestComponent>();
		composition.Components().Append<Test2Component>();
	}
}

public class TestComponent : IComponent
{
	public void Initialize()
	{ }

	public void Terminate()
	{
		throw new NotImplementedException();
	}
}

public class Test2Component : IComponent
{
	public void Initialize()
	{ }

	public void Terminate()
	{
		throw new NotImplementedException();
	}
}
```

Run the website, make it restart (e.g. save the `Web.config`) and go to the log viewer: there's only a single error logged from `Test2Component`. As terminating is done in a reverse order, the `TestComponent` isn't terminated (and the second exception isn't logged):

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Core/Composing/ComponentCollection.cs#L37-L51

### Expected result
Make sure all components can terminate and log all exceptions thrown while doing so.

### Actual result
Terminating of the components stops at the first exception.

PR #6617, opened 2019-10-08
Log exceptions thrown while terminating components

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6616.

### Description
This PR makes sure the `Terminate` method of all components are executed and any exceptions are logged as errors.

I first combined all exceptions into a single `AggregateException`, but that would still result in an unhandled exception while terminating the runtime and not run all logic within the `HandleApplicationEnd`:

https://github.com/umbraco/Umbraco-CMS/blob/27223738c6af045b637bb9944b53134e681ce5e5/src/Umbraco.Core/Runtime/CoreRuntime.cs#L281-L284

https://github.com/umbraco/Umbraco-CMS/blob/3332e93d4c2b44fbd11fb86fe89e62b61b28f96d/src/Umbraco.Web/UmbracoApplicationBase.cs#L109-L150

Testing can be done by following the reproduction steps from the linked issue. It should log both exceptions as errors and still log the detailed shutdown message.

### DIFF
--- a/src/Umbraco.Core/Composing/ComponentCollection.cs
+++ b/src/Umbraco.Core/Composing/ComponentCollection.cs
@@ -39,8 +39,6 @@ namespace Umbraco.Core.Composing
         {
             using (_logger.DebugDuration<ComponentCollection>($"Terminating. (log components when >{LogThresholdMilliseconds}ms)", "Terminated."))
             {
-                var exceptions = new List<Exception>();
-
                 foreach (var component in this.Reverse()) // terminate components in reverse order
                 {
                     var componentType = component.GetType();
@@ -53,14 +51,9 @@ namespace Umbraco.Core.Composing
                         }
                         catch (Exception ex)
                         {
-                            exceptions.Add(ex);
+                            _logger.Error(componentType, ex, "Error while terminating component.");
                         }
                     }
-                }
-
-                if (exceptions.Count > 0)
-                {
-                    throw new AggregateException("One or more errors occurred while terminating components.", exceptions);
                 }
             }
         }
--- a/src/Umbraco.Core/Composing/ComponentCollection.cs
+++ b/src/Umbraco.Core/Composing/ComponentCollection.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using Umbraco.Core.Logging;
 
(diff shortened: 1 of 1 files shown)


ronaldbarendse:
To explain the reasoning behind this change: components represent single pieces of logic and shouldn't affect each-other. So when one component throws an exception while terminating, all others should not be impacted.

Besides: the application is stopping, so even if components did rely on successful terminating of another component, it shouldn't impact a running application.

ronaldbarendse:
@nul800sebastiaan What's the reply from HQ? This PR is really small, but fixes a major problem.

If a custom `IComposer`/`IUserComposer` throws an exception, the following code within Umbraco core will definitely not run (as terminating is done in reverse order and core components are added/initialized first):

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Core/Compose/ManifestWatcherComponent.cs#L38-L44

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web/SignalR/PreviewHubComponent.cs#L30-L33

nul800sebastiaan:
@ronaldbarendse Hmm, I got a promise that someone would reply but they didn't ;-) I've asked again!

warrenbuckley:
Code looks/reads fine - off to test :)

src/Umbraco.Core/Composing/ComponentCollection.cs
```diff
@@ -43,8 +44,15 @@ public void Terminate()
...
-                        component.Terminate();
-                        component.DisposeIfDisposable();
+                        try
+                        {
+                            component.Terminate();
+                            component.DisposeIfDisposable();
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.Error(componentType, ex, "Error while terminating component.");
```
warrenbuckley:
```suggestion
                            _logger.Error(componentType, ex, $"Error while terminating component '{componentType.FullName}'.");
```

warrenbuckley: Just a super minor thing to help with readability of logs that the rendered message log contains the class/type of the component. Yes, it already exists as the `SourceContext` on the log, but this makes it easier to read than just `Error while terminating component.`

warrenbuckley: Was hoping to update your PR directly @ronaldbarendse (I don't have permissions to push/update your PR)
I will merge this PR but will amend with the minor change I have suggested

ronaldbarendse: That probably makes it easier to search indeed, although I would then change the source context to ‘ComponentCollection‘, as that is the calling type and the same as the other logging calls...

ronaldbarendse: @warrenbuckley Can we change this minor thing before it gets released? As the error itself isn't logged from the component and all other logging within this file use `ComponentCollection` as source context, it's best to change this into:
```suggestion
                            _logger.Error<ComponentCollection>(ex, "Error while terminating component {ComponentType}.", componentType.FullName);
```

warrenbuckley: Good spot @ronaldbarendse I have fixed in this commit here - d378495942fe1828571444aeb45e84143059f3c4

Issue #6620, opened 2019-10-08
Remove usage of obsolete string extension method ToSHA1


prjseal:
In the StringExtensions class, it says that the string extension method `.ToSHA1()` is now obsolete and may be removed in future versions. It advises the developer to use `.GenerateHash()` instead.

However, there are still usages of `.ToSHA1()` in the code base and they show up in the output window as warnings.

These need to be updated to use GenerateHash()

I have created a PR for this already.

Shazwazza:
@nul800sebastiaan shouldn't this be marked as breaking?

Shazwazza:
And marked for some release?

PR #6621, opened 2019-10-08
- removed usage of obsolete ToSHA1() and replaced it with GenerateHash()

prjseal:
In the StringExtensions class, it says that the string extension method `.ToSHA1()` is now obsolete and may be removed in future versions. It advises the developer to use `.GenerateHash()` instead.

However, there are still usages of `.ToSHA1()` in the code base and they show up in the output window as warnings.

These need to be updated to use GenerateHash()

This PR fixes the issue #6620

### DIFF
--- a/src/Umbraco.Core/MainDom.cs
+++ b/src/Umbraco.Core/MainDom.cs
@@ -65,7 +65,7 @@ namespace Umbraco.Core
             // a new process for the same application path
 
             var appPath = HostingEnvironment.ApplicationPhysicalPath;
-            var hash = (appId + ":::" + appPath).ToSHA1();
+            var hash = (appId + ":::" + appPath).GenerateHash();
 
             var lockName = "UMBRACO-" + hash + "-MAINDOM-LCK";
             _asyncLock = new AsyncLock(lockName);
--- a/src/Umbraco.Core/Models/UserExtensions.cs
+++ b/src/Umbraco.Core/Models/UserExtensions.cs
@@ -67,7 +67,7 @@ namespace Umbraco.Core.Models
 
             if (user.Avatar.IsNullOrWhiteSpace())
             {
-                var gravatarHash = user.Email.ToMd5();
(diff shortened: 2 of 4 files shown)


nul800sebastiaan:
I'm not sure this is a good idea.. I don't know if the hash will be different and if the new hashes (or the old hashes) can still be read properly after this. I'll have to ask around! :)

ronaldbarendse:
Gravatar requires a MD5 hash, so changing that to `GenerateHash` will not work in most cases (as MD5 isn't the default hashing algorithm). See https://nl.gravatar.com/site/implement/hash/.

The other problem is that the hashing algorithm of `GenerateHash` may change, causing redirects to not work anymore and it might have other effects (e.g. an hash is also created on the client side, not matching the server side anymore).

prjseal:
I agree with your points but it seems odd that you are encouraging the end user to use GenerateHash but in fact it gives 2 different results.

ronaldbarendse:
The obsolete messages are quite confusing indeed and are probably the actual bug :wink:

The `GenerateHash` function uses a setting to determine whether MD5 or SHA1 is used:
https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Core/StringExtensions.cs#L729-L734

This is perfectly fine if you don't depend of the algorithm and only use it temporarily (e.g. user sessions).

I'm not sure, but we could change this to a private method:

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Core/StringExtensions.cs#L763

nul800sebastiaan:
Thanks, I'll close this one. Thanks for the first attempt @prjseal! 👍

prjseal:
We got there in the end anyway. Thanks for the final fix Ronald

ronaldbarendse:
And credits to you for raising the issue @prjseal :wink:

Issue #6622, opened 2019-10-08
Pass in user id when using EmptyRecycleBin method


prjseal:
When you build the core project you get warnings about needing to pass in the userId to the EmptyRecycleBin method as the method without userId in it is obsolete and marked as so.

There are 3 tests which use this method without passing the user in. If we update these then the messages will go away.

nul800sebastiaan:
Thanks @prjseal !

Note: there's is no requirement any more to first create an issue and then add a PR, only a PR is just as good. 👍

prjseal:
Thanks for letting me know about not needing to create the issue first.

PR #6623, opened 2019-10-08
- updated usage of EmptyRecycleBin to pass in the user id as instructed

prjseal:
When you build the core project you get warnings about needing to pass in the userId to the EmptyRecycleBin method as the method without userId in it is obsolete and marked as so.

There are some tests which use this method without passing the user in. If we update these then the messages will go away.

This PR updates them and therefore fixes the issue #6622


nul800sebastiaan:
Thanks @prjseal - looks like the build succeeded! 👍

PR #6625, opened 2019-10-09
Data type usage reporting and deletion warning

Shazwazza:
Displays a report when viewing a data type to show where it is used: Document types, media types, member types and optionally for which property editors.  This feature also adds a new deletion prompt when deleting a data type to let you know exactly where it is used before you agree to delete it.

image

image

## Changes

* data type ui updated to support content app layout - this means moving some stuff around since each view is now responsible for the parts of the data type layout
* data type angular updated to use more modern apis (controller as, etc...)
* report data is only requested when the releations tab is active
* new IDataTypeService.GetReferences
* new DataTypeController.GetReferences endpoint for angular
* Fixes navigating in angular when navigating via an html local url (there was a bug in this)
* updated delete dialog/controller to show where the data type is used, this now uses more modern apis  (controller as, etc...)

## Testing

* Ensure unit tests all pass
* Review code
* Ensure data type creation & editing all works as expected
* Ensure the relations report is working and is accurate when clicking the relations tab
* Ensure you can click through to the related content type items
* Ensure the relations delete report is working and is accurate when trying to delete a data type and make sure you can still delete the data type


src/Umbraco.Web.UI.Client/src/views/datatypes/datatype.edit.controller.js
```diff
@@ -177,26 +134,102 @@ function DataTypeEditController($scope, $routeParams, appState, navigationServic
...
+
+                });
+        }
+        else {
+            loadDataType();
+        }
+        
+        var labelKeys = [
+            "general_settings",
+            "relations_tabName"
```
leekelleher: I wonder if "Usage" could be friendlier than "Relations"?

With "Settings", my first thought was it should be "Configuration" or "Configure", (mostly because the API refers to `ConfigurationEditor`, `ConfigurationField`) - but I'm less precious about this one. ☺️

ault: "Usage" or "Used By" much friendlier imho.

Shazwazza: I had Usage before, but in some theories it should be Usages and then I didn't want to have an english war, so I chose References ... which has now been updated in JS to be consistent with c# as References.

bergmania:
@nielslyngsoe 
I couldn't find the unit tests of the `umbCmsJoinArray` filter? You probably forgot to add the file to the commit

I added some tests myself and think some of the behavior was a bit funny. 
If I used `null` as seperator, then the output uses `'null'` as sperator. I expected `''`. So I changed this

Hopefully you can figure out how to merge my tests with the one you forgot to push.

bergmania:
Seems to work as expected :)

Do you agree on my change to the filter, and comments?

src/Umbraco.Web/Models/ContentEditing/DataTypeReferences.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+
+namespace Umbraco.Web.Models.ContentEditing
+{
+    [DataContract(Name = "dataTypeUsages", Namespace = "")]
```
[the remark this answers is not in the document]
Shazwazza: @bergmania that is a mistake, the class used to be usages too, this should be changed - it won't affect JS

PR #6627, opened 2019-10-09
Unpublished content appearing in macro results

stevemegson:
### Prerequisites

### Description
`UmbracoHelper.RenderMacro` always renders the macro as though in preview mode. Also, the results of macros rendered inside grid editors or RTEs are cached even though they're based on preview content. That cached result can then be shown when viewing the page.

`MacroRenderingController` explicitly requests the preview version of the content, which is correct since it's used by the editors. However, `UmbracoContext.InPreviewMode` is still false, so `MacroRenderer` will cache the result. Calling `UmbracoContext.ForcedPreview` fixes this.

`UmbracoComponentRenderer` also explicitly requests the preview version. This is necessary when being called from `MacroRenderingController`, but also affects `UmbracoHelper`.

### Testing
UmbracoHelper:
- Create a doctype with a text string property `text` and an RTE.
- Create a macro which displays the value of `text`.
- Include that macro in the template with `Umbraco.RenderMacro`.
- Create a document and set the text property to "Published value". Save and publish.
- Set the text property to "Unpublished value" and Save.
- Check that previewing the page shows "Unpublished value" and viewing the page normally shows "Published value".

RTE:
- Enable inserting and rendering the macro in rich text editors, and set a long cache duration.
- Remove `RenderMacro` from the template and add the RTE value.
- Insert the macro in the RTE, check that "Unpublished value" is shown in the editor.
- View the page and check that "Published value" is shown.

---
_This item has been added to our backlog [AB#3414](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3414)_

### DIFF
--- a/src/Umbraco.Web/UmbracoComponentRenderer.cs
+++ b/src/Umbraco.Web/UmbracoComponentRenderer.cs
@@ -92,12 +92,12 @@ namespace Umbraco.Web
             if (contentId == default)
                 throw new ArgumentException("Invalid content id " + contentId);
 
-            var content = _umbracoContextAccessor.UmbracoContext.Content?.GetById(true, contentId);
+            var content = _umbracoContextAccessor.UmbracoContext.Content?.GetById(contentId);
 
             if (content == null)
                 throw new InvalidOperationException("Cannot render a macro, no content found by id " + contentId);
 
-            return RenderMacro(alias, parameters, content);
+            return RenderMacro(content, alias, parameters);
         }
 
         /// <summary>
@@ -107,7 +107,7 @@ namespace Umbraco.Web
         /// <param name="parameters">The parameters.</param>
         /// <param name="content">The content used for macro rendering</param>
         /// <returns></returns>
-        private IHtmlString RenderMacro(string alias, IDictionary<string, object> parameters, IPublishedContent content)
+        public IHtmlString RenderMacro(IPublishedContent content, string alias, IDictionary<string, object> parameters)
         {
             if (content == null) throw new ArgumentNullException(nameof(content));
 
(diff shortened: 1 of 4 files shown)


emmaburstow:
Hi Steve! Nice to see so much of you here. Thanks for the pr. We'll look and if we need anything, we'll let you know. 

Em

src/Umbraco.Web/UmbracoComponentRenderer.cs
```diff
@@ -117,51 +117,7 @@ private IHtmlString RenderMacro(string alias, IDictionary<string, object> parame
                 x => x.Key.ToLowerInvariant(),
                 i => (i.Value is string) ? HttpUtility.HtmlDecode(i.Value.ToString()) : i.Value);
             
-            var macroControl = _macroRenderer.Render(alias, content, macroProps).GetAsControl();
-
-            string html;
-            if (macroControl is LiteralControl control)
```
Shazwazza: haha yes! all of this is totally gargabe in v8 🎉  nice find to get rid of legacy. `MacroContent.Control` is actually never used and is only ever assigned if GetAsControl() is used which then just creates an arbitrary LiteralControl. This will save a little bit of overhead for folks that use a ton of macros :)

stevemegson:
Unfortunately I don't think we can avoid the parameter parsing. To build the cache key we need the parsed values.

Thinking about it again now, I'm actually leaning toward getting rid of the breaking change and accepting the double lookup. By passing an `IPublishedContent` to `RenderMacro` we make it possible for other callers to pass in preview content without also forcing the `UmbracoContext` into preview mode, and it's not obvious that you shouldn't do that. That would again lead to macro results based on preview content being cached and possibly used outside preview mode. By only accepting an ID we can be sure that the `IPublishedContent` we look up matches the context's preview state.

Shazwazza:
@stevemegson 

> Unfortunately I don't think we can avoid the parameter parsing. To build the cache key we need the parsed values.

aha, i didn't read further to see they are used for the cache key, no problemo.

Would you be able to update the PR to not have the breaking change with based on your last comment?

src/Umbraco.Web/UmbracoHelper.cs
```diff
@@ -146,7 +146,7 @@ public IHtmlString RenderTemplate(int contentId, int? altTemplateId = null)
         /// <returns></returns>
         public IHtmlString RenderMacro(string alias)
         {
-            return ComponentRenderer.RenderMacro(AssignedContentItem?.Id ?? 0, alias, new { });
+            return ComponentRenderer.RenderMacro(AssignedContentItem?.Id ?? 0, alias, new Dictionary<string, object>());
```
Shazwazza: Ideally we use a static reference to an `EmptyDictionary` reference so we aren't allocating a new instance everytime a macro is rendered, else we can pass in null and then deal with null in the `RenderMacro`. It's minor so i'll fix up.

Issue #6624, opened 2019-10-08
EnableComposer and DisableComposer attributes only work for assemblies containing IComposer


ronaldbarendse:
Following the [documentation](https://our.umbraco.com/documentation/implementation/composing/#disablecomposer--disable), you should be able to add a `DisableComposer` attribute to any assembly to disable a composer.

Going through the composing code, this attribute is only retrieved from assemblies containing at least one `IComposer`:

https://github.com/umbraco/Umbraco-CMS/blob/747a8cba2ef22a66c0f73cf7396a0ec50e5211e1/src/Umbraco.Core/Composing/Composers.cs#L243-L257

So this attribute might not always be 'seen' by the composing code (and thus not disable the composer)!

Reproduction
------------
### Bug summary
Disabling a composer using the `DisableComposer` assembly attribute does not work when used in an assembly that doesn't contain an `IComposer`.

### Specifics
Umbraco 8.1.5

### Steps to reproduce
If Umbraco is installed in a clean website project, add the following code:

```c#
[assembly: DisableComposer(typeof(Umbraco.Web.Runtime.WebInitialComposer))]

public class TestComposer : IComposer
{
	public void Compose(Composition composition)
	{ }
}
```

You should see a `BootFailedException`, as the `WebInitialComposer` isn't run.

Now comment out the `TestComposer`, rebuild and Umbraco should start up without the exception, as the `DisableComposer` attribute is not 'seen' anymore.

### Expected result
All `EnableComposer` and `DisableComposer` attributes should be considered while composing and not require adding an empty/dummy composer within the same assembly (which is the current work-around).

### Actual result
The attributes are not considered at all, thus not enabling/disabling any of the requested composers.

zpqrtbnk:
Yes - it's considering assemblies of those types that implement `IComposer` instead of considering all assemblies where we looked for `IComposer` in the first place - even those that do not contain any `IComposer`. Confirmed bug.

zpqrtbnk:
Fix implies not passing only the types to the `Composers` ctor but also the assemblies... in fact, it'd be better to just pass assemblies to `Composers` ctor and let it do what the `TypeLoader` does, in a proper way.

PR #6630, opened 2019-10-09
Get EnableComposer and DisableComposer attributes from all assemblies

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6624.

### Description
The changes in this PR makes sure all `EnableComposer` and `DisableComposer` attributes are loaded from all assemblies and not only those containing `IComposer`s.

I've done what @zpqrtbnk suggested first (https://github.com/umbraco/Umbraco-CMS/issues/6624#issuecomment-539911444): passing all assemblies in the `Composers` constructor. This made testing a lot harder and didn't feel right, as you're also passing in assemblies that aren't even used...

So I've introduced `GetAssemblyAttributes<T>()` and `GetAssemblyAttributes(param Type[] attributeTypes)` to `TypeLoader`, making it easy to retrieve all assembly attributes (from the same assemblies the types are loaded, e.g. without the known exclusions).

The final fix only passes the attributes to the `Composers` constructor. This is more in-line with how the `IComposer` types are passed in and makes testing easier (although I'm not sure how to test loading the attributes from external/empty assemblies).

### DIFF
--- a/src/Umbraco.Core/Composing/Composers.cs
+++ b/src/Umbraco.Core/Composing/Composers.cs
@@ -18,19 +18,31 @@ namespace Umbraco.Core.Composing
         private readonly Composition _composition;
         private readonly IProfilingLogger _logger;
         private readonly IEnumerable<Type> _composerTypes;
+        private readonly IEnumerable<Assembly> _assemblies;
 
         private const int LogThresholdMilliseconds = 100;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="Composers"/> class.
+        /// Initializes a new instance of the <see cref="Composers" /> class.
         /// </summary>
         /// <param name="composition">The composition.</param>
         /// <param name="composerTypes">The composer types.</param>
+        /// <param name="assemblies">The assemblies.</param>
         /// <param name="logger">A profiling logger.</param>
-        public Composers(Composition composition, IEnumerable<Type> composerTypes, IProfilingLogger logger)
+        /// <exception cref="ArgumentNullException">
+        /// composition
+        /// or
+        /// composerTypes
+        /// or
+        /// assemblies
+        /// or
+        /// logger
+        /// </exception>
+        public Composers(Composition composition, IEnumerable<Type> composerTypes, IEnumerable<Assembly> assemblies, IProfilingLogger logger)
         {
             _composition = composition ?? throw new ArgumentNullException(nameof(composition));
             _composerTypes = composerTypes ?? throw new ArgumentNullException(nameof(composerTypes));
+            _assemblies = assemblies ?? throw new ArgumentNullException(nameof(assemblies));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
@@ -103,7 +115,7 @@ namespace Umbraco.Core.Composing
                 .ToList();
 
             // enable or disable composers
-            EnableDisableComposers(composerTypeList);
+            EnableDisableComposers(composerTypeList, _assemblies);
 
             void GatherInterfaces<TAttribute>(Type type, Func<TAttribute, Type> getTypeInAttribute, HashSet<Type> iset, List<Type> set2)
                 where TAttribute : Attribute
@@ -218,7 +230,7 @@ namespace Umbraco.Core.Composing
             return text.ToString();
         }
 
-        private static void EnableDisableComposers(ICollection<Type> types)
+        private static void EnableDisableComposers(ICollection<Type> types, IEnumerable<Assembly> assemblies)
         {
             var enabled = new Dictionary<Type, EnableInfo>();
 
@@ -240,7 +252,6 @@ namespace Umbraco.Core.Composing
                 enableInfo.Weight = weight2;
             }
 
-            var assemblies = types.Select(x => x.Assembly).Distinct();
             foreach (var assembly in assemblies)
             {
                 foreach (var attr in assembly.GetCustomAttributes<EnableComposerAttribute>())
--- a/src/Umbraco.Core/Composing/Composers.cs
+++ b/src/Umbraco.Core/Composing/Composers.cs
@@ -18,7 +18,7 @@ namespace Umbraco.Core.Composing
         private readonly Composition _composition;
         private readonly IProfilingLogger _logger;
         private readonly IEnumerable<Type> _composerTypes;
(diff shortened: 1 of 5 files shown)


src/Umbraco.Core/Composing/TypeLoader.cs
```diff
@@ -505,6 +505,38 @@ private void DeleteFile(string path, int timeoutMilliseconds)
...
+
+        /// <summary>
+        /// Gets the assembly attributes.
+        /// </summary>
+        /// <param name="attributeTypes">The attribute types.</param>
+        /// <returns>
+        /// The assembly attributes of the specified types.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">attributeTypes</exception>
+        public IEnumerable<Attribute> GetAssemblyAttributes(params Type[] attributeTypes)
```
ronaldbarendse: This method currently returns zero attributes if no types are supplied (empty array/no arguments). Should we instead return all assembly attributes in that case? That would align more with how `GetCustomAttributes()` works, but might cause issues if the `attributeTypes` is dynamically generated... Thoughts?

src/Umbraco.Core/Runtime/CoreRuntime.cs
```diff
@@ -147,8 +147,9 @@ protected virtual IFactory Boot(IRegister register, DisposableTimer timer)
 
                 // get composers, and compose
                 var composerTypes = ResolveComposerTypes(typeLoader);
+                var enableDisableAttributes = typeLoader.GetAssemblyAttributes(typeof(EnableComposerAttribute), typeof(DisableComposerAttribute));
                 composition.WithCollectionBuilder<ComponentCollectionBuilder>();
```
ronaldbarendse: Any reason this method call is placed between `composerTypes` and the initialization of `Composers`?

src/Umbraco.Core/Composing/TypeLoader.cs
ronaldbarendse: I've fixed this by adding a parameterless overload 👍

src/Umbraco.Core/Composing/Composers.cs
```diff
@@ -18,19 +18,32 @@ public class Composers
...
+        /// composition
+        /// or
+        /// composerTypes
+        /// or
+        /// enableDisableAttributes
+        /// or
+        /// logger
+        /// </exception>
+
+        public Composers(Composition composition, IEnumerable<Type> composerTypes, IEnumerable<Attribute> enableDisableAttributes, IProfilingLogger logger)
```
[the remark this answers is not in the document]
ronaldbarendse: I've added the specific types in the documentation, seems like that should be sufficient 👍

ronaldbarendse:
I've done some minor changes (to make sure `GetAssemblyAttributes()` does not use deferred execution and a parameterless overload returning all attributes is available) and fixed the failing test 👍

src/Umbraco.Core/Runtime/CoreRuntime.cs
Shazwazza: @ronaldbarendse AFAIK calls to `WithCollectionBuilder` as empty calls (i.e. nothing proceeding it) adds this collection builder to the composition so that it can be resolved again with subsequent calls to `WithCollectionBuilder` in order to add to the collection. So potentially its so that this collection is added before the composers are created so the composers have access to this collection? I would assume this would need to be added before `composers.Compose()`

src/Umbraco.Core/Composing/Composers.cs
```diff
@@ -18,19 +18,30 @@ public class Composers
...
+        /// <param name="logger">The profiling logger.</param>
+        /// <exception cref="ArgumentNullException">composition
+        /// or
+        /// composerTypes
+        /// or
+        /// enableDisableAttributes
+        /// or
+        /// logger</exception>
+
+        public Composers(Composition composition, IEnumerable<Type> composerTypes, IEnumerable<Attribute> enableDisableAttributes, IProfilingLogger logger)
```
Shazwazza: @ronaldbarendse can we make this non-breaking - so that we obsolete the old constructor and also if the old constructor is used it fallbacks to the old way of finding the attributes?

Shazwazza: @ronaldbarendse what do you think abuot this breaking change?

src/Umbraco.Core/Runtime/CoreRuntime.cs
ronaldbarendse: Looks like it's not needed here, as the `composers.Compose();` method does the same:
https://github.com/umbraco/Umbraco-CMS/blob/932f7297f3473957c074325f72f632728eb8426a/src/Umbraco.Core/Composing/Composers.cs#L57-L60

src/Umbraco.Core/Composing/Composers.cs
ronaldbarendse: I've added the obsolete constructor for backwards compatibility, as both `CoreRuntime` and `Composers` classes are public and the `Boot` method can be overridden (and indeed still call this constructor overload) 👍

src/Umbraco.Core/Runtime/CoreRuntime.cs
ronaldbarendse: I just confirmed this gets called twice, so it's not required: the last commit removes the call within the `CoreRuntime`.

Shazwazza:
Looks good, i'll merge and re-run some tests. Thanks!

Shazwazza:
Works, only thing of slight concern is an impact on startup perf since there's a tiny bit more scanning. I think scanning for assembly attributes is very quick (it's been years since i benchmarked that stuff) so prob not a prob but I've added a profiling/timer in debug logging to the call that does the scanning so we can track it just in case.

Shazwazza:
The debug timing indicates this particular scanning is roughly 1ms, i think we can live with that :)

ronaldbarendse:
@Shazwazza Even if getting the custom attributes from all assemblies took longer than 1ms, it would probably be faster or at least more efficient then getting the assemblies from all `IComposer` types, filtering distinct ones and getting the attributes from those...

As this is now merged, I'll update PR https://github.com/umbraco/Umbraco-CMS/pull/6654 next 👍

PR #6631, opened 2019-10-09
V7: Don't call ClearPreviewXmlContent from ClearDocumentXmlCache

### LINKED ISSUE #6629: Error when previewing pages (under a unpublished parent) on 7.15.3
When trying to "Preview" a child node of a unpublished parent, i get a 404, unless i "preview" the unpublished parent first.

There seems to be a relation to "Parent" that gets lost, or cut off, when unpublishing. 

Reproduction
------------

### Bug summary

My site structure has multiple levels of nodes. When using the "Preview-mode", I experience issues if the page I am attempting to preview, has a unpublished parent.

This problem only persists until I have clicked "Preview" on the unpublished page, and on all the ancestors between the unpublished page, and the page I actually want to preview.

### Specifics

My current umbraco version:
Umbraco version 7.15.3 assembly: 1.0.7180.24662

My site structure looks roughly like this:
chrome_difAok6eaH

I am currently route-hijacking all the nodes in the chain of pages that experience this.

I have a custom 404-page.

The site was upgraded (nuget) from 7.15.0 to 7.15.3, but I don't know if this issue was present on the site when it was running 7.15.0.

### Steps to reproduce

Create a root node (A)
Create a child node (B) on (A).
Create a child node (C) on (B).
Create a child node (D) on (C).

At this point, I can preview all pages.

Click "Unpublish" on (B).

At this point, I can not preview C or D.

If i click "Preview" on B, the preview stops giving 404 on C. 
If I then click "Preview" on C, the preview stops giving 404 on D.

### Expected result

Pages should be preview-able without having to chain-preview all the way down from the unpublished parent

### Actual result

My custom 404 page is displayed unless i click "Preview" all the way down from the unpublished parent, in the right order.

---
_This item has been added to our backlog [AB#3416](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3416)_

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #6629

### Description
Removing a document from the in-memory XML cache also removes it (and all its descendants) from the in-memory XML cache for preview, so it is no longer possible to preview the descendants.

Testing:
- Unpublish a page which has some descendants
- Preview one of the descendants, you should get a correct preview rather than a 404.

### DIFF
--- a/src/Umbraco.Web/umbraco.presentation/content.cs
+++ b/src/Umbraco.Web/umbraco.presentation/content.cs
@@ -478,8 +478,6 @@ namespace umbraco
                     safeXml.AcceptChanges();
                 }
             }
-
-            ClearPreviewXmlContent(id);
         }
 
         /// <summary>


emmaburstow:
Hey Steve,

Thanks for the work here. We'll take a look and shout if we need you.

Em

Shazwazza:
Yep I think this change is correct, the call to this method must have been an oversight. The only calls to `ClearPreviewXmlContent(int id)` are done from the `UnpublishedPageCacheRefresher` which is responsible for evicting caches when unpublished documents are totally deleted which seems correct. So it does seem odd that this was also being called when evicting a document from the published XML (since that document wasn't actually deleted).

Just for reference here... we call the overload `ClearPreviewXmlContent()` from `SortNodes(int parentId)`, however this overload differs in the way it works - It removes the entire in memory XML preview cache from memory which would force it to be rebuilt from scratch the next time it is needed. This is a 'safe' way to totally rebuild the xml preview cache and I think that is fine. The issue with `ClearPreviewXmlContent(int id)` is that it does not force any rebuilding of the cache, it just removes one of the xml node elements from the xml cache and it would never come back unless the whole cache was evicted or that single item was added back in.

@stevemegson do you think the above is accurate?

stevemegson:
@Shazwazza that looks right to me.

It might have originally been added expecting that the single preview mode would still update the preview for ancestors and descendants of a node when it is previewed, which would have dealt with any missing elements.

nul800sebastiaan:
Sorry for the wait @stevemegson - no news as to when 7.15.4 will be out, but this PR is good to go for when that happens! 👍

PR #6643, opened 2019-10-10
Clean up of change password section in member details

poornimanayar:
### Prerequisites

I have done a little clean up of the change password section in member details. I have made the button styling in line with rest of the site and used the `umb-checkbox` instead of the [input checkbox](url) to make it consistent. The "Change Password" button has a dark blue background now like the Change Password button in the user dialog and the Cancel button is "info" style.

image

image


emmaburstow:
Hey P,

Thanks for the work you've done here. A quick scan of the list tells me you'll be getting a hacktoberfest tee. Lovely change here. Great work.

Em

src/Umbraco.Web.UI.Client/src/views/components/users/change-password.html
```diff
@@ -6,19 +6,15 @@
...
         <div ng-switch-when="true">
 
             <ng-form name="changePasswordForm">
                 <umb-control-group alias="resetPassword" label="@user_resetPassword" ng-show="vm.config.enableReset">
-                    <input type="checkbox" ng-model="vm.passwordValues.reset"
-                           name="resetPassword"
-                           val-server-field="resetPassword"
-                           no-dirty-check
-                           ng-change="vm.showReset = !vm.showReset" />
+                    <umb-checkbox  model="vm.passwordValues.reset" />
```
elit0451: @poornimanayar I think you've forgotten to include some of the previous attributes like:
`<umb-checkbox  model="vm.passwordValues.reset" name="resetPassword" val-server-field="resetPassword" on-change="vm.showReset = !vm.showReset"/>`

poornimanayar:
Oh dear! Let me check!

src/Umbraco.Web.UI.Client/src/views/components/users/change-password.html
elit0451: @BatJan how about `no-dirty-check` attribute?

BatJan: @elit0451 I'll need to have a look at the directive later. It's been extended a bit since I initially introduced it to core so don't have the full overview of what's possible and not 😅 

Apart from that the name attribute should not be necessary though since angularjs only cares about the model value @poornimanayar 👍

src/Umbraco.Web.UI.Client/src/views/components/users/change-password.html
```diff
@@ -6,19 +6,15 @@
     </div>
     <div ng-switch="vm.changing">
         <div ng-switch-when="false">
-            <a href="" ng-click="vm.doChange()" class="btn btn-small">
+            <a href="" ng-click="vm.doChange()" class="btn umb-button__button btn-action umb-button--">
```
BatJan: Hey @poornimanayar - It would be awesome if you could also change this to `<buttton type="button" ng-click="vm.doChange()" class="btn umb-button__button btn-action">...</button>`

Maybe you could even have a look at using the `<umb-button>` component/directive instead. There are a few examples on how this can be setup if you search the source code. Only thing I can't tell currently is what the styling needs to look like in this scenario... :-)

src/Umbraco.Web.UI.Client/src/views/components/users/change-password.html
```diff
@@ -62,7 +58,7 @@
                     </span>
                 </umb-control-group>
 
-                <a href="" ng-click="vm.cancelChange()" ng-show="vm.showCancelBtn()" class="btn btn-small">
+                <a href="" ng-click="vm.cancelChange()" ng-show="vm.showCancelBtn()" class="btn umb-button__button btn-cancel umb-button--">
```
BatJan: @poornimanayar My above comment also goes for this line 😃 - Let me know if you need further pointers. I'll be happy to assist 👍

src/Umbraco.Web.UI.Client/src/views/components/users/change-password.html
BatJan: @elit0451 @poornimanayar Just had a look at the file and the no-dirty-check directive is probably not needed since the checkbox should accept a boolean value.

poornimanayar:
@elit0451 @BatJan , incorporated your comments. I havent changed the button styles. As I feel they are in line with the rest of the application already. The "Change Password" button has been already changed to an action with this PR, in line with the "Change Password" for users. The "Cancel" is the info style grey. 

Here is a screenshot.
member-change-password

nul800sebastiaan:
Thanks very much @poornimanayar! 👍

PR #6644, opened 2019-10-10
Changed the styling of Delete button in the language section

poornimanayar:
Made the button styling on the Delete button in the languages section more consistent.

**Before**
image

**After**
image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/languages/overview.html
+++ b/src/Umbraco.Web.UI.Client/src/views/languages/overview.html
@@ -63,7 +63,7 @@
                         <td style="text-align: right;">
                             <umb-button ng-if="!language.isDefault"
                                         type="button"
-                                        size="xxs"
+                                        size="sm"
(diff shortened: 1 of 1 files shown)


emmaburstow:
Another lovely change - we'll let you know if we need you for anything.

Em

elit0451:
Thanks for the suggestion @poornimanayar :blush:

However, after having a closer look at the dedicated space for the **default** language and the **added ones** it turns out that it looks more proportionate if we keep the size as `xxs`. I have made some screenshots so you can better see the difference :slightly_smiling_face:

## Size **XXS**
LangBtnXXS

## Size **SM**
LangBtnSM


I have considered that you'd like to have more consistency but the other buttons on the `overview.html` page you want to implement the change have their size as **xs**. I've also tried that but you can still see the difference between the languages space division.

Therefore, I will, unfortunately, close this PR. But would love to see any further suggestions :sparkles: H5YR

PR #6646, opened 2019-10-10
Changed the styling of change password button in users section to be action style

poornimanayar:
The "Change Password" button in the user dashboard is "info" style. I have made it action style to make it consistent along the "Change Password" button in the user profile dialog. There is a similar PR for the members section #6643 

**Before**
Before

**After**
After


emmaburstow:
Great work here P. Thanks and we'll let you know if we need you to do anything else.

Em

elit0451:
Looks good @poornimanayar 👍 Merging

PR #6649, opened 2019-10-10
Better exception handling for templates when they exceed 255 characters

poornimanayar:

If there's an existing issue for this PR then this fixes <https://github.com/umbraco/Umbraco-CMS/issues/4633>

In this PR I have fixed the exception around template name and member name. When saving a new template or renaming an existing template, if there are over 255 characters in the name an exception is thrown. I have also added a `maxlength` attribute to the header so that more than 255 characters cannot be inputted


ronaldbarendse:
The null and length checks should be separated and throw an `ArgumentNullException` or `ArgumentOutOfRangeException`. As empty names are also not valid, an `ArgumentException` could be added in-between. See also https://github.com/umbraco/Umbraco-CMS/pull/6599.

nul800sebastiaan:
@poornimanayar Can you have a look at Ronald's suggestions please?

poornimanayar:
@nul800sebastiaan sure!

src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-header.html
```diff
@@ -37,7 +37,7 @@
                                umb-auto-focus
                                val-server-field="Name"
                                required
-                               autocomplete="off" />
+                               autocomplete="off" maxlength="255"/>
```
ronaldbarendse: Just a little suggestion to align the new attribute on a new line:
```suggestion
                               autocomplete="off"
                               maxlength="255"/>
```

src/Umbraco.Core/Services/Implement/FileService.cs
```diff
@@ -405,6 +405,11 @@ public long GetScriptFileSize(string filepath)
         /// <returns></returns>
         public ITemplate CreateTemplateWithIdentity(string name, string alias, string content, ITemplate masterTemplate = null, int userId = Constants.Security.SuperUserId)
         {
+            if (name != null && name.Length > 255)
+            {
+                throw new InvalidOperationException("Name cannot be more than 255 characters in length.");
+            }
```
ronaldbarendse: So this check should be separated into:
```c#
if (name == null) throw new ArgumentNullException(nameof(name));
if (string.IsNullOrWhiteSpace(name)) throw new ArgumentException("Value can't be empty or consist only of white-space characters.", nameof(name));
if (name.Length > 255) throw new ArgumentOutOfRangeException(nameof(name), "Value can't be more than 255 characters in length.");
```

poornimanayar: Thanks @ronaldbarendse , working on it now :-)

src/Umbraco.Core/Services/Implement/FileService.cs
```diff
@@ -539,6 +544,12 @@ public IEnumerable<ITemplate> GetTemplateChildren(int masterTemplateId)
         /// <param name="userId"></param>
         public void SaveTemplate(ITemplate template, int userId = Constants.Security.SuperUserId)
         {
+            if (template.Name != null && template.Name.Length > 255)
+            {
+                throw new InvalidOperationException("Name cannot be more than 255 characters in length.");
```
ronaldbarendse: This one is a little trickier, as `template.Name` isn't an argument and therefore shouldn't throw the same exceptions as `CreateTemplateWithIdentity`.

It should either throw an `ArgumentException` referencing `template` or keep this `InvalidOperationException`. Either way: the message should indicate it can't be null, empty/white space or more than 255 characters in length.

src/Umbraco.Core/Services/Implement/FileService.cs
```diff
@@ -405,6 +405,21 @@ public long GetScriptFileSize(string filepath)
         /// <returns></returns>
         public ITemplate CreateTemplateWithIdentity(string name, string alias, string content, ITemplate masterTemplate = null, int userId = Constants.Security.SuperUserId)
         {
+            if(name == null)
+            {
+                throw new ArgumentNullException(nameof(name), "Name cannot be null or empty");
```
ronaldbarendse: It's better to use the default message 'Value cannot be null.' here:

```suggestion
                throw new ArgumentNullException(nameof(name));
```

src/Umbraco.Core/Services/Implement/FileService.cs
```diff
@@ -539,6 +554,12 @@ public IEnumerable<ITemplate> GetTemplateChildren(int masterTemplateId)
         /// <param name="userId"></param>
         public void SaveTemplate(ITemplate template, int userId = Constants.Security.SuperUserId)
         {
+            if (template.Name != null && template.Name.Length > 255)
+            {
+                throw new InvalidOperationException("Name cannot be null, empty, white-space characters only or more than 255 characters in length.");
```
ronaldbarendse:
```suggestion
                throw new InvalidOperationException("Name cannot be null, empty, contain only white-space characters or be more than 255 characters in length.");
```

src/Umbraco.Core/Services/Implement/FileService.cs
ronaldbarendse: Just to clarify: not passing a message will use the aforementioned default message 👍

poornimanayar: Aah! Missed my eyes :(

src/Umbraco.Core/Services/Implement/FileService.cs
```diff
@@ -539,6 +554,12 @@ public IEnumerable<ITemplate> GetTemplateChildren(int masterTemplateId)
         /// <param name="userId"></param>
         public void SaveTemplate(ITemplate template, int userId = Constants.Security.SuperUserId)
         {
+            if (template.Name != null && template.Name.Length > 255)
```
ronaldbarendse: Let's also change this to:
```suggestion
            if (template == null)
            {
                throw new ArgumentNullException(nameof(template));
            }
            
            if (string.IsNullOrWhiteSpace(template.Name) || template.Name.Length > 255)
```

PR #6650, opened 2019-10-10
Update the create dialog on Relation Types to use umb-radiobutton

poornimanayar:
I have updated the Create Dialog on Relation Types to use `umb-radiobutton`. It is currently using `input type="radio"` at the moment

**Before**
image

**After**
image


emmaburstow:
So 👏 much 👏 better 👏 

Great work P. We'll shout if we need a change.

Em

PR #6651, opened 2019-10-10
Handle duplicate keys in DataTypeMigration

### LINKED ISSUE #6588: DB migration failure
Hi all

we are trying to upgrade our 7.15.1 to 8.1.1 we get the following failure:

Error during installation

The database failed to upgrade. ERROR: The database configuration failed with the following message: An item with the same key has already been added. Please check log file for additional information (can be found in '/App_Data/Logs/')

See the log for full details (logs can typically be found in the App_Data\Logs folder).

when going to the log i can see the following: {"@t":"2019-09-24T14:05:03.4542856Z","@mt":"Installation step {Step} failed.","@l":"Error","@x":"System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> Umbraco.Web.Install.InstallException: The database failed to upgrade. ERROR: The database configuration failed with the following message: An item with the same key has already been added.\n Please check log file for additional information (can be found in '/App_Data/Logs/')\r\n at Umbraco.Web.Install.InstallSteps.DatabaseUpgradeStep.ExecuteAsync(Object model)\r\n --- End of inner exception stack trace ---\r\n at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)\r\n at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)\r\n at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)\r\n at Umbraco.Web.Install.Controllers.InstallApiController.

any idea?

thanks E

stevemegson:
### Prerequisites
If there's an existing issue for this PR then this fixes #6588

### Description
DataTypeMigration assumes that it can call `ToDictionary(x => x.Alias, x => x)` on the prevalues for each data type, but dropdowns created on older versions may have blank aliases.

This PR modifies it to use an `IEnumerable<PreValueDto>` instead.

### DIFF
--- a/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/DataTypeMigration.cs
+++ b/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/DataTypeMigration.cs
@@ -74,9 +74,18 @@ namespace Umbraco.Core.Migrations.Upgrade.V_8_0_0
                     .From<DataTypeDto>()
                     .Where<DataTypeDto>(x => x.NodeId == group.Key)).First();
 
+                // check for duplicate aliases
+                var aliases = group.Select(x => x.Alias).Where(x => !string.IsNullOrWhiteSpace(x)).ToArray();
+                if (aliases.Distinct().Count() != aliases.Length)
+                    throw new InvalidOperationException($"Cannot migrate prevalues for datatype id={dataType.NodeId}, editor={dataType.EditorAlias}: duplicate alias.");
+
+                // handle null/empty aliases
+                int index = 0;
+                var dictionary = group.ToDictionary(x => string.IsNullOrWhiteSpace(x.Alias) ? index++.ToString() : x.Alias);
+
                 // migrate the preValues to configuration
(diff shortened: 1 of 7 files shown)


ronaldbarendse:
This will be a breaking change though (the public method signature of `GetConfiguration` changes)... If the empty aliases can't be filtered out or get a new generated alias, there's probably no other way to migrate the data types.

stevemegson:
Ah yes, I thought I'd checked that everything is internal, but of course the interface is public so that people can provide migrations for other property editors.

Plan B assigns ascending keys for empty aliases, as v7's multivalues prevalue editor does. The `Alias` property on the `PreValueDto` is left blank because `DefaultPreValueMigrator` expects to see that.

I've also moved some of the error checking from `DefaultPreValueMigrator` out into `DataTypeMigration` to give a more descriptive message if there are duplicate non-empty aliases.

nul800sebastiaan:
Cool fix there @stevemegson - thanks! 👍

PR #6654, opened 2019-10-10
Add automatically composed components

ronaldbarendse:
### Prerequisites

### Description
This feature PR is based on the problems with wiring up `IComponent`s, as it [requires a composer to add the component](https://our.umbraco.com/documentation/implementation/composing/#example---creating-a-component-to-listen-for-contentservicesaving-events) and you have to implement the interface without keeping the `throw new NotImplementedException();` on either the `Initialize` or `Terminate` method (see issue https://github.com/umbraco/Umbraco-CMS/issues/6616 and PR https://github.com/umbraco/Umbraco-CMS/pull/6617). Even the Umbraco documentation had multiple examples where this went wrong (see https://github.com/umbraco/UmbracoDocs/pull/1956 and https://github.com/umbraco/UmbracoDocs/pull/1957).

Continuing on the work done in https://github.com/umbraco/Umbraco-CMS/pull/6630, I've added the following interface and accompanying base class:

```c#
public interface IUserComponent : IComponent, IDiscoverable
{ }

public abstract class UserComponentBase : IUserComponent
{
    public abstract void Initialize();

    public virtual void Terminate()
    {
        // Nothing to terminate
    }
}
```

The user components are automatically discovered, enabled/disabled like composers and added to the composition. As most components only require initialization code, the base class already implements an empty `Terminate` method (we could also do this for `IComponent`, if desired: thoughts?).

The `ComponentComposer<T>` still has its use, as some components also require more composing then only adding itself.

I've picked `IUserComponent` as name for this interface/feature, as that seemed _okay_. Any thoughts about a better name? I've thought about `IComposedComponent`, but that sounds definitive, like it can't be enabled/disabled, and as this feature effectively hides the composing part, I don't think it's a great name...

And although the order of components shouldn't matter, we might want them added at the end of the components composition (`UserComponentComposer` only needs to be decorated with `IUserComposer` for this). Any thoughts about that too?

I've created this as a draft PR, because it needs some additional input and probably also unit tests (similar to testing the composers).


zpqrtbnk:
Interesting. I've always had the feeling we were discovering (magically) too many things, so I implemented `IComposer` discovery and the idea was that everything else would eventually have to be registered explicitly.

Now, I understand how that can be annoying when you don't want to "compose" anything but just register a `IComponent`. Feels awkward to have to write the code below, just to register a component.

    public class SomeComposer : ComponentComposer<SomeComponent>
    {  }

If I understand your PR correctly, there is a `UserComponentComposer` that will discover and register all user components. User components can enable/disable each others, but there is no ordering mechanism, they just start in some unspecified order. Is that correct?

And I assume the reasoning is... that must be enough for most user components, and if anyone requires precise ordering, then they can obtain it by using a non-user component + a composer.

Is there anything I am missing?

It may be a cool way to make people's lifes simpler. My only wonder is that the name (user component) does not indicate what it's doing (auto-registering). Would `IAutoComposedComponent` or something less ugly but carrying the same (more explicit) meaning make sense?

*also: need to deal with ppl trying to enable/disable non-auto components...*

stevemegson:
I've also been trying to decide whether there's a better option than one-line ComponentComposers. Could we provide a base class for "self-composing" components? Something like

```c#
public abstract class SelfComposingUserComponent : IComponent, IUserComposer
{
    {
        composition.Components().Append(GetType());
    }

    public abstract void Initialize();

    public virtual void Terminate()
    {
    }
}
```

One advantage would be that we don't need a separate mechanism for discovering, enabling and disabling them - the existing EnableComposer/DisableComposer/ComposeBefore/ComposeAfter attributes would apply. No doubt there are also disadvantages.

ronaldbarendse:
@zpqrtbnk You're spot on! The main use case for this will be a quick way to subscribe to events (like `ApplicationEventHandler` in V7), so I see this used more in application code then packages, etc. (hence the name)... As components should be independent of each other, ordering shouldn't be that important and we still have the current, verbose way of composing, if otherwise required.

@stevemegson I've actually tried the exact same thing, but it has 2 major problems: the class is initialised twice (once when composing and again when initializing the component) and because of that you can't inject dependencies in the constructor.

stevemegson:
@ronaldbarendse Ah yes, I knew I had to be missing an obvious problem. It all looks fine until you want your component to actually do anything interesting.

ronaldbarendse:
This feature currently only requires a nice, descriptive name for the interface/base class. As already noted by @zpqrtbnk, 'user component' does not indicate it's (automatically) composed and might add some confusion with `IUserComposer`.

I think `IComposedComponent` is actually the best name, because it's basically a combination of an `IComposer` that composes the `IComponent`. Although the name suggests it's always composed and can't be enabled/disabled, it seems better than `IAutoComposedComponent` (auto is redundant, as composers are always automatically discovered) or `IComposingComponent`/`ISelfComposingComponent` (since the instance itself doesn't do the composing). Any thoughts?

ronaldbarendse:
If you look at it from another perspective, `IApplicationComponent` might also be a good name. The `Initialize` and `Terminate` methods refer to the application and it will be familiar when coming from version 7.

protherj:
I like this idea.  However, I also wonder if having more options will mean that new people will have a harder time learning how to do it properly.  Sometimes more options means less clarity.  Just thought I'd throw that out there.  

If the change is made, the documentation likely needs a page on when to use each option perhaps with examples.

src/Umbraco.Core/Composing/DisableApplicationComponentAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// Indicates that an <see cref="IApplicationComponent" /> should be disabled.
+    /// </summary>
+    /// <remarks>
+    /// This attribute has greater priority than the <see cref="EnableAttribute" /> when it's marking the component itself, but lower priority when it's referencing another.
+    /// </remarks>
+    /// <seealso cref="System.Attribute" />
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
+    public class DisableApplicationComponentAttribute : Attribute
```
ronaldbarendse, line 13: Just like the `DisableComposerAttribute`, this can be added as an assembly attribute to disable the specified `IApplicationCompontent` type.

src/Umbraco.Core/Composing/DisableAttribute.cs
```diff
@@ -3,35 +3,33 @@
 namespace Umbraco.Core.Composing
 {
     /// <summary>
-    /// Indicates that a composer should be disabled.
+    /// Indicates that a type should be disabled.
```
ronaldbarendse, line 6: This is not specific to composers anymore...

ronaldbarendse (commented):
I've renamed this to `IApplicationComponent`, changed the base branch to `v8/contrib` and added the finishing touches, so it's ready for review!

This will primarily be a quick way to add events on application startup, without the need to compose the component (either manually using an `IUserComposer` or using the `ComponentComposer<T>` base class) and not having to explicitly deal with implementing the `Terminate` method:

```c#
public class ApplicationComponent : ApplicationComponentBase
{
    public void Initialize()
    {
        // Add some application events here
    }
}
```

PR #6481, opened 2019-09-30
V8/fix section selection

dawoe:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes

### Description

This PR is similar to #6480

But this time for the section selection and user selection on user groups.

Dave

PR #6659, opened 2019-10-11
V8: Fix section and user assignment to groups

kjac:
### Prerequisites

### Description

#6481 seems to have introduced an unfortunate bug: You can no assign users and sections to groups:

group-section-and-users-before

This PR fixes it; here are the same operations with this PR applied:

group-section-and-users-after


emmaburstow:
Hey Kenn,

Thanks for the impeccable documenting here. Thanks also for fixing the bug. We'll let you know if we need something.

Em

nul800sebastiaan:
Ooh ouch, thanks so much for noticing @kjac nice fix! 👍

PR #6660, opened 2019-10-11
Show the "remove" action for assigned user groups when editing users

kjac:
### Prerequisites

### Description

You can't remove the assigned user groups directly when editing a user; the workaround is to select "Add" and then deselect the groups you want removed:

edit-user-remove-groups-before

By the looks of the code, there's supposed to be a "Remove" option per group - it just hasn't been enabled. So... this PR enables that:

edit-user-remove-groups


emmaburstow:
Ah amazing. Thanks for fixing a proper pain point. Great work a ever. If we've any questions, we'll ask.

Em

nul800sebastiaan:
Fantastic, cheers! 👍

Issue #6663, opened 2019-10-11
Deletion of a language causes unintended behavior on Content tree


elit0451:
When you have content varying by culture, you can select a language you want to see your Content tree in (_ex: Danish_) other than your default one and if you delete that language afterwards:
- the Content section will still present you with the content variants names in the language you lastly viewed the tree _(in the example case it will be in Danish - the language that we just deleted)_
- if you select your variant node in the default language and you click **Save** without having made any changes you will get the error: `Code` **X** `does not correspond to an existing language. Parameter name: isoCode`

Umbraco V8.1.5

### Steps to reproduce
1 - Go to the **Settings** section and enable the toggle that will allow a Document type to vary by culture (_ex: Home node_)
2 - In the same section, add a few languages _(ex: Danish, German, etc.)_
3 - Go back to the **Content** section and fill in the necessary values for the cultures you added _(ex: Just fill in the name of the Home node in the different languages - Home DK; Home DE)_
4 - Click **Save**
5 - Choose to see the **Content** tree in any of the languages that you added _(ex: Danish)_

![image](https://user-images.githubusercontent.com/21998037/66654235-12d94d00-ec3a-11e9-84ae-7dfddeb5a521.png)

6 - Notice that the selected language will update the `mculture` attribute in the url _(ex: `mculture=da`)_
7 - Next navigate to **Setting** > **Languages** and delete the language indicated in the `mculture` attribute _(ex: Danish)_
8 - Go back to the **Content** section and verify that your tree hasn't been updated -> even after refreshing your `mculture` will still be set to the deleted language
9 - If you click on the node varying by culture _(ex: Home)_ and **Save** without any changes you will get the error due to that you try to save the node in the language you've just deleted - `Code dk does not correspond to an existing language. Parameter name: isoCode`

### Expected result
Content tree to be updated after the language it was last viewed in is deleted. No error to be caused after saving without any changes made.

### Actual result
Content tree still being shown in the language which was deleted. Error `Code` **X** `does not correspond to an existing language. Parameter name: isoCode` is caused.

---
_This item has been added to our backlog [AB#3225](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3225)_

elit0451:
Working on a fix 👍

PR #6664, opened 2019-10-11
Language variant deletion

elit0451:
Fixes: #6663 

Testing:
- Go to the **Settings** section and enable the toggle that will allow the **Home** Document type to vary by culture
- In the same section, add Danish language
- Go back to the **Content** section and fill in name of the Home node after you choose to see the tree in Danish - _Home DK_
- Click **Save**
- Choose to see the **Content** tree in Danish and notice that the selected language will update the `mculture` attribute in the url to `mculture=da`
- Navigate to **Setting** > **Languages** next and delete the Danish language, leaving you just with the default one
- Go back to the **Content** section and verify that your tree has been updated as well as the `mculture` (it would disappear if this is the last non-default language or it will be set to the default language if you have more non-default languages)
- Clicking on the Home node and saving without any changes will be successful


warrenbuckley:
Code reads fine & tested 🎉

PR #6665, opened 2019-10-11
Tweak the examine management dashboard

poornimanayar:
As per @ronaldbarendse 's suggestion [here](https://github.com/umbraco/Umbraco-CMS/pull/6225#issuecomment-540471223) , I have updated the rebuild index button on the examine management dashboard to the red colour. Also took out the heading on the dashboard. Given that we have tabs to go from one dashboard to another we probably dont need it.
image

image


emmaburstow:
Great! Really wonderful change. 

We'll ask if we have any questions. 

Em

poornimanayar:
Good catch will do it.

nul800sebastiaan:
Looking good, thanks @poornimanayar !

PR #6666, opened 2019-10-11
Clean up of the published cache status dashboard

poornimanayar:
As per @ronaldbarendse 's suggestion here, tried to do the dashboard up and bring it in line with the Examine Management Dashboard
**Before**
8d982ea9e765fed27502480b258fd5f74d5db1e3
image

**After**
image


ronaldbarendse:
This looks way better! :+1: As discussed on the other PRs, maybe the grey boxes are more suited to display output (published cache status) and we should only add a subtle divider between the buttons/titles of the Caches box?

nul800sebastiaan:
I agree with Ronald here @poornimanayar - it looks a bit weird that the button gets a full width box. Could you update that part?

poornimanayar:
Hi @nul800sebastiaan,

Updated! This is how it looks now. 
image

Poornima

ronaldbarendse:
We're getting there! Just some (more) minor remarks:
- The boxes should not be on a white background
- The divider between the boxes and the last one below 'Internals' should be removed
- The 'Published Cache Status' text might need some additional styling to represent some kind of output (like the search results in the 'Examine Management' dashboard, e.g. `<p class="umb-panel-group__details-status-action">...</p>`)

ronaldbarendse:
> @ronaldbarendse When you say "The boxes should not be on a white background" do you mean that the outer box should be removed. If so, I think thats what is causing your second point.

Yes, looks like that was not a divider, but the shadow the the last box 😉

I can only spot a grey 1 pixel line below the 'Published Cache Status' title (the 'Caches' box doesn't have this) 🔍😆

poornimanayar:
@ronaldbarendse The 1px line is the panel box shadow, like we have on the Examing Management Dashboard now :-)

nul800sebastiaan:
Nice one for consistency, thanks for all the work @poornimanayar and the feedback @ronaldbarendse! 💪

PR #6667, opened 2019-10-11
Buttons on the healthcheck dashboard made more consistent

poornimanayar:
Made the buttons on healthcheck dashboard made more consistent. I have also removed the `h3` heading.

**Before**
image

image

**After**
image

image

There was also a big space by the description which I have removed
**Before**
image

**After**
image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-panel-group.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-panel-group.less
@@ -107,4 +107,5 @@
 .umb-panel-group__details-status-action-description {
     margin-top: 5px;
     font-size: 12px;
+    padding-left:165px;
 }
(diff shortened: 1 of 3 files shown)


ronaldbarendse:
Hmm, looks like we need to make some UI conventions... :innocent:

I rather like the buttons in the header (this is used in other places, such as Rollback in the info/audit log/history). And I don't think the grey boxes around the buttons are really required here, as they're more suited for displaying output (like the search results in the Examine dashboard).

poornimanayar:
I will make the changes as per the suggestion from @ronaldbarendse . And also make similar tweaks to #6665 to get rid of the background the button. What do you think?

poornimanayar:
I have updated the dashboard as shown below. Just noticed that the background strip for the Fix button is not very visible in the gif or a screenshot for some reason. So I have attached it as a PDF.

Let me know if I have got it okay now :-)

healthcheck-dashboard

[Screenshot.pdf](https://github.com/umbraco/Umbraco-CMS/files/3785908/Screenshot.pdf)

nul800sebastiaan:
Thanks @poornimanayar !

Issue #6669, opened 2019-10-11
Consider removing css classnames starting with two hyphens to support IE11


robmaas:
In a couple of places classnames starting with two hyphens (--) are used.

The specs only allow this since the introduction of CSS Custom Properties.

This is not supported in (amongst others) IE11 and could be fairly easily fixed by changing the classnames or coming up with a new naming convention. Even though IE11 is not officially supported it would be good to fix it.

For example:
The class --notInFront causes the umb-editor__overlay to be ignored in IE11 after PR #6668 

Right now there only seem to be a handful of those classes:
--error
--notInFront
--disabled
--selected
--initialized
--not-allowed
--current
--state-selection

**Proposal for new names:**
has--error
is--notInFront
is--disabled
is--selected
is--initialized
is--not-allowed
is--current
is--state-selection

nul800sebastiaan:
As detailed on https://github.com/umbraco/Umbraco-CMS/pull/6671, we don't plan to support IE11 at all any more in v8.

Issue #5662, opened 2019-06-13
V8: Cannot build links in Rich Text Editor in IE11


john-mckillip:
Insert/Edit links in the rich text editor not displaying UI in IE11.

**Bug summary**

**Specifics**
Umbraco version: v8.0.2
Browser IE v11
FF, Chrome, and Edge work well

**Steps to reproduce**
Visit rich text editor in Umbraco back office on any page type and highlight text to hyperlink
Click Insert/Edit Link button in RT toolbar

**Expected result**
Right-side Select Link panel will open revealing options to build links to pages and assets

**Actual result**
Nothing happens; right side panel doesn't display

umbrabot:
Hi @john-mckillip,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

nul800sebastiaan:
Please note: we don't officially support IE11 (https://our.umbraco.com/documentation/Getting-Started/Setup/Requirements/) but it would be good to have this one fixed.

nul800sebastiaan:
Contradicting myself above, I know, but: we do not plan to add any more code specifically to IE11, it is not supported in v8 and we are not planning on working on that at all. Sorry for the confusion on this one, from now on this will be the consistent answer. :)

PR #6671, opened 2019-10-12
Added polyfills and fixes for older browsers (#6688).

robmaas:
Added some polyfills and changed some JS functions to use jQuery functions instead of native JS functions that are less supported (for example in IE11).

This fixes a couple of things:
- Not possible to add/edit links in the RTE editor in IE11 (#5662)
- LogViewer wouldn't work at all
- Some errors in directives

Regarding #5662: The UI does not show the grey overlay because css custom properties are not supported in IE11. Issue #6669

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbtabbedcontent.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbtabbedcontent.directive.js
@@ -6,12 +6,12 @@
 
         function link($scope, $element, $attrs) {
             
-            var appRootNode = $element[0];
+            var $appRootNode = $element.first();
             
             // Directive for cached property groups.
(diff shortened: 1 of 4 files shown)


nul800sebastiaan:
Hi @robmaas and thanks for your first contribution to Umbraco CMS! 

Unfortunately, we do not support IE11 at all in v8, adding fixes and polyfills will send the wrong message and will force us to keep supporting some of these things, which we have not committed to doing for v8.

I appreciate that you put a lot of work in for this and I am sorry to have to close this one. From looking at the related issue numbers I can see that I'm sending conflicting messages in encouraging someone to add IE11 support. From now on, I will close IE11 requests instead so that we are consistent on the message to potential contributors.

Once again, thanks so far for your contribution and we look forward to seeing more of your (non-IE11-related) handywork! 👍

PR #6672, opened 2019-10-12
V8: Handle permissions for users in multiple groups

### LINKED ISSUE #6647: V8: Tree node menu actions permissions are incorrectly recognized when assigning multiple User Groups
When dealing with multiple user groups, available node actions are incorrectly determined and may result in either too low or too high permissions (depending on the order in which groups were created).

Problem was observed and tested on Umbraco 8.1.4, however from what I could see, it still persists in newest version of the code.

Setup:

Content tree:
image

User groups:
image

Permissions (left - RO, right - Ho):
image

Assigning both groups to a single user results in him seeing (and editing if possible) both nodes in content, however action menu items are only shown based on permissions of group with lower Id (since list of user groups is always sorted by Id). The root of the problem seems to lie in _UserServiceExtensions.cs_ (19):
`return userService.GetPermissions(user, ids[ids.Length - 1]).FirstOrDefault();`

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #6647

### Description

This PR fixes an issue with item context menus permissions for users that are assigned multiple groups. See #6647 for an excellent description of the issue.

#### Testing this PR

1. Create a new user and assign two groups with different permissions defined; e.g. the Writers and the Editors groups.
2. Log in as this user and verify that all actions defined for both groups are available in the context menu for content items.
3. As the administrator, assign specific permissions to a content node, changing the access for the chosen groups.
4. As the new user, verify that the specific node permissions are applied.

### DIFF
--- a/src/Umbraco.Core/Services/UserServiceExtensions.cs
+++ b/src/Umbraco.Core/Services/UserServiceExtensions.cs
@@ -104,5 +104,17 @@ namespace Umbraco.Core.Services
 
             return found;
         }
+
+        /// <summary>
+        /// Gets the concrete assigned permissions for the provided user and node
+        /// </summary>
+        /// <param name="userService"></param>
+        /// <param name="user"></param>
+        /// <param name="nodeId"></param>
+        internal static string[] GetAssignedPermissions(this IUserService userService, IUser user, int nodeId)
+        {
+            var permissionCollection = userService.GetPermissions(user, nodeId);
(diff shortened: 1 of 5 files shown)


emmaburstow:
Hello Kenn,

Thanks for the testing steps. We'll take a look and we'll ask if we have any questions.

Em

poornimanayar:
Merged! Thanks @kjac !

PR #6677, opened 2019-10-12
Fixes #6648: MultipleTextstring validation failure

### LINKED ISSUE #6648: V8: MultipleTextstring validation failure
When creating a MultipleTextstring property with minimum 0 items, which are not mandatory but have validation (email address), Umbraco tries to validate and fails.

image
image

AndyButland:
As reported in #6648, a regular expression added for a multiple text string data type doesn't work (as it attempts to validate the JSON for the full set of text strings).

In this PR I've added a custom `FormatValidator` for this property editor, that will validate each string separately, and fail validation if at least one string doesn't match the configured regular expression.

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/MultipleTextStringPropertyEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/MultipleTextStringPropertyEditor.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hi Andy,

Thanks for the work here. Regex? Not working? Are we sure 😂 ? We'll take a good look at it and then let you know if we need anything. 

Emma

michielDB:
@AndyButland , thank you for the fix. I had lost track of this a bit.
For now I have switched off the validation.

nul800sebastiaan:
Thanks very much for completing this big puzzle @AndyButland! 👍 🏅

PR #6678, opened 2019-10-12
V8: Add validation to user groups when creating new users

kjac:
### Prerequisites

### Description

A new user can't be created without at least one group assignment. But there is no clientside validation of this, only serverside validation. And there is no meaningful feedback of what went wrong when user creation fails because of missing group assignments:

create-user-group-validation-before

This PR adds clientside validation to the user group assignment, just as there is clientside validation for the user name and email:

create-user-group-validation


nul800sebastiaan:
Thanks @kjac !

PR #6679, opened 2019-10-12
Allow hash type on GenerateHash and remove obsoleted ToMd5 and ToSHA1 methods

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6620.

### Description
These are all the changes needed to remove the obsoleted `ToMd5()` and `ToSHA1()` internal methods and replace them with a generic `GenerateHash()` method that accepts a hash type (https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.hashalgorithm.create#System_Security_Cryptography_HashAlgorithm_Create_System_String_). See the comments on PR https://github.com/umbraco/Umbraco-CMS/pull/6621.

I've kept the `GenerateHash()` method internal, but maybe this can also be publicly exposed, so everyone can easily create hashes from strings?

### DIFF
--- a/src/Umbraco.Core/MainDom.cs
+++ b/src/Umbraco.Core/MainDom.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
+using System.Security.Cryptography;
 using System.Threading;
 using System.Web.Hosting;
 using Umbraco.Core.Logging;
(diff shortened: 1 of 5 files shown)


ronaldbarendse:
I've thought about introducing constants, but I think that defeats the generic part of this method.

We could do this instead though:

```c#
internal static string GenerateHash<T>(this string str)
   where T : HashAlgorithm
{
   return str.GenerateHash(typeof(T).FullName);
}
```

So you could do `"string to hash".GenerateHash<SHA1>()` and change the generic type to any hash algorithm (from the `System.Security.Cryptography` namespace).

ronaldbarendse:
@dawoe I've implemented the above changes, so the `GenerateHash` can now accept any `HashAlgorithm` as generic type and you don't have to create/use constants 👍

PR #6685, opened 2019-10-14
Ability to change the culture of an existing language

### LINKED ISSUE #6503: Allow changing the language culture
When you've added a Language under settings it's not possible to change the assigned culture, this could be problematic if for example you have built your website with `English (United States)` (which is the default language), but later realising it should have been `English` or `English (United Kingdom).

Right now you'd have to either write a migration script or go to the database and manually changing the culture, then rebuilding the database cache and the memory cache.

It would be nice if this was supported directly in the Backoffice

---
_This item has been added to our backlog [AB#3057](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3057)_

Shazwazza:
Fixes #6503 

## Changes:

* Allow modifying a culture in the UI for an existing lang
* Allow modifying a culture for an existing lang in the controller
* Update lang repo to validate changing to an existing culture + tests
* Warn the user about expensive operation if they choose to change the culture of an existing language
* Change lang refresher to use payloads instead of just IDs
* When a lang is removed or has a culture change then lots of caches need to be evicted or rebuilt: Domains (all caches), IContent, nucache json DB store, nucache in memory, all indexes

## Testing

* Make sure tests pass + code review
* Have a 2-3 langs with variant content + domains assigned that render
* Change a language's properties but NOT the culture - you should not have a warning displayed
* Change a language's culture to one that already exists - you will get a warning, accept it, and then you will get a validation error
* Change a language's culture to one that doesn't exist - you will get a warning, accept it, it should save and return to the list showing the new language name
* Navigate to a variant content item that is under a node with domains assigned, cycle through each culture via the drop down and look at the Info tab/app to verify that all URLs are correct and routable
* Take note of this content ID and go to the Examine dashboard, Internal index, and search it with this ID. Click on the search result to show all of the fields in that document index, you will see fields that have culture suffixes, ensure that all culture suffixes are correct with the new culture suffix that you changed your lang too.


src/Umbraco.Web/Editors/LanguageController.cs
```diff
@@ -99,24 +99,28 @@ public Language SaveLanguage(Language language)
...
+            var existingByCulture = Services.LocalizationService.GetLanguageByIsoCode(language.IsoCode);
 
             // the localization service might return the generic language even when queried for specific ones (e.g. "da" when queried for "da-DK")
             // - we need to handle that explicitly
-            if (existing?.IsoCode != language.IsoCode)
+            if (existingByCulture?.IsoCode != language.IsoCode)
             {
-                existing = null;
+                existingByCulture = null;
             }
```
bergmania: This is IMO not the controller's responsibility. 

I think we should make an overload/extension method on the service that only returns the language, if the IsoCode matches 100% (Basically executing the code you do here)

src/Umbraco.Web/Editors/LanguageController.cs
```diff
@@ -99,24 +99,28 @@ public Language SaveLanguage(Language language)
...
-            if (existing != null && language.Id != existing.Id)
+            if (existingByCulture != null && language.Id != existingByCulture.Id)
             {
                 //someone is trying to create a language that already exist
                 ModelState.AddModelError("IsoCode", "The language " + language.IsoCode + " already exists");
                 throw new HttpResponseException(Request.CreateValidationErrorResponse(ModelState));
             }
 
-            if (existing == null)
+            var existingById = language.Id != default ? Services.LocalizationService.GetLanguageById(language.Id) : null;
```
bergmania: What are the reason to use `default` and not `null` ?

bergmania: Ahh I see it is `0`

src/Umbraco.Web/Editors/LanguageController.cs
Shazwazza: I didn't change this code, just the var names, I don't exactly know what this is for so I'd prefer to not have to change this in this PR.

src/Umbraco.Web/Editors/LanguageController.cs
Shazwazza: We do use a viewmodel, if it's a brand new language, then there is no ID, and therefore it is `0` because that is the default value of int.

bergmania: 👍 
I know, I just think it is a code smell, to use an `int` and not `int?` when the property also have to represent the absence of a value. But that's not part of this PR, more a general issue.

bergmania:
All tests parse.. Seems to work as expected..

I tested with up to 1500 content items that should have updated the language. Take a few minutes time, but the endpoint do not timeout 👍 

@rasmusjp, will you ensure the changes fit your needs. If so then merge?

rasmusjp:
Looks good, only thing I've noticed is if you're changing the language you are currently editing `?mculture` the parameter is not updated, resulting in the selected language in the content tree is still showing the old language. Not sure if this causes any problems tho

Shazwazza:
@rasmusjp that is a general issue with changing langs, it is starting to be addressed in https://github.com/umbraco/Umbraco-CMS/issues/6663 - which might fix this issue with changing the culture too, but we'll have to wait and see once this PR and that one is merged in.

PR #6686, opened 2019-10-14
V8: Fix sorting of the immediate children of editor start nodes

### LINKED ISSUE #6682: Editor unable to sort nodes in content tree
A content editor is unable to sort the order of the nodes directly under the content start node in the content tree. 

Reproduction
------------

### Bug summary

We have several user groups where the editors only have access to one page and its subpages in the content tree. When an editor tries to sort the level directly underneath the content start node (right clicking on the content start node and choosing sort) Umbraco gives an error message.  
For example, we have a user group called ”Kommun och Politik” with access to content and media, content start node is a page with the same name as the group (Kommun och Politik) and all the default permissions are set to true. In the content tree, when an editor right clicks on ”Kommun och Politik” and chooses sort, Umbraco will give an error message when the editor clicks ”save”. However, if the editor right clicks on a subpage to Kommun och Politik and chooses sort, that’ll workout fine. 
The error appears when trying to change sort order on the content start node. It doesn’t matter if the user clicks on ”Name” to change the sort order or tries to change sort order for only one of the nodes, the error message still appears. I also tried to recreate the issue on an Umbraco 7.15.2 installation but there the sorting worked as expected which make be believe this is bug in Umbraco 8. 

### Specifics

Umbraco version 8.1.3
Chrome 77.0.3865.120
Also checked in Firefox 69.0.1

Img1
Img2
Img4
Img3

### Steps to reproduce

- Create a new user group
 - Give the group access to content and media
 - Set content start node to a specific page
 - Set media start node to root
 - Set all default permissions to true
 - Add a user to the group
 - Make sure the user isn’t a member of any other groups
 - Sign in with the new user
 - Right click on the content start node and click ”Sort”
 - Change the sort order in any way
 - Click ”Save”
 - The error message appears

### Expected result
That the pages are sorted in the order the editor chooses

### Actual result
The editor is presented with an error message:

Authorization error: Unauthorized access to URL:
/umbraco/backoffice/UmbracoApi/Content/PostSort
Contact your administrator for information.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #6682

### Description

See #6682 for a detailed description; the gist of it is as follows:

When an editor has a start node defined, she can't sort the immediate children of that start node despite having permissions to do so:

sort-start-node-content-before

Same problem exists in the media library, although the error is less gracefully handled: The editor appears to simply be logged out (although she really isn't, it just looks like it):

sort-start-node-media-before

It all boils down to a faulty perception of which node is the "parent node" for the sorting operation. The JS incorrectly assumes it's the parent node of the start node, when in reality it is the start node itself. This PR corrects that assumption.

When this PR is applied, sorting works for start nodes in both the content and media sections:

sort-start-node-content-after

sort-start-node-media-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/content.sort.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/content.sort.controller.js
@@ -4,7 +4,6 @@
     function ContentSortController($scope, $filter, $routeParams, contentResource, navigationService) {
 
         var vm = this;
-        var parentId = $scope.currentNode.parentId ? $scope.currentNode.parentId : "-1";
         var id = $scope.currentNode.id;
 
         vm.loading = false;
@@ -42,7 +41,7 @@
             vm.saveButtonState = "busy";
             
             var args = {
-                parentId: parentId,
+                parentId: id,
                 sortedIds: _.map(vm.children, function(child){ return child.id; })
             };
 
--- a/src/Umbraco.Web.UI.Client/src/views/media/media.sort.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/media/media.sort.controller.js
@@ -4,7 +4,6 @@
     function MediaSortController($scope, $filter, mediaResource, navigationService) {
 
         var vm = this;
-        var parentId = $scope.currentNode.parentId ? $scope.currentNode.parentId : "-1";
         var id = $scope.currentNode.id;
 
         vm.loading = false;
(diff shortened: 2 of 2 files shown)


emmaburstow:
Thanks for fixing this @kjac - it's an irritating bug. We'll let you know if we need anything more from you but for now, this all looks great.

Em

nul800sebastiaan:
Lovely, thanks very much! 👍

PR #6687, opened 2019-10-14
V8: Prevent/remove duplicate nodes for editors with start nodes

kjac:
### Prerequisites

### Description

When an editor has start nodes, there's something really weird going on in the content tree. The start nodes are loaded as you'd expect, but the entire tree above the start nodes is also loaded and inserted below the start nodes.

Apparently this is by design to provide the editors with some context as to where they're editing. To me it seems like an annoyance than a genuine help. Besides having all nodes duplicated in the tree, the behavior is also very strange once you start editing:

1. If you select a node in the tree that is built from your start node, the corresponding node underneath the entire tree is highlighted instead of the one you selected.
2. If you right click a node anywhere, both that node and the corresponding duplicate is highlighted.
3. All nodes in the "context" tree have a "not-allowed" cursor and are dimmed down - even the ones you can actually edit.

While none of these are really wrong (except perhaps item 3), it just seems like a strange, confusing editor experience:

start-nodes-remove-duplicates-before

This PR removes the additional "context" tree, leaving behind only the start nodes. While the editor obviously looses the context, I believe the editing experience less confusing this way:

start-nodes-remove-duplicates-after

Clearly this PR is mostly a matter of opinion. I wonder if @nielslyngsoe has some input here?

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/tree.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/tree.service.js
@@ -810,8 +810,17 @@ function treeService($q, treeResource, iconHelper, notificationsService, eventsS
             var node = args.node;
 
             var doSync = function () {
-                //check if it exists in the already loaded children
-                var child = self.getChildNode(node, args.path[currPathIndex]);
+                // dig through the rest of the path to find the applicable child (the tree path may not be complete for users with start nodes)
+                var child = null;
+                _.find(_.rest(args.path, currPathIndex), function (id) {
+                    child = self.getChildNode(node, id);
+                    if (child != null) {
+                        // fast forward the current index to the index of the found child
+                        currPathIndex = _.indexOf(args.path, id);
+                        return true;
+                    }
+                    return false;
+                });
                 if (child) {
                     if (args.path.length === (currPathIndex + 1)) {
                         //woot! synced the node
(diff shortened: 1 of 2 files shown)


emmaburstow:
Hey Kenn, 

Weird bug! Thanks for finding and fixing. We'll be in touch soon,

Emma

poornimanayar:
Hey @kjac,

I can see that this works. But as you rightly said, it is a matter of opinion. So I will leave it to @nielslyngsoe for any input here :-)

Poornima

nielslyngsoe:
Hi @kjac

Definitely a bad experience. I did a little research to figure out the intension of this feature. And the conclusion is that we would need the context. A reason 
for this is visible in this example from a v7 site:

image (2)
Here it would be quite confusing if we started mixing products and blog posts, so we should re-create the experience to fit this example. Always showing the context :-)

I do understand that in some situations it could make sense to hide the context, but we should aim for the original experience in regards to this. And if we see the need for hiding the context, we could make that as a new feature.

kjac:
@nielslyngsoe Ah... so I guess the actual problem is that the user start nodes are pulled out as separate entries above the "context" tree. In reality only the "context" tree should be there, and the editors should drill down through that. I'll have a look at mending the PR accordingly.

nielslyngsoe:
@kjac Cool, thanks!

kjac:
@nielslyngsoe I have created a new PR in #6815. I'll close this one in favor of #6815.

PR #6688, opened 2019-10-15
Long running operations causes SQL timeout

Shazwazza:
Moves read/write locking and isolation level to the responsibility of the sql provider based on the db type and changes Sql server to use `ReadCommitted` for normal SQL calls and `RepeatableRead` for when working with distributed locks.

This is a trimmed down version of this PR https://github.com/umbraco/Umbraco-CMS/pull/6029

## Changes

* Adds methods to ISqlSyntaxProvider for dealing with read/write locks and for returning the provider's own default isolation mode
* Changes SQL Server default isolation mode to `ReadCommitted` and then ensures we only use `RepeatableRead` when creating/reading the distributed locks using the ReadLock/WriteLock methods. This means that long running operations/transactions are not going to block all reads from the database which have resulted in timeout errors
* No changes to SQLCE - this will continue to use `RepeatableRead` since it does not support isolation mode per query

## Testing

* Ensures all tests pass
* Code review
* Test Umbraco with SQLCE and SQL Server - only needs some simple tests, installing with starter kit, creating content types and content should be all that is needed


src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
```diff
@@ -246,6 +249,41 @@ public override bool DoesTableExist(IDatabase db, string tableName)
...
+
+            if (db.Transaction.IsolationLevel < IsolationLevel.ReadCommitted)
+                throw new InvalidOperationException("A transaction with minimum ReadCommitted isolation level is required.");
+
+
+            // *not* using a unique 'WHERE IN' query here because the *order* of lockIds is important to avoid deadlocks
+            foreach (var lockId in lockIds)
+            {
+                db.Execute(@"SET LOCK_TIMEOUT 1800;");
+                var i = db.Execute(@"UPDATE umbracoLock WITH (REPEATABLEREAD) SET value = value*-1 WHERE id=@id", new { id = lockId });
```
zpqrtbnk: if by accident the value has been set to zero this does not lock - hence the original CASE

bergmania: @zpqrtbnk, is there ways this can happen other than people manipulating the database them self? I wonder why we handle this special case?

zpqrtbnk: no, it was "assume garbage in" but not critical - you decide

nul800sebastiaan:
Will this help for https://github.com/umbraco/Umbraco-CMS/issues/6735 as well?

bergmania:
@nul800sebastiaan probably not.. This fixes unintended locking.. Slow queries are still slow.

PR #6692, opened 2019-10-15
V8: Move misplaced colon in language selector

kjac:
### Prerequisites

### Description

As an unintended side effect of #6538, preceding colons have appeared in the language selector:

image

The colons are intended for screen readers, so this PR moves these colons inside the `sr-only` block, and thus the language selector looks normal again:

image


nul800sebastiaan:
Ha! I just fixed this one before I saw this.. https://github.com/umbraco/Umbraco-CMS/commit/4e345a8f796370b78c7c01246dd2da74342ce6a0

This fix seems even better, so I'll merge it on top of that. Thanks @kjac!

PR #6693, opened 2019-10-15
Properly set / unset Culture flag when "vary by Culture" is changed

### LINKED ISSUE #6691: "Allow varying by culture" should set/unset Culture flag only
> Requesting the `category/segments` label on this 

Currently, toggling the "Allow varying by culture" setting on a Content Type or Property Type will set its `Variations` property to either `ContentVariation.Nothing` or `ContentVariation.Culture`. 

However, this disregards the `ContentVariation.Segment` flag that can also be set and always _removes_ it. While segments are not currently supported completely in Umbraco, we are working on it through PRs (such as [this one](https://github.com/umbraco/Umbraco-CMS/pull/6209) which is merged in v8/dev). 

The correct behavior of toggling "Allow varying by culture" would be to set or unset the `ContentVariation.Culture` flag. This change maintains the current behavior when segments are not involved (it switches between `Nothing` and `Culture`), but will also work properly when the `Segment` flag is already set via code on the ContentType / PropertyType.

__I have a PR ready which I believe fixes the issue, I will link it soon.__

PerplexDaniel:
### Prerequisites

Fixes #6691 

### Description
Toggling the "Allow varying by culture" setting on a content type or property type now only touches the `ContentVariation.Culture` flag, rather than directly setting either `ContentVariation.Nothing` or `ContentVariation.Culture`. This means the `ContentVariation.Segment` flag is always cleared which is not correct.

### How to test
This fix should not break any existing functionality (it will still switch between `Nothing` and `Culture` until segments are available in Umbraco), so the test would be to run whatever existing tests are available to test the "Allow varying by culture" behavior. It should be the same.

---
_This item has been added to our backlog [AB#3855](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3855)_


emmaburstow:
Hey @PerplexDaniel 

Thanks for the work here and also thanks for the testing steps. We'll take a look and let you know if we have any questions.

Thanks,

Emma

poornimanayar:
Hi @PerplexDaniel ,

I have tested this around various languages and content editing. All looks fine! Is there anything more specific to test?

Poornima

PerplexDaniel:
@poornimanayar Thanks for testing :)

>  Is there anything more specific to test?

Not that I could think of. This change should not have any effect in the currently released version of Umbraco, it is simply more correct and necessary in the near future when Segments become available for use (currently, only the Culture variation is completely supported).

To provide some context: we are working (together with Umbraco HQ) on getting Segments into the core at some point (see other PRs [here](https://github.com/umbraco/Umbraco-CMS/pull/6208) and [here](https://github.com/umbraco/Umbraco-CMS/pull/6209)) and this was a small change we also need to make sure the "Segment"-flag would not be cleared when saving a content type / property type from the UI (Settings > Document Types). As I described, the current code only switches the enum value between `Nothing` and `Culture`, but it ignores (and always removes) the `Segment` flag that can already be there. This PR fixes that.

src/Umbraco.Web/Models/Mapping/ContentTypeMapDefinition.cs
```diff
@@ -223,7 +223,11 @@ private static void Map(PropertyTypeBasic source, PropertyType target, MapperCon
             target.DataTypeKey = source.DataTypeKey;
             target.Mandatory = source.Validation.Mandatory;
             target.ValidationRegExp = source.Validation.Pattern;
-            target.Variations = source.AllowCultureVariant ? ContentVariation.Culture : ContentVariation.Nothing;
+            target.Variations = source.AllowCultureVariant
+                // Set culture flag
+                ? target.Variations | ContentVariation.Culture
+                // Unset culture flag
+                : target.Variations & ~ContentVariation.Culture;
```
nul800sebastiaan: I know this looks beautiful but it is extremely unreadable and I personally am not familiar with this notation, so I have no idea what is going on. For readability can we please turn this into multiple `if`s and whatever the `~` means? :-)

src/Umbraco.Web/Models/Mapping/ContentTypeMapDefinition.cs
```diff
@@ -395,9 +399,11 @@ private static void Map(MemberPropertyTypeBasic source, MemberPropertyTypeDispla
...
             if (!(target is IMemberType))
             {
-                target.Variations = ContentVariation.Nothing;
-                if (source.AllowCultureVariant)
-                    target.Variations |= ContentVariation.Culture;
+                target.Variations = source.AllowCultureVariant
+                    // Set culture flag
+                    ? target.Variations | ContentVariation.Culture
+                    // Unset culture flag
+                    : target.Variations & ~ContentVariation.Culture;
```
nul800sebastiaan: This is repeated from above, can we turn this into a small method instead so when we need to make changes, we make them the same everywhere please?

src/Umbraco.Web/Models/Mapping/ContentTypeMapDefinition.cs
PerplexDaniel: Thanks for the review. The original code also used a ternary `if`-statement which is why I kept it that way. But will change soon then.

Regarding the `~` operator; this is the [bitwise complement operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#bitwise-complement-operator-), so it inverts all the bits (e.g. `1100` would become `0011`). So by performing an AND (`&`) with the complement of `Culture` we remove the `Culture` flag.

Using bitwise operations is the normal way of manipulating flags enum types, like `target.Variations`. This is also explained by [Microsoft here](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/enumeration-types#enumeration-types-as-bit-flags). 

I would advice against removing these operators just because bitwise operators are not very well known. 

With respect to readability it could however be nice to add an extension method like `UnsetFlag()` in the [`EnumExtensions`](https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Core/EnumExtensions.cs) that you guys already have.

nul800sebastiaan: Thanks @PerplexDaniel! 
Cool, thanks for the insight!

I agree, that the `bitwise complement operator` is absolutely fine then, make sure to leave a code comment. 

And yes, I really like the idea of adding an extension method, with a description of what it does and why.

I love ternary operators, but it gets a bit unreadable when they seemingly start to do more than one thing in the if/else clauses. Clearly my lack of knowledge on bitwise operators led me to believe there was multiple nested if/else statements here. So they're still fine to use! 

If you could refactor this into an extension method with some method documentation then this should be good to go! 👍

src/Umbraco.Web/Models/Mapping/ContentTypeMapDefinition.cs
PerplexDaniel: It indeed looks identical but the involved types are not actually the same. The type of `target` is either `IContentTypeComposition` or `PropertyType` which do not seem to share any class or interface. The duplicated code just sets/unsets the `ContentVariation.Culture` based on the `AllowCultureVariant`, which originally was also duplicated of course. But I will check if it can be improved with a method.

src/Umbraco.Web/Models/Mapping/ContentTypeMapDefinition.cs
```diff
@@ -224,11 +224,9 @@ private static void Map(PropertyTypeBasic source, PropertyType target, MapperCon
...
             target.ValidationRegExp = source.Validation.Pattern;
             target.Variations = source.AllowCultureVariant
-                // Set culture flag
-                ? target.Variations | ContentVariation.Culture
-                // Unset culture flag
-                : target.Variations & ~ContentVariation.Culture;
-
+                ? target.Variations.SetFlag(ContentVariation.Culture)
+                : target.Variations.UnsetFlag(ContentVariation.Culture);
+
```
PerplexDaniel: I added the extension methods `SetFlag` and `UnsetFlag` to make it more readable, and also added unit tests for those extensions. Hope that's OK.

JeffreyPerplex:
@nul800sebastiaan Any update on this :)?

Shazwazza:
HI all, I'm scheduled to re-review this this sprint, are you able to fix up the merge conflicts for me though?

PerplexDaniel:
@Shazwazza Just merged v8/dev and resolved the conflict

Shazwazza:
Great stuff!

nul800sebastiaan:
Hi there @PerplexDaniel,

Kind regards,

ronaldbarendse:
Great fix, although I don't understand why the `SetFlag` and `UnsetFlag` extension methods are added!

Adding or removing an enum flag can be done very easily:
```
target.Variations |= ContentVariation.Culture;
target.Variations &= ~ContentVariation.Culture;
```

I can only see *some* added value for a `SetFlag` extension method that accepts a boolean, like shown in this discussion: https://stackoverflow.com/questions/5850873/enum-hasflag-why-no-enum-setflag.

The performance overhead of these extension methods shouldn't be a problem in this case, but because they're available, they might get (unknowingly/mistakenly) used in performance critical code.

PerplexDaniel:
@ronaldbarendse I agree, [my original commit](https://github.com/umbraco/Umbraco-CMS/pull/6693/commits/1a92dbfd8fee30689822a09a57a2365d40c1bc4e) did not include any extension methods, but Sebastiaan thought the syntax might not be very well known thus the extension methods were added (check the conversation around Oct 29).

ronaldbarendse:
Hmm, I see... Well, I would argue the custom extension methods will be even more unknown compared to the bitwise operators 😉 If you don't understand this basic concept, you're probably also not the person that should touch this part of Umbraco anyway 🙄

Might be good to performance test this before it's released (and removing would mean a breaking change)! Otherwise, at least change it to a single `SetFlag<T>(T flag, bool value)` extension method or make these methods internal...

nul800sebastiaan:
Thanks @ronaldbarendse, I am the person to suggest this since I was until then utterly unaware of bitwise operators! 🎉 

And this is why I let @PerplexDaniel, a much smarter person than me, touch that part of Umbraco. 

I am unashamed about my ignorance and your comments don't bother me, but please don't presume to know how smart other contributors are, that is not part of our friendly ethos. Thanks.

PerplexDaniel:
I cannot imagine this has any significant performance impact, sure there's overhead with an extension method but this is probably in the order of micro or nano seconds. You are welcome to run some tests of course :) 

The existing enum extension methods were all public which is why the `SetFlag` / `UnsetFlag` methods were made public as well. 

Anyway, this pull request is already merged and the issue is closed so if you want to change something I think it's best to submit another PR and discuss there with HQ.

ronaldbarendse:
@nul800sebastiaan I totally understand your reaction, as the extension methods were added after your advice and my comment might seem directed to you, but that was certainly not the case 😉 I'm glad it didn't bother you personally anyway. It's also not about how smart someone is, but more about understanding concepts. Going against others recommendations and actions (as it's already merged) might inherently be a little bit unfriendly, so allow me to make my case below...

Not everyone will be familiar with bitwise operators and that's fine, but adding more code doesn't necessarily make it simpler to understand. It will however add:
- More code to maintain
- A public API (that's specific to Umbraco)
- Documentation for the code
- Tests to verify the written code
- Performance problems (down the road, if it gets used in the wrong place)

@PerplexDaniel Yes, it won't be noticeable in this case, as it only toggles a single flag. But the performance difference between the bitwise operators and using the extension methods is significant! And most of that is actually because of the overhead (method call, casting, converting, additional variables).

I've whipped up a Benchmark.NET project and there's a **BIG** difference, more than 10.000 times slower. It actually couldn't correctly calculate the ratios and showed a warning for the bitwise tests: _The method duration is indistinguishable from the empty method duration_.

|                 Method |        Mean |     Error |    StdDev |      Median |
|----------------------- |------------:|----------:|----------:|------------:|
|                SetFlag | 200.7955 ns | 1.3842 ns | 1.2271 ns | 200.6699 ns |
|        SetFlagExisting | 200.6694 ns | 1.6062 ns | 1.4239 ns | 200.4420 ns |
|         SetFlagBitwise |   0.0143 ns | 0.0070 ns | 0.0058 ns |   0.0161 ns |
| SetFlagBitwiseExisting |   0.0049 ns | 0.0073 ns | 0.0068 ns |   0.0019 ns |

Even though Microsoft added the `HasFlag` extension method and that probably got some performance reviews before it got shipped, it's still considerably slower (again almost 10.000 times):

|              Method |       Mean |     Error |    StdDev |     Median |
|-------------------- |-----------:|----------:|----------:|-----------:|
|         HasFlag | 17.2275 ns | 0.1370 ns | 0.1214 ns | 17.2089 ns |
|  HasFlagBitwise |  0.0023 ns | 0.0042 ns | 0.0039 ns |  0.0000 ns |

A single line comment would in this case be more than sufficient to make the code understandable:

```c#
if (source.AllowCultureVariant)
{
    target.Variations |= ContentVariation.Culture; // Set flag using bitwise logical OR
}
else
{
    target.Variations &= ~ContentVariation.Culture; // Remove flag using bitwise logical AND and complement (reversing the bit)
}
```

PerplexDaniel:
@ronaldbarendse I appreciate the tests you have done, but keep in mind we are talking about nanoseconds here, 1 ns == 1/1,000,000,000 of a second. So while the difference is certainly significant, the end result is still in the order of nanoseconds. The readability argument would certainly win over a performance gain of nanoseconds, even though I agree people should probably be familiar with the bitwise operations but the reality is they are unknown and not used much in the web development field.

Anyway, of course the decision is not mine so perhaps just open a new issue and let Umbraco decide on this?

nul800sebastiaan:
This is certainly a "pick your battles" issue. I this has an actual significant impact that is not on the order of nanoseconds it's worth looking at. 

Your other arguments do make sense and we accept the added "work", which is on the order of 5 lines of code. Thanks!

ronaldbarendse:
Besides the increased speed, using bitwise operators also doesn't require a memory allocation (because of the boxing/unboxing).

Performance differences aside, we shouldn't add unneeded code to Umbraco (especially extension methods within the quite common `Umbraco.Core` namespace), as they might result in other problems when unknowingly used in the wrong place...

I've already made the PR to change back to bitwise operators (like @PerplexDaniel initially used): https://github.com/umbraco/Umbraco-CMS/pull/7430. Maybe I just like to pick these battles, but IMHO we should all strive to keeping the Umbraco code-base clean and performing: that includes these little additions/changes 😉

Shazwazza:
I would like to keep the ext methods, it helps for readability and we can make it have no impact on performance. It is unfortunate that there's no good generic way to deal with setting flags for Enums without boxing and mem allocations. People have gone to great lengths to attempt this, here's a great SO thread https://stackoverflow.com/questions/1189144/c-sharp-non-boxing-conversion-of-generic-enum-to-int. (NOTE: the benchmarks don't show mem allocations due to this)

IMO for Enums that we want to use these ext methods on we can just make non-generic versions of them which will have no perf impact at all, for example:

```
public static ContentVariation SetFlag(this ContentVariation input, ContentVariation flag) => input | flag;
```

nul800sebastiaan:
@ronaldbarendse As per Shannon's feedback I will close your PR. Keep in mind that Umbraco is not exactly a pure architect's dream, we try to provide helpful extension points for anybody working with the code as well. Yes that sometimes come with a big overhead and sometimes with nano-overhead. It's something we think is often worth it.

JimBobSquarePants:
Best of both worlds...

``` c#
[MethodImpl(MethodImplOptions.AggressiveInlining)]
public static T SetFlag<T>(this T input, T flag)
    where T : unmanaged, Enum
{
    // Bitwise OR to set flag f of enum i. 
    // We avoid allocations by using Unsafe over Convert.
    // Branching should be removed by the Jitter.
    if (Unsafe.SizeOf<T>() == 1)
    {
        byte r = (byte)(Unsafe.As<T, byte>(ref input) | Unsafe.As<T, byte>(ref flag));
        return Unsafe.As<byte, T>(ref r);
    }
    else if (Unsafe.SizeOf<T>() == 2)
    {
        short r = (short)(Unsafe.As<T, short>(ref input) | Unsafe.As<T, short>(ref flag));
        return Unsafe.As<short, T>(ref r);
    }
    else if (Unsafe.SizeOf<T>() == 4)
    {
        int r = Unsafe.As<T, int>(ref input) | Unsafe.As<T, int>(ref flag);
        return Unsafe.As<int, T>(ref r);
    }
    else
    {
        // size = 8
        long r = Unsafe.As<T, long>(ref input) | Unsafe.As<T, long>(ref flag);
        return Unsafe.As<long, T>(ref r);
    }
}

[MethodImpl(MethodImplOptions.AggressiveInlining)]
public static T UnsetFlag<T>(this T input, T flag)
        where T : unmanaged, Enum
{
    // Bitwise AND combined with bitwise complement to unset flag f of enum i.
    // We avoid allocations by using Unsafe over Convert.
    // Branching should be removed by the Jitter.
    if (Unsafe.SizeOf<T>() == 1)
    {
        byte r = (byte)(Unsafe.As<T, byte>(ref input) & ~Unsafe.As<T, byte>(ref flag));
        return Unsafe.As<byte, T>(ref r);
    }
    else if (Unsafe.SizeOf<T>() == 2)
    {
        short r = (short)(Unsafe.As<T, short>(ref input) & ~Unsafe.As<T, short>(ref flag));
        return Unsafe.As<short, T>(ref r);
    }
    else if (Unsafe.SizeOf<T>() == 4)
    {
        int r = Unsafe.As<T, int>(ref input) & ~Unsafe.As<T, int>(ref flag);
        return Unsafe.As<int, T>(ref r);
    }
    else
    {
        // size = 8
        long r = Unsafe.As<T, long>(ref input) & ~Unsafe.As<T, long>(ref flag);
        return Unsafe.As<long, T>(ref r);
    }
}
```

Shazwazza:
Oooh la la! 🎉 I've not seen that before, looks fancy though :) @ronaldbarendse any chance you can give that a spin?

ronaldbarendse:
@JimBobSquarePants Yes, I've seen that type of code in this .NET Core issue: https://github.com/dotnet/corefx/issues/693#issuecomment-557362093. This will require compiling `Umbraco.Core` with the `unsafe` flag set, which might not be desired.

These extension methods are also available in the [Enums.NET](https://github.com/TylerBrinkley/Enums.NET) library, with code and performance tests, but adding another dependency might also not be desired.

IMHO Umbraco should not include these type of generic extension methods, as they may clash with other libraries (requiring namespace aliases), pollute all public APIs with methods only available in Umbraco and not easily re-used in other projects, etc.

If readability was such an important requirement, these flags should just be booleans on the types, optionally storing the actual value as a flags enum... I think the solution in this Shannons comment https://github.com/umbraco/Umbraco-CMS/pull/6693#issuecomment-571807022 is the best: have extension methods on the specific enum types. The extensions class already exists:
https://github.com/umbraco/Umbraco-CMS/blob/3a1078824d44a082eefe0dc1b4e6f9ad73e362dd/src/Umbraco.Core/ContentVariationExtensions.cs#L108-L128

ronaldbarendse:
I've updated my branch with these changes: https://github.com/umbraco/Umbraco-CMS/compare/v8/dev...ronaldbarendse:v8/bugfix/enum-extensions. As PR #7430 is already closed, it will need to be reopened to see the changes/merge them in...

JimBobSquarePants:
> This will require compiling Umbraco.Core with the unsafe flag set, which might not be desired.

You're confusing `unsafe` with `Unsafe<T>`.

I totally agree regarding Enums.NET btw. Very well put together and tested and supports all the target frameworks Umbraco would ever need. That would certainly be my preference going forward.

ronaldbarendse:
@JimBobSquarePants You're right (I wouldn't expect otherwise from you 😉), the example I linked used the `unsafe` keyword in the method declaration and that's indeed not required... Still, I don't think Umbraco should ship these extension methods, as they're too generic.

PR #6696, opened 2019-10-15
V8: Use member type icon in member pickers

kjac:
### Prerequisites

### Description

The member pickers (both the dedicated one and MNTP) do not show the correct member type icons for members of different types. 

In this case I have the members "Hello Mister" and "Hello to you too" of type "Test 1234" (so imaginative) which uses a different icon than the default one:

image

...yet in the member pickers, all members look the same:

image

Indeed had I changed the icon for the default member type, the pickers would still show the default icon because it is hard coded 😞 

This PR updates the mapping of the member entities so their member type icons are used in pickers:

image


emmaburstow:
Lovely work! Thanks for doing your bit to make the back office more consistent (and colourful)!

Em

nul800sebastiaan:
Looking great, cheers!

PR #6697, opened 2019-10-15
V8: Use the member type icon in the Members section

kjac:
### Prerequisites

### Description

In the Users section we display the user groups with their respective group icons:

image

But in the Members section we do not; in fact the icon displayed gives next to no value other than to look consistent with the rest of the trees:

image

This PR ensures that we use the member type icon in the Members section:

image


emmaburstow:
Also brilliant. Thanks Kenn.

Em

nul800sebastiaan:
Perfect!

PR #6704, opened 2019-10-15
Add "umb-auto-focus" on media and link pickers

BatJan:
### Prerequisites

### Description
When I worked on #4526 this weekend I discovered that some of the editors that open don't have focus set on the first focusable item.

I have fixed it for the "media picker" and the "link picker" with this PR. If the first element does not have focus then the focus lock will not work as expected 😃


nul800sebastiaan:
Lovely, thank you kindly @BatJan - that was bugging me as well! 👍

PR #6707, opened 2019-10-15
Media picker: Fix semantics, missing dictionary items and hide icons non-visually

BatJan:
### Prerequisites

### Description
When dealing with a previous PR I noticed that some empty `<a>` elements are being used, which is not ideal. One of them had `role="button"` set but I have changed the element to be a real `<button>` instead and moved the text from the `aria-label` into a `.sr-only` span since that makes the translation a bit easier to deal with. Some restyling after changing the `<a>` elements to `<button>` elements overall.

This also fixes an issue where the label did not get announced by NVDA at least.

**Before**
media-picker-before

**After**
media-picker-after


poornimanayar:
Hey @BatJan ,

Testing this now, will let you know how it goes :-)

Poornima

nul800sebastiaan:
Woohoo, this is excellent, thanks @BatJan! 👍

PR #6708, opened 2019-10-15
Don't hide content in umb-editors while animating

fALKENdk:
Right now content inside `umb-editors` first appear when the `umb-editor` is done animating.
With this PR, content is shown while the `umb-editor` is animating.

Testet in Firefox 69, Chrome 77, Safari 13 on macOS.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editors.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editors.html
@@ -18,7 +18,7 @@
         'umb-editor--level3': model.level === 3}"
         >
         <div ng-if="!model.view && !model.animating" ng-transclude></div>
-        <div ng-if="model.view && !model.animating" ng-include="model.view"></div>
+        <div ng-if="model.view" ng-include="model.view"></div>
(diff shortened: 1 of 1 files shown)


clausjensen:
Well.
It sort of works better.
There's some .. quirks though.

When adding a property and picking an editor - this works a whole lot better. Very snappy and I can type instantly without having to wait. 🎉 

However - when adding a property, it is like this slide-in doesn't really animate correctly anymore. From a first glance - it seems to just snap in place with no animation, instantly. But if you then look closer, it actually does this - and then the slide-in kinda shakes for a bit like it is trying to animate something not really right.

Another issues is also that when I go from the "Add Property" slide-in to the "Add Editor" slide-in, as soon as the "Add Editor" slide-in shows, the previous one (in the back) is displayed 100% to the left of the current one (in front) - and then animates to the right and hides behind the current one.
Actually this affects a few more of the places where a slide-in is supposed to cover a previous slide-in ... so more of a general problem.

nul800sebastiaan:
Thanks @fALKENdk - I have these same issues that @clausjensen mentions, I think this needs to be rethought. The animations are quite weird.. I'll close this one for now.

PR #6715, opened 2019-10-16
V8: Fix RTE full screen mode

### LINKED ISSUE #6701: 8.2: Full screen options in RTE doesn't work correctly
I've added the "Full screen" option to an RTE.
When I try to use it in the content section, it doesn't go full screen, but instead gives me this:

RTE-fullscreen

Reproduction
------------

### Bug summary

The full screen option of RTE's doesn't work properly.
It worked fine in Umbraco 8.1.5

### Specifics

* Umbraco 8.2
* Google Chrome, 77.0.3865.120
* Checked Edge (44.18362.387.0) - doesn't work there either

### Steps to reproduce

* Create a Document Type
* Add an RTE to the Document Type
* Make sure that the "Full screen" option is enabled on the RTE
* Create a content node
* Click the full screen option in the RTE

### Expected result

I expected that the editor would go into full screen mode.

### Actual result

RTE-fullscreen

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #6701

### Description

The recent RTE styling updates unfortunately broke the full screen mode:

rte-fullscreen-before

This PR fixes it - when applied, RTE full screen mode works like it did in 8.1.5:

rte-fullscreen-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/rte.less
+++ b/src/Umbraco.Web.UI.Client/src/less/rte.less
@@ -83,10 +83,6 @@
     }
 }
 
-.umb-rte .mce-fullscreen {
-    position:absolute;
-}
-
 .umb-rte .mce-toolbar .mce-btn-group {
     padding: 0;
 }
@@ -164,4 +160,8 @@
 .umb-grid .umb-rte {
     border: 1px solid #d8d7d9;
     max-width: none;
+}
+
+.mce-fullscreen {
+    position: absolute;
 }


kjac:
Yeah that app can't really make up it's mind 😛

nul800sebastiaan:
Well at least it fixes the fullscreen button but not the things that Ronald is mentioning, also I can't click on "view source", "insert image", etc. That's a different issue though and I am not sure how to fix it since those open in an infinite editor which is going to look weird in this fullscreen view.. 🤔 

Anyway, for now you can at least type in full screen mode, let's tackle the other problems separately.

PR #6719, opened 2019-10-16
Use umb-checkbox in doctype collection

bjarnef:
### Prerequisites

### Description
When using document type collection to create a Parent and Child document type the selection of whether templates also should be created use traditional checkboxes.

I have replaced these with `umb-checkbox` so it is more consistent with rest of the UI.

I have added some more utility css classes for spacings. We had for `margin-top`, `margin-bottom` and `margin-left` but not for `margin-right`.

I have also added a few to reset padding since I have seens place where inline styling are used to set e.g. `padding-left: 0` where we can use these instead.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/documenttypes/create.html
+++ b/src/Umbraco.Web.UI.Client/src/views/documenttypes/create.html
@@ -28,7 +28,7 @@
                         <i class="large icon icon-thumbnail-list"></i>
                         <span class="menu-label">
                             Document Type Collection...
-                            <!--                        <localize key="content_documentType_collection">Document Type Collection</localize>-->
+                            <!--<localize key="content_documentType_collection">Document Type Collection</localize>-->
                         </span>
                     </button>
                 </li>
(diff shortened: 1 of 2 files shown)


emmaburstow:
Hey @bjarnef 

Lovely change here! We'll let you know if we need anything but all looks good to me.

Emma

src/Umbraco.Web.UI.Client/src/views/documenttypes/create.html
```diff
@@ -81,18 +81,20 @@ <h5><localize key="create_createUnder">Create an item under</localize> {{current
 
                 <umb-control-group label="Name of the Parent Document Type" hide-label="false">
                     <input type="text" name="collectionName" ng-model="model.collectionName" class="umb-textstring textstring input-block-level" umb-auto-focus required />
-                    <span ng-if="model.disableTemplates === false">
-                    <input id="collectionCreateTemplate" name="collectionCreateTemplate" type="checkbox" ng-model="model.collectionCreateTemplate" style="margin-top: 0;" />
-                    <label for="collectionCreateTemplate" style="margin-bottom: 0; padding-left: 2px;">Create template for the Parent Document Type</label>
-                    </span>
+
+                    <div ng-if="model.disableTemplates === false" class="flex">
+                        <umb-checkbox input-id="collectionCreateTemplate" name="collectionCreateTemplate" model="model.collectionCreateTemplate" />
```
BatJan: Hey @bjarnef - The label is not neccessary when you use the umb-checkbox directive. The directive puts in a label for you. All that is really needed is this

```
                   <umb-checkbox
                        ng-if="model.disableTemplates === false"
                        model="model.collectionCreateTemplate"
                        text="Create template for the Parent Document Type"
                    />
```
In this case I have added the `ng-if` check on the directive as well - Currently the checkbox will render with two labels, which is not ideal for accessibility and the `input-id` and `name` attribues are not neccessary 😃

src/Umbraco.Web.UI.Client/src/views/documenttypes/create.html
```diff
@@ -81,18 +81,20 @@ <h5><localize key="create_createUnder">Create an item under</localize> {{current
...
 
                 <umb-control-group label="Name of the Item Document Type" hide-label="false">
                     <input type="text" name="collectionItemName" ng-model="model.collectionItemName" class="umb-textstring textstring input-block-level" required />
-                    <span ng-if="model.disableTemplates === false">
-                    <input id="collectionItemCreateTemplate" name="collectionItemCreateTemplate" type="checkbox" ng-model="model.collectionItemCreateTemplate" style="margin-top: 0;" />
-                    <label for="collectionItemCreateTemplate" style="margin-bottom: 0; padding-left: 2px;">Create template for the Item Document Type</label>
-                    </span>
+
+                    <div ng-if="model.disableTemplates === false" class="flex">
+                        <umb-checkbox input-id="collectionItemCreateTemplate" name="collectionItemCreateTemplate" model="model.collectionItemCreateTemplate" />
```
BatJan: Same as my comment above :-)

abjerner:
@bjarnef I also gave this a test, and the suggestions by @BatJan makes good sense. Could you update your PR to reflect Jan's suggestions?

bjarnef:
@abjerner I have merged latest changes from `v8/dev` branch and removed the wrapping ´<div>` elements.

nul800sebastiaan:
Lovely, thanks for the help here @bjarnef!

PR #6722, opened 2019-10-16
V8: General UX overhaul of the media picker in the RTE

### LINKED ISSUE #6712: V8.2 RTE image dialog doesn't display title
The image dialog in the RTE misses some styling and the title. It feels like it's not a finished dialog.

image

image

### Bug summary
* Modal dialog misses Title
* Not sure if it also misses some styling, like some padding below the title box.
* Missing some labels?

### Steps to reproduce
I have inserted an image in the RTE using the drag and drop functionality. Then I select the image in the RTE and click in the Media Picker button and a dialog opens on the right which is a bit confusing as it doesn't says what it's for and misses some styling I think:

### Expected result
A more clear meaning of what that dialog is for.

### Actual result
Confusing, the first time it opened I didn't know what that was for.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #6712

### Description

The media picker seems to be in need of an overhaul - it has quite a few little quirks going on.

When you pick an image to insert into an RTE, you're presented with a dialog where you can edit the image before inserting it. In this dialog:

1. The header is empty, leaving behind an odd space.
2. A focal point editor is displayed, but the focal point is never used.
3. There's a read-only field with the image URL for no particular reason.
4. If you hit "Close" instead of "Submit" you're brought back to the media picker. This may be an unintended feature, but it's not terribly intuitive.

image

Now if you insert an image and subsequently choose to edit it:

1. The image (focal point editor) is no longer visible.
2. The image URL is not shown - the read-only field just reads "Url".
3. If you hit "Close" instead of "Submit" you're brought back to the media picker... same behavior as above, only here it's just plain weird.

rte-media-picker-before

This PR addresses all the points above. I have opted to add a preview image instead of the focal point editor, and I have removed the read-only URL as it really adds no value.

Here's how things work with this PR applied (apologies in advance for a somewhat huge GIF):

rte-media-picker-after

#### Backwards compatibility 

While I can't find any use cases for it in the current code base, I have left both the focal point editor and the URL field in the media picker: 

- An extra configuration option has been introduced so the media picker consumers can choose to disable the focal point editor and show the preview image instead. For backwards compatibility this option is assumed to be `false`.

- Currently the media picker only conditionally turns the URL field read-only, namely when the edited media item has no ID. Again, I can't find any current uses for this but someone somewhere might rely on that functionality so I left it there; the URL field will be shown (as editable) if the edited media item has no ID.


nul800sebastiaan:
Some quick observations from the couch:

- Maybe the alt text should be above the image? It is the only thing you can interact with in that overlay so makes sense that you see it first
- There's a nice edge case since 8.2.0: drag an image into the RTE and then click on the image button in the toolbar, it hasn't saved the image on disk yet, so the only way to show a preview would be to get the value of the blob and copy it into the overlay panel.. which if it's a huge image might be too much

kjac:
@nul800sebastiaan good pointers.

I was a bit torn about the alt text location myself, but I think you're right. I'll move it.

The edge case you're referring to is very valid. However in this PR it's handled more or less exactly as it is today - which isn't great and which is also on my to-do list to fix (although this PR is a bit in the way in terms of merge conflicts).

kjac:
All done; it now looks like this:

image

nul800sebastiaan:
It's absolutely wonderful, thanks very much @kjac! ⭐️

PR #6725, opened 2019-10-16
Change the style of the confirm button on delete media

poornimanayar:
Changed the style on the Confirm button on delete media to `danger` . We already have this in the Content section on deleting a node. So I think this can make it consistent.

**Before**
image

**After**
image


emmaburstow:
Lovely. We'll take a look soon and let you know if we need you. Consistency is really being improved here.

Em

nul800sebastiaan:
Thank you kindly!

PR #6726, opened 2019-10-16
Change the confirm button style on delete dictionary item

poornimanayar:
The delete button should use the `danger` style across the board. I noticed this while I was doing #6725 I dont know whether this is a good move, hence doing this as a separate PR. Will flow it across the board if this is okay.

**Before**
image

**After**
image


emmaburstow:
So nice to keep seeing your name here. Speak soon,

Em

nul800sebastiaan:
Good one! 👍

PR #6727, opened 2019-10-16
V8: Style the umb-overlay header and footer like the rest of the overlays

kjac:
### Prerequisites

### Description

The header and footer of the `umb-overlay` dialog look rather out of place when the dialog is used in "slide-in-from-right" mode; I think they may be leftovers from V7.

Here's a V8 style dialog:

image

...and here are a few select samples of `umb-overlay`:

image

image

Notice:

- The height of the header and footer (and how they do not align nicely with the main window header and footer)
- The font size of the header
- The border color that separates the dialog content from the header and footer
- The background color of the footer
- The lack of padding in the top of the dialog content 
- The white border near the footer when the dialog has a scrollbar

This PR updates the styling of `umb-overlay` so it looks like a V8 dialog in "slide-in-from-right" mode:

image

image

The styling of `umb-overlay` when used e.g. for picking items types in Grid or Nested Content remains (largely) unchanged:

image

#### Testing this PR

Verify that the `umb-overlay` look and feel is V8-ish 😄 

At the time of writing, `umb-overlay` is used for:

- Edit dialog when inserting or editing images from RTE
- User profile dialog when clicking the user avatar in the upper left corner
- File type picker when uploading new images via the dropzone (provided you have multiple "image" media types)
- Fields list per search result in the Examine Management dashboard 
- New item type picker in the Grid
- New item type picker in Nested Content
- List view YSOD details dialog (I'm not entirely sure how to test this one...)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/overlays.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/overlays.less
@@ -20,11 +20,11 @@
 }
 
 .umb-overlay .umb-overlay-header {
-    border-bottom: 1px solid @purple-l3;
+    border-bottom: 1px solid @gray-9;
     margin-top: 0;
     flex-grow: 0;
     flex-shrink: 0;
-    padding: 20px 30px 0;
+    padding: 20px 20px 0;
 }
 
 .umb-overlay__section-header {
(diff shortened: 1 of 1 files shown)


emmaburstow:
This is just right. Ok, so we'll check it over and merge if it's all good.

Em

PR #6728, opened 2019-10-16
Remove key from the user details screen

poornimanayar:
I don't know whether my thought process is right here but user does not have a key, so it should not show on the details screen.

**Before**
image

**After**
image


emmaburstow:
Ah yes. This makes sense. No point in showing it if there isn't one... We'll let you know if we need anything,

Em

nul800sebastiaan:
Minor detail, the key is predictable and there is a 1 in there somewhere ;-)

`00000001-0000-0000-0000-000000000000` 

We should keep showing it sill, so I'll close this PR, thanks anyway!

PR #6729, opened 2019-10-16
Notifications span full width on user group save or delete

poornimanayar:
Noticed that on user group save or delete the notifications do not span the full width. I have fixed that.

**Before**
image

**After**
User group Save
image

User Group Delete
image


emmaburstow:
Good spot! :D

Em

poornimanayar:
This is an issue on the Create Users/Invite Users form as well.

nul800sebastiaan:
Lovely, thanks Poornima!

PR #6733, opened 2019-10-17
Added apostrophe to Filtrer... in danish

nfJakobsen:
image


nul800sebastiaan:
And for the hattrick! Thanks again! 👍

PR #6736, opened 2019-10-17
V8: Lazy loaded properties can be interacted with when creating content types

kjac:
### Prerequisites

### Description

When creating content types using the keyboard (i.e. tab a lot), some of the property editors that use lazy loading can actually be interacted with - for example the RTE and Nested Content:

content-type-tabindex-before

This PR ensures that we observe changes in the editors marked with the `disable-tabindex` directive and apply the tabindex disabling whenever these editors are done loading. Also the RTE `iframe` has been added explicitly to the whitelist of "tab-able elements".

content-type-tabindex-after

/cc @clausjensen, a.k.a. Sir Tabs-A-Lot 😆


emmaburstow:
👋 🙇‍♀ 🗣 

Em

nul800sebastiaan:
Haha! Wow, I've never tried this, good find!

clausjensen:
Thanks @kjac 😃

Issue #3022, opened 2018-09-24
Media uploader in RTE doesn't select uploaded image


JorickDJ:
This issue was already submitted on 12th of September 2017 and we are still having problems with this issue in Umbraco version 7.10.5.

The previous bug submission can be found here:
https://issues.umbraco.org/issue/U4-10408

When I want to add a new image to the RTE I click on the Image Picker and upload a new image. The Image Picker indicates that the image is selected and when you click "Submit" it adds a "nothing.jpg" to the RTE.

If you upload a new image in the RTE Image Picker and then scroll to the uploaded image and select it again, then it will be added correctly.

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/3117

jannikanker:
@KimHolzmann @nul800sebastiaan I'm still seeing this happen in 7.12.4 :-s Also the browse dialog seems to pop up three times, seemingly unmotivated (although that might be an entirely different issue).

Any idea what is happening?

![3022](https://user-images.githubusercontent.com/1769849/57017787-f1068380-6c20-11e9-850c-6c4944ecce05.gif)

KirschnerMatt:
This is still happening on a fresh install of 7.15.3

enkelmedia:
Same here!

enkelmedia:
One idea, at least is to ship the "nothing.jpg"-image so that the editors can see that something went wrong.

jannikanker:
@nul800sebastiaan Might there be grounds for reopening this case? I know full well that I reported it using an older version than the fix, but people seem to still be experiencing it in the latest (v7) version :-/

nul800sebastiaan:
Create a new issue for it please, whatever was fixed in this PR is what this issue is marking, apparently there's something else / additional wrong or it wasn't fixed properly. We can't re-open issues that have a release tag on them. :)

PR #6739, opened 2019-10-17
V7: Media uploader in RTE doesn't select uploaded image

### LINKED ISSUE #6720: V7: Media uploader in RTE doesn't select uploaded image
This is a repost of the following issue: https://github.com/umbraco/Umbraco-CMS/issues/3022

While the issue might have been fixed before, it seems to have come back for one reason or another. This may be reproduced by uploading images to an RTE in a fresh install of V7.15.3

The original comment by JorickDJ is still relevant:

> This issue was already submitted on 12th of September 2017 and we are still having problems with this issue in Umbraco version 7.10.5.
> 
> The previous bug submission can be found here:
> https://issues.umbraco.org/issue/U4-10408
> 
> When I want to add a new image to the RTE I click on the Image Picker and upload a new image. The Image Picker indicates that the image is selected and when you click "Submit" it adds a "nothing.jpg" to the RTE.
> 
> If you upload a new image in the RTE Image Picker and then scroll to the uploaded image and select it again, then it will be added correctly.

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #6720 

### Description
This was previously fixed as #3022, but a further change is needed after 7.15's changes to the media tree.

Testing:
- In 7.15.x, open the media picker from an RTE and upload a new image.
- The image should be correctly inserted, rather than a reference to "nothing.jpg".

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/overlays/mediaPicker/mediapicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/overlays/mediaPicker/mediapicker.controller.js
@@ -251,7 +251,12 @@ angular.module("umbraco")
                     if (files.length === 1 && $scope.model.selectedImages.length === 0) {
                         var image = $scope.images[$scope.images.length - 1];
                         $scope.target = image;
-                        $scope.target.url = mediaHelper.resolveFile(image);
+                        // handle both entity and full media object
+                        if (image.image) {
+                            $scope.target.url = image.image;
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hiya Steve,

So good to see so much of you here. Thanks for the PR. We'll let you know if we need anything.

Emma

elit0451:
Tested your fix @stevemegson works as intended! 👍 
Merging!

PR #6749, opened 2019-10-17
Hide the drag and drop on screenreaders in the media section file upload and media picker overlay

poornimanayar:
I have hidden the drag and drop on screenreaders in the media section file upload and the media picker overlay. Also corrected the text on the upload link to make it not sound too off in such an instance


poornimanayar:
Sure, can do that @nul800sebastiaan . Will take a look later on today :-)

nul800sebastiaan:
Actually, come to think of it, the screen reader will never even see the dropzone, you just tab the "Upload" button and get asked for a file, I don't imagine a visually impaired person dragging a file in.

So I've removed that button and the translations again and that leaves us with a dropzone which again will never appear for a screen reader anyway, it only appears if you hover a file onto the browser.

I think there's not change needed here, I'll close this one. Thanks Poornima and sorry for making you do extra work! :-)

PR #6751, opened 2019-10-17
Add additional null conditionals to the FromEditor method of RichTextPropertyValueEditor

Jeavon:
Adds additional null conditionals to the FromEditor method as this method is sometimes executed for front end rendering and in that scenario CurrentUser is null, this throws an exception.

See the Doc Type Grid Editor issue for details of when this issue occurs https://github.com/skttl/umbraco-doc-type-grid-editor/issues/159

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
@@ -89,7 +89,7 @@ namespace Umbraco.Web.PropertyEditors
 
                 var grid = DeserializeGridValue(rawJson, out var rtes);
 
-                var userId = _umbracoContextAccessor.UmbracoContext?.Security.CurrentUser.Id ?? Constants.Security.SuperUserId;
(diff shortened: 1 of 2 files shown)


milandt:
This fix solved this exact problem for me. The PR should be accepted :)

clausjensen:
@Jeavon .. and thanks for the PR 😄 things went a little fast yesterday so completely forgot to write here!

FransdeJong:
Is this fix included in 8.3? I still run into issues with DTGE.

clausjensen:
@FransdeJong yes - the fix is applied to 8.3 as well. You however can't see it by looking at the commit hash as it was manually merged over.
https://github.com/umbraco/Umbraco-CMS/blob/release-8.3.0/src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs#L92

You can see the merged commit here:
https://github.com/umbraco/Umbraco-CMS/commit/5e299d7880ca979797720dd1276b7bf9b92b90c8

So it must be a different bug you're seeing.

PR #6752, opened 2019-10-17
Boot processing cleanup.

John-Blair:
### Prerequisites

### DIFF
--- a/src/Umbraco.Core/MainDom.cs
+++ b/src/Umbraco.Core/MainDom.cs
@@ -172,17 +172,27 @@ namespace Umbraco.Core
                 // if more than 1 instance reach that point, one will get the lock
                 // and the other one will timeout, which is accepted
 
-                //TODO: This can throw a TimeoutException - in which case should this be in a try/finally to ensure the signal is always reset?
-                _asyncLocker = _asyncLock.Lock(LockTimeoutMilliseconds);
+                //This can throw a TimeoutException - in which case should this be in a try/finally to ensure the signal is always reset.
+                try
+                {
+                    _asyncLocker = _asyncLock.Lock(LockTimeoutMilliseconds);
+                }
+                catch
+                {
+                    throw;
+                }
+                finally
+                {
+                    // we need to reset the event, because otherwise we would end up
+                    // signaling ourselves and committing suicide immediately.
+                    // only 1 instance can reach that point, but other instances may
+                    // have started and be trying to get the lock - they will timeout,
+                    // which is accepted
+
+                    _signal.Reset();
+                }
                 _isMainDom = true;
-
-                // we need to reset the event, because otherwise we would end up
-                // signaling ourselves and committing suicide immediately.
-                // only 1 instance can reach that point, but other instances may
-                // have started and be trying to get the lock - they will timeout,
-                // which is accepted
-
-                _signal.Reset();
+               
 
                 //WaitOneAsync (ext method) will wait for a signal without blocking the main thread, the waiting is done on a background thread
 
--- a/src/Umbraco.Core/Runtime/CoreRuntime.cs
+++ b/src/Umbraco.Core/Runtime/CoreRuntime.cs
@@ -140,23 +140,25 @@ namespace Umbraco.Core.Runtime
                 Compose(composition);
 
                 // acquire the main domain - if this fails then anything that should be registered with MainDom will not operate
-                AcquireMainDom(mainDom);
+                if (AcquireMainDom(mainDom))
+                {
+                    // determine our runtime level
+                    DetermineRuntimeLevel(databaseFactory, ProfilingLogger);
 
-                // determine our runtime level
-                DetermineRuntimeLevel(databaseFactory, ProfilingLogger);
+                    // get composers, and compose
+                    var composerTypes = ResolveComposerTypes(typeLoader);
+                    composition.WithCollectionBuilder<ComponentCollectionBuilder>();
+                    var composers = new Composers(composition, composerTypes, ProfilingLogger);
+                    composers.Compose();
 
-                // get composers, and compose
-                var composerTypes = ResolveComposerTypes(typeLoader);
-                composition.WithCollectionBuilder<ComponentCollectionBuilder>();
-                var composers = new Composers(composition, composerTypes, ProfilingLogger);
-                composers.Compose();
+                    // create the factory
+                    _factory = Current.Factory = composition.CreateFactory();
 
-                // create the factory
-                _factory = Current.Factory = composition.CreateFactory();
+                    // create & initialize the components
+                    _components = _factory.GetInstance<ComponentCollection>();
+                    _components.Initialize();
+                }
 
-                // create & initialize the components
-                _components = _factory.GetInstance<ComponentCollection>();
-                _components.Initialize();
             }
             catch (Exception e)
             {
--- a/src/Umbraco.Web/PublishedCache/NuCache/ContentStore.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/ContentStore.cs
@@ -235,11 +235,17 @@ namespace Umbraco.Web.PublishedCache.NuCache
             var lockInfo = new WriteLockInfo();
             try
             {
-                Lock(lockInfo);
-
-                if (_localDb == null) return;
-                _localDb.Dispose();
-                _localDb = null;
+                try{
+                    // Trying to lock could throw exceptions so always make sure to clean up.
+                    Lock(lockInfo);
+                }
+                catch 
+                {
+                    if (_localDb == null) return;
+                    _localDb.Dispose();
+                    _localDb = null;
+                }
+                
             }
             finally
             {
--- a/src/Umbraco.Web/PublishedCache/NuCache/ContentStore.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/ContentStore.cs
@@ -239,11 +239,21 @@ namespace Umbraco.Web.PublishedCache.NuCache
                     // Trying to lock could throw exceptions so always make sure to clean up.
                     Lock(lockInfo);
                 }
-                catch 
+                catch { throw; }
+                finally 
                 {
-                    if (_localDb == null) return;
-                    _localDb.Dispose();
-                    _localDb = null;
+                    if (_localDb != null)
+                    {
+                        try
+                        {
+                            _localDb.Dispose();
+                        }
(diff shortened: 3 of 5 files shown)


src/Umbraco.Web/PublishedCache/NuCache/ContentStore.cs
```diff
@@ -235,11 +235,17 @@ public void ReleaseLocalDb()
             var lockInfo = new WriteLockInfo();
             try
             {
-                Lock(lockInfo);
-
-                if (_localDb == null) return;
-                _localDb.Dispose();
-                _localDb = null;
+                try{
```
Shazwazza: For this you mentioned

> The lock call could throw an exception in the Lock attempting to do a Monitor.Enter that is not caught and won't cause the nucache db  to be properly closed, suggest you add a catch and always dispose  of the _localDB

Why would the Lock call throw? And if it did we would have seen this exception occuring in the past but we've never seen this happen. 

Also, this code is now removing the disposal of the _localDb **unless** an exception occurs which is not what we want. It might also be possible that this whole thing is wrapped in a try/finally because disposing the _localDb might throw?

src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
```diff
@@ -140,8 +142,11 @@ internal class PublishedSnapshotService : PublishedSnapshotServiceBase
                         lock (_storesLock)
                         {
                             _contentStore?.ReleaseLocalDb(); //null check because we could shut down before being assigned
+                            // Make sure the local dbs are disposed of if the content store was not created.
+                            _localContentDb?.Dispose();
```
Shazwazza: This is unecessary since this gets done as part of `_contentStore.ReleaseLocalDb`

src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
```diff
@@ -128,10 +128,12 @@ internal class PublishedSnapshotService : PublishedSnapshotServiceBase
                         var path = GetLocalFilesPath();
                         var localContentDbPath = Path.Combine(path, "NuCache.Content.db");
                         var localMediaDbPath = Path.Combine(path, "NuCache.Media.db");
-                        _localDbExists = File.Exists(localContentDbPath) && File.Exists(localMediaDbPath);
```
Shazwazza: You've removed using the flag `_localDbExists` but the actual variable still exists

src/Umbraco.Core/MainDom.cs
```diff
@@ -172,17 +172,27 @@ internal bool Acquire()
...
-                // have started and be trying to get the lock - they will timeout,
-                // which is accepted
+                //This can throw a TimeoutException - in which case should this be in a try/finally to ensure the signal is always reset.
+                try
+                {
+                    _asyncLocker = _asyncLock.Lock(LockTimeoutMilliseconds);
+                }
+                catch
+                {
+                    throw;
```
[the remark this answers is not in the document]
John-Blair: The catch; throw is necessary to replicate the original logic exactly where the error is thrown - we don't want to swallow it. The only change is to guarantee the signal is reset.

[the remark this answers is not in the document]
John-Blair: Timeout is thrown as an exception - its not an error - im still waking up.

src/Umbraco.Core/Runtime/CoreRuntime.cs
```diff
@@ -140,23 +140,25 @@ protected virtual IFactory Boot(IRegister register, DisposableTimer timer)
                 Compose(composition);
 
                 // acquire the main domain - if this fails then anything that should be registered with MainDom will not operate
-                AcquireMainDom(mainDom);
+                if (AcquireMainDom(mainDom))
```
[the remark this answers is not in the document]
John-Blair: The main time it will return false is when it has been signaled.  When the components run and call register - the register checks for _signaled being true and returns immediately without doing the register. So you are partially not doing the register. The app domain  is not the main dom - it should be doing as little as possible and bailing out as early as possible to minimise the chances of causing weird errors. Now i accept there may be unforseen side effects and dev errors - but that is just a sign that other things need cleaning up. I can see both sides of the argument.

src/Umbraco.Web/PublishedCache/NuCache/ContentStore.cs
John-Blair: Apologies - you are correct - it should have been in a finally - i have updated the code.
The interface signature on the Monitor.Enter shows it can throw exceptions - so we should handle them to make sure _localDB is always cleaned up. Yes disposing might throw so I've now wrapped that.

src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
John-Blair: _contentStore has the Elvis check ...uuuhuh? so it may not have been created to release the local db so this is defensive to make sure of cleanup.

John-Blair: Added further defences around the dispose calls.

src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
John-Blair: _localDbExists is still used elsewhere by LoadCachesOnStartup , so keeping this value the same as present to not affect behaviour elsewhere.

src/Umbraco.Core/MainDom.cs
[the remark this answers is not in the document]
Shazwazza: You can just do a try/finally, it is the same as this. A try/finally doesn't swallow anything, there's now need to catch/throw.

src/Umbraco.Core/Runtime/CoreRuntime.cs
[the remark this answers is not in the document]
Shazwazza: Sure, but this appdomain is definitely going to serve at least one request. If we just terminate now, those requests will definitely get errors, whereas there's a chance they won't if we let the app bootup. The only thing that will definitely not work is Examine, the rest will continue to operate as normal, except that nucache will be read into memory via calls to the DB.

src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
Shazwazza: However, you've updated how the 2 different stores are checked and then we only set _localDbExists to true if **both** exist, later we check this flag in `LoadCaches`. It would seem to me that `LoadCaches` should use individual checks of whether the media or content caches exist and proceed accordingly instead of using this single field. I think it would make more sense to have the fields: `_localMediaDbExists` and `_localContentDbExists` and use these fields in place of wherever `_localDbExists` is used and remove `_localDbExists` entirely.

src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
```diff
@@ -140,8 +142,19 @@ internal class PublishedSnapshotService : PublishedSnapshotServiceBase
                         lock (_storesLock)
                         {
                             _contentStore?.ReleaseLocalDb(); //null check because we could shut down before being assigned
+                            // Make sure the local dbs are disposed of if the content store was not created.
+                            try
+                            {
+                                _localContentDb?.Dispose();
```
Shazwazza: I know you've added additional try/catch's here, but my first comment was saying that you don't need to dispose of this here at all since it's definitely going to be disposed of in `_contentStore?.ReleaseLocalDb` above.

src/Umbraco.Web/PublishedCache/NuCache/ContentStore.cs
```diff
@@ -235,11 +235,27 @@ public void ReleaseLocalDb()
...
+                    // Trying to lock could throw exceptions so always make sure to clean up.
+                    Lock(lockInfo);
+                }
+                catch { throw; }
```
Shazwazza: Again, no need to do a try/catch/throw in a try/finally, it does not swallow anything in a try/finally

src/Umbraco.Core/MainDom.cs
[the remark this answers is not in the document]
John-Blair: Agreed! You wouldn't think I've been doing c# a long time XD
I generally use a global error handler on my websites , and rarely find the need to use try/catch/finally - keeps my code clean - ....but i do see the need/use of them here.

src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
John-Blair: agreed

src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
John-Blair: Your call - just noticed the use of ?. and catering for the null scenario.

src/Umbraco.Core/Runtime/CoreRuntime.cs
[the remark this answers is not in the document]
John-Blair: Your call. If you really think a false return should be ignored you should really change the method signature to return void so a reviewer knows all callers should continue processing regardless of whether the main dom was acquired or not - which does seem odd.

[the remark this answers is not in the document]
Shazwazza: We can't change the signature = breaking change and perhaps in some cases a caller will want to know if Acquiring was true. In any case, we cannot just explode if we don't Acquire the MainDom, the normal boot process must continue.

src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs
Shazwazza: I suppose it's safer, fine to leave the additional disposal call

Shazwazza:
I have merged in all of these changes along with updates based on our converstations in this code. All changes now exist in this branch/PR: https://github.com/umbraco/Umbraco-CMS/pull/6757

nul800sebastiaan:
Hi there @John-Blair,

Kind regards,

PR #6748, opened 2019-10-17
Fix for #3510 - health check for http headers was case sensitive, but…

tompipe:
This came up in todays Umbraco Security training course. 

The health checks look for HttpHeaders but was case sensitive, but HTTP spec states headers are case insensitive. Looks like this was raised long ago in #3510 - hopefully it can now be merged and closed

PR #3551, opened 2018-11-08
3510 - Fixes Health check false negative in X-XSS-Protection header check 

MMasey:
This PR fixes #3510 by adding a warning message when an X-XSS-Protection header exists in the web.config, but is in the wrong case. It also provides a button to fix the issue.

@Matthew-wise also helped with this.

PR #6753, opened 2019-10-17
V8: Health check http header case sensitivity

tompipe:
Slight update for #6748 to address the issue raised in https://github.com/umbraco/Umbraco-CMS/pull/3551#issuecomment-438000340


emmaburstow:
Hiya Tom!

Hope you're well. Thanks for doing this and linking the issue for us to look at too. We'll shout if we need you

Em

nul800sebastiaan:
Thanks again @tompipe ! 👍

PR #6755, opened 2019-10-17
V8: Make the color class prefix configurable for <umb-color-swatches/>

### LINKED ISSUE #6740: v8 UX: Swatch picker is hard coded to prefix class names with `btn-`
I was looking to reuse the swatch picker control in a custom package so fed it a series of colours and told it to treat the values as class names. When I then tried to use the control it didn't work (all the swatches were grey).

In looking into the source, it appears the swatch picker is built only to work with Umbraco colour class names as it assumes the values will be `color-SOMETHING` and to get a background colour for the swatch it then prefixes those with `btn-` to use the Umbraco `btn-color-SOMETHING` classes.

The control should not assume the colours will be coming from Umbraco. If you need to prefix the css classes, then the control should accept a `bg-class-name-prefix` setting to allow this to be customized.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #6740

### Description

This PR makes it possible to configure the prefix applied to the class name for each color option in the color swatch directive.

The prefix is defined by supplying `color-class-name-prefix="my-prefix"` in the directive. If none is supplied, the directive performs a fallback to "btn" to ensure backwards compatibility.

#### Testing this PR

First and foremost check that the color swatch performs the expected fallback - for example by opening the icon picker for a doctype. If it still looks really colorful, the fallback works.

image

Now testing it with a customized prefix isn't straightforward. What I did was to edit the color swatch used in the icon picker:

image

...which caused it to look like this:

image

Success! It's colorless 😆

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbcolorswatches.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbcolorswatches.directive.js
@@ -18,6 +18,7 @@ Use this directive to generate color swatches to pick from.
 @param {string} size (<code>attribute</code>): The size (s, m).
 @param {string} useLabel (<code>attribute</code>): Specify if labels should be used.
 @param {string} useColorClass (<code>attribute</code>): Specify if color values are css classes.
+@param {string} colorClassNamePrefix (<code>attribute</code>): Specify the prefix used for the class for each color (defaults to "btn").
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Ooh, I kinda like the colourless swatches! ;-)

Thanks very much, such a simple update!

PR #6756, opened 2019-10-17
V8: Handle save keyboard shortcut from within the RTE

kjac:
### Prerequisites

### Description

I keep getting annoyed with how the system save dialog appears whenever I hit Ctrl+S in the RTE:

rte-save-keyboard-shortcut-before

Turns out there is an easy fix that makes Ctrl+S save the current content instead, just like it should. So this PR implements just that:

rte-save-keyboard-shortcut

#### What about Publish?

While it would be just as easy to handle the shortcut for publishing (Ctrl+P), I have opted not to do it. The editors might hit the shortcut by accident and trigger an unwanted publish of the entire page. 

Of course the editors could just as easily hit the shortcut for save by accident and trigger an unwanted save, but that's nowhere near as problematic. And I do believe the advantage of having a functional shortcut outweighs this potential problem.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
@@ -201,6 +201,10 @@
                 $scope.page.buttonGroupState = "success";
             }));
 
+            evts.push(eventsService.on("rte.shortcut.save", function(){
+                $scope.save();
+            }));
+
             evts.push(eventsService.on("content.saved", function(){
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
YAAAY! I've been annoyed by this for years! Why did I never create an issue for it 🤦‍♂ 

I don't see why CTRL+P shouldn't work though, it does work everywhere else you use it. I'd rather see it working consistently.

nul800sebastiaan:
Sure.. let's try it out with only CTRL+S for now and see how it goes! 

I **LOVE** this update! Thank you! 🎉

PR #6767, opened 2019-10-18
Added thumbs.db to .gitignore

sidhyatikku:
### Description

`thumbs.db` is a windows directory cache file which sometimes gets pushed to repos, comes in commits, and can trouble other users. Now windows users can send PRs without having to worry about `thumbs.db`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@
 ._.DS_Store
 .vs/
 /local/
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks! 👍 This has never been a problem for us, we don't need this addition, sorry! 😊

PR #6768, opened 2019-10-18
V8: Make Nested Content resilient to missing property editors

kjac:
### Prerequisites

### Description

If you remove a custom property editor that's currently used by a property on a content type, the content editor handles this rather gracefully by simply displaying the raw property value instead of the missing property editor:

image

Nested Content... not so much 😆

image

This is particularly annoying when you're building custom editor experiences in different branches against the same database.

This PR makes Nested Content mimic the content editor handling for missing property editors:

image

#### Testing this PR

Here's a simple custom property editor you can use for testing: [TestPropertyEditor.zip](https://github.com/umbraco/Umbraco-CMS/files/3745083/TestPropertyEditor.zip)

Steps:

1. Add the property editor to `App_Plugins`.
2. Create a Nested Content configuration with an element type that uses this property editor.
3. Create some content based on this Nested Content configuration. 
4. Remove the property editor from `App_Plugins` and restart the site.
5. Verify that Nested Content loads without errors and displays the raw property value instead of the (now missing) property editor.

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
@@ -124,7 +124,6 @@ namespace Umbraco.Web.PropertyEditors
                                 var propEditor = _propertyEditors[propType.PropertyEditorAlias];
                                 if (propEditor == null)
                                 {
-                                    propValues[propAlias] = null;
                                     continue;
                                 }
                                 var tempConfig = dataTypeService.GetDataType(propType.DataTypeId).Configuration;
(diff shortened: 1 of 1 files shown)


poornimanayar:
Hi @kjac,

I have tested what you have mentioned above, works okay. I can also see that when the property editor is deleted the temp value is shown and I can now publish the page without any errors. Now, do a second publish again and move back and forth between the nodes and you can see that the raw property value is also gone.  While this is a great improvement from what we have now, do you think we should have some messaging rather than the raw property value? 

Poornima

kjac:
@poornimanayar good find. Turns out the raw property value is deleted on the first publish; you just can't see it in the UI until you reload the browser. 

I found the issue - the raw property value is now preserved on publish:

nested-content-missing-property-editor

poornimanayar:
Also tested that once the deleted PE is added back it "just works", top class work @kjac !

nul800sebastiaan:
This is amazing! Errors begone! 💯

PR #6769, opened 2019-10-18
V8: Style and translate the composition removal confirmation dialog

kjac:
### Prerequisites

### Description

Considering the impact of removing compositions from a content type, the confirmation dialog isn't really as foreboding as it should be:

image

It's also a bit off styling wise, compared to the other dialogs. And as it turns out it's not localizable either.

With this PR the dialog becomes a bit more ominous 😨 

image

This dialog is also in line with other dialogs, and it's localizable.


poornimanayar:
Thanks again @kjac 

Poornima

nul800sebastiaan:
Nice one @kjac! 👍 

Still not sure what is going on with these language files, hope this merge might even it out again.. only en_us is "different" now but I can not at all see a difference in my diff tooling nor on GitHub.. weird.

PR #6771, opened 2019-10-18
V8: Harmonize content save/publish dialogs with mandatory languages

kjac:
### Prerequisites

### Description

The save/publish dialogs for multilingual content are a bit inconsistent when you have a mandatory language. In some dialogs an asterisk is displayed next to the language, in some the label "Mandatory language" is added - and some have both. 

On top of that, #5635 has the unintended side effect of pushing the asterisk down below the checkbox, leaving the variant status label somewhat floating.

image

image

image

Considering how the asterisk usually denotes "required" and that it's not necessarily required to tick the mandatory languages in these dialogs, I think we should simply remove them and add the "Mandatory language" everywhere instead. So this PR does just that:

image

image

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/overlays/publish.html
+++ b/src/Umbraco.Web.UI.Client/src/views/content/overlays/publish.html
@@ -24,8 +24,6 @@
                                   text="{{ variant.language.name }}"
                     />
                     <div>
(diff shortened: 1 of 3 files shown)


poornimanayar:
Keen eye to detail as usual @kjac !

Poornima

nul800sebastiaan:
Love it!

PR #6773, opened 2019-10-19
6772 Img is self closing

JohJohan:
### Prerequisites


nul800sebastiaan:
Thanks! I wish you would've asked before putting the effort in. We like closing `img` tags and don't want to change this.

PR #6775, opened 2019-10-19
V8: Datetime picker should submit its value on enter when entering time

kjac:
### Prerequisites

### Description

When you have a datetime picker, I would expect it to submit its value on enter after you input the time. But it doesn't. You have to hit escape to close the datepicker (which incidentally does submit the value... that's arguable):

datepicker-enter-before

This PR ensures that the datetime picker submits its value when enter is pressed. It works like this:

datepicker-enter-after


poornimanayar:
Hi @kjac ,

Thank you once again!

Poornima

nul800sebastiaan:
Fantastic addition! 👍

PR #6776, opened 2019-10-19
V8: Do not allow deleting inherited content type groups

kjac:
### Prerequisites

### Description

The backoffice lets us delete inherited groups from content types. Fortunately the server doesn't actually delete the inherited groups, but still... the client shouldn't allow it in the first place:

delete-inherited-groups-before

Fortunately it's a simple fix. With this PR, inherited groups can't be deleted clientside:

delete-inherited-groups-after


poornimanayar:
Thanks once again @kjac !

Poornima

nul800sebastiaan:
Beauty!

PR #6777, opened 2019-10-19
V8: Remove "double tabbing" and duplicate editor for datatype picker

kjac:
### Prerequisites

### Description

If you're prone to tabbing your way through the content type creation, it's a bit of a hassle picking the right datatype for properties. You have to tab twice for each datatype when you're moving through the list of datatypes; once to highlight the datatype - the outer `li` element - and once more to highlight the contained `a` element.

What's worse is if you hit enter when the inner `a` element is highlighted, the "new datatype" editor is displayed twice (it's only displayed once if you hit enter on the outer `li` though):

property-type-tabbing-before-annotated

This PR fixes it all up so you can tab your way through the datatype picker like you'd expect:

property-type-tabbing-after


poornimanayar:
👍 :-)

Poornima

nul800sebastiaan:
I love it!

Issue #6410, opened 2019-09-23
Add 'Add as Element' to Doc Type Create menu


philipdanielhayton:
Could we have the ability to add a new doc type as an Element. This would save me a load of clicks each time I forget to set new types as elements.

FransdeJong:
As I see there are two issues this would solve.

1. It removes the extra clicks needed to create a element type
2. If I want to quickly create a nested content Item I regularly find myself creating a DocType, creating a DataType, go back to the DocType because I forgot to set it to element, go back to DataType.


If for whatever reason a extra Item in the create documenttype menu is unwanted this issue could also be solved by moving the toggle to the top of the design page.

skttl:
I would love this, 80% of the doc types I create is elements, and I usually follow @FransdeJong s pattern :)

nul800sebastiaan:
We'll have a chat at HQ, seems like an okay addition to me, but we'll soon get loads of choices and I'm afraid newcomers will have no understanding of what they mean and why they should choose one over the other.

There's a PR for this (and more) as well: https://github.com/umbraco/Umbraco-CMS/pull/5580

nul800sebastiaan:
Thanks @thehaytonproject - we've had a chat about this today and we'll put this on our longer terms "ideas" list for when we start working on a UI update for document types.

Our discussion involved:
- Too many options in this menu confuses new users, it's already difficult to learn that you need a document type to get started and "bombarding" people with choices is only going to make that harder
- An idea we'd like to test is to set up default folders under "Document types" (`Elements`, `Compositions`, `List views`) under which you can only create document types of that very specific type. We'd also have a change to add help text on a dashboard while clicking each of these items. Just an idea, it might work.

thehaytonproject:
Hi @nul800sebastiaan 

Thanks to yourself and the team for taking the time to review my suggestion.

I can't help but feel that you've over thought this one. It sounds like there are lots of big changes coming in a future update, which I look forward to hearing more about.... but until then, is it really so risky to add one additional button?

Does the *potential* issue of confusing new developers out weight the opportunity to provide a *real* QoL improvement?

Is adding one additional button *really* going to baffle new developers? (Keen to stress that we're talking about developers here)

Is going from 3 buttons to 4 really bombarding users? 

I'm a little sleep deprived so apologies if my post seems narky 😅 I don't mean to sound bitter, I just disagree with what your justification. I do value and appreciate the work you and the guys do.

[...]

PR #6779, opened 2019-10-19
Allow querystring parameters to set IsElementType checkbox on create document type route

marcemarc:
### Prerequisites

- [  + ] I have added steps to test this contribution in the description below

Relates to issue https://github.com/umbraco/Umbraco-CMS/issues/6410

### Description

Some of the cool Umbraco folk I know were chatting around the issue #6410 which is essentially is 'all about' adding to the Create Document Type options, a super fast way to create an Element Type, instead of having to create first a Document Type Without Template, and then remember to manually check the Is An Element Type checkbox on the permissions tab...

... why can't they just have a new option to do that on that menu ... speed up everyone's daily workflow...

image

... well there are concerns on how to present these options, in a way that doesn't confuse people new to Umbraco (what is a Document Type Collection anyway? etc)... which will involve some due consideration and perhaps an RFC...

... anyway in the meantime, devs are pulling their hair out, everyday, right click create, document type without template, permissions check is an element type and they are getting frustrated, and well I don't know just how far this is going to escalate...

So this PR doesn't address the menu, or adding that in... there is a whole lot of discussion required first - so what does it do? and why am I proposing it?

This PR, updates the DocumentTypes edit.controller.js to look on the querystring for two new parameters

iselement
&
culturevary

if these parameters are set to true on the querystring, then the checkbox toggle for 'Is an Element Type' and 'Allow varying by culture' will by default be also set to true.

image

image

image

This changes nothing for day to day creating of Document Types! - nothing is visible to anyone via the menu, shhhsshsh nobody needs to know this change has happened... just merge it in... nobody will know... 

... but what it enables is a cool kid package developer to experiment with the options for this menu... to allow a 'power tools' document type options plugin to be created - that can perhaps enable evolution in the realm of the package sphere to identify and test the long term improvements required to be made to this menu in the core product... in the short term, devs can have a boost to their productivity, whilst buying some time for the longer term considerations...

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/documenttypes/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/documenttypes/edit.controller.js
@@ -62,6 +62,8 @@
                 documentTypeId = $scope.model.id;
                 create = $scope.model.create;
                 noTemplate = $scope.model.notemplate;
+                isElement = $scope.model.iselement;
+                allowVaryByCulture = $scope.model.allowvarybyculture;
                 vm.submitButtonKey = "buttons_saveAndClose";
                 vm.generateModelsKey = "buttons_generateModelsAndClose";
             }
--- a/src/Umbraco.Web.UI.Client/src/views/documenttypes/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/documenttypes/edit.controller.js
@@ -62,8 +62,8 @@
                 documentTypeId = $scope.model.id;
                 create = $scope.model.create;
                 noTemplate = $scope.model.notemplate;
-                isElement = $scope.model.iselement;
-                allowVaryByCulture = $scope.model.allowvarybyculture;
+                isElement = $scope.model.isElement;
+                allowVaryByCulture = $scope.model.allowVaryByCulture;
                 vm.submitButtonKey = "buttons_saveAndClose";
                 vm.generateModelsKey = "buttons_generateModelsAndClose";
             }
--- a/src/Umbraco.Web.UI.Client/src/views/documenttypes/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/documenttypes/edit.controller.js
@@ -18,6 +18,8 @@
         var documentTypeId = $routeParams.id;
         var create = $routeParams.create;
         var noTemplate = $routeParams.notemplate;
+        var isElement = $routeParams.iselement;
+        var allowVaryByCulture = $routeParams.culturevary;
         var infiniteMode = $scope.model && $scope.model.infiniteMode;
 
         vm.save = save;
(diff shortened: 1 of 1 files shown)


poornimanayar:
Hey @marcemarc ,

Thank you for the PR. We shall review it and get back to you if anything more is needed. 

Poornima

marcemarc:
@abjerner Yes I spotted that too, to be honest, I wasn't sure how the new create Document Type view could be created from the context of an infinite editor? so wasn't quite sure how to test making that change, or whether it made sense.

abjerner:
@marcemarc it appears that you can call `editorService.documentTypeEditor(options)`:

https://our.umbraco.com/apidocs/v8/ui/#/api/umbraco.services.editorService#methods_documenttypeeditor

But the documentation doesn't really have any more information beyond that the function exists. I'll try to look a bit more into it ;)

abjerner:
@marcemarc I did some digging, and it seems that a create dialog can be opened by specifying both `id` and `create` as:

```javascript
editorService.documentTypeEditor({
	id: -1,
	create: true,
	noTemplate: true
});
```

From what I can tell, there is no part of the Umbraco source code that actually opens a document type create dialog like this. But there might be in the future, or a package might use it, so I think adding `isElement` and `allowVaryByCulture` makes sense. The create dialog is also the only way I can get passing the `noTemplate` option to the editor service to make sense.

marcemarc:
@abjerner thanks for the digging, I couldn't envisage where it would be used... so it's good to know it isn't... less worried about changing it now - so I have pushed a change, that I think would make this possible via the infinite editor, though I'm not sure how best to test ...

src/Umbraco.Web.UI.Client/src/views/documenttypes/edit.controller.js
```diff
@@ -60,6 +62,8 @@
                 documentTypeId = $scope.model.id;
                 create = $scope.model.create;
                 noTemplate = $scope.model.notemplate;
+                isElement = $scope.model.iselement;
+                allowVaryByCulture = $scope.model.allowvarybyculture;
```
abjerner: @marcemarc thanks 😄 

I may be a bit picky, but could you change `$scope.model.iselement` and $scope.model.allowvarybyculture` to `$scope.model.isElement` and $scope.model.allowVaryByCulture` respectively?

Ideally these should be camel cased, as this follows the convention used for the parameters for the other overlays. I know `$routeParams.notemplate` is all lowercase, but IMO that is also wrong (we can perhaps fix that later).

nul800sebastiaan:
Thanks @marcemarc and @abjerner ! 👍 


👀 
image
👀

marcemarc:
sorry... @nul800sebastiaan but I think you are expecting me to do something different now each time, and I don't want to disappoint :-P

nul800sebastiaan:
😂 I was indeed waiting for that!

PR #6781, opened 2019-10-20
Rollback dropdown - convert version dates to local

### LINKED ISSUE #6738: Rollback issue in 8.2.0
Hi,

There is a bug in the rollback feature in umbraco v 8.2.0. The time in the rollback drop-down is not in sync with the history time details. There is a 2 hours time difference between them.

Reproduction
------------

* go to the "Info" tab of any content page which has been saved and published multiple times
* in the "History" section click the "Rollback" button
* expand the "Rollback to" drop down
* the time mentioned in the drop down is different to the time listed in the History section, there is 2 hours time difference between them

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #6738

### Description
The History section of a document's Info tab converts times to the client's local time, but the "Rollback to" dropdown displays times in the server's timezone.

This PR converts the times in the dropdown to local times.

Testing:
- With a server in a different timezone from the client, compare the times displayed in the History section and the "Rollback to" dropdown. They should match, both in the client's local time.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/rollback/rollback.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/rollback/rollback.controller.js
@@ -1,7 +1,7 @@
 (function () {
     "use strict";
 
-    function RollbackController($scope, contentResource, localizationService, assetsService) {
+    function RollbackController($scope, contentResource, localizationService, assetsService, dateHelper, userService) {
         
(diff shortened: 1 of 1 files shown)


poornimanayar:
Hi again @stevemegson ,

Thank you for the work, we shall get back to you.

Poornima

nul800sebastiaan:
Super, thanks very much @stevemegson! 👍

PR #6782, opened 2019-10-20
Cache lookup of macro by alias

stevemegson:
### Description
Rendering a macro always queries the database to get the macro details from the alias. This is true even if the macro result is already cached, since we need the macro's details to generate the correct cache key.

This PR adds caching of the `IMacro` results within `MacroRenderer`.

It may be safe to cache within MacroService.GetByAlias instead, but it seemed better to make MacroRenderer responsible for its own cache, with no risk of side effects for other code using MacroService.


poornimanayar:
Hi @stevemegson ,

Many thanks for your work. We shall review it and get back to you :-)

Poornima

nul800sebastiaan:
I love it, thanks very much @stevemegson!

Issue #6408, opened 2019-09-23
Media subfolder not shown in media picker window


milance-feelgood:
Hello,

I wanted to report a bug which is related to media folders (subfolders) which are shown in media picker window. 

Reproduction
------------

Please take a look at attached screenshots which will give you full picture of media folders structure and what is missing in media picker window when we want to choose an image.

### Bug summary

We have user which has assigned media nodes which can be accessible for it.
Our media structure:
We have 3 main media folders: 
**Common Data** 
**Common Images** and
**Local Media** which consists of subfolders (for each of domain languages we have created subfolder here).

For this specific user we have selected in media nodes subfolder **at** under **Local Media** folder. 
When we are logged in with this user and we try to choose an image, we can not see this subfolder **at** in media picker window until we have selected **Include subfolders in search**
checkbox.

### Specifics

**Umbraco Version 8.1.2**

### Steps to reproduce

- Click on media picker (media picker window would open)
- In media picker window you will have overview of all folders which user can see (based on assigned user permissions and picked media nodes).

### Expected result

We expect that in media picker window user can see all media folders which are selected in its user media nodes even they are subfolders. So, we don't need to select first **Include subfolders in search** to see correct media structure.

### Actual result
 
When we open media picker window we can not see subfolder which would be possible to see based on picked media nodes for the current user.

Notes: Because this bug consists of problem with media access in media picker which is also related to user permissions and selected media nodes, maybe is related to 2 different topics: media picker and user permissions media access.

Thanks.

![2-content-media-picker-items](https://user-images.githubusercontent.com/54978926/65417236-e0d28900-ddf9-11e9-88ff-6bc81c3d75c1.jpg)
![1-user-permission-media-access-and-structure](https://user-images.githubusercontent.com/54978926/65417237-e16b1f80-ddf9-11e9-9d4e-5634ee2915ca.jpg)

---
_This item has been added to our backlog [AB#3406](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3406)_

kjac:
Yeah... it's got nothing to do with Examine. It's a bug on the clientside of things.

PR in #6786

umbrabot:
Hiya @milance-feelgood,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

[...]

PR #6786, opened 2019-10-20
V8: Make sure all user start nodes are shown in the media picker

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6408

### Description

See a detailed issue description in #6408

For some reason it's up to the `umb-media-grid` directive to filter out items that do not match the current media picker configuration and state. And this is what causes the problem: The `umb-media-grid` directive makes a few failed assumptions about the items passed to it, and thus ends up filtering out user start nodes that aren't on the same level as the first start node encountered.

This whole setup makes very little sense; `umb-media-grid` should simply display the items being passed to it. If any filtering of the items is required, the caller (e.g. the media picker) should be responsible for this. So... this PR moves the filtering to the media picker.

#### ⚠️ Possible breaking change ⚠️

This change may be considered breaking because the built-in filtering functionality of the `umb-media-grid` directive changes. Any 3rd party packages depending on this functionality will have to implement their own filtering logic.

Problem is... we can't fix the failed assumptions in `umb-media-grid` without introducing additional parameters to the directive. As the functionality in question would rely on these parameters it would be a functional breaking change regardless. And then I'd rather opt to clean up `umb-media-grid` and move the filtering to the media picker.

#### The result

With this PR applied, a user with these media start nodes:

image

...is able to navigate the media picker as is expected in #6408:

media-picker-user-start-nodes


poornimanayar:
Hi @kjac ,

Thanks once again!

Poornima

kjac:
PR updated with latest v8/dev

poornimanayar:
Hi @kjac ,

That totally works and I agree with you on what you have mentioned about the directive. Since its a potential breaking change I will let @nul800sebastiaan or someone else from the HQ to have a look before its all approved and merged in :-)

Poornima

kjac:
PR updated with latest v8/dev

poornimanayar:
@nul800sebastiaan May be someone from the HQ can cast an eye on this one as this has a potential breaking change as @kjac  suggests.

nul800sebastiaan:
We'll have a chat about it and see if it's an acceptable breaking change or not! Functionality looks great though! 👍

Issue #6785, opened 2019-10-20
V7: Grid editor config not loading in backoffice


rickliveshere:
I have found that intermittently the backoffice will fail to load grid configuration for a content node when making a call to the /umbraco/GetGridConfig endpoint. Either an empty result is returned, or only configuration for grid editors whose configuration is derived from manifests for plugins from the ~/App_Plugins directory.

This indicates there is a problem either reading the file ~/config/grid.editors.config, or a problem with the file itself (permissions, corrupt) and producing a problem reading it.

### Steps to reproduce

Generate a document type that uses the grid, and add some row configurations that uses a mixture of editors defined in both ~/config/grid.editors.config and in App_Plugins manifest files. In my case, I am using:

- Terratype map grid editor (https://our.umbraco.com/packages/backoffice-extensions/terratype/) for an example from App_Plugins

- Rich Text Editor defined in the ~/config/grid.editors.config file.

Create a new piece of content using this document type and save the result.

Reload the content in the backoffice to force a new call to /umbraco/GetGridConfig

### Expected result

The content node should load in the backoffice, with the grid displaying any populated content correctly.

### Actual result

An error is shown for each grid editor where the call to /umbraco/GetGridConfig has not returned a configuration for how to load that particular editor.

It should be noted that this happens intermittently, and it's not fully understood as to why. I can't seem to replicate this consistently in a local environment, but I appear to have more success when hosting in an Azure App Service.

rickliveshere:
Update: after looking at the ordering of calls in the stacktrace, I've found the system32 path is being set by the Nexu package installed in my project: https://github.com/dawoe/umbraco-nexu/blob/develop-v7/Source/Our.Umbraco.Nexu.Parsers/PropertyParsers/Core/GridParser.cs. I will raise the issue in that repository.

PR #6787, opened 2019-10-20
#6785 - Alters use of Server.MapPath to IOHelper.MapPath

rickread:
Relates to : https://github.com/umbraco/Umbraco-CMS/issues/6785

I put some logging statements in the call to /umbraco/GetGridConfig - and noted that the path for the ~/config directory was being resolved to 'D:\Windows\system32\~\config\grid.editors.config.js'.

This meant that eventually when Umbraco.Core.Configuration.Grid.GridEditorsConfig.Editors is called, the config path is not resolved on line 37 `(File.Exists(gridConfig))`, and the property resolves just those grid editors that have configuration in App_Plugin manifests.

Notably, these issues were only replicated in an Azure App Service as locally the code always resolved to the correct config path. However, I noted that other parts of the code base use Umbraco.Core.IO.IOHelper.MapPath instead of Server.MapPath, with the difference being that the path is being resolved using `HostingEnvironment.MapPath(path)`. Changing to use this instead appears to resolve the correct path.

As to _why_ this is occurring, I'm still not sure. I'm thinking whether it's to do with the /umbraco/GetGridConfig being a Web API call, and `Server.MapPath(path)` being dependent on HttpContext? It should be noted that I'm running the project in .Net 4.7.1 - whether this is generating the issue in itself?

### DIFF
--- a/src/Umbraco.Web/Editors/BackOfficeController.cs
+++ b/src/Umbraco.Web/Editors/BackOfficeController.cs
@@ -239,8 +239,8 @@ namespace Umbraco.Web.Editors
             var gridConfig = UmbracoConfig.For.GridConfig(
                 Logger,
                 ApplicationContext.ApplicationCache.RuntimeCache,
-                new DirectoryInfo(Server.MapPath(SystemDirectories.AppPlugins)),
-                new DirectoryInfo(Server.MapPath(SystemDirectories.Config)),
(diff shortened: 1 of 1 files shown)


rickliveshere:
Update: I've found the issue is being caused by the Nexu package installed in my project by initialising Grid Editor config with a system32 path. I'll raise this as an issue in that repo.

nul800sebastiaan:
👍 OK!

PR #6790, opened 2019-10-21
V8: Update TinyMCE language files

### LINKED ISSUE #6230: V8 : TinyMCE language not correctly set with xx_xx format culture codes
When setting backoffice UI in a language wich as a two part culture code (ex: 'Français (France)' : fr_fr), TinyMCE does not use right language. 

For instance in culture 'Français (France)' the language assigned on TinyMCE init is 'fr' wich sould be 'fr_FR' (https://www.tiny.cloud/docs/configure/localization/) to work right.

stevemegson:
### Prerequisites
If there's an existing issue for this PR then this fixes #6230

### Description
As discovered in #6230, we have some out-of-date TinyMCE language files which are not in the format expected by the current version. This PR takes the latest language pack from https://www.tiny.cloud/docs-4x/configure/localization/#language.


nul800sebastiaan:
Thanks @stevemegson ! Do you know if all of these files will be loaded when you run Umbraco or is it clever and does it only load the 1 required language file?

stevemegson:
It only loads the one it needs. I guess we're including quite a few that can never be loaded in practice, since I think TinyMCE can only ever be configured with a culture that has a matching Umbraco language file. But it seemed easier to have the full set available in case they're useful in the future.

poornimanayar:
Hi @stevemegson ,

So this is what I did to verify this

- Set the UI langauge to fr-FR in web.config
- Changed the language of my logged in back office user to French

Now when I hover over the options in an RTE I can see the titles in French. Is there anything more I need to do here?

Poornima

stevemegson:
@poornimanayar I think that covers it. Previously you'd get English titles as a fallback because it couldn't understand the old language file. You could test some other languages, but if it's working for French then the other files should also work.

poornimanayar:
And merged :-)

PR #6793, opened 2019-10-21
Fix wrong assembly name for UmbracoModule

rasmusjp:
Noticed that the assembly name in the xdt file still was `umbraco` but that assembly doesn't exist anymore and the module is now located in Umbraco.Web`.

It's not causing any problems which I guess is because it's being replaced somewhere else in the powershell install scripts


nul800sebastiaan:
👍

PR #6795, opened 2019-10-21
V8: Embed models builder

bergmania:
#### Changes by Stephane:
- Removed the nuget reference to ModelsBuilder
- Removed dependency to roslyn
- Removed support for custom attributes
- Removed support for modifying models
- Removed support for events
- Removed stuff to support API (`src/Umbraco.ModelsBuilder/Api/**/*`)
- Removed support for dll model (`LiveDll` and `Dll`)

#### Changes by me:
- Updated the `GetModelContentType` and `GetModelPropertyType` from the auto-generated files to return interfaces instead of concrete types.
- Added overload of `GetModelPropertyType` that takes interface instead of concreate type, and obsolete the overload with concrete type.
- Removed unused classes `ClrNameSource` and  `EnumerableExtensions`, such that we should not support these
- Internalized `TextHeaderWriter`
- Fixed build pipeline issues
  - `Microsoft.CodeAnalysis.CSharp` version aligned with the rest of the projects
  - Specified the C# version used in the `Umbraco.ModelsBuilder` project, to `7.3` to support the features used
- Refactored use of `new Exception(...)` to more explicit exception types.
- Removed option to disable ModelsBuilder. Do not make sense for V8 to allow this, and everything is broken if you do that today.

#### Architecture:
- Interfaces to make it possible to replace with a external modelsbuild already exists in `Umbraco.Core`
- Other ModelsBuilder stuff is placed in `Umbraco.ModelsBuilder` that is also added to the `Umbraco.Web` nuget package.


src/Umbraco.ModelsBuilder/Building/TypeModel.cs
```diff
@@ -204,5 +185,21 @@ public IEnumerable<TypeModel> EnumerateBaseTypes(bool andSelf = false)
                 typeModel = typeModel.BaseType;
             }
         }
+
+        /// <summary>
+        /// Maps ModelType.
+        /// </summary>
+        public static void MapModelTypes(IList<TypeModel> typeModels, string ns)
```
Shazwazza: leave this as public or internal?

src/Umbraco.ModelsBuilder/Dashboard/BuilderDashboardHelper.cs
```diff
@@ -1,21 +1,17 @@
...
 
 namespace Umbraco.ModelsBuilder.Dashboard
 {
     internal static class BuilderDashboardHelper
     {
-        public static bool CanGenerate()
-        {
-            return UmbracoConfig.For.ModelsBuilder().ModelsMode.SupportsExplicitGeneration();
-        }
+        private static Config Config => Current.Configs.ModelsBuilder();
```
Shazwazza: inject? means making this not static, or passing this in to the methods, also ideally we'd find a better name for this than "Helper"

src/Umbraco.ModelsBuilder/ReferencedAssemblies.cs
```diff
@@ -31,19 +25,68 @@ static ReferencedAssemblies()
...
+
+            return null;
+        }
+
+        public static Assembly GetNetStandardAssembly()
+        {
+            // in PreApplicationStartMethod we cannot get BuildManager.Referenced assemblies, do it differently
+            try
+            {
+                var someAssembly = Assembly.Load("System.Reflection.Metadata");
```
Shazwazza: code duplication, should prob be shared properly

src/Umbraco.ModelsBuilder/Umbraco/HashHelper.cs
```diff
@@ -6,13 +6,10 @@ namespace Umbraco.ModelsBuilder.Umbraco
 {
     class HashHelper
```
Shazwazza: "Helper", could do with a better name

src/Umbraco.ModelsBuilder/Umbraco/PureLiveModelFactory.cs
```diff
@@ -300,19 +328,12 @@ internal Infos EnsureModels()
 
         private Assembly GetModelsAssembly(bool forceRebuild)
         {
-            var modelsDirectory = UmbracoConfig.For.ModelsBuilder().ModelsDirectory;
+            var modelsDirectory = _config.ModelsDirectory;
             if (!Directory.Exists(modelsDirectory))
                 Directory.CreateDirectory(modelsDirectory);
 
-            // must filter out *.generated.cs because we haven't deleted them yet!
-            var ourFiles = Directory.Exists(modelsDirectory)
```
Shazwazza: I wonder why we are not using this as part of the hash any longer? Seems like an odd change since this is most likely how it is working in the original MB

src/Umbraco.ModelsBuilder/Umbraco/LiveModelsProvider.cs
```diff
@@ -21,15 +21,10 @@ public sealed class LiveModelsProvider
...
-        internal static bool IsEnabled
-        {
-            get
-            {
-                var config = UmbracoConfig.For.ModelsBuilder();
-                return config.ModelsMode.IsLiveNotPure();
-                // we do not manage pure live here
-            }
-        }
+        private static Config Config => Current.Configs.ModelsBuilder();
```
Shazwazza: inject? means less statics

src/Umbraco.ModelsBuilder/Umbraco/LiveModelsProvider.cs
```diff
@@ -133,9 +127,7 @@ public void Dispose()
 
         public static void Install()
         {
-            if (!LiveModelsProvider.IsEnabled)
-                return;
-
+            // always - don't read config in PreApplicationStartMethod
             HttpApplication.RegisterModule(typeof(LiveModelsProviderModule));
```
Shazwazza: This is what we'll most likely have to alter a bit to detect the original MB is active.

src/Umbraco.ModelsBuilder/Umbraco/ModelsBuilderInitializer.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Web;
+using System.Web.Compilation;
+using Umbraco.ModelsBuilder.Umbraco;
+
+[assembly: PreApplicationStartMethod(typeof(ModelsBuilderInitializer), "Initialize")]
+
+namespace Umbraco.ModelsBuilder.Umbraco
+{
+    public static class ModelsBuilderInitializer
```
Shazwazza: doesn't need to be public

src/Umbraco.ModelsBuilder/Umbraco/ModelsGenerationError.cs
```diff
@@ -1,13 +1,15 @@
...
 using System.Text;
-using Umbraco.Core.Configuration;
+using Umbraco.Core.Composing;
 using Umbraco.ModelsBuilder.Configuration;
 
 namespace Umbraco.ModelsBuilder.Umbraco
 {
     internal static class ModelsGenerationError
     {
+        private static Config Config => Current.Configs.ModelsBuilder();
```
Shazwazza: inject?

src/Umbraco.ModelsBuilder/Umbraco/OutOfDateModelsStatus.cs
```diff
@@ -1,19 +1,18 @@
...
+﻿using System.IO;
+using Umbraco.Core.Composing;
 using Umbraco.ModelsBuilder.Configuration;
 using Umbraco.Web.Cache;
 
 namespace Umbraco.ModelsBuilder.Umbraco
 {
     public sealed class OutOfDateModelsStatus
     {
+        private static Config Config => Current.Configs.ModelsBuilder();
```
Shazwazza: inject?

src/Umbraco.ModelsBuilder/Umbraco/PublishedModelUtility.cs
```diff
@@ -40,7 +41,15 @@ public static PublishedContentType GetModelContentType(PublishedItemType itemTyp
             }
         }
 
-        public static PublishedPropertyType GetModelPropertyType<TModel, TValue>(PublishedContentType contentType, Expression<Func<TModel, TValue>> selector)
+        [Obsolete("Use the overload taking IPublishedContentType instead")]
```
Shazwazza: Do we need to obsolete anything here? This is brand new code/APIs that has never been shipped so we shouldn't really need to ship an obsoleted method.

src/Umbraco.ModelsBuilder/Umbraco/PureLiveModelFactory.cs
```diff
@@ -24,14 +23,14 @@
 
 namespace Umbraco.ModelsBuilder.Umbraco
```
Shazwazza: Just noticed this namespace, it's not very pretty and IMO we should change this.

src/Umbraco.ModelsBuilder/Umbraco/PureLiveModelFactory.cs
```diff
@@ -182,14 +197,27 @@ private void RazorBuildProvider_CodeGenerationStarted(object sender, EventArgs e
...
+
+                var modelsDirectory = _config.ModelsDirectory;
+                if (!Directory.Exists(modelsDirectory))
+                    Directory.CreateDirectory(modelsDirectory);
+
+                // clear stuff
+                var modelsHashFile = Path.Combine(modelsDirectory, "models.hash");
+                //var modelsSrcFile = Path.Combine(modelsDirectory, "models.generated.cs");
+                //var projFile = Path.Combine(modelsDirectory, "all.generated.cs");
+                var dllPathFile = Path.Combine(modelsDirectory, "all.dll.path");
```
Shazwazza: This seems odd - not sure if this is intended but since we are supporting PureLive and LiveAppData, there is no "DLL" but there is generated code, need to investigate to see if this is a mistake. Same goes with removing the files from being hashed.

src/Umbraco.ModelsBuilder/Umbraco.ModelsBuilder.csproj
```diff
@@ -110,14 +92,23 @@
     </PackageReference>
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\Umbraco.Abstractions\Umbraco.Abstractions.csproj">
```
Shazwazza: Hrm, this isn't a project ;) I'll fixup

src/Umbraco.ModelsBuilder/Umbraco/LiveModelsProvider.cs
```diff
@@ -133,9 +127,7 @@ public void Dispose()
 
         public static void Install()
         {
-            if (!LiveModelsProvider.IsEnabled)
```
Shazwazza: Seems odd this was removed, i'll update with recent changes. The module should not be enabled if pure live is not enabled.

src/Umbraco.ModelsBuilder/Umbraco/PublishedModelUtility.cs
bergmania: You're right.. We don't..

src/Umbraco.ModelsBuilder/Umbraco/PureLiveModelFactory.cs
bergmania: Do you have any suggestions?

src/Umbraco.ModelsBuilder/Umbraco.ModelsBuilder.csproj
bergmania: whoops.. (I tested on the unicore project, must have forgot to revert that change)

src/Umbraco.ModelsBuilder/Umbraco/PureLiveModelFactory.cs
Shazwazza: I don't believe this should have been removed, I will re-implement these checks in my new branch.

Shazwazza: Actually, on second thought, the reason this is removed is because it's not really including any files since the only files that will exist in this folder that do not end with .generated.cs is just the hash file. So i guess inluding these files is useless. Perhaps this is for when DLL mode is enabled, but it seems not relevant for either purelive appdata or liveappdata.

src/Umbraco.ModelsBuilder/Building/Builder.cs
```diff
@@ -273,20 +201,10 @@ internal void PrepareAmbiguousSymbols()
...
-
-            // only 1 - ensure it matches
-            var found = symbols[0].ToDisplayString();
-            var pos = found.IndexOf('<'); // generic?
-            if (pos > 0) found = found.Substring(0, pos); // strip
-            return found != match; // and compare
+            // cannot figure out is a symbol is ambiguous without Roslyn
+            // so... let's say everything is ambiguous - code won't be
+            // pretty but it'll work
+            return true;
```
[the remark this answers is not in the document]
Shazwazza: I've added notes to this in the code, it just means that the generated csharp will have `global::` syntax for most of it's types.

src/Umbraco.ModelsBuilder/Umbraco/PureLiveModelFactory.cs
Shazwazza: This is how the source is in the orig MB too so it's been working like that for a long time. The reason is because we don't need to take into account these commented out files when checking if models need to be rebuilt and they can be left there for reference just in case.

src/Umbraco.Web/Properties/AssemblyInfo.cs
```diff
@@ -12,7 +12,7 @@
 
 // Umbraco Cms
 [assembly: InternalsVisibleTo("Umbraco.Web.UI")]
-
+[assembly: InternalsVisibleTo("Umbraco.ModelsBuilder.Embedded")]
```
zpqrtbnk: why is this needed?

src/Umbraco.Core/Properties/AssemblyInfo.cs
```diff
@@ -14,6 +14,7 @@
 [assembly: InternalsVisibleTo("Umbraco.Web")]
 [assembly: InternalsVisibleTo("Umbraco.Web.UI")]
 [assembly: InternalsVisibleTo("Umbraco.Examine")]
+[assembly: InternalsVisibleTo("Umbraco.ModelsBuilder.Embedded")]
```
zpqrtbnk: why is this needed?

bergmania: We use `PanicException` + `UrlExtensions.IsClientSideRequest` from Core

src/Umbraco.Web/Properties/AssemblyInfo.cs
bergmania: We use `UmbracoFeatures.Disabled` to removed the embedded modelsbuilder dashboard if we detect the full version.

src/Umbraco.Core/Properties/AssemblyInfo.cs
zpqrtbnk: ok

src/Umbraco.Web/Properties/AssemblyInfo.cs
zpqrtbnk: never mind, found it. this should remove the dashboard and the controller:

    // remove the embedded dashboard
    composition.Dashboards().Remove<ModelsBuilderDashboard>();

    // remove the embedded controller
    composition.WithCollectionBuilder<UmbracoApiControllerTypeCollectionBuilder>()
        .Remove<ModelsBuilderDashboardController>();

warrenbuckley:
## Tested
* Vanilla 8.5 Install with SK & various MB modes
* Tested an 8.4 to 8.5 upgrade (with new & old DLLs)
* Tested by removing old DLL after upgrade
* Tested doing fundamentals course on 8.5

dawoe:
@bergmania 

I know I'm late to the party but....

Where was the RFC for this ? So we as community members who use MB on a daily basis could have given input on how they use it.

With this PR modelsbuilder has become very limited in functionality. 

What was the reasoning behind this ?

Dave

bergmania:
Hi @dawoe

You are able to use the existing (external) models builder with full functionality. You just need to add the nuget package. We detect the dll.

Box-Of-Hats:
Are there any blogs posts explaining more about the shift for the ModelBuilder? I've had a look around and I haven't been able to find any reasoning behind it or how this will affect me as an Umbraco developer

nul800sebastiaan:
Please note that the full version of ModelsBuilder isn't going anywhere, you can keep using that in exactly the same way. 

We talked about this in a bit more detail in the latest product update (scroll to "Umbraco 8.5"): https://umbraco.com/blog/umbraco-product-update-december-18th-2019/

lukehook:
It's mentioned that installing the full version of models builder is still supported and that the .dll is detected. I've just installed the models builder from nuget and now I'm getting an error that Dll mode is on a valid option. What else do I need to do to get the previous models builder to work or is this not possible afterall?

bergmania:
Hi @lukehook 

Can you confirm the `Umbraco.ModelsBuilder.dll` is in your bin folder?

When I try to reproduce, I can't get any error when the `Umbraco.ModelsBuilder.dll`  is in the bin folder. But if I delete this dll - and thereby running on the embedded, I get the following error:
```
System.Configuration.ConfigurationErrorsException: ModelsMode "Dll" is not a valid mode. Note that modes are case-sensitive. Possible values are: Nothing, PureLive, AppData, LiveAppData
```

lukehook:
@bergmania sorry I should have been more specific on when I'm seeing the error. The dll is definitely in the /bin folder. I'm seeing the error whenever I attempt to save a doc type. I can generate the models absolutely find in the ModelsBuilder dashboard. But when I save a doc type I'm seeing the error you are referring to

bergmania:
Thanks @lukehook, now I was able to reproduce 😢

PR #6798, opened 2019-10-21
V8: Reflect inherited permissions in child node menus

### LINKED ISSUE #6788: Unauthorized access when attempting to delete node
I have created a user group with the intention that users of that group can create and delete nodes below a specific node, but cannot delete that node. (ie, cannot delete the parent but can create/delete children)

### Specifics

- doesn't work in umbraco 8.2.0
- doesn't work in umbraco 8.1.4
- did work in umbraco 7.12.4

### Steps to reproduce

- Create a doctype for the root, for example Homepage
- Create a doctype to act as the parent, for example: News List.
- Create a doctype to act as the child, for example: News Item
- Set the News Item to be an Allowed child of News List
- Set the News List to be an Allowed child of Homepage
- Add the Homepage to the content tree, then the News List and add some News Items below it.

- Create a group - for example News Editor

#### Assign access

- For Sections add Content and Media.
- Set Content Start Node to Homepage
- Set Media Start Node to Media Root.

#### Default permissions
- select Browse Node
- select Delete
- select Create
- select Publish

#### Granular permissions

- Select the News List node
- give permissions to Browse Node, Create, Publish

- Create a Test user and assign them to the News Editor group.
- Login as the Test user and try and delete one of the child nodes.

### Expected result

It's expected that the item would be deleted.

### Actual result

When attempting to delete a Child node you get the following error message:

`Authorization error: Unauthorized access to URL: 
/umbraco/backoffice/UmbracoApi/Content/DeleteById
Contact your administrator for information.`

### Why am i doing this?

Users need to right-click (or select the parent node) in order to Create a node beneath it, sometimes users are used to this pattern and accidentally right-click on the parent and choose delete or they did not realize they had the parent node focused.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #6788

### Description

When you assign explicit group permissions to a node, all children inherit these permissions - at least until you break the inheritance with another explicit permission assignment. 

Whereas the content editor reflects these inherited permissions for the children (i.e. it removes the option to publish if publishing is not allowed on the parent), the menus for the children do not. However, if you attempt to perform a menu action that isn't allowed due to restrictions on a parent, you'll get an unauthorized error (as reported in #6788) and an infinite spinning confirmation button:

image

This PR ensures that the permissions used to build the menus are the same that are used when executing the menu actions (and also the same used when rendering the content editor).

In other words: With this PR applied, the parent restrictions are applied in the child node menus:

action-permissions-by-path

*Note: #6788 has it backwards; it does not expect the permissions to be inherited. I did a double check with the latest V7 and this is how permissions work in V7.*


poornimanayar:
Hi @kjac ,

👍 

Poornima

poornimanayar:
Okay @kjac ,

I have tested this and here are my comments and some test notes
- Create a new user group with create, delete, publish and browse permissions
- Assign create, pubish and browse to the "About Us" section and I can no longer delete the parent or child page
- Now, assign only create, publish and browseto "Blog" node in starter kit
- I expect child items not to have the delete option in the list view but I can see that and clicking delete gives me the Unauthorised error. However, if I go into the child node I cannot see the delete option.

**Listview**
image

**Child node**
image


Any questions let me know.

Poornima

kjac:
Hi @poornimanayar 😄 

Sounds like the PR works as intended. This PR doesn't concern the listview options for selected items 👍

poornimanayar:
My bad! @kjac has pointed me in the right direction with the issue around permissions on listview, so this fix is defenitely an improvement over what we have now.

Poornima

poornimanayar:
@kjac , can you resolve the merge conflicts against `v8/contrib`

kjac:
@poornimanayar I sure would, but there aren't any conflicts? 😄

poornimanayar:
hi @kjac,

That’s weird.. I ll try again.

Poornima

poornimanayar:
Hmm, @kjac, I changed the base to `v8/contrib` and now I see a merge conflict.

kjac:
@poornimanayar now I saw the conflicts too 👍 they're resolved now.

poornimanayar:
Merged! Thanks @kjac

PR #6799, opened 2019-10-21
V7: Editors should always be allowed to delete their own content

kjac:
### Prerequisites

### Description

Editors should always be able to delete the content they create - even if deletion is explicitly disallowed by means of group permissions assigned to a (parent) node. That's why the "Delete" action remains in the node menu. 

It just doesn't work, because this special exemption isn't honored when the delete action is executed:

delete-own-content-before

This PR adds this special case handling to the action execution, and thus you're always able to delete the content you have created:

delete-own-content-after

### DIFF
--- a/src/Umbraco.Web/Editors/ContentController.cs
+++ b/src/Umbraco.Web/Editors/ContentController.cs
@@ -28,6 +28,7 @@ using Constants = Umbraco.Core.Constants;
 using umbraco.cms.businesslogic;
 using System.Collections;
 using umbraco;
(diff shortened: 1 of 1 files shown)


poornimanayar:
Hi @kjac ,

You know the drill!

Poornima

elit0451:
Tested and it looks good @kjac :raised_hands: Thanks for your contribution :star2:
Merged!

elit0451:
Cherry picked for 7.15.4 in commit f172760f961218760ce4f9e8dc0e9b7908ad90b0

PR #6800, opened 2019-10-21
V8: Even out the padding in dialog footers

kjac:
### Prerequisites

### Description

Yeah so this is a super minor thing, but it's been bugging me so... here goes: The dialog footer padding is all in the bottom and nothing in the top. It's clearly visible when you have a scrollbar in the dialog.

Besides the lack of spacing above the footer buttons, these buttons are also misaligned with the content editor buttons:

image

As you might have guessed, this PR moves the dialog footer buttons a bit down and to the right to match the spacing of the content editor buttons. This also adds an appropriate amount of spacing above the dialog footer buttons:

image


poornimanayar:
@kjac ,
 👍

abjerner:
@kjac thanks a lot 👍

This fixes the difference between the dialogs (left) and the overlays (right), so your PR is hereby approved.

Somewhat related, but probably something for another PR, there is however also some difference between the dialogs/overlays and the normal buttons in the bottom right corner:

image

kjac:
@abjerner I'm not sure I follow... but there is also https://github.com/umbraco/Umbraco-CMS/pull/6727 which centers around styling the overlays more V8-ish, does that solve it?

abjerner:
@kjac good question 😄 

It don't think it does for this particular case. But I also noticed that the overlay I was checking had the old look and feel, so it probably shouldn't be used for testing.

Anyways - here is a bigger screenshot of what I'm referring to (there is a 2 pixel vertical difference):

image

Therefore I actually think the padding should be `8px` instead of `10px` - here's the result with `8px`:

image

When changed to `8px`, the buttons also align vertically with the image picker overlay as shown here:

image

What I was comparing against initially was the second overlay when inserting an image. The UI here more resembles the UI from Umbraco 7 overlays (which is a whole other issue - it may be what you're fixing in #6727):

image

So my conclusion is that I believe the padding should be `8px` instead of `10px` (although I approved the latter 🙃).

src/Umbraco.Web.UI.Client/src/less/modals.less
```diff
@@ -111,7 +111,7 @@
 	left: 0px;
 	right: 0px;
 	bottom: 0px;
-    padding: 20px;
+    padding: 10px;
```
abjerner: Hence my reply from yesterday, shouldn't the padding be `8px` instead?

kjac: @abjerner you're right. That however left behind a gap between the dialog body and footer (the scroll bar didn't align with the footer), but I believe I got that fixed as well.

Here's a zoomed in view of dialog buttons (left) and content editor buttons (right):

image

nul800sebastiaan:
Wonderful fixing of the details! 👍

PR #6804, opened 2019-10-21
Improvements to grid accessibility

MMasey:
## Description

This PR fixes the following accessibility issues 

- 32 & 60 - "+" icon is identified as "link" - screen readers
- 46 & 54 - Once new row has been added to page grid, unable to navigate to using keyboard

It also fixed an issue with the ItemPickerOverlay where `tab-order` was being added to the list items because the items we using anchor tags rather than button element.

## Steps to test

1. Using the starter them navigate to a doctype making use of the grid property editor e.g. the Home page
2. Tab through the grid options until to reach the Add new row button (+)
3. You should be able to tab to it
4. You should be able to use the keyboard to tab through and select a row type.
5. When you hit enter, it should automatically focus on the newly created row.

Bonus
1. When using the keyboard to navigating through the ItemPicker, it should only take on hit of the tab key to go through each of the options.

## Demo gif

grid-accessibility-improvements


poornimanayar:
Hi @MMasey ,

Thank you for this. I spotted this the other day while I was tweaking around. Well done for getting this fixed 👍 

Regards,
Poornima

poornimanayar:
Works as advertised! Happy to approve @MMasey 👍

PR #6807, opened 2019-10-22
Splits up TemplateUtilities into testable parts, creates new GetReferences API

Shazwazza:
This is one of many PRs that will be for the Media Tracking feature. This is just a start though and I wanted to get this PR reviewed before we start on more of the implementation details since a lot of this PR is just cleanup.

* Splits up `TemplateUtilities` into testable parts and obsoletes it
* Renames some things
* Adds `IDataValueReference` which will be used to resolve references
* Implements IDataValueReference for the RTE as an example
* Adds unit tests
* Fixes the Image source parsing to use the media url provider, not the IPublishedContent.Url result which is actually incorrect

Merging this is fine since it's targeting a main feature branch, not v8/dev


src/Umbraco.Web/Templates/LocalLinkParser.cs
```diff
@@ -0,0 +1,121 @@
...
+using Umbraco.Core.Logging;
+using Umbraco.Web.PublishedCache;
+using Umbraco.Web.Routing;
+
+namespace Umbraco.Web.Templates
+{
+    /// <summary>
+    /// Utility class used to parse internal links
+    /// </summary>
+    public sealed class LocalLinkParser
```
bergmania: I really don't like this name, when it is not related to `Umbraco.Core.Deploy.ILocalLinkParser`. 
Can we find an alternative even that it actually parses local links.. 😕

src/Umbraco.Web/Templates/TemplateUtilities.cs
```diff
@@ -15,16 +14,11 @@
 
 namespace Umbraco.Web.Templates
 {
-    //NOTE: I realize there is only one class in this namespace but I'm pretty positive that there will be more classes in
-    //this namespace once we start migrating and cleaning up more code.
 
-    /// <summary>
-    /// Utility class used for templates
-    /// </summary>
+    [Obsolete("This class is obsolete, all methods have been moved to other classes such as InternalLinkHelper, UrlResolver and MediaParser")]
```
bergmania: MediaParser is named ImageSourceParser.. 

There also is a lot of variables that in named inconsistent now.. Think we need to clean this up, before we continue.

src/Umbraco.Web/Templates/TemplateUtilities.cs
```diff
@@ -15,16 +14,11 @@
...
     public static class TemplateUtilities
     {
-        const string TemporaryImageDataAttribute = "data-tmpimg";
-
+        [Obsolete("Inject and use an instance of InternalLinkParser instead")]
```
bergmania: I guess this is not `LocalLinkParser` ??

src/Umbraco.Web/Templates/TemplateUtilities.cs
```diff
@@ -35,262 +29,28 @@ internal static string ParseInternalLinks(string text, bool preview, UmbracoCont
...
-        }
-
-        /// <summary>
-        /// Removes media urls from &lt;img&gt; tags where a data-udi attribute is present
-        /// </summary>
-        /// <param name="text"></param>
-        /// <returns></returns>
+            => Current.Factory.GetInstance<ImageSourceParser>().EnsureImageSources(text);
+        
+        [Obsolete("Use MediaParser.RemoveImageSources instead")]
```
bergmania: Again this is renamed.. IIRC we can use `nameof(ImageSourceParser)` in locations like this.

src/Umbraco.Web/Templates/TemplateUtilities.cs
```diff
@@ -35,262 +29,28 @@ internal static string ParseInternalLinks(string text, bool preview, UmbracoCont
...
         internal static string RemoveMediaUrlsFromTextString(string text)
-            // see comment in ResolveMediaFromTextString for group reference
-            => ResolveImgPattern.Replace(text, "$1$3$4$5");
+            => Current.Factory.GetInstance<ImageSourceParser>().RemoveImageSources(text);
 
+        [Obsolete("Use MediaParser.RemoveImageSources instead")]
```
bergmania: same as above

Shazwazza:
Well, naming is hard.

Yes - we have the same names as the Deploy interfaces - because i just made them consistent, since they basically do the same thing, except the deploy implementations return 'artifacts' instead.

I will spend time today thinking about names ...

Shazwazza:
Have pushed new changes, i didn't go crazy with trying to think of some new zany names but did make them more specific. Happy to keep thinking if you still don't like them. They are now:

* HtmlImageSourceParser
* HtmlLocalLinkParser
* HtmlUrlParser

Because, these exist in the Umbraco.Web.Templates namespace which essentially has to do with working templating (and html), and all of these parsers are specifically working with HTML content. Alternatively, could have them all suffixed with "Translator" instead of "Parser" ? ... i dunno.

I've renamed all variables and comments to be consistent.

bergmania:
I'm happy with the new names.. :)

Jeavon:
There are some method signature changes in this PR that should be listed as breaking changes for v8.6. For example, the RteMacroRenderingValueConverter now has 3 new parameters. It might be that it's covered by https://github.com/umbraco/Umbraco-CMS/pull/7303 which is already listed as a breaking change...

Shazwazza:
For ctors for injected objects we will also need to add that to the breaking change policy, as APIs evolve ctors will change. Listing as breaking is still a good thing though for info purposes.

PR #6815, opened 2019-10-22
V8: Prevent/remove duplicate nodes for editors with start nodes (take two)

kjac:
### Prerequisites

### Description

*This is a redo of #6687 based on the comments from @nielslyngsoe*

There's something weird going on when editors have start nodes; the same content nodes are being loaded twice, and things generally feels broken.

Consider an editor with these start nodes configured:

image

Here's the corresponding editor experience:

editor-with-start-nodes-before

Personally I'd be genuinely confused about the duplicate nodes - even more so because the the root nodes isn't duplicated. And also:

- When you select a node in the tree that is built from your start node, the corresponding node underneath the "full" tree is highlighted instead of the one you selected.
- When you right click a node anywhere, both that node and the corresponding duplicate are highlighted.
- All nodes in the "full" tree have a "not-allowed" cursor and are dimmed down - even the ones you can actually edit.

This PR removes the duplicate start nodes, leaving behind only the "full" tree. This I believe is also the behavior in V7.

Here's the editor experience for the same setup with this PR applied:

editor-with-start-nodes-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less
@@ -289,9 +289,8 @@ body.touch .umb-tree {
 }
 
 .no-access {
-    .umb-tree-icon,
-    .root-link,
-    .umb-tree-item__label {
+    > .umb-tree-item__inner .umb-tree-icon,
+    > .umb-tree-item__inner .umb-tree-item__label {
         color: @gray-7;
         cursor: not-allowed;
     }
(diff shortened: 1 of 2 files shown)


nielslyngsoe:
@kjac lovely, looks great — I haven't tested jet, but from the GIF it looks like the right solution.

nul800sebastiaan:
This looks great indeed! Does it apply to media as well @kjac?

joepvtl:
Hey @kjac ,

Is there a quick fix for this I can use? Or do I need to wait for a new version?

-Joep

kjac:
@joepvtl you can't patch an existing version with this PR I'm afraid. But now it's been approved I'll merge it in and then it should be included in 8.6

LCichanowiczPMG:
I do not see this issue listed in [the release notes for v8.6.0](https://our.umbraco.com/download/releases/860).  Is it part of the 8.6.0 release?

nielslyngsoe:
@LCichanowiczPMG Thanks for letting us know. The GitHub labels are used for release notes. I really don't know why it was removed. But thanks to you I have added it again.

PR #6817, opened 2019-10-23
Dim down the alias of locked/inherited properties

kjac:
### Prerequisites

### Description

The aliases of inherited and locked properties ought to be dimmed down with a "not allowed" hover cursor, but they aren't (you can't see the cursor from these screenshots though):

image

image

This PR updates the styling, so it's evident you can't mess around with the aliases of these properties (again... you can't see the hover cursor, but it's supposed to be "not allowed").

image

image


nul800sebastiaan:
Nice one, makes sense to me! 👍

PR #6821, opened 2019-10-24
V8: Make "add property" shortcut work with multiple groups

kjac:
### Prerequisites

### Description

If you have multiple groups in the content type builder, the "add property" shortcut (alt+shift+p) only works for the last group:

add-property-shortcut-before

This is because the shortcut is registered for every group and the last registration simply wins.

This PR moves the shortcut registration outside the property groups and adds logic to figure out which group is currently targeted the "add property" action. It works like this:

add-property-shortcut-after


emmaburstow:
@kjac... your work here just gets better and better. Subtitles on the key strokes.... Genius. 

em

nul800sebastiaan:
Oh that's a bit silly, yes, that was always the intention. This works great, cheers!

Issue #6638, opened 2019-10-10
Enable copy/paste for grid items


skttl:
Just like nested content, it could be very useful to have the ability to copy and paste grid items using the clipboardService.

nul800sebastiaan:
Yes, we're on the road to enabling copy/paste for anything that makes sense. 👍

skttl:
Is it something that you are working on in HQ, or is it open for PRs?

nul800sebastiaan:
We still have some things to do at HQ before we can ask for contributions, give @nielslyngsoe a few more weeks to make this a nice experience please! :-)

skttl:
Thanks :)

nielslyngsoe:
You can follow the progress on this issue: https://github.com/umbraco/Umbraco-CMS/issues/6816

I will keep this issue open, since it requests the copy feature specifically for the Grid.

skttl:
@nielslyngsoe I was thinking about copying specific controls in the grid. Your PR adds the ability to copy the complete value - which also is very nice.

nielslyngsoe:
@skttl Oh, I see, my mistake. But good, we could use that as well.
Can I ask you to change the description to be more explicit in terms of copy/pasting individual items. Thanks

nielslyngsoe:
Reopen issue, still very relevant

skttl:
This feature needs to emit events, so custom editors can react on being copied. As an example, Doc Type Grid Editor stores a guid key in each editor, and this needs to be regenerated when copied for this to work correctly.

lars-erik:
Here's what I did to enable events in the grid editor:  
https://github.com/umbraco/Umbraco-CMS/pull/2640

#h5yr all. Love to see this finally happen.

nielslyngsoe:
Hi @skttl 

It's up to the Custom Editor to implement the copy functionality and as well the paste functionality — thereby to ensure that it's done the right way.

Currently, We suggest removing unique data when making a copy, and ensuring keys are begin generated when pasting.

It's not enough to generate new keys for the copied data. Since the same data could be pasted multiple times.

I hope it makes sense, let us know if you see any potential problems with the current strategy or implementation.

Thanks

skttl:
Oh sorry, meant pasted of course :)

I'm not sure how to implement this in grid, as it is the grid that takes care of copying/pasting as I see it. I guess the copy function should be in the blue title bar on each grid control. 

Which was why I suggested an event you can subscribe to.

[...]

PR #6823, opened 2019-10-24
PropertyActions + Copy all entries for Nested Content

nielslyngsoe:
Implementation of PropertyActions plus prove of concept by adding "Copy all entries" for Nested Content.

For more details on this matter please read the issue:
fixes https://github.com/umbraco/Umbraco-CMS/issues/6638

image

**Test notes:**
Prepare test by creating two ElementTypes, called ElementType A and ElementType B.
Then create a documentType with three nested content properties. Two based on the same DataType that allows both ElementType A and ElementType B to be inserted. And a different DataType that only allows for ElementType A.

— Test that Property Actions can be accessed by mouse.
— Test that Property Actions can be accessed by keyboard.
— Create two different elementTypes in the Nested Content property.
— Test that the Copy action can be executed on the Nested Content property.
— Test that the copied entries can be inserted/pasted into the same nested content property.
— Test that the copied entries can be inserted/pasted into the same nested content property in split-view.
— Test that the copied entries can be inserted/pasted into the same nested content property in a new document.
— Test that the copied entries can be inserted/pasted into another nested content property.
— Test that the copied entries can be inserted/pasted into another nested content property configured with a different data-type.
— Test that the copied entries cannot be inserted/pasted into another nested content property that does not accept one or more of the ElementTypes copied.
— Test a copy of a nested content property that only contains elementType A can be copied over to nested content property that only allows for elementType B.


nielslyngsoe:
We have now agreed on a solution, and the current state of the UI can be seen here:

image

src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbproperty.directive.js
```diff
@@ -32,6 +35,20 @@ angular.module("umbraco.directives")
...
                     $scope.property.propertyErrorMessage = errorMsg;
                 };
+                
+                var unsubscribe = $scope.$on("ExposePropertyEditorAPI", function(event, api) {
+                    
+                    //avoid eventual parent properties to capture this.
+                    event.stopPropagation();
+                    
+                    $scope.propertyEditorAPI = api;
+                    //$scope.propertyActions = api.propertyActions;
```
warrenbuckley:
```suggestion
```
Lets not leave commented code in @nielslyngsoe

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
```diff
@@ -1,203 +1,276 @@
...
+    *
+    * @param {string} type A string defining the type of data to storing, example: 'elementType', 'contentNode'
+    * @param {string} alias A string defining the alias of the data to store, example: 'product'
+    * @param {object} entry A object containing the properties to be saved, this could be the object of a ElementType, ContentNode, ...
+    * @param {string} displayLabel (optional) A string swetting the label to display when showing paste entries.
+    *
+    * @description
+    * Saves a single JS-object with a type and alias to the clipboard.
+    */
+    service.copy = function(type, alias, data, displayLabel) {
```
warrenbuckley: Method/function arguments changed do we know if anyone used this in their own packages as this would be considered a breaking change @nielslyngsoe

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
```diff
@@ -1,203 +1,276 @@
...
+            }
+        );
+        
+        var entry = {unique:uniqueKey, type:type, alias:alias, data:prepareEntryForStorage(data), label:displayLabel || data.name, icon:iconHelper.convertFromLegacyIcon(data.icon)};
+        storage.entries.push(entry);
+        
+        if (saveStorage(storage) === true) {
+            notificationsService.success("Clipboard", "Copied to clipboard.");
+        } else {
+            notificationsService.success("Clipboard", "Couldnt copy this data to clipboard.");
```
warrenbuckley: Is this a bit confusing/misleading UX to give a green/success message when an error happened @nielslyngsoe ?

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
```diff
@@ -1,203 +1,276 @@
...
+        // remove previous copies of this entry:
+        storage.entries = storage.entries.filter(
+            (entry) => {
+                return entry.unique !== uniqueKey;
+            }
+        );
+        
+        var entry = {unique:uniqueKey, type:type, aliases:aliases, data:copiedDatas, label:displayLabel, icon:displayIcon};
+
+        console.log(entry);
```
warrenbuckley: Do we still need this console.log @nielslyngsoe ?

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
```diff
@@ -1,203 +1,276 @@
...
+    * @param {string} aliases A array of strings providing the alias of the data you want to remove.
+    * 
+    * @description
+    * Removes entries matching the given type and one of the provided aliases.
+    */
+    service.clearEntriesOfType = function(type, allowedAliases) {
+        
+        var storage = retriveStorage();
+        
+        console.log("BEFORE: ", storage.entries)
```
warrenbuckley: Same do we need debug/testing stuff?

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
```diff
@@ -1,203 +1,276 @@
...
+                        || 
+                        (entry.aliases && entry.aliases.filter(entryAlias => allowedAliases.filter(allowedAlias => allowedAlias === entryAlias).length > 0).length > 0)
+                    )
+                );
+            }
+        );
+        
+        storage.entries = filteretEntries;
+        
+        console.log("AFTER: ", storage.entries)
```
warrenbuckley: Ditto - more console.logs

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
nielslyngsoe: No worries

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
nielslyngsoe: Yep, should be changed

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
nielslyngsoe: No, thats a mistake

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
nielslyngsoe: They are gone in a shift move.

warrenbuckley:
Updated code reads fine, off to re-test :)

warrenbuckley:
OK found a new bug/issue @nielslyngsoe 

* Set up two nested content properties that Allow A & B element types
* One that has no max items
* The other with say 3 max items
* Copy 4 or more items from nested content

It does not remove or restrict the paste and the data is saved without errors/validation kicking in.

warrenbuckley:
Also I have this current UI bug (not sure if related or a recent merge conflict or....?)
image

nul800sebastiaan:
Hmm, was just double checking I didn't miss something while merging as I can repro Warren's the media picker problem. It seems localized to this PR as that works find on `v8/dev` and I just merged `v8/dev` into this PR.

nielslyngsoe:
> OK found a new bug/issue @nielslyngsoe
> 
> * Set up two nested content properties that Allow A & B element types
> * One that has no max items
> * The other with say 3 max items
> * Copy 4 or more items from nested content
> 
> It does not remove or restrict the paste and the data is saved without errors/validation kicking in.

Dada! 😧 We didn't consider this, good catch!
There is a solution for it ready now.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
```diff
@@ -208,6 +211,39 @@ angular.module("umbraco").controller("Umbraco.PropertyEditors.NestedContent.Prop
...
+            syncCurrentNode();
+
+            // list aliases
+            var aliases = $scope.nodes.map((node) => node.contentTypeAlias);
+
+            // remove dublicates
+            aliases = aliases.filter((item, index) => aliases.indexOf(item) === index);
+
+            // Retrive variant name
+            var culture = $routeParams.cculture ? $routeParams.cculture : $routeParams.mculture;
```
Shazwazza: This isn't a safe way to get the current culture. We should never rely on the $routeParams except at the top level. This won't work when it's in split view for example. To get the current culture requires a component model, for example see the `valPropertyMsg` directive requires `^^umbVariantContent` ... which exposes the current editor's language (i.e. `umbVariantContent.editor.content`). This is the only way to get the current editors culture when in split view since the routeParams are not sufficient for that and should generally be avoided entirely unless at the highest editor level. This does mean that the nested content controller/editor should be adjusted to be a component. Components can now be stored anywhere so a component for nested content can just exist beside this controller/view it doesn't have to go to the weird components folder.

src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbproperty.directive.js
```diff
@@ -32,6 +35,19 @@ angular.module("umbraco.directives")
                 self.setPropertyError = function (errorMsg) {
                     $scope.property.propertyErrorMessage = errorMsg;
                 };
+                
+                var unsubscribe = $scope.$on("ExposePropertyEditorAPI", function(event, api) {
+                    
+                    //avoid eventual parent properties to capture this.
+                    event.stopPropagation();
+                    
+                    $scope.propertyEditorAPI = api;
```
Shazwazza: The name `propertyEditorApi` is too generic and doesn't really explain what this is doing. This is really completely specific to "Property Actions" (if that is the term we are going with)

src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
nielslyngsoe: I think this argument makes it clear that we should go with the Component Solution. I'm on it!

PR #6824, opened 2019-10-24
Create Interface for Member and Content Index to use in Populators

### LINKED ISSUE #6705: ContentPopulators are using UmbracoContentIndex
@Shazwazza 
Actually, I look into the source code of Umbraco and I am trying to figure out why IndexPopulators are using UmbracoContentIndex as the type to discover instead of some Interface which could be inherited by external indexes with a different type than Lucene index? As UmbracoContentIndex inherits from  UmbracoExamineIndex, which is based on LuceneIndex, IUmbracoIndex, IIndexDiagnostics. Actually I would probably propose to create IUmbracoContentIndex which will allow using build-in Content Populators instead of push to create own. It will make easier for future developers which want to create custom Examine Provider for Umbraco.

bielu:
### Prerequisites

If there's an existing issue for this PR then this fixes #6705 

### Description
In Following of conversation with Shannon I prepared interfaces to use in Populators instead of using Lucene base type.
@Shazwazza as promised :)

### DIFF
--- a/src/Umbraco.Examine/ContentIndexPopulator.cs
+++ b/src/Umbraco.Examine/ContentIndexPopulator.cs
@@ -15,7 +15,7 @@ namespace Umbraco.Examine
     /// <summary>
     /// Performs the data lookups required to rebuild a content index
     /// </summary>
-    public class ContentIndexPopulator : IndexPopulator<UmbracoContentIndex>
+    public class ContentIndexPopulator : IndexPopulator<IUmbracoContentIndex>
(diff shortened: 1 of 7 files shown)


emmaburstow:
Thanks for working on this @bielu 

We'll take a look and let you know if we're happy with the solution as is. Otherwise, we'll let you know if there are any changes you could make.

Emma

Shazwazza:
Great stuff, looks good :)

PR #6826, opened 2019-10-24
V8: Performance boost for loading and saving Nested Content

kjac:
### Prerequisites

### Description

The more items you add to a Nested Content property, the longer the containing page takes to load and save. Here's a comparison between an empty Nested Content property and one with some items:

nested-content-performance-before

This PR optimizes the way `NestedContentPropertyValueEditor` fetches the element type blueprints of the contained items. This yields a surprisingly high performance gain:

nested-content-performance-after

To validate this I created a silly little profiling session within the save operation of `ContentController`. Here's the time spent collecting Nested Content property values before this PR is applied:

image

...and here is after:

image

That's something! And that's only for saving. A similar gain can be found in both the validation and the load operations.


emmaburstow:
Ooo nice! Even a little improvement will help. Especially if you have a lot of items. 

Emma

nul800sebastiaan:
Love it!

Issue #6827, opened 2019-10-24
FileUploadCleanupFilterAttribute contains mixed references to Request/Response


ronaldbarendse:
The `FileUploadCleanupFilterAttribute` seems to contain a bug, as it first null-checks the `actionExecutedContext.Request.Content` and then uses `actionExecutedContext.Response.Content`:

https://github.com/umbraco/Umbraco-CMS/blob/5cfdfd6e9cb96ffba1f9b0f0c21041683037262b/src/Umbraco.Web/WebApi/Filters/FileUploadCleanupFilterAttribute.cs#L68-L94

Later on, the warning messages refer to the request content, so it seems that's what it should be using:

https://github.com/umbraco/Umbraco-CMS/blob/5cfdfd6e9cb96ffba1f9b0f0c21041683037262b/src/Umbraco.Web/WebApi/Filters/FileUploadCleanupFilterAttribute.cs#L139-L147

The logging is also quite chatty with all the warnings... And all the nested conditions could be flattened by first validating all values.

Shazwazza:
The reason this is all wierd is because in previous versions of .net and mvc there was/is all sorts of problems with the async nature of what is going on here. For the most part this was a bug in one of the .net framework versions. There's some huge threads about this in numerous places (probably on the old tracker). IIRC part of the problem was that during this processing the HttpContext or something like that would 'disapear'

ronaldbarendse:
But does that still apply for .NET Framework 4.7.2 and ASP.NET MVC/Web API 5? And regardless of that: it's still null-checking something that isn't used, plus the warning messages (in the second snippet) are wrong.

I've found the following related issues on the old tracker:
- https://issues.umbraco.org/issue/U4-4049
- https://issues.umbraco.org/issue/U4-6334
- https://issues.umbraco.org/issue/U4-7082

Scanning through them, I see you've added all the logging because of weird `NullReferenceException`s, but that's probably because it's not correctly validating the `actionExecutedContext.Response.Content` in the first place. It's also not using the `HttpContext`, only the `HttpActionExecutedContext`.

Shazwazza:
[...]

PR #6829, opened 2019-10-24
Fix FileUploadCleanupFilterAttribute

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6827.

### Description
The `FileUploadCleanupFilterAttribute` only did null-checks on `actionExecutedContext.Request.Content`, but then used `actionExecutedContext.Response.Content`.

This fixes all the checks and:
- Removes redundant logging: if there's nothing we can do, just skip cleaning up the uploaded files (the log messages do not provide real help anyway)
- Check all action arguments for `IHaveUploadedFiles` models (not only the first)
- Remove unused folder tracking

We could use the `actionExecutedContext.Response.TryGetContentValue<T>()` extension method instead of getting the `ObjectContent` and then the `Value`, but the current logic will probably be exactly the same...

When testing, the following should still work:
- Upload of media, media on content, local package install, document type import, user avatar (current or other), member import and blueprint/content template import
- Temporary files should be cleaned up (`App_Data\Temp\FileUploads`)

### DIFF
--- a/src/Umbraco.Web/WebApi/Filters/FileUploadCleanupFilterAttribute.cs
+++ b/src/Umbraco.Web/WebApi/Filters/FileUploadCleanupFilterAttribute.cs
@@ -6,7 +6,6 @@ using System.Net.Http;
 using System.Web.Http.Filters;
 using Umbraco.Core;
 using Umbraco.Core.Logging;
-using Umbraco.Core.Models;
 using Umbraco.Web.Composing;
 using Umbraco.Web.Models.ContentEditing;
 using File = System.IO.File;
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web/WebApi/Filters/FileUploadCleanupFilterAttribute.cs
```diff
@@ -1,149 +1,85 @@
...
+                    // Cleanup any files associated
+                    foreach (var uploadedFile in haveUploadedFiles.UploadedFiles)
+                    {
+                        try
+                        {
+                            File.Delete(uploadedFile.TempFilePath);
+                        }
+                        catch (Exception ex)
+                        {
+                            Current.Logger.Error<FileUploadCleanupFilterAttribute>(ex, "Could not delete temp file {FileName}", uploadedFile.TempFilePath);
```
ronaldbarendse: Should this be logged as an error? Maybe a warning in more appropriate?

src/Umbraco.Web/WebApi/Filters/FileUploadCleanupFilterAttribute.cs
```diff
@@ -1,149 +1,85 @@
...
-                        }
+                    try
+                    {
+                        File.Delete(uploadedFile.TempFilePath);
                     }
-                    else
+                    catch (Exception ex)
                     {
-                        Current.Logger.Warn<FileUploadCleanupFilterAttribute>("The actionExecutedContext.Request.Content.Value is not IHaveUploadedFiles, it is {ObjectType}", objectContent.Value.GetType());
+                        Current.Logger.Error<FileUploadCleanupFilterAttribute>(ex, "Could not delete temp file {FileName}", uploadedFile.TempFilePath);
```
ronaldbarendse: Same here: probably better to log this as a warning...

src/Umbraco.Web/WebApi/Filters/FileUploadCleanupFilterAttribute.cs
```diff
@@ -1,149 +1,85 @@
...
                     }
-                }
-                else
-                {
-                    Current.Logger.Warn<FileUploadCleanupFilterAttribute>("The actionExecutedContext.Request.Content is not ObjectContent, it is {RequestObjectType}", actionExecutedContext.Request.Content.GetType());
+
+                    // Clear out the temp path, so it's not returned in the response
+                    uploadedFile.TempFilePath = null;
```
ronaldbarendse: Clearing out should set this to the initial value of `null` (same as when initiating a new `ContentPropertyFile`).

emmaburstow:
Morning @ronaldbarendse,

Thanks for the PR. We'll take a look at your questions and get an answer to you soon. 

Emma

PR #6832, opened 2019-10-24
V8: Don't YSOD when publishing content if a property editor has been removed

kjac:
### Prerequisites

### Description

If you remove custom property editors that are in use on one or more content types, content of these types will replace the missing property editor with a Label property editor that displays the raw property value. This works really well - until you attempt to publish the content:

content-with-removed-pe-before

This is particularly annoying when you're building custom editor experiences in different branches against the same database.

Turns out the validation routine doesn't expect the property editor to be missing... so this PR adds some defensive programming to counter this scenario. When applied, properties with missing property editors always pass validation, and thus things start working again:

content-with-removed-pe-after

#### Testing this PR

Here's a simple custom property editor that can be used for testing: [TestPropertyEditor.zip](https://github.com/umbraco/Umbraco-CMS/files/3768784/TestPropertyEditor.zip)

Steps:

1. Add the property editor to `App_Plugins`.
2. Create a content type that uses this property editor.
3. Create some content based on this content type. 
4. Remove the property editor from `App_Plugins` and restart the site.
5. Verify that you can publish content that contains the now missing property editor.

### DIFF
--- a/src/Umbraco.Core/Services/PropertyValidationService.cs
+++ b/src/Umbraco.Core/Services/PropertyValidationService.cs
@@ -131,6 +131,12 @@ namespace Umbraco.Core.Services
         private bool IsPropertyValueValid(PropertyType propertyType, object value)
         {
             var editor = _propertyEditors[propertyType.PropertyEditorAlias];
+            if (editor == null)
+            {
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
I like it! 👍

PR #6833, opened 2019-10-24
V8: Set focus on content name when opening content for editing

kjac:
### Prerequisites

### Description

If you're a big fan of tabbing around in the UI, it's pretty annoying that the input focus remains in the content tree after opening content for editing (or anything else for that matter):

editor-focus-on-load-before

I would expect the input focus to shift to the content editor, to allow for quick tabbing down through the content. So this PR ensures that input focus is applied to the content item name on load:

editor-focus-on-load-after


nul800sebastiaan:
Awesome, this is definitely a must-have for us keyboard enthusiasts! Thanks!

PR #6834, opened 2019-10-24
V8: Limit the number of tabbable elements in tree nodes

kjac:
### Prerequisites

### Description

The number of elements you have to tab your way through in order to navigate down a tree is really quite excessive... and the most prominent looking one of those elements (the outermost node container) has no action attached, despite it looking like it would actually open the node for editing:

tab-tree-before

I have removed the tab index from a few of these elements to allow for speedier navigation through the tree, without compromising with the functionality at hand. Here's how it works:

tab-tree-after


nul800sebastiaan:
Fantabulous! _See what I did there? 😏_

PR #6835, opened 2019-10-24
Replace old checkbox markup with umb-checkbox

BatJan:
### Prerequisites

### Description
I have changed the good ol' standard checkbox in the media pickers "Include subfolders in search" to use the umb-checkbox directive instead.

**Before**
media-picker-checkbox-before

**After**
media-picker-checkbox-after


nul800sebastiaan:
Beauty!

PR #6838, opened 2019-10-24
Document type collection: Refactor labels to reference the proper input fields

BatJan:
### Prerequisites

### Description
This PR is hard to visualize since the label looks the exact same as before. But the difference is that it has a reference to the id of the input field it's attached too, which improves accessibility and makes the input field focused whenever the label is clicked.

Let me know if I need to ellaborate further on this 😃


elit0451:
Cool change @BatJan :ok_hand:
Tested the functionality :star2: Merging!

PR #6839, opened 2019-10-24
Template render section: Replace checkbox with umb-checkbox

BatJan:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
I have replaced the old checkbox with the umb-checkbox directive.

**Before**
template-sections-before

**After**
template-sections-after

PR #6841, opened 2019-10-24
Template insert: Replace checkbox with umb-checkbox and remove wrong label

BatJan:
### Prerequisites

### Description
I have changed the old checkbox to make use of the umb-checkbox instead.

Also I have changed a `<label>` to `<span>` since the label only contains a text, which also improves the visuals imho.

**Before**
template-insert-before

**After**
template-insert-after


bjarnef:
@BatJan I have already submitted a PR :)
https://github.com/umbraco/Umbraco-CMS/pull/6531

emmaburstow:
Another double! We'll take a look at both and let you know what happens next 🙊

BatJan:
Same comment as on #6839 @emmaburstow 😃

nul800sebastiaan:
Same closing reason as on #6839 then! ;-)

PR #6843, opened 2019-10-25
Media tracking prototype

Shazwazza:
This is still part of a larger work in progress for media (and other entity) tracking.

Changes in this PR (work will continue in another branch):

* New Relation Types including migrations
* Changes to Relation Types - which now allow NULL values for child/parent entity types - these values are not actually used for anything. Well, they were but unecessarily, they were used to resolve the IEntity for an ID and the entity type which was silly since we can just get the entity type for the ID to look it up ... which we do now. The only API 'breaking changes' is that the `Guid` is now `Guid?` for parent/child object types on IRelationType and the RelationType ctor has been updated to be useful, it was not really working before
* IRelation now returns the actual parent/child object types
* No need to keep allocating new `RelationFactory`, all mappings are just static
* Content, media and member repositories now all auto-track relations based on the relations returned by property editors
* Fixes the relation type angular editor, before it was hacking the planet to try to resolve the relations for that type since apparently the c# implementation was simply not finished. So it all works now without hacking things and it supports nullable child/parent object types
* Some code has been moved around
* A few TODO notes added which well need to review when we're further along with the prototype
* Unit tests added

## Testing

Only the RTE has been updated to return references right now, so you can have any content item with an RTE, add some images to it via the image picker and create some internal links to other documents, save it, then go see that relation type (Related Media and Related Documents) unders settings and click on the 'relations' app and it will show you the relations created. You can also remove/change those items in the RTE and those relations will be removed/updated.


src/Umbraco.Core/PropertyEditors/DataEditor.cs
```diff
@@ -19,6 +19,7 @@ namespace Umbraco.Core.PropertyEditors
     public class DataEditor : IDataEditor
     {
         private IDictionary<string, object> _defaultConfiguration;
+        private IDataValueEditor _nonConfigured;
```
bergmania: I don't understand the variable name, why not just `_valueEditor` or `_dataValueEditor`?

src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
```diff
@@ -798,5 +818,65 @@ public virtual IEnumerable<TEntity> GetRecycleBin()
...
+                var refs = reference.GetReferences(val);
+                trackedRelations.AddRange(refs);
+            }
+
+            //First delete all auto-relations for this entity
+            RelationRepository.DeleteByParent(entity.Id, Constants.Conventions.RelationTypes.AutomaticRelationTypes);
+
+            if (trackedRelations.Count == 0) return;
+
+            var udiToGuids = trackedRelations.Select(x => x.Udi as GuidUdi)
```
bergmania:
```suggestion
            var udiToGuids = trackedRelations.Select(x => x.Udi as GuidUdi).Distinct()
```

We need to district here, otherwise ToDictionary throws if I have multiple relations to the same item

I tested this by adding both a link to an image and the image itself into the RTE

bergmania:
When I add a link to a media item, it is put under "Related documents". I would expect it as related media.

image

src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
bergmania: I see we actually need to distinct on trackedRelations, otherwise we have `A duplicate value cannot be inserted into a unique index` error from the database

src/Umbraco.Web/PropertyEditors/RichTextPropertyEditor.cs
```diff
@@ -131,15 +136,19 @@ public override object FromEditor(Core.Models.Editors.ContentPropertyData editor
             /// </summary>
             /// <param name="value"></param>
             /// <returns></returns>
-            public IEnumerable<Udi> GetReferences(object value)
+            public IEnumerable<UmbracoEntityReference> GetReferences(object value)
             {
                 var asString = value == null ? string.Empty : value is string str ? str : value.ToString();
 
+                //TODO: FindUdisFromDataAttributes will return UDIs of any type found, typically these will always be "media" but
```
bergmania: We basically needs a mapping from `Udi.EntityType` into the correct `Constants.Conventions.RelationTypes.*` Maybe with a fallback to something defined outside.

bergmania: What is the reason to explicit set the `RelationTypes` instead of using the info from the `udi`? And only take the UDI as input parameter in the ctor of  `UmbracoEntityReference`

src/Umbraco.Core/Models/Relation.cs
```diff
@@ -17,12 +17,37 @@ public class Relation : EntityBase, IRelation
         private IRelationType _relationType;
         private string _comment;
 
+        /// <summary>
+        /// Constructor for constructing the entity to be created
+        /// </summary>
+        /// <param name="parentId"></param>
+        /// <param name="childId"></param>
+        /// <param name="parentObjectType"></param>
```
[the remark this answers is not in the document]
Shazwazza: I haven't used any GUID for relations because that is not how they are actually related in the DB. If we use GUID then you will need to lookup the INT as an extra DB call.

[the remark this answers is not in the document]
Shazwazza: But yes, the code comments are wrong - sorry VS 2019 hasn't got there yet to warn me since i have r# turned off

src/Umbraco.Core/Persistence/Dtos/RelationTypeDto.cs
```diff
@@ -9,7 +9,7 @@ namespace Umbraco.Core.Persistence.Dtos
     [ExplicitColumns]
     internal class RelationTypeDto
     {
-        public const int NodeIdSeed = 4;
+        public const int NodeIdSeed = 10;
```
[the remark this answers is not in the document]
Shazwazza: That's necessary because we are adding more relation dtos during install which exceed the default Seed value. If we don't change this, then the next relation type that is added is going to throw an exception since the Seed will have been configured lower than what is now available. It's one of those annoying 'gotchas' if we are adding more data during the default installation to watch out for.

src/Umbraco.Core/Persistence/Repositories/Implement/DocumentRepository.cs
```diff
@@ -468,6 +484,8 @@ protected override void PersistNewItem(IContent entity)
                 ClearEntityTags(entity, _tagRepository);
             }
 
+            PersistRelations(entity);
```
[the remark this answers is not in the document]
Shazwazza: I don't think we should be using an event at all since we aren't doing that anywhere else with the repositories in the code base. I've decided to explicitly call this base method from each inherited repository just to make it explicit, however, we could easily add this to the base classes methods of `PersistUpdatedItem` and `PersistNewItem` to have this automatically taken care of. I just thought for now to make it explicit but we can change if we like.

PR #6847, opened 2019-10-25
Request datatype by guid

bjarnef:
### Prerequisites

### Description
At the moment it is possible to get content, media and member by id, but also by guid.

For datatypes it request datatype by id and if you try to request it by guid, it returns a console error.

chrome_2019-10-23_00-32-02

chrome_2019-10-23_00-32-54

### DIFF
--- a/src/Umbraco.Web/Editors/MacrosController.cs
+++ b/src/Umbraco.Web/Editors/MacrosController.cs
@@ -112,19 +112,7 @@ namespace Umbraco.Web.Editors
                 return this.ReturnErrorResponse($"Macro with id {id} does not exist");
             }
 
-            var macroDisplay = Mapper.Map<IMacro, MacroDisplay>(macro);
-
-            var parameters = macro.Properties.Values
-                                .OrderBy(x => x.SortOrder)
-                                .Select(x => new MacroParameterDisplay()
-                                {
-                                    Editor = x.EditorAlias,
-                                    Key = x.Alias,
-                                    Label = x.Name,
-                                    Id = x.Id
-                                });
-
-            macroDisplay.Parameters = parameters;
+            var macroDisplay = MapToDisplay(macro);
 
             return this.Request.CreateResponse(HttpStatusCode.OK, macroDisplay);
         }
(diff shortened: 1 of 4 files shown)


bjarnef:
It is now possible to request datatype in backoffice by `Guid`.
This is especially useful on Umbraco Cloud if you get errors regarding a datatype with a specific `Guid` or `Udi` and is a bit more consistent with other parts after backoffice UI.

image

Maybe we also want this in another parts of backoffice like on document type, media type, member type etc.?

emmaburstow:
Nice work @bjarnef! This is a fantastic improvement. It's exactly what that guid should be for, right?

Emma

src/Umbraco.Web/Editors/MacrosController.cs
```diff
@@ -97,39 +112,79 @@ public HttpResponseMessage GetById(int id)
...
+                                .OrderBy(x => x.SortOrder)
+                                .Select(x => new MacroParameterDisplay()
+                                {
+                                    Editor = x.EditorAlias,
+                                    Key = x.Alias,
+                                    Label = x.Name,
+                                    Id = x.Id
+                                });
+
+            macroDisplay.Parameters = parameters;
```
bjarnef: @Shazwazza I guess the `Parameters` could be set in the mapping (similar to how property editors are mapped on datatype), but it is nicer than the existing where it populated `MacroDisplay` properties in the controller.

bjarnef:
@kjac it can wrap this in a method like in ContentController.
https://github.com/umbraco/Umbraco-CMS/blob/c1bffefd236d69803e7937a02fd0462cac7fb1ba/src/Umbraco.Web/Editors/ContentController.cs#L2158-L2163

bjarnef:
@kjac I have updated the PR, where the parameters mapping is moved to a private method.

nul800sebastiaan:
Can we still use an int Id? 

What's your concern @kjac ? I think this might be perfectly fine, if the original way using the int Id still works!

kjac:
@nul800sebastiaan yes we can still use an int 😄 

I think it's perfectly fine too, and it's non-breaking so that's good. I just wanted to sanity check as it's principally a new feature.

I'll have a look at it again tonight and get it merged in.

bjarnef:
Yes, it works using id and `int`, `guid` and `udi`, not sure how the request looks with the `udi` but it cast the `Udi` id to `GuidUdi` and then access the `Guid` property, but it is something similar in ContentController.

kjac:
Merged! Thanks again, @bjarnef 😄

PR #6848, opened 2019-10-25
Removed the extra padding on the left on the overlay header

poornimanayar:
There are a number of overlays in the backoffice which has a en extra padding on the header by the left.

image

Another example is "You have unsaved changes" overlay
image

I have fixed the padding so the header and the confirmation text align up.

image

image


emmaburstow:
👍

nul800sebastiaan:
Lovely, thanks Poornima!

PR #6853, opened 2019-10-25
V8: Make the listview delete dialog warn about irreversible deletions

kjac:
### Prerequisites

### Description

_If it feels like deja-vu here, it's because #6600 is a lot along the same lines as this one..._

When you delete items from the recycle bin, the warning dialog is the same as is used when deleting items from any other listview. Among other things it tells you to unpublish languages (for culture variant content), which is meaningless in this context:

image

This PR adds a proper warning message to the delete dialog when it's used in context of the recycle bin:

image

...while retaining the warning when deleting (culture variant) content from other list views:

image


nul800sebastiaan:
Works great! Thanks! 👍

PR #6854, opened 2019-10-25
V8: Fix styling for publish/unpublish dialogs with mandatory language in listviews

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description

When you have a mandatory language, the publish/unpublish dialogs from listviews lack a bit in styling:

image

image

This PR makes them a bit prettier:

image

image

PR #6855, opened 2019-10-25
V8: Listview unpublish dialog should automatically select all languages

kjac:
### Prerequisites

### Description

When you have a mandatory language, unpublishing content in that language will unpublish the content in all languages. 

The default unpublish dialog handles this by forcefully selecting and disabling all other language checkboxes if the mandatory languages is selected for unpublishing. But the listview unpublish dialog doesn't handle it at all:

listview-unpublish-before

This PR adds the same handling to the listview unpublish dialog:

listview-unpublish-after

See also #6854 with regards to the mandatory language styling in the dialog 😄


nul800sebastiaan:
Wonderful! Thanks very much! 👍

PR #6858, opened 2019-10-25
Confirm uninstall package: Change old checkbox to use umb-checkbox

BatJan:
### Prerequisites

### Description
I have done 2 things in this PR

1: I extended the umb-checkbox directive with a parameter called "cssClass" making it possible to pass override classes as needed
2: I changed the old checkbox to using the umb-checkbox directive passing css modifiers for making the text smaller and bold like it was using the old pattern

The "no-dirty-check" directive has been removed since the checkbox is a boolean so it should not really be necessary unless I misunderstand something 😅 

**Before**
package-uninstall-before

**After**
package-uninstall-after


nul800sebastiaan:
Thanks very much @BatJan!

PR #6859, opened 2019-10-25
Grid layout config: Replace the old checkbox with the umb-checbox directive

BatJan:
### Prerequisites

### Description
I have replaced the old checkbox with the umb-checkbox directive - I have also temporarily added some inline styling to align the checkbox. But that's only until this PR is merged #6858

So before merging this PR please merge the other one and let me update this PR with a proper modifier CSS class 😉 

You might also wonder "Why not change the other checkboxes that are revealed when the new one is toggled?" - Well, that requires an extension to the umb-checkbox directive, which I'm having a think about these days to make it as flexible as we need it to be. Stay tuned 👀 

**Before**
grid-layout-before

**After**
grid-layout-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/layoutconfig.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/layoutconfig.html
@@ -78,17 +78,14 @@
 
 
           <umb-control-group hide-label="true">
-             <ul class="unstyled">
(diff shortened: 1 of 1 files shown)


bjarnef:
Maybe `umb-toggle` would be better in this case since it is more a setting.

BatJan:
@nul800sebastiaan Done 😉

nul800sebastiaan:
Lovely, thanks! 👍

PR #6860, opened 2019-10-25
Improved the editor service documentation a bit

abjerner:
The documentation for the `documentTypeEditor` function in the editor service was missing the `id` and `create` properties, so I've added them as well as improved the surrounding text a bit.

<hr />

I haven't been able to build the documentation locally. The guide on how to do so mentions Grunt, which is no longer in use. There used to be a `docs` task for Grunt, but this also seems to be gone now. I was curious to see how the back ticks would look, and if they would brake anything.

P.S.: Shouldn't the `documentTypeEditor` function be marked as deprecated, and then replaced by a `contentTypeEditor` function instead?

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/editor.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/editor.service.js
@@ -606,10 +606,12 @@ When building a custom infinite editor view you can use the same components as a
          * @methodOf umbraco.services.editorService
          *
          * @description
-         * Opens the document type editor in infinite editing, the submit callback returns the saved document type
+         * Opens the document type editor in infinite editing, the submit callback returns the alias of the saved document type.
          * @param {Object} editor rendering options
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
Yo @abjerner 

I found the build of the API docs bad to work with the last time I had a look as well, you can utilize DocFX's serve command to serve a locally built static site. I wrote how to do this under how to test in this PR: https://github.com/umbraco/Umbraco-CMS/pull/6532

Hope it helps 😁

nul800sebastiaan:
Different kind of docs @jmayntzhusen! 

The docs build has moved into a dedicated build, away from the rest of the "Belle" build since we updated the main build to versions that the old docs builder could not handle. 

The docs are now being built from `src\Umbraco.Web.UI.Docs` and there's a helpful README ;-)
https://github.com/umbraco/Umbraco-CMS/tree/v8/dev/src/Umbraco.Web.UI.Docs

This change looks good to me!

PR #6861, opened 2019-10-25
Log viewer: Localize texts and titles

BatJan:
### Prerequisites

### Description
Browsing through some files i the backoffice I noticed the log viewer lacked some translations so I added a new ara in the language files called "logViewer" and added the different keys in there and mapped the texts and titles in the view to using the language keys.

Keys are added for en, en_us and da.


nul800sebastiaan:
Awesome, thanks very much!

PR #6862, opened 2019-10-25
Replaced umb-overlay with editorService for listing fields of an Examine search result

abjerner:
Warren did a fantastic job of listing the fields of the search result, but I've always felt the overlay was a bit too small for content like this. And not to mention the `umb-overlay` directive is obsolete.

So with this PR, I'm turning this:

image

into this:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/dashboard/settings/examinemanagement.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/dashboard/settings/examinemanagement.controller.js
@@ -26,17 +26,15 @@ function ExamineManagementController($scope, $http, $q, $timeout, $location, umb
 
     function showSearchResultDialog(values) {
         if (vm.searchResults) {
-
             localizationService.localize("examineManagement_fieldValues").then(function (value) {
-
-                vm.searchResults.overlay = {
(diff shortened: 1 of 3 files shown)


emmaburstow:
👍

abjerner:
@nul800sebastiaan I only tested with a browser window taking up half the width of my monitor, so 1720 pixels. I don't mind the extra space here. But I can see the overlay looks ridiculous when the windows has full width (3440 pixels), so I've updated the overlay to `medium` size as suggested.

The issue with the scroll bar is because the browser has no natural way of wrapping things like URLs and JSON structures. I've added `word-break: break-word;` to the cells so the browser forces line breaks when needed. I couldn't find an obvious place in the LESS to add a rule for this, so I've gone with adding this as an inline style (💣 😱). If you can help with a better place to add the rule, I'd be happy to update my PR 😉

Pretty much [all of the "modern browsers" support `word-break`](https://caniuse.com/#feat=word-break), so it shouldn't be necessary to use CSS prefixes.

New screenshot with `size: "medium"` and `word-break: break-word;`:

image

kjac:
...and merged. Thanks, @abjerner ⭐️

PR #6863, opened 2019-10-25
Log viewer: Filter toggle

BatJan:
### Prerequisites

### Description
I have improved the filter dropdown, converting the `<a>` to `<button>` and adding appropiate aria attributes as well.

I have also added some empty `alt` attributes on the images rendering the favicons for different searches. This is done in order to let screen readers know they can safely ignore the image since the provided text is adequate in a non-visual context.

Furthermore I have modified some styling so the control aligns with the left edge and looks better in general.

There is more stuff to be done in this view but I'm going to split it all up in smaller PR's since doing one big file, would be confusing for everyone involved 😃 

**Before**
log-filter-before

**After**
log-filter-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/belle.less
+++ b/src/Umbraco.Web.UI.Client/src/less/belle.less
@@ -77,6 +77,7 @@
 @import "listview.less"; 
 @import "gridview.less";
 @import "footer.less";
+@import "filter-toggle.less";
 
 @import "forms/umb-validation-label.less";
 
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Thanks!

FYI This one got a lot trickier to merge @BatJan due to the search.html now having translations and this PR not having those yet, so I had to be careful. Think I managed though 🤞

BatJan:
@nul800sebastiaan Ah, sorry about that - Tried doing the thing where I did more minor PR's to make the review process easier... turns out it made the merging experience worse... 🤷‍♂ 😁 

I'm confident you managed to get it right though 😃

PR #6865, opened 2019-10-26
Languages: Add missing "none" translation

BatJan:
### Prerequisites

### Description
I noticed that the "(none)" text was not localized - Therefore I have added this translation key in the "languages" section for en, en_us and da files and mapped it in the view.


nul800sebastiaan:
Good stuff! 👍

Issue #6869, opened 2019-10-26
v8 - Installation of Machine Key Fails if web.config has multiple system.web sections


KevinJump:
If you run the Umbraco Installation on a site where the web.config has multiple system.web sections (if for example there is one in a `<location>` element, then the installation fails.

Reproduction
------------

on a new Umbraco project before installation add a system.web section to the config file inside a `<location>` element e.g

```
    <location path="some/path">
        <!-- Up the file upload limit for this path only -->
        <system.web>
            <httpRuntime maxRequestLength="512000" />
        </system.web>
    </location>

```
Run the installation and you will get a failure, 

### Bug summary

The installation process should look for the system.web under the root Configuration element and not any descendants as there may be many.

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/6870

PR #6870, opened 2019-10-26
look for the system.web section under the root configuration node

KevinJump:
use Element to get the root system.web. DescendantsAndSelf can return multiple

### Prerequisites

Fixes : #6869 Installation of Machine Key fails if config has multiple system.web sections.

### Description

When the machine key is added to the config, the code looks for DecendantsOrSelf("system.web").Single - if the file has multiple system.web elements this will throw an exception and installation will fail. 

This PR changes the lookup from DecendantsOrSelf to xml.Root.Element which will only find the `System.Web` element in the root `<configuration>` node. 

### To Test

-  NuGet Umbraco
- Add a extra system.web to the config 
e.g
```
<location path="some/path">
    <!-- Up the file upload limit for this path only -->
    <system.web>
        <httpRuntime maxRequestLength="512000" />
    </system.web>
</location>
```

Install Umbraco, ask for a machine key - key will be created in the main `system.web` and will not fail. 

- Equally normal installation will still work


nul800sebastiaan:
Wonderful @KevinJump, this works like a charm! 💪

PR #6886, opened 2019-10-26
Fix bug when saving a search in the logviewer

skttl:
### Prerequisites

### Description
When you try to save a search in the logviewer you get an error, because you are sending the complete json object with dates to the controller.

I fixed this by only sending the filterexpression, so it now works.

To test, open the logviewer, perform a search, try to save it.
Find the search in the dropdown of saved searches and verify that it works.


poornimanayar:
Hi @skttl ,

The fix works but will there be a scenario where the dates need to be served. I am not sure whether/how to search by dates but in such a case shouldn't the saving of dates be allowed?

Poornima

kjac:
> I am not sure whether/how to search by dates but in such a case shouldn't the saving of dates be allowed?

@poornimanayar the dates are selected on the logviewer frontpage and shouldn't be saved as part of the query 😄 

image

PR #6888, opened 2019-10-26
Content templates dashboard: Add translations and move styles external

BatJan:
### Prerequisites

### Description
I have added the missing translations for this dashboard and also moved all of the inline styles into an external stylesheet so we keep things clean.


BatJan:
Hey @poornimanayar - Hmm, I just checked on both en and en-us contain the new area for the translations at the bottom of the file so it should work I think?

If you inspect the text is it wrapped in `<localize>` tags? And is all the previous inline styling removed from `<p>` and `<ul>`? If so then it should be fine? 😃

nul800sebastiaan:
Note that most translations won't get picked up until the app pool recycles (if you touch the web.config and save it that will be enough).

poornimanayar:
@BatJan @nul800sebastiaan  I did a full rebuild. Testing it again :-)

poornimanayar:
Hi @BatJan,

So this is what prompted me to ask you to re-check. It might very well be my lack of knowledge, apologies if thats the case :-(  I did a front-end build and went into the Healthcheck dashboard and I see the below. I would expect all the text to show the localization keys but there is one line of text which shows up. And yes, nothing in code that triggers it! I do a full build and I can see all content. But apart from this all okay and happy to approve although there is a merge conflict to be resolved.

image

BatJan:
Hey @poornimanayar - Hmm, that's really weird. I have just double checked and it works as expected on my machine.

Whenever I need to check that translations are added and working I find that it's usually neccessary to first bump the version number in the ClientDependency.config and then recycle the app pool before the change takes effect.

Maybe you also need to do this in this particular case?

BatJan:
@poornimanayar @nul800sebastiaan Is this one still not acting right on your machines? 😃

nul800sebastiaan:
Sorry for the delay @BatJan! This is working great here! 👍  Thanks!

PR #6892, opened 2019-10-27
Changed the styling of Resend Invitation button and fixed the margin

poornimanayar:
The "Resend Invitation" button is a currently an `info` style button (grey). I think it should be an `action` style button as it is an action. It also brings it in line with the **Invite User** screen where the Send Invite button is `action` style. There was also a margin to the left of all the buttons in the sidebar on user details screen which were causing the buttons to shift a little. In addition, I have also introduced a bit of margin after the buttons and before the "Last Login" as it was appearing too close and squished.

Let me know if this is okay :-)

**Before**
image

**After**
image


emmaburstow:
Lovely work!!! Thank you :D

PR #6898, opened 2019-10-27
Log Viewer - Fix search toggle accessibility

BatJan:
### Prerequisites

### Description
When you expand a line in the log viewer you will see a "Search" button, which will list some options of searching the issue on different search engines, on Github etc. I've improved the accessibility and the UX doing the following

- Convert `<a>` to `<button>`
- Added `aria-haspopup="true"` and `aria-expanded="true/false"`
- Removed `data-toggle="dropdown`, which has no impact at all. I think it's an old Bootstrap pattern that has been abandoned at some point
- Added the `deep-blur` directive to the parent element so that the "dropdown" will disappear when it's tabbed out of

**Before**
logviewer-search-dropdown-before

**After**
logviewer-search-dropdown-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/logviewer/search.html
+++ b/src/Umbraco.Web.UI.Client/src/views/logviewer/search.html
@@ -151,11 +151,18 @@
                                                 </tbody>
                                             </table>
 
-                                            <div class="btn-group">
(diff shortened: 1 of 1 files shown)


emmaburstow:
Lovely work, thanks Jan

PR #6904, opened 2019-10-27
Listview: Various corrections

BatJan:
### Prerequisites

### Description
Ok, so quite a few small corrections in this PR - It started out with me wanting to remove the `data-toggle` attribute since it does not do anything anyway but then I discovered some other stuff that I could change too now that I was at it anyway 😃 

The following has happened
- Removal of `data-toggle="dropdown` - It's an old Bootstrap leftover that does not do anything
- addition of `aria-hidden="true"` on `<i>` elements where it was missing ensuring no screen reader will try to announce it
- Changing some `<a>` tags to `<button>`
- Adding a `<label>` for the listview search instead of just an icon with a click event. Using a label makes the click event unnecessary since the label references the input field
- Adding a sr-only text for the `<label>`

The icon/label change looks like this (Basically the same as before but with improved semantics and accessibility 😉 )

label-containing-icon


poornimanayar:
Hi @BatJan ,

I think the code changes look fine to me. I have given the listview a good test to make sure everything is in order. Is there anything else I need to test?

Poornima

BatJan:
@poornimanayar Sounds good - I don't think so, sounds like you covered it all 😃

PR #6906, opened 2019-10-27
V8: Correct path to datatype references that are not document types

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description

The "Open" links on the new datatype references tab all lead to `#/settings/documentTypes/edit/{{::reference.id}}` - also for media and member types. This obviously leads to a 404 error if you attempt to open a reference that is _not_ a document type:

datatype-references-path-before

This PR updates the paths according to reference type:

datatype-references-path-after

PR #6670, opened 2019-10-12
Added ContentApps for Members

robmaas:
This PR enables ContentApps for Members.

Both the Content and the Info ContentApps have been implemented as well.
Some additional work can be done on the content of those tabs, for example Infinite Editing is not yet working for the MemberType in the Info ContentApp.

It is possible to limit the visibility based on the memberType (both manifest and c# definitions), like so:

    "show": [
            "-member/testType",
            "+member/*"
        ]

**BREAKING**
Please note that this can be a breaking change for sites if they throw Exceptions like Core does, this is discussed in issue: #5843 
The way to fix these issues is to include a IMember case to the switch statements:

    switch (source)
    {
        case IMedia media:
            // dosomething(media);
            break;
        case IContent content:
            // dosomething(content);
            break;
        case IMember _:
            // don't show for member
            return null;

        default:
            throw new NotSupportedException($"Object type {source.GetType()} is not supported here.");
    }

Screenshots:
1: Default view (no custom ContentApps)
image

2: Added two custom ContentApps (1 from manifest, 1 from code)
image

3: Removed the (internal) Info ContentApp from code using:

    composition.ContentApps().Remove<ContentApps.ContentInfoContentAppFactory>();

image
Since there is only 1 ContentApp remaing (Content) it does not show.

---
_This item has been added to our backlog [AB#3417](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3417)_

PR #6907, opened 2019-10-27
V8: Add infinite editing to datatype references tab

kjac:
### Prerequisites

### Description

From the datatype references tab you can open the referencing types (although it's slightly broken, see #6906). But once you do so, you loose your context. And browser back doesn't get you back to the datatype you were looking at, there's some redirecting going on that prevents it.

So. This PR introduces infinite editing for all referencing types. This also includes adding member types to infinite editing (which might be a feature on its own, but hey... two for one). It works like this:

datatype-references-infinite-editing

Notice how the datatype references tab reloads if the infinite editor submits, thus reflecting any changes made to the referencing type.


kjac:
Updated with latest v8/dev

dawoe:
I saw infinite editing for member types in this PR as well : https://github.com/umbraco/Umbraco-CMS/pull/6670/

You both named in the same way so merging should be pretty easy

kjac:
Updated with latest v8/dev

poornimanayar:
@kjac, looks like this is a duplicate of #7211 . So I will close this PR down. Thanks for your stellar work as always!

kjac:
@poornimanayar I don't think this is a duplicate of #7211 ... this PR adds infinite editing to the datatype references (info) tab - see the GIF

poornimanayar:
Bad Poornima! I got confused by [6906](https://github.com/umbraco/Umbraco-CMS/pull/6906), I should have been more careful. Apologies @kjac! I ll review this one shortly :-)

poornimanayar:
And merged :-)

PR #6908, opened 2019-10-27
Listview: Add table class

BatJan:
### Prerequisites

### Description
This PR adds the already existing `.table` class to the column table making it look much better.

Before
column-table-before

After
column-table-after


poornimanayar:
Thanks a lot for the work @BatJan .. Merged in now 👍

PR #6911, opened 2019-10-27
Macro parameters: Improve accessibility

BatJan:
### Prerequisites

### Description
In this PR I have added `aria-hidden="true"` on an icon to ensure screen readers don't try to announce them.

I have also changed `<a>` to `<button>`, which makes the "edit" and "remove" options possible to navigate using keyboard.

`Before`
macro-parameters-before

`After`
macro-parameters-after


poornimanayar:
Thanks once again for the work @BatJan .. Merged in now 👍

PR #6913, opened 2019-10-27
Image cropper config: Improve semantics

BatJan:
### Prerequisites

### Description
I have changed some `<a>` elements that did not have a valid ng-href/href attribute to `<button>` elements instead to improve the semantics and make screen readers announce "button" instead of link.

Visually things look the same as before the PR 😃


BatJan:
Hi @poornimanayar 

It's been a while so can't remember what it looks like currently but I would say that's for another PR :-)

emmaburstow:
Lovely work, Jan. Thanks

PR #6914, opened 2019-10-27
V8: Set tab focus on newly created nested content items

kjac:
### Prerequisites

### Description

Nested Content is fairly good when it comes to tabbing. But it fails when creating new items: The input focus should be applied to the newly created item, but clearly it isn't:

nested-content-new-item-focus-before

As you can see, input focus ends up down in the breadcrumb, and you have to shift+tab your way back up into the editor.

This PR applies automatic input focus to newly created items:

nested-content-new-item-focus-after


kjac:
Updated with latest v8/dev

poornimanayar:
And merged :-)

PR #6915, opened 2019-10-27
V8: Add :root LESS partial for setting useful CSS variables

nathanwoulfe:
### Prerequisites

### Description

As a package developer, I'm often needing to use Umbraco brand colors (and others) in the backoffice - currently that means finding the hex code either in source or dev tools in the backoffice, and referencing same in my project. That's fine until the Umbraco color palette changes, and suddenly my package is using outdated/obsolete colors.

Enter CSS variables, :root pseudo-element and this PR.

I've added a new LESS partial for defining CSS variables on :root, so that developers need not look up colors but can reference these directly:

```CSS
.my-element { color: var(--brandPrimary) }
.my-other-element { padding: var(--paddingMini) }
```

This can be verified in the browser console, by inspecting the HTML element - will see a list of available :root variables in the console.

See also https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties

No functional changes in this PR, but the variables are exposed for anyone wanting to use them. Can readily add others to the root.less partial, and make life a little easier for package devs.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/belle.less
+++ b/src/Umbraco.Web.UI.Client/src/less/belle.less
@@ -16,7 +16,7 @@
 @import "../../lib/bootstrap/less/type.less";
 @import "../../lib/bootstrap/less/code.less";
 @import "tables.less";
-
+@import "root.less";
 
 // Components: common
 @import "../../lib/bootstrap/less/dropdowns.less";
--- a/src/Umbraco.Web.UI.Client/src/less/root.less
+++ b/src/Umbraco.Web.UI.Client/src/less/root.less
@@ -1,0 +1,32 @@
+/* contains variables to make available on the :root pseudo element - ie colors, base sizing/padding units 
+   makes package developers happy :)
+*/
+
+:root {
+    --blue: @blue;
+    --blueMid: @blueMid;
+    --blueDark: @blueDark;
+    --blueExtraDark: @blueExtraDark;
+    --blueLight: @blueLight;
+    --blueNight: @blueNight;
+    
+    --pink: @pink;
+    --pinkLight: @pinkLight;
+    --pinkRedLight: @pinkRedLight;
+    
+    --brown: @brown;
+    --brownLight: @brownLight;
+    --brownGrayLight: @brownGrayLight;
+    --brownGrayExtraLight: @brownGrayExtraLight;
+        
+    --error: @red;    
+    --warning: @orange;
+    --success: @green;
+    
+    --brandPrimary: @blueExtraDark;
+    --brandSecondary: @pinkLight;
+    
+    --paddingMini: @paddingMini;
+    --paddingSmall: @paddingSmall;
+    --paddingLarge: @paddingLarge;
+}


nathanwoulfe:
@nul800sebastiaan 👋 merging this one would take but a minute, and you'd forever be in the good graces of package developers and backoffice extenders...

ronaldbarendse:
It would be great to have all other variables available too, but that could be added in another PR (this will indeed be enough for most package developers).

This might also make changing the back-office theme quite easy, although the variables would need to be more specific to their usage to be able to override correctly. Bulma does this by having initial and derived [variables](https://bulma.io/documentation/customize/variables/), so the initial variable is e.g. `$red: #D42054;` and the derived `$error: $red;`. Changing the shade of red should then be done by changing the `$red` variable, but changing the error color by changing `$error`. The same applies to link, button, text, title, section, etc. colors/spacing/fonts...

The whole codebase would need to use these new derived variables though, so all (preprocessed) LESS variables should be updated to CSS variables. This would drop IE11 [support](https://caniuse.com/#feat=css-variables), so fallbacks should be added using e.g. [`postcss-css-variables`](https://www.npmjs.com/package/postcss-css-variables) with `preserve: true` to mitigate this.

One thing to note: when package developers use these CSS variables and the browser doesn't support that (IE11), there's no fallback! This shouldn't be an issue if we're targeting only the supported browsers from the [documentation](https://our.umbraco.com/documentation/getting-started/setup/requirements/#browsers):
* Firefox (Latest)
* Chrome (Latest)
* Safari (Latest)
* Edge

nielslyngsoe:
Hi @nathanwoulfe I think this would be really good. But I have a few thoughts that I would like to share and then let's talk about how we can make this happen, cause it would truly be a better world like that.
— First I think we need to be specific that we are dealing with colors in the naming, just cause CSS-variables can be more and some usage-named variables might not be that obvious, even collide with other interests.. like "--border" or "--shadow"
— One of the reasons to make this available is to make sure that packages don't break when colors are updated, aka. colors will change. So it might not be the smartest to have color-named variables, cause we can't change orange to blue if its named orange. So naming for this is super hard, I'm not super fond of the way it's done today, but that at least isn't exposed and therefore it can be changed. But when making these CSS-Variable we have to stick with them cause packages will be dependent on them. So I would really like us to put a few thoughts into what would be the most optimal naming convention, with the future in mind. I'm both concerned with sticking to color-named variables but I'm also concerned with usage-named variables Like let's say "outline-color". Cause the outline color might work well on a white and blue background today, but in the future when a new color is assigned it might not work in all scenarios. Where I'm getting with this is that I don't have the right solution at hand, but I would love us to put some thoughts in the name-system — cause too abstract is not good and too detailed isn't either.
Btw. I think its good to have the color-named variables at hand, but we need to notch devs to use usage-vars, so it might be an idea to not expose the color-named variables only usage-named variables, or name them in a way there its clear that the color named variable isn't the right choice. My concern is that devs use the blue color instead of --color-brand-primary.

And to make it even harder we don't have a design system, aka. we don't have a systematic way to use colors as it is today, but we hopefully get closer to that as we aim for getting a component library and getting things aligned. In other words, we should strike to find a naming system that can evolve. With that in mind, a very simple system might be the way to start, especially if we prefix things in a way that allows for future detail in naming?

The main problem, as I see it, is to make sure that the variables in the future don't change in a way where we end up with low or no contrast in some cases. So we might need to think about contrasts in our naming-system. It might be good enough with a somehow well defined usage-named variables that make it clear what goes against what, like "--color-button-primary-background" together with "--color-button-primary-text". But in some cases we are not ready to be this specific about the usage jet, so we might not introduce names like foreground/background to enable us to get in a good place with the names today.

Those where my thoughts, sorry for the length, I hope you are up for helping out making sure we get a good start on this :-D Thanks in advance and feel free to express your opinion freely, I'm open :-)

nathanwoulfe:
@nielslyngsoe I completely agree - whatever is done needs to be future-proof, otherwise it defeats the intention of making life easier for package devs (no point in making these changes if it means devs need to check every Umb release to ensure --green is still green). I don't think we could ever make it completely foolproof, and at the end of the day, there's nothing stopping a dev from using their own colors/values, and ignoring the available values.

I don't think we'd need to expose absolutely every variable in the existing codebase. Devs should be able to leverage the existing directives/components to build their UI, which removes the need for writing much (if any) CSS, but for styling custom elements, a basic set of safe variables would be ideal. 

Angular's Material theming manages color nicely, using weighting to manage color palettes. Could use a similar approach, and expose a series of colors for primary, secondary, accent etc:

primary-500 => the brand blue aka @blueExtraDark. 500 is the base color for the palette
primary-100 => the lightest blue
...
primary-800 => the darkest blue

That abstracts the actual color so that if Umbraco decided blue was too 2020 and wanted to use green instead, the primary palette would continue to work. Would be possible too to define palettes for text, which could be accompanied by usage rules (id text-800 should be used on colors < 400, text-100 on colors > 500).

It's that big old naming challenge. Implementing these changes isn't anything complex, but it needs to be named correctly else it's wasted effort 😄 For me personally, colors are a time consuming step in package work - I usually end up with a SCSS partial which duplicates the key Umbraco colors as variables  ($danger, $warning, $success, $umbPrimary, $umbHighlight and so on).

nathanwoulfe:
This PR was from a fork I've since destroyed, so I'll create a new one with updated variable names, and migrated this discussion into the comments.

PR #6917, opened 2019-10-28
Code cleanup for MB embed

Shazwazza:
* no more statics or service locator
* some new namespaces, no more Umbraco.ModelsBuilder.Umbraco
* removed obsolete classes
* imports some tests from MB original where it made sense
* interface for IModelsBuilderConfiguration
* Commented out `StaticMixinMethods` - since it's not used, i don't think MB 'lite' needs this?


bergmania:
Looks good.

PR #6919, opened 2019-10-28
Feature/media provide screen reader context remove files

poornimanayar:
In the media section when you edit a media item "Remove files" can be tabbed to, but does not give context to the screen reader users. I have added additional buttons and translation keys for screen reader users to help with this.


poornimanayar:
Hi @kjac ,

I am not sure whether that is needed in this instance. In the below screenshot I think it is clear to non-screen reader editors that the link should be used to remove files, but if you still think we need to improve on it, happy to add that on to the same PR.

image

Poornima

poornimanayar:
Yes, i think that was a merge gone bad. I had trouble with some of my merges yday. I have reverted the changes now

poornimanayar:
All done @kjac

kjac:
Thank you @poornimanayar - merged 😄

PR #6921, opened 2019-10-28
Removed Members Dashboard Intro

poornimanayar:
I might be wrong, but looks like the `membersdashboardintro` is not used. So removed it :-)


emmaburstow:
Thanks for the work, P

PR #6926, opened 2019-10-28
V8: Hide the "blob:" URL for images uploaded via the RTE

kjac:
### Prerequisites

### Description

If you drag an image from your computer to the RTE and subsequently go to edit it (before saving), you're presented with an ominous looking URL:

image

I'm not entirely sure what happens if you edit this URL, but you shouldn't be allowed to do so. Thus... this PR hides the "blob:" URL:

rte-image-upload-hide-url


kjac:
PR updated with latest v8/dev

poornimanayar:
And merged :-)

PR #6927, opened 2019-10-28
V8: Auto-resize the RTE after inserting images

kjac:
### Prerequisites

### Description

😱 Large GIFs ahead 😱 

When inserting images in the RTE it doesn't always resize itself to fit the new content; in fact I find it fails to do so more often than not:

rte-auto-resize-before

As soon as you click virtually anywhere in the UI, the editor auto-resizes to fit... but it kinda feels like it should do that on its own when the image is inserted. So! This PR forces the RTE to auto-resize after an image is inserted. It works like this:

rte-auto-resize-after

_In these GIFs I've used the new RTE image upload to demo the issue, but it also happens when inserting images from the media library._


poornimanayar:
@kjac , I assume that if a height is set on the RTE itself in the datatype it wont resize beyond the specified height?

kjac:
@poornimanayar you assume correct 😄 this PR simply triggers a resize of the RTE; it doesn't manipulate the inserted image in any way.

poornimanayar:
And merged..

PR #6928, opened 2019-10-28
V8: Remove "double tabbing" through the itempicker

kjac:
### Prerequisites

### Description

If you use tab to navigate the item picker overlay (e.g. for Nested Content item selection), each item in the picker has two tab stops:

nested-content-double-tab-before

Seems a bit excessive! This PR boils it down to just one tab stop per item:

nested-content-double-tab-after


poornimanayar:
@kjac , can you fix the conflicts on this?

kjac:
@poornimanayar looks like recent changes to the item picker has made this a non-issue, so I'm closing the PR 👍

PR #6929, opened 2019-10-28
User group tab: Accessibility improvements

BatJan:
### Prerequisites

### Description
Since the introduction of the `ngAria` module into core every non-interactive element that has a `ng-click` attribute will automagically be given the `role="button"` and `tabindex="0"` attributes. One downside to this is that in some scenarios keyboard navigation becomes a bit clunky/weird. But nonetheless it's still better than it was before! However we should aim to eliminate ng-click handlers on non-interactive elements in general. Therefore I added a `<button>` element, whenever a row should be selectable meaning only row's that are selectable will get focused instead of all row's.

In this PR the following has happened

- Added a `<button>` element with an ng-if deciding whether to show it or not if the row is selectable or not. The button contains a visually hidden text saying "Select row: " + the computed value of group.name
- Aria-hidden has been added to the `<i>` elements except for the one in the search, which is on purpose since I have another PR open where I need a certain class to exists in the core in order to properly fix that particular icon 😉 
- Added a missing translation key for "All sections"

This is what the before and after visually looks like. Notice how all the row's got highlighted before even when they were not selectable. The PR address and fixes this.

**Before**
user-group-before

**After**
user-group-after


BatJan:
@kjac Sorry about the very late response for this one... So much stuff has happened the past few months 😅 

I hear you... And maybe this particular case is where I need to bite the dust and use something like `role="button"` on the div's instead and some other trickery in order to ensure maximum accessibility.

Currently time is an issue so I'm not sure when I will have a free moment to look further into this. Please keep this open for now - I will find some time eventually 😃

BatJan:
@kjac @nul800sebastiaan Oops, got did something very very wrong - Can you help me reopen this PR please? 🙏

nul800sebastiaan:
@BatJan  Ehm yeah.. there's 0 commits in this one and you force pushed, so there's nothing I can actually do. Maybe you still have the commit with the changes you made so that you can push them? But seems unlikely.. 🙈

nul800sebastiaan:
Maybe @poornimanayar still has the PR branch on her machine? That could help then we at least have the changes.

nul800sebastiaan:
Ah, I still had a copy of that branch these are the changed file below @BatJan - can you make a new branch locally, copy these files in (make new changes?) and push for a new PR?

[6929.zip](https://github.com/umbraco/Umbraco-CMS/files/4456570/6929.zip)

poornimanayar:
Sorry @nul800sebastiaan just saw this email trail now! But glad you found the branch :-)

PR #6931, opened 2019-10-28
V8: Fix JS error when using the hide search function in treepicker

kjac:
### Prerequisites

### Description

The hide search function in the treepicker (the left arrow in the treepicker search field) fails miserably with a JS error in the following case:

1. You have some list view enabled content which has children of the same type (list view enabled being the culprit here).
2. You search for and pick one of these children in the tree picker.
3. You click the left arrow to go hide the search results.

It goes like this:

treepicker-hide-search-before

This PR fixes the problem by applying a bit of defensive coding:

treepicker-hide-search-after

**Note:** I'm not entirely sure that the faulty code is even supposed to be there. It seems as if it's trying to append the selected tree node children to the parent list view node, presumably in order to display the selected list view child in the tree.


nathanwoulfe:
Hey @kjac, hope you've been keeping well. Found this one in my digging through old V8 PRs, was still an issue despite the controller code being substantially updated. I've resolved the merge issues and updated to apply the same logical change.

nathanwoulfe (approved):
JS had changed substantially since this PR was created, but the logical issue remained - I've updated to apply the same guard to prevent attempting to iterate the children array when it may be null/undefined

PR #6933, opened 2019-10-28
V8: Display picked item state when navigating the treepicker

kjac:
### Prerequisites

### Description

The treepicker isn't super great at displaying the current picked item state when you're moving around in the tree. It's particularly bad when navigating listview nodes and when searching:

treepicker-browse-selected-state-before

This PR makes a brave attempt at fixing these issues, so the currently picked items are always marked with the checkbox on matter how you navigate the tree:

treepicker-browse-selected-state-after


poornimanayar:
And merged ... Thanks as always 👍

PR #6934, opened 2019-10-28
V8: Add autofocus to treepicker search box

kjac:
### Prerequisites

### Description

Media Picker applies autofocus to the search box when the picker is opened, which means the editors can instantly search for the media they want. 

The treepicker component (which powers MNTP and Content Picker) does not apply the same autofocus:

treepicker-search-autofocus-before

Being a huge fan of the search boxes, I think that's a darn shame. So this PR ensures that the treepicker search box automatically receives focus when the treepicker is opened:

treepicker-search-autofocus-after


poornimanayar:
And merged :-) Thanks @kjac

PR #6935, opened 2019-10-29
Allow legacy Models Builder to continue to work if detected

Shazwazza:
## IMPORTANT

This PR: https://github.com/umbraco/Umbraco-CMS/pull/6917 **must** be reviewed and merged first!

## Changes

* Ensures that if the legacy Umbraco.ModelsBuilder.dll exists in the /bin that the embedded MB is disabled
  * Disabling includes disabling the embedded dashboard/manifest/controller too
* Re-adds the `Enabled` MB flag that was previously removed
* Changes embedded assembly and namespace to Umbraco.ModelsBuilder.Embedded since we cannot have the same DLL/namespace as the legacy one
* Changes the to use the App_Plugins/UmbModelsBuilder folder since we cannot have the same folder as the legacy one
* Changes the controller name to ModelsBuilderDashboardController  since it cannot be the same as the legacy controller name
* Ensures that the MB assembly version is included in the models hash so that anytime the MB version changes, the models are re-generated, this also means that changing between legacy and embedded will seamlessly work
* Splits up a few files

## Testing

Currently the legacy MB nuget package is still installed, which means when you build the solution, you will actually be using the legacy MB, not the embedded one. 

* Make sure you have MB enabled with PureLive
* Start your site with the Starter Kit (which uses MB) and it should continue to work
* In the folder: App_Data/Models, there's a file: all.generated.cs which includes all generated models. If you open it you will see that there's an assembly attribute near the top: `[assembly: PureLiveAssembly]` ... this is a good indicator that the legacy MB is running. This attribute has been removed from the embedded MB since it was obsoleted.
* While the site is running, delete the file: /bin/Umbraco.ModelsBuilder.dll and refresh the home page - it should continue to work as expected. Go check the all.generated.cs file again and you will notice that the declaration `[assembly: PureLiveAssembly]` doesn't exist, which is a good indicator that the embedded MB is running
* recompile the solution - this will re-deploy the Umbraco.ModelsBuilder.dll to the bin, refresh the home page and it should all still work, you can double check the all.generated.cs file to see that you are again running with the legacy MB version

I'll create a separate task to remove the legacy MB nuget package from being part of the solution and that will be a separate PR.


bergmania:
Now that MB is embedded, I think we should move the stuff from `App_Plugins` and into `Umbraco/`

Do you see any reasons not to do this?

PR #6936, opened 2019-10-29
 media tracking references

bergmania:
# NOTE: This PR need to be merged first: https://github.com/umbraco/Umbraco-CMS/pull/6843

Track references of documents and media from

- Media Picker
- Content Picker
- Multi Node Tree Picker
- Multi Url Picker
- Grid
- Nested Content

Fixes AB#3325
Fixes AB#3326
Fixes AB#3327
Fixes AB#3328
Fixes AB#3329
Fixes AB#3331

## Tests

- Added media and/or content to all of the listed property types.. 
- Also try double nested content.


src/Umbraco.Web/PropertyEditors/ContentPickerPropertyEditor.cs
```diff
@@ -25,5 +26,23 @@ protected override IConfigurationEditor CreateConfigurationEditor()
...
+            {
+            }
+
+            public IEnumerable<UmbracoEntityReference> GetReferences(object value)
+            {
+                var asString = value is string str ? str : value?.ToString();
+
+                if (string.IsNullOrEmpty(asString)) yield break;
+
+                yield return new UmbracoEntityReference(Udi.Parse(asString));
```
Shazwazza: Maybe we should do a TryParse just in case?

src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
```diff
@@ -131,16 +145,41 @@ public override object ToEditor(Property property, IDataTypeService dataTypeServ
...
             }
 
-            private GridValue DeserializeGridValue(string rawJson, out IEnumerable<GridValue.GridControl> richTextValues)
+            private GridValue DeserializeGridValue(string rawJson, out IEnumerable<GridValue.GridControl> richTextValues, out IEnumerable<GridValue.GridControl> mediaValues)
             {
                 var grid = JsonConvert.DeserializeObject<GridValue>(rawJson);
 
                 // Find all controls that use the RTE editor
-                var controls = grid.Sections.SelectMany(x => x.Rows.SelectMany(r => r.Areas).SelectMany(a => a.Controls));
+                var controls = grid.Sections.SelectMany(x => x.Rows.SelectMany(r => r.Areas).SelectMany(a => a.Controls)).ToArray();
```
Shazwazza: Only a nitpicky thing, i used to do ToArray as opposed to ToList by default, but it turns out that ToArray is slower and allocates more mem than ToList

src/Umbraco.Web/PropertyEditors/MediaPickerPropertyEditor.cs
```diff
@@ -17,14 +19,36 @@ namespace Umbraco.Web.PropertyEditors
...
+            {
+            }
+
+            public IEnumerable<UmbracoEntityReference> GetReferences(object value)
+            {
+                var asString = value is string str ? str : value?.ToString();
+
+                if (string.IsNullOrEmpty(asString)) yield break;
+
+                yield return new UmbracoEntityReference(Udi.Parse(asString));
```
Shazwazza: also should do TryParse just to be safe that things don't explode

src/Umbraco.Web/PropertyEditors/MultiNodeTreePickerPropertyEditor.cs
```diff
@@ -18,5 +20,29 @@ public MultiNodeTreePickerPropertyEditor(ILogger logger)
...
+            }
+
+            public IEnumerable<UmbracoEntityReference> GetReferences(object value)
+            {
+                var asString = value == null ? string.Empty : value is string str ? str : value.ToString();
+
+                var udiPaths = asString.Split(',');
+                foreach (var udiPath in udiPaths)
+                {
+                    yield return new UmbracoEntityReference(Udi.Parse(udiPath));
```
Shazwazza: TryParse as above

src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
```diff
@@ -266,6 +266,45 @@ public override object FromEditor(ContentPropertyData editorValue, object curren
                 return JsonConvert.SerializeObject(value);
             }
 
+            public IEnumerable<UmbracoEntityReference> GetReferences(object value)
+            {
+                var rawJson = value == null ? string.Empty : value is string str ? str : value.ToString();
+
+                var result = new List<UmbracoEntityReference>();
+
+                var list = JsonConvert.DeserializeObject<List<object>>(rawJson);
```
Shazwazza: Would it be better to just Deserialize as `List<JObject>` then you don't have to cast below

src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
```diff
@@ -266,6 +266,45 @@ public override object FromEditor(ContentPropertyData editorValue, object curren
...
+                if (list == null)
+                    return result;
+
+                foreach (var o in list)
+                {
+                    var propValues = (JObject) o;
+
+                    var contentType = GetElementType(propValues);
```
Shazwazza: This is an N+1 call, but I realize that this is being done in the `ConvertDbToString` method too. I also realize that all content types are cached, so it's not hitting the DB each time but still something to ponder ;)

Shazwazza: On that note, this code is basically all duplicated from the ConvertDbToString method. We should prob make a re-usable method to do this iteration

bergmania: Actually this was also what I did first and I ended up in something similar to what you did, but I was not proud of the readability of the delegate similar to this: `Action<string, PropertyType, JObject, int> onIteration`. 😢 

But the code duplication was not good either.

Shazwazza: Yeah, it's not the prettiest but I'll take it over the duplication, at least its isolated to this editor.

PR #6938, opened 2019-10-29
Remove inline styling from the tour intro

poornimanayar:
There was some inline styling in the tour intro overlay around the "Dont show this tour again" button which I have removed

image


poornimanayar:
@kjac, there was a reason but I cant remember what now. But i have tweaked it to remove the id now.

kjac:
Aaaand merged 🎉

PR #6942, opened 2019-10-29
V8: Use "not allowed" cursor for locked type aliases

kjac:
### Prerequisites

### Description

OK so this is a tiny thing, but...

Some type aliases are uneditable because they're reserved by the system (e.g. the built-in user group aliases). They appear in the UI without the little padlock that usually lets you "unlock" the alias to edit it. 

image

For some reason these uneditable type aliases do not have a "not allowed" cursor when you hover the mouse above them - they use the default "text highlight" cursor. This PR changes the cursor to "not allowed".


poornimanayar:
And merged

Issue #6632, opened 2019-10-09
Can't configure alias for Media Type "Folder", dialogs not working.


enkelmedia:
We wanted to prefix all our models generated by Models Builder with "Cms" and currently the only way to do this was to change the alias of the doc-types, media-types etc.

It looks like it all worked but when creating a new folder in the "Selected Media"-dialog. We're getting an error that there is no media type with the alias "Folder". 

![image](https://user-images.githubusercontent.com/1782524/66497822-5c505d80-eabd-11e9-8ec0-2f25b41e73b8.png)

Looking at the code in the "PostAddFolder"-method we can see that this is because the alias "Folder" is hardcoded as a constant.

Controller:
https://github.com/umbraco/Umbraco-CMS/blob/56545ad7f1fdee3c2b27e725c89378ccb947f65a/src/Umbraco.Web/Editors/MediaController.cs#L602

Constant:
https://github.com/umbraco/Umbraco-CMS/blob/6f5665f6a547881f85b393b7865723152c51a21e/src/Umbraco.Core/Constants-Conventions.cs#L114

Other types of "conventions" like the default alias of a member is something that we can configured but as far as I can understand it the default alias for a Folder (file,image) can't be configured?

nul800sebastiaan:
Hey @enkelmedia - I am having a really hard time understand what you're trying to accomplish and why? Do you have some background info for me please?

enkelmedia:
Well, the stuff about Models Builder was maybe just making it more confusing =D

Basiclly, if I change the alias of the media type "Folder" to something else, the built in views will blow up since the alias "Folder" is hard coded as a constant. What I'm saying is that there should be a way to configure what the defualt alias for a Folder, File or Image should be.

stevemegson:
Rather than add settings to change the media types that we try to create, could we come up with sensible rules to pick a media type from those allowed under the parent? Perhaps File would use the first type with an upload property, Image would use the first type with an image cropper property, and Folder would use the first type with neither.

enkelmedia:
That would also be a good option, I ended up reverting the changes but I still think that it would be good to be able to have a custom alias for a media type.

ronaldbarendse:
As https://github.com/umbraco/Umbraco-CMS/pull/6943 protects renaming these media types from the back-office (it's still possible using the services), this probably won't be an issue anymore.

[...]

PR #6943, opened 2019-10-29
V8: Protect "system media types" from alias changes and deletion

kjac:
### Prerequisites

### Description

The default media types "Image", "File" and "Folder" are used in a number of places and expected to be part of any Umbraco installation - things go horribly wrong if they are not present (see for example #6632).

While this dependency on "known media types" is not ideal, we should at least protect these media types from deletion and/or alias changes until a less hardcoded solution can be implemented in the core, so newcomers do not accidentally destroy their Umbraco installation simply by using the backoffice. 

This PR ensures that:

1. "Image", "File" and "Folder" can't be deleted.
2. "Image", "File" and "Folder" can't have their aliases changed (see also #6942).

image

image

### DIFF
--- a/src/Umbraco.Core/Models/MediaTypeExtensions.cs
+++ b/src/Umbraco.Core/Models/MediaTypeExtensions.cs
@@ -1,0 +1,10 @@
+﻿namespace Umbraco.Core.Models
+{
+    internal static class MediaTypeExtensions
+    {
+        internal static bool IsSystemMediaType(this IMediaType mediaType) =>
+            mediaType.Alias == Constants.Conventions.MediaTypes.File
+            || mediaType.Alias == Constants.Conventions.MediaTypes.Folder
+            || mediaType.Alias == Constants.Conventions.MediaTypes.Image;
+    }
+}
--- a/src/Umbraco.Core/Umbraco.Core.csproj
+++ b/src/Umbraco.Core/Umbraco.Core.csproj
@@ -268,6 +268,7 @@
     <Compile Include="Models\Entities\IMemberEntitySlim.cs" />
     <Compile Include="Models\Entities\MediaEntitySlim.cs" />
     <Compile Include="Models\Entities\MemberEntitySlim.cs" />
+    <Compile Include="Models\MediaTypeExtensions.cs" />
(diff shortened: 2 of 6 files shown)


nul800sebastiaan:
I always have a last look, so if it looks good them merge away @poornimanayar

ronaldbarendse:
Having files and folders in the media library is indeed a general concept, so protecting them seems logical (and great work, again, @kjac 👍).

But in some cases you might want to remove the `Image` media type and just use the `File` type instead (e.g. if you don't need image cropping or the additional metadata - and therefore the additional media type). Might be a nice feature to allow this kind of fallback, any thoughts?

kjac:
@ronaldbarendse this PR is all about safeguarding the current implementation where certain media type aliases are hardcoded in the codebase. It would certainly be beneficial to get rid of these hardcoded aliases, but that's another scope.

I think for the time being your best bet would be to replace the image cropper on `Image` with a file upload property.

ronaldbarendse:
As the media types are conventions (the aliases are within `Constants.Conventions.MediaTypes`), I would argue requiring them is contradictory to being a convention 🤷‍♂ On the other hand, not having a basic folder/file structure in the media section or doing a lot of lookups/checks whether they exist also isn't very useful 🙄

So my main point is that images can also be stored as a `File` media type and having an `Image` media type should therefore not be required. Making all properties the same between `File` and `Image` is certainly an option, but that only makes it more confusing to editors (and developers).

I've searched for references to `Constants.Conventions.MediaTypes.Image` and besides the extension method from this PR, the `DatabaseDataCreator` and tests, it's used in the following places:
https://github.com/umbraco/Umbraco-CMS/blob/b68f1d44f1c5c640c0c49264b34ecd6a9caa6a44/src/Umbraco.Web/Editors/MediaController.cs#L704-L714
https://github.com/umbraco/Umbraco-CMS/blob/f2d8d3f44f70c05ee283ac26747d767b50c3f10f/src/Umbraco.Web/Templates/TemplateUtilities.cs#L228-L231

There might be other references to the `Image` media type (because they don't use this constant, e.g. within AngularJS), but adding a check in these places would allow lifting the protection of the `Image` media type... I'll put something together 👍

PR #6903, opened 2019-10-27
V8: UX updates to the listview copy dialog + add option to include descendants

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description

The copy dialog used by listviews needs a little TLC:

1. It does not have any contextual help text. The copy dialog used by trees has a text like "Choose where to copy [node name] to in the tree structure below". 
2. The "Submit" button is always enabled. If you click it without choosing a target for the copy operation, the dialog simply closes without any user feedback.
3. There is no option to include or leave out descendants. The copy dialog used by trees has this option.

image

This PR addresses all of the issues mentioned above:

listview-copy-update

PR #6944, opened 2019-10-29
Replaced umb-checkbox with umb-toggle in listview copy overlay

poornimanayar:
I noticed that when you try to copy items in a list view the "Relate to Original" option uses `umb-checkbox`. I have replaced it with `umb-toggle` to make it consistent.

**Before**
listview-copy-before

**After**
listview-copy-after


bjarnef:
I wonder if the "Relate to original" and "Include descendants" should be `umb-checkbox` instead of `umb-toggle`.

E.g. in link picker we have "open in new window" as checkbox.

Not sure if these are classified as settings.

Here are some great articles about toggle vs checkbox.
https://www.nngroup.com/articles/toggle-switch-guidelines/
https://uxmovement.com/buttons/when-to-use-a-switch-or-checkbox/
https://uxplanet.org/checkbox-vs-toggle-switch-7fc6e83f10b8

poornimanayar:
@kjac ,

Thanks for the review. I have made the amends. Regd #6903 I had the exact same thoughts as yours while I was reviewing it :-) 

Poornima

kjac:
Merged! Thanks again ⭐️

PR #6946, opened 2019-10-29
Move the embedded Modelsbuilder files away from App_Plugins

bergmania:
## Changes 
- Removed the package.manifest - Dashboard implemented in code
- Moved the js and html files into `Umbraco.Web.UI.Client`
  - Renamed the stuff to `ModelsbuilderManagement` to avoid collisions and align with examine.
- Removed the `DisableModelsBuilderManifestFilter` because the package.manifest is removed.

#### Pros
- Even more embedded into the core
- We should not handle collisions with packages

#### Cons
- The embedded js for modelsbuilder dashboard cannot be removed
  - Is minified and bundled with the rest of our code

## Testing
Same testing as here: https://github.com/umbraco/Umbraco-CMS/pull/6935
+ Test that the dashboard works for both modes.

### OBS
I had some problems with nuget not copying the js files into `App_Plugins` when doing it directly on the branch and project `Umbraco.Web.UI` (Probably because we use PackageReferences).
I can be easier to test the builded project that uses legacy nuget `packages.config`


Shazwazza:
@bergmania This is awesome, i totally didn't think of this :P

PR #6949, opened 2019-10-29
Improve accessibility of grid prevalues screen

poornimanayar:
This is a WIP. Will add more changes later on. Some of the changes I am trying to address

- All icons to have `aria-hidden=true`
- Change the Delete, Edit links to be `link` style buttons


src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.prevalues.html
```diff
@@ -108,18 +107,18 @@ <h4><localize key="grid_rowConfigurations" /></h4>
...
                     </a>
                 </li>
             </ul>
 
             <ul class="unstyled list-icons">
                 <li>
-                    <i class="icon icon-settings-alt-2 turquoise"></i>
+                    <i class="icon icon-settings-alt-2 turquoise" aria-hidden="true"></i>
 
                     <a href="#" ng-click="editConfig()" prevent-default>
```
BatJan: This needs to be `<button>` too 😃

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.prevalues.html
```diff
@@ -137,10 +136,10 @@ <h4><localize key="grid_rowConfigurations" /></h4>
                 ng-model="model.value.styles">
 
                 <li ng-repeat="style in model.value.styles">
-                    <i class="icon icon-navigation handle"></i>
+                    <i class="icon icon-navigation handle" aria-hidden="true"></i>
 
                     <a href="#" prevent-default ng-click="removeConfigValue(model.value.styles, $index)">
```
BatJan: This should be a `<button>` as well

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.prevalues.html
```diff
@@ -149,7 +148,7 @@ <h4><localize key="grid_rowConfigurations" /></h4>
 
             <ul class="unstyled list-icons">
                 <li>
-                    <i class="icon icon-settings-alt-2 turquoise"></i>
+                    <i class="icon icon-settings-alt-2 turquoise" aria-hidden="true"></i>
                     <a href="#" ng-click="editStyles()" prevent-default>
```
BatJan: And finally this should be a `<button>` too :-) - Also it's a good idea to wrap a fallback text in the `<localize>` element since it will currently not hold a fallback value if  general_edit is missing in a language file. So it's better to have `<localize key="general_edit">Edit</localize>` for instance ensuring that if the key is missing in Danish then at least "Edit" will be rendered instead of nothing

poornimanayar:
@BatJan Thanks for your input! I shall work through them :-)

bjarnef:
@poornimanayar ensure also to add `type` attribute since browsers may have different default value for type attribute on button element.
https://stackoverflow.com/a/31644856

abjerner:
@bjarnef great point 👍 

I'm not sure why I missed this when testing, but clicking any the new delete buttons will save the data type as default type for a button is `submit`, but should be `button`.

Unfortunately I was a bit too quick to approve and merge this PR, so we can't really modify it now. @poornimanayar would it be possible for you to create a new PR making the buttons have `type="button"`?

poornimanayar:
Of course I will be doing them @bjarnef and @abjerner

PR #6951, opened 2019-10-29
V8: Re-apply input focus to the last focused element when closing dialogs

kjac:
### Prerequisites

### Description

Tabbing through the content editor is made really hard by pickers and other dialogs spawned by `editorService`. When they're closed, input focus is set to the first applicable element on the page, which is in the top menu:

reapply-focus-before

This PR makes `editorService` attempt to re-apply input focus to the element that had input focus before the dialog was opened:

reapply-focus-after

See also #6934.


poornimanayar:
And merged :-) Thank you for the work as always :-)

PR #6954, opened 2019-10-29
V8: Fix broken grid when reordering grid rows that contain RTE's

### LINKED ISSUE #6947: Reorder grid row containing RTE editor
Reordering grid rows containing the RTE editor error and can't be reordered.

Reproduction
------------
- Open a document containing the Grid editor
- Select the 'Reorder option
- Attempt to drag a row containing the RTE editor. It throws multiple ' Uncaught TypeError: Cannot read property 'settings' of undefined' script errors and freezes the reordering before complete. The page has to be refreshed to restore it to a workable state and the reordering is incomplete.

umbraco-grid-editor-reorder

The full script errors are as follows: 
umbraco.controllers.js?cdv=211:16506 Uncaught TypeError: Cannot read property 'settings' of undefined
    at HTMLDivElement.<anonymous> (umbraco.controllers.js?cdv=211:16506)
    at Function.each (jquery.min.js?cdv=211:2)
    at k.fn.init.each (jquery.min.js?cdv=211:2)
    at HTMLDivElement.stop (umbraco.controllers.js?cdv=211:16504)
    at HTMLDivElement.<anonymous> (sortable.js?cdv=211:58)
    at HTMLDivElement.<anonymous> (sortable.js?cdv=211:58)
    at HTMLDivElement.<anonymous> (sortable.js?cdv=211:59)
    at t.<computed>.<computed>._trigger (jquery-ui.min.js?cdv=211:6)
    at t.<computed>.<computed>._trigger (jquery-ui.min.js?cdv=211:13)
    at t.<computed>.<computed>._trigger (jquery-ui.min.js?cdv=211:6)

---
_This item has been added to our backlog [AB#3436](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3436)_

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #6947

### Description

While I don't exactly pride myself of being any kind of a Grid expert, I do believe this PR fixes the issues concerning reordering grid rows containing RTE's (described in #6947).

With this PR applied at least I seem able to drag RTE's around without breaking things:

grid-rte-reordering

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js
@@ -73,7 +73,10 @@ angular.module("umbraco")
                     ui.item.find(".umb-rte").each(function (key, value) {
                         // remove all RTEs in the dragged row and save their settings
                         var rteId = value.id;
-                        draggedRteSettings[rteId] = _.findWhere(tinyMCE.editors, { id: rteId }).settings;
+                        var editor = _.findWhere(tinyMCE.editors, { id: rteId });
+                        if (editor) {
+                            draggedRteSettings[rteId] = editor.settings;
+                        }
                     });
                 },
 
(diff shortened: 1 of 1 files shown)


stevemegson:
It may be worth also renaming the `umb-rte` class. Since 8.2.0 it's being used in two places - to style the whole RTE via [this div](https://github.com/umbraco/Umbraco-CMS/blob/e90670425e6b6cc712172d3a48ccb04bd811bf0c/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/rte.html#L1) and also to identify [this div](https://github.com/umbraco/Umbraco-CMS/blob/66e3869b06d334b0b698b8d5430aab2bf7f1a345/src/Umbraco.Web.UI.Client/src/views/components/grid/grid-rte.html#L3) in the code that's now failing. The code is now finding the first div as well as the one it expects. Although this change allows the code to skip over those divs without error, the clashing class name may cause other confusion and bugs in future.

clausjensen:
@stevemegson Thanks for mentioning it. Could you create a separate issue for this? It is not as critical as this one and this one will be going to a patch, where we only want to fix specific critical things without too much else 😃 
The other one could be targetted for 8.3 or a later minor then.

PR #6956, opened 2019-10-29
Replace compositions checkboxes

bjarnef:
### Prerequisites

### Description
This PR replaces the traditional checkboxes in compositions overlay on a document type with `umb-checkbox` component.

The function `selectCompositeContentType` update the `allowed` property, so we need to set `selected="false"` if `allowed="false"`.
https://github.com/umbraco/Umbraco-CMS/blob/d3344e2cd400e7aee3678cf4d4755bb80fb29b04/src/Umbraco.Web.UI.Client/src/common/directives/components/umbgroupsbuilder.directive.js#L303

With these changes (and in the current code with traditional checkboxes and if you are fast, you can select the checkboxes, which are not allowed, but the selecting is removed shortly after).

E.g. in default starterkit when selecting document type "Home", this should disallow "Product" and "Content Base" to be selected.


poornimanayar:
Haaa, great one @bjarnef, It was bugging me and I was sure you would pick this up. Well done you 👍

poornimanayar:
Hi @bjarnef ,

Good to see you picking this up. So here is what I have tested. Please let me know if I have missed out on anything

1) I can add an element type as a composition
2) I cannot add doc types or element types more than once if they have a property with same alias
3) I can add a composition
4) Once I select "Home" I cannot choose "Product" doc type as a composition.

I have tried to test and make sure that the existing functionality is as close to the new one as possible. All good I think :-)

Poornima

bjarnef:
@kjac yes it was a bit hacky because it call this function, which need to finish to disable checkboxes before selecting other compositions, where this function have a few other requests.

https://github.com/umbraco/Umbraco-CMS/blob/d3344e2cd400e7aee3678cf4d4755bb80fb29b04/src/Umbraco.Web.UI.Client/src/common/directives/components/umbgroupsbuilder.directive.js#L282-L325

kjac:
@bjarnef yep. I think that function could return a promise and resolve it where the `// TODO:` comments  are:

https://github.com/umbraco/Umbraco-CMS/blob/d3344e2cd400e7aee3678cf4d4755bb80fb29b04/src/Umbraco.Web.UI.Client/src/common/directives/components/umbgroupsbuilder.directive.js#L311 

...and: 

https://github.com/umbraco/Umbraco-CMS/blob/d3344e2cd400e7aee3678cf4d4755bb80fb29b04/src/Umbraco.Web.UI.Client/src/common/directives/components/umbgroupsbuilder.directive.js#L321 

But if that isn't feasible I'm still happy with this PR, it's a definitive step up from what we have. 

Simplifying the `$timeout` handler would also be good if possible.

bjarnef:
Yes, I guess it could have something like this with a `resolve` and `reject`.

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web.UI.Client/src/common/mocks/services/assetsService.mocks.js#L6-L8

kjac:
@bjarnef definitively 👍 are you OK with making that change?

bjarnef:
@kjac I have updated the PR where `selectCompositeContentType` now returns a promise, where `selectedContentType` and `newSelection` is available in response.

I have also added a loading indicator.

I wonder if we should have an overlay over the compostions document types while loading since some of the options might be disabled after the response is returned.

In starterkit if you select Home and then Product fast, it is diabled afterwards.

I was able to select both, but only one of them added to the document type after submit.

image

and after re-open compositions only Home was selected.

image

kjac:
@bjarnef the promise solution works really well 👍 

The load indicator... well, not so much if you have enough doctypes that you get a scrollbar in the compositions dialog. But I really like the idea of the load indicator - I too managed to select multiple doctypes that were invalid in combination (didn't attempt saving though!).

One possible approach could be something along these lines:

compositions-loading

Also: You left a `console.log("selectCompositeContentType", response);` in the PR 😄

bjarnef:
Yes, the `console.log` should indeed be removed 😁 

Not sure we want the hide the list, maybe just a "muted/dimmed" layer on top and show the loader in the top of this?

kjac:
@bjarnef I don't know if we have that ability on hand in the dialogs as-is? We certainly shouldn't introduce it "just" for this PR, but if it's already there then by all means.

nul800sebastiaan:
Whoa, blast from the past! Alright, I've updated this with the new `Utilities` calls, removed the `console.log` and tested it out, everything looks great, happy to merge this one! Thanks @bjarnef and @kjac! 👍

PR #6957, opened 2019-10-29
Change the style of delete confirm buttons at various places across back office

poornimanayar:
Similar to what I attempted in #6726 I have changed the delete confirm buttons to be`danger` style (red background, white text) to make it consistent. The confirm button on the delete overlay for the following features must the `danger` style

- Member group
- Document Type
- Document Type Folder
- Media Type
- Member Type
- Macro
- Relation Types
- Content Templates
- Templates
- Partial View
- Partial View Macro Files
- Stylesheets
- Scripts 
image

In addition to the above the overlay for deleting saved search in the log viewer must also be `danger` style
logviewer-overlay


kjac:
Merged. Thank you ⭐️

PR #6962, opened 2019-10-30
V8: Styling updates to image cropper and upload property editors

kjac:
### Prerequisites

### Description

The upload areas of the image cropper and upload property editors are a bit excessive in size, compared to most other editors:

image

Also when an image is uploaded to the upload property editor, the applied image border is a tad out of place:

image

With this PR applied, the upload areas have the same width as other property editors:

image

...and when images are uploaded, the upload property border doesn't go full width:

image

Notice how the image cropper is still allowed to span the full browser width when an image is uploaded. I chose this approach in order to leave room for those image cropper configurations with a lot of defined crops.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-property-file-upload.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-property-file-upload.less
@@ -1,6 +1,7 @@
 ﻿.umb-property-file-upload {
 
     .umb-upload-button-big {
+        max-width: (@propertyEditorLimitedWidth - 40);
         display: block;
         padding: 20px;
         opacity: 1;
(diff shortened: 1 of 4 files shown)


poornimanayar:
Thanks @kjac , Thats bringing more consistency in. I have tested it and can see it works without any issues. Works well on smaller screens as well!

poornimanayar:
@kjac , Merged now and thanks once again for your work!

bjarnef:
@kjac couldn't this use the existing `umb-property-editor--limit-width` which is used in other property editors?

https://github.com/umbraco/Umbraco-CMS/search?utf8=%E2%9C%93&q=umb-property-editor--limit-width&type=

PR #6527, opened 2019-10-01
V8: Make upload and image cropper mandatory validation work clientside

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description

Mandatory properties of types Upload and Image Cropper currently aren't validated clientside - only serverside:

upload-imagecropper-mandatory-before

This PR adds mandatory validation clientside as well:

upload-imagecropper-mandatory-after

PR #6965, opened 2019-10-30
V8: Allow file upload and image cropper properties to be optional

kjac:
### Prerequisites

### Description

Seems _someone_ (ahem... me) made a tiny mistake when making upload and image cropper mandatory validation work clientside (#6527)... they can no longer be optional:

image

This PR fixes it right up.

#### Testing this PR

1. Ensure that clientside mandatory validation works for image cropper and upload properties.
2. Ensure that non-mandatory properties of type image cropper and upload do not require uploaded files in order to save the content.


nul800sebastiaan:
D'oh! Of course I only tested mandatory ones, oops! 

Thanks for the correction! 👍

Issue #6703, opened 2019-10-15
V8, After you invite a user, a checkbox is shown along with a success message, and (wrongly?) given a 'cursor: pointer' style


kieronbos:
I realise this is incredibly minor, but after you invite a user, a checkbox is shown along with a success message, and (wrongly?) given a cursor: pointer style, by the class `.umb-checkmark`

I could be wrong, in which case what should the checkmarks action be? 

Otherwise, its incorrectly implying the checkbox is actionable when it isn't. 

I'm happy to try and do this but it'd be my first issue and Id need time to figure that out.

Maybe this comes under the accessibility umbrella.

Seen here: https://gyazo.com/783d08c7ec9a38573507e907579931f2

Reproduction
------------

Invite a user
Then checkbox appears on URL: https://x.x/umbraco#/users/users/users?invite=true

The html: 
`<i class="icon-check umb-checkmark umb-checkmark--m umb-checkmark--checked" ng-class="{'umb-checkmark--checked': checked}" checked="checked" size="m"></i>`

with style:

`.umb-checkmark {
    width: 25px;
    height: 25px;
    border: 1px solid #bbbabf;
    border-radius: 3px;
    box-sizing: border-box;
    display: flex;
    justify-content: center;
    align-items: center;
    color: #bbbabf;
    **cursor: pointer;**
    font-size: 15px;
}`

nul800sebastiaan:
Ah sure, should not have a pointer. Maybe you can help fix it?

umbrabot:
Hi @phunkb,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

phunkb:
Yes, I would love to fix it!

I will start investigating how first.

kjac:
Fixed in #6985

PR #6967, opened 2019-10-30
Remove unneccessary cursor:pointer on checkmark

kieronbos:
This fixes [6703](https://github.com/umbraco/Umbraco-CMS/issues/6703)

### Description
Editing css file umb-checkmark.less to remove cursor: pointer style, I searched all references to this class and think that i checked them all, and removing the style had no adverse affects on the interface.

Locations checked were tours, user group adding, user invite and creation, most cases simply didnt need cursor pointer, however user group adding did need it, but the parent anchor was providing the cursor.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-checkmark.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-checkmark.less
@@ -9,7 +9,6 @@
     justify-content: center;
     align-items: center;
     color: @gray-7;
-    cursor: pointer;
     font-size: 15px;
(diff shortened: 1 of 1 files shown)


kjac:
@kieronbos thank you for your investigation work on this one. However, #6985 has been accepted as the fix for this #6703. As `umb-checkmark` is built to work as a toggle (or toggle indicator, at least), it would be reasonable to believe it has been used by 3rd party plugins in a context where the pointer cursor is indeed required. 

Thank you once again. Keep up the good work 👏

PR #6970, opened 2019-10-30
V8: Make multiple media property editor same width as other property editors

kjac:
### Prerequisites

### Description

Media picker properties that allow multiselect take up the full width of the screen. This doesn't look terribly good when they're used alongside other properties that have the default max width of 800px:

image

There is a functional merit when you pick a lot of images, but still... it looks a bit awkward:

image

This PR enforces the same max width of 800px to the media picker:

image

Obviously this means it takes up more vertical space when you pick a lot of images, but I think it's worth it:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/property-editors.less
+++ b/src/Umbraco.Web.UI.Client/src/less/property-editors.less
@@ -232,6 +232,7 @@
 }
 .umb-mediapicker-multi > div {
     width:100%;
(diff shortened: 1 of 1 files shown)


nielslyngsoe:
Cool, I think its a good proposal! 👍  And if anyone seems to miss the full-width we could just consider making that an option, since this solution would be the best default.

poornimanayar:
Thanks for the work @kjac ! Merged :-)

PR #6979, opened 2019-10-30
Extend umb-checkbox to allow icons

BatJan:
### Prerequisites

### Description
I noticed that the rich text editor datatype was using the umb-checkbox directive but that there was a `<label>` wrapped around it, which should not be neccessary. But it was in order to ensure that clicking the associated config icon and text would tick the box.

Now that's not needed anymore since I have enabled the directive to accept an `iconClass` parameter, and when it's passed a value it will be rendered. if it's left empty it will not. Then things will look and act like we're used to.

Doing this fixes the markup pattern so we don't have `<label>` nested inside a `<label>` - Visually things look the same as before.

This can be tested in the "rich text editor" datatype 👍

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html
@@ -16,5 +16,12 @@
         </span>
     </span>
 
+    <i class="{{vm.iconClass}}" style="margin-left: 26px;" aria-hidden="true"></i>
+
     <span class="umb-form-check__text">{{vm.text}}</span>
 </label>
+
+
+<!-- TODO
+    1: Make sure to add the margin-left class on <i> if it should be displayed otherwise add it to the text element :thumbs
+-->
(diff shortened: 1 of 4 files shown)


src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html
```diff
@@ -16,5 +16,12 @@
         </span>
     </span>
 
+    <i class="{{vm.iconClass}}" style="margin-left: 26px;" aria-hidden="true"></i>
+
     <span class="umb-form-check__text">{{vm.text}}</span>
 </label>
+
+
+<!-- TODO
```
poornimanayar: Is this note needed here as you seem to have handled this in a further commit?

BatJan: Oops! Removed now :-)

BatJan:
@poornimanayar - Thank you for testing 😃 Removed the comment 👍

poornimanayar:
No worries my friend :-) Happy to help!

poornimanayar:
Thanks @BatJan ! Merged :-)

PR #6983, opened 2019-10-30
V8: Do not allow rollback of deleted content

kjac:
### Prerequisites

### Description

At the moment we allow rollback of content in the recycle bin. While it sort of works, it makes very little sense. And you also really quickly end up with something like this:

image

So... This PR removes the rollback button from elements in the recycle bin, and also (as a precaution) denies rollback of trashed items serverside.


poornimanayar:
👍 @kjac

emmaburstow:
Lovely work, as usual @kjac :)

PR #6315, opened 2019-09-09
V8: Accessibility Changes For umbEditorHeader Directive

RachBreeze:
This PR applies the accessibility amends made in https://github.com/umbraco/Umbraco-CMS/pull/6236 and https://github.com/umbraco/Umbraco-CMS/pull/5760 to the umEditorHeader Directive.

Specifically it aims to:

- Set the <title> attribute on the page based on the user's location
- Allow screen reader users to understand when they are creating an item what item they are creating
- Allow screen reader users to understand when they are editing an item what item they are editing

PR #6986, opened 2019-10-30
V8: Accessibility Changes For umbEditorHeader Directive (settings section)

RachBreeze:
This PR extends #6315 adding the screen reader properties and setting the page title for almost all of the remaining menu items in the settings section.   It leaves the language and the log viewer as the remaining items to update within this section


poornimanayar:
Hi @RachBreeze ,

Thank you, you know what happens next!

Poornima

RachBreeze:
Hi @kjac 
Yep I can look at the merge conflicts.
I did try and set the properties on `umb-editor-header` originally but it wouldn't pass through to the header controller

RachBreeze:
Hi @kjac 
Thank you once again for reviewing.
I've merged the language files, if you have a sec it would be good to know how I can set VS to keep the formatting the same as those in the xml files, even when I copy and paste the indentation I can't keep it the same :-(

kjac:
As for the language files, well... that's a longer story. 

At some point in the past someone formatted one or more of them in a manner that doesn't conform to the default indentation which 4 spaces, as defined in `.editorconfig`:

```
# Default settings:
# A newline ending every file
# Use 4 spaces as indentation
[*]
insert_final_newline = true
end_of_line = crlf
indent_style = space
indent_size = 4
```

As a result, VS fails miserably at indenting these files.

What I usually do is hit Ctrl+Z immediately after copying and pasting an entry in the language files if the indentation looks wrong. VS will usually undo it's failed auto-formatting then, leaving behind something that at least *looks* correct (even if it is wrong according to `.editorconfig`).

Logic would dictate that we should simply force format those files in a separate commit to `v8/dev` and be done with it. Unfortunately this would leave a whole lot of existing PRs with a merge conflict 😞

kjac:
Merged. Thanks again, @RachBreeze 👍

PR #6987, opened 2019-10-30
Removes continually loading banner if there is no current editor state

RachBreeze:
The loading banner continues to display when in the following sections:
- When the creating/editing users
- When the creating/editing user groups
- When the creating/editing languages
- When in the log viewer

The continually loading banner introduced in  #6315 and this PR removes it


nul800sebastiaan:
Whoops! Thanks for fixing it so quickly @RachBreeze ! 👍

PR #6991, opened 2019-10-30
Icon picker button: Fix accessibility

BatJan:
### Prerequisites

### Description
The button one clicks on a document type has been changed from `<div>` to `<button>` and the icon has been hidden from screen readers.

Looks the same as before the PR 😃


poornimanayar:
:+1 @BatJan

PR #6993, opened 2019-10-30
Editor service shouldn't overwrite the size if already set in the options

abjerner:
For most of the functions in the editor service, the `size` option is hardcoded, making it impossible to set another size.

Especially with the introduction of a `medium` size, it may be favorable to only set a size if one hasn't already been specified in the editor options.


poornimanayar:
Makes sense @abjerner! We shall review and get back to you :-)

Poornima

kjac:
..and merged!

PR #6994, opened 2019-10-30
User group picker: Fix accessibility issues

BatJan:
### Prerequisites

### Description
I noticed that the user group picker had a `ng-click` event on a `<div>` element. I added a `<button>` inside the div instead and moved the click event on to it and made some styling fixes so everything looks and works like it did before with the added bonus of some visually hidden texts indicating whether a group has been selected or not.

I also added a translation for a text that was not translated already.

user-group-picker


poornimanayar:
Good one @BatJan , do you want to resolve the merge conflicts though?

Poornima

BatJan:
Should be resolved now @poornimanayar 👍

kjac:
Merged 👍

PR #6995, opened 2019-10-30
V8: Automatically select uploaded media in media pickers

### LINKED ISSUE #6844: UX Media Picker - V8 : auto selection for the uploaded images and then scroll down to them
When uploading images from the device using the media picker, the users might not know if the uploaded images are really uploaded or not, because there was no action happened after uploading the images. 

as we can see in the below example, I am uploading some images to my media folder:

MediaPicker

What I think that would be better if applied on the above process:

- When a user upload an image, we can place the recent uploaded images at the top of the dialog or at least we can apply an auto scroll to the bottom of the dialog, so the user can see the uploaded images.

- The uploaded images should be selected by default because it would be an extra step especially when we upload many images, we have to select them again.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes (part of) https://github.com/umbraco/Umbraco-CMS/issues/6844

### Description

The media picker used to select media automatically when they were uploaded within the picker. Somewhere along the line this got broken.

This PR fixes it - and thus in part also fixes #6844.

With this PR applied, here's how the media picker works:

select-uploaded-media

This works both for media pickers in "multiple images" and in "single image" mode; for "single image" mode the last uploaded media is automatically selected.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js
@@ -1,7 +1,7 @@
 //used for the media picker dialog
 angular.module("umbraco")
     .controller("Umbraco.Editors.MediaPickerController",
-        function ($scope, mediaResource, entityResource, userService, mediaHelper, mediaTypeHelper, eventsService, treeService, localStorageService, localizationService, editorService) {
+        function ($scope, $timeout, mediaResource, entityResource, userService, mediaHelper, mediaTypeHelper, eventsService, treeService, localStorageService, localizationService, editorService) {
 
             var vm = this;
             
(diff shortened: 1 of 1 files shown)


poornimanayar:
Hi @kjac ,

Thank you once again! We will be in touch :-)

P

saifobeidat:
Hi @poornimanayar , sorry but could you please give me an update for this PR?

saifobeidat:
Hi @kjac , I tried your solution it works perfectly for a separate multi media picker, but I tried to use the media picker in RTE the same issue still exists.

poornimanayar:
Hi @kjac ,

I have tested the work, i can see the PR works as advertised. But the difficulty I have is when you are uploading media to a folder with a large number of images its still not quite visual enough for me, I have to scroll down all the way to the bottom. As @nul800sebastiaan says on the issue #6844 we either need a way to sort it or show the most recent uploads at the top or some visual messaging to say its been uploaded.

Poornima

kjac:
@poornimanayar agreed, it would be nice to change the sort order. However that change requires a significant refactoring of the entity system (there is no sorting atm), so I'd prefer to leave that in the capable hands of HQ.

saifobeidat:
Hi @kjac , does your solution works for 8.4.0 ?

kjac:
@saifobeidat nope, this PR hasn't been merged in yet.

emmaburstow:
Thank you for the work, Kenn!

PR #7000, opened 2019-10-30
Added dialog aria labelling

Matthew-Wise:
Added aria role for dialog with label and description.

This provides context of the dialog. However I couldn't workout a way to focus the dialog on click. (so you need to tab after selecting permissions

Also this could be done for the other dialogs but felt getting the focus working with the help of some others was a better way to go.


poornimanayar:
Whooooaaaaaa @Matthew-Wise , proud contributor of the 7000th PR(I assume thats gramatically correct!) I have given it a quick test and it works. But you are right, you need to retain the focus on the dialog. Also noted that once you add permissions to a group you cannot tab to edit/remove it the focus goes from the Set permissions for.. button to Close button upon tabbing.

poornimanayar:
Hmm looks like the Edit/Remove by the group are anchor tags, I might submit a PR to change it to link buttons.

PR #7002, opened 2019-10-31
Fixes a race condition with nucache

Shazwazza:
There is 2x issues fixed here:

## Race Condition

Is it possible that the cache file is empty but the database is not? YES... (well, it used to be possible until this fix)
* A new file is created when one doesn't exist, this will only be done when `MainDom` is acquired
* The new file will be populated as soon as `LoadCachesOnStartup` is called
* If the appdomain is going down the moment after `MainDom` was acquired and we've created an empty cache file, then the MainDom release callback is triggered and executes on a different thread, which will close the file and set the cache file reference to `null`. At this moment, it is possible that the file is closed and the reference to the file set to `null` **before** `LoadCachesOnStartup` which would mean that the current appdomain would load in the in-mem cache via DB calls and would not attempt to write back to the cache file during site operation, **but** this now means that there is an empty cache file which will be loaded by the next appdomain and it won't check if it's empty, it just assumes that since the cache file is there, that is correct.

This fixes this by:

* We'll return `false` from `LoadContentFromLocalDbLocked` if there are no items in the local cache file. This is a safety check. The site could very well have zero content items, but that's ok. This check will at least force the content to be loaded in from the database instead of the cache file. 
* Fixed the locking of the PublishedSnapshotService initialization to avoid the race condition mentioned above. This also means that the safety check above shouldn't be needed but we'll leave it there anyways. The lock fix is: 
  * We were locking on `_storesLock` to populate the cache, but not when we were creating the cache files. We were also locking on the same `_storesLock` when releasing from MainDom. The problem was we were not using the same lock to lock the whole sequence from creating a new cache file to populating the cache. This PR fixes that, this lock will be used when creating or opening a cache file and then populating the cache which will prevent the releasing/closing of the cache file before the cache is populated

## SetAllFastSorted 

`SetAllFastSorted` was also used when loading nucache from the database, it was intended originally to only be used when loading nucache from the cache files, therefore it would never also update the nucache file. This was an oversight when using SetAllFastSorted when loading from the database since when doing that we weren't writing back to the cache file ... fixed in this PR

### DIFF
--- a/src/Umbraco.Web/Scheduling/SchedulerComponent.cs
+++ b/src/Umbraco.Web/Scheduling/SchedulerComponent.cs
@@ -18,6 +18,11 @@ namespace Umbraco.Web.Scheduling
 {
     internal sealed class SchedulerComponent : IComponent
     {
+        private const int DefaultDelayMilliseconds = 180000; // 3 mins
+        private const int OneMinuteMilliseconds = 60000;
+        private const int FiveMinuteMilliseconds = 300000;
+        private const int OneHourMilliseconds = 3600000;
+
         private readonly IRuntimeState _runtime;
         private readonly IContentService _contentService;
         private readonly IAuditService _auditService;
@@ -111,7 +116,7 @@ namespace Umbraco.Web.Scheduling
         {
             // ping/keepalive
             // on all servers
-            var task = new KeepAlive(_keepAliveRunner, 60000, 300000, _runtime, _logger);
+            var task = new KeepAlive(_keepAliveRunner, DefaultDelayMilliseconds, FiveMinuteMilliseconds, _runtime, _logger);
             _keepAliveRunner.TryAdd(task);
             return task;
         }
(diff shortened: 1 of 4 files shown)


src/Umbraco.Web/Scheduling/SchedulerComponent.cs
```diff
@@ -120,7 +125,7 @@ private IBackgroundTask RegisterScheduledPublishing()
         {
             // scheduled publishing/unpublishing
             // install on all, will only run on non-replica servers
-            var task = new ScheduledPublishing(_publishingRunner, 60000, 60000, _runtime, _contentService, _umbracoContextFactory, _logger);
+            var task = new ScheduledPublishing(_publishingRunner, DefaultDelayMilliseconds, OneMinuteMilliseconds, _runtime, _contentService, _umbracoContextFactory, _logger);
```
bergmania: Is this change from 60000 ms to 180000 ms intended?

Shazwazza: Yes, we start these background tasks far too quickly after startup, sometimes on cloud it takes a minute to even boot (depending on what folks are doing with their code) and we should initiate BG tasks before boot is fully complete. In seb's logs I saw calls for running through all of the healthchecks before the whole boot process was done which is insane, especially when you consider overlapping app domains and multiple restarts.

PR #7003, opened 2019-10-31
V8: Fix button style when opening doctype in infinite editing with unsaved changes

kjac:
### Prerequisites

### Description

The button style should probably be `warning` for the dialog that alerts you how changes will be discarded if you open the doctype in infinite editing without saving your changes first:

image

So... this PR updates the button style:

image


poornimanayar:
Thanks for this @kjac! 

Poornima

PR #7004, opened 2019-10-31
V8: Mark the current variant as dirty when its name changes

kjac:
### Prerequisites

### Description

Content variants have an *isDirty* flag that's set whenever the variant content changes. Unfortunately it doesn't work when the variant name changes - the code that's put in place to ensure this is broken due to some componentization of the backoffice.

The issue is most noticeable if you change the content name and then go to edit the doctype without saving your changes. You ought to be alerted that any unsaved changes will be undone, but you're not:

variant-name-dirty-before

Efficiently this means that you'll waste all the precious time you spend changing content names before editing doctypes 😲 or worse, if something a bit more critical starts depending on the `isDirty` flag (perhaps a content app?).

This PR obviously fixes the broken code, enforcing `isDirty` whenever the content name changes:

variant-name-dirty-after


poornimanayar:
Hi @kjac ,

To be reviewed and replied :-) Watch this space!

Poornima

PR #7006, opened 2019-10-31
V8: Fix Nested Content configuration overflow on smaller screens

kjac:
### Prerequisites

### Description

The new Nested Content configuration doesn't work very gracefully on smaller screens:

image

Also there's really no need to invent new styling etc. for displaying an icon and a label, as we have the `umb-node-preview` for just that... so this PR cleans up the code a little bit and ensures we handle smaller screens gracefully:

image


poornimanayar:
Thanks @kjac !

Poornima

PR #7011, opened 2019-10-31
Accessibility improvements to Multivalues view

poornimanayar:
Currently in the `multivalues.html` view the Remove option is an `<a>`. I have changed it to a `button`. 

I have also made the drag and drop hide on-screen readers using `aria-hidden=true`. Don't know whether that is okay on second thoughts..

To test, add prevalues to the following data types and make sure the Remove button can be tabbed into using keyboard. 
- Checkbox list
- Radiobutton list
- Dropdown
- Dropdown multiple
These are the only ones I can think of now. Color Picker has its own prevalues screen.

An example of how it should work..
prevalues


kjac:
This works brilliantly, @poornimanayar ... merged, thank you 👏

PR #7014, opened 2019-10-31
Accessibility improvements & clean up of mini listiview

poornimanayar:
In the mini listview I have made some improvements. I have cleaned up and moved some inline styling to stylesheets, added some fallback localization text and made the icons hidden on screen readers.

I think it will be good if #6997 can also be merged in together with this as it addresses another mini listview issue :-)


poornimanayar:
Hmm, merge gone bad I think. let me look at it again

poornimanayar:
Hi @kjac ,

I have made the `nth-child` a class as you suggested but the table cell with `display:none` seems to be used in rendering as the table body seems to have two cells as well and removing the empty one is causing it to break badly.

Poornima

poornimanayar:
I couldnt agree with you more @kjac , I think i need to delete my local `v8/contrib` and fetch it all over again.  The worst part is I could not see these locally. So something is definitely wrong. I will start this PR all over again.

poornimanayar:
PR re-done in https://github.com/umbraco/Umbraco-CMS/pull/7398

kjac:
@poornimanayar really weird things going on indeed then :) make sure you close this one when you have a new one ready.

kjac:
...oh shoot I was too quick to comment there 😆

PR #7016, opened 2019-10-31
Set permission button accessibility fixes

BatJan:
### Prerequisites

### Description
I noticed that the "Set permission" button could use some accessibility improvements so in this PR I have
- Added the deep-blur directive meaning that the popup will disappear when it's tabbed out of
- Added the aria attributes for `aria-haspopup` and `aria-expanded`
- Refactored the `<a>` elements in the dropdown to `<button>`
- Hidden the `<i>` element using `aria-hidden="true`

**Before**
permission-dropdown-before

**After**
permission-dropdown-after


poornimanayar:
@BatJan  👍 

Poornima

kjac:
...and merged 👍

PR #7017, opened 2019-10-31
V8: Make macro partial view mandatory

kjac:
### Prerequisites

### Description

When you create new macros, the macro partial view isn't set. If you then allow those macros in the RTE and enabled rendering within the RTE, this is what happens:

image

Of course it makes no sense to have a macro without a view, but for newcomers that kind of error probably isn't terribly comforting. So this PR makes it mandatory to pick a view for a macro.

image

Obviously the macro can still be created without the view (that's how it's created after all), but at least you can't enable it for RTE without picking a view.


poornimanayar:
Keen eye to detail as usual @kjac ! Talk soon!

Poornima

poornimanayar:
And merged :-)

PR #7020, opened 2019-10-31
Create Content Blueprints accessibility improvements

poornimanayar:
With this PR I have targeted the following
- Accessibility improvement by hiding the icon to screen readers
- Clean up moving inline style to stylesheet


skttl:
Doc Type Grid Editors uses element type content templates. I don't see why this should be disabled. Could come in handy for Nested content and Proteus too 🙂

poornimanayar:
@skttl Yes, I was unsure as well, but always worth knowing opinions ;-) Happy to revert that change..

poornimanayar:
I have re-enabled the element type on content blueprint as there were reservations about it :-)

bjarnef:
@poornimanayar ensure also to add `type` attribute since browsers may have different default value for type attribute on button element.
https://stackoverflow.com/a/31644856

The `prevent-default` directive on the button element can also be removed.

abjerner:
Having the `prevent-default` attribute makes this work, so it gives the same end result as having `type="button"`.

But from a semantic point of view, it would make good sense to change it to `type="button"`. @poornimanayar is this something you could create another PR for?

poornimanayar:
Yes, I am sorry I missed out on the `type` attribute. I will be doing the fixes for both as a PR

PR #7022, opened 2019-10-31
Fixed casing for the "noTemplate" option

abjerner:
The option should be named using camel casing similar to pretty much all other options. The fix is made so either "notemplate" or "noTemplate" can be specified, making the fix backwards compatible instead of creating a breaking change.

While I had the controller open, I also fixed so the `id` option is no longer required when the `create` option is is set to `true`. When the latter is the case, the ID is automatically set to `-1` if not already specified.

The PR can be tested by running the following code:

```js
$scope.danger = function() {
    editorService.documentTypeEditor({
        create: true,
        noTemplate: true
    });
};
```


poornimanayar:
Thanks for the PR @abjerner , we will be in touch!

Poornima

kjac:
Merged. Thanks again, @abjerner ⭐️

PR #7023, opened 2019-10-31
Improved the ngdocs for the editor service

abjerner:
- The `id` and `create` options where wrongly specified as being of type `Callback`. They are now `Number` and `Boolean` respectively

- The `noTemplate` was added. This change depends on #7022 being merged.

- The `isElement` and `allowVaryByCulture` was added following #6779

image


poornimanayar:
👍 @abjerner

kjac:
Merged along with #7022 😄 thanks again, @abjerner

PR #7029, opened 2019-10-31
Angular API documentation: Adds styling for <code> elements

abjerner:
`<code>` elements in the descriptions are now properly styled 🎉 

The `gulp-ngdocs` task will by default remove all styling from `<code>` elements - probably because it uses them throughout the generated documentation where the styling doesn't make sense (eg. in the `<h1>` element). This PR will therefore only target `<code>` elements in the list of methods.

## Before

image

## After

image


poornimanayar:
Thanks again @abjerner

poornimanayar:
and merged

PR #7030, opened 2019-10-31
Responsive Angular API documentation

abjerner:
As the Angular API documentation uses nested tables for listing options for the options and parameters, the space quickly feels a bit cramped.

With this PR I'm making the documentation fluid. The documentation site uses Bootstrap for creating columns, but only seems to use `.span3` and `.span9`, so I've only updated those to be responsive. The other `.span{n}` classes are still fixed with.

I've set the container to a maximum width of 1500 pixels, so the site doesn't become super wide on wide monitors.

This PR will have a merge conflict with #7029. In my head it make sense to make the site responsive, so as I'm not 100% this will be merged, I've created the changes as two PRs instead of one 😮

## Before

image

## After

With a browser window with a width of 1700 pixels, the same block now looks like:

image

### DIFF
--- a/src/Umbraco.Web.UI.Docs/umb-docs.css
+++ b/src/Umbraco.Web.UI.Docs/umb-docs.css
@@ -7,6 +7,21 @@ body {
     
 }
 
+.container, .navbar-static-top .container, .navbar-fixed-top .container, .navbar-fixed-bottom .container {
+    max-width: 1500px;
+    width: 95%;
+}
+
+.span3 {
+    width: 220px;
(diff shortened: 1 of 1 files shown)


poornimanayar:
Thanks @anders! Speak to you soon!

poornimanayar:
This definitely looks better. @nul800sebastiaan, can I merge this?

nul800sebastiaan:
@poornimanayar go for it 👍

poornimanayar:
Merged @abjerner ! Thanks for the work as always :-)

PR #7031, opened 2019-10-31
V8: Make it possible to hide anchor/querystring input in the linkpicker

kjac:
### Prerequisites

### Description

I find some editors get confused with the "Anchor / querystring" input field in the linkpicker:

image

And more often than not the editors don't really need this option at all. So this PR makes it possible to remove "Anchor / querystring" from the linkpicker.

You must explicitly select to do so in the confguration (thus it's also backwards compatible):

image

Once enabled the "Anchor / querystring" input disappears from the linkpicker:

image


kjac:
@poornimanayar that's a really excellent question!

The RTE linkpicker will certainly still have the anchor/querystring input, unless we actively do something to remove it. It might be worth looking into a configuration option on the RTE config along the same lines as this PR.

poornimanayar:
Yes, that makes sense! I But cant it get a bit more complicated given we can add the "Insert Link" from the RTE datatype but then you need to pay with the config to make this take effect. IMO it should be something that can be configured on the data type itself rather than in the config. What do you think?

poornimanayar:
@kjac ,

Can you fix merge conflicts against `v8/contrib` on this one?

Poornima

kjac:
@poornimanayar all done 👍

ronaldbarendse:
One thing to keep in mind when hiding the anchor/query string input: if you enter a URL with either, it's saved within the URL and when the link is edited again, the anchor/query string is split from the URL and stored in the hidden anchor field, making it uneditable!

I've fixed this in PR #7533 (commit https://github.com/umbraco/Umbraco-CMS/pull/7533/commits/7c4603d4694956fe9fbe7cba4d4578136f56adab).

PR #7036, opened 2019-11-01
Property-Action for Remove All Entries in Nested Content

nielslyngsoe:
**Feel free to finish it!**

Remove all entries from Nested Content (Draft PR).

With this feature, I would like editors to be able to remove all entries of a Nested Content property.

image

The action will be available through Property Actions, and the action should be confirmed via a Confirm Dialog before begin executed.

I didn't have the time to finish it, so feel free to make it happen.

Thanks


kjac:
Ooh neat! I've got a bit of time on my hands, I'll have a look.

kjac:
PR in #7039

PR #7038, opened 2019-11-01
Fix ContentQuery.Search not skipping results and return wrong total records

ronaldbarendse:
While using the `ContentQuery.Search()` methods with `skip > 0`, I've noticed it doesn't skip the results and only increases the amount of returned results:

https://github.com/umbraco/Umbraco-CMS/blob/fca74ad8d7e72e4a5981cd90e4725ad0f7cd4a82/src/Umbraco.Web/PublishedContentQuery.cs#L192-L234

https://github.com/umbraco/Umbraco-CMS/blob/fca74ad8d7e72e4a5981cd90e4725ad0f7cd4a82/src/Umbraco.Web/PublishedContentQuery.cs#L243-L251

This is a functional breaking change, because if you're currently manually skipping the results you won't get any results anymore:
```c#
ContentQuery.Search("test", 10, 20, out _).Skip(10)
```

As both methods don't expose the `ISearchResults`, manually skipping would still allocate the result objects (and we [don't want that](https://shazwazza.com/post/paging-with-examine/)). Because of this, I think it's justified to introduce this change.


src/Umbraco.Web/PublishedContentQuery.cs
```diff
@@ -191,46 +191,52 @@ public IEnumerable<PublishedSearchResult> Search(string term, string culture = "
         /// <inheritdoc />
         public IEnumerable<PublishedSearchResult> Search(string term, int skip, int take, out long totalRecords, string culture = "*", string indexName = null)
```
ronaldbarendse: We can already set the default argument value, so the used index is visible to callers:

```suggestion
         public IEnumerable<PublishedSearchResult> Search(string term, int skip, int take, out long totalRecords, string culture = "*", string indexName = Constants.UmbracoIndexes.ExternalIndexName)
```

The check within the method should be kept, so passing `null` explicitly still uses the external index... Thoughts?

src/Umbraco.Web/PublishedContentQuery.cs
```diff
@@ -191,46 +191,52 @@ public IEnumerable<PublishedSearchResult> Search(string term, string culture = "
...
 
             if (!_examineManager.TryGetIndex(indexName, out var index) || !(index is IUmbracoIndex umbIndex))
+            {
                 throw new InvalidOperationException($"No index found by name {indexName} or is not of type {typeof(IUmbracoIndex)}");
+            }
 
             var searcher = umbIndex.GetSearcher();
 
-            // default to max 500 results
-            var count = skip == 0 && take == 0 ? 500 : skip + take;
```
ronaldbarendse: There's no need to also have a hardcoded `maxResults` here, so I've changed it to be the same as the overload below.

ronaldbarendse:
I've found another bug: the `totalRecords` also includes media results (folders)! This completely breaks paging, as the total returned results (containing only content) might be lower than this value...

ronaldbarendse:
When executing `ContentQuery.Search("new", 0, 2, out var totalRecords)`, I only get 1 result, but `totalRecords` is 3.

I can confirm this within the back-office; it returns 3 records, but only 1 is content and the other 2 are media:

ExternalIndex


afbeelding

afbeelding

afbeelding

ronaldbarendse:
Manually creating the query with `searcher.CreateQuery("content")` also doesn't add a filter on the `__IndexType` field and still returns media! Only after adding `Field("__IndexType", "content")`, the returned results are exclusively content types.

@Shazwazza Any thoughts, as it seems most search overloads are completely broken (including those within the `Examine` package)?

src/Umbraco.Web/PublishedContentQuery.cs
ronaldbarendse: This should then also be done on `IPublishedContentQuery`, as the interface is used/visible to callers 😉

ronaldbarendse:
I've made the neccesary changes within Umbraco to make sure only content is searched and the correct total record count should be returned (as the `PublishedSearchResult` uses `IPublishedContentCache` and thus only returns content), but this currently depends on a bug within Examine to be fixed first: https://github.com/Shazwazza/Examine/issues/143.

ronaldbarendse:
I've updated the dependency on Examine to require version 1.0.2, so the search query category/index type is used and the correct amount of total records is returned. Thanks @Shazwazza for fixing this bug and releasing a new version so fast!

Because `PublishedSearchResult` can also be used for media and members, the `ToPublishedSearchResults()` extension method now also accepts an `IPublishedSnapshot` that uses the correct cache depending on the category/index type. This allows `ContentQuery.Search(IQueryExecutor query)` to return not only content, but also media or (depending in the used index) members.

ronaldbarendse:
Looks like the sorting also needs to be fixed, as it currently does this in the `ToPublishedSearchResults()` extension method. This will only sort the items of the current page, besides not being as efficient as Lucene and possibly even overrides any existing sorting!

ronaldbarendse:
I've removed the sorting within the `ToPublishedSearchResults()` extension method, as the results were already sorted on score by default.

Both `skip` and `take` parameters now also have some additional validation, so negative values are not allowed, which could trip up the results:
- A negative `skip` would throw an exception when iterating the `ISearchResults`;
- A negative `take` would return the maximum amount of results `LuceneSearcher.MaxDoc`:
https://github.com/Shazwazza/Examine/blob/864168c7f019908242dd56197e075edea5ee77cf/src/Examine/LuceneEngine/LuceneSearchResults.cs#L76.

ronaldbarendse:
@nul800sebastiaan or @Shazwazza: Can you take a look at the changes and provide any feedback? I don't think this PR needs more to get merged... And as it fixes some major bugs in Umbraco, it would be great to ship this ASAP 😄

Shazwazza:
@ronaldbarendse this all works great, thanks a ton for the effor this is a much needed fix.

src/Umbraco.Web/PublishedContentQuery.cs
```diff
@@ -191,46 +191,44 @@ public IEnumerable<PublishedSearchResult> Search(string term, string culture = "
...
             if (culture == "*")
             {
-                //search everything
-
-                results = searcher.Search(term, count);
+                // Search everything
+                queryExecutor = query.ManagedQuery(term);
             }
-            else if (culture.IsNullOrWhiteSpace())
+            else if (string.IsNullOrWhiteSpace(culture))
```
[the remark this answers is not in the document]
ronaldbarendse: Because the extension method doesn't do anything else 😉 And the static method is always available (with auto-completion), without requiring the `Umbraco.Core` namespace or even Umbraco itself...

I would argue this makes it more readable, as it uses standard C# methods and you're sure it won't throw a `NullReferenceException` (with seems to be the biggest 'issue' of using this type of extension method: https://codereview.stackexchange.com/questions/6602/helper-class-for-null-and-empty-checks).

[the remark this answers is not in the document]
Shazwazza: Sure if you don't like it then don't use it ... but you changed what was already there :P anyways, it's fine.

ronaldbarendse:
@nul800sebastiaan I see this PR doesn't have the breaking change label and therefore also isn't mentioned in the [8.6 release blogpost](https://umbraco.com/blog/umbraco-86-release/).

Just like the cleaned up exceptions, this contains functional changes that might affect custom code:
- Manually skipping the results should be removed when using the overload accepting an amount to skip, as this is now fixed and works as expected. So `ContentQuery.Search("test", 10, 20, out _).Skip(10)` will yield zero results in Umbraco 8.6.0 and this might not be that easy to spot, as it's only visible when skipping items (e.g. on the second 'page' of results).
- The total records returned might be less, as it previously also included media, even though media items weren't part of the results, as it only used the content cache.
- The results aren't sorted per page anymore, as that would override any sorting done within Examine (see https://github.com/umbraco/Umbraco-CMS/pull/7038#issuecomment-570824492).
- Using invalid `skip` or `take` arguments will now throw `ArgumentOutOfRangeException`s before executing the search (which might otherwise have thrown an exception when iterating the results or return the maximum amount of results, see https://github.com/umbraco/Umbraco-CMS/pull/7038#issuecomment-570952926).
- Examine is updated to ensure the supplied category is used: https://github.com/Shazwazza/Examine/issues/143.

nul800sebastiaan:
Okidoki, added the breaking label so that will appear on the release notes on Our soon.

Shazwazza:
seems like every bug fix can end up being a 'breaking change' :P

https://xkcd.com/1172/

Issue #2925, opened 2018-09-06
Nested Content PropertyValueConverter is missing DefaultPropertyValueConverter attribute


ronaldbarendse:
I've created this issue to surface and centrally discuss PRs #2918 and #2922.

Creating a custom `PropertyValueConverter` for the built-in `Umbraco.NestedContent` editor is not possible, because the built-in PVCs are not marked with a `DefaultPropertyValueConverter` attribute.

Reproduction
------------
Create a custom PVC for `Umbraco.NestedContent`: instead of overriding the default ones, an error is thrown indicating more than 1 PVC is registered.

### Specifics
This happens on all released versions of Umbraco 7.7+, when Nested Content was integrated. The exact exception is:

`InvalidOperationException: Type 'Umbraco.Web.PropertyEditors.ValueConverters.NestedContentManyValueConverter' cannot be an IPropertyValueConverter for property 'socialMedia' of content type 'home' because type 'TestProject.Web.Core.NestingContentlyManyValueConverter' has already been detected as a converter for that property, and only one converter can exist for a property.. This usually indicates that the content cache is corrupt; the content cache has been rebuilt in an attempt to self-fix the issue.`

ronaldbarendse:
I've updated both PR #2918 and #2922 to fix an issue with the `JsonValueConverter` (as that converter automatically registers itself for all properties which store their value as JSON - and as both are now default converters, the same conflict happened as described in the issue).

Although PR #2922 changes public classes (by merging `NestedContentSingleValueConverter` and `NestedContentManyValueConverter` into a single `NestedContentValueConverter` class) and that should be marked as a breaking change, I think that it's justified and should be the real fix for this issue.

Having a single converter for a single property editor also makes more sense. And especially now we can override the default converter, it's way more convenient to do so. For example, if you'd like to filter out hidden/disabled nested content items (by adding a `umbracoNaviHide` checkbox or better yet, using https://github.com/nathanwoulfe/NestingContently), it's this simple:

```c#
public class IsVisibleNestedContentValueConverter : NestedContentValueConverter
{
    public override object ConvertDataToSource(PublishedPropertyType propertyType, object source, bool preview)
    {
        var obj = base.ConvertDataToSource(propertyType, source, preview);
        if (obj is IPublishedContent content && !content.IsVisible())
        {
            return null;
        }
        else if (obj is IEnumerable<IPublishedContent> contents)
        {
            return contents.Where(c => c.IsVisible());
        }

[...]

PR #2918, opened 2018-09-05
Nested Content PropertyValueConverter is missing DefaultPropertyValueConverter attribute

ronaldbarendse:
Creating a custom `PropertyValueConverter` for the built-in `Umbraco.NestedContent` editor is not possible, because the built-in PVCs is not marked with a `DefaultPropertyValueConverter` attribute.

This functionality (and the attribute) were added in issue [U4-3591](http://issues.umbraco.org/issue/U4-3591), but because NC was added later and originated from a community package, marking it as a default PVC was probably overlooked.

I've also converted the `IPropertyValueConverterMeta` interface to attributes, since the values are static. Implementing the interface would be the way to go if the PVCs (`NestedContentSingleValueConverter` and `NestedContentManyValueConverter`) were merged into a single class, but changing that would be a breaking change.

Connected to #2925

PR #7045, opened 2019-11-01
Add DefaultPropertyValueConverterAttribute to core Nested Content PropertyValueConverters

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/2925.

### Description
The is the V8 version of PR https://github.com/umbraco/Umbraco-CMS/pull/2918. By adding the `DefaultPropertyValueConverterAttribute` to the PVCs, adding a custom one doesn't require [manually removing them](https://gist.github.com/stevemegson/33c24a4b36bf9d7381aa4dc9cc01b117).

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/ValueConverters/NestedContentManyValueConverter.cs
+++ b/src/Umbraco.Web/PropertyEditors/ValueConverters/NestedContentManyValueConverter.cs
@@ -6,6 +6,7 @@ using Newtonsoft.Json.Linq;
 using Umbraco.Core.Logging;
 using Umbraco.Core.Models.PublishedContent;
 using Umbraco.Core.PropertyEditors;
+using Umbraco.Core.PropertyEditors.ValueConverters;
 using Umbraco.Web.PublishedCache;
 
 namespace Umbraco.Web.PropertyEditors.ValueConverters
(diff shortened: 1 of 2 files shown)


poornimanayar:
Hey @ronaldbarendse ,

I remember having a conversation on this one some time back! Thank you again for your work, speak to you soon!

Poornima

ronaldbarendse:
Any progress on merging these (quite small) changes?

poornimanayar:
Hi @ronaldbarendse ,

Apologies for the delay. We have been having chats within the team around this one. And one question that came up was how breaking this is for people who have their own workaround? Can you put down your thoughts around that here?

Poornima

poornimanayar:
@ronaldbarendse , Can you also add any code to test this alongside any notes here?

ronaldbarendse:
You can test this with the following gist: https://gist.github.com/stevemegson/33c24a4b36bf9d7381aa4dc9cc01b117.

Currently you're required to remove the built-in PVCs (in the `NestedContentComposer`), as you'll otherwise receive an `InvalidOperationException` (see one of the comments). With this PR applied, removing isn't needed anymore, as Umbraco knows the built-in PVCs are the default ones and should be automatically replaced with the custom ones.

To verify this works correctly, it should not throw the exception if you remove the `NestedContentComposer` from the gist and still use the custom PVCs (add a breakpoint to `GetPropertyCacheLevel()` and check if this is hit when rendering NC content on a page). To check doesn't break anything, it should still work with the `NestedContentComposer`, even if it doesn't remove anything, as they're already removed/replaced.

poornimanayar:
And merged! Thanks for your work as always @ronaldbarendse

PR #7049, opened 2019-11-04
V8: Reset paging in user search when free text searching and filtering

### LINKED ISSUE #7047: Free-text filtering users when on page 2+ doesn't show results
If you have 2 pages of users, go to page 2 and then start filtering. It will say there are several results but they won't show. Presumably because it's trying to show page 2 of the filtered results and there are less than 1 page worth of results.

Filtering should reset the page back to page 1 so that it shows the first page of the filtered results.

## Umbraco version

I am seeing this issue on Umbraco version: 8.1.5

### Steps to reproduce

Create more than 1 page of users.
Go to users list
Switch to page 2
Enter text to filter users so that less than 1 page of results are shown

### Expected result

I expected to see the users that matched the filter

### Actual result

I see no results but I do get the correct "Users (2)" in the top left

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7047

### Description

As described in #7047 - if you have multiple pages of users (more than 25 users), the paging doesn't reset when you perform free text search for users.

The problem isn't isolated to free text, however; it also happens when filtering users:

users-search-pagination-before

This PR fixes it all by resetting the current page number when the free text search input or any filter changes:

users-search-pagination-after


poornimanayar:
haaaaaa, its one of those errors! Thanks @kjac !

kjac:
Updated with latest v8/dev

poornimanayar:
@kjac , Before I merge this once can you cast your eyes quickly over this PR. Just because I see a lot of commits once i changed the base to `v8\contrib`

kjac:
@poornimanayar you're right; this one needs to wait 'till we have `v8/dev` merged to `v8/contrib`

kjac:
@poornimanayar I have merged `v8/contrib` into the PR again, it looks good now 👍

PR #7057, opened 2019-11-05
Only enumerate the app plugins folder if it exists

bergmania:
In MB replacement, we have removed the MB plugin, and therefore the folder do not exists..

When having a new user, the tours fail without this fix.

### Considerations
- Do we need to handle this elsewhere?
- Should we have a dummy file to ensure the folder exists?


Shazwazza:
All good. Have double checked in the core to see if we need other checks on this folder and all checks are already taken care of as far as the usage of `SystemDirectories.AppPlugins` goes. I don't think we should ship an empty directly with a dummy file. I think this change is all that's needed.

PR #6634, opened 2019-10-09
Adds nametemplates for grid editors

skttl:
#6633

### Description
This adds name templates to grid editors, that you can use like in Nested Content.

I have added some templates to the default grid editors too, as an example.

If there is no template available, the name will be used instead.

billede

PR #7059, opened 2019-11-05
Grid editor name templates

skttl:
### Prerequisites

If there's an existing issue for this PR then this fixes #6634 (https://github.com/umbraco/Umbraco-CMS/pull/6634#issuecomment-549643188)

### Description
This one changes the logic in name templates a bit, moving the editor name part out of the template, so it is alwasy visible.

In addition to that I added some null checks to the default editor name templates.

To test, add a new image to the grid. Notice no errors in the console, and the name will just be "Image". When you add an image, the name will be "Image: [the name of the image]"


bergmania:
Thanks @skttl..

The commits are part of `umbraco:v8/dev`, so I will close the PR.

I don't understand why github says there still are changes..

Issue #7052, opened 2019-11-04
Image cropper issues


Luksor:
Image cropper is broken on 8.2.1.

## Umbraco version
* Umbraco 8.2.1

Reproduction
------------

### Bug summary

When saving a doctype with a configured cropper I'm getting "System.NotSupportedException: The given path's format is not supported."

### Specifics

Tested on clean Umbraco installation, with the starter kit installed.

### Steps to reproduce

* Install Umbraco with the starter kit.
* Replace the property editor for the Photo field in the Person doctype to an Image Cropper and define a crop.
* Upload an image to it and set a crop.

### Expected result

* Document is saved.

### Actual result

* Exception is thrown.

Stack trace:
````
at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
   at System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPath)
   at Umbraco.Core.IO.PhysicalFileSystem.GetFullPath(String path) in D:\a\1\s\src\Umbraco.Core\IO\PhysicalFileSystem.cs:line 301
   at Umbraco.Core.IO.PhysicalFileSystem.DeleteFile(String path) in D:\a\1\s\src\Umbraco.Core\IO\PhysicalFileSystem.cs:line 219
   at Umbraco.Core.IO.ShadowWrapper.DeleteFile(String path) in D:\a\1\s\src\Umbraco.Core\IO\ShadowWrapper.cs:line 170
   at Umbraco.Core.IO.FileSystemWrapper.DeleteFile(String path) in D:\a\1\s\src\Umbraco.Core\IO\FileSystemWrapper.cs:line 73
   at Umbraco.Web.PropertyEditors.ImageCropperPropertyValueEditor.FromEditor(ContentPropertyData editorValue, Object currentValue) in D:\a\1\s\src\Umbraco.Web\PropertyEditors\ImageCropperPropertyValueEditor.cs:line 117
   at Umbraco.Web.Editors.ContentControllerBase.MapPropertyValuesForPersistence[TPersisted,TSaved](TSaved contentItem, ContentPropertyCollectionDto dto, Func`3 getPropertyValue, Action`3 savePropertyValue, String culture) in D:\a\1\s\src\Umbraco.Web\Editors\ContentControllerBase.cs:line 94
[...]

PR #7061, opened 2019-11-05
V8: Make upload and image cropper interchangeable

kjac:
### Prerequisites

### Description

While investigating #7052 I realized that the upload and image cropper datatypes aren't completely interchangeable:

- When changing a property from upload to image cropper, things usually go just fine.
- When changing a property from image cropper to upload, things go horribly wrong:
image

This PR makes the two datatypes interchangeable - or rather, it enables us to swap an image cropper with an upload datatype, as it already works the other way around.

Here's how it works when changing from image cropper to upload:

upload-imagecropper-interchangeable-1

...and for good measure, here's the other way around:

upload-imagecropper-interchangeable-2


emmaburstow:
Heyyyy Kenn, thanks for this work. Sorry it's taken a while for me to say so. We'll check it out and let you know if we need changes.

Em

PR #7066, opened 2019-11-06
New service/repo endpoints for efficient paging of entities based on relations data

Shazwazza:
This adds new service endpoints:

* IRelationService.GetPagedParentEntitiesByChildId
* IRelationService.GetPagedChildEntitiesByParentId

Which allows for efficient single query paging to get entities based on relationship data (parent or child ids). We could not do this before since the only way to get entity data for relation data was to first query for all relations for parent or child ids and then pass all of those relations in to get the entities. This was innefficient and also didn't support paging.

To make this work required some plumbing changes:

* The EntityRepository has been updated to allow a mixed collection of entity types to be returned, previously it only allowed for a single entity type to be queried and returned
* The EntityRepository.GetPagedResultsByQuery has a new overload that allows for a sub-query of relations data
* The EntityRepository now uses mostly Left Joins for it's data because it now supports mixed entity return types

links [AB#3449](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3449)

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/IEntityRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/IEntityRepository.cs
@@ -55,8 +55,9 @@ namespace Umbraco.Core.Persistence.Repositories
         /// A collection of mixed entity types which would be of type <see cref="IEntitySlim"/>, <see cref="IDocumentEntitySlim"/>, <see cref="IMediaEntitySlim"/>,
         /// <see cref="IMemberEntitySlim"/>
         /// </returns>
-        IEnumerable<IEntitySlim> GetPagedResultsByQuery(IQuery<IUmbracoEntity> query, Guid[] objectTypes, long pageIndex, int pageSize, out long totalRecords,
-            IQuery<IUmbracoEntity> filter, Ordering ordering);
+        IEnumerable<IEntitySlim> GetPagedResultsByQuery(
+            IQuery<IUmbracoEntity> query, Guid[] objectTypes, long pageIndex, int pageSize, out long totalRecords,
+            IQuery<IUmbracoEntity> filter, Ordering ordering, IQuery<IRelation> relationQuery = null);
 
         /// <summary>
         /// Gets paged entities for a query and a specific object type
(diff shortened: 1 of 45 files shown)


src/Umbraco.Core/Persistence/Repositories/IEntityRepository.cs
```diff
@@ -30,7 +42,37 @@ public interface IEntityRepository : IRepository
...
+        /// <param name="pageSize"></param>
+        /// <param name="totalRecords"></param>
+        /// <param name="filter"></param>
+        /// <param name="ordering"></param>
+        /// <returns>
+        /// A collection of mixed entity types which would be of type <see cref="IEntitySlim"/>, <see cref="IDocumentEntitySlim"/>, <see cref="IMediaEntitySlim"/>,
+        /// <see cref="IMemberEntitySlim"/>
+        /// </returns>
+        IEnumerable<IEntitySlim> GetPagedResultsByQuery(
+            IQuery<IUmbracoEntity> query, Guid[] objectTypes, long pageIndex, int pageSize, out long totalRecords,
```
warrenbuckley: XML signature does not match, missing the array for `objectTypes`

src/Umbraco.Core/Persistence/Repositories/Implement/EntityRepository.cs
```diff
@@ -70,35 +92,13 @@ public EntityRepository(IScopeAccessor scopeAccessor)
...
+            var page = Database.Page<GenericContentEntityDto>(pageIndexToFetch, pageSize, sql);
+            dtos = page.Items;
+            totalRecords = page.TotalItems;
 
-            var entities = dtos.Select(x => BuildEntity(isContent, isMedia, isMember, x)).ToArray();
+            var entities = dtos.Select(BuildEntity).ToArray();
 
-            if (isContent)
-                BuildVariants(entities.Cast<DocumentEntitySlim>());
+            BuildVariants(entities.OfType<DocumentEntitySlim>());
```
warrenbuckley: Learnt something new `.OfType<T>` as opposed to Cast 👍

PR #7068, opened 2019-11-06
Adds ability to bulk insert relations and adds tests

Shazwazza:
Adds ability to bulk insert relations and adds tests. This is so we don't end up with an N+1 issue when saving content, media or members since we first clear auto-found relations and then re-add them. Before this, we would add a relation one at a time which is a perf problem, now we bulk insert them.

linked to [AB#3448](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3448)

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
@@ -853,21 +853,22 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
             var allRelationTypes = RelationTypeRepository.GetMany(Array.Empty<int>())
                 .ToDictionary(x => x.Alias, x => x);
 
-            foreach(var rel in trackedRelations)
-            {
-                if (!allRelationTypes.TryGetValue(rel.RelationTypeAlias, out var relationType))
-                    throw new InvalidOperationException($"The relation type {rel.RelationTypeAlias} does not exist");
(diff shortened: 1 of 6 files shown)


src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
```diff
@@ -853,21 +853,22 @@ protected void PersistRelations(TEntity entity)
...
-                    continue; // This shouldn't happen!
+                    if (!keyToIds.TryGetValue(guid, out var id))
+                        return null; // This shouldn't happen!
 
-                //Create new relation
-                //TODO: This is N+1, we could do this all in one operation, just need a new method on the relations repo
-                RelationRepository.Save(new Relation(entity.Id, id, relationType));
-            }
+                    return new Relation(entity.Id, id, relationType);
+                }).WhereNotNull();
```
bergmania: In this case everything is correct, but maybe it is time to update the `WhereNotNull` extension method to test using `is null` instead of `!= null`. 
According to Mads Torgersen, it is much faster and more secure as it does not have to test for and possibly execute the overloads of the `!=` Operator

Shazwazza: ooh very nice, didn't know about that new feature, excellent!

bergmania:
Tested, works as expected

Issue #6324, opened 2019-09-11
V8: Broken link in RTE media picker for media files hosted on S3


bengyap:
We have an issue with the media files links in the RTE media picker. In details, we have installed the umbraco S3 plugin which uploads media files to the S3 bucket. In the result, the absolute urls of media files pointing to S3 are stored in our umbraco DB.

This worked fine in the 8.0.2 version. However since we upgraded to latest version 8.1.4, we noticed that media files links are broken in the RTE media picker because the links have become relative path instead of full url in 8.0.2 version. The relative path will not work for our case since our media files are hosted on S3.

When we select and insert a media file into RTE, we see broken link in the RTE as well as it inserts the relative path of the media file.

Also, fyi, the S3 plugin doesn't do anything when it comes to retrieving or displaying the media files, as the absolute urls (pointing to S3) are already stored in database. So I guess it's umbraco to handle the absolute urls of media files stored in the DB

**Step to Reproduce**

For Umbraco version since 8.1.0 with the Umbraco S3 plugin installed, login to the CMS backend.

1. Edit a page with a rich text editor, 
2. Go to the RTE, click media picker to insert an image

### Expected result

All media files stored in DB with absolute urls (pointing the S3), should appear in the RTE media picker sidebar, as well as after they are inserted to RTE.

### Actual result

In the sidebar displaying image list, images don't appear as they are broken links with relative path pointing to local site instead of S3. Clicking on the image, we can see relative path of the image in the "url" field. Inserting the image will result in relative path of image being inserted to the RTE, hence showing broken link in the RTE as well.  However after I save or publish the page, the inserted image appear in RTE with absolute url pointing to S3.

nul800sebastiaan:
Is the conclusion here that, with the S3 plugin installed:

1. Publishing a page in 8.1 with a broken image link fixes the image link? Or do you HAVE to replace the image first?
2. In 8.1 when you upload a new image, everything works?

rasmusjp:
I've located the problem, this method:

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Core/Persistence/Factories/ContentBaseFactory.cs#L314-L327 

which uses this regex https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Core/Persistence/Factories/ContentBaseFactory.cs#L13

[...]

PR #7075, opened 2019-11-06
Wrong media path stored in database

rasmusjp:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6324

I've consollidated how media paths are resolved, instead it being done in multiple places it's now up to the Property Editor to handle this.

A new interface has been introduced `IDataEditorWithMediaPath` which has one new method `string GetMediaPath(object value)` that the needs to be implemented, both `FileUploadPropertyEditor` and `ImageCropperPropertyEditor` implements the interface.

---
_This item has been added to our backlog [AB#3971](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3971)_


src/Umbraco.Core/PropertyEditors/IDataEditorWithMediaPath.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Umbraco.Core.PropertyEditors
+{
+    public interface IDataEditorWithMediaPath : IDataEditor
```
Shazwazza: It's probably easier/nicer and more flexible that IDataEditorWithMediaPath doesn't inherit from IDataEditor and it's just a standalone interface.

rasmusjp: I've removed the inheritance

PR #7088, opened 2019-11-07
Allows members to be ordered by last edited field

AndyFelton:
It was raised in #6477 that the last edited field on the default members list view doesn't allow sorting.

This PR adds that ability.

Also fixes a minor issue that when adding the content type.


emmaburstow:
Andyyyyyyyy! Thanks for the work.

nul800sebastiaan:
Brilliant! Sorry @AndyFelton for the delay, we're all good to go for 7.15 with this one. Thank you very much! 👍 ⭐

PR #7089, opened 2019-11-07
Added the correct sources to the buildscripts. If you have internal p…

JeffreyPerplex:
If you have an internal nugetpackagesource specified (like I have) and NuGet cannot connect to that packagesource you'll get an error message that it cannot connect (for example; if you're outside of your company network).

Of course Umbraco does not need that source, but someway it still fails... Now I've added the correct sources to the different files.

One source is the Umbraco myget-url, and one source is a NuGet-url

### DIFF
--- a/build/build-bootstrap.ps1
+++ b/build/build-bootstrap.ps1
@@ -22,6 +22,8 @@
   # get NuGet
   $cache = 4
   $nuget = "$scriptTemp\nuget.exe"
+  # ensure the correct NuGet-source is used. This one is used by Umbraco
+  $nugetsourceUmbraco = "https://www.myget.org/F/umbracocore/api/v3/index.json"
   if (-not $local)
   {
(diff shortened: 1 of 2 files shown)


JeffreyPerplex:
Error

Here you can see the error by the way

JeffreyPerplex:
And if you agree with this; you should also merge this one in: https://github.com/umbraco/Umbraco-Build/pull/3. This is downloaded by the powershell-script and in a different repository (not branch :P)

JeffreyPerplex:
Solves this issue by the way: https://github.com/umbraco/Umbraco-CMS/issues/7090

emmaburstow:
Thanks Jeffrey!

nul800sebastiaan:
Sorry for the delay @JeffreyPerplex and thanks so much for making it easier to build umbraco with a different configuration, great stuff! 👍

PR #7091, opened 2019-11-07
- added delete title to trash icons in the grid

prjseal:
### UK Fest Hackathon PR

In the Grid Editor, when you hover over the trash icon there is no title which tells you what it is for, but there is one for the Confirm and for Cancel.

## Before

This PR adds the Delete title to the trash icon.


emmaburstow:
Paul, this is fab. Nice spot.

PR #7094, opened 2019-11-07
Accessibility Improvement - Open Template in info content app changed from anchor tag to button

poornimanayar:
Currently, in the Info content app on any node the "Open" template is an `<a>` . Changed it to a `<button>` so thats its more accessible.

umbraco-template-open


emmaburstow:
Lovely

PR #7098, opened 2019-11-07
V8: Make nested content item picker slide in from right

kjac:
### Prerequisites

### Description
 
_Slight disclaimer: This PR was done on a phone because Em dared me to. Blame Em. I know I do._

One thing that's annoyed me a lot about nested content is how the overlay for picking new items work.  

1. The centered dialog style overlay isn't really used anywhere else in the UI except the grid (also for picking new items), so it is defacto not a recognizable editing experience for the editors.
2. The first time you open the overlay, it is usually not at all centered close to the "add" botton, specially on smaller screens. The second time you click "add" it will center just fine, unless of course you reload the page.
3. This is the worst one: When you have more than a few item types (off the top of my head I think it's 15+), a search field is displayed. When you search for items, the dialog starts shrinking to fit the resulting items, which is confusing to look at and somewhat of an eyesore to the editors.

This PR makes the overlay slide in from the right side, just like content pickers and similar editors. This solves all of the pain points listed above.

You could argue that the existing overlay works just great if you have just a few item types and that we should spend the energy fixing the centering instead. But I really do feel that the overlay is out of place in the whole editing experience.

I really wish I could create some GIFs to explain myself better. Darn phone.

Cheers!


emmaburstow:
👍

kjac:
Updated with latest v8/dev

poornimanayar:
Adding a screenshot so that this helps

NC-Slide

@kjac , I agree with everything that you have put down in the description. I have tested this and this works well. I tested with around 25 element types, I am assuming that the dialog does not page?

Poornima

kjac:
@poornimanayar awesome 😄 

Nope the dialog doesn't page... nor does the current "dialog overlay" solution. It simply shows a scroll bar - I do believe the "slide-in overlay" is better with scroll bars than the "dialog overlay", though. Also the search bar is there to help the editors find the element type they need, so I guess paging would be a bit over the top to do.

poornimanayar:
I agree with what you said @kjac its a lot more consistent.  I wanted to clarify regd the paging:-)

nul800sebastiaan:
After a bit of discussion at HQ I'll close this for now. While this is a good solution for people with many choices, it's not great if there's only a few choices.
However, related to this it would be nice to have space to explain more what each of the options mean. This will be handled as part of the Block Editor work as well, so we'll revisit when a solution has been found there.

Too bad, always love merging PRs made on a dare! 😉

PR #7101, opened 2019-11-07
Corrected the model variable name

richarth:
The model variable name is registerModel rather than profileModel.

This change corrects the comment's example code.  There is nothing to test as this is just some commented out example code.


emmaburstow:
Thanks for that change @richarth - if you're here at the Hackathon in London, come see me or Seb for a sticker

richarth:
Thanks, but I'm not at the Hackathon. I'll consider this a virtual sticker.

PR #7102, opened 2019-11-07
V8: Accessibility Changes For umbEditorHeader Directive (edit user) 

RachBreeze:
This PR extends #6315 by adding the $setAccessibleHeader event to enable the umbHeader Directive to set the screen reader properties and setting the page title when a user edits an existing user.


emmaburstow:
Hey Rachel,

Thanks for the work here :) We'll shout if we need you.

Em

RachBreeze:
Hi
I've got a couple of changes I'd like to make based on this  PR and on #7104  it would be lovely to get some feedback please
Thank you in advance

RachBreeze:
Hi,  I'd like to look to finish of the title work I've been doing and would be good to get feedback on this PR please, there's a few places in the CMS where as a developer we can't just set the title and screen reader properties properties containing the name are loaded after the header is loaded, and ng-hide won't resolve this issue.  $`setAccessibleHeader` attempts to resolve this by allowing them to be set after the header directive is loaded.

RachBreeze:
Hiya 
Have you managed to review this yet please?  Is there anything I can do to assist?  As mentioned above I've got a few more tbd issues which would be dependent on this PR so it would be great to get feedback.

nul800sebastiaan:
Hi @RachBreeze and sorry for the long delay! I've merged this with the latest `v8/contrib` so it's up to date. I'm a bit confused about this change here to the loader that seems unrelated, why is that one changed please? 

https://github.com/umbraco/Umbraco-CMS/pull/7102/files#diff-cc14314b8965974db240d069e00e28d0R2-R3

nul800sebastiaan:
Turns out that change came from v7 like with some of your other PRs, so I've reverted that for now.

Looks good otherwise! 👍

PR #7103, opened 2019-11-07
Added a required property to properties

richarth:
I've added a required property to UmbracoProperty and set this for the member property view models.  We can then use this in views such as the registration macro partial view to determine whether a property/field is required.


emmaburstow:
Hey @richarth,

Thanks for the work - we'll take a look and let you know if we need anything.

Emma

kjac:
I may be wrong here, but I do believe this constitutes a breaking change. For good measure I have tagged this as breaking; feel free to prove me wrong 😄

PR #7106, opened 2019-11-07
Don't keep prepending login to the page title

RachBreeze:
Now uses https://github.com/umbraco/Umbraco-CMS/pull/6456  to prevent the page title being appended to the previous page title when the user is automatically logged out.


emmaburstow:
Thanks Rach. You're a star ⭐️ 

Em

RachBreeze:
Hi @kjac 
Thank you very much for reviewing

For testing purposes I've set the timeout to one minute:
**Prior to this PR**
Go to a content page:
image

The logout displays:
image

**This PR**
Go to a content page:
image

image

There were two reasons for submitting the PR, the first was as discussed above, to prevent "Login"being prepended to the existing page name, the second was to use the `$scope.$emit("$changeTitle", title);` logic used elsewhere in the site.

Reviewing the 2 screen shots of the login page, and listening to the title on a screen reader both before and after this change, I think the Login - x-  site needs changing to Login - site (where x is the tree item to be logged into e.g. content and site is the site being displayed eg localhost)

Fixing the display of "login" in the title after logging in is something I hadn't spotted and I'd need to investigate further.

kjac:
Merged. Thanks again @RachBreeze 😄

RachBreeze:
Hi @kjac 
Thanks for reviewing and Happy New Year

PR #7107, opened 2019-11-07
- only show remove photo on user avatar when there is an avatar

prjseal:
### Umbraco UK Fest Hackathon PR

On the user profile screen, under the user avatar preview, the remove photo link should only show if there is an avatar already picked. This PR fixes that.


emmaburstow:
Thanks for the work, Paul. We'll let you know if we need anything done and if we're happy, as usual, we'll merge.

Em

prjseal:
Cool, thanks

kjac:
Merged 👍 thank you again, @prjseal

PR #7111, opened 2019-11-08
NetCore: Static IOHelper -> IIOHelper

bergmania:
Refactored static `IOHelper` into interface and class. 

Added IOHelper on current to make it easy for now


Shazwazza:
Straight forward, will merge. I created a sub task to change usages of IIOHelper to be injected AB#3667

PR #7123, opened 2019-11-11
V8: Use current client culture when searching content pickers

kjac:
### Prerequisites

### Description

If you choose a non-default culture as your main culture or edit content in another culture than the default one, the selected culture isn't reflected when using the search option of content pickers (including MNTP). But once you've picked content from the search results, the list of picked content is displayed in the correct culture - including the URL to the content:

picker-search-culture-before

This PR ensures that we utilize the client culture when performing content searches in the entity API. It looks like this:

picker-search-culture-after


kjac:
PR updated with latest v8/dev

emmaburstow:
Eyyy, thanks for the work Kenn. We'll shout if we need anything.

Em

poornimanayar:
@kjac , Just to clarify this is only to do with searching. If i have different node names for two languages they wont reflect in the tree as such, it only shows the url in current culture. also this is only for ContentPicker?

For eg I have About Us (en-US) and About Us UK (en-GB) as the node names. With this fix in place the search should show the correct url, but node names will be shown in default language.

kjac:
Hi @poornimanayar 

This PR ensures that the displayed node names match the current culture when searching in the content picker (and MNTP - I believe they share the same angular tree component). 

The GIFs attempt to illustrate this: Soon as I start searching in Danish, the search results are shown in English (the default culture). The results are correct (they match the Danish search phrase), they are just displayed incorrectly.

poornimanayar:
And merged :-) Thanks @kjac

PR #7130, opened 2019-11-12
V8: Support culture variant URLs in multi URL picker

### LINKED ISSUE #7114: Multi URL Picker doesn't display culture variant URL in backoffice
When picking a content URL in the Multi URL Picker in a multilinguagl website the default language URL is displayed in the backoffice instead of the culture variant URL:

image

## Umbraco version: 

I am seeing this issue on Umbraco version: 8.2.2

Reproduction
------------

Add two (or more) languages to the website with one default:

image

Add a document type with **Allow varying by culture** enabled and add a Multi URL Picker property to this document also with **Allow varying by culture** enabled:

image

Add internal content links to the page in the non-default language:

image

The property shows the URL's for the default language instead of the culture variant URL, this is very confusing for the editors. When rendering the links in a view (frontend) it does show the proper culture variant URL, so it's only an issue of the backoffice.

I've found similar issues in Github. In issue [#6145](https://github.com/umbraco/Umbraco-CMS/issues/6145) @nul800sebastiaan states this issue was fixed in Umbraco 8.1.4, but I'm still seeing it in version 8.2.2. I've also tried it on a clean install which you can download here (backoffice credentials are in the readme.txt):

https://www.dropbox.com/s/h5hhsztgxv9bvpm/umbraco-822.zip?dl=0

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7114

### Description

#7114 describes how the multi URL picker doesn't display URLs in the current editor culture, if you're editing content in another culture than the default one. The problem is actually two-fold:

1. When you add a new item to the multi URL picker, the item URL is using the default culture.
2. When the multi URL picker loads and contains existing items, these item URLs are also shown using the default culture.

Here's a GIF of me editing in the Danish culture (where English is default). The Danish culture is configured with the host name "/da" and the English with "/en":

multi-url-cultures-before

This PR addresses both issues by applying the same pattern for URL loading that is used in the content pickers. With this PR, the multi URL picker displays the URLs with respect to the current editor culture:

multi-url-cultures-after

See also #7123 which addresses the content picker search always using the default culture (it's related to this since you can search the content tree in the multi URL picker).


emmaburstow:
Hey Kenn,

Lots of work here and lots of documentation. I am truly sorry it's taken so long to say so but thank you. We are very grateful for all that you do 🤗 

Emma

ronaldbarendse:
@kjac Looks like this breaks external links. as the `item.udi` is not set: `Cannot read property 'indexOf' of null`.
https://github.com/umbraco/Umbraco-CMS/blob/4f7b62f3692aa123bda3238f3f5dced874db7e73/src/Umbraco.Web.UI.Client/src/views/propertyeditors/multiurlpicker/multiurlpicker.controller.js#L150

ronaldbarendse:
The real fix is by passing the culture to the `Url()` methods in the `MultiUrlPickerValueEditor`, as that already fetches the current URL of the saved UDI:

https://github.com/umbraco/Umbraco-CMS/blob/a9d45413c89d67280ce54c7fd4182e90323bab6c/src/Umbraco.Web/PropertyEditors/MultiUrlPickerValueEditor.cs#L84

https://github.com/umbraco/Umbraco-CMS/blob/a9d45413c89d67280ce54c7fd4182e90323bab6c/src/Umbraco.Web/PropertyEditors/MultiUrlPickerValueEditor.cs#L92

PR #7135, opened 2019-11-12
Changed SystemDirectories from static to instance,

bergmania:
- Changed `SystemDirectories` from static to instance
- Added interface `ISystemDirectories` in abstractions 
- Created a static property on current.

This is required for config movement.

I create a up for graps task about injecting where possible


src/Umbraco.Core/IO/SystemDirectories.cs
```diff
@@ -4,50 +4,50 @@
...
+        public string MvcViews => "~/Views";
 
-        public static string PartialViews => MvcViews + "/Partials/";
+        public string PartialViews => MvcViews + "/Partials/";
 
-        public static string MacroPartials => MvcViews + "/MacroPartials/";
+        public string MacroPartials => MvcViews + "/MacroPartials/";
 
-        public static string Media => Current.IOHelper.ReturnPath("umbracoMediaPath", "~/media");
+        public string Media => Current.IOHelper.ReturnPath("umbracoMediaPath", "~/media");
```
Shazwazza: This is going to get ugly. If IOHelper requires ISystemDirectories then ISystemDirectories is going to require `IIOHelper`. We'll need to sort this out before we can continue and mark anything as up for grabs.

src/Umbraco.Core/Configuration/GlobalSettings.cs
```diff
@@ -378,5 +378,43 @@ public bool UseHttps
...
+        {
+            if (backingField != null) return backingField;
+
+            if (ConfigurationManager.AppSettings.ContainsKey(appSettingKey))
+            {
+                try
+                {
+                    var value = ConfigurationManager.AppSettings[appSettingKey];
+
+                    backingField = (T)Convert.ChangeType(value, typeof(T));
```
Shazwazza: Might be better to use our `TryConvertTo` ext method which is a lot smarter than the ChangeType conversion, plus then you don't need to try/catch

Issue #7126, opened 2019-11-11
Image properties window not opening for Images in RTE of a Grid


russellshome:
When trying to view the properties of an image in the RTE of a Grid the properties window does not load due to a javascript error:

Possibly unhandled rejection: The object in the array was not formatted as a key/value pair

## Umbraco version

I am seeing this issue on Umbraco version: 8.2.1 and 8.2.2

Reproduction
------------

Add Starter site to new install of Umbraco

Under settings - data types - Content Base - Content - Grid layout
Click Image (button to allow adding images to the RTE)
Add an image to the RTE on the about us page
LOG OUT (This is important as the issue will not otherwise happen)
Log In
Select the image that was added earlier and click the image button

===

Note that the error persists until the image button is clicked without selecting an image after which time it will not occur again until after Log out/Log in.

### Bug summary

the properties window does not load due to a javascript error:

### Specifics

Borwsers: Chrome and Edge tested

videoi is attached
[2019-11-12 07-44-05.zip](https://github.com/umbraco/Umbraco-CMS/files/3832788/2019-11-12.07-44-05.zip)

### Steps to reproduce

see above

### Expected result

Expected no error and property window to display at minimum

### Actual result

See video

---
_This item has been added to our backlog [AB#5387](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/5387)_

nul800sebastiaan:
That's annoying @russellshome! I can reproduce in 8.1.4 as well, so this problem might have been around for a while. Very curious that this only happens after a logout.

If you have any clue as to what is going wrong, maybe you can find a fix for it as well?

umbrabot:
Hi @russellshome,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

kjac:
This issue is not limited to grid RTE's; I can reproduce in a regular RTE. I'm trying to trace the source of the issue 👍

kjac:
Right! PR in #7140

PR #7137, opened 2019-11-12
V8: Fix mapping YSOD for culture variant content after logout and login

kjac:
### Prerequisites

### Description

I came across this while investigating #7126: After logout and a subsequent login, `mculture` is never re-initialized. This causes list views for culture variant content to fail miserably:

culture-variant-content-logout-login-before

This PR ensures that the navigation controller initialization is always performed after login, thus `mculture` is always set - and thus the problem goes away:

culture-variant-content-logout-login-after

#### Testing this PR

1. Enable list view on a culture variant content type.
2. Ensure one item of this content type has one or more children.
3. Logout.
4. Login.
5. Open the content item and verify that no mapping error occurs.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/navigation.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/navigation.controller.js
@@ -249,6 +249,8 @@ function NavigationController($scope, $rootScope, $location, $log, $q, $routePar
 
     //when a user logs out or timesout
     evts.push(eventsService.on("app.notAuthenticated", function () {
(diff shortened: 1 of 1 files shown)


emmaburstow:
Ahhh nice fix! Thanks for this Kenn. "Fail miserably" indeed,

Em

poornimanayar:
Hi @kjac ,

I cannot reproduce this on the latest v8/contrib. I can see the culture getting initialised upon login. 

Poornima

kjac:
@poornimanayar you're right, I guess the issue has been resolved by another PR 👍

PR #7139, opened 2019-11-12
AB3621 - Paging Relations UI

warrenbuckley:
This adds pagination to the Relation Type on the 'Relations' content app with a new API controller, Angular JS resource & other UI/JS changes as needed.

Fixes [AB#3621](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3621)

## Testing
* Verify code implemented correctly
* Test you can page correctly & the correct data is returned

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/RelationRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/RelationRepository.cs
@@ -253,7 +253,7 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
             // apply ordering
             ApplyOrdering(ref sql, ordering);
 
-            var pageIndexToFetch = pageIndex;
+            var pageIndexToFetch = pageIndex + 1;
             var page = Database.Page<RelationDto>(pageIndexToFetch, pageSize, sql);
             var dtos = page.Items;
             totalRecords = page.TotalItems;
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/RelationRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/RelationRepository.cs
@@ -253,7 +253,7 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
             // apply ordering
             ApplyOrdering(ref sql, ordering);
 
-            var pageIndexToFetch = pageIndex + 1;
+            var pageIndexToFetch = pageIndex;
             var page = Database.Page<RelationDto>(pageIndexToFetch, pageSize, sql);
             var dtos = page.Items;
             totalRecords = page.TotalItems;
--- a/src/Umbraco.Web.UI.Client/src/common/resources/relationtype.resource.js
+++ b/src/Umbraco.Web.UI.Client/src/common/resources/relationtype.resource.js
@@ -118,13 +118,9 @@ function relationTypeResource($q, $http, umbRequestHelper, umbDataFormatter) {
 
         getPagedResults: function (id, options) {
 
-            console.log('options in', options);
-
             var defaults = {
-                pageSize: 100,
-                pageNumber: 1,
-                orderDirection: "Ascending",
-                orderBy: "SortOrder"
+                pageSize: 25,
+                pageNumber: 1
             };
             if (options === undefined) {
                 options = {};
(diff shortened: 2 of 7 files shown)


warrenbuckley:
@Shazwazza any chance you have got time to review & test this please mate, then I can do the other pagination tasks if this implementation/approach is OK

src/Umbraco.Core/Persistence/Repositories/Implement/RelationRepository.cs
```diff
@@ -253,7 +253,7 @@ public IEnumerable<IRelation> GetPagedRelationsByQuery(IQuery<IRelation> query,
             // apply ordering
             ApplyOrdering(ref sql, ordering);
 
-            var pageIndexToFetch = pageIndex + 1;
+            var pageIndexToFetch = pageIndex;
```
Shazwazza: This shouldn't be changed. The var name `pageIndexToFetch` could prob be changed to `pageNumberToFetch`. When you see a term called "Index" for paging in our codebase it means it's a zero based index, when you see "Number" it means it's a 1 based number. So you should be passing in the zero based index to this method just like all of our other paging methods.

src/Umbraco.Web.UI.Client/src/common/resources/relationtype.resource.js
```diff
@@ -114,6 +114,49 @@ function relationTypeResource($q, $http, umbRequestHelper, umbDataFormatter) {
                 $http.post(umbRequestHelper.getApiUrl("relationTypeApiBaseUrl", "DeleteById", [{ id: id }])),
                 "Failed to delete item " + id
             );
+        },
+
+        getPagedResults: function (id, options) {
+
+            console.log('options in', options);
```
Shazwazza: remove console.log

src/Umbraco.Web.UI.Client/src/common/resources/relationtype.resource.js
```diff
@@ -114,6 +114,49 @@ function relationTypeResource($q, $http, umbRequestHelper, umbDataFormatter) {
...
+            options = defaults;
+            //change asc/desct
+            if (options.orderDirection === "asc") {
+                options.orderDirection = "Ascending";
+            }
+            else if (options.orderDirection === "desc") {
+                options.orderDirection = "Descending";
+            }
+
+            console.log('options before HTTP', options);
```
Shazwazza: no console.log

src/Umbraco.Web/Editors/RelationTypeController.cs
```diff
@@ -50,6 +52,23 @@ public RelationTypeDisplay GetById(int id)
...
+        public PagedResult<RelationDisplay> GetPagedResults(int id, int pageNumber = 1, int pageSize = 100)
+        {
+
+            if (pageNumber <= 0 || pageSize <= 0)
+            {
+                throw new NotSupportedException("Both pageNumber and pageSize must be greater than zero");
+            }
+
+            // Ordering do we need to pass through?
+            var relations = Services.RelationService.GetPagedByRelationTypeId(id, pageNumber, pageSize, out long totalRecords);
```
Shazwazza: you need to pass a pageIndex to this service not a 1 based number (see prev comments)

Shazwazza: I wouldn't worry about ordering for now, people can't sort the results currently.

src/Umbraco.Web.UI.Client/src/views/relationtypes/edit.controller.js
```diff
@@ -21,6 +21,11 @@ function RelationTypeEditController($scope, $routeParams, relationTypeResource,
 
     function init() {
         vm.page.loading = true;
+        vm.relationsLoading = true;
+
+        vm.changePageNumber = changePageNumber;
+        vm.options = {};
+        vm.options.pageSize = 3;
```
Shazwazza: This seems to be a testing page size, prob should just be the default :)

warrenbuckley: Yes a page size of 100 seems excessive and I wanted to verify paging was kicking, hence it was set to three for testing.

warrenbuckley:
Should be ready for re-review @Shazwazza 
One thing to note is that the main content/app tab for RelationType listed the count of relations (was done by a simple count `.length` of the array of relations coming back from the API which we have now removed

PR #7141, opened 2019-11-12
WIP: Add missing database indexes for foreign keys

### LINKED ISSUE #6601: Database indexes missing on foreign keys
## Description

_**UPDATE: I've added notes about specific indexes that need to be created [below](https://github.com/umbraco/Umbraco-CMS/issues/6601#issuecomment-573564629)**_

SQL server does not enforce that indexes are created on FK columns but in many (most) cases they should have indexes since quite often is the case that these columns are queried against.

One important one that is missing is: 

```sql
CREATE NONCLUSTERED INDEX TEST_IX_umbracoContentVersion_nodeId ON dbo.umbracoContentVersion (nodeId)
```

We query against this column a **lot**!

So we definitely need that Index created but there's probably lots of others...

I found an SQL script that lists all FK columns that are missing indexes from this site: https://www.mssqltips.com/sqlservertip/5004/script-to-identify-all-nonindexed-foreign-keys-in-a-sql-server-database/

For reference, here's an article describing the benefits of indexing FK columns: https://sqlperformance.com/2012/11/t-sql-queries/benefits-indexing-foreign-keys

Here are the results of that script:

| Table_Name|Column_Name|
| ------------- | -----|
|cmsContentType2ContentType	|childContentTypeId|
|cmsContentTypeAllowedContentType	|AllowedId|
|cmsDictionary	|parent|
|cmsDocumentType	|templateNodeId|
|cmsLanguageText	|languageId|
|cmsLanguageText	|UniqueId|
|cmsMember2MemberGroup	|MemberGroup|
|cmsMemberType |	NodeId|
|cmsPropertyType|	contentTypeId|
|cmsPropertyType|	dataTypeId|
|cmsPropertyType|	propertyTypeGroupId|
|cmsPropertyTypeGroup	|contenttypeNodeId|
|cmsTagRelationship	|propertyTypeId|
|cmsTagRelationship	|tagId|
|cmsTask	|nodeId|
|cmsTask	|parentUserId|
|cmsTask	|taskTypeId|
|cmsTask	|userId|
|umbracoAccess	|loginNodeId|
|umbracoAccess	|noAccessNodeId|
|umbracoAccessRule|	accessId|
|umbracoContent	|contentTypeId|
|umbracoContentSchedule	|languageId|
|umbracoContentSchedule	|nodeId|
|~~umbracoContentVersion~~	|nodeId|
|~~umbracoContentVersion~~	|userId|
|umbracoContentVersionCultureVariation	|availableUserId|
|umbracoDocumentVersion	|templateId|
|umbracoDomain	|domainRootStructureID|
|umbracoLog	|userId|
|umbracoNode|	nodeUser|
|umbracoRedirectUrl	|contentKey|
|umbracoRelation	|childId|
|umbracoRelation	|relType|
|umbracoUser2NodeNotify	|nodeId|
|umbracoUser2UserGroup	|userGroupId|
|umbracoUserGroup	|startContentId|
|umbracoUserGroup	|startMediaId|
|umbracoUserLogin	|userId|
|umbracoUserStartNode	|startNode|
|umbracoUserStartNode	|userId|

## What to fix

We definitely need the umbracoContentVersion.nodeId indexed but we should review all of the above and determine if adding indexes to any of these will be beneficial. We only really want to add indexes if they are used in queries in our codebase. Adding too many indexes will slow down data insertion which we don't want. 

In most cases you can find the "DTO" object for a table such as: `Umbraco.Core.Persistence.Dtos.ContentVersionDto` and then you can righ click on it's column that you want to search for to see if it's used in queries. For example, for the ContentVersionDto.NodeId, i right click this member and say find references in VS, then i can see how many queries it's used in

## How to fix

For indexes that should exist, the DTO classes will need to be updated to declare the index and then a migration will need to be written to add all of the new indexes.

---
_This item has been added to our backlog [AB#4131](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4131)_

rbottema:
### Prerequisites

If there's an existing issue for this PR then this fixes #6601 

### Description

In this PR I have implemented the missing indexes as reported by @Shazwazza, using the very helpful list compiled by @liamlaverty. I have only implemented indexes for the columns that @liamlaverty noted as definitely needing an index. I also took the liberty of including new indexes of which there was a `TODO` comment, and including the `TableName` field when not present (so it can be used in index names and the reflection in index creation). This should improve performance of various queries around the application.

#### Steps to test:

- Perform migration
- Verify if the indexes were created succesfully

#### TODO

This is still a WIP PR because of the following reasons:

- Someone (from HQ?) should verify @liamlaverty's table in #6601, also checking if some of the 'maybe's actually do need an index as well.
- I have not tested the functionality myself, as I do not know how to test a migration.. Any help guiding me in this is much appreciated!
- Is my approach in creating the indexes this way correct, or should I just use the `CreateKeysAndIndexes` migration that is already there? I opted for this approach for safety, as I don't know exactly how rigorous the `CreateKeysAndIndexes` migration is in deletion of already existing (possibly added by the user) indexes.
- Related, what should be done if a user already created an index we are defining here? In particular, on the same column, but with a different name.

### DIFF
--- a/src/Umbraco.Core/Migrations/Upgrade/V_8_4_0/AddForeignKeyIndexes.cs
+++ b/src/Umbraco.Core/Migrations/Upgrade/V_8_4_0/AddForeignKeyIndexes.cs
@@ -1,0 +1,75 @@
+﻿using NPoco;
+using System;
+using System.Linq;
+using Umbraco.Core.Persistence.DatabaseAnnotations;
+using Umbraco.Core.Persistence.Dtos;
+
+namespace Umbraco.Core.Migrations.Upgrade.V_8_4_0
+{
+    public class AddForeignKeyIndexes : MigrationBase
+    {
+        public AddForeignKeyIndexes(IMigrationContext context)
+            : base(context)
+        { }
+
+        public override void Migrate()
+        {
+            var newIndexes = new[]
+            {
+                (typeof(DictionaryDto), nameof(DictionaryDto.Parent)),
+                (typeof(LanguageTextDto), nameof(LanguageTextDto.LanguageId)),
+                (typeof(LanguageTextDto), nameof(LanguageTextDto.UniqueId)),
+                (typeof(Member2MemberGroupDto), nameof(Member2MemberGroupDto.MemberGroup)),
+                (typeof(PropertyTypeDto), nameof(PropertyTypeDto.DataTypeId)),
+                (typeof(PropertyTypeDto), nameof(PropertyTypeDto.ContentTypeId)),
+                (typeof(PropertyTypeDto), nameof(PropertyTypeDto.PropertyTypeGroupId)),
+                (typeof(PropertyTypeGroupDto), nameof(PropertyTypeGroupDto.ContentTypeNodeId)),
+                (typeof(TagRelationshipDto), nameof(TagRelationshipDto.TagId)),
+                (typeof(TagRelationshipDto), nameof(TagRelationshipDto.PropertyTypeId)),
+                (typeof(AccessRuleDto), nameof(AccessRuleDto.AccessId)),
+                (typeof(ContentDto), nameof(ContentDto.ContentTypeId)),
+                (typeof(ContentScheduleDto), nameof(ContentScheduleDto.NodeId)),
+                (typeof(ContentVersionDto), nameof(ContentVersionDto.NodeId)),
+                (typeof(ContentVersionDto), nameof(ContentVersionDto.UserId)),
+                (typeof(ContentVersionDto), nameof(ContentVersionDto.Current)),
+                (typeof(LogDto), nameof(LogDto.UserId)),
+                (typeof(LogDto), nameof(LogDto.Datestamp)),
+                (typeof(LogDto), nameof(LogDto.Header)),
+                (typeof(NodeDto), nameof(NodeDto.UserId)),
+                (typeof(RedirectUrlDto), nameof(RedirectUrlDto.ContentKey)),
+                (typeof(User2NodeNotifyDto), nameof(User2NodeNotifyDto.NodeId)),
+                (typeof(User2UserGroupDto), nameof(User2UserGroupDto.UserGroupId)),
+                (typeof(UserGroupDto), nameof(UserGroupDto.StartContentId)),
+                (typeof(UserGroupDto), nameof(UserGroupDto.StartMediaId)),
+                (typeof(UserLoginDto), nameof(UserLoginDto.UserId)),
+                (typeof(UserStartNodeDto), nameof(UserStartNodeDto.UserId)),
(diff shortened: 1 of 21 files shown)


src/Umbraco.Core/Migrations/Upgrade/V_8_4_0/AddForeignKeyIndexes.cs
```diff
@@ -0,0 +1,75 @@
...
+            foreach (var (type, propertyName) in newIndexes)
+            {
+                CreateIndexIfNotExists(type, propertyName);
+            }
+        }
+
+        private void CreateIndexIfNotExists(Type dto, string propertyName)
+        {
+            var property = dto.GetProperty(propertyName);
+            var indexName = property.GetCustomAttributes(false).OfType<IndexAttribute>().Single().Name;
```
ronaldbarendse: Make sure this also supports zero attributes, as that might change in the future. This migration will still run when upgrading the current to a higher version that removed the `IndexAttribute` 😉

src/Umbraco.Core/Migrations/Upgrade/V_8_4_0/AddForeignKeyIndexes.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            if (IndexExists(indexName))
+                return;
+
+            var tableName = dto.GetField("TableName").GetValue(null) as string;
```
ronaldbarendse: Shouldn't this use the `TableNameAttribute` instead?

rbottema:
Thanks for your comments @ronaldbarendse! I will wait until I get more feedback (especially about the testing and about `CreateKeysAndIndexes`) before I start processing your feedback.

emmaburstow:
Hey @rbottema - I can see this is a work in progress and that there's some discussion going on so I will say thanks and then wait for you to be ready  for us to review it before we do.

Thanks!

Emma

ronaldbarendse:
@emmaburstow This PR actually requires some input from HQ, probably from @Shazwazza or @liamlaverty...

@rbottema To make sure all indexes are added according to the attributes (same as when creating the tables in the [DatabaseInstallStep](https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web/Install/InstallSteps/DatabaseInstallStep.cs)), you can do the following:

```c#
foreach (var x in DatabaseSchemaCreator.OrderedTables)
{
   Delete.KeysAndIndexes(x).Do();
   Create.KeysAndIndexes(x).Do();
}
```

Deleting and re-creating the indexes might be unnecessary (and negatively impact database performance) and indexes that are renamed/removed from the DTOs will not be deleted though. Maybe the `CreateKeysAndIndexesBuilder` should be extended to allow conditionally creating the indexes?

PR #7146, opened 2019-11-13
Allow replacing UmbracoPlan through DI

stevemegson:
Allows the standard database migration plan to be modified from outside Umbraco.Core, adding custom steps. Modifies `DatabaseUpgradeStep` to request an implementation of `UmbracoPlan` via DI, allowing another assembly to replace it with `composition.RegisterUnique<UmbracoPlan, ModifiedPlan>()` in a composer. 

UmbracoPlan is also used in other places, but only to read the FinalState. We check that the replacement plan has the same FinalState as the standard plan, so I've left these using the standard plan. Some would be inconvenient to change.

Also makes `NPocoDatabaseTypeExtensions` and `DefaultPreValueMigrator` public so that they can be used in custom migrations. These aren't essential, but I don't think there's any need for them to be internal.


emmaburstow:
Hey @stevemegson - thanks for the work here. Looks really interesting. We'll take a look and let you know if we have any questions.

Emma

PR #7150, opened 2019-11-13
v7: Fix false matches for AltTemplate convention Urls

marcemarc:
See https://our.umbraco.com/forum/using-umbraco-and-getting-started/98220-help-with-error-message

It's possible to load a published node via an alternative template by requesting it in the format:

/about-us/our-staff/staffgridlayout

where '/about-us/our-staff/' is the node in Umbraco and 'staffgridlayout' is the alias of the alternative template you want Umbraco to load the content with.

This is implemented by the ContentFinderByNiceUrlAndTemplate Content Finder...

... this is kind of a hidden feature of Umbraco that's been there for years...

Anyway the way this is implemented assumes that if the 'last portion of a url' matches a template, then the request must be a valid request trying to make use of this convention so

/orangey/crows/staffgridlayout

would cause the Content Finder to try and load the '/orangey/crows/' node with the 'staffgridlayout' template...  but of the /orangey/crows/ node doesn't exist...

... there is no null check, and so the node.IsAllowedTemplate(template.Id) triggers an error, and the request doesn't fall through to a proper 404 page, (perhaps set later in the ContentFinder stack...)

This commit, adds the null check in.

Don't think this is an issue on V8, where the convention has also been removed by default.

### Prerequisites

- [|] I have added steps to test this contribution in the description below

### Testering

Install the starter kit
Create a new additional template for a BlogPostPage, call it 'SuperBlogPost'
Verify you can visit the blog post page url , with /SuperBlogPost appended to load the blog post in the alternative template.
Now visit another made up url with /SuperBlogPost appended eg
/orangey/crows/SuperBlogPost/
The page should 404 instead of YSOD
You should get an entry in the debug log (if logs are set to debug), explaining what just happened.

### DIFF
--- a/src/Umbraco.Web/Routing/ContentFinderByNiceUrlAndTemplate.cs
+++ b/src/Umbraco.Web/Routing/ContentFinderByNiceUrlAndTemplate.cs
@@ -43,16 +43,23 @@ namespace Umbraco.Web.Routing
 
                     var route = docRequest.HasDomain ? (docRequest.Domain.RootNodeId.ToString() + path) : path;
                     node = FindContent(docRequest, route);
-
-                    if (node.IsAllowedTemplate(template.Id))
-                    {
-                        docRequest.TemplateModel = template;
+                    //not guaranteed to find a node - just because last portion of url contains a template alias, doesn't mean remaining part of the url is a published node
+                    if (node !=null){
+                        if (node.IsAllowedTemplate(template.Id))
+                        {
+                            docRequest.TemplateModel = template;
+                        }
+                        else
+                        {
+                            LogHelper.Warn<ContentFinderByNiceUrlAndTemplate>("Configuration settings prevent template \"{0}\" from showing for node \"{1}\"", () => templateAlias, () => node.Id);
+                            docRequest.PublishedContent = null;
+                            node = null;
+                        }
                     }
-                    else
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hullo Marc,

Thanks for the PR. Thanks for the explanation of what you've done here too. Gives me plenty to work with. I'll take a look and let you know if I have any questions. Looks legit to me so far,

Em

emmaburstow:
This works nicely - I'm getting the 404 when I put in a made up URL but I'm also getting the alt template when I add /superblogpost/ to any other url in the site. Is that expected behaviour?

marcemarc:
@emmaburstow Yes, that's how the AltTemplate convention works out of the box (there is an option to turn it off in the umbracoSettings.config routing section... ie it hasn't ever, as far as I know, carried out a check to see if the template is 'allowed' for the particular node. This PR just fixes people getting tons of errors in their logs instead of a 404 if someone spamming the site happens to use a url that matches an existing template...

emmaburstow:
Brilliant. Thanks for that Marc. I'll approve :D
Happy new year!

nul800sebastiaan:
Sorry for the wait @marcemarc - no news as to when 7.15.4 will be out, but this PR is good to go for when that happens! 👍

PR #7155, opened 2019-11-14
set a minimum version requirement for node

nielslyngsoe:
To avoid people using old version of NodeJS, I would like to set a minimum version.


warrenbuckley:
Yep all fine :)

PR #7160, opened 2019-11-15
AB3610 - Pagination of Media References

warrenbuckley:
This adds in pagination to media references (each group/entity type has its own pager)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/resources/media.resource.js
+++ b/src/Umbraco.Web.UI.Client/src/common/resources/media.resource.js
@@ -554,28 +554,88 @@ function mediaResource($q, $http, umbDataFormatter, umbRequestHelper) {
                 'Failed to retrieve media items for search: ' + query);
         },
 
-        /**
-         * @ngdoc method
-         * @name umbraco.resources.mediaResource#getReferences
-         * @methodOf umbraco.resources.mediaResource
-         *
-         * @description
-         * Retrieves references of a given media item.
-         *
-         * @param {Int} id id of media node to retrieve references for
-         * @returns {Promise} resourcePromise object.
-         *
-         */
-        getReferences: function (id) {
+        getPagedContentReferences: function (id, options) {
+
+            var defaults = {
+                pageSize: 25,
+                pageNumber: 1
+            };
+            if (options === undefined) {
+                options = {};
+            }
+            //overwrite the defaults if there are any specified
+            angular.extend(defaults, options);
+            //now copy back to the options we will use
+            options = defaults;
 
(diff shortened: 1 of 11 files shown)


src/Umbraco.Web.UI.Client/src/common/resources/media.resource.js
```diff
@@ -554,28 +554,88 @@ function mediaResource($q, $http, umbDataFormatter, umbRequestHelper) {
...
+                            pageNumber: options.pageNumber,
+                            pageSize: options.pageSize
+                        }
+                    )),
                 "Failed to retrieve usages for media of id " + id);
+        },
+
+        getPagedMediaReferences: function (id, options) {
+
+            var defaults = {
```
Shazwazza: There's a lot of repitition here, getting the defaults can just be one method that is shared.

src/Umbraco.Web/Editors/MediaController.cs
```diff
@@ -945,70 +945,73 @@ internal static bool CheckPermissions(IDictionary<string, object> storage, IUser
             return hasPathAccess;
         }
 
-        /// <summary>
-        /// Returns the references (usages) for the media item
-        /// </summary>
-        /// <param name="id"></param>
-        /// <returns></returns>
-        public MediaReferences GetReferences(int id)
+        public PagedResult<EntityTypeReferences> GetPagedContentReferences(int id, int pageNumber = 1, int pageSize = 100)
```
Shazwazza: Seems like a lot of repitition here for 3 methods that all essentially do the same thing apart from a couple of references like the type of entity and the Udi, seems like this could all exist under one method with an object type parameter and another generic method to create the result?

src/Umbraco.Web/Models/ContentEditing/MediaReferences.cs
```diff
@@ -1,24 +1,9 @@
...
-using System.Linq;
-using System.Runtime.Serialization;
+﻿using System.Runtime.Serialization;
 
 namespace Umbraco.Web.Models.ContentEditing
 {
-    [DataContract(Name = "mediaReferences", Namespace = "")]
-    public class MediaReferences
+    [DataContract(Name = "entityType", Namespace = "")]
+    public class EntityTypeReferences : EntityBasic
```
Shazwazza: Is there any reason to keep this empty class?

warrenbuckley:
@Shazwazza ready for re-review & testing please :)

warrenbuckley:
Thanks for the feedback learned some new things & great point about RouteParams & infinite editing I had not thought about that.

PR #7162, opened 2019-11-16
removed the duplicate canAdd method from Nested Content

techmind-git:
### Prerequisites

### Description
I came across a duplicate method (method **canAdd**) in the angular controller for nested content which I removed.

Functionality should remain unchanged, which I tested with datatype "Product - Features - Nested Content" from the default starter kit.

_Note: this is my first pull request so if I didn't follow the contribution guidelines properly or something can be improved please let me know :-)._


techmind-git:
I updated the branch with all latest changes but this isn't really the history you'd want I believe so I'll close this PR as suggested before and open a new one that will only contain what is needed

Issue #6894, opened 2019-10-27
Opening Help Panel Without Connection to Internet Throws Exception


idseefeld:
Within a customers intranet solution, where the server has no internet access due to security policies, opening the help panel throws an exception and a red error message is shown.

Reproduction
------------
1. Install an Umbraco version 8.
2. Disable your network adaptors or unplug
3. Goto back office and open the help panel

### Bug summary

I located a server side HttpClient calling: https://our.umbraco.com/Umbraco/Documentation/Lessons/GetContextHelpDocs?sectionAlias=...
 
The frontend API call was:
http://localhost:52341/umbraco/BackOffice/Api/Help/GetContextHelpForPage?section=...

idseefeld:
There are more issues in other sections. 
I think all HttpClient use cases should be checked.

kjac:
Fixed in #7163

PR #7163, opened 2019-11-16
Fix HelpController for isolated intranet solutions

idseefeld:
Fix for issue: #6894 

External HttpClient call wrapped by try catch with log entry.


idseefeld:
@kjac, here comes the change ;-)

kjac:
Merged. Thanks again, @idseefeld 👍

PR #7165, opened 2019-11-17
removed unused _domainService

prjseal:
Removed unused _domainService which shows as warning in the build output.


kjac:
Merged. Thanks again, @prjseal 😄

Issue #7133, opened 2019-11-12
NC keys: Pasting Nested Content does not generate new keys correctly for subitems


jbjc:
Copying and pasting nested content that contains nested content does not work as expected. 

In the back end it seems to work and update fine, however: on the actual site the nested content's nested content seems to be still using the data from the original copied content.

## Umbraco version

I am seeing this issue on Umbraco version: 8.2.2 and 8.3.0

Reproduction
------------

### Bug summary

It seems like pasting nested content that contains nested content only generates a new key for the top level nested content. The embedded nested content that is pasted retains the same key as the original copied content.

### Specifics

No error messages, the content in the backend seems to update and save fine. It is only when viewing the content on the site when you notice the issue. I have looked at the keys for the nested content and the pasted content is definitely using the same key as the original. 

### Steps to reproduce

Create a document type with a nested content picker, this picker should pick a type that contains another nested content picker. This second nested content picker should another type that contains something like a text property.

1. Create the page
2. add the nested content to the page 
3. add nested content (populate the text) to this nested content
4. publish this
5. copy the top level nested content
6. paste the top level nested content 
7. modify the text in the new nested content
8. publish this

The text will appear saved in the backend, however, the keys for the original and copied text will be the same.

![image](https://user-images.githubusercontent.com/57661759/68659775-d0a07580-052f-11ea-8f96-dc60a2bd5c51.png)

(original content was 'test', as you can see the updated content is still showing as the original)
![image](https://user-images.githubusercontent.com/57661759/68659799-d9914700-052f-11ea-8576-b81f3b149243.png)

---
_This item has been added to our backlog [AB#3838](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3838)_

jason721:
My temporary fix, adding this to the end of the `pasteFromClipboard` function should give everything a new guid. 

```
              function setNewKey(values) {
                    for (var i = 0; i < values.length; i++) {
                        var obj = values[i];
                        if (obj.key) {
                            console.log("Setting key", obj)
                            obj.key = String.CreateGuid();
                            for (var k in obj) {
                               // if this has a property that is an array, we need to check if there's more keys to give a new guid to
                                if (Array.isArray(obj[k])) {
                                    setNewKey(obj[k])
                                }
                            }
                        }
                    }
                }

                for (var t = 0; t < newNode.variants[0].tabs.length; t++) {
[...]

PR #7166, opened 2019-11-18
NC keys: Do not copy keys for Nested Content (Fix for #7133)

nielslyngsoe:
To ensure unique keys for Nested Content entries, this PR will remove keys from nested-content properties when copying. So entries get new keys when pasted/created. This might be needed for other data types, but since there is no common practice for this I have chosen to only implement it for Nested Content. Meaning we might need to extend with further types, like Stacked Content etc.

This only works for Nested Content properties that have been initialized by the AngularJS Property Controller. Which could result in some entries not having a key. Therefore I propose that we ensure a key for each nested-content property in C#. A similar solution is needed to solve the issues in regard to the duplication of pages with NC and Blueprints with NC.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/7133

---
_This item has been added to our backlog [AB#4349](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4349)_


src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
```diff
@@ -53,11 +53,40 @@ function clipboardService(notificationsService, eventsService, localStorageServi
         return false;
     }
 
+    function removeKeys(values) {
```
Shazwazza: We need to make this more generic. This services should not know anything about nested content. The nested content property editor needs to be responsible for clearing these values. To do that we need to allow registering a callback or something where the property editor can say something like "for alias 'nestedcontent' call this method"

I've added a proposal for filling in blank keys here https://github.com/umbraco/Umbraco-CMS/issues/5913#issuecomment-568611712

nielslyngsoe: I agree. But we have an architectural problem. 

Cause, if we imagine that Stacked Content also had implemented a copy action, and we had a Stacked Content Element Type has a Nested Content Property, then we need to make sure that the Nested Content entry keys are handled, thought its the Stacked Content Actions handling this.

Might only be a problem with this solution, we might need to look at other solutions on this. The problem is that we cant be sure that any property Angular Code is initialized since the user might never navigate to the content of a copied entry. That's where I thought removing Keys on copy, would be good enough to make the Backend-code know that it should generate keys.

But let me know if you have a better idea.

Shazwazza: Yes removing the keys and having the back end deal with it is fine and is what I've mentioned in the other task. What I'm saying here is that the clipboard.service.js shouldn't need to know about property editors. And sure it's odd that one property editor might need to know about another one based on these nested things (in the future when we have a streamlined data structure we can prob handle that) but for now, Nested Content should be responsible for clearing it's own key, maybe based on a callback or something but this service shouldn't know what nested content is or any other property editors.

nielslyngsoe: Sounds like a good solution, let's get some time dedicated for that solution.

nielslyngsoe: Now I have provided a solution for this.

nielslyngsoe:
We still have an issue with Unit Tests, fixing it now.

nielslyngsoe:
Fixed

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
```diff
@@ -75,6 +97,21 @@ function clipboardService(notificationsService, eventsService, localStorageServi
...
+    * @ngdoc method
+    * @name umbraco.services.clipboardService#registrerPropertyClearingResolver
+    * @methodOf umbraco.services.clipboardService
+    *
+    * @param {string} function A method executed for every property and inner properties copied.
+    *
+    * @description
+    * Executed for all properties including inner properties when performing a copy action.
+    */
+   service.registrerClearPropertyResolver = function(resolver) {
```
Shazwazza, line 110: Unfortunately this is a spelling mistake: `registrer`

warrenbuckley: Fixed in this PR here - https://github.com/umbraco/Umbraco-CMS/pull/8218

Issue #7147, opened 2019-11-13
Embed function in RTE removes previous content from RTE


TimOldenkamp:
The old content in the RTE is deleted after use of the embed function.

## Umbraco version

I am seeing this issue on Umbraco version: 8.2.2

Reproduction
------------

### Bug summary

The embed function removes the content already inserted in the RTE.

### Steps to reproduce

1.  Add text to the RTE (and save the page)
2.  Go to the same RTE and use the Embed function. For example add a Youtube-video.

![Vimeo-movie](https://user-images.githubusercontent.com/960810/68747348-a1a30600-05fa-11ea-8ad7-2cf4eaa2d267.gif)

### Expected result

The old content (for example text) is still in the RTE and the embedded item is added.

### Actual result

Only the new embedded item is in the RTE. The old content (for example text) is deleted from the RTE.

---
_This item has been added to our backlog [AB#3794](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3794)_

nul800sebastiaan:
Unfortunately for you, I cannot reproduce the issue on a clean 8.2.2 install, maybe you've upgraded this from an older version and there's a problem with the upgrade? If you have better steps to reproduce in a clean install of 8.2. then let us know, but for now I'll have to close this one as I don't know how to make it fail. :-)

TimOldenkamp:
Steps to reproduce:
1. Do a clean install of version 8.2.x or version 8.3.0 with The Starter Kit
2. Change the _Contact - Contact Intro - Rich Text Editor_ to support the Embed
3. Go to the page _Contact_. The default text at the property _Contact Intro_ is 'Lorem ipsum dolor sit amet...' Now press the Embed and insert a YouTube video.
4. The result in the property is the YouTube video but the original 'Lorem ipsum dolor sit amet...' is gone...

Tested en reproduced with Chrome (v78.0.3904.97 64-bit) and Firefox (v70.0.1 64-bit)

nul800sebastiaan:
Right, I see what you mean now! I actually typed words in the RTE before inserting the youtube video. 

It seems to work fine if you use an embed when on a new line but indeed I can see it removes the line that you are on if it's not a new line. We'll have to have a look at this!

PR #7167, opened 2019-11-18
Embed function in RTE removes previous content from RTE

nielslyngsoe:
Only replace content if the active element is of type EmbedItem.
If a selection has been we will still be replacing the selection.

Test notes:
One document type with a Rich Text Editor property.

test A:
— Write some text, keep the pointer after the last character (do not make a new line) and then add an EmbedItem. See that text is still present.

test B: 
— Add an image, and then add a EmbedItem, see that Image is still present.

test C:
— Add an embed item, and then add another embed item.

test D:
— Add an embed item, and then select it, for then to be able to edit the EmbedItem by clicking the Embed-button. See the EmbedItem begin replaced with the updated one when you save the EmbedEdit changes.

Fixes https://github.com/umbraco/Umbraco-CMS/issues/7147

---
_This item has been added to our backlog [AB#3856](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3856)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
@@ -544,7 +544,7 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
                 },
                 embed.preview);
             
-            // Only replace if actieElement is an Embed element.
+            // Only replace if activeElement is an Embed element.
             if (activeElement && activeElement.nodeName.toUpperCase() === "DIV" && activeElement.classList.contains("embeditem")){
                 activeElement.replaceWith(wrapper); // directly replaces the html node
             } else {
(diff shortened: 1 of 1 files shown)


src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
```diff
@@ -543,11 +543,11 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
                     'contenteditable': false
                 },
                 embed.preview);
-
-            if (activeElement) {
+            
+            // Only replace if actieElement is an Embed element.
```
warrenbuckley:
```suggestion
            // Only replace if activeElement is an Embed element.
```

warrenbuckley:
Code looks fine, off to test this out...

warrenbuckley:
All good 👍

PR #7179, opened 2019-11-19
Netcore: Movement of package abstractions

bergmania:
**This needs to be merged first: https://github.com/umbraco/Umbraco-CMS/pull/7176**

Movement of package stuff.

Had to remove the default value of `UmbracoVersion` from `PackageDefinision`, and handle the null value in `PackageDefinitionXmlParser` instead. There I could inject the IUmrbacoVersion.


Shazwazza:
@bergmania looks good, i realized the build was failing after i merged so will check it out to make sure everything is ok.

Shazwazza:
@bergmania it's the damn csproj file that has merge issues all the time, have fixed and pushed.

Issue #7181, opened 2019-11-19
Infinite editing doc type throws JS error when template is allowed


hemraker:
When editing a Document Type with Infinite Editing, a JS error is thrown in the console if a template is allowed on the Document Type.

## Umbraco Version
8.1.5+ (Tested 8.1.5, 8.2.1, 8.3)

### Specifics

Edit any document type, that has templates allowed, through infinite editing. Saving the doct type will produce following JS error:
```
Possibly unhandled rejection: Could not find the tree templates angular.js:15536:41
    Angular 8
        consoleLog
        get
        processChecks
        $digest
        $apply
        done
        completeRequest
        requestLoaded
```

### Steps to reproduce

1. Install 8.1.5+ with starterkit
2. Go to Home content node
3. Go to the Info tab and open the Home document type
4. Press Save and Close
5. Error is thrown in console

### Expected result

I expected the doctype to be saved without throwing an error.

### Actual result

Doc type is saved but with error in the JS console

---
_This item has been added to our backlog [AB#3877](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3877)_

kjac:
I can reproduce this... and I've found the error. I'll submit a PR tonight 👍

kjac:
PR in #7183

nielslyngsoe:
Though KJACs request follows the way this issue has been solved before in that file, I have chosen to grab the option to solve this in the right way to avoid having issues in the future.

So a new PR for this issue:
https://github.com/umbraco/Umbraco-CMS/pull/7207

warrenbuckley:
Closed by PR #7207

PR #7183, opened 2019-11-19
V8: Fix JS error when editing content type in infinite mode

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7181

### Description

See steps to reproduce in https://github.com/umbraco/Umbraco-CMS/issues/7181 - the error manifests like this:

infinite-edit-content-type-js-error

It happens because we attempt to reload the templates tree when the content type is saved, and of course this won't work very well in infinite editing. So this PR ensures that we do not attempt to reload the templates tree when in infinite editing mode.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/documenttypes/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/documenttypes/edit.controller.js
@@ -519,7 +519,8 @@
         }));
 
         evts.push(eventsService.on("editors.documentType.saved", function(name, args) {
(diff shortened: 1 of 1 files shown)


nielslyngsoe:
Hi @kjac, Thanks as always for your PRs! 
I do though not think it's a stable solution to check for infinityMode then the problem occurs because the tree isn't present.
Therefore I will close this one and instead I have made a PR that tests for the presence of template tree.
Thanks for your suggestion :-)

PR #7184, opened 2019-11-19
V8: Make the sticky header span the full width of its container

kjac:
### Prerequisites

### Description

⚠️ sort of large GIFs ahead, apologies in advance ⚠️ 

The sticky header bar doesn't span the full width of the content editor. This is most notable when using the grid. The sticky header _does_ however span the full width when editing content types, both in and out of infinite editing mode:

sticky-header-before

This PR forces the sticky header to be just slightly wider than the parent container in any context it's shown. This seems to do the trick:

sticky-header-after


emmaburstow:
Urgh - great change. That did not look good before. You're a 🌟 Kenn,

Em

kjac:
Thanks @emmaburstow 😄 

I've changed the base of the PR to `v8/contrib` and fixed the merge conflicts. Should be all good to go.

poornimanayar:
Thanks again and merged !

PR #7186, opened 2019-11-20
add disable-hotkeys attribute to disable hotkeys

### LINKED ISSUE #6508: Pressing '1' or '2' inside a Tiny MCE Rich Text Area running in 'distraction free' mode causes the page to jump up to the top and 'focus' on  'Content' and 'Info' Content Apps
Using Umbraco 8.2RC

Create a Property based upon a Rich Text Area data type.
Set the data type to run in 'distraction free' mode
Type some text, and press the '1' key or the '2' key, the focus is removed from the rich text area (and if your editor is a little bit down the page the page will scroll up) and passed onto either the 'Content' or the 'Info' Content App Icon...

This is probably a recent accessibility change to allow you to toggle between the two using access keys, and probably something is added to tiny mce to prevent this happening when the numbers are entered into the rich text area, but the same workaround is not applied in 'distraction free mode'...

... but I am guessing...

I would expect the number 1 or the number 2 to appear inside the rich text area instead...

image

### LINKED ISSUE #7168: Pressing '1' or '2' in RTE results in selecting Content or Info tab
When adding a property of type RTE (Distraction free) and typing in '1' or '2' the current behavior is a bit strange. Instead of displaying '1' or '2' Umbraco selects the 'Content' (1) or 'Info' (2) tab. So it seems that when you have RTE of type distraction free on you content page that pressing 1 or 2 is acting as a shortcut to Content and/or Info.

Umbraco 8.2.1

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #7168 and fixes #6508 

### Description
Issue here was in keyboard service, where keypresses are checked before propagating events for triggering hotkeys. Keypresses originating in INPUT or TEXTAREA nodes were explicitly prevented from bubbling/propagating, but the distraction-free rte is a content-editable DIV, so always allows propagation. Classic RTE is an iframe, so no propagation.

To fix this, I've added an attribute - `disable-hotkeys` - to the RTE element. This could be used in other places too if needed, for now only added to fix this issue. The same check for INPUT or TEXTAREA tag name now checks for that attribute, and the issue is gone (line 216 in keyboard.service.js - encoding/whitespace blew up the diff). 

To verify, add a distraction-free RTE to a content type, and note typing 1 or 2 doesn't trigger hotkeys, but adds the value to the RTE.


nathanwoulfe:
@kjac nah, that should not have been included. Just tried reverting the changes, didn't go well.

My git-fu is weak.

kjac:
Uh ouh. Perhaps your GitHub-fu is better! I believe you can just delete the entire file from the PR in the changes section? Look for three little dots to the top right of the file, and perhaps they shall illuminate your path. 

Wax off.

nathanwoulfe:
Not sure that did what we thought it would do.

kjac:
🙈 my bad... sorry mate. My GitHub-fu is poor too! 

Right... so guess the only thing left to do is to copy in the `package-lock.json` as it looked previous to your PR (in 5a78c28d). 

I have created a gist with it here. Could you try downloading it and copying it into `src/Umbraco.Web.UI.Client/`?

https://gist.githubusercontent.com/kjac/3677d5281dc961ce0e733f393174d821/raw/3a9e09d27b8e6bbe0bf87f5a52e279ee1fba9241/package-lock.json

nul800sebastiaan:
That took a bit of wrangling, but all fixed now.

nathanwoulfe:
And this is why we all love @nul800sebastiaan

nul800sebastiaan:
I tried to make a gif to show how I normally do it but it didn't work this time because of the file being deleted. I'll try to do it on another PR when I come across it so I can document it! :)

kjac:
Thanks again, @nathanwoulfe ⭐️

nathanwoulfe:
Not a problem my good man 😁

nul800sebastiaan:
Hi @nathanwoulfe and thanks again for this fix! I just noticed that when the RTE is in a Grid then pressing `1` and `2` is still problematic! Also, and only in an RTE in the Grid, you cannot type a space..! 🤯

There's an issue for it - https://github.com/umbraco/Umbraco-CMS/issues/7572

Any chance you have some time to look into those cases as well please? 🙏

nathanwoulfe:
Who needs spaces anyway... Hopefully adding the attribute to the grid rte will fix it all. Will have a look this morning

nathanwoulfe:
disable-hotkeys attribute fixes the 1 and 2 issue, will push that here as it's directly related to this issue. #7572 is related, but will be a different solution.

nathanwoulfe:
New PR is #7604

PR #7188, opened 2019-11-20
V8: Do not submit the current content editor when editing a media picker item

### LINKED ISSUE #7185: Clicking edit on media picker launches the save dialog
If you have a media picker and you click the edit (pencil) button the document is Saved or the Save dialog is launched (if it's a variant)!

See gif: 

EDC6226A-8623-4637-8E65-67041DD59F43

If its invariant, clicking the pencil actually just automatically goes and saves the document without any warning.

Version: Latest dev/v8 branch (currently set to 8.4.0)

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7185

### Description

See #7185 for an issue description. To reproduce it you need a content type where the media picker is *not* in multipicker mode and is somewhat topmost in the list of property editors.

Here's a GIF of the issue in action:

edit-media-item-before

It is a side effect of turning the buttons into actual <button> tags instead of <a> tags (in an effort to make the media picker more accessible, see #6805). Depending on your property editor order and configuration this means that the browser might see the link picker edit media button as the first <button> element in the form and thus treat it as a submit button.

This PR fixes it by applying an explicit type to the edit button (`type="button"`), thus explicitly instructing the browser how to interpret it. The `prevent-default` directive could also have been used here, but the proposed solution leverages the built-in browser functionality and I do think that's preferable.

Here's a GIF of the fix in action:

edit-media-item-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.html
@@ -36,7 +36,7 @@
                 </umb-file-icon>
 
                 <div class="umb-sortable-thumbnails__actions" data-element="sortable-thumbnail-actions">
-                    <button aria-label="Edit media" ng-if="allowEditMedia" class="umb-sortable-thumbnails__action btn-reset" data-element="action-edit" ng-click="vm.editItem(media)">
+                    <button type="button" aria-label="Edit media" ng-if="allowEditMedia" class="umb-sortable-thumbnails__action btn-reset" data-element="action-edit" ng-click="vm.editItem(media)">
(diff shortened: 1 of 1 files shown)


MMasey:
Thanks for picking this up @kjac, the bug was my bad. You may want to add the `type=“button”` to the delete button too to cover all grounds. 

Thanks again

kjac:
Aye. The remove button seems to work just fine also after changing the edit one 👍

bjarnef:
@kjac I think the delete button should have `type="button"`. Without this attribute some browsers has "submit" as default behaviour, while other have "button" as default.

Note how the validation is triggered in the "content" tab, when removing the media. I think the validation only should trigger when saving the node.

kjac:
Fixed in #7651 so I'm closing this PR.

PR #5312, opened 2019-04-23
Refactor IPublishedContent

zpqrtbnk:
Fixes #5170

PR #7189, opened 2019-11-20
Make sure the configured url provider mode is used when getting a url

dawoe:
### Prerequisites

### Description

You can configure Umbraco to generate absolute urls out of the box by changing the web.Routing section umbracoSetttings.config file.  You can do this by settting `urlProviderMode="Absolute"`

So by setting this I would expect umbraco to generate absolute urls when I request the url through the url property of a IPublishedContent item.

However on 8.3.0 it still generates relative urls after setting this in the config. This is cased by the default value for mode being Auto on the Url method in PublishedContentExtensions. 

I changed the default value of the mode to UrlMode.Default. And when it is set it will get the mode configured on the url provider. This is backwards compatible because when there is no urlProvider mode configured in the web.Routing section it will default to Auto.

Can this be part of 8.3.1 ? This is blocking me now. I don't want to write a lot of custom code just to have absolute urls

Dave

### DIFF
--- a/src/Umbraco.Web/PublishedContentExtensions.cs
+++ b/src/Umbraco.Web/PublishedContentExtensions.cs
@@ -1193,7 +1193,7 @@ namespace Umbraco.Web
         /// if any. In addition, when the content type is multi-lingual, this is the url for the
         /// specified culture. Otherwise, it is the invariant url.</para>
         /// </remarks>
-        public static string Url(this IPublishedContent content, string culture = null, UrlMode mode = UrlMode.Auto)
+        public static string Url(this IPublishedContent content, string culture = null, UrlMode mode = UrlMode.Default)
         {
             var umbracoContext = Composing.Current.UmbracoContext;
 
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Quote from Shannon: 
> I don't think his extra if statement is required though, the mode should just be passed to the UrlProvider.GetUrl which handles the mode in there and already does the default check - you can have a look by following the call into UrlProvider.GetUrl and UrlProvider.GetMediaUrl (both of which have Default as the default)

Other than that I can confirm this is a regression from 8.0 created in 8.1 in this PR: https://github.com/umbraco/Umbraco-CMS/pull/5312

dawoe:
Hi @nul800sebastiaan and @kjac 
I removed the if statement

kjac:
Thanks @dawoe 👍 I'll get on it.

kjac:
Merged 😄 thanks again, @dawoe

nul800sebastiaan:
Note: I've marked this as breaking, not because I think it will break anything but if someone has accidentally been using `Absolute` but not realized that it doesn't do anything, they'll see a bit of different behavior now.

Issue #7173, opened 2019-11-19
.Net Core transition: Inject IUmbracoVersion wherever possible


bergmania:
**This task is up for grabs, for contributing to the .Net Core transition project**

On the development branch of the .net core transition project [`netcore/dev`](https://github.com/umbraco/Umbraco-CMS/tree/netcore/dev), we have already abstracted the static `UmbracoVersion`, into an interface (`IUmbracoVersion`) and an instance class (`UmbracoVersion `).

This was required to unblock some other tasks. To finish that task as fast as possible, we just added a new static property on the static helper `Current` (`Current.UmbracoVersion`). This new static property is used all over the system, instead of the original static `UmbracoVersion`.

This task is about eliminating the usage of `Current.UmbracoVersion`, by replacing it with a constructor injection of `IUmbracoVersion`.

Be aware, `Current` exists in both `Umbraco.Core.Composing` and `Umbraco.Web.Composing`.

Expected Procedure:
- Find an instance class that uses `Current.UmbracoVersion`.
- Add `IUmbracoVersion umbracoVersion` to the constructor.
- Save the umbracoVersion in a `private readonly IUmbracoVersion _umbracoVersion;` field.
- Replace the usages of `Current.UmbracoVersion.*` with `_umbracoVersion.*`.
- Make sure the project builds
- Ensure unit tests all pass
- Repeat the process

If any questions, feel free to ask :)

___
Related: [AB#3865](https://umbraco.visualstudio.com/D-Team%20Tracker/_workitems/edit/3865)

crgrieve:
Hello @bergmania,

I am happy to help on this one if it's still up for grabs? If so, I have done one instance of it as an example to make sure I am understanding ok: https://github.com/crgrieve/Umbraco-CMS/commit/015d5882184b0ae43d9745cc42dc54c614941b3e

Is this what you are looking for? If so, I am happy to work through the other instances of Current.UmbracoVersion

Also, on your description, I assume you mean _umbracoVersion.* rather than _ioHelper.*?

Thanks,
Carole

bergmania:
Hi @crgrieve,
Yep, it should have been `_umbracoVersion.*`, good catch. Note to self always triple check when you copy/paste 😄 

Yes, it is still up-for-grabs.
Your example is exactly what to do - Good work 😃 

Feel free to split the task in multiple PRs, if you think that make sense.

crgrieve:
Thanks @bergmania 😊

[...]

PR #7190, opened 2019-11-20
Injecting umbraco version

crgrieve:
As discussed on issue #7173 with @bergmania, here is a first attempt at replacing the current.umbracoVersion with an injected IUmbracoVersion.

Thanks,
Carole


bergmania:
Thanks @crgrieve.

I tested and took a few more.. If CI is still happy I will merge :)

PR #7192, opened 2019-11-20
V8: Retain mculture when clicking results from global search

kjac:
### Prerequisites

### Description

The backoffice URL looses the current `mculture` when you open a search result. If you're editing a non-default language, this leads to a quite inconsistent editor experience: The initialized part of the backoffice still respects the previous `mculture`, but anything loaded subsequent to opening a search result performs a fallback to the default culture:

retain-mculture-from-search-before

Historically there have been all sorts of mishaps when `mculture` goes missing in a multi lingual setup, so this should be prevented at all cost. 

It seems this is caused by the routing changes introduced in #6664.

This PR applies a slightly different approach to these changes, thus ensuring we can still retain the current `mculture` across the backoffice on route change if no `mculture` is explicitly specified in the new route. 

Here's the fix in action:

retain-mculture-from-search-after


emmaburstow:
👌

mikecp (approved):

mikecp:
Hello @kjac, better late than never, I just checked and the problem still exists 😅
Thanks for the fix ! 👍

PR #5580, opened 2019-06-04
V8: Update create options for content types

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description

The current options for creating a content type is very V7-ish:

image

The way I see it, almost of the use cases for "Document type without a template" has been replaced by "Element type" (a.k.a. blueprints for Nested Content etc.).

On top of that I feel like we should help the users create culture variant content types right from the "create" context menu. This would at least partially mend the issues that have popped up since the release of V8 with people realizing too late that their content type was *not* culture variant after creating a bunch of properties, and thus having to re-configure the culture variance on each property.

With this PR applied, the "create" context menu removes "Document type without a template" and introduces "Document Type - varying by culture" and "Element Type":

create-content-types

Ideally I'd like to get rid of "Document Type Collection" too as it is utterly un-intuitive, but that's another discussion 😄

PR #7196, opened 2019-11-20
Introduced new "Element Type" option for the create document type dialog

abjerner:
When using Nested Content or similar and creating many element types, it would be useful with an **Element Type** option in the create dialog. So here is a PR for that.

image


emmaburstow:
👌 Niiiiice! This is a fab change

kjac:
@abjerner I agree, there is a real need to have a look at the create options here. But HQ is currently working on a new layout for this menu - see @nul800sebastiaan's comment [here](https://github.com/umbraco/Umbraco-CMS/issues/6410#issuecomment-536640835). #5580 was closed for that same reason.

nul800sebastiaan:
I'll close this one in favor of https://github.com/umbraco/Umbraco-CMS/issues/6410 - which is something we want to work on in the future but is not quite on the roadmap at the moment, sorry.

PR #7197, opened 2019-11-21
V8: Use the current client culture in global search

kjac:
### Prerequisites

### Description

_Note: This PR builds upon and utilizes the changes in #7123, which should then probably be merged before this one._

When editing a non-default culture and using the global search to find content, the search results are displayed in the default culture. This is somewhat confusing to the editors:

global-search-culture-before

This PR ensures that we use the currently active client culture when searching content in the global search:

global-search-culture-after

The implementation isn't the prettiest. I would have preferred to add `culture` to the `Search` method of `ISearchableTree` and let the caller (`EntityController`) pass the current client culture explicitly. But that would be a breaking change to anyone having created their own `ISearchableTree` implementations.


emmaburstow:
👌

poornimanayar:
Hi @kjac ,

Been testing this. I am so sorry I am stuck with the same kind of issue that I had with another PR of yours! So the question is if I have a different node name in a particular language do you expect the global search to return the node name in that language or the default one? The code implies that I should see the node name in the language although what I see is the node name in default language and url in current language. I am not sure about this, but looking at internal index the node names are stored as `nodeName_en-gb` (culture in lowercase) as opposed to `nodeName_en-GB`. So you might need a `ToLower()` where you do the string interpolation in `EntityMapDefinition`.  

Poornima

kjac:
@poornimanayar I'd expect the search results to be presented in my currently chosen main language. Instead the results are always displayed in the default language. Here's how the search results are currently being presented for non-default languages in the latest `v8/contrib`:

image

And here are the same search results with this PR applied:

image

poornimanayar:
Hi @kjac ,

Here is a screenshot from my testing. The url returned from search is in the current language but the node corresponds to the name from default language. 

Content - umbracov8 local

Poornima

kjac:
@poornimanayar that's wonky; but all the same, if the PR fixes your search so your node name is correct according to the current culture, it should be all good?

Issue #5259, opened 2019-04-14
U4-10518 - Adding a ListView to the Media Picker


marcemarc:
Moving this issue from the old issue tracker...

... not least as I can see that @kjac is updating ux for the Media Picker in another PR, and also updating sorts in listviews ... and I'm confident I can pique his interest just enough to roll together a smoother interaction than what I put together as a proof of concept back in v7! (I'm aware of how machiavellian this is of me :-P)

Essentially navigating through the Media section from a picker, when there are lots of folders, can be slow, and it can be hard to pick an image if you only know the filename... allowing the editor to toggle between listview (A-Z and no images, just filenames) and then back to images, speed up these common journeys, I have it implemented on a number of V7 sites, but it's not really releasable as a package without maintaining a seperate Media Picker

There is a video here: https://youtu.be/NsmOnXJY1gg

Original issue: https://issues.umbraco.org/issue/U4-10518

kjac:
@marcemarc I see what you did there 😆 

It's not a bad idea. But does it really solve the actual editorial challenge "I don't know where the stuff is"? The editors still have to find the correct folder first. If the editors have been good at subdividing the media library into folders, finding the correct folder can be a hassle in itself.

My initial thought was "why not just use the search field?". After a bit of testing that still stands. The search results include *both* folders and files, which means you can find both the correct folder and the correct file if you have any idea of their respective names - or at least narrow down the options pretty quickly.

That being said, a listview toggle could still make sense, specially for folders with many items in them. However... folders can be explicitly configured as listviews, and this is reflected in the media picker. So the listview toggle should only be shown for folders that aren't configured as listviews, and this raises a bit of a consistency concern in my book.

Thoughts?

marcemarc:
@kjac first of all, thanks for thinking about it, I raised it again mainly as it seemed the two things you were looking at in other PRs, might coalesce into this old issue. The issue was raised back in the day before 'Search' searched the whole of the Media section - so agree search does mitigate a number of the picking issues.

The search does not search filenames (as far as I can tell, only the names of the Media Items - now these will probably be the same if they are bulk uploaded, and the name automatically assigned from the filename - but not always the case...

Search can return duplicate items, thinking particularly 'files' rather than 'images' without the context of the folder they are in, it's really difficult to tell 'which is actually the one' to pick. (Think folders within sections all called 'Page Banners' etc)

[...]

PR #7198, opened 2019-11-21
V8: Add alternate 'List View' to Media Picker 

marcemarc:
### Prerequisites

- [~] I have added steps to test this contribution in the description below

https://github.com/umbraco/Umbraco-CMS/issues/5259

### Description

I have had a go at this... as directed on the issue tracker... see (https://github.com/umbraco/Umbraco-CMS/issues/5259) but think there are some outstanding points to consider before it could be merged, and oh of course it's unlikely my attempts at angularJS and the styling of the interaction follow 'best guidelines'... but hopefully by attempting this, the idea can be realised with other people's input/help.

What does this Pull Request do?

It adds a little icon in the Media Picker to enable editors to 'toggle' between the normal view of images to pick from and an alternate listview (the idea is we think more like a file system, when we are picking assets, and if you know the name of the file you want to pick, it's quicker to pick from a list than scroll through possibly lots of similar images with similar filenames).

The idea is to never 'crop' the name of the Media Item, so the editor can always see the whole name of the item without the ... obscuring the last bit (often the bit that distinguishes one file from another)

It makes the Media Picker slide out panel wider using the new 'medium' setting

It allows the list view of media items to be sorted by 'Name' or by 'Last Update Date'*

There is a video here:

http://snowmoon.co.uk/mediapickerlistview2.gif

The Media Picker 'should' just work in terms of its existing functionality, and it should be tested with different configurations, eg picking folders, images only etc, and also should be tested from inside the Rich Text Editor and Grid.

I've been using this on V7 for two years now, but a simpler version.

## Caveats Things to Consider

1) EntityBasic - in V7 you had the UpdateDate and CreateDate supplied to the items in the MediaPicker (largely unnecessarily!), in V8 this has been streamlined and they have been removed - in This PR I add ~~CreateDate~~ UpdateDate into the EntityBasic class and the EntityMapDefinition - I have no idea if this is a good idea or breaks something else using the EntityBasic class elsewhere in Umbraco.

2) UpdateDate - ~~I want to display the Last Updated Date in the ListView, however in investigating the disappearance of the CreateDate and UpdateDate from the MediaItem, I found that the UpdateDate was always null (1st of Jan 0001) in the EntityService, but the CreateDate was populated - I've used CreateDate to illustrate the interaction in this PR - and if this is something that we want to progress further, would create a different issue in the tracker as to why UpdateDate is null, so it can be used here instead... broadly speaking CreateDate isn't tooo far away from the UpdateDate, when files are uploaded etc~~ I've added VersionDate from the ContentVersion table into the select and groupby clause when bringing back entities in the EntityRepository, so now can display the UpdatDate for the media item... but is this ok to do? 

3) With the Media Picker you can configure to 'pick folders' but also no 'pick folders' when I built this for V7 I had a Caret character (little triangle) indicating the option to expland a folder rather than pick it... this worked reasonably well as the V7 picker I built looked a bit like the content tree - but I've struggled to make sense of this in a table format... if anyone has any ideas...

image

4) Switching Icon - Have I got the Icons 'around the right way'?- should this interaction be using the layout switching component used in the List View?

5) Sorting - I've kept 'folders' always together when sorting, so A-Z is A-Z of Folders followed by A-Z of files - this feels like how a file system works.. but does it make sense?

6) Go further - should the filename / file extension be visible in an additional column in the ListView - should you be able to 'select' all the items in the ListView with one checkbox in the header?

7) I didn't use the umbTable component, as I couldn't work out how to vary the contents of the cells, but I have tried to use the markup, didn't know whether the whole list view should be a seperate component or try to use the mini list view... I just don't know... my thinking was this is always a variation of the 'Media Grid' - and was ok to put in here... 

8) Position of the switching icon - the direction was to put this level with the breadcrumb... which I think I've done by putting it in the mediapicker html rather than inside the umbMediaGrid... but not sure if this is correct, also there is no breadcrumb when searching, so I've just made it appear on the right...

9) there is sometimes a weird blue border when you select things in the list view, that disappears when focus moves away, can't find where it's coming from.

@kjac this was the thing I mentioned at the UK festival


marcemarc:
So I've started updating other entities that inherit entitybasic to have an updateDate property to make the warnings on the build go away... but I've stopped as this feels wrong.

I'm thinking I need to implement a 'MediaEntityBasic' class that has the UpdateDate, (or could I use MediaSlim - not sure if that is something different>) and then somehow update the entityResource (or should it be in the mediaResource) so that the mediapicker.controller.js can call an endpoint that ultimately returns MediaEntityBasic instead of EntityBasic... so I'm not affecting anything else that inherits EntityBasic in Umbraco... ?

abjerner:
@marcemarc didn't get to test this over the weekend, but found some time now. When building, I'm getting the following errors:

image

Fixing those let me build again, but I'm really not sure how to test this.

Anyways, I was testing something on `v8/contrib`, so my clone is now running 8.7 opposed to 8.6, and therefore not possible to downgrade. I'll see if I can test it a bit further tomorrow 😉

marcemarc:
@abjerner Thanks for having a look, yes this is the dilemma... the initial feedback on the issue was to add in sorting by Last Updated Date. (https://github.com/umbraco/Umbraco-CMS/issues/5259#issuecomment-546268094)

To do that requires retrieving that column from the database.

The media picker uses the entityrepository, which returns an EntityBasic...

... so I updated this to add the UpdateDate.

This seemed to work, eg didn't flag any issues with build until I created the PR, and of course, there is a warning for mapping rules if a property isn't mapped, and lots of other stuff inherits from entitybasic...

... so dilemma I have for this PR is

1) implement UpdatedDate for all things that inherit EntityBasic, and their Mapping rules
2) update entityrepository for media requests to return a new LastUpdatedEntityBasic, which is the EntityBasic with the LastUpdated date, and corresponding mapping, so that other mappings and objects that inherit from entitybasic are unaffected
3) create a new LastUpdatedRepository to return the LastUpdatedEntityBasic, so it's not confusing that the entityrepository is now returning something other than entitybasic...
4) something else
So all options... quite a bit of work, so I'm loathed to do one way, and then find the other way or another way would be preferable for handling this in the core... but maybe I have to choose one so people can see that it's sooo wrong? etc and then be able to point out better ways. (and what to name things)

But that is why I'm stuck!

marcemarc:
Closed in favour of https://github.com/umbraco/Umbraco-CMS/pull/7912

PR #7201, opened 2019-11-21
AB3875 untouchability checker: do not check style for document fragments

nielslyngsoe:
Do not test style on DocumentFragment.

Test notes: Go to DocumentType Editor and see if error still appears in console.

Mark AB3875 as done when merged, thanks!

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/tabbable.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/tabbable.service.js
@@ -184,7 +184,13 @@
               });
               if (cached) return cached[1];
 
-              nodeComputedStyle = nodeComputedStyle || this.doc.defaultView.getComputedStyle(node);
+              if (!nodeComputedStyle) {
+                if (node instanceof DocumentFragment) {
+                  return false;
+                } else {
+                  nodeComputedStyle = this.doc.defaultView.getComputedStyle(node);
+                }
+              }
 
               var result = false;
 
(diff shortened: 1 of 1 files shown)


src/Umbraco.Web.UI.Client/src/common/services/tabbable.service.js
```diff
@@ -184,7 +184,13 @@
               });
               if (cached) return cached[1];
 
-              nodeComputedStyle = nodeComputedStyle || this.doc.defaultView.getComputedStyle(node);
+              if (!nodeComputedStyle) {
+                if (node instanceof DocumentFragment) {
+                  return true;// though DocumentFragment dosnt directly have display 'none', we know that it will never be visible, and therefor we return true. (and do not cache this, cause it will change if appended to the DOM)
```
warrenbuckley:
```suggestion
                  return true;// though DocumentFragment doesn't directly have display 'none', we know that it will never be visible, and therefore we return true. (and do not cache this, cause it will change if appended to the DOM)
```

warrenbuckley:
OK gonna close this PR and I will cherry pick your commits @nielslyngsoe so it targets just `v8/8.4`

warrenbuckley:
Updated PR that cherry picked these commits against v8/8.4 branch as opposed v8/dev aka 8.6.0 is here - https://github.com/umbraco/Umbraco-CMS/pull/7203

PR #7205, opened 2019-11-22
Moves remaining service abstractions

Shazwazza:
Moves the remaining service abstractions and ext classes to Abstractions project.

Few things that changed:

* Registers `ICultureDictionary` in DI so it can be resolved directly instead of having to create a new one each time from `ICultureDictionaryFactory`
* Removes more usages of Current
* Creates `IPropertyValidationService` (based on TODO notes which removes more usages of Current), this unfortunatley needs to be injected as lazy since it requires PropertyEditorCollection and thus we'd get a circular reference but we can live with this for now
* Removes unused methods that relied on System.Web (membership)

links [AB#3915](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3915)


bergmania:
Code looks fine 👍

PR #7207, opened 2019-11-22
check if template tree exists, to avoid error if its not present.

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/7181


warrenbuckley:
Code looks fine and has been tested before and after 👍

PR #7211, opened 2019-11-22
Rename reference to info

nielslyngsoe:
Renaming "Reference" Tab to "Info" Tab for DataType Editor Plus moved two informatic properties to the Info tab.


warrenbuckley:
Code move about looks fine, off to test to check it looks all a-okay :)

warrenbuckley:
It works fine but the UI bugs me a tiny bit that it does not line up (Maybe we can fix it up somehow)

image

bjarnef:
It would be great to be able to request the datatype from the guid shown, see PR https://github.com/umbraco/Umbraco-CMS/pull/6847

PR #7212, opened 2019-11-22
Segments in Backoffice: Server code

PerplexDaniel:
### Description
Server-side changes for segment support in the Backoffice. This makes sure segments will be sent to the backoffice when there is segment data in the database, and also makes sure segment data is saved back to the database on save/publish.

Note this cannot really be tested in isolation (barring perhaps some regression tests). To test this in the interface you would also need to merge #7213 and perform some additional steps, which are described in #7213.

---
_This item has been added to our backlog [AB#4124](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4124)_


src/Umbraco.Web/Models/Mapping/ContentVariantMapper.cs
```diff
@@ -21,52 +21,110 @@ public ContentVariantMapper(ILocalizationService localizationService)
...
-                var variants = langs.Select(x =>
+                if (languages.Count == 0 || segments.Count == 0)
                 {
-                    //We need to set the culture in the mapping context since this is needed to ensure that the correct property values
-                    //are resolved during the mapping
-                    context.SetCulture(x.IsoCode);
-                    return context.Map<ContentVariantDisplay>(source);
-                }).ToList();
+                    // This should not happen
+                    throw new ArgumentException("No languages or segments available");
```
Shazwazza: Probably InvalidOperationException here, don't think ArgumentException is the right one

PerplexDaniel: Agree, pushed the change

src/Umbraco.Web/Editors/Binders/ContentItemBinder.cs
```diff
@@ -49,7 +49,7 @@ public bool BindModel(HttpActionContext actionContext, ModelBindingContext bindi
             {
                 foreach (var variant in model.Variants)
                 {
-                    if (variant.Culture.IsNullOrWhiteSpace())
+                    if (variant.Culture.IsNullOrWhiteSpace() && variant.Segment.IsNullOrWhiteSpace())
```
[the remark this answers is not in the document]
PerplexDaniel: The original code treated invariant code separately from variant code, so I updated the "is it invariant"-check to include segments. That is, it is only invariant if both the culture and segment are null/whitespace. Perhaps invariant does not have to be treated differently, then this could indeed be removed.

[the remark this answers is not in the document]
Shazwazza: From what I can tell in code, any call to context.GetCulture would also check if the underlying thing it's working with is variant or not so seems to me like we could avoid this check and just SetCulture/SetSegment regardless since if they are null that is fine. 

I'll test out these changes - i'll merge the client side stuff in temporarily but will revert it afterwords and we can deal with the client side stuff later.

src/Umbraco.Web/Models/Mapping/ContentVariantMapper.cs
```diff
@@ -21,52 +21,110 @@ public ContentVariantMapper(ILocalizationService localizationService)
...
+            else
+            {
+                return context.MapEnumerable<ILanguage, Language>(allLanguages).ToList();
             }
-            return result;
+        }
+
+        private IEnumerable<string> GetSegments(IContent content)
+        {
+            return content.Properties.SelectMany(p => p.Values.Select(v => v.Segment)).Distinct();
```
Shazwazza: @PerplexDaniel  Just testing some stuff now, is it expected that this will return a `null` value as part of the result? or should `null` be filtered from here? I can't quite tell if that's expected or not, else this needs a WhereNotNull appended to it. I've merged and tested this in my temp branch and it works, i'll merge into dev/v8 once you let me know about this null.

PerplexDaniel: Yeah, segment `null` would be the "default" segment, sort of like the default language when a content type varies by language. In the above context the variants of a page are created based on the segments, and segment `null` should definitely be part of it as one of the variants will have segment `null` and another could have a segment `segment-b` or whatever alias is chosen. So when one segment is added the page would have at least 2 variants (one with segment `null` and another one with an actual segment alias).

I believe existing core code elsewhere will convert `null` to the empty string, and this is also how you can explicitly read the unsegmented property value in the front-end: `Model.Value("myProp", segment: "")`. Passing in `null` would fallback to the current variation context instead.

Shazwazza: Ok cool, i'll leave it as-is and add a note to it.

Shazwazza:
@PerplexDaniel that's all pulled in 🎉 thanks!

PR #7213, opened 2019-11-22
Segments in Backoffice: Client code

PerplexDaniel:
### Description
Proof of concept code to be able to display/edit segments in Umbraco backoffice. This PR is so you can guys can easily check the code changes through the diffs.

If you actually want to run this you would also need to merge #7212 for server side support,  configure a doctype to use segment variations, and insert a segmented property value with the content service. In addition, the splitview UI + variant dropdown will only show when culture variations are also enabled for the doctype.

To see it running it is much easier to clone [this repo](https://github.com/PerplexDaniel/Umbraco-CMS/tree/feature/backoffice-segment-support), follow the README and execute it.

---
_This item has been added to our backlog [AB#4165](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4165)_


src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontenteditors.directive.js
```diff
@@ -9,7 +9,7 @@
         bindings: {
             page: "<",
             content: "<", // TODO: Not sure if this should be = since we are changing the 'active' property of a variant
-            culture: "<",
+            variantId: "<",
```
Shazwazza: This could be considered a breaking change if some deeper components require this component and then access it's `culture` property. Currently I don't think anything does that in our core now though. Thoughts?

src/Umbraco.Web.UI.Client/src/common/directives/components/editor/umbeditorcontentheader.directive.js
```diff
@@ -15,36 +15,30 @@
...
-                    title += scope.content.name;
+
                 } else {
                     var name = editorState.current.contentTypeName;
                     scope.a11yMessage += " " + name;
                     scope.a11yName = name + " " + scope.a11yName;
-                    title += name;
                 }
-                
-                scope.$emit("$changeTitle", title);
```
Shazwazza: Is this a breaking change since there's no longer an event emitted? Was something relying on this?

src/Umbraco.Web.UI.Client/src/common/services/umbdataformatter.service.js
```diff
@@ -425,7 +426,15 @@
                         var variant = displayModel.variants[j];
 
                         _.each(invariantProperties, function (invProp) {
-                            variant.tabs[invProp.tabIndex].properties[invProp.propIndex] = invProp.property;
+                            var tab = variant.tabs[invProp.tabIndex];
+                            var prop = tab.properties[invProp.propIndex];
+
+                            if (prop.segment) {
+                                // Do not touch segmented properties
```
Shazwazza: Not sure I understand this check. If we have a non-culture + segmented property on a culture document then the segmented property would need to be the same reference across all cultures?

src/Umbraco.Web.UI.Client/src/common/services/varianthelper.service.js
```diff
@@ -0,0 +1,61 @@
...
+        } else if (!hasLanguage && hasSegment) {
+            // Segment only
+            return sep + variant.segment;
+        } else {
+            // Culture and Segment
+            return variant.language.culture + sep + variant.segment;
+        }
+    }
+
+    function getDisplayName(variant) {
```
Shazwazza: would be good to have some code comments on this one and hopefully unit tests

src/Umbraco.Web.UI.Client/src/common/services/varianthelper.service.js
```diff
@@ -0,0 +1,61 @@
...
+function variantHelper() {
+    /**
+     * Returns the id for this variant
+     * @param {any} variant
+     */
+    function getId(variant) {
+        var hasLanguage = variant.language && !!variant.language.culture;
+        var hasSegment = !!variant.segment;
+
+        var sep = ";";
```
Shazwazza: could be a const outside of this capture to avoid re-allocating strings

src/Umbraco.Web.UI.Client/src/common/services/varianthelper.service.js
```diff
@@ -0,0 +1,61 @@
...
+* @ngdoc service
+* @name umbraco.services.variantHelper
+* @description A helper service for dealing with variants
+**/
+function variantHelper() {
+    /**
+     * Returns the id for this variant
+     * @param {any} variant
+     */
+    function getId(variant) {
```
Shazwazza: Ideally some unit tests would be made for these methods.Angular services are the easiest to test so should be relatively easy to setup.

src/Umbraco.Web.UI.Client/src/common/services/varianthelper.service.js
```diff
@@ -0,0 +1,61 @@
+﻿/**
+* @ngdoc service
+* @name umbraco.services.variantHelper
+* @description A helper service for dealing with variants
+**/
+function variantHelper() {
```
Shazwazza: We are trying to keep away from naming new things `Helper` since these just become buckets of "stuff" that don't make any real sense. Naming is hard but if you can think of a better service name we should try to not use Helper. Maybe

* `contentVariantIdentifier`
* `contentVariantUtility` (I know, not much better than helper ;) 
etc.. ?

src/Umbraco.Web.UI.Client/src/views/components/content/edit.html
```diff
@@ -9,7 +9,7 @@
             <umb-variant-content-editors
                 page="page"
                 content="content"
-                culture="culture"
+                variant-id="culture"
```
Shazwazza: Is it meant to be only passing in culture for the variant-id ?

src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-content-header.html
```diff
@@ -40,18 +40,18 @@
                     </ng-form>
 
                     <a ng-if="content.variants.length > 0 && hideChangeVariant !== true" class="umb-variant-switcher__toggle" href="" ng-click="vm.dropdownOpen = !vm.dropdownOpen" ng-class="{'--error': vm.errorsOnOtherVariants}">
-                        <span>{{vm.currentVariant.language.name}}</span>
+                        <span ng-bind="getVariantDisplayName(vm.currentVariant)"></span>
```
Shazwazza: we try not to use methods in angular binding where possible because it's not good for performance since angular will call this method on every digest. Not sure we can avoid it since if we one-way bind it we'll have to handle refreshing it somehow, but maybe thats a possibility?

src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-content-header.html
```diff
@@ -40,18 +40,18 @@
...
-
+                    
                     <span ng-if="hideChangeVariant" class="umb-variant-switcher__toggle">
-                        <span>{{vm.currentVariant.language.name}}</span>
+                        <span ng-bind="getVariantDisplayName(variant)"></span>
                     </span>
 
                     <umb-dropdown ng-if="vm.dropdownOpen" style="min-width: 100%; max-height: 250px; overflow-y: auto; margin-top: 5px;" on-close="vm.dropdownOpen = false" umb-keyboard-list>
-                        <umb-dropdown-item class="umb-variant-switcher__item" ng-class="{'--current': variant.active, '--not-allowed': variantIsOpen(variant.language.culture), '--error': variantHasError(variant.language.culture)}" ng-repeat="variant in content.variants">
+                        <umb-dropdown-item class="umb-variant-switcher__item" ng-class="{'--current': variant.active, '--not-allowed': variantIsOpen(variant), '--error': variantHasError(variant.language.culture)}" ng-repeat="variant in content.variants">
```
Shazwazza: variantHasError is still only checking culture, have you dealt with validation for segments at this stage?

src/Umbraco.Web.UI.Client/src/views/content/overlays/publish.controller.js
```diff
@@ -12,6 +12,8 @@
         vm.dirtyVariantFilter = dirtyVariantFilter;
         vm.pristineVariantFilter = pristineVariantFilter;
 
+        $scope.getVariantDisplayName = variantHelper.getDisplayName;
```
Shazwazza: This controller is configured to use `vm` so this should be assigned to vm not $scope.

src/Umbraco.Web.UI.Client/src/views/content/overlays/publish.html
```diff
@@ -20,8 +20,8 @@
                                   model="variant.publish"
                                   on-change="vm.changeSelection(variant)"
                                   disabled="(variant.canPublish === false)"
-                                  server-validation-field="{{variant.htmlId}}" 
-                                  text="{{ variant.language.name }}"
+                                  server-validation-field="{{variant.htmlId}}"
+                                  text="{{getVariantDisplayName(variant)}}"
```
Shazwazza: should be `vm.getVariantDisplayName` once the js part is updated.

src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontenteditors.directive.js
PerplexDaniel: Actually for this PoC I hijacked the `culture` property and turned it into `<culture>SEMICOLON<segment>` and renamed it to `variantId`. It could for example be "en-US;segment-1" . However, this is just a hack and `culture` and `segment` should likely be passed in separately in production code, also to be able to update the querystring properly (`cculture` AND `csegment`). 

Most of these client changes were really just to be able to test the server changes and see if we can actually edit segmented property values. I should have described that more clearly in the description of this PR.

Shazwazza: Ok so are you saying that this should be changed from PoC code to production code? I can't really merge this until it's production ready, or is the intention that this PR is not to be merged?

PerplexDaniel: Yes, definitely. When we (me + Jeffrey) were discussing our changes with Claus he requested to submit PRs so you guys could have a look, but this front-end Umbraco stuff is not complete and certainly not production quality, so this is not an actual "request to pull" in that sense.

The server side code is intended to be merged, but this branch is really just the quickest way to make the front-end capable of editing segmented values to get a sense of it all. In my mind your front end guys would think about how everything is supposed to work UX wise and probably do these changes, but if not we probably should do another pass here though that will be at a later moment, probably not this month.

Shazwazza:
I have tested this locally with the server side stuff to confirm things do work. I will close this PR though, it won't be merged (as noted)

PR #7216, opened 2019-11-24
V8: #7209 fixes filter in User Management not keeping state

### LINKED ISSUE #7209: Filter in User Management not keeping state
Under the section User Management, the Status filter is not keeping its state when going back to list.

## Umbraco version

I am seeing this issue on Umbraco version: 7.15.2

Reproduction
------------

### Bug summary

filter-bug-7152

When i first do the filtering, the "GetPagedUsers" endpoint is called like this: `/umbraco/backoffice/UmbracoApi/Users/GetPagedUsers?pageNumber=1&pageSize=25&orderBy=Name&orderDirection=Ascending&filter=&userStates=Active`

When i come back from to the list from a user, the call look lige this: `/umbraco/backoffice/UmbracoApi/Users/GetPagedUsers?pageNumber=1&pageSize=25&orderBy=Name&orderDirection=Ascending&filter=`

So this means that `&userStates=Active` is missing in the 2nd call.

AndyButland:
Fixes #7209.

Currently when you apply a filter or sort to the user management screen, click to view a user's detail and then click to return to the list of users, the state of the filter/sort isn't saved, and you see the default view.

In this PR I've saved these selections to `$rootScope` and, if present, restored the user view filter and sort from here when returning to the list of users.  `$rootScope` isn't something to overuse of course as it's effectively a global variable, but it seems an appropriate use case here.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/users/views/users/users.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/users/views/users/users.controller.js
@@ -1,13 +1,12 @@
 (function () {
     "use strict";
 
-    function UsersController($scope, $timeout, $location, $routeParams, usersResource, 
-        userGroupsResource, userService, localizationService, contentEditingHelper, 
-        usersHelper, formHelper, notificationsService, dateHelper, editorService, 
+    function UsersController($scope, $timeout, $location, $routeParams, $rootScope, usersResource,
+        userGroupsResource, userService, localizationService,
+        usersHelper, formHelper, dateHelper, editorService,
         listViewHelper) {
 
         var vm = this;
-        var localizeSaving = localizationService.localize("general_saving");
 
         vm.page = {};
         vm.users = [];
@@ -128,8 +127,7 @@
 
         function init() {
 
-            vm.usersOptions.orderBy = "Name";
-            vm.usersOptions.orderDirection = "Ascending";
+            initViewOptions();
 
             if ($routeParams.create) {
                 setUsersViewState("createUser");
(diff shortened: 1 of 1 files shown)


kjac:
@AndyButland I did something similar for content list views, but using the querystring. Might be worth looking at as inspiration?

https://github.com/umbraco/Umbraco-CMS/pull/6189

AndyButland:
Thanks @kjac - I'll take a look.  Agree it might be nicer if this was reflected in the URL (would allow sharing of direct links to particular filters/sorts).

kjac:
Well it ain't a do or die kinda thing, but at least for consistency it's worth digging into. Alternatively redoing the content list view sorting to match the user one, whatever works 😀

AndyButland:
Have updated to use the querystring to retain state, as per discussion above.

Shazwazza:
I'm late to the party here but mculture should be automatically retained when changing routes with our magic route handling that re-adds it. Maybe there's an edge case here that it's not working?

kjac:
@Shazwazza good point, let me have a look at the actual current implementation and see what's going on since it works.

kjac:
@AndyButland ok so the culprit seems to be that you're now using `$location.url()` instead of `$location.path()` in `goToUser()`. This apparently bypasses the magic route handling.

A quick PoC shows that this works:

```js
$location.path("/users/users/user/" + user.id)
  .search("orderBy", vm.usersOptions.orderBy)
  .search("orderDirection", vm.usersOptions.orderDirection)
  .search("pageNumber", vm.usersOptions.pageNumber)
  .search("userStates", vm.usersOptions.userStates.join(","))
  .search("userGroups", vm.usersOptions.userGroups.join(","))
  .search("create", null).search("invite", null);
```

...when inserted instead of: 

```js
$location.url(pathToUser(user)).search("create", null).search("invite", null);
```

Of course my pretty PoC should probably be refactored a fair bit using some of the helpers you have introduced.

AndyButland:
Thanks for diagnosing that @kjac - I've reworked to use that method.  Think I changed it originally as I was having trouble with this `mculture` getting added and corrupting the querystring, but must have been as I was trying to initiate the change of route by applying the querystring as a string, rather than as you've shown above.

kjac:
Merged 😄 thanks again, @AndyButland

AndyButland:
You're welcome.  Curious though - what's the intention of `v8/contrib` branch?

kjac:
@AndyButland it's the future, my good man 😆 

It's supposed to be the base of future contribs. The good people at HQ are working out some docs and defaults, I believe.

PR #7217, opened 2019-11-25
V8: Add indicators to required properties within nested content

### LINKED ISSUE #7191: Mandatory field indication missing from Nested Content
### Bug summary
When a field within nested content is set as mandatory, the red star to indicate it as mandatory is missing.

The correct error message is shown when you save the page, but client-side validation is not performed as it is when the field is outside of nested content.

## Umbraco version
I am seeing this issue on Umbraco version: 8.3.0

### Setup
------------
Create a document type with a textstring.
Set the textstring 'Field is mandatory' to true.
Set the document type 'Is an Element type' to true.
Save the document type.

Create a new data type.
Set 'Nested Content' as the property editor.
Set the Element type to the newly created document type (and set the Group accordingly).
Save the data type.

In a document type (one that can be created as a node)
Add a new property.
Set the property name.
Select the newly created data type from the editor list.
Click on submit.
Save the document type.
------------

### Steps to reproduce
Add an instance of the document type above.
Add a new instance of the nested content property.

### Expected result
The mandatory field should have a red star indicating it is mandatory.

### Actual result
The mandatory field shows no indication of being mandatory.
image

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7191

### Description

This PR adds asterisks to required properties within nested content, as discussed in #7191. Note that this PR does _not_ implement clientside validation of these properties.

It looks like this:

#### Simple first-level required properties

image

#### Nested nested content with mixed required and non-required

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-nested-content.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-nested-content.less
@@ -14,6 +14,17 @@
     pointer-events: none;
 }
 
+.umb-nested-content--mandatory {
+    /*
+        yeah so this is a pain, but we must be super specific in targeting the mandatory property labels,
+        otherwise all properties within a reqired, nested, nested content property will all appear mandatory
+    */
+    > ng-form > .control-group > .umb-el-wrap > .control-header label:after {
+        content: '*';
+        color: @red;
+    }
+}
+
 .umb-nested-content-overlay {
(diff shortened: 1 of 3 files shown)


nielslyngsoe:
Super cool to get this indication in here, a small step making it easier for editors 👏

colin-tactuum:
Any idea when this will be included in an Umbraco release?

PR #7220, opened 2019-11-25
Remove membership providers from Core

Shazwazza:
https://github.com/umbraco/Umbraco-CMS/pull/7205 *must* be reviewed and merged first.

* Removing all refs of membership providers (System.Web.Security) from Umbraco.Core
* Removing the entire user membership provider
* Start prep work for removing members membership provider including removing support for old legacy things, custom membership providers, password question/answer, non-hashed passwords, etc... 
* Abstracting some required password security mechanisms

The membership provider is still in use for members which only operates within Umbraco.Web and is still used for both front-end and back office member editing. The next phase of removing membership providers will be to see what is involved in not referencing membership providers in the back office which will require some simple abstractions. (will be the next PR). The harder challenge is to replace front-end memberhip with asp.net identity (borrowing from UmbracoIdentity package).

## Testing

* Users and members should continue to work as per normal, creating, modifying, changing passwords for both
* Member front-end functionality should work - you can create partial view macros for login/register/info and test on the front-end
* New installs should work


src/Umbraco.Configuration/ConfigsFactory.cs
```diff
@@ -25,10 +25,26 @@ public Configs Create()
...
             return configs;
         }
     }
+
+    // Default/static user password configs
+    // TODO: Make this configurable somewhere - we've removed membership providers, so could be a section in the umbracosettings.config file?
+    // keeping in mind that we will also be removing the members membership provider so there will be 2x the same/similar configuration
+    internal class DefaultUserPasswordConfig : IUserPasswordConfiguration
+    {
+        public int RequiredLength => 12;
```
bergmania: Any reason to increase this from 10 to 12?

src/Umbraco.Web/Security/ActiveDirectoryBackOfficeUserPasswordChecker.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Umbraco.Web.Security
 {
-    // TODO: This relies on an assembly that is not .NET Standard :(
+    // TODO: This relies on an assembly that is not .NET Standard (at least not at the time of implementation) :(
```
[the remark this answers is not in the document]
Shazwazza: All good, i don't really think this belongs in Core anyways and should just be a package

src/Umbraco.Configuration/ConfigsFactory.cs
Shazwazza: haha, nope, just thought it was 12 off the top of my head, we can change when we implement the actual config.

PR #7221, opened 2019-11-25
Thread safety in RuntimeState.EnsureApplicationUrl

stevemegson:
`RuntimeState.EnsureApplicationUrl` uses a `HashSet<string>` to store previously-seen URLs, which is not thread-safe. It is possible to corrupt the HashSet in a way that causes calls to `Contains` to enter an infinite loop (I think it creates a loop in a linked list). Since `EnsureApplicationUrl` is called in `BeginRequest`, this breaks everything.

Since there is no `ConcurrentHashSet`, this PR uses a `ConcurrentDictionary<string, byte>`.

### DIFF
--- a/src/Umbraco.Core/Collections/ConcurrentHashSet.cs
+++ b/src/Umbraco.Core/Collections/ConcurrentHashSet.cs
@@ -70,7 +70,23 @@ namespace Umbraco.Core.Collections
         /// The number of elements contained in the <see cref="T:System.Collections.ICollection"/>.
         /// </returns>
         /// <filterpriority>2</filterpriority>
-        public int Count => GetThreadSafeClone().Count;
+        public int Count
+        {
+            get
+            {
+                try
+                {
+                    _instanceLocker.EnterReadLock();
+                    return _innerSet.Count;
+                }
+                finally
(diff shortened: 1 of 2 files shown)


Shazwazza:
We have a `Umbraco.Core.Collections.ConcurrentHashSet` in the core that you can use

stevemegson:
@Shazwazza thanks, I didn't know that was there.

While I'm sure it's unlikely to affect performance in practice here, does `Contains` really need to take a write lock and clone the inner HashSet? Would this not do the job?

```c#
public bool Contains(T item)
{
    try
    {
        _instanceLocker.EnterReadLock();
        return _innerSet.Contains(item);
    }
    finally
    {
        if (_instanceLocker.IsReadLockHeld)
            _instanceLocker.ExitReadLock();
    }
}
```

Shazwazza:
@stevemegson yep i haven't looked at this code for ages and don't fully recall the history there but you are correct, there's certainly no need to clone to get a contains (or a count), feel free to make those updates as part of this pr.

Shazwazza:
Looks good! thanks @stevemegson !

Issue #7219, opened 2019-11-25
Add attribute based route hijacking support


mattbrailsford:
Generally in MVC the Controller is meant to orchestrate functionality within a specific topic with the controller taking the topic name, ie `OrderController`. Within that controller it might expose different actions / views related to that topic.

In Umbraco, we have the concept of route hijacking to allow us to override the default route controller and have it perform some custom logic. To do this, we create a controller named after the doc type, for example a doc type of MyPage would require a controller of `MyPageController` inheriting from `RenderMvcController` or `IRenderController`. 

Whilst this works, it does however break the MVC convention as it forces a controller to be about a single function, but you may have multiple pages related to the same topic that you would like to group within the single controller. This currently isn't possible.

To solve this, I propose we add the ability to define hijacked routes via attribute (much in the same way in MVC you can define routes via attributes too). An example of this syntax could be as follows:

````csharp
public class MyController : SurfaceController, IRenderController
{
    // Hijack route for doc type MyPage
    [HijackRoute("MyPage")]
    public ActionResult MyPage(ContentModel model)
    {
        // Do some stuff here, then return the base method
        return View("MyPage", model);
    }
    
    // Hijack route for doc type MyPage using template MyTemplate
    [HijackRoute("MyPage", "MyTemplate")]
    public ActionResult MyOtherPage(ContentModel model)
    {
        // Do some stuff here, then return the base method
        return View("MyOtherPage", model);
    }
}
````

By implementing a `HijackRouteAttribute` we can now name our controller whatever we want and have it hijack multiple content types, grouping the logic together.

umbrabot:
Hiya @mattbrailsford,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #7223, opened 2019-11-25
Implemented route hijacking via attributes

mattbrailsford:
PR for issue #7219 adding route hijacking support using attributes.

I don't think my approach is that optimal, but I don't really fully understand all the moving parts so I just implemented it as best as I could. Please let me know if you know of any better approach.

To test, create a basic install using standard starter kit, then in `App_Code` folder create a `MyPeopleController` containing the following:

````csharp
using System;
using System.Web.Mvc;

namespace Demo
{
    public class MyPeopleController : Umbraco.Web.Mvc.RenderMvcController
    {
        [Umbraco.Web.Mvc.HijackRoute("People")]
        public ActionResult Index2(Umbraco.Web.Models.ContentModel model)
        {
            // Do some stuff here, the return the base Index method
            return View("People", model);
        }
    }
}
````

If everything works, navigating to `/people/` on the front end should hit your controller action (even though the controller name no longer matches the name convention).

---
_This item has been added to our backlog [AB#4649](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4649)_

### DIFF
--- a/src/Umbraco.Web/Mvc/HijackRouteAttribute.cs
+++ b/src/Umbraco.Web/Mvc/HijackRouteAttribute.cs
@@ -1,0 +1,28 @@
+﻿using System;
+
+namespace Umbraco.Web.Mvc
+{
+    /// <summary>
+    /// Used to define a controller action as being
+    /// an entry point for a hijacked route
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    public class HijackRouteAttribute : Attribute
+    {
+        public string ContentTypeAlias { get; set; }
+        public string TemplateAlias { get; set; }
+
+        public HijackRouteAttribute(string contentTypeAlias)
+        {
+            ContentTypeAlias = contentTypeAlias;
+            TemplateAlias = null;
+        }
+
+        public HijackRouteAttribute(string contentTypeAlias, string templateAlias)
+        {
+            ContentTypeAlias = contentTypeAlias;
+            TemplateAlias = templateAlias;
+        }
+    }
+
+}
--- a/src/Umbraco.Web/Mvc/RenderControllerFactory.cs
+++ b/src/Umbraco.Web/Mvc/RenderControllerFactory.cs
@@ -1,4 +1,5 @@
-﻿using System.Web.Mvc;
+﻿using System.Collections.Generic;
+using System.Web.Mvc;
 using System.Web.Routing;
 
 namespace Umbraco.Web.Mvc
@@ -10,6 +11,10 @@ namespace Umbraco.Web.Mvc
     /// <remarks></remarks>
     public class RenderControllerFactory : UmbracoControllerFactory
     {
+        public RenderControllerFactory(IEnumerable<IRenderController> renderControllers)
+            : base(renderControllers)
+        { }
+
         /// <summary>
         /// Determines whether this instance can handle the specified request.
         /// </summary>
--- a/src/Umbraco.Web/Mvc/RenderRouteHandler.cs
+++ b/src/Umbraco.Web/Mvc/RenderRouteHandler.cs
@@ -13,6 +13,7 @@ using Umbraco.Web.Routing;
 using System.Collections.Generic;
 using Current = Umbraco.Web.Composing.Current;
 using Umbraco.Web.Features;
+using System.Reflection;
 
 namespace Umbraco.Web.Mvc
 {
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by emmaburstow** [comment]
Hey mb, thanks for the work. I'll take a look and let you know if I have any questions but from here, all looks well :)

Em

File: src/Umbraco.Web/Mvc/RenderControllerFactory.cs [hunk 1]
```diff
@@ -10,6 +11,10 @@ namespace Umbraco.Web.Mvc
     /// <remarks></remarks>
     public class RenderControllerFactory : UmbracoControllerFactory
     {
+        public RenderControllerFactory(IEnumerable<IRenderController> renderControllers)
```
Shazwazza: This will be a breaking change, you'll prob have to create an empty obsolete ctor that passes in this collection from the DI Current.Factory

src/Umbraco.Web/Mvc/RenderRouteHandler.cs
```diff
@@ -279,10 +280,40 @@ internal virtual RouteDefinition GetUmbracoRouteDefinition(RequestContext reques
...
                     def.ControllerType = controllerType;
                     def.ControllerName = ControllerExtensions.GetControllerName(controllerType);
+
                     if (def.ControllerName != defaultControllerName)
                     {
                         def.HasHijackedRoute = true;
                     }
+
+                    // Look for hijacked action
+                    var hijackActions = controllerType
```
Shazwazza: We will need to perf optimize this before release we don't want all this reflection happening on each request. 
//cc @nul800sebastiaan + @clausjensen - we can add a task for this to a sprint.

src/Umbraco.Web/Mvc/RenderRouteHandler.cs
```diff
@@ -279,10 +280,40 @@ internal virtual RouteDefinition GetUmbracoRouteDefinition(RequestContext reques
...
+                        .GetMethods(BindingFlags.Public | BindingFlags.Instance)
+                        .Select(x => Tuple.Create(x, x.GetCustomAttribute<HijackRouteAttribute>(true)))
```
Shazwazza: The `true` parameter here will mean that it will look into base classes/inherited structures which i really don't think you would want in this case? This will also add performance overhead.

src/Umbraco.Web/Mvc/RenderRouteHandler.cs
```diff
@@ -279,10 +280,40 @@ internal virtual RouteDefinition GetUmbracoRouteDefinition(RequestContext reques
...
+                        .ToList();
+
+                    if (hijackActions.Count > 0)
+                    {
+                        MethodInfo hijackAction = null;
+
+                        if (request.HasTemplate)
+                        {
+                            hijackAction = hijackActions
+                                .FirstOrDefault(x => x.Item2
```
Shazwazza: Since it will be possible to add the same attributes with the same template alias/action name to multiple actions, using FirstOrDefault here will mean developers don't really know which one would execute. 
Do we need to consider what should happen in this scenario? Is there a case where a developer would want/need multiple attributes with the same values? Else should an exception throw to inform a developer that they have duplicates?

src/Umbraco.Web/Mvc/UmbracoControllerFactory.cs
```diff
@@ -13,12 +16,27 @@ namespace Umbraco.Web.Mvc
     public abstract class UmbracoControllerFactory : IFilteredControllerFactory
     {
         private readonly OverridenDefaultControllerFactory _innerFactory = new OverridenDefaultControllerFactory();
+        private readonly IEnumerable<IRenderController> _renderControllers;
+
+        public UmbracoControllerFactory(IEnumerable<IRenderController> renderControllers)
```
Shazwazza: Breaking change (see prev note)

src/Umbraco.Web/Mvc/UmbracoControllerFactory.cs
```diff
@@ -13,12 +16,27 @@ namespace Umbraco.Web.Mvc
...
 
         public abstract bool CanHandle(RequestContext request);
 
         public virtual Type GetControllerType(RequestContext requestContext, string controllerName)
         {
-            return _innerFactory.GetControllerType(requestContext, controllerName);
+            var controllerType = _innerFactory.GetControllerType(requestContext, controllerName);
+            if (controllerType == null)
+            {
+                controllerType = _renderControllers
```
Shazwazza: I'm unsure about the addition of this code, is it necessary? We didn't require this code before for route hijacking so not sure why we require it now.

src/Umbraco.Web/Mvc/UmbracoControllerFactory.cs
```diff
@@ -13,12 +16,27 @@ namespace Umbraco.Web.Mvc
...
+                    .Select(x => x.GetType())
+                    .FirstOrDefault(x => x.GetMethods(BindingFlags.Public | BindingFlags.Instance)
```
Shazwazza: If we need to keep this code it will need to be perf optimized before release, we can't have this reflection running on each request.
//cc @nul800sebastiaan + @clausjensen - we can add a task for this to a sprint.

src/Umbraco.Web/Mvc/HijackRouteAttribute.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+
+namespace Umbraco.Web.Mvc
+{
+    /// <summary>
+    /// Used to define a controller action as being
+    /// an entry point for a hijacked route
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    public class HijackRouteAttribute : Attribute
```
Shazwazza: Attribute classes should be sealed

src/Umbraco.Web/Mvc/RenderRouteHandler.cs
mattbrailsford: Yea, I think we can remove it and if someone wants to inherit a base controller they will have to override the child action and add the attribute on the override 👍

src/Umbraco.Web/Mvc/UmbracoControllerFactory.cs
mattbrailsford: It's needed because the controllers are only globally locatable via their actual names, but we need a way to say "if one isn't found, also look for controllers with the relevant attribute". From what I could find, the looking up all happens inside MS core code, so the only place I could think to add the additional lookup was here, using controllers in DI. 

This code is locating the relevant controller, then the other code `RenderRouteHandler` digs deeper to look for the relevant action.

PR #7224, opened 2019-11-26
V8: Update the descriptions of the Nested Content configuration options

### LINKED ISSUE #7218: Nested Content configuration wording
We should stop saying Document Types and start using Element Type, and thereby it shouldn't be needed to say only Element Types. Also, the word 'Blueprint' could be misleading, I think it's better without it.

Feel welcome to provide a new formulation for this Description.

image

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7218

### Description

This PR updates the descriptions (and one label) of the Nested Content configuration options, hopefully making them a bit clearer (and all in present tense)

These are the configuration options as-is:

image

And here are the configuration options with this PR applied. I have marked the changed ones:

image


nielslyngsoe:
I cut a few words off to make it even simpler, Thanks!

PR #7225, opened 2019-11-26
V8: Filter itempicker items explicitly by their names

kjac:
### Prerequisites

### Description

The itempicker overlay filtering has a weird behavior: It always returns all items when filtering for "Icon":

image

It happens because all items have an icon property with a textual value that contains "icon".

This is particularly annoying if you have a lot of element types and you're trying to find the "Icon list". And if you have 12 or more element types available in a Nested Content configuration, the editors will also experience this, as the filtering option is enabled at 12+ element types when editing Nested Content.

This PR ensures that the filtering is done explicitly on the item names, thus fixing this issue:

itempicker-filtering-after


emmaburstow:
👌

PR #7226, opened 2019-11-26
V8: Automatically open the itempicker when configuring Nested Content

kjac:
### Prerequisites

### Description

When you add new element types to the Nested Config configuration, you _first_ have to click "Add" and _then_ click "Add element type": 

nested-content-add-before

It makes sense that you have to click "Add element type" for the first (default) element in the list of element types, but I can't think of any reason why "Add element type" isn't invoked automatically when adding additional element types. So this PR fixes just that:

nested-content-add-after


kjac:
Updated with latest v8/dev

PR #7228, opened 2019-11-26
V8 - Community package tracking of references

warrenbuckley:
This is a WIP/draft PR to allow the following scenario:

> A good example of this would be something like DTGE, if this existed in v8 but the developers on this project didn't want to update their package to support the new `IDataValueReference`, I should be able to create my own implementation of `IDataValueReference` and assign this to be used for the existing DTGE.


src/Umbraco.Core/PropertyEditors/DataValueReferenceCollection.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections.Generic;
+using Umbraco.Core.Composing;
+
+namespace Umbraco.Core.PropertyEditors
+{
+    public class DataValueReferenceCollection : BuilderCollectionBase<IDataValueReference>
```
Shazwazza: Like i mentioned in the spec and on Slack - this collection does not track `IDataValueReference`, we need a new `IDataReferenceValueFactory` (or whatever you want to call it). `IDataValueReference` is an interface that can be imlemented directly by property editors, `IDataValueReferenceFactory` is something that can create an instance of `IDataValueReference`. 

Prob best to re-read the entire spec again :P

warrenbuckley:
@Shazwazza I wasn't sure about the Factory approach. My brain couldn't get my head around it at the time, but your spec notes talk about us potentially tracking/adding all our core implementations to the collection so developers can pick & choose what they want to disable in their own composers hence its a more simpler approach of tracking the `IDataValueReferences` directly.

This is still a WIP/draft but I would :heart: for your input/early review before I spend anymore time on this, to ensure I have gone about this in the right way please.

warrenbuckley:
@Shazwazza I have done your suggestions as we talked about yesterday after standup, but the unit tests & LightInject do not seem to like me. Any ideas/thoughts as to what I have done wrong please mate?!

warrenbuckley:
Fixed the unit tests failing. Can you take a look at this for a review please @Shazwazza

warrenbuckley:
## Test Notes
* Verify code is implemented correctly
* Install the SK & verify the content/document & media relations are created as they were before
* Use the example below to add your own implementation to **ADD** to the reference list 

### Example for testing
```csharp
using System;
using System.Collections.Generic;
using Umbraco.Core;
using Umbraco.Core.Models;
using Umbraco.Core.Models.Editors;
using Umbraco.Core.PropertyEditors;

namespace Umbraco.Web.PropertyEditors
{
    public class DumbExample : IDataValueReferenceFor
    {
        public IDataValueReference GetDataValueReference() => new AddToContentPicker();

        public bool IsForEditor(IDataEditor dataEditor) => dataEditor.Alias.InvariantEquals(Constants.PropertyEditors.Aliases.ContentPicker);

        private class AddToContentPicker : IDataValueReference
        {
            public IEnumerable<UmbracoEntityReference> GetReferences(object value)
            {
                // Dumb example - always ADD a media reference to the collection list
                // When it's a ContentPicker datatype
                var references = new List<UmbracoEntityReference>();
                var udiType = ObjectTypes.GetUdiType(UmbracoObjectTypes.Media);
                var udi = Udi.Create(udiType, Guid.Parse("fbbaa38d-bd93-48b9-b1d5-724c46b6693e"));
                var entityRef = new UmbracoEntityReference(udi);
                references.Add(entityRef);
                return references;
            }
        }
    }
```

**Remember to add this class in some composer**
```csharp
composition.DataValueReferenceFors()
                .Append<DumbExample>();
```

src/Umbraco.Core/Runtime/CoreInitialComposer.cs
```diff
@@ -70,11 +70,16 @@ public override void Compose(Composition composition)
...
-            composition.WithCollectionBuilder<DataEditorCollectionBuilder>()
+            composition.DataEditors()
                 .Add(() => composition.TypeLoader.GetDataEditors());
             composition.RegisterUnique<PropertyEditorCollection>();
             composition.RegisterUnique<ParameterEditorCollection>();
 
+            // Used to determine if a datatype/editor should be storing/tracking
+            // references to media item/s
+            composition.DataValueReferenceFors();
+                //WB:TODO Try me out & overide/add to an existing list
```
Shazwazza: will need to remove comment

src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
```diff
@@ -835,6 +837,20 @@ protected void PersistRelations(TEntity entity)
                 var val = p.GetValue(); // get the invariant value
                 var refs = reference.GetReferences(val);
                 trackedRelations.AddRange(refs);
+
```
Shazwazza: This whole section needs to be dealt with differently. I would like to have all of the logic to return `IEnumerable<UmbracoEntityReference>` as tracked relations in a method... probably a method on top of `DataValueReferenceForCollection`. There's a problem with this current logic as you'll see above, it first does this check `if (!(valueEditor is IDataValueReference reference)) continue;` ... and if that doesn't succeed it then moves to the next property which means it entirely ignores any custom `IDataValueReferenceFor` implementations.

So ideally, all of this logic is handled with a call to a new method: `_dataValueReferenceFors.GetAllReferences(entity.Properties);`

warrenbuckley: OK the renames is done to `IDataValueReferenceFactory` and the example code to inherit from both interfaces from your second suggestion.

Why is the method on the Collection as opposed to the Interface/item in the collection @Shazwazza

Shazwazza: @warrenbuckley I don't really understand your question. What Needs to happen is that this whole highlighted code area: https://github.com/umbraco/Umbraco-CMS/pull/7228/files#diff-83e79f7101cd6797ca42ad8ed07f5835R828-R854 needs to be in a centralized method. A single method used to get all references from all properties for this content item. Not really sure what interface you are referring to here.

warrenbuckley:
Ready for a re-review & test please @Shazwazza 

## Updated Example

```csharp
using System;
using System.Collections.Generic;
using Umbraco.Core;
using Umbraco.Core.Models;
using Umbraco.Core.Models.Editors;
using Umbraco.Core.PropertyEditors;

namespace Umbraco.Web.PropertyEditors
{
    public class DumbExample : IDataValueReferenceFactory, IDataValueReference
    {
        public IDataValueReference GetDataValueReference() => this;

        public bool IsForEditor(IDataEditor dataEditor) => dataEditor.Alias.InvariantEquals(Constants.PropertyEditors.Aliases.ContentPicker);

        public IEnumerable<UmbracoEntityReference> GetReferences(object value)
        {
            // Dumb example - always ADD a media reference to the collection list
            // When it's a ContentPicker datatype
            var references = new List<UmbracoEntityReference>();
            var udiType = ObjectTypes.GetUdiType(UmbracoObjectTypes.Media);
            var udi = Udi.Create(udiType, Guid.Parse("fbbaa38d-bd93-48b9-b1d5-724c46b6693e"));
            var entityRef = new UmbracoEntityReference(udi);
            references.Add(entityRef);
            return references;
        }
    }
}
```

src/Umbraco.Core/PropertyEditors/DataValueReferenceFactoryCollection.cs
```diff
@@ -8,5 +10,43 @@ public class DataValueReferenceFactoryCollection : BuilderCollectionBase<IDataVa
...
         { }
+
+        public IEnumerable<UmbracoEntityReference> GetAllReferences(PropertyCollection properties, PropertyEditorCollection propertyEditors)
+        {
+            var trackedRelations = new List<UmbracoEntityReference>();
+
+            foreach (var p in properties)
+            {
+                // Injecting propertyEditorCollection is causing LightInject to throw a Recursive Dependancy error
+                // Hence using Current.PropertyEditors
```
Shazwazza: comment can be removed

warrenbuckley: Done in this commit 4d78a2c848d7399ddaa13257a7d13526205f5b31 @Shazwazza

PR #7229, opened 2019-11-26
Service worker for the backoffice

Matthew-Wise:
This PR allows the backoffice to run a service worker. 

It can be enabled using a new umbraco configuration settings:

```
<backOffice>
        <serviceWorker>            
            <domain>example.com</domain>
        </serviceWorker>
    </backOffice>
```

I am unclear on its effect when upgrading, as for install I had a look over the paths and see no issues.

A test using chrome's fast 3G simulation resulted in 0.5 - 1 second difference in load times.

**Testing notes**
Using localhost is fine if you want to use a custom domain it needs to have a valid ssl.

Thanks 
Matt


nul800sebastiaan:
For a bit of context: 

Background info as to why service workers are great: https://medium.com/@the_ozmic/improving-perceived-performance-images-on-the-web-and-more-7761d6847669

> **Caching with Service Workers.**
You can take advantage of service workers and make your web app/website incredibly more performant through the use of client-side caching. When a request is made by the client, using your service worker, you can intercept that request and serve a cached version of it instead if that request has been made before and your cache is not out-dated. This not only makes serving images and other assets seem instant, but you also save the user’s data because they make far fewer requests that actually get to the server.

However, a service worker runs for a full domain and you can only run one per domain. Which means you need to explicitly opt-in to running the service worker for the Umbraco backoffice by whitelisting it in your config. 

If you are already running a service worker for your front-end then you should NOT whitelist the backoffice one or you'll get conflicts.

Alternatively, you can choose to run the Umbraco backoffice on a subdomain and the frontend on the main domain. At that point you can whitelist the subdomain (`admin.domain.com` for example) to use the backoffice service worker.

Matthew-Wise:
With SVG icons been added and the service workers been mentioned. Is this PR something you would like fixing up?

PR #7232, opened 2019-11-26
V8: Disable Nested Content add button when max items is reached (regression)

kjac:
### Prerequisites

### Description

I found a small regression issue in the recent rewrite of Nested Content: The "add" button isn't disabled when the maximum allowed number of items is reached.

With this PR the "add" button goes dimmed (disabled) when you're no longer allowed to add anymore items:

image


emmaburstow:
👌 I really like this one

kjac:
PR updated with latest v8/dev

PezCo:
While on this may i suggest adding a check for single item mode too? vm.singleMode is available possibly on line 39 to hide the whole footer?

nathanwoulfe:
@kjac this appears to be fixed on latest v8/contrib - given that, I'm hitting the bit red button.
image

Issue #6758, opened 2019-10-18
Calling RemoveTreeNodeLocked removes the root node of the current gen


clausjensen:
Follow up for: https://github.com/umbraco/Umbraco-CMS/issues/6698

>What I find a bit strange is that in the ContentStore.SetBranch method, it first clears the branch with ClearBranchLocked which is fine, this essentially zero's out the new Gen for the branch since it's going to fill in these null values after that ... but it then calls RemoveTreeNodeLocked which is going to actually remove the root node of this particular branch for the current Gen (not the new one that is being created) which seems odd since this branch shouldn't be modified for the current Gen, only for the new Gen but this effectively makes this branch disapear for the current Gen. 🤔 I'm still investigating what changes are needed to fix this correctly.

---
_This item has been added to our backlog [AB#3252](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3252)_

nul800sebastiaan:
Was fixed in #7233

PR #7233, opened 2019-11-27
When nucache updates are made the previous Gen root node is cleared

Shazwazza:
Fixing https://github.com/umbraco/Umbraco-CMS/issues/6758

## Testing notes

see comment https://github.com/umbraco/Umbraco-CMS/pull/7233#issuecomment-573523237

## Original note:

As noted in the issue above - there's an odd thing happening when we refresh a branch in nucache, it will remove the root node that is being updated from the branch for the previous gen. This is problematic on a high traffic site in the case that there are active requests being served while a change is made to the same content being served. It would be a timing issue but could affect some requests not being able to find the content being attempted to route.

... Still investigating but I've created a unit test to show that the hiearchy values are being zero'd out which shouldn't happen since you should be able to request a previous gen's state which should be 'frozen' but this is not the case.


Shazwazza:
The issue is these lines of code:

https://github.com/umbraco/Umbraco-CMS/blame/v8/dev/src/Umbraco.Web/PublishedCache/NuCache/ContentStore.cs#L941

They are changing the values of an already existing content item for the current generation but the current generation should never be modified. Previous to this change this code wasn't there.

The changes were from several months ago when we were fixing perf issues with new cache, I started fixing the initial loading issue one way and Stephane had told me he had previously worked on a double linked list implementation which would be the next perf fix. That code lives here and this is where that change was introduced https://github.com/umbraco/Umbraco-CMS/commit/e7e40b20b5f7d007fcd37fc98d0397350d4bad88#diff-1983717019d1fd1da9fa80f10ab00b38R787. I think this change was accidental and without me fully undertanding the impact of it I had merged it into our current code.

Turns out, just removing these 2x lines of code works and all tests pass. 

Have just pushed a change with notes and we'll see if all tests pass.

bergmania:
Changes looks good and make sense.. Nice with the unit test

PR #7235, opened 2019-11-27
V8: Display published state of mixed variance correctly in multi URL picker

kjac:
### Prerequisites

### Description

If a multi URL picker on culture variant content contains culture *in*variant items, these always list as being in an unpublished state - despite them actually being published:

image

This is due to a slightly wrong assumption in the property value editor. This PR fixes that assumption; once applied the published state is correct for invariant items as well:

image

#### Steps to test

1. Create culture variant content that contains a multi URL picker.
2. Pick published items of both culture variant and culture invariant types.
3. Save the content.
4. Reload the browser. Verify that the published states of the picked items are still listed correctly.
5. Pick some unpublished content as well, and verify that it still lists as being unpublished in the multi URL picker.


emmaburstow:
👌

emmaburstow:
👌

PR #7250, opened 2019-11-29
V8: 7249 - pointer events on data type label

### LINKED ISSUE #7249: Clicking a doctype's property data type doesn't open the editor for the property
In a fresh install of 8.4RC. in the latest Chrome on Windows 10 pro, when I click on the data type of a property within the doctype editor I would expect to open the editor (as worked previously in v7 and v8). in 8.4RC clicking the data type (as shown in the attached screenshot) does nothing.

I now have to click on the cog icon to open the editor for a property.

image

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #7249

### Description
Added pointer-events:none to the inner element so that it is never targeted by the click event. Added a bit of structure to the SASS file, to nest modifier classes rather than repeating the full selector.

Verify: go to settings, doc types, click the data type label for a property, infinite editor will open (provided the property is not inherited/composed)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-group-builder.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-group-builder.less
@@ -1,133 +1,134 @@
 /* ---------- GROUPS ----------  */
 
 .umb-group-builder__groups {
-  list-style: none;
-  margin: 0;
-  padding: 0;
+    list-style: none;
+    margin: 0;
+    padding: 0;
 }
 
 .umb-group-builder__group {
-  min-height: 86px;
-  border: 1px solid transparent;
-  border-radius: @baseBorderRadius;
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks @nathanwoulfe - why can't we make the inner element actually clickable though?

nathanwoulfe:
We could, by binding another handler, but by using the pointer-event attribute instead, we can avoid that and let the click pass through to the parent div to fire the click event and open the editor. 

So it's still clickable in the sense that the user can click on it and have the editor open, but the actual click handler is bound to the parent element.

nathanwoulfe:
@nul800sebastiaan we'd agreed to merge this, but looks like I've refreshed my repo in the 2 years since I opened it, so can't deal with the conflicts. I'm going to close this, can open again against v9.

PR #7251, opened 2019-11-29
7247 - restores missing doctype alias

### LINKED ISSUE #7247: 8.4RC - missing doctype alias, no way view to edit it after creation
In a fresh install of 8.4RC the doctype alias is missing

This is in the latest Chrome on Windows 10 Pro

The padlock shows (and you can click on it), but the alias does not show, so there's no way to edit it. See attached screenshot.

image

---
_This item has been added to our backlog [AB#4000](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4000)_

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #7247 

### Description

Issue here was the autoResize directive was watching model values, and calling an update on change, but that was happening before the updated value was bound so the resulting width was always zero. 

Change here adds a timeout in resize directive to ensure changed value has bound before calculating size. Have also removed a stack of Internet Explorer specific stuff, since IE is not supported in 8.

Readily validated by refreshing the edit doc type view - alias is visible and editable.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbautoresize.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbautoresize.directive.js
@@ -1,157 +1,66 @@
 angular.module("umbraco.directives")
-   .directive('umbAutoResize', function($timeout) {
-      return {
-         require: ["^?umbTabs", "ngModel"],
-         link: function(scope, element, attr, controllersArr) {
+    .directive('umbAutoResize', function($timeout) {
+        return {
+            require: ["^?umbTabs", "ngModel"],
+            link: function (scope, element, attr, controllersArr) {
 
-            var domEl = element[0];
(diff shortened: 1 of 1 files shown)


nielslyngsoe:
Hi @nathanwoulfe Thanks for your effort in making a fix for this issue. But we cannot remove the IE support since it also covers MS Edge. So I have ported your fix into a new PR with only the timeout correction.

https://github.com/umbraco/Umbraco-CMS/pull/7267

nathanwoulfe:
Thanks Niels - I'd figured the IE issue wouldn't exist in Edge anymore given it's built on Chromium. All good though, as long as the bug is fixed

nielslyngsoe:
@nathanwoulfe no worries, it was with good intention 😀  Sadly the Chromium version of Edge is first released at the start of 2020 😐

PR #7253, opened 2019-11-29
 8.4RC: MediaIndexPopulator should use IUmbracoContentIndex instead of Umbrac…

bielu:
Hi @Shazwazza ,
as I was doing changes in ContentPopulators I notice I didn't make a change for Media and fixing bug which will be introduced for custom examine providers as Media wouldn't be populated.
that issue is not affecting Umbraco with basic Examine.
Related PR:
https://github.com/umbraco/Umbraco-CMS/pull/6824

---
_This item has been added to our backlog [AB#4028](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4028)_


Shazwazza:
Thanks @bielu good catch, sorry I missed that one!

PR #7257, opened 2019-12-02
V8: Do not save content when opening the content template

kjac:
### Prerequisites

### Description

In an effort to make things more accessible, the link that used to open the current content template has been turned into a button (see #7094). As a result, the content editor now saves the current content when you open its template in infinite editing:

edit-template-prompt-save-before

If you're editing invariant content you probably won't even notice it being saved. That's not so good.

This issue is much along the lines of #7185, and the fix is the same: Applying an explicit type to the edit button (type="button"), thus explicitly instructing the browser how to interpret it.

edit-template-prompt-save-after


poornimanayar:
And merged :-)

PR #7260, opened 2019-12-02
Listview styling adjustments

nielslyngsoe:
This PR updates the looks of items in the subheader in List View, to look as following screenshot:

image

The reason for this is to keep making the looks of Umbraco simpler and friendlier by ensuring a calm look. Notice how the search input is collapsed until its activated, in this way we make it easier for users to decode the view.

Any thoughts and considerations are welcome.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/belle.less
+++ b/src/Umbraco.Web.UI.Client/src/less/belle.less
@@ -132,6 +132,7 @@
 @import "components/umb-content-grid.less";
 @import "components/umb-contextmenu.less";
 @import "components/umb-layout-selector.less";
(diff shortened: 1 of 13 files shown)


bjarnef:
@nielslyngsoe it doesn't seem to be consistent that the search is collapsed in listview (content, media and member), but not in users section.

image

I think the search input (until focussed) looks a bit disabled and the "create" button as well, because of the grey background.

nielslyngsoe:
@bjarnef correct, we should get that one adjusted as well.

I have created a up-for-graps issue, so we can get that sorted in near future. :-)
https://github.com/umbraco/Umbraco-CMS/issues/7517

PR #7263, opened 2019-12-03
Removing more membership providers from being used

Shazwazza:
With this update, membership providers are no longer used for back office operations for members (already have been removed for users). So now membership provider functionality only remains for the front-end which will be easier to manage when we split the web projects into front-end and back end. This also means there's a lot less usages of System.Web.Security.

Eventually membership will need to be replaced with aspnet identity in the front-end which is fine but for now, the back office allows editing member data by talking directly to our services and security/password handling is dealt with by some abstractions.

This also removes all of the old legacy logic of having to deal with the huge amount of legacy membership provider settings and password changing. For example, all of this is removed:

* enablePasswordRetrieval
* requiresQuestionAnswer
* enableReset
* resetPassword
* generatedPassword / GeneratedPassword

And sorts out the simple usage of allowManuallyChangingPassword / AllowManuallyChangingPassword

Changing passwords  in the back office is now simple:

* You can change any user's password apart from your own if you have access to the user section
* You can change your own password if you supply your existing one
* You can change any members password if you have access to edit them

There's no more password resetting/generating via the back office for members. We could bring this back if we want to but that was more of a work around for legacy behavior. For now it's gone though which is fine for the time being.

You'll notice there's a new IMemberPasswordConfiguration and it's implementation is just a static config class just like Users. Eventually when we get aspnet identity in for members, we'll need to replace that implementation.

## Testing

* Make sure you can login
* Change your own password
* Create other users
* Login as them
* Change other users' passwords
* Create members in the back office, assign roles, change passwords
* Create 3x partial view macros for the front end - register member, login member, member status
   * Allow these macros to be rendered in an RTE
   * Put these macros in an RTE then test that you can register, login, logout on the front-end


src/Umbraco.Web/Editors/MemberController.cs
```diff
@@ -186,53 +191,31 @@ public MemberDisplay GetEmpty(string contentTypeAlias = null)
...
-                    generatedPassword = UpdateWithMembershipProvider(contentItem);
+                    UpdateMemberData(contentItem);
                     break;
                 case ContentSaveAction.SaveNew:
-                    MembershipCreateStatus status;
-                    CreateWithMembershipProvider(contentItem, out status);
-
-                    // save the ID of the creator
-                    contentItem.PersistedContent.CreatorId = Security.CurrentUser.Id;
+                    contentItem.PersistedContent = CreateMemberData(contentItem);
```
bergmania: There is something wrong going on here.. Before this `contentItem.PersistedContent.RawPasswordValue` contains the value inputted in the UI. After this line the value is empty.

The result is also that the password i Use when I create the member, cannot be used to sign in the member

Shazwazza: Thanks and good catch, have fixed this up and removed more membership provider stuff that I overlooked.

src/Umbraco.Web/Editors/AuthenticationController.cs
```diff
@@ -60,9 +60,9 @@ public AuthenticationController(IUserPasswordConfiguration passwordConfiguration
         /// </summary>
         /// <returns></returns>
         [WebApi.UmbracoAuthorize(requireApproval: false)]
-        public IDictionary<string, object> GetPasswordConfig()
+        public IDictionary<string, object> GetPasswordConfig(int userId)
         {
-            return _passwordConfiguration.GetConfiguration();
+            return _passwordConfiguration.GetConfiguration(userId != UmbracoContext.Security.CurrentUser.Id);
```
bergmania: I'm not a fan of this logic in a controller. It is business logic that we old allow `allowManuallyChangingPassword` for the current user..

src/Umbraco.Web/Models/ContentEditing/MemberSave.cs
```diff
@@ -27,16 +31,27 @@ public class MemberSave : ContentBaseSave<IMember>
...
+
+        /// <summary>
+        /// Returns the value from the IsLockedOut property
+        /// </summary>
+        public bool IsLockedOut => GetPropertyValue<bool>(Constants.Conventions.Member.IsLockedOut);
+
+        /// <summary>
+        /// Returns the value from the IsApproved property
+        /// </summary>
+        public bool IsApproved => GetPropertyValue<bool>(Constants.Conventions.Member.IsApproved);
```
bergmania: These are calculated each time. Could just be a simple getter and set the values in the default constructor 🙂 

Something like

```
...
        public MemberSave()
        {
            Comments = GetPropertyValue<string>(Constants.Conventions.Member.Comments);
            IsLockedOut = GetPropertyValue<bool>(Constants.Conventions.Member.IsLockedOut);
            IsApproved = GetPropertyValue<bool>(Constants.Conventions.Member.IsApproved);
        }
...
        /// <summary>
        /// Returns the value from the Comments property
        /// </summary>
        public string Comments { get; }

        /// <summary>
        /// Returns the value from the IsLockedOut property
        /// </summary>
        public bool IsLockedOut { get; }

        /// <summary>
        /// Returns the value from the IsApproved property
        /// </summary>
        public bool IsApproved { get; }
...
```

src/Umbraco.Web/Models/ContentEditing/MemberSave.cs
```diff
@@ -27,16 +31,27 @@ public class MemberSave : ContentBaseSave<IMember>
...
-
-        [DataMember(Name = "isApproved")]
-        public bool IsApproved { get; set; }
-        
-
-        // TODO: Need to add question / answer support
+        /// <summary>
+        /// Returns the value from the Comments property
+        /// </summary>
+        public string Comments => GetPropertyValue<string>(Constants.Conventions.Member.Comments);
```
bergmania: Are it intended the DataMemberAttributes is removed?

src/Umbraco.Web/Editors/MemberController.cs
```diff
@@ -496,111 +354,48 @@ private static void UpdateName(MemberSave memberSave)
...
-        ///     that we then use to create the member in the custom membership provider. This acts as the link between Umbraco data and
-        ///     the custom membership provider data. This gives us the ability to eventually have custom membership properties but still use
-        ///     a custom membership provider. If there is no 'Member' member type, then we will simply just create the membership provider member
-        ///     with no link to our data.
-        ///
-        /// If this is successful, it will go and re-fetch the IMember from the db because it will now have an ID because the Umbraco provider
-        /// uses the umbraco data store - then of course we need to re-map it to the saved property values.
-        /// </remarks>
-        private MembershipUser CreateWithMembershipProvider(MemberSave contentItem, out MembershipCreateStatus status)
+        // TODO: This logic should be pulled into the service layer
```
bergmania: 👍

src/Umbraco.Web/Editors/AuthenticationController.cs
Shazwazza: Yep i agree, I forgot we can access the current user on a thread without any http stuff so i can change that.

src/Umbraco.Web/Models/ContentEditing/MemberSave.cs
Shazwazza: We can't set the values in the ctor because the `Properties` value isn't set in the constructor, it is a get/set property which these values rely on. I was going to just have a lazy field value set to prevent recalculation but then I stopped caring :P

src/Umbraco.Web/Models/ContentEditing/MemberSave.cs
Shazwazza: Yes, these aren't set on the front-end.

src/Umbraco.Web/Editors/AuthenticationController.cs
Shazwazza: I will leave this as-is for now, we'll be changing more later with passwords, auth, members, users and i'm positive we'll be re visiting this.

PR #7267, opened 2019-12-03
adding timeout for Chrome to render before making the correction

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/7247

Now you can see and edit the DocType Alias in Chrome.


warrenbuckley:
Tested and this all works as expected

PR #7274, opened 2019-12-04
Checking nodeType on activeNode to ensure that nodes originate from same treeAlias

nielslyngsoe:
This fix ensures that the method 'getNodeCssClass' checks whether the given node and the activeNode have the same nodeType. aka. TreeAlias, that's at least what I could conclude. Quite some naming confusion.

This change only affects what nodes get the active class, so it should have a minor impact. But still important to test firmly if anything is broken by this change.


clausjensen:
Change looks good and I've tested that it seems to work as expected - nothing seemed broken 👍

PR #7193, opened 2019-11-20
Netcore: Abstract System.Web stuff

bergmania:
This PR is about abstracting some of the usage of System.Web, such that we can end up moving Umbraco.Core to netstandard

- Created an `IHostingEnvironment` interface and `AspNetHostingEncironment` implementation
- Created an `IIPResolver` interface and `AspNetIpResolver` implementation
- Created an `IBackOfficeInfo` interface and `AspNetBackOfficeInfo` implementation
- Created an `ISessionIdResolver` interface and `AspNetSessionIdResolver` implementation
- Created an `IRequestCache` and uses this instead of HttpContext.Current.Items
- Created an `IHtmlEncodedString` and uses this instead of System.Web.IHtmlString

PR #7276, opened 2019-12-04
Netcore: IOHelper cannot use System.Web

bergmania:
## Cannot be merged before https://github.com/umbraco/Umbraco-CMS/pull/7193

- Injecting `IHostingEnvironment` into `IOHelper`
- Split `IHostingSettings` out from `IGlobalSettings` to avoid a circular dependency between some config needed for `AspNetHostingEnvironment`
- Handle `IHtmlEncodedString` in views when typing `@VariableName`, instead of forced to `@Html.Raw(VariableName)`
- Moved more methods from StringExtensions in Core to Abstractions (No changes)


src/Umbraco.Web/Mvc/UmbracoViewPageOfTModel.cs
```diff
@@ -235,6 +236,18 @@ public override void WriteLiteral(object value)
             base.WriteLiteral(value);
         }
 
+        public override void Write(object value)
+        {
+            if (value is IHtmlEncodedString)
+            {
+                base.WriteLiteral(value);
```
Shazwazza: Should this be `value.ToHtmlString()` ? Since that is what the `System.Web.HttpUtility.HtmlEncode` is doing when writing out these values

src/Umbraco.Web/Mvc/UmbracoViewPageOfTModel.cs
```diff
@@ -235,6 +236,18 @@ public override void WriteLiteral(object value)
             base.WriteLiteral(value);
         }
 
+        public override void Write(object value)
+        {
+            if (value is IHtmlEncodedString)
```
Shazwazza: I guess the issue with our own abstraction on this is that any other type of view isn't going to deal with this correctly. For example, in partial views or just normal views that aren't using this one then strange things will happen. I realize MS didn't make this hugely flexible since the underlying logic for all views uses `HttpUtility.HtmlEncode` which explicilty checks for `IHtmlString`. 

Not really sure the best answer to that apart from configuring the common base razor view type in each web.config in the views folder to be this one (i.e. https://haacked.com/archive/2011/02/21/changing-base-type-of-a-razor-view.aspx/) . Maybe there's another way to do that programatically with some sort of extension point for views.

src/Umbraco.Web/Mvc/UmbracoViewPageOfTModel.cs
bergmania: I do that because this is safer. The only reason it worked now is that `ToString` and `ToHtmlString` is doing the same.

src/Umbraco.Core/Logging/Serilog/Enrichers/HttpRequestNumberEnricher.cs
```diff
@@ -22,11 +24,10 @@ internal class HttpRequestNumberEnricher : ILogEventEnricher
         /// </summary>
         private const string _httpRequestNumberPropertyName = "HttpRequestNumber";
 
-        private readonly Lazy<IAppCache> _requestCache;
 
-        public HttpRequestNumberEnricher(Lazy<IAppCache> requestCache)
+        public HttpRequestNumberEnricher(Func<IFactory> factoryFunc)
```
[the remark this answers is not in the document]
bergmania: The issue is that we don't have IRequestCache when we need to create the logging..

I realized I could change to a `Func<IRequestCache>` instead of `Func<Factory>`..

src/Umbraco.Abstractions/StringExtensions.cs
```diff
@@ -1213,5 +1214,80 @@ private static void SwapBytes(byte[] guid, int left, int right)
         /// </summary>
         public static string NullOrWhiteSpaceAsNull(this string text)
             => string.IsNullOrWhiteSpace(text) ? null : text;
+
+        /// <summary>
+        /// Ensures that a path has `~/` as prefix
+        /// </summary>
+        /// <param name="path"></param>
+        /// <returns></returns>
+        public static string EnsurePathIsApplicationRootPrefixed(this string path)
```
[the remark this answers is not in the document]
bergmania: I agree.. I'm already confused by all these extension methods on string and object 😱

src/Umbraco.Web/Mvc/UmbracoViewPageOfTModel.cs
bergmania: I'm not sure this is a real problem? 
Our views is generated with the `@inherits Umbraco.Web.Mvc.UmbracoViewPage<ContentModels.MyDocType>` in the top. 

You need this to use modelsbuilder right?

Shazwazza: Our normal template views are but that's only if  you create templates via the back office. You don't need to use that for models builder, the base class is just a helper. You can just as easily do `@model ContentModels.MyDocType`

Similarly this would be the approach for many people's partial views since most people aren't creating partial views via the back office but in VS instead. Many people will opt to using the base view class but it's not a requirement.

Shazwazza:
@bergmania if you can fix up the merge conflicts, then please merge this PR

PR #7290, opened 2019-12-06
Use input-id on toggle in profiler dashboard

bjarnef:
### Prerequisites

### Description
It seems the label next to toggle didn't changed the toggle anymore. It is because the rendered id on the button was `id="profilerAlwaysOn "` (with a whitespace last) instead of `id="profilerAlwaysOn"`.

I guess the `id` attribute on the toggle in some way conflicted with the internal `id` attribute on the `button` element inside `umb-toggle`.

Instead it now use `input-id="profilerAlwaysOn"`.

The `input-id` property has been added in this PR https://github.com/umbraco/Umbraco-CMS/pull/6512


kjac:
Thanks again, @bjarnef - it's merged 😄

Issue #7060, opened 2019-11-05
Include Prism for code snippets


bjarnef:
I wonder if it would be useful to include Prism for code snippets, e.g. to create a code snippet component, which could be used in query builder: https://github.com/umbraco/Umbraco-CMS/pull/6584

It could also be useful in packager view after installation to show code snippets (sometimes developers need to insert a few code snippets in e.g. views after installation of a package.

This could also be useful in TinyMCE editor, which support Prism be default to allow inserting code samples: https://www.tiny.cloud/docs-4x/plugins/codesample/

bjarnef:
@nielslyngsoe any thoughts on this?

nielslyngsoe:
@bjarnef interesting, I see that Sebastiaan already put it in for our talk on monday, so after that you will hear from us. :-)

bjarnef:
@nielslyngsoe I agree using the code samples in TinyMCE is mostly for developers/geeks, but it might also be useful in the new block editor, e.g. for blog posts https://shazwazza.com/post/how-to-register-mvc-controllers-shipped-with-a-class-library-in-aspnet-core/

nul800sebastiaan:
Hi @bjarnef - we're a little reluctant to add another library, although this is a small one, but we're considering it as an option. Unfortunately Razor support has been a feature request for 2 years (https://github.com/PrismJS/prism/issues/1225), which is quite limiting in the context of Umbraco.

The other option we thought of, and this would probably be preferred, is to look at re-using Ace in read-only mode. Could you play with that first and see if it makes sense to use that?

As for the block editor, I would think it's very often used, only few people use Umbraco to publish code snippets in their content, so I would see that as a plugin for Umbraco.

emma-hq:
Hey folks,

The fix for this was merged in #7294 so I'm going to close this one. Thanks for your work @bjarnef 

Emma

PR #7294, opened 2019-12-08
Add code snippet component

bjarnef:
### Prerequisites

### Description
This PR add a `umb-code-snippet` component, which includes a copy button to copy the code and enhance this feature from this PR https://github.com/umbraco/Umbraco-CMS/pull/6584#issuecomment-549855172

I have been inspired by these code samples https://docs.microsoft.com/en-us/odata/webapi/netcore (which are using highlight.js)

The idea is that package developers also can re-use this component in e.g. the view after installation, in dashboard and in another part of backoffice UI in core.

At the moment this doesn't use third parthy plugins like [highlight.js](https://highlightjs.org/) or [Prism](https://prismjs.com/). https://github.com/umbraco/Umbraco-CMS/issues/7060

Prism is lightweight and it supported by default in TinyMCE, but doesn't out of the box support razor (cshtml/vbhtml)

Highlight.js seems to have some support for razor (cshtml - not sure about vbhtml) https://github.com/highlightjs/highlightjs-cshtml-razor

It could use one of these or maybe the Ace Editor in readonly mode (https://ace.c9.io/#nav=api&api=editor under setReadOnly) https://github.com/Microsoft/monaco-editor/issues/54

It is also possible in Monaco Editor to set readonly mode https://github.com/Microsoft/monaco-editor/issues/54 if Umbraco choose to switch editor https://github.com/umbraco/Umbraco-CMS/pull/6267

Some enhancements that can be added/improved later

- Syntax highlighting: Prism, Highlight.js, Ace Editor, Monaco Editor
- Choose to show/hide copy button or entire "header" (maybe this should be shown by default).

### DIFF
--- a/src/Umbraco.Web.UI.Client/lib/bootstrap/less/code.less
+++ b/src/Umbraco.Web.UI.Client/lib/bootstrap/less/code.less
@@ -9,14 +9,13 @@ pre.code {
   padding: 0 3px 2px;
   #font > #family > .monospace;
   font-size: @baseFontSize - 2;
-  color: @grayDark;
+  color: @blueExtraDark;
   .border-radius(3px);
 }
 
 // Inline code
 code {
   padding: 2px 4px;
-  color: #d14;
   background-color: #f7f7f9;
   border: 1px solid #e1e1e8;
   white-space: nowrap;
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbcodesnippet.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbcodesnippet.directive.js
@@ -1,0 +1,100 @@
+﻿/**
+@ngdoc directive
+@name umbraco.directives.directive:umbCodeSnippet
+@restrict E
+@scope
+
+@description
+
+<h3>Markup example</h3>
+<pre>
+	<div ng-controller="My.Controller as vm">
+
+        <umb-code-snippet 
+            language="charp">
+            {{code}}
+        </umb-code-snippet>
+
+	</div>
+</pre>
+
+<h3>Controller example</h3>
+<pre>
+	(function () {
+		"use strict";
+
+		function Controller() {
+
+            var vm = this;
+
+        }
+
+		angular.module("umbraco").controller("My.Controller", Controller);
+
+	})();
+</pre>
+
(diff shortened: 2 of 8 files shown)


bjarnef:
@nielslyngsoe what do you think on this?

nielslyngsoe:
Hi @bjarnef 

I think it looks good, but I do feel that we should find a different color than red. It might give the impression that the code is wrong(has errors)?

And now that we are at it, I think it would be nice to use a brighter gray, what do you think?

bjarnef:
@nielslyngsoe the red color is how the `<code>` element is styled. Also the gray color is from the `<pre>` element, so I guess it is some general styles that need to be changed?

nielslyngsoe:
@bjarnef ah, that originates from Bootstrap CSS Library, which the BackOffice originally was build upon. But now we have decoupled the project so we have our own copy of the project, which we can change as wished. Feel free to change it.

bjarnef:
@nielslyngsoe we have this part for styling of `<pre>` element https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/ApiDocs/umbracotemplate/styles/main.css#L201-L211

and this for styling of `<code>` https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI/Content/bootstrap.css#L1536-L1542 https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web.UI/Content/bootstrap.css#L1572-L1579

bjarnef:
Should it be a neutral text color like this (black or dark gray'ish)?

https://docs.microsoft.com/en-us/odata/webapi/netcore

bjarnef:
It might also be possible to re-use the syntax highlighting from ace editor (which already is included):
https://stackoverflow.com/questions/8832320/how-can-i-highlight-code-with-ace-editor

nielslyngsoe:
@bjarnef would be interesting to investigate if ACE is a possibility.

If not possible with ACE then I think in terms of color that maybe this one would fit:
@blueExtraDark (#1b264f)
on a background color of:
@brownGrayLight (#f6f4f4)

But feel free to make a choice that you like, you can see color variables in variables.less

bjarnef:
@nielslyngsoe I have updated color of `code` element and adjusted the styling a bit.

image

PR #7222, opened 2019-11-25
Refactorying of PropertyActions

nielslyngsoe:
I have refactored the property actions, as requested by Shannon in this PR: https://github.com/umbraco/Umbraco-CMS/pull/6823

Changes include:
- Turning Nested Content Property Editor Controller into Component.
- Using UmbProperty directly for communication between PropertyActions and Property Editors.
- Using UmbVariantContent directly to get Content Name.
- Moved Umbraco.PropertyEditors.NestedContent.DocTypePickerController into its own file

This PR should be used for further review.

---
_This item has been added to our backlog [AB#3937](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3937)_

PR #7301, opened 2019-12-10
Make Nested Content actually save the item content again

kjac:
### Prerequisites

### Description

After #7222, Nested Content no longer saves of the item content 😲 only when you copy and paste items is the content saved:

nc-save-content-before

Specifically the refactoring in 503fd188 seems to be the culprit. Basically what happens is the controller wipes its "current node" from state before attempting to sync back the item content into the very same "current node".

This PR fixes things up:

nc-save-content-after

When testing this PR, make sure that all Nested Content operations save back the item content to the correct item - in particular:

1. Create new items with and without collapsing previously edited items
2. Edit existing items with and without collapsing previously edited items
3. Reorder items without collapsing previously edited items
4. Delete items with and without collapsing previously edited items
5. Copy and paste items

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.component.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.component.js
@@ -53,8 +53,8 @@
         });
 
         function setCurrentNode(node) {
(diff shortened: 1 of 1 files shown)


nielslyngsoe:
😵

@kjac Thanks, this change fixed the issue.

PR #7306, opened 2019-12-10
Introduce a new IMainDomLock and both default and sql implementations

Shazwazza:
Related to https://github.com/umbraco/Umbraco-CMS/issues/6546

The default MainDom implementation of using a system-wide semaphore lock doesn't natively work on Azure because Azure doesn't respect system-wide semaphore or event wait handles. So we want to introduce a MainDom that can work based on SQL/Db which is what this does.

* Adds a new MainDom lock to the locks table including migrations to add it for upgrades
* Abstracts out the locking mechanism for MainDom so we don't have to completely replace `IMainDom` and instead just replace the locking/signal part of it
* Implements the default abstraction for IMainDomLock which uses the normal semaphore lock
* Implements a new Sql IMainDomLock which uses SQL to acquire the main dom and to signal to release the main dom
* Adds an appSetting to be able to use the sql main dom

The way that the SQL main dom lock works is:

* On startup we get a distributed write lock (which has a timeout)
* we insert or update a row in the keyvalue table for the main dom with the current app domain Id
  * If the row was inserted then it means there wasn't an existing one and we've acquired MainDom. We update the row to the appDomain id.
  * If the row was updated, it means there was an existing MainDom and we need to signal to it that we are waiting and it should shut down, to do that we update the record with a temp GUID. The active MainDom then recognizes a change and starts unwinding. When it's done, it then update the row value and adds a suffix '_updated'. The appDomain coming online is polling for this value, if it detects that the value is it's temp ID + _updated, then it knows it can acquire. We update the row to the appDomainId. If the value has an _updated prefix but it doesn't match the ID it put there, it means another appdomain has come online and in this case we cannot acquire.
* On a natural appdomain shutdown, we will delete the row

There's a TODO note in this PR - which is that we would have the option to delete the maindom database row when maindom is acquired and this is the marker that maindom is acquired. The benefit of this is that if the appdomain is terminated and it doesn't wind down, then there is no orphan database row which means there isn't going to be a 1min 30sec wait for the next maindom to forcibly take over. But having a non db row as an indicator seems weird and possibly might not work.


src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
```diff
@@ -250,6 +250,11 @@ public override bool DoesTableExist(IDatabase db, string tableName)
         }
 
         public override void WriteLock(IDatabase db, params int[] lockIds)
+        {
+            WriteLock(db, 1800, lockIds);
+        }
+
+        public void WriteLock(IDatabase db, int millisecondsTimeout, params int[] lockIds)
```
bergmania: This seems to be a breaking change..

Before `WriteLock(db, 42)` would hit `public override void WriteLock(IDatabase db, params int[] lockIds)` now it hits 
` public void WriteLock(IDatabase db, int millisecondsTimeout, params int[] lockIds)`

I think it is cleaner to change the signature to 
```
  public void WriteLock(IDatabase db, TimeSpan timeout, params int[] lockIds)
```

src/Umbraco.Core/Runtime/SqlMainDomLock.cs
```diff
@@ -0,0 +1,407 @@
...
+namespace Umbraco.Core.Runtime
+{
+    internal class SqlMainDomLock : IMainDomLock
+    {
+        private string _lockId;
+        private const string MainDomKey = "Umbraco.Core.Runtime.SqlMainDom";
+        private readonly ILogger _logger;
+        private IUmbracoDatabase _db;
+        private CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
+        private SqlServerSyntaxProvider _sqlServerSyntax = new SqlServerSyntaxProvider();
```
bergmania: I think we need a bit more smooth handing if you are setting
```
<add key="Umbraco.Core.MainDom.Lock" value="SqlMainDomLock" />
```
while running SqlCE. It just explodes with `System.TimeoutException: Cannot acquire MainDom`
But the underlying issue is because SqlCE doesn't understand the Query Hints.

I discussed with @clausjensen, whether this needs to support SqlCe 
While I can't see any technical reasons it can't, we both agreed on it is okay if it just is a clear error message. You should never run `SqlMainDomLock` and `SqlCe` together.

For now, I suggest ensuring the DB is running SqlServer, otherwise throw an NotSupportedException.

src/Umbraco.Core/Runtime/SqlMainDomLock.cs
```diff
@@ -0,0 +1,407 @@
...
+
+        /// <summary>
+        /// Wait for any existing MainDom to release so we can continue booting
+        /// </summary>
+        /// <param name="tempId"></param>
+        /// <param name="millisecondsTimeout"></param>
+        /// <returns></returns>
+        private Task<bool> WaitForExistingAsync(string tempId, int millisecondsTimeout)
+        {
+            var updatedTempId = tempId + "_updated";
```
bergmania: I think the string `"_updated"` should be saved in a `const` we use that multiple times, and it is important it is the same

src/Umbraco.Core/Runtime/SqlMainDomLock.cs
```diff
@@ -0,0 +1,407 @@
...
+
+        /// <summary>
+        /// Checks if the exception is an SQL timeout
+        /// </summary>
+        /// <param name="exception"></param>
+        /// <returns></returns>
+        private bool IsLockTimeoutException(Exception exception) => exception is SqlException sqlException && sqlException.Number == 1222;
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
```
bergmania:
```suggestion
        private bool _disposedValue = false; // To detect redundant calls
```

Should follow normal naming convensions

src/Umbraco.Core/Runtime/SqlMainDomLock.cs
```diff
@@ -0,0 +1,407 @@
...
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposedValue)
```
bergmania:
```suggestion
            if (!_disposedValue)
```

src/Umbraco.Core/Runtime/SqlMainDomLock.cs
```diff
@@ -0,0 +1,407 @@
...
+using System.Threading.Tasks;
+using Umbraco.Core.Logging;
+using Umbraco.Core.Persistence;
+using Umbraco.Core.Persistence.Dtos;
+using Umbraco.Core.Persistence.Mappers;
+using Umbraco.Core.Persistence.SqlSyntax;
+
+namespace Umbraco.Core.Runtime
+{
+    internal class SqlMainDomLock : IMainDomLock
```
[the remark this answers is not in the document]
Shazwazza: yes but remember this service/object is created pre-bootup since it's required by bootup and at that state there is no UmbracoDatabase, repositories, services, etc... so We can't really get our normal KeyValueRepository into this class without service locators or something.

src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
Shazwazza: Ah yes good catch, will fix up

src/Umbraco.Core/Runtime/SqlMainDomLock.cs
Shazwazza: Yep, there's no reason why this should be used with SqlCe and it wasn't written with an intention to do so. Will fix up code to throw.

Shazwazza:
@bergmania have committed some changes and added some comments here. 

Regarding the complexity of the class and a KeyValueRepository type of class, we can do that but that will also require for us to overhaul the KeyValueService since that is basically acting like a repository currently. I could look into re-using this class for now? Would probably require some wrangling with a IScopeProvider instance. I'm happy to do what needs to be done but will mean this won't be done for a bit, what do you think?

bergmania:
Changes look good,

I don't understand why you changed `_sqlServerSyntax` to `_dbFactory.SqlContext.SqlSyntax`, in only some of the cases.., I guess it should be all or nothing?

I understand the issues with KeyValueRepository in this scenario - I turn a blind eye for now 🙈

bergmania:
@Shazwazza I have merged.. 
Regarding the TODOs I can't see any good ways to do this.. I also don't like a missing value as an indicator.

For now I think this is okay - As far as I can see we only have to wait 40 sec and not 90

Shazwazza:
Yep I agree and in theory it shouldn't happen anyways, not unless the process is completely terminated which would be rare.

teeto:
Hi, @Shazwazza , so in Azure hostings we must add this to the web.config or is active by default?
<add key="Umbraco.Core.MainDom.Lock" value="SqlMainDomLock" />

Shazwazza:
@teeto yes, please see updated docs for running on azure https://our.umbraco.com/documentation/Getting-Started/Setup/Server-Setup/azure-web-apps

teeto:
Thanks, i saw the docs after writing here and updated my web.confif to include it.
Also, i have added this line:
<add key="Umbraco.Core.LocalTempStorage" value="EnvironmentTemp" />
But after some errors i changed it for this:
<add key="Umbraco.Examine.LuceneDirectoryFactory" value="Examine.LuceneEngine.Directories.TempEnvDirectoryFactory, Examine" />

Whats the difference?

Shazwazza:
(your code snippets didn't come through)

teeto:
Thanks i didnt insert code, now it is edited

Shazwazza:
@teeto you need both of those appSettings (there are 3 appSettings required as per docs) 

```xml
<add
    key="Umbraco.Core.MainDom.Lock"
    value="SqlMainDomLock" />
```

Uses SQL as the distributed locking mechanism for synchronizing app domains instead of system-wide semaphore locks

```xml
<add
    key="Umbraco.Core.LocalTempStorage"
    value="EnvironmentTemp" />
```

Anything that can use temporary storage will be stored in the %temp% folder which is on the local fast drive in Azure. This is things like nucache cache files, cdf cache files, type loader cache files, and others. 

```xml
<add
    key="Umbraco.Examine.LuceneDirectoryFactory"
    value="Examine.LuceneEngine.Directories.SyncTempEnvDirectoryFactory, Examine" />
```

This is a special lucene "directory" that reads the active Lucene index files in the local fast drive in the %temp% folder and writes to both the %temp% folder and the main storage folder (App_Data, which is stored on the network share). When Azure moves your site to another worker it will sync the Lucene files on-demand from the main store (network share) to the local fast drive and read the index from there. This is so your index doesn't have to rebuild from scratch whenever azure moves your site. There's docs about this here (they're kinda hidden away) https://our.umbraco.com/Documentation/Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication#examine-directory-factory-options. The docs have changed a bit so need to figure out why this is only mentioned here now because it's also important that if you are load balancing and auto-scaling, you need to use `TempEnvDirectoryFactory` not `SyncTempEnvDirectoryFactory` on your front-end web app.

teeto:
Sorry i didnt write it properly, i meaned that i at first added SyncTempEnvDirectoryFactory and after some errors i changed to TempEnvDirectoryFactory, and it seemed to work fine but i dident know the difference it makes between the two. On umbraco 7 i used useTempStorage="Sync" which i think is the equivalent.

stodolos:
> Sorry i didnt write it properly, i meaned that at first i added SyncTempEnvDirectoryFactory and after some errors i changed to TempEnvDirectoryFactory, and it seemed to work fine but i didnt know the difference it makes between the two. On umbraco 7 i used useTempStorage="Sync" which i think is the equivalent.
> I use Azure web apps but i dont use auto scaling and only using one instance, but i guess it is using load balancing anyways so thats the reason that i better use TempEnvDirectoryFactory

I also use TempEnvDirectoryFactory ever since I read [this article](https://www.intelegencia.com/blog/technology/architecting-the-umbraco-platform).  I also upgraded my Examine when I did this.

stodolos:
I did come across an issue with the Umbraco.Core.MainDom.Lock in place on my staging app service before the DB had actually been upgraded.  Here are the steps I took:

1. stopped the app service
2. applied the setting in prep for the new code to be deployed
3. deployed the 8.6.3 code to my app service
4. started the app service

Received the following error:
> Umbraco.Core.Exceptions.BootFailedException: Boot failed. 
> ---> System.TimeoutException: Cannot acquire MainDom
> 
>    at Umbraco.Core.Runtime.MainDom.Acquire() in d:\\a\\1\\s\\src\\Umbraco.Core\\Runtime\\MainDom.cs:line 160
> 
>    at Umbraco.Core.Runtime.MainDom.<get_IsMainDom>b__14_0() in d:\\a\\1\\s\\src\\Umbraco.Core\\Runtime\\MainDom.cs:line 186
> 
>    at System.Threading.LazyInitializer.EnsureInitializedCore[T](T& target, Boolean& initialized, Object& syncLock, Func`1 valueFactory)
> 
>    at Umbraco.Core.Runtime.MainDom.get_IsMainDom() in d:\\a\\1\\s\\src\\Umbraco.Core\\Runtime\\MainDom.cs:line 186
> 
>    at Umbraco.Core.Runtime.CoreRuntime.AcquireMainDom(IMainDom mainDom) in d:\\a\\1\\s\\src\\Umbraco.Core\\Runtime\\CoreRuntime.cs:line 263
> 
>    at Umbraco.Core.Runtime.CoreRuntime.Boot(IRegister register, DisposableTimer timer) in d:\\a\\1\\s\\src\\Umbraco.Core\\Runtime\\CoreRuntime.cs:line 166
> 
>    --- End of inner exception stack trace ---","FailMessage":"Boot failed.","Duration":967,"TimingId":"...","SourceContext":"Umbraco.Core.Runtime.CoreRuntime","ProcessId":...,"ProcessName":"w3wp","ThreadId":1,"...":2,"AppDomainAppId":"...","MachineName":"...","Log4NetLevel":"ERROR"

To get it working again, I had to:

1. stop the app service
2. remove the setting
3. started the app service back up
4. went through the upgrade process
5. stopped the site
6. added the setting back

zacharybil:
I encountered the same error as stodolos when upgrading to 8.6 (from 8.3) and adding the SqlMainDomLock setting in the same release.

Following stodolos steps sorted it out for me

> To get it working again, I had to:
> 1. stop the app service
> 2. remove the setting
> 3. started the app service back up
> 4. went through the upgrade process
> 5. stopped the site
> 6. added the setting back

Perhaps a note about this should be added to the docs:
https://github.com/umbraco/UmbracoDocs/blob/main/Getting-Started/Setup/Server-Setup/azure-web-apps.md

xwj-june:
Same error as stodolos and zacharybil happened to my local web app when upgrading 8.5.3 to 8.6.6. I sorted it out follows stodolos's steps but with some changes since my issue is in local:

1. Taking off SqlMainDomLock setting (if you don't have one, continue with the next step)
2. Complete the Umbraco database update (in case you don't know, you will need to go to the back office where it will prompt you to upgrade the database
3. Add back SqlMainDomLock setting, the site should be working now

PR #6998, opened 2019-10-30
Nested Content: Fix controls to be more accessible 

BatJan:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
I have changed the `<a>` to `<button>` and added some non-visual texts and fixed the styling.

Below is an image that indicates what controls I'm refering too - Visually it looks and acts the same as before.

nested-content-controls

PR #7307, opened 2019-12-10
V8: Don't add new Nested Content items on enter key

kjac:
### Prerequisites

### Description

#6998 had an unintended side effect: Whenever you hit the enter key in a simple property editor within a Nested Content item, the "Add" button is invoked, displaying the Nested Content item picker (or just adding another item when in single item mode):

nc-add-on-enter

This PR fixes it by adding the missing `type="button` to the "Add" button (thus form submit doesn't invoke the button).


poornimanayar:
@kjac ,

I think it will be worth getting this into the `v8/contrib` branch. so can you fix the merge conflicts against the branch.

Poornima

kjac:
@poornimanayar all done 👍

poornimanayar:
And merged :-) Thanks as always @kjac

PR #7308, opened 2019-12-11
V8: Prevent the Nested Content item picker from crashing

kjac:
### Prerequisites

### Description

If you're tabbing your way through content editing and you accidentally hit the Enter key twice when adding new items to Nested Content, the itempicker overlay crashes hard:

nc-itempicker-double-enter-before

This is an unintended side effect of #6998 - the new `<button>` element for "Add content" retains focus whereas the previous `<a>` element did not. Thus the Nested Content controller attempts to open the overlay twice, which causes it to fail.

I have tried applying autofocus to the overlay as a solution, but `umb-auto-focus` doesn't really work in this case. So I have opted to let the Nested Content controller handle it instead by ensuring it doesn't open the overlay twice.


poornimanayar:
And merged :-)

PR #7310, opened 2019-12-11
V7: Discard MiniProfiler results when not in debug mode

ronaldbarendse:
### Prerequisites

### Description
This PR undos the fix (ugly hack IMHO) applied in https://github.com/umbraco/Umbraco-CMS/pull/6114 and correctly discards the MiniProfiler results when not in debug mode. This ensures the results aren't even stored, which seems the best thing to do (instead of removing the MVC routes to disable accessing them via the results URL).

Because of [this bug in MiniProfiler](https://github.com/MiniProfiler/dotnet/issues/96), the results URL would throw a `NullReferenceException` when nothing is stored and no storage is set. This PR also explicitly sets the storage to the default `HttpRuntimeCacheStorage`, but with a cache duration of 30 minutes (instead of 1 day). This should be enough time to view the profiling results and is actually the [default since MiniProfiler 4.1.0](https://github.com/MiniProfiler/dotnet/blob/6b3a42a65fed15b8f8b35804668fbc2ea429402b/src/MiniProfiler/MiniProfilerOptions.cs#L25-L26).

Steps to test this:
- Set `system.web/compilation@debug=true`, start the site and check the profiling results URL: you should see results
- Set `system.web/compilation@debug=false`, start the site and check whether the results URL returns a 404 status code and does not return any results

---
_This item has been added to our backlog [AB#4561](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4561)_

### DIFF
--- a/src/Umbraco.Web/Profiling/WebProfiler.cs
+++ b/src/Umbraco.Web/Profiling/WebProfiler.cs
@@ -3,6 +3,7 @@ using System.Threading;
 using System.Web;
 using StackExchange.Profiling;
 using StackExchange.Profiling.SqlFormatters;
+using StackExchange.Profiling.Storage;
 using Umbraco.Core;
 using Umbraco.Core.Configuration;
 using Umbraco.Core.Logging;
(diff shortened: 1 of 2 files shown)


emmaburstow:
Hi @ronaldbarendse,

thanks for another PR and also thanks for the info above. Gives us plenty to work with.

Emma

emmaburstow:
Hi @ronaldbarendse 

Thanks for this work and thanks for the testing steps. We'll take a look soon and shout if we need anything else from you.

Emma

ronaldbarendse:
@emmaburstow I've created this PR for Umbraco 7 after having contact with @clausjensen and he indicated this might get merged in, even though PR #6114 already fixes the problem. As these changes ensure the MiniProfiler results aren't even stored (when not in debug mode), it's a better and more secure solution, so it may be included in the next 7.15 patch release (if/when that will get released).

warrenbuckley:
Thanks @ronaldbarendse 
All tested & merged in :)

PR #7311, opened 2019-12-11
V8: Make the Multi URL Picker sorting behave the same as MNTP

kjac:
### Prerequisites

### Description

The Multi URL Picker sorting doesn't quite behave like sortable lists generally does in Umbraco (e.g. in MNTP):

multi-url-sortable-before

Most noticeable is probably how the Multi URL Picker sorting doesn't limit the item movement to the y axis only. But also the Multi URL Picker items behave differently among themselves (when the sort is triggered) and they're not slightly transparent as is the case with other sortables. 

So this PR basically just makes the Multi URL Picker sorting behave the same way as it does for MNTP:

multi-url-sortable-after


poornimanayar:
And merged :-)

PR #7312, opened 2019-12-11
Don't show actions menu if there are no actions

mattbrailsford:
The umb-editor-menu directive is responsible for fetching menu items based on a current node, but if there are no menu items returned for that node an empty menu is displayed.

This PR prevents the menu from being rendered if there are no menu items to display in the menu.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-menu.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-menu.html
@@ -1,4 +1,4 @@
-<div class="pull-right" style="position: relative;">
+<div class="pull-right" style="position: relative;" ng-if="actions && actions.length > 0">
 
     <umb-button
         type="button"


nul800sebastiaan:
Seems to make sense, the only thing I'm questioning is the "hiding" part. Maybe it is strange to not have it there for consistency? So instead of hiding it, it could be a disabled item? Not sure.

mattbrailsford:
I personally think it's ok to hide it as I think it would be more confusing to see something you can't interact with taking up mental space, when you could remove it and know that everything you see is something you can use.

</2cents>

bjarnef:
I think it would be confusing to editors that the actions menu sometimes is clickable and sometimes not. It would be better just to only show the button if there are any relevant actions.

nul800sebastiaan:
All good, we should indeed hide the actions menu of there are no actions in it! 👍

poornimanayar:
Hi @mattbrailsford ,

Do you have any test plugin using which we can test this?

Poornima

kjac:
@mattbrailsford while the actions menu button indeed disappears if there are no actions available, there are a few issues with this PR:

1. The actions button now glitches when you browse around, as it is now only displayed after the XHR to `GetMenu` completes:
actions-menu-glitch

2. When there are actions available, the actions menu button is shown but the menu doesn't expand (replacing `ng-if` with `ng-show` in your PR fixes this): 
actions-menu-unresponsive

I think the glitching is really super annoying 😞 perhaps it was better to disable the actions menu button rather than hiding it?

mattbrailsford:
@kjac hmm, I see what you mean, though it would feel like just disabling the action menu would be a “fix” just to stop it glitching.

I wonder how it would feel if you waited till after the menu is loaded before rendering the header / editor? Effectively this is what you are wanting in a sense “dhow the header buttons/ actions in one go” which requires you to know everything before you render. 

I think disabling will still have a level of glitch if you are in a section that has no actions it will be accessible then quickly flash to disabled.

kjac:
@mattbrailsford waiting until everything is loaded would indeed be a fix. I can't really judge the impact on the editor experience, though. My gut tells me it'll feel slower (even if it's not) because the editor seems to load longer before anything (everything) happens. I'm thinking @nielslyngsoe should get in on that discussion.

I was thinking we could disable the actions menu until the actions have loaded - not disable it *after* if no actions are available. The glitching would be minimal if the disabling is made somewhat subtle, and in reality it makes sense - you can't really operate the actions menu until the actions have loaded.

c9mbundy:
FWIW: My 2c. 
Disabled buttons and dropdowns are a common-place UX that users won't be confused by, provided the UX is appropriately styled. 
However, the current 'Actions' button is breaking the UX pattern of the buttons to the left of it - an Icon with Text below. 
If there was an appropriate 'Actions' icon, then it becomes easier to place the text 'No Actions' below as the default (both icon and texted grayed) and update to 'Actions' with an enabled style, as actions are loaded. 
Not having an Actions button leaves the user in doubt as to whether or not there should be one, having a grayed out button saying 'No Actions' leaves them in no doubt.

mattbrailsford:
@kjac yea, that's understandable. 

This isn't the hill I'm going to die on so if a disabled action button is the lesser evil, then that's fine by me. My general principle is that of "don't make me think" which would suggest don't show something you can't interact with, but I understand the reasoning for it. Either way, it's better than what we have now 👍

kjac:
@mattbrailsford I'm all for incremental changes 😄 

FYI there is already an option in the `umb-editor-header` directive to hide the actions menu altogether. I don't know if this is applicable to your use case? If I'm not mistaken you can apply `hide-actions-menu="true"` to hide it.

c9mbundy:
@mattbrailsford - From a UX point of view, hiding a button sometimes and not others, actually leaves the user needing to think about it... 'where's the button?' ...because a missing button provides no explanation for it's non existence and so creates a level of inconsistency. A disabled button - and one that clearly indicates that it is disabled - provides the consistency of being where the user expected to see it, and a level of feedback that it is disabled in this case. How good the job is of conveying the 'not available for this situation' depends on a combination of iconography, text, color and even hover/tap dialogs.

nielslyngsoe:
@mattbrailsford It would be nice if you could present the case where this is needed. Since we will each be guessing a bit on the context of this request.
If I'm guessing correctly, then your context is not in the Content-section, but in a custom section?
If so, and if no actions is a consistent fact for that section, then it makes most sense not to see the actions button at any time for that section.
I would love you to investigate the option that @kjac presents. That would, from my current perspective, be the best solution.

In regards to delaying the display of the sub-header, that would not be preferable. Most users will be navigating content nodes, and we should strive to make content editing as fast as possible, with a few glitches as possible.

mattbrailsford:
@nielslyngsoe I'm fine with what @kjac has presented so it's fine to go ahead with that approach.

As far as the use case goes, yes this is in the context of a custom section but I think it sill holds true to standard Umbraco sections.

In my situation I have some areas that present list views for that area which I don't personally have any action items for and so when the view renders, the action dropdown is displayed with no items.

Now I know what you are going to say "but if you don't have any actions, don't show the menu" BUT for my given context at least, and I would claim the same for Umbraco core, action menus are an important extension point where someone might want to inject a custom action to perform additional tasks relevant to that area and so me closing off the menu explicitly because I personally don't have any actions completely cuts this off for others.

Now I could fetch the action menu items and see if there are any and then hide the menu myself, but this as well would be a waste because I would look them up, and if I do have them, I'd tell the header area about it, then the header area would again go look up the menu items wasting a request. 

And thus, my request is that I can just render the actions menu, whether I know I have any menu actions myself or not and have the actions menu determine itself whether to display or not.

nielslyngsoe:
@mattbrailsford Its a good point, And I think we all agree, it comes down to the technical problem of it not immediately begin determined whether there are actions. Said in other words, we would have a glitch or slower presentation, not very attractive options.

I'm happy that the solution works for you for now, and I understand the downside of this.

Let us know if the extensibility problem scales, then we can take a deeper look into a solution.

kjac:
> I'm fine with what @kjac has presented so it's fine to go ahead with that approach.

@mattbrailsford just to clarify here... can you use the `hide-actions-menu` option on `umb-editor-header` or did you mean _"It'd be fine to implement the disabled-until-ready approach"_?

mattbrailsford:
@kjac I'm saying I can't use `hide-actions-menu` is it blocks off that extension point for others. I'm happy to go with your "disable till ready" approach 👍

nul800sebastiaan:
I'll re-open this one, so what needs to be done to finish this, or is the current PR sufficient? (sorry, did not follow the full conversation that closely).

kjac:
I'll give the alternative approach a try tomorrow 👍

kjac:
New PR in #7446

mattbrailsford:
Closed my PR as happy @kjac's new PR resolves the issue

PR #7318, opened 2019-12-11
Updated content edit controller to hide save button when update permission is removed

### LINKED ISSUE #7314: Buttons panel display incorrect state after permissions change (after save event)
If we add **OnSaved** events in the backend to remove **Read** permission of the saved item (for example, changing Document permissions from **Read + Update** to **Read**), after the save **UI** still shows **Save** button even though it shouldn't.

## Umbraco version 8

I am seeing this issue on Umbraco version: 8.2.1

Reproduction
------------

### Bug summary

**Content** returned from the **API** (`.../content/getById?id=...`) returns proper permissions.
`Actions` menu contains correct actions.
`Buttons panel` still contains `Save` button.

### Specifics

I'm pretty sure the issue is with this code:
`Umbraco-CMS/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js`

```
...
            if ($scope.app && $scope.app.alias !== "umbContent" && $scope.app.alias !== "umbInfo" && $scope.app.alias !== "umbListView") {
                $scope.defaultButton = null;
                $scope.subButtons = null;
                $scope.page.showSaveButton = false;
                $scope.page.showPreviewButton = false;
                return;
            }

            // create the save button
            if (_.contains($scope.content.allowedActions, "A")) {
                $scope.page.showSaveButton = true;
                // add ellipsis to the save button if it opens the variant overlay
                $scope.page.saveButtonEllipsis = content.variants && content.variants.length > 1 ? "true" : "false";
            }
...
```

What happens is that since initially  **Document** contains **Update**  permissions the field `$scope.page.showSaveButton` is set to true.
After the content refresh there is no code to set the value to false.
Essentially this fixes the issue:
```
            // create the save button
            if (_.contains($scope.content.allowedActions, "A")) {
                $scope.page.showSaveButton = true;
                // add ellipsis to the save button if it opens the variant overlay
                $scope.page.saveButtonEllipsis = content.variants && content.variants.length > 1 ? "true" : "false";
            } else {
                $scope.page.showSaveButton = false;
            }
```

### Steps to reproduce

1. Add component which attaches itself to the ContentSaved event
2. In the handler update saved content permissions and remove "A" permission - update permission.
3. Login to the backoffice with an appropriate user.
4. Save document

### Expected result

Save button is removed from the **UI** since the user shouldn't have ability to save the document anymore.

### Actual result

Save button is still displayed

asmailov:
Fixes #7314


kjac:
Hi @asmailov - thank you for your PR 👍 

Could you please add a code snippet as you mention on #7314 to allow us to test this change?

asmailov:
> Hi @asmailov - thank you for your PR 👍
> 
> Could you please add a code snippet as you mention on #7314 to allow us to test this change?

Are you asking for the code for "ContentSave" event in which I change permissions?

kjac:
@asmailov yes please, in a somewhat condensed form that allows us to test your PR 😄

asmailov:
@kjac 

Steps to reproduce:

1. Install basic Umbraco
2. Add a test user group (with alias: `testGroup` or other, but then you need to update snippet) with following configuration:
image
image
3. Add a test user to the newly created user group
4. Create a document (any document is fine)
5. Add this **Composer** and **Component**
```
namespace Umbraco_1
{
    public class PermissionsComposer : IComposer
    {
        {
            composition
                .Components()
                .Append<PermissionsComponent>();
        }
    }

    public class PermissionsComponent : IComponent
    {
        private readonly IUserService userService;

        public PermissionsComponent(IUserService userService)
        {
            this.userService = userService;
        }

        public void Initialize()
        {
            ContentService.Saved += ContentService_Saved;
        }

        private void ContentService_Saved(IContentService sender, ContentSavedEventArgs e)
        {
            var groupAlias = "testGroup"; // <- insert group alias
            var testGroup = userService.GetUserGroupByAlias(groupAlias);

            if (testGroup == null)
            {
                throw new Exception($"Group with a name: '{groupAlias}' was not found!");
            }

            var groupId = testGroup.Id;
            var entitiesIds = GetEntitiesIds(e.SavedEntities);

            userService.ReplaceUserGroupPermissions(
                testGroup.Id,
                "F",
                entitiesIds
            );

            int GetId(IContent content) => content.Id;
            int[] GetEntitiesIds(IEnumerable<IContent> contents) =>
                contents
                    .Select(GetId)
                    .ToArray();
        }

        public void Terminate() { }
    }
}
```
6. Login to backoffice with the test user
7. Open the document mentioned in step 4
8. Save the document

**Expected:** all the buttons disappear except for **Preview**
**Actual:** all the buttons disappear except for **Preview** and **Save**.

On refresh, save buttons disappear.

kjac:
Merged 👍 thanks again!

Issue #7317, opened 2019-12-11
V8: Lowercase media-path hardcoded in regex


enkelmedia:
We had a strange issue with a Umbraco 8.4-site where images uploaded would not be selectable in a Media Picker when "select only images" was activated.

When I looked closer at it, the "Media Grid" is setting up the selectable-state based on the property mediaEntity.metaData.MediaPath which was always null in our case. It also turns out that all rows in the umbracoMediaVersion-table had the value "null" for the path-column.

After some digging around it turned out the we've had this setting in web.config

`<add key="umbracoMediaPath" value="~/Media" />`

Note the uppercase M in Media, then when the MediaFactory is trying to use regex to find a media-url inside a property value this fails since the regex MediaPathPattern is hardcoded to /media (lowercase).

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Core/Persistence/Factories/ContentBaseFactory.cs#L13

The same would also be true if one where to rename the media-folder to something like "assets", then this regex would not work, umbracoMediaVersion.parth would be null and the images would not be selectable in the "Image-Media-Picker" or in the RTE/Grid so it's not just a matter of making it case-insensible, it needs to respect the settings in web.config.

nul800sebastiaan:
Looks like this regex has been there for over 2 years now, while the hardcoding is not great, you should also wonder why someone found it necessary to update this default setting to capital M? Althouugh I now fully expect you to come back and say: this site started on Umbraco 4.5.2 and that was the default web.config at the time! 😉

Anyway, not important. Do you think we can update the regex to include the configured `umbracoMediaPath` and strip the leading `~` off so that the regex works?

Would be good to get that fixed in one of the upcoming v8 releases.

umbrabot:
Hi @enkelmedia,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

enkelmedia:
[...]

PR #7321, opened 2019-12-12
Temp 7317

enkelmedia:
### Prerequisites

Fixes: https://github.com/umbraco/Umbraco-CMS/issues/7317

### Description
Fixes an issue in ContentBaseFactory where the path to the media-folder was hardcoded to /media (lowercase), with this fix we respect the setting in web.config when looking for media-paths inside a property.

I've also made it IgnoreCase to support scenarios where one might have a mix of /Media and /media-paths in the database. This also means removing "Compiled" from the RegEx which probably have a extremely small impact, maybe even boosts performance depending on how many times media files are saved. (Read about regex compile here: https://rimdev.io/regex-performance-with-and-without-regexoptions-compiled-using-dotnet-framework-48-and-net-core-31-december-2019/).

Cheers!


enkelmedia:
Must have branched from the 8.4 branch when I created this, closing this PR and getting back with a new one.

PR #7322, opened 2019-12-12
Fixes #7317 issue with RegEx for media in ContentBaseFactory

### LINKED ISSUE #7317: V8: Lowercase media-path hardcoded in regex

enkelmedia:
Fixes: #7317

Description
Fixes an issue in ContentBaseFactory where the path to the media-folder was hardcoded to /media (lowercase), with this fix we respect the setting in web.config when looking for media-paths inside a property.

I've also made it IgnoreCase to support scenarios where one might have a mix of /Media and /media-paths in the database. This also means removing "Compiled" from the RegEx which probably have a extremely small impact, maybe even boosts performance depending on how many times media files are saved. (Read about regex compile here: https://rimdev.io/regex-performance-with-and-without-regexoptions-compiled-using-dotnet-framework-48-and-net-core-31-december-2019/).

Cheers!

### DIFF
--- a/src/Umbraco.Core/Persistence/Factories/ContentBaseFactory.cs
+++ b/src/Umbraco.Core/Persistence/Factories/ContentBaseFactory.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
+using Umbraco.Core.IO;
 using Umbraco.Core.Models;
 using Umbraco.Core.Persistence.Dtos;
 using Umbraco.Core.Persistence.Repositories;
@@ -10,7 +11,7 @@ namespace Umbraco.Core.Persistence.Factories
 {
     internal class ContentBaseFactory
     {
-        private static readonly Regex MediaPathPattern = new Regex(@"(/media/.+?)(?:['""]|$)", RegexOptions.Compiled);
+        private static readonly Regex MediaPathPattern = new Regex("(" + SystemDirectories.Media.TrimStart("~").EnsureEndsWith("/") + ".+?)(?:['\"]|$)", RegexOptions.IgnoreCase);
 
         /// <summary>
         /// Builds an IContent item from a dto and content type.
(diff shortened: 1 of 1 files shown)


enkelmedia:
Great feedback @kjac =D I've updated the code, added back Regex.Compiled and used your regex, tried it out and it worked nice for me.

kjac:
Hi @enkelmedia,

In the meantime someone else added changes to `ContentBaseFactory`, leaving your PR with a merge conflict 😞 could you please have a look at resolving that?

enkelmedia:
Seems like there is another solution for this problem coming:

https://github.com/umbraco/Umbraco-CMS/commit/26f1e77813642f3c2badc9bf3ef2a1cd7e1a0383

There where comments about this in the code before and the RegEx has been removed in favor of something that used the dataEditor to get the URL. 

I guess this leaves this PR useless? =D

kjac:
@enkelmedia by the looks of it... it would seem so 🙈

PR #7325, opened 2019-12-12
V8: Don't load languages in treepicker unless they're needed

### LINKED ISSUE #7313: Multi node tree picker (MNTP) start node always seems to be default language - not variant aware
Context: a multilingual Umbraco 8.4.0 site with multiple languages. When I set the start node for a MNTP it always shows the nodes for the default culture, even when editing content in a different culture. So if I am editing content with German culture then the MNTP shows the nodes as if I were viewing them as the default culture (en-US).

## Umbraco version

I am seeing this issue on Umbraco version: 8.4.0

Reproduction
------------

### Bug summary

MNTP doesn't respect current editor culture.

### Specifics

MNTP lets you pick nodes that aren't published in the editor culture you are using.

### Steps to reproduce

* Create a site with multiple language cultures and set "en-US" as default language.
* Create a document type with "Allow varying by culture" enabled
* Create a property that can also be varied by culture and use the MNTP as the property editor
* Set the MNTP start node to a listing page
* On that listing page create some child nodes that vary by culture. Publish some of these nodes in the default language and some in, say, "de-DE" culture. Just ensure the DE tree is different to the US tree.
* As an editor set the tree to DE and then go to the page with the MNTP. It will show a list of pages from the US tree, rather than the DE tree. It will show as published nodes that aren't published in the DE language, which are selectable. (If you try and access these nodes in code they have a URL set to "#")

**Note:** This happens if you set the start node by picking it _and_ also if you use XPath to set it. The result is the same.

### Expected result

I'd expect the MNTP to show me the DE version of the tree when editing in DE, with any nodes that are not published in DE greyed-out and unselectable.

### Actual result

You see the US tree. This lets you select nodes that are not published in DE (and also greys out nodes that are published in DE but not in US).

### Screenshots

German tree for "Kunder" (Customers) as seen in the content tree view:

image

The view the MNTP shows when editing using DE culture. Note it doesn't show the same view. The nodes "Practicus", "Advantage Solutions" and "Balfour Beatty" are not published in DE version, yet are shown as published and selectable.

image

Here is the US tree view for "Customers". You can see the MNTP is showing this view, even when editing in DE.

image

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7313

### Description

The problem described in #7313 basically happens because the treepicker loads all languages by default, in case it is configured to display an embedded language picker. 

If it isn't configured to do that (which is the default), those loaded languages are never used. So it seems really silly to load them, specially as they're now causing problems. 

This PR ensures that the languages are only loaded if they're required by the treepicker configuration. When applied, the MNTP starts behaving like you'd expect it to - both with and without start nodes configured:

treepicker-load-languages

*Note: In reality the problem described in #7313 is dependent on client rendering timing. Sometimes the languages are loaded before the treepicker fetches its nodes, sometimes after - and this determines if the nodes are fetched in the default language or the current editor language.*

### To clean up, or not to clean up?

That is the question!

I would really like to clean up the whole language selector feature of the treepicker. It isn't used anywhere anymore, it's a reminiscent of previously being able to toggle languages explicitly in the treepicker. 

However - cleaning it all up would be considered a breaking change on the clientside. 3rd party packages *could* in theory be using the feature.

But then again: If the feature isn't used by the core, it won't get regression tested and is bound to stop working at some point (as a matter of fact I don't even know if it works now).

I'll raise another issue where we can debate this.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/treepicker/treepicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/treepicker/treepicker.controller.js
@@ -75,18 +75,20 @@ angular.module("umbraco").controller("Umbraco.Editors.TreePickerController",
          */
         function onInit () {
 
-            // load languages
-            languageResource.getAll().then(function (languages) {
-                vm.languages = languages;
+            if (vm.showLanguageSelector) {
+                // load languages
+                languageResource.getAll().then(function (languages) {
+                    vm.languages = languages;
 
-                // set the default language
-                vm.languages.forEach(function (language) {
-                    if (language.isDefault) {
-                        vm.selectedLanguage = language;
-                        vm.languageSelectorIsOpen = false;
-                    }
+                    // set the default language
+                    vm.languages.forEach(function (language) {
+                        if (language.isDefault) {
+                            vm.selectedLanguage = language;
+                            vm.languageSelectorIsOpen = false;
+                        }
+                    });
                 });
-            });
+            }
 
             if (vm.treeAlias === "content") {
                 vm.entityType = "Document";
@@ -211,7 +213,7 @@ angular.module("umbraco").controller("Umbraco.Editors.TreePickerController",
             if (vm.dataTypeKey) {
                 queryParams["dataTypeKey"] = vm.dataTypeKey;
             }
-                
+
             var queryString = $.param(queryParams); //create the query string from the params object
             
             if (!queryString) {


ronaldbarendse:
I'd like to know why the language selector in the link picker isn't currently used... It would be a great feature to be able to select a node and store the culture, so you can render the content in the requested culture or just create links between cultures. I'm doing some Multi URL Picker enhancements (https://github.com/umbraco/Umbraco-CMS/pull/7533), so I'm looking at all the different ways it's currently used and what's still missing ;-)

ronaldbarendse:
Ah, just read https://github.com/umbraco/Umbraco-CMS/issues/7326#issuecomment-566496511 👍 Great to see the ability to select different nodes/cultures isn't forgotten!

poornimanayar:
And merged! Thanks @kjac ⭐️

PR #7330, opened 2019-12-13
V8: Wildcards in tree searches causes YSODs

kjac:
### Prerequisites

### Description

If you try to use a `*` in any search field (global search, copy/move dialog search, treepicker search, ...), things tend to go horribly wrong:

image

This PR explicitly removes the `*` character in the search queries to avoid that:

image

This doesn't change the fact that the tree searcher still performs wildcard queries:

image

### DIFF
--- a/src/Umbraco.Core/Persistence/Querying/ExpressionVisitorBase.cs
+++ b/src/Umbraco.Core/Persistence/Querying/ExpressionVisitorBase.cs
@@ -744,7 +744,9 @@ namespace Umbraco.Core.Persistence.Querying
 
             var c = exp[0];
             return (c == '"' || c == '`' || c == '\'') && exp[exp.Length - 1] == c
-                ? exp.Substring(1, exp.Length - 2)
+                ? exp.Length == 1
+                    ? string.Empty
+                    : exp.Substring(1, exp.Length - 2)
                 : exp;
         }
     }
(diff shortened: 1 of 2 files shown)


bjarnef:
@kjac I could also trigger a ysod by including `_` (underscore) in search term.

kjac:
@bjarnef good grief! I will amend the PR. Thanks 👍

bjarnef:
@kjac `-` (hyphen) also trigger the ysod ⚠️

kjac:
Magic markdown code block char ` also does it.

kjac:
Aha! So as it turns out, the problem is two-fold:

1. Lucene search does not like the chars `* - _` so those are now stripped from the Lucene searches.
2. SQL queries containing only one quotation mark (``' " ` ``) throws a logic error when stripping those quotation marks.

The PR has been updated to handle both scenarios. In summary:

### Testing this PR

Global search, copy/move dialog search, treepicker search etc. should *not* throw a YSOD when entering one and only one char of the following: ``* - _ ` ' "``

marcemarc:
Is it specific to those characters or all the Lucene 'Special Characters'?

+ - && || ! ( ) { } [ ] ^ " ~ * ? : \

http://lucene.apache.org/core/2_9_4/queryparsersyntax.html#N10180

kjac:
@marcemarc thanks for the input 😄 but no, for the Lucene searches it's only `* - _`

nul800sebastiaan:
Whoop! 🎉  Much better now, cheers! 👍

PR #7331, opened 2019-12-13
V8: Don't append empty file names to search results

kjac:
### Prerequisites

### Description

With #6579 we started appending the actual file names of uploaded files in the search result. Unfortunately this wasn't tested with media folder search results:

image

...so this PR ensures that we only append file names that aren't empty:

image

### DIFF
--- a/src/Umbraco.Web/Models/Mapping/EntityMapDefinition.cs
+++ b/src/Umbraco.Web/Models/Mapping/EntityMapDefinition.cs
@@ -177,12 +177,9 @@ namespace Umbraco.Web.Models.Mapping
 
             target.Name = source.Values.ContainsKey("nodeName") ? source.Values["nodeName"] : "[no name]";
 
-            if (source.Values.TryGetValue(UmbracoExamineIndex.UmbracoFileFieldName, out var umbracoFile))
+            if (source.Values.TryGetValue(UmbracoExamineIndex.UmbracoFileFieldName, out var umbracoFile) && umbracoFile.IsNullOrWhiteSpace() == false)
             {
-                if (umbracoFile != null)
-                {
-                    target.Name = $"{target.Name} ({umbracoFile})";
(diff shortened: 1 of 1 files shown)


poornimanayar:
Thanks once again @kjac

nul800sebastiaan:
Lovely, thanks for that one! 👍

src/Umbraco.Web/Models/Mapping/EntityMapDefinition.cs
```diff
@@ -186,7 +186,7 @@ private static void Map(ISearchResult source, SearchResultEntity target, MapperC
                 target.Name = source.Values.ContainsKey($"nodeName_{culture}") ? source.Values[$"nodeName_{culture}"] : target.Name;
             }
 
-            if (source.Values.TryGetValue(UmbracoExamineIndex.UmbracoFileFieldName, out var umbracoFile))
+            if (source.Values.TryGetValue(UmbracoExamineIndex.UmbracoFileFieldName, out var umbracoFile) && umbracoFile.IsNullOrWhiteSpace() == false)
             {
                 if (umbracoFile != null)
```
bjarnef, line 191: @kjac the nested if-condition checking umbracoFile not is redundant now with this change 😃

nul800sebastiaan: @bjarnef Duh.. my bad! I removed it but forgot to push before removing it 😂  I'll remove it now, thanks! 👍

nul800sebastiaan: Fixed in https://github.com/umbraco/Umbraco-CMS/commit/e390820f2e1719c54d3521a77a8201b4d5062ed2?w=1 - thanks for noticing @bjarnef! ⭐

PR #7344, opened 2019-12-17
V8: Display MNTP "Allow items of type" correctly when a start node is set

### LINKED ISSUE #7328: Multinode Treepicker (MNTP) allow items of type not persisted on save
When creating a multinode treepicker (MNTP) datatype in 8.4 you are able to select one or more document types via the "Allow items of type" selector. This works. However, if you set a "start node" by picking a node then as soon as you save the datatype the allowed items disappear and are not saved.

## Umbraco version

I am seeing this issue on Umbraco version: 8.4

Reproduction
------------

### Bug summary

The MNTP "Allow items of type" picker doesn't persist selected items upon saving when a start-node is present.

### Specifics

Think this is an 8.4 regression. I have language variants enabled (if that is relevant).

### Steps to reproduce

* Create an MNTP
* Select a start node via the "choose" picker dialogue
* Use the "Allow items of type" selector to select any document type
* Save

As mentioned, language variants are enabled on the site, if that has any bearing. So the start node can be varied by culture.

### Expected result

The items of type selected are persisted.

### Actual result

The items are not persisted.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7328

### Description

If you've set a start node for an MNTP config, the node types in "Allow items of type" aren't displayed correctly. Every so often you might see them flicker into view before they disappear, but most often they'll just appear to be removed:

allow-items-of-type-before

It happens because the "Node type" configuration and the "Allow items of type" configuration are connected via events, so "Allow items of type" automatically resets when the "Node type" changes. Unfortunately this event is emitted twice when a start node is present, forcing an unfortunate reset for "Allow items of type".

This PR refactors the event handling somewhat so things start working again:

allow-items-of-type-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/treesource.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/treesource.controller.js
@@ -19,15 +19,16 @@ angular.module('umbraco')
             };
         }
 
-		if($scope.model.value.id && $scope.model.value.type !== "member"){
-			entityResource.getById($scope.model.value.id, entityType()).then(function(item){
+        if($scope.model.value.id && $scope.model.value.type !== "member"){
+            entityResource.getById($scope.model.value.id, entityType()).then(function(item){
                 populate(item);
-			});
+            });
         }
-
-	    $timeout(function () {
-	        treeSourceChanged();
-	    }, 100);
+        else {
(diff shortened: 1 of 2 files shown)


poornimanayar:
Thanks @kjac

poornimanayar:
And merged :-)

PR #7348, opened 2019-12-17
V8 Hide the Nested Content "add" button until the editor has loaded

kjac:
### Prerequisites

### Description

Just noticed this on 8.4... the Nested Content "Add content" button is always displayed and active, even while the editor is loading (fetching element types). As a result the load indicator is displayed on top of the add button, which looks rather odd:

nc-hide-add-until-loaded-before

This PR fixes it:

nc-hide-add-until-loaded-after


poornimanayar:
@kjac ,

👍

poornimanayar:
Aannndddd merged :-)

PR #7350, opened 2019-12-17
Netcore/feature/ab3791 move scoping to infrastructure and replace callcontext

bergmania:
## NOTE: Build on top of this PR: https://github.com/umbraco/Umbraco-CMS/pull/7309

- Moved the scoping namespace from Core to Infrastructure
- Build our own CallContext replacement.
  - Had to to make it explicit `Guid?` instead of `object` to avoid boxing/unboxing - otherwise test becomes minutes slower.
  - Be aware I also tried using AsyncLocal without any lock.
  - It has to be a `ConcurrentDictionary` otherwise the tests are failing. But I'm really unsure why when we use the thread-id as part of the keys.


src/Umbraco.Tests/Testing/UmbracoTestBase.cs
```diff
@@ -64,6 +65,7 @@ namespace Umbraco.Tests.Testing
     /// provides all the necessary environment, through DI. Yes, DI is bad in tests - unit tests.
     /// But it is OK in integration tests.
     /// </remarks>
+    [Apartment(ApartmentState.STA)]
```
Shazwazza: Is this necessary? This seems to indicate that we will prohibit async threads?

src/Umbraco.Infrastructure/Scoping/CallContext.cs
```diff
@@ -0,0 +1,41 @@
...
+
+
+        /// <summary>
+        /// Retrieves an object with the specified name from the <see cref="CallContext"/>.
+        /// </summary>
+        /// <param name="name">The name of the item in the call context.</param>
+        /// <returns>The object in the call context associated with the specified name, or <see langword="null"/> if not found.</returns>
+        public static Guid? GetData(string name)
+        {
+            return _state.TryGetValue(name + Thread.CurrentThread.ManagedThreadId, out var data) ? data : null;
```
Shazwazza: This isn't going to work as expected, this isn't going to flow at all with async operations and will only work for non-async/singlethreaded operations which means it won't work for any async operations like webapi calls so our Scopes will not be tracked properly and we'll leak transactions, etc... . 

The CallContext.LogicalGetData/LogicalSetData which is what we were using **does** flow with async. Here's the info https://blog.stephencleary.com/2013/04/implicit-async-context-asynclocal.html this link is referenced in our codebase in `HybridAccessorBase`. 

IIRC AsyncLocal is the replacement for CallContext.LogicalSetData/LogicalGetData. I know you've seen a few topics about this such as https://stackoverflow.com/questions/31707362/how-do-the-semantics-of-asynclocal-differ-from-the-logical-call-context and I know you've seen this one too http://www.cazzulino.com/callcontext-netstandard-netcore.html (also note the strongly typed example at the bottom which doesn't require boxing/unboxing).

We will need to use AsyncLocal in place of CallContext since we must have the data flow with async behaviors else our scopes/transactions will not work.

bergmania: I tried using implementation like here http://www.cazzulino.com/callcontext-netstandard-netcore.html but that did not pass our tests..

src/Umbraco.Tests/Testing/UmbracoTestBase.cs
bergmania: I guess not, it was part of my fight with AsyncLocal - thanks.

src/Umbraco.Infrastructure/Scoping/CallContext.cs
Shazwazza: Cool, i'll play around with this today and report back.

Shazwazza:
@Bjarke Berg I've made some significant changes which are all working now and tests passing. 

We now use AsyncLocal to manage the data in a 'call context'. I can't see your original implementation of the CallContext work around using AsyncLocal but I suspect you did the same thing as me when trying it and adding a 'Remove' method to it... which would mean you would end up with tons of errors in the tests saying that the DB is locked and that's because there shouldn't be any Remove method. AsyncLocal itself manages it's own reference to what you put into it. The ConcurrentDictionary just holds a key + AsyncLocal reference and the value within AsyncLocal is managed internally, we never actually want to remove the items/keys from the dictionary. I've added a big note in the new `CallContext<T>` class that you can see.

There was a lot of code in the codebase to manage the quirks of the old `CallContext` and we don't need to worry about these same quirks anymore since AsyncLocal isn't the same as the old CallContext which was orig made for remoting. These work arounds have been in place for quite a few versions (since v7) and the CallContext itself in .NET framework as been 'fixed' in later version of .NET Framework. So while in v7 (net 452) the quirks existed whereas in net 472 they don't (in theory) but in net 472 AsyncLocal exists and it should definitely be used intead plus it supports netstandard.

So I've gone ahead and removed all of these work arounds. We no longer need to store the references that are supposed to go into the CallContext into a weird static collection and then only manage the GUID ids in the CallContext (this is a strange work around). Now we use AsyncLocal and just put the actual objects we want to flow in there and it just works. Lots of code has been removed (a good thing) and it's all a little simpler. It could prob be further simplified but we can handle that later, but it's not a priority.

All tests pass in my local machine, i will wait till the build server passes all tests and then will merge.

Issue #7343, opened 2019-12-16
Change colors on Event Notifications


bmargevicius:
Nothing super important but on ContentService events we have the ability to inject in messages into the backoffice with several "types":

EventMessageType.Default
EventMessageType.Info
EventMessageType.Success
EventMessageType.Error
EventMessageType.Warning

It seems that Info and Success have the same green color. 
I'd like to see the EventMessageType.Info have a different color. I *think* informational messages are often colored some shade of blue? [Bootstrap as an example](https://getbootstrap.com/docs/4.4/components/alerts/)

![Umbraco Message Types](https://user-images.githubusercontent.com/24920023/70943306-89217380-201e-11ea-8b3f-012cce977530.gif)

nul800sebastiaan:
Sounds good! I don't think "blue" is in our current pallet, maybe we can find a similar inconspicuous color that we already use elsewhere that is not pink? :-)

nul800sebastiaan:
Maybe @nielslyngsoe has a suggestion!

nielslyngsoe:
@nul800sebastiaan @bmargevicius @liamlaverty I've added a comment on the PR. 😀
https://github.com/umbraco/Umbraco-CMS/pull/7351

bmargevicius:
I was looking to pr this, I just don't know where to begin to look to make this change. Is there some sort of process to finding things in the CMS?

nul800sebastiaan:
Wow, didn't see there was a PR for it already, thanks @liamlaverty!

@bmargevicius looks like Liam beat you to it! I usually just start by inspecting the div want to change and do a global search in the code. Not the most efficient, but usually gets me where I need to go. Umbraco is a huge app with a lot of legacy code, so it's not always logical to find things.

bmargevicius:
@nul800sebastiaan I know. I was trying to hunt this down, and found classes no longer in use...

umbrabot:
Hi @bmargevicius,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

[...]

PR #7351, opened 2019-12-17
fixes info notifications to appear in blue

liamlaverty:
https://github.com/umbraco/Umbraco-CMS/issues/7343

### Prerequisites

- Added a test to the .spec file

### Description
Update the notifications so that they appear in the standard umbraco info blue

info-fix

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/notifications.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/notifications.service.js
@@ -148,7 +148,7 @@ angular.module('umbraco.services')
                     break;
                 case 1:
                     //info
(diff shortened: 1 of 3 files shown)


nielslyngsoe:
Hi @liamlaverty, I love this change, but we do not use this turquoise color anymore. So could you try one of our blue colors from variables.less, I suggest one of these:

@blueMid:               #2152A3;
@blueDark:              #3544b1;
@blueExtraDark:         #1b264f;

liamlaverty:
Set the variable for `@infoBackground:@blueDark;`, in 284bc4a which will catch the few other places this is used in the cms too. 

info-variable-update

kjac:
Merged 👍

PR #7356, opened 2019-12-18
Netcore/feature/ab4227 move files to infrastructure

bergmania:
## NOTE: Build on top of this PR: https://github.com/umbraco/Umbraco-CMS/pull/7350

- Moved lots and lots of files
- Introduced `IMarchal`. `System.Runtime.InteropServices.Marchal` is not part of netstandard, but exists in both Fullframework and .Net core 3+. So, in theory, this should be each to implement when moved to net core also
- Introduced `IPasswordHasher` and `AspNetPasswordHasher ` that uses `Microsoft.AspNet.Identity.PassworkHasher`
- More dependency injenctions


src/Umbraco.Infrastructure/Persistence/Dtos/DocumentDto.cs
```diff
@@ -7,7 +7,7 @@ namespace Umbraco.Core.Persistence.Dtos
     [TableName(TableName)]
     [PrimaryKey("nodeId", AutoIncrement = false)]
     [ExplicitColumns]
-    internal class DocumentDto
+    public class DocumentDto
```
Shazwazza: How come we have to publicize the DTOs?

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/DocumentRepository.cs
```diff
@@ -21,7 +21,7 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
     /// <summary>
     /// Represents a repository for doing CRUD operations for <see cref="IContent"/>.
     /// </summary>
-    internal class DocumentRepository : ContentRepositoryBase<int, IContent, DocumentRepository>, IDocumentRepository
+    public class DocumentRepository : ContentRepositoryBase<int, IContent, DocumentRepository>, IDocumentRepository
```
Shazwazza: how come we are publicizing our repositories? while others are internal

src/Umbraco.Infrastructure/Sync/DatabaseServerMessenger.cs
```diff
@@ -123,10 +125,12 @@ protected void Boot()
             // the service will *not* be able to properly handle our notifications anymore
             const int weight = 10;
 
-            if (!(_runtime is RuntimeState runtime))
-                throw new NotSupportedException($"Unsupported IRuntimeState implementation {_runtime.GetType().FullName}, expecting {typeof(RuntimeState).FullName}.");
 
-            var registered = runtime.MainDom.Register(
+            //TODO Why do we have interface if we expect to be exact type!!!?
```
Shazwazza: No idea about this one, prob best to just remove the comment and do what makes sense.

src/Umbraco.Web/UmbracoDbProviderFactoryCreator.cs
```diff
@@ -40,5 +42,11 @@ public ISqlSyntaxProvider GetSqlSyntaxProvider(string providerName)
                     throw new InvalidOperationException($"Unknown provider name \"{providerName}\"");
             }
         }
+
+        public void CreateDatabase()
+        {
+            var engine = new SqlCeEngine(DatabaseBuilder.EmbeddedDatabaseConnectionString);
```
Shazwazza: This seems odd since we aren't checking for a provider type here or anything, are we checking what is specified before we call this?

PR #7358, opened 2019-12-18
Change the icon in the Document Types Tree to be the chosen icon instead of a default one.

prjseal:
In the Settings section, Document Types tree we see all document types with the exact same default icon like this

image

In the content type picker that was recently added we can see the icon for the doc type and its colour like this:

image

As a developer / architect of an umbraco site it would be so useful to be able to see which icons and colours have been chosen for a document type at the tree level. It would save us clicking through to each document type to see which icon has been chosen for it. Also it would give continuity to what we have with the content type picker, showing the icon and colour.

My PR fixes this issue and shows us the chosen icon with its colour instead of just a black default doc type icon. Here is what it will look like:

image

### DIFF
--- a/src/Umbraco.Web/Trees/ContentTypeTreeController.cs
+++ b/src/Umbraco.Web/Trees/ContentTypeTreeController.cs
@@ -69,15 +69,18 @@ namespace Umbraco.Web.Trees
                     .OrderBy(entity => entity.Name)
                     .Select(dt =>
                     {
+                        // get the content type here so we can get the icon from it to use when we create the tree node
+                        // and we can enrich the result with content type data that's not available in the entity service output
+                        var contentType = contentTypes[dt.Id];
+
                         // since 7.4+ child type creation is enabled by a config option. It defaults to on, but can be disabled if we decide to.
                         // need this check to keep supporting sites where children have already been created.
                         var hasChildren = dt.HasChildren;
-                        var node = CreateTreeNode(dt, Constants.ObjectTypes.DocumentType, id, queryStrings, Constants.Icons.ContentType, hasChildren);
+                        var node = CreateTreeNode(dt, Constants.ObjectTypes.DocumentType, id, queryStrings, contentType.Icon, hasChildren);
 
                         node.Path = dt.Path;
 
-                        // enrich the result with content type data that's not available in the entity service output
-                        var contentType = contentTypes[dt.Id];
+                        // now we can enrich the result with content type data that's not available in the entity service output
                         node.Alias = contentType.Alias;
                         node.AdditionalData["isElement"] = contentType.IsElement;
 
--- a/src/Umbraco.Web/Trees/ContentTypeTreeController.cs
+++ b/src/Umbraco.Web/Trees/ContentTypeTreeController.cs
@@ -76,7 +76,7 @@ namespace Umbraco.Web.Trees
                         // since 7.4+ child type creation is enabled by a config option. It defaults to on, but can be disabled if we decide to.
                         // need this check to keep supporting sites where children have already been created.
                         var hasChildren = dt.HasChildren;
-                        var node = CreateTreeNode(dt, Constants.ObjectTypes.DocumentType, id, queryStrings, contentType.Icon, hasChildren);
+                        var node = CreateTreeNode(dt, Constants.ObjectTypes.DocumentType, id, queryStrings, contentType?.Icon ?? Constants.Icons.ContentType, hasChildren);
 
                         node.Path = dt.Path;
 
--- a/src/Umbraco.Web/Trees/MemberTypeTreeController.cs
+++ b/src/Umbraco.Web/Trees/MemberTypeTreeController.cs
@@ -23,7 +23,7 @@ namespace Umbraco.Web.Trees
         {
             return Services.MemberTypeService.GetAll()
                 .OrderBy(x => x.Name)
-                .Select(dt => CreateTreeNode(dt, Constants.ObjectTypes.MemberType, id, queryStrings, dt.Icon, false));
+                .Select(dt => CreateTreeNode(dt, Constants.ObjectTypes.MemberType, id, queryStrings, dt?.Icon ?? Constants.Icons.MemberType, false));
         }
     }
 }
--- a/src/Umbraco.Web/Trees/MemberTypeTreeController.cs
+++ b/src/Umbraco.Web/Trees/MemberTypeTreeController.cs
@@ -23,7 +23,7 @@ namespace Umbraco.Web.Trees
         {
             return Services.MemberTypeService.GetAll()
                 .OrderBy(x => x.Name)
-                .Select(dt => CreateTreeNode(dt, Constants.ObjectTypes.MemberType, id, queryStrings, Constants.Icons.MemberType, false));
+                .Select(dt => CreateTreeNode(dt, Constants.ObjectTypes.MemberType, id, queryStrings, dt.Icon, false));
         }
     }
 }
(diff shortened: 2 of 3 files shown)


bjarnef:
What happens if the developer use a folder icon as a repository, archieve or similar for e.g. document type or media type (and not a different color)?

In that case it would looks like a folder in the tree, which it isn't.
Does that mean developers shouldn't be able to use "folder" icons for e.g. document types?

prjseal:
@bjarnef this isn't stopping anyone from using icons, this is showing them which icons they have used. In your scenario above, they could use the folder icon and maybe use a different colour for it to differentiate it from the actual folders.

bjarnef:
@prjseal yes, I don't say it doesn't work, but if they don't use a different color and a folder icon (core folder icon or another folder icon), it might be confusing in the tree because it looks like a folder, but it isn't.

With the content tree it is different, because the developer define the structure. In document, media and member types the "folder" nodes and the icon of these are defined in core.

Imagine the "Article list" was using a folder icon and just default color (black). It would look like the other tree folders, but the behaviour is not the same, e.g. there is no right-click action to create a folder on this document type (which looks like a folder).

For newbies this could be confusing and personally I like the building blocks are neutral.

jacklawry:
Seams to make sense to me… If there is an icon why not use it.

OwainWilliams:
We were actually talking about this in the office today. I think it makes sense. 

I get what @bjarnef is saying thought about the defaults and if you then use the folder icon for something else it could cause confusion. Is it maybe a thought to change the defaults to be something that can't be used by the developer so there is no confusion? 

I guess the problem with changing the defaults though is that a folder works well for the current purpose. It is a difficult thing to get right but I do think changing the icons is a good idea and helps from a visual point of view to find things quicker.

bjarnef:
@OwainWilliams how to ensure developers doesn't use same folder icon? At the moment it is possible to add your own custom font icons, so you can probably always use same icon, e.g. folder icon from Font Awesome package.

Although the folder icon not is exactly the same, it might still cause some confusion.

prjseal:
@bjarnef if this is a problem, why do they use the actual icons on the content type picker? That has been approved and merged into live. This is being consistent with that. It's better to be consistent than to do things differently. That can cause more confusion.

deMD:
I love it. I am a pretty visual person so if I can find the doc type I want to edit by just seeing the icon I become very happy.

OwainWilliams:
> @OwainWilliams how to ensure developers doesn't use same folder icon? At the moment it is possible to add your own custom font icons, so you can probably always use same icon, e.g. folder icon from Font Awesome package.
> 
> Although the folder icon not is exactly the same, it might still cause some confusion.

I was just voicing a thought - not a solution :D  I don't know how you ensure a developer doesn't use a folder icon.

bjarnef:
@prjseal I can't say why it has been approved to use icons in the content type picker, just that there ser potential issues using the selected icons for content types.

The content tree and content type trees (document type, media type and member type are two different things).

In content tree you as developer decide/define the icons and where you want to use folder icons. In content type tree the folder icon is defined in core, so other kind of folder icons shown in the tree could cause confusion. In section picker we don't use icons either.

CarlSargunar:
I really love this idea - making searching for types visually makes it much faster and easier, removes mental load. 

And I think the solutions suggested for folders work too. All for this

ReVoid:
I remember when I develop my very first document types with Umbraco, I wondered why it possible to set an icon, but it doesn't reflect to document type tree like everywhere else? 

I think it is a good idea! DocTypes tree becomes much more intuitive and friendly. 

@bjarnef All you told sound reasonable, but if I have to choice of two evils, I would prefer different icons. May be a compromise would solve the problem, like make all the icons in one color except folder icon. @prjseal What do you think about it?

bjarnef:
@nielslyngsoe has some thoughts of if de should limit the number of colors to use:
https://github.com/umbraco/Umbraco-CMS/issues/7360

Furthermore if we change this, it should also be reflected on media types and member types as well, where the default "folder" media type use a folder icon.

In settings under media types the folder media type will then look like a subfolder, but it isn't. That is where I think it will be inconsistent, because they look similar but have different behaviour.

skttl:
I've been thinking of doing the same. I like it, and it makes it a lot easier to find the desired doctype in a hughe pile of doctypes.

I see the concern of having doctypes with the folder icon clash with doc type containers. But we already allow doctypes without folder-icons to be expandable in the content tree, so I don't think the presence of a folder icon says "expand me" that much more than the little arrow besides it. I don't think folder icon clashes is a problem in this regard.

Rockerby:
I really like this. When making content for the site you use the icon to see what type of content you've created, I think this change ties the two in really nicely.

skttl:
Hi @prjseal 

I think you should do the same thing to Media and Member types for consistency.

prjseal:
@skttl I agree that would be good. What do HQ think? @nielslyngsoe @nul800sebastiaan

nul800sebastiaan:
Agreed on the folder icon, let's not build a special case for now and see how it goes.

Yep, we'd be happy if you did the same for Media/Member Types @prjseal!

greystate:
Thanks for this @prjseal — Can't tell you how many times I've had to cancel building (or prematurely save) a Doctype because I had to go check the existing one's icons and colors first :D 

I have no issues with the folder icon - being able to see icons & colors many times outweigh being confused about a specific icon for a couple of seconds once in a while.

👍

prjseal:
No worries. I’ll finish this PR this week.

prjseal:
@nul800sebastiaan sorry for the delay, I have finished this PR now.
So we have icons for doc types, media types and member types now, as requested in the conversation above.

src/Umbraco.Web/Trees/MemberTypeTreeController.cs
```diff
@@ -23,7 +23,7 @@ protected override IEnumerable<TreeNode> GetTreeNodesFromService(string id, Form
         {
             return Services.MemberTypeService.GetAll()
                 .OrderBy(x => x.Name)
-                .Select(dt => CreateTreeNode(dt, Constants.ObjectTypes.MemberType, id, queryStrings, Constants.Icons.MemberType, false));
+                .Select(dt => CreateTreeNode(dt, Constants.ObjectTypes.MemberType, id, queryStrings, dt.Icon, false));
```
poornimanayar: @prjseal Is there a reason why you are not using the null coalescing here on the icon like you do with the media type? Apologies if I have missed out something..

src/Umbraco.Web/Trees/ContentTypeTreeController.cs
```diff
@@ -69,15 +69,18 @@ protected override TreeNodeCollection GetTreeNodes(string id, FormDataCollection
...
                     {
+                        // get the content type here so we can get the icon from it to use when we create the tree node
+                        // and we can enrich the result with content type data that's not available in the entity service output
+                        var contentType = contentTypes[dt.Id];
+
                         // since 7.4+ child type creation is enabled by a config option. It defaults to on, but can be disabled if we decide to.
                         // need this check to keep supporting sites where children have already been created.
                         var hasChildren = dt.HasChildren;
-                        var node = CreateTreeNode(dt, Constants.ObjectTypes.DocumentType, id, queryStrings, Constants.Icons.ContentType, hasChildren);
+                        var node = CreateTreeNode(dt, Constants.ObjectTypes.DocumentType, id, queryStrings, contentType.Icon, hasChildren);
```
poornimanayar: Same comment as on the `MemberTypeTreeController` applies here @prjseal

poornimanayar:
@prjseal I have tested the work and I must say its a fantastic addition. Thank you as always. Can you get back to us about my comments when you have a minute?

Poornima

prjseal:
@poornimanayar thanks for the review. I've updated the code with your great suggestions.

poornimanayar:
And.................... merged!!!!!  Great work!🥇

bjarnef:
I have submitted a PR to update tree icon on save if icon has changed: https://github.com/umbraco/Umbraco-CMS/pull/8585

Issue #7178, opened 2019-11-19
Embed control is grid layout creates invalid embed url (A potentially dangerous Request.Path value was detected from the client (:))


junkin77:
Replicated on Umbraco versions 8.2.2 and 8.3

Embed video in a Grid Layout and save - can be YouTube video or Vimeo e.g. https://www.youtube.com/watch?v=__G4RrlGmVk . If you view on the site it errors as follows: "A potentially dangerous Request.Path value was detected from the client (:)" [as described here](https://our.umbraco.com/forum/using-umbraco-and-getting-started/99963-a-potentially-dangerous-requestpath-value-was-detected-from-the-client).

The outputted html code looks like this : 
```
<div class="video-wrapper">
	{
  "constrain": true,
  "height": 240,
  "width": 360,
  "url": "https://www.youtube.com/watch?v=__G4RrlGmVk",
  "info": "",
  "preview": "<iframe width="\&quot;320\&quot;" height="\&quot;240\&quot;" src="\&quot;https://www.youtube.com/embed/__G4RrlGmVk?feature=oembed\&quot;" frameborder="\&quot;0\&quot;" allow="\&quot;accelerometer;" autoplay;="" encrypted-media;="" gyroscope;="" picture-in-picture\"="" allowfullscreen=""></iframe>"
}
</div>
```

Existing embedded videos still display ok but adding a new one results in this error.

---
_This item has been added to our backlog [AB#3893](https://umbraco.visualstudio.com/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/3893)_

nul800sebastiaan:
Something seems to have changed in 8.2.0 with returning the stored data, it no longer returns just the html for the raw HTML for embed, but it now returns some json data.

For now, to work around the issue, I think this should do the trick, update `Views\Partials\Grid\Editors\Embed.cshtml` like so:

```csharp
@model dynamic
@using Umbraco.Web.Templates
@{ 
	var embedValue = "test";
	try {
		embedValue = Model.value.preview;
	} catch(Exception ex) {
		embedValue = Model.value;
	}
}

<div class="video-wrapper">
	@Html.Raw(embedValue)
</div>
```

junkin77:
That works. Thanks Sebastian.

warrenbuckley:
@nul800sebastiaan it was this PR where this was changed it seems -  https://github.com/umbraco/Umbraco-CMS/pull/4899

[...]

PR #4899, opened 2019-03-06
v8: Edit embed in grid

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes: https://github.com/umbraco/Umbraco-CMS/issues/4178

### Description
This PR allow to edit an existing Embed grid editor instead of deleting af inserting af new.

PR #7372, opened 2019-12-19
V8: AB3893 Updates the HTML Razor Partial View for grid cell items that use Embed

warrenbuckley:
Updates the HTML Razor Partial View for grid cell items that use Embed

It has been recently updated in 8.2.0 that the Value stored is a JSON object containing width, height & the raw HTML returned by the oEmbed that was done in PR #4899

This fixes reported bug #7178

### DIFF
--- a/src/Umbraco.Web.UI/Views/Partials/Grid/Editors/Embed.cshtml
+++ b/src/Umbraco.Web.UI/Views/Partials/Grid/Editors/Embed.cshtml
@@ -1,7 +1,14 @@
 ﻿@model dynamic
 @using Umbraco.Web.Templates
-
+@{
+    var embedValue = string.Empty;
+    try {
+        embedValue = Model.value.preview;
+    } catch(Exception ex) {
+        embedValue = Model.value;
+    }
+}
 
 <div class="video-wrapper">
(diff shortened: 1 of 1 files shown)


src/Umbraco.Web.UI/Views/Partials/Grid/Editors/Embed.cshtml
```diff
@@ -1,7 +1,14 @@
...
 @using Umbraco.Web.Templates
-
+@{
+    var embedValue = string.Empty;
+    try {
+        embedValue = Model.value.preview;
+    } catch(Exception ex) {
+        embedValue = Model.value;
+    }
+}
```
bjarnef: @warrenbuckley couldn't this be simplified to: 

```
var embedValue = Model.value.preview ? Model.value.preview : Model.value;
```

TimGeyssens:
Just tested on 8.5.3 and get the json instead of the embed code... doesn't seem to have Warren's fix...

nul800sebastiaan:
@TimGeyssens See the related issue #7178 - this is included in the upcoming 8.6.0

TimGeyssens:
@nul800sebastiaan ok sweet :)

PR #7373, opened 2019-12-20
Use the editor icon for data type trees

skttl:
### Prerequisites

### Description
With the same arguments as in #7358 I would propose to use the editor icons in the data type tree. This will make it a lot easier to find data types, and make a better overview of what kind of editors you are using.

billede

To test:
Verify that the data type tree uses different icons for the data types.


kjac:
@skttl this looks great! We'll review and get back to you 🌟

Issue #7374, opened 2019-12-20
Add default id to checkboxes and radiobuttons


bjarnef:
At the moment checkboxes and radiobuttons rendered the `id` attribute with an empty value if not a value `input-id` value is set, where e.g. Chrome warn about this:

![image](https://user-images.githubusercontent.com/2919859/71247459-216a7180-2319-11ea-9f5a-7c03046d0987.png)

https://github.com/umbraco/Umbraco-CMS/blob/bfe359bd1e22856f4560e41056033157e7357c78/src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html#L3

Could we add a default value for `id` if `input-id` is empty?
In tags editor it seems we does domething like this:

https://github.com/umbraco/Umbraco-CMS/blob/990d14e7919b9e802c804f1b6eea1417dc6a39a2/src/Umbraco.Web.UI.Client/src/common/directives/components/tags/umbtagseditor.directive.js#L53

bjarnef:
@nielslyngsoe any thoughts on this? Maybe with a Umbraco prefix and also in tags editor, e.g. `umb_check_{guid}`, `umb_radio_{guid}`, `umb_tag_{guid}`?

nielslyngsoe:
@bjarnef I think it makes sense to create a new GUID if input-id is empty.
We do not need to prefix the GUID, collisions shouldn't be possible (The clever .Net guys say so...)

bjarnef:
It would just make it more clear that it is an internal generated GUID. Furthermore the tags editor has prefix "t", it would be great to have something consistent here. It not already we have another compoent named something with "t", we might have this in future, so "tag" might be better (or "umb_tag" as prefix), but it is a breaking change.

bjarnef:
@nielslyngsoe I have submitted a PR https://github.com/umbraco/Umbraco-CMS/pull/7375

kjac:
Fixed by #7375

PR #7375, opened 2019-12-20
Add component unique ids

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7374

### Description
This PR add a unique id when `input-id` is empty, so it doesn't render empty `id` attributes. I have also added a prefix for these.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbtoggle.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbtoggle.directive.js
@@ -75,7 +75,7 @@
             scope.displayLabelOff = "";
 
             function onInit() {
-                scope.inputId = scope.inputId || String.CreateGuid();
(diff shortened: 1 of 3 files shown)


kjac:
This looks great, @bjarnef - it's been annoying me a lot lately. We'll review this and get back to you 👍

kjac:
Merged 👍

bjarnef:
@kjac I guess the ids can conflict in e.g. `umb-toggle-group` and inside Nested Content, but we can take a further look at it later.

kjac:
@bjarnef indeed. This change is a great step in the right direction and it solves the problem at hand. Should another one arise from IDs within Nested Content that will have to be adressed separately.

bjarnef:
@kjac I haven't cheked this, but the guid might be different when generated in each toggle in toggle group or for each item in Neste Content.

PR #7376, opened 2019-12-20
Property Actions for Property Editor with Controllers

nielslyngsoe:
This PR exposes a reference to umbProperty on the property scope, so Property Editors can access properties on umbProperty. This will allow Property-Editors made with a Controller to add Property Actions.

This PR as well implements a Property Action for Media Picker in multi-mode, to prove that the implementation works.

Screenshot from solution:
image

Test notes:
— Have a Document Type with a Media Picker configured to select multiple media.
— See that Property Actions appear on the Media Picker property.
— Test that the action can be used to clear your picks.
— Test that the action is disabled when the Media Picker is empty.
— Test that picks made after a clear is begin saved and that the correct data is still present after a reload.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyeditor.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyeditor.directive.js
@@ -26,7 +26,7 @@ function umbPropEditor(umbPropEditorHelper) {
                 //we'll also maintain the current form name.
                 scope[ctrl[0].$name] = ctrl[0];
 
-                // We will capture a reference to umbProperty in this Directive and parse it on to the Scope, so Property-Editor controllers can use it.
+                // We will capture a reference to umbProperty in this Directive and pass it on to the Scope, so Property-Editor controllers can use it.
                 scope["umbProperty"] = ctrl[1];
 
                 if(!scope.model.alias){
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyeditor.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyeditor.directive.js
@@ -15,7 +15,7 @@ function umbPropEditor(umbPropEditorHelper) {
                 preview: "<"
             },
             
-            require: "^^form",
+            require: ["^^form", "?^umbProperty"],
             restrict: 'E',
             replace: true,      
             templateUrl: 'views/components/property/umb-property-editor.html',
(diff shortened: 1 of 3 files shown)


src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyeditor.directive.js
```diff
@@ -24,7 +24,10 @@ function umbPropEditor(umbPropEditorHelper) {
                 //we need to copy the form controller val to our isolated scope so that
                 //it get's carried down to the child scopes of this!
                 //we'll also maintain the current form name.
-                scope[ctrl.$name] = ctrl;
+                scope[ctrl[0].$name] = ctrl[0];
+
+                // We will capture a reference to umbProperty in this Directive and parse it on to the Scope, so Property-Editor controllers can use it.
```
warrenbuckley:
```suggestion
                // We will capture a reference to umbProperty in this Directive and pass it on to the Scope, so Property-Editor controllers can use it.
```

src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.controller.js
```diff
@@ -247,6 +248,31 @@ angular.module('umbraco').controller("Umbraco.PropertyEditors.MediaPickerControl
             return true;
         }
 
+        function removeAllEntries() {
+            $scope.mediaItems.length = 0;// AngularJS way to empty the array.
```
warrenbuckley: Can we not just set `$scope.mediaItems = [];` and `$scope.ids = [];` never seen this in our code so far to set the array length to be 0 @nielslyngsoe ?

src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.controller.js
```diff
@@ -247,6 +248,31 @@ angular.module('umbraco').controller("Umbraco.PropertyEditors.MediaPickerControl
...
+            $scope.ids.length = 0;// AngularJS way to empty the array.
+            sync();
+            setDirty();
+        }
+
+        var removeAllEntriesAction = {
+            labelKey: 'clipboard_labelForRemoveAllEntries',
+            labelTokens: [],
```
warrenbuckley: Is the property `labelTokens` required to be set, even if it is an empty array? It would be nicer to omit it surely?

warrenbuckley:
Code mostly looks all fine, with a few minor questions @nielslyngsoe 

I am off to test this out and if so, then I will check the docs you have written and update them if needed to show an example of using the reference to the parent directive/component of `umbProperty` so that the method `$scope.umbProperty.setPropertyActions();` can be called.

warrenbuckley:
This all works fine @nielslyngsoe  :+1:
Will merge this in, but maybe you can look at my comments/suggestions still when your back in 2020 :)

PR #7381, opened 2019-12-22
V8: Remove "double tabbing" on context menu items (regression)

kjac:
### Prerequisites

### Description

#5729 adds some really nice improvements to keyboard navigation through the trees. Unfortunately an unintended side effect is that the tree context menu items now require two tab strokes each:

context-menu-tab-before

With this PR we're back to one per item:

context-menu-tab-after


poornimanayar:
Thanks @kjac , we will have a look and get back to you :-)

nul800sebastiaan:
I can't repro this in the latest `v8/contrib` so I believe it's been fixed somewhere else since then!

Issue #7367, opened 2019-12-19
.NET Core transition: Introduce configuration of IUserPasswordConfiguration and IMemberPasswordConfiguration


bergmania:
**This task is up for grabs, for contributing to the .NET Core transition project**

On the development branch of the .net core transition project [`netcore/dev`](https://github.com/umbraco/Umbraco-CMS/tree/netcore/dev), we introduced new interfaces for password configuration, These interfaces is left with just a default hard coded implementation `DefaultPasswordConfig`.

This task is about introducing configuration values from a config file. Such that these values can be changed without a recompile of the binaries.

It is important that the actual implementation is located in the `Umbraco.Configuration` project

If any questions, feel free to ask :)

JimBobSquarePants:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.iconfiguration?view=dotnet-plat-ext-3.1

You're literally reinventing the wheel.

Shazwazza:
Yep we're fully aware how configuration is supposed to be handled in .net core. We aren't there yet, this is still all transition phase of splitting up projects, making as much net standard as possible before getting the runtimes working with net core. During the transition, the net framework web app needs to continue to work as expected. 
Eventually there will be a new implementation of the configuration project and config interfaces using the desired techniques.

PR #7384, opened 2019-12-23
.NET Core transition: Populated user and member password configuration from settings file (#7367)

AndyButland:
This PR is for #7367 "Introduce configuration of IUserPasswordConfiguration and IMemberPasswordConfiguration".  I've added the necessary settings to `umbracoSettings.config` and amended the configuration loading so they are are read from there.

I wondered about the seeming duplication between `IPasswordConfiguration` and `IPasswordConfigurationSection` (and hence the mapping of fields from one to the other).  It seems like we'd always expect these to have the same values, but being consistent with existing patterns for configuration suggests it's probably best they are separate,

Have checked affected unit tests, and confirmed can still login.  Changing a user's password in the back-office shows the settings are in use.

### DIFF
--- a/src/Umbraco.Web.UI/config/umbracoSettings.config
+++ b/src/Umbraco.Web.UI/config/umbracoSettings.config
@@ -59,10 +59,14 @@
   <security>
     <!-- set to true to auto update login interval (and there by disabling the lock screen -->
     <keepUserLoggedIn>false</keepUserLoggedIn>
+
     <!-- by default this is true and if not specified in config will be true. set to false to always show a separate username field in the back office user editor -->
     <usernameIsEmail>true</usernameIsEmail>
+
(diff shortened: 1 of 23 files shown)


src/Umbraco.Web.UI/config/umbracoSettings.config
```diff
@@ -63,6 +63,12 @@
     <usernameIsEmail>true</usernameIsEmail>
     <!-- change in 4.8: Disabled users are now showed dimmed and last in the tree. If you prefer not to display them set this to true -->
     <hideDisabledUsersInBackoffice>false</hideDisabledUsersInBackoffice>
+    <userPasswordConfiguration
```
Shazwazza: We'll need to update this in the umbracoSettings.Release.config file too (it's hidden away underneath this file, you'll need to click the expand button beside it)

AndyButland: Thanks, have updated.

Issue #7369, opened 2019-12-19
.NET Core transition: Inject service interfaces wherever possible


bergmania:
**This task is up for grabs, for contributing to the .NET Core transition project**

On the development branch of the .net core transition project [`netcore/dev`](https://github.com/umbraco/Umbraco-CMS/tree/netcore/dev), we have used `Current.Services.*` a lot.

When this is used, we cannot move the class into Umbraco.Abstractions or Umbraco.Infrastructure.

This task is about eliminating the usage of `Current.Services.*` in general, by replacing it with a constructor injection of the interface of the given service that is in use.  E.g. `Current.Services.LocalizationService` has to be replaced with injection of `ILocalizationService`.

Expected Procedure:
- Find an instance class that uses `Current.Services.*`.
- Add the interface of the service to the constructor.
- Save the variable in a private readonly field.
- Replace the usages of `Current.Services.*` with the field.
- Make sure the project builds
  - Often you will need to inject this into classes that uses this specific class
- Ensure unit tests all pass
- Repeat the process

If any questions, feel free to ask :)

Feel free to resolve this task in multiple smaller PRs, Due to the fact long running feature branch most likely have merge conflict with [`netcore/dev`](https://github.com/umbraco/Umbraco-CMS/tree/netcore/dev)

hishamco:
I'm an Umbraco fan from version 3 or 4, so I decide to pick this one as my first contribution to the great ecosystem

hishamco:
While this is a huge change I will break this task into many PRs per service

@Shazwazza where should I register the services, I don't see `IStartup` implementation or the web app is not created yet?

hishamco:
- [ ] Inject IAuditService
- [ ] Inject IConsentService
- [ ] Inject IContentService
- [ ] Inject IContentTypeBaseServiceProvider
- [ ] Inject IContentTypeService
- [ ] Inject IDataTypeService
- [ ] Inject IDomainService
- [ ] Inject IEntityService
- [ ] Inject ExternalLoginService
- [ ] Inject IFileService
- [ ] Inject ILocalizationService #7388
- [ ] Inject IMacroService
- [ ] Inject IMediaService
- [ ] Inject IMediaTypeService
- [ ] Inject IMemberGroupService
- [ ] Inject IMemberService
- [ ] Inject IMemberTypeService
- [ ] Inject INotificationService
- [ ] Inject IPackagingService
- [ ] Inject IPublicAccessService
[...]

PR #7387, opened 2019-12-25
Inject ILocalizationService

hishamco:
Address #7369


hishamco:
It's the time to fix few broken unit tests, also I may need to prettify the `ILocalizationService` in test projects

PR #7392, opened 2019-12-25
Inject IContentService

hishamco:
Address #7369


hishamco:
There's still four attributes which using `IContentService`, we may need to refactor them or find some way to inject the service properly

emmaburstow:
Hey @hishamco 

Thanks for picking up this issue for the .net core transition project :D We'll take a look and let you know if we need anything changed. Looks good from here. 

Emma

nul800sebastiaan:
Closing this one since the issue has since been resolved in an alternative way. See my comment here: https://github.com/umbraco/Umbraco-CMS/issues/7369#issuecomment-590262272

PR #6959, opened 2019-10-30
Grid Datatype- Delete Row Configuration - Changed the styles from warning to danger for the confirm delete button, added localization keys

poornimanayar:
In the grid datatype when you try to delete a row config, the overlay uses `warning` style all over. Since the application uses `danger` for deletes I have changed it to `danger` style. I have also added localization keys

**Before**
image

**After**
image

PR #7399, opened 2020-01-02
Changed the styles from warning to danger for the confirm delete button, added localization keys

poornimanayar:
This is the PR https://github.com/umbraco/Umbraco-CMS/pull/6959 re-done as the merge went bad there again.

In the grid datatype when you try to delete a row config, the overlay uses warning style all over. Since the application uses danger for deletes I have changed it to danger style. I have also added localization keys


emmaburstow:
👌

kjac:
All merged in. I'll let this one count as a Hacktoberfest PR, as the original one was indeed from Hacktoberfest.

PR #7400, opened 2020-01-02
Add possibility to override internal hardcoded configs for backoffice search

bielu:
### Prerequisites

### Description
 I had already a few situations when my client wants to be able to search by some custom fields which were in the index but weren't used in the back-office search. I prepare small PR which will allow overriding default hardcoded constants with new one.

### How to test that:
Create some custom composer which will register your CustomInternalSearchConstants which implements IInternalSearchConstants.

` composition.RegisterUnique<IInternalSearchConstants, InternalSearchConstants>();`

Create CustomInternalSearchConstants, example:

```
namespace Umbraco.Web.Search
{
    public class CustomInternalSearchConstants : InternalSearchConstants, IInternalSearchConstants
    {
        public List<string> GetBackOfficeMediaFields()
        {

            var list = base.GetBackOfficeMediaFields();
            list.Add("parentID");
            return list;
        }
    }
}
```
and now search Media by parentId(or any another fields which could be defined instead).


emmaburstow:
Hi @bielu - this looks really helpful. I can think of a few use cases. We'll take a look and let you know if we need any changes.

Emma

Shazwazza:
Hi @bielu one of the reasons we have ISearchableTree is so that you can replace it so you can replace the media implementation of ISearchableTree. Would it make sense to update the the UmbracoTreeSearcher to allow passing in field names for searching? Then you can just replace the media ISearchableTree which uses the same `UmbracoTreeSearcher`?

Just looking to see if we can re-use what we already have without adding new services. But maybe the approach is both of these things since I understand replacing the new service you created is simpler for just modifying fields.

Thoughts?

src/Umbraco.Web/Search/IInternalSearchConstants.cs
```diff
@@ -0,0 +1,13 @@
+using System.Collections.Generic;
+
+namespace Umbraco.Web.Search
+{
+    public interface IInternalSearchConstants
+    {
+        List<string> GetBackOfficeFields();
```
Shazwazza: These all need to be immutable (i.e. IEnumerable<string> or IReadOnlyList<string> or IReadOnlyCollection<string>)

src/Umbraco.Web/Search/IInternalSearchConstants.cs
```diff
@@ -0,0 +1,13 @@
+using System.Collections.Generic;
+
+namespace Umbraco.Web.Search
+{
+    public interface IInternalSearchConstants
```
Shazwazza: I think this could have a friendlier name. Since this is explicit to the UmbracoTreeSearcher this could be something like `UmbracoTreeSearcherFields`

src/Umbraco.Web/Search/IInternalSearchConstants.cs
bielu: Sure I will change them :) to IEnumerable

src/Umbraco.Web/Search/IInternalSearchConstants.cs
bielu: Actually I was trying to find the best name, but I am open for the new name :), Actually good idea with that proposition.

bielu:
yeah, Shannon, we have ISearchableTree but this is a case where we want to add some fields to back-office search, it is overkill, agree we should allow for both options.

src/Umbraco.Web/Search/IInternalSearchConstants.cs
bielu: Changed @Shazwazza

bielu:
@Shazwazza actually I looked now into UmbracoTreeSearcher but I am not seeing how to pass field names as we are using one UmbracoTreeSearcher for all types it is why my service has  
```
IEnumerable<string> GetBackOfficeFields();
IEnumerable<string> GetBackOfficeMembersFields();
IEnumerable<string> GetBackOfficeMediaFields();
IEnumerable<string> GetBackOfficeDocumentFields();
```
as we have actually 3 different hardcoded list of fields, I add one more list BackOfficeDocumentFields to allow add for all 4 types of searchers: All, Media, Members,Documents

src/Umbraco.Web/Search/IUmbracoTreeSearcherFields.cs
```diff
@@ -0,0 +1,13 @@
+using System.Collections.Generic;
+
+namespace Umbraco.Web.Search
+{
+    public interface IUmbracoTreeSearcherFields
```
Shazwazza: We need some code documentation on all of this please

src/Umbraco.Web/Search/UmbracoTreeSearcherFields.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+        private IReadOnlyList<string> _backOfficeMediaFields = new List<string> {UmbracoExamineIndex.UmbracoFileFieldName };
+        public IEnumerable<string> GetBackOfficeMediaFields()
+        {
+            return _backOfficeMediaFields;
+        }
+        private IReadOnlyList<string> _backOfficeDocumentFields = new List<string> ();
+        public IEnumerable<string> GetBackOfficeDocumentFields()
+        {
+            return _backOfficeDocumentFields;
```
Shazwazza: We don't need to allocate anything here, this can just be `return Enumerable.Empty<string>()`

bielu:
@Shazwazza changed. :)

PR #7403, opened 2020-01-02
Fix : Do not open media editing dialog in infinite editing mode when hitting enter in a textbox

poornimanayar:
Currently, when you hit enter in a textbox the media editing dialog opens up in infinite editing mode. This is caused by the edit, delete or add buttons by the media picker on the page getting triggered when you hit enter. This PR aims at fixing that.

**Before**
media-edit-before

**After**
media-edit-fix


kjac:
@poornimanayar thank you for this 👍 excellent GIF'ing too!

Unfortunately this has already been reported and a PR has been submitted - see #7188

poornimanayar:
Ahhh, poor me! I should have thought about that @kjac is in town :D

PR #7406, opened 2020-01-03
V8: Don't allow editing trashed media items from the media picker

kjac:
### Prerequisites

### Description

The media picker shows an edit button even for trashed and deleted items:

image

It makes little sense to edit a trashed item. And if the item is completely deleted, the media editor throws an error and keeps loading indefinitely: 

image

This PR removes the media picker edit button for trashed (and deleted) media items:

image

...and it works for both single and multi modes:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.html
@@ -36,7 +36,7 @@
                 </umb-file-icon>
 
                 <div class="umb-sortable-thumbnails__actions" data-element="sortable-thumbnail-actions">
-                    <button aria-label="Edit media" ng-if="allowEditMedia" class="umb-sortable-thumbnails__action btn-reset" data-element="action-edit" ng-click="vm.editItem(media)">
+                    <button aria-label="Edit media" ng-if="allowEditMedia && !media.trashed" class="umb-sortable-thumbnails__action btn-reset" data-element="action-edit" ng-click="vm.editItem(media)">
                         <i class="icon icon-edit" aria-hidden="true"></i>
                     </button>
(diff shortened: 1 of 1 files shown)


emmaburstow:
👍

abjerner:
Hi @kjac,

I agree that the edit button shouldn't be there for media that has been deleted, as there is no way to edit a deleted item.

Not that this necessarily has something to do with your PR, but I think there is a bit of an UX issue for media that has been trashed, but not yet deleted.

Say I'm an editor coming across a page with a trashed media, I can currently click to open the media, and see some more details about. But I can't restore or delete it from infinite editing, and the breadcrumb doesn't mention the recycle bin (which are issues on it's own, not because of this PR). By removing the edit button, the editor will have very little clue about the trashed media. So until the UX is improved around this part, I feel it may be wrong to remove the edit button for trashed media.

Any thoughts?

kjac:
@abjerner I think I kinda get your point. The infinite editing of trashed media isn't the greatest solution to find the trashed media though, even if it does provide more context to the media than the media picker itself. 

Thing is we can't tell trashed and entirely deleted media apart with the properties currently available to us. Thus the 404 error can't be fixed unless we sacrifice the edit button for both trashed and entirely deleted media.

abjerner:
@kjac I know it's not the greatest solution, but still something I think it's worth considering.

Regarding telling trashed and deleted media apart, couldn't we set a `deleted` property here:

https://github.com/umbraco/Umbraco-CMS/blob/67168068f63429cedb07251b2f3a6d4938d1d23b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.controller.js#L67

The name is already set to `vm.labels.deletedItem`, so there is some "delete handling" going on here.

nul800sebastiaan:
@kjac Thanks! Merged something similar from @bjarnef the other day, sorry I didn't remember this one was still open 🙈

PR #7407, opened 2020-01-03
Inject IUser service

hishamco:
Address #7369

### DIFF
--- a/src/Umbraco.Web/PublishedCache/NuCache/ContentNode.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/ContentNode.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Diagnostics;
 using Umbraco.Core.Models.PublishedContent;
+using Umbraco.Core.Services;
 using Umbraco.Web.PublishedCache.NuCache.DataSource;
 
 namespace Umbraco.Web.PublishedCache.NuCache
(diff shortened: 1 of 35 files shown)


hishamco:
Again there 're few places such as attributes & extension methods, that need to refactor

@Shazwazza is there a way to use `IFactory` in the extension methods that you 're using? What about the attributes, coz they 're trick :)

src/Umbraco.Web/PublishedCache/NuCache/ContentNode.cs
```diff
@@ -149,7 +150,7 @@ private IPublishedContent GetModel(ref IPublishedContent model, ContentData cont
             // more than 1 instance, but the lock below ensures we only ever return
             // 1 unique instance - and locking is a nice explicit way to ensure this
 
-            var m = new PublishedContent(this, contentData, _publishedSnapshotAccessor, _variationContextAccessor, _publishedModelFactory).CreateModel(_publishedModelFactory);
+            var m = new PublishedContent(this, contentData, _publishedSnapshotAccessor, _variationContextAccessor, _publishedModelFactory, null).CreateModel(_publishedModelFactory);
```
hishamco: I set this to `null` for now, because I think the better way here is to inject `PublishedContext` from DI instead of newing up here

hishamco:
I think I need to revise the failed unit tests

nul800sebastiaan:
Closing this one since the issue has since been resolved in an alternative way. See my comment here: https://github.com/umbraco/Umbraco-CMS/issues/7369#issuecomment-590262272

Issue #7409, opened 2020-01-03
Default-value of a Checkbox-property is not used in YesNoValueConverter


jerpenol:
Hi,

I am seeing this issue on Umbraco version: 7 & 8 

If I add a Checkbox-property with a default value to an already existing and in-use Doctype, the default value is not used as a fallback in the ValueConverter of the property.

This is not a problem when the default value of the data type is ``` false ```, because in the ```Umbraco.Core.PropertyEditors.ValueConverters.YesNoValueConverter``` it ultimately (when ```source``` is ```null```) falls back to ``` false ```. However when the default value is set to ``` true``` the ValueConverter still defaults to ``` false ```, which is not expected.

The solution I see is adding some logic to the ```Umbraco.Core.PropertyEditors.ValueConverters.YesNoValueConverter``` which falls back to the default value of the data type when ```source``` is ```null```.

kjac:
Hi @jerpenol,

This is by design and how Umbraco works. If you add new properties to your content types, you need to apply some defensive coding around their usage, as the content based on these content types won't be given any default values for these properties. 

On the upside this means that you *know* if an editor has actually filled out your property. This is useful in some cases, and thus adding a fallback to the datatype default value will be a behavioral breaking change. 

You can use `Model.HasValue(string alias)` to figure out if a property has been assigned a value. Get in touch with the people on [the forum](https://our.umbraco.com/forum/) if you need any help with this.

jerpenol:
Hi @kjac ,

Thanks for your feedback.

My problem is I cannot know when a checkbox is not filled out, because it defaults to ```false``` in the ValueConverter as of now.
So this means the default value of the Checkbox is only to be used when creating new content? Because when I add the property after it is already in use, existing content is not modified at the moment, which renders the checkbox useless with existing content (unless you save your content one-by-one).

nul800sebastiaan:
Thanks Kenn, that's indeed how it works. :-)

PR #7411, opened 2020-01-03
yesno-property-with-default-value (issue #7409)

jerpenol:
Hi,

**References**
Issue #7409 

**Problem**
If I add a Checkbox-property with a default value to an already existing and in-use Doctype, the default value is not used as a fallback in the ValueConverter of the property.

This is not a problem when the default value of the data type is ```false```, because in the ```Umbraco.Core.PropertyEditors.ValueConverters.YesNoValueConverter it ultimately``` (when ```source``` is ```null```) falls back to ```false```. However when the default value is set to ```true``` the ValueConverter still defaults to ```false```, which is not expected.

**Solution**
What I did to solve this was, I added logic to the   ```Umbraco.Core.PropertyEditors.ValueConverters.YesNoValueConverter``` that gets and falls back to the default value of the data type when ```source``` is ```null```.

Since there is no ```TrueFalseConfiguration``` in ```Umbraco.Core``` (only in ```Umbraco.Web``` which I cannot reference from within ```Umbraco.Core```) I used ```dynamic``` to get the default value of the data type. I know ```dynamic``` is a bit "filthy", so a cleaner way to get the default value would be neat!

### DIFF
--- a/src/Umbraco.Core/PropertyEditors/ValueConverters/YesNoValueConverter.cs
+++ b/src/Umbraco.Core/PropertyEditors/ValueConverters/YesNoValueConverter.cs
@@ -22,6 +22,12 @@ namespace Umbraco.Core.PropertyEditors.ValueConverters
             // typically the converter does not need to handle anything else ("true"...)
             // however there are cases where the value passed to the converter could be a non-string object, e.g. int, bool
 
+            if (source == null)
+            {
+                // if source is null fall back to default-value of data type
+                source = propertyType?.DataType?.ConfigurationAs<dynamic>()?.Default;
+            }
+            
             if (source is string s)
             {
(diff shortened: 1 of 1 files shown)


kjac:
Hi @jerpenol,

Thank you for this 😄 

Apart from being "filthy" as you phrase it, the use of `dynamic` is generally speaking also synonymous with bad performance, so it should be avoided at all cost.

I also daresay that this change is unlikely to be accepted in any form, even if a way around the `dynamic` can be found. See my comment [here](https://github.com/umbraco/Umbraco-CMS/issues/7409#issuecomment-570564319).

kjac:
With #7409 being closed I'm closing this one as well.

PR #7412, opened 2020-01-03
v8: Fix for login screen title display

RachBreeze:
As discussed with @kjac on https://github.com/umbraco/Umbraco-CMS/pull/7106 the current login screen prepends "Login" to the item being edited when a user is logged out of the system:
image
_screen shot 1_
Then when the user logins they still see login prepended :

image
_screen shot 2_

This PR refactors the calculation of the page title allowing two properties to be used:
-  `pageTitle` when setting the page title in the routine $changeTitle on init.js, this is an existing routine and should not change any existing page title calculations
-  `rootTitle` when setting the page title in the routine $setRootTitle on init.js, this is a new routine and is only currently used in the umblogin.directive.js, it allows the title "Login" to be displayed without any content prefixes.  So rather than _screen shot 1_ the user sees
image
_screen shot 3_
This should provide the screen reader user better understanding of what the page is for

And finally when a user logs in they no longer see "login prepended" to the page title


poornimanayar:
Thank you for the work @RachBreeze . We shall review it and get back to you.

RachBreeze:
Hiya how did the review go?  Is there anything I can do to assist?

poornimanayar:
Hey @RachBreeze ,

I gave this a quick spin and this is what I can see

- Go to /umbraco I see Login in the title
- Login to the backoffice, go edit some pages and login is not prepended
- Now when I log back out shouldnt the title say Login again? Its saying Umbraco. Apologies if I misunderstood the work.

Can you merge the latest `v8/contrib` into your branch if you are testing this your end? I did that and saw the above results.

Poornima

RachBreeze:
Hi @poornimanayar 
I've spent a while checking this out, and I can reproduce but I'm struggling to fix, when the user logs out the code to set the title is fired, and then redirects to the login page (it does the same thing when login fails).  The redirect isn't hitting the `UmbLoginController` initialization events.  I don't believe it's possible to resolve this issue :-(
The good news is I've tided the code :-)

poornimanayar:
Hi Rachel,

Thats fine! What we have now is clearly an improvement.

Poornima

nul800sebastiaan:
Hi @RachBreeze - I can see two whole new files (or they look new) `src/Umbraco.Web.UI/Umbraco/js/navigation.controller.js` / `src/Umbraco.Web.UI/Umbraco/js/main.controller.js` in this PR. Do you know where they came from? It feels like they should not have been committed?

RachBreeze:
Hi @nul800sebastiaan  I worked on v7 and v8 on April 10th, those files sound like v7 and shouldn't have been committed

nul800sebastiaan:
@RachBreeze Ahaaa! I like to check out the v8 and the v7 branch in 2 different directories. Switching between v7 and v8 is prone to errors, it's a nightmare to rebuild the gulp stuff and you get stray files as you've noticed.

nul800sebastiaan:
Thanks @RachBreeze, this works as described. There was a lot of changes of code along the way, so in the end we only ended up with a few lines of code changed. You might want to have a second look at it to see that this is what you meant to do. This has been merged though and an automatic logout due to a timeout gives the behavior as you described above, so I think this is all good! 👍

RachBreeze:
Hi @nul800sebastiaan 
Thank you that's a great shout on v7 and v8 thank you. I felt like the first PR was overly complicated hence the rewrite in the follow up

PR #7426, opened 2020-01-07
Netcore: Move indentity struff from core to web

bergmania:
## Based on this PR, that need to be merged first: https://github.com/umbraco/Umbraco-CMS/pull/7425

Moved classes and dependencies from core to web, that uses Identity.


src/Umbraco.Infrastructure/Composing/CompositionExtensions/FileSystems.cs
```diff
@@ -1,9 +1,12 @@
...
+﻿using Umbraco.Core.Composing;
+using Umbraco.Web.Composing;
+using Umbraco.Core.Configuration;
 using Umbraco.Core.IO;
 using Umbraco.Core.IO.MediaPathSchemes;
 using Umbraco.Core.Logging;
+using Umbraco.Core;
 
-namespace Umbraco.Core.Composing.CompositionExtensions
+namespace Umbraco.Web.Composing.CompositionExtensions
```
Shazwazza: Any reason this namespace is changed?

src/Umbraco.Infrastructure/Composing/CompositionExtensions/Repositories.cs
```diff
@@ -1,7 +1,9 @@
-﻿using Umbraco.Core.Persistence.Repositories;
+﻿using Umbraco.Core;
+using Umbraco.Core.Composing;
+using Umbraco.Core.Persistence.Repositories;
 using Umbraco.Core.Persistence.Repositories.Implement;
 
-namespace Umbraco.Core.Composing.CompositionExtensions
+namespace Umbraco.Web.Composing.CompositionExtensions
```
Shazwazza: This namespace change seems odd, this has to do with Repositories which are most definitely at the core level

src/Umbraco.Infrastructure/Composing/CompositionExtensions/Services.cs
```diff
@@ -11,7 +13,7 @@
 using Umbraco.Core.Services;
 using Umbraco.Core.Services.Implement;
 
-namespace Umbraco.Core.Composing.CompositionExtensions
+namespace Umbraco.Web.Composing.CompositionExtensions
```
Shazwazza: Also an odd namespace change?

src/Umbraco.Infrastructure/Models/Membership/User.cs
```diff
@@ -322,7 +322,7 @@ public string Language
         }
 
         [IgnoreDataMember]
-        internal bool DefaultToLiveEditing
+        public bool DefaultToLiveEditing
```
Shazwazza: Wow blast from the past, pretty sure this can be completely deleted we don't have 'live editing' since like v4 or something

src/Umbraco.Infrastructure/Composing/CompositionExtensions/Configuration.cs
```diff
@@ -1,6 +1,8 @@
-﻿using Umbraco.Core.Configuration.UmbracoSettings;
+﻿using Umbraco.Core;
+using Umbraco.Core.Composing;
+using Umbraco.Core.Configuration.UmbracoSettings;
 
-namespace Umbraco.Core.Composing.CompositionExtensions
+namespace Umbraco.Web.Composing.CompositionExtensions
```
Shazwazza: Is this namespace change meant to be done?

src/Umbraco.Infrastructure/Models/Membership/User.cs
bergmania: You are right.. It is always set to false.. And never read (in backend code atleast)

bergmania:
You are right about the namespaces should not have been updated from `Umbraco.Core.Composing.CompositionExtensions` to `Umbraco.Web.Composing.CompositionExtensions`

PR #7427, opened 2020-01-07
Netcore: Move SqlCe classes to separate project

bergmania:
## Based on this PR, that need to be merged first: #7426

- Moved SqlCe classes to separate project running net472 - The `.csproj` is using the new structure
  - `SqlCeBulkSqlInsertProvider`
  - `SqlCeSyntaxProvider`
- Deleted unused files from Umbraco.Core (`FileResources/*`)
- Added references to the new project when needed
  - Umbraco.Web
  - Umbraco.Tests
  - Umbraco.Tests.Benchmarks


Shazwazza:
Just trying to run this locally now - from scratch without a DB and cannot run it, will debug this now.

Shazwazza:
@bergmania there were a couple problems (from previous PRs) preventing the site from starting which I've fixed up (as part of this PR) in these revs:

* https://github.com/umbraco/Umbraco-CMS/pull/7427/commits/c18b6e6c82ac5f5ed7e905add1f748ef91b6feec
  * Have moved the creation of `IBulkSqlInsertProvider` to the `IDbProviderFactoryCreator` which means we no longer eagerly create this and it's not the bootup code's responsibility (`UmbracoApplication`), this also simplifies a bunch of ctors. I think it makes sense but let me know and we can change stuff later.
* https://github.com/umbraco/Umbraco-CMS/pull/7427/commits/11f5d2f80022d5c99ee24a711ee3cdf51c909174
  * Not initializing the `Current.Factory` before components are initialized bit us, but that's a good thing since `Current` should defo not be used in components so I've fixed the component that had the issue

All works now and I can install a new site with SqlCe. will merge.

PR #7430, opened 2020-01-07
Clean up EnumExtensions

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6691.

### Description
As discussed in PR https://github.com/umbraco/Umbraco-CMS/pull/6693, the new `SetFlag()` and `UnsetFlag()` enum extension methods are not necessary, as this can be done with bitwise operations way more efficiently (more than 10.000 times faster, without memory allocations, boxing/unboxing, converting to `ulong` or requiring garbage collection).

Logic operators (AND, OR, XOR and NOT) are a basic programming concept, although it might not be understood by everyone... I've added comments for this, as that helps you understand the code without:
- Adding more code to maintain
- Extend the public API (that's specific to Umbraco)
- Require documentation for the code
- Writing tests to verify the code
- All mentioned performance problems (especially down the road, if it gets used in performance critical code)

To avoid making this a breaking change, this will need to be merged before Umbraco 8.6.0 is released. @Shazwazza: you merged PR #6693, what's your take on this?


ronaldbarendse:
I've also obsoleted the existing `HasFlagAll()` extension method, as this does exactly the same as the [`HasFlag()`](https://docs.microsoft.com/en-us/dotnet/api/system.enum.hasflag) method or using a bitwise operation (`(value & flag) == flag`).

nul800sebastiaan:
As discussed on https://github.com/umbraco/Umbraco-CMS/pull/6693, we'll keep the extension methods around.

ronaldbarendse:
@nul800sebastiaan I've updated my branch to use non-generic extension methods for `ContentVariation`, so you don't get the performance problems of boxing/unboxing generic types and Umbraco doesn't add unnecessary global extension methods: https://github.com/ronaldbarendse/Umbraco-CMS/commit/c38240c872428b366cd716c34c8ce468ced91271.

In my opinion, this is the best way to keep the pros of the extension methods, without the cons of performance/namespace pollution and aligns with the comment from @Shazwazza: https://github.com/umbraco/Umbraco-CMS/pull/6693#issuecomment-571807022.

ronaldbarendse:
I've also added the `VariesBySegment()` and `VariesByCultureAndSegment()` extension methods for `ISimpleContentType`, as those didn't yet exist: https://github.com/umbraco/Umbraco-CMS/commit/05f7d24a9a4b8bfea675b70243c396408e7cfce5. Can this be re-opened or should I create a new PR @nul800sebastiaan?

nul800sebastiaan:
@ronaldbarendse Weird! I don't see your changes showing up on this PR.. maybe because it's closed? I'll reopen and see what happens.

nul800sebastiaan:
Yep! That did the trick! I'll hand it over to the team.

nul800sebastiaan:
@ronaldbarendse Unfortunately it looks like that broke the build though, could you have a look?

src/Umbraco.TestData/SegmentTestController.cs
```diff
@@ -50,7 +50,7 @@ public ActionResult DisableDocTypeSegments(string alias)
             if (!ct.VariesBySegment())
                 return Content($"The document type {alias} does not allow segments, nothing has been changed");
 
-            ct.Variations = ct.Variations.UnsetFlag(ContentVariation.Segment);
+            ct.Variations = ct.Variations.SetVariesBy(ContentVariation.Nothing);
```
nul800sebastiaan: @ronaldbarendse Alright I've fixed the build now, I had to also update this file. This particular change I am not 100% sure about, in this method we're trying to set the variation to `Nothing`, is that correct?

And general note, it would be good if you could compile before committing next time.

ronaldbarendse:
Yes, looks like @PerplexDaniel added the `VaryesBySegment` overload and this commit is just merged: https://github.com/umbraco/Umbraco-CMS/commit/45db8fd7. I've updated this PR with the latest changes and fixed the build.

I've also renamed the `SetVariesBy()` extension method on the enum overload, as it was too easy to write `contentType.Variations.SetVariesBy(..)` and not actually setting the variations on the content type (as this method returns the changed enum value). So this extension method is renamed to `SetFlag()` and accepts a boolean to set/remove the flag...

src/Umbraco.TestData/SegmentTestController.cs
ronaldbarendse: `Nothing` is not a flag, because the value is `0` (this is the same as expecting `1 - 0 = 0`, where 1 is the current flag and 0 the flag you're trying to remove). Nonetheless, it should not set the variation to `Nothing`, but only remove the `Segment` flag. I've updated this PR to do exactly this 👍

Shazwazza:
Looks good to me, cheers!

PR #7432, opened 2020-01-08
Netcore: Handle IRegisteredObject and remove Umbraco.Core project

bergmania:
## Based on this PR, that need to be merged first: #7427

- Implemented our own `IRegisteredObject` and `AspNetHostingEnvironment` wraps the (Un)RegisterObject using a private wrapper class.
- Updated the following to use our own `IRegisteredObject` 
  - `MainDom`
  - `PureLiveModelFactory`
  - `IBackgroundTaskRunner` and `BackgroundTaskRunner`
- Moved `MainDom` to `Umbraco.Infrastructure`
- Removed the `Umbraco.Core` project


src/Umbraco.Web/Hosting/AspNetHostingEnvironment.cs
```diff
@@ -44,6 +47,23 @@ public void LazyRestartApplication()
             HttpRuntime.UnloadAppDomain();
         }
 
+        private IDictionary<IRegisteredObject, RegisteredObjectWrapper> RegisteredObjects { get;  } = new Dictionary<IRegisteredObject, RegisteredObjectWrapper>();
```
Shazwazza: Why isn't this just a private readonly field and should probably be ConcurrentDictionary since we want to make sure this is 100% threadsafe.

bergmania:
I updated the code

Shazwazza:
works great!

PR #7433, opened 2020-01-08
V8: Fix for "no search results" text shown on top of the search input field

kjac:
### Prerequisites

### Description

If you search for a page in the linkpicker (e.g. from an RTE) and the page can't be found, the "no search results" text is shown rather clumsily on top of the search input field:

image

It happens because the search results directive is hardcoded to display the empty state at position `center`. This PR lets us specify a position for the empty state (it still defaults to `center`), and thus the empty state for the linkpicker can be positioned better:

image


emmaburstow:
👍

nul800sebastiaan:
Perfect! Thanks again! 👍

PR #7435, opened 2020-01-08
Git ignorer files in src/Umbraco.Web.UI/data

nielslyngsoe:
Adding ignorer for "src/Umbraco.Web.UI/data"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -163,3 +163,4 @@ build/temp/
 # eof
 /src/Umbraco.Web.UI.Client/TESTS-*.xml
 /src/ApiDocs/api/*
+src/Umbraco.Web.UI/data


clausjensen:
I've added this without the whitespace changes :)
https://github.com/umbraco/Umbraco-CMS/commit/99bc508b68ad70dd2df601cc4096d621f3a64b00

nul800sebastiaan:
We really shouldn't be adding random folders to the ignore file. We don't generate files in the `/data` folder from CMS and if there are files there then you should be aware as to WHY they are there. The danger of ignoring things is that you suddenly have a folder with files that does something useful and in 3 months we're left puzzling as to why the behavior on your computer is so much different than on other people's computers. 

In this case, I understand that Niels will now forever have Deploy installed in his local copy of the source code, which means he will see deviant behavior that nobody else will see.

I guess this is fine for now but please be aware that you are deliberately creating a different version of the CMS source on your local machine.

clausjensen:
@nielslyngsoe just in case you didn't get Sebastiaans point here 😄 :
What he's saying is that instead of needing this ignore, you should remove Deploy from your Umbraco source when you don't need it there (and then you wouldn't have the data folder either).

You are no longer working on a "vanilla" Umbraco installation and functionality/features/errors you could be seeing (or not seeing) could be tainted by the fact that you are actually using Umbraco + Deploy and not only Umbraco.

So... as an Umbraco CMS developer you would normally not want to run and develop on the source _with Deploy in there as well_ unless you are actually working on a Deploy feature.

nielslyngsoe:
@clausjensen I got that.

PR #7436, opened 2020-01-08
V8: Fix JS error when trying to insert a link in an empty RTE

kjac:
### Prerequisites

### Description

If you attempt to insert a link in an empty RTE, a JS error is thrown:

empty-rte-linkpicker-error-before

Turns out `getAnchors.getAnchors(...)` has varying return types; it returns an empty array if the RTE is empty and a promise if the RTE is _not_ empty. 

This PR ensures that `getAnchors.getAnchors(...)` always returns a promise, and thus the linkpicker no longer throws an error:

empty-rte-linkpicker-error-after

_Note: Inserting a link in the RTE without a selection results in a sort of weird behavior where the link text is empty. This should probably be addressed in another PR as it affects non-empty RTEs as well._


emmaburstow:
👍

nul800sebastiaan:
Works like a charm, cheers @kjac!

PR #7442, opened 2020-01-09
#7441 Make backoffice CSS/LESS DRY(er) => box-shadows

### LINKED ISSUE #7441: Make CSS/LESS more DRY
Looking through the backoffice front-end project, we have a lot of repeated CSS rules (box-shadows being the one that started me on this issue). I'd like to spend some time refactoring existing CSS to use mixins/variables instead of repeating property values, which becomes error prone and harder to maintain.

As an example, I've refactored the healthchecks dashboard LESS file (purely because it appeared in a particular box-shadow search). Changes improve the structure of the file to namespace the component with sensible nesting, and uses a couple of class mixins to apply box-shadow (including hover state and transitions).

Will create a PR with the changes, for discussion before I rip through any others. Color and type are already pretty good, but there's lots of room for improvement by replacing repetition with generated styles. Happy to chip away at this

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #7441 

### Description

Changes here are an example of what I'd like to do to start tidying up some of the backoffice LESS files - removing duplication across files by using class mixins instead. Example here uses a box-shadow class rather than hardcoding the rule (this particular box-shadow value is used 20+ times)

Not wanting to merge this just yet, wanted to show my intentions before chopping up lots of files... 😄


emmaburstow:
Hi Nathan,

Thanks for the pr - nice bit of refactoring! I can see it's a draft so will hold off reviewing.

Em

nathanwoulfe:
Thanks @emmaburstow I'll finish off the box-shadow stuff here, and will add separate prs for similar changes rather than bloating this into a behemoth.

src/Umbraco.Web.UI.Client/src/less/components/application/umb-drawer.less
```diff
@@ -7,50 +7,46 @@
...
-    box-shadow: inset 5px 0 20px rgba(0,0,0,.3);
-}
+    box-shadow: inset 5px 0 20px rgba(0, 0, 0, .3);
 
-.umb-drawer-view {
-    display: flex;
-    flex-direction: column;
-    height: 100%;
-    overflow: hidden;
+    &-view {
```
skttl: I don't like this. This is going to make it a lot harder to find .umb-drawer-view in the sourcecode.

nathanwoulfe: Yup, fair call - I may have gone a bit heavy-handed. Have changed this to have separate classes for .umb-drawer-view, .umb-drawer-content and .umb-drawer-footer, to be grep-friendly. 

Makes better sense to split on the BEM modifier:
```
 .umb-group-panel {
    // stuff
    &__header { 
        // more stuff
    }
}
```

I still feel though there's benefit to embracing a degree of nesting in the LESS files, particularly as the project structure improves (ie component per file, sensibly namespaced etc).

nathanwoulfe:
There's a heap of different shadows in use throughout the backoffice - wondering if there's scope to standardise these using an approach similar to Angular Material's (https://material.angular.io/guide/elevation)[elevation helpers], where a set of mixins are used to create z-axis separation between elements.

I'm seeing a lot of shadow definitions which are pretty similar, so might be able to be merged/combined. @nielslyngsoe might have an opinion?

nielslyngsoe:
Hi @nathanwoulfe 

Good initiative, I think it's great to get shadows aligned. But I think you should know we aren't using the Bootstrap Shadows that much. It's much more common that custom implementation is made. That said, I think we should aim to align those. So we might need to re-think how would we like Shadows, how many, what names make sense? Like, thinking about the Developer Experience in picking a shadow?

I only think we need a few, at least it would be nice to try to keep it to a bare minimum. At many places, the shadows are a bit different than other shadows, but from a visual point, they don't need to be different.

I hope my thoughts make sense to you and I'm looking forward to seeing progress on this.

BTW. you could start out making the framework for shadows, and then I can help out implement those. :-)

nathanwoulfe:
Sounds good to me. What I've done so far is take the existing box-shadow utility classes (.shadow-level-1 through .shaow-level-5) and created a mixin for using these  in CSS (.umb-box-shadow(n) where n is the level). I'll keep chipping away at standardising what I can. 

Five levels of shadow depth is probably plenty, and can add additional parameters to the mixin eg colour to make it more useful. 

Ideally devs would never need to think about the implementation of things shadows, border radius etc but instead apply a mixin. Makes it easier to change things in the future by tweaking variables.

nielslyngsoe:
@nathanwoulfe sounds good. I do think we aswell should make shadows available as classes, I think there are a few developers out there who like the ability to just add a class to get shadows. How do you feel about that?

nathanwoulfe:
@nielslyngsoe absolutely agree - I'm using the mixin to generate the .shadow-depth-x classes too, so devs can either use the class in HTML, or the mixin in LESS.

Borders are another one worth looking at - mixin for thin, mid, thick, axes, color, to save too much thinking for devs.

nielslyngsoe:
@nathanwoulfe love it, looking forward to see your approach.
Please keep in mind that we properly have options that can be removed, to make the code and choices simpler :-)  Like we might only need 3 shadows? meaning some shadows should be adjusted to fit with our choice. Cause there might be 20+ presented today.

PR #7444, opened 2020-01-09
added round corners in umb-media-grid for media-picker

nielslyngsoe:
Added round corners on items in media-picker.

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-media-grid.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-media-grid.less
@@ -19,11 +19,19 @@
     align-items: center;
     align-self: stretch;
 
+    border-radius: @baseBorderRadius;
+
     margin: 10px;
     position: relative;
     user-select: none;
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
@nielslyngsoe As discussed here: https://github.com/umbraco/Umbraco-CMS/pull/8686#issuecomment-680700097

Is rounded corners something you want or don't want? ;-)

nielslyngsoe:
@nul800sebastiaan I know exactly what I want, I want consistency and a friendly look. 🤩

To answer your concern: I would like round corners on these as that unifies the look of 'cards', makes them more friendly.

bjarnef:
@nielslyngsoe in that case it would be great to update all pickers "add" buttons to use rounded corners as well. I think would should have a New class instead of `umb-node-preview-add` but keep this for legacy support.

In the HTML pickers will use this class or in less extend from this class, so we only need to update this one place.

nielslyngsoe:
@bjarnef I think that sounds like a good plan, it sounded like you had a good idea about how to go about it, so feel free to make your suggestion as a draft PR. then we can finish it together?

PR #7446, opened 2020-01-10
V8: Disable the actions menu button until there are actions available

kjac:
### Prerequisites

This PR replaces #7312 

### Description

As discussed on #7312, the  actions menu button in the editor header shouldn't be enabled until there are any actions to show. 

This PR implements that. When applied, the actions menu button is disabled and subtly dimmed down until the actions for the current content loads:

disable-actions-button-until-actions-load

If no actions are loaded, the actions menu button simply remains disabled and dimmed down.


emmaburstow:
👍

PR #7448, opened 2020-01-11
use default color when previous selection removed

### LINKED ISSUE #7445: If picking no icon color for DocumentTypes it should save the default color
If the color of a DocumentType is removed it should be set to the default color.
Currently, if you de-select your current color and save, you will still have the same color as before.

image

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #7445 

### Description
I've added a property to the colors array to flag a color as the default - when the existing selection changes (and when setting the default if no value exists), we now grab the default color from the colors array. Went this way rather than assuming the first entry is the default, and to decouple the configuration from the implementation.

Also removed a usage of the underscore library, replacing it with vanilla js (`_.findWhere()` becomes `Array.find()`)

To test, edit a document type icon. Change the color to red, submit, save, color is persisted. Edit the icon again, clicking the red color swatch, color will default back to black, save, color is persisted.


emmaburstow:
Lovely stuff, Nathan. This is a really nice improvement for the editor. Code looks fab too. We'll test it all works and let you know if we need anything.

Em

nathanwoulfe:
Always happy to help 👍

PR #7449, opened 2020-01-12
V8: fix route change issue when editing image on new node

### LINKED ISSUE #7377: V8: Failed to retrieve data after editing an uploaded image while creating a new node
I tried to edit an uploaded image while creating a new node and once I finish editing, it fails and I got a loading screen with errors on the console:

> "errorMsg":"Failed to retrieve data for content id 1096","data":"","status":404

and on my screen I got:

> Request error: The URL returned a 404 (not found):
/umbraco/backoffice/UmbracoApi/Content/GetById

## Umbraco version

I am seeing this issue on Umbraco version: 8.4

Reproduction
------------

### Steps to reproduce

1- Create a new node
2- Upload an image
3- Edit the uploaded image ( fill any text in a new property ) and click on "Save and Close"

(GIF)

Note that this issue doesn't occur when you edit an image from the media picker dialog

image

And also it doesn't occur when you try to edit an image of existing nodes (not new).
### Expected result
Should be able to continue on node creation process 

### Actual result
Got a blank screen

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #7377 

### Description
Per the issue description, this PR fixes the bug where (infinite) editing an image from a new (unsaved) content item causes a broken route change, causing a 404, causing a UI error message to display.

The bug was caused by calls to a few methods not required when editing in infinite mode (eg syncing the tree, updating editorState), which then caused unexpected behaviour by updating the ID in the route to that of the edited media item rather than the current content item, which threw the 404. I've fixed this by shifting the non-infinite bits behind the existing infiniteMode check. All seems to be working fine, both editing media in infinite and inline mode, and from new or existing content items.

Other changes in the PR remove Lodash functions replacing them with native JS, renaming some variables (from `image` to `media` as the variable is not always an image) and adding type attribute to buttons to prevent unwanted form submits.

There was also some unreachable code in the mediapicker controller, where a couple of loops referenced $scope.images, which didn't exist. I'm assuming that should be $scope.mediaItems, as that's the collection of media items assigned in the picker...

To test:
- create a new content node, DO NOT SAVE
- add a media item to the node
- edit the media item
- save and close the infinite editor
- get a happy green message
- edit media item again, to confirm changes saved

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/media/media.edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/media/media.edit.controller.js
@@ -73,14 +73,14 @@ function mediaEditController($scope, $routeParams, $q, appState, mediaResource,
         
         var content = $scope.content;
         
-        // we need to check wether an app is present in the current data, if not we will present the default app.
+        // we need to check whether an app is present in the current data, if not we will present the default app.
         var isAppPresent = false;
         
         // on first init, we dont have any apps. but if we are re-initializing, we do, but ...
         if ($scope.app) {
             
             // lets check if it still exists as part of our apps array. (if not we have made a change to our docType, even just a re-save of the docType it will turn into new Apps.)
-            _.forEach(content.apps, function(app) {
+            content.apps.forEach(app => {
                 if (app === $scope.app) {
                     isAppPresent = true;
                 }
@@ -88,7 +88,7 @@ function mediaEditController($scope, $routeParams, $q, appState, mediaResource,
             
             // if we did reload our DocType, but still have the same app we will try to find it by the alias.
             if (isAppPresent === false) {
-                _.forEach(content.apps, function(app) {
+                content.apps.forEach(app => {
                     if (app.alias === $scope.app.alias) {
                         isAppPresent = true;
                         app.active = true;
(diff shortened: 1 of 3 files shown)


emmaburstow:
Hullo Nathan,

Thanks for the work. We shall take a look and do a thing with it. As ever. 

Em 👍

nathanwoulfe:
Cheers @emmaburstow. As ever, shout if I need to do more things at my end.

N

nul800sebastiaan:
Oh no.. another deleted repo @nathanwoulfe? 🙈 

I'm testing this now, will merge when ready but I don't know if that will automatically close this PR .. let's see.

nathanwoulfe:
Yeah, I got a bit carried away and blasted my fork last week. I think this should merge fine, changed files and commits are still visible...

Would be nice if GitHub warned when deleting a repo with pending PRs...

nul800sebastiaan:
Guess it couldn't resolve the reference any more, but it's definitely part of this really awkward merge I did https://github.com/umbraco/Umbraco-CMS/commit/67168068f63429cedb07251b2f3a6d4938d1d23b 

So.. ehm, yeah I did it wrong but it ended up right in the end anyway, so all good, I'll close this one! Thanks Nathan, great work! ⭐

PR #7450, opened 2020-01-13
add dialog for change password, shift form from user details view int…

### LINKED ISSUE #7013: Change password on users should open a dialog
The current user experience when changing the password on a User in the User Section is not optimal.
My 50 cents on what's wrong with the current solution:
— The inputs appear in the existing context leaving a lot of noise around the simple task of changing the password.
— We only show a cancel button, leaving the user with doubt if they want to save the change.

Here a GIF of the current experience:
BwMcW4Tx7p

Instead, I propose that the form for changing the password appears in a dialog/overlay, looking something like this:
image

The confirm password, should save the password on the user, so your not relying on pressing Save for the user.

nathanwoulfe:
…o dialog

### Prerequisites

If there's an existing issue for this PR then this fixes #7013 

### Description
PR does what @nielslyngsoe demoed in the gif attached to the issue - change password opens in a dialog rather than inline in the sidebar.

To test, edit a user, click change password, update details in the dialog (check validation all works), click confirm, user is saved with new password.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/resources/users.resource.js
+++ b/src/Umbraco.Web.UI.Client/src/common/resources/users.resource.js
@@ -405,6 +405,43 @@
                     formattedSaveData),
                 "Failed to save user");
         }
+        
+        /**
+          * @ngdoc method
+          * @name umbraco.resources.usersResource#changePassword
+          * @methodOf umbraco.resources.usersResource
+          *
+          * @description
+          * Changes a user's password
+          *
+          * ##usage
+          * <pre>
+          * usersResource.changePassword(changePasswordModel)
+          *    .then(function() {
+          *        // password changed
+          *    });
+          * </pre>
+          * 
+          * @param {Object} model object to save
+          * @returns {Promise} resourcePromise object containing the updated user.
+          *
+          */        
+        function changePassword(changePasswordModel) {
+            if (!changePasswordModel) {
+                throw "password model not specified";
+            }
+
+            //need to convert the password data into the correctly formatted save data - it is *not* the same and we don't want to over-post
+            var formattedPasswordData = umbDataFormatter.formatChangePasswordModel(changePasswordModel);
+
+            return umbRequestHelper.resourcePromise(
+                $http.post(
+                    umbRequestHelper.getApiUrl(
+                        "userApiBaseUrl",
+                        "PostChangePassword"),
+                    formattedPasswordData),
+                "Failed to save user");
+        }        
 
         /**
           * @ngdoc method
@@ -447,6 +484,7 @@
             createUser: createUser,
             inviteUser: inviteUser,
             saveUser: saveUser,
+            changePassword: changePassword,
             deleteNonLoggedInUser: deleteNonLoggedInUser,
             clearAvatar: clearAvatar
         };
--- a/src/Umbraco.Web.UI.Client/src/common/services/umbdataformatter.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/umbdataformatter.service.js
@@ -152,8 +152,7 @@
             formatUserPostData: function (displayModel) {
 
                 //create the save model from the display model
-                var saveModel = _.pick(displayModel, 'id', 'parentId', 'name', 'username', 'culture', 'email', 'startContentIds', 'startMediaIds', 'userGroups', 'message', 'changePassword');
-                saveModel.changePassword = this.formatChangePasswordModel(saveModel.changePassword);
+                var saveModel = _.pick(displayModel, 'id', 'parentId', 'name', 'username', 'culture', 'email', 'startContentIds', 'startMediaIds', 'userGroups', 'message');
 
                 //make sure the userGroups are just a string array
                 var currGroups = saveModel.userGroups;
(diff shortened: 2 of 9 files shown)


nathanwoulfe:
That is not ideal. I'll review...

nathanwoulfe:
Working now - I hadn't mapped the dialog response back onto the user model in the controller. Have verified by changing password and logging out of the backoffice. Old password no longer works, new password works.

kjac:
@nathanwoulfe thank you for that 👍 it does indeed save now. But it saves the entire user and not just the user password... I can see both pros and cons for that solution, but it might be unexpected to the admins.

Ref the original issue:

> The confirm password, should save the password on the user, so your not relying on pressing Save for the user.

@nielslyngsoe could you please pitch in here? Should the popup save all changes made to the user or _only_ the changed password?

nathanwoulfe:
@kjac that makes sense. Can add a new endpoint for this purpose - there's no current endpoint for an admin to change a user password (endpoint exists for user to change their own).

kjac:
@nathanwoulfe I was thinking along the same lines. But let's get @nielslyngsoe in on this before we do anything - it would be silly to build a new endpoint, only to realize that the admins would be better off with the current solution.

nathanwoulfe:
@kjac sounds good to me. Will hold fire.

nielslyngsoe:
Hi guys, thanks for your concerns. I will just investigate, will be back.

nielslyngsoe:
Do you see any potential issues or do you have any specific concerns regarding saving the hole user? It was what happened before, correct? :-)

kjac:
@nielslyngsoe it's a good question... I can see both pros and cons.

When you're changing the password in a popup dialog, you might not expect the password change to save any changes made to the user details "below" the popup.

On the other hand, if you make changes to the user details, then go change the password and lastly go back to the list of users without saving the user details, you'll be prompted to save changes if the password change didn't save them for you. That might also be a bit counter intuitive.

My take on this is that we should just save the user details along with the password change (as this PR does it) and worry about separating the two later on if it causes any grief. My only reason for bringing all this to your attention was your wording in the original issue: _The confirm password should save the password on the user_ - which sounded like you explicitly wanted to separate the two.

nielslyngsoe:
@kjac I see. Thanks for the insight. 😀

I would say its a bit dangerous, cause the user won't expect anything else than the password to be changed.

Imagine you considered changing the e-mail, but realized that you just needed to change the password. So you go change the password and closes your browser right after. So you would think the e-mail wasn't changed.

So I would say we need to ensure only the password is updated. And I think the endpoint that Nathan proposes is the right way to go about it.

Thanks

nathanwoulfe:
Sounds sensible to me. Changing password is an action that makes sense to be isolated too, it's likely to happen more often than an email change, for example. Given that, it follows that saving the user will not save the password, since it must be saved when the modal submits, so any logic related to password management in both front and back end can be moved too (eg the save method in the JS controller checks some password-related stuff, that can move into the dialog submit handler).

Other benefit of separating the behaviors is that persisting the password change won't require refreshing the view model or rebinding anything, so there's trivial (almost imperceptible) performance benefit. 😸 

Will get to it in the next day or so.

bjarnef:
@nathanwoulfe did you see my comment here to add this for change password on member as well? https://github.com/umbraco/Umbraco-CMS/issues/7013#issuecomment-573579379
Just to keep in mind if we can re-use this and in another PR have a change password overlay for member 👍

nielslyngsoe:
@nathanwoulfe sounds really good.

@bjarnef I think we should consider this too, And if @nathanwoulfe is up for it that would be cool. But otherwise, let's create an up-for-graphs issue when the groundwork has been done by Nathan. :-)

nathanwoulfe:
@bjarnef that makes good sense - that would be all the password-change places right? member, user and user drawer?

I'll check how each is implemented, would be good to consolidate the methods if possible...

bjarnef:
@nathanwoulfe I think it would be fine to keep the change password in user drawer as it is now, but we can change last later if @nielslyngsoe agree with this?
First it would be great to change it on user and member edit views :)

nathanwoulfe:
@bjarnef I wasn't clear - was meaning the server-side methods, UI can stay as is. There's a change password method in three different controllers, need to check if they can be consolidated...

nathanwoulfe:
Latest commit adds a new endpoint PostChangePassword to UsersController, to manage the password changing. Updates the backoffice to add new resource endpoint and removes password properties from the user model. Have tested, seems to be working.

Members is a slightly different gig, since the change password control is a property editor. Same logic should apply though - submitting the modal performs the post, removing the need to save the member after updating the password.

nathanwoulfe:
Rather than holding this up waiting on member view changes, this should go as a standalone change, members need a bit more time/thought. 

Ping @kjac @nielslyngsoe @bjarnef

nielslyngsoe:
@nathanwoulfe I agree. If you still think we should do something about the Members Password, please create an issue for that specifically.

kjac:
@nathanwoulfe this works brilliantly ⭐️ thanks a lot, mate!

nathanwoulfe:
No worries at all, will try to find some time to look at doing the members view too. 🕖🕗🕕

Issue #7452, opened 2020-01-13
Scheduled Publishing not working with multiple cultures


Serder:
## Umbraco version

I am seeing this issue on Umbraco version: 8.4

Reproduction
------------
### Bug summary
I have an Umbraco site that has two cultures: French and English. I have a ListView of Articles (document type) and I want to use the Scheduled Publishing feature for these articles. However I get the following error when I try to do it.

{Message: "No variants flagged for saving"}
Message: "No variants flagged for saving"
![image](https://user-images.githubusercontent.com/8854467/72290476-536eab00-361b-11ea-9fb2-82de5fc3d4b4.png)

![image](https://user-images.githubusercontent.com/8854467/72290528-6e411f80-361b-11ea-9bb5-85ae100788d7.png)

this happen on this URL on chrome: http://localhost:50896/Umbraco#/content/content/edit/2739?mculture=fr-CA

### Steps to reproduce

1. Have a document type that has culture sharing enabled
![image](https://user-images.githubusercontent.com/8854467/72290578-8dd84800-361b-11ea-8096-b027dca9cca7.png)

2. Create some content with this document type and fill in the content fields
3. Click Scheduled Publishing and try to publish in the future
4. Get the error 

I discovered this last week and investigated. I found no reference to that bug in the 8.4 version. I tested with a document type in which the "Allow varying by culture" is not checked and the scheduled publishing worked. However, for our project, we need that Varied cultures to be enabled.

We have another project using Umbraco 8.2 in which the scheduled publishing works with the varied cultures enabled.

### Expected result
The element is set to be published in the future date
### Actual result
Error as shown before.

stevemegson:
`ContentSaveValidationAttribute` is checking that at least one variant has the `save` property set, which isn't true when scheduling content which varies by culture.

We have this code to deal with invariant content before the schedule dialog is opened:

```javascript
if (!isContentCultureVariant()) {
    //ensure the flags are set
    $scope.content.variants[0].save = true;
}
```

When the dialog is submitted we copy the dates over from `model` to `$scope.content`. I guess we need to check here which variants have changed and set the `save` property accordingly. Or would it be safe to just set `save` on all variants even if we're just re-saving the existing dates?

dawoe:
I stumbled on this my self today in 8.4. It is definitly a bug. The website I saw this is on is also a multilingual website

Dave

dawoe:
[...]

PR #7456, opened 2020-01-14
#7452 Fixes schedule publish dialog for multilanguage sites

dawoe:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7452

### Description
This fixes the error when you try to schedule content in a multilanguage site.

In the scheduled publishing dialog the save flag was set correctly on the variant. But when the schedule button is clicked the new value for this was not stored in the orignal variant. This PR fixes that.


nul800sebastiaan:
👍

Issue #7457, opened 2020-01-14
Can't create or save document types in 8.5 with full ModelsBuilder installed


sniffdk:
Just upgraded from 8.3 to 8.5 with the full ModelsBuilder installed. I now can't create or update document types.
I get the following error:
![image](https://user-images.githubusercontent.com/1074004/72351081-f688f880-36df-11ea-88fa-7a021dac0588.png)

I also tried to install an empty 8.5 from nuget. When full ModelsBuilder is installed, the same error is shown. 
If I uninstall ModelsBuilder creating and saving document types works again.

nul800sebastiaan:
😱
We're working on it!

nul800sebastiaan:
Thanks for the report, can reproduce unfortunately, working on a PR now. Sorry about this, we did test this scenario of course, I'm not sure what went wrong during testing! 😰

sniffdk:
No worries, just glad you are working on a quick fix 😄

PR #7458, opened 2020-01-14
Hotfix - Missing registration in IOC when running external MB

bergmania:
Fixes: https://github.com/umbraco/Umbraco-CMS/issues/7457

We need to register the config no matter what which MB is used. The config is used to avoid execution of methods in the internal MB


Shazwazza:
Merged and tested:

using:

```xml
<add key="Umbraco.ModelsBuilder.Enable" value="true"/>
<add key="Umbraco.ModelsBuilder.ModelsMode" value="PureLive"/>
```

* Creating/Updating/Deleting document type and Rendering Starter Kit (since it uses MB models) with embedded MB (without the Umbraco.ModelsBuilder.dl) in the bin)
* Creating/Updating/Deleting document type and Rendering Starter Kit (since it uses MB models) with Umbraco.ModelsBuilder 8.1.0 (https://www.nuget.org/packages/Umbraco.ModelsBuilder/8.1.0, with the Umbraco.ModelsBuilder.dl) in the bin)

PR #7459, opened 2020-01-14
Netcore: Move property editors to infrastructure (1st part)

elit0451:
Moved half of the Property Editors from **Umbraco.Web** into **Umbraco.Infrastructure**.
Removed the usages of `Current.*` and instead injected their dependencies in the constructor.
Fixed tests to work with the changes.


bergmania:
Looking super good.. No comments at all :)

Issue #7469, opened 2020-01-17
V8.5.1: Ambiguous extension method signature between embedded and external ModelsBuilder


sniffdk:
The IPublishedContent .Value extension method that takes an Expression<Func<TModel, TValue>> parameter is duplicated in the new embedded ModelsBuilder.
Due to them being in the same namespace, there is an ambiguous call between the two, as the method signatures are identical.
Screenshots below:
![image](https://user-images.githubusercontent.com/1074004/72601047-f6257300-3914-11ea-9f52-8d3fe89d1621.png)

![image](https://user-images.githubusercontent.com/1074004/72601066-fd4c8100-3914-11ea-9029-666eaa6fb0d9.png)

nul800sebastiaan:
Thanks again Mads, we're looking into it!

ronaldbarendse:
This issue is now fixed in PR #7470 by renaming the Umbraco.ModelsBuilder.Embedded version of `Value(...)` to `ValueFor(...)`:

https://github.com/umbraco/Umbraco-CMS/blob/41dc8d915bdcadb628e6dde9422586310ac66a1b/src/Umbraco.ModelsBuilder.Embedded/PublishedElementExtensions.cs#L20

clausjensen:
Closing this one - fixed (final fix) in 8.5.3 😄

bielu:
@nul800sebastiaan This one should be flagged as braking change name of Extension method was changed, which means we break code which was using that.

nul800sebastiaan:
Okay, done.

PR #7470, opened 2020-01-17
https://github.com/umbraco/Umbraco-CMS/issues/7469 - Fixed  Ambiguous…

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/7469

- Renames our extension method from `Value` to `ValueByExpression`
  - There really is no good solution to this - An alternative is to remove the extension method from the embedded modelsbuilder

### DIFF
--- a/src/Umbraco.ModelsBuilder.Embedded/PublishedElementExtensions.cs
+++ b/src/Umbraco.ModelsBuilder.Embedded/PublishedElementExtensions.cs
@@ -17,14 +17,13 @@ namespace Umbraco.Web
         /// <summary>
         /// Gets the value of a property.
         /// </summary>
-        public static TValue Value<TModel, TValue>(this TModel model, Expression<Func<TModel, TValue>> property, string culture = null, string segment = null, Fallback fallback = default, TValue defaultValue = default)
+        public static TValue GetValue<TModel, TValue>(this TModel model, Expression<Func<TModel, TValue>> property, string culture = null, string segment = null, Fallback fallback = default, TValue defaultValue = default)
             where TModel : IPublishedElement
         {
             var alias = GetAlias(model, property);
             return model.Value<TValue>(alias, culture, segment, fallback, defaultValue);
         }
 
-        // fixme that one should be public so ppl can use it
         private static string GetAlias<TModel, TValue>(TModel model, Expression<Func<TModel, TValue>> property)
         {
             if (property.NodeType != ExpressionType.Lambda)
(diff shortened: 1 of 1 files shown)


sniffdk:
Perhaps it could be named .Value and .ValueFor
This would put the naming in line with how mvc already does it eg. Html.Label and Html.LabelFor

Shazwazza:
I'm pretty sure we can solve this with AppDomain events, also considering that `Umbraco.ModelsBuilder.Embedded` is a separate assembly may make it easier. I'll test some things out.

ronaldbarendse:
Having the same namespace (and methods) in different assemblies will inevitably cause these kind of problems, but instead of changing the method, you should use `extern alias` (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/extern-alias) or better yet: rename the namespace 😉

@Shazwazza Isn't this a compiler problem, so how would AppDomain events help prevent this?

Shazwazza:
@ronaldbarendse yes you are right but the immediate issue is that of dynamic compilation in razor and app_code. I pushed a temp fix here for razor compilation work arounds https://github.com/umbraco/Umbraco-CMS/pull/7475 but there isn't a possible work around for app_code. These were just interim hotfix's, we will be pushing a change to the `Umbraco.ModelsBuilder` package in the next release that won't have the overlap/duplication of the code in the CMS which will solve the problem. Namespace rename isn't a solution either because these APIs are used in razor views mostly and to have a seamless upgrade experience both namespaces would need to be imported and we'd have the same problem.

ronaldbarendse:
Hmm, seems releasing new versions of both Umbraco and ModelsBuilder is still the only real solution to this problem (as you mentioned in the linked PR)... That will break the supported versions of ModelsBuilder though (@zpqrtbnk be sure to make this at least a new minor release, so older Umbraco versions can still get patch updates).

Why wasn't the full ModelsBuilder package transferred to Umbraco? It was already renamed from Zbu.ModelsBuilder, included as core package and it's even MIT licensed... API support and the VS extension could then still be supported by Stephan (possibly as paid extensions if that's required to keep it updated).

Shazwazza:
If people are using this ext method in compiled code and we release a new MB version, yes they will need to re-compile but they will need to do that now anyways with this conundrum. The new `Umbraco.ModelsBuilder` package will target umbraco 8.5.3. There's a blog post about what parts of MB were transfered here https://umbraco.com/blog/umbraco-85-release/

This next `Umbraco.ModelsBuilder` release to solve this single namespace /ext issue will most likely be the last MB release under this package name, but I guess we'll see what transpires with the community package.

Issue #7368, opened 2019-12-19
.NET Core transition: Clean up obsoletes


bergmania:
**This task is up for grabs, for contributing to the .NET Core transition project**

On the development branch of the .net core transition project [`netcore/dev`](https://github.com/umbraco/Umbraco-CMS/tree/netcore/dev), we have a lot of classes, methods and properties that are marked with the `ObsoleteAttribute`

Some of these are very easy to remove because there are no internal usages. Others are still in use internal, and will need some more refactor, to be able to remove.

If any questions, feel free to ask :)

Feel free to resolve this task in multiple smaller PRs, Due to the fact that long running feature branch most likely have end up having merge conflict with [`netcore/dev`](https://github.com/umbraco/Umbraco-CMS/tree/netcore/dev)

hishamco:
If not one will pick this one, I will add it to my list :)

benjaminc:
I'll start working on this.

benjaminc:
I was able to get most of the obsolete items removed cleanly, however there was one I had questions about.  The Umbraco.Core.Constants.ObjectTypes.Strings.Stylesheet constant in Umbraco.Abstractions is marked as obsolete with the note "This no longer exists in the database".  It is used only one place, in the UmbracoObjectTypeAttribute on the Umbraco.Core.Models.UmbracoObjectTypes.Stylesheet enum value.  That enum value is used for UDI type lookups, audit entries, and relation object types.  Can all of these be deleted, or should I take a different approach here?

bergmania:
@Shazwazza, do you know the history of this `Umbraco.Core.Constants.ObjectTypes.Strings.Stylesheet`

Shazwazza:
These can be deleted, Stylesheet data (believe it or not) used to be stored in the database 😱 many many moons ago, i'm amazed that we still have references to that, please remove :)

benjaminc:
I've submitted pull request #7491 to remove them.

PR #7472, opened 2020-01-18
Netcore/feature/7368 cleanup obsoletes

benjaminc:
### Prerequisites
This is one of multiple pull requests for #7368

### Description
This removes unused, obsolete methods and classes in the Umbraco.Web project.  None of the code changed was referenced anywhere in the solution.  It also updates the SourceLink package references to the latest beta version, because it wouldn't build on my machine without doing that, and so I imagine others will have a similar problem.  Given that the Umbraco.Infrastructure project was already using the more recent version, this shouldn't cause any problems.


bergmania:
Looks very good.. 
There was a single view using `Html.ReplaceLineBreaksForHtml`, that I pushed an update for.

I will spend some time testing

PR #7475, opened 2020-01-20
Fixing Ambiguous extension method signature between embedded and external ModelsBuilder

Shazwazza:
fixes: https://github.com/umbraco/Umbraco-CMS/issues/7469

This is a work around which works for razor views. If the legacy MB is detected this will remove the embedded MB from the referenced DLLs for razor compilation.

This works fine in my tests, however this issue would still be present if there was code in App_Code since that is dynamically compiled and unfortunately there's no way to override the behavior of how that compilation occurs... I've gone very deep to try and although we can get close unfortunatley MS has hard coded type checks for `SourceFileBuildProvider` which is what they use to compile App_Code so it cannot be overridden. 

For compiled code, if this error occurs then the full extension method class name can be used to call the method instead. This PR changes the class name so that they can be differentiated.

In the near future, i would propose that we ship a new version of the legacy `Umbraco.ModelsBuilder` package that doesn't ship with this extension. We should actually ship both an Umbraco version and a Umbraco.ModelsBuilder version at the same time (lets say it's 8.5.3), then this Umbraco.ModelsBuilder version can have a min requirement of 8.5.3 and 8.5.3 can remove this code check. Make sense?

### DIFF
--- a/src/Umbraco.ModelsBuilder.Embedded/Compose/ModelsBuilderComposer.cs
+++ b/src/Umbraco.ModelsBuilder.Embedded/Compose/ModelsBuilderComposer.cs
@@ -1,5 +1,8 @@
 ﻿using System;
+using System.CodeDom;
+using System.CodeDom.Compiler;
 using System.Collections;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Web;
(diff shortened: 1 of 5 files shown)


ronaldbarendse:
As PR #7470 is merged, these changes would not be required, right?

Shazwazza:
@ronaldbarendse correct, the name change was `ValueFor` i'll close this PR ... was an interesting challenge :)

PR #7479, opened 2020-01-20
Fixing ContentFinderByUrlAlias for multilingual site with a domain

### LINKED ISSUE #7383: UmbracoUrlAlias is not fully working on multi language website
Hello, I've just identified what seems to big a bug with the UmbracoUrlAlias on mutlilanguage website.

I have tried to reproduce this issue on a clean installation of Umbraco 8.4.0

## Umbraco version

I am seeing this issue on Umbraco version: 8.3.0 and 8.4.0

Reproduction
------------

umbracoAlias

As you can see the URL generated in the backoffice is wrong, it shouldn't include the domain part.

In this example, the cultures and domain configuration is like this: 
image

Then when you want to add an umbraco URL alias to a page in another language (in this example FR), it does not work at all:

umbracoAliasFR

### Bug summary

Umbraco Alias URL is not working as expected with multi language websites.

filefontaine:
### Prerequisites

If there's an existing issue for this PR then this fixes #7383 

### Description

Need some help for this don't know if we need to go more in the refactoring of FindContentByAlias(...)

### DIFF
--- a/src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
+++ b/src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
@@ -77,7 +77,7 @@ namespace Umbraco.Web.Routing
             var testList = new List<string>
             {
                 trimAlias.TrimStart('/'), // is "alias"
-                $"{trimAlias}", // is "/alias"
+                trimAlias, // is "/alias"
                 $"{trimAlias.TrimStart('/')}/", // is "alias/"
                 $"{trimAlias}/" // is "/alias/"
             };
--- a/src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
+++ b/src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Text;
 using System.Linq;
 using Umbraco.Core.Logging;
@@ -67,11 +68,17 @@ namespace Umbraco.Web.Routing
 
             var hasDomain = !domaineName.IsNullOrWhiteSpace();
 
-            var test1 = $"{alias.TrimStart('/')}";
-            var test2 = $"/{test1}"; // test2 is "/alias"
-            test1 = $"{test1}"; // test1 is "alias"
+            // Create the list of potential alias that can be found on umbracoAliasUrl
+            var trimAlias = alias.TrimStart('/');
+            var testList = new List<string>
+            {
+                trimAlias, // is "alias"
+                $"/{trimAlias}", // is "/alias"
+                $"{trimAlias}/", // is "alias/"
+                $"/{trimAlias}/" // is "/alias/"
+            };
 
-            bool IsMatch(IPublishedContent c, string a1, string a2)
+            bool IsMatch(IPublishedContent c)
             {
                 // this basically implements the original XPath query ;-(
                 //
@@ -96,8 +103,10 @@ namespace Umbraco.Web.Routing
                 if (string.IsNullOrWhiteSpace(v)) return false;
                 v = v.Replace(" ", "");
                 // Split UrlAlias to a list and add the domain name if hasDomain
-                var vList = v.Split(',').Select(x => hasDomain ? $"{domaineName.TrimStart('/')}/{x}" : x).ToList();
-                return vList.InvariantContains(a1) || vList.InvariantContains(a2);
+                var t = v.Split(',');
+                var vList = t.Select(x => hasDomain ? $"{domaineName.TrimStart('/').TrimEnd('/')}/{x}" : x).ToList();
+                vList.AddRange(t);
+                return vList.ContainsAny(testList);
             }
 
             // TODO: even with Linq, what happens below has to be horribly slow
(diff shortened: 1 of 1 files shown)


filefontaine:
I find why some check fail, i forgot that domain can be domain1.com/ or domain1.com/en
try to fix this

src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
```diff
@@ -64,11 +67,22 @@ private static IPublishedContent FindContentByAlias(IPublishedContentCache cache
...
+            var match = Regex.Match(domaineName, @"(?<!http:\/|http:|https:\/|https:)\/.+");
+
+            // Create the list of potential alias that can be found on umbracoAliasUrl
+            var trimAlias = hasDomain ? alias.TrimStart(match.Value) : alias;
+            var testList = new List<string>
+            {
+                trimAlias.TrimStart('/'), // is "alias"
+                $"{trimAlias}", // is "/alias"
+                $"{trimAlias.TrimStart('/')}/", // is "alias/"
+                $"{trimAlias}/" // is "/alias/"
```
ronaldbarendse: Instead of using string interpolation, there's also an Umbraco extension method to conditionally add characters to the end of a string: `EnsureEndsWith()`.

src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
```diff
@@ -64,11 +67,22 @@ private static IPublishedContent FindContentByAlias(IPublishedContentCache cache
...
-            bool IsMatch(IPublishedContent c, string a1, string a2)
+            // Try to find any first relative path
+            var match = Regex.Match(domaineName, @"(?<!http:\/|http:|https:\/|https:)\/.+");
+
+            // Create the list of potential alias that can be found on umbracoAliasUrl
+            var trimAlias = hasDomain ? alias.TrimStart(match.Value) : alias;
+            var testList = new List<string>
+            {
+                trimAlias.TrimStart('/'), // is "alias"
+                $"{trimAlias}", // is "/alias"
```
ronaldbarendse:
```suggestion
                trimAlias, // is "/alias"
```

src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
```diff
@@ -64,11 +67,22 @@ private static IPublishedContent FindContentByAlias(IPublishedContentCache cache
...
+                $"{trimAlias.TrimStart('/')}/", // is "alias/"
```
ronaldbarendse:
```suggestion
                trimAlias.EnsureEndsWith('/'), // is "alias/"
```

src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
ronaldbarendse:
```suggestion
                trimAlias.EnsureEndsWith('/') // is "/alias/"
```

src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
```diff
@@ -64,11 +67,22 @@ private static IPublishedContent FindContentByAlias(IPublishedContentCache cache
...
             const string propertyAlias = Constants.Conventions.Content.UrlAlias;
 
-            var test1 = alias.TrimStart('/') + ",";
-            var test2 = ",/" + test1; // test2 is ",/alias,"
-            test1 = "," + test1; // test1 is ",alias,"
+            var hasDomain = !domaineName.IsNullOrWhiteSpace();
 
-            bool IsMatch(IPublishedContent c, string a1, string a2)
+            // Try to find any first relative path
+            var match = Regex.Match(domaineName, @"(?<!http:\/|http:|https:\/|https:)\/.+");
```
ronaldbarendse: There's a DomainAndUri class in Umbraco: it's probably easier to use that instead of parsing the domain manually here.

src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
```diff
@@ -64,11 +67,22 @@ private static IPublishedContent FindContentByAlias(IPublishedContentCache cache
...
+
+            // Create the list of potential alias that can be found on umbracoAliasUrl
+            var trimAlias = hasDomain ? alias.TrimStart(match.Value) : alias;
```
ronaldbarendse: `TrimStart()` might remove the value multiple times, so this should be replaced with a check/substring instead.

src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
ronaldbarendse: Ah, looks like my suggestion removed the `TrinStart()` 🙄 This should be completely rewritten either way, as this kind of matching seems unneeded and obsolete if this content finder correctly uses the current `DomainAndUri`...

filefontaine: Yes i have see that, i will try to add the DomainAndUri object

Issue #6421, opened 2019-09-25
Allow adding custom install/upgrade steps and fix ordering


ronaldbarendse:
It's currently not possible to add custom install/upgrade steps to the setup/upgrade screen, as the ordering of the steps filter out any discovered/registered steps:

https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web/Install/InstallStepCollection.cs#L13-L35

It looks like it should just order the steps based on the `ServerOrder`:
https://github.com/umbraco/Umbraco-CMS/blob/853087a75044b814df458457dc9a1f778cc89749/src/Umbraco.Web/Install/Models/InstallSetupStepAttribute.cs#L35

But maybe a better solution would be to change this to an `OrderedCollectionBuilderBase` or `WeightedCollectionBuilder`?

One use-case for this might be to ask for the Models Builder mode (and other settings) on install, as that's currently an external package that's installed by default.

sitereactor:
Would love to see this as well.
Would also give more flexibility in terms of customizing an install up-front or even unattended installs of both umbraco and starterkits if its taken a bit further.

nul800sebastiaan:
We'll have a chat at HQ soon about this. Obviously the installer was never intended to be extensible and I am not sure we would want to support that scenario since we've always wanted the installer to ask as few questions as possible. 

I can see the point though on it being something you would add for your internal projects before running the installer.

umbrabot:
Hiya @ronaldbarendse,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #7480, opened 2020-01-20
Allow for add additional steps/remove steps on installation

### LINKED ISSUE #7464: Umbraco exposes attribute for creating an installation step, but doesn't allow for customization of installation
Hi all,

I was looking into a way to unregister default installation starter kit with v8 installation, but actually I noticed a few things:
`InstallSetupStep<T>` is public
`InstallSetupStepAttribute` is public and sealed
Which is fine, but here we could notice, something that in my opinion is weird:
`InstallStepCollection` is public and sealed
`public static class Installer` is a static helper class

And from my perspective, I don't really understand why:

1. InstallStepCollection is sealed and not written in a way where we could add/remove steps, which are safe to remove, as they are not required for successful installation (I.e. Starter kit installation)?
2. Installer class: why it was done as a static helper, instead of smaller Components/ Composer to allow to remove not needed parts?

Thanks,
Arkadiusz

bielu:
### Prerequisites

If there's an existing issue for this PR then this fixes #7464 & #6421

As we discussed in that issues it is really helpful to give the option to customize the installation process of Umbraco.
My PR is adding an event and 2 methods to allow for that.
Example code for the remove starter kit installation
```
using Umbraco.Core.Composing;
using Umbraco.Web.Install;
using Umbraco.Web.Install.InstallSteps;

namespace Umbraco.Web.UI
{
    public class RemoveStarterKitComponent : IComponent
    {
        private InstallStepCollection _installStepCollection;

        public RemoveStarterKitComponent(InstallStepCollection installStepCollection)
        {
            _installStepCollection = installStepCollection;
        }

        public void Initialize()
        {
            _installStepCollection.PrepareListEvent += InstallStepCollectionOnPrepareListEvent;
        }

        private void InstallStepCollectionOnPrepareListEvent(object sender, InstallStepCollection e)
        {
          e.RemoveStepByType(typeof(StarterKitDownloadStep));
          e.RemoveStepByType(typeof(StarterKitCleanupStep));
          e.RemoveStepByType(typeof(StarterKitInstallStep));

        }

        public void Terminate()
        {
        }
    }
}
```
```
using Umbraco.Core.Composing;

namespace Umbraco.Web.UI
{
    public class RemoveStarterKitFromInstallationComposer : IUserComposer
    {
        {

        }
    }
}

```
One of the things which were mention in one of the issues is ordering, where I decide to use a simpler version. It means I am staying with hardcoded order, and expose the only starter kit to be removed and later all new steps could be registered before the SetUmbracoVersionStep.

I am really happy to make a lot changes to that PR as I am aware not everything here is done in best way, but it was the "easiest" way. Probably proposition of ServerOrder would be better but actually I wasn't sure how it should work.

### DIFF
--- a/src/Umbraco.Web.UI/Umbraco.Web.UI.csproj
+++ b/src/Umbraco.Web.UI/Umbraco.Web.UI.csproj
@@ -135,9 +135,6 @@
     <Compile Include="..\SolutionInfo.cs">
       <Link>Properties\SolutionInfo.cs</Link>
     </Compile>
-    <Compile Include="App_Code\AddCustomFieldWithAnalyzer.cs" />
-    <Compile Include="App_Code\RemoveStarterKitComponent.cs" />
-    <Compile Include="App_Code\RemoveStarterKitFromInstallationComposer.cs" />
     <Compile Include="Config\splashes\NoNodes.aspx.cs">
       <DependentUpon>noNodes.aspx</DependentUpon>
       <SubType>ASPXCodeBehind</SubType>
--- a/src/Umbraco.Web.UI/Umbraco.Web.UI.csproj
+++ b/src/Umbraco.Web.UI/Umbraco.Web.UI.csproj
@@ -135,6 +135,9 @@
     <Compile Include="..\SolutionInfo.cs">
       <Link>Properties\SolutionInfo.cs</Link>
     </Compile>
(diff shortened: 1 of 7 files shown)


emmaburstow:
Hi @bielu 

We'll have a chat about this one and get back to you. Thanks for the work,

Emma

PR #7486, opened 2020-01-21
NetCore:  Moved CacheRefreshers to abstractions

bergmania:
Moved cache refreshers to abstractions...

- Needed to inject IJsonSerializer in a couple of classes to avoid using newtonsoft json directly.
- Introduced interface on `IdkMap` - The interface is located in abstractions, the class still exists in web
- Needed to split `CacheHelperExtensions` such that a single extension method is available in Abstractions

[AB#4635](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4635)


src/Umbraco.Web/Cache/DistributedCacheBinder_Handlers.cs
```diff
@@ -390,13 +390,13 @@ private void FileService_SavedTemplate(IFileService sender, SaveEventArgs<ITempl
         private void MacroService_Deleted(IMacroService sender, DeleteEventArgs<IMacro> e)
         {
             foreach (var entity in e.DeletedEntities)
-                _distributedCache.RemoveMacroCache(entity);
+                _distributedCache.RemoveMacroCache(_jsonSerializer, entity);
```
Shazwazza: I don't think we should have to pass in a serializer here, i think the cache refresher should have it's own serializer injected and calling code shouldn't pass one in. This is also so that some serializer requirements aren't broken since the cache refresher should serialize and deserialize always with the exact same one whereas with these APIs that may not happen.

src/Umbraco.Web/Cache/DistributedCacheExtensions.cs
```diff
@@ -189,16 +190,16 @@ public static void RefreshAllPublishedSnapshot(this DistributedCache dc)
 
         #region MacroCache
 
-        public static void RefreshMacroCache(this DistributedCache dc, IMacro macro)
+        public static void RefreshMacroCache(this DistributedCache dc, IJsonSerializer jsonSerializer, IMacro macro)
```
Shazwazza: We shouldn't pass in a serializer here (see prev comments). The `MacroCacheRefresher.Serialize` shouldn't be static, instead we should resolve the cache refresher instance from the collection and call a non-static Serialize method on it. Same with the RemoveMacroCache method. 

Should be the same for all json cache refreshers, we should remove their static methods and require resolving the instance to access it's serialize/deserialize methods and the cache refresher has it's own injected serializer to do the work.

src/Umbraco.Abstractions/Services/IIdkMap.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using Umbraco.Core.Models;
+
+namespace Umbraco.Core.Services
+{
+    public interface IIdkMap
+    {
+        void SetMapper(UmbracoObjectTypes umbracoObjectType, Func<int, Guid> id2key, Func<Guid, int> key2id);
```
Shazwazza: This should not be included on the interface. The usage of this is strange and the way to work around it is to have a non-web based implementation of IIdkMap and a web based version of it. The non-web based version doesn't have to change, however the Web based version would have a custom implementation for `GetIdForKey` and `GetKeyForId` where it checks specifically for `UmbracoObjectTypes.Document` and `UmbracoObjectTypes.Media` and uses a different lookup mechanism for it - which would use the PublishedSnapshotService instead, then we would remove the code in PublishedSnapshotService where it calls SetMapper.

bergmania: I'm not 100% how you see the implementation, can you do this part? - The end result with a slimmer interface sounds good.

bergmania:
I pushed changed requested, except the one with a slimmer `IIdkMap` - Hopefully, you have time to pick this up.

I'm not completely happy with the changes in `DistributedCacheBinder_Handlers.cs` The way I get the `MacroCacheRefresher` can maybe be done in a better way?  https://github.com/umbraco/Umbraco-CMS/blob/47c211907a5668aad1e37211d9a1ea9216f1e70d/src/Umbraco.Web/Cache/DistributedCacheBinder_Handlers.cs#L392-L404

Feel free to fix up

src/Umbraco.Abstractions/Services/IIdkMap.cs
Shazwazza: Just pushed a change for this, will have a look at the `DistributedCacheBinder_Handlers` now

bergmania: Looks good :)

PR #7489, opened 2020-01-21
Remove Image alias from protected/system media types

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6632.

### Description
This undoes a part of PR https://github.com/umbraco/Umbraco-CMS/pull/6943, so only `File` and `Folder` media type aliases are protected, as those are absolutely required for Umbraco (and the media section) to work correctly. The `Image` media type isn't required, as images can also be saved as a `File` media type.

The `MediaController` did an automatic selection between `File` and `Image`, so that now always selects `File` if the `Image` media type is not available. Pasting images into the RTE always saved them as `Image` types, but now falls back to `File`.

The changes can be tested using these steps:
- Remove the `Image` media type (and double-check that you can't remove `Folder` and `File` types)
- Upload an image by dropping it into a folder (from the media section and media picker), no errors should occur and the uploaded image is saved as a `File` media type
- Paste an image into a RTE (with image enabled), no errors should occur and the uploaded image is saved as a `File` media type

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
@@ -25,15 +25,17 @@ namespace Umbraco.Web.PropertyEditors
         Group = Constants.PropertyEditors.Groups.RichContent)]
     public class GridPropertyEditor : DataEditor
     {
-        private IMediaService _mediaService;
-        private IContentTypeBaseServiceProvider _contentTypeBaseServiceProvider;
-        private IUmbracoContextAccessor _umbracoContextAccessor;
-        private ILogger _logger;
+        private readonly IMediaService _mediaService;
+        private readonly IMediaTypeService _mediaTypeService;
+        private readonly IContentTypeBaseServiceProvider _contentTypeBaseServiceProvider;
+        private readonly IUmbracoContextAccessor _umbracoContextAccessor;
+        private readonly ILogger _logger;
 
-        public GridPropertyEditor(ILogger logger, IMediaService mediaService, IContentTypeBaseServiceProvider contentTypeBaseServiceProvider, IUmbracoContextAccessor umbracoContextAccessor)
+        public GridPropertyEditor(ILogger logger, IMediaService mediaService, IMediaTypeService mediaTypeService, IContentTypeBaseServiceProvider contentTypeBaseServiceProvider, IUmbracoContextAccessor umbracoContextAccessor)
             : base(logger)
         {
             _mediaService = mediaService;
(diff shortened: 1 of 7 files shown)


src/Umbraco.Web/PropertyEditors/GridPropertyEditor.cs
```diff
@@ -25,15 +25,17 @@ namespace Umbraco.Web.PropertyEditors
...
-        private IUmbracoContextAccessor _umbracoContextAccessor;
-        private ILogger _logger;
+        private readonly IMediaService _mediaService;
+        private readonly IMediaTypeService _mediaTypeService;
+        private readonly IContentTypeBaseServiceProvider _contentTypeBaseServiceProvider;
+        private readonly IUmbracoContextAccessor _umbracoContextAccessor;
+        private readonly ILogger _logger;
 
-        public GridPropertyEditor(ILogger logger, IMediaService mediaService, IContentTypeBaseServiceProvider contentTypeBaseServiceProvider, IUmbracoContextAccessor umbracoContextAccessor)
+        public GridPropertyEditor(ILogger logger, IMediaService mediaService, IMediaTypeService mediaTypeService, IContentTypeBaseServiceProvider contentTypeBaseServiceProvider, IUmbracoContextAccessor umbracoContextAccessor)
```
ronaldbarendse: The additional parameter on this public constructor is automatically injected, but should we keep the original constructor/overload with an obsolete message?

nul800sebastiaan:
I am probably missing something but numerous packages also rely on the `Image` type to exist, so we're trying to be helpful here by not letting people delete it and thus making fewer things fail.

If you really must mess with this then of course you're free to fire off a call to the `MediaTypeService` to delete this item, which is not recommended.

ronaldbarendse:
You should be able to tailor Umbraco to your own specific needs, right? There's no reason to protect the `Image` media type, as you can just as well use the `File` media type to store images. And like @kjac mentioned in https://github.com/umbraco/Umbraco-CMS/pull/6943#issuecomment-575168946, it would be beneficial to eventually get rid of the hard-coded aliases altogether, so this just leads the way with the `Image` alias.

This should at least be supported by Umbraco, as the alias is only stored as a convention within the `Constants.Conventions.MediaTypes` class. Package authors might indeed depend on specific types to be available, but if that's the case, you either report it as a bug or don't use the package at all...

Maybe the UI should still protect you from deleting the `Image` media type, but either way: Umbraco should make sure it doesn't throw exceptions when it's not available and use the `File` media type instead.

PR #7496, opened 2020-01-23
Replace Current.Services references with DI where possible

benjaminc:
### Prerequisites
Resolves some of #7369 

### Description
Replaces Current.Services references with dependency injection where possible in the Umbraco.Web project.


bergmania:
Looks good

PR #7497, opened 2020-01-23
Netcore: Moves services and dependencies from Web to Abstractions

bergmania:
- Moved services and dependencies
  - `DashboardService`
  - `SectionService`
  - `TreeService`
- Refactored MenuItem to not include add method that support un-localized string, and fixed references
- Needed to make `DashboardService` public
- Removed some unused field (`private bool _defaultToLiveEditing;`)
- Removed unsued class (`MvcVersionCheck`)
- Removed support for vbhtml views


elit0451:
Good changes 🎉 Merging

PR #7499, opened 2020-01-23
V8 Enable segments in Client-code

nielslyngsoe:
Draft PR, still in progress


src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontent.directive.js
```diff
@@ -7,12 +7,11 @@
     var umbVariantContent = {
         templateUrl: 'views/components/content/umb-variant-content.html',
         bindings: {
-            content: "<",
+            content: "=",
             page: "<",
-            editor: "<",
+            editor: "=",
```
Shazwazza: Having 2 way binding goes against how components 'should' work. The standard approach to components is to have one way binding and then callbacks else we fall into the previous angular traps where you don't know what is actually happening. Any reason why we need this? https://docs.angularjs.org/guide/component#component-based-application-architecture

src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontenteditors.directive.js
```diff
@@ -122,157 +118,61 @@
...
+                    segment: variantSegment
                 });
             }
             else {
                 //replace the editor for the same culture
                 vm.editors[index].content = variant;
             }
         }
-
-        function initVariant(variant, editorIndex) {
```
Shazwazza: So where has this code gone? There are vast amounts of changes in this PR so i can't really follow along very well but I fear a lot of stuff that we discovered was necessary is now being undone. There's a ton of test cases that needs to be covered regarding split views, invariant properties, editing values and ensure they are syncing between split views, ensuring that the same model is not being used between split views except for invariant properties, etc...

src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontent.directive.js
nielslyngsoe: Good input, let me have a look to see what we can do about that.

src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontenteditors.directive.js
nielslyngsoe: This method basically added copies of other parts of the data to the variant object, to enable sub-parts of the editor to use that data. But the cause of this hack we had to maintain that data from multiple places. Resulting in an overly complex code, that is hard to read and not very open for change.

That's basically why I have chosen to refactor this. Let's have a talk to go through changes, cause I like to get your view on the whole thing.

src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontent.directive.js
nielslyngsoe: Gone now, I will mark this feedback as resolved.

nielslyngsoe:
fixes https://github.com/umbraco/Umbraco-CMS/issues/7610

nielslyngsoe:
fixes https://github.com/umbraco/Umbraco-CMS/issues/7611

nielslyngsoe:
fixes https://github.com/umbraco/Umbraco-CMS/issues/7900

nielslyngsoe:
Merged into this PR https://github.com/umbraco/Umbraco-CMS/pull/7973

Issue #7504, opened 2020-01-24
PartialViewMacroModel should implement IContentModel


tompipe:
## Umbraco version

I am seeing this issue on Umbraco version: 8.4.0

Reproduction
------------

### Bug summary

Within the UmbracoViewPage, when first requesting the current UmbracoHelper, there's some logic to create one from the ViewData.Model.

If the Model implements IContentModel, it sets up the umbraco helper with the AssignedContentItem from the Model.

https://github.com/umbraco/Umbraco-CMS/blob/0b6f1f6e117d1cb475230073699c725ea7b2dd4a/src/Umbraco.Web/Mvc/UmbracoViewPageOfTModel.cs#L82-L100

However, PartialViewMacroModel doesn't implement IContentModel, so the UmbracoHelper in a PartialViewMacroPage doesn't have an AssignedContentItem set correctly, resulting in the following error:

![image](https://user-images.githubusercontent.com/629719/73086881-60b45100-3ec9-11ea-876f-c38b45a006b5.png)

### Steps to reproduce

1. Create a new partial view macro (with macro)
2. Enable use in RTE & grid
3. Enable render in RTE & grid
4. Add `@Umbraco.AssignedContentItem.Name` to the macro partial
5. Configure a grid layout on a page, and allow macro editor.
6. Add the macro to the grid cell
7. Observe the error

### Expected result

The UmbracoHelper correctly initialised with the current content item.

### Actual result

The UmbracoHelper is not correctly configured, resulting in an error

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/7505

PR #7505, opened 2020-01-24
Ensure macromodel implements IContentModel, fixing #7504

tompipe:
Implement IContentModel on PartialViewMacroModel - fixing #7504

### DIFF
--- a/src/Umbraco.Web/Models/PartialViewMacroModel.cs
+++ b/src/Umbraco.Web/Models/PartialViewMacroModel.cs
@@ -6,7 +6,7 @@ namespace Umbraco.Web.Models
     /// <summary>
     /// The model used when rendering Partial View Macros
     /// </summary>
-    public class PartialViewMacroModel
+    public class PartialViewMacroModel : IContentModel
     {
 
         public PartialViewMacroModel(IPublishedContent page,


emmaburstow:
Brill! Lovely spot and thanks for making it a PR. 

Speak soon,

Em

tompipe:
What's the reason you consider it a breaking change?

kjac:
I may be wrong (and by all means please prove me wrong here), but far as I know.. adding an interface to a type is a binary breaking change to that type..?

tompipe:
That would usually be the case, but perhaps it's not breaking as such. It doesn't add/remove any  properties, and it's public 'surface' is effectively unchanged, it already contained the Content property, so any existing code should remain unaffected (in theory 😂)

Checking guidance for .net core's definitions of acceptable/breaking changes, I guess it falls under 'REQUIRES JUDGMENT' 🤷‍♂️

https://docs.microsoft.com/en-us/dotnet/core/compatibility/#types

kjac:
@tompipe it's not breaking on code level as such, since you can just recompile and be on your way. However if you upgrade without recompiling, I do believe things will break "just" because the interface was added. At least that's how I conceive it.

Perhaps @Shazwazza has a perspective on this?

Shazwazza:
Additive changes are fine and will not require you to recompile.

Shazwazza:
see https://our.umbraco.com/Documentation/Development-Guidelines/breaking-changes

nul800sebastiaan:
Thanks @tompipe! 👍 

And thanks Shan and Kenn for the discussion, good refresher on breaking changes!

PR #7508, opened 2020-01-25
Use strongtyped Models as grid and RTE config

bielu:
### Prerequisites

### Description
When I was working for one of the internal projects in my Company I had a need to use strong type models in a grid configuration and I notice in Umbraco source code comment which was saying:
`// TODO: Make these strongly typed, for now this works though`
And I decide to make PRs with my models, which I am using to set up grid configs.

### DIFF
--- a/src/Umbraco.Web/Models/PropertyEditorsConfigs/GridArea.cs
+++ b/src/Umbraco.Web/Models/PropertyEditorsConfigs/GridArea.cs
@@ -1,0 +1,11 @@
+using System.Collections.Generic;
+
+namespace Umbraco.Web.Models.PropertyEditorsConfigs
+{
+    public class GridArea
+    {
+        public int Grid { get; set; }
+        public bool AllowAll { get; set; }
+        public List<string> Allowed { get; set; }
+    }
+}
--- a/src/Umbraco.Web/Models/PropertyEditorsConfigs/GridBasic.cs
+++ b/src/Umbraco.Web/Models/PropertyEditorsConfigs/GridBasic.cs
@@ -1,0 +1,15 @@
+using System.Collections.Generic;
+
+namespace Umbraco.Web.Models.PropertyEditorsConfigs
+{
+    public class GridBasic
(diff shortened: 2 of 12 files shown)


tompipe:
Nice. But is there some overlap with the models at `Umbraco.Core.Models.GridValue` etc?

bielu:
@tompipe will look into that :)

bielu:
A quick overview of what is in models, which are overlapping:

| Model Name  | Umbraco.Core.Models                                                                                                                                                                               | Umbraco.Web.Models.PropertyEditorsConfigs |
|-------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------|
| GridSection |  public string Grid, public IEnumerable<GridRow> Rows |   public int grid,public bool AllowAll,public List<string> Allowed                                         |
| GridArea    | public string Grid, public IEnumerable<GridControl> Controls,public JToken Styles,public JToken Config                                                                                                                                                                                                  |   public int Grid,public bool AllowAll,public List<string> Allowed                                         |

bielu:
@ronaldbarendse 
>I've added some suggestions on the models, but they should probably be added as child classes of GridConfiguration (like ListViewConfiguration),
I don't really agree they should be child classes here and tbh I think Listview subclasses should be extracted and be separate classes but yeah here we will have a different opinion about things. Also, they shouldn't inherit from Jobject as we actually convert JTokens/Jobject to real object somewhere in code via Newtonsoft.Json.  Rest Of Comment in code looks ok and going to merge them into the commit. :)

bielu:
>The Umbraco.Core.Models.GridValue should be used with an IPropertyValueConverter, but that's also something on the to-do list:
True it should use also that will add that soon to my PR and update also GridValueConverter.cs

bielu:
@ronaldbarendse I just look into values which are propagated in both and they are different, ares in grid values != ares in backoffice :) it is why I didn't changed gridvalue :)

PR #7514, opened 2020-01-26
Access document, media and member types by guid and udi

bjarnef:
### Prerequisites

### Description
This PR enable the option to access document, media and member types by guid and udi and not only by id.

At the moment this it already possible for content, media and member nodes (and in Umbraco Forms as I remember) ... and also for datatypes and macros https://github.com/umbraco/Umbraco-CMS/pull/6847

### DIFF
--- a/src/Umbraco.Web/Editors/ContentTypeController.cs
+++ b/src/Umbraco.Web/Editors/ContentTypeController.cs
@@ -77,15 +77,58 @@ namespace Umbraco.Web.Editors
             return Services.ContentTypeService.HasContentNodes(id);
         }
 
+        /// <summary>
+        /// Gets the document type a given id
+        /// </summary>
+        /// <param name="id"></param>
+        /// <returns></returns>
         public DocumentTypeDisplay GetById(int id)
         {
-            var ct = Services.ContentTypeService.Get(id);
-            if (ct == null)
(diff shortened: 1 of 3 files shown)


emmaburstow:
Hiya @bjarnef,

thanks for the work here. It'll be really useful to be able to access document types by udi and guid! We'll take a look and let you know if we need anything.

Emma

bjarnef:
@nul800sebastiaan do you see any issues with this and similar PRs to request document by `Guid` and `Udi`?

bjarnef:
@emmaburstow it seems these PR's are a bit forgotten compared to new incoming PR's. Anything you need from me regarding these "access nodes from guid and udi" PR's? 😊🦄

nul800sebastiaan:
Lovely work @bjarnef - happy to merge!

ChrisMullaley:
I'm not sure if this is the right place to mention this but we are looking at upgrading or Umbraco version and hit an issue relating to this change.

We have client code that was calling ContentTypeController.GetById(int id). Those calls started failing after the update and I tracked it down to the fact that our client code adds a trailing slash to the URL. That was preventing the custom action selector from finding the right action method.

We've fixed it in the client code but it might be useful for others to know, It also means that there is a small inconsistency in how URLs are resolved between standard controller actions and these overloaded ones.

PR #7528, opened 2020-01-27
Access member group by id, guid and udi

bjarnef:
### Prerequisites

### Description
This PR enable the option to access member groups by guid and udi and not only by id in url.

### DIFF
--- a/src/Umbraco.Core/Services/Implement/MemberGroupService.cs
+++ b/src/Umbraco.Core/Services/Implement/MemberGroupService.cs
@@ -70,6 +70,14 @@ namespace Umbraco.Core.Services.Implement
             }
         }
 
+        public IMemberGroup GetById(Guid id)
+        {
+            using (var scope = ScopeProvider.CreateScope(autoComplete: true))
+            {
+                return _memberGroupRepository.GetMany().FirstOrDefault(x => x.Key == id);
+            }
+        }
+
         public IMemberGroup GetByName(string name)
         {
             using (var scope = ScopeProvider.CreateScope(autoComplete: true))
--- a/src/Umbraco.Core/Services/IMemberGroupService.cs
+++ b/src/Umbraco.Core/Services/IMemberGroupService.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using Umbraco.Core.Models;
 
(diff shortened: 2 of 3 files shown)


emmaburstow:
I see a theme 🤔 Thanks for this @bjarnef - speak soon.

Em

bjarnef:
@nul800sebastiaan one last og those access by id, guid and udi 😁😎👌

nul800sebastiaan:
@bjarnef Oh I missed this one! Thanks!

bjarnef:
Woop, more or less all is now accessible via id, guid and udi 👍🎉

Let me know if something is missing 😁

src/Umbraco.Core/Services/Implement/MemberGroupService.cs
```diff
@@ -70,6 +70,14 @@ public IMemberGroup GetById(int id)
             }
         }
 
+        public IMemberGroup GetById(Guid id)
+        {
+            using (var scope = ScopeProvider.CreateScope(autoComplete: true))
+            {
+                return _memberGroupRepository.GetMany().FirstOrDefault(x => x.Key == id);
```
nul800sebastiaan, line 77: @bjarnef I seem to have missed this line! I realize you had no other choice here, but this is not a friendly with regards to performance! It will get ALL the row from SQL and do filtering in C#, that can potentially be many rows and this repository doesn't cache the results I think.

Can you please look into adding a method to the `MemberGroupRepository` to do a single SQL query to get a group by it's `Key`?

bjarnef: Ahh yes. Need to have another look at this. I guess for most projects it won't be a huge performance issues, since no or a few member groups are used, but some projects might have a huge among of member groups.

Found this for media:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Services/Implement/MediaService.cs#L346-L358

nul800sebastiaan: Sure, I am more concerned about people using that method in all of their views so it will run on each request! Something like MediaService or ContentService would be good yep!

bjarnef: I have submitted a PR here: https://github.com/umbraco/Umbraco-CMS/pull/8496

Issue #7386, opened 2019-12-24
V8.4 - Loading or publishing a node that has a base64 image takes a long time


saifobeidat:
I tried to insert an image inside the RTE which its source is base64 code like the below: 

`<img src="data:image/jpeg;base64,/9j/4AAQSkZJRgABAQEAYABgAAD/2wB........">`

 The image immediatly appeared in the RTE, but when I saved & published the node, it took about 15 minutes to finish.

Then I tried to open that node, and the same time has been taken to to be loaded

## Umbraco version

I am seeing this issue on Umbraco version:  8.4

### Steps to reproduce
1- Create a new node with a RTE property
2- Insert an a base64 image inside the RTE
3- Save & publish

![2019-12-24_18-37-17](https://user-images.githubusercontent.com/24841143/71420184-8d800900-267c-11ea-8276-1acac42aa9cf.gif)

### Expected result
Saving & Publishing / Loading a node should be done quickly

### Actual result
Long time to finish

garpunkal:
I've noticed previously that if a base64 image is embedded in the RTE, it can cause random CPU spikes. This can cause websites to become unresponsive.  I'm guessing it's down to indexing. Once I identified this image was causing the problem, we removed it and the CPU spikes disappeared. 

I'd recommend not adding base64 encoded images into the RTE editor for this reason, or making sure this issue is addressed too.

saifobeidat:
> I've noticed previously that if a base64 image is embedded in the RTE, it can cause random CPU spikes. This can cause websites to become unresponsive. I'm guessing it's down to indexing. Once I identified this image was causing the problem, we removed it and the CPU spikes disappeared.
> 
> I'd recommend not adding base64 encoded images into the RTE editor for this reason, or making sure this issue is addressed too.

Yes, I wouldn't add base64 images to my website, but I am migrating content form external DB, and it has a lot of records containing base64 images, so it's not easy to remove those images.

nul800sebastiaan:
We don't specifically support base64 images in the RTE, and have seen this problem before when attempting to add them. we never investigated why this would create a CPU spike like this but we'd be happy if someone wanted to figure that one out.

umbrabot:
Hi @saifobeidat,

[...]

PR #7530, opened 2020-01-28
Improve RTE performance when using base64 images

asmailov:
Link to the issue: #7386 

### Description

**Changes:**

1. Changed image tag parsing **Regex** to be more performant
2. Added more **tests** (including one which tests performance)

**How to test:**

1. Add any document with a default rich text editor property
2. Use [this page](https://www.base64-image.de/) or any other to transform an image into a **base64** characters.
3. Insert `` into your document's property source.
4. Ensure that **Saving**, **Publishing**, **Reopening**, **Previewing** works fast.
5. Additionally, add media images and ensure that they are still working correctly.

### DIFF
--- a/src/Umbraco.Tests/Templates/HtmlImageSourceParserTests.cs
+++ b/src/Umbraco.Tests/Templates/HtmlImageSourceParserTests.cs
@@ -1,7 +1,6 @@
 ﻿using Umbraco.Core.Logging;
 using Moq;
 using NUnit.Framework;
-using Umbraco.Core.Services;
 using Umbraco.Tests.Testing.Objects.Accessors;
 using Umbraco.Web.Templates;
 using Umbraco.Web;
@@ -13,12 +12,10 @@ using System;
 using System.Linq;
 using Umbraco.Core.Models;
 using Umbraco.Core;
-using Umbraco.Web.PropertyEditors;
+using System.Diagnostics;
 
 namespace Umbraco.Tests.Templates
 {
-
-
     [TestFixture]
     public class HtmlImageSourceParserTests
     {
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Great stuff @asmailov! Thank you very much and wonderful to be adding some unit tests for this, top job! 🌟

I see this is your first PR for Umbraco-CMS as well, congrats on making it an aweseome one. Processing the Html is going down from ~50 seconds to just ~45 milliseconds on my machine, amazing! 👍 

Congrats on your first accepted PR for Umbraco-CMS, you've earned yourself a well deserved Contributor badge on your Our Umbraco profile: https://our.umbraco.com/members/id:279447/ 🎖 🏅

asmailov:
@nul800sebastiaan Glad, I could help 👍

nul800sebastiaan:
Hi again @asmailov - this seems to have broken the ability to insert an image into the Rich Text Editor, so either by using the insert button or by drag & dropping the image into the RTE. Any chance you can try that and see what is going wrong? I would've thought with all those unit tests we'd have some great coverage, but it seems like something is still missing!

nul800sebastiaan:
It's because `umbracoContextAccessor.UmbracoContext` is null so we never get to set up the `_getMediaUrl` Func.

image

nul800sebastiaan:
Fixed by updating the usage to be like the way `HtmlLocalLinkParser` uses the context. 

See commit https://github.com/umbraco/Umbraco-CMS/commit/fbd95f1c8edfc3afc45f3e758b7894a393a40b86

asmailov:
> Fixed by updating the usage to be like the way `HtmlLocalLinkParser` uses the context.
> 
> See commit [fbd95f1](https://github.com/umbraco/Umbraco-CMS/commit/fbd95f1c8edfc3afc45f3e758b7894a393a40b86)

Thank you, @nul800sebastiaan !
I don't often check the GitHub, so I only noticed the notification today 😞 

The mistake was assuming that the **UmbracoContext** will be available in the constructor. Previously we checked the existence of **UmbracoContext** in the public method which is run at a much later time. By moving validation into the constructor it broke the methods.

Issue #7507, opened 2020-01-24
Umbraco.Core/Udi.cs bug with Udi Create(Uri uri)


rbproworks:
I noticed what looks like a simple copy/paste issue or typo in the code for Umbraco-CMS/src/Umbraco.Core/Udi.cs, specifically the internal static Udi Create(Uri uri) method.

There are two if statements and the second one will currently never get called:

 if (udiType == UdiType.GuidUdi)
                return new GuidUdi(uri);
            **if (udiType == UdiType.GuidUdi)
                return new StringUdi(uri);**
The second "if" statement should presumably be:
**if (udiType == UdiType.StringUdi)
                return new StringUdi(uri);**

See also:
https://github.com/umbraco/Umbraco-CMS/blob/5a07e34fb4756004bc7175f28a96332bbe84b7ae/src/Umbraco.Core/Udi.cs/#L274

## Umbraco version
I am seeing this issue on Umbraco version: 8.x

I don't know the ramifications of this - it's an internal method and I'm not sure how StringUdi's are or are not being used currently.
![Umbraco-v8-Create-bug](https://user-images.githubusercontent.com/6393405/73112103-dfa58b80-3ec1-11ea-8110-875e4bd6a3f7.png)

nul800sebastiaan:
Interesting! Yes, that definitely seems like a copy/paste error. I have a lot of random sites laying around on my machine and can't find the error (Uri .. is not a valid udi) so I suspect this codepath is not often hit. 

But it would be very good to fix this nonetheless!

umbrabot:
Hi @rbproworks,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/7531

PR #7531, opened 2020-01-28
[FIX] Corrected Create method so that String UDI is correctly identified

NikRimington:
- [FIX] Changed check of UDI type to check for GUID type and then String type as previously was checking for
GUID type twice. This resulted in the method never creating a String UDI if needed.
 - [ISSUE] Github Issue #7507 https://github.com/umbraco/Umbraco-CMS/issues/7507

### Description

The change is exactly that described in the issue, however could not find where the code is used within the core to fully test it, but appears related to the UdiRangeJsonConverter

### DIFF
--- a/src/Umbraco.Core/Udi.cs
+++ b/src/Umbraco.Core/Udi.cs
@@ -316,7 +316,7 @@ namespace Umbraco.Core
 
             if (udiType == UdiType.GuidUdi)
                 return new GuidUdi(uri);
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks @NikRimington! I don't think there's any code in our codebase that will actually hit this path but in any case, we can clearly see this was a copy/paste error that needs to be fixed. All good, thanks again! ⭐️

PR #5449, opened 2019-05-11
Multi URL Picker enhancements: Content property and only save manually entered link name

ronaldbarendse:
This PR adds two enhancements to the recently integrated Multi URL Picker (https://github.com/umbraco/Umbraco-CMS/pull/2323):

**1. `Content` property on the `Link` model (as commented on https://github.com/umbraco/Umbraco-CMS/pull/2323#discussion_r249716153)**
Adding the `Content` property was quite simple, as the `MultiUrlPickerPropertyConverter` already gets the content or media as `IPublishedContent`. This allows easy access to any additional properties, e.g. an overview image that's rendered with the linked content or a thumbnail (with alt-text) of the linked media item. I think the ease of use justifies doing an additional query using the `Udi` (e.g. `Umbraco.TypedContent(link.Udi)` or `Umbraco.TypedMedia(link.Udi)`).

**2. Only save manually entered link name**
The current link picker overlay sets the link name to the selected content or media item, but this has the following problems:
- Entering a (custom) link name and then selecting a node overwrites the entered text (as the name/title input is shown before the node pickers, this is a quite common/understandable order of doing things);
- If the linked node name is changed, the link name is not automatically updated, so editors might have to find all referencing links to update the link name;
- Because of the above, translating content (by duplicating and editing or using other tools) requires more text to be changed and when changing the links to point to the duplicated child nodes, require the custom text to be stored somewhere, so it's not overwritten.

02qB9Ll3tS

PR #2323, opened 2017-11-13
Add Multi Url Picker to core

rasmusjp:
PR for [U4-8930](http://issues.umbraco.org/issue/U4-8930)

PR #7533, opened 2020-01-28
V8/feature/multiurlpicker enhancements

ronaldbarendse:
### Prerequisites

### Description
This is a Umbraco 8 re-implementation of PR https://github.com/umbraco/Umbraco-CMS/pull/5449.

**1. `Content` property on the `Link` model (as commented on https://github.com/umbraco/Umbraco-CMS/pull/2323#discussion_r249716153)**
Adding the `Content` property was quite simple, as the `MultiUrlPickerPropertyConverter` already gets the content or media as `IPublishedContent`. This allows easy access to any additional properties, e.g. an overview image that's rendered with the linked content or a thumbnail (with alt-text) of the linked media item.

Steps to test:
- Add links to content, media and external URL on a 'Umbraco.MultiUrlPicker' property
- Check whether the correct properties on the `Link` model that's returned from the PVC are set:
  - `Content` should only be `null` for external URLs
  - `Content.Key` should have the same GUID as in the `Udi`

**2. Only save manually entered link name**
The current link picker overlay sets the link name to the selected content or media item, but this has the following problems:
- Entering a (custom) link name and then selecting a node overwrites the entered text (as the name/title input is shown before the node pickers, this is a quite common/understandable order of doing things);
- If the linked node name is changed, the link name is not automatically updated, so editors might have to find all referencing links to update the link name;
- Because of the above, translating content (by duplicating and editing or using other tools) requires more text to be changed and when changing the links to point to the duplicated child nodes, require the custom text to be stored somewhere, so it's not overwritten (this obviously only applies to translations using multiple root nodes and not when using culture variants).

The interface/AngularJS part of this feature need to be changed, but the PVC already sets the content name when the link title/name is empty. When this is ready, I'll add testing steps and mark this PR as ready to review. Stay tuned!

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/ValueConverters/MultiUrlPickerValueConverter.cs
+++ b/src/Umbraco.Web/PropertyEditors/ValueConverters/MultiUrlPickerValueConverter.cs
@@ -50,6 +50,7 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
 
                 foreach (var dto in dtos)
                 {
+                    var name = dto.Name;
                     var type = LinkType.External;
                     var url = dto.Url;
                     IPublishedContent content = null;
@@ -77,12 +78,17 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
                             continue;
                         }
 
+                        if (string.IsNullOrEmpty(name))
+                        {
+                            name = content.Name;
+                        }
+
                         url = content.Url();
                     }
 
                     links.Add(new Link
                     {
-                        Name = dto.Name,
+                        Name = name,
                         Target = dto.Target,
                         Type = type,
                         Udi = dto.Udi,
--- a/src/Umbraco.Web/PropertyEditors/ValueConverters/MultiUrlPickerValueConverter.cs
+++ b/src/Umbraco.Web/PropertyEditors/ValueConverters/MultiUrlPickerValueConverter.cs
@@ -1,8 +1,7 @@
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Newtonsoft.Json;
 using Umbraco.Core;
 using Umbraco.Core.Logging;
 using Umbraco.Core.Models.PublishedContent;
@@ -15,30 +14,29 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
     public class MultiUrlPickerValueConverter : PropertyValueConverterBase
     {
         private readonly IPublishedSnapshotAccessor _publishedSnapshotAccessor;
-        private readonly IProfilingLogger _proflog;
+        private readonly IProfilingLogger _profilingLogger;
 
-        public MultiUrlPickerValueConverter(IPublishedSnapshotAccessor publishedSnapshotAccessor, IProfilingLogger proflog)
+        public MultiUrlPickerValueConverter(IPublishedSnapshotAccessor publishedSnapshotAccessor, IProfilingLogger profilingLogger)
         {
             _publishedSnapshotAccessor = publishedSnapshotAccessor ?? throw new ArgumentNullException(nameof(publishedSnapshotAccessor));
-            _proflog = proflog ?? throw new ArgumentNullException(nameof(proflog));
+            _profilingLogger = profilingLogger ?? throw new ArgumentNullException(nameof(profilingLogger));
         }
 
         public override bool IsConverter(IPublishedPropertyType propertyType) => Constants.PropertyEditors.Aliases.MultiUrlPicker.Equals(propertyType.EditorAlias);
 
-        public override Type GetPropertyValueType(IPublishedPropertyType propertyType) =>
-            propertyType.DataType.ConfigurationAs<MultiUrlPickerConfiguration>().MaxNumber == 1 ?
-                typeof(Link) :
-                typeof(IEnumerable<Link>);
+        public override Type GetPropertyValueType(IPublishedPropertyType propertyType) => propertyType.DataType.ConfigurationAs<MultiUrlPickerConfiguration>().MaxNumber == 1
+            ? typeof(Link)
+            : typeof(IEnumerable<Link>);
 
         public override PropertyCacheLevel GetPropertyCacheLevel(IPublishedPropertyType propertyType) => PropertyCacheLevel.Snapshot;
 
-        public override bool? IsValue(object value, PropertyValueLevel level) => value?.ToString() != "[]";
+        public override bool? IsValue(object value, PropertyValueLevel level) => value?.ToString() is var jsonValue && !string.IsNullOrEmpty(jsonValue) && jsonValue != "[]";
 
         public override object ConvertSourceToIntermediate(IPublishedElement owner, IPublishedPropertyType propertyType, object source, bool preview) => source?.ToString();
 
         public override object ConvertIntermediateToObject(IPublishedElement owner, IPublishedPropertyType propertyType, PropertyCacheLevel referenceCacheLevel, object inter, bool preview)
         {
-            using (_proflog.DebugDuration<MultiUrlPickerValueConverter>($"ConvertPropertyToLinks ({propertyType.DataType.Id})"))
+            using (_profilingLogger.DebugDuration<MultiUrlPickerValueConverter>($"ConvertPropertyToLinks ({propertyType.DataType.Id})"))
             {
                 var maxNumber = propertyType.DataType.ConfigurationAs<MultiUrlPickerConfiguration>().MaxNumber;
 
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Thanks so far! Before you continue on this: both of these sound like breaking changes in behavior might cause problems in upgraded website.

I also quickly skimmed the old PR and saw a new setting was added which we generally try to avoid.

ronaldbarendse:
Currently, there aren't any breaking changes: only an additional `Content` property on the `Link` model and better handling of unsupported UDIs in the PVC.

I've only added the additional overlay setting `useNodeName` in the V7 PR to keep backwards-compatibility in case the overlay is used in other places that don't expect a non-empty link title (e.g. the RJP Multi URL Picker). I'll rewrite this in a way that doesn't require this setting, e.g. by inferring this when the `currentTarget` object contains a `title` property (besides `name`)...

The additional 'Hide target' setting on the data type can't be avoided and seems like a very sensible feature, as you might not want to use this and/or want to make it easier/cleaner/friendlier for editors.

The PVC can also be slightly optimized to only process the configured `maxNumber` of (valid) links, instead of adding all processed links to a list and then returning only a subset. I understand it needs to switch from a list to single item (as the return type changes when `maxNumber` is set to `1`), but I don't get why it returns a subset: lowering the `maxChars` on a textbox property also doesn't truncate the text, right?

src/Umbraco.Web/PropertyEditors/ValueConverters/MultiUrlPickerValueConverter.cs
```diff
@@ -15,77 +14,102 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
...
+                            case Core.Constants.UdiEntityType.Media:
+                                type = LinkType.Media;
+                                content = _publishedSnapshotAccessor.PublishedSnapshot.Media.GetById(preview, dto.Udi.Guid);
+                                break;
+                            default:
+                                // Skip unsupported entity types
+                                continue;
+                        }
 
-                        if (content == null || content.ContentType.ItemType == PublishedItemType.Element)
```
ronaldbarendse: How can the content type ever be an element if it's retrieved from the snapshots content/media cache?

src/Umbraco.Web/PropertyEditors/ValueConverters/MultiUrlPickerValueConverter.cs
```diff
@@ -15,77 +14,102 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
...
+                    {
+                        Name = name,
+                        Target = dto.Target,
+                        Type = type,
+                        Udi = dto.Udi,
+                        Content = content,
+                        Url = url + dto.QueryString
+                    });
+
+                    if (maxNumber > 0 && links.Count == maxNumber)
```
ronaldbarendse: This currently ensures that no more links are processed then needed. However, this behavior is a little bit strange, as lowering the `maxNumber` will truncate the list of already selected, valid links. Lowering the `maxChars` on a Textbox or `maxItems` on Nested Content won't do this. Yes, it should validate the maximum number when publishing, but let's keep the behavior of all default Umbraco PVCs the same.

So we only have to short-circuit when we're in single item mode:
```suggestion
                    if (maxNumber == 1)
```

src/Umbraco.Web/PropertyEditors/ValueConverters/MultiUrlPickerValueConverter.cs
```diff
@@ -15,77 +14,102 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
...
+                    {
+                        // Stop processing
+                        break;
+                    }
+                }
+
+                // Return the correct value type
+                if (maxNumber == 1)
```
ronaldbarendse: We're only returning here (and not in the foreach loop above) to make sure it always returns the correct value type. If all links are unpublished content, the list will be empty and it will correctly return `null`.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/multiurlpicker/multiurlpicker.controller.js
```diff
@@ -144,16 +144,6 @@ function multiUrlPickerController($scope, angularHelper, localizationService, en
...
-
-        _.each($scope.model.value, function (item){
-            // we must reload the "document" link URLs to match the current editor culture
-            if (item.udi.indexOf("/document/") > 0) {
-                item.url = null;
-                entityResource.getUrlByUdi(item.udi).then(function (data) {
-                    item.url = data;
-                });
-            }
-        });
```
ronaldbarendse: The PR also contains the fix applied in PR https://github.com/umbraco/Umbraco-CMS/pull/7543, so the Multi URL Picker works again with external URLs.

src/Umbraco.Web/PropertyEditors/MultiUrlPickerConfiguration.cs
```diff
@@ -11,13 +10,12 @@ public class MultiUrlPickerConfiguration : IIgnoreUserStartNodesConfig
...
-        [ConfigurationField(Core.Constants.DataTypes.ReservedPreValueKeys.IgnoreUserStartNodes,
-            "Ignore User Start Nodes", "boolean",
+        [ConfigurationField(Core.Constants.DataTypes.ReservedPreValueKeys.IgnoreUserStartNodes, "Ignore user start nodes", "boolean",
             Description = "Selecting this option allows a user to choose nodes that they normally don't have access to.")]
         public bool IgnoreUserStartNodes { get; set; }
 
-        [ConfigurationField("hideAnchor",
-            "Hide anchor/query string input", "boolean",
+        // TODO Change this to hideQueryString, as that's the name used in the picker
+        [ConfigurationField("hideAnchor", "Hide anchor/query string input", "boolean",
```
ronaldbarendse: This configuration key doesn't correspond to the Multi URL Picker name, both `LinkDisplay` and `LinkDto` saves it as `queryString`... Although the linkpicker overlay calls it `anchor` again 🤔

src/Umbraco.Web/PropertyEditors/MultiUrlPickerValueEditor.cs
```diff
@@ -133,16 +133,16 @@ public override object FromEditor(ContentPropertyData editorValue, object curren
...
+                    JsonConvert.DeserializeObject<List<LinkDisplay>>(value).Select(link => new LinkDto
                     {
                         Name = link.Name,
-                        QueryString = link.QueryString,
                         Target = link.Target,
+                        QueryString = link.QueryString,
                         Udi = link.Udi,
-                        Url = link.Udi == null ? link.Url : null, // only save the url for external links
-                    },
+                        Url = link.Udi == null ? link.Url : null // Only save the url for external links
```
ronaldbarendse, line 144:
```suggestion
                        Url = link.Udi == null ? link.Url : null // Only save the URL for external links
```

src/Umbraco.Web.UI.Client/src/views/propertyeditors/multiurlpicker/multiurlpicker.controller.js
ronaldbarendse: I've now applied the merged commit from PR https://github.com/umbraco/Umbraco-CMS/pull/7623.

src/Umbraco.Web/PropertyEditors/MultiUrlPickerConfiguration.cs
ronaldbarendse: Let's keep this the way it is, as changing the name would cause a breaking change.

ronaldbarendse:
I've added the option to hide the target checkbox and fixed the link picker display when the anchor/query is hidden, so this PR is ready to get reviewed! (tagging @nul800sebastiaan and @poornimanayar as this isn't a new PR)

FYI the option to only save the manually entered link name requires a lot more work, so I've removed that from this PR...

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/linkpicker/linkpicker.controller.js
```diff
@@ -92,9 +92,8 @@ angular.module("umbraco").controller("Umbraco.Editors.LinkPickerController",
                         $scope.anchorValues = resp.anchorValues;
                         $scope.model.target.url = resp.url;
                     });
-
                 }
-            } else if ($scope.model.target.url && $scope.model.target.url.length) {
+            } else if ($scope.model.target.url && $scope.model.target.url.length && !dialogOptions.hideAnchor) {
```
ronaldbarendse, line 96: This ensures the anchor is not populated from the URL when the anchor/query string field is hidden. Otherwise it's possible to enter a URL with an anchor/query string, submit and then not be able to edit this anymore (as it's split from the URL and that field is hidden).

An existing anchor value is kept however (it's only hidden after all), so this might result in duplicate anchor/query strings (the one in the URL, because it's not split anymore and the previously saved value).

ronaldbarendse:
@nul800sebastiaan I've fixed the conflict on `en.xml`, although it now shows additions that are already in `v8/contrib`, so I don't know how that will merge now (just beware 😉)...

ronaldbarendse:
> @nul800sebastiaan I've fixed the conflict on `en.xml`, although it now shows additions that are already in `v8/contrib`, so I don't know how that will merge now (just beware 😉)...

I've manually removed the additions that were not a part of this PR and force-pushed the existing merge commit, so everything is in order again and ready for review/merge 👍

ronaldbarendse:
Looks like the `MultiUrlPickerValueConverter` is also missing the `DefaultPropertyValueConverter`, so that results in the same issue as for Nested Content (https://github.com/umbraco/Umbraco-CMS/pull/2918).

ronaldbarendse:
> Looks like the `MultiUrlPickerValueConverter` is also missing the `DefaultPropertyValueConverter`, so that results in the same issue as for Nested Content (#2918).

This is now also fixed in the latest commit 👍

SimonAntony:
Looks like a pretty solid implementation here @ronaldbarendse - thanks for sorting it, any idea when it maybe approved?

PR #7535, opened 2020-01-29
Adding more fine colors for UI

nielslyngsoe:
Adding a more bright pink for fills on white, to make it less aggressive, less dangerous.

See example here, from the content-tree:

image


kjac:
Merged ⭐️

PR #7536, opened 2020-01-29
make icon of umb-preview-node align vertically.

nielslyngsoe:
Making the icon of umb-node-preview align vertically.

image


nielslyngsoe:
@kjac cool, thanks for your thorough test!

PR #7542, opened 2020-01-30
Moved PublishedCache from Umbraco.Web into .Net standard project (#7551)

AndyButland:
This is very rough - not yet for review/merging - sharing for other eyes to see.

Related issue: #7551

In this update we've taken the _PublishedCache_ folder from Umbraco.Web and moved into a new .Net standard project.  Returned to compiling state with a number of things to check and some temporarily commented out functionality.

Things to consider:

- Caching of the creator and writer names.  Retrieving these require `IUserService`, but we don't want to inject that to avoid a reference to it being cached with the `PublishedContent` object.
- Caching of the URL - requires access to a `UrlProvider`
- There's one use of a `PublishedContentExtension` in `PublishedCacheBase.GetByContentType`, which are in `Umbraco.Web` and have some uses of service location that need removing.
- There's some use of the concrete type `PublishedSnapshotService` instead of `IPublishedSnapshotService` - some of these have been changed for the interface, some we need to check does the calling code really need to know about the implementation class?


AndyButland:
Have pushed an update resolving the first issue by removing the `WriterName` and `CreatorName` from `IPublishedContent` and replacing functionality with 2 new extension methods in `Umbraco.Core.PublishedContentExtensions`.

AndyButland:
Further update to similarly remove `Url` from `IPublishedContent` and update usages to use the existing `Url()` extension method.

stevetemple:
Create a PR for marking WriterName & CreatorName in v8 #7544 which this then removes

stevetemple:
There was a new'ing of the concrete type PublishedElement inside the Nested Content code. This should be managed through something like a factory class.

AndyButland:
Further update for the third issue noted in the original PR description, have migrated various `IPublishedContent` extension methods from web to abstractions project, removing use of service location and adding the dependencies to the method signature.

Then have restored the original methods, passing through to the migrated ones, providing the service located dependency.

Finally have used one of these extensions in the new `Umbraco.Infrastructure.PublishedCache` project.

AndyButland:
I think this PR is at a point where it can be reviewed now, so I'll recap what I've done, issues found, and how they've been resolved in this comment.

- Have created a new .Net standard project `Umbraco.Infrastructure.PublishedCache` (but not changed any namespaces for now... think I understood that was better to leave at the moment to aid merges from V8, but can update to the new location if that's better).
- Moved the `Umbraco.Web.PublishedCache` folder into here, with the exception of `UmbracoContextPublishedSnapshotAccessor` that has dependencies that suggest it still needs to be in `Umbraco.Web`.
- Have updated project references such that the new project only references other "core level" projects (i.e. `Abstractions` and `Infrastructure`).
- Added the external reference to the forked .Net standard B-tree implementation.

In doing this I came across the following issues:

- Caching of the creator and writer names. Retrieving these require `IUserService`, but we don't want to inject that to avoid a reference to it being cached with the `PublishedContent` object (there was a comment flagging that we want to avoid this, and currently do so by using service location via `Current`, but as that's defined in `Umbraco.Web`, it's not available any more.
    - To resolve this I've removed `CreatorName` and `WriterName` from `IPublishedContent` and classes implementing this interface.  I've created a new `CreatorName() / WriterName()` extension methods in `Umbraco.Web` to be used as alternatives, that delegate to further extension methods in `Umbraco.Abstraction`, `GetCreatorName(IUserService userService) / GetWriterName(IUserService userService)` (so we have a "friendly" extension that uses service location, but have a more correct one where we pass the dependency to use within the code-base).
- Similar issue with the `Url` property on `IPublishedContent`.
    - Given this is context dependent and requires a `UrlProvider`, I've resolved this in a similar way: removing the field from `IPublishedContent` and creating a `Url()` extension method.
- There was one use of a `PublishedContentExtension` in `PublishedCacheBase.GetByContentType`, which was defined in `Umbraco.Web` and so no longer accessible.
	- For this one I've migrated the extension method (along with a load of other related ones) into `Umbraco.Abstractions`, and added additional parameters to the signatures where dependencies are required.  I've kept the original, "friendly", extension methods that use service location though, but changed their implementation to delegate to the new ones.
	
	
So for example we now have in `Umbraco.Web`:

```
private static IVariationContextAccessor VariationContextAccessor => Current.VariationContextAccessor;

public static IPublishedContent DescendantOrSelf(this IPublishedContent content, int level, string culture = null)
{
    return content.DescendantOrSelf(VariationContextAccessor, level, culture);
}
```

That delegates to this in `Umbraco.Abstractions`:

```
public static IPublishedContent DescendantOrSelf(this IPublishedContent content, IVariationContextAccessor variationContextAccessor, int level, string culture = null)
{
    ...
}
```
	
- There were a couple of instances where the concrete type `PublishedSnapshotService` is used, rather than `IPublishedSnapshotService`.  Where it seems to make sense I've added a couple of methods to the interface, and where it looks to be very "nucache specific", I've left the type checking.

All tests are passing, a cursory test of the back-office and template rendering seems OK, and I've tested out the new extension methods in a front-end template.

// @bergmania @Shazwazza @stevetemple

bergmania:
I also fixed some views that should reference new extension methods..

src/Umbraco.Web/Editors/NuCacheStatusController.cs
```diff
@@ -17,24 +17,24 @@ public NuCacheStatusController(IPublishedSnapshotService publishedSnapshotServic
...
-        private PublishedSnapshotService PublishedSnapshotService
+        private IPublishedSnapshotService PublishedSnapshotService
         {
             get
             {
-                var svc = _publishedSnapshotService as PublishedSnapshotService;
-                if (svc == null)
-                    throw new NotSupportedException("Not running NuCache.");
-                return svc;
+                // TODO: do we need this?
```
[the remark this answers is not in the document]
stevetemple: I think this controller might need renaming if it's more generic. 

Perhaps the IPublishedSnapshotService should have a method, or preferably there is another interface that indicates whether it allows for a UI and provides the methods the UI uses?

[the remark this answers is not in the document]
bergmania: I also thought about renaming to a more generic - E.g. `PublishedSnapshotCacheStatusController`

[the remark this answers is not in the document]
AndyButland: Have removed and set to use interface, and renamed controller to more generic name as suggested (and related front-end assets).

src/Umbraco.Web.UI/Umbraco.Web.UI.csproj
```diff
@@ -382,7 +382,9 @@
     <Message Text="Skip Belle because UmbracoBuild is '$(UmbracoBuild)' (this is not Visual Studio)." Importance="High" Condition="'$(UmbracoBuild)' != ''" />
     <Message Text="Skip Belle because $(ProjectDir)Umbraco\lib exists." Importance="High" Condition="Exists('$(ProjectDir)Umbraco\lib')" />
     <Message Text="Build Belle because UmbracoBuild is empty (this is Visual Studio), and $(ProjectDir)Umbraco\lib does not exist." Importance="High" Condition="!Exists('$(ProjectDir)Umbraco\lib') and '$(UmbracoBuild)' == ''" />
+    <!--
```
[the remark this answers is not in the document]
AndyButland: Restored (had removed as was getting some problems locally, and didn't mean to check-in).

src/Umbraco.Infrastructure.PublishedCache/PublishedCacheBase.cs
```diff
@@ -10,6 +10,13 @@ namespace Umbraco.Web.PublishedCache
 {
     abstract class PublishedCacheBase : IPublishedCache
     {
```
Shazwazza: This class can exist in abstractions, it doesn't need to exist in the PublishedCache project, it has no ties to NuCache and could be used by other implementations

src/Umbraco.Infrastructure.PublishedCache/PublishedElement.cs
```diff
@@ -14,7 +14,7 @@ namespace Umbraco.Web.PublishedCache
     // an entirely new models factory + not even sure it makes sense at all since
     // sets are created manually todo yes it does! - what does this all mean?
     //
-    internal class PublishedElement : IPublishedElement
```
Shazwazza: This should exist in abstractions, it has no ties to NuCache and would be used by other implementations

src/Umbraco.Infrastructure.PublishedCache/PublishedMember.cs
```diff
@@ -141,10 +141,6 @@ private void EnsureMemberProperty(List<IPublishedProperty> properties, List<stri
 
         public override string UrlSegment => throw new NotSupportedException();
```
Shazwazza: This class should exist in Abstractions, it has no ties to nucache and could be re-used for a different implementation.

src/Umbraco.Infrastructure.PublishedCache/PublishedSnapshotServiceBase.cs
```diff
@@ -38,5 +38,13 @@ public virtual void Rebuild()
 
         public virtual void Dispose()
```
Shazwazza: Should this exist in Abstractions?

bergmania:
I agree with the naming of the project `Umbraco.Infrastructure.PublishedCache` => `Umbraco.PublishedCache`.

But we already have the SqlCe project, that also has `Infrastructure` in the name, I fix up in dev branch..

bergmania:
Alright, I addressed all of @Shazwazza's comments..

- The project is now named: `Umbraco.PublishedCache.NuCache`
- Classes in the root are moved to `Abstractions` or `Infrastracuture`
  - Then I was able to remove the reference from `Umbraco.Web` to `Umbraco.PublishedCache.NuCache`
- The content of the NuCache folder is moved to the root
- I introduced an `IPublishedCacheComposer` in Abstractions
  - The `NuCacheComposer` now implements this
  - I updated the `ModelsBuilderComposer ` to compose before `IPublishedCacheComposer` instead of `NuCacheComposer`
    - Then I was able to remove the reference from `Umbraco.ModelsBuilder.Embedded` to `Umbraco.PublishedCache.NuCache`
- I added a reference from `Umbraco.Web.UI` to `Umbraco.PublishedCache.NuCache` just to ensure the dll is copied to the bin folder

src/Umbraco.Abstractions/PublishedCache/IPublishedElementFactory.cs
```diff
@@ -0,0 +1,7 @@
+namespace Umbraco.Web.PublishedCache
+{
+    public interface IPublishedElementFactory
```
Shazwazza: What is this? this seems to be a brand new addition but is not used anywhere. Perhaps this was the start of trying to resolve this coment? https://github.com/umbraco/Umbraco-CMS/pull/7542#issuecomment-580715415 

In the future, and something we wanted to do in v8 but didn't get a chance - was to overhaul IPublishedContent/IPublishedElement - so that it's just a poco model. Ideally then all traversing and 'doing' operations are done with a service. This would make it so much easier to create a new published content implementation since we would just re-use the poco models and so the models are not also services themselves. But anyways, that's for later. In the meantime, i'm not sure what this is.

bergmania: You are right, it was my first attempt to fix https://github.com/umbraco/Umbraco-CMS/pull/7542#issuecomment-580715415. 
Realised that it was not necessary after I moved PublishedElement to Abstraction. I removed the interface again

Issue #7326, opened 2019-12-12
V8: Clean up language selectors in the treepicker


kjac:
In the early days of V8 we could switch between languages in the treepicker when selecting content e.g. from MNTP. This just wasn't terribly useful nor user friendly, so it was changed into always displaying the content nodes in the language currently being edited (`cculture`/`mculture`).

However... the code for the treepicker language selector still lingers, and has most recently caused the issue described in #7313.

I would really like to remove the whole thing from the treepicker. It isn't used anywhere in the core anymore, it won't get regression tested and is bound to stop working at some point (as a matter of fact I don't even know if it works now).

But... removing it is a breaking change on the clientside, as 3rd party packages *could* be using the feature.

Paging @nul800sebastiaan and @nielslyngsoe for some thoughts on this.

nul800sebastiaan:
Thanks @kjac - we had a chat about this and I learned that we're looking into re-introducing the language picker at some point (you might have very good reasons to link to a Spanish page from the English site). So we'll leave the code for now for when we are ready to start using it again! 👍

PR #7543, opened 2020-01-30
8.6RC: Support culture variant URLs in multi URL picker

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7114.

### Description
This corrects PR https://github.com/umbraco/Umbraco-CMS/pull/7130, as that throws JavaScript errors after adding an external link: `Cannot read property 'indexOf' of null`.

The Multi URL Picker already fetches the current URL in the `MultiUrlPickerValueEditor`, but didn't pass the culture to the `Url()` method. So this also removes the need for an additional API call (per linked content/media item!) to fetch the URL in the current culture...

While testing, I also noticed culture invariant content was displayed as unpublished on a culture variant node. This PR also ensures the `MultiUrlPickerValueEditor` checks whether the linked content is culture variant and only then checks the published cultures.

---
_This item has been added to our backlog [AB#4906](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4906)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/multiurlpicker/multiurlpicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/multiurlpicker/multiurlpicker.controller.js
@@ -144,16 +144,6 @@ function multiUrlPickerController($scope, angularHelper, localizationService, en
         if ($scope.model.validation && $scope.model.validation.mandatory && !$scope.model.config.minNumber) {
             $scope.model.config.minNumber = 1;
         }
-
-        _.each($scope.model.value, function (item){
-            // we must reload the "document" link URLs to match the current editor culture
-            if (item.udi.indexOf("/document/") > 0) {
-                item.url = null;
-                entityResource.getUrlByUdi(item.udi).then(function (data) {
-                    item.url = data;
-                });
-            }
-        });
     }
 
     init();
--- a/src/Umbraco.Web/PropertyEditors/MultiUrlPickerValueEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/MultiUrlPickerValueEditor.cs
@@ -79,7 +79,7 @@ namespace Umbraco.Web.PropertyEditors
                         if (entity is IDocumentEntitySlim documentEntity)
                         {
                             icon = documentEntity.ContentTypeIcon;
-                            published = culture == null ? documentEntity.Published : documentEntity.PublishedCultures.Contains(culture);
+                            published = culture == null || !documentEntity.Variations.VariesByCulture() ? documentEntity.Published : documentEntity.PublishedCultures.Contains(culture);
                             udi = new GuidUdi(Constants.UdiEntityType.Document, documentEntity.Key);
                             url = _publishedSnapshotAccessor.PublishedSnapshot.Content.GetById(entity.Key)?.Url(culture) ?? "#";
                             trashed = documentEntity.Trashed;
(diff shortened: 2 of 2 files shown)


kjac:
@ronaldbarendse thank you for this. We'll have a look at it as soon as possible 👍

ronaldbarendse:
@kjac The real issue is not storing the culture of the selected content (https://github.com/umbraco/Umbraco-CMS/issues/7326) or making the selected back-office culture available to the `IDataValueEditor.ToEditor()` method. So the URL can only be fetched in the same culture as the content/property or the default language when it's invariant.

The selected back-office culture is available in the `MapperContext`, but as the culture is explicitly set to `null` to ensure it's getting the correct stored value of the property, this is not passed to the `ToEditor()` method:
https://github.com/umbraco/Umbraco-CMS/blob/33f73379c501d30d6d6ffa3931f47eddb1bd08d4/src/Umbraco.Web/Models/Mapping/ContentPropertyBasicMapper.cs#L61-L79

Maybe this could be fixed by injecting the `IVariationContextAccessor` into the `MultiUrlPickerValueEditor` and making that also work within the back-office?

Reducing XHRs can be done by only fetching the URLs when the content/property is invariant...

ronaldbarendse:
Also, the current fix that's also used in the MNTP doesn't work correctly when in split-view: it shows the URL of the first selected content culture (left side) on the opened split-view culture (right side).

Shazwazza:
I've mentioned this on other issues... property editors can *not* use the mculture $routeParams to detect culture. If any property editor requires to know what culture is being edited, the property editor must become an angular component and then use required directives to get the umbraco content directive to know which culture is being edited. I'm assuming thats the issue you are seeing with split view.

Shazwazza:
Hi @kjac + @ronaldbarendse I haven't had time to fully digest this whole issue and the new issues caused by the 'fixes' and now the fix for the fix :) But I'm hoping you guys can provide me with a status and if this is blocking a release of 8.6?

kjac:
@Shazwazza #7623 fixes the issue for now. As far as I know it'll be a sufficient fix for 8.6. Granted it might work badly in split view like MNTP but at least the multi url picker doesn't crash on non-local links. 

This PR is being kept open in the hope that a better and generic solution can be devised to handle the localization challenge for both editors on the server side - preferably in a non-breaking manner.

Shazwazza:
@kjac ok thanks, i'll update our internal notes with this then. Side note - I don't know if part of the cause for split view issues that you are referring to are caused by using mculture route params.... if so, it's because absolutely no editor should ever rely on mculture, unless it's a top level editor. If you need access to the current culture in a property editor, your property editor must be a component so it can reference the higher up `umbProperty` or `umbVariantContent` component and use it's referenced API to get the current edited content's culture.

kjac:
@Shazwazza agreed. I haven't seen the split view issue myself (haven't tested it), but it's been reported for MNTP.

ronaldbarendse:
@Shazwazza It looks like the change in https://github.com/umbraco/Umbraco-CMS/pull/7130 doesn't fetch the current culture in the browser, but gets it on the server side using `Request.ClientCulture()`:

https://github.com/umbraco/Umbraco-CMS/blob/4f7b62f3692aa123bda3238f3f5dced874db7e73/src/Umbraco.Web.UI.Client/src/views/propertyeditors/multiurlpicker/multiurlpicker.controller.js#L152

https://github.com/umbraco/Umbraco-CMS/blob/4f7b62f3692aa123bda3238f3f5dced874db7e73/src/Umbraco.Web/Editors/EntityController.cs#L250

Shazwazza:
@ronaldbarendse it's been a very long time since i looked at any of that but the `ClientCulture` code gets it's value from a header `X-UMB-CULTURE`  which is set with an angular interceptor. I still haven't looked into this issue on this link so i don't understand the underlying issue (forgive me!), i'm just giving you guys some info as I have it :)

PR #7544, opened 2020-01-31
Obsolete CreatorName and WriterName on IPublishedContent

stevetemple:
_Creating this PR to trigger a discussion on the impact of removing CreatorName and WriterName from IPublishedContent_

### Description

We've been looking at things that make the move to netcore more difficult. One of the things that came up is PublishedContent's dependency on IUserService. 

It seemed to the Unicore team that CreatorName and WriterName are rarely used and could easily be replaced by an extension method. This would remove service location from IPublishedContent implementations and a bunch of TODOs and other code that feels a bit hacky.

If we were to remove them in future making them obsolete now would seem sensible so Umbraco developers could see that these are going to vanish and update now.

### DIFF
--- a/src/Umbraco.Web/PublishedCache/PublishedMember.cs
+++ b/src/Umbraco.Web/PublishedCache/PublishedMember.cs
@@ -140,9 +140,13 @@ namespace Umbraco.Web.PublishedCache
         public override string UrlSegment => throw new NotSupportedException();
 
         // TODO: ARGH! need to fix this - this is not good because it uses ApplicationContext.Current
+        [Obsolete("Use WriterName(IUserService) extension instead")]
+
         public override string WriterName => _member.GetCreatorProfile().Name;
 
         // TODO: ARGH! need to fix this - this is not good because it uses ApplicationContext.Current
+        [Obsolete("Use CreatorName(IUserService) extension instead")]
+
         public override string CreatorName => _member.GetCreatorProfile().Name;
 
         public override int WriterId => _member.CreatorId;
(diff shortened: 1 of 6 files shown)


src/Umbraco.Web/PublishedCache/PublishedMember.cs
```diff
@@ -140,9 +140,11 @@ private void EnsureMemberProperty(List<IPublishedProperty> properties, List<stri
         public override string UrlSegment => throw new NotSupportedException();
 
         // TODO: ARGH! need to fix this - this is not good because it uses ApplicationContext.Current
+        [Obsolete("Use WriterName(IUserService) extension instead")]
         public override string WriterName => _member.GetCreatorProfile().Name;
 
         // TODO: ARGH! need to fix this - this is not good because it uses ApplicationContext.Current
```
stevetemple: Stuff like this highlights to me that these need refactoring because they've proven tough to implement nicely in the past

ravimotha:
Can this information be recorded in a format that is not dependent on iuser..

Created by Ravi 31012020 as a string.. or two strings

It is not useful for most instances but when inheriting a site or new dev it’s sometimes useful to know who did what ie created a do type or a page..(mostly for blame)
Other wise it’s mostly redundant

stevetemple:
> Can this information be recorded in a format that is not dependent on iuser..
> 
> Created by Ravi 31012020 as a string.. or two strings
> 
> It is not useful for most instances but when inheriting a site or new dev it’s sometimes useful to know who did what ie created a do type or a page..(mostly for blame)
> Other wise it’s mostly redundant

We have the Id so it's just the friendly name that we're dropping which feels to me redundant if you have the id. But maybe a bit slower to get the name if you're doing dozens of nodes.

stevetemple:
Now that #7542 is merged, this should be merged

PR #7550, opened 2020-02-01
Obsolete Url on IPublishedContent

stevetemple:
### Description

Part of the asp-netcore dev work

The Url property on published content, which is very context dependant causes issue to move to netcore. There is already an extension method to replace it so seems sensible to remove it on the next major release. 

Marking obsolete here to start letting developers know that it's likely to be removed.

### DIFF
--- a/src/Umbraco.Core/Models/PublishedContent/IPublishedContent.cs
+++ b/src/Umbraco.Core/Models/PublishedContent/IPublishedContent.cs
@@ -97,6 +97,7 @@ namespace Umbraco.Core.Models.PublishedContent
         /// <para>The value of this property is contextual. It depends on the 'current' request uri,
         /// if any.</para>
         /// </remarks>
+        [Obsolete("Use the Url() extension instead")]
         string Url { get; }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


stevetemple:
Now that #7542 is merged this should be merged

rasolo:
Hello, where is this extension method that I can replace with? I can't figure out how to find it lol.

rasolo:
I saw the extension method being used in the code for the commit here and I can find it in the Umbraco source project but I can't find it and use it in my personal project which uses Umbraco 8.7.0

ronaldbarendse:
What part of the `Url` property is _very context dependant_ to cause issues? Is it because it returns different values depending on the culture (from `VariationContext`), parent nodes and domain name? The supplied `Url()` extension method does not allow specifying the current URL (like `UrlProvider.GetUrl(...)` does), so its still context dependant, right?

@rasolo The `Url()` extension method is in `Umbraco.Web.PublishedContentExtensions`, so you need to reference `Umbraco.Web` 👍 

Looks like the `UrlSegment` and `Name` properties are also context dependant (they can vary by culture) and have similair extension methods, although in the `Umbraco.Core` assembly. So this all seems more like a problem with seperating the Core and Web assemblies and having an `Url` property in Core that actually requires compoments in Web breaks the [clean architecture](https://docs.microsoft.com/en-us/dotnet/architecture/modern-web-apps-azure/common-web-application-architectures#clean-architecture).

rasolo:
Thanks I was able to use the extension method but it breaks my unit tests. Moq does not allow for using extension methods. If I don't mock the url property it will throw an error with the Umbraco Mapper library created by Andy Butland, it will say that the object (ipublishedcontent) is null because the url property is not set. If I use the extension method wrapped in a new class it will complain about the IPublishedContent.

My unit tests in the future will be broken once the old .Url property is removed :( I will ask Andy too if it can be solved

PR #7553, opened 2020-02-02
Remove Empty Statements

bielu:
Hello Guys,
I was reviewing code and notice a lot of empty statements in code.
I decide to clean up that little :)


kjac:
Hi @bielu! Thank you for taking the time to clean up. We'll have a look soon as we can and get back to you 😀

ronaldbarendse:
That's a lot of unnecessary semicolons: I've reviewed the changes and these can all be safely removed 👍 Also a good catch to remove the semicolons in the Razor views, as that would end up in the resulting HTML...

PR #7554, opened 2020-02-02
Add an Active State for Expand (...) in the Top Section Navigation

### LINKED ISSUE #7473: Consider Adding an Active State for Expand (...) in the Top Section Navigation
> **UI/UX Improvement**
> Umbraco v8+

If you were to log into the Umbraco v8 Backoffice you'd be presented with a top navigation, dark blue, consisting of all the sections you have been granted access to. For an administrator with access to them all, some sections may only be accessible through the expand (...) link. 

My suggestion is that if such a section is currently active, that the expand link (...) should have the same active state as other sections.

Step-by-Step | Screenshot
------------ | -------------
Here we have the content section, active by default: | image
And if we have too many sections, they will be hidden behind the expand (...) menu link: | image
And if we were to navigate to one of these sections, we lose the "active state" from before: | image
I know this is a trivial, nit-picky, low-value ask – but I propose we apply an active state to the expand (...) menu item in this scenario: | image

The solution I hope is simple:

1. Apply the `current` class to the expand (...) menu item if a child is currently active.
2. Update the CSS for the expand (...) link so that if the `current` class is applied, each dot matches the active color `#f5c1bc` and the opacity is set to `1.0`.

markadrake:
### Prerequisites

- This PR is intended to fix #7473.

### Description

- The `expand` element now receives the `current` class if the currently active section is found in the overflow.

- If the active section is found in the overflow, the `expand` element shares the same visual appearance.

- Made some of the CSS rulesets more specific, so they didn't cascade to all links underneath the `expand` element.

### Questions / Comments

- To determine if the `current` element should have the class `current` applied, I couldn't find a way to do this with a simple one-liner in the template. I wrote a function `currentSectionInOverflow`. If this is not ideal please give me feedback on how I might update this to a preferred solution.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/application/umbsections.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/application/umbsections.directive.js
@@ -133,6 +133,17 @@ function sectionsDirective($timeout, $window, navigationService, treeService, se
                 }
             };
 
+            scope.currentSectionInOverflow = function () {
+                if (scope.overflowingSections === 0) {
+                    return false;
+                }
+
+                var currentSection = scope.sections.filter(s => s.alias === scope.currentSection);
+
+                return (scope.sections.indexOf(currentSection[0]) >= scope.maxSections);
+
+            };
+
             loadSections();
 
         }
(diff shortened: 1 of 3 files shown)


markadrake:
Screenshots:

image

Selected `Forms`. The `expand` element is highlighted.

image

Showing the dropdown. 

image

kjac:
Hi there @markadrake 🙂 this looks ace, we'll review it as soon as possible.

markadrake:
@kjac I took care of a merge conflict that arouse this week.

kjac:
Good job @markadrake, it's merged!

PR #7555, opened 2020-02-02
Allow Macro's to be cached differently across variants

marcemarc:
### Prerequisites

- [ I] I have added steps to test this contribution in the description below

There is no issue created to link to, I can create one if that helps

### Description

Currently, if you have a Macro used in a content item, that varies by culture, and you set the macro to 'cache' per page and the macros share the same parameters, then the output of the macro is the same for all culture variations ...but this causes an issue if the Macro implementation displays content that varies by culture, eg different child nodes or uses a Dictionary item, ... only the first request to the Macro is cached and the culture's content for that first request is displayed from cache on the other variant pages - and therefore the correct cultural content is not shown!... 

This PR at least attempts to illustrate the problem, It makes a change to the way the macro cache key is generated for Macros to always use the current culture as part of the Key - this ensures each macro  is cached separately per variant...

But is this too naive? and is this the best way to get the current culture to use in the key?

I can't think of a compelling scenario where you'd want to cache the macro the same output across variants, and therefore need any granular control over this eg having a 'Cache By Variant' option in the UI? and I don't think it 'harms' to add the culture name to the cache key for non-varying environments too... but I guess it depends on how expensive the culture lookup is?

To Test this

- Install the Starter Kit
- Make the Homepage document type vary by culture
- make the 'grid' property vary by culture
- add a new language to the site eg English
- add two hostnames to the homepage node one for American culture, one for English culture, using Cultures and Hostnames setting
- add the 'LatestBlogposts' Macro to the grid property of the English version of the page - with 3 news items and the blogs section picked as the parameters.
- Set the 'LatestBlogposts' macro for listing out blog posts to 'cache by page' for 1 minute
- Add a new dictionary item called 'title' in the Translation section of the site, and provide a different value for English and American translations
- Update the 'LatestBlogposts' Macro implementation to include a time stamp, and also write out the dictionary item value:
```
<h1>@DateTime.Now</h1>
<h2>Dictionary Item: @Umbraco.GetDictionaryValue("title")</h2>
```
Request each page separately, the output should cache differently for each language variation, eg different dictionary item and different time stamp - and further requests within the minute for each page should serve the culture specific version of the macro from cache.

NB - without the PR, the first request would cache the output of the first language and display that on the second language variation page.

It works - but just not sure if this is the best way or if           
`var cultureName = content.GetCultureFromDomains(); `
is the best way to get the culture in this context?


emmaburstow:
Hola Marc!

Thanks for the work and the detailed documentation of it. We'll take a look and let you know if we have questions.

Emma

nul800sebastiaan:
Thanks @marcemarc - this is lovely! Yes, as you noticed, getting the culture should be done differently, Umbraco is already responsible for setting the culture properly on requests, so we ca use the one that we set in `System.Threading.Thread.CurrentThread.CurrentUICulture`. I've updated the PR for that.

Other than that, this works well, thanks again Marc! 👍

marcemarc:
thanks @nul800sebastiaan !

PR #7557, opened 2020-02-03
c# Block List models and property value converter with tests created

Shazwazza:
* Strongly typed models for the Block List editor and some base classes/interfaces
* Property Value converter for the Block List editor
* Unit tests
* Adds `element` as a known UDI type
* Shares code between block editor and nested content (i.e. BlockEditorConverter)

Please have a look at the unit tests to see how the model works. You can do this:

```cs
// sudo code to get a `BlockListModel`
var model = GetBlockListModel();

// using the model...

// you can iterate over the data (IPublishedElement) 
foreach(var data in model.Data)
{
    
}
// you can iterate over each item in the layout
foreach(var item in model.Layout)
{
    // get the settings (IPublishedElement) if there is one, it could be null
    var settings = item.Settings;
    // The UDI that refers to the data item in the model.Data collection
    var udi = item.Udi;

    // But you can also directly reference the data block (IPublishedElement)
    var data = item.Data;
    // TODO: Is this confusing? although this is easy to write, does it make more sense to be 
    // explicit and do this instead?
    var data = model.GetData(item.Udi);
}
```

There are some TODOs in the code about naming conventions, let me know what you think about those and the TODO above.


src/Umbraco.Web/PropertyEditors/BlockListConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Umbraco.Web.PropertyEditors
+{
+
+    /// <summary>
+    /// The configuration object for the Block List editor
+    /// </summary>
+    public class BlockListConfiguration
+    {
+        [ConfigurationField("elementTypes", "Element Types", "views/propertyeditors/blocklist/blocklist.elementtypepicker.html", Description = "Select the Element Types to use as models for the items.")]
+        public ElementType[] ElementTypes { get; set; }
```
bjarnef: @Shazwazza I guess this "element type picker" will be a global prevalue editor, which can be re-used in e.g. Nested Content and in third party packages?

warrenbuckley:
Unit Test Failing - but will fix up
>Umbraco.Tests.Composing.TypeLoaderTests.GetDataEditors
>Expected: 38
>But was:  39

src/Umbraco.Core/PropertyEditors/BlockListEditor.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Umbraco.Core.PropertyEditors
+{
+    public class BlockListEditor
```
warrenbuckley: @Shazwazza intended to be empty?

src/Umbraco.Web/Runtime/WebInitialComposer.cs
```diff
@@ -108,6 +108,7 @@ public override void Compose(Composition composition)
             composition.RegisterUnique<HtmlUrlParser>();
             composition.RegisterUnique<HtmlImageSourceParser>();
             composition.RegisterUnique<RichTextEditorPastedImages>();
+            composition.RegisterUnique<PropertyEditors.ValueConverters.BlockEditorConverter>();
```
warrenbuckley: Any reason registered unique here & not put into the `PropertyValueConverterCollection`  on line 177?

src/Umbraco.Web/PropertyEditors/BlockListConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        // TODO: Fill me in
```
warrenbuckley: Is this a leftover TODO @Shazwazza or this for a next phase/iteration of work once synced up with Niels L or is something important missing here?!

warrenbuckley:
## Feedback: Accessing data in the layout
In regards to your question @Shazwazza regarding fetching `Data` whilst iterating over the `layout` items to me, it may be confusing for some (myself included) as the underlying JSON data stored does not exactly match as the layout array does not contain a data property.

Where a more explicit method of `GetData` or similarly named method is more verbose and easier to understand that it has been retrieved from the array of items stored in the Data property of the JSON.

Again it's all personal opinion & preference, to some they would obviously prefer the property.

_This may evolve & change when we get a chance to use Niels UI to build something and then trying to build some real-world use cases on the frontend of the site._

## Review
Can you answer the few small questions that I have put inline in the code please, then I am happy to merge it in :)

src/Umbraco.Web/PropertyEditors/BlockListConfiguration.cs
Shazwazza: @bjarnef The "Block Editor" RFC is really stating that we are building the underlying functionality for implementation of block editors. For the first editor, we are creating the Block List editor (name may change) which is what the RFC is talking about. In the future there may be other editors based on the block editor functionality which is what the RFC also says. So yes, in the future part of the code in this PR will be re-used. As for BlockListConfiguration ... maybe, maybe not, this is just the data type structure for the block list editor which as it stands now is incomplete ... this is just stub code until we create the whole thing.

src/Umbraco.Web/PropertyEditors/BlockListConfiguration.cs
Shazwazza: This is for the next phase of work to be done, we haven't created the property editor yet at thsi stage.

src/Umbraco.Core/PropertyEditors/BlockListEditor.cs
Shazwazza: Doh, will delete, not supposed to be there

src/Umbraco.Web/Runtime/WebInitialComposer.cs
Shazwazza: This is not a property value converter, it is just a service that a couple of them depend on. Make sense?

Shazwazza:
Thanks @warrenbuckley , have pushed some changes and left you some comments. I've changed it to GetData instead since i agree it is more explicit and isn't confusing. Feel free to merge if you're ok with changes.

src/Umbraco.Web/Runtime/WebInitialComposer.cs
warrenbuckley: I think the more verbose namespace made me think it was a ValueConvertor itself just because of where it lives @Shazwazza

warrenbuckley:
All good - merged in :)

PR #7558, opened 2020-02-03
Html encoding document names when rendered in the Relation Types views

elit0451:
_Testing notes in AB task._


warrenbuckley:
Code looks fine & verified it works based on testing notes on internal AB card

PR #7561, opened 2020-02-03
Migrations - Use sp_rename in ReplaceColumn when available

stevemegson:

In SQL Server, use sp_rename in ReplaceColumn rather than adding a new column and copying the data.

### DIFF
--- a/src/Umbraco.Core/Migrations/MigrationBase_Extra.cs
+++ b/src/Umbraco.Core/Migrations/MigrationBase_Extra.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using Umbraco.Core.Persistence;
 using Umbraco.Core.Persistence.DatabaseModelDefinitions;
 using Umbraco.Core.Persistence.SqlSyntax;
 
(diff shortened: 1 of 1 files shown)


src/Umbraco.Core/Migrations/MigrationBase_Extra.cs
```diff
@@ -84,10 +85,18 @@ protected void AlterColumn<T>(string tableName, string columnName)
...
+            if (DatabaseType.IsSqlCe())
+            {
+                AddColumn<T>(tableName, newName, out var sqls);
+                Execute.Sql($"UPDATE {SqlSyntax.GetQuotedTableName(tableName)} SET {SqlSyntax.GetQuotedColumnName(newName)}={SqlSyntax.GetQuotedColumnName(currentName)}").Do();
+                foreach (var sql in sqls) Execute.Sql(sql).Do();
+                Delete.Column(currentName).FromTable(tableName).Do();
+            }
+            else
+            {
+                Execute.Sql(SqlSyntax.FormatColumnRename(tableName, currentName, newName)).Do();
```
Shazwazza: Wow, learn something new everyday. I didn't realize we had a FormatColumnRename on sql syntax and already had sp_rename on there. It's not actually used anywhere but nice to know we support the Migration syntax `Rename.Column(....).To(....)`

PR #7562, opened 2020-02-03
Migrations - Create temporary index on umbracoPropertyData

stevemegson:

Create a temporary index on umbracoPropertyData to speed up remaining migrations.

### DIFF
--- a/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/VariantsMigration.cs
+++ b/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/VariantsMigration.cs
@@ -19,6 +19,7 @@ namespace Umbraco.Core.Migrations.Upgrade.V_8_0_0
         public override void Migrate()
         {
             MigratePropertyData();
+            CreatePropertyDataIndexes();
             MigrateContentAndPropertyTypes();
             MigrateContent();
             MigrateVersions();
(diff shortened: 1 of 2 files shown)


src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/VariantsMigration.cs
```diff
@@ -90,6 +91,15 @@ private void MigratePropertyData()
             Rename.Table(PreTables.PropertyData).To(Constants.DatabaseSchema.Tables.PropertyData).Do();
         }
 
+        private void CreatePropertyDataIndexes()
+        {
+            // Creates a temporary index on umbracoPropertyData to speed up other migrations which update property values.
+            // It will be removed in CreateKeysAndIndexes before the normal indexes for the table are created
+            var tableDefinition = Persistence.DatabaseModelDefinitions.DefinitionFactory.GetTableDefinition(typeof(PropertyDataDto), SqlSyntax);
+            Execute.Sql(SqlSyntax.FormatPrimaryKey(tableDefinition)).Do();
+            Execute.Sql($"CREATE UNIQUE NONCLUSTERED INDEX IX_umbracoPropertyData_Temp ON {PreTables.PropertyData} (versionId,propertyTypeId,languageId,segment)").Do();
```
Shazwazza: Makes sense. I'm fairly positive that this is the structure that we actually want the index to exist on this table anyways instead of having 4x individual indexes which i think not actually helping us at all and is probably causing slower performance than faster perf. Something we should look into soon. Have created a separate issue for this https://github.com/umbraco/Umbraco-CMS/issues/7570

Shazwazza: I'll need to fix this up - at this stage this will throw an exception because we've already changed the table name in `MigratePropertyData`. I'm also unsure why the format PK line is here?

Shazwazza:
@stevemegson I've pushed an update to this here a426a15b5a0cec83f5c630b06a87330eb7e85d8b but still unsure about the need for the FormatPrimaryKey call if you can let me know what you think about that. Thanks!

stevemegson:
@Shazwazza Damn, I'd gone through the original PR getting rid of some hard-coded table names, and managed to use the wrong constant here. Your update is obviously the better solution anyway.

Adding the primary key will be to speed up updates to single rows, since they'll generate `UPDATE umbracoPropertyData SET ... WHERE id=x`.

Shazwazza:
@stevemegson aha right and that's because we first clear all keys/indexes. ok makes sense thanks! i'll keep working through your other PRs

PR #7563, opened 2020-02-03
Migrations - SQLCE does not support UPDATE...FROM

stevemegson:

Several places use a SELECT followed by many UPDATES in a foreach loop because SQLCE doesn't support UPDATE...FROM. It's worth the effort to include the UPDATE...FROM version for SQL Server.

### DIFF
--- a/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/VariantsMigration.cs
+++ b/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/VariantsMigration.cs
@@ -74,10 +74,18 @@ HAVING COUNT(v2.id) <> 1").Any())
             {
                 Alter.Table(PreTables.PropertyData).AddColumn("versionId2").AsInt32().Nullable().Do();
 
-                // SQLCE does not support UPDATE...FROM
-                var temp = Database.Fetch<dynamic>($"SELECT id, versionId FROM {PreTables.ContentVersion}");
-                foreach (var t in temp)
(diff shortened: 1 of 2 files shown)


src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/VariantsMigration.cs
```diff
@@ -153,22 +161,40 @@ private void MigrateVersions()
...
-            foreach (var t in temp1)
-                Database.Execute($@"UPDATE {PreTables.ContentVersion} SET text=@text, {SqlSyntax.GetQuotedColumnName("current")}=@current, userId=@userId WHERE versionId=@versionId",
-                    new { text = t.text, current = t.newest, userId=t.documentUser, versionId=t.versionId });
+            if (Database.DatabaseType.IsSqlCe())
+            {
+                // SQLCE does not support UPDATE...FROM
+                var documents = Database.Fetch<dynamic>($"SELECT versionId, text, published, newest, documentUser FROM {PreTables.Document}");
+                foreach (var t in documents)
+                    Database.Execute($@"UPDATE {PreTables.ContentVersion} SET text=@text, {SqlSyntax.GetQuotedColumnName("current")}=@current, userId=@userId WHERE versionId=@versionId",
+                        new { text = t.text, current = t.newest && !t.published, userId = t.documentUser, versionId = t.versionId });
```
stevemegson: The change from `current = t.newest` to current = `t.newest && !t.published` comes from #7569

TimZander:
This is not in any releases I have seen. I would like to use it to upgrade my database as it is currently taking hours to do so, but without it in a released package I am nervous

PR #7569, opened 2020-02-03
Migrations - Add extra "published=newest" versions with INSERT...SELECT

stevemegson:

When creating the new versions for content with published=newest, replace foreach loops with three INSERTs to umbracoContentVersion, umbracoDocumentVersion and umbracoPropertyData.

### DIFF
--- a/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/VariantsMigration.cs
+++ b/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/VariantsMigration.cs
@@ -179,31 +179,33 @@ SELECT cver.id, doc.templateId, doc.published
 FROM {SqlSyntax.GetQuotedTableName(PreTables.ContentVersion)} cver
 JOIN {SqlSyntax.GetQuotedTableName(PreTables.Document)} doc ON doc.nodeId=cver.nodeId AND doc.versionId=cver.versionId");
 
+
             // need to add extra rows for where published=newest
             // 'cos INSERT above has inserted the 'published' document version
             // and v8 always has a 'edited' document version too
-            var temp3 = Database.Fetch<dynamic>($@"SELECT doc.nodeId, doc.updateDate, doc.documentUser, doc.text, doc.templateId, cver.id versionId
+            Database.Execute($@"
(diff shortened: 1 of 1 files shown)


Shazwazza:
Code looks good, just testing it all out now, it's certainly the way forward! Couple questions though:

the old code:

```cs
Database.Execute($"UPDATE {SqlSyntax.GetQuotedTableName(PreTables.ContentVersion)} SET {SqlSyntax.GetQuotedColumnName("current")}=0 WHERE nodeId=@0 AND id<>@1", (int) t.nodeId, id);
```

Is this now unnecessary because of the change to setting current above like `current = t.newest && !t.published` ?

src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/VariantsMigration.cs
```diff
@@ -179,31 +179,33 @@ private void MigrateVersions()
...
+
+            Database.Execute($@"
+INSERT INTO {SqlSyntax.GetQuotedTableName(Constants.DatabaseSchema.Tables.DocumentVersion)} (id, templateId, published)
+SELECT cverNew.id, doc.templateId, 0
+FROM {SqlSyntax.GetQuotedTableName(PreTables.Document)} doc
+JOIN {SqlSyntax.GetQuotedTableName(PreTables.ContentVersion)} cverNew ON doc.nodeId = cverNew.nodeId
+WHERE doc.newest=1 AND doc.published=1 AND cverNew.{SqlSyntax.GetQuotedColumnName("current")} = 1");
+
+            Database.Execute($@"
+INSERT INTO {SqlSyntax.GetQuotedTableName(PreTables.PropertyData)} (propertytypeid,languageId,segment,textValue,varcharValue,decimalValue,intValue,dateValue,versionId)
```
Shazwazza: This table has already been renamed, i'll fix up

stevemegson:
> Is this now unnecessary because of the change to setting current above like `current = t.newest && !t.published` ?

Yes, the only row that update should be changing is the published version, and the change to `current = t.newest && !t.published` means that there'll be nothing to change.

PR #7573, opened 2020-02-04
ImageSharp implementation

elit0451:
Using **ImageSharp** to manipulate an image using information defined in a query string.

### DIFF
--- a/src/Umbraco.Abstractions/Models/ImageFileModel.cs
+++ b/src/Umbraco.Abstractions/Models/ImageFileModel.cs
@@ -1,8 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-
-namespace Umbraco.Core.Models
+﻿namespace Umbraco.Core.Models
 {
     public class ImageFileModel
     {
--- a/src/Umbraco.Abstractions/Models/ImageFileModel.cs
+++ b/src/Umbraco.Abstractions/Models/ImageFileModel.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Umbraco.Core.Models
+{
+    public class ImageFileModel
+    {
+        public byte[] Data { get; }
+        public string MimeType { get; }
+
+        public ImageFileModel(byte[] imageData, string mimeType)
+        {
+            Data = imageData;
+            MimeType = mimeType;
+        }
+    }
+}
--- a/src/Umbraco.Abstractions/Services/IImageSharpImageService.cs
+++ b/src/Umbraco.Abstractions/Services/IImageSharpImageService.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Umbraco.Core.Models;
+
+namespace Umbraco.Core.Services
+{
+    public interface IImageSharpImageService
+    {
+        ImageFileModel GetImage(string imageOriginalPath, int width, int height);
+    }
+}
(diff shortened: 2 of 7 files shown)


src/Umbraco.Web/Editors/ImageSharpImageController.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Web.Mvc;
+using Umbraco.Core.Services;
+using Umbraco.Web.Mvc;
+
+namespace Umbraco.Web.Editors
+{
+    public class ImageSharpImageController : PluginController
+    {
+        private readonly IImageSharpImageService _imageSharpImageSertvice;
```
stevetemple: Typo `Sertvice`!

elit0451: Nice spot @stevetemple 👏 will update

src/Umbraco.Abstractions/Services/IImageSharpImageService.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Umbraco.Core.Models;
+
+namespace Umbraco.Core.Services
+{
+    public interface IImageSharpImageService
```
JimBobSquarePants: `IImageSharpService` Why would you create an interface specific to a specific library then abstract out the model? 
This makes no sense.

src/Umbraco.Abstractions/Models/ImageFileModel.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Umbraco.Core.Models
+{
+    public class ImageFileModel
+    {
+        public byte[] Data { get; }
```
JimBobSquarePants: This is an incredibly allocation heavy pattern.

src/Umbraco.Infrastructure/Media/ImageSharpImageService.cs
```diff
@@ -0,0 +1,39 @@
...
+using SixLabors.ImageSharp.Processing;
+using Umbraco.Core.Models;
+using Umbraco.Core.Services;
+
+namespace Umbraco.Infrastructure.Media
+{
+    public class ImageSharpImageService : IImageSharpImageService
+    {
+
+        public ImageFileModel GetImage(string imageOriginalPath, int? width, int? height)
```
JimBobSquarePants: How do you know the images are always going to be stored in a file system?

src/Umbraco.Infrastructure/Media/ImageSharpImageService.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            if (string.IsNullOrEmpty(imageOriginalPath)) return null;
+
+            using (var image = Image.Load(imageOriginalPath, out imageFormat))
+            {
+                image.Mutate(x =>
+                {
+                    if (width != null)
+                        if (height != null)
+                            x.Resize((int) width, (int) height);
```
JimBobSquarePants: You want to avoid null checks in your interface design. Each implementation now requires null checks.

src/Umbraco.Web/Editors/ImageSharpImageController.cs
```diff
@@ -0,0 +1,21 @@
...
+        public ImageSharpImageController(IImageSharpImageService imageSharpImageSertvice)
+        {
+            _imageSharpImageSertvice = imageSharpImageSertvice;
+        }
+        public ActionResult ManipulateImage(string imagePath, int? width, int? height)
```
JimBobSquarePants: Someone could now repeatedly call this method and DDOS the site.

benjaminc:
I think any implementation of this really should use the ImageSharp middleware instead of adding our own.  Is this just a stop-gap until we get to .NET Core, or is the intent here to provide long-term ImageSharp access from a .NET Framework project?

src/Umbraco.Abstractions/Models/ImageFileModel.cs
elit0451: Thank you @JimBobSquarePants! The implementation will be refactored

PR #7577, opened 2020-02-04
Refactor: PublishedSnapshotService 3 near identical functions

stevetemple:
### Description

Whilst looking at this code for migration to asp-netcore I've noticed quite a few bits that could be tidied up. This is the first of them.

There are 3 virtually identical methods in this class that are running the same SQL with different parameters. I've refactored these so that they call private methods passing in the differences with different parameters. This will reduce the likelihood that these will accidentally diverge over time introducing bugs. 

For future refactoring this function is a candidate to move to it's own class


emmaburstow:
Hiya Steve!

Thanks for the PR. I'll label this too.

Em

stevetemple:
FYI This isn't really dependent on project/net-core it's just a general tidy up

emmaburstow:
I did wonder whether it was a tidy up in the service of net-core or a tidy up that was more general. Ok - label removed 👍

PR #7579, opened 2020-02-05
Disable clicking on disabled checkboxes/radios

nathanwoulfe:
Adding pointer-events:none so that the disabled checkbox/radio is not clickable. Without this rule, the disabled item is opaque but still clickable, so not really disabled.


emmaburstow:
ooh good spot. Thanks for this Nathan! 

Em

nathanwoulfe:
👍👍

emmaburstow:
heyyyy 

I wanted to merge this but we have a problem. 

image

Thoughts?

Emma

nathanwoulfe:
Ah yes. I might have killed my fork. Don't ask, it was not a good day. 

It's a one-liner, I'll redo.

nathanwoulfe:
https://github.com/umbraco/Umbraco-CMS/pull/7689

PR #7580, opened 2020-02-05
NetCore: Replaced the Umbraco.Examine.Lucene.csproj with the new format

bergmania:
- Updated the format of the Umbraco.Examine.Lucene.csproj
- Updated and aligned versions of NuGet packages 
- Removed unnecessary InternalVisibleTo


bergmania:
Yeah, the test web config is auto generated and for some reason not ignored in our gitignore. I see it committed sometimes if we commit while running tests locally.

PR #4927, opened 2019-03-09
V8: Allow reordering of allowed child types

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/4923

### Description

This PR adds the ability to explicitly sort the allowed child content types as described in #4923.

It works for both content and media.

For content/media types allowed at root, the current sort order (by name) is preserved.

#### Demo: Content

sort-allowed-children

#### Demo: Media

sort-allowed-children-media

PR #7582, opened 2020-02-05
Adjust child selector sorting

bjarnef:
### Prerequisites

### Description
This PR https://github.com/umbraco/Umbraco-CMS/pull/4927 introduced sorting on the child elements in `umb-child-selector` component. However with these changed add also added the `move` cursor on the add button.

The changes in this PR ensure only the child elements that can be sorted get the `move` cursor, while the add-button still has the `pointer` cursor.
It also add documentation of the previous added `onSort` function.

Furthermore it has a few changed to update buttons using `button` element and add `aria-hidden="true"` to icons.


poornimanayar:
Hi @bjarnef ,

Many thanks for the PR. We shall review it and get back to you. Watch this space.

Poornima

PR #7583, opened 2020-02-05
Slider format function should return a number

### LINKED ISSUE #7564: Slider with negative value weird frontend behaviour
When using the slider data type with negative min value the frontend behaviour goes weird. However the input is actually is saved.

image

[Image from Gyazo](https://gyazo.com/f8b94c03280b311e50be8c77a62b7476)

## Umbraco version

I am seeing this issue on Umbraco version: 8.5.3

Reproduction
------------

### Specifics
Using Brave Version 1.2.43 Chromium: 79.0.3945.130 (Official Build) (64-bit)
Same on Edge

No errors in console

### Steps to reproduce

1. Create a Slider data type
2. Set min value to a negative value
3. Create a property with the data type
4. In content try to edit the value

### Expected result
The slider value to stay put

### Actual result
The slider does not stay put

stevemegson:
### Prerequisites

If there's an existing issue for this PR then this fixes #7564

### Description
In noUiSlider configuration, the format `from` function should return a number. Returning a string works as long as the minimum is non-negative, but causes strange behaviour when minimum is negative.

Testing:
- Create a slider property with the minimum set to any negative number.
- The slider should behave normally, not jump to the maximum and minimum values as seen in #7564.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/slider/slider.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/slider/slider.controller.js
@@ -56,7 +56,7 @@
                     return value.toFixed(stepDecimalPlaces);
                 },
                 from: function (value) {
-                    return value;
+                    return Number(value);
                 }
             },
             "range": {


poornimanayar:
Hi @stevemegson ,

Many thanks for the work as always. We shall review the PR and get back to you.

Regards,
Poornima

kjac:
Thanks a bunch, @stevemegson

PR #7584, opened 2020-02-05
Use button element in user detail and user group views

bjarnef:
### Prerequisites

### Description
This PR replaces add buttons in user detail and user group views to use `button` element instead of anchor element and improves accessibility.


poornimanayar:
Hi again @bjarnef ,

Many thanks for the work. We shall review it and get back to you :-)

Regards,
Poornima

PR #6955, opened 2019-10-29
Package repo: Accessibility improvements

BatJan:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
I this PR I have made the following changes
- Changed `<a>` without a `href` value to `<button>`
- Added a visually hidden `<label>` for the search field
- Added `aria-hidden="true"` to `<i>` elements
- Fixed styling accordingly with the changed elements

The repo looks the same as before the changes so makes no sense to add before/after images 😃

PR #7586, opened 2020-02-05
Update button in package overview not to use flexbox

bjarnef:
### Prerequisites

Tested in Chrome Version 79.0.3945.130

### Description
This PR update the button element in package overview to not use flexbox, but instead use flexbox for a inner wrapping div. The issues where the "umb-package-icon" overflow the container was affected after changes in this PR https://github.com/umbraco/Umbraco-CMS/pull/6955 where the anchor elements was changed to `button` elements.
It also ensure the border color change on hover works as previous.

In some browers it is not supported to use `button` element as flexbox container or doesn't allow to change its `display` value.
https://stackoverflow.com/a/35466231/1693918

> `<button>` is not implementable (by browsers) in pure CSS, so they are a bit of a black box, from the perspective of CSS. This means that they don't necessarily react in the same way that e.g. a `<div>` would.

https://github.com/philipwalton/flexbugs#flexbug-9

**Before**
image

image

**After**

image

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-packages.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-packages.less
@@ -59,15 +59,12 @@
 
 .umb-package-link {
     display: block;
-    flex-wrap: wrap;
-    flex-direction: column;
-    justify-content: center;
     position: relative;
     box-sizing: border-box;
     height: 100%;
     width: 100%;
     border-radius: 3px;
-    border: 0 none;
+    border: 1px solid transparent;
     text-decoration: none !important;
     transition: border-color 100ms ease;
     background-color: @white;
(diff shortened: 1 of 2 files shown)


poornimanayar:
Hi again @bjarnef ,

Thank you for the PR! Watch this space while we review the work :-)

Regards,
Poornima

bjarnef:
@poornimanayar any update on this one?

bjarnef:
@nul800sebastiaan do need more details on this? It would be great with a follow-up when the PR team add a comment, but nothing further has happened within 1, 2 or 3 months :)
I feel many of these get a comments, but is then a bit forgotten as a lot of other PR's are coming in. Would love to get some of these closed to limit the number of open PR's 😁😎

nul800sebastiaan:
Thanks @bjarnef - sometimes we can't get to everything in good time unfortunately, and you have so many open...

Only 200 open PRs to go in total, one day! 🙈 

In any case, this one is great, works for me! 👍

PR #7588, opened 2020-02-05
Bugfix: Allow users to explicitly scan assmblies with ServiceContaine…

### LINKED ISSUE #7502: LightInject Registering by convention is not possible as you have replaced the default AssemblyScanner.
I can't register dependencies by convention.

container.RegisterAssembly() is completely unusable as the AssemblyScanner has been replaced with a NOP stub.

See https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Composing/LightInject/LightInjectContainer.cs

Line 61
Line 254

## Umbraco version

I am seeing this issue on Umbraco version: 8.5.2

Reproduction
------------
### Steps to reproduce

Spin up umbraco 8.5.2 site

Add the following to a c# file

```csharp
    public class Composer : IUserComposer
    {
        {
            var container = composition.Concrete as ServiceContainer;
            Debug.Assert(container != null, nameof(container) + " != null");
            container.RegisterAssembly(typeof(IFoo).Assembly);

            // The following throws
            container.GetInstance<IFoo>();
        }
    }

    public interface IFoo
    {
        string Ping();
    }

    public class Foo : IFoo
    {
        public string Ping()
        {
            return "Pong";
        }
    }
```

### Expected result

I expect to be able to resolve an instance implementing IFoo

### Actual result

It doesn't work.

However if I blast over the NOP scanner with the following it works as one would expect.
```csharp
        {
            var container = composition.Concrete as ServiceContainer;
            Debug.Assert(container != null, nameof(container) + " != null");

            // Replace NOP scanner
            container.AssemblyScanner = new ServiceContainer().AssemblyScanner;

            container.RegisterAssembly(typeof(IFoo).Assembly);
            // The following does not throw
            container.GetInstance<IFoo>();
        }
```

rustybox:
…r.RegisterAssembly

Closes #7502


rustybox:
I much prefer the follow as it allows removal of the IAssemblyScanner implementation **### **however**,
whilst this works really nicely when running the site, it at present breaks a bunch of the tests, for example failing to resolve PropertyEditorCollection in Umbraco.Core.Composing.Current

```csharp
            bool PreventScanForUmbracoCompositionRootsPredicate(Type type, string s)
            {
                var fullName = type?.FullName;

                // ReSharper disable once ConvertIfStatementToReturnStatement
                if (string.IsNullOrEmpty(fullName))
                    return false;

                // The fallback factory method only applies to types in Umbraco namespaces
                return fullName.StartsWith("Umbraco", StringComparison.OrdinalIgnoreCase);
            }

            // Explicitly registering a fallback factory method prevents ServiceContainer.GetEmitMethod
            // from scanning the Umbraco assemblies for composition roots.
            container.RegisterFallback(
                predicate: PreventScanForUmbracoCompositionRootsPredicate,
                factory: request => null
            );
```

I have some errands to run, but will attempt to fix tests later and update PR.

rustybox:
OK this is much nicer now.

nul800sebastiaan:
Please note: as I mentioned in issue #7502 - I will need to ask at Umbraco HQ is this is a change that we actually want. So we might not be able to accept this PR 🙈
I really don't want to waste your time so please hold off on doing more work on this until we know this is where we want to go!

rustybox:
No worries, nothing left to do now.

WRT if HQ want to allow users to use the container as they see fit, It's certainly something the community are looking for:

https://our.umbraco.com/forum/umbraco-8/98728-dependency-injection-umbraco-8
https://our.umbraco.com/forum/using-umbraco-and-getting-started/99810-lightinject

src/Umbraco.Tests/Testing/UmbracoTestBase.cs
```diff
@@ -418,15 +421,25 @@ protected virtual void Reset()
...
-                    scope.Dispose();
+                    Scope scope;
+                    while ((scope = scopeProvider.AmbientScope) != null)
+                    {
+                        scope.Reset();
+                        scope.Dispose();
+                    }
                 }
             }
+            catch (ResolveUnregisteredDependencyException)
```
Shazwazza: Is this something we'd have to worry about during regular application shutdown?

src/Umbraco.Core/Exceptions/ResolveUnregisteredDependencyException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Umbraco.Core.Exceptions
+{
+    class ResolveUnregisteredDependencyException : ApplicationException
```
Shazwazza: If we are going to have a custom exception, we'll need to follow the best practices, see https://github.com/umbraco/Umbraco-CMS/issues/5947 and https://github.com/umbraco/Umbraco-CMS/pull/6599

That said, we already have LightInjectException, perhaps we can re-use that?

src/Umbraco.Core/Composing/LightInject/LightInjectContainer.cs
```diff
@@ -56,9 +57,24 @@ protected static ServiceContainer CreateServiceContainer()
...
+            // ensure that we do *not* scan our assemblies
+            // but do not prevent users from doing so if they so desire.
+            bool PreventScanForUmbracoCompositionRootsPredicate(Type type, string s)
+            {
+                // Tests expect the other registered fallback to run, see RegisterAuto below.
+                if (!type.IsInterface)
+                    return false;
+
+                // The fallback factory method only applies to types in Umbraco namespaces
+                return type.FullName?.StartsWith("Umbraco", StringComparison.OrdinalIgnoreCase) ?? false;
```
Shazwazza: Is this too restrictive? Some user's may have assembly names that start with Umbraco (i.e. `UmbracoIdentity`). Maybe it should be "Umbraco." ?

src/Umbraco.Core/Composing/LightInject/LightInjectContainer.cs
```diff
@@ -56,9 +57,24 @@ protected static ServiceContainer CreateServiceContainer()
...
+            }
+
+            // Explicitly registering a fallback factory method prevents ServiceContainer.GetEmitMethod
+            // from scanning the Umbraco assemblies for composition roots.
+            container.RegisterFallback(
+                PreventScanForUmbracoCompositionRootsPredicate,
+                request => throw new ResolveUnregisteredDependencyException(request.ServiceType)
```
Shazwazza: Do we want to throw an exception here? else can just be null and we don't have to worry about this accidentally throwing for weird reasons.

src/Umbraco.Core/Exceptions/ResolveUnregisteredDependencyException.cs
rustybox: Reusing sounds good, will update PR

src/Umbraco.Core/Composing/LightInject/LightInjectContainer.cs
rustybox: Agreed 100%

src/Umbraco.Core/Composing/LightInject/LightInjectContainer.cs
rustybox: I can argue it both ways, null from resolve is current behaviour so it's less of an impact which is nice, but i'm a big fan of fail hard/fast. When something goes wrong i'd much rather be thinking "oh there's something wrong with the ioc container/registrations let's go check that out" than "eh why have I got a random null reference exception".

I went with null at first but opted for an exception as I believe it is better, but happy to switch it to returning null.

src/Umbraco.Tests/Testing/UmbracoTestBase.cs
rustybox: I don't believe so no. I triggered a shutdown with a web.config change whilst debugger attached and I can't get the fallback to throw. Logs look good also.

Essentially it just occurs in the tests now because previously line 427 in UmbracoTestBase would get a null and carry on whilst skipping scope.Reset() && Dispose in tests that handn't wired up a mock scope provider

```csharp
var instance = Factory?.TryGetInstance<IScopeProvider>();
```

A good demo of the issue occuring is in

Umbraco.Tests.Models.ContentTests #Can_Change_Property_Value 

TearDown in UmbracoTestBase will trigger the exception with "Unable to resolve type: Umbraco.Core.Scoping.IScopeProvider"

src/Umbraco.Core/Composing/LightInject/LightInjectContainer.cs
rustybox: On second thoughts, LightInject.TryGetInstance is explicitly called which is documented as returning null on failure.

So I think returning null is better otherwise things will become very confusing.

rustybox:
```
Umbraco.Tests.Models.PropertyTypeTests.Can_Deep_Clone
Expected: <11>
But was:  <11>
```

OK then...

Shazwazza:
@rustybox sorry i think that error is from another merged in PR, we'll get that fixed and you can re-merge and hopefully all tests pass. I'm happy to go ahead with your PR now that I understand what is happening, glad you got all those tests in there so if anything changes in LI we will know if our work around fails.

rustybox:
@Shazwazza is there a concern arround the following 

> With some additional investigation, it turns out that assembly scan is called primarily for these assemblies: mscorlib, System.Web.Mvc, System.Web.Http (and friends), all compiled razor view assemblies and it occurs on startup and pretty much anytime a controller is created which is a lot. When controllers are created then it starts trying to scan Umbraco.Web too.

Because this only prevents scan for "Umbraco." namespaces, if that's a concern https://github.com/umbraco/Umbraco-CMS/pull/7588/commits/abfacf78d7efefda4c2d9ce6538f8ce060d4c2e2 or your own PR may be better choice

Shazwazza:
@rustybox aha yes forgot about that one thanks. OK i'll have a look at your orig commit + your tests and see how that goes.

rustybox:
Sadly tests are lacking in original commit, but in theory the tests from the later commits should still "just work"

Shazwazza:
yep no worries :)

Shazwazza:
I know this is still outstanding 😢 It's been starred in my inbox for ages to try to get around to but my time keeps slipping away. I've pinged some folks on the team to see if someone can assist with having a look.

rustybox:
I no longer think this deserves any attention, resolved by the work for #8653

Shazwazza:
@rustybox Sounds good to me, so far nobody else is reporting on this one. I'll close for now, thanks for all your effort!

PR #7592, opened 2020-02-06
Installer adjustments

bjarnef:
### Prerequisites

### Description
A few adjustments to installer where some properties in `installerService` get assigned a value `null` instead of `undefined`. Furhermore the anchor element on starterkit view is changed to a `button` element.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/installer/installer.service.js
+++ b/src/Umbraco.Web.UI.Client/src/installer/installer.service.js
@@ -322,8 +322,8 @@ angular.module("umbraco.install").factory('installerService', function ($rootSco
 			service.status.feedback = null;
 			service.status.fact = null;
 
-			if(factTimer){
-				clearInterval(factTimer);
+			if (factTimer) {
+			    clearInterval(factTimer);
 			}
 		},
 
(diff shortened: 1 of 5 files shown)


poornimanayar:
Hi @bjarnef ,

Thanks again for the PR. We shall review it and get back to you :-)

Poornima

kjac:
@bjarnef thanks 👍 could you add some steps to test this, please?

bjarnef:
@kjac the best way to test this is the set `umbracoDbDSN` and `Umbraco.Core.ConfigurationStatus` to empty string in `web.config`, which should trigger the installer.

Then go through the installer steps and ensure Umbraco is installed correct, eventually the step to install a starterkit, where `<a>` element is changed to a `<button>` element.

I saw somewhere @Shazwazza mentioned variables shouldn't be assigned to a `undefined` value since you actually assign a value, where it is better to assign `null`.

kjac:
@bjarnef aah I get it, it's a bit housekeeping + changing that link to a button - thanks, much appreciated 👍

bjarnef:
@kjac did you had a further look at this?

kjac:
@bjarnef I did... I started a review and tagged you in it. Does it not show for you?

nul800sebastiaan:
@kjac I think you only "self-requested a review" but never saved your comment. Let us know if you have anything to add. From my point of view this looks good and the installer works as expected.

Thanks @bjarnef! 👍 

Note: some more stuff for @nathanwoulfe & friends to refactor in due time for #7663

PR #7595, opened 2020-02-06
Missing onClose for tour

bjarnef:
### Prerequisites

### Description
In a clean Umbraco install with no content nodes and running the tour "Create content", it stops after step 2 because it can't find a homepage node and you get the tour step error "Oh, we got lost!". The button "End tour" close the overlay, while the close button does nothing, because there is no `on-close` implemented for this tour step.

chrome_2020-02-06_13-22-21

chrome_2020-02-06_13-23-24

Not sure if it makes sense this tour look for a homepage node? I guess some might take this tour to know how to get started creating content?

It would be great if it also was possible to close the tour by using `esc` shortcut like with other overlays in backoffice.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/application/umbtour/umbtourstep.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/application/umbtour/umbtourstep.directive.js
@@ -18,9 +18,10 @@
 
         function link(scope, element, attrs, ctrl) {
 
-            scope.close = function() {
-                if(scope.onClose) {
-                    scope.onClose();
+
+            scope.close = function () {
+                if (scope.onClose) {
+                   scope.onClose();
                 }
             }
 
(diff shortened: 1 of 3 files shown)


bjarnef:
I have now added the option to close the tour overlay using `esc` key. However for some reason it doesn't work in this step, where an element is highlighted.

image

poornimanayar:
Hi @bjarnef ,

Thank you for the PR. Watch this space while we review the changes :-)

Poornima

bjarnef:
@kjac actually it was also happening for some other tour steps. Not sure if it is related to if `event` or `eventElement` is defined. Maybe this block key events somewhere, which might be fine - but should probably exclude `esc` then.

https://github.com/umbraco/Umbraco-CMS/blob/163c1f7028eb419eda33e8a0c0548f66fd1ee8e2/src/Umbraco.Web.UI/config/BackOfficeTours/getting-started.json#L261-L262

PR #7598, opened 2020-02-06
Update NestedContentPropertyEditorsMigration.cs

jmayntzhusen:
Exact same issue here as https://github.com/umbraco/Umbraco-CMS/pull/6368 🙂


nul800sebastiaan:
I did something weird and could not push to your branch, so it's being replace by this https://github.com/umbraco/Umbraco-CMS/pull/7609

PR #7599, opened 2020-02-06
UI: corrected button style to match new style

nielslyngsoe:
Corrected the button style of the dropdown version of the create button on the Media section.

image

---
_This item has been added to our backlog [AB#8130](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8130)_


abjerner:
I like it 👍 

I know the old button had this as well, but can we do something about the margin to the left of the button? Ideally the button should align with the left side of the dropzone (IMO).

nathanwoulfe:
That margin is because there are hidden button-groups preceding this one in the DOM, and sibling button-groups have a margin-left to create gutters between the groups. 

Could change that to margin-right, with margin-right:0 on last-child...

nielslyngsoe:
@abjerner It's super annoying with the way margins have been approached in this project. It would require a larger general change of almost everything to make this work appropriate. But thanks for the awareness, it would be good to get that corrected, as it annoys the eye, but its not our first priority at the moment. But do feel free to provide a take it. :-)

abjerner:
@nielslyngsoe the view has three `.btn-group` elements, but as far as I can see, their respective `ng-show` attributes ensure that only one of them are shown at any giving time. So wouldn't putting a `<div>` around each of them solve this? It at least seems so, but not sure whether I've missed something.

warrenbuckley:
CSS/HTML class name changes look fine - will wait for the CI build to be happy & will merge in 👍

warrenbuckley (approved):

PR #7600, opened 2020-02-06
Fix jumping section name on focus

bjarnef:
### Prerequisites

### Description
When using keyboard to navigate through the sections, these are jumping a bit, because the section name get a border on focus, but no by default. So on focus it is moved a few pixels.
Adding a transparent border ensure the section name not jump on focus.

**Before**

**After**


abjerner:
...aaand it's merged - thanks a lot 🥇 

wvrn7MMemcOB2-robert-downey-jr

PR #7602, opened 2020-02-07
Ensures that we track relations for values on all variant property values

Shazwazza:
With item tracking in 8.6 we weren't tracking references for values in variants/segments. We did this originally because we are planning on tracking references directly to variants/segments but we are not there yet since that requires db changes. For now, we will track item refrences based on any valid variant/invariant property values.

This PR adds support for that and is based on the TODO in this comment https://github.com/umbraco/Umbraco-CMS/blob/v8/feature/media-tracking/src/Umbraco.Core/PropertyEditors/DataValueReferenceFactoryCollection.cs#L22

Unit tests have been added too.

### DIFF
--- a/src/Umbraco.Tests/PropertyEditors/DataValueReferenceFactoryCollectionTests.cs
+++ b/src/Umbraco.Tests/PropertyEditors/DataValueReferenceFactoryCollectionTests.cs
@@ -1,0 +1,223 @@
+﻿using Moq;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Umbraco.Core;
+using Umbraco.Core.Logging;
+using Umbraco.Core.Models;
+using Umbraco.Core.Models.Editors;
+using Umbraco.Core.PropertyEditors;
+using Umbraco.Web.PropertyEditors;
+using static Umbraco.Core.Models.Property;
+
+namespace Umbraco.Tests.PropertyEditors
+{
+    [TestFixture]
+    public class DataValueReferenceFactoryCollectionTests
+    {
+        [Test]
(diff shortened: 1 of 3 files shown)


src/Umbraco.Tests/PropertyEditors/DataValueReferenceFactoryCollectionTests.cs
```diff
@@ -0,0 +1,223 @@
...
+{
+    [TestFixture]
+    public class DataValueReferenceFactoryCollectionTests
+    {
+        [Test]
+        public void GetAllReferences_All_Variants_With_IDataValueReferenceFactory()
+        {
+            var collection = new DataValueReferenceFactoryCollection(new TestDataValueReferenceFactory().Yield());
+
+            // label does not implement IDataValueReference
```
warrenbuckley: @Shazwazza like your comment says a label is just a string and does not store a reference to anything, but this unit test expects two results/items to be references?!

warrenbuckley:
Code reads fine but just a question about one of the unit tests is a bit confusing as its a Label property editor which itself does not track references.

src/Umbraco.Tests/PropertyEditors/DataValueReferenceFactoryCollectionTests.cs
Shazwazza: Yes, the point of thi test is to show references resolved from `IDataValueReferenceFactory` which is implemented with `TestDataValueReferenceFactory`, label is used here because it does not implement `IDataValueReference`

warrenbuckley: Thanks did not spot the `TestDataValueReferenceFactory` at the bottom of the test class. All good will merge in :)

PR #7604, opened 2020-02-07
7168 hotkeys in rte

nathanwoulfe:
Adds directive to grid rte to fix the 1/2 key press issue previously fixed in the standalone rte
Have merged v8/contrib into my local branch, so should be smooooth sailing.


nul800sebastiaan:
Thanks very much @nathanwoulfe - works like a charm! 👍

Issue #7594, opened 2020-02-06
As developer I want to be able to select no user in a user picker, so the field can be empty.


ThomasHerckenrath:
When I add a user picker to a document type, initially no user is set as value for the field. When I choose a user as value I want to be able to select no user again, wich is currently not possible.

nul800sebastiaan:
Good point! It would be great if someone can have a look at updating the user picker to do this. 

What Umbraco version are you using by the way? For v7, we're not building any new features, so you'd have to implement a custom user picker that includes a way to remove the picked user. 
We'd love to have this feature added to v8 though.

umbrabot:
Hi @ThomasHerckenrath,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

bjarnef:
At the moment the User Picker is just a dropdown. It would be great to make it more consistent with other pickers, e.g. Member Picker.

![image](https://user-images.githubusercontent.com/2919859/74021437-d6c1b900-499b-11ea-957f-b01ca15958d6.png)


Furthermore User Picker doesn't have any specific configuration, but it would be nice to be able to configurate it to allow picking specific user roles (user grops), but the same seems to be the case for Member Picker.

bjarnef:
Okay, the User Picker is actually not a specific User Picker, but a general Entity Picker: https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/entitypicker where Member Picker is its own property editor: https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/memberpicker

It would be great to create a specific user picker property editor (maybe we can do that and change the view without changing the data format it save, so it doesn't break on upgrade). As a start it might be possible to change the entity picker to be more consistent with other pickers and allow the remove the selected item.

bjarnef:
It seems the `entitypicker` is used by User Picker + some parameter editors:
[...]

PR #7613, opened 2020-02-07
Add user picker property editor

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7594

### Description
This PR add a new user picker property editor instead of using enity picker. At the moment we can't use `treePicker` in `editorService` similar to member picker, because mapping of `User` entity is not handled like `Member`.
https://github.com/umbraco/Umbraco-CMS/blob/a9d45413c89d67280ce54c7fd4182e90323bab6c/src/Umbraco.Web/Editors/EntityController.cs#L921

However `editorService` already have a `userPicker` function used in users section. It would be great to extend this to support multiple like `mediaPicker` (maybe also for `memberPicker later) and being able to configurate this on the configuration of the property editor. For now we can just keep it as a single picker.

A quick demo:


bjarnef:
It is now possible to select a user and also remove the selected user. I have also made a few changes on `umb-user-preview` to make it a bit more consistent with `umb-node-preview`.

The `userPicker` now also works as a single and multi picker (default is multi picker since it is the current behaviour). Later the we could add a the configuration option on the user picker property editor to decide whether it should be a single or multi picker (we can do something similar for member picker).

At the moment the user picker just stores an id like the exiting picker (using entity picker), but if we later allow this configuration, we also need to update the property value converterhttps://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/UserPickerPropertyEditor.cs to return a single object or a collection simialar to media picker, MNTP, url picker, etc.

Since `entityResource.getByIds` doesn't support "User" and we would like the avatars as well to user `umb-user-preview` like the user picker on user groups, I have added a new function `getUsers` in `usersResource`.



**Test notes**
- Select a user via user picker and ensure the picked user is shown after selection. Also ensure that user is shown after save and refresh.
- Ensure multiple users still can be selected in user picker on a user group

poornimanayar:
Hey @bjarnef ,

You seem very busy! Thank you for the work. We shall review your changes and get back to you :-)

Poornima

nul800sebastiaan:
What a lovely update and very useful to have a dedicated user picker! Thanks very much @bjarnef! 👍 

> The userPicker now also works as a single and multi picker (default is multi picker since it is the current behaviour). Later the we could add a the configuration option on the user picker property editor to decide whether it should be a single or multi picker (we can do something similar for member picker).

Yes! Would be great to be able to configure the datatype as single or multiple

> At the moment the user picker just stores an id like the exiting picker (using entity picker), but if we later allow this configuration, we also need to update the property value converterhttps://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/UserPickerPropertyEditor.cs to return a single object or a collection simialar to media picker, MNTP, url picker, etc.

Yep, that would be excellent as well! 👍

bjarnef:
@nul800sebastiaan I would like if the different pickers could use `Minimum` and `Maximum` prevalues to configurate is the picker is multipicker or not.
See this PR for multiple member picker https://github.com/umbraco/Umbraco-CMS/pull/7634 and this feature request for media picker https://github.com/umbraco/Umbraco-CMS/issues/8471

At the moment the user picker use `Interger` but need to be `String` instead to store multiple komma delimited values and maybe `Udi` instead, but at the moment I think the API only support to get an user by `int` id.
Furthermore do we have an frontend model for user entity like `IPubslihedContent`. It would be great with a property value converter for user picker, but it probably need to returns something else than `IUser` or `IEnumerable<IUser>`.

PR #7614, opened 2020-02-07
V8/feature/ab4550 segments ui variant picker

PerplexDaniel:
### Description
@nielslyngsoe Amazing work already, looks great! I was checking it for a bit and in the process already made some fixes for stuff I ran into hope that's OK. Sending you these as a PR so you can easily view / test them yourself.

The fixes are mainly for situations where "Vary by culture" is not enabled on property-level or on content-type level.  With certain combinations the UI regarding "Inherited by <language>" was not correct and fields were not properly disabled. In addition, in splitview mode where multiple variants are saved/published a variant with culture `null` and some segment `s` was not being saved at all. 

There are probably some more things that need to be refined, for example the `getVariantDisplayName` function in Angular might be removed if we introduce a `DisplayName` (or similar) property on the `ContentVariantDisplay` class in C# instead. This saves binding to functions in AngularJS which isn't great for performance.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbtabbedcontent.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbtabbedcontent.directive.js
@@ -137,6 +137,24 @@
                     }
                 }
             );
+
+            $scope.propertyEditorDisabled = function (property) {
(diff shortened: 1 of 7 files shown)


nielslyngsoe:
@PerplexDaniel Looks good, I'm merging. :-)

Issue #7538, opened 2020-01-30
NetCore: Introduce Image Url Generator Abstractions


bergmania:
### Background History
For the .NET Core transformation, we need to be able to abstraction the image URL generation, because ImageProcessor is not supported.

### The task
This task is about introducing an `IImageUrlGenerator ` and an `ImageProcessorImageUrlGenerator`.

The interface only requires a single method
```
string GetImageUrl(ImageUrlGenerationOptions options);
```
And `ImageUrlGenerationOptions` needs to have all the options used for image generation in Umbraco. Like:
```
int Width;
int Height;
AnimationProcessMode(Enum) animationProcessMode;
ImageCropMode CropMode
ImageCrop AnchorAnchor
...
```
The `GetImageUrl` implementation is not responsible of looking up information from a named crop. This look up needs to be done before calling `GetImageUrl`

The `ImageProcessorImageUrlGenerator` need to take the functionality from `ImageCropperTemplateExtensions.GetCropUrl` and `ImageCropperTemplateExtensions.GetCropUrl` should call the new service with the options.

It is important that it is easy/possible to replace the `ImageProcessorImageUrlGenerator` with a custom implementation of `ImageUrlGenerationOptions`.

We will need to get the `IImageUrlGenerator` from the `Current` service locator in Razor views.

Occurrences that needs to replace hardcoded url generation:

| File        |
| ------------- |
| UserExtensions.cs    | 
| ImageCropperValue.cs    | 
| ImagesController.cs    | 
| ImageCropperTest.cs    | 
| ImageCropperTemplateExtensions.cs    | 
| RichTextEditorPastedImages.cs    | 
| HtmlImageSourceParserTests.cs    | 
| Media.cshtml    | 

Even though this change is for the .NET Core transition, it should be a non-breaking change, that makes sense to implement on Umbraco 8 ([v8/dev)](https://github.com/umbraco/Umbraco-CMS/tree/v8/dev).

benjaminc:
I'll take a stab at this

Issue #7540, opened 2020-01-30
NetCore: Use API to get Image URLs in BackOffice instead of hardcoded querystrings


bergmania:
### Background History
For the .NET Core transformation, we need to be able to abstract the image URL generation, because ImageProcessor is not supported in .NET Core.

### The task
This task is about changing a variety of Umbraco back-office JavaScript to use the `ImageUrlGenerationController` introduced in issue #7539. This will generate image URLs with various parameters in a way that doesn’t hardcode the parameter names or format. The following JavaScript files currently generate image URLs by appending query string parameters and will need to be changed to use the controller method:

```
Media.controller.js - lines 68-75, 82
Tinymce.service.js - line 308, 1496
Fileupload.controller.js - line 54
Imagecropper.controller.js - line 222
```

Even though this change is for the .NET Core transition, it should be a non-breaking change that makes sense to implement on Umbraco 8 ([v8/dev](https://github.com/umbraco/Umbraco-CMS/tree/v8/dev)).

bielu:
@bergmaniain .net core we could use that instead:
https://github.com/SixLabors/ImageSharp/issues
maybe it makes sense to look how we could use ImageSharp .net core and Image processor still in framework?

bergmania:
Hi @bielu.. we are aware of ImageSharp, but the change in license is not a perfect match. We expect that it will be possible to implement multiple different image manipulation libraries, such that a ImageSharp could make sense for companies that would like the commercial license.

bielu:
@bergmania from my understanding of change of licence as long they will not release RC, all OSS could reference them and use them even after release RC on same Apache licence, based on that what was said here:
https://github.com/SixLabors/ImageSharp/issues/1024
@JimBobSquarePants could you correct me if I am wrong about that?

JimBobSquarePants:
@bielu Introduce it as a dependency now and you get to preserve the Apache license.

bielu:
@JimBobSquarePants thanks for confirmation :)

bielu:
@bergmania, in that case, should we look into how to integrate imagesharp into Umbraco with the possibility to switch to any image processor?

bergmania:
@bielu,
I think it makes sense to do.
I theory it is a task a little bit like this https://github.com/umbraco/Umbraco-CMS/issues/7541, just with ImageSharp instead of SkiaSharp.

[...]

Issue #7539, opened 2020-01-30
NetCore: Introduce API to generate image URLs for BackOffice


bergmania:
### Background History
For the .NET Core transformation, we need to be able to abstract the image URL generation, because ImageProcessor is not supported in .NET Core.

### The task
This task is about introducing an `ImageUrlGenerationController` to generate URLs for images within the Umbraco back-office. The controller will extend the UmbracoAuthorizedJsonController class, and will implement a single method that should call `ImageCropperTemplateExtensions.GetCropUrl` to generate a URL given the arguments. At a minimum, the controller method should take as arguments the `width`, `height`, `mode`, and `animationProcessMode` arguments of the `GetCropUrl` method.

Even though this change is for the .NET Core transition, it should be a non-breaking change that makes sense to implement on Umbraco 8 ([v8/dev](https://github.com/umbraco/Umbraco-CMS/tree/v8/dev)).

benjaminc:
This has been implemented in PR #7616

PR #7616, opened 2020-02-07
Introduce Image URL Generator abstraction

benjaminc:
### Prerequisites

- [ X ] I have added steps to test this contribution in the description below

This resolves issue #7538 

### Description

I created an IImageUrlGenerator interface, along with an ImageProcessorImageUrlGenerator implementation.  I injected that implementation at startup, reworked the ImageCropperTemplateExtensions and ImageCropperValue classes to use it, and then updated other classes and tests that were hard-coded with image URL parameters so that they used the implementation.  This included adding constructors to various classes so that they could accept an injected IImageUrlGenerator implementation, and then obsoleting the previous constructor so that we can delete it in the .NET Core project.

To test this you should spin up a new site with the starter pack, and verify that cropped media show up correctly on the front end.  Specifically, test the following:

- A simple cropped image used directly in a template
- A cropped image used within an RTE
- A cropped image used within an RTE inside a grid
- A custom user avatar in the back-office

Existing unit tests for ImageCropper have been updated to use the new class, but no new unit tests were added.

### DIFF
--- a/src/Umbraco.Core/Composing/Current.cs
+++ b/src/Umbraco.Core/Composing/Current.cs
@@ -5,6 +5,7 @@ using Umbraco.Core.Dictionary;
 using Umbraco.Core.IO;
 using Umbraco.Core.Logging;
 using Umbraco.Core.Mapping;
+using Umbraco.Core.Models;
 using Umbraco.Core.Models.PublishedContent;
 using Umbraco.Core.PackageActions;
 using Umbraco.Core.Packaging;
@@ -208,6 +209,8 @@ namespace Umbraco.Core.Composing
         public static IVariationContextAccessor VariationContextAccessor
             => Factory.GetInstance<IVariationContextAccessor>();
 
+        public static IImageUrlGenerator ImageUrlGenerator
+            => Factory.GetInstance<IImageUrlGenerator>();
         #endregion
     }
 }
--- a/src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
+++ b/src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
@@ -1,8 +1,17 @@
 ﻿namespace Umbraco.Core.Models
 {
+    /// <summary>
+    /// These are options that are passed to the IImageUrlGenerator implementation to determine
+    /// the propery URL that is needed
+    /// </summary>
     public class ImageUrlGenerationOptions
     {
-        public string ImageUrl { get; set; }
+        public ImageUrlGenerationOptions (string imageUrl)
+        {
+            ImageUrl = imageUrl;
+        }
+
+        public string ImageUrl { get; }
         public int? Width { get; set; }
         public int? Height { get; set; }
         public decimal? WidthRatio { get; set; }
@@ -18,18 +27,40 @@
         public bool UpScale { get; set; } = true;
         public string AnimationProcessMode { get; set; }
 
+        /// <summary>
+        /// The focal point position, in whatever units the registered IImageUrlGenerator uses,
+        /// typically a percentage of the total image from 0.0 to 1.0.
+        /// </summary>
         public class FocalPointPosition
         {
-            public decimal Left { get; set; }
-            public decimal Top { get; set; }
+            public FocalPointPosition (decimal top, decimal left)
+            {
+                Left = left;
+                Top = top;
+            }
+
+            public decimal Left { get; }
+            public decimal Top { get; }
         }
 
+        /// <summary>
+        /// The bounds of the crop within the original image, in whatever units the registered
+        /// IImageUrlGenerator uses, typically a percentage between 0 and 100.
+        /// </summary>
         public class CropCoordinates
         {
-            public decimal X1 { get; set; }
-            public decimal Y1 { get; set; }
-            public decimal X2 { get; set; }
-            public decimal Y2 { get; set; }
+            public CropCoordinates (decimal x1, decimal y1, decimal x2, decimal y2)
+            {
+                X1 = x1;
+                Y1 = y1;
+                X2 = x2;
+                Y2 = y2;
+            }
+
+            public decimal X1 { get; }
+            public decimal Y1 { get; }
+            public decimal X2 { get; }
+            public decimal Y2 { get; }
         }
     }
 }
--- a/src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
+++ b/src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
@@ -1,0 +1,35 @@
+﻿namespace Umbraco.Core.Models
+{
+    public class ImageUrlGenerationOptions
+    {
+        public string ImageUrl { get; set; }
+        public int? Width { get; set; }
+        public int? Height { get; set; }
+        public decimal? WidthRatio { get; set; }
+        public decimal? HeightRatio { get; set; }
+        public int? Quality { get; set; }
+        public string ImageCropMode { get; set; }
+        public string ImageCropAnchor { get; set; }
+        public bool DefaultCrop { get; set; }
+        public FocalPointPosition FocalPoint { get; set; }
+        public CropCoordinates Crop { get; set; }
+        public string CacheBusterValue { get; set; }
+        public string FurtherOptions { get; set; }
+        public bool UpScale { get; set; } = true;
+        public string AnimationProcessMode { get; set; }
+
+        public class FocalPointPosition
+        {
+            public decimal Left { get; set; }
+            public decimal Top { get; set; }
+        }
+
+        public class CropCoordinates
+        {
+            public decimal X1 { get; set; }
+            public decimal Y1 { get; set; }
+            public decimal X2 { get; set; }
+            public decimal Y2 { get; set; }
+        }
+    }
+}
--- a/src/Umbraco.Core/Models/UserExtensions.cs
+++ b/src/Umbraco.Core/Models/UserExtensions.cs
@@ -106,13 +106,14 @@ namespace Umbraco.Core.Models
 
             //use the custom avatar
             var avatarUrl = Current.MediaFileSystem.GetUrl(user.Avatar);
+            var urlGenerator = Current.ImageUrlGenerator;
             return new[]
             {
-                avatarUrl  + "?width=30&height=30&mode=crop",
-                avatarUrl  + "?width=60&height=60&mode=crop",
-                avatarUrl  + "?width=90&height=90&mode=crop",
-                avatarUrl  + "?width=150&height=150&mode=crop",
-                avatarUrl  + "?width=300&height=300&mode=crop"
+                urlGenerator.GetImageUrl(new ImageUrlGenerationOptions { ImageUrl = avatarUrl, ImageCropMode = "crop", Width = 30, Height = 30 }),
+                urlGenerator.GetImageUrl(new ImageUrlGenerationOptions { ImageUrl = avatarUrl, ImageCropMode = "crop", Width = 60, Height = 60 }),
+                urlGenerator.GetImageUrl(new ImageUrlGenerationOptions { ImageUrl = avatarUrl, ImageCropMode = "crop", Width = 90, Height = 90 }),
+                urlGenerator.GetImageUrl(new ImageUrlGenerationOptions { ImageUrl = avatarUrl, ImageCropMode = "crop", Width = 150, Height = 150 }),
+                urlGenerator.GetImageUrl(new ImageUrlGenerationOptions { ImageUrl = avatarUrl, ImageCropMode = "crop", Width = 300, Height = 300 })
             };
 
         }
--- a/src/Umbraco.Core/Models/UserExtensions.cs
+++ b/src/Umbraco.Core/Models/UserExtensions.cs
@@ -109,11 +109,11 @@ namespace Umbraco.Core.Models
             var urlGenerator = Current.ImageUrlGenerator;
             return new[]
             {
-                urlGenerator.GetImageUrl(new ImageUrlGenerationOptions { ImageUrl = avatarUrl, ImageCropMode = "crop", Width = 30, Height = 30 }),
-                urlGenerator.GetImageUrl(new ImageUrlGenerationOptions { ImageUrl = avatarUrl, ImageCropMode = "crop", Width = 60, Height = 60 }),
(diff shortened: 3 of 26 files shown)


src/Umbraco.Web/Models/ImageProcessorImageUrlGenerator.cs
```diff
@@ -0,0 +1,108 @@
...
+                imageProcessorUrl.Append("&cropmode=percentage");
+            }
+            else if (options.DefaultCrop)
+            {
+                imageProcessorUrl.Append("?anchor=center");
+                imageProcessorUrl.Append("&mode=crop");
+            }
+            else
+            {
+                imageProcessorUrl.Append("?mode=" + options.ImageCropMode.ToString().ToLower());
```
AndyButland: Think you should probably continue to use the `StringBuilder` here (and in similar constructs), i.e. `imageProcessorUrl.Append("?mode=").Append(options.ImageCropMode.ToString().ToLower());`

src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public class ImageUrlGenerationOptions
+    {
+        public string ImageUrl { get; set; }
+        public int? Width { get; set; }
+        public int? Height { get; set; }
+        public decimal? WidthRatio { get; set; }
+        public decimal? HeightRatio { get; set; }
+        public int? Quality { get; set; }
+        public string ImageCropMode { get; set; }
```
AndyButland: Think it's worth making this into an `enum`?  Just as it has a fixed set of valid values.  Not sure though, as I guess we want this class to be generic for image processor and any other potential implementations.

AndyButland:
Good to see this @benjaminc - nice work.  I had a look over and left a couple of minor comments.

benjaminc:
Thanks @AndyButland.  I considered both those things, and initially I was proceeding on trying to make as few changes as possible (so the StringBuilder issues, for example were in the original), it probably does make sense to just do it right.  No sense leaving a mess if we can clean it up without breaking anything.  I'll fix those and resubmit, and I think I'm going to submit fixes for #7539 and #7540 in this same pull request so that it can all get done at once, since they depend on each other.

src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
benjaminc: Yeah, on this one I think we probably shouldn't go with an Enum, just because the exact values that are allowed for any of the three enums are processor dependent, i.e. ImageCropMode, ImageCropAnchor, and AnimationProcessMode.

src/Umbraco.Web/Models/ImageProcessorImageUrlGenerator.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System.Globalization;
+using System.Text;
+using Umbraco.Core;
+using Umbraco.Core.Models;
+
+namespace Umbraco.Web.Models
+{
+    internal class ImageProcessorImageUrlGenerator : IImageUrlGenerator
+    {
+        public string GetImageUrl(ImageUrlGenerationOptions options)
```
bergmania: I think we should separate several private methods from this method..

Eg:
```csharp
        private void AppendCrop(StringBuilder imageProcessorUrl, ImageUrlGenerationOptions.CropCoordinates crop)
        {
            imageProcessorUrl.Append("?crop=");
            imageProcessorUrl.Append(crop.X1.ToString(CultureInfo.InvariantCulture)).Append(",");
            imageProcessorUrl.Append(crop.Y1.ToString(CultureInfo.InvariantCulture)).Append(",");
            imageProcessorUrl.Append(crop.X2.ToString(CultureInfo.InvariantCulture)).Append(",");
            imageProcessorUrl.Append(crop.Y2.ToString(CultureInfo.InvariantCulture));
            imageProcessorUrl.Append("&cropmode=percentage");
        }

        private void AppendFocalPoint(StringBuilder imageProcessorUrl, ImageUrlGenerationOptions.FocalPointPosition focalPoint)
        {
            imageProcessorUrl.Append("?center=");
            imageProcessorUrl.Append(focalPoint.Top.ToString(CultureInfo.InvariantCulture)).Append(",");
            imageProcessorUrl.Append(focalPoint.Left.ToString(CultureInfo.InvariantCulture));
            imageProcessorUrl.Append("&mode=crop");
        }
```

src/Umbraco.Web/PropertyEditors/RichTextPropertyEditor.cs
```diff
@@ -63,14 +75,16 @@ internal class RichTextPropertyValueEditor : DataValueEditor, IDataValueReferenc
             private readonly HtmlImageSourceParser _imageSourceParser;
             private readonly HtmlLocalLinkParser _localLinkParser;
             private readonly RichTextEditorPastedImages _pastedImages;
+            private readonly IImageUrlGenerator _imageUrlGenerator;
 
-            public RichTextPropertyValueEditor(DataEditorAttribute attribute, IUmbracoContextAccessor umbracoContextAccessor, HtmlImageSourceParser imageSourceParser, HtmlLocalLinkParser localLinkParser, RichTextEditorPastedImages pastedImages)
+            public RichTextPropertyValueEditor(DataEditorAttribute attribute, IUmbracoContextAccessor umbracoContextAccessor, HtmlImageSourceParser imageSourceParser, HtmlLocalLinkParser localLinkParser, RichTextEditorPastedImages pastedImages, IImageUrlGenerator imageUrlGenerator)
```
bergmania: We need to have the old constructor with ` [Obsolete("Use the constructor which takes an IImageUrlGenerator")]`

src/Umbraco.Core/Composing/Current.cs
```diff
@@ -208,6 +209,8 @@ public static IPublishedValueFallback PublishedValueFallback
         public static IVariationContextAccessor VariationContextAccessor
             => Factory.GetInstance<IVariationContextAccessor>();
 
+        public static IImageUrlGenerator ImageUrlGenerator
```
bielu: Is any reason why we are using that in current instead of direct injecting in places of use?

src/Umbraco.Core/PropertyEditors/ValueConverters/ImageCropperValue.cs
```diff
@@ -59,77 +61,72 @@ public ImageCropperCrop GetCrop(string alias)
...
         /// Gets the value image url for a specified crop.
         /// </summary>
-        public string GetCropUrl(string alias, bool useCropDimensions = true, bool useFocalPoint = false, string cacheBusterValue = null)
+        [Obsolete("Use the overload that takes an IImageUrlGenerator")]
+        public string GetCropUrl(string alias, bool useCropDimensions = true, bool useFocalPoint = false, string cacheBusterValue = null) => GetCropUrl(alias, Current.ImageUrlGenerator, useCropDimensions, useFocalPoint, cacheBusterValue);
+
+        /// <summary>
+        /// Gets the value image url for a specified crop.
+        /// </summary>
+        public string GetCropUrl(string alias, IImageUrlGenerator imageUrlGenerator, bool useCropDimensions = true, bool useFocalPoint = false, string cacheBusterValue = null)
```
bielu: Is any reason to insert ImageurlGenator instead of using internal variable, as we register that as Unique, it means we probably could use one ImageUrlGenator, instead of allowing inject few of them in the same time?

src/Umbraco.Web/Models/ImageProcessorImageUrlGenerator.cs
bielu: Is any reason why we set up ImageProcessorImageUrlGenerator as internal? I think it should be public to allow for overrides :)

src/Umbraco.Web/PropertyEditors/RichTextPropertyEditor.cs
benjaminc: The reason I didn't keep the old constructor was because this was an Internal class, so it can't be referenced outside the solution.  Since we updated all the references in the solution, there is no reason to keep the old constructor.

src/Umbraco.Core/Composing/Current.cs
benjaminc: I could use Current.Factory instead, but all the existing extension methods need some way to get the current generator without already having an IImageUrlGenerator instance.

src/Umbraco.Core/PropertyEditors/ValueConverters/ImageCropperValue.cs
benjaminc: I'm not sure what you mean by this.  The ImageCropperValue class is a model class either manually constructed or deserialized from JSON.  Since we can't pass it in through the constructor due to the deserialization, and we don't want a static reference for injection purposes, we need to pass this in to the method.

bielu: I didn't notice it is class which is deserialized from JSON my comment is not really relevant in that case :)

Shazwazza:
@bergmania  + @benjaminc ... this has been merged into v8/contrib! .... pretty sure we definitely don't want that? Or are we planning on adding this to v8 and not just the netcore branch? I wonder if anything else netcore related has accidentally been merged to v8 too?

Shazwazza:
On another note, we shouldn't keep these old branches around, normally if a PR is merged and the branch is part of this repo, we need to click the Delete Branch button

src/Umbraco.Core/Models/UserExtensions.cs
```diff
@@ -106,13 +106,14 @@ internal static string[] GetUserAvatarUrls(this IUser user, IAppCache cache)
...
             var avatarUrl = Current.MediaFileSystem.GetUrl(user.Avatar);
+            var urlGenerator = Current.ImageUrlGenerator;
             return new[]
             {
-                avatarUrl  + "?width=30&height=30&mode=crop",
-                avatarUrl  + "?width=60&height=60&mode=crop",
-                avatarUrl  + "?width=90&height=90&mode=crop",
-                avatarUrl  + "?width=150&height=150&mode=crop",
-                avatarUrl  + "?width=300&height=300&mode=crop"
+                urlGenerator.GetImageUrl(new ImageUrlGenerationOptions(avatarUrl) { ImageCropMode = "crop", Width = 30, Height = 30 }),
```
Shazwazza: I think it would be useful to have some static factory methods on `ImageUrlGenerationOptions` to avoid this type of duplication for common tasks like cropping. For example: `ImageUrlGenerationOptions.CropUrl(avatarUrl, 30, 30)` would be much simpler.

Shazwazza: We could even go so far as to make a fluent pattern for this, for example:

`ImageUrlGenerationOptions.Create(avatarUrl).Width(30).Height(30).Crop().Upscale()`

benjaminc:
@shazwazza this was intentionally merged into v8/contrib. We wanted to introduce the part that was non-breaking at v8 so that it could be used and tested more, and so that it would be easier to keep the netcore branch up to date with v8. I also intentionally didn't delete the branch in merged because I needed to manually merge it into netcore/dev before the next v8 to netcore merge. Now that I've done that, I would like to delete the branch, but don't see where to do that in GitHub. Any help would be greatly appreciated!

Shazwazza:
@benjaminc oh phew 😅 good stuff then, thought we'd have to do some git magic. All good then :) I can delete the branch now, maybe it's only admins.

src/Umbraco.Core/Models/UserExtensions.cs
benjaminc: I like the fluent idea.  What is the final Upscale call for?  Are you thinking the Create method would return a builder object, and the Upscale call would return the Options object?

If you want to leave the branch open, I can go ahead and implement this today.

Shazwazza: It's better to submit a new PR on a new branch, makes it easier to review. This was just an example, but `Upscale` would just be setting `upscale=true`. All of this could mostly be done with extension methods since they are just writing to properties so Create could really just return an instance of ImageUrlGenerationOptions. I'm just proposing the idea, not the implementation though, will leave that up to you :)

src/Umbraco.Web/Models/ImageProcessorImageUrlGenerator.cs
ronaldbarendse: I agree with @bielu: make this class public and (most) helper methods protected, so it can be easily overwritten. One _very_ useful example is to add a token to all generated image URLs that can be validated using `ImageProcessingModule.OnProcessQuerystring`, so it's possible to protect against a DoS attack (requesting a lot of images with different parameters that require a lot of server processing power to generate), see: https://github.com/jimbobsquarepants/imageprocessor/issues/49#issuecomment-199337644

src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+        public string ImageCropAnchor { get; set; }
+        public bool DefaultCrop { get; set; }
+        public FocalPointPosition FocalPoint { get; set; }
+        public CropCoordinates Crop { get; set; }
+        public string CacheBusterValue { get; set; }
+        public string FurtherOptions { get; set; }
+        public bool UpScale { get; set; } = true;
+        public string AnimationProcessMode { get; set; }
+
+        public class FocalPointPosition
```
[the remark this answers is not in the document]
ronaldbarendse: It's probably better to make this a `struct` and ensure it has the correct `GetHashCode` and `Equals` (or better yet: `IEquatable<T>`) implementations. This makes it almost the same as: https://docs.microsoft.com/en-us/dotnet/api/system.drawing.pointf.

src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            public decimal Left { get; set; }
+            public decimal Top { get; set; }
+        }
+
+        public class CropCoordinates
```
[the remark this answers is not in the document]
ronaldbarendse: Just like `FocalPointPosition`: make this a struct, so it's almost the same as: https://docs.microsoft.com/en-us/dotnet/api/system.drawing.rectanglef.

PR #7618, opened 2020-02-08
Netcore: Migrate various extension methods on IPublishedContent, IPublishedElement and IPublishedProperty to Umbraco.Abstractions

AndyButland:
Implements ##7617.

In the [previous PR](https://github.com/umbraco/Umbraco-CMS/pull/7542) I worked on I found I had a need to move an extension method on `IPublishedContent` into `Umbraco.Abstractions` to be able to access it from an "infrastructure" project.  To to that I:

- Moved it (and a few others) to `Umbraco.Abstractions`, introducing parameters to the method signature for any service located dependencies
- Kept the original extension method as was, and delegated to the new one, passing the service located dependencies.

In this PR I've continued that for other extension methods on `IPublishedContent`, `IPublishedElement`, and `IPublishedProperty`.

For each extension method I've either:

- Simply moved it, as it had no service located dependencies and didn't depend on anything that wasn't already in `Umbraco.Abstractions`
- Done what I've described above (i.e. copied, introduced method parameters, and delegated from the original extension method in `Umbraco.Web` - so we keep the "friendly" one that uses service location).
- Left as they are as there were dependencies I couldn't remove or move (e.g. use of `UrlProvider` or `UmbracoContext` defined in `Umbraco.Web`, a dependency on `IExamineManager`, which, although an abstraction, it's an interface defined within the Examine library which, presumably, we don't want to add a reference to `Umbraco.Abstractions` for.

Hopefully this all makes sense and aligns with the aims we discussed last week.


bergmania:
Looks perfect - Thanks, this will make my abstraction and movement of UmbracoContext easier  😄

PR #7619, opened 2020-02-10
NetCore: Untangle and abstract UmbracoContext

bergmania:
- Introduced an interface on `UmbracoContext`
- Introduced an interface on `WebSecurity`
- Introduced an interface on `PublishedRequest`
- Introduced an interface on `UrlProvider` (Can't be named `IUrlProvider`) 
  - Used name `IPublishedUrlProvider` - Suggestions?
- Removed `UmbracoContext.HttpContext` 
  - Replaced with injections of `IHttpContextAccessor`
- Finally moved interfaces to `Umbraco.Abstractions`
  - Separated `UriUtility` in a core version that lives in `Umbraco.Abstractions`

- Also find a bug with macros. I needed to update the some `CacheRefresher` to use the `PayloadCacheRefresherBase`


src/Umbraco.Abstractions/Security/IWebSecurity.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// </summary>
+        /// <value>The current user.</value>
+        IUser CurrentUser { get; }
+
+        /// <summary>
+        /// Logs a user in.
+        /// </summary>
+        /// <param name="userId">The user Id</param>
+        /// <returns>returns the number of seconds until their session times out</returns>
+        double PerformLogin(int userId);
```
Shazwazza: We need to remove this, it's not actually used for login, i'll mark as obsolete for now

src/Umbraco.Web/UmbracoContextFactory.cs
```diff
@@ -79,14 +80,26 @@ public UmbracoContextReference EnsureUmbracoContext(HttpContextBase httpContext
...
 
             var umbracoContext = CreateUmbracoContext(httpContext);
             _umbracoContextAccessor.UmbracoContext = umbracoContext;
 
             return new UmbracoContextReference(umbracoContext, true, _umbracoContextAccessor);
         }
 
+        public static HttpContextBase EnsureHttpContext(HttpContextBase httpContext = null)
+        {
+            if (Thread.GetDomain().GetData(".appPath") is null || Thread.GetDomain().GetData(".appVPath") is null)
```
Shazwazza: prob best to store the result of `Thread.GetDomain()` instead of re-calling the method. I'll update

Shazwazza: Not quite sure what this code is doing though, seems to be new?

src/Umbraco.Abstractions/IUmbracoContext.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        /// <summary>
+        /// Boolean value indicating whether the current request is a front-end umbraco request
+        /// </summary>
+        bool IsFrontEndUmbracoRequest { get; }
+
+        /// <summary>
+        /// Gets the url provider.
+        /// </summary>
+        IPublishedUrlProvider UrlProvider { get; }
```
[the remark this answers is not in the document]
bergmania: I think it makes sense to try it out, but that should be a separate PR IMO.

src/Umbraco.Web/UmbracoContextFactory.cs
bergmania: Understandable - If also more or less found the solution using trial-and-error.
After I refactored all of this, we use the fake HttpContext more often in tests.
Some tests fail if we only have the original code
```
return httpContext ?? new HttpContextWrapper(HttpContext.Current ??
                                                         new HttpContext(new SimpleWorkerRequest("null.aspx", "", NullWriterInstance)));
```

because the constructor of `SimpleWorkerRequest` throws a `NullReferenceException`. if `Thread.GetDomain().GetData(".appPath")` or `Thread.GetDomain().GetData(".appVPath")` is `null`.  The constructor is here:
```
    public SimpleWorkerRequest(string page, string query, TextWriter output)
      : this()
    {
      this._queryString = query;
      this._output = output;
      this._page = page;
      this.ExtractPagePathInfo();
      this._appPhysPath = Thread.GetDomain().GetData(".appPath").ToString();
      this._appVirtPath = Thread.GetDomain().GetData(".appVPath").ToString();
      this._installDir = HttpRuntime.AspInstallDirectoryInternal;
      this._hasRuntimeInfo = true;
    }
```

bergmania:
Your changes looks good..  I just solved the merge conflict after you merged the other PR

src/Umbraco.Abstractions/IUmbracoContext.cs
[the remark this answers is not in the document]
bergmania: PR: https://github.com/umbraco/Umbraco-CMS/pull/7664/

It was a bit complicated, but seems to work :)

src/Umbraco.Web/UmbracoContextFactory.cs
bergmania: Btw - The method is completely removed in https://github.com/umbraco/Umbraco-CMS/pull/7664/, because we there is more abstraction, so we don't need a httpContext directly

Shazwazza:
Great stuff!

PR #7623, opened 2020-02-10
8.6RC: Prevent multi URL picker from crashing for non-local links

kjac:
### Prerequisites

### Description

As described in #7543, the changes in #7130 causes the multi URL picker to crash with JS errors when it contains non-local (absolute) links.

#### Steps to reproduce

For a piece of content with a multi URL picker:

1. Add a new item to the multi URL picker and enter e.g. "https://somewhere.com" in the link field:
image
2. Save the content item.
3. Reload the browser.
4. The multi URL picker is not displays as empty, and the JS console is full of errors:
image

When this PR is applied, the multi URL picker displays all added items after saving the content and reloading the browser:

image


nul800sebastiaan:
Thanks! I've cherry picked this for the upcoming 8.6 as well! 👍

PR #7624, opened 2020-02-10
V8: Rollback fails for content with numeric or decimal property values

kjac:
### Prerequisites

### Description

If content contains properties of type numeric or decimal (and possibly more), the rollback dialog logs errors in the JS console and the "Rollback" button never gets enabled:

image

image

#### Steps to reproduce

For content with a numeric or decimal property:

1. Enter a value in the numeric/decimal property.
2. Save the content.
3. Reload the browser window.
4. Attempt to rollback the content to a previous version.

This PR ensures that the diff is always performed on string values, thus fixing the error and re-enabling rollback for the content:

image


nul800sebastiaan:
Ha! That was a tricky one! All good 👍

PR #7627, opened 2020-02-10
AB4828 - Reset Password Email

warrenbuckley:
This PR will check if the security stamp has not been changed since the forgotten password email has been sent out.

Such as the email/username changing etc...

## Testing Notes
See original notes on the Azure Board task/card 4828

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/UserRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/UserRepository.cs
@@ -557,6 +557,16 @@ ORDER BY colName";
                 }
             }
 
+            // If userlogin or the email has changed then need to reset security stamp
+            if (changedCols.Contains("userLogin") || changedCols.Contains("userEmail"))
+            {
+                userDto.EmailConfirmedDate = null;
+                userDto.SecurityStampToken = entity.SecurityStamp = Guid.NewGuid().ToString();
+                
+                changedCols.Add("emailConfirmedDate");
+                changedCols.Add("securityStampToken");  
+            }
+
             //only update the changed cols
             if (changedCols.Count > 0)
             {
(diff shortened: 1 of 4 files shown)


warrenbuckley:
@Shazwazza when testing & writing this PR against the test/repro notes on the Azure Board Card #AB4828 

I was under the impression that `SecurityStamp` would be regenerated/invalidated when the user changes their username/email address from the call at sprint planning this morning?

Can you give me a pointer or two please mate 😄

src/Umbraco.Web/Editors/AuthenticationController.cs
```diff
@@ -516,7 +516,8 @@ private string ConstructCallbackUrl(int userId, string code)
                 {
                     area = GlobalSettings.GetUmbracoMvcArea(),
                     u = userId,
-                    r = code
+                    r = code,
```
Shazwazza: We don't want to pass the security stamp separately, we don't want people to know about a security stamp at all. The `code` here should be the hash of the `code` + the current security stamp.

warrenbuckley: Thanks will revert these changes back and look into the SecurityStamp stuff with the email from your notes 💪

src/Umbraco.Core/Persistence/Repositories/Implement/UserRepository.cs
```diff
@@ -557,6 +557,16 @@ protected override void PersistUpdatedItem(IUser entity)
                 }
             }
 
+            // If userlogin or the email has changed then need to reset security stamp
+            if (changedCols.Contains("userLogin") || changedCols.Contains("userEmail"))
```
Shazwazza: Looks good, all that's missing is a unit test :) Should be an easy one to write in the UserRepositoryTests, you can have a look at `Can_Perform_Update_On_UserRepository` for inspiration, but for this one we'll just be testing/asserting this logic.

PR #7630, opened 2020-02-10
V8: Nested Content can't add multiple items 

kjac:
### Prerequisites

### Description

If Nested Content is configured with only one element type, the editor only lets you add a single item - regardless of any upper (or lower) limit being set or not.

The workaround: Add the items one at a time and reload the browser window between adds 😆 

This PR fixes the problem.


kjac:
So this was reported as an issue now... Cross-linking to #7698

warrenbuckley:
Thanks @kjac all tested :)

warrenbuckley:
Cherry-picked into v8/8.6 as well @nul800sebastiaan 
https://github.com/umbraco/Umbraco-CMS/commit/78a46677e5b0937a210d786c5703b74a6935d0e2

PR #7631, opened 2020-02-10
V8: Only show culture for content links if there is more than one culture

kjac:
### Prerequisites

### Description

The content links on the Info app for culture invariant content always seem to show specific culture indicators, even there is only one culture that applies:

content-link-culture-before

This seems a bit silly, so this PR removes those culture indicator unless there's an actual culture difference between the content links:

content-link-culture-after

#### Steps to reproduce

1. Create a site with only one language.
2. Edit any content on the site.
3. The culture indicators are shown.

#### Stuff to test

1. The steps above should not yield any culture indicators for the content.
2. Adding a second language *and* binding a hostname to that language should result in hostname specific culture indicators for the same content.


nul800sebastiaan:
Nice touch! Yep, this works as advertised! 👍

PR #7632, opened 2020-02-10
V8: Align "Field is mandatory" label with the toggle button

kjac:
### Prerequisites

### Description

I'm not sure why the "Field is mandatory" label for properties isn't set as a label for the toggle button... it seems odd, specially with the recently introduced custom error message text field below:

property-mandatory-label-position-before

This PR changes things around so the label becomes part of the toggle button:

property-mandatory-label-position-after


nul800sebastiaan:
Great, looks much better, cheers!

Issue #7607, opened 2020-02-07
Allow to configurate Member Picker


bjarnef:
At the moment Member Picker doesn't have any specific configuration.
It would be great to allow the configurate this (although MNTP can be used) to allow picking members of a specific type and maybe although only members who belong to a specific member group? However member groups can be added and removed from members, so not sure if it could cause some confusing since a member then could be picked, but no longer belongs to a specific member group?

See also https://github.com/umbraco/Umbraco-CMS/issues/7608

nul800sebastiaan:
Thanks, I suspect the answer is: sure, but if you have specific needs you'd be better off making a custom member picker. Not sure, we'll discuss at HQ!

As always, if you have a specific use-case that you're running into then that always a good way to figure out if something makes sense to add to a built-in datatype.

bjarnef:
It would be great to decide whether it should be a single or multi picker similar to media picker, MNTP, multi url picker.

bjarnef:
PR: https://github.com/umbraco/Umbraco-CMS/pull/7634

umbrabot:
Hiya @bjarnef,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

bjarnef:
@umbrabot still relevant
I think this is still relevant.

@nul800sebastiaan is it possible to configurate the bot not to close issues with linked PR? :)

nul800sebastiaan:
No :-( The GitHub API doesn't let me see these links.

nul800sebastiaan:
The bot does look at the `community/pr` label though, but it wasn't present for this one unfortunately.

umbrabot:
Closing this in relation to #7634 - make sure to read the close reason on #7634

marlonscloud:
Why was this never implemented?

marlonscloud:
[...]

Issue #6181, opened 2019-08-23
Feature - Reduce the number of "content" pickers down to only Multi-Node Tree Picker


NikRimington:
So I've been thinking about whether it is worth having the following 3 pickers:

- Content Picker
- Media Picker
- Multi-Node Tree Picker

The functionality of both the content picker and the media picker are present in the Multi-Node Tree Picker and it would make sense if we removed the Content and Media pickers. MNTP is a much more flexible picker and with a few tweaks to it (for example enabling media preview) it could be a direct replacement for both of them. It has better support for defining pickable content types / media types, start nodes (e.g. use of xpath).

MNTP could also have its Property Value Converter enhanced to work in a similar way to Nested Content where if only a single type is available to it then it strongly types the return object. At the minute it always returns IPublishedContent.

I was wondering what peoples thoughts were on this?

nul800sebastiaan:
Initial thoughts: maybe in the future.. 

- Currently MNTP can not filter out media/member types, which is a bug that needs fixing. 
- Then we need to make sure MNTP has the best user experience. In the past it used to be less elegant than the separate pickers (but I haven't compared them side-by-side recently)
- Phasing out the old ones can be done quite easily, we did this in v7 as well, Legacy datatypes would not be available in fresh installs but people who upgraded would still keep their existing datatypes
- We can not mess with PVCs at all, if the result is different than it is today, it will break the frontend of any site that is using the current PVCs.

I am looking forward to community comments as well as we'll have a chat about this at HQ!

Matthew-Wise:
Have to say I cant recall the last time  I used a Content Picker. Like this idea of one picker to rule them all!

mzajkowski:
The only reason why I sometimes used a single one, was a need for strongly casting it into a single object rather than collection of objects. Of course it might be sorted using the value converter, but might be also worth to adjust the MNTP "global" converter to return single object when settings assumes to select only 1 item. Just an idea.

+1 for the unification!

ronaldbarendse:
Just like the `Umbraco.MultiUrlPicker`, this should be added as new data editor with a new value converter (the other editors can then optionally be marked as depreciated). The same features could also be added to the Multi URL Picker: https://github.com/umbraco/Umbraco-CMS/pull/2323#pullrequestreview-194926635 and https://github.com/umbraco/Umbraco-CMS/pull/5449 (this PR got closed as v7 won't include any new features anymore).

[...]

PR #7634, opened 2020-02-10
Multiple member picker

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7607

### Description
This PR add a multi picker configuration option similar to the configuration in media picker. When the member picker is configurated as multi picker it returns `IEnumerable<IPublishedContent>` otherwiser `IPublishedContent`.

image

image

At the moment media picker has a similar option, but maybe two configuration options for minimum and maximum items as in multi url picker is better and more flexible?

https://github.com/umbraco/Umbraco-CMS/blob/eed94d8cd03b04385c244a8f593e01e95f349218/src/Umbraco.Web/PropertyEditors/MultiUrlPickerConfiguration.cs#L8-L12


poornimanayar:
Hey @bjarnef ,

Thats such a fantastic feature! We shall review your work and get back to you.

Regards,
Poornima

bjarnef:
Great! 👍 
Maybe @nul800sebastiaan, @kjac or @nielslyngsoe have some feedback on the multi picker toggle vs minimum and maximum items as on multi url picker.

On some projects I have also had the need to limit media picker to just a few images instead of selecting unlimited number on images. You can of course limit the images in the code, but it would be great if we used same approach across different pickers which can be single or multi pickers.

poornimanayar:
Apologies I didnt get back to you earlier on this @bjarnef . We had an team chat around this and @nul800sebastiaan asked the following

- Before you could only pick 1 member, correct? Now you can configure it to pick multiples?

- There's a property value converter for it, but what happens when you switch from single to multiple? And what happens for people with a custom PVC for member picker?

- Additionally, instead of a checkbox for "Multiple", then just leave it to the options Min and Max, if max is > 1 then it's automatically a Multiple picker.

bjarnef:
Yes, at the moment the member picker only allow picking a single member and with this PR it allow to configurate the editor as single or multi member picker.

I haven't checked specific what happens when changing an existing member picker to be configurated as multi picker, but the PVC is much similar to `MediaPickerValueConverter` https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerValueConverter.cs

`Min` and `Max` would make it a bit more flexible than just `Multiple` (like on media picker). We already have this for MNTP, Multi Url Picker, etc... it would be great with a consistent way to handle single/multi picker via configuration.
See more about this in these issues:
https://github.com/umbraco/Umbraco-CMS/issues/7636
https://github.com/umbraco/Umbraco-CMS/issues/6181

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
```diff
@@ -27,44 +30,77 @@ public override PropertyCacheLevel GetPropertyCacheLevel(IPublishedPropertyType
...
-            var attemptConvertInt = source.TryConvertTo<int>();
-            if (attemptConvertInt.Success)
-                return attemptConvertInt.Result;
-            var attemptConvertUdi = source.TryConvertTo<Udi>();
-            if (attemptConvertUdi.Success)
-                return attemptConvertUdi.Result;
-            return null;
+            var nodeIds = source.ToString()
+                .Split(new[] { "," }, StringSplitOptions.RemoveEmptyEntries)
+                .Select(Udi.Parse)
```
nul800sebastiaan, line 52: @bjarnef Hmm, previously we could be getting an `int` or a `Udi`, now you're just assuming this is always a `Udi`, that can't be right?

nul800sebastiaan: I know it will be a `Udi` list for newly picked values but if people already have a single `int` identifier stored then we break that here.

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
```diff
@@ -27,44 +30,77 @@ public override PropertyCacheLevel GetPropertyCacheLevel(IPublishedPropertyType
...
+                    if (guidUdi == null) continue;
+                    var item = _publishedSnapshotAccessor.PublishedSnapshot.Members.GetById(guidUdi.Guid);
+                    if (item != null)
+                        memberItems.Add(item);
                 }
+
+                return isMultiple ? memberItems : FirstOrDefault(memberItems);
             }
 
+            //if (Current.UmbracoContext != null)
```
nul800sebastiaan, line 81: Make sure to delete commented out code.

src/Umbraco.Web/PublishedCache/NuCache/MemberCache.cs
```diff
@@ -69,6 +69,18 @@ public IPublishedContent GetById(bool preview, int memberId)
                 });
         }
 
+        public IPublishedContent /*IPublishedMember*/ GetById(Guid memberId)
```
nul800sebastiaan, line 72:
```suggestion
        public IPublishedContent GetById(Guid memberId)
```

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
```diff
@@ -27,44 +30,77 @@ public override PropertyCacheLevel GetPropertyCacheLevel(IPublishedPropertyType
...
+            return nodeIds;
         }
 
         public override object ConvertIntermediateToObject(IPublishedElement owner, IPublishedPropertyType propertyType, PropertyCacheLevel cacheLevel, object source, bool preview)
         {
-            if (source == null)
-                return null;
+            var isMultiple = IsMultipleDataType(propertyType.DataType);
+
+            var udis = (Udi[])source;
```
nul800sebastiaan, line 62: Can we just cast `source` to `Udi[]`? Even if there's only one?

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
bjarnef: Yeah, I think I just was inspired from the media picker https://github.com/umbraco/Umbraco-CMS/blob/51c1234db3d747cdb609236daabb5625f8058da0/src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerValueConverter.cs#L59-L62 ... but I guess this already existed since Media Picker V2.

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
bjarnef: Not sure - at the moment it is like media picker https://github.com/umbraco/Umbraco-CMS/blob/51c1234db3d747cdb609236daabb5625f8058da0/src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerValueConverter.cs#L72 .. but maybe it could try cast to different types, e.g. start with cast to `Udi`, then `int`?
Do we have something else in core trying resolving string as `Udi` or `Int` for legacy support?

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
bjarnef: @nul800sebastiaan not sure if this is a left over from previous when member picker V1 stored id, but when the new pickers was refactored in the V2 pickers Content Picker, Media Picker, Media Picker, MNTP, Member Picker stored an `Udi` instead. In Umbraco v8 the "2" in the naming was removed.

Hasn't member picker stored an `Udi` since the initial release of Umbraco v8?

bjarnef: It looks like the `{"idType", "udi"}` has been there for a while.

https://github.com/umbraco/Umbraco-CMS/commit/a97900515b0e7bbfba6e7d32d8175505ab2bf863

bjarnef: I have tested with Umbraco v8.0.0 where it also stored an `Udi`, so I think it is an oversight this part wasn't changed like in Content Picker and Media Picker value converters.

image

nul800sebastiaan: Sure, but there's also people migrating from v7 which is why I'm assuming that int still needs to be there..

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
nul800sebastiaan: Many places I think, something like this: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/ValueConverters/ContentPickerValueConverter.cs#L39

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
bjarnef: @nul800sebastiaan sure, but isn't it an issue with Media Picker as well? https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerValueConverter.cs#L59-L63

or is it what this migration handle? https://github.com/umbraco/Umbraco-CMS/blob/2f978e96d4fbf84d8786c3245566c8800916a74b/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/UpdatePickerIntegerValuesToUdi.cs

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
bjarnef: Is there a migration handling this for MediaPicker, since the value converter doesn't handle `int`?
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerValueConverter.cs#L54-L64

If developers are migrating from v7 using `Umbraco.MediaPicker2` it probably isn't an issue (since this stored an `Udi` and was renamed to just `Umbraco.MediaPicker` before initial release of Umbraco v8). But if people used the obsolete `Umbraco.MediaPicker` in v7 which stored `int` ids and migrate to v8 I guess it is an issue with Media Picker as well?

bjarnef:
Instead of a "Multiple" configuration, I have changed it to use "Minimum" and "Maximum" configuration like most other picker, e.g. Multi Url Picker and MNTP. It would be great if Media Picker supported this as well, but we can look at that in another issue/PR.

image

image

bjarnef:
Now it is also possible to sort the selected members like in MNTP.

bjarnef:
The member picker property editor now also support "Show open button" like MNTP using `memberEditor` infinite editing.

image

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
bjarnef: @Shazwazza do the recall if is neccesary to handle the `int` value in the value converter for migration from old MemberPicker (not MemberPicker2) in v7 to v8. The media picker value converter doesn't seem to handle this anyway, so I guess it would be an issue så well if migrating from v7 to v8 from old MediaPicker which stored `int` ids to the newer one storing 'Udi` ids.

nul800sebastiaan: @bjarnef It's just that the memberpicker doesn't have a migration, content, media and MNTP have been converted to Udi in the v8.0.0 migration - https://github.com/umbraco/Umbraco-CMS/blob/2f978e96d4fbf84d8786c3245566c8800916a74b/src/Umbraco.Core/Migrations/Upgrade/V_8_0_0/UpdatePickerIntegerValuesToUdi.cs#L21-L23

So yes, we'll need to expect int identifiers here.

nul800sebastiaan:
> Instead of a "Multiple" configuration, I have changed it to use "Minimum" and "Maximum" configuration like most other picker 

As mentioned on https://github.com/umbraco/Umbraco-CMS/issues/8471 - this is a breaking change, there's code relying on the "is multiple" variable so we'd need to preserve that one.

nul800sebastiaan:
Ah but of course we never had a multiple member picker before so in this case that would be fine! 👍

src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs
bjarnef: Yeah, but it seems the migration only update the startNodeId i in the configuration. Where did it update the `int` value saved in content from MediaPicker v1 to `Udi` in v2?
The existing member picker didn't have configuration prevalues, but acted a single picker, so by default I have made it work as single picker (where media picker ny default is multi picker).

nul800sebastiaan: 🤦‍♂️  duh.. okay, I didn't read that very closely did I? 😂 

Anyway, would be good to test this out, upgrade from v7 with an int id stored.. not sure what will happen, maybe none of that is supported, maybe not. I really don't know.

bjarnef: It seems Media Picker doesn't support it when migrating from v7 to v8 going from v1 to v2 picker.. but it the v2 media picker was already used in v7, it shouldn't cause any issues since it already store `Udi`.

Member Picker is probably used, but I guess it would be a bigger issue with Media Picker since it is used more and not all projects use Member Picker.

bjarnef:
@nul800sebastiaan just working on a project where I could use this to store multiple member references 🙈

bjarnef:
@nul800sebastiaan do we still want this in v8 or maybe in v9?

Prologh:
So, has this PR been abandoned or has it been merged into Umbraco 9?

biapar:
Where is this datatype? How to add multiple member? Into U10 there isn't.

bjarnef:
This pull request wasn't merged, so I doesn't exists in core. 

For now the MNTP can be used to select mulitple members.

image

image

PR #7635, opened 2020-02-11
V8: Fix the create dialog on empty installs (help newcomers)

### LINKED ISSUE #7571: Describe why there is no create options at the content root
When a user opens the create dialog on the root of the tree, but doesn't have any DocumentTypes allowed for creation in the root, the user sees the following screen:

image

Instead, we should allow all Document Types to be created at root, and thereby list them in the dialog. Or mention that a Document Type should be allowed.

And if there are no Document Types in the solution, we should show a screen describing that and link to the Settings Section, potential a direct link to create a new Document Type.

Something similar to this dialog, this example is taken from the create dialog on a content-node:
image

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7571

### Description

The changes added to aid newcomers in #6573 don't exactly work; the create dialog on empty installs still looks very very empty:

image

This PR fixes it, so the create dialog actually helps newcomers get on with creating content types:

image

The "Create a new document type" of course only shows if the current editor has access to the Settings section.

If a one or more content types do exist but none are marked as allowed at root, the create dialog contains this message:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/content.create.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/content.create.controller.js
@@ -22,10 +22,14 @@ function contentCreateController($scope,
     function initialize() {
         $scope.loading = true;
         $scope.allowedTypes = null;
-        $scope.countTypes = contentTypeResource.getCount;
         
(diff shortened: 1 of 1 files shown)


nielslyngsoe:
@kjac looks really good!

nul800sebastiaan:
Sweet, thanks @kjac !

PR #7637, opened 2020-02-11
V8/bug/ab4899 remember user filtering

nielslyngsoe:
Remember user filtering, by storing it in the URL parameters, like the other filtering options.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/umb-mini-search/umbminisearch.component.js
+++ b/src/Umbraco.Web.UI.Client/src/views/components/umb-mini-search/umbminisearch.component.js
@@ -10,7 +10,8 @@
             bindings: {
                 model: "=",
                 onStartTyping: "&?",
-                onSearch: "&?"
+                onSearch: "&?",
+                onBlur: "&?"
             }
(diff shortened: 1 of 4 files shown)


src/Umbraco.Web.UI.Client/src/views/components/umb-mini-search/umbminisearch.component.js
```diff
@@ -10,7 +10,8 @@
             bindings: {
                 model: "=",
                 onStartTyping: "&?",
```
warrenbuckley: Do we drop `on-start-typing` from this component/directive and get all implementations to use the `on-blur` approach @nielslyngsoe ?

nielslyngsoe: No, just for this specific usage of umb-mini-search. So still used in other places.

warrenbuckley:
Code seems fine to me, but off to test now :)

warrenbuckley:
All tested and works fine, once I was testing on the right branch 😃

PR #7644, opened 2020-02-11
Netcore: Use ServiceContext property from base class instead of Current.Services

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/7369

- Implements the missing usages in tests

Test notes: 
- The tests should pass


elit0451:
Tests pass 💯 Changes look good! 
Merging

PR #7646, opened 2020-02-11
Removed circular dependency between data type service and repository

AndyButland:
Resolves one instance of #4237.

I came across this issue when looking at feasibility of moving models and services into `Umbraco.Abstractions`.  That was too deep a rabbit hole so came up for air again and fixed one small piece - where the data type service depends on the repository (as you'd expect) but there's also a dependency the other way.

It's only needed as part of the process that converts data types with missing property editors into labels.

I've resolved it by instead of creating a label in the repository layer - which then requires lots of dependencies, including the data type service - creating a simpler "marker" editor, which is then converted into a label in the service.


src/Umbraco.Infrastructure/Services/Implement/DataTypeService.cs
```diff
@@ -281,7 +300,31 @@ public IEnumerable<IDataType> GetAll(params int[] ids)
...
+        {
+            if (dataType == null)
+            {
+                return;
+            }
+
+            ConvertMissingEditorsOfDataTypesToLabels(new[] { dataType });
+        }
+
+        private void ConvertMissingEditorsOfDataTypesToLabels(IEnumerable<IDataType> dataTypes)
```
Shazwazza: maybe this can just be `params IDataType[] dataTypes` and then we only need the single method `ConvertMissingEditorsOfDataTypesToLabels`?

PR #7650, opened 2020-02-12
NetCore: Removed legacy way of handling alternative templates..

bergmania:
This code removes some legacy code to support alternative templates.

Today we need to use `ContentFinderByUrlAndTemplate`
See note here: https://our.umbraco.com/Documentation/Reference/Routing/Request-Pipeline/IContentFinder#adding-and-removing-icontentfinders

This PR changes:
- Removes the `"altTemplate"` key from the HttpContext.Items. This value was always `null`
  - The only way this value was set was in: https://github.com/umbraco/Umbraco-CMS/pull/7650/files#diff-61e00a0739f023734bbeda8103c96f24L196 
  - The value is always the result from the cache itself, can is never set. https://github.com/umbraco/Umbraco-CMS/pull/7650/files#diff-61e00a0739f023734bbeda8103c96f24L99-R113

### DIFF
--- a/src/Umbraco.Web/Routing/PublishedRouter.cs
+++ b/src/Umbraco.Web/Routing/PublishedRouter.cs
@@ -20,7 +20,6 @@ namespace Umbraco.Web.Routing
     /// </summary>
     public class PublishedRouter : IPublishedRouter
     {
-        private readonly IWebRoutingSection _webRoutingSection;
         private readonly ContentFinderCollection _contentFinders;
         private readonly IContentLastChanceFinder _contentLastChanceFinder;
         private readonly ServiceContext _services;
@@ -34,7 +33,6 @@ namespace Umbraco.Web.Routing
         /// Initializes a new instance of the <see cref="PublishedRouter"/> class.
         /// </summary>
         public PublishedRouter(
-            IWebRoutingSection webRoutingSection,
             ContentFinderCollection contentFinders,
             IContentLastChanceFinder contentLastChanceFinder,
             IVariationContextAccessor variationContextAccessor,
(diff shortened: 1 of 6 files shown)


src/Umbraco.Web/Routing/PublishedRouter.cs
```diff
@@ -644,74 +641,14 @@ private void FindTemplate(PublishedRequest request)
...
-            // only if the published content is the initial once, else the alternate template
-            // does not apply
-            // + optionally, apply the alternate template on internal redirects
-            var useAltTemplate = request.IsInitialPublishedContent
-                || (_webRoutingSection.InternalRedirectPreservesTemplate && request.IsInternalRedirectPublishedContent);
-            var altTemplate = useAltTemplate
-                ? request.UmbracoContext.HttpContext.Request[Constants.Conventions.Url.AltTemplate]
-                : null;
-
-            if (string.IsNullOrWhiteSpace(altTemplate))
```
bergmania: Here the value `altTemplate` was always `null`

src/Umbraco.Web/Templates/TemplateRenderer.cs
```diff
@@ -172,28 +172,25 @@ private string RenderUmbracoRequestToString(RequestContext requestContext)
...
         /// <summary>
         /// Save all items that we know are used for rendering execution to variables so we can restore after rendering
         /// </summary>
-        private void SaveExistingItems(out PublishedRequest oldPublishedRequest, out object oldAltTemplate)
+        private void SaveExistingItems(out PublishedRequest oldPublishedRequest)
         {
             //Many objects require that these legacy items are in the http context items... before we render this template we need to first
             //save the values in them so that we can re-set them after we render so the rest of the execution works as per normal
             oldPublishedRequest = _umbracoContextAccessor.UmbracoContext.PublishedRequest;
-            oldAltTemplate = _umbracoContextAccessor.UmbracoContext.HttpContext.Items[Core.Constants.Conventions.Url.AltTemplate];
```
bergmania: This value is `null` if never set

Shazwazza:
seems fine to me! :)

PR #7652, opened 2020-02-12
Update PropertyType.cs

benjaminhowarth1:
Please, *please* stop making things internal when it breaks legacy behaviour (such as obtaining properties by tab group, which is precisely what this allows for).

### Prerequisites
- [X] Increasing an access modifier from internal to public will not break any tests.

### Description
Makes PropertyGroupId public.

### DIFF
--- a/src/Umbraco.Core/Models/PropertyType.cs
+++ b/src/Umbraco.Core/Models/PropertyType.cs
@@ -178,7 +178,7 @@ namespace Umbraco.Core.Models
         /// </summary>
         /// <remarks>For generic properties, the value is <c>null</c>.</remarks>
         [DataMember]
-        internal Lazy<int> PropertyGroupId
+        public Lazy<int> PropertyGroupId
         {
             get => _propertyGroupId;
             set => SetPropertyValueAndDetectChanges(value, ref _propertyGroupId, nameof(PropertyGroupId));


poornimanayar:
Hi @benjaminhowarth1 ,

Many thanks for your PR. We shall review it and get back to you.

Poornima

Shazwazza:
This is internal because it's weird and has been marked internal forever. The resaon is because when a property group and types are being added to a content type for persisting the resulting values need to be lazily calculated when the repository first persists the group so that when the properties are persisted the associated group has a valid ID since that will get inserted into the DB first.

Making this public means we need to support this weirdness and things can go wrong if used incorrectly such as creating a new group with properties and then resolving the Lazy ID before persisting would result in a YSOD because a 0 would have been resolved for the group Id instead of the real one. 

We have to live with this weirdness since it would be a bit of work to change this behavior so instead i would propose adding a public get property that will first check if the `PropertyGroupId.IsValueCreated` and if not it will return `default(int)`. 

In the meantime you can resolve properties by groups by iterating over the groups and accessing it's `PropertyTypes` collection.

Shazwazza:
Hrm, i suppose that won't really work unless we change the ctor and populate the group Id via the ctor when it's resolved from the DB. That's when we know it definitely has a valid group Id assigned and is readable from the DB. That is certainly a possibilty but would need a bit more work since we'd need to change every place where a new PropertyType is constructed based on existing data. 

I suppose just making this public is the easiest... hopefully people don't fall into the trap mentioned above though! I'll merge.

PR #7654, opened 2020-02-13
V8: Ensure that the recycle bin doesn't delete content at root

### LINKED ISSUE #7590: Recycle bin removing content on the website
We currently have a multilanguage website that has a problem with the recycle bin. If I click on the recycle bin, it will go to the following url: /umbraco#/content/content/recyclebin?mculture=en-GB and show all the current items in the recycle bin. However if I try to go to /umbraco#/content/content/recyclebin, it lists the items in my root.
image

And when I now delete something from the recycle bin, it will delete it from the content tree and remove it entirely from Umbraco. Meaning I cannot recover it anymore from the actual recycle bin.

I am currently using Umbraco version 8.4.1 and also encountered the same problem on Umbraco 8.5.2

The problem occurs because it uses GetAllowedChildren?contentId=-1 & GetChildren?id=-1, instead of using the actual recycle bin ID.

It also seems like this problem doesn't occur on a single language website.

---
_This item has been added to our backlog [AB#4955](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4955)_

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7590

### Description

#7590 outlines the issue quite well. 

The root cause is that the listview assumes it should show content at root if no ID is passed to the listview in `$routeParams.id` (kudos to @stevemegson for investigating). _That is a really dangerous assumption!_

This PR removes that assumption. As a result the listview remains in a loading state because nothing ever loads. 

image

It is also possible to make the listview appear empty (remove the loading state), but the loading state seems like the better alternative. After all, this is an exception state that _Should Not Happen™_.

I tried getting rid of the layout switcher and search input while the listview is in loading state, but that makes it impossible to search for anything at all in any listview so... that's not gonna happen, at least not in this PR 😆 

The only place I could find that exploits the assumption of `-1` is the listview on the media root dashboard. To counter that I have explicitly defined `-1` for `$routeParams.id` in the dashboard controller, that seems to work just fine.

#### Testing this PR

First and foremost you must have a culture variant setup to test this.

1. Verify that the recycle bin still works for both media and content.
2. Go to `/umbraco#/content/content/recyclebin` (without `mculture` in the querystring) - you may have to do this in a new tab. Verify that the recycle bin either remains in a loading state or lists the correct recycle bin content (there is a timing thing going on, sometimes it actually does list the recycle bin content). _Under no circumstance should it list anything from the content root._
3. Open the media root and verify that it lists folders and files at root as usual.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/dashboard/dashboard.tabs.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/dashboard/dashboard.tabs.controller.js
@@ -220,7 +220,7 @@ function startupLatestEditsController($scope) {
 }
 angular.module("umbraco").controller("Umbraco.Dashboard.StartupLatestEditsController", startupLatestEditsController);
 
-function MediaFolderBrowserDashboardController($rootScope, $scope, $location, contentTypeResource, userService) {
+function MediaFolderBrowserDashboardController($scope, $routeParams, $location, contentTypeResource, userService) {
 
     var currentUser = {};
 
@@ -251,6 +251,8 @@ function MediaFolderBrowserDashboardController($rootScope, $scope, $location, co
                         view: dt.view
                     };
 
+                    // tell the list view to list content at root
+                    $routeParams.id = -1;
             });
 
         } else if (currentUser.startMediaIds.length > 0){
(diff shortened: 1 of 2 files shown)


warrenbuckley:
@kjac this fixes the bug, but is the intention to have it continuously show the loading bar?

kjac:
Aye sure is. As noted in the description, the alternative is to show an empty list which would be misleading.

warrenbuckley:
OK will merge this in for now, but would be nice if it showed the results as you would expect if you were to visit the recycle bin link directly in the browser

warrenbuckley:
@nul800sebastiaan I have cherry-picked this into V8/8.6 as well :)

PR #7656, opened 2020-02-13
V8: Correct cursor for list view search icon

kjac:
### Prerequisites

### Description

When the listview search input field is collapsed it should probably have a pointer cursor because it looks and acts like a button - right now the cursor is the default input field cursor:

listview-search-before

With this PR applied the cursor is a pointer by default and turns into an input field cursor when the search field is expanded:

listview-search-after


poornimanayar:
@kjac 👍

PR #7659, opened 2020-02-13
Review Perplex change for Segments

PerplexDaniel:
@nielslyngsoe 
I ran into an exception when trying to add a new node when it was set to vary by segment only.
This latest commit should fix it. Server-side code would create 0 variants in this case, but it should always create 1 variant (the default / unsegmented variant).

---
_This item has been added to our backlog [AB#5083](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/5083)_

### DIFF
--- a/src/Umbraco.Core/Constants-AppSettings.cs
+++ b/src/Umbraco.Core/Constants-AppSettings.cs
@@ -44,6 +44,21 @@ namespace Umbraco.Core
             public const string Path = "Umbraco.Core.Path";
 
             /// <summary>
+            /// Gets the path to the css directory (/css by default).
+            /// </summary>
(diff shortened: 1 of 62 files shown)


nielslyngsoe:
The code looks good, whether this is the right approach for variants I need someone else to check. But I will merge this into our branch, so you have this available.

PerplexDaniel:
Yeah actually the whole `GetSegments()` method is from my earlier server side code PR so this is just a small addition for the case of creating new nodes that apparently failed. I suppose this stuff isn't covered yet by unit tests otherwise this should have been caught already. But thanks in advance for merging this.

PR #7660, opened 2020-02-13
Fixing when a start node is a content item in a nested list view

Shazwazza:
If you have a user with a start node that is nested within 2 or more list views, the tree will not expand when the user logs in and therefore the user cannot actually navigate to or edit their start node.

## Testing

Create a tree structure like:

* Home
  * List View
    * List View
      * Content

And then create a new user and set their start node to `Content`. When you log in the tree will expand to show `Content`. Be creative, try nesting a start node in a bunch of list views, maybe even have a content -> list view -> content -> list view type scenario, it should all work.

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/IContentTypeRepositoryBase.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/IContentTypeRepositoryBase.cs
@@ -28,6 +28,13 @@ namespace Umbraco.Core.Persistence.Repositories
         bool HasContainerInPath(string contentPath);
 
         /// <summary>
+        /// Gets a value indicating whether there is a list view content item in the path.
+        /// </summary>
+        /// <param name="ids"></param>
+        /// <returns></returns>
+        bool HasContainerInPath(params int[] ids);
+
+        /// <summary>
         /// Returns true or false depending on whether content nodes have been created based on the provided content type id.
         /// </summary>
         bool HasContentNodes(int id);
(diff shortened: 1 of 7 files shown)


warrenbuckley:
Code changes makes sense/read fine 👍 
Off to test this out from your notes :)

warrenbuckley:
## Test Feedback
@Shazwazza a few weird quirks from testing, see my attached video for reference
https://youtu.be/r4wx4BFJ9qs

- The content tree does not sync up correctly with the node selected, it seems to highlight the parent.
- Back arrow in a list view should be removed/disabled if you can not visit its parent.

Shazwazza:
@warrenbuckley that seems odd since i fixed that tree sync issue since it was previously syncing to the parent like in your video (what is up with your audio btw , lol) . Anyways, i'll try to have a look, i had the same tests as you and seemed to be working on my machine.

Shazwazza:
@warrenbuckley right! I've fixed it all up. It was a pretty crazy thing going on indeed. My orig tests were pretty simple but the code didn't cover all scenarios like yours. It should be good now if you can please test. You can have a look at the latest commit to see what was changed:

* Now ensures that all start nodes are used (calculated) since before it was only using start nodes directly assigned to the user which was wrong
* Deals with cases where the start node is a non-list view that exists underneath a start node along with dealing with when there's a child/descendant of this non-list view start node that is a list view

warrenbuckley:
Code changes look/read fine. Off to retest this

warrenbuckley:
Retested and all works as expected - good stuff 💪

PR #7662, opened 2020-02-14
v8: 7647 Can't save content after clearing property with validation error

### LINKED ISSUE #7647: Can't save content after clearing property with validation error
When you save a content item with a non-required field, but with url validation, and fill in a wrong format, the validation error comes up and blocks the save. When you clear the field, the validation error does not go away, and you have to type something new in the field, even though it's non-required.

I am seeing this issue on Umbraco version: 8.5.3

Steps to reproduce:
1) Create new document type with a property thats not required, but has validation (url validation).
2) Create new content of this type, enter a wrong url format, and save the content.
3) Validation message comes up and blocks the save, this is correct behaviour.
4) Now clear the field with the validation message
5) Try to save again, Umbraco still displays the validation message and blocks the save, even though the field is not required and has an empty value.

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #7647 

### Description
This was due to validation not checking for an empty value, which is a valid state for non-mandatory fields.

Changes here ensure that when we set the validation state on a property form, we check that the property is not mandatory and has no value - if both those flags are true, existing errors are cleared and form controls all reset. 

I'm not really across the intricacies of the client-side validation, but from what I can see, the changes fix the problem, and don't impact on existing validation logic. Needs a good sanity check though from someone saner than I.

I've tested by reproducing the scenario from the original bug report. All works as expected - saving with an invalid value displays the validation messages, clearing the value clears the validation.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/validation/valpropertymsg.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/validation/valpropertymsg.directive.js
@@ -24,8 +24,7 @@ function valPropertyMsg(serverValidationManager, localizationService) {
             var hasError = false;
 
             //create properties on our custom scope so we can use it in our template
-            scope.errorMsg = "";
-            
+            scope.errorMsg = "";            
             
             //the property form controller api
             var formCtrl = ctrl[0];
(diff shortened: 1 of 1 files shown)


poornimanayar:
Thank you @nathanwoulfe ! We shall review the work and get back to you if we need anything more.

Regards,
Poornima

nathanwoulfe:
Useful bits are down around line 142 - rest is mainly whitespace/line-length stuff.

poornimanayar:
I am sure I have seen this during my v7 training days, so well done on getting this fixed. On it now!

poornimanayar:
And merged :-) ⭐️ Thanks for the work once again

nathanwoulfe:
No worries at all! Was a tricky one to track down :)

PR #7664, opened 2020-02-14
Netcore: Removed `UrlProvider` from `UmbracoContext`

bergmania:
### Merge https://github.com/umbraco/Umbraco-CMS/pull/7619 first!

- Removed `UrlProvider` from `UmbracoContext`
  - Now you need to inject the `IPublishedUrlProvider`
- Also moved some files that could now be moved into `Umbraco.Abstractions`
- Uses `IRequestCache` instead of `IHttpContextAccessor.HttpContext.Items`


src/Umbraco.Abstractions/Models/PublishedContent/HttpContextVariationContextAccessor.cs
```diff
@@ -7,22 +8,22 @@ namespace Umbraco.Web.Models.PublishedContent
     /// </summary>
     public class HttpContextVariationContextAccessor : IVariationContextAccessor
     {
-        public const string ContextKey = "Umbraco.Web.Models.PublishedContent.DefaultVariationContextAccessor";
-        public readonly IHttpContextAccessor HttpContextAccessor;
+        private readonly IRequestCache _requestCache;
+        private const string _contextKey = "Umbraco.Web.Models.PublishedContent.DefaultVariationContextAccessor";
```
Shazwazza: This is where naming conventions is confusing between r#, VS, etc... and I don't think we have a rule for this in our editorconfig - should constants always be ProperCase? Sames goes for static readonly field names, what convention should we follow? r# generally wants constants and static readonly to be ProperCase

src/Umbraco.Web/Install/InstallSteps/NewInstallStep.cs
```diff
@@ -138,7 +137,7 @@ public override bool RequiresExecution(UserModel model)
 
             // In this one case when it's a brand new install and nothing has been configured, make sure the
             // back office cookie is cleared so there's no old cookies lying around causing problems
-            _http.ExpireCookie(_umbracoSettingsSection.Security.AuthCookieName);
+            _httpContextAccessor.HttpContext.ExpireCookie(_umbracoSettingsSection.Security.AuthCookieName);
```
Shazwazza: Was this meant to use the new ICookieManager.ExpireCookie?

src/Umbraco.Web/Mvc/MinifyJavaScriptResultAttribute.cs
```diff
@@ -24,7 +38,7 @@ public override void OnResultExecuting(ResultExecutingContext filterContext)
             if (filterContext.Result == null) return;
             var jsResult = filterContext.Result as JavaScriptResult;
             if (jsResult == null) return;
-            if (filterContext.HttpContext.IsDebuggingEnabled) return;
+            if (_hostingEnvironment.IsDebugMode) return;
```
Shazwazza: I assume this change is really just pre-emptively knowing that these APIs will be different in netcore MVC?

src/Umbraco.Web/PropertyEditors/ValueConverters/MarkdownEditorValueConverter.cs
```diff
@@ -1,53 +0,0 @@
-﻿using System;
-using System.Web;
-using HeyRed.MarkdownSharp;
-using Umbraco.Core;
-using Umbraco.Core.Models.PublishedContent;
-using Umbraco.Core.PropertyEditors;
-using Umbraco.Web.Composing;
-using Umbraco.Web.Templates;
-
-namespace Umbraco.Web.PropertyEditors.ValueConverters
```
Shazwazza: Was this file supposed to be deleted?

src/Umbraco.Web/Routing/IMediaUrlProvider.cs
```diff
@@ -26,6 +25,6 @@ public interface IMediaUrlProvider
         /// e.g. a cdn url provider will most likely always return an absolute url.</para>
         /// <para>If the provider is unable to provide a url, it returns <c>null</c>.</para>
         /// </remarks>
-        UrlInfo GetMediaUrl(IUmbracoContext umbracoContext, IPublishedContent content, string propertyAlias, UrlMode mode, string culture, Uri current);
```
Shazwazza: Seems like by changing this signature we've invented new circular dependencies?

src/Umbraco.Web/Routing/IUrlProvider.cs
```diff
@@ -24,19 +23,18 @@ public interface IUrlProvider
         /// when no culture is specified, the current culture.</para>
         /// <para>If the provider is unable to provide a url, it should return <c>null</c>.</para>
         /// </remarks>
-        UrlInfo GetUrl(IUmbracoContext umbracoContext, IPublishedContent content, UrlMode mode, string culture, Uri current);
+        UrlInfo GetUrl(IPublishedContent content, UrlMode mode, string culture, Uri current);
```
Shazwazza: Seems like by changing this signature we've invented new circular dependencies and same with the overload?

src/Umbraco.Web/Security/AuthenticationExtensions.cs
```diff
@@ -169,18 +169,6 @@ public static bool RenewUmbracoAuthTicket(this HttpContextBase http)
             return true;
         }
 
-        /// <summary>
-        /// This will force ticket renewal in the OWIN pipeline
-        /// </summary>
-        /// <param name="http"></param>
-        /// <returns></returns>
-        internal static bool RenewUmbracoAuthTicket(this HttpContext http)
```
Shazwazza: I assume we don't use this anymore?

src/Umbraco.Web/Security/WebSecurity.cs
```diff
@@ -74,29 +74,29 @@ private BackOfficeSignInManager SignInManager
...
 
             return TimeSpan.FromMinutes(_globalSettings.TimeOutInMinutes).TotalSeconds;
         }
 
         [Obsolete("This needs to be removed, ASP.NET Identity should always be used for this operation, this is currently only used in the installer which needs to be updated")]
         public void ClearCurrentLogin()
         {
-            _httpContext.UmbracoLogout();
-            _httpContext.GetOwinContext().Authentication.SignOut(
+            _httpContextAccessor.HttpContext.UmbracoLogout();
```
Shazwazza: The one part about using httpcontextaccessor and umbracocontextaccessor is that the result might be null and we are no doing null checks where necessary but we should be. This could be encapsulated in a property getter with a null check. There's probably a few places where these things need to be fixed up.

src/Umbraco.Abstractions/UmbracoContextReference.cs
```diff
@@ -20,9 +20,8 @@ public class UmbracoContextReference : IDisposable //fixme - should we inherit f
         /// <summary>
         /// Initializes a new instance of the <see cref="UmbracoContextReference"/> class.
         /// </summary>
-        internal UmbracoContextReference(IUmbracoContext umbracoContext, bool isRoot, IUmbracoContextAccessor umbracoContextAccessor)
+        internal UmbracoContextReference(bool isRoot, IUmbracoContextAccessor umbracoContextAccessor)
```
Shazwazza: This change i'm just unsure of due to how UmbracoContextReference and IUmbracoContextFactory works. This reference should return the literal reference of IUmbracoContext not the result of the accessor since there's no assurance that the accessor is returning the correct reference. Thoughts?

src/Umbraco.Abstractions/Models/PublishedContent/HttpContextVariationContextAccessor.cs
bergmania: Looks like we can add 
```
dotnet_naming_symbols.private_fields.required_modifiers = abstract,async,readonly,static # all except const
```

src/Umbraco.Web/Mvc/MinifyJavaScriptResultAttribute.cs
bergmania: I know these will change, I just try to align and abstract our direct usage of `HttpContext`

src/Umbraco.Web/PropertyEditors/ValueConverters/MarkdownEditorValueConverter.cs
bergmania: Yes, at some point this was duplicated, and also lives in Infrastructure

src/Umbraco.Web/Install/InstallSteps/NewInstallStep.cs
bergmania: Yep, good catch

src/Umbraco.Web/Security/AuthenticationExtensions.cs
bergmania: It is still used in the  `AuthenticateCurrentRequest` method

src/Umbraco.Abstractions/UmbracoContextReference.cs
bergmania: I can change it back to inject the `IUmbracoContext`..

src/Umbraco.Web/Security/WebSecurity.cs
bergmania: I inserted a null check in the property `AspNetHttpContextAccessor.HttpContext`.

src/Umbraco.Web/Routing/IUrlProvider.cs
bergmania: Not directly this - The cycle is between these file:
```
DefaultMediaUrlProvider
- PropertyEditorCollection
-- DataEditorCollection
--- GridPropertyEditor
---- HtmlImageSourceParser
----- UrlProvider
------ MediaUrlProviderCollection
------- DefaultMediaUrlProvider
```

src/Umbraco.Web/Routing/IMediaUrlProvider.cs
bergmania: This is the cycle..  
```
DefaultMediaUrlProvider
- PropertyEditorCollection
-- DataEditorCollection
--- GridPropertyEditor
---- HtmlImageSourceParser
----- UrlProvider
------ MediaUrlProviderCollection
------- DefaultMediaUrlProvider
```

I can't find a good way to break it currently.  I think `DefaultMediaUrlProvider` - where it is currently solved by having a lazy dependency is the correct place to break the cycle.

bergmania:
I have implemented most of your comments..
I'm in doubt how to break the cycle in a better manner then what I already did by using Lazy.

bergmania:
The real issue with the cycle is IMO that `GridPropertyEditor` needs to have dependencies to everything required by `GridPropertyValueEditor`, which also needs all dependencies of `RichTextPropertyValueEditor` and `MediaPickerPropertyValueEditor`...

src/Umbraco.Abstractions/CompositionExtensions.cs
```diff
@@ -22,6 +23,15 @@ public static ComponentCollectionBuilder Components(this Composition composition
         public static DashboardCollectionBuilder Dashboards(this Composition composition)
             => composition.WithCollectionBuilder<DashboardCollectionBuilder>();
 
+
+        /// <summary>
+        /// Gets the content finders collection builder.
+        /// </summary>
+        /// <param name="composition">The composition.</param>
+        /// <returns></returns>
+        public static DataEditorWithMediaPathCollectionBuilder DataEditorsWithMediaPath(this Composition composition)
```
Shazwazza: Will think of a different name :) I'll push something shortly and let you know

src/Umbraco.Abstractions/UmbracoContextReference.cs
```diff
@@ -51,7 +49,7 @@ public void Dispose()
             if (IsRoot)
             {
                 UmbracoContext.Dispose();
-                _umbracoContextAccessor.UmbracoContext = null;
+                UmbracoContext = null;
```
Shazwazza: I think this also changes the logic of how this works, when the reference is disposed it is supposed to clear the context on the accessor. See the code comments on the class: 

```
/// <summary>
/// Represents a reference to an <see cref="UmbracoContext"/> instance.
/// </summary>
/// <remarks>
/// <para>A reference points to an <see cref="UmbracoContext"/> and it may own it (when it
/// is a root reference) or just reference it. A reference must be disposed after it has
/// been used. Disposing does nothing if the reference is not a root reference. Otherwise,
/// it disposes the <see cref="UmbracoContext"/> and clears the
/// <see cref="IUmbracoContextAccessor"/>.</para>
/// </remarks>
```

I'll revert

src/Umbraco.Web/Net/AspNetHttpContextAccessor.cs
```diff
@@ -10,7 +10,13 @@ public HttpContextBase HttpContext
             get
             {
                 var httpContext = System.Web.HttpContext.Current;
-                return httpContext is null ? null : new HttpContextWrapper(httpContext);
+
+                if (httpContext is null)
```
Shazwazza: I wonder if this goes against the 'accessor' pattern? I think most accessors will just return null if there isn't one which also provides a way to determine if there is an object at the current state. By throwing here it means you can't check if there isn't one without a try/catch. I know that means we'd have to put null checks everywhere throughout the code.... else does it make sense to have an ext method for this like `httpContextAccessor.GetRequired()` which has this null check?

.editorconfig
```diff
@@ -25,16 +25,22 @@ dotnet_naming_rule.private_members_with_underscore.severity = suggestion
 
 dotnet_naming_symbols.private_fields.applicable_kinds           = field
 dotnet_naming_symbols.private_fields.applicable_accessibilities = private
+dotnet_naming_symbols.private_fields.required_modifiers = abstract,async,readonly,static # all except const
```
Shazwazza: This breaks how field names are generated. With this change when creating new fields based on ctor params it will be `this.someProperty` instead of `_someProperty`

.editorconfig
```diff
@@ -25,16 +25,22 @@ dotnet_naming_rule.private_members_with_underscore.severity = suggestion
...
 csharp_style_var_for_built_in_types = true:suggestion
 csharp_style_var_when_type_is_apparent = true:suggestion
 csharp_style_var_elsewhere = true:suggestion
-csharp_prefer_braces = false : none
+csharp_prefer_braces = false : none
 
-[*.{js,less}]
-trim_trailing_whitespace = false
+[*.{js,less}]
+trim_trailing_whitespace = false
```
Shazwazza: I know this hasn't changed but i have no idea why this is false?

.editorconfig
```diff
@@ -25,16 +25,22 @@ dotnet_naming_rule.private_members_with_underscore.severity = suggestion
 
 dotnet_naming_symbols.private_fields.applicable_kinds           = field
 dotnet_naming_symbols.private_fields.applicable_accessibilities = private
+# dotnet_naming_symbols.private_fields.required_modifiers = abstract,async,readonly,static # all except const
```
Shazwazza: Commented this out because it causes fields to not be prefixed with underscores and creating new fields from ctor params results in `this.myProperty` instead of `_myProperty`

src/Umbraco.Abstractions/CompositionExtensions.cs
Shazwazza: I have pushed a new rev: 13a56eb08549f196eeb8039b656c24cf91c452aa which renames and refactors this a little to simplify this and decouple it from property editors since the logic was still relying on property editors having this interface which isn't what we want.

.editorconfig
bergmania: Damn, worked in Rider - But I guess that's because the other things are default conventions

.editorconfig
bergmania: 🤷

src/Umbraco.Web/Net/AspNetHttpContextAccessor.cs
bergmania: I like the idea of an extension method.. I will introduce that

bergmania:
I'm in doubt about this one in `WebInitailComposer.cs`. Should this HttpContext throw if it is `null`

```
  composition.Register<HttpContextBase>(factory => factory.GetInstance<IHttpContextAccessor>().HttpContext, Lifetime.Request);
```

src/Umbraco.Web/HttpContextAccessorExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Umbraco.Web
+{
+    public static class HttpContextAccessorExtensions
+    {
+        public static HttpContextBase GetRequiredHttpContext(this IHttpContextAccessor httpContextAccessor)
+        {
+            var httpContext = httpContextAccessor.HttpContext;
+
+            if(httpContext is null) throw new IOException("HttpContext is null");
```
Shazwazza: How come it's IOException?

bergmania: Dooh'  Should have been InvalidOperationException..  I must have chosen the wrong one, when the suggestions came..  Do you agree on InvalidOperationException? - Then I fix in dev

Shazwazza: Yep sounds good

Issue #7668, opened 2020-02-15
.NetCore: Remove direct database queries from services and controllers


AndyButland:
Umbraco services shouldn't be doing direct queries of the database, rather they should go through their respective repository for database access.  This is a dependency we'd need to remove in order to migrate to a "clean architecture", with all models and services (business logic) in "core" and persistence concerns in "infrastructure".

There look to be a couple of services where this occurs:

- `KeyValueService`
- `UserService`

Task is to remove these concerns, migrating to an existing or new repository layer.

kashfshah:
Any other services still need this treatment?

AndyButland:
I'm not aware of any, no.  Looks like you closed this @bergmania, and re-opened, but I think it can actually be closed now.  I believe that at least any issues of SQL statements in the services were removed from the linked PR.

bergmania:
Hi @kashfshah  and @AndyButland .
At least `EntityService` are still building SQL and using `Scope.Database` directly.

bergmania:
Also `LogController` makes SQL

umbrabot:
Hi @AndyButland,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

ageroh:
Hi,
Still need help?
I willing to take over one this one

bergmania:
Hi @ageroh...

Yes, this is still unresolved and it would be super nice if you would take a look. The issue will not be closed by a single PR.

I'm aware of multiple usages of `ISqlContext` in controllers.  These are all Fat controllers, where business logic should be handled in services and database queries should be handled in repositories. 

Consider `LogController` as an example. All logic could be refactored to be handled by a new LogService. 

Let me know if you have any questions :)

ageroh:
I will have a look and let you know! Happy to help!

ageroh:
@bergmania could you check the below:

https://github.com/umbraco/Umbraco-CMS/pull/8968

lars-erik:
After the unicorner with @bergmania on Unicore, this issue has piqued my interest due to the possible long run outcome.  
[...]

PR #7669, opened 2020-02-15
.NetCore: Remove direct database queries from services

AndyButland:
Issue references:: #7668 

In this PR I've removed some direct database requests being made in Umbraco services, and moved those concerns up to a repository layer.

There look to be a couple of services where this occurs:

- `KeyValueService`
- `UserService`

The `KeyValueService` I've created a new `IKeyValueRepository`, along with implementation and creation of appropriate models.  Then I've added methods to this for getting and setting key values, and referenced them from the service.

For `UserService` I've added one additional method to the existing (and renamed another for clarity), and used this to move the database access out of the service and up to the repository layer.


src/Umbraco.Infrastructure/Persistence/Repositories/Implement/KeyValueRepository.cs
```diff
@@ -0,0 +1,233 @@
...
+
+namespace Umbraco.Core.Persistence.Repositories.Implement
+{
+    internal class KeyValueRepository : NPocoRepositoryBase<string, IKeyValue>, IKeyValueRepository
+    {
+        public KeyValueRepository(IScopeAccessor scopeAccessor, ILogger logger)
+            : base(scopeAccessor, AppCaches.NoCache, logger)
+        { }
+
+        public void Initialize()
```
Shazwazza: I'm not so sure about this one, i don't really think that `Initialize` makes sense on this interface since it's not really clear what it actually does. Since this is specific to migrations i think this code needs to live outside of this repository and the call made elsewhere.

src/Umbraco.Abstractions/Persistence/Repositories/IKeyValueRepository.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Umbraco.Infrastructure.Persistence.Repositories
+{
+    public interface IKeyValueRepository
```
Shazwazza: This doesn't really follow any of the current repository patterns that we have. All repsositories implement at least `IRepository`, then we have `IReadRepository`, `IWriteRepository`, and `IQueryRepository` which are all implemented by `IReadWriteQueryRepository`. Whether it makes sense here to implement all of them I'm unsure but we should try to be consistent. Most repositories inherit from `RepositoryBase` which has the normal plumbing wired up for Scopes, etc... but requires your entity to be of type `IEntity`. You could even return a consistent GUID Key for the entity based on the `string` key using the string extension `ToGuid` method that we have.  We should also have a task to update this DB table to have both a Created and Updated date time so we can be more consistent with the EntityBase model. Perhaps go as far as to add a real GUID column to the DB table so it's actually get/set instead of just get.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/KeyValueRepository.cs
```diff
@@ -0,0 +1,233 @@
...
+using Umbraco.Core.Migrations;
+using Umbraco.Core.Models;
+using Umbraco.Core.Persistence.Dtos;
+using Umbraco.Core.Persistence.Querying;
+using Umbraco.Core.Scoping;
+using Umbraco.Infrastructure.Persistence.Repositories;
+
+namespace Umbraco.Core.Persistence.Repositories.Implement
+{
+    internal class KeyValueRepository : NPocoRepositoryBase<string, IKeyValue>, IKeyValueRepository
```
Shazwazza: See other comment about repository design, if we are consistent with the `IReadRepository` and `IWriteRepository`, etc... then the repository methods are already there to deal with. Having GetValue, SetValue aren't consistent with what we have.

AndyButland:
Thanks for the feedback.  Please see what you think of the latest.  I've made some amends to align more with the existing repository patterns.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/KeyValueRepository.cs
AndyButland: Have moved into a new component: `IKeyValueServiceInitialization` (and implementation).

src/Umbraco.Abstractions/Persistence/Repositories/IKeyValueRepository.cs
AndyButland: Yes, point noted.  I've updated the `IKeyValueRepository` to use these base interfaces, and make the amends that follow from that.  It's slightly clunky in the sense that the "identifier" for this entity is the key of the key value, but `Key` is already defined on the `IEntity`.  But follows existing patterns better as you noted.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/KeyValueRepository.cs
AndyButland: Have removed these now, they are still in the service layer and implemented using methods that act on the `IKeyValue/KeyValue` entity class now.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/KeyValueRepository.cs
```diff
@@ -0,0 +1,133 @@
...
+    {
+        public KeyValueRepository(IScopeAccessor scopeAccessor, ILogger logger)
+            : base(scopeAccessor, AppCaches.NoCache, logger)
+        { }
+
+        /// <summary>
+        /// Gets a value directly from the database, no scope, nothing.
+        /// </summary>
+        /// <remarks>Used by <see cref="Runtime.CoreRuntime"/> to determine the runtime state.</remarks>
+        internal static string GetValue(IUmbracoDatabase database, string key)
```
Shazwazza: I think this can be moved out to an extension method on IUmbracoDatabase. There's just something funky about this being an internal class and an internal method and that we don't want to really ever call a Repository class directly. Or maybe this can remain a static method here to return the `Sql` object used to do the query, that way all SQL production is still in a repository class. But the calling of this method from `RuntimeState` could then be `CurrentMigrationState = database.GetFromKeyValueTable(stateValueKey);` (or similar) and that method could call this method which could be changed to `GetKeyValueSql`. That way we can have the `database.GetFromKeyValueTable` as a public method in case it's needed elsewhere or by people.

src/Umbraco.Core/Models/KeyValue.cs
```diff
@@ -0,0 +1,32 @@
...
+using Umbraco.Core.Models.Entities;
+
+namespace Umbraco.Core.Models
+{
+    /// <summary>
+    /// Implements <see cref="IKeyValue"/>.
+    /// </summary>
+    [Serializable]
+    [DataContract(IsReference = true)]
+    public class KeyValue : EntityBase, IKeyValue
```
Shazwazza: This should explicitly implement `IEntity` and then explicitly implement `IEntity.HasIdentity` to return it's own logic of whether this entity has an identity. (explicit implementation is the same as overriding a base class). Then we don't have to make the `Save` method in RepositoryBase virtual and we don't have to override that method in the KeyValueRepository.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/RepositoryBaseOfTIdTEntity.cs
```diff
@@ -132,7 +132,7 @@ protected virtual RepositoryCachePolicyOptions DefaultOptions
         /// </summary>
         /// <remarks>This method is backed by an <see cref="IAppPolicyCache"/> cache</remarks>
         /// <param name="entity"></param>
-        public void Save(TEntity entity)
+        public virtual void Save(TEntity entity)
```
Shazwazza: Not needed, see previous comment about explicitly implementing HasIdentity.

src/Umbraco.Infrastructure/Migrations/Custom/IKeyValueServiceInitialization.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Umbraco.Core.Persistence;
+
+namespace Umbraco.Infrastructure.Migrations.Custom
+{
+    public interface IKeyValueServiceInitialization
```
Shazwazza: We actually don't need this (or the implementation) at all, this was a hack/workaround for v7 -> v8 migrations

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/KeyValueRepository.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        protected override string GetBaseWhereClause()
+        {
+            return Constants.DatabaseSchema.Tables.KeyValue + ".key = @id";
+        }
+
+        protected override IEnumerable<string> GetDeleteClauses()
+        {
+            return new List<string>();
```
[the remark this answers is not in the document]
AndyButland: Have amended to this.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/KeyValueRepository.cs
```diff
@@ -0,0 +1,133 @@
...
+        protected override IEnumerable<IKeyValue> PerformGetAll(params string[] ids)
+        {
+            var sql = GetBaseQuery(false).WhereIn<KeyValueDto>(x => x.Key, ids);
+            var dtos = Database.Fetch<KeyValueDto>(sql);
+            return dtos.WhereNotNull().Select(Map);
+        }
+
+        protected override IEnumerable<IKeyValue> PerformGetByQuery(IQuery<IKeyValue> query)
+        {
+            throw new NotImplementedException();
```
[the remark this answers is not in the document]
AndyButland: Done.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/KeyValueRepository.cs
AndyButland: Agree that's better.  Have implemented.

src/Umbraco.Infrastructure/Migrations/Custom/IKeyValueServiceInitialization.cs
AndyButland: Have removed this and all uses of it.

src/Umbraco.Core/Models/KeyValue.cs
AndyButland: Had some trouble with this one.  The problem being that the key value "entity" doesn't actually have an identity like an integer (that will be zero if the record doesn't already exist) - there's no "id" primary key in `umbracoKeyValue`, the key itself is the primary key.

So when saving an instance of a key value pair, the only way to see if it's a new or updated entity, is to attempt to retrieve it by key and see if it exists.  If so, it's an update, if not an insert.

And that retrieval needs to be in the repository values - the `KeyValue` model class doesn't know from it's own properties if it's a new or updated value, and hence can't implement `HasIdentity`.

So have left this as is for now, pending any further thoughts or feedback...

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/RepositoryBaseOfTIdTEntity.cs
AndyButland: Left in place for now, see response to comment above.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/KeyValueRepository.cs
```diff
@@ -0,0 +1,133 @@
...
+                PersistNewItem(entity);
+            else
+                PersistUpdatedItem(entity);
+        }
+
+        #endregion
+
+        #region Overrides of NPocoRepositoryBase<string, IKeyValue>
+
+        protected override Guid NodeObjectTypeId => throw new NotImplementedException();
```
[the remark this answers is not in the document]
AndyButland: Done.

src/Umbraco.Core/Models/KeyValue.cs
Shazwazza: Ah yes i see what you mean because even if its not persisted you of course have to assign a key and we cannot just know if the entity exists if that key is null or not, makes sense! Oh well, would have been nice :) we'll keep what you have done

AndyButland:
OK - conflicts resolved.

Shazwazza:
Thanks!!

PR #7676, opened 2020-02-17
V8/feature/ab4550 segments ui variant picker

PerplexDaniel:
Hi @nielslyngsoe,

I added some additional commits to fix a few things. The C# changes will probably have to reviewed by someone else at some point, but perhaps that is better to do when we feel it's good enough and can be checked. I now simply fix stuff I run into and sometimes it has to be changed again later (like the `GetSegments()` method). 

New changes now:
* `GetSegments()` updated (again) to always return the default segment (`null`). I ran into a case where there was no data yet for the default segment but there was data for other segments and in that case the default segment was not returned from the `GetSegments()` method, meaning in the Umbraco frontend you could no longer edit the default segment at all.
* Null reference exception fix related to the assumption a culture would be present
* Small optimization in the `requestSplitView` function to prevent looping through all variants after variant has been found
* Experimental fix to prevent segments to be deleted each save when their value is `null`. 
  * I noticed whenever a Save or Publish is performed Umbraco does something like "delete all property data of version X" followed by "insert all relevant property data of version X", and decides that property data with a `null` value is not relevant so is skipped. However, for segments this means the whole segment is removed because we only have the property data to know which segments exists. So when I want to create a new empty segment I just write an empty value to some property but Umbraco was removing old segments when saving a new one (i.e., if I create segment `A`, then `B`, then `C` all with `null` or `""` values, only segment `C` would be present after the last save/publish).

---
_This item has been added to our backlog [AB#5131](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/5131)_

### DIFF
--- a/src/Umbraco.Core/Constants-AppSettings.cs
+++ b/src/Umbraco.Core/Constants-AppSettings.cs
@@ -44,6 +44,21 @@ namespace Umbraco.Core
             public const string Path = "Umbraco.Core.Path";
 
             /// <summary>
+            /// Gets the path to the css directory (/css by default).
+            /// </summary>
+            public const string CssPath = "umbracoCssPath";
+
+            /// <summary>
+            /// Gets the path to the scripts directory (/scripts by default).
+            /// </summary>
+            public const string ScriptsPath = "umbracoScriptsPath";
+
+            /// <summary>
+            /// Gets the path to  media directory (/media by default).
+            /// </summary>
+            public const string MediaPath = "umbracoMediaPath";
+
+            /// <summary>
             /// The reserved urls from web.config.
             /// </summary>
             public const string ReservedUrls = "Umbraco.Core.ReservedUrls";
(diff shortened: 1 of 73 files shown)


nielslyngsoe:
Hi @PerplexDaniel 

Looking at your changes in src/Umbraco.Web/Editors/ContentController.cs, I start thinking that does not handle Segments since the code seems to only deal with Culture. Is that correctly understood? I'm trying to conclude whether we need to revisit this part later?

PerplexDaniel:
Yes, looking at that code we probably also need some `segmentErrors` to include those. Or perhaps just transform this `cultureErrors` into `variantErrors` and include errors on segmented variants. I only fixed a crash there and did not change any of the functionality.

nielslyngsoe:
Thanks for the clarification, I have noted that.

PR #7677, opened 2020-02-17
ContentVariantDisplay: add DisplayName property and use in Umbraco in…

PerplexDaniel:
@nielslyngsoe Seperate branch this time, probably easier to check. This is a small change which adds a `DisplayName` property to the `ContentVariantDisplay` class so we can control the variant display name on the server, and in AngularJS do not have to bind to functions which is slower and I know Shannon does not like that either. 

In addition, developers can now alter the `DisplayName` property if desired by using the existing Umbraco event `EditorModelEventManager.SendingContentModel` which is also useful for the uMarketingSuite.

FYI, because this made the `contentvariantutilities.service.js` entirely obsolete (because it was empty) I removed it. 

---
_This item has been added to our backlog [AB#5130](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/5130)_


nielslyngsoe:
Still unsure whether we like to have this DisplayName located here. But I think it makes good sense to have it originating from BackEnd. It could be useful for validation messages as well.

I will merge it for now, and we will see if we need to change later.

PR #7680, opened 2020-02-18
Update CONTRIBUTING.md

PerplexDaniel:
Summary link and actual header were different, link failed because of that too. Changed header to the summary + link content. Could also be done the other was around if preffered (update summary + link to match header).

Also fixes link syntax (removes space that breaks link functionality)


PerplexDaniel:
Hi @poornimanayar,

Sure no problem, that's indeed what I put in the description of the PR -- it could be changed the other way around. Just updated the summary and reverted header change.

poornimanayar:
And merged :-) ⭐️

PR #7684, opened 2020-02-18
Change web service invocation to new Our rest endpoints

elit0451:
The purpose of this PR it to replace the invocation of a SOAP web service for Our with 2 new rest APIs.

In order to test this PR you need to be familiar with how to run Our locally and be able to access `upgradechecker` database.

---

## Testing
After you make sure you can run Our locally, go ahead and replace the [RestApiInstallUrl](https://github.com/umbraco/Umbraco-CMS/blob/a058703004c748997f56e027b299228938ebde33/src/Umbraco.Core/Persistence/Repositories/Implement/InstallationRepository.cs#L15) with:
 `private const string RestApiInstallUrl = "http://localhost:24292/umbraco/api/Installation/Install";`

and [RestApiUpgradeChecklUrl](https://github.com/umbraco/Umbraco-CMS/blob/a058703004c748997f56e027b299228938ebde33/src/Umbraco.Core/Persistence/Repositories/Implement/UpgradeCheckRepository.cs#L13) with:
`private const string RestApiUpgradeChecklUrl = "http://localhost:24292/umbraco/api/UpgradeCheck/CheckUpgrade";`
</br>

### Install endpoint
❗️ You need to test this on a fresh Umbraco install!

You need to be monitoring _[upgradechecker].[dbo].[installs]_ table throughout the installation process. 
- When you get the page asking you for your name, email and password, if you check the database, you will have a record inserted in there. 
- **Make sure `[completed]` and `[installEnd]` fields** are `0` and `Null` respectively
- When the installation process completes ⬆️ the previously mentioned fields should have different values
</br>

### Upgrade Check endpoint

This time it will be the _[upgradechecker].[dbo].[pings]_ table that you would be checking!
- Log out of Umbraco and clear all the cache and cookies or at least the `UMB_UPDCHK` one.
- Log in again and check the records in the db.
- You should see a new record with your Umbraco version and the time you logged in `[lastPing]` field.

---
❕ Make sure that Umbraco doesn't break in Our is not running locally anymore ❕


bergmania:
Hi @elit0451.

I pushed a few changes:
- I postfixed `async` the method names with Async to align with the codebase and conventions
- I removed the explicit json handling, and used the features from the framework
- I fixed a bug (not introduced by this PR) with the install cookie (See `InstallHelper`)
- I introduced a private DTO class in `UpgradeCheckRepository` instead of using an anonymous type

PR #7689, opened 2020-02-19
add pointer-events:none on disabled checkbox/radio

nathanwoulfe:
Hey @emmaburstow, this is the re-do of the one I destroyed. Same change, same file, same result.


nul800sebastiaan:
Wonderful, thank you @nathanwoulfe ! 👍

PR #7690, opened 2020-02-19
Fix regression from valpropertymsg directive changes

nathanwoulfe:
Fixes regression where validation object doesn't always exist on the property, depending on context - when adding an editor to a content type, the validation logic runs, but the validation property on the editor doesn't exist.

Changes here add an explicit check for `currentProperty.validation` before trying to access `.mandatory`

This relates to the changes from #7662 - ping @poornimanayar since she's across those changes 😄 

### Prerequisites

To test, edit a content type, adding a new property editor, should be no JS errors logged into the console. Validation should clear as expected when editing the property value on a content item.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/validation/valpropertymsg.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/validation/valpropertymsg.directive.js
@@ -39,7 +39,9 @@ function valPropertyMsg(serverValidationManager, localizationService) {
             scope.currentProperty = currentProperty;
 
             var currentCulture = currentProperty.culture;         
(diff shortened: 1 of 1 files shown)


nathanwoulfe:
Try adding a text box to an existing type, before applying this change - will throw a js error when the editor config overlay opens, because `validation` isn't a property on the property editor. If there are other places where the property editor model doesn't include the `validation` object, this fix will manage those too

poornimanayar:
And merged your work ⭐️ ! Thanks as always :-)

nathanwoulfe:
No worries - happened to have Dev tools open and noticed the error. I don't think it actually broke anything since the model value never changed so was never validated, but still needed to be fixed

PR #7691, opened 2020-02-20
NetCore: Move more files

bergmania:
- Moved `Umbraco.Examine` into `Umbraco.Infrastructure`
- Moved more files
  - Features
    - Introduced `IUmbracoFeature` that our `UmbracoApiControllerBase ` implements, instead of replying directly on the `UmbracoApiControllerBase`
  - Mappings
  -  Editor stuff (not controllers)
  - Install stuff
     - changed `InstallStatusTracker` from static to instance
- Uses Cookie Manager in more files
  - Expanded functionality
- Uses config project instead of `WebConfigurationManager`.
- FilePermissionHelper 
  - Moved some static helper methods out of `FilePermissionHelper into
- FilePermissionDirectoryHelper`
  - Changed FilePermissionHelper from static to instance and include interface
- Introduced `MenuItemCollectionFactory` to still have an easy way to create `MenuItemCollection` even that we need `ActionCollection` injected.


src/Umbraco.Core/Features/IUmbracoFeature.cs
```diff
@@ -0,0 +1,7 @@
+namespace Umbraco.Web.Features
+{
+    public interface IUmbracoFeature
```
Shazwazza: Could do with some code comments here explaining what this is

src/Umbraco.Core/Install/FilePermissionDirectoryHelper.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.IO;
+using Umbraco.Core.IO;
+
+namespace Umbraco.Web.Install
+{
+    public class FilePermissionDirectoryHelper
```
Shazwazza: Try to avoid "Helper" class names, maybe this can just be IOHelper extension methods? depending on if CreateRandomFileName is used outside of this class.

src/Umbraco.Core/UmbracoEvents.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+using Umbraco.Web.Routing;
+
+namespace Umbraco.Core
+{
+    public interface IUmbracoRouteEventSender
```
Shazwazza: Code comments? need to explain what this is

src/Umbraco.Core/Trees/IMenuItemCollectionFactory.cs
```diff
@@ -0,0 +1,9 @@
+using Umbraco.Web.Models.Trees;
+
+namespace Umbraco.Web.Trees
+{
+    public interface IMenuItemCollectionFactory
```
Shazwazza: Code comments? need to explain what this is

src/Umbraco.Core/Sync/IBatchedDatabaseServerMessenger.cs
```diff
@@ -0,0 +1,7 @@
+namespace Umbraco.Core.Sync
+{
+    public interface IBatchedDatabaseServerMessenger : IServerMessenger
```
Shazwazza: Code comments? need to explain what this is ... though this might have been empty before, if so not to worry

src/Umbraco.Web/Install/Controllers/InstallApiController.cs
```diff
@@ -216,11 +218,11 @@ internal bool StepRequiresExecution(InstallSetupStep step, JToken instruction)
         }
 
         // executes the step
-        internal async Task<InstallSetupResult> ExecuteStepAsync(InstallSetupStep step, JToken instruction)
+        internal async Task<InstallSetupResult> ExecuteStepAsync(InstallSetupStep step, object instruction)
         {
             using (_proflog.TraceDuration<InstallApiController>($"Executing installation step: '{step.Name}'.", "Step completed"))
             {
-                var model = instruction?.ToObject(step.StepType);
+                var model = Convert.ChangeType(instruction, step.StepType);
```
Shazwazza: Does this always work? Our extension method TryConvertTo is a superior conversion tool than .net's built in ones like this

src/Umbraco.Web/Trees/FileSystemTreeController.cs
```diff
@@ -4,14 +4,46 @@
...
 using Umbraco.Web.Models.Trees;
+using Umbraco.Web.Routing;
 
 namespace Umbraco.Web.Trees
 {
     public abstract class FileSystemTreeController : TreeController
     {
+        private readonly IMenuItemCollectionFactory _menuItemCollectionFactory;
+
+        protected FileSystemTreeController()
```
Shazwazza: how come we have this ctor?

src/Umbraco.Web/Trees/FileSystemTreeController.cs
```diff
@@ -4,14 +4,46 @@
...
 using Umbraco.Web.Actions;
+using Umbraco.Web.Composing;
 using Umbraco.Web.Models.Trees;
+using Umbraco.Web.Routing;
 
 namespace Umbraco.Web.Trees
 {
     public abstract class FileSystemTreeController : TreeController
     {
+        private readonly IMenuItemCollectionFactory _menuItemCollectionFactory;
```
Shazwazza: Should all TreeControllers just expose this as a protected property?

src/Umbraco.Core/Trees/MenuItemList.cs
```diff
@@ -44,7 +47,7 @@ public MenuItem Add<T>(ILocalizedTextService textService, bool hasSeparator = fa
         private MenuItem CreateMenuItem<T>(ILocalizedTextService textService, bool hasSeparator = false, bool opensDialog = false)
             where T : IAction
         {
-            var item = Current.Actions.GetAction<T>();
+            var item = _actionCollection.GetAction<T>();
```
Shazwazza: I'm assuming this is the underlying reason why we need the MenuItemCollectionFactory?

src/Umbraco.Core/Install/FilePermissionDirectoryHelper.cs
bergmania: I moved both methods into extensions of `IIOHelper`. Even that only one of them uses `IIOHelper`. The reason to move the other method also (`CreateRandomFileName()`) is because I think it logically belongs to `IOHelper`, but we don't need to implement this method for every implementation.

src/Umbraco.Core/UmbracoEvents.cs
bergmania: I could remove this interface..

src/Umbraco.Web/Trees/FileSystemTreeController.cs
bergmania: The controller is abstract, so this ctor is introduced to avoid all inheritors to implement the full ctor. - This is the same way all abstract controllers constructors are handled.

src/Umbraco.Web/Trees/FileSystemTreeController.cs
bergmania: I change that - It is currently not used, but then we follow our pattern.

Issue #5892, opened 2019-07-15
v8 GetUrlAndAnchors doesn't seem to work for variant property values


Shazwazza:
When using the link picker and using variant content, the anchor values don't get returned for variant properties. 

The problem seems to be in the call to `ContentService.GetAnchorValuesFromRTEs(id, culture);` and within that a call is made to `var value = contentProperty.GetValue(culture)?.ToString();` which by default the culture will be `"*"`, however no value will get returned for this because `PropertyType.SupportsVariation(culture, segment)` will fail because the `wildcards=false` parameter is always false.

bjarnef:
I guess it also should have a null check here https://github.com/umbraco/Umbraco-CMS/blob/ade9bb73246caf25a7073f2b9e5262641a201863/src/Umbraco.Core/Services/ContentServiceExtensions.cs#L22 before accessing properties from `content` although it probably wouldn't happen since it pass in the node id from the selected node in linkpicker.

Shazwazza:
For v8, we'll need to pass up the explicit culture for this i think, passing in * for a property value doesn't really work, plus we need to consider what variant we are on, splitview, etc....

marcoteodoro:
@Shazwazza there is also another related issue. if you have nested content blocks, lets say one RTE in nested content one, and another RTE in nested content 2 and you try to create an anchor link from this RTE's nothing is listed. I think the issue is that you are just looking for property types of RTE, but if it's an IElement i think we must query sub-properties of RTE as well to get the anchor links. 
https://github.com/umbraco/Umbraco-CMS/blob/ade9bb73246caf25a7073f2b9e5262641a201863/src/Umbraco.Core/Services/ContentServiceExtensions.cs#L26

marcoteodoro:
@Shazwazza not sure how should look at the pull request but I've ask your review on https://github.com/umbraco/Umbraco-CMS/pull/7695 this will fix this bug.

umbrabot:
Hiya @Shazwazza,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
[...]

PR #7695, opened 2020-02-20
Allow Link Picker to get links from RTE's inside nested content

marcoteodoro:
### Prerequisites

- Nested content with an RTE
- Doctype with a property type that allows multiple nested content items.
- Document type and properties allow variants 
- At least to cultures 
Steps to test this pull request.

- Add a content node with two nested content items. 
- Add one anchor link in one of the RTE's
- Save or Save and Publish your changes
- Add a link on the other RTE. 
- When the link editor dialog opens, click on the node where you have both RTE's and look at the top-right text box where you can append query string parameters, you should see the arrow icon on mouse hover, on click you should be able to see the recently created anchor link for your specific culture. 

note: If you don't set a culture this will still work and return the anchor links that are not assigned to a specific language. 

this fixes https://github.com/umbraco/Umbraco-CMS/issues/5892

Please check the [gif file](https://drive.google.com/file/d/1PPBUO_fR3Jaq8coVX4cD83UTQzKw42VA/view?usp=sharing) as this will make things much clear for you when testing the PR.

### DIFF
--- a/src/Umbraco.Core/Services/ContentServiceExtensions.cs
+++ b/src/Umbraco.Core/Services/ContentServiceExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
@@ -21,6 +21,7 @@ namespace Umbraco.Core.Services
             var result = new List<string>();
             var content = contentService.GetById(id);
 
+
             foreach (var contentProperty in content.Properties)
             {
                 if (contentProperty.PropertyType.PropertyEditorAlias.InvariantEquals(Constants.PropertyEditors.Aliases.TinyMce))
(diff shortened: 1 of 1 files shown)


emmaburstow:
Hey @marcoteodoro 

Thanks for the work here and also, a great big congrats on your first PR to this repo! If you let us know what your username is on Our.Umbraco.com we will add a new contrib badge to your profile once we've merged. 

In the meantime, thanks for the testing steps above! We'll let you know if we have any questions.

Emma

marcoteodoro:
Great @emmaburstow! I think my username is marcoteodoro@double.pt, 
image

Hope this code is accepted and the pull request merged soon as this is an important feature for me. 

Many Thanks

nul800sebastiaan:
@marcoteodoro Just to set some expectations: it will be a while before we can release a version with this fix in it, if this is important then, for now, it would be best for you to create a custom PropertyValueConverter that can do what you need it to.

The main concern I have with this PR at the moment is that it's very specific for Nested Content and I'm not sure this code should be making all kinds of workarounds for different property editors. Did you test this with the Grid as well? We might need to fix the Grid with something as well, in case it doesn't yet work there. If it does work there, then maybe we need to look into how the Grid does it currently.

Note: I know very little or nothing about this part of the codebase, so I might be way off :-)

marcoteodoro:
@nul800sebastiaan i'm not sure the PropertyValueConverter will work in this case. The issue for me is not when rendering CMS property data on templates but when you try to create anchor links on umbraco. Unless i missunderstood your sugestion (if so please explain how that will help) i think we need to fix the source code. 

I didn't test the grid but I will prepare a demo project to check that as well. thanks. meanwhile, if you can explain how can I use the property value converter would be great. 

Thanks,
Marco

nul800sebastiaan:
I'm not sure either, but I was under the impression that `ConvertSourceToIntermediate` would convert what is in the database to what the backoffice will load.. I might be way off though! 

https://our.umbraco.com/documentation/extending/property-editors/value-converters#convertsourcetointermediateipublishedelement-owner-ipublishedpropertytype-propertytype-object-source-bool-preview

marcoteodoro:
@nul800sebastiaan  I will try that. I guess I need to replace the link property editor?

PR #7700, opened 2020-02-21
Add missing methods "GetChildren(Guid id)" & "HasChildren(Guid id)" to "IContentTypeBaseService"

Jeavon:
This method exists (see link below) in the ContentTypeService however it cannot be used as it is missing from the interface, hence adding so it can be used.

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Services/Implement/ContentTypeServiceBaseOfTRepositoryTItemTService.cs#L255

Test code:
`IEnumerable<IContentType> contentTypes = Services.ContentTypeService.GetChildren(Guid.Parse("ca4249ed-2b23-4337-b522-63cabe5587d1"));
`


bjarnef:
@Jeavon should the other method with `int` id parameter in this interface has an overload with `Guid` id as well, e.g. `HasChildren(int id)` and `HasChildren(Guid id)`?

Jeavon:
@bjarnef yes, I think it should, will add

emmaburstow:
Thanks for the work, Jeavon! Will test and merge 

Em

Issue #7688, opened 2020-02-19
Umbraco.UploadField based media item causes an error when using URL based custom file systems


whatwasmyusername:
Unable to re-save a file media item after the file has been uploaded to a remote file system.

## Umbraco version

I am seeing this issue on Umbraco version: 8.5.3

Reproduction
------------

### Bug summary

The `UploadFileTypeValidator` assumes a physical file system based file which causes and error when attempting to save the media item after initial upload.

### Specifics

When using file system providers such as [Azure Blob Storage](https://our.umbraco.com/documentation/Extending/FileSystemProviders/Azure-Blob-Storage/) or [Umbraco S3 Provider](https://github.com/ElijahGlover/Umbraco-S3-Provider), the value saved to the database is an URL rather than a path to a file in the media storage folder.

The method to determine the file extension: `new FileInfo(fileName)` causes a "System.NotSupportedException: The given path's format is not supported." error when the value being evaluated is a URL instead of a file system path.

This uncaught exception prevents changes to the media item from being saved.

### Steps to reproduce

* Setup a custom file system provider for media which saves its paths as URLs (two examples given above).
* Upload a non-image file (the default Image media type uses the Image Cropper property editor, which does not use this validator in the same way as FileUploadPropertyEditor).
* Navigate to the media item in the back office.
* Click the "Save" button for the media item.

### Expected result

* Media item is saved

### Actual result

* Media item is not saved
* Dialog is displayed with message from the server
* Save button is left in "busy" state.

An abbreviated error message displayed is as follows:

````An error occured
The given path's format is not supported.

Exception Details
System.NotSupportedException: The given path's format is not supported.

Stacktrace
at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
   at System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPath)
   at System.IO.FileInfo.Init(String fileName, Boolean checkHost)
   at Umbraco.Web.PropertyEditors.UploadFileTypeValidator.IsValidFileExtension(String fileName) in d:\a\1\s\src\Umbraco.Web\PropertyEditors\UploadFileTypeValidator.cs:line 51
   at Umbraco.Web.PropertyEditors.UploadFileTypeValidator.<Validate>d__0.MoveNext() in 

[Stacktrace has been trimmed]
````

[...]

PR #7701, opened 2020-02-21
Fix for exception throw when validating URLs

whatwasmyusername:
### Prerequisites

This fixes https://github.com/umbraco/Umbraco-CMS/issues/7688

### Description

* Replaced the use of FileInfo with Path.GetExtension to support URL-based filenames
* This can be tested by:
  * Setting up an Azure or S3 media file system
  * Uploading a media item with a Umbraco.FileUpload property (i.e. the default File type)
  * Re-save the media item


emmaburstow:
Hello hello!

Thanks for the PR and congrats on the first PR to this repo! We'll take a look and when we merge you'll get a nice new contrib badge on your our.umbraco.com profile so let us know where to find you.

We'll be in touch if we need anything,

Emma

nul800sebastiaan:
I would love to hear what @kimschurmann thinks as well, since they changed this in PR https://github.com/umbraco/Umbraco-CMS/issues/7688 and I would like to make sure that we're not breaking that scenario again with this update. I can't see why it would break though. 🤔

kimschurmann:
Hi @nul800sebastiaan 

I currently do not have the time to look at it, but my colleague @kjac probably knows what is going on and could perhabs be talked into looking at the issue?

rasmusjp:
I don't think the fix breaks anything as `Path.GetExtension(filename)` and `new FileInfo(filename).Extension` essentially does the same, the different is that `new FileInfo(filename)` tries to demand some file permissions which it can't because the filename isn't a valid path.

As far as I can tell this is only a problem with the UmbracoFileSystemProviders.Azure if you disable the virtual path provider by adding the `AzureBlobFileSystem.DisableVirtualPathProvider=true` app setting, then it stores the full Url to the media item instead of the relative one.

stevetemple:
Should do the same thing, except the Path.GetExtension is more lightweight, doesn't do anything with the actual file system, just string manipulation so is just a better choice for this anyway. 

A unit test around this, would prove that this works identically to the old version, happy to knock something up if that's desired.

nul800sebastiaan:
@rasmusjp Oh, fantastic feedback, in that case neither of these methods would work! 

So it would be better to get the extension through a method like this: https://stackoverflow.com/a/52748299/5018

As has been pointed out to me as well, this also includes a few unit tests, which is great.

So it looks like the solution should be: to get the file extension of a URL or file name, feed it into this method as a string. Then we don't need to worry about filesystem operations. Additionally, this method has unit tests which we can use and expand upon. 

Could you make those changes @whatwasmyusername ? I would add this to `Umbraco.Core.StringExtensions` so it can be used easily. And there's a bunch of unit tests for that class already as well so they can be expanded upon.

nul800sebastiaan:
@rasmusjp Oh, fantastic feedback, in that case neither of these methods would work! 

So it would be better to get the extension through a method like this: https://stackoverflow.com/a/52748299/5018 - but without the `FromURL`, it's getting the extension from a string, whether it's a URL or not.

As has been pointed out to me as well, this also includes a few unit tests, which is great.

So it looks like the solution should be: to get the file extension of a URL or file name, feed it into this method as a string. Then we don't need to worry about filesystem operations. Additionally, this method has unit tests which we can use and expand upon. 

Could you make those changes @whatwasmyusername ? I would add this to `Umbraco.Core.StringExtensions` so it can be used easily. And there's a bunch of unit tests for that class already as well so they can be expanded upon.

whatwasmyusername:
Sounds great @nul800sebastiaan .  I'll go the route of adding the string extension.

whatwasmyusername:
Thanks for the recommendations.  

I've added a few extra test cases based on your suggestion and took a look at using your regex.  The regex you threw out there didn't work for strings without question marks in them, and it assumed letters and numbers only in the extension, so I made a couple of changes to get it to pass all of the tests.  

I didn't see any performance changes (all tests were already executing in the 4-10ms range for me, though admittedly that doesn't really constitue extensive testing around memory usage or big pieces of data), but If you really think that's a better route to go, I will push up a new commit.

kjac:
@whatwasmyusername I do believe it's a better route to go. My concern with the current implementation is that someone will use the extension method to process larger sets of data and thus put an unnecessary strain on the CPU.

whatwasmyusername:
Thanks for moving this along!  Here's [my profile](https://our.umbraco.com/members/whatwasmyusername/).

nul800sebastiaan:
@whatwasmyusername Sweet, badge added! 👍 Thanks again!

PR #7703, opened 2020-02-24
Update editor.service.js ngdoc for templateEditor fn

nathanwoulfe:
Fix ngdoc description for templateEditor fn => description previously referred to user group picker, should be `template editor`. Quick n easy.


nathanwoulfe:
Taking on the big challenges this week... 😃

PR #7705, opened 2020-02-24
NetCore: Executable in .NET Core

bergmania:
## Notes
- Added an executable in .NET Core 3.1
  -  Named `Umbraco.Web.Ui.NetCore` for now
- Added Umbraco.Web.Website (.NET Core 3.1)
  - Added extension methods as proxies for more specific commands.
    - Added ImageSharp.Web.
 - Trimed duplicate stuff from IHostingEnvironment
 - Implemented the ASP.NET Core versions of the following interfaces:
   - `IBackOfficeInfo`
   - `IHostingEnvironment`
   - `IIpResolver`
   - `IMarchal`
   - `ISessionIdResolver`
   - `IUmbracoApplicationLifetime`
   - `IUserAgentProvider`
 - Prepared a AspNetCoreHttpContextAccessor
  - Not completely ready, as we can't really abstract HttpContextBase (especially the Request)
## Test Notes:
- Run the new  `Umbraco.Web.Ui.NetCore` project
- Add an image file into the `wwwroot/Media` folder and request it from the website
  - https://localhost:5001/Media/{filename}
- Test that ImageSharp manipulates the file if you add query string params
  - E.g. https://localhost:5001/Media/{filename}?width=100


src/Umbraco.Web.BackOffice/Umbraco.Web.BackOffice.csproj
```diff
@@ -1,11 +1,13 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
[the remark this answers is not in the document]
Shazwazza: Unless your plan is to have 3x executable websites but 2 of them (Website and Backoffice) are also used like class libraries when hosting both parts on the same website? In which case things could get a bit weird since we won't know where to put assets/views. 
I think we should probably have netcore3.1 class libs for the back office and frontend and then also have web executables for both the backoffice, frontend, and combined. 
For now, this is a super easy change, i'll change these over to class libs

src/Umbraco.Web.BackOffice/AspNetCore/AspNetCoreUserAgentProvider.cs
```diff
@@ -0,0 +1,20 @@
+using Microsoft.AspNetCore.Http;
+using Umbraco.Net;
+
+namespace Umbraco.Web.BackOffice.AspNetCore
+{
+    public class AspNetCoreUserAgentProvider : IUserAgentProvider
```
Shazwazza: Any reason this is public and some of the other implementations are internal?

src/Umbraco.Web.Website/AspNetCore/UmbracoWebsiteApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+        public static IApplicationBuilder UseUmbracoWebsite(this IApplicationBuilder app)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+
+            // Important we handle image manipulations before the static files, otherwise the querystring is just ignored.
+            app.UseImageSharp();
```
Shazwazza: We'll need this in the back office too... i know this is early days and thsi is POC work :) just mentioning it. We'll probably have some other common ext methods like UseUmbracoWebCommon. Also on that note, we'll probably end up with another netcore 3.1 class library for common things between the back office and front-end.

Issue #7709, opened 2020-02-24
Reintroduce `ConfigureForDisplay` for `DataValueEditor`?


leekelleher:
When developing a custom Property Editor in v8, if the display is configurable by its Data Type, (e.g. hiding the label, or potentially changing the view), the only place to do this in the value editor's `Configuration` setter method.

Here's an example of how Nested Content does this for hiding the property's label: https://github.com/umbraco/Umbraco-CMS/blob/release-8.5.4/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs#L81-L91

_IMO, I don't think "display logic" should be done inside a property's setter method._

In v7, there was a method available called [`ConfigureForDisplay`](https://github.com/umbraco/Umbraco-CMS/blob/release-7.15.3/src/Umbraco.Core/PropertyEditors/PropertyValueEditor.cs#L70-L74), that could be used to config the value editor.

This is how Nested Content did it back in v7, using the `ConfigureForDisplay` method: https://github.com/umbraco/Umbraco-CMS/blob/release-7.15.3/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs#L101-L113

---

I have 2 questions...

1. **Was there a reason why `ConfigureForDisplay` was removed in v8?**
  _(I've been digging through the git history - [found where it initially was commented out](https://github.com/umbraco/Umbraco-CMS/blob/cfd2ba3b5a2c6a381aab3babe603b9bcaa70f61b/src/Umbraco.Core/PropertyEditors/DataValueEditor.cs#L85-L97) - alas no answers about why)_

2. **Would it be up for consideration to reintroduce `ConfigureForDisplay`?** 😃 

My reason for asking is that in Contentment I have a couple of property editors where a different view can be selected, e.g. say swapping out a dropdown list with a radiobutton list. _(I'm currently doing this via the `Configuration` setter method, it's not pretty code!)_

abjerner:
Hi @leekelleher,

The `DataEditor` class has a virtual `CreateValueEditor` method, which could then be overriden like:

```csharp
protected override IDataValueEditor CreateValueEditor()  {

    var editor = new DataValueEditor(Attribute)
    {
        HideLabel = true,
        View = "/My/Custom/View.html"
    };

    return editor;

}
```

It let's you set both the `HideLabel` and `View` properties, but from what I can tell, it doesn't have access to the configuration.

[...]

PR #7710, opened 2020-02-24
Making DataEditor.GetValueEditor method virtual

abjerner:
Based on #7709, I think it makes good sense to make the `DataEditor.GetValueEditor` method virtual, so it can be overridden by classes inheriting from the `DataEditor` class.

Overriding the method could then be something like:

```csharp
public override IDataValueEditor GetValueEditor(object configuration)
{

    var editor = base.GetValueEditor(configuration);

    if (editor is DataValueEditor valueEditor && configuration is MyConfiguration config)
    {
        valueEditor.HideLabel = config.HideLabel;
    }

    return editor;

}
```

For testing that the overriden method is actually hit, I made this custom property editor:

```csharp
using Umbraco.Core.Logging;
using Umbraco.Core.PropertyEditors;
using Umbraco.Web.PropertyEditors;

namespace Umbraco.Web.Test{

    [DataEditor("Test", EditorType.PropertyValue, "Test", "mediapicker", ValueType = ValueTypes.String)]
    public class TestEditor : DataEditor {

        public TestEditor(ILogger logger) : base(logger)  { }

        protected override IConfigurationEditor CreateConfigurationEditor() => new MediaPickerConfigurationEditor();

        public override IDataValueEditor GetValueEditor(object configuration) {

            var editor = base.GetValueEditor(configuration);

            if (editor is DataValueEditor valueEditor && configuration is MediaPickerConfiguration config)  {
                valueEditor.HideLabel = true;
                valueEditor.View = "nope.html";
            }

            return editor;

        }

    }

}
```

It uses the view and configuration of the media picker, and adding it to a document type, the changes made in the `GetValueEditor` method should be visible to the user.

### DIFF
--- a/src/Umbraco.Core/PropertyEditors/DataEditor.cs
+++ b/src/Umbraco.Core/PropertyEditors/DataEditor.cs
@@ -105,7 +105,7 @@ namespace Umbraco.Core.PropertyEditors
         /// <para>Technically, it could be cached by datatype but let's keep things
         /// simple enough for now.</para>
         /// </remarks>
-        public IDataValueEditor GetValueEditor(object configuration)
+        public virtual IDataValueEditor GetValueEditor(object configuration)
         {
             // if an explicit value editor has been set (by the manifest parser)
             // then return it, and ignore the configuration, which is going to be


nul800sebastiaan:
After seeing Shannon's reply, it seems that implementing the interface is the preferred way to go, so we'll go with that for now, I'll close this one. Thanks though! 👍

Shazwazza:
You can still make it virtual, would be easier for devs and less code to write

nul800sebastiaan:
Super! Done!

PR #7712, opened 2020-02-25
Netcore: Move macros to Abstractions and Infrastructure

elit0451:
This PR allows for migrating the rest of the PropertyEditors, that today have references to `IMacroRenderer` and `MacroTagParser`.
</br>

Additionally:
- The usage of WebForms controls is eliminated from **MacroContent.cs**;
- The enum for **MacroTypes** is removed as we only have 1 type and at the moment we don't plan to add other types;
- Removing references to `pageElements` from **MacroRenderer.cs** - this is an old legacy which comes from `PublishedContentHashtableConverter` (which is also removed);

Test notes:
- Before checking out to the PR branch create a Macro in the back-office and insert it in a RTE property of a Content node;
- Check out to the beforementioned branch and verify that the Macro is still rendering and that we still allow for modifications to the Macros


src/Umbraco.Core/Macros/MacroContent.cs
```diff
@@ -0,0 +1,20 @@
...
+    public class MacroContent
+    {
+        // gets or sets the text content
+        public string Text { get; set; }
+
+        // gets or sets the date the content was generated
+        public DateTime Date { get; set; } = DateTime.Now;
+
+        // a value indicating whether the content is empty
+        public bool IsEmpty => Text == null;
```
bergmania:
```suggestion
        public bool IsEmpty => Text is null;
```

test for `is null` is faster because you can't override the `is` operator

bergmania:
Seems to work as expected.. Good work.

@Shazwazza, will you have a look also

PR #7722, opened 2020-02-26
NetCore: Moved install steps + dashboards

bergmania:
- Moved dashboards to Core
- Introduced IUserAgentProvider, instead of using HttpContext directly
- Introduced IUmbracoApplicationLifetime, instead of using UmbracoApplication directly
  - Moved/refactored UmbracoApplication.Restart
  - Uses a property to store IsRestarting info - Works as long as the IUmbracoApplicationLifetime is registered as a singleton
- Introduced IMachineKeyConfig instead of using WebConfigurationManager directly
- Move information about dbProvider to the SqlSyntax
- Moved install steps


src/Umbraco.Infrastructure/Intall/InstallHelper.cs
```diff
@@ -22,32 +22,38 @@ public sealed class InstallHelper
...
+        public InstallHelper(DatabaseBuilder databaseBuilder,
             ILogger logger,
             IGlobalSettings globalSettings,
             IUmbracoVersion umbracoVersion,
             IConnectionStrings connectionStrings,
             IInstallationService installationService,
-            ICookieManager cookieManager)
+            ICookieManager cookieManager,
+            IUserAgentProvider userAgentProvider,
+            Lazy<ISqlContext> sqlContext,
```
Shazwazza: Introducing a new circular ref? aha, i see it's in order to get rid of the `Current` usage. Perhaps we should add TODO notes about these things to see if we can un-circular ref these.

src/Umbraco.Infrastructure/Intall/InstallHelper.cs
```diff
@@ -107,19 +111,7 @@ public async Task InstallStatus(bool isCompleted, string errorMsg)
...
-            //
-            // doing it differently
-            var syntax = sqlContext.SqlSyntax;
-            if (syntax is SqlCeSyntaxProvider)
-                dbProvider = "SqlServerCE";
-            else if (syntax is SqlServerSyntaxProvider)
-                dbProvider = (syntax as SqlServerSyntaxProvider).ServerVersion.IsAzure ? "SqlAzure" : "SqlServer";
-
-            return dbProvider;
+            return sqlContext.SqlSyntax.DbProvider;
```
Shazwazza: Maybe we don't even need the method `GetDbProviderString` anymore and instead just use SqlSyntax to resolve this info?

src/Umbraco.Infrastructure/Intall/InstallSteps/ConfigureMachineKey.cs
```diff
@@ -27,10 +29,11 @@ public ConfigureMachineKey(IIOHelper ioHelper)
         /// Don't display the view or execute if a machine key already exists
         /// </summary>
         /// <returns></returns>
-        private static bool HasMachineKey()
+        private bool HasMachineKey()
         {
-            var section = (MachineKeySection) WebConfigurationManager.GetSection("system.web/machineKey");
-            return section.ElementInformation.Source != null;
+            return _machineKeyConfig.HasMachineKey;
+            // var section = (MachineKeySection) WebConfigurationManager.GetSection("system.web/machineKey");
```
Shazwazza: can remove commented out code

src/Umbraco.Infrastructure/Intall/InstallSteps/SetUmbracoVersionStep.cs
```diff
@@ -1,44 +1,31 @@
...
-            _userService = userService;
             _umbracoVersion = umbracoVersion;
-            _ioHelper = ioHelper;
         }
 
         public override Task<InstallSetupResult> ExecuteAsync(object model)
         {
-            var security = new WebSecurity(_httpContextAccessor, _userService, _globalSettings, _ioHelper);
-
+            var security = _umbracoContextAccessor.UmbracoContext.Security;
```
Shazwazza: null checks? Should we have a GetRequiredContext ext method like we have for the http context accessor?

src/Umbraco.Infrastructure/Intall/InstallHelper.cs
bergmania: In general, I do not introduce new circular refs, but the circles become more explicit by the move into constructor injection :)

I changed this case to inject IUmbracoDatabaseFactory instead, and the call SqlContext on that dependency. I'm not a big fan of that solution either, But as long as all of our interfaces and classes has such big responsibility, the cicular refs are hard to avoid.

PR #7725, opened 2020-02-26
7720 - make internal UserService events public

### LINKED ISSUE #7720: Internal events in UserService
Umbraco.Core/Serviced/Implement/UserService raises a bunch of events when users and/or groups are modified, however SavingUserGroup and SavedUserGroup are both internal. 

Wondering if there's a reason for this? If there's not, I'll make em public.

Example use case: in Plumber, I want to add a caching layer around approval groups to avoid db load. I can cache my native groups, but am also adding a feature to allow using Umbraco groups in the workflow, so need to know when these are changed so that I can invalidate cache...

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #7720 

### Description

Changes [Saving|Saved]UserGroup events in UserService from internal to public

Looks like these were always intended to be public, but got mixed up in the 7=>8 changes. Have also had to make UserGroupWithUsers public as it's referenced by the event so must be equally accessible.

Have verified by adding an event handler in an existing Composer in Umbraco.Web, events are raised, parameters are correct.

### DIFF
--- a/src/Umbraco.Core/Events/UserGroupWithUsers.cs
+++ b/src/Umbraco.Core/Events/UserGroupWithUsers.cs
@@ -3,7 +3,7 @@ using Umbraco.Core.Models.Membership;
 
 namespace Umbraco.Core.Events
 {
-    internal class UserGroupWithUsers
+    public class UserGroupWithUsers
     {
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Thanks @nathanwoulfe! Good stuff! 👍

nathanwoulfe:
👍👍

PR #7729, opened 2020-02-27
7728 - show elipsis only when > length

### LINKED ISSUE #7728: v8 TruncateByWords always adds elipisis, even if word count is less than truncate count
When I use the the TruncateByWords method I expect an ellipsis to be added only when the original word count exceeds the truncate count I pass in.

At the moment the ellipsis is always shown, so if I want to truncate to 10 words but my text is 4 words long then it still applies an ellipsis to the end.

## Umbraco version

I am seeing this issue on Umbraco version: 8.5.4 

Reproduction
------------

Pass an html text to the HtmlHelper.TruncateByWords() method eg:
```
var text = <p>this is four words</p> 

@Html.TruncateByWords(text, 10) 
```

### Expected result

<p>this is four words</p>

### Actual result

<p>this is four words...</p>

I had a look at the logic:

Can change to on line 246 of HtmlStringUtilities.cs
```
if (!lengthReached)
{
    if (currentTextLength == length)
    {
        // if the last character added was the first of a two character unicode pair, add the second character
        if (Char.IsHighSurrogate((char)ic))
        {
            var lowSurrogate = tr.Read();
            outputtw.Write((char)lowSurrogate);
        }

    }
    // only add elipsis if current length greater than original length
    if (currentTextLength > length)
    {
        if (addElipsis)
        {
            outputtw.Write(hellip);
        }
        lengthReached = true;
    }
}  
                      
```

I can add a PR if this looks fine.

mjbarlow:
Fixed the elipsis issue, only adds an elipisis if the text length exceeds the length that is being truncated too.

### Prerequisites

If there's an existing issue for this PR then this fixes #7728 

### Description
Updates the Truncate method, only adds the elipsis if the length of the string passed in exceeds the truncate amount.  If they are equal then no elipsis is added.

### DIFF
--- a/src/Umbraco.Web/HtmlStringUtilities.cs
+++ b/src/Umbraco.Web/HtmlStringUtilities.cs
@@ -227,21 +227,27 @@ namespace Umbraco.Web
                                         }
                                     }
 
-                                    if (!lengthReached && currentTextLength >= length)
+                                    if (!lengthReached)
                                     {
-                                        // if the last character added was the first of a two character unicode pair, add the second character
-                                        if (Char.IsHighSurrogate((char)ic))
+                                        if (currentTextLength == length)
                                         {
-                                            var lowSurrogate = tr.Read();
-                                            outputtw.Write((char)lowSurrogate);
-                                        }
+                                            // if the last character added was the first of a two character unicode pair, add the second character
+                                            if (char.IsHighSurrogate((char)ic))
+                                            {
+                                                var lowSurrogate = tr.Read();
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Sorry for the late reply @mjbarlow - we'll have a look at this soon! 👍

mjbarlow:
No worries, it's a minor issue! :)

nul800sebastiaan:
Sweet, good fix @mjbarlow! 👍 

And congrats on your first contribution to this repository as well, top job! ⭐️

PR #7731, opened 2020-02-27
Some decoupling from underscorejs

warrenbuckley:
A LAB PR to tidy up the dependencies used in the Umbraco backoffice JS code & try to remove the usage of underscorejs as most of this stuff is now available natively in JS

This is a WIP/Draft until the following list & all references is gone

**Reference/guide**
https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

- [x] _.first
- [x] _.last
- [ ] _.find _[127 results in 55 files]_
- [ ] _.each [121 _results in 53 files]_
- [ ] _.map _[59 results in 32 files]_
- [ ] _.filter _[37 results in 17 files]_
- [ ] _.contains _[34 results in 12 files]_
- [ ] _.reject _[22 results in 13 files]_
- [ ] _.forEach _[13 results, in 6 files]_
- [ ] _.debounce _[10 results in 10 files]_
- [ ] _.reduce _[6 results in 4 files]_
- [ ] _.values _[6 results in 6 files]_
- [ ] _.union _[5 results in 5 files]_
- [ ] _.isEqual _[5 results in 3 files]_
- [ ] _.where _[4 results in 4 files]_
- [ ] _.flatten _{4 results in 2 files]_

### DIFF
--- a/src/Umbraco.Web.UI.Client/gulp/tasks/dependencies.js
+++ b/src/Umbraco.Web.UI.Client/gulp/tasks/dependencies.js
@@ -232,11 +232,6 @@ function dependencies() {
             "name": "typeahead.js",
             "src":  ["./node_modules/typeahead.js/dist/typeahead.bundle.min.js"],
             "base": "./node_modules/typeahead.js/dist"
-        },
-        {
-            "name": "underscore",
-            "src":  ["node_modules/underscore/underscore-min.js"],
-            "base": "./node_modules/underscore"
         }
     ];
 
--- a/src/Umbraco.Web.UI.Client/gulp/tasks/dependencies.js
+++ b/src/Umbraco.Web.UI.Client/gulp/tasks/dependencies.js
@@ -105,7 +105,7 @@ function dependencies() {
             "name": "angular-messages",
             "src":  ["./node_modules/angular-messages/angular-messages.js"],
             "base": "./node_modules/angular-messages"
-        },        
+        },
         {
             "name": "angular-mocks",
             "src":  ["./node_modules/angular-mocks/angular-mocks.js"],
@@ -252,7 +252,7 @@ function dependencies() {
 
     //Copies all static assets into /root / assets folder
     //css, fonts and image files
-    
+
     var assetsTask = gulp.src(config.sources.globs.assets, { allowEmpty: true });
     assetsTask = assetsTask.pipe(imagemin([
         imagemin.gifsicle({interlaced: true}),
@@ -265,9 +265,9 @@ function dependencies() {
             ]
         })
     ]));
-    
+
     assetsTask = assetsTask.pipe(gulp.dest(config.root + config.targets.assets));
-    
+
     stream.add(assetsTask);
 
     // Copies all the less files related to the preview into their folder
--- a/src/Umbraco.Web.UI.Client/gulp/tasks/js.js
+++ b/src/Umbraco.Web.UI.Client/gulp/tasks/js.js
@@ -19,12 +19,12 @@ function js() {
         gulp.src(config.sources.globs.js).pipe( gulp.dest(config.root + config.targets.js) )
     );
 
-    for(const group in config.sources.js){
-        const groupItem = config.sources.js[group];
+    Object.keys(config.sources.js).forEach(key => {
+        const groupItem = config.sources.js[key];
         stream.add(processJs(groupItem.files, groupItem.out));
-    }
+    });
 
-     return stream;
+    return stream;
 };
 
 module.exports = { js: js };
--- a/src/Umbraco.Web.UI.Client/gulp/tasks/js.js
+++ b/src/Umbraco.Web.UI.Client/gulp/tasks/js.js
@@ -3,7 +3,6 @@
 var config = require('../config');
 var gulp = require('gulp');
 
-var _ = require('lodash');
 var MergeStream = require('merge-stream');
 
 var processJs = require('../util/processJs');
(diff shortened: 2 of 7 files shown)


src/Umbraco.Web.UI.Client/gulp/tasks/js.js
```diff
@@ -20,11 +19,10 @@ function js() {
...
 
-    _.forEach(config.sources.js, function (group) {
-        stream.add(
-            processJs(group.files, group.out)
-        );
-    });
+    for(const group in config.sources.js){
+        const groupItem = config.sources.js[group];
+        stream.add(processJs(groupItem.files, groupItem.out));
+    }
```
mattbrailsford: Not sure how true this still is, but it used to be the case at least that `for ... in` looping on an objects keys was not the best option, due to the fact it's prototype chain might include props you don't want to loop. Instead, requesting it's keys and forEeaching should be better.

    Object.keys(config.sources.js).forEach(key => {
         const groupItem = config.sources.js[key];
         stream.add(processJs(groupItem.files, groupItem.out));
    });

src/Umbraco.Web.UI.Client/src/views/propertyeditors/slider/slider.controller.js
```diff
@@ -41,11 +41,11 @@
         }
         // convert to array
         $scope.sliderValue = $scope.model.value ? $scope.model.value.split(',') : null;
-        
+
         // don't render values with decimal places if the step increment in a whole number
         var stepDecimalPlaces = $scope.model.config.step % 1 == 0
             ? 0
-            : _.last($scope.model.config.step.toString().replace(",", ".").split(".")).length;
+            : $scope.model.config.step.toString().replace(",", ".").split(".")[$scope.model.config.step.toString().replace(",", ".").split(".") - 1].length;
```
mattbrailsford: You should probably only do `$scope.model.config.step.toString().replace(",", ".").split(".")` once and store the value so you don't calculate it twice. 

Or if you want to get fancy, you could do `$scope.model.config.step.toString().replace(",", ".").split(".").slice(-1)[0]`

src/Umbraco.Web.UI.Client/gulp/tasks/js.js
warrenbuckley: The power of the HIVE mind

src/Umbraco.Web.UI.Client/src/views/propertyeditors/slider/slider.controller.js
warrenbuckley: Yeh I was trying to not change too much and be quite literal with my first pass/conversions, but good idea :)

src/Umbraco.Web.UI.Client/gulp/tasks/dependencies.js
```diff
@@ -232,11 +232,6 @@ function dependencies() {
             "name": "typeahead.js",
             "src":  ["./node_modules/typeahead.js/dist/typeahead.bundle.min.js"],
             "base": "./node_modules/typeahead.js/dist"
-        },
-        {
-            "name": "underscore",
```
dawoe: Does this mean underscore js will not be shipped with the backoffice anymore ? This will be massive breaking change for package developers. So this needs to be made clear way in advance. And have some kind of upgrade instructions for package developers available

mattbrailsford: I believe the point of this PR is to START to decouple Umbraco from its dependency, as many functions are now natively supported. Once Umbraco Core no longer needs it, a deprication strategy can be formed. On Twitter yesterday it was suggest to create a deprication layer that proxies the new methods outputting warnings to the console log to upgrade. Ultimately the intent is not to just “rip it out” but to just make a start to gradually remove these unnecessary dependencies.

warrenbuckley: Yeh I was maybe a bit too aggressive on the removal - so this will get rolled back @dawoe & @mattbrailsford after finding out package devs use it more than I thought.

So goal of this will be to remove our own usage of underscore.js

nathanwoulfe:
This relates nicely to the changes I just added in #7738 - if there are cases where the underscore dependency can't be readily worked-around or removed, abstract it away in utility.js. I definitely agree with removing usages from the Umbraco codebase, but continuing to ship the library.

I have two usages in there already, for `copy()` and `isEqual()`. I don't see any issue with maintaining the dependency if it's used where absolutely necessary, and ideally behind an abstraction - there's a huge difference between carrying a library dependency and a tight coupling to a framework 😄

nathanwoulfe:
I've started pulling out _.each usages - 11 files done, 40 to go...

bjarnef:
I have noticed the function `_.omit` which is used a few places [6 results in 3 files].
@warrenbuckley can you add this to the list in case someone want to look at this?

bjarnef:
I have also noticed `_.throttle` is used in a few files [2 results in 2 files].
Replaced with vanilla javascript function in https://github.com/umbraco/Umbraco-CMS/pull/8523

bjarnef:
I have replaced underscore `_.debounce` function with ulility function in https://github.com/umbraco/Umbraco-CMS/pull/8524

nul800sebastiaan:
Just FYI: we (HQ) need to talk about a strategy here, the goal should not be to write our own library (Utilities.js) and I think we don't want to get rid of ALL of underscore, just the lines that can be easily replaced without having to create a new Utility function for it.

We definitely don't want to maintain a full library that does exactly what underscore does, there's a tool for that, it's called underscore and it's pretty good! 😅

bjarnef:
@nul800sebastiaan sure underscore is great, but today much of this can be done with vanilla javascript or possible with a small function, so maybe we don't need to include underscore in future releases (Umbraco v9?). Furthermore browser support is much better than 5 years ago, where I guess the functions in underscore have support for legacy browsers, which Umbraco 8 doesn't support anyway.

PR #7738, opened 2020-02-29
adds utility.js as facade to generic javascript utility functions

nathanwoulfe:
### Prerequisites

This is the first step required for the AngularJs divorce proceedings...

### Description

Changes here introduce `utility.js` to either replace AngularJs ng-function stuff, or to provide a facade to underscore functions (which in turn replace AngularJs ng-function stuff). Ideally at some point in the future the underscore dependency can be removed too, but for now it's required to manage some of the more complex bits. 

Functions from AngularJs' ng-functions not included here are either not used in the current codebase, or can be replaced with vanilla JS (things like angular.forEach, for example).

Open to changing the naming, but felt `umb` was an appropriate namespacing, particularly as `Umbraco` is already used to provide access to the server variables collection. I guess we could attach methods to that object, but that might prove confusing.

Given all that, this change doesn't actually do anything, just adds a collection of new methods. Can verify these are available globally by bashing `umb` into the browser console, to return the object and its contained methods. Can even test the methods if you're so inclined, most are obvious in their intent.


nathanwoulfe:
Annnd had to rework things a bit, the first way I'd added utility.js didn't play nice with Karma, so tests were failing. Not sure why, but the `umb` global couldn't be found.

Instead, I'm now bundling the utility file into the umbraco.services module. Tests all pass, umb utility object is available everywhere. It doesn't feel like the best solution, am open to other ways.

nielslyngsoe:
Hi @nathanwoulfe 
Good start — I do think we need to consider making the name a bit more explicit.
Not that I have a great name up my sleeve...

I don't think it needs to be named anything in relation to Umbraco, so I would say we could get rid of the "umb".
Utility is a good name but maybe too general?

Any thoughts, ideas to elaborate on?

nathanwoulfe:
@nielslyngsoe I jumped between `utility` and `helpers` and those generic type names. Definitely happy to change from `umb` too - was really a case of having to call it something! 

Maybe `utility` isn't too bad... After all, that is what it is

nielslyngsoe:
@nathanwoulfe if you like it, I think we can go with it :-D. Then we can always be enlightened later if someone has a great name. 🌟

nathanwoulfe:
Descriptive naming is good naming. I'll update the PR in the morning.

nielslyngsoe:
Cool, and just let me know if you need anything or have some thoughts you like some perspectives to 😊 then I'm right here.

src/Umbraco.Web.UI.Client/.eslintrc
```diff
@@ -19,10 +19,10 @@
         "tinyMCE": false,
         "FileReader": false,
         "Umbraco": false,
+        "umb": false,
```
abjerner: @nathanwoulfe seems you missed the new name here:

```suggestion
        "Utilities": false,
```

src/Umbraco.Web.UI.Client/gulp/config.js
```diff
@@ -28,9 +28,9 @@ module.exports = {
             installer: { files: "./src/installer/**/*.js", out: "umbraco.installer.js" },
             filters: { files: "./src/common/filters/**/*.js", out: "umbraco.filters.js" },
             resources: { files: "./src/common/resources/**/*.js", out: "umbraco.resources.js" },
-            services: { files: "./src/common/services/**/*.js", out: "umbraco.services.js" },
+            services: { files: ["./src/common/services/**/*.js", "./src/utility.js"], out: "umbraco.services.js" },
```
abjerner: @nathanwoulfe I may be a little picky, but as the global variable name is now in plural, shouldn't the file name also be? Both the actual file name and the reference here.

src/Umbraco.Web.UI.Client/src/utility.js
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Equivalent to angular.noop
+     */
+    const noop = () => { };
+
+    /**
+     * Equivalent to angular.copy
+     * Abstraction of underscore.clone
+     */
+    const copy = val => _.clone(val);
```
abjerner: @nathanwoulfe `angular.copy` does a deep clone, while `_.clone` only does a shallow clone. From what I can tell, underscore doesn't have a deep clone function, so we'll most likely have to build this our selves. Or as Angular.js is MIT, we may borrow it from their source code (they seem to cover a lot of different cases, and it's probably well tested).

Exhibit A:

```javascript
var a = { hello: { world: true } };

var b = angular.copy(a);

a.hello.world = false;

console.log(b.hello.world); // outputs "true"
```

Exhibit B:

```javascript
var a = { hello: { world: true } };

var b = _.clone(a);

a.hello.world = false;

console.log(b.hello.world); // outputs "false"
```

src/Umbraco.Web.UI.Client/src/utility.js
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Equivalent to angular.isArray
+     */
+    const isArray = val => Array.isArray(val) || val instanceof Array;
+
+    /**
+     * Equivalent to angular.equals
+     * Abstraction of underscore.isEqual
+     */
+    const equals = (a, b) => _.isEqual(a, b);
```
abjerner: `angular.equals` and `_.isEqual` seem to have very different implementations, so this part probably requires some more testing.

src/Umbraco.Web.UI.Client/gulp/config.js
nathanwoulfe: @abjerner yup, I agree, naming should be consistent. Will fix that.

As for the _ vs Angular bits, I think we'll need to make those changes case-by-case, in some places it will be fine to swap, others, like you said, the different implementation may return unexpected results...

abjerner: @nathanwoulfe great 👍 

I think the copy function is the most critical. For various infinite editors and property editors, Umbraco will make a deep clone of the selection so the user makes changes to the copy rather than the original selection. So I think this is the most important one to have working as-is, as it *will* break existing functionality.

I don't know the impact of the other functions, but I suspect they are less important.

src/Umbraco.Web.UI.Client/src/utility.js
nathanwoulfe: Need to check too how we actually use it - might be more efficient to include a method rather than importing from _

src/Umbraco.Web.UI.Client/src/utility.js
nathanwoulfe: @abjerner I think the most robust way would be to include the AngularJs version - removes the risk of things breaking due to differing implementations. That said, that's a complex piece of code, so wondering if instead, for now at least, Utilities.copy() is simply a facade to angular.copy() - would serve the same purpose of isolating references, and can be dealt with in due course without blocking other changes. Same for angular.equal.

We could look at importing individual functions from lodash (where cloneDeep is a thing), but that means changes to the gulp build to introduce browserify or similar to manage getting the CommonJs stuff into the browser...

abjerner: It's not like we're skipping Angular for Umbraco 8 anytime soon, so I agree that having `Utitilities.copy` call `angular.copy()` for now would be an acceptable solution. Adding another dependency should be avoided if possible.

nathanwoulfe:
@abjerner, yeah, had assume `Array.prototype.forEach` in place of `angular.forEach`

Had intended using `Object.assign` in place of `angular.extend`, but that might have issues with potentially overriding hashkeys - extend protects $$props, which is required when working with two AngularJs objects... Will add it in, but can probably apply it case-by-case - depending on the context, `Object.assign` might be perfectly ok.

nul800sebastiaan:
Hey @nathanwoulfe - great start! Sorry for the delay but world events kind of interfered 🙈 

Thanks again!! 👍 ⭐️

nathanwoulfe:
@nul800sebastiaan these are strange times indeed... 

Having this merged means the other related tasks can now be tackled 👍

PR #7744, opened 2020-03-02
NetCore: Refactor internal usages of UmbracoHelper

elit0451:
* Removed several properties from **UmbracoHelper** class:
  - `ContentQuery`
  - `TagQuery`
  - `CultureDictionaryFactory`
  - `ComponentRenderer`
* Removed Media related methods from there as well
* Removed `UmbracoHelper` from **UmbracoApiControllerBase** 
* Refactored all `UmbracoHelper` references (also in tests)
* Injected the needed Interfaces to classes that previously used UmbracoHelper to get the same functionality

## Test:
- Make sure all tests pass
- No back-office functionality is changed


src/Umbraco.Web/UmbracoHelper.cs
```diff
@@ -180,61 +180,7 @@ public string GetDictionaryValue(string key, string altText)
...
-        /// <returns>The content, or null of the content item is not in the cache.</returns>
-        public IPublishedContent Content(string id)
-        {
-            return ContentForObject(id);
-        }
-
-        public IPublishedContent Content(Udi id)
-        {
-            return _publishedContentQuery.Content(id);
-        }
```
elit0451: Have these been removed by mistake?

src/Umbraco.Web/UmbracoHelper.cs
```diff
@@ -247,127 +193,87 @@ public IPublishedContent ContentSingleAtXPath(string xpath, params XPathVariable
...
         /// </summary>
         /// <param name="ids">The content identifiers.</param>
         /// <returns>The existing contents corresponding to the identifiers.</returns>
         /// <remarks>If an identifier does not match an existing content, it will be missing in the returned value.</remarks>
-        public IEnumerable<IPublishedContent> Content(IEnumerable<object> ids)
-        {
-            return ContentForObjects(ids);
-        }
+        public IEnumerable<IPublishedContent> Content(IEnumerable<object> ids) => _publishedContentQuery.Content(ids);
+
```
elit0451: Shouldn't this also be removed as it is now declared in **IPublishedContentQuery.cs**  and implemented in **PublishedContentQuery.cs**?

src/Umbraco.Web/UmbracoHelper.cs
bergmania: By mistake.. I reintroduce..

src/Umbraco.Web/UmbracoHelper.cs
```diff
@@ -497,69 +315,37 @@ public IPublishedContent Media(object id)
...
-        {
-            if (ConvertIdObjectToInt(id, out var intId))
-                return _publishedContentQuery.Media(intId);
-            if (ConvertIdObjectToGuid(id, out var guidId))
-                return _publishedContentQuery.Media(guidId);
-            if (ConvertIdObjectToUdi(id, out var udiId))
-                return _publishedContentQuery.Media(udiId);
-            return null;
-        }
+        private IPublishedContent MediaForObject(object id) => _publishedContentQuery.Media(id);
```
elit0451: There is a similar case related to Content (`IPublishedContent ContentForObject(object id)`) where the method was entirely removed as the code block was moved into the [`IPublishedContent Content(object id)`](https://github.com/umbraco/Umbraco-CMS/blob/28bccd4ab2458d4bd9529879479639d43bc9b4b9/src/Umbraco.Infrastructure/PublishedContentQuery.cs#L104) of **PublishedContentQuery.cs**. Regarding Media we also have similar behaviour into https://github.com/umbraco/Umbraco-CMS/blob/28bccd4ab2458d4bd9529879479639d43bc9b4b9/src/Umbraco.Infrastructure/PublishedContentQuery.cs#L149. Could you just clarify what is the right approach here?

bergmania: Should be reintroduced..

The right approach for now is that everything is available on UmbracoHelper, but we should internally not use it. We should use the underlying service. 

The implementations in UmbracoHelper should just wrap the underlying services..

UmbracoHelper cannot be removed for now, as the query builder in templates builds queries  using the UmbracoHelper

src/Umbraco.Web/UmbracoHelper.cs
elit0451: Got the answer in https://github.com/umbraco/Umbraco-CMS/pull/7744#discussion_r386907044 -> resolving

src/Umbraco.Web/UmbracoHelper.cs
elit0451: Got it, thanks 💪

elit0451: Reintroduced it in https://github.com/umbraco/Umbraco-CMS/pull/7744/commits/0fdd0e9d646d019882fb828d4ae9dacb1ebdf57e :slightly_smiling_face: After seeing the tests pass will merge if you don't mind

PR #7746, opened 2020-03-02
V8: Do not allow removing all items in NC when a minimum is being enforced

kjac:
### Prerequisites

### Description

If Nested Content is configured with:

- Only one element type.
- A minimum number of items required.

...the editor automatically creates and enforces this minimum number of items - meaning you can't delete items beyond the minimum number of items required:

image

But you can circumvent this by using the property action "Remove all entries":

image

What's more... Nested Content actually lets you save and publish without the minimum number of items (probably because it shouldn't be possible to remove them in the first place):

image

This PR disables "Remove all items" when a minimum number of items is being enforced by the editor:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
@@ -592,7 +592,8 @@
 
         function updatePropertyActionStates() {
             copyAllEntriesAction.isDisabled = !model.value || !model.value.length;
-            removeAllEntriesAction.isDisabled = copyAllEntriesAction.isDisabled;
+            // in addition to the copy action state, remove all entries should be disallowed if a minimum number of items is being enforced
+            removeAllEntriesAction.isDisabled = copyAllEntriesAction.isDisabled || (model.config.contentTypes.length == 1 && vm.minItems !== 0);
         }
 
 


nielslyngsoe:
Hi @kjac 

Good catch and I agree that the inconsistent experience should be corrected. I do though advocate for a different solution. But I'm open for other reasons if you have a case to present or if you are fond of a different solution?

Overall I advocate for correcting the system for editors to work as they like, begin free in their approach to the data, meaning they can work without living up to requirements/validation. When writing content.

In regards to this, the case could be that an Editor needs to replace 4? features in a Nested Content list, with a minimum requirement of 4?. Maybe the editor did copy the full content-node or they had dummy data or another reason for this scenario to happen.
One way to work with the data is by replacing values in each entry, but in this way, you might forget to replace on the field, especially if you are leaving your desk or navigating to retrieve the correct data while performing this change.
To avoid this, I would argue that the editor would like to delete all entries before creating the new ones. To ensure a clean slate, avoiding overseen values existing in the new content.

So that said I would go for enabling the user to remove all entries at any time. As well saving this — if they had to leave midway.
But validation should be preventing publishing.

I think this is the scenario today, except what you discovered, that the button gets disabled. This should be changed so the editor always can delete an item.

Let me know what you think about this approach, Thanks in advance and thanks for begin aware of this detail! 🌟👏

kjac:
Hi @nielslyngsoe 

I'm completely onboard with your vision for how item based editors in general should be less restrictive and more informative in their approach to enforcing validation of min and max number of items.

This PR is simply a patch for a regression error. As-is the editors can bypass the validation completely, which won't do. 

Once validation for min and max number of items is reimplemented along your guidelines, this PR (and many other working towards fixing the NC validation) becomes void.

PR #7747, opened 2020-03-02
Create Content Blueprints - Semantic fixes

poornimanayar:
As suggested [here](https://github.com/umbraco/Umbraco-CMS/pull/7020#issuecomment-593055598) I have added `type` attribute to `button` and taken away the `prevent-default` attribute.

Mentioning @abjerner as he tested the other PR :-)


abjerner:
Thanks @poornimanayar 👍 

It looks like discussed in #7020, so I can't imagine it wouldn't work. But I'll make sure to test it as well 😉

PR #7750, opened 2020-03-03
Netcore: Untangle membership helper and trim Web even more

bergmania:
# Changes
- Introduced `IRequestAccessor`
  - Methods to get values posted or in query string
  - Proxy the `UmbracoModule` request events in `IRequestAccessor`
- Replaced direct usages of HttpContext.Request with the new IRequestAccessor

- Introduced `IMemberUserKeyProvider`
- Introduced `IPublicAccessChecker`
  - Removes some complexity with dependency on `MembershipHelper` from `PublishedRouter`
- Introduced `ISessionManager`
  - Get/Set value
- Deleted `WebCachingAppCache` and replaced with the `ObjectCacheAppCache`. These should be identical function-wise.
- Merged `AspNetSessionIdResolver` into the new `AspNetSessionManager`
- Removed unused functionality in Models about `doLookup`
  - `RegisterModel`

# Tests 
- Backoffice should work en general


bergmania:
Good catches, thanks 💪

PR #7752, opened 2020-03-03
Obsoleting UmbracoHelper properties

elit0451:
This PR obsolete some of the properties on UmbracoHelper that is removed in the .NET Core branch.

- `ContentQuery`
- `TagQuery`
- `CultureDictionaryFactory`
- `ComponentRenderer`
- `MembershipHelper`

### DIFF
--- a/src/Umbraco.Web/UmbracoHelper.cs
+++ b/src/Umbraco.Web/UmbracoHelper.cs
@@ -94,7 +94,7 @@ namespace Umbraco.Web
         /// <summary>
         /// Gets the membership helper.
         /// </summary>
-        [Obsolete("Inject and use an instance of " + nameof(Security.MembershipHelper) + " in the constructor instead")]
+        [Obsolete("Inject and use an instance of " + nameof(Security.MembershipHelper) + " in the constructor instead. In views you can use @Members.")]
         public MembershipHelper MembershipHelper => Ensure(_membershipHelper);
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Umbraco.Web/UmbracoHelper.cs
```diff
@@ -70,24 +70,31 @@ internal UmbracoHelper()
...
         /// Gets the query context.
         /// </summary>
+        [Obsolete("Inject and use an instance of " + nameof(IPublishedContentQuery) + " in the constructor for using it in classes or get it from Current.PublishedContentQuery in views")]
         public IPublishedContentQuery ContentQuery => Ensure(_publishedContentQuery);
 
         /// <summary>
         /// Gets the membership helper.
         /// </summary>
+        [Obsolete("Inject and use an instance of " + nameof(Security.MembershipHelper) + " in the constructor instead")]
         public MembershipHelper MembershipHelper => Ensure(_membershipHelper);
```
AndyButland: Just noted that there's not an "or get it from Current. to use it in views" option on this property, so wondering what the approach would be here if someone wanted to use `MembershipHelper` in views?  If it's planned to support that, then probably should update the Obsolete attributes description to explain how.

bergmania: The only reason not to add it to Current, is because it is available from `UmbracoViewPage`

So you should already be able to just use `@Members` in views.. I will add that to the message

PR #7753, opened 2020-03-04
Removed duplications of tree routing

Matthew-Wise:
Ran a duplication report against the backoffice code. - https://gistpreview.github.io/?600a4f0c2847e37197beba1579403bf6 expand all works no other buttons do.
 (raw: https://gist.github.com/Matthew-Wise/600a4f0c2847e37197beba1579403bf6)
Which highlights "copy paste" code. 

Whilst finding a good location to remove the duplication I also found similar code in the navigation service so created a new method they could all use.
Tested this change using the members -> All members route


emmaburstow:
Heyyyy,

Thanks ever so much for the PR. We'll take a look and let you know if we have any questions.

Em

nul800sebastiaan:
Very nice @Matthew-Wise - great cleanup! 👍 ⭐️

PR #7755, opened 2020-03-04
V8: Fix completely broken(tm) media pickers

kjac:
### Prerequisites

### Description

#7449 introduced some rewriting of the media picker controller to replace lodash methods with native es6 equivalents. For reasons I can't really explain, this has resulted in a completely broken(™) media picker where every picked item renders as trashed even though they're really not:

image

Moving the code comment outside the one-liner in `.find()` solves it, and brings back the picked items:

image


nul800sebastiaan:
WEIRD! 

@nathanwoulfe take note.. code comments within lambdas do NOT work in JS like in C# 😱

nathanwoulfe:
I guess that makes sense - js will treat the first line after the arrow as a return statement, even if it's a comment. 

Strange though that I didn't notice that in the UI, it's pretty obvious...

Issue #7754, opened 2020-03-04
umb-overlay inside umb-editor is hidden


enkelmedia:
We're using some "Nested Nested Content" inside Doc Type Grid Editor, this works fine until there is more than one Document Type configured on the Nested Content Data Type. In these cases a "umb-overlay" is shown so that the editor can pick which DocType to add, like this:

![screen - nested content choose type](https://user-images.githubusercontent.com/1782524/75871313-a5a69e00-5e0c-11ea-80c1-e31a0fe69828.PNG)

This works all fine, but when Nested Content is shown inside a EditorService-dialog, the umb-overlay will set it's style to left:xxx px; so that it ends up outside the screen, this is how it looks when clicking "Add Content"

![screen - nested in editor service](https://user-images.githubusercontent.com/1782524/75871616-21a0e600-5e0d-11ea-9c61-1a04de8da3c9.PNG)

After removing the "left"-style from the umb-overlay the box shows again:

![screen nested inside editorservice removed left](https://user-images.githubusercontent.com/1782524/75871666-34b3b600-5e0d-11ea-8de7-1d6688aa1fec.PNG)

This issue is present only with "Doc Type Grid Editor" but also when for example configuring a "Multi Node Tree Picker" to have the "Open"-button. When opening an Node and trying to choose a DocType for Nested Content it looks like this:

![screen - omb-overlay in editor service edit content](https://user-images.githubusercontent.com/1782524/75872082-ddfaac00-5e0d-11ea-8765-e6e7b36dd77f.PNG)

In other words, the umb-overlay is placed to much to the right so that the editor can't choose a DocType. The issue is specially big when the editorService is opening a window that is of type "small".

We've solved this with some CSS-hacks but I guess it would be good to come up with something in Core?

Our solution was to put any umb-overlays to the right using CSS like this but there might be better solutions =D

```css
.umb-editors {
    .umb-overlay {
        left: inherit !important;
        right: 0 !important;
        top: 0 !important;
        bottom: 0 !important;
    }
}
```

nul800sebastiaan:
Just make sure to always mention the Umbraco version you're on please so we can try to reproduce!

abjerner:
We have experienced this as well, but I haven't gotten around to creating a PR or even creating an issue yet.

I think this is in 8.5, but not in 8.4. But I haven't really tested beyond this.

[...]

PR #7759, opened 2020-03-04
Implemented fix for #7754

abjerner:
Following the discussion in #7754, I've updated the grid property editor so it's now using the `overlayService` instead of the deprecated `umbOverlay` directive.

overlayService


nul800sebastiaan:
Lovely, thanks @abjerner! 👍

PR #7760, opened 2020-03-05
Creates a new base class UmbracoRenderController

Shazwazza:
This becomes the base class for RenderMvcController and the utlity methods are moved there. This also makes IRenderController discoverable instead of only discovering IRenderMvcController. This all should have been done before v8 release along with removing ContentModel but we are stuck with that. In the meantime, this makes it nicer to hijack routes.

This is a follow up to this issue here https://github.com/umbraco/UmbracoDocs/issues/2026 in that some things are a little confusing which is mostly due to some left over bits from v7 and not getting around to changing this behavior for 8.0 release.

With this change it means that you can implement your own render controller without being forced to know about the legacy `ContentModel` because you aren't forced to use or override the Index action which binds the ContentModel. With this change it means that you can reference the currently rendering content as it was supposed to be based on an injected service instead of being magically bound to a parameter.

For example, a custom render controller with this change could look like this:

```cs
    public class MyRenderController : UmbracoRenderController
    {
        public MyRenderController(IGlobalSettings globalSettings, IUmbracoContextAccessor umbracoContextAccessor, ServiceContext services, AppCaches appCaches, IProfilingLogger profilingLogger, UmbracoHelper umbracoHelper)
            : base(globalSettings, umbracoContextAccessor, services, appCaches, profilingLogger, umbracoHelper)
        {
        }

        public ActionResult Index()
        {
            // get the current content in the request
            var model = CurrentPage;
            return CurrentTemplate(model);
        }
    }
```

I will write up more info/details on the issue mentioned above.

### DIFF
--- a/src/Umbraco.Web/Composing/CompositionExtensions/Controllers.cs
+++ b/src/Umbraco.Web/Composing/CompositionExtensions/Controllers.cs
@@ -60,9 +60,9 @@ namespace Umbraco.Web.Composing.CompositionExtensions
             var nonUmbracoWebPluginController = composition.TypeLoader.GetTypes<PluginController>().Where(x => x.Assembly != umbracoWebAssembly);
             composition.RegisterControllers(nonUmbracoWebPluginController);
 
-            // can and register every IRenderMvcController in everything (IRenderMvcController is IDiscoverable)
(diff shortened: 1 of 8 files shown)


enkelmedia:
When will we see this make it to a release? :)

Shazwazza:
This is marked for the community to review and test so that needs to be done first

PR #7761, opened 2020-03-05
Remove ContentFinderByRedirectUrl if RedirectUrlTracking is disabled

marcemarc:
We have a config setting for turning off RedirectUrlTracking in UmbracoSettings Routing section - if RedirectUrlTracking is disabled, we don't add redirects to the dashboard, when Urls change, however any existing 301 redirects in the database table continue to work, as the ContentFinderByRedirectUrl is still registered, regardless of the value of the setting. in V7 the ContentFinder was removed when the RedirectUrlTracking setting was disabled: https://github.com/umbraco/Umbraco-CMS/blob/efbe97a01da63d34a505dad8b3e8ae1fe0729621/src/Umbraco.Web/Routing/RedirectTrackingEventHandler.cs#L36

This PR just adds the check for the setting before the ContentFinderByRedirectUrl is registered, and it is only added if the setting is not disabled (I know it's round way and I'm testing for something 'not being disabled' - but I didn't name the setting!)

In order to test, install a fresh version of Umbraco with the starter kit, change some of the names of pages to generate redirects in the redirect tracking database table.

Test the redirects work.

Now disable RedirectUrlTracking, by adding the attribute in the routing section of the UmbracoSettings.config and setting it to true:

 ` <web.routing trySkipIisCustomErrors="true" internalRedirectPreservesTemplate="false" disableAlternativeTemplates="false" validateAlternativeTemplates="false" disableFindContentByIdPath="false" umbracoApplicationUrl="" disableRedirectUrlTracking="true">`

Now revisiting the redirects you created earlier, they should not 'redirect' (make sure the browser isn't caching them)


emmaburstow:
Hullo Marc,

Thanks for the work. As ever, thanks for the excellent documentation of the work too. One of us will take a look and let you know if we've got questions or comments. 

In the meantime, hope you're well.

Em

nul800sebastiaan:
Thanks @marcemarc  - this works great.

I'll mark it as a breaking change since people could be relying on the "wrong" behavior where their existing redirects would keep working after disabling URL tracking.

marcemarc:
Thanks @nul800sebastiaan !!

ronaldbarendse:
IMHO this was the correct behavior, as it disabled _tracking_ URL changes and creating new redirects, not stop redirecting... I would recommend adding a setting `disableRedirectUrls` to explicitly disable all redirecting logic, including not adding the dashboard.

We can always revert to the previous behavior by adding back the `ContentFinderByRedirectUrl`:

```c#
[ComposeAfter(typeof(WebInitialComposer))]
public class EnableRedirectUrlComposer : IComposer
{
   {
      if (composition.Configs.Settings().WebRouting.DisableRedirectUrlTracking)
      {
         composition.ContentFinders().Append<ContentFinderByRedirectUrl>();
      }
   }
}
```

Also, the confirmation dialog that's shown when clicking `Disable URL tracker` (only visible for admins) should at least warn about stopping all existing redirects and this also restarts the site (because of a config file change):

Disable URL tracking

PR #7763, opened 2020-03-05
No more TypedContentAtRoot

bjarnef:
### Prerequisites

### Description
Just a minor fix in the comments mentioning `TypedContentAtRoot`, which no longer exists.


abjerner:
@bjarnef thanks 👍 

As `ContentAtRoot` is a specific method, wouldn't it make sense to add it as a reference? Eg. `<see cref="UmbracoHelper.ContentAtRoot"/>`? Then it will also result in a warning should be method be deleted in the future.

bjarnef:
@abjerner not sure. Which class should it reference to since it exists in three different classes:
https://github.com/umbraco/Umbraco-CMS/search?q=ContentAtRoot&unscoped_q=ContentAtRoot

abjerner:
@bjarnef I thought about that as well, but I think ´UmbracoHelper` will be the most likely one to use.

But perhaps the remarks could be something like this?
```
This can be useful in order to return all nodes in an entire site by a type when combined with <see cref="UmbracoHelper.ContentAtRoot"/> or similar.
```

nul800sebastiaan:
Updated as per Anders' suggestion. Thanks @bjarnef and @abjerner! 👍

PR #7765, opened 2020-03-06
Allows assembly scanning when using RegisterAssembly

Shazwazza:
This PR fixes this issue https://github.com/umbraco/Umbraco-CMS/issues/7502 but please read that thread and we'll need to determine if we use this PR or this one https://github.com/umbraco/Umbraco-CMS/pull/7588/files or a combination

Have moved the container customization to UmbracoServiceContainer. This captures the original assembly scanner but continues to replace it with a noop assembly scanner. Then explicitly override all calls to RegisterAssembly where we replace the assembly scanner with the original non-noop one in a using clause which when disposed re-replaces again with the noop one. This is not thread safe but that is ok since in no case is the container ever being configured by multiple threads. This is a reasonably simple approach to allow assembly scanning for the RegisterAssembly methods.

### DIFF
--- a/src/Umbraco.Core/Composing/LightInject/LightInjectContainer.cs
+++ b/src/Umbraco.Core/Composing/LightInject/LightInjectContainer.cs
@@ -1,11 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Reflection;
 using System.Threading;
 using LightInject;
 
 namespace Umbraco.Core.Composing.LightInject
 {
+
     /// <summary>
     /// Implements DI with LightInject.
     /// </summary>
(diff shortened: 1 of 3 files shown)


rustybox:
Did you see this also https://github.com/umbraco/Umbraco-CMS/pull/7588/commits/abfacf78d7efefda4c2d9ce6538f8ce060d4c2e2

I tried this out before switching to GetEmitter.

It's less explicit but achieves the same as this PR by abusing the fact that LightInject uses a different method signature in fallback scenario to explicit request to register assembly.

Shazwazza:
@rustybox will close my PR in favor of yours (have commented)

PR #7778, opened 2020-03-09
Creates a netstandard compliant TypeFinder

Shazwazza:
Actually, the TypeFinder itself hasn't changed much but this introduces a new abstraction: `IAssemblyProvider` which is the thing that returns the list of assemblies to scan for type scanning.

Previously we would:

* If hosted in netframework, use the BuildManager's list of assemblies which would have all assemblies pre-loaded even if the appdomain hadn't loaded them yet and this was also based on referenced types in the config xml
* If not hosted we would 'brute force' load all assemblies in the /bin folder and in the current AppDomain ... which would include all GAC assemblies since the AppDomain has those all pre-loaded 

There is no build manager in netcore or netstandard so that doesn't work.  We cannot just use the appdomain, there's many notes in our codebase about this because although the AppDomain may seem to return more DLLs than BuildManager, it's only because it includes all Gac assemblies but it will not include any non-loaded assembly which could be transitive dependencies since assemblies are actually lazily loaded into an appdomain when types are referenced which means type scanning on startup will miss some things in some cases.

So what to do? I've blogged about this previously a couple of times and in the past and there is actually a way to do this in netstandard which is how they used to do this in netcore 2 which scans the output `.deps.json` file produced from netcore builds. This is the code that could work: https://github.com/Shazwazza/Umbraco-CMS/blob/dev-v9/src/Umbraco.Core/Plugins/DefaultUmbracoAssemblyProvider.cs **But** this is still only going to return candidate assemblies that are directly or transitively referenced by the entry assembly, that won't actually work for assemblies that might be 'dropped' into the /bin. 

Having a look at what netcore 3 does for it's runtime scanning.... Netcore 3 limits the amount of runtime type scanning as much as possible by doing the type scanning at build time (see https://github.com/dotnet/aspnetcore-tooling/pull/598/files). This dynamically embeds assembly attributes in your code `ApplicationPartAttribute`) which pre-determines if your DLL should be scanned. So that when MVC scans assemblies it's only scanning the entry point assembly and looking for other ApplicationPartAttributes to include in the scanning. 

We still need to dynamically scan assemblies and cannot do any fancy build level scanning because Umbraco's plugin model still means that assemblies could be 'dropped' into the /bin without being referenced. 

So we borrow some concepts from netcore3's build time and runtime scanning and come up with a hybrid. The new default `IAssemblyProvider` will use the entry point assembly to see if it references directly or transitively Umbraco.Core/Umbraco.Web, it will then include Umbraco.Core and Umbraco.Web as entry point assemblies too (since these need to be scanned), and then it will get all unique locations of these entry point assemblies and include DLL files found there but all of this will use a big ugly list of exclusions to exclude anything we know not to scan.

Questions/Comments:

* During type scanning in TypeFinder we also do a check for `TypeHelper.GetReferencingAssemblies` but I think we can remove this overhead now because we are only including assemblies to scan that directly or transitively reference Umbraco.Web and Umbraco.Core ... We can actually add a flag for this to disable or enable this feature with the TypeFinder options, then we can make a benchmark.net test and see the results
* We can remove the BuildManagerTypeFinder, BuildManagerAssemblyProvider and BruteForceAssemblyProvider since this single default assembly provider will just work
* How can we test performance of this compared to the original BuildManager reference list? I don't think there's a good way because we have no control over the exact moment that build manager finds assemblies in a net framework web app vs using this in a netcore web app. Even if we tried to test the difference in a net framework web app I'm not sure we can because we still have no control over when buildmanager runs. ... but something tells me that BuildManager was slower, at least on a cold boot because it would at runtime need to go figure out all DLLs being referenced including going through xml config. It's probable that this output was cached but not sure. Our TypeFinder caches things anyways on disk so all were worried about is cold boot but since the logic to include DLLs that we are using is very simple it should be very fast too.


bergmania:
> During type scanning in TypeFinder we also do a check for TypeHelper.GetReferencingAssemblies but I think we can remove this overhead now because we are only including assemblies to scan that directly or transitively reference Umbraco.Web and Umbraco.Core ... We can actually add a flag for this to disable or enable this feature with the TypeFinder options, then we can make a benchmark.net test and see the results

I'm really unsure here, I can't come up with a use case we this code is necessary..


> We can remove the BuildManagerTypeFinder, BuildManagerAssemblyProvider and BruteForceAssemblyProvider since this single default assembly provider will just work

I think we should just remove these when the default should work.

> How can we test performance of this compared to the original BuildManager reference list? I don't think there's a good way because we have no control over the exact moment that build manager finds assemblies in a net framework web app vs using this in a netcore web app. Even if we tried to test the difference in a net framework web app I'm not sure we can because we still have no control over when buildmanager runs. ... but something tells me that BuildManager was slower, at least on a cold boot because it would at runtime need to go figure out all DLLs being referenced including going through xml config. It's probable that this output was cached but not sure. Our TypeFinder caches things anyways on disk so all were worried about is cold boot but since the logic to include DLLs that we are using is very simple it should be very fast too.

I'm not that worried about performance here..  I also can't find a good way to test this directly, 
I think it is good enough, just to compare the cold start time of Umbraco before and after.  But I know there are lots of other elements involved in the total execution time then.

Shazwazza:
So i ran a benchmark test with and without the referenced assemblies lookup when resolving types and the differences is super minimal so I'm just gonna leave it as-is. 

|                               Method |     Mean |    Error |   StdDev |   Median | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------------------- |---------:|---------:|---------:|---------:|------:|--------:|------------:|------------:|------------:|--------------------:|
|    WithGetReferencingAssembliesCheck | 294.1 ms | 5.713 ms | 29.11 ms | 288.0 ms |  1.00 |    0.00 |           - |           - |           - |             1.37 MB |
| WithoutGetReferencingAssembliesCheck | 288.5 ms | 4.708 ms | 24.29 ms | 286.2 ms |  0.99 |    0.12 |           - |           - |           - |             1.33 MB |

I agree about performance, i think it will be fine but we can and should benchmark startup times when we can between the netcore and v8 version when it's possible and can look into this to see if it's a bottleneck (but i highly doubt it).

I have removed the other classes.

bergmania:
Do we have an alternative to `Assembly.GetEntryAssembly()` that returns `null`? - At least in my case..

Shazwazza:
yes, in all tests (nunit) that will return null. You can just give it the assembly you want, like we do in the unit tests.

bergmania:
I know, but currently, I can't start the solution unless I change the line in CoreRuntime:
`new TypeFinder(Logger, new DefaultUmbracoAssemblyProvider(Assembly.GetEntryAssembly()));` to `new TypeFinder(Logger, new DefaultUmbracoAssemblyProvider(Assembly.GetEntryAssembly() ?? GetType().Assembly));` or similar, due to `Assembly.GetEntryAssembly()` is returning `null`

Shazwazza:
@bergmania ok lets chat today, i'll have a look shortly once i get through all my emails, etc... This should always return something according to all the netcore stuff i was reading. They actually exposed this API specifically for this at the request of the aspnet team.

Shazwazza:
@bergmania right so this doesn't return null in netcore3 app but does in net framework :/ 
I'll investigate options in netframework but we cannot use GetType().Assembly since that is not the entry assembly which is always that main executable (i.e. will be someones own assembly that references umbraco)

Shazwazza:
@bergmania have pushed an update with a bunch of notes for you to read :)

bergmania:
@Shazwazza, changes looks good and my tests did not find anything not working. 

Do you want me to merge, or should we resolve any of the TODOs before. The one with passing in an explicit list of assembly names should be possible to resolve already

Shazwazza:
@bergmania lets leave it for now, i'll merge this in and will keep a note about this as there may be a need to tweak typefinder later anyways.

PR #7782, opened 2020-03-10
Added GetSegment() method to VariationContext and uses it when conten…

PerplexDaniel:
#### Description of PR
Adds a `GetSegment(int contentId)` method to `VariationContext`. This allows implementations to vary the segment based on the `contentId` instead of only being able to provide 1 segment for a request.

This code change was proposed by @Shazwazza after a call with us. We initially wanted to change the `IVariationContextAccessor` but changes to public interfaces are not allowed as it would break all existing code.

We have also updated the usage of `VariationContext.Segment` to use `GetSegment()` with the appropriate `contentId` whenever possible.

#### Why is this useful
Being able to render segmented property data from a detail page (e.g. Product) on a page that itself does not use segments (e.g. Product overview page). So when the properties of the Product are rendered in the overview, those should use some segment (e.g. `"segment-xyz"`) whereas properties of the overview page itself should be rendered with the default segment (`""`).
Without this PR, you can only define 1 segment in total for the entire request. So all properties of the current page would rendered using a single segment.

### DIFF
--- a/src/Umbraco.Core/Models/PublishedContent/VariationContextAccessorExtensions.cs
+++ b/src/Umbraco.Core/Models/PublishedContent/VariationContextAccessorExtensions.cs
@@ -3,13 +3,32 @@
     public static class VariationContextAccessorExtensions
     {
         public static void ContextualizeVariation(this IVariationContextAccessor variationContextAccessor, ContentVariation variations, ref string culture, ref string segment)
+            => variationContextAccessor.ContextualizeVariation(variations, null, ref culture, ref segment);
+
+        public static void ContextualizeVariation(this IVariationContextAccessor variationContextAccessor, ContentVariation variations, int contentId, ref string culture, ref string segment)
+            => variationContextAccessor.ContextualizeVariation(variations, (int?)contentId, ref culture, ref segment);
+
+        private static void ContextualizeVariation(this IVariationContextAccessor variationContextAccessor, ContentVariation variations, int? contentId, ref string culture, ref string segment)
         {
             if (culture != null && segment != null) return;
 
             // use context values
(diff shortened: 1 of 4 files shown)


src/Umbraco.Core/Models/PublishedContent/VariationContextAccessorExtensions.cs
```diff
@@ -3,13 +3,32 @@
     public static class VariationContextAccessorExtensions
     {
         public static void ContextualizeVariation(this IVariationContextAccessor variationContextAccessor, ContentVariation variations, ref string culture, ref string segment)
```
Shazwazza: Does this method get used anymore? Else we should obsolete it

PerplexDaniel: It's still used in 2 locations in the `PublisedValueFallback` ([here](https://github.com/umbraco/Umbraco-CMS/blob/3eb4c642c39709840ce117a00a64c63f2526c7a4/src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs#L36) and [here](https://github.com/umbraco/Umbraco-CMS/blob/3eb4c642c39709840ce117a00a64c63f2526c7a4/src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs#L76)) as there does not seem to be a `contentId` available in those contexts. So I suppose in some fallback scenarios perhaps this new functionality will not be used yet.

Shazwazza: Ok no problemo we'll leave as-is for now

PR #7783, opened 2020-03-10
Remove Start Nodes from Cookie

### LINKED ISSUE #7727: Umbraco stores permissions and start Nodes in cookie
When you have a big solution with a lot of sites on one instance(+100) and you decide to manage permissions, start nodes, start media nodes you will hit an issue with an oversized cookie which will cause permanent block for login to back office. 

## Umbraco version

I am seeing this issue on all Umbraco versions(7-8)

Reproduction
------------

### Steps to reproduce

1) Add ~150 start nodes/media startnodes to the user
2) Try login as a user

### Expected result

User will be able to log in and manage all 150 start nodes

### Actual result

User will be log out from backoffice and will be not able to log in again as the cookie is too big

bielu:
### Prerequisites

If there's an existing issue for this PR then this fixes #7727 

This is a fix quite unusual issue with cookie size.

### DIFF
--- a/src/Umbraco.Core/Constants-Security.cs
+++ b/src/Umbraco.Core/Constants-Security.cs
@@ -48,8 +48,9 @@ namespace Umbraco.Core
             /// providers need to be setup differently and each auth type for the back office will be prefixed with this value
             /// </remarks>
             public const string BackOfficeExternalAuthenticationTypePrefix = "Umbraco.";
-
+            [Obsolete("Start Content Nodes Claims are not anymore in use")]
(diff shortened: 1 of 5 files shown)


emmary2020:
Hey @bielu 

Thanks for the PR. We'll take a look and shout if we need you

Emma

bielu:
@Shazwazza I think that should be merge asap, if we introduce information from External logins into identity as it will make space for that  information in cookie :P

Shazwazza:
Hi, unfortunately (but fortunately for me) I'm on holidays for 3 weeks. I won't be able to get to this one until I'm back. I've asked @nul800sebastiaan to flag this one as something to look at when I return. Cheers!

Issue #7780, opened 2020-03-09
Visible filters on user section: User filter when 0 character in Count makes filter disapear


alexanderkarlsson:
A brief description of the issue goes here.

When filtering on users, we can filter on disabled, active and so on. When count has a 0 character the filter alternative is not visible.

## Umbraco version

I am seeing this issue on Umbraco version: Umbraco 7.15.3

Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

Go to User section and go to Users. Create nine users and disable them. See that there is a filter Status filter = disabled. Create one more and disable it. The disabled filter is gone.  

### Specifics

In the code:
`<umb-dropdown-item ng-repeat="userState in vm.userStatesFilter | filter:{ count: '!0', key: '!All'}" style="padding: 8px 20px 8px 16px;">`

There is a angular filter and removes items if a 0 character is there. 

Do something of this to fix it.
https://stackoverflow.com/questions/24081004/angularjs-ng-repeat-filter-when-value-is-greater-than

nul800sebastiaan:
Thanks @alexanderkarlsson! Sounds like you know how to help, would you like to make the changes? We're always happy for a PR!

umbrabot:
Hi @alexanderkarlsson,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

alexanderkarlsson:
@nul800sebastiaan hi, i have made a PR to this now. I hope i have made the right steps to this issue

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/issues/7780

PR #7784, opened 2020-03-10
#7780 Show all filters that is more then 0 hits

alexanderkarlsson:
### Prerequisites

Add exakt 10 users and disable them. The disable filter should be there. 

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7780 

### Description
There where a ng-repeat filter that removed filters that hade count = a 0 in it. So 0, 10, 20, 30 and so on did not be visible. Now we romove that filter and use a ng-show with a count more then zero. We could fix this with a custom filter but i don´t think that is not nessasary for this problem. ng-if dont work for we cant have both ng-repeat and ng-if.


nul800sebastiaan:
Thanks @alexanderkarlsson - in v8 we have this as `filter:{ count: '!0', key: '!All'}` - does that work as well?

alexanderkarlsson:
@nul800sebastiaan Same problem there, so no. I did not look in v8. If you dont like to show filters when there is a no 0 character in count :)

PR #7786, opened 2020-03-10
Package actions not run when Umbraco Package is uninstalled

peterjrichards:
### Description
The changes are to allow Umbraco package actions to work as intended. Adding any undoable Install or Uninstall actions caused an exception because in the original code was concatinating the XML string of each action to a closed root, <actions></actions>. The exception indicated there were multiple roots and this prevented the uninstall / undo actions from being stored.

Also the PackageInstallation class was not differenciating between undoable Install and Uninstall actions, so Uninstall actions would never have run, even if the aforementioned exception did not occur.

I have therefore left the original UndoPackageActions method for its intended purpose but added an additional condition to ensure only undoable Install actions were executed; this is because Uninstall package actions would require the Execute method to be called rather than the Undo method.

There are further improvements to be made around Package Actions but this seems like a good interim solution.


emmary2020:
Hiya @peterjrichards 

Thanks ever so much for the work here. We'll take a good look and let you know if we need anything else from you. 

Speak soon,

Emma

peterjrichards:
@nul800sebastiaan I will certainly look at creating unit tests for the after behaviour. In terms of the current behaviour, you can see the warning in the logs after installing any package which has any package actions in it.

peterjrichards:
Hi @nul800sebastiaan 

Please could you review the unit tests?

Kind regards,
Peter

Issue #7789, opened 2020-03-11
UX Bug: Submitting Invalid Grid JSON Config Loses Changes


SvenPam:
When editing the editor config for grid content; validation happens on blur. This allows you to click submit straight away, subsequently closing the dilogue and (frustatingly) losing your changes.

## Umbraco version

I am seeing this issue on Umbraco version: 8.5.4

Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

### Bug summary

This is more of a UX bug than an actual issue. Given that data is lost, it can be frustrating.

### Specifics

URL: /umbraco#/settings/dataTypes/edit/{your grid data type ID}
Version: 8.5.4
Browser: Firefox 73
Other Browser's Checked: No - UX issue
Error message (which flashes up): "This configuration is not valid json, and will not be saved."
Screenshot: ![image](https://user-images.githubusercontent.com/3789646/76409817-d5165700-6386-11ea-8055-96ab07ba5e54.png)

### Steps to reproduce
1. Go to a grid data type settings
2. Under settings, click edit
3. Add some invalid JSON
4. Click "Submit"

### Expected result

Modal should not close - text box should retain changes.

### Actual result

Error flashes and hard work disappear's with the modal. 😥

SvenPam:
This issues was bugging me. Tried my best to follow the contrib guide - advanced apologies if this is totally wrong.

nul800sebastiaan:
Thanks @SvenPam - that PR looks quite alright to me, we'll have to have a proper look at it!

BatJan:
@SvenPam Looks like your PR for this issues has already been merged so closing this issue now - Thanks for your contribution 🙌🏻 🎉

PR #7790, opened 2020-03-11
#7789 Prevent grid config being submitted if not valid

SvenPam:
Prerequisites

    I have added steps to test this contribution in the description below

Fixes issue: umbraco#7789
Description

I have disabled the submit button until JSON is valid.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/editconfig.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/editconfig.html
@@ -5,54 +5,50 @@
         <form novalidate name="EditConfigForm" val-form-manager>
 
             <umb-editor-header
-                name="model.title"
-                name-locked="true"
-                hide-alias="true"
-                hide-icon="true"
-                hide-description="true">
+                               name="model.title"
+                               name-locked="true"
+                               hide-alias="true"
+                               hide-icon="true"
+                               hide-description="true">
             </umb-editor-header>
 
-            <umb-editor-container>
-                <umb-box>
-                    <umb-box-content>
-
-                        <ng-form name="gridConfigEditor">
-
+            <ng-form name="gridConfigEditor">
+                <umb-editor-container>
+                    <umb-box>
+                        <umb-box-content>
                             <h4>{{model.name}}</h4>
                             <p>Settings will only save if the entered json configuration is valid</p>
                             <textarea name="configSource"
-                                validate-on="'keyup'"
-                                rows="20" class="umb-property-editor umb-textarea"
-                                umb-raw-model="model.config"></textarea>
+                                      validate-on="'keyup'"
+                                      rows="20" class="umb-property-editor umb-textarea"
+                                      umb-raw-model="model.config"></textarea>
 
                             <div class="alert alert-error" ng-show="gridConfigEditor.$invalid === true">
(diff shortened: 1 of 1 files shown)


abjerner:
Thanks a lot @SvenPam - and congrats with your first PR for Umbraco 🎉 

I had a quick look at the code, and it seems to do the trick - but we'll have a extra look to make sure that it does 😉

matterlab-admin:
Think I butchered the template sorry!

SvenPam:
I butchered the template, sorry!

abjerner:
@SvenPam don't worry about that 😉 

You done a good job with describing what is wrong in the issue, which should be sufficient 👍

nul800sebastiaan:
Thanks @SvenPam - unfortunately this doesn't seem to work for me, I think it's because we check if the model is valid on `'blur'` but that might not be in time to prevent the button from being disabled?.
I am not sure how to fix this so I can't guide you unfortunately. Happy for better suggestions!

SvenPam:
Sorry for taking a while to get back - the world needs a PR I think!

I'm really sorry, I'm having to do this blind as I can't run the back office. I've followed the contribution guide as best I can with a fresh install, but get errors like: "No url found for api name updateCheckApiBaseUrl" as soon as I log in.

I've pushed a possible solution to what you have described. Validation of JSON should happen on keyup, which should disable the submit button in the code from my previous commit. I imagine the error message will appear as you type, but this is - in my opinion - is better than losing data when you try to save it.

nul800sebastiaan:
@SvenPam Alright, we need to fix your setup then, what's the problem in getting it to run the backoffice?

I think it's a TINY bit better, for sure, but unfortunately not good enough 🙈. We really just want to prevent the submit button from actually submitting anything until the validation errors are fixed. 

Let me know if I can help get your solution to run so you can do some proper debugging. In general it should be enough to run the solution from VS by pressing F5, you might want to go into the Task Runner Explorer and run the build task:

image

image

nul800sebastiaan:
Hi @SvenPam did you happen to try again and make some progress? Would be nice to wrap this one up!

SvenPam:
Hi @nul800sebastiaan! Hope you're safe and well.

The issue was server-side so Gulp, unfortunately, didn't help. Sorry that wasn't clear in my last comment. I've had another stab at this today, I deleted Umbraco temp data and reset the version and the API issue appeared went away, I guess it was an issue during installation.

I'm not familiar with Angular, so I hope this update is OK: it looks like I was referencing the state of an internal Angular component `ng-form` from outside its scope. It made sense to move the form scope higher and closer to the parent where the actual HTML form is. Subsequently, this solved the submit button not getting disabled as it was able to pick up `gridConfigEditor`.

SvenPam:
Just to clarify: I did get it working and was able to test in my browser this time around.

nul800sebastiaan:
Hi again @SvenPam - I seem to have forgotten to reply to this! 😅 

I think this doesn't really solve the problem unfortunately, it's quite annoying that on each keypress you get a validation error until you get the json to be valid. It's also a lot of overhead to keep checking that on each keypress. I think the real solution needs to be that we validate the json when pressing the Submit (should be "Save") button. 

I'll close this one for now but if you'd like to update this one then we'll be happy to give it another look! Thanks for the help so far! 👍

SvenPam:
Ah, swing and a miss! Thanks for your time, glad to be of help.

Issue #7803, opened 2020-03-13
Missing usergroup aliases


erikjanwestendorp:
The editor group and the writers group aliases are missing in [Constants-Security.cs](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Constants-Security.cs)

Not a big issue of course, but would be nice to have them there 👍

nul800sebastiaan:
@erikjanwestendorp Just so you know, when making a PR you don't have to link it to an issue. If an existing issue is present then of course link it, but no need to make a fresh issue for a new PR! 👍

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/7804

PR #7804, opened 2020-03-13
#7803 Add EditorGroupAlias and WriterGroupAlias to Constants class

erikjanwestendorp:
Fixes issue: #7803

**Description**
I have added EditorGroupAlias and WriterGroupAlias to Constants-Security.cs


nul800sebastiaan:
Thanks @erikjanwestendorp - but how about actually replacing hardcoded strings as well then? 😁

erikjanwestendorp:
@nul800sebastiaan Good point! I replaced the hardcoded strings as well! 😄

nul800sebastiaan:
Super, thanks very much @erikjanwestendorp - lovely update! 👍

PR #7806, opened 2020-03-13
GetAllIds for Roles

bielu:
### Prerequisites

Quite often you need a list of IDs instead of names for groups, but actually Umbraco didn't have to retrieve that using member service,  I added them to Umbraco as I think it could be actually helpful to have them instead of using directly repository


nul800sebastiaan:
Thanks @bielu - unfortunately you've made it really hard to do a compare by reformatting the whole document. Could you please undo all the formatting changes so we can actually see what changed?

bielu:
@nul800sebastiaan reverted :)

nul800sebastiaan:
Thanks @bielu - all good! 👍

PR #7810, opened 2020-03-16
NetCore: Refactor config/setting to not use dependencies

bergmania:
This PR refactor config/settings, to avoid using dependencies in Config/Settings.

For connectionstings:
- Changed to method injection of `IIOHelper` on the two methods to save and remvoe the connectionstring

For ModelsBuilder
- Handle the replace of `~\` with the virual path of the app outside of the settings

For GlobalSettings
- Moved the Path to IIOHelper


src/Umbraco.Configuration/ConfigsFactory.cs
```diff
@@ -61,7 +58,6 @@ public Configs Create(IIOHelper ioHelper, ILogger logger)
             configs.Add<IMemberPasswordConfiguration>(() => MemberPasswordConfigurationSettings);
             configs.Add<IContentSettings>(() => ContentSettings);
 
-            configs.AddCoreConfigs(ioHelper);
```
elit0451: Is it that for the moment we don't add the `GridConfig`?

src/Umbraco.Configuration/GlobalSettings.cs
```diff
@@ -227,7 +214,7 @@ public string ConfigurationStatus
             }
             set
             {
-                SaveSetting(Constants.AppSettings.ConfigurationStatus, value, _ioHelper);
+                SaveSetting(Constants.AppSettings.ConfigurationStatus, value, Current.IOHelper); //TODO remove
```
elit0451: Is this TODO for later use?

src/Umbraco.Core/Configuration/GlobalSettingsExtensions.cs
```diff
@@ -28,14 +28,15 @@ public static string GetUmbracoMvcArea(this IGlobalSettings globalSettings, IIOH
             return _mvcArea;
         }
 
+        //TODO Move to IOHelper
         internal static string GetUmbracoMvcAreaNoCache(this IGlobalSettings globalSettings, IIOHelper ioHelper)
```
elit0451: `globalSettings` is no longer needed and can be safely removed

src/Umbraco.Configuration/GlobalSettings.cs
bergmania: This is a class that will be removed in the future.. I thing the TODO is fine for now

src/Umbraco.Configuration/ConfigsFactory.cs
bergmania: Yes.. `IGridConfig` is not a real config file as the rest.. It is just registered in a normal way.

PR #7811, opened 2020-03-16
Puts in place the code to 'boot' Umbraco Core in netcore

Shazwazza:
(Draft PR)

* New netcore integration test project
* New netstandard Umbraco.Tests.Common class library for shared resources between test projects and moves a bunch of mocks/stub stuff there
* New ext methods for startup
* LightInject cross-wire container (we can decide more on DI later, for now this is easiest). This is done with host builder extensions and following LI's own procedures but tweaked for our own use


src/Umbraco.Tests.Integration/RuntimeTests.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            // MSDI
+            var services = new ServiceCollection();
+            // These services are required
+            services.AddSingleton<IHttpContextAccessor>(x => testHelper.GetHttpContextAccessor());
+            services.AddSingleton<IWebHostEnvironment>(x => testHelper.GetWebHostEnvironment());
+            services.AddSingleton<IHostApplicationLifetime>(x => Mock.Of<IHostApplicationLifetime>());
+
+            // LightInject / Umbraco
+            var container = new ServiceContainer(ContainerOptions.Default.Clone().WithMicrosoftSettings().WithAspNetCoreSettings());
```
bergmania: Maybe we need TestHelper.GetServiceContainer..

`ServiceContainer(ContainerOptions.Default.Clone().WithMicrosoftSettings().WithAspNetCoreSettings());` is repeated over again and again

src/Umbraco.Web.BackOffice/AspNetCore/AspNetCoreSessionIdResolver.cs
```diff
@@ -12,6 +13,17 @@ public AspNetCoreSessionIdResolver(IHttpContextAccessor httpContextAccessor)
             _httpContextAccessor = httpContextAccessor;
         }
 
-        public string SessionId => _httpContextAccessor?.HttpContext.Session?.Id;
+        
+        public string SessionId
+        {
+            get
+            {
+                // If session isn't enabled this will throw an exception so we check
```
bergmania: We should save the httpcontext in a variable here, instead of accessing twise..

src/Umbraco.Web.BackOffice/AspNetCore/UmbracoBackOfficeServiceCollectionExtensions.cs
```diff
@@ -30,53 +42,116 @@ public static IServiceCollection AddUmbracoConfiguration(this IServiceCollection
...
 
-            var httpContextAccessor = serviceProvider.GetService<IHttpContextAccessor>();
-            var webHostEnvironment = serviceProvider.GetService<IWebHostEnvironment>();
-            var hostApplicationLifetime = serviceProvider.GetService<IHostApplicationLifetime>();
+            // TODO: Get rid of this 'Current' requirement
+            var globalSettings = Current.Configs.Global();
+            var umbracoVersion = new UmbracoVersion(globalSettings);
 
-            var configs = serviceProvider.GetService<Configs>();
+            // TODO: Currently we are not passing in any TypeFinderConfig (with ITypeFinderSettings) which we should do, however
```
bergmania: After the changes to config, this should be possible now

src/Umbraco.Web.BackOffice/AspNetCore/UmbracoBackOfficeServiceCollectionExtensions.cs
```diff
@@ -30,53 +42,116 @@ public static IServiceCollection AddUmbracoConfiguration(this IServiceCollection
...
             var globalSettings = configs.Global();
 
-            var hostingEnvironment = new AspNetCoreHostingEnvironment(hostingSettings, webHostEnvironment,
-                httpContextAccessor, hostApplicationLifetime);
+            var hostingEnvironment = new AspNetCoreHostingEnvironment(hostingSettings, webHostEnvironment, httpContextAccessor, hostApplicationLifetime);
             var ioHelper = new IOHelper(hostingEnvironment, globalSettings);
             var logger = SerilogLogger.CreateWithDefaultConfiguration(hostingEnvironment,
                 new AspNetCoreSessionIdResolver(httpContextAccessor),
-                () => services.BuildServiceProvider().GetService<IRequestCache>(), coreDebug, ioHelper,
+                () => serviceProvider.GetService<IRequestCache>(), coreDebug, ioHelper,
```
bergmania: I don't think the `serviceProvider`, build at that time, has a `IRequestCache` availble, that why it was done the other way.

src/Umbraco.Web.BackOffice/AspNetCore/UmbracoBackOfficeServiceCollectionExtensions.cs
```diff
@@ -30,53 +42,116 @@ public static IServiceCollection AddUmbracoConfiguration(this IServiceCollection
...
                 new AspNetCoreMarchal());
 
-            var backOfficeInfo = new AspNetCoreBackOfficeInfo(globalSettings);
+            var backOfficeInfo = new AspNetCoreBackOfficeInfo(configs.Global());
```
bergmania: Why not use the `globalSettings` variable?

src/Umbraco.Web.UI.NetCore/Program.cs
```diff
@@ -14,7 +15,9 @@ public static void Main(string[] args)
         }
 
         public static IHostBuilder CreateHostBuilder(string[] args) =>
-            Host.CreateDefaultBuilder(args)
-                .ConfigureWebHostDefaults(webBuilder => { webBuilder.UseStartup<Startup>(); });
+            Host.CreateDefaultBuilder(args)                
+                .ConfigureWebHostDefaults(webBuilder => { webBuilder.UseStartup<Startup>(); })
+                .UseUmbraco()
+                .UseSerilog();
```
bergmania: I think we should do the `UseSerilog` inside `UseUmbraco`..

Shazwazza: How do you mean by this? This is Serilog's own setup requirement, see https://nblumhardt.com/2019/10/serilog-in-aspnetcore-3/ and https://github.com/serilog/serilog-aspnetcore

bergmania: By calling `builder.UseSerilog` in `HostBuilderExtensions.UseUmbraco` before returning.. 

Serilog is a dependency due to Umbraco, so I think it is better we set it up, other then require all solutions to do the same

src/Umbraco.Web.BackOffice/AspNetCore/UmbracoBackOfficeServiceCollectionExtensions.cs
Shazwazza: Ok, i will make this change in the next PR since I'm refactoring a little bit of this already.

src/Umbraco.Web.BackOffice/AspNetCore/UmbracoBackOfficeServiceCollectionExtensions.cs
Shazwazza: ah right I see. I'll change back. I've added a couple more notes in a newer branch about the whole BuildServiceProvider() while we are building the container. I'm not convinced this will work correctly because it will end up being different instances of a service that is supposed to be a singleton. The main problem is going to be `IHostApplicationLifetime` I think. Each time you call `BuildServiceProvider` and then resolve a service from that, even if the service is registered as a singleton, it will be a new instance for each service provider created which is really not what we want to have happen. I don't think we can solve this in this PR but we'll need to considering this going forward.

src/Umbraco.Web.UI.NetCore/Program.cs
Shazwazza: Right i see. OK i'll remove this for now and create a new task for that.

PR #7813, opened 2020-03-16
NetCore: Config implementations for .NET Core

bergmania:
## Summary
Implemetation of config/settings interfaces in .NET Core using `AppSettings.json` (Default)

## Details
- Introduced an AspNetCoreConfigsFactory, implemting the IConfigsFactory
  - All settings are implemented to support hot-swap of settings (Without recycle)
- Seperated the actual implementation of the config interfaces from the sections (the old implementation)
- Move the legacy stuff into a Legacy folder/namespace
- Removed some unused legacy config stuff
- 
## Questions
We need to handle the following:
- Debug
  - Debug mode cannot be retrived from the host. In current implementation I have introcuced our own debug flag
- ConnectionStrings
  - We need a provider name. We can vary from best practice having an object in the `"ConnectionStrings"` section instead of a string? Or we could try to use the name/key of the connectionstring to lookup the provider elsewhere in the config?
- Machine Key
  - We need the implement this the new way:  https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/compatibility/replacing-machinekey?view=aspnetcore-3.1
- Config modifications
  - In general we can't save values back into the config. Today we do this for ConnectionStrings and ConfigurationVersion.
  - We could look for the first `JsonConfigurationProvider` and if we find one, we can get the filepath from that before we update? But then throw if we cannot find a  `JsonConfigurationProvider`?


src/Umbraco.Configuration/Models/ConnectionStrings.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public ConfigConnectionString this[string key]
+        {
+            get => new ConfigConnectionString(_configuration.GetConnectionString(key), "sasd", key);
+            set => throw new NotImplementedException();
+        }
+
+        public void RemoveConnectionString(string umbracoConnectionName, IIOHelper ioHelper)
+        {
+            //TODO We need to figure out what to do here.. We cond have another config setting, that tells with file(s) to update? or should we assume appsettings.json
```
emmagarland: Agreed this is a bigger question - maybe one for a wider GitHub ticket/RFC?

src/Umbraco.Configuration/Models/ContentSettings.cs
```diff
@@ -0,0 +1,89 @@
...
+            public Guid ContentKey { get; set;}
+            public string ContentXPath { get; set;}
+            public bool HasContentId { get; set;}
+            public bool HasContentKey { get; set;}
+            public string Culture { get; set; }
+        }
+
+        private class ImagingAutoFillUploadField : IImagingAutoFillUploadField
+        {
+            public string Alias { get; set; }
```
emmagarland: Just checking for my own understanding - the decision for both these classes ImagingAutoFillUploadField and ContentErrorPage to be nested classes within the Internal ContentSettings class, is this to control the usage to be internal to ContentSettings?

src/Umbraco.Configuration/Models/HealthChecksSettingsSettings.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Configuration;
+using Umbraco.Core.Configuration.HealthChecks;
+
+namespace Umbraco.Configuration.Models
+{
+    internal class HealthChecksSettingsSettings : IHealthChecksSettings
```
emmagarland: Low priority: naming comment as above

src/Umbraco.Configuration/Models/MachineKeyConfig.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public class MachineKeyConfig : IMachineKeyConfig
+    {
+        private readonly IConfiguration _configuration;
+        public MachineKeyConfig(IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        //TODO all the machineKey stuff should be replaced: https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/compatibility/replacing-machinekey?view=aspnetcore-3.1
```
emmagarland: New GitHub issue to open if not already?

src/Umbraco.Configuration/Models/RequestHandlerSettings.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            _configuration = configuration;
+        }
+
+        public bool AddTrailingSlash   => _configuration.GetValue<bool?>("Umbraco:CMS:RequestHandler:AddTrailingSlash") ?? true;
+        public bool ConvertUrlsToAscii   => _configuration.GetValue<string>("Umbraco:CMS:RequestHandler:ConvertUrlsToAscii").InvariantEquals("true");
+        public bool TryConvertUrlsToAscii   => _configuration.GetValue<string>("Umbraco:CMS:RequestHandler:ConvertUrlsToAscii").InvariantEquals("try");
+
+
+        //We need to special handle ":", as this character is special in keys
```
emmagarland: (minor) Just to clarify, are we removing the : from the character collection as it is used as an identifier? If so, it might be good to mention in the comments e.g. "Remove instances of the ':' character, because...".

src/Umbraco.Configuration/Models/UserPasswordConfigurationSettings.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        private readonly IConfiguration _configuration;
+        public UserPasswordConfigurationSettings(IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        public int RequiredLength => _configuration.GetValue<int?>("Umbraco:CMS:Security:UserPassword:RequiredLength") ?? 10;
+        public bool RequireNonLetterOrDigit =>  _configuration.GetValue<bool?>("Umbraco:CMS:Security:UserPassword:RequireNonLetterOrDigit") ?? false;
+        public bool RequireDigit => _configuration.GetValue<bool?>("Umbraco:CMS:Security:UserPassword:RequireDigit") ?? false;
```
emmagarland: Side note, MemberPasswordConfigurationSettings very similar to this now, noted for the user/membership work

src/Umbraco.Configuration/ConfigsFactory.cs
```diff
@@ -27,7 +29,7 @@ public class ConfigsFactory : IConfigsFactory
         public IMemberPasswordConfiguration MemberPasswordConfigurationSettings { get; } = new MemberPasswordConfigurationSettings();
         public IContentSettings ContentSettings { get; } = new ContentSettings();
         public IGlobalSettings GlobalSettings { get; } = new GlobalSettings();
-        public IHealthChecks HealthChecksSettings { get; } = new HealthChecksSettings();
+        public IHealthChecksSettings HealthChecksSettingsSettings { get; } = new HealthChecksSettingsSettings();
```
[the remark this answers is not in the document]
emmagarland: Just seen it is because of the other file named HealthCheckSettings

src/Umbraco.Web.UI.Client/src/views/components/content/umb-content-node-info.html
```diff
@@ -26,7 +26,7 @@
             <umb-box-header title-key="redirectUrls_redirectUrlManagement"></umb-box-header>
             <umb-box-content class="block-form">
                 <div style="position: relative;">
-                    <div ng-if="loadingRedirectUrls" style="background: rgba(255, 255, 255, 0.8); position: absolute; top: 0; left: 0; right: 0; bottom: 0;"></div>
+                    <div ng-if="loadingRedirectUrls" style="background: rgba(255: 255, 255, 0.8); position: absolute; top: 0; left: 0; right: 0; bottom: 0;"></div>
```
emmagarland: Is the colon replacing the comma for a reason? (not sure if this is something new I am unaware of)

src/Umbraco.Configuration/Models/GlobalSettings.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace Umbraco.Configuration.Models
+{
+    /// <summary>
+    /// The GlobalSettings Class contains general settings information for the entire Umbraco instance based on information from  web.config appsettings
+    /// </summary>
+    public class GlobalSettings : IGlobalSettings
+    {
+        private readonly IConfiguration _configuration;
+
+        internal const string StaticReservedPaths = "~/app_plugins/,~/install/,~/mini-profiler-resources/,"; //must end with a comma!
```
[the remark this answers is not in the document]
bergmania: I Just took over the existing config settings. I guess we could change it to an array later when we only has to support the new config implementation

src/Umbraco.Configuration/ConfigsFactory.cs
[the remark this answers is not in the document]
bergmania: No, that was a refacroting mistake.. Thanks

src/Umbraco.Configuration/Models/ContentSettings.cs
bergmania: The only reason is because they are only used by `ContentSettings` and everything people should care about is available on the interface..

src/Umbraco.Configuration/Models/RequestHandlerSettings.cs
bergmania: No the issue is, if/when we use ":" as key in the json-config file. it is not understood by the framework.

There are two alternatives
1. Special handle `":"` and always replace with `""`
1. Implement the config more verbose, such that we don't use `":"` as key.
E.G
```
{
    "Umbraco": {
        "CMS": {
            "RequestHandler": {
                "AddTrailingSlash": true,
                "CharCollection": [
                    {"Char": ":", "Replacement": ""},
                    {"Char": "\"", "Replacement": ""},
                    {"Char": "'", "Replacement": ""},
                    {"Char": "%", "Replacement": ""},
                   ...
                ]
            }
        }
    }
}
```
instead of the current
```
{
    "Umbraco": {
        "CMS": {
            "RequestHandler": {
                "AddTrailingSlash": true,
                "CharCollection": {
                    ":": "",
                    "\"":"",
                    "'":"",
                    "%":"",
                   ...
                }
            }
        }
    }
}
```

src/Umbraco.Web.UI.Client/src/views/components/content/umb-content-node-info.html
bergmania: Nope, good catch 👍

src/Umbraco.Configuration/Models/GlobalSettings.cs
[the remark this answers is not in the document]
emmagarland: Yes - seems like a good task to do

bergmania:
### Update:
Regarding config updates, I implemented a version that takes the first (lowest priority) JSON config provider, and update the json file asociated with this provider.

So, in the default setup this updates `appSettings.json`.

src/Umbraco.Configuration/Models/ContentSettings.cs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        public string NotificationEmailAddress =>
+            _configuration.GetValue<string>("Umbraco:CMS:Content:Notifications:Email");
+
+        public bool DisableHtmlEmail =>
+            _configuration.GetValue("Umbraco:CMS:Content:Notifications:DisableHtmlEmail", false);
+
+        public IEnumerable<string> ImageFileTypes => _configuration.GetValue(
+            "Umbraco:CMS:Content:Imaging:ImageFileTypes", new[] { "jpeg", "jpg", "gif", "bmp", "png", "tiff", "tif" });
```
Shazwazza: Its a bit of work but each of these prefix parts be constants in ase we need to change them?

src/Umbraco.Configuration/Models/MachineKeyConfig.cs
Shazwazza: We should remove Machine Key config entirely, there is no need for this at all in aspnetcore for our needs.

src/Umbraco.Configuration/Models/MemberPasswordConfigurationSettings.cs
```diff
@@ -0,0 +1,39 @@
...
+        public bool RequireDigit =>
+            _configuration.GetValue("Umbraco:CMS:Security:MemberPassword:RequireDigit", false);
+
+        public bool RequireLowercase =>
+            _configuration.GetValue("Umbraco:CMS:Security:MemberPassword:RequireLowercase", false);
+
+        public bool RequireUppercase =>
+            _configuration.GetValue("Umbraco:CMS:Security:MemberPassword:RequireUppercase", false);
+
+        public bool UseLegacyEncoding =>
```
Shazwazza: This needs to be removed

src/Umbraco.Configuration/Models/MemberPasswordConfigurationSettings.cs
```diff
@@ -0,0 +1,39 @@
+﻿using Microsoft.Extensions.Configuration;
+using Umbraco.Core.Configuration;
+
+namespace Umbraco.Configuration.Models
+{
+    internal class MemberPasswordConfigurationSettings : IMemberPasswordConfiguration
```
Shazwazza: We'll prob need to modify these settings to be inline with identity, but we can deal with that later.

src/Umbraco.Configuration/Models/RequestHandlerSettings.cs
Shazwazza: I would prefer that we don't need to hack this with special chars and rules. Although the Char/Replacement is more verbose it doesn't need any hacks so I vote for that.

src/Umbraco.Configuration/Models/RuntimeSettings.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    internal class RuntimeSettings : IRuntimeSettings
+    {
+        private readonly IConfiguration _configuration;
+        public RuntimeSettings(IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        public int? MaxQueryStringLength => _configuration.GetValue<int?>("Umbraco:CMS:Runtime:MaxRequestLength");
```
Shazwazza: Not sure where we even use these in the CMS? seems like we don't need these.

src/Umbraco.Configuration/Models/UserPasswordConfigurationSettings.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public bool RequireDigit => _configuration.GetValue("Umbraco:CMS:Security:UserPassword:RequireDigit", false);
+
+        public bool RequireLowercase =>
+            _configuration.GetValue("Umbraco:CMS:Security:UserPassword:RequireLowercase", false);
+
+        public bool RequireUppercase =>
+            _configuration.GetValue("Umbraco:CMS:Security:UserPassword:RequireUppercase", false);
+
+        public bool UseLegacyEncoding =>
```
Shazwazza: similar to the member config, this should be removed

src/Umbraco.Configuration/Models/WebRoutingSettings.cs
```diff
@@ -0,0 +1,40 @@
...
+    internal class WebRoutingSettings : IWebRoutingSettings
+    {
+        private readonly IConfiguration _configuration;
+
+        public WebRoutingSettings(IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        public bool TrySkipIisCustomErrors =>
```
Shazwazza: Don't think this is going to make a lot of sense? ... maybe something to revisit later on

bergmania: Agree.. This in just to have the same interfaces as before, for now.. Make sense to clean up when we don't have to care about the .NET Framework executable

src/Umbraco.Configuration/Models/RuntimeSettings.cs
bergmania: Currently used by ClientDependency

Issue #7777, opened 2020-03-08
Factory Extensions CreateInstance with arguments doesn't work correctly


tompipe:
When registering a service with an implmentation factory using the below extension method

https://github.com/umbraco/Umbraco-CMS/blob/a0aad045fb7267650b4330c26d391d1788cec7fd/src/Umbraco.Core/FactoryExtensions.cs#L63-L89

It doesn't seem to work correctly if multiple parameters of the same type are passed through. The FirstOrDefault on line 85 seems to be the issue, it's passing through the value of the first parameter for all subsequent parameters of the same type

To reproduce, define and register a new service with an implementation factory:

```C#
composition.Register<ITestService>(factory => 
{   
    var param1 = "param1";
    var param2 = "param2";

   return factory.CreateInstance<TestService>(param1, param2);
});

public class TestService : ITestService
{
    public TestService(UmbracoHelper umbraco, string param1, string param2)
    {
         Debug.Assert(param1.Equals(param2));
    }
}
```

I can work around it by manually fetching each dependency from the factory instead, creating the service via new, rather than calling CreateInstance, but I would have expected the values passed through via the CreateInstance method to match the arguments correctly. 

## Umbraco version

I am seeing this issue on Umbraco version: 8.4.0

nul800sebastiaan:
Not sure what to do with this one, the code has been like this since 8.0.0 and I am not sure what the impact would be of changing it. Hopefully that can be done in a backwards compatible way! I'll have to ask around. Will get back to you!

Shazwazza:
If it can be changed to support multiple parameters of the same type that is fine but in order to do that this will require unit tests since there currently are none... well, sort of. There is ContainerConformingTests.CanRegisterSingletonWithCreate which tests the generic version of this method which does call this method. So if this is changed to support more scenarios then this test needs to be revamped to test all cases or another one created.

I've marked this up for grabs.

umbrabot:
Hi @tompipe,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

[...]

PR #7814, opened 2020-03-16
Makes sure supplied arguments cannot be used more than once.

dknippers:
### Prerequisites
PR for #7777 to discuss a possible solution.

### Description
This commit simply makes sure none of the supplied arguments can be used more than once, which I believe is the issue here.


emmary2020:
Hey @dknippers 

Thanks for the PR. Much appreciated. We'll take a look and let you know if we've any questions. 

All the best,

Em

dknippers:
Hi @nul800sebastiaan,

You register it with the container as an `IThing`, but you later request a `Thing4` which is not registered. So this crash makes sense. You have 2 options:

1) Request an instance of `IThing` (it will not have `.Thing` and `.Thing2` properties though).
2) Register it as a `Thing4`

Regardless, the test should be done with 2 parameters of the same type. @tompipe used 2 parameters of type `string` in his issue, for example. In your code it uses `Thing1` and `Thing2`, which are not the same type so the Assert will always be true, even without this PR.

I have pushed a version of `CanRegisterSingletonWithCreate` to this PR that will fail without this PR and succeed after this PR. In there, I use the same type for both parameters (`Thing1`). That is what the issue was about (multiple parameters of the same type, but all parameters get set to the first value of that type).

Let me know if that unit test was helpful, or if you wanted to test something else!

nul800sebastiaan:
Duh.. well that makes total sense now, thanks!

Yay, it fails! And is fixed with this PR. 👍 

```csharp
        [Test]
        public void CanRegisterMultipleSameTypeParametersWithCreateInstance()
        {
            var register = GetRegister();
            var factory = register.CreateFactory();
            register.Register<Thing4>(factory =>
            {
                var param1 = "param1";
                var param2 = "param2";

                return factory.CreateInstance<Thing4>(param1, param2);
            });
            var instance = factory.GetInstance<Thing4>();
            Assert.AreNotEqual(instance.Param1, instance.Param2);
        }
```

image

nul800sebastiaan:
Ah I see you already updated the tests @dknippers - wonderful! ⭐ 

Great to have tried it myself as well as I saw the error in me trying to use 2 params of a different type, which was not the problem!

nul800sebastiaan:
I've [moved the test into a separate test and used the strings for a clearer error](https://github.com/umbraco/Umbraco-CMS/pull/7814/commits/558ffaeba70e922240a124a47faa1718a8f3e0be), this error now is a bit confusing: 

```
Umbraco.Tests.Composing.ContainerConformingTests.CanRegisterMultipleSameTypeParametersWithCreateInstance

  Expected: not equal to <Umbraco.Tests.Composing.ContainerConformingTests+Thing1>
  But was:  <Umbraco.Tests.Composing.ContainerConformingTests+Thing1>
```

Other than that, this is great, thanks very much @dknippers!

PR #7819, opened 2020-03-17
v8: Bug fix/make groups to use paginated list of users

### LINKED ISSUE #6112: V7 User groups display all assigned users
In Umbraco back office groups, we are displaying all users and for websites with a large number of content editors it takes ages to load:

Here is an example from a user group where we have around 700 users:
image

The 30s are a still optimistic situation I had already few times about a few minutes after a fresh run of Umbraco.

bielu:
### Prerequisites

If there's an existing issue for this PR then this fixes #6112 
To see the differences you need:
Add more than 10 users into 1 user group and you should be to see newly paginated users list like here:
image
This will actually prevent really big performance degradation described in the issue.

I will backport that for v7 as soon @nul800sebastiaan will confirm it will be accepted for v7.

### DIFF
--- a/src/Umbraco.Core/Models/Membership/IUserGroup.cs
+++ b/src/Umbraco.Core/Models/Membership/IUserGroup.cs
@@ -20,6 +20,8 @@ namespace Umbraco.Core.Models.Membership
         /// </summary>
         string Name { get; set; }
 
+        int MaxUsers { get; set; }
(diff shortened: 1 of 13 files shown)


emmary2020:
Hi @bielu 

Thanks for the PR. If we need anything or have any questions, we'll be sure to get in touch. Speak soon,

Emma

PR #7827, opened 2020-03-20
remove some redundant usings

SimonCropp:


nul800sebastiaan:
Hi @SimonCropp 👋! I love your enthusiasm and wouldn't want to put up any barriers for contributing.. however.. 😅 

We unfortunately can't merge PRs like this where there's an enormous sweep over our entire code base. The biggest problem here is that this specific PR will cause many unnecessary merge conflicts along the way. Other than that, we love a bit of cleanup, but this doesn't actually change a single thing for the codebase itself. Unused usings are just.. "unused", not causing any problems. 

We do fully support the "scout rule": "Leave your code better than you found it." - just like you would learn if you were part of the scouts, always to leave your camp site cleaner than you found it. So if you're actually making improvements to one part of a class, then clean up around you as well.

I'm going to close this broad sweep PR for now, sorry if you put a lot of energy in it! 🙈 

There's [a list of issues that are up for grabs](https://github.com/umbraco/Umbraco-CMS/issues?q=is%3Aissue+is%3Aopen+label%3Acommunity%2Fup-for-grabs) if you want to dig in and help fix something!

PR #3460, opened 2018-10-27
EntityRepository is fetching all property data for media

clausjensen:
connect #3457 

### Changes

- Adds internal methods for getting media entities through the entity repository/service without fetching all property data.
- Ensures the trees use this method to avoid fetching property data in the media tree.

This is not the prettiest way to do it since it requires casting of the service/repo to use the internal methods. It is however a way of preventing this becoming a breaking change in the APIs.
This should be fixed when merged to V8, to we make sure we do not do this in the `EntityRepository` for media items.

### Test

- Ensure that the content and media tree renders as expected (they render nodes without throwing errors).
- Ensure that the "folder browser" feature on media folders still work (this feature may be the reason why this property data was added initially? - and it should still work fine since this feature uses paging)

PR #7837, opened 2020-03-24
Netcore: Use API to get Image URLs in BackOffice instead of hardcoded querystrings

AndyButland:
Implements #7540

In this PR I've created an API controller that can be called from the back-office to retrieve the URL to use to request an image with various processing parameters (resize, crops etc.).

This means that rather than using hard-coded values that will only work for image processor, we can use the abstraction introduced in #7539 to get an appropriate URL for whatever image processing provider is in use.

I've also done a little refactoring of `IImageUrlGenerator` and it's implementation - just to choose what seemed a more appropriate namespace, and to make some null/empty value checks more complete/explicit.

_Caveats:_ although 4 places have been identified where hard-coded URLs were in place, and all have been updated, I've only been able to test one - the one added to the tinymce controller.  For the others I simply can't find what triggers them!  So this needs identifying and testing before merging.  The implementation can still be reviewed though.


src/Umbraco.Web.UI.Client/src/views/propertyeditors/fileupload/fileupload.controller.js
```diff
@@ -44,15 +44,22 @@
...
                         if (mediaHelper.detectIfImageByExtension(property.value)) {
-                            //get default big thumbnail from image processor
-                            var thumbnailUrl = property.value + "?rnd=" + moment(entity.updateDate).format("YYYYMMDDHHmmss") + "&width=500&animationprocessmode=first";
-                            return thumbnailUrl;
+
+                            // Get default big thumbnail from image processor
+                            return mediaHelper.getProcessedImageUrl(property.value,
+                                {
+                                    animationprocessmode: "first",
+                                    cacheBusterValue: moment(entity.updateDate).format("YYYYMMDDHHmmss"),
```
Shazwazza: I feel like it would be better for the server to auto-cache bust this than the responsibility to be on the client?

Shazwazza:
(also need to resolve conflicts)

src/Umbraco.Web.UI.Client/src/views/propertyeditors/fileupload/fileupload.controller.js
AndyButland: I kind of agree.  But it's only really the client here that knows the entity's update date, and hence what the appropriate cache buster value is.  By the time the request gets to the server it's just a path to an image file.  We could use a random value cache buster at the server, but then we'd be doing it far more than we want to (i.e. if the entity hasn't changed, we want to use the cache if it's available.

AndyButland:
Merge conflicts resolved.

AndyButland:
I've made a couple of updates, and have verified that two of the instances are working correctly: in `media.controller.js` and `Tinymce.service.js`.

The other two - in `fileupload.controller.js` and `imagecropper.controller.js` - both are called within a function provided to `mediaHelper.registerFileResolver`.  I haven't been able to find how to trigger this code to test what I've done, but am pretty sure I haven't got it quite right, given that previously the call returned a string, and now in some routes, it needs to get that string following an HTTP call - so via a promise.  As such I think it'll need someone who knows JavaScript and the Umbraco implementation a bit better than me to finish this up.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/fileupload/fileupload.controller.js
Shazwazza: ok no worries then :)

Shazwazza:
IIRC `mediaHelper.registerFileResolver` is a legacy thing and we should have removed it in v8. I found a link of myself saying to do so here: https://github.com/umbraco/Umbraco-CMS/pull/3460#issuecomment-434903930 haha :) 

I think we should just remove that and then it's less to worry about here. Else the way to trigger it is to register your own thing but i can't actually remember how to do that or why, there's prob some comments in the code for it but like i said i think it should just be removed... it is clearly not documented for use anyways.

AndyButland:
OK, have removed.  Have checked both property editors still function as expected.

Shazwazza:
great stuff!

PR #7838, opened 2020-03-24
Netcore: Smidge implementation

elit0451:
### Details
- Introduced Umbraco abstractions for the concepts for runtime minification
  - Introduced the interface `IRuntimeMinifier` and implemented that for both _ClientDependency_ and _Smidge_
  - and concepts like `IAssetFile`, `AssetFile`, `AssetType`, `CssFile`, `JavaScriptFile`,...
 - Moved some of the `CompositionExtensions` to boot the SmidgeRuntimeMinifier in .net core. (These will be cleaned up in another task, which will also compose more classes.)
- Introduced a new `SaveConfigValue` on `IConfigManipulator` and only implemeneted for Json, as this is not used in framework (xml)
- Moved some of the common parts into `Umbraco.Infrastracure.Javascript`
  - `Resources.resx`, `AssetInitialization`, `CssInitialization`,  `JsInitialization`, the `*.js` files and a `JavaScriptHelper` (I really dont have a better name for this collection of helper methods)
- Introduced a simple `BackOfficeController` in .Net Core, and moved/reimplemented the `MinifyJavaScriptResultAttribute`
- Updated gulp stuff to copy the client files into multiple roots, and now inserts into both executables.

Other notes:
- Reset in Smidge, basically just updates the `Umbraco:CMS:RuntimeMinification:Version` config value. Smidge uses this value as cache buster
- JS minification currently uses `JsMinifier` and not `NuglifyJS` due to issues in our JS code
- In general please review that async/await is used correct.

### Test:
- Verify backoffice still works in both debug mode and not, when running on full framework.
- On the .NET Core executable, test the following endpoint `/umbraco/application`
  - When in debug mode, the endpoint should return 
```javascript
LazyLoad.js([
"/umbraco/lib/jquery/jquery.min.js",
"/umbraco/lib/jquery-ui/jquery-ui.min.js",
"/umbraco/lib/jquery-ui-touch-punch/jquery.ui.touch-punch.min.js",
"/umbraco/lib/angular/angular.js",
"/umbraco/lib/underscore/underscore-min.js",
"/umbraco/lib/moment/moment.min.js",
"/umbraco/lib/flatpickr/flatpickr.js",
"/umbraco/lib/animejs/anime.min.js",
"/umbraco/lib/angular-route/angular-route.js",
"/umbraco/lib/angular-cookies/angular-cookies.js",
"/umbraco/lib/angular-aria/angular-aria.min.js",
"/umbraco/lib/angular-touch/angular-touch.js",
"/umbraco/lib/angular-sanitize/angular-sanitize.js",
"/umbraco/lib/angular-animate/angular-animate.js",
"/umbraco/lib/angular-messages/angular-messages.js",
"/umbraco/lib/angular-ui-sortable/sortable.js",
"/umbraco/lib/angular-dynamic-locale/tmhDynamicLocale.min.js",
"/umbraco/lib/ng-file-upload/ng-file-upload.min.js",
"/umbraco/lib/angular-local-storage/angular-local-storage.min.js",
"/umbraco/lib/chart.js/chart.min.js",
"/umbraco/lib/angular-chart.js/angular-chart.min.js",
"/umbraco/lib/umbraco/Extensions.js",
"/umbraco/lib/umbraco/NamespaceManager.js",
"/umbraco/lib/umbraco/LegacySpeechBubble.js",
"/umbraco/js/app.js",
"/umbraco/js/umbraco.resources.js",
"/umbraco/js/umbraco.directives.js",
"/umbraco/js/umbraco.filters.js",
"/umbraco/js/umbraco.services.js",
"/umbraco/js/umbraco.interceptors.js",
"/umbraco/js/umbraco.controllers.js",
"/umbraco/js/routes.js",
"/umbraco/js/init.js"], function () {
    //we need to set the legacy UmbClientMgr path
    if ((typeof UmbClientMgr) !== "undefined") {
        UmbClientMgr.setUmbracoPath('/umbraco');
    }

    jQuery(document).ready(function () {

        angular.bootstrap(document, ['umbraco']);

    });
});
```
  - When not in debug mode, the returned file should be something like this
```javascript
LazyLoad.js(["/sc/80d3896f.fb8e5b2a.daf64d0d.63dc80f8.f142a940.d1aa3809.aa3c40bc.c53cdd84.5f4a2348.ddf7f64a.c3fc92ad.da0e5fc7.717925a2.512fe428.e1519986.527d537f.a139e144.32a48b01.33d92ffb.de682689.e12dee6b.08d18f13.fc8a39cc.46c812c6.f48f9c37.bee60e6f.5058ea9f.40b550d3.6d7fdc6f.cd415aa0.bfa798db.a176e6b8.c980e938.js.v637212411755687059"],function(){if((typeof UmbClientMgr)!=="undefined"){UmbClientMgr.setUmbracoPath('/umbraco');}
jQuery(document).ready(function(){angular.bootstrap(document,['umbraco']);});});;
```
- Please follow the js file and ensure it exists (Note that the file length is too long for IISExpress - Use the build in server)


src/Umbraco.Web.UI/Umbraco/Views/Default.cshtml
```diff
@@ -17,11 +16,17 @@
             isDebug = true;
         }
     }
+    var runtimeMinifier = Current.RuntimeMinifier;
+
+    runtimeMinifier.RequiresCss("assets/css/umbraco.css", "Umbraco");
+    runtimeMinifier.RequiresCss("lib/bootstrap-social/bootstrap-social.css", "Umbraco");
+    runtimeMinifier.RequiresCss("assets/css/umbraco.css", "Umbraco");
```
elit0451: We have that one on ln 21
```suggestion

```

src/Umbraco.Web.UI/Umbraco/PartialViewMacros/Templates/EditProfile.cshtml
```diff
@@ -10,15 +9,19 @@
 
     Html.EnableClientValidation();
     Html.EnableUnobtrusiveJavaScript();
-    Html.RequiresJs("https://cdnjs.cloudflare.com/ajax/libs/jquery/3.4.1/jquery.min.js");
-    Html.RequiresJs("https://cdnjs.cloudflare.com/ajax/libs/jquery-validate/1.19.0/jquery.validate.min.js");
-    Html.RequiresJs("https://cdnjs.cloudflare.com/ajax/libs/jquery-validation-unobtrusive/3.2.11/jquery.validate.unobtrusive.min.js");
+
+    var runtimeMinifier = Current.RuntimeMinifier;
+
+    runtimeMinifier.RequiresJs("umbraco",
```
Shazwazza, line 15: as mentioned in my previous comment for these macro scripts we shouldn't use the minifier to register scripts, the user should choose if they want to use a runtime minifier. Also because this won't work OOTB if there are more than one of these on the page or another cshtml that has RenderJsHere which forces users to update all these macros anyways. We should just render normal script tags and keep it simple. i'll update. (plus like mentioned before this wont work with load balancing)

src/Umbraco.Web.UI/Umbraco/Views/AuthorizeUpgrade.cshtml
```diff
@@ -1,16 +1,17 @@
...
 @{
     Layout = null;
 
-    Html
-        .RequiresCss("assets/css/umbraco.css", "Umbraco")
-        .RequiresCss("lib/bootstrap-social/bootstrap-social.css", "Umbraco")
-        .RequiresCss("lib/font-awesome/css/font-awesome.min.css", "Umbraco");
+    var runtimeMinifier = Current.RuntimeMinifier;
+
+    runtimeMinifier.RequiresCss("umbraco",
```
Shazwazza, line 11: This needs to be a pre-defined bundle... as in, defined in c# on startup. View based declarations will not work in load balancing (see prev comments). I'll update.

src/Umbraco.Web.UI/Umbraco/Views/Default.cshtml
```diff
@@ -17,11 +16,12 @@
             isDebug = true;
         }
     }
+    var runtimeMinifier = Current.RuntimeMinifier;
 
-    Html
-        .RequiresCss("assets/css/umbraco.css", "Umbraco")
-        .RequiresCss("lib/bootstrap-social/bootstrap-social.css", "Umbraco")
-        .RequiresCss("lib/font-awesome/css/font-awesome.min.css", "Umbraco");
+    runtimeMinifier.RequiresCss("umbraco",
```
Shazwazza, line 21: This needs to be a pre-defined bundle... as in, defined in c# on startup. View based declarations will not work in load balancing (see prev comments). I'll update.

src/Umbraco.Web.UI/Umbraco/Views/Preview/Index.cshtml
```diff
@@ -1,14 +1,13 @@
...
+﻿@using Umbraco.Web.Composing
 
 @inherits System.Web.Mvc.WebViewPage<Umbraco.Web.Editors.BackOfficePreviewModel>
 @{
     var disableDevicePreview = Model.DisableDevicePreview.ToString().ToLowerInvariant();
 
-    Html.RequiresCss("assets/css/canvasdesigner.css", "Umbraco");
+    var runtimeMinifier = Current.RuntimeMinifier;
+
+    runtimeMinifier.RequiresCss("umbraco-preview",
```
Shazwazza, line 9: This needs to be a pre-defined bundle... as in, defined in c# on startup. View based declarations will not work in load balancing (see prev comments). I'll update.

src/Umbraco.Core/Runtime/IRuntimeMinifier.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using Umbraco.Core.Assets;
+
+namespace Umbraco.Core.Runtime
+{
+    public interface IRuntimeMinifier
+    {
+        string GetHashValue { get; }
```
Shazwazza, line 9: Method should be more descriptive as to what is doing... this is a cache buster value so should just be called CacheBuster

src/Umbraco.Core/Runtime/IRuntimeMinifier.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using Umbraco.Core.Assets;
+
+namespace Umbraco.Core.Runtime
```
Shazwazza, line 5: This shouldn't be in this namespace, although it's called "RuntimeMinifier" its not part of the actual Umbraco "Runtime". 

There's a few namespaces here that should just be one: Umbraco.Core.Assets is specifically to do with this so they should all be in the same namespace. This is also specifically for "Web" so I'd suggest it be part of an Umbraco.Web.RuntimeMinification namespace which sort of corresponds to the Umbraco.Web.Common.RuntimeMinification. But I'd go a step further, we have other web namespaces in Umbraco.Core that are new that should be consolidated:

* Cookie
* Request
* Search (SearchableTreeAttribute move to Trees)
* Session

I'll create a task for that but will move this and Assets into one namespace

src/Umbraco.Web/JavaScript/CDF/ClientDependencyRuntimeMinifier.cs
```diff
@@ -0,0 +1,228 @@
...
+            }
+
+            var renderer = ClientDependencySettings.Instance.MvcRendererCollection["Umbraco.DependencyPathRenderer"];
+            renderer.RegisterDependencies(dependencies, new HashSet<IClientDependencyPath>(), out var scripts, out var stylesheets, _httpContextAccessor.HttpContext);
+
+            var toParse = assetType == AssetType.Javascript ? scripts : stylesheets;
+            return Task.FromResult<IEnumerable<string>>(toParse.Split(new[] { DependencyPathRenderer.Delimiter }, StringSplitOptions.RemoveEmptyEntries));
+        }
+
+        public async Task<string> MinifyAsync(string fileContent, AssetType assetType)
```
Shazwazza, line 138: There is no await in this method so async shouldn't be here

src/Umbraco.Core/Composing/DefaultUmbracoAssemblyProvider.cs
```diff
@@ -18,7 +18,6 @@ public class DefaultUmbracoAssemblyProvider : IAssemblyProvider
             {
                 "Umbraco.Core",
                 "Umbraco.Web",
-                "Umbraco.Web.BackOffice",
```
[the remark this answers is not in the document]
bergmania: Duplicate, look a couple of lines below

src/Umbraco.Core/Composing/DisableAttribute.cs
```diff
@@ -21,6 +21,11 @@ public class DisableAttribute : Attribute
         public DisableAttribute()
         { }
 
+        public DisableAttribute(string fullTypeName)
+        {
+            DisabledType = Type.GetType(fullTypeName);
```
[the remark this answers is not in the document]
bergmania: This was actually added by mistake, it is not used in our code, but I think it is a good addition. Feel free to add the assemblyname as param also

src/Umbraco.Infrastructure/RuntimeMinification/CssInitialization.cs
```diff
@@ -2,38 +2,47 @@
...
+            var files = await GetStylesheetFilesAsync(requestUrl);
             return WriteScript(files);
         }
 
-        public IEnumerable<string> GetStylesheetFiles(HttpContextBase httpContext)
+        public async Task<IEnumerable<string>> GetStylesheetFilesAsync(Uri requestUrl)
         {
             var stylesheets = new HashSet<string>();
-            var optimizedManifest = OptimizeAssetCollection(_parser.Manifest.Stylesheets, ClientDependencyType.Css, httpContext);
+            var optimizedManifest = await JavaScriptHelper.OptimizeAssetCollectionAsync(_parser.Manifest.Stylesheets, AssetType.Css, requestUrl, _runtimeMinifier);
```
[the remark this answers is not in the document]
bergmania: AssetHelper 🙈

src/Umbraco.Web.Common/Runtime/AspNetCoreComposer.cs
```diff
@@ -34,6 +42,38 @@ public new void Compose(Composition composition)
...
+
+
+            // register the http context and umbraco context accessors
+            // we *should* use the HttpContextUmbracoContextAccessor, however there are cases when
+            // we have no http context, eg when booting Umbraco or in background threads, so instead
+            // let's use an hybrid accessor that can fall back to a ThreadStatic context.
+            composition.RegisterUnique<IUmbracoContextAccessor, HybridUmbracoContextAccessor>();
+
+            // register the umbraco context factory
+            // composition.RegisterUnique<IUmbracoContextFactory, UmbracoContextFactory>();
```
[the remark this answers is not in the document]
bergmania: I don't think it was necessary.. I have the upcoming task about cleaning this up, and "implement as much as possible from WebInitialComposer" I guess a lot of stuff will also be moved into the core composer, as both interface and implementation lives in core/infrastructure.

src/Umbraco.Web.Common/RuntimeMinification/SmidgeComposer.cs
```diff
@@ -0,0 +1,14 @@
...
+using Umbraco.Core.Composing;
+using Umbraco.Core.Runtime;
+
+namespace Umbraco.Web.Common.RuntimeMinification
+{
+    public sealed class SmidgeComposer : IComposer
+    {
+        public void Compose(Composition composition)
+        {
+            composition.Register<IRuntimeMinifier, SmidgeRuntimeMinifier>(Core.Composing.Lifetime.Scope);
```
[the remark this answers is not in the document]
bergmania: Yes, `SmidgeHelper` is scope

src/Umbraco.Web.Common/RuntimeMinification/SmidgeRuntimeMinifier.cs
```diff
@@ -0,0 +1,156 @@
...
+            _ioHelper = ioHelper;
+            _hostingEnvironment = hostingEnvironment;
+            _smidgeConfig = smidgeConfig;
+            _configManipulator = configManipulator;
+        }
+
+        private PreProcessPipeline JsPipeline  => _jsPipeline ??= _preProcessPipelineFactory.Create(typeof(JsMinifier));
+        private PreProcessPipeline CssPipeline  => _cssPipeline ??= _preProcessPipelineFactory.Create(typeof(NuglifyCss));
+
+        private Uri GetRequestUrl() => new Uri(_httpContextAccessor.HttpContext.Request.GetEncodedUrl(), UriKind.Absolute);
```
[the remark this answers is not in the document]
bergmania: Yes guess a null check could be good..

Shazwazza:
## Rework changes

There's a ton of changes here but most are pretty simple. I'll try to explain the main bits:

The re-work was first required because the implementation would support Load Balancing. All bundles need to be registered up-front. Once that refactoring started I found other issues, but in the end the minification refactoring makes things a heck of a lot simpler. Many old files have been removed and many of the abstraction surface area has been reduced.

### IRuntimeMinifier updates

* All bundles are registered on startup with a component and a call to the new `BackOfficeWebAssets` which does the bundle creation. There's a TODO there about eagerly scanning stuff. Right now we do eagerly scan stuff to create the bundle but ideally we don't do this. To fix it is relatively easy and we don't have to change any of the implementation. All we need to do is just ensure that BackOfficeWebAssets.CreateBundles runs at the beginning of the first request made to the website. This will mean we aren't slowing down startup and it will still work in load balancing. But i've left as a TODO now.
* Things like CssInitialize, JsInitialize, etc... are gone and functionality has been simplified and merged into other classes. The namespace for all this has changed to `Umbraco.Web.WebAssets` and `Umbraco.Core.WebAssets`
* The implementation in CDF and Smidge is simplified and there's much less reflection :P 
* There's a work around for the SmidgeHelper since we cannot register our implementation as Per Scope, so we have a new SmidgeHelperAccessor

### Other updates

after this was all done and the latest changes merged to this branch a lot of stuff was breaking. As it turns out our usages of paths was kinda messed up, especially since netcore has a content root and a web root which we need to consider but we weren't. Then after some digging it turns out there's just lots of confused API usages between IIOHelper and IHostingEnvironment (and others). There's a significant overlap between these two things and there shouldn't be. So IIOHelper is far more simplified, lots of comments have been added to code, the usage of IIOHelper has been greatly reduced and IHostingEnvironment used instead since that's generally all that's needed. Much more work can be done here but big progress has been made. I've added TODOs in places as well. We'll still need to be on the lookout for `MapPath` usages which can be done by both IIOHelper and IHostingEnvironment (again, we should remove this overlap, it shouldn't exist on IIOHelper probably). MapPath will also map to the web content root (i.e. www) and in many cases this will be wrong in netcore but we'll need to find that out as we go. In those cases, IHostingEnvironment.ApplicationPhysicalPath should be used.

This is where most of the changes in this code is, just small ctor refactorings really.

I've fixed up the deadlocks with integration tests. There was a couple different issues leading to that problem which are fixed now and I've made it slightly more robust to make sure we aren't closing databases that shouldn't be closed while tests are running.

TypeLoader was an issue because it wasn't rescanning when there were binary changes, this is because it was looking in the wrong folder for binary files (another IOHelper issue). So I've fixed up a TODO that was in there to get around this and we have a new abstraction which is used by TypeFinder called IRuntimeHash which is a thing that builds a hash value for the current runtime which is used to determine if the runtime has changed and used to determine if we should rescan. This is exposed now by ITypeFinder so we didn't have to re-inject a whole bunch of other dependencies. This is working well now.



https://github.com/umbraco/Umbraco-CMS/compare/32deaa12d2d36b0eda4fc25338486074b0d91c1f...c48f36f6a23e771e1e583cff5b20fdcb2ffae82a

PR #7840, opened 2020-03-25
Netcore - Runtime booting and shutting down (and more)

Shazwazza:
This is a rather large PR but a lot of it is just small stuff:

* Splits apart some interfaces into separate components: 
  * `IHostingEnvironment` has been split into 2: `IHostingEnvironment` and `IHostingEnvironmentLifetime` where `IHostingEnvironmentLifetime` is responsible only for the `RegisterObject`/`UnregisterObject` of `IRegisteredObject`. This is required because the netcore version of this requires netcore's `IHostApplicationLifetime` and if we didn't split this, it meant that we would have to resolve `IHostApplicationLifetime` from the netcore service container before the Host was done with it's configuring process which leads to problems because then more than once instance of IHostApplicationLifetime is created which cannot be done.
  * `IRuntimeState` has properties/methods removed from it since these were just wrapping other dependencies which also required Lazy injecting them: Debug, IsMainDom, ServerRole, MainDom ... so wherever these were required, now the appropriate other services are injected: IMainDom, IServerRegistrar, IHostingEnvironment. This simplifies the ctor and dependencies need to pass around.
* IMainDom has been modified to include a new method: `bool Acquire(IHostingEnvironmentLifetime hostingEnvironment);` . This should really have been there anyways since before we were implicitly requiring the first call to `IsMainDom` to actually do the acquiring, not its explicitly done. The main reason for this though is MainDom is a top level, manually created service and we cannot resolve `IHostingEnvironmentLifetime` at the time of it's creation but it's a requirement of MainDom. So now this is passed in to the Acquire method which is done after the configuration process.
* IRuntime has been modified
  * We have split the runtime process into 2 parts like netcore does: Configure (renamed from Boot) and Start. So basically Configure is called (for example) during AddUmbracoCore (in Startup.ConfigureServices) and Start is called in UseUmbracoCore (in Startup.Configure)
  * This makes a lot more sense now especially with regards to netcore, but actually makes more sense in general too. Configure still needs to assembly scan to add things to the container but that's pretty much all it does, Start is the thing that acquires MainDom, runs registered callbacks (i.e. for deploy), initializes components
* Integration tests
  * The integration runtime tests have been changed to use Generic Host which is much nicer/cleaner
  * There is now tests for AddUmbracoCore and UseUmbracoCore along with another test to show that we can boot and install the Umbraco DB

This is a draft since there's still some TODOs to cleanup and need to check if all tests pass on the build server.

Questions/Comments:

* The name of IHostingEnvironmentLifetime ... ? We also have IUmbracoApplicationLifetime, maybe these should be combined?
* I have left quite a number of comments in the code, some with questions
* We need to stop using BuildServiceProvider (as per my notes) one way or another so we should think about that as another task


src/Umbraco.Tests.Integration/Implementations/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Umbraco.Tests.Integration.Implementations
+{
+    public static class HostBuilderExtensions
+    {
+        
+        public static IHostBuilder UseLocalDb(this IHostBuilder hostBuilder, string dbFilePath)
+        {
+            // Need to register SqlClient manually
+            // TODO: Move this to someplace central
+            DbProviderFactories.RegisterFactory("System.Data.SqlClient", SqlClientFactory.Instance);
```
bergmania:
```suggestion
            DbProviderFactories.RegisterFactory(Constants.DbProviderNames.SqlServer, SqlClientFactory.Instance);
```

src/Umbraco.Tests.Integration/Implementations/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                if (!Directory.Exists(dbFilePath))
+                    Directory.CreateDirectory(dbFilePath);
+
+                var dbName = Guid.NewGuid().ToString("N");
+                var instance = TestLocalDb.EnsureLocalDbInstanceAndDatabase(dbName, dbFilePath);
+
+                x.AddInMemoryCollection(new[]
+                {
+                    new KeyValuePair<string, string>("ConnectionStrings:umbracoDbDSN", instance.GetConnectionString(dbName))
```
bergmania:
```suggestion
                    new KeyValuePair<string, string>($"ConnectionStrings:{Constants.System.UmbracoConnectionName}", instance.GetConnectionString(dbName))
```

src/Umbraco.Tests.Integration/Implementations/TestHelper.cs
```diff
@@ -60,7 +63,7 @@ public TestHelper() : base(typeof(TestHelper).Assembly)
 
         public IWebHostEnvironment GetWebHostEnvironment() => _hostEnvironment;
 
-        public override IDbProviderFactoryCreator DbProviderFactoryCreator => new TestDbProviderFactoryCreator();
+        public override IDbProviderFactoryCreator DbProviderFactoryCreator => new SqlServerDbProviderFactoryCreator("System.Data.SqlClient", DbProviderFactories.GetFactory);
```
bergmania:
```suggestion
        public override IDbProviderFactoryCreator DbProviderFactoryCreator => new SqlServerDbProviderFactoryCreator(Constants.DbProviderNames.SqlServer, DbProviderFactories.GetFactory);
```

src/Umbraco.Web.BackOffice/AspNetCore/AspNetCoreHostingEnvironmentShutdown.cs
```diff
@@ -0,0 +1,59 @@
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using Microsoft.Extensions.Hosting;
+using Umbraco.Core;
+using Umbraco.Core.Hosting;
+
+namespace Umbraco.Web.BackOffice.AspNetCore
+{
+    public class AspNetCoreHostingEnvironmentLifetime : IHostingEnvironmentLifetime
```
bergmania: File name should be aligned with class name

bergmania:
Regarding the name of IHostingEnvironmentLifetime, i don't think we should merge, as there are zero classes injecting both. But maybe we can come up with better names ? 🙈

src/Umbraco.Core/Net/IUmbracoApplicationLifetime.cs
```diff
@@ -1,5 +1,6 @@
 namespace Umbraco.Net
 {
+    // TODO: This shouldn't be in this namespace?
```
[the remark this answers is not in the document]
Shazwazza: Could be an option, i'll create a task for that but won't do that in this PR

src/Umbraco.Core/Runtime/MainDom.cs
```diff
@@ -42,17 +42,24 @@ public class MainDom : IMainDom, IRegisteredObject, IDisposable
...
             _logger = logger;
-            _hostingEnvironment = hostingEnvironment;
             _mainDomLock = systemLock;
         }
 
         #endregion
 
+        public bool Acquire(IHostingEnvironmentLifetime hostingEnvironment)
+        {
+            _hostingEnvironment = hostingEnvironment;
```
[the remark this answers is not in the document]
Shazwazza: Yep good call, i'll add null checks in both places

src/Umbraco.Infrastructure/Runtime/CoreRuntime.cs
```diff
@@ -25,12 +23,26 @@ public class CoreRuntime : IRuntime
...
         private ComponentCollection _components;
         private IFactory _factory;
-        private RuntimeState _state;
+        private readonly RuntimeState _state;
         private readonly IUmbracoBootPermissionChecker _umbracoBootPermissionChecker;
         private readonly IGlobalSettings _globalSettings;
         private readonly IConnectionStrings _connectionStrings;
 
 
+        /// <summary>
```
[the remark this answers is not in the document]
Shazwazza: Lol! i think i put that there to add some comments to a parameter or 2 but must have forgot :P I'll remove

Shazwazza:
Have pushed updates to your comments and renamed `IHostingEnvironmentLifetime` to `IApplicationShutdownRegistry` which is exactly what it does so that's a better name.

I'll merge 🎉

Issue #7773, opened 2020-03-06
Memory Leak in GetCacheItem cache dependency


techcolin:
#Summary

We noticed memory creep in our Umbraco instances and started investigating.
Over time our sites were filling RAM on the instance and only a restart would clear it down - often by around 50%.

Running dotMemory to analyse memory dumps pointed towards millions of duplicate strings from the Umbraco Forms module config file path.  While each is tiny the volume amounts to significant RAM.

We have done some local analysis and created a test controller to see what was going on.  I have attached this controller for your interest.  All it does is run a number of GetCacheItem calls in the same way the Umbraco Forms module calls this to load in config settings in:

Umbraco.Forms.Core.Cofiguration.GetFormsSettingFromCache

I raised this as an Umbraco Forms bug however the developer explained their cache extensions are built on top of Umbraco cache extensions, so I investigated further.

## Umbraco version

We have replicated this in both Umbraco 7.14 and also in Umbraco 8.5.1

Reproduction
------------

[MemLeakController.txt](https://github.com/umbraco/Umbraco-CMS/files/4299575/MemLeakController.txt)

Test 1
Run the controller method a number of times while running a memory profiler (ideally dotMemory)
TestBasicCache?useFileDependency=true&nIterations=100000

Each time the number of objects in memory will increase by roughly 100,000
Check the report and you will see these are all duplicate strings for the file dependency path (with a number appended).  In this case:  C:\Temp\tempyfile.config8278623472

Test 2
Restart and run the controller a number of times with useFileDependency=false and you'll see the duplicate strings are no longer created

Forcing Garbage Collection has no impact, as these are all in Unmanaged Memory.

We feel there is an issue with string interning inside the cache extensions somewhere that needs addressed.  

Our dotMemory profiles for these tests on U7 and U8 are available on request, and I have attached a memory graph from the Azure app service which shows the creep with organic traffic over 5 days.

I hope you can help as we are having to restart our app pools each week as these are high volume sites.

![MemoryCreep-5 days](https://user-images.githubusercontent.com/40235909/76108056-c79d4d80-5fd1-11ea-910f-7b50969817d9.png)

---
_This item has been added to our backlog [AB#5774](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/5774)_

techcolin:
[...]

PR #7847, opened 2020-03-26
Fixes Memory Leak in GetCacheItem cache dependency #7773

Shazwazza:
## Fixes Memory Leak in GetCacheItem cache dependency #7773

This fixes a memory leak in the `WebCachingAppCache` which is the implementation of `IAppPolicyCache`

It turns out we were allocating an instance of `CacheDependency` any time `Insert` or `Get` was called even if we weren't going to insert/add to the cache. So a new instance of `CacheDependency` was allocated for every Get. But one object allocation isn't the end of the world ... normally. Turns out that this particular ugly object in .NET Framework runs a lot of code in it's constructor and allocates more objects. 

I ran the test mentioned in #7773 with 1,000,000 iterations before the fix and the code took roughly 20 seconds to execute. With the fix and 3,000,000 iterations it took a couple seconds and there's no more memory leak. It seems that just creating a CacheDependency instance allocates and pins stuff in memory. 

If you are interested in this nasty code it's here https://referencesource.microsoft.com/#system.web/Cache/CacheDependency.cs,304 and this `Init` method is called as part of the ctor! yikes.

Here's the before and after snapshots of the memory and the before is only with 1,000,000 iterations, the after is with 3,000,000. 

image

This is a good candidate to backport to a patch release of 7.15.x IMO.

### DIFF
--- a/src/Umbraco.Core/Cache/WebCachingAppCache.cs
+++ b/src/Umbraco.Core/Cache/WebCachingAppCache.cs
@@ -37,23 +37,28 @@ namespace Umbraco.Core.Cache
         /// <inheritdoc />
         public object Get(string key, Func<object> factory, TimeSpan? timeout, bool isSliding = false, CacheItemPriority priority = CacheItemPriority.Normal, CacheItemRemovedCallback removedCallback = null, string[] dependentFiles = null)
         {
-            CacheDependency dependency = null;
-            if (dependentFiles != null && dependentFiles.Any())
-            {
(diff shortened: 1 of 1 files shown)


SteveVaneeckhout:
> This is a good candidate to backport to a patch release of 7.15.x IMO.
Please do, we have a couple of big websites with Forms that are regularly running into memory issues.

Shazwazza:
@SteveVaneeckhout yep it's here https://github.com/umbraco/Umbraco-CMS/pull/7848

PR #7851, opened 2020-03-28
Fix #7521 - Runtime error if removing email notificationMethod

### LINKED ISSUE #7521: Runtime error if removing email notificationMethod
If you remove the `<notificationMethod alias="email" ...>` element from the `HealthChecks.config` file, Umbraco crashes with a `NullReferenceException`:

```
  at Umbraco.Web.HealthCheck.NotificationMethods.EmailNotificationMethod..ctor(ILocalizedTextService textService, IRuntimeState runtimeState, ILogger logger) in c:\projects\Umbraco-CMS\src\Umbraco.Web\HealthCheck\NotificationMethods\EmailNotificationMethod.cs:line 23
```

The error happens because Umbraco creates a new instance of the `EmailNotificationMethod` class, but when it is not declared in `HealthChecks.config`, the `Settings` property is null.

I personally think that it should be possible to remove the element and maybe replace it with another notifier. But that's just my opinion. Under all circumstances, throwing a `NullReferenceException` seems pretty unfriendly :smile:

## Umbraco version

I am seeing this issue on Umbraco version: 8.5.2 (pretty sure this happens in any Umbraco version containing health checks and health checks notifiers.

### Steps to reproduce

Create a new Umbraco website, comment out the mentioned element and hit <kbd>F5</kbd>.

### Expected result

Umbraco start up without the email notifier configured.

### Actual result

Umbraco crashes.

olibos:
### Description
If we remove the `<notificationMethod alias="email" ...>` element from the `HealthChecks.config` then Settings is null in EmailNotificationMethod constructor.

Just added a check on Settings if it's null then the NotificationMethod disabled itself (same behavior that when there is no `recipientEmail`)

### DIFF
--- a/src/Umbraco.Web/HealthCheck/NotificationMethods/EmailNotificationMethod.cs
+++ b/src/Umbraco.Web/HealthCheck/NotificationMethods/EmailNotificationMethod.cs
@@ -20,7 +20,7 @@ namespace Umbraco.Web.HealthCheck.NotificationMethods
 
         public EmailNotificationMethod(ILocalizedTextService textService, IRuntimeState runtimeState, ILogger logger)
         {
-            var recipientEmail = Settings["recipientEmail"]?.Value;
+            var recipientEmail = Settings?["recipientEmail"]?.Value;
             if (string.IsNullOrWhiteSpace(recipientEmail))
             {
                 Enabled = false;


nul800sebastiaan:
Well that was easier than expected! Wonderful find, thanks very much @olibos! 👍

PR #7853, opened 2020-03-29
Fix #7672 - Image can be added to root of Media when AllowAtRoot is false

### LINKED ISSUE #7672: Image can be added to root of Media when AllowAtRoot is false
Images can be added to the root of the media section even when the AllowAtRoot Property is set to false.

## Umbraco version

I am seeing this issue on Umbraco version: 8.5.3 (also seen in 8.3.0)
 

Reproduction
------------

### Steps to reproduce

1. Set the Image Media Types AllowAtRoot property to false in Setting>Media Types>Image
2. Create a Doc Type a Media Picker property on it and create and node with this Doc Type
3. Use the Media Picker
4. Click upload while viewing the root. This will not open the file upload window
5. Create a new media folder
6. Click into the folder
7. Click to return to the root
8. Click upload while viewing the root. This will open the file upload window

### Expected result

Clicking upload should not open a the file upload window if they are viewing the root, and a warning message should inform the user that images are not allowed at the root.

### Actual result

Clicking the upload button provides no warning message and the file upload window opens

olibos:
This fixes #7672
With @mmrauta's message suggestion

### Description
Set the Image Media Types AllowAtRoot property to false in Setting>Media Types>Image.
Create a Doc Type with a Media Picker property on it and create a node with this Doc Type.
Open the Media Picker.
Click upload while viewing the root of the Media folder. This will not open the file upload window.
Create a new media folder.
Go into the folder.
Click to return to the root.
Click upload while viewing the root. This will no longer open the file upload window.


src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js
```diff
@@ -157,7 +157,13 @@ angular.module("umbraco")
             }
 
             function upload(v) {
-                angular.element(".umb-file-dropzone .file-select").trigger("click");
+                var fileSelect = angular.element(".umb-file-dropzone .file-select");
+                if (fileSelect.length === 0){
+                    localizationService.localize('media_uploadNotAllowed').then(function (message) { notificationsService.warning(message); });
+                }
+                else{
+                    angular.element(".umb-file-dropzone .file-select").trigger("click");
```
bjarnef: @olibos you don't need find the element again. It is already available in `fileSelect` variable.

olibos: My bad! First victim of daylight saving time 😂
Fixed in new version 😉

olibos: @bjarnef have you seen the new commit?

bjarnef: @olibos yes I did... much better 😁 👍

emmary2020:
Hey @olibos 

Thanks for the work here - lovely to see the community collaborating :) 

We'll take a look and get in touch soon,

Emma

olibos:
It seems that the latest updated breaks it
I'll have a look.

nul800sebastiaan:
Apologies for the delay on this one! 🙈  
Thanks also @olibos for your help on this one - together with @mmrauta's fix this is very a very nice combo! ⭐

PR #7856, opened 2020-03-30
V8: Angular Divorce: isUndefined

warrenbuckley:
Uses @nathanwoulfe new Utilities global JS to replace **angular.isUndefined** with **Utilities.isUndefinied**

* Also adds the Utilities.js file to the installer LazyLoad (As a replacement was done in installer for DB type)

This is an unscheduled PR done in my own time, so if the lovely PR team can do their normal thing that would be ace @nul800sebastiaan


nathanwoulfe:
@warrenbuckley, I updated base to v8/contrib, will review and merge. Will be ace to get these changes all sorted.
🎅

nathanwoulfe (approved):
merged latest v8/contrib, no issues/conflicts. Changes here all look good, site runs, no errors.

bjarnef:
@warrenbuckley @nathanwoulfe would it make to add a `Utilities.isUndefiniedOrNull()` function as well?
https://stackoverflow.com/questions/17910192/undefined-or-null-for-angularjs

since we are doing this a lot and could simplify it further.

A few examples:

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/resources/auth.resource.js#L240

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/resources/log.resource.js#L89

PR #7860, opened 2020-03-30
V8: Angular Divorce: equals

warrenbuckley:
Uses @nathanwoulfe new Utilities global JS to replace **angular.equals** with **Utilities.equals**


nathanwoulfe (approved):
Better late than never, hey @warrenbuckley...

Fixed up the expect merge conflicts, all works fine. Time for the merge button...

PR #7865, opened 2020-03-31
Getting Integration tests running on Devops

Shazwazza:
This tries to resolve a few things mostly to do with getting root paths. IOHelper is a bit of a mess so this cleans up a bunch of stuff that shouldn't be used. 

* if using IOHelper to resolve the root it is just ioHelper.MapPath("~") we don't want to use ioHelper.Root there
* GetRootDirectorySafe is an ext method, not part of IOHelper anymore, only used in certain circumstances
* Removes unnecessary methods from IIOHelper
* There are cases where we are using ioHelper.MapPath("~") which will work but it would be better to use IHostingEnvironment.ApplicationPhysicalPath ... but didn't want to go changing all of it right now, would take too long
* Ensures tests use known paths for their files, for tests there's a WorkingDirectory which is the test proj root + TEMP instead of using the CurrentAssemblyDirectory ... some test files are copied to the /bin on build and for those we use MapPathForTestFiles
* For netcore integration tests, we detect if we are using Azure DevOps and change the WorkingDirectory to the %temp% folder. This is required because it's the only read/writable place where LocalDb files can be used. If LocalDb files are used outside of this folder you'll get read/write exceptions. 

When this is merged we need to enable the these items in the netcore build 

You can also see these previously passed for this branch when i had them enabled https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=37392&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b

image


bergmania:
Looks good :)

PR #7880, opened 2020-04-01
7879: Handle for multiple picked media relations

### LINKED ISSUE #7879: Media Picker doesn't record relations in multiple-media mode
In Umbra v8.6.0 media relationships were introduced.  The media picker will record an entry in the umbracoRelations table any time a media item is picked on a page.  This works fine unless the media picker's "Pick multiple items" property is set to true, and you then pick multiple items.  In this case it records that no media is related to the page.

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.0

### Bug summary

This is a bug in the MediaPickerPropertyValueEditor.GetReferences method.  It assumes the string has only a single UDI, and so doesn't split it on a comma before trying to parse the UDI.

### Steps to reproduce

1. Create a new v8.6.0 site with the Starter Pack
2. Connect to the database and execute the following query, which should return 1 row for the image picked as the Hero Background property - `SELECT ur.*, urt.name, p.text, c.text FROM umbracoRelation ur JOIN umbracoRelationType urt ON ur.relType = urt.id JOIN umbracoNode p ON ur.parentId = p.id JOIN umbracoNode c ON ur.childId = c.id WHERE p.text = 'Home' AND urt.alias = 'umbMedia'`
3. In the Umbraco back-office, go to the Settings section, to the Media Picker data type
4. Change the "Pick multiple items" setting to true, and save the data type
5. Go to the Content section and select the Home page
6. Scroll down to the Hero Background property and select any other media item, so that there are two items now selected
7. Save and publish the change
8. Execute the query in the database again, and you'll now see no media relationships for the home page

### Expected result

When you have picked 2 or more media items in a media picker, you should see relationships for all of them in the database.

### Actual result

When you pick more than 1 media item, no relationships are recorded in the database.

benjaminc:
### Prerequisites

Resolves #7879 

### Description
When trying to resolve the media references for a Media Picker property editor, it splits the value on commas before trying to create a UDI.  If there are multiple media items picked, this will ensure that all picked items get recorded as media references.

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/MediaPickerPropertyEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/MediaPickerPropertyEditor.cs
@@ -45,8 +45,11 @@ namespace Umbraco.Web.PropertyEditors
 
                 if (string.IsNullOrEmpty(asString)) yield break;
 
-                if (Udi.TryParse(asString, out var udi))
-                    yield return new UmbracoEntityReference(udi);
+                foreach (var udiStr in asString.Split(','))
+                {
+                    if (Udi.TryParse(udiStr, out var udi))
+                        yield return new UmbracoEntityReference(udi);
+                }
             }
         }
     }


emmary2020:
Hey @benjaminc,

Thanks for the work here! We'll take a look and let you know if we have any feedback.

Emma

nul800sebastiaan:
Good one @benjaminc ! Thanks very much 👍

PR #7884, opened 2020-04-02
Revert GetData() in Block Editor List model

warrenbuckley:
This reverts this final commit 45e892f3505059674779c6e1a43084a367c2862f
in this PR #7557 so you use .Data property and not a GetData() method


warrenbuckley:
The tests for BlockEditor fail - but looking at previous commits in `Block-Editor-List` branch it seems the test has recently stopped working anyway.

So merging this in - heads up for you all @nielslyngsoe @Shazwazza & @benjaminc

PR #7885, opened 2020-04-02
AB4901: UI for notifying user they have no start nodes

warrenbuckley:
@nielslyngsoe this is a rough work in progress PR for you to pick up and fix up as needed Niels :)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/user.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/user.service.js
@@ -185,7 +185,19 @@ angular.module('umbraco.services')
             authenticate: function (login, password) {
 
                 return authResource.performLogin(login, password)
-                    .then(this.setAuthenticationSuccessful);
+                    .then(function(data) {
+
+                        // Check if user has a start node set.
+                        if(data.startContentIds.length === 0 && data.startMediaIds.length === 0){
+                            var errorMsg = "User has no start-nodes";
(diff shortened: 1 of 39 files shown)


src/Umbraco.Web.UI.Client/src/common/services/user.service.js
```diff
@@ -185,7 +185,19 @@ angular.module('umbraco.services')
...
                 return authResource.performLogin(login, password)
-                    .then(this.setAuthenticationSuccessful);
+                    .then(function(data) {
+
+                        // Check if user has a start node set.
+                        if(data.startContentIds.length === 0 && data.startMediaIds.length === 0){
+                            var errorMsg = "User has no start-nodes";
+                            var result = { errorMsg: errorMsg, user: data, authenticated: false, lastUserId: lastUserId, loginType: "credentials" };
+                            eventsService.emit("app.notAuthenticated", result);
+                            throw result;
```
Shazwazza, line 195: Do we want to 'throw' here? or return `$q.reject(result)` ?

nielslyngsoe: Well, I guess we could change the flow and use reject. But for this, I use the Throw to break the promise flow. This way we avoid calling setAuthenticationSuccessful.

But if you feel like I guess it could be remodeled a bit, but this has the least impact on the code.

Shazwazza: ok cool, just wanted to check if there was a reason, just not sure if there's really a diff between the two as I would suspect throwing within a promise would call the failed promise result anyways. but all good.

PR #7889, opened 2020-04-03
Update .editorconfig

warrenbuckley:
## What
This PR contains an update to the **.editorconfig** found at the root of the repository with more verbose .NET & CSharp settings for VS & Rider.

In addition the Umbraco.Web.UI.Client folder where the JS/CSS & HTML lives for the backoffice SPA application to recommend that the EditorConfig extension for VSCode is installed as unfortunately VSCode does not use EditorConfig this out of the box in the core product.

## Who
This is PR is to help anyone who works on the Umbraco CMS source code be it HQ employees or external opensource contributors to the project.

## Why
This is to help uniform our codebase and to ensure anyone working with the CMS codebase edits files in a consistent way. A simple example is that across MacOS & Windows generally use different line endings in files, in an editorconfig we can specify the line ending type to be CRLF.

Another common example is to ensure consistency with indentation type. Some people have their editors set up to use tabs and others spaces, combinded with different indent size. This way we can ensure everyone uses 4 spaces as an indentation & thus when PRs come into GitHub we will have PRs with less whitespace changes

The VSCode specific settings was created for the Umrbaco,Web.UI.Client folder as this seems to the the most common editor when users are working with clientside files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,24 +14,29 @@ indent_size = 4
 trim_trailing_whitespace = true
 
 
-# Code files
-[*.{cs,csx,vb,vbx}]
-indent_size = 4
-insert_final_newline = true
-charset = utf-8-bom
+###############################
+# JS & Less                   #
+###############################
+
+[*.{js,less}]
+trim_trailing_whitespace = false
+
 
 ###############################
-# .NET Coding Conventions     #
+# .NET & CSharp               #
 ###############################
-[*.{cs,vb}]
+
+[*.cs]
+
 # Organize usings
-# dotnet_sort_system_directives_first = true
+# dotnet_separate_import_directive_groups = false
+# dotnet_sort_system_directives_first = false
 
-# this. preferences
-# dotnet_style_qualification_for_field = false:silent
-# dotnet_style_qualification_for_property = false:silent
-# dotnet_style_qualification_for_method = false:silent
+# this. and Me. preferences
 # dotnet_style_qualification_for_event = false:silent
+# dotnet_style_qualification_for_field = false:silent
+# dotnet_style_qualification_for_method = false:silent
+# dotnet_style_qualification_for_property = false:silent
 
 # Language keywords vs BCL types preferences
 dotnet_style_predefined_type_for_locals_parameters_members = true:error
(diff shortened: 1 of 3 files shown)


warrenbuckley:
@nul800sebastiaan updated editorconfig to be in sync with netcore branch in the end but kept the VSCode extension recommendation to install the editorconfig extension if you use that to edit the JS/client side files of Umbraco

.editorconfig
```diff
@@ -1,40 +1,46 @@
...
+
+
+
+
+# https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/ide/editorconfig-code-style-settings-reference.md
+[*.cs]
+csharp_style_var_for_built_in_types = true:suggestion
+csharp_style_var_when_type_is_apparent = true:suggestion
+csharp_style_var_elsewhere = true:suggestion
 csharp_prefer_braces = false : none
```
JimBobSquarePants, line 43: This is such a dangerous setting. 

https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/

Shazwazza: Was added here https://github.com/umbraco/Umbraco-CMS/commit/1d9bb4605bba814d8954b204f1dd5ea099680031 I assume because many moons ago when v8 was being developed by community members if statements with braces were being commonly used. I've been stung with almost the same problem because of this. This rule would also allow for things like if/foreach without any braces... which is far too nasty. 

I think at the time of this rule/development r# was heavily relied upon which would warn of these mistakes. I don't use r# anymore for various reason and I know plenty of others don't either so we can't rely on that. 

I am a fan of terseness but that is only my personal opinion in my own code. I agree that its much safer to be consistent and require braces since there doesn't seem to be a rule that would prevent those types of mistakes.

nathanwoulfe: @Shazwazza what's the resolution here? Is this cool to merge with prefer false?

Shazwazza: @nathanwoulfe we should align whatever is in the codebase in v8 with what is in netcore since that is what we will be doing from here on in: 

So we shouldn't merge these at all IMO unless they are inline with what is in the netcore proj: https://github.com/umbraco/Umbraco-CMS/blob/netcore/netcore/src/.editorconfig

nathanwoulfe:
@Shazwazza should we then go the other direction and merge editorconfig back from netcore? Makes sense to adopt the same rules, and start to get incoming v8 code adhering.

Shazwazza:
@nathanwoulfe yeah I suppose that would be the right thing to do :)

Shazwazza:
@nathanwoulfe that said... I'm not sure we want to do that at this stage since that will probably make merging v8 -> netcore even more miserable since it will cause a huge amount of changes. Maybe we should just leave editorconfig as-is in v8?

nul800sebastiaan:
There's 0 changes in `.editorconfig` in this PR when you disregard whitespace changes.

image

I would suggest reverting `.editorconfig` and the other 2 changes seem nice. It will make changes in each js/html/css file that is touched but I don't think that's a problem with merging and merging up to vNext.

nul800sebastiaan:
Ah but there are additional lines.. maybe I can't see them because of the conflict? I would indeed suggest not changing the `.editorconfig` in the root.

warrenbuckley:
I done this so long ago so feel free to kill this PR off, if its causing headaches

PR #7890, opened 2020-04-03
NetCore: Initialize more services for .net core

bergmania:
This PR cleans up our `AspNetCoreComposer`, and `WebInitialComposer`.

All registrations that could be moved to `CoreInitialComposer` is moved from the two above mentioned composers. The reason so many registrations could be moved, is becuase we have moved that many files from `Umbraco.Web` into either `Umbraco.Core` or `Umbraco.Infrastructure`

Also introduced the following:
- `AspNetCoreCookieManager` 
- `AspNetCoreSessionManager`


src/Umbraco.Core/ContentApps/ContentEditorContentAppFactory.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Umbraco.Web.ContentApps
 {
-    internal class ContentEditorContentAppFactory : IContentAppFactory
+    public class ContentEditorContentAppFactory : IContentAppFactory
```
Shazwazza, line 10: Any reason these are all be made public, there's quite a few in this PR but not sure how that's related to this PR? Just wondering if we are ever going to re-review what should and shouldn't be public if we are just changing everything to public

src/Umbraco.Infrastructure/Runtime/CoreInitialComposer.cs
```diff
@@ -177,6 +196,170 @@ public override void Compose(Composition composition)
 
             // Config manipulator
             composition.RegisterUnique<IConfigManipulator, JsonConfigManipulator>();
+
+
+            // register the http context and umbraco context accessors
```
Shazwazza, line 201: Should all this 'web' stuff be put into a composition extension like we have for ComposeRespositories, etc.. ?

src/Umbraco.Infrastructure/Runtime/CoreInitialComposer.cs
```diff
@@ -177,6 +196,170 @@ public override void Compose(Composition composition)
...
+
+
+            composition.Register<ITagQuery, TagQuery>(Lifetime.Request);
+
+            composition.RegisterUnique<HtmlLocalLinkParser>();
+            composition.RegisterUnique<HtmlUrlParser>();
+            composition.RegisterUnique<HtmlImageSourceParser>();
+            composition.RegisterUnique<RichTextEditorPastedImages>();
+
+            // we should stop injecting UmbracoContext and always inject IUmbracoContextAccessor, however at the moment
```
Shazwazza, line 336: Is this comment still relevant? Thought this was already taken care of?

src/Umbraco.Web.Common/AspNetCore/AspNetCoreCookieManager.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using Microsoft.AspNetCore.Http;
+using Umbraco.Core.Cookie;
```
Shazwazza, line 3: I had a task to cleanup this namespace since it shouldn't exist like that, just not sure where those notes went. There's a few of new namespaces in Core that have 'one' item which shouldn't exist and needs to be better organized. These are mostly to do with 'web' things. Anyways, i'll create/find the task for that or just do it after review.

src/Umbraco.Web.Common/Runtime/AspNetCoreComposer.cs
```diff
@@ -39,41 +41,19 @@ public new void Compose(Composition composition)
...
-            composition.MediaUrlProviders()
-                .Append<DefaultMediaUrlProvider>();
+            composition.RegisterUnique<ICookieManager, AspNetCoreCookieManager>();
 
-            composition.RegisterUnique<ISiteDomainHelper, SiteDomainHelper>();
+            composition.RegisterMultipleUnique<ISessionIdResolver, ISessionManager, AspNetCoreSessionManager>();
 
-            // register properties fallback
-            composition.RegisterUnique<IPublishedValueFallback, PublishedValueFallback>();
+            composition.RegisterUnique<ISessionIdResolver, AspNetCoreSessionManager>();
```
Shazwazza, line 55: Is this a duplicate since above there is already `composition.RegisterMultipleUnique<ISessionIdResolver, ISessionManager, AspNetCoreSessionManager>();` ?

src/Umbraco.Infrastructure/Runtime/CoreInitialComposer.cs
bergmania: It only seemed to be nessasary due to `RenderNoContentController`, but I fixed that up

src/Umbraco.Infrastructure/Runtime/CoreInitialComposer.cs
bergmania: I'm not sure exactly where to do the cut between "Web" and non web. Feel free to give it a shot, but don't see it as an important thing to do

src/Umbraco.Core/ContentApps/ContentEditorContentAppFactory.cs
bergmania: It is related because internals from core is not visible to infrastructure. IMO a good thing, because it makes it harder to test.

I think it is very important to take revisit the internal/public discussion. But whether we should use real `internal` is most likely the big question. The only reason to do it now is to avoid introducing breaking changes, right?

Shazwazza: Yes without 'real' internal makes things very difficult to deal with breaking changes in the future. There's a few 'challenges' internally that we already need to deal with in regards to breaking changes so if we even further restrict what we can change it's going to be very very difficult to change things into the future without a new major version or cutting through a lot of red-tape. 

On the note of 'real' vs non real internal things, i don't agree that non-real internal things works such as having an 'internal' namespace. This is what Microsoft tried with netcore and it failed which is why they've gone back to real internals since everyone was just using the internal namespaces and then complaining when they broke it. 

But anyways, we just need to make sure that we somehow save some time to go through all the APIs that we don't actually need to have as public before release else we may end up shooting ourselves in the foot.

PR #7895, opened 2020-04-04

AndyButland:

- Stylesheet
- Template
- Relation

I've made a small, simplifying amend to the `RelationTypeBuilder` in the process.


src/Umbraco.Tests.Common/Builders/TemplateBuilder.cs
```diff
@@ -0,0 +1,116 @@
...
+        public override Template Build()
+        {
+            var id = _id ?? 0;
+            var key = _key ?? Guid.NewGuid();
+            var name = _name ?? Guid.NewGuid().ToString();
+            var alias = _alias ?? name.ToCamelCase();
+            var createDate = _createDate ?? DateTime.Now;
+            var updateDate = _updateDate ?? DateTime.Now;
+            var path = _path ?? string.Empty;
+            var content = _content;
```
elit0451, line 54: Why are we skipping the null check here?
```suggestion
            var content = _content ?? string.Empty;
```

AndyButland: I found we needed to retain this as null, otherwise we'd get a test failure here; `Assert.AreEqual(clone, template);`.  Seems null specifically is expected here, rather than empty string.

bergmania:
Looks good :)

PR #7896, opened 2020-04-04
Netcore: Further model unit test migrations into new project and builder pattern (2)

AndyButland:

- PropertyGroup
- PropertyType
- Property

I've also tweaked the setup where we have some builders that might be used in different contexts - whether standalone, created as a child of another builder and, in particular, when created as a child of more than one builder (so using generics).  I've extended this using marker interface constraints (see `PropertyGroupBuilder` and `PropertyTypeBuilder`.


src/Umbraco.Tests.Common/Builders/PropertyBuilder.cs
```diff
@@ -0,0 +1,71 @@
+using System;
+using Umbraco.Core.Models;
+using Umbraco.Tests.Common.Builders.Interfaces;
+
+namespace Umbraco.Tests.Common.Builders
+{
+    public class PropertyBuilder
+        : BuilderBase<Property>,
```
bergmania, line 8: Any reason, this cannot be a `IProperty`

src/Umbraco.Tests.UnitTests/Umbraco.Infrastructure/Models/PropertyTests.cs
```diff
@@ -0,0 +1,64 @@
...
+                Assert.AreEqual(propertyInfo.GetValue(clone, null), propertyInfo.GetValue(property, null));
+            }
+        }
+
+        private Property BuildProperty()
+        {
+            return _builder
+                .WithId(4)
+                .WithKey(Guid.NewGuid())
+                .WithCreateDate(DateTime.Now)
```
bergmania, line 46: We can use defalts for some of these.. `Key`, Dates, ...

src/Umbraco.Tests.Common/Builders/PropertyBuilder.cs
AndyButland: No, seems to be fine with `IProperty`, have updated.

src/Umbraco.Tests.UnitTests/Umbraco.Infrastructure/Models/PropertyTests.cs
AndyButland: Agreed. have removed.

AndyButland:
Have made the two suggested amends.  I agree on the extension method for a complex object, but suggest will save that for another PR?

Getting a "Packages failed to restore" though now I've merged the latest `netcore/dev` in to resolve conflicts...

AndyButland:
Have hopefully fixed that... merge issue.

bergmania:
Yes, a separate PR is fine for a good way to create a complex object.

PR #7909, opened 2020-04-07
Netcore: Alter ResXFileRef to match filesystem

rustybox:
Resolves build error on case sensitive file systems

ResXFileRef should match path on filesystem

```
error MSB3103: Invalid Resx file. System.IO.FileNotFoundException: Could not find file '/home/example/Dev/Umbraco-CMS/src/Umbraco.Infrastructure/WebAssets/jsinitialize.js'.
```


Shazwazza:
thanks!

PR #7910, opened 2020-04-07
Add migrations for new block editor

benjaminc:
NOTE: This is a review PR to the block-editor-list branch, not a PR to the v8/contrib or other such branch.

This PR adds three database migrations.  Two are necessary for converting stacked content data to block list data.  The third is one for the color picker, which apparently never got its pre-values converted correctly, and which I found while testing and working with the block list data.  I thought about moving that last one into its own issue and PR, but then there would be code merges and DB state merge paths in the UmbracoPlan class, and if no one else is complaining about it yet, I didn't think it worth it.

---
_This item has been added to our backlog [AB#6195](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6195)_


benjaminc:
For reference, I've tested all three migrations on a database that was at 8.6.0 (having previously been upgraded from a 7.15.3 database with Stacked Content and color pickers in it), and straight from 7.15.3 to this latest.  There were no issues either way, at least with the data I have.

benjaminc:
I'm going to pull the Color Picker migration out of this PR, and move it into a PR for issue #7939, along with migrations for the other items mentioned in that PR.

nielslyngsoe:
Stil needs review from @Shazwazza

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/ConvertToElements.cs
```diff
@@ -0,0 +1,78 @@
...
+    public class ConvertToElements : MigrationBase
+    {
+        public ConvertToElements(IMigrationContext context) : base(context)
+        {
+        }
+
+        public override void Migrate()
+        {
+            // Get all document type IDs by alias
+            var docTypes = Database.Fetch<ContentTypeDto>();
```
Shazwazza, line 19: why not just do `Database.Fetch<ContentTypeDto>().ToDictionary(x => x.Alias, x => x.NodeId)` ?

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/ConvertToElements.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Find all compositions those document types use
+            var parentElementTypeIds = Database.Fetch<ContentType2ContentTypeDto>(Sql()
+                .Select<ContentType2ContentTypeDto>()
+                .From<ContentType2ContentTypeDto>()
+                .WhereIn<ContentType2ContentTypeDto>(c => c.ChildId, elementTypeIds)
+                ).Select(c => c.ParentId);
+
+            elementTypeIds = elementTypeIds.Union(parentElementTypeIds).ToList();
+
+            // Convert all those document types to element type
```
Shazwazza, line 38: Is there any chance this can break people's sites? Like if potentially they were using a doc type as both an element type and a doc type?

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/ConvertToElements.cs
```diff
@@ -0,0 +1,78 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System.Collections.Generic;
+using System.Linq;
+using Umbraco.Core.Persistence;
+using Umbraco.Core.Persistence.Dtos;
+
+namespace Umbraco.Core.Migrations.Upgrade.V_8_7_0
+{
+    public class ConvertToElements : MigrationBase
```
Shazwazza, line 10: Would be good to have some code comments on this class and maybe renamed to ConvertToElementTypes ?

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/StackedContentToBlockList.cs
```diff
@@ -47,12 +48,40 @@ private List<DataTypeDto> GetDataTypes(string alias)
                 .On<ContentTypeDto, NodeDto>(c => c.NodeId, n => n.NodeId);
 
             var types = Database.Fetch<ContentTypeDto>(sql);
-            var map = new Dictionary<Guid, string>(types.Count);
-            types.ForEach(t => map[t.NodeDto.UniqueId] = t.Alias);
-            return map;
+            var typeMap = new Dictionary<int, ContentTypeDto>(types.Count);
```
Shazwazza, line 51: As above, a ToDictionary call seems easier

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/StackedContentToBlockList.cs
```diff
@@ -47,12 +48,40 @@ private List<DataTypeDto> GetDataTypes(string alias)
...
+                .From<PropertyTypeDto>()
+                .InnerJoin<DataTypeDto>()
+                .On<PropertyTypeDto, DataTypeDto>(c => c.DataTypeId, n => n.NodeId)
+                .WhereIn<DataTypeDto>(d => d.EditorAlias, new[] { Constants.PropertyEditors.Aliases.NestedContent, Constants.PropertyEditors.Aliases.ColorPicker });
+            var props = Database.Fetch<PropertyTypeDto>(sql);
+            // Get all nested content and color picker property aliases by content type ID
+            var propLk = props.ToLookup(p => p.ContentTypeId, p => p.Alias);
+
+            var knownMap = new Dictionary<Guid, KnownContentType>(types.Count);
+            types.ForEach(t => knownMap[t.NodeDto.UniqueId] = new KnownContentType
```
Shazwazza, line 76: ToDictionary?

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/StackedContentToBlockList.cs
```diff
@@ -73,16 +102,16 @@ private bool Migrate(IEnumerable<DataTypeDto> dataTypesToMigrate, Dictionary<Gui
...
             var old = JsonConvert.DeserializeObject<StackedContentConfiguration>(dataType.Configuration);
             var config = new BlockListConfiguration
             {
                 Blocks = old.ContentTypes?.Select(t => new BlockListConfiguration.BlockConfiguration
                 {
-                    Alias = knownDocumentTypes[t.IcContentTypeGuid],
+                    Alias = knownDocumentTypes.TryGetValue(t.IcContentTypeGuid, out var ct) ? ct.Alias : null,
                     Label = t.NameTemplate
-                }).ToArray(),
+                }).Where(c => c.Alias != null).ToArray(),
```
Shazwazza, line 114: null or whitespace?

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/StackedContentToBlockList.cs
```diff
@@ -141,11 +170,11 @@ private void UpdateDataType(DataTypeDto dataType)
         private class BlockListConfiguration
         {
 
-            [ConfigurationField("blocks", "Available Blocks", "views/propertyeditors/blocklist/prevalue/blocklist.elementtypepicker.html", Description = "Define the available blocks.")]
+            [JsonProperty("blocks")]
```
Shazwazza, line 173: Not sure why these changes are in this PR? Also i might just not be in the loop on this one but why are we removing the [ConfigurationField] attributes for things that are configuration fields?

Shazwazza: oooooh .... haha, these are copied locally here as private classes specific to the migration, ok i gotcha

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/ConvertToElements.cs
benjaminc: I had written an app a couple years back that did a lot of list to dictionary conversions, and it was performing horribly.  After doing using VS's performance analyzer, we found that the dictionary resize operations are extremely expensive, especially as the list grows to 1,000+ elements.  If you size the dictionary appropriately at the start and then just insert into it, it was lightning fast.  Ever since then, I've always tried to make it a habit to size the dictionary appropriately and then insert into it.  Since the ToDictionary extension is on the IEnumerable interface, it can't do that, and so just iterates and adds to the dictionary.  If they created a ToDictionary that was an extension on the ICollection interface, that would be ideal, but without that I just create my own, properly sized dictionary and add to it.

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/ConvertToElements.cs
benjaminc: There is, and I thought about that further when I was writing the migrations that I did for #7939 where I decided that if there were any element types already (i.e. this wasn't a v7 to v8 migration), then I wouldn't mess with retyping any document types.  I'll port that change back to here, to minimize the effect on anyone upgrading from a previous v8 version.  We are dealing just with those upgrading Stacked Content, so likely they are still coming straight from v7 anyway, but it would be a good quick check.

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/ConvertToElements.cs
benjaminc: Can do, I've got some busy projects for the next couple of days at work, but I'll add some comments when I get a chance next week.

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/StackedContentToBlockList.cs
benjaminc: That would imply that there was a content type alias in the database that was whitespace, and which we didn't want to map here, because if the GUID matched a known content type it would get the alias in the DB, and if it didn't it would always return null.  I don't think whitespace aliases are valid for content types, but if they were, I'd want to allow them here.

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/ConvertToElements.cs
Shazwazza: Fair enough! Might be worth making an ext method for this then based on ICollection<T>, not required for this PR but in the future. And good to know! Wonder if there are places in our codebase where we are creating large dictionaries with ToDictionary based on collections, might have a look today and see.

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/ConvertToElements.cs
Shazwazza: Cool so since this is already merged i supposed we will just wait until you have a little bit of time to push a new PR, just let us know :) I'll add some notes to the code.

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/ConvertToElements.cs
Shazwazza: I decided to create a PR for changing some of our internal usages of ToDictionary and discovered we should be using ToLookup in various places too instead of GroupBy/ToDictionary which is much slower. I've left this open for the community to review but please have a look i fyou have time https://github.com/umbraco/Umbraco-CMS/pull/7969

src/Umbraco.Core/Migrations/Upgrade/V_8_7_0/ConvertToElements.cs
benjaminc: Just as a follow-up to this, I ended up pulling the ConvertToElement migration in PR #8336 since the NestedContent portion is PR'd in #7957 and without the StackedContent to BlockEditorList migration, there is nothing else to convert.

Issue #7054, opened 2019-11-05
Media picker paging when searching doesn't page


Shazwazza:
If you use the media picker and then search for an item, the search works and the GetPagedDescendants call returns the correct elements, but the paging doesn't work properly:

* The first page only shows a few items instead of 100 (which is what is being returned in my tests)
* When going to the 2nd page, it shows nothing

You can see the results of GetPagedDescendants is correct but the UI paging is doing something funky.

See video: https://www.screencast.com/t/VGGvawt02

_Note: in the video you'll see JS errors but this is not this bug, it is this one: https://github.com/umbraco/Umbraco-CMS/pull/6634#issuecomment-549643188_

kjac:
@Shazwazza that's funky... I can't replicate it at all 🤔 

Here's how it looks on my end:

![media-picker-pagination](https://user-images.githubusercontent.com/7405322/68205491-c24fd800-ffca-11e9-8fa7-8efc4a15c0c4.gif)

Shazwazza:
@kjac Do you have folders/subfolder/etc... in your media lib? I only have a few in the root but thousands in some sub folders.

kjac:
@Shazwazza I do. In that gif I'm emulating your screen cast, searching I'm a sub folder with some 120 images in it. The folder does not have sub folders though, but your cast didn't show you including sub folders in your search.

I wonder if this is a volume related issue? Can't see why it should be but.. never say never.

Could you post the json response from your first result page? Maybe that could give me some hints as to what's happening.

umbrabot:
Hiya @Shazwazza,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

Issue #2957, opened 2018-09-11
Mini list view in content picker uses fixed sort order / field instead of sort options from associated (custom) list view


PGGMTeam5:
The mini list view in content picker uses a fixed sort order / sort field instead of using the sort options from the associated (custom) list view.
See the screen shots below.

### Steps to reproduce
Create a standard new Umbraco (v7.10.2) installation, with the supplied standard starter kit.
Change sort order in de List View - products to "Last edited" and "Descending"
Edit one of the Product child items, for instance the "Unicorn" item, save and publish
In the Products list view the edited item ("Unicorn") is now displayed as the first item, conform the changed sort order
Now go to some content page with a rich text editor
Insert a link in the rich text editor:
In the tree view in the link picker expand the Home and then the Products content item

### Expected result
I expected the child Product items in the mini list view to be sorted based on the "Last edited" field, descending.

### Actual result
The product items are sorted based on "SortOrder", ascending.

### Proposed solution
In the "MiniListViewDirective" there is a function "open(node)", in which there is now a var 'miniListView' being declared, with a "pagination" property, which in turn contains the properties "orderDirection" and "orderBy". These properties now contain the fixed values "Ascending" and "SortOrder". However, these properties should be read and set based on the list view that is associated with the node parameter.

### Screen shots
![change 20sort 20order 20in 20list 20view](https://user-images.githubusercontent.com/41419466/45344712-f0d34e00-b593-11e8-8ea9-5dd056229c5d.png)
![sort 20order 20in 20link 20picker 20not 20correct](https://user-images.githubusercontent.com/41419466/45344713-f16be480-b593-11e8-971e-7ccbd77725ef.png)
![sort 20order 20in 20products 20content 20item 20is 20ok](https://user-images.githubusercontent.com/41419466/45344714-f16be480-b593-11e8-963e-800a8d00e787.png)

marcemarc:
Have had this same issue on a site, when a content editor is 'picking from a listview' in most scenarios I find they most likely want to be picking something that has 'just been created' and therefore having the default sort order of the MiniListView picker being 'descending' by sort order, would I think be much more preferable for the 'default' experience than having the items in Ascending order... and asking the editor to 'go to the last page' to get the 'latest' thing to pick. When this is explained to editors they often ask why they can't change the sort order manually?

So couple of possibilities of steps in the right direction:
[...]

PR #7912, opened 2020-04-08
V8: Add alternate 'List View' to Media Picker - Take 2

marcemarc:
This is a second attempt at: https://github.com/umbraco/Umbraco-CMS/pull/7198

Basically the original PR got woefully out of date, and my attempt to bring it back up-to-date was signaling thousands of changes after the rebase... so impossible to see what the PR was actually changing...

Anyway this is the original Issue: https://github.com/umbraco/Umbraco-CMS/issues/5259

### Recap: What does this Pull Request do?

It adds a little icon in the Media Picker to enable editors to 'toggle' between the normal view of images to pick from and an alternate listview (the idea is we think more like a file system, when we are picking assets, and if you know the name of the file you want to pick, it's quicker to pick from a list than scroll through possibly lots of similar images with similar filenames).

The idea is to never 'crop' the name of the Media Item, so the editor can always see the whole name of the item without the ... obscuring the last bit (often the bit that distinguishes one file from another)

It makes the Media Picker slide out panel wider using the new 'medium' setting

It allows the list view of media items to be sorted by 'Name' or by 'Last Update Date'*

There is a video here:

The Media Picker 'should' just work in terms of its existing functionality, and it should be tested with different configurations, eg picking folders, images only etc, and also should be tested from inside the Rich Text Editor and Grid.

### This attempt

I got stuck previously retrieving the UpdateDate for the media item, in the sense that I added an UpdateDate property to the EntityBasic, and then all the Mapping functions screamed about that field not being mapped everywhere else EntityBasic was in use... I've now discovered the AdditionalInfo property where you can stuff additional data, so no need to extend EntityBasic, no issues with the Mapping functions.

### Other things

I listed lots of other things on the original PR https://github.com/umbraco/Umbraco-CMS/pull/7198 that might be considerations, but maybe just pull this and see how it feels, rather than be put off by my long ramblings of concerns...


tcmorris:
Hey @marcemarc 

Thanks for the PR! I pulled this down to take a look. 

Functionality wise, it seems to work as described. Things I tried...
- Select a media item and then toggle, it maintains the currently selected media item and switches to the list or grid view. 
- Change the sort column on list view and then toggle, the sort column is kept the same.
- Select a media item, search for that item and ensure it is still selected.
- View a list of folders and toggle between list/grid view.
- Used it within the grid and rich text editor
- Upload a media item with long name, it was trimmed on grid and then included all of the content in list view as mentioned in the ticket. 

Things I noticed...
- The order of items is not the same. The list view by default is ordered by name. The ordering in grid view appears to be create date asc.
- The order of the items in media picker is also different to the ordering in media section, which appears to be create date desc.
- The toggle is not the same as the toggle in the media section. Might be good to update for consistency. Is the toggle in the media section better for usability?
- The name isn't trimmed in list view on media picker, which perhaps is unlikely to be a problem, but in the media section the name is trimmed so I would suggest matching that for consistency.

toggle-in-media-section
media-picker-list-view
media-section-list-view

Overall, looks like a useful feature. Nice work 👍

marcemarc:
Thanks @tcmorris for giving it a spin... much appreciated...following up on your observations..

The image view on the picker, I think, is fitting the images into the space available an so is only loosely following create date, but not necessarily. my thinking is you switch to ListView when you want to see the name of the file and so therefore A-Z makes more sense - but clicking the headings does reorder... so you can choose to order by latest date or Z-A etc - I've made Folders sorted separately from Files, to be more consistent with a File System.

There is another issue about trying to connect the settings of a ListView to the settings of a picker that is picking from it: https://github.com/umbraco/Umbraco-CMS/issues/2957 but I think the context of picking media items is different to the context of browsing the Media section... my thinking was to go A-Z as the use case is focussed on not being able to tell the difference between similar images, only the filename is the distinguisher...

Purposefully the filename isn't trimmed here, my thinking is because one of the use cases is to be able to tell the difference between similarly named filenames eg photoshoot_20042020_1.jpg, photoshoot_20042020_2.jpg, photoshoot_20042020_3.jpg - the last bit of the filename is often the bit you need to see to be able to pick the right one...

Agree about the toggle, I did look at it briefly but wasn't sure how exactly to wire it up, will have a think.

Thanks!

Marc

marcemarc:
@tcmorris ok have pushed an update - it now uses the umbLayoutSelector directive for a consistent UI toggle between views - but not sure the directive is intended for how I've used it here :-) but you are right it looks much better

marcemarc:
image

tcmorris:
Hey @marcemarc 

Pulled the update down this afternoon. Looking much better. From a UI perspective makes more sense and good to have that consistency. 

I think the naming and order is fine as it is based on what you described. The use cases seem reasonable and could be part of later PR if that needs to be updated. 

Looks good to me!

mzajkowski:
Hello there!

It's great to see the progress here about which I wasn't aware of totally. I came into this issue as there is something else bothering me and our project recently.

During the media picker load, there is a call to get the collection of all media items and it's not paginated by default, what - as you can imagine - is causing a 'small' issues when we're dealing e.g. with 19k++ documents in the media library (as we do at the moment).

Does anyone faced this issue? What we're currently hacking is the extension for the default media picker where we'll replace the calls to our custom endpoints to save us and our editors from dealing with frozen browsers and hundreds of documents being loaded all the time.

Maybe it'll be a good opportunity to combine efforts and also merge it here to generally "improve" the journey of picking medias.

marcemarc:
Hi @mzajkowski 

I'm wondering if you are experiencing something like this issue here: https://github.com/umbraco/Umbraco-CMS/issues/7054 ?

But yes, paging tons of thumbnail images isn't a great experience, and being able to switch to a ListView of just filenames temporarily, may prevent the browser freezing?, helps editors pick the image they want, particularly if there are lots of them and similar in nature - but it all depends on the context of the site and it's use of media. The history this PR is based on is from super huge sites we've had with tons of images - feedback from editors using the contents of this PR is that it works well, that sometimes they prefer to see images, but in some situations having the filenames, 'like a filesystem' is faster, particularly being able to see the date modified to confirm they are picking the 'right' image.

It would be interesting to see if you could run this branch against a copy of your db + media to see if it helped? and report back, eg give an indication if this is worth pursuing further?

But in terms of this PR, I am not sure where it is going! It has **no 'state label'**, but it isn't a speculative PR either, original issue was created here: https://issues.umbraco.org/issue/U4-10518 but this was closed when moving to the new issue tracker, following a flurry of Media Picker UI Improvements I recreated the issue for V8 https://github.com/umbraco/Umbraco-CMS/issues/5259 - (sneakily hoping @kjac would do it in his sleep!) The issue was discussed in HQ and feedback from @nul800sebastiaan was:

> We had a look at this at HQ and think this is a really cool and handy feature and if you're still up for it, this would make a great core feature.

and further feedback on the UI from @nielslyngsoe in Oct 2019

> I think enabling a listview for the MediaPicker makes a ton of sense. I do though think we should aim for a table layout for the list, So we have columns for each value. Each columns should be sortable, enabling sorting by creation-date or name.

So I took that feedback and worked up a new PR, but I got stuck and nobody could help :-( so it languished...

Just before the Candid Contributions hackathon in April, with fresh eyes I think I worked out the blocking issue and @tcmorris helped review, and suggest improvements (see above) - so unofficially I'm told this is awaiting UI review, that there is a team that meets regularly to discuss community PRs that affect the UI of Umbraco, we know at least one of these meetings have taken place, (@hartvig has tweeted about them!) but we don't know if this PR was considered,

It's tricky as further changes have been made to the core media picker, meaning this PR is slightly out of date, which may mean it can't be considered anymore for UI review until it is updated, but it would be good to know whether the idea in this state has been reviewed and this is something that is still 'really cool and handy feature'... and on the right tracks, and just in a backlog - but it's helpful and worth me keeping the branch up to date.

What I'm saying is don't get your hopes up!

Either this has just fallen off the radar, or more likely there are more important things ahead of it, Umbraco cannot do 'everything' - and taking time to explain or label an open PR takes time too. 

There is a bit of a pandemic at the moment! so nothing is normal, procedure and time wise, and this is slightly less important than staying safe!

Maybe it has been reviewed and it's been decided not to proceed with this, but nobody really wants to close a PR knowing work has gone it to it, and that work has previously been encouraged by HQ - there is an argument to say this shouldn't be a core feature, perhaps most sites only pick from a list of 4 or 5 images and this idea is misplaced, and only seems necessary on 'larger sites' - but it would be good to know that, as we can perhaps look at alternative ways of working with Media outside of Umbraco for these larger implementations.

The other alternative is the idea is still nice, but what I've tried to implement isn't very good, and isn't really salvageable - but it is totally ok to close the PR with that feedback and keep the issue open with the community/upforgrabs label, or an internal sprint.

But I find it fascinating to see how this plays out, it's a good example of what happens when a PR stalls, and it's not easy for anyone to know what to do with them!

poornimanayar:
Hi @marcemarc ,

Apologies for the delay. I have tested this feature.. Really nice work. The UI is very consistent with the rest of the back office and the sorting , searching, all of it works very well. I have also followed the conversation here and tested everything as mentioned by @tcmorris. Great work with the feedback @tcmorris and I can confirm that all of it works

There is one tiny little thing which I noticed. When you search for a media name the results table seems to jump up a bit.  I am not sure whether its something that has been introduced here or an issue across the board. 

**In this image there is a margin between the layout selector and table**
image

**In this image its much closer**
image

Its really a tiny one considering the grand scale of things but I think its good to cross that off too.

I have gone through the code as well. I think I see quite a breaking change happening around this commit https://github.com/umbraco/Umbraco-CMS/pull/7912/commits/66934b8bb4b4b14d3a20cc873ba2400e6adad165

We are changing from `CreateDate` to `UpdateDate` in the `IEntityBasic` and that is then flowing through all the other entities. Can we work it out another way? I am thinking from package devs perspective and I feel that it could be a breaking change.

Let me know @marcemarc , speak to you soon!

Poornima

marcemarc:
Hi @poornimanayar 

Thanks for the feedback, yes I think there isn't a breadcrumb element on the left, in the search scenario, so you are right it doesn't look consistent in terms of spacing in that context!

With regards to the breaking change, that isn't being made in this PR, if you look at the Files that have changed in this PR: https://github.com/umbraco/Umbraco-CMS/pull/7912/files you can see that change isn't made.

As part of the first attempt I made, I thought that was the way, but I soon realised how much that would change following that path! and then discovered the MetaData of the MediaItem to be able to pass through the UpdateDate without adding it to EntityBasic!

So let me know if that is ok?

regards

Marc

poornimanayar:
Apologies @marcemarc , I realise that now! Lazy eyes. I didn’t go through the whole list of commits. But if you can fix the little thing that will be great. I will have a look at this PR and go through the commits once again. Let’s look to get this merged then :-)

Poornima

src/Umbraco.Web.UI.Client/src/views/components/umb-media-grid.html
```diff
@@ -8,6 +9,7 @@
         ng-class="{'-selected': item.selected, '-file': !item.thumbnail, '-folder': item.isFolder, '-svg': item.extension == 'svg', '-selectable': item.selectable, '-unselectable': !item.selectable}"
     >
         <div>
+                <a ng-if="allowOnClickEdit === 'true'" ng-click="clickEdit(item, $event)" ng-href="" class="icon-edit umb-media-grid__edit"></a>
```
nul800sebastiaan, line 12: Following the rest of the accessibility updates, can you make this a `<button type="button" .. etc></button>` please?

An example update from anchor to button is here: https://github.com/umbraco/Umbraco-CMS/pull/8162/files#diff-6dadd238fd47b0ef06ceec320c1325adR27

src/Umbraco.Web.UI.Client/src/views/components/umb-media-grid.html
```diff
@@ -19,26 +21,79 @@
...
+                                <i class="umb-table-head__icon icon" aria-hidden="true" ng-class="{'icon-navigation-up': isSortDirection('name', 'asc'), 'icon-navigation-down': isSortDirection('name', 'desc')}"></i>
+                            </button>
+                        </div>
+                        <div class="umb-table-cell">
+                            <button type="button"
+                                    class="umb-table-head__link sortable"
+                                    ng-click="setSort('updateDate')"
+                                    prevent-default>
+                                Last Updated
+                                <i class="umb-table-head__icon icon" ng-class="{'icon-navigation-up': isSortDirection('updateDate', 'asc'), 'icon-navigation-down': isSortDirection('updateDate', 'desc')}"></i>
```
nul800sebastiaan, line 74: Can add `aria-hidden="true"` here as well

src/Umbraco.Web.UI.Client/src/views/components/umb-media-grid.html
```diff
@@ -19,26 +21,79 @@
...
+                        </div>
+               
+                    </div>
+                </div>
+                <div class="umb-table-body">
+                    <div class="umb-table-row umb-media-grid__list-item -selectable" ng-click="clickItem(item, $event, $index)" ng-repeat="item in items | filter:filterBy | orderBy:sortBy:sortReverse" ng-class="{'-selected': item.selected}">
+
+                        <div class="umb-table-cell">
+                            <i ng-show="item.selected" class="umb-table-body__icon umb-table-body__checkicon icon-check" aria-hidden="true"></i>
+                            <i class="{{item.icon}}" ng-if="!item.thumbnail && item.extension != 'svg' && !item.selected"></i><i class="icon-picture" ng-if="item.thumbnail && !item.selected"></i>
```
nul800sebastiaan, line 85: Can add `aria-hidden="true"` here as well for both of them

marcemarc: > Can add `aria-hidden="true"` here as well for both of them

Added (one was there already)

src/Umbraco.Web.UI.Client/src/views/components/umb-media-grid.html
marcemarc: > Can add `aria-hidden="true"` here as well

added!!

src/Umbraco.Web.UI.Client/src/views/components/umb-media-grid.html
marcemarc: > Following the rest of the accessibility updates, can you make this a `<button type="button" .. etc></button>` please?

Well this is a strange thing... I didn't add that 'allowonclickedit' link, it's not part of the List View bits I'm adding... looking at the history it was removed in commit 0992bd8 as part of the Media Picker UX Updates - so I think when that has been merged into this branch, the conflict hasn't been resolved correctly, and it's been accidentally reintroduced.

So I've taken the existing HTML from the v8/contrib brand for the umb-media-grid.html to ensure, it's completely correct and up-to-date and so that removes this link, so no need to transform it into a button, as it no longer exists!

marcemarc:
> In this image its much closer

I think I've updated the way this is laid out in the search instance, so that it's not tight to the line, but it's the absence of the breadcrumb markup that makes it tricky to make it exactly the same? -  without restyling the breadcrumb a little too... 'flex all the things' but I might not be the best person to tackle that!

marcemarc:
@nul800sebastiaan thanks!, I think I've addressed the requested updates...

src/Umbraco.Web.UI.Client/src/views/components/umb-media-grid.html
nul800sebastiaan: @marcemarc What I meant.. instead of having actions happen by clicking on a fake link, we should change it into a button. I'll update this one!

nul800sebastiaan: Aha.. I see what happened now.. indeed, no need to turn it into a button if it's not there any more! Merges get confusing haha!

nul800sebastiaan:
Alright, I went through it all again, tested, all good now, thanks very much for the work and your patience! ⭐ 

Great addition, bring on v8.8! 🎉

marcemarc:
Yey, Thanks @nul800sebastiaan !

bjarnef:
@nul800sebastiaan I have submitted a PR to remove this function: https://github.com/umbraco/Umbraco-CMS/pull/8497

Issue #7914, opened 2020-04-08
Upgrade from v7 to v8.6+ throws exception on media URL migration


benjaminc:
## Overview
In trying to upgrade a database that uses RTEs and Grid heavily, the upgrade fails if I got directly to v8.6.0.  It works if I first go to an earlier version, and then try to go to v8.6.0.  The error is in the ConvertTinyMceAndGridMediaUrlsToLocalLink migration, on line 113.

### Umbraco version

I am seeing this issue on Umbraco version: v8.6.0, from v7.15.3

### Bug summary

This is caused by the ConvertTinyMceAndGridMediaUrlsToLocalLink migration using a service reference to retrieve data instead of using direct database access.  Thus when new columns were introduced in the AddPropertyTypeValidationMessageColumns migration, the service is trying to read from these new columns that don't yet exist in the database.

### Steps to reproduce

Create a v7.15.3 database, and add Grid and RTE data that contain media links.  Then try to upgrade the database to v8.6.0.

### Expected result

The site is upgraded to v8.6.0 successfully.

### Actual result

The upgrade wizard fails with the message "The database failed to upgrade. ERROR: The database configuration failed with the following message: Invalid column name 'mandatoryMessage'. Invalid column name 'validationRegExpMessage'. Please check log file for additional information (can be found in '/App_Data/Logs/')".  Reviewing the trace log shows the following stack trace for that error:
```
System.Data.SqlClient.SqlException (0x80131904): Invalid column name 'mandatoryMessage'.
Invalid column name 'validationRegExpMessage'.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   at System.Data.SqlClient.SqlDataReader.get_MetaData()
[...]

PR #7915, opened 2020-04-08
Remove dependency on Media Service in migration

benjaminc:
### Prerequisites

This fixes the bug at #7914 

### Description
Replaces a call to the Media Service with direct database access, so that the migration code is more resilient to changes introduced by newer versions.


emmaburstow:
Hi @benjaminc,

Thanks for the work here. We'll take a look and let you know if we've any questions or comments. 

Emma

Issue #7919, opened 2020-04-09
Models Builder fails on Member after deleting 'Date Picker with time' data type


ronaldbarendse:
After removing the default 'Date Picker with time' data type, Models Builder fails to build the Member model with this error:
```
Failed to build models.
Could not find a datatype with identifier -36.
Parameter name: id

   at Umbraco.Core.Models.PublishedContent.PublishedContentTypeFactory.GetDataType(Int32 id) in D:\a\1\s\src\Umbraco.Core\Models\PublishedContent\PublishedContentTypeFactory.cs:line 95
   at Umbraco.Core.Models.PublishedContent.PublishedPropertyType..ctor(String propertyTypeAlias, Int32 dataTypeId, Boolean isUserProperty, ContentVariation variations, PropertyValueConverterCollection propertyValueConverters, IPublishedModelFactory publishedModelFactory, IPublishedContentTypeFactory factory) in D:\a\1\s\src\Umbraco.Core\Models\PublishedContent\PublishedPropertyType.cs:line 69
   at Umbraco.Core.Models.PublishedContent.PublishedPropertyType..ctor(IPublishedContentType contentType, String propertyTypeAlias, Int32 dataTypeId, Boolean isUserProperty, ContentVariation variations, PropertyValueConverterCollection propertyValueConverters, IPublishedModelFactory publishedModelFactory, IPublishedContentTypeFactory factory) in D:\a\1\s\src\Umbraco.Core\Models\PublishedContent\PublishedPropertyType.cs:line 47
   at Umbraco.Core.Models.PublishedContent.PublishedContentTypeFactory.CreateCorePropertyType(IPublishedContentType contentType, String propertyTypeAlias, Int32 dataTypeId, ContentVariation variations) in D:\a\1\s\src\Umbraco.Core\Models\PublishedContent\PublishedContentTypeFactory.cs:line 67
   at Umbraco.Core.Models.PublishedContent.PublishedContentType.EnsureMemberProperties(List`1 propertyTypes, IPublishedContentTypeFactory factory) in D:\a\1\s\src\Umbraco.Core\Models\PublishedContent\PublishedContentType.cs:line 98
   at Umbraco.Core.Models.PublishedContent.PublishedContentType..ctor(IContentTypeComposition contentType, IPublishedContentTypeFactory factory) in D:\a\1\s\src\Umbraco.Core\Models\PublishedContent\PublishedContentType.cs:line 30
[...]

PR #7920, opened 2020-04-09
Fix EnsureMemberProperties data type IDs

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes: https://github.com/umbraco/Umbraco-CMS/issues/7919.

### Description
This updates the `EnsureMemberProperties` method to use the correct data type IDs, so no exceptions are thrown while generating models using Models Builder.


nul800sebastiaan:
Makes sense, thanks @ronaldbarendse! 👍

PR #7923, opened 2020-04-10
Move the infinite overlay to the same source order as regular overlays

BatJan:
### Prerequisites

### Description
I have just decided to close #4526 without having it merged in it's current state since the stuff I made no longer works.

Therefore I decided to have a deeper look at some of the root causes to some of the issues I faced when trying to implement the focus lock. One of them being that regular overlays and infinite overlays got triggered from different places within the source code meaning I had to do so weird little hacks to get it working. Move the infinite overlay out of the source order it will be a lot simpler to implement a well working focus lock - So this is what I have done with this PR 😃 

The `<umb-editors>` component has been moved to the same level in the source code as the `<umb-overlay>` component, since it's an overlay too really. I have also removed the `<div class="umb-editor__overlay" ng-if="infiniteMode"></div>` and added `<umb-backdrop ng-if="backdrop.show || infiniteMode">...<div>` to make the backdrop behave much more aligned with what you would expect from an overlay meaning that you now can't click on the content tree using your mosue - Just like with regular overlays making the overall experience much more consistent and improves the UX.

When this PR is hopefully accepted I will start looking at adding the focus lock again since this change will make it a lot easier to achieve 😄 

Visually this is what it acts like before and after

**Before**
umb-editors-before

**After**
umb-editors-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/editor/umb-editor.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/editor/umb-editor.less
@@ -1,6 +1,7 @@
 .umb-editors {
     .absolute();
     overflow: hidden;
+    z-index: 7501;
 
     .umb-editor {
         box-shadow: 0px 0 30px 0 rgba(0,0,0,.3);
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Sweet, works well, thanks @BatJan! ⭐

PR #7924, opened 2020-04-10
Replaced angular.toJson with Utilities.toJson [#CanConHackathon]

Matthew-Wise:
https://trello.com/c/UTjBLFQd/17-replace-instances-of-angulartojson

Decouples angulars toJson to es6 this is a direct port of the angular code to ensure we keep its features. I also ported the unit tests to make make testing easier for everyone.

Matt

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbrawmodel.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbrawmodel.directive.js
@@ -49,7 +49,7 @@ angular.module("umbraco.directives")
 				function jsonToString(object) {
 					// better than JSON.stringify(), because it formats + filters $$hashKey etc.
 					// NOTE that this will remove all $-prefixed values
-					return angular.toJson(object, true);
+					return Utilities.toJson(object, true);
 				}
 
 				function isValidJson(model) {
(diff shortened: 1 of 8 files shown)


nul800sebastiaan:
Great! And the tests work in the DevOps pipeline as well!

image

bjarnef:
@Matthew-Wise do we have an equivalent to `angular.fromJson()` as well?
It seems this PR replacing `angular.fromJson()` with `JSON.Parse()` is causing issues, when inserting an image grid editor in grid content.
https://github.com/umbraco/Umbraco-CMS/pull/7952

`angular.fromJson()` is also checking if value is a string:

It seems to work when replacing the getter here with the following:
https://github.com/umbraco/Umbraco-CMS/blob/a5ff8a429b83fb6263617eb2396aa1e80d1e55a4/src/Umbraco.Web.UI.Client/src/common/services/util.service.js#L208-L210

```
get: function (key) {
      var val = storage["umb_" + key];
      return Utilities.isString(val) ? JSON.parse(val) : val;
}
```

But I think we should have an `Utilities.fromJson` as well and use this here instead.

Matthew-Wise:
Jan had already started doing from :) https://github.com/umbraco/Umbraco-CMS/pull/7952

bjarnef:
@Matthew-Wise the PR https://github.com/umbraco/Umbraco-CMS/pull/7952 just replaced `angular.fromJson()` with `JSON.parse`. It didn't added a equivalent `Utilities.fromJson`.

It might work in some cases when wrapped inside `try/catch`, but here it broke inserting an image grid editor in grid content. https://github.com/umbraco/Umbraco-CMS/pull/7952/files#diff-0f50c5d5842e7132955874d51f872a84L209

PR #7928, opened 2020-04-10
Replace instances of angular.forEach with vanilla JS forEach

BatJan:
### Prerequisites
https://trello.com/c/71Cjd6cq/8-replace-instances-of-angularforeach

### Description
This PR refactors each instance of angular.forEach to vanilla JS forEach as mentione din the Trello link above 😃 

I realize it's a TON of files to review so please let me know if there is anything I can do to make your jobs easier 👼

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/localization/localize.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/localization/localize.directive.js
@@ -30,7 +30,7 @@ angular.module("umbraco.directives")
     .directive('localize', function ($log, localizationService) {
         return {
             restrict: 'E',
-            scope:{
+            scope: {
                 key: '@',
                 tokens: '=',
                 watchTokens: '@'
@@ -40,13 +40,13 @@ angular.module("umbraco.directives")
             link: function (scope, element, attrs) {
                 var key = scope.key;
                 scope.text = "";
-                
+
                 // A render function to be able to update tokens as values update.
                 function render() {
                     element.html(localizationService.tokenReplace(scope.text, scope.tokens || null));
                 }
-                
-                localizationService.localize(key).then(function(value){
+
+                localizationService.localize(key).then(function (value) {
                     scope.text = value;
                     render();
                 });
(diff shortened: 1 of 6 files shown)


emmaburstow:
Thanks Jan!!!

BatJan:
Ok, so I was a bit optimistic and naive about this one so I just did a hard reset to the initial commit. Started seeing a lof of "undefined" error messages in the console window - Some places we need to wrap the stuff we loop inside if statements it appears - I suppose angular.forEach does some kind of magic behind the scenes so it's not neccessary to check... It's quite a big task with 156 entries containing "angular.forEach" 😅

BatJan:
@emmaburstow @nul800sebastiaan @nathanwoulfe Ok, I think I will leave this one at the few replacements I've already done.

According to my search for angular.forEach in vs code there is still 143 instances in 55 files where this need to be fixed. But I'm feeling insecure about doing them all - I actually HAD done them all but when I started poking around the backoffice I would get many `.forEach is undefined` error messages - So reset the whole thing and did a few where I knew how to test if stuff still worked. I need to aim for simpler things to pickup where I feel more confident I won't break everything because I might need to add an `if` check as well 😅 

But please do review the few changes I've made 😃 

@nathanwoulfe I will put the card back to the "Things to do" once this has been merged so someone braver than me can pickup the task again 👼

nul800sebastiaan:
Thanks @BatJan, these are indeed good to go! 👍

bjarnef:
@nul800sebastiaan maybe some references on `angular.forEach` wasn't replaced or some PR's have been merged including `angular.forEach` instead the use of `Utilities.forEach`.

At least of seems there are still some use of `angular.forEach` in `v8/contrib` branch:
https://github.com/umbraco/Umbraco-CMS/search?p=1&q=angular.forEach

bjarnef:
@nul800sebastiaan I have submitted a new PR here: https://github.com/umbraco/Umbraco-CMS/pull/10759

nathanwoulfe:
Ideally, eventually, these would all be replaced with native forEach. No library/framework necessary, but requires a little more code change than substituting for Utilities...

bjarnef:
@nathanwoulfe the `Utilities.forEach` should also be native js forEach - just with an extra check that array not is null or undefined :)
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/utilities.js#L109-L117

PR #7931, opened 2020-04-10
angular.isObject replace to Utilities.isObject

adam-werner:
https://trello.com/c/ZQ6Pv900/13-replace-instances-of-angularisobject

Replaces the angular.isObject to the Utilities.isObject as part of the Umbraco's AngularJS divorce


nul800sebastiaan:
And this one is also great, thanks again @adam-werner! 👍

PR #7936, opened 2020-04-10
V8-Layout Selector Directive - Button tag isn't closed? leaving invalid markup

marcemarc:
It looks like to me there is an 'extra' closing div in this markup... but I think this is because the button isn't closed, there is a closing div where the closing button should be? - eg the icon and span are part of thee button... I've made this change locally and it operates fine, but then it operating fine before hand too...

### Prerequisites

If there's an existing issue for this PR then this fixes - there isn't one

### Description
Was just looking at the markup for the umb layout selector directive to try to fathom how to use it elsewhere, and spotted an additional closing div... which led me to conclude there is in fact a missing closing button tag...

You can view the layout selector in the Media section of Umbraco, it's used to switch between grid and list view...


emmaburstow:
Thank you Marc! We'll take a look asap 

Em

nul800sebastiaan:
Good catch @marcemarc, thanks! 🍻

PR #7938, opened 2020-04-10
Remove code that does UA sniffing and explicit stuff for IE11 only

BatJan:
### Description
Working on #7928 I noticed that there was some IE11 related code in this directive and I could not resist the urge to remove it 😄 

However it appears that back in 2015 a bugfix was made adding a check to Edge in the code as well. So I'm not 100% sure it's safe to remove. However I'm not sure how to test if the original Edge issue is still there since https://issues.umbraco.org/issue/U4-7541 is not very descriptive.

But I have not noticed anything looking/acting weird in regards to the resizing in Edge after removing this functionality. But someone might need to do an extra thourough check 😅


nul800sebastiaan:
All good @BatJan! Thanks! 👍

Issue #7939, opened 2020-04-10
Various Pre-values and Data not migrated


benjaminc:
## Overview
I came across some data that was not being migrated correctly for color pickers.  On doing further investigate with a test database, I found multiple cases where pre-values were not correctly converted, and a few cases where data was not being correctly converted, all for Umbraco property editors.

### Umbraco version

I am seeing this issue from Umbraco v7.15.3 to Umbraco v8.6.0

### Bug summary

Various pre-values and some data values are not converted correctly.

### Specifics

Using the test DB and media folder that can be found at https://proworkscorporation.sharepoint.com/:u:/s/ProWorks/Eaz2pIHhfpNEqhRd7sSIoGkBr3fUaW2QOWT4v7NO0o0g4g?e=xv3q5t

The following was not converted correctly:
- Pre-values
    1. **Color Picker** - Colors not converted
    2. **Grid** - RTE Code option is not converted to Source Code Editor
    3. **Multinode Treepicker** - Does not preserve "Allow items of type"
    4. **Nested Content** - Document Type and Tab are not converted to Element Type and Group properly
    5. **Repeatable Textstrings** - Does not preserve minimum and maximum values
    6. **Slider** - Does not preserve initial value, initial value 2, minimum value, maximum value, or step increments

- Data Values
    1. Nested Content
        - Checkbox List data is not converted
        - Dropdown data is not converted
        - Used document types are not marked as element data types

### Steps to reproduce

1. Create an Umbraco v7.15.3 site.
2. Copy the media files from the attached file into the media folder
3. Import the database from the attached file
4. Point the Umbraco site at the imported database
5. Start the site, and view the back-office
6. In a new folder, create an Umbraco v8.6.0 site.
7. Copy the media files from the attached file into the media folder
8. Import the database from the attached file into a new database
9. Point the Umbraco site at the second imported database
10. Start the site, and allow it to upgrade
11. In the back-office, view the data types and compare with the v7.15.3 data types
12. In the back-office, view the content nodes and compare with the v7.15.3 site

### Expected result

Data types pre-values should match when correlating pre-values exist in v8.

Content should be visible and match.

### Actual result

Various data type pre-values do not match (see specifics above)

The Extended content page throws an error due to Nested Content data not being mapped properly, and cannot even be edited.

benjaminc:
[...]

PR #7940, opened 2020-04-11
v8.6 Migration plan fix for error "Invalid column name 'mandatoryMess…

hfloyd:
v8.6 Migration plan fix for error "Invalid column name 'mandatoryMessage'. Invalid column name 'validationRegExpMessage'"

### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7914

### Description
To test: Create a v 7.15.4 website with some content, then follow the Migration instructions (https://our.umbraco.com/documentation/Getting-Started/Setup/Upgrading/migrating-to-v8#step-2-migrating-content-to-umbraco-8). The Migration should complete without throwing the "Invalid column name 'mandatoryMessage'. Invalid column name 'validationRegExpMessage'" error.

### DIFF
--- a/src/Umbraco.Core/Migrations/Upgrade/UmbracoPlan.cs
+++ b/src/Umbraco.Core/Migrations/Upgrade/UmbracoPlan.cs
@@ -179,6 +179,9 @@ namespace Umbraco.Core.Migrations.Upgrade
             To<ChangeNuCacheJsonFormat>("{80C0A0CB-0DD5-4573-B000-C4B7C313C70D}");
             // release-8.0.1
 
+            //Gotta do this first...even though it wasn't introduced until 8.6
+            To<AddPropertyTypeValidationMessageColumns>("{3D67D2C8-5E65-47D0-A9E1-DC2EE0779D6B}");
+
             // to 8.1.0...
             To<ConvertTinyMceAndGridMediaUrlsToLocalLink>("{B69B6E8C-A769-4044-A27E-4A4E18D1645A}");
             To<RenameUserLoginDtoDateIndex>("{0372A42B-DECF-498D-B4D1-6379E907EB94}");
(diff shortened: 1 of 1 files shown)


jaddie:
Hmm interesting that you've got this error, as I've just used the v8/v7-to-v8-migrations branch to migrate from 7.15.3 to "8.7", only when I got this error you mentioned did I compare the live 8.6 schema to the schema of my converted database did I notice loads of missing fields etc only to realise the migrations branch is massively out of date & hasn't seen a merge in over 800 commits or so. Are you sure you didn't end up following the same path perhaps?

nul800sebastiaan:
Thanks @hfloyd! I've left you a comment on the issue. 👍

tombille:
I tried this - migrating a cloud project running umbraco 7.15.4 to 8.6 - same error and the fix works

jaddie:
Can also confirm that this is an issue and moving the mentioned migration solves it, tested with the performance enhancement branch merged into latest V8 branch so running on V8.7.0

jaddie:
@Shazwazza Is there any kind of idea / plan when this is going to get merged in since anything related to the migrations not working is rather significant tbh when trying to adopt v8. Thankfully stevemegson has done stellar work in sorting out the mess of migrating and the performance issues, now there are just a few quick fixes that need merging in, its making it quite a pain to try and go with the idea of upgrading client's websites and working on the upsell in that area, only to be edgy because there is still not a proper working version of V8 with migrations in it yet.

benjaminc:
The only problem with this solution is that it breaks the state machine that is the migration plan.  If someone has already upgraded to something between 8.1 and 8.6, and then tries to go straight to 8.7 or wherever this will be included, then the `AddPropertyTypeValidationMessageColumns` migration never runs.

benjaminc:
I have created a proposed alternative PR #8337 which allows for successful upgrade both from v7 and from v8.1-v8.5.

nzdev:
Any progress on this issue? Now that v8.7 is out it would be good to have this fixed. Migrating large amounts of media / content is too slow till 8.7, but you can't migrate from v7 and take advantage of the speed up with this bug present.

benjaminc:
No progress yet, but if you want to work around this issue you can use the [ProWorks.Umbraco8.Migrations](https://www.nuget.org/packages/ProWorks.Umbraco8.Migrations) package (with [source code here](https://bitbucket.org/proworks/proworks.umbraco8.migrations/src/master/)) I wrote since I needed to get clients up to 8.7 and it hadn't included these changes yet.  It works around this issue with a custom IMediaService implementation that is used only during the initial v7 to v8 upgrade, and isn't used after that.  It also includes a bunch of fixes for the #7939 issue related to content that didn't get migrated properly from v7 to v8, which is another PR that hasn't been included yet, #7957.

PR #7942, opened 2020-04-11
Added elements to improve default security and caching for performance

jameswilki:
### Prerequisites

My first PR and there's no existing issue, but these web.config changes are something that I implement on the majority of Umbraco sites I build. I'm interested in any feedback to know if they could be included in the template release or not.

### Description - Security changes
Added `<httpCookies>` element to set `httpOnlyCookies="true"` to mark cookies as HttpOnly to help mitigate cross-site scripting (XSS) attacks
For a production site running HTTPS, the `requireSSL` attribute should be changed to "true" to ensure that cookies are marked as secure to help mitigate man-in-the-middle attacks.

Added more customHeaders to further tighten security, including cross-site scripting prevention, strict-transport-security and X-Frame options. Attempts to remove the server and ASP.Net version headers.

Added outbound rewrite rules to remove server header and enforce Strict-Transport-Security when the site is running HTTPS. This helps mitigate man-in-the-middle attacks.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security

### Testing - Security changes
Security changes can be tested before and after on https://securityheaders.com by entering your site URL. Tick the "Hide results" option to exclude your results from the public list.

### Description - Mime type changes
Added mime types for common media types and updated the existing font mime types to the newer values. This helps webservers serve files using the correct mime types.

### Testing - Mime type changes
Using the Network tab in browser developer tools, check the Type column before and after applying the changes.

### Description - Caching and compression
Added caching and compression for static and dynamic files to improve site performance.
The compression settings are IIS-specific, so if this causes problems, maybe they could be included, but commented out inititially?

### Testing - Caching and compression
Using the Network tab in browser developer tools, check the "Size" column before and after applying the changes. If set, you'll also need to uncheck the "Disable cache" option temporarily. Refreshing the page should result in a lot of files showing "cached" (Firefox) / "memory cache" (Chrome).
Firefox developer tools also lists the "Transferred" size next to the resource "Size".

Thanks, and kind regards,

James.

### DIFF
--- a/src/Umbraco.Web.UI/web.Template.config
+++ b/src/Umbraco.Web.UI/web.Template.config
@@ -87,6 +87,8 @@
 
         <httpRuntime requestValidationMode="2.0" enableVersionHeader="false" targetFramework="4.7.2" maxRequestLength="51200" fcnMode="Single" />
 
+        <httpCookies httpOnlyCookies="true" requireSSL="false"/>
+
         <httpModules>
             <add name="ScriptModule" type="System.Web.Handlers.ScriptModule, System.Web.Extensions, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31BF3856AD364E35" />
             <add name="UmbracoModule" type="Umbraco.Web.UmbracoModule,Umbraco.Web" />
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Hey thanks @jameswilki - unfortunately we can't just add loads of things to the default web.config as it will a ton of issues on different configurations. For sure a lot of these things are great to add for a lot of people but they should be in the documentation as recommended security practices instead.

The current web.config default have been carefully built up after years and years of trying out little changes and making sure they are the minimally necessary entries that we need. These additional steps will quite drastically change the behavior of people's sites and they haven't "opted in" so to speak. Most of these changes need to be something the people consciously do so that they know what they're doing and why. 

Thanks for making a detailed PR with a great description of the changes. Sorry we can't accept it but I would encourage you to describe them in the documentation so people know they can opt in to them and why they would opt in to them.

Do note that all of the caching changes are probably not something we should recommend, you really need to know what you're doing there and how to clear that cache, etc. Also the gzip options will not work on every configuration so we probably can't recommend those either.

Thanks again!

jameswilki:
Thanks for the detailed feedback @nul800sebastiaan, I figured there'd be good reasons for some of the items not being there by default. I'll look at the documentation side of things.

Kind regards,

James.

PR #7946, opened 2020-04-11
Append "active" to currently selected section for screen readers

glombek:
### Prerequisites

### Description

Accessibility fix.

Active tab is visually identified but not for screen readers, this commit provides some auditory cues to the user to ensure they are not lost within the navigation.

The currently selected section is marked as the ARIA current page `aria-current="page"`. This has limited browser/screen reader support so he word "active" is appended to currently selected section for screen readers (with `aria-label`) as well.

To test, enable a screen reader and move along the section links along the top of the page. The currently selected section will indicate it is selected in the auditory description. Changing section should also update the description of the pages. Also test at a smaller resolution so that the section navigation items collapse into a "..." menu as shown below.

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/application/umb-sections.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/application/umb-sections.html
@@ -7,7 +7,7 @@
                    ng-dblclick="sectionDblClick(section)"
                    ng-click="sectionClick($event, section)"
                    prevent-default
-                   aria-current="{{section.alias == currentSection ? 'page' : null}}"
+                   ng-attr-aria-current="{{section.alias == currentSection ? 'page' : undefined}}"
                    aria-label="{{section.name + (section.alias == currentSection ? ' (active)' : '')}}">
                     <span class="section__name">{{section.name}}</span>
                 </a>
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks @glombek - this is working great so far! The only thing I tried to have a look at was making the `aria-current` appear only conditionally. Usually I'd do that by doing something link `ng-attr-aria-current="{{section.alias == currentSection ? 'page' : undefined}}"` but that really doesn't seems to want to work. 

If you have any ideas then that would be good, but since it's only about max 10 elements in the DOM I don't think we need to take heroic measures to fix that. Let me know if you can make it work, but don't spend too much time on it!

glombek:
Thanks @nul800sebastiaan, I wasn't aware of the `ng-attr` prefix! It was annoying me that the attribute was being rendered without a value, though it was still valid according to the spec.

I couldn't find any official documentation that an attribute wouldn't be added if the value was `null` or `undefined`, only anecdotal references to it!

However, with my testing it appears `null` renders the attribute as empty, while `undefined` does not output the attribute 😕

So, I've amended my solution.

nul800sebastiaan:
Thanks again @glombek! I just tried out this build and I see the same problem I saw when I tried to update this locally, it looks like this for me: 

image

So I am not sure how you've concluded that it does not output the attribute? 😁

glombek:
Yes, the `ng-attr-aria-current` attribute exists, but `aria-current` does
not, which is the expected behaviour.

nul800sebastiaan:
Aaah I see, that makes total sense! Then it works exactly right, awesome! 

Woohoo, congrats on your first accepted PR to Umbraco! 🎉

I've added a shiny new contributor badge to your profile on Our https://our.umbraco.com/members/glombek 🏅 

Keep up the good work, hope to see more from you in the near future!

PR #7948, opened 2020-04-11

AndyButland:

- Users and related extension methods
- Path validation

I found an issue with some existing user tests where we re-use the same builder for creating multiple entities.  To resolve this I've introduced a change where the builders reset themselves after building the object, so they are "fresh" to construct a new one.


bergmania:
@AndyButland 
Why do you think it is a good idea with a reset method, compared to just new up a new builder? It should be very cheep to new up a builder?

Also if it resets every time, we cannot build many nodes with the same settings. E.g:
```
     var builder = new LanguageBuilder().WithFallbackLanguageId(1);
     var languages = Enumerable.Range(1, 10).Select(x => builder.Build());
```

In this (bad) example, the builder should create 10 languages, with `1` as fallback language Id, but random data for the key value. In theoery the CultureInfo could also be chosen random.

AndyButland:
Yes, I debated that - but went with the reset mainly to support the pattern of tests we currently have, which is a single, class level builder (e.g. `private readonly MacroBuilder _builder = new MacroBuilder();`), and it seems just a little less neat to have some tests like this, and others, where they need multiple entities created, have to have to remember to new up new builders inside the test.

From a little reading on the pattern too it seemed to be a common, albeit optional, thing to do.

The way I have it would make your example something like:

```
    var languages = Enumerable.Range(1, 10)
        .Select(x =>_builder.WithFallbackLanguageId(1).Build());
```

Other than newing up a new builder, another way might be to have a reset method that you have to call if you want to use it, so you could do:

```
var entity1 = _builder.WithId(11).WithParentId(10).Build();
var entity2 = _builder.WithId(12).Build();	        // Would have ParentId = 10
var entity3 = _builder.Reset().WithId(13).Build();	// Would have ParentId = whatever default is
```

But overall just thought having the fact that builders are always "use once" was a little cleaner than either of these two options.

Not sold 100% though... seems nicer to me but you can convince me otherwise :-).

bergmania:
I'm not sure I can convince you, as it overall is not such a big deal.

My only issue with the approach here 
```
   var languages = Enumerable.Range(1, 10)
        .Select(x =>_builder.WithFallbackLanguageId(1).Build());
```
will be with complex scenarios where we call many `With*` or `Add*` methods.

IMO, it is just more complex to implement a builder, if I also have to implement the `Reset` method. I'm also afraid of the risk for not remember all fields when resetting.

AndyButland:
Yes, both valid points - though in the case of the more complex object we could pull out a helper method, so only defined the `With*`s and `And*`s once.

I put the `Reset` on the abstract class so you can't forget it, but it is possible to forget to reset certain fields of course (perhaps no more than forgetting to build them, but less obvious certainly).  I did wonder if the `Reset` could be implemented on the abstract base class, using reflection, but it doesn't seem straightforward.  Whilst you could likely find all the private fields, for some we set to null, others to a new empty object or collection.

bergmania:
Hmm.. 
The fields is maybe easy to do, but the builders are complex to do..  I still think it is more clean without a reset method, but don't have more points and can live with the reset method. Maybe we need a tiebreaker :D 
@Shazwazza thoughts?

Shazwazza:
I'm not a huge fan of `Reset` as to me it seems to make things less simple. For this example:

```
var entity1 = _builder.WithId(11).WithParentId(10).Build();
var entity2 = _builder.WithId(12).Build();	        // Would have ParentId = 10
```

I would have thought that WithId(12) would overwrite the ID ?

Shazwazza:
oh... don't mind me... that example makes sense, i misread :P

But IMO that example makes sense.

I would think this would just work:

```
var entity1 = _builder.WithId(11).WithParentId(10).Build();
var entity2 = _builder.WithId(12).Build();	        // Would id 12 and ParentId = 10
var entity3 = _builder.WithId(13).Build();	// Would have id 13 and ParentId = 10
```

If I wanted to 'clear' something i would just make a new builder. To me that makes more sense.

AndyButland:
OK - consider the tie-break broken!  Have updated to remove the `Reset()` method.

Please have a look at the test `PathValidationTests.Ensure_Path_Entity_Valid_Recursive_Parent()` now though, as you'll see here an example where we need to reset - or rather recreate - the builder each time.  And it's not just that the builder's state is used on multiple calls within the test, it also leaks from what's set in other tests too - meaning you can get odd errors depending on what order the tests get run (or they work when run on their own, but not as a set).

Given that, without the resetting, we might be better not doing this...

```
private EntitySlimBuilder _builder = new EntitySlimBuilder();
```

... in each test class.

But rather this:

```
private EntitySlimBuilder _builder;

[SetUp]
private void SetUp()
{
   _builder = new EntitySlimBuilder();
}
```

Thus ensuring we have a "clean" builder for each test.  Thoughts?

bergmania:
I agree or even skip the field completely and new up a builder, when you need it.

AndyButland:
OK - have updated the tests to use this `SetUp` method, ensuring all tests within the class, including any that might get added in future, will get a "clean" builder.

bergmania:
The nuget issue was temporary. I'm happy with the changes now :) Good work @AndyButland  💪

PR #7951, opened 2020-04-11
Replace instances of angular.element()

BatJan:
### Prerequisites

This PR addresses the issue mentioned here https://trello.com/c/yxsY0JYW/5-replace-instances-of-angularelement

### Description
I have replaced all instances of `angular.element()` except from the `LegacySpeechBubble.js` file since I'm not sure where this might be used if it's still being used at all. At a first glance it seems like we could actually delete this file instead?


nul800sebastiaan:
Good stuff @BatJan! 👍

PR #7952, opened 2020-04-11
Replace angular.fromJson with JSON.parse

BatJan:
### Prerequisites

Fixing this issue https://trello.com/c/uPQwJmDQ/9-replace-instances-of-angularfromjson

### Description
I have been replacing all instances of `angular.fromJson` with `JSON.parse`. 😃


nul800sebastiaan:
Thanks! 👍

bjarnef:
It seems this is causing issues now, when inserting e.g. image grid editor in grid content.
https://github.com/umbraco/Umbraco-CMS/blob/a5ff8a429b83fb6263617eb2396aa1e80d1e55a4/src/Umbraco.Web.UI.Client/src/common/services/util.service.js#L209

where `angular.fromJson()` is a bit different:

With `angular.fromJson()` I can open the image dialog:

image

with just the current change using `JSON.Parse()`:
image

It seems to be an empty when session storage is empty, where JSON.Parse throw an error.

nul800sebastiaan:
@bjarnef Hmmm.. Maybe it's the same issue as https://github.com/umbraco/Umbraco-CMS/pull/8013 though? And not so much an issue with this change?

bjarnef:
@nul800sebastiaan not sure if the issue @lars-erik is having is the same since it is in v8.6.x ...
anyway I am working on a PR regarding the issue, when inserting an image grid editor in grid content:
https://github.com/umbraco/Umbraco-CMS/pull/8014

lars-erik:
@nul800sebastiaan @bjarnef This is not related to #8013, but makes it so you can't even get to #8013. 😆  
I didn't notice since I copied the change to the controller into built output in a soon-to-be-prod-ready 8.6.1 site, so didn't get the services update that blocks it. Now doing more work for a possible second PR based on the contrib branch where I merged #8014 and about to test more.

PR #7953, opened 2020-04-11
Remove unused "umb-password-toggle" directive

BatJan:
### Description
When working on some other PR's I came noticed this directive does not appear to be used anywhere when searching through the source code - So I suggest we remove it for good 😃


nul800sebastiaan:
Thanks @BatJan! I've marked this as breaking on the off chance that someone is using this directive somewhere in their custom code.

PR #7954, opened 2020-04-12
Make the 'GetBackOfficeFields' methods virtual

marcemarc:
When implementing your own IUmbracoTreeSearcherFields, would it be good to be able to inherit from this default UmbracoTreeSearcherFields implementation, and then override whichever method you'd like to manipulate to add an additional field to, instead, and I'm not super sure, would you have to implement all four methods in your custom implementation?

I was just looking at @bielu 's PR for the docs, and was trying to work out how best to suggest a realworld example, and from @Shazwazza 's comments on the original PR the intention with this approach is to make it easy to allow people to override the SearcherFields list as required... so was thinking the sort of example to put into the docs would be something like this:

```
public class CustomTreeSearcherFields : UmbracoTreeSearcherFields
    {
        public override IEnumerable<string> GetBackOfficeDocumentFields()
        {
            var currentDocumentFields = base.GetBackOfficeDocumentFields().ToList();
            currentDocumentFields.Add("bodyText");
            return currentDocumentFields;
        }
    }
```
But I don't think that's possible if those methods aren't virtual?
But also I may not have understood correctly!

To test this PR Create the class as above and register using a composer:

```
        {
            composition.RegisterUnique<IUmbracoTreeSearcherFields, CustomTreeSearcherFields>();
        }
```
Using the starter kit, you would be able to 'search' the backoffice for terms included in any bodyText fields in the site.

### Prerequisites

### DIFF
--- a/src/Umbraco.Web/Search/UmbracoTreeSearcherFields.cs
+++ b/src/Umbraco.Web/Search/UmbracoTreeSearcherFields.cs
@@ -7,23 +7,23 @@ namespace Umbraco.Web.Search
     public class UmbracoTreeSearcherFields : IUmbracoTreeSearcherFields
     {
         private IReadOnlyList<string> _backOfficeFields = new List<string> {"id", "__NodeId", "__Key"};
-        public IEnumerable<string> GetBackOfficeFields()
+        public virtual IEnumerable<string> GetBackOfficeFields()
         {
             return _backOfficeFields;
         }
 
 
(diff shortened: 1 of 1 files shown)


bielu:
@marcemarc they are implementing an interface and could be overridden as @Shazwazza show here:
https://github.com/Shazwazza/Examine/pull/135
but anyway that way which I showed in that code will work anyway, as we are using interface everywhere :)

marcemarc:
@bielu awesome, I got the VS warning but didn't think! thanks!

PR #7955, opened 2020-04-12
Activate backdrop when drawer is open

BatJan:
### Prerequisites

### Description
I noticed that when hitting the "Help" link in the header menu the drawer would open without the backdrop being activated. So with this PR I fixed this so it acts more like whenever we open a dialog withing the rest of the backoffice UI - In example overlays, infinite overlays etc.

This also means that it's not possible to click on anything else using your mouse - Tabbing is currently still another story 😃 

**Before**
help-before

**After**
help-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/application/umb-drawer.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/application/umb-drawer.less
@@ -4,7 +4,7 @@
     bottom: 0;
     left: auto;
     right: 0;
-    z-index: 10;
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Thanks @BatJan. since this is not an overlay but a panel that lives next to the backoffice, we don't want to prevent interaction. In fact in the screenshot below you can see we're even encouraging interaction with the backoffice while the panel is open, which is now no longer possible! ;-)

So we'll keep it as it is and close this PR!

image

PR #7956, opened 2020-04-12
Add backdrop to "user overlay"

BatJan:
### Prerequisites

### Description
This PR does the same as in #7955 - Only that this is when the "User overlay" is activated 😃 

**Before**
user-details-before

**After**
user-details-after


nul800sebastiaan:
Thanks @BatJan, this one makes sense! 👍

PR #7957, opened 2020-04-13
Add migrations for various pre-value errors

benjaminc:
### Prerequisites

This resolves the issue at #7939 

### Description
Add migrations for color picker pre-values, slider pre-values, and multiple-textstring pre-values.  Also migrations to fix a couple pre-values that weren't migrated correctly on the grid, nested content, and multi-node tree picker.  Document types used on nested content data types were marked as elements if elements hadn't been setup in the DB already.  Finally, added a migration to update data embedded in nested content properties for checkbox lists, radio button lists and dropdowns.


emmaburstow:
Hiya @benjaminc,

Thanks for your work here and for the description. It gives us plenty to work with. Seb has linked it to issue #7939 too. We'll take a look and let you know if we need anything.

Emma

benjaminc:
Are these migration fixes targeted for 8.7, or were they not going to be included in that release?

benjaminc:
If anyone finding this wants to get fixes for these issues before the PR is merged, as well as a fix for #7914 which prevents you from going straight from v7 to v8.6+, you can use the ProWorks.Umbraco8.Migrations package (with source code here). It includes fixes for both issues while the PRs for them are under review.

PR #7961, opened 2020-04-15
Nucache NullReferenceException when copying

Shazwazza:
The issue was originally reported as part of another one here https://github.com/umbraco/Umbraco-CMS/pull/7907#issuecomment-610259393

It is not always possible to replicate this issue since it depends on the state that nucache is in and the issue could 'go away' upon site restart or cache rebuilding in some cases.

I've added a unit test to show the problem and then a commit that fixes it. 

The issue is because when a branch is refreshed and there is already more than 1 cache generation created and when there's a hierarchy in the branch, during the recursive call to `ClearBranchLocked` it was passing in `link.Value` but since we are clearing that `Value` within the method, in this exact case `link.Value` will end up being null once the logic returns from the recursion since it's a reference that was changed. The fix is to capture the reference locally and use that `Value` for the remainder of the loop since within that loop we aren't expecting the value to change.

### DIFF
--- a/src/Umbraco.Tests/PublishedContent/NuCacheChildrenTests.cs
+++ b/src/Umbraco.Tests/PublishedContent/NuCacheChildrenTests.cs
@@ -44,7 +44,7 @@ namespace Umbraco.Tests.PublishedContent
             _snapshotService?.Dispose();
         }
 
-        private void Init(IEnumerable<ContentNodeKit> kits)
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Cherry picked for 8.6.1 in b30db05cc7a1b69c7e9d9990ea2677f18473bae3

stefankip:
Just wanted to let you know it fixed our issue! Thanks again @Shazwazza 👍

Issue #7898, opened 2020-04-05
Netcore: migrate unit and integration tests into new projects


AndyButland:
Tangentially related to the .Netcore transition, and being worked on in the same `netcore/dev` branch, are some efforts to clean-up the Umbraco tests.  Currently they all sit within the `Umbraco.Tests` project, and are mix of unit and integration tests.

We'd like to move out the true unit tests into a new project `Umbraco.Tests.UnitTests`, making it easier for developers to run those tests that can run more quickly, more often.

We're also taking the time to refactor them, by introducing a [builder pattern](https://refactoring.guru/design-patterns/builder) for the construction of model classes used in tests.

A number have already been transitioned, so the necessary patterns have been established and various supporting classes created.  But there's quite a number still to do, hence this issue.

The idea is that this issue won't be resolved by a single PR, rather a number of smaller ones, each focused on migrating and refactoring one or a few classes of tests.  Given that, it's worth flagging on this issue any tests anyone plans to look at migrating and updating, so effort doesn't get duplicated.

Unit tests can generally be identified as those that either don't inherit from a base class (though there are some that inherit from `UmbracoTestBase` - normally putting them in the "integration test" category - but that don't really need to, so it's worth checking the test implementation).

The process I've been following when working on the model class tests, using as an example `MemberGroupTests`, is as follows:

- Pick a unit test class and move it from `Umbraco.Tests` to `Umbraco.Tests.UnitTests`.
- Adjust the namespace.
- Remove the inherited base class (if there is one).
- If there are multiple tests all creating the same object to test against, refactor this into a private helper method.
- Add a class level reference to a builder class - e.g. `private readonly MemberGroupBuilder _builder = new MemberGroupBuilder();`.  It's likely this class won't exist yet.
- Change the implementation of model classes construction to use the builder pattern.

E.g. this:

```
var group = new MemberGroup()
{   
    CreateDate = DateTime.Now,  
    CreatorId = 4,  
    Id = 6, 
    Key = Guid.NewGuid(),   
    UpdateDate = DateTime.Now,  
    Name = "asdf"   
};  
group.AdditionalData.Add("test1", 123); 
group.AdditionalData.Add("test2", "hello");
```

Becomes:

```
return _builder
    .WithId(6)
    .WithName("Test Group")
    .WithCreatorId(4)
    .AddAdditionalData()
        .WithKeyValue("test1", 123)
        .WithKeyValue("test2", "hello")
        .Done()
    .Build();
```

[...]

PR #7963, opened 2020-04-15
Netcore: Migrate UmbracoSettings tests from UnitTests to Integration Tests

emmagarland:
### Prerequisites

Based on the work to migrate and optimise unit tests https://github.com/umbraco/Umbraco-CMS/issues/7898

### Description

- I have moved UmbracoSettings tests from Unit Tests to Integration Tests, since they are reading from disk.
- I've fixed the issue where PHP and BMP cases weren't being set as expected.
- Moved the test helper method I needed for these tests into the Integration TestHelper. It may be that we can remove the use for this entirely but currently the config is being published and read from that path.
- Most other aspects of the tests currently remain as they were. There was no need for a builder as no complex object was being built.

### Comments

- In ContentElements.cs line 32, we default to a set of DisallowedUploadFiles if this is empty, whereas in the AllowedUploadFiles we default to an empty list. The tests are checking the default values of the UmbracoSettings file so I guess this makes sense, but just wanted to make sure this was intended behaviour - AllowedUploadFiles defaults to none, and disallowed defaults to the ones you listed. Same in ContentSettings.cs. This seems the safer way round, but since I'm in the tests I thought I'd mention
- Also, the tests were writing a load of info out to Debug, presumably because they weren't working. I have switched it to Console but it maybe we can delete it now there is no need for debugging those failing test cases
- I was getting issues with building the tests due to the Umbraco.Core namespaces locally, so I amended some of the existing ones. Let me know if that isn't the right convention


emmagarland:
Back to draft as getting unexpected behaviour with 2 tests - not ready for merge yet

emmagarland:
Ready again

PR #7965, opened 2020-04-16
Segments

nielslyngsoe:
This PR is for merging the Segments UI into our Dev branch.

**Description:**

This PR makes Segments available.
When enabled, UI will appear for properties, making it possible to determine whether a property will vary by segments.
The creation of segments is as well done by code.

To enable UI for enabling segmentation for a DocumentType, the server-variable "showAllowSegmentationForDocumentTypes" can be set to true.

But the UI still contains no way to create Segments, to do so read the description about test endpoints.

**UI Updates**
When segmentation is enabled for a DocumentType, the option to allow a property to vary by segments because available in the DocType Editor:
image

In DocumentType Editor, properties now have a label for Culture varying and Segmentation:
image

Variant selector is updated to show segment variants. If both culture and segment variants are present the segment-variants are placed as children of the culture variants.
image

Save, Publish, Unpublish etc. Dialogs have been updated to only show relevant choices:
image

**Test endpoints:**
For testing purpose segmentation can be enabled by end-points:

In web.config you should make sure you have the following <add key="Umbraco.TestData.Enabled" value="true"/> in the appSettings.

And you need to check that you have the Umbraco.TestData.dll
Otherwise, you can copy it from src\Umbraco.TestData\bin\Debug

To enable segments for a DocumentType call this endpoint:
http://localhost:8600/umbraco/surface/SegmentTest/EnableDocTypeSegments?alias=umbTestDataContent&propertyTypeAlias=desc

Creation of Segments can be done through this endpoint:
http://localhost:8600/umbraco/surface/SegmentTest/AddSegmentData?contentId=41366&propertyAlias=desc&value=hello%20world&segment=Test1

**Test notes:**
We need to test almost everything that deals with Variants.
Client-code for variants have been refactored to support an additional property to determine which variant to deal with.

This means that all editor parts dealing with variant content should be tested.

Known parts:
Content Editor
— Property Editors
— Variant Selector
— Save, publish, ... action dialogs

Content Apps


nielslyngsoe:
I'm closing this, to open vary similar PR based on a Feature Branch.
https://github.com/umbraco/Umbraco-CMS/pull/7973

PR #7969, opened 2020-04-17
New ToFastDictionary ext method and changes some GroupBy/ToDictionary to ToLookup

Shazwazza:
Based on a discussion about ToDictionary usages here https://github.com/umbraco/Umbraco-CMS/pull/7910/files#r409689045

I decided to just have a look to see where we could micro optimize some things and created a new ToFastDictionary ext method and another CopyToDictionary ext method. I also realized we are using GroupBy/ToDictionary in many places which in most cases can be better resolved and is much better for performance to use ToLookup so have changed a few of those (though there are more that could be done too).

### DIFF
--- a/src/Umbraco.Core/DictionaryExtensions.cs
+++ b/src/Umbraco.Core/DictionaryExtensions.cs
@@ -14,6 +14,20 @@ namespace Umbraco.Core
     ///</summary>
     internal static class DictionaryExtensions
     {
+        /// <summary>
+        /// Efficiently copies a readonly dictionary to a normal dictionary
+        /// </summary>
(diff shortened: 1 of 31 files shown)


Shazwazza:
@benjaminc this is pretty old but wondering if you or anyone else wanted to have a look at these changes, it's based on your comments on another PR from a while back: https://github.com/umbraco/Umbraco-CMS/pull/7910/files#r409959396

benjaminc:
@Shazwazza absolutely, I'll do that now.  Sorry I didn't see your comment about creating this back on Apr 16.  I saw the first one about thinking about creating it, then somehow completely missed the second about the PR!

PR #7973, opened 2020-04-17
Segment feature

nielslyngsoe:
This PR is for merging the Segments UI into our Dev branch.

**Description:**

This PR makes Segments available.
When enabled, UI will appear for properties, making it possible to determine whether a property will vary by segments.
The creation of segments is as well done by code.

To enable UI for enabling segmentation for a DocumentType, the server-variable "showAllowSegmentationForDocumentTypes" can be set to true.

But the UI still contains no way to create Segments, to do so read the description about test endpoints.

**UI Updates**
When segmentation is enabled for a DocumentType, the option to allow a property to vary by segments because available in the DocType Editor:
image

In DocumentType Editor, properties now have a label for Culture varying and Segmentation:
image

Variant selector is updated to show segment variants. If both culture and segment variants are present the segment-variants are placed as children of the culture variants.
image

Save, Publish, Unpublish etc. Dialogs have been updated to only show relevant choices:
image

**Test endpoints:**
For testing purpose segmentation can be enabled by end-points:

In web.config you should make sure you have the following <add key="Umbraco.TestData.Enabled" value="true"/> in the appSettings.

And you need to check that you have the Umbraco.TestData.dll
Otherwise, you can copy it from src\Umbraco.TestData\bin\Debug

To enable segments for a DocumentType call this endpoint:
http://localhost:8600/umbraco/surface/SegmentTest/EnableDocTypeSegments?alias=umbTestDataContent&propertyTypeAlias=desc

Creation of Segments can be done through this endpoint:
http://localhost:8600/umbraco/surface/SegmentTest/AddSegmentData?contentId=41366&propertyAlias=desc&value=hello%20world&segment=Test1

**Test notes:**
We need to test almost everything that deals with Variants.
Client-code for variants have been refactored to support an additional property to determine which variant to deal with.

This means that all editor parts dealing with variant content should be tested.

Known parts:
Content Editor
— Property Editors
— Variant Selector
— Save, publish, ... action dialogs

Content Apps


Shazwazza:
There's quite a lot of issues when testing this and saving cultures and segments. This is going to require more than just a 'review', it will require quite a lot of bug fixing and testing and I think the bugs will just need to be fixed as they are found. 

Segments published state isn't correct and just seems to show the published state of it's culture parent. 

When Saving a culture and/or segment that has validation issues after the result from the server for some reason the entire culture disappears in the culture drop down along with within the save/publish overlay causing the view to change to another culture which means you don't see validation problems. Not sure what happens to the culture but it's gone from the editor. 

Validation results are not working properly. There's quite a lot of issues here that just needs to be worked through. The keys from the server seem to be returning ok but are not being wired up in angular to it's property, culture (in the overlay), etc... when editing segments - of course need to test this all on normal cultures and non-cultures and non-culture with segments.

warrenbuckley:
Peer reviewed on a very long call with @nielslyngsoe

ronaldbarendse:
@nielslyngsoe Just a note on naming: aren't culture variants also part of the segmentation? So instead of 'Allow segmentation', it's actually more accurate (and consistent) to name this checkbox 'Allow vary by segment'?

https://github.com/umbraco/Umbraco-CMS/blob/c4456c53a054069e1d37c8813b07f02f39bfcfd3/src/Umbraco.Web.UI/Umbraco/config/lang/en.xml#L1666-L1675

Zweben:
Question for those involved with the segmentation feature: once this PR makes its way into a release, would you consider the underlying segmentation feature safe to use in production? I have a project that is ideally suited for segments (content localization), so I'm considering setting up a simple project-specific UI to enable their use, backed by code based off of SegmentTestController.cs. It would sure be a lot cleaner than the alternative I would have to use.

nielslyngsoe:
@Zweben there are parts of the content workflow that aren't fully working with segments. Meaning that if you want to use the Umbraco UI for publishing your segments etc. you will be experiencing weirdness.

Chris-N1:
Just a note based on my experience trying to get this to work as this seems to be missing from the PR above - you need the following 2 settings in web.config in appSettings section for this to work:
```
<add key="Umbraco.TestData.Enabled" value ="true" />
<add key="showAllowSegmentationForDocumentTypes" value="true" />
```

ronaldbarendse:
@Chris-N1 The `Umbraco.TestData.Enabled` setting is only required to be able to generate test content and is [documented alongside the code](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.TestData/readme.md).

The `showAllowSegmentationForDocumentTypes` setting does not seem to be honored, as it's fetched from the `ServerVariables` in the [document type permissions AngularJS controller](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/documenttypes/views/permissions/permissions.controller.js#L20) and that's [always set to `false`](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Editors/BackOfficeServerVariables.cs#L360).

Chris-N1:
@ronaldbarendse  OK - not sure what was happening then as I could not turn segmentation on at all for any document type/alias without that setting present. Possibly a bug? There's also this code in the SegmentTestController:

`if (ConfigurationManager.AppSettings["Umbraco.TestData.Enabled"] != "true")
                return HttpNotFound();`

ronaldbarendse:
You can't enable segmentation from the UI yet (that's still experimental and behind a fixed flag in code), but only from the `ContentTypeService` API. This is what the controllers in `Umbraco.TestData` do and they only work when you enable them using that application setting.

michaelchart:
Is there a plan for when the rest of the UI for working with segmented content might be implemented? i.e. when this feature is no longer experimental but fully available. We're trying to determine whether to wait for this or to implement a different approach for now.

nielslyngsoe:
> Is there a plan for when the rest of the UI for working with segmented content might be implemented? i.e. when this feature is no longer experimental but fully available. We're trying to determine whether to wait for this or to implement a different approach for now.

There is no current plan to develop this into a full feature. It only here to enables packages to make segmented data.

ronaldbarendse:
@nielslyngsoe Creating the segment names/content variants would indeed be something packages could enable, as segments could be different between nodes (e.g. for A/B testing, you can have different variants per node).

However, the document type configuration UI is already available within Umbraco, just hidden behind a flag. This flag currently can't be changed, as it's retrieved from the following server variable:

https://github.com/umbraco/Umbraco-CMS/blob/6da85dd47d0fe4dff7750755c637afddf0afa5fc/src/Umbraco.Web.UI.Client/src/views/documenttypes/views/permissions/permissions.controller.js#L20

But that's always set to `false`:
https://github.com/umbraco/Umbraco-CMS/blob/0bd4dced0b3e9205660114406b7e814f817179c7/src/Umbraco.Web/Editors/BackOfficeServerVariables.cs#L362

It would make sense to remove this flag (and always show the `Allow segmentation` checkbox (which should be updated to `Allow vary by segment` to be consistent with cultures IMHO) or actually make it configurable by editing the UmbracoSettings.config file.

PR #7977, opened 2020-04-19

AndyButland:
Another round of migration in this PR for issue #7898 - moving tests into the new unit tests project and builder pattern.

Don't think there's anything particularly controversial in this one; I've not adjusted any patterns.

Only things I can think of that might be worth calling out are:

- I removed one test (which was ignored and looked like was there for some one-off profiling),  It needed various dependencies that we don't currently have, so didn't think it was worth preparing them for this (in `ContentTypeTests`).
- I've introduced a few extension methods in `ContentTypeBuilderExtensions` that can be used to create full, sample content types for tests with much shorter syntax.


src/Umbraco.Tests.Common/Builders/Extensions/ContentTypeBuilderExtensions.cs
```diff
@@ -0,0 +1,143 @@
+﻿using Umbraco.Core;
+using Umbraco.Core.Models;
+
+namespace Umbraco.Tests.Common.Builders.Extensions
+{
+    public static class ContentTypeBuilderExtensions
+    {
+        public static ContentType BuildSimpleContentType(this ContentTypeBuilder builder)
```
bergmania, line 8: This method should in theory be closer to what a `new ContentTypeBuilder().Build()` should do.

I think we can use more default value in the builders: E.g.
- icon = `icon-document`
- thumbnail = `folder.png`
- path = `_path ?? $"-1,{GetId()}"`

src/Umbraco.Tests.Common/Builders/Extensions/ContentTypeBuilderExtensions.cs
```diff
@@ -0,0 +1,143 @@
...
+                .WithPropertyTypeIdsIncrementingFrom(200)
+                .AddPropertyGroup()
+                    .WithName("Content")
+                    .AddPropertyType()
+                        .WithPropertyEditorAlias(Constants.PropertyEditors.Aliases.TextBox)
+                        .WithValueStorageType(ValueStorageType.Nvarchar)
+                        .WithAlias("title")
+                        .WithName("Title")
+                        .WithSortOrder(1)
+                        .WithDataTypeId(-88)
```
bergmania, line 127: Why do we need to specific this explicit, and not use the default values?

src/Umbraco.Tests.Common/Builders/Extensions/ContentTypeBuilderExtensions.cs
```diff
@@ -0,0 +1,143 @@
...
+                        .Done()
+                    .AddPropertyType()
+                        .WithPropertyEditorAlias(Constants.PropertyEditors.Aliases.TextBox)
+                        .WithValueStorageType(ValueStorageType.Ntext)
+                        .WithAlias("bodyText")
```
bergmania, line 132: Why do we need to specific this explicit, and not use the default values?

src/Umbraco.Tests.Common/Builders/Extensions/ContentTypeBuilderExtensions.cs
```diff
@@ -0,0 +1,143 @@
...
+                .WithId(10)
+                .WithAlias("memberType")
+                .WithName("Member type")
+                .WithCreatorId(22)
+                .WithDescription("test")
+                .WithIcon("icon")
+                .WithThumbnail("thumb")
+                .WithSortOrder(5)
+                .WithLevel(3)
+                .WithPath("-1,4,10")
```
bergmania, line 117: Any reason default values is not good enougth?

src/Umbraco.Tests.Common/Builders/Extensions/ContentTypeBuilderExtensions.cs
AndyButland: In general the reason I've done this is just to make sure I'm not changing what the test was previously doing - i.e. if it was setting a value before, I've left it doing it still.  But I do agree default values would serve the purpose of the tests in most cases just as well.  I'll have a look to where we can use these more.

AndyButland: Have made some updates to address this.

src/Umbraco.Tests.Common/Builders/Extensions/ContentTypeBuilderExtensions.cs
AndyButland: Have made some updates to address this.

src/Umbraco.Tests.Common/Builders/Extensions/ContentTypeBuilderExtensions.cs
AndyButland: Have made some updates to address this.

src/Umbraco.Tests.Common/Builders/Extensions/ContentTypeBuilderExtensions.cs
AndyButland: Have made some updates to address this.

bergmania:
Nice work, again :)

PR #7978, opened 2020-04-19
Update Models Builder model namespace in Views/web.config

callumbwhyte:
The Models Builder namespace was changed from `Umbraco.Web.PublishedContentModels` to `Umbraco.Web.PublishedModels` in V8...

It seems one of the transforms for `Views/web.config` wasn't updated - so I've done just that!

Weirdly the line I have updated doesn't appear in the `Views/web.config` in my projects, so I don't think this file is even used anymore... 🤷‍♂️


nul800sebastiaan:
It's there when I do a fresh install, and it's correct.. But it can't hurt to run this transform if it's not correct. Thanks @callumbwhyte!

PR #7983, opened 2020-04-20
ModelsBuilder enabled flag not respected

Shazwazza:
When MB is disabled we were still rebuilding all of nucache linked list data when schema changes are made which is a requirement of PureLive models, but if MB is disabled then PureLive models don't matter. This was causing unnecessary performance and db overhead when modifying document types when mb is disabled.

This is all done with no breaking changes by adding an additive interface: `ILivePublishedModelFactory2` which fixes a TODO that should have been cleared in 8.3 but we are getting to it now and also allows for us to know if the ILivePublishedModelFactory is enabled or not.

### DIFF
--- a/src/Umbraco.Web/PublishedCache/PublishedSnapshotServiceBase.cs
+++ b/src/Umbraco.Web/PublishedCache/PublishedSnapshotServiceBase.cs
@@ -6,16 +6,6 @@ using Umbraco.Web.Cache;
 
 namespace Umbraco.Web.PublishedCache
 {
-    internal class NotifiedEventArgs<T> : EventArgs
-    {
-        public NotifiedEventArgs(T[] payloads)
-        {
-            Payloads = payloads;
-        }
-
-        public T[] Payloads { get; }
-    }
-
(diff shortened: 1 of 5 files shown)


src/Umbraco.Web/PublishedCache/PublishedSnapshotServiceBase.cs
```diff
@@ -1,11 +1,22 @@
...
-    abstract class PublishedSnapshotServiceBase : IPublishedSnapshotService
+    internal class NotifiedEventArgs<T> : EventArgs
+    {
+        public NotifiedEventArgs(T[] payloads)
+        {
+            Payloads = payloads;
+        }
+
+        public T[] Payloads { get; }
+    }
```
bergmania, line 17: Why is this added but not used?

Shazwazza: mistake

Shazwazza: I've remove this now

PR #7984, opened 2020-04-20
Fixes #7967 - cannot start site with SqlMainDomLock  when db isn't configured

### LINKED ISSUE #7967: SqlMainDomLock when configured via appSettings cannot be used unless umbraco is installed
This is a setting for running Umbraco on Azure webapps but currently in order to use this setting Umbraco needs to first be installed otherwise a YSOD will occur on startup. 

Should be relatively easy to check the runtime state within SqlMainDomLock and ignore any database connections if the db is not available. 

---
_This item has been added to our backlog [AB#6274](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6274)_

Shazwazza:
Fixes SqlMainDomLock when configured via appSettings cannot be used unless umbraco is installed #7967


bergmania:
Looks good

PR #7989, opened 2020-04-21
Netcore: Implemented DisableBrowserCacheAttribute in .Net core.

AndyButland:
Towards #7991.

In this PR I've taken a copy of `DisableBrowserCacheAttribute` into `Umbraco.Web.BackOffice` and re-implemented using .Net Core.

I suspect we might actually be able to get rid of this and use the built-in `ResponseCache`, but we may want it for more control and is at least useful at the moment as a simple example reference for other migrations.

### DIFF
--- a/src/Umbraco.Web.BackOffice/Filters/DisableBrowserCacheAttribute.cs
+++ b/src/Umbraco.Web.BackOffice/Filters/DisableBrowserCacheAttribute.cs
@@ -10,19 +10,12 @@ namespace Umbraco.Web.BackOffice.Filters
     {
         public override async Task OnResultExecutionAsync(ResultExecutingContext context, ResultExecutionDelegate next)
         {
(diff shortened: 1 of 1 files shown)


AndyButland:
Just had a realisation on this - there are synchronous and asynchronous options for filter implementation (see [here](https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/filters?view=aspnetcore-3.1#result-filters)),

I've used the asynchronous version here, following the pattern used in `MinifyJavaScriptResultAttribute` - but actually given this one doesn't need to call any asynchronous code like that one does, it's clearer just to use the synchronous pattern.  So have updated to that.

elit0451:
Hi @AndyButland, I am closing this PR as it duplicates part of https://github.com/umbraco/Umbraco-CMS/pull/7992

PR #7992, opened 2020-04-21
Netcore: Reimplementing action filters

elit0451:
## Action filters implemented in .net core :
- `StatusCodeResultAttribute`
- `DisableBrowserCacheAttribute`
-  `PreRenderViewActionFilterAttribute`
- `EnsurePartialViewMacroViewContextFilterAttribute`

## Testing
- You need to use one of the existing controllers in .net core solution (ex: BackOfficeController)
- To test the individual attributes place the [_action filter_] before an action, ex:
  - `[StatusCodeResult(System.Net.HttpStatusCode.Accepted)]`, the implementation in the attribute will be executed **after** the action is executed
  - `[DisableBrowserCache]`, the implementation in the attribute will be executed **before** the action is called
   -> To check the result, open the network tab and check the request of the action that you called
   -> Have a look at the Response Headers, you need to have the following:
```
cache-control: no-store, must-revalidate, no-cache, max-age=0
...
expires: Mon, 01 Jan 1990 00:00:00 GMT
last-modified: Tue, 21 Apr 2020 10:39:28 GMT (different value but similar format)
pragma: no-cache
```
  - `[PreRenderViewActionFilter]`, the implementation in the attribute will be executed **after** the action is executed
  - `[EnsurePartialViewMacroViewContextFilter]`, for this attribute, you need to change the base class of the controller you are using; `RenderController` needs to be the base class

### DIFF
--- a/src/Umbraco.Web.BackOffice/ActionExecutedEventArgs.cs
+++ b/src/Umbraco.Web.BackOffice/ActionExecutedEventArgs.cs
@@ -1,0 +1,17 @@
+﻿using System;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Umbraco.Web.BackOffice
+{
+    public class ActionExecutedEventArgs : EventArgs
+    {
+        public Controller Controller { get; set; }
+        public object Model { get; set; }
+
+        public ActionExecutedEventArgs(Controller controller, object model)
+        {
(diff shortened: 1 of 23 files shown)


bergmania:
Lovely work 💪

Issue #5151, opened 2019-04-04
Nasty Exception with Scope/Provider


zpqrtbnk:
As reported in #5121 some nasty exceptions were found in logs:

~~~~
System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at Umbraco.Core.Scoping.ScopeProvider.SetCallContextObject(String key, IInstanceIdentifiable value)
   at Umbraco.Core.Scoping.ScopeProvider.SetAmbient(Scope scope, ScopeContext context)
   at Umbraco.Core.Scoping.ScopeProvider.CreateScope(IsolationLevel isolationLevel, RepositoryCacheMode repositoryCacheMode, IEventDispatcher eventDispatcher, Nullable`1 scopeFileSystems, Boolean callContext, Boolean autoComplete)
   at Umbraco.Core.Services.Implement.UserService.GetUserById(Int32 id)
   at Umbraco.Core.Services.Implement.UserService.GetProfileById(Int32 id)
   at Umbraco.Core.ContentExtensions.GetCreatorProfile(IContentBase content, IUserService userService)
   at Umbraco.Examine.ContentValueSetBuilder.<GetValueSets>d__3.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at Examine.LuceneEngine.Providers.LuceneIndex.ForceProcessQueueItems(Boolean block) in C:\projects\examine\src\Examine\LuceneEngine\Providers\LuceneIndex.cs:line 879
~~~~

~~~~
System.NullReferenceException: Object reference not set to an instance of an object.
   at Umbraco.Core.Scoping.Scope.<>c__DisplayClass72_0.<RobustExit>b__2()
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions)
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions)
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions)
   at Umbraco.Core.Scoping.Scope.RobustExit(Boolean completed, Boolean onException)
   at Umbraco.Core.Scoping.Scope.DisposeLastScope()
   at Umbraco.Core.Scoping.Scope.Dispose()
[...]

PR #7994, opened 2020-04-22
Fixing Nasty Exception with Scope/Provider #5151 - when indexing data

Shazwazza:
Fixing Nasty Exception with Scope/Provider #5151

To replicate this problem it seems that you would need to individually re-index a lot of documents in one operation (i.e. like transferring a lot with Deploy). 

This is actually only a guess at the problem/solution but I think it's correct. The underlying exception is: 

```
System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at Umbraco.Core.Scoping.ScopeProvider.SetCallContextObject(String key, IInstanceIdentifiable value)
   at Umbraco.Core.Scoping.ScopeProvider.SetAmbient(Scope scope, ScopeContext context)
   at Umbraco.Core.Scoping.ScopeProvider.CreateScope(IsolationLevel isolationLevel, RepositoryCacheMode repositoryCacheMode, IEventDispatcher eventDispatcher, Nullable`1 scopeFileSystems, Boolean callContext, Boolean autoComplete)
   at Umbraco.Core.Services.Implement.UserService.GetUserById(Int32 id)
   at Umbraco.Core.Services.Implement.UserService.GetProfileById(Int32 id)
   at Umbraco.Core.ContentExtensions.GetCreatorProfile(IContentBase content, IUserService userService)
   at Umbraco.Examine.ContentValueSetBuilder.<GetValueSets>d__3.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at Examine.LuceneEngine.Providers.LuceneIndex.ForceProcessQueueItems(Boolean block) in C:\projects\examine\src\Examine\LuceneEngine\Providers\LuceneIndex.cs:line 879
```

and 

```
   at Umbraco.Core.Scoping.Scope.<>c__DisplayClass72_0.<RobustExit>b__2()
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions)
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions)
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions)
   at Umbraco.Core.Scoping.Scope.RobustExit(Boolean completed, Boolean onException)
   at Umbraco.Core.Scoping.Scope.DisposeLastScope()
   at Umbraco.Core.Scoping.Scope.Dispose()
   at Umbraco.Core.Services.Implement.UserService.GetUserById(Int32 id)
   at Umbraco.Core.Services.Implement.UserService.GetProfileById(Int32 id)
   at Umbraco.Core.ContentExtensions.GetCreatorProfile(IContentBase content, IUserService userService)
   at Umbraco.Examine.ContentValueSetBuilder.<GetValueSets>d__3.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at Examine.LuceneEngine.Providers.LuceneIndex.ForceProcessQueueItems(Boolean block) in C:\projects\examine\src\Examine\LuceneEngine\Providers\LuceneIndex.cs:line 879
```

But i think the null ref is caused by the first one probably on a second pass of indexing another document.

I 'think' what is happening here is that when a user object hasn't been cached, then a call during a document indexing operation will be made to the IUserService to get the user profile information for that user which will cause a DB query and then it's cached. But when `ForceProcessQueueItems` is called, this is occurring on a background thread and we are not explicitly creating a Scope which means it will try to use an ambient scope but since this isn't in a request this will use a CallContext ambient scope and I think because one doesn't exist on each call during the indexing operation it will try to create one by calling `StaticCallContextObjects` which fwill use the same, which is always the static key of "Umbraco.Core.Scoping.ScopeContext". This is because it's assuming that only one thread would access this same CallContext at once so when the call to `SetAmbient` it first nullifies the CallContext Scope and then adds it. I suspect however that there are multiple threads using this same CallContext at once and when the user isn't cached it needs to make a DB call so it needs to create an Ambient Scope and this happens concurrently, then the null exception will occur because another thread has set the Ambient Scope to null while trying to create it meanwhile another thread is trying to complete the scope.

To avoid this problem, the rule of thumb is that an explicit Scope needs to be used when accessing the database, via services, or whatever when running on a background (non-request) thread else it will try to create an Ambient Scope in the CallContext which in some cases will work fine but in cases like this where somehow multiple threads are accessing the same CallContext it will not work.

... at least that is my theory here.

### DIFF
--- a/src/Umbraco.Examine/ContentValueSetBuilder.cs
+++ b/src/Umbraco.Examine/ContentValueSetBuilder.cs
@@ -53,9 +53,9 @@ namespace Umbraco.Examine
             // processing below instead of one by one.
             using (var scope = _scopeProvider.CreateScope())
             {
-                creatorIds = content.Select(x => x.CreatorId).Distinct().Select(x => _userService.GetProfileById(x))
+                creatorIds = _userService.GetProfilesById(content.Select(x => x.CreatorId).ToArray())
                     .ToDictionary(x => x.Id, x => x);
-                writerIds = content.Select(x => x.WriterId).Distinct().Select(x => _userService.GetProfileById(x))
+                writerIds = _userService.GetProfilesById(content.Select(x => x.WriterId).ToArray())
                     .ToDictionary(x => x.Id, x => x);
                 scope.Complete();
             }
--- a/src/Umbraco.Examine/ContentValueSetBuilder.cs
+++ b/src/Umbraco.Examine/ContentValueSetBuilder.cs
@@ -1,9 +1,13 @@
 ﻿using Examine;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Umbraco.Core;
+using Umbraco.Core.Composing;
 using Umbraco.Core.Models;
+using Umbraco.Core.Models.Membership;
 using Umbraco.Core.PropertyEditors;
+using Umbraco.Core.Scoping;
 using Umbraco.Core.Services;
 using Umbraco.Core.Strings;
 
@@ -16,20 +20,46 @@ namespace Umbraco.Examine
     {
         private readonly UrlSegmentProviderCollection _urlSegmentProviders;
         private readonly IUserService _userService;
+        private readonly IScopeProvider _scopeProvider;
+
+        [Obsolete("Use the other ctor instead")]
+        public ContentValueSetBuilder(PropertyEditorCollection propertyEditors,
+            UrlSegmentProviderCollection urlSegmentProviders,
+            IUserService userService,
+            bool publishedValuesOnly)
+            : this(propertyEditors, urlSegmentProviders, userService, Current.ScopeProvider, publishedValuesOnly)
+        {
+        }
 
         public ContentValueSetBuilder(PropertyEditorCollection propertyEditors,
             UrlSegmentProviderCollection urlSegmentProviders,
             IUserService userService,
+            IScopeProvider scopeProvider,
             bool publishedValuesOnly)
             : base(propertyEditors, publishedValuesOnly)
         {
             _urlSegmentProviders = urlSegmentProviders;
             _userService = userService;
+            _scopeProvider = scopeProvider;
         }
 
         /// <inheritdoc />
         public override IEnumerable<ValueSet> GetValueSets(params IContent[] content)
         {
+            Dictionary<int, IProfile> creatorIds;
+            Dictionary<int, IProfile> writerIds;
+
+            // We can lookup all of the creator/writer names at once which can save some
+            // processing below instead of one by one.
+            using (var scope = _scopeProvider.CreateScope())
+            {
+                creatorIds = content.Select(x => x.CreatorId).Distinct().Select(x => _userService.GetProfileById(x))
+                    .ToDictionary(x => x.Id, x => x);
+                writerIds = content.Select(x => x.WriterId).Distinct().Select(x => _userService.GetProfileById(x))
+                    .ToDictionary(x => x.Id, x => x);
+                scope.Complete();
+            }
+
             // TODO: There is a lot of boxing going on here and ultimately all values will be boxed by Lucene anyways
             // but I wonder if there's a way to reduce the boxing that we have to do or if it will matter in the end since
             // Lucene will do it no matter what? One idea was to create a `FieldValue` struct which would contain `object`, `object[]`, `ValueType` and `ValueType[]`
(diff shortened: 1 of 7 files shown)


src/Umbraco.Examine/ContentValueSetBuilder.cs
```diff
@@ -58,8 +88,8 @@ public override IEnumerable<ValueSet> GetValueSets(params IContent[] content)
                     {"urlName", urlValue?.Yield() ?? Enumerable.Empty<string>()},                  //Always add invariant urlName
                     {"path", c.Path?.Yield() ?? Enumerable.Empty<string>()},
                     {"nodeType", c.ContentType.Id.ToString().Yield() ?? Enumerable.Empty<string>()},
-                    {"creatorName", (c.GetCreatorProfile(_userService)?.Name ?? "??").Yield() },
-                    {"writerName",(c.GetWriterProfile(_userService)?.Name ?? "??").Yield() },
+                    {"creatorName", (creatorIds.TryGetValue(c.CreatorId, out var creatorProfile) ? creatorProfile.Name : "??").Yield() },
+                    {"writerName", (writerIds.TryGetValue(c.CreatorId, out var writerProfile) ? writerProfile.Name : "??").Yield() },
```
bjarnef, line 92: @Shazwazza shouldn't it be `(writerIds.TryGetValue(c.WriterId, out var writerProfile)` here instead of `c.CreatorId`?

bergmania: I think so @bjarnef

src/Umbraco.Examine/ContentValueSetBuilder.cs
```diff
@@ -16,20 +20,46 @@ public class ContentValueSetBuilder : BaseValueSetBuilder<IContent>, IContentVal
...
         public override IEnumerable<ValueSet> GetValueSets(params IContent[] content)
         {
+            Dictionary<int, IProfile> creatorIds;
+            Dictionary<int, IProfile> writerIds;
+
+            // We can lookup all of the creator/writer names at once which can save some
+            // processing below instead of one by one.
+            using (var scope = _scopeProvider.CreateScope())
+            {
+                creatorIds = content.Select(x => x.CreatorId).Distinct().Select(x => _userService.GetProfileById(x))
```
bergmania, line 56: I suggest we create an extension method like `IEnumerable<IProfile> GetProfilesById(this IUserService userService, params int[] ids)`

I'm not sure how this can save processing , when it still look up one by one

sitereactor: Makes sense to me 👍

sitereactor: I assume it handles batches if too many ids are passed in?

bergmania: @sitereactor, the underlying `GetUsersById` handles if there are 2000+ ids, by loading all items into the cache.

sitereactor: Yeah, that was a 'tanketorsk' so I deleted my comment. All good 😉 👍

bergmania: @sitereactor  👍 

Will you review the changes:
https://github.com/umbraco/Umbraco-CMS/pull/7994/commits/27f7d5efae744e9e4d227e3d526ec83113ded80f

And merge if happy

Shazwazza: @bergmania 

> I'm not sure how the current implementation can save processing , when it still look up one by one.

It's because this is doing a `Distinct` so "in theory" it could reduce processing. All users are cached. I don't think `GetUsersById` goes into cache since the repository cache policy is just the default so this will cause even more queries to execute

Shazwazza: All users are cached only when they are resolved by single ID based on the UserRepository cache policy

deekoulos:
Hi @Shazwazza,

I am using version 8.7.0 and getting exactly the same exception while running background tasks doing save and publish of multiple nodes. 

According to your version assignments, this bug should have been fixed in 8.6.4 and I guess also in 8.7.0, right? Why am I then still getting this error?

Thanks
Dee

ksingercoffey:
We are seeing this same stack trace on our load balanced sites. It only seems to be thrown on the replica servers after the distributed cache tells them to re-index updated content.

The area of Scope.cs that is throwing the error is:

starting around line 436
```
                // if *we* created it, then get rid of it
                if (_scopeProvider.AmbientContext == Context)
                {
                    try
                    {
                        _scopeProvider.AmbientContext.ScopeExit(completed);
                    }
                    finally
                    {
                        // removes the ambient context (ambient scope already gone)
                        _scopeProvider.SetAmbient(null);
                    }
                }
```

When debugging, Context is null and _scopeProvider.AmbientContext is null so it continues and then the .ScopeExit() line fails with the null reference.

It's been pretty tough to replicate, though. It doesn't always become null.

deekoulos:
after upgrading to 8.7.1 the issue is resolved for me!

sbosell:
I am seeing this on 8.9.1 with a recurring task running (no load balancing) on azure web app:

image

PR #7996, opened 2020-04-22
Fixes up all things logging in netcore

Shazwazza:
There's quite a lot of (good) changes here. Now logging actually works in the netcore web project. If you start it up you'll now get logs output to the content root (not the www root) at /App_Data/Logs 🎉 

The notable changes:

* New ILoggingConfiguration which is used to specify in ONE place the paths for logs and log config files. There's a TODO and a remaining question we'll need to answer eventually ... should we keep the current xml logging files? This now means we aren't ever hard coding the App_Data/Logs path everywhere, it's just in one place.
* Fixes the paths for logging, no more using IIOHelper for this since that will always map to the www path with MapPath.  Removes all instances of `AppDomain.CurrentDomain.BaseDirectory` since that would resolve differently between net framework and netcore. This also meant cleaning up the log viewer stuff. 
* Moves the MiniDump logic into a separate class so it's not coupled to our logger implementation which means we don't need so many dependencies used to create our logger - it needs to be super slim. This logic now exists as part of a serilog Enricher.
* Dynamically adds enrichers to Serilog after the container is built, we can do this because we are using serilog's `Enrich.FromLogContext()`, now we don't have to pre-resolve things from the container before its ready 
* splits IProfiler into 2 bits with a new IProfilerHtml which is used only to render the Html this means we don't have to pre-resolve httpcontextaccsesor before the container is built (pretty sure this would cause some weird problems because it would mean there were 2x httpcontextaccessors in our app)
* Adds our serilogger config files to netcore to be read from the /Config root


src/Umbraco.Web.Common/Middleware/UmbracoRequestLoggingMiddleware.cs
```diff
@@ -0,0 +1,41 @@
...
+            _next = next;
+            _sessionIdEnricher = sessionIdEnricher;
+            _requestNumberEnricher = requestNumberEnricher;
+            _requestIdEnricher = requestIdEnricher;
+        }
+
+        public async Task Invoke(HttpContext httpContext)
+        {
+            // TODO: Need to decide if we want this stuff still, there's new request logging in serilog:
+            // https://github.com/serilog/serilog-aspnetcore#request-logging which i think would suffice and replace all of this?
```
bergmania, line 31: I guess it is better to use their implementation, instead of maintaining our own.

PR #8002, opened 2020-04-22
Fix for tiny MCE when language is en-US

RachBreeze:
### Prerequisites
When opening a document type with the tinyMCE on it users are warned about a missing en-US tinyMCE file.  This PR resolves that


src/Umbraco.Web.UI.Client/lib/tinymce/langs/en_US.js
```diff
@@ -0,0 +1,261 @@
+tinymce.addI18n('en_CA',{
```
bjarnef, line 1: @RachBreeze I think this should be `en_US`.
We already have `en_CA` language here: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/lib/tinymce/langs/en_CA.js

RachBreeze: Hi @bjarnef 

Oof yes :-) good spot thank you

emmaburstow:
Hola Rachel,

Thanks for this work. Lovely to see you and Bjarne collaborating too. We'll shout if we need you,

Emma

src/Umbraco.Web.UI.Client/lib/tinymce/langs/en_US.js
bjarnef: @RachBreeze did you gad any time to update the PR? 😎👍

RachBreeze:
Many thanks @bjarnef  for fixing

nul800sebastiaan:
Thanks @RachBreeze! 👍

PR #8008, opened 2020-04-24
Netcore: #7898 Migrate Misc tests to new test project

electricsheep:
### Prerequisites

This fixes part of https://github.com/umbraco/Umbraco-CMS/issues/7898

### Description

Migrating unit tests in the Misc folder to the new test project.

Started off with a relatively easy bunch in the Misc folder:
- `DateTimeExtensionsTests`, `HashCodeCombinerTests` & `HashGeneratorTests` - just moved the classes
- `XmlHelperTests` - moved and created `XmlDocumentBuilder` (is this ok as a builder?). There are tests marked as Benchmark here, do they need to move somewhere else?
- `UriUtilityTests` - moved and refactored to remove dependency on `TestHelper` and SettingsForTests`. Didn't create a builder as needed to mock and setup a dependency for the class.


AndyButland:
Looks good to me @electricsheep.  I wonder on the `XmlHelper` if it's worth a look as to whether or if this is still used, just in the sense that that "big XML cache" in Umbraco is no more?  So it might be one we could simply get rid of unless there's some other current use for it.  But I wouldn't do that unless someone from HQ agrees.

bergmania:
Looks good @electricsheep. Thanks a lot 🚀 

@AndyButland, `XmlHelper` is still used in `MacroTagParser` and `GridPropertyIndexValueFactory`

PR #8009, opened 2020-04-24
Netcore: Install steps in dotnet core executable

bergmania:
Fixes [AB#6233](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6233) 

### Details
-  Needed to update some dictionarys in config to be case insensitive
-  Small changes to the typefinder
     - Needed to explicit add "System.Data.SqlserverCe" assembly to the list of `DefaultUmbracoAssemblyProvider.UmbracoCoreAssemblyNames` to ensure the assembly is found 
     - Using ` Assembly.LoadFrom(dll)` instead of `Assembly.Load(assemblyName)`
    - Changed `KnownAssemblyExclusionFilter` because the "System." is too generelt now that we need to find `System.Data.SqlserverCe`.
- Created a new `IDbProviderFactoryCreator ` where we inject all possible creators but find the one based on the provider name.. Input to this is very welcome.
    - Added `ProviderName` property on
         - `IBulkSqlInsertProvider`
         - `ISqlSyntaxProvider`
- Removed the version from Config and the `UmbracoVersion.LocalVersion`
- Implemented our own `HttpResponseException ` and a filter to catch these exceptions and return the expected status code.
- Turned off the `WithAspNetCoreSettings` on the container, such that we can register multiple implementations on a single interface, and then inject these as `IEnumerable<...>`
- Macros
    - Due to dependencies we needed the `MacroRenderer` to be ported. I did that, by using a ViewComponent instead of controller with child only action.
- Had to add `AppDomain.CurrentDomain.SetData("DataDirectory", IOHelper.MapPath(Constants.SystemDirectories.Data));` in CoreRuntime, to ensure the `DataDirectory` part of the embedded connectionstrings are correct. 

### Known issues
- I introduced a hack to the BTree.  There is comment in code.
- We still need to implement `MemberUserKeyProvider` for .net core
- We still need to determine the server address. Right now the first request will fail because we try to start the background tasks
- Restarts only works when using IIS/IISExpress not Kestrel directly, due to the fact that we close the app and use the host to start a new on next request.
- It will not be possible to install Umbraco on the old .NET Framework executable

### Testing
- Start the dotnet core executable using IISExpress
- Go to `/Install`
- Follow the steps
- Wait for install and restart
- Verify redirected to `/umbraco`
- Verify connection string is updated in `appsettings.json`
- Verify when restarting, the runtime state is `Run`
- Verify `/install` redirects to `/umbraco`


src/Umbraco.Web.Common/Macros/PartialViewMacroViewComponent.cs
```diff
@@ -0,0 +1,38 @@
...
+using Umbraco.Core.Composing;
+using Umbraco.Core.Models.PublishedContent;
+using Umbraco.Web.Mvc;
+
+namespace Umbraco.Web.Macros
+{
+    /// <summary>
+    /// Controller to render macro content for Partial View Macros
+    /// </summary>
+    //[MergeParentContextViewData] // TODO is this requeired now it is a ViewComponent?
```
Shazwazza, line 14: This depends on my prev question and whether view contexts are nested and wired up, i have no idea how view components now work. I'm hoping a lot has changed but who knows. This is getting into that 'mvc magic' part of our RFC and we're going to run into a lot of this now since we do a lot of old 'mvc magic'. This all has to do with wiring up validation and ModelState during post backs. That in itself with SurfaceControllers and front-end routes is going to be challenging. It took me quite some time to get that all working with MVC. I worked with MS on making that work with dnx (< netcore 1) and it was weird. Now with endpoint routing it will be all changed again. I really want to work on that part. We won't really know abuot this stuff until we dive into it

src/Umbraco.Web.Common/UmbracoContext/UmbracoContextFactory.cs
```diff
@@ -0,0 +1,109 @@
...
+using Umbraco.Web.Security;
+
+namespace Umbraco.Web
+{
+    /// <summary>
+    /// Creates and manages <see cref="IUmbracoContext"/> instances.
+    /// </summary>
+    public class UmbracoContextFactory : IUmbracoContextFactory
+    {
+        private static readonly NullWriter NullWriterInstance = new NullWriter();
```
Shazwazza, line 19: doesn't seem like this is used?

src/Umbraco.Web.UI.Client/src/installer/installer.service.js
```diff
@@ -237,7 +237,7 @@ angular.module("umbraco.install").factory('installerService', function ($rootSco
 			service.status.feedback = getDescriptionForStepAtIndex(service.status.steps, 0);
 			service.status.progress = 0;
 
-			function processInstallStep() {
+			function processInstallStep(retry) {
```
Shazwazza, line 240: what is `retry` ?

src/Umbraco.Web.UI.Client/src/installer/installer.service.js
```diff
@@ -329,20 +304,52 @@ angular.module("umbraco.install").factory('installerService', function ($rootSco
...
-			service.status.loading = false;
-
-			if (factTimer) {
-			    clearInterval(factTimer);
-			}
-
-			$timeout(function(){
-				window.location.href = Umbraco.Sys.ServerVariables.umbracoBaseUrl;
-			}, 1500);
+            $http.post(Umbraco.Sys.ServerVariables.installApiBaseUrl + "CompleteInstall", _installerModel)
```
Shazwazza, line 307: How come this is required? by the time the installer gets to this `complete` stage the site will have already shut down and then we are redirecting to the website below which will restart it

src/Umbraco.Web.UI.Client/src/installer/steps/user.html
```diff
@@ -1,7 +1,7 @@
 ﻿<div ng-controller="Umbraco.Install.UserController">
-    <h1>Install Umbraco 8</h1>
+    <h1>Install Umbraco 9</h1>
```
Shazwazza, line 2: Now you've made it official :P

src/Umbraco.Core/Composing/DefaultUmbracoAssemblyProvider.cs
```diff
@@ -25,6 +25,8 @@ public class DefaultUmbracoAssemblyProvider : IAssemblyProvider
                 "Umbraco.Web.Common",
                 "Umbraco.Web.BackOffice",
                 "Umbraco.Web.Website",
+                "Umbraco.Web.NetCore",
+                "System.Data.SqlServerCe"
```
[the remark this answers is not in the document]
Shazwazza: See below... i think i figured it out, but have added further commments...

src/Umbraco.PublishedCache.NuCache/DataSource/BTree.cs
```diff
@@ -22,6 +22,13 @@ internal class BTree
                 // default is 4096, min 2^9 = 512, max 2^16 = 64K
                 FileBlockSize = GetBlockSize(settings),
 
+
+                //HACK: Forces FileOptions to be WriteThrough here: https://github.com/mamift/CSharpTest.Net.Collections/blob/9f93733b3af7ee0e2de353e822ff54d908209b0b/src/CSharpTest.Net.Collections/IO/TransactedCompoundFile.cs#L316-L327,
+                // as the reflection uses otherwise will failed in .NET Core as the "_handle" field in FileStream is renamed to "_fileHandle".
+                StoragePerformance = StoragePerformance.CommitToDisk,
```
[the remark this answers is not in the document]
bergmania: I dont have a lot of knowledge about the b-tree package, so I don't know whether `StoragePerformance.CommitToDisk,` is a bad thing, but I will make a PR to the project, about changing the `_handle` variable

src/Umbraco.Core/Composing/ReferenceResolver.cs
```diff
@@ -65,7 +65,7 @@ public IEnumerable<Assembly> ResolveAssemblies()
                     if (assemblyName.FullName.StartsWith("Umbraco."))
                         continue;
 
-                    var assembly = Assembly.Load(assemblyName);
+                    var assembly = Assembly.LoadFrom(dll);
```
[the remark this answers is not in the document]
bergmania: The reasoning was because `Assembly.Load(assemblyName);` could not load the SqlCE assembly

src/Umbraco.Core/Exceptions/HttpResponseException.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Runtime.Serialization;
+
+namespace Umbraco.Core.Exceptions
+{
+    [Serializable]
+    public class HttpResponseException : Exception
```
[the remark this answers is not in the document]
bergmania: There is no netstandard or NetCore version of it

src/Umbraco.Infrastructure/Persistence/IEmbeddedDatabaseCreator.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Umbraco.Core.Persistence
+{
+    public interface IEmbeddedDatabaseCreator
+    {
+        string ProviderName { get; }
+        void Create();
```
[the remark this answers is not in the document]
bergmania: We need to have the functionality implemented in the `Umbraco.Persistance.SqlCe` assembly.

src/Umbraco.Web.Common/AspNetCore/AspNetCoreHostingEnvironment.cs
```diff
@@ -26,6 +23,7 @@ public AspNetCoreHostingEnvironment(IHostingSettings hostingSettings, IWebHostEn
             ApplicationPhysicalPath = webHostEnvironment.ContentRootPath;
 
             ApplicationVirtualPath = "/"; //TODO how to find this, This is a server thing, not application thing.
+            ApplicationServerAddress = "https://localhost:44354"; // TODO how to find this?
```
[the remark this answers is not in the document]
bergmania: It does not need to be part of IHostingEnvironment interface, but IMO it is an okay fit based on the name of the interface?

Do you have ideas to how to get this value, as far as I can see, the only safe way, is as a config value, because reverse proxies will never be handled otherwise.

src/Umbraco.Web.Common/Attributes/AngularJsonOnlyConfigurationAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Text.Json;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Mvc.Formatters;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+using Umbraco.Web.Common.Formatters;
+
+namespace Umbraco.Web.Common.Attributes
```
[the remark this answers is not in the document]
bergmania: I agree, not perfect, but pretty aligned with the MVC pattern, "Views", "Controllers", ...

The issue here is it is not all filters, or related to MVC.

I will give it a try, to find better name or locations of the files

src/Umbraco.Web.UI.Client/src/installer/steps/user.html
bergmania: 🙈 Not more then the assembly version. But actual I guess we should use the assembly version instead of a hardcoded value.

src/Umbraco.Web.UI.Client/src/installer/installer.service.js
bergmania: good catch, some code I removed again

src/Umbraco.Web.UI.Client/src/installer/installer.service.js
bergmania: The site will not restart in the installer (Only if the starterkit steps is applied).

This is a more explicit restart, and I also had to increase the wait time ,because IIEexpress is slower to do the restart stuff. I had an 502 error Bad request throwed by the AspNetCoreModule, outside of our application.

src/Umbraco.Core/Composing/ReferenceResolver.cs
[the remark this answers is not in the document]
Shazwazza: I think we need to revert this, loading LoadFrom vs Load are 2 different things into 2 different assembly load contexts. The Assembly.Load uses fusion binding and we definitely don't want to muck with this when loading plugins. Here's what I propose... this could be done after this PR is merged too, let me know what you think. I would also be more than happy to do this stuff (i.e. I'd like to)

* We revert all of the TypeFinder, etc... changes in this PR that deal with loading in SqlCe
* We manually load in SqlCe in a different way than is done here and we don't need to use TypeFinder for it. SqlCe might need to load with Assembly.LoadFrom because it's not actually compatible but we should not be loading in compatible assemblies that way

src/Umbraco.PublishedCache.NuCache/DataSource/BTree.cs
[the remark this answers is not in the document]
Shazwazza: Sounds good but do note that the orig maintainer of this lib prob won't release anything but he may respond. Stephane had been in touch with him in the past and he was responsive but said he isn't actually 'maintaining' the project. He said it was used in hundreds+ different projects and it's stable, however with the netcore change this might of course be different. Would be interesting to see what they know about this one.

src/Umbraco.Web.Common/AspNetCore/AspNetCoreHostingEnvironment.cs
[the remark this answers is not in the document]
Shazwazza: We already offer this as part of config: `IWebRoutingSettings.UmbracoApplicationUrl`. Perhaps this config could/should move elsewhere. But i guess the real question is why we need this value before a request begins? Currently in v8 we don't. We dynamically set this value based on the initial request if the `IWebRoutingSettings.UmbracoApplicationUrl` isn't specified. (there used to be a 3rd place where this was configured https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Sync/IServerRegistrar.cs#L27 ... but we've thankfully removed that since this should just be consolidated to one place)

src/Umbraco.Web.Common/Attributes/AngularJsonOnlyConfigurationAttribute.cs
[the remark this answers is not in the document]
Shazwazza: Cool, though some argue against the MVC pattern of "Views", "Controllers", etc... .too :) which is why there's feature folders https://codingsight.com/asp-net-core-mvc-feature-folders/ 

I'd also prefer we don't just use "Controllers" and instead break that up into it's features. Could be 'BackOffice', or like we do currently for back office editors, those all exist in `Editors`.

I guess so long as we have a few tasks later on about re-adjusting all namespaces we don't have to 'solve' this today but I just worry about leaving it until later because based on past predictions these cleanup tasks won't actually get done.

src/Umbraco.Infrastructure/Persistence/IEmbeddedDatabaseCreator.cs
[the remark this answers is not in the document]
Shazwazza: I guess I don't quite understand this one, maybe we can chat about it. How come IDbProviderFactoryCreator can't be implemented from the Umbraco.Persistance.SqlCe assembly?

src/Umbraco.Web.UI.Client/src/installer/installer.service.js
Shazwazza: Cool all good then

src/Umbraco.Web.UI.NetCore/Umbraco.Web.UI.NetCore.csproj
```diff
@@ -4,35 +4,52 @@
...
+        <Content Remove="wwwroot\Umbraco\views\install\error.html" />
+        <Content Update="Views\Install\Index.cshtml">
+          <ExcludeFromSingleFile>true</ExcludeFromSingleFile>
+          <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+        </Content>
+    </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.3" />
+      <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.7.0" />
```
Shazwazza, line 52: why are we referencing `System.Configuration.ConfigurationManager` ?

src/Umbraco.Web.UI.NetCore/Umbraco.Web.UI.NetCore.csproj
```diff
@@ -4,35 +4,52 @@
         <TargetFramework>netcoreapp3.1</TargetFramework>
         <RootNamespace>Umbraco.Web.UI.NetCore</RootNamespace>
     </PropertyGroup>
-
     <ItemGroup>
+      <ProjectReference Include="..\Umbraco.Examine.Lucene\Umbraco.Examine.Lucene.csproj" />
```
Shazwazza, line 8: We need a task to change this to work properly since this in theory shouldn't actually be working :) We won't be able to ship with Lucene 3.0.3, we'll need to update Examine to Lucene 4.8.0 (netstandard) which is currently in beta00008 (though it is considered stable)

src/Umbraco.Web/AspNet/AspNetHostingEnvironment.cs
```diff
@@ -25,20 +26,25 @@ public AspNetHostingEnvironment(IHostingSettings hostingSettings)
                                      ?? HostingEnvironment.ApplicationVirtualPath?.EnsureStartsWith("/")
                                      ?? "/";
             IISVersion = HttpRuntime.IISVersion;
+
+            ApplicationServerAddress = "https://localhost:44354"; // TODO how to find this?
```
Shazwazza, line 30: See my prev comment about this

src/Umbraco.Web/Install/UmbracoInstallArea.cs
```diff
@@ -1,47 +1,40 @@
...
-                new[] { typeof(InstallApiController).Namespace });
-        }
-
-        public override string AreaName
-        {
-            get { return "UmbracoInstall"; }
-        }
-    }
-}
+﻿//
```
Shazwazza, line 1: Are we keeping this?

src/Umbraco.Web/Runtime/WebInitialComponent.cs
```diff
@@ -131,7 +131,7 @@ private void SetupMvcAndWebApi()
             RouteNoContentController(umbracoPath);
 
             // register install routes
-            RouteTable.Routes.RegisterArea<UmbracoInstallArea>();
+           // RouteTable.Routes.RegisterArea<UmbracoInstallArea>();
```
Shazwazza, line 134: I assume we're just commenting this out because with these changes installation wont run in netframework?

src/Umbraco.Web.UI.Client/src/common/interceptors/culturerequest.interceptor.js
```diff
@@ -10,7 +10,7 @@
             //dealing with requests:
             'request': function (config) {
 
-                if (!Umbraco.Sys.ServerVariables.umbracoSettings.umbracoPath) {
+                if (!Umbraco.Sys.ServerVariablesUmbraco.Sys.ServerVariables.umbracoSettings.umbracoPath) {
```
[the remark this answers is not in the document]
bergmania: 🙈  Thanks, my bad

src/Umbraco.Web.UI.NetCore/Umbraco.Web.UI.NetCore.csproj
bergmania: Currently a hack due to Lucene, can be removed when Examine is updated to netstandard

src/Umbraco.Web.UI.NetCore/Umbraco.Web.UI.NetCore.csproj
bergmania: Isn't this the task? you Created it yourself?
https://dev.azure.com/umbraco/D-Team%20Tracker/_workitems/edit/4911

Shazwazza: haha, well i guess so :)

src/Umbraco.Web/Runtime/WebInitialComponent.cs
bergmania: Yes, the current implementation cannot run in netframework. There controllers and everything is moved.

src/Umbraco.PublishedCache.NuCache/DataSource/BTree.cs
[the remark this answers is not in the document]
bergmania: Right now we use a fork of the original project. That fork is hopefully more active

src/Umbraco.Web.Common/Macros/MacroRenderer.cs
```diff
@@ -0,0 +1,423 @@
...
+            }
+
+            if (model.CacheByPage)
+                id.AppendFormat("{0}-", pageId);
+
+            if (model.CacheByMember)
+            {
+                object key = 0;
+
+                if (_umbracoContextAccessor.UmbracoContext.Security.IsAuthenticated())
```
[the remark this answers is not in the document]
bergmania: GetRequiredUmbracoContext

src/Umbraco.Core/Composing/ReferenceResolver.cs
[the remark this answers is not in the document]
bergmania: Now SqlCE is manual loaded, and we just ignore if it is not avaiable. The implementation works if there is no project reference to the `Umbraco.Persistence.SqlCe` project. 

In this case, we also "remove" the fast install option, so you need to go to the next step, and type in db info. On the other hand, if SqlCE is available, the fast install is still possible.
SqlCe is available if all the required dlls are in the bin folder. (e.g. by making a reference to the  `Umbraco.Persistence.SqlCe`  project)

src/Umbraco.Web.Common/AspNetCore/AspNetCoreHostingEnvironment.cs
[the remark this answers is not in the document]
bergmania: In v8, we called `EnsureApplicationUrl` on each request, that set the application url, if it was changed. 

This was pretty nasty, as we casted the IRuntimeState to RuntimeState just to call that method 😭 

Now I implemented a `GetApplicationUrl()` on `IRequestAccessor` that does this whenever the application url is used. In general this is `< 1` time per request.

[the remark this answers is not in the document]
bergmania: The original code in `EnsureApplicationUrl` had a fixme comment. I currently keep this comment, but i'm unsure if this is still relevant

src/Umbraco.Web.Common/Install/InstallApiController.cs
```diff
@@ -69,10 +79,18 @@ public IEnumerable<Package> GetPackages()
...
+            _umbracoApplicationLifetime.Restart();
+            return NoContent();
+        }
+
         /// <summary>
-        /// Installs.
+        ///     Installs.
         /// </summary>
-        public async Task<InstallProgressResultModel> PostPerformInstall(InstallInstructions installModel)
+        public async Task<InstallProgressResultModel> PostPerformInstall([FromBody] InstallInstructions installModel)
```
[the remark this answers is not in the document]
bergmania: Dont seems to be enough to add `[HttpPost]`. In that case I still need `[FromBody]`

src/Umbraco.Web.Common/Macros/MemberUserKeyProvider.cs
```diff
@@ -0,0 +1,14 @@
+using Umbraco.Core.Security;
+
+namespace Umbraco.Web.Common.Macros
+{
+    internal class MemberUserKeyProvider : IMemberUserKeyProvider
+    {
+        public object GetMemberProviderUserKey()
```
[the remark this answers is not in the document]
bergmania: Will be part of a revisit macros task

src/Umbraco.Web.Common/Macros/MacroRenderer.cs
```diff
@@ -0,0 +1,423 @@
...
+        }
+
+        // gets the macro source file
+        // null if macro is not file-based
+        private FileInfo GetMacroFile(MacroModel model)
+        {
+            var filename = GetMacroFileName(model);
+            if (filename == null) return null;
+
+            var mapped = _ioHelper.MapPath(filename);
```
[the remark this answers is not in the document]
bergmania: Will be seperate task

src/Umbraco.Web.Common/AspNetCore/AspNetCoreHostingEnvironment.cs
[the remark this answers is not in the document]
Shazwazza: Yep, the problem is when we dynamically assign this URL which has issues with things like slot swapping and a bunch of other things. I think the reality is that for people hosting on Azure or doing things where the URL can actually change hosts, etc... they will need to configure the URL. I'll have a look at the changes.

src/Umbraco.Infrastructure/Install/InstallSteps/DatabaseConfigureStep.cs
```diff
@@ -64,8 +65,37 @@ private void ConfigureConnection(DatabaseModel database)
...
+                return new
+                {
+                    databases = databases
+                };
+            }
+        }
+
+        public static bool IsSqlCeAvailable()
+        {
+            return !(Type.GetType("Umbraco.Persistance.SqlCe.SqlCeSyntaxProvider, Umbraco.Persistance.SqlCe") is null);
```
Shazwazza, line 93: This may not always work unless the type is loaded into the appdomain already (perhaps it is, ... i'll continue checking in the PR). The alternative way to force it Assembly.Load("....").GetType("....") then you are forcibly loading it into the appdomain.

src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
```diff
@@ -24,14 +25,65 @@
...
+            try
+            {
+                var binFolder = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+                if (binFolder != null)
+                {
+                    var dllPath = Path.Combine(binFolder, "Umbraco.Persistance.SqlCe.dll");
+                    Assembly.LoadFrom(dllPath);
+
+
+                    var sqlCeSyntaxProviderType = Type.GetType("Umbraco.Persistance.SqlCe.SqlCeSyntaxProvider, Umbraco.Persistance.SqlCe");
```
Shazwazza, line 48: We should also use Assembly.Load("...").GetType("...") if this ever returns null which is possible.

src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
```diff
@@ -122,8 +174,17 @@ public static IServiceCollection AddUmbracoCore(this IServiceCollection services
             if (container is null) throw new ArgumentNullException(nameof(container));
             if (entryAssembly is null) throw new ArgumentNullException(nameof(entryAssembly));
 
+            services.AddSingleton<IDbProviderFactoryCreator>(x => new DbProviderFactoryCreator(
```
Shazwazza, line 177: This is another case where we are resolving things from a pre-resolved container. This means that this instance of IDbProviderFactoryCreator and all of it's dependencies at this stage will not be the same instances as ones resolved after the real container is build so they aren't actually "singletons." There is an ugly work around for this that I've created which is to re-register all of these types in the container as a delegate that points to the pre-resolved services but of course that is a nasty hack. 

I'm unsure how we can work around this but i'll see what i can find.

Shazwazza: It seems that we are doing this purely to use this object for the SqlMainDomLock BUT we don't want to do that, that is why we manually created these things before, just like we do in v8. We don't support SqlMainDomLock with SqlCe, only SQL server which was configured before. I don't think we need this here.

src/Umbraco.Web.Common/Install/InstallApiController.cs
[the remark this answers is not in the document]
Shazwazza: That is quite strange since it's just a normal posted model 🤔  I'll have a look at why that might be

src/Umbraco.Web.Common/Runtime/AspNetCoreComponent.cs
```diff
@@ -1,18 +1,30 @@
...
 using Umbraco.Net;
+using Umbraco.Web.Common.Lifetime;
 
 namespace Umbraco.Web.Common.Runtime
 {
     public sealed class AspNetCoreComponent : IComponent
     {
         private readonly IHostApplicationLifetime _hostApplicationLifetime;
         private readonly IUmbracoApplicationLifetimeManager _umbracoApplicationLifetimeManager;
+        private readonly IUmbracoRequestLifetime _umbracoRequestLifetime;
```
Shazwazza, line 15: These are not used

src/Umbraco.Web.UI.NetCore/Umbraco.Web.UI.NetCore.csproj
```diff
@@ -4,35 +4,52 @@
...
 
     <ItemGroup>
-      <None Remove="wwwroot\Umbraco\**" />
         <None Remove="App_Data\**" />
     </ItemGroup>
 
     <ItemGroup>
-        <Content Remove="wwwroot\Umbraco\**" />
         <Content Remove="App_Data\**" />
+        <Content Remove="wwwroot\Umbraco\views\install\machinekey.html" />
```
Shazwazza, line 34: We can remove the machinekey install step since that doesn't actually exist in netcore

src/Umbraco.Web/Install/UmbracoInstallArea.cs
```diff
@@ -1,47 +0,0 @@
...
-using System.Linq;
-using System.Web.Http;
-using System.Web.Mvc;
-using System.Web.Routing;
-using Umbraco.Core;
-using Umbraco.Core.Configuration;
-using Umbraco.Web.Editors;
-using Umbraco.Web.Install.Controllers;
-
-namespace Umbraco.Web.Install
```
Shazwazza, line 11: We'll need to undelete this but that will be part of my new tasks.

src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
Shazwazza: In this case Assembly.Load from returns the assembly that was loaded into the appdomain so to be safe its changed to use `assembly.GetType`. I've updated.

src/Umbraco.Infrastructure/Install/InstallSteps/DatabaseConfigureStep.cs
Shazwazza: I've left this one but added notes saying that we know we've forcibly loaded this assembly into the appdomain before this is ever called.

src/Umbraco.Web.UI.NetCore/Startup.cs
```diff
@@ -42,12 +51,27 @@ public Startup(IWebHostEnvironment webHostEnvironment, IConfiguration config)
         // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
         public void ConfigureServices(IServiceCollection services)
         {
+            services.AddUmbracoSqlCeSupport();
```
Shazwazza, line 54: This should be part of our CoreComposer, i'll have a look at fixing this up

Shazwazza: I understand why some of this seems to need to be setup before the rest, mostly to do with the SqlMainDom but that only requires that this is registered `DbProviderFactories.RegisterFactory(Core.Constants.DbProviderNames.SqlServer, SqlClientFactory.Instance);` which could easily be part of it's own construction. It just doesn't seem right that we need to boot things before we boot the core which is also why we need to pre-add things for the integrations tests. Maybe you have other thoughts about this one but i'll cleanup things how i think it could work and maybe you can let me know if you disagree.

Shazwazza:
@bergmania merging now, i've added notes/tasks for the remaining bits

PR #8010, opened 2020-04-26
Theme backoffice docs for v8 to clearly differentiate it from the v7 docs

BatJan:
### Prerequisites

### Description
I have caught myself needing to double check the url for the v8 backoffice documentation to make sure I was not looking at the v7 docs since they currently share the same theme and both headlines are identical saying "Umbraco Backoffice UI API Documentation".

Therefore I have changed the headline to say "Umbraco 8 Backoffice UI API Documentation" and then I have themed it so the styles differentiate from the v7 version as well making it very explicit what version the documentation is for - The colors etc. are reused from our.umbraco.com.

I'm of course open for suggestions for changes 😃 

Oh, and btw - It would be nice if the docs could be updated as well since for instance the umbDateTimePicker is still displayed as umbFlatPicker https://our.umbraco.com/apidocs/v8/ui/#/api/umbraco.directives.directive:umbFlatpickr even though the documentation for it has changed - There might be other directives etc. that would be nice to have updated as well 😃 
**Before**
backoffice-docs-before

**After**
backoffice-docs-after

### DIFF
--- a/src/Umbraco.Web.UI.Docs/gulpfile.js
+++ b/src/Umbraco.Web.UI.Docs/gulpfile.js
@@ -16,7 +16,7 @@ gulp.task('docs', [], function (cb) {
     var options = {
         html5Mode: false,
         startPage: '/api',
-        title: "Umbraco Backoffice UI API Documentation",
+        title: "Umbraco 8 Backoffice UI API Documentation",
         dest: './api',
         styles: ['./umb-docs.css'],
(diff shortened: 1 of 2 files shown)


emmaburstow:
Hey Jan,

A mahoosive #h5yr to you. Thanks for the work. We'll test and let you know if we have questions. Lovely images, by the way. 

Em

nul800sebastiaan:
Thanks @BatJan! This will take a while to roll out, but we'll get there 😅

PR #8013, opened 2020-04-27
Grid media picker doesn't enable select (at all)

lars-erik:
When adding images to the grid using the built-in media grid-editor, the select button is never enabled. When adding `config.size`, the details dialog shows, but no longer in its own dialog, so the submit/close methods aren't used.  
I've moved the selection statement up to enable selection.  
Historically, the crop-mechanism here works rather bad as well, so hope to have a quick look at it, but this can be merged as-is to make the editor work at all.

Dropping a mention to @bjarnef and @kjac since their names are all over the blame for these files. 😉  
Would appreciate feedback/thoughts/hidden knowledge.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js
@@ -347,23 +347,11 @@ angular.module("umbraco")
 
             function openDetailsDialog() {
                 localizationService.localize("defaultdialogs_editSelectedMedia").then(function (data) {
+                    $scope.model.selection.push($scope.target);
                     vm.mediaPickerDetailsOverlay = {
                         show: true,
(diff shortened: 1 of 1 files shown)


lars-erik:
Doh, after a rebase to the v8/dev merge @nul800sebastiaan did at the same time as I pushed the PR, the PR is invalid. :P The details overlay is back as a dialog with its own submit/close buttons and have started working again. However, the crop feature is broken. Gonna have a look at that, which was my next goal in any case. Unless it's being worked at by someone else?

lars-erik:
HA! This was a styling issue all along. Currently hunting purp in 8.6.1 while trying to get #8023 into it as a temp patch with custom view for grid editor.

lars-erik:
I'm still completely stumped what makes the difference between 8.6.x and 8.7 (contrib) except that the overlays are now on top of the main menu as well. Can't see _any_ relevant difference. 😱
Really need to fix this in 8.6.x, but totally clueless.

PR #8014, opened 2020-04-27
Add utility function equivalent to angular.fromJson

bjarnef:
### Prerequisites

### Description
This previous PR https://github.com/umbraco/Umbraco-CMS/pull/7952 replacing `angular.fromJson` from `JSON.parse` seems to fail as `angular.fromJson` also checked in the value passed in was equal to an string value.

E.g. when inserting an image editor in grid content it is causing an console error, when session storage is empty.

Furthermore we now have a `Utilities.toJson()` function https://github.com/umbraco/Umbraco-CMS/pull/7924 so I think we should have an `Utilities.fromJson` as well.

You can of course always use the traditional `JSON.parse` inside a `try/catch` block in e.g, but this method (similar to the `angular.fromJson()` only try to parse the value if it is an string value).

```
try {
    var obj = JSON.parse(value);
}
catch { 

}
```

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbrawmodel.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbrawmodel.directive.js
@@ -39,7 +39,7 @@ angular.module("umbraco.directives")
 
                 function stringToJson(text) {
                     try {
-                        return JSON.parse(text);
+                        return Utilities.fromJson(text);
                     } catch (err) {
                         setInvalid();
                         return text;
@@ -55,7 +55,7 @@ angular.module("umbraco.directives")
                 function isValidJson(model) {
                     var flag = true;
                     try {
-                        JSON.parse(model)
+                        Utilities.fromJson(model)
                     } catch (err) {
                         flag = false;
                     }
--- a/src/Umbraco.Web.UI.Client/src/common/mocks/resources/_utils.js
+++ b/src/Umbraco.Web.UI.Client/src/common/mocks/resources/_utils.js
@@ -90,7 +90,7 @@ angular.module('umbraco.mocks').
                                             name: "1 column layout",
                                             sections: [
                                                 {
-                                                    grid: 12,
+                                                    grid: 12
                                                 }
                                             ]
                                         },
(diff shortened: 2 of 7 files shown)


bjarnef:
**Before**


**After**

bjarnef:
@nul800sebastiaan it seems there have been some other underlying issues regarding when having an image grid editor in grid and saving the content, but we can look at this in another PR.

lars-erik:
I just hit the same issue and was about to fix that as well. Will merged yours to test mine. When cropping, you hit another issue. See #8013. I wanna see about using the cropper from the imagecropper datatype instead as well, maybe in another PR tho.

bjarnef:
The YSOD I get when saving grid content with image grid editor (or just a regular property using media picker) seems to be because I am missing some relations, e.g. `umbMedia`.
https://github.com/umbraco/Umbraco-CMS/commit/193892f0847bec49f4f7891e8dd296c10627a7a1

Maybe the upgrader didn't run as expected after I previous pulled some of the latest changes in `v8/contrib` branch.

bjarnef:
I installed in clean Umbraco install and installed the starter kit. I now have the "Related Media" and "Related Content" relations. However I think these shouldn't have the "delete" actions since it would break stuff if these are deleted. Maybe these should be "locked" similar to default listview datatypes used in content, media and member?

nielslyngsoe:
This addition to Utilities is great! Thanks

bjarnef:
@nielslyngsoe shouldn't this be merged into `v8/contrib` which is this default branch now?

nielslyngsoe:
> @nielslyngsoe shouldn't this be merged into `v8/contrib` which is this default branch now?

Yes and no — Normally I wouldn't be merging these PRs. But we needed it straight away so I just decided to merge it directly to the dev branch. The Dev branch should be merged back to contrib branch each day so hopefully, no one was waiting on this one. Hence no one did merge it from the contrib-team.

bjarnef:
@nielslyngsoe yeah, I just needed this fix when submitting other PRs, but I had the change locally and only comitted the relevant changes for the specific feature/bug. It seems this is merged to `v8/contrib` now, so that's great 👍

Issue #8021, opened 2020-04-27
Lock core relation types


bjarnef:
We now get two new relations types "Related Media" and "Related Content" added, which is used by e.g. media picker property editor.
https://github.com/umbraco/Umbraco-CMS/commit/193892f0847bec49f4f7891e8dd296c10627a7a1

I noticed YSOD when saving property using media picker or save grid content using image grid editor after pulling latest changes, where I didn't had this relations added.
https://github.com/umbraco/Umbraco-CMS/pull/8014

It seems these relations have the delete action (but I think it would break stuff if these are deleted). I think something would happen if deleting datatypes used in core.
Therefore e.g. listviews are "locked".

I think something similar should be applied to these relations.

![image](https://user-images.githubusercontent.com/2919859/80373759-d8619580-8895-11ea-80f0-de8eee022a35.png)

![image](https://user-images.githubusercontent.com/2919859/80373799-e7484800-8895-11ea-95d8-c27867792601.png)

![image](https://user-images.githubusercontent.com/2919859/80373904-165eb980-8896-11ea-858d-2ca9211cba2e.png)

PR #8022, opened 2020-04-27
Lock core relation types

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8021

### Description
This PR lock the core relation types to track content and media, so alias of these relation types is locked and delete action is not available for these. This works similar to the changes in https://github.com/umbraco/Umbraco-CMS/pull/6943

### DIFF
--- a/src/Umbraco.Core/Models/RelationTypeExtensions.cs
+++ b/src/Umbraco.Core/Models/RelationTypeExtensions.cs
@@ -1,0 +1,9 @@
+﻿namespace Umbraco.Core.Models
+{
+    internal static class RelationTypeExtensions
+    {
+        internal static bool IsSystemRelationType(this IRelationType relationType) =>
+            relationType.Alias == Constants.Conventions.RelationTypes.RelatedDocumentAlias
+            || relationType.Alias == Constants.Conventions.RelationTypes.RelatedMediaAlias;
+    }
+}
--- a/src/Umbraco.Core/Models/RelationTypeExtensions.cs
+++ b/src/Umbraco.Core/Models/RelationTypeExtensions.cs
@@ -4,6 +4,9 @@
     {
         internal static bool IsSystemRelationType(this IRelationType relationType) =>
             relationType.Alias == Constants.Conventions.RelationTypes.RelatedDocumentAlias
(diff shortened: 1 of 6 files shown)


emmaburstow:
Hi Bjanre,

As usual, thanks for the work :) We'll take a look and let you know if we need anything.

Em

bjarnef:
@kjac this is pretty much the same functionality as you added in this PR https://github.com/umbraco/Umbraco-CMS/pull/6943

Since these relation types is used for internal use, I don't see the reason having the delete action in the menu for these relation types.

kjac:
@bjarnef couldn't agree more. Maybe I didn't make myself clear: I think the PR is great, and I think you should expand on this PR to ensure that none of the built-in relation types can be deleted 😄 

Here's my reasoning:

As it is right now, Umbraco will fail miserably if either of `Constants.Conventions.RelationTypes.RelatedDocumentAlias` or `Constants.Conventions.RelationTypes.RelatedMediaAlias` is deleted - the ones you have protected with this PR. 

If you delete any of the other built-in relation types, Umbraco will automatically recreate them on demand - i.e. when relating a copy to its original in the content tree. In other words, deleted relation types will eventually reappear again. So I see no point in being able to delete them - thus expanding this PR to include all built-in relation types makes sense to me.

Thoughts?

bjarnef:
@kjac ahh okay, not sure if Umbraco always has recreated those relation types?
What happens og renaming alias of these? Does it re-create these as new relation types with the original alias?
I guess the alias should be locked as well then.

kjac:
@bjarnef not sure if it was always like that... I am assuming as much though 😄 

And yes, makes perfect sense to lock the alias of those types; essentially handle them all in the same way.

bjarnef:
@kjac in https://github.com/umbraco/Umbraco-CMS/pull/8022/commits/2755d29c4e0a118c7d55a1c23fee66806fe82fa3 I have added a few more relation types, which are locked.

image

@Shazwazza do we ever need to delete or rename alias of these relation types, which are added by default?

kjac:
Merged ⭐

Shazwazza:
@bjarnef nope don't think so, thanks!

PR #8025, opened 2020-04-28
Migrated back office to ASP.NET Core Identity

scottbrady91:
### Prerequisites

### Description

Migration from ASP.NET Identity 2 to ASP.NET Core Identity as part of Rock Solid Knowledge's participation in the Unicore project.

**This PR does not include any changes to the angular code or the database.** I have written tests for implementations new to this PR.

#### Changes:
- Removed use of `Microsoft.AspNet.Identity.Owin` within Umbraco
- Removed use of `Microsoft.AspNet.Identity.Core` within Umbraco back office. Full removal requires an update of the members implementation
- Added the following OWIN adaptors:
  - `BackOfficeSignInManager`
  - `SecurityStampValidator`
  - `OwinDataProtectorTokenProvider`
  - `NopLookupNormalizer`
  - `OwinDataProtectorTokenProvider`
- Ported code from the `Microsoft.AspNet.Identity.Owin` library such as `IdentityFactoryMiddleware` and various `IAppBuilder` extensions
- Removed `EmailService`, `UmbracoEmailMessage`, and `BackOfficeSignInManager.SendTwoFactorCodeAsync`. Email and SMS notifications are no longer part of the ASP.NET Identity library and **will need to be replaced**

@bergmania @emmagarland


src/Umbraco.Web/Editors/UsersController.cs
```diff
@@ -709,22 +704,26 @@ public async Task<HttpResponseMessage> PostUnlockUsers([FromUri]int[] userIds)
...
-                var user = await UserManager.FindByIdAsync(userIds[0]);
+
                 return Request.CreateNotificationSuccessResponse(
                     Services.TextService.Localize("speechBubbles/unlockUserSuccess", new[] { user.Name }));
             }
 
             foreach (var u in userIds)
             {
-                var unlockResult = await UserManager.SetLockoutEndDateAsync(u, DateTimeOffset.Now);
+                var unlockResult = await UserManager.SetLockoutEndDateAsync(user, DateTimeOffset.Now);
```
bergmania, line 726: Not sure we can use `user` here.. Think we need to get the user using `u` before.

src/Umbraco.Web/Security/AppBuilderExtensions.cs
```diff
@@ -464,5 +419,41 @@ public static void SanitizeThreadCulture(this IAppBuilder app)
...
+        {
+            return CreatePerOwinContext<T>(app, (options, context) => createCallback());
+        }
+
+        public static IAppBuilder CreatePerOwinContext<T>(this IAppBuilder app,
+            Func<IdentityFactoryOptions<T>, IOwinContext, T> createCallback) where T : class, IDisposable
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException("app");
```
bergmania, line 433:
```suggestion
                throw new ArgumentNullException(nameof(app));
```

src/Umbraco.Web/Security/UmbracoBackOfficeIdentity.cs
```diff
@@ -204,17 +203,17 @@ private UmbracoBackOfficeIdentity(ClaimsIdentity identity)
         private string[] _allowedApplications;
         public string[] AllowedApplications => _allowedApplications ?? (_allowedApplications = FindAll(x => x.Type == Constants.Security.AllowedApplicationsClaimType).Select(app => app.Value).ToArray());
 
-        public int Id => int.Parse(this.FindFirstValue(ClaimTypes.NameIdentifier));
+        public int Id => int.Parse(this.FindFirst(ClaimTypes.NameIdentifier)?.Value);
```
bergmania, line 206: We could consider making an extention method for `FindFirstValue`, to ensure we dont forget the null check etc.

src/Umbraco.Web/Security/AuthenticationManagerExtensions.cs
```diff
@@ -82,5 +84,36 @@ public static async Task<ExternalLoginInfo> GetExternalLoginInfoAsync(this IAuth
             }
             return GetExternalLoginInfo(await manager.AuthenticateAsync(authenticationType));
         }
+        
+        public static IEnumerable<AuthenticationDescription> GetExternalAuthenticationTypes(this IAuthenticationManager manager)
+        {
+            if (manager == null) throw new ArgumentNullException(nameof(manager));
+            return manager.GetAuthenticationTypes(d => d.Properties != null && d.Properties.ContainsKey("Caption"));
```
bergmania, line 91: Question: What is the different between `d.Properties["Caption"]` and [`d.Caption`](https://docs.microsoft.com/en-us/previous-versions/aspnet/mt152177%28v%3dvs.113%29)

src/Umbraco.Web/Security/UmbracoSecurityStampValidator.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            if (getUserIdCallback == null) throw new ArgumentNullException(nameof(getUserIdCallback));
+
+            return async context =>
+            {
+                var currentUtc = DateTimeOffset.UtcNow;
+                if (context.Options != null && context.Options.SystemClock != null)
+                {
+                    currentUtc = context.Options.SystemClock.UtcNow;
+                }
```
bergmania, line 31:
```suggestion
                 var currentUtc = context.Options?.SystemClock?.UtcNow ?? DateTimeOffset.UtcNow;
```

src/Umbraco.Web/Security/OwinDataProtectorTokenProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+            _protector = protector ?? throw new ArgumentNullException(nameof(protector));
+            TokenLifespan = TimeSpan.FromDays(1);
+        }
+
+        public async Task<string> GenerateAsync(string purpose, UserManager<TUser> manager, TUser user)
+        {
+            if (manager == null) throw new ArgumentNullException(nameof(manager));
+            if (user == null) throw new ArgumentNullException(nameof(user));
+
+            var ms = new MemoryStream();
```
bergmania, line 32: Think guess we need to wrap this in a using statement to ensure is is disposed after the `ms.ToArray()`

src/Umbraco.Web/Security/OwinDataProtectorTokenProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+        public async Task<bool> ValidateAsync(string purpose, string token, UserManager<TUser> manager, TUser user)
+        {
+            if (string.IsNullOrWhiteSpace(token)) throw new ArgumentNullException(nameof(token));
+            if (manager == null) throw new ArgumentNullException(nameof(manager));
+            if (user == null) throw new ArgumentNullException(nameof(user));
+
+            try
+            {
+                var unprotectedData = _protector.Unprotect(Convert.FromBase64String(token));
+                var ms = new MemoryStream(unprotectedData);
```
bergmania, line 60: missing using?

src/Umbraco.Web/Security/IdentityFactoryMiddleware.cs
```diff
@@ -0,0 +1,117 @@
...
+        ///     Constructor
+        /// </summary>
+        /// <param name="next">The next middleware in the OWIN pipeline to invoke</param>
+        /// <param name="options">Configuration options for the middleware</param>
+        public IdentityFactoryMiddleware(OwinMiddleware next, TOptions options)
+            : base(next)
+        {
+            if (options == null)
+            {
+                throw new ArgumentNullException("options");
```
bergmania, line 25:
```suggestion
                throw new ArgumentNullException(nameof(options));
```

src/Umbraco.Web/Security/IdentityFactoryMiddleware.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+            if (options.Provider == null)
+            {
+                throw new ArgumentNullException("options.Provider");
```
bergmania, line 29: I guess this is more an `ArgumentException` then it is not the direct argument that is null.

src/Umbraco.Web/Security/BackOfficeUserStore.cs
```diff
@@ -245,25 +286,29 @@ public Task<bool> HasPasswordAsync(BackOfficeIdentityUser user)
...
+        /// <param name="cancellationToken"></param>
         /// <returns/>
-        public Task SetEmailAsync(BackOfficeIdentityUser user, string email)
+        public Task SetEmailAsync(BackOfficeIdentityUser user, string email, CancellationToken cancellationToken = default(CancellationToken))
         {
+            cancellationToken.ThrowIfCancellationRequested();
             ThrowIfDisposed();
             if (user == null) throw new ArgumentNullException(nameof(user));
-            if (email.IsNullOrWhiteSpace()) throw new ArgumentNullException("email");
+            // if (email.IsNullOrWhiteSpace()) throw new ArgumentNullException("email");
```
Shazwazza, line 296: do we need this check? else we can remove the comment

src/Umbraco.Web/Security/BackOfficeUserStore.cs
```diff
@@ -454,36 +528,49 @@ public Task<IList<string>> GetRolesAsync(BackOfficeIdentityUser user)
...
+            cancellationToken.ThrowIfCancellationRequested();
             ThrowIfDisposed();
             if (user == null) throw new ArgumentNullException(nameof(user));
-            return Task.FromResult(user.Roles.Select(x => x.RoleId).InvariantContains(roleName));
+            return Task.FromResult(user.Roles.Select(x => x.RoleId).InvariantContains(normalizedRoleName));
+        }
+
+        public Task<IList<BackOfficeIdentityUser>> GetUsersInRoleAsync(string normalizedRoleName, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
```
Shazwazza, line 545: We could implement this, but can see it's a new thing so can leave for another task. @bergmania  could be an up for grabs task?

src/Umbraco.Web/Security/OwinDataProtectorTokenProvider.cs
scottbrady91: Agreed. I ported this directly from the MS implementation, so this is what you were using before. Let me see if there is a reason that they've left it open so that I don't somehow create a regression.

Shazwazza: I wouldn't worry too much about it, it will be deleted anyways :) If it's working in the back office now then I'd say it doesn't matter too much.

src/Umbraco.Web/Security/AppBuilderExtensions.cs
```diff
@@ -197,11 +152,11 @@ public static class AppBuilderExtensions
...
                 // logs in. This is a security feature which is used when you
                 // change a password or add an external login to your account.
-                OnValidateIdentity = SecurityStampValidator
-                    .OnValidateIdentity<BackOfficeUserManager, BackOfficeIdentityUser, int>(
-                        TimeSpan.FromMinutes(30),
-                        (manager, user) => manager.GenerateUserIdentityAsync(user),
-                        identity => identity.GetUserId<int>()),
+                OnValidateIdentity = UmbracoSecurityStampValidator
+                    .OnValidateIdentity<BackOfficeSignInManager, BackOfficeUserManager, BackOfficeIdentityUser>(
+                        TimeSpan.FromMinutes(3),
```
emmagarland, line 157: Changed from 30 to 3. Would this be something to put in config?

src/Umbraco.Web/Security/OwinDataProtectorTokenProvider.cs
scottbrady91: https://stackoverflow.com/questions/234059/is-a-memory-leak-created-if-a-memorystream-in-net-is-not-closed/234071

src/Umbraco.Web/Security/BackOfficeUserManager.cs
```diff
@@ -366,65 +325,67 @@ public override Task<IdentityResult> ChangePasswordAsync(int userId, string curr
         /// <param name="user"></param>
         /// <param name="password"></param>
         /// <returns></returns>
-        protected override async Task<bool> VerifyPasswordAsync(IUserPasswordStore<T, int> store, T user, string password)
+        protected override async Task<PasswordVerificationResult> VerifyPasswordAsync(IUserPasswordStore<T> store, T user, string password)
         {
-            var userAwarePasswordHasher = PasswordHasher as IUserAwarePasswordHasher<BackOfficeIdentityUser, int>;
+            var userAwarePasswordHasher = PasswordHasher;
```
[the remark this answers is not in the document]
scottbrady91: Good catch. This override is no longer necessary

src/Umbraco.Web/Security/BackOfficeUserStore.cs
scottbrady91: I forgot to mention this one. Due to the Umbraco bootstrap process, a user with a null email address is created (it's later updated). What's Umbraco's policy with email? Is it required? Have I managed to change the bootstrap user creation somehow?

src/Umbraco.Web/Security/AppBuilderExtensions.cs
scottbrady91: Good catch. Some test settings left in 🙄

src/Umbraco.Web/Security/BackOfficeUserStore.cs
Shazwazza: @scottbrady91  This is a bit of a legacy story I suppose. Today, email is the login so it's required. But there are 2 fields in the db: `userLogin` and `userEmail`. There's a config switch somewhere IIRC that forces the username to be the email but if this is turned off they act independently. So by default, email cannot be blank. I guess I'm unsure where you are referring to where an empty email is set, have any pointers?

src/Umbraco.Web/Editors/UsersController.cs
scottbrady91: Well spotted. I have written some basic tests and refactor of this controller action

src/Umbraco.Web/Security/AuthenticationManagerExtensions.cs
scottbrady91: Looks like the `d.Caption` calls down into `d.Properties["Caption"]`. I ported across the MS implementation, so I assume that the Caption property was added after their implementation was written 🤷‍♂️ Either way, I've updated it 🙂

src/Umbraco.Web/Security/BackOfficeUserStore.cs
scottbrady91: During `NewInstallStep.ExecuteAsync`, a user is retrieved from the database that has an empty email address. This then fails validation during the `UserManager`s `ChangePasswordWithResetAsync` method.

Are you happy creating a new task for investigating this and leaving the check disabled for this PR?

Here's the callstack:
```
Umbraco.Web.Security.BackOfficeUserStore.SetEmailAsync(Umbraco.Web.Models.Identity.BackOfficeIdentityUser user, string email, System.Threading.CancellationToken cancellationToken) Line 296	C#
Umbraco.Web.Security.BackOfficeUserStore.SetNormalizedEmailAsync(Umbraco.Web.Models.Identity.BackOfficeIdentityUser user, string normalizedEmail, System.Threading.CancellationToken cancellationToken) Line 372	C#
[External Code]	
Umbraco.Web.Security.BackOfficeUserManager<Umbraco.Web.Models.Identity.BackOfficeIdentityUser>.ChangePasswordWithResetAsync(int userId, string token, string newPassword) Line 311	C#
Umbraco.Web.Install.InstallSteps.NewInstallStep.ExecuteAsync(Umbraco.Web.Install.Models.UserModel user) Line 73	C#
[External Code]	
Umbraco.Web.Install.Controllers.InstallApiController.ExecuteStepAsync(Umbraco.Web.Install.Models.InstallSetupStep step, object instruction) Line 256	C#
Umbraco.Web.Install.Controllers.InstallApiController.PostPerformInstall(Umbraco.Web.Install.Models.InstallInstructions installModel) Line 106	C#
[External Code]
```

src/Umbraco.Web/Editors/AuthenticationController.cs
```diff
@@ -315,25 +313,26 @@ public async Task<HttpResponseMessage> PostRequestPasswordReset(RequestPasswordR
...
+                    var code = await UserManager.GeneratePasswordResetTokenAsync(identityUser);
                     var callbackUrl = ConstructCallbackUrl(identityUser.Id, code);
 
                     var message = Services.TextService.Localize("resetPasswordEmailCopyFormat",
                         // Ensure the culture of the found user is used for the email!
                         UmbracoUserExtensions.GetUserCulture(identityUser.Culture, Services.TextService, GlobalSettings),
                         new[] { identityUser.UserName, callbackUrl });
 
-                    await UserManager.SendEmailAsync(identityUser.Id,
+                    // TODO: Port email service to ASP.NET Core
```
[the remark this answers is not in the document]
scottbrady91: I did have these initially, but I backed out so that the back office functionality still worked. What would you prefer?

src/Umbraco.Web/Security/BackOfficeUserStore.cs
bergmania: Alright, the issue seems to be because the superadmin (`UserId = -1`) has `email = ""`. This user is inserted into the database in `DatabaseDataCreator`

bergmania: @scottbrady91, as far as I can see, we can just move the following code in `NewInstallStep.ExecuteAsync`:
```
            admin.Email = user.Email.Trim();
            admin.Name = user.Name.Trim();
            admin.Username = user.Email.Trim();

            _userService.Save(admin);
```

up before the password is reset. Then the user has an email when `ChangePasswordWithResetAsync` is called

src/Umbraco.Web/Editors/AuthenticationController.cs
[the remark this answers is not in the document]
bergmania: I have created tasks to reimplement these, so we don't forget it, so I think it is fine with the TODO.

src/Umbraco.Web/Security/BackOfficeUserStore.cs
bergmania: I can do this change, and then we can merge :)

scottbrady91: Thanks, Bjarke

src/Umbraco.Core/ClaimsIdentityExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+            if (identity is ClaimsIdentity claimsIdentity)
+            {
+                username = claimsIdentity.FindFirstValue(ClaimTypes.Name)
+                           ?? claimsIdentity.FindFirstValue("preferred_username");
+            }
+
+            return username;
+        }
+
+        public static string FindFirstValue(this ClaimsIdentity identity, string claimType)
```
JimBobSquarePants, line 37: Public extensions methods like this should really be documented. There is no indication here that the method can return `null`.

PR #8027, opened 2020-04-28
Member type infinite editing

bjarnef:
### Prerequisites

### Description
I noticed a few differences on members compared to content and media, where members didn't had an option to open member type and use infinite editing. Part of the code was prepared to use infite editing.

I have compared with media and tried to keep as much as possible consistent with these controllers for media:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/mediatypes/edit.controller.js

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/media/media.edit.controller.js


bjarnef:
I need to replace hyphen in `args.memberType.key` since this contained "-" while it didn't in `$scope.content.contentType.key`.

Not sure why this is different from media type?
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/media/media.edit.controller.js#L123

image

bjarnef:
Demo of these changes.



@nielslyngsoe it seems some of inifite editing for members was implemented, but not completely done. It would be great to have this to make it more consistent with content and media.

emmaburstow:
Hey Bjarne,

Thanks for the hard work here! We'll let you know if we have any questions, as ever.

Em

nielslyngsoe:
@bjarnef I agree, could you create issues on each specific part. That would be nice. Thanks

bjarnef:
@nielslyngsoe which part? This part?

I need to replace hyphen in `args.memberType.key` since this contained "-" while it didn't in `$scope.content.contentType.key`.

bjarnef:
@nielslyngsoe it would be great if we could have infinite editing for members as well, like we also have content apps for members making it more consistent with content and media.

nul800sebastiaan:
This works brilliantly, thanks very much again @bjarnef! 👍 👍

PR #8031, opened 2020-04-29
Add varchar(max) support in DTO attributes

### LINKED ISSUE #5204: v8: Add varchar(max) support
We need to be able to support NvarcharMax for SQL server in Umbraco. Currently there isn't any support for this in the code, only NText which is deprecated. 

We need to somehow support NvarcharMax but also keeping in mind that however this works would need to fallback to NText for SqlCe. 

The migration syntax would need to support NvarcharMax along with DTO attributes.

## Original post:

In v8 there is an ability to create DB tables using a builder interface. As part of this you can do things like define columns and their types. These then get built into a SQL statement to run on the various database types.

There is currently an ability to define columns as strings and provide an int size, such as

    Create.Table("myTable").WithColumn("myColumn").AsString(255)

This will then get converted into a SQL statement that creates a column of type `nvarchar(255)`. 

Ones thing missing though is an ability to create a string column of type `nvarchar(max)` as max is clearly not an int value which the .AsString method requires.

We could really do with an additional means of defining a string value with it's size set to `max`

JoseMarcenaro:
This fixes #5204

### Description

A `SpecialDbTypes.NVARCHARMAX` has been added to the enum in order to allow the definition of nvarchar(max) fields in Migration schemas:
```
[SpecialDbType(SpecialDbTypes.NVARCHARMAX)]
public string Body {get; set;}
```

The implementation for this new enum member has been added in `SqlSyntaxProviderBase`, and overriden in `SqlCeSyntaxProvider`  because SqlCE does not implement nvarchar(max) - ntext should be used instead.

### DIFF
--- a/src/Umbraco.Core/Persistence/DatabaseAnnotations/SpecialDbTypes.cs
+++ b/src/Umbraco.Core/Persistence/DatabaseAnnotations/SpecialDbTypes.cs
@@ -7,6 +7,7 @@
     public enum SpecialDbTypes
     {
         NTEXT,
-        NCHAR
+        NCHAR,
+        NVARCHARMAX
     }
 }
--- a/src/Umbraco.Core/Persistence/PocoDataDataReader.cs
+++ b/src/Umbraco.Core/Persistence/PocoDataDataReader.cs
@@ -79,6 +79,9 @@ namespace Umbraco.Core.Persistence
                         case SpecialDbTypes.NCHAR:
                             sqlDbType = SqlDbType.NChar;
                             break;
+                        case SpecialDbTypes.NVARCHARMAX:
+                            sqlDbType = SqlDbType.NVarChar;
+                            break;
                         default:
                             throw new ArgumentOutOfRangeException();
                     }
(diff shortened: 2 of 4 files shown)


emmaburstow:
Hiya @JoseMarcenaro,

Thanks for the PR and congrats on your first PR to this repo. We'll take a look at what you've done here and ask if we have any questions. Once it's been merged, we'll add a shiny new contributor badge to your our.umbraco.com profile so please let us know your username there :)

Speak soon,

Emma

JoseMarcenaro:
Hi @emmaburstow  - thanks to you and the huge community actively supporting Umbraco!
My username in our.umbraco.com is Jose Marcenaro, id 156385.

nul800sebastiaan:
Hi there @JoseMarcenaro and sorry for the long delay on this one! This looks great to me and will be available in v8.8!

Thanks again and congrats on your first PR for Umbraco CMS! Congrats on your contributors badge! 🏅 https://our.umbraco.com/members/id:156385/

Shazwazza:
Great stuff! Now we just need to get the NTEXT columns converted and migrations added , not sure if we have a task for that, anyone know?

stevetemple:
> Great stuff! Now we just need to get the NTEXT columns converted and migrations added , not sure if we have a task for that, anyone know?

Did this every happen?

Shazwazza:
@stevetemple not that i know of

Issue #8035, opened 2020-04-30
umbSessionStorage getter needs to check wether the key exists


nielslyngsoe:
Currently our getter for umbSessionStorage doesn't perform a check for the key, resulting in the coding trying to parse 'undefined'. resulting in this JS error:

![image](https://user-images.githubusercontent.com/6791648/80692958-9f623480-8ad2-11ea-90d0-306489fc1aa8.png)

PR #8036, opened 2020-04-30
UmbSessionStorage: Make sure key exists before using it as a JSON string

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/8035

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/util.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/util.service.js
@@ -206,7 +206,11 @@ function umbSessionStorage($window) {
     return {
 
         get: function (key) {
-            return JSON.parse(storage["umb_" + key]);
+            if (storage["umb_" + key]) {
+                return JSON.parse(storage["umb_" + key]);
+            } else {
+                return null;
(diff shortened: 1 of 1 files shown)


bjarnef:
@nielslyngsoe https://github.com/umbraco/Umbraco-CMS/pull/8014 also fixes this although it would be great with check if the key exists.

nielslyngsoe:
@bjarnef great. I think we should have both.

bjarnef:
If `storage["umb_" + key]` is null, undefined, an object it just return this value :)

bjarnef:
I think the changes in this PR would return null if `storage["umb_" + key]` was `0` or `false`.

nielslyngsoe:
@bjarnef I think you are right, let me have another look.

bjarnef:
The changes in PR #8014 does the same as in `angular.fromJson()` so it should break anything of the existing behaviour.

nielslyngsoe:
@bjarnef good point, lets keep existing as-is. It would just have been nicer to be consistent in terms of whether you get null or undefined back. But I will save that for another day.

PR #8043, opened 2020-05-01
Upload media and select media type

bjarnef:
### Prerequisites

### Description
When having multiple allowed media types (e.g. default "Image" and a "Special Image" under e.g. "Folder" media type) and dragging files/images to dropzone, it shows an overlay to select media type. This is however the older/legacy `umb-overlay` instead of using `overlayService` or `editorService`.

In this case we could use `editorService.mediaTypePicker()` however I have chosen to use `overlayService` using `itempicker` view, because this is used when inserting grid editors (where you choose which editor to use).
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js#L320-L341

This is how it looks at the moment:

chrome_2020-05-01_12-32-21

and when inserting an item in grid:

image

With these changes after uploading files to dropzone in e.g. "Folder" with "Image" and "Special Image" allowed. For this case "Special Image" (alias: specialImage) has just a single property Upload image (umbracoFile) using Image Cropper - just like the default "Image" media type.

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbfiledropzone.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbfiledropzone.directive.js
@@ -20,7 +20,7 @@ TODO
 
 angular.module("umbraco.directives")
     .directive('umbFileDropzone',
-        function($timeout, Upload, localizationService, umbRequestHelper) {
+        function ($timeout, Upload, localizationService, umbRequestHelper, overlayService) {
             return {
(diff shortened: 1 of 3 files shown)


emmaburstow:
Morning Bjarne! 

Thanks for the work here. And lovely documentation of it too. Speak soon,

Emma

nul800sebastiaan:
This is really lovely, thanks so much @bjarnef ! 👍

PR #8044, opened 2020-05-01
Adjust nested content overlay

bjarnef:
### Prerequisites

### Description
This PR replace the older/legacy `umbOverlay` with an overlay using `overlayService` and makes it more consistent with overlay shown in grid editor.

At the moment there is only a few references left to `umb-overlay` in core, so maybe we can completely remove the references from core and maybe in Umbraco v9 removed `umbOverlay` directive. Until the we can remove the references in core, where it makes sense.

One of these references have been removed in this PR and another in https://github.com/umbraco/Umbraco-CMS/pull/8043.

For this PR and have used default starter kit and created a new nested content element type "Special Feature" with a property "USP", which is just a child the document type of "Feature".

**Before**

chrome_2020-05-01_13-46-11

**After**

image


bjarnef:
Besides the two mentioned PR's we have these 5 references left (actually only 4 since one of them is a comment). Furthermore there's the directive itself.

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/application/umb-app-header.html#L48-L54

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI/Umbraco/Views/Default.cshtml#L99-L104

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/umbchildselector.directive.js#L24-L30 (comment)

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.html#L139-L195

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/listview/listview.html#L233-L238

@nielslyngsoe could we look at removing the last few references for `umb-overlay` in core?

bjarnef:
Okay, I don't think we can remove `umb-overlay` here https://github.com/umbraco/Umbraco-CMS/blob/3e9c759426e5b852f0ea643a26b68c73f4048343/src/Umbraco.Web.UI/Umbraco/Views/Default.cshtml#L99-L104 as I think `overlayService` use this, but we can replace most of the `umb-overlay` in the views in favor of using `overlayService`.

nielslyngsoe:
> @nielslyngsoe could we look at removing the last few references for `umb-overlay` in core?

@bjarnef definitely. Let's get rid of as many references as possible, to simplify the code. We cant remove it completely because we might have packages using it. But it would be a great start to align the CMS code.

nul800sebastiaan:
Sorry for the delay @bjarnef - this one is good to go, thanks very much!

PR #8045, opened 2020-05-01
Adjust example for overlay in umb-child-selector

bjarnef:
### Prerequisites

### Description
This PR shouldn't break anything. The `umb-child-selector` component is used two places in core: document type and media type permission views.

These already used `overlayService`. I have adjusted these a bit and updated the example the use `overlayService` instead since `umb-overlay` is legacy and might be removed in a future version of Umbraco.

If developers are using this in packages and have added `umb-overlay` to their views and implemented show/hide of this overlay in the `add` function in the controller, this would still work.

It is however recommended to use `overlayService` now, so I have therefore updated the example shown in the directive.


emmaburstow:
Hey Bjarne,

You've been busy! Thanks for this work. We'll check it out and shout if we have questions.

Emma

nul800sebastiaan:
Wonderful, works like a charm! Thanks @bjarnef! 👍

PR #8046, opened 2020-05-01
Replace ysod umb-overlay using overlayService.ysod

bjarnef:
### Prerequisites

### Description
This PR replace the legacy `umb-overlay` here to show ysod and instead use `overlayService.ysod()` just like here: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/media/media.delete.controller.js#L62


emmaburstow:
Hi Bjarne,

Lovely! A step in the right direction :) Thanks for the work,

Emma

nathanwoulfe:
How good are changes that result in reduced LOC... Thanks @bjarnef for another great update.

PR #8047, opened 2020-05-01
Use overlayService for user overlay

bjarnef:
### Prerequisites

### Description
This PR has a few changes so user dialog use `overlayService` and add a new property `name` to be able to set value of `data-element` in overlay, when using the `overlayService`.

The `data-element` is used in the `Get Started` tour after the step clicking on user avatar.


emmaburstow:
Thanks again Bjarne! Lovely work as usual. We'll take a look and let you know if we need anything,

Emma

nul800sebastiaan:
Good stuff @bjarnef - thanks very much! 👍

Issue #8050, opened 2020-05-02
Missing localization text for custom audit events.


KevinJump:
Custom audit events are missing a label and text in the back end 

![image](https://user-images.githubusercontent.com/431231/80863111-d7ec4480-8c71-11ea-87c6-cc4126e323b0.png)

Ideally you should be able to set custom label / colour for custom events, but a quick fix is to add the labels in the Lang files so it shows parameters like the other audit events do.

nul800sebastiaan:
Fixed in #8051

PR #8051, opened 2020-05-02
Add keys for custom audit events #8050

KevinJump:
Adding keys in lang file for custom events. so we can have text and parameters show. 

I've set the custom key to %0% which means the text will be what ever is entered in as parameters in the umbracoLog table for the custom event.

image

Issue #8050


emmaburstow:
Oooh fab. Thanks Kevin! We'll take a look.

Em

nul800sebastiaan:
Thanks @KevinJump - that seems like an oversight and these should have been there from the beginning! 👍

PR #8057, opened 2020-05-04
Don't re-create indexes twice when rebuilding in the back office

Shazwazza:
when we rebuild indexes in the back office we are actually calling index.CreateIndex() 2 times before populating.

This code is simple, don't call `index.CreateIndex();` because it is called inside `_indexRebuilder.RebuildIndex(indexName);`


nul800sebastiaan:
Duh, that's definitely some unneeded overhead!

nul800sebastiaan:
Cherry picked for 8.6.2 in 9f2c1cd3e633936f990e09d779dfd76e47365c65

Issue #8038, opened 2020-04-30
SqlMainDom setting for Azure does not work in Load Balancing


Shazwazza:
From 8.6.0, for running on Azure we have a new setting as documented here https://our.umbraco.com/documentation/Getting-Started/Setup/Server-Setup/azure-web-apps#recommended-configuration 

the issue is this setting:

```
<add
    key="Umbraco.Core.MainDom.Lock"
    value="SqlMainDomLock" />
```

Which should be applied to all servers when running on azure but in load balancing this causes problems because that setting doesn't support multiple servers. So we need to fix this so that configuration for running on azure is simple - like the docs say :)

bug found from here https://github.com/umbraco/Umbraco-CMS/issues/8006#issuecomment-620328022

---
_This item has been added to our backlog [AB#6441](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6441)_

Shazwazza:
PR is here https://github.com/umbraco/Umbraco-CMS/pull/8058

PR #8058, opened 2020-05-04
Fixes SqlMainDomLock setting for Azure does not work in Load Balancing #8038

Shazwazza:
Fixes: SqlMainDomLock setting for Azure does not work in Load Balancing #8038

The reason this doesn't work in LB is because there is only a single 'key' for the database table `umbracoKeyValue` to use for this lock which means only a single server/app can be registered for this distributed lock. 

The purpose of MainDom is to synchronize access to certain resources between concurrent AppDomains. This means that the lock needs to be specific to one web application on one web server. This is exactly what the original MainDom did with it's system wide semaphore lock. This would work in load balancing (outside of azure) because implicitly a system-wide semaphore lock is machine specific. This Sql lock needs to take into account this same uniqueness.

So now it does, the key in the DB is a combination of the original MainDom's semaphore lock Id (which takes into account the web app's Id and physical path) and the current server's name.

The changes are minimal but should make sense.

For testing, use the SqlMainDomLock config:

```xml
<add
    key="Umbraco.Core.MainDom.Lock"
    value="SqlMainDomLock" />
```

when you run the site have a look at the data in the table `[umbracoKeyValue]` and you'll see an entry for: `Umbraco.Core.Runtime.SqlMainDom-****` which is the lock for your current machine/app. If you bump the web.config and refresh you'll notice the value for this row changes. Rows do not keep getting added to the table because the hash is always consistent. If you only have the front-end of your app open (not the back office since that polls), then bump the web.config, that will shut down the app and you'll see that row be removed. That is exactly how SqlMainDomLock works and that's what should be happening. When the app shutsdown (without restarting with new requests), the row will be removed, when a new request comes in it will be re-added. When the app restarts the value will be changed. (though that is all just testing the original SqlMainDomLock stuff which we know works, for this we just want to test that it's not creating more than one row per machine/app)

### DIFF
--- a/build/NuSpecs/tools/Views.Web.config.install.xdt
+++ b/build/NuSpecs/tools/Views.Web.config.install.xdt
@@ -11,7 +11,7 @@
         <host factoryType="System.Web.Mvc.MvcWebRazorHostFactory, System.Web.Mvc, Version=5.2.7.0, Culture=neutral, PublicKeyToken=31BF3856AD364E35" xdt:Locator="Match(factoryType)" xdt:Transform="SetAttributes(factoryType)" />
         <pages pageBaseType="System.Web.Mvc.WebViewPage">
             <namespaces>
-                <add namespace="Umbraco.Web.PublishedContentModels" xdt:Transform="InsertIfMissing" />
+                <add namespace="Umbraco.Web.PublishedModels" xdt:Transform="InsertIfMissing" />
             </namespaces>
         </pages>
     </system.web.webPages.razor>
--- a/src/Umbraco.Core/Runtime/SqlMainDomLock.cs
+++ b/src/Umbraco.Core/Runtime/SqlMainDomLock.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Web;
 using Umbraco.Core.Logging;
 using Umbraco.Core.Persistence;
 using Umbraco.Core.Persistence.Dtos;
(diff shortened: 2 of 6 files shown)


bergmania:
It is not directly related to this task, but I cannot upgrade Umbraco from a version before SqlLock (e.g. 8.3) when I have `<add key="Umbraco.Core.MainDom.Lock" value="SqlMainDomLock" />` in the web.config. Due to 
```
System.Data.SqlClient.SqlException (0x80131904): Invalid object name 'umbracoLock'.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
```

If I remove it, I can upgrade Umbraco and then re-insert it.

bergmania:
@Shazwazza this fix seems to work as expected, and changes make sense.

I think it is okay, I could not update from the early version. I also used a config value not available in the version I updated from

Shazwazza:
@bergmania aha, i guess that's another issue similar to this one https://github.com/umbraco/Umbraco-CMS/issues/7967 I can see about a fix for that separately too.

bergmania:
@Shazwazza, I guess you are right. This issue could very well be the same

Shazwazza:
@bergmania Damnit, this was targeting v8contrib but it was supposed to be v8 dev, now the commits and changes it shows are not actually correct. but it was merged?

Issue #7758, opened 2020-03-04
V8: If a user has no Dashboards, Sections don't load


joepvtl:
When a user has no Dashboard the Sections don't load and give an error once you login.

## Umbraco version

I have tested this on: Umbraco 8.4.0 and 8.6.0-rc

### Bug summary

It gives a 500 because the querystings parameter is empty when adding the recyclebin to the TreeNodes.

### Steps to reproduce

1. Install a clean version of Umbraco.
2. Create a user that has the default 'Writer' group.
3. Make sure the user has the Content root as startnode.
4. Remove the RedirectUrlDashboard in a composer.
5. Login as the user.

### Expected result

No errors when logging in and Sections in the navigation header.

### Actual result

A 500 error when logging in with no Sections.

---
_This item has been added to our backlog [AB#5716](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/5716)_

nul800sebastiaan:
Really weird! I'm looking into it.

joepvtl:
Hey @nul800sebastiaan,

Any news regarding this issue?

nul800sebastiaan:
Sorry @joepvtl it kinda fell off my radar when I couldn't figure out what to do about it. The code seems very weird to me and I don't understand it. I'm sure there's someone at HQ who can help.. :-)

In any case, this is something we'll look at in the future and we'll try to fix it for 8.7.0, so it's going to take a while. Until then I'm afraid you'll have to live with adding at least some kind of dashboard that all your users can see. Might I suggest a friendly dashboard telling your writers and editors where they can reach someone if they have some problems using the system, or for example something with useful links: https://our.umbraco.com/packages/backoffice-extensions/the-dashboard/ (I know there's no v8 version of this yet, but you could use it for inspiration! 👍

joepvtl:
No problem! I already worked around it by just creating a Dashboard with an empty view.

PR #8061, opened 2020-05-05
Preventive check that will enable sections load for users with no dashboards

elit0451:
Fix for #7758 

Description about the problem and steps to reproduce/test can be found in #7758!

The error that a user with no dashboards was getting when logging in, was that we were trying to get a missing required value parameter (`application`) from the query string -> [code ref](https://github.com/umbraco/Umbraco-CMS/blob/d7469e6576fbe4bf4d23f9046399b3899c0ff58c/src/Umbraco.Web/Trees/ContentTreeControllerBase.cs#L335). 

I suspect that passing `null` as the value of the `queryString` when `hasDashboards` is **false** (in `GetSections()` in the [SectionController.cs](https://github.com/umbraco/Umbraco-CMS/blob/d7469e6576fbe4bf4d23f9046399b3899c0ff58c/src/Umbraco.Web/Editors/SectionController.cs#L60)) might not be the right approach but just to be sure we now check if the `queryString` has the key `application` before we try to get its value.

### DIFF
--- a/src/Umbraco.Web/Trees/ContentTreeControllerBase.cs
+++ b/src/Umbraco.Web/Trees/ContentTreeControllerBase.cs
@@ -322,8 +322,10 @@ namespace Umbraco.Web.Trees
 
                 var nodes = GetTreeNodesInternal(id, queryStrings);
 
-                //only render the recycle bin if we are not in dialog and the start id id still the root
-                if (IsDialog(queryStrings) == false && id == Constants.System.RootString)
+                //only render the recycle bin if we are not in dialog and the start id is still the root
+                //we need to check for the "application" key in the queryString because its value is required here,
+                //and for some reason when there are no dashboards, this parameter is missing  
+                if (IsDialog(queryStrings) == false && id == Constants.System.RootString && queryStrings.HasKey("application"))
                 {
                     nodes.Add(CreateTreeNode(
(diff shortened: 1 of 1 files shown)


Shazwazza:
@elit0451 ok so here's what is going on. I'm unsure if you had a look at the Stack Trace window when debugging but this is one of the most useful things you can use when debugging. 

image

By double clicking on that line in the Call Stack it will take you to the code at that snapshot, you can even inspect variables and their values in that snapshot. 

image

So the reason why it's null is because we are passing in null :) This logic that executes the content tree getting nodes isn't actually being used in the UI to get the nodes. This is being called as a call to SectionController.GetSections() and the purpose is stated in the comments there:

```cs
//now we can add metadata for each section so that the UI knows if there's actually anything at all to render for
//a dashboard for a given section, then the UI can deal with it accordingly (i.e. redirect to the first tree)
```

The result of GetSections() is to prepare the angular app so it knows how to deal with sections. 

image

I'm unsure why it's being called twice ... i think that's a bug :) 

So your fix is a good fix since as it turns out query strings can be null in this case because we are actually passing in null. In a normal case this can never be null, but this is one of those weird things.

Shazwazza:
I've cherry picked this to the 8.6 branch

Issue #7964, opened 2020-04-15
error icon doesn't disappear after correcting validation


shearer3000:
Reported for umbraco 8.6.0

When attempting to publish a page where a property doesn’t meet validation (e.g. mandatory) then publishing will fail and the ‘error icon’ will show as expected:
![failed-validation](https://user-images.githubusercontent.com/2433680/79380809-136dea00-7fb5-11ea-8823-2673e1618bfe.png)

However, correcting the property will mean republishing once more succeeds, but the icon still remains (in the language dropdown too):
![dropdown-after-validation](https://user-images.githubusercontent.com/2433680/79380821-17017100-7fb5-11ea-8d43-72ab0e98072c.png)

Expected  
Error styling disappears after correcting property validation and republishing

Actual
Icon and red text remains

---
_This item has been added to our backlog [AB#6284](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6284)_

nul800sebastiaan:
Thanks for reporting, I see this is a new icon added to v8.6 and could do with some improvement for sure! I'll let the team know that we should have a look at it.

nielslyngsoe:
This will be fixed soon, PR here: https://github.com/umbraco/Umbraco-CMS/pull/8064

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8064

PR #8064, opened 2020-05-05
UI: Corrected validation indication in variant-picker

nielslyngsoe:
This PR makes sure that validation indication in variant-picker only stays as long as an error is present. Plus avoid displaying field-errors, we only want to use property errors for this indication.

This PR is an addition to the work for Segments, the reason for this PR to target this is that Segments will soon be merged and include a ton of changes to variant-picker. This way we minimize merge errors.

fixes https://github.com/umbraco/Umbraco-CMS/issues/7964

---
_This item has been added to our backlog [AB#6512](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6512)_


warrenbuckley:
All good :)

PR #8065, opened 2020-05-05
Validation errors reworked to incorporate segment errors

PerplexDaniel:
Hi @nielslyngsoe ,

I did some initial work to have validation messages show up in the Save/Publish dialog.
This seems to work pretty well already. I do run into other issues related to Saving/Publishing in general, as you pointed out the `ContentService` does not seem to take segments into account much and will just publish all variants of a culture. So my attention will move to that area next.

However, perhaps you can have a quick look at this when you have time (no rush). Shannon already did some work to add the segment to the ModelState returned from the server. I expanded it a bit, reworked some stuff from `CultureXYZ` to `VariantXYZ` and I believe this already improves things. 

For example, this now renders properly when there are server side validation errors for a segment (except the message format still refers to every variant as a "language"...):
segment-error

Summary of changes:

Server:
- In general we have added a "segment" parameter to the validation related methods.
  The items themselves (ContentVariantSave) already had this parameter available.
- A variant was identified by the culture only in many places.
  This has been updated to a tuple of (culture, segment).

Client:
- Small bugfix for the "htmlId" property used in the client to properly match the ModelState key
  returned from the server (added "_" suffix).
- Updated "compositeId" to use "null" when the segment is NULL, like was done by Shannon in ModelState
  on the server.

TODO:
- Look into the places where mandatory cultures are being used, i.e.
  `_allLangs.Value.Values.Where(x => x.IsMandatory).Select(x => x.IsoCode)`
  It is unclear to me how that should interact with segments. Should we consider all
  variants with a mandatory culture actually mandatory, or perhaps only the variants
  that have segment NULL?

---
_This item has been added to our backlog [AB#6515](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6515)_

### DIFF
--- a/src/Umbraco.Web/Editors/ContentController.cs
+++ b/src/Umbraco.Web/Editors/ContentController.cs
@@ -37,6 +37,7 @@ using Umbraco.Core.Models.Entities;
 using Umbraco.Core.Persistence;
 using Umbraco.Core.Security;
 using Umbraco.Web.Routing;
+using Umbraco.Core.Collections;
 
 namespace Umbraco.Web.Editors
 {
@@ -706,12 +707,19 @@ namespace Umbraco.Web.Editors
                     {
                         if (variantCount > 1)
                         {
-                            var cultureErrors = ModelState.GetCulturesWithErrors(Services.LocalizationService, cultureForInvariantErrors);
-                            foreach (var c in contentItem.Variants.Where(x => x.Save && !cultureErrors.Contains(x.Culture)).Select(x => x.Culture).ToArray())
+                            var variantErrors = ModelState.GetVariantsWithErrors(cultureForInvariantErrors);
+
+                            var validVariants = contentItem.Variants
+                                .Where(x => x.Save && !variantErrors.Contains((x.Culture, x.Segment)))
+                                .Select(x => (culture: x.Culture, segment: x.Segment));
+
+                            foreach (var (culture, segment) in validVariants)
                             {
-                                AddSuccessNotification(notifications, c,
+                                var variantName = GetVariantName(culture, segment);
+
+                                AddSuccessNotification(notifications, culture, segment,
                                     Services.TextService.Localize("speechBubbles/editContentSendToPublish"),
-                                    Services.TextService.Localize("speechBubbles/editVariantSendToPublishText", new[] { _allLangs.Value[c].CultureName }));
+                                    Services.TextService.Localize("speechBubbles/editVariantSendToPublishText", new[] { variantName }));
                             }
                         }
                         else if (ModelState.IsValid)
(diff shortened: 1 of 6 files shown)


nielslyngsoe:
Hi @PerplexDaniel since we now have merged the segment feature into dev, you can target your PR towards dev as well. (I have re-target this PR)

I will have a look at the changes

nielslyngsoe:
@PerplexDaniel in regards to your question about mandatory, the mandatory language should only make the none-segmented culture variant mandatory. So only affect variants with segment = null.
We might introduce the option to make segments mandatory at one point, but for now they are optional.

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -1040,14 +1049,16 @@ private bool SaveScheduleVariant(ContentItemSave contentItem)
                 if (!isPublished && releaseDates.Count == 0)
                 {
                     //can't continue, a mandatory variant is not published and not scheduled for publishing
-                    AddCultureValidationError(culture, "speechBubbles/scheduleErrReleaseDate2");
+                    // TODO: Add segment
```
Shazwazza, line 1052: Do we need this TODO?

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -1040,14 +1049,16 @@ private bool SaveScheduleVariant(ContentItemSave contentItem)
...
+                    AddVariantValidationError(culture, null, "speechBubbles/scheduleErrReleaseDate2");
                     isValid = false;
                     continue;
                 }
                 if (!isPublished && releaseDates.Any(x => nonMandatoryVariantReleaseDates.Any(r => x.Date > r.Date)))
                 {
                     //can't continue, a mandatory variant is not published and it's scheduled for publishing after a non-mandatory
-                    AddCultureValidationError(culture, "speechBubbles/scheduleErrReleaseDate3");
+                    // TODO: Add segment
```
Shazwazza, line 1060: Do we need this TODO?

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -1376,14 +1386,40 @@ private bool PublishCulture(IContent persistentContent, IEnumerable<ContentVaria
...
+        /// Creates the human readable variant name based on culture and segment                
+        /// </summary>
+        /// <param name="culture">Culture</param>
+        /// <param name="segment">Segment</param>
+        /// <returns></returns>
+        private string GetVariantName(string culture, string segment)
+        {
+            if(culture.IsNullOrWhiteSpace() && segment.IsNullOrWhiteSpace())
+            {
+                // TODO: Get name for default variant from somewhere?
```
Shazwazza, line 1414: What do we expect to happen here?

src/Umbraco.Web/Editors/ContentController.cs
nielslyngsoe: Yes, if we want to deal with segments then we need to parse the segment not 'null'.

nielslyngsoe: (Sorry, I'll edit my comment to this:) If we want to support mandatory segmented variants we should add segments. But currently its not part of the feature therefor it can stay 'null' and we can remove the Todo.

src/Umbraco.Web/Editors/ContentController.cs
nielslyngsoe: I would propose we return a translated/localized name here.

PR #8070, opened 2020-05-06
Added support for segments to File Uploads

PerplexDaniel:
As per @nielslyngsoe's [comment](https://github.com/umbraco/Umbraco-CMS/pull/8065#issuecomment-624531758) I'm targeting this to `v8/dev`. 

This adds a "segment" parameter to posted file data, so you can post files for multiple variants (segmented / non segmented) at the same time. I did not completely test all upload related editors yet, but I tested with "File Upload" and "ImageCropper" and it looked good.

As an aside, it seems there is a bug in Umbraco stable when uploading the same file for the same property of different cultures that leads to only 1 file being generated on disk.  The filename of a media file seems determined only based on the content id + property id + sanitized filename, so if I upload `some-image.jpg` for the `image` property of a doctype for cultures `nl-NL` and `en-US` there will only be 1 file total. Which means if I delete it for `nl-NL` it will also be gone for `en-US` and gives a 404 in Umbraco... 

At least that's something I saw when testing this, and I'm pretty sure this was due to code already shipped in a public release. `UniqueMediaPathScheme.GetFilePath`. I can look into that later, I have not filed a bug report for that because I haven't checked this in a released version yet.

---
_This item has been added to our backlog [AB#6523](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6523)_

### DIFF
--- a/src/Umbraco.Core/Models/Editors/ContentPropertyFile.cs
+++ b/src/Umbraco.Core/Models/Editors/ContentPropertyFile.cs
@@ -17,6 +17,11 @@
         public string Culture { get; set; }
 
         /// <summary>
+        /// When dealing with content variants, this is the segment for the variant
+        /// </summary>
+        public string Segment { get; set; }
+
+        /// <summary>
         /// An array of metadata that is parsed out from the file info posted to the server which is set on the client. 
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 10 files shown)


Shazwazza:
Thanks @PerplexDaniel !

The code looks good - approved! 

regarding the issue you found, prob best to log an issue for it :) 

regarding the different file upload editors, there are only 2 that you mentioned :)

I won't merge yet until more testing is done.

nielslyngsoe:
I've tested this feature for invariant, culture variant, segment variant and segmented + culture variant document types.
Done by uploading different files for each variant for file-upload property editor.

PR #8071, opened 2020-05-06
Adjust property settings

bjarnef:
### Prerequisites

### Description
This adjust property settings for datatypes and macro parameter settings to use `<button>` elements instead and furthermore ensure the settings icon on datatype property editor always is in top-right corner, when name is a longer value.

**Before**

**After**


emmaburstow:
Hey Bjarne,

thanks for the work here, another fab improvement to the user journey. We'll take a look and let you know if we need anything. 

Emma

bjarnef:
Unfortunately this wasn't merged in time and now this has changed in `v8/contrib` which making this obsolete, so I'll close this one.

PR #8074, opened 2020-05-07
Update slider to handle merging of overlapping tooltips

bjarnef:
### Prerequisites

### Description
This PR update noUiSlider to latest version, which have fixes some bugs. Furthermore it has added default styling to support merging overlapping tooltips. The documentation has a example to merge overlapping tooltip (like some other slider plugin does, e.g [Ion.RangeSlider](http://ionden.com/a/plugins/ion.rangeSlider/)).

**Before**

**After**

### DIFF
--- a/src/Umbraco.Web.UI.Client/package.json
+++ b/src/Umbraco.Web.UI.Client/package.json
@@ -38,7 +38,7 @@
     "lazyload-js": "1.0.0",
     "moment": "2.22.2",
     "ng-file-upload": "12.2.13",
-    "nouislider": "14.1.1",
+    "nouislider": "14.4.0",
     "npm": "6.13.6",
     "signalr": "2.4.0",
     "spectrum-colorpicker": "1.8.0",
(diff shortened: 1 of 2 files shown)


bjarnef:
@nielslyngsoe this seems to work very well 👍

bjarnef:
In https://github.com/umbraco/Umbraco-CMS/pull/8074/commits/5a3389e731b98a242961b9272f8f6fd0b167163f I have adjusted this a bit so it filter to unique values, which mean when two handles have exact same value, it just show a single tooltip value (instead of showing same value twice separated with `-`).

emmaburstow:
Thanks for the explanation and the work, Bjarne. We'll check the work through and reach out if we need more input from you.

Em

nielslyngsoe:
@bjarnef I love it, Thanks!

nul800sebastiaan:
Thanks @bjarnef - this definitely helps a lot with the user experience! 👍

PR #8079, opened 2020-05-08
Add aria-hidden to icons in url lists

bjarnef:
### Prerequisites

### Description
This PR adjust the url lists a bit adding `aria-hidden="true"` to these icons. Furthermore in urllist it has simplified this: `<span>{{value.linkText || value.url}}</span>`


emmaburstow:
Thanks for the work here @bjarnef :)

We'll take a look and shout if we need you 

Em

nul800sebastiaan:
Lovely, thanks very much @bjarnef !

Issue #7799, opened 2020-03-12
Error while migrating from v-7.15.3 to v-8.5.4 - String or binary data truncated


guruprasad83:
My Current Umbraco version is 7.15.3. I am trying to migrate it to 8.5.4
I followed this blog -https://our.umbraco.com/documentation/getting-started/setup/upgrading/migrating-to-v8 
By creating new site and then started migration to the old database.
But am getting error - 

> The database configuration failed with the following message: String or binary data would be truncated.The statement has been terminated. Please check log file for additional information

Any help would be greatly appreciated

Snippet from my log file below

```
[...]

PR #8080, opened 2020-05-08
Netcore: Implemented content model binder.

AndyButland:
Relating to issue #7799 

In this PR I've taken the `ContentModelBinder` and related components and migrated them to .Net core.

Am a bit limited by time/understanding/current state of project in terms of what's ported already to properly test it, so really have just translated into the new framework, got everything building and added some unit tests,  Hopefully it's either right or not far off for when other parts are migrated and it can be verified in real use.

### DIFF
--- a/src/Umbraco.Web.Common/Filters/ModelBindingExceptionFilter.cs
+++ b/src/Umbraco.Web.Common/Filters/ModelBindingExceptionFilter.cs
@@ -1,13 +1,15 @@
 ﻿using System;
-using System.Configuration;
 using System.Net;
 using System.Text.RegularExpressions;
-using System.Web.Mvc;
-using Umbraco.Core.Configuration;
+using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
 using Umbraco.Core;
-using Umbraco.Web.Composing;
+using Umbraco.Core.Configuration;
+using Umbraco.Core.Models.PublishedContent;
+using Umbraco.Web.Common.ModelBinders;
 
-namespace Umbraco.Web.Mvc
+namespace Umbraco.Web.Common.Filters
 {
     /// <summary>
     ///     An exception filter checking if we get a <see cref="ModelBindingException" /> or <see cref="InvalidCastException" /> with the same model.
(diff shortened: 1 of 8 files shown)


src/Umbraco.Web.Common/Filters/ModelBindingExceptionFilter.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            var disabled = _exceptionFilterSettings?.Disabled ?? false;
+            if (_publishedModelFactory.IsLiveFactory()
+                && !disabled
+                && !filterContext.ExceptionHandled
+                && ((filterContext.Exception is ModelBindingException || filterContext.Exception is InvalidCastException)
+                    && IsMessageAboutTheSameModelType(filterContext.Exception.Message)))
+            {
+                filterContext.HttpContext.Response.Headers.Add(HttpResponseHeader.RetryAfter.ToString(), "1");
+                filterContext.Result = new RedirectResult(filterContext.HttpContext.Request.GetDisplayUrl(), false);
```
bergmania, line 44: I think we need to use `GetEncodedUrl()` here:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.extensions.urihelper.getdisplayurl?view=aspnetcore-3.1

src/Umbraco.Web.UI.NetCore/Startup.cs
```diff
@@ -47,7 +48,10 @@ public void ConfigureServices(IServiceCollection services)
             services.AddUmbracoCore(_env, out var factory);
             services.AddUmbracoWebsite();
 
-            services.AddMvc();
+            services.AddMvc(options =>
```
bergmania, line 51: I think this should be removed.. I think we can explicit add this modelbinder to the required actions, like we do for `ContentItemBinder`

src/Umbraco.Web.Common/Filters/ModelBindingExceptionFilter.cs
AndyButland: Agreed, have updated.

src/Umbraco.Web.UI.NetCore/Startup.cs
AndyButland: Have removed.

bergmania:
As always, thank you for the good work :) 🚀

AndyButland:
You're welcome.

Issue #8082, opened 2020-05-08
Textarea character limit trimming trailing spaces


alanwthom:
When adding a character limit to the textarea property, the character count only updates when the last character of the input is not a space.

## Umbraco version

I'd originally experienced this on v8.5.2, but have also been able to reproduce on v8.6.1.

Reproduction
------------

1. Install Umbraco 8.6.1 with starter kit
2. Update textarea datatype to have a character limit of 300
3. Go to Home node and update description field with something like "      foo"
4. Character limit does not update when space characters are added, only when a non-space character is the last character in the input

### Bug summary

Having looked at the source, it seems like the textarea input is missing an `ng-trim="false"` attribute. This issue doesn't look to apply to the textstring property as it already has an `ng-trim="false"` attribute. I'll try and submit a PR this afternoon. 🤞

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8083

PR #8083, opened 2020-05-08
Adding ng-trim="false" attribute to textarea property editor

alanwthom:
### Prerequisites

This PR addresses issue #8082 

### Description
Adds an `ng-trim="false"` attribute to the textarea property editor so that character count functionality updates when the last character of the input is a space. This attribute is present on the textstring property editor which works as expected. 

This can be tested by creating a textarea property with a character limit and entering a number of spaces once the character count appears and observing that it updates with each new character entered.

Newbie PR here, so please shout if I've done something wrong or put something in the wrong place! 👍


nul800sebastiaan:
Aha, good catch @alanwthom, thanks for that! 👍 

Congrats on your first PR to Umbraco-CMS, that is perfect the way it is now, so not to worry, you didn't do anything wrong!

I've added a shiny new contributor badge on your Our Umbraco profile as well: https://our.umbraco.com/members/id:152642/ 🎖️ 

Thanks again!

Issue #8081, opened 2020-05-08
V8: Nested Content validation message does not translate property names


joepvtl:
We use a lot of translations for property names and we came across this issue that the validation messages of Nested Content doesn't translate the property names.

## Umbraco version

I have tested this on 8.6.1

Reproduction
------------

### Steps to reproduce

1. Setup a clean install of Umbraco 8.6.1 and use the default starter kit.
2. Create a new doctype as an element and add a **mandatory** TextString property to it and name the property '#p_NameTest'.
3. Create a new Nested Content datatype and add the new doctype to it.
4. Add a new property with the new datatype to the homepage.
5. Create a dictionary item called 'p_NameTest' and add a translation to it.
6. Add an item to the Nested Content and don't fill in the value.
7. Press 'Save and Publish' and you will see the validation message popup.

### Expected result

A validation message with a translated property name.
![image](https://user-images.githubusercontent.com/42842226/81403220-34000e80-9133-11ea-9f02-434d96c70227.png)

### Actual result
A validation message without a translated property name
![image](https://user-images.githubusercontent.com/42842226/81403338-76c1e680-9133-11ea-9075-9bbe0c4d078d.png)

nul800sebastiaan:
Sure thing @joepvtl, would be good to have that fixed.

umbrabot:
Hi @joepvtl,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

umbrabot:
Hiya @joepvtl,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

[...]

PR #8092, opened 2020-05-11
Added translation of property name for nested content property editors.

cheeseytoastie:
Mandatory messages are not changed and the text "Cannot be empty" remains untranslated which seems wrong? But this fixes the issue reported by 8081

### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8081

### Description
The issue described in 8081 needed the LocalizedTextService passed in through to the mapper so we can map the "label" of the property name in the same way it works for the vanilla property editor.  

I didn't attempt to translate ".... cannot be empty" though it seems like this would be a good addition to the Umbraco dictionary. 

To test:
1) Create two users, EN / DK
2) Create a Nested Content Doc type with a mandatory field name as per the instructions in the issue (namely named #p_NameTest and create corresponding dictionary item "p_NameTest" with EN and DK translations. 
3) Attempt to save and publish a new NC item without filling in the mandatory field. 
4) Check that the warning has the translated field name instead of #p_NameTest rendered.

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
@@ -290,7 +290,7 @@ namespace Umbraco.Web.PropertyEditors
                     {
                         foreach (var result in validator.Validate(row.JsonRowValue[row.PropKey], propertyEditor.GetValueEditor().ValueType, config))
                         {
-                            result.ErrorMessage = "Item " + (row.RowIndex + 1) + " '" + row.PropType.Name + "' " + result.ErrorMessage;
+                            result.ErrorMessage = "Item " + (row.RowIndex + 1) + " '" + _textService.UmbracoDictionaryTranslate(row.PropType.Name) + "' " + result.ErrorMessage;
                             validationResults.Add(result);
                         }
                     }
@@ -301,7 +301,7 @@ namespace Umbraco.Web.PropertyEditors
                         if (row.JsonRowValue[row.PropKey] == null)
                         {
                             var message = string.IsNullOrWhiteSpace(row.PropType.MandatoryMessage)
-                                                      ? $"'{row.PropType.Name}' cannot be null"
+                                                      ? $"'{_textService.UmbracoDictionaryTranslate(row.PropType.Name)}' cannot be null"
                                                       : row.PropType.MandatoryMessage;
                             validationResults.Add(new ValidationResult($"Item {(row.RowIndex + 1)}: {message}", new[] { row.PropKey }));
                         }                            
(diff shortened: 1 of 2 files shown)


cheeseytoastie:
Oops I've broke a test - will fix tomorrow

joepvtl:
Hi,

Whilst you are at it, shouldn't you also add the translations to these two lines aswell?
https://github.com/umbraco/Umbraco-CMS/blob/b2a9e17c90ab5c9d2bfa7c7b6d2e738a152073e0/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs#L287
https://github.com/umbraco/Umbraco-CMS/blob/b2a9e17c90ab5c9d2bfa7c7b6d2e738a152073e0/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs#L298
-Joep

cheeseytoastie:
> Hi,
> 
> Whilst you are at it, shouldn't you also add the translations to these two lines aswell?
> https://github.com/umbraco/Umbraco-CMS/blob/b2a9e17c90ab5c9d2bfa7c7b6d2e738a152073e0/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs#L287
> 
> 
> https://github.com/umbraco/Umbraco-CMS/blob/b2a9e17c90ab5c9d2bfa7c7b6d2e738a152073e0/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs#L298
> 
> 
> -Joep

Good spot - added those

src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
```diff
@@ -31,14 +31,16 @@ public class NestedContentPropertyEditor : DataEditor
         private readonly Lazy<PropertyEditorCollection> _propertyEditors;
         private readonly IDataTypeService _dataTypeService;
         private readonly IContentTypeService _contentTypeService;
+        private readonly ILocalizedTextService _textService;
         internal const string ContentTypeAliasPropertyKey = "ncContentTypeAlias";
 
-        public NestedContentPropertyEditor(ILogger logger, Lazy<PropertyEditorCollection> propertyEditors, IDataTypeService dataTypeService, IContentTypeService contentTypeService)
+        public NestedContentPropertyEditor(ILogger logger, Lazy<PropertyEditorCollection> propertyEditors, IDataTypeService dataTypeService, IContentTypeService contentTypeService, ILocalizedTextService textService)
```
ronaldbarendse, line 37: This should be added as a new constructor and the old constructor should be obsoleted (and get the `ILocalizedTextService` instance from `Current.Services.TextService`):
```suggestion
        [Obsolete("Use constructor with all dependencies")]
        public NestedContentPropertyEditor(ILogger logger, Lazy<PropertyEditorCollection> propertyEditors, IDataTypeService dataTypeService, IContentTypeService contentTypeService)
            : this(logger, propertyEditors, dataTypeService, contentTypeService, Current.Services.TextService)
        { }
        
        public NestedContentPropertyEditor(ILogger logger, Lazy<PropertyEditorCollection> propertyEditors, IDataTypeService dataTypeService, IContentTypeService contentTypeService, ILocalizedTextService textService)
```

PR #8094, opened 2020-05-12
Don't modify global MVC options (and more)

Shazwazza:
This is the first of what will be a few PRs but want to keep these smaller where possible, here's the gist:

## New "Application Model" conventions

This is how we will customize only our controllers and our MVC implementation and it's called MVCs "Application Model". I just wrote a blog post about this here https://shazwazza.com/post/custom-body-model-binding-per-controller-in-asp-net-core/ which is essentially how we are doing the below. I've also logged an issue with aspnetcore https://github.com/dotnet/aspnetcore/issues/21724 though for our purposes moving forward using the "application model" is a nicer way anyways... just took a while to figure out.

* New IApplicationModelProvider which mimics MVC's own application model that powers their [ApiController] attribute. Their conventions are a little too opinionated for our needs so we've made our own and we now have an [UmbracoApi] attribute. This application model applies most of the same conventions as [ApiController] but excludes Api visibility and attribute routing requirements. And then this adds a custom convention which will always use newtonsoft json as the IInputFormatter for binding the request body as json and this will only be applied to actions within controllers (or from base controllers) marked with [UmbracoApi]. Moving forward I'm sure we'll add more conventions (i.e. authorization policies, filters, and more)
* Removes `[FromBody]` from `PostPerformInstall` since it's not needed. This is because this convention automatically makes all complex models as parameters implicitly bound from the request body. This makes migrating our api controllers waaaay easier.
* Removes `[HttpPost]` from `PostPerformInstall` since it's not needed. Our `Post`, `Get`, etc... prefixes DO work with .net core 3 so we don't need to explicitly attribute methods. [FromBody] is no longer required (see next few points)

## Cleanup stuff

* "Api controllers" should inherit from `ControllerBase` not `Controller`, the `Controller` class provides UI based methods and api controllers don't have that, so have adjusted our base class
* `AddUmbracoWebsite` -> `AddUmbracoWebComponents` since this ext method will configure web based things
* Changed `UseUmbracoRequestLifetime` to `UseUmbracoRouting` because this will become a more encompassing thing needed for being able to use routing/requests with Umbraco in general
* new `UmbracoWebServiceCollectionExtensions` which site alongside `UmbracoCoreServiceCollectionExtensions` which provides service extensions for web things and moves web based ext methods there. ImageSharp init moved to here instead of being in the Website project.
* Adds stub code in `UmbracoMvcConfigureOptions` where we could modify global MVC options using the `IOptions` pattern - but currently we don't need to (yet)
* Fixes InstallController.Index not being async but calling into async code
* Don't process logic in some middlewares if we are just dealing with client side requests
* Ensure an UmbracoContext is created on each (non client side) request and that it's disposed at the end of it. 
* Create a stub of IWebSecurity - it still needs to be implemented but at least we don't get null refs anymore
* try/catch on `_umbracoRequestLifetimeManager.InitRequest` since this raises an event and we have no control over if any consumers of that event will throw. If that happened then the entire request would die and I don't think we want that? If we want that optional, then the event args will need to provide that switch and we can check for it
* Move some files `Umbraco.Web.Common.Runtime.Profiler` -> `Umbraco.Web.Common.Profiler`
* No more global MvcOptions changes in Startup


src/Umbraco.Web.Common/UmbracoContext/UmbracoContextFactory.cs
```diff
@@ -16,8 +17,6 @@ namespace Umbraco.Web
     /// </summary>
     public class UmbracoContextFactory : IUmbracoContextFactory
     {
-        private static readonly NullWriter NullWriterInstance = new NullWriter();
```
bergmania, line 19: I guess the NullWriter class can be removed now?

bergmania:
Install etc. seems to still work 💪 . I have merged.

PR #8095, opened 2020-05-12
Redirect to installer and ext method/startup cleanup

Shazwazza:
This cleans up a bunch of our ext methods during startup and cleans up Startup in general. This starts paving the way for getting the routing to work properly. 

Again, just want to keep the PRs smaller if possible so there's less chance of us overlapping things. 

* Moves current init to where the container is cross wired
* makes our middleware use the interface so it's factory middleware instead of by convention, it's actually slightly faster this way as it doesn't use reflection to create it 
* Ensures we enable the dependencies within the ext methods else we'll just end up with broken sites
* New ext methods to enable installer
* Moves ext methods and names ext method classes to make more sense
* Redirects to installer when in install or upgrade mode and adds more notes about that


src/Umbraco.Web.Website/Extensions/UmbracoWebsiteApplicationBuilderExtensions.cs
```diff
@@ -4,10 +4,11 @@
 using SixLabors.ImageSharp.Web.DependencyInjection;
 using Umbraco.Core;
 using Umbraco.Extensions;
+using Umbraco.Web.Common.Middleware;
 
-namespace Umbraco.Web.Website.AspNetCore
+namespace Umbraco.Web.Website.Extensions
```
bergmania, line 9: I guess this should be `Umbraco.Extensions`

PR #8099, opened 2020-05-13
Netcore/feature/auto routing

Shazwazza:
* streamlines routing, makes it consistent 
* fix catch all routing for install redirect
* moves authorizeattribute and preview controller
* auto-routes back office api controllers
* fixing up some naming 
* adds tests for all the routes
* removes code from WebInitialComposer of things that have been migrated to make it easier for us to know since we can't execute the .net framework app right now anyways
* lots of notes

## Testing

* Ensure unit tests pass
* Test that the whole installation process works
* Test that you can go to /umbraco when it's installed and it will render a test view

we of course can't test the preview controller and some other things


src/Umbraco.Web.UI.NetCore/Controllers/TestBackOfficeUmbracoApiController.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Umbraco.Web.Common.Attributes;
+using Umbraco.Web.Common.Controllers;
+
+namespace Umbraco.Web.UI.NetCore.Controllers
+{
+    [PluginController("Test")]
+    [IsBackOffice]
+    public class TestBackOfficeUmbracoApiController : UmbracoApiController
```
bergmania, line 13: I guess this not should have been committed

src/Umbraco.Web.Common/Extensions/LinkGeneratorExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        /// <summary>
+        /// Returns the URL for the installer
+        /// </summary>
+        /// <param name="linkGenerator"></param>
+        /// <returns></returns>
+        public static string GetInstallerUrl(this LinkGenerator linkGenerator)
+        {
+            return linkGenerator.GetPathByAction("Index", ControllerExtensions.GetControllerName<InstallController>(), new { area = Constants.Web.Mvc.InstallArea });
```
bergmania, line 40:
```suggestion
            return linkGenerator.GetPathByAction(nameof(InstallController.Index), ControllerExtensions.GetControllerName<InstallController>(), new { area = Constants.Web.Mvc.InstallArea });
```

src/Umbraco.Web.Common/Extensions/LinkGeneratorExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            Type backOfficeControllerType;
+            try
+            {
+                backOfficeControllerType = Assembly.Load("Umbraco.Web.BackOffice")?.GetType("Umbraco.Web.BackOffice.Controllers.BackOfficeController");
+                if (backOfficeControllerType == null) return "/"; // this would indicate that the installer is installed without the back office
+            }
+            catch (Exception)
+            {
+                return "/"; // this would indicate that the installer is installed without the back office
```
bergmania, line 27: In theory this should be the virtual path, right?

src/Umbraco.Web.Common/Extensions/LinkGeneratorExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+        public static string GetBackOfficeUrl(this LinkGenerator linkGenerator)
+        {
+
+            Type backOfficeControllerType;
+            try
+            {
+                backOfficeControllerType = Assembly.Load("Umbraco.Web.BackOffice")?.GetType("Umbraco.Web.BackOffice.Controllers.BackOfficeController");
+                if (backOfficeControllerType == null) return "/"; // this would indicate that the installer is installed without the back office
+            }
+            catch (Exception)
```
bergmania, line 25:
```suggestion
            catch
```

src/Umbraco.Web.BackOffice/Filters/UmbracoAuthorizeFilter.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// <param name="umbracoContext"></param>
+        /// <param name="runtimeState"></param>
+        /// /// <param name="linkGenerator"></param>
+        /// <param name="redirectUrl">If specified will redirect to this URL if not authorized</param>
+        public UmbracoAuthorizeFilter(
+            IUmbracoContextAccessor umbracoContext, IRuntimeState runtimeState, LinkGenerator linkGenerator,
+            string redirectUrl)
+            : this(umbracoContext, runtimeState, linkGenerator)
+        {
+            _redirectUrl = redirectUrl;
```
bergmania, line 71: I'm not a fan of multiple constructors setting different fields.. I think we should make a private construtor handling all fields, and use that for each of them..

So something like this:

```
   privateUmbracoAuthorizeFilter(
            IUmbracoContextAccessor umbracoContext, IRuntimeState runtimeState, LinkGenerator linkGenerator,
 bool redirectToUmbracoLogin, string redirectUrl)
        {
            _umbracoContext = umbracoContext ?? throw new ArgumentNullException(nameof(umbracoContext));
            _runtimeState = runtimeState ?? throw new ArgumentNullException(nameof(runtimeState));
            _linkGenerator = linkGenerator ?? throw new ArgumentNullException(nameof(linkGenerator));
            _redirectToUmbracoLogin = redirectToUmbracoLogin;
            _redirectUrl = redirectUrl;
        }
```

Having multiple constructors setting fields are doomed to go wrong at some time

bergmania:
I guess the pattern `umbraco/BackOffice/{action}/{id?}` is wrong?

In v8, the requests is to `/umbraco/application` which is routed to `BackofficeController` and action `Application`, but now it seems to be required the request is to `/umbrco/backoffice/application`

src/Umbraco.Web.UI.NetCore/Controllers/TestBackOfficeUmbracoApiController.cs
Shazwazza: I left it there just for testing but will remove, if you want to test you can add your own.

src/Umbraco.Web.Common/Extensions/LinkGeneratorExtensions.cs
Shazwazza: yep, will fix to use `return hostingEnvironment.ApplicationVirtualPath;`

Shazwazza:
@bergmania I cannot reproduce.

Also the `AssertMinimalBackOfficeRoutes` test asserts that this pattern is correct for the BackOfficeController: `Assert.AreEqual($"umbraco/{{action}}/{{id?}}", endpoint1.RoutePattern.RawText);`

If I run the project and go to /umbraco it renders the test view, if I go to /umbraco/application it renders the JS output.

The pattern `umbraco/BackOffice` is for the authentication controller and potentially other non-autorouted controllers. There's only one exception to the /umbraco/backoffice route and that is the BackOfficeController which is routed correctly.

PR #8102, opened 2020-05-13
Remove "double tabbing" in the datatype picker

kjac:
### Prerequisites

### Description
The (very nicely) revamped data type picker suffers from "double tabbing" when you tab your way through the available datatypes:

double-tab-datatype-picker-before

And if you happen to hit enter on the innermost focused element of a datatype, you'll get duplicate "Select configuration" dialogs:

image

This PR shuffles some click handlers around, ensuring that only the innermost elements (the `a` tags) can receive keyboard focus. When applied it's much easier to tab your way through the datatype picker:

double-tab-datatype-picker-after


nielslyngsoe:
@kjac perfect! Thanks for your awareness :-)

PR #8103, opened 2020-05-13
Enable navigating datatype configuration picker by keyboard

kjac:
### Prerequisites

### Description

The datatype configuration picker has a few minor issues:

1. When the datatype configuration picker opens, it does not apply focus to itself. Thus if you use tab to navigate through the available datatype configurations, you're tabbing around behind the overlay. 

2. If you manually set focus in the dialog, it suffers from the same "double tabbing" issues as the datatype picker (see #8102).

3. Lastly the create button hover title is "Create a new configuration of" - it's missing the name of the data type editor.

double-tab-datatype-configuration-picker-before

This PR ensures that the dialog captures focus on load, and fixes the other small issues as well:

double-tab-datatype-configuration-picker-after


nielslyngsoe:
@kjac Again thanks for being aware. I see that we should also get around the mark-up here to replace a-tags with buttons, but that's another PR. :-)

kjac:
@nielslyngsoe yeah I did try replacing `a` with `button` as it would indeed be more correct, but the styling went haywire so I figured the same thing.... new PR 😆

PR #8104, opened 2020-05-13
Reload property preview when changing property type

### LINKED ISSUE #8068: DocumentTypeEditor does not update preview of property-editor when changed
When changing a properties data-type, the preview of the property-editor is not updated to reflect the new property-editor.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8068

### Description

When changing a property type on an existing property, the property preview doesn't reload in the content type builder (like it does in V7):

groups-builder-preview-reload-before

This PR ensures that we watch for changes and update the property preview accordingly:

groups-builder-preview-reload-after


nul800sebastiaan:
Works like a charm, thank you kindly @kjac! 👍

PR #8105, opened 2020-05-13
Fix "Edit: undefined" in content editor title

kjac:
### Prerequisites

### Description

The content editor title seems broken; it reads "Edit: undefined" when editing an existing content item:

image

This happens for both culture variant and invariant content. But still works fine when creating a new content item:

image

With this PR, the title goes back to "Edit: [name of content item]" like it used to be:

content-editor-title-undefined-name


nul800sebastiaan:
Great! I just saw this one as well! Thanks @kjac! 👍

PR #8107, opened 2020-05-13
Prevent property previews from having focus

kjac:
### Prerequisites

### Description

When tabbing through the content type editor, inputs/buttons/etc. within the property previews can receive focus:

property-preview-catches-focus-before

You can't actually interact with the focused elements within the property previews. If you hit enter with one of them is in a focused state, the "edit property" dialog opens up (as you'd expect it to), so the applied focus within the preview doesn't break things as such. It just feels weird when navigating. 

Anyway. This PR fixes it:

property-preview-catches-focus-after


nul800sebastiaan:
Such a simple little fix, such an improvement, lovely, thanks @kjac!

Issue #5533, opened 2019-05-28
Cannot define own RoleProvider for public access


Appstract-fjellvang:
In Umbraco 8 it doesnt seem to be possible to add your own RoleProvider any longer - therefor you cannot roll with an external data source for your member groups.

Reproduction
------------

Create a class that implements RoleProvider. Define "GetAllRoles" so that it returns an array of strings of test user groups.

Change the reference in the Web.config as done in umbraco 7.

### Bug summary

The custom defined groups doesn't show up in during "Role based protected" nor in the Member group dropdown of the Member dashboard

### Specifics

Found this code in the source:
https://github.com/umbraco/Umbraco-CMS/blob/5a07e34fb4756004bc7175f28a96332bbe84b7ae/src/Umbraco.Web/Editors/MemberGroupController.cs#L66

Seems like when your're using a custom membership provider, it just returns an empty array.

### Expected result

I expect a list of custom membership groups

### Actual result

Empty list

nul800sebastiaan:
Hi @Appstract-fjellvang - the code you reference has been like this since v7.5.0, released in August 2016. I can't remember exactly why it changed but I am sure it was for good reason. 

I'll assign this to @Shazwazza to see what he says when he comes back from his holidays, maybe @AndyButland can chime in here as well?

AndyButland:
I did some of the early work on user groups back around then, but nothing on member groups as I recall.  So don't think I've got any insight here.

Appstract-fjellvang:
@AndyButland So you're saying the MemberGroupController has been like that since 7.5 ? Because i didn't succeed in finding any equivalent in the umbraco 7.14 release, and custom member groups are working as expected in our umbraco 7 solutions

Appstract-fjellvang:
I figured it out.

Seems like you don't use the RoleProvider any longer.

I made it work by implementing my own MemberGroupService, then RegisterUnique in an IUserComposer
and finally made my membership provider implement UmbracoMembershipProviderBase instead of the old MembershipProvider

Shazwazza:
Replacing the MemberGroupService isn't what we want to do here, i'll re-open.

Shazwazza:
In this case for role based public access it should be querying the role provider but will need to investigate if there's a reason this is not done.

Appstract-fjellvang:
Yeah that does make more sense, but i didn't investigate further why this didnt happen

Appstract-fjellvang:
[...]

PR #8114, opened 2020-05-14
Proof of concept Public Access with Custom Role Provider

enkelmedia:
### Prerequisites

The issues is well described in here: https://github.com/umbraco/Umbraco-CMS/issues/5533

### Description
This is a initial "proof of concept" for supporting Public Access with a Custom Role Provider. It's a draft and there will be refactoring and documentation needs.

Short summery of changes:

* Introduced a non breaking extension to editorService.itemPicker, it now supports selecting multiple items and displaying the picker as a "tree-view", defaults to backwards compatible settings.
* Introduced a new model "RoleDisplay" that can represent both a MemberGroup and an external Role from a Role Provider. The "Id" is of type string since this is what's stored in the PublicAccessRule and this is needed since the Custom Role Provider only supports string roles.
* Updated ContentController.GetPublicAccess (and its model) to return all groups in the first load to avoid having to fetch them later on. This avoids changes in JS-services.

### Setting up Public Access with Umbraco Member Group:
umb role member group

### And with a Custom Role Provider:
umb role custom

I wanted to initiate the PR to get guidance and opinions, is this the right direction? 

### Improvements
I do know that there is a need for refactoring, ie property names on the PublicAccessModel returned by ContronController should probably be "Roles", "AllRoles" and not "Groups".

More than this I see some things:
* Model-mappings needs refactoring and to be moved into mappers.
* We could skip the contentService.itemPicker and just make it possible to select/toggle groups inside the dialog. But if there is a lot of roles then this might not be a very nice experience.
image
* We should probably exclude roles that is already selected from the list of selectable roles in the dialog if we need the itemPicker.
* Would also be nice to provide a extension point for injecting/modifying RoleDisplay-models when running a custom RoleProvider - since the GetAllRoles()-method on the role provider-abstraction only expose a string[] this would allow a custom RoleProvider to display friendly names for the Roles as well, or maybe we could use localisations in a smart way to provide friendly names?

Let me know what you think about this and if I should take if further.

Cheers!


KeithBoynton:
I'm encouraged to see this, it's a showstopper for me at the moment

enkelmedia:
Not sure if I can do anything to get feedback on this one?

nul800sebastiaan:
@enkelmedia Thanks! You will have noticed I'm a bit behind.. 🙈 

This looks pretty good and quite easy to me. The only problem I have is I don't really know what the property `hasProtection` means, maybe you can add a code comment to clarify so that future reader of the code know too?

Other than that, you know it needs cleaning up so please go forward with that and then we can review and test it! 👍 

@KeithBoynton I have no timeline for when we can get this merged, so if this is a showstopper then you need to look for alternative ways to protect pages for example by doing checks in your controllers for those pages.

enkelmedia:
@nul800sebastiaan I've made some changes now, but som questions

I've updated the UI and variable names so it feels better, I actually decided to keep the mappings in the controller method as they are only used there. Let me know if this is bad.

1. Merge...? =/
I wanted to get the branch up to date with v8/contrib so i merged in v8/contrib into my branch... so now there's a ugly merge-commit in the branch... maybe I should have rebased? Or what is the best way to do this?

2. Interface for friendy names
To address the fact that the build in RoleProvider in .NET only exposes string[] I added a interface that can be implemented with the RoleProvider to facilitate for friendly names.

So the code checks if this interface is implemented and uses that if it can, this way an external RoleProvider can still provide friendly names for roles if needed.

```csharp
 public class MarkusRoleProvider : RoleProvider, IRoleProviderWithDisplayNames
{
  public RoleDisplay[] GetAllRolesWithDisplayName()
  {
    ....
  }
}
```

So the clean up should be in a quite good state, I need to remove the "MarkusRoleProvider.cs"-file as well, managed to commit that... =/ But maybe it's good to keep for testing and remove it before it's merged.

KeithBoynton:
@nul800sebastiaan Yes, it's fine I understand that, I'm doing my own manual checks in the controllers etc. for now but the more it goes on the messier the workaround becomes. I commented to raise the visibility and show it's impacting more than just enkelmedia. Thanks for responding, it's great that it's got visibility.

@enkelmedia Great work! Thank you for raising this.

KeithBoynton:
@nul800sebastiaan @enkelmedia What needs to happen on this to progress it, just wondering if there anything I can do to help

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/itempicker/itempicker.controller.js
```diff
@@ -12,24 +12,57 @@ function ItemPickerController($scope, localizationService) {
...
             });
         }
+
+        if (vm.multiPicker) {
+            vm.selectedItems = [];
+
+            //if (Utilities.isArray($scope.model.selectedItems)) {
+            //    vm.selectedItems = $scope.model.selectedItems;
+            //}
+            //TODO: Push selected items from model
```
nul800sebastiaan, line 36: Can you explain what the idea is here? And is this a new feature, should we worry about this later maybe?

nul800sebastiaan (changes requested):
@enkelmedia 

Thanks for the cleanup, I did another merge and some more cleaning up! I removed your MarkusRoleProvider, for future tests I would recommend dropping something in the App_Code folder and documenting it on the PR. 

I've left you a question on the itempicker controller!

Yes, adding ugly merge commits is absolutely fine, in the end we'll just squash this and pretend they never happened!

>  I actually decided to keep the mappings in the controller method as they are only used there. Let me know if this is bad.

I'll need to check with @Shazwazza 

> 2. Interface for friendly role names
> What do we think about this approach?

Also will need to check with @Shazwazza

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/itempicker/itempicker.controller.js
enkelmedia: If i recall this right I was thinking that it could be good to be able to initialize the picker with selected items, but it’s not needed for this so I drop that - forgot to remove the commented out code.

enkelmedia:
This one really died? I’m still running this code in a “custom build” for one client. Far from ideal. 

Since this is for V8 I’m not sure if this will ever be included? 

It’s just sad, spending a lot of time and effort on something that just ends up waiting for feedback and answers that never comes.

KeithBoynton:
Yes I wonder, has it been fixed in v9? Or can this be transferred for v9 and merged in?

enkelmedia:
@KeithBoynton I think V9 will be a totally different story as the underlying "RoleProvider" that this solution is using does not exist in .NET5 - I haven't looked closer but I guess/think that it might be easier with ASP.NET Core Identity.

KeithBoynton:
Thanks for that info @enkelmedia that's interesting to know, that explains why you think it might not get the attention now

PR #8119, opened 2020-05-16
Fix label alignment in umb-checkbox

BatJan:
### Prerequisites

### Description
I noticed the label text for the checkbox was not vertically centered - This is adjusted with this small PR.

**Before**
checkbox-label-before

**After**
checkbox-label-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-form-check.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-form-check.less
@@ -18,6 +18,8 @@ label.umb-form-check--checkbox{
     }
     .umb-form-check__info {
         margin-left:20px;
+        position: relative;
+        top: 3px;
     }
     
 


nul800sebastiaan:
Subtle but great! 👍

bjarnef:
It seems this is causing other side effects, e.g. for radiobuttons on relation types.

image

image

BatJan:
Good catch @bjarnef - It appears another PR is needed for getting to the root of the underlying issue. Even after removal of the top: 3px it's still a bit off. Will see if I can find some time to look deeper into it 👍

BatJan:
@bjarnef There appears to be more issues due to some selectors taking over specificity of the `label`, which is causing some of the issues at least. Also some weirdness around margins that are only set for the checkbox and not the radio button - Hope to submit a PR after this evenings a11y meeting fixing all the instances 😃

PR #8120, opened 2020-05-16
Add disable-dirty-check option

BatJan:
### Prerequisites

### Description
I have extended the `<umb-checkbox>` directive with a `disable-dirty-check` parameter since I have recently worked on a custom content app where I reused the directive and we wanted to be able to disable the dirty check. But it's not possible to do currently out of the box. I'm reusing the already build in `no-dirty-check` directive if the `disable-dirty-check` is set to true, which will help out others too if they need an easy way to prevent the model from becoming dirty and therefore trigger the "unsaved data overlay" in scenarios where it does not make sense 😃 

I used the "checkboxlist" datatype for demo purposes only and it's not part of this PR.

**Default directive behavior**
disable-dirty-check-false-default

**Directive behavior when `disable-cdirty-check="true"` is set**
disable-dirty-check-true


nul800sebastiaan:
Great! Thanks @BatJan!

PR #8121, opened 2020-05-16
Add backdrop when menu items are opened

BatJan:
### Prerequisites

### Description
In this PR a backdrop is added whenever we open a menu - I have done this to align the UI behavior throughout the entire backoffice. Should this PR end up being accepted I will do a follow up PR where I refactor the "closeBackdrop" function I have added in 3 files so the close method of the backdrop service can handle the removal of the class based on a parameter instead so it's handled by the service instead of the places where it's called.

I hope you like it 😃 

~~**EDIT**: Ok, so I appear to have missed some spots where the backdrop does is not removed using the context menu - I'm looking into and will update this PR when it's ready for review again 👍~~

**EDIT:** Ok, this is good to go again! 😄 - Ensure to test it thourougly though - I think I covered all the edge cases now and I have tested that it acts like you would expect when interacting with the various context menu dialogs, methods etc. 👍  

**Before**
without-backdrop

**After**
with-backdrop


nul800sebastiaan:
This worked great in my testing as well @BatJan and is a really useful addition, well done, keep up the great work! 👍

PR #8124, opened 2020-05-18
boot failed exception handling

Shazwazza:
Ensures we deal with boot failed exceptions like we used to and show the exception instead of silently failing. Also ports over a couple other things from the umbraco module.

* Register our middleware services in a composer and make them a singleton since they can be (no need for transient)
* Use `BootFailedMiddleware` when boot fails
* UmbracoRequestMiddleware  properly handles exceptions and ports logic from umb module to dispose any IDisposeOnRequestEnd
* removes code from the umb module that has been ported

to test, you can just add an exception in code into one of the composers this will result in a boot failed and the exception will be thrown/shown.


bergmania:
Code looks good. 
I think we need to handle the issue, when we cannot boot due to connection isssues to the DB.

In this case it still throws an ugly error:
image

The reason is because the AspNetCoreComposer has not been executed.

Shazwazza:
@bergmania good catch, have pushed a fix for that.

bergmania:
Nice work, works as expected 💪

PR #8129, opened 2020-05-18
Added ASP.NET Core Identity to Umbraco.Web.UI.NetCore

scottbrady91:
### Prerequisites

### Description
Addition of ASP.NET Core Identity to Umbraco.Web.UI.NetCore as part of Rock Solid Knowledge's participation in the Unicore project.

I have not added any cookie auth for now. You could either add your own authentication handler or introduce the ASP.NET Core Identity default cookies and use the `SignInManager`. I'm preferring the idea of Umbraco managing its own, but I thought I would leave the decision to you.

#### Changes
- Extracted re-usable ASP.NET Identity code from previous PR into `Umbraco.Infrastructure`
  - `Umbraco.Infrastructure` used due to existing back office code using `Umbraco.Core.Models.Membership.User`. This will take better Umbraco knowledge than I have to move into `Umbraco.Core`.
- ASP.NET Core Identity registered using `AddIdentityCore` to prevent the use of its default cookies
- Re-added code in `NewInstallStep` to create user

#### Notes & Future TODOs
- ASP.NET Core host uses the default ASP.NET Core Identity password hasher due to `Umbraco.Core.Security.ConfiguredPasswordValidator` having a dependency on ASP.NET Identity 2. It looks like this configuration style needs unpicking to use the ASP.NET Core options functionality
- `UmbracoMapper` currently injected by the back office. I assume you'll update this as part of another ticket
- Internal methods on `BackOfficeIdentityUser` and `BackOfficeUserManager` made public due to migration from `Umbraco.Web`
- Thread.CurrentPrincipal is no longer supported in ASP.NET Core: https://docs.microsoft.com/en-us/aspnet/core/migration/claimsprincipal-current?view=aspnetcore-3.1. `IdentityAuditEventArgs.GetCurrentRequestBackofficeUserId` needs refactoring to no longer use this.


src/Umbraco.Infrastructure/BackOffice/Extensions/ClaimsPrincipalExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            //Otherwise convert to a UmbracoBackOfficeIdentity if it's auth'd and has the back office session
+            if (user.Identity is ClaimsIdentity claimsIdentity && claimsIdentity.IsAuthenticated && claimsIdentity.HasClaim(x => x.Type == Constants.Security.SessionIdClaimType))
+            {
+                try
+                {
+                    return UmbracoBackOfficeIdentity.FromClaimsIdentity(claimsIdentity);
+                }
+                catch (InvalidOperationException)
+                {
```
emmagarland, line 37: Do we want to record/do something in this instance if an InvalidOperationException occurs?

src/Umbraco.Infrastructure/BackOffice/IdentityAuditEventArgs.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        /// <summary>
+        /// Returns the current logged in backoffice user's Id logging if there is one
+        /// </summary>
+        /// <returns></returns>
+        protected int GetCurrentRequestBackofficeUserId()
+        {
+            var userId = -1;
+            /*var backOfficeIdentity = Thread.CurrentPrincipal.GetUmbracoIdentity();
```
emmagarland, line 111: Is this one of the TODOs?

src/Umbraco.Infrastructure/Install/InstallSteps/NewInstallStep.cs
```diff
@@ -60,25 +62,21 @@ public override async Task<InstallSetupResult> ExecuteAsync(UserModel user)
             _userService.Save(admin);
 
             //TODO: This needs to be reintroduced, when users are compatible with ASP.NET Core Identity.
```
Shazwazza, line 64: we can remove the TODO

src/Umbraco.Infrastructure/BackOffice/Extensions/ClaimsPrincipalExtensions.cs
Shazwazza: This code existed before - was just moved here. I'll make a note about looking into this and where/why it's there and at least add notes

src/Umbraco.Web.BackOffice/Extensions/UmbracoBackOfficeServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            services.AddDataProtection();
+
+            // UmbracoMapper - hack?
+            services.TryAddSingleton<IdentityMapDefinition>();
+            services.TryAddSingleton(s => new MapDefinitionCollection(new[] {s.GetService<IdentityMapDefinition>()}));
+            services.TryAddSingleton<UmbracoMapper>();
+
+            services.TryAddScoped<IIpResolver, AspNetCoreIpResolver>();
+
+            services.AddIdentityCore<BackOfficeIdentityUser>();
```
Shazwazza, line 27: Is this duplication intended?

src/Umbraco.Infrastructure/BackOffice/IdentityAuditEventArgs.cs
Shazwazza: Uncommenting this code still builds so I'm wondering if it was just part of the refactoring process?

scottbrady91: Sorry, I missed this one when I reviewed my changes. It had to be removed since Thread.CurrentPrincipal is no longer supported in ASP.NET Core: https://docs.microsoft.com/en-us/aspnet/core/migration/claimsprincipal-current?view=aspnetcore-3.1. Again, this requires more Umbraco knowledge than I have to refactor this. I've added it to the notes in the PR write up.

Since both ASP.NET and ASP.NET Core are using this method, you could multi-target by wrapping it in an ifdef for now.

src/Umbraco.Web.BackOffice/Extensions/UmbracoBackOfficeServiceCollectionExtensions.cs
scottbrady91: No, sorry

src/Umbraco.Infrastructure/Install/InstallSteps/NewInstallStep.cs
scottbrady91: Of course :p

scottbrady91:
> With the normal cookie auth we would have to "adjust" it in our own way anyways just like we do in v8. There's quite a few things to consider since we need to cater for front-end and back-end auth separation along with dealing with preview authentication, etc... My intention was to do a similar thing as we had in v8 where we use the underlying plumbing in the cookie authentication code but it exists side by side of the default cookie auth. But am interested in what you mean by 'managing it's own'?

When you use `AddIdentity` & `SignInManager`, I feel like they hijack your auth system a little. They set their own cookies as the default auth schemes, and I don't believe you can change the auth scheme names (the `SignInManager` is hardcoded to use `IdentityConstants.ApplicationScheme`).

By managing your own cookies, I mean just calling `AddAuthentication().AddCookie()` yourself. This way you get full control over auth schemes and don't have to deal with an opinionated/restricted `SignInManager`. I think this is the safer bet for having front-end and back-end auth live side by side.

Shazwazza:
HI @scottbrady91 thanks for the info and yep that makes sense so i think we're on the same page about the cookie auth. We'll need to have our own implementation one way or another and if SignInManager is in-flexible in that regard we'll use our own service. Just wanted to check if you were referring to something entirely different :) 

I'll get this further reviewed today and then start on some auth/cleanup/etc... Thanks!

src/Umbraco.Infrastructure/BackOffice/IdentityAuditEventArgs.cs
Shazwazza: thanks for the link, i'll have a look at this one 👍

scottbrady91:
@Shazwazza Awesome. Feel free to reach out over slack if you want to bounce any ideas off of me 🙂

PR #8140, opened 2020-05-19
RelationService - Log to audit on RelationType Save

poornimanayar:
When a relation type is saved, it is not logged to the `umbracoLog` table. I have tried to enable logging when a relation type is saved.


nul800sebastiaan:
Thanks for this one @poornimanayar! 👍

Issue #7025, opened 2019-10-31
Umbraco 8 Macro parameter support for “Umbraco.MultiUrlPicker” editor


shearer3000:
The datatypes available for Umbraco 8 Macro parameters is missing the core type of “Umbraco.MultiUrlPicker” editor.

_Background
In umbraco 7, a method for achieving this was a package.manifest to define a custom datatype. This works in umbraco 8 too to an extent, but the macro in the RTE doesn’t actually save the links that are picked._

```
       "propertyEditors": [
              {
                     "alias": "MyCustom.MultiUrlPicker",
                     "name": "Macro - Multi Url Picker",
                     "isParameterEditor": true,
                     "editor": {
                           "view": "~/umbraco/Views/propertyeditors/multiurlpicker/multiurlpicker.html"
                     }
              }
```

Thanks
Andrew

ref: original forum topic
https://our.umbraco.com/forum/templates-partial-views-and-macros//98546-macro-parameter-editor-in-umbraco-8

shearer3000:
hopefully this could be considered for 8.5 :)

mzajkowski:
#7534

marcemarc:
Can confirm this is still the case in V8.6.1
Essentially the use case for this is having a Macro that can insert links as 'buttons' in the grid or a rich text area.
'you can ask the editor to insert a link and apply a tinymce 'format' or use a tinymce 'template' to achieve 'a version of this' - but usually there is more complicated markup around the buttons or you want to programmatically apply some logic, eg first link picked is the 'primary' button the second and third picked items are 'secondary buttons' and are styled differently.

In V7 you could achieve this by using the defaultConfig in the package.manifest file to specify the min / max number of links - along with setting isParameterEditor flag - and the Url Picker would follow that configuration and function correctly as a macro parameter...

... the same trick in V8, doesn't work, defaultConfig isn't honored, and the whilst the picker appears and allows you to enter links and pick items... those picked items and links are saved when the macro parameter dialog is closed :-(

What I'm not sure is if this occurred when the Url Picker was moved into core in V7 - eg something in the core implementation prevents the saving of the picked items to the right scope - or whether it's to do with the way V8 is handling configuration of macro parameters... eg it worked by luck in V7! and this kind of thing is no longer supported.

marcemarc:
Ok Update on this!, I think I understand why this is occurring....
Essentially the UrlPicker has a 'renderModel' that tracks which links are currently picked in the picker...

[...]

PR #8144, opened 2020-05-20
Broadcast 'formSubmitting' event when Macro Parameter Overlay is submit

marcemarc:
Some property editors - eg MultiUrlPicker listen out for the broadcast of the 'formSubmitting' event before setting their value of $scope.model.value... if these editors are used as Macro Parameters they do not save their value. This commit adds the broadcasting of the event when the 'submit' button is called and any parameters have been set.

The MultiUrlPicker is a popular MacroParameter candidate + setting defaultConfig for min/max items, and in V7 this worked... (Possibly only in the RJP package?). 

Having added this broadcast of the 'formSubmitting' event to the closing of the Macro Parameter overlay, the MultiUrlPicker functions as you would expect in the context of a Macro Parameter!

The existing issue that this PR relates to is: https://github.com/umbraco/Umbraco-CMS/issues/7025

## How to test?

Add a new folder to your App_Plugins folder called 'MultiUrlPickerMacroParameter'
Add a package.manifest file to the folder with the following config:
```
{
    "propertyEditors": [
        {
            "alias": "MacroParameter.UrlPicker",
            "name": "Url Picker Macro Parameter",
            "isParameterEditor": true,
	    "Group": "Macro Parameters",
	    "Icon": "icon-link",
            "editor": {
                "view": "multiurlpicker",
                "valueType": "JSON"
            },
             //can set min and max or other default config for the picker
	     "defaultConfig": {
			 "minNumber": 0,
			  "maxNumber": 5
		  }
               }
	  ]
  }
```
Rebuild your site, then create a New macro - add a parameter called 'Urls' - you'll find 'Url Picker Macro Parameter' is a new option. Allow the macro to be inserted in the Rich Text Area + Grid.
Visit a page with a rich text area - and insert the macro - add some links, press submit. The links should be saved and accessible to use in your Macro partial view in a big blob of JSON!

Without the 'broadcast' event in place, you'll find the URL picker will appear as a macro parameter, but none of the picked links will be saved :-(

(this may not be the right way to do it of course!)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/macropicker/macropicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/macropicker/macropicker.controller.js
@@ -22,6 +22,7 @@ function MacroPickerController($scope, entityResource, macroResource, umbPropEdi
         if ($scope.wizardStep === "macroSelect") {
             editParams(true);
         } else {
+            $scope.$broadcast("formSubmitting", { scope: $scope });
             $scope.model.submit($scope.model);
         }
     };


nul800sebastiaan:
I love it, seems to work a treat, thanks @marcemarc! 👍

amirkhan81:
Interesting caveat with this on 8.17.0 is that it will store the value if you save the page, then save and publish, but not if you just hit save and publish...

PR #8147, opened 2020-05-22
Stabilising acceptance tests + Bugfix for hanging saves of media.

bergmania:
## Notes
- Fixed bug with parallel execution of the cache refreshers

- Fixed issue in template test that called postSave twice.

- Updated Cypress to latest + Updated Umbraco Test helpers with longer timeout on close tours modal.

- Added new acceptance test for delete templates


bergmania:
Added no-notes, as the bug with  cache refreshers was never released.

PR #8148, opened 2020-05-22
Grid: adds row immediately if only one layout is allowed

### LINKED ISSUE #8039: In a Grid Layout add row automatically when plus icon is clicked if there is only one row configuration
Let say you have configured a simple Grid Layout with only one generic row configuration. 

Now, if the editor wants to add a row, s/he clicks the plus icon and is then asked to select what row configuration to insert. In this case there is only one to select so why ask for it? Why not insert a row of the only allowed on directly?

With this feature request implemented I belive the editor flow will be a little smoother with less clicks in cases where you configure the Grid Layout somewhat similar to a Stacked Layout scenario.

skttl:
### Prerequisites

If there's an existing issue for this PR then this fixes #8039

### Description
This changes the +-button in grids, if you only have one allowed layout to choose from. If you only have one allowed layout, the row gets inserted on the first click of the + button. In addition to that, a tooltip is added with the name of the layout.

How it looks for one allowed layout:
rRdAGIsTwb

How it looks for more than one allowed layout:
OIoUuAgkmq

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.html
@@ -251,7 +251,9 @@
 
                        <button
                           class="iconBox btn-reset"
-                          ng-click="toggleAddRow()"
+                          type="button"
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Lovely, thanks for that one @skttl 👍

PR #8150, opened 2020-05-22
Add missing focus style in .buttonbackground mixin

BatJan:
### Prerequisites

### Description
While I was working on #8141 I noticed some issues with the focus styling in the user dialog overlay for instance - Turns out it was missing the :focus pseudo class, which I have now added.

**Before**
edit-focus-before

**After**
edit-focus-after


nul800sebastiaan:
Look at that, much nicer @BatJan, thank you! 👍

PR #8151, opened 2020-05-23
Allowing null parameters in factory.CreateInstance

callumbwhyte:
I've been doing a bit with `factory.CreateInstance` recently and found it wasn't possible to pass nullable arguments, for a number of reasons.

If you pass a null argument in to the `CreateInstance` extension method a `NullRefrenceException` is thrown.

If an argument is null it tries to get an instance of that parameter type from the IoC container. The `GetInstance` method throws an exception if that type is not registered. By calling the `TryGetInstance` method we have the option of returning null - this will then be handled within the class being instantiated (e.g. via a check in the constructor)

I did consider introducing a `TryCreateInstance` method with these changes but don't think this creates a breaking change as previously this would have thrown an exception.


nul800sebastiaan:
Thanks @callumbwhyte! Since we recently merged https://github.com/umbraco/Umbraco-CMS/pull/7814 this resulted in a merge conflict which I have now resolved for you, I am pretty sure we still want to do that null coalesce here on line 88 right? https://github.com/umbraco/Umbraco-CMS/pull/8151/files#diff-efb2b0ba882ec9b7d11af970c6657eddR88 Otherwise we'd surely get a null ref error again.

Let me know if the merge looks good to you too! 👍

callumbwhyte:
Thanks @nul800sebastiaan!

Yes that null coalesce is needed still, merge looks great 👍

nul800sebastiaan:
Does look great @callumbwhyte but.. I will need you to have a look at the unit tests, some are failing now 🙈 

https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=42981&view=ms.vss-test-web.build-test-results-tab&runId=1050874&resultId=100500&paneView=debug

callumbwhyte:
Doh! Indeedy.. I'll take a look and update the PR

nul800sebastiaan:
@callumbwhyte Guessing you haven't had a chance to take a look yet? 😅  (take your time, no rush, just wanted to check in on this one).

callumbwhyte:
I've got this fixed up (or almost) on my local just been a bit busy! Will revisit shortly!

PR #8152, opened 2020-05-23
Removing references to legacy Examine config

callumbwhyte:
This removes the last remaining reference to the legacy Examine config sections.

I encountered some problems while testing my package against for configuring Examine against the Umbraco source as I have a config section named the same thing 🤪


nul800sebastiaan:
Hehe very good, thanks very much @callumbwhyte ! 👍

Issue #8153, opened 2020-05-23
V8: HasValue() & IsValue() behave inconsistently for 'Umbraco.Dropdown.Flexible' property editor


marcemarc:
If you create a Property, of type 'Umbraco.Dropdown.Flexible' and provide several dropdown options, when the editor opens the content item featuring this property, and saves and publishes without choosing an item from the dropdown, the value stored in the database for the dropdown Is NULL - and HasValue & IsValue function correctly for the property. (using the implementations from PropertyValueConverterBase).

If you select an item for the Dropdown property, save and publish - the selected item is stored in the database, and HasValue & IsValue function correctly still.

However if you return to the Dropdown property and 'unselect' the selected item - saving and publishing stores '[]' or weirdly '[null]' as the value for the dropdown in the database - and now HasValue and IsValue report as 'true' even though 'no value is selected'...

Dropdowns are often used in a scenario where you want a setting to be applied to a section but overridable on the page level - see the 'Search Engine Change Frequency' example in the razor SiteMap tutorial in the documentation for an example: https://our.umbraco.com/documentation/Tutorials/Creating-an-XML-Site-Map/#2-create-xmlsitemapsettings-composition

Umbraco Version 8.6.1

I would expect HasValue and IsValue to return false, if no item is selected in the Dropdown property.

nul800sebastiaan:
Ultimately, this is a problem with how we store empty values, so we should definitely just null out the field in the database instead of storing empty js arrays or `[null]`. 

We'd love some help in getting this fixed up if you or someone coming along gets a chance to look at this!

umbrabot:
Hi @marcemarc,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

marcemarc:
@nul800sebastiaan agreed! - I've updated the PR to prevent the dropdown flexible property editor from saving [] or [null] when there are no items selected... and indeed this resolves the IsValue HasValue problem without updating the property value converter

nul800sebastiaan:
[...]

PR #8154, opened 2020-05-23
Fix HasValue and IsValue for Dropdown Property Types with no Values

marcemarc:
### Prerequisites

- [yep] I have added steps to test this contribution in the description below

Fixes: https://github.com/umbraco/Umbraco-CMS/issues/8153

### Description
An 'empty' dropdown (Umbraco.Dropdown.Flexible) property -  will store a value as a string in the database in the format '[]' or '[null]' when actually 'nothing is selected'- this is after first selecting an item in the dropdown, saving, then unselecting the selected item reverting back to the state where nothing is selected... 

this breaks IsValue and HasValue in 'PropertyValueConverterBase' because they are only determining whether something has a value on the basis of whether it is a string and isn't whitespace... so this also breaks ToAncestors() on a recursive property that uses the dropdown.

~Hopefully this PR illustrates the problem, I'm not saying this is the only way or best way to fix it... eg you could look from the property editor perspective and try to ensure that [] && [null] is never actually saved to the database, or you could move this fix into the 'PropertyValueConverterBase' if it's a common issue across other similar property editors - for now, I've added the check for [] and [null] in the specific converter for the dropdown: `FlexibleDropdownPropertyValueConverter`~

To test the solution, create a property with a dropdown with alias 'searchEngineChangeFrequency', give it several values add it to a document type, in the template for that document type write out:

`<h2>@Model.HasValue("searchEngineChangeFrequency") - @Model.Value("searchEngineChangeFrequency")</h2>`

In the backoffice 'save and publish' a page featuring the dropdown editor 
Initially have no value selected.

With this PR in place you should see
False - 
Now select a value in the list and save and publish you should see
True - daily
Now return and unselect the value in the dropdown, you should see
False - 
again... without the PR you will see 
True -

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/dropdownFlexible/dropdownFlexible.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/dropdownFlexible/dropdownFlexible.controller.js
@@ -16,12 +16,13 @@ angular.module("umbraco").controller("Umbraco.PropertyEditors.DropdownFlexibleCo
         //ensure this is a bool, old data could store zeros/ones or string versions
         $scope.model.config.multiple = Object.toBoolean($scope.model.config.multiple);
         
-    //ensure when form is saved that we don't store [] or [null] as string values in the database when no items are selected
+        //ensure when form is saved that we don't store [] or [null] as string values in the database when no items are selected
         $scope.$on("formSubmitting", function () {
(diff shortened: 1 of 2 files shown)


marcemarc:
I've updated the PR so that the fix is now to prevent the dropdown flexible property editor from saving [] or [null] into the database when not dropdown items are selected... this allows IsValue and HasValue to work as expected for a dropdown with no items selected

nul800sebastiaan:
Good stuff, thanks @marcemarc!

PR #8155, opened 2020-05-23
Accessibility improvement: umb-context-menu

BatJan:
### Prerequisites

-  ✔ I have added steps to test this contribution in the description below

### Description
In this PR I have optimized accessibility a little bit by moving the click handler onto the `<button>` element and added `aria-hidden="true"` on the `<i>` element. I have also optimized the code by refactoring the ng-if check so we instead pass `true` or `false` to the `umb-auto-focus` directive instead of duplicating the code blocks 😃


nul800sebastiaan:
Thanks @BatJan! Works like a charm. Don't forget to add `type="button"` to most of these buttons, I've done it for you here 👍

PR #8156, opened 2020-05-23
Sort dialog: Accessibility optimizations

BatJan:
### Prerequisites

### Description
Within this PR the following has been changed

- `<a>` is converted to `<button>` since they don't contain any `href` values
- `aria-hidden="true"` has been added to the section containing the arrow icons
- Fallback texts have been added in case localization of the dictionary key has not been added for a certain culture
- Also changed the error display to render the text within a `<p>` and moved the `ng-if` onto the wrapping `<div>`, meaning I removed the outer `<div>`, which is not needed

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/main.less
+++ b/src/Umbraco.Web.UI.Client/src/less/main.less
@@ -347,8 +347,14 @@ label:not([for]) {
   border: none !important;
 }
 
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Looks great @BatJan - thanks very much! 👍

PR #8157, opened 2020-05-24
Use element type description as tooltip in the Nested Content "Add content" dialog

### LINKED ISSUE #8073: nested content editor doesn't show doctype description
When setting up a datatype using the nested content editor, You specify the element docytypes that should be available to the content editor.

If you enter a description for the doctype like so:
definition

Its not presented to the content editor in the UI:
nested-content-picker

Would it be possible to add this enhancement? This would help inform/educate the content editor on which option they should be picking (for those more 'involved' cases)

Thanks

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8073

### Description

As discussed in #8073 it makes sense to use the element type descriptions in the Nested Content "Add content" dialog.

This PR shows the element type descriptions as tooltips on hover. I have opted to only show the tooltips when hovering the element type names, as it got too aggressive otherwise.

Obviously no tooltip is displayed if the element type does not have a description. 

Here's a GIF that shows the feature in action (only some of the element types have descriptions):

element-type-description-as-tooltip

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/overlays/itempicker/itempicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/overlays/itempicker/itempicker.controller.js
@@ -19,6 +19,31 @@ function ItemPickerOverlay($scope, localizationService) {
         $scope.submitForm($scope.model);
     };
 
+    $scope.tooltip = {
+        show: false,
+        event: null
+    };
+
+    $scope.showTooltip = function(item, $event) {
+        if (!item.tooltip) {
+            $scope.mouseLeave();
+            return;
+        }
+        $scope.tooltip = {
+            show: true,
+            event: $event,
+            text: item.tooltip
+        };
+    }
+
+    $scope.hideTooltip = function () {
+        $scope.tooltip = {
+            show: false,
+            event: null,
+            text: null
+        };
+    }
+
     onInit();
 
 }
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Lovely addition here, thanks very much @kjac! 👍

bjarnef:
@nul800sebastiaan @kjac it seems the `tooltip: scaffold.documentType.description` in nested controller is missing after merging https://github.com/umbraco/Umbraco-CMS/pull/8044

I noticed the child node picker on a document type is returning a console error because of `$scope.mouseLeave()` is undefined.

image

It seems it need some adjustments to work correct in this overlay.

image

bjarnef:
The strange position of the tooltip seems to be how it calculate the `tooltipStyles` in the tooltip directive, where it rely on `#contentwrapper`:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/umbtooltip.directive.js#L82

In overlays it should probably used the parent overlay container - maybe an option to specify this, e.g.

```
<umb-tooltip 
           ng-if="tooltip.show"
           event="tooltip.event"
           container="tooltip.container">
        {{ tooltip.text }}
    </umb-tooltip>
```

bjarnef:
I have submitted a PR https://github.com/umbraco/Umbraco-CMS/pull/8416 to fix this, while the current behavior with tooltips in e.g. media grid should still work. It could probably be improved further in future.

Issue #8033, opened 2020-04-29
Copy content node with descendants is not preserving sort


jpinto3488:
Umbraco 8.6.1 running on Umbraco Cloud.

Copy a node with some children, options:
- Relate to original: false
- Include descendants: true

The descendants of the copied element will have a different sort than the original.

nul800sebastiaan:
Ah yes, it looks like we order them by create date, would be good to get that fixed.

Note: not a Cloud problem but a CMS problem,

kjac:
Good one! PR in #8158

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8158

nul800sebastiaan:
False alarm! This was actually not fixed in that PR unfortunately. 🙈

vladicaivanovski:
So what is the status of this issue, when can we expect it to be fully resolved? 

Currently running on version 8.9.1, it's still present and currently a big issue for our site.

mathew65:
I was running version 8.10.1 and problem was still present. After upgrading to 8.12.1 it seems it's still not solved.

RyanHelmn:
Happy to take a look at this, will update here of any progress ✌

riccardocorazza:
I'm running version 8.12.1 and problem still present.
Any solutions?

nul800sebastiaan:
This is still "up for grabs", happy to look at any pull requests for this.

riccardocorazza:
Is this Pull Request not the solution?
[https://github.com/umbraco/Umbraco-CMS/pull/8158/files](https://github.com/umbraco/Umbraco-CMS/pull/8158/files)

ricbrady:
Hi @nul800sebastiaan 
I've fixed and tested the Bug.
How can I push it and create a pull request?

nul800sebastiaan:
Awesome @ricbrady! There's some explanation in our contributing docs: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/.github/CONTRIBUTING.md#how-do-i-begin

In short:

- Fork the Umbraco repository
- Clone your fork locally
- Make a branch for the fix
- Fix the problem
- Push the branch to your clone
- On GitHub you can now send us a pull request with that fix

ricbrady:
I've done. #10091

OwainWilliams:
Fixes #10091

nul800sebastiaan:
[...]

PR #8158, opened 2020-05-24
Preserve sort order when copying content

### LINKED ISSUE #7199: Change Name of node in split mode doesn't work when the language is not the default language
Hello,

My website has two languages, Spanish (default) and English.

My problem is that when I try two change **only** the Name of the node in both languages **in Split mode**, I publish the node but the change of the Name in the non default language (English) doesn't work.

Otherwise, if I change another property this doesn't happen, is only in split mode and with the Name of the node.

Regards!

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7199 8033

### Description

As described in #8033, the sort order isn't preserved when copying content; instead the copied content is sorted by creation order of the original content:

copy-descendants-before

This PR ensures that we copy descendants by order of their original sort order. Here's the same operation with this PR applied:

copy-descendants-after

### DIFF
--- a/src/Umbraco.Core/Services/Implement/ContentService.cs
+++ b/src/Umbraco.Core/Services/Implement/ContentService.cs
@@ -2174,7 +2174,7 @@ namespace Umbraco.Core.Services.Implement
                     var total = long.MaxValue;
                     while (page * pageSize < total)
                     {
-                        var descendants = GetPagedDescendants(content.Id, page++, pageSize, out total);
+                        var descendants = GetPagedDescendants(content.Id, page++, pageSize, out total, ordering: Ordering.By("sortOrder"));
                         foreach (var descendant in descendants)
                         {
                             // if parent has not been copied, skip, else gets its copy id
--- a/src/Umbraco.Tests/Services/ContentServiceTests.cs
+++ b/src/Umbraco.Tests/Services/ContentServiceTests.cs
@@ -1959,6 +1959,33 @@ namespace Umbraco.Tests.Services
             Assert.AreNotEqual(childCopy.Key, child.Key);
         }
 
+        
+        [Test]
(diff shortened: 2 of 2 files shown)


ltctech:
It default sorts by Path instead of SortOrder because it needs the parent to exist before the child. If a child appears before the parent it will skip and not copy the child.

The descendants probably need to be sorted by Path depth shallow first and then by SortOrder.

nul800sebastiaan:
Aha.. well I see what the problem is! It only works with children but not with descendants. I'll revert this for now 😅

kjac:
Ah! I'm on it 👍

ronaldbarendse:
`GetPagedDescendants()` will sort on path by default, so that ensures the parent is always copied before any child/descendant:

https://github.com/umbraco/Umbraco-CMS/blob/703037195aabd9237ad6af0e649333f2494ff2ba/src/Umbraco.Core/Services/Implement/ContentService.cs#L585-L589

But because the path also includes the current node ID, we can't add a second sorting on the sort order. If the recursive copy uses `GetPagedChildren()` instead of `GetPagedDescendants()` and does that in a recursive private/local function (so all descendants are still copied, not only direct children), that will keep the copies in the same order. One note though: the copies will not get the exact same sort order, as it gets renumbered from 0.

Just to check: can't we manually set the sort order to be the same (`descendantCopy.SortOrder = descendant.SortOrder`) if this isn't already done in `DeepCloneWithResetIdentities()` or does the `Save` method reset it?

goldmelodyvn:
I got the same issue, how is this status now? thanks

vladicaivanovski:
So what is the status of this issue, when can we expect it to be fully resolved?

Currently running on version 8.9.1, it's still present and currently a big issue for our site.

mathew65:
Still not solved on 8.12.1.

PR #8159, opened 2020-05-24
Infinite overlay: Datatype picker - Semantics and Accessibility fixes

BatJan:
### Prerequisites

- ✔️ I have added steps to test this contribution in the description below

### Description
In this PR I have changed the following
- Turned the `<a>` into a `<button>` since the `href` was empty
- Hidden the `<i>` from screen readers by adding `aria-hidden="true"`
- Changed `<h5>` to `<h2>` and `<h3>` to make it easier for screen readers to navigate (In a later PR I will add wrap the "Select editor" text in `<h1>` but it's in a different file)
- Added default texts in case no localization entry exists for certain languages

I hope you like it 😃


nul800sebastiaan:
I like it @BatJan! 👍

PR #8160, opened 2020-05-24
Umb-editor directive: Fix semantics

BatJan:
### Prerequisites

### Description
With this PR I have added a `<h1>` for the editors headline since this will improve semantics and accessibility as well. Furthermore I have fixed some stuff around valid HTML and wrapped the description text in a `<p>` tag as well.

I hope you like it 😃

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-header.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-header.html
@@ -12,16 +12,16 @@
         <div class="flex items-center" style="flex: 1;">
 
             <ng-form data-element="editor-icon" name="iconForm">
-                <button 
-                    type="button" class="umb-panel-header-icon" 
-                    ng-if="!hideIcon" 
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Makes sense to me, thanks @BatJan!

KevinJump:
Hi,

I don't know if this was intentional, but in checking the umb-panel-header-name to a h1, new styling has been introduced, as such if you have name and description locked. you get strangly spaced layout in the header.

so : 
```
        <umb-editor-header name="vm.page.title"
                           description="vm.page.description"
                           name-locked="true"
                           description-locked="true"
```

look like this pre 8.7 
image

in 8.7 it becomes. 
image

If this isn't intentional, i can log it as an issue ?

BatJan:
@KevinJump Ouch, that was cetainly not intentional - I'll do a PR fixing this straight way. Good catch 🙌

BatJan:
@KevinJump Proposed a fix for this in #8754

PR #8161, opened 2020-05-24
Update listviews after sorting children

### LINKED ISSUE #7995: Sorting items in listview via Actions menu does not update the list
When viewing children in a listview, if you use the actions menu to drag-sort the items and save, the listview does not update the sort numbers or order of items in the list.

## Umbraco version

I am seeing this issue on Umbraco version: 8.6

Reproduction
------------

### Bug summary

When viewing children in a listview, if you use the actions menu to drag-sort the items and save, the listview does not update the sort numbers or order of items in the lsit until you click the sort header to toggle the sort direction, or leave and come back to the listview.

### Steps to reproduce

When viewing a listview use the Actions menu to drag-sort the items and save.

### Expected result

After sorting items the listview should refresh to reflect the changes to the sort order.

### Actual result

Original list order.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7995

### Description

As discussed in #7995, the listviews do not update after performing a sort of the children displayed, leaving them somewhat out of sync.

This PR ensures that listviews are updated after a sort is performed on the children of the currently displayed content item. This applies to content and media listviews alike.

Since the sort order column is not part of the default listview, I have opted to only update the listviews after sorting if the sort column is indeed part of the listview. Thus the behavior for default listviews remain unchanged, and any performance penalty introduced by this PR will only affect those listviews where sort order is explicitly added as a column.

Here's a quick demo of this PR:

listview-reload-after-sort

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/content.sort.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/content.sort.controller.js
@@ -1,7 +1,7 @@
 (function () {
     "use strict";
 
-    function ContentSortController($scope, $filter, $routeParams, contentResource, navigationService) {
+    function ContentSortController($scope, $filter, $routeParams, contentResource, navigationService, eventsService) {
 
         var vm = this;
         var id = $scope.currentNode.id;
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Very slick, works like a charm, thanks @kjac ! 👍

PR #8162, opened 2020-05-24
Datatype configuration: Improve accessibility and UX

BatJan:
### Prerequisites

### Description
In this PR the following has been done
- Convert `<a>` to `<button>` since the `href` was empty
- Hide the icon by setting `aria-hidden="true"` on it
- Add some visuallyhidden texts to improve the screen reader experience
- Add the umb-auto-focus directive on the button and set the first button to be focused when the overlay opens instead of relying on an empty anchor tag - If no editor exists to configure then the "Add new" button will be focused instead - Below you can see the before and after effect.

I hope you like it and look forward to receive any feedback you might have as usual 😃 

**Before**
datatype-config-before

**After**
datatype-config-after


nul800sebastiaan:
Makes total sense to me, cheers @BatJan! 👍

PR #8163, opened 2020-05-25
Clone properties before manipulating them in the rollback window

### LINKED ISSUE #8146: V8: Repeatable textstring property grows with blank entries when open Rollback menu
Repeatable textstring datatype works not as expected when trying to navigate some other menus (detected a Rollback menu for now)

## Umbraco version 

Umbraco 8.6 (Pure Starter kit)

Reproduction
------------

### Steps to reproduce

1. Add property editor of "Repeatable textstring" type to any document type.
2. Focus the property **UPD:** or don't focus, it doesn't matter
3. Open a sidebar menu, "Info > Rollback" in my case
4. Return to the property, you will see it grown with a lot of empty entries

### Expected result

Nothing changed with the property

### Actual result

It has grown with lot of blank entries

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8146

### Description

#8146 describes how the rollback window causes the Repeatable Textstrings property editor to go bananas. The same thing happens to Nested Content (although not nearly as bad) and likely a bunch of other editors too.

Specifically this happens when doing a diff compare in the rollback window. Here's a GIF of the issue:

rollback-clone-before

This PR ensures that the rollback window performs diff compares on clones of the content properties, thus ensuring that any property manipulation performed for the sake of the diff does not affect the content editor. 

rollback-clone-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/rollback/rollback.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/rollback/rollback.controller.js
@@ -116,6 +116,11 @@
             // extract all properties from the tabs and create new object for the diff
             currentVersion.tabs.forEach((tab, tabIndex) => {
                 tab.properties.forEach((property, propertyIndex) => {
+
+                    // this is a reference to the property being displayed in the editor - we need to clone it before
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks @kjac - turns out this was fixed by @bjarnef already last year! 🙈 

I'll close this one, have merged his PR!

PR #8167, opened 2020-05-25
Ensure that "sort" is available as "order by" for custom listviews

### LINKED ISSUE #7340: V8 - List View - Default Sort Options
In v8, when creating a custom list view though the back office, it doesn't seem possible to set the default sort order to be based on "sort order". The drop down only consists of Name, Last Edited, and Created By options, however Sort Order seems perfectly reasonable option.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7340

### Description

See #7340 for a detailed issue description. 

TL;DR: When creating a custom listview for a content type, "Sort" is not available as an "Order By" option. It only becomes available after saving and reloading the browser window:

create-listview-orderby-missing-sortorder-before

This happens because the listview editor reverts to the built-in listview configuration (formerly known as prevalues) when creating a new custom listview, instead of adhering to the default configuration as dictated by Umbraco. Since the built-in listview configuration is missing "Sort", it appears to be missing when creating the new custom listview. 

This PR fixes it; here's the PR applied in action:

create-listview-orderby-missing-sortorder-after

As you can tell, the listview editor reverts back to the built-in listview configuration again after removing the custom listview.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umblistviewsettings.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umblistviewsettings.directive.js
@@ -12,16 +12,18 @@
 
             /* ---------- INIT ---------- */ 
 
+            const setDataType = (dataType) => {
+                scope.dataType = dataType;
+                listViewPrevalueHelper.setPrevalues(dataType.preValues);
+            }
+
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Great stuff, thanks @kjac!

PR #8169, opened 2020-05-25
Listview "Order by" should not contain removed columns

kjac:
### Prerequisites

### Description

The "Order by" selector does not update when removing listview columns:

listview-sortby-removed-properties-before

This PR fixes it:

listview-sortby-removed-properties-after


nul800sebastiaan:
Oh yes, that was a bit silly, thank you for the help @kjac! 👍

PR #8170, opened 2020-05-25
Fix a few localization glitches in the listview "Order by" selector

kjac:
### Prerequisites

### Description

There are a few issues in the "Order by" selector when editing listviews:

1. If you remove and add the "Last edited" column, its name won't get translated like the rest of the system fields - it shows as `undefined`.
2. Non-system fields also show as `undefined` until you fill out a heading for them.

listview-orderby-localization-before

I also found some unused code that could be deleted while I was at it. Here's a GIF of the same operations with this PR applied:

listview-orderby-localization-after


nul800sebastiaan:
Sweet, that helps indeed! 👍

PR #8175, opened 2020-05-26
Fix glitch in umb-checkbox background

kjac:
### Prerequisites

### Description

_Note: This is a regression, it was also fixed in #4981_

On some screens I'm seeing a white border in some usages of `umb-checkbox`. It might be a screen resolution/interpolation thing that sets it off. Curiously enough it's on every other checked checkbox on the RTE configuration:

image

Anyway... the fix is quite harmless - it's adding a background color to the checked state of the checkbox. It looks like this:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-form-check.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-form-check.less
@@ -45,6 +45,7 @@ label.umb-form-check--checkbox{
         }
         &:checked ~ .umb-form-check__state .umb-form-check__check {
             border-color: @ui-option-type;
+            background-color: @ui-option-type;
         }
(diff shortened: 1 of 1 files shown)


clausjensen:
cherrypicked to 8.6.3

bjarnef:
@kjac how does the `umb-radiobutton` look with this change?

wouldn't it make it look like this?

image

clausjensen:
@bjarnef absolutely correct. I had not seen that 😢 Great catch.
@kjac can you make the selector more specific, while ensuring it still works? 😃

kjac:
@clausjensen oh shoot... good catch, @bjarnef - I'll have a look at it tonight 👍

clausjensen:
@kjac could you tag me when you submit a fix - I need to cherrypick for 8.6.3 again so just to be sure I'm notified

PR #8177, opened 2020-05-26
Adds in NC Property Component to deal with keys that were duplicated

warrenbuckley:
When copying a node only the Copying & Copied events are emitted and not the Saving & Saved events that were assumed would fire.

When copying a content node, it will create new GUID string keys for all nested content items, including a recursive if one of the JSON properties of the nested content, included another nested content item, as opposed to removing/clearing the keys as originally suggested for the solution.

Saving & Publishing nodes will only add new keys if they are missing, which requires @nielslyngsoe #7166 JS PR as he removes the key property & further nested content inside nested content key properties from the clientside code when a nested content item/s are copied & used.

## Note
This PR is built upon @nielslyngsoe JS changes needed to deal with copying nested content items, so be mindful that this PR targets his PR as opposed to V8/dev - https://github.com/umbraco/Umbraco-CMS/pull/7166

## Test Notes
* Use a package such as Igloo which uses Nested Content alot
* Create a page with a hero and add buttons to it (which is also another nested content)
* Ensure all keys are unique
* Copy the hero inside the nested content UI
* Create a new copy with slightly different content
* Save or Save & Publish Node
* Reload node in backoffice & verify XHR request in DevTools for node that the NC property items all have unique keys

* Copy the entire content node with the right click context menu
* Ensure all NC items have new/unique keys different from the original content item (please compare)

## Other Tests

* Verify this works OK with segments/culture variants and each language variant of the nested content gets unique keys as well.

Fixes AB#6684

### DIFF
--- a/src/Umbraco.Web/Compose/NestedContentPropertyComponent.cs
+++ b/src/Umbraco.Web/Compose/NestedContentPropertyComponent.cs
@@ -1,0 +1,174 @@
+﻿using System;
+using System.Linq;
+using Newtonsoft.Json.Linq;
+using Umbraco.Core;
+using Umbraco.Core.Composing;
+using Umbraco.Core.Events;
+using Umbraco.Core.Models;
+using Umbraco.Core.Services;
+using Umbraco.Core.Services.Implement;
+using Umbraco.Web.PropertyEditors;
+
+namespace Umbraco.Web.Compose
+{
+    public class NestedContentPropertyComponent : IComponent
+    {
+        public void Initialize()
+        {
+            ContentService.Copying += ContentService_Copying;
+            ContentService.Saving += ContentService_Saving;
+            ContentService.Publishing += ContentService_Publishing;
+        }
+
+        private void ContentService_Copying(IContentService sender, CopyEventArgs<IContent> e)
+        {
+            // When a content node contains nested content property
+            // Check if the copied node contains a nested content
+            var nestedContentProps = e.Copy.Properties.Where(x => x.PropertyType.PropertyEditorAlias == Constants.PropertyEditors.Aliases.NestedContent);
+
+            // Each NC Property on a doctype
+            foreach (var nestedContentProp in nestedContentProps)
+            {
+                // A NC Prop may have one or more values due to cultures
+                var propVals = nestedContentProp.Values;
+                foreach (var cultureVal in propVals)
+                {
+                    // Remove keys from published value & any nested NC's
+                    var updatedPublishedVal = CreateNewNestedContentKeys(cultureVal.PublishedValue.ToString());
+                    cultureVal.PublishedValue = updatedPublishedVal;
+
+                    // Remove keys from edited/draft value & any nested NC's
+                    var updatedEditedVal = CreateNewNestedContentKeys(cultureVal.EditedValue.ToString());
+                    cultureVal.EditedValue = updatedEditedVal;
+                }
+            }
+        }
+
+        private void ContentService_Saving(IContentService sender, ContentSavingEventArgs e)
+        {
+            // One or more content nodes could be saved in a bulk publish
+            foreach(var entity in e.SavedEntities)
+            {
+                // When a content node contains nested content property
+                // Check if the copied node contains a nested content
+                var nestedContentProps = entity.Properties.Where(x => x.PropertyType.PropertyEditorAlias == Constants.PropertyEditors.Aliases.NestedContent);
+
+                // Each NC Property on a doctype
+                foreach (var nestedContentProp in nestedContentProps)
+                {
+                    // A NC Prop may have one or more values due to cultures
+                    var propVals = nestedContentProp.Values;
+                    foreach (var cultureVal in propVals)
+                    {
+                        // Remove keys from published value & any nested NC's
+                        var updatedPublishedVal = CreateMissingNestedContentKeys(cultureVal.PublishedValue.ToString());
+                        cultureVal.PublishedValue = updatedPublishedVal;
+
+                        // Remove keys from edited/draft value & any nested NC's
+                        var updatedEditedVal = CreateMissingNestedContentKeys(cultureVal.EditedValue.ToString());
+                        cultureVal.EditedValue = updatedEditedVal;
+                    }
+                }
+            }
+        }
+
+        private void ContentService_Publishing(IContentService sender, ContentPublishingEventArgs e)
+        {
+            // One or more content nodes could be saved in a bulk publish
+            foreach (var entity in e.PublishedEntities)
+            {
+                // When a content node contains nested content property
+                // Check if the copied node contains a nested content
+                var nestedContentProps = entity.Properties.Where(x => x.PropertyType.PropertyEditorAlias == Constants.PropertyEditors.Aliases.NestedContent);
+
+                // Each NC Property on a doctype
+                foreach (var nestedContentProp in nestedContentProps)
+                {
+                    // A NC Prop may have one or more values due to cultures
+                    var propVals = nestedContentProp.Values;
+                    foreach (var cultureVal in propVals)
+                    {
+                        // Remove keys from published value & any nested NC's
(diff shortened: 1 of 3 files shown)


clausjensen:
Tested this and confirmed it works.

- Copying items in the UI.
- Copying items in the UI with nested items inside them.
- Copying entire node with nested items and nested items inside.
- All of the above with variants enabled (and tested in a variant not the default language as well).

All tests confirmed to have new keys assigned on items that should be considered new and unique.

- Also tested that existing items in the above cases, being resaved - keep their existing keys.

src/Umbraco.Web/Compose/NestedContentPropertyComponent.cs
```diff
@@ -0,0 +1,180 @@
...
+            ContentService.Saving -= ContentService_Saving;
+            ContentService.Publishing -= ContentService_Publishing;
+        }
+
+        private string CreateNewNestedContentKeys(string ncJson)
+        {
+            if (string.IsNullOrWhiteSpace(ncJson))
+                return ncJson;
+
+            // Try & convert JSON to JArray (two props we will know should exist are key & ncContentTypeAlias)
```
Shazwazza, line 115: This isn't really 'trying' though is it :) if this fails it will throw.

src/Umbraco.Web/Compose/NestedContentPropertyComponent.cs
```diff
@@ -0,0 +1,180 @@
...
+                        ncItemProp.Value = Guid.NewGuid().ToString();
+
+                    // No need to check this property for JSON - as this is a JSON prop we know
+                    // That onyl contains the string of the doctype alias used as the NC item
+                    if (ncItemProp.Name == NestedContentPropertyEditor.ContentTypeAliasPropertyKey)
+                        continue;
+
+                    // As we don't know what properties in the JSON may contain the nested NC
+                    // We are detecting if its value stores JSON to help filter the list AND that in its JSON it has ncContentTypeAlias prop
+                    if (ncItemProp.Value.ToString().DetectIsJson() && ncItemProp.Value.ToString().Contains(NestedContentPropertyEditor.ContentTypeAliasPropertyKey))
```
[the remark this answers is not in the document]
Shazwazza: This isn't a very nice way of recursing. I realize this will 'work' but what if someone types in the value of "ncContentTypeAlias" in one of their textbox values? This will think it needs to recurse and will probably explode because we aren't error handling the serialized value when trying to parse into JSON. Also this will not work if someone has a nested content cell with another non-nested content editor that also contains a nested content editor. 

IMO the way to do this is to 'just' recurse with the `ncItemProp.Value.ToString()` (if it's not null), don't do this check since it will be unnecessary and properly parse out the json defensively. But you would need to keep recursing even if it's not nested content and you'd probably want to recurse on a JToken not a string. 

There's probably a nicer way to do that too and use JsonPath instead https://www.newtonsoft.com/json/help/html/QueryJsonSelectTokenJsonPath.htm to select all tokens that you are looking for and then update from there. If you can make that work that would be far better.

src/Umbraco.Web/Compose/NestedContentPropertyComponent.cs
```diff
@@ -0,0 +1,180 @@
...
+                        // Recurse & update this JSON property
+                        ncItemProp.Value = CreateNewNestedContentKeys(ncItemProp.Value.ToString());
+                    }
+                }
+            }
+
+            return ncItems.ToString();
+        }
+
+        private string CreateMissingNestedContentKeys(string ncJson)
```
Shazwazza, line 144: This is almost a total duplication of the above :/

warrenbuckley: Fixed duplication, the method takes in a bool to know if it's replacing the key values OR only updating missing key properties to have a new key

warrenbuckley:
@Shazwazza can you do a re-review please, hopefully, I understood your points yesterday/today
If still needs re-work can we take a call so I can get it sorted & ensure I have a stronger coffee ☕

src/Umbraco.Web/Compose/NestedContentPropertyComponent.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        private string CreateNestedContentKeys(string rawJson, bool onlyMissingKeys)
+        {
+            if (string.IsNullOrWhiteSpace(rawJson))
+                return rawJson;
+
+            // Parse JSON
+            var ncJson = JToken.Parse(rawJson);
```
Shazwazza, line 75: This should be even more safe, you don't really know if this string is json or if it is going to parse. You can at least use our string extension DetectIsJson to check if it might be json first

src/Umbraco.Web/Compose/NestedContentPropertyComponent.cs
```diff
@@ -0,0 +1,128 @@
...
+
+                    if (ncItemPropVal.DetectIsJson())
+                    {
+                        // Parse the nested JSON (complex editor)
+                        var complexEditorJson = JToken.Parse(ncItemPropVal);
+
+                        // Verify the complex editor is nested content (Will have one or more ncContentTypeAlias)
+                        // One for each NC item/row
+                        var hasNestedContentJsonProp = complexEditorJson.SelectTokens($"$..['{NestedContentPropertyEditor.ContentTypeAliasPropertyKey}']", false);
+                        if (hasNestedContentJsonProp.Count() > 0)
```
Shazwazza, line 116: This check will work if we have NC nested within another, but if we are currently in an editor that isn't NC that is within NC this won't work. Where are the unit tests? I don't understand how you are testing this stuff without that, it would be such a time consumer to re-test this over and over in real life plus you will never be sure you are covering all cases. It will be much easier to develop this against tests!  So this still wont work if you have NC -> some other editor -> NC especially when the sub json is escaped.

This is not doing what we talked about in our last call. If this is false, then we know the current JSON doesn't contain a direct reference to NC but we don't know if theres a sub-reference. What you had before where you just do a brute force string check if it contains `NestedContentPropertyEditor.ContentTypeAliasPropertyKey` is correct because we have to deal with escaped JSON. So do the string.invariantcontains check, if that is true, then you know somewhere nested within this json there's probably an NC, though it could also just be a string of that alias someone typed in, but at least at that stage you know you need to recurse further. I think this is just gonna be faster if i re-do this for you.

Shazwazza:
@warrenbuckley i have pushed changes along with unit tests. I have added TODO notes about what other tests there are to write. This must be done. I cannot fathom how you have got to this point without unit tests. The logic is much simpler now like we talked about: 

* Safely get a JToken (any JToken)
* Flatten it with jsonpath to get all values
* Iterate each value and check if it's an NC property, if it is get it's siblings 'key' and update it
* if its not originally a NC editor val and if the property isn't key/ncContentTypeAlias recurse if there's potential that it contains a nested NC (at any level via any complex editor)

I've written tests for a simple case, a nested case without escaping and a nested case with escaping. you need to write more tests, see TODO notes.

And then once those are written and passing test again in real life.

warrenbuckley:
@Shazwazza seems you forgot to add the test file `PropertyEditors\NestedContentPropertyComponentTests.cs` to the commit  - do you mind pushing it if you get a chance please

warrenbuckley:
Manually tested with Igloo/UNO, which uses NestedContent, with other NestedContent items inside it without issues.

Copying NC items to ensure new keys for these items, when node saved/published
Copying the entire node ensures all NC items get new keys

warrenbuckley:
Tested and reviewed by @clausjensen as well
Aiming to prep a release for 8.6.2 on Monday.

>Can add in tests (and more tests) when Shannon is back

warrenbuckley:
This PR along with Niels PR with JS clipboard service #7166
Has been cherry picked into v8/8.6 for an 8.6.2 release with commit - 1fe7de0b5f78d3d2d7843a02f520fe65b21a2952

PR #8178, opened 2020-05-26
Make SQL mapper functionality public

### LINKED ISSUE #8176: Make MapperForAttribute/SqlTranslator public
I'm looking at creating a queryable interface for my entities in Vendr and would like to make use of Umbracos LINQ visitors that convert entity linq expressions into SQL statements.

In order to do this it looks like I can use the `IQuery<>` interface and define my own custom maps from the entity object to the state object. This all appears to be public apart from one element which is the `MapperForAttribute` which is required in order to be able to resolve a given mapper for a type. Without this attribute I can define my mapper, but it will never get picked up.

I would therefore like to request that the `MapperForAttribute` be made public so that 3rd parties may be able to make use of this query interface for their own needs.

mattbrailsford:
### Description
Fixes #8176 making the umbraco sql mapper functionality public for third parties to use


mattbrailsford:
I'm not sure if `SqlExpressionExtensions` also wants making public 🤔

PR #8179, opened 2020-05-26
Add umb-search-filter component

BatJan:
### Prerequisites

### Description
I noticed some repeatable patterns regarding the filter search boxes found in for instance the "redirect dashboard" and within the "Editor" overlay when setting up new document types and adding datatypes etc.

So instead of fixing the same accessibility issues 3 times I figured it would be nicer to wrap it in a reusable component instead - So that is what I have done 😃 

Hopefully the component can be modified at a later stage to cater for the same behavior one sees in the "mini-list-view" where the input field expands when it recieves focus - But I think that's for another PR 👍 

Looking forward to receive any feedback you might have regarding this one.


nul800sebastiaan:
Great work @BatJan - works as expected, nice to consolidate the same boxes! 👍 There's probably a few more that can be replaced.

PR #8185, opened 2020-05-27
Logging content template save in back office to log table

poornimanayar:
Similar to #8140 I am trying to log the saving of a content template into the `umbracoLog` table. Currently it does not save into the `umbracoLog` table.

with this PR applied, when a content template is saved the following SQL query should bring back results on content template as well.

SELECT TOP (10) [id]
      ,[userId]
      ,[NodeId]
      ,[entityType]
      ,[Datestamp]
      ,[logHeader]
      ,[logComment]
      ,[parameters]
  FROM [dbo].[umbracoLog]
  order by Datestamp desc


nul800sebastiaan:
Awesome, thanks for this @poornimanayar 👍

PR #8188, opened 2020-05-27
Added GetLog method to LogController

poornimanayar:
Currently the `getLog` method in `logResource`  https://our.umbraco.com/apidocs/v8/ui/#/api/umbraco.resources.logResource errors out as the `LogController` does not have a `GetLog` method. I have created a method in the `LogController` to overcome the issue. I am assuming that the method needs to be added rather than the method in the resource marked obsolete. If that's the case let me know. I will amend it accordingly.

You can test this change using the attached dashboard I have created. Without this change in place it should show an error notification. With the change in place it should show some results as long as some content or any other entity has been saved.

[TestDashboard.zip](https://github.com/umbraco/Umbraco-CMS/files/4689187/TestDashboard.zip)


nul800sebastiaan:
Sweet!! Thanks very much @poornimanayar 👍

PR #8189, opened 2020-05-27
Fix HTML markup of umb-list-view-settings.html

rbottema:
I've noticed this file was missing a closing div tag, so I added it.


nul800sebastiaan:
Good catch @rbottema, thanks! 👍

PR #8190, opened 2020-05-27
Fix HTML markup of includeproperties.prevalues.html

rbottema:
I've noticed this span was missing a closing >, so I added it.


nul800sebastiaan:
Lovely, thank you kindly @rbottema!

PR #8191, opened 2020-05-27
umb-lightbox improvements

BatJan:
### Prerequisites

### Description
In this PR I have done some wee styling changes, which are not really noticeable but had to be done after the switch from `<div>` to `<button>` 😃 

So what has happened is

- The `<div>` elements have been converted to `<button>`
- Missing localization texts have been added
- Styling has been modified

Works and looks the same as before. At a later point it would be good to align the backdrop so it's triggered by the backdrop service instead.


nul800sebastiaan:
Looking good, cheers @BatJan!

PR #8195, opened 2020-05-28
Improve dictionary dashboard performance significantly

kjac:
### Prerequisites

### Description

As the dictionary grows larger, the dictionary dashboard turns slower. When you have a somewhat sizable dictionary, the dashboard takes ages to load.

This PR improves the dashboard load speed significantly... in my test scenario about 75 times - load time is cut from ~30s to ~400ms on average:

Measurements

Those ~30s can mainly be credited to recursive operations on the database whereas this PR fetches all dictionary items in one go. So there is a significant performance improvement towards the database as well.

### Testing this PR

First and foremost you must have a somewhat large dictionary to test this. You can use the following SQL to generate one (do backup your database first, though!).

[Add dictionary items.txt](https://github.com/umbraco/Umbraco-CMS/files/4693318/Add.dictionary.items.txt)

The SQL generates the dictionary structure used in the performance measurements above:

image

Now load the dashboard a few times. Without this PR applied it'll take a loooong time to load. With this PR applied it is pretty darn fast 😄 

I have compared the output of the `GetList()` API call and it remains unchanged (the size on the measurements above also indicate this).

### DIFF
--- a/src/Umbraco.Web/Editors/DictionaryController.cs
+++ b/src/Umbraco.Web/Editors/DictionaryController.cs
@@ -207,44 +207,31 @@ namespace Umbraco.Web.Editors
         /// </returns>
         public IEnumerable<DictionaryOverviewDisplay> GetList()
         {
-            var list = new List<DictionaryOverviewDisplay>();
+            var items = Services.LocalizationService.GetDictionaryItemDescendants(null).ToArray();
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Sweet improvements!! 😲

Thanks @kjac! ⭐

PR #8196, opened 2020-05-28
GetControllerTypeInternal should release the controller instance if it creates one

stevemegson:
### Prerequisites

### Description
`ControllerFactoryExtensions.GetControllerTypeInternal` and `MasterControllerFactory.GetControllerTypeInternal` will fall back to calling `CreateController` if a custom controller factory is being used.

In this case, it should then call `ReleaseController`. The factory may have retained a reference to the instance, leading to a memory leak if the instance is never released.


nul800sebastiaan:
I was about to say "but we end up here and there's nothing to release! https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Composing/LightInject/LightInjectContainer.cs#L131"

But yes, then I noticed the fact that you mentioned a custom controller factory, then it makes sense!  😉

PR #4720, opened 2019-02-24
V8: Change "Include subfolders in search" to "Search only in this folder"

kjac:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes: https://github.com/umbraco/Umbraco-CMS/issues/4707

### Description

#4707 outlines the requirements for this PR 😄 here's how it looks when applied:

search-only-this-folder

While I was at it I found some unused localization going on in the media picker controller, so I removed that as well.

PR #8197, opened 2020-05-28
Search the entire subtree in media pickers

kjac:
### Prerequisites

### Description

As discussed [here](https://github.com/umbraco/Umbraco-CMS/pull/4720#issuecomment-635301456) the "Search only this folder" option in the media picker should search the current folder _and_ all sub folders:

image

As-is it _only_ searches the current folder - all sub folders are omitted from the search results.

With this PR applied, the media picker lists search results form the entire subtree below the current folder:
 
media-picker-search-subtree

I have taken the opportunity to clean up a bit, as some of the variable/label naming around this feature doesn't make sense right now. Also it turns out that the removal of sub folder results is actually performed clientside (!) in `<umb-media-grid />`, and since media picker is the only one currently using that feature in the entire codebase, I have removed the feature altogether.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbmediagrid.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbmediagrid.directive.js
@@ -126,20 +126,6 @@ Use this directive to generate a thumbnail grid of media items.
                         scope.items.splice(i, 1);
                         i--;
                     }
-
-                    // If subfolder search is not enabled remove the media items that's not needed
-                    // Make sure that includeSubFolder is not undefined since the directive is used
-                    // in contexts where it should not be used. Currently only used when we trigger
-                    // a media picker
-                    if (scope.includeSubFolders !== undefined) {
-                        if (scope.includeSubFolders !== 'true') {
-                            if (item.parentId !== parseInt(scope.currentFolderId)) {
-                                scope.items.splice(i, 1);
-                                i--;
-                            }
-                        }
-                    }
-
                 }
                 
                 if (scope.items.length > 0) {
@@ -344,7 +330,6 @@ Use this directive to generate a thumbnail grid of media items.
                 disableFolderSelect: "@",
                 onlyImages: "@",
                 onlyFolders: "@",
-                includeSubFolders: "@",
                 currentFolderId: "@"
             },
             link: link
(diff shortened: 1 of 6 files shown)


nul800sebastiaan:
Thank you @kjac for the prompt update! 

We're still discussing what the checkbox name should be, I think we've landed on "Search My Folder Name", but I can see if I can make that update as well. 👍

nielslyngsoe:
@kjac @nul800sebastiaan I'm cool with both, so up to you to pick what you think is best. I can add that "Search __FolderName__" fits with the UI in Windows, so it might be more recognizable for users. It also lets us get rid of the "only" word which might be what would trigger the interpretation of a very limited search.

ronaldbarendse:
This makes much more sense indeed, but like @nul800sebastiaan suggests, the UI needs a way to show the current folder name when searching! Currently, the breadcrumb disappears when searching, even when 'Search only this folder' is checked.

IMHO the best solution would be to always search from the current folder/tree and show a button below the search results to search the whole media library (if you're not already searching from the root). That makes the checkbox obsolete (cleaning up the UI and saving some vertical space) and will create a better editor workflow:
- Only showing more results if desired, instead of having to enable the checkbox to show less
- First enabling the checkbox below the search field, enter the search term above and having the results show below: think about accessibility and the amount of tabbing!

nielslyngsoe:
Great point @ronaldbarendse, no need to show the checkbox unless you are searching. Actually I would argue that we could get rid of it completely.
If we changed placeholder text of the search-bar, to "Type to search in __FolderName__", and we always perform a local search.

Then we could keep breadcrumbs when searching.

Might need to add a symbol, in the end, to indicate that the shown items not necessary are direct children. Something like this maybe:

`root / someFolder / __currentFolder__ / SearchResult`
or
`root / someFolder / __currentFolder__ / 〜`

By still having the breadcrumbs available the scope of your search becomes much more clear, and you can easily broaden your search by clicking one of the parent folders.

Thanks for your inspiration!

nielslyngsoe:
@ronaldbarendse great summery, I think its a really good approach, I do though think we should keep the first attempt simple. Meaning lets keep out the last two points. I'm quite certain that most users will understand the context, so I won't be affright of that. But to help the users we could spice the breadcrumbs up, with a little label/text above it saying:

Searching in:
Media / Products / (Search-icon)

I think that should make it work. :-)

I will greater an up-for-grabs issue, to keep it separate from this thread (since this PR is closed)

nielslyngsoe:
Issue created https://github.com/umbraco/Umbraco-CMS/issues/8219

nathanwoulfe:
This has been resolved elsewhere - searching media in v8/contrib shows the 'Search only this folder' checkbox. Other points raised (and covered in #8219) remain.

Will close this one, in favor of #8219.

PR #8194, opened 2020-05-27
Umb-confirm-directive: Hide icons and add missing screen reader friendly texts

BatJan:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
In this PR I have added screen reader friendly texts on the buttons and added `aria-hidden="true"` on the `<i>` element ensuring no screen reader announces them.

PR #8198, opened 2020-05-28
umb-confirmation directive: move trashcan into directive and address accessibility issues

BatJan:
### Prerequisites

### Description
While I did the PR for #8194 I afterwards noticed the pattern for using the directive was a bit weird/disconnected... Currently it relies on an outside button to trigger the "confirm" or "undo" icons... Which has then led to different patterns of implementing. Some places a `<button>` has been used, some places a `<i>` and some places dictionary items have been added, some places not and then it hit me that the "trigger" should also be part of the directive making the usage easier and more consistent.

I have done the following
- Extended the directive to contain "show" and "on-delete" properties
- Replaced all old patterns with the new pattern ensuring consistency
- **Edit:** Ensured that the implementations in 3rd party packages will still work using the old pattern 👍 

~~I'm considering if perhaps the directive should be renamed to something else in order to maintain backwards compatibility since the changes I have made are breaking. I'm thinking that perhaps it would also be a good idea to consider the look and feel to be more in line with the backoffice in general so perhaps renaming the directive would be best?~~

~~Thoughts @nul800sebastiaan and @nielslyngsoe ? 😃~~ Nevermind this comment I figured out a solution that avoids breaking it.


nul800sebastiaan:
Wonderful, thanks again @BatJan - this works great, cheers! 🍻

nul800sebastiaan:
Unfortunately this was a breaking change for people using the `umb-confirm-action` directive, we'll get this reverted in v8.8.1 so it works again for people relying on it.

nul800sebastiaan:
Reverted in f9b49fc5f35dfd331388df76b24b91d6c11c9b2b

nielslyngsoe:
@BatJan @nul800sebastiaan okay, so I've updated this PR to be backwards compatible. Please re-review :-D

PR #8200, opened 2020-05-29
V8/bugfix/gulp update inline views on watch

nielslyngsoe:
I changed a few things in the Gulp script to make it work with watching and updating inline views.
And corrected the setup for Unit Test server


nul800sebastiaan:
👍

PR #8202, opened 2020-05-29
fix: remove auto focus from package search box

MMasey:
This PR removes the auto-focus on the package search box. 

image

This issue was identified in the  Accessibility audit for version 8.4

**Rational**
> The search field receives focus on page load.
> 
> This differs from other pages within the menu as their focus remains within the menu when the page loads.
> 
> It is good that the focus goes into the page when the menu item is activated, but because of the focus it misses the secondary nav (Packages, Installed etc.)

This PR makes the experience consistent with the rest of the pages in the backoffice.


nul800sebastiaan:
Lovely, all good on my end, thanks @MMasey! 👍

PR #8205, opened 2020-05-30
Random fixes for nested content "add" button

BatJan:
### Prerequisites

### Description
I noticed some fallback texts were missing if the dictionary labels for certain language don't exists so I added some fallback texts.

Apart from that I made a tiny accessibility improvement adding the `aria-disabled` attribute and whenever the max limit of allowed blocks in nested content screen readers will announce either "disabled" or "dimmed" letting screen reader users know the button won't be doing anything.

In order to disable it totally from being clicked we would need to set `disabled="true"` but screen readers currently have varying support for this, so it works inconsistently and could leave some users in the dark. So click events will still be fired but since there is a limit no new blocks will be added anyway 👍


nul800sebastiaan:
So random... but no, very good! 👍 Thanks @BatJan!

PR #8206, opened 2020-05-30
Improve the accessibility of the cog icon

BatJan:
### Prerequisites

### Description
Currently the settings cog icon in the grid is wrapped in a div and has no screen reader friendly text, which this PR aims to fix.

- I have wrapped the icon in a `<button>` and hidden the icon using `aria-hidden="true"`
- I have also added a generic text for screen readers announcing "Open column settings" - At a later stage we need to make this a tad more specific towards the context of the cog's placement. But baby steps! 👶 😃 
- I have also added some fallback texts and wrapped them in `<p></p>` in case dictionary items have not been added for some reason while I was at it 👍


nul800sebastiaan:
Great, SuperTak @BatJan! 👍

Issue #8115, opened 2020-05-14
notificationsService documentation specifies an "actions" option but value isn't used


abjerner:
https://our.umbraco.com/apidocs/v8/ui/#/api/umbraco.services.notificationsService#methods_add

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/services/notifications.service.js#L56

The documentation mentions a `actions` option you can use to add one or more buttons to a notification. But neither the [service](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/services/notifications.service.js), [directive](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/notifications/umbnotifications.directive.js) or [view](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/notifications/umb-notifications.html) have any references to the `actions` option. At least when checking the latest code for both V7 and V8. So not sure if it was ever been supported 😮 

So the million dollar questions now: should the option be removed from the documentation, or should support for these actions buttons be implemented?

When reading up on the documentation for the `notificationsService` for a package, my idea was that an error notification could feature a "read more" action/button leading to the package documentation or an overlay within Umbraco so that users can read more about what the error actually means.

nul800sebastiaan:
Thanks @abjerner - I have no idea, but since this [was added 6 years ago with no supporting code](https://github.com/umbraco/Umbraco-CMS/commit/6f28a0a6864371cb8f6f6aad4d8b62c901612b0b#diff-49740a97c5186ae9750702e133f9fac1R44), I think this was just an idea that never came to fruition. I think removing it from the docs is the only thing  to do here. Wanna help? :-)

umbrabot:
Hi @abjerner,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

abjerner:
[...]

PR #2965, opened 2018-09-11
Removed the anchor tag around notification

poornimanayar:
### Prerequisites

- [ ] I have [created an issue](https://github.com/umbraco/Umbraco-CMS/issues) for the proposed changes in this PR, the link is: https://github.com/umbraco/Umbraco-CMS/issues/2964

### Description
Removed the <a> around notifications. The notifications should not have a hover state now.

PR #8207, opened 2020-05-30
umb-notifications.html - refactored switch logic

leekelleher:
### Prerequisites

### Description

A while ago I was trying to figure out why I couldn't get the `url` property to render in the `umb-notifications.html` view.

I commented about it here: https://github.com/umbraco/Umbraco-CMS/pull/2965#discussion_r401619104

I couldn't see how the `ng-switch` statement worked, (as it's intended to work with literal values, not objects).  So I swapped out the `ng-switch` statement with a couple of `ng-if` conditions.

In terms of testing, I made a test dashboard harness...

**package.manifest**

```json
{
    "javascript": [
        "~/App_Plugins/LeeTest/dashboard.js"
    ],
    "dashboards": [{
            "alias": "leeTestDashboard",
            "view": "/App_Plugins/LeeTest/dashboard.html",
            "sections": ["content"],
        }
    ]
}
```

**dashboard.html**

```html
<div ng-controller="LeeTest.Controller as vm">
    <button type="button" ng-click="vm.click()">Click me</button>
</div>
```

**dashboard.js**

```js
angular.module("umbraco").controller("LeeTest.Controller",[
    "$scope",
    "notificationsService",
    function ($scope, notificationsService) {

        var vm = this;

        vm.types = ["success", "warning", "error", "info"];

        vm.click = function() {

            vm.types.forEach(function(type) {

                notificationsService.add({
                    headline: type,
                    message: "This a message.",
                    url: "/umbraco/#/settings",
                    type: type,
                    sticky: false
                });

            });

        };
    }
]);
```

---
I've added an additional commit to change couple of quotes from single to double, for consistency.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/notifications/umb-notifications.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/notifications/umb-notifications.html
@@ -7,13 +7,14 @@
             <div ng-if="notification.view">
                 <div ng-include="notification.view"></div>
             </div>
-            <div ng-if="notification.headline" ng-switch on="{{notification}}">
-                <a ng-href="{{notification.url}}" ng-switch-when="{{notification.url && notification.url.trim() != ''}}" target="_blank">
-                    <strong>{{notification.headline}}</strong>
+
+            <div ng-if="notification.headline">
+                <a ng-if="notification.url" ng-href="{{notification.url}}" href="" target="_blank">
+                    <strong ng-bind="notification.headline"></strong>
                     <span ng-bind-html="notification.message"></span>
                 </a>
-                <div ng-switch-default>
-                    <strong>{{notification.headline}}</strong>
+                <div ng-if="!notification.url">
+                    <strong ng-bind="notification.headline"></strong>
                     <span ng-bind-html="notification.message"></span>
                 </div>
             </div>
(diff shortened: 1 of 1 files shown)


leekelleher:
Thanks @abjerner!

nul800sebastiaan:
So does this resolve #8115? Or is more needed for #8115?

leekelleher:
@nul800sebastiaan IMO this PR isn't related to #8115 - that one appears to be about the `actions` property.
This PR made the `url` property work again.

abjerner:
Yes, it's two different things, but both related to notifications. I plan on submitting a PR for `actions`, but hopefully Seb can provide some feedback there.

Anyways - I played around with `actions` after merging this PR. In my proposal for `actions`, I have a `target` property to compliment the `url` property of the action - eg. if it is an external link, it may be best to open in a new new tab so the user doesn't leave Umbraco.

Wouldn't it make sense to have a `target` property on the notification object to compliment it's `url` property?

nul800sebastiaan:
A `target` would be good indeed!

rasmusjp:
The link should probably have a `rel="_noopener"` attribute too (see https://web.dev/external-anchors-use-rel-noopener/).

nul800sebastiaan:
Great minds! I was just thinking the same, you leak some info if you link to external sites.

abjerner:
@rasmusjp does the underscore mean something I don't know about?

I had thought about adding `noreferrer` instead, which also implies `noopener`. If the backoffice links to some external page, it would might be a good idea to hide the referer as well.

rasmusjp:
@abjerner no it’s a typo, the underscore shouldn’t be there.

Yeah `noreferrer` might be better in this case.

Issue #8165, opened 2020-05-25
Image Cropper - uploading same image in 2+ cultures creates only 1 physical file


PerplexDaniel:
## Umbraco version

I am seeing this issue on Umbraco version: 8.6.1

### Bug summary
Uploading a single image file to multiple cultures using an Image Cropper creates only 1 physical file. This leads to issues when removing the media in 1 culture, as it removes the file used by other cultures as well.

The root cause is Umbraco creates some "unique" name for the directory of the media file, but this name is based on (from the top of my head) the content id / property alias / hash of filename. So the culture does not matter, which is why the same file will generate the same directory name for any culture, leading to the exact same physical file in `/media`.

### Steps to reproduce
1. Create an Umbraco site with at least 2 different cultures
2. Add an Image Cropper property
3. Upload the __same image file__ in multiple cultures
4. Observe on disk there is only 1 physical file being created.
5. In Umbraco, clear the image for 1 culture.
6. Observe the media file has been removed from disk
7. Observe the other cultures now give a 404 on the image file.

### Expected result
Expected result after `5.` is that the image file of the specific culture is removed, but other cultures are not affected.

### Actual result
The image file is removed for all cultures, leading to 404 errors in Umbraco when viewing the other cultures.

nul800sebastiaan:
Ah yes.. it would be great to get this fixed and use the culture (if the field is culture variant) in the creation of the directory name. 👍

umbrabot:
Hi @PerplexDaniel,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

Rockerby:
I'll have a crack at this!

Rockerby:
[PR-8208](https://github.com/umbraco/Umbraco-CMS/pull/8208)

Issue #4771, opened 2019-02-27
V8: Allow varying culture for media types


vonbarnekowa:
The new feature for varying the culture for document types is great! Would it be possible to add this feature for media types?

My use case:

We added a text box to the image document type for the copyright. The thing is that the copyright can vary from one language to another.

pralthom:
Hi,

We have the same requirement as we need to have different cultures for the medias.
It would be great if you can add this in a future release.

Thanks

nul800sebastiaan:
Hi! We understand the feature request and have the same wish. This is a longer-term project which we can't work on immediately.

I've put this on the "idea" list for inspiration when we pick up work on the media library in the future.

tom-ingeniuus:
I'd like to add that this would be very useful for adding variant descriptions to be used as alt attributes for images **for accessibility purposes**

ronaldbarendse:
We've used Vorto properties on media in V7 to add alternate texts, so not having culture variant media types and no other alternative (Vorto isn't compatible with V8 and probably won't ever) is a major issue 😢 

Adding the alternate texts within the content as additional property isn't always possible (e.g. when picking multiple images or a folder), makes editing unintuative (no image, but filled text; changing the image doesn't update/reset the text, etc.) and would also mean the text needs to be entered every time it's picked.

I would seriously consider adding this feature to the next minor (8.2.0), most infrastructure already seems in place...

ronaldbarendse:
As a temporary workaround, I'm now just adding a property per culture to the media type, e.g. `alternateText_nl` and `alternateText_fr` (same format as Examine adds culture variant fields to the index). To make accessing the specific values a little easier (and hide this workaround to the front-end), the following custom Models Builder class/extension method can be added:

```c#
using Umbraco.Core.Composing;
using Umbraco.ModelsBuilder;
using Umbraco.Web;

[IgnorePropertyType("alternateText_nl")]
[IgnorePropertyType("alternateText_fr")]
public partial class Image
{
	public string AlternateText => this.AlternateText();
}

public static partial class ImageExtensions
{
	public static string AlternateText(this Image image, string culture = null)
	{
		var alias = "alternateText";
		if (culture == null)
		{
			culture = Current.VariationContextAccessor?.VariationContext?.Culture;
		}
		if (culture != null)
		{
			alias += "_" + culture;
		}

		return image.Value<string>(alias);
	}
}
```

jvtroyen:
[...]

PR #8208, opened 2020-05-30
Store Image Cropper media files in specific culture folders

Rockerby:
### Prerequisites

This is linked to the PR [8165](https://github.com/umbraco/Umbraco-CMS/issues/8165)

### Description

The change passes through the culture of the property down to the IMediaPathScheme to store files in culture specific folders. An example would be `/media/hsudhsie/un-UK/image.jpg`. This will only happen if the site has been setup to be multi-lingual.

Steps to test:

1. Create an Umbraco site with at least 2 different cultures
2. Add an Image Cropper property
3. Make the Image Cropper to vary by culture
4. Upload the same image file in multiple cultures
5. Observe on disk the file is created within a folder named with the culture
6. In Umbraco, clear the image for 1 culture.
7. Observe the media file has been removed from disk for that culture only
8. Observe the other cultures still have their file

cultured_image

Other tests included uploading an image via the Image Cropper when the property is not vary by culture. This stores the image in the usual media folder, e.g. `/media/gagdhd/test.png`.

Switching to multi-cultured after uploading an image has also been tested. This stores the image in the usual media folder, still visible as the path has already been saved. If another image is uploaded it puts it into the culture specific folder.

### DIFF
--- a/src/Umbraco.Core/IO/IMediaFileSystem.cs
+++ b/src/Umbraco.Core/IO/IMediaFileSystem.cs
@@ -22,9 +22,10 @@ namespace Umbraco.Core.IO
         /// <param name="filename">The file name.</param>
         /// <param name="cuid">The unique identifier of the content/media owning the file.</param>
         /// <param name="puid">The unique identifier of the property type owning the file.</param>
+        /// <param name="culture">The culture to save the file against.</param>
         /// <returns>The filesystem-relative path to the media file.</returns>
         /// <remarks>With the old media path scheme, this CREATES a new media path each time it is invoked.</remarks>
-        string GetMediaPath(string filename, Guid cuid, Guid puid);
+        string GetMediaPath(string filename, Guid cuid, Guid puid, string culture = null);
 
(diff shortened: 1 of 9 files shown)


ronaldbarendse:
This is a major breaking change, as you're adding a new parameter to methods of public interfaces (even though they are optional/have a default value, this is binary breaking)! ⚠️ 

Although this is currently only an issue for file uploads/image croppers added to culture variant content, it will become a bigger issue when media also gets variants (https://github.com/umbraco/Umbraco-CMS/issues/4771).

Segments would probably have the same issue, so files will also need to get stored based on segment. It's probably best to include that in the same PR.

Another thing to consider is renaming cultures (or segments), as that's supported since Umbraco 8.3.0 (https://github.com/umbraco/Umbraco-CMS/issues/6503). AFAIK the culture is only used when the media URL is generated, so this might still result in overwriting files if you've stored files in `en-US`, renamed the culture to `en` and then add a new culture with `en-US` again.

PR #8211, opened 2020-05-30
Multi Url picker datatype: a11y fixes

BatJan:
### Prerequisites

### Description
The `<a>` has been converted to a `<button>` and non-visual texts have been added in order to make provide a more specific context to screen reader users instead of just "Add" it will now read the computed label name for the property + "Add url(s)", which is much more specific.


nul800sebastiaan:
Lovely work @BatJan - thanks very much! 👍

ronaldbarendse:
Looks like the localization keys in this PR aren't correct:

general_addUpTo

https://github.com/umbraco/Umbraco-CMS/blob/3441200a7ff0b46e91737115ec50a9f58cadc367/src/Umbraco.Web.UI.Client/src/views/propertyeditors/multiurlpicker/multiurlpicker.html#L50

Other validation messages aren't localized and instead of splitting the texts, it should use placeholders (`%0%`) for the counts (as other languages could use a different format/sentence structure).

ronaldbarendse:
The localization key is fixed in https://github.com/umbraco/Umbraco-CMS/pull/9239 👍 Now only make sure placeholders are used and URL has the correct casing (https://github.com/umbraco/Umbraco-CMS/pull/9080) 😉

Issue #8090, opened 2020-05-11
Saving members without HttpContext throwing errors


runegronkjaer:
When saving members og adding roles to members throws an error when there's no HttpContext available.

## Umbraco version

I am seeing this issue on Umbraco version: 8.5.2
But it does not look like the issue have been fixed in newer versions

Reproduction
------------

### Bug summary

The save of the member or the adding of the role works like it should but one of the events afterwards throws and error.
More specific PerformingIp property on AuditEventsComponent.cs throws a null pointer exception. In the Exception below the error is thrown when trying to add user to roles:

`InnerException: System.NullReferenceException: Object reference not set to an instance of an object.
   at Umbraco.Core.Compose.AuditEventsComponent.get_PerformingIp() in D:\a\1\s\src\Umbraco.Core\Compose\AuditEventsComponent.cs:line 69
   at Umbraco.Core.Compose.AuditEventsComponent.OnAssignedRoles(IMemberService sender, RolesEventArgs args) in D:\a\1\s\src\Umbraco.Core\Compose\AuditEventsComponent.cs:line 109
   at Umbraco.Core.Events.EventDefinition2.RaiseEvent() in D:\a\1\s\src\Umbraco.Core\Events\EventDefinition.cs:line 69
   at Umbraco.Core.Events.QueuingEventDispatcher.ScopeExitCompleted() in D:\a\1\s\src\Umbraco.Core\Events\QueuingEventDispatcher.cs:line 23
   at Umbraco.Core.Events.QueuingEventDispatcherBase.ScopeExit(Boolean completed) in D:\a\1\s\src\Umbraco.Core\Events\QueuingEventDispatcherBase.cs:line 337
   at Umbraco.Core.Scoping.Scope.<>c__DisplayClass71_0.<RobustExit>b__1() in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 434
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions) in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 472
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions) in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 476
   at Umbraco.Core.Scoping.Scope.RobustExit(Boolean completed, Boolean onException) in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 420
[...]

PR #8212, opened 2020-05-30
`IP` null when `GetCurrentRequestIpAddress` called from spoofed context

Rockerby:
### Prerequisites

This fixes issue [8090](https://github.com/umbraco/Umbraco-CMS/issues/8090).

### Description
When calling `httpContext.GetCurrentRequestIpAddress()` from a spoofed HTTP context then the returned IP is `null`. A use case for this would be calling the MemberService.Save from a Hangfire scheduled task.

Testing is tricky; the way I tested was to make these changes locally then reference the DLL in a project where I had this error. Using the current Core DLL I receive the `NullReference` on line 71, using the updated DLL with this patch completes the MemberService.Save completely.

Since IP can return an empty string anyway this won't have any adverse affect.


emmaburstow:
Morning @Rockerby!

Thanks for the PR. We'll take a look and shout if we have feedback. Was fab to 'see' you this weekend. Speak soon,

Emma

nul800sebastiaan:
Lovely, thanks very much @Rockerby! 👍

PR #8224, opened 2020-06-02
umb-notifications.html - added option for setting the target attribute

abjerner:
Following the discussion in https://github.com/umbraco/Umbraco-CMS/pull/8207, I've now added a `target` property to control the corresponding target attribute of the `<a/>` element of the notification.

With this PR, the target will no longer be `_blank` by default, but instead leaves it up to the developer to set so if needed.

The PR also adds `rel="noreferrer"` to the `<a/>` element for security and privacy reasons. While this is primarily for when the link opens in a new window/tab, I don't see a problem always having it on the link.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/notifications.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/notifications.service.js
@@ -52,6 +52,7 @@ angular.module('umbraco.services')
 		* @param {String} item.message longer text for the notication, trimmed after 200 characters, which can then be exanded
 		* @param {String} item.type Notification type, can be: "success","warning","error" or "info" 
 		* @param {String} item.url url to open when notification is clicked
+		* @param {String} item.target the target used together with `url`. Empty if not specified.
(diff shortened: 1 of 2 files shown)


abjerner:
Oh - and for testing the PR, Lee's example in https://github.com/umbraco/Umbraco-CMS/pull/8207 still applies, but with the addition of the `target` property:

```javascript
                notificationsService.add({
                    headline: type,
                    message: "This a message.",
                    url: "/umbraco/#/settings",
                    target: "_blank",
                    type: type,
                    sticky: false
                });
```

nul800sebastiaan:
Thanks @abjerner! This is going to be a bit shitty if we don't include it in 8.7.0, so I'll cherry pick it there as well, otherwise changing the behavior would be a "breaking" change. All good though! 👍

nul800sebastiaan:
Cherry picked for 8.7.0 in 16769e95043e460b3c3e2c9dcbc350eb0f3dcd7b

PR #8227, opened 2020-06-03
V8/bugfix/7981 limiting password length

### LINKED ISSUE #7981: Limiting max char length for member passwords
Currently there is no max length enforced for member passwords, and that is in many ways a good thing, as long passwords are more secure.

However it is currently possible in Umbraco to set a password that is for example 100k characters long which can become a resource issue when the membership provider tries to hash such a long string.

A good solution would be some server side validation that would kick in an return an error if you try to set too long a password. 256 chars could be a suggested limit as described in https://www.troyhunt.com/passwords-evolved-authentication-guidance-for-the-modern-era/

redmorello:
I've added an upper password limit to the umb-login form and also the Members password reset form in the backoffice.

Nothing added to the Membership Provider, I tried but it was not possible due to the existing provider.

Fixes #7981


nul800sebastiaan:
Thanks @redmorello! We will need to reject longer passwords on the server as well, this is where all the processing is being done and where people with malicious intent can try to overwhelm the server. We should do absolutely no hashing when the password is longer than this. Can you look into adding that as well please? 
I think that should be the same for the member membership provider? Not sure.

redmorello:
@nul800sebastiaan I haven't forgotten, just had a massive increase in workload the last few weeks. It's still on my todo list 👍

redmorello:
@nul800sebastiaan spent some time over the weekend changing the membership provider, and updating some tests.

src/Umbraco.Core/Security/MembershipProviderBase.cs
```diff
@@ -296,7 +316,7 @@ public override void Initialize(string name, NameValueCollection config)
         /// <param name="e"></param>
         protected override void OnValidatingPassword(ValidatePasswordEventArgs e)
         {
-            var attempt = IsPasswordValid(e.Password, MinRequiredNonAlphanumericCharacters, PasswordStrengthRegularExpression, MinRequiredPasswordLength);
+            var attempt = IsPasswordValid(e.Password, MinRequiredNonAlphanumericCharacters, PasswordStrengthRegularExpression, MinRequiredPasswordLength, MinRequiredPasswordLength);
```
nathanwoulfe, line 319: Shouldn't this be MaxRequiredPasswordLength?

src/Umbraco.Core/Security/MembershipProviderBase.cs
```diff
@@ -249,6 +268,7 @@ public override void Initialize(string name, NameValueCollection config)
             _maxInvalidPasswordAttempts = GetIntValue(config, "maxInvalidPasswordAttempts", 5, false, 0);
             _passwordAttemptWindow = GetIntValue(config, "passwordAttemptWindow", 10, false, 0);
             _minRequiredPasswordLength = GetIntValue(config, "minRequiredPasswordLength", DefaultMinPasswordLength, true, 0x80);
+            _maxRequiredPasswordLength = 256;
```
nathanwoulfe, line 271: Could the default value here come from a constant? It's reused in a few places so would be safer

src/Umbraco.Core/Security/MembershipProviderPasswordValidator.cs
```diff
@@ -27,7 +27,7 @@ public override async Task<IdentityResult> ValidateAsync(string item)
             var result = await base.ValidateAsync(item);
             if (result.Succeeded == false)
                 return result;
-            var providerValidate = MembershipProviderBase.IsPasswordValid(item, Provider.MinRequiredNonAlphanumericCharacters, Provider.PasswordStrengthRegularExpression, Provider.MinRequiredPasswordLength);
+            var providerValidate = MembershipProviderBase.IsPasswordValid(item, Provider.MinRequiredNonAlphanumericCharacters, Provider.PasswordStrengthRegularExpression, Provider.MinRequiredPasswordLength, 256);
```
nathanwoulfe, line 30: Related to earlier - can this be constant?

src/Umbraco.Tests/Membership/MembershipProviderBaseTests.cs
```diff
@@ -150,6 +150,7 @@ public void Sets_Defaults()
             Assert.AreEqual(5, provider.MaxInvalidPasswordAttempts);
             Assert.AreEqual(10, provider.PasswordAttemptWindow);
             Assert.AreEqual(provider.DefaultMinPasswordLength, provider.MinRequiredPasswordLength);
+            Assert.AreEqual(provider.DefaultMaxPasswordLength, 256);
```
nathanwoulfe, line 153: Related to earlier - can this be constant?

nathanwoulfe:
@redmorello, appreciating this is a pretty old PR, it's still a beneficial change. I've left a couple of comments.

Will re-target to v8/contrib and deal with the conflicts.

nathanwoulfe (changes requested):

PR #8229, opened 2020-06-04
Add config to define true and false value

bjarnef:
### Prerequisites

### Description
At the moment the boolean (toggle) property editor (previous checkbox) store the value as 0 and 1.
However when re-using the property editor in e.g. third party packages, a custom dashboard or similar it would be great to set another true/false value similar to `ng-true-value` and `ng-false-value`.

Currently the boolean property editor only works with 0 and 1 value, but it would be great to support `True` and `False` as well. Maybe sometime later we can also make it work with e.g. `On/Off`, `Yes/No` as it is possible with `ng-true-value` and `ng-false-value` in Angular.

E.g. in [Vendr](https://vendr.net/) for Umbraco in its order.editor.config.js in allow to re-use existing property editors for order and order line properties, e.g.

```
{
    additionalInfo: [
        { alias: "ipAddress", label: "IP Address", isReadOnly: true },
        { alias: "notificationSent", label: "Notification Sent", isReadOnly: false, view: "boolean" }
    ]
}

```

With this PR it would be possible to use the boolean property editor for a property with a `true/false` value.

```
{
    additionalInfo: [
        { alias: "ipAddress", label: "IP Address", isReadOnly: true },
        { alias: "notificationSent", label: "Notification sent", isReadOnly: false, view: "boolean", config: { truevalue: "True", falsevalue: "False" } }
    ]
}
```

Besides this it would also allow to toggle a `true/false` property value in a custom dashboard or similar re-using the `boolean` property editor.

To test this PR unzip the following zip-file and put the folder "MyDashboard" in App_Plugins folder.
[MyDashboard.zip](https://github.com/umbraco/Umbraco-CMS/files/4729242/MyDashboard.zip)


emmaburstow:
Good morning Bjarne, 

Thanks, as ever, for the work here. I like this! This is a fantastic improvement. We'll take a look and if there's any questions, we'll be sure to ask.

Emma

bjarnef:
@nielslyngsoe what do you think of this?

nul800sebastiaan:
Updated to `Utilities.extend` instead of `angular.extend` - all good to go! Thanks @bjarnef! 👍

bjarnef:
Thanks again @nul800sebastiaan
I am sure @mattbrailsford will love this for configuration in Vendr 😎👍🎉

nielslyngsoe:
@bjarnef great!

PR #8220, opened 2020-06-02
Netcore: Migration of more controllers

bergmania:
# Notes

- Migrated the following controllers:
    - `DictionaryController`
    - `TinyMceController`
    - `UpdateCheckController`
- I updated a lot of wrong namespaces
- I added `IWebSecurity` to the DI container, so we can inject `IWebSecuity` directly instead of injecting `IUmbracoContextAccessor` and use `_umbracoContextAccessor.UmbracoContext.Security`

# Tests
We currently have issues when returning a forbidden request. So to be able to test, start by commenting out the `context.Result = new ForbidResult();` in `UmbracoTreeAuthorizeFilter.OnAuthorization`.
- `DictionaryController`
   - Setup dictionary items in `netcore/dev` (Framework executable) and use the same database when you test this branch (dotnetcore executable)
    - https://localhost:44354/umbraco/backoffice/UmbracoApi/Dictionary/DeleteById?id=1
    - https://localhost:44354/umbraco/backoffice/UmbracoApi/Dictionary/Create?parentId=1&key=test
    - https://localhost:44354/umbraco/backoffice/UmbracoApi/Dictionary/GetById?id=1
    - https://localhost:44354/umbraco/backoffice/UmbracoApi/Dictionary/PostSave
    - https://localhost:44354/umbraco/backoffice/UmbracoApi/Dictionary/GetList

- `TinyMceController`
    - https://localhost:44354/umbraco/backoffice/UmbracoApi/TinyMce/UploadImage 
    - I used postman 
image

- `UpdateCheckController`
    - https://localhost:44354/umbraco/backoffice/UmbracoApi/UpdateCheck/GetCheck 

Related to [AB#6587](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6587)

PR #8230, opened 2020-06-04
Netcore: Removing UmbracoContext.Security usages

elit0451:
## Notes:

Once `IWebSecurity` was added to the DI Container (#8220) we can now directly inject `IWebSecurity` and eliminate the usages of `UmbracoContext.Security`.


bergmania:
Hi @elit0451.

Good work. I fixed the failing test. It failed because the mock of the `ServiceProvider` was not setup to return `IWebSecturity` when requested, but only `IUmbracoContext`. Thats also way you had to add a null check. Changes can be found here: https://github.com/umbraco/Umbraco-CMS/pull/8230/commits/6a560d634059c4146b9586a66816a0b13872ed2f

PR #8231, opened 2020-06-05
no more 401 when trying to create content where no permitted types exist

### LINKED ISSUE #8062: User gets 401 Unauthorized if no root  start node is set
When a user with access to the Content section (apart from the administrator) clicks on the 3 dots next to the tree root name (**Content**) to check if there is a root node that can be created, gets thrown out of the back-office with **401 Unauthorized** error. This is happening because no Document type was set as a start root node. After that, if the user tries to login again, the Unauthorized error keeps preventing the login and the cache and cookies need to be cleared for a successful login.

## Umbraco version

Tested on the latest v8/dev branch, also on 8.6.1
This is not a problem in v7.

### Bug summary
Getting 401 because there isn't any start root node set. Can't login after getting the error.

### Steps to reproduce
1. Install a clean version of Umbraco.
2. Create a user that has the default 'Writer' or 'Editor' group.
3. Login as the user.
4. Click on the 3 dots next to the tree root name (Content).
5. After being thrown out of the back-office, try to login again.

### Expected result
No Unauthorized error. The result should be similar to what we get with an Admin user in the same scenario. We should see a column that opens with the text:
> Create an item under Content
There are no allowed document types available...

### Actual result
A 401 Unauthorized error when clicking on the 3 dots. Can't login again without clearing the cache and cookies.

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #8062

### Description
Issue here was caused by calling `getCount` on `contentTypeResource` as the API method is restricted to users with access to the document types tree, so returns the 401 for editor/writer user types.

I've fixed it by refactoring `Umbraco.Editors.Content.CreateController` to make better use of `$q.all` - we now wait for all promises to resolve, which gives us access to the currentUser object, so can check the user's permissions before calling getCount. Users without appropriate permissions see the basic message "The selected page in the content tree doesn't allow for any pages to be created below it.".

Verify by first logging in as an admin and ensuring no doc types are permitted at root. Log in as writer/editor, and attempt to create at the root (ie click ... next to content). Should not be logged out, but should see the text message re no permitted content.

Because I can't help myself, I cleaned up the `createOrSelectBlueprintIfAny` method in the same controller - removes the underscore dependency in favor of boring old vanilla javascript and removes some nesting by returning early when no blueprints exist for the doctype.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/content.create.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/content.create.controller.js
@@ -28,7 +28,9 @@ function contentCreateController($scope,
         
         $q.all([contentTypeResource.getAllowedTypes($scope.currentNode.id), authResource.getCurrentUser()])
             .then(resolvedPromises => {
-                [let types, let currentUser] = resolvedPromises;
+                let types;
+                let currentUser;
+                [types, currentUser] = resolvedPromises;
             
                 $scope.allowedTypes = iconHelper.formatContentTypeIcons(types);                
             
--- a/src/Umbraco.Web.UI.Client/src/views/content/content.create.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/content.create.controller.js
@@ -22,33 +22,32 @@ function contentCreateController($scope,
     function initialize() {
         $scope.loading = true;
         $scope.allowedTypes = null;
-        
-        var getAllowedTypes = contentTypeResource.getAllowedTypes($scope.currentNode.id).then(function (data) {
-            $scope.allowedTypes = iconHelper.formatContentTypeIcons(data);
-            if ($scope.allowedTypes.length === 0) {
-                contentTypeResource.getCount().then(function(count) {
-                    $scope.countTypes = count;
-                });
-            }
-        });
-        var getCurrentUser = authResource.getCurrentUser().then(function (currentUser) {
-            if (currentUser.allowedSections.indexOf("settings") > -1) {
-                $scope.hasSettingsAccess = true;
-                if ($scope.currentNode.id > -1) {
-                    contentResource.getById($scope.currentNode.id).then(function (data) {
-                        $scope.contentTypeId = data.contentTypeId;
-                    });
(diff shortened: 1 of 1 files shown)


nathanwoulfe:
Annnnd looks like I have some tests to fix

src/Umbraco.Web.UI.Client/src/views/content/content.create.controller.js
```diff
@@ -73,25 +74,24 @@ function contentCreateController($scope,
...
-            if (blueprintConfig.skipSelect) {
-                createFromBlueprint(blueprints[0].id);
-            } else {
-                $scope.selectContentType = false;
-                $scope.selectBlueprint = true;
-                $scope.selectableBlueprints = blueprints;
-            }
-        } else {
+
+        const keys = Object.keys(docType.blueprints);
```
lars-erik, line 79: You removed a null propagation.

nathanwoulfe: Yup, it wasn't required - blueprints will be an empty object if none exists for the doctype.

I think the failing test is related to using q to resolve the API requests and/or destructing that response. Test fails when allowedTypes is null, so looks to me like that test case is being asserted before the promises are resolved.

lars-erik: Ah. Need $httpBackend.$flush() and/or $rootScope.$appy()?
Otherwise might be that the test lacks a "done" parameter to do the "old-style" async wait in the test.
Haven't got time to dig in, just glanced at the code from the phone. ☺

emmaburstow:
Hey Nathan,

Thanks for the work! 🙌 Speak soon,

Emma

nathanwoulfe:
Hello @emmaburstow 
I still haven't been able to sort the test failures... Trying, not succeeding. Sad face.

nathanwoulfe:
Looks like this is no longer required - this issue it was solving is no longer happening on v8/contrib.

PR #8242, opened 2020-06-07
Netcore: Migrated some tree related classes

AndyButland:
In this PR I've made a start on migration of trees to .NET Core.  Have just picked the base controller classes, looked to migrate them, and brought over various related attributes and extension methods.

No testing done other than making sure everything builds.

There's one point I'm stuck on progressing (will add an inline comment) and I've got a fairly busy week lined up so could be I won't be too responsive on taking it further, so if it's on the core road-map, feel free to just use and progress anything that looks useful from here (if of course there is).

### DIFF
--- a/src/Umbraco.Core/Trees/ITree.cs
+++ b/src/Umbraco.Core/Trees/ITree.cs
@@ -3,7 +3,7 @@
     // TODO: we don't really use this, it is nice to have the treecontroller, attribute and ApplicationTree streamlined to implement this but it's not used
     // leave as internal for now, maybe we'll use in the future, means we could pass around ITree
     // TODO: We should make this a thing, a tree should just be an interface *not* a controller
-    internal interface ITree
+    public interface ITree
     {
         /// <summary>
         /// Gets or sets the sort order.
--- a/src/Umbraco.Core/Trees/Tree.cs
+++ b/src/Umbraco.Core/Trees/Tree.cs
@@ -45,7 +45,7 @@ namespace Umbraco.Web.Trees
         /// </summary>
         public Type TreeControllerType { get; }
 
-        internal static string GetRootNodeDisplayName(ITree tree, ILocalizedTextService textService)
+        public static string GetRootNodeDisplayName(ITree tree, ILocalizedTextService textService)
         {
             var label = $"[{tree.TreeAlias}]";
 
--- a/src/Umbraco.Web.BackOffice/Trees/TreeController.cs
+++ b/src/Umbraco.Web.BackOffice/Trees/TreeController.cs
@@ -1,48 +1,38 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using Microsoft.Extensions.DependencyInjection;
 using Umbraco.Core;
-using Umbraco.Core.Cache;
-using Umbraco.Core.Configuration;
-using Umbraco.Core.Logging;
-using Umbraco.Core.Mapping;
-using Umbraco.Core.Persistence;
 using Umbraco.Core.Services;
-using Umbraco.Web.Routing;
+using Umbraco.Web.Trees;
 
-namespace Umbraco.Web.Trees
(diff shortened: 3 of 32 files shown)


src/Umbraco.Core/Trees/ITree.cs
```diff
@@ -3,7 +3,7 @@
     // TODO: we don't really use this, it is nice to have the treecontroller, attribute and ApplicationTree streamlined to implement this but it's not used
     // leave as internal for now, maybe we'll use in the future, means we could pass around ITree
     // TODO: We should make this a thing, a tree should just be an interface *not* a controller
-    internal interface ITree
+    public interface ITree
```
AndyButland, line 6: Have made this public such that its accessible to `Umbraco.Web.BackOffice`.  We could use `InternalsVisibleTo` here if it's preferred not to make this public.

src/Umbraco.Core/Trees/Tree.cs
```diff
@@ -45,7 +45,7 @@ public Tree(int sortOrder, string applicationAlias, string group, string alias,
         /// </summary>
         public Type TreeControllerType { get; }
 
-        internal static string GetRootNodeDisplayName(ITree tree, ILocalizedTextService textService)
+        public static string GetRootNodeDisplayName(ITree tree, ILocalizedTextService textService)
```
AndyButland, line 48: Similarly here - have made this public such that its accessible to `Umbraco.Web.BackOffice`.  We could use `InternalsVisibleTo` here if it's preferred.

src/Umbraco.Web.BackOffice/Trees/TreeControllerBase.cs
```diff
@@ -0,0 +1,399 @@
...
+    /// </summary>
+    /// <remarks>
+    /// Developers should generally inherit from TreeController.
+    /// </remarks>
+    [TypeFilter(typeof(AngularJsonOnlyConfigurationAttribute))]
+    public abstract class TreeControllerBase : UmbracoAuthorizedApiController, ITree
+    {
+        // TODO: Need to set this, but from where?
+        //       Presumably not injecting as this will be a base controller for package/solution developers.
+        private readonly UmbracoApiControllerTypeCollection _apiControllers;
```
AndyButland, line 30: This is the point I'm stuck on - where to do get this from?  It's needed to be passed into the `GetUmbracoApiService` method in `UrlHelperExtensions`.

src/Umbraco.Web.BackOffice/Trees/TreeController.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        private static readonly ConcurrentDictionary<Type, TreeAttribute> _treeAttributeCache = new ConcurrentDictionary<Type, TreeAttribute>();
+
+        private readonly TreeAttribute _treeAttribute;
+
+        private readonly ILocalizedTextService _textService;
+
+        protected TreeController()
+        {
+            var serviceProvider = HttpContext.RequestServices;
```
AndyButland, line 23: Wasn't too sure about this - should we support a parameterless constructor to make things easier for package/solution developers (and perhaps other core trees)?  Or is it better to expect them to use a constructor where the dependencies (i.e. `ILocalizedTextService`) are provided?

src/Umbraco.Core/Trees/ITree.cs
bergmania: When used by our own non-test dlls, we prefere it to be public :)

src/Umbraco.Web.BackOffice/Trees/TreeController.cs
bergmania: For now, we remove the parameterless constructors. These make testing harder, and package developers should be interested in having testable code. That said, maybe we reintroduce them later. For now, it is most important, that we do not use these untestable parts ourselves.

PR #8246, opened 2020-06-08
update date-format matching for formats with am/pm

### LINKED ISSUE #8241: Having Datepicker Show time with AM/PM

This is purely a back office issue:

You can change the "format" of the Date Format field in the configuration of the Date/Time Field type. It says on https://momentjs.com that "h:mm:ss a" is the format.. But it shows the calendar. It does not seem to display the picker correctly in this case..

## Umbraco version

Version: 8.6.1

Reproduction
------------
see above

### Expected result

I would expect it to popup a time editor with 3 fields Hour, Minute and am/pm
### Actual result

Shows the calendar with the time at the bottom even though the date format string only has time.

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #8241 

### Description
I've updated the date-format string check to catch all (most) time variants, replacing the old string comparisons with a regex match to catch lots more for less effort. Also updated the view to show clock icon when the picker is time-only, rather than the calendar. This won't catch absolutely every time format, but it's a lot broader than the previous solution which forced 24-hour formatting with a leading 0

Verify by updating the date picker config to `h:mm:ss a` or any combination of 1-2 `H/h`, 1-2 `m`, 1/2 `s`, all colon-separated, with or without a trailing `a`, with or without a leading space (aka any valid [momentjs time format](https://momentjs.com/docs/#/displaying/format/)). Picker will not display the calendar, and will reformat times with respect to am/pm when the format starts with `h`.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/datepicker.controller.js
@@ -55,7 +55,9 @@ function dateTimePickerController($scope, angularHelper, dateHelper, validationM
         };
 
         // Don't show calendar if date format has been set to only time
-        if ($scope.model.config.format === "HH:mm:ss" || $scope.model.config.format === "HH:mm" || $scope.model.config.format === "HH") {
+        const timeFormat = $scope.model.config.format.toLowerCase();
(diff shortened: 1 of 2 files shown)


Rizen-Kris:
Any updates on this? It needs an approving review

nul800sebastiaan:
Very nice! Good Regex skills @nathanwoulfe 👍

PR #8250, opened 2020-06-08
Use property editor width limit for slider property editor

bjarnef:
### Prerequisites

### Description
This PR re-use the `umb-property-editor--limit-width` mixin to set the max width and not specific to `66.6%` width also mean it in overlay use more of the width instead only 2/3 of the width in overlays.

**Before**

image

**After**

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/property-editors.less
+++ b/src/Umbraco.Web.UI.Client/src/less/property-editors.less
@@ -707,8 +707,9 @@
             margin-top: 0;
         }
     }
+
 //
-// folder-browser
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
So much nicer, can't believe I never noticed! Thanks @bjarnef!

bjarnef:
@nul800sebastiaan I noticed you have added the `category/modelsbuilder` label to this. It hasn't much to do with ModelsBuilder though 😁

nul800sebastiaan:
lol! not sure where that came from, thanks!

PR #8253, opened 2020-06-08
Fix checkbox and radio label alignment

BatJan:
### Prerequisites

### Description
As pointed out by @bjarnef in #8119 I did unfortunately not get to the root of the problem with the proposed PR, which affected the display of radio buttons making them look worse and the rich text editor settings did not look to good either - They did not look good before but looked worse after, which was not the intent 😅 

With this PR I managed to address the root of the problem and there were several issues that needed to be fixed styling wise. Now both radio buttons and checkboxes look good 👍 

### Before
_Relation types_
radio-relate-before

_Rich text editor_
rte-before

_Radio- and checkboxes_
radio-checkboxes-before

### After
_Relation types_
radio-relate-after

_Rich text editor_
rte-after

_Radio- and checkboxes_
radio-checkboxes-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-form-check.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-form-check.less
@@ -1,25 +1,18 @@
 @checkboxWidth: 18px;
 @checkboxHeight: 18px;
-label.umb-form-check--checkbox{
-    margin:3px 0;
-}
 
 .umb-form-check {
     display: flex;
+    align-items: center;
     position: relative;
-    padding-left: 0px;
-    margin: 0;
-    min-height: 22px;
+    padding-left: 0;
+    margin: 5px 0;
+    min-height: 20px;
     cursor: pointer !important;
 
-    .umb-form-check__symbol {
-        margin-top: 4px;
-        margin-right: 10px;
-    }
     .umb-form-check__info {
-        margin-left:20px;
+        margin-left: 30px;
         position: relative;
-        top: 3px;
     }
     
 
@@ -81,7 +74,6 @@ label.umb-form-check--checkbox{
     }
 
     .tabbing-active &.umb-form-check--radiobutton &__input:focus ~ .umb-form-check__state .umb-form-check__check {
-        //outline: 2px solid @inputBorderTabFocus;
         border: 2px solid @inputBorderTabFocus;
         margin: -1px;
     }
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Looking great! Thanks @BatJan! 👍

bjarnef:
It seems this affect some dialogs like nodepermissions, where it always add left and right padding to these labels.

image

image

BatJan:
@bjarnef I don't think that's the culprit - All the :not selector does is to ensure the label styling is not applied to labels containing the `.umb-form-check` so `<label>` or `<label class="anything-other-than-umb-form-check>` will be hit by the general label rule but `<label class="umb-form-check">` won't - I'm not sure what has changed when and where but it's not this part of the code I can assure you 😃

bjarnef:
@BatJan as in the screenshots it is clear that this styling comes from this this line https://github.com/umbraco/Umbraco-CMS/pull/8253/files#diff-7f35832f3c1700489408cc3a179e1802R32

However it seems the following (previous) styles would also target this label. Maybe something else previous overwrite this since I haven't noticed this previous.

```
.controls-row label{padding: 0 10px 0 10px; vertical-align: middle;}
```

I also noticed some of this in placed in `forms.less` and in `main.less`. Would be great to keep this together.

```
.controls-row label:not(.umb-form-check) {
    display: inline-block;
    padding: 0 10px 0 10px;
    vertical-align: middle;
}
```

BatJan:
@bjarnef In your screendump you disable the styling. You need to remove the :not() part of the selector - When it is removed the result is still the same. So the issue is not related to this change since the `.control-row label` selector has not been introduced here. It's only been modified to ensure that if the targeted label contains the `.umb-form-check` class then the styling will not apply making sure the rule won't affect the checkbox directive.

In the below screendump you can see the issue still remain after removing the `:not()` part of the selector. Adding the :not() selector is the only modification I have done with this PR.

without-not-selector

BatJan:
@bjarnef Seems like the use of the `.controls-row` class is the issue and it does not seem to be necessary - Made PR fixing the alignment issue here https://github.com/umbraco/Umbraco-CMS/pull/9069

bjarnef:
@BatJan there are two screendumps in https://github.com/umbraco/Umbraco-CMS/pull/8253#issuecomment-703254642
The first one shows how it currently looks, the second one shows how it looks when disabling this styling :)

I noticed the styling, but when looking at the history it seems it previous had this:

```
.controls-row label{padding: 0 10px 0 10px; vertical-align: middle;}
```

I guess somehow else might has regarding labels - maybe something overwrite this padding in dialogs.

It actually seems it was affected by a change from `<div>` in v8.7.0 to `<label>` in v8.8.0 

image

bjarnef:
It seems to be related to this instead https://github.com/umbraco/Umbraco-CMS/commit/c81266f5dd9c7181491986a42102a325f7970df7#diff-554f0d5110d05a4f16f86254fd35631dR10 where a `<label>` element was added and since the `<label>` with ancestor with `.controls-row` by default has left/right padding.

Issue #8256, opened 2020-06-09
Raise an event when deleting a document type to its dependents


thomasandresen:
If a document type is allowed as a child in other document types, and the document type is being deleted, the dependents gets updated in the database and data is consistent. However since there is no event being triggered that the dependent document type has been changed, other won't be able to react on such change, e.g. when using deploy.

Fixes umbraco/Umbraco.Cloud.Issues#66

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8257

PR #8257, opened 2020-06-09
If content type is deleted and other document types have a reference to it, raise an event

thomasandresen:
Fixes umbraco/Umbraco-CMS#8256

When a document type is being deleted, all document types are being iterated to lookup if they have a reference to the document type. If it has, simply update the AllowedContentTypes to reflect the change.

To test:
- Compile the CMS and copy the DLL's from deploy project.
- Run the solution
- Add three document types (dta, dtb, dtc)
- Let the third document types (dtc) be Allowed as child for each of the other two (dta and dtb)
- Locate the uda files and verify that for both dta and dtb, that there is a reference to dtc
- Delete dtc
- Locate the uda files and verify that for both dta and dtb, that the reference to dtc has been deleted.


warrenbuckley:
Code reads fine & works as described

PR #8263, opened 2020-06-09
Use code snippet component for insert field

bjarnef:
### Prerequisites

### Description
This PR re-use the `umb-code-snippet` component which is also used in querybuilder to show the code snippet when inserting af field value to make it a bit more consistent.


nul800sebastiaan:
Cool! Thanks Bjarne!

PR #8266, opened 2020-06-10
Allows unpublished content to render a macro HTML in the RTE

warrenbuckley:
Fixes https://github.com/umbraco/Umbraco.Forms.Issues/issues/354
However, this is a core CMS issue as opposed to a Forms issue

## Test Notes
* Test before & after PR applied to verify it works
* Have Umbraco Forms installed
* Create a doctype with one property that is a RTE that allows macros to be inserted
* Create new content node and use the Forms Macro in the RTE (Do NOT publish it. only save it)
* Reload/come back to the node & see the error or verify its gone all OK & the macro can render in the RTE

### DIFF
--- a/src/Umbraco.Web/Editors/MacroRenderingController.cs
+++ b/src/Umbraco.Web/Editors/MacroRenderingController.cs
@@ -142,7 +142,7 @@ namespace Umbraco.Web.Editors
                 //need to create a specific content result formatted as HTML since this controller has been configured
                 //with only json formatters.
                 result.Content = new StringContent(
-                    _componentRenderer.RenderMacro(pageId, m.Alias, macroParams).ToString(),
+                    _componentRenderer.RenderMacroForContent(publishedContent, m.Alias, macroParams).ToString(),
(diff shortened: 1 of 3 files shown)


warrenbuckley:
@nul800sebastiaan marked this for 8.7.0

nul800sebastiaan:
Make sure to give it a `type` and let the original bug reporter know what's happening, now it just looks like you closed their issue with no reason why.

nul800sebastiaan:
Cherry picked for 8.6.3 in commit c99798ac1b9e236cd56a4375f41ea7738eb30a0c

StuartPaterson:
I still appear to be seeing this error with - 

Umbraco v8.13.0 and Umbraco Forms v8.7.6

Also posted on the community (https://our.umbraco.com/forum/using-umbraco-and-getting-started//106758-umbraco-forms-v876-webviewpage-error)

Should the renderUmbracoForm error be resolved in the above versions? I was doing the same, just inserting the macro in an RTE when I get the error.

PR #8268, opened 2020-06-10
Fix radio button checked appearance

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8175 (side effects on radio button - [reference here](https://github.com/umbraco/Umbraco-CMS/pull/8175#issuecomment-641404313))

### Description

#8175 had the unintended side effect of messing up the radio button checked appearence:

image

This PR turns the radio button checked state back to what it should be:

image

...while retaining fix for the checkbox checked state originally intended by #8175:

image

Please note that this PR targets `v8/dev` on purpose as #8175 was merged into that branch.


clausjensen:
Awesome! 😃

Issue #8238, opened 2020-06-05
V8: Repeatable textstring property editor saves whitespaces


ReVoid:
If get try to enter whitespace to `Repeatable textstring` property editor as the second row, it saves whitespaces, but UI does not show empty entries

## Umbraco version

Umbraco 8.6.2

Reproduction
------------
![1](https://user-images.githubusercontent.com/12345710/83892422-4e8bce80-a757-11ea-8a47-e8c2ffdfdfcf.gif)
![image](https://user-images.githubusercontent.com/12345710/83892524-75e29b80-a757-11ea-9495-dfeaff98e366.png)

### Steps to reproduce

1. Create a property of `Repeatable textstring` data type
2. Enter something in the **first** row
3. Enter whitespace in the **second** row or others
4. Click "Save and publish"
5. Check the result in your razor view or another place

### Expected result
Whitespaces would not be saved, or, as an alternative, UI shows empty entries.

### Actual result
The entry with whitespace saved, but not shown.

nul800sebastiaan:
Yeah that's right, we really don't need to save the value if it's just whitespaces! We'd love a little help to get this fixed if you (or someone else can).

umbrabot:
Hi @ReVoid,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8269

PR #8269, opened 2020-06-10
Remove empty strings from repeatable textstrings whitespace

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8238

### Description
This PR excludes empty string values from saved model, so the underlying json doesn't contain empty values, which made the property value converter return an array with empty values.


nul800sebastiaan:
Very lovely update, thanks very much @bjarnef!

Issue #8259, opened 2020-06-09
PropertyValueConverterBase contains unused HasValue method


ronaldbarendse:
The `PropertyValueConverterBase` contains an unused `HasValue` method (that's not a member of the `IPropertyValueConverter` interface): https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Core/PropertyEditors/PropertyValueConverterBase.cs#L25-L31

This method can probably be obsoleted (or removed, but that would be a breaking change, as it's a public, virtual method). The same code is already used in the `IsValue` method: https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Core/PropertyEditors/PropertyValueConverterBase.cs#L14-L23

The same code is also used in `PublishedPropertyType`, but that returns a boolean for every type of `PropertyValueLevel` (not the same as the code above, that only returns true/false for the source value): https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Core/Models/PublishedContent/PublishedPropertyType.cs#L179-L190

ronaldbarendse:
Also, according to the documentation the `IsValue` method should return `null` for levels it can't determine: https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Core/PropertyEditors/IPropertyValueConverter.cs#L20-L30

But `PropertyValueConverterBase` currently throws a `NotSupportedException` for the `Inter` and `Object` types!

nul800sebastiaan:
What is the result of having an unused method? Is it in the way? People can inherit that class and use it? I don't really understand the problem.

ronaldbarendse:
The biggest problem is probably confusion, as developers might override the `HasValue` (that doesn't do anything) instead of the `IsValue` method. This might cause bugs and/or wasted development time, which can be easily prevented by adding an `Obsolete` attribute indicating this.

Besides that, the `IsValue` method also throws exceptions when it shouldn't, so that's definitely a bug (the documentation explicitly states it should return `null` and the exception message indicates it should only be thrown on invalid enum values, not unhandled ones).

PR #8270, opened 2020-06-10
PropertyValueConverterBase cleanup: fix IsValue, obsolete HasValue, add documentation

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8259.

### Description
This PR cleans up the `PropertyValueConverterBase` by:
- Fixing `IsValue`, so it doesn't throw `NotSupportedException` for valid enum values, but returns `null` when it can't determine whether it's an actual value;
- Obsoleting `HasValue`, as it's not included in the `IPropertyValueConverter` contract and might confuse developers overriding this unused method;
- Simplifying the value check using type pattern matching (`value is string stringValue`);
- Adding XML documentation

### DIFF
--- a/src/Umbraco.Core/PropertyEditors/PropertyValueConverterBase.cs
+++ b/src/Umbraco.Core/PropertyEditors/PropertyValueConverterBase.cs
@@ -31,7 +31,7 @@ namespace Umbraco.Core.PropertyEditors
             }
         }
 
-        [Obsolete("This method is not part of the IPropertyValueConverter contract and therefore not used, use IsValue instead.")]
+        [Obsolete("This method is not part of the IPropertyValueConverter contract, therefore not used and will be removed in future versions; use IsValue instead.")]
         public virtual bool HasValue(IPublishedProperty property, string culture, string segment)
         {
             var value = property.GetSourceValue(culture, segment);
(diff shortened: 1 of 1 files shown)


src/Umbraco.Core/PropertyEditors/PropertyValueConverterBase.cs
```diff
@@ -4,44 +4,57 @@
...
+                case PropertyValueLevel.Inter:
+                    return null;
+                case PropertyValueLevel.Object:
+                    return null;
                 default:
                     throw new NotSupportedException($"Invalid level: {level}.");
             }
         }
 
+        [Obsolete("This method is not part of the IPropertyValueConverter contract and therefore not used, use IsValue instead.")]
```
Shazwazza, line 34: Might be good to also mention that this will be removed in later versions since we'll prob just remove it in the netcore branch.

ronaldbarendse: Just by marking it obsolete [implicitly informs uses that the element will be removed in future versions of the product](https://docs.microsoft.com/en-us/dotnet/api/system.obsoleteattribute#remarks). But we can always change the message to make the more clear:

```suggestion
        [Obsolete("This method is not part of the IPropertyValueConverter contract, therefore not used and will be removed in future versions; use IsValue instead.")]
```

Issue #4419, opened 2019-02-05
Add property in package.manifest to re-use property value converters for simple types


bjarnef:
At the moment when you have a custom property editors and are working with either strongly typed models or using ModelsBuilder in Visual Studio, one has to implement a property value converter to return a specific type. It does make sense in many cases e.g. when storing json and one want to convert this to a specific model.

However as a frontend developer this might be an extra step and the cool thing about property editors in Umbraco v7 is that you can do a lot just with html, css and javascript without knowing about C#.

For example in my own package Switcher https://our.umbraco.org/projects/backoffice-extensions/switcher/ I can implement a property value converter to return either false or true (while ModelsBuilder at the moment just return an object). However it might be a bit overkill to add an extra assembly just for simple types and checkbox property editor already has a converter for this to return true/false from a stored int 0/1 - it would be cool if you just could re-use built-in property value converters in Umbraco core for simple types:

- String
- Int32
- Boolean
- (decimal/double/float) - e.g. for property editors allowing to store decimal values, like Umbraco.Decimal property editor.
- (DateTime)
- (String/int array) - e.g. as Umbraco.MultipleTextstring

I imagine the package.manifest could have a property for return type or output type like in the screenshot. https://our.umbraco.org/documentation/extending/property-editors/package-manifest

Original issue: https://issues.umbraco.org/issue/U4-10005

nul800sebastiaan:
Sounds like a good idea, I'll have to discuss with the team soon, will get back to you!

bjarnef:
Thanks @nul800sebastiaan 
It seems a bit overkill that each smaller packages / property editors should implement there own property value converter and add an additional assembly to the project or the developer should implement a property value converter like this https://our.umbraco.com/packages/backoffice-extensions/switcher/bugs-feedback-and-suggestions/95608-using-switcher-with-modelsbuilder#comment-302289 where these already exists in core.

I guess you could end up with a bunch of int and bool property value converters then.

bjarnef:
Hi @nul800sebastiaan

Any update on the outcome of the internal discussion? 😁
Maybe @Shazwazza @zpqrtbnk or @clausjensen have some ideas on this?

nul800sebastiaan:
Sorry for the delay, this sounds good! Not sure how this can be implemented yet but maybe PVCEs could look at an editor's valueType (in the package.manifest file)?.

I've marked this as "Up for grabs" so that you or someone else coming along could create a pull request for it.

[...]

PR #8271, opened 2020-06-10
Add generic PropertyValueConverterBase classes

ronaldbarendse:
### Prerequisites

### Description
This is currently WIP and I'm creating a draft PR to discuss whether adding generic versions of `PropertyValueConverterBase` might help with creating custom PVCs and add some more type checking between source, intermediate and object values.

These generic base classes can be used as follows:

```c#
public class YesNoValueConverter : PropertyValueConverterBase<bool>
{
    public override bool IsConverter(IPublishedPropertyType propertyType)
        => propertyType.EditorAlias == Constants.PropertyEditors.Aliases.Boolean;

    public override PropertyCacheLevel GetPropertyCacheLevel(IPublishedPropertyType propertyType)
        => PropertyCacheLevel.Element;

    protected override bool ConvertSourceToIntermediate<T>(IPublishedElement owner, IPublishedPropertyType propertyType, object source, bool preview)
    {
        if (source is string s)
        {
            if (s.Length == 0 || s == "0")
                return false;

            if (s == "1")
                return true;

            return bool.TryParse(s, out var result) && result;
        }

        if (source is int i)
            return i == 1;

        if (source is long l)
            return l == 1;

        if (source is bool b)
            return b;

        return false;
    }

    protected override object ConvertIntermediateToXPath<T>(IPublishedElement owner, IPublishedPropertyType propertyType, PropertyCacheLevel referenceCacheLevel, bool inter, bool preview)
        => inter ? "1" : "0";
```

As you can see, there's no need to specify the value type (as that's already set to the generic type parameter) and the intermediate value doesn't need casting to a boolean.

This is a very simple PVC (and thus a very small code reduction compared to https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Core/PropertyEditors/ValueConverters/YesNoValueConverter.cs), but having the type safety/enforcement on more complex PVCs will ensure the correct values/types are returned!

If the intermediate and object values are different, `PropertyValueConverterBase<TIntermediate, TObject>` can be used as base class and will ensure the correct types are available on the `ConvertIntermediateToObject<T>()` method.

### DIFF
--- a/src/Umbraco.Core/PropertyEditors/PropertyValueConverterBase.cs
+++ b/src/Umbraco.Core/PropertyEditors/PropertyValueConverterBase.cs
@@ -45,4 +45,61 @@ namespace Umbraco.Core.PropertyEditors
         public virtual object ConvertIntermediateToXPath(IPublishedElement owner, IPublishedPropertyType propertyType, PropertyCacheLevel referenceCacheLevel, object inter, bool preview)
             => inter?.ToString() ?? string.Empty;
     }
+
+    public abstract class PropertyValueConverterBase<TIntermediate, TObject> : PropertyValueConverterBase
+    {
+        public sealed override bool? IsValue(object value, PropertyValueLevel level)
+        {
+            switch (level)
+            {
+                case PropertyValueLevel.Source:
+                    return IsSourceValue(value);
+                case PropertyValueLevel.Inter:
+                    return IsIntermediateValue((TIntermediate)value);
+                case PropertyValueLevel.Object:
+                    return IsObjectValue((TObject)value);
+                default:
+                    return base.IsValue(value, level);
+            }
+        }
+
+        protected virtual bool? IsSourceValue(object value)
+            => base.IsValue(value, PropertyValueLevel.Source);
+
+        protected virtual bool? IsIntermediateValue(TIntermediate value)
+            => base.IsValue(value, PropertyValueLevel.Inter);
+
+        protected virtual bool? IsObjectValue(TObject value)
+            => base.IsValue(value, PropertyValueLevel.Object);
+
+        public override Type GetPropertyValueType(IPublishedPropertyType propertyType)
+            => typeof(TObject);
+
+        public sealed override object ConvertSourceToIntermediate(IPublishedElement owner, IPublishedPropertyType propertyType, object source, bool preview)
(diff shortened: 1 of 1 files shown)


ronaldbarendse:
To give an example if a core PVC that returns an invalid object value (`ConvertIntermediateToObject` should return either `IPublishedContent` or `IEnumerable<IPublishedContent>`):

https://github.com/umbraco/Umbraco-CMS/blob/df8f28114160271ed4ab5db4519b4d13c3acccfc/src/Umbraco.Web/PropertyEditors/ValueConverters/MultiNodeTreePickerValueConverter.cs#L65-L127

This returns the first `Udi` from the intermediate value for the excluded property aliasses (`umbracoInternalRedirectId` and `umbracoRedirect`) or the original `Udi[]` when no UmbracoContext is available.

This makes the PVC return unexpected values and will be hard to debug e.g. why is the `IPublishedContent` property generated by ModelsBuilder always returning `null` for a property with `umbracoInternalRedirectId`?

Using these generic `PropertyValueConverterBase` classes will mean you'd need 3 PVCs though (similair to the Nested Content PVCs):
- `MultiNodeTreePickerSingleValueConverter : PropertyValueConverterBase<Udi, IPublishedContent>` - for single item MNTP (maxItems 1), excluding the `umbracoInternalRedirectId` and `umbracoRedirect` property aliasses
- `MultiNodeTreePickerManyValueConverter : PropertyValueConverterBase<Udi[], IEnumerable<IPublishedContent>>` - for multi item MNTP (maxItems != 1), excluding the `umbracoInternalRedirectId` and `umbracoRedirect` property aliasses
- `MultiNodeTreePickerExcludedValueConverter : PropertyValueConverterBase<Udi>` - only for the `umbracoInternalRedirectId` and `umbracoRedirect` property aliasses and always returning a single `Udi`

abjerner:
@ronaldbarendse this sounds a lot like what @bjarnef suggested here in https://github.com/umbraco/Umbraco-CMS/issues/4419, so you may be able find some useful hints in the discussion 😉

abjerner:
@ronaldbarendse if you look at Bjarne's suggestions, his idea is to look at the `valueType` property instead of the alias. This may give the PVC a wider use, as you can then also use it for your own property editors.

PR #8272, opened 2020-06-11
8245 allow string templates in overlays

### LINKED ISSUE #8245: Allow template string in overlayService
To avoid adding files with only a few lines of HTML when creating simple overlays, it would be awesome to be able to render the overlay (not an infinite editor) by providing the view HTML rather than the view path.

```
this.overlayService.open({
    view: '<div>My simple overlay has HTML instead of a path</div>`,
    submit: () => {
        this.doSomething();
        this.overlayService.close();
    },
    close: () => this.overlayService.close()
});
```
This would be a similar behaviour to how AngularJs manages directives/components, where the object has either the `template` or `templateUrl` property. I haven't looked at how the current overlays work, so this could be a trivial task or a not-so-trivial task.

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #8245 

### Description

Changes allow passing an HTML string as the overlay view as an alternate to a view path - means that for simple overlays there's no need to add an additional HTML file to hold that content. 

Testing requires a couple of changes to existing content, given this is a new thing. Doesn't need to be these files, I just grabbed something existing to save spinning up a new dashboard/view.

- edit Umbraco.Web.UI.Client/src/views/dashboard/settings/examinemanagement.html, adding the below (anywhere, it's just a button to trigger our overlay): 

`<button ng-click="vm.overlay()" type="button">Overlay</button>`

- edit Umbraco.Web.UI.Client/src/views/dashboard/settings/examinemanagement.controller.js, adding the following (also doesn't matter where, just smoosh it into the controller):

```javascript
    vm.overlay = overlay;
    function overlay() { 
        overlayService.open({
            view: '<div>value can be localized <localize key="general_delete"></localize> or include other <b>html</b></div>',
            submitButtonLabelKey: 'general_delete',
            submitButtonStyle: 'danger',
            submit: () => overlayService.close(),
            close: () => overlayService.close()
        });
    }
```

To note here is the string being passed as the view - it contains HTML and an AngularJs directive, both of which will be compiled and rendered into our overlay. Save both files, navigate to the Examine management dashboard in settings, click your button and whammo, overlay from a string.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/overlays/umboverlay.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/overlays/umboverlay.directive.js
@@ -189,7 +189,7 @@ Opens an overlay to show a custom YSOD. </br>
 </table>
 
 @param {object} model Overlay options.
-@param {string} view Path to view or one of the default view names.
+@param {string} view Path to view, one of the default view names or an HTML string.
 @param {string} position The overlay position ("left", "right", "center": "target").
 **/
 
@@ -262,9 +262,8 @@ Opens an overlay to show a custom YSOD. </br>
             function setView() {
                 if (scope.view) {
 
-                    // if the view value has a space, treat it as an inline value
-                    // since setting the view by alias will not have a space
-                    if (scope.view.indexOf(' ') > -1) {
+                    // an inline template must be HTML, so must start and end with < and > respectively
+                    if (scope.view.startsWith('<') && scope.view.endsWith('>')) {
                         // reuse the existing scoped-view element, and compile our view into it
                         const element = el.find('.scoped-view');
                         element.html(scope.view);
@@ -274,7 +273,7 @@ Opens an overlay to show a custom YSOD. </br>
                         // ensure the ng-include view isn't rendered 
                         scope.viewIsLiteral = true;
                     }
-                    
+                    // when passed a view alias eg confirm, ysod
                     else if (scope.view.indexOf(".html") === -1) {
                         var viewAlias = scope.view.toLowerCase();
                         scope.view = "views/common/overlays/" + viewAlias + "/" + viewAlias + ".html";
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/overlays/umboverlay.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/overlays/umboverlay.directive.js
@@ -260,10 +260,22 @@ Opens an overlay to show a custom YSOD. </br>
             }
 
             function setView() {
-
                 if (scope.view) {
 
-                    if (scope.view.indexOf(".html") === -1) {
+                    // if the view value has a space, treat it as an inline value
+                    // since setting the view by alias will not have a space
+                    if (scope.view.indexOf(' ') > -1) {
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web.UI.Client/src/common/directives/components/overlays/umboverlay.directive.js
```diff
@@ -260,10 +260,22 @@ Opens an overlay to show a custom YSOD. </br>
             }
 
             function setView() {
-
                 if (scope.view) {
 
-                    if (scope.view.indexOf(".html") === -1) {
+                    // if the view value has a space, treat it as an inline value
+                    // since setting the view by alias will not have a space
+                    if (scope.view.indexOf(' ') > -1) {
```
bjarnef, line 267: @nathanwoulfe couldn't this potential break stuff if developers have specified a view like `~/App_Plugins/My Plugin/view.html`?

Maybe it is fine to assume at least any html it added, e.g. it contains at least `<` and `>`?
Does it still work using the shorthand view?

nathanwoulfe: @bjarnef good point - I could check for spaces and the absence of `.html`, although that would fail if core added/changed existing overlay naming to have spaces. Maybe require the literal view be wrapped in an HTML element, and can then checks `startsWith('<')` and `endsWith('>')`. It probably should require HTML rather than a bare string, in the interest of semantic HTML and to be consistent with Angular directive templates, which require a single root element.

I don't want to introduce a new property for the literal view, really want to manage it in code - cleaner and friendly for developers...

bjarnef:
I think it is also possible to specify a shorthand for the view like `boolean` for property editors or `confirm` in overlayService as I remember.

nathanwoulfe:
Yup, that will work fine, even better if I require html as the literal template string

nul800sebastiaan:
Might it be better to add a property `viewLiteral` instead? Or maybe prefix a literal with `@` or something like that? I know both are not super transparent but if documented shouldn't that be fine? Not sure about the implications of adding a new property though?

nathanwoulfe:
@nul800sebastiaan for simplicity's sake for the developer, I wanted to keep the single property for view - I considered `viewLiteral` but felt that then introduced complexity where a dev could set both properties (one would need to take priority of the other), or could mix up the values and send a view path as the literal value, or html as the view value. Also means some additional ng-if checks in the overlay directive view.

I'd maybe go with prefixing the literal view - that works nicely in that it expresses intent. Or could always ensure the documentation is robust and explicitly requires the literal be a single HTML element (ie must start and end with an HTML tag). As it stands now, passing an HTML string will render that string, passing a view path will render the view and passing a known view alias will render the aliased view. The only failure point would be someone passing a `string that meets none of those`, which would fail in the current implementation anyway.

Whatever change is made would need to be documented, so maybe we're chasing our tails 😃. The more that can be made to Just Work for devs, the better...

nathanwoulfe:
Hey @nul800sebastiaan @bjarnef, I've run into this issue again on some package work, where I have a single-line overlay template which would be much better off inlined as a literal value. If I update this PR to prefix the view value for literals, would there be any objections to merging? Will update docs too.

PR #8277, opened 2020-06-11
Migrate unit tests to new test project

electricsheep:
### Prerequisites

This fixes part of #7898

### Description

Moved a bunch more test classes over to the new project - some more simple ones that don't require much refactoring.

- `TestClone` - pulled out into separate class in the `Umbraco.Test.Common` project since it is used in multiple test classes
- Removed a test from `ReflectionTests` which seems be testing `Type.GetInterfaces`. Doesn't seem like that is useful? But can restore if so.


bergmania:
@electricsheep, Looks good.. Thank you very much 💪

Issue #8280, opened 2020-06-12
Log configuration with AppDomain.CurrentDomain.BaseDirectory does not support Virtual Directories.


enkelmedia:
We're using Octopus Deploy for our release management, each time new release is pushed a new directory is created with the version number of that release, and the IIS website is pointed to the new folder.

/wwwroot/website1/releases/1.0.0
/wwwroot/website1/releases/1.0.1
/wwwroot/website1/releases/1.0.2

We want files like processed images, cache and log files to "survive" these releases which we do by creating virtual directories for folders that we need to keep after each release. 

So ie App_Data\Temp points to:

/wwwroot/website1/resources/App_Data/Temp

Currently we've got everything except the log files working. The problem seems to be the way that the path for the log directory is created. The paths are created using a string contact with  AppDomain.CurrentDomain.BaseDirectory + "\App_Data\Logs\...".

In a scenario when ie /App_Data/Logs/ is a virtual directory this will not work as the directory will be created in side the "release"-folder, in our case

/1.0.0/App_Data/Logs

What we would like is for this to respect the Virtual Directory and put the logs in 

/wwwroot/website1/resources/App_Data/Logs

It looks like AppDomain.CurrentDomain.BaseDirectory is used in some places in the code base but mostly around the logging:

https://github.com/umbraco/Umbraco-CMS/blob/62dc68e7b2c75a5205bb486674b994587f555dd1/src/Umbraco.Core/Logging/Serilog/LoggerConfigExtensions.cs#L29

https://github.com/umbraco/Umbraco-CMS/blob/62dc68e7b2c75a5205bb486674b994587f555dd1/src/Umbraco.Core/Logging/Viewer/JsonLogViewer.cs#L19

I'm proposing that we should use something like HostingEnvironment.MapPath("~/App_Data/Logs") to get the logs folder, this way the virtual directory would be respected.

It's possible for me to work around this at the moment by:
* Adding a EnvironmentVariable called "APPDATA" with this value: HostingEnvironment.MapPath("~/App_Data/Logs") and use that in serilog.config %APPDATA%.
* Implementing a custom "ILogViewer" that used the same method to create the paths.

It really feels bad to have to create a bunch of custom stuff just to get log files where we need them to. I also know that I can configure a absolut path to the log file in serilog.config but that would not be recpected by the ILogViewer used in the backoffice so I think that the easiest solution is to use HostingEnvironment.MapPath.

[...]

PR #8281, opened 2020-06-12
Using IOHelper for Log files

enkelmedia:
### Prerequisites

#8280

### Description
Adjusted code around logging to use IOHelper over hardcoded paths with AppDomain.CurrentDomain.BaseDirectory ie. to support virtual directories for log files.

I've adjusted the environment variable called "BASEDIR" to also use the IOHelper, I had to trim the string in the end since the IOHelper return a trailing slash but the configuration in serilog.config does not expect this. So the Trim is to keep backwards compatibility for sites that is upgrading.

Also added a property for the LogFiles in the SystemDirectories-class to avoid having it hardcoded in lots of places. 

Since the JsonLogViewer reads this setting we still have an "issue" if someone wants to store log files in a custom folder (ie out of the root), we could address this by making SystemDirectories.LogFile simular to the SystemDirectories.Media where one could also configure this in web.config app settings.

Let me know if that is something that we want to do?

### DIFF
--- a/src/Umbraco.Core/IO/SystemDirectories.cs
+++ b/src/Umbraco.Core/IO/SystemDirectories.cs
@@ -11,6 +11,8 @@ namespace Umbraco.Core.IO
 
         public static string Data => "~/App_Data";
 
+        public static string LogFiles => Data + "/Logs";
+
         public static string TempData => Data + "/TEMP";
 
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Looking good, thanks @enkelmedia! 👍

Issue #8264, opened 2020-06-09
NetCore: Use TypeFilterAttribute with nested filter for attributes


AndyButland:
There are a few examples in the code base of .NET Core MVC attributes implemented as filters to allow the use of constructor based dependency injection and applied to controllers using `[TypeFilter]` attribute.

We'd like to migrate all of these instances to a new pattern, where the filter is nested within an attribute inheriting from `TypeFilterAttribute` (as explained in [this blog post](https://www.andybutland.dev/2020/06/dependency-injection-in-aspnet-core-attributes.html)).

A solution-wide search for "[TypeFilter" will reveal all the instances.

An example of one of these is `AngularJsonOnlyConfigurationAttribute`, which was converted in [this commit](https://github.com/umbraco/Umbraco-CMS/pull/8242/commits/af382e5e71c5fa672bc29066958c8530b03e501d).

This issue can likely best be handled via multiple PRs.

umbrabot:
Hi @AndyButland,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

ronaldbarendse:
Just to add some additional information, this blog post explains why DI can't be used with attributes: https://blogs.cuttingedge.it/steven/posts/2014/dependency-injection-in-attributes-dont-do-it/ (TL;DR: the attributes are not constructed/resolved by the DI container). The proposed design separates the data/behavior a little bit better then the `TypeFilterAttribute`, but in essence it's the same: let the attribute resolve the type from the DI container when it's needed by the ASP.NET MVC framework.

AndyButland:
If anyone is planning to work on this as part of the upcoming [Umbraco Hackathon](https://candidcontributions.com/umbrackathon/), there's a Google Doc [here](https://docs.google.com/spreadsheets/d/1udNKPQ7Iu5sYQf5q98S4sjo2SAT39Z4n64XxVYUhUx8/edit?usp=sharing) to track what you and others are working on.  The idea being that people don't accidentally end up overlapping and working on the same classes.

Matthew-Wise:
[...]

PR #8284, opened 2020-06-12
Have the HttpResponseException filter use the new pattern

louisjacksonrees:
### Prerequisites

https://github.com/umbraco/Umbraco-CMS/issues/8264

### Description
Ensuring that the `HttpResponseExceptionFilter` meets the new pattern as described in #8264 

To test ensure that is there is a HttpResponseException, the correct status code is applied,


AndyButland:
Thanks for submitting this @louisjacksonrees  (first PR to Umbraco I think? Welcome if so!).

I think this just needs a couple of minor amends.

The nested filter class within the attribute should be marked as `private` - just so it's clear it's only intended to be used from within the attribute.

And I think - checking with @bergmania  to confirm - we probably want the `Order` property available when using the attribute (and in fact we have it, because it's defined in `TypeFilterAttribute`.  Which means we can remove the `IOrderedFilter` interface, and set the default order in the constructor, i.e.:

```
    public HttpResponseExceptionAttribute() : base(typeof(HttpResponseExceptionFilter))
    {
        Order = int.MaxValue - 10;
    }
```

PR #8290, opened 2020-06-15
Importing a doctype leaves behind temp files

nul800sebastiaan:
Importing a doctype would never assign a value to UploadedFiles, turns out we don't need UploadedFiles on the model

Additionally added a check to see if temp files were left behind and delete them to prevent errors when the same file has just been uploaded


nul800sebastiaan:
Note the `FileCleanupFilter` was removed because we want to clean up the files later in the process, this was being done in the `Import` method already.

bergmania:
Also tested.. Work as expected :)

PR #8302, opened 2020-06-17
Netcore: Content/Media controller + ContentType+MediaType controller

bergmania:
Notes:
- Fixed issue with `ConnectionStrings.ParseProvider` so the connection string from the integration tests are passed as `SqlServer`
- Added more default values to the `ImagingAutoFillUploadField`, to avoid null reference issues.
- Had to make a couple of things public, to access them from the web projects
- Fixed issue with the filesystem, to ensure the root is wwwroot
- Extended the `ContentBuilder` +related and created a `ContentItemSaveBuilder` + related to make test data.
- Introduced out variable for `connectionstring` in `ApplicationBuilderExtensions.UseTestLocalDb`
- Added a super simple test using the testserver `BackOfficeAssetsControllerTests`
- Migrated ContentControllerTests (Testserver), but added explicit for now. We need to figure out why these tests are not stable in a later PR
- Had to improve the `DetermineAmbiguousActionByPassingParameters` to handle arrays of types also
- Had to uncomment `InvalidModelStateFilterConvention` as we handle bad requests in a different way our selves in the content/media/member controllers
- Pulled `SetReasonPhrase` into an extension method

Migrated the following filters
- `PrefixlessBodyModelValidator` 
     - As far as I can see there is not good why to do this anymore, unless you change it for all controllers. So this implementation is more like a hack.

Migrated the following Controllers:
- `ContentControllerBase`
   - Including validators and model binders
- `ContentController`
- `MediaController`
- `MemberController`
- `ContentTypeControllerBase`
- `ContentTypeController`
- `MediaTypeController`
- `MemberTypeController`
- `EntityController`

# Tests
- Should be handled as one combined tests after [next PR](https://github.com/umbraco/Umbraco-CMS/pull/8323), where backoffice should be usable.


src/Umbraco.Web.BackOffice/ModelBinders/FromJsonPathAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// </summary>
+    /// <remarks>
+    ///     An example would be if you had json like:
+    ///     { ids: [1,2,3,4] }
+    ///     And you had an action like: GetByIds(int[] ids, UmbracoEntityTypes type)
+    ///     The ids array will not bind because the object being sent up is an object and not an array so the
+    ///     normal json formatter will not figure this out.
+    ///     This would also let you bind sub levels of the JSON being sent up too if you wanted with any jsonpath
+    /// </remarks>
+    internal class FromJsonPathAttribute : ModelBinderAttribute
```
benjaminc, line 26: Is there a reason to make this internal?  This one seems like a nice, general-purpose attribute that could be reused by other packages.

src/Umbraco.Web.Common/Extensions/HttpRequestExtensions.cs
```diff
@@ -52,6 +55,21 @@ private static bool IsSet(this IPAddress address)
         {
             const string NullIpAddress = "::1";
             return address != null && address.ToString() != NullIpAddress;
+        }
+
+        public static string GetRawBodyString(this HttpRequest request, Encoding encoding = null)
```
benjaminc, line 60: Should there be an async version of this?  Then the JsonPathBinder could use the async version in its BindModelAsync method.

src/Umbraco.Web.BackOffice/ActionResults/UmbracoNotificationSuccessResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class UmbracoNotificationSuccessResponse : OkObjectResult
+    {
+        public UmbracoNotificationSuccessResponse(string successMessage) : base(null)
+        {
+            var notificationModel = new SimpleNotificationModel
+            {
+                Message = successMessage
+            };
+            notificationModel.AddSuccessNotification(successMessage, string.Empty);
```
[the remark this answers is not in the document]
bergmania: Logically I agree, but this is the replacement of `HttpRequestMessageExtensions.CreateNotificationSuccessResponse` which does the exact same

src/Umbraco.Web.BackOffice/Filters/FileUploadCleanupFilterAttribute.cs
```diff
@@ -0,0 +1,163 @@
...
+                    {
+
+                        if (context.ActionArguments.First().Value is IHaveUploadedFiles contentItem)
+                        {
+                            //cleanup any files associated
+                            foreach (var f in contentItem.UploadedFiles)
+                            {
+                                //track all temp folders so we can remove old files afterwards
+                                var dir = Path.GetDirectoryName(f.TempFilePath);
+                                if (tempFolders.Contains(dir) == false)
```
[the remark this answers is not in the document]
bergmania: I just copied the file from the old Umbraco.Web project. I'm pretty sure it is just coding preferences.

The only thing that can be performance optimised is null checks. `x is null` vs `x == null`.  The `is` operator is way faster because it does not check operator overload - But this also makes them a bit different.

src/Umbraco.Web.BackOffice/Filters/FileUploadCleanupFilterAttribute.cs
```diff
@@ -0,0 +1,163 @@
...
+                                        File.Delete(f.TempFilePath);
+                                    }
+                                    catch (Exception ex)
+                                    {
+                                        _logger.Error<FileUploadCleanupFilterAttribute>(ex,
+                                            "Could not delete temp file {FileName}", f.TempFilePath);
+                                    }
+
+                                    //clear out the temp path so it's not returned in the response
+                                    f.TempFilePath = "";
```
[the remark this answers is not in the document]
bergmania: Again, I haven't changed anything that was not required from the original file `src/Umbraco.Web/WebApi/Filters/FileUploadCleanupFilterAttribute.cs`

src/Umbraco.Web.BackOffice/ModelBinders/FromJsonPathAttribute.cs
bergmania: Not really, other then we don't have to maintain that functionality public. I change to public for now, but the reason it was internal, was because the original file was internal `src/Umbraco.Web/Editors/FromJsonPathAttribute.cs`

src/Umbraco.Web.Common/Extensions/HttpRequestExtensions.cs
bergmania: Added one

src/Umbraco.Web.BackOffice/Filters/EnsureUserPermissionForMediaAttribute.cs
```diff
@@ -0,0 +1,181 @@
...
+{
+    /// <summary>
+    /// Auth filter to check if the current user has access to the content item
+    /// </summary>
+    /// <remarks>
+    /// Since media doesn't have permissions, this simply checks start node access
+    /// </remarks>
+    internal sealed class EnsureUserPermissionForMediaAttribute : TypeFilterAttribute
+    {
+        public EnsureUserPermissionForMediaAttribute(int nodeId)
```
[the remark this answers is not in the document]
bergmania: I don't think it is required. If the input is a parameter, we use the the overload taking a parameter, and that handle UDIs. 
The overload using int is only for known nodes, like the RecycleBin. In that case I think int is okay.

No matter what, It was not part of the attributes I migrated, so if we need it, it could be in its own PR. UDIs are not used internally.

src/Umbraco.Web.BackOffice/Filters/EnsureUserPermissionForContentAttribute.cs
```diff
@@ -0,0 +1,252 @@
...
+    ///     By default the permission that is checked is browse but this can be specified in the ctor.
+    ///     NOTE: This cannot be an auth filter because that happens too soon and we don't have access to the action params.
+    /// </remarks>
+    public sealed class EnsureUserPermissionForContentAttribute : TypeFilterAttribute
+    {
+
+        /// <summary>
+        ///     This constructor will only be able to test the start node access
+        /// </summary>
+        public EnsureUserPermissionForContentAttribute(int nodeId)
```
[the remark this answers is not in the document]
bergmania: I don't think it is required. If the input is a parameter, we use the the overload taking a parameter, and that handle UDIs. 
The overload using int is only for known nodes, like the RecycleBin. In that case I think int is okay.

No matter what, It was not part of the attributes I migrated, so if we need it, it could be in its own PR. UDIs are not used internally.

bergmania:
Thanks for review/feedback @benjaminc .

Please read my comments and see if you agree..

benjaminc:
Looks great!

src/Umbraco.Web.BackOffice/ModelBinders/BlueprintItemBinder.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Umbraco.Web.BackOffice.ModelBinders
+{
+    internal class BlueprintItemBinder : ContentItemBinder
+    {
+        private readonly ContentService _contentService;
+
+        public BlueprintItemBinder(IJsonSerializer jsonSerializer, UmbracoMapper umbracoMapper, IContentService contentService, IContentTypeService contentTypeService, IHostingEnvironment hostingEnvironment, ContentService contentService2) : base(jsonSerializer, umbracoMapper, contentService, contentTypeService, hostingEnvironment)
+        {
+            _contentService = contentService2;
```
emmagarland, line 17: This maybe a historical thing but why do we need 2 content services?

src/Umbraco.Web.BackOffice/Controllers/MediaTypeController.cs
```diff
@@ -0,0 +1,351 @@
...
+using Umbraco.Web.Common.Exceptions;
+using Umbraco.Web.Editors;
+using Umbraco.Web.Models.ContentEditing;
+using Umbraco.Web.Security;
+
+namespace Umbraco.Web.BackOffice.Controllers
+{
+    // TODO:  We'll need to be careful about the security on this controller, when we start implementing
+    // methods to modify content types we'll need to enforce security on the individual methods, we
+    // cannot put security on the whole controller because things like GetAllowedChildren are required for content editing.
```
emmagarland, line 23: Definitely one to zoom in on

src/Umbraco.Web.BackOffice/Filters/ContentModelValidator.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// <param name="actionContext"></param>
+        /// <returns></returns>
+        protected bool ValidateProperties(List<ContentPropertyBasic> postedProperties, List<IProperty> persistedProperties, ActionExecutingContext actionContext)
+        {
+            foreach (var p in postedProperties)
+            {
+                if (persistedProperties.Any(property => property.Alias == p.Alias) == false)
+                {
+                    // TODO: Do we return errors here ? If someone deletes a property whilst their editing then should we just
+                    //save the property data that remains? Or inform them they need to reload... not sure. This problem exists currently too i think.
```
emmagarland, line 99: Interesting, personally I'd like to be notified that it has happened, but not lose data...

src/Umbraco.Web.BackOffice/Filters/ContentSaveValidationAttribute.cs
```diff
@@ -0,0 +1,244 @@
...
+                int contentIdToCheck;
+                switch (contentItem.Action)
+                {
+                    case ContentSaveAction.Save:
+                        permissionToCheck.Add(ActionUpdate.ActionLetter);
+                        contentToCheck = contentItem.PersistedContent;
+                        contentIdToCheck = contentToCheck.Id;
+                        break;
+                    case ContentSaveAction.Publish:
+                    case ContentSaveAction.PublishWithDescendants:
```
emmagarland, line 116: I know this was moved but seems like a future candidate for a refactor

src/Umbraco.Web.BackOffice/Filters/EnsureUserPermissionForMediaAttribute.cs
```diff
@@ -0,0 +1,181 @@
...
+                }
+                else
+                {
+                    throw new HttpResponseException(System.Net.HttpStatusCode.Unauthorized);
+                }
+            }
+
+            public void OnActionExecuted(ActionExecutedContext context)
+            {
+
```
emmagarland, line 176: I can see this has shuffled around a little bit, so there are now some empty methods and we aren't needing to call the base method(s) here since we derive from ActionFilterAttribute. Just checking this is a conscious decision not to call it

src/Umbraco.Web.BackOffice/Filters/FileUploadCleanupFilterAttribute.cs
```diff
@@ -0,0 +1,163 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+                else
+                {
+                    if (context == null)
+                    {
+                        _logger.Warn<FileUploadCleanupFilterAttribute>("The context is null!!??");
```
emmagarland, line 89: Very expressive warning message!

src/Umbraco.Web.BackOffice/Filters/FileUploadCleanupFilterAttribute.cs
```diff
@@ -0,0 +1,163 @@
...
+                                    {
+                                        tempFolders.Add(dir);
+                                    }
+
+                                    _logger.Debug<FileUploadCleanupFilterAttribute>(
+                                        "Removing temp file {FileName}", f.TempFilePath);
+
+                                    try
+                                    {
+                                        File.Delete(f.TempFilePath);
```
emmagarland, line 128: Ends up quite nested but I think sometimes that is clearer than returning after the opposite case

src/Umbraco.Web.BackOffice/Filters/FilterAllowedOutgoingMediaAttribute.cs
```diff
@@ -135,5 +141,11 @@ internal dynamic GetValueFromResponse(ObjectContent objectContent)
 
             return null;
         }
+
+
+        public void OnActionExecuting(ActionExecutingContext context)
+        {
```
emmagarland, line 147: (same comment from me as before re base methods)

src/Umbraco.Web.BackOffice/Filters/PrefixlessBodyModelValidatorAttribute.cs
```diff
@@ -0,0 +1,52 @@
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+
+namespace Umbraco.Web.BackOffice.Filters
+{
+    public class PrefixlessBodyModelValidatorAttribute : TypeFilterAttribute
```
emmagarland, line 9: Can this summary still have comments in it describing the purpose?

src/Umbraco.Web.BackOffice/Filters/FileUploadCleanupFilterAttribute.cs
```diff
@@ -0,0 +1,163 @@
...
+using Umbraco.Core.Logging;
+using Umbraco.Web.Models.ContentEditing;
+
+namespace Umbraco.Web.WebApi.Filters
+{
+    /// <summary>
+    ///     Checks if the parameter is IHaveUploadedFiles and then deletes any temporary saved files from file uploads
+    ///     associated with the request
+    /// </summary>
+    internal sealed class FileUploadCleanupFilterAttribute : TypeFilterAttribute
```
Shazwazza, line 18: This attribute was removed from some actions in this PR but this was also moved? How come? I thought we didn't need this filter anymore?

src/Umbraco.Web.BackOffice/Filters/FileUploadCleanupFilterAttribute.cs
Shazwazza: Yeah, due to bugs in old mvc this used to throw because of bugs in async support in old net framework versions, was hard to track the bug down, these should never be thrown. Pretty sure thsi whole filter can be removed.

src/Umbraco.Web.BackOffice/Filters/FilterAllowedOutgoingMediaAttribute.cs
```diff
@@ -2,57 +2,65 @@
...
+
 
 namespace Umbraco.Web.WebApi.Filters
 {
     /// <summary>
     /// This inspects the result of the action that returns a collection of content and removes
     /// any item that the current user doesn't have access to
     /// </summary>
-    internal class FilterAllowedOutgoingMediaAttribute : ActionFilterAttribute
+    internal class FilterAllowedOutgoingMediaAttribute : TypeFilterAttribute
```
Shazwazza, line 22: I thought I saw in the media controller that these attributes were commented out with a note saying to re-enable them when this is transfered to netcore?

src/Umbraco.Web.BackOffice/Filters/FilterAllowedOutgoingMediaAttribute.cs
Shazwazza: @bergmania ?

src/Umbraco.Web.BackOffice/ModelBinders/BlueprintItemBinder.cs
Shazwazza: @bergmania yes good point, i assume it was due to internals but these should be fixed in netcore?

src/Umbraco.Web.BackOffice/Umbraco.Web.BackOffice.csproj
```diff
@@ -12,7 +12,9 @@
 
     <ItemGroup>
       <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="3.1.4" />
+      <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
       <PackageReference Include="Serilog.AspNetCore" Version="3.2.0" />
+      <PackageReference Include="System.IO.Pipelines" Version="4.7.1" />
```
Shazwazza, line 17: what do we do with this one?

src/Umbraco.Web.Common/ApplicationModels/UmbracoApiBehaviorApplicationModelProvider.cs
```diff
@@ -36,7 +36,7 @@ public UmbracoApiBehaviorApplicationModelProvider(IModelMetadataProvider modelMe
             ActionModelConventions = new List<IActionModelConvention>()
             {
                 new ClientErrorResultFilterConvention(), // Ensures the responses without any body is converted into a simple json object with info instead of a string like "Status Code: 404; Not Found"
-                new InvalidModelStateFilterConvention(), // automatically 400 responses if ModelState is invalid before hitting the controller
+                //new InvalidModelStateFilterConvention(), // automatically 400 responses if ModelState is invalid before hitting the controller
```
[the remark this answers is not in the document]
Shazwazza: I think we should just remove it since we won't be able to support this globally in our controllers.

src/Umbraco.Web.Common/Extensions/HttpRequestExtensions.cs
```diff
@@ -53,5 +56,27 @@ private static bool IsSet(this IPAddress address)
             const string NullIpAddress = "::1";
             return address != null && address.ToString() != NullIpAddress;
         }
+
+        public static string GetRawBodyString(this HttpRequest request, Encoding encoding = null)
+        {
+            request.Body.Seek(0, SeekOrigin.Begin);
+
+            var reader = new StreamReader(request.Body, encoding ?? Encoding.UTF8);
```
Shazwazza, line 64: Do these readers need to be disposed? I assume we don't want to dispose the stream?

src/Umbraco.Web.Common/Filters/AngularJsonOnlyConfigurationAttribute.cs
```diff
@@ -13,6 +13,7 @@ public class AngularJsonOnlyConfigurationAttribute : TypeFilterAttribute
     {
         public AngularJsonOnlyConfigurationAttribute() : base(typeof(AngularJsonOnlyConfigurationFilter))
         {
+            Order = -2400;
```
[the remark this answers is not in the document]
Shazwazza: @bergmania ?

src/Umbraco.Web.BackOffice/Filters/PrefixlessBodyModelValidatorAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        public PrefixlessBodyModelValidatorAttribute() : base(typeof(PrefixlessBodyModelValidatorFilter))
+        {
+        }
+
+        private class PrefixlessBodyModelValidatorFilter : IActionFilter
```
Shazwazza, line 15: Would like to understand more about your comment: "As far as I can see there is not good why to do this anymore, unless you change it for all controllers. So this implementation is more like a hack." Maybe we need some TODO notes in here for what you tried, what might not work now, etc... and maybe a follow up task?

src/Umbraco.Web.BackOffice/Controllers/BackOfficeNotificationsController.cs
```diff
@@ -7,6 +8,7 @@ namespace Umbraco.Web.BackOffice.Controllers
     /// resulting message is INotificationModel in which case it will append any Event Messages
     /// currently in the request.
     /// </summary>
+    [PrefixlessBodyModelValidator]
```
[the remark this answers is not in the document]
bergmania: It was removed earlier. In v8 its also on this class

src/Umbraco.Web.Common/Filters/AngularJsonOnlyConfigurationAttribute.cs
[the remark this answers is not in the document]
bergmania: Seems not to be needed. I remove it. Was part of my tests of getting stuff working

src/Umbraco.Web.BackOffice/ModelBinders/BlueprintItemBinder.cs
bergmania: Good point.. I have removed the concrete type

src/Umbraco.Web.BackOffice/Filters/EnsureUserPermissionForMediaAttribute.cs
bergmania: The reason is that we now implement `IActionFilter` which requires us to have both methods, even that we only use one.

src/Umbraco.Web.BackOffice/Filters/FileUploadCleanupFilterAttribute.cs
bergmania: The filter cannot be removed.. We still use it when we upload a file in one action and handle it in another action. Like import of doc types (next PR)

src/Umbraco.Web.BackOffice/Filters/FileUploadCleanupFilterAttribute.cs
bergmania: Some upload methods handle the stream in the action, in these cases we don't need this attribute anymore. As the stream is not saved to a temp folder that needs to be cleaned up. But we still have some examples are we use the temp folder.

src/Umbraco.Web.BackOffice/Umbraco.Web.BackOffice.csproj
bergmania: Removed it again.. Thanks

src/Umbraco.Web.Common/Extensions/HttpRequestExtensions.cs
bergmania: If we dispose the StreamReader it disposed the stream.. And no, we don't want the stream to be disposed

src/Umbraco.Web.BackOffice/Controllers/ContentController.cs
```diff
@@ -1710,45 +1727,40 @@ public ContentDomainsAndCulture GetCultureAndDomains(int id)
...
         [HttpPost]
-        public DomainSave PostSaveLanguageAndDomains(DomainSave model)
+        public ActionResult<DomainSave> PostSaveLanguageAndDomains(DomainSave model)
         {
             foreach (var domain in model.Domains)
             {
                 try
                 {
-                    var uri = DomainUtilities.ParseUriFromDomainName(domain.Name, Request.RequestUri);
+                    var uri = DomainUtilities.ParseUriFromDomainName(domain.Name, new Uri(Request.GetEncodedUrl(), UriKind.RelativeOrAbsolute));
```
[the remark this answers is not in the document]
bergmania: You are right..

src/Umbraco.Tests.Integration/TestServerTest/UmbracoTestServerTestBase.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        [TearDown]
+        public void TearDown()
+        {
+
+            Factory.Dispose();
+
+            if (Current.IsInitialized)
+            {
+                typeof(Current).GetProperty(nameof(Current.IsInitialized), BindingFlags.Public | BindingFlags.Static).SetValue(null, false);
```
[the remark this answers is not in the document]
bergmania: That's like choosing between the plague and cholera.. But have updated..

src/Umbraco.Infrastructure/Composing/CompositionExtensions/FileSystems.cs
```diff
@@ -92,7 +92,17 @@ public static Composition ComposeFileSystems(this Composition composition)
...
+            composition.SetMediaFileSystem(factory =>
+            {
+                var ioHelper = factory.GetInstance<IIOHelper>();
+                var hostingEnvironment = factory.GetInstance<IHostingEnvironment>();
+                var logger = factory.GetInstance<ILogger>();
+                var globalSettings = factory.GetInstance<IGlobalSettings>();
+
+                var rootPath = hostingEnvironment.MapPathWebRoot(globalSettings.UmbracoMediaPath);
+                var rootUrl = hostingEnvironment.ToAbsolute(globalSettings.UmbracoMediaPath);
+                return new PhysicalFileSystem(ioHelper, hostingEnvironment, logger, rootPath, rootUrl);
```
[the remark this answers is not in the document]
bergmania: This should be equivalent to the functionality before.

src/Umbraco.Web.Common/Extensions/HttpRequestExtensions.cs
benjaminc: Could we not use the overload below to dispose of the StreamReader while leaving the stream open?

https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=netcore-3.1#System_IO_StreamReader__ctor_System_IO_Stream_System_Text_Encoding_System_Boolean_System_Int32_System_Boolean_

bergmania: Nice.. Guess that will work :)

src/Umbraco.Web.BackOffice/Filters/PrefixlessBodyModelValidatorAttribute.cs
bergmania: Added todo

src/Umbraco.Tests.Integration/TestServerTest/UmbracoTestServerTestBase.cs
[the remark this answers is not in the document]
Shazwazza: lol, well which disease performs better? 😂

src/Umbraco.Web.Common/Filters/OutgoingNoHyphenGuidFormatAttribute.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            public void OnResultExecuting(ResultExecutingContext context)
+            {
+                if (context.Result is ObjectResult objectResult)
+                {
+                    var serializerSettings = new JsonSerializerSettings();
+                    serializerSettings.Converters.Add(new GuidNoHyphenConverter());
+
+                    objectResult.Formatters.Clear();
+                    objectResult.Formatters.Add(new AngularJsonMediaTypeFormatter(serializerSettings, _arrayPool, _options.Value));
```
Shazwazza, line 45: @bergmania I don't understand why this and the JsonDateTimeFormatAttribute both clear and add the AngularJsonMediaTypeFormatter? That's is the responsibility of AngularJsonOnlyConfigurationAttribute. Is there a reason why this same logic is duplicated to these other 2 attributes?

bergmania: That's true.. The issue is if they don't it will be overridden.  I'm not sure what is the best way to "merge" this info when the formatters are set in different attributes.. Any suggestions? Basically we need to manipulate the JsonSerializerSettings of the existing formatters, if they are of type `NewtonsoftJsonOutputFormatter` 😕

PR #8304, opened 2020-06-17
Improve jsDocs

nielslyngsoe:
I've had trouble running jsDocs and fixed it by adding this pre-install script.

I was as well missing the ability to run a server and watch for file changes, so this goes as well.

:-)


warrenbuckley:
Looks good

PR #8306, opened 2020-06-17
Media picker folder create loader

bjarnef:
### Prerequisites

### Description
When creating a folder from media picker overlay it use the `umb-loader` class. However this is causing horizontal overflow because of the animationand therefore horizontal scrollbar shown/hidden while animating.

That's why I originally added the `umb-loader` directive which wrap this inside another `<div>` element. It also ensure consistency.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/umb-loader.html

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.html
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.html
@@ -20,15 +20,14 @@
                         <div class="umb-mediapicker-upload">
                             <div class="form-search">
                                 <i class="icon-search" aria-hidden="true"></i>
-                                <input class="umb-search-field search-query -full-width-input"
+                                <input type="text"
+                                    class="umb-search-field search-query -full-width-input"
                                     ng-model="vm.searchOptions.filter"
                                     localize="placeholder"
                                     placeholder="@placeholders_search"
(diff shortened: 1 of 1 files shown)


bjarnef:
@nielslyngsoe it would be great to use the `umb-loader` directive here instead 👍

bjarnef:
@abjerner not really any changes except this line:  https://github.com/umbraco/Umbraco-CMS/pull/8306/files#diff-8adc6680306361edd7adcd7a27dd4c7aR84

Besides that I just formatted the document and maybe just me preferring `type` attribute in start of the element :)

abjerner:
@bjarnef yes, it might not be a problem, which is also why I approved your PR 😉

But I can see reasons for considering it either way, and as I'm unsure about the policy for such changes, I think it's best to leave it up to Seb to merge it.

abjerner:
Thanks again @bjarnef 👍 

To explain my reasons for waiting to merge, an argument could be made that those extra changes were not directly related to the PR, and therefore should have their own PR.

But a counter argument could be to follow what I think it called the boy scout rule, meaning you leave the campsite (here the code) in a better condition than which you found it in. Following this rule, those extra changes are very welcome.

I couldn't remember whether this was something we had discussed in the PR team, so I had to check with Seb first. And the conclusion is that it's absolutely okay to do so 😉

Issue #8274, opened 2020-06-11
Make it possible to open user groups in infinite editor


skttl:
When working with users and user groups, it would be nice if you could open up the user group straight from the user editor, by clicking the group name in the selected groups list.

![image](https://user-images.githubusercontent.com/3726467/84357878-9e8aeb00-abc6-11ea-9b39-7f5fc70fbb9d.png)

nul800sebastiaan:
I would love that, it's annoying having to switch over.

One little caveat.. I would also like a big screaming warning (not really.. ;-) ) when you change a user group that those changes will be applied to all users of that group. But I think that's a separate feature request.
It should probably be an extra confirmation saying: if you save these changes they will apply to user x, y and z as well, are you sure?

umbrabot:
Hi @skttl,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8308

nielslyngsoe:
@skttl Great thanks!

PR #8308, opened 2020-06-18
8274: Make it possible to open user groups in infinite editor

patrickdemooij9:
PR for https://github.com/umbraco/Umbraco-CMS/issues/8274

One thing that I noticed in the v8/contrib branch is that adding new groups isn't working properly. Instead of selecting the groups that you already have, it selects nothing. Dunno where this came from, but already had the problem before my PR.

In the user detail, it'll show an edit button for each user group that you have permission to. Permission logic has been taken from the group list. When clicking on the edit, it'll open up the infinite window with the user group edit. After editing and saving, it'll also update the user group on your user screen.


bielu:
I love idea of that feature but I think it will kill Backoffice in case of 1k+ users in group in actual implementation of creating Model of group, it is why I created PR (I dont like way but it is only not breaking way I think)
https://github.com/umbraco/Umbraco-CMS/pull/7819
But really happy to see that feature!

nul800sebastiaan:
@patrickdemooij9 Sorry that this one didn't get on the radar @patrickdemooij9 - this is all good! And the selector has been fixed in the mean time, so the previous sections are pre-selected when you open the editor.

PR #8311, opened 2020-06-18
Use correct logging overload in InstructionProcessTask

joepvtl:
### Description
When a error occures whilist running the InstructionProcessTask it only logs the message and not the exception. This is because it was using the wrong overload of the Error method.


joepvtl:
I see this is already fixed in an other PR.

Issue #8299, opened 2020-06-16
Sort domains and/or languages


ronaldbarendse:
It's currently not possible to sort/order the domains of a content node (under 'Culture and Hostnames') or the languages themselves, which makes the following impossible:
- Configure the primary domain to generate absolute URLs, used when no/multiple domains are found: https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Web/Routing/DomainUtilities.cs#L194
- Change the order of the content tree and content language dropdowns (it currently only sets the default language as first)
- Automatically generate a language selector on the front-end based on the published cultures/variants in a non-random order (in Umbraco 7 we used multiple root nodes per culture, so the order could be retrieved from the content tree)

abjerner:
Personally I'd like to see this very much 👍 

Currently the domains have no particular order, but the domain added first typically is treated as the primary one. But at least in Umbraco 7, we've had some sites where Umbraco at random would consider one of the other domains to be the first domain instead. So being able to sort them should prevent that from happen 😄 

Just for reference, @skttl started a PR for this in Umbraco 7.

Old issue: https://github.com/umbraco/Umbraco-CMS/issues/6591

Even older issue: https://github.com/umbraco/Umbraco-CMS/issues/3210

And the PR: https://github.com/umbraco/Umbraco-CMS/pull/3295

skttl:
I also did a PR for v8, but got rejected as I couldn't find a way to do it without making a breaking change.
#6611

ronaldbarendse:
It should be possible without making breaking changes, but that would require adding new interfaces with the added sorting functionality, e.g.:

```c#
public interface ISortableDomainRepository : IDomainRepository
{
    bool Sort(IEnumerable<IDomain> items);
}

public interface ISortableDomainService : IDomainService
{
    bool Sort(IEnumerable<IDomain> items);
}
```

The `DomainDto` is an internal class, so adding a new `int SortOrder { get; set; }` property is not a problem. This can then be used internally (within the repository) for returning the `IDomain`s in the right order, so that this interface doesn't need to change.

The same can be done for `ILanguageRepository` and `ILocalizationService` (which handles both dictionary items and languages) if we want to (primarily) be able to sort the languages within the back-office dropdowns.

[...]

PR #8316, opened 2020-06-19
Sort domains (Culture and Hostnames)

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8299.

#### Domain service and repository (database)
I've started creating all the necessary plumbing to allow sorting domains (set under 'Culture and Hostnames'):
- Adding the `SortOrder` property to the service level models (`IDomain` and `UmbracoDomain`) and database model (`DomainDto`);
- Adding the new `sortOrder` column to `umbracoDomain` and setting initial values in a migration;
- Add `Sort` method to `DomainService` (similar to those on other services);
- Ensured a new, incremental, sort order is set for new domains;
- Wildcard domains (setting the language of a content node) are not sorted by setting the sortOrder to -1, as only a single one of this type should be available per content node.

As per https://github.com/umbraco/Umbraco-CMS/issues/8299#issuecomment-645748580, I've ensured to limit breaking changes to repository/service interfaces and have only done additions to them.

These changes can be tested by following these steps:
1. Setup an Umbraco instance with the Starter Kit, but without this PR applied and:
   - Add additional languages (besides `en-US`);
   - Set the language on the root node and add some additional domains/languages:
     Culture and Hostnames
   - Do the same on another node, so ordering per node can be tested later;
   - Note the current order of domains.
2. Apply this PR and complete the upgrade, the migration should have executed and:
    - The log should include the executed SQL commands:
      ```
      Database requires upgrade
      "Executing installation step: 'DatabaseUpgrade'." [Timing "c8007c9"]
      Database upgrade started
      Running 'Upgrade' service
      Starting '"Umbraco.Core"'...
      Execute "AddDomainSortOrder"
      At "{a78e3369-8ea3-40ec-ad3f-5f76929d2b20}"
      SQL [0]: "ALTER TABLE [umbracoDomain] ADD [sortOrder] INTEGER NOT NULL CONSTRAINT [DF_umbracoDomain_sortOrder] DEFAULT ('0')"
      SQL [1]: "ALTER TABLE [umbracoDomain] DROP CONSTRAINT [DF_umbracoDomain_sortOrder]"
      SQL [2]: UPDATE [umbracoDomain] SET sortOrder = id
      SQL [3]: UPDATE [umbracoDomain] SET [umbracoDomain].[sortOrder]=@0 WHERE (LEN(domainName) = 0 OR CHARINDEX('*', domainName) = 1) -- @0:-1
      At "{ace18303-f74a-40d2-83c4-4ab24b33cbfb}"
      PostMigration: Umbraco.Web.Migrations.PostMigrations.ClearCsrfCookies.
      Done (pending scope completion).
      ```
      - The `umbracoDomain` table should have the additional `sortOrder` column with initial sort orders.
3. Check whether the domains on the configured nodes are sorted the same as before (as the auto-incremented id is set as default sort order in the migration).
4. Check whether deleting and re-adding domains will always put them last with a single incremented sort order _for that node_.
5. Check whether the wildcard domains/node languages sort order is always set to -1, even after setting it to `Inherit` and back to a language.
6. Change/reverse the sort order by editing the values in the database, restart the application (to clear the cache) and ensure the domains are returned/shown in the correct order 🎉 

#### Sorting domains in back-office
Next up is updating the `ContentController` and UI to allow re-ordering/sorting the domains. This is done by keeping track of the saved domains and sort them according to the view model that's made sort-able:

Sortable Culture and Hostnames

#### Ensuring sorting is honoured within Umbraco
The last step is to make sure the sorting is honoured where needed. For now, I've ensured the `DomainCache` and `DomainUtilities` keeps the correct sort order. Because `DefaultUrlProvider` depends on these classes, it should return URLs in the correct order/with the first domain. Custom `IUrlProvider`s might return the URLs in a different order though, but that's out of this PRs scope.


src/Umbraco.Core/Migrations/Upgrade/V_8_8_0/AddDomainSortOrder.cs
```diff
@@ -0,0 +1,42 @@
...
+                // Delete the default value constraint
+                Delete
+                    .DefaultConstraint()
+                    .OnTable(Constants.DatabaseSchema.Tables.Domain)
+                    .OnColumn("sortOrder")
+                    .Do();
+
+                // Keep exising sort order by setting it to the id
+                var updateSortOrder = Sql().Update<DomainDto>().Append("SET sortOrder = id");
+                Execute.Sql(updateSortOrder).Do();
```
ronaldbarendse, line 34: As the domains were not sorted, it used the default database order (based on automatically incremented ID), so use that as starting point for the new `sortOrder` column.

src/Umbraco.Core/Migrations/Upgrade/V_8_8_0/AddDomainSortOrder.cs
```diff
@@ -0,0 +1,42 @@
...
+
+                // Set sort order of wildcard domains to -1
+                var updateWildcardSortOrder = Sql().Update<DomainDto>(d => d.Set(f => f.SortOrder, -1)).Where("LEN(domainName) = 0 OR CHARINDEX('*', domainName) = 1");
+                Execute.Sql(updateWildcardSortOrder).Do();
```
ronaldbarendse, line 38: Content can only have a single wildcard domain (node language), so we don't require sorting those and use -1.

src/Umbraco.Core/Persistence/Factories/DomainFactory.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Umbraco.Core.Models;
+using Umbraco.Core.Persistence.Dtos;
+
+namespace Umbraco.Core.Persistence.Factories
+{
+    internal static class DomainFactory
```
ronaldbarendse, line 6: Moved from internal sub-class in `DomainRepository` and now aligns with other factories within this namespace.

src/Umbraco.Core/Persistence/Repositories/Implement/DomainRepository.cs
```diff
@@ -8,45 +8,39 @@
...
 using Umbraco.Core.Models.Entities;
 using Umbraco.Core.Persistence.Dtos;
+using Umbraco.Core.Persistence.Factories;
 using Umbraco.Core.Persistence.Querying;
 using Umbraco.Core.Scoping;
 
 namespace Umbraco.Core.Persistence.Repositories.Implement
 {
-    // TODO: We need to get a readonly ISO code for the domain assigned
-
```
ronaldbarendse, line 17: This is already fixed by doing a left join on the language table in `GetBaseQuery()`.

src/Umbraco.Core/Services/Implement/DomainService.cs
```diff
@@ -91,11 +92,55 @@ public Attempt<OperationResult> Save(IDomain domainEntity)
...
                 scope.Complete();
+
+                saveEventArgs.CanCancel = false;
+                scope.Events.Dispatch(Saved, this, saveEventArgs);
+            }
+
+            return OperationResult.Attempt.Succeed(evtMsgs);
+        }
+
+        public Attempt<OperationResult> Sort(IEnumerable<IDomain> items)
```
ronaldbarendse, line 109: This works exactly like media: if the sort order doesn't match the index of the items, it's updated and saved.

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -1763,85 +1766,89 @@ public DomainSave PostSaveLanguageAndDomains(DomainSave model)
...
 
-            // process domains
-            // delete every (non-wildcard) domain, that exists in the DB yet is not in the model
-            foreach (var domain in domains.Where(d => d.IsWildcard == false && model.Domains.All(m => m.Name.InvariantEquals(d.DomainName) == false)))
+            // Delete every domain that's in the database, but not in the model
+            foreach (var assignedDomain in assignedDomains.Where(d => (d.IsWildcard && language == null) || (d.IsWildcard == false && model.Domains.All(m => m.Name.InvariantEquals(d.DomainName) == false))))
             {
-                Services.DomainService.Delete(domain);
+                Services.DomainService.Delete(assignedDomain);
             }
```
ronaldbarendse, line 1774: This now removes all domains not required by the model, so including the wildcard domain.

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -1763,85 +1766,89 @@ public DomainSave PostSaveLanguageAndDomains(DomainSave model)
...
+                    response.Content = new StringContent("Saving domain failed");
+                    response.ReasonPhrase = saveAttempt.Result.Result.ToString();
+                    throw new HttpResponseException(response);
+                }
+
+                savedDomains.Add(domain);
             }
 
+            // Sort saved domains
+            Services.DomainService.Sort(savedDomains);
```
ronaldbarendse, line 1850: This ensures the domains sort order is updated correctly! 🚀

src/Umbraco.Web/Routing/Domain.cs
```diff
@@ -61,5 +78,13 @@ protected Domain(Domain domain)
...
         /// </summary>
         public bool IsWildcard { get; }
+
+        /// <summary>
+        /// Gets the sort order.
+        /// </summary>
+        /// <value>
+        /// The sort order.
+        /// </value>
+        public int SortOrder { get; }
```
ronaldbarendse, line 88: The `Domain` model (used in `IDomainCache`) also requires a sort order, as the cache doesn't keep the order of the collections.

src/Umbraco.Web/Routing/DomainUtilities.cs
```diff
@@ -150,15 +150,10 @@ internal static DomainAndUri SelectDomain(IEnumerable<Domain> domains, Uri uri,
         {
             // sanitize the list to have proper uris for comparison (scheme, path end with /)
             // we need to end with / because example.com/foo cannot match example.com/foobar
-            // we need to order so example.com/foo matches before example.com/
-            var domainsAndUris = domains
-                .Where(d => d.IsWildcard == false)
-                .Select(d => new DomainAndUri(d, uri))
-                .OrderByDescending(d => d.Uri.ToString())
```
ronaldbarendse, line 157: Ordering by URL is only required for `SelectByBase()`, so more specific URLs get matched first and therefore removed here to keep the correct domain sort order.

src/Umbraco.Web/Routing/DomainUtilities.cs
```diff
@@ -264,7 +262,7 @@ private static DomainAndUri GetByCulture(IReadOnlyCollection<DomainAndUri> domai
                 if (domainAndUri != null) return domainAndUri;
             }
 
-            return domainsAndUris.First(); // what else?
+            return domainsAndUris.FirstOrDefault(); // what else?
```
ronaldbarendse, line 265: Although 0 domains is already handled within `SelectDomain()` (returning null), let's just ensure this does the same...

src/Umbraco.Web/Routing/DomainUtilities.cs
```diff
@@ -345,14 +331,17 @@ internal static Domain FindDomainInPath(IEnumerable<Domain> domains, string path
...
             var stopNodeId = rootNodeId ?? -1;
 
             return path.Split(',')
                        .Reverse()
                        .Select(int.Parse)
                        .TakeWhile(id => id != stopNodeId)
-                       .Select(id => domains.FirstOrDefault(d => d.ContentId == id && d.IsWildcard))
+                       .Select(id => domains.FirstOrDefault(d => d.ContentId == id && d.IsWildcard == isWildcard))
                        .FirstOrDefault(domain => domain != null);
         }
```
ronaldbarendse, line 346: I've created this private method to remove the duplicated code in `FindDomainInPath()` and `FindWildcardDomainInPath()`.

src/Umbraco.Core/Persistence/Mappers/DomainMapper.cs
```diff
@@ -19,6 +19,7 @@ protected override void DefineMaps()
             DefineMap<UmbracoDomain, DomainDto>(nameof(UmbracoDomain.RootContentId), nameof(DomainDto.RootStructureId));
             DefineMap<UmbracoDomain, DomainDto>(nameof(UmbracoDomain.LanguageId), nameof(DomainDto.DefaultLanguage));
             DefineMap<UmbracoDomain, DomainDto>(nameof(UmbracoDomain.DomainName), nameof(DomainDto.DomainName));
+            DefineMap<UmbracoDomain, DomainDto>(nameof(UmbracoDomain.SortOrder), nameof(DomainDto.SortOrder));
```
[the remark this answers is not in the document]
ronaldbarendse: Seems like a mapper can be used instead of the factory methods, but I can't find documentation about the differences (most factories in `Umbraco.Core.Persistence.Factories` also have a mapper in `Umbraco.Core.Persistence.Mappers`).

ronaldbarendse:
The changes in `DomainUtilities` do seem to work, as the `Umbraco.Tests.Routing.UrlsProviderWithDomainsTests.Get_Url_Alternate` test now returns the domains in a different order 😏 I'll add explicit sort orders in the `UmbracoDomain` mocks (instead of setting them all to 0) and fix the assertions order!

ronaldbarendse:
@nul800sebastiaan I've merged the latest changes from `v8/contrib` again and verified everything still works. Would be great to have this feature added to v8 and merged to v9 before the final version is released. The migration currently targets 8.16, but changing this to a later version would be a small change to the `UmbracoPlan` and the directory/namespace of the migration 😉

Although this PR adds a new property on the `IDomain`/`UmbracoDomain` models, that shouldn't be a major breaking change (it is even [documented as non-breaking](https://our.umbraco.com/documentation/development-guidelines/breaking-changes#non-breaking)).

------------------------

This feature allows changing the primary domain that's returned when generating URLs (if the current request isn't associated to a domain):

Demo with domain names

------------------------

But I recon the best thing about this would be to allow changing the order of a language dropdown for culture variant content:

Demo with culture variants

In its most simplistic form, this could be achieved using the following code (inside the master template):
```cshtml
@foreach (var domain in UmbracoContext.PublishedSnapshot.Domains.GetAssigned(Model.Root().Id))
{
    <a href="@Model.Url(domain.Culture.Name)">@domain.Culture.DisplayName</a>
}
```

A more concise way would be to use the domain from the current request (and only fall-back to the root node ID when not found) and check whether the culture is actually published for that content. I would recommend adding this code in a child action of a surface controller, populate a view model and render this using a partial view. This code would then be something like this:

```cshtml
@{
    var currentPage = Model; // Only this variable and the UmbracoContext are required for the following code
    
    var rootId = UmbracoContext.PublishedRequest.HasDomain ? UmbracoContext.PublishedRequest.Domain.ContentId : currentPage.Root().Id;
    foreach (var domain in UmbracoContext.PublishedSnapshot.Domains.GetAssigned(rootId))
    {
        var culture = domain.Culture.Name;
        string url;
        if (currentPage.IsPublished(culture))
        {
            // Use the URL of the current page in the requested culture
            url = currentPage.Url(culture);
        }
        else if (currentPage.Id != rootId && UmbracoContext.PublishedSnapshot.Content.GetById(rootId) is IPublishedContent root && root.IsPublished(culture))
        {
            // Fallback to the root page, so even if the current page isn't available in the requested culture, we can switch to this culture
            url = root.Url(culture);
        }
        else
        {
            // Both the current and root page aren't published in the requested culture
            continue;
        }
        
        // Do something with the URL
        <a class="nav-link" href="@url">@domain.Culture.DisplayName</a>
    }
}
```

bjarnef:
@ronaldbarendse @nul800sebastiaan is this something that is planned to include in any near future? 🙈 
I can't count the number of times I have wanted to sort the domains added on a site/root node.
Often I prepare domains on the site node in Umbraco close the launch of a site on Umbraco Cloud, but it seems if the `project-alias.s1.umbraco.io` or `project-alias.euwest01.umbraco.io` domain isn't the first one, menu item in frontend use the real domain. So currently they need to be added in the correct order.

PR #8323, opened 2020-06-22
Netcore: Migrated remaining backoffice controllers

bergmania:
# Notes
- Updated generated templates to use the current namespace of `UmbracoViewPage` and `PartialViewMacroPage`
- Fixed some issues with content-root vs web-root
- Fixed some test builders
- Migrated TestServer tests (Explicit until we have a solution for build server)
- Added AutoFixture and created an `AutoMoqDataAttribute` in the unit test project, to simplify tests of UsersController and can be used in the future
- Fixed issues when services injected `IPasswordConfiguration`. Which is not registered in the container. Instead, I inject `IUserPasswordConfiguration`
- Fixed issue in `DetermineAmbiguousActionByPassingParameters` when null-values are valid.
- Added filter to output exceptions as JSON on `UmbracoAuthorizedJsonController`
- Added new options to `LinkGeneratorExtensions` to generate links with multiple query params
- Fixed issue with the `UmbracoJsonModelBinder` so now `[Required]` attributes are not required to model bind
   - The issue here is e.g. `UserSave` is not model bound and controller throws ``ArgumentNullException`.
- Migrated the following controllers:
    - `CurrentUserController`
    - `MemberGroupController`
    - `TemplateQueryController`
    - `UserGroupsController`
    - `UsersController`

- Migrated the following actions into existing controllers
    - `AuthenticationController.GetPasswordConfig`

- Finally also removed a lot of files that is not used anymore in the `Umbraco.Web` project

### Questions
- How can we test FIPS compliance for users? Old tests commented out (`UsersControllerTests`)

### Tests
- In general, BackOffice should work. Espespecially error/validation messages.
- Known issues
   - The Cultures set on the thread is not chosen from the user signing in but is using the machine default (I have danish 👎 )
   - The users are not updated on login, so the user is set marked as inactive


src/Umbraco.Web.Common/Formatters/IgnoreRequiredAttributsResolver.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Reflection;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Umbraco.Web.Common.Formatters
+{
+    public class IgnoreRequiredAttributsResolver : DefaultContractResolver
```
stevetemple, line 7:
```suggestion
    public class IgnoreRequiredAttributesResolver : DefaultContractResolver
```

src/Umbraco.Tests.Common/Builders/UserBuilder.cs
```diff
@@ -76,12 +79,20 @@ public UserBuilder<TParent> WithLanguage(string language)
             return this;
         }
 
+   public UserBuilder<TParent> WithUsername(string username)
```
stevetemple, line 82: Formatting: spacing

src/Umbraco.Web.Common/ModelBinding/UmbracoJsonModelBinderProvider.cs
```diff
@@ -25,6 +29,11 @@ private static IInputFormatter[] GetNewtonsoftJsonFormatter(ILoggerFactory logge
             {
                 AllowInputFormatterExceptionMessages = true
             };
+
+            var ss = jsonOptions.SerializerSettings; // Just use the defaults as base
+
+            // We need to ignore required attributes when serializing. E.g UserSave.ChangePassword. Otherwise the model is not model bound.
+            ss.ContractResolver = new IgnoreRequiredAttributsResolver();
```
stevetemple, line 36:
```suggestion
            ss.ContractResolver = new IgnoreRequiredAttributesResolver();
```

src/Umbraco.Web.BackOffice/Filters/UserGroupValidateAttribute.cs
```diff
@@ -0,0 +1,101 @@
...
+            //now assign the persisted entity to the model so we can use it in the action
+            userGroupSave.PersistedUserGroup = persisted;
+
+            var existing = _userService.GetUserGroupByAlias(userGroupSave.Alias);
+            if (existing != null && existing.Id != userGroupSave.PersistedUserGroup.Id)
+            {
+                context.ModelState.AddModelError("Alias", "A user group with this alias already exists");
+            }
+
+            // TODO: Validate the name is unique?
```
stevetemple, line 83: TODO

src/Umbraco.Web.BackOffice/Filters/UserGroupValidateAttribute.cs
```diff
@@ -0,0 +1,101 @@
...
+                    }
+
+                    if (persisted.Alias != userGroupSave.Alias && persisted.IsSystemUserGroup())
+                    {
+                        var message = $"User group with alias: {persisted.Alias} cannot be changed";
+                        context.Result = new UmbracoErrorResult(HttpStatusCode.BadRequest, message);
+                        return;
+                    }
+
+                    //map the model to the persisted instance
```
stevetemple, line 61: These comments here and just below are more what rather than why

src/Umbraco.Web.BackOffice/Filters/UserGroupValidateAttribute.cs
```diff
@@ -0,0 +1,101 @@
...
+              private readonly IUserService _userService;
+              private readonly UmbracoMapper _umbracoMapper;
+
+              public UserGroupValidateFilter(
+            IUserService userService,
+            UmbracoMapper umbracoMapper)
+        {
+            _userService = userService ?? throw new ArgumentNullException(nameof(userService));
+            _umbracoMapper = umbracoMapper ?? throw new ArgumentNullException(nameof(umbracoMapper));
+        }
```
stevetemple, line 32: Formatting: Spacing

src/Umbraco.Web.BackOffice/Security/PasswordChanger.cs
```diff
@@ -0,0 +1,102 @@
...
+            IUser savingUser,
+            ChangingPasswordModel passwordModel,
+            BackOfficeUserManager userMgr)
+        {
+            if (passwordModel == null) throw new ArgumentNullException(nameof(passwordModel));
+            if (userMgr == null) throw new ArgumentNullException(nameof(userMgr));
+
+            if (passwordModel.NewPassword.IsNullOrWhiteSpace())
+            {
+                return Attempt.Fail(new PasswordChangedModel { ChangeError = new ValidationResult("Cannot set an empty password", new[] { "value" }) });
```
stevetemple, line 42: Can these strings be localised?

src/Umbraco.Web/Editors/PasswordChanger.cs
```diff
@@ -9,6 +9,7 @@
 using Umbraco.Web.Security;
 using IUser = Umbraco.Core.Models.Membership.IUser;
 
+//Migrated to .NET CORE
```
stevetemple, line 12: Can this be deleted

src/Umbraco.Tests.Common/Builders/UserBuilder.cs
```diff
@@ -113,7 +124,7 @@ public UserBuilder<TParent> WithSuffix(string suffix)
 
         public override User Build()
         {
-            var id = _id ?? 1;
+            var id = _id ?? 0;
```
stevetemple, line 127: This feels intentional but possibly breaking, not sure why it was 1 previously but thought I'd raise in case it's accidental

src/Umbraco.Web.BackOffice/Filters/UserGroupValidateAttribute.cs
bergmania: Not a new TODO.. Just copied over from v8

src/Umbraco.Web.BackOffice/Filters/UserGroupValidateAttribute.cs
bergmania: Agree.. But again, these are just copied over one to one

src/Umbraco.Web.BackOffice/Security/PasswordChanger.cs
bergmania: I think yes, but not really related to this PR. This is just copied over. We could make an up-for-grabs task about localizing them

src/Umbraco.Web/Editors/PasswordChanger.cs
bergmania: Nope, not yet.. Still used in members implementation

src/Umbraco.Tests.Common/Builders/UserBuilder.cs
bergmania: Yes.. But this is just the Test builder. I think it was by mistake 1 before.. All tests updated

bergmania:
Thanks, @stevetemple..

Good feedback. I think all of the concerns are relevant, but not necessarily for this PR, as all the issues existed before.

I fixed the misspelling and formatting issues.

src/Umbraco.Web.Common/Formatters/IgnoreRequiredAttributesResolver.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Reflection;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Umbraco.Web.Common.Formatters
+{
+    public class IgnoreRequiredAttributesResolver : DefaultContractResolver
```
emmagarland, line 7: Can we create tests for this new resolver?

src/Umbraco.Web.BackOffice/Controllers/MemberGroupController.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        public IEnumerable<MemberGroupDisplay> GetByIds([FromQuery]int[] ids)
+        {
+            return _memberGroupService.GetByIds(ids)
+                    .Select(_umbracoMapper.Map<IMemberGroup, MemberGroupDisplay>);
+        }
+
+        [HttpDelete]
+        [HttpPost]
```
emmagarland, line 59: I know this was already here, just wondered why we might have POST here too

src/Umbraco.Web/Runtime/WebInitialComposer.cs
```diff
@@ -31,7 +31,6 @@ public override void Compose(Composition composition)
 
             composition.Register<UmbracoInjectedModule>();
 
-            composition.ComposeWebMappingProfiles();
```
emmagarland, line 34: I don't know what the implications of this will be since it has been deleted, but are we registering all the previous components somewhere else now (or future task)?

src/Umbraco.Web.BackOffice/Controllers/UsersController.cs
```diff
@@ -147,30 +172,24 @@ internal static async Task<HttpResponseMessage> PostSetAvatarInternal(HttpReques
...
-                using (var fs = System.IO.File.OpenRead(file.LocalFileName))
+                using (var fs = file.OpenReadStream())
                 {
                     mediaFileSystem.AddFile(user.Avatar, fs, true);
                 }
 
                 userService.Save(user);
-
-                //track the temp file so the cleanup filter removes it
-                tempFiles.UploadedFiles.Add(new ContentPropertyFile
```
emmagarland, line 158: Was this not needed then?

emmagarland:
Awesome to have the .NET Core Umbraco back-office running locally! 🥇 

Will put any issues found so far on here. Let it be known I have break on **_all_** exceptions switched on!! Some of these maybe known or wider issues we can break out.

**Testing Results so far (only just dipped into the back-office atm)**

1. I usually default to SQLCE and not SQLServer when testing/running locally, but this didn't work as it needs a SQLCE project reference to do so. However, SQLCE being the default Umbraco DB is usual behaviour for the current v8 install. As you prob have thought of some people who run Umbraco for dev may not have full SQL or just need something spun up quickly, so perhaps will expect the same as I did by default, like previous Umbraco versions.
2. When I customise the install and choose a brand new database, I see some `SqlServerSyntaxProvider.cs` exceptions being thrown when initially creating the new SQLServer database. Things like `GetColumnsInSchema` method throwing an exception because `TABLE_NAME`, `COLUMN_NAME`  doesn't exist etc. I believe this was line 190 `GetColumnsInSchema`
3. All the database tables were created, along with my default user, which was cool. However, then I seemed to get into some sort of install loop. The page kept saying _"You see this screen because your Umbraco installation did not complete correctly"_. I tried to debug why but it just went through some checks in `NewInstallStep.cs` and then kept returning that view `continueinstall.cshtml`. I ended up stopping the site, clearing cookies, rerunning the app and logging in with the credentials I'd used - I logged in fine. Now I am in the .NET Core Umbraco back-office! Woohoo!
4. When logging in the first time, there were a few errors being thrown behind the scenes, but they didn't break me out of the login process so I temporarily passed over them to focus on the goal to look at the back-office - we can look at them later. Some might already exist - an example is an exception on an index not being present yet. 
5. While debugging with break all exceptions on, BackgroundTaskRunner kicks in with:

`System.IO.DirectoryNotFoundException: 'Could not find a part of the path 'D:\Source\Repos\Umbraco-v8\src\Umbraco-CMS\src\Umbraco.Web.UI.NetCore\wwwroot\App_Data\TEMP\FileUploads'.' from BackgroundTaskRunner`


Thanks

src/Umbraco.Web.Common/Formatters/IgnoreRequiredAttributesResolver.cs
bergmania: I added one :)

src/Umbraco.Web/Runtime/WebInitialComposer.cs
bergmania: I moved some of the registrations to the `ComposeCoreMappingProfiles`, and the rest is added in `BackOfficeComposer`

src/Umbraco.Tests/TestHelpers/ControllerTesting/TestStartup.cs
```diff
@@ -39,7 +40,7 @@ public void Configuration(IAppBuilder app)
             // Add in a simple exception tracer so we can see what is causing the 500 Internal Server Error
             httpConfig.Services.Add(typeof (IExceptionLogger), new TraceExceptionLogger());
 
-            httpConfig.Services.Replace(typeof (IAssembliesResolver), new SpecificAssemblyResolver(new[] { typeof (UsersController).Assembly }));
+            httpConfig.Services.Replace(typeof (IAssembliesResolver), new SpecificAssemblyResolver(new[] { typeof (AspNetApplicationShutdownRegistry).Assembly }));
```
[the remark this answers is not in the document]
bergmania: We just need the assembly of `Umbraco.Web`. `UsersController` is moved away from  `Umbraco.Web`, so needed just a new class from  `Umbraco.Web`.

src/Umbraco.Web.BackOffice/Controllers/MemberGroupController.cs
bergmania: I'm not really sure, but at least the client are using POST. Whether delete is used, I don't know.

src/Umbraco.Web.BackOffice/Controllers/UsersController.cs
bergmania: It is not needed anymore. The reason is because in framework all files were uploaded to a temp location before the controller action was executed. Therefore we had to clean up (using the `FileUploadCleanupFilterAttribute`).

This is not the case anymore, and we handle the stream directly in the action, so nothing to clean up,

bergmania:
Thanks, @emmagarland ..

I have added the SqlCE project reference to the netcore executable so now you can use SqlCE as default (on windows). The issue now is when you try from non-windows, this will end up failing. That's a problem for later.

I can't repro the install loop. I tried to install using LocalDb and SqlCE - Both succeed.

Regarding the review, I added the suggested tests and answered on the comments.

src/Umbraco.Tests.UnitTests/Umbraco.Web.Common/Formatters/IgnoreRequiredAttributesResolverUnitTests.cs
```diff
@@ -0,0 +1,41 @@
...
+
+                Assert.IsNotNull(actual);
+                Assert.IsNull(actual.Property);
+            });
+        }
+
+        [DataContract(Name = "objectWithRequiredProperty", Namespace = "")]
+        private class ObjectWithRequiredProperty
+        {
+            [DataMember(Name = "property", IsRequired = true)]
```
Shazwazza, line 37: So am I correct in assuming that this new IgnoreRequiredAttributesResolver is specifically for deserializing JSON so it actually deserializes without an exception even when IsRequired is there *but* the IsRequired will still be used during validation? Or is the validation Required attribute different? If so, is there ever a need to use the IsRequired param?

src/Umbraco.Tests/Web/Controllers/UsersControllerTests.cs
```diff
@@ -1,526 +1,526 @@
...
-        {
-            return new Mock<BackOfficeOwinUserManager>(
-                new Mock<IPasswordConfiguration>().Object,
-                new Mock<IIpResolver>().Object,
-                new Mock<IUserStore<BackOfficeIdentityUser>>().Object,
-                null, null, null, null, null, null, null);
-        }
-    }
-}
+﻿// using System;
```
Shazwazza, line 1: Remove this file?

src/Umbraco.Web.BackOffice/Filters/UserGroupValidateAttribute.cs
```diff
@@ -0,0 +1,101 @@
...
+                throw HttpResponseException.CreateValidationErrorResponse(context.ModelState);
+            }
+        }
+
+        public void OnActionExecuted(ActionExecutedContext context)
+        {
+
+        }
+
+        }
```
Shazwazza, line 97: Is this an extra } ? don't think the build is failing but seems odd

src/Umbraco.Web.UI.NetCore/Umbraco.Web.UI.NetCore.csproj
```diff
@@ -61,6 +62,23 @@
           <SubType>Designer</SubType>
         </Content>
         <Content Remove="wwwroot\Web.config" />
+        <Content Remove="Umbraco\PartialViewMacros\Templates\Gallery.cshtml" />
```
Shazwazza, line 65: Do we need all these remove statements?

src/Umbraco.Web.UI.NetCore/Umbraco/PartialViewMacros/Templates/Breadcrumb.cshtml
```diff
@@ -18,7 +18,7 @@
         @* For each page in the ancestors collection which have been ordered by Level (so we start with the highest top node first) *@
         @foreach (var item in selection.OrderBy(x => x.Level))
         {
-            <li><a href="@item.Url()">@item.Name</a> <span class="divider">/</span></li>
+            <li><a href="@item.Url(PublishedUrlProvider)">@item.Name</a> <span class="divider">/</span></li>
```
Shazwazza, line 21: This will be an interesting change for people if we make this a requirement. I'm all for it since the model shouldn't be responsible for generating the URL but I think if we make the change we shouldn't use ext methods on the model to do it and instead just call the service. But we can discuss later.

bergmania: I guess we will end up having an extra nuget package. with the old codesmells. So having the old extention methods available. but we should not use these our self, or in templates.. Code is untestable using these..

`Umbraco.Friendly` or similar sounds good ;)
At least better than `Umbraco.ReintroduceCodeSmells` or `Umbraco.ReintroduceUntestableCode` :P

src/Umbraco.Web.UI.NetCore/Umbraco.Web.UI.NetCore.csproj
bergmania: Nope.. I will clean up in a later PR regarding the publish for alpha

src/Umbraco.Web.BackOffice/Filters/UserGroupValidateAttribute.cs
bergmania: I will reformat entire fill to fix up :)

src/Umbraco.Tests.UnitTests/Umbraco.Web.Common/Formatters/IgnoreRequiredAttributesResolverUnitTests.cs
bergmania: Yes..
The `RequiredAttribute` is different. I guess that is just used invalidation. 

Maybe we could just remove the `DataMember(IsRequired = true)`. I assumed it was there because we require it hen we output the same model.

src/Umbraco.Tests.Common/Builders/UserGroupBuilder.cs
```diff
@@ -92,7 +92,7 @@ public IReadOnlyUserGroup BuildReadOnly(IUserGroup userGroup)
 
         public override IUserGroup Build()
         {
-            var id = _id ?? 1;
+            var id = _id ?? 0;
```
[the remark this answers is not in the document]
bergmania: 0 is intended.

This is the userGroup builder... But for user builder 0 is also intended, so it can be created by default. If you need an admin, then you can use WithId(Constants.Security.SuperUserId)

src/Umbraco.Infrastructure/Scheduling/BackgroundTaskRunner.cs
```diff
@@ -137,17 +137,17 @@ public BackgroundTaskRunner(BackgroundTaskRunnerOptions options, ILogger logger,
         /// <param name="name">The name of the runner.</param>
         /// <param name="options">The set of options.</param>
         /// <param name="logger">A logger.</param>
-        /// <param name="hostingEnvironment">The hosting environment</param>
+        /// <param name="applicationShutdownRegistry">The hosting environment</param>
```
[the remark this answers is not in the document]
bergmania: Fixing.. But not really sure what value these param-xml docs are given in 99% of the cases, we just humanize the variable name.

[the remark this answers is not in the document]
Shazwazza: i know, it's just they get used to generate the csharp docs https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Core.html

src/Umbraco.Tests.UnitTests/Umbraco.Web.Common/Formatters/IgnoreRequiredAttributesResolverUnitTests.cs
Shazwazza: maybe, i think we could revisit it, maybe for now we'll just leave this as-is. seems strange that it worked in v8, or maybe that's only because we switched everything to DataMember stuff in netcore and we didn't have that setting?

src/Umbraco.Web.UI.NetCore/Umbraco/PartialViewMacros/Templates/Breadcrumb.cshtml
Shazwazza: Yep we'll see how it goes, i think the more we move away from the hacks the better and then can re-implement 'friendly' things later if necessary. In some cases though i think we'll realize that some of the 'friendly' things don't actually make much sense. For example, when (one day) we decouple the content cache from URLs, then item.Url() probably won't make sense anymore.

src/Umbraco.Tests.UnitTests/Umbraco.Web.Common/Formatters/IgnoreRequiredAttributesResolverUnitTests.cs
bergmania: I think it is because we have our own UmbracoJsonModelBinderConvention now..

Issue #7675, opened 2019-11-28
"[LogScrubber] " Task has failed   if \App_Data\TEMP\FileUploads doesn't exist


mistyn8:
Umbraco 8.3.0 Forms 8.2.0 Deploy 3.2.0

```
Message:
"[LogScrubber] " Task has failed

Exception:
System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\home\site\wwwroot\App_Data\TEMP\FileUploads'.
```
Seems to be if a new cloud site created and no interaction with media has caused the FileUploads folder to be created?

umbrabot:
Hi @mistyn8,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

ronaldbarendse:
@nul800sebastiaan The task is indeed added to the wrong runner, so that should get fixed, but the main issue is in:

https://github.com/umbraco/Umbraco-CMS/blob/5cfdfd6e9cb96ffba1f9b0f0c21041683037262b/src/Umbraco.Web/Scheduling/TempFileCleanup.cs#L50-L56

This still gets all files from a non-existant directory and throws the exception (see stack trace):
```
System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\home\site\wwwroot\App_Data\TEMP\FileUploads'.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileSystemEnumerableIterator`1.CommonInit()
   at System.IO.FileSystemEnumerableIterator`1..ctor(String path, String originalUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler`1 resultHandler, Boolean checkHost)
   at System.IO.DirectoryInfo.InternalGetFiles(String searchPattern, SearchOption searchOption)
   at System.IO.DirectoryInfo.GetFiles(String searchPattern, SearchOption searchOption)
   at Umbraco.Web.Scheduling.TempFileCleanup.CleanupFolder(DirectoryInfo dir)
[...]

PR #8326, opened 2020-06-23
Fix LogScrubber exception when cleaning up temp files

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7675.

### Description
This adds the `TempFileCleanup` task to the right background runner, skips nonexistent directories and also allows read-only files to be deleted (by removing this flag before deletion).

Testing can be done by ensuring the `~\App_Data\TEMP\FileUploads` directory doesn't exist, optionally set the logging level to debug (so you can see the `The cleanup folder doesn't exist...` message), wait for the background task to run (default 3 minutes delay) and check whether the exception/error isn't logged anymore.

### DIFF
--- a/src/Umbraco.Web/Scheduling/SchedulerComponent.cs
+++ b/src/Umbraco.Web/Scheduling/SchedulerComponent.cs
@@ -177,7 +177,7 @@ namespace Umbraco.Web.Scheduling
                 new[] { new DirectoryInfo(IOHelper.MapPath(SystemDirectories.TempFileUploads)) },
                 TimeSpan.FromDays(1), //files that are over a day old
                 _runtime, _logger);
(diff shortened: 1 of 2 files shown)


ronaldbarendse:
Regarding read-only files: trying to delete those will throw an `UnauthorizedAccessException` with the message `path specified a read-only file.` (see: https://docs.microsoft.com/en-us/dotnet/api/system.io.file.delete?redirectedfrom=MSDN&view=netframework-4.7.2#exceptions).

nul800sebastiaan:
Love it @ronaldbarendse! Thanks for the fixes! 👍 Good one on the read-only files as well!

PR #8339, opened 2020-06-26
Complex Editor Validation

Shazwazza:
This PR adds support for server side validation for complex editors. Complex editors are those that store JSON structures that are based on Element Types.

## How it works?

We needed to do this in a way that wasn't going to have breaking changes and try to do it with the least code changes involved. The original idea(s) were documented here: https://gist.github.com/Shazwazza/ad9fcbdb0fdacff1179a9eed88393aa6 which talks about using JSON Path's as validation 'keys'. That general idea was still followed but we don't use JSON Paths and the way it works in the end is quite a bit different from that doc, however that doc is a nice guide on how the current server validation works and the end result of adding complex server validation doesn't really change how non-complex server validation works which is nice.

### A re-cap (see gist if you want full info):  

* angularjs sends http post to the server, .NET webapi does stuff with the json model posted and for any errors it returns in the standard webapi `ModelState` dictionary
* For content editing, this ModelState dictionary is populated based on `ValidationResult` (part of .NET's `System.ComponentModel.DataAnnotations`, which is just member name + error message) objects returned from property editors
* This dictionary is key value and for non-content editors, like the data type or template editor, the keys returned are 'simple', they might be named something like "dataTypeAlias" and the error is a string which contains the error message. On the client side a server side validator directive called `valServerField` which is attached to any input control with `ngModel` (i.e. `<input type='text' name='dataTypeAlias' val-server-field="dataTypeAlias" />`) will wire up the validation returned in ModelState to the value defined in val-server-field. There's a few services like serverValidationManager that manages this data.
* Content editing is slightly different because of it's complexity and due to legacy implementations from v7 has an interesting format. It still uses ModelState but the dictionary's key's can contain special keys that are treated differently. They string key can be a "." delimited string where each part represents what property the error belongs to. If the first part is equal to the string `__Properties` then we know it's a property error. An example key looks like `_Properties.title.invariant.null.value` meaning it's for the `title` property for the `invariant` culture, non segment and for an input field with a directive of `val-server` applied to it with a value of `value` (i.e.  `<input type='text' name='textField' val-server="value" />` ). Only the first 2 parts are required, the rest are optional (culture, segment, field) and culture and segment can both be the value 'null'. `val-server` is different from `val-server-field` and is used only for content property validation. There is also `val-property-msg` which is the big red validation bar that commonly will say "property has errors". If a property editor doesn't have a specific sub field to wire up an error message to, server validation can be assigned to the whole property, for example, with this key: `_Properties.title`. 

### How does complex validation fit in?

The same ModelState key/value system is used, however for the value part this can now be JSON instead of a simple error msg string. We still use `ValidationResult` to populate the ModelState, except now we have some sub-classes of `ValidationResult` which allow us to nested validation results within validation results to any level. Then a custom JSON serializer will serialize that custom validation result into the JSON value for a ModelState key/val. The key for the ModelState remains the same and as you'll see we re-use the same concept for nested errors. An example/description of this this JSON value is:

image

The requirements for this to work is that each 'block' (or 'row' in nested content) has it's own GUID id since it's essentially a piece of content. 

On the client side we need to transform this data into a flat structure. Each server error in the client side app is tracked as a "validation path". Previously this was simply the property alias but now that value is a path that defines where the error exists within a property. For example, lets say the above JSON value was returned for this property key: "_Properties.myProperty.en-US.null", then the flattened format of this would be:

```json
[
  { "path": "myProperty", "culture": "en-US", "segment": null, "field": null, "error": "" },
  { "path": "myProperty/f8bf0a48-4f4d-4fea-8d49-1e863c6ebce1/featureName", "culture": "en-US", "segment": null, "field": null, "error": "Must be the word Hello" },
  { "path": "myProperty/f8bf0a48-4f4d-4fea-8d49-1e863c6ebce1/subFeatures", "culture": "en-US", "segment": null, "field": null, "error": "" },
  { "path": "myProperty/f8bf0a48-4f4d-4fea-8d49-1e863c6ebce1/subFeatures/979a5c0c-3a4d-4801-bccf-c19558f90c13/featureName", "culture": "en-US", "segment": null, "field": null, "error": "Must be the word Hello"},
  { "path": "myProperty/f8bf0a48-4f4d-4fea-8d49-1e863c6ebce1/subFeatures/979a5c0c-3a4d-4801-bccf-c19558f90c13/featureDetails", "culture": "en-US", "segment": null, "field": null, "error": "Must be the word Hello"},
  { "path": "myProperty/f8bf0a48-4f4d-4fea-8d49-1e863c6ebce1/subFeatures/f077fd75-23fc-4cfa-b037-3db60539f54d/featureName", "culture": "en-US", "segment": null, "field": null, "error": "Must be the word Hello"},
  { "path": "myProperty/f8bf0a48-4f4d-4fea-8d49-1e863c6ebce1/subFeatures/f077fd75-23fc-4cfa-b037-3db60539f54d/featureDetails", "culture": "en-US", "segment": null, "field": null, "error": "Must be the word Hello"}
]
```

With this structure in JS we can track if there's still server error registered for a given path based on a path match which could be an exact match, a prefix, suffix or contains match. For example, the highest level property will know if there's still server side errors applied to it if any errors match the path prefix of "myProperty". 

With complex editors, whether they are edited 'inline' or in infinite editing mode, the angular form isn't rendered as an entire form so we cannot rely only on angular validation. Also the DOM inheritance with infinite editing is broken, therefore the form and $scope inheritance is broken which means that it's not possible at all to rely on angular validation to track the form validation hierarchy.  This is now fixed (see below) and we can force angular to track the hierarchy properly which is semi integral to this all working. 

So now between having angular behave correctly with forms even in infinite editing and being able to 'watch' the server side keys based on path matching, we can achieve what we had before which is still using `val-server` and the `serverValidationManager`. A lot has needed to be changed to fully support this, see below!

### .Net

* There's some notes in classes that don't have anything to do with the changes done but we should leave these here for future reference:
  * PureLiveModelFactory.cs
  * CtorInvokeBenchmarks.cs
* Moves some c# validation logic to a centralized place: PropertyValidationService
* Creates new sub-classes of ValidationResult to support complex/nested validation structures. We then have a custom serializer for these structures to serialize the json structure that we want. This allows us to not have to modify much in c# as far as returning the current validation structure since the ToString() of validation result will normally return it's error message (which is what it was doing before) but for ToString() for these custom ValidationResult instances it will serialize to a JSON structure.
* Refactoring of the c# implementation of some parts of the block editor
  * Adds classes and logic for handling data conversions correctly and strongly typed and no more weird `IBlockEditorDataHelper` (`DataHelper`) that wasn't really a service and less non strongly typed things like JObject.
  * New base class `ComplexEditorValidator` to be used for any complex value editor which contains centralized logic for dealing with server side validation. Nested Content now also uses this.
* Fixes issues with Nested Content and refactors a few things:
  * `_nestedContentValues.GetPropertyValues` was coded in a way that was nearly impossible to follow the logic since at a glance when using this method in methods like https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs#L137, https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs#L189 it would seem this method doesn't do anything but in a strange way, the out param used there was being exploited because it was a reference to other values so changing the out parameter was changing the data !?!! Further, this strange code meant it was creating an object per json row of NC instead of creating a model and then strangely updating the JSON data in the background based on data references.  This is the most confusing code ever so it is now fixed and readable and strongly typed.
* Adds unit tests for new validation structure

### AngularJs

* Fixes error handling in promises in the content editor so that we aren't swallowing errors which made things near impossible to debug. This also fixes an issue with the underlying requesthelper service to ensure that http promise errors are bubbled up correctly.
* Some error handling checking in umbeditorcontentheader.directive.js - we were never checking if `scope.vm.defaultVariant` was available and in some cases it was not and would throw a JS error. We should have been checking this before. I unfortunately can't remember how to reproduce the exception i was seeing but it was there and you can see by the code changes this would need to be checked, it was definitely missing.
* Adds functionality to infinite editing to allow maintaining the parent $scope between the normal editor which launched the infinite editor. Without this the $scope hierarchy is totally lost which means you cannot traverse scopes but it also means that events emitted or broadcast do not follow the correct $scope chain. By keeping the scope chain connected events also continue to work as expected. Also adds functionality to maintain the FormController hierarchy between the normal and and the launched infinite editor. This means that when the child form (in the infinite editor) is connected, it's changes will affect the parent form which is required if you have forms that launch sub forms. This means changes to the child forms and the $pristine or $dirty states are tracked upwards correctly. This is in umbeditors.directive.js and another sub directive `EditorRepeaterDirective` was needed to make this work so that we could control the linking/$compiling of each editor in the list of infinite editors.
* A new option for the formHelper.submitForm: `keepServerValidation` which is needed in order to not always clear all server validation errors anytime a form is closed - which would occur anytime an infinite editor is closed! In almost all cases we don't want to clear server validation.
* umbProperty directive converted to a real component. Complex validation works by having a 'key' for a complex property and umbProperty is responsible for generating that key based on the rendered editor by directive inheritance. This is what the new `getValidationPath` method is for. All complex editors based on element types use umbProperty to render each property. However, due to the way AngularJs works with directive inheritance, it relies on the DOM but with infinite editing the DOM hierarchy is broken but because we've now allowed for maintaining the $scope chain, we can lookup the inherited umbProperty directive in it's $scope chain. umbProperty now takes in an `elementKey` value which is used to build up the validation path.
* valFormManager refactored a bit. This now all also uses the $scope chain to lookup it's parent in case the DOM hierarchy is broken with infinite editing. Moves the logic to watch form errors to angularHelper.countAllFormErrors and makes that logic readable since it was totally not before and in fact it was broken and wasn't traversing the form errors correctly. the API `showValidation` was never working because it was only assigning a primitive value to `this` which would never change so that public property has been converted to a method: `isShowingValidation()`, I suppose this is a "breaking change" but we never used this API within our own code and it's not a documented API so according to our guidelines it's not actually a "breaking" public API (see https://our.umbraco.com/apidocs/v8/ui/#/api/umbraco.directives.directive:valFormManager ... nothing is documented here as a public API)
* `valPropertyMsg` has a lot of changes. This is responsible for showing the big red error on top of properties that will say "this property has errors' if there are many or it will contain the actual error message if it can detect it. This required quite a lot of changes because it's one of the more complicated issues to handle because for complex server errors these errors need to be cleared when all other child errors are cleared and errors attached to itself are cleared. A good change for this however is that it no longer needs to watch the entire nested property model which it was doing before. Instead this now takes advantage of angularjs ngModel $validators and now we simply shallow watch the current formController (FormController) object and if any property values change, we find all ngModel controls attached to this form and sub forms (but only up until the next umbProperty control) and attach a custom $validator. This means that when any of these controls are changed, it automatically clears this validator. The other piece of the puzzle is to subscribe to all server validation changes for a validation key 'prefix'. this means the subscription callback will fire for all child server validation changes which it can then use to clear itself if there is only one error (itself) remaining - this same check takes advantage of the fix for infinite editors and linking parent/child forms because it allows us to check the $pristine setting on the form. 
* new `valServerMatch` directive, this is used to put the border around the invalid block. This uses new functionality in the serverValidationManager to subscribe to either an exact match, prefix, suffix or 'contains' match of a validation key. If it matches it will set the attached FormController's validity state for a validator called `valServerMatch` to invalid. This can then be used in html to check the validity state of that validator to toggle html like css classes
* `contenteditinghelper.service.js` - updated to parse the complex json validation message to show each single string message instead of a giant json blob
* `serverValidationManager` - probably the biggest update (but there's some tests!). This service has been refactored quite a bit so sorry for it looking like it's entirely changed but I moved most of the functions to be normal functions instead of functions declared as properties. This is much more readable and maintainable and allows for much nicer javascript variable scoping and not needing to rely on `this` and assigning that to `self`. Not all functions have been moved though (you'll see). 
  * Added functionality to subscribe to validation changes by matching a validation key with exact match, suffix, prefix or contains matching
  * We now notify callbacks anytime changes are made
  * adds logic to parse the complex validator json blob coming back in ModelState from the server
* Adds functionality to the udiservice for building and parsing udis
* umbBlockListPropertyEditor directive 
  * removes propertyForm property. There is no reason to have this property as it's resolved through directive inheritance so we just add a require `propertyForm: "^form"` to do this correctly
  * resolves umbVariantContent via $scope chain if not found in the DOM inheritance (i.e. in infinite editing) since we always need a reference to it
  * When creating the block object we must always pass along the current culture/segment data to the block otherwise validation fails (since it relies on culture/segment) but I think other things would fail too
  * ensure the control does property $dirty setting when blocks are deleted and ensure the serverValidationManager is updated with removed keys when this happens
  * always pass along the $parentScope and $parentForm when launching infinite editing to maintain the inheritance
  * always pass along the block index - we actually no longer need this for validation but I think it's probably useful for something in the future
  * No longer render everything in this directive, instead there's a new sub directive: umbBlockListRow which allows us to wrap the row in an ng-form with the correct $scope.
* Updates nested content stuff to work with complex server validation. Not much was needed but there was a few things like updating serverValidationManager when things are deleted and passing along the correct key to the umbProperty.
* Removes `umbBlockListScopedBlock` because it's totally unnecessary, there's no reason to have 2x directives that do the exact same thing based on one property. Instead we have one directive `umbBlockListBlock` which renders each block, if it has a stylesheet we $compile with shadow dom, if it doesn't we $compile normally. Also have fixed this directive to watch primitive values like `index` which would never have worked to be passed along if it's not watched. Note: the `umbBlockListBlock` isn't new, it's just that I fixed the file naming convention so it looks like it's been re-created.
* Fixes file naming conventions. We don't name JS directive files with snake-case, only its views.

### Dotless

* Because umb-property has been changed to a real component, the element <umb-property> is output in the html. Because of that, some css had to be reconfigured
* Adds the red border to blocks in error for both the block list and nested content
*


src/Umbraco.Web/Editors/Binders/ContentModelBinderHelper.cs
```diff
@@ -15,9 +15,9 @@ namespace Umbraco.Web.Editors.Binders
     /// <summary>
     /// Helper methods to bind media/member models
     /// </summary>
-    internal class ContentModelBinderHelper
+    internal static class ContentModelBinderHelper
```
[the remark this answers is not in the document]
Shazwazza: @warrenbuckley do you understand statics or not? there's reasons they should exist and there's reasons they should not exist. You can see here that all methods in this class are static, before we were allocating new instances of this object (allocations should be avoided if possible). This class is also internal and not a public API, it holds no state, its exactly the type of thing that should be static. I've never said "never" use statics, they have their place but don't ever create one unless you mean to do it for a reason.

src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
```diff
@@ -89,41 +98,37 @@ public override object Configuration
...
+                            var convValue = valEditor.ConvertDbToString(prop.Value.PropertyType, prop.Value.Value, dataTypeService);
+
+                            // update the raw value since this is what will get serialized out
+                            row.RawPropertyValues[prop.Key] = convValue;
                         }
                         catch (InvalidOperationException)
                         {
                             // deal with weird situations by ignoring them (no comment)
-                            row.JsonRowValue[row.PropKey] = null;
+                            row.RawPropertyValues.Remove(prop.Key);
```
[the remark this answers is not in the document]
Shazwazza: Will add this to my re-work task

Shazwazza:
@warrenbuckley have created a re-work task to follow up on a few things. I 'think' the build and tests should pass now, will check back after dinner

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js
```diff
@@ -17,18 +17,18 @@
             controller: BlockListController,
             controllerAs: "vm",
             bindings: {
-                model: "=",
-                propertyForm: "="
+                model: "="
             },
             require: {
+                propertyForm: "^form",
```
nielslyngsoe, line 23: Happy that you know this is right for all property-editors, I guess I didn't have that big picture :-)

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js
```diff
@@ -176,6 +185,20 @@
...
+                block.content.language = vm.umbVariantContent.editor.content.language;
+                // currently we only ever deal with invariant content for blocks so there's only one
+                block.content.variants[0].tabs.forEach(tab => {
+                    tab.properties.forEach(prop => {
+                        prop.culture = vm.umbVariantContent.editor.content.language.culture;
+                    });
+                });
+            }
+
+            // TODO: Why is there a '/' prefixed? that means this will never work with virtual directories
```
nielslyngsoe, line 201: This might be wrong, maybe it has to do with my setup. If you think its wrong, then let's make a PR with it corrected and have someone testing that it works. It does work, but that might be luck?

nielslyngsoe: Btw. there is multiple instances of similar appending.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js
```diff
@@ -261,10 +298,13 @@
             }
             
             var blockEditorModel = {
+                $parentScope: $scope, // pass in a $parentScope, this maintains the scope inheritance in infinite editing
+                $parentForm: parentForm || vm.propertyForm, // pass in a $parentForm, this maintains the FormController hierarchy with the infinite editing view (if it contains a form)
                 hideContent: blockObject.hideContentInOverlay,
                 openSettings: openSettings === true,
                 liveEditing: liveEditing,
                 title: blockObject.label,
+                index: blockIndex,
```
nielslyngsoe, line 307: I´m not super happy about this 'index', I think that reflects that block editor has to be editing blocks of a list, but I don't think we should dictate that. :-)
If someone wants to have blocks in a _Dictionary_ or just single properties(like a 'left' and 'right' block) they should be able to do so. If I read and understood correctly we are not going to use this Index currently, and reading the _BlockEditorController_ its not begin used. So I think it would be good to not have this.

Currently, the _editBlock_ method is exposed thought vm.blockEditorApi, and is used by customViews to open the infinite editor, for those developers it's not so nice that they have to parse these many attributes, would be nice to keep it as simple as possible. So I would love to hear your thoughts on the options to simplify this as much as possible. One though could be that the _umbBlockListPropertyEditor_ appends a _edit_() method to the block-objects, which gathers the references, index, etc. required for this.

let me know your thoughts, thanks :-)

src/Umbraco.Web/PropertyEditors/BlockEditorPropertyEditor.cs
```diff
@@ -66,210 +64,146 @@ public IEnumerable<UmbracoEntityReference> GetReferences(object value)
...
-                        var regex = new Regex(row.PropType.ValidationRegExp);
-                        if (!regex.IsMatch(row.JsonRowValue[row.PropKey].ToString()))
-                        {
-                            var message = string.IsNullOrWhiteSpace(row.PropType.ValidationRegExpMessage)
-                                                      ? $"'{row.PropType.Name}' is invalid, it does not match the correct pattern"
-                                                      : row.PropType.ValidationRegExpMessage;
-                            validationResults.Add(new ValidationResult($"Item {(row.RowIndex + 1)}: {message}", new[] { row.PropKey }));
-                        }
+                        elementValidation.AddPropertyTypeValidation(
+                            new PropertyTypeValidationModel(prop.Value.PropertyType, prop.Value.Value));
```
nielslyngsoe, line 105: Yes yes yes, super nice. Love this refactoring. — I´m not reading all the C# but just the BlockEditor changes.

src/Umbraco.Web.UI.Client/src/common/directives/validation/valpropertymsg.directive.js
```diff
@@ -84,43 +83,109 @@ function valPropertyMsg(serverValidationManager, localizationService) {
...
+                    }
+                });
+            }
+
+            // We start the watch when there's server validation errors detected.
+            // We watch on the current form's properties and on first watch or if they are dynamically changed
+            // we find all ngModel controls recursively on this form (but stop recursing before we get to the next)
+            // umbProperty form). Then for each ngModelController we assign a new $validator. This $validator 
+            // will execute whenever the value is changed which allows us to check and reset the server validator
+            // TODO: Is this even needed?
```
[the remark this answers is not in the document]
Shazwazza: Ah yes i will remove, there's prob some TODOs left behind while working on the code that should be removed

src/Umbraco.Web.UI.Client/src/common/directives/validation/valservermatch.directive.js
```diff
@@ -0,0 +1,89 @@
...
+            
+            var unsubscribe = [];
+
+            //subscribe to the server validation changes
+            function serverValidationManagerCallback(isValid, propertyErrors, allErrors) {
+                if (!isValid) {
+                    formCtrl.$setValidity('valServerMatch', false);                    
+                }
+                else {
+                    formCtrl.$setValidity('valServerMatch', true);
```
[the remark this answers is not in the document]
Shazwazza: @nielslyngsoe yeah not sure why our .editorconfig isn't kicking in here.... oh, i just looked it's because this is the setting: 🤦‍♂️ 

```
[*.{js,less}]
trim_trailing_whitespace = false
```

I have no idea why this is set to false do you? Seems like it should absolutely be set to true?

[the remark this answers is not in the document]
Shazwazza: I'm going to push a change to keep the whitespace for .less but remove it for .js (Since I'm not sure what requirements there are for .less). Let me know if you want to change that.

src/Umbraco.Web.UI.Client/src/common/services/contenteditinghelper.service.js
```diff
@@ -614,10 +634,9 @@ function contentEditingHelper(fileManager, $q, $location, $routeParams, editorSt
                     formHelper.handleServerValidation(args.err.data.ModelState);
 
                     //add model state errors to notifications
+                    // TODO: Need to ignore complex messages
```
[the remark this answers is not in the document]
Shazwazza: yep good catch

src/Umbraco.Web.UI.Client/src/common/services/servervalidationmgr.service.js
```diff
@@ -68,82 +130,468 @@ function serverValidationManager($timeout) {
...
+
+            var matchProp = cb.propertyAlias === propertyAlias
+                ? true
+                : cb.options.matchType === "prefix"
+                    ? propertyAlias.startsWith(cb.propertyAlias + '/')
+                    : cb.options.matchType === "suffix"
+                        ? propertyAlias.endsWith('/' + cb.propertyAlias)
+                        : cb.options.matchType === "contains"
+                            ? propertyAlias.includes('/' + cb.propertyAlias + '/')
+                            : false;
```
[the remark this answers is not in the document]
Shazwazza: Yep I agree it's ugly, but it's actually not the same logic as the other one so it's not a duplication. I did try to split this out into if statements but it ends up being a super huge nested bit of code which i thought was much uglier than this. I'll give it another try.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js
Shazwazza: @nielslyngsoe It works currently for internal editors and will work for App_Plugin editors but we are 'rooting' the path. Websites aren't always hosted in the root path. Umbraco is possible to be hosted in a virtual folder like /cms/ instead of at / in which case this will never work because we are rooting this. This is why we have `umbRequestHelper.convertVirtualToAbsolutePath` and this code should use that instead. 

I need to move focus from 8.7 for now as we have other critical issues/support i need to work on. Are you able to create tasks for this?

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js
Shazwazza: @nielslyngsoe Actually the index probably no longer needs to exist and be passed everywhere. This was an initial requirement when thinking and prototyping some json path stuff with validation. I actually think it can be removed. The reason i kept it in there was because I figured we might end up using the index for other purposes and more information would be better but as you note that might not be required for other editor types. I know we use index for some other things but IIRC we no longer require or use it within the editBlock/editSettings stuff. Are you able to make tasks for this to be cleaned up?

src/Umbraco.Web.UI.Client/src/common/services/servervalidationmgr.service.js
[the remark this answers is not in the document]
Shazwazza: @nielslyngsoe I've pushed a change you can see in this rev: f0cc7a493d68eb8171b6f32b6f39556369074e1e, have moved to 2 different methods with comments and changed to simple if statements so it's easier to follow.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umb-block-list-property-editor.html
```diff
@@ -18,47 +18,10 @@
                     <div class="__plus" ng-style="{'left':inlineCreateButtonCtrl.plusPosX}">+</div>
                 </button>
 
-                <div class="umb-block-list__block" ng-class="{'--active':layout.$block.active}">
-
-                    <umb-block-list-scoped-block
-                        ng-if="layout.$block.config.stylesheet"
-                        class="umb-block-list__block--content blockelement__draggable-element"
```
nielslyngsoe, line 25: @Shazwazza this class "blockelement__draggable-element" did slip our review, I´m adding it in a new PR.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js
nielslyngsoe: Good, Ive made a task for it in AB.

PR #8344, opened 2020-06-26
Fix folder uploads when existing image names clash with the uploaded folder name

### LINKED ISSUE #8325: Media item acts as folder if a folder of the same name as the media item is being dragged into a folder
If you drag and drop a folder into another folder in the media section, and a media item in the same directory has the same name as the folder, this media item will be used as a folder (the content of the dragged folder is created as children of the media item).
This is an issue since the media picker don't treat the image as a folder, so the children can't be picked.

## Umbraco version
I am seeing this issue on Umbraco version: 8.6.2 on a blank install with starter kit.

Reproduction
------------
### Steps to reproduce

1. Go to the media section of Umbraco, and create a folder
2. In this folder upload an image called 'Image'
3. On your desktop, create a folder called 'Image' and add an image called 'Folder Image' to it
4. Drag and drop this folder into the media section of Umbraco
5. Notice that 'Folder Image' is now a child of 'Image'

### Expected result
When I upload a folder, I'd expect a folder in the media section to be created. If the issue is the dublicate name, then the folder should have the (1) suffix you normally get, when trying to create an item with the same name as another item.
Seems harder to reproduce in the root of the media section than inside a folder.

### Actual result
Content of the dragged folder is created as childs of the media item with the same name as the folder.
image

### LINKED ISSUE #8342: Member fields of type label are not shown in the backoffice
When a membertype contains fields of the type label, they are not shown in de backoffice.

afbeelding

afbeelding

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.2 and 8.6.3 (maybe earlier versions > 8.2.2)

Reproduction
------------

Create under de "Settings"-section a membertype and add a field of the type "Label". 
Go to the section "Members" and add an new member.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8325

### Description

See #8325 for steps to reproduce.

When this PR is applied, the uploaded folder is created correctly with the "(1)" suffix, and the uploaded images are created inside the folder as one would expect:

folder-upload-name-clash

### DIFF
--- a/src/Umbraco.Web/Editors/MediaController.cs
+++ b/src/Umbraco.Web/Editors/MediaController.cs
@@ -774,10 +774,13 @@ namespace Umbraco.Web.Editors
             var total = long.MaxValue;
             while (page * pageSize < total)
             {
-                var children = Services.MediaService.GetPagedChildren(mediaId, page, pageSize, out total,
+                var children = Services.MediaService.GetPagedChildren(mediaId, page++, pageSize, out total,
                     SqlContext.Query<IMedia>().Where(x => x.Name == nameToFind));
-                foreach (var c in children)
-                    return c; //return first one if any are found
+                var match = children.FirstOrDefault(c => c.ContentType.Alias == contentTypeAlias);
+                if (match != null)
+                {
+                    return match;
+                }
             }
             return null;
         }


nul800sebastiaan:
Wow that is a good find, thanks for the fix @kjac!! 👍

PR #8346, opened 2020-06-26
Show custom label properties on members

### LINKED ISSUE #8342: Member fields of type label are not shown in the backoffice

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8342

### Description

As described in #8342 ... custom label properties on a member type are not rendered when editing a member of said type.

For example this member type:

image

...renders like this in the editor (note the missing label property):

image

This happens because we don't want to show the system label properties (prefixed with "umbracoMember", e.g. "umbracoMemberFailedPasswordAttempts") when editing the member. So we filter out all label properties. Bit harsh, really.

This PR adds a custom filter which only filters out label properties if they're carrying the "umbracoMember" prefix. When the PR is applied, the custom label property renders in the editor as expected:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/member/apps/content/content.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/member/apps/content/content.controller.js
@@ -5,6 +5,14 @@
 
         var vm = this;
 
+        vm.hideSystemProperties = function (property) {
+            // hide some specific, known properties by alias
+            if (property.alias === "_umb_id" || property.alias === "_umb_doctype") {
+                return false;
+            }
+            // hide all label properties with the alias prefix "umbracoMember" (e.g. "umbracoMemberFailedPasswordAttempts")
+            return property.view !== "readonlyvalue" || property.alias.indexOf("umbracoMember") !== 0;
+        }
     }
 
     angular.module("umbraco").controller("Umbraco.Editors.Member.Apps.ContentController", MemberAppContentController);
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web.UI.Client/src/views/member/apps/content/content.controller.js
```diff
@@ -5,6 +5,14 @@
 
         var vm = this;
 
+        vm.hideSystemProperties = function (property) {
+            // hide some specific, known properties by alias
+            if (property.alias === "_umb_id" || property.alias === "_umb_doctype") {
+                return false;
+            }
+            // hide all label properties with the alias prefix "umbracoMember" (e.g. "umbracoMemberFailedPasswordAttempts")
+            return property.view !== "readonlyvalue" || property.alias.indexOf("umbracoMember") !== 0;
```
nathanwoulfe, line 14: Hi @KennethJakobsen, this is me being picky, but I'd use `!property.alias.startsWith('umbracoMember')` rather than `indexOf`, for those extra readability points. Like I said, me being picky and preferring the natural language function names 😄 Is there a JS style guide for the UI?

nul800sebastiaan: Even more picky, we like using `=== false` instead of the easily overlooked `!` in front of something 😅 

```suggestion
            return property.view !== "readonlyvalue" || property.alias.startsWith('umbracoMember') === false;
```

There is no JS style guide, yet ... ;-)

nul800sebastiaan:
Works like a charm! Thanks @kjac and @nathanwoulfe! 👍

PR #8348, opened 2020-06-28
Add the current culture to the cache key  for the CachedPartial Html Helper

marcemarc:
When using variants it's unlikely to need the same Html.Partial result to be cached across all language variations, if we always add the current culture name to the cache key, we ensure this doesn't accidentally happen, and when not using variants, no harm to have the current culture name for the single site included in the key.

We made a very similar change for Macro caching here:
https://github.com/umbraco/Umbraco-CMS/pull/7555/files

I'm not sure why at the time, I didn't think to check whether Html.CachePartial would have the same issue.


nul800sebastiaan:
Thanks very much @marcemarc, good little fix! 👍

PR #8354, opened 2020-06-29
Remove sorting in grid layout configuration

bjarnef:
### Prerequisites

### Description
It seems like the grid layout configuration is using `ui-sortable` and `model.value.templates`, but this is undefined and therefore causing a lot of console errors.

Not sure if this has worked previous. I tried in an older Umbraco install using v8.5.5 with same console errors.
I also added a few accessibility enhancements:
https://github.com/umbraco/Umbraco-CMS/pull/8354/files?diff=unified&w=1


bjarnef:
@nielslyngsoe do you recall if this has ever worked or the `ui-sortable` by mistake have been copied to this?

PR #8355, opened 2020-06-30
Use eventsService to tell everyone the current culture has changed

### LINKED ISSUE #8345: Nested content is not saved when you switch to another language and back
When we change nested content on a page in the primary language, then switch to another language (not split view) and back to english, the change will be gone.

That is not expected.

## Umbraco version
I am seeing this issue on Umbraco version: Umbraco 8.6.3

Reproduction
------------
In a multilanguage setup, edit nested content, change language and back and it is gone.
I expect that the content will be kept in browsermemory while switching through the different languages.

### Specifics
I am using Chrome on mac

nathanwoulfe:
### Prerequisites

### Description

To save having to manually track the current variant (`$watch` === 👿), this PR uses the events service to broadcast the update of the selected content variant. How that event is used is up the subscribing code, but will be useful in variant-aware content apps, and can be used when working on #8345 (by triggering Nested Content's sync method when the variant changes). Event is raised when the culture changes and when returning from split view. It does not fire on initial load, as the active variant is already available from editorState.

To test, unzip [CultureClub.zip](https://github.com/umbraco/Umbraco-CMS/files/4849928/CultureClub.zip) into App_plugins. It adds a quick content app, displaying the current culture. Switch cultures, app updates 🍾 

For those playing along at home and not wanting to install my beautiful app, this is the important bit:

```javascript
angular.module("umbraco")
    .controller("CultureClubController", function ($scope, eventsService, editorState) {
        // changing the culture in the header emits 'editors.content.cultureChanged'
        const cultureChanged = eventsService.on('editors.content.cultureChanged', 
            (_, args) => this.culture = args);
    
        this.culture = editorState.current.variants.find(x => x.active).language.culture;
    
        // MUST be disposed otherwise ever culture change adds a new listener which is horrible if you're responding to the 
        // change by sending a request - suddenly you have lots of requests instead of 1.
        $scope.$on('$destroy', () => cultureChanged());
});
```

Used `editors.content.cultureChanged` to be consistent with existing events `editors.content.splitViewChanged` and `editors.content.splitViewRequest`

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontenteditors.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontenteditors.directive.js
@@ -20,7 +20,7 @@
         controller: umbVariantContentEditorsController
     };
 
-    function umbVariantContentEditorsController($scope, $location, contentEditingHelper) {
+    function umbVariantContentEditorsController($scope, $location, eventsService) {
 
         var prevContentDateUpdated = null;
 
@@ -36,6 +36,7 @@
         vm.selectVariant = selectVariant;
         vm.selectApp = selectApp;
         vm.selectAppAnchor = selectAppAnchor;
+        vm.requestSplitView = requestSplitView;
 
         //Used to track how many content views there are (for split view there will be 2, it could support more in theory)
         vm.editors = [];
@@ -66,7 +67,7 @@
 
         /** Allows us to deep watch whatever we want - executes on every digest cycle */
         function doCheck() {
-            if (!angular.equals(vm.content.updateDate, prevContentDateUpdated)) {
+            if (!Utilities.equals(vm.content.updateDate, prevContentDateUpdated)) {
                 setActiveVariant();
                 prevContentDateUpdated = Utilities.copy(vm.content.updateDate);
             }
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks very much @nathanwoulfe - happy with this one! 👍

src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontenteditors.directive.js
```diff
@@ -101,13 +103,14 @@
...
-                        return (!v.language || v.language.culture === vm.editors[s].content.language.culture) && v.segment === vm.editors[s].content.segment;
-                    });
+                    var variant = vm.content.variants.find(v =>
+                        (!v.language || v.language.culture === vm.editors[s].content.language.culture) && v.segment === vm.editors[s].content.segment);
+
                     vm.editors[s].content = variant;
                 }
             }
 
+            eventsService.emit('editors.content.cultureChanged', activeVariant.language.culture);
```
bjarnef, line 113: @nul800sebastiaan @nathanwoulfe with latest changes in `v8/contrib` I get an console error on e.g. HomePage node, because the document type doesn't vary by culture and `language` therefore is null.

image

image

nathanwoulfe: Of course. Maybe just using active variant.language instead - that way the entire language object is emitted, or it's null.

nathanwoulfe: Always wondered too why invariant content didn't populate the default language...

bjarnef: Yes, it might be better with the entire language object instead and could be useful if the language object later is extended with more properties.

nathanwoulfe: #8474 fixes this

PR #8357, opened 2020-06-30
Grid config allow editors toggle

bjarnef:
### Prerequisites

### Description
This PR add a toggle to "Allow all editors" in row configuration similar to "Allow all row configurations" in layout configuration changed in https://github.com/umbraco/Umbraco-CMS/pull/6859

chrome_2020-06-30_11-41-54

chrome_2020-06-30_11-42-47

However the toggle changed in "Allow all row configurations" just changed the toggle state using `currentSection.allowAll` variable, but didn't use the existing function in the controller https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/layoutconfig.controller.js#L58-L65 but it didn't reset previous specific selected row configuration as before changes in https://github.com/umbraco/Umbraco-CMS/pull/6859

To test this
- Edit layout, select specific row configurations and submit
- Edit layout, toggle "Allow all row configurations" so it is enabled and submit
- Edit layout, toggle "Allow all row configurations" so it is disabled. Now the previous specific selected row configurations are still still and not cleared as previous. This PR also fix this issue.

This "Allow all editors" toggle looks like the following.

image

image


bjarnef:
@nielslyngsoe it would be great to use a toggle for "Allow all editors" as well. It also fixes an issue introduced in https://github.com/umbraco/Umbraco-CMS/pull/6859 where "Allow all row configurations" didn't reset previous specific selected row configurations as described. It would also be great to use `umb-checkbox` component instead of the traditional checkboxes for row configurations and grid editors, but we can have a look at this in another PR.

Much of this could use an overhaul to match the rest of the UI including a better class names using `uSky-` prefix.

bjarnef:
Hi @nul800sebastiaan

This PR should only be related to the toggle "Allow all editors". The change of the checkboxes come from this PR https://github.com/umbraco/Umbraco-CMS/pull/8363

I can take a look. Not sure if some other changes had an effect on this.

bjarnef:
@nul800sebastiaan it looks like this to me, but will have a look with latest changes from `v8/contrib` branch.
 
image

bjarnef:
@nul800sebastiaan okay, I can reproduce this with latest changes merged from `v8/contrib` and re-compiling views, styling...

it seems to be related to `.umb-form-check` styling.

image

image

bjarnef:
It seems some changes was made in this PR https://github.com/umbraco/Umbraco-CMS/pull/8253
which changed `label.umb-form-check` not to use `display: inline-block` but `display: flex` from `.umb-form-check`.

bjarnef:
@nul800sebastiaan since the `umb-checkbox` and `umb-radiobutton` use `ng-transclude` I have changed this in https://github.com/umbraco/Umbraco-CMS/pull/8357/commits/bcf46878fd36791fa475df09debef822b2e6c666 which add this content in the label for the checkbox.

It now looks like this:

image

Off topic: It would be great when new features are added to either `umb-checkbox` or `umb-radiobutton` that both of these are as consistent as possible, but we can look at that in another PR.

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js#L24-L35

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js#L24-L32

bjarnef:
I have submitted another PR regarding the differences between `umb-checkbox` and `umb-radiobutton` components.
https://github.com/umbraco/Umbraco-CMS/pull/8441

nul800sebastiaan:
Super updates @bjarnef! I agree in a perfect world all the updates would be made to both check/radio at the same time but nothing is super perfect at the moment so we have to look out for that. I'll be on the lookout for changes to either one and make sure they get applied to the other as well. 

In any case, this now works wonderfully and am happy to merge, thanks again!! 👍

bjarnef:
@nul800sebastiaan thanks champ! 😁😎👍🎉

PR #8363, opened 2020-06-30
Grid config row and editor checkboxes

bjarnef:
### Prerequisites

### Description
This PR replace the traditional checkboxes with `umb-checkbox` component in grid configuration when selecting row configuration for layout and grid editors for cells.

A minor issue with "Allow all row configurations" toggle which doesn't clear previous selected specific row configurations for the layout is mentioned and fixed in https://github.com/umbraco/Umbraco-CMS/pull/8357


nul800sebastiaan:
Happy with this one! 👍

PR #8204, opened 2020-05-30
Shorter JSON property names for nucache

nzdev:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes N/A

### Description

What:
Changes the property names in json for properties on ContentNestedData, CultureVariation and PropertyData types to shorter names.
Also  changes Culture and Segment on PropertyData to not be serialized to json when the value is an empty string, these properties will still deserialize to an empty string.

Why?:
When you have a large number of content/media items with several properties on them, cold boot up time (no NuCache.*.db files on disk) can be very slow. By shortening the json you reduce the number of bytes that need to be read from the data base into memory and then to the nucache.db files on disk. For ~26,000 published content and ~110,000 media items this saved ~2 mins of time for cold boots and reduced NuCache.Content.db from 114mb to 98mb for my scenario.

How to test?
Create a large number of media and published content items with 10+ properties. 

1. Delete the local nucache files (Nucache.*.db and Nucache.*.tlog).
2. Start the website, time how long it takes to boot
3. Website should load existing nucache serialization format successfully.
4. Rebuild the cmsContentNu table using the backoffice. The cmsContentNu.data column should be in the more compact representation
5. Stop the site.
6.  Delete the local nucache files (Nucache.*.db and Nucache.*.tlog).
7. Start the website, time how long it takes to boot
8. Website should load existing nucache serialization format successfully.

Issue #7968, opened 2020-04-16
umbracoNode index refactor


Shazwazza:
We currently have a bunch of individual indexes on umbracoNode:

* IX_umbracoNode_ObjectType
* IX_umbracoNode_ParentId
* IX_umbracoNode_Path
* IX_umbracoNode_Trashed
* IX_umbracoNode_UniqueId

and we do query on all of these and sort on a few but SQL Server doesn't use ALL indexes when it does a query, it's only going to use one of them and in many cases even when querying on these values it's not going to use any and instead choose to do a Clustered Index Scan on PK_umbracoNode. 

In some SQL tests I've run recently with the Execution Plan window open it tells me things like: 

```
Missing Index Details from SQLQuery3.sql 
The Query Processor estimates that implementing the following index could improve the query cost by 11.5743%.
```

```sql
CREATE NONCLUSTERED INDEX [<Name of Missing Index, sysname,>]
ON [dbo].[umbracoNode] ([nodeObjectType],[trashed])
INCLUDE ([uniqueId],[parentId],[level],[path],[sortOrder],[nodeUser],[createDate])
```

Depending on the query this report is saying that if that index existed it would use it. However we can't just go create every index that the execution plan suggests and we currently don't support having `INCLUDE` columns in indexes (one day we will).

But if we instead look at the current indexes we have and then look at the majority of queries we execute against umbracoNode (LOTS) we can see which columns should be in a single index. It just might be that objectType, trashed, path, uniqueId, parentId should all just be in one index.... this I'm just not sure of. 

I know a bit about this stuff but I'm no expert so hoping that someone out there might be able to add some insight to this and possibly help out. 

Thanks!

umbrabot:
Hiya @Shazwazza,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

Shazwazza:
@umbrabot still relevant

[...]

Issue #7570, opened 2020-02-04
umbracoPropertyData index refactor


Shazwazza:
Currently we have 4x individual indexes on umbracoPropertyData: languageId, propertyTypeId, segment, versionId

and although we probably query and join on all of these columns based on different service methods, there's a good chance that having 4x different indexes is not helping so much and pay actually be causing worse performance than having a single index.

An example of a single index is here which is a new temporary index created while doing migrations since it makes querying property data faster: https://github.com/umbraco/Umbraco-CMS/pull/7562/files#diff-96f37baf15cfa95daa5d26d5be5af8a5R100

I think this is probably more like what we require but we need to do a bit of research to figure out how we are querying, joining and sorting on umbracoPropertyData.

Slightly related is this issue https://github.com/umbraco/Umbraco-CMS/issues/6601

__UPDATE:__ actually it turns out we already have this index on this table. So we have a unique non-clustered index across all of these columns AND we have individual non-unique non-clustered indexes on each column. That doesn't seem right, there's no way that all of these indexes will actually get used. I'm just not 100% positive that only keeping the unique non-clustered index is all we need ... needs some research

stevemegson:
As far as I can see, only the unique index should be needed most of the time - we're almost always just querying the properties for known version IDs. The propertyTypeId index is used when deleting a property or a content type, and it may be worth keeping for that reason. I may be missing other uses, but I can't see why anything would use the other two indexes.

The odd case is looking up property values to sort a list view. I'd expect it to use the unique index to find just the rows for the required versions and property type(s), but I'm seeing a seek on the propertyTypeId index before joining to the set of relevant node IDs. With some manual editing of the SQL to simplify the joins, I can get it to use the unique index and my test cases are 4-5 times quicker that way. I get the same results for my tests, but I'd need to dig a bit deeper to convince myself that there's no weird edge case that requires the current query.

Since the primary key is hardly ever used (outside of migrations), I wonder if we'd be better with a nonclustered primary key and making the unique index clustered?

Shazwazza:
Would love to know a bit more on this one, i feel like some of the index changes we want in the CMS could see big performance improvements in some areas. I agree about the PK since it's not really there for any particular reason.

Shazwazza:
FYI i created another index refactor one here https://github.com/umbraco/Umbraco-CMS/issues/7968 for umbracoNode

umbrabot:
Hiya @Shazwazza,

[...]

PR #8365, opened 2020-07-01
Introduce IContentNestedDataSerializer to allow injecting a custom serializer for nucache

nzdev:
### Prerequisites

### Description
What:
Introduces IContentNestedDataSerializer interface to allow developers to replace the current serialization logic for nucache by replacing the NucacheComposer registration. Existing logic moved to JsonContentNestedDataSerializer class that implements IContentNestedDataSerializer and is registered by default.

Why?:
Once you have a large number of content/media items the size of the nucache json makes an significant impact on cold boot times and the size of the nucache db files. Any optimisations can add up by the time you have a few thousand content/media items. By allowing replacement of the serialization logic for nucache it can be possible to reduce the size of the json using strategies such as:
- Map the names of document properties to shorter alias's and back. E.x. "umbracoInternalRedirectId" becomes "12". For your own custom properties you could simply use shorter alias's, but for umbraco conventions it's not possible without some mapping like this.
- Don't serialize to json empty/ default property values in the PropertyData dictionary (On deserialization they are expected to exist, so you will need to know per installation what properties you need to deserialize and deserialize a default value to not break nucache)
- Use compression. Html in the rich text editor compresses very well using LZ4 and is fast https://github.com/MiloszKrajewski/K4os.Compression.LZ4
- Remove large property values that are not used in queries for older/ less frequently used content. This data could be looked up from the cmsContentNu table in the database or redis/blob storage etc and updated on publish/unpublish. This could save on memory use if the content is large enough.
- Use string interning on custom property names to save on memory.

How to test?

1. Existing nucache tests should pass.
2. Existing nucache content should deserialize successfully.
3. An alternative IContentNestedDataSerializer can be registered in a IUserComposer, replacing the default JsonContentNestedDataSerializer


nzdev:
The build error does not seem related to the code change.

Shazwazza:
> Use string interning on custom property names to save on memory.

YES! we've been planning on exploring this since before v8 was launched but haven't had the time, it would be interesting to see what results this could yield. Some folks are skeptical but I think it could be significant.

I'm very interesting in all of the things you've mentioned too and wondering if you have been running any benchmarking tools to find some of these bottlenecks? And if so if you could share any of that info? Mostly about this comment

> the size of the nucache json makes an significant impact on cold boot times and the size of the nucache db files.

I'm curious where the bottlenecks are for these things and how the size of the actual nucache db files affects the performance.

I think it would be wonderful if we could include many of these ideas in the core directly too. One thought is that maybe we can maintain a dictionary of actual aliases to shortened aliases directly within nucache. This can be used to serialize/deserialize the objects which would greatly reduce the size. 

Depending on where the bottlenecks are, maybe it would make sense to compress any string/html/json value if it exceeds a certain length. 

Regarding

> Remove large property values that are not used in queries for older/ less frequently used content. This data could be looked up from the cmsContentNu table in the database or redis/blob storage etc and updated on publish/unpublish. This could save on memory use if the content is large enough.

I mentioned in my comment here https://github.com/umbraco/Umbraco-CMS/issues/8367#issuecomment-652761645 about the umbraco routing behavior. I've really wanted to be able to properly be able to lazy load documents from the content cache for a long time but routing within Umbraco would need an overhaul to do that so I'm not sure if even with this PR you would really be able to do this since if any of the URLs inbound or outbound for 'old' content needed to be created it would automatically load in this data anyways.

Let me know what you think. I would love to start tinkering around with some of this with benchmarks to see what could be done.

nzdev:
@Shazwazza  Thanks for responding.

Regarding 

> the size of the nucache json makes an significant impact on cold boot times and the size of the nucache db files.

I had a look at the existing tests but didn't have a go at writing a proper benchmark. I have some really really rough notes from running with shorter json property keys for the fixed keys like "propertyData" => "pd" "segment" => "s" as well as simply mapping the custom properties from alias to id.

Fogive me for these really rough notes from when I was looking at this over a month ago, if any of this makes no sense let me know and I will elaborate:

Archived: I added a "archived" bit column to cmsContentNucache and excluded those rows in a join so no custom properties from the json for these in the DatabaseSource class, idea being there could be a built in "IsArchived" property so you know to look up custom properties from the db, leaving enough built in data for routing / filtering.

Times are mins:seconds, all times are running locally with the vs debugger attached. Estimates given for how this may perform on azure based on my testing for time for the first page to load (homepage)

Local times on i5-8265U 16GB Ram
On Azure at 50DTUs P1V2 Appservice plan ~ 3.14 times slower than local for first request.
On Azure at 100DTUs P1V2 Appservice plan ~ 2.3 times slower than local for first request.

Starting state:
6:22 mins locally, timed at around 19 mins on azure @50DTU, 15 mins on azure @100DTU
content documents: 114,688
media documents: 106,496
Content nucache db: 115mb
Media nucache db: 107mb

Only shrink built in json keys (changed "propertyData" to "pd" etc in previous pr)
Boot: 3:50 Estimated to be  x 3.14  12.0 mins on azure@50DTU, x2.3 = 8.8mins on azure@100DTU
Content: 98mb
Media: 106mb

shrink built in json keys and custom properties (map "umbracoRedirectUrl" to it's id)
Boot: 3:20 Estimated to be  x 3.14  10.4 mins on azure@50DTU, x2.3 = 7.6mins on azure@100DTU
Content: 81mb
Media: 106mb


1/3 media archived (not loaded in join on nucache table)
Boot: 3:52
Content:115mb nucache.db file
Media: 70mb nucache.db file

1/3 media archived, no content archived, load from sqldb
Boot: 3:55 Estimated to be  x 3.14  12.5 mins on azure@50DTU, x2.3 = 9.2mins on azure@100DTU
Content:115mb
Media: 70mb

1/3rd media archived and content leaves older than 1 year archived, load from sqldb
Boot: 2:24 Estimated to be x 3.14 7.5 mins on azure@50DTU, x2.3 = 5.5mins on azure@100DTU
Content: 70mb
Media: 70mb

70% media archived and content leaves older than 1 year archived, load from sqldb
Boot: 1:30 Estimated to be x 3.14 4.7 mins on azure@50DTU, x2.3 = 3.5mins on azure@100DTU
Content: 70mb
Media: 32mb

70% media archived and no content archived, load from sqldb
Boot: 1:44 Estimated to be x 5.5 mins on azure@50DTU, x2.3 = 4.0mins on azure@100DTU
Content: 98mb
Media: 32mb

No archiving, boot from disk nucache.db files
boot 1:32

1/3rd media archived and content leaves older than 1 year archived, boot from disk nucache.db files
boot 1:12

Shazwazza:
@nzdev great stuff :) So the tests in this case showed a direct correlation between the amount of documents in the nucache files and boot times. I think this is expected since it needs to read in more document structures if there are more. I'm certainly intrigued about the possibility of having this data read faster with smaller keys and smaller data but with the same document counts. I might try to put some benchmark's together to see if that's a real possibility as i think that would be a reasonably 'quick' win. Will reply if i get some time today/tomorrow to do that.

Shazwazza:
aha, i just remembered your previous PR about the same sort of thing https://github.com/umbraco/Umbraco-CMS/pull/8204#issue-425443767 I think this is they key part: "By shortening the json you reduce the number of bytes that need to be read from the data base into memory and then to the nucache.db files on disk" That makes sense. I for some reason had it in my head that we were talking about the performance from nucache.db -> in memory but we are mostly talking about "cold boot" like you said from db -> in memory. But bytes are bytes so will be interesting to see different benchmarks, even just raw json deserialization benchmarks with smaller json for the same doc. I'll report back findings when i have them.

nzdev:
For 

> Use string interning on custom property names to save on memory

It might make sense to store the short property names in bplustree then map to the interned string rather than in an IContentNestedDataSerializer. I'm currently tempted to just use 1, 2, 3... as property names and use constants in code everywhere. It might make a difference for linq queries. On the todo list to try out. Here's where you would intern.
image


That LZ4 library seems to break umbraco's build, but if that was fixed, it appears to be a worthwhile tradeoff if it was opt in.

nzdev:
The charts for messagepack performance seem impressive though

nzdev:
@Shazwazza sorry, yes this is about cold boots from db, new app service / new code deployments.

nzdev:
With the media archiving testing, the thought was that we have a lot of unpublished content we need to keep, and that's not loaded into nucache, but the now unused media it was using still is being loaded into nucache, slowing down boot times significantly when we don't need it, but don't want to delete. For the testing results in the previous comment only ~26000 content documents are published.

Having a way to create a media archive policy so we don't load archived media would make a big difference. With the new media tracking feature you could set a policy like archive/don't load into nucache at all if not tracked as in use, and not in a galleries folder (to avoid breaking custom queries). Then on publish/ unpublish of content, archive and unarchive media. It would have to be opt in though with a custom policy as then it's on the developer if they don't load media that they need.

Shazwazza:
Ah yes media is actually a great candidate for true lazy loading 'today' because it doesn't have a nested URL structure that needs to be calculated. That is quite interesting actually because if we could make media truly lazy loaded that would save quite a lot of memory for large sites if media is 'archived', but if we did it for real, it wouldn't matter if it was flagged as archived or not, it just wouldn't be loaded unless requested either from nucache.db or from the db.

I'll defo have a look at the LZ4 and message pack stuff.

Just running some memory and cpu benchmarks on the cold boot/rebuild cache process to see what I can.

nzdev:
One other thing to note was that I disabled all the examine index's for those test runs. If these ideas get implemented it may make sense to remove the internal index for frontend servers when load balancing to further save on boot times as you shouldn't really need to be searching the back-office on these instances and media is no longer cached using the internal index. Thanks for helping out with these ideas, I was hoping to get the interface in at least before v8.7.0 as I haven't had much of a chance to implement anything other than property mapping.

Shazwazza:
@nzdev Some quick tests with string intern and using dotmemory to profile this stuff, for property aliases we save about 2MB of memory based on interning those aliases at that location and I have roughly 10k content + media items combined in these tests. There's other interning that should be done in other places to and these are super quick wins. We should also intern all aliases throughout the codebase but will focus on the cache side of things for now. I suppose in a cold boot scenario we'd want to try to have more control over the `DatabaseDataSource.DeserializeNestedData` to ensure that property aliases are interned when deserialized. I'm just testing out a new serialization converter/attribute to intern some of the things in `Umbraco.Web.PublishedCache.NuCache.DataSource.PropertyData` and their associated dictionary keys upon deserialization.

As for the compression side of things - there's a few different areas where compression can work so just wanted to get your thoughts on what you were thinking:

a) We can compress the entire json block data in the contentNu table - this will improve the IO time of cold booting when loading data directly from the DB
b) We can compress individual property data at the property level. This would mean less data stored in the umbracoPropertyData table along with less data in memory when working with IContentBase instances that are cached. This compressed property data would also make it's way into the contentNu table and like the previous point would improve the IO read time from the DB. This compressed data would also be what is cached within IPublishedContent therefor saving quite a lot of memory. We would then decompress it when sending this data to an editor or the front-end (maybe property value converter). We could go even further and keep our own 'local intern pool' to be sure we aren't re-creating large string instances for property values.

Option a) is super easy, option b) is more involved but could be huge in potential memory saving if it would work (off the top of my head seems like it could)

> If these ideas get implemented it may make sense to remove the internal index for frontend servers when load balancing to further save on boot times as you shouldn't really need to be searching the back-office on these instances and media is no longer cached using the internal index.

It is basically possible to do this today but you'd have to jump through a few hoops to do it, we can certainly make it much much easier to just disable an entire index (basically making it a noop index along with it's populators, etc...). The nice thing about v8 though compared to v7 is that there is no longer double lookups to populate the indexes, the lookups are shared which is the premise of populators. So if you have the ExternalIndex still, you'll get the same data lookups you'd get anyways if you had both the Internal and External because they are both populated with the same 'populator' so maybe it won't have much impact to disable it. Easily done though.

nzdev:
Option b has one potential issue, if you are writing SQL and working with the property data and it's compressed, that might be a deal breaker unless SQL server has lz4 using sqlclr. But the memory savings would be worth it at my scale $.

nzdev:
Equally, compressing at the property data level could  save on SQL storage, especially if there are multiple versions

Shazwazza:
Yep though luckily there's nothing in the core that searches the DB on property data so that would only be if someone is working directly with SQL. If it were configurable it could be quite good!

Another feature we've been talking about for a very very very long time which is actually quite easy to do is to automatically limit the versions in the DB. I know there's an UnVersion package that sort of does this but this should be baked into the core with a few options like: Max versions to keep, Min versions to keep, Max age to keep which could be set per doc type. But anyways, that's another topic :P

nzdev:
I think it could be configurable to have compression off/ nucache table/ property data. Even better if you could switch between. 

Another topic im interested in, versions/SQL.😀
As for unversion/ old versions. On v7 I had a look into simply moving old versions to another set of tables, I assumed the old versions were used only during rollback, so that would be nicer if you have to maintain all versions for legal reasons. Another modern approach is using temporal tables https://docs.microsoft.com/en-us/sql/relational-databases/tables/temporal-tables?view=sql-server-ver15. If you have a ton of versions and content the scheduled publishing query is ineffecient. Using filtered SQL indexes like only including the newest/published documents had benefits for v7. Using include for some sql indexes also make a worthwhile difference.

nzdev:
Option c would be to compress at the property level in the nucache table.

nzdev:
Interning/ string pool for the property values would also work. It might be worth just interning "0", "1", "" and not any other string.

nzdev:
I've got a ton of ideas, if these should all be separate topics, great, but while we are talking data and performance. The audit log table could also use int codes for each event rather than text and use format strings and only store the variables. Then it's possible to reconstruct the messages in c# and still allow for custom messages.

Shazwazza:
Love all the ideas :) we'll eventually need to move these to separate topics but for me it's fine for now just chatting through this stuff here. It's just lucky that today and tomorrow are our 'end of cycle' hack days so I actually have time to do some of this stuff and have been wanting to for ages... so very good timing :) I've pushed a branch with some WIP changes here https://github.com/umbraco/Umbraco-CMS/pull/8376

I like all options a, b and c and i think we can make them all configurable, that's my next attempt but i want to start with a) just for simplicity to see how either message pack or the LZ4 lib behaves but will hopefully prototype the others tomorrow. 

> Another modern approach is using temporal tables

I didn't know about this, would be very interesting! But the challenge is to stop support for SqlCe and/or other databases entirely in order to do that. Perhaps one day since I really want to look into leveraging graph querying in SQL server since that seems like an interesting fit https://medium.com/@davidglass_49320/graph-processing-in-sql-server-2017-8291a3a5122a . Even before going into uncharted SQL server territory I really want to overhaul the DB and use JSON documents for most things. For history we don't need to keep all property versions in the property versions table, just the edited/published. All other document versions could just be a single row of a serialized json document. Rolling back would mean just deserializing and replaying it so long as the schema hasn't changed (but that's the same problem today). 

It sound like you understand more about SQL server than I do and we've been wanting to make sure we have the correct indexes since I know we don't. There's a couple open issues to track some of this but the research has fallen flat since I don't think many people truly understand the implications of SQL indexes: https://github.com/umbraco/Umbraco-CMS/issues/6601, https://github.com/umbraco/Umbraco-CMS/issues/7570, https://github.com/umbraco/Umbraco-CMS/issues/7968 

> Interning/ string pool for the property values would also work. It might be worth just interning "0", "1", "" and not any other string.

I'm unsure exactly what you mean here?

nzdev:
I wouldn't really want to drop support for all other databases. PostgreSQL would potentially be a lower cost/better alternative and is supported as a managed database on azure. Once a .net core version is available, a managed PostgreSQL database + Linux app service plans would be a huge cost saving for umbraco hosting of large / multiple sites.

nzdev:
> Interning/ string pool for the property values would also work. It might be worth just interning "0", "1", "" and not any other string.

> I'm unsure exactly what you mean here?

It's not be appropriate to intern all string custom property values, but interning a few common values might be worth doing. Examples of values potentially worth interning:
"0", "1", "true", "false", empty string (might not be necessary).
I imagine these would appear frequently in peoples data.

Shazwazza:
I've merged your changes into my WIP branch (which is why it says merged :) 

I've made some fun progress today and it's all in here: https://github.com/umbraco/Umbraco-CMS/pull/8376

* Added MessagePack (fixed the build issue)
* Created a message pack serializer and added a test to make sure the same data is serialized/deserialized between that and the default one
* Made it configurable as an app setting: `<add key="Umbraco.Web.PublishedCache.NuCache.Serializer" value="MsgPack"/>`

If you want to give it a spin I'd be interested to see any benchmark figures on your end with your data set size. The test data set size I'm currently using isn't huge, only about 7000k docs but the rich text size isn't that big either. If I run `sp_spaceused cmsContentNu` in SQL mgmt studio it will give me the data size of this table. With Message Pack + LZ4 compression on it reduces this table size by a couple of MB which of course means saving that amount on data transfer on cold boot. But for a much larger data set this could be quite significant. 

I've added a ton of TODO notes in that PR and want to start exploring option c as it's the next easiest option. 

> It's not appropriate to intern all string custom property values

aha i missed the part about property 'values' i was still thinking about the aliases. And yep that's probably true! Just unsure how to check a string value before you intern it since that would just allocate the string?

nzdev:
The string would already be allocated, when you intern the string, it updates the reference to the string in the interned set of string. The copy would then be available for garbage collection. A switch statement might suffice?

nzdev:
I'll give it a go. Is there an easy way to generate a nuget package with a beta version number? ~397mb for my cmsContentNu currently

Shazwazza:
Yep, once i fix the build that's breaking currently it'll produce a nuget package, i'll put the link here when thats done

nzdev:
Rebuilt with the json serializer, Data 560616 KB for sp_spaceused cmsContentNu, 3:40 cold boot in vs. I assume it timed out the last time I rebuilt the cmsContentNu table.

Rebuilt with the messagepack serializer 440576 KB  for sp_spaceused cmsContentNu ~3:40 cold boot as well. It will likely have more of an impact when running on Azure given the database won't be on the same server. The big gain is still in lazy loading media, but the next one would be to map the alias' to shorter ones

Shazwazza:
Well 100MB of IO is pretty good :) but yep there's plenty more to do. I'm just adjusting the serialization/deserialization and the cmsContentNu table to support a 'raw' varbinary(max) column to store the serialized bytes instead of base64 string since I'd like to see what diff that makes too.

nzdev:
It should make some difference as there is less allocations due to having to use base64.

My previous attempt at mapping the properties https://github.com/umbraco/Umbraco-CMS/pull/8379

Shazwazza:
Wow, have just pushed that change with a migration to add the binary column. The MsgPack serializer now stores it's data in raw bytes. My table size went from 25000KB to 7448KB ! If you can replicate that kind of change on your end i think that will make quite a big difference 🤞

Shazwazza:
The timing in the logs with this now when PublishedSnapshotService does "Loading content from database" is consistently about 500ms faster than before so was normally around 2000ms with my data set and is now about 1500ms. I think that's reasonably significant. Also considering this is is all just 'option a'

nzdev:
222976 KB down from 440576 KB

nzdev:
I'm still getting sql lock timeouts though when rebuilding the database cache in the backoffice

Shazwazza:
> I'm still getting sql lock timeouts though when rebuilding the database cache in the backoffice, seems to recover though

I changed the row count for the paging to be smaller and configurable since it was trying to page over too much . The culprit is this: https://github.com/umbraco/Umbraco-CMS/pull/8376/files#diff-83e79f7101cd6797ca42ad8ed07f5835R633

Would love it if we could figure out a nicer way to get the property data without doing huge "IN" clauses which results in an index scan on the property data table - which if you have a lot of versions would get even worse. 

> I had to rebuild the table as the data column was not nullable.

 good catch, i'll update the migration

nzdev:
@Shazwazza do you have an example of that query? I'm having trouble tracking it down with the sql profiler

Shazwazza:
damn, i actually just closed the sql profiler with it, it's a simple query that just selects rows from umbracoPropertyData but it does a WHERE versionID IN (.......) and will include up to 2000 ids (since 2100 params is max). Id searches would be fast but because we pass in so many it just does an index scan which means iterating every single row in the table, at least according to the query execution plan on my machine.

nzdev:
Not really sure without further context. Reducing the number of rows in umbracoPropertyData by only keeping the newest and published in this table and shifting the rest to a umbracoPropertyDataHistory table (ideally using a schema that could be compatible with temporal tables in futures) would help.
Looks like the test issue is sorted, where would I grab the nuget packages to try out?

Shazwazza:
Yep that can certainly be one part of the solution.

I think you should have access to the build itself since it's a public pipeline: https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=43806&view=results and you can grab the artifacts from there. Else i think the pipeline package feed would work too which is https://pkgs.dev.azure.com/umbraco/0fe81d3e-adbd-40f4-af17-796887948029/_packaging/UmbracoCms/nuget/v3/index.json 

the version built in that pipeline is `UmbracoCms.8.7.0-alpha.9764`

On that note, unfortunately these changes won't make it into 8.7.0 and will need to wait until 8.8.0 or later. 8.7.0 is locked as of a week ago.

nzdev:
Some of these transient errors are not retried:
https://docs.microsoft.com/en-us/azure/azure-sql/database/troubleshoot-common-errors-issues#transient-fault-error-messages-40197-40613-and-others

nzdev:
@Shazwazza fair enough. MessagePack should also be added as a dependency in nuspec

nzdev:
When transitioning between json and message pack, the site won't boot as the dataRaw column is null and the data column contains json. It might be appropriate for the messagepack serializer to fallback to the json serializer during deserialization if dataRaw is null. Serialization should still be messagepack in this scenario so new/rebuilt cmsContentNu records are in messagepack.

nzdev:
@Shazwazza WIP for option c https://github.com/umbraco/Umbraco-CMS/pull/8400

nzdev:
Further looking into boot up, DatabaseDataSource.ContentSourcesSelect could be optimised. Line 71
`  .LeftJoin<ContentNuDto>("nuEdit").On<NodeDto, ContentNuDto>((left, right) => left.NodeId == right.NodeId && !right.Published, aliasRight: "nuEdit")`

which results in:
`LEFT JOIN [cmsContentNu] [nuEdit]
ON (([umbracoNode].[id] = [nuEdit].[nodeId]) AND NOT ([nuEdit].[published] = 1))`

Where it should be:
``LEFT JOIN [cmsContentNu] [nuEdit]
ON (([umbracoNode].[id] = [nuEdit].[nodeId]) AND ([nuEdit].[published] = 0))``

cmsContentNu.Published is a not null bit column, so it's only 0 or 1. Usually doing NOT where clauses results in indexes not being used.

Same with:
`WHERE ((([umbracoNode].[nodeObjectType] = @4) AND NOT ([umbracoNode].[trashed] = 1)))`

Should be:
`WHERE ((([umbracoNode].[nodeObjectType] = @4) AND ([umbracoNode].[trashed] = 0)))`

These indexes also help:
`CREATE NONCLUSTERED INDEX [IX_umbracoContentVersion_nodeId_current_include] ON [dbo].[umbracoContentVersion]
(
	[nodeId] ASC,
	[current] ASC
)
INCLUDE([id],[VersionDate],[text],[userId])
GO
`

`CREATE NONCLUSTERED INDEX [IX_umbracoNode_nodeObjectType_trashed_include] ON [dbo].[umbracoNode]
(
	[level] ASC,
	[parentID] ASC,
	[sortOrder] ASC,
	[nodeObjectType] ASC,
	[trashed] ASC
)
INCLUDE([nodeUser],[path],[uniqueID],[createDate])`


This is the query I'm referring to:
`exec sp_executesql N'SELECT * FROM (SELECT poco_base.*, ROW_NUMBER() OVER (ORDER BY [level], [parentId], [sortOrder]) poco_rn 
FROM ( 
SELECT [umbracoNode].[id] AS [Id], [umbracoNode].[uniqueId] AS [Uid], [umbracoNode].[level] AS [Level], [umbracoNode].[path] AS [Path], [umbracoNode].[sortOrder] AS [SortOrder], [umbracoNode].[parentId] AS [ParentId], [umbracoNode].[createDate] AS [CreateDate], [umbracoNode].[nodeUser] AS [CreatorId]
, [umbracoContent].[contentTypeId] AS [ContentTypeId]
, [umbracoDocument].[published] AS [Published], [umbracoDocument].[edited] AS [Edited]
, [umbracoContentVersion].[id] AS [VersionId], [umbracoContentVersion].[text] AS [EditName], [umbracoContentVersion].[versionDate] AS [EditVersionDate], [umbracoContentVersion].[userId] AS [EditWriterId]
, [umbracoDocumentVersion].[templateId] AS [EditTemplateId]
, [pcver].[id] AS [PublishedVersionId], [pcver].[text] AS [PubName], [pcver].[versionDate] AS [PubVersionDate], [pcver].[userId] AS [PubWriterId]
, [pdver].[templateId] AS [PubTemplateId]
, [nuEdit].[data] AS [EditData]
, [nuPub].[data] AS [PubData]
FROM [umbracoNode]
INNER JOIN [umbracoContent]
ON ([umbracoNode].[id] = [umbracoContent].[nodeId])
INNER JOIN [umbracoDocument]
ON ([umbracoNode].[id] = [umbracoDocument].[nodeId])
INNER JOIN [umbracoContentVersion]
ON (([umbracoNode].[id] = [umbracoContentVersion].[nodeId]) AND [umbracoContentVersion].[current] = @0)
INNER JOIN [umbracoDocumentVersion]
ON ([umbracoContentVersion].[id] = [umbracoDocumentVersion].[id])
LEFT JOIN [umbracoContentVersion] [pcver]
INNER JOIN [umbracoDocumentVersion] [pdver]
ON (([pcver].[id] = [pdver].[id]) AND [pdver].[published] = @1)
ON ([umbracoNode].[id] = [pcver].[nodeId])
LEFT JOIN [cmsContentNu] [nuEdit]
ON (([umbracoNode].[id] = [nuEdit].[nodeId]) AND NOT ([nuEdit].[published] = @2))
LEFT JOIN [cmsContentNu] [nuPub]
ON (([umbracoNode].[id] = [nuPub].[nodeId]) AND [nuPub].[published] = @3)
WHERE ((([umbracoNode].[nodeObjectType] = @4) AND NOT ([umbracoNode].[trashed] = @5)))
) poco_base ) poco_paged 
WHERE poco_rn > @6 AND poco_rn <= @7 
ORDER BY poco_rn',N'@0 int,@1 int,@2 int,@3 int,@4 uniqueidentifier,@5 int,@6 bigint,@7 bigint'`


These changes and indexes take the duration from 12521 down to 181 in the sql profiler

Shazwazza:
That's a big perf improvement! I'm sure the queries are reasonably easy to change but the indexes not so much because there is currently no support for Sql Server specific type indexes. There's a task lying around somewhere to make that happen but not sure where. The ROW_NUMBER() OVER syntax shouldn't be seen though, all paging is done with the more modern SQL server syntax so long as your aren't running a super old Sql server version.

I'd love to be able to have Sql server specific indexes in there if we can, it will require a bit of extra plumbing with the DTO configuration, etc... but should be doable. There's also a task somewhere about looking at many of the current DB indexes we have and seeing what should be changed since I'm sure we can get a lot more perf out of common queries if they can be done better.

nzdev:
Seems easy enough to add support.
https://github.com/umbraco/Umbraco-CMS/pull/8630

The indexes are doing most of the work to speed things up.

nzdev:
Many of the indexes are greatly improved by using include. Saves scanning the index + keylookup for the table, when you can just include what you need in the index. It's more space, but for boot times it's worth the trade off.

Shazwazza:
WOW! that was ridiculously fast 🎉 Yep I'm happy with the size tradeoff. Do you know if include columns also slow down updates/inserts? There's probably quite a lot of other queries for content and properties that could be much faster. I think there's also a bunch of indexes that never actually get used either. 
I have that PR starred to look at next week but we could look at including those changes and these additional indexes in the working nucache-perf branch too?

nzdev:
Another good one:
`CREATE UNIQUE NONCLUSTERED INDEX [IX_umbracoRedirectUrl] ON [dbo].[umbracoRedirectUrl]
(
	[urlHash] ASC,
	[contentKey] ASC,
	[culture] ASC,
	[createDateUtc] ASC
) INCLUDE ([url])
`

nzdev:
I'm happy to help get them out sooner, even if it's not on this branch. Though I'm very much waiting till this is done before I upgrade. It would also be nice to apply the same principals for those on v7. I've seen similar query issues.

Shazwazza:
Might be best then to update #8630 with the index changes you see fit with descriptions and then those can be part of an 8.8 release, I can also pick those changes into the `nucache-perf` branch too. v7 unfortunately won't really get updates unless they are critical.

nzdev:
It's more work for the sql server than not including them on updates/inserts. But those should happen far less often than reads, and if the reads are faster then that leaves more io/cpu for writes :)

nzdev:
Another thing to note is if you are on sql azure, some of the indexes that help will be automatically applied. It's a good place to start for looking at some improvements that should be brought in.

nzdev:
Locally I'm using sql server 2017 compatibility level re: ROW_NUMBER syntax

nzdev:
The way the paging sql is generated to get the content in DatabaseDataSource could be optimised. Any hints on why I'm not getting OFFSET FETCH generated would be good. 

At the moment it looks like QueryPaged<> just runs the whole query wrapped in Count(*). It would be more efficient to use a different query to get the count so it doesn't need to join the nucache table as it's only needed to retrieve the results.

`
 public IEnumerable<ContentNodeKit> GetAllContentSources(IScope scope)
        {
            var sql = ContentSourcesSelect(scope)
                .Where<NodeDto>(x => x.NodeObjectType == Constants.ObjectTypes.Document && !x.Trashed)
                .OrderBy<NodeDto>(x => x.Level, x => x.ParentId, x => x.SortOrder);
            foreach (var row in scope.Database.QueryPaged<ContentSourceDto>(PageSize, sql))
                yield return CreateContentNodeKit(row);
        }
`


Results in:
`SELECT COUNT(*) FROM (SELECT [umbracoNode].[id] AS [Id], [umbracoNode].[uniqueId] AS [Uid], [umbracoNode].[level] AS [Level], [umbracoNode].[path] AS [Path], [umbracoNode].[sortOrder] AS [SortOrder], [umbracoNode].[parentId] AS [ParentId], [umbracoNode].[createDate] AS [CreateDate], [umbracoNode].[nodeUser] AS [CreatorId]
, [umbracoContent].[contentTypeId] AS [ContentTypeId]
, [umbracoDocument].[published] AS [Published], [umbracoDocument].[edited] AS [Edited]
, [umbracoContentVersion].[id] AS [VersionId], [umbracoContentVersion].[text] AS [EditName], [umbracoContentVersion].[versionDate] AS [EditVersionDate], [umbracoContentVersion].[userId] AS [EditWriterId]
, [umbracoDocumentVersion].[templateId] AS [EditTemplateId]
, [pcver].[id] AS [PublishedVersionId], [pcver].[text] AS [PubName], [pcver].[versionDate] AS [PubVersionDate], [pcver].[userId] AS [PubWriterId]
, [pdver].[templateId] AS [PubTemplateId]
, [nuEdit].[data] AS [EditData]
, [nuPub].[data] AS [PubData]
, [nuEdit].[dataRaw] AS [EditDataRaw]
, [nuPub].[dataRaw] AS [PubDataRaw]
FROM [umbracoNode]
INNER JOIN [umbracoContent]
ON ([umbracoNode].[id] = [umbracoContent].[nodeId])
INNER JOIN [umbracoDocument]
ON ([umbracoNode].[id] = [umbracoDocument].[nodeId])
INNER JOIN [umbracoContentVersion]
ON (([umbracoNode].[id] = [umbracoContentVersion].[nodeId]) AND [umbracoContentVersion].[current] = @0)
INNER JOIN [umbracoDocumentVersion]
ON ([umbracoContentVersion].[id] = [umbracoDocumentVersion].[id])
LEFT JOIN [umbracoContentVersion] [pcver]
INNER JOIN [umbracoDocumentVersion] [pdver]
ON (([pcver].[id] = [pdver].[id]) AND [pdver].[published] = @1)
ON ([umbracoNode].[id] = [pcver].[nodeId])
LEFT JOIN [cmsContentNu] [nuEdit]
ON (([umbracoNode].[id] = [nuEdit].[nodeId]) AND NOT ([nuEdit].[published] = @2))
LEFT JOIN [cmsContentNu] [nuPub]
ON (([umbracoNode].[id] = [nuPub].[nodeId]) AND [nuPub].[published] = @3)
WHERE ((([umbracoNode].[nodeObjectType] = @4) AND NOT ([umbracoNode].[trashed] = @5)))
) npoco_tbl`

Where it would be better if it was:

`SELECT COUNT(*) FROM (SELECT [umbracoNode].[id] AS [Id]
FROM [umbracoNode]
INNER JOIN [umbracoContent]
ON ([umbracoNode].[id] = [umbracoContent].[nodeId])
INNER JOIN [umbracoDocument]
ON ([umbracoNode].[id] = [umbracoDocument].[nodeId])
INNER JOIN [umbracoContentVersion]
ON (([umbracoNode].[id] = [umbracoContentVersion].[nodeId]) AND [umbracoContentVersion].[current] = @0)
INNER JOIN [umbracoDocumentVersion]
ON ([umbracoContentVersion].[id] = [umbracoDocumentVersion].[id])
LEFT JOIN [umbracoContentVersion] [pcver]
INNER JOIN [umbracoDocumentVersion] [pdver]
ON (([pcver].[id] = [pdver].[id]) AND [pdver].[published] = @1)
ON ([umbracoNode].[id] = [pcver].[nodeId])
WHERE ((([umbracoNode].[nodeObjectType] = @4) AND NOT ([umbracoNode].[trashed] = @5)))
) npoco_tbl`

Which avoids the expensive index seek on cmsContentNu.
image

nzdev:
I'm using SQL Server 2019, which is not supported correctly. https://github.com/umbraco/Umbraco-CMS/pull/8634

Shazwazza:
Just looking into the COUNT  now

nzdev:
@Shazwazza I think that count query isn't called too often. Might not be worth the effort vs return.

Shazwazza:
@nzdev I've merged #8630 into the nucache-perf branch. I've just pushed some changes to regarding the COUNT query and the usage of NOT listed above - just waiting to see if the build server is happy with those changes. 

Regarding this index https://github.com/umbraco/Umbraco-CMS/pull/8365#issuecomment-673419356, any chance you could make a PR for that one with a description of it's importance? 

I'll look into the other ones you mentioned above

> CREATE NONCLUSTERED INDEX [IX_umbracoContentVersion_nodeId_current_include] ON [dbo].[umbracoContentVersion] ( [nodeId] ASC, [current] ASC ) INCLUDE([id],[VersionDate],[text],[userId]) GO

> CREATE NONCLUSTERED INDEX [IX_umbracoNode_nodeObjectType_trashed_include] ON [dbo].[umbracoNode] ( [level] ASC, [parentID] ASC, [sortOrder] ASC, [nodeObjectType] ASC, [trashed] ASC ) INCLUDE([nodeUser],[path],[uniqueID],[createDate])

nzdev:
@Shazwazza Happy to, though I might wait till you have written one of the migrations for the other two indexes mentioned so I have some migration code to go on.

Shazwazza:
@nzdev Have pushed a commit with the new indexes and migration here https://github.com/umbraco/Umbraco-CMS/pull/8376/commits/77e5a41077ea85ed1b8d9a9c2ace687275741d1e

I have removed the additional indexes on umbracoNode that are now all part of the same index. AFAIK SQL Server will use an index if a column is contained within it even if not all columns are being used in the query so seems silly to have single indexes on single columns along with one index that includes these columns. I could be wrong though, what do you think?

nzdev:
@Shazwazza Include columns are not sorted, so if removed from the key columns of an index it results in a sort having to be done. I'd probably still keep the single indexes, as any custom sql might be using them.

Shazwazza:
@nzdev yep that's cool but the indexes I've removed are just part of the new composite index columns, not the include columns. For example, I removed the umbracoNode indexes on parentId and trashed which existed individually and now there's a single index that contains both of these https://github.com/umbraco/Umbraco-CMS/commit/77e5a41077ea85ed1b8d9a9c2ace687275741d1e#diff-e4709993d1cc3013c994674ee9289084R57 I think I read something a while back that SQL server will still use a composite index for an individual column if it's part of the index. I guess that would just need some testing.

> It would probably be a matter of finding all the queries that used the existing index and seeing how they perform.

yep that's part of the other task I've mentioned before, there's quite a lot to review!

nzdev:
@Shazwazza I think you will find that's only true if the composite key is used in order.
Example:
```
CREATE NONCLUSTERED INDEX [NonClusteredIndex-20200825-164225] ON [dbo].[umbracoNode]
(
	[parentId] ASC,
	[path] ASC,
	[sortOrder] ASC
)
```

I can use order by [parentId] ASC, or [parentId] ASC, [path] ASC or [parentId] ASC ,[path] ASC ,[sortOrder] ASC but not any other combination as otherwise the ordering is wrong.

Shazwazza:
@nzdev thanks and yep that rings a bell. I found another missing index with the count query too so have pushed changes for that. 

I'm now back to just testing things and have realized that the effort so far with double compression only takes into account when this data actually gets into the bplustree file, it doesn't take into account when the data is already in memory. For example, this Optimize call https://github.com/umbraco/Umbraco-CMS/pull/8376/files#diff-cee6dacfc05fce993194a208fd939d12R81 is done when writing to the cmsContentNu table and based on config it will compress a property value. Then that data is read back out and Deserialize https://github.com/umbraco/Umbraco-CMS/pull/8376/files#diff-cee6dacfc05fce993194a208fd939d12R98 is called converting the msgpack bytes into ContentNestedData. But the property value that was compressed remains as compressed bytes.  This is taken care of when reading from bplustree with the `Lz4DictionaryOfPropertyDataSerializer` which sets the property Value to a `LazyCompressedString`, or immediate decompresses.

What I'm fairly sure needs to happen is that the Deserialize method https://github.com/umbraco/Umbraco-CMS/pull/8376/files#diff-cee6dacfc05fce993194a208fd939d12R98 needs to basically do what is in the `Lz4DictionaryOfPropertyDataSerializer` is doing. Just thought I'd ping you about this in case I'm not thinking clearly :P

nzdev:
Some other potentially worthwhile indexes. Sorry, I don't have the queries that call them handy.

```
CREATE NONCLUSTERED INDEX [IX_cmsContentNu_published] ON [dbo].[cmsContentNu]
(
	[published] ASC
)
INCLUDE([nodeId])
```

```
CREATE NONCLUSTERED INDEX [IX_umbracoContent_contentTypeId] ON [dbo].[umbracoContent]
(
	[contentTypeId] ASC
)
```

nzdev:
@Shazwazza 

I don't know enough about how bplustree is being used, how snapshot generations, how lookups / querying work on the tree etc to make this a really informed comment, however...

I,m not too sure at the moment around your comment on deserialization, but it sounds about right, the work the Lz4DictionaryOfPropertyDataSerializer is doing before the data gets into bplustree could be moved to the messagepack deserialization. It may then make sense to use a different prefix for the primitive serializer to mark the type as an lz4 compressed string so umbraco knows to lazily decompress the string.

Another option would be to just copy the messagepack bytes from the database directly to the bplustree stream, skipping deserializing to NestedContentData and have a prefix for messagepack bytes. This would mean deserializing/decompressing the whole message when the data is read from bplustree, so you most likely wouldn't compress a few individual properties in this scenario. But it would reduce cpu usage on boot as no deserialization/compression has to happen at boot.

Shazwazza:
Thanks will follow up with those indexes. 

So after going through much of this code and the various options, I just wanted to write what we are currently doing and a hope to simplify. I've made changes locally now and things are working. I'll push up these changes and probably some of the below tomorrow. 

## Current state of Proof of concept

### New cmsContentNu serialization

* We serialize and compress data in cmsContentNu as bytes using MessagePack

= OK (we'll probably remove configs for this and just make it the way it is with data migrations)

### property compression at cmsContentNu level

* We optionally with NucachePropertyCompressionLevel.SQLDatabase also compress individual propertly values before being serialized with LZ4 (this essentially 'double compresses' the property data).   
  * These bytes end up as the property data in memory and __directly transfer to bplustree already compressed__
  * These bytes can be set as LazyCompressedString which will be resolved by the IPropertyValueConverter calling ToString() for rendering (TODO: we should not rely on ToString(), implicit cast to string?)

### property compression at bplustree level

* We can optionally with NucachePropertyCompressionLevel.NuCacheDatabase compress individual property values into the bplustree file
  * and then on deserialization, eagerly expand or set as a lazy expanded
  * The string values will make their way to the IPropertyValueConverter for rendering, or the LazyCompressedString will be used and resolved with ToString() in IPropertyValueConverter

## Proposal

We remove the NucachePropertyCompressionLevel.NuCacheDatabase option since it makes more sense to do this from the top down. I realize that this option is here so that you can easily switch between compressed/non-compressed for testing but I feel like the end result of all of this will be that compression works great and if it needs to be disabled, it can be done so at the property level (entirely disabled) or a developer can implement their own `IContentNestedDataByteSerializer` and/or `IDictionaryOfPropertyDataSerializer` for full flexibility/control. 

Remove the Lz4DictionaryOfPropertyDataSerializer since I don't think we need this and we just treat all compressed bytes with LazyCompressedString. I don't really see the benefit of having an option to eagerly expand but if it's required then custom implementation of the serializers above could be done. 

I'm unsure about the property alias mapping since the in-memory cache will always need to have it's full alias, so this mapping only affects the size of the bplustree and cmsContentNu data and I'm unsure it's worth it. I think for now to simplify what is here we should remove it and if this level of customization is required then custom serializers can be implemented. 

What do you think?

nzdev:
Happy to have the NucachePropertyCompressionLevel.NuCacheDatabase left open as an option for someone to implement themselves. I think enough of the classes are public?

Implicit conversion to string sounds like a better approach than ToString() for LazyCompressedString.

Property alias mapping is likely not worthwhile given LZ4, interning the alias should suffice.

I'd like to open up this further by :
1. Inject ISerializer<ContentNodeKit> instead of ContentDataSerializer into  Umbraco.Web.PublishedCache.NuCache.DataSource.BTree.GetTree(). ContentNodeKit and ContentData would need to be public.
2. Make all the properties on ContentData virtual or use an interface for ContentData.
2.  Having Umbraco.Web.PublishedCache.NuCache.DataSource.BTree.GetTree() be an interface that's injected into PublishedSnapshotService and have MainDomRegister() call the interface.

This gives enough flexibility to change up how serialiation is done. There are some other options for serialization I'd like to try out that would require these changes.

Shazwazza:
@nzdev Ok great, i'll let you know when I've pushed some updates and then we can keep iterating from there! :)

Shazwazza:
@nzdev Have pushed changes. This strips back quite a lot but I think still leaves the fundamentals in place and there's probably more abstraction work to do like you mentioned in this comment https://github.com/umbraco/Umbraco-CMS/pull/8365#issuecomment-679893062. 

The main things being

* Changed interface of `IPropertyCompressionOptions`, there is no factory now and it just has a single method call of whether to compress the property or not, this also needs to take into account the content type which it now does
* Default implementation of this to compress data for all complex property editors

but there's a bunch more cleanup, etc.. you can see in rev: 9a06b6291ae0a636108727cc4af6fe1f28009d3c

I think there's some tests failing so will fix those up but otherwise this is all working and content is rendering. Would be interesting to run some benchmarks, etc...

nzdev:
Thanks @Shazwazza  This pr might help to get more benchmarks written. https://github.com/umbraco/Umbraco-CMS/pull/8728

nzdev:
Any interest in merging a more basic version of this for v7?. It would be good to be able to just use lz4 on the XML cache in the database, without message pack or the property compression etc. It would make a big difference.

Shazwazza:
Hi @nzdev as for v7 we aren't going to introduce any new features, any updates there will only be critical bug/security fixes.

nzdev:
Without providing results, I think compressing every string is hurting performance when I've been trying this code out.
I think opting in on a per document basis would be the way to go.

Shazwazza:
@nzdev I assume you mean for the entire changes done to date here? https://github.com/umbraco/Umbraco-CMS/pull/8376 and that it is compressing for every `CompressedStorage` editor?

Tests should be passing, its definitely time to start running benchmarks to see where it's at (by that I mean load testing mostly and comparing between this version and the current 8.x version). But wondering based on your gut feel where the perf issue is? Just rendering or saving or both?

nzdev:
Querying is slow if you are querying on string properties. It would be enough to remove the attribute from the properties. Seems to  hurt rendering significantly.

nzdev:
The double decompression is where I think the slowness is. I think the compression on save is ok, and decompressing the whole nucache bytes for each document on startup is fine as it's a tradeoff between cpu and database and decompressing works out faster. But having the overhead of decompressing after startup is not ideal.

PR #8368, opened 2020-07-01
V8/feature/block list hide contentin overlay v2

nielslyngsoe:
(New PR to avoid dragging old stuff with us.)

This PR is adding the ability to hide the content editor from the infinite editing overlay in the Block List property editor.
The content editor is already hidden if the editor is set to inline mode. But for the default mode, this enables the specific block type to force the content editor to be hidden. This is preferred if the custom view for the given block type delivers a UI that provides editing of the content.


warrenbuckley:
Code changes look fine - I will retest this new PR

warrenbuckley:
All good with testing - just waiting for the CI build to be happy

warrenbuckley:
I think the CI build server struggling as the C# unit test `Perform_Scheduled_Publishing` is now failing but run locally in this branch is fine & the note next to this test does mention this can be slow to run.

PR #8370, opened 2020-07-01
Adjust border radius of button in button group if having sub buttons

bjarnef:
### Prerequisites

### Description
When SMTP mail settings has been configurated in web.config the "Invite user" button is shown with a "Create user" subbutton.
However when this isn't configurated only "Create user" button is show, where the button has rounded corners on the left, but no border radius on the right.

**Before**

image

image

**After**

image

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/buttons/umb-button-group.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/buttons/umb-button-group.less
@@ -14,7 +14,7 @@
     display: flex;
 }
 
-.umb-button-group {
+.umb-button-group.-with-button-group-toggle {
 
     .umb-button__button {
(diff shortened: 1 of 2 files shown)


bjarnef:
@nathanwoulfe yes, I agree the buttons layout selector + search on the right should have same height, but whether these should be rounded as well is a decision HQ should make. For this I made the corners rounded as other button elements and since it looks strange with rounded corners on left side of button, but not on the right side of the button.

nathanwoulfe:
@bjarnef sounds sensible to me - other buttons should be a separate PR.

nul800sebastiaan:
> wondering though if it now makes sense for the layout selector buttons on the right-side of view to also be rounded? And now that I've noticed that, those buttons are different sizes - one is 30px, other is 32px (which is the same as the Create user button).

yes and yes! :-)

Thanks very much @bjarnef  👍

Issue #8372, opened 2020-07-02
e.CancelOperation() does not display message on ContentService.Moving event


creativesuspects:
When using e.CancelOperation(new EventMessage(...)) on ContentService.Moving event the move action is cancelled, but the message is not displayed in the back-office. The user still sees a success message. On the ContentService.Trashing event the custom message is being displayed. I haven't tested any of the other "ing" events yet.

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.3 (clean install)

Reproduction
------------

### Steps to reproduce

Hook in to the ContentService events to prevent a document from being moved:

```
namespace Application.Core.Composing
{
    public class PublishedContentEventsComposer : IUserComposer
    {
        public void Compose(Composition composition)
        {
            composition.Components().Append<PublishedContentEventsComponent>();
        }
    }

    public class PublishedContentEventsComponent : IComponent
    {
        public PublishedContentEventsComponent()
        {
        }

        public void Initialize()
        {
            ContentService.Moving += ContentService_Moving;
            ContentService.Trashing += ContentService_Trashing;
        }

        private void ContentService_Moving(IContentService sender, MoveEventArgs<IContent> e)
        {
            foreach (var moveInfo in e.MoveInfoCollection)
            {
                e.CancelOperation(new EventMessage("Unauthorized", "This document type can't be moved", EventMessageType.Error));
            }
        }

        private void ContentService_Trashing(IContentService sender, MoveEventArgs<IContent> e)
        {
            foreach (var moveInfo in e.MoveInfoCollection)
            {
                e.CancelOperation(new EventMessage("Unauthorized", "This document type can't be moved", EventMessageType.Error));
            }
        }

        public void Terminate()
        {
        }
    }
}
```

When trying to delete a document the user see the error message that is passed in through the e.CancelOperation() method. When trying to move a document the user doesn't see the error message, but instead sees the default success message.

![Video_2020-07-02_120708](https://user-images.githubusercontent.com/161342/86346975-f369d080-bc5d-11ea-8580-edf28fb644f5.gif)

### Expected result

Show custom message on ContentService.Moving event.

creativesuspects:
I think the following change in [ContentService.cs](https://github.com/umbraco/Umbraco-CMS/blob/d65041c2503fc85ac4619a4abcd703dcc20f067d/src/Umbraco.Core/Services/Implement/ContentService.cs) could fix this:

Line 1947: return;  // causes rollback

Should be:

[...]

PR #8375, opened 2020-07-02
Show custom EventMessage when cancelling ContentService.Moving event

creativesuspects:
### Prerequisites

https://github.com/umbraco/Umbraco-CMS/issues/8372

### Description

This PR fixes an issue where the `ContentService.Moving` event does not show a custom `EventMessage` to the user when the event is cancelled using the `e.CancelOperation(new EventMessage(...))` method. The user would always see the default success message even if the Moving event was cancelled.

I have basically just copied pieces of logic from the `ContentService.Trashing` event which is almost identical to the Moving event. You can use the following code to test the fix, just add the code to a PublishedContentEvents.cs file in the src/Umbraco.Web.UI/App_Code/ folder:

```
using System;
using Umbraco.Core;
using Umbraco.Core.Composing;
using Umbraco.Core.Events;
using Umbraco.Core.Models;
using Umbraco.Core.Services;
using Umbraco.Core.Services.Implement;

namespace Application.Core.Composing
{
    public class PublishedContentEventsComposer : IUserComposer
    {
        {
            composition.Components().Append<PublishedContentEventsComponent>();
        }
    }

    public class PublishedContentEventsComponent : IComponent
    {
        public PublishedContentEventsComponent()
        {
        }

        public void Initialize()
        {
            ContentService.Moving += ContentService_Moving;
        }

        private void ContentService_Moving(IContentService sender, MoveEventArgs<IContent> e)
        {
            foreach (var moveInfo in e.MoveInfoCollection)
            {
                e.CancelOperation(new EventMessage("Unauthorized", "This document type can't be moved", EventMessageType.Error));
            }
        }
        public void Terminate()
        {
        }
    }
}
```

Video_2020-07-02_152157

### DIFF
--- a/src/Umbraco.Core/Services/IContentService.cs
+++ b/src/Umbraco.Core/Services/IContentService.cs
@@ -290,7 +290,12 @@ namespace Umbraco.Core.Services
         /// <summary>
         /// Moves a document under a new parent.
         /// </summary>
-        OperationResult Move(IContent content, int parentId, int userId = Constants.Security.SuperUserId);
+        void Move(IContent content, int parentId, int userId = Constants.Security.SuperUserId);
+
+        /// <summary>
+        /// Moves a document under a new parent and shows event messages (if any).
+        /// </summary>
+        OperationResult MoveWithResult(IContent content, int parentId, int userId = Constants.Security.SuperUserId);
 
         /// <summary>
         /// Copies a document.
--- a/src/Umbraco.Core/Services/IContentService.cs
+++ b/src/Umbraco.Core/Services/IContentService.cs
@@ -290,7 +290,7 @@ namespace Umbraco.Core.Services
         /// <summary>
         /// Moves a document under a new parent.
         /// </summary>
-        void Move(IContent content, int parentId, int userId = Constants.Security.SuperUserId);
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Thanks @creativesuspects! Unfortunately we can't just update this method with a new return type, that will require everyone who upgrades to this version to recompile the code. Instead I would recommend creating a new method like `MoveWithResult` - it's ugly and you need to know it exists, but that can be documented.. This will be the only way we can maintain backwards compatibility and not break everything for a lot of people. 

Thanks for contributing so far!

creativesuspects:
@nul800sebastiaan: Thanks for the feedback. I will try to make the necessary to the PR.

creativesuspects:
@nul800sebastiaan: I've updated the PR. Question though: Can I use the new `MoveWithResult()` method in `ContentController.PostMove()` or does this method also need some extra work to maintain backwards compatibility?

creativesuspects:
@nul800sebastiaan Any news on this?

creativesuspects:
Any news on this? This issue is still present in Umbraco 8.10.1.

creativesuspects:
Any news on this? This issue is still present in Umbraco 8.10.1.

PR #8376, opened 2020-07-02
Performance improvements mainly around front-end caching

Shazwazza:
This is a branch of some ideas being chatted about in https://github.com/umbraco/Umbraco-CMS/pull/8365

* Reduce the size of data stored in the cmsContentNu table
  * Reducing this data size reduces the amount of data that needs to be transferred over the network which reduces the latency to load it
  * The data is now serialized with MessagePack and the keys used for the data format are single chars
  * All aliases used , including things like culture names to consume this format are now string interned
  * The data storage for this format is now raw bytes instead of serialized strings, this greatly reduces the data size but also removes the overhead for serialization and deserialization. This can be disabled to return to original behavior by: `<add key="Umbraco.Web.PublishedCache.NuCache.Serializer" value="JSON"/>`. MessagePack will be the default that we ship with. There's a check in place during start to detect if the config switch has changed and it will rebuild the nucache tables
  * Allows for custom pluggable serializers for this data
* Ability to compresses property values in memory/caches (i.e. RTE and complex json values) 
  * By default property level compression is turned off but it can be implemented in a custom way by registering a custom `IPropertyCacheCompressionOptions` which will determine if a property should be compressed in memory. Example: `composition.RegisterUnique<IPropertyCacheCompressionOptions, CustomPropertyCacheCompressionOptions>();`
* Improved SQL indexes which now also utilizes SQL Server INCLUDE columns (#8630)
* Improved SQL queries when rebuilding the in memory caches from the cmsContentNu table
* Adds string interning for common aliases like property aliases
* More primitive types supported in nucache (#9440)

Other changes:

* Fixes a startup issue that doesn't bubble the underlying boot failed exception if the container fails.
* Ability to have SQL Server indexes with INCLUDE columns 
* Still need to do some benchmarks between this and previous versions for
  * [x] Cold Boot time
  * [x] DB table size of cmsContentNu table
  * [x] Normal Boot time - need to compare what a cold boot vs normal boot looks like with these changes (in hopes we could eventually just get rid of the local cache files)
  * [ ] Rendering times with in-memory compressed properties (first load)
  * [x] Time to takes to rebuild the DB nucache table
  * [x] Running mem consumption

## Benchmark 

The raw numbers here are not the interesting part since that is largely dependent on my machine. It's the % that are the important figures. This Umbraco install has ~ 16k nodes.

### Table Sizes

|  | name | rows | reserved | data | index_size | unused
-- | -- | -- | -- | -- | -- | --
BEFORE | cmsContentNu | 15983 | 43344 KB | 42728 KB | 16 KB | 600 KB
AFTER | cmsContentNu | 15983 | 11664 KB | 10096 KB | 48 KB | 1520 KB

With MsgPack binary serialization the database table size is reduced by ~ __73%__

### Cold Boot time

_An average based on 10 runs_

  | ave | max | min
-- | -- | -- | --
BEFORE | 18586.63636 | 33267 | 5675
AFTER with cmsContentNu as   JSON | 6586.444444 | 7365 | 5970
AFTER with cmsContentNu as   MsgPack | 5615.090909 | 6331 | 4627

The new code changes and with the MsgPack system performs ~ __70%__ better than the original implementation. It's also far more consistent. Before on average every 2nd cold boot would have a very long boot time.

### Normal Boot time

This is based on the normal boot process (non cold boot) to see what the perf difference is of having the local nucache files vs loading directly from the database (i.e. `IgnoreLocalDb = true`). With the new DB indexes and code changes we perform about 15% faster when loading in memory nucache from the DB (this figure would be larger if running the database over the network/internet). But as you can see it's still quite a lot faster to have the local nucache files. 

  | ave | max | min
-- | -- | -- | --
BEFORE Boot without nucache   files | 5321.909091 | 7007 | 4394
AFTER Boot without nucache   files | 4520.454545 | 5115 | 3911
AFTER Boot with nucache files | 2449.777778 | 3085 | 1949

## Cold boot memory

This is a comparison of mem consumption during cold booting between this codebase and 8.13. This was run several times with the same result. The cold boot memory footprint of this code base is lower than 8.13

From bootup until page render. Significantly less memory.

image

After first page render and navigating several pages. This one is the most surprising because it seems that a cold boot before these changes even after boot requires more memory than normal to run.

image

## Warm boot memory

This is a comparison of mem consumption during warm booting between this codebase and 8.13. This was run several times with the same result. The cold boot memory footprint of this code base is lower than 8.13

From bootup until page render. .Not a huge change but still less memory.

image

After first page render and navigating several pages. Not a huge change but still less memory.

image

## String Duplicates

In this codebase there is some usages of string interning. Based on profiling you can see that this is working a bit but it's not perfect. We are trying to intern all common strings that are dynamic such as property type aliases, content type aliases, etc... Here's a comparison of this codebase vs 8.13 with the same data in the database and sorted by the highest duplicate string count. You can see that a lot of alias duplicates are reduced but they certainly still exist. We should continue to string intern these values throughout the codebase in the future to further reduce this.

This codebase:

image

8.13:

image


src/Umbraco.Web/PublishedCache/NuCache/DataSource/CultureVariation.cs
```diff
@@ -1,35 +1,45 @@
...
+        [DataMember(Order = 2)]
         [JsonProperty("dt")]
         public DateTime Date { get; set; }
 
+        [DataMember(Order = 3)]
         [JsonProperty("isd")]
         public bool IsDraft { get; set; }
 
         //Legacy properties used to deserialize existing nucache db entries
+        [DataMember(Order = 4)]
```
nzdev, line 30: I don't think those legacy properties should be serialized for message pack [IgnoreMember]

src/Umbraco.Web/PublishedCache/NuCache/DataSource/PropertyData.cs
```diff
@@ -28,22 +33,26 @@ public string Segment
...
         }
 
+        [DataMember(Order = 2)]
         [JsonProperty("v")]
         public object Value { get; set; }
 
         //Legacy properties used to deserialize existing nucache db entries
+        [DataMember(Order = 3)]
         [JsonProperty("culture")]
         private string LegacyCulture
```
nzdev, line 43: These legacy properties could be removed if a migration forced a rebuild of the nucache table in the database. But they are currently necessary as umbraco will fail to boot otherwise

src/Umbraco.Web/PublishedCache/NuCache/DataSource/CultureVariation.cs
Shazwazza: yep i can remove that but they don't actually serialize when they aren't there but happy to add the attribute

src/Umbraco.Web/PublishedCache/NuCache/DataSource/PropertyData.cs
Shazwazza: yep we'll need to keep them unfortunately

nzdev:
Further looking into boot up, DatabaseDataSource.ContentSourcesSelect could be optimised. Line 62
`  .LeftJoin<ContentNuDto>("nuEdit").On<NodeDto, ContentNuDto>((left, right) => left.NodeId == right.NodeId && !right.Published, aliasRight: "nuEdit")`

which results in:
`LEFT JOIN [cmsContentNu] [nuEdit]
ON (([umbracoNode].[id] = [nuEdit].[nodeId]) AND NOT ([nuEdit].[published] = 1))`

Where it should be:
``LEFT JOIN [cmsContentNu] [nuEdit]
ON (([umbracoNode].[id] = [nuEdit].[nodeId]) AND ([nuEdit].[published] = 0))``

cmsContentNu.Published is a bit column, so it's only 0 or 1. Usually doing NOT where clauses results in indexes not being used.

These indexes also help:
`CREATE NONCLUSTERED INDEX [IX_umbracoContentVersion_nodeId_current_include] ON [dbo].[umbracoContentVersion]
(
	[nodeId] ASC,
	[current] ASC
)
INCLUDE([id],[VersionDate],[text],[userId])
GO
`

`CREATE NONCLUSTERED INDEX [IX_umbracoNode_nodeObjectType_trashed_include] ON [dbo].[umbracoNode]
(
	[nodeObjectType] ASC,
	[trashed] ASC
)
INCLUDE([parentID],[nodeUser],[level],[path],[sortOrder],[uniqueID],[createDate])`

nzdev:
My comments on this have been way off. There are two enums. I've added descriptions of what I had intended. The NucachePropertyCompressionLevel.SQLDatabase  is the preferred option. I was mixing it up with decompression.
public enum NucachePropertyCompressionLevel
{
None = 0, //No property compression
SQLDatabase = 1, //Compress the property before being written to the database. Idea being we only compress this once. All the records in cmsContentNu need to be rebuilt when this gets enabled.
NuCacheDatabase = 2 // Compress the property when writing to nucache bplustree after reading from the database. Idea being we compress this at rebuild / boot. This option supports older items not being compressed already.
}
///


/// If/where to decompress custom properties for nucache
///
public enum NucachePropertyDecompressionLevel
{
NotCompressed = 0, // No decompression required
Immediate = 1, //Decompress the property after it's read from the database (at boot or rebuild). Idea being we do the decompression at boot.
Lazy = 2 //Decompress the property when .ToString() is called. Idea being we decompress when the value is read.
}

nzdev:
The sql database option should probably be put back in.

nzdev:
Made a few updates to my comments on the enum.

Shazwazza:
Hrm, so are you saying you would prefer not to use the appSetting `Umbraco.Web.PublishedCache.NuCache.Serializer="MsgPack"` to serialize the nucache db document to binary? Didn't we determine that this is much better/faster for startup https://github.com/umbraco/Umbraco-CMS/pull/8365#issuecomment-653362591 

If we store the data as binary than there's no reason for the `NucachePropertyCompressionLevel.SQLDatabase` because it will never be used to compress property data at the SQL database level since that is only done when serializing the nucache DB document as json and not binary.

nzdev:
At the moment, both options only work for msg pack. With the sql database option we would compress the property to a byte[] then compress that along with the other properties using message pack. It's a bit of a double up, but given LZ4 is fast it's worthwhile.

nzdev:
Without the SQLDatabase option time spent on compression at boot will grow with the content, and cpu / memory would be used on potentially infrequently accessed content. If you have a large number of documents, this could be significant.

Shazwazza:
Aha, i finally got it. I was looking in the PublishedSnapshotService where it uses `IContentNestedDataByteSerializer` or `IContentNestedDataSerializer`, I forgot that we can actually write to the 2x columns, but I get it now. I'll add that option back in!

It's getting late in AU too, won't ask too many more questions :P 

> You would probably only have this on for richtext content or json.

yep for sure

>  I'd suggest in future the json strategy is dropped entirely given our observations. 

I assume you just mean we just would make the binary option the norm? We'd still keep the JSON column since if we removed that you won't be able to have the SQLDatabase option?

nzdev:
Unless I've misread the code again that I wrote, I wrote the SQLDatabase option to work with the messagepack option, with no intention of using it with the json serializer. Ideally a migration would drop the cmsContentNu table  and rebuild it without the json column for messagepack serializer as switching between the message pack and json is no fun.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/SerializerBase.cs
```diff
@@ -6,59 +6,95 @@ namespace Umbraco.Web.PublishedCache.NuCache.DataSource
...
-                case 'F':
+                case PrefixFloat:
                     return PrimitiveSerializer.Float.ReadFrom(stream);
-                case 'B':
+                case PrefixDouble:
                     return PrimitiveSerializer.Double.ReadFrom(stream);
-                case 'D':
+                case PrefixDateTime:
                     return PrimitiveSerializer.DateTime.ReadFrom(stream);
+                case PrefixByteArray:
```
nzdev, line 96: It might make sense to have another prefix for an LZ4 compressed byte array.

Shazwazza:
> Unless I've misread the code again that I wrote, I wrote the SQLDatabase option to work with the messagepack option, with no intention of using it with the json serializer

From what I can tell, the `MsgPackContentNestedDataSerializer.Serialize` which then calls the `MsgPackContentNestedDataSerializer.Optimize` method which is the code that checks for the `SQLDatabase` option would only be called if the `_contentNestedDataSerializer` is *not* `IContentNestedDataByteSerializer`. When `IContentNestedDataByteSerializer` is being used, only `SerializeBytes` is called and in that method it doesn't do anything with properties it just uses MsgPack to write the bytes for the whole doc: https://github.com/umbraco/Umbraco-CMS/pull/8376/files#diff-a7f4263c248ded990a4eac4c9943276fR1462

If you wanted to double compress the property data and also use MsgPack than `Optimize` would need to be used within `MsgPackContentNestedDataSerializer.SerializeBytes`?

src/Umbraco.Web/PublishedCache/NuCache/DataSource/NuCacheCompressionOptions.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Umbraco.Web.PublishedCache.NuCache.DataSource
+{
+    public struct NuCacheCompressionOptions : IEquatable<NuCacheCompressionOptions>
```
nzdev, line 6: This could also be an interface that takes the document / media and lets the implementer determine per document whether to compress or not. E.x. compress documents older than 90 days.

Shazwazza: Is there any reason you would want to compress property values based on time based logic like that? Seems to me like having all complex editors that store JSON or rich text should just be compressed - either on or off. I guess I'm just wondering what the benefit of not compressing 'new' rich text would be?

nzdev: Saves on decompressing frequently used content when requests arrive. LZ4 seems fast enough though that this may not be an issue, but I can't say for sure for everyone.

nzdev: Potentially you could even remove properties and retrieve them from elsewhere.

Shazwazza: > Saves on decompressing frequently used content when requests arrive. LZ4 seems fast enough though that this may not be an issue, but I can't say for sure for everyone.

Cool I just wanted to check if I was missing something. I'll need to do some rendering benchmarks with the decompression happening for properties and see what kind of difference that makes.

src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
```diff
@@ -630,6 +630,13 @@ private static void AppendNodeToFix(IDictionary<int, List<NodeDto>> nodesToRebui
...
 
+            // TODO: This is a bugger of a query and I believe is the main issue with regards to SQL performance drain when querying content
+            // which is done when rebuilding caches/indexes/etc... in bulk. We are using an "IN" query on umbracoPropertyData.VersionId
+            // which then performs a Clustered Index Scan on PK_umbracoPropertyData which means it iterates the entire table which can be enormous!
+            // especially if there are both a lot of content but worse if there is a lot of versions of that content.
+            // So is it possible to return this property data without doing an index scan on PK_umbracoPropertyData and without iterating every row
+            // in the table?
+
             // get all PropertyDataDto for all definitions / versions
             var allPropertyDataDtos = Database.FetchByGroups<PropertyDataDto, int>(versions, 2000, batch =>
```
nzdev, line 641: I don't think this is called during startup, just from the rebuild database cache. 
Changing the batch size to 500 results in a index seek instead of an index scan. This call often causes timeouts. 

I think this method could be reordered. It may be better to query the database first for the PropertyTypeDto's rather than getting the ids from the propertyDataDtos. This method is also really memory expensive as it loads all published and new versions of all content into memory at the same time, it may be better to page through them, adding them to the dictionary to smooth out the load on the database server.

Shazwazza: > I don't think this is called during startup, just from the rebuild database cache.

Yep this is called whenever IContent or collections of it is queried. The grouped results to 500 using an index seek is interesting though. Do you know if there's a certain limit to when it decides to start scanning instead of seeking?

Changing this to not load everything into memory could only be done by refactoring a lot of how this method gets used in the first place. For queries where tons are returned - like when rebuilding the db cache or when displaying content in list views it could be interesting to see how this could be improved. 

For now however do you think it's worth just changing the 2000 -> 500 ?

nzdev: I'm not really sure when seek to scan change happens, 500 seemed to result in an index seek. It may need bench marking to know if its better to make 4x the queries vs the table scan, but given query at 2000 this times out frequently, I would consider it worthwhile using the smaller batch size.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/NuCacheCompressionOptions.cs
nzdev: Potentially it might be better to just copy the compressed messagepack bytes straight to bplustree rather than deserializing messagepack, serialising again using messagepack and or the primitive serializers. It would need benchmarking, but given the data would still be compressed it may be faster to read off the disk doing it this way. Alternatively, if [ZeroFormatter](https://github.com/neuecc/ZeroFormatter) is suitable (not sure, never used it), it supports lazy deserialization at the cost of a larger size, could be used as well.

src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
Shazwazza: Yep will look into it, thanks!

Shazwazza: When testing locally and changing to 500 it unfortunately continues to use an index scan when a lot of versions are involved. A big part of the timeout issue was that it was trying to page over page sizes of 5000 for the entire content/media data, this has now been changed to be 1000 or configurable https://github.com/umbraco/Umbraco-CMS/pull/8376/files#diff-a7f4263c248ded990a4eac4c9943276fL1480  which helped with the timeouts I was getting locally. I think for individual content items or even a few items at a time these queries work ok but when we know we are going to be looking up *all* content or media I think perhaps this entire lookup process should be reconsidered. (ideally the database would just be overhauled ... but that's a whole other story)

src/Umbraco.Web/PublishedCache/NuCache/DataSource/SerializerBase.cs
```diff
@@ -98,7 +98,9 @@ protected object ReadObject(char type, Stream stream)
                 case PrefixDateTime:
                     return PrimitiveSerializer.DateTime.ReadFrom(stream);
                 case PrefixByteArray:
-                    return PrimitiveSerializer.Bytes.ReadFrom(stream);
+                    // When it's a byte array always return as a LazyCompressedString
+                    // TODO: Else we need to make a different prefix for lazy vs eager loading
+                    return new LazyCompressedString(PrimitiveSerializer.Bytes.ReadFrom(stream));
```
nzdev, line 103: I think a different prefix must be used otherwise you can't serialize/deserialize a byte array that's not a string

src/Umbraco.Web/PropertyEditors/ComplexEditorPropertyCompressionOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+
+                if (!_propertyEditors.TryGet(propertyType.PropertyEditorAlias, out var propertyEditor)) return null;
+
+                var editor = propertyEditor.GetValueEditor();
+                if (editor == null) return null;
+
+                return editor.ValueType;
+            });
+
+            return valueType == ValueTypes.Json || valueType == ValueTypes.Xml || valueType == ValueTypes.Text;
```
nzdev, line 45: If possible, I think only compressing json, xml, and rich text (html) would be better than compressing every string type.

src/Umbraco.Web/PublishedCache/NuCache/NuCacheComposer.cs
```diff
@@ -14,36 +15,22 @@ public override void Compose(Composition composition)
             base.Compose(composition);
 
             var serializer = ConfigurationManager.AppSettings["Umbraco.Web.PublishedCache.NuCache.Serializer"];
-            composition.Register<INuCachePropertyOptionsFactory, AppSettingsNuCachePropertyMapFactory>();
-            composition.Register<Lz4DictionaryOfPropertyDataSerializer, Lz4DictionaryOfPropertyDataSerializer>();
-
-            // TODO: Based on our findings it seems like this should not be configurable, we should just be using this because it's better
-            if (serializer == "MsgPack")
+            if (serializer != "MsgPack")
```
nzdev, line 18: By default it's not messagepack, unless a migration adds the appsetting

src/Umbraco.Web/PropertyEditors/ComplexEditorPropertyCompressionOptions.cs
Shazwazza: Yep Text = NText, String is a short string but yep the plan later is to possibly attribute a property editor that it supports compression so it's defined by the property editor not directly by it's value type. This is sort of a temporary thing that can evolve.

src/Umbraco.Core/Persistence/Dtos/NodeDto.cs
```diff
@@ -20,7 +20,7 @@ internal class NodeDto
 
         [Column("uniqueId")]
         [NullSetting(NullSetting = NullSettings.NotNull)]
-        [Index(IndexTypes.UniqueNonClustered, Name = "IX_" + TableName + "_UniqueId")]
+        [Index(IndexTypes.UniqueNonClustered, Name = "IX_" + TableName + "_UniqueId", IncludeColumns = "parentId,level,path,sortOrder,trashed,nodeUser,text,createDate")]
```
nzdev, line 23: This index can't be modified using the current code as it's a unique constraint used for a foreign key. It may be better to add a second index.

nzdev: Breaks upgrade from 8.6.4

nzdev:
A wip PR for LiteDb as a nucache provider, as an extension of the pr to add interfaces for nucache #8898. Could be a reasonable starting point for discussing querying nucache. The PR doesn't as written provide an interface for querying, but one could open a second litedb connection in readonly and run the query. @Shazwazza

src/Umbraco.Web/PublishedCache/NuCache/NuCacheComposer.cs
Shazwazza: Yep, we'll most likely ship with this setting which means it will be opt-in so turned on for new sites but for upgrades people will have to opt in manually. I'll add this to the TODO notes on this issue (which will keep growing). Possibly we make it default for everything including upgrades which will then require migrations.

src/Umbraco.Core/Persistence/Dtos/NodeDto.cs
Shazwazza: The migration is fixed in the latest rev

Shazwazza:
@nzdev I've just pushed a large-ish revision: 67a9b5bb9770908f431760f88783b00d64839fae 

> Refactors the serialization of the content data that is stored in the nucache table. This had to change because we need to resolve content type data in order to check if the property should be compressed and we cannot do that data lookup while the data is being processed since we get an open data reader exception. This is fixed now by using a serializer factory instead so the Create method can do any initialization needed prior to running any serialization operation. Renames a few things so we dont have ContentNested (whatever that meant )

The way it was before was not working with my previous changes because the IPropertyCompressionOptions was performing DB lookups while nucache might have been rebuilding which would mean an open db reader exception because nucache doesn't fetch all the rows into memory. So to fix this the serializer is now a factory which is what is registered in DI and this will be used to resolve a serializer for a serialization session which gives us a chance to initialize any data before any lookups start. 

I've also changed the default IPropertyCompressionOptions implementation to now look at property editors that have been attributed as [CompressedStorage]. So now any property editor that has this attribute will have it's data compressed. 

Let me know any questions/concerns/comments/etc... Cheers! I'll hopefully get more time to look at a few of these things tomorrow and in the coming couple weeks.

nzdev:
No concerns except the failing test

src/Umbraco.Core/Migrations/Upgrade/UmbracoPlan.cs
```diff
@@ -202,6 +202,9 @@ protected void DefinePlan()
             // to 8.9.0...
             To<AddCmsContentNuByteColumn>("{8DDDCD0B-D7D5-4C97-BD6A-6B38CA65752F}");
```
nzdev, line 203: Order needs changing

Shazwazza: haha yes working on it :P

src/Umbraco.Web/PublishedCache/NuCache/DataSource/MsgPackContentNestedDataSerializer.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        /// <summary>
+        /// Used during serialization to compress properties
+        /// </summary>
+        /// <param name="model"></param>
+        private void Compress(int contentTypeId, ContentCacheDataModel model)
+        {
+            foreach(var propertyAliasToData in model.PropertyData)
+            {
+                if (_propertyOptions.IsCompressed(contentTypeId, propertyAliasToData.Key))
```
nzdev, line 86: I don't think this should be default as it's quite slow on decompression

Shazwazza: I'm assuming you mean the double compression as the default? Right now if you want to test the difference, you can replace `IContentCacheDataSerializerFactory` in DI with an overridden version of `MsgPackContentNestedDataSerializerFactory` and just return a new instance of `MsgPackContentNestedDataSerializer` with empty `IPropertyCompressionOptions` that always returns false. 

Is it just the first published query that is slow since it needs to decompress things? Since once the first one is done it will be decompressed in memory right?

nzdev: If you have a few different queries it's slow for a while

Shazwazza: OK cool, will change how this works with the ability to opt-in to manage what might be compressed in memory but will leave it uncompressed (so we only compress with MsgPack at the db level). This task is still ongoing, sorry the work is slow but I can only allocate a bit of time each sprint.

nzdev: Perharps we could break this up into multiple smaller prs.

Shazwazza: Yes that could be possible, I've just finished re-reviewing a lot of it and making some notes. I need to resolve this one https://github.com/umbraco/Umbraco-CMS/pull/9440/files since this https://github.com/umbraco/Umbraco-CMS/pull/9440#discussion_r544006996 isn't quite right and have run into issues there. 

* Compression is one thing - which is a majority of these changes
* Improved Sql indexes + INCLUDE support could be separated 
* Improved Sql queries could prob be separated out but that does have some overlap with the compression db changes
* Some of the interning could be separated out if it's easy but there's a lot of stuff for that which relates to the changes for compression
* Other "Fixes a startup issue that doesn't bubble the underlying boot failed exception if the container fails." could be separated

If you have time let me know but I will continue to resolve the last bits here regarding compression

nzdev:
I don't know exactly how it works, but Umbraco.Web.PublishedCache.NuCache.Property is a memory hog.

Shazwazza:
> I don't know exactly how it works, but Umbraco.Web.PublishedCache.NuCache.Property is a memory hog.

I assume that's what shows up as a high number of items in a mem profiler? One of these `Umbraco.Web.PublishedCache.NuCache.Property` will exist in memory for every property for every IPublishedContent. That is the implementation of `IPublishedProperty` for IPublishedContent. Not sure about a 'hog' but yes if you have a huge amount of content then there will be a lot of these. In v7 the XML document was in memory and traversing that would created new ones of these for every IPublishedContent traversed resulting in huge amounts of memory used and heavy pressure on the GC/CPU. In v8 the actual IPublishedContent objects are in memory, not an xml document so traversing doesn't create new objects. I'm sure you know all/most of this but just explaining again.

nzdev:
The count is large which is to be expected. But the total memory is surprising and correlates with the size of the rte content I have which is odd as I'm only reading the value of the rte property for a few of the instances. Given the size of the property data, it's odd.
image

Shazwazza:
That is a lot of mem! But you have a huge amount of content too so I don't have enough context of your site to know if that is odd or not. 

> as I'm only reading the value of the rte property for a few of the instances

Doesn't really matter if you are reading the data or not, all property data is in memory. We don't support lazy loading this into memory due to a ton of factors (like how routing works, etc...). Part of the work being done here that you've done is to help reduce this. Is this the mem usage with the double compression enabled? 

The other thing to note - which may be directly related to what you are seeing is that there can be 2x copies of each property in memory if there is a draft version of the content - the published content and the edited version. I've never really dived into this area but it is true that if you had draft versions of all of your content items, you could essentially have 2x the amount of data for a single content item in mem. This is set here https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PublishedCache/NuCache/DataSource/BTree.ContentNodeKitSerializer.cs#L28 . There doesn't seem to be any easy way to lazy load this data if we wanted to only load the draft data if/when preview is requested. And then ideally it would make a lot of sense to have that draft data removed from the in-mem cache when not in use (sliding time). Though unless preview is requested, the actual `IPublishedContent` model isn't created from the raw draft data. That is only done here lazily https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PublishedCache/NuCache/ContentNode.cs#L164

It would probably be worth investigating this at some point and maybe you would be able see in your memory snapshots if a lot of this data is draft data or not?

I also noticed this setting: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PublishedCache/NuCache/PublishedSnapshotService.cs#L70 and I think this was meant to be revisited prior to v8 launch but it was not ... hence there is no configuration for this. From a quick look it seems that the field controls the property level and route caches when in preview mode. 

I'm just about to push some changes that removes the `[CompressedStorage]` so no double compression is on by default. We can ship with simple extension points to configure double compression to easily write your own options in code to determine what to double compress. I'm also updating the code for the IsCompressed check to take an object that will provide a lot more context/info about the content item to decide if it's double compressed since before there was very limited data to determine that.

Shazwazza:
FYI I've updated the description for this PR to include some more details. I've removed the `CompressedStorageAttribute`, in-memory property compression is turned off but can easily be configured in code by registering your own `IPropertyCacheCompressionOptions` which will also let you configure it based on the age of the content item since we now pass in IReadOnlyContentBase  to that service.

nzdev:
Hi @Shazwazza  any thoughts on what's still outstanding on this branch before merging in? I'm thinking perhaps we remove messagepack from being bundled and providing either documentation or another package on how to get messagepack set up. There are plenty of improvements in this branch that will speed things up without switching to messagepack just yet.

Shazwazza:
@nzdev hey, yes it still a task on my list and it does get assigned to sprints but a few things have pushed my time out. I will get to it and there's only a few things to test that are remaining. It will remain scheduled until I can get to it. With Easter break however it won't be until after that.

src/Umbraco.Core/Sync/DatabaseServerMessenger.cs
```diff
@@ -110,14 +111,14 @@ protected override bool RequiresDistributed(ICacheRefresher refresher, MessageTy
...
-        /// </summary>
-        /// <remarks>
-        /// Thread safety: this is NOT thread safe. Because it is NOT meant to run multi-threaded.
-        /// Callers MUST ensure thread-safety.
-        /// </remarks>
+        [Obsolete("This is no longer used and will be removed in future versions")]
         protected void Boot()
+        {
+            // if called, just forces the boot logic
+            _ = GetSyncBootState();
```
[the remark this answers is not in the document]
Shazwazza: It won't, that's not what is happening. The published cache will only load when it's accessed

Shazwazza:
I've added more benchmarks (above). The running memory and rebuild times aren't significantly different so nothing to really report on there. Though, that might be different if there are tens of thousands of nodes. 

The only thing I'd still like to benchmark is the effect of having the compressed property data and rendering.

I've moved this as ready to review. I can see that v8/contrib needs to be re-merged in though.

src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
```diff
@@ -630,6 +630,13 @@ private static void AppendNodeToFix(IDictionary<int, List<NodeDto>> nodesToRebui
             }
             if (versions.Count == 0) return new Dictionary<int, PropertyCollection>();
 
+            // TODO: This is a bugger of a query and I believe is the main issue with regards to SQL performance drain when querying content
+            // which is done when rebuilding caches/indexes/etc... in bulk. We are using an "IN" query on umbracoPropertyData.VersionId
+            // which then performs a Clustered Index Scan on PK_umbracoPropertyData which means it iterates the entire table which can be enormous!
+            // especially if there are both a lot of content but worse if there is a lot of versions of that content.
+            // So is it possible to return this property data without doing an index scan on PK_umbracoPropertyData and without iterating every row
+            // in the table?
```
stevetemple, line 638: What's the oldest version of SQL server Umbraco supports. Essentially are indexed views a possibility?

Shazwazza: Currently says 

> SQL Server 2012 and higher

https://our.umbraco.com/documentation/Fundamentals/Setup/Requirements/

src/Umbraco.Tests/Umbraco.Tests.csproj
```diff
@@ -78,6 +78,8 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Castle.Core" Version="4.4.1" />
+    <PackageReference Include="Examine" Version="1.0.2" />
     <PackageReference Include="Castle.Core" Version="4.3.1" />
     <PackageReference Include="Examine" Version="1.2.0" />
```
bergmania, line 84: Duplicates

src/Umbraco.Tests/Umbraco.Tests.csproj
```diff
@@ -100,6 +102,9 @@
     <PackageReference Include="Microsoft.Owin.Testing" Version="4.0.1" />
     <PackageReference Include="Microsoft.Web.Infrastructure" Version="1.0.0.0" />
     <PackageReference Include="MiniProfiler" Version="4.0.138" />
+    <PackageReference Include="Moq">
+      <Version>4.14.5</Version>
+    </PackageReference>
     <PackageReference Include="Moq" Version="4.10.1" />
```
bergmania, line 108: Duplicate

src/Umbraco.Web/PublishedCache/NuCache/DataSource/MsgPackContentNestedDataSerializer.cs
```diff
@@ -0,0 +1,207 @@
...
+        //        }
+
+        //        options.Security.DepthStep(ref reader);
+
+        //        var path = reader.ReadString();
+
+        //        reader.Depth--;
+        //        return new FileInfo(path);
+        //    }
+        //}
```
bergmania, line 205: I guess this should be removed..

bergmania (approved):
Looks good.. Just small comments. I fear the merge to v9 🙈

src/Umbraco.Core/Migrations/Upgrade/UmbracoPlan.cs
```diff
@@ -198,10 +199,14 @@ protected void DefinePlan()
...
             // to 8.9.0
             To<ExternalLoginTableUserData>("{B5838FF5-1D22-4F6C-BCEB-F83ACB14B575}");
-			
+
             // to 8.10.0
             To<AddPropertyTypeLabelOnTopColumn>("{D6A8D863-38EC-44FB-91EC-ACD6A668BD18}");
 
+            // to 8.12.0...
+            To<AddCmsContentNuByteColumn>("{8DDDCD0B-D7D5-4C97-BD6A-6B38CA65752F}");
+            To<UpgradedIncludeIndexes>("{4695D0C9-0729-4976-985B-048D503665D8}");
```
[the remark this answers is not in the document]
Shazwazza: yep, I've updated to 8.15 for now.

Shazwazza:
Happy to help with merge to v9. I've updated the PR based on your comments.

Shazwazza:
One thing to test that I don't know if you did is to change the config:

<add key="Umbraco.Web.PublishedCache.NuCache.Serializer" value="MsgPack"/>

to 

<add key="Umbraco.Web.PublishedCache.NuCache.Serializer" value="JSON"/>

which will force the nucache table to rebuild on startup with the correct format. You can verify this by looking in the nucache DB table. WIth MsgPack it will populate the binary column, with JSON it will populate the string column.

bergmania:
> One thing to test that I don't know if you did is to change the config:
> ....


I tested this and it seems to work as expected 🎉 

I also executed the acceptance tests.. A Single test does not pass, but that is not related to this PR.

Shazwazza:
Please remember to mark with a release version. I've marked as 8.15.

nul800sebastiaan:
Thanks @Shazwazza I was about to do this as well! 👍

ronaldbarendse:
@Shazwazza I just noticed some separate commits to fix the migration from this PR to support SQL CE, but that doesn't seem to drop the temp table after moving all rows back to the original table:

https://github.com/umbraco/Umbraco-CMS/blob/7dc03256edf1bbe15b5b44c5d44a033369ce0e07/src/Umbraco.Core/Migrations/Upgrade/V_8_15_0/AddCmsContentNuByteColumn.cs#L22-L32

Shazwazza:
@ronaldbarendse good catch - I guess we're too late now and it will need to be in the next minor release 🤦‍♂️ 🤦‍♂️ 🤦‍♂️  

@nul800sebastiaan any thoughts?

ronaldbarendse:
We can add a release note mentioning you can manually remove the table (`DROP TABLE [umbracocmsContentNuTEMP]`) _if you're using SQL CE and upgraded your site to 8.15.0_.

Because it's only on SQL CE, this should only happen on development databases, so it's not a big issue anyway. And if we fix up this migration in a next patch release, upgrading a pre 8.15.0 site to 8.15.1 won't have this problem 👍🏻 

---------------
Also note the double prefix of `Constants.DatabaseSchema.TableNamePrefix`/`umbraco` and `cms`:
https://github.com/umbraco/Umbraco-CMS/blob/7dc03256edf1bbe15b5b44c5d44a033369ce0e07/src/Umbraco.Core/Migrations/Upgrade/V_8_15_0/AddCmsContentNuByteColumn.cs#L42

This is probably because of the comment on the original `cmsContentNu` table constant:
https://github.com/umbraco/Umbraco-CMS/blob/40dbfa7b8846c2835ddce08cdf690a9dc8719a94/src/Umbraco.Core/Persistence/Constants-DatabaseSchema.cs#L17

We can also fix this up, but it shouldn't matter, as it's a temporary table anyway 😉

PR #8379, opened 2020-07-03
WIP: Map property keys in nucache

nzdev:
@Shazwazza here's what I was thinking for mapping the property keys to shorter values back in may.


nzdev:
slower.

PR #8381, opened 2020-07-03
V8/feature/block list more solid

nielslyngsoe:
Making Block List code more solid.

Improvements include
- Actually has a scopeOfExistance for infinite editing layers that's different than the property-editor scope.
- Destroy cleans up even more.
- Watchers only react to changes.
- The unit test improved a little bit to capture a new scenario.


warrenbuckley:
All good & discussed with @nielslyngsoe over zoom

PR #8382, opened 2020-07-03
V8/feature/block list data structure v3

nielslyngsoe:
https://github.com/umbraco/Umbraco-CMS/pull/8339 ~**Must be reviewed + merged first!**~

Changed the code data structure of Block Editors to fit with data structure proposal v.3, see https://github.com/umbraco/rfcs/pull/23

This PR changes the data structure for both JS and c#. 

* ~Ensure all c# and JS tests pass~ ✔️ 

This also:

* Fixes/streamlines how the block editor data is serialized/deserialized
* Implements the ToEditor/FromEditor logic to ensure that these methods are called for all child controls
* Fixes some other minor issues found when testing 

## Testing

* If you already have a block editor and there's data in it, you won't get any errors, you're data will basically just be cleared because it will detect an invalid structure
* Ensure the block editor works as expected
* Ensure the multi text line editor works in the block editor

### Test case 1

This code should also deal with this [AB#7185](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7185):

> If a user removes the a Block Type from Property Editor Configuration(Prevalues) I would like all the data based on that Block Type to be removed.
Be aware about that Block Editors can be nested inside each other.
> I do not think we need to re-publish data, as the item aren't rendered if configuration doesn't exists.
(Claus: correct - if we want to republish for any reason - we really need to consider unpublished data not getting published)

### Test case 2:

This code should also deal with this [AB#7187](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7187):

> If an element type doesn't exist anymore we should not render the content of a block in the front-end(Website front-end). Same goes for content that does not match a block configuration from that property-editor.


Shazwazza:
FYI I've merged my validation changes into this branch and will now update the c# parts of the data format directly on this branch too.

src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
```diff
@@ -532,6 +542,7 @@
 
                     delete this.layout;
                     delete this.data;
+                    delete this.settingsData;
```
Shazwazza, line 545: Is this correct it doesn't seem right? I'm trying to debug some issues with these changes since it's not submitting data properly. We have this.data.settingsData and this.data.contentData, yet this is deleting this.settingsData and this.content but I don't even thing this.content exists? hrm, there's also a few other references to this.content. I'll keep debugging but i think some things in this code are missing from moving to the new structure  correctly.

warrenbuckley:
Code reads fine to me - off to test & verify :)

warrenbuckley:
@Shazwazza is there a preferred/recommended way to write/iterate over blocks in Razor?
As both approaches work for me fine `@Model.Widgets.ContentData` and `@Model.Widgets.Layout`

```cshtml
@inherits Umbraco.Web.Mvc.UmbracoViewPage<ContentModels.Home>
@using ContentModels = Umbraco.Web.PublishedModels;
@{
    Layout = null;
}

<h4>Normal Repeatable Textstring</h4>
<ul>
    @foreach (var item in Model.Features)
    {
        <li>@item</li>
    }
</ul>


@foreach (var item in Model.Widgets.ContentData)
{
    var blockElItem = (ContentModels.Product)item;
    <h3>Product Name: @blockElItem.ProductName</h3>
    <ul>
        @foreach (var feature in blockElItem.Features)
        {
            <li>@feature</li>
        }
    </ul>
}

@foreach (var block in Model.Widgets.Layout)
{
    var blockEl = (ContentModels.Product)block.Content;
    <h3>Product Name: @blockEl.ProductName</h3>
    <ul>
        @foreach (var feature in blockEl.Features)
        {
            <li>@feature</li>
        }
    </ul>
}
```

warrenbuckley:
Grr - pushed the wrong button I only wanted to make a comment & not close the PR :S

warrenbuckley:
- [x] Tests pass
- [x] Code reads fine
- [x] Test with repeatable textstring & verifying its own PropertyEditor.FromEditor calls are made etc 
- [ ] Test with existing block data before PR - ensure data is cleared
- [ ] Remove block editor from config - removes data
- [ ] Remove block editor from config - no longer renders in razor frontend

Shazwazza:
Thanks @warrenbuckley !

> Data is not removed but we are still shown in with an unsupported block message.

I think this is expected. I didn't change the code to remove them entirely because I saw we had code in there to inform the user visually about what happens. Happy to make it 'remove' the data before it goes to the editor but maybe we can get @nielslyngsoe opinion on that one?

I'll fixup the other issue where the data ends up on the front-end, i seem to have forgot to update the property value converter for this.

> is there a preferred/recommended way to write/iterate over blocks in Razor?
As both approaches work for me fine @Model.Widgets.ContentData and @Model.Widgets.Layout

I think we'll probably want to change the c# razor model a bit since before @nielslyngsoe went on holidays he mentioned this to me but we haven't made any changes to this. I think we should chat about this early this week - perhaps both of you could have a quick call today to see what the best solution is? and then i can make the changes wed or thur?

Shazwazza:
@warrenbuckley I've pushed changes for this now but also see my comments above. I 'think' that your razor was outputing stale data because maybe you were iterating over the Content blocks directly instead of the layout since I think the layout items were being filtered. In any case, i've pushed code to filter out the content/settings blocks directly too and added a test.

src/Umbraco.Web/PropertyEditors/BlockListConfiguration.cs
```diff
@@ -27,11 +27,12 @@ public class BlockConfiguration
             [JsonProperty("thumbnail")]
             public string Thumbnail { get; set; }
 
+            // TODO: This is named inconsistently in JS but renaming it needs to be done in quite a lot of places, this should be contentElementTypeKey
```
nielslyngsoe, line 30: Let me understand this correctly. Is this inconsistent in the new Block Editor code, or is it inconsistent in general in the JS?

I think naming it explicitly with Content, makes it clear that it's for the Content part of the Block, not the Block itself. What I mean is that a Block specifically is not based on an Element-Type directly, the Element-Type is used for the Content or the Settings part of the Block. :-). This will be more clear when we start having Block data in the DB, then the Block itself is then just a reference/usage of an Element-Type from the DB. (If I understood our approach correctly)

Shazwazza: @nielslyngsoe it's because it's inconsistent with the naming we've just added for `settingsElementTypeKey`. In c# in this model we have: `ContentElementTypeKey` and `SettingsElementTypeKey` ... these are consistently named. The corresponding serialized JS names are: `contentTypeKey` and `settingsElementTypeKey` these are not consistently named, they should be: `contentElementTypeKey` and `settingsElementTypeKey`. One is for "content" one is for "settings" and both are based off of Element Types.

nielslyngsoe: Oh, yes, I agree.  Lets get this sorted. Ive making a task for it.
https://dev.azure.com/umbraco/D-Team%20Tracker/_sprints/taskboard/CMS%20Team/D-Team%20Tracker/Sprint%20142?workitem=7621

PR #8383, opened 2020-07-05
v8: Improvements to accessibility of translations section

RachBreeze:
**Dictionary List Page**
On the dictionary list page it wasn't possible for a screen reader to determine correctly the number of rows in the dictionary list.  By changing from a click event on the table row to hrefs and adding a table caption this has been resolved

Previously colours were used to denote the state of the translation
image

Screen reader only text has been added to the columns to indicate the state of the translation

**Dictionary Create**
A label for has been added to the name label.  And the screen reader user can now tab onto the `<h5>` text box. Giving more indication of where they are in the dictionary flow.
An id has been added to the name textbox allowing the label for to correctly identify the textbox.
Name box now has maxlength field matching match length in db.

**Dictionary Edit**
The id has been added to the text areas allowing the label fors to correctly identify the textarea.


nul800sebastiaan:
Wonderful, great update @RachBreeze - thanks very much! 👍

PR #8387, opened 2020-07-06
Show SVG preview on upload

bjarnef:
### Prerequisites

### Description
When uploading creating a new media e.g. "File" or "Image" and selecting or dragging a SVG file to the file upload a preview is not shown, because the `fileSrc` property only is set for images https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbpropertyfileupload.directive.js#L248, where SVG at the moment by default not is considered as image and for example is creating as "File" media type under a folder, when uploading SVG files to a folder.

The `fileSrc` is however set for existing uploaded SVG files on init:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbpropertyfileupload.directive.js#L121

**Before**

**After**

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbpropertyfileupload.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbpropertyfileupload.directive.js
@@ -116,9 +116,10 @@
                     var f = {
                         fileName: file,
                         isImage: mediaHelper.detectIfImageByExtension(file),
-                        extension: getExtension(file)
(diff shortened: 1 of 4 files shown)


bjarnef:
@nielslyngsoe it would be awesome to show a preview for SVG files here like for images.

nul800sebastiaan:
Wonderful @bjarnef - so much better! 👍

PR #8390, opened 2020-07-06
Register dragover state of file upload

bjarnef:
### Prerequisites

### Description
At the moment when dragging file over dropzone in e.g. image cropper a `drag-over` class is added and return feedback to the user. However when dragging a file over a single file upload field on a File media type the user get no feedback on drag over.

I was a bit inspired from this article https://css-tricks.com/drag-and-drop-file-uploading/ and have for a while noticed the file upload in Umbraco didn't provide any feeback to user on dragover.

This is how the dragover change look for the image cropper dropzone:

This PR add a `drag-over` class here as well, which style the upload button with a border color as in the image cropper dropzone.

**Before**

**After**


bjarnef:
@nielslyngsoe it would be great with some feedback to user here on dragover 👍

nul800sebastiaan:
Great stuff @bjarnef - happy to merge! 👍

PR #8394, opened 2020-07-07
Netcore: More tests migrated..

bergmania:
## Details

- Migrated `AuditServiceTests`
    - Added `AuditEntryBuilder`
- Made an option to now boot umbraco core, in the integration tests.. (Testserver tests has its own boot stuff, but resues the db stuff)
- Moved the `[UmbracoTest(Database = UmbracoTestOptions.Database.NewSchemaPerTest)]` into the base class of the testserver tests
 
- Bug fix found by the test of TemplateQueryController. 
   - Output was not camelcased. 
   - Fix in AngularJsonOnlyConfigurationAttribute.cs


src/Umbraco.Web.Common/Formatters/AngularJsonMediaTypeFormatter.cs
```diff
@@ -21,8 +21,7 @@ public class AngularJsonMediaTypeFormatter : NewtonsoftJsonOutputFormatter
         public AngularJsonMediaTypeFormatter(JsonSerializerSettings serializerSettings, ArrayPool<char> charPool, MvcOptions mvcOptions)
             : base(serializerSettings, charPool, mvcOptions)
         {
-            serializerSettings.Converters.Add(new VersionConverter());
```
Shazwazza, line 24: I assume this change is for your note: 

> Bug fix found by the test of TemplateQueryController.
> Output was not camelcased.
> Fix in AngularJsonOnlyConfigurationAttribute.cs

But... does this now mean everything is now camel cased by default? If so, we cannot have that because angular expects many things to not be auto-camel cased. Also what about the VersionConverter?

bergmania: This change bacially just forces the `serializerSettings` to be injected..

So the `VersionConverter` is still used by `AngularJsonOnlyConfigurationAttribute` and camelcase is only used by `JsonCamelCaseFormatterAttribute`

Shazwazza:
I have merged the netcore/netcore branch into this one but the build is failing with one test, not sure if it's part of this PR or not, any ideas?

bergmania:
I know all tests passed before :) but I can take a look

PR #8396, opened 2020-07-07
NuGet package specifications for SQL CE

elit0451:
## Test
- In Powershell execute `$uenv=build/build.ps1 -get ` at the root folder to get the environment variable and download all the necessary tools
  - OBS: this needs to be done just the first time, after that, you can just use `$uenv=build/build.ps1 -get -loc` so you stop downloading over and over again and assume everything is ready (_build.temp_ folder will be deleted every time you run this command)
- Then `$uenv.CompileUmbraco()` will create a **bin** and **WebApp** folder into _build.temp_ with their contents
- Run `$uenv.CompileTests()` to get the _Umbraco.Persistance.SqlCe.*_ files into **WebApp\bin** folder 
- Next command is `$uenv.PreparePackages()` 
- Finally, `$uenv.PackageNuGet()` will generate the 4 NuGet packages (one of them being **UmbracoCms.SqlCe**) in **build.out** folder

---
❗ Make sure that you create XML file for the _Umbraco.Persistance.SqlCe_ project (it is copied in the NuGet package)

### DIFF
--- a/build/NuSpecs/UmbracoCms.SqlCe.nuspec
+++ b/build/NuSpecs/UmbracoCms.SqlCe.nuspec
@@ -3,7 +3,7 @@
     <metadata minClientVersion="4.1.0">
         <id>UmbracoCms.SqlCe</id>
         <version>9.0.0</version>
-        <title>Umbraco Cms Core Binaries</title>
+        <title>Umbraco Cms Sql Ce Add-on</title>
         <authors>Umbraco HQ</authors>
         <owners>Umbraco HQ</owners>
         <license type="expression">MIT</license>
@@ -17,13 +17,15 @@
         <repository type="git" url="https://github.com/umbraco/umbraco-cms" />
         <dependencies>
 
-            <group targetFramework="net472">
+            <group targetFramework="netstandard2.0">
                 <!--
                   note: dependencies are specified as [x.y.z,x.999999) eg [2.1.0,2.999999) and NOT [2.1.0,3.0.0) because
                   the latter would pick anything below 3.0.0 and that includes prereleases such as 3.0.0-alpha, and we do
                   not want this to happen as the alpha of the next major is, really, the next major already.
                 -->
                 <dependency id="UmbracoCms.Core" version="[$version$]" />
+            </group>
+            <group targetFramework="net472">
                 <dependency id="Umbraco.SqlServerCE" version="[4.0.0.1,4.999999)" />
             </group>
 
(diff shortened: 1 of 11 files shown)


build/NuSpecs/UmbracoCms.SqlCe.nuspec
```diff
@@ -0,0 +1,42 @@
...
+        <repository type="git" url="https://github.com/umbraco/umbraco-cms" />
+        <dependencies>
+
+            <group targetFramework="net472">
+                <!--
+                  note: dependencies are specified as [x.y.z,x.999999) eg [2.1.0,2.999999) and NOT [2.1.0,3.0.0) because
+                  the latter would pick anything below 3.0.0 and that includes prereleases such as 3.0.0-alpha, and we do
+                  not want this to happen as the alpha of the next major is, really, the next major already.
+                -->
+                <dependency id="UmbracoCms.Core" version="[$version$]" />
```
bergmania, line 26: I guess this reference has to be netstandard, as `UmbracoCms.Core` will be netstandard 2.0

build/NuSpecs/UmbracoCms.SqlCe.nuspec
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata minClientVersion="4.1.0">
+        <id>UmbracoCms.SqlCe</id>
+        <version>9.0.0</version>
+        <title>Umbraco Cms Core Binaries</title>
```
bergmania, line 6: The Title needs to incode something about SqlCE

elit0451: Thought about it but I saw that on the other nuspec files that we had previously we kept the same title, but can definitely change that!

build/NuSpecs/UmbracoCms.SqlCe.nuspec
elit0451: You are right! Will refactor, thanks 👍

build/NuSpecs/UmbracoCms.SqlCe.nuspec
bergmania: I think it is important to change this one, because this is an add-on to the rest of Umbraco. Not a necessary part.

elit0451: Makes sense!

Issue #8392, opened 2020-07-07
Scheduled content leads to SQL error when rebuilding cache 


mcl-sz:
When the content cache needs to rebuild and content contains items that scheduled, an error occures:

2020-07-07 11:13:25.512 +02:00 [Error] Exception ("0043983c").
System.Data.SqlClient.SqlException (0x80131904): Lock request time out period exceeded.
The statement has been terminated.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString, Boolean isInternal, Boolean forDescribeParameterEncryption, Boolean shouldCacheForAlwaysEncrypted)
   at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, Boolean inRetry, SqlDataReader ds, Boolean describeParameterEncryptionRequest)
   at System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method, TaskCompletionSource`1 completion, Int32 timeout, Task& task, Boolean& usedCache, Boolean asyncWrite, Boolean inRetry)
   at System.Data.SqlClient.SqlCommand.InternalExecuteNonQuery(TaskCompletionSource`1 completion, String methodName, Boolean sendToPipe, Int32 timeout, Boolean& usedCache, Boolean asyncWrite, Boolean inRetry)
   at System.Data.SqlClient.SqlCommand.ExecuteNonQuery()
[...]

Issue #8215, opened 2020-06-01
8.6.1 - Azure Web App - Indexes disappear completely


NikRimington:
Hey Everyone,

So we have a multi-lingual site running in an Azure Web App and every so often (no pattern found yet) all the indexes completely disappear. By disappear, I mean that the examine dashboard errors as none of the indexes exist on disk at all. As various sections of the front end are powered by these indexes they result in errors being thrown.

At this point in time I'm looking for advise on things to look for in order to identify what is going on as I've never seen behaviour like this and there is nothing in the logs to indicate it has happened except errors when the front end tries to access the indexes?

## Umbraco version

8.6.1

Reproduction
------------

Unable to reproduce at this time

### Specifics

I've verified that all the correct settings are in place for Azure web apps based on the following docs:
https://our.umbraco.com/documentation/Getting-Started/Setup/Server-Setup/azure-web-apps#recommended-configuration

![image](https://user-images.githubusercontent.com/13313745/83389608-63c7cc80-a3e8-11ea-8bf6-99d395d5d856.png)

![image](https://user-images.githubusercontent.com/13313745/83389642-73dfac00-a3e8-11ea-820e-f7f67b2a2b92.png)

Shazwazza:
Are you load balancing? or slot swapping (which will act like load balancing if pointed to the same database). There's an issue with SqlMainDomLock for LB which you could be affected by which can cause all sorts of strange behavior. Its fixed in the upcoming 8.6.2. Perhaps you've seen/read this gigantic thread here? https://github.com/umbraco/Umbraco-CMS/issues/8006

NikRimington:
Hey Shannon :-) 

Nope, not load balanced or slot swapping (Useful to know that slot swapping behaves like load balancing though as are looking to move to that in the future).

We are using Azure Pipelines to deploy straight to the live environment at the minute.
When the issue occurs (which is twice in the past 2 weeks so far), we fix it by touching web.config to trigger a restart.

I did just check the umbracoServer table though and there are a lot of entries in there but they are all (as far as I can tell) for the same slot, so I'm guessing the changing names are when Azure moves things around.

![image](https://user-images.githubusercontent.com/13313745/83393055-1189aa00-a3ee-11ea-95a1-f84bbb7fb53a.png)

Shazwazza:
[...]

PR #8398, opened 2020-07-08
Adjustments for SqlMainDomLock and others to make azure operations more resilient

Shazwazza:
Details are here: https://github.com/umbraco/Umbraco-CMS/issues/8215

This should also resolve https://github.com/umbraco/Umbraco-CMS/issues/8392

* Improvements to SqlMainDomLock
  * Better management of db instances and transactions - previously this was trying to manage a single database instances which isn't what should be done, db instances should exist for a short period of time and disposed of. Changed to more explicit code for managing transactions
  * Increase polling times - though according to logs this wasn't causing problems but is still too aggressive for what it's doing
  * Don't stop listening (shut down maindom) if there are SQL exceptions, only if the appdomain has actually triggered a shutdown. We cannot just stop listening even if there are errors since it will lead to premature MainDom shutdown when the app is not actually being shutdown. This means it will just retry on the next poll. 
* Investigation into scheduled publishing lock request timeouts (also reported here #8392) which occurs around the same time as when SqlMainDomLock fails
  * The PerformScheduledPublishInternal was using yield returns with a wrapped Scope/IDisposable which is a bit hard to follow along. One issue is that we were yield returning when the Saving event was canceled for an individual item but not continuing so the logic would have proceeded anyways for that item even if that event was canceled and found other similar issues where it would actually break out of the entire process instead of skip to the next record
  * I believe one of the main issues seen with the lock request timeouts like those in #8392 and in the logs for #8215 is because the PerformScheduledPublishInternal takes a WriteLock at the very beginning of the method. This is the write lock that times out according to the stack traces. This is probably because scheduled publishing runs every minute which is actually a bit crazy to think we lock the whole content tree every minute even if we aren't writing anything. And because of that I think this may also be one of the other main suspects as to why we saw SqlMainDomLock error around the same time we see the sql lock timeout error for scheduled publishing. To fix this:
    * We need a new and much faster method: DocumentRepository.HasContentForRelease instead of only relying on the potentially expensive DocumentRepository.GetContentForRelease, similarly for GetContentForExpiration we should also have HasContentForExpiration. Then we separate the job of PerformScheduledPublishInternal into 2 parts: First, without any read or write locks just check if either HasContentForRelease or GetContentForExpiration  is true, then we can take a write lock and proceed as we are today. This will mean that there isn't a writelock on all content attempted to be made every minute.
* Don't short circuit index rebuilding if one of the populators throws an exception, continue with the other populators
* Brings our sql transient fault handling in-line with the latest specs from MS

### DIFF
--- a/src/Umbraco.Core/Services/Implement/ContentService.cs
+++ b/src/Umbraco.Core/Services/Implement/ContentService.cs
@@ -30,7 +30,7 @@ namespace Umbraco.Core.Services.Implement
         private IQuery<IContent> _queryNotTrashed;
         //TODO: The non-lazy object should be injected
         private readonly Lazy<PropertyValidationService> _propertyValidationService = new Lazy<PropertyValidationService>(() => new PropertyValidationService());
-        
+
 
         #region Constructors
 
@@ -875,7 +875,7 @@ namespace Umbraco.Core.Services.Implement
                     throw new NotSupportedException($"Culture \"{culture}\" is not supported by invariant content types.");
             }
 
-            if(content.Name != null && content.Name.Length > 255)
+            if (content.Name != null && content.Name.Length > 255)
             {
                 throw new InvalidOperationException("Name cannot be more than 255 characters in length.");
             }
@@ -1243,7 +1243,7 @@ namespace Umbraco.Core.Services.Implement
                     if (culturesUnpublishing != null)
                     {
                         // This will mean that that we unpublished a mandatory culture or we unpublished the last culture.
-                        
+
                         var langs = string.Join(", ", allLangs
                                     .Where(x => culturesUnpublishing.InvariantContains(x.IsoCode))
                                     .Select(x => x.CultureName));
@@ -1252,7 +1252,7 @@ namespace Umbraco.Core.Services.Implement
                         if (publishResult == null)
                             throw new PanicException("publishResult == null - should not happen");
 
-                        switch(publishResult.Result)
+                        switch (publishResult.Result)
                         {
                             case PublishResultType.FailedPublishMandatoryCultureMissing:
                                 //occurs when a mandatory culture was unpublished (which means we tried publishing the document without a mandatory culture)
@@ -1266,7 +1266,7 @@ namespace Umbraco.Core.Services.Implement
                                 Audit(AuditType.Unpublish, userId, content.Id, "Unpublished (last language unpublished)");
                                 return new PublishResult(PublishResultType.SuccessUnpublishLastCulture, evtMsgs, content);
                         }
-                        
+
                     }
 
                     Audit(AuditType.Unpublish, userId, content.Id);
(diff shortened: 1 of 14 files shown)


src/Umbraco.Core/Services/Implement/ContentService.cs
```diff
@@ -1387,11 +1453,14 @@ private IEnumerable<PublishResult> PerformScheduledPublishInternal(DateTime date
                             .ToList();
 
                         if (pendingCultures.Count == 0)
-                            break; //shouldn't happen but no point in continuing if there's nothing there
+                            continue; //shouldn't happen but no point in continuing if there's nothing there
```
clausjensen, line 1456: maybe update/remove these comments .. someone will be confused in the future if it says no point in continuing and we then `continue` 😃 ..

src/Umbraco.Core/Services/Implement/ContentService.cs
```diff
@@ -1403,94 +1472,51 @@ private IEnumerable<PublishResult> PerformScheduledPublishInternal(DateTime date
...
-                            var impact = CultureImpact.Explicit(culture, IsDefaultCulture(allLangs, culture));
+                            var impact = CultureImpact.Explicit(culture, IsDefaultCulture(allLangs.Value, culture));
                             var tryPublish = d.PublishCulture(impact) && _propertyValidationService.Value.IsPropertyDataValid(d, out invalidProperties, impact);
                             if (invalidProperties != null && invalidProperties.Length > 0)
                                 Logger.Warn<ContentService>("Scheduled publishing will fail for document {DocumentId} and culture {Culture} because of invalid properties {InvalidProperties}",
                                     d.Id, culture, string.Join(",", invalidProperties.Select(x => x.Alias)));
 
                             publishing &= tryPublish; //set the culture to be published
-                            if (!publishing) break; // no point continuing
+                            if (!publishing) continue; // no point continuing
```
clausjensen, line 1482: same ^^

src/Umbraco.Web/Compose/DatabaseServerRegistrarAndMessengerComponent.cs
```diff
@@ -218,7 +218,7 @@ public override bool PerformRun()
                 }
                 catch (Exception e)
                 {
-                    _logger.Error<InstructionProcessTask>("Failed (will repeat).", e);
+                    _logger.Error<InstructionProcessTask>(e, "Failed (will repeat).");
```
clausjensen, line 221: nice catch :)

src/Umbraco.Web.UI/Umbraco.Web.UI.csproj
```diff
@@ -12,7 +12,8 @@
     <ProjectTypeGuids>{349c5851-65df-11da-9384-00065b846f21};{fae04ec0-301f-11d3-bf4b-00c04f79efbc}</ProjectTypeGuids>
     <RunPostBuildEvent>OnBuildSuccess</RunPostBuildEvent>
     <UseIISExpress>true</UseIISExpress>
-    <IISExpressSSLPort>44331</IISExpressSSLPort>
```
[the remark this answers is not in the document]
Shazwazza: Nope not intentional, Visual Studio or something in our git repo constantly changes this everytime I build.

PR #8400, opened 2020-07-08
WIP Option C - Compress custom properties for nucache to reduce memory consumption

nzdev:
### Prerequisites

RE: https://github.com/umbraco/Umbraco-CMS/pull/8365

### Description
What?
Compress custom properties at cmsContentNu table level or nucache.db level.
Decompress properties when reading the property from nucache.db or lazily when the property is read / converted.
Map property alias to a shorter key in nucache

Why?
Reduce memory consumption by keeping large string / rich text properties compressed till read.

TODO: Update propertyconverterbase logic for IsValue and HasValue as these check for strings which may not work with lazy decompression.
TODO: Configuration per property editor / per property.

How to test:
Enable MessagePack Nucache Serializer
`<add key="Umbraco.Web.PublishedCache.NuCache.Serializer" value="MsgPack"/>
`
Enable LZ4 DictionaryOfPropertiesSerializer
`<add key="Umbraco.Web.PublishedCache.NuCache.DictionaryOfPropertiesSerializer" value="LZ4Map"/>`

Configure Map
Format:
`<add key="Umbraco.Web.PublishedCache.NuCache.PropertySerializationMap" value="{propertyAlias},{NucachePropertyCompressionLevel},{NucachePropertyDecompressionLevel},{shorterPropertyAlias};{propertyAlias2},{NucachePropertyCompressionLevel2},{NucachePropertyDecompressionLevel2}{shorterPropertyAlias2};"/>`

Example:
Compress richText1 in cmsContentNu table, Lazy decompression when property read, shorten property key to rte1.
Compress textString2 in nuCache.db, Immediately decompress property when deserialized, don't change property key.
Don't compress/decompress umbracoNaviHide, shorten property key to unh.

`<add key="Umbraco.Web.PublishedCache.NuCache.PropertySerializationMap" value="richText1,SQLDatabase,Lazy,rte1;textString2,NucacheDatabase,Immediate,textString2;umbracoNaviHide,None,NotCompressed,unh"/>`

Rebuild nucache

### DIFF
--- a/src/Umbraco.Web/PublishedCache/NuCache/DataSource/AppSettingsNucachePropertyMapFactory.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/DataSource/AppSettingsNucachePropertyMapFactory.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Configuration;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Umbraco.Web.PublishedCache.NuCache.DataSource
+{
+    public class AppSettingsNucachePropertyMapFactory : INucachePropertyOptionsFactory
+    {
+        public NucachePropertyOptions GetNucachePropertyOptions()
+        {
+            NucachePropertyOptions options = new NucachePropertyOptions
+            {
+                PropertyMap = GetPropertyMap(),
+                LZ4CompressionLevel = K4os.Compression.LZ4.LZ4Level.L10_OPT,
+                MinimumCompressibleStringLength = null
+            };
+            return options;
+        }
+
+        public IReadOnlyDictionary<string, (NucachePropertyCompressionLevel compress, NucachePropertyDecompressionLevel decompressionLevel, string mappedAlias)> GetPropertyMap()
+        {
+            var propertyMap = new Dictionary<string, (NucachePropertyCompressionLevel compress, NucachePropertyDecompressionLevel decompressionLevel, string mappedAlias)>();
+            // TODO: Use xml/json/c# to define map
+            var propertyDictionarySerializerMap = ConfigurationManager.AppSettings["Umbraco.Web.PublishedCache.NuCache.PropertySerializationMap"];
+            if (!string.IsNullOrWhiteSpace(propertyDictionarySerializerMap))
+            {
+                //propertyAlias,CompressionLevel,DecompressionLevel,mappedAlias;
+                propertyDictionarySerializerMap.Split(';')
+                    .Select(x =>
+                    {
+                        var y = x.Split(',');
+                        (string alias, NucachePropertyCompressionLevel compressionLevel, NucachePropertyDecompressionLevel decompressionLevel, string mappedAlias) v = (y[0],
+                        (NucachePropertyCompressionLevel)System.Enum.Parse(typeof(NucachePropertyCompressionLevel), y[1]),
+                        (NucachePropertyDecompressionLevel)System.Enum.Parse(typeof(NucachePropertyDecompressionLevel), y[2]),
+                        y[3]
+                        );
+                        return v;
+                    })
+                    .ToList().ForEach(x => propertyMap.Add(x.alias, (x.compressionLevel, x.decompressionLevel, x.mappedAlias)));
+            }
+            return propertyMap;
+        }
+    }
+}
--- a/src/Umbraco.Web/PublishedCache/NuCache/DataSource/AppSettingsNucachePropertyMapFactory.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/DataSource/AppSettingsNucachePropertyMapFactory.cs
@@ -7,9 +7,9 @@ using System.Threading.Tasks;
 
 namespace Umbraco.Web.PublishedCache.NuCache.DataSource
 {
-    public class AppSettingsNucachePropertyMapFactory : INucachePropertyOptionsFactory
+    public class AppSettingsNuCachePropertyMapFactory : INuCachePropertyOptionsFactory
     {
-        public NucachePropertyOptions GetNucachePropertyOptions()
+        public NucachePropertyOptions GetNuCachePropertyOptions()
         {
             NucachePropertyOptions options = new NucachePropertyOptions
             {
--- a/src/Umbraco.Web/PublishedCache/NuCache/DataSource/AppSettingsNucachePropertyMapFactory.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/DataSource/AppSettingsNucachePropertyMapFactory.cs
@@ -9,9 +9,9 @@ namespace Umbraco.Web.PublishedCache.NuCache.DataSource
 {
     public class AppSettingsNuCachePropertyMapFactory : INuCachePropertyOptionsFactory
     {
-        public NucachePropertyOptions GetNuCachePropertyOptions()
+        public NuCachePropertyOptions GetNuCachePropertyOptions()
         {
-            NucachePropertyOptions options = new NucachePropertyOptions
+            NuCachePropertyOptions options = new NuCachePropertyOptions
             {
                 PropertyMap = GetPropertyMap(),
                 LZ4CompressionLevel = K4os.Compression.LZ4.LZ4Level.L10_OPT,
--- a/src/Umbraco.Web/PublishedCache/NuCache/DataSource/INucachePropertyOptionsFactory.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/DataSource/INucachePropertyOptionsFactory.cs
@@ -6,8 +6,8 @@ using System.Threading.Tasks;
 
 namespace Umbraco.Web.PublishedCache.NuCache.DataSource
 {
-    public interface INucachePropertyOptionsFactory
+    public interface INuCachePropertyOptionsFactory
     {
-        NucachePropertyOptions GetNucachePropertyOptions();
+        NucachePropertyOptions GetNuCachePropertyOptions();
     }
 }
--- a/src/Umbraco.Web/PublishedCache/NuCache/DataSource/INucachePropertyOptionsFactory.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/DataSource/INucachePropertyOptionsFactory.cs
@@ -8,6 +8,6 @@ namespace Umbraco.Web.PublishedCache.NuCache.DataSource
 {
     public interface INuCachePropertyOptionsFactory
     {
-        NucachePropertyOptions GetNuCachePropertyOptions();
+        NuCachePropertyOptions GetNuCachePropertyOptions();
     }
 }
--- a/src/Umbraco.Web/PublishedCache/NuCache/DataSource/INucachePropertyOptionsFactory.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/DataSource/INucachePropertyOptionsFactory.cs
@@ -1,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Umbraco.Web.PublishedCache.NuCache.DataSource
+{
+    public interface INucachePropertyOptionsFactory
+    {
+        NucachePropertyOptions GetNucachePropertyOptions();
+    }
+}
--- a/src/Umbraco.Web/PublishedCache/NuCache/DataSource/LazyCompressedString.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/DataSource/LazyCompressedString.cs
@@ -8,15 +8,23 @@ using System.Threading.Tasks;
 
 namespace Umbraco.Web.PublishedCache.NuCache.DataSource
 {
-    public class LazyCompressedString
+    /// <summary>
+    /// Lazily decompresses a LZ4 Pickler compressed UTF8 string
+    /// </summary>
+    internal class LazyCompressedString
     {
         private byte[] _bytes;
         private string _str;
 
+        /// <summary>
+        /// Constructor
+        /// </summary>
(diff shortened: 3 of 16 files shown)


src/Umbraco.Web/PublishedCache/NuCache/DataSource/AppSettingsNucachePropertyMapFactory.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Configuration;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Umbraco.Web.PublishedCache.NuCache.DataSource
+{
+    public class AppSettingsNucachePropertyMapFactory : INucachePropertyOptionsFactory
```
Shazwazza, line 10: This will be ok for testing for now until we progress this further but ultimately a lot of how property values are serialized/deserialized will need to be up to the property editor itself. We can of course have other config bits and pieces

src/Umbraco.Web/PublishedCache/NuCache/DataSource/INucachePropertyOptionsFactory.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Umbraco.Web.PublishedCache.NuCache.DataSource
+{
+    public interface INucachePropertyOptionsFactory
```
Shazwazza, line 9: minor: NuCache instead of Nucache

src/Umbraco.Web/PublishedCache/NuCache/DataSource/LazyCompressedString.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Umbraco.Web.PublishedCache.NuCache.DataSource
+{
+    public class LazyCompressedString
```
Shazwazza, line 11: keep things internal for now that we don't need to make public

src/Umbraco.Web/PublishedCache/NuCache/DataSource/INucachePropertyOptionsFactory.cs
nzdev: Fixed spelling

src/Umbraco.Web/PublishedCache/NuCache/DataSource/AppSettingsNucachePropertyMapFactory.cs
nzdev: Fixed spelling. Not currently sure how to get the UI / property editor to configure, I took this approach as I wasn't sure how this feature would be supported by non-core property editors.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/LazyCompressedString.cs
nzdev: changed to internal

src/Umbraco.Web/PublishedCache/NuCache/DataSource/AppSettingsNucachePropertyMapFactory.cs
Shazwazza: Yep no worries gotta start somewhere. I have a couple hours tonight so gonna merge this in and then can go from there.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/Lz4DictionaryOfPropertyDataSerializer.cs
```diff
@@ -0,0 +1,177 @@
...
+                foreach (var pdata in values)
+                {
+                    // everything that can be null is read/written as object
+                    //  even though - culture and segment should never be null here,
+                    //  see note in ReadFrom() method above
+                    WriteObject(pdata.Culture ?? string.Empty, stream);
+                    WriteObject(pdata.Segment ?? string.Empty, stream);
+
+                    //Only compress strings
+                    if (pdata.Value is string stringValue && pdata.Value != null && map.Compress.Equals(NucachePropertyCompressionLevel.NucacheDatabase)
```
Shazwazza, line 135: @nzdev I've merged this in and just cleaning up a few things to understand this all a little more. I've noticed that in this method we are only checking for this type NucachePropertyCompressionLevel.NucacheDatabase and not also NucachePropertyCompressionLevel.SQLDatabase but we are checking for both of those in the above ReadFrom method. Just wondering if this is on purpose or not?

src/Umbraco.Web/PublishedCache/NuCache/DataSource/Lz4DictionaryOfPropertyDataSerializer.cs
```diff
@@ -0,0 +1,177 @@
...
+                        && (_nucachePropertyOptions.MinimumCompressibleStringLength == null
+                        || !_nucachePropertyOptions.MinimumCompressibleStringLength.HasValue
+                        || stringValue.Length > _nucachePropertyOptions.MinimumCompressibleStringLength.Value))
+                    {
+                        var stringBytes = Encoding.UTF8.GetBytes(stringValue);
+                        var compressedBytes = LZ4Pickler.Pickle(stringBytes, _nucachePropertyOptions.LZ4CompressionLevel);
+                        WriteObject(compressedBytes, stream);
+                    }
+                    WriteObject(pdata.Value, stream);
```
Shazwazza, line 144: Was this supposed to be an else?

src/Umbraco.Web/PublishedCache/NuCache/DataSource/MsgPackContentNestedDataSerializer.cs
```diff
@@ -48,10 +53,42 @@ public ContentNestedData Deserialize(string data)
...
+                            foreach (var property in properties.Where(x => x.Value != null && x.Value is string))
+                            {
+                                property.Value = LZ4Pickler.Pickle(Encoding.UTF8.GetBytes(property.Value as string), _propertyOptions.LZ4CompressionLevel);
+                            }
+                        }
+                    }
+                    if (map.Value.mappedAlias != null && !map.Key.Equals(map.Value.mappedAlias)
+                        && nestedData.PropertyData.Remove(map.Key) && nestedData.PropertyData.TryGetValue(map.Key, out PropertyData[] properties2))
+                    {
+                        nestedData.PropertyData.Remove(map.Key);
```
Shazwazza, line 85: This 2nd Remove call isn't necessary I don't think since it's already being removed in the if statement?

src/Umbraco.Web/PublishedCache/NuCache/DataSource/Lz4DictionaryOfPropertyDataSerializer.cs
nzdev: It's been a while, but the idea is potentially compressing twice. At the property level (NucachePropertyCompressionLevel.NucacheDatabase) as well as compressing the entire json (NucachePropertyCompressionLevel.SQLDatabase ). Compressing at the property level as well means the strings in bplustree are compressed.

nzdev: When reading back, we can decide if we want to decompress the properties or not before they go into bplustree.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/Lz4DictionaryOfPropertyDataSerializer.cs
nzdev: yes

src/Umbraco.Web/PublishedCache/NuCache/DataSource/Lz4DictionaryOfPropertyDataSerializer.cs
Shazwazza: Ok cool, yep also wanted to check on what these meant as I'm adding some code comments:

* NucachePropertyCompressionLevel.NucacheDatabase = compression for properties at the nucache level (in memory + nucache.db)
* NucachePropertyCompressionLevel.SQLDatabase = compression for the entire document at the nucache DB level

I 'think' however, that on this line https://github.com/umbraco/Umbraco-CMS/pull/8400/files#diff-48a0f486bb419f4223394accbe88f205R85 we shouldn't be checking for NucachePropertyCompressionLevel.SQLDatabase? To me it seems like NucachePropertyCompressionLevel.SQLDatabase isn't really used quite yet since that is pretty much the only usage of it in the code. I could very well be missing something though :)

nzdev: Ah, sorry. It's where it's decompressed that makes a difference. NucachePropertyCompressionLevel has no effect on compressing the whole document.
NucachePropertyCompressionLevel.NucacheDatabase = compress the property before it's written to the database, store the compressed value in bplustree, lazy decompress the string on .ToString().

NucachePropertyCompressionLevel.SQLDatabase compress the property before it's written to the database, decompress when reading the document from the database, store the uncompressed value in bplustree. It's late in NZ, but I think you would only do this if you were not compressing the whole document.

Shazwazza: Ah, I think I understand a little more here, in the MsgPackContentNestedDataSerializer.Optimize we check for NucachePropertyCompressionLevel.SQLDatabase and we are compressing the property data if we aren't using the byte serialization. 

I'm just not sure there's any benefit in having that option since it seems storing the document as bytes with MessagePack is much better?

nzdev: NucachePropertyCompressionLevel.SQLDatabase is may be worth dropping. I think the thought at the time was it may be better to do all the decompression at boot.

Shazwazza: Cool just wanted to make sure i got it and makes sense now. Still cleaning up a few things, will post some revision changes when I'm done

src/Umbraco.Web/PublishedCache/NuCache/DataSource/MsgPackContentNestedDataSerializer.cs
nzdev: I think this alias mapping wasn't working correctly. I remember having to set the map to map to the same name in order to git it to work last time. This may be why.

nzdev: It saves a few bytes, but it might be worth removing this if block for now.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/MsgPackContentNestedDataSerializer.cs
```diff
@@ -48,10 +53,42 @@ public ContentNestedData Deserialize(string data)
...
+        /// Compress properties and map property names to shorter names
+        /// </summary>
+        /// <param name="nestedData"></param>
+        private void Optimize(ContentNestedData nestedData)
+        {
+            if (_propertyOptions.PropertyMap != null && _propertyOptions.PropertyMap.Any())
+            {
+                foreach (var map in _propertyOptions.PropertyMap)
+                {
+                    if (map.Value.compress.Equals(NucachePropertyCompressionLevel.SQLDatabase))
```
nzdev, line 72: maybe this should also do the nucache option as well

nzdev: My comments on this have been way off. There are two enums.
    public enum NucachePropertyCompressionLevel
    {
        None = 0, //No property compression
        SQLDatabase = 1, //Compress the property before being written to the database
        NuCacheDatabase = 2 // Compress the property when writing to nucache bplustree after reading from the database
    }
    /// <summary>
    /// If/where to decompress custom properties for nucache
    /// </summary>
    public enum NucachePropertyDecompressionLevel
    {
        NotCompressed = 0, // No decompression required
        Immediate = 1, //Decompress the property after it's read from the database (at boot or rebuild)
        Lazy = 2 //Decompress the property when .ToString() is called.
    }

src/Umbraco.Web/PublishedCache/NuCache/DataSource/MsgPackContentNestedDataSerializer.cs
Shazwazza: Yep have added notes, i'll test to see how it goes.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/MsgPackContentNestedDataSerializer.cs
Shazwazza: Yep all good, have figured it out :) I've removed the SQLDatabase  option for now along with the 'MsgPackContentNestedDataSerializer.Optimize'  since I don't really think anyone would choose to not use the binary storage format since it's much faster and smaller. We can add back later if we think it's important.

nzdev: The SQLDatabase option is still binary, and my preferred option for my scenario. The goal with it was to save on compression and memory on boot as it does not need to hold both the string value and the pickled byte[] at boot. I updated further explanation as a comment on the main thread instead of here (my apologies, it's midnight here).

PR #8403, opened 2020-07-08
8192 with integration test

ismailmayat:
### Prerequisites

Added working integration test for recent PR for issue 8192


ismailmayat:
want to refactor test code

nul800sebastiaan:
Thanks Ismail! I've updated the test a little bit to be able to create many users, I wanted to check that the performance impact of hydrating the user objects wasn't super horrible. Luckily it's negligible! 

Thanks again, unit tests are wonderful! 👍 

Next time you're not done yet with a PR, you can create it as a Draft as well, so @poornimanayar doesn't swoop in too quickly! 😉

ismailmayat:
@nul800sebastiaan noted about draft, @poornimanayar gave it special attention as it was my first v8 pr hopefully first of many.  Regarding performance testing is there something in the code base that gives you those metrics or are you using third party tool?

nul800sebastiaan:
Cool and congrats on that first one! 👍 

We do have perf tests in `Umbraco.Tests.Benchmarks`. But this was just a very simple manual check for the moment, what happens if I create a 100 users before your hydration fix, and what happens if I create a 100 after your fix. Tried that 5 times and the timings were pretty much the same!

Issue #8405, opened 2020-07-09
Nested Content - Add content dialog doesn't render html (icon)


stefankip:
Using Umbraco v8.6.3

When one adds content to a Nested Content property and have multiple content types to add, or have content on the clipboard, the dialog for selecting the content type is shown.
This dialog has a heading from the language files and the default translation contains HTML to show an icon.
This HTML and thus icon isn't rendered as you can see here:

![image](https://user-images.githubusercontent.com/46313552/87015083-392e1800-c1cd-11ea-8717-83ca6fd388b7.png)

bjarnef:
Do you use any third party packages? I have looked at latest changes in `v8/contrib` branch, where I can't reproduce this issue. Also in `v8/8.6` branch, it seems there shouldn't be this text above "Paste from clipboard".
https://github.com/umbraco/Umbraco-CMS/blob/v8/8.6/src/Umbraco.Web.UI.Client/src/views/common/overlays/itempicker/itempicker.html#L14-L19

stefankip:
Hmm it seems to be this line: https://github.com/umbraco/Umbraco-CMS/blob/7634b376769b1965f5425129f3f3865bab23d301/src/Umbraco.Web.UI.Client/src/views/components/overlays/umb-overlay.html#L4
I'm using the Perplex ContentBlocks package and I'm adding content inside a NC inside one of those blocks.

bjarnef:
I tried with a clean Umbraco v8.6.3 with default starter kit.

In English:

![image](https://user-images.githubusercontent.com/2919859/87041547-ac965080-c1f2-11ea-861b-0a9dbd547b45.png)

In Dutch:
![image](https://user-images.githubusercontent.com/2919859/87041671-dfd8df80-c1f2-11ea-98ad-67c814d20db5.png)

![image](https://user-images.githubusercontent.com/2919859/87041704-ea937480-c1f2-11ea-94da-56f9fc087291.png)

In seems the Dutch translation includes icons, which it shouldn't.

stefankip:
But it looks pretty neat, so maybe we can add the icon to other languages as well and support that? 😎

bjarnef:
Looking at the "addElement" key in also seems to be an issue with another language:
https://github.com/umbraco/Umbraco-CMS/blob/4dc5b02b89fdfa019b30eaaa6ff41f65010611ab/src/Umbraco.Web.UI/Umbraco/config/lang/tr.xml#L799

[...]

PR #8408, opened 2020-07-09
Fix translations files including icon HTML

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8405

### Description
A few translation files included icon HTML for a translation key using for Nested Content add button and also in overlay title. No other languages than `nl` and `tr` did include this.

In general I think HTML should be avoided when possible in translations, since we just want the raw translation and when people contribute with translations they often only make changes to a few languages - not all of them.

This should probably be cherry picked in `v8/8.6`


nul800sebastiaan:
Great, thank you! 👍

PR #8410, opened 2020-07-09
Allow editor info to shrink for longer datatype names

bjarnef:
### Prerequisites

### Description
In property settings overlay when datatype has a longer name the editor info need to shrink, so it doesn't push the remove button out of screen real estate.

**Before**

image

After

image


nul800sebastiaan:
Ha! Nice one Bjarne, cheers 🍻

PR #8411, opened 2020-07-09
Align group builder property actions to top

bjarnef:
### Prerequisites

### Description
This PR align the property actions to top in the group builder instead of centered. In might work okay with these centered, but if adding a longer description, it push down these actions.
It seems nicer to have these at same position next to the property editor preview.

**Before**

image

**After**

image


nul800sebastiaan:
Love it! Thanks again!

Issue #8412, opened 2020-07-10
Tag length is limited in the database but not in the ui


agrath:
When using the built-in tag picker, the tag database table is limited to 200 characters. Shorter tags work fine.

![image](https://user-images.githubusercontent.com/78284/87112765-c9995500-c2c0-11ea-8eac-60dc024559c8.png)

## Umbraco version

I am seeing this issue on Umbraco version: 8.7.0 (latest v8/contrib)

Reproduction
------------

Enter a long tag into a tag picker,
I use a password generator of more than 200 characters and you will get the following exception:

![image](https://user-images.githubusercontent.com/78284/87112825-efbef500-c2c0-11ea-8fde-f4a654be7260.png)

Rest of template truncated, I'm going to push a pull request shortly to set the maxlength on the tag input field

PR #8413, opened 2020-07-10
#8412 Set ng-maxlength and maxlength attributes on the tag input 

agrath:
...on umb-tags-editor.html to avoid a string or binary data would be truncated error when the entered tag exceeds the database column length

This fixes https://github.com/umbraco/Umbraco-CMS/issues/8412

### Description

Sets max length on the tag input
To test this, set up a blank site, paste a tag longer than 200 characters in the tag picker
It will be truncated in the UI
Prior to this fix it would throw a server error on save & publish


nul800sebastiaan:
Hey there @agrath - hope you're well! Good fix here, thanks, this is helpful! 👍

PR #8416, opened 2020-07-10
Adjust tooltip position inside overlay

bjarnef:
### Prerequisites

### Description
This PR fix the issue, where tooltip wasn't positioned correct when used inside an overlay, e.g. in itempicker, which is now used a overlay in nested content. The tooltip is at the moment using `#contentwrapper` as container to calculate the position of the tooltip: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/umbtooltip.directive.js#L82

Also note it is now using `getBoundingClientRect()` since `container[0].offsetLeft` and `container[0].offsetTop` return the left and top position, but not the "real" position, when the overlay is using `transform: translate(-50%,-50%);`
https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

Furthermore on mouseover in allow child nodes overlay on a document type, it was throwing an console error, because in called a undefined function `$scope.mouseLeave()`.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbtooltip.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbtooltip.directive.js
@@ -77,13 +77,17 @@ Use this directive to render a tooltip.
          scope.tooltipStyles.left = 0;
          scope.tooltipStyles.top = 0;
 
-         function setTooltipPosition(event) {
+          function setTooltipPosition(event) {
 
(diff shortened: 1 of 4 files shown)


bjarnef:
@kjac maybe you want to have a look at this one? 😎👌

nul800sebastiaan:
Once again, lovely stuff! Thanks @bjarnef 👍  Sorry I didn't see this problem before merging #8044!

PR #8419, opened 2020-07-11
Sort on y-axis in sort dialog

bjarnef:
### Prerequisites

### Description
This PR restrict the sorting to y-axis in sort dialog and within parent element.
Furthermore if remove unnecessary use of `ng-if` and use `ng-class` to toggle asc/desc sort icon instead.

One of the button elements also had a wrong end-closing tag, so fixed that as well.


nul800sebastiaan:
Lovely work, @bjarnef strikes again! 👍

PR #8423, opened 2020-07-11
Listview columns grouping

bjarnef:
### Prerequisites

### Description
This PR use `<optgroup>` to group fields for listview columns in two groups: "System Fields" and "Custom Fields".

Furthermore is remove unnecessary use of `ng-if`.
The following part:

```
<span class="list-view-layout__name-text" ng-if="!val.isSystem" ng-bind="val.alias"></span>
<span class="list-view-layout__name-text" ng-if="val.isSystem == 1" ng-bind="val.alias"></span>
```

can be simplified to:

```
<span class="list-view-layout__name-text" ng-bind="val.alias"></span>
```

image


nul800sebastiaan:
Good old `optgroup`! Thanks very much @bjarnef ! 👍

Issue #8424, opened 2020-07-11
Overlay positioned top left when using hotkey


bjarnef:
In child node selector under document type and media type an overlay is shown at target position on click.

![chrome_2020-07-11_18-30-38](https://user-images.githubusercontent.com/2919859/87229158-21f65280-c3a6-11ea-815d-9e683bf4e619.png)

However when using the hotkey `alt+shift+c` the overlay is positioned in top left corner.

![chrome_2020-07-11_18-30-23](https://user-images.githubusercontent.com/2919859/87229171-405c4e00-c3a6-11ea-9c67-5c0af682b2b8.png)

On click you might also briefly note the overlay "jumps" from top left to target position.

Tested with latest code changes in `v8/contrib` branch.

emmaburstow:
Hey Bjarne! Thanks for reporting. I can reproduce so I've labelled :)

Emma

umbrabot:
Hi @bjarnef,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
PR submitted: https://github.com/umbraco/Umbraco-CMS/pull/8431

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8431

PR #8431, opened 2020-07-13
Adjust target overlay position using hotkey

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8424

### Description
This PR use similar to this other PR https://github.com/umbraco/Umbraco-CMS/pull/8416 `getBoundingClientRect()` which get the "real" position, when e.g. uisng transform styles via css.

In this case `setTargetPosition()` is only using for overlays having position set to `target`.

When using hotkey the `mousePositionClickX` and `mousePositionClickY` was undefined/null, so in this case it fallback to center of container element.

Furthermore the position was calculate based on container `#contentwrapper`, but like `umb-tooltip` in https://github.com/umbraco/Umbraco-CMS/pull/8416, this would probably be positioned wrong if opening an `target` overlay from an existing overlay.


nul800sebastiaan:
I love this, thanks a bunch @bjarnef ! 👍

Issue #8244, opened 2020-06-08
v7.15.5 - Failed to delete item xxxx - Not a writer.


c9mbundy:
Since upgrading to v7.15.5 - I'm noticing a constant error when deleting nodes.
On refreshing the page, it appears that the node has actually been deleted, and lands in the recycle-bin as expected, but I don't know if the error indicates other issues that need to be addressed.
I'm not aware of any changes on my side that could be triggering this, and it was not apparent with v7.15.4
-----------------
Received an error from the server
Failed to delete item xxxx
Not a writer.
----------------
Exception Details:
System.InvalidOperationException: Not a writer.
Stacktrace:
at umbraco.SafeXmlReaderWriter.set_Xml(XmlDocument value)
   at umbraco.content.UpdateDocumentCache(Document d)
   at Umbraco.Web.Cache.UnpublishedPageCacheRefresher.Refresh(String jsonPayload)
   at Umbraco.Core.Sync.ServerMessengerBase.DeliverLocal(ICacheRefresher refresher, MessageType messageType, IEnumerable`1 ids, String json)
   at Umbraco.Core.Sync.ServerMessengerBase.Deliver(IEnumerable`1 servers, ICacheRefresher refresher, MessageType messageType, IEnumerable`1 ids, String json)
   at Umbraco.Core.Sync.ServerMessengerBase.PerformRefresh(IEnumerable`1 servers, ICacheRefresher refresher, String jsonPayload)
   at Umbraco.Web.Cache.DistributedCache.RefreshByJson(Guid factoryGuid, String jsonPayload)
   at Umbraco.Web.Cache.DistributedCacheExtensions.RefreshUnpublishedPageCache(DistributedCache dc, IContent[] content)
   at Umbraco.Core.Events.TypedEventHandler`2.Invoke(TSender sender, TEventArgs e)
   at Umbraco.Core.Events.EventDefinition`2.RaiseEvent()
   at Umbraco.Core.Events.ScopeEventDispatcher.ScopeExitCompleted()
   at Umbraco.Core.Events.ScopeEventDispatcherBase.ScopeExit(Boolean completed)
   at Umbraco.Core.Scoping.Scope.<>c__DisplayClass70_0.<RobustExit>b__1()
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions)
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions)
[...]

PR #8436, opened 2020-07-14
v7: fix content service - unpublish

### LINKED ISSUE #8393: V7: it is impossible to remove node with children after unpublish only parent.
v7: 7.15.4
Steps to reproduce:
1) install Umbraco with Starter Kit
2) Unpublish about Us
3) Try to remove About Us
Expected result:
About us with childrens will be moved to trashbin
Actual result:
image

bielu:
### Prerequisites

- [ ] Step to test are describe in following issue

If there's an existing issue for this PR then this fixes #8393

I added logic if there is any children to unpublishing all children.

### DIFF
--- a/src/Umbraco.Core/Services/ContentService.cs
+++ b/src/Umbraco.Core/Services/ContentService.cs
@@ -2544,7 +2544,7 @@ namespace Umbraco.Core.Services
 
             if (content.HasChildren())
             {
-                foreach (var subcontent in content.Ancestors())
+                foreach (var subcontent in content.Children())
                 {
                     UnPublishDo(subcontent, false, userId);
                 }
--- a/src/Umbraco.Core/Services/ContentService.cs
+++ b/src/Umbraco.Core/Services/ContentService.cs
@@ -2542,6 +2542,14 @@ namespace Umbraco.Core.Services
                 uow.Commit();
             }
 
+            if (content.HasChildren())
+            {
+                foreach (var subcontent in content.Ancestors())
+                {
+                    UnPublishDo(subcontent, false, userId);
+                }
+
+            }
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Hmm, I actually don't think it is supposed to work this way, for performance reasons we specifically do not do this and unpublishing a parent should just move all descendants to the recycle bin based on if any ancestors have -20 in the path.. 

Additionally I've tried to reproduce the original problem and I can't. I have created a node with children and descendants, throw it in the recycle bin and am then perfectly able to delete it from the bin. I also tried with the starter kit but also no problem..



I'll close this one and the related issue. Of course if you have other steps to repro or can find the root cause of your problem we're happy to have another look!

bielu:
@nul800sebastiaan you actually missed one step! 
I describe that issue occurs when you unpublish node with children first.

nul800sebastiaan:
@bielu Ah yes, okay, now I can reproduce! Unfortunately this is not how it should be fixed, unfortunately I also don't have guidance on how it really should be fixed 🙈

bielu:
@nul800sebastiaan the issue here is we need go through all children (perf issue) or change how removing things works(breaking change) I was looking into that over last week and it is the best what I found we can do without breaking staff around but maybe @Shazwazza will have better idea.

nul800sebastiaan:
Something must have changed along the way, I just tried in 7.12 and there I can perfectly delete it.. strange!

bielu:
I think it is related to changes of cache in 7.14/7.15(not sure now)

c9mbundy:
Perhaps related to #8244 ? I closed this issue as reproducing was inconsistent, but I also suspected something to do with cache.

bielu:
@c9mbundy this is same bug but in different context.

bielu:
@nul800sebastiaan any ideas or maybe information from internal conversations?

PR #8437, opened 2020-07-14
Fix for issue #7461 ("Inserting a macro in the rich test editor (RTE) in the Grid makes it disappear on Save & Publish")

### LINKED ISSUE #7461: Inserting a macro in the rich test editor (RTE) in the Grid makes it disappear on Save & Publish
The scenario is simple and reproducible, as mentioned here:
https://github.com/umbraco/Umbraco-CMS/issues/6846#issuecomment-546675121

> If you click into another area of the RTE after inserting the macro, it will save / publish with no issues, but if you insert the macro into the position you want within the RTE and then save, it gets stripped out / removed.

We're seeing this in all versions (or at least I went back to 8.4.0, 8.3.0, 8.2.2 and 8.1.5).

xdjoshuaaz:
### Description
Fixes #7461

This PR contains two changes to grid RTE editor code;
a) GridRteDirective now finds the `currentForm` using the parent scope from the GridController above the GridRte directive. The GridRte directive creates its own isolated scope and therefore we need to search for the form in `$scope.parent` and not the isolated `$scope`. This does NOT fix the issue mentioned but fixes the dirty event handler in the RTE so that `form.$setDirty()` is called correctly.
b) `syncContent()` is now called on TinyMCE Dirty events so that model.value is set to the RTE's contents whenever a macro is edited or inserted. This fixes the issue mentioned above.

GIF before where the text isn't saved after saving and publishing content:
2020-07-14T14-41-06--18

GIF after the fix:
2020-07-14T14-42-15--20

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/grid/grid.rte.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/grid/grid.rte.directive.js
@@ -87,7 +87,9 @@ angular.module("umbraco.directives")
                         tinyMceService.initializeEditor({
                             editor: editor,
                             model: scope,
-                            currentForm: angularHelper.getCurrentForm(scope)
+                            // Form is found in the scope of the grid controller above us, not in our isolated scope
+                            // https://github.com/umbraco/Umbraco-CMS/issues/7461
+                            currentForm: angularHelper.getCurrentForm(scope.$parent)
                         });
 
                         //custom initialization for this editor within the grid
--- a/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
@@ -1504,6 +1504,8 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
             });
 
             args.editor.on('Dirty', function (e) {
+            	syncContent(); // Set model.value to the RTE's content
+
                 //make the form dirty manually so that the track changes works, setting our model doesn't trigger
                 // the angular bits because tinymce replaces the textarea.
                 if (args.currentForm) {


nul800sebastiaan:
Hi @xdjoshuaaz - first off apologies for the delay! I meant to get to your PR sooner but got a bit delayed!

This is great and works just fine, so all good. 👍 

I see this is your first PR for Umbraco CMS, congratulations and great job, this is not an easy part of the code to decipher! 🎖️ 

We'd love to give you a contributor badge on [Our Umbraco](https://our.umbraco.com) if you have an account there, couldn't find you at the moment. 

Thanks again and great job, hope to see more contributions from your hand! 🎉

anna-Kop:
Hello!

In which version of Umbraco will this be fixed? I'm using 8.7.0 and I still have the problem.

Is there a way I can use this fix already?

nul800sebastiaan:
@anna-Kop this is fixing issue https://github.com/umbraco/Umbraco-CMS/issues/7461 - which will be released with Umbraco 8.8.0, which is due out on October 1: https://our.umbraco.com/download/releases/880

MrCSharp22:
I've upgraded to 8.8.0 RC to test this out and the issue still persists:

umbraco-8 8 0-rc


is there anything I might've missed?

nul800sebastiaan:
@MrCSharp22 Usually it's browser caching. Try updating the version in ClientDependency.config (just make it 1 higher or lower) and try again.

MrCSharp22:
Thanks @nul800sebastiaan 
I did as you suggested but that didn't fix the issue. This was tested on Firefox and Chromium Edge too. I saw that the ```cdv``` parameter had the new version in the dev tools window and also made sure cache was disabled when testing this.

However, I noticed that once I insert the macro, I have to click inside the RTE then click on "Save and Publish". I can see CMS editors missing this and losing some content, but this is a definitely an improvement over the old behavior.

TrilochanTripathi-01:
I've upgraded to version 13.4.0 and same issue, "Once I've inserted macro in RTE it disappears but if I click back inside RTE it saves without issues."

PR #8441, opened 2020-07-15
Streamline umb-checkbox and umb-radiobutton components

bjarnef:
### Prerequisites

### Description
Recently features has been added to `umb-checkbox`, but not to `umb-radiobutton`.
I think these two components should be as consistent as possible. If new features are added to one of them, it should probably also been included on the other one.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
@@ -31,8 +31,9 @@
 @param {boolean} disabled Set the checkbox to be disabled.
 @param {boolean} required Set the checkbox to be required.
 @param {callback} onChange Callback when the value of the checkbox change by interaction.
-@param {string} cssClass Set a css class modifier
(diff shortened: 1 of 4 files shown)


bjarnef:
It would have been great the `umb-checkbox` and `umb-radiobutton` just used `icon` instead of `icon-class` to be consistent with other components in core using icons as well. Unfortunately we can't change that now as it was introduced in https://github.com/umbraco/Umbraco-CMS/pull/6979

**umbButton**
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbutton.directive.js#L65

**umbFileIcon**
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/umbfileicon.directive.js#L41

**umbNodePreview*
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/umbnodepreview.directive.js#L79

nul800sebastiaan:
Wonderful!! Thanks again @bjarnef 👍

PR #8444, opened 2020-07-15
Return Image only type when datatype is configured as such

patrickdemooij9:
Currently when you have a media data type with "Images only" enabled, the result will still be IPublishedContent, even though everything is an Image object. We can easily return the data as IEnumerable<Image> or Image when the option is turned on.

Took most of the code from the nested content property value converter:
https://github.com/umbraco/Umbraco-CMS/blob/df8f28114160271ed4ab5db4519b4d13c3acccfc/src/Umbraco.Web/PropertyEditors/ValueConverters/NestedContentManyValueConverter.cs
https://github.com/umbraco/Umbraco-CMS/blob/df8f28114160271ed4ab5db4519b4d13c3acccfc/src/Umbraco.Web/PropertyEditors/ValueConverters/NestedContentSingleValueConverter.cs

Way to test:
Create a model with modelsbuilder that has a mediapicker with the option to only choose images. It should come out as an Image object now instead of IPublishedContent. If the media picker has multiple enabled, it'll return an IEnumerable<Image>

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerValueConverter.cs
+++ b/src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerValueConverter.cs
@@ -79,8 +79,8 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
             PropertyCacheLevel cacheLevel, object source, bool preview)
         {
             var config = GetConfiguration(propertyType.DataType);
-            var isMultiple = config.Multiple;
-            var isImagesOnly = config.OnlyImages;
+            var isMultiple = IsMultipleDataType(config);
+            var isImagesOnly = IsImagesOnlyDataType(config);
 
             var udis = (Udi[])source;
             var mediaItems = isImagesOnly
--- a/src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerValueConverter.cs
+++ b/src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerValueConverter.cs
@@ -26,7 +26,8 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
         {
             _publishedSnapshotAccessor = publishedSnapshotAccessor ??
                                          throw new ArgumentNullException(nameof(publishedSnapshotAccessor));
-            _publishedModelFactory = publishedModelFactory;
+            _publishedModelFactory = publishedModelFactory ??
+                                     throw new ArgumentNullException(nameof(publishedModelFactory));
         }
 
         public override bool IsConverter(IPublishedPropertyType propertyType)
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks @patrickdemooij9 - but what happens to existing sites, expecting an `IPublishedContent` now.. seem like they will need to change their templates to expect a new type?

patrickdemooij9:
Good question @nul800sebastiaan. I believe that for the most part it should go fine as Image inherits from IPublishedContent, however there may still be situations where some special logic is used that would break with this change.

nul800sebastiaan:
I am not so sure, I am currently thinking this will be a large breaking change for people somehow working with the returned type of 
`IPublishedContent` and before we can evaluate this PR we'd need to have some re-assurance that existing usages won't break. 

This is of course unfortunate but sometimes the mistakes we put out there can't be simply corrected a year and a half later without causing problems for a significant amount of users. 

What can we do to make sure this is a non-breaking update?

patrickdemooij9:
But will it really break the IPublishedContent usages as Image is still an IPublishedContent itself. For example, this will still work:
```
public Test(IPublishedContent image){
}

var test = new Test(image) //image being of type Image
```
Because Image will just cast itself down to IPublishedContent right? Every Image is an IPublishedContent. However, there could be a way to make this non-breaking and that is to introduce it as a PropertyValueConverter that you have to append yourself in 
a Composer (Or the other way around) if you want to use it.

kjac:
I do agree with @nul800sebastiaan - this is going to be breaking for a number of usages.

It is also a false assumption to believe that "Images only" will return only Image media types. You can create multiple media types that will register as "image" media types.

callumbwhyte:
**Everyone** is right here! Yes this is annoying. Yes this would be a breaking change. 😊

This is a long standing issue, way back from when Value Converters were first introduced! This has been broken in recent memory too, when it was unexpectedly changed in v8.1.4 (see #6034)... Ignoring how we got here, there's definitely an issue frustrating / confusing people here that can be fixed.

## The issue

@kjac is absolutely correct that it's wrong to assume selecting "Images only" means only the `Image` media type can be selected - I think this is an oversight in the original logic.

In fact the [picker logic](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js#L53) is based on the file extension of the media item selected - this allows it to support custom media types, but even means if you uploaded a JPG as a `File` media type you can select this too _(...I think this might actually create other issues... 🤯)_

It would be sensible that the logic of the property value converter should also work the same way as the picker... But that's not how Value Converters work. Value Converters determine the type at generation time, while the picker is based on the selected content.

## Proposal

I believe the ideal solution would be that the media picker is updated to work in a similar way to other pickers with the option to choose which media types can be selected.

We do away with the concept of an "image" and allow this to be controlled by the allowed media types. _This also gives a bit more flexibility - imagine I only want allow PDFs to be uploaded... I create a custom media type with that restriction and bingo._

Not only does it give flexibility, but also makes things less confusing for beginners and unlocks potential to make the developer experience better (e.g. by package devs, see below).

Of course removing the "image only" checkbox completely is a breaking change, but replacing it with a media type picker and populating it the relevant media types (e.g. as part of a migration) might be enough to mitigate issues in the majority of cases? Regardless it would be classed as a minor breaking change and should be documented as such...

Doing this wouldn't directly fix the issue - but it does make pickers behave more consistently:

- [MNTP converter](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/ValueConverters/MultiNodeTreePickerValueConverter.cs#L45) returns `IPublishedContent` / `IEnumerable<IPublishedContent>`
- [Nested Content converter](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/ValueConverters/NestedContentManyValueConverter.cs#L37) returns `IEnumerable<IPublishedElement>` / `IEnumerable<T>` _(yay, also inconsistent...)_
- [Member Picker converter](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/ValueConverters/MemberPickerValueConverter.cs#L29) returns `IPublishedContent`

## Solution

Once we have consistent behaviour for pickers it makes room for custom implementations, or packages like my [SuperValueConverters](https://github.com/callumbwhyte/super-value-converters) package to step in and do it's thing - it's sole job is returning the specific type that you have picked, where Core can just return the generic interface.

I have a version of this package running in my own projects with a "hack" that looks up which media types have an Image Cropper property editor and treats them as "images" in order to return the correct type. I never released it because it's a bit of a workaround, the above proposal would make things much nicer...

This gives people choice, and means we can maybe think about bringing something like this into Core as part of a major release in future (it would be a big breaking change).

stefankip:
Thanks @patrickdemooij9, implemented this in my project and boilerplate, really saves me valuable time!

nul800sebastiaan:
Thanks for the feedback all! I will close this one for now, we do need to rethink how these pickers work first indeed! Ah the curse of the legacy.. 😟

PR #8445, opened 2020-07-15
Adjust toggle styling when checked and disabled

bjarnef:
### Prerequisites

### Description
At the moment when `umb-toggle` is checked and disabled, it is not clear the toggle is checked. This was changed in PR https://github.com/umbraco/Umbraco-CMS/pull/7626

For example in languages where default language here is checked and disabled.

chrome_2020-07-15_18-57-29

I have adjusted the toggle background color, so it looks more look the active checked state and a bit dimmed, which is a common pattern used accross different web interfaces, e.g https://css-tricks.com/custom-styling-form-inputs-with-modern-css-features/

image

Finally it removes unneccessary use of `ng-if`.


bjarnef:
@nielslyngsoe it would be great with this update.

nul800sebastiaan:
Good stuff @bjarnef - thanks very much! 👍

PR #8458, opened 2020-07-21
NetCore: Migrating Macro Templates

elit0451:
Migrating as many of the templates from `Umbraco.Web.UI\Umbraco\PartialViewMacros\Templates` to `Umbraco.Web.UI.NetCore\Umbraco\PartialViewMacros\Templates`.

The ones that left are listed below:
- [ ] **EditProfile.cshtml** - making use of `MembershipHelper`, therefore cannot be moved yet
- [ ] **Login.cshtml** - uses unmigrated controller and other
- [ ] **LoginStatus.cshtml** - making use of `MembershipHelper`, therefore cannot be moved yet
- [ ] **RegisterMember.cshtml** - making use of `MembershipHelper`, therefore cannot be moved yet

----
### Note:

We need to figure out **which extension method** we need to use to get the cropUrl in [Gallery.cshtml](https://github.com/umbraco/Umbraco-CMS/blob/717d8e6156d49665a53d230608e38ee276e49c68/src/Umbraco.Web.UI.NetCore/Umbraco/PartialViewMacros/Templates/Gallery.cshtml#L55) and then ln.53-ln.57 can be uncommented on that file.


src/Umbraco.Web.UI.NetCore/Umbraco/PartialViewMacros/Templates/Gallery.cshtml
```diff
@@ -1,10 +1,14 @@
...
-@inherits Umbraco.Web.Macros.PartialViewMacroPage
-
+@using Umbraco.Core.Composing
+@using Umbraco.Core.PropertyEditors.ValueConverters
+@using Umbraco.Web.Routing
+@inherits Umbraco.Web.Common.Macros.PartialViewMacroPage
+@inject IPublishedContentQuery PublishedContentQuery
+@inject IVariationContextAccessor VariationContextAccessor
+@inject IPublishedUrlProvider PublishedUrlProvider
+@inject IFactory Factory
```
bergmania: Why is Factory injected?  You can inject the `IImageUrlGenerator` instead

elit0451: Makes sense! But as I mentioned after we uncomment ln.[53](https://github.com/umbraco/Umbraco-CMS/blob/717d8e6156d49665a53d230608e38ee276e49c68/src/Umbraco.Web.UI.NetCore/Umbraco/PartialViewMacros/Templates/Gallery.cshtml#L53)-57, we will have problems with the `GetCropUrl`, so perhaps we need to introduce a new extension method or get the crop URL differently?

PR #8463, opened 2020-07-22
Improving umb-button link docs

callumbwhyte:
While implementing a new backoffice UI I found the docs for the `umb-button` to be inaccurate.

In order for the `href` attribute to work the `type` attribute must be set to "link", else you end up with a button that doesn't do anything on click...

This was missing from the docs, it now isn't! 🎉


nul800sebastiaan:
Lovely, thanks @callumbwhyte ! 👍

PR #8455, opened 2020-07-20
Fix issue with Removing after unpublish parent

bielu:
### Prerequisites

- [ ] I have added steps to test this contribution in description of related issue

If there's an existing issue for this PR then this fixes #8393

### Description

@nul800sebastiaan  here you have PR with proposed fixed

PR #8464, opened 2020-07-22
v7: Revert previous changes, force writer in update of cache

bielu:
### Description
In previous PR #8455 , we fixed issue which was occurring but in time of QA for client we find out that change is actually causing more issues in different places, it is why I continue to research on issue. and find out it is enough to force Xml writer to be writer and all issues are solved.
@nul800sebastiaan  I am doing PR to check again unit tests results as I already notice they could be different on local and CI env (not sure why)


nul800sebastiaan:
Thanks @bielu - this makes sense and the unit tests are happy again, also the backoffice is now happy when trying to delete nodes! 👍 

I am super annoyed that I can't find a way to make a failing unit test for this, I don't understand why [this unit test](https://github.com/umbraco/Umbraco-CMS/pull/8455#issuecomment-661912223) works before applying your changes.

I've added the unit test now so we can at least detect any problems in the future. Let us know if you find a good way to make a test that would fail on the error that was encountered that is now fixed!

PR #8467, opened 2020-07-23
Make icon property for checkbox and radiobutton consistent with other components

bjarnef:
### Prerequisites

### Description
It would have been great if the `umb-checkbox` and `umb-radiobutton` just used `icon` instead of `icon-class` attribute to be consistent with other components in core using icons as well. This was introduced in https://github.com/umbraco/Umbraco-CMS/pull/6979

At the moment this is only used for icons in richtext editor configuration in core.

image

This PR add an `icon` parameter as other components and and mark `iconClass` as deprecated. Maybe this can be removed in a minor version. Note sure if the `@deprecated` in `ngdoc` is entirely correct.

**umbButton**
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbutton.directive.js#L65

**umbFileIcon**
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/umbfileicon.directive.js#L41

**umbNodePreview**
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/umbnodepreview.directive.js#L79

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
@@ -32,7 +32,8 @@
 @param {boolean} required Set the checkbox to be required.
 @param {callback} onChange Callback when the value of the checkbox change by interaction.
 @param {string} cssClass Set a css class modifier.
-@param {string} iconClass Set an icon next to checkbox.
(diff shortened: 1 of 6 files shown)


nul800sebastiaan:
Lovely stuff indeed, thanks again @bjarnef! 👍

PR #8472, opened 2020-07-24
Add missing esc shortcut to close examine result field values overlay

bjarnef:
### Prerequisites

### Description
When opening member search result overlay with title "Field values" from MembersIndex it isn't possible to close this overlay via `esc` shortcut.


nul800sebastiaan:
Ha! Very good, that has annoyed me before! Thanks! 👍

PR #8474, opened 2020-07-24
Fix null language in variant change event

nathanwoulfe:
### Prerequisites

Resolves issue #8355 where language is null on invariant content, so attempting to access language.culture throws an exception. Changed to pass the language object as the event data, rather than language.culture


nul800sebastiaan:
@nathanwoulfe Thanks! I know this will fix the error but semantically it looks a bit weird: has the culture actually changed if there is no culture?

nathanwoulfe:
@nul800sebastiaan if a tree falls in the forest... I can wrap the event in a variant length check - if there's more than one variant, the culture has changed. That would also have been a solution for language.culture being null (but still makes better sense to emit the entire language object rather than just the culture property).

Hold the line caller, I'll update.

nul800sebastiaan:
@nathanwoulfe I'm also happy with always emitting the language, but then the event needs to be renamed! 🙂

nul800sebastiaan:
All good like this as well, thanks @nathanwoulfe 👍

PR #8475, opened 2020-07-24
7731 bye bye underscore => _.each

nathanwoulfe:
Changes replace `_.each` with native `.forEach` in directives (components) and resources - I've sanity checked these locally, all seems to be fine. This is related to #7731 but does not yet complete the _.each removal (figured it's better to have multiple smaller PRs rather than one big one with 50 files modified).

Most are like-for-like replacements, with the exception of some like the changes in umbsearch.directive.js where the value is a dictionary not an array, so uses `Object.keys().forEach()` because the value is not directly iterable.

Made related syntax changes to replace some indexOf calls with includes (easier to read `x.includes(y)` compared to `x.indexOf(y) !== -1`), and converted a few functions to single-line fat-arrows where it made sense to be terser.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/search.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/search.service.js
@@ -43,9 +43,7 @@ angular.module('umbraco.services')
                 }
 
                 return entityResource.search(args.term, "Member", args.searchFrom).then(function (data) {
-                    _.each(data, function (item) {
-                        searchResultFormatter.configureMemberResult(item);
-                    });
+                    data.forEach(item => searchResultFormatter.configureMemberResult(item));
                     return data;
                 });
             },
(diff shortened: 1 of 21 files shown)


nul800sebastiaan:
Looks good, thanks @nathanwoulfe! 👍

src/Umbraco.Web.UI.Client/src/common/services/search.service.js
```diff
@@ -119,7 +113,7 @@ angular.module('umbraco.services')
 
                 return entityResource.searchAll(args.term, args.canceler).then(function (data) {
 
-                    _.each(data, function (resultByType) {
+                    data.forEach(resultByType => {
```
nul800sebastiaan, line 116: @nathanwoulfe Uh oh! This one seems to have broken! 

image


Little help please? 🙏

nul800sebastiaan: Reproduce by trying to search in the backoffice:

image

nathanwoulfe: Hmm. I thought I'd sorted that... I'm guessing data isn't an array, but an object/dictionary. I'll grab the laptop...

PR #8476, opened 2020-07-25
Show open button for media and member in MNTP

bjarnef:
### Prerequisites

### Description
At the moment when selecting "Show open button" in configuration of Content Picker and MNTP, because it also rely on this variable https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/contentpicker/contentpicker.controller.js#L142

image

However infinite editing also works for media and member, so I have modified it to open `mediaEditor` and `memberEditor` as well based on the entity type.

**Media**

**Member**

I also noticed a missing option to close infinite editor via `esc` shortcut in content editor.

One minor thing missing is when clicking "Save" or "Save and publish" from content editor, it doesn't close the overlay like media and member editor do, but we can look at this in another PR.


bjarnef:
The contentpicker property editor (used for content picker and MNTP) also have `allowEditButton`:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/contentpicker/contentpicker.controller.js#L143

and `showEditButton` and `showPathOnHover`:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/contentpicker/contentpicker.controller.js#L90-L91

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/contentpicker/contentpicker.controller.js#L134-L135

but I can't see these are used anywhere. I guess this is a leftover from the old pickers, which had an option to show open, edit and path on hover as I remember.
@nielslyngsoe do you recall if these are even needed anymore?

bjarnef:
I also fixed a minor issue, where the mini listview was causing horizontal scroll while animating forward, when having e.g. "People" as start node.

**Before**

**After**

nul800sebastiaan:
@bjarnef Nice! But we need to check if the current user has access to these sections before showing the button. Media might be mostly fine, Members is not accessible for everyone though.

bjarnef:
@nul800sebastiaan ahh yes. Do you know where we do something similar to check if the user has access to these sections?

Furthermore do we need `showEditButton` and `showPathOnHover`. Seems to be a leftover from old pickers?

bjarnef:
Okay, we can fetch to current user and see the user groups the user belongs to and which sections the user group contains.

Maybe something like this. Not sure if the user object directly contains the sections, the user has access to.
https://github.com/umbraco/Umbraco-CMS/blob/b082d40921d3ccc939c28f1adda7973729b63b16/src/Umbraco.Web.UI.Client/src/views/users/views/groups/groups.controller.js#L33

bjarnef:
It seems the user object has a property `allowedSections`:
https://github.com/umbraco/Umbraco-CMS/blob/994c379a946fa7e36eed669e429ab26fa977b00c/src/Umbraco.Web.UI.Client/src/routes.js#L42

nul800sebastiaan:
@bjarnef Yes, we do it here, for example: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/mediapicker/mediapicker.controller.js#L148

> Furthermore do we need `showEditButton` and `showPathOnHover`.

Yeah, they seem completely unused, think they're fine to remove!

Although.. this is where it got removed from the view: https://github.com/umbraco/Umbraco-CMS/commit/95217422546ba71227bfaa550cc3a291a8b357b8

But added as additional data.. so I have no idea now.. Maybe just best to leave it.

nul800sebastiaan:
> It seems the user object has a property `allowedSections`:

You found the same thing as I did 😉

bjarnef:
@nul800sebastiaan I have update this to check to current user has access to the section.
Previous it always showed "Open" button for document entity type, but I guess you could use MTNP on media or member and a user doesn't have access to content section? So I have made the check for content section as well.

While testing I removed "Media" from "Administrators" group I confirmed the open button wasn't shown. When going back to user group sections and picked "Media" again I noticed it only selected "Media" and the existing ones was removed. I think the existing ones should be preselected (didn't they used to be?).

Funny enough I succeed to save the "Administrators" user groups with just only "Media" section and was logged out. When logging in again I only have access to "Media" 🙈 woops!

image

bjarnef:
@nul800sebastiaan do you know if I someway can reset access for "admin" user group via db to default sections without doing a clean Umbraco install?

bjarnef:
@nul800sebastiaan okay, I could modify the data for admin (super user) to get access as editors who have access to users sections.

bjarnef:
It seems we also have these only set from code.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/MultiNodePickerConfigurationEditor.cs#L33-L34

nul800sebastiaan:
> Previous it always showed "Open" button for document entity type, but I guess you could use MTNP on media or member and a user doesn't have access to content section? So I have made the check for content section as well.

Fantastic, never thought of that! 👍 

As for `showEditButton` and `showPathOnHover` let's leave them for now and do some further exploration of what happens if we remove those.

bjarnef:
@nul800sebastiaan I have reverted the commit to re-add https://github.com/umbraco/Umbraco-CMS/pull/8476/commits/35b82151232ce2494d7189dc31fdf21206859bd6 `showEditButton` and `showPathOnHover`.

bjarnef:
I don't think `showPathOnHover` is used anymore at least not available in `umbNodePreview`.
https://github.com/umbraco/Umbraco-CMS/blob/a4274ba76727273eb8d7cfb02b97fc5bdcd3e9cc/src/Umbraco.Web.UI.Client/src/common/directives/components/umbnodepreview.directive.js#L79-L93

The "edit" can be used in `umbNodePreview`, but didn't seem to be used here.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/umb-node-preview.html#L19-L39

nul800sebastiaan:
Lovely work @bjarnef and it works well! I think a new PR to remove the two variables would be good so we can do a proper little investigation before merging! 👍 

Thanks again!

PR #5606, opened 2019-06-06
V8: Add umb-icon directive to backoffice to begin transition from icon fonts to svgs

MMasey:
_This was initially a much larger PR so I have decided to recreate it with far fewer commits and code changes to hopefully make it easier to merge The existing icon system is still fully functional so once this is in, we could gradually go through the backoffice and make the changes one by one, rather than one huge monolith of a PR (which is what this one ended up being #5293). The majority of the changes that are part of this PR are from adding the 605 new individual svg icons._ 😜

### Summary
This PR updates almost all of the icons using the icon font found throughout the backoffice with SVGs. Each icon is now it's own svg file with the following naming convention icon-[name]. I have created an IconController that currently pointed to a folder containing these icons and makes them usable via the an angular directive call umb-icon. This not only improves accessibility, but also make it much easier to customise the icons available in the back office.

For this PR I have only updated the IconPicker, and a few other icons that were using the `<ins>` or `<span>` tag. The rest of the changes from `<i>` to `<umb-icon>` can happen later.

One thing that I think needs to be added is the ability to sanitise the SVG's as the IconController reads the svg file directly in order for the angular directive to make use of it. I've found a package that can do HTML Sanitizing that can have additional tags and attributes added to the list of allowed data so that it can work with SVGs too. I have this running without issue on a seperate project and can confirm that it works great.

The package is https://github.com/mganss/HtmlSanitizer

I suppose the issue with is it that it has a dependency on AngleSharp, and I know we're trying to reduce dependencies in v8. 

### `<umb-icon />` directive

```
// The most basic `umb-icon` directive looks something like this
<umb-icon icon-name="icon-profile" />


// You can pass any other attribute through to it as you would any other directive, such as...
<umb-icon icon-name="icon-profile" class="icon icon-profile large" ng-show="model.show"></umb-icon>
```

All of the individual icon svgs now live under `src\Umbraco.Web.UI.Client\src\assets\icons`. When the frontend build process is run these get copied over to the `src\Umbraco.Web.UI\Umbraco\assets` folder, in the same way the other frontend source files do.

These are 2 API endpoints that can be used to get the SVG markup. The controller for these can be found at `v8\src\Umbraco.Web\Editors\IconController.cs`

### GetIcon
returns a single IconModel
```
e.g. - http://umbraco.v8/umbraco/backoffice/UmbracoApi/Icon/GetIcon?iconName=icon-navigation-right

returns
{
    "Name":"icon-navigation-right",
[...]

PR #8477, opened 2020-07-25
Update boolean/toggle/checkbox label prevalues

bjarnef:
### Prerequisites

### Description
The True/False property editor aka Checkbox has at the moment a prevalue in the configuration to set a label text, but "Write a label text" is not very informative. Furthermore it both used this for "label-on" and "label-off".

Instead it now has two prevalues to set text for "Label On" and "Label Off". If the developer want the same text whether the toggle is checked or not, is it quite easy to set same label text in both prevalues.

Maybe some even prefer it only show a specific label when the toggle in checked of not checked?

At the moment it would only show labels if the toggle has a "labelOn" value and not if "labelOff" actual has a value:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/boolean/boolean.html#L6
So the configuration now has a property "Show labels" to active decide whether labels should be shown or not.
Furthermore if either labelOn or labelOff didn't had a value `umb-toggle` has fallback to `On` and `Off`. I have changed this so it only use this fallback if both fields are empty - or if both `label-on` and `label-off` in the `umb-toggle` component are empty. With this you can e.g. show a label text only if the toggle is checked or unchecked. e.g. if you specify a custom label on text "Day" a no value for label off text (Night), it shouldn't default to "Off" right? Better only use the default values is none label text has been filled.

I have already updated the property editor name from "Checkbox" to "Toggle" since it hasn't much to do with a checkbox anymore. If replacing the font icons with SVG icons as in this PR https://github.com/umbraco/Umbraco-CMS/pull/5606 we might even use something like `icon-toggle` or `icon-switch` instead of `icon-checkbox` sometime later.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbtoggle.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbtoggle.directive.js
@@ -87,22 +87,28 @@
             }
 
             function setLabelText() {
-
-                // set default label for "on"
+                
                 if (scope.labelOn) {
                     scope.displayLabelOn = scope.labelOn;
-                } else {
-                    localizationService.localize("general_on").then(function (value) {
-                        scope.displayLabelOn = value;
-                    });
                 }
-
-                // set default label for "Off"
+                
                 if (scope.labelOff) {
                     scope.displayLabelOff = scope.labelOff;
-                } else {
-                    localizationService.localize("general_off").then(function (value) {
-                        scope.displayLabelOff = value;
+                }
+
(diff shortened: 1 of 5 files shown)


nul800sebastiaan:
Lovely! I made it a little less "checkboxy":

image

I feel enabled/disabled is a better description if we're no longer referring to this as a checkbox.

We'll probably need to update some documentation somewhere to change from Checkbox to Toggle.

bielu:
@nul800sebastiaan change of type is breaking change, it should be flagged as breaking.
Default change from string to boolean, and if someone worked directly, will have issues here.

PR #8480, opened 2020-07-25
Richtext rules adjustments

bjarnef:
### Prerequisites

### Description
A few adjustments to richtext styles editor using `<button>` elements instead of `<a>` elements.
Furthermore is replace a few fonts using `@monoFontFamily` variable, where it makes the textarea look more like an css editor.

image

I considered it ace editor code be used here as well, but I think it need to include the selector as readonly line. https://stackoverflow.com/a/39640987/1693918
Might be possible, but a bit more complex and would require another PR.


nul800sebastiaan:
Nice one! Would be really great if ace can be used, nice autocomplete and highlighting would be available!

Thanks @bjarnef! 👍

PR #8132, opened 2020-05-18
Add "umb-button-ellipsis" component

BatJan:
### Prerequisites

- [X] I have added steps to test this contribution in the description below

### Description
~~This PR is still WIP but I decided to create it after noticing 4-5 different variants of achieving the same thing in the backoffice.~~

**EDIT:** This PR is ready for review now. The aim is to ensure that whenever an ellipsis button is needed it can be reused with some CSS modifications to change the colors etc. but at the same time ensure proper accessibility is baked in etc.

I besides creating this component I have also replaced the old markup patterns with this newly created component as well. Below is a series of GIF's of what the ellipsis looks like the places where I have changed them - This also serves as a reference of where to look in the main branch to review and compare the ellipsises after my proposed changes.

**Content section**
content-section-ellipsis-after

**Section and tabs ellipsis**
section-and-tabs-ellipsis-after

**Content app ellipsis after**
content-app-ellipsis-after

PR #8481, opened 2020-07-26
Hide action button in tree

bjarnef:
### Prerequisites

### Description
This previous PR https://github.com/umbraco/Umbraco-CMS/pull/8132 introduced a `umb-button-ellipsis` components and replaced with this various placed with "ellipsis" button. One of these placed is the action button in tree.

However `umb-tree` has an options to `hideoptions` which set a class `hide-options` and the following styling which hide the button with class `umb-options`: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less#L217-L219

The new button in tree does however no longer has this class, so it is always shown on hover (e.g. in macro partial view picker) even `hideoptions` is set. Note this is used quite many places: https://github.com/umbraco/Umbraco-CMS/search?q=hideoptions&unscoped_q=hideoptions where this is the one when using treepicker infinite editor https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/treepicker/treepicker.html#L59

This PR add this class to the component here (because we might not want to use this class on all `umb-button-ellipsis` components and furthermore it make it easier to target some specific styling to this inside tree.

**Before**

image

**After**

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/buttons/umb-button-ellipsis.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/buttons/umb-button-ellipsis.less
@@ -1,7 +1,7 @@
 .umb-button-ellipsis{
     padding: 0 5px;
     text-align: center;
-    margin: 0 10px 0 auto;
+    margin: 0 auto;
     cursor: pointer;
     border-radius: @baseBorderRadius;
     color: black;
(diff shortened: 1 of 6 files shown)


nul800sebastiaan:
Wonderful, great spot @bjarnef!! 👍 

I'm just going to copy in @BatJan so he can have a little look. I've merged this and I think all of this now seems fixed but an extra set of eyes helps as well.

BatJan:
@bjarnef @nul800sebastiaan Looks fine to me as well - Good job 👍

PR #8485, opened 2020-07-27
Do not rebind content types on save errors

### LINKED ISSUE #8465: Groups in Document Types are being removed if there is a duplicate
## Umbraco version

I am seeing this issue on Umbraco version: 8.5.4

Reproduction
------------

### Bug summary

Groups in Document Types are being removed if there is a duplicate!

### Specifics

I detected this when I tried to add a Group, followed by a Property, and I got the "Duplicate group names not allowed". But this only happens if I do it through a Content of that same Document Type (going in "Info" and then editing/adding through there). It does not happen if we do it through Settings -> Document Types.

I used Umbraco 8.5.4, with Google Chrome Version 84.0.4147.89 (Official Build) (64-bit). I have not checked in other browsers

### Steps to reproduce

Create a Document Type (in my case, it is one that has a Template associated).
Add a Group and a Property in that Group. Save it.

Create a content of that same Document Type, fill out fields and save.
Then, go to the "Info" tab, and edit the Document Type through there.
Add a Group with the name of the previously created Group and add a Property. The objective is to provoke a "duplicate error".
After the error, rename the Group and then save.

After all this, refresh the page of the content, and the original Group should disappear, along with the properties and the content (re-added the properties and the values were gone).

### Expected result

The expected result was to keep the original Group, and not delete it, along with the Properties and it's values.

Expected the same result if I made the same steps through Settings -> Document Types.

### Actual result

Message saying "Duplicate group names not allowed", and deletion of Group that already existed, along with Properties and it's values.

---
_This item has been added to our backlog [AB#7549](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7549)_

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8465

### Description

See #8465 for an issue description and a screencast of the issue as it plays out.

The issue happens because the content type editor is rebound with data from the server after an unsuccessful save event. The way the rebind callback is implemented, it accidentally assigns duplicate IDs to the groups with matching names, and thus one of them ends up overwritten on the next successful save.

I don't believe it makes sense to rebind the content type editor if the save fails. Nothing has been persisted on the server anyway. So this PR simply removes the rebinding upon failed saves.

With this PR applied, the editor behaves like you'd expect - preserving the previously duplicated groups after renaming them appropriately:

do-not-rebind-on-save-error

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/contenteditinghelper.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/contenteditinghelper.service.js
@@ -100,10 +100,7 @@ function contentEditingHelper(fileManager, $q, $location, $routeParams, editorSt
                         self.handleSaveError({
                             showNotifications: args.showNotifications,
                             softRedirect: args.softRedirect,
-                            err: err,
-                            rebindCallback: function () {
-                                rebindCallback.apply(self, [args.content, err.data]);
-                            }
+                            err: err
                         });
 
                         //update editor state to what is current


nul800sebastiaan:
Well isn't that lovely! Thanks a bunch @kjac! 👍 ⭐

PR #8486, opened 2020-07-27
Fix broken "more" button in editor app navigation

kjac:
### Prerequisites

### Description

The "more" button in the editor app navigation has gone and gotten slightly broken after merging in #8132. Actually it looks pretty good when there are no apps displayed at all, but soon as only some apps are there (i.e. when using content apps), it starts looking really weird:

image

image

image

This PR reverts the editor app navigation part of #8132 and redoes it to be keyboard navigation friendly. Once applied, the "more" button looks like the rest of the content app buttons once again:

image

image

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-editor-navigation-item.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-editor-navigation-item.less
@@ -142,21 +142,7 @@
     // --------------------------------
 
     &-more__icon {
-        margin-bottom: 10px;
-
-        i {
-            height: 5px;
-            width: 5px;
-            border-radius: 50%;
-            background: @ui-active-type; // fallback if browser doesnt support currentColor
-            background: currentColor;
-            display: inline-block;
-            margin: 0 5px 0 0;
-        }
-
-        i:last-of-type {
-            margin-right: 0;
-        }
+        margin-bottom: 6px;
     }
 
     &-more__dropdown {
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Thanks! Of course the point of having a `umb-button-ellipsis` was to use it everywhere instead of reverting to a button that does

nul800sebastiaan:
Thanks! Of course the point of having a umb-button-ellipsis was to use it everywhere instead of reverting to a button that will not be changed if the directive gets changed. Was it not possible to update the directive to fix this specific problem?

kjac:
Nope unfortunately not. At least not without breaking it everywhere else. Also I'm suspecting that even if we managed, it'd only be a matter of time before other changes to the same directive would break the navigation. As you've noticed it's not the first thing you see on a vanilla install, so things might slip by testing.

nul800sebastiaan:
Surely we don't have to throw the baby out with the bathwater, it seems to me that an optional variable for the directive could do the trick just fine? 😅

kjac:
@nul800sebastiaan parameterizing adding a parameter/variable to the directive won't do any good. One would have to rewrite the entire directive with a big ol' if/else to apply the necessary styling to the directive contents, and then we'd be even worse off 😄 

I have created an alternative approach in #8492 where the lost styling is re-applied to the "more" button without removing the directive usage. I honestly think this PR is a better approach though, as the changes in #8492 are equally prone to regression issues.

BatJan:
@kjac I've added yet another proposed fix for this in #8500 - Could you please have a look at it? I think it's cleaner and I don't think that it should be prone to regression issues unless I'm missing something? 😅

nul800sebastiaan:
Cheers! Picking #8500 over this one! 👍

Issue #8487, opened 2020-07-27
Existing sections in user group section picker should be preselected


bjarnef:
I noticed while editing sections for "Administrators" user group after removing media and wanted to re-add media section, the existing sections wasn't preselected, which they used to be.
Instead the list was cleared after selecting "Media" and only "Media" was added to sections list.

I was a bit too fast to save, so I was logged out. When logging back in I only had access to media section.
Tested with latest changes in `v8/contrib` branch.

![image](https://user-images.githubusercontent.com/2919859/88570641-b3b7be00-d03c-11ea-99e6-b30b0047df58.png)

It's quite unfortunate that you as an administrator is able to lock yourself out of the system 🙈 

Here is a demonstration - if saving after these steps I can't access these sections again via UI.

![2020-07-27_19-51-30](https://user-images.githubusercontent.com/2919859/88574524-c1704200-d042-11ea-8edd-4e3343fe59d2.gif)

I think the super user (user id -1) should always have access to users section and should not be able to remove this.

bjarnef:
This is how it looked in Umbraco v8.4.0

![2020-07-27_19-59-12](https://user-images.githubusercontent.com/2919859/88575243-dbf6eb00-d043-11ea-87b1-62ef4afd0cb2.gif)

bjarnef:
It seems `Utilities.copy` doesn't copy the items in the array to `currentSelection` as `angular.copy` did.


**Now**

```
function openSectionPicker() {
            var currentSelection = [];
            
            Utilities.copy(vm.userGroup.sections, currentSelection);

            var sectionPicker = {
                selection: currentSelection,
                submit: function (model) {
                    vm.userGroup.sections = model.selection;
                    editorService.close();
                },
                close: function () {
                    editorService.close();
                }
            };
            editorService.sectionPicker(sectionPicker);
}
```

![image](https://user-images.githubusercontent.com/2919859/88577235-fda5a180-d046-11ea-8c98-75ef9bb054e7.png)


**Before**

```
function openSectionPicker() {
            var currentSelection = [];
            
            angular.copy(vm.userGroup.sections, currentSelection);

            var sectionPicker = {
                selection: currentSelection,
                submit: function (model) {
                    vm.userGroup.sections = model.selection;
                    editorService.close();
                },
                close: function () {
                    editorService.close();
                }
            };
            editorService.sectionPicker(sectionPicker);
}
```

[...]

PR #7950, opened 2020-04-11
Replace angular.copy with Utilities.copy

BatJan:
### Prerequisites

This PR fixes https://trello.com/c/kiVAk6fb/1-replaces-instances-angularcopy

If there's an existing issue for this PR then this fixes 

### Description
I've replaced every relevant instance of `angular.copy` with `Utilties.copy`

PR #8488, opened 2020-07-27
Ensure replacement of angular.copy still works with second parameter…

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8487

### Description
At the moment when opening section picker from user group, none of the current sections where preselected. This happens because `currentSelection` here always is empty.

https://github.com/umbraco/Umbraco-CMS/blob/bd26cb36ecce2bbe86f11b61dd1a01c314378c89/src/Umbraco.Web.UI.Client/src/views/users/group.controller.js#L104

The `Utilities.copy` works as `angular.copy` but unlike `angular.copy` it doesn't accept a second optional paramter to copy to a destination object/array.

This was cause by the changes in https://github.com/umbraco/Umbraco-CMS/pull/7950 where it a few places use the second optional parameter.

Note the issue also affected user picker in a user group, so when picking new users in a group (and you forget to select the existing users) they are cleared from the list, which could be problematic since these users no longer has access to what they used to have access to.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/utilities.js
+++ b/src/Umbraco.Web.UI.Client/src/utilities.js
@@ -18,7 +18,7 @@
     /**
      * Facade to angular.copy
      */
-    const copy = val => angular.copy(val);
+    const copy = (src, dst) => angular.copy(src, dst);
 
     /**
      * Equivalent to angular.isArray


nul800sebastiaan:
Thanks so much for digging into this one @bjarnef - nasty bug and lovely simple fix, well done! ⭐ 

I'll make sure it's cherry picked into 8.7.0!

PR #8490, opened 2020-07-28
7731 bye bye underscore => _.each pt2 => content overlays

nathanwoulfe:
### Prerequisites

### Description
This is an addition to #8475, splitting into multiple PRs to avoid giant merges. These changes remove underscore from the content overlay controllers (the ones used for the publish/save/unpublish overlays). Also fixes the sort order of variants displayed in the overlays (were previously in reverse alphabetical order, now in alpha order).

Change replace all underscore methods with native JS, and makes some general improvements (avoids iterating collections multiple times where possible, syntax fixes etc). Functionally, nothing has changed.

Testing requirement is to verify everything still works. Can save, publish, save+publish, unpublish, schedule with variant and invariant content. Changes are pretty safe and predictable as most relate to iterating the variants collection, which is always an array even if content is invariant.

There's a lot of duplication across these controllers, as there's a lot of common functionality (sorting, filtering) which would be really good to refactor. Hypothetically (😃) Typescript would make this more manageable as a variant object could become a variant class with its own methods, so duplicated controller functions would no longer be required to check properties as the variant would provide those values...

````javascript
export class Variant {
  // lots of variant properties
  language: VariantLanguage;
  segment: VariantSegment;
  state: string;

  get isMandatory(): boolean {
    return this.language && this.language.isMandatory && this.segment;
  }
}

// elsewhere
const mandatoryVariants = vm.variants.filter(v => v.isMandatory());
````

Oh so lovely!

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/overlays/save.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/overlays/save.controller.js
@@ -58,57 +58,47 @@
         function onInit() {
             vm.variants = $scope.model.variants;
             vm.availableVariants = vm.variants.filter(saveableVariantFilter);
+            vm.isNew = vm.variants.some(variant => variant.state === 'NotCreated');
 
-            if(!$scope.model.title) {
-                localizationService.localize("content_readyToSave").then(function(value){
+            if (!$scope.model.title) {
+                localizationService.localize("content_readyToSave").then(value => {
                     $scope.model.title = value;
                 });
             }
 
-            _.each(vm.variants,
-                function (variant) {
+            vm.variants.forEach(variant => {
 
-                    //reset state:
-                    variant.save = false;
-                    variant.publish = false;
+                //reset state:
+                variant.save = variant.publish = false;
+                variant.isMandatory = isMandatoryFilter(variant);
 
-                    variant.isMandatory = isMandatoryFilter(variant);
-
-                    if(variant.state !== "NotCreated"){
-                        vm.isNew = false;
-                    }
-                });
-
-            _.each(vm.variants,
-                function (variant) {
-                    if(vm.isNew && hasAnyData(variant)){
-                        variant.save = true;
-                    }
-                });
+                if(vm.isNew && hasAnyData(variant)){
+                    variant.save = true;
+                }
+            });
 
             if (vm.variants.length !== 0) {
+                        
+                //ensure that the current one is selected
+                var active = vm.variants.find(v => v.active);
+                if (active) {
+                    active.save = true;
+                }
 
-                _.find(vm.variants, function (v) {
-                    if(v.active) {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/Umbraco.Web.UI.Client/src/views/content/overlays/save.controller.js [hunk 1]
```diff
@@ -58,57 +58,47 @@
...
+                vm.availableVariants.sort((a, b) => {
                     if (a.language && b.language) {
-                        if (a.language.name > b.language.name) {
+                        if (a.language.name < b.language.name) {
                             return -1;
                         }
-                        if (a.language.name < b.language.name) {
+                        if (a.language.name > b.language.name) {
                             return 1;
                         }
```
nul800sebastiaan, line 95: I'm seeing this reversal a few times @nathanwoulfe - that's scary! Can you explain? I don't get why it's the other way around now..

src/Umbraco.Web.UI.Client/src/views/content/overlays/save.controller.js
```diff
@@ -58,57 +58,47 @@
...
                     }
                     if (a.segment && b.segment) {
-                        if (a.segment > b.segment) {
+                        if (a.segment < b.segment) {
                             return -1;
                         }
-                        if (a.segment < b.segment) {
+                        if (a.segment > b.segment) {
```
nul800sebastiaan, line 101: I'm seeing this reversal a few times @nathanwoulfe - that's scary! Can you explain? I don't get why it's the other way around now..

nathanwoulfe: Hey Seb, I'll double check to make sure I wasn't dreaming, but this change fixes the sort order of the variants listed in the save/publish overlay. They were previously in reverse alpha order, this flips them to ascending order.

nathanwoulfe: Would be nice actually to add this as a filter, given the same code is being used in a few places

nul800sebastiaan: @nathanwoulfe There's a lot of repetition going on here for sure and it's super fragile that this code is copy/pasted all over the place!

nathanwoulfe: Might look into adding some filters. Other option might be to import common stuff as es6 modules to avoid the angular coupling. 

I'll see what I can make work.

nul800sebastiaan: >  fixes the sort order of the variants listed in the save/publish overlay

Well whaddayaknow! That totally fixes it haha! 👍  


I'll merge this for now, refactoring can be done in a new PR! 🎉

PR #8492, opened 2020-07-28
Re-style the "more" button in editor app navigation

kjac:
### Prerequisites

### Description

This is an alternate approach to fixing #8486.

As-is the "more" button in the editor app navigation looks rather off when you have content apps and not enough space for them all:

more-button-before-1

more-button-before-2

This PR re-applies the styles for the "more" button that were lost in #8132. Once applied the "more" button looks like it used to:

more-button-after-1

more-button-after-2


BatJan:
Yet another fix proposed in #8500 - Let's keep the discussion there or in #8486 😃 - Just adding this comment for reference.

nul800sebastiaan:
Cheers! Picking #8500 over this one! 👍

PR #8493, opened 2020-07-28
Block list editor - settings validation

Shazwazza:
**Ensure that https://github.com/umbraco/Umbraco-CMS/pull/8382 is reviewed and merged first!**

This ensures that server validation for the settings part of a block works and visibly shows where the error is to the end user.

* Fixes val-sub-view. This directive I believe has been broken since 8.0 but we didn't realize it. This is the directive that sets a hasError value on a model in order to show that a content app tab has an error (red). It doesn't use normal validators to do this because a content app's label and it's view are separate so it relies on setting model data directly. It's a quirky directive that works in conjunction with umb-editor-sub-view or umb-editor-sub-views and valFormManager. Prior to 8.0 we didn't have umb-editor-sub-views (plural) and this was working as expected however with the umb-editor-sub-views this no longer worked because it relies upon an inherited $scope and the value it was relying on was $scope.model, however with umb-editor-sub-views that model isn't the model used to display the tabs, instead it is $scope.subView. Now this directive takes a value to be told which $scope property to bind to.
* Enhances val-server-match to allow more precise validation matching. It still matches on exact, prefix, suffix or contains but now we can specify a different validation key which is assigned to the ng-form it's attached to. Previously we'd just use a validation key like "valServerMatch" which would be the $error key for the form, but we need more control over knowing if it's a content or a settings error so we can highlight the settings cog if there's a settings error. So now we can specify different validation keys in the value of val-server-match. I've added ngdocs about this.
* Moves the logic that was swapping/replacing the content apps from the scaffold to a centralized place which now occurs when we get the scaffold. Previously this logic was hidden within an infinite editor controller and also mean we created a separate object reference to the data.
* Removes the "_private" methods from the blockeditormodelobject.service.js  since they should just be functions at a higher scope, no need for "_private" style declarations and exposing a public API if we don't want to
* Removes the TestData reference from Umbraco.Web.UI so we don't end up shipping that DLL and bogus.dll with the build output
* Adds server side validation to the min/max blocks and fixes displaying the error in the html
* Adds some logging for when data is skipped
* Ensures that settings data is validated on the server side
* Ensures that the 'settings' cog is highlighted red when there is settings errors in inline mode and when in infinite mode ensures that the content app icons for either content/settings are highlighted when they are in error

## Testing

* Prob easiest to look at the slack video https://umbraco.slack.com/archives/CAF1E7HBK/p1595901372136600 and reproduce those test cases
* You will need to have server side validation applied to fields on both content and settings then test server side validation for both content and settings and in both inline and infinite editing modes.
* Test min/max validation. By default this will just kick in on the client side but you'll want to test server side validation, to do that comment out the JS code in the `onAmountOfBlocksChanged` in the `// validate limits:` block of code in  `umbBlockListPropertyEditor.component.js`, then the min/max validation is server side and should work


warrenbuckley:
Some failed JS unit tests that need fixing up please @Shazwazza 

image

warrenbuckley:
Code seems fine to me - off to test this out

warrenbuckley:
Testing seems fine - if you can investigate failing JS tests please @Shazwazza then this is all good to go 👍

Shazwazza:
@warrenbuckley have just pushed fixes for tests if you want to take a look. Cheers!

warrenbuckley:
Congrats @Shazwazza for fixing it - all merged in now

Congrats

src/Umbraco.Web.UI.Client/src/common/directives/validation/valservermatch.directive.js
```diff
@@ -40,41 +52,63 @@ function valServerMatch(serverValidationManager) {
...
-                else {
-                    formCtrl.$setValidity('valServerMatch', true);                    
+            function bindCallback(validationKey, matchVal, matchType) {
+
+                if (Utilities.isString(matchVal)) {
+                    matchVal = [matchVal]; // normalize to an array since the value can also natively be an array
                 }
+
+                // match for each string in the array
+                matchVal.forEach(m => {
```
nielslyngsoe, line 62: I've experience that 'matchVal' is undefined, could you please see if it make sense to add a undefined check? :-) @Shazwazza

PR #8494, opened 2020-07-28
Don't auto-expand all content anchors on small screens

kjac:
### Prerequisites

### Description

While fixing the "more" button styling in the content app navigation, I noticed how the content anchor list kept expanding on small screens, making the content apps behind it unreachable:

app-navigation-anchor-toggle-before

Turns out the CSS selector that expands the anchor list is a tad too greedy. This PR narrows it down a tad, and once applied the content app navigation works somewhat nicer on small screens:

app-navigation-anchor-toggle-after-1

...and of course it still works on larger screens:

app-navigation-anchor-toggle-after-2


nul800sebastiaan:
Sweet! I don't have that many apps so didn't see this before, thanks for that @kjac!

PR #8495, opened 2020-07-28
Don't allow the grid to steal focus on init

### LINKED ISSUE #8324: Focus issue when creating new page w. grid in backoffice
When you have setup a new page with a grid-datatype with only one Grid Layout and one Rows configuration, the grid-datatype takes focus instead of the normal "Enter a name..." field.

focusbug-griddemo

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.2

Reproduction
------------

### Steps to reproduce

* add a new dataType of type "Grid layout"
* remove Grid layouts so that there is only one
* remove Rows configrations so that there is only one
* save datatype
* insert datatype on doctype
* go to Content section and add a new node (of that doctype)

datatype-gridtype-setup

### Expected result

I would expect that the "Enter a name..." field would have focus

### Actual result

Grid area gets focus

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8324

### Description

_See also #8324 for a brilliant issue description_

If a Grid configuration has only one layout and one row configuration, the property editor will automatically initialize itself with one layout + row. This makes sense. However, in combination with the excellent accessibility effort in #6804, this causes the grid to steal focus on init - most noticeably when creating new content:

grid-init-focus-before

This PR ensures that the auto-focus is not performed on init, but is retained on subsequent row additions to preserve the accessibility:

grid-init-focus-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js
@@ -405,14 +405,16 @@ angular.module("umbraco")
 
                 eventsService.emit("grid.rowAdded", { scope: $scope, element: $element, row: row });
 
-                // TODO: find a nicer way to do this without relying on setTimeout
-                setTimeout(function () {
-                    var newRowEl = $element.find("[data-rowid='" + row.$uniqueId + "']");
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Nice one @kjac! Thanks for this! 👍

PR #8496, opened 2020-07-28
Fetch membergroup by guid

bjarnef:
### Prerequisites

### Description
This PR adjust previous merged PR https://github.com/umbraco/Umbraco-CMS/pull/7528 to fetch member group by Guid (uniqueId in umbracoNode) table.

It seems many of the repository classes use a `GetBaseWhereClause()` which query by id. Some of these use the constant for table name, while others don't.
https://github.com/umbraco/Umbraco-CMS/search?p=3&q=GetBaseWhereClause&unscoped_q=GetBaseWhereClause

I have updated this in https://github.com/umbraco/Umbraco-CMS/commit/d5d5d077c5d5bfc8634fc57745429cd81300f6e1 (but we could probably do with the other repositories as well).

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs
@@ -69,7 +69,7 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
 
         protected override string GetBaseWhereClause()
         {
-            return "umbracoNode.id = @id";
+            return $"{Constants.DatabaseSchema.Tables.Node}.id = @id";
         }
 
         protected override IEnumerable<string> GetDeleteClauses()
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs
@@ -119,7 +119,7 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
         {
             var sql = GetBaseQuery(false);
             //sql.Where(GetBaseWhereClause(), new { uniqueId = uniqueId });
-            sql.Where("umbracoNode.uniqueId = @uniqueId)", new { uniqueId = uniqueId.ToString() });
+            sql.Where("umbracoNode.uniqueId = @uniqueId", new { uniqueId });
 
             var dto = Database.Fetch<NodeDto>(SqlSyntax.SelectTop(sql, 1)).FirstOrDefault();
 
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs
@@ -118,7 +118,6 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
         public IMemberGroup Get(Guid uniqueId)
         {
             var sql = GetBaseQuery(false);
-            //sql.Where(GetBaseWhereClause(), new { uniqueId = uniqueId });
             sql.Where("umbracoNode.uniqueId = @uniqueId", new { uniqueId });
 
             var dto = Database.Fetch<NodeDto>(SqlSyntax.SelectTop(sql, 1)).FirstOrDefault();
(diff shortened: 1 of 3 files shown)


bjarnef:
At the moment the method in `MemberGroupRepository` generate the following sql query, but it seems I have a syntax error:

```
SELECT [umbracoNode].[id] AS [NodeId], [umbracoNode].[uniqueId] AS [UniqueId], [umbracoNode].[parentId] AS [ParentId], [umbracoNode].[level] AS [Level], [umbracoNode].[path] AS [Path], [umbracoNode].[sortOrder] AS [SortOrder], [umbracoNode].[trashed] AS [Trashed], [umbracoNode].[nodeUser] AS [UserId], [umbracoNode].[text] AS [Text], [umbracoNode].[nodeObjectType] AS [NodeObjectType], [umbracoNode].[createDate] AS [CreateDate]
FROM [umbracoNode]
WHERE (([umbracoNode].[nodeObjectType] = @0))
AND (umbracoNode.uniqueId = @1))
```

The current one to get by id use the following method:

```
protected override string GetBaseWhereClause()
{
     return $"{Constants.DatabaseSchema.Tables.Node}.id = @id";
}
```

Could we extend this with an overload to use `uniqueId` instead with `id` as default?

bjarnef:
@nul800sebastiaan can you have a look at this? 😁 



we could probably re-use an overload of `GetBaseWhereClause()`
https://github.com/umbraco/Umbraco-CMS/blob/5992f8ecbeda0117f5264a617500afc09218938f/src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs#L70-L73

or use `Constants.DatabaseSchema.Tables.Node` in the following query:

https://github.com/umbraco/Umbraco-CMS/blob/5992f8ecbeda0117f5264a617500afc09218938f/src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs#L121

nul800sebastiaan:
Awesome, this is perfect! 👍

bjarnef:
@nul800sebastiaan I noticed we have a similar use of `GetMany()` in `RelationTypeRepository` here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Persistence/Repositories/Implement/RelationTypeRepository.cs#L32-L42

Isn't it a potentially issue as well?

//cc @Shazwazza

src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs
```diff
@@ -115,6 +115,16 @@ protected override void PersistUpdatedItem(IMemberGroup entity)
             entity.ResetDirtyProperties();
         }
 
+        public IMemberGroup Get(Guid uniqueId)
+        {
+            var sql = GetBaseQuery(false);
+            sql.Where("umbracoNode.uniqueId = @uniqueId", new { uniqueId });
```
Shazwazza, line 121: This could be strongly typed with `Where<NodeDto>` like we do in lots of places

Shazwazza:
@bjarnef you can see the comments in those repos like `// use the underlying GetAll which will force cache all content types`

because these repos have a different cache policy where they cache everything so everything does a GetAll() and then post filters because we know the entire data set is in memory. In some repositories they override CreateCachePolicy and return FullDataSetRepositoryCachePolicy to indicate this.

bjarnef:
@Shazwazza ahh okay. In this case for member groups it is direct SQL-queries without caching on server side.

src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs
bjarnef: @Shazwazza yes it could, but I think there is several other places where the SQL-queries can be re-written with strongly typed syntax. Currently we have many of these or combination of strongly typed query combined with string SQL-queries where it could use `.Where<SomeDto>` or `.WhereIn<SomeDto>`.

Could be enhanced further in some other PR's to clean this up 👍 

At least I replaced the "umbracoNode" table with the constant value 😁

Shazwazza: yep, just wanted to mention it :)

Shazwazza:
yep that is fine here, the repository by default provides caching per id but since it's a different type of id, a custom cache is necessary but this all starts getting a little messy since we have to deal with cache refreshers, etc... I don't think the amount that a member group will be queried by GUID will have a huge overhead. We can deal with it later if necessary.

src/Umbraco.Core/Persistence/Repositories/Implement/MemberGroupRepository.cs
bjarnef: @Shazwazza I have updated this as part of https://github.com/umbraco/Umbraco-CMS/pull/9467 🎄

PR #8500, opened 2020-07-28
umb-ellispsis-component: Fix visuals when used in the content app section

BatJan:
### Prerequisites

### Description
@nul800sebastiaan This is my approach to fix the issue filed by @kjac in #8486 and the alternate approach in #8492 

I have passed the css classes used on the "Content app" buttons, which fixes the height issue and also fixed the general styling of the "more" button actually using the "css-class" attribute provided by the directive.

Furthermore in order to vertically align the ellispsis and the text I've added a "content" wrapper around the 2 existing spans in order to be able to do this in the directive itself. This does not affect the usage where there is no text to display.

I can't see any downsides to this approach since the only change added to the component is a span that wraps the icon and text and otherwise it's just accepting the passed css class, which was the intention from the very beginning. Please let me know if you think I'm missing something 😃

This is what it looks like when these changes are applied

directive-fix

**UPDATE** - Fixed the issues with this PR that @kjac pointed out in his comment below 👍 

ellipsis-component-content-app-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbuttonellipsis.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbuttonellipsis.directive.js
@@ -29,6 +29,7 @@
 @param {boolean} showText Set to <code>true</code> to show the text. <code>false</code> by default
 @param {domElement} element Highlights a DOM-element (HTML-selector)
 @param {string} state Set the initial state of the component. To have it hidden use <code>hidden</code>
+@param {string} mode Set the mode, which decides how to style the component. Currently only "default" and "tab" are supported
 **/
 
 (function () {
@@ -82,7 +83,8 @@
             color: "@?",
             showText: "<?",
             element: "@?",
-            state: "@?"
+            state: "@?",
+            mode: "@?"
         }
     };
 
--- a/src/Umbraco.Web.UI.Client/src/less/components/buttons/umb-button-ellipsis.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/buttons/umb-button-ellipsis.less
@@ -15,7 +15,7 @@
         justify-content: center;
     }
 
-    .umb-button-ellipsis__content-app,
+    .umb-button-ellipsis--tab,
     .umb-tour-is-visible .umb-tree &,
     &:hover,
     &:focus{
@@ -40,7 +40,7 @@
         color: inherit;
         flex-basis: 100%;
 
-        .umb-button-ellipsis__content-app & {
+        .umb-button-ellipsis--tab & {
             margin: 0 0 7px
         }
     }
@@ -51,7 +51,7 @@
         line-height: 1em;
         flex-basis: 100%;
 
-        .umb-button-ellipsis__content-app & {
+        .umb-button-ellipsis--tab & {
             position: absolute;
             right: 0;
             left: 0;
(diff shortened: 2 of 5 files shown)


kjac:
Hi @BatJan, thanks for weighing in :heart:

What you're doing here is along the lines of what I did in #8492. From what I can see in your screenshot, and from what I gather when reading through your changes, there are a few issues:

1. There's a vertical misalignment between the "More" label and the labels of the visible content apps. 
2. I suspect the "More" button isn't marked as active (pink-ish bottom border/indicator) when a content app behind the "More" button is active. 

Both of these are solved in #8492. 

The second issue isn't hard to fix, you can add a conditional `is-active` class to the `css-class` directory parameter (see [here](https://github.com/umbraco/Umbraco-CMS/pull/8492/files#diff-dfb0b07dc0f08a455bb925bca3c84448R25)).

However. The first issue requires you to target one or more of the CSS classes within the directory (`.umb-button-ellipsis__icon`, `.umb-button-ellipsis__text` from the outer scope (`.umb-sub-views-nav-item`), in order to achieve the correct vertical alignment (and possibly other styling requirements to make the "More" button look and behave like the visible content apps). 

This is where I foresee regression issues. The `umb-button-ellipsis` directive has its own styling, and anyone making changes to that in order to achieve something where the directive is very visible, might overlook the implications of its usage in the content app navigation. After all, the "More" button is rarely visible - you'll only see it if you have content apps installed, or if you are testing on a very small screen. 

This is why I have mixed feelings about #8492, and (as it's essentially the same) this PR as well. I'm usually all for abstracting and refactoring where possible, but in this case I really do believe we'd be better of having a dedicated implementation of the "More" button for the content app navigation.

BatJan:
@kjac Oops, yes I missed a few details with this PR indeed - Those are fixed now though 😅 Thank you for pointing them out. Icon and text now aligns and the hover/active states acts as expected as well.

Forgive me if I misunderstand your concern but in regard to the concerns you have about targeting icon and text. The way I have done it it's done within the scope of the component using CSS class modifiers to overwrite some of the defaults when it's being used in the context of a content app.

But yes it might still be easy to overlook any potential impact when modifications are made to the component in the future - I'm a bit torn about this... Because the underlying pattern is identical so I would prefer to keep it within one component and while it's true the text is rarely visible with the current uses I don't know if that will change for some scenarios in the future.

If we decide to keep the usage of the ellipsis component then I would prefer keeping things in scope of the component like I have proposed with this PR - If the decision ends up being to roll back like you suggest in #8486 then I'm OK with that 😄 - Please just change that line 24 that reads `<div class="umb-sub-views-nav-item-more__icon">&bull;&bull;&bull;</div>` to `<span class="umb-sub-views-nav-item-more__icon">&bull;&bull;&bull;</span>` since div's inside buttons are not valid 👼

kjac:
@BatJan this is indeed not a hill worth dying on. My sole concern is that of core stability. And since the original PR introducing the ellipsis directive made it through without anyone noticing the "More" button got me worried, thus why I did the revert in #8486. 

I'll leave the decision in @nul800sebastiaan's capable hands 👍

BatJan:
@kjac Yeah I'm thinking the same - And your point and concern are very valid indeed 😃 🙌 

I have updated this PR adding a "mode" parameter, which makes it cleaner so we don't need to remember to pass certain css-class modifiers but can set the mode instead and handle it inside the component should this end up being the fix that is chosen by @nul800sebastiaan 👼

nul800sebastiaan:
Thanks for the discussion gentlemen! I feel like it will be pretty "easy" to remember that when changes come in to ellipsis that the component is used in different ways. I love the DRY approach and thinks it's pretty nice to have different "modes".

I'll merge this one and reject the other two from Kenn. Thanks again!!

PR #8501, opened 2020-07-28
Remove usages of angular.forEach in users and groups editing

### LINKED ISSUE #7718: Start decoupling from AngularJs
Might be controversial, but here we go...

I'm proposing we start refining the backoffice javascript to remove/reduce the coupling to AngularJs. The quickest win here (IMO) is to remove all usages of [Angular functions](https://docs.angularjs.org/api/ng/function).

Most (if not all) of these exist as native JS methods in ES6, or are provided in Lodash. Umbraco currently takes a dependency on Underscore, of which Lodash is a superset. Benefit of Lodash is that it allows individual functions to be included, greatly reducing the size of the dependency. I don't see any issue in replacing a framework dependency with a library dependency.

So I guess this is two-fold - replace AngularJs fns with vanilla JS where possible, else import the appropriate Lodash method. At the same time, replace Underscore with vanilla, falling back to Lodash.

Why? It will need to happen eventually. Regardless of what direction the backoffice takes, the AngularJs functions will stop working. These changes can be addressed now as incremental improvements rather than leaving until later as part of sweeping backoffice changes.

## TODO:

- [x] add utility javascript bits - resolved in #7738, updated in #7975, #8014 
- [x] replace instances of angular.noop() - PR #7738 
- [x] replace instances of angular.copy() - PR #7950
- [x] replace instances of angular.isFunction() - PR #7861
- [x] replace instances of angular.element() - PR #7951 
- [x] replace instances of angular.equals() - PR #7860
- [ ] replace instances of angular.extend()
- [x] replace instances of angular.forEach() - PR #7928
- [x] replace instances of angular.fromJson() - PR #7952
- [x] replace instances of angular.isArray() - PR #7934 
- [x] replace instances of angular.isDefined() - PR #7927
- [x] replace instances of angular.isNumber() - PR #7925
- [x] replace instances of angular.isObject() - PR #7931
- [x] replace instances of angular.isString() - PR#7929
- [x] replace instances of angular.isUndefined() - PR #7856 
- [x] replace instances of angular.toJson() - PR #7924
- [ ] profit!

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #7718 (in part)

### Description

Continuing the effort to minimize the angular dependency, I have rewritten all usages of angular.forEach with the native equivalent in the users and groups section.

## Testing this PR

Basically there should be no noticeable difference when applying this PR; users and groups editing and management should "just work". Here are a few things to test and try out:

### Groups

image

image

image

### Users

image

image

image

image

image

image

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/users/group.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/users/group.controller.js
@@ -185,7 +185,7 @@
          * however the list to display the permissions isn't via the dictionary way so we need to format it
          */
         function formatGranularPermissionSelection() {
-            angular.forEach(vm.userGroup.assignedPermissions, function (node) {
+            vm.userGroup.assignedPermissions.forEach(function (node) {
                 formatGranularPermissionSelectionForNode(node);
             });
         }
@@ -193,8 +193,8 @@
         function formatGranularPermissionSelectionForNode(node) {
             //the dictionary is assigned via node.permissions we will reformat to node.allowedPermissions
             node.allowedPermissions = [];
-            angular.forEach(node.permissions, function (permissions, key) {
-                angular.forEach(permissions, function (p) {
+            Object.values(node.permissions).forEach(function (permissions) {
+                permissions.forEach(function (p) {
                     if (p.checked) {
                         node.allowedPermissions.push(p);
                     }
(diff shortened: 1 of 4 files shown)


nul800sebastiaan:
[click], [click], [click] - looking good!

PR #8502, opened 2020-07-28
Preserve superuser access

bjarnef:
### Prerequisites

### Description
At the moment the user sections have some issues regarding administrator access as mentioned in 
https://github.com/umbraco/Umbraco-CMS/issues/8487

Highlight of the current issues

- It is possible to remove users section from Administrators ("admin") group which mean if logged in as super admin (user id -1), the user is logged out and when logging back in the user has no longer access to users section and there is no way to grant access to users section again (at least not if no other users has access to users section) or without modifying data in database.

- It is possible to remove (super) user from Administrators group under Users -> Groups

- The (super) user can remove Administrators group under user details page

This PR adjust this so the remove button isn't shown these places for super user and furthermore for Administrators group doesn't allow to remove the following sections: Content, Media, Settings, Users, Members.

The prevent removing these by mistake and saving the user/group. However it is still possible via userpicker and sectionpicker to remove these existing sections/users. It would be great if sectionpicker and userpicker supported `filter` and maybe `filterCssClass` similar to treepicker, so it can disable selection/deselection of the items in the overlays.

https://github.com/umbraco/Umbraco-CMS/blob/0bd9e3ca9955844389556ce512fa0aab181703fb/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/treepicker/treepicker.controller.js#L448-L500

But it think this is a bit better than the current behavior and prevents that you by mistake remove access to mainly users section.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/users/group.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/users/group.controller.js
@@ -19,6 +19,7 @@
         vm.openUserPicker = openUserPicker;
         vm.removeSelectedItem = removeSelectedItem;
         vm.clearStartNode = clearStartNode;
+        vm.allowRemoveUserFromGroup = allowRemoveUserFromGroup;
         vm.save = save;
         vm.openGranularPermissionsPicker = openGranularPermissionsPicker;
         vm.setPermissionsForNode = setPermissionsForNode;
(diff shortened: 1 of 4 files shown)


nul800sebastiaan:
Maybe I didn't make it clear but I want everybody to be able to lock down the "superuser" account, we should not preserve anything at all. 🙂 

Sorry if I didn't make it clear enough, but this PR is not what we should do.

bjarnef:
Okay, but even without using super user, users who belongs to "admin" group are able to remove "users" section from this group and thus prevent any administrators from accessing "users" section.

nul800sebastiaan:
Yes, we want other admins to limit any admin user. 👍 Everything is working as expected.

PR #8504, opened 2020-07-29
Fix the alignment of the "delete property" button

kjac:
### Prerequisites

### Description

Somewhere along the lines, the "delete property" button on the content type builder has gone and got itself a bit crooked:

image

This PR fixes the button so it goes back to being centered under the "property settings" button:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-group-builder.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-group-builder.less
@@ -355,7 +355,7 @@ input.umb-group-builder__group-title-input:disabled:hover {
     position: relative;
     margin: 5px 0;
     
-    > button {
+    button {
         border: none;
         
         font-size: 18px;


nul800sebastiaan:
Yikes, that does look a bit wonky. Love 1 character fixes! 👍 Cheers!

nul800sebastiaan:
Ah, just noticed another little annoyance after this @kjac 

image

The confirm/delete actions should be next to the recycle bin instead of on top of it. Now.. let's see who can fix it faster, you or @bjarnef 😁😉

bjarnef:
@nul800sebastiaan I fixed some issues regarding the styling of tags editor.
https://github.com/umbraco/Umbraco-CMS/pull/8528

might fix some of the styling, but can have another look at this as well.

I would probably use `transform` here, so we don't need to adjust left position when/if sizes of these buttons change. The action button also used to be "circle" 😁  

image

PR #8506, opened 2020-07-29
Performance improvements to content type editor

kjac:
### Prerequisites

### Description

I stumbled across this while mucking about with the angular.forEach refactorings in #8505 ... the content type editor performs lookups for each and every property on load, in order to grab the corresponding datatype names and icons. 

This means an XHR is performed for each property on the content type, which obviously causes a great deal of requests to the server - the more properties, the more requests.

Here's a sample of the requests currently performed when editing a single document type:

image

And similarly for the built-in "Image" media type:

image

On top of increasing the requests made to the server, these lookups also cause additional wait time for the editors. And it's a bit silly, really, because we have all the data on hand when mapping the content type to the editor; we're just not returning it on first load, and thus we're forcing the editor to fetch it in these subsequent lookups.

This PR adds the datatype name and icon to the output when loading content types (document/media/member types) from the server, thus removing the need to load these data as subsequent lookups. 

When the PR is applied, the number of requests go down in a magnitude equal to the number of properties on the content type being edited. Here's a screenshot of the request made for same document type as shown above:

image

...and here are the requests for the built-in "Image" media type:

image


nul800sebastiaan:
This works great, though I got really stumped on how to build this.. I kept running into:

image

But the "code" was just code comments!

image

Adding those two properties, however.. worked and it stopped complaining. It's not something I've ever encountered before, this magic seems to do the trick though 🙈 

Thanks @kjac! 👍

nielslyngsoe:
@kjac WOW, this is great, thanks a lot for spotting this and doing something about it!

PR #8509, opened 2020-07-29
Query builder buttons

bjarnef:
### Prerequisites

### Description
I have replaced a few anchor elements with `<button>` elements in template query builder.
Furthermore the samle result didn't use the icon from the json, so I changed that and use `icon-document` since `icon-file` doesn't exist.

It would be nice if it could use the respective icons for the nodes, but `IPublishedContentType` doesn't have a `Icon` property like `IContentType`. So I guess that would require looking up the specific content types from the results.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/querybuilder/querybuilder.html
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/querybuilder/querybuilder.html
@@ -71,7 +71,7 @@
 
                                     <umb-dropdown ng-if="vm.propertyFilterOpen[$index]" on-close="console.log(1);vm.propertyFilterOpen[$index] = false">
                                         <umb-dropdown-item ng-repeat="property in vm.properties">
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Thanks @bjarnef! Yeah for the icons we'd need some more info on the model, not sure if we want to add that though. A lookup per  item seems a bit overkill.

bjarnef:
@nul800sebastiaan yeah, or not necessarily per item. For each item `x.ContentType.Alias` is available, so it could find distinct document types and lookup the icons for these, but still a lookup per document type (potentially 20 since results are limited to 20 and each could be different document type).

Not sure if it would make sense `Icon` was available on `ContentType` on `IPublishedContentType` like in `IContentType`?
Maybe @Shazwazza knows?

Shazwazza:
"published" models are purely for front-end and they live 100% in memory, we can't add arbitrary values there like icons that aren't explicitly required for the front-end else we fill up more memory unnecessarily ... plus that would require some large breaking changes.

bjarnef:
Thanks for the explanation @Shazwazza 👍

PR #8510, opened 2020-07-29
Replace splitview close button with button element

bjarnef:
### Prerequisites

### Description
This PR replace the close button when splitview is opened with `<button>` element and futhermore use `umb-icon`.

It also removes a unnecessary use of `ng-if`.

image

It does however seems to be a bit strange I need to set `icon-delete` twice. Maybe when setting `icon` attribute it should set this as class in the `umb-icon` component.

```
<umb-icon icon="icon-delete" class="icon-delete"></umb-icon>
```

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js
@@ -32,7 +32,7 @@ This format is only used in the iconpicker.html
 (function () {
     "use strict";
 
-    function UmbIconDirective($http, $sce, iconHelper) {
+    function UmbIconDirective(iconHelper) {
 
         var directive = {
             replace: true,
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js
@@ -44,7 +44,8 @@ This format is only used in the iconpicker.html
             },
 
             link: function (scope) {
-                if (scope.svgString === undefined) {
+                
+                if (scope.svgString === undefined && scope.icon !== undefined) {
                     var icon = scope.icon.split(" ")[0]; // Ensure that only the first part of the icon is used as sometimes the color is added too, e.g. see umbeditorheader.directive scope.openIconPicker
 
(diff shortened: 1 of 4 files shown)


bjarnef:
I noticed @MMasey renamed the scope property from `iconName` to `icon`. In the component view it did however still use `iconName`.
https://github.com/umbraco/Umbraco-CMS/pull/8510/commits/47d0cdbfd4328b0ae7394a7e17830e94cc3be701

However with the following:

```
<umb-icon icon="icon-delete" class="icon-delete"></umb-icon>
```

chrome_2020-07-29_13-32-22

it always seems to render `svgString`, because `svgString` is undefined and end up here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js#L50

and continue to this:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js#L67

So in this case it end in this condition:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/umb-icon.html#L3

@MMasey is this only used as fallback in case the `getIcon()` request fails?
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/umb-icon.html#L4

MMasey:
Hi @bjarnef, that's correct. Essentially the icon directive will use the `iconHelper` to try and get the svg string for requested icon. If it can't return an svg string, it should fallback to the `<i>` element. This is to support the legacy icons.

My bad on missing the `iconName` to `iconName` change on the directive too 🤦‍♂️ I has noticed it when I was trying to get the transclude stuff to work, but simple forgot to push it.

bjarnef:
@MMasey okay, I have changed `iconName` to `icon` in the directive view as part of this PR.

nul800sebastiaan:
Thanks @bjarnef - you're creating a lot of overlap with different PRs introducing the same code, please be careful with that since I can't keep up with checking out and merging them. The overlap leads to merge conflicts that are getting hard to fix 😬 

In any case, this one is now good to go! Thanks again! 👍

PR #8512, opened 2020-07-29
Make sure the RTE always loads minified resources

### LINKED ISSUE #8328: TinyMCE loads non-minified plugins in non-debug mode and vice-versa
When running Umbraco on a live environment (with `debug="false"`), I see the following plugin is loaded when TinyMCE/the RTE is initialized in the back-office: `/umbraco/lib/tinymce//plugins/autoresize/plugin.js?umb__rnd=7de6bda4e77ab561e3b19a72a59cdb67`. This loads the non-minified version of the plugin (5,58 kB).

When running in debug mode, the minified version is loaded (2,03 kB - 63% smaller size), so this logic seems mixed up: `/umbraco/lib/tinymce//plugins/autoresize/plugin.min.js?umb__rnd=64f3d65f34f0673f8fc49a60d0d01780`.

## Umbraco version

I am seeing this issue on Umbraco 8.6.2.

Reproduction
------------
### Steps to reproduce
Switch the debug attribute in the web.config and check the loaded plugins in the developers tools network tab.

### Expected result
Load the normal version in debug mode and minified ones in live/release builds.

### Actual result
TinyMCE loads the minified version in debug mode and normal version on live/release builds.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8328 

### Description

As described in #8328 the RTE loads un-minified resources in release mode. As it turns out it's not only plugins but also themes that are loaded un-minified:

image

This issue _does_ indeed affect both normal and grid RTEs.

This PR forces TinyMCE to always load minified resources, no matter if Umbraco runs in debug or release mode. When applied the loaded resources are significantly smaller:

image

While I was at it I also redid the way the grid RTE loads TinyMCE resources to match that of the normal RTE (as implemented in #6147).

#### Testing this PR

After building the frontend (`gulp dev`), be sure to set `debug="false"` in web.config:

image

Then increment `version` in ClientDependency.config:

image

Now reload any content item with an RTE with the Network tab open in Chrome (or whatever browser), and verify that only minified TinyMCE resources are loaded.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/grid/grid.rte.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/grid/grid.rte.directive.js
@@ -1,5 +1,5 @@
 angular.module("umbraco.directives")
-    .directive('gridRte', function (tinyMceService, angularHelper, assetsService, $q, $timeout, eventsService) {
+    .directive('gridRte', function (tinyMceService, angularHelper, assetsService, $q, $timeout, eventsService, tinyMceAssets) {
         return {
             scope: {
                 uniqueId: '=',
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Looks good to me, cheers @kjac! 👍

PR #8515, opened 2020-07-29
Use umb-icon in listview layout selector

bjarnef:
### Prerequisites

### Description
This PR use `<umb-icon>` in the listview layout selector and removes a few unnecessary `prevent-default` on the button elements.

I also noticed a few missing icons, while looking at the listview.

### Document type properties

**Before**

chrome_2020-07-29_22-43-52

**After**

image

### Listview configuration

**Before**

chrome_2020-07-29_23-00-32

**After**

image


nul800sebastiaan:
Wonderful, you beat me to it as I was just now looking into those missing icons! Saves me some time, awesome work! 👍 ⭐

PR #8518, opened 2020-07-30
Show keyboard input focus on file property editors

kjac:
### Prerequisites

### Description

The property editors for file upload and image cropper do not show the keyboard input focus when you're tabbing through the editor interface:

file-upload-keyboard-focus-before

...but with this PR - they do 😄 

file-upload-keyboard-focus-after


nul800sebastiaan:
Yay!! Good job! 👍

PR #8519, opened 2020-07-30
Adjustments to umb-icon with custom font icons

bjarnef:
### Prerequisites

### Description
After merging PR https://github.com/umbraco/Umbraco-CMS/pull/5606 I have noticed some issues with custom font icons. At the moment it expect the font-family is `icomoon`, but it could be something else like `icomoonforms` shipped with Umbraco Forms.

Furthermore these fonts didn't look nice in the iconpicker.

**Before**

**After**

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js
@@ -32,7 +32,7 @@ This format is only used in the iconpicker.html
 (function () {
     "use strict";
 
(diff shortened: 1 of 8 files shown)


bjarnef:
@MMasey I noticed it had some issues, when using a custom font icons, e.g. from Umbraco Forms which is shipped with a few custom font icons, but could also be an external font package or just including a custom font icons in your project.

nul800sebastiaan:
Sweet that looks and feels a lot nicer, many thanks @bjarnef! 👍

Issue #6441, opened 2019-09-26
MediaPicker treats file as folder


roeermose:
It is not possible to select the media in the mediapicker when using belowconfiguration.
The media is treated as a folder eventhough it is a file.

### Specifics
Version 7.15.3
Chrome 76.0.3809.132
Checked in Firefox 67.0.4

Have been tested on a clean 7.15.3 by support with same results.

### Steps to reproduce

Create a doc-type, create property with MediaPicker editor, set "Pick multiple items" = true, "Pick only images" = false, "Disable folder select" = true
Create a media where you do not upload a file.
Create a node using the new doc-type and try to select the media.

### Expected result
I expected to be able to select the media.

### Actual result
The media was not selectable, and when you tried to select it, then you would be presented with an empty folder.

---
_This item has been added to our backlog [AB#7087](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7087)_

nul800sebastiaan:
Hey @roeermose - apart from the selected file not really having a file extension, this all works for me, not sure what I'm doing differently?

![image](https://user-images.githubusercontent.com/304656/65744349-6440f280-e0f8-11e9-9627-d0df6227b9c8.png)

roeermose:
Hey @nul800sebastiaan - have you set the specfic configuration for the MediaPicker (default works fine, but the specific configuration does not)

nul800sebastiaan:
Aha! Yeah, I did misread the configuration necessary, after updating it, I can't pick the empty item either. 

We'll have a look!

stevemegson:
In 7.15, the media picker's definition of "is a folder" is effectively just "has no umbracoFile value". The current test can't tell the difference between an item which has no `umbracoFile` property and one where the `umbracoFile` property has no file selected. Both will behave as folders.

I'd need to check, but my PR #6195 to support thumbnails from other properties may also handle having no file selected. I think it will make mediaPath null when there's no property and an empty string when the property has no selected file.

There may also be a javascript tweak needed make sure that an empty string as mediaPath is enough to make the item "not a folder".

andersbrohall:
Is this related?
[#7317](https://github.com/umbraco/Umbraco-CMS/issues/7317?fbclid=IwAR0ctuDwA7JfBwqfgd7EAy1S8sovKQq0OG_KMm0Jm85qr4NiP7GOLKPumjg)
:)

[...]

PR #8520, opened 2020-07-30
Fixing #6441 so empty media items are not detected as folders.

Shazwazza:
Fixing #6441 so empty media items are not detected as folders.

Few things involved here:

* hack implemented in EntityController for media items to ensure that an IsFolder metadata value is set based on whether the media type has an umbracoFile property type (basically the previously logic from < 7.15 for folder detection)
* when resolving media items from the entity controller (these are 'slim' media items), always pass this data through mediaHelper.formatMediaEntityData to ensure that isFolder and image/thumbnail properties are correctly set. Now we don't have to do this logic differently in all the places. This method has a few fallback mechanisms for backwards compat, etc... 

Testing: 

* See #6441 
* Create a media item file but don't upload a file, then try to pick this in a picker with images only = true
* Ensure all media pickers work correctly with all the different types of options
* Ensure all media grid views work correctly


Shazwazza:
@nul800sebastiaan I have assigned you as reviewer and tagged as community/review in hopes that this can get reviewed since people are still expecting a v7.x released right?

hfloyd:
@Shazwazza I wasn't able to get to this before, and now the Build seems to have disappeared. Can you direct me to the build and I will do a community review? Thanks!

Shazwazza:
@hfloyd if you can build from source the branch is this one https://github.com/umbraco/Umbraco-CMS/tree/v7/bugfix/6441-media-items-folders

else I've just kicked off another build: https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=49528&view=results when that's done the artifacts will be there to download.

hfloyd:
%Thanks so much, @Shazwazza  - I got em ;-)

hfloyd:
Here are things I tested in an existing v7.15.6 site, using published artifacts "UmbracoCms.7.15.5-alpha202010080001"

### Setup:
- Created a new Image media node at Media root with an uploaded image (name: "Test Pic")
- Created a new Image media node at Media root without any uploaded image file (name: "Missing Test Pic")

### Observations - Previous Behaviors:

**In an RTE Image Insert:**

- Both nodes were visible in the picker/available for picking.
- Inserting the "Test Pic" worked fine.
- Selecting "Missing Test Pic" treated it as a folder - by "opening" it inside the picker:
image
Clicking the green "Submit" button didn't really do anything.

**On a "Single-Image-Only" Media picker**
([Pick multiple items = false], [Pick only images = true], [Disable folder select = true])
- Similar behavior as RTE: Both nodes were visible, selecting "Missing Test Pic" treated it as a folder - by "opening" it inside the picker 
image

**On an "Multi + No folders" Media picker**
([Pick multiple items = true], [Pick only images = false], [Disable folder select = true])
- Same behavior as above: Both nodes were visible, selecting "Missing Test Pic" treated it as a folder - by "opening" it inside the picker, clicking "Submit" returns to the main content with no Media attached to the property.

**On an "Single/No folders" Media picker**
([Pick multiple items = false], [Pick only images = false], [Disable folder select = true])
- Same behavior as above: Both nodes were visible, selecting "Missing Test Pic" treated it as a folder - by "opening" it inside the picker, clicking "Submit" returns to the main content with no Media attached to the property.

### Observations - New  Behaviors:

**In an RTE Image Insert:**

- Both nodes were visible in the picker/available for picking.
- Inserting the "Test Pic" worked fine - the post-select popup displayed the selected image an provided an opportunity to add ALT tag:
image
- Selecting "Missing Test Pic" showed the same post-select pop-up as for the "Test Pic":, but the Url was blank:
image

- It displays as a broken image in the RTE:
image
And is visible with the UDI in the Sourcecode:
image
*Notice that the provided ALT tag is preserved*

**On a "Single-Image-Only" Media picker**
([Pick multiple items = false], [Pick only images = true], [Disable folder select = true])
- Both nodes were visible, able to select the "Missing Test Pic" node:
image
- Green "submit" button resulted in a visible link to the media node on the property:
image

**On an "Multi + No folders" Media picker**
([Pick multiple items = true], [Pick only images = false], [Disable folder select = true])
- Both nodes were visible, able to select both nodes:
image
- Both nodes displayed on property:
image

**On an "Single/No folders" Media picker**
([Pick multiple items = false], [Pick only images = false], [Disable folder select = true])
- Same behavior as above: Both nodes were visible, able to select "Missing Test Pic", which is shown attached to the property.



### Random thoughts about the bigger picture...
Considering the ability to create multiple "non-standard" MediaTypes, it seems the only way to truly handle Media Picker configurations would be for Media Types to have some additional properties - like  "Treat as Folder"/"Treat as Image"/"Treat as File" sort of options... otherwise how would the MediaPicker ever "know" if a custom media Type is a "folder" and thus should be non-pickable in the event that [Disable folder select = true], similarly, if [Pick only images = true] is configured... How will the MediaPicker determine what MediaTypes count as "Images"?

Shazwazza:
Thanks @hfloyd!

Based on your test notes above, can you confirm that the only 'issue' is that when selecting a blank image and inserting into the RTE weird things happen because there isn't an image to load. I'm unsure what happens in v7.14 or v8 when we do that, might be the same issue so not sure if that's a 'new' issue.

> How will the MediaPicker determine what MediaTypes count as "Images"?

Yes this is tricky business and also not consistent 😢 

The way it works now in this v7 code is that a media entity will have a media file path stored in the DB. If that path doesn't exist we fallback to the behavior of checking if the media type contains a property called `umbracoFile` and it will be treated as a file, anything else will be a folder. So in this case for v7, a custom media type will just need to have a property type of this name. 

*But* there's also another convention that any media type that is suffixed with the name "Folder" is treated as a folder in the case where we are displaying the media grid and getting child folders, see https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Web/Editors/MediaController.cs#L215

There's also code for this same convention in JS ... but I don't think this method is actually used https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/Umbraco.Web.UI.Client/src/common/services/mediatypehelper.service.js#L10

Things get more complicated when we start taking into account Media "Slim" entities resolved from the EntityService/EntityController vs full Media objects resolved from the MediaService/MediaController. This fix and many of the work arounds are to deal with this difference. A slim media entity is used throughout the back office in things like pickers, etc... it doesn't contain all media data which there can be a ton of. A full media object is used mostly when just editing a media item and there's also legacy/inconsistent ways of determining a full media object's media path and whether it's a folder that is different from what is listed above ☹️ 

Then... v8 is also slightly different. A property editor can implement `IDataEditorWithMediaPath` which is how the path of a media item is resolved and stored in the database. Media entities are also a bit more capable with v8. **BUT** (and I've just tested this), this problem totally exists in v8 too and it's all the same issue. There's still nothing in v8 that specifically denotes a folder or a file or an image. Files and images will behave differently too because the media picker has an option to only allow image selection. That is currently taken care of by checking the file extension... but if there is no file chosen then there is no way to know (this is probably also an issue in v7 too).

@nul800sebastiaan This issue isn't just v7, it is also v8. So while this fix does 'solve' a regression issue it doesn't actually solve the issue and v8 has the same problem. Moving forward - do we keep hacking the planet and coming up with weird ways to fix this or do we fix this properly? I would like to think we should make this all consistent so we can actually understand what is going on.  For v7, I'm fine with leaving it with this fix, it is the same experience now as before the regression. For v8+ we should fix this properly

v8+ Options:

* Easy (maybe hacky): Determine media types (either Folder, Image or File) by naming convention. Suffixed with "Folder" = folder, suffixed with "Image" = image, suffixed with "File" = file. This also works for enabling other types in the future without doing any work like "Video" if that became a thing
* Have an option on the media type to choose: Folder, Image or File. Adding more options would relatively trivial since it's just a string or an INT. Requires DB and UI changes along with data migrations
* Others? 

Then we also want to cleanup and deprecate a bunch of stuff that currently exists to work around this issue that has been there for many versions (and delete that code in netcore). 

Thoughts?

hfloyd:
**RE:** 
> Based on your test notes above, can you confirm that the only 'issue' is that when selecting a blank image and inserting into the RTE weird things happen because there isn't an image to load.

I actually wouldn't call that an 'issue' - I think it is actually the appropriate behavior - the UDI is there, in case a file ever gets attached to the Media node. I'm not sure how else it could reasonably be handled. Also, showing the "nothing.jpg" img acts as a reminder to the Editor that there won't be a valid image rendered there, because it doesn't exist.

**Regarding your other notes...**
I agree that it probably doesn't make sense to change the setup/behavior for v7 at this point. For those with v7 sites, they are obviously working with it as-is - and fixing this issue will handle the real problem being experienced.

**In terms of v8+...**
Anything which cleans up code or makes something operate consistently is a good investment, in my mind. In general, I prefer obvious and explicit settings rather than assumptions or mysterious "conventions" (which then need to be documented somewhere), however, I can see the need for assumptions when explicit data doesn't exist.

**If the goals are thus:**
1. Clarity & consistency around rules for what constitutes an image or folder MediaType - in terms of the Picker and any other places this distinction is important
2. Minimizing problems with back-wards compatibility and data migrations/upgrades

**My personal "perfect scenario"** (keep in mind - I'm not nearly as conversant of all the internal Core code moving parts here - as you detailed above, and tend to be a bit "old fashioned" when thinking about code problems):
- I think adding a drop-down sort of option for "MediaType Classification" or something could make sense. My suggestion would be to provide an option for "Other" (all options: Folder, File, Image, Other) so people have that as an option for things like "Video" (which I personally HAVE used as a Media Type - tends to have a textfield for a Youtube/Vimeo link or ID and a file upload for a thumbnail image). 
- This new field should also be BLANK by default on non-standard MediaTypes - so upgrades/migrations will just have a blank value for existing MediaTypes until the dev updates it to take advantage of the improved functionality.
- Ideally, there is a single "function" to determine what the "official" classification for the MediaType is. Any implementation (MediaService/IPubContent/"Slim"Media/etc) should use that function. 
- To support "unknown" (blank) Classifications, the function could first check for an explicit Classification value, and if missing, it can fallback to using the mediaFile/umbracoFile check and/or naming-convention-based determination - ideally the fall-back would be as close to what is currently in use, so devs who don't pay any attention to the additional "Classification" data won't have ugly surprises on upgrade.

hfloyd:
@Shazwazza - Looking at this again today, I do notice a strange issue CAUSED by this update.

In my Media section, I had added those two "testing" files, but since this is a "real" site, there are a bunch of folders containing other media items.

**On a "Single-Image-Only" Media picker**
([Pick multiple items = false], [Pick only images = true], [Disable folder select = true])

I notice that none of the folders are visible for navigating into - only those two "root" level images are present:
image

On the live version of this site, that same Picker (without the update installed), allows navigating through all the folders.

If I change the config for the Picker to [Pick only images = false], then I can see the folders:
image
I can click into the folders and pick a deeper-level image.

Do you see the same behavior?

src/Umbraco.Web.UI.Client/src/common/directives/components/umbmediagrid.directive.js
```diff
@@ -120,7 +120,7 @@ Use this directive to generate a thumbnail grid of media items.
                     setOriginalSize(item, itemMaxHeight);
 
                     // remove non images when onlyImages is set to true
-                    if(scope.onlyImages === "true" && !item.isFolder && !item.thumbnail){
+                    if (scope.onlyImages === "true" && item.isFolder === true) {
```
drangus, line 123: I think this should be:
if (scope.onlyImages === "true" && item.isFolder === false && item.thumbnail == null) {

src/Umbraco.Web.UI.Client/src/common/services/mediahelper.service.js
```diff
@@ -5,10 +5,37 @@
...
+
+                if (mediaEntity.metaData.MediaPath) {
+                    mediaEntity.thumbnail = this.resolveFileFromEntity(mediaEntity, true);
+                    mediaEntity.image = this.resolveFileFromEntity(mediaEntity, false);
+                    if (mediaEntity.isFolder === undefined) {
+                        mediaEntity.isFolder = mediaEntity.image ? false : true;
+                    }
+                }
+                else if (mediaEntity.isFolder === undefined) {
+                    mediaEntity.isFolder = mediaEntity.metaData.ContentTypeAlias && mediaEntity.metaData.ContentTypeAlias.toLowerCase() === "folder";
```
drangus, line 34: This may be too restrictive as it prevents additional, bespoke folder types being catered for. It could at least be based on these types having aliases beginning with 'folder', i.e., 
`mediaEntity.isFolder = mediaEntity.metaData.ContentTypeAlias && mediaEntity.metaData.ContentTypeAlias.toLowerCase().startsWith("folder");`

nul800sebastiaan:
😓 😓 Wow reviewing was quite a ride! 

First off, I can't believe this is also an issue in v8.. but it is! The changes here are mostly good, @drangus had a good suggestion which I've implemented, that fixed the problem @hfloyd that was CAUSED by this PR.

The second suggestion from @drangus was also implemented, from the comments here I gather our "logic" has always been that any media type with the suffix `folder` is a folder, [so that's added now](https://github.com/umbraco/Umbraco-CMS/pull/8520/files#diff-ac3413db0c33316d01137763c530965d23507eaaf203d9eaf492259fe03f1ea2R36).

However... ! I found a weird problem in `private T HackForMediaEntity<T>(T entity, UmbracoEntityTypes entityType, Lazy<IEnumerable<IMediaType>> allMediaTypes)`. In there we look at each entity and ask for `entity.Alias`. For some reason that alias is always `null` 😮 

This meant that the code to check if the entity has an `umbracoFile` property never ran. I've updated the code to [fetch the alias from the `AdditionalData` collection](https://github.com/umbraco/Umbraco-CMS/pull/8520/files#diff-478f0608a2e0cf8ea0496ba56fd37f3cbe1dd882062a86ca7bcb08da5c9c7296R1084-R1094).  

That update makes sure that `IsFolder` should be always set, meaning that the code in `mediahelper.service.js` never gets reached, [I've left a comment](https://github.com/umbraco/Umbraco-CMS/pull/8520/files#diff-ac3413db0c33316d01137763c530965d23507eaaf203d9eaf492259fe03f1ea2R34) as I'm not 100% sure about that. 

I'll await some feedback from @Shazwazza to make sure I didn't miss anything silly, but after that this one can (finally!) get merged. Then we'll need to have another look at v8 because this is (indeed) a mess.

nul800sebastiaan (approved):

hfloyd:
Thanks for pushing this forward, @nul800sebastiaan !

src/Umbraco.Web.UI.Client/src/common/services/mediahelper.service.js
```diff
@@ -31,7 +31,9 @@ function mediaHelper(umbRequestHelper, $log) {
                     }
                 }
                 else if (mediaEntity.isFolder === undefined) {
-                    mediaEntity.isFolder = mediaEntity.metaData.ContentTypeAlias && mediaEntity.metaData.ContentTypeAlias.toLowerCase() === "folder";
+                    // We should never get here, resolveFileFromEntity should have worked
+                    var alias = mediaEntity.metaData.ContentTypeAlias.toLowerCase();
+                    mediaEntity.isFolder = mediaEntity.metaData.ContentTypeAlias && (alias.endsWith("folder"));
```
Shazwazza, line 36: @nul800sebastiaan This check is weird, there is no point in checking the existence of `mediaEntity.metaData.ContentTypeAlias` since you've already `toLowerCase` it above. If it didn't exist, then an exception would have already been thrown. This null check needs to be fixed.

src/Umbraco.Web.UI.Client/src/common/directives/components/umbmediagrid.directive.js
```diff
@@ -120,7 +120,7 @@ Use this directive to generate a thumbnail grid of media items.
                     setOriginalSize(item, itemMaxHeight);
 
                     // remove non images when onlyImages is set to true
-                    if (scope.onlyImages === "true" && item.isFolder === true) {
+                    if (scope.onlyImages === "true" && item.isFolder === false && item.thumbnail == null) {
```
Shazwazza, line 123: Don't we still also want to check `isFolder === true`? If that is the case then it should definitely be ignored right?

maybe like?

```js
if (scope.onlyImages === "true" && (item.isFolder === true || (item.isFolder === false && item.thumbnail == null)))
```

Shazwazza (commented):
@nul800sebastiaan couple of things to fixup, have left comments/questions inline.

src/Umbraco.Web.UI.Client/src/common/directives/components/umbmediagrid.directive.js
drangus: Agreed!

drangus: If we include item.isFolder === true, all folders will be removed, thereby preventing navigation down the folder tree?

PR #8521, opened 2020-07-30
Don't log message if svg icon isn't found

bjarnef:
### Prerequisites

### Description
At the moment the changes regarding `umb-icon` directive extended IconController to look for svg icons and log an error if it isn't found.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Editors/IconController.cs#L95

However this will fill log with errors, when using custom font icons where svg icons are not available, e.g. when Umbraco Forms is installed.

chrome_2020-07-30_13-30-20

### DIFF
--- a/src/Umbraco.Web/Editors/IconController.cs
+++ b/src/Umbraco.Web/Editors/IconController.cs
@@ -90,9 +90,8 @@ namespace Umbraco.Web.Editors
 
                 return svg;
             }
-            catch (Exception ex)
+            catch
             {
-                Logger.Error<IconController>(ex, $"Could not load {iconName}.svg");
                 return null;
             }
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks @bjarnef - it's definitely an exception though, the icon file should be there. But it would be fine to log it as a `Debug` level instead, so it's not logged by default but will be logged if you increase the log level to debug. Can you update it to that please?

nul800sebastiaan:
Ah.. I didn't read your last sentence.. yeah that might be quite a lot of logging anyway if the svg is not implemented yet, better to ignore it for now indeed! 👍

PR #8523, opened 2020-07-30
Replace underscore _.throttle function with utility function

bjarnef:
### Prerequisites

### Description
I noticed the underscore `_.throttle` function is used a few places.
This PR replaces this with an utility function using vanilla javascript.

To test this it is easiest to increase timeout to e.g. 2000 milliseconds (2 seconds) and in query builder check in the browser network panel that `PostTemplateQuery` isn't requested each time when you click ascending/descending button fast.


nul800sebastiaan:
As noted before: while we love the idea of decoupling from AngularJS/underscore we don't want to then maintain our own custom library that we'd need to maintain and find and fix bugs in etc.
So we won't be able to detach this aggressively at the moment unfortunately. Therefore we'll close this PR and please note that for the moment, we won't consider future PRs that introduce new methods in the Utilities file.

PR #8524, opened 2020-07-30
Replace underscore _.debounce function with utility function

bjarnef:
### Prerequisites

### Description
I noticed the underscore `_.debounce` function is used a few places.
This PR replaces this with an utility function using vanilla javascript.

To test this check mainly search, e.g. listview search, that search happens with a delay of few milliseconds and therefore no search on each keypress.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/imaging/umbimagegravity.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/imaging/umbimagegravity.directive.js
@@ -195,7 +195,7 @@
             //vm.center.top = (offsetY + 10) / scope.dimensions.height;
         };
 
-        var lazyEndEvent = _.debounce(function () {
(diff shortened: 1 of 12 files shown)


nul800sebastiaan:
@bjarnef Can you hold off with replacing more of the underscore bits please? We need to have a strategy with this one and I think we don't want to get rid of ALL of underscore, just the lines that can be easily replaced without having to create a new Utility function for it.

We definitely don't want to maintain a full library that does exactly what underscore does, there's a tool for that, it's called underscore and it's pretty good! 😅 

cc: @warrenbuckley @nathanwoulfe

nul800sebastiaan:
As noted before: while we love the idea of decoupling from AngularJS/underscore we don't want to then maintain our own custom library that we'd need to maintain and find and fix bugs in etc.
So we won't be able to detach this aggressively at the moment unfortunately. Therefore we'll close this PR and please note that for the moment, we won't consider future PRs that introduce new methods in the Utilities file.

PR #8527, opened 2020-07-30
Update model name in logviewer saved search

bjarnef:
### Prerequisites

### Description
I noticed a minor issue when saving a search query in logviewer, where it by default use "overlay" and name, which isn't friendly.
This is because we now have this default name in overlay used for `data-element` value.

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/overlays/umboverlay.directive.js#L224-L226

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/overlays/umb-overlay.html#L2

I have updated this so then model use `queryName` as model value and added a bit localization of the overlay as well.

**Before**

**After**


nul800sebastiaan:
Good stuff Bjarne, thanks!

Issue #8526, opened 2020-07-30
New icon directive makes tags look a little weird


nul800sebastiaan:
Now that #5293 has been merged some icons behave a little different than they did before. I just noticed that the tags delete icon is a too dark and is also causing a lot of extra padding on the tag itself.

![image](https://user-images.githubusercontent.com/304656/88939666-14d2d200-d287-11ea-9bb5-7e23b1f596d5.png)

![image](https://user-images.githubusercontent.com/304656/88939673-17352c00-d287-11ea-9ba2-4d8a46985c9c.png)

bjarnef:
It does indeed not look as it should.
PR submitted: https://github.com/umbraco/Umbraco-CMS/pull/8528

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8528 - thanks again Bjarne!

PR #5293, opened 2019-04-17
V8: Update backoffice icon fonts to use svgs

MMasey:
**Summary**

This PR updates almost all of the icons using the icon font found throughout the backoffice with SVGs. Each icon is now it's own svg file with the following naming convention `icon-[name]`. I have created an IconController that currently pointed to a folder containing these icons and makes them usable via the an angular directive call `umb-icon`. This not only improves accessibility, but also make it much easier to customise the icons available in the back office.

This is a pretty big PR and is currently still a work in progress, so i'd really appreciate some input/feedback whether it's around the coding style, tackling the SVG sanitization (I have a potential solution but don't just want to throw it in there) or just general questions. I think it's almost there but I may be wrong.

I've put a fair amount of detail in the issue https://github.com/umbraco/Umbraco-CMS/issues/3560 so I would recommend having a read through to get up to speed. 

I'm creating this as a draft PR for the time being as I don't think it's ready to be properly submitted yet.

Many thanks

PR #8528, opened 2020-07-30
Fix styling of confirm actions in tags editor

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8526

### Description
This PR fixes some styling issues after merging https://github.com/umbraco/Umbraco-CMS/pull/5293 and moved trash icon inside `umb-confirm-action` in https://github.com/umbraco/Umbraco-CMS/pull/8198

It also fixes issues where no outline was shown on focus.
Since `umb-tags-editor` is a separate component I have moved the styling to `umb-tags-editor.less`, so it re-used developers will get the same styling. The property editor still has the `umb-tags` class which has styling with the limited width.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/belle.less
+++ b/src/Umbraco.Web.UI.Client/src/less/belle.less
@@ -175,6 +175,7 @@
 @import "components/umb-dropdown.less";
 @import "components/umb-range-slider.less";
 @import "components/umb-number.less";
+@import "components/umb-tags-editor.less";
 
 @import "components/buttons/umb-button.less";
 @import "components/buttons/umb-button-group.less";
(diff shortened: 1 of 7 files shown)


bjarnef:
@nul800sebastiaan sure, will take a look at that as well 👍

bjarnef:
@nul800sebastiaan actually the icons are still there, just too much padding.

image

It seemed @kjac changed this in https://github.com/umbraco/Umbraco-CMS/pull/8504 but it also affect these confirm and cancel buttons. Looking at a fix for this.

bjarnef:
@nul800sebastiaan I have reverted the fix from @kjac so it only target `<button>` element directly below `umb-group-builder__property-action` (not delete button since this is moved inside `umb-confirm-action`), but both buttons has the `btn-icon` class.

So it should be good now 👍 

image

image

nul800sebastiaan:
Yep, works great again now! 👍 ⭐

PR #8529, opened 2020-07-30
Buttons cleanup

bjarnef:
### Prerequisites

### Description
This PR fixes a few oversights after changing anchor elements to `<button>` elements.:

- Remove `href` attributes from `<button>` elements.
- Remove `prevent-default` directive on `<button>` elements.
- Adding missing type attributes (browsers may have different default behavior).
- Replace anchor elements in pagination with `<button>` elements + disable button for current page and prev/next when on first/last page. This also ensure when using keyboard it doesn't focus "inactive" buttons.


nul800sebastiaan:
Fantastic, great cleanup, thanks!!

PR #8533, opened 2020-07-31
Some UI update to validation messages

nielslyngsoe:
image
Making the property-error message into a speech bubble style, so it yells less in your face.

Shows a error-badge on content-apps if no badge already is present, and colors the badge in red to make it even more visible that there is issues.
image


warrenbuckley:
Code looks fine with it being 90% CSS
Off to test/verify

warrenbuckley:
All good and even got the subtle bounce animation to try & get my attention when switched to the info tab

PR #8534, opened 2020-07-31
Fix object iteration in underscore _.each changes

nathanwoulfe:
@nul800sebastiaan this will fix the issue raised on #8475 - data returned for searchAll is a dictionary so I've updated to use Object.values(). The other methods on the service are all arrays so will be fine.


nul800sebastiaan:
It does a lot more than that, but all good! Thanks again @nathanwoulfe! 👍

PR #8540, opened 2020-07-31
Languages names do not show in publish/unpublish dialogs

kjac:
### Prerequisites

### Description

The listview publish/unpublish dialogs are Completely Broken™ (the language names are missing):

listview-publish-unpublish-languages-before

I'm not entirely sure why 🤷 but it seems related to the frontend changes that were (accidentally?) included in #7700 (possibly the addition of `transclude` in the checkbox directive) ... or maybe it's something entirely different. Whatever the case is, it's fixable by _not_ auto-closing the `<umb-checkbox/>` element, so that's exactly what this PR does to fix it:

listview-publish-unpublish-languages

Also the "mandatory language" label was a bit off, so I fixed that - and I removed the usage of `angular.forEach` just because.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-list.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-list.less
@@ -40,7 +40,7 @@ a.umb-list-item:focus {
 }
 
 .umb-list-item__description--checkbox{
-    margin: 0 0 0 26px;
+    margin: 0 0 0 30px;
 }
 
 .umb-list-checkbox {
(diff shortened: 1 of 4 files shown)


bjarnef:
@kjac yes, custom angular elements can't be self-closing elements, so they need a closing end tag 👍 
here is a great explanation of why it can't be self-closing: https://stackoverflow.com/a/25012451/1693918

nul800sebastiaan:
Cheers @kjac, good stuff!

clausjensen:
Manually merged to 8.7 - updated release tag.

nul800sebastiaan:
Guessing this is part of commits d00c07592e73318bbc90bfe6fc571dce406cb680 and f25ee0a369e503efe635e5c8762ee99b624d1001

PR #8541, opened 2020-07-31
Ask user if they are sure about discarding changes

nielslyngsoe:
If form is dirty, aka. has been changed, we now ask the user if they are sure about discarding those changes in a dialog.

### DIFF
--- a/src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
+++ b/src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
@@ -647,7 +647,7 @@
     <key alias="design">Design</key>
     <key alias="dictionary">Ordbog</key>
     <key alias="dimensions">Dimensioner</key>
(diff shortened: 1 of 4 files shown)


src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
```diff
@@ -647,6 +647,7 @@
     <key alias="design">Design</key>
     <key alias="dictionary">Ordbog</key>
     <key alias="dimensions">Dimensioner</key>
+    <key alias="discard">Discard</key>
```
bjarnef, line 650: @nielslyngsoe "Kassér" would be a better for Danish translation 👍

nielslyngsoe: @bjarnef What a bummer, I fully agree. :-)

PR #8546, opened 2020-07-31
Grid Textstring editor - Bug with unique ID.

leekelleher:
### Prerequisites

### Description

Whilst using the textstring grid editor (aka. Headline), I'd noticed in Chrome Dev Tools there was a warning of multiple/duplicated ID attributes of `id="_text"`.

Digging deeper, turns out to be a small bug when setting the ID in [views/propertyeditors/grid/editors/textstring.html#L6](https://github.com/umbraco/Umbraco-CMS/blob/release-8.6.4/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/textstring.html#L6), the `uniqueId` property is missing the leading `$` symbol, (a-la `$uniqueId`, [as it's set in the Grid controller](https://github.com/umbraco/Umbraco-CMS/blob/release-8.6.4/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js#L917).)

Steps for reproduction, add multiple Headline editors to a Grid, view Chrome Dev Tools _(other web browsers are available 🤑)_ see the multiple same IDs warning.

Once patched, the ID is generated as `id="{GUID}_text"`, making the ID unique on the page.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/textstring.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/textstring.html
@@ -3,7 +3,7 @@
 		rows="1"
 		umb-auto-resize
 		umb-auto-focus
-		class="textstring input-block-level" id="{{control.uniqueId}}_text"
+		class="textstring input-block-level" id="{{control.$uniqueId}}_text"
 		ng-model="control.value"
 		ng-attr-style="{{control.editor.config.style}}" localize="placeholder" placeholder="@grid_placeholderWriteHere"></textarea>
 </div>


leekelleher:
I've not idea why the build failed, it doesn't appear to be related to my code changes. I wonder if the build failure is why this PR isn't being reviewed? Should I rebase my branch? 🤷‍♂️

abjerner:
@leekelleher the builds fail now and then. I normally don't pay much attention to it for Angular-related PRs.

I think people are generally quite busy - either due to the effects of COVID-19, summer vacation or similar. This probably also affects the PR team.

So while you ought to have received a reply by now, I think the above is more the reason why we haven't replied than because of the build errors.

I'll try to have a look and get your PR reviewed 😉

abjerner:
Weird. I can see ID is wrong/duplicate when inspecting the DOM (so you're right that it should be fixed), but for some reason my Chrome doesn't complain about it (which is usually does for that kind of errors).

I'll try on another machine tomorrow to see if I can reproduce it there 😮

abjerner (approved):
I've now tried in Chrome, Edge and Firefox, but I still can't reproduce the error with duplicate IDs. But I think your fix is correct, so it's hereby approved 👍 

I know the overlays from the editor service has problems duplicate **headerName** IDs, and these show up in the console just fine:

image

abjerner:
@leekelleher thanks a lot 👍 

I somehow still can't reproduce the console error, but believing the change is correct, I've now merged your PR 🎉

leekelleher:
Thanks @abjerner . I'd been meaning to get a screengrab for you, _(but hadn't got around to it yet)._ Thanks for merging in.

PR #8478, opened 2020-07-25
Remove button for custom listview layouts isn't rendered

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
Custom angular element directives/components can't be self-closing elements like navtive elements, e.g. `<input type="checkbox value="" />`:
https://stackoverflow.com/a/25012451/1693918

This was previous fixed in https://github.com/umbraco/Umbraco-CMS/pull/8422 but overwritten in https://github.com/umbraco/Umbraco-CMS/commit/1cd79d81af821470de2123e1c6febe2980d98111#diff-c84512d9cc059db603911cd477e1d82e

**Before**

image

**After**

image

PR #8547, opened 2020-07-31
Fix self closing angular elements

bjarnef:
### Prerequisites

### Description
This PR ensure all `<umb-checkbox>` and `<umb-radiobutton>` not are void elements (self-closed elements) since this turns out to  cause issues and in these examples:
https://github.com/umbraco/Umbraco-CMS/pull/8540

Here is a great explanation of why it can't be self-closing: https://stackoverflow.com/a/25012451/1693918


clausjensen:
I've manually merged these changes over to 8.7 and marked as fixed in 8.7. This bug existed in 8.6 as well so it is not something we have introduced right now, and it should be fixed before we send out a 8.7 RC.

nul800sebastiaan:
Guessing this is part of commits d00c07592e73318bbc90bfe6fc571dce406cb680 and f25ee0a369e503efe635e5c8762ee99b624d1001

PR #8548, opened 2020-07-31
Grid cell error could use "umb-code-snippet" directive?

leekelleher:
### Prerequisites

### Description

Thanks to @bjarnef's excellent `umb-code-snippet` component #7294, it could be used when displaying errors with Grid editors cells?

Currently the [error.html#L2](https://github.com/umbraco/Umbraco-CMS/blob/release-8.6.4/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/error.html#L2) is using a `<pre>`, where as with `<umb-code-snippet>` it would also offer the *Copy* button - useful for sharing/debugging purposes.

**with `<pre>`**

**with `<umb-code-snippet>`**

_A small incremental improvement._


nul800sebastiaan:
Thanks @leekelleher! Lovely with the copy button indeed, makes total sense to me, thanks! 👍

leekelleher:
Cool, thanks @nul800sebastiaan!

Issue #8440, opened 2020-07-14
v8 - Add support for openAnchorItem in umbEditorHeader


jamiepollock:
The ``umb-editor-header`` [does not support](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-header.html#L88-L92) the [openAnchorItem property](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/editor/umbeditornavigation.directive.js#L139) on the ``umb-editor-navigation`` directive.

Including support for this in the ``umb-editor-header`` directive would mean uses of this in custom sections could leverage this callback to interact with subviews or jump elements in the view like the content editor currently does for property groups.

In Umbraco v8.6.4 implementing a navigation definition like so, also does not update the anchors state:

```
vm.navigation = [
    {
        "name": "First",
        "alias": "first",
        "icon": "icon-pie-chart",
        "view":  "/App_Plugins/Path/To/First/View.html",,
        "active": true
    },
    {
        "name": "Second",
        "alias": "second",
        "icon": "icon-shorts",
        "view": "/App_Plugins/Path/To/Second/View.html",
        "active": false,
        "anchors": [
            { "label": "Anchor 1", "id": 0 },
            { "label": "Anchor 2", "id": 1 },
            { "label": "Anchor 3", "id": 2 }
        ]
    }
];
```

How can you help?
-------------------------------

I would be happy to implement a solution which makes this available to developers.

nul800sebastiaan:
Sure thing, that would be lovely! 👍

umbrabot:
Hi @jamiepollock,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

jamiepollock:
@nul800sebastiaan sorry for the lack of acknowledgement to your response. It is still on my radar. I hope to start work on this in the next week.

umbrabot:
Hiya @jamiepollock!

[...]

PR #8549, opened 2020-07-31
Added hook for onSelectAnchorItem in umb-editor-header.

jamiepollock:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8440

### Description

This adds a hook for onSelectAnchorItem for umb-editor-navigation to umb-editor-header.

Here is a [simple dashboard demo](https://github.com/umbraco/Umbraco-CMS/files/5008439/demo-8440.zip) which can be dropped into an Umbraco solution which uses this code base. 

The demo app dashboard utilizes this new hook so it can be tested by HQ. This was used as part of my own testing during development. The demo app reports the navigation item & anchor when clicked in the browser developer console.

**Note:** The current implementation does not modify the active state of the anchor item. If this is required we can discuss the best place to implement this.


nul800sebastiaan (changes requested):
Thanks @jamiepollock and apologies for the delay! I'm having a heck of a time trying out this PR, I am expecting to see the subviews when I switch between the "section" buttons but nothing is happening, what am I missing?

nathanwoulfe (approved):
Hey @jamiepollock, hope you remember raising this almost a year ago... Works really nicely, I can certainly see how this will be useful.

@nul800sebastiaan not sure why the dashboard views don't load, I see the same output as your example. However, the new function is called, and the expected values are logged. Looking on a content node, the new function is called when selecting an anchor in the content dropdown, so it follows then that passing in a function on the new attribute will see that function called correctly.

I'm happy to merge this, assuming there are no concerns from the example not working as expected?

jamiepollock:
@nathanwoulfe @nul800sebastiaan hey both, I am sorry I've been ghosting this ticket for so long. Life has been just crazy this past year due to multiple factors, I've not had the energy to deal with anything outside of work. I'll try my best to make amends going forward. I could go on but I'm sure we've all had a rough year or so with COVID.

If you & Seb are happy to merge this then that's great. Otherwise I can try and pick up where I left off. I don't know why the example isn't working as expected. I apologise about that.

nathanwoulfe:
Hey @jamiepollock all good my friend, completely understand. I'll try pick it up again next week, see if I can sort the dashboard demo

PR #8550, opened 2020-07-31
Fix media picker clickhandler

bjarnef:
### Prerequisites

### Description
When open media picker and clicking media item name, I noticed a console error. This is because it call `$scope.clickHandler()` but it should just be `clickHandler()`.

image


nul800sebastiaan:
Nice one, thanks @bjarnef!

PR #8543, opened 2020-07-31
Fix styling of buttons in grid cells

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
This PR adjust styling of buttons in grid cells and optimize the buttons for accessibility.

**Before**

**After**

PR #8551, opened 2020-08-01
Fix broken grid LESS causing the client build to fail

kjac:
### Prerequisites

### Description

A slight error snuck into the grid LESS in #8543. Unfortunately this error causes the entire client build to fail:

image

I believe this PR fixes it; the grid styling at least looks like what's shown in #8543

/cc @bjarnef

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-grid.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-grid.less
@@ -754,11 +754,10 @@
     display: inline-block;
 }
 
-.umb-grid .cell-tools {
-    .umb-grid .umb-control-tool {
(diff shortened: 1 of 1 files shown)


bjarnef:
@kjac ahh, it was just missing a comma in 
https://github.com/umbraco/Umbraco-CMS/pull/8543

It should have been this instead:

```
.umb-grid .cell-tools,
.umb-grid .umb-control-tool {
    .btn-icon {
        padding: 0;
    }
}
```

kjac:
@bjarnef that works too 👍 I have updated the fix accordingly.

nul800sebastiaan:
D'oh! 🤦‍♂️ Thanks gentlemen!

PR #8554, opened 2020-08-02
Update prevalues multivalues sortable styling

bjarnef:
### Prerequisites

### Description
At the moment is was assumed that left column in multivalues prevalue editor is always sortable and therefore added `cursor:move`.
However that is not always the case e.g. in macro parameter editors, where only the icon was acting as sortable handle.

I have updated this to be move specific on `handle` class which is often used or `ui-sortable-handle` which is added on sortable items.

image

Also note the move cursor is still shown in colorpicker prevalue editor, when hover the color preview and hex code.

image


nul800sebastiaan:
Good stuff @bjarnef - thanks! 👍

PR #8558, opened 2020-08-03
Bugfix: Tour's next button not clickable when infinite editing bar is open

bergmania:
### Details
- When the Get started tour is executed, the tour is locked at step 8 of 12 because the next button is not clickable

### Changes
- Moved the `<umb-tour>` tag out of the current scope. 

### Test
- Run the get started tour. Before the next butten of step `8/12` was not clickable.
- Now
tour


nul800sebastiaan:
Cherry-picked for 8.7.0 in https://github.com/umbraco/Umbraco-CMS/commit/bd1aac7b36283a0843c96582beac18863333a03a

PR #8559, opened 2020-08-03
Fix selectors causing acceptance test to give false errors.

nikolajlauridsen:
### Prerequisites

### Description
I Changed the selectors in the cypress tests to match the selectors on the site.


bergmania:
Looks good :)

PR #8561, opened 2020-08-03
Changes c# model for block list editor to be much simpler.

Shazwazza:
For the block list editor we don't need all of the other content/settings data, we only need to know what is in the layout. This simplifies the whole model and makes it easier to work with.

The model itself is now an `IReadOnlyList<BlockListItem>` This means you can iterate the model directly and you have access to get an item by its index but I've also created other indexers too. 

This is before:

```cshtml
@using Umbraco.Core.Models.Blocks
@inherits UmbracoViewPage
@{
    int index = 0;
    var blocks = Model.Value<BlockListModel>("widgetsBlock");
    foreach (var layout in blocks.Layout)
    {
        var content = layout.Content;
        @Html.Partial("UnoBlocks/" + content.ContentType.Alias, layout, new ViewDataDictionary { { "order", index } })
        index++;
    }
}
```

This is after:

```cshtml
@using Umbraco.Core.Models.Blocks
@inherits UmbracoViewPage
@{
    var blocks = Model.Value<BlockListModel>("widgetsBlock");
    foreach (var index = 0; index < blocks.Count; index++)
    {
        var block = blocks[index];
        @Html.Partial("UnoBlocks/" + block.Content.ContentType.Alias, block, new ViewDataDictionary { { "order", index } })
    }
}
```

You can still iterate directly over the BlockListModel value but since we need an index anyways, we might as well use a for loop and because we can access the item by it's index like block[index] it saves a bit of code.

You can also access block items by other index types too:

* `blocks[INT]` (the index of the content item)
* `blocks[GUID]` (the key of the content item)
* `blocks[UDI]` (the element udi of the content item)

The model type for each block is `BlockListItem`


nielslyngsoe:
- "The model type for each block is BlockListItem"

Fantastic, this one was "BlockListLayoutReference" before, right? I´m happy with this name, simple

warrenbuckley:
Code changes look fine
Unit tests pass

PR #8562, opened 2020-08-03
Add missing backdrop when using the "Actions" button

kjac:
### Prerequisites

### Description

When invoking a menu item from the tree, a nice backdrop is added behind the resulting dialog. The same backdrop does not appear if you invoke the same menu item from the "Actions" button:

actions-menu-backdrop-before

...but this PR takes care of it:

actions-menu-backdrop-after


nul800sebastiaan:
Ha! I never noticed! Good one @kjac! 👍

PR #8564, opened 2020-08-03
Make the entire tree menu ellipsis button clickable

kjac:
### Prerequisites

### Description

It's a tad annoying how the tree menu ellipsis button is only clickable at the very center, despite the hover effect and mouse cursor indicating that the entire button should be clickable:

tree-menu-clickable-before

Kinda get the feeling that there are no menu items available when the menu doesn't appear 🤔 

This PR makes the entire button clickable:

tree-menu-clickable-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less
@@ -96,7 +96,6 @@ body.touch .umb-tree {
             width: auto;
             height: auto;
             margin: 0 5px 0 auto;
-            padding: 7px 5px;
             overflow: visible;
             clip: auto;
         }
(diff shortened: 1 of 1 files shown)


bjarnef:
@kjac how would it look when using tab key? I think it would make the action button move, because of the padding here:
https://github.com/umbraco/Umbraco-CMS/blob/2653fcb2b4b6543a506cee166a9c5efe279afd22/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less#L224

kjac:
@bjarnef hmmm I did test that, and just did a double take; it looks just fine. But thanks for reviewing it ❤️ 

tree-menu-clickable-tabbing

nul800sebastiaan:
Looks good to me too! Thanks for the update, good stuff @kjac! 👍

Issue #8537, opened 2020-07-31
Property actions require additional tab to reach first menu item


bjarnef:
With latest changes in `v8/contrib` when using tab key it seems to require an additional keypress to reach first menu label.
This seems to be because there a hidden close button, but the existing button ellipsis button already act as toggle.

Does it need this close button? Regarding accessibility should the button text for screen readers just toggle between `Open Property Actions` to `Close Property Actions` on open/close?

![2020-07-31_14-01-10](https://user-images.githubusercontent.com/2919859/89033267-b6652c80-d336-11ea-8011-49583112466a.gif)

https://github.com/umbraco/Umbraco-CMS/blob/0e4979bd7f0f7fae2effc09cca672f259582fdf1/src/Umbraco.Web.UI.Client/src/views/components/property/property-actions/umb-property-actions.html#L13-L20

cc: @BatJan @nielslyngsoe

nul800sebastiaan:
Agreed, for me there's no need to have an open/close, toggling is fine and tabbing should not require that extra key press! 👍

umbrabot:
Hi @bjarnef,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
Yes, regarding accessibility it can however toggle the `text` here based on state of `vm.isOpen`.
https://github.com/umbraco/Umbraco-CMS/blob/0e4979bd7f0f7fae2effc09cca672f259582fdf1/src/Umbraco.Web.UI.Client/src/views/components/property/property-actions/umb-property-actions.html#L6

bjarnef:
PR submitted: https://github.com/umbraco/Umbraco-CMS/pull/8565

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8565

PR #8565, opened 2020-08-03
Update umb-property-actions toggle button

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8537

### Description
This PR removes the additional close button and use the toggle button, which change the text when dropdown is open/closed.
For some reason this component was located in [src/views/components/property/property-actions](https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/property/property-actions) folder and not consistent with all other components/directives which are located in [/src/common/directives/components/](https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components)

For the view it was located in a subfolder, but I don't think we need an additional folder for just this single file, so I have moved it to this folder instead: [src/views/components/property](https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/property)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyactions.component.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyactions.component.js
@@ -70,8 +70,6 @@
             localizationService.localizeMany(labelKeys).then(values => {
                 vm.labels.openText = values[0];
                 vm.labels.closeText = values[1];
-
-                console.log("vm.labels", vm.labels);
             });
         }
     }
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyactions.component.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyactions.component.js
@@ -5,14 +5,22 @@
      * A component to render the property action toggle
      */
     
-    function umbPropertyActionsController(keyboardService) {
+    function umbPropertyActionsController(keyboardService, localizationService) {
 
         var vm = this;
 
-        vm.labels = {};
         vm.isOpen = false;
+        vm.labels = {
+            openText: "Open Property Actions",
+            closeText: "Close Property Actions"
+        };
 
+        vm.open = open;
+        vm.close = close;
+        vm.executeAction = executeAction;
 
+        vm.$onDestroy = onDestroy;
+        vm.$onInit = onInit;
 
         function initDropDown() {
             keyboardService.bind("esc", vm.close);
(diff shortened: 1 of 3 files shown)


bjarnef:
@nielslyngsoe it would be great if we could ensure the js-part of the directives/components is located same place.

nielslyngsoe:
Hi @bjarnef we have two buttons to make the button go out of the paper. To provide this connected look between the dialog and the options. This PR changes that look. Any specific reason why you want to only have one button? :-)

bjarnef:
@nielslyngsoe have a look at the description in the issue :)
At the moment the close button isn't visible, but because of it exists, it requires two hits on tab key to reach the first menu action.
Furthermore the existing button already act as toggle (open/close), so no need to have the additional hidden close button.

nielslyngsoe:
@bjarnef maybe I misunderstood the description, my concern was in regards to the visual result. Its suppose to look like this:

image

Compared to the closed state:
image

Notice that the button has a shadow and is white when open. In the closed state, it's greyish and has no shadow.

bjarnef:
@nielslyngsoe this PR was done in `v8/contrib` branch and before the block list was merged to this branch, so the styling may have changed. But in my tests I didn't noticed any changes visually before and after this change. See a before gif in the issue.

nul800sebastiaan (changes requested):
@bjarnef I love it!! But yes, if you could make some style updates like @nielslyngsoe suggests then it would be perfect! 👍

bjarnef:
@nielslyngsoe @nul800sebastiaan I have adjusted this in https://github.com/umbraco/Umbraco-CMS/pull/8565/commits/a1ea42665169319b27a120d23ee408d1b18857af which clean up some styling for the extra close button on top of the existing toggle button.

**Current**

image

**Now**

image

bjarnef:
When using tab key to navigate the toggle button and hitting enter it opens the context menu. Then subsequent tab keypress navigate the menu items and shift+tab key is navigating back.
Currently in v8.7 from first menu action and one step back closes the context menu. This is consistent with the action menu in top right corner in editing area.
However in latest changes in `v8/contrib` branch it explicit requires a enter keypress on close button or an additional step back to close the context menu.

With these changes this behaviour it similar to in v8.7.

nul800sebastiaan:
It's great! And it feels natural to use to me, thanks @bjarnef, great job!

PR #8567, opened 2020-08-03
Fix JS error in empty listviews

kjac:
### Prerequisites

### Description

Slight unintended side effect of #8538 is the JS error that occurs when a listview is empty - try searching the "all members" view for a member that does not exist and see the JS console:

image

This PR adds some defensive coding around the failing method to remove the JS error.


nul800sebastiaan:
Whoops, good catch @bjarnef  and thanks for the fix @kjac ! 👍

PR #8569, opened 2020-08-03
Only show mediapicker edit button if media is found

bjarnef:
### Prerequisites

### Description
At the moment when a media is deleted, but the media picker still has the stored reference, the edit button is shown and when clicking this button the infinite mode of media is opened. However because the media isn't available `$scope.page.loading` is never set to false and `$scope.content` is not set.

So basically it keeps showing the loading indicator
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/media/edit.html#L3

and because content isn't shown while loading, content editors doesn't have the option to click close or using `esc` shortcut.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/media/edit.html#L9

the only option is to refresh page.

I have improved the error handling and only show the edit button in media picker, when the media is actually found (also true if in media recycle bin).

I tried a different approach to keep showing edit button for not found media items and showing an empty state message, but it would require hide e.g. header, main content and submit button.

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/media/edit.html#L9

```
<div data-element="editor-media" ng-controller="Umbraco.Editors.Media.EditController">

    <umb-load-indicator ng-if="page.loading"></umb-load-indicator>

    <form novalidate name="contentForm"
          ng-submit="save()"
          val-form-manager>

        <umb-editor-view ng-if="!page.loading">

            <umb-editor-header ng-if="content"
                               name="content.name"
                               menu="page.menu"
                               hide-icon="true"
                               hide-description="true"
                               hide-alias="true"
                               navigation="content.apps"
                               on-select-navigation-item="appChanged(item)"
                               show-back-button="showBack()"
                               on-back="onBack()"
                               setpagetitle="header.setPageTitle">
            </umb-editor-header>

            <umb-editor-container>

                <div class="umb-editor-sub-views" ng-if="!page.loading && content">
                    <div id="sub-view-{{$index}}" ng-repeat="app in content.apps track by app.alias">
                        <umb-editor-sub-view model="app" content="content" />
                    </div>
                </div>

                <umb-empty-state ng-if="!page.loading && !content"
                                 position="center">
                    <localize key="media_itemNotFound">The media item was not found</localize>
                </umb-empty-state>

            </umb-editor-container>

            <umb-editor-footer>

                <umb-editor-footer-content-left>

                    <umb-breadcrumbs ng-if="ancestors && ancestors.length > 0"
                                     ancestors="ancestors"
                                     entity-type="media">
                    </umb-breadcrumbs>

                </umb-editor-footer-content-left>

                <umb-editor-footer-content-right>

                    <umb-button ng-if="model.infiniteMode"
                                type="button"
                                button-style="link"
                                label-key="general_close"
                                shortcut="esc"
                                action="close()">
                    </umb-button>

                    <umb-button alias="returnToList"
                                ng-if="page.listViewPath && !model.infiniteMode"
                                type="link"
                                href="#{{page.listViewPath}}"
                                label-key="buttons_returnToList">
                    </umb-button>

                    <!-- label-key="buttons_save" -->
                    <umb-button alias="save"
                                type="submit"
                                button-style="success"
                                ng-if="content"
                                state="page.saveButtonState"
                                label-key="{{page.submitButtonLabelKey}}"
                                shortcut="ctrl+s">
                    </umb-button>

                </umb-editor-footer-content-right>


            </umb-editor-footer>

        </umb-editor-view>

    </form>
</div>

```

In the end I don't think it add any value to the content editor, so I felt it was a better option to hide to edit button.

I have however improved the error handing in media.edit.controller.js in case we want to make improvements on this later.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/media/media.edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/media/media.edit.controller.js
@@ -51,22 +51,25 @@ function mediaEditController($scope, $routeParams, $location, $http, $q, appStat
 
         $scope.page.loading = true;
 
-        mediaResource.getScaffold(nodeId, $routeParams.doctype)
-            .then(function (data) {
-                $scope.content = data;
+        mediaResource.getScaffold(nodeId, $routeParams.doctype).then(function (data) {
+            $scope.content = data;
 
-                init();
+            init();
 
-                $scope.page.loading = false;
-
-            });
+            $scope.page.loading = false;
+        }, function () {
+            $scope.page.loading = false;
+        });
     }
     else {
         $scope.page.loading = true;
-        loadMedia()
-            .then(function(){
-                $scope.page.loading = false;
-            });
+
+        loadMedia().then(function(){
+            $scope.page.loading = false;
+        }, function (error) {
+            console.log("loadMedia error", error);
+            $scope.page.loading = false;
+        });
     }
 
     function init() {
@@ -121,6 +124,8 @@ function mediaEditController($scope, $routeParams, $location, $http, $q, appStat
             if(args && args.mediaType && args.mediaType.key === $scope.content.contentType.key) {
                 $scope.page.loading = true;
                 loadMedia().then(function() {
+                    $scope.page.loading = false;
+                }, function () {
                     $scope.page.loading = false;
                 });
             }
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
I absolutely agree with this one @bjarnef - it doesn't make sense to edit something that's really not there any more, sweet update, love it! 👍

PR #8571, opened 2020-08-04
Ensure all back office controllers are plugin controllers and use constant

Shazwazza:
ref [AB#6332](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6332)

PluginController attributes means controllers are routed with MapAreaControllerRoute instead of MapControllerRoute which the latter means can be routed with the default route which we don't want.

### DIFF
--- a/src/Umbraco.Web.UI.NetCore/appsettings.json
+++ b/src/Umbraco.Web.UI.NetCore/appsettings.json
@@ -1,6 +1,6 @@
 {
     "ConnectionStrings": {
-        "umbracoDbDSN": "server=teamcanada3;database=v9;user id=sa;password='test'"
+        "umbracoDbDSN": ""
     },
     "Logging": {
         "LogLevel": {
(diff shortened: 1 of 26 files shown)


src/Umbraco.Web.UI.NetCore/appsettings.json
```diff
@@ -1,123 +1,130 @@
...
-          "MaxWidth": 5000,
-          "MaxHeight": 5000
-        },
-        "Cache": {
-          "Folder": "../App_Data/Cache",
-          "MaxBrowserCacheDays": 7,
-          "MaxCacheDays": 365,
-          "CachedNameLength": 8
+    "ConnectionStrings": {
+        "umbracoDbDSN": "server=teamcanada3;database=v9;user id=sa;password='test'"
```
bergmania, line 3: I connection string needs to be rolled back

Shazwazza:
yep means HealthCheckController is a bug in v8

PR #8572, opened 2020-08-04
Migrates remaining non-oauth actions on back office controller to netcore

Shazwazza:
ref [AB#6970](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6970)

Migrates remaining non-oauth actions on back office controller to netcore including the authorize upgrade cshtml view.

## Testing

* Log out of the back office, or clear cookies
* Update the table in the Db `umbracoKeyValue`, change the value for `Umbraco.Core.Upgrader.State+Umbraco.Core` to be a previous step found in the `UmbracoPlan.cs`. For example, use the step before the final: "{2AB29964-02A1-474D-BD6B-72148D2A53A2}" 
* go to the back office, it will redirect to the authorize upgrade page, ensure it works, login and it will redirect to the upgrader

As for the invite user i think we'll need to test that later but the code is basically identical, almost no changes.

### DIFF
--- a/src/Umbraco.Web.BackOffice/Controllers/BackOfficeController.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/BackOfficeController.cs
@@ -88,7 +88,7 @@ namespace Umbraco.Web.BackOffice.Controllers
         public async Task<IActionResult> VerifyInvite(string invite)
         {
             //if you are hitting VerifyInvite, you're already signed in as a different user, and the token is invalid
-            //you'll exit on one of the return RedirectToAction("Default") but you're still logged in so you just get
+            //you'll exit on one of the return RedirectToAction(nameof(Default)) but you're still logged in so you just get
             //dumped at the default admin view with no detail
             if (_webSecurity.IsAuthenticated())
             {
@@ -138,7 +138,7 @@ namespace Umbraco.Web.BackOffice.Controllers
             //sign the user in
             DateTime? previousLastLoginDate = identityUser.LastLoginDateUtc;
             await _signInManager.SignInAsync(identityUser, false);
-            //reset the lastlogindate back to previous as the user hasn't actually logged in, to add a flag or similar to SignInManager would be a breaking change
+            //reset the lastlogindate back to previous as the user hasn't actually logged in, to add a flag or similar to BackOfficeSignInManager would be a breaking change
             identityUser.LastLoginDateUtc = previousLastLoginDate;
             await _userManager.UpdateAsync(identityUser);
 
@@ -158,7 +158,7 @@ namespace Umbraco.Web.BackOffice.Controllers
             return await RenderDefaultOrProcessExternalLoginAsync(
                 //The default view to render when there is no external login info or errors
                 () => View(viewPath),
-                //The ActionResult to perform if external login is successful
+                //The IActionResult to perform if external login is successful
                 () => Redirect("/"));
         }
 
--- a/src/Umbraco.Web.BackOffice/Controllers/BackOfficeController.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/BackOfficeController.cs
@@ -13,14 +13,17 @@ using Umbraco.Core.Cache;
 using Umbraco.Core.Configuration;
 using Umbraco.Core.Configuration.Grid;
 using Umbraco.Core.Hosting;
+using Umbraco.Core.Logging;
 using Umbraco.Core.Services;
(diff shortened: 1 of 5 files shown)


src/Umbraco.Web.BackOffice/Controllers/BackOfficeController.cs
```diff
@@ -74,6 +84,84 @@ public async Task<IActionResult> Default()
...
+                await _signInManager.SignOutAsync();
+            }
+
+            if (invite == null)
+            {
+                _logger.Warn<BackOfficeController>("VerifyUser endpoint reached with invalid token: NULL");
+                return RedirectToAction(nameof(Default));
+            }
+
+            var parts = System.Net.WebUtility.UrlDecode(invite).Split('|');
```
elit0451, line 104: I can see that we have been using both `WebUtility.UrlEncode/UrlDecode` or `HttpUtility.UrlEncode/UrlDecode` throughout the code. Would be nice to just stick to, for example, `WebUtility.UrlEncode/UrlDecode` as there is a difference in the casing for the encoded characters. 

For Windows, it does not matter but in Linux they are different

elit0451:
Hey @Shazwazza 

Checked the code, it looks fine. Have spotted a few minor outdated comments/unused var that I have fixed now!

BUT.. even though the test scenario that you suggested works, I am able to access the backoffice without any authentication. Tried running on both IIS Express and Umbraco.Web.UI.Netcore project -> same issue. Even after stopping the site on IIS Express or closing the executable console, running again will **not** end up on the Login page.

Any ideas what have caused this?

src/Umbraco.Web.BackOffice/Controllers/BackOfficeController.cs
Shazwazza: We cannot use HttpUtility, that is part of the System.Web namespace which is .net framework only.

Shazwazza:
Hi @elit0451 ,

> I am able to access the backoffice without any authentication

So you never get prompted to login? Even after you logout? What if you clear your cookies? There's nothing in this code change that would affect this so if this is a problem (i cannot replicate) i don't think it's due to this code, but can you confirm?

src/Umbraco.Web.BackOffice/Controllers/BackOfficeController.cs
elit0451: Think we can still use it in .net core - [link](https://docs.microsoft.com/en-us/dotnet/api/system.web.httputility.urlencode?view=netcore-3.1) -> for example we use it in **Umbraco.Web.BackOffice** (for some controllers or extensions) and in **Umbraco.Web.Website** (for action result) projects

elit0451:
Not quite, I get the login screen and the problem happens after I log in. After logout or clearing the cookies, I am presented with the login page but once more 1 login is needed and from there on stopping and running the website allows me to access the backoffice without credentials 🤷

src/Umbraco.Web.BackOffice/Controllers/BackOfficeController.cs
Shazwazza: We must absolutely not use it in our .net core projects. This is problematic! I'm unsure how this has happened. All references to `using System.Web;` in our netcore/netstandard projects must not exist. @bergmania did you know about this? We still have references to `System.Web` in netcore/netstandard projects 😥

bergmania: I was not aware... From what I can see, it is luckily only `HttpUtility` that is used. I think the UrlEncode/UrlDecode can be replaced one to one, but we need to find an alternative to `HttpUtility.ParseQueryString`

Shazwazza:
@elit0451 

> Not quite, I get the login screen and the problem happens after I log in. After logout or clearing the cookies, I am presented with the login page but once more 1 login is needed and from there on stopping and running the website allows me to access the backoffice without credentials

Sounds like this problem isn't related to this PR?

elit0451:
Haven’t seen it in the past couple of days but perhaps.. or maybe it is a problem just in my set up. As I said code and tests looks alright apart from that, you can merge if you decide to :)

Shazwazza:
@bergmania AFAIK the login issues isn't caused by this PR but we should investigate the login issues reported here. Perhaps you can replicate?

bergmania:
I create a task to investigate

PR #8574, opened 2020-08-04
append lost drag class

nielslyngsoe:
The "blockelement__draggable-element" class was present on the scoped block list block components before validation was merged in, but it somehow got lost. Its appended cause the scoping makes it unable to search and find it inside the scoped component, therefore we need to make the full entry draggable.


warrenbuckley:
Trivial change - just verifying its fixed

PR #8578, opened 2020-08-04
Don't retain "move media" in busy state on error

kjac:
### Prerequisites

### Description

If you attempt to move a media item to its parent folder, the operation will fail with a big ol' red error message. But the cancel/move buttons are never re-enabled, thus leaving the editor with no visible ways out (you can still click outside the dialog to close it, but still... not nice):

image

This PR ensures that the "busy" state (which removes the cancel button and disables the move button) is not retained when an error occurs during the move operation:

media-move-to-same-parent-after


nul800sebastiaan:
Ha! Well that's silly! Good update, cheers @kjac! 👍

PR #8582, opened 2020-08-04
Fix the arrow direction in the content language picker

kjac:
### Prerequisites

### Description

The introduction of SVG icons has had a slight unintended side effect: The language selector arrow always points to the right:

language-selector-arrow-direction-before

This PR fixes it:

language-selector-arrow-direction


nul800sebastiaan:
Oh very good, nice one @kjac! 👍 🔼 ▶ 🔽

PR #8584, opened 2020-08-04
Netcore: Added request localization from the current user.

bergmania:
## Details
Added a custom `RequestCultureProvider` to use the BackOffice user's language as culture.

Even that we already call `EnsureCulture` from the `OnValidatePrincipal` event. This is not enough, as the culture is reset when the thread hits the controller. 

## Test
1. Set the users language to a non-English language. E.g. danish.
1. Sign out
1. Sign in again
1. Verify the localization are updated. 

Note: I tested with multiple users with different languages.

### DIFF
--- a/src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
+++ b/src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
@@ -156,24 +156,7 @@ namespace Umbraco.Extensions
                 out factory);
 
             return services;
-        }
-
-        public static IServiceCollection AddUmbracoRequestLocalization(this IServiceCollection services)
-        {
-            services.Configure<RequestLocalizationOptions>(options =>
-            {
-                var supportedCultures = CultureInfo
-                    .GetCultures(CultureTypes.AllCultures & ~ CultureTypes.NeutralCultures)
-                    .Where(cul => !String.IsNullOrEmpty(cul.Name))
-                    .ToArray();
-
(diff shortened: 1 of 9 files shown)


src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
```diff
@@ -156,6 +158,23 @@ public static IServiceCollection AddUmbracoCore(this IServiceCollection services
             return services;
         }
 
+        public static IServiceCollection AddUmbracoRequestLocalization(this IServiceCollection services)
+        {
+            services.Configure<RequestLocalizationOptions>(options =>
```
Shazwazza, line 163: So that we don't conflict with global aspnet things i think we need to sub-class this so that it's BackOfficeRequestLocalizationOptions just like we do for all other services. This will also mean looking at the RequestLocalizationMiddleware  since we may need to sub-class that and have our own as well. We can't just globally set these options since this affects every request that goes to a site, we need to only apply these types of things to our own requests and not affect global settings. Just like everything else the aspnetcore team hasn't made that perfectly flexible wo we'll need to work around it. 

I'm happy to look into this tonight?

bergmania: You are right - Forgot about that.. You are very welcome to make the changes 💪

Shazwazza:
@bergmania have pushed a commit, all my tests work and my back office is localized correctly

bergmania:
Changes looks good :) 💪

PR #8585, opened 2020-08-04
Update content type icon on save

bjarnef:
### Prerequisites

### Description
Now that content types (document, media and member types) use the specific content type icon in tree, it would be great if the node icon updated if icon is changing (icon or color).

With the PR it update the node icon in tree if the content type icon is different from the current content type icon. At the moment the event name is `editors.tree.icon.changed` - not sure if we have something better?


poornimanayar (approved):
Hi @bjarnef ,

This totally works! I have tested it and it works well. Many thanks for your work :-)

Poornima

poornimanayar:
And merged! Thanks once again!

PR #8587, opened 2020-08-04
Update user filter buttons

bjarnef:
### Prerequisites

### Description
This PR updates the filters in user section to use `<button>` elements instead of anchor elements.
Furthermore the markup and inline styling was re-used a few places and in log viewer filter, where it was a bit different from filters in users section.

I have added some shared styling in a `umb-filter.less` file. At the moment we still need to ensure the markup is identical, but my thought is we can make this easier (and easier to package developers to re-use) by adding a `umb-filter` component.

So for example something like this:

```
<div class="umb-filter">
    <button type="button"
            class="btn btn-link dropdown-toggle umb-filter__toggle"
            ng-click="vm.page.showLevelFilter = !vm.page.showLevelFilter"
            aria-haspopup="true"
            aria-expanded="{{vm.page.showLevelFilter === undefined ? false : vm.page.showLevelFilter}}">

        <span><localize key="logViewer_logLevels">Log Levels</localize>:</span>
        <span class="bold truncate dib umb-filter__label">{{ vm.getFilterName(vm.logLevels) }}</span>
        <span class="caret" aria-hidden="true"></span>
    </button>
    <umb-dropdown class="pull-left" ng-if="vm.page.showLevelFilter" on-close="vm.page.showLevelFilter = false;">
        <umb-dropdown-item ng-repeat="level in vm.logLevels" class="dropdown-item">
            <div class="flex items-center">
                <umb-checkbox input-id="loglevel-{{$index}}" name="loglevel" model="level.selected" on-change="vm.setLogLevelFilter(level)"></umb-checkbox>
                <label for="loglevel-{{$index}}">
                    <umb-badge size="s" color="{{level.logTypeColor}}">{{level.name}}</umb-badge>
                </label>
            </div>
        </umb-dropdown-item>
    </umb-dropdown>
</div>
```

could be simplified to this:

```
<umb-filter name="Groups" multiselect="true" on-select="vm.select()">
    <umb-filter-item>Administrators</umb-filter-item>
    <umb-filter-item>Editors</umb-filter-item>
    <umb-filter-item>Translators</umb-filter-item>
    <umb-filter-item>Writers</umb-filter-item>
</umb-filter>
```

The `multiselect` property could decide whether the component should render the `<umb-checkbox>` or a `<button>` as in **Order By** filter.

However this would be in another PR.

For this PR the changes looks like this:


bjarnef:
@nielslyngsoe what do you think of a `umb-filter` component? wouldn't it make it much cleaner and ensure consistency? 😎

nul800sebastiaan:
Yay, lovely changes here @bjarnef and I totally agree with an `umb-filter` component, good idea! 👍

PR #8590, opened 2020-08-05
Netcore: Get rid of System.Web in netcore and netstandard projects

bergmania:
- Replaced `System.Web.HttpUtility` with `System.Net.WebUtility`
- Replaced querystrings as `String`s and `NameValueCollection`s with the `QueryString` type


Shazwazza:
@bergmania file changes look good and tests pass, will merge and re-test. I find it strange that the solution compiled with `using System.Web;` declarations when the projects are not specifically referencing any net framework assemblies. That seems like a bug to me in netcore?

PR #8591, opened 2020-08-05
Netcore: Let build.ps1 create the dotnet new templates

bergmania:
# Details
When executing the `build.ps1` script, it now also creates an `Umbraco.Templates` nuget package. This package can be used to install the templates using `dotnet new -i Umbraco.Templates`.

The templates package contains one package
1. `umbraco`
    - An empty umbraco solution, but with options to add SqlCE (Use `dotnet new umbraco --help` to see how later)
 
The templates are build using the following current files from the `Umbraco.Web.UI.Netcore` project
- `Program.cs`
- `Startup.cs`
- `appsettings.json`
- `appsettings.Development.json`

These files are combined with some static resources, for the templates located in `~/build/templates`

# Tests
To test this you can type `dotnet new -i Umbraco.Templates.9.0.0.nuspec` form the `build.out` folder after the `build.ps1` script is executed.

Then you install a project using `dotnet new umbraco -n {Name}`

To test the generated project, you need to have your NuGet sources to point to the `build.out` folder to find Umbraco in version 9.0.0


build/build.ps1
```diff
@@ -326,10 +328,23 @@
     $this.CopyFiles("$src\Umbraco.Web.UI.NetCore\wwwroot\umbraco\js", "*", "$tmp\WebApp\wwwroot\umbraco\js")
     $this.CopyFiles("$src\Umbraco.Web.UI.NetCore\wwwroot\umbraco\lib", "*", "$tmp\WebApp\wwwroot\umbraco\lib")
     $this.CopyFiles("$src\Umbraco.Web.UI.NetCore\wwwroot\umbraco\views", "*", "$tmp\WebApp\wwwroot\umbraco\views")
+
+
+
+    # Prepare templates
+    Write-Host "Copy template files"
+    $this.CopyFiles("$templates", "*", "$tmp\Templates")
```
elit0451, line 336: The bin folder gets copied too. We can exclude it by adding a line below ⬇️

build/build.ps1
```diff
@@ -326,10 +328,23 @@
...
+
+    Write-Host "Copy program.cs and startup.cs for templates"
+    $this.CopyFiles("$src\Umbraco.Web.UI.NetCore", "Program.cs", "$tmp\Templates\UmbracoSolution")
+    $this.CopyFiles("$src\Umbraco.Web.UI.NetCore", "Startup.cs", "$tmp\Templates\UmbracoSolution")
+    $this.CopyFiles("$src\Umbraco.Web.UI.NetCore", "appsettings.json", "$tmp\Templates\UmbracoSolution")
+    $this.CopyFiles("$src\Umbraco.Web.UI.NetCore", "appsettings.Development.json", "$tmp\Templates\UmbracoSolution")
   })
```
elit0451, line 343:
```suggestion

  $this.RemoveDirectory("$tmp\Templates\UmbracoSolution\bin")
  })
```

src/Umbraco.Tests.Integration/TestServerTest/UmbracoWebApplicationFactory.cs
```diff
@@ -20,7 +20,7 @@
 using Umbraco.Core.Models.Membership;
 using Umbraco.Core.Services;
 using Umbraco.Web.Common.Security;
-using Umbraco.Web.UI.BackOffice;
+using Umbraco.Web.UI.NetCore;
 
 namespace Umbraco.Tests.Integration.TestServerTest
```
elit0451, line 25: This change slipped into this PR 😁

elit0451:
Overall, the changes look good! 

But when I tried installing the 2 templates (**umbraco** and **umbraco -U**), the elements on the login page and the back-office haven't quite rendered _/check picture ⬇️ /_. Seems like we are missing something on the UmbracoCms nuget package. 
Does it happen for you too?

image

----

And also the solution that I get from the **umbraco -U** contains **amd64** and **x86** folders 🤔 

image

src/Umbraco.Tests.Integration/TestServerTest/UmbracoWebApplicationFactory.cs
bergmania: This is expected.. The namespace of startup.cs and program.cs was wrong.

bergmania:
Based on the screenshot, it looks like the `umbraco/config/lang` folder is not there. I will investigate, as I was sure I fixed this :)

bergmania:
The `amd64 ` and `x86` folders is expected. These are part of SqlCE and required

build/build.ps1
bergmania: Good catch

elit0451:
Happy with the changes 💪 Merging

PR #8592, opened 2020-08-05
Do not require the umbVariantContent in the context of the DocumentTypeEditor

nielslyngsoe:
When preview the Block List editor in the DocumentTypeEditor we do not have umbVariantContent available which resulted in this error: "Could not find umbVariantContent in the $scope chain".
But we can use "vm.umbProperty" to check whether we are in this weird context of the DocumentTypeEditor, and then not start the scope searching.


warrenbuckley:
Code change seems fine - waiting for the build to complete before its merged

PR #8594, opened 2020-08-06
Fix server validation issue with block editor due to cultures not being carried over correctly

Shazwazza:
This fixes an issue where property validators were not being cleared for server errors. The problem was because we were not transferring the correct culture values to the sub-block properties.

## Testing

* Make sure server validation works for invariant content for both content/settings and that validation errors clear when you change data
* Make sure server validation works for variant content but for an invariant block editor property for both content/settings and that validation errors clear when you change data. Test without split view open and with split view open.
* Make sure server validation works for variant content for an variant block editor property for both content/settings and that validation errors clear when you change data. Test without split view open and with split view open.


nielslyngsoe:
Very good! Thanks!

PR #8597, opened 2020-08-06
Fix to ensure we use virtual paths for any blocks with custom views

warrenbuckley:
Fixes AZ#7611
Will use a virtual path in case Umbraco put in an unusual place

## Test Notes
* Add a block to block list property editor that uses a custom HTML view
* Ensure it renders
* Add a block without a custom HTML view - ensure that renders the default one

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/umbrequesthelper.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/umbrequesthelper.service.js
@@ -19,20 +19,18 @@ function umbRequestHelper($http, $q, notificationsService, eventsService, formHe
          * @param {string} a virtual path, if this is already an absolute path it will just be returned, if this is a relative path an exception will be thrown
          */
         convertVirtualToAbsolutePath: function(virtualPath) {
-            // This will convert a local path to virtual path (i.e. /App_Plugins/Blah/Test.html ) to an virtual path (i.e. ~/App_Plugins/Blah/Test.html)
-            if(virtualPath.startsWith("~/") === false) {
-                if(virtualPath.startsWith("/")) {
-                    virtualPath =  "~" + virtualPath;
-                } else {
-                    virtualPath = "~/" + virtualPath;
-                }
+            if (virtualPath.startsWith("/")) {
+                return virtualPath;
             }
-
+            if (!virtualPath.startsWith("~/")) {
+                throw "The path " + virtualPath + " is not a virtual path";
+            }
             if (!Umbraco.Sys.ServerVariables.application.applicationPath) {
                 throw "No applicationPath defined in Umbraco.ServerVariables.application.applicationPath";
             }
             return Umbraco.Sys.ServerVariables.application.applicationPath + virtualPath.trimStart("~/");
         },
+
 
         /**
          * @ngdoc method
--- a/src/Umbraco.Web.UI.Client/src/common/services/umbrequesthelper.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/umbrequesthelper.service.js
@@ -19,39 +19,19 @@ function umbRequestHelper($http, $q, notificationsService, eventsService, formHe
          * @param {string} a virtual path, if this is already an absolute path it will just be returned, if this is a relative path an exception will be thrown
          */
         convertVirtualToAbsolutePath: function(virtualPath) {
-            if (virtualPath.startsWith("/")) {
-                return virtualPath;
+            // This will convert a local path to virtual path (i.e. /App_Plugins/Blah/Test.html ) to an virtual path (i.e. ~/App_Plugins/Blah/Test.html)
+            if(virtualPath.startsWith("~/") === false) {
+                if(virtualPath.startsWith("/")) {
+                    virtualPath =  "~" + virtualPath;
+                } else {
+                    virtualPath = "~/" + virtualPath;
+                }
             }
-            if (!virtualPath.startsWith("~/")) {
-                throw "The path " + virtualPath + " is not a virtual path";
-            }
+
             if (!Umbraco.Sys.ServerVariables.application.applicationPath) {
                 throw "No applicationPath defined in Umbraco.ServerVariables.application.applicationPath";
             }
             return Umbraco.Sys.ServerVariables.application.applicationPath + virtualPath.trimStart("~/");
-        },
-
-        /**
-         * @ngdoc method
-         * @name umbraco.services.umbRequestHelper#ensureVirtualPath
-         * @methodOf umbraco.services.umbRequestHelper
-         * @function
-         *
-         * @description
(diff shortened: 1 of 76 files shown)


src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js
```diff
@@ -220,8 +220,13 @@
                 });
             }
 
-            // TODO: Why is there a '/' prefixed? that means this will never work with virtual directories
-            block.view = (block.config.view ? "/" + block.config.view : getDefaultViewForBlock(block));
+            if(block.config.view) {
+                block.config.view = umbRequestHelper.ensureVirtualPath(block.config.view);
```
Shazwazza, line 224: This shouldn't need to be done in 2 calls, IMO the convertVirtualToAbsolutePath should implicitly call ensureVirtualPath so there's no need to explicitly call that.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbblocklistblock.component.js
```diff
@@ -48,8 +48,7 @@
             $scope.valFormManager = model.valFormManager;
 
             if (model.stylesheet) {
-                // TODO: Not sure why this needs a prefixed /? this means it will never work in a virtual directory
-                model.stylesheet = "/" + model.stylesheet;
+                model.stylesheet = umbRequestHelper.ensureVirtualPath(model.stylesheet);
```
Shazwazza, line 51: I assume somewhere along the way that `model.stylesheet` then gets converted with umbRequestHelper.convertVirtualToAbsolutePath ?

warrenbuckley: No this got missed - good catches will fix up

nielslyngsoe:
@Shazwazza now ensureVirtualPath is part of convertVirtualToAbsolutePath, meaning that convertVirtualToAbsolutePath does not fire errors anymore, since the conversion would cover those cases. I´m not sure this is right, but that's what you basically asked for, I´m not sure what correct in this case, what are the options, and why do we even fix this in the client. why doesn't the C# output something useful? That my thoughts, let me know what you think.

src/Umbraco.Web.UI.Client/src/common/services/umbrequesthelper.service.js
```diff
@@ -15,16 +15,19 @@ function umbRequestHelper($http, $q, notificationsService, eventsService, formHe
...
+         *
          * @param {string} a virtual path, if this is already an absolute path it will just be returned, if this is a relative path an exception will be thrown
          */
         convertVirtualToAbsolutePath: function(virtualPath) {
-            if (virtualPath.startsWith("/")) {
-                return virtualPath;
-            }
-            if (!virtualPath.startsWith("~/")) {
-                throw "The path " + virtualPath + " is not a virtual path";
+            // This will convert a local path to virtual path (i.e. /App_Plugins/Blah/Test.html ) to an virtual path (i.e. ~/App_Plugins/Blah/Test.html)
```
clausjensen, line 22: I don't really get why we're doing this?
We take the `virtualPath` and we do these to things to ensure it conforms to a format of `~/App_Plugins/` etc. 
Then we take this and trim away that part we just made sure was there (`~/`) and then concat it with the `applicationPath`.

Wouldn't it be enough to simply do this:
``` 
if (!Umbraco.Sys.ServerVariables.application.applicationPath) {
    throw "No applicationPath defined in Umbraco.ServerVariables.application.applicationPath";
}
return Umbraco.Sys.ServerVariables.application.applicationPath + virtualPath.trimStart("~").trimStart("/");
```

clausjensen: Scratch that... there's something going on in this whole PR that isn't really right.

src/Umbraco.Web.UI.Client/src/common/services/umbrequesthelper.service.js
```diff
@@ -15,16 +15,19 @@ function umbRequestHelper($http, $q, notificationsService, eventsService, formHe
          *
          * @description
          * This will convert a virtual path (i.e. ~/App_Plugins/Blah/Test.html ) to an absolute path
-         * 
+         *
          * @param {string} a virtual path, if this is already an absolute path it will just be returned, if this is a relative path an exception will be thrown
          */
         convertVirtualToAbsolutePath: function(virtualPath) {
```
clausjensen, line 21: We can't really make this change as the method is no longer doing what it should be doing.
Before the change, this method would throw an exception given a relative path. It no longer does. Instead it just converts this into a (most likely) wrong/non-existing path `Umbraco.Sys.ServerVariables.application.applicationPath + relativePath`.

src/Umbraco.Web.UI.Client/src/views/packages/options.controller.js
```diff
@@ -21,13 +21,6 @@
 
                 vm.loading = false;
 
-                //make sure the packageView is formatted as a virtual path
```
clausjensen, line 24: This should not be moved to the `convertVirtualToAbsolutePath` method, as that method should not be handling relative paths. It might work for this, but if used elsewhere that method would need to throw exceptions and not just "handle" a relative path.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js
```diff
@@ -220,8 +220,12 @@
                 });
             }
 
-            // TODO: Why is there a '/' prefixed? that means this will never work with virtual directories
-            block.view = (block.config.view ? "/" + block.config.view : getDefaultViewForBlock(block));
+            if(block.config.view) {
+                block.config.view = umbRequestHelper.convertVirtualToAbsolutePath(block.config.view);
```
clausjensen, line 224: We should not be pushing back this to config. If the config is wrong - well then it should not work and a fix should be applied to make sure it doesn't contain bad data, instead of simply "fixing things" here if the data is wrong.

clausjensen:
It would also be really great to add a unit test to this, so we're sure a change like this would be caught by it in the future. Testing with relative, absolute and virtual paths.

nielslyngsoe:
Code seems good and I have corrected Pre Value editor to store the correct value. Asked @warrenbuckley to do testing on Windows Filesystem.

warrenbuckley:
Yep @nielslyngsoe it works, as expected on Windows & JS tests are all happy.
Can you fix up the merge conflict for me & get it merged in, please

Thumbs Up

PR #8601, opened 2020-08-06
Fixing publish descendants dialog

clausjensen:
### Description
The publish descendants dialog was broken at some point going from 8.6->8.7.

I've fixed it up, using the label on the toggle directive instead of using a separate label.

Also updated the localized texts to better reflect what is happening here:
- The toggle allows you to include or exclude publishing of items that have a published state.
- The old help text could indicate that items in a published state but with a "draft" on top would not be published if deselected ... they would be - so this is now more clear.

The last fix done here is that the selected value would not be pushed back to the scope which means your selection would never make it back.

---
_This item has been added to our backlog [AB#7682](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7682)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/overlays/publishdescendants.html
+++ b/src/Umbraco.Web.UI.Client/src/views/content/overlays/publishdescendants.html
@@ -6,15 +6,13 @@
         </div>
 
         <div class="flex mb3">
-
             <umb-toggle checked="vm.includeUnpublished"
                         on-click="vm.toggleIncludeUnpublished()"
                         class="mr2"
                         label-on="{{vm.labels.includeUnpublished}}"
                         label-off="{{vm.labels.includeUnpublished}}"
                         label-position="right"
-                        show-labels="true"/>
-
+                        show-labels="true"></umb-toggle>
         </div>
 
     </div>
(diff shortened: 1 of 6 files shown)


clausjensen:
When merged, should also be cherrypicked for 8.7 as it is a regression.

src/Umbraco.Web.UI.Client/src/views/content/overlays/publishdescendants.html
```diff
@@ -6,12 +6,15 @@
...
-            <localize key="content_includeUnpublished"></localize>
+
+            <umb-toggle checked="vm.includeUnpublished"
+                        on-click="vm.toggleIncludeUnpublished()"
+                        class="mr2"
+                        label-on="{{vm.labels.includeUnpublished}}"
+                        label-off="{{vm.labels.includeUnpublished}}"
+                        label-position="right"
+                        show-labels="true"/>
+
```
bjarnef, line 17: @clausjensen notice angular directives/components can't be void elements (self-closing), so it needs a end-closing `</umb-toggle>`.

There is a great explanation of it here: https://stackoverflow.com/a/25012451/1693918

Also here is an example of what can happen when a component/directive is self-closed. Of course native HTML5 elements can still be self-closed like ``, `<br />` etc.

Here is a PR updating this for `<umb-checkbox>` and `<umb-radiobutton>`, but should also be done for `<umb-toggle>`.
https://github.com/umbraco/Umbraco-CMS/pull/8547

bjarnef:
@clausjensen I have submitted the following PR: https://github.com/umbraco/Umbraco-CMS/pull/8605

But it seems it was only in this overlay `<umb-toggle>` was void element. In the PR I removed some whitespace where toggle was used in macros, but since you have localized the text in the controller it should be fine and don't need the `<localize>` directive in view. It is just missing the  end-closing `</umb-toggle>` instead.

src/Umbraco.Web.UI.Client/src/views/content/overlays/publishdescendants.html
clausjensen: I've merged your PR as it might be good just for history .. and the whitespace change is fine. Merging this afterwards will overwrite some things in yours which is fine.

clausjensen:
I've removed version tag on this as I'm fairly sure this was never actually released in a release.

Issue #8602, opened 2020-08-06
Localize element not rendering after umb-toggle


clausjensen:
## Umbraco version

Latest v8/dev.

Reproduction
------------

Well... this is a bit funky, but do this:

- Edit `publishdescendants.html` (it's in `views/content/overlays`).
- Locate the first `umb-toggle` element in there and place this on the line directly above it:
`<localize key="content_includeUnpublished"></localize>`
- Open a content node, select to publish descendants and verify that you see this text now showing up in the dialog.
- Now go back to the file and move the localize tag _below_ the `umb-toggle` directive.
- Open the same dialog ... and verify that the localized text no longer renders.

I haven't tested if this applies to other directives as well ... I just noticed it with the toggle.

### Bug summary

Localized text/tag seems to not render when put in _after_ a umb-toggle element. It works fine if placed before umb-toggle.

clausjensen:
Here's a challenge for you @kjac 😄 Have fun!

_(I've got no clue what is going on here!!!)_

bjarnef:
@clausjensen I think this is because of the `<umb-toggle>` is added as a void element (self-closing element), which angular directives/components can't be as they are custom HTML elements.

It's a similar issue as in https://github.com/umbraco/Umbraco-CMS/pull/8478

There is a great explanation of it here: https://stackoverflow.com/a/25012451/1693918

I have also added a comment here: https://github.com/umbraco/Umbraco-CMS/pull/8601#discussion_r466515477

We should ensure all `<umb-toggle>` use an end-closing `</umb-toggle>`.

I have already updated this for `<umb-checkbox>` and `<umb-radiobutton>`, but should also be done for <umb-toggle>.
#8547

clausjensen:
I had no idea about the thing with closing elements .. Great find 😄 👍 and thanks!!

PR #8605, opened 2020-08-06
Fix self closing umb-toggle

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8602

### Description
The `umb-toggle` is a custom angular directive, where custom angular directives/components can't be void elements (self-closing), so it needs to be non-void element - in this case has a end-closing `</umb-toggle>`.

There is a great explanation of it here: https://stackoverflow.com/a/25012451/1693918

**Before**

chrome_2020-08-06_19-50-02

**After**

image


bjarnef:
In https://github.com/umbraco/Umbraco-CMS/pull/8605/commits/deef02bd3265c981a2038eabbe6ae4119b935973 I have added a change so you can toggle the setting by clicking the label as well.

PR #8606, opened 2020-08-06
Various fixes for macroparameterpicker view

BatJan:
### Prerequisites

### Description
During the work I'm currently doing on #8522 I noticed that there were some issues in this view where tabbing between the items would have a weird double effect. This PR addresses those issues + hides the `<i>` element from screen readers and provides an English fallback text in case the dictionary key does not exist for a certain language.

In the gif's below I display what the filter acts like as well since it's 2 different parts of the code.

**Before**
macropickerview-before

**After**
macropickerview-after


nul800sebastiaan:
Thanks @BatJan, lovely work 👍

Issue #8596, opened 2020-08-06
Infinite editor when creating new items, does not handle variants


pgregorynz:
If you are using the editor service to open a new document, and you click save and close or publish and close it does not understand anything about variants.

## Umbraco version

I am seeing this issue on Umbraco version 8 (all versions)

Reproduction
------------

This is not so easy to reproduce unless you are creating a new content items using the editor service.  Here is the basics of the code that is used to create new items using the editor service.

```      
        var options = {
		create: true,
		documentTypeAlias: 'someTypeAlias',
		parentId: 1234,
		allowSaveAndClose: true,
		allowPublishAndClose: true,
		submit: function (model) {
			editorService.close();
		},
		close: function (model) {
			editorService.close();
		}
	};

	editorService.contentEditor(options);
```

### Bug summary

When you have more than one language variant and you are using the saveAndClose and publishAndClose option it will cause an error and the item will not be created.

### Expected result

Item should be saved and published and the editor pane close down.

### Actual result

It doesnt do as expected.

I have a PR for this issue that I will put forward to resolve this that takes into account language variations.

umbrabot:
Closing this in relation to PR #8607 - make sure to read the close reason on #8607

PR #8607, opened 2020-08-07
Fixes issues #8596 with infinite editor

pgregorynz:
When using the infinite editor with the publish and close or save and close option it would fail when there are language variants present.  This fix resolves this.

Fixes issue #8596

### Description

The editorService allows users to create plugins for umbraco that launch the infinite editor with the options for save and close and publish and close.  These two methods in the edit.controller.js file do not handle language variants.  This PR resolves that issue. 

There is a bit of duplication of code from the publish and save methods, and there could be a minor refactor to consolidate this code to simply the file however I did not want to introduce new bugs or issues into the main standard publishing workflows opting to duplicate the pattern of those methods but taking into account the infinite editor aspects related to "closing" the editor.

The following code snippet could be used to test.  Or a good example of where we are using this feature of umbraco is in the Bento editor which is what we need this fix in to resolve.

```
        var options = {
		create: true,
		documentTypeAlias: 'someTypeAlias',
		parentId: 1234,
		allowSaveAndClose: true,
		allowPublishAndClose: true,
		submit: function (model) {
			editorService.close();
		},
		close: function (model) {
			editorService.close();
		}
	};

	editorService.contentEditor(options);
```


nul800sebastiaan:
Hi @pgregorynz, sorry for the late reply, I was away for a few days. I am not super familiar with this part of the code so I'll need to find someone to look at it. 

The first thing I can see is that a lot of new code is introduced and that will make this harder to review. However you said there's some duplication going on, can you explain that? Instead of duplicating code we'd really want repeating code to be centralized so we don't forget to update it in one part of the system while forgetting about another part.

pgregorynz:
Hi Seb

I fully understand the position you have on this, and I would love to consolidate the code into a more DRY approach.  My main concern is that I didnt want to mess to much with the standard Publish or Save features of content but I will rework the pull request and resubmit as this part of the CMS will need resolution to enable some pretty cool features to work for infinite editing.  EG the possibility of listviews that can create and edit items without changing view in the CMS.

nul800sebastiaan:
Hi @pgregorynz - is this what you need help with? https://twitter.com/pgregorynz/status/1295982996479021056

I'd much rather mess "with the standard Publish or Save features of content" to make sure we don't have code that easily gets out of sync.

Anything that can help show off the functionality in a zip from an `App_Plugins ` folder would help, preferably not the whole of Bento but a simplified example. 

We'd love to help, the easier you make it for us, the easier you make it for you 🙂

pgregorynz:
@nul800sebastiaan I have updated the PR with my latest attempt to consolidate.  I have an issue that I am not sure how to overcome and would love some help.  I will work on an example App_Plugin however Bento with language variants demonstrates the issue perfectly but requires a bit of setup.  I will see if I can zip up the required bits (SQLCE db and bin/app_plugin bits).

nul800sebastiaan:
@pgregorynz it's in my list and I'll have a look at the merge issue soon 👍

nul800sebastiaan:
@pgregorynz Alright, merge issue resolved. We added more complex error handling for the block editor and that was causing some of your code to have to move! I recommend you go through it again and check that it all still works. Let us know how you get on!

PR #8613, opened 2020-08-10
UI for highlighting validation issues

nielslyngsoe:
Update for Validation UI changes how issues in content of a block are presented.

With the default look, the items will look like this:
image

With a custom view, we have this look instead:
image


warrenbuckley:
Code changes seem fine - even though CSS overly specific but as you told me Niels its to do with not effecting nested blocks styling & incorrectly highlighting valid blocks

warrenbuckley:
Tested and all OK

image

PR #8617, opened 2020-08-11
Block editor - Fix for when you have no access to settings

warrenbuckley:
### Test Notes
* Create a block list editor datatype on a content node
* Create a new user with NO access to the Settings section
* Add a block to the block list editor in a content node
* Verify you have a list of empty block cards with no names, description or icons
* Apply PR & retest to ensure name, icon & description is visible

### Problem
The problem was that we were trying to access JSON properties in DataType that would return null

### Fix
* Adds a new property `ContentTypeDescription` to SimpleContentType
* Updates JS to select the properties needed fro the Block Card component (Icon, Name & Description) which are available as root properties on the request made to `/umbraco/backoffice/UmbracoApi/Content/GetEmptyByKey?contentTypeKey=`

### DIFF
--- a/src/Umbraco.Core/Models/ISimpleContentType.cs
+++ b/src/Umbraco.Core/Models/ISimpleContentType.cs
@@ -15,6 +15,8 @@ namespace Umbraco.Core.Models
         /// </summary>
         string Alias { get; }
 
+        string Description { get; }
+
         /// <summary>
         /// Gets the default template of the content type.
         /// </summary>
--- a/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
@@ -447,7 +447,7 @@
                     if (scaffold) {
                         blocks.push({
                             blockConfigModel: blockConfiguration,
-                            elementTypeModel: scaffold.documentType
+                            elementTypeModel: {icon: scaffold.icon, name: scaffold.contentTypeName, description: scaffold.contentTypeDescription}
                         });
                     }
                 });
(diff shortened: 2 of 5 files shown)


nielslyngsoe:
I can't tell whether the code corrections are correct, but the fix solves the issue in my local test project. So I will say its work as expected.
@warrenbuckley maybe you can have Claus or Shannon review the code changes?

src/Umbraco.Core/Models/ISimpleContentType.cs
```diff
@@ -15,6 +15,8 @@ public interface ISimpleContentType : IUmbracoEntity
         /// </summary>
         string Alias { get; }
 
+        string Description { get; }
```
Shazwazza, line 18: Hrm, we'll have to see what implications this has since it's a breaking change. In theory we are allowed to change this in accordance with our breaking change policies so just need to see if people implement this... i don't think so, but could be wrong.

src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
```diff
@@ -447,7 +447,7 @@
                     if (scaffold) {
                         blocks.push({
                             blockConfigModel: blockConfiguration,
-                            elementTypeModel: scaffold.documentType
+                            elementTypeModel: {icon: scaffold.icon, name: scaffold.contentTypeName, description: scaffold.contentTypeDescription}
```
Shazwazza, line 450: So was scaffold.documentType null here? Where is scaffold.documentType assigned? I feel like all of this can be resolved without any of these changes by using the EntityController (i.e. entityResource) since you should be able to get this basic information about a Document Type there even without access to Settings. That is how most property editors work.

warrenbuckley: Yes scaffold.documentType was null from the request made to `/umbraco/backoffice/UmbracoApi/Content/GetEmptyByKey?contentTypeKey=XXX` but the top level properties were mostly there with the exception of the datatype description hence I added it.

warrenbuckley:
@Shazwazza closing this PR as easier to just do a new commit rather than revert what I have done already & opened this one instead - https://github.com/umbraco/Umbraco-CMS/pull/8627

As discussed on Slack today/Wednesday we thought that the security check in the model mapping seemed unnecessary

Issue #7552, opened 2020-02-01
Umbraco lack of Consistency for Camel/Pascal case in writing ID


bielu:
When I was looking into code I could notice the same aliases could be written in a different way as for example:
 **Parent ID**
Examine alias:
`"parentID"`
In same time DB:
`"ParentId"`
And reading the whole subject about how should be ID used in camel/pascal case this is quite complicated:
https://softwareengineering.stackexchange.com/questions/186277/should-the-variable-be-named-id-or-id
But still, we should be consistent about using aliases everywhere, it means if we are using ID we should use 
`"ParentID"` and `"parentID"` 
or for Id:
`"ParentId"` and `"parentId"` 

That issue is occurring for all aliases with the acronym "ID".

Code Examples:
`[DataMember(Name = "parentId")]`
` new FieldDefinition("parentID", FieldDefinitionTypes.Integer),`
`if (entity.IsPropertyDirty("ParentId"))`

markadrake:
I think some of the confusion comes from the English language, and well the rest may come from the varying standards in languages and frameworks. 

In writing, we should use capital letters. The term **ID** stands for _**identifier document**_ (or similar). The term _id_ is an abbreviation already for _idem_.

So what are the standards proposed by and followed for C#, .NET Framework, SQL, and don't forget JavaScript? Often it can depend on where the letters fall at. In Javascript, if the variable name is `id` then obviously, I would not capitalize it, nor camel case it, or anything - I'd allow it to stay lower case.

I'm with you @bielu, if it makes sense to update it for consistency while following the rules of syntax and language. If my opinion matters at all, in every case you made above, `ID` should be written as **ID**.

I often find myself wishing to ignore the [naming conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions) and want to capitalize other acronyms like `html` and `url` or `uri` and the such, but it's supposedly only for identifiers of 2 letters.

nul800sebastiaan:
Welcome to the world of legacy code! 

Unfortunately, we have a long history of naming things and they aren't consistent at the moment. It's a good point to keep in mind for the [Unicore project](https://umbraco.com/blog/the-unicore-team/) if we can so I've sent this off to the Unicore team for their consideration. For the current version, we won't be making any changes since this has the great potential to breaking all kinds of integrations where people are counting on the current inconsistencies.

bielu:
[...]

PR #8619, opened 2020-08-11
Update ContentTypeRepositoryBase.cs

erolakgul:
propertytypeid => propertyTypeId

### Prerequisites

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
@@ -1184,7 +1184,7 @@ AND umbracoNode.id <> @id",
         {
             // first clear dependencies
             Database.Delete<TagRelationshipDto>("WHERE propertyTypeId = @Id", new { Id = propertyTypeId });
-            Database.Delete<PropertyDataDto>("WHERE propertytypeid = @Id", new { Id = propertyTypeId });
+            Database.Delete<PropertyDataDto>("WHERE propertyTypeId = @Id", new { Id = propertyTypeId });
 
             // then delete the property type
             Database.Delete<PropertyTypeDto>("WHERE contentTypeId = @Id AND id = @PropertyTypeId",


erolakgul:
It is necessary you to approve the changes i have made in ContentTypeRepositoryBase class for turkish users. Many people can be use turkish_ci or turkish_cs collation their database. thank you..

ronaldbarendse:
Issue https://github.com/umbraco/Umbraco-CMS/issues/7552 contains a related discussion about changing all these different casings into constants and the conclusion is to do this in small PRs, so you're on the right track.

Regarding the issue you're trying to fix: this probably isn't the only column name that contains a different casing. You're better off using a database with the generic `Latin1_General_CI_AS` collation (see https://our.umbraco.com/forum/using-umbraco-and-getting-started/86852-umbraco-sql-collation-issue - keep in mind that changing the database collation doesn't change existing column collations though).

erolakgul:
Maybe you are right may be not. 

I dont agree with because;
In turkish, "I" and "i" are different characters. İ is upper case of i. Then  "I" is upper case of ı. If we chosed turkish_ci... it does not work 1187. line in ContentTypeRepositoryBase.cs because of propertytypeid. It must be  propertytypeid => propertyTypeId.

I agree with because;
If we chosed Latin1_General_CI_AS, it works.

> Issue #7552 contains a related discussion about changing all these different casings into constants and the conclusion is to do this in small PRs, so you're on the right track.
> 
> Regarding the issue you're trying to fix: this probably isn't the only column name that contains a different casing. You're better off using a database with the generic `Latin1_General_CI_AS` collation (see https://our.umbraco.com/forum/using-umbraco-and-getting-started/86852-umbraco-sql-collation-issue - keep in mind that changing the database collation doesn't change existing column collations though).

nul800sebastiaan:
Thanks @erolakgul, this change will be shipped with Umbraco 8.8.0! 

As Ronald noted, this is likely not the only problem you'll encounter but I hope this is enough for you for now. The related discussion details that it's going to be difficult to change all of the problems without making breaking changes, but updating the SQL query here is fine of course.

Looks like this is your first contribution to Umbraco CMS, congrats! 👍 

If you have an account on [Our Umbraco](https://our.umbraco.com) I'd love to give you a contributor badge, just send me a link to your profile!

PR #8620, opened 2020-08-11
Eliminates error/no-preview when grid media picker have no crop set

### LINKED ISSUE #8545: Console error when opening media picker crop from grid
With latest changes I noticed some console errors, when opening image crop via media picker in the grid editor.

It seems `scope.dimensions` has some `NaN` values in its nested object properties, e.g. scale `min` and `max`.

image

lars-erik:
Should fix #8545

Will see about moving the expressions into functions on the controller. Possibly just elimiate the bool flag set by RTE picker and check for crop config instead.


src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js
```diff
@@ -29,7 +29,6 @@ angular.module("umbraco")
             var dialogOptions = $scope.model;
 
             $scope.disableFolderSelect = (dialogOptions.disableFolderSelect && dialogOptions.disableFolderSelect !== "0") ? true : false;
-            $scope.disableFocalPoint = (dialogOptions.disableFocalPoint && dialogOptions.disableFocalPoint !== "0") ? true : false;
```
bjarnef, line 32: @lars-erik note that this is passed in from here and it would be a breaking change to remove as devlopers might use the infinite `mediapicker`.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js#L1564

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/overlays/mediacropdetails.controller.js
```diff
@@ -56,4 +58,14 @@
...
+            function cropNotSet() {
+                var model = $scope.model;
+                return !(model.cropSize || {}).width && model.target.thumbnail;
+            }
+
+            function cropSet() {
+                var model = $scope.model;
+                return (model.cropSize || {}).width && model.target.thumbnail;
+            }
+
```
bjarnef, line 70: Do we need two functions for this? Can't er just set a boolean scope property?
E.g. `vm.cropSet`, `vm.cropsExists` or similar?

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js
lars-erik: I'm aware, but it's not used for anything beyond this point. If you want the cropper, you provide a cropSize, otherwise you only get a preview. I don't think the focal point was ever saved. I'm hoping to look into whether it's possible to allow cropping from the RTE as well. In any case I think this field is obsolete and will remove it completely if we do the above.

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/overlays/mediacropdetails.controller.js
lars-erik: Not really, but then again "cropNotSet" is easier on the eye than "!cropSet". Also, if going on the path mentioned in the other comment, one of the two might get renamed, refactored and extended individually.

bjarnef: Sure, but we could use same function which return true or false.
You can then use two different scope properties or explicit specify the value, e.g. "cropSet === true" and "cropSet === false".

nul800sebastiaan:
I've updated this PR a little bit and we'll need to revisit it before the final release, but it mostly works now and we'll need to look at if the focal point is now missing somewhere that it shouldn't be missing.

Also note: it does not work when you add this config to a `package.manifest`. This is our attempt:

```
{
    "gridEditors": [
        {
            "name": "Updated Media with Config",
            "alias": "media.portrait2",
            "view": "media",
            "icon": "icon-picture",
            "config": {
                "width": 800,
                "height": 450
            }
        }
    ]
}
```

This config doesn't come through and you don't get a cropper. It's going to be a known issue for the RC, not a showstopper, but will need to be fixed for the final release.

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/overlays/mediacropdetails.html
```diff
@@ -26,15 +26,15 @@ <h5>
...
+                <div ng-if="vm.cropSet() === false">
                     <h5>
                         <localize key="general_preview">Preview</localize>
                     </h5>
 
                     <img ng-src="{{model.target.thumbnail}}" alt="{{model.target.name}}" />
                 </div>
 
-                <div ng-if="!model.disableFocalPoint">
+                <div ng-if="vm.cropSet()">
```
bjarnef, line 37: @nul800sebastiaan couldn't this just run the function in the controller?

```
vm.hasCrops = cropSet() === true;
```

Then use `vm.hasCrops` in view and remove this in controller `vm.cropSet = cropSet;`

With the current logic it needs to evaluate `cropSet()` twice.

nul800sebastiaan:
New issue for the missing part: https://github.com/umbraco/Umbraco-CMS/issues/8926

@bjarnef I will add that to the new issue!

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/overlays/mediacropdetails.html
nul800sebastiaan: Thanks @bjarnef - I've fixed this up now. https://github.com/umbraco/Umbraco-CMS/commit/e1ef5f4e6b6b28715d965a9a3d2dac951f251de6

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/overlays/mediacropdetails.controller.js
```diff
@@ -56,4 +57,8 @@
                 }
             }
 
+            function cropSet() {
+                var model = $scope.model;
+                return (model.cropSize || {}).width && model.target.thumbnail;
```
nul800sebastiaan, line 62: @lars-erik I didn't look very closely when merging this. This would return either an empty string or the `model.target.thumbnail` value. I've changed it to check for a configure width/height:  https://github.com/umbraco/Umbraco-CMS/commit/e1ef5f4e6b6b28715d965a9a3d2dac951f251de6#diff-c9ab1fa1ff35ffa6b4551dd5b51a95aeR62

I must admit I don't actually understand what the `|| {}` means in `(model.cropSize || {})`. I've left it for now, but can you clarify please?

nathanwoulfe: The `(x.y || {}).z` is the same as `x.y && x.y.z` - in cases where y is undefined, the RHS assigns an empty object so the check for z won't fail.

Given that this case is checking truthiness, not using the actual value of z, either way works the same - if we needed the value of z, it's a handy pattern.

Javascript does have optional chaining too, so we could do `x?.y?.z` if the build had the required transpiling for browser support...

nul800sebastiaan:
FYI: I've reintroduced the focal point, it was stored just fine and people might be using it. When you have an image without crops and edit it, the focal point now also gets shown (this never worked before).

https://github.com/umbraco/Umbraco-CMS/commit/df0222b09e5b86f2bb75439d5c8bccfaedf32bf8

nul800sebastiaan:
focal

src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/overlays/mediacropdetails.controller.js
nul800sebastiaan: @nathanwoulfe Thanks! I guess I kinda mixed and matched this in my update then 🙈 

https://github.com/umbraco/Umbraco-CMS/commit/e1ef5f4e6b6b28715d965a9a3d2dac951f251de6#diff-c9ab1fa1ff35ffa6b4551dd5b51a95aeR62

Are you saying that: `(model.cropSize || {}).width !== undefined` will give the same result as `(model.cropSize || {}).width`?

lars-erik: @nul800sebastiaan JavaScript works with truthy and falsy things.
Null, undefined, 0, undefined, "" etc. are all "falsy".
So you can just test if it's anything except those.
|| is just the same as C# ??.

nathanwoulfe: Should do, yes, thanks to javascript coercing to boolean, but I'd keep the version with the explicit comparison since it's easier to read.

lars-erik:
> FYI: I've reintroduced the focal point, it was stored just fine and people might be using it. When you have an image without crops and edit it, the focal point now also gets shown (this never worked before).
> 
> [df0222b](https://github.com/umbraco/Umbraco-CMS/commit/df0222b09e5b86f2bb75439d5c8bccfaedf32bf8)

I still have an issue with this making no sense at all to editors and should be discouraged.  
I'd still be very surprised if anyone _actually_ use it, but I respect the decision to let it be.  
Hopefully the "only" use case of media from grid will be with crop set.

nul800sebastiaan:
@lars-erik I know 😅 but we really want to make this upgrade as safe, none-breaking and predictable as possible. It was previously possible to move the focal point so I'm SURE people have queried it in the frontend somehow and will be unhappy if we remove the option 🙈

lars-erik:
@nul800sebastiaan I get it. Almost tempted to introduce "yet-a-setting" to disable it if you find it annoying when not having crops.

nul800sebastiaan:
@lars-erik Settings are evil! 🤬😁

bjarnef:
@nul800sebastiaan in case you didn't notice it seems  this can be simplified a bit: https://github.com/umbraco/Umbraco-CMS/commit/df0222b09e5b86f2bb75439d5c8bccfaedf32bf8

PR #8626, opened 2020-08-12
umb-editor-header name-required attribute

nielslyngsoe:
Added the option to define name-required, which sets required on the name-input. This defaults to true and therefor this does not change anything. unless the attribute is defined.

This fix corrects an issue with form-validation begin invalid when the name is an empty string, though the name is locked.


warrenbuckley:
Code change looks fine - need to test this

nielslyngsoe:
New solution...

PR #8627, opened 2020-08-12
(Rework) Block editor - Fix for when you have no access to settings

warrenbuckley:
## Test Notes
* Create a block list editor datatype on a content node
* Create a new user with NO access to the Settings section
* Add a block to the block list editor in a content node
* Verify you have a list of empty block cards with no names, description or icons
* Apply PR & retest to ensure name, icon & description is visible


warrenbuckley:
As discussed on Slack today/Wednesday we thought that the security check in the model mapping seemed unnecessary @Shazwazza

Shazwazza:
@warrenbuckley i trust that you ran those tests :) that code is much much simpler!!

PR #8630, opened 2020-08-13
Support include in sql server indexes

nzdev:
### Prerequisites

RE https://github.com/umbraco/Umbraco-CMS/pull/8365

### Description
Use the new IncludeColumns property in the same way as ForColumns property for the Index attribute during a db migration. This will allow for inclusion of columns that don't make up the key of an index


Shazwazza:
aha here was some info about the other DB SQL tasks https://github.com/umbraco/Umbraco-CMS/pull/8365#issuecomment-653046613 in case you were interested :)

PR #8631, opened 2020-08-13
Accept PublishedPendingChanges state as a published state

nielslyngsoe:
Ability to perform publish on a single variant when other variants is in the "PublishedPendingChanges" state.

Plus consider the name as data, if new this ensures to save that variant no matter the choice.

**Steps to reproduce UX problem:**

Have a variant document and two languages that both are mandatory.

save and publish content for the secondary language.
make new data for the secondary language and then just save it.
Refresh browser
Try to save and publish data for the primary language. (Before the secondary language would be required/mandatory, but it is already published)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/overlays/publish.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/overlays/publish.controller.js
@@ -78,7 +78,7 @@
         }
 
         function notPublishedMandatoryFilter(variant) {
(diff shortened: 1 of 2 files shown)


warrenbuckley:
Code looks fine, but had an issue repro before & after the PR applied @nielslyngsoe 
8631_before

warrenbuckley:
All fine & re-tested on a call with @nielslyngsoe

PR #8636, opened 2020-08-17
#6472 update the preview xml on publish actions

### LINKED ISSUE #6472: Preview mode unable to retrieve other pages from cache
### Bug summary

Since Umbraco 7.15.0 we are unable to use the preview function in at least 3 of our websites. While loading the previewed page I try to retrieve other pages using `content.AncestorOrSelf()` or `content.Siblings()` and I receive null or unexpected results (see screenshots).

### Specifics

- 7.15.0 up to 7.15.3
- The bug disappears when `Umbraco.Preview.Mode` is set to `Legacy`

RenderModel

Siblings

This is interesting, only previously previewed pages are showing up in the list but the page with ID 1050 should **not** be a sibling.

db-cmsPreviewXml

### Expected result

A webpage that loads in successfully without exceptions.

### Actual result

NullReferenceException in the code that runs fine outside the preview mode.

dawoe:
### Prerequisites

If there's an existing issue for this PR then this fixes #6472

### Description
This PR makes sure the preview xml get's updated when saving and publishing

Steps to reproduce and test can be found here : https://github.com/umbraco/Umbraco-CMS/issues/6472#issuecomment-576614477

Make sure that in step 11 and 12 you publish nodes C, D and the website node.

To test the fix the easiest is to build the umbraco zip package and copy the dll's over the provided test website.

Dave

### DIFF
--- a/src/Umbraco.Web/Cache/PageCacheRefresher.cs
+++ b/src/Umbraco.Web/Cache/PageCacheRefresher.cs
@@ -51,7 +51,11 @@ namespace Umbraco.Web.Cache
         public override void RefreshAll()
         {
             if (Suspendable.PageCacheRefresher.CanRefreshDocumentCacheFromDatabase)
+            {
                 content.Instance.RefreshContentFromDatabase();
+                content.Instance.ClearPreviewXmlContent();
+            }
+               
             ClearCaches();
             base.RefreshAll();
         }
@@ -63,7 +67,12 @@ namespace Umbraco.Web.Cache
         public override void Refresh(int id)
         {
             if (Suspendable.PageCacheRefresher.CanUpdateDocumentCache)
-                content.Instance.UpdateDocumentCache(id);
+            {
+                var d = new Document(id);
+                content.Instance.UpdateDocumentCache(d);
+                content.Instance.UpdatePreviewXmlContent(d);
+            }
+               
             ClearCaches();
             base.Refresh(id);
         }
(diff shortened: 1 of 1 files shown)


SteveVaneeckhout:
I checked Dave's fix and cannot reproduce the issue anymore.

src/Umbraco.Web/Cache/PageCacheRefresher.cs
```diff
@@ -75,23 +84,36 @@ public override void Refresh(int id)
...
             ClearCaches();
             base.Refresh(instance);
         }
 
         public override void Remove(IContent instance)
         {
             if (Suspendable.PageCacheRefresher.CanUpdateDocumentCache)
+            {               
                 content.Instance.ClearDocumentCache(new Document(instance), false);
+                content.Instance.ClearPreviewXmlContent(instance.Id);
```
ronaldbarendse, line 114: Any reason why `ClearDocumentCache` also doesn't use `instance.Id`, but creates a new instance of `Document`? Looks like the overload with the documentId does it a little bit safer: https://github.com/umbraco/Umbraco-CMS/blob/d11a1946f573f055c88a219bee3f7f1d6b058aa7/src/Umbraco.Web/umbraco.presentation/content.cs#L404-L420

dawoe: Don't know why. This was already in place, so I did not touch it.

nul800sebastiaan:
This looks great, thanks very much for the help @dawoe! 👍

PR #2476, opened 2018-02-25
(uDuf) Prevent unsaved dialogue appearing when no changes made on FileUpload

barryokane:
Addresses this issue in media section (U4-9996) and content (U4-7484)

Three things changed:

* FileUpload has a hidden field for 'fileCount' which is updated in controller & this was causing the issue : added no-dirty-check attribute
* Changed the no-dirty-check directive so the instead of relying on 'onFocus' event to override dirty value it is always false (also should avoid confusion if setPristine is ever called) : see http://issues.umbraco.org/issue/U4-4902#comment=67-43816 and https://stackoverflow.com/questions/17089090/prevent-input-from-setting-form-dirty-angularjs
* Explicitly call $setDirty in fileupload Angular controller when a fileSelected event is fired.

PR #8638, opened 2020-08-17
Allow to pass in boolean to noDirtyCheck directive

bjarnef:
### Prerequisites

### Description
At the moment we have some duplicated logic to be able to set `no-dirty-check` directive on e.g. `<umb-checkbox>` and `<umb-radiobutton>`.

I would however like just to pass in a boolean value whether to set this or not.

For example:

- `no-dirty-check` without attribute value set `$dirty` to false.
- `no-dirty-check="true"` with attribute value `true` works as before mentioned.
- `no-dirty-check="false"` with attribute value `true` works as default behaviour.

The directive was originally created based on this https://stackoverflow.com/a/27430980 but note sure why we need the local variable with getter and setter. Furthermore why it change `$pristine`. It seems this was changed in https://github.com/umbraco/Umbraco-CMS/pull/2476

I found some different variants setting `$dirty` to `false` or `angular.noop`:
https://stackoverflow.com/a/29225603

### Test notes

- Check that changes to input fields in content trigger the "Unsaved changes" dialog
- Add a custom dashboard with use on `no-dirty-check`,  `no-dirty-check="true"` and `no-dirty-check="true"` on `<input>` field works as expected. Also check with`<umb-checkbox>` and `<umb-radiobutton>` using `disable-dirty-check` property with `true/false` value.

```
<div>
    <label>Text 1</label>:<br>
    <input type="text" name="text1" placeholder="Enter value..." ng-model="text1" no-dirty-check />
</div>

<div>
    <label>Text 2</label>:<br>
    <input type="text" name="text2" placeholder="Enter value..." ng-model="text2" no-dirty-check="false" />
</div>

<div>
    <label>Text 3</label>:<br>
    <input type="text" name="text3" placeholder="Enter value..." ng-model="text3" no-dirty-check="true" />
</div>

<div>
    <label>Checkbox</label>:<br>
    <umb-checkbox model="check1" name="check1" disable-dirty-check="false">Test Check</umb-checkbox>
</div>

<div>
    <label>Radiobutton</label>:<br>
    <umb-radiobutton model="radio1" name="radio1" disable-dirty-check="true">Test Radio</umb-radiobutton>
</div>
```

[MyDashboard.zip](https://github.com/umbraco/Umbraco-CMS/files/5084038/MyDashboard.zip)


bjarnef:
@nul800sebastiaan @nielslyngsoe I believe this will clean up code a bit and we don't need some duplicated code.
Something similar could be done for `prevent-enter-submit`:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/forms/umb-search-filter.html#L5-L26

bjarnef:
However for `preventDefault` this seems to be used here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/preventdefault.directive.js#L33

bjarnef:
I have submitted another PR to fix the issue with `preventEnterSubmit`: https://github.com/umbraco/Umbraco-CMS/pull/8639

bjarnef:
@kjac maybe you want to test this? 😎

nul800sebastiaan:
This one is too intricate for me to evaluate, I will need some help from @kjac, @nathanwoulfe or if neither can help then @nielslyngsoe. Any takers? 🤞

kjac:
Sure I'll give it a spin in the morning 👍

src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
```diff
@@ -50,7 +50,7 @@
 
         function onInit() {
             vm.inputId = vm.inputId || "umb-check_" + String.CreateGuid();
-
+            vm.disableDirtyCheck = vm.disableDirtyCheck === true;
```
kjac, line 53: Is this necessary? And won't it break backwards compat with people using a truthly value that's not strictly `true`?

src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
```diff
@@ -49,7 +49,7 @@
 
         function onInit() {
             vm.inputId = vm.inputId || "umb-radio_" + String.CreateGuid();
-
+            vm.disableDirtyCheck = vm.disableDirtyCheck === true;
```
kjac, line 52: Same as above: Is this necessary? And won't it break backwards compat with people using a truthly value that's not strictly `true`?

src/Umbraco.Web.UI.Client/src/common/directives/validation/nodirtycheck.directive.js
```diff
@@ -10,13 +10,12 @@ function noDirtyCheck() {
...
         link: function (scope, elm, attrs, ctrl) {
 
-            var alwaysFalse = {
-                    get: function () { return false; },
-                    set: function () { }
-                };
-            Object.defineProperty(ctrl, '$pristine', alwaysFalse);
-            Object.defineProperty(ctrl, '$dirty', alwaysFalse);
+            // If no attribute value is "false", then skip and use default behaviour.
+            var dirtyCheck = scope.$eval(attrs.noDirtyCheck) === false;
```
kjac, line 14: There must be a native JS function to evaluate a boolean instead of depending on AngularJS to do it for us?

src/Umbraco.Web.UI.Client/src/common/directives/validation/nodirtycheck.directive.js
```diff
@@ -10,13 +10,12 @@ function noDirtyCheck() {
...
+            if (dirtyCheck)
+                return;
 
+            ctrl.$setDirty = Utilities.noop;
```
kjac, line 18: While I'm sure this works, I'm a tad worried it's too good to be true. The quite intricate solution with the `alwaysFalse` binding to `$pristine` and `$dirty` must be there for a reason... this is just a hunch, but I'm afraid your solution over simplifies things. Wouldn't it be prudent to keep the old solution here, just after the conditional dirty check handling?

kjac (dismissed):
@bjarnef I see where you're going with this. Code cleanup is always nice 👍 I have left a bunch of comments for you to have a look at.

src/Umbraco.Web.UI.Client/src/common/directives/validation/nodirtycheck.directive.js
bjarnef: Not sure if there is a specific reason it change `$pristine`. In other alternatives of `no-dirty-check` directives I have seen in only change `$dirty`. Regarding the "Unsaved changes" dialog it only rely on `$dirty` property as far as I know.

src/Umbraco.Web.UI.Client/src/common/directives/validation/nodirtycheck.directive.js
bjarnef: Yes, probably. However I have noticed the `$eval` function is used several other places to parse attribute values.

src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
bjarnef: Not sure if we need to support both `true/false` and `'true'`/`'false'` as string values. Here is just do the comparison in the controller so the view is a bit cleaner :)

kjac: While certainly cleaner it's also more stringent. My point is that the directive (presumably) supports truthly values today, so it should keep on doing that to preserve compatibility with current usages in 3rd party extensions etc.

src/Umbraco.Web.UI.Client/src/common/directives/validation/nodirtycheck.directive.js
kjac: Yep I agree with you. As I tried to explain, it _should_ work fine, but the current solution seems quite explicitly done as-is. There can be two reasons for this: Either the current solution is over complicated and implemented as such due to lack of knowledge/understanding, or else the current solution covers edge cases we can't right now. My point is it can't hurt to retain the current solution from a better-safe-than-sorry standpoint?

src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
bjarnef: We can add more checks here to make it work with other truthly values. E.g.

```
vm.disableDirtyCheck = vm.disableDirtyCheck === true || vm.disableDirtyCheck === 'true' || vm.disableDirtyCheck === 1;
```

bjarnef: I noticed we have this `isBool` function in color picker:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/colorpicker/colorpicker.controller.js#L142-L145

Maybe we should have a global function somewhere to check for truthly values?

kjac: Come to think of it we already do - there's an `Object.toBoolean(value)` method (defined in Extensions.js).

src/Umbraco.Web.UI.Client/src/common/directives/validation/nodirtycheck.directive.js
bjarnef: Yeah, not sure if `$pristine` has any effect on the "Unsaved changes" dialog though. The reason we have the `no-dirty-check` directive is to prevent the "Unsaved changes" dialog to show in certain circumstances.

It seems it only rely on `$dirty` before opening the overlay here:
https://github.com/umbraco/Umbraco-CMS/blob/0140d906a5a66d3ba512332e845a64d36169f911/src/Umbraco.Web.UI.Client/src/common/directives/validation/valformmanager.directive.js#L175-L177

bjarnef:
@kjac it should now check for truthly value 👍

src/Umbraco.Web.UI.Client/src/common/directives/validation/nodirtycheck.directive.js
kjac: With the above changes in mind, shouldn't this be `var dirtyCheck = Object.toBoolean(attrs.noDirtyCheck) === false;` ?

kjac:
Thanks @bjarnef . I left another remark about the usage of `scope.$eval()` for you to consider, now that we discovered `Object.toBoolean()` 😄 

Also would you please revert the dirty handling to the previous solution? If you feel very strongly for changing it, you can do so in a new PR, but please be absolutely certain that you know exactly why the previous solution is written in such a cumbersome way before attempting to change it.

bjarnef:
@kjac I have reverted the dirty handling for now and changes to `Object.toBoolean()` 😊

Regarding the "Unsaved changes" I don't think the `$pristine` has any effect on this, since it only check if the form is  `$dirty`.
https://github.com/umbraco/Umbraco-CMS/blob/0140d906a5a66d3ba512332e845a64d36169f911/src/Umbraco.Web.UI.Client/src/common/directives/validation/valformmanager.directive.js#L175-L177

To me it seems it was just taken from one of the first results on Google, but no comment left, why it need to change `$pristine`.
https://stackoverflow.com/a/27430980

kjac:
@bjarnef finally got around to giving this a test run. It seems your solution didn't take into account the default (undefined) values when the directives were applied without true/false. I have pushed a fix in https://github.com/umbraco/Umbraco-CMS/pull/8638/commits/0a5b442bf0fb7a99268178f942620ddc0cc2e0ca - have a look.

Here's a test dashboard you can use with all combinations of dirty checking:

```html
<div ng-controller="MyDashboardController as vm">
    <div>
        <input type="text" name="text1" placeholder="no-dirty-check" ng-model="text1" no-dirty-check />
    </div>

    <div>
        <input type="text" name="text2" placeholder="no-dirty-check=&quot;false&quot;" ng-model="text2" no-dirty-check="false" />
    </div>

    <div>
        <input type="text" name="text3" placeholder="no-dirty-check=&quot;true&quot;" ng-model="text3" no-dirty-check="true" />
    </div>

    <div>
        <umb-checkbox model="check0" name="check0" disable-dirty-check>disable-dirty-check</umb-checkbox>
    </div>

    <div>
        <umb-checkbox model="check1" name="check1" disable-dirty-check="false">disable-dirty-check="false"</umb-checkbox>
    </div>

    <div>
        <umb-checkbox model="check2" name="check2" disable-dirty-check="true">disable-dirty-check="true"</umb-checkbox>
    </div>

    <div>
        <umb-radiobutton model="radio0" name="radio0" disable-dirty-check>disable-dirty-check</umb-radiobutton>
    </div>

    <div>
        <umb-radiobutton model="radio1" name="radio1" disable-dirty-check="false">disable-dirty-check="false"</umb-radiobutton>
    </div>

    <div>
        <umb-radiobutton model="radio2" name="radio2" disable-dirty-check="true">disable-dirty-check="true"</umb-radiobutton>
    </div>
</div>
```

bjarnef:
@kjac strange `Object.toBoolean()` should already handle `undefined` and `null` values passed in right?
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/lib/umbraco/Extensions.js#L344

Thus returning `false` and the evaluating the condition would be `false`.

kjac:
@bjarnef exactly; but "no value" is supposed to mean `true` in this case, as adding the directive without a value should enable it 😄

bjarnef:
@kjac okay, then we might need a `null` check as well?

kjac:
@bjarnef that was my thought too, and surely it can't hurt to add them, but try as I might I could only manage to pass `undefined` to the directives.

bjarnef:
@kjac anything else we need to handle in this PR?

bjarnef:
@nathanwoulfe maybe you could also take a look at this one as well, which would clean up some duplicated code 👍

bjarnef:
@nielslyngsoe @madsrasmussen maybe this is something you guys want to review as it would cleanup some redundant code.

bjarnef:
@emma-hq can and if we can look at this since it seems to be forgotten and it will make the code much cleaner ⌛ 🥇

mikecp:
Hello @bjarnef ,
Thanks for this PR!

I have checked it and I noticed that the tests that were made on `vm.disableDirtyCheck === undefined` to set the final value of `vm.disableDirtyCheck` to true/false did not take into account the difference between having the attribute with no value, in which case we need an end value "true" (as mentioned earlier in the conversation), and not having the attribute at all, in which case we need an end value of "false" if I am not mistaken.
This resulted into not displaying the "Unsaved changes" dialog when the attribute was not specified.

So, I added a commit to update those tests and differentiate the two situations. If you are OK with them, I'll make the merge.

mikecp:
And here is the dashboard I used for testing:

```
<div ng-controller="MyDashboardController as vm">
    <div>
        <input type="text" name="text0" placeholder="no-dirty-check-attribute" ng-model="text0"/>
    </div>
    <div>
        <input type="text" name="text1" placeholder="no-dirty-check" ng-model="text1" no-dirty-check />
    </div>

    <div>
        <input type="text" name="text2" placeholder="no-dirty-check=&quot;false&quot;" ng-model="text2" no-dirty-check="false" />
    </div>

    <div>
        <input type="text" name="text3" placeholder="no-dirty-check=&quot;true&quot;" ng-model="text3" no-dirty-check="true" />
    </div>


    <div>
        <umb-checkbox model="check99" name="check99">no attribute disable-dirty-check</umb-checkbox>
    </div>

    <div>
        <umb-checkbox model="check0" name="check0" disable-dirty-check>disable-dirty-check</umb-checkbox>
    </div>

    <div>
        <umb-checkbox model="check1" name="check1" disable-dirty-check="false">disable-dirty-check="false"</umb-checkbox>
    </div>

    <div>
        <umb-checkbox model="check2" name="check2" disable-dirty-check="true">disable-dirty-check="true"</umb-checkbox>
    </div>

    <div>
        <umb-radiobutton model="radio99" name="radio99"> no attribute disable-dirty-check</umb-radiobutton>
    </div>
    <div>
        <umb-radiobutton model="radio0" name="radio0" disable-dirty-check>disable-dirty-check</umb-radiobutton>
    </div>

    <div>
        <umb-radiobutton model="radio1" name="radio1" disable-dirty-check="false">disable-dirty-check="false"</umb-radiobutton>
    </div>

    <div>
        <umb-radiobutton model="radio2" name="radio2" disable-dirty-check="true">disable-dirty-check="true"</umb-radiobutton>
    </div>
</div>
```

mikecp:
Reminder to myself: check why the tests are failing

bjarnef:
@mikecp I tested with my original dashboard example and your example which seems to work correct.

However I added a few more check, which seem to trigger the "Unsaved changes dialog" - shouldn't this work the same as just `no-dirty-check` attribute without a value?

```
<div>
        <input type="text" name="text4" placeholder="no-dirty-check=&quot;undefined&quot;" ng-model="text4" no-dirty-check="undefined" />
</div>
<div>
        <input type="text" name="text5" placeholder="no-dirty-check=&quot;null&quot;" ng-model="text5" no-dirty-check="null" />
</div>
```

bjarnef:
Regarding the `no-dirty-check="undefined"` and `no-dirty-check="null"` we get this and thus it works as the default without the attribute. 

image

bjarnef:
Actually I think the noDirtyCheck directive need this instead, where `Object.toBoolean()` returns false for `null` or `undefined` value.
So it need to explicit check the value is `true` instead of `false`, so any "invalid" values works as `no-dirty-check` without a value. 

```
Object.toBoolean(attrs.noDirtyCheck) === true
```

image

bjarnef:
Regarding the console errors the it may has something to do with the values from the attibute is parses as a string. Not sure why I get the non-assigned error though, maybe it has something to do with we now use a newer Angular version than when I originally submitted the PR.
https://docs.angularjs.org/error/$compile/nonassign

Furthermore I wonder if we need the changed in https://github.com/umbraco/Umbraco-CMS/pull/8638/commits/115a475f80ab85733cedf82204244ede6cdb6423

Should it instead explicit check on `false` value instead as is returned if attribute value is `null` or `undefined`.

```
vm.disableDirtyCheck = Object.toBoolean(vm.disableDirtyCheck) === false;
```

bjarnef:
@nathanwoulfe I wonder if you may have an idea why I am starting to see this console error? I wonder if something has changed in Angular v1.8.0?
https://github.com/umbraco/Umbraco-CMS/blob/97ad14c8958489bfa52458dc9c9757185366f15a/src/Umbraco.Web.UI.Client/package.json#L17

It would be an issue if `disableDirtyCheck` wasn't optional, but it is :)

nathanwoulfe:
Not sure, but it could be due to using attrs rather than scope to access the value - if the value is bound in the view, and has changes, perhaps Angular is aware that it has changed, but can't propagate the change to the directive since it's not a scoped input. 

That's a guess, and a big one...

bjarnef:
@nathanwoulfe it doesn't seem to be related to the use of `$attrs` since I get the console errors without the use of this change.

If I comment out the line where `vm.disableDirtyCheck` is assigned in controller I don't get the console errors, but of course the logic doesn't work as expected then.

bjarnef:
@mikecp yeah the `Object.toBoolean()` extension may not fit the needs here, so it is great explicit to check the value.

Regarding `noDirtyCheck` I think we want e.g. this `no-dirty-check="something"` to behave like `no-dirty-check` or `no-dirty-check=""`

So the value need to be "0" (int or string) or "false" (bool or string) to "disable" this (like if this attribute isn't added which make the form field dirty and trigger the "Unsaved changes" dialog.

mikecp:
Hey @bjarnef ,

I start to mixup what was said in the different messages😃 . So if I try to summarize, what we want is:

- no-dirty-check
    * not there => display unsaved changes warning
    * there without value => no display
    * there with value true, 1 or '1' => no display
    * there with value false, 0 or '0' => display unsaved changes warning
    * there with value null, undefined or any other invalid value => no display

Regarding disable-dirty-check, the more I look at it, the more it looks like just a pass-through to the "no-dirty-check" attribute of the form control beneath it. Son I am starting to think that we might just pass its value as-is to the no-dirty-check directive, and let that one interpret it following the rules above...

Since no-dirty-check now handles all possible values sent, is that not the easiest way to keep the logic homogeneous?

Cheers!

bjarnef:
@mikecp correct about the `no-dirty-check` attribute 👍

Yeah, er han probably just pass the value through. Not that important as it from start was supposed to use true/false values and is a property, while the other is a directive. It might could to check what happens with no `dissble-dirty-check`, empty value, null, undefined though 👍

bjarnef:
@mikecp did you had a change to look at this again? would be great to close this and make the code a bit cleaner 😎

mikecp:
Hey @bjarnef ,
Thanks for the reminder 😉I'll have a look back at this so we can finalize it!
Cheers!

bjarnef:
@mikecp did you found some time to look at this again? Let me know if you need further info or I can help with anything 😎👍

bjarnef:
@mikecp anything I can help with to get this closed/merged. I would really like to cleanup some of the redundant code 🙈🤓

mikecp:
Hi @bjarnef ,
My apologies for my very slow reaction, it seems I needed a longer summer break than expected, but I'll definitely look at this and finalize this in the coming days 😅
Cheers!

mikecp (approved):

mikecp:
Hi @bjarnef ,

I finally found the time to make the last checks on this PR 🎉😂

Testing on undefined, null and empty values for `disable-dirty-check` gave indeed some last glitches but I made some last adjustments and now everything is working as it should, meaning that `disable-dirty-check` gives the same behavior as `no-dirty-check` for the same values/bound values.

So I can finally merge this, with all my apologies for this long delay 😅

Thanks again for this PR and for your patience 👍

bjarnef:
Hi @mikecp

Thanks for testing it again and reviewing the code. Great to have this redundant code cleaned up and will be useful in future development easy to enable/disable the dirty check. 👍

PR #8640, opened 2020-08-17
Use button elements in schedule overlay

bjarnef:
### Prerequisites

### Description
This PR replaces a few `<a>` (anchor) elements in schedule overlay with `<button>` elements for accessibility.

At the moment it seems the datepicker doesn't open when focusing the button and hitting enter, which might because of the transcluded content, but we can look at this in another PR.

image


poornimanayar (approved):
Hi @bjarnef ,

Many thanks for your work. I have tested this and I can see that this works and doesnt break any existing functionality. 

Poornima

poornimanayar:
And merged! Thanks once again :-)

PR #8641, opened 2020-08-18
Fixed log viewer level filter by correcting binding for provided array of selected values to filter by.

AndyButland:
Resolves an issue caused by the `[FromQuery]` parameter binding for arrays, which [needs the parameter name explicitly provided](https://stackoverflow.com/a/55516012/489433).

Without this the array comes through as empty so no level filter is applied.


bergmania:
Thanks @AndyButland 💪

PR #8652, opened 2020-08-18
Contrast change for Umbraco button styling for action link cont

RachBreeze:
This PR attempts to improve the contrast when the user tabs to an `umb-button` with the type `link` and button-style `action`.

The current contrast is

image

This screen shot is taken from  \Umbraco.Web.UI.Client\src\views\common\overlays\user\user.html, the edit button is defined as a link with the action style class:
`        <umb-button
            alias="editUser"
            type="link"
            href="#/users/users/user/{{user.id}}"
            action="model.close()"
            button-style="action"
            label="Edit"
            label-key="general_edit"
            ng-if="canEditProfile">
       </umb-button>`

This PR changes the contrast as shown, and removes the underline when the user tabs to it. This is to style it the same as the "Change Password" button next to the Edit Button
image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/buttons.less
+++ b/src/Umbraco.Web.UI.Client/src/less/buttons.less
@@ -204,7 +204,16 @@ input[type="button"] {
 }
 // Made for Umbraco, 2019
 .btn-action {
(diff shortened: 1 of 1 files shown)


RachBreeze:
Further investigations needed following feedback from @poornimanayar , so hold this thought thank you

Issue #3013, opened 2018-09-21
If Umbraco.EmailAddress isn't mandatory it should be posible to be saved as empty


nielslyngsoe:
Reproduction
-----------

I have created a DocType with the Umbraco.EmaillAdress DataType, and the e-mail is not set to required.

### Expected result

I expected the validation to only be invoked when the field had content.

### Actual result

Umbraco wont save the doctype without a valid e-mail, it says: "Property has errors".

renetruelsen:
I was able to reproduce this issue. 
It seems that there are some conflicts between different levels of validations on the Umbraco.Email field vs RegEx. I'll try and dig into it

renetruelsen:
Ok. Here is what I got so far. The problem with the Email PropertyEditor is that it by default undergoes two different checks for correct email syntax. 
1) It uses a client side angular regex: /^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+@[a-z0-9]([a-z0-9-]*[a-z0-9])?(\.[a-z0-9]([a-z0-9-]*[a-z0-9])?)*$/i  (this specific validation rule has been changed a bit forth and back in 2016 and again in september 2017). This client side validation is used several places in the backend, for example when creating a user.

2) And the propertyeditor also uses the build in .NET "EmailAddressAttribute". 

The problem is basically that these two validate differently and is thus handled differently. 
For example "test@test" is by 1) considered valid, but by 2) not considered valid. The result is that the data is stored but when the page is updated, it gives an "Email is invalid"-error. 
But notice! The data is actually stored, and on a node refresh (F5), there is no information that the email is invalid.

...for now I've decided to keep the custom regex validation out of the discussion, but this is also serverside validation, and thus acts similar to 2).

My question is - what is the proper solution for this? 
My suggestion is either to align the email-syntax validation between the angular front-end and "EmailAttribute" back-end, and at the same time change the default "Validate as an email address"-expression to be the same as the one used in the angular check.
Or - maybe completely remove the backend syntax check on email addresses, since it's redundant?

What do you think?

renetruelsen:
...submitted a PR for this issue

stevetemple:
If the PR is being rejected on the grounds of it being for v7.x and that not having any new bug fixes/features. Is this still happening in v8 or can we close this as won't be done.

nul800sebastiaan:
[...]

PR #8655, opened 2020-08-18
Deprecate IsRequired on Email datatype

stevetemple:
### Prerequisites

For issue https://github.com/umbraco/Umbraco-CMS/issues/3013

### Description

On the email datatype there is a field IsRequired, which when true makes the field mandatory overriding the mandatory selection when the datatype is added to a doctype making it mandatory without showing the * etc for a mandatory field. 

This then has confusing and unexpected validation. In discussion with @nul800sebastiaan we thought it would be best to hide the IsRequired field for now. The best I could do was make it hidden and including a deprecated message.

To reproduce the issue, add an email datatype with Is Required set to true to a doctype with mandatory set to false. You'll see it doesn't have the UI of a mandatory field but does have validation for a mandatory field.

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/EmailAddressConfiguration.cs
+++ b/src/Umbraco.Web/PropertyEditors/EmailAddressConfiguration.cs
@@ -1,4 +1,5 @@
-﻿using Umbraco.Core.PropertyEditors;
+﻿using System;
+using Umbraco.Core.PropertyEditors;
 
 namespace Umbraco.Web.PropertyEditors
 {
@@ -8,6 +9,7 @@ namespace Umbraco.Web.PropertyEditors
     public class EmailAddressConfiguration
     {
         [ConfigurationField("IsRequired", "Required?", "hidden", Description = "Deprecated; Make this required by selecting mandatory when adding to the document type")]
+        [Obsolete("No longer used, use `Mandatory` for the property instead. Will be removed in the next major version")]
         public bool IsRequired { get; set; }
     }
 }
--- a/src/Umbraco.Web/PropertyEditors/EmailAddressConfiguration.cs
+++ b/src/Umbraco.Web/PropertyEditors/EmailAddressConfiguration.cs
@@ -7,7 +7,7 @@ namespace Umbraco.Web.PropertyEditors
     /// </summary>
     public class EmailAddressConfiguration
     {
-        [ConfigurationField("IsRequired", "Required?", "boolean")]
+        [ConfigurationField("IsRequired", "Required?", "hidden", Description = "Deprecated; Make this required by selecting mandatory when adding to the document type")]
         public bool IsRequired { get; set; }
     }
 }


poornimanayar:
Hi @stevetemple ,

Thank you for the work. To begin with I couldnt reproduce or get the red error ribbon, it was giving me text which says "Required" for a fraction of a second as seen in the gif. I have tested this on a new data type and an existing data type (with required on). For new data types it works fine. For existing data types where email address was marked required it still gives me that momentary text. Any way we can hide this in the view? Is it worth hiding the label in the edit view of the datatype? What do you think

email

Poornima

stevetemple:
It's a good point, I've been thinking about this. We discussed not removing this as it's a breaking change but defaulting it to false and hiding it. 

Maybe we should just add the deprecated message for v8 and I'll create a v9 PR where we remove the field altogether?

poornimanayar:
Yeah , it is a bit confusing. Just looking at the view, there is an `ng-required` which checks for the `model.config`, will getting rid of that be a breaking change? I am assuming no.

src/Umbraco.Web/PropertyEditors/EmailAddressConfiguration.cs
```diff
@@ -7,7 +7,7 @@ namespace Umbraco.Web.PropertyEditors
     /// </summary>
     public class EmailAddressConfiguration
     {
-        [ConfigurationField("IsRequired", "Required?", "boolean")]
+        [ConfigurationField("IsRequired", "Required?", "hidden", Description = "Deprecated; Make this required by selecting mandatory when adding to the document type")]
```
nul800sebastiaan, line 10:
```suggestion
        [ConfigurationField("IsRequired", "Required?", "hidden", Description = "Deprecated; Make this required by selecting mandatory when adding to the document type")]
        [Obsolete("No longer used, use `Mandatory` for the property instead. Will be removed in the next major version")]
```

nul800sebastiaan: Will need to add `using System;` at the top too.

stevetemple:
> Yeah , it is a bit confusing. Just looking at the view, there is an `ng-required` which checks for the `model.config`, will getting rid of that be a breaking change? I am assuming no.

I think if we stop the field being required even though it was on the version of umbraco before the upgrade it'll be subtlety breaking in that a field that you expected to be required was no longer required and now you might get empty strings in it.

nul800sebastiaan (commented):
> Maybe we should just add the deprecated message for v8 and I'll create a v9 PR where we remove the field altogether?

Yep that's what I was alluding to in the issue. 
I just realized we can't quite remove the "required" label, but I think this is good enough for now so people don't actually use it.

src/Umbraco.Web/PropertyEditors/EmailAddressConfiguration.cs
poornimanayar: Won't marking as `Obsolete` hide it if the `showDeprecatedPropertyEditors` in `umbracosettings.config` is set to true?

nul800sebastiaan: This is only obsoleting the `IsRequired` property -  https://github.com/umbraco/Umbraco-CMS/pull/8655/files#diff-54cf08e9c3a954b718f21f33ca16968aR11

poornimanayar (approved):
Hey @stevetemple ,

I think this is all good at the moment. Approving and merging your work! Thanks once again for your PR 👍 

Poornima

poornimanayar:
And merged :-)

PR #8659, opened 2020-08-19
Use convertVirtualToAbsolutePath for all file URLs of a block config

nielslyngsoe:
Fixing the issue of missing the use of ´convertVirtualToAbsolutePath´. Plus refactored to the init of a BlockObject, so Devs don't need to worry about this.


warrenbuckley:
All good & merged in.
Tested on a pair programming call with @nielslyngsoe

Issue #8651, opened 2020-08-18
Refactor configuration to take advantage of Options Pattern


JimBobSquarePants:
Current .NET Core configuration utilizes a non-conventional pattern to load configuration values.

https://github.com/umbraco/Umbraco-CMS/blob/43efee9647ed8f2eaab3ec2b71d3dd3eb46b9017/src/Umbraco.Configuration/AspNetCoreConfigsFactory.cs

ASP.NET Core already has an established pattern to load options. 
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.1

There are several advantages to using this pattern.

1. It's common convention. No surprises for future contributors.
2. It allows options monitoring and dynamic reloading when required. 
3. It negates the requirement for each superfluous interface definition and implementation. 

Binding a section becomes trivial. 

``` c#
services.Configure<MyOptions>(Configuration.GetSection(nameof(MyOptions)));
``` 

Consuming it equally so.

``` c#
public class MyOptionsConsumer(IOptions<MyOptions> options)
{
    Guard.NotNull(options, nameof(options));

    this.options = options.Value;    
}
```

Unit testing again is trivial.

``` c#
var myOptions = new MyOptions();
var myOptionsConsumer = new MyOptionsConsumer(Options.Create(myOptions));
```

bergmania:
Hi @JimBobSquarePants..

Overall seems like a good idea. The only thing I don't like is if we have to know the `IOptions<T>` all over our own codebase.  
I'm sure there is a way to get the best of both worlds.  
E.g https://www.strathweb.com/2016/09/strongly-typed-configuration-in-asp-net-core-without-ioptionst/

JimBobSquarePants:
I think that post is a little naïve and doesn't take into account any of the the highlighted benefits of `IOptions<T>`. 

Having `IOptions<T>` passed to your constructors isn't a bad thing. As I noted before.

>It's common convention. No surprises for future contributors.

bergmania:
We decide to allow the MS Config interfaces in `Umbraco.Core`. @AndyButland already started a draft PR. Good start, and we can continue on work.

I think we should make a `netcore/config` branch as the base branch for all these smaller config PRs, so we can merge these into `netcore/netcore` as one big PR, but review it as several smaller PRs.

umbrabot:
Hi @JimBobSquarePants,

[...]

PR #8660, opened 2020-08-19
Migrating to IOptions pattern for configuration.

AndyButland:
Just a tiny POC for now - referencing issue #8651.  Demonstrates that the `IOptions` approach will support:

- Default values if no configuration provided
- Overrides via configuration added in `appSettings.json`
- Hot-reloading works if `IOptionsSnapshot` used

It we proceed, need to handle for all other settings objects and apply to tests.


AndyButland:
I've taken this to a point now where I think it can be reviewed and perhaps merged (post-alpha).  It's been a bit tricky to do given we're sort of in a part-way transition point, with a .NET executable back-office, but with a fair bit of code in `Umbraco.Web` and `Umbraco.Tests` that should still of course compile (and for the tests, pass).  But I think this is a reasonable way to proceed - the idea being that we port over the .NET executable call stack to `IOptions` based configuration, but for now still have to maintain some additional parameters and/or conversion functions to ensure the parts in other areas of the solution still work.  In time, as they are migrated into an appropriate web or tests project, they'll need amending too.

The summary of what I've done is:

- For each existing configuration model class (e.g. `GlobalSettings : IGlobalSettings`), I've created a new class in `Umbraco.Core.Configuration.Models`.  The new versions are pure POCOs; they have no interface and no logic to populate the fields from configuration.
   - A few - `HealthCheckSettings`, `ContentSettings` and `RequestHandlerSettings` - have one or more fields not yet implemented (usually as it's more complex configuration than simple fields, and probably we need to consider what this will look like in the new JSON based configuration).
- In `UmbracoCoreServiceCollectionExtensions.AddUmbracoConfiguration()` I've registered all of these with the service collection.
- I've then found all the instances of the previously injected model classes and changed them to inject e.g. `IOptions<GlobalSettings>`
    - I've done this with `IOptions` over `IOptionsSnapshot` partly as the latter didn't work in a lot of cases (as not in a scoped context), and partly as only the former seems to have the `Options.Create` method available, which is useful in unit test code and in avoiding duplicate constructors (one for where we inject the parameter, and one for where we new up the class in code).
- I've done this within all projects within the .NET Core executable call-stack (i.e. all projects except `Umbraco.Web`).
   - My thinking is that these should be converted as and when the files from this project are adapted and migrated into the appropriate `Umbraco.Web.*` project.
- In order to maintain everything compiling and tests passing, including in `Umbraco.Web` and `Umbraco.Tests` that broadly continue to use the existing configuration model classes and methods, I've created a couple of temporary classes, responsible for converting from one configuration model into the other (and vice versa).  See `ConfigModelConversionsFromLegacy` and `ConfigModelConversionsToLegacy`.

In terms of testing, I've got all projects building and tests passing.  I've also checked the web application still boots and back-office works  appears to function as it did before.

src/Umbraco.Core/BackOffice/BackOfficeIdentityUser.cs
```diff
@@ -39,7 +40,7 @@ public class BackOfficeIdentityUser : IdentityUser<int, IIdentityUserLogin, Iden
         /// <param name="email">This is allowed to be null (but would need to be filled in if trying to persist this instance)</param>
         /// <param name="culture"></param>
         /// <returns></returns>
-        public static BackOfficeIdentityUser CreateNew(IGlobalSettings globalSettings, string username, string email, string culture)
+        public static BackOfficeIdentityUser CreateNew(GlobalSettings globalSettings, string username, string email, string culture)
```
bergmania, line 43: I guess this should be  `IOptions<GlobalSettings>`? . Same in other locations?

AndyButland: I think this is OK as is.  We only really need `IOptions<GlobalSettings>` if we're injecting them in via the constructor.  For this static factory method, we can just pass in the POCO directly from the calling code (where it's probably been unpacked from `IOptions<GlobalSettings>`).

Shazwazza: But... if we want to support hot reloading of config then we need to pass in IOptionsSnapshot (if in request based/transient lifetime) or use IOptionsMonitor, if we pass in the real instance, that instance will never change

AndyButland: That's true, though I think this parameter is still OK typed as just `GlobalSettings`.  This `CreateNew` method is called (for example) from `UsersController`, that takes an `IOptions<GlobalSettings>` parameter (and so it's there that, if we wanted to, we could instead use `IOptionsSnapshot<GlobalSettings>`.  This can be unpacked though for calling this `CreateNew` method and so we can just pass the POCO (it doesn't need to be passed with it's `IOptions`, or `IOptionsSnapshot`, wrapper).

See my comment above too about why I've used `IOptions`, at least for now, over the other two possibilities (starts: _"I've done this with IOptions over IOptionsSnapshot partly..."_).

src/Umbraco.Tests/TestHelpers/TestObjects.cs
```diff
@@ -180,11 +184,11 @@ public void RegisterServices(IRegister register)
...
+                        new EntityXmlSerializer(contentService.Value, mediaService.Value, dataTypeService.Value, userService.Value, localizationService.Value, contentTypeService.Value, urlSegmentProviders, TestHelper.ShortStringHelper, propertyEditorCollection), logger, umbracoVersion, Options.Create(globalSettings), "installedPackages.config"),
                     new PackageInstallation(
-                        new PackageDataInstallation(logger, fileService.Value, macroService.Value, localizationService.Value, dataTypeService.Value, entityService.Value, contentTypeService.Value, contentService.Value, propertyEditorCollection, scopeProvider, shortStringHelper, GetGlobalSettings(), localizedTextService.Value),
+                        new PackageDataInstallation(logger, fileService.Value, macroService.Value, localizationService.Value, dataTypeService.Value, entityService.Value, contentTypeService.Value, contentService.Value, propertyEditorCollection, scopeProvider, shortStringHelper, Microsoft.Extensions.Options.Options.Create(globalSettings), localizedTextService.Value),
```
elit0451, line 191:
```suggestion
                        new PackageDataInstallation(logger, fileService.Value, macroService.Value, localizationService.Value, dataTypeService.Value, entityService.Value, contentTypeService.Value, contentService.Value, propertyEditorCollection, scopeProvider, shortStringHelper, Options.Create(globalSettings), localizedTextService.Value),
```

elit0451 (changes requested):
Really amazing work @AndyButland 🎉💪 
I have left some comments around the code, but it would have definitely been much easier to review those changes in separate small PRs - at least to have the tests changes into a different PR. _Just a suggestion for next time 🙂_

src/Umbraco.Core/Configuration/Models/GlobalSettings.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public string DefaultUILanguage { get; set; } = "en-US";
+
+        public bool HideTopLevelNodeFromPath { get; set; } = false;
+
+        public bool UseHttps { get; set; } = false;
+
+        public int VersionCheckPeriod { get; set; } = 7;
+
+        public string Path { get; set; } = "~/umbraco";
```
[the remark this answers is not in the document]
AndyButland: Agreed, have updated.

src/Umbraco.Core/Configuration/Models/ImagingCacheSettings.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Umbraco.Core.Configuration.Models
+{
+    public class ImagingCacheSettings
+    {
+        public int MaxBrowserCacheDays { get; set; } = 7;
+
+        public int MaxCacheDays { get; set; } = 365;
+
+        public uint CachedNameLength { get; set; } = 7;
```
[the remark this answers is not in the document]
AndyButland: Great spot!  Have amended.

src/Umbraco.Core/Routing/UrlProvider.cs
```diff
@@ -24,7 +26,7 @@ public class UrlProvider : IPublishedUrlProvider
         /// <param name="mediaUrlProviders">The list of media url providers.</param>
         /// <param name="variationContextAccessor">The current variation accessor.</param>
         /// <param name="propertyEditorCollection"></param>
-        public UrlProvider(IUmbracoContextAccessor umbracoContextAccessor, IWebRoutingSettings routingSettings, UrlProviderCollection urlProviders, MediaUrlProviderCollection mediaUrlProviders, IVariationContextAccessor variationContextAccessor)
+        public UrlProvider(IUmbracoContextAccessor umbracoContextAccessor, IOptionsSnapshot<WebRoutingSettings> routingSettings, UrlProviderCollection urlProviders, MediaUrlProviderCollection mediaUrlProviders, IVariationContextAccessor variationContextAccessor)
         {
             if (routingSettings == null) throw new ArgumentNullException(nameof(routingSettings));
```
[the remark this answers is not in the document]
AndyButland: From what I read [here](https://tpodolak.com/blog/2018/01/29/asp-net-core-make-sure-ioptionst-initialized/), and in testing it out, I'm not sure we actually need a guard clause here at all any more.  It seems even if we've omitted to configure this particular class (in `AddUmbracoConfiguration`), due to the fact we're injecting in via `IOptions`, it'll never be null (if unconfigured, it would be non-null, with default values).  So have just removed this check.

src/Umbraco.Infrastructure/Persistence/UmbracoDatabaseFactory.cs
```diff
@@ -70,7 +71,7 @@ private DbProviderFactory DbProviderFactory
         /// Initializes a new instance of the <see cref="UmbracoDatabaseFactory"/>.
         /// </summary>
         /// <remarks>Used by core runtime.</remarks>
-        public UmbracoDatabaseFactory(ILogger logger, IGlobalSettings globalSettings, IConnectionStrings connectionStrings, Lazy<IMapperCollection> mappers,IDbProviderFactoryCreator dbProviderFactoryCreator)
+        public UmbracoDatabaseFactory(ILogger logger, GlobalSettings globalSettings, ConnectionStrings connectionStrings, Lazy<IMapperCollection> mappers,IDbProviderFactoryCreator dbProviderFactoryCreator)
```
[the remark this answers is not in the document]
AndyButland: Yes, I think you are right.  Have updated.

src/Umbraco.Tests.Common/Builders/GlobalSettingsBuilder.cs
```diff
@@ -175,12 +161,10 @@ public override IGlobalSettings Build()
...
             var installMissingDatabase = _installMissingDatabase ?? false;
-            var isSmtpServerConfigured = _isSmtpServerConfigured ?? false;
-            var path = _path ?? "/umbraco";
             var registerType = _registerType ?? null;
-            var reservedPaths = _reservedPaths ?? "~/app_plugins/,~/install/,~/mini-profiler-resources/,";
-            var reservedUrls = _reservedUrls ?? "~/config/splashes/noNodes.aspx,~/.well-known,";
-            var umbracoPath = _umbracoPath ?? "~/umbraco";
+            var reservedPaths = _reservedPaths ?? GlobalSettings.StaticReservedPaths;
+            var reservedUrls = _reservedUrls ?? GlobalSettings.StaticReservedUrls;
+            var path = _path ?? "~/umbraco";
```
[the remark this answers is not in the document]
AndyButland: Have updated this too.

src/Umbraco.Tests/TestHelpers/TestObjects.cs
AndyButland: Yes, we don't need the full namespace here - have removed..

AndyButland:
Thanks for looking over @elit0451 - I've addressed your comments.  And point noted on the PR size, when I started I hadn't planned on it being quite so big... unfortunately though this seemed a tricky one to break up, as there was quite a lot of change needed, for example in the tests, just to bring things back to a compiling/test passing state,

bergmania:
Hi @AndyButland 

I see this error when i click the package section. Not sure how this is related to this PR, but I don't see in on `netcore/netcore`

image

AndyButland:
Should be OK now @bergmania - these are tricky to resolve, as the stack trace isn't very useful.  The issue is some dependency up the chain from `PackagingService` itself has a dependency that can't be resolved, but the error details don't tell you which one.  Have found and fixed this one though.

bergmania:
@AndyButland 

Just merged netcore/netcore into this and also made a simple commit cleaning a bit up
https://github.com/umbraco/Umbraco-CMS/pull/8660/commits/90d14158ee14d1bff3dfac4c831d8a47293040cf

src/Umbraco.Core/Routing/UrlProvider.cs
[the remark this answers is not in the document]
elit0451: Makes sense! 🙂

elit0451 (approved):

bergmania:
@AndyButland,
Overall changes look good. But some of the configurations need to be "updateable". E.g. `HostingSettings.Debug` is used in the very beginning and "Saved" in `AspNetCoreHostingEnvironment`. Before this values was received from config every time.

I think it should be possible to override most configs in code. Currently in startup but later also in composers:
E.g.
```
        {
            services.AddUmbraco(_env, _config);

            services.Configure<HostingSettings>(settings => // Or PostConfigure
            {
                settings.Debug = false;
            });
        }
```

If this happens we need to use this value. I know it is not possible for all configurations, but the ones used in composition root e.g. from `IHostingEnvironment`, `IIOHelper`, `IBackOfficeInfo` etc. needs to be updateable.  Maybe we need to use `IOptionsMonitor` in these?

AndyButland:
I've pushed some amends to use `IOptionsMonitor` for those classes you've mentioned, constructed in `AddUmbracoCore()`.  All seems to work as it did before.

bergmania:
Thanks @AndyButland, I will test and look into the build server issue

src/Umbraco.Infrastructure/Models/ContentEditing/UserInvite.cs
```diff
@@ -33,7 +33,8 @@ public IEnumerable<ValidationResult> Validate(ValidationContext validationContex
             if (UserGroups.Any() == false)
                 yield return new ValidationResult("A user must be assigned to at least one group", new[] { nameof(UserGroups) });
 
-            if (Current.Configs.Security().UsernameIsEmail == false && Username.IsNullOrWhiteSpace())
+            // TODO: this will need another way of retrieving this setting if and when Configs are removed from Current.
```
AndyButland, line 36: You can remove this `TODO` now @bergmania (now you've removed `Current.Configs`.

src/Umbraco.Infrastructure/Trees/TreeNode.cs
```diff
@@ -113,7 +114,11 @@ public string IconFilePath
                     return Current.IOHelper.ResolveUrl("~" + Icon.TrimStart('~'));
 
                 //legacy icon path
-                var backOfficePath = Current.Configs.Global().GetUmbracoMvcArea(Current.HostingEnvironment);
+
+                // TODO: replace this when we have something other than Current.Configs available
```
AndyButland, line 118: This `TODO` is also no longer required.

src/Umbraco.Infrastructure/Models/ContentEditing/UserInvite.cs
bergmania: These TODOs are still required, as I just replaced `Current.Configs` with their specific config class on `Current`

src/Umbraco.Infrastructure/Trees/TreeNode.cs
bergmania: These TODOs are still required, as I just replaced `Current.Configs` with their specific config class on `Current`

src/Umbraco.Infrastructure/Models/ContentEditing/UserInvite.cs
AndyButland: Not sure.  Just to be clear, I added these, they weren't there before.  And I added them to note that we'd need to changes this when `Current.Configs` is no longer available.  And now it's not.  So it could be you want to add a new TODO about refactoring from use of `Current` per so (i.e. no longer using service location if we can avoid it), but that wasn't there before and wasn't what I intended with the comment..

bergmania:
Commits part of this PR https://github.com/umbraco/Umbraco-CMS/pull/8891

PR #8661, opened 2020-08-19
Adjust block list configuration

bjarnef:
### Prerequisites

### Description
I have added a few changes to the configuration of the block list

- Ensure edit overlay can be closed via `esc` shortcut
- Remove the white color in the "Add" buttons in the configuration overlay so the text is visible and more consistent with `umb-node-preview`.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/components/umb-node-preview.less

Furthermore in this overlay these "Add" buttons has border radius which the default `umb-node-preview` doesn't have. Make we should start using a new class name which can be used for this picker button and re-used throughout the UI since it is re-used in many pickers, sometimes with copied styling to another class, so when the styling on the original `umb-node-preview` class change a bit, the others buttons starts being inconsistent.

I also noticed when using tabs key to to focus "delete" action it trigger both "delete" overlay and "edit" overlay, where it seems to bubble to parent `ng-click` although the "delete" button event should stop event bubbling (it should on mouse click, but not via enter key).

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
@@ -30,7 +30,6 @@
             }
 
             loadElementTypes();
-
         }
 
         function loadElementTypes() {
@@ -47,9 +46,10 @@
                 }
             }
         }
+
         unsubscribe.push(eventsService.on("editors.documentType.saved", updateUsedElementTypes));
 
-        vm.requestRemoveBlockByIndex = function (index) {
+        vm.requestRemoveBlockByIndex = function ($event, index) {
             localizationService.localizeMany(["general_delete", "blockEditor_confirmDeleteBlockMessage", "blockEditor_confirmDeleteBlockNotice"]).then(function (data) {
                 var contentElementType = vm.getElementTypeByKey($scope.model.value[index].contentTypeKey);
                 overlayService.confirmDelete({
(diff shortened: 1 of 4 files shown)


bjarnef:
@nielslyngsoe do you know why we have both html, js and less inside https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist and not separating as with other part of the existing code?

**Directives/components**
https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives

**Less**
https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/less

**Component views**
https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components

nielslyngsoe (approved):
Looking good

nielslyngsoe:
@bjarnef So the reasoning for having both HTML, JS, and LESS inside is to gather the files of one concern to provide a better overview. We have a few other things that does it this way. I'm fully aware that we break the original file structure, but the intention is for the better, otherwise, you can only find related files through references.

bjarnef:
@nielslyngsoe maybe regarding the property editor and prevalue editor, but for the component itself it seems a bit inconsistent where more or less all other components have the HTML, JS, LESS part structured in separate folders.

bjarnef:
Replaced with PR https://github.com/umbraco/Umbraco-CMS/pull/8686

PR #8664, opened 2020-08-19
Add filter css class for not allowed thumbnail extensions

bjarnef:
### Prerequisites

### Description
This PR add the `not-allowed` cursor style when however not allowed files in block thumbnail picker similar to file picker in packager: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/packages/edit.controller.js#L286


nielslyngsoe:
@bjarnef could you try the latest version, this should have been fixed now.

bjarnef:
@nielslyngsoe it doesn't seem the changes are merged to `v8/contrib` branch yet.

nielslyngsoe:
@bjarnef Ah, yes, you are right. They are only in the 8.7 branch.

bjarnef:
@nielslyngsoe okay, so I guess if we want to make changes to the upcoming block list editor, we need to work on 8.7 branch for now or target changes to 8.7 branch.

bjarnef:
@nielslyngsoe I have tested with v8.7-RC .. the icon color seems to work, but when selecting an SVG as thumbnail, the inline css become `background-image: url(~/Media/0mil3yiu/android.svg?upscale=false&width=400);` so it is looking for a file `/umbraco/~/Media/0mil3yiu/android.svg`.

It should be `~/umbraco/Media/0mil3yiu/android.svg` instead and without `?upscale=false&width=400` for SVG files.

bjarnef:
Replaced with PR https://github.com/umbraco/Umbraco-CMS/pull/8684

PR #8667, opened 2020-08-19
Update description for block list live editing

bjarnef:
### Prerequisites

### Description
The block list live editing property in prevalues says it is for live updated custom views. However it also works without custom views, when using dynamic expression in label, e.g. `{{myProperty}}`.

When live editing is enabled the label text updates as you type. When it is disabled it only update the label text after submitting the overlay.

chrome_2020-08-19_21-31-13


bjarnef:
Replaced with PR https://github.com/umbraco/Umbraco-CMS/pull/8687

PR #8671, opened 2020-08-20
Add more tags on document types to show whether the property has custom mandatory message and validation

poornimanayar:
Hi,

With this PR I have tried to add more tags to the property view when there is a custom validation message or pattern on the field. I think it will be good if we can get an overview of fields with custom mandatory messages and custom validation pattern.

tags

Poornima


nul800sebastiaan:
Thanks @poornimanayar - We see where you're going with this one but it's actually making it less clear to see what's going on. @nielslyngsoe has ideas on how to improve this in other ways, but it's a tough one to crack. One thing is for sure, we don't want to add more "noise" to this view at the moment! Thanks though! 👍

PR #8674, opened 2020-08-20
Netcore: Update user on successful login

bergmania:
## Notes
Move the code that updates the user on successful login into the netcore `BackofficeSignInManager`

From here https://github.com/umbraco/Umbraco-CMS/blob/netcore/netcore/src/Umbraco.Web/Security/BackOfficeSignInManager.cs#L204-L209

Even that it works on this location, please review it is the correct spot to do this.

## Test
- Login the user, and go to user section and ensure the user last login is updated (and the user is not inactive anymore)


Shazwazza:
Seems fine, i think there's still more to migrate with all of that so might figure out other things along the way but seems good to me for now.

PR #8678, opened 2020-08-20
Fixes issue with scopes being disposed or referenced incorrectly

Shazwazza:
See TODO notes in code for some explanation. I will need to fix this up and have a look at v8 next week.

Also fixed a critical issue with DatabaseServerMessenger.LocalIdentity being set incorrectly and not as a property. This means it was recalculated every time with a new GUID so it was changing every time. This meant that this local identity was processing it's own DB instructions which meant that it was processing it's own events twice - essentially double publishing, double indexing, etc... every time. (careful using => assignments! ;)

### DIFF
--- a/src/Umbraco.Infrastructure/Sync/DatabaseServerMessenger.cs
+++ b/src/Umbraco.Infrastructure/Sync/DatabaseServerMessenger.cs
@@ -63,6 +63,13 @@ namespace Umbraco.Core.Sync
             _lastPruned = _lastSync = DateTime.UtcNow;
             _syncIdle = new ManualResetEvent(true);
             _distCacheFilePath = new Lazy<string>(() => GetDistCacheFilePath(hostingEnvironment));
+
+            // See notes on LocalIdentity
+            LocalIdentity = NetworkHelper.MachineName // eg DOMAIN\SERVER
+                + "/" + _hostingEnvironment.ApplicationId // eg /LM/S3SVC/11/ROOT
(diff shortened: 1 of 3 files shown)


src/Umbraco.Infrastructure/Sync/DatabaseServerMessenger.cs
```diff
@@ -526,11 +533,7 @@ private void SaveLastSynced(int id)
         /// <para>Practically, all we really need is the guid, the other infos are here for information
         /// and debugging purposes.</para>
         /// </remarks>
-        protected string LocalIdentity => NetworkHelper.MachineName // eg DOMAIN\SERVER
-            + "/" + _hostingEnvironment.ApplicationId // eg /LM/S3SVC/11/ROOT
-            + " [P" + Process.GetCurrentProcess().Id // eg 1234
-            + "/D" + AppDomain.CurrentDomain.Id // eg 22
-            + "] " + Guid.NewGuid().ToString("N").ToUpper(); // make it truly unique
+        protected string LocalIdentity { get; }
```
benjaminc, line 536: I think you can do this inline initialization if you wanted to, with `{ get; } = NetworkHelper...`.  That would do what I think you were intending from the inline method call.  It would do the same as moving it to the constructor, but would keep the calculation near the property for reference.

Shazwazza:
see https://github.com/umbraco/Umbraco-CMS/pull/8772

PR #8684, opened 2020-08-20
V8.7RC Add filter css class for not allowed thumbnail extensions

bjarnef:
### Prerequisites

### Description
This PR add the `not-allowed` cursor style when however not allowed files in block thumbnail picker similar to file picker in packager: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/packages/edit.controller.js#L286


bjarnef:
@nielslyngsoe it would be great with this addition, so it is clear when entities not ser allowed to select.

nul800sebastiaan:
Sweet, thanks @bjarnef! And thanks also for the missing semicolons!! 👍

PR #8687, opened 2020-08-20
V8.7RC Update block list live editing description

bjarnef:
### Prerequisites

### Description
The block list live editing property in prevalues says it is for live updated custom views. However it also works without custom views, when using dynamic expression in label, e.g. `{{myPropertyAlias}}`.

When live editing is enabled the label text updates as you type. When it is disabled it only update the label text after submitting the overlay.

chrome_2020-08-19_21-31-13


nul800sebastiaan:
Looks good to me, cheers @bjarnef! 👍

PR #8690, opened 2020-08-20
V8.7RC Block list sort within containment

bjarnef:
### Prerequisites

### Description
This PR makes the sorting of items in block list and nested content works within parent container instead of dragging items way out of the block list or nested content area.


nul800sebastiaan:
Yay! 👍 Thanks again @bjarnef!

Issue #8679, opened 2020-08-20
Feature Request : Use the reference tracking to provide a better editor experience


dawoe:
With the release of 8.6 media tracking was introduced. But the backoffice UI could do with some improvements to provide a better experience.

Since I have a popular package (Nexu) that does the same so tried to see how it lines up.

Having tested it noticed that also content usage is being tracked, not only media. 

On a media item you can see where the item is being used

![image](https://user-images.githubusercontent.com/1193822/90784926-12055300-e302-11ea-969e-40e2772b2081.png)

But on content this information is missing. And it is useful there as well.

Another great feature I have in my package is that when editor tries to delete or unpublish something he or she get's warned that the item is in use, with a list of where it is being used.

This is completly missing in backoffice at the moment. So a editor can still delete stuff and potentialy breaking his website.

How can you help?
-------------------------------

I think these problems can be tackled in a serie of small pull requests.

1. Display the references on a content item
2. Warn a editor during the deletion of a media item.
3. Warn a editor during the deletion of a content item
4. Warn a editor when he tries to unpublish.
5. Implement step 2-4 for bulk actions in list views

I don't think the references need to be shown in the dialogs for unpublish/delete because the list is available on the info tab.

When we have this in place we can think about future features like when a user wants to delete a image, let him choose a new image that should be used instead.

But I would like to hear if HQ is open for this before doing any effort on this.

TODO List
-------------------------------

- [x] https://github.com/umbraco/Umbraco-CMS/issues/8692
- [x] https://github.com/umbraco/Umbraco-CMS/issues/8699
- [x] https://github.com/umbraco/Umbraco-CMS/issues/8693
- [x] https://github.com/umbraco/Umbraco-CMS/issues/8694
- [x] https://github.com/umbraco/Umbraco-CMS/issues/8695
- [ ] https://github.com/umbraco/Umbraco-CMS/issues/8696
- [ ] https://github.com/umbraco/Umbraco-CMS/issues/8697
- [ ] https://github.com/umbraco/Umbraco-CMS/issues/8698

**Update** 
Until we are feature complete al the PR's will be targeted at v8/item-tracking branch. Once everything is done this can go into v8/contrib

Dave

nielslyngsoe:
Hi @dawoe

If we already have trustworthy data about the references I don't see any reason why we don't show this. And use it to warn users about the references when unpublishing or deleting content.

I will go ahead and say that PRs for such improvements are very VERY welcome!

[...]

PR #8703, opened 2020-08-21
Show tracked references on a content item

### LINKED ISSUE #8692: Display tracked references on content items
This is part of #8679

We need to display the tracked references of a content item on the info tab, so a editor can see where this is used.
Just like it currently is being done on media items.

Testing should be done with and without variants enabled.

Personally I think this should be shown between the links box and the audit trail. I think this something editors need more than the audit trail.

dawoe:
### Prerequisites

If there's an existing issue for this PR then this fixes #8692 

### Description

This is the first PR in a series of PR's related to this bigger task #8679 

Starting out I noticed that all the code related to showing references on a media item was strongly coupled to the media section.

To decouple this I first created a dedicated api controller and angular resource for getting references.
After that I created a reusable angular component to show the references.

I refactored the info tab for media to decouple showing of the references by using this component. After that I also used the component on the info tab of content items.

Now references to the content item are visible on the info tab.

image

### Testing

- Check if media item references are shown correctly on the info tab
- Check if content item references are shown correctly on the info tab

Both can be verified by checking the relations "Related Document" and "Related Media" 

Probably best to test with multi variant site as well


poornimanayar:
Fantastic work @dawoe ! I will give this a spin and let you know how it goes :-) Watch this space!

poornimanayar:
Hi @dawoe ,

Many thanks for your PR. I started testing your PR. I started testing the media first and here are my findings. In a single language environment all the below works and shows the media refs.

1. Media Picker - Works , shows references correctly on media
2. Multiple media picker - works and shows media references if media items are specified, not folders
3. MNTP Media - works well
4) All the above in NC - works

But in a variant set up its not getting picked up. I tried with the same image I used above and expected it to show 2 refs (US & UK, am I correct in my understanding here) but I had only one. I tried adding new images as well, it shows me a reference but to the wrong language. I added a new image to the UK language page and expected it to show me refs to the UK page but I have US language page as reference. 

Poornima

dawoe:
I don't know if the core version tracks in which variant it is used. 

This is how it worked before my PR by the way. I guess because it is because even if a item has language variants, it's still one umbraco item.

Maybe @Shazwazza  can give some insights.

Dave

dawoe:
I just checked the relation table, and there is no reference to the variant stored with the relation.

Shazwazza:
Hi @dawoe ! Yep we have some backlog tasks to get variant support for tracking. The proposed way to do this was to create another DB table which extended the relations table. We don't want to break the current abilities and APIs of how relations operates today by now adding new arbitrary columns to the relations tables and APIs that don't make any sense for other entities. So to support variant tracking for variant based entities we can create another DB table with links to the relations table and variants data and then create new APIs for this type of support. We we're recently speaking about this backlog item since we know it is still outstanding.

dawoe:
Hey @Shazwazza 

Thanks for the feedback. Good to see it is still on the radar.

Dave

poornimanayar:
Hey @dawoe ,

I have tested the content tracking as well now. The issue I have is around the variants again. Content tracking on variant is very confusing from an editor's perspective. On a single site the content tracking works very well. No issues around that at all. But on variants here is what I noted down.

I have English UK and US with US as the default language
1) I link to a blog page in US site from various fields on content page in the US site . The blog page shows Content Page US as reference - Fine
2) I switch to UK site and link from Contact page to Blog page (both in UK site), the blog page now shows Contact Us US as reference. It should be UK 
3) I now go to blog page on US site and when I see reference I see Contact page US as a reference which is totally wrong as I have linked on UK site and not US site.

I think tracking for variants is what is really lacking for media and content tracking. 

Poornima

dawoe:
Hi @poornimanayar 

That is not something I can solve in this PR. To make the variant tracking work we will need some guidance from HQ on how to do this.

Dave

src/Umbraco.Web.UI.Client/src/common/resources/media.resource.js
```diff
@@ -550,36 +550,6 @@ function mediaResource($q, $http, umbDataFormatter, umbRequestHelper) {
                         "Search",
                         args)),
                 'Failed to retrieve media items for search: ' + query);
-        },
-
-        getPagedReferences: function (id, options) {
```
Shazwazza, line 555: This is a breaking change

Shazwazza: I think it's fine for a breaking change, i don't believe that this method is documented so people probably haven't used it but would mean this needs to be marked as a breaking change.

dawoe: I could put it back and call the new api method or the new resource. So it would not break, but we can mark the method as obsolete

dawoe: I put the method back so the change is non-breaking. The method calls the new resource. Also marked as deprecated for the docs

dawoe:
@poornimanayar can this be merged ?

nul800sebastiaan:
FYI, as discussed with Dave, I've updated the base branch for this for now, upcoming PRs can target `v8/item-tracking`. There will for now be a known limitation on variants, which is fine for now, this will need to be addressed at some point. 

Let's work on this separate branch for now so it doesn't immediately end up in dev and then we can merge it into a release when it feels more complete.

poornimanayar (approved):
Merging the work :-)

PR #8710, opened 2020-08-21
Add `!important' to the color in the style attribute

### LINKED ISSUE #8688: V8.7RC - BlockList Editor: Icon color does not override color defined on DocumentType
In the BlockList Editor it is possible to define colors for the background and the icon of the item to represent it in the "Add content" dialog. But only the background color is changed. The icon keeps the color that was set on the DocumentType it represents.

## Umbraco version

I am seeing this issue on Umbraco version: 8.7.0-rc

Reproduction
------------

### Steps to reproduce

   * Create a new property on a document type and use the BlockList property editor for it
   * In the Editor settings, create a new block and assign it an icon using the red color
   * In the Configuration for the block, in its "Catalogue appearance" section, set its Background color to `#ff8000` and the Icon color to `#000000`
   * When using the blockeditor in the Content section, observe that the block's icon color will be the red color initially chosen for the document type's icon.

### Expected result

I expected to see the block as a black icon on an orange background

### Actual result

The icon is shown using the red color that has been defiend on the document type.

### Cause of bug

(This is very likely a CSS specificity issue where the class for the icon has a higher specificity than the injected style (or that the injected style is applied to an ancestor of the icon, whereas the icon gets the specific icon color class.)

greystate:
### Why have I changed from `ng-style` attribute to `ng-attr-style` ?

It turns out that `!important` doesn't work with the `ng-style` attribute, so I have switched to use `ng-attr-style` instead, which works by setting the exact string, instead of using an object.

### Prerequisites

Fixes #8688 

### Description

* Create a new property on a document type and use the BlockList property editor for it
* In the Editor settings, create a new block and assign it an icon using the red color
* In the Configuration for the block, in its "Catalogue appearance" section, set its **Background color** to `#ff8000` and the **Icon color** to `#000000`
* When using the block editor in the Content section, observe that the block's icon color will be the black color, thus correctly overriding the red color initially chosen for the document type's icon.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/blockcard/umb-block-card.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/blockcard/umb-block-card.html
@@ -1,6 +1,6 @@
 
     <div class="__showcase" ng-style="{'background-color':vm.blockConfigModel.backgroundColor, 'background-image': vm.blockConfigModel.thumbnail ? 'url('+vm.blockConfigModel.thumbnail+'?upscale=false&width=400)' : 'transparent'}">
-        <i ng-if="vm.blockConfigModel.thumbnail == null && vm.elementTypeModel.icon" class="__icon {{ vm.elementTypeModel.icon }}" ng-style="{'color':vm.blockConfigModel.iconColor}" aria-hidden="true"></i>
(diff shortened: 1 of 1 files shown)


nielslyngsoe (approved):
Looks good

greystate:
Didn't this PR make it into the 8.7.0 release?

PR #8711, opened 2020-08-22
Accessibility: Fixes for "insert code snippet" view in macro pickers

BatJan:
### Prerequisites

### Description
During the work on #8522 I discovered that this view did not contain proper `<button>` elements, for the different insert options, which means the focus lock won't be able to pickup the first focusable element properly and puts the focus on the "close" button making one unable to tab to the options inside the overlay. This is fixed in this PR with the added buttons. While I had the view opened I fixed some HTML semantics as well adding `h2` for the title and wrapped the description in `<p>` - I also added proper fallback texts in case the dictionary keys have not been translated for a certain language.

The view visually appears the same as before these changes - I decided to make this an independent PR in an attempt to keep the #8522 as clean as possible 👍

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-insert-code-box.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-insert-code-box.less
@@ -21,6 +21,7 @@
     font-size: 15px;
     margin-bottom: 5px;
     font-weight: bold;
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Sweet, looks good @BatJan - great job! 👍

PR #8712, opened 2020-08-22
Accessibility: Grid config datatype improvements

BatJan:
### Prerequisites

### Description
During my work on the focus lock for the infinite editor in #8522 I noticed during the testing that the "button" for adding layout or row configurations on the grid content datatype was not tabable since it was wrapped in `<a>` without a `href` attribute. But since it's not really a link I've refactored it to be `<button>` instead and fixed the CSS to add hover and focus styles for the button element.

Furthermore I have added some screen reader friendly text labels as well.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/layoutconfig.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/layoutconfig.html
@@ -33,17 +33,19 @@
                                 <div class="tb" style="height: 70px; border-width: 2px; padding: 2px">
                                     <div class="tr">
 
-                                        <a class="td uSky-templates-column"
+                                        <button class="btn-reset td uSky-templates-column"
                                            ng-class="{last:$last, selected:section==currentSection}"
                                            ng-repeat="section in currentLayout.sections"
                                            ng-click="configureSection(section, currentLayout)"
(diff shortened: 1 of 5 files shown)


src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/layoutconfig.html
```diff
@@ -33,17 +33,23 @@
                                 <div class="tb" style="height: 70px; border-width: 2px; padding: 2px">
                                     <div class="tr">
 
-                                        <a class="td uSky-templates-column"
+                                        <button class="btn-reset td uSky-templates-column"
```
nathanwoulfe, line 36: Hey @BatJan should this button also have the type attribute?

BatJan: @nathanwoulfe Ah good catch - Yes I think it's better to be explicit even though it works fine without it. But I'll add it here too 👍

nul800sebastiaan:
Good stuff! Thanks again @BatJan! 👍

bjarnef:
@BatJan it seems you missed changing the end tag to a `</button>` closing tag here: https://github.com/umbraco/Umbraco-CMS/pull/8712/files#diff-f76bf6866ae588340b2f93d1b918aa53R53

PR #8713, opened 2020-08-24
Adding `display: none;` prevents `ng-show` from working

### LINKED ISSUE #8705: Redirects are not shown on info tab of a content item
Somewhere in V7 @marcemarc introduced showing redirects on the info tab (https://github.com/umbraco/Umbraco-CMS/pull/2452)

Today I was working on #8703 and I noticed these are not shown anymore

## Umbraco version

I am seeing this issue on Umbraco version: latest version of v8/contrib branch

Reproduction
------------

Do a clean install of the latest v8/contrib branch and use the starter kit. Rename a node. Check if the redirect is added by looking at the redirects dashboard.

Now go to the item and check the info tab. No redirects are shown

This is probably because of the style display:none on the div element https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/content/umb-content-node-info.html#L23

But if you use a debugger you can see the code to get the redirects is still executed each time you switch to the info tab.

I don't know the reason why this is not shown anymore.

But either fix this by removing the code entirely to reduce unneeded api requests. Or show the redirects again.

nul800sebastiaan:
### Prerequisites

If there's an existing issue for this PR then this fixes #8705

### Description
As described in #8705: 

- Rename a node while the built-in URL tracker is enabled
- Go to the info panel and see if the redirect panel (see below) shows up
- Before applying this change it won't show up, after applying this it will show up
- Also test that the redirect panel is still hidden on items that have no redirects.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/content/umb-content-node-info.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/content/umb-content-node-info.html
@@ -22,7 +22,7 @@
                 </ul>
             </umb-box-content>
         </umb-box>
-        <umb-box data-element="node-info-redirects" style="display:none;" ng-cloak ng-show="!urlTrackerDisabled && hasRedirects">
+        <umb-box data-element="node-info-redirects" ng-cloak ng-show="!urlTrackerDisabled && hasRedirects">
             <umb-box-header title-key="redirectUrls_redirectUrlManagement"></umb-box-header>
             <umb-box-content class="block-form">
                 <div style="position: relative;">


nathanwoulfe:
@nul800sebastiaan shouldn't this be targeting v8/contrib?

nul800sebastiaan:
It really should huh.. not sure what happened there 😅

nul800sebastiaan:
I'll merge it to the appropriate branches of course!

Issue #8725, opened 2020-08-25
umb-multiple-textbox does not have an accessible label on input elements


lordscarlet:
The input fields created by _Umbraco.Multipletextstring_ do not have an accessible label

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.1

Reproduction
------------

The property editor for _Multiple Text String_ makes no attempt to provide an accessible label. 

### Specifics

https://github.com/umbraco/Umbraco-CMS/blob/44b84c4cddb44a60b6ea73fe5870c251328fa857/src/Umbraco.Web.UI.Client/src/views/propertyeditors/multipletextbox/multipletextbox.html#L5-L9

### Steps to reproduce

1. Create a document type containing a property using the _Umbraco.Multipletexstring_ editor
2. Create a node using the created content type
3. Add a test string to reveal an input

### Expected result

Each input field should have an associated `<label>` with a `for` attribute referencing the `id` of the `<input>` field

### Actual result

The `<input>` has no label

umbrabot:
Hi @lordscarlet,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

lordscarlet:
@nul800sebastiaan at risk of being confusing, I am going to do somewhat of a cross-post here. On #8726 we are discussing a similar situation, and I mentioned two options. From a UI perspective, option 1 (`aria-labelledby`) would look like this:

![image](https://user-images.githubusercontent.com/17951/91844117-e00abf80-ec24-11ea-962f-a1b248df8147.png)

It passes [ANDI](https://github.com/SSAgov/ANDI) validation (which matters for my implementation, others may have different standards)

The alternative would look like this (possibly with some different formatting in the end? Maybe it's fine):

![image](https://user-images.githubusercontent.com/17951/91849138-91135900-ec29-11ea-86ef-ffec10dedb41.png)

I don't know what the preference is from a design/UX side in balance with accessibility.

lordscarlet:
[...]

PR #8726, opened 2020-08-25
Added screen reader only labels for domain and language fields in culture and hostnames overlay

### LINKED ISSUE #8724: Culture and Hostnames inputs do not have a label
Under _Culture and Hostnames_ for a root node, the _Domain_ and _Language_ inputs do not have an accessible label

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.1

Reproduction
------------

### Bug summary

The _Domain_ and _Language_ inputs do not have an accessible name

### Steps to reproduce

1. Navigate to the _Content_ section
2. View the actions on a root node
3. Choose _Culture and Hostnames_

### Expected result

A screen reader should be able to view the name for each input field, typically via a `<label>` with a `for` attribute.

### Actual result

The input fields do not have an associated label

poornimanayar:
Hi,

I have added a screen reader only label to domain and language fields in culture and hostnames overlay.

If there's an existing issue for this PR then this fixes #8724 

Poornima


lordscarlet:
Why not use the existing contextual name, further up in the HTML? 

```<h5 class="umb-pane-title"><localize key="assignDomain_setDomains">Domains</localize></h5>```

**Comment by poornimanayar** [comment]
Hey @lordscarlet ,

I did consider that. Let me see again whether I can play around with the localize directive and use that as `aria-labelledby`. Thanks for the comment :-)

Poornima

**Comment by lordscarlet** [comment]
Ah, I see. The issue is that the one "Domain" `<h5>` can apply to multiple instances of the input.

**Comment by poornimanayar** [comment]
I think I have found a way to do that. Watch this space :-)

File: src/Umbraco.Web.UI.Client/src/views/content/assigndomain.html [hunk 1]
```diff
@@ -32,11 +32,15 @@ <h5 class="umb-pane-title"><localize key="assignDomain_setDomains">Domains</loca
                         <thead>
                             <tr>
                                 <th>
-                                    <localize key="assignDomain_domain"></localize>
+                                    <div id="assignDomainLabel" localize="title" title="@assignDomain_domain">
```
nul800sebastiaan, line 35: Is this the way to do this? It seems a bit "heavy" to me to add a div here which is a block-level implement, maybe a span will do or maybe this can be even more clever? 🤔

nul800sebastiaan (commented):
@lordscarlet I'm wondering if we need to add divs with the same content that's already there in the `<localize>` elements. Is adding a `<label>` instead of `aria-labelledby` and option? Forgive my questions, this is a new pattern I haven't seen yet so I'm a bit out of my depth. 

@poornimanayar I've updated your PR to not introduce a whole lot of whitespace changes, make sure to pay attention to that when you commit your changes. It looked like a huge change even if there was onl actually a few lines changed. Also the indentation had changed in the html file, try to keep those consistent please! 👍

src/Umbraco.Web.UI.Client/src/views/content/assigndomain.html
lordscarlet: I agree, however, this seems to be existing code that was just reformatted?

lordscarlet:
> @lordscarlet I'm wondering if we need to add divs with the same content that's already there in the `<localize>` elements. Is adding a `<label>` instead of `aria-labelledby` and option? Forgive my questions, this is a new pattern I haven't seen yet so I'm a bit out of my depth.

@nul800sebastiaan You would need to add a `label` for each row of data entry. In this case, the existing `label`, to the sighted user, applies to all fields. a `label`, by definition, can only refer to _one_ `input` field. So right now you have:

```
Domains

[ blah.com ]
[ other.com ]
[ site.com ]
```

If you wanted to use `label` to solve this, you would need:

```
Domains

Domain 1 [ blah.com ]
Domain 2 [ other.com ]
Domain 3 [ site.com ]

```
or something similar. Ideally the experience would be the same/similar for both those using a screen reader and those that are not. `aria-labelledby` is new to me as well, but it seems to get around this situation. It is not ideal, but the alternative is a bit odd as well. It is closer to changing the intent of the design/etc, so I think community contributors are a little hesitant to make such a change. Having said that, from an accessibility point of view, the second option would be great. I am dealing with the same scenario in attempting to address https://github.com/umbraco/Umbraco-CMS/issues/8725.

src/Umbraco.Web.UI.Client/src/views/content/assigndomain.html
BatJan: @poornimanayar I think you should be able to add the "id" attribute for the `<localize>` directive, which if I remember correctly would then be put on the output element so the `<span>` would get the `id` attribute, so you don't need to wrap it in the `<div>` and make it inline :-)

PR #8728, opened 2020-08-26
Make it possible to use existing base tests from the unit test project in the BenchmarkTests project.

nzdev:
### Prerequisites

### Description
Adds a benchmark test for examine index population. Makes it possible to reuse the existing test classes in the benchmark project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -174,3 +174,4 @@ cypress.env.json
 /src/Umbraco.Tests.AcceptanceTest/package-lock.json
 /src/Umbraco.Tests.AcceptanceTest/cypress/videos/
 /src/Umbraco.Tests.AcceptanceTest/cypress/screenshots/
+/src/Umbraco.Tests.Benchmarks/App_Data/TEMP/TypesCache
--- a/src/Umbraco.Tests/Testing/TestOptionAttributeBase.cs
+++ b/src/Umbraco.Tests/Testing/TestOptionAttributeBase.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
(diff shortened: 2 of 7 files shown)


src/Umbraco.Tests/Testing/TestOptionAttributeBase.cs
```diff
@@ -24,13 +26,69 @@ public static TOptions GetFixtureOptions<TOptions>(Type type)
         {
             return Get<TOptions>(type, null);
         }
+        /// <summary>
+        /// Find the class name and method name that has the attribute
+        /// </summary>
+        /// <param name="callerName">Out Caller Method Name</param>
+        /// <returns>Caller Class Name</returns>
+        public static string GetBenchmarkTestCaller(out string callerName)
```
Shazwazza, line 34: This is pretty nasty, do we really need this? Surely we can achieve what is needed in a nicer way?

.gitignore
```diff
@@ -174,3 +174,4 @@ cypress.env.json
 /src/Umbraco.Tests.AcceptanceTest/package-lock.json
 /src/Umbraco.Tests.AcceptanceTest/cypress/videos/
 /src/Umbraco.Tests.AcceptanceTest/cypress/screenshots/
+/src/Umbraco.Tests.Benchmarks/App_Data/TEMP/TypesCache
```
Shazwazza, line 177: Is this required?

Shazwazza (changes requested):
Sorry for the delay! I have left a couple of comments inline. I'm just not entirely sold on the GetBenchmarkTestCaller, there must be a nicer way? Does the benchmark test need to use this attribute? Could it instead bootstrap itself just by calling code?

PR #8732, opened 2020-08-26
V8.7RC Prevent usage of NestedContent in BlockEditors

nielslyngsoe:
We have several issues in regards to use NC in BL. But there is no reason to have NC now that BL is available. Therefore we want to nudge Devs to use Block List, and mark the usage of NestedContent as unsupported in the context of Block Editors.

---
_This item has been added to our backlog [AB#7946](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7946)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
@@ -231,7 +231,8 @@
         var notSupportedProperties = [
             "Umbraco.Tags",
             "Umbraco.UploadField",
-            "Umbraco.ImageCropper"
+            "Umbraco.ImageCropper",
+            "Umbraco.NestedContent"
         ];
 
 


dawoe:
What is the reason for not allowing NC in block editor ?

I could imagine having a FAQ block. Where the different questions (and answers) n the block are managed using nested content.

Using a nested block editor seems overkill. A block editor requires more configuration as a nested content property editor.

I can come up with numerous scenario's where having a nested content in side a block editor can be usefull.

nielslyngsoe:
Hi @dawoe 

Thanks for your input, I'm sorry but I don't understand the argument so maybe I can ask you to elaborate a bit here, so I can understand your point of view?

Maybe I can start by adding some perspectives from my point of view:

Using Nested Content inside Block List or using Block List inside Block List is basically the same thing.

Looking at minimal configuration of Block List is kind of simpler than NC, as you don't need to pick a property-group. I see there is one more click to edit the label, but I don't think that's the point you are going for, right? :-)

Block List is here to replace Nested Content, so when you create new projects there should be no need for you to use Nested Content anymore, anything we can do to make BL as easy to use as you find NC please let us know so we can get there. Thanks :-)

dawoe:
Hi @nielslyngsoe 

I only had a quick play with the block list editor. Maybe I need a more in depth look. But what I have seen is there are a lot of config options. But maybe they are not all needed to mimick NC behavior.

Dave

nielslyngsoe:
@dawoe yep, so it is a new config screen that is different from NC. But you reach the same setup as NC with just turning on "inline editing", and then the only disadvantage coming from NC is that labels can first be edited when opening the configuration overlay for each individual block.

Hopefully learning this new config will save you time overall, as this config can be used for more scenarios than NC.

markprouse:
The only problem with removing NC from BL is if you require a sub list which only has and only ever will have one type in it then having Block List is very weighty. With NC in C# I can go; item.ncItems.propertyOfNc which is nice and simple. Where if items in this example was a block list its a lot more complex with casting and other stuff to get to those properties. Unless you know of a nice way to access them? Also if you use block list, there is surely no point showing the pick block item library window if you only want to allow one type to be added? Like i say the whole thing is very weighty in the use of the list only ever being required to hold one type.

nielslyngsoe:
@markprouse we need a single item mode for Block List Editor, to support the UX flow you request.
There are several problems with how Nested Content is made, which will not work in collaboration with Block Editors. Therefore it's not supported.

Issue #8721, opened 2020-08-25
V8.7RC - Grid layout leaves visible div tags


nikolajlauridsen:
When using the grid layout and rendering it with ```@Html.GetGridHtml(Model, "propertyName")``` the grid layout leaves a visible ```<div class='container'>``` and ```</div>``` around every column section in the view, see screenshot.

![div tag issue2](https://user-images.githubusercontent.com/16456704/91150232-6884df80-e6bc-11ea-9e83-3370de70b4a1.PNG)

## Umbraco version

I am seeing this issue on Umbraco version: 8.7RC

Reproduction
------------

1. Create a document type
2. Add a gridlayout editor
3. Create a piece of content with the document type
3. Add 1 or more column sections to the grid layout editor, containing a piece of data, I tested it with headline and rte
4. Render it to the front end with: ```@Html.GetGridHtml(Model, "propertyName")```

---
_This item has been added to our backlog [AB#7920](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7920)_

nul800sebastiaan:
Aha.. well that is definitely wrong, it should fall back to the `bootstrap3` layout if you don't specify a layout (https://our.umbraco.com/documentation/getting-started/backoffice/property-editors/built-in-property-editors/grid-layout/render-grid-in-template).  

This is a regression bug, this used to work in 8.6.

ronaldbarendse:
This is because the `Bootstrap3.cshtml` grid view still uses `@Html.If()` for conditionally rendering HTML, but that helper method didn't correctly HTML encode the value (and that's fixed in https://github.com/umbraco/Umbraco-CMS/pull/6545).

The fix is already mentioned in the issue: https://github.com/umbraco/Umbraco-CMS/issues/6275#issuecomment-650399031 (that's also correctly marked as a breaking change). PR https://github.com/umbraco/Umbraco-CMS/pull/8735 fixes this 👍

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/issues/8727

Cherry picked for 8.7 in 8a074616d38bedaf1a8266d2e5dd23a15a651272

bielu:
@nul800sebastiaan shouldn't be that tagged as breaking? I am not seeing that label in either on pr or on that issue

nul800sebastiaan:
[...]

PR #8735, opened 2020-08-26
V8.7RC: Change grid view @Html.If to conditional statement

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8721.

### Description
This changes the `@Html.If()` helper method that now correctly HTML encodes the passed in value to prevent XSS attacks with a normal conditional statement.


nul800sebastiaan:
Thanks @ronaldbarendse! Of course that was the problem, almost forgot about it! 👍

PR #8736, opened 2020-08-26
V8.7RC: Recreate UDIs for pasted Block Editor entries

nielslyngsoe:
Our front-end rendering uses the UDIs of ElementTypes for caching, meaning that we have to ensure these are unique. Before this fix, we just copied the raw data of an elementType without manipulating it. Though pasted entries are created as new, their data is parsed on, leaving the inner properties of Block Editors with the same UDIs as their source. This problem is first visible when rendering these in the front-end as the first rendered ElementType with that UDI will be cached and stay as the data shown for other Blocks/ElementTypes using the same UDI.
This PR replaces UDIs on Paste, ensuring they are unique.

To reproduce, make this scenario happen:
image

You need three levels of Block Editors to imitate the scenario of copying data that contains an inner property of Block List Editor. The reason for this is because when copying a Block List Entry you get the ContentTypeModel, meaning the second-level property appears as normal properties of a node {alias: .., editor: .., value: ...} The value of the Second Block Editor will be the raw value we store of a Block Editor, aka. {layout: ..., contentData:, ..., settingsData: ...} This we need to ensure we interpret and update UDIs of.

Check that the ´contentUdi´ and ´settingsUdi´ are changed when pasting this entry.

Notice this feature is solving the broad need for Block Editors in general, meaning it's not specific towards the Block List property editor.

---
_This item has been added to our backlog [AB#7967](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7967)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/blockeditor.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/blockeditor.service.js
@@ -4,26 +4,106 @@
  *
  * @description
  * <b>Added in Umbraco 8.7</b>. Service for dealing with Block Editors.
- * 
+ *
  * Block Editor Service provides the basic features for a block editor.
  * The main feature is the ability to create a Model Object which takes care of your data for your Block Editor.
- * 
- * 
+ *
+ *
  * ##Samples
  *
  * ####Instantiate a Model Object for your property editor:
- * 
+ *
  * <pre>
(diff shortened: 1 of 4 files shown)


nikolajlauridsen:
All Blocks get a new UDI now
image

Both when pasted into the same BlockList, and when pasted into a new content node 👍

nikolajlauridsen (approved):

PR #8738, opened 2020-08-26
Userpicker view: Change anchor element to button

BatJan:
### Prerequisites

### Description
This is a very very small PR changing the `<a>` tag to a `<button>` since it's not a link pointing to anywhere - Improves accessibility and screen reader announcement.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/users/umb-user-picker-list.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/users/umb-user-picker-list.less
@@ -8,6 +8,8 @@
     margin-bottom: 5px;
     padding: 10px;
     align-items: center;
+    width: 100%;
+    text-align: left;
 }
 
 .umb-user-picker-list-item:active,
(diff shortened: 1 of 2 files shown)


abjerner:
Hi @BatJan - I hope your doing well 😉 

Changing the `<a>` to a `<button>` indeed makes sense, and is also in line with some of your past PRs, and it looks absolutely correct when skimming through the changes. Nevertheless, I'll try to give it a spin as well to be sure, and report back here with my findings.

Thanks 👍

BatJan:
Hey @abjerner Thank you mate 😃  I'm doing good thanks - Hope you're are well too 😎

abjerner:
Thanks @BatJan. Although a but to the busy side, I'm doing good as well 😄 

The `<button>` element appears to be missing some styling to look like the `<a>` element:

1. The `<button>` element doesn't take full width, so you can't toggle the user to by clicking the empty area to the left
2. There is a bit of extra space to the left of the user name

image

In the screenshot above, the first item is with the `<button>` element, while the second item is with the `<a>` element (both shown with hover).

If you can fix this, it should be good to merge 😉

abjerner (changes requested):
Aaaand I should have posted by comment as a review instead of a regular comment. If you can make the changes that I described in my previous comment, I'd be happy to merge 😉

BatJan:
@abjerner Good catch mate 👍 I've updated the styling it looks and acts like before the conversion to `<button>` 😎

nul800sebastiaan:
Looking good now indeed! 👍 Thanks @BatJan and @abjerner!

PR #8739, opened 2020-08-26
Userpicker: Add warning overlay when the "Remove" button is hit

BatJan:
### Prerequisites

### Description
I have added a warning dialog when one hits the "Remove" button next to a user listed in the user picker so that the user is not removed immediately but you have to confirm your action - This is aligning the UI with what happens in other places in the backoffice as well.

**Before**
userpicker-before

**After**
userpicker-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/users/group.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/users/group.controller.js
@@ -288,7 +288,7 @@
                     view: "views/users/views/overlays/remove.html",
                     username: selection[index].username,
                     userGroupName: vm.userGroup.name.toLowerCase(),
-                    submitButtonLabelKey: "contentTypeEditor_yesDelete",
+                    submitButtonLabelKey: "defaultdialogs_yesRemove",
                     submitButtonStyle: "danger",
 
                     submit: function () {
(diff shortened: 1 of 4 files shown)


nul800sebastiaan (changes requested):
Thanks @BatJan - to make it consistent can you change the "Yes, delete" button to "Yes, remove" instead please? It's a confusing right now to mix the terms.

bjarnef:
I recently changes the user picker property editor to use userpicker infinite editor instead or entity picker. Does the user picker property editor also shows this warning?

BatJan:
@nul800sebastiaan Yes of course - Intended to do that but missed it doh! Will have a look later 👍

BatJan:
@bjarnef Not sure - Can you share a screendump or gif of where I need to look please? 😃

bjarnef:
@BatJan I am not at my computer at the moment, but you could create an datatype instance of user picker and then use this on a new property on a document type.

BatJan:
@nul800sebastiaan The changes you requested have been pushed 👍 

@bjarnef Nope, this PR does not display it there yet since the "remove" button is not displayed in the infinite overlay but once you have picked the user. So it's in a different view that needs to be changed. I'd like to make that a separate PR at some point.

bjarnef:
@BatJan yes, I just meant both use the user picker infinite editor, but the remove function is implemented where the infine editor is used.

BatJan:
@bjarnef Yes it will require another PR :-)

nul800sebastiaan:
Wonderful, thanks @BatJan! 👍

PR #8743, opened 2020-08-27
Update WebSecurity.cs to fix back office culture

idseefeld:
### Issue
Back office culture is currently not set. 
### Fix
ValidateCurrentUser gets called from Umbraco.Web.BackOffice.Filters.UmbracoAuthorizeFilter and maybe this could be the right place to ensure culture as set by user's property - right?

_One problem remains: The user needs to logout and login again to see the changes._

### DIFF
--- a/src/Umbraco.Web.Common/Security/WebSecurity.cs
+++ b/src/Umbraco.Web.Common/Security/WebSecurity.cs
@@ -6,6 +6,7 @@ using Umbraco.Core.Configuration;
 using Umbraco.Core.Hosting;
 using Umbraco.Core.Models.Membership;
 using Umbraco.Core.Services;
+using Umbraco.Core.Security;
 using Umbraco.Extensions;
 using Umbraco.Web.Security;
 using Umbraco.Core.Models;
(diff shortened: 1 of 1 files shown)


Shazwazza:
Hi @idseefeld !

The logic for the setting the user's culture is part of the authentication mechanism which occurs at the beginning of a request/middleware. Setting the culture for the back office controllers is set via `BackOfficeCultureFilter` which is based on the user's culture in the authentication ticket. Do you have a test case for where this is not working?

idseefeld:
Hi @Shazwazza,
`BackOfficeCultureFilter` sounds good. I haven't found that one.

1. I found the issue on German Windows 10 system. 
2. I have just downloaded the netcore/netcore branch, build the solution an started the `Umbraco.Web.UI.NetCore` project within Visual Studio 2019.
3. The initial user was configured for English/US, but the back office was displayed in `de-DE`. This was probably caused by my system default language.

I have also checked, that **not** the browser default language caused the issue.
Setting the Windows language to English/US did change the back office to English.

Within one of the Api controllers (e.g. `Umbraco.Web.BackOffice/Controllers/SectionController.cs` method `GetSections`, at least after `sectionModels line 51` was set) you can see the culture is wrong.

Hopefully this gives you enough details!

Dirk

nul800sebastiaan:
I'll close this one for now as it's not the right location to do this, we can try again once we get some clarification from @Shazwazza 👍

PR #8744, opened 2020-08-27
V8.7RC: Moved validation indication one element up.

nielslyngsoe:
In regards to a Block of Bloc List with a custom stylesheet.
Before this fix, the validation indication was not visible cause it was target to an element inside the scoped stylesheet.

This fix moves the CSS to the parent element, making it display cause it's then out of the scoped element.

Result:
image

---
_This item has been added to our backlog [AB#7978](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7978)_


nul800sebastiaan:
Sweet! Thanks Niels!

PR #8745, opened 2020-08-27
User property editor: Add "Remove" warning dialog

BatJan:
### Prerequisites

### Description
This will add a "remove warning" dialog when one clicks the "Remove" button on the user property editor (If it's been setup on a document type of course).

**Before**
userpicker-property

**After**
userpicker-property-after

@nul800sebastiaan During the work on this I was wondering... Did it not use to be possible to select more than one user using the picker? It appears to me that it's currently only possible to select one user - But in the userpicker property editor view's code it appears the selected users are pushed into an array so I suspect the intention is that it should be possible to select X amount of users? I'll see if I can do a PR to fix this if my observations are correct 😃


nul800sebastiaan:
@BatJan Sure, our wish for the future is for all our pickers to have two default options: min and max that can be picked.

nul800sebastiaan:
Thanks @BatJan! I reverted the 2 files that only had whitespace changes and removed a console.log for you! ;-)

Other than that this looks good to me! 👍

BatJan:
@nul800sebastiaan I'm getting sloppy... Will try to double check before pushing the next PR 😅

PR #8747, opened 2020-08-28
Resolves issue with rendering back-office media selections when synchronous IO is not allowed.

AndyButland:
This allows picked media to render in the back-office if we don't have `options.AllowSynchronousIO = true;` configured (which I understand in the long-run, we don't want to do).


src/Umbraco.Web.BackOffice/Controllers/DetermineAmbiguousActionByPassingParameters.cs
```diff
@@ -102,7 +103,11 @@ private IEnumerable<string> GetValue(ParameterDescriptor descriptor, RouteContex
                 // IMPORTANT: Ensure the requestBody can be read multiple times.
                 routeContext.HttpContext.Request.EnableBuffering();
 
-                var body = _requestBody ??= routeContext.HttpContext.Request.GetRawBodyString();
+                // We need to use the asynchronous method here if synchronous IO is not allowed (it may or may not be, depending
+                // on configuration in UmbracoBackOfficeServiceCollectionExtensions.AddUmbraco()).
+                // We can't use async/await due to the need to override IsValidForRequest, which doesn't have an async override, so going with
+                // this, which seems to be the least worst option for "sync to async" (https://stackoverflow.com/a/32429753/489433).
+                var body = _requestBody ??= Task.Run(() => routeContext.HttpContext.Request.GetRawBodyStringAsync()).GetAwaiter().GetResult();
```
bergmania, line 110: Out of curiosity, whats the differnece between:

```
var body = _requestBody ??= Task.Run(() => routeContext.HttpContext.Request.GetRawBodyStringAsync()).GetAwaiter().GetResult();
```

and 

```
var body = _requestBody ??= routeContext.HttpContext.Request.GetRawBodyStringAsync().GetAwaiter().GetResult();
```

We might have some other stuff to update, if this is better :)

AndyButland: To be honest I've read about this 100s of times and still never quite sure what the best way to handle this need for "sync over async" :-).  There seems to be so many questions and suggestions out there, all of which say, "ideally use async/await".  But as we can't here, based on the SO reference in the comments, I went with this construct as the suggestion is this is less likely to suffer from deadlocks.

PR #8748, opened 2020-08-28
Fix: v7 BeginTransaction null reference after custom migrations

JustinGeschiere:
### Problem
After executing custom migrations, the following null reference exceptions was often encountered.
image

This caused the migration version to not be updated and all custom migrations to run again on start-up. The null reference could be easily fixed by using an existing property that has a built in null check. However, this was not the underlying issue since another exception would take its place:
image

### Cause
The underlying issue is that PetaPoco's _transactionDepth could be a negative number. This was caused by an IDisposable implementation on the database to always close the transaction and reducing the _transactionDepth, even if no transaction was active/present.

### Changes
- Replace direct object reference with property that has built in null check
image

- Add a check to close transaction only if db is in transaction
image

### Test
Unfortunately, this is not an easy one to test. We could reproduce this issue by running our custom migrations and there would be a big chance (almost always) this issue would occur. After this change this issue never occurred again.

The general steps would be:
1. Test with latest v7: run custom migrations and encounter the error (might require multiple tests before occurrence)
2. Test with branch: run the same custom migrations and never encounter the error (test multiple times)

### DIFF
--- a/src/Umbraco.Core/Persistence/PetaPoco.cs
+++ b/src/Umbraco.Core/Persistence/PetaPoco.cs
@@ -344,7 +344,7 @@ namespace Umbraco.Core.Persistence
                 _transactionCancelled = false;
 				OnBeginTransaction();
 			}
-            else if (isolationLevel > _transaction.IsolationLevel)
+            else if (isolationLevel > CurrentTransactionIsolationLevel)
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Makes total sense to me @JustinGeschiere - thanks for the help!

PR #8750, opened 2020-08-28
Netcore/fix install api url

idseefeld:
### Prerequisites

The install routing is broken

### Description
The current extension method GetInstallerApiUrl returns a url without / at its end. As umbraco.install.js combines "GetSetup" with this url, the route is broken.

@bergmania I hope this helps you to fix the issue;-)


bergmania:
Thanks @idseefeld...

Nice catch :)

Issue #8706, opened 2020-08-21
Mini search does not have <label>


lordscarlet:
The primary search has a label, which is good for accessibility. However, the mini search does not.

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.1

Reproduction
------------

### Bug summary

When searching the _Users_ (or any other) section, the mini search does not have a `<label>`

### Specifics

This effects anywhere with a mini-search, such as `/umbraco/#/users/users/users?filter=xyz`

![image](https://user-images.githubusercontent.com/17951/90908729-b939c700-e3a2-11ea-86d2-aea6e099382e.png)

### Expected result

The `<input>` should have an associated `<label>` with `for` attribute that matches the `id` of the `<input>`

### Actual result

There is no `<label>` or corresponding `id`.

umbrabot:
Hi @lordscarlet,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

nikcio:
@nul800sebastiaan @lordscarlet I believe that this can be closed. Was just looking through some old issues and this one seems solved in #10583

kjac:
Indeed this was fixed by https://github.com/umbraco/Umbraco-CMS/pull/10583 👍

PR #8752, opened 2020-08-28
Feature/temp 8706

lordscarlet:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8706

### Description
Added a `<label>` for `umb-mini-search` that mimics the implementation of `umb-search`

image

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/umb-mini-search/umb-mini-search.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/umb-mini-search/umb-mini-search.html
@@ -1,9 +1,10 @@
 <ng-form class="umb-mini-search" ng-class="{'--has-value': vm.model !== null && vm.model !== ''}" novalidate>
-    <i class="icon icon-search" aria-hidden="true"></i>
+    <label for="mini-search"><i class="icon icon-search"></i><span class="sr-only"><localize key="general_typeToSearch">Search...</localize></span></label>
     <input
         class="form-control search-input"
         type="text"
         localize="placeholder,label"
+        id="mini-search"
         label="@general_typeToSearch"
         placeholder="@general_typeToSearch"
         ng-model="vm.model"
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-mini-search.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-mini-search.less
@@ -48,4 +48,8 @@
         padding-left:30px;
     }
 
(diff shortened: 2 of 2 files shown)


src/Umbraco.Web.UI.Client/src/views/components/umb-mini-search/umb-mini-search.html
```diff
@@ -1,9 +1,10 @@
 <ng-form class="umb-mini-search" ng-class="{'--has-value': vm.model !== null && vm.model !== ''}" novalidate>
-    <i class="icon icon-search" aria-hidden="true"></i>
+    <label for="mini-search"><i class="icon icon-search"></i><span class="sr-only"><localize key="general_typeToSearch">Search...</localize></span></label>
```
nul800sebastiaan, line 2: Any reason you've removed `aria-hidden="true"` on the icon? These icons can not be properly articulated by screen readers and should be hidden. Especially now that you're adding a nice label around them! 🙂

src/Umbraco.Web.UI.Client/src/views/components/umb-mini-search/umb-mini-search.html
```diff
@@ -1,9 +1,10 @@
...
     <input
         class="form-control search-input"
         type="text"
         localize="placeholder,label"
+        id="mini-search"
```
nul800sebastiaan, line 7: I wonder what happens if we have multiple mini searches on one page.. not sure it can happen? Can we make this unique for each instance somehow you think?

nul800sebastiaan (changes requested):
Thanks @lordscarlet - I've left you some questions!

src/Umbraco.Web.UI.Client/src/views/components/umb-mini-search/umb-mini-search.html
lordscarlet: I wondered the same. I know currently I don't have any situation where two show up at the same time. I wasn't cure where to get an id to make it unique, but as I am working on another issue I have learned a little more about the structure of AngularJS and I think I can pull out something to make it unique. I will circle back to this after I finish that issue.

nul800sebastiaan:
This was superseded by: https://github.com/umbraco/Umbraco-CMS/pull/10583 which I've just merged, so I'll close this one! Better find something new to work on @lordscarlet! 😉

PR #8753, opened 2020-08-28
Removing "Shift-Tab" focus trap in Rich Text Editor

azdyrski:
Prevents focus trap on "shift tab" in RTE

### Prerequisites

### Description
Tab navigating through the properties when creating a new node does not allow Shift-Tab to navigate back to the previous property on a Rich Text Editor field.

rte-focus-trap

Changes sets display:none on the already non-visible input element that RTE is invoked using.  This removes it from the tab navigation that was trapping the focus.


nul800sebastiaan (changes requested):
Thanks @azdyrski! I think there might be a reason this element is not set to `display: none;`, likely that it wont initialize correctly in some browsers / scenarios if the input is set to this. Can you think of another way of doing this? Tabindex updated maybe? Not sure..

Issue #8689, opened 2020-08-20
V8.7RC Label with custom expression


bjarnef:
With the new Block List we can use a custom expression for label, e.g. "Block {{myPropertyAlias}}".

However it would be great if we could use basic angular expression like we can in Nested Content, e.g. "Block {{$index + 1}}".
This expression seems just to render "Block 1" for all blocks.

nielslyngsoe:
@bjarnef you should be able to use angular expressions, but $index is not available, you can only choose values of your content properties(by their alias). If you think index is a missing option, I think it would be fairly simple to make that available as well. :-)

bjarnef:
@nielslyngsoe yes I think this is a missing option :) I would be great this is a bit consistent with Nested Content name template. See the link getName function in Nested Content 🙂

nul800sebastiaan:
Additionally, @liquidkool is asking for the following feature:

> I want to populate the label field with content from the settings tab. I can get content from the content tab using `{{aliasName}}`, however this doesn't work with content from the settings tab. I've tried `{{settings.aliasName}}`. Not sure if its possible?

ronaldbarendse:
Adding support to access both data/content and settings would mean the existing expressions need to change from `{{myPropertyAlias}}` to something like `{{content.myPropertyAlias}}`, so you can access settings with the same alias using `{{settings.myPropertyAlias}}`.

I don't know what data the custom AngularJS views (for displaying previews within the back-office) have access to, but it would make a lot of sense to make the namings/models the same for the name template!

bjarnef:
The block label is generated in `getBlockLabel` method here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/8.7/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js#L102-L112

Where the config label is being interpolated here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/8.7/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js#L556-L558

However in Nested Content it pass in the index. Not sure if it is possible the generate the index without passing in the index or retrieve this index by the position of the item in the array.

Furthermore in Nested Content when using `$index` it actually render `$index + 1`:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js#L358

[...]

PR #8763, opened 2020-08-30
V8.7RC Block list adjustments

bjarnef:
### Prerequisites

### Description
While working on a solution for https://github.com/umbraco/Umbraco-CMS/issues/8689 I noticed various issues, which I have fixed in this PR.

- Accessibility enhancements.
- Added `rel="noopener"` for security reasons.
- Stored path `views/propertyeditors/blocklist/blocklistentryeditors/` in a variable.
- Added properties to `vm.labels` after localization.
- The add/plus button didn't had same width and height. Furthermore it looked like you never hit the button because it is moving when the cursor position change. Instead this is updated in `transform` property where it is vertically centered and moved 80% to the left, so the cursor doesn't cover the icon, but it still feels like you hit the button.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
@@ -543,7 +543,6 @@
                         unsupported: true
                     };
                     contentScaffold = {};
-
                 }
 
                 var blockObject = {};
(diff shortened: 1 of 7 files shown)


bjarnef:
@nielslyngsoe this seems to be a bit nicer. Futhermore maybe `umb-code-snippet` component could be used in the unsupportedblock editor?
https://github.com/umbraco/Umbraco-CMS/blob/v8/8.7/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/blocklistentryeditors/unsupportedblock/unsupportedblock.editor.html#L11

E.g. like this change for v8.8?
https://github.com/umbraco/Umbraco-CMS/pull/8548

nul800sebastiaan:
Wonderful updates @bjarnef - thanks very much! 👍

Issue #8727, opened 2020-08-25
V8.7RC Remove BlockListTemplateExtensions and default view/template


ronaldbarendse:
PR https://github.com/umbraco/Umbraco-CMS/pull/8273 added the new Block List editor including an extension class [`BlockListTemplateExtensions`](https://github.com/umbraco/Umbraco-CMS/blob/0140d906a5a66d3ba512332e845a64d36169f911/src/Umbraco.Web/BlockListTemplateExtensions.cs) and a default view/template [`/Views/Partials/BlockList/Default.cshtml`](https://github.com/umbraco/Umbraco-CMS/blob/0140d906a5a66d3ba512332e845a64d36169f911/src/Umbraco.Web.UI/Views/Partials/BlockList/Default.cshtml).

In my opinion, these should be removed because there are multiple things wrong with this approach and the implementation:
- The default view includes a wrapper `<div class="umb-block-list">` that doesn't make sense in any frontend framework (or custom styling for that matter);
- Customizations in this view might get overwritten when Umbraco is updated (e.g. when a bugfix needs to be applied to this file);
- Just like the Grid and Forms views, you might not use it and because of the previous reason shouldn't customize this view either, so it doesn't make sense to ship it (as content);
- The extension class and default view both check whether any layouts exist (which is redundant);
- The `GetBlockListHtml()` methods don't return an `IHtmlString`, but the concrete, ASP.NET specific `MvcHtmlString`;
- It renders all HTML into a string, instead of using the slightly more performant `RenderPartial()` that directly writes to the output stream (and because this editor might contain whole page layouts, this might result in a significant increase in memory usage/object allocations);
- It's just a little bit of boilerplate that even throws exceptions when you haven't created a partial view in a specific folder (by default `BlockList/Components/{ContentTypeAlias}`).

This also breaks the 'clean canvas' philosophy of Umbraco and there are multiple, better ways to help get people started with this editor:
- Remove the view and include the HTML generation/partial rendering into the extension method (e.g. like the MVC form control helpers `Html.DropDownList()`);
- Include the default view as example in the documentation;
- Only include the view in a starter kit.

The same arguments apply to the Grid and Forms views: do not pollute client projects with views that aren't used or shouldn't be modified. Lets keep it simple and clean as possible, just like how Nested Content currently is!

## Umbraco version

I am seeing this issue on Umbraco version: 8.7RC.

ronaldbarendse:
[...]

PR #8767, opened 2020-08-31
V8/bugfix/return IHtmlString from GetBlockListHtml

nikolajlauridsen:
Changed the return type from MvcHtmlString to IHtmlString in the BlockListTemplateExtensions as suggested in #8727. 

I'm having a hard time testing it since I still don't really understand how to get blocklist to actually work.

---
Edit:
Managed to figure out Block List and get it tested. It works just fine apart from having to edit the /Views/Partials/BlockList/Default.cshtml  from:

```
@inherits UmbracoViewPage<BlockListModel>
@using Umbraco.Core.Models.Blocks
@{
    if (Model?.Layout == null || !Model.Layout.Any()) { return; }
}
<div class="umb-block-list">
    @foreach (var layout in Model.Layout)
    {
        if (layout?.Udi == null) { continue; }
        var data = layout.Data;
        @Html.Partial("BlockList/Components/" + data.ContentType.Alias, layout)        
    }
</div>
```

to: 
```
@inherits UmbracoViewPage<BlockListModel>
@using Umbraco.Core.Models.Blocks
@{
    if (Model?.Layout == null || !Model.Layout.Any()) { return; }
}
<div class="umb-block-list">
    @foreach (var layout in Model.Layout)
    {
        if (layout?.ContentUdi == null) { continue; }
        var data = layout.Content;
        @Html.Partial("BlockList/Components/" + data.ContentType.Alias, layout)        
    }
</div>
```

But that seems very much unrelated to this PR

### DIFF
--- a/src/Umbraco.Web/BlockListTemplateExtensions.cs
+++ b/src/Umbraco.Web/BlockListTemplateExtensions.cs
@@ -4,6 +4,7 @@ using System.Web.Mvc;
 using System.Web.Mvc.Html;
 using Umbraco.Core.Models.Blocks;
 using Umbraco.Core.Models.PublishedContent;
+using System.Web;
 
 namespace Umbraco.Web
 {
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Cherry picked for 8.7 in 3d385cc1db49b73316b505c1fa654bd19736c039

PR #8768, opened 2020-08-31
Sort dictionary translations by display name

kjac:
### Prerequisites

### Description

It's been bugging me for a bit now... when you have more than a few languages, the dictionary translations become annoying to manage, because they're not sorted by language display name but rather by the order of which the languages were created (apparently):

image

image

This PR ensures that both the dictionary overview and the dictionary item editing sorts the translations by language display name:

image

image


nul800sebastiaan:
Wonderful, I'm happy with this! 👍

PR #8769, opened 2020-08-31
Sort the languages by their display name

kjac:
### Prerequisites

### Description

Following up on #8768, the list of languages seems to be sorted by the order of which they're created. 

image

This probably doesn't make a whole lot of sense to most people, so this PR ensures that the languages are explicitly sorted by name:

image


nul800sebastiaan:
Love it! Cheers @kjac! 👍

PR #8772, opened 2020-08-31
Umbraco to re-index data on background thread and fix potential Scope issue

Shazwazza:
Fixes issue with scopes being disposed or referenced incorrectly due to yield returns as this can capture a scope in the enumerator which will get passed to a background thread in Examine and cause some issues.

We saw this issue in netcore but the issue must still exist in v8 but we don't visibly see it for some reason.

The other issue is that the ValueSet lookup for content was done 3x when an IContent is saved and it should only be done 2x, one for published, one for unpublished.

The other issue is that the data lookups to build a ValueSet are intended to be done on a background thread. This is the case in v7 because the IEnumerable is lazy and passed all the way down to Examine's background thread but this doesn't occur in v8 because we need to iterate/split that value set before it's sent to Examine so the ValueSet is eagerly built within the request.

We can easily resolve this by using the background task manager and just pushing a task when IContent/IMedia/IMember is saved. This will return the execution to the UI quicker.

---
_This item has been added to our backlog [AB#8064](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8064)_

### DIFF
--- a/src/Umbraco.Examine/ContentValueSetBuilder.cs
+++ b/src/Umbraco.Examine/ContentValueSetBuilder.cs
@@ -60,6 +60,11 @@ namespace Umbraco.Examine
                 scope.Complete();
             }
 
+            return GetValueSetsEnumerable(content, creatorIds, writerIds);
+        }
+
+        private IEnumerable<ValueSet> GetValueSetsEnumerable(IContent[] content, Dictionary<int, IProfile> creatorIds, Dictionary<int, IProfile> writerIds)
+        {
             // TODO: There is a lot of boxing going on here and ultimately all values will be boxed by Lucene anyways
             // but I wonder if there's a way to reduce the boxing that we have to do or if it will matter in the end since
             // Lucene will do it no matter what? One idea was to create a `FieldValue` struct which would contain `object`, `object[]`, `ValueType` and `ValueType[]`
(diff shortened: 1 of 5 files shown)


clausjensen:
@Shazwazza I've reviewed and the code looks fine - tested and didn't see any problems.

I can however tell you that last week I had errors happening on a local install while testing something completely unrelated. It was originating from Examine parts and in particular the DeferedReIndex. So while you say we haven't seen this on v8 and that you didn't understand why because it should be there, I'm thinking you are exactly right, the problem does exist on v8 as well.

The errors were very *very* random and I could not consistently reproduce - and at some point I ended up not seeing it at all through 25 consecutive attempts at reproducing. So I can't really say whether this has fixed it or not, but I choose to believe it does.

nul800sebastiaan:
Cherry picked for 8.7.1 in 041d0469b618fa70c0ff07a3141f16a1e8d5b918

Shazwazza:
@clausjensen yes this absolutely affects some existing sites in v8 and could be the cause for all sorts of strange errors. I had a customer email me directly last week with a similar problem and I pointed to this PR which solved their issue so it's an important one.

PR #8774, opened 2020-08-31
Ensure directives use end-closing tag

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
I just noticed a few other issues with void elements (self-closing) with directives, which are custom HTML elements.

In this case is was causing layout issues in grid configuration settings/styles.

**Before**

chrome_2020-08-31_18-27-48

**After**

image

I have the following json configuration in grid settings:

```
[
  {
    "label": "Background color of row",
    "description": "Choose background color",
    "key": "background-color",
    "view": "colorpicker",
    "applyTo": "row",
    "prevalues": [
      "#D30535",
      {
        "value": "#32CD32"
      },
      "#81E6E0",
      "#109D95",
      "red",
      {
        "label": "White",
        "value": "#fff"
      },
      "#53B3AE",
      {
        "value": "#f00",
        "label": "Red"
      }
    ]
  },
  {
    "label": "Position",
    "description": "Set position",
    "key": "position",
    "view": "radiobuttonlist",
    "applyTo": "row",
    "prevalues": [
      "left",
      {
        "value": "center"
      },
      {
        "label": "Right",
        "value": "right"
      }
    ]
  }
] 
```

PR #1496, opened 2016-09-21
U4-9003 - Add checkboxlist prevalue editor

bjarnef:
Issue: http://issues.umbraco.org/issue/U4-9003

Currently there is a radiobuttonlist prevalue editor, but not a checkboxlist prevalue editor. A scenario where a checkboxlist prevalue editor would be useful it for e.g. a weekdays property.

Actually I was missing this in this PR https://github.com/umbraco/Umbraco-CMS/pull/984 to extend datepicker options. At the time of submitting the PR it created this weedays checkboxlist prevalue editor, but it would be great to have it in core: https://github.com/bjarnef/Umbraco-CMS/blob/0dbf686a0fa3bf8c3f4c701498b7b457ee9998f6/src/Umbraco.Web.UI.Client/src/views/propertyeditors/datepicker/weekdays.prevalues.html

Umbraco seems to include checklist-model, but I am not sure how much it is used: https://vitalets.github.io/checklist-model/ 

As an alternative it could have a controller to split the selected values.

It could also be useful it grid editor settings. I also included the option to define the labels , which @clausjensen added in this PR https://github.com/umbraco/Umbraco-CMS/pull/1486

image

Both radiobuttonlist and checkboxlist is missing the option to specify prevalues from package.manifest:
http://issues.umbraco.org/issue/U4-7030

PR #8775, opened 2020-08-31
Use umb-radiobutton component in grid configuration

bjarnef:
### Prerequisites

### Description
I noticed the grid configuration at settings/styles was still using a traditional radiobutton when using `radiobuttonlist` prevalue editor. This is now replaced with `<umb-radiobutton>` component.
In this gif I have rendered `model.value` to show the value changes.

To test this the following json can be used in grid editor settings/styles:

```
[
  {
    "label": "Background color of row",
    "description": "Choose background color",
    "key": "background-color",
    "view": "colorpicker",
    "applyTo": "row",
    "prevalues": [
      "#D30535",
      {
        "value": "#32CD32"
      },
      "#81E6E0",
      "#109D95",
      "red",
      {
        "label": "White",
        "value": "#fff"
      },
      "#53B3AE",
      {
        "value": "#f00",
        "label": "Red"
      }
    ]
  },
  {
    "label": "Position",
    "description": "Set position",
    "key": "position",
    "view": "radiobuttonlist",
    "applyTo": "row",
    "prevalues": [
      "left",
      {
        "value": "center"
      },
      {
        "label": "Right",
        "value": "right"
      }
    ]
  }
]
```

The layout issues with the box headers (note how the header take up valuable space on the left) have been fixed in this PR https://github.com/umbraco/Umbraco-CMS/pull/8774


bjarnef:
I still wonder if we should have a `checkboxlist` prevalue editor as well - both to use in grid settings/styles or in a custom property editor, e.g. it could be used for a weekday property in a custom datepicker property editor listing Monday, Tuesdag, etc. and only the selected days was enabled in a datepicker.

I submitted a PR for here https://github.com/umbraco/Umbraco-CMS/pull/1496 but was rejected at that time.

@nielslyngsoe is this something we want to add in Umbraco v8?

nul800sebastiaan:
Wonderful, thanks @bjarnef! 👍 

I would say a checkboxlist property could certainly be useful! But I'll defer to @nielslyngsoe for a definitive answer. 🙂

PR #8776, opened 2020-08-31
Query builder: Convert `<a>` to `<button>`

BatJan:
### Prerequisites

### Description
In this PR I have converted the `<a>` tags containing an empty `href` attribute to `<button>` instead improving accessibility and semantics. I have checked that the styling remains intact after the change 😉 

It's hard to show using a GIF but it's the query builder view that opens in the template editor when clicking the "Query builder" button and the changes affect the highlighted dropdowns. It also affects some of the dropdowns that appear depending on, what you select in the first dropdown.

querybuilder


nul800sebastiaan:
Thanks @BatJan 👍

PR #8778, opened 2020-09-01
Users - fix semantics + fix focus on umb-checkmark directive

BatJan:
### Prerequisites

### Description
I converted 2 `<a>` elements with empty `href` attributes to be `<button>` instead. I also added som styling in order to ensure that the "checkbox" in the "list layout" of the users would receive focus styles when tabbing. Currently the `<umb-checkmark>` directive only consists of a single icon - I will soon address this specifically in another PR unless someone beats me to it wrapping the icon in the directive inside a button by default and adjust the styling accordingly.

Attached are some screendumps of the areas that have been changed and need to be tested 😄 

**Create user add button**
create-user-add-button

**View users in list layout**
users-view-list-layout

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/users/views/users/users.html
+++ b/src/Umbraco.Web.UI.Client/src/views/users/views/users/users.html
@@ -211,9 +211,9 @@
                 <div class="umb-table-head">
                     <div class="umb-table-row">
                         <div class="umb-table-cell umb-user-table-col-avatar not-fixed" style="width: 70px; padding:10px 15px;">
-                            <a href="" style="text-decoration: none;" ng-click="vm.selectAll()">
+                            <button type="button" class="btn-reset umb-checkmark__action" style="outline: none;" ng-click="vm.selectAll()">
                                 <umb-checkmark checked="vm.areAllSelected()" size="xs"></umb-checkmark>
-                            </a>
+                            </button>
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web.UI.Client/src/views/users/views/users/users.html
```diff
@@ -211,9 +211,9 @@
                 <div class="umb-table-head">
                     <div class="umb-table-row">
                         <div class="umb-table-cell umb-user-table-col-avatar not-fixed" style="width: 70px; padding:10px 15px;">
-                            <a href="" style="text-decoration: none;" ng-click="vm.selectAll()">
+                            <button type="button" class="btn-reset umb-checkmark__action" style="outline: none;" ng-click="vm.selectAll()">
```
bjarnef, line 214: @BatJan could `umb-outline` or `btn-outline` class be used here?

BatJan: @bjarnef I'm not sure - Maybe. But I use this temporary class to fix some underlying states too - This fix is temporary untill I have a stab at refactoring the `<umb-checkmark>` anyway. But that is for another PR. Once I have a look at that it might be worth exploring the `umb-outline` or `btn-outline` classes 👍 😃

PR #8779, opened 2020-09-01
Macropicker: Fix semantics

BatJan:
…ckable items

### Prerequisites

### Description
With this PR I have changed the `<a>` element containing an empty `href` attribute to be `<button>`. I have also made a little styling adjustment so that there is 10px button margin from the search form and down to the card items so when one hovers the items there is a little space between the search box and the items.


nul800sebastiaan:
Happy days, cheers mate! 👍

PR #8780, opened 2020-09-01
Dictionary list - convert a to button and remove wrapping span element

BatJan:
### Prerequisites

### Description
This PR converts 2 `<a>` elements with an empty `href` to `<button>` and cleans up the markup a tiny little bit removing a wrapping `<span>` element since it's not needed nor valid to wrap a `<button>` inside it. The styling handled by the `<span>` has been attached to the `<button>` instead.


nul800sebastiaan:
Lovely, thanks @BatJan!

PR #8781, opened 2020-09-01
Introduce interfaces to allow customisation of NuCache

nzdev:
### Prerequisites

### Description

What?
Allow replacement of BPlusTree implementation, Serializers for BPlusTree, ContentData.

Why?
Out of the box works well but can be optimised further in specific scenarios using alternative serialization libraries and techniques.

How:
Introduce ITransactableDictionary interface for BPlusTree.
Introduce ITransactableDictionaryFactory interface for creating ITransactableDictionary.
Inject ISerializer<IContentNodeKit> to BPlussTreeTransactableDictionaryFactory to allow full control over serialization to and from BPlusTree.
Introduce IContentData for ContentData.

Testing:
Replace any of the interfaces mentioned above using a composer.


nzdev:
As discussed in https://github.com/umbraco/Umbraco-CMS/pull/8365 with @Shazwazza

src/Umbraco.Web/PublishedCache/NuCache/ITransactableDictionary.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Umbraco.Web.PublishedCache.NuCache
+{
+    public interface ITransactableDictionary<TKey, TValue> :
+        ITransactable,
+        CSharpTest.Net.Collections.IConcurrentDictionary<TKey, TValue>,
```
Shazwazza, line 9: I don't think we want to use an abstraction that is part of a library that isn't part of our Core. Ideally these interfaces will move to the Core project. Can we create our own abstraction if that's necessary?

Shazwazza (changes requested):
I'll need to dive deeper into this tomorrow or next week but i think it mostly makes sense. I've left one comment in there about the dictionary abstraction. Also with the latest commits to the nucache-perf branch there's some conflicts if you could fix those up when you have a chance? Cheers! :)

Shazwazza:
@nzdev oops, i think the latest commit(s) have pushed all netcore changes to this branch? Or something along those lines, there's a huge amount of changes now

nzdev:
Replacing with https://github.com/umbraco/Umbraco-CMS/pull/8978 due to merge mess

Issue #8723, opened 2020-08-25
V8.7RC: Can't create content if a content template of that document type exists


nikolajlauridsen:
If I create a content template of a given document type I can no longer create content of that type, you can however still create content of other types that has no content templates. 

It does not matter if I create the template from the context menu in the content section or through the settings section. If you delete all content templates of the given type you can create content of that type again. 

Honnestly not 100% sure how content templates works, but that does not seem intentional.

![contentTemplate](https://user-images.githubusercontent.com/16456704/91155427-20b58680-e6c3-11ea-9868-b40c00227c3c.gif)

---
_This item has been added to our backlog [AB#7922](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7922)_

nul800sebastiaan:
Was fixed in https://github.com/umbraco/Umbraco-CMS/pull/8783

PR #8783, opened 2020-09-01
Should not call hideDialog

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/8723

There is no need to call hideDialog, as the two options that need to hide the menu does that already. The third option is a shift of UI, so it should not hide the menu.

This PR removes that method call and thereby solving the above issue.


nul800sebastiaan:
Great, cheers!

Issue #8676, opened 2020-08-20
V8.7RC: Getting logged out when clicking on the tripple dots for a content item, if no doc types are allowed to be created and you don't have access to settings


ksuvec:
### Description
When you want to create a content item under a current content item, you click to 3 little dots that appears when you have your content item. If there are no content items, a box will appear with a button that says 'Edit permissions for this document type` (see img)
![image](https://user-images.githubusercontent.com/28726487/90769100-fb560080-e2ef-11ea-9953-87f312126591.png)
This is all fine, but when you don't have access to Settings duo to your user group settings, clicking the 3 buttons will just log you out (umbraco/backoffice/api/contenttype - 401 unauthorized). I assume this is because Umbraco asks for the specific doc type, which you don't have access to.

---
_This item has been added to our backlog [AB#7881](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7881)_

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8785

PR #8785, opened 2020-09-01
V8.7RC Only request getCount() if the current user has access to settings.

nielslyngsoe:
fixes https://github.com/umbraco/Umbraco-CMS/issues/8676

Only request getCount() if the current user has access to settings.


nul800sebastiaan:
Works like a charm, cheers! 👍

Issue #8709, opened 2020-08-21
V8.7RC - Removing an available Block from the Block List editor breaks the editor if content has been entered for the removed block


marcemarc:
If you've configured the BlockList editor to use several different kinds of blocks, and content items have been created for those items - and subsequently a block is removed - eg. you no longer want editors to use that block in this list - then at the point of removal, you get a warning that data based on this block 'will be deleted' - if you accept this warning and continue. Revisiting a content page that had content entered for the block type that was removed, then the BlockList Editor does not load, you just get the three 'loading dots'...

(the content hasn't been deleted as, if you add the block type back to the available list the content reappears).

![image](https://user-images.githubusercontent.com/260802/90930171-fda08380-e3f1-11ea-955b-9b9e0615e336.png)

The console logs that the block is not initialized because it's content type is not allowed for this Property Editor, so could we handle this event better in the UI?

Could we allow the editor to still function showing 'other blocks' based on other Block Types in the list? and allowing new Block Types to be added - when the page is subsequently saved the 'old ghost block content' is silently removed?

Another option would be to show some kind of message in place of the block explaining there is content here but the block type has been removed... but whilst that might save accidental removal scenarios - maybe the silent removal is more in keeping, given that the person who removed it accepted their content would be deleted! and there is rollback functionality...

But I think the main thing is the editor should still be functional with the new configuration, if a block option is removed.

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/8786

nul800sebastiaan:
Merged #8786 now 👍

PR #8786, opened 2020-09-01
V8.7RC Show Blocks that does not have a BlockConfiguration as Unsupported Bl…

nielslyngsoe:
fixes https://github.com/umbraco/Umbraco-CMS/issues/8709

Reworked the impact of removing a block configuration. Since we can't remove the data of a block we show it as unsupported and tells the developer that this is the result of removing a block configuration.

AB: https://dev.azure.com/umbraco/D-Team%20Tracker/_workitems/edit/8000


nul800sebastiaan (changes requested):
Much better, I've commented out the link for now as we'll need to have somewhere to actually link to. We can add it back once we have some docs for it.

Unfortunately I can't actually delete the unsupported block.




The code which it points to is:

````
            function deleteBlock(block) {
                var layoutIndex = vm.layout.findIndex(function(entry) {
                    return entry.contentUdi === block.content.udi;
                });
```

Can you please see if that's fixable? 🙂

nielslyngsoe:
@nul800sebastiaan I made a fix to this one.

nul800sebastiaan:
All good in my testing now, did some testing of rollbacks as well and everything seems to work correctly this time around! 👍

PR #8791, opened 2020-09-01
Make the page name label visible at all times

lordscarlet:
### Prerequisites

### Description

I have no idea if this is controversial. This was clearly deliberately made not visible to users not using a screen reader (see: `sr-only` class). However, for 508 Compliance in the United States, the guidelines specifically state "The label or instruction must be visible when the form field has focus". https://www.dhs.gov/sites/default/files/publications/trusted_tester_test_process_v5_0_aug_16_2019.pdf#page=29

image

Before:

image

After: 

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-content-header.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-content-header.html
@@ -17,7 +17,7 @@
                         {{a11yMessage}}
                     </p>
                 </div>
-                <label for="headerName" class="control-label">{{a11yName}}</label>
+                <label for="headerName" class="control-label">{{a11yName}} <span><strong class="umb-control-required">*</strong></span></label>
                 <div class="umb-editor-header__name-wrapper">
                     <ng-form name="headerNameForm">
                         <input data-element="editor-name-field"
--- a/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-content-header.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-content-header.html
@@ -17,8 +17,8 @@
                         {{a11yMessage}}
                     </p>
                 </div>
+                <label for="headerName" class="control-label">{{a11yName}}</label>
                 <div class="umb-editor-header__name-wrapper">
-                    <label for="headerName" class="sr-only">{{a11yName}}</label>
                     <ng-form name="headerNameForm">
                         <input data-element="editor-name-field"
                                type="text"


nul800sebastiaan:
Ha! Well this is an interesting one, how forceful is this requirement? Yes it will be a bit controversial, especially since screen readers will be able to tell you exactly what this field is for. I'm just trying to figure out what the use case is regarding accessibility for showing the label? 
It's certainly a conscious decision from our end not to show it as it should be pretty clear what this field is for, but maybe I'm missing something 😅

lordscarlet:
I am not the 508 compliance expert on our team, but.. It's serious enough to cause some to say Umbraco cannot be used inside the US Federal Government. The guidelines I linked are the standard for determining 508 compliance, which is a federal law and can cause agencies to be sued if they do not meet them.

lordscarlet:
..one thing to be clear on is that 508 compliance (and accessibility) is not just about the blind. It can also be those with limited sight, color blindness, limited motor skills, and other categories that are not specifically those that require screen readers. A lot of what we are being hit with in our reviews are related to this other category of users that do not use a screen reader but may not be able to use a mouse, or have a very low pixel resolution, or other limitations. I will see if I can get our certified tester to take a look at this thread and give me a better explanation.

nathanwoulfe:
This is an interesting one... Why wouldn't ARIA labelling be sufficient? As a contrived example, I ran the ANDI tool against the Google search page - the search field is not visibly labelled, but does use appropriate ARIA labelling, and passes the ANDI check, but would fail the requirements in the first screenshot.

lordscarlet:
> This is an interesting one... Why wouldn't ARIA labelling be sufficient? As a contrived example, I ran the ANDI tool against the Google search page - the search field is not visibly labelled, but does use appropriate ARIA labelling, and passes the ANDI check, but would fail the requirements in the first screenshot.

Definitely an interesting case and one I have been struggling with. Luckily, we want to user Plumber as well, so we will be looking at it with a fine-toothed comb as well 🤣

jaim0419:
Hi there! 508 Compliance expert of the team here. Just wanted to give a little additional context that may be helpful. As the author stated above, accessibility isn't just for those who use a screen reader, but also those who have mobility issues, visual issues that don't require a screen reader, are deaf or hard of hearing, and those with cognitive disabilities, such as memory issues, language, or learning disabilities. While it may be clear to a user with no cognitive issues what a field with no label at the top of a page is for, that may be quite difficult for someone with a memory issue or a learning disability. We help them out, by ensuring that visual labels are present for all form fields. The tie to WCAG 2.0, see notes at the bottom: https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html

In the example of Google, there's a little bit of a gray area here (isn't there always), but technically this is alright as the magnifying glass serves as an accepted, visual cue of what this form field does. Now one can argue that it could be better and there would be no harm in keeping the text visible in the field when focus is on it, but technically it does pass. Also, ANDI is a great tool, and much better than the previous set, however several tests are still done manually, either visually, with a keyboard, or by looking at the code.

Hope this helps!

nathanwoulfe:
> > This is an interesting one... Why wouldn't ARIA labelling be sufficient? As a contrived example, I ran the ANDI tool against the Google search page - the search field is not visibly labelled, but does use appropriate ARIA labelling, and passes the ANDI check, but would fail the requirements in the first screenshot.
> 
> Definitely an interesting case and one I have been struggling with. Luckily, we want to user Plumber as well, so we will be looking at it with a fine-toothed comb too 🤣

Should have kept my head down...

Absolutely appreciate any feedback though!

Issue #8787, opened 2020-09-01
Splitview editing with a radiobuttonlist does not work


JeffreyPerplex:
I have Umbraco 8.7RC and the default starter kit installed. The only things I added is that 

- I have added the language Dutch to the set of languages
- set the "Allow vary by culture" flag on the homepage to "True"
- set some properties on this document type to "Allow vary by culture" (The property "Hero background" (media picker - mandatory), "Font" (radiobuttonlist - mandatory), "Color theme" (radiobuttonlist - mandatory)

What I then do:

- I go to the homepage and edit the content in splitview (left English, right Dutch) 
- I click on the right side on the radiobuttonlist for "Font" and pooffff.... The left side selected radiobutton disappears

See this .gif:

![splitviewradiobuttonlist](https://user-images.githubusercontent.com/12457148/91856404-d5711a00-ec66-11ea-8fc4-b9205e0eac6c.gif)

What I expected:

- The original value on the left side stayed selected

nul800sebastiaan:
I can also reproduce in 8.5.5. Looks like it thinks the options are part of the same list.. strange!

umbrabot:
Hi @JeffreyPerplex,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
Could it be because the form wrapping radiobuttons property editor has the name `radioButtonsFieldForm`, so in splitview they might conflict with each other because they have same group name.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/radiobuttons/radiobuttons.html#L2

It could add something like this `radioButtonsFieldForm_{{vm.uniqueId}}` and in controller has something like this:

```
function init() {
     vm.uniqueId = String.CreateGuid();
}
```

but it could also be related to the name attribute on the radiobutton which in this case is `font` by not unique per culture.

![image](https://user-images.githubusercontent.com/2919859/91903540-acbb4580-eca3-11ea-9113-6e2110332968.png)

[...]

PR #8792, opened 2020-09-01
Create more unique id for radiobutton list when using splitview

bjarnef:
Create more unique id for radiobutton list when using splitview and where two properties on each culture has same alias

### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8787

### Description
When using splitview and the property using radiobuttons property editor, it is shown for both cultures and with same value of radiobutton `name` attribute, because the property alias is the same.

However this also cause the two radiobutton lists being seen as same group, so when changing selected option it remove the selected option in the another culture.

Note it is both a issue when the property vary by culture and don't vary by culture.
It could most likely in this case also use value from `mculture` and `cculture` querystrings, but it seems a bit safer to use `String.CreateGuid()` to create a unique identifier when the radiobutton list is rendered.

The value in `name` attribute on `<ng-form>` should probably also be more unique, but it doesn't seem to be crucial now that the value in `ng-messages` later in view is more unique. With an empty radiobutton list property and mandatory+ vary by culture, it seems to show the validation at the correct culture. We can always look at this in another PR if it turns out to be a problem: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/radiobuttons/radiobuttons.html#L2

Should probably be cherry picked to v8.7 as well 👍

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/radiobuttons/radiobuttons.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/radiobuttons/radiobuttons.controller.js
@@ -7,6 +7,8 @@ angular.module("umbraco").controller("Umbraco.PropertyEditors.RadioButtonsContro
 
         function init() {
 
+            vm.uniqueId = String.CreateGuid();
+
             //we can't really do anything if the config isn't an object
             if (Utilities.isObject($scope.model.config.items)) {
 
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Looks good and works like a charm, thanks @bjarnef!

PR #8793, opened 2020-09-02
V8.7RC Block Editor: Use correct label for custom view headline

kjac:
### Prerequisites

### Description

It looks like the wrong label is being used in the "Select view" dialog when selecting custom block editor views:

image

Perhaps `headlineSelectView` was renamed to `headlineAddCustomView` along the way to ensure consistency... at least that makes sense in my head 😆 `headlineAddCustomView` seems to be a better fit for the dialog title:

image


nul800sebastiaan:
Thanks for the help, this one was already fixed! 👍

PR #8794, opened 2020-09-02
V8.7RC Block Editor: Fix JS validation error for missing settings block

kjac:
### Prerequisites

### Description

If you configure a block in Block Editor without a settings model, the editor throws a nasty JS error:

image

It happens when the editor displays server side validation errors. Obviously there can be no server side validation of the settings model if there is none, but Block Editor assumes there is one - and then things fail.

This PR ensures that we only show server side validation errors for the settings model if there indeed is one configured for the block.


nul800sebastiaan:
Hmmm, try as I might, I can't reproduce this! It could be that this was already fixed and today's merge brought those fixes over. Could you try again on the latest and see if you can repro @kjac ?

kjac:
@nul800sebastiaan after merging the 8.7 branch into v8/contrib the error is gone; seems the whole validation thing has had a redo in 8.7 😐

nul800sebastiaan:
It sure did! Thanks for verifying, sorry to waste your time! 🙏

PR #8795, opened 2020-09-02
V8.7RC Block Editor: Cannot reuse existing element types

kjac:
### Prerequisites

### Description

It seems we're quite unable to use existing element types in block editor:

image

Turns out a JS error is making a mess of things:

image

This PR fixes the JS error, thus bringing back selection of existing element types:

be-reuse-element


nul800sebastiaan:
I can't reproduce this one any more on the current codebase, all looks good! I'll close this one!

PR #8797, opened 2020-09-02
V8.7RC Block Editor: Cannot select an existing element type as settings model

kjac:
### Prerequisites

### Description

Somewhat similar to #8795 but the source of the error is different... as an unintended side effect of #8552, the `compareArrays` filter throws an error if the array being compared to is not defined. Previously this was safeguarded by AngularJS (when AngularJS was used to perform the array iteration) .

image

It becomes quite evident in the Block Editor when selecting a model for block settings - you're unable to pick any existing element types for the settings model:

be-reuse-element-settings-before

This PR amends the `compareArrays` filter, and now the element settings picker works as you'd expect:

be-reuse-element-settings-after


nul800sebastiaan:
Thanks a bunch @kjac! 👍

nul800sebastiaan:
Cherry picked for 8.7 in 31e69ad1e2b13a78b05219af5b3d173df7c1c6c0

PR #8798, opened 2020-09-02
Fixes integration tests bootstrapping

Shazwazza:
* Adds IUmbracoBuilder to build the config so it's easier to split apart plus this is a much nicer pattern for us to boot/configure and allows us to potentially supply additional options for each method
  * This pattern will be expanded on with the MSDI changes and is more or less what James was proposing but will be further tweaked. He seemed happy with this approach so far and it does make it easier to boot things up so you don't have to know every single ext method needed to boot umbraco. 
* ensures background tasks don't run in integration tests
* uses a custom CoreRuntime to ensure we use SimpleMainDom
* re-wires the WebApplicationFactory so that duplicate Hosts are not running at the same time which was the biggest problem. This class automatically will use the supplied <T> parameter as the "Startup" class and we were giving it the real Startup.cs class which is not configurable to be run as part of tests. So we were booting our own Generic Host in the base class and then booting another host with the default usage of WebApplicationFactory. Instead we are just in charge of our own Generic Host for maximum flexibility, we don't use UseStartup and let it discover methods, we manage this ourselves. 
* fixes issues with overriding/shadowing which mean base class logic wasn't being called or sub class logic wasn't entirely overriding either in the base class integration tests
* uses RAMDirectory in lucene for tests
* fixes CoreRuntime's events that we're moved and changed incorrectly, Deploy compat would be broken with that as it was not doing what it was meant to do
* adds new real non static events which we can use in integration tests to install the DB before the runtime state is calculated
* adds custom integration test composers/components to modify services that are not compatible with tests 
* Fixes IRuntime to be correctly DI and exposes the missing method on the interface
* Custom ILocalizedTestService registration for integration tests to load the lang xml files from their correct file system location
* Ensures all static events are unregistered on terminate ... SO important (have backported this to 8.6)
* Ensures the IRuntime is Terminated in integration tests


src/Umbraco.Infrastructure/PropertyEditors/PropertyEditorsComponent.cs
```diff
@@ -27,32 +32,48 @@ public void Initialize()
         }
 
         public void Terminate()
-        { }
+        {
+            foreach (var t in _terminate) t();
```
bergmania, line 36: I like this pattern, even that it is not perfect. IMO it would be better to have a method taking both subscribe and unsubscribe actions, to ensure you always unsubscribe also.

We could then handle the `foreach` in a base class for both a `_terminateActions` and a `_initializeActions`

bergmania (approved):
Looks good.. Only small suggestions that can be updated later if you agree

PR #8800, opened 2020-09-02
Switch from the Chrome runner to JSDom

nul800sebastiaan:
### TL;DR; 

We shouldn't rely on software being installed on each machine to make the frontend build succeed. This removes the dependency on Chrome.

### The Long story

Recently the test runner for the frontend tests was switched out to use the Chrome browser: 
https://github.com/umbraco/Umbraco-CMS/commit/378118236345b7d3008935d43319d1be755320fa

This is great... except when you don't have Chrome installed. Running the `build.ps1` script doesn't work when Chrome is not installed and `npm run build`/`npm run test` will also result in errors.

image

It's only a small population of course that have this problem. So I re-installed Chrome and `npm run build`/`npm run test` works again. Hurray! 

However `build.ps1` is still failing, I presume it's because we reset the path so powershell has no idea where to find Chrome.exe

I tried adding the default path to our build scripts but unfortunately that didn't works. 

Then I considered switching back to `PhantomJS` like we had before, but this one is discontinued and doesn't understand some of our modern syntax now. 

Finally, I found `JSDom`, which is perfect: we don't require anything installed on the local machine (no Chrome needed) and JSDom is a package we can get from NPM.

`npm run build`/`npm run test` ran well, time to party! 🎉 

Oh wait.. `build.ps1` is failing now.. whaaaaat..? 😢 

Turns out out build module uses Node version 10, in that version `The fs.promises API is experimental`. This is registered as an error in powershell, failing the build. We could update to Node version 12, but... the build module for Umbraco is used for other projects as well, I definitely don't want to do that without a lot of testing.

The solution is to ignore that single error in the powershell script. I tested by changing `expect(scope.selectContentType).toBe(true);` to `expect(scope.selectContentType).toBe(false);` in the content controller spec and... hurray! The build fails. Correcting the spec made the build succeeed and we've proven that the `fs.promises` error is the only one that gets filtered out.


nul800sebastiaan:
Ps. Note that I've also removed the dependencies for PhantomJS, they were no longer used.

nul800sebastiaan:
Pps. Note that I'm not just doing this because it didn't work on my machine, I know I am very contrarian in not having Chrome installed. The biggest problem is that we can't rely on all contributors having Chrome installed and if we can't fix the Chrome path problem in `build.ps1` then it's going to be very difficult for new contributors to even get started.

nielslyngsoe (approved):
Works

PR #8803, opened 2020-09-02
Convert a to button

BatJan:
### Prerequisites

### Description
I have changed the `<a>` to `<button>` since it had an empty `href` attribute. I have also attached the click event to the `<button>` element instead of on the `<li>` element making the tabbing behave better since the `<li>` will no more be band-aided by the ngAria component 😃


nathanwoulfe (approved):
👍

nathanwoulfe:
Thanks @BatJan must be getting close to all of these changed?

BatJan:
@nathanwoulfe Yes very close now. One or two more to go 👍 - PR for one of them incoming 😉

PR #8806, opened 2020-09-03
Use correct woff and woff2 MIME types

ronaldbarendse:
### Prerequisites

### Description
The MIME types for `.woff` and `.woff2` file extensions are updated by Umbraco and set to outdated/incorrect values:

https://github.com/umbraco/Umbraco-CMS/blob/44fbc29541604973df5311af9898fc4ceaad583d/src/Umbraco.Web.UI/web.Template.config#L196-L197

https://github.com/umbraco/Umbraco-CMS/blob/3f5729767d362c86011865845cebea28f1ea5c92/src/Umbraco.Web.UI/web.Template.config#L320-L321

The correct MIME types are `font/woff` and `font/woff2` (see [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) and the [official IANA list](http://www.iana.org/assignments/media-types/media-types.xhtml#font)). This PR makes sure these are set in the default `web.Template.config` (new installs) and updated in `Web.config.install.xdt` (when upgrading via NuGet).

To test this PR, make sure that on new installs and after upgrading using NuGet, the `Content-Type` response header (visible in Developers Tools\Network) of requests like `/umbraco/assets/fonts/lato/LatoLatin-Regular.woff` and `/umbraco/assets/fonts/lato/LatoLatin-Regular.woff2` return the correct MIME types.

### DIFF
--- a/build/NuSpecs/tools/Web.config.install.xdt
+++ b/build/NuSpecs/tools/Web.config.install.xdt
@@ -84,9 +84,9 @@
       <remove fileExtension=".svg" xdt:Locator="Match(fileExtension)" xdt:Transform="InsertIfMissing" />
       <mimeMap fileExtension=".svg" mimeType="image/svg+xml" xdt:Locator="Match(fileExtension)" xdt:Transform="InsertIfMissing" />
       <remove fileExtension=".woff" xdt:Locator="Match(fileExtension)" xdt:Transform="InsertIfMissing" />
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Just had to have a look as of when the woff type changed, 2017 seems like a long enough time ago to update this indeed! 👍 Thanks!

PR #8807, opened 2020-09-03
AC7976 dont place runtime vars in model

nielslyngsoe:
We should not place vars use internally by property-editor in the model. The model will be replaced with a model from server on save, so we will loose these vars. They can be maintained by `model.onValueChanged` but there is no need for this. Instead we just keep them in $scope.

This issue was seen on TextArea and TextBox, Which this PR fixes.

Previously validation errors would appear on model after model has been returned from server, as the maxChar variable was stored on model and therefor lost.

Screenshot of issue appearing in TextBox:
image


nul800sebastiaan:
Noticed the configuration still said 500, changed that as well, other then that this is good to go! 👍 

image

PR #8808, opened 2020-09-03
Update config comments and fix transforms

ronaldbarendse:
### Prerequisites

### Description
While fixing the WOFF/WOFF2 MIME types (PR https://github.com/umbraco/Umbraco-CMS/pull/8806), I noticed some outdated links in the config comments and an invalid character in the config transform:
https://github.com/umbraco/Umbraco-CMS/blob/54642590cca333fd20d1ee4f8fd9e263c596f5ff/build/NuSpecs/tools/Web.config.install.xdt#L49

I've also ensured the config transforms are not nested, since that results in wrong transforms getting applied. In the following example, it's not clearly defined whether the `urlCompression` attributes should only be set if the `system.webServer` element is not missing:
```xml
<system.webServer xdt:Transform="InsertIfMissing">
    <urlCompression doStaticCompression="false" doDynamicCompression="false" dynamicCompressionBeforeCache="false" xdt:Transform="SetAttributes(doStaticCompression,doDynamicCompression,dynamicCompressionBeforeCache)" />
</system.webServer>
```

This is easily fixed by seperating the transforms:
```xml
<system.webServer xdt:Transform="InsertIfMissing" />
<system.webServer>
    <urlCompression doStaticCompression="false" doDynamicCompression="false" dynamicCompressionBeforeCache="false" xdt:Transform="SetAttributes(doStaticCompression,doDynamicCompression,dynamicCompressionBeforeCache)" />
</system.webServer>
```

To test this, make sure new installs get the changes from the `web.Template.config` and NuGet upgrades still do the right transforms.


nul800sebastiaan:
Great, good update, thanks!! 👍

Issue #8801, opened 2020-09-02
Using `clipboardService.copy` only for other things than Element Types


leekelleher:
Is the `clipboardService.copy` function designed to only be used with Element Type schema data?

Can it be used for any JSON data structure?

**Back story**

For my [Contentment package](https://github.com/leekelleher/umbraco-contentment/), I'd developed my own block editor against earlier versions, (8.4, 8.5, 8.6). One of my features was to [copy block items using the `clipboardService.copy`](https://github.com/leekelleher/umbraco-contentment/blob/dev/v1.1/src/Umbraco.Community.Contentment/DataEditors/ContentBlocks/content-blocks.js#L202) - which worked great.  Prior to my first beta release, I'd tested against v8.6.1 and the copy feature was good.

I've noticed (whilst reviewing the v8.7-RC) that since v8.6.2 the `clipboardService.copy` now assumes that the JSON data has a `variants/tabs/properties` structure, (e.g. that of a Document/Element Type schema). _(As far as I can tell, added in PR #7133)_  This works well for Nested Content (and new Block List), but if the JSON data being copied doesn't fit the same structure, then an error is thrown.

> Here's added line of code in the `prepareEntryForStorage ` function, that explicitly assumes for `variants/tabs/properties` structure: <https://github.com/umbraco/Umbraco-CMS/blob/release-8.6.2/src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js#L76>

The problem I have on my Contentment package is that my JSON data structure doesn't follow the same structure as Nested Content/Block List. I've removed the `variants/tabs/properties` structure, storing a reduced set of property data.

If were are able to copy any JSON structure with the `clipboardService`, I had plans for using it on other property-editors in Contentment.  But if it's been designed to only work with Document/Element Type schema, then I'll park my ideas. 😕 

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.2, 8.6.3, 8.6.4, 8.7.0-RC

Reproduction
--------------

_I've snipped out the reproduction steps, as it's more a question of intent._

---
_This item has been added to our backlog [AB#8090](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8090)_

nielslyngsoe:
[...]

PR #8809, opened 2020-09-03
V8.7RC: Clipboard resolvers should target specific entry types

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/8801

Mistakenly we have forgotten to make resolvers specific to entry types. Making it impossible to use the clipboard service for other types than ElementTypes. This is an error that will be corrected by this PR.

-- Breaking Change --
This PR is slightly breaking as copy and pasting elementTypeArray has changed.

In general, this PR requires resolvers to provide the entry-type that it will be used for, this simplifies writing resolvers as you can define the type it should be used for.

Due to this, copying an array of elementTypes should be using the entry-type ElementType, as we will start using this type for which resolvers that match. Meaning that array and single entries will have the same type. and code for pasting should be checking whether data is an array or a single object.

Test notes:
We need to test both our implementations, Block List Editor(with inner blocks 3 levels) and Nested Content with inner Nested Content. The important part is checking wether the Resolvers does their job of replacing UDIs(BL) or removing keys(NC)
Please have a bigger picture in mind when testing this, do you think we have solved the needs for package devs well enough?

---
_This item has been added to our backlog [AB#8093](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8093)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/blockeditor.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/blockeditor.service.js
@@ -49,6 +49,7 @@
                 } else if(k === "settingsUdi") {
                     replaceUdi(obj, k, propValue.settingsData);
                 } else {
+                    // lets crawl through all properties of layout to make sure get captured all `contentUdi` and `settingsUdi` properties.
                     var propType = typeof obj[k];
                     if(propType === "object" || propType === "array") {
                         replaceUdisOfObject(obj[k], propValue)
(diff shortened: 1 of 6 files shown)


nul800sebastiaan (changes requested):
Much better, I've commented out the link for now as we'll need to have somewhere to actually link to. We can add it back once we have some docs for it.

Unfortunately I can't actually delete the unsupported block.




The code which it points to is:

````
            function deleteBlock(block) {
                var layoutIndex = vm.layout.findIndex(function(entry) {
                    return entry.contentUdi === block.content.udi;
                });
```

Can you please see if that's fixable? 🙂

nul800sebastiaan:
whoops, commented on the wrong PR!

warrenbuckley:
Code explained to me over a call with @nielslyngsoe 
Seems OK to me now I have a bit of context to understand it

✔️ Tested with block list inside block list copying & pasting items and all UDIs are still unique

warrenbuckley:
✔️  Same with Nested Content inside Nested Content items copied & pasted correctly and still have correct unique keys for each item after node save

PR #8821, opened 2020-09-04
Fix the Nested Content "add content" dialog handling of the clipboard

kjac:
### Prerequisites

### Description

The refactoring of the Nested Content "add content" dialog in #8044 has left the handling of the clipboard somewhat broken.

First of all the header ("Add content") should be hidden if the clipboard contains any applicable items, as the other dialog headers ("Paste from clipboard" and "Create new") take its place:

nc-clipboard-01

Second when you attempt to clear the clipboard you'll see an error in the JS console (even if the clipboard is actually cleared):

nc-clipboard-02

This PR fixes the clipboard handling - the JS error is removed and the header is hidden when the clipboard contains items:

image


nul800sebastiaan:
Good stuff, thanks again @kjac - works like a charm now!

PR #8822, opened 2020-09-04
Adjust tags editor typeahead to fit within container in splitview

bjarnef:
### Prerequisites

### Description
When opening splitview the tags editor overflow its container a bit which in causing horizontal scroll.

I have adjusted the styling, so the input fit within its container.

**Before**

**After**


nul800sebastiaan:
Sweet, nice one @bjarnef, many thanks! 👍

PR #8824, opened 2020-09-04
Different user flow when creating a new blocks

nielslyngsoe:
When user creates a new block, not in inline-editing mode, the editor overlay will appear after block type selection. If user wants to cancel at this point, we want to cancel the whole creation and not just the editing of its properties.

This PR changes the buttons when in creation flow, showing "cancel" and "create". If cancel the user will be asked if they are sure about discarding this creation.

As seen in this screenshot:
image


nul800sebastiaan (changes requested):
The only problem is that it doesn't remove the new block.. see gif!

nielslyngsoe:
@nul800sebastiaan could you make sure you have rebuild your thing, the Gif clearly shows a different scenario than what I have. Note the text on the buttons, it should say "cancel" and "create"

block_editor_new_create_flow

nul800sebastiaan:
Whoops, totally right, I never switched to the PR branch it looks like, my bad! 

All good, works great! 👍

PR #8826, opened 2020-09-04
Associate existing label to the button for a umb-toggle-group item (#8716)

### LINKED ISSUE #8716: umb-toggle-group does not have an accessible name
The `umb-toggle-group` directive does not have a `<label>` associating the `item.name` with the associated `button` (generated by `umb-toggle`).

## Umbraco version

I am seeing this issue on Umbraco version:  8.6.1

Reproduction
------------

Navigate to `/umbraco/#/users/users/group/-1?create=true`

Under _Default permissions_ investigate the toggles and their associated names.

### Expected result

The toggle should have an associated name. Because the UI presents one, populated by `item.name`, there should be a `<label>` around that name with a `for=` attribute indicated the id of the button that generates the toggle.

### Actual result

There is no association between the two elements

lordscarlet:
### Prerequisites

If there's an existing issue for this PR then this fixes #8716 

### Description

Specify `input-id` for `umb-toggle` and add a label to refer to that id.

Before:

image

After:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/buttons/umb-toggle-group.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/buttons/umb-toggle-group.html
@@ -3,11 +3,11 @@
         <umb-toggle class="umb-toggle-group-item__toggle"
                     checked="item.checked"
                     disabled="item.disabled"
-                    input-id="{{$id}}-{{$index}}"
+                    input-id="{{item.inputId}}"
                     on-click="change(item)">
         </umb-toggle>
         <div class="umb-toggle-group-item__content" ng-click="change(item)">
-            <div><label for="{{$id}}-{{$index}}">{{ item.name }}</label> </div>
+            <div><label for="{{item.inputId}}">{{ item.name }}</label> </div>
             <div class="umb-toggle-group-item__description">{{ item.description }}</div>
         </div>
     </div>
--- a/src/Umbraco.Web.UI.Client/src/views/components/buttons/umb-toggle-group.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/buttons/umb-toggle-group.html
@@ -3,10 +3,11 @@
         <umb-toggle class="umb-toggle-group-item__toggle"
                     checked="item.checked"
                     disabled="item.disabled"
+                    input-id="{{$id}}-{{$index}}"
                     on-click="change(item)">
         </umb-toggle>
         <div class="umb-toggle-group-item__content" ng-click="change(item)">
-            <div>{{ item.name }} </div>
+            <div><label for="{{$id}}-{{$index}}">{{ item.name }}</label> </div>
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web.UI.Client/src/views/components/buttons/umb-toggle-group.html
```diff
@@ -3,10 +3,11 @@
         <umb-toggle class="umb-toggle-group-item__toggle"
                     checked="item.checked"
                     disabled="item.disabled"
+                    input-id="{{$id}}-{{$index}}"
```
bjarnef, line 6: Maybe something like this would be better?

```
input-id="{{item.inputId}}"
```

and then use `String.CreateGuid()` in controller.

It would be a bit more consistent with this:
https://github.com/umbraco/Umbraco-CMS/commit/124415e413a7f716b220a126513d1b967da73cff

and e.g. `<umb-checkbox>`
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js#L52

The `<umb-toggle>` set an id here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbtoggle.directive.js#L78

But in these changes it overwrite the id.

It could probably extend each item in the directive to set something like this:
```
item.inputId = "umb-toggle-group-item_ + String.CreateGuid();
```

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbtogglegroup.directive.js#L73-L86

lordscarlet: Thank you, @bjarnef. I planned to do this, but I spent an inordinate amount of time because I am not familiar with AngularJS conventions, and I did not know that `inputId` is referenced as `input-id` in the view. Anyway, I made an update based on your feedback. If anything else could be improved, please let me know.

nul800sebastiaan:
Apologies about the delay on this one! That is lovely work @lordscarlet and thanks for the assist @bjarnef 👍  (sorry about AngularJS.. it has it's moments but.. 🙈 ).

Congrats on your first merged PR for Umbraco-CMS, let's hope there's many more to come 🎉 

You've just earned your contrib badge on Our, as well, woohoo! 🎖️  https://our.umbraco.com/members/id:37850/

lordscarlet:
Thank you and not a problem at all! I would have a lot more submitted already, but we got into too much of a rush and had to just apply a bunch of stuff locally and quickly. Hopefully over the next month or two we will find time to backtrack all of those as Prs.

nathanwoulfe:
Changes in #9199 fix the regression 👍

PR #8827, opened 2020-09-04
Show only settings and styles boxes when each of them has configuration

bjarnef:
### Prerequisites

### Description
When there is no grid configuration for settings and styles two empty boxes are shown for these. It would be nicer to show an empty message here.
Furtheremore if you only use e.g. styles there is no need to use valuable space to show an empty settings box. So instead we can just show each box it any items are available. I have also changes, so the submit button only is visible when any configuration is available.

chrome_2020-09-04_14-28-16

image


nul800sebastiaan:
What a great idea, thank you @bjarnef!

PR #8828, opened 2020-09-04
Use umb-code-snippet in unsupported block

bjarnef:
### Prerequisites

### Description
When a block is unsupported in block list editor, we can use `<umb-code-snippet>` to show the json, which also makes it easy to copy the snippet.
It is also a bit more consistent with grid error https://github.com/umbraco/Umbraco-CMS/pull/8548 and query builder https://github.com/umbraco/Umbraco-CMS/pull/7294

For testing I has just comment out this line locally https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js#L200

**Before**

chrome_2020-09-04_14-57-02

**After**


bjarnef:
@nielslyngsoe what do you think? 😎

nielslyngsoe:
Neither of them is pretty ;-), but this is definitely an improvement :-D

bjarnef:
Not sure it needs to be pretty as long it is useful 😁
I guess it would also help with less vertical scroll if you have a lot of text in the block 👍

nielslyngsoe:
Yep, it gives a better overview. And nice that you can do a "safe" copy of the JSON 👍

nul800sebastiaan:
Great stuff @bjarnef ! 👍 

Cherry picked for 8.7.1 in d55916e1b8538dca0adbff81bc6b9615793a3d30

PR #8829, opened 2020-09-04
Block Editor: Automatically add item if there is only one item type

### LINKED ISSUE #8760: V8.7RC When only one block available, don't show the catalogue
When you try to add content to a block list, you get the catalogue dialog shown in order to select the block type.
image

When you only have one block type configured, and nothing in the clipboard, this step should be omitted, and the block should just be added IMO.

kjac:
### Prerequisites

Fixes #8760

### Description

If you have a Block Editor with only one item type, you're still prompted to pick that item type when adding new items, despite that item type being the only choice (i.e. even when there is nothing in the clipboard).

I reckon the block editor should automatically add a new item to the item list if there is only one available item type (and nothing in the clipboard):

be-auto-add-item

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbBlockListPropertyEditor.component.js
@@ -420,23 +420,7 @@
                         blockPickerModel.close();
                     }
                 },
-                submit: function(blockPickerModel, mouseEvent) {
-                    var added = false;
-                    if (blockPickerModel && blockPickerModel.selectedItem) {
-                        added = addNewBlock(createIndex, blockPickerModel.selectedItem.blockConfigModel.contentElementTypeKey);
-                    }
-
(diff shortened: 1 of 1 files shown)


abjerner:
Thanks @kjac 👍 

This seems to be what @skttl created an issue for here: https://github.com/umbraco/Umbraco-CMS/issues/8760

I've only looked through your changes here on GitHub, but as far as I can tell, this doesn't handle the clipboard scenario that Søren describes. Is this correct?

kjac:
@abjerner hah! Yeah certainly looks like it!

I'll comment a bit on the issue, but at least this is (I think) a step forward compared to what is. Also... I don't see why the item picker should *not* contain the single allowed item type if the clipboard has items, that seems really counter intuitive - what if I wanted to create a new item?

Anyway I'll throw some remarks on the issue 👍 thanks for pointing me in that direction.

nathanwoulfe:
Looks like this has been resolved in #9277 - when only one block type exists, the Add button adds the new item directly. If content exists on the clipboard, the Add button is suffixed with a clipboard button which opens the overlay to the clipboard view, which can be switch to Add empty via the header tabs.

PR #8833, opened 2020-09-05
Checkboxlist and radiobuttonlist validation

bjarnef:
### Prerequisites

### Description
Recently I fixed a issue with radiobuttonlist in splitview in this PR https://github.com/umbraco/Umbraco-CMS/pull/8792
But I then noticed that when the property was mandatory and using default or a custom validation message, this wasn't shown.

This has been fixed in https://github.com/umbraco/Umbraco-CMS/commit/f5aeca5f3dacae8016488139d38ca14dba39681d

Furthermore I noticed checkboxlist didn't show this message either and the value of `name` attribute was identical in splitview although this didn't cause issues as with radiobuttons as multiple options are allowed, but could cause issues with validation of the correct "group".

So I have now made this consistent with radiobuttonlist.

image

image


bjarnef:
@nul800sebastiaan should probably be cherry picked to v8.7 as well ☺️

nul800sebastiaan:
Good catch @bjarnef! 
Cherry picker for 8.7 in 86d105dcf383939222f3dea07699c8f609a080da

PR #8840, opened 2020-09-07
Add Partials and RenderPartials extension methods to replace GetBlockListHtml

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8727.

### Description
This PR adds extension methods to `HtmlHelper` for easy rendering of partials views from a list of items, like from the Nested Content (`IEnumerable<IPublishedElement>`) or Block List (`IEnumerable<BlockListItem>` aka `BlockListModel`) editors.

Instead of using `@Html.GetBlockListHtml(Model.BlockListItems)`, you can now use `@Html.Partials(Model.BlockListItems, "BlockList")` and it will render partials views for every block list item using the views in `~\Views\Partials\BlockList\{ContentAlias}.cshtml`. This doesn't render the `<div class="umb-block-list">` wrapper though, so to keep the exact same output, you'd have to (conditionally) add that yourself...


ronaldbarendse:
Here are some examples of how these extension methods can be used and the verbose code they replace.

-------------

For Nested Content, you manually need to loop over the elements and determine which partial view must be used to render that type:
```cshtml
@if (Model.Elements != null)
{
    foreach (var element in Model.Elements)
    {
        Html.RenderPartial("NestedContent/" + element.ContentType.Alias, element);
        @Html.Partial("NestedContent/" + element.ContentType.Alias, element)
    }
}
```

The generic version of both `RenderPartials<T>()` and `Partials<T>()` allow a function to be passed that must return the partial view name. The null-check and iteration is all done within the extension method (and if no partial view name is returned, the element is skipped):
```cshtml
@{ Html.RenderPartials(Model.Elements, element => "NestedContent/" + element.ContentType.Alias); }
@Html.Partials(Model.Elements, element => "NestedContent/" + element.ContentType.Alias)
```

For Nested Content (or any other editor that returns `IEnumerable<IPublishedElement>`, you can (optionally) provide a path that's prepended to the content type alias, so the above code can be written as:
```cshtml
@{ Html.RenderPartials(Model.Elements, "NestedContent"); }
@Html.Partials(Model.Elements, "NestedContent")
```

-------------

The same concept works for the new Block List editor; the initial 8.7 release provides a `GetBlockListHtml()` that basically does the same as:
```cshtml
@if (Model.Blocks != null)
{
    @foreach (var block in Model.Blocks)
    {
        Html.RenderPartial("BlockList/" + block.Content.ContentType.Alias, block);
        @Html.Partial("BlockList/" + block.Content.ContentType.Alias, block)
    }
}
```

The generic version of this would still be a little verbose:
```cshtml
@{ Html.RenderPartials(Model.Blocks, block => "BlockList/" + block.Content.ContentType.Alias); }
@Html.Partials(Model.Blocks, block => "BlockList/" + block.Content.ContentType.Alias)
```

So for Block List (or any other editor the returns `IEnumerable<BlockListItem>`), this can be written as:
```cshtml
@{ Html.RenderPartials(Model.Blocks, "BlockList"); }
@Html.Partials(Model.Blocks, "BlockList")
```

-------------

Optionally you can also:
- Manually change the way the partial view name is generated (e.g. remove a generic prefix `@Html.Partials(Model.Elements, e => "Home/" + e.ContentType.Alias.TrimStart("Home"));` - using the `TrimStart` from Umbraco's `StringExtensions`)
- Pass in a different `ViewDataDictionary` to the partial (the same for all or different per item/index)
- Transform the model that's passed to the partial (e.g. to wrap the item in a view model or only use the `IPublishedElement` of a `BlockListItem` for example)

NikRimington:
@ronaldbarendse I think these are cool extension methods, however, I think the order of the parameters should be the other way around if possible. The reason for my thoughts on this is that it would follow the same pattern as the MVC method `Html.Partial()`, where the path to partial comes first and then the model. Would just feel more consistent.

ronaldbarendse:
@NikRimington I totally understand where you're coming from, but having the parameters in a similar order as the `Partial`/`RenderPartial` methods would actually make it more confusing:
1. The partial view name is based on iterating the items, so it's logical to have the items as first parameter
2. The partial view path is combined with the content type, so the following renders totally different partials views:
   - `@Html.Partial("NestedContent", Model.Elements)` - renders the `NestedContent` partial view using the `IEnumerable<IPublishedElement>` as model
   - `@Html.Partials(Model.Elements, "NestedContent")` - renders different partial views within the `NestedContent` folder based on the content type alias and using the individual `IPublishedElement`s as model
3. There's an overload that can transform the model passed into the partial view (`getModel`), so the items don't always end up as the model. This parameter is still in front of the `partialViewPath` or `getPartialViewName`, as the model is also available when getting the partial view name and it would otherwise cause unambiguity with the `getViewData` parameter.

The second point is a major one, as (accidentally) adding or removing a single letter would end up as valid code, but do completely different things!

src/Umbraco.Web.UI/Umbraco.Web.UI.csproj
```diff
@@ -285,7 +285,6 @@
     <Content Include="Views\Partials\Grid\Bootstrap3-Fluid.cshtml" />
     <Content Include="Umbraco\Views\Preview\Index.cshtml" />
     <Content Include="Umbraco\Views\web.config" />
-    <Content Include="Views\Partials\BlockList\Default.cshtml" />
```
ronaldbarendse, line 288: Although this Razor view is removed, it's still left untouched within existing projects (that installed a previous Umbraco 8.7+ version) and because the `BlockListTemplateExtensions` methods are only obsoleted, rendering of the Block List items will continue to work.

ronaldbarendse:
@nul800sebastiaan Do you and/or the (newly formed 🎉) Core Contributors Team have enough info and examples to review this PR? Together with https://github.com/umbraco/Umbraco-CMS/pull/9031, this would make rendering Nested Content, Block List and Multi URL Picker/Link items a lot easier!

PR #8841, opened 2020-09-07
8.7RC Add generic BlockListItem classes

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8727.

### Description
Currently, you have to cast the content and settings from the `BlockListItem` to the correct type, before you can access the strongly typed models, e.g.:

```cshtml
@model BlockListItem
@{
   var content = Model.Content as Feature;
   var settings = Model.Settings as FeatureSettings;
}
<h1 class="@settings.CssClass">@content.FeatureName</h1>
```

This PR adds the generic classes `BlockListItem<T>` and `BlockListItem<TContent, TSettings>` that both inherit from `BlockListItem`. The PVC is updated to create instances of the most explicit, matching generic type, so that can be used in the (partial) views:

```cshtml
@model BlockListItem<Feature, FeatureSettings>
<h1 class="@Model.Settings.CssClass">@Model.Content.FeatureName</h1>
```

Also, if you don't have settings configured (or you have multiple block configurations with different setting types), you can use `BlockListItem<Feature>`. The settings will be exposed as `IPublishedElement`, just like the non-generic `BlockListItem` and you have to manually null-check/cast it.

This will make working with the Block List editor a lot easier and probably also reduce bugs in the views, as you get a `ModelBindingException` with a nice message with the model/content types (instead of ` InvalidCastException` or `NullReferenceException`s).

### DIFF
--- a/src/Umbraco.Core/Models/Blocks/ContentAndSettingsReference.cs
+++ b/src/Umbraco.Core/Models/Blocks/ContentAndSettingsReference.cs
@@ -1,5 +1,5 @@
-﻿using System.Collections.Generic;
-using System;
+﻿using System;
+using System.Collections.Generic;
 
 namespace Umbraco.Core.Models.Blocks
 {
@@ -12,26 +12,16 @@ namespace Umbraco.Core.Models.Blocks
         }
 
         public Udi ContentUdi { get; }
+
         public Udi SettingsUdi { get; }
 
-        public override bool Equals(object obj)
-        {
-            return obj is ContentAndSettingsReference reference && Equals(reference);
-        }
+        public override bool Equals(object obj) => obj is ContentAndSettingsReference reference && Equals(reference);
 
-        public bool Equals(ContentAndSettingsReference other)
-        {
-            return EqualityComparer<Udi>.Default.Equals(ContentUdi, other.ContentUdi) &&
-                   EqualityComparer<Udi>.Default.Equals(SettingsUdi, other.SettingsUdi);
-        }
+        public bool Equals(ContentAndSettingsReference other) => other != null
+            && EqualityComparer<Udi>.Default.Equals(ContentUdi, other.ContentUdi)
+            && EqualityComparer<Udi>.Default.Equals(SettingsUdi, other.SettingsUdi);
 
-        public override int GetHashCode()
-        {
-            var hashCode = 272556606;
-            hashCode = hashCode * -1521134295 + EqualityComparer<Udi>.Default.GetHashCode(ContentUdi);
-            hashCode = hashCode * -1521134295 + EqualityComparer<Udi>.Default.GetHashCode(SettingsUdi);
-            return hashCode;
-        }
+        public override int GetHashCode() => (ContentUdi, SettingsUdi).GetHashCode();
 
         public static bool operator ==(ContentAndSettingsReference left, ContentAndSettingsReference right)
         {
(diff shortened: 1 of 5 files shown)


src/Umbraco.Web/PropertyEditors/ValueConverters/BlockListPropertyValueConverter.cs
```diff
@@ -120,7 +120,9 @@ public override object ConvertIntermediateToObject(IPublishedElement owner, IPub
                             settingsData = null;
                     }
 
-                    var layoutRef = new BlockListItem(contentGuidUdi, contentData, settingGuidUdi, settingsData);
+                    var layoutType = typeof(BlockListItem<,>).MakeGenericType(contentData.GetType(), settingsData?.GetType() ?? typeof(IPublishedElement));
+                    var layoutRef = (BlockListItem)Activator.CreateInstance(layoutType, contentGuidUdi, contentData, settingGuidUdi, settingsData);
```
ronaldbarendse, line 124: Because `IPublishedModelFactory` only allows creating generic lists, this currently uses basic reflection/activation (`PublishedModelFactory` emits constructor functions and caches them). We might want to do the same here (eventually), because that might gain some performance (especially when lots of items are the same content/settings type).

src/Umbraco.Core/Models/Blocks/ContentAndSettingsReference.cs
```diff
@@ -12,26 +12,16 @@ public ContentAndSettingsReference(Udi contentUdi, Udi settingsUdi)
...
-            return obj is ContentAndSettingsReference reference && Equals(reference);
-        }
+        public override bool Equals(object obj) => obj is ContentAndSettingsReference reference && Equals(reference);
 
-        public bool Equals(ContentAndSettingsReference other)
-        {
-            return EqualityComparer<Udi>.Default.Equals(ContentUdi, other.ContentUdi) &&
-                   EqualityComparer<Udi>.Default.Equals(SettingsUdi, other.SettingsUdi);
-        }
+        public bool Equals(ContentAndSettingsReference other) => other != null
```
ronaldbarendse, line 20: This was missing a null-check 😉

ronaldbarendse:
The current implementation retrieves the content/setting types from the stored property data, so that would cause problems when changing the Block List editor configuration, like adding settings 💥 

It's better to retrieve the types from the `BlockListConfiguration`: I'll do another PR to rewrite this...

nul800sebastiaan:
Thanks very much @ronaldbarendse !
Cherry picked for 8.7 in ddc922a583e24582d9c1918667d50d91a7e5921b

ronaldbarendse:
I've done some digging and it's not so easy to get the model type of the settings, as the Block List configuration only stores the settings content type GUID and the `IPublishedModelFactory` doesn't have a way to get the model type without having an instance of `IPublishedElement` 😞

This means whenever there are no settings stored (in the block list data/property), but a settings type is configured or the settings type is changed afterwards, the generic type used for the settings will always be `IPublishedElement`. This will result in a `ModelBindingException` when the partial view uses `@model BlockListItem<Feature, FeatureSettings>`, but the actual type is `BlockListItem<Feature, IPublishedElement>`. This is not a critical issue, but one that definitely needs some fixing!

ronaldbarendse:
@nul800sebastiaan I've created a fix for the above issue, which should be included in the final 8.7 release or otherwise the next patch release: https://github.com/umbraco/Umbraco-CMS/pull/8854.

src/Umbraco.Web/PropertyEditors/ValueConverters/BlockListPropertyValueConverter.cs
ronaldbarendse: So it's possible to get the actual/CLR model type using `ModelType.For("contentTypeAlias")` and `IPublishedModelFactory`s `MapModelType(modelType)` (see PR https://github.com/umbraco/Umbraco-CMS/pull/8854), but that still doesn't do the caching...

PR #8842, opened 2020-09-08
Ensure new UDIs for blocks when copying

Shazwazza:
This is when we do a full node copy (not a JS copy). When that occurs we need to re-generate new UDIs for each block in the block list value. We do this already for Nested Content too and just like NC this code will deal with nested block editors and even nested block editors within nested other complex editors that are in a block editor.

I've come to realize however that this is still problematic and it's the same for Nested Content. If there's a block editor, or nested content within another complex editor where the root editor is not itself then the GUIDs/UDIs won't be replaced in that nested json data. I'm not too worried about that scenario for now for the 8.7 release since that scenario probably won't exist very often but it's something to consider in the future. 

All of this logic will need to be entirely deleted and/or refactored when blocks of data are stored as centralized/shareable element blocks in the database https://github.com/umbraco/rfcs/pull/24. That will have it's own challenges for what 'Copy' actually means.

For testing:

* Review the code
* Ensure all tests pass
* Test that copying a node in the back office that contains block editor data works. Verify the data stored for that property in the [umbracoPropertyData] table for the original content item against the property data stored for that property for the copied content item. The value should be the same however the UDIs/GUIDs will be changed.

### DIFF
--- a/src/Umbraco.Core/ContentExtensions.cs
+++ b/src/Umbraco.Core/ContentExtensions.cs
@@ -120,6 +120,15 @@ namespace Umbraco.Core
         }
 
         /// <summary>
+        /// Returns all properties based on the editorAlias
+        /// </summary>
+        /// <param name="content"></param>
+        /// <param name="editorAlias"></param>
+        /// <returns></returns>
+        public static IEnumerable<Property> GetPropertiesByEditor(this IContentBase content, string editorAlias)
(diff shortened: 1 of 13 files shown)


warrenbuckley:
✔️ Ensure all tests pass

warrenbuckley:
✔️  Review the code

Off to test this out manually & verify 😄

warrenbuckley:
All works as expected with test
image

PR #8850, opened 2020-09-08
Review Feature: umb-property-info-button

nielslyngsoe:
Info button, to provide more into next to property editors.

Video of demo (Internal link):
https://umbraco.slack.com/files/UFCA59C8Z/F019192RCHL/block_editor_prevalue_more_info.mov

image

**Test notes**
There is a test implementation of this in Block List PreValue Configuration UI, inside the overlay for editing a blocks configuration.

---
_This item has been added to our backlog [AB#8132](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8132)_


bjarnef:
@nielslyngsoe any reason this need its own folder? Couldn't it be placed in e.g. "buttons" folder otherwise at root level?

It just seems a bit inconsistent that we now have most less, views and directive/components in separated folder, but now start to mix views, scripts, less in their own folders, while the existing components have separated less, scripts and html.

warrenbuckley:
@bjarnef I can maybe answer for @nielslyngsoe 
I think moving forward in the future views, JS & styles should be together with a component, as someone one who works here at the HQ finding the associated view, JS or CSS file can be tricky at times. So with the living side by side, it is a lot easier to find the files you need to be editing & looking at, in my opinion.

warrenbuckley (approved):

PR #8851, opened 2020-09-08
Netcore/feature/models builder embedded migration roslyn compiler

nikolajlauridsen:
* Move model directory to Umbraco.Web.UI.NetCore/Umbraco/Models
* Add RoslynCompiler and use it to compile models
* Add UmbracoAssemblyLoadContext and use it to load and unload the compiled models
* Use IHostingEnvironment instead of IIOHelper
* Inject IPublishedValueFallback into models
* Use string extension to generate type hash instead of HashCombiner for PureLiveMode
* Remove a lot of unused stuff (AssemblyInfo, HashCombiner etc.)
* Added ModelsBuilder section to appsettings (is this right) 
* Excluded PureLive generated models from csproj.
* Fixed tests to reflect changes.


bergmania (approved):

bergmania:
@nikolajlauridsen, I changed a few things.. Most imporant is the hash generated I added seperators (newlines) between the items. So  Name="TestTest" and Description="" is different from Name="Test" and Description="Test"

Issue #8835, opened 2020-09-05
Dotnet core alpha: cannot disable url tracker


crgrieve:
## Umbraco version

I am seeing this issue on Umbraco version: Dotnet core alpha

### Bug summary

Error when choosing to "Disable URL tracker".

### Expected result

Like on v8 we should have a message saying it worked.
![image](https://user-images.githubusercontent.com/9353655/92304903-30ae4f80-ef7a-11ea-8c61-454caddc9aca.png)

### Actual result
Error saying it didn't work and to check logs. There isn't anything in logs to indicate what is wrong.
![image](https://user-images.githubusercontent.com/9353655/92304911-3efc6b80-ef7a-11ea-9115-b7a47fb34949.png)

elit0451:
Thank you for reporting this issue @crgrieve! The cause of it was due to trying to update the value of _DisableRedirectUrlTracking_ in **~/config/umbracoSettings.config** instead of also doing so in **appsettings.json**

A solution has been implemented in this [PR](https://github.com/umbraco/Umbraco-CMS/pull/8852) 🙂

PR #8852, opened 2020-09-08
NetCore: Fix for disabling/enabling url tracking

elit0451:
Issue: https://github.com/umbraco/Umbraco-CMS/issues/8835 reported on alpha001.

Testing notes:
- Navigate to the _Redirect Url Management_ dashboard in the **Content** section;
- Clicking "Disable URL tracker" and confirming will present you with a success notification stating _"**URL tracker has now been disabled.**"_ 
  - check appsettings.json ➡️ there needs to be a new property object like:
```
"Umbraco": {
    "CMS": {
      . . .
      "WebRouting": {
        "DisableRedirectUrlTracking": "true"
      }
   }
}
```
- Clicking then "Enable URL tracker" will present you once more with a success notification this time saying _"**URL tracker has now been enabled.**"_ 
  - check appsettings.json ➡️ the value of `DisableRedirectUrlTracking` should be set to `"false"`

### DIFF
--- a/src/Umbraco.Core/Configuration/IConfigManipulator.cs
+++ b/src/Umbraco.Core/Configuration/IConfigManipulator.cs
@@ -7,5 +7,6 @@ namespace Umbraco.Core.Configuration
         void RemoveConnectionString();
         void SaveConnectionString(string connectionString, string providerName);
         void SaveConfigValue(string itemPath, object value);
+        void SaveDisableRedirectUrlTracking(bool disable);
     }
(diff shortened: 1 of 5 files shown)


bergmania:
Looks really good and works as expected 💪 ..

I added a small commit, that moved the checks for this config into the events instead of not initializing the events if disabled. The reason is that we can change the config value without recycling the app in netcore, so it is better if the events are always executed,.

PR #8854, opened 2020-09-08
8.7RC Get generic BlockListItem settings model type from configuration

ronaldbarendse:
### Prerequisites

### Description
This fixes an issue in PR https://github.com/umbraco/Umbraco-CMS/pull/8841 that could result in using the wrong generic type for the settings of `BlockListItem<TContent, TSettings>`. Instead of using the type from the stored settings data, this now uses the settings type from the block configuration.

To test this, configure a block without settings (e.g. using content model `Feature`) and create/publish some blocks. The partial view rendering the block should have a model type of `BlockListItem<Feature, IPublishedElement>` (where `Model.Settings` is `null`).

Add a settings model to that block configuration (e.g. using `FeatureSettings`): this directly changes the partial view model to `BlockListItem<Feature, FeatureSettings>`, even though no settings data is stored (`Model.Settings` will still be `null`). Previously, it would still be `BlockListItem<Feature, IPublishedElement>` and throw a `ModelBindingException` if you updated the partial view model to access the settings in a strongly typed way.

This also ensures that after rolling back content or adding/changing the settings model of a block, the returned type is always exactly like configured and thus does not depend on the stored/published data.


src/Umbraco.Web/PropertyEditors/ValueConverters/BlockListPropertyValueConverter.cs
```diff
@@ -120,7 +122,18 @@ public override object ConvertIntermediateToObject(IPublishedElement owner, IPub
...
+                    {
+                        settingsType = _blockConverter.GetModelType(blockConfig.SettingsElementTypeKey.Value);
+                    }
+                    else
+                    {
+                        settingsType = typeof(IPublishedElement);
+                    }
+
+                    var layoutType = typeof(BlockListItem<,>).MakeGenericType(contentData.GetType(), settingsType);
                     var layoutRef = (BlockListItem)Activator.CreateInstance(layoutType, contentGuidUdi, contentData, settingGuidUdi, settingsData);
```
ronaldbarendse, line 137: This should use compiled lambda expressions, as that's _way_ faster than `Activator.CreateInstance()`, especially with the parameterized constructor!

If we cache on `contentTypeKey` and `blockConfig.SettingsElementTypeKey`, getting the settings type could be included in the cache, which reduces calls to `GetModelType()` 👍

Any pointers on how this is best achieved?

ronaldbarendse:
@nul800sebastiaan This PR should also get merged ASAP, because using the generic `BlockListItem<TContent, TSettings>` as (partial) view model will currently give a `ModelBindingException` for the stored blocks that don't have settings (yet).

This happens when adding settings to a block after it's already used: the PVC still returns `BlockListItem<T>` for the stored block items, so you can't update the (partial) view model to `BlockListItem<TContent, TSettings>` to access the settings in a strongly typed way.

src/Umbraco.Web/PropertyEditors/ValueConverters/BlockListPropertyValueConverter.cs
```diff
@@ -51,66 +49,70 @@ public override object ConvertIntermediateToObject(IPublishedElement owner, IPub
...
             using (_proflog.DebugDuration<BlockListPropertyValueConverter>($"ConvertPropertyToBlockList ({propertyType.DataType.Id})"))
             {
-                var configuration = propertyType.DataType.ConfigurationAs<BlockListConfiguration>();
-                var blockConfigMap = configuration.Blocks.ToDictionary(x => x.ContentElementTypeKey);
-                var validSettingElementTypes = blockConfigMap.Values.Select(x => x.SettingsElementTypeKey).Where(x => x.HasValue).Distinct().ToList();
-
-                var contentPublishedElements = new Dictionary<Guid, IPublishedElement>();
-                var settingsPublishedElements = new Dictionary<Guid, IPublishedElement>();
-
-                var layout = new List<BlockListItem>();
```
ronaldbarendse, line 61: I've moved this below the short-circuiting code to prevent doing unnecessary work.

nikolajlauridsen (approved):
Code looks good to me, and it tests out good, I was unable to break anything 😆👍

PR #8859, opened 2020-09-10
V8: stabilize cypress tests and add content with contentpicker test 

nikolajlauridsen:
I added the content with contentpicker test. 

I also did some tweaks to the existing tests, to make them more stable. 

### Content tests
I fetch the content node tree after refreshing content, this is just to wait for the content to appear after refreshing if we don't do this the tests will fail because it tries to click an item that does not exist yet, while doing so I also put in in a function since it's used in almost all test.

A very minor thing but I indented function calls to child builders, making it easier to see what belongs where

### Content rollback test
I also changed the content rollback test, we no longer expect the badges to be in a certain order since it's not deterministic in this version of the UI, I managed to get this result when doing rollback manually 
non determenistic badges
The first time it's Rollback -> Save
The other time it's Save -> Rollback.
Since the order is not deterministic we cannot assume that one badge appears before the other, the test does, however, assert that the title is actually being rolled back, so it's still a valid test. 

I also changed the selector when selecting the rollback version since this has been changed in the UI.

### Create template test
I did some tweaks to the create template test as well, instead of trying to write something in the editor to force the autosave, we just remove focus from it by focusing on something else, this triggers the autosave, after that, we wait for the autosave to complete by waiting for the template to appear in the tree view. If we don't do all this there will at best be saved a duplicate of the template, and at worst, it will error out because of the template file being in use. 
At one point I thought this was an actual bug, but I cannot reproduce it manually anymore.

### Tours tests
I added a timeout to the ```contains``` call when getting the percentages before running a test, once in a while, it took a little too long before the value got loaded into the progress tracker causing the test to fail 

### Members tests
I add an increased timeout when typing in the password. The password is very long and may exceed the default 4-second timeout for typing causing the last portion of the password to be cut off on slower systems.

### Needs to be fixed in testhelpers package
Having run the tests a multitude of times one of the tests failed once due to umbracoEditorHeaderName, this issue is related to the one in member tests, the issue occurs because the type command in
```cy.get('#headerName', { log: false }).type(typedText).should('have.value', typedText);``` 
times out before the whole name has been typed, meaning that the should assert failes, because it does not match the expected.

This is easily fixed by adding a longer timeout on the type command, but can't be done in this PR

### DIFF
--- a/src/Umbraco.Tests.AcceptanceTest/cypress/integration/Content/content.ts
+++ b/src/Umbraco.Tests.AcceptanceTest/cypress/integration/Content/content.ts
@@ -331,8 +331,8 @@ context('Content', () => {
         cy.reload();
 
         // Assert
-        cy.get('.history').find('.umb-badge').eq(0).should('contain.text', "Save");
-        cy.get('.history').find('.umb-badge').eq(1).should('contain.text', "Rollback");
+        cy.get('.history').find('.umb-badge').contains('Save').should('be.visible');
+        cy.get('.history').find('.umb-badge').contains('Rollback').should('be.visible');
         cy.get('#headerName').should('have.value', initialNodeName);
 
         // Clean up (content is automatically deleted when document types are gone)
--- a/src/Umbraco.Tests.AcceptanceTest/cypress/integration/Content/content.ts
+++ b/src/Umbraco.Tests.AcceptanceTest/cypress/integration/Content/content.ts
@@ -48,8 +48,8 @@ context('Content', () => {
                     .withContentTypeAlias(rootDocTypeAlias)
                     .withAction("saveNew")
                     .addVariant()
-                    .withName(nodeName)
-                    .withSave(true)
+                        .withName(nodeName)
+                        .withSave(true)
                     .done()
                     .build();
 
@@ -60,8 +60,8 @@ context('Content', () => {
                         .withAction("saveNew")
                         .withParent(contentNode["id"])
                         .addVariant()
-                        .withName(childNodeName)
-                        .withSave(true)
+                            .withName(childNodeName)
+                            .withSave(true)
                         .done()
                         .build();
 
@@ -72,8 +72,8 @@ context('Content', () => {
                     .withContentTypeAlias(rootDocTypeAlias)
                     .withAction("saveNew")
                     .addVariant()
-                    .withName(anotherNodeName)
-                    .withSave(true)
+                        .withName(anotherNodeName)
+                        .withSave(true)
                     .done()
                     .build();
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/Umbraco.Tests.AcceptanceTest/cypress/integration/Content/content.ts [hunk 1]
```diff
@@ -326,8 +330,8 @@ context('Content', () => {
         cy.reload();
 
         // Assert
-        cy.get('.history').find('.umb-badge').eq(0).should('contain.text', "Save");
-        cy.get('.history').find('.umb-badge').eq(1).should('contain.text', "Rollback");
+        cy.get('.history').find('.umb-badge').eq(0).should('contain.text', "Rollback");
```
elit0451, line 333: When I did the tests I have manually tried Rolling back content and on the audit trail, I was always the "Save" label before the "Rollback" one.  When doing the tests sometimes it happened that the operation was too fast and "Rollback" was coming first (differing from what I observed during the manual reproduction steps), that's why I am reloading the page before asserting. And then after running the test a few times the ordering was as expected. So I am not sure about changing it as I would think that the typical scenario would be to rollback and then save the changes. 

cc @bergmania

nikolajlauridsen: Hmm that's very strange, when I manually try and roll back I see the rollback label before the save label like in the screenshot, and likewise when I run the test.
It would make sense to me if it did the rollback and then saved it hpwever, meaning the Save should be before the rollback.

nikolajlauridsen: Coming back now close to a month later it's back to being save first then rollback, this is an odd one.

I spoke to Niels a while ago and we talked about how it, in reality, should just be rollback only, since it's kind of confusing that it's both save and rollback since that suggests that you've performed two actions, but that's a separate issue really. I'm resolving this conversation for now :)

src/Umbraco.Tests.AcceptanceTest/cypress/integration/Members/members.js
```diff
@@ -24,8 +25,8 @@ context('Members', () => {
 
         cy.get('input#_umb_login').clear().type(email);
         cy.get('input#_umb_email').clear().type(email);
-        cy.get('input#password').clear().type(password);
-        cy.get('input#confirmPassword').clear().type(password);
+        cy.get('input#password').clear().type(password, { timeout: passwordTimeout });
```
bergmania, line 28: I think it is okay in these tests to just type a shorter password. We wait too long for a simple password :)

nikolajlauridsen: Ah that's true, it does seem like an unnecessarily long password for no particular reason, I'll just cut it half. 
It might be a good idea to keep  the extended timeout since it doesn't cause the test to take longer, and I have had some weird cases when cypress stops typing for a little bit causing even small strings to take longer than 4 seconds, as mentioned in the part about the helper package in my description.

PR #8866, opened 2020-09-11
Netcore: Handle tree authentication

bergmania:
Handles tree auth in the same was as in v8.


elit0451 (changes requested):
Code looks alright, but we are missing something...

The following test scenario **failed**:

When I have logged in as **admin** (_a user with almost no permissions have been created_) 
and I am at the **Users section**,
I logout
and log in with the other user,
I get 500 error.

What happened, in essence, is that since I was at the Users' section just before logging out, the next time I log in with any user, it will try to send me to that section (including the users' tree) which I don't have access with the second user.

Request URL: https://localhost:44331/umbraco/backoffice/umbracotrees/applicationtree/GetApplicationTrees?application=users&tree=&use=main

Back-Office look: 
image

elit0451 (approved):

PR #8878, opened 2020-09-11
Fixes #8876 and #8872

### LINKED ISSUE #8872: Block action is inherting css it shouldn't
image

The action bar at the right looks wrong. the icons are too big, and there is a right margin on each of them making it look broken.

I think this comes from this css, responsible for styling the icon to the left of the label in the default block preview:
```css
.blockelement-labelblock-editor i {
    font-size: 22px;
    margin-right: 5px;
    display: inline-block;
    vertical-align: middle;
}
```

I'm only getting this, because I'm doing this [Block List in Block List approach for making a grid](https://twitter.com/skttl/status/1303675858494861312), which puts the action bar inside a `.blockelement-labelblock-editor`.

A simple fix is to change the selector to `.blockelement-labelblock-editor > i`

Version 8.7.0

### LINKED ISSUE #8876: Default block preview doesn't handle text overflow well
The default block preview kind of breaks, when the text-label exceeds the available space.
image

Can be fixed by adding the following css:

```css
.blockelement-labelblock-editor {
display:flex;
padding:12px 20px;
}
.blockelement-labelblock-editor i {
margin-right:10px;
}
.blockelement-labelblock-editor span {
padding-top:2px;
padding-bottom:2px;
}
```

Then it looks like this:
image

skttl:
If there's an existing issue for this PR then this fixes #8876 
Fixes #8872

### Description
This updates the css as described in the issues

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/blocklistentryeditors/labelblock/labelblock.editor.less
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/blocklistentryeditors/labelblock/labelblock.editor.less
@@ -1,6 +1,6 @@
 .blockelement-labelblock-editor {
     position: relative;
-    display: block;
+    display: flex;
     margin-bottom: 4px;
     margin-top: 4px;
     width: 100%;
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
It's fixed in #8986 - @nielslyngsoe please please let me teach you how to contribute to community pull requests, much nicer to credit community work and not have to re-do everything yourself.

PR #8882, opened 2020-09-11
Section picker: Fix accessibility and semantic issues

BatJan:
### Prerequisites

### Description
Depending on whether or not the "mediatypepicker.html" and controller is actively being used or not this PR is probably the final of it's kind turning `<a>` containing and empty `href` into a proper `<button>` and adding `aria-hidden="true"` on the icon 😃 

I've fixed some styling as well.

This is the view in question and the screendump is taken after the changes were made 👍 

section-picker


nul800sebastiaan:
Thanks @BatJan (and if this is truly the last one.. we need a party! 🎉 )

Cherry picked for 8.7.1 in f2ce837ca2f6287c20fa9d55c542a99fa0ddd339

PR #8883, opened 2020-09-11
Remove unused legacy mediatypepicker overlay

BatJan:
### Prerequisites

### Description
After some very thorough investigation it occurs to me that this overlay is no longer being used in the back office. Therefore I suggest we remove it - I realize this is probably a breaking change but it's usage seems to have been very niche in it's use case since it has only been used for displaying possible media types when uploading media to a folder that allowed for different media types like the built in "image" type and perhaps a custom made "Special image" type - Then this overlay would be displayed during the upload session. Maybe I suffer from poor imagination but I have a hard time imagining that 3rd party package authors or people who have done property editors or content apps have been using this specific overlay over some of the more general ones.

I suspect it became deprecated after #8043 was merged changing the overlay to be triggered using the overlay service and serve the "itempicker" overlay instead. I have not been able to find any references for this overlay so it should be safe to remove... Famous last words™ 😅


nul800sebastiaan:
I also can't imagine people using it, but let's be cautious and mark it as a breaking one! Thanks @BatJan!

Issue #8825, opened 2020-09-04
Modifier for colorpicker prevalue editor not working


rndfm:
When using the colorpicker prevalue editor with a modifier the modifier is ignored and the whole value object is returned.

I am seeing this issue on Umbraco version: 8.6.4

Reproduction
------------

Add a colorpicker styling setting to a grid editor. eg:

```
[
  {
    "label": "Background colour",
    "key": "background-color",
    "view": "colorpicker",
    "modifier": "{0}",
    "prevalues": [
      "#23265A",
      "#034EA2"
    ]
  }
]
```

The editor is shown correctly with the two colours.
But the modifier is not used when the page is rendered and the whole object is used.
```
"{
  "value": "23265a",
  "label": "#23265A",
  "id": "0"
}"
```

Expected result
Modifier should be used and output would have been: "#23265a"
(or maybe "23265a" as that is the first property)

Actual result
```
"{
  "value": "23265a",
  "label": "#23265A",
  "id": "0"
}"
```

bjarnef:
How does it render the grid layout? do you use the default Bootstrap 3 view?
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI/Views/Partials/Grid/Bootstrap3.cshtml

As I remember the prevalue editor should store a hex code without "#".
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/colorpicker.html#L8

@kjac have you noticed this issue with colorpicker prevalue editor? It seems this hasn't changed for a long time, so it would probably more be something related to `<umb-color-swatches>` component.

bjarnef:
I have checkted this with latest changes in `v8/contrib` where it seems to return the following, when using the rule in Styles and when default Bootstrap 3 view.

```
style="background-color:[object Object];
```

bjarnef:
The colorpicker property editor store a JSON format like this:

```
{"value":"034ea2","label":"034ea2","sortOrder":1,"id":"10"}
```

The colorpicker prevalue editor store a JSON format like this:

```
{"value":"034ea2","label":"#034EA2","id":"1"}
```

bjarnef:
It does seem the color is set correct on intial, when not using modifier.

Log og `$scope.model.value` in start of controller returns the json.

```
{value: "034ea2", label: "#034EA2", id: "1"}
``` 

[...]

PR #8885, opened 2020-09-12
Adjustments of colorpicker prevalue and property editor

bjarnef:
### Prerequisites

### Description
This PR has a few adjustments to colorpicker prevalue and property editors.
While looking at this issue https://github.com/umbraco/Umbraco-CMS/issues/8825 (although I didn't find a solution for now) I noticed a few other things to update.

- `$scope.isConfigured` was defined twice in colorpicker property editor (exact same line)
- Update to use `vm` in colorpicker property editor
- In colorpicker property editor we have an `isTrue()` function - however we have a `Object.toBoolean()` extension method, so we can use this instead.


nathanwoulfe (approved):
Awesome, as always. Thanks @bjarnef.

nul800sebastiaan:
This part is working nicely at least! 👍 Thanks @bjarnef!

PR #8886, opened 2020-09-12
Use String.CreateGuid in grid editor

bjarnef:
### Prerequisites

### Description
At the moment grid editor is using its own local function to create a guid for an unique id of e.g. controls.
However most places in backoffice we already use `String.CreateGuid()` extension method: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/lib/umbraco/Extensions.js#L49-L59

So I think we should remove this function and use the extension method instead to create an unique id. No need to reinvent the wheel 😄🎡

Here is an example with the current function vs `String.CreateGuid()`:

```
$scope.setUniqueId = function () {
    var guid1 = guid();
    var guid2 = String.CreateGuid();

    console.log("guid1", guid1);
    console.log("guid2", guid2);

    return guid1;
};
```

chrome_2020-09-12_17-14-59

Furthermore I couldn't find any references where the parameters in `setUniqueId()` function was used, so I have removed these.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js
@@ -627,21 +627,8 @@ angular.module("umbraco")
 
             }
 
-
-            var guid = (function () {
-                function s4() {
-                    return Math.floor((1 + Math.random()) * 0x10000)
-                        .toString(16)
-                        .substring(1);
-                }
-                return function () {
-                    return s4() + s4() + "-" + s4() + "-" + s4() + "-" +
-                        s4() + "-" + s4() + s4() + s4();
-                };
-            })();
-
-            $scope.setUniqueId = function (cell, index) {
-                return guid();
+            $scope.setUniqueId = function () {
+                return String.CreateGuid();
             };
 
             $scope.addControl = function (editor, cell, index, initialize) {


abjerner:
Great catch Bjarne 👍 

This also ensures that generated GUIDs follow the `XXXXXXXX-XXXX-4XXX-XXXX-XXXXXXXXXXXX` pattern.

I'll try to pull this down and give it a spin, but it shouldn't be a problem getting this approved and merged 😉

bjarnef:
Yeah, I noticed the format `XXXXXXXX-XXXX-4XXX-XXXX-XXXXXXXXXXXX` - not sure if there is a specific reason for the number `4` in the Guid (probably) 😄

abjerner:
There is standard for UUIDs/GUIDs (RFC 4122), so the `4` indicates the UUID version. This is similar to GUIDs generated by `Guid.NewGuid()` in C#.

I wasn't aware of this, but the format from `String.CreateGuid` is `xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx`. The `y` part should then be either `8`, `9`, `A` or `B`. I'm not really sure why this matter, but it's in line with the specification and the implementation in C#, so all good 👍

bjarnef:
Okay, maybe it should have a short comment around here:
https://github.com/umbraco/Umbraco-CMS/blob/0140d906a5a66d3ba512332e845a64d36169f911/src/Umbraco.Web.UI.Client/lib/umbraco/Extensions.js#L54

e.g. with link to https://tools.ietf.org/html/rfc4122

Here is also a bit more info about UUID compliant with RFC 4122: https://www.cryptosys.net/pki/uuid-rfc4122.html

> The digit at position 1 above is always "4" and the digit at position 2 is always one of "8", "9", "A" or "B".

abjerner:
@bjarnef I couldn't find information about this in https://tools.ietf.org/html/rfc4122, but it might be worth adding a comment with https://www.cryptosys.net/pki/uuid-rfc4122.html (I also found that when I did a quick readup).

Adding the comment is not directly related to this PR, so feel free to either add it via a new PR or as part of this one 😉

abjerner (approved):
Looks good 👍

PR #8892, opened 2020-09-14
Netcore: Migrate to use MS ILogger<T>

nikolajlauridsen:
Migrating ILogger to use Microsofts ILogger

IProfileLogging gets injected on its own and does not serve as a normal logger anymore, but that's to be expected since it's still our own logging implementation, it does, however, use a MS ILogger as the underlying logger abstractions.

The global logger in Current uses ILogger\<object\>, which is unfortunate, but there's not much to do about that, and it should generally be avoided using it.

Other than that the logger logs as expected, all the tests pass, and the logs are displayed in the log viewer of the back office as well. 

Steps to test:
Run the project and check that it logs as expected


src/Umbraco.Core/Logging/ILogger.cs
```diff
@@ -25,158 +27,67 @@ public interface ILogger
...
-        void Warn(Type reporting, string messageTemplate, params object[] propertyValues);
-
-        /// <summary>
-        /// Logs a warning message with an exception.
-        /// </summary>
-        /// <param name="reporting">The reporting type.</param>
-        /// <param name="exception">An exception.</param>
-        /// <param name="message">A message.</param>
-        void Warn(Type reporting, Exception exception, string message);
+        void LogWarning(string messageTemplate, params object[] propertyValues);
```
nzdev, line 62: The code should avoid the use of params keyword as it causes an array allocation. It would be better to also include overloads for 1, 2,3 and 4 parameters without the use of the params key.

nzdev:
Could this have supported the old method names as well, rather than updating every usage? Would be far less changes, easier to keep in sync with v8. 
Using params for all method signatures is inefficient as it always allocates an array when you only have 1 param. Serilog and the Microsoft logger have overloads for using 1, 2  and 3 params that avoid the allocation. 
I also wonder if this could have targeted v8 instead and have this get into the netcore branch that way. My thought would have been it's easier to introduce the abstractions/interfaces in v8 and provide a new implementation in .net core only when the v8 implementation could not be used.

bergmania:
@nzdev This is an intermediate state. We want to use MS ILogger abstraction and not our own.

Logging is one of the areas we have identified, where we want to align as much as possible with best practices in .NET Core.

src/Umbraco.Infrastructure/Umbraco.Infrastructure.csproj
```diff
@@ -14,6 +14,8 @@
       <PackageReference Include="LightInject" Version="6.3.4" />
       <PackageReference Include="LightInject.Annotation" Version="1.1.0" />
       <PackageReference Include="LightInject.Microsoft.DependencyInjection" Version="3.3.1" />
+      <PackageReference Include="LightInject.Microsoft.Hosting" Version="1.2.0" />
+      <PackageReference Include="MailKit" Version="2.6.0" />
```
nikolajlauridsen, line 18: I'm not really sure about this? Maybe something that happened in a merge?

src/Umbraco.Tests.Common/Umbraco.Tests.Common.csproj
```diff
@@ -6,6 +6,9 @@
 
   <ItemGroup>
     <PackageReference Include="AutoFixture.AutoMoq" Version="4.14.0" />
+    <PackageReference Include="AutoFixture.NUnit3" Version="4.11.0" />
```
nikolajlauridsen, line 9: Same thing here, merge problem perhaps?

src/Umbraco.Tests/TestHelpers/TestWithDatabaseBase.cs
```diff
@@ -150,7 +152,7 @@ protected virtual string GetDbProviderName()
 
         protected virtual string GetDbConnectionString()
         {
-            return @"Datasource=|DataDirectory|UmbracoNPocoTests.sdf;Flush Interval=1;";
+            return @"DataSource=|DataDirectory|UmbracoNPocoTests.sdf;Flush Interval=1;";
```
nikolajlauridsen, line 155: Don't think I changed this, is it supposed to be `Datasource` or `DataSource`?

src/Umbraco.Infrastructure/PropertyEditors/DataValueEditor.cs
```diff
@@ -217,7 +217,7 @@ public virtual object FromEditor(ContentPropertyData editorValue, object current
             var result = TryConvertValueToCrlType(editorValue.Value);
             if (result.Success == false)
             {
-                Current.Logger.Warn<DataValueEditor>("The value {EditorValue} cannot be converted to the type {StorageTypeValue}", editorValue.Value, ValueTypes.ToStorageType(ValueType));
+                Current.Logger.LogWarning("The value {EditorValue} cannot be converted to the type {StorageTypeValue}", editorValue.Value, ValueTypes.ToStorageType(ValueType));
```
AndyButland, line 220: Should we be injecting the typed logger rather than using `Current.Logger` here?

src/Umbraco.Infrastructure/PropertyEditors/ValueConverters/ImageCropperValueConverter.cs
```diff
@@ -44,7 +44,7 @@ public override object ConvertSourceToIntermediate(IPublishedElement owner, IPub
             catch (Exception ex)
             {
                 // cannot deserialize, assume it may be a raw image url
-                Current.Logger.Error<ImageCropperValueConverter>(ex, "Could not deserialize string '{JsonString}' into an image cropper value.", sourceString);
+                Current.Logger.LogError(ex, "Could not deserialize string '{JsonString}' into an image cropper value.", sourceString);
```
AndyButland, line 47: As above - should we be injecting here?

src/Umbraco.Infrastructure/PropertyEditors/ValueConverters/JsonValueConverter.cs
```diff
@@ -58,7 +58,7 @@ public override object ConvertSourceToIntermediate(IPublishedElement owner, IPub
                 }
                 catch (Exception ex)
                 {
-                    Current.Logger.Error<JsonValueConverter>(ex, "Could not parse the string '{JsonString}' to a json object", sourceString);
+                    Current.Logger.LogError(ex, "Could not parse the string '{JsonString}' to a json object", sourceString);
```
AndyButland, line 61: One more - so guess there's a few.  General point then is whether we should be looking to remove use of `Current.Logger` as part this PR.

src/Umbraco.Infrastructure/Examine/IndexRebuilder.cs
```diff
@@ -14,12 +15,14 @@ namespace Umbraco.Examine
     /// </summary>
     public class IndexRebuilder
     {
-        private readonly IProfilingLogger _logger;
+        private readonly IProfilingLogger _pLogger;
```
AndyButland, line 18: Just noticed that most of this time this variable is called `_profilingLogger` (or `profilingLogger` in the constructor), which is a little more readable/explanatory and easier to see which one is being used.  So small thing, but maybe could do a find/replace to standardise?

AndyButland (commented):
Looks good to me.  I got to the end!  Just left a couple of minor inline points.

src/Umbraco.Infrastructure/PropertyEditors/DataValueEditor.cs
bergmania: It's a good idea, but I think we need to have this PR merged, and then we could look into it afterwards..

bergmania (approved):
@nikolajlauridsen  I reviewed and fixed issues/TODOs.. Please take a look at my commits

PR #8898, opened 2020-09-15
LiteDb Nucache Provider

nzdev:
### Prerequisites

### Description
Draft implementation of a LiteDb Nucache provider. Drop in replacement for BPlusTree.

This is an example implementation of nucache using LiteDb. This might be better suited as a package. It may be a reasonable starting point for discussing plugging in NoSQL implementations other than BPlusTree that would allow for queries / or an IQueryable implementation so umbraco does not load the entire set of content properties when filtering. Perhaps CQRS would be appropriate for ContentStore queries and commands so it's easy to swap out with something like cosmos db.

A known limitation is litedb has a 1mb limit per doc, however it has a file api so large properties could use the file api (not implemented).

How to test:
Currently the nucachecomposer is set to use litedb. Run the Web.UI project with a starter kit 2x, the first time will write to litedb. The second would read from.


Shazwazza:
As always, your ideas and trials to push Umbraco to better places is awesome  :) 🎉 

It's been on our list for a long time to be able to lazy load things instead of having everything in memory (like we've previously discussed). The biggest barrier to this currently is routing because incoming and outgoing routing in umbraco is dynamic and is entirely based on the tree hierarchy. Ideally we refactor routing entirely so it's not dependent on this graph, else if it's not always in memory it will need to be recalculated/queried each time which happens quite often.

I have a lot of tasks/PRs starred in my inbox from you to follow up on, i will get there eventually!

nzdev:
@Shazwazza this pr decouples the routing, but I need help with the unit tests https://github.com/umbraco/Umbraco-CMS/pull/8927.

nul800sebastiaan:
FYI: looks like this one got closed automatically because https://github.com/umbraco/Umbraco-CMS/pull/8376 was merged. I've re-targeted it now and I'll attempt to fix the merge conflicts!

nul800sebastiaan:
Wow I'm sorry @nzdev - I did try to do a merge but that was WAAAAAY to brave of me to attempt. I am hoping you have some time to do it yourself, this is a really tough one with all the other changes on `PublishedSnapshotService.cs` and me not understanding all of the intentions.

Shazwazza:
I think this is still a proof of concept though and there will be other API changes required before we can support something like this. Not sure if it's worth keeping open now? Might be better to re-attempt this when the other nucache API changes are looked into?

nul800sebastiaan:
Ah I didn't realize how old this was! Happy to close for now but if you think it's worth preserving @nzdev let us know!

PR #8899, opened 2020-09-15
Prevent InvalidCastException being thrown when ContentPicker has picked a Udi

nzdev:
### Prerequisites

### Description
Prevent InvalidCastException being thrown when ContentPicker has picked a Udi. Trying to convert to an int will fail for Udis with an InvalidCastException.

Checks if the value starts with umb to determine if it is a udi, if Udi, don't try cast to int.

How to test?

Use a content picker on a document, pick a udi. InvalidCastException should not be thrown when the property converter runs

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/ValueConverters/ContentPickerValueConverter.cs
+++ b/src/Umbraco.Web/PropertyEditors/ValueConverters/ContentPickerValueConverter.cs
@@ -36,9 +36,14 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
         {
             if (source == null) return null;
 
-            var attemptConvertInt = source.TryConvertTo<int>();
-            if (attemptConvertInt.Success)
-                return attemptConvertInt.Result;
+            //Don't attempt to convert to int for UDI
+            if(!(source is string) || source is string strSource && !string.IsNullOrWhiteSpace(strSource) && !strSource.StartsWith("umb"))
+            {
+                 var attemptConvertInt = source.TryConvertTo<int>();
+                    if (attemptConvertInt.Success)
+                        return attemptConvertInt.Result;
+            }
+
             var attemptConvertUdi = source.TryConvertTo<Udi>();
             if (attemptConvertUdi.Success)
                 return attemptConvertUdi.Result;


nathanwoulfe (commented):
Hi @nzdev I've had a look at this, it does exactly what you've said, but I'm wondering why it matters - the exception is managed by the `TryConvertTo` method, and execution continues correctly. I'd expect that a lot of the 100+ references to `TryConvertTo` do similar, where the exception is swallowed and the success property used as a condition...

nzdev:
It matters. Throwing exceptions is not free. It breaks the flow and slows program execution. I get that they are being swallowed so it may appear to be no harm, but it does matter. If you are trying to debug umbraco itself you will see the exceptions being thrown in the debug window. Then you may wonder if you broke something. Then you spend time investigating whether you broke something. Alternatively, you are looking for where a InvalidCastException was thrown and set the debugger to break on that type and now you have to contend with ok vs not ok exceptions. I know, this happened to me.

The design of tryconvertto goes against the pattern in .net. Every other try convert returns a bool on whether the conversion succeeded and the converted value is set on an out parameter and they don't throw exceptions. The better change would be to fix the design issue.


We should try to make umbraco as a framework perform the best it can so it's the best choice for people to increase it's adoption.

nathanwoulfe:
Cool, that makes total sense. Given that, is a generic conversion function really a good idea, since it's relying on throwing the exception when unsuccessful? Couldn't something like `TryConvertToInt(object)` internally implement the guards from your change and reduce/remove the use of try-catch for controlling flow?

nathanwoulfe (approved):

Shazwazza:
TryConvertTo shouldn't throw - it is 'trying' to do something and if not successful will return no successful. The typical pattern of returning a bool with an out parameter is essentially the same as here with returning an `Attempt<T>` which is not an uncommon pattern, other systems have a different name, other systems return a Tuple. This type of pattern is actually required for async Try methods since out params are not supported on those. We have many unit tests for TryConvertTo, it seems to me like whatever is throwing that exception within TryConvertTo shouldn't? It's been a long time since I looked at that code but IIRC TryConvertTo should never throw.

nzdev:
I'm not sure exactly what tryconvert is doing, it looks complicated. But eventually it kind of gives up and trys to cast without using is or as. Adding an is check makes sense, just not sure if that would not work with dynamic types.

Shazwazza:
Basically tries the fastest way to convert from one type to another using the many different conversion ways built into .net. This fix in this PR is good though we don't need to go changing underlying stuff to make this better. At some stage it will be worth looking into TryConvertTo for some specific types. Using 'is' won't work I don't think because I don't think that runs implicit cast operators? I could be wrong though but I assume 'as' would?

nzdev:
I wasn't too sure how it would work for implicit casts and dynamic types. With class and struct generic constraints you could work around the limitation on as, but that would require new methods.

nathanwoulfe:
@Shazwazza @nzdev as far as the intent of this PR, should be ok to merge, yeah? Changes fix the issue as raised, anything else needs more discussion and would be a substantive change.

nzdev:
Ok to merge @nathanwoulfe

PR #8904, opened 2020-09-15
Reduce memory use and speed up queries by only fetching required properties and documents from examine

nzdev:
### Prerequisites

### Description
What?
Optimize examine queries to fetch the least number of documents/fields required to keep the same result.

Why?
Improve query performance, reduce memory use.

Requires this PR for Examine to be merged and a new examine package to be released / umbraco update to use new examine package:
https://github.com/Shazwazza/Examine/pull/184

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -4,11 +4,11 @@
 
 We are the friendly CMS. And our friendliness stems from our values. That's why we have set for ourselves, Umbraco HQ, and the community, five values to guide us in everything we do:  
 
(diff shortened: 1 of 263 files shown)


Shazwazza:
Makes sense! Just need to get that examine release out. Only issues here is that old breaking change thing again. We can't really change interfaces so we'll have to figure out a different way to. Sorry i know that's such a pain.

nzdev:
Changes to interface made @Shazwazza

Shazwazza:
Great stuff, just FYI I'm on holidays for the next 3 weeks so will follow up on all of this when I return.

Issue #7622, opened 2020-02-10
TagsValueConverter throws exception when deserialization fails


rasmuseeg:
I changed the property from DataType Content Picker, to Tags.
This caused the picked DateType value of UDI to fail during deserialization from string of "umb://.." to JArray, on the following line:

https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Core/PropertyEditors/ValueConverters/TagsValueConverter.cs#L38

How should this exception be handled?

## Umbraco version

8.4.1

### Steps to reproduce
1. Create property of DateType Content Picker
2. Go to content and select content, and save. (A value must be saved in db)
3. Goto Document Types, and change property datatype to tags.

### Expected result
I expected the result to be null or empty array.

### Actual result

The exception is thrown on the following line:
https://github.com/umbraco/Umbraco-CMS/blob/3bfd9b71e290354744d677440983ecd06c5c0788/src/Umbraco.Core/PropertyEditors/ValueConverters/TagsValueConverter.cs#L38

nul800sebastiaan:
You're causing data to be unexpected, which is indeed an exception. 

You're also going to get an exception on the frontend, as the valueconverter for Tags will not work on the stored data from the pickers.

For performance reasons, when you change a datatype on a document type, we don't change the data on each content item. 

I would suggest we do a try/catch of this error but it is specifically an exception so that you will know when something is corrupt in your data (if you didn't switch datatypes, so errors in your actual tags data). 
So I'm not sure what to do here.. I would like to do the friendly thing and give you an empty array, but will it be an unfriendly thing for people with problematic tags data?

rasmuseeg:
Hi @nul800sebastiaan ,

Thanks for your quick reply :-)

I do understand that this an edge case where taking sides causes problems elsewhere.

But when changing a datatype causing the system to fail, but i wasn't warned. I would think that Umbraco knows that i dont know what i'm doing, and handles that for me.
Otherwise it should warn about doing that change.

But personnally i would prefer to have one of the following options:
* Umbraco just catches the errors and logs them
* Custom implementation of a property exception handler to catch these exceptions.

What do you think?

Cases where the current setup causes problems, is if the developer only uses the backoffice to develop and change documents. This would cause their site to break, which is not warned before saving changes to a document. Like how removing compositions shows a warning.

stevemegson:
[...]

PR #8906, opened 2020-09-15
Fix NullReferenceException while deserializing empty value in TagsValueConverter

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7622.

### Description
This change ensures that stored values (especially empty strings) for the `Umbraco.Tags` editor do not result in a `NullReferenceException`, as mentioned in the linked issue.

I've also removed the dependency on the `IDataTypeService`, as getting the data type configuration can be done without services/database lookups in Umbraco 8.

### DIFF
--- a/src/Umbraco.Core/PropertyEditors/ValueConverters/TagsValueConverter.cs
+++ b/src/Umbraco.Core/PropertyEditors/ValueConverters/TagsValueConverter.cs
@@ -1,8 +1,6 @@
 ﻿using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using Umbraco.Core.Models;
 using Umbraco.Core.Models.PublishedContent;
 using Umbraco.Core.Services;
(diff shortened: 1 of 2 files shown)


src/Umbraco.Core/PropertyEditors/ValueConverters/TagsValueConverter.cs
```diff
@@ -12,69 +10,42 @@ namespace Umbraco.Core.PropertyEditors.ValueConverters
...
-        private readonly IDataTypeService _dataTypeService;
+        public TagsValueConverter()
+        { }
 
+        [Obsolete("This constructor isn't required anymore, because we don't need services to lookup the data type configuration.")]
         public TagsValueConverter(IDataTypeService dataTypeService)
-        {
-            _dataTypeService = dataTypeService ?? throw new ArgumentNullException(nameof(dataTypeService));
-        }
+        { }
```
ronaldbarendse, line 18: I've obsoleted the original constructor, so it's clear to not use this one (although it doesn't really matter, as it doesn't do anything with the parameter) and added a parameterless one that can be used instead.

src/Umbraco.Core/PropertyEditors/ValueConverters/TagsValueConverter.cs
```diff
@@ -12,69 +10,42 @@ namespace Umbraco.Core.PropertyEditors.ValueConverters
...
             => PropertyCacheLevel.Element;
 
         public override object ConvertSourceToIntermediate(IPublishedElement owner, IPublishedPropertyType propertyType, object source, bool preview)
         {
-            if (source == null) return Array.Empty<string>();
-
-            // if Json storage type deserialize and return as string array
-            if (JsonStorageType(propertyType.DataType.Id))
+            var sourceString = source?.ToString();
+            if (string.IsNullOrEmpty(sourceString))
```
ronaldbarendse, line 32: This is the actual fix, as it now short-circuits on `null` _or_ empty values.

src/Umbraco.Core/PropertyEditors/ValueConverters/TagsValueConverter.cs
```diff
@@ -12,69 +10,42 @@ namespace Umbraco.Core.PropertyEditors.ValueConverters
...
-            // refresh in DataTypeCacheRefresher
-
-            return Storages.GetOrAdd(dataTypeId, id =>
+            switch (propertyType.DataType.ConfigurationAs<TagConfiguration>().StorageType)
             {
-                var configuration = _dataTypeService.GetDataType(id).ConfigurationAs<TagConfiguration>();
-                return configuration.StorageType == TagsStorageType.Json;
-            });
+                case TagsStorageType.Csv:
+                    return sourceString.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
```
ronaldbarendse, line 40: I'm not sure why empty entries/tags need to be removed: they shouldn't get stored in the first place and the JSON storage type doesn't do this either. I've kept this for now though...

src/Umbraco.Core/PropertyEditors/ValueConverters/TagsValueConverter.cs
```diff
@@ -12,69 +10,42 @@ namespace Umbraco.Core.PropertyEditors.ValueConverters
...
-        /// The data type id.
-        /// </param>
-        /// <returns>
-        /// The <see cref="bool"/>.
-        /// </returns>
-        private bool JsonStorageType(int dataTypeId)
-        {
-            // GetDataType(id) is cached at repository level; still, there is some
-            // deep-cloning involved (expensive) - better cache here + trigger
-            // refresh in DataTypeCacheRefresher
```
ronaldbarendse, line 64: This isn't the case anymore in V8, so lets keep it simple and remove this unnecessary caching.

src/Umbraco.Core/PropertyEditors/ValueConverters/TagsValueConverter.cs
ronaldbarendse: This is probably because the tag names can also include comma's, which could result in empty entries when stored as CSV.

PR #8907, opened 2020-09-15
Fix IndexOutOfRangeException when converting single value to range in SliderValueConverter

ronaldbarendse:
### Prerequisites

### Description
While doing PR https://github.com/umbraco/Umbraco-CMS/pull/8906, I noticed `SliderValueConverter` could also throw an `IndexOutOfRangeException` if the saved value was still a single value, but the configuration changed to return a range. It will now return a `Range<decimal>` with both minimum and maximum set to the same value.

I've also removed the dependency on the `IDataTypeService`, as getting the data type configuration can be done without services/database lookups in Umbraco 8.

### DIFF
--- a/src/Umbraco.Core/PropertyEditors/ValueConverters/SliderValueConverter.cs
+++ b/src/Umbraco.Core/PropertyEditors/ValueConverters/SliderValueConverter.cs
@@ -1,7 +1,4 @@
 ﻿using System;
-using System.Collections.Concurrent;
-using System.Collections.Generic;
-using System.Linq;
 using Umbraco.Core.Models;
 using Umbraco.Core.Models.PublishedContent;
 using Umbraco.Core.Services;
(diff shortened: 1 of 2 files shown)


src/Umbraco.Core/PropertyEditors/ValueConverters/SliderValueConverter.cs
```diff
@@ -11,77 +8,73 @@ namespace Umbraco.Core.PropertyEditors.ValueConverters
...
-            _dataTypeService = dataTypeService ?? throw new ArgumentNullException(nameof(dataTypeService));
-        }
+        { }
 
         public override bool IsConverter(IPublishedPropertyType propertyType)
             => propertyType.EditorAlias.InvariantEquals(Constants.PropertyEditors.Aliases.Slider);
 
         public override Type GetPropertyValueType(IPublishedPropertyType propertyType)
-            => IsRangeDataType(propertyType.DataType.Id) ? typeof (Range<decimal>) : typeof (decimal);
+            => IsRangeDataType(propertyType) ? typeof(Range<decimal>) : typeof(decimal);
```
ronaldbarendse, line 22: This PVC can also return `null` (for no or an invalid value), so the type for single values should actually be `typeof(decimal?)`, right? This will probably be a breaking change, so I haven't included it in this PR.

Although changing the value type is done previously for the Media Picker (https://github.com/umbraco/Umbraco-CMS/issues/6033), it's not without problems, as it changes the generated models and all code that expects a non-nullable value.

src/Umbraco.Core/PropertyEditors/ValueConverters/SliderValueConverter.cs
```diff
@@ -11,77 +8,73 @@ namespace Umbraco.Core.PropertyEditors.ValueConverters
...
-                if (minimumAttempt.Success && maximumAttempt.Success)
+                if (rangeRawValues.Length == 1 && minimumAttempt.Success)
+                {
+                    // Configuration is probably changed from single to range, return range with same min/max
+                    return new Range<decimal>
+                    {
+                        Minimum = minimumAttempt.Result,
+                        Maximum = minimumAttempt.Result
+                    };
+                }
```
ronaldbarendse, line 49: When changing the configuration to enable ranges, the saved data is still a single value, so let's use that as both minimum and maximum value.

src/Umbraco.Core/PropertyEditors/ValueConverters/SliderValueConverter.cs
```diff
@@ -11,77 +8,73 @@ namespace Umbraco.Core.PropertyEditors.ValueConverters
...
+                else if (rangeRawValues.Length == 2)
```
ronaldbarendse, line 50: If the saved data isn't exactly 2 values, it should be considered invalid (and return `null` below).

src/Umbraco.Core/PropertyEditors/ValueConverters/SliderValueConverter.cs
```diff
@@ -11,77 +8,73 @@ namespace Umbraco.Core.PropertyEditors.ValueConverters
...
-        /// </param>
-        /// <returns>
-        /// The <see cref="bool"/>.
-        /// </returns>
-        private bool IsRangeDataType(int dataTypeId)
-        {
-            // GetPreValuesCollectionByDataTypeId is cached at repository level;
-            // still, the collection is deep-cloned so this is kinda expensive,
-            // better to cache here + trigger refresh in DataTypeCacheRefresher
-            // TODO: this is cheap now, remove the caching
```
ronaldbarendse, line 70: This isn't the case anymore in V8 (as the TODO comment confirms), so lets keep it simple and remove this unnecessary caching.

PR #8908, opened 2020-09-15
Change wrong URL in KeepAlive XML Comments

Swimburger:
The XML comments for KeepAlive mention the URL for the KeepAlive.Ping action but incorrectly omit 'umbraco' from the path.
The change adds the 'umbraco' section to the path.


nathanwoulfe (approved):
👍

nul800sebastiaan:
@nathanwoulfe  We're always targeting the next minor version! If we think it's urgent enough for a patch release we can cherry-pick to that. So 8.8 it is for this one.

nathanwoulfe:
@nul800sebastiaan all good - 8.8 it is.

PR #8917, opened 2020-09-16
Improve speed and reduce allocations by implementing logger overloads to avoid params [] allocation

nzdev:
### Prerequisites

### Description
What?
Improve logging performance and reduce memory allocations.
Implement overloads for logging that take 1, 2 or 3 parameters so that the params array method is not used when there are 1 - 3 parameters. The non params[] overloads are used automatically by implementing in this way so no need to use find and replace.

Why?
Logging is faster and allocates no memory for a params array. This is important as even if you have log level set to info, the log debug method is still called and an array is still allocated.

Benchmark: Synthetic, 1000 calls to log debug with 2 parameters for the message template
```
//        BenchmarkDotNet=v0.11.3, OS=Windows 10.0.18362
        //Intel Core i5-8265U CPU 1.60GHz(Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
        // [Host]     : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.8.4180.0
        //  Job-JIATTD : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.8.4180.0

        //IterationCount=3  IterationTime=100.0000 ms LaunchCount = 1
        //WarmupCount=3

        //       Method |      Mean |     Error |    StdDev | Ratio | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
        //------------- |----------:|----------:|----------:|------:|------------:|------------:|------------:|--------------------:|
        //     Baseline | 14.599 us | 1.0882 us | 0.0596 us |  1.00 |     10.0420 |           - |           - |             32048 B |
        // NewOverload2 |  1.775 us | 0.4056 us | 0.0222 us |  0.12 |           - |           - |           - |                   - |
```

How to test?
Existing unit tests pass. New method signatures are used / have references.


Shazwazza:
This makes a lot of sense of course! Only problem is the dreaded breaking changes 😢 
Do you think it would be possible to do the interface shuffle for this and have something like ILogger2 and have ext methods that cast? In the netcore branch logging is overhauled to use MS logging directly so will ask the team to ensure that arrays aren't being allocated all the time there too.

nzdev:
@Shazwazza Made the necessary changes. The test failure looks to be unrelated to the code changes.

Shazwazza (approved):
Makes so much sense and can definitely save on an enormous amount of allocations. I'm all for this one. 

@bergmania IIRC to port this to netcore would basically be deleting all of this code and starting from scratch using https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?view=aspnetcore-5.0 I think

@nul800sebastiaan just pinging you about this one too :)

nzdev:
Any chance of getting this one merged as well as it's going to keep encountering conflicts.

bergmania:
I'm not sure how this should be merged to the netcore branch. We are not using our own interface anymore.

I guess, on that branch, we should ignore these changes and then at some point look into [LoggerMessage](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?view=aspnetcore-5.0)?

nzdev:
I would strongly recommend ensuring the logging pattern is a high performance one such as LoggerMessage **before** releasing v9 @bergmania, given logging is used everywhere and in performance critical code. It will be an unpleasant breaking change otherwise to get this right later, or will compromise on performance. For example, this implementation is suboptimal as there is another if condition called for every call to the logger through the extension method workaround where this checks for ILogger2. This could have been faster if it was written as a breaking change by avoiding the branching.

Issue #8367, opened 2020-07-01
V8: LINQ Published Content query memory use high / initial query slow.


nzdev:
Might not be a issue for most, however, when you have a large number of content nodes and try to run a linq query such as get the top 10 published pages that start with some path ordered by a date property descending, the first time the query is very slow and memory use increases significantly.

For one scenario with a database of ~26000 pages, it takes ~ 2000 milliseconds to get the top 10 most recently published pages for the first linq query, and 94 milliseconds for subsequent queries. Using SQL to run an equivalent query to get the ids then just look up the PublishedContent by id's runs locally in ~330ms and saves about 100mb of memory for this site/query.

I have another site with around 80,000 pages I'm trying to work my way up to being viable in v8, the goal being to try reduce memory used by older content while still having the content available.

It would be ideal to only load the properties used in the where/select clause until / ToList/ToArray etc are called so that less memory is used. 
Having a more memory efficient way similar to v7, where you could use an XpathNavigator instead of IPublishedContent / examine would help, something like another content nucache database that only contains the properties you have marked as usable in queries you could use to find id's then look up the iPublishedContent from the main nucache. Examine might be ideal, but it's a bit hit and miss on Azure and Azure Search is an additional cost that could be avoided given cpu usage isn't high for this type of query. This could also avoid the need to increase virtual machine/app service plan costs due to needing additional memory to store less frequently used content when you have low cpu usage. 

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.2

Reproduction
------------

### Bug summary
Entire set of property data is loaded from nucache for each content item into memory when you are not filtering on these properties and the properties may not be required as the documents won't be in the result set.

### Specifics

Umbraco V8.6.2
Chrome latest

### Steps to reproduce

1. Set up a database with a large number of pages (10,000+) with long html bodies in a property (not used by where clause in query).
2. Run a linq query such as (Top 10 most recent articles in a section of the site ordered by a custom publishDate property, falling back to CreateDate where a custom publish date property is unavailable)
`var articles = CurrentPage.Descendants().Where(x => x.ContentType.Alias == "article")
                .OrderByDescending(x => x.Value<DateTime>("publishDate",defaultValue: x.CreateDate))
                .Skip(0)
                .Take(10).ToList()`
3. The first time the query is run it can take seconds to complete and appears to load the entire set of descendants into memory.
4. Rerun the linq query.

### Expected result

[...]

PR #8927, opened 2020-09-17
Decouple routing from content cache

nzdev:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8367 

### Description
What?
Decouples routing from the content cache by introducing IContentRouter.

Why?
Allow for implementing routing without traversing the content cache which is really not efficient. One step closer to https://github.com/umbraco/Umbraco-CMS/issues/8367.

How to test?
I'm stuck on getting a few unit tests to work as they still use the xml cache. Some assistance with changing them to nucache would be awesome.
If the unit tests pass then it works :)


src/Umbraco.Web/PublishedCache/NuCache/ContentCache.cs
```diff
@@ -29,194 +25,12 @@ internal class ContentCache : PublishedCacheBase, IPublishedContentCache2, INavi
...
-        {
-            return GetByRoute(PreviewDefault, route, hideTopLevelNode, culture);
-        }
-
-        public IPublishedContent GetByRoute(bool preview, string route, bool? hideTopLevelNode = null, string culture = null)
-        {
-            if (route == null) throw new ArgumentNullException(nameof(route));
-
-            var cache = preview == false || PublishedSnapshotService.FullCacheWhenPreviewing ? _elementsCache : _snapshotCache;
-            var key = CacheKeys.ContentCacheContentByRoute(route, preview, culture);
```
Shazwazza, line 68: The usage of `CacheKeys.ContentCacheContentByRoute` seems to be removed in the refactoring. Was this intended? Do we still use this cache key?

src/Umbraco.Web/UmbracoContext.cs
```diff
@@ -37,7 +38,8 @@ public class UmbracoContext : DisposableObjectSlim, IDisposeOnRequestEnd
             IEnumerable<IUrlProvider> urlProviders,
             IEnumerable<IMediaUrlProvider> mediaUrlProviders,
             IGlobalSettings globalSettings,
-            IVariationContextAccessor variationContextAccessor)
+            IVariationContextAccessor variationContextAccessor,
+            IPublishedSnapshotAccessor publishedSnapshotAccessor)
```
Shazwazza, line 42: This is a circular refernce. The IPublishedSnapshotAccessor depends on the UmbracoContext, the default implementation is UmbracoContextPublishedSnapshotAccessor. The UmbracoContext should not depend on or expose this service.

src/Umbraco.Web/UmbracoContextFactory.cs
```diff
@@ -31,11 +31,16 @@ public class UmbracoContextFactory : IUmbracoContextFactory
         private readonly UrlProviderCollection _urlProviders;
         private readonly MediaUrlProviderCollection _mediaUrlProviders;
         private readonly IUserService _userService;
+        private readonly IPublishedSnapshotAccessor _publishedSnapshotAccessor;
```
Shazwazza, line 34: See other comment

src/Umbraco.Web/PublishedCache/IContentRouter.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <summary>
+        /// Gets the route for a content identified by its alias.
+        /// </summary>
+        /// <param name="preview">In Preview</param>
+        /// <param name="rootNodeId">Domain content id</param>
+        /// <param name="culture">Request culture</param>
+        /// <param name="alias">Url Alias</param>
+        /// <returns>Routing Outcome</returns>
+        ContentRoutingResult GetIdByAlias(bool preview, int rootNodeId, string culture, string alias);
```
Shazwazza, line 57: (Brain dump...)

This abstraction will be able to replace the logic for the built in IContentFinders which is nice but any custom IContentFinders will still directly access the cache. As an example, Articulate has a custom one: https://github.com/Shazwazza/Articulate/blob/master/src/Articulate/Routing/DateFormattedPostContentFinder.cs ... that's a pretty custom one that needs to query the content cache for property values. I guess IContentFinders can be replaced so there is still an opportunity for devs to replace custom ones too. In the future, or as an increment to this work, I think it would be good to have routes persisted to a DB table and that persisting would occur when content is saved where we could ask all IUrlProvider for the content's URLs and save those. In that case IContentFinders won't really execute, only as a last resort if no persisted URLs are matched. 

Some notes on existing abstractions:

* IPublishedRouter - which is a large service and too big to re-implement nicely but potentially the logic within it's `bool TryRouteRequest(PublishedRequest request)` is what needs to be examined and could be abstracted. This does a few things:
  * Looks up domains - this can also lookup items in the content cache
  * Handles redirect results
  * Runs the IContentFinders
* IContentFinder - these are replaceable today so in theory you could already replace the routing lookups for everything contained in this new interface - So then it's a decision of if it makes sense to have this new interface or just replace the existing IContentFinders. I guess it's easier to replace one service than many, but if we had persisted URLs maybe it wouldn't matter?
* IUrlProvider - outbound routing, generates URLs for content. I 'think' if we persisted the result of GetOtherUrls in a DB in a format that was easily queryable/indexed (possibly similar to the hashing mechanism done in the UrlRedirect table) it could work if executed on content save. 

If we had persisted URLs/routes then we could probably have an IContentFinder for that which would be the first one - this would satisfy the statement above that 

> IContentFinders won't really execute, only as a last resort if no persisted URLs are matched. 

But it would probably still be good to abstract out and split up the IPublishedRouter.TryRouteRequest

src/Umbraco.Web/PublishedCache/ContentCacheContentRouter.cs
```diff
@@ -0,0 +1,267 @@
...
+namespace Umbraco.Web.PublishedCache
+{
+    internal class ContentCacheContentRouter : IContentRouter
+    {
+        private readonly IPublishedSnapshotAccessor _snapshot;
+        private readonly IGlobalSettings _globalSettings;
+        private readonly IAppCache _snapshotCache;
+        private readonly IAppCache _elementsCache;
+
+        public ContentCacheContentRouter(IPublishedSnapshotAccessor snapshot, IGlobalSettings globalSettings,
```
Shazwazza, line 21: I'm don't think this works as intended. The previous usage of the snapshotCache and elementsCache are not singleton instances and are not injected. The instances of snapshotCache and elementsCache are managed via the PublishedSnapshotService. The PublishedSnapshotService creates a new snapshot cache instance for each generation it creates. This is sort of like a 3 dimensional dictionary and each snapshot/generation is a new layer so that any existing requests continue to use the existing snapshot data that it's operating on. These snapshots are cleaned up when they are no longer in use. Whenever a call to PublishedSnapshotService.GetElements is made, a new snapshotCache instance is used and the existing elementsCache instance is used if a new generation is not needed. In this case these caches will be singletons that are never cleared or updated and will clash between generations. This object should probably not be responsible for any of this caching and the caching responsibility should be where it was before. Also note that `IPublishedSnapshotAccessor` is a request based object that is based on the UmbracoContext, if there is no request, then it will be null so it might not be ideal for this to be based on that. 

I've added notes to this interface declaration as well. I think that for this to work each method would need to take an isntance of `ContentStore.Snapshot` to work on and it would only deal with lookups, no caching. Or something along those lines.

Shazwazza (changes requested):
Great to get the conversation started on this, it will be an interesting task. There's a ton of options/ideas. Yeah the tests are going to be a pain converting to nucache, it's been on the list for a real long time and we're going to need to do it for netcore anyways so will probably be best to do that in the v8 branch first and merge up. I've left a bunch of notes inline, including some brain dumps, etc... let me know what you think when you get a chance.

src/Umbraco.Web/PublishedCache/NuCache/ContentCache.cs
nzdev: No longer used. CacheKeys.ContentCacheContentIdByRoute is used instead.

src/Umbraco.Web/PublishedCache/ContentRoutingResult.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Umbraco.Web.PublishedCache
+{
+    public enum RoutingOutcome
```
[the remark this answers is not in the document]
nzdev: Moved to separate file

src/Umbraco.Web/PublishedCache/ContentRoutingResult.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        NotFound = 0,
+        Found = 1
+    }
+    public struct ContentRoutingResult
+    {
+        public int Id { get; set; }
```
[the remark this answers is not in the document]
nzdev: Changed constructor. set no longer available.

src/Umbraco.Web/UmbracoContext.cs
nzdev: Removed

src/Umbraco.Web/UmbracoContextFactory.cs
nzdev: Removed

src/Umbraco.Web/PublishedCache/NuCache/ContentCache.cs
nzdev: Need to reimplement this cache somewhere now that CacheKeys.ContentCacheContentIdByRoute is removed

src/Umbraco.Web/PublishedCache/ContentCacheContentRouter.cs
nzdev: I removed the caching, but it will still need to be added back somewhere.

src/Umbraco.Web/PublishedCache/IContentRouter.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Gets the route for a content identified by its alias.
+        /// </summary>
+        /// <param name="preview">In Preview</param>
+        /// <param name="rootNodeId">Domain content id</param>
+        /// <param name="culture">Request culture</param>
+        /// <param name="alias">Url Alias</param>
+        /// <returns>Routing Outcome</returns>
+        ContentRoutingResult GetIdByAlias(IPublishedContentCache currentSnapshot, bool preview, int rootNodeId, string culture, string alias);
```
Shazwazza, line 46: The other overloads were changed to IPublishedCache2, wondering if this should also be changed so it's consistent?

src/Umbraco.Web/PublishedCache/IContentRouter.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <returns>The route.</returns>
+        /// <remarks>The value of <paramref name="preview"/> overrides defaults.</remarks>
+        string GetRouteById(IPublishedCache2 snapshot, IDomainCache domainCache, bool preview, int contentId, string culture = null);
+        // <summary>
+        /// Gets the route for a content identified by its unique identifier.
+        /// </summary>
+        /// <param name="contentId">The content unique identifier.</param>
+        /// <returns>The route.</returns>
+        /// <remarks>Considers published or unpublished content depending on defaults.</remarks>
+        string GetRouteById(bool defaultPreview, IPublishedCache2 snapshot, IDomainCache domainCache, int contentId, string culture = null);
```
Shazwazza, line 36: The other methods have a `bool preview` should `defaultPreview` here be changed to `preview` and moved after the `IPublishedCache2` declaration for consistency? but then if that happens this method becomes identical to the above method?

src/Umbraco.Web/PublishedCache/IContentRouter.cs
```diff
@@ -0,0 +1,48 @@
...
+        ContentRoutingResult GetIdByRoute(IPublishedCache2 currentSnapshot,bool preview, string route, bool? hideTopLevelNode, string culture);
+       
+        /// <summary>
+        /// Gets the route for a content identified by its unique identifier.
+        /// </summary>
+        /// <param name="preview">A value indicating whether to consider unpublished content.</param>
+        /// <param name="contentId">The content unique identifier.</param>
+        /// <returns>The route.</returns>
+        /// <remarks>The value of <paramref name="preview"/> overrides defaults.</remarks>
+        string GetRouteById(IPublishedCache2 snapshot, IDomainCache domainCache, bool preview, int contentId, string culture = null);
```
Shazwazza, line 29: IIRC this string value is like a 'magic' string that is prefixed with the Domain INT id or something like that. I think it would make sense to un-magic this and return an object/struct with the relevant data otherwise it's not clear what this string is supposed to be

src/Umbraco.Web/PublishedCache/ContentCacheAdapter.cs
```diff
@@ -0,0 +1,187 @@
...
+using System.Collections.Generic;
+using System.Xml.XPath;
+using Umbraco.Core;
+using Umbraco.Core.Models.PublishedContent;
+using Umbraco.Core.Xml;
+using Umbraco.Web.PublishedCache.NuCache;
+
+namespace Umbraco.Web.PublishedCache
+{
+    internal class ContentCacheAdapter : IPublishedContentCache, IPublishedContentCache2
```
Shazwazza, line 11: Doesn't need to implement both since IPublishedContentCache2 implements IPublishedContentCache?

src/Umbraco.Web/PublishedCache/ContentCacheAdapter.cs
```diff
@@ -0,0 +1,187 @@
...
+    {
+        private readonly IPublishedCache2 _contentCache;
+        private readonly ContentCache _contentCache1;
+        private readonly IContentRouter _contentRouter;
+        private readonly IDomainCache _domainCache;
+
+        public ContentCacheAdapter(ContentCache contentCache, IContentRouter contentRouter,IDomainCache domainCache)
```
Shazwazza, line 18: Should this be `ContentCache` or `IPublishedCache2` ?

src/Umbraco.Web/PublishedCache/IContentRouter.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="preview">A value indicating whether to consider unpublished content.</param>
+        /// <param name="route">The route</param>
+        /// <param name="hideTopLevelNode">A value forcing the HideTopLevelNode setting.</param>
+        /// <returns>The content, or null.</returns>
+        /// <remarks>
+        /// <para>A valid route is either a simple path eg <c>/foo/bar/nil</c> or a root node id and a path, eg <c>123/foo/bar/nil</c>.</para>
+        /// <para>If <param name="hideTopLevelNode" /> is <c>null</c> then the settings value is used.</para>
+        /// <para>The value of <paramref name="preview"/> overrides defaults.</para>
+        /// </remarks>
+        ContentRoutingResult GetIdByRoute(IPublishedCache2 currentSnapshot,bool preview, string route, bool? hideTopLevelNode, string culture);
```
Shazwazza, line 20: IContentRouter is a singleton and so is `IPublishedCache2` so this can just be ctor injected?

Shazwazza (changes requested):
I've left some more comments inline. Decided to try to write down the changes to see an overview so here's how it goes:

* IContentRouter (new) (singleton), responsible for:
  * resolving an document Id based on a route (the most common routing)
  * getting a qualified route for a document Id
  * more or less replaces IPublishedContentCache2 logic for routing logic in ContentFinderByUrlAlias
  * For other types of routing requirements, those will continue to fallback to IContentFinders which may traverse the cache/tree
* IPublishedCache2 (scoped)
  * Is just the content cache, things like GetById and also resolving by Xpath
* ContentCacheAdapter (new) (internal) (scoped)
  * implements IPublishedContentCache2  (the routing interface parts but is now legacy with these changes ... probably should be removed in future version)
  * wraps functionality of both IPublishedCache2 (ContentCache) and IContentRouter
  * used by published snapshot, a new one of these is created per PublishedSnapshotElements
* ContentCache (internal) (scoped)
  * Implementation of IPublishedContentCache2
  * Previously responsible for routing (i.e. IPublishedContentCache2 )
  * Previously responsible for caching the route results based on the cache instances passed to it
* PublishedSnapshotService
  * Manages 'snapshots' (like a 3d dictionary)
  * responsible for creating the ContentCacheAdapter/ContentCache
  * Each level of the dictionary (snapshot) would have it's own snapshot cache and potentially share the previous elements cache which were passed to the ContentCache instance created for this snapshot  
  
Where does caching fit in? I think it could just be put back in more or less the same was as it was but inside the Adapter?

src/Umbraco.Web/PublishedCache/PublishedCacheBase.cs
```diff
@@ -100,5 +100,7 @@ public virtual IEnumerable<IPublishedContent> GetByContentType(IPublishedContent
                 .SelectMany(x => x.DescendantsOrSelf())
                 .Where(x => x.ContentType.Id == contentType.Id);
         }
+
+        public abstract void Dispose();
```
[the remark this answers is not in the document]
nzdev: IPublishedCache2 now inherits from IDisposable. So the base class needs to abstractly implement IDisposable

src/Umbraco.Web/PublishedCache/IContentRouter.cs
nzdev: _contentRouter.GetIdByAlias(frequest.UmbracoContext.PublishedSnapshot.Content is IPublishedContentCache so this method needs to take IPublishedContentCache  I believe

src/Umbraco.Web/PublishedCache/IContentRouter.cs
nzdev: Good catch. I'll remove.

src/Umbraco.Web/PublishedCache/IContentRouter.cs
nzdev: I don't think IPublishedCache2  is a singleton.

src/Umbraco.Web/PublishedCache/ContentCacheAdapter.cs
nzdev: Introduced IPublishedCache3

src/Umbraco.Web/PublishedCache/ContentCacheAdapter.cs
nzdev: Changed as suggested

src/Umbraco.Web/PublishedCache/IContentRouter.cs
nzdev: Introduced RouteByIdResult, though it's being 'magic''d again in ContentCacheAdaptor

src/Umbraco.Web/PublishedCache/NuCache/ContentCache.cs
nzdev: Implemented in ContentCacheAdapter

src/Umbraco.Web/PublishedCache/ContentCacheContentRouter.cs
nzdev: Implemented caching

src/Umbraco.Web/PublishedCache/IContentRouter.cs
Shazwazza: Ah I see. Wonder if it makes sense to (just in case) make these the latest numbered interfaces to-date, like IPublishedCache3, IPublishedContentCache2 (man that's ugly! will be good to clean this up in netcore branch)

src/Umbraco.Web/PublishedCache/IContentRouter.cs
Shazwazza: yep you are correct

src/Umbraco.Web/PublishedCache/ContentCacheAdapter.cs
```diff
@@ -0,0 +1,218 @@
...
+            var cache = (preview == false || PublishedSnapshotService.FullCacheWhenPreviewing) ? _elementsCache : _snapshotCache;
+            var key = NuCache.CacheKeys.ContentCacheRouteByContent(contentId, preview, culture);
+            return cache.GetCacheItem<string>(key, () =>
+            {
+                var result = _contentRouter.GetRouteById(_contentCache, _domainCache, preview, contentId, culture);
+                if (result.Outcome == RoutingOutcome.NotFound)
+                {
+                    return null;
+                }
+                return (result.DomainId?.ToString(CultureInfo.InvariantCulture) ?? "") + result.Url;
```
Shazwazza, line 169: Guess we really need code comments/docs for what this is. We don't have them anywhere that I can see :( I'll update the current source with something. I'll make another internal note for this to clean up in netcore too since we don't need to worry about breaking changes and can hopefully just deal with RouteByIdResult instead.

src/Umbraco.Web/PublishedCache/ContentCacheContentRouter.cs
```diff
@@ -139,7 +134,7 @@ private string GetRouteByIdInternal(IPublishedCache2 publishedCache, IDomainCach
             }
 
             // at this point this will be the urlSegment of the root, no segment indicates this is not published when this is a variant
-            if (urlSegment.IsNullOrWhiteSpace()) return null;
+            if (urlSegment.IsNullOrWhiteSpace()) return new RouteByIdResult(RoutingOutcome.NotFound);
```
Shazwazza, line 137: This could be a static instead of a new struct like return `RouteByIdResult.NotFound;`

src/Umbraco.Web/PublishedCache/IContentRouter.cs
Shazwazza: Yep, we can remove the magic in netcore

src/Umbraco.Web/PublishedCache/IContentRouter.cs
nzdev: Currently it makes sense to me to support the older interface due to the callers not requiring the newer interfaces.

Issue #8919, opened 2020-09-16
8.8RC: The login screen shows 404s on two icons


nul800sebastiaan:
The [login button](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/buttons/umb-button.html#L3) is requesting two icons in the login screen, but the icon endpoint can only be used when you're logged in, so right now we're getting 404s in the console: 

![image](https://user-images.githubusercontent.com/304656/93321197-388bb080-f812-11ea-849c-8dfdfec12b58.png)

Suggestion: the `umb-icon` should lazy load icons when they're needed, these two icons are not needed on the login screen so we never need to load them there.

---
_This item has been added to our backlog [AB#8305](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8305)_

ronaldbarendse:
Why is this controller/action authenticated in the first place? And secondly, why is it used at all? Isn't it a lot easier to default to a specific path if the `umb-icon` doesn't specify a full path (e.g. including a slash or extension) and directly load that SVG file?

That would also make it a lot easier to add your own icons in packages, as you don't need to put the SVG files in a specific path, make the controller extensible and/or account for duplicate names...

ronaldbarendse:
@bergmania Can you give a follow-up on my previous comment?

IMHO we shouldn't need an `IIconService` for this, as it introduces more problems than it solves: it's currently basically a wrapper to a physical directory with SVG files (without correct exception handling or caching) and any attempt to replace this service might break icons that are used in the back-office.

As the icons are only used in the back-office UI, having an AngularJS resource/controller should be enough, right? And as already mentioned, this could directly load the SVG data from a specific URL...

nul800sebastiaan:
You're (partially) right, I've [moved the interface to `Umbraco.Core.Services`](https://github.com/umbraco/Umbraco-CMS/commit/fe3f60412208473fc94c899917036028788262c0), so we can break it if we have to https://our.umbraco.com/Documentation/Development-Guidelines/breaking-changes#non-breaking

I don't foresee many changes to this one in the future, but better save than sorry!

There will be more code in this service to deal with looking up icons from different directories than the default, so it's not just a path that people can specify. Makes total sense to me to have an interface that we can use to mock in unit tests. 

And I sense you're not happy with the current implementation, we'd love some help of course! 😄

PR #8930, opened 2020-09-17
Inject icons used in login screen to the runtime icon cache

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/8919

As you are required to be authorized to retrieve icons through the getIcon end-point, we cannot retrieve icons on the client-side when the user is not authorized. Meaning that the icons used in the login view cannot be loaded in the same manner.
To fix this we will inject the two known icons into the runtime cache through the razor document, which is allowed to retrieve the icons no matter authorization.

This Draft PR does not retrieve the icons from disc, meaning that we have to replace line 128 and 129 in Default.cshtml with razor code that gets the correct SVG.

This draft PR solve the client-side part and needs some Razor work to be done.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/iconhelper.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/iconhelper.service.js
@@ -209,7 +209,7 @@ function iconHelper($http, $q, $sce, $timeout, umbRequestHelper) {
                             ,'Failed to retrieve icon: ' + iconName)
                         .then(icon => {
                             if(icon) {
-                                var trustedIcon = this.defineIcon(icon.name, icon.SvgString);
+                                var trustedIcon = this.defineIcon(icon.Name, icon.SvgString);
 
                                 liveRequests = _.filter(liveRequests, iconRequestPath);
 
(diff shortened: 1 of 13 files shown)


bergmania:
@nielslyngsoe  is it expected that we make lots of requests to the icon service?

image

src/Umbraco.Web.UI.Client/src/common/services/iconhelper.service.js
```diff
@@ -209,15 +209,11 @@ function iconHelper($http, $q, $sce, $timeout, umbRequestHelper) {
                             ,'Failed to retrieve icon: ' + iconName)
                         .then(icon => {
                             if(icon) {
-                                var trustedIcon = {
-                                    name: icon.Name,
-                                    svgString: $sce.trustAsHtml(icon.SvgString)
-                                };
-                                this._cacheIcon(trustedIcon);
+                                var trustedIcon = this.defineIcon(icon.name, icon.SvgString);
```
bergmania, line 212: This leads to errors
```suggestion
                                var trustedIcon = this.defineIcon(icon.Name, icon.SvgString);
```

PR #8932, opened 2020-09-17
NetCore: Update NuGet packages to latest versions

elit0451:
- This PR contains 42 NuGet packages updated to their latest version. 
The last ones left are related to **Examine.***, **Umbraco.Examine.Lucene**, **Umbraco.Tests.Benchmarks**, or **Umbraco.Web.UI** and **Umbraco.Web** only.

The patch update on _System.Threading.Tasks.Extensions_ NuGet failed many tests with the error: *The located assembly's manifest definition does not match the assembly reference*. So, after some research, it is because the *Moq* package uses this as an underlying dependency and there is a conflict in the updated version and the one used in *Moq*.

- Nuspec files are also updated with the correct version.

---
### Note: 
I added: 
- `<dependency id="Serilog.Filters.Expressions" version="[2.1.0,2.999999)" />` to **build/NuSpecs/UmbracoCms.Core.nuspec**
- `<dependency id="Microsoft.CodeAnalysis.CSharp" version="[3.7.0,3.999999)" />` to **build/NuSpecs/UmbracoCms.Web.nuspec**

because those packages were updated but were not present in our nuspec files.

Perhaps the same should be done for **Microsoft.AspNet.SignalR.Core** (UmbracoCms.Web.nuspec) ?


bergmania:
`Microsoft.AspNet.SignalR.Core` should not be added.. We need to migreate this to aspnetcore signalR, and part of that task will be to update it

Updates looks good.. Now on to testing

PR #8933, opened 2020-09-18
Block List - hide add-content button when maximum items is reached

### LINKED ISSUE #8894: Block List editor Amount Range not applied
When creating a block list editor (following the documentations), I added 0 - 1 in the Amount range fields. However, I was able to add more than one Block into the page.

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #8894 

### Description
When the block list has maximum items, the add content button should not be shown when the number of blocks is equal to the maximum permitted - this PR makes that happen. Means that editors won't be able to add 10 blocks then get a nasty surprise on saving, when the validation says no, too many.

Test by setting max amount to 1, block list will hide add-content button when one block exists.

If the maximum amount is reduced (say from 5 to 2), the UI will display only the valid number of blocks (ie 3, 4 and 5 will be discarded, 1 and 2 will be displayed). Given that, the validation messages for exceeding the maximum number of blocks will never be shown (but have not been changed as part of this PR). 

I still think it would be useful to show a count on the block list component so editors are aware of how many blocks are permitted, but this is a one-line PR that's taken long to write about than to make the change so I'm stopping now.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/validation/showvalidationonsubmit.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/validation/showvalidationonsubmit.directive.js
@@ -11,6 +11,7 @@
             link: function (scope, element, attr, ctrl) {
 
                 var formMgr = ctrl.length > 1 ? ctrl[1] : null;
+                const cssHide = 'ng-hide';
 
                 //We can either get the form submitted status by the parent directive valFormManager
                 //or we can check upwards in the DOM for the css class... lets try both :)
@@ -18,17 +19,17 @@
                 //reset the status.
                 var submitted = element.closest(".show-validation").length > 0 || (formMgr && formMgr.showValidation);
                 if (!submitted) {
-                    element.hide();
+                    element[0].classList.add(cssHide);
                 }
 
                 var unsubscribe = [];
 
                 unsubscribe.push(scope.$on("formSubmitting", function (ev, args) {
-                    element.show();
+                    element[0].classList.remove(cssHide);
                 }));
 
                 unsubscribe.push(scope.$on("formSubmitted", function (ev, args) {
-                    element.hide();
+                    element[0].classList.add(cssHide);
                 }));
 
                 //no isolate scope to listen to element destroy
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umb-block-list-property-editor.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umb-block-list-property-editor.html
@@ -10,6 +10,7 @@
 
                 <button type="button"
                         class="btn-reset umb-block-list__block--create-button"
+                        ng-hide="vm.validationLimit.max && vm.layout.length >= vm.validationLimit.max"
                         ng-click="vm.showCreateDialog($index, $event)"
                         ng-controller="Umbraco.PropertyEditors.BlockListPropertyEditor.CreateButtonController as inlineCreateButtonCtrl"
                         ng-mousemove="inlineCreateButtonCtrl.onMouseMove($event)">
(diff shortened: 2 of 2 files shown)


ronaldbarendse:
Will this also hide the add button in-between blocks? And what happens to the `BlockListModel` that's returned from the PVC, will that also get truncated to the maximum allowed items?

IMHO the editor should show all blocks when the maximum is lowered, so you can remove the ones you don't want anymore (or get the same validation message as we currently see).

Also, having the possibility to add more blocks than the maximum, allows you to change a block type by adding another, copy/pasting the content and then delete the old block.

Another thing to consider is the copy/paste all items property action: does this also truncate to the maximum items? If it does, this has the same issue as lowering the maximum...

nathanwoulfe:
@ronaldbarendse you've considered this in a lot more detail than I did... 👍 

I don't know what happens to the excess blocks - if I set max=10, add 10, then change max = 5, the UI renders the first 5 blocks, but I haven't checked what happens if the max is changed back to 10. Ideally, the original 10 blocks would be visible again (and in that case, like you suggested, they should be visible in the UI at all times, but marked as not available for front-end display).

However, does that open the door to editors creating dozens of blocks in a list with a low limit, then using the ordering to essentially disable blocks? Is that an issue? For example, I could have a collection of 30 blocks in a list with max=5, knowing only the first 5 will ever be rendered.

All the original min/max validation is still in place, so I'd expect this would manage PVC, other add button, clipboard etc. Will check though, and see what can be improved.

nathanwoulfe:
@ronaldbarendse, I checked a couple of things:
- If I reduce the maximum to a number lower than the current number of blocks (ie 3 created, max changed to 1), I still see all the blocks in the backoffice. I can't save/publish the page until validation passes. Given that, the rendered page will show the invalid number of blocks, since that's the last published state. Not sure what I was seeing in my previous comment, UI retains all blocks.
- floating add button was still visible, I have a change pending to fix that (hide it when block count >= validation.max)

I still agree though that it makes sense to be able to add additional blocks beyond the validation maximum, but for that to work there needs to be a visible count/indicator that a limit exists, and that it has (or has not) been reached, otherwise users can potentially add a stack of blocks that can't be saved.

nathanwoulfe:
Based on @nielslyngsoe's comments on the issue, latest commit always shows the add-content button, but also displays the validation text when the content is invalid. 

To make that work, I've modified the show-validation-on-submit directive which was previously using `.hide()` and `.show()` to control the message visibility, but that doesn't play nice with AngularJs. I've changed the directive to instead add/remove the `ng-hide` class name, which has the same end result (sets the element visibility to `hidden!important`) but means we can still control the element visisbility via ng-show to display it when the list is invalid.

nielslyngsoe:
Hi @nathanwoulfe thanks for taking the effort to provide a solution to this problem. Looking at the code I spotted that we just need to remove show-validation-on-submit directive from the messages and things work as they should.
Therefor I will close this one to make a fix in a new PR.

nathanwoulfe:
@nielslyngsoe all good, but still worth considering the changes in the directive - not the best idea to be mixing jQuery .hide() in with AngularJs ng-show/ng-hide as it will cause specificity issues elsewhere. Better to toggle the ng-hide classname to keep it all angular.

nielslyngsoe:
@nathanwoulfe Oh, yes! that would be good. Can I ask you to minimize the scope of this PR to that?

nathanwoulfe:
@nielslyngsoe absolutely. I'll make it happen.

nielslyngsoe:
@nathanwoulfe thanks a lot — I can easily end up spending most my time just managing PRs, so minimizing that work is always nice.

nathanwoulfe:
@nielslyngsoe I'll close this one and open a clean PR - probably better to keep the history clean.

PR #8934, opened 2020-09-18
Netcore/feature/healthcheck replaceconfiglogic

emmagarland:
NB:l I wanted to stop before going too far - more fixes can follow afterwards

I will do a file move at the end in a separate PR to avoid loss of history, as some health checks are in a different folder.

### Prerequisites

Will add issue if this PR covers it 

### Description

* Mild refactoring but overall keeping it the same, except for injecting IOptionsMonitor<T> for config, and JSON parsing instead of XML (after discussing with team) 
* No longer using the IConfigManipulator to update settings as required - this PR does not allow for fixing health check settings. That can be a future discussion, but for various reasons we won't be updating the config values here.
* Health check tests currently unaffected - still need to increase coverage
* Edited health-checks to make them .NET Core compatible, switching from config to appsettings.json. 
* There will also be new ones needed I expect, as well as some redundant ones since the move to .NET core
* We will need to edit the xml translation files to reflect the move from config and XPath to JSON settings. At the moment, some of the health-check messages won't reflect this. A good community up-for-grabs for future.
* Have not tested the other side of health check, e.g. the HealthCheckNotifier, yet


src/Umbraco.Core/Configuration/HealthChecks/CompilationDebugCheck.cs
```diff
@@ -1,33 +1,39 @@
...
-            : base(textService, hostingEnvironment, logger)
-        { }
+        private readonly HostingSettings _hostingSettings;
 
-        public override string FilePath => "~/Web.config";
+        public CompilationDebugCheck(ILocalizedTextService textService, ILogger logger,IConfigurationService configurationService, IOptions<HostingSettings> hostingSettings)
+            : base(textService, logger, configurationService)
+        {
+            _hostingSettings = hostingSettings.Value;
+        }
```
bergmania, line 20: Here do we need to inject IOptionsMonitor<HostingSettings> to ensure the value is updated on the fly

Also save the  IOptionsMonitor<HostingSettings> in the field and below use `_hostingSettings.CurrentValue.Debug` instead

src/Umbraco.Core/Configuration/HealthChecks/ConfigurationService.cs
```diff
@@ -1,113 +1,65 @@
...
+                }
 
-                if (node.NodeType == XmlNodeType.Element)
-                    node.InnerText = value;
-                else
-                    node.Value = value;
-
-                xmlDocument.Save(_configFilePath);
-                return new ConfigurationServiceResult { Success = true };
+                _configManipulator.SaveConfigValue(itemPath, value);
```
bergmania, line 54: I dont this this can be used directly. There are multiple possible issues..

1) This do not work if the value do not exists in the config. E.g using default value. This is also why we "merge" json objects when we insert connection string.
1) This only saves to the json file lowest in the config hierarchy (appsettings.json). But the value could be overridden in multiple other sources higher in the hierarchy (e.g. appsettings.Development.json).

Not sure how we can do, or maybe we should just remove the "Fix" option from some of the healthchecks?

src/Umbraco.Core/Configuration/HealthChecks/MacroErrorsCheck.cs
```diff
@@ -1,27 +1,42 @@
...
-        { }
+        private readonly ILocalizedTextService _textService;
+        private readonly ILogger _logger;
+        private readonly ContentSettings _contentSettings;
 
-        public override string FilePath => "~/Config/umbracoSettings.config";
-
-        public override string XPath => "/settings/content/MacroErrors";
+        public MacroErrorsCheck(ILocalizedTextService textService, ILogger logger,
+            IConfigurationService configurationService, IOptions<ContentSettings> contentSettings)
```
bergmania, line 22: Also here we need to use IOptionMonitor<T>

src/Umbraco.Core/Constants-Configuration.cs
```diff
@@ -11,11 +11,21 @@ public static class Configuration
...
-            public const string ConfigModelsBuilderPrefix = ConfigPrefix+"ModelsBuilder:";
-            public const string ConfigRuntimeMinification = ConfigPrefix+"RuntimeMinification";
-            public const string ConfigRuntimeMinificationVersion = ConfigRuntimeMinification+":Version";
+            public const string ConfigContentPrefix = ConfigPrefix + "Content:";
+            public const string ConfigModelsBuilderPrefix = ConfigPrefix + "ModelsBuilder:";
+            public const string ConfigContentNotificationsPrefix = ConfigContentPrefix + "Notifications:";
+            public const string ConfigHostingPrefix = ConfigPrefix + "Hosting:";
+            public const string ConfigCustomErrorsPrefix = ConfigPrefix + "CustomErrors:";
+
+            public const string ConfigRuntimeMinification = ConfigPrefix + "RuntimeMinification:";
```
bergmania, line 22: This should be named "*Prefix" when it ends with ":"

bergmania (changes requested):
Hi @emmagarland 

I just shortly reviewed the PR. And gave some feedback. Hopefully it make sense :)

src/Umbraco.Core/Configuration/HealthChecks/CompilationDebugCheck.cs
emmagarland: Will do!

src/Umbraco.Core/Configuration/HealthChecks/ConfigurationService.cs
emmagarland: Good points. I think it is best if I don't try to fix in this PR, remove the fix option for the next push, and it can be considered again in a future product conversation - as discussed, whether we show a "How to fix" or actually try to write the fix.

emmagarland: This is no longer being referenced, so:

- I've kept the code changes as if we were going to write the values
- I haven't deleted the service as it was pre-existing and created in 2016 for the first health checks

src/Umbraco.Core/HealthCheck/Checks/AbstractSettingsCheck.cs
```diff
@@ -0,0 +1,160 @@
+﻿using System;
```
emmagarland, line 1: This is the same file as AbstractConfigCheck, but renamed (probably shouldn't have done that here)

elit0451 (approved):

bergmania (approved):

PR #8935, opened 2020-09-18
V8/fix tracked references query

dawoe:
While working on the UI improvements for the item tracking feature (#8679) I noticed that all relations types are shown on the info tab.

To reproduce you can take the following scenario's (based on the v8/item-tracking branch) using the default starter kit.

**Scenario 1**
1. Copy a existing page and check relate to original.
2. On the new page you will see the original page in the list of Used in documents

**Scenario 2**
1. Delete a existing page that is under a other page.
2. Locate the deleted page in the recycle bin and check the info tab. You will see the parent of the deleted page under used in documents.

This one is also happening with media

This PR applies filtering on the relation types that are used for filtering. Everything is backwards compatible and covered with unit tests. I discussed this with @Shazwazza


poornimanayar (approved):

poornimanayar:
Hi @dawoe ,

Many thanks for this. I have tested this and it works!. Can see that the bugs have been fixed. Merging it in!

Poornima

Issue #8654, opened 2020-08-18
Use Test Environment Settings in Unit and Integration Tests.


JimBobSquarePants:
In ASP.NET Core environmental settings are loaded without transforms. 

The environment itself is determined by the present of the environmental setting [`ASPNETCORE_ENVIRONMENT`](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-3.1). By default three values are supported. _Development, Staging_, and _Production_. 

It is often a requirement though to use custom setting during testing. This is possible via `IWebHostBuilder.ConfigureAppConfiguration` used currently in the source...

https://github.com/umbraco/Umbraco-CMS/blob/43efee9647ed8f2eaab3ec2b71d3dd3eb46b9017/src/Umbraco.Tests.Integration/TestServerTest/UmbracoWebApplicationFactory.cs#L52-L59

...but there's a better way which allows test settings to sit in an environmental specific `appsettings.Testing.json` file in the same manner as other environments.

This is massively beneficial for value discovery for developers contributing to the library since values are easy to find and compare.

Implementing it is trivial. First define an extension environment.

```c#
/// <summary>
/// Extension methods for <see cref="IHostEnvironment"/>.
/// </summary>
public static class HostEnvironmentExtensions
{
    /// <summary>
    /// The environment name for the "Testing" environment.
    /// </summary>
    public const string Testing = nameof(Testing);

    /// <summary>
    /// Checks if the current host environment name is <see cref="Testing"/>.
    /// </summary>
    /// <param name="hostEnvironment">An instance of <see cref="IHostEnvironment"/>.</param>
    /// <returns>
    /// True if the environment name is <see cref="Testing"/>, otherwise false.
    /// </returns>
    public static bool IsTesting(this IHostEnvironment hostEnvironment)
        => hostEnvironment.IsEnvironment(Testing);
}
```

For integration tests wiring it up is as simple as follows.

```c#
/// <summary>
/// Gives a fixture an opportunity to configure the application before it gets built.
/// </summary>
/// <param name="builder">The <see cref="IWebHostBuilder" /> for the application.</param>
protected override void ConfigureWebHost(IWebHostBuilder builder)
{
    // Use environment settings
    builder.UseEnvironment(HostEnvironmentExtensions.Testing);

    // Use the same startup class as your application.
    builder.UseStartup<Startup>();

    // TODO: Add any other setting like URLs.

    base.ConfigureWebHost(builder);
}
[...]

Issue #4998, opened 2019-03-15
Exception in BPlusTree


zpqrtbnk:
See #4992, one that issue was fixed, the following happened:

~~~~
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. 
Parameter name: length
 at CSharpTest.Net.IO.TransactedCompoundFile.Write(UInt32 handle, Byte[] bytes, Int32 offset, Int32 length)
 at CSharpTest.Net.Storage.BTreeFileStoreV2.Update[T](IStorageHandle handleIn, ISerializer`1 serializer, T node)
 at CSharpTest.Net.Collections.BPlusTree`2.StorageCache.OnItemRemoved(KeyValuePair`2 item)
 at CSharpTest.Net.Collections.LurchTable`2.TryDequeue(Predicate`1 predicate, KeyValuePair`2& value)
 at CSharpTest.Net.Collections.BPlusTree`2.StorageCache.Flush()
 at CSharpTest.Net.Collections.BPlusTree`2.StorageCache.Commit()
 at CSharpTest.Net.Collections.BPlusTree`2.CommitChanges(Boolean requiresLock)
 at Umbraco.Web.PublishedCache.NuCache.ContentStore.Release(WriteLockInfo lockInfo, Boolean commit)
 at Umbraco.Web.PublishedCache.NuCache.ContentStore.Set(ContentNodeKit kit)
 at Umbraco.Web.PublishedCache.NuCache.PublishedSnapshotService.NotifyLocked(IEnumerable`1 payloads, Boolean& draftChanged, Boolean& publishedChanged)
 at Umbraco.Web.PublishedCache.NuCache.PublishedSnapshotService.Notify(JsonPayload[] payloads, Boolean& draftChanged, Boolean& publishedChanged)
 at Umbraco.Web.Cache.ContentCacheRefresher.Refresh(JsonPayload[] payloads)
 at Umbraco.Core.Sync.ServerMessengerBase.DeliverLocal[TPayload](ICacheRefresher refresher, TPayload[] payload)
 at Umbraco.Core.Sync.ServerMessengerBase.Deliver[TPayload](ICacheRefresher refresher, TPayload[] payload)
 at Umbraco.Core.Sync.ServerMessengerBase.PerformRefresh[TPayload](ICacheRefresher refresher, TPayload[] payload)
[...]

PR #8938, opened 2020-09-21
NetCore: Added validation for configuration

AndyButland:
In this PR I've added validation for the new configuration classes, using `IValidateOptions` (for background, see [this blog post](https://code-maze.com/aspnet-configuration-options-validation/)).

By doing this we catch early, on start-up, if a configuration entry has been made that's not valid.

See what you think about the change I've made to configuration values that were Enums (for example `ContentSettings.MacroErrorsBehaviour`.  Although we can bind these directly, I've chosen to do it via a string such that we can use this validation method - catch all at the same time and use consistent error messaging.  With an enum binding directly we can't validate like this (so an invalid value will still be caught, but in a different way/at a different time).

Relates to #8651


src/Umbraco.Core/Configuration/Models/ContentSettings.cs
```diff
@@ -16,11 +15,25 @@ public class ContentSettings
...
+        // via this string to allow for validation to occur on start-up that the configured value does
+        // match one of the enum values.
+        // Without this it'll fail on first use, and we'd have less control over the error message.
+        internal string MacroErrors { get; set; } = MacroErrorBehaviour.Inline.ToString();
+
+        public MacroErrorBehaviour MacroErrorsBehaviour
+        {
+            get
+            {
+                return Enum.TryParse<MacroErrorBehaviour>(MacroErrors, true, out var value)
```
JimBobSquarePants, line 32: I'm not sold on this. I would want the framework to throw on invalid configuration not set a value for me.

AndyButland: Thanks for taking a look over @JimBobSquarePants.

In practice, the answer to _What happens if I explicitly chose different behavior but spelt it incorrectly?_ would be that this would be caught by the validation I've added in this PR (see `ContentSettingsValidator`).  That's going to make sure that if you've provided a value, it must be one of the accepted enum values, hence you should never get the "assumed" value set here.

But that said, it follows - and I take your point - that when this code is run that validation should have occurred, and so if `TryParse` failed here, it would be an exception (i.e. something has gone wrong if an invalid value hasn't already been caught by the validation), so it's best to throw here rather than silently using a default value.  Have updated.

src/Umbraco.Core/Configuration/Models/ModelsBuilderSettings.cs
```diff
@@ -18,10 +20,23 @@ public class ModelsBuilderConfig
         /// </remarks>
         public bool Enable { get; set; } = false;
 
+        // See note on ContentSettings.MacroErrors
+        internal string ModelsMode { get; set; } = Configuration.ModelsMode.Nothing.ToString();
```
bergmania, line 24: Can these be private like we have for connection string? I'm afraid we will use this value where it is not intended.

src/Umbraco.Core/Configuration/Models/Validation/ContentSettingsValidator.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        private bool ValidateError404Collection(IEnumerable<ContentErrorPage> values, out string message)
+        {
+            return ValidateCollection("Content:Error404Collection", values, "Culture and one and only one of ContentId, ContentKey and ContentXPath must be specified for each entry", out message);
+        }
+
+        private bool ValidateAutoFillImageProperties(IEnumerable<ImagingAutoFillUploadField> values, out string message)
+        {
+            return ValidateCollection("Content:Imaging:AutoFillImageProperties", values, "Alias, WidthFieldAlias, HeightFieldAlias, LengthFieldAlias and ExtensionFieldAlias must be specified for each entry", out message);
```
bergmania, line 42: I really think we need constants for the values like "Content:Imaging:AutoFillImageProperties".
 Maybe build from multiple constants..

src/Umbraco.Tests.Common/Builders/ConfigurationModelBuilders/ContentSettingsBuilder.cs
```diff
@@ -0,0 +1,56 @@
+using System.Collections.Generic;
+using System.Linq;
+using Umbraco.Core.Configuration.Models;
+using Umbraco.Core.Macros;
+
+namespace Umbraco.Tests.Common.Builders
```
bergmania, line 6: Not sure I see the value of the builder, when ContentSettings  has valid and good default values and is a poco?

src/Umbraco.Web.UI.NetCore/appsettings.json
```diff
@@ -24,6 +24,10 @@
         "DisableKeepAliveTask": false,
         "KeepAlivePingUrl": "{umbracoApplicationUrl}/api/keepalive/ping"
       },
+      "ModelsBuilder": {
+        "ModelsMode": "Nothing",
```
bergmania, line 28: I agree this should be nothing as default value

bergmania (commented):
Looks very good, but I have a few very common comments..

src/Umbraco.Core/Configuration/Models/ContentSettings.cs
JimBobSquarePants: Doesn't that now mean double validation?

You've got code now for a scenario that should never present itself and that a unit test should prevent someone ever changing the code to avoid the default validation.

src/Umbraco.Core/Configuration/Models/ContentErrorPage.cs
```diff
@@ -1,15 +1,30 @@
...
+
+        public bool HasContentXPath => !string.IsNullOrEmpty(ContentXPath);
+
+        [Required]
         public string Culture { get; set; }
+
+        internal override bool IsValid()
+        {
+            return base.IsValid() &&
+                ((HasContentId ? 1 : 0) + (HasContentKey ? 1 : 0) + (HasContentXPath ? 1 : 0) == 1);
```
JimBobSquarePants, line 27: So... We need exactly one of these states? 

Personally I would refactor to only ever allow one type of configuration. Too many options can be confusing.

src/Umbraco.Core/Configuration/Models/ContentImagingSettings.cs
```diff
@@ -17,17 +14,8 @@ public class ContentImagingSettings
             }
         };
 
-        public IEnumerable<string> ImageFileTypes { get; set; } = new[] { "jpeg", "jpg", "gif", "bmp", "png", "tiff", "tif" };
-
-        public IEnumerable<IImagingAutoFillUploadField> AutoFillImageProperties { get; set; } = DefaultImagingAutoFillUploadField;
+        public string[] ImageFileTypes { get; set; } = new[] { "jpeg", "jpg", "gif", "bmp", "png", "tiff", "tif" };
```
JimBobSquarePants, line 17: Just so everyone knows. ImageSharp does not support Tiff yet and ImageProcessor is not supported on NET Core (And is a complete (_read dead_) project). 

Image format configuration should be bound to whatever the lowest common denominator of the various potential solutions available not hard coded.

src/Umbraco.Core/Configuration/Models/ModelsBuilderSettings.cs
JimBobSquarePants: The json parser won't be able to see private fields most likely.

src/Umbraco.Core/Configuration/Models/NuCacheSettings.cs
```diff
@@ -2,6 +2,6 @@
 {
     public class NuCacheSettings 
     {
-        public string BTreeBlockSize { get; set; }
+        public int? BTreeBlockSize { get; set; }
```
JimBobSquarePants, line 5: Does NuCache provide a default value? Does it already validate input and clamp it to a range? Is this something that should even be set?

src/Umbraco.Tests.Common/Builders/ConfigurationModelBuilders/ContentSettingsBuilder.cs
JimBobSquarePants: I don't actually see the benefit in ANY builders. The whole point of using `Options<T>` is that they are injected directly into services. 

Generally you want each type to be as simple as possible with sensible defaults so that external configuration covers edge case use only.

AndyButland: It's partly to help with creating invalid setups so we can unit test that the validators work (see e.g. ContentSettingsValidationTests).  And partly just for consistency with what's gone before (we have builders for most of the configuration models - so probably we should have for all or none of them).

src/Umbraco.Core/Configuration/Models/ModelsBuilderSettings.cs
AndyButland: Unfortunately not - as I'm looking to do the validation on these string representations, that validator classes need to be able to read them.  As currently do the builders used in the unit tests..

src/Umbraco.Core/Configuration/Models/Validation/ContentSettingsValidator.cs
AndyButland: Yes, agreed.  Have updated to populate from a combination of constants and values derived from `nameof` expressions.

src/Umbraco.Web.UI.NetCore/appsettings.json
AndyButland: Just to make sure it's clear here, I had to change this as I found out that currently in `netcore/netcore` we aren't loading the configuration in appSettings (as it's looking for it in the wrong place).  When I fixed that, and `PureLive` was getting picked up, I was getting errors so changed the default to `Nothing`.

JimBobSquarePants: Is `Nothing` one of the enum values? Not sold on the naming here or the need for separate values.

`None` or `Disabled` would be a better choice.

src/Umbraco.Core/Configuration/Models/ContentSettings.cs
AndyButland: Unfortunately I've had to revert this - it seems that this property is getting evaluated at start-up before the validation occurs, so an invalid entry in configuration will cause this to throw an exception before we get to the validation.

Arguably we could just not validate enum settings and rely on this built-in behaviour if we try to bind directly from configuration to an enum value on our configuration POCO.  Which would also mean we could get rid of the extra string field.  Only reason not to is to be able to handle the validation all in the same way, with similar error messages etc. - which I quite like, but perhaps it's not adding that much value.

src/Umbraco.Core/Configuration/Models/ContentErrorPage.cs
AndyButland: That seems to be the current expectation yes (have implemented this from a `TODO`).  Would suggest that refactor if done should be outside of this PR (and I guess there's potential that different solutions will have used different "identifiers", and as such we should support them all).

src/Umbraco.Tests.Common/Builders/ConfigurationModelBuilders/ContentSettingsBuilder.cs
JimBobSquarePants: Seems like a lot of ceremony. 

In fitting with ASP.NET Core I would build something around a  `ConfigurationBuilder` that parses a json source and creates an `IConfiguration` instance from that source. Much like the approach I demoed to create test configuration instances in #8654

src/Umbraco.Core/Configuration/Models/NuCacheSettings.cs
AndyButland: It does, yes: `4096`.  Not going to pretend I particularly understand it!  But yes, there's some validation of it already under `BTree.GetBlockSize()`.  We could consider migrating this into a configuration validator class too, but I've not done that as it's a little more involved/specific than your typical validation.

src/Umbraco.Core/Configuration/Models/ContentImagingSettings.cs
AndyButland: Agreed, that sounds better, but probably for something outside of this PR.  Perhaps could add to `IImageUrlGenerator` (the abstraction introduced for different "image processors"?

src/Umbraco.Core/Configuration/Models/ContentSettings.cs
JimBobSquarePants: I would say go with the default framework behavior for configuration driven scenarios. 

For code driven configuration I would normally handle property validation within the setter of the property itself without the need for an additional method to call from somewhere else.

src/Umbraco.Core/Configuration/Models/NuCacheSettings.cs
JimBobSquarePants: It seems like the kind of configuration I would leave well alone as setting an arbitrary value without profiling could be harmful.

src/Umbraco.Web.UI.NetCore/appsettings.json
AndyButland: It is yes.  Like the example below, we'd need to consider whether changing the name here that might be in various people's solution config is worthwhile.

src/Umbraco.Core/Configuration/Models/NuCacheSettings.cs
AndyButland: Found a reference to it being made configurable here: https://github.com/umbraco/Umbraco-CMS/issues/4998

JimBobSquarePants: Whoa!  That means there’s an issue requiring a fix.

src/Umbraco.Web.UI.NetCore/appsettings.json
JimBobSquarePants: I very much doubt there’ll be an upgrade path from V8 that allows preserving exisiting config

src/Umbraco.Tests.Common/Builders/ConfigurationModelBuilders/ContentSettingsBuilder.cs
bergmania: I still sees alot of benefits of some of the builders, just not for configuration (or any other POCO with default ctor and good default values).
E.g. `DataTypeBuilder` still makes it easier to create DataTypes.

Not sure I see the benefits of the builder in `ContentSettingsValidationTests` as you still use a "builder" method too. That method could just be replaced with this right?

```cs
 private static ContentSettings BuildContentSettings(string macroErrors = "inline", string culture = "en-US", string contentXPath = "", string autoFillImagePropertyAlias = "testAlias")
        {
            return new ContentSettings()
            {
                MacroErrors = macroErrors,
                Error404Collection = new[]
                {
                    new ContentErrorPage()
                    {
                        Culture = culture,
                        ContentId = 1,
                        ContentXPath = contentXPath
                    },
                },
                Imaging = new ContentImagingSettings()
                {
                    AutoFillImageProperties = new ImagingAutoFillUploadField[]
                    {
                        new ImagingAutoFillUploadField()
                        {
                            Alias = autoFillImagePropertyAlias
                        },
                    }
                }
            };
        }
```

src/Umbraco.Web.UI.NetCore/appsettings.json
bergmania: The reason you had errors is most likely because PureLive and Precompiled views are not a good combi :)  So if you use PureLive you will also have to disable the precompiled views.

src/Umbraco.Core/Configuration/Models/ContentImagingSettings.cs
bergmania: Agree this value is very related to the implementation of `IImageUrlGenerator`

src/Umbraco.Web.UI.NetCore/appsettings.json
AndyButland: _I very much doubt there’ll be an upgrade path from V8 that allows preserving exisiting config_ - no, agreed.  I was more meaning though that if someone has this config in the existing XML, then it'll be a copy/paste process to transfer it into JSON.  If the names change, then it's a little bit more.

src/Umbraco.Tests.Common/Builders/ConfigurationModelBuilders/ContentSettingsBuilder.cs
AndyButland: OK, I'll at an update later or tomorrow to remove the builders we've introduced for these configuration models.  Wasn't really seeing why they were considered a good idea for domain model classes, but not for these, but I see the distinction you are drawing, and makes sense.

src/Umbraco.Core/Configuration/Models/ContentSettings.cs
AndyButland: OK - will do this (update coming later or tomorrow).  Possibly I'm getting a little too attached to the custom code... but it's probably not adding that much value, and the workarounds introduced are a little confusing.

src/Umbraco.Tests.Common/Builders/ConfigurationModelBuilders/ContentSettingsBuilder.cs
AndyButland: Updated now to that effect.

src/Umbraco.Core/Configuration/Models/ContentSettings.cs
AndyButland: Updated now to that effect.

src/Umbraco.Core/Configuration/Models/ContentImagingSettings.cs
AndyButland: Have created a new PR for this amend: https://github.com/umbraco/Umbraco-CMS/pull/8948

bergmania:
@JimBobSquarePants  I merged this PR as both Andy and I thinks all the comments related to this PR is resolved.  Feel free to write me, if you disagree 😄

JimBobSquarePants:
@bergmania I'll be honest and say I really do not like the introduction of this pattern. 

I think It's solving a problem that I don't think truly exists and introducing massive swathes of undocumented code in order to do so.

PR #8953, opened 2020-09-23
Re-use svg opener on the upload propertytype

nul800sebastiaan:
Following the update in https://github.com/umbraco/Umbraco-CMS/pull/6137 we didn't upload the generic fileupload control, that is fixed in this PR.

You can test this by adding an upload datatype to one of your document types, uploading both a non-svg image and an svg image to that document and then click on the image. The svg image will open in a new tab and when you inspect it with your browser's dev tools you will see it's been wrapped in an `img` tag.

image

This is reusing the exact same approach as was used in the umb-media-info directive (see previous PR https://github.com/umbraco/Umbraco-CMS/pull/6137).


nul800sebastiaan:
Note: this will need to be fixed in v8 as well, not sure if a cherry pick will work.

PR #8960, opened 2020-09-25
create enviroment configuration through install script

nielslyngsoe:
Ask user about username and password for the Umbraco User to be used for Cypress as part of the installation script.

This is how it will look running the npm installation script, which will result in an environment configuration file that contains the username and password.

image

Test notes:

- Test that "npm i" askes for user credentials if NO cypress.env.json exist.
- Test that "npm i" DOES NOT askes for user credentials if the cypress.env.json already exist.
- Test that you can change credentials by running "npm run config"
- Test that baseUrl does not get set when leaving prompt question 'baseUrl' empty.
- Test that defining baseUrl in prompt gets set in cypress.env.json
- Test that cypress uses the environment baseUrl, if defined.
- Read the readme file


bergmania:
Changes looks good and my tests works.. 

For some (unrelated) reason the build server failed. I'm just re running before merging..

Issue #8359, opened 2020-06-30
Nodes get wrong sortorder


joepvtl:
When you first create a media item then a new content item, the content item gets the wrong sort order. It gets fixed after an AppPool recycle. We made this conclusion because we added the SqlProfiler when creating a new content node and saw that it looked at the NodeObjectId of an media item to determine the SortOrder, but it did create the content correctly.
There aren't any error logs before this bug occurs.

## Umbraco version

I have tested this on 8.6.2

Reproduction
------------

### Steps to reproduce

Content gets wrong SortOrder:
1. Setup a fresh install of umbraco and use the starter kit.
2. Create a new media item.
3. Create a new conten item under the 'About us' node.

Media gets wrong SortOrder:
1. Setup a fresh install of umbraco and use the starter kit.
2. Create a new content item.
3. Create a new media item under the 'People' node.

After the last step you can see that the node doesn't get placed at the bottom where it should be. When you look in the database you can see that the node got a SortOrder 0, instead of the expected 2.

### Expected result

Nodes get the correct sortorder.

### Actual result

Nodes get the wrong sortorder.

---
_This item has been added to our backlog [AB#7545](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/7545)_

nul800sebastiaan:
Thanks @joepvtl - I fail to see how this could possibly get fixed after the app pool recycles? It's not like we go in the database after an app pool recycle and update the sortorder column.. can you clarify?

joepvtl:
Well, I think it has something to do with SqlTemplates. Since they both use the same template key to determine the SortOrder and the template gets put into a dictionary. When the content gets created first the query for determining the SortOrder of content gets put into the dictionary, then the media uses that same key and basically uses the same query.

alexroode:
We are seeing a similar issue. In our case we have a news feed node configured as a List View whose child posts are sorted by their sort order descending, to show the newest children first while still allowing editors to manually sort if needed.  Sometimes our site seems to get into a state where each new child item gets created with sort order 0 instead of incrementing like we expect. In this case editors are forced to manually sort to make them appear before the older children with nonzero sort order, so this causes a problem for us.

If we restart the app pool, then create a new node, it seems to properly increment the next new node's sort order. But we seem to eventually get back to the state described above - I have not been able to pinpoint when this happens.

joepvtl:
[...]

PR #8961, opened 2020-09-25
Fix SQL parameters in GetNewChildSortOrder

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8359.

### Description
Because the `NodeObjectTypeId` of the current implementing repository (`DocumentRepository`, `MediaRepository` and `MemberRepository`) ended up in the SQL template of the base class/method `ContentRepositoryBase.GetNewChildSortOrder()`, the sort order was only correctly returned for the repository that did the first database query (as that cached the SQL template).

This PR moves the `NodeObjectTypeId` to an SQL parameter and removes the `first` parameter from the SQL query (as that was also incorrectly cached in the template). This ensures the sort order is correctly retrieved from the database for all repositories and if none are found, the correct first sort order/number is returned.

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
@@ -910,7 +910,7 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
             var sql = template.Sql(NodeObjectTypeId, parentId);
             var sortOrder = Database.ExecuteScalar<int?>(sql);
 
-            return sortOrder.HasValue ? sortOrder.Value + 1 : first;
+            return (sortOrder + 1) ?? first;
         }
 
         protected virtual NodeDto GetParentNodeDto(int parentId)
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
@@ -900,11 +900,16 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
 
         protected virtual int GetNewChildSortOrder(int parentId, int first)
         {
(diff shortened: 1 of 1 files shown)


src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
```diff
@@ -900,28 +901,43 @@ protected virtual string EnsureUniqueNodeName(int parentId, string nodeName, int
...
+                .Select("MAX(sortOrder)")
+                .From<NodeDto>()
+                .Where<NodeDto>(x => x.NodeObjectType == SqlTemplate.Arg<Guid>("nodeObjectType") && x.ParentId == SqlTemplate.Arg<int>("parentId"))
             );
 
-            return Database.ExecuteScalar<int>(template.Sql(new { parentId })) + 1;
+            var sql = template.Sql(NodeObjectTypeId, parentId);
+            var sortOrder = Database.ExecuteScalar<int?>(sql);
+
+            return sortOrder.HasValue ? sortOrder.Value + 1 : first;
```
ronaldbarendse, line 913: This can also be written a little more verbose:
```suggestion
            return (sortOrder + 1) ?? first;
```

Because unary operators always return `null` if one or both of the operators are `null` (see [lifted operators](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types#lifted-operators)), this gives the same result 👍

ronaldbarendse:
@nul800sebastiaan Any change on reviewing this PR and including it into all upcoming patch releases (8.6.6, 8.7.1 and 8.8.1)? It's causing quite a big issue with content, media and members not getting the right sort order...

nul800sebastiaan:
Sorry, this will have to wait.

mikecp (approved):

mikecp:
Hello @ronaldbarendse ,
Thanks for this great PR, I just went through it and it works perfectly, items get the expected sort order again 👍 
It will probably solve "mysterious sorting logic" issues for quite some people! 💯

PR #8968, opened 2020-09-27
LogController: move Query calls from controller to AuditService, usin…

ageroh:
This is a small contribution, of a larger issue, I should have created a draft PR, but I dont by mistake.


bergmania:
Hi @ageroh..

First of all, we will need to base the PR on the `netcore/netcore` branch, as the LogController is migrated to ASP.NET Core in that. 

I think this is a small step in the right direction, but I still think we should move all the logic from the LogController into the service layer. This could be part of `IAuditService` or preferable part of a new "service" interface, to avoid having all this logic in the controller.

Does it make sense?

PR #8993, opened 2020-09-30
Fix the method to return if the directory doesn't exist

failureflawless:
Fixes it!


emmaburstow (approved):
Happy with it

emmaburstow:
This was already fixed 🤦

PR #8799, opened 2020-09-02
Allow to pass in boolean to umbAutoFocus directive

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
This PR allow to pass in truthy values (0/1, 'true'/'false', true/false) to `umbAutoFocus` similar to `preventEnterSubmit ` and `preventDefault` directives.

Originally added to this PR https://github.com/umbraco/Umbraco-CMS/pull/8639 but moved to a separate PR instead.

PR #9004, opened 2020-10-01
Fix broken auto-focus

kjac:
### Prerequisites

### Description

Seems like #8799 broke the `umb-auto-focus` directive... it now assumes it's enabled by default, causing focus to be applied to all the wrong things out of order.

As an example: The data type selector should automatically be focused when it's opened, so you can start typing right away; but it's not:

umb-auto-focus-before

This PR fixes the directive, so it only applies focus if `umb-auto-focus` is actually meant to be applied (either by supplying `umb-auto-focus` without a value or by using `umb-auto-focus="true"`). Now the data type selector does receive focus automatically:

umb-auto-focus-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbautofocus.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbautofocus.directive.js
@@ -10,19 +10,18 @@ angular.module("umbraco.directives")
                 }
             };
 
-            var enabled = true;
             //check if there's a value for the attribute, if there is and it's false then we conditionally don't
             //use auto focus.
             if (attrs.umbAutoFocus) {
                 attrs.$observe("umbAutoFocus", function (newVal) {
-                    enabled = (newVal === "false" || newVal === 0 || newVal === false) ? false : true;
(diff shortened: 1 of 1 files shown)


bjarnef:
@kjac isn't that similar to `prevent-deault` directive? https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/preventdefault.directive.js

And `prevent-enter-submit` (if the enable/disable actually worked) https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/prevententersubmit.directive.js

It would be great to streamline these three directives to handle enable/disable in same way.

kjac:
@bjarnef you could argue that. The other directives do work like that though, because they're reactive. The auto focus one is proactive, it instantly does it job (applies focus) on load... and thus when assuming it's enabled by default, it applies focus by default before it can digest the `observe`.

poornimanayar (approved):
Hi @kjac ,

Tested this and it works fine!

Poornima

poornimanayar:
I am merging this in as it works well!  Thanks for the work once again @kjac ! ⭐ 

Poornima

PR #9005, opened 2020-10-01
Creating a "System" content type breaks Umbraco

kjac:
### Prerequisites

### Description

Creating a content type named "System" has the quite unintended side effect of breaking Umbraco entirely:

image

This was achieved with ModelsBuilder in `PureLive` (default) mode:

1. Create the content type "System" - an empty content type is fine.
2. Reload the backoffice.

This PR adds validation to prevent this specific case of unfortunate content type naming; it looks like this when you attempt to create a "System" content type:

image

There are likely other cases to be found down the road, and once found we can add them to the `reservedModelAliases` collection in the validator.


poornimanayar (approved):
Hey @kjac ,

Many thanks for the PR! It works as advertised. Happy to merge :-)

Poornima

poornimanayar:
And merged!

PR #8384, opened 2020-07-05
Make language name editable

ronaldbarendse:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7478.

### Description
The language name is already stored in the database and initially set to the display name of the culture (when adding the language). Since Umbraco 8.3.0 it's possible to change the culture (https://github.com/umbraco/Umbraco-CMS/issues/6503), but that doesn't change the language name (https://github.com/umbraco/Umbraco-CMS/issues/7478). This PR allows the language name of existing languages to be edited (so you don't have to manually enter the language name while adding it).

I've also cleaned up the language edit and overview AngularJS views and controllers, as they contained some quirks and unconventional styling. There are 2 commits that do change some existing design, namely removing the globe icon in the language overview (https://github.com/umbraco/Umbraco-CMS/commit/22039ad332f16cdc608bc3860225e088d4ad36c6) and making sure the styling of the `umb-control-group` is the same as the `umb-property` (https://github.com/umbraco/Umbraco-CMS/commit/b1f0647539e1f236ab1a32c17ed9bfda66bb13a6); these can be reverted/not merged (cherry-picked) though.

This PR can be tested by:
- Editing an existing language and make sure it's correctly stored and retrieved (shown in the back-office);
- Adding a new language and checking whether it's correctly storing the culture name as initial language name.

Edit language

PR #9006, opened 2020-10-01
Sort the variant picker entries by display name

kjac:
### Prerequisites

### Description

In the previous versions of Umbraco, the variant picker entries were sorted alphabetically:

image

Somewhere along the line this stopped working, leaving the variants to appear in a somewhat random order (I suspect by order of which the languages were created):

image

This PR re-introduces the alphabetical sorting in the variant picker:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-content-header.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-content-header.html
@@ -51,7 +51,7 @@
 
                     <umb-dropdown ng-if="vm.dropdownOpen" class="umb-variant-switcher" ng-class="{'--has-sub-variants': vm.hasSubVariants === true}" on-close="vm.dropdownOpen = false" umb-keyboard-list>
                         <umb-dropdown-item
-                            ng-repeat-start="entry in vm.variantMenu track by entry.key"
+                            ng-repeat-start="entry in vm.variantMenu | orderBy:'variant.displayName' track by entry.key"
                             class="umb-variant-switcher__item"
                             ng-class="{'--current': entry.variant === editor.content, '--active': entry.variant.active && vm.dropdownOpen, '--error': entry.variant.active !== true && entry.variant.hasError, '--state-notCreated':entry.variant.state==='NotCreated' && entry.variant.name == null, '--state-draft':entry.variant.state==='Draft' || (entry.variant.state==='NotCreated' && entry.variant.name != null)}"
                         >


poornimanayar (approved):
Hey @kjac ,

Tested and this works well 👍 

Poornima

poornimanayar:
And merged :-)

ronaldbarendse:
Languages currently don't have a specific sort order, so I would recommend making sure the `ILanguageRepository` actually returns the languages in a consistent order and make sure the AngularJS controllers/views maintain that instead. The same is probably also true for the segments below the languages/variants (and the introduction of segments might actually have caused this change).

This would also make it possible to add a sort order to languages later (without having to remove this custom ordering from the view, including the one from PR https://github.com/umbraco/Umbraco-CMS/pull/9007), just like I did for domains in PR https://github.com/umbraco/Umbraco-CMS/pull/8316.

nul800sebastiaan:
Yep agreed @ronaldbarendse but until we have proper sorting, this is an improvement already, and yes it will need to be amended when there's a sort order coming back from the controller at some point.

ronaldbarendse:
While testing this and removing all `orderBy` filters (from the `umb-navigation`, `umb-editor-content-header` and language overview), I've noticed it already sorts the languages alphabetically, but the default language is always displayed first.

Replicating this is easy: add random languages to a clean install (with en-US as default) and check whether the default is always first and the rest alphabetically:

Languages English default

Then change the default language and check whether that's now first:

Languages Danish default

So this PR and #9007 are actually regressions, as it makes total sense to have the default language displayed first, because that's the language all invariant properties can be edited in...

nathanwoulfe:
Would it be useful for the UI to indicate that the first language is the default? Might avoid potential confusion as to why the list appears unsorted...

ronaldbarendse:
@nathanwoulfe That's a great idea, although it should be a very clear indicator (e.g. not only a checkmark) without taking too much space (e.g. not a label with `(default language)`). I too see the potential confusion this creates and the reason for @kjac to create the linked PRs, so there's no questioning about that.

This is just me thinking, but wouldn't it make much more sense if the first is always the default, then the languages that use the default as fallback language and continue this fallback 'tree', alphabetically sorting the languages with the same fallback?

kjac:
I'm a firm believer in letting the server serve data and let the UI handle the data presentation. The way I see it, the server shouldn't dictate orders of anything unless they have a functional meaning (i.e. concrete sort orders for content nodes). 

In other words I don't think there should be a set sort order for languages, unless it has a functional purpose.

For now I think sorting these things by alphabetical order makes sense, at least more sense that previously. To build upon that, perhaps we should show the variant picker entries in two sections - the ones that have been created/edited and the ones that haven't ... each section again sorted alphabetically. And yes, a clear indication of the default language might be good too.

Anyway - those were my two cents 👍

ronaldbarendse:
@kjac I totally understand what you're saying: languages themselves don't have a specific order, but depending on the context, should be ordered in the most logical way.

In this case though, the server _is_ giving the languages a specific order: default first, then alphabetically (in the `LanguageMapDefinition` mapping, see PR https://github.com/umbraco/Umbraco-CMS/pull/9165). The `LanguageRepository` however sorts the languages on ID (as part of some legacy reason), so that already seems odd:
https://github.com/umbraco/Umbraco-CMS/blob/13488be0a3280c966ca5115c1cf75cc1979e5819/src/Umbraco.Core/Persistence/Repositories/Implement/LanguageRepository.cs#L50-L55

I would argue the alphabetical sorting doesn't make a sense at all, as you probably want all related languages together (e.g. Scandinavian languages, like Danish and Norwegian, or Wester/Eastern languages) and the default always at the top (as invariant properties can only be edited in the default language).

The other thing to consider is the language fallback chain, it might make sense to have a tree like:
- English - default
  - English (United Kingdom) - fallback to English
  - English (United States) - fallback to English
- Dutch
  - Dutch (Belgium) - fallback to Dutch
  - Dutch (The Netherlands) - fallback to Dutch
- French
  - French (Belgium) - fallback to French
  - French (France) - fallback to French

This tree can have deeper fallback chains, so that would be something the UI needs to take into account (including the segments that are already displayed below the languages). If you're primarily targeting the US, it would also make sense to have that language above UK, same for The Netherlands and France, as Dutch and French are respectively the main languages for those countries.

These are all arguments to give languages a specific order (to group related ones) _and_ let the UI group the languages based on the fallback chain (in the tree/content culture pickers).

@kjac Grouping based on whether the content is created for that language would cause the order of languages to differ between nodes, which totally breaks the 'muscle memory' for editors that expect a specific language to always be in a specific place...

@nathanwoulfe If the default language is always displayed first, having the indicator would not add much. Languages can also be mandatory though and that's a lot more valuable to display here, as it's only shown in the dialog when publishing the content.

ronaldbarendse:
@kjac Also note that if language names become editable (PR https://github.com/umbraco/Umbraco-CMS/pull/8384), sorting alphabetically might result in changing the order when renaming (e.g. 'English (United Stated)` to `American`) and cause the same issue for editors (changing the place they expect a language to be).

PR #9007, opened 2020-10-01
Sort the languages in the global language selector

kjac:
### Prerequisites

### Description

This is somewhat related to #9006 - the languages in the global language selector aren't sorted in any particular way:

image

The languages used to be sorted alphabetically, which I think makes perfect sense...to this PR adds sorting to the language selector:

image


poornimanayar (approved):
Hi @kjac ,

Tested and this works well :-)

Poornima

poornimanayar:
And merged :-) Many thanks for the work as always!

Issue #8955, opened 2020-09-24
NetCore: Introduction tour can't be completed


elit0451:
### Bug summary

(1) Introduction tour cannot be completed due to a lost step on **alpha02**.

### Steps to reproduce

Navigate to Getting Started tours and begin the introduction tour.

### Expected result

We can complete the Introduction tour.

### Actual result

The tour cannot continue after step 7.

![image](https://user-images.githubusercontent.com/21998037/94125252-69916400-fe56-11ea-8b25-28d2f986244b.png)

benjaminc:
It looks like the PR for this has been merged, and the subsequent PR that fixes an issue found after the first PR is also merged.  Is there anything still outstanding on this issue, or should this be closed?

bergmania:
Thanks @benjaminc .. It should be closed - My bad

PR #9008, opened 2020-10-01
NetCore: Fix for Back-Office tours

elit0451:
Solves issue: https://github.com/umbraco/Umbraco-CMS/issues/8955

### Test:
Make sure all tours can be successfully completed.


nikolajlauridsen:
I just ran through the tours on a clean install and it gets further, but unfortunately, there's still some issues. This PR does fix the issue reported in #8955 so I don't know if these other things should be a separate task? Let me know if that's the case :) 

#### Create content
After running the "Create content" an empty box appears and the screen is whited out, when you click close it's back to normal
After running create content node

#### Render in template
When doing the render in template portion of the tour you're asked to expand the template folder, however, since the template folder is expanded by default you'll end up closing the folder and locking the test. 

If you manually close the folder before running the test the tests runs fine.
Render in template

#### View home page
The View home page part of the tour gets lost before the third step with an "oh we got lost message"

However, this might be because my content tree doesn't automatically update for some reason when I manually clicked "Reload" the test ran through with no issues, so I'm not sure if this is related to the tours at all.
View home page

elit0451:
Thanks for checking the PR @nikolajlauridsen 😊 

I have managed to reproduce the issue happening during the execution of the **Create content** tour.  However, after running `npm install` and `npm run build` in *~/src/Umbraco.Web.UI.Client* the issue didn't appear again (also the reason why the Content tree wasn't updated -> what you reported in **View home page** paragraph)

It turns out it is not related to the tour itself. And the empty box you saw was an error not rendering. I managed to get the error - it is related to the PostSave action that is executed when clicking **Save and publish** of the content. Need to check with @bergmania if a new issue should be created for it (on the NetCore project)
> URL: https://localhost:44331/umbraco/backoffice/umbracoapi/content/PostSave
>> error:



And when refreshing the tree it seems like the save and publish operation was successful (too many times - got the same node created depending on how many times I clicked the btn):




</br>

Regarding the **Render in template** tour, when checking with *v8/dev*, the behaviour seems identical. As you mentioned, the tour locks and it seems like this has happened after v8.7 (tested also on the 8.8.0 release - same bug). On the tour step with a title "*Expand the Templates node*" we have **"skipStepIfVisible"** attribute which will prevent the tour from such locking. Basically, we will skip this step if we find a button with `data-element=tree-item-expand`, which is there, but we are also looking for a class `icon-navigation-down` on this same button element. The class however, is now present as a class on the span child of the button. So this needs to have a new issue defined (for v8)!

image

bergmania:
After merging in latest v8, I can only repro the issue with templates tree already expanded..

bergmania:
I created a PR to v8 to fix the issue https://github.com/umbraco/Umbraco-CMS/pull/9070

PR #9010, opened 2020-10-01
V8 - Don't index descendants of unpublished page

### LINKED ISSUE #8879: Examine is indexing inherited unpublished nodes when the external index is rebuilt
Content nodes that have an unpublished ancestor, and are therefore unpublished by inheritance, are being added to the Examine ExternalIndex when the index is rebuilt.

This is a similar issue to this old one (the problem may have returned):
https://issues.umbraco.org/issue/U4-9318

I'm seeing this issue on Umbraco version:  8.6.4

Steps to reproduce:

Create two content nodes of any type, one below the other, making a parent and a child node. Save and publish both nodes and note down their node IDs.
Go to Settings > Examine Management > ExternalIndex and search for the parent and the child using node ID to confirm that they are both in the index (as expected).

Unpublish the parent node, but leave the child node published.
Go back to Settings > Examine Management > ExternalIndex and search for the parent and the child again using node ID to confirm that neither are in the index (as expected).

Click the "Rebuild index" button and wait for the ExternalIndex to finish rebuilding.

Search for the parent and the child again using node ID. This time the parent is not in the index (as expected) but the child is.

Expected outcome:

I would not expect the published child to be in the external index after the rebuild (because the parent is unpublished).

dawoe:
### Prerequisites

If there's an existing issue for this PR then this fixes #8879 

### Description

Like the issue mentions, published descendants of unpublished item end up in the ExternalIndex when you rebuild it from the dashboard.

So I refactored the code to go recursively (yeah I know) through the children. This code works and the descendants should not end up in the index anymore

But they still do.  

When I put a breakpoint in the ContentIndexer PopulateIndexes method I see this get hit twice when rebuilding the index. Once with _publishedValuesOnly set true and once with it set to _false.

This because the list of _populators (https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Examine/IndexRebuilder.cs#L36) used when rebuilding the index has 4 populators (content, publishedcontent, media and members)

During the populate method that is getting called there is a check if the populator is registered for a index (https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Examine/IndexPopulator.cs#L41)

I would expect that would for the ExternalIndex only the MediaIndexPopulator and the PublishedContentIndexPopulator would be registered. But the ContentIndexPopulator is registered as well.

Problem is I can't seem to find where they are registered. The index populators have RegisterIndex method. But that only seems to be called  in Unit tests.

I also tried so if this could be fixed in ContenSetValidator, because that has a validatPath method. But that always returns true. Because the ParentId is never set.

I could use some help on this one from HQ... probably @Shazwazza


src/Umbraco.Examine/ContentIndexPopulator.cs
```diff
@@ -70,31 +70,65 @@ protected override void PopulateIndexes(IReadOnlyList<IIndex> indexes)
...
             IContent[] content;
 
             do
             {
-                if (!_publishedValuesOnly)
-                {
-                    content = _contentService.GetPagedDescendants(contentParentId, pageIndex, pageSize, out _).ToArray();
-                }
-                else
+                content = _contentService.GetPagedDescendants(contentParentId, pageIndex, pageSize, out _).ToArray();
```
Shazwazza, line 90: ToList is faster than ToArray

src/Umbraco.Examine/ContentIndexPopulator.cs
```diff
@@ -70,31 +70,65 @@ protected override void PopulateIndexes(IReadOnlyList<IIndex> indexes)
...
+
+                if (content.Length > 0)
```
Shazwazza, line 92: There is no need to do a ToArray( ) and then check the length before doing a foreach. The foreach will not iterate if there is no items

Shazwazza (changes requested):
Have left minor notes inline but we need to tackle this a different way. We need to have the minimal amount of queries possible and querying per parent could be poor performance for some (it will be worse than it is currently). This should be handled like it is done in v7, see here https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/UmbracoExamine/UmbracoContentIndexer.cs#L495 

v7 orders by level descending but we can keep using the Path ordering since it will achieve the same thing. The way it works is that it tracks which paths are unpublished and then doesn't continue for the descendants of that path. More code here https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/UmbracoExamine/UmbracoContentIndexer.cs#L638

this should be fairly easy to replicate back to v8

dawoe:
Hi @Shazwazza 

I know my query for getting the pages can be optimized. But I stopped working on it because of the issue the ContentIndexPopulator also runs for the ExternalIndex. So even if optimize the query for getting published pages when the PublishedContentIndexpPopulator runs, it still will get the the unpublished ones and index them as well.

Dave

Shazwazza:
@dawoe Right, I see that now. That's a bit of an oversight 😢 I've just pushed a PR for that fix here https://github.com/umbraco/Umbraco-CMS/pull/9137 if you can have a look?

src/Umbraco.Examine/ContentIndexPopulator.cs
dawoe: I think an array is used here because the  GetValueSets method of  ValueSetBuilder needs an array. 
Don't know if using a list and then calling ToArray is faster ?

dawoe:
Hi @Shazwazza or other reviewers,

I merged PR #9137 in to this PR because I needed to changes to fix the issue. I tested that PR and that works.

See the issue for steps to reproduce. With this PR the unpublished descendants don't show up in the index anymore

src/Umbraco.Examine/ContentIndexPopulator.cs
```diff
@@ -70,31 +77,89 @@ protected override void PopulateIndexes(IReadOnlyList<IIndex> indexes)
...
+            IContent[] content;
+
+            var publishedPages = new HashSet<int>();
+
+            do
+            {
+                //add the published filter
+                //note: We will filter for published variants in the validator
+                content = _contentService.GetPagedDescendants(contentParentId, pageIndex, pageSize, out _, _publishedQuery,
+                    Ordering.By("Level", Direction.Ascending)).ToArray();
```
Shazwazza, line 122: Did you test with ordering by Path? IIRC ordering by path will order it by level implicitly

src/Umbraco.Examine/ContentIndexPopulator.cs
Shazwazza: Ah i see, no worries then. And nope, calling ToList() then ToArray() isn't faster, it will be slow and also allocate another unnecessary object.

src/Umbraco.Examine/ContentIndexPopulator.cs
```diff
@@ -70,31 +77,89 @@ protected override void PopulateIndexes(IReadOnlyList<IIndex> indexes)
...
                 if (content.Length > 0)
                 {
+                    var indexableContent = new List<IContent>();
+
+                    // get the max level in this result set
+                    int maxLevel = content.Max(x => x.Level);
+
+                    // gets the first level pages, these are always published because _publishedQuery filter
+                    // and store the id in a hash set so we can track them
+                    var firstLevelContent = content.Where(x => x.Level == 1);
```
Shazwazza, line 133: Why do we need to do this? This means that we are doing to iterate over the entire result set more than once which we don't want to do. This also means we are going to iterate over all subsequent paged results sets even though we would know we no longer have any level 1s remaining. We already know we are ordering by level so we know the first content is going to be level one which can be tracked. We should be able to achieve what is needed without iterating the result set more than once. We manage this this way in v7 https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/src/UmbracoExamine/UmbracoContentIndexer.cs#L497

src/Umbraco.Examine/ContentIndexPopulator.cs
```diff
@@ -70,31 +77,89 @@ protected override void PopulateIndexes(IReadOnlyList<IIndex> indexes)
...
+                    {
+                        publishedPages.Add(item.Id);
+                        indexableContent.Add(item);
+                    }
+
+                    // get content per level so we can filter the pages that don't have a published parent
+                    // this because the paged descendants method does not take unpublished ancestors into account
+                    for (var level = 2; level <= maxLevel; level++)
+                    {
+                        var levelContent = content.Where(x => x.Level == level && publishedPages.Contains(x.ParentId));
```
Shazwazza, line 145: We cannot do this, this means we are iterating the entire result set again. We need to be able to do all of this without re-iterating the result set more than once (i.e. we only do a foreach on the result and we track what is needed within the single loop)

Shazwazza (changes requested):
Thanks! I've added some notes inline.

src/Umbraco.Examine/ContentIndexPopulator.cs
dawoe: I did this check because we are in a loop. If somebody more than 10000 root level nodes we need to keep track of this somehow. But I will see if I can optimize

src/Umbraco.Examine/ContentIndexPopulator.cs
dawoe: Works also by path, did it by level because that was how the v7 example did it

src/Umbraco.Examine/ContentIndexPopulator.cs
dawoe: Refactored to everything in a single foreach

src/Umbraco.Examine/ContentIndexPopulator.cs
dawoe: Refactored to everything in a single foreach

src/Umbraco.Examine/ContentIndexPopulator.cs
```diff
@@ -70,31 +77,83 @@ protected override void PopulateIndexes(IReadOnlyList<IIndex> indexes)
...
+
+                if (content.Length > 0)
                 {
-                    //add the published filter
-                    //note: We will filter for published variants in the validator
-                    content = _contentService.GetPagedDescendants(contentParentId, pageIndex, pageSize, out _,
-                        _publishedQuery, Ordering.By("Path", Direction.Ascending)).ToArray();
+                    // ReSharper disable once PossibleMultipleEnumeration
+                    foreach (var index in indexes)
+                        index.IndexItems(_contentValueSetBuilder.GetValueSets(content));
```
Shazwazza, line 103: This can be improved. There's no reason to call the same method to build value sets for each index when the result is the same for all of them: `_contentValueSetBuilder.GetValueSets(content)`. I realize this was like this before but it's a very easy win to not have to allocate a whole bunch of objects for each index. So before we enumerate we can store `var valueSets = _contentValueSetBuilder.GetValueSets(content).ToList()` and then just `index.IndexItems(valueSets);`

Shazwazza (approved):
One minor thing that we could change but happy to change that later on since it was like that already, otherwise looks great!! awesome work. I'll test!

src/Umbraco.Examine/ContentIndexPopulator.cs
dawoe: Fixed it

Shazwazza:
Not sure why this PR didn't auto close, it was merged in rev 35bf59dee41bd6033abecfd56e95d5ac794e5d84

Shazwazza:
ha! well i know why it didn't automatically close, because you added a commit very recently :P ok cool, i'll re-merge

Shazwazza:
there we go, great stuff :)

nul800sebastiaan:
Cherry picked for 8.6.6+ in a6101ce499749d5f569537d3519dbcfe3c07dde1

src/Umbraco.Examine/ContentIndexPopulator.cs
```diff
@@ -15,7 +16,7 @@ namespace Umbraco.Examine
     /// <summary>
     /// Performs the data lookups required to rebuild a content index
     /// </summary>
-    public class ContentIndexPopulator : IndexPopulator<IUmbracoContentIndex>
+    public class ContentIndexPopulator : IndexPopulator<IUmbracoContentIndex2>
```
bielu, line 19: @nul800sebastiaan that one is breaking change as change behavior of umbraco can we get that PR tagged as breaking change please?

Shazwazza:
Have marked this as a breaking change because:

This check https://github.com/umbraco/Umbraco-CMS/pull/9137/files#diff-ec2f6e77662a173dae15422b1557c62bace65290edc56f55bebed82642b1d413R19 actually fails if the index is not specifically of type `IUmbracoContentIndex2` and is only `IUmbracoContentIndex`, even though `IUmbracoContentIndex2` inherits from `IUmbracoContentIndex`.

Shazwazza:
It's really this change that is broken: https://github.com/umbraco/Umbraco-CMS/pull/9137/files#diff-7f317f8357d3cc94abec7725c38886f8617ae3fa7a345c82973235cafa302ae0R18

PR #9011, opened 2020-10-01
use 'contact links' config feature to replace issue templates that link elsewhere

LottePitcher:
Three of the github issue templates (Support, Documentation, and Security) ask people to go elsewhere.

This PR adds a config.yml file that uses the 'contact links' feature, so the links appear on the new issue page directly, no need for separate templates (so this PR deletes those too).

I've seen other repos that use emojis in the yml so these should be ok.

My only concern is that there is some information in the security template that doesn't appear to be discussed on the official 'report a vulnerability' page.


nul800sebastiaan:
Fantastic, thanks again @LottePitcher! ⭐

PR #8994, opened 2020-09-30
fixes 8924, Always clear both

nielslyngsoe:
fixes https://github.com/umbraco/Umbraco-CMS/issues/8924

+ added word-break for description to avoid having long words in description overlapping the property controls.

---
_This item has been added to our backlog [AB#8649](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8649)_

PR #9012, opened 2020-10-01
Fix : Member Info Content App looks broken

poornimanayar:
The members info content app looks quite funny in the current `v8/contrib` branch. This PR fixes it. I am not sure whether there is a better way to fix this though!

image


bjarnef:
@poornimanayar I think @nielslyngsoe has already fixed this in https://github.com/umbraco/Umbraco-CMS/pull/8994

poornimanayar:
Thanks @bjarnef for pointing this out :-)

PR #9013, opened 2020-10-01
Replaced <i> with umb-icon in umb-node-preview

poornimanayar:
Another `umb-icon` from me. I have replaced the `<i>` with `umb-icon` in `umb-node-preview` . This is quite a sweeping change as the `umb-node-preview` is used in many many places. Would recommend testing the following
1) Info content apps in content, media, members
2) Icons in grid editor config
3) icons in content picker in content edit
4) icons for user groups in users section


nathanwoulfe:
Hi @poornimanayar should this be setting all the icons in the umb-node-preview directive to the svg value or is that a separate change? All looks fine otherwise, no more `<i>` in the node preview elements.

poornimanayar:
Hey @nathanwoulfe , my work replaces the `<i>`  in the `umb-node-preview`  to use the `umb-icon` directive. But the node preview is used in many many places across the application as I said before. I have done a search on where it is used and tested all of it. It all seems to be working okay

nathanwoulfe:
Yup, all works fine from what I've seen 👍👍

bjarnef:
@poornimanayar when using `<umb-icon>` it doesn't need `aria-hidden` attribute. The component takes care of this 👍

poornimanayar:
Thanks for pointing out @bjarnef, missed that detail.. will correct it :-)

nathanwoulfe (approved):
Hey @poornimanayar, took me a while to get back to this, but it's allllll good 👍

PR #9014, opened 2020-10-01
run JS unit test in development script

nielslyngsoe:
This PR makes the JS Unit test server run when you run the dev npm script, in this way it becomes more clear if Unit Tests are broken.

---
_This item has been added to our backlog [AB#8663](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8663)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/gulp/tasks/test.js
+++ b/src/Umbraco.Web.UI.Client/gulp/tasks/test.js
@@ -23,7 +23,6 @@ function runUnitTestServer() {
         autoWatch: true,
         port: 9999,
         singleRun: false,
-        browsers: ['ChromeDebugging'],
         keepalive: true
     })
     .start();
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
@nielslyngsoe I would argue the opposite: when developing I want to make some quick changes and see if it works, having to wait for the whole test suite to run is a barrier for me to try and develop anything. I don't know if something can be done async? In general, the `dev` task is quite slow for me and I never know when it's done copying all the things, I would love to see more logging to know that I can refresh my browser to see the results of my edits. I also would love it if we stop doing some things in the `dev` script: image minifying is killing me, splicing the views into the js files is not needed when I'm just doing dev work.. maybe more. We have a need for speed here! 😅

nielslyngsoe:
@nul800sebastiaan more speed would be super great. This is not slowing anything down as it runs after the compilation is performed, so it's just an extra thing.

Issue #8916, opened 2020-09-16
Ability to add "umb-app-header__actions".


nielslyngsoe:
I would like Extensions/Packages/"code in general" to be able to add app-header actions. An app header action should not be specifically defined and therefore I as a package developer would like to be able to inject the HTML I like to append.

![image](https://user-images.githubusercontent.com/6791648/93311610-813d6c80-f806-11ea-8b80-f2ae53cf09b7.png)

The reason for this is that a header-action might need to have a badge or other features on its menu-item. The action of a Header-Action should not be specific to showing a view in a specific spot. I want devs to be able to make their own interactions. So they can choose whether it should open dialogs, sidebars, or appear as a section.

nielslyngsoe:
This request was posted previously but I can't seem to find it. I would love to see this feature in our CMS. I think devs who make extensions or just use Umbraco together with other software can use this spot for a more appropriate/subtle integrations.
Let me know any thoughts, Especially how you think this can be useful?
Thanks

skttl:
Yes, this would be great for something like https://our.umbraco.com/packages/backoffice-extensions/deploy-assistant/

I guess it should be done in a package.manifest, like it is done for content apps?

patrickdemooij9:
Love the idea. Played around a bit and have gotten "apps" working in the header:
![image](https://user-images.githubusercontent.com/11466511/94802539-2ac85480-03e8-11eb-970e-b9587b326dd7.png)
Now, just have to connect it up to an API request.

nielslyngsoe:
@patrickdemooij9 Looks good, you might want to make a draft PR then others can help out with parts of it if you like?

Ps. I guess @skttl can read my mind... ;-P

patrickdemooij9:
@nielslyngsoe always open to having other people help me or just review my code! Created the draft PR here: https://github.com/umbraco/Umbraco-CMS/pull/9016

umbrabot:
Closing this in relation to PR #9016 - make sure to read the close reason on #9016

PR #9016, opened 2020-10-01
Ability to add "umb-app-header__actions"

patrickdemooij9:
Issue: https://github.com/umbraco/Umbraco-CMS/issues/8916


bjarnef:
I just skimmed the code a bit. Do we really need a view for each app or should we limit it to some specific features using a shared view? E.g. a `<button>` element using `<umb-icon>` (font icon or SVG icon). It could have a name used as tooltip and for screen readers. Maybe a badge to make a notification similar to content apps.

@nielslyngsoe @skttl thoughts?

patrickdemooij9:
@bjarnef good point. It was one of the things I wasn't sure about yet. This way, the developers have more freedom, but that of course comes with the danger of less consistency and more work as you also have to make a view. While if we enforce the view and only let them enter the action, name, alias you'll keep the consistency and it's easier to add but also has less freedom for the developer.

I think the second option is better for now, just because it's also easier to add for the developers. But would love to hear what other people think about it!

bjarnef:
@patrickdemooij9 sometimes it is great to keep it simple and don't offer to many options to avoid developers implementing very different views. With a specific format it will also ensure the accessibility is in top, which the developer otherwise should handle.

skttl:
I think we need to look at how Content Apps are implemented, and make this work similar.

This means a header action could be defined in a package.manifest or from C#.

There could be a default view, for simple stuff, like just adding a button with a badge. Could look like this
```
{
    // define the headerActions you want to create
    "headerActions": [
    {
        "name": "Deploy Assistant",
        "alias": "deployAssistant",
        "weight": 0,
        "icon": "icon-calculator",
        "controller": "DeployAssistantHeaderActionController",
            "show": [
                "+role/admin"  // show for 'admin' user group
            ]
    }
    ],
    // array of files we want to inject into the application on app_start
    "javascript": [
        "~/App_Plugins/DeployAssistant/deployassistantheaderactioncontroller.js"
    ]
}
```

The controller would then be able to add notification badges, just like you can do with content apps (https://our.umbraco.com/documentation/extending/Content-Apps/#notification-badges)

The controller would also be responsible for handling click actions etc. on the headerAction.

skttl:
Can you add an example on how to add your own header action?

patrickdemooij9:
@skttl I haven't gotten around to wiring the clicks up to any actions. Mostly because I still have to figure out how to be able to do that. There are 3 things left to do:
- Connect a controller to each action so that they are able to add badges and handle their own clicks. (No idea how to do that in Angular, any pointers would be appreciated 👍 )
- Make sure to be able to read header apps from the manifest
- Figure out what to do with the localized text. Right now I use the alias and name for it, but the localized text is more like a description, so I am not quite sure if it fits there.

skttl:
I found this blog post showing how to bind a controller from a variable (aka dynamic controller binding :)), I haven't tried it myself though, and after reading this, I wonder if we should opt for the simple way, of just having a view for each header action, like you suggested at first.

I think your handling of localized text is good - IIRC this is how section texts are done too.

bjarnef:
@skttl although a simple view file could be used, I fear this won't be consistent between the different header apps.
In my option I would be simpler and ensure consistency with a few limited options.

- Name
- Alias
- Icon
- Action (link to a route or similar or an action opening a modal/overlay).
- Badge (count, color, etc)

I don't think we need much else than this. So better start simple and enhance later.

skttl:
Okay, so instead of calling it `action`, it should be `view`, or `drawerView`. I think it should open in the same drawer as the help function today.

We need to figure out how to handle the search part then, as that one has its own way of working (the search overlay)

skttl:
But wouldn't you still need to be able to inject some JS into the header action icon itself, in order to update the badge?

Take the example of my Deploy Assistant package. I would like to be able to have a deploy assistant icon in the header, and have it continuously update with the current deploy status (in progress, failed or complete).

bjarnef:
@skttl some might want it to open a drawer like the help panel, but it could also open an overlay (like the search kind of does) or maybe just navigate to a route?

patrickdemooij9:
@bjarnef @skttl, reading the comments I feel like the Header apps should have the following functionality:

- Should have the ability to hook up some angular code to the specific app (apart from the action). This to make sure that it can be used within packages like the Deploy Assistant where you want to update the badge based on the progress of a specific task on the server.
- Should be easy to add and follow the style of the current header. 
- Should not be restricted to a certain type of click action. It should be able to open an overlay, drawer, link. Whatever the user would want.

So, looking at the requirements above, I think there is one solution that we could use here, but please give feedback on what you think about it. We'll go back to the point where there were different views for the various buttons. However, we'll also implement a default view that handles any simple app that only needs to do the following:

- Name
- Alias
- Icon
- Action (link to a route or similar or an action opening a modal/overlay).
- Badge (count, color, etc)
That default view can then probably be used for most of the scenarios that the user would want. However, we'll also have some more advanced scenarios like @skttl mentioned. For this, we should have the programmer be able to set their own view. This will then override the default view and let you have more control over this. We'll have to see how much control we should give the user here, as we do want the styling to match up with the rest of the header.

For the default/override view logic, we can use a combination of an interface and an abstract class. We'll have the IHeaderApp, which requires you to give you a view, but we'll also have a default abstract BaseHeaderApp (that implements IHeaderApp) which will always set the view to the base view mentioned above. Users that want to only show an icon with simple action, can implement the BaseHeaderApp. And people who want more freedom can decide to implement the IHeaderApp.

Of course, always open to feedback and maybe there is a different and better way to solve the issue.

skttl:
I think the most simple way of adding this is by having a view for each header app. So your manifest would be nothing more than name, alias, view (and optional weight and show). This is also what @nielslyngsoe is suggesting in #8916. It would then be up to the dev to add interaction (whether it should link to a route, open a modal/drawer/overlay, or something else).

We could add an umb-header-action directive to make it easier to implement the view.

If you add action as a manifest option, you will miss a way to add custom actions, because these are bound to the controller controlling the action. So you need to hijack the default header app controller which is messy, or rely on vanilla javascript methods which is hacky when working with an angularjs app.

If you add badge as a manifest option, you won't have a way to update the badge, It will only ever be defined in the manifest. So even if you could update it, if your manifest is in code, the only way to update the badge in the frontend is by reloading the browser.

By having "just" a view, it's up to the developer to define an action for the item, the icon can be picked by the developer (or be some text or whatever you like). And the badge could be implemented in any way necessary.

nielslyngsoe:
@bjarnef @patrickdemooij9 @skttl interesting to read your thoughts on the matter, It's not like I have the right answer lying around for you, so I trust you will find a good solution that suites the different needs.

To put some perspective to the request I will add a few thoughts on how this can be used:

1. It would be nice if I, as a developer, easily(through the use of package.manifest) can add a simple umb-header-actions. Where I don't need to make a controller or a view, and this approach would be consistent with the rest in terms of UI. (I imagine when controller or view or ... isn't defined you get this default)

2. It would be nice if I, as a pro-developer, are able to make my own experience. Let's use the Deploy Assistant package example. I would like to control the badge, when the badge appears but also how it appears. In this example, I do not want a consistent experience cause we cannot at this stage predict what the developer wants to do. I will trust developers at this stage are making something that will appear consistent. For the Deploy Assistant package I want to have a progress-bar-border around my badge to show how the transfer progress is coming along. When I add things to my Deploy Queue I want my icon and badge to do a little jump, or an animation, to indicate that something was put into the "basket". I also imagine that the icon changes look depending on whether I as a user have things in my Deploy queue. And I want different looks depending on which environment I'm visiting.
I also what to control what the click-action does. It should not open a section or a sidebar-panel. I would like to do a Modal view that appears beneath the header-action. 🚀

I hope that helps a bit with figuring out the scope of the feature. 🤹‍♀️

skttl:
For the simple header action, I think they should only be able to change the location then.

So a package manifest would be something like

```json
    {
        "name": "My simple header action",
        "alias": "mySimpleHeaderAction",
        "weight": 0,
        "icon": "icon-calculator",
        "location": "#/content/1234",
            "show": [
                "+role/admin"  // show for 'admin' user group
            ]
    }
```

Other packages who could use a custom view for a header action:
- https://our.umbraco.com/packages/backoffice-extensions/ourumbracoenvironmentdashboard/
- https://our.umbraco.com/packages/backoffice-extensions/maintenance-manager/
- https://our.umbraco.com/packages/backoffice-extensions/impersonator/
- https://our.umbraco.com/packages/developer-tools/sitelock/

The "advanced" manifest would then be
```
    {
        "name": "Deploy Assistant",
        "alias": "deployAssistant",
        "weight": 0,
        "view": "~/App_Plugins/DeployAssistant/HeaderAction.html",
            "show": [
                "+role/admin"  // show for 'admin' user group
            ]
    }
```

So the only required options is name, alias and weight.

If a view is defined, the view is used. If not, the icon is used for a default view, where the action changes the location to the one configured.

patrickdemooij9:
What does the "location" do in the simple header app? I assume that's an easy way to jump to any page? We could also still keep the action there and follow the same logic as the menu items currently use. Then it won't restrict the user on the simple app to only have a redirect header app.

But I'll try implementing something that (hopefully) matches most of the above wishes when I get to my lunch 👍

skttl:
But then you are bound to use the actions provided by the default controller, which already is implemented.

patrickdemooij9:
Not really, I believe. It'll only be something for the default header app, so you can still have your advanced ones. And you should be able to reach other services in the angular application as well. Custom menu items let you choose your own action which (I believe) lets you call methods on things like the navigation service.

patrickdemooij9:
I added the logic for the manifest now, so you should be able to add a new header app with the following manifest:
```
{
    "headerApps": [
        {
            "name": "Test App",
            "alias": "testApp",
            "weight": 0,
            "icon": "icon-wine-glass",
            "view": "views/header/apps/help/help.html",
            "show": [
                "+role/admin"
            ]
        }
    ]
}
```
I also made a default header app view that doesn't require a view but works with just a action.

skttl:
Just tried it out - works really well #h5yr!

I noticed the show parameter in the manifest is required for the app to show up. Is that the same for content apps?

My app:

package.manifest
```json
{
    "headerApps": [
        {
            "name": "My Header App",
            "alias": "myHeaderApp",
            "weight": 0,
            "view": "~/App_Plugins/MyHeaderApp/MyHeaderApp.html",
            "show": [
                "+role/admin"
            ]
        }
    ],
    "javascript": [
        "~/App_Plugins/MyHeaderApp/controller.js"
    ]
}
```

view:
```html
<div ng-controller="MyHeaderAppController" data-header-action-alias="MyHeaderApp">
    <button type="button" class="umb-app-header__button btn-reset" ng-click="sayHello()">
        <umb-icon icon="icon-traffic" class="umb-app-header__action-icon icon-traffic"></umb-icon>
        <div class="badge">{{ count }}</div>
    </button>
</div>
```

controller:
```js
angular.module("umbraco").controller("MyHeaderAppController",
    function ($scope, notificationsService, $timeout) {
        $scope.count = 0;

        var inc = function() {
            $scope.count++;
            $timeout(inc, 1000);
        };

        inc();

        $scope.sayHello = function() {
            notificationsService.info("Hello", $scope.count);
        };
    });
```

EA71Qyd94P

bjarnef:
From what I can read here, it seems the `view` property here is not consistent with how it works for content apps, where the `view` property refer to the view being loaded in the editing area: https://our.umbraco.com/documentation/extending/Content-Apps/#creating-the-view-and-the-controller and the `icon` is part of the package.manifest https://our.umbraco.com/documentation/extending/Content-Apps/#setting-up-the-plugin

@skttl from your example it seems the "view" doesn't need much more than and `name`, `alias`, `icon` and `badge` similar to this for content apps: https://our.umbraco.com/documentation/extending/Content-Apps/#notification-badges

```
<div ng-controller="MyHeaderAppController" data-header-action-alias="MyHeaderApp">
    <button type="button" class="umb-app-header__button btn-reset" ng-click="sayHello()">
        <umb-icon icon="icon-traffic" class="umb-app-header__action-icon icon-traffic"></umb-icon>
        <div class="badge">{{ count }}</div>
    </button>
</div>
```

The `view` property could however be used when clicking the button, e.g. open the view html in an overlay (e.g. centrered or on right side), drawer (like help panel), etc.

Not sure how to implement the button `sayHello()` function without this view. If it does need this view, maybe it should be named something else to be more consistent with content apps?
Some might want to use header apps for like in your example, while other maybe just want to use it to load a view in a modal/overlay/drawer like help panel?

skttl:
I think, if you want to open up something in an a drawer, upon clicking the header-app, it would be up to you as a developer to make that happen. Just like it is for anything else in the backoffice.

Header apps should just enable you to add parts to the top right corner.

It doesn't IMO make sense to tie the developer into opening stuff in a dialog or loading a location. I think we should keep this simple, and just allow specifying a view for the header app, and optionally weight and the show option. The other options is at best useless IMO, because you either need to figure out some complex way of making default behavior, or only be able to specify an action from the core header apps, that in most (if not all) cases already is implented by the core header apps.

bjarnef:
Sure, but altough the header app and content app definitions in package.manifest looks very similar,  the behaviour would be different, since e.g. content app doesn't allow the customize the view of the content app button itself, which may be confusing to some.

I think we should allow to specify the `icon` as well in the package.manifest so this is used when view is not specified (default view is used) and this is available in model in the custom app view.

nielslyngsoe:
@skttl @patrickdemooij9 @bjarnef just a thought, but maybe if we didn't name it "view" it wouldn't be misunderstood.

What if it was named "appview"? or "iconview"?. If not defined you could define an "icon" and then you will be using "location" to define the action of it.

I can see the misunderstanding so if we can avoid that easily I think that would be preferable :-)

patrickdemooij9:
> Sure, but altough the header app and content app definitions in package.manifest looks very similar, the behaviour would be different, since e.g. content app doesn't allow the customize the view of the content app button itself, which may be confusing to some.
> 
> I think we should allow to specify the `icon` as well in the package.manifest so this is used when view is not specified (default view is used) and this is available in model in the custom app view.
This should work now. The default view will use the icon if there is one in the manifest.


> @skttl @patrickdemooij9 @bjarnef just a thought, but maybe if we didn't name it "view" it wouldn't be misunderstood.
> 
> What if it was named "appview"? or "iconview"?. If not defined you could define an "icon" and then you will be using "location" to define the action of it.
> 
> I can see the misunderstanding so if we can avoid that easily I think that would be preferable :-)
Yeah, maybe that would help. Maybe it shouldn't even be called HeaderApp as it's so close to ContentApp.

nielslyngsoe:
> Yeah, maybe that would help. Maybe it shouldn't even be called HeaderApp as it's so close to ContentApp.

Yeah, but then again it's just two words and from just those, it's understandable where and what it is. I would argue that from a user perspective it's the same. It's "just" for the developer that it's not. I would say we just need to make the technical differences stand-out so you as a developer understand that it's not the same.

I think naming wise it would be a good rule-of-thumb if we asked our self, what if we liked the be able to overwrite the "icon-view" of ContentApps, how would that come in play and what should that be called. This would lead us to a name that can be reused across concepts.

PR #9017, opened 2020-10-01
Prevents templates being created when adding DocTypes through the block list config

### LINKED ISSUE #8646: V8.7RC: Adding a new block definition creates a template on disk
In 8.7, you can add a Block List datatype to your document type. You need to configure the Block List and in the process add existing element types to it or create a new one. 

image

When you do this, a new document type editor will appear for you which is automatically set to be an element type:

image

However, when you give this new element type a name, it will still create a template for you on disk. We don't need a template for an element type, it's not going to be an actual page. So this needs to be prevented.

We'd love some community help with fixing this, for now this will be a known bug for 8.7.0.

---
_This item has been added to our backlog [AB#8006](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8006)_

skttl:
Fixes #8646 

### Prerequisites

If there's an existing issue for this PR then this fixes #8646

### Description
I added the noTemplate property to the model used for creating the document type.

To test:
Create a block list, add a block, add a new doctype for the block. Verify that no templates are being created for the new doctype.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
@@ -141,6 +141,7 @@
             const editor = {
                 create: true,
                 infiniteMode: true,
+                noTemplate: true,
                 isElement: true,
                 submit: function (model) {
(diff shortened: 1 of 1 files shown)


poornimanayar (approved):
Hey @skttl ,

This totally works! Happy to approve. Thanks a lot for the work :-)

Poornima

PR #9019, opened 2020-10-01
Do not save media items when hitting enter

kjac:
### Prerequisites

### Description

For whatever reason the "Save" button on media items is of type "submit". This causes the editor to save automatically if you hit enter after loading the editor:

edit-media-prevent-enter-submit

For content items, the "Save" button is of type "button", which prevents this behavior... so I have changed the "Save" button on media items to be of type "button" as well.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/media/edit.html
+++ b/src/Umbraco.Web.UI.Client/src/views/media/edit.html
@@ -59,7 +59,7 @@
 
                     <!-- label-key="buttons_save" -->
                     <umb-button alias="save"
-                                type="submit"
+                                type="button"
                                 button-style="success"
                                 state="page.saveButtonState"
                                 label-key="{{page.submitButtonLabelKey}}"


nathanwoulfe (approved):

nathanwoulfe:
Thanks @kjac - well spotted! Strangely though I wasn't able to reproduce consistently... Will #9020 too.

bjarnef:
I think this should be a submit button.
With latest change in `v8/contrib` branch it seems nothing happens when the button is of type `button`, but changing the type to `submit` trigger the submit of the page.

75Ch7hjjhu

Tested in Chrome Version 86.0.4240.183

bjarnef:
Not sure why "save" and "save + publish" (both button elements with type button) do work.
In users section the button type is "button" as well, but in members section a submit button.

It would be great if this could be consistent across the sections for the primary save/submit button.
Maybe the submit button just need the `prevent-enter-submit` directive as well?

bjarnef:
Besides that I think `<button type="submit">` has something to say regarding accessibility.

and will enable the submission via common input gesture (ENTER press on text input elements for example):
https://itnext.io/form-accessibility-a-practical-guide-4062b7e2dd14

@BatJan @RachBreeze maybe you have some feedback on this?

BatJan:
@bjarnef I'm currently on leave so don't have much time to look into this - But a general rule of thumb is that if the button is placed inside a form and it's purpose is to submit the form, then it should be `type="submit"` - If it's a button placed inside a form that triggers a modal overlay then it should be `type="button"` - In this case it appears it should remain a "submit" and thus when you're inside the form it will be submitted when the enter key is hit.

PR #9020, opened 2020-10-01
Do not remove files from media editors when hitting enter

### LINKED ISSUE #8988: v8: Pressing "ENTER" key on a Media Node property removes the associated File
I have several custom properties on a MediaType. There is also the Upload control for attaching a file.

A few of the properties are Repeatable Textstrings type (Umbraco.MultipleTextstring). When using the "Add" button, this works fine, but if the ENTER key is pressed to get a "new line", the uploaded file gets removed.

Here is a small video demo: [
Umbraco 8 - Media Node ENTER bug.mp4 - With captions](https://drive.google.com/file/d/1SiRYUjRbQmFmyRGco_e8CG7PACtfWM7t/view?usp=sharing )
Edit_ Test - (Local) Media - hind-hpi local

### Umbraco version

I am seeing this issue on Umbraco version:  8.7.0

### Steps to reproduce

1. Create a MediaType with a Upload control and a repeatable textstring control.
2. Add a node of the type, add a file via the Upload. 
3. Add some textstring data, click the enter key after adding a line.
4. Notice the uploaded file gets cleared.

### Expected result

I'd expect that the file (and any other property data) would still be present.

### Actual result

The file gets removed.

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8988

### Description

As described in #8988, the media editors tend to remove their associated files when enter is pressed elsewhere when editing media items:

media-editor-enter-removes-media-before

Looks like this is caused by the accessibility changes in #6919 (approved by yours truly 🙈)... the `<button>` tags need to have `type="button"` to stop acting like a submit button. So this PR adds `type="button"` to those buttons.

Note that the issue fixed in #9019 makes this a tad hard to test, but if you look closely on this GIF, the issue *is* indeed fixed by this PR:

media-editor-enter-removes-media-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/upload/umb-property-file-upload.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/upload/umb-property-file-upload.html
@@ -43,9 +43,9 @@
                     </div>
                 </div>
                 <div>
-                    <button class="btn btn-link btn-crop-delete" aria-hidden="true" ng-click="vm.clear()"><i class="icon-delete red"></i> <localize key="content_uploadClear">Remove file</localize></button>
-                    <button class="sr-only" ng-if="file.isImage" ng-click="vm.clear()"><localize key="content_uploadClearImageContext">Click here to remove the image from the media item</localize></button>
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
When I say `button` you say: `TYPE="BUTTON"`!,When I say `button` you say: `TYPE="BUTTON"`!


Thanks @kjac 👍

Issue #8838, opened 2020-09-07
Node Actions menu should be disabled when user has selected items in the list.


nielslyngsoe:
If a user has selected items in the list-view, the "Action" button, top-right should be disabled.
That at least my preferred solution, if not posible, we can disabled the "Actions" all the time in the ListView tab.

![image](https://user-images.githubusercontent.com/6791648/92379997-ab0cda00-f108-11ea-8400-efb4b41623d1.png)

BatJan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/9021

PR #9021, opened 2020-10-01
8838: Node Actions menu should be disabled on listview

patrickdemooij9:
Based on issue: https://github.com/umbraco/Umbraco-CMS/issues/8838

I also saw that the _hide-menu="vm.page.hideActionsMenu"_ wasn't working because it had the wrong name, so I also fixed that. I wasn't able to get the actions disabled if you have items selected. Maybe someone else has an idea on how to do that. For now, it'll disable it when the listview app is shown and can be extended on later.


poornimanayar:
Hi @patrickdemooij9  ,

I have tested this and can see the Actions menu disabled on listview tab or selecting an item. I am going to ask @nielslyngsoe to cast an eye on this one as well :-)

Attached is a gif from my testing
actions

Poornima

nul800sebastiaan:
@patrickdemooij9 Thanks! The behavior is almost as intended, what was meant here is that the Actions dropdown should be disabled when you have selected one or more of the listview items, otherwise it should stay available. Would you be able to update this behavior please?

patrickdemooij9:
Hey @nul800sebastiaan! Yeah, that's what I was trying to achieve initially, but I couldn't figure out a way to see if an item was selected. So I went for this solution as that was the second option in the issue: ' if not posible, we can disabled the "Actions" all the time in the ListView tab.'

However, if there are some pointers on how to figure out if there is an item selected, I would love to finish it the intended way.

nathanwoulfe:
There's potentially an event emitted when a list view item is selected, which would provide a hook for toggling the state of the actions button. If there's no event, would be good to add one.

It makes better sense (in my head) that disabling happens when items are selected, not when the list view is the active app - it's an extra step to flip to the info app to enable actions, which is inconsistent with non-listview content.

patrickdemooij9:
Added an event for the selecting/unselecting of items and connected that with the logic of the enabling/disabling of the actions menu.

nathanwoulfe (approved):
@patrickdemooij9 this all looks good to me - works as expected for selecting, deselecting, select all, clear selection. Nice job!

PR #9022, opened 2020-10-01
Installer: Starterkit - Convert <a> to <button>

BatJan:
### Prerequisites

### Description
Noticed that we also had an `<a>` tag without an `href` attribute in the installer process where one picks the starterkit. Converted to `<button>` with this PR 👍


nul800sebastiaan:
Looking good!

image


Thanks @BatJan!

PR #9025, opened 2020-10-01
umb-tree-search-results directive view fixes

BatJan:
### Prerequisites

### Description
In this PR I have fixed the following
- Converted `<a>` to `<button`
- Added english fallback for "no search result" localization
- Changed `<i>` to use `<umb-icon>` - Please not an extra icon directive has been added to display the green checkmark when an item is checked. Previously this was handled using CSS on the `<i>` tag
- Some minor styling issues

** Where to find it **
The directive is being used in more places but one of them is in the "Copy" dialog on a node when using the search box.

umb-tree-search-results

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less
@@ -155,6 +155,14 @@ body.touch .umb-tree {
         &-link {
             display: block;
         }
+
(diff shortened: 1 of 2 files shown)


nathanwoulfe (approved):
Hey @BatJan this looks great - I can't get it running in the browser (nothing to do with the changes, just some super aggressive backoffice caching), but looking at the changes it all makes good sense.

PR #9027, opened 2020-10-01
Contextmenu: Add focus lock

BatJan:
### Prerequisites

### Description
In this PR I have done some cleanup putting the markup that used to wrap the `<umb-context-menu>` directive into the directive template since it's where it really belongs. I have then added some attributes to define the overlay as a dialog using `role="dialog"` attaching some aria labels for title and description as well.

Finally I have added the focus lock directive to keep the tabbing locked inside this overlay. You might wonder if it should be possible to tab out into the node tree but since it's an overlay of the available edit options for the node it should not 😃 

It's by purpose that I have not switched the `<i>` tag to `<umb-icon>` since there appears to be an issue with the icon used for the "Create content template" item in the menu where it gets smaller. So that's for another PR sometime.

Ideally when one triggers the "outsideClick" function or hits the "Reload" option the focus would be reset on the node in the content tree but that's for another PR since we need to keep track of a few things.

~~@nul800sebastiaan I added the tranlsation key to both `en.xml` and `en_us.xml` but for some reason the US language file appears to be ignored somehow? But I have not touched anything to do that so don't know if it's a general issue btw?~~

**Edit:** @nul800sebastiaan Nevermind me! It's been updated now - Not sure why it was not picked up in the first place but worked this morning 👍 

**Before**
context-menu-overlay-before

**After**
context-menu-overlay-after


nathanwoulfe (approved):
Lovely work as always, @BatJan. Given this was 2 months old (blame Hacktoberfest), there were conflicts in the lang files, which I've resolved for you. 🎅

Issue #8977, opened 2020-09-28
Multi Url Picker does not render external links correctly if user does not include http as part of Link Url


lesley-w:
A brief description of the issue goes here.

If a user creates a link using Multi Url Picker but does not include http or https in the url, the link renders the url as "siteurl" + "linkurl" e.g. http://www.mysite.com/www.facebook.com, which results in a broken link.

The url picker know that it is an external link, so should add http:// as a default if it has not been specified e.g. http://www.facebook.com

Please fill out the rest of the details in the issue template below. 
The more details you can give us, the easier it will be for us
to determine the cause of a problem.

-->

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.4

Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

### Bug summary

If a user creates a link using Multi Url Picker but does not include http or https in the url, the link renders the url as "siteurl" + "linkurl" e.g. http://www.mysite.com/www.facebook.com.
The url picker know that it is an external link, so should add http:// as a default if it has not been specified e.g. http://www.facebook.com

### Steps to reproduce

Create a link using Multi Url Picker data type, but don't inlcude http  in the url.
Render link. It will render Url as http://www.mysite.com/www.facebook.com, which creates a broken link

### Expected result

If Link.Type = LinkType.External and url does not begin with http or https, add http:// to start of url
Render link will render Url as http://www.facebook.com

### Actual result

Create a link using Multi Url Picker data type, but don't inlcude http  in the url e.g. www.facebook.com
Render link will render Url as http://www.mysite.com/www.facebook.com, which creates a broken link

nathanwoulfe:
Hi @lesley-w I can confirm the same happening on the latest v8 code - no protocol on the external URL results in a relative link being created.

Will mark as up-for-grabs.

umbrabot:
Hi @lesley-w,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

jordanc44:
I got this one!

[...]

PR #9029, opened 2020-10-02
[8977] prepending http to external links that have it missing

jordanc44:
### Prerequisites

https://github.com/umbraco/Umbraco-CMS/issues/8977

### Description
A conditional has been added. If a URL in the multiurl picker does not start with http, http:// is appended to the start of the URL string.

This is a fix for the issue here: https://github.com/umbraco/Umbraco-CMS/issues/8977

Steps to test: 
Create a link using Multi Url Picker data type, but don't inlcude http in the url.
The url when rendered on the page  should link to the external link.

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/ValueConverters/MultiUrlPickerValueConverter.cs
+++ b/src/Umbraco.Web/PropertyEditors/ValueConverters/MultiUrlPickerValueConverter.cs
@@ -53,7 +53,7 @@ namespace Umbraco.Web.PropertyEditors.ValueConverters
                 foreach (var dto in dtos)
                 {
                     var type = LinkType.External;
-                    var url = dto.Url;
+                    var url = dto.Url.StartsWith("http") ? dto.Url : "http://" + dto.Url;
 
                     if (dto.Udi != null)
                     {


c9mbundy:
Would '//' be an option here - to avoid potential mixed-content issues?

ronaldbarendse:
This should absolutely _not_ be merged as-is, as it completely breaks relative URLs, `mailto:`, etc. and also doesn't null-check the `dto.Url` property (causing a `NullReferenceException` is the property is null)!

The fix should be to ensure editors enter correct URIs, e.g. showing a suggestion when entering a URL that doesn't contains a scheme.

ronaldbarendse:
BTW, if you really want to use this logic, you can inherit from the default `MultiUrlPickerValueConverter` and override the `ConvertIntermediateToObject` method 👍

lesley-w:
How about just adding some info text below the Link box instead of, or in addition to, validation warnings? 
e.g. 
Links to external website should start with http:// or https://
Email links should start with mailto:
Phone number links should start with tel:

Content editors change over time, and usually without our knowledge, so it's hard to ensure that information regarding how to enter a link correctly does not get lost over time

ronaldbarendse:
@lesley-w Adding that much explanation text shouldn't be necessary; a lot of editors already know this and it's not even relevant if you're selecting content or media.

Like discussed on the issue, adding validation might be hard to get right (and we don't want to block correct input that just doesn't pass the validation tests).

IMHO we should help editors prevent making the most obvious mistake: adding a domain without a scheme. This could be as easy as showing a simple warning message when the link starts with `www.`.

lesley-w:
This is what I have done for the current project I am working on:
image

c9mbundy:
If the editor ignores warning and has still neglected to provide the protocol, then a non-specific // might be a safe option - i.e. until they tell us otherwise, use whatever the site is using.

ronaldbarendse:
> non-specific `//` might be a safe option

The only 'safe' option is to use `http://`, as you don't know if the linked site actually supports HTTPS (even though it absolutely should by now). And having it depend on how you're accessing the current site might then even cause the links to only work when accessing it using HTTP and not when using HTTPS (assuming the current site not automatically redirect HTTP to HTTPS). Mixed content also isn't the issue here, as it's not embedded content, but just a link to an external site.

Either way, this PR breaks links with relative URLs (`/manually-entered-link`) or other schemes like `mailto:` and `tel:`. Keep in mind: 'external' links don't have to go to other sites; they are links that don't have a content or media item picked.

nathanwoulfe:
Does this need then a more detailed resolution - checking the URL starts with `/`, `//`, `http://`, `https://`, `mailto:`, `tel:` etc to determine what (if anything) needs to be appended?

PR #9030, opened 2020-10-02
Replace obsolete usages with injected services in UmbracoHelper

kjac:
### Prerequisites

### Description

47 of 123 (or roughly 40%) of the solution build warnings all originate from internal usages of `Obsolete` marked properties in `UmbracoHelper.cs`. 

image

Since the relevant services are already injected into `UmbracoHelper`, I figure there is no reason not to use them directly and thus get rid of those warnings. So I have done just that. 

There are no noticeable changes to test. But in order to validate the change, you can use this template (it assumes the content being rendered has a published and render-able child). If the template renders successfully, everything works 😄 

```cshtml
@{
  Layout = null;
}
<html>
  <body>
    <ul>
      <li>What does Culture Dictionary think the current culture is? <b>@Umbraco.CultureDictionary.Culture</b></li>
      <li>Does Membership Helper think I have access? <b>@Umbraco.MemberHasAccess(Model.Path)</b></li>
      <li>Does Membership Helper think I am logged in? <b>@Umbraco.MemberIsLoggedOn()</b></li>
      <li>What does Content Query think the current content name is? <b>@Umbraco.Content(Model.Id).Name</b></li>
    </ul>
    <p>
    Here is the Compoment Rendererer at work:
    </p>
    <xmp>
      @Umbraco.RenderTemplate(Model.FirstChild().Id)
    </xmp>
  </body>
</html>
```

### DIFF
--- a/src/Umbraco.Web/UmbracoHelper.cs
+++ b/src/Umbraco.Web/UmbracoHelper.cs
@@ -141,7 +141,7 @@ namespace Umbraco.Web
         /// <returns></returns>
         public IHtmlString RenderTemplate(int contentId, int? altTemplateId = null)
         {
-            return ComponentRenderer.RenderTemplate(contentId, altTemplateId);
+            return _componentRenderer.RenderTemplate(contentId, altTemplateId);
         }
 
         #region RenderMacro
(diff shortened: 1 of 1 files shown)


nathanwoulfe (approved):
Hi @kjac all works fine, changes make sense to me.

nul800sebastiaan:
Thanks @kjac! 👍

PR #9031, opened 2020-10-02
Add Link and BeginLink extension methods

ronaldbarendse:
### Prerequisites

### Description
Similar to PR https://github.com/umbraco/Umbraco-CMS/pull/8840, this adds additional `HtmlHelper` extension methods to render links (`<a href=""></a>`) in Razor views, either using the Umbraco `Link` model or only a URL and text.

The `Link` extension methods that accept a `Link` model automatically use the URL, name and even target (open in new tab or window) to generate the link. When the target is set to `_blank`, it automatically adds `rel="noopener"` (unless there's already a `rel` attribute added to the `htmlAttributes` parameter), [avoiding performance and security issues](https://web.dev/external-anchors-use-rel-noopener/). Whenever the `Link` is `null`, nothing will be rendered, so no additional null-checks have to be added to the views. It even allows using a Razor template (with the `Link` assigned to the special `item` variable):
```cshtml
@Html.Link(Model.Button, htmlAttributes: new { @class = "button" })
@Html.Link(Model.Button, @<text><i class="fab fa-umbraco"></i> @item.Name</text>, new { @class = "button" })
```

The overloads that accept a URL and text work very similar:
```cshtml
@Html.Link(Model.Url, "Example text", new { @class = "button" })
@Html.Link(Model.Url, @<text><i class="fab fa-umbraco"></i> Example text</text>, new { @class = "button" })
```

----

Whenever you need to wrap the link around more HTML, you can use the `BeginLink` extension methods and wrap it in a `using` block (just like `BeginForm`). If the `Link` is `null`, the link is not rendered, but the content within is.

```cshtml
@using (Html.BeginLink(Model.Button, new { @class = "button" }))
{
    <i class="fab fa-umbraco"></i> Example text
}

@using (Html.BeginLink(Model.Url, new { @class = "button" }))
{
    <i class="fab fa-umbraco"></i> Example text
}
```

-----

The `BeginLink` uses a `DisposableTagScope` to render the start and end tags using an `IDisposable`. Because this can be used in very useful ways, I've added a bonus extension method: `BeginTagScope` 🎉 This allows conditionally rendering start/end tags by using a conditional attribute or manually calling the `Start` (or `End`) methods:

```cshtml
@using (Html.BeginTagScope("strong", new { @class = "root-node" }, Model.Parent == null)) {}
```

```cshtml
@using (var list = Html.BeginTagScope("dl", new { @class = "contact-details" }))
{
    if (!string.IsNullOrEmpty(Model.Telephone))
    {
        list.Start();

        <dt>Telephone</dt>
        <dd>@Model.Telephone</dd>
    }

    if (!string.IsNullOrEmpty(Model.Email))
    {
        list.Start();

        <dt>Email</dt>
        <dd>@Model.Email</dd>
    }
}
```

### DIFF
--- a/src/Umbraco.Web/Mvc/DisposableTagScope.cs
+++ b/src/Umbraco.Web/Mvc/DisposableTagScope.cs
@@ -1,0 +1,160 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Web.Mvc;
+using Umbraco.Core;
+
+namespace Umbraco.Web.Mvc
+{
+    /// <summary>
+    /// Class for creating HTML elements with seperate start and end tags.
+    /// </summary>
+    /// <seealso cref="Umbraco.Web.Mvc.IDisposableTagScope" />
+    public class DisposableTagScope : IDisposableTagScope
+    {
+        #region Fields
+
+        /// <summary>
+        /// The tag builder.
+        /// </summary>
+        private readonly TagBuilder tagBuilder;
+
+        /// <summary>
+        /// Indicates whether the tag is started.
+        /// </summary>
+        private bool started = false;
+
+        /// <summary>
+        /// Indicates whether this instance is disposed.
+        /// </summary>
+        private bool disposed = false;
+
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Gets the view context.
+        /// </summary>
+        /// <value>
+        /// The view context.
+        /// </value>
+        protected ViewContext ViewContext { get; }
+
+        #endregion
+
+        #region Constructors
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DisposableTagScope" /> class.
+        /// </summary>
+        /// <param name="viewContext">The view context.</param>
+        /// <param name="tagName">Name of the tag.</param>
+        /// <param name="htmlAttributes">The HTML attributes.</param>
+        /// <exception cref="System.ArgumentNullException">viewContext</exception>
+        public DisposableTagScope(ViewContext viewContext, string tagName, object htmlAttributes = null)
+        {
+            this.ViewContext = viewContext ?? throw new ArgumentNullException(nameof(viewContext));
+
+            var tagBuilder = new TagBuilder(tagName);
+            var htmlAttributesDictionary = htmlAttributes as IDictionary<string, object> ?? HtmlHelper.AnonymousObjectToHtmlAttributes(htmlAttributes);
+            htmlAttributesDictionary.RemoveAll(kvp => kvp.Value == null);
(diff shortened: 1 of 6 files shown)


DanDiplo:
@ronaldbarendse @nul800sebastiaan Could you think about how you name any new HTML helpers to avoid clashing with existing helpers people might have used? I've already written an `@Html.Link()` helper which I've used in dozens of projects and I'm sure I'm not the only one. I'm worried that if you add this to core it will clash with that, causing compilation and runtime errors.

Maybe could you prefix any Umbraco helpers with `Umb` or similar? eg. `@Html.UmbLink()` - this would help differentiate them from existing helpers. Or, at the very least, place them in a new namespace? Thanks for your consideration.

ronaldbarendse:
@DanDiplo That will always be the case for new things that are added to Umbraco and the same advice could be applied to your own extension methods 😉 That said, these extension methods are already defined in a new namespace: `Umbraco.Web.Mvc.Html` 🎉 

@nul800sebastiaan We should probably also add this new namespace to the `Views\Web.config` file (initial install) and config transform (on upgrade), so you don't have to manually do this (and require additional documentation just to get started with these new extension methods):
https://github.com/umbraco/Umbraco-CMS/blob/83395eb9943c51178ec2ccbc8bc9f281eb5b83b2/src/Umbraco.Web.UI/Views/Web.config#L23
https://github.com/umbraco/Umbraco-CMS/blob/83395eb9943c51178ec2ccbc8bc9f281eb5b83b2/build/NuSpecs/tools/Views.Web.config.install.xdt#L13-L15

If you already have an existing `Link` extension method, you'd either remove your own implementation or remove the newly added namespace. I think that's a better approach then renaming it to a more obscure name (which might still conflict with custom code, although I can imagine the chances might be smaller for someone to have named it `UmbLink`)...

ronaldbarendse:
For Umbraco 9, we could convert these HTML helpers into tag helpers (in the `Umbraco.Web.Mvc.TagHelpers` assembly), so they can all be added using `@addTagHelper *, Umbraco.Web.Mvc.TagHelpers` and/or specific ones deleted using `@removeTagHelper Umbraco.Web.Mvc.TagHelpers.LinkTagHelper, Umbraco.Web.Mvc.TagHelpers`.

AaronSadlerUK:
Picking up on this...

Surely Umbraco specific helpers should live within the Umbraco helper... That way it lives within the Umbraco eco-system and would stop these clashes...

ronaldbarendse:
@AaronSadlerUK Not really, as some extension methods actually require the `HtmlHelper` instance (e.g. to get the `ViewContext`). They should (and do) live in their own Umbraco namespace, which you would need to import (add a `using` for) before they're available...

PR #9032, opened 2020-10-02
Fix, update and extend Range model

ronaldbarendse:
### Prerequisites

### Description
The `Range` model is used in Umbraco as return type in the `Slider` data type, but doesn't allow an easy way to format the minimum and maximum value.

This PR adds 2 additional `ToString` overloads accepting either a single format or a range format, differentiating based on whether the minimum and maximum are equal or not, making the following possible:

```cshtml
@Model.Range.ToString("only {0}", "from {0} to {1}")
```

Besides that, I've also done the following:
- Update XML documentation
- Implement `IEquatable<Range<T>>` and override `Equals` and `GetHashCode`
- Fix `ToString()` to use culture invariant formatting 🐞 
- Add tests 🎉


mikecp:
Hello @ronaldbarendse 
Thank you for this PR! 
Some nice additions there 👍,  and tests 🥇 !!!

Maybe just one small suggestion: it could be interesting to call the "generic" ToString() method on line 57 from the default ToString() on line 36 as well, so that there is only one place where the string.Format logic is called. Something like:

`public override string ToString() => this.ToString("{0},{1}", CultureInfo.InvariantCulture); `

What do you think?

mikecp (approved):

mikecp:
Hi @ronaldbarendse ,
I saw that you did not have time to react to my suggestion yet, so I took the liberty of making the small suggested change myself, in order to be able to merge this interesting PR. I hope you won't mind.

Thanks again for this great work 👍 !

PR #9033, opened 2020-10-02
Add Root<T> and Breadcrumbs extension methods for IPublishedContent

ronaldbarendse:
### Prerequisites

### Description
This adds a generic `Root<T>` extension method to `IPublishedContent` to get the root content in a specific (ModelsBuilder generated) content type, e.g. to get the `Home` content in a Razor master view:

```cshtml
@{
    var home = Model.Root<Home>();
}
```

It also adds a `Breadcrumbs` extension method that basically gets the ancestors in reverse order, until a specific level or content type is found and with/without itself:

```cshtml
@foreach (var breadcrumb in Model.Breadcrumbs<Home>())
{
    <a href="@breadcrumb.Url()">@breadcrumb.Name</a>
}
```

I've also fixed the obsolete usage of both `CreatorName` and `WriterName` properties (https://github.com/umbraco/Umbraco-CMS/pull/7544) with the extension methods, which was very easy, as the `IUserService` was already available 🎉

### DIFF
--- a/src/Umbraco.Web/PublishedContentExtensions.cs
+++ b/src/Umbraco.Web/PublishedContentExtensions.cs
@@ -1155,8 +1155,8 @@ namespace Umbraco.Web
                                     { "NodeTypeAlias", n.ContentType.Alias },
                                     { "CreateDate", n.CreateDate },
                                     { "UpdateDate", n.UpdateDate },
-                                    { "CreatorName", n.CreatorName },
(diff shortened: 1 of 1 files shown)


nathanwoulfe (approved):
Hi @ronaldbarendse, apologies for taking so long to get to this... All works nicely, breadcrumbs in particular will be a useful addition.

PR #9034, opened 2020-10-02
Remove build warnings for internal classes

kjac:
### Prerequisites

### Description

There is a bunch of obsolete usage going on internally in the core. I have tried to root some of them with this PR - the ones that are non-breaking and "safe" to do. There probably are loads more that could be safely fixed, but I reckon it's best to do this in baby steps 😄


ronaldbarendse:
I've had this exact PR in mind for this month after seeing all the obsolete warnings after a build 😉

Just note that some internal things might also be used in Deploy, Forms and/or Headless (which shouldn't be required: either make some things public of move them into the CMS):
https://github.com/umbraco/Umbraco-CMS/blob/ad53b7630837de8db7bc7871dacea9af2a51e4f4/src/Umbraco.Core/Properties/AssemblyInfo.cs#L25-L36
https://github.com/umbraco/Umbraco-CMS/blob/ad53b7630837de8db7bc7871dacea9af2a51e4f4/src/Umbraco.Web/Properties/AssemblyInfo.cs#L22-L33

PR #9037, opened 2020-10-02
umb-search-filter: Convert i to use umb-icon

BatJan:
### Prerequisites

### Description
I had a look around in the backoffice and was suprised to see that the `<umb-search-filter>` directive was not being used more since I had a vague memory that I switced some of the old patterns to use this directive - Turns out I was wrong. So some more PR's changing those patterns to use the `<umb-search-filter>` will be incoming 😄 

**Update:** The directive has also been extended to accept a css-class parameter, since I foresee we might need it in some scenarios 😃 

**Update 2:** I realised that we needed to add some more "scoped" styling for the directive since currently we would run into issues with using CSS modifiers in other context - Therefore the 3 places where the directive is being used already has also been updated.

Therefore I want to start out by changing the `<i>` to `<umb-icon>` in the directive so we get it out of the box with the other incoming PR's 👍

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-search-filter.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-search-filter.less
@@ -1,0 +1,28 @@
+html .umb-search-filter {
+    position: relative;
+    height: 30px;
+
+    &__input {
+        padding-left: 30px;
+        padding-right: 6px;
+        width: 190px;
+        margin: 0;
+
+        &.w-100 {
+            width: 100%;
+        }
+
+        &.mb-15 {
+            margin-bottom: 15px;
+        }
+    }
+
+    .icon-search {
+        color: #d8d7d9;
+        position: absolute;
+        top: 0;
+        bottom: 0;
+        left: 10px;
+        margin: auto 0;
+    }
+}
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbsearchfilter.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbsearchfilter.directive.js
@@ -75,7 +75,8 @@
             labelKey: "@?",
             onChange: "&?",
             autoFocus: "<?",
-            preventSubmitOnEnter: "<?"
+            preventSubmitOnEnter: "<?",
+            cssClass: "@?"
         }
     };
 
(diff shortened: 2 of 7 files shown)


poornimanayar:
Hi @BatJan ,

The changes look fine. I checked and verified the change against datatypepicker, macro picker and redirecturls dashboard which seems unaffected by this. Is this correct? Let me know and I can merge this in 👍 

Poornima

poornimanayar (approved):

poornimanayar:
Approved and merged! Thanks for the work @BatJan !

src/Umbraco.Web.UI.Client/src/less/components/umb-search-filter.less
```diff
@@ -0,0 +1,28 @@
+html .umb-search-filter {
```
bjarnef, line 1: @BatJan is there a specific need to include `html` here? Unless `umb-search-filter` can be a class on `html` element itself?

BatJan:
@bjarnef It's due to specificity. I would like to avoid using `!important` and adding html as the first selector makes that possible and provides the needed specificity. It's been a long while and I have not touched any code for a little over 2 months but I think I faced an issue where the declaration was too generic changing the appearance elsewhere. Maybe in a scenario where the class was used some place where it might not have belonged... Can't really remember it now 🤷 😄

src/Umbraco.Web.UI.Client/src/less/components/umb-search-filter.less
BatJan: @bjarnef It's due to specificity. I would like to avoid using `!important` and adding html as the first selector makes that possible and provides the needed specificity. It's been a long while and I have not touched any code for a little over 2 months but I think I faced an issue where the declaration was too generic changing the appearance elsewhere. Maybe in a scenario where the class was used some place where it might not have belonged... Can't really remember it now 🤷 😄

bjarnef: @BatJan it just seemed a bit unnecessary and the `w-100` which already exists as a utility class. However I tried removing this, but then the `forms.less` overwritten to padding on the input styling.
Anyway I fixed some issues after the `icon-search` class was removed from `<umb-icon>` in another PR.
https://github.com/umbraco/Umbraco-CMS/pull/9631

PR #9039, opened 2020-10-02
Improved UX for deleting properties and groups in content types

### LINKED ISSUE #8946: Improve UX of deleting properties of a DocumentType in the DocumentTypeEditor.
I see two sub tasks to make this happen:

**A.**

I would like to use the Delete Dialog for deleting a property of a DocumentType in the DocumentTypeEditor.

This screenshot shows the current experience:
image

Instead, we should use this Dialog with some proper text (let the user know that the data will be removed etc.), and the buttons should say "Cancel", "Delete" (not Yes, Delete).
image

Hopefully, this dialog sets focus on the "Delete" button and enable the user to press "Enter" to confirm.

**B.**

If user has focus on the PropertyEditor-button and press the key "Delete" we want to open this dialog as well..
image

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8946

### Description

This PR updates the UX for deleting properties and groups in the content type builder. See #8946 for more details - I expanded it to apply to property groups as well.

With this PR, a delete confirmation dialog is displayed whenever you attempt to delete a property or a group. Once the dialog is confirmed or cancelled, the content type editor will resume a proper keyboard focus, making it possible to continue from where you left. Please note that #9004 needs to be merged in for that to work.

Here's a GIF that shows the PR in action:

delete-ux


kjac:
@nielslyngsoe for your reviewing pleasure 😄

nielslyngsoe (approved):
I love this, great that you brought the focus into this as well.

nielslyngsoe:
@kjac thanks for pointing me out, Then I better do this review. ;-)
Its work well, BUT the focus doesn't work for me, I never have the focus changed by code, neither when 
I interact with mouse or keyboard, same for both "Confirm" and "Cancel". Could you test that it still works for you?

kjac:
@nielslyngsoe I think you might have been seeing this because #9004 wasn't merged into the branch (see also initial issue description). I have merged from contrib now to include the fixes in #9004, and the PR WorksOnMyMachine out of the box now:

delete-property-auto-focus

nul800sebastiaan (changes requested):
It's looking good @kjac but I really think it should specify what will be deleted, the message `Are you sure you want to delete?` is a bit vague now. Not sure what I'm deleting (yes, I know I JUST clicked the delete button). Maybe just `Are you sure you want to delete this property?` (or `this group`) is enough?

kjac:
@nul800sebastiaan I'd be happy to amend, but @nielslyngsoe might need to have a say here - the message is pretty strictly defined in the linked issue (#8946)

nielslyngsoe:
Hi @kjac, looks good, and great to hear that the focus works.
In regards to the message, it was not intended to be specific from my point of view. So I think Sebastiaan is right that it would be help full if the message was more descriptive.
I think it would be even better is the name of the prop was present in the message, along these lines:

"Are you sure you will delete the property *MyProp*"

I think you might come up with something even better, I'm not sure whether data is lost by doing this(like can you save and than create this same property again on a later stage and still have the data?), but if so we should make that clear as well.

nathanwoulfe:
@kjac if you have a bit of spare time (ha, who does?!) to make the updates here, I can merge when ready.

OwainWilliams:
@kjac Have you had a chance to look in to this further? 

Thanks.

PR #9045, opened 2020-10-02
ContextDialog: Accessibility improvements

BatJan:
### Prerequisites

### Description
I have improved the accessibility of the context dialog overlay adding `role="dialog`, ´aria-labelledby´, `aria-describedby` and a description text + of course added keys in the languages files for `en` and `en_us`.

I would have loved to add the focus lock as well but in it's current state it's not quite ready yet - I will give it a try again once I've finished the #8522, which contain some changes that takes view changes into account. View changes are exactly the issue here - The current focus lock is not aware if something change state and becomes visible. But once it's fixed I'll give this another shot :-)


nathanwoulfe (approved):
Thanks @BatJan, lovely work as always.

Issue #8353, opened 2020-06-29
8.6.3 install screen responsive?


mistyn8:
Hopefully a picture speaks a thousand words?

![image](https://user-images.githubusercontent.com/12418214/86053096-e7a8bf00-ba4f-11ea-809e-51d0580a32e6.png)

nathanwoulfe:
Thanks @mistyn8, could certainly do with some work. Question for you - how do you rate your CSS chops? Feel free to tidy this up and submit a PR. Marking it as up-for-grabs in case any one feels like picking this up.

cheers
Nathan

umbrabot:
Hi @mistyn8,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
I just grabbed the Umbraco v8.6.3 from https://our.umbraco.com/download/ and can't reproduce the issue.

![image](https://user-images.githubusercontent.com/2919859/86259716-a71a8400-bbbc-11ea-9f7e-306ea9db75c8.png)

Are you sure you don't have any browser extensions which are causing this?

bjarnef:
Okay, it looks like it is an issue, when viewport width < 768px.

mistyn8:
@bjarnef yep should have put more than just responsive in the title. :-)

blueboxes:
This seems to be an issue on all forms as the media query that puts the label on new line does not override the text-alignment.

This PR fixes it https://github.com/umbraco/Umbraco-CMS/pull/9047

PR #9047, opened 2020-10-02
Issue 8353 - Fix of issue that caused small screens labels alignment

blueboxes:
There's an existing issue for this PR see https://github.com/umbraco/Umbraco-CMS/issues/8353

To reproduce the issue reduce the issue on the install screen width to less than 767px and form labels are placed on a new line, however the labels are still right aligned.

before

With this fix it sets them back to default:

after-fix


blueboxes:
I have added a second fix. The width was fixed so I have set it to a % for smaller screens but with a max width for larger screens

Broken

After Umbraco

nathanwoulfe (approved):
Thanks for being patient @blueboxes - I've verified this one, works nicely.

PR #9048, opened 2020-10-02
Do not show cryptic confirm dialog when editing group

### LINKED ISSUE #9036: Trying to remove a section from a user group tries to remove user from group
If I try to remove the Translation section from the Administrator Group I get an error indicating that I want to remove the user "undefined" from the actual group.

## Umbraco version

I am seeing this issue on Umbraco version: 8.8.0

Reproduction
------------

### Bug summary

Deleting a section from a user group shows a weird error message related to user deletion.

### Specifics

* Umbraco Version 8.8.0
* Chrome Brwoser 83.0.4103.116
* not checked in other browsers
image

### Steps to reproduce

* Go to the User section
* Switch to the user groups
* Select the administrator group
* Click on "remove" of a section
* The error message should appear

### Expected result

The section should be removed from the editor without any error message

### Actual result

An error message is shown related to user not to user group indicating a user will be deleted

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #9036

### Description

Looks like #8739 was a tad too eager in showing confirmation dialogs. It resulted in somewhat cryptic dialogs when removing sections or granular permissions from a user group:

image

image

This PR ensures that we retain the confirmation dialog when removing users from groups, but avoid showing confirmation dialogs when removing sections or granular permissions:

group-delete-items-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/users/group.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/users/group.controller.js
@@ -23,7 +23,9 @@
         vm.openContentPicker = openContentPicker;
         vm.openMediaPicker = openMediaPicker;
         vm.openUserPicker = openUserPicker;
-        vm.removeSelectedItem = removeSelectedItem;
+        vm.removeSection = removeSection;
+        vm.removeAssignedPermissions = removeAssignedPermissions;
+        vm.removeUser = removeUser;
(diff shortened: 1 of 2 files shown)


nathanwoulfe (approved):
@kjac works perfectly, thank you good sir.

PR #9050, opened 2020-10-02
Listview: Fix icon-picker a11y

BatJan:
### Prerequisites

### Description
In this PR I have made the following changes

- Converted `<div>` to `<button>`
- Changed `<i>` to `<umb-icon>`
- Made an English fallback text to "Add icon" in case a tranlsation key does not exist on a given language

**View after changes**

iconpicker-listview-changes


nathanwoulfe:
Hi @BatJan I've pushed an update to this, to split model.value into the icon and color parts, so that we can bind only the icon to the icon attribute, and only the color to the class attribute in umb-icon. Nothing major, just cleans up the bindings on the icon directive

nathanwoulfe (approved):

PR #9051, opened 2020-10-02
Grid editors: Accessibility improvements

BatJan:
### Prerequisites

### Description
In my quest to improve both the semantics and accessibility of the backoffice I noticed that on the "media", "embed" and "macro" editors for the grid `ng-click` handlers where added to either `<div>` or `<button>` elements directly among other things that have now been addressed and fixed.

In the "macro" and "embed" views the changes have mainly been converting `<div>` to `<button>` and convert `<div>` elements inside `<button>` to be `<span>` since having `<div>` inside them is not valid.

In the "media" view the above has also been fixed and there have been added a visually hidden label for the caption input as well.

**The canged areas in the grid**

**Macro**
grid-macro-change

**Embed**
grid-embed-change

**Media**
grid-media-change


nathanwoulfe (approved):
@BatJan awesome, love it, amazing.
Fixed up the merge conflicts, all good to go.
🎅

PR #9052, opened 2020-10-02
Grid prevalues: Improve accessibility and semantics

BatJan:
### Prerequisites

### Description
I noticed that that `ng-click` was being used on a `div` some places here so it's been changed to be using `<button>` elements instead and the child elements have been changed to be `<span>` elements since having block level elements as childs of `<button>` elements is invalid markup.

I've also added a `<p>` around some texts and added a fallback value, which was missing in the view as well.

I tried changing from `<i>` to `<umb-icon>` but there were some styling issues that requires a bit more effort to resolve properly. So that's for another PR some time by someone.

**What things look like after the changes and what I've been changing the markup for**
grid-prevalue-changes

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.prevalues.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.prevalues.html
@@ -17,20 +17,20 @@
 
                 <li ng-repeat="template in model.value.templates" class="clearfix">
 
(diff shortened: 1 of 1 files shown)


nathanwoulfe:
@BatJan I've got a pending PR (#9410) which takes care of the other bits you've mentioned and cleans up all the styling. I'll merge this, then deal with the conflicts in my PR...

nathanwoulfe (approved):

PR #9053, opened 2020-10-02
Grid: Layoutconfig improvements

BatJan:
### Prerequisites

### Description
With this PR I have changes a `<div>` with an `ng-click` handler to a `<button>` instead and I fixed the semantics accordingly. I also added a proper fallback text to a localization element, which was missing.

**The changed area**
grid-layout-changes


nathanwoulfe (approved):
@BatJan I'm running out of ways to say thanks! 🤷‍♂️

nathanwoulfe:
@BatJan annnnnd just noticed the new button has no type attribute - would you mind adding that, and I'll merge it too

PR #9055, opened 2020-10-03
Make JsonLogViewer public to allow changing log directory

### LINKED ISSUE #7908: Log Viewer in backoffice breaks when changing log location
### Bug summary

When changing the location of the logs in serilog.config, the Log Viewer in the backoffice does not work anymore. 

### Steps to reproduce

* Start with clean Umbraco 8 site
* Change `serilog:write-to:File.path` in config/serilog.config to something else than the default
* Make sure no old logs file remain (i.e. delete App_Data/Logs/)
* Open Log Viewer in the backoffice

### Expected result

Log Viewer works

### Actual result

Log Viewer doesn't work: No logs are displayed and the following error is shown: "Failed to retrieve state if logs can be viewed".

In the logs the following error is logged: "System.IO.DirectoryNotFoundException: Could not find a part of the path 'D:\home\site\wwwroot\App_Data\Logs'" (at `Umbraco.Core.Logging.Viewer.JsonLogViewer.CheckCanOpenLogs(LogTimePeriod logTimePeriod)`)

### Specifics

Also (kind of) reported before [on the forums](https://our.umbraco.com/forum/using-umbraco-and-getting-started/99251-change-log-files-output-directory).

I know we can [replace the build-in LogViewer with our own](https://our.umbraco.com/documentation/getting-started/Backoffice/LogViewer/#implementing-your-own-log-viewer), but I feel this functionality is basic enough to work 'out-of-the-box' or with minimal extra effort.

Looking at the [JsonLogViewer](https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Core/Logging/Viewer/JsonLogViewer.cs) code, it looks like the constructor supports adding in a different `logsPath` than the default. However, this parameter is not used in [its only usage](https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Core/Logging/Viewer/LogViewerComposer.cs). We also cannot change this in our own code, as the `JsonLogViewer` class is internal, meaning we would have to copy the entire code to change this parameter.

Additionally, [line 65](https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Core/Logging/Viewer/JsonLogViewer.cs#L65) does not use the `_logsPath` field for some reason.

### Possible solutions

The preferred solution would be that the Core `LogViewerComposer` would build the `JsonLogViewer` in a way that passes the `logsPath`. I do not see an obvious way to do so at this time.

Another solution would be to allow users to use the `JsonLogViewer` to set a different `logsPath`. I think the steps would be:

* Make class `JsonLogViewer` public.
* Change line 65 to use the private field instead of the hardcoded path
* Update docs to reflect changes needed when changing Serilog location

The steps users would have to make would be:

* Add new Composer that calls `composition.SetLogViewer(_ => new JsonLogViewer(composition.Logger, logsPath: yourNewLogsPath));`

I'm not sure if any of the above would work, but if that is the preferred direction to take, I can look into creating a PR for this.

## Umbraco version

I am seeing this issue on Umbraco version: 8.6

rbottema:
### Prerequisites

If there's an existing issue for this PR then this fixes #7908.

### Description
As described in #7908, changing the log directory in serilog.config breaks the log viewer in the backoffice. Making the `JsonLogViewer` class public allows user code to change the `logsPath` so the log viewer works again.

#### Testing steps:

- Change the value of `serilog:write-to:File.path` in config/serilog.config to something else than the default (e.g.: `%BASEDIR%\App_Data\LogsTest\UmbracoTraceLog.%MACHINENAME%..json`
- Create a new Composer, replacing the default LogViewer with a new one with the file path changed, e.g.:
```
public class ChangeLogPathComposer : IUserComposer
{
    {
        composition.SetLogViewer(_ => new JsonLogViewer(composition.Logger, logsPath: IOHelper.MapPath("~/App_Data/LogsTest")));
    }
}
```
- (Optional) Delete the contents of the default folder (~/App_Data/Logs) to trigger an error if the logs are loaded incorrectly (otherwise it will just show the wrong log files)
- Start the site and open the log viewer in the backoffice. Verify that it works and shows the correct logs.

### Notes:

- The hardcoded path was already fixed in #8281 
- The Composer needs to use `IOHelper.MapPath`, is this desired, or should we call this method on the input in the constructor of `JsonLogViewer`?
- After this is merged, I will have a look at updating the docs to document this.

### DIFF
--- a/src/Umbraco.Core/Logging/Viewer/JsonLogViewer.cs
+++ b/src/Umbraco.Core/Logging/Viewer/JsonLogViewer.cs
@@ -9,7 +9,7 @@ using Umbraco.Core.IO;
 
 namespace Umbraco.Core.Logging.Viewer
 {
-    internal class JsonLogViewer : LogViewerSourceBase
+    public class JsonLogViewer : LogViewerSourceBase
     {
         private readonly string _logsPath;
         private readonly ILogger _logger;


nul800sebastiaan:
Thanks @rbottema - I haven't looked into this yet, but I have a question: wouldn't it be better if the logviewer reads the location of the log files from the config? It feels unintuitive to have to update both the config file and then have to add code to the site in order for the location to update completely.

rbottema:
That does make more sense! I will have to look into the best way to do this, as it doesn't seem that 'the location of the log files' is readily available in the code at the moment.

enkelmedia:
I would agree with this as well!

Maybe one solution would be to use a Environement Variable called %LOGBASE% or something and use this in the serilog.config and also read this in the JsonLogViewer

Serilog.config:
```
<add key="serilog:write-to:File.path" value="%LOGBASE%\UmbracoTraceLog.%MACHINENAME%..json" />
```

A drawback with this would be that one might think that they could set the path in serilog.config to %LOGBASE%\foo\UmbracoTraceLog.. which would break the JsonLogViewer... either way it would be bettern than the current setup.

nul800sebastiaan:
Ultimately, the LogViewer should "just" use whatever is configured in the config file, so somehow we need to figure out why it doesn't and fix that. I think we don't really want the LogViewer to be public at this time.

rbottema:
From the research I have done so far for this, 'whatever is configured in the config' is not that simple unfortunately, because of the way Serilog works with sinks and whatnot. I believe that the LogViewer is the only location in the code where the location of the logs matters/is used directly.

enkelmedia:
I've proposed a solution to this issue here: https://github.com/umbraco/Umbraco-CMS/issues/10143#issuecomment-848847021

And also, this could be combined with another idea to be able to override all these configurations: https://github.com/umbraco/Umbraco-CMS/issues/5035#issuecomment-848799230

All these hardcoded paths is really an issue in all most any V8-project that we run.

PR #8812, opened 2020-09-03
Add color picker component

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
Currently we have color picker using spectrum plugin and used in prevalues of Color Picker property editor. It would howeever be great to re-use this various places in backoffice. E.g. in a custom dashboard, a custom property editor or as prevalue editor in grid configuration. We have a `colorpicker` prevalue editor, but with this it would make it easier to add a new prevalue editor (or your own prevalue editor) allowing to re-use this.

So this it adding a new component `<umb-color-picker>`.

I have recently released a new property editor `Color Pickr` https://our.umbraco.com/packages/backoffice-extensions/color-pickr/ inspired from this vanilla JS plugin called Pickr: https://simonwep.github.io/pickr/ 🥳🎉🚀

I could use much of the same approach to create this component wrapping spectrum color picker and then use this instead in the prevalue editor for color picker property editor.

PR #9056, opened 2020-10-03
Use color picker component in block list editor configuration

bjarnef:
### Prerequisites

### Description
At the moment when the UX for icon color and background color in block list editor isn't great. There is not much guidedance regarding the input format or validation of this.

This PR re-use the `<umb-color-picker>` component indroduced in this PR: https://github.com/umbraco/Umbraco-CMS/pull/8812

r0vqMfSRiU


bjarnef:
With PR https://github.com/umbraco/Umbraco-CMS/pull/9094 we might want to allow empty (clear) in configuration of color picker component here, so it is possible to clear an previous selected color.

nielslyngsoe:
Oh yes, this would be great!

I think we should allow the user to have empty color, just an easy way to get back to 'default' :-)

bjarnef:
@nielslyngsoe it should now be possible to clear any existing selected color.

HioyHMMBrM

bjarnef:
@nul800sebastiaan @nielslyngsoe is it too late for this to be part of v8.10 with the other block list improvements? 😁🙈

nathanwoulfe (approved):
A nice UX improvement, thanks @bjarnef. I've tested and it works perfectly - makes sense to cherry pick for 8.10, but I'm not sure on the policy given the RC is released...

nielslyngsoe:
@bjarnef i see it has ben tagged with 8.11, so it didnt become part of 8.10. But with the new release cycles, 8.11 will be out in 6 weeks. So waiting for a release should hopefully not be a topic of much focus going forward. :-)

nathanwoulfe:
That tagging was me - given 8.10 RC is out, any new merge is labelled for 8.11. 

I guess if there was an RC2 due to finding something requiring a fix for release it could be pulled in, but that's not my call (and probably a risky precedent to set)

nielslyngsoe:
@nathanwoulfe all good, my conclusion was since this did not have a 8.10 tag it has missed the train. But we will have one shortly in the next year 👍  and yep, we don’t include new stuff after RC.

PR #9057, opened 2020-10-03
Make sure external links are not attempted to be opened inside preview iframe

### LINKED ISSUE #6960: v8: Preview mode cannot open external URLs in same tab
See title. Opening external links in the preview mode (only when they would open in the same tab) gives a browser error.

### Specifics

Tested with Umbraco 8.2.1 and Chrome 77.

### Steps to reproduce

- Add an external link on a page (e.g.: external link in a RTE)
- Open preview mode
- Click the external link

### Expected result

The link is opened, either within or outside the preview mode. Outside is acceptable, as the user can click 'back' in their browser anyway.

### Actual result

In Chrome, a gray page is shown with the test 'www.google.com refused to connect.'. Console shows the following: `Refused to display 'https://www.google.com/' in a frame because it set 'X-Frame-Options' to 'sameorigin'.`

rbottema:
### Prerequisites

If there's an existing issue for this PR then this fixes #6960

### Description
As discussed in #6960, when attempting to open an external link in the same tab inside preview, a browser error is displayed. This is because the link is attempted to be opened inside the iframe, which will not be allowed for most external URLs.

This PR modifies all external links that did not have their target set already to open in the same browser window, outside the iframe. This is not 100% ideal UX, but as per the discussion in #6960, it's likely the best option for now.

#### Testing notes

- Place a link to an external page (without 'open in new tab' set). I used the About Us page of the Starter Kit.
- Preview the page and click the link. Verify that you do not get a browser error, but the link is opened instead.

I have also tested the following scenarios I could think of, and they all seem to work fine:

- Internal links should open in the iframe
- Links with a target set (e.g. `_blank`) should not have their behavior overwritten
- Special links like `mailto:` should keep working
- Pages without links should not result in browser errors

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/preview/preview.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/preview/preview.controller.js
@@ -85,6 +85,13 @@ var app = angular.module("umbraco.preview", ['umbraco.resources', 'umbraco.servi
                 .fail(function () { console.log("Could not connect to SignalR preview hub."); });
         }
 
+        function fixExternalLinks(iframe) {
+            // Make sure external links don't open inside the iframe
+            Array.from(iframe.contentDocument.getElementsByTagName("a"))
+                .filter(a => a.hostname !== location.hostname && !a.target)
+                .forEach(a => a.target = "_top");
+        }
+
         var isInit = getParameterByName("init");
         if (isInit === "true") {
             //do not continue, this is the first load of this new window, if this is passed in it means it's been
@@ -188,6 +195,7 @@ var app = angular.module("umbraco.preview", ['umbraco.resources', 'umbraco.servi
         $scope.onFrameLoaded = function (iframe) {
             $scope.frameLoaded = true;
             configureSignalR(iframe);
+            fixExternalLinks(iframe);
 
             $scope.currentCultureIso = $location.search().culture || null;
         };


nathanwoulfe:
@rbottema is there a reason for using _top rather than _blank? _top updates the current tab, but if we used _blank the opened link doesn't mess with the preview view. Would be good too to add the `rel="noopener"` attribute too when setting the target

rbottema:
I used `_top` to make the link work as close to the actual functionality as possible: to open the link in the same tab. If we would open in a new tab, we would change the specified functionality of the link, which may also cause confusion ("I defined this to open in the same tab, why does it open in a new tab?").

If we settle for `_blank`, I agree that we should include `rel="noopener"`.

nathanwoulfe:
Cool cool, sounds good to me - noopener would be great, thanks

rbottema:
I'm confused: I propose we stay at `_top` to keep the behavior as close as possible. If we use `_top`, there is no need for `rel="noopener"`: that only applies for `_blank`.

nathanwoulfe:
Right you are - ignore me, it's been a long day...

nathanwoulfe (approved):

nathanwoulfe:
Thanks @rbottema, not sure why I didn't merge this after our discussion... Done now though. 🎄

PR #9058, opened 2020-10-03
Fix void element of umb-checkbox component

bjarnef:
### Prerequisites

### Description
Angular elements are custom HTML elements, which are not meant to be void elements (self-closing) although it did seem to work in previous Angular versions.

I already fixed this issue in https://github.com/umbraco/Umbraco-CMS/commit/12a4dbc7052aad1e0018396be27be82feb5b7827#diff-144204f44d67201e4a96574595d21b13 but is seems this was reverted when merging https://github.com/umbraco/Umbraco-CMS/pull/8213

In the case it only cause some addtional spacing in the logviewer filter.

**Current**

image

**After**

image


nathanwoulfe (approved):

nathanwoulfe:
Thank you @bjarnef, good find 👍

PR #9059, opened 2020-10-03
Don't log query result to avoid confusion

### LINKED ISSUE #7108: Possibly Incorrect Index Names Logs?
In the logs, anything that logs an index creation logs the index name as the result of the query to create the index, rather than the index name. So for example you get:

{"@t":"2019-11-07T12:11:24.9705346Z","@mt":"Create Index ({CreatedIndex}):\n {Sql}","CreatedIndex":-1,"Sql":"CREATE UNIQUE NONCLUSTERED INDEX [IX_umbracoDocumentCultureVariation_NodeId] ON [umbracoDocumentCultureVariation] ([nodeId],[languageId])","SourceContext":"Umbraco.Core.Migrations.Install.DatabaseSchemaCreator","ProcessId":14716,"ProcessName":"iisexpress","ThreadId":16,"AppDomainId":4,"AppDomainAppId":"LMW3SVC2ROOT","MachineName":"FRUITOPIA","Log4NetLevel":"INFO ","HttpRequestNumber":2,"HttpRequestId":"3d076713-ad77-4a30-a9ec-c938985d3dd7"}

Should "CreatedIndex" be the name of the index, or the result of the query? The naming is ambiguous.

rbottema:
The query result will always be -1, as the result is the amount of rows affected, but that is not applicable in these create queries. This may cause confusion as devs may wonder why for example 'CreatedIndex' is -1. This also seemed to be the only place where we actually use the result of `database.Execute`.

### Prerequisites

If there's an existing issue for this PR then this fixes #7108

### Description
Logging these parameters leads to confusion, so let's remove them. Logging the name of the created indexes and keys requires some rework of the code and I'm not sure it's really needed at this point, as the SQL already contains the name if needed.

#### Test notes

- Run the site on a clean database or without database
- Ensure the database is constructed correctly and the log entries do not contain the removed parameters

### DIFF
--- a/src/Umbraco.Core/Migrations/Install/DatabaseSchemaCreator.cs
+++ b/src/Umbraco.Core/Migrations/Install/DatabaseSchemaCreator.cs
@@ -445,14 +445,14 @@ namespace Umbraco.Core.Migrations.Install
             }
 
             //Execute the Create Table sql
-            var created = _database.Execute(new Sql(createSql));
-                    _logger.Info<DatabaseSchemaCreator>("Create Table {TableName} ({Created}): \n {Sql}", tableName, created, createSql);
+            _database.Execute(new Sql(createSql));
+            _logger.Info<DatabaseSchemaCreator>("Create Table {TableName}: \n {Sql}", tableName, createSql);
 
             //If any statements exists for the primary key execute them here
             if (string.IsNullOrEmpty(createPrimaryKeySql) == false)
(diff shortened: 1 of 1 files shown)


mikecp (approved):

mikecp:
Thanks @rbottema for this PR!
Looks all good to me :-)
And it removes some confusion indeed!

PR #9060, opened 2020-10-03
Introduce CachedPartial that takes a Func instead of an object

### LINKED ISSUE #7142: Allow passing a Func instead of object in CachedPartial
The [CachedPartial](https://github.com/umbraco/Umbraco-CMS/blob/a2551521e45e47da4318012704d4c129e659aaa8/src/Umbraco.Web/HtmlHelperRenderExtensions.cs#L78) extension is a great way to easily achieve partial caching. However, in our architecture, it has a great flaw. Calling the method requires passing in an already built model, and that is exactly where the possibly complex logic is that we want to cache!

Ideally, we pass a `Func` that builds the model, that is only executed if the partial is not cached already.

We have created our own extension for that, but unfortunately it is quite hacky, as there is no way to check if a certain partial is already cached. Instead, what we do is pass first `null` as model to `CachedPartial`. This will throw an exception if there is no cache yet built, but will return successfully if there is already a cache. We then catch the exception, building the model and calling `CachedPartial` again. The code is as follows (edited for brevity):

```CSharp
public static IHtmlString FuncCachedPartial(this HtmlHelper htmlHelper, string partialViewName, Expression<Func<object>> builder, int cachedSeconds = 3600)
{
    IHtmlString result = null;

    try
    {
        object model = null;
        result = htmlHelper.CachedPartial(partialViewName, model, cachedSeconds);
    }
    catch (Exception ex)
    {
        bool isCachedException = (ex.InnerException is InvalidOperationException && ex.InnerException.Message.Contains("but this dictionary requires a model item of type"));

        if (!isCachedException)
        {
            Diagnostics.Logging.Error(ex);
        }
    }

    if (result == null)
    {
        object model = builder.Compile()();
        result = htmlHelper.CachedPartial(partialViewName, model, cachedSeconds);
    }

    return result;
}
```

rbottema:
This allows postponing the building of the model passed to the partial to when it is actually needed, as the model really only needs to be built when the partial is not in cache (and when the contextualKeyBuilder doesn't use the model as input). This can be a great performance improvement if building the model is very expensive.

### Prerequisites

If there's an existing issue for this PR then this fixes #7142.

### Description
See above.

Usage is similar to the existing `CachedPartial`, but then a Func is passed instead, e.g.:
```
@Html.CachedPartial("~/Views/Partials/Home/Hero.cshtml", () => HeroBuilder.Build(Model))
```

#### Testing notes
To test the difference between this and the regular `CachedPartial`, we need to verify whether the method is actually only called once.

- Make sure caching is enabled, so `debug="false"` in the web.config
- Take an existing usage of `CachedPartial` like `@Html.CachedPartial("~/Views/Partials/Home/Hero.cshtml", Model)` and verify that it works
- Replace with the usage with a Func that has a side effect like logging. For example:
```
@Html.CachedPartial("~/Views/Partials/Home/Hero.cshtml", () => {
    global::Umbraco.Web.Composing.Current.Logger.Warn(typeof(object), "Builder called");
    return Model;
})
```
- Refresh the page a few times and verify that the log entry only appears once.

### DIFF
--- a/src/Umbraco.Web/CacheHelperExtensions.cs
+++ b/src/Umbraco.Web/CacheHelperExtensions.cs
@@ -36,16 +36,44 @@ namespace Umbraco.Web
             string cacheKey,
             ViewDataDictionary viewData = null)
         {
+            return appCaches.CachedPartialView(htmlHelper,
+                partialViewName,
+                () => model,
+                cachedSeconds,
+                cacheKey,
+                viewData);
+        }
+
+        /// <summary>
+        /// Outputs and caches a partial view in MVC
+        /// </summary>
+        /// <param name="appCaches"></param>
+        /// <param name="htmlHelper"></param>
+        /// <param name="partialViewName"></param>
+        /// <param name="builder"></param>
+        /// <param name="cachedSeconds"></param>
+        /// <param name="cacheKey">used to cache the partial view, this key could change if it is cached by page or by member</param>
+        /// <param name="viewData"></param>
+        /// <returns></returns>
+        public static IHtmlString CachedPartialView(
+            this AppCaches appCaches,
+            HtmlHelper htmlHelper,
+            string partialViewName,
+            Func<object> builder,
+            int cachedSeconds,
+            string cacheKey,
+            ViewDataDictionary viewData = null)
+        {
             //disable cached partials in debug mode: http://issues.umbraco.org/issue/U4-5940
             if (htmlHelper.ViewContext.HttpContext.IsDebuggingEnabled)
             {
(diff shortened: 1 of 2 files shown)


nzdev:
This func should have the UmbracoHelper as a parameter as otherwise people will end up capturing UmbracoContext and break caching

rbottema:
I'm not sure I understand, can you elaborate? If the `builder` Func is to be executed, it will always do so in the same request.

nzdev:
I think you are right. Nevermind :)

PR #9062, opened 2020-10-04
Adjustments of block list configuration

bjarnef:
### Prerequisites

### Description
This PR add a few adjustments to block list editor configuration.
It seems the event bubbling is stopped when clicking delete button. However when focus the delete button and hitting enter, it does seem to trigger the parent click the open the block configuration.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/blockcard/umb-block-card.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/blockcard/umb-block-card.html
@@ -1,4 +1,4 @@
-
+<div class="umb-block-card">
     <div class="__showcase" ng-style="{'background-color':vm.blockConfigModel.backgroundColor, 'background-image': vm.styleBackgroundImage}">
         <i ng-if="vm.blockConfigModel.thumbnail == null && vm.elementTypeModel.icon" class="__icon {{ vm.elementTypeModel.icon }}" ng-attr-style="{{'color:'+vm.blockConfigModel.iconColor+' !important'}}" aria-hidden="true"></i>
     </div>
@@ -6,6 +6,5 @@
         <div class="__name" ng-bind="vm.elementTypeModel.name"></div>
         <div class="__subname" ng-if="vm.elementTypeModel.description" ng-bind="vm.elementTypeModel.description"></div>
     </div>
-
     <ng-transclude></ng-transclude>
-
+</div>
--- a/src/Umbraco.Web.UI.Client/src/views/components/blockcard/umb-block-card.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/blockcard/umb-block-card.html
@@ -1,4 +1,4 @@
-<div class="umb-block-card">
+<div>
     <div class="__showcase" ng-style="{'background-color':vm.blockConfigModel.backgroundColor, 'background-image': vm.styleBackgroundImage}">
         <i ng-if="vm.blockConfigModel.thumbnail == null && vm.elementTypeModel.icon" class="__icon {{ vm.elementTypeModel.icon }}" ng-attr-style="{{'color:'+vm.blockConfigModel.iconColor+' !important'}}" aria-hidden="true"></i>
     </div>
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
@@ -106,7 +106,7 @@
                 return obj;
             });
 
-            var availableItems = vm.getAvailableElementTypes()
+            var availableItems = vm.getAvailableElementTypes();
 
             localizationService.localizeMany(["blockEditor_headlineCreateBlock", "blockEditor_labelcreateNewElementType"]).then(function(localized) {
 
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
@@ -67,8 +67,8 @@
                 });
             });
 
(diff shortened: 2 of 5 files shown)


bjarnef:
@nielslyngsoe do you have any idea why the hitting delete button using enter key also trigger the parent block card click event?

nathanwoulfe (approved):
Hi @bjarnef, this all looks good to me.
I've resolved the merge conflicts and added a fix for the event bubbling issue on the the remove button - it's a bit hacky, but it works. When the openBlockOverlay function runs, it now checks the original event and returns if it came from the delete button. Not the most elegant, but ok for now.

Not sure why stopPropagation wouldn't work on the key press - probably something to do with how the events are raised. I suspect AngularJs fires the ng-click on keydown, but keyup is the event that propagates and triggers the openBlockOverlay function. Maybe.

nathanwoulfe:
Will leave this unmerged for now, in case @nielslyngsoe has some wisdom to fix the events properly...

src/Umbraco.Web.UI.Client/src/views/components/blockcard/umb-block-card.html
```diff
@@ -1,10 +1,17 @@
-<div>
-    <div class="__showcase" ng-style="{'background-color':vm.blockConfigModel.backgroundColor, 'background-image': vm.styleBackgroundImage}">
-        <i ng-if="vm.blockConfigModel.thumbnail == null && vm.elementTypeModel.icon" class="__icon {{ vm.elementTypeModel.icon }}" ng-attr-style="{{'color:'+vm.blockConfigModel.iconColor+' !important'}}" aria-hidden="true"></i>
+
+    <div class="__showcase"
+        ng-class="{'--error':vm.elementTypeModel === null}"
+        ng-style="{'background-color': vm.blockConfigModel ? vm.blockConfigModel.backgroundColor : '', 'background-image': vm.styleBackgroundImage}">
+        <i ng-if="vm.blockConfigModel == null || vm.blockConfigModel.thumbnail == null" class="__icon {{ vm.elementTypeModel ? vm.elementTypeModel.icon : 'icon-block' }}" ng-attr-style="{{'color:'+vm.blockConfigModel.iconColor+' !important'}}" aria-hidden="true"></i>
```
bjarnef: @nathanwoulfe we can probably use `<umb-icon>` here as well and I wonder if we still need `!important` to force this color style if using `<umb-icon>` component 👍

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
```diff
@@ -188,28 +165,51 @@
...
-                editorService.open(overlayModel);
-            });
+                    var overlayModel = {
+                        block: clonedBlockData,
+                        title: data,
+                        view: "views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.overlay.html",
+                        size: "small",
+                        submit: overlayModel => {
+                            loadElementTypes()// lets load elementType again, to ensure we are up to date.
+                            TransferProperties(overlayModel.block, block);// transfer properties back to block object. (Doing this cause we dont know if block object is added to model jet, therefor we cant use index or replace the object.)
```
bjarnef, line 192: @nathanwoulfe I know this isn't specific related to the recent changes, but I think this function `TransferProperties()` should use camel case like `transferProperties()`. I also wonder if this couldn't be placed in a helper/service.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js#L13-L17

Or maybe we can just use `Utilities.copy(src, dst)` function?
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/utilities.js#L21

nielslyngsoe: Hi Bjarne

I think that Utilities.copy is doing a deep clone. Its important that we don't do a deep copy here, as we want the references to float over.

The reason for it to be with a capitalized first letter is that it's a Static method. As we do with Classes etc.

I think you should hold back a bit with all these micro-optimizations, as it becomes hard to track the code backwards when the code has been changed just for styling purposes. I'm mostly concerned about the changing of the methods to arrow functions without any other change or reason to do so.

bjarnef: @nielslyngsoe okay, makes sense. I haven't looked further at the differences on the "block" object and the "blockConfiguration" object, but I guess this object is simpler.

Anyway I was wondering if there isn't anywhere else in code, where we also do a "flat" copy as well and therefore this might be something to re-use.

nielslyngsoe: @bjarnef would be nice to investigate. Could for sure be useful in other places so refactoring it into a utility method would be nice.

bjarnef: @nielslyngsoe sure, but for now I think @nathanwoulfe added a workaround in this PR, to fix the annoying issue, where keypress on block delete button both trigger delete overlay and edit overlay 😅🙈

nielslyngsoe: @bjarnef Oh, yes. We should still have that. :-) But we should in general try to keep one PR for each feature/bug. So we should extract that anyway. That would make @nul800sebastiaan happy. Because we use PRs/Issues to get an overview of the changes and fixes made in a release.
In this way, it's also easier to later return to a feature/bug-fix, evolve it, pull it for a hotfix in an earlier version for a patch releases. Or eventually to revert it if it causes trouble. :-) I hope that makes sense :-)

PR #9063, opened 2020-10-04
V8/feature/update icon picker to use umb icon

MMasey:
This PR reintroduces the svg icons into the icon picker. 

This differs to the original approach in that I haven't made any changes to the original `getIcons` method in the `iconhelper.service.js`. That way, if anyone was utilising it for their package etc it won't break anything.

I've also made an update to the `umb-icon` directive, adding lazyload functionality. This paired with the built in caching from the icon service also meant that it was no longer necessary to manually add the svgstring value in the icon picker view so i've removed that from the "docs" too.

Fingers crossed this should cover all the bases.


src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/iconpicker/iconpicker.controller.js
```diff
@@ -43,10 +43,24 @@ function IconPickerController($scope, localizationService, iconHelper) {
...
+            .then(icons => {
+                vm.icons = icons;
+
+                // Get's legacy icons, removes duplicates then maps them to IconModel
+                iconHelper.getIcons()
+                    .then(icons => {
+                        if (icons && icons.length > 0) {
+                            let legacyIcons = icons
+                                .filter(icon => !vm.icons.find(x => x.name == icon))
+                                .map(icon => { return { name: icon, svgString: undefined }; });
```
bjarnef, line 56: I would use `null` instead of `undefined` here since you actually assign a value to the property.

MMasey: Agreed, makes a lot of sense 👍

nul800sebastiaan:
@MMasey Thanks! Have you tested this with https://our.umbraco.com/packages/backoffice-extensions/content-node-icons/ to make sure that one doesn't break?

MMasey:
@nul800sebastiaan Ah perfect, not yet but now I know one to test against I shall.

MMasey:
> @MMasey Thanks! Have you tested this with https://our.umbraco.com/packages/backoffice-extensions/content-node-icons/ to make sure that one doesn't break?

Right, I've good a double whammy test for you.

First of installed the package you mentions and all seems to work find. I had a look at in the console and there does seems to be some code in the package itself that throws a type error, but the functionality still works as expected, so I imagine it just needs a bit of tweaking. From what I can tell though, nothing breaks.

icon plugin test

I then had a quick look for another icon package and installed this one https://our.umbraco.com/packages/backoffice-extensions/line-awesome-icon-pack/. As you can see it all seems to be working fine too, along side the other package too.

line awesome icon plugin test

Tagging @markadrake & @adam-werner for reference

nul800sebastiaan:
Good news/bad news for @markadrake - the same error appears in v8.7.0, so it's not due to these changes 

image

It's coming from this line:

```js
if (!data.icon || !data.iconColor) reject(null);
``` 

Not sure that `reject` is working as it shoud've stopped trying a few lines before on 

```js 
if (!response.data) reject(null);
```

src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js
```diff
@@ -43,13 +34,19 @@ This format is only used in the iconpicker.html
                 svgString: "=?"
             },
 
-            link: function (scope) {
-                
+            link: function (scope, element) {
                 if (scope.svgString === undefined && scope.svgString !== null && scope.icon !== undefined && scope.icon !== null) {
-                    var icon = scope.icon.split(" ")[0]; // Ensure that only the first part of the icon is used as sometimes the color is added too, e.g. see umbeditorheader.directive scope.openIconPicker
+                    const observer = new IntersectionObserver(_lazyRequestIcon, {rootMargin: "100px"});
```
nul800sebastiaan, line 39: What does this mean? I can not decipher it 😅

src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js
```diff
@@ -15,16 +15,7 @@ Simple icon
...
-    <umb-icon icon="icon-alert" class="icon-class" ng-click="doSomething()"></umb-icon>
-</pre>
-
-Manual svg string
-This format is only used in the iconpicker.html
-<pre>
-    <umb-icon 
-        icon="icon-alert"
-        svg-string='<svg height="50" width="50"><circle cx="25" cy="25" r="25" fill="red" /></svg>'>
-    </umb-icon>
```
nul800sebastiaan, line 27: So does this notation no longer work? Technically that's already a breaking change! 😅

nul800sebastiaan (commented):
Sweet! Well the good news is that this all works and it also works with @skttl's PR #8884 with no problems (yes I can override any icon I want! 🎉).

Can you talk a little bit about how the caching works please? I do see a lot fewer `GetIcon` and `GetAllIcons` requests, which is great. 

I left you some more comments and I want to avoid the wrath of @nielslyngsoe when I merge it so I'll ask him to have a look as well!

src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js
MMasey: No worries, the IntersectionObserver is a sweet native API supported by all modern browsers that will fire a callback when the observed DOM element comes into view. The `rootMargin: "100px"` essentially means that it will add an additional 100px around the element when it triggers the callback.

The `_lazyRequestIcon`  checks to see if the element is intersecting with the view according to the options, and if it sees that is it, it will fetch make the request, and then disconnect the observer as it is no longer needed.

src/Umbraco.Web.UI.Client/src/common/directives/components/umbicon.directive.js
MMasey: The notation will still work, but it's isn't being used anywhere in the umbraco backoffice code anymore. 

When you define a value for the `icon` attribute in the `umb-icon` directive, it will make a request looking for that icon. If it finds one it will return an svg string and will assign the value of that string to the `svg-string` attribute, which the directive uses internally to render out the svg. So the attribute is used by the directive anyway, but it just doesn't really make sense to set it directly anymore.

MMasey:
@nul800sebastiaan The caching happens client side by the `iconhelper.service`. When a request is made for either a single icon or all of them, if one or more icons are returned it will add it to an array in the service called `iconCache`. Every time a request is made for an icon, the service looks for it in the cache first, and if it can find it, will then make an actually request to the IconController API.

This means that if a person opens the icon picker, all of the icons are added to the cache. The main difference with the update is that rather then chucking them all into the DOM at once, which can be a bit much, it only renders the icons that are visible on the screen.

There are a number of things we could potentially do to take this further in the future too such as defining a list of the most commonly used icons and always loading them on initial page load. We could also look at integrating with a service worker in the Umbraco backoffice that could cache the icons locally, as opposed to that page load, it isn't too bad at the moment as angular does a fair amount of routing for us, but it can always be improved.

Hope this helps 🙂

nielslyngsoe (approved):
Looks good I'm going to test the code

nielslyngsoe:
The code works fine, it is though a bit slower to load the SVGs than the old solution. Do we have a concept to improve this in the future? :-) just asking :-)

nul800sebastiaan:
Cherry picked for 8.8.1 in 879d54b5259b237d7c6b82c285bee28c2f64dd7f

bjarnef:
@MMasey I just had a look at using `<umb-icon>` in help drawer as replacement of the traditional font icon using `<i>` element, but it doesn't seem to work well here, where it in general seems the request takes 3-5 seconds.

Not sure if it is specific related to this PR though.

Xl8bWOl7L2

bjarnef:
It seems `GetAllIconNames()` in a clean install request 606 icons and it doesn't take that long.
image

Not sure if could be an issue with multiple requests to /umbraco/backoffice/UmbracoApi/Icon/GetIcon?iconName=xxx and each of these need to lookup all icons:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Services/IconService.cs#L71-L72

image

MMasey:
It looks like the `ToList()` on the `iconNames` is really slowing things down. e.g.

**With `ToList()`**

get-all-icons-perf

**Without `ToList()`**

get-all-icons-perf-no-tolist

The ToList also isn't required here too. I've go a branch where I have removed it which I could use to make a PR but It would be good to tackle the clientside loading stuff too. Whilst the `iconhelper.service.js` does have some logic to help reduce the number of repeat requests I think that it still misses some when many icons are loaded concurrently which could be why we are seeing multiple requests for the same icon. I think the best thing to do would be to integrate a proper queue based or event based request system for requesting single icons.

On top of this I also noticed that @clausjensen (I think) added some code on the 22nd of September to help preload some icons in the `Umbraco/Views/Default.cshtml`. We could potentially use this approach to preload some more icons if required, or at least a similar approach.

nathanwoulfe:
Could this be done similar to the language files and made available from a client-side cache? Or punt em all into a server-side cache on startup? 

Shouldn't need to look up everything in each request (unless I'm missing something 🤷‍♂️)

clausjensen:
@MMasey The preloading done there was to ensure we do not try to request the icons from an authenticated endpoint when they are to be shown/used before you are logged in 😃 We don't want to expose the endpoint as a public one as it could be misused/introduce potential security issues, but we also need just a few of the icons available for the login screen where you are not yet authenticated. So preloading those few ones by outputting them via serverside rendering and then injecting them into the javascript service was a solution.

So I don't really think it makes a difference since the icons preloaded like this, would still be loaded through the same icon service back there. 

Also ..I'm pretty sure language files are cached somewhere as I think you wouldn't actually see a change to a language file reflected until you restart the application... to be honest I can't remember how this all works though as it is not something I've looked at for a long time ;)

PR #9054, opened 2020-10-02
Package section: Edit view

BatJan:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
With this PR I have changed the clickable headers from being `<div>` element to be `<button>` elements and fixing the styling accordingly. I have also added the `aria-expanded` attribute, to indicate whether the button is expanded or not to screen readers.

Furthermore I fixed a bug where brackets were missing in the `<umb-icon icon="">` attribute meaning that the arrow icon for "Package content" and "Package actions" was missing. Please see the below before and after screendumps.

**Before**
missin-arrow-icon

**After** 
not-missing-arrow-icon

PR #9065, opened 2020-10-04
Fix missing icon shown for package actions group

bjarnef:
### Prerequisites

### Description
In https://github.com/umbraco/Umbraco-CMS/pull/5606 `<umb-icon>` component was adding and the icon in the icon in header of the expansion panels was replaced with `<umb-icon>`.

However for package actions it wasn't rendered correct, which this PR fixes.


bjarnef:
@nathanwoulfe maybe you would like to test this one? I guess we don't need the second part regarding setting the icon classes anyway :)

nathanwoulfe:
@bjarnef looks like @BatJan beat you to it - he fixed this issue as part of #9054

PR #9066, opened 2020-10-04
Use ace xml mode for package actions

bjarnef:
### Prerequisites

### Description
At the moment package actions input in packages section is a simple textarea. However since package actions at the moment is in xml format, is would be great to use the xml mode in ace editor.
https://our.umbraco.com/documentation/Extending/Packages/Package-Actions/

image


nathanwoulfe (approved):
Hey @bjarnef, all looks good to me, definitely an improvement.

The documentation link annoys me a bit - thought maybe it would be better in the Actions description, but that means messing with escaping text and probably isn't worth it for a one-off. If the control group description supported html, it could be useful...

PR #9069, opened 2020-10-04
umb-control-group directive: Fix styling issue

BatJan:
### Prerequisites

- ✔ I have added steps to test this contribution in the description below

### Description
@bjarnef found this issue and we had a chat about whether it could be introduced by a former PR of mine but we figured out it was not but I of course got a bit curious to figure out what was going on and after a little digging it appears to be the use of `.controls-row` class that offsets the padding on the "Title" by 10 pixels. Removing the class appears to be fixing the issue.

Therefore I have simply removed this class and that seems to fix the issue - I have checked the view that appears when you right click on a content node and select "Do something else" and select "Permissions" and then choose a group to setup permissions for in the dropdown. I have checked it in the users section as well. Can't remember the actions to take to see it but it looked fine there as well.

**Before**
control-group-before

**After**
control-group-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/main.less
+++ b/src/Umbraco.Web.UI.Client/src/less/main.less
@@ -276,7 +276,7 @@ label:not([for]) {
 }
 
 .controls-row {
-  padding-bottom: 5px;
+  padding-bottom: 5px; 
   margin-left: 240px;
 }
 
--- a/src/Umbraco.Web.UI.Client/src/less/main.less
+++ b/src/Umbraco.Web.UI.Client/src/less/main.less
@@ -276,7 +276,7 @@ label:not([for]) {
 }
 
 .controls-row {
-  padding-bottom: 5px; 
+  padding-bottom: 5px;
   margin-left: 240px;
 }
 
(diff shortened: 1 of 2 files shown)


bjarnef:
It seems to be related to this instead https://github.com/umbraco/Umbraco-CMS/commit/c81266f5dd9c7181491986a42102a325f7970df7#diff-554f0d5110d05a4f16f86254fd35631dR10 where a `<label>` element was added and since the `<label>` with ancestor with `.controls-row` by default has left/right padding. 🤦‍♂️

nathanwoulfe (approved):
Thanks @BatJan, works perfectly - no more misaligned labels

mattbrailsford:
This is causing regression issues in Vendr. We use the `umb-control-block` directive in a `umb-box` directive using the `.block-form` css class on the umb-box. With these in place, there was a CSS rule for `.block-form .controls-row` which would remove the left margin + top padding on the control row, but without this class on the element any more, this falls back to the `.form-horizontal .controls` css class which has 180px left margin. So what was aligned correctly before, now is not.

Before:
image

After:
image

mattbrailsford:
I think maybe there needs to be a rule that says a class shouldn't be removed from a components dom unless there are no other CSS rules targeting it.

bjarnef:
@mattbrailsford I think this PR might solve this issue https://github.com/umbraco/Umbraco-CMS/pull/9632

mattbrailsford:
@bjarnef yea, this is essentially what I've had to add to my CSS file.

bjarnef:
The PR link to a similar issue for member group https://github.com/umbraco/Umbraco-CMS/pull/9469 but that solution can't be used under datatypes.

/cc @nielslyngsoe

PR #9071, opened 2020-10-05
Update 'working with the code' to remove reference to VS2017

marcemarc:
I'm not sure you need to restate the version here, having got it specified in the list above, but if you do, then a vanilla 2017 won't build the source because of the c#8 stuff introduced recently.


nul800sebastiaan:
Good one, cheers @marcemarc!

PR #9075, opened 2020-10-05
V8/bugfix/9074 current urls segments

### LINKED ISSUE #9074: Culture invariant node with segment variant: "This item is not published"
When a segment variant is created for a node that does not allow culture variants the Info tab shows "This item is not published", even though the node is published. This _only_ affects the backoffice UI, the node is still published and reachable.

not published
segments not published

## Umbraco version

I am seeing this issue on Umbraco version: 8.7 + 8.8

### Steps to reproduce

1. Enable segmentation for a document type + 1 property type
2. Create a segment for a node of this document type using SetValue() on the IContent with some value for the "segment" parameter (e.g. `"tmp_segment"`)
3. Go to the Info tab and observe the message.

### Expected result

The URL for the Default variant and the segment variant should be the URL of the Default variant.
Segment variants do not have their own URLs like culture variants have.

### Actual result

Umbraco shows "This item is not published".

### Cause of bug
So this issue is [caused by the code](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbcontentnodeinfo.directive.js#L324) for the `currentUrls` property in the `<umb-content-node-info>` directive. 
This code is checking the number of variants and if there is more than 1, assumes we are dealing with culture variation. However, now that segments are added this is no longer always the case.

```js
if (scope.node.variants.length > 1) {
    // nodes with variants
    scope.currentUrls = _.filter(scope.node.urls, (url) => (scope.currentVariant.language && scope.currentVariant.language.culture === url.culture));
} else {
    // invariant nodes
    scope.currentUrls = scope.node.urls;
}
```

So we end up in the `scope.node.variants.length > 1` case, which will then lead to an empty array of currentUrls because `currentVariant.language` is always null.

I will submit a PR shortly that will fix this issue by simplifying the `if/else` logic to 1 case that will handle everything.
Basically, we should just filter all URLs for the current language, but if there is not current language show all URLs.
This should work in all cases.
```js
scope.currentUrls = _.filter(scope.node.urls, (url) => (scope.currentVariant.language == null || scope.currentVariant.language.culture === url.culture));
```

PerplexDaniel:
### Prerequisites

### Description
Fixes #9074 
Explanation & steps to reproduce are listed in the issue.


nul800sebastiaan:
Thanks @PerplexDaniel - sorry for the delay on this one! I've never played with Segments so I needed to create some kind of test environment to figure out what your code was doing.

For future reference, I enabled and created segments like so, that was a fun one to figure out! 😁 

```csharp
{
    var contentType = _contentTypeService.Get(1092); // Contact doctype Id of the default starter kit
    contentType.Variations = ContentVariation.Segment;
    var pageTitleProperty = contentType.PropertyTypes.FirstOrDefault(x => x.Alias == "pageTitle");
    if (pageTitleProperty != null)
        pageTitleProperty.Variations = ContentVariation.Segment;
    _contentTypeService.Save(contentType);
    scope.Complete();
}

{
    var contactItem = _contentService.GetById(1118); // Id of the content item for Contact page in the default starter kit
    contactItem.SetValue("pageTitle", "Test Segment", segment: "tmp_segment");
    _contentService.Save(contactItem);
    scope.Complete();
}
```

I am not familiar with the `_.filter` notation either so I had to put a breakpoint in to see what it was doing.. This behaves as expected, if there's only segment variations, it displays all URLs available, if there's also language segments it picks the URLs applicable to the current language.

So all good, merging for 8.12!

PR #9080, opened 2020-10-05
Fix casing of 'URL'

ronaldbarendse:
### Prerequisites

### Description
I've always been annoyed by the fact that 'URL' isn't written in capital letters ([because it's an abbreviation for Universal Resource Locator](https://www.dictionary.com/browse/url)) when managing redirects or using the link picker.

This PR updates all translations to use capitals and also (because I was already using find/replace):
- TinyMCE language files
- XML documentations
- In-code messages (e.g. exceptions, logs, etc.)
- Comments

The same is probably true for e.g. HTTP/HTTPS, API or UDI, but those won't be as visible as URL (I haven't seen those used incorrectly in the translations) and are probably not worth the effort to change...

Testing can by done by checking whether the casing is now correct in the 'Redirect URL Management' dashboard and 'Multi URL Picker' editor, although checking the changes is probably enough to validate for correctness.


src/Umbraco.Core/GuidUdi.cs
```diff
@@ -34,7 +34,7 @@ public GuidUdi(Uri uriValue)
         {
             Guid guid;
             if (Guid.TryParse(uriValue.AbsolutePath.TrimStart('/'), out guid) == false)
-                throw new FormatException("Url \"" + uriValue + "\" is not a guid entity id.");
+                throw new FormatException("URI \"" + uriValue + "\" is not a GUID entity id.");
```
ronaldbarendse, line 37: 'id' is probably also an abbreviation of 'identifier', so this can be updated to:
```suggestion
                throw new FormatException("URI \"" + uriValue + "\" is not a GUID entity ID.");
```

src/Umbraco.Core/GuidUdi.cs
```diff
@@ -48,7 +48,7 @@ public new static GuidUdi Parse(string s)
         {
             var udi = Udi.Parse(s);
             if (udi is GuidUdi == false)
-                throw new FormatException("String \"" + s + "\" is not a guid entity id.");
+                throw new FormatException("String \"" + s + "\" is not a GUID entity id.");
```
ronaldbarendse, line 51:
```suggestion
                throw new FormatException("String \"" + s + "\" is not a GUID entity ID.");
```

src/Umbraco.Web/Media/EmbedProviders/EmbedProviderBase.cs
```diff
@@ -25,7 +25,7 @@ public abstract class EmbedProviderBase : IEmbedProvider
         public virtual string GetEmbedProviderUrl(string url, int maxWidth, int maxHeight)
         {
             if (Uri.IsWellFormedUriString(url, UriKind.RelativeOrAbsolute) == false)
-                throw new ArgumentException("Not a valid Url");
+                throw new ArgumentException("Not a valid URL.", nameof(url));
```
ronaldbarendse, line 28: I've also added the parameter name...

src/Umbraco.Web/PropertyEditors/MultiUrlPickerPropertyEditor.cs
```diff
@@ -13,7 +13,7 @@ namespace Umbraco.Web.PropertyEditors
     [DataEditor(
         Constants.PropertyEditors.Aliases.MultiUrlPicker,
         EditorType.PropertyValue,
-        "Multi Url Picker",
+        "Multi URL Picker",
```
ronaldbarendse, line 16: I don't think changing the data editor name should give any issues, as it's only used for displaying in the back-office, but maybe this can be double checked?

src/Umbraco.Core/Services/IRedirectUrlService.cs
```diff
@@ -10,78 +10,78 @@ namespace Umbraco.Core.Services
...
         /// </summary>
-        /// <param name="url">The Umbraco redirect url route.</param>
-        /// <returns>The most recent redirect urls corresponding to the route.</returns>
+        /// <param name="url">The Umbraco redirect URL route.</param>
+        /// <returns>The most recent redirect URLs corresponding to the route.</returns>
         IRedirectUrl GetMostRecentRedirectUrl(string url);
 
         /// <summary>
-        /// Gets all redirect urls for a content item.
+        /// Gets all redirectURLs for a content item.
```
nul800sebastiaan, line 52:
```suggestion
        /// Gets all redirect URLs for a content item.
```

nul800sebastiaan (approved):

nul800sebastiaan:
Thanks @ronaldbarendse. I don't want to sound ungrateful, but do you have a list of things that annoy you? 

Generally a cleanup like this, while it seems nice, has a big impact on merging and existing open PRs. We rather avoid it cleanup as we go along. I appreciate that this is not always possible and a cleanup task is necessary sometimes. 
I'll merge this one, but let's talk if you have a list of other one of these that you want to fix.

Regarding the TinyMCE files: I can't guarantee that they won't be overwritten the next time we upgrade Tiny, apologies in advance.

ronaldbarendse:
> do you have a list of things that annoy you?

I do keep a list of minor issues I encounter that aren't that important to fix, but after a while indeed start to annoy me 😉 This was one of them, so I'm grateful for merging it so fast! 😄 Like I already indicated, I don't think fixing other abbreviations/casings are worth the effort. It might be good to keep this in mind when reviewing though, especially when it's visible to the end-user/editor.

I purposely created separate commits, so you would be able to cherry pick only the changes to e.g. the translations. I do know that changing a lot of files might cause merge conflicts, especially on open PRs (like my own PR https://github.com/umbraco/Umbraco-CMS/pull/7533 😞), but as it's only changing the casing, will be easy to resolve.

Regarding the TinyMCE files: you will at least see the changes in the diff when the files are overwritten on an update. I haven't seen those texts in the editor though, so it's probably only for the tooltips. Thinking about that, this PR might also improve accessibility if screen readers now correctly pronounce URL?

bergmania:
> Thanks @ronaldbarendse. I don't want to sound ungrateful, but do you have a list of things that annoy you?
> 
> Generally a cleanup like this, while it seems nice, has a big impact on merging and existing open PRs. We rather avoid it cleanup as we go along. I appreciate that this is not always possible and a cleanup task is necessary sometimes.
> I'll merge this one, but let's talk if you have a list of other one of these that you want to fix.

Exactly.. I feel the pain now... 😢 
image

Issue #9078, opened 2020-10-05
Don't overwrite custom branding on upgrade


craigs100:
This is just a plea from the heart. Would it be possible to not overwrite the login screen Logo and background Image during an upgrade?  Thought I'd get this in early as there's a concentration on the Login area at the moment.

It's a PITA to have to work out where the changes go every time when the client asks where their shiny logo and background have gone after an upgrade.

Thanks.

nul800sebastiaan:
You can change the background image in `umbracoSettings.config` (I am assuming you just overwrite `login.jpg` it in the umbraco folder at the moment). 

```
    <!-- You can specify your own background image for the login screen here. This path is relative to the ~/umbraco path. The default location is: /umbraco/assets/img/login.jpg -->
    <loginBackgroundImage>assets/img/login.jpg</loginBackgroundImage>
```

If you put it in the same folder and name it `customlogin.jpg` it shouldn't be overwritten. we have no such config setting for the logo but that would be a good pull request to get in, it will even earn you a t-shirt if you can do it in October! https://umbraco.com/blog/hacktoberfest-2020/

umbrabot:
Hi @craigs100,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

ukdiverboi:
@nul800sebastiaan see PR I have added with screenshots, etc. that implements the functionality for this ticket #9163

PR #9083, opened 2020-10-06
9078: Allow changing the login logo through umbracoSettings

patrickdemooij9:
Issue: https://github.com/umbraco/Umbraco-CMS/issues/9078

Just wondering if it's ok to set the default in the umbracoSettings.config instead of falling back to it if the element doesn't exist. For now, I just decided to stick to the same implementation for the background, but let me know what you would rather want!

Can be tested by setting the loginLogoImage element in the UmbracoConfig


nul800sebastiaan:
@patrickdemooij9 - Thanks! We should definitely set a default if there is nothing defined in the config file, could you add that please? 🙏

patrickdemooij9:
@nul800sebastiaan, I updated it so that it has the default set in the code!

mikecp (approved):

mikecp:
Thanks @patrickdemooij9 for this PR, it adds a nice customization option 👍

PR #9085, opened 2020-10-06
NetCore: Migrate further integration tests

AndyButland:
In this PR I've migrated a few more integration tests:

- AmbiguousEventTests (made this one a unit test)
- ContentServiceEventTests
- MacroServiceTests

Have also made some minor amends to the existing ones (adhering naming convention for private variables and making use of the existing builders for creating user models).


src/Umbraco.Tests.Integration/Services/ContentServiceEventTests.cs
```diff
@@ -7,24 +7,29 @@
...
         PublishedRepositoryEvents = true,
         WithApplication = true,
         Logger = UmbracoTestOptions.Logger.Console)]
-    public class ContentServiceEventTests : TestWithSomeContentBase
+    public class ContentServiceEventTests : UmbracoIntegrationTest
     {
+        private IContentTypeService _contentTypeService => GetRequiredService<IContentTypeService>();
+        private IContentService _contentService => GetRequiredService<IContentService>();
+        private ILocalizationService _localizationService => GetRequiredService<ILocalizationService>();
+        private IFileService _fileService => GetRequiredService<IFileService>();
```
bergmania, line 26: All of these do not follow the convension for private properties. Should be Title case

src/Umbraco.Tests.Integration/Services/ContentServiceEventTests.cs
```diff
@@ -7,24 +7,29 @@
...
         private GlobalSettings _globalSettings;
 
-        public override void SetUp()
+        public override void Setup()
         {
-            base.SetUp();
+            base.Setup();
             ContentRepositoryBase.ThrowOnWarning = true;
             _globalSettings = new GlobalSettings();
         }
```
bergmania, line 35: IMO it is nicer to use [Setup] and call the mehtod something else to not collide with name from the base class. Then we donøt need to remember to call `base.Setup();`

src/Umbraco.Tests.Integration/Services/MacroServiceTests.cs
```diff
@@ -6,27 +6,33 @@
...
+namespace Umbraco.Tests.Integration.Services
 {
     [TestFixture]
     [Apartment(ApartmentState.STA)]
     [UmbracoTest(Database = UmbracoTestOptions.Database.NewSchemaPerTest)]
-    public class MacroServiceTests : TestWithSomeContentBase
+    public class MacroServiceTests : UmbracoIntegrationTest
     {
-        public override void CreateTestData()
+        private MacroService _macroService => (MacroService)GetRequiredService<IMacroService>();
```
bergmania, line 25: Not following convension for properties

src/Umbraco.Tests.Integration/Services/MediaServiceTests.cs
```diff
@@ -19,16 +19,16 @@ namespace Umbraco.Tests.Integration.Services
     [UmbracoTest(Database = UmbracoTestOptions.Database.NewSchemaPerTest, PublishedRepositoryEvents = true)]
     public class MediaServiceTests : UmbracoIntegrationTest
     {
-        private IMediaService MediaService => GetRequiredService<IMediaService>();
-        private IMediaTypeService MediaTypeService => GetRequiredService<IMediaTypeService>();
+        private IMediaService _mediaService => GetRequiredService<IMediaService>();
+        private IMediaTypeService _mediaTypeService => GetRequiredService<IMediaTypeService>();
```
bergmania, line 23: Not following convensions

src/Umbraco.Tests.Integration/Services/PublicAccessServiceTests.cs
```diff
@@ -15,20 +15,20 @@ namespace Umbraco.Tests.Integration.Services
...
     public class PublicAccessServiceTests : UmbracoIntegrationTest
     {
-        private IContentService ContentService => GetRequiredService<IContentService>();
-        private IContentTypeService ContentTypeService => GetRequiredService<IContentTypeService>();
-        private IFileService FileService => GetRequiredService<IFileService>();
-        private IPublicAccessService PublicAccessService => GetRequiredService<IPublicAccessService>();
+        private IContentService _contentService => GetRequiredService<IContentService>();
+        private IContentTypeService _contentTypeService => GetRequiredService<IContentTypeService>();
+        private IFileService _fileService => GetRequiredService<IFileService>();
+        private IPublicAccessService _publicAccessService => GetRequiredService<IPublicAccessService>();
```
bergmania, line 21: Not following convensions

bergmania (changes requested):
Hi @AndyButland.. Nice work. 

I disagree with your changes of private properties.  Remember it is still properties, even they use the syntactical sugar "=>".  If they were fields ("=") then I agree in using _camelCase.

src/Umbraco.Tests.Integration/Services/MacroServiceTests.cs
```diff
@@ -6,27 +6,33 @@
...
+    public class MacroServiceTests : UmbracoIntegrationTest
     {
-        public override void CreateTestData()
+        private MacroService MacroService => (MacroService)GetRequiredService<IMacroService>();
+
+        [SetUp]
+        public override void Setup()
         {
-            base.CreateTestData();
+            base.Setup();
```
bergmania, line 30: Still thinks this should only use the [SetUp] attribute and not override the `base.Setup`

src/Umbraco.Tests.Integration/Services/ContentServiceEventTests.cs
AndyButland: Agreed, have updated.

src/Umbraco.Tests.Integration/Services/ContentServiceEventTests.cs
AndyButland: Have updated to this.

src/Umbraco.Tests.UnitTests/Umbraco.Web.BackOffice/Filters/FilterAllowedOutgoingContentAttributeTests.cs
```diff
@@ -78,10 +77,7 @@ public void GetValueFromResponse_Returns_Null_Not_Found_Property()
         [Test]
         public void Filter_On_Start_Node()
         {
-            var userMock = MockedUser.GetUserMock();
-            userMock.Setup(u => u.Id).Returns(9);
-            userMock.Setup(u => u.StartContentIds).Returns(new[] { 5 });
-            var user = userMock.Object;
+            var user = CreateUser(id: 9, startContentId: 5);
```
bergmania, line 80:
```suggestion
            var user = new UserBuilder()
                .WithId(9)
                .WithStartContentIds(new[] { 5 })
                .Build();
```

IMO, this is a good example where it is an anti-pattern not using the builder directly. Because we build a model that is more complex than needed for the specific tests. 
That is why I like the builder pattern because you can have a minimal model, with just the changes required for the test.

None of the tests in this class seems to need the user group to be set?

src/Umbraco.Infrastructure/Runtime/CoreInitialComposer.cs
```diff
@@ -204,13 +203,6 @@ public override void Compose(Composition composition)
             // Config manipulator
             composition.RegisterUnique<IConfigManipulator, JsonConfigManipulator>();
 
-
-            // register the http context and umbraco context accessors
-            // we *should* use the HttpContextUmbracoContextAccessor, however there are cases when
-            // we have no http context, eg when booting Umbraco or in background threads, so instead
-            // let's use an hybrid accessor that can fall back to a ThreadStatic context.
-            composition.RegisterUnique<IUmbracoContextAccessor, HybridUmbracoContextAccessor>();
-
```
bergmania, line 213: Not sure I understand why this is removed?

AndyButland: It was registered twice.

bergmania: Then it makes perfect sense 💪  😄

src/Umbraco.Tests.UnitTests/Umbraco.Web.BackOffice/Filters/FilterAllowedOutgoingContentAttributeTests.cs
AndyButland: They don't no - have removed.

src/Umbraco.Tests.Integration/Services/MacroServiceTests.cs
AndyButland: Agreed, have changed this too.

bergmania (approved):

bergmania:
Just waits for the buildserver to ensure all tests passes, before I merge

PR #9091, opened 2020-10-06
Adding a cache buster to the scoped stylesheet in Blocklist editor.

### LINKED ISSUE #9079: Custom stylesheet in Block Editor should have a cache buster
I really like the custom view/stylesheet you can select for editor appearance in block editor, but if I make changes to the stylesheet it doesn't show in backoffice.

## Umbraco version

I am seeing this issue on Umbraco version: Umbraco 8.7 and later

Reproduction
------------

### Bug summary

The problem is that the file is cached by the browser, and isn't reloaded by any mechanisms in Umbraco, like the client dependency version that the rest of the backoffice uses. Since it is loaded asyncronously it will also not reload with CTRL-F5 or similar. The only way is to delete browser cache, or rename the file.

### Steps to reproduce

Select a custom stylesheet file in the editor appearance dialog. Load the view in backoffice. Then go back and edit the stylesheet. The edit will not show up in backoffice unless you clear browser cache.

### Expected result

There should be a good way to invalidate this file, so that it is reloaded. With multiple editors it is not a good solution to ask them to clear the browser cache. Renaming the file with each edit is possible, but a rather impractical solution.

### Proposed solution

Probably the best way to deal with this is to use the client dependency version that controls the rest of the backoffice resources. That would mean that the stylesheet should be included as something like:

```css
@import 'stylesheet.css?v=#######';
```

ArnsteinBerg:
Adding a cache buster to the scoped stylesheet in Blocklist editor.

### Prerequisites

Fixes #9079

### Description

This adds the standard cache buster used in the assets service, to the scoped stylesheet in the Blocklist editor blocks. This stylesheet is loaded separately from the rest of Umbraco's assets and will not be refreshed if the site has long cache timeouts for static resources.

Steps to test:
1. Make sure the site has long cache timeout for static resources:
`    <staticContent>
      <clientCache cacheControlMode="UseMaxAge" cacheControlMaxAge="365.00:00:00" />
    </staticContent>
`
1. Create a Block editor with a block that has a custom view and stylesheet configured.
1. Open a document in the backoffice with this editor, and watch the custom view and stylesheet load.
1. Edit both the stylesheet and the view
1. Reload the backoffice, and notice that the custom stylesheet is not updated since it is already cached by the browser.
1. Change the client dependency version number.
1. Reload the backoffice again, and the edits should now be visible.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbblocklistblock.component.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbblocklistblock.component.js
@@ -58,7 +58,7 @@
                     <style>
                     @import "{{$scope.stylesheet}}"
                     </style>
-                    <div ng-include="'${$scope.view}'"></div>
+                    <div ng-include="view"></div>
                 `;
                 $compile(shadowRoot)($scope);
             }
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbblocklistblock.component.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbblocklistblock.component.js
@@ -29,7 +29,7 @@
         }
     );
 
-    function BlockListBlockController($scope, $compile, $element, umbRequestHelper) {
+    function BlockListBlockController($scope, $compile, $element, assetsService, umbRequestHelper) {
         var model = this;
 
         model.$onInit = function () {
@@ -49,26 +49,20 @@
             $scope.index = model.index;
             $scope.parentForm = model.parentForm;
             $scope.valFormManager = model.valFormManager;
+            $scope.view = assetsService.appendCacheBuster(model.view);
 
             if (model.stylesheet) {
-                var stylesheetUrl = model.stylesheet;
-                if (Umbraco.Sys.ServerVariables.application) {
-                    var rnd = Umbraco.Sys.ServerVariables.application.cacheBuster;
-                    var _op = (stylesheetUrl.indexOf("?") > 0) ? "&" : "?";
-                    stylesheetUrl = stylesheetUrl + _op + "umb__rnd=" + rnd;
-                }
-
                 var shadowRoot = $element[0].attachShadow({ mode: 'open' });
                 shadowRoot.innerHTML = `
                     <style>
-                    @import "${stylesheetUrl}"
+                    @import "${assetsService.appendCacheBuster(model.stylesheet)}"
                     </style>
-                    <div ng-include="'${model.view}'"></div>
+                    <div ng-include="'${$scope.view}'"></div>
                 `;
                 $compile(shadowRoot)($scope);
             }
             else {
-                $element.append($compile('<div ng-include="model.view"></div>')($scope));
+                $element.append($compile('<div ng-include="view"></div>')($scope));
             }
         };
 
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbblocklistblock.component.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbblocklistblock.component.js
@@ -52,6 +52,7 @@
             $scope.view = assetsService.appendCacheBuster(model.view);
 
             if (model.stylesheet) {
+                $scope.stylesheet = assetsService.appendCacheBuster(model.stylesheet);
                 var shadowRoot = $element[0].attachShadow({ mode: 'open' });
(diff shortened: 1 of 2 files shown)


poornimanayar:
Hello @ArnsteinBerg ,

Many thanks for your work. I am going to give the PR a test and see how it all works. I will be in touch if I need any more information on this.

Regards,
Poornima

poornimanayar:
Hi @ArnsteinBerg ,

I have tested this and I want to confirm that what I have done is right. I have got an extended cache timeout in web.config, added a block list with a custom view and stylesheet. I went in and edited the stylesheet from settings section, updated the client dependency config by 1, restarted, and saw your fix working! Is that correct?

Regards,
Poornima

Poornima

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbblocklistblock.component.js
```diff
@@ -51,10 +51,17 @@
             $scope.valFormManager = model.valFormManager;
 
             if (model.stylesheet) {
+                var stylesheetUrl = model.stylesheet;
+                if (Umbraco.Sys.ServerVariables.application) {
+                    var rnd = Umbraco.Sys.ServerVariables.application.cacheBuster;
+                    var _op = (stylesheetUrl.indexOf("?") > 0) ? "&" : "?";
+                    stylesheetUrl = stylesheetUrl + _op + "umb__rnd=" + rnd;
+                }
```
ronaldbarendse, line 59: Instead of duplicating code, the following private function could be made public as e.g. `assetsService.appendCacheBuster(url)`:

https://github.com/umbraco/Umbraco-CMS/blob/cb47e0f1f6c4832ed4f1c90c4541c74adabea923/src/Umbraco.Web.UI.Client/src/common/services/assets.service.js#L48-L58

ArnsteinBerg: Yes, I thought about this too, but I didn't want to touch more code than necessary to resolve the issue. I see now that this issue also apply to the custom view of the blocklistblock, so I think I will propose a new solution using the function from assetsService.

ArnsteinBerg:
@poornimanayar You have tested correctly. However I see that this issue also involves the custom view for blocklistblock, so I think I will propose a new solution for both of these issues, and you might want to wait for that.

ArnsteinBerg:
@poornimanayar Now I have added the fix for both custom stylesheet and custom view. I have also made use of the cache buster function from assetsService, as suggested.

The test to verify the fix is the same as before, but now you should also try to make a change to the custom view, and see that both this and the stylesheet is reloaded when you increase the client dependency version.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umbblocklistblock.component.js
```diff
@@ -49,19 +49,20 @@
...
 
             if (model.stylesheet) {
                 var shadowRoot = $element[0].attachShadow({ mode: 'open' });
                 shadowRoot.innerHTML = `
                     <style>
-                    @import "${model.stylesheet}"
+                    @import "${assetsService.appendCacheBuster(model.stylesheet)}"
                     </style>
-                    <div ng-include="'${model.view}'"></div>
+                    <div ng-include="'${$scope.view}'"></div>
```
ronaldbarendse, line 60: This doesn't require escaping the template literal and should be the same as below:
```suggestion
                    <div ng-include="view"></div>
```

ronaldbarendse: If these changes are made, the [template literals](https://caniuse.com/template-literals) (backticks) aren't needed anymore.

ronaldbarendse (changes requested):

poornimanayar:
Hi @ArnsteinBerg ,

I have tested this again but I am struggling to reproduce the fix for custom stylesheet. Is it possible you can record your screen while you test and post here? I have tested custom view and it looks good. It would also be good to post the custom stylesheet and views here.

Poornima

nul800sebastiaan:
Hi again @ArnsteinBerg! We're wrapping up the Hacktoberfest rewards list and haven't heard from you yet, make sure to fill in the form at https://docs.google.com/forms/d/e/1FAIpQLSdvfuqHFn7ib_xxMbVZ3nuyUgUTUCGPwmw0y83kiY4lP3OzKQ/viewform or if you want us to plant a tree on your behalf then you can reply as well and we'll take care of it. Thanks! 👍

Issue #8950, opened 2020-09-23
Resizing content tree messes up Users, Packages & Forms section


JeffreyPerplex:
1. Open Umbraco. Resize the contenttree panel (drag it bigger or smaller)

![afbeelding](https://user-images.githubusercontent.com/12457148/93978444-4e522600-fd7c-11ea-9188-9a9ab3afc179.png)

2. Go to another section like Users / Package / Forms... These sections don't have the tree visible, but you see that there is a grey panel on the left side (where the content tree supposed to be)... 

![afbeelding](https://user-images.githubusercontent.com/12457148/93978476-59a55180-fd7c-11ea-8748-647dd31383b5.png)

3. As soon as you hit F5 or resize your window everything is fine, but it's a bit strange

And if you want a demo in a gif... That is of course possible as well:

![trucdelatruc2](https://user-images.githubusercontent.com/12457148/93980252-ae49cc00-fd7e-11ea-8a8d-2d756a001ea8.gif)

JeffreyPerplex:
Umbraco 8.7 and higher (and probably below as well :))

nul800sebastiaan:
Ha! Never seen that before, interesting! Might be a good one for Hacktoberfest.

umbrabot:
Hi @JeffreyPerplex,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
It seems the left position isn't reset, when switching to a different section, where it should use full width like packages and users sections.

![image](https://user-images.githubusercontent.com/2919859/95226658-7e2e0e80-07fd-11eb-9250-dcdece0b4743.png)

PR #9092, opened 2020-10-06
Ensure content wrapper is positioned to left in section without tree

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8950

### Description
This PR a probably a temporary workaround, but it would be great to investigate further how to reset the inline left position on `contentwrapper`, when using the vertical size handle to change size of navigation panel and afterwards switch to another section without trees like packages, users, forms, etc.

It does seems there is still a tree in background but this is hidden here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/application/grid.less#L180

For now the workaround in this PR seems to do the trick, where the left position with `!important` gets a higher priority than the inline left position.

DMjlSMFH3U

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/application/grid.less
+++ b/src/Umbraco.Web.UI.Client/src/less/application/grid.less
@@ -171,13 +171,13 @@ body.umb-drawer-is-visible #mainwrapper{
 }
 
 @media (min-width: 1101px) {
-  #contentwrapper, #speechbubble {left: 360px;}
-  .emptySection #contentwrapper {left:0px;}
+  #contentwrapper, #speechbubble { left: 360px; } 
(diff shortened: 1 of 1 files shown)


nathanwoulfe (approved):
As much as !important and id-bound CSS makes my heart hurt, this does fix the issue, at least until we fix it properly.

I think there's something fishy happening in the `appState.globalState.changed` event, where `showNavigation` is always sent as true, then sometimes again as false...

bjarnef:
@nathanwoulfe yeah, not pretty, but it works for now 😁

PR #9094, opened 2020-10-06
Combine background image and background color in block card

bjarnef:
### Prerequisites

### Description
This PR allow to combine background image (thumbnail) and background color in block list editor using `linear-gradient` as in the example here: https://stackoverflow.com/a/36679903

https://caniuse.com/mdn-css_types_image_gradient_linear-gradient

iLIYi38Dei

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/blockcard/umb-block-card.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/blockcard/umb-block-card.html
@@ -1,8 +1,8 @@
-
+<div>
     <div class="__showcase"
-        ng-class="{'--error':vm.elementTypeModel === null}"
-        ng-style="{'background-color':vm.blockConfigModel.backgroundColor, 'background-image': vm.styleBackgroundImage}">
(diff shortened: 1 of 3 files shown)


bjarnef:
@nielslyngsoe what do you think of this approach?

nielslyngsoe:
Hi @bjarnef if I got this one right, you basically moved the background-color on top of the image, right?
So that will accomplish a sort of blend-feature?
I think this will be miscommunication, as I imagine some users will be using an SVG as a thumbnail or a transparent image and thereby trust the background-color to be in the back. Anh thoughts on that? :-)

bjarnef:
@nielslyngsoe yes, or actually not an additional layer, but same layer. I think it could work with SVG thumnail as well, but we might need to have another look at this scenario with the current feature and with these changes.

nielslyngsoe:
So the question is how much do we want to be able to do from the BackOffice UI, cause the look you created here would be possible with just tinting/blending the thumbnail. I could see that we end up having many different requests to achieve different looks but I'm not sure we want to support these features. I think it's better we keep it simple and then the ones who need these looks should apply this look to their thumbnails, if you have a lot this can be handled with a Photoshop script or similar. :-) at least that's the right way for now, but please enlighten me if you have specific cases where this is useful — always creat to hear about the real-life stories :-)

nielslyngsoe:
I will close this for now :-) towards the effort of keeping things simple :-)

PR #9098, opened 2020-10-07
remove unnecessary classnames/markup

nathanwoulfe:
### Prerequisites

### Description

Noticed some odd styling in the umb-confirm directive - buttons had extra margin and border (see pics). This change removes the unnecessary classes and an html element, to kill off the additional margins and border. The y-axis margins collapse so the change won't be noticeable, but pulls the buttons back to the left a little. `umb-control-group control-group` classes weren't doing anything as the `umb-button-toolbar` class controlled button positioning.

I've checked all the instances of umb-confirm, looks fine in all cases.

Examples
Delete content - before
delete-before

Delete content - after
delete-after

Empty recycle bin - before
recycle-before

Empty recycle bin - after
recycle-after


mikecp (approved):

mikecp:
Looks all good @nathanwoulfe :-) !

PR #9100, opened 2020-10-07
Adds support for the super old password format so we can handle upgrades

Shazwazza:
For [AB#6964](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6964) and [AB#6965](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/6965)

* Adds support for validating the very old password format
* Fixes an issue i had with the gulp build so that it works on linux (package update) 
* Passwords will now all roll forward to the latest hashing algorithm
* Removes some tests that are testing code that will be deleted

## Testing

In your v8 database, find the row of your user and copy the `passwordConfig` value which should look like this: `{"hashAlgorithm":"HMACSHA256"}` and replace your netcore database user value with that. Do the same for the `userPassword` value. Then restart your netcore site and try to log into the back office with the password you were using in your v8 site and it should work. Once you are logged in go check that your password and hashing algorithm for your user row has been updated to the latest algorithm which is `PBKDF2.ASPNETCORE.V3`


nikolajlauridsen:
Tested it out and it all works 👍

PR #9102, opened 2020-10-07
NetCore: make use of builder classes for model creation in migrated tests

AndyButland:
Relates to issue #7898.

This PR is addressing the issue where we currently have builders for most model classes and have been using them in the refactored tests, in the last batch of integration and unit test migration, we've also brought over various classes like `MockContentTypes`, `MockContent` etc. that are doing a similar job - constructing model instances for tests - but in a different way (vis constructors and direct property setting of the model classes).

So here I've moved these classes that had been moved to `Umbraco.Tests.Common` back into the legacy `Umbraco.Tests` projcet, so they are no longer used by the migrated tests, but can still be used by the non-migrated ones.  The idea being that they will stay there, and eventually be retired along with any tests we aren't going to migrate.  And any further migrated tests should be amended to no longer use them.

To keep the easy use of a static method to create a particular structure for testing, I've recreated the ones we need for the current tests as static methods available off of the builders, and then changed all the migrated tests that used the `Mock*` classes to use these.  These static methods internally use the builders to create the model classes.

In doing this I've found a few subtle differences leading to failing tests, and have made amends to resolve them.


AndyButland:
4 failing tests - all failing with `The factory has not been configured with a proper connection string`.  Not sure what's happening here - doesn't seem to be related to my changes, and I don't see these fail locally.

src/Umbraco.Tests.Common/Builders/MediaBuilder.cs
```diff
@@ -33,36 +35,53 @@ public class MediaBuilder
...
             var level = _level ?? 1;
             var path = _path ?? $"-1,{id}";
             var sortOrder = _sortOrder ?? 0;
             var trashed = _trashed ?? false;
 
-            if (_mediaTypeBuilder == null)
+            if (_mediaTypeBuilder is null && _mediaType is null)
             {
-                throw new InvalidOperationException("A member cannot be constructed without providing a member type. Use AddMediaType().");
+                throw new InvalidOperationException("A media item cannot be constructed without providing a member type. Use AddMediaType() or WithMediaType().");
```
nikolajlauridsen, line 79:
```suggestion
                throw new InvalidOperationException("A media item cannot be constructed without providing a media type. Use AddMediaType() or WithMediaType().");
```

src/Umbraco.Tests.Common/Builders/ContentBuilder.cs
```diff
@@ -48,53 +54,119 @@ public ContentTypeBuilder AddContentType()
...
             var trashed = _trashed ?? false;
             var culture = _cultureInfo?.Name ?? null;
+            var propertyValues = _propertyValues ?? null;
+            var propertyValuesCulture = _propertyValuesCulture ?? null;
+            var propertyValuesSegment = _propertyValuesSegment ?? null;
 
             if (_contentTypeBuilder is null && _contentType is null)
             {
-                throw new InvalidOperationException("A member cannot be constructed without providing a member type. Use AddContentType() or WithContentType().");
+                throw new InvalidOperationException("A content item cannot be constructed without providing a member type. Use AddContentType() or WithContentType().");
```
nikolajlauridsen, line 124:
```suggestion
                throw new InvalidOperationException("A content item cannot be constructed without providing a content type. Use AddContentType() or WithContentType().");
```

src/Umbraco.Tests.Common/Builders/ContentBuilder.cs
```diff
@@ -48,53 +54,119 @@ public ContentTypeBuilder AddContentType()
...
             {
-                var propertyData = _propertyDataBuilder.Build();
-                foreach (var kvp in propertyData)
+                if (_propertyDataBuilder != null)
+                {
+                    var propertyData = _propertyDataBuilder.Build();
+                    foreach (var kvp in propertyData)
+                    {
+                        content.SetValue(kvp.Key, kvp.Value);
+                    }
```
nikolajlauridsen, line 165:
```suggestion
                    foreach (var keyValuePair in propertyData)
                    {
                        content.SetValue(keyValuePair.Key, keyValuePair.Value);
                    }
```

kvp can be a bit confusing at first requiring you to do a mental double take, keyValuePair is more clear :)

nikolajlauridsen:
I cloned down you branch and ran the tests locally. When I run just the ```ContentServiceEventTests``` they pass, but when I run all the integration tests the ```ContentServiceEventTests``` unfortunately fail locally as well, we need to figure out why this is. 

My immediate guess would be something with ```InstallTestLocalDb``` in the ```UmbracoIntegrationTests``` base class but I'm not sure as of yet.

nikolajlauridsen:
Turns out it was because the ContentService raised an event with ContentTypeService, IPublishedSnapShotService which hasn't been properly migrated for nucache yet, I fixed it by clearing the scoped events before running the test 😄 

Other than that I just did some minor tweaks, I fixed a single typo, removed a few imports, and removed Entities and Objects from the Tests.Common csproj file, since the folders no longer exist.

Great work 👍 Just need to wait for the CI now 😄

AndyButland:
Nice - thanks for finding and fixing that, think I may have been looking for a while!

PR #9103, opened 2020-10-07
Hide property actions for culture invariant properties

### LINKED ISSUE #8845: Multi media picker data can be deleted in secondary translation, even if varying by culture disabled
I've just noticed if I have a multiple media picker where the images are picked in the default language, then I go to my secondary translation page, where I can't add any images because that prop editor is not varied by culture.
Note the greyed out image below
image
I can still hover over the three dots and click "Remove all items"
image
which does what it says! Removes all items!

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.4

Reproduction
------------

### Specifics

Using chrome, umbraco version 8.6.4

### Steps to reproduce

Put a multiple media picker property editor on a document type with multiple translations, which is not variable by culture.
Add an image to the main translation page. Go to a secondary translation page, hover over the three dots on this property editor, and remove the images. 
I don't think this should happen?

### Expected result

I expect that a property editor that is not variable by culture *cannot* be changed if you're not in the main translation page for that document type.

### Actual result

You can remove all images from multiple media picker in the

kjac:
### Prerequisites

If there's an existing issue for this PR then this fixes #8845

### Description

Property actions are usually used to edit the content of a property in one way or another... this is troublesome if said properties are culture invariant; you end up being able to edit the property in other languages than the default language:

propery-actions-inherited-properties-before

This is not limited to the Media Picker only, it applies to any property editor with property actions (i.e. Nested Content, Block Editor etc.).

This PR removes the property actions for culture invariant properties outside of the default language:

propery-actions-inherited-properties-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/property/umb-property.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/property/umb-property.html
@@ -22,7 +22,7 @@
 
                     </label>
 
-                    <umb-property-actions actions="vm.propertyActions"></umb-property-actions>
+                    <umb-property-actions ng-if="!vm.showInherit" actions="vm.propertyActions"></umb-property-actions>
 
                     <small class="control-description" ng-if="vm.property.description" ng-bind-html="vm.property.description | preserveNewLineInHtml"></small>
                 </div>


nathanwoulfe (approved):
Thanks @kjac, works perfectly. I'd expect nothing less.

PR #9105, opened 2020-10-07
Add confirmation dialog to media picker "remove all entries" action

kjac:
### Prerequisites

### Description

If you attempt to remove all items of a Nested Content or Block Editor property (by use of the property actions), you're asked to confirm this. However, this is not the case when you do the same for a Media Picker:

propery-actions-media-picker-before

This PR adds the same confirmation dialog to Media Picker:

propery-actions-media-picker-after


nathanwoulfe (approved):
That's a winner, thanks @kjac

PR #9117, opened 2020-10-08
Netcore: migrate relation tests + Introduced new bulksave and fixed issue in save

bergmania:
### Details
- Updated nuget packages
- Found an issue in `RelationRepository` when tests was moved to LocalDB instead of SqlCE. 
  - `RelationRepository.Save(IEnumerable<IRelation> relations)` did not populate ids back on relations when using LocalDB/SqlServer
- Migrated `RelationRepositoryTests` and `RelationServiceTests`
  - Minimized the amount of data created in some of the tests. These tests took ~25 sec each on the build server before
  - `RelationTypeRepository.PerformGetAll` did not support running tests with no cache
- Found another issue in other PR https://github.com/umbraco/Umbraco-CMS/pull/9136 about save did create duplicate keys..
  - Fixed by introducing `IRelationService.GetByParentChildAndType`
  - Tests migrated in https://github.com/umbraco/Umbraco-CMS/pull/9136 proffs this solved the issue

### Tests
1. Create doc type with a media data type
1. Create a content of that doc type and set media
1. Check the relation is created under Settings > Relations > Relate Media > Relations


src/Umbraco.Core/Models/IRelation.cs
```diff
@@ -4,37 +4,81 @@
...
+        new string Comment { get; set; }
+
+        /// <summary>
+        /// Gets the Id of the <see cref="RelationType"/> that this Relation is based on.
+        /// </summary>
+        [IgnoreDataMember]
+        new int RelationTypeId { get; }
+    }
+
+    public interface IRelationReadOnly : IEntity
```
Shazwazza, line 52: What I was proposing was: 

> a new ReadOnlyRelation (or something) that is a simplified/slim version of what is needed to create/update a relation which doesn't need to implement all of the IEntity/IRememberBeingDirty, etc... since it doesn't need anything and all of it's properties are readonly

This shouldn't inherit from IEntity since it doesn't need to be and it's underlying properties are mutable (and also contain Id). It should just contain the properties needed to persist a relation without inheriting from anything. Then IRelation  should not inherit from this, the `new` hiding properties is pretty ugly IMO.

Shazwazza (changes requested):
Test are passing which is good but left a comment about the models.

src/Umbraco.Core/Models/IRelation.cs
bergmania: Please re-review with latest changes

src/Umbraco.Core/Services/IRelationService.cs
```diff
@@ -341,5 +341,15 @@ public interface IRelationService : IService
...
         void DeleteRelationsOfType(IRelationType relationType);
+
+        /// <summary>
+        /// Gets a relation by the unique combination of parentId, childId and relationType.
+        /// </summary>
+        /// <param name="parentId">The id of the parent item.</param>
+        /// <param name="childId">The id of the child item.</param>
+        /// <param name="relationType">The RelationType.</param>
+        /// <returns>The relation or null</returns>
+        IRelation GetByParentChildAndType(int parentId, int childId, IRelationType relationType);
```
Shazwazza, line 352: Should be named GetByParentAndChild to be consistent with the `GetByParentOrChildId` method

src/Umbraco.Infrastructure/Compose/RelateOnTrashComponent.cs
```diff
@@ -68,47 +76,55 @@ private void MediaService_Moved(IMediaService sender, MoveEventArgs<IMedia> e)
...
+                                item.Entity.Id, originalParentId));
+                    }
                 }
+
+                scope.Complete();
             }
         }
 
-        private void MediaService_Trashed(IMediaService sender, MoveEventArgs<IMedia> e)
+        public void MediaService_Trashed(IMediaService sender, MoveEventArgs<IMedia> e)
```
Shazwazza, line 127: why public?

src/Umbraco.Tests.Common/Builders/MediaBuilder.cs
```diff
@@ -221,5 +222,30 @@ string IWithPathBuilder.Path
...
+
+        public static IMedia CreateSimpleMedia(IMediaType contentType, string name, int parentId)
+        {
+            var media = new MediaBuilder()
+                .WithMediaType(contentType)
+                .WithName(name)
+                .WithParentId(parentId)
+                .WithCreatorId(0)
+                .Build();
+                ;
```
Shazwazza, line 234: extra semi colon

Shazwazza (approved):
Looks good, left a couple comments about super minor things but i'll fix those up after i merge. Very glad to see these tests cleaned up, it's been in need of this for many years.

src/Umbraco.Infrastructure/Compose/RelateOnTrashComponent.cs
bergmania: Just revert, my bad.. This was justed tempory to test I found have the same error before I moved the tests

src/Umbraco.Tests.Common/Builders/MediaBuilder.cs
bergmania: 🤦

PR #9123, opened 2020-10-08
Remove .woff font

BatJan:
### Prerequisites

### Description
With this PR I have removed the .woff font and the references to it - It's no longer needed since we only support the so-called "Evergreen" browsers, which all have support for the better compressed .woff2 format according to https://caniuse.com/?search=woff2 😃


emmaburstow:
Jan!!! Thanks for the work. Speak soon,

Em

nathanwoulfe (approved):
Makes sense to me - browsers won't be downloading the woff files when woff2 exist, so it's just deadweight in the codebase. Another winner for @BatJan

Issue #8673, opened 2020-08-20
Open preview in split view


skttl:
When opening the preview, you get a SignalR connection, that makes it update automatically every time you save your content. This is awesome, and IMO an unsung hero feature.
![UnaMZTQ8FR](https://user-images.githubusercontent.com/3726467/90763486-11ab8e80-e2e7-11ea-8872-24546111b44c.gif)

But, you need to manager browser windows and set them up next to each other in order to get real value from it.

So I would suggest that we utilize the split view feature and open the preview in there.

In addition to that, we could make the split view zoom the preview, so it matches the breakpoint selected (so if you ie. choose the desktop breakpoint, but doesn't have screen real estate to show it, the preview will be zoomed out to fit. The same way browser dev tools can do).

nul800sebastiaan:
Ha! Thanks, I learn something new every day, we do refresh the page when something new is saved, very cool.

Personally I am a little concerned that it might not look too great in split view as the viewport might get a little squished for the editing side. 
Is there a clever way to zoom out using just javascript? That does sound cool. 

Additionally, I would probably make an option to still be able to open the preview in a new tab for people who want to use more space but that can be part of this too.

I'll need to discuss this at HQ but it's pretty exciting idea to me. Is there any chance you would be able to hack together a quick prototype? 
I can guess one of the objections might be "that's not what split view is for!" so we'll need to see how to handle that as well.

skttl:
I'll see what I can do. Is there any kind of documentation on hooking up
the split view?

tor. d. 20. aug. 2020 15.30 skrev Sebastiaan Janssen <
notifications@github.com>:

nul800sebastiaan:
Hehe, now that would be the biggest problem. It is not meant for anything other than side by side editing of the same document, so you're probably going to have to hack the planet for now. 🙈

nul800sebastiaan:
Just a thought, it could even be an App_Plugin if that's possible, no need to try and wrangle the core code for a simple POC.

nul800sebastiaan:
Hey there @skttl - we had a chat about this at HQ and I've posted the notes of that conversation on the related issue. I'll close this one and we can continue adding ideas to issue #8303

Issue #8303, opened 2020-06-17
Preview as a Content App


pgregorynz:
Preview is currently tided to a button and has been for a very long time.  Now that content apps have been added to Umbraco it should be possible to make the preview experience load in the back office in a content app.

How can you help?
-------------------------------

I have the dynamic loading of previews in the back office of Umbraco for Bento Editor (in dev but prototyped and working) and the same process could be used to load an entire view of the front end in a content app in a dynamically generated iframe.

JeffreyPerplex:
I would love to see it as a content app-like functionality as well. How we think about this now is that it's just some kind of content app that behaves like a checkbox. You can enable it by clicking on it, and then it appears in the backoffice everywhere on the right side for example.

And once you click it again it disappears. 

This is not something you could achieve with a content app right now, but it does sound cool, doesn't it :)

How our current functionality looks like is this (implemented by the fantastic @PerplexDaniel in Perplex.ContentBlocks):

![afbeelding](https://user-images.githubusercontent.com/12457148/84894230-c08ddd00-b0a0-11ea-804d-c875d91c35bb.png)

- It's rendering the preview in a iframe for this specific pageId & culture
- You can see it in desktop & mobile mode
- If you hit save (or the big button "Save & Refresh") it reloads

And a pretty cool feature is that the page scrolls with the editor according to the content block that is shown. So if you have a really long page and you scroll to the 4th contentblock the preview will scroll there as well. This of course requires some magic in the preview rendering and I don't know whether this could be generalized so that it works for every datatype...

But it would be cool :),

Jeffrey Schoemaker

nul800sebastiaan:
Just linking this to the other feature request around previews: https://github.com/umbraco/Umbraco-CMS/issues/8673

nul800sebastiaan:
Hey there @pgregorynz - we have had a chat at HQ about this and I just wanted to drop some of the notes from that conversation here. Note that we got very excited by the possibilities and rethinking things, so it's not to shoot down your specific idea, but a record of what our thinking was when considering it! 

- We talked about the complexities of what you see is what you get and how difficult it is to really nail that one
- We had a talk about: why do we need previews at all? Maybe the editing experience in the backoffice needs to be improved instead? 
    - Or, instead, maybe actual live editing on the frontend should be an actual thing 
[...]

PR #9131, opened 2020-10-10
Scale the preview iframe to fit the viewport

skttl:
### Prerequisites

If there's an existing issue for this PR then this fixes related, but not complete fix for #8673 and #8303 

### Description
This PR adds some automatic scaling of the preview iframe in the preview viewer, to make it fit the current viewport. This way you can preview your page as if it was shown in ie. a desktop sized browser window, but you don't need a window of the same size.

Modern browser dev tools (seen in Chrome, Edge and Firefox) does the same in responsive mode.

To test:
Open a node in the previewer, make sure your browser window is reasonably small. Change the viewmode in the bottom right corner, and notice how the iframe scales up and down when needed, to fit in your viewport. When in fullscreen, no scaling should be applied though.

R1pzPilAEH

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/canvas-designer.less
+++ b/src/Umbraco.Web.UI.Client/src/less/canvas-designer.less
@@ -192,6 +192,7 @@ body {
 
 #demo-iframe-wrapper {
     transition: all 240ms cubic-bezier(0.165, 0.84, 0.44, 1);
+    flex-shrink:0;
 }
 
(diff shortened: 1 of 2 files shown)


emmaburstow:
Hi @skttl 

Thanks for the PR. You're a ⭐ . We'll take a look and shout if we need you,

Emma

nathanwoulfe (approved):
@skttl this is really slick, nice work

PR #9132, opened 2020-10-10
Itempicker: Fix filter check

BatJan:
### Prerequisites

### Description
@nul800sebastiaan This PR fixes the issue where the search filter was not being displayed when used in the "template" context - Since the model is not immutable it's not enough to assume that the filter should be displayed or hidden if model.filter is true or false since it can also be undefined. Therefore it's now changed from `ng-if="model.filter"` to `ng-if="model.filter !== false` 👍 

**Before**
item-picker-template-before

**After**
item-picker-template-after


emmaburstow:
Hey @BatJan!

Thanks for the PR. We'll take a look and shout if we need anything. Happy Hacktoberfest!

Em

emmaburstow (approved):
Perfect :)

nul800sebastiaan:
This broke in v8.8.0, have cherry picked for 8.8.1 (and higher versions) in 7d3df87b7fd72429c5ebd44edb14896c82a973be

PR #9136, opened 2020-10-11
NetCore: Further integration tests migrations

AndyButland:
In this PR I've migrated some more integration tests, related to issue #7898.

Mostly was fairly straightforward, changed to new base class and pattern for service reference and updated model creation to use the builders.

I found an issue related to how NPoco handles setting IDs on inserted DTOs when doing bulk uploads (this only works for the existing tests using SQL CE, as this doesn't have the SQL bulk copy feature available, and hence falls-back to inserting one record at a time, when it can set the ID on the inserted DTO; with LocalDb, it uses SQL bulk copy, and doesn't set the IDs).  I don't think we're using this feature within the Umbraco code, though we were relying on it in a test with the relations service - but  I could amend the test to assert in a different way, so think that's all fine. 

I removed some tests based on "content XML" (as the DTO is not available in `Umbraco.Core`, and so presume we no longer need this).

Unfortunately there are still [three failing tests](https://github.com/umbraco/Umbraco-CMS/pull/9136/checks?check_run_id=1240599549) that I've not been able to resolve:

- Two are related to some relation amends being triggered by an event when media is deleted, that doesn't seem to be fired when the original test is run from `Umbraco.Tests`.
- One hits a thrown exception with the comment "should not happen due to the cache policy".


bergmania:
Hi @AndyButland.. As always thanks..

I updated with latest netcore/netcore, and solved an issue in the code base found by the migrated tests. The fix is part of this https://github.com/umbraco/Umbraco-CMS/pull/9117 witch is also merged into this PR, so we need that PR merged before this. Thats why I updated the base branch of this PR.

bergmania:
Doh, when this is an external PR, it seems like Github closes the PR when the base branch is closed/merged. 

@AndyButland, We should just reopen it :)

bergmania:
The test `ContentServiceTests.Can_Rollback_Version_On_Multilingual` seems to be a bit unstable.. I often get a "Expected: 3, but was:  4 on the line `Assert.AreEqual(3, versionsSlimFr.Length);`

We need to investigate that

AndyButland:
Reopened here: https://github.com/umbraco/Umbraco-CMS/pull/9149

PR #9141, opened 2020-10-12
Cleanup examine search results, and adds ability to toggle fields

skttl:
I cleaned up the search results table in Examine Dashboard a little, by moving the columns, and enforcing their widths to make it more predictable. In addition to that, I fixed the Score column to have 4 digits only (the rest is in the title attribute).

I also added a new feature. When clicking the Fields column heading, you can add fields to the table. Not all fields are obviously fit for this, but I see a lot of usecases where this is useful. If you add more columns than the viewport can hold, the table will scroll on the x-axis. To make this a bit more useable, I fixed the Id-column, and the score column so they are always visible.


poornimanayar:
This is ace @skttl ... Let me give this a test :-)

poornimanayar:
@skttl , I have tried getting your PR checked out twice! I cant see the toggle overlay at all. I dont get it in the project at all. Any ideas what is missing?
image

skttl:
Did you try to refresh?
image

When I checked out the branch again, I was missing the file at first - but after a refresh it showed up.

poornimanayar:
Having a look now :-)

poornimanayar:
I still cant reproduce this :-(

skttl:
Can you see the files on the disk?

poornimanayar:
Yes Soren I can. I have asked the team as well to see if anyone can reproduce it

nathanwoulfe:
Hey @skttl apologies for taking so long to get to this - it's a really cool change and makes the search results way more useful.

I've merged latest from v8/contrib, added the joinArray filter to the table cells to ensure they render correctly (otherwise values are rendered as a stringified array), added one-time binding to the cells to avoid recalculating that filter on every digest (which is a lot), and then had to make an update to the filter to ensure it didn't explode when the array was empty/null, which is the case when a field has no data.

Only other bit I'm not sure on is the link on the Fields header - without knowing it's there, it's not obvious for a user. I had a fiddle with adding an icon-button, but it didn't really work. Could be that my choice of icons wasn't great. Regardless, I think this should be merged, can always review the Fields link in future.

nathanwoulfe (approved):

nathanwoulfe:
Had some failing tests where the joinArray filter was expected to fail (!!) - I've updated the filter to ensure the failing cases no longer throw errors, and updated the tests accordingly.  Also fixed some typos in the jsdocs.

Failures were caused by passing a non-object argument (in which case the test were expecting the filter to throw errors). These cases are now handled elegantly (null/undefined return an empty string, number/string values are returned unmodified). This means that using the filter on non-object data (ie a string) won't break the UI, it will work as if the data was a single-item array.

Will wait for green from the build server, then merge.

Issue #9128, opened 2020-10-09
Make All Umbraco Composer Public


bielu:
Hi, 
I really don't understand why we have internal composer and components, which make really hard to manage order, if we need to make sure some composers has to be register before our ones.
As example:
PropertyEditorsComponent and PropertyEditorsComposer 
make me really hard time as it is random if that component will be register before mine or after.
here is full list of composers:
WebProfilerComposer - internal
PropertyEditorsComposer - internal
SchedulerComposer - internal
list of components:
WebProfilerComponent
DisabledModelsBuilderComponent
PropertyEditorsComponent
SchedulerComponent
Maybe I am missing some point, but I really think they should be public.

nul800sebastiaan:
Seems like a good idea, public but sealed would be our preference for now, care to shoot us a PR? 🙏 😄

umbrabot:
Hi @bielu,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bielu:
@nul800sebastiaan send as always :)

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/9142 and cherry picked for 8.6.6 in b96776949ffbe4dc7562b12e4a4366e56ad320e6 - will also be available in 8.7.1, 8.8.1 and 8.9.0.

ronaldbarendse:
@bielu All these composers inherit from the `ICoreComposer` interface, so they always run before the `IComposer` and `IUserComposer` implementations. If you want to run your component before any of the core ones, you can use `Insert(0, ...)` instead of `Append(...)` while adding the component, right?

Making these composers/components public would also mean they can be disabled, so Umbraco should be able to run when this is the case:

```c#
[assembly:DisableComposer(typeof(WebProfilerComposer))]
[assembly:DisableComposer(typeof(PropertyEditorsComposer))]
[assembly:DisableComposer(typeof(SchedulerComposer))]
```

[...]

PR #9142, opened 2020-10-12
Make composers and components public

bielu:
This pr solves https://github.com/umbraco/Umbraco-CMS/issues/9128
Testing: Nothing really changes, as we only change accessibility of composers.


emmaburstow:
Hey Arek,

Thanks for this PR. Looks ace :) We'll take a look at the work and make sure that as you say 'nothing really changes'. If all is good, we'll merge,

Emma

PR #9144, opened 2020-10-12
V8/Fixes preview of segmented content

### LINKED ISSUE #9143: Preview of segment content does not work
Currently previewing of segmented content does not work.
When having only one language it will result in a error. When you have multiple languages it will show a preview, but you don't see your segmented content

## Umbraco version

I am seeing this issue on Umbraco version: 8.7 and higher

## Steps to reproduce

- Add the following app setting to web.config `<add key="Umbraco.TestData.Enabled" value="true"/>`
- And you need to check that you have the Umbraco.TestData.dll in the bin folder. Otherwise, you can copy it from src\Umbraco.TestData\bin\Debug
- Add a property with the alias testProperty to the Content Page doctype. Use a textbox as datatype. Make sure this doctype is not allowed to vary by culture
- Call this url in your browser : http://localhost:8900/umbraco/surface/SegmentTest/EnableDocTypeSegments?alias=contentPage&propertyTypeAlias=testProperty. The output should be The document type contentPage and property type testProperty now allows segments
- Get the id of a Content Page
- Call this url from your browser. Update it with your id. http://localhost:8900/umbraco/surface/SegmentTest/AddSegmentData?contentId=1213&propertyAlias=testProperty&value=hello%20world&segment=TestSegment
- The output should be Segment value has been set on content 1213 for property testProperty
- Now refresh your page in the backend and check if there is segment variant avaiable
- Select the variant testsegment and hit save & preview
- This will result in a error. Because the value "invariant" is passed as culture

dawoe:
### Prerequisites

This fixes #9143 

### Description
This PR makes sure you don't get a error when previewing segmented content and that you actually see the values of your segment content.

Steps to reproduce can be found in the related issue


emmaburstow:
Hey Dave!

You've been busy ;) Thanks for the work. We'll take a look and merge if all is cool.

Emma

src/Umbraco.Web/Routing/ContentFinderByIdPath.cs
```diff
@@ -54,14 +54,40 @@ public bool TryFindContent(PublishedRequest frequest)
 
                     if (node != null)
                     {
+                        string culture = null;
```
nul800sebastiaan, line 57: Looking at the code below, the culture will never be null and this would be a lot easier to read, am I missing something?

```csharp
var culture = frequest.Culture.Name;
//if we have a node, check if we have a culture in the query string
if (frequest.UmbracoContext.HttpContext.Request.QueryString.ContainsKey("culture"))
{
    culture = frequest.UmbracoContext.HttpContext.Request.QueryString["culture"];

    if (culture != "invariant")
    {
        //we're assuming it will match a culture, if an invalid one is passed in, an exception will throw (there is no TryGetCultureInfo method), i think this is ok though
        frequest.Culture = CultureInfo.GetCultureInfo(frequest.UmbracoContext.HttpContext.Request.QueryString["culture"]);
    }
}
```

src/Umbraco.Web/Editors/PreviewController.cs
```diff
@@ -99,7 +99,7 @@ public JavaScriptResult Application()
         /// </summary>
         /// <returns></returns>
         [UmbracoAuthorize]
-        public ActionResult Frame(int id, string culture)
+        public ActionResult Frame(int id, string culture, string segment)
```
nul800sebastiaan, line 102: This is a breaking change, we'll need to obsolete the old method signature and tell people to use the new one including the segment.

nul800sebastiaan (changes requested):
Thanks @dawoe! I have left a couple of requests for changes, after which we can test and merge! 👍

src/Umbraco.Web/Editors/PreviewController.cs
dawoe: Hi @nul800sebastiaan 

I don't think many people will be using this method directly in their code. And it is a route...so if the parameter is not present it will be empty by default.

So should I fix this ?

Dave

nul800sebastiaan: Ah yes, I forgot how the routing for webapi works! Agreed that nobody should be relying on this in their custom c# code. All good then! 👍

src/Umbraco.Web/Routing/ContentFinderByIdPath.cs
dawoe: Updated

PR #8525, opened 2020-07-30
Fixes #8433 ensures there are no write locks when updating members during login

Shazwazza:
Fixes #8433 

Ensures there are no write locks when updating members during login

This also finalizes an outstanding TODO that has been there for a very very long time that will hugely boost performance for sites that are membership heavy and have a lot of active logins at once. We never cached members by username but that is how they login and how they re-authenticate each request which is a huge amount of database calls. 

---
_This item has been added to our backlog [AB#8415](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8415)_

PR #9146, opened 2020-10-13
Fixes regression issue with property values being updated in content repositories

### LINKED ISSUE #9093: The value for custom member properties cannot be removed in the back office of v8.8
Hey Guys,

this bug is new in 8.8. I tested this with 8.7 and it works fine there. 

Steps to Reproduce:

Add a custom property editor to the default member type:
image

Create a new member and add a string to the new property Phone
image

Save and refresh the member editor page
The entered value should stay, which is fine.

Change the value, hit save and refresh
The entered value should stay, which is fine.
image

Now remove the entered value, hit save and refresh
The removed value is not removed
image

I tested this with other property types like member group picker or numerics. Same behaviour. So you can add values to custom properties, you can change them but you cant remove the values. 

I tested this behaviour on fresh installed solutions and it came up out of nowhere with the v8.8. Earlier versions work properly in this regard.

Really painful and critical bug in our case, since there are so many business logic related actions for editing members in the BO, which require to remove values from property types.... 

What I also see is that after every edit of a member, no matter if you have added new custom property types or you just edit the default properties, a warning is logged:
image

First I thought this has something to do with our solution, then I tested this behaviour with a fresh installed solution 8.8, edited the member and saw the warning log after every edit and save. 

---
_This item has been added to our backlog [AB#8780](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8780)_

### LINKED ISSUE #9095: Member Index does not get updated on member updates in v8.8
This worked fine in 8.7 but is not working anymore since updated to 8.8.

I tested this behaviour on a fresh installed solution in version 8.8, which failed. No issue in 8.7

Steps to reproduce:
Create a test member of the default member type and save
image

Go to Settings > Examine Management > Member Index
search for the member name --> Member ist not listed
image

Rebuild the member index and search again --> member gets listed
image

Go to members, edit the username for the created member, save
image

Go back to Settings > Examine > Member Index > search for the member > click on member fields
image

Check the edited login name value > its not updated
image

Rebuild the member index, search for the member again, click on the fields >  now its updated
image

The members index should be updated automatically on member updates.

Something happend to the members with the upgrade to 8.8, see my other bug report: https://github.com/umbraco/Umbraco-CMS/issues/9093

I retested this scenario on a fresh installed 8.7 to be sure that it worked fine there and the tests passed.

---
_This item has been added to our backlog [AB#8783](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8783)_

Shazwazza:
Fixes #9095 - caused by #8525 with changes here https://github.com/umbraco/Umbraco-CMS/pull/8525/files#diff-2ab7be5e592894a2e3dfac615d7a94a5ae6ab11e1948312fb515749caca04b79R12 because the ExamineComponent.MemberCacheRefresherUpdated event wasn't updated to handle the RefreshByPayload message type

Fixes #9093 - caused by #8525 with changes here https://github.com/umbraco/Umbraco-CMS/pull/8525/files#diff-0f4c2578706d75656978e4cccd48671380730283de0ec98f0d8e79ef7d7d0ef7R388. The problem was a change to the Member Repository that was not taking into account removed property values. The change was required because the previous way that content properties were persisted was not done atomically and was unsafe with high DB concurrency. This change was also required because a new member service method was added to deal with high concurrency issues which requires this other operation to be isolated correctly. It is not possible without a serializable transaction (which we don't use) to delete a bunch of rows and then re-add them since this is not isolated from other DB operations and weird things can happen along with deadlocks. To fix this we use a SELECT statement with an UPDLOCK hint on the rows returned and then either Insert, Update (or Delete) was is required based on the existing data compared to the data we're saving. The missing part here was the Delete part.

This has been fixed and the method to do this moved to the underlying ContentRepositoryBase so it can be re-used by the DocumentRepository and the MediaRepository because both of those are prone to the same concurrency problem and were also trying to just clear rows and re-insert which is unsafe. These repositories now also use this same shared method. I have also moved the logic for inserting property value rows to a shared method when new documents, media or members is created. I've put a TODO there because we can speed that operation up but would like to run some benchmarks on that before we do (that's for a later date).

I've added tests for the ContentService, MediaService and MemberService to ensure that property values can all be added, cleared, updated. 

## Testing

Ensure that #9093 and #9095 problems are fixed and that all tests pass

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
@@ -919,6 +919,8 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
                 }
                 else
                 {
+                    // TODO: we can speed this up: Use BulkInsert and then do one SELECT to re-retrieve the property data inserted with assigned IDs.
+                    // This is a perfect thing to benchmark with Benchmark.NET to compare perf between Nuget releases.
                     Database.Insert(propertyDataDto);
                 }
 
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
@@ -863,6 +863,75 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
 
         }
 
+        /// <summary>
+        /// Inserts property values for the content entity
+        /// </summary>
+        /// <param name="entity"></param>
+        /// <param name="publishedVersionId"></param>
+        /// <param name="edited"></param>
+        /// <param name="editedCultures"></param>
+        /// <remarks>
+        /// Used when creating a new entity
+        /// </remarks>
+        protected void InsertPropertyValues(TEntity entity, int publishedVersionId, out bool edited, out HashSet<string> editedCultures)
+        {
+            // persist the property data
+            var propertyDataDtos = PropertyFactory.BuildDtos(entity.ContentType.Variations, entity.VersionId, publishedVersionId, entity.Properties, LanguageRepository, out edited, out editedCultures);
+            foreach (var propertyDataDto in propertyDataDtos)
+            {
+                Database.Insert(propertyDataDto);
+            }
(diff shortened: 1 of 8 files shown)


nul800sebastiaan:
Functionality is restored for both linked issues, tests pass, looks good to me (and yay for making it DRY!).

PR #9150, opened 2020-10-13
Netcore: Migrate authorized api controller attributes

nikolajlauridsen:
I migrated UmbracoUserTimeoutFilter, and UnhandledExceptionLoggerConfiguration.

UnhandledExceptionLoggerConfiguration is now a middleware filter that applies the UnhandledExceptionLoggerMiddleware to the UmbracoAuthorizedApiController only and will log unhandled errors, regardless of whether the site is in debug or not. 

I also removed the TODO of ```[EnableDetailedErrors]``` in ```UmbracoAuthorizedApiController``` since doing this does not require a filter in netcore, it should be sufficient to change ```JsonExceptionFilterAttribute``` instead. 

I also added a fix to our logging. The ServiceProvider in ```UmbracoCoreServiceCollectionExtensions.AddUmbracoCore``` was built before logging was registered, meaning that there was no logger registered when we tried to create a LoggerFactory for getRuntime/CreateTypeFinder, which meant no logging happened at all. We rebuild the ServiceProvider now after having registered the logging, but it would be optimal if we could move the registration of logging out of CreateCompositionRoot entirely.


bergmania:
Nice work @nikolajlauridsen

PR #9154, opened 2020-10-13
Improvements to validation and accessibility of relationship create view

RachBreeze:
Prior to this change mandatory fields were not being marked as required which caused errors on save, I have now marked these fields as mandatory.

Whilst this is an improvement the accessibility and validation the use of umb-radiobutton on this view is something that requires revisiting, as in order to mark this as mandatory both buttons needed to be marked as required which is a confusing experience to a screen reader user.  Further investigation is required to determine if this is an issue with the create view or umb-radiobutton


RachBreeze:
Hi @nul800sebastiaan  would it be possible to add the accessibility label to this, please as it's an accessibility fix :-)?

nathanwoulfe (changes requested):
Hi @RachBreeze this all looks good, validation fires as expected. Noticed though some odd formatting of the radio buttons, where the text wraps to sit below the input (not related to your changes) - this can be fixed by passing the localization key to the `umb-radiobutton` directive using the `labelKey` parameter. Can then delete the `localize` directives in the radio UL.

Would you mind making that change?

RachBreeze:
@nathanwoulfe sure not a problem :-)

RachBreeze:
Hi @nathanwoulfe 
Thank you for reviewing, I've updated as requested, 
Kind regards

nathanwoulfe:
Thanks @RachBreeze, much appreciated.

nathanwoulfe (approved):

Issue #9155, opened 2020-10-13
Remove legacy Instagram oEmbed provider


ronaldbarendse:
The Instagram oEmbed provider provided by Umbraco uses the endpoint that will be deprecated on October 24, 2020: 
https://github.com/umbraco/Umbraco-CMS/blob/e9627f2ee962a73cd8c007a42845c9129e3339ee/src/Umbraco.Web/Media/EmbedProviders/Instagram.cs#L7

More information can be found on https://developers.facebook.com/docs/instagram/oembed-legacy and about the new endpoint on https://developers.facebook.com/docs/instagram/oembed.

Because the new endpoint requires either an App Access Token (recommended) or Client Access Token, Instagram oEmbed support can't be provided out of the box anymore (and would otherwise at least require rewriting this provider).

The same applies to the Facebook page/post/video oEmbed endpoints, but it looks like Umbraco never provided support for those (see https://developers.facebook.com/docs/plugins/oembed-legacy and https://developers.facebook.com/docs/plugins/oembed).

nul800sebastiaan:
Well that's unfortunate. The recommended replacement is not a very workable solution either as it requires a lot of setup work, unlike this one which worked without authentication whatsoever. 

I'll mark this as breaking because there's no alternative at the moment.

This is also quite disastrous for existing installs and I don't see any fixes apart from someone building a custom OEmbed provider that can take in keys as a config etc, super cumbersome. 😢

These are the new requirements: 
![image](https://user-images.githubusercontent.com/304656/95997436-10549900-0e34-11eb-9540-945b982fd7da.png)

ronaldbarendse:
Very unfortunate indeed, especially because the depreciation period is very short, as the change is introduced as part of the new v8 API endpoints in August: https://developers.facebook.com/blog/post/2020/08/04/Introducing-graph-v8-marketing-api-v8/. Other CMSes also seem to drop core support and leave implementing the authorization to 3rd party plugins.

And I can say from experience that configuring a Facebook app per site, waiting for Facebook to review it before it's live, making sure the access tokens are automatically renewed (even long-lived tokens can expire), etc. is way too much work for having oEmbeds work. This is only gets worse when you need to consider GDPR/tracking laws when displaying the embedded HTML...

ronaldbarendse:
[...]

PR #9156, opened 2020-10-13
Remove legacy Instagram oEmbed provider

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9155.

### Description
The Instagram oEmbed endpoint will be deprecated on October 24, 2020 and return an error code 400. This makes the current Instagram oEmbed provider useless and because the new endpoint requires configuring tokens, can't be fixed without requiring additional (site/environment specific) configuration.

This breaking behavior can already be tested by adding `breaking_change=oembed` as query string parameter (see links in related issue), which can be done by adding this to the current `RequestParams` property of the provider. This will result in the 400 error and not returning the embed HTML markup.

After applying this PR, the Instagram oEmbed provider is removed and will not recognize the instagram.com/instagr.am URLs anymore, skipping the needless request to the API endpoint.


nathanwoulfe (approved):
@ronaldbarendse I see no issues with this - it's a breaking change, technically, but the deprecation of the entire oEmbed endpoint is also a breaking change, so I think this becomes a breaking breaking change. Or a breaking change breaker.

Existing sites won't be any more broken by removing this than they are already, given we've past the Instagram deprecation date.

bjarnef:
Would we implement the new Instagram oEmbed provider instead as mentioned here?
https://developers.facebook.com/docs/instagram/oembed-legacy/

The new one:
https://developers.facebook.com/docs/instagram/oembed
https://developers.facebook.com/docs/graph-api/reference/instagram-oembed/

Looking at the registred providers it doesn't look like we have a Facebook oEmbed provider either:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Runtime/WebInitialComposer.cs#L271-L284

https://oembed.com/
https://developers.facebook.com/docs/plugins/oembed/

bjarnef:
Actually it seems there has been made some changes recently:
https://github.com/iamcal/oembed/issues/468
https://github.com/iamcal/oembed/pull/470

where the new https://graph.facebook.com/v8.0/instagram_oembed API endpoint has been added.

nathanwoulfe:
From the discussion in #9155, there's a fair bit work left for devs to implement the new embed (active, approved Facebook app for starters), whereas the deprecated embed was essentially plug n play. 

I think the additional work required to set up the new embed negates the benefit of having it out of the box in the backoffice. That's my take, at least.

PR #9160, opened 2020-10-14
Added documentation for umbControlGroup directive

callumbwhyte:
I have added the metadata required to the `umbControlGroup` directive so that the attributes available can be listed in the backoffice documentation.

I did my best to give an accurate description of the purpose for each attribute, but this was inferred from how they are used in the code itself. Perhaps someone can double-check my understanding is correct, and propose any better descriptions if anything is unclear.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/html/umbcontrolgroup.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/html/umbcontrolgroup.directive.js
@@ -3,12 +3,12 @@
 * @name umbraco.directives.directive:umbControlGroup
 * @restrict E
 
-@param {string=} label The label for the control group field
-@param {string=} description The description for the control group field
-@param {boolean=} hideLabel Set to <code>true</code> to hide the label
-@param {string=} alias The alias of the field within the control group
-@param {string=} labelFor The alias of the field that the label is for, used for validation
-@param {boolean=} required Set to <code>true</code> to mark the field as required
+@param {string=} label The label for the control group field.
+@param {string=} description The description for the control group field.
+@param {boolean=} hideLabel Set to <code>true</code> to hide the label.
+@param {string=} alias The alias of the field within the control group.
+@param {string=} labelFor The alias of the field that the label is for, used for validation.
+@param {boolean=} required Set to <code>true</code> to mark the field as required.
 
 **/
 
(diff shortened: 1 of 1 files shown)


abjerner (changes requested):
Not really consistent throughout the docs, but could you add a period at the end of each line?

abjerner:
Hi Callum and thanks for helping out with the docs 👍 

I'm not really sure what the `=` means for each parameter? I don't recall seeing those used with ngdocs before.

callumbwhyte:
Hey @abjerner,

PR updated to add a period for consistency!

The `=` is JSDoc syntax for optional parameters. See https://jsdoc.app/tags-param.html It's used in other backoffice UI docs too, such as in [umb-button](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbutton.directive.js#L58).

Cheers,
Callum

abjerner:
@callumbwhyte great. Thanks for clarifying - I didn't know that 👍

IIRC there was some problems in the past with using `<code>` in the comments. I'm not sure whether I remember incorrectly, or this have been fixed since then, but I'll have to test that before approving 😉

callumbwhyte:
@abjerner 👍 Again the use of `<code>` was copied from `umb-button`... Doesn't look like the element is styled differently but [it does render](https://our.umbraco.com/apidocs/v8/ui/#/api/umbraco.directives.directive:umbButton#usage_directive-info)...

abjerner (approved):

abjerner:
@callumbwhyte the missing styling was probably the issue then.

For instance for the `editorService`, the list of parameters looks like on the screenshot. I think I did a few PRs for this for Hacktoberfest last year. The comments are using back ticks, but they are also rendered as `<code>` elements in the output, so the issue seems to depending on the container.

Eg. as the `editorService` has methods, there is a `.methods` container. The fix must then be to add similar styling for `.content .usage_directive-info code`.

```css
.content .methods code {
    border: 1px solid #e1e1e8;
    background-color: #f7f7f9;
    padding: 0px 3px;
    font-size: 85%;
    color: #d14;
    white-space: nowrap;
}
```

image

As the issue seems to be with the styling, I've gone ahead and approved your PR. We can then have a look at the styling through another PR. If you don't beat me to it, I might have a look at that 😉 

Thanks again 👍

PR #9161, opened 2020-10-14
Styling all code elements in backoffice UI docs

callumbwhyte:
As highlighted by @abjerner in #9160 there's no styling for `<code>` blocks in the Umbraco backoffice docs styles. Well, there is, but it's only for method info.

This change makes sure all `<code>` blocks in the "content" section of a docs page are formatted correctly. I feel this is more future proof than specifically targeting which section of the page the block appears in.

I cleaned up some of the styles whilst I was at it too...

### DIFF
--- a/src/Umbraco.Web.UI.Docs/umb-docs.css
+++ b/src/Umbraco.Web.UI.Docs/umb-docs.css
@@ -2,9 +2,9 @@
 html {
     font-family: 'Segoe UI', Tahoma, Helvetica, sans-serif;
 }
+
 body {
     font-family: 'Segoe UI', Tahoma, Helvetica, sans-serif;
-
 }
 
 .container, .navbar-static-top .container, .navbar-fixed-top .container, .navbar-fixed-bottom .container {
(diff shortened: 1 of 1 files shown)


abjerner:
Awesome @callumbwhyte 👍 

I agree that this seems more future proof. It also doesn't style the annoying `<code>` element inside the `<h1>`, which I think was my reason for creating the initial rule.

nathanwoulfe (approved):
@callumbwhyte easiest way to see this in action? Changes read fine, just want to eyeball it before I merge

callumbwhyte:
Hey @nathanwoulfe,

You can build and run the docs locally as per the [README](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Docs/README.md)

Worth checking a few different docs like `umbControlGroup` and `editorService` to see a range of different uses.

Cheers

nathanwoulfe:
Read the readme? That's genius!

nathanwoulfe:
@callumbwhyte finally got back around to this - all good, looks great. Did notice a build issue with the docs project, related to (I think) a mismatch between my local node version (14.15.0) and the gulp version referenced in the docs package.json (3.9.1). Resolved it locally with a npm-shrinkwrap file, but will need to look at updating gulp...

PR #9165, opened 2020-10-14
Remove orderBy filter for languages/variants

ronaldbarendse:
### Prerequisites

This fixes regressions introduced in PRs #9006 and #9007.

### Description
As per https://github.com/umbraco/Umbraco-CMS/pull/9006#issuecomment-708682932, the languages/variants were correctly sorted alphabetically, but with the default language first. This reverts the custom `orderBy` filters in the AngularJS views and ensures the ordering retrieved from the server is used instead.

This ordering is done on the server in the `LanguageMapDefinition`, so even though the `ILanguageRepository` always returns the languages sorted on ID (which can be different per environment), this ensures the first is always the default and the rest is sorted on name:

https://github.com/umbraco/Umbraco-CMS/blob/b29fd4ada3c1eb7a703c3a3a0131ecfa990c3f97/src/Umbraco.Web/Models/Mapping/LanguageMapDefinition.cs#L41-L56

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/application/umb-navigation.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/application/umb-navigation.html
@@ -23,7 +23,7 @@
                         class="umb-language-picker__dropdown-item"
                         ng-class="{'umb-language-picker__dropdown-item--current': language.active}"
                         ng-click="selectLanguage(language)"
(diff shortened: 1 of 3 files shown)


nathanwoulfe (approved):
@ronaldbarendse this makes sense. In 8.12, I'm fairly sure I saw the default language in the content header is annotated as the default, which helps to clarify why it is first.

PR #9169, opened 2020-10-15
Migrated RenderMvcController and RenderIndexActionSelectorAttribute

bergmania:
- Migrated RenderMvcController and related
  - RenderIndexActionSelectorAttribute
     - Had to reimplement the tests too


src/Umbraco.Tests.Integration/Services/ContentServiceEventTests.cs
```diff
@@ -20,7 +20,7 @@ namespace Umbraco.Tests.Integration.Services
         Logger = UmbracoTestOptions.Logger.Console)]
     public class ContentServiceEventTests : UmbracoIntegrationTest
     {
-        private IContentTypeService ContentTypeService => GetRequiredService<IContentTypeService>();
+        private IContentTypeService IContentTypeService => GetRequiredService<IContentTypeService>();
```
AndyButland, line 23: This naming jumps out as a bit funny to me - i.e. a property with an `I` prefix, when the other similar properties don't.  Might be better as it was before, and use the fully qualified namespaces when you need to call the static methods on `ContentTypeService`?

src/Umbraco.Tests.Integration/Services/MemberTypeServiceTests.cs
```diff
@@ -17,22 +18,22 @@ namespace Umbraco.Tests.Integration.Services
     public class MemberTypeServiceTests : UmbracoIntegrationTest
     {
         private IMemberService MemberService => GetRequiredService<IMemberService>();
-        private IMemberTypeService MemberTypeService => GetRequiredService<IMemberTypeService>();
+        private IMemberTypeService IMemberTypeService => GetRequiredService<IMemberTypeService>();
```
AndyButland, line 21: Same comment on naming for this.

src/Umbraco.Web.Website/Controllers/RenderMvcController.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        /// <typeparam name="T">The type of the model.</typeparam>
+        /// <param name="model">The model.</param>
+        /// <returns>The action result.</returns>
+        /// <remarks>If the template found in the route values doesn't physically exist, then an empty ContentResult will be returned.</remarks>
+        protected IActionResult CurrentTemplate<T>(T model)
+        {
+            var template = ControllerContext.RouteData.Values["action"].ToString();
+            if (EnsurePhsyicalViewExists(template) == false)
+                throw new Exception("No physical template file was found for template " + template);
```
AndyButland, line 79: I realise this is just copied, but while you are at it... probably should be `InvalidOperationException` as you've used above (to adhere to [this recommendation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/exceptions/creating-and-throwing-exceptions#things-to-avoid-when-throwing-exceptions)).

Could use string interpolation here too.

AndyButland (commented):
Looks good to me - I added a couple of minor comments.

PR #9171, opened 2020-10-15
Speed up indexing for Media and reduce memory use by reducing array allocations and caching writer name look up.

nzdev:
### Prerequisites

### Description
What?
Reduce memory allocations and save on profile lookups for writerId and creator Id while indexing.
Reuse arrays that are commonly instantiated. Retrieve the names for writers for media indexing the same way this is being done for content.

Why?
Save memory, database, improve performance.

How to test?
Existing tests pass.


nathanwoulfe:
@nzdev let's get this into 10 - I can't change the target branch, likely because v8/contrib no longer exists. If you could retarget a fresh PR to v10/contrib, we can tee it up for review.

Not sure why this wasn't merged, possible I was waiting for a response on my comment.

PR #9173, opened 2020-10-15
Ability to add notifications for users and members

patrickdemooij9:
This problem was pointed out by @abjerner on the Umbraco slack channel. You were able to add EventMessages for members, but they wouldn't be displayed on the front-end.

I added the logic for Members, but then I also had to implement them for Users.

672e412b76ee62fa3fe3bee27e49a766

This is the piece of code that I used for testing. (Content was used for just figuring out the logic for notifications)
```
    {
        {
            MemberService.Saved += MemberServiceOnSaved;
            ContentService.Saved += ContentServiceOnSaving;
            UserService.SavedUser += UserService_SavedUser;
        }

        private void UserService_SavedUser(IUserService sender, SaveEventArgs<Core.Models.Membership.IUser> e)
        {
            e.Messages.Add(new EventMessage("Hello", "Hello World for user", EventMessageType.Info));
        }

        private void ContentServiceOnSaving(IContentService sender, ContentSavedEventArgs e)
        {
            e.Messages.Add(new EventMessage("Hello", "Hello World for content", EventMessageType.Info));
        }
        private void MemberServiceOnSaved(IMemberService sender, SaveEventArgs<IMember> e)
        {
            e.Messages.Add(new EventMessage("Hello", "Hello World for member", EventMessageType.Info));
        }
    }
```


mikecp:
Hello @patrickdemooij9 ,
Thanks for this PR, it is indeed better if the messages are actually displayed :-)
I had a look at it and I noticed a strange behavior though: when I save without making any change, I get the same display as in your video, but if I actually update a field (e.g. comments) and then save, I get the "blue" message displayed twice.

Could you maybe check if you have the same issue at your side? Otherwise I guess it will have to do with my local install/debug run somehow...

Thanks and cheers!

patrickdemooij9:
Hi @mikecp 

I was able to reproduce the problem, but it's not something that has been caused by my code changes. When editing the comment, the system will save the member twice (not quite sure why)

The first save happens at MembershipHelper.UpdateMember (or UpdateWithMembershipProvider in the MemberController)
The second save happens at Services.MemberService.Save(contentItem.PersistedContent) in the MemberController.

I am not quite sure why the system is saving the member twice. It sounds like it has something to do with the default properties and the custom properties of a member.

PR #9175, opened 2020-10-16
Update and fix Umbraco's SmidgeRuntimeMinifier implementation and update to latest Smidge

Shazwazza:
SmidgeRuntimeMinifier  was not using Nuglify to process the files, that is because the Smidge.Nuglify project wasn't updated to netcore 3 yet which is now done along with a bunch of other fixes some of which stemmed from testing in umbraco netcore, see https://github.com/Shazwazza/Smidge/milestone/13?closed=1

This hasn't been released yet but this PR is targeting the pre-release version of Smidge 3.2

Prior to all of this we were getting unhandled exceptions occurring during js processing because we are not including anything already minified in the minification process by way of Smidge's processor conventions. When this happens it means that existing sourcemap declarations for minified files remain as-is which are relative paths of the current request. when these are bundled that relative path is Smidge's endpoint for minifying and it will throw an exception for .map requests and also because those don't exist, so then chrome also has exceptions.

Smidge 3.2 fixes this by ensuring that all remaining sourcemap declarations are translated to their correct locations and this is now working well. You can see in this image which shows that we now have source maps for all minified resources! 🎉 including our own resources which is possible with Nuglify.

image

There's some custom settings that are required for our purposes and Nuglify because we don't name our angular injection correctly with regards to minification and the angularjs standards which require naming your injected services with strings else minification engines can rename those - so we turn off renaming in Nuglify to work around that.

There's still a lot of work to do to make this all work nicely though because Nuglify is a fully fledged minifier that creates a abstract syntax tree (AST) to process the JS. This is way slower than the original JSMin implementation but is far more effective. See this old benchmark for an idea https://github.com/Shazwazza/Smidge/wiki/Custom-pre-processing#minification-benchmarks . You will notice that when running with non-debug mode and changing the RuntimeMinification:Version to something new and refreshing the back office that it will take a few seconds longer than before to load the first time while Nuglify processes all of  these files. The result is a much smaller payload compared to JsMin and we get source maps and the delay is only one time... but we can do even better:

It is possible to just keep using JsMin so we can always have that as a fallback option but what we really need to do is fix our gulp build to do the following:

* In our release build (as opposed to `gulp dev`):
* Export all production libs that we use - the minified versions. We are currently not doing this for tons of things.
* Bundle all of our required production libraries into one file.
* Minify and bundle all of our own angular code.
* (All of the above should have source maps working and also translate the existing source map declarations to their absolute paths like Smidge is now doing, I'm sure that's a normal thing to do in gulp/js minification/bundling tools)
* Minify and bundle all of our own css files
* Do not use Smidge to load all of the above, just load in these 3x bundled/minified resources directly
* Use Smidge to load in all other assets which should mainly just be dynamic assets that are declared in the package.manifest files

Then everything is waaaay faster to load and we aren't runtime bundling anything that we don't need to. I don't think the above is a huge job and someone in the community can surely help out with this one.

## Testing

Change your RuntimeMinification:Version  in config to something new and have Hosting:Debug set to true in your appsettings.Development.json and ensure the back office still works as you expect. Then change to Hosting:Debug to false and reload the back office, it should still work as you expect except not everything is bundled with Nuglify and there's no exceptions, and source maps will work in the "Sources" tab of your chrome/edge dev tools.


src/Umbraco.Web.Common/RuntimeMinification/SmidgeRuntimeMinifier.cs
```diff
@@ -21,31 +21,39 @@ public class SmidgeRuntimeMinifier : IRuntimeMinifier
...
             _smidge = smidge;
-            _preProcessPipelineFactory = preProcessPipelineFactory;
             _hostingEnvironment = hostingEnvironment;
             _smidgeConfig = smidgeConfig;
             _configManipulator = configManipulator;
-        }
 
-        private PreProcessPipeline JsPipeline => _jsPipeline ??= _preProcessPipelineFactory.Create(typeof(JsMinifier));
-        private PreProcessPipeline CssPipeline => _cssPipeline ??= _preProcessPipelineFactory.Create(typeof(NuglifyCss));
+            _jsMinPipeline = new Lazy<PreProcessPipeline>(() => _bundles.PipelineFactory.Create(typeof(JsMinifier)));
```
bergmania, line 48: Should this not be NugilifyJs now?

src/Umbraco.Web.Common/Umbraco.Web.Common.csproj
```diff
@@ -26,8 +26,8 @@
       <PackageReference Include="NETStandard.Library" Version="2.0.3" />
       <PackageReference Include="Serilog.AspNetCore" Version="3.4.0" />
       <PackageReference Include="SixLabors.ImageSharp.Web" Version="1.0.0" />
-      <PackageReference Include="Smidge" Version="3.1.1" />
-      <PackageReference Include="Smidge.Nuglify" Version="2.0.0" />
+      <PackageReference Include="Smidge" Version="3.2.0-build.244" />
+      <PackageReference Include="Smidge.Nuglify" Version="3.2.0-build.244" />
```
bergmania, line 30: When do you expect these are released to the official nuget feed?. Do we need to upload them to our own alpha feed like we have done with examine, if we ship a new alpha?

src/Umbraco.Web.Common/RuntimeMinification/SmidgeRuntimeMinifier.cs
Shazwazza: It could be, i haven't tested that. This is purely used only for minifying Action results which IMO will be better to use JsMinifier since it's a simple minifier and very fast compared to nuglify and normally this is only used to minify small json results.

src/Umbraco.Web.Common/Umbraco.Web.Common.csproj
Shazwazza: I can push this week

PR #9181, opened 2020-10-16
9157: Use the sorting event instead of the saving

### LINKED ISSUE #9157: Sorting event not fired in ContentService
## Umbraco version

The offending code can be seen at https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Services/Implement/ContentService.cs#L2359 where it references the Sorting event name, but actually passes in the Saving event handler.

Reproduction
------------

### Bug summary

When sorting content through the Umbraco back-office, it should first fire the Sorting event, then the Saving event, then the Saved event, and finally the Sorted event.  It instead fires the Saving event twice.

### Steps to reproduce

1. Create a Composer and Component that registers an event handler method for the ContentService.Sorting event and writes out a log message
2. Open the back-office and sort some nodes

### Expected result

There should be a log message written out that shows the sorting event fired.

### Actual result

No message is written

patrickdemooij9:
Fixes #9157 

Can be tested with the following class:
```
    {
        {
            ContentService.Sorted += ContentService_Sorted;
            ContentService.Sorting += ContentService_Sorting;
        }

        private void ContentService_Sorting(IContentService sender, SaveEventArgs<IContent> e)
        {
            e.Messages.Add(new EventMessage("Test", "Sorting"));
        }

        private void ContentService_Sorted(IContentService sender, SaveEventArgs<IContent> e)
        {
            e.Messages.Add(new EventMessage("Test", "Sorted"));
        }
    }
```

However, sorting actually doesn't have a way to show the event messages on the frontend. I check it by putting some breakpoints on the functions and seeing if they would be hit.

### DIFF
--- a/src/Umbraco.Core/Services/Implement/ContentService.cs
+++ b/src/Umbraco.Core/Services/Implement/ContentService.cs
@@ -2356,7 +2356,7 @@ namespace Umbraco.Core.Services.Implement
             if (raiseEvents)
             {
                 //raise cancelable sorting event
-                if (scope.Events.DispatchCancelable(Saving, this, saveEventArgs, nameof(Sorting)))
+                if (scope.Events.DispatchCancelable(Sorting, this, saveEventArgs, nameof(Sorting)))
                     return OperationResult.Cancel(evtMsgs);
 
                 //raise saving event (this one cannot be canceled)


patrickdemooij9:
@nul800sebastiaan is this something that could be merged? We would love to use the sorting event

nathanwoulfe (approved):
Hey @patrickdemooij9, hopefully this one was worth the wait, sorry it took so long to get a simple change merged 😃

Issue #8466, opened 2020-07-23
.Url returning url from the wrong language


patrickdemooij9:
I've been having this bug for a very long time, but was never able to figure out what happend. I was finally able to figure it out and reproduce it with a new Umbraco installation:

How to reproduce:

1. Install new Umbraco starter site (I've encountered this in 8.5.5 and in 8.6.4, but I am sure that there are more versions affected here)
2. Go to the backoffice
3. Create a new language that's different from the main language. It doesn't matter what language it is.
4. Go to the "Products" document type and turn "Allow varying by culture" to on.
5. Go to your content tree and add 2 domains to the home page, I used subfolders for this: 
 ![image](https://user-images.githubusercontent.com/11466511/88289113-0881ce80-ccf5-11ea-8502-e4fc900ff4c8.png)
6. Go to the products page and publish both language variants for it.
7. Go to the products page of your first language.
8. Go to the products page of your second language.
9. As you should see, the products page of your second language has the wrong urls that go to the products. Your first language however is still fine.
10. Go back to the backoffice and publish both variants again.
11. Go to the products page of your second language.
12. Go to the products page of your first language.
13. You should now see the bug that we had before, but the other way around. Your second language products should now be fine, while the first ones just result in 404.

Here is a video showcasing the bug and why it can easily break your website:
https://share.vidyard.com/watch/MRCqs8AKx4uVZZUTJZU4Mt?

Now why does this happen:
At first, I didn't know why this happend (Hence making this ticket a long time ago: https://github.com/umbraco/Umbraco-CMS/issues/7930). But after looking further into it, I managed to figure out exactly what is going wrong.

The structure is very important here. We have an overview page (Products) that has culture variants and we have the children (Product) that does not have culture variants. When we do an .Url call to the product page, no culture is passed with it. The first important point in the code is here: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Routing/UrlProvider.cs#L127. Because the product page is no culture variant, the NULL culture will not be set to the current culture being used! For the first request, this is fine because the current culture will be used further down the code. However, the problem lies in combination with https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PublishedCache/NuCache/ContentCache.cs#L132. Here, the result of our product page with culture NULL will be cached. This is why the first request goes fine, but the second request shows the results of the first request.

[...]

PR #9183, opened 2020-10-16
9153: Only show the cultures of the node in preview

### LINKED ISSUE #9153: Preview shows all configured languages even when not varying by culture
The preview shows all configured languages even when a item is not varying by culture.

## Umbraco version

I am seeing this issue on Umbraco version: 8.7 and up, could be on lower onces as well.

Reproduction
------------

### Bug summary

When previewing all the languages are shown that are configured. I would only expect the language choice to popup when a item is configured to vary by culture. And only see the languages that are already created.

This is somehow how related to #7612. Would be nice if we can fix them in one go.

### Specifics

The languages are retreived here :
https://github.com/umbraco/Umbraco-CMS/blob/eec1e95eb29d9589ecb1b2de395dc9e5a91256b7/src/Umbraco.Web/Editors/PreviewController.cs#L67

### Steps to reproduce

1. Install a new Umbraco installation and use the default starter kit.
2. Add a additional language.
3. Preview a item. 

### Expected result

You will see no language choice in the preview

### Actual result

You see a language choice in the preview

image

patrickdemooij9:
Fixes #9153 

This uses the .Cultures of the content node instead of always using all the languages on the website.

I tested it the following way:

1. Using the Umbraco starter kit.
2. Set the People document type to Varies by Culture.
3. Add another language to the website.
4. Add two domains to the root node.
5. Preview the people node and see 2 languages listed in preview modes.
6. Preview the products node and see that no languages are shown.

Now, this will work most of the time, but there is one situation where this doesn't work. If I make the People document Varies by Culture then you'll also have 2 versions of the children under there (Like Matt Brailsford), but those 2 languages will not show up in the language selector for the preview version. This is because .Cultures only returns a single language in that case. I can probably work around it with domains, but I think the issue lays deeper within Umbraco for that. Also see this issue that is related to it: https://github.com/umbraco/Umbraco-CMS/issues/8466

Please let me know if you would still want this PR through even with the problem that I explain above here.

### DIFF
--- a/src/Umbraco.Web/Editors/PreviewController.cs
+++ b/src/Umbraco.Web/Editors/PreviewController.cs
@@ -26,7 +26,6 @@ namespace Umbraco.Web.Editors
         private readonly IUmbracoContextAccessor _umbracoContextAccessor;
         private readonly ILocalizationService _localizationService;
         private readonly IIconService _iconService;
-        private readonly IDomainService _domainService;
 
         [Obsolete("Use the constructor that injects IIconService.")]
         public PreviewController(
@@ -40,8 +39,7 @@ namespace Umbraco.Web.Editors
                 publishedSnapshotService,
                 umbracoContextAccessor,
                 localizationService,
-                Current.IconService,
-                Current.Services.DomainService)
+                Current.IconService)
         {
 
         }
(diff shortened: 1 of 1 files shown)


dawoe:
Hi @patrickdemooij9 

This works. And like you said the solution is not perfect. But improves on how it is now.

I would be happy to see this merged already. 

Dave

dawoe (approved):

nathanwoulfe (approved):
Like @dawoe said, this works nicely and is an improvement on how it currently works.
🎅

PR #9184, opened 2020-10-16
More blue added to the dashboard because we all like more aqua

heyemmaburstow:
### Prerequisites

To Test: load the back office and check the dashboard out :)

### Description

It looks better 

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/dashboard.html
+++ b/src/Umbraco.Web.UI.Client/src/views/common/dashboard.html
@@ -20,7 +20,7 @@
 
                     <div ng-repeat="property in tab.properties">
 
-                        <div class="clearfix">
+                        <div class="" style="color: red; background-color: aqua">
                             <div ng-include="property.view"></div>
                         </div>
 


src/Umbraco.Web.UI.Client/src/views/common/dashboard.html
```diff
@@ -20,7 +20,7 @@
 
                     <div ng-repeat="property in tab.properties">
 
-                        <div class="clearfix">
+                        <div class="" style="color: red; background-color: aqua">
```
nul800sebastiaan, line 23: I don't see the red, but the blue is cool. Maybe turquoise would be better?

src/Umbraco.Web.UI.Client/src/views/common/dashboard.html
```diff
@@ -20,7 +20,7 @@
 
                     <div ng-repeat="property in tab.properties">
```
nul800sebastiaan, line 22: Why is there whitespace here?

nul800sebastiaan (changes requested):
I really think the colour is a bit off, can you update that please? 👍

deanleigh:
Missing ; after aqua

PR #9185, opened 2020-10-16
Refactoring server names from Master to Primary

### LINKED ISSUE #9088: Rename ServerRole.Master in vNext?
Just had a thought while reacting to a tweet today.

Would it be prudent to rename `ServerRole.Master` to `ServerRole.Main` in vNext? (Unicore?)  
Would also mandate renaming in the docs:  
https://our.umbraco.com/documentation/getting-started/setup/server-setup/load-balancing/flexible-advanced

#BLM!

lars-erik:
This is as much work that can be done to fix #9088 in V8.

There's one usage of `ServerRole.Master` and quite a lot of `IsMaster` properties that can't be changed without breaking comptibility. Rest will have to be done to unicore. 🤷‍♂️

I've set up two identical websites and tested publishing from the primary one using the new enum value. Seems legit:

image

Docs have a sibling PR at https://github.com/umbraco/UmbracoDocs/pull/2779.

### DIFF
--- a/src/Umbraco.Core/Models/IServerRegistration.cs
+++ b/src/Umbraco.Core/Models/IServerRegistration.cs
@@ -21,6 +21,7 @@ namespace Umbraco.Core.Models
         /// <summary>
         /// Gets or sets a value indicating whether the server is master.
         /// </summary>
+        [Obsolete("Replaced with IsPrimary. Will be removed from a future version.")]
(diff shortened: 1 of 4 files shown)


lars-erik:
There are two warnings of possible breaking changes in here for now. Dunno how far to go with this. Happy to discuss.

PR #9186, opened 2020-10-16
Add new "umb-confirm-remove" directive

BatJan:
### Prerequisites

### Description
Unfortunately the changes that got merged with #8198 broke forms and probably other packages and therefore the changes got reverted. Therefore I have now been redoing the changes in a new directive called "umb-confirm-remove" instead.

~~I will mark the umb-confirm-action directive as being obsolete and still need to do some further enhancments before this one is ready for review. But got the primary stuff sorted.~~

**Update**: @nul800sebastiaan Ok, so this one is ready for being reviewed :-) Basically I've copy/pasted the changes that was originally made in umb-confirm-action into this directive and ensured it looks and acts as expected the different places where it's being used and I've added an "Obsoleted" text to the documentation for the old directive as well.

Let me know if you have any comments 👍

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbconfirmaction.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbconfirmaction.directive.js
@@ -5,6 +5,7 @@
 @scope
 
 @description
+<p><b>OBSOLETE!!! - Please use <umb-confirm-remove> instead :-)</b></p>
 <p>Use this directive to toggle a confirmation prompt for an action.
 The prompt consists of a checkmark and a cross to confirm or cancel the action.
 The prompt can be opened in four direction up, down, left or right.</p>
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umbconfirmremove.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umbconfirmremove.directive.js
@@ -10,15 +10,13 @@ The prompt can be opened in four direction up, down, left or right.</p>
 <h3>Markup example</h3>
 <pre>
     <div ng-controller="My.Controller as vm">
-        <div class="my-action" style="position:relative;">
-            <umb-confirm-action
-                show="vm.promptIsVisible"
-                direction="left"
-                on-delete="vm.showPrompt()"
-                on-confirm="vm.ConfirmRemove()"
-                on-cancel="vm.hidePrompt()">
-            </umb-confirm-action>
-        </div>
+        <umb-confirm-remove
+            show="vm.promptIsVisible"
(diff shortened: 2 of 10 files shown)


nielslyngsoe:
Hi @BatJan, thanks for making this one.
I'm actually not a fan of the UI of this Confirm-Action. I think we would be better of using the standard dialog. Like when we confirm other things in general, let me know if you see an advantages of this one that I'm not ware off, but otherwise, I would go for this one:

image

:-D

BatJan:
@nielslyngsoe I've actually been meaning to ask you about this - I guess you just confirmed my suspicion that we should indeed make the other overlay the standard now 😂 - Guess I could have saved some work by just asking if we should go down that path instead. But I actually agree - I think the ✅ and ❌ have had their time 😃 All though I can also be nervous that in some cases it might be a bit much "in your face" with the overlay sometimes as opposed to this more subtle approach if I need to play the devils avocado here 🥑 🔱 😅 

I'll be happy to help converting wherever this might be needed 👍 and you can close this PR

nul800sebastiaan:
Just make sure that it's 100% backwards compatible.. Forms loves this umb-confirm directive apparently. 😅

nielslyngsoe:
@BatJan you should always play the devil's avocado 🥑🥑🥑. Plus I love avocadoes. .oO(though not that good for the environment with all this transport, but that's a problem in 2030...)

Never the less you are right, it is up in your face. But also not to be missed, or misread. Which I think is the strength of the standard overlay, nothing special about that as you 'know it' and most users would not even read or barely "see"(obtain the impression) but just press Enter (or click ...) and get on with their things.

Never the less if you experience any software/UI that has a good shift confirm action thingy that made your avocado dance, I would love to hear about it and see if we could be inspired by it.

Thanks 🥑🦖

BatJan:
@nielslyngsoe 😂 Yes, just double checking if that is the direction we want to go in. If I remember correctly right now the places where the overlay is being used is actually different views... When I find some time I might help looking into if we could have a generic view instead and then start replacing / adding it where appropriate so we have a consitent UI 👍 

@nul800sebastiaan Don't worry, refactoring the current confirm action to be using the overlay instead won't affect forms 👼

nathanwoulfe:
Hey @BatJan, this looks great and I'm keen to have proper look and get it merged - is it ready to go? No pending changes from the conversation above?

BatJan:
Hey @nathanwoulfe sorry for the slow reply! I think we actually need to close this one since it would make more sense to make a new one for triggering the standard overlay as Niels and I discussed above 👍🏻 So at some point I will probably do a new PR for that instead 😊

nathanwoulfe:
All good @BatJan!

PR #9187, opened 2020-10-16
Added a new Input to Image Cropper for User Friendly Image Crop Names

deanleigh:
## Reason

Add a user friendly crop name 2 image cropper for the end user well still keeping the crop alias for developers.

This means we can make better hooks for developers that may not be meaningful to the end user.

## Collaborators:

- @callumbwhyte 
- @benjaminc 
- @FransdeJong


deanleigh:
image

deanleigh:
image

deanleigh:
Working on getting the inputs to align with the fields, currently they do not
image

deanleigh:
The first task is to make the container for the inputs behave the same as the individual crops below by making it flex instead of inline-flex
image

PR #9192, opened 2020-10-16
User group picker: Use <umb-search-filter> directive

BatJan:
### Prerequisites

### Description
I have replaced the search markup with the `<umb-search-filter>` directive making maintaining the search box easier not re-implementing the same markup pattern multiple times.

**Usergroup search field after the change has been made**
usergroup-umb-search-filter


nathanwoulfe (approved):
@BatJan lovely jubbly, works perfectly

PR #9194, opened 2020-10-16
Added char counter for dictionary items

### LINKED ISSUE #8668: If a field (data type or dictionary item) has a character limit, there should be some kind of warning or a counter
When I try to save a lengthier dictionary item containing some HTML, ie. I only get the error that something is wrong and not that I have exceeded the max allowed number of characters.

https://snipboard.io/VE9SBH.jpg

Mantus667:
### Prerequisites

fixes #8668 

### Description

This PR adds a char counter to the textareas for dictionary items. Each textarea has its own char counter.
Dictionary items have a maxlength-attribute of 1000 as the DB field has this size. Thus the char counter only counts to zero and does not go to negative values (like the normal textarea one does.

image

To test this issue create a new dictionary item and add text until you reach 1000 chars.
If you save with a few chars less then 1000 and reload the item the counter should also show


nathanwoulfe (approved):
Thanks @Mantus667, this is a really useful addition. I've added a small change to use a single instance of the localize directive for the counter, rather than having a separate screen-reader-only element. Otherwise, all good and works perfectly.

🎅

Issue #8425, opened 2020-07-12
Public Access / Membership misconceptions


marcemarc:
When reviewing Uno from the perspective of somebody new to Umbraco, trying to work out what things are from 'just' what is presented in the Umbraco backoffice, menus and sections. 

Eg. the context of somebody quickly trying to work out the capabilities of Umbraco from spinning up a trial of Uno and clicking around. One thing that is an anomaly, is the presentation of 'Members' in the backoffice and how this relates to the protection of content.

This could probably be addressed by a Tour + Members dashboard explaining the concepts to people are new to Umbraco, how to setup a login and register page with widgets etc - but the anomaly I'm flagging here is that the 'right click' action menu for protecting content is called **'Public Access'**.

This doesn't give off a good scent that this option is related to membership and the Members section.
This doesn't give off a very good scent that this is an option for protecting or restricting content.

The phrase 'Public Access' - possibly means different things in different cultures? but generally is a term that means members of the public have the right to access and view things... https://www.oxfordlearnersdictionaries.com/definition/english/public-access and https://www.lawinsider.com/dictionary/public-access

Whereas you only need to choose this option in Umbraco if you want to 'restrict' access. 

Which means people can struggle to find this option on the menu, it not mapping to the mental model of what they are trying to achieve - and have to learn it's an association with the function of Membership and Content protection.

If you've been using Umbraco for a while you've learned this and don't notice the anomaly.

With the target audience of Uno, (and for everyone really) it might make more sense for people getting started, if 'Public Access' was renamed... either to be:

"Restrict Access"
or
"Protect Content"

to strengthen the association with the 'function' of the action

or

"Member Access"

to strengthen the association with the Members section (and also I think that 'granting access to a member' is conceptually close to the way the functionality works)

nathanwoulfe:
@marcemarc this makes good sense to me. I'd lean to Restrict Access because that's the most accurate descriptor of the action - you use that menu item to restrict access to the content. That, and the text in the overlay uses the term 'restrict access' so it's consistent too.

I don't know who would make this decision, but it seems completely reasonable to me.

nul800sebastiaan:
Restrict access sound perfect to me.. that's exactly what you're doing with that action.

There's a few places where this could need a change but starting with the translation files might be good enough for now.

![image](https://user-images.githubusercontent.com/304656/87319871-804c3e00-c52a-11ea-986e-9fa942932788.png)

umbrabot:
[...]

PR #9197, opened 2020-10-17
Update Language keys for 'Public Access' to become 'Restrict Public Access' …

marcemarc:
…on the action menu

### Prerequisites

- [@ ] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/8425

### Description
The issue https://github.com/umbraco/Umbraco-CMS/issues/8425 describes the misconception over the title 'Pubic Access' for an action menu option which does the opposite, eg denys access to the public!

The discussion came to the conclusion that 'Restrict Access' would be a better option - to lead people to understand the setting meant to restrict access...

This PR updates only the language files to use Restrict Access instead of Public Access, so at least this can be seen/experienced to decide if this change makes sense - or if 'Restrict Access' feels right.

### Screenshots:

RestrictAccessOption

My only concern here with 'Restrict Access' - is it too similar to 'Permissions'? - in the original issue I was suggesting 'Member Access' as the name of the option, because this would 'map' the option to the 'Members' section - which it is implicitly connected to, 

RestrictMemberAccess

certainly though in isolation 'Restrict Access' is a good name to describe the function of the action!

RestrictActionMenu

### Other considerations

Throughout the codebase we refer to 'PublicAccess' as being a thing, eg we have clientside service with methods like

```
   getPublicAccess: function (contentId) {
            return umbRequestHelper.resourcePromise(
                $http.get(
                    umbRequestHelper.getApiUrl("contentApiBaseUrl", "GetPublicAccess", {
                        contentId: contentId
                    })
                ),
                "Failed to get public access for content item with id " + contentId
            );
        },
```
and we have C# events from the PublicAccessService, etc

PublicAccessMentionsInTheCodeBase

This doesn't effect editors, but would lead to developers to question - why is this called PublicAccess?
You could take the view that changing these to match RestrictAccess etc would make sense to do if the menu name does change - but it would also trigger breaking changes etc - really they should be based on the 'alias' of the action menu and not worry about the label that is translatable anyway - the 'alias' for the option though is called 'protect' and not 'public access' :-(

### Audit Trail

From an editor facing perspective the only other mention of 'Public Access' is in the audit trail - where a Public Access changed event is recorded - but in a way that still makes sense - the 'public access' has changed by restricting access to the node...

or we could edge our bets and call the menu item 'Restrict Public Access' ...

RestrictPublicAccess


dawoe (approved):

dawoe:
@marcemarc Looks good to me.

LottePitcher:
@marcemarc 
So if the menu item was changed to "Restrict Public Access..." then having services, audit entries relating to PublicAccess would still make sense, right?  It IS about restricting public access after all. So that 'edge betting' option gets my vote!

And thank you for looking at this, that menu item has always made me think for slightly longer than I wanted to.

marcemarc:
Updated to be Restrict Public Access, I think that makes the most sense in the context, improves the understanding of what the option is, but persists the notion of 'PublicAccess' being a thing in the Umbraco codebase and audit trail

nathanwoulfe (approved):
Thanks @marcemarc, I think the final resolution here was the right one 👍 
I've tweaked the casing on the longer phrases to use `access` rather than `Access` since it's not being used as a proper noun.

There's a whole stack of similar changes that could probably be made throughout the language files - ie sometimes it's Content Type, sometimes it's content type, etc etc et al. Most cases these should be treated as common, not proper nouns. But that's for another day.

PR #9198, opened 2020-10-17
Cache the SVG icons serverside to boost performance

kjac:
### Prerequisites

### Description

There seems to be a severe performance penalty with the wider adaption of SVG icons in the backoffice. The icons are loaded with an XHR, and serverside that XHR does a _lot_ of work in order fetch the SVG:

- Seek out all applicable SVG files on the file system to find the requested one
- Read the SVG file from disk as text
- Perform HTML sanitation on the SVG file contents

...and all of this done for _every single icon_ requested by the backoffice, resulting in a massive CPU/time consumption on the server. The following screenshot is taken after expanding a single node with a bunch of different node types (icons) below - notice the time consumed for each icon being loaded:

image

This PR adds caching to `IconService` to ensure that we at least only scan the file once for all applicable SVG files. Of course we still have to read the files and perform the HTML sanitation, but the result is still remarkably better:

image

I will look into optimizing the cache clientside as well 👍

### DIFF
--- a/src/Umbraco.Web/Services/IconService.cs
+++ b/src/Umbraco.Web/Services/IconService.cs
@@ -2,7 +2,9 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using Ganss.XSS;
+using Microsoft.CodeAnalysis.Operations;
 using Umbraco.Core;
+using Umbraco.Core.Cache;
 using Umbraco.Core.Configuration;
 using Umbraco.Core.IO;
 using Umbraco.Core.Models;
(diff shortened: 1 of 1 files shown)


Shazwazza:
Is it also worth looking into the logic that performs this to figure out why it's slow and how we can make it perform better? Sometimes caching things is a band aid fix and also increase mem consumption. Would be good to know what mem overhead this has, but also good to know if the perf bottlenecks could be fixed or improved without caching. Just throwing out some ideas.

kjac:
@Shazwazza completely agree, caching is a last resort. 

The first thing I did was to dig into the implementation, but I drew a blank there. The SVG icon implementation is based on two relatively expensive operations: Disk access (traversing the file system) and HTML cleanup (`HtmlSanitizer`). I just can't see any way around those, when we have to support an extendable set of file based icons. 

@nul800sebastiaan did analyze the memory usage of this PR, and with the default icon set it's about 1.3 MB:

image

From that output I did the linked PR (#9200) to see if the memory consumption would be noticeably lower using an array as cache storage. It was not.

I'd love to see another way around this. I just can't see how to improve the performance (both in time and CPU utilization) without introducing caching at some level; be it the result of the file system traversal or the final result of the `HtmlSanitizer` output (this PR).

Shazwazza:
@kjac ok cool, was just checking if other options were investigated 👍

nathanwoulfe:
Closing this in favor of #9200, which by all reports is a more efficient approach. Will re-review that one today, and merge if all good.

PR #9199, opened 2020-10-17
Make toggle group labels clickable and aligned correctly

kjac:
### Prerequisites

### Description

The changes in #8826 has unfortunately resulted in misaligned labels for toggle groups:

image

Also the checkbox label is no longer clickable - or rather, clicking it does not have any effect (because the label element swallows up the click event on the parent, which is supposed to be toggling the checkbox):

toggle-group-before

This PR removes the additional padding and margin on the checkbox label, and ensures that it doesn't steal click events from the parent element:

toggle-group-after


nathanwoulfe (approved):

nathanwoulfe:
@kjac works as advertised. Shall merge.

PR #9203, opened 2020-10-17
When a user switches their own language the menu doesn't update to reflect the language change

RachBreeze:
If a user edits their profile and changes the language the menu doesn't change to reflect the language change, as shown below

image

I did consider refreshing the page to reload the menu, however that breaks the paradigm of all other messages which display a success/ fail message after saving.

Instead I've added a localised message `operationSavedHeaderReloadUser` which will prompt the user to reload their browser after the user has been saved.

I have also ensured that the messages are localised to the new language after saving.

Whilst this is an edge case, it's not obvious that the page needs to be reloaded in order for all content to be localised :-)

Created as part of the CanConHackathon


mikecp (approved):

mikecp:
Thanks @RachBreeze for this PR.
It is indeed not obvious that you need to refresh the page !
I took the liberty to already add the French and Dutch translation while I was at it ;-)

RachBreeze:
> Thanks @RachBreeze for this PR.
> It is indeed not obvious that you need to refresh the page !
> I took the liberty to already add the French and Dutch translation while I was at it ;-)

Thank you @mikecp much appreciated :-)

PR #9206, opened 2020-10-17
Change comments to use allowed and disallowed

BatJan:
### Prerequisites

### Description
During the work on some other PR's I noticed that while the variables names had changed the comments had not so I've done that with this PR ☺️


nathanwoulfe (approved):
Thanks @BatJan, I'm 100% confident this will not break anything 😄

Issue #7215, opened 2019-11-22
Allow hiding Links box in Info app for variants


ronaldbarendse:
The following event should hide the 'Links' box on the 'Info' app:
```c#
EditorModelEventManager.SendingContentModel += (sender, e) => e.Model.Urls = null;
```

This works for invariant document types, but when using variants, the box is still shown:

![Links box](https://user-images.githubusercontent.com/1051287/69439460-4afaa200-0d47-11ea-820e-b4b7a1fef6e7.png)

It looks like the following `_.filter` is the culprit, as `_.filter(null)` will return an empty array:
https://github.com/umbraco/Umbraco-CMS/blob/b8e5146baa5b9432bc2c090736f0f58bf6087876/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbcontentnodeinfo.directive.js#L323-L329

And the `ng-if` will evaluate to true for the empty array and show the box:

https://github.com/umbraco/Umbraco-CMS/blob/52dcdb396ff7d68bd76b8f276384f018fd542887/src/Umbraco.Web.UI.Client/src/views/components/content/umb-content-node-info.html#L5

## Umbraco version

I am seeing this issue on Umbraco version: 8.3.0

umbrabot:
Hi @ronaldbarendse,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

ronaldbarendse:
The following also adds a custom message when no URLs are available (empty array):

https://github.com/umbraco/Umbraco-CMS/blob/b8e5146baa5b9432bc2c090736f0f58bf6087876/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbcontentnodeinfo.directive.js#L71-L80

So to completely remove the URL box, it is _required_ to set `e.Model.Urls = null`.

nathanwoulfe:
Fixed in #9207

PR #9207, opened 2020-10-18
Check if urls is null

Mantus667:
Check if node.urls is null and set scope.currentUrls to null and return

### Prerequisites

Related to issue #7215

### Description

When setting node urls to null using EditorModelEventManager.SendingContentModel += (sender, e) => e.Model.Urls = null; then urls box is shown if the node has variants. A check is now added to set scope.currentUrls to null same as for Elements.

To test create a component with the above-mentioned code and open a content node that has variants. The urls box should be hidden.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbcontentnodeinfo.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbcontentnodeinfo.directive.js
@@ -315,7 +315,7 @@
 
             function updateCurrentUrls() {
                 // never show URLs for element types (if they happen to have been created in the content tree)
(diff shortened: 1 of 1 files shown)


ronaldbarendse:
@Mantus667 Thanks for creating this PR, now let's make sure this gets fixed/merged before the issue's first anniversary! 🎉 

I can confirm the issue is still present in Umbraco 8.8.0, as I noticed this again yesterday!

nathanwoulfe (approved):
Hi @Mantus667, better late than never, yeah? This works really well - can confirm the following:
- urls are displayed correctly for variant content
- urls are displayed correctly for invariant content
- nothing displayed for variants when the component is registered
- nothing displayed for invariant content when the component is registered

PR #9064, opened 2020-10-04
Add umb-icon to umb-tree-item and umb-tree-search-box

MMasey:
This PR updates the `umb-tree-item` and `umb-tree-search-box` views to use the `umb-icon` directive. 

umb-tree-svg-update

PR #9210, opened 2020-10-18
Conditionally show check icon if umb-tree-item is selected

### LINKED ISSUE #9209: Tree view items do not show selected icon when selected
A regression caused by https://github.com/umbraco/Umbraco-CMS/pull/9064 The tree views no longer show the green tick icon when you select an item.

The check icon was previously enabled/disabled by css (line 177 of umb-tree.less) which targeted `<i>`, but the icon now uses `<umb-icon>` so the CSS no longer targets it.

## Umbraco version

I am seeing this issue on Umbraco version: v8/contrib

Reproduction
------------
### Steps to reproduce

- Create any picker data type that uses tree view (MNTP, member group picker etc)
- Select an item

### Expected result

- Icon should change to a tick

Capture

### Actual result

- Icon does not change/no indication item is selected

matthew-darby:
### Prerequisites

Fixes #9209 

### Description

A regression caused by #9064 The tree views no longer show the green tick icon when you select an item.

The check icon was previously enabled/disabled by css (line 177 of `umb-tree.less`) which targeted `<i>`, but the icon now uses `<umb-icon>` so the CSS no longer targets it.

I've changed it so the icon is changed in the view if the node is selected rather than via CSS.

However I'm unsure if:

- This is the correct approach
- The icon needs to be green (it is on some views but isn't on others)

@MMasey @poornimanayar I would appreciate your input on this as I understand you have done alot of work on icons recently - thank you!

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/tree/umb-tree-item.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/tree/umb-tree-item.html
@@ -9,7 +9,7 @@
             <span class="sr-only">Expand child items for {{node.name}}</span>
         </button>
 
-        <umb-icon icon="{{node.icon}}" class="icon umb-tree-icon sprTree" ng-class="node.cssClass" title="{{::node.title}}" ng-click="select(node, $event)" ng-style="::node.style" tabindex="-1"></umb-icon>
+        <umb-icon icon="{{node.selected ? 'icon-check' : node.icon}}" class="icon umb-tree-icon sprTree" ng-class="node.cssClass" title="{{::node.title}}" ng-click="select(node, $event)" ng-style="::node.style" tabindex="-1"></umb-icon>
         <span class="umb-tree-item__annotation"></span>
         <a class="umb-tree-item__label" ng-href="#/{{::node.routePath}}" ng-click="select(node, $event)" title="{{::node.title}}">{{node.name}}</a>
 


nathanwoulfe:
@matthew-darby all works as advertised - selecting items in a tree view changes the icon. 

Re the color, I think the check icon should always be green (contextually, it's most appropriate). That said, they're explicitly set to `#3544b1` in a list view, so maybe for consistency we use that color everywhere. @nielslyngsoe might have an opinion too.

nathanwoulfe (approved):

nielslyngsoe:
@matthew-darby great awareness, and thanks for the PR.

I haven't tested this PR, but based on @nathanwoulfe ´s comment I think it would be good to align the visual look when selecting items. But that's maybe for another PR/task — just in order to be able to move on with this fix.

Actually, this selection-state is something I haven't quite settled just jet.

Generally, the purple color is used for Selections(what is marked) and the Pink color is used for the Active state(Where you are). So with that rule, it makes sense to use the #3544b1 color.
But currently, the selected color has just been used in situations where the selection has no direct impact on the data, but just as a tool-selection, a choice of which elements to perform an action on.
As it had to go for the temporary state look.

This is different here as the selected item is the one stored for reference, therefore I would go for a more impactful look, with the background of the item turned into the #3544b1 color. Text and icons are then turned into white.

image

Let me know what you think.

matthew-darby:
@nielslyngsoe thanks for the detailed feedback!

For this PR, should I change the icon to `#3544b1` and create a separate task to standardise the select state as you suggested?

PR #9211, opened 2020-10-18
NetCore: migrated further unit tests

AndyButland:
Migrated the following three unit test classes:

- ContentTests
- VariationTests
- ImageSharpUrlGeneratorTests (renamed from ImageProcessorUrlGeneratorTest).

And moved some tests into the correct folder/namespace.


src/Umbraco.Tests.Integration/Umbraco/logs/UmbracoTraceLog.DESKTOP-HBIG8BG.20201007_4119.json
```diff
@@ -0,0 +1 @@
+{"@t":"2020-10-07T08:32:38.1888337Z","@mt":"{StartMessage} [Timing {TimingId}]","StartMessage":"Examine shutting down","TimingId":"f8a5ef0","SourceContext":"Umbraco.Web.Search.ExamineComponent","ProcessId":44024,"ProcessName":"testhost.x86","ThreadId":13,"AppDomainId":1,"AppDomainAppId":"1","MachineName":"DESKTOP-HBIG8BG","Log4NetLevel":"INFO "}
```
elit0451, line 1: This file has been mistakenly committed ;)

src/Umbraco.Tests.Integration/Services/ContentTypeServiceTests.cs
```diff
@@ -1200,7 +1200,7 @@ public void Can_Rename_PropertyGroup_With_Inherited_PropertyGroups()
             // create 'page' content type with a 'Content_' group
             var template = TemplateBuilder.CreateTextPageTemplate();
             FileService.SaveTemplate(template);
-            var page = ContentTypeBuilder.CreateSimpleContentType("page", "Page", null, false, "Content_", defaultTemplateId: template.Id);
+            var page = ContentTypeBuilder.CreateSimpleContentType("page", "Page", propertyGroupName: "Content_", defaultTemplateId: template.Id);
```
elit0451, line 1203: I saw that you liked the cleaner version of calling `CreateSimpleContentType()` where you remove redundant params due to being the same as the default values of those parameters. I think this is a good approach. There are several other places **in this file** where we can easily remove the `null` value passed on the method call.

Let me know if you would be up for fixing this small "cosmetic" details 😁

src/Umbraco.Tests.UnitTests/Umbraco.Core/Models/DataTypeTests.cs
```diff
@@ -4,7 +4,7 @@
 using Umbraco.Tests.Common.Builders;
 using Umbraco.Tests.Common.Builders.Extensions;
 
-namespace Umbraco.Tests.UnitTests.Umbraco.Infrastructure.Models
+namespace Umbraco.Tests.UnitTests.Umbraco.Core.Models
```
elit0451, line 7: One would think that due to the similarities with the other Test classes that you successfully moved from <code>Umbraco.Tests.UnitTests.Umbraco.<b>Infrastructure</b>.Models</code> to <code>Umbraco.Tests.UnitTests.Umbraco.<b>Core</b>.Models</code> this file can also be migrated in a similar manner.

However, this is not the case here 🙂 

- Even though `DataType` class has a `Umbraco.Core` namespace, it is located in _Umbraco.Infrastructure_ folder
  - It also has a _Newtonsoft.Json_ using, meaning that we are going to keep it in this location.

At this point, we don't have all the correct namespaces but we should trust the files location

src/Umbraco.Tests.UnitTests/Umbraco.Core/Models/PathValidationTests.cs
```diff
@@ -7,7 +7,7 @@
 using Umbraco.Tests.Common.Builders;
 using Umbraco.Tests.Common.Builders.Extensions;
 
-namespace Umbraco.Tests.UnitTests.Umbraco.Infrastructure.Models
+namespace Umbraco.Tests.UnitTests.Umbraco.Core.Models
```
elit0451, line 10: Same case as with **DataTypeTests.cs**

-  `PathValidationExtensions` class 
   - `Umbraco.Core` namespace
   - located in _Umbraco.Infrastructure_ folder
   - _Microsoft.Extensions.Logging_ using

elit0451 (changes requested):
Great work as always @AndyButland 🎉 

I left a few comments for you to check.

Regarding **DataTypeTests** and **PathValidationTests** I would suggest to keep the tests in <code>Umbraco.Tests.UnitTests.Umbraco.<b>Infrastructure</b>.Models</code> as this will reflect the real location of the underlying model classes. 🙂

src/Umbraco.Tests.Integration/Services/ContentTypeServiceTests.cs
AndyButland: Thanks, for sure - have updated.

src/Umbraco.Tests.UnitTests/Umbraco.Core/Models/DataTypeTests.cs
AndyButland: Good spot - have moved to the right place now.

src/Umbraco.Tests.UnitTests/Umbraco.Core/Models/PathValidationTests.cs
AndyButland: Also moved to the right place.

src/Umbraco.Tests.Integration/Umbraco/logs/UmbracoTraceLog.DESKTOP-HBIG8BG.20201007_4119.json
AndyButland: Have removed.

AndyButland:
Thanks @elit0451 - I've made the suggested amends.

elit0451 (approved):
Great, thanks @AndyButland 

I will wait for the build server and merge afterwards 😉

Issue #9129, opened 2020-10-09
Loss of Block List data after deleting and creating items


HannesWinston:
The `PostSave` event does not include the value of a Block List after the deleting and creating Block List items.

## Umbraco version

Umbraco v8.8.0

Reproduction
------------

### Bug summary

The deletion of an item in a Block List followed by publishing seems to break empty Block Lists on a page. 

After the deletion of a Block List item, publishing and without reloading the page, adding new items to an empty list will result in those items not being saved on the page. But the publishing event is still successful, giving the editor the impression that the content was saved to the page when in in fact - it was not. This is an issue that can result in editors loosing data.

### Specifics

As of now, I have found two ways to produce the error. The first way uses only one Block List property on a page, and the second uses two or more.

Tested in:
- Chrome v85.0.4183.121
- Edge v85.0.564.70
- Firefox v81.0.1

### Steps to reproduce
Using a clean install of Umbraco v8.8.0 with starter kit.
1. Create a new `Block List` property on `Document Type` "Home".
2. Create a new `Element Type` containing a `Textbox` and add it to the newly created `Block List`.
3. Save `Document Type` and go to Content and edit the page "Home".
4. Add one item with any text to the `Block List` property and publish the page.
5. Delete the item and publish the page.
6. Without reloading the browser or going to another page and back to the page "Home", create a new item with any text and publish the page.
7. The page saves successfully, however the `PostSave` event does not send the correct `Block List` value
8. Reload the page.
9. The `Block List` property value is now empty, even though last published version should contain an item.

OR

1. Create two new `Block List` properties on `Document Type` "Home".
2. Create a new `Element Type` containing a `Textbox` and add it to both of the newly created `Block List` properties.
3. Save `Document Type` and go to Content and edit the page "Home".
4. Add two items with any text to the first `Block List` property and publish the page.
5. Delete **one** of the items in the first `Block List`, then create a new item in the second empty `Block List`.
6. Publish the page
7. The page saves successfully, however the `PostSave` event does not send the correct `Block List` value for the second `Block List`.
8. Reload the page.
9. The second `Block List` property value is now empty, even though last published version should contain an item.

### Expected result

The Block List should contain the last published item.

### Actual result

The Block List is empty and the PostSave event sends the following:
`{"id":0,"alias":"blockList","value":""}`

### How to avoid

After deleting an item in a Block List, save the page and reload the browser.

---
[...]

PR #9215, opened 2020-10-19
9129 keep model value reference after empty

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/9129

When server resets the value to a primative value (null or undefined, does not matter) we lose the reference between the BlockEditorModelObject and the Data Model of the Content. This update makes sure when resetting the value its bound on the model and that reference is parsed.

Read test notes on issue.


warrenbuckley:
Code reads fine & tested with the original test notes on issue 9129

PR #9216, opened 2020-10-19
8198 making umb-confirm-action backwards compatible

nielslyngsoe:
This PR takes the PR https://github.com/umbraco/Umbraco-CMS/pull/8198 and makes it backward compatible. As some developers already have implemented this update. So we had to continue with this.

It is tough my recommendation that we stop using this UI and uses the standard dialog for confirmations.


BatJan:
@nielslyngsoe Just gave it a quick look and looks good to me 👍

nielslyngsoe:
@nul800sebastiaan please review this one, thanks :-)

nul800sebastiaan:
FYI: we decided that @BatJan was right all along, there's no way to make this backwards compatible and non-breaking. We'll need to start over on this one! 😅

Issue #8936, opened 2020-09-18
Logic error in ContentController not figuring out default culture correctly


xdjoshuaaz:
There is a logic error in the Umbraco.Web.Editors.ContentController class inside of the PublishCulture method. The issue is on [line 1397](https://github.com/umbraco/Umbraco-CMS/blob/deb03e4d82e48674c6d8bdc7926e820e17454d9e/src/Umbraco.Web/Editors/ContentController.cs#L1397):
```
var valid = persistentContent.PublishCulture(CultureImpact.Explicit(variant.Culture, defaultCulture.InvariantEquals(variant.Culture)));
```

The expression `defaultCulture.InvariantEquals(variant.Culture)` is comparing two different strings, `defaultCulture` is a Culture Name e.g. "English (United States)" and `variant.Culture` is a Culture ISO code e.g. "en-US".

This means this expression always return `false` as shown in the IDE screenshot below:
![image](https://user-images.githubusercontent.com/5199871/93600301-50e90000-f9b7-11ea-8dcf-74bf2af8366d.png)

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.4 and 8.7.0

Reproduction
------------

This issue does not result in any visible bugs out-of-the-box. We've discovered this issue through custom code that declutters content variants that are not applicable to that content.

While Umbraco doesn't support the custom scenario mentioned above out-of-the-box and therefore any bugs that arise from it won't be fixed, the logic error itself is reproducible in any Umbraco installation.

### Specifics
I believe that this issue was caused by a change in 2019. Commit [here](https://github.com/umbraco/Umbraco-CMS/commit/fe2c519d75ff5b8baac14116ac35118a6d28e4ee).

As seen in that commit, the culture Key is no longer used and CultureName is used instead.

### Steps to reproduce

1. Install Umbraco 8 in VS
2. Create a new document type  
    - Needs to vary by culture
    - Needs to be allowed at root
    - Should have at least 1 property, e.g. a true/false property editor
3. Create new content in the site root
4. Add a breakpoint in VS on `ContentController.PublishCulture`
5. Publish the new content, the breakpoint should be hit
6. Step into the foreach-loop and add watches for `variant.Culture` and `defaultCulture`

### Expected result
`variant.Culture` and `defaultCulture` should be similar kinds of values i.e. "en-US" and "en-US"

### Actual result
`variant.Culture` and `defaultCulture` are different kinds of values i.e. "en-US" and "English (United States)"

[...]

PR #9220, opened 2020-10-19
Get default culture comparable to the variant culture

benjaminc:
This fixes issue #8936.  The testing steps are specified there.

The fix was correcting defaultCulture from using the culture's display name to is ISO name.  The one consideration I had to look into was to make sure the one other place default culture was used would handle it correctly.  This was on line 703 of ContentController where it called CultureImpact.GetCultureForInvariantErrors.  However in my debugging, the other values passed in, which would have been used if not equal to the default culture, were also in the ISO name format, not the display name format, and so this change probably fixes another potential bug in that call.

The unit tests all passed locally, and when running through the test scenario from the issue above, I was able to verify that the comparison was between two equivalent strings now.


mikecp (approved):

mikecp:
Hello @benjaminc ,
Thanks for your PR 👍 
That was a well hidden bug there ;-)

PR #9221, opened 2020-10-19
Unique node name unit test fix & refactor

anthonydotnet:
This PR fixes a failing unit test, while also fixing a unit test which was incorrectly passing. 

### 1. TestMany() in SimilarNodeNameTests.cs was failing
Given the names:
- Test (1)
- Test (2)
- Test (1) (1)

When "Test" is supplied to the method, then Test (3) **should** be returned. However, it was returning Test (2), which is obviously a conflict, as it already exists in the list.

Comment in the code was: //fixme - this will yield "Test (2)" which is already in use

### 2. Test() in SimilarNodeNameTests.cs was incorrectly passing
Given the names:
- Kilo
- Kilo (1)

When "Kilo (1)" is supplied to the method, then we would expect "Kilo (2)" to be returned. In this scenario it makes sense to increment the suffix. However, the method was incorrectly returning "Kilo (1) (1)" as a "passing" valid result. 

The result of "Kilo (1) (1)" only makes sense if there is NO existence of Kilo without the suffix.
Given the name:
- Kilo (1)

When "Kilo (1)" is supplied to the method, then we would expect "Kilo (1) (1)". In this scenario, "Kilo (1)" is treated as a single string by itself. Not a string with a suffix. 

### 3. BDD Unit Tests & SimilarNodeName rewrite
To accomplish the above, I implemented BDD style unit tests, and rewrote the unique name generation to be easier to understand.


src/Umbraco.Infrastructure/Persistence/Repositories/Implement/SimilarNodeName.cs
```diff
@@ -1,118 +1,249 @@
...
         }
 
-        // gets a unique name
-        public static string GetUniqueName(IEnumerable<SimilarNodeName> names, int nodeId, string nodeName)
+        internal class StructuredName
         {
-            var uniqueNumber = 1;
-            var uniqueing = false;
-            foreach (var name in names.OrderBy(x => x, new Comparer()))
+            private const string Suffixed_Pattern = @"(.*)\s\((\d+)\)$";
```
bergmania, line 164: Not sure we should use "\s" here.  Why not force it to be only " " (Space)
```suggestion
            private const string Suffixed_Pattern = @"(.*) \((\d+)\)$";
```

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/SimilarNodeName.cs
```diff
@@ -1,118 +1,249 @@
...
-                ? string.Concat(nodeName, " (", uniqueNumber.ToString(), ")")
-                : nodeName;
+            internal bool HasSuffix()
+            {
+                return Number > 0;
+            }
+
+            internal void SetSuffix(int number)
+            {
+                Number = number;
```
bergmania, line 212: We should check that number > 0. Or at least not negative :)
```suggestion
            {
                if(number <= 0) {
                    throw new ArgumentException("Suffix must be a positive number", nameof(number));
                }
                Number = number;
```

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/SimilarNodeName.cs
```diff
@@ -1,118 +1,249 @@
...
-            var uniqueing = false;
-            foreach (var name in names.OrderBy(x => x, new Comparer()))
+            private const string Suffixed_Pattern = @"(.*)\s\((\d+)\)$";
+            internal string Text { get; private set; }
+            internal int Number { get; private set; }
+            public string FullName
+            {
+                get
+                {
+                    string text = string.IsNullOrWhiteSpace(Text) ? Text.Trim() : Text;
```
bergmania, line 171: If `Text` is `null` then Trim throws? Should this not have been the opposite?

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/SimilarNodeName.cs
```diff
@@ -1,118 +1,249 @@
...
+            {
+                Number = number;
+            }
+
+            internal void SetNoSuffix()
+            {
+                Number = 0;
+            }
+
+            internal void SetText(string text)
```
bergmania, line 220: Why not having the property setter as internal?

bergmania (commented):
Hi @anthonydotnet...

Lovely work. The new code is easier to read 💪 

I have a few comments. Let's see if you agree :)

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/SimilarNodeName.cs
anthonydotnet: Agree... we can force a space character.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/SimilarNodeName.cs
anthonydotnet: This is yet another edge case which probably no one has considered. If a user decides to name their node "Test (-1)" then we should allow it. We should assume that "Test (-1)" is a complete string which the user wants to use. 

The suffix detecting regex will just need to be tweaked to only match on positive numbers. No match will default to using the entire string as the text. If you're ok with this, I will make the change.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/SimilarNodeName.cs
anthonydotnet: In this case I should not have used IsNullOrWhiteSpace. Due to the replacement with Space_Character constant, Text will always at least have a value of a space character. It should only need to test for whitepsace. 

So it should be something like this...
var text = Text = Space_Character : Text.Trim() : Text

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/SimilarNodeName.cs
anthonydotnet: TBH I'm on the fence about this one. The only reason I did it was to follow the same pattern as the SetSuffix() and SetNoSuffix(). For these 2 other methods it makes the intention a lot clearer for developers in the future.

I'm splitting hairs right now regarding a simple text setter. What do you want? :)

bergmania: Any reason `Suffix` is not a `uint?` (or `ushort?`) ? Then the property setter could be used for everything. 
- `SetNoSuffix` would be `Suffix = null`
- `SetSuffix` would be `Suffix = 1`
- `HasSuffix` would be `Suffix.HasValue`.

I just this the current code looks more like Java than C# 😄

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/SimilarNodeName.cs
bergmania: `string text = (Text == Space_Character) ? Text.Trim() : Text;` make sense, yes :)

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/SimilarNodeName.cs
anthonydotnet: I can live with that. The only tweak I would make is that the null and 1 should be well named constants.

anthonydotnet:
PR is now updated.

bergmania:
Thank you very much @anthonydotnet 💪

anthonydotnet:
My pleasure. This specific bit of code has been really fun!

Thanks!

PR #9222, opened 2020-10-19
Added support for svgs on the rte prevalue controller

RachBreeze:
Following @MMasey development of `umb-icon`
This PR adds support for svgs on the rte prevalue controller.

Because not all icons in the control currently have an svg, and because we want to support historic values there was a need to continue to support this class styling:

`"mce-ico " + (icon.isCustom ? " mce-i-custom " : " mce-i-") + icon.name`

This means that `rte.prevalues.controller.js` retrieves a list of all svgs to determine which icons have svg equivalents and use them.

My first attempt in `rte.prevalues.controller.js` was pass either the svg name or the class name into `rete.prevalues.html` and  to try and replace `<i ng-if="vm.icon.length" class="{{vm.icon}}" aria-hidden="true"></i>` with `<umb-icon ng-if="vm.icon.length" icon="{{vm.icon}}" aria-hidden="true"></umb-icon>` in `umb-checkbox.thml` however this lost the styling of the icons which don't have an svg equivalent.  I have had to introduce a separate svg property to enable the checkbox to support both `<umb-icon>` moving forward and leave the icon property in umb-checkbox to ensure backwards compatabilty

PR started at CanConHackaton :-)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
@@ -52,7 +52,7 @@
             vm.inputId = vm.inputId || "umb-check_" + String.CreateGuid();
 
             vm.icon = vm.icon || vm.iconClass || null;
-            vm.svg = vm.svg || null;
+      
             // If a labelKey is passed let's update the returned text if it's does not contain an opening square bracket [
             if (vm.labelKey) {
                  localizationService.localize(vm.labelKey).then(function (data) {
@@ -91,7 +91,6 @@
             cssClass: "@?",
             iconClass: "@?", // deprecated
             icon: "@?",
-            svg: "@?",
             disableDirtyCheck: "=?"
         }
     };
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
@@ -52,7 +52,7 @@
             vm.inputId = vm.inputId || "umb-check_" + String.CreateGuid();
 
             vm.icon = vm.icon || vm.iconClass || null;
-
+            vm.svg = vm.svg || null;
             // If a labelKey is passed let's update the returned text if it's does not contain an opening square bracket [
             if (vm.labelKey) {
                  localizationService.localize(vm.labelKey).then(function (data) {
@@ -91,6 +91,7 @@
             cssClass: "@?",
             iconClass: "@?", // deprecated
             icon: "@?",
+            svg: "@?",
             disableDirtyCheck: "=?"
         }
     };
--- a/src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html
@@ -32,7 +32,7 @@
         </span>
     </span>
     <span class="umb-form-check__info" ng-transclude>
-        <i ng-if="vm.icon.length" class="{{vm.icon}}" aria-hidden="true"></i>
+        <umb-icon ng-if="vm.icon.length" icon="{{vm.icon}}" aria-hidden="true"></umb-icon>
         <span ng-if="vm.text.length" class="umb-form-check__text">{{vm.text}}</span>
     </span>
 </label>
--- a/src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html
@@ -32,8 +32,7 @@
         </span>
     </span>
     <span class="umb-form-check__info" ng-transclude>
-        <umb-icon ng-if="vm.svg.length" icon="{{vm.svg}}"> </umb-icon>
-        <i ng-if="vm.icon.length" class="{{vm.icon}}" aria-hidden="true"></i>
+        <umb-icon ng-if="vm.icon.length" icon="{{vm.icon}}"> </umb-icon>
         <span ng-if="vm.text.length" class="umb-form-check__text">{{vm.text}}</span>
     </span>
 </label>
--- a/src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html
@@ -1,29 +1,29 @@
 <label class="checkbox umb-form-check umb-form-check--checkbox {{vm.cssClass}}" ng-class="{ 'umb-form-check--disabled': vm.disabled }">
     <span class="umb-form-check__symbol">
         <input ng-if="vm.disableDirtyCheck"
-            type="checkbox"
-            id="{{vm.inputId}}"
-            name="{{vm.name}}"
-            value="{{vm.value}}"
-            class="umb-form-check__input"
-            val-server-field="{{vm.serverValidationField}}"
-            ng-model="vm.model"
-            ng-disabled="vm.disabled"
-            ng-required="vm.required"
-            ng-change="vm.change()"
-            no-dirty-check />
+               type="checkbox"
+               id="{{vm.inputId}}"
+               name="{{vm.name}}"
+               value="{{vm.value}}"
+               class="umb-form-check__input"
+               val-server-field="{{vm.serverValidationField}}"
+               ng-model="vm.model"
+               ng-disabled="vm.disabled"
+               ng-required="vm.required"
+               ng-change="vm.change()"
+               no-dirty-check />
 
         <input ng-if="!vm.disableDirtyCheck"
-            type="checkbox"
-            id="{{vm.inputId}}"
-            name="{{vm.name}}"
-            value="{{vm.value}}"
-            class="umb-form-check__input"
-            val-server-field="{{vm.serverValidationField}}"
-            ng-model="vm.model"
-            ng-disabled="vm.disabled"
-            ng-required="vm.required"
-            ng-change="vm.change()"/>
+               type="checkbox"
+               id="{{vm.inputId}}"
+               name="{{vm.name}}"
+               value="{{vm.value}}"
+               class="umb-form-check__input"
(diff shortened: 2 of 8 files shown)


bjarnef:
@RachBreeze I think it would be ideal if we only have a single property `icon` and somehow can handle both svg and font icon. Furthermore it would be great if we can avoid having both `<i>` and `<umb-icon>` since the latter one already has fallback using `<i>` element inside the component.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/rte/rte.prevalues.controller.js
```diff
@@ -25,22 +25,34 @@ angular.module("umbraco").controller("Umbraco.PrevalueEditors.RteController",
...
+                    function(obj) {
+                        var icon = getFontIcon(obj.alias);
+                        var iconName = "";
+                        var svgName = "";
+                        var testName = icon.name.startsWith("icon-") ? icon.name : "icon-" + icon.name;
+                        if (data.some(e => e.name === testName)) {
+                            svgName = testName;
+                        } else {
+                            iconName = "mce-ico " + (icon.isCustom ? " mce-i-custom " : " mce-i-") + icon.name;
+                        }
```
bjarnef, line 40: It should be sufficient to set just the `icon` property here.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/rte/rte.prevalues.html
```diff
@@ -7,6 +7,7 @@
                 on-change="selectCommand(cmd)"
                 icon="{{cmd.icon}}"
                 text="{{cmd.name}}"
+                svg="{{cmd.svg}}"
```
bjarnef, line 10: This shouldn't be neccesary.

src/Umbraco.Web.UI.Client/src/views/components/forms/umb-checkbox.html
```diff
@@ -32,6 +32,7 @@
         </span>
     </span>
     <span class="umb-form-check__info" ng-transclude>
+        <umb-icon ng-if="vm.svg.length" icon="{{vm.svg}}"> </umb-icon>
         <i ng-if="vm.icon.length" class="{{vm.icon}}" aria-hidden="true"></i>
         <span ng-if="vm.text.length" class="umb-form-check__text">{{vm.text}}</span>
```
bjarnef, line 37: We should only use `<umb-icon>` here then and remove the old `<i>` element.

RachBreeze:
Hi @bjarnef 
I totally agree it would be great to only use `umb-icon`, and I did hope that my PR explained why I couldn't do this.
Essentially the icon font elements passed into umb-checkbox by the `rte.prevalues.controller.js` don't display on `umb-icon` when they are passed in as an icon font, which is why I had to leave it in.

RachBreeze:
The issue I had was that the `umb-icon` class on the `umb-icon directive` essentially over road the display of the font icons being passed in by the `rte.prevalues.controller.js` and I tried a number of options with formatting on that class which failed to fix the issue.
In all honesty the only thing I didn't try was removing the umb-icon class if an svg wasn't found.

bjarnef:
So is it an issue than the rte icons not use the current `<i>` element inside `<umb-icon>` when passing in the font icon?

I would like we only have `<umb-icon>` here an investigate further if SVG or font icon is not rendered.
```
<umb-icon ng-if="vm.svg.length" icon="{{vm.svg}}"> </umb-icon>
<i ng-if="vm.icon.length" class="{{vm.icon}}" aria-hidden="true"></i>
```

What happens if you set `icon` property in rte.prevalues.controller.js?
Something like this:

```
var icon = getFontIcon(obj.alias);
var iconName = icon.name.startsWith("icon-") ? icon.name : "icon-" + icon.name;
if (data.some(e => e.name === iconName)) {
      icon = iconName;
} else {
     icon = "mce-ico " + (icon.isCustom ? " mce-i-custom " : " mce-i-") + icon.name;
}

var objCmd = Utilities.extend(obj,
{
    fontIcon: icon.name,
    isCustom: icon.isCustom,
    selected: $scope.model.value.toolbar.indexOf(obj.alias) >= 0,
    icon: icon
});

return objCmd;
```

RachBreeze:
In that case the font icons are not displayed on `umb-icon`

bjarnef:
Yeah, I just think we should find the right solution as the component otherwise could be cluttered ☺
Using `<umb-icon>` also lazy loaded the icons, so it some cases it is shown up with some delay.

I might have a look later today.

RachBreeze:
Hi @bjarnef I have found an issue in `umb-button` where the name of the icon is also passed into the class and tied this up.  Having both the icon name in `umb-button` and in the class resulted in the icon being displayed twice when it was resolved to an icon font

bjarnef:
@RachBreeze okay, so I have tested this with latest changes in `v8/contrib` branch and it does kinda seems to work with `<umb-icon>`.

image

In `<umb-checkbox>` (and in `<umb-radiobutton>`) replace the following:

```
<i ng-if="vm.icon.length" class="{{vm.icon}}" aria-hidden="true"></i>
```

With
```
<umb-icon ng-if="vm.icon.length" icon="{{vm.icon}}"></umb-icon>
```

and in rte.prevalues.controller.js:

```
tinyMceService.configuration().then(function(config){
    $scope.tinyMceConfig = config;

    iconHelper.getAllIcons().then(data => {

        // extend commands with properties for font-icon and if it is a custom command
        $scope.tinyMceConfig.commands = _.map($scope.tinyMceConfig.commands, function (obj) {

            var fontIcon = getFontIcon(obj.alias);
            var isCustom = fontIcon.isCustom;

            var iconName = fontIcon.name.startsWith("icon-") ? fontIcon.name : "icon-" + fontIcon.name;

            var icon = data.some(e => e.name === iconName)
                ? iconName
                : "mce-ico " + (isCustom ? "mce-i-custom " : "mce-i-") + fontIcon.name;

            var objCmd = Utilities.extend(obj, {
                isCustom: isCustom,
                selected: $scope.model.value.toolbar.indexOf(obj.alias) >= 0,
                icon: icon
            });

            return objCmd;
        });

    });
});
```

The reason some icons doesn't show is because we inherit font family for `<i>` element inside `<umb-icon>` component in order to make it backwards compatible.

image

We have a few options here:

- Specific font family for `mce-ico` class:

```
.umb-icon>i.mce-ico {
    font-family: 'tinymce',Arial;
}
```

- However we probably prefer this instead using `:not()` css selector:

```
.umb-icon>i:not(.mce-ico) {
    font-family: inherit;
}
```
This part can be found here: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/components/umb-icon.less#L29-L31

I also noticed we have this styling. It is probably because of the toolbar in TinyMCE. We could play with the font size here, so it keep using 12px in TinyMCE toolbar for now but inside `.umb-icon` or `.rte-editor-preval` maybe use default (16px) or 14px?

image

RachBreeze:
it was the issue with the duplicate class and icon names that was throwing me a bit as it was causing the icons to display like this when the `umb-icon` class was removed from the `umb-icon` directive

image

There are a lot of umb-icon entries with the unneeded class

bjarnef:
@RachBreeze okay, I don't see that issue with latest changes in `v8/contrib` branch.

image

But I guess this isn't specific for this PR regarding the rte prevalue icons?

RachBreeze:
Hi @bjarnef 
Thank you for looking at this it was exactly the way I had seen the PR working originally!

In all honesty I hadn't thought to check the CSS so from my perspective that's an ace spot thank you.
I think 14 px is probably in line with the current sizing.  And in terms of the styling I do prefer:

```
.umb-icon>i:not(.mce-ico) {
    font-family: inherit;
}

```
I'll look at the duplicate icon and class names in a separate PR, as they really shouldn't be needed  :-)

Thank you for your time

RachBreeze:
In order to reproduce the duplicate display if you remove the class `umb-icon` from the `umb-icon` directive it can be reproduced, so in effect the class is hiding the issue

bjarnef:
@RachBreeze you're welcome 😊
you might also want to look at the updates in the controller here:
https://github.com/umbraco/Umbraco-CMS/pull/9222#issuecomment-713099039

Just cleaned up some of the code and we didn't really used the `fontIcon` property in `objCmd` since we just use the `icon` property. The view is much simpler than it used to be 😁😋

src/Umbraco.Web.UI.Client/src/views/propertyeditors/rte/rte.prevalues.controller.js
```diff
@@ -29,22 +29,21 @@ angular.module("umbraco").controller("Umbraco.PrevalueEditors.RteController",
...
+                            : "mce-ico " + (isCustom ? "mce-i-custom " : "mce-i-") + fontIcon.name;
+
                         var objCmd = Utilities.extend(obj,
                             {
                                 fontIcon: icon.name,
                                 isCustom: icon.isCustom,
                                 selected: $scope.model.value.toolbar.indexOf(obj.alias) >= 0,
-                                icon: iconName,
-                                svg: svgName
+                                icon: icon
```
bjarnef, line 46: @RachBreeze you can probably also remove the `fontIcon` property here since we don't use this and `icon` variable is now a string, so `isCustom` property should use `isCustom` variable defined above.

RachBreeze:
Hi @bjarnef that's a great shout thank you, I've tidied up the code and removed the unused variables

src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js
```diff
@@ -91,6 +91,7 @@
             cssClass: "@?",
             iconClass: "@?", // deprecated
             icon: "@?",
+            svg: "@?",
```
bjarnef, line 94: @RachBreeze it seems you forgot to remove `svg` added in the component here.

RachBreeze: Oops!  Sorry on it

RachBreeze:
Hi this PR has been open a while now is there anything I can do to help?

PR #9227, opened 2020-10-20
Netcore: Migrate rendering related unit tests

nikolajlauridsen:
I migrated the tests in RenderModelBinderTests and UmbracoViewPageTests.

When migrating RenderModelBinderTests, I found an issue in the ContentModelBinder, if types already matched we just stopped it with `Task.CompletedTask` this caused the binding to be marked as failed, whereas previously we just set the result to be source model, I corrected this to behave as before and adjusted one test in ContentModelBinderTests to reflect this behaviour.

When migrating UmbracoViewPageTests I ran into a lot more trouble. The tests seem to test `SetViewData` in `UmbracoViewPage` however this method does not exist on `RazorPage` and therefore `UmbracoViewPage`. there were two primary problems that showed up frequently, and I think it's the easiest to try and explain them with an example.

Take this test from the old tests
```c#
        [Test]
        public void RenderModelOf_ContentType1_To_ContentType1()
        {
            var content = new ContentType1(null);
            var model = new ContentModel<ContentType1>(content);
            var view = new ContentType1TestPage();
            var viewData = new ViewDataDictionary(model);

            view.ViewContext = GetViewContext();
            view.SetViewDataX(viewData);

            Assert.IsInstanceOf<ContentType1>(view.Model);
        }
```

When migrating it would look something like this 

```c#
        [Test]
        public void RenderModelOf_ContentType1_To_ContentType1()
        {
            var content = new ContentType1(null);
            var model = new ContentModel<ContentType1>(content);
            var view = new ContentType1TestPage();
            var viewData = GetViewDataDictionary<ContentModel<ContentType1>>(model);
        
            view.ViewData = viewData;
        
            Assert.IsInstanceOf<ContentType1>(view.Model);
        }
```

However this gives us a compile time error because it's strongly typed so we can't set `view.ViewData` from `ContentModel<ContentType1>` because the page expects `ContentType1` and not `ContentModel<ContentType1>`

And we can't create `var viewData` with `ContentType1` from the `ContentModel<ContentType1>` model object because `ViewDataDictionary` is strongly typed as well and the content model it doesn't actually inherit `ContentType1`

The last option would be to rewrite it a bit more, but then it would be identical to `RenderModel_ContentType1_To_ContentType1`

The other issue is related as well and affects all the tests that expect an exception like 

```c#
        [Test]
        public void RenderModelOf_ContentType1_To_ContentType2()
        {
            var content = new ContentType1(null);
            var model = new ContentModel<ContentType1>(content);
            var view = new ContentType2TestPage();
            var viewData = new ViewDataDictionary(model);
        
            view.ViewContext = GetViewContext();
            Assert.Throws<ModelBindingException>(() => view.SetViewDataX(viewData));
        }
```

The problem is again that everything is strongly typed so we're not able to try and set the viewData with the wrong type, we'll just get a compile-time error. 

As a positive though, the tests don't actually need the huge `GetViewContext` helper method because these tests are not affected and do not use the viewcontext

I left the tests I couldn't migrated as commented out and didn't delete them since I figured it might be easier to follow my thoughts that way.


bergmania (approved):
I fixed the tests.. Please look at my commit. One thing that I needed to do in another way, compared to v8, is if the ViewDataDictionary is not generic, then we need to use a new one, as it has to be generic in netcore.

PR #9230, opened 2020-10-20
Update umb-button directive parameter usage to reflect accurate imple…

lssweatherhead:
### Prerequisites

### Description

Whilst working on the umbSense tags for VS Code, using the API docs as a reference, it appears to me that the usage parameters for the umb-button directive are not totally accurate: https://our.umbraco.com/apidocs/v8/ui/#/api/umbraco.directives.directive:umbButton

I have updated them (along with some typos that were causing ambiguity in the docs) to reflect that properties like `isExpanded`, `hasPopup`, `addEllipsis` etc should really be expected as boolean inputs in the directive. If this is not the case/not consistent then I am happy to revert those changes - just let me know!

Thanks,
Laura


nathanwoulfe (approved):
Thanks @lssweatherhead!

PR #9233, opened 2020-10-21
Fix OutOfMemory exception being thrown during VariantsMigration from v7 to V8

nzdev:
### Prerequisites

### Description
VariantsMigration throws OutOfMemory when migrating a large number of propertyData records.
Updated to run query on database to avoid outofmemory exception

How to test?
Migrate to V8 with a very large number of cmsPropertyData records with large text column values


nzdev:
I've used this to migrate sites successfully.

PR #9234, opened 2020-10-21
Change to MemberHasAccess to return true for unprotected paths.

MrJackWilson:
### Prerequisites

### Description

The method as it currently stands will only return true for a path that is protected _and_ the member has access. However, it should return true for paths which are also unprotected.

To test, call the method passing in paths of nodes which are unprotected. It should return false for all of these paths.

There are 3 possible scenarios that this method should deal with:
1. Node is unprotected
2. Node is protected and member does not have access
3. Node is protected and member does have access

At the moment the code only satisfies options 2 and 3.

### DIFF
--- a/src/Umbraco.Web/Security/MembershipHelper.cs
+++ b/src/Umbraco.Web/Security/MembershipHelper.cs
@@ -119,7 +119,7 @@ namespace Umbraco.Web.Security
             {
                 pathsWithAccess.TryGetValue(path, out var hasAccess);
                 // if it's not found it's false anyways
-                result[path] = hasAccess;
+                result[path] = !pathsWithProtection.Contains(path) || hasAccess;
             }
             return result;
         }


nathanwoulfe:
Hi @MrJackWilson, thanks for taking the time to contribute, much appreciated. Your thinking here makes sense, I'll test that it works as expected, get it merged, then @nul800sebastiaan can hook you up with a fancy new contributor badge for your our.umbraco profile (assuming of course you have one - if you don't, go get one 😄)

nathanwoulfe (commented):
I'm not seeing this method hit at all - the access check for username restriction is happening in `PublicAccessServiceExtensions`, not `MembershipHelper`. @MrJackWilson can you confirm that you're able to step through this method?

MrJackWilson:
Hi @nathanwoulfe, I can confirm that I am able to step through the MemberHasAccess(IEnumerable<string> paths) method on the MembershipHelper.

nathanwoulfe:
Hi @MrJackWilson, all good - I was expecting MemberHasAccess to be called as part of the internal access check, didn't realise it needed to be called explicitly from the helper.

Works nicely, especially when I use the id-based path, not the URL path as the parameter. My bad.

Will merge.

nathanwoulfe (approved):

MrJackWilson:
Hi @nathanwoulfe, no problem, glad to know it works as it should.

Here's my profile: https://our.umbraco.com/members/MrJackWilson/. Thanks!

PR #9236, opened 2020-10-21
Update NestedContentPropertyEditor.cs to fix 2 bugs

### LINKED ISSUE #9235: V8 Bug, If a property where nested content is used has a value that isn't json... the content editor breaks...
Say you are working on a site, you first have a multi media picker, that stored the media ids... then switch this to a nested content since you need more then just an image...

Data stored is media ids but when changing to nested content it expects json... tries to DeserializeObject but fails... (leaving the editor with a error message and an non working content editor)...

 var rowValues = JsonConvert.DeserializeObject<List<NestedContentRowValue>>(propertyValue.ToString());

### LINKED ISSUE #9237: V8 Bug, uncaught null reference expection in nested content
When you have a nested content item that doesn't have any properties... it breaks nested content..leaving the editor with a broken content editor

TimGeyssens:
Make sure when trying to JsonConvert.DeserializeObject it is json...

Fixes #9235 
Fixes #9237

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
@@ -363,7 +363,10 @@ namespace Umbraco.Web.PropertyEditors
             {
                 if (propertyValue == null || string.IsNullOrWhiteSpace(propertyValue.ToString()))
                     return new List<NestedContentRowValue>();
-
+                
+                 if (!propertyValue.ToString().DetectIsJson())
+                    return new List<NestedContentRowValue>();
+                    
                 var rowValues = JsonConvert.DeserializeObject<List<NestedContentRowValue>>(propertyValue.ToString());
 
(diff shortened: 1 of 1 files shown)


nathanwoulfe (approved):
@TimGeyssens this works as expected (no more errors). However, because I was curious, tried changing in the other direction (from NC back to a media picker), that works fine when the original media picker value hasn't changed to NC, but if we set an NC value then change the type back to a picker, we get an error (which is expected, and happens for the same reason that the original NC=>picker change was broken).

Probably doesn't need to be fixed here, but might be something to look at for all editors to see if there's scope to improve managing data when changing types. Could be messy though...

TimGeyssens:
Good thinking! Maybe a try parse option per data editor? To make sure it doesn't kill the backend...

nathanwoulfe:
@TimGeyssens probably part of a larger conversation, and shouldn't let that hold up merging these changes. Will get back around to it 😄

Issue #8963, opened 2020-09-25
Mandatory fields in nested content not marked as required in 8.7.0


jesperweber:
After upgrading from 8.6.4 to 8.7.0 the red asterisk is now missing for mandatory fields in nested content.

## Umbraco version

I am seeing this issue on Umbraco version: 8.7.0

Reproduction
------------

1. Installed a clean Umbraco 8.6.4
2. Created an element doc type with one mandatory textstring field
3. Created a nested content data type and allowed the element type I just created
4. Created a page doc type with the following fields
   1. A textstring field marked as mandatory 
   2. A field for the nested content data type created in step 3
5. Created a page and got the following (as expected)
![image](https://user-images.githubusercontent.com/2575817/94272338-37116500-ff43-11ea-8527-bc2245a320fa.png)

6. Upgrade Umbraco from 8.6.4 to 8.7.0 from NuGet
7. Now the red asterisk in on the nested content field is missing, but still there for the page field
![image](https://user-images.githubusercontent.com/2575817/94272515-72139880-ff43-11ea-8f4b-873e81248b80.png)

 

Also it doesn't tell which field in the nested element that's failing the validation
![image](https://user-images.githubusercontent.com/2575817/94428314-cd869600-0190-11eb-8c93-6f3d8501d06d.png)

---
_This item has been added to our backlog [AB#8782](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/8782)_

limpa123:
Getting this one as well. Setting up new doctypes and datatype in the updated version gives the same result. 
Adding a value to the required field once the validation message has been shown does not get rid of the message. The only way to save is to remove the nested content item and then add it again filling out the value from start.

jesperweber:
@limpa123 Yes that's also what I experience.

filefontaine:
I found that in my 8.5.5 prod website, the red * is not set either so i think is a regression of 8.6. 

But the error appear correctly :
![image](https://user-images.githubusercontent.com/22192319/95058114-5a80a080-06f7-11eb-94b7-b8c6edc82ae2.png)

(**Current Navigation Title** is mandatory)

nul800sebastiaan:
First of your nested validation in 8.7+ should look a bit better, like this: 

![image](https://user-images.githubusercontent.com/304656/95102777-1790ee00-0734-11eb-9ef3-6f88e645f1b8.png)

[...]

PR #9249, opened 2020-10-22
Fix of highlight mandatory properties of NC

nielslyngsoe:
fixes https://github.com/umbraco/Umbraco-CMS/issues/8963

Makes the star(*) appear on mandatory fields of NC, as previously.

image


nielslyngsoe:
WUPS, stupid GH UI begin a bit slow.

PR #9252, opened 2020-10-22
NetCore: Email message abstraction

elit0451:
Abstracting `MailMessage`, so we only expose our own types.

## Test
- Configure SMTP settings (uncomment them in _appsettings.Development.json_)
- Test the email sending functionality (it should be the same as before)

### DIFF
--- a/src/Umbraco.Core/Models/EmailMessage.cs
+++ b/src/Umbraco.Core/Models/EmailMessage.cs
@@ -8,19 +8,28 @@ namespace Umbraco.Core.Models
         public string To { get; }
         public string Subject { get; }
         public string Body { get; }
-        public bool IsBodyHtml { get; set; } = false;
+        public bool IsBodyHtml { get; }
 
-        public EmailMessage(string from, string to, string subject, string body)
+        public EmailMessage(string from, string to, string subject, string body, bool isBodyHtml)
         {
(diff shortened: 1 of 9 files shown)


src/Umbraco.Core/Models/EmailMessage.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Umbraco.Core.Models
+{
+    public class EmailMessage
+    {
+        public string From { get; }
+        public string To { get; }
+        public string Subject { get; }
+        public string Body { get; }
+        public bool IsBodyHtml { get; set; } = false;
```
bergmania, line 11: From what  I can see `IsBodyHtml ` could as well be part of the ctor.

src/Umbraco.Core/Models/EmailMessage.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public EmailMessage(string from, string to, string subject, string body)
+        {
+            if (string.IsNullOrEmpty(from)) throw new ArgumentException("Value cannot be null or empty.", nameof(from));
```
bergmania, line 15: I think these should be separated into two ifs for each

bergmania (approved):
Only minor things.. I will fix up

Issue #9097, opened 2020-10-06
Add a contextual password helper


TimGeyssens:
When you set or reset your pw... the pw entered has to meet a certain criteria... think it is just checking the number of chars.

It is a bit frustration that you can only see this is required (on pw reset) after you set a pw with not enough chars...

A nice touch would be to add  a contextual password helper

Like the one outlined here: https://dev.to/lisasy/what-does-it-take-to-create-a-strong-password-5ff3

nul800sebastiaan:
Yes, I think it would be great to add a help text indicating what the minimal expected password length is and even see if we can disable the save button until that length is reached. It would be good if it was a live counter actually, that shows how many more characters are needed (you've typed 5, the min length is 12 so it will indicate: you need to add 7 more characters).

Password indicators are notoriously bad at indicating if a password is actually good: https://www.troyhunt.com/password-strength-indicators-help-people-make-dumb-choices and we don't think that should be part of the CMS. 

Bonus: optionally, check the password using the Pwnd Passwords API: https://haveibeenpwned.com/Passwords

umbrabot:
Hi @TimGeyssens,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

TimGeyssens:
Yes agree about the idicator, just lenght would be ace! Good idea with  Pwnd Passwords API

dawoe:
Should this be used in the install screen as well ?

TimGeyssens:
I would say yes, everywhere a pw needs to be set that has a policy... to avoid frustrations

TimGeyssens:
Some AngularJs code for inspiration http://embed.plnkr.co/UJ8mRWJ6EL94i64h5fah/

mikecp:
I have made a first PR for this. Next steps: see comments in PR :-)

TimGeyssens:
Woot!

mikecp:
So now working on reset password screen as well.

@nul800sebastiaan What is your idea/intention about the Pwnd Password check? Display a warning? Refuse the password?

[...]

PR #9256, opened 2020-10-23
9097 add contextual password helper

mikecp:
### Prerequisites

This PR is a first fix for issue   #9097

### Description
I have added a "password tip" text mentioning the minimum number of characters remaining to type, as well as a reminder of the number of non-alphanumeric characters required (cf screenshot)  

password-tip

I have been able to do this for the password changes in the backoffice: members, users, and own profile.

I have not been able to do this on the password change on the login form (this is the one in the "forgot password" mails) because I need to call the following method to get the info, and you need to be authenticated to have access to that (got 401 error when calling this from login page)

```
                authResource.getMembershipProviderConfig().then(function(data) {
                    vm.invitedUserPasswordModel.passwordPolicies = data;
                });

```

If somebody can point me on how to get that info for that screen, that would be great :-)

### Steps to test: 

- adapt the values of the  minRequiredNonalphanumericCharacters and minRequiredPasswordLength attributes on the MembershipProviders registration in the web.config file (both UmbracoMembershipProvider and UsersMembershipProvider to test all sections)
- go to any of the "change password" sections/screesn in the backoffice

### Next steps; 

- try to get this working in the loin screen for the "forgot password emails"
- integrate the Pwnd Passwords API as bonus, as suggested in the issue


protherj:
Hey @mikecp!  We may take this PR and try to help move the needle on it during our Hackathon.  Sound OK?

mikecp:
Hey @protherj Sure no problem. I was planning on having a look at it tonight as well, so maybe just drop a comment here to let me know if you start working on it, then I'll keep my hands off it ;-)

protherj:
You could join our Zoom :)

mikecp:
In fact I could, I'll try to drop by :-)

mikecp:
And now with display on the password reset screen as well ;-)

password-tip2

protherj:
Hey, we tried to convert the validation to a directive, but ran out of time as it was more complex than we expected.  Sorry!

mikecp:
Hey @protherj, no problem ;-) 
I thought about that as well, but indeed more complex than I thought. Currently copy-pasted, so I'll give another try at directive if I find time :-)

TimGeyssens:
any update on merge @nul800sebastiaan ?

mikecp:
Haven't done the check with Pwnd Passwords API yet but I guess this can be taken over later on?

TimGeyssens:
is copy paste ok? Or does it need to be directive?

TimGeyssens:
is copy paste ok? Or does it need to be directive?

TimGeyssens:
@mikecp yeah would agree, merge now and then it can be added later

mikecp:
@TimGeyssens Indeed copy-pasted, i will have a look to move to directive.

suedeapple:
Hmmm.. I'm not too sure if it's safe to disclose how long a password needs to be on the "Public" side of the site.  It opens itself to a hacker to determine the min length of other users passwords.

TimGeyssens:
It's for pw change and set ... Not when logging in

TimGeyssens:
@nathanwoulfe how about this one? Simple one to avoid some frustration...

nathanwoulfe:
@mikecp @TimGeyssens @protherj this looks great - I'm going to have a look at creating a directive to avoid the duplication. Otherwise, this is a valuable UX addition

mikecp:
Thanks @nathanwoulfe , much appreciated.
I kept that directive stuff on my todo list but somehow did not find the time to actually do it yet...
Would indeed be better to avoid code duplication!

nathanwoulfe:
@mikecp directive is in my last commit - I've implemented it the user password change overlay only, but if we're happy with the approach it's easy enough to update the other locations.

Only bit that's potentially a bit messy is having to pass a reference to the password input element, rather than a scoped variable - this is because the password fields have a min-length attribute set, and potentially some logic implemented too, which prevents the scoped variable from being set until the min-length is met. That then means the helper text isn't displayed until the password is n characters long, which defeats the purpose (I'm guessing that's why you have the key-up event?). By passing a selector for the password input, the directive can then watch the value, and show the helper as needed.

mikecp:
Hey @nathanwoulfe ,
I was just having a look at it, it looks great! Thanks! I would probably never have thought of that "element-selector-attribute" to get the password textbox through 👍 

Indeed, the "key-up" event was my workaround for that RawValue not being passed until the password had actually met the criteria's 😂
And I don't see how we could go without some level of "messiness" since we cannot really change the behavior of that password textbox...

So for me it is good to go with the directive as you wrote it.
I can change the rest myself, but then tomorrow because it's getting late for me here 😴 But feel free to do it yourself if you feel like it.
Cheers!

mikecp:
@nathanwoulfe still needs some cleaning up, I'll finish tonight 😀

nathanwoulfe:
Legendary, thanks @mikecp. I had intended coming back to finish up, but Monday got away from me

I'm not sure that the passing the element selector as the directive binding is necessarily the best idea, but it works, and saves the repeated code.

mikecp:
Hey @nathanwoulfe ,
I think this time we're there 😅
There was one major issue that I identified by testing, due to the fact that we have separate Members and Users Manager, which can each have their own settings for min length and non alpha-num.

This was handled "behind the scenes" in my "duplicate code" version because the appropriate values were passed during the binding to the component, but with the separate directive, we always used the Users settings via the two constants, and therefore this could display misleading when changing a Member password.

Anyway, there are a bunch of changes that I made so that specific min length and non alpha-num values can now be bound to the password tip component.
Also, doing these changes brought me to an alternative solution for the "watcher", by using the "onchanges" event.
Yes, the ng-keyup is back on the password inputs, so if you think that the "watcher" remains better, we can revert to it 😉 

Cheers!

nathanwoulfe (approved):
@mikecp looks good to me - there's a tradeoff with the keyup vs watch, where the former is potentially more efficient and avoids the angularjs coupling, but it does mean the component is more involved since it needs to manage the keyup. The way I'd done the watch though was a bit dodgy as it meant the directive was operating outside it's own scope, so probably best to not be doing that...

Otherwise, all works nicely, it's a really useful addition 👍

mikecp:
Thx @nathanwoulfe !
As you said, none of the options was optimal, but in the end I thought it made a little more sense to let the component be responsible to send its "state change" to the directive.

Cheers!

PR #9257, opened 2020-10-23
Merges netcore/dev (from merge_umbraco_id_into_netcore_dev) into the 6973-migrating-authenticationcontroller branch

Shazwazza:
Merges netcore/dev (from merge_umbraco_id_into_netcore_dev) into the 6973-migrating-authenticationcontroller branch

Basically migrating the functionality from 8.9 into netcore. The project builds, i haven't had time to see if all the tests are passing. But I can install and login, etc... 

This does not implement any of the OAuth stuff for real, it's just a means of getting it all merged. When I'm back from holidays I need to take some time and implement all of this and then test. There's still a bunch more to do to get all the functionality that we used to have in there working.


src/Umbraco.Core/Configuration/Models/GlobalSettings.cs
```diff
@@ -46,13 +49,17 @@ internal const string
...
+        public DatabaseServerMessengerSettings DatabaseServerMessenger { get; set; } = new DatabaseServerMessengerSettings();
+
         public string RegisterType { get; set; } = string.Empty;
 
         public string DatabaseFactoryServerVersion { get; set; } = string.Empty;
 
         public string MainDomLock { get; set; } = string.Empty;
 
-        public string NoNodesViewPath { get; set; } = "~/config/splashes/NoNodes.cshtml";
+        public string NoNodesViewPath { get; set; } = "~/umbraco/UmbracoWebsite/NoNodes.cshtml";
```
Shazwazza, line 62: @bergmania we will need to fix these hard coded ~/umbraco paths, they cannot be hard coded like this since the umbraco path comes from the UmbracoPath value. We can change later (not in this PR)

PR #9260, opened 2020-10-23
Enable auto-focus for confirm button in umb-confirm

### LINKED ISSUE #8945: Put focus on Ok button for Tree Delete Dialog
￼image

This "Ok" button should receive focus when the dialog opens, so the user can press enter immediately.

It would also be nice to rename the button to "Delete", this makes it more clear for the user what the action is.

daniel-shuy:
### Prerequisites

If there's an existing issue for this PR then this fixes #8945

### Description

This PR enables `auto-focus` for confirm button (`OK` button) in `umb-confirm`.

This mainly enables autofocus for the contextual menu delete (see #8945), but also the Empty Recycle Bin confirmation.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/umb-confirm.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/umb-confirm.html
@@ -15,7 +15,8 @@
                         button-style="{{confirmButtonStyle || 'primary'}}"
                         state="confirmButtonState"
                         disabled="confirmDisabled === true"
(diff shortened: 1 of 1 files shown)


nathanwoulfe (approved):

nathanwoulfe:
Hi @daniel-shuy this is all good, works perfectly. Changing the button text is a bigger job, since `general_ok` is the default key, so would mean updating the individual instances of `umb-confirm` to set the desired text. Still a good idea though, it's not ideal to have OK in a red button, sends conflicting messages...

Will merge this change though, text can be considered separately.

Issue #9172, opened 2020-10-15
change data type for delayMilliseconds and periodMilliseconds from int to long


deekoulos:
Hey y'all,

working with the Umbraco background tasks I recognized that there are 2 important parameters, to schedule the first run and the reccuring cycle:

1. delayMilliseconds 
2. periodMilliseconds 

Both are measured in milliseconds and there are ways to use these parameters to have exact schedules, for running them initially and afterwards on certain dates and times. 

![image](https://user-images.githubusercontent.com/55948134/96122716-35a7dc80-0ef2-11eb-8f0d-9489bd08735e.png)

However, since these parameters are int values, containing milliseconds, they are a bit limited. 

I hope I am doing the math right --> Int32 Max Value is 2.147.483.647

Which means the first delay as well as the recurring cycle cannot be later than ~25 days in the future:

2.147.483.647 / 1000 = 2.147.483 seconds
2.147.483 / 3600 = ~596 hours
596 / 24 = ~25 days

I think it would be better and safer to change the datatype to long to provide more flexibility in scheduling tasks. 

In our case, we need some tasks to be scheduled once in a month.

Thanks
Dee

nul800sebastiaan:
Hint: we have women working on CMS too, I've changed your first sentence! 🙂

deekoulos:
just for the record @nul800sebastiaan , guy is male specific but guys is gender neutral :D 

https://www.dictionary.com/browse/guy?s=t

nul800sebastiaan:
Regardless of a dictionary definition (I wish I could send a PR to dictionary.com!), it would be nice to use inclusive language as most people who don't identify as a guy definitely don't feel included when you address them with "guys". It might seem like a small thing, but it really helps. Thanks!

deekoulos:
alright, check

deekoulos:
Hey @nul800sebastiaan,

what is the current status of this ticket? Will this be implemented? We need to run tasks every months which will exceed the capacity of the integer values.

Best
Dee

nul800sebastiaan:
Whoops. sorry, forgot to follow up on this one. Happy for some help here if you want to make this happen. 👍

umbrabot:
Hi @deekoulos,

[...]

PR #9263, opened 2020-10-23
Add overloaded method for RecurringTaskBase

MaikelVeen:
As discussed in issue #9172 it is useful for recurring task to be able to repeat them longer than int milliseconds allows. This pull request adds an overloaded method for the constructor and changes the underlying property. This should be safe since Timer has a overloaded constructor aswell that accepts a long.


mikecp (approved):

mikecp:
Thanks for this PR @MaikelVeen . It does the work indeed 👍 
And since it is your first PR here, @nul800sebastiaan will have the pleasure to assign you your contributor badge on Our 🥇

PR #9264, opened 2020-10-23
[Issue 5277-146] accessibility - Close 'X' icon next to language drop…

justinsane98:
From the main accessibility task 5277:
https://github.com/umbraco/Umbraco-CMS/issues/5277

subtask 146:
Close 'X' icon next to language drop down is identified as "link" - screen reader


glombek:
Hi Justin,

Thanks for the pull request, I'm just taking a look at it with the team.

For reference, the button that this PR changes is the close button which appears next to the name and language select dropdown when a variant is edited in split view:
image

Joe

glombek:
Hi @justinsane98 ,

Thanks again for your PR.

I was just wondering whether "Close window" was the most appropriate label for this - it feels like more of a pane to me? Would the simple "Close" label suit better? Or should we have a "Close variant pane" (or similar) item added to the dictionary?

I don't think this is a blocker, but would like to hear your thoughts on the matter!

Thanks,

Joe

justinsane98:
I admit I was trying to use an existing key and I was looking perhaps for something more specific than just "close" because it really could be closing any number of things. I think if you go with something as specific as "Variant Pane" then you need to have "X Pane", "Y Pane", etc and I could see that getting cumbersome to manage. How about something in the middle like "Close Pane"?

nathanwoulfe:
@justinsane98 I'd go with `Close pane` - it's descriptive enough without being too specific to prevent reusing the key. If you could add the new key and update the PR, we can merge for 8.12.

mikecp (approved):

mikecp:
Hello @justinsane98 ,
I took the liberty to create the missing key so that we could merge your PR, I hope you won't mind. Thanks for your contribution 👍

PR #9265, opened 2020-10-23
[Issue 5277-145] accessibility - Language drop downs do not advise as drop…

justinsane98:
From the main accessibility task 5277:
https://github.com/umbraco/Umbraco-CMS/issues/5277

subtask 145:
Language drop downs do not advise as drop downs - screen readers

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-content-header.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/editor/umb-editor-content-header.html
@@ -40,7 +40,7 @@
                                maxlength="255" />
                     </ng-form>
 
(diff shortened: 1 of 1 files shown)


nathanwoulfe (changes requested):
Hi @justinsane98 given the button is hidden by the ng-if directive when there are no variants (or there are variants and hideChangeVariant is not true), could we set aria-haspopup to true and avoid the additional property binding? Doesn't hurt to be explicit when we can.

nathanwoulfe (approved):
Thanks @justinsane98, will get this merged.

nul800sebastiaan:
Hi again @justinsane98! We're wrapping up the Hacktoberfest rewards list and haven't heard from you yet, make sure to fill in the form at https://docs.google.com/forms/d/e/1FAIpQLSdvfuqHFn7ib_xxMbVZ3nuyUgUTUCGPwmw0y83kiY4lP3OzKQ/viewform or if you want us to plant a tree on your behalf then you can reply as well and we'll take care of it. Thanks! 👍

Issue #8539, opened 2020-07-31
V8 Image Cropper doesn't save image if property alias contains underscore


ReVoid:
If `Image Cropper` property editor alias contains an underscore char `_` (possible, other rarely used chars too), the editor doesn't save the image in backoffice, even in **`Media` folder**.

![error](https://user-images.githubusercontent.com/12345710/89034148-0fd25900-d341-11ea-9257-d069f0e45405.gif)

## Umbraco version

8.6.3 clean install

Reproduction
------------

### Steps to reproduce

1. Create custom property editor of `Image Cropper`
2. Create a property of that type and set alias containing underscore char `_`
3. Try to upload an image, save and publish the document
4. Go away from the document, and return to it again, the image is missing now
5. Go to your property editor from step 1, and remove the underscore from it alias
6. Return to the document and try to upload the image again, everything works fine now

### Expected result

An Image saved

### Actual result

An image was not saved :(

nul800sebastiaan:
I can reproduce this. It looks like it gets stored just fine but when adding the image to the view something is going wrong, in my case it was missing the `/media/wualp1sx/` part from the image `src` attribute. 

So, when not using an underscore in the property alias:

![image](https://user-images.githubusercontent.com/304656/89180903-fb859a80-d592-11ea-92c8-231e9b308a25.png)


And when the alias does use an underscore: 

![image](https://user-images.githubusercontent.com/304656/89180986-19eb9600-d593-11ea-99af-915895cecac7.png)

nul800sebastiaan:
Of course the workaround for now is: don't use underscores in your aliases. 🙂

umbrabot:
Hi @ReVoid,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

blueboxes:
I am happy to take a look at a fix for this if that is ok.

blueboxes:
The issue is that the JS uploads new files with a name (file.Headers.ContentDisposition.Name) that is made up 4 parts separated by underscore which is then split out on the server. 

[...]

PR #9266, opened 2020-10-23
8539: Allow underscore in alias of image cropper property

benjaminc:
This fixes issue #8539

### Description
After this fix, an image cropper property can have an underscore in its alias.  See the screenshots below for examples.

image

image


OwainWilliams:
Hi @benjaminc 
Sorry it's taken so long to look at this but thanks for spending the time fixing this issue. 
I've now had a chance to test this and it does indeed fix issue #8539

Thank you!

OwainWilliams (approved):

Issue #8719, opened 2020-08-25
V8.7RC Show alias when multiple blocks have the same name


ronaldbarendse:
Like the issues raised for the Nested Content editor in https://github.com/umbraco/Umbraco-CMS/pull/6092 and https://github.com/umbraco/Umbraco-CMS/pull/5430, there's currently no way to differentiate element types with the same name when adding a new block:

![Create new block](https://user-images.githubusercontent.com/1051287/91146926-de3a7c80-e6b7-11ea-8927-4afe8aa911b6.png)

Or even after the blocks are added:

![Available blocks](https://user-images.githubusercontent.com/1051287/91147107-1772ec80-e6b8-11ea-95ca-cdf3002d568f.png)

## Umbraco version

I am seeing this issue on Umbraco version: 8.7RC.

### Expected result
I would expect the alias to be displayed at least in the selected/available blocks (e.g. below the name in a smaller, grey text).

Another nice feature would be if the search would also consider the alias, not only the name.

nielslyngsoe:
@ronaldbarendse I think this is spot on and very useful. Feel free to provide a PR of how you see this implemented. A loose idea on how to keep this visually simple could be to just show the Alias as a tooltip on hover. This together with the ability to search on Alias.
If you don't feel for that idea, I think you should look into providing more visual space to avoid it begin too cramped. Thanks :-)

umbrabot:
Hi @ronaldbarendse,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

ronaldbarendse:
Looking at the available blocks/thumbnails, the element type description is already shown below the name in a grey text. Having at least the alias shown on hover (as title) would fix this, otherwise it will indeed begin to look too cramped very fast (especially when a long description is present).

protherj:
I took a crack at this today with an idea.  I suspect there is more work on this one, but I wanted to see if this solution is kind of what you're looking for.  

I submitted a PR that a
[...]

PR #9267, opened 2020-10-24
8719 - Show block item alias when no element description is present

protherj:
### Prerequisites

If there's an existing issue for this PR then this fixes [8719](https://github.com/umbraco/Umbraco-CMS/issues/8719)

### Description

Adds the Block alias in parenthesis below the Block name if there is no element description present.  This allows blocks that have the same name/icon to be selected by the editor.  

See the screenshot below.  The first block has an element description.  The second block doesn't so it displays the alias.

BlockEditorAlias


ronaldbarendse:
@protherj While it's a good attempt, this doesn't fix the issue in a reliable, consistent way. Just like @nielslyngsoe suggested in the linked issue, adding it as title would at least always allow a developer to see/get the alias.

I think Nested Content addressed this issue by adding the alias (with square brackets) only to duplicate names. Maybe that's a better way, especially if it's also consistent with other property editors.

protherj:
I'll look into how NestedContent did it and try to be consistent with that

Issue #8076, opened 2020-05-07
8.6 Upload files in Media Picker is disabled for first time visitors


nielslyngsoe:
Users cannot upload files in Media Picker for the first visit of that property-editor.

Reproduce: Clear your cache and cookies, then open the Media Picker and see that the upload-button is gray. Closing and re-opening the media-picker fixes this problem.

First time visit:
![image](https://user-images.githubusercontent.com/6791648/81282988-3cd1e100-905c-11ea-9483-f7f4b53a11a4.png)

Second time:
![image](https://user-images.githubusercontent.com/6791648/81283058-53783800-905c-11ea-9abf-cbb7d0e8fd2a.png)

KeithBoynton:
I can confirm I've just experienced this too, in v8.6.1

deladroid:
I can confirm on Umbraco version 8.6.3 as well.

umbrabot:
Closing this in relation to PR #9270 - make sure to read the close reason on #9270

PR #9270, opened 2020-10-24
Fix disabled upload button in media picker on initial open

matthew-darby:
### Prerequisites

Fixes https://github.com/umbraco/Umbraco-CMS/issues/8076

### Description

See original issue for more details - how to test:

- Create a media picker that **has a start node assigned**
- Open the media picker, the upload button should not be greyed out

The issue was caused by the `hasStartNode` check failing, because the `node.path` only contains the start node ID on initial load.

When it is not an intial load, a different method is used which populates the whole node path. I have copied this to resolve the issue.

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js#L358

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js
@@ -133,7 +133,8 @@ angular.module("umbraco")
                         entityResource.getById($scope.lastOpenedNode, "media")
                             .then(ensureWithinStartNode, gotoStartNode);
                     } else {
-                        gotoStartNode();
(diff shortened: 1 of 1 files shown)


nathanwoulfe (changes requested):
Hi @matthew-darby this change works fine when a start node is set, but it throws an error when the start node is -1 (ie not set)

Could you please review with a mediapicker with no start node?

ta
Nathan

nathanwoulfe:
Pinging @matthew-darby...

matthew-darby:
@nathanwoulfe Sorry for the delay, really busy at the moment so not sure I'll have time to finish it - happy for anyone else to pick it up!

PR #9272, opened 2020-10-25
Returning response from downloadFile promise

callumbwhyte:
When working with the `umbrequesthelper.downloadFile` in the backoffice it would be handy to be able to access the response object, e.g. to check values from the response headers. Currently this function doesn't return anything...

This change makes the function return the `response` object on success. The `error` promise already returns the `response` so it's a little unclear why `success` doesn't too...

This change also brings it inline with the other methods in this helper.

The impact is thankfully minimal and won't won't break any existing implementations as promise parameters are optional.


lssweatherhead:
Thanks Callum - just taking a look at this and will get back to you!

lssweatherhead (approved):
Tested from the member export action and looks good to me! Makes sense to be returning the object and most other promise resolutions do also.

Thanks Callum!

PR #9273, opened 2020-10-25
Including ModelsBuilder.Umbraco in list of external Models Builders

callumbwhyte:
As part of the [Community Models Builder Team](https://github.com/modelsbuilder)'s work we will be releasing a new version of Models Builder.

We cannot continue using the `Umbraco.ModelsBuilder` namespace, for many obvious reasons, and so we'll be introducing a new namespace: `ModelsBuilder.Umbraco`. This offers the flexibility we need to drive the project in the direction we want.

Umbraco identifies whether to use the built in "Embedded" Models Builder or the "legacy" / external one by scanning for the `Umbraco.ModelsBuilder` assembly...

Longer term I think we need a better way of registering a Models Builder, but for now this change adds the new namespace to the list treated as external.

### DIFF
--- a/src/Umbraco.ModelsBuilder.Embedded/Compose/ModelsBuilderComposer.cs
+++ b/src/Umbraco.ModelsBuilder.Embedded/Compose/ModelsBuilderComposer.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Reflection;
 using Umbraco.Core;
 using Umbraco.Core.Logging;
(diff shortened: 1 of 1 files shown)


dawoe (approved):
Hi @callumbwhyte 
I have tested this with the "old" version of Umbraco.Modelsbuilder and it works. Is there a place where I can find the version with new namespace ?

But I assume that will work as well.

Dave

callumbwhyte:
There's nothing you can download yet... I tested by creating a dummy project with the same name and dropping the assembly in the `bin`, the existing Models Builder disappears.

nathanwoulfe:
Could the assembly names come from config/app-settings? If I then decided to roll my own builder, I could register it by updating that registration. Would make a namespace change like this easier too, since it could be managed by the package installer...

Regardless, this change is fine. Will merge.

nathanwoulfe (approved):

PR #9275, opened 2020-10-26
Make index, settings and layout props available for labels

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/8689

Make properties from Block settings available for the Block label on $settings.

**Test notes**
Use {{$settings.myprop}} in a label.

Test that the label does not fail for Blocks created before settings ElementType has been added for a Block.
— create a block with only Content-ElementType.
- create content of it.
- add Settings ElementType to your block (settings should have a property with alias "myProp")
- write a label like this. {{$settings.myProp}}
- view previously created content, and notice that no errors appear in the console.

Use {{$index}} in a label.
Test that the index is correct on init. When reordering and for nested Block Editors.

### Note
The index starts from 1 as opposed to a zero based index so that it is more editor friendly

---
_This item has been added to our backlog [AB#9064](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/9064)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
@@ -104,8 +104,8 @@
          */
         function getBlockLabel(blockObject) {
             if (blockObject.labelInterpolator !== undefined) {
-                // We are just using the data model, since its a key/value object that is live synced. (if we need to add additional vars, we could make a shallow copy and apply those.)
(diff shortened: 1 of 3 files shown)


warrenbuckley:
Code changes seems fine - off to test

warrenbuckley:
All tested and good and thanks for fixing to be a more editor friendly index number starting from 1 as opposed to 0.
Just waiting for the CI builds to go green...

PR #9279, opened 2020-10-26
When creating a new dictionary item clears out the page title

RachBreeze:
Prior to this PR the page title would reflect the previous item that had been displayed.

The page title is now reset to"Translation" when a new dictionary item is added.  This follows the same design pattern as when creating a new relationship in settings relationships, which also clears out the page title.  

I did consider setting the page title to "New Translation" or similar but create new translation does not open a full size window, so I felt this could be confusing


RachBreeze:
Hi @nul800sebastiaan  would it be possible to add the accessibility label to this, please as it's an accessibility fix  :-)?

mikecp (approved):

mikecp:
Hi @RachBreeze ,

Thanks a lot for this new accessibility improvement 👍 Looking forward to the next one :-)

Regarding the title, I think both options made sense, but it is probably better to follow already existing patterns to keep consistency across screens.

PR #9280, opened 2020-10-26
NetCore: MSDI refactor phase 2

rustybox:
Further progress for #8653

Less IFactory more MSDI

All tests passing locally (and can run UI)


rustybox:
@bergmania any chance of priority merging this when you get a chance

CoreRuntime, UmbracoBuilderExtensions and ApplicationBuilder extensions seem to have regular conflicts so be good to get this in soonish!

bergmania:
> @bergmania any chance of priority merging this when you get a chance

I will review and test this morning

src/Umbraco.Web/CompositionExtensions.cs
```diff
@@ -111,12 +117,7 @@ public static void SetDefaultRenderMvcController<TController>(this Composition c
         /// <remarks>The controller type is registered to the container by the composition.</remarks>
         public static void SetDefaultRenderMvcController(this Composition composition, Type controllerType)
         {
-            composition.OnCreatingFactory["Umbraco.Core.DefaultRenderMvcController"] = () =>
-            {
-                // no need to register: all IRenderMvcController are registered
-                //composition.Register(controllerType, Lifetime.Request);
-                Current.DefaultRenderMvcControllerType = controllerType;
-            };
+           // NOOP
```
bergmania, line 120: I guess at least we need a TODO here.. 
Not sure why this is so complicated and not just something like
 ```
composition.Services.AddScoped(typeof(IRenderMvcController), controllerType);
```

bergmania (commented):
Only small comments.. I did not found any issues in my test

src/Umbraco.Web/CompositionExtensions.cs
rustybox: I'm not sure I follow this, what's the TODO for?

This sets Current.DefaultRenderMvcControllerType but current wont exist.

rustybox: There's a TODO higher up the call stack to handle this, just removed the two extensions for now, they aren't called

bergmania: I know current won't exist in the long run, but currently, it is still used in "RenderRouteHandler", which is a class that soon we need to be migrated.

I guess in the migrated version, we will just inject the `IRenderMvcController` instead of using it from `Current.DefaultRenderMvcControllerType`, but right now we have some functionality that seems to work, but is just a noop-version.

No matter what, I don't see the idea of a noop-implementation? Either we can removed it (and replace with something similar to `composition.Services.AddScoped(typeof(IRenderMvcController), controllerType);`
otherwise I guess we at some point will need to implemented => Todo :)

bergmania: I see you removed the methods.. That makes more sense to me :)

rustybox: Oh this is a different "Current"

> composition.Services.AddScoped(typeof(IRenderMvcController), controllerType);

Wouldn't be anything like what it did though that code was commented out, all it did was set DefaultRenderMvcControllerType on Umbraco.Web.Composing.Current

rustybox: Enhanced the existing TODO in WebInitialComposer now also to give context later.

rustybox:
Added a bonus commit to make life easier later

bergmania:
Fantastic work again @rustybox 💪

PR #9281, opened 2020-10-27
Improve database query performance by introducing Count and Page to Member/Media/Document repositories so you don't redo the count(*) for every page fetched.

nzdev:
### Prerequisites

### Description
What?
Add Count() and a GetPage overload to media/document/member repositories that does not do a count. 
Replace use of existing GetPage() calls in loops with one call to Count() then looped calls to GetPage() that does not return a count.

Why?
Existing calls to GetPage() are inefficient. Every time you fetch a page you run a count(*) on the entire result set. This count(*) is also inefficient as it selects on unnecessary columns and may join unnecessary tables. Existing calls to GetPage() should be replaced with a single count(*) that performs no unnecessary joins/selects for the count then calling the new GetPage() overload.

How to test?
Existing tests pass.

```
/// <summary>
        /// Gets Count for the given query
        /// </summary>
        /// <remarks>Here, <paramref name="filter"/> can be null</remarks>
        long Count(IQuery<TEntity> query, IQuery<TEntity> filter = null);

        /// <summary>
        /// Gets paged content items.
        /// </summary>
        /// <param name="query">Query</param>
        /// <param name="filter">Filter</param>
        /// <param name="ordering">Ordering</param>
        /// <param name="pageIndex">Page Index. Indexing starts at zero</param>
        /// <param name="pageSize">Page size</param>
        /// <remarks>Here, <paramref name="filter"/> can be null but <paramref name="ordering"/> cannot.</remarks>
        IEnumerable<TEntity> GetPage(IQuery<TEntity> query, long pageIndex, int pageSize, Ordering ordering,
            IQuery<TEntity> filter = null);
```


src/Umbraco.Core/Persistence/Repositories/IMediaRepository.cs
```diff
@@ -3,6 +3,10 @@
 
 namespace Umbraco.Core.Persistence.Repositories
 {
+    public interface IMediaRepository2 : IMediaRepository, IContentRepository2<int, IMedia>
```
nzdev, line 6: Make services implement v2 interface or move v2 interface methods to v1(prefer)

Shazwazza: This is one of those places in the codebase we are allowed to break interfaces, the db repositories and services - only additive changes allowed though.

nzdev: awesome

nzdev: updated to use existing interface

Shazwazza:
Is this superseded by this https://github.com/umbraco/Umbraco-CMS/pull/9450?

src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
```diff
@@ -1055,6 +1055,55 @@ protected void ReplacePropertyValues(TEntity entity, int versionId, int publishe
...
+            var pagedResult = Database.SkipTake<TDto>((pageIndex)*pageSize, pageSize, sql);
+
+            // map the DTOs and return
+            return mapDtos(pagedResult);
+        }
+        protected int Count(IQuery<TEntity> query, Sql<ISqlContext> filter = null)
+        {
+            
+            // start with base query, and apply the supplied IQuery
+            if (query == null) query = Query<TEntity>();
```
nzdev, line 1084: This does not generate a count query

src/Umbraco.Core/Persistence/Repositories/Implement/DocumentRepository.cs
```diff
@@ -884,6 +884,73 @@ public void AddOrUpdatePermissions(ContentPermissionSet permission)
...
+                filterSql,
+                ordering);
+        }
+
+        public override int Count(IQuery<IContent> query, IQuery<IContent> filter = null)
+        {
+            Sql<ISqlContext> filterSql = null;
+
+            // if we have a filter, map its clauses to an Sql statement
+            if (filter != null)
```
nzdev, line 928: This does not generate a count query

nzdev:
> Is this superseded by this #9450?

No

PR #9284, opened 2020-10-27
Improve GetPagedDescendants() query performance with an overload that does not do a count on each fetch

nzdev:
### Prerequisites

### Description
What?
Add  GetPagedDescendants overload to media/document/repositories that does not do a count. 
Replace the existing GetPagedDescendants() where the count is not used.

Why?
Existing calls to GetPagedDescendants() are inefficient. Every time you fetch a page you run a count(*) on the entire result set. This count(*) is also inefficient as it selects on unnecessary columns and may join unnecessary tables.

How to test?
Existing tests pass.

```
/// <summary>
        /// Gets a collection of <see cref="IMedia"/> objects by Parent Id
        /// </summary>
        /// <param name="id">Id of the Parent to retrieve Descendants from</param>
        /// <param name="pageIndex">Page number</param>
        /// <param name="pageSize">Page size</param>
        /// <param name="totalRecords">Total records query would return without paging</param>
        /// <param name="ordering"></param>
        /// <param name="filter"></param>
        /// <returns>An Enumerable list of <see cref="IContent"/> objects</returns>
        IEnumerable<IMedia> GetPagedDescendants(int id, long pageIndex, int pageSize,
            IQuery<IMedia> filter = null, Ordering ordering = null);
    }
```


src/Umbraco.Core/Services/Implement/ContentService.cs
```diff
@@ -606,6 +606,29 @@ public IEnumerable<IContent> GetPublishedChildren(int id)
                 return GetPagedLocked(null, pageIndex, pageSize, out totalChildren, filter, ordering);
             }
         }
+        public IEnumerable<IContent> GetPagedDescendants(int id, long pageIndex, int pageSize,
```
Shazwazza, line 609: This code can be shared in one method which can accept a bool as to whether or not to include the count

src/Umbraco.Core/Services/Implement/ContentService.cs
```diff
@@ -624,6 +647,15 @@ private IQuery<IContent> GetPagedDescendantQuery(string contentPath)
 
             return _documentRepository.GetPage(query, pageIndex, pageSize, out totalChildren, filter, ordering);
         }
+        private IEnumerable<IContent> GetPagedLocked(IQuery<IContent> query, long pageIndex, int pageSize,
```
Shazwazza, line 650: This code can be shared in one method which can accept a bool as to whether or not to include the count

src/Umbraco.Core/Services/Implement/MediaService.cs
```diff
@@ -535,6 +535,30 @@ public IEnumerable<IMedia> GetAncestors(IMedia media)
                 return GetPagedLocked(GetPagedDescendantQuery(null), pageIndex, pageSize, out totalChildren, filter, ordering);
             }
         }
+        /// <inheritdoc />
+        public IEnumerable<IMedia> GetPagedDescendants(int id, long pageIndex, int pageSize,
```
Shazwazza, line 539: This code can be shared in one method which can accept a bool as to whether or not to include the count

src/Umbraco.Core/Services/Implement/MediaService.cs
```diff
@@ -553,6 +577,16 @@ private IQuery<IMedia> GetPagedDescendantQuery(string mediaPath)
 
             return _mediaRepository.GetPage(query, pageIndex, pageSize, out totalChildren, filter, ordering);
         }
+        private IEnumerable<IMedia> GetPagedLocked(IQuery<IMedia> query, long pageIndex, int pageSize,
```
Shazwazza, line 580: This code can be shared in one method which can accept a bool as to whether or not to include the count

src/Umbraco.Core/Persistence/Repositories/Implement/MemberRepository.cs
```diff
@@ -569,6 +569,36 @@ public void SetLastLogin(string username, DateTime date)
                 filterSql,
                 ordering);
         }
+        public override IEnumerable<IMember> GetPage(IQuery<IMember> query, long pageIndex, int pageSize, Ordering ordering, IQuery<IMember> filter = null)
```
Shazwazza, line 572: This code can be shared in one method which can accept a bool as to whether or not to include the count

Shazwazza (changes requested):
I started adding comments inline but it's the same comment in all scenarios: "This code can be shared in one method which can accept a bool as to whether or not to include the count"

There's a lot of code duplication that I think can be solved with a private shared method with a bool param.

src/Umbraco.Core/Persistence/Repositories/Implement/ContentRepositoryBase.cs
```diff
@@ -1071,6 +1071,55 @@ protected void ReplacePropertyValues(TEntity entity, int versionId, int publishe
...
+            // map the DTOs and return
+            return mapDtos(pagedResult);
+        }
+        protected int Count(IQuery<TEntity> query, Sql<ISqlContext> filter = null)
+        {
+            
+            // start with base query, and apply the supplied IQuery
+            if (query == null) query = Query<TEntity>();
+            var sql = new SqlTranslator<TEntity>(GetBaseQuery(QueryType.Many), query).Translate();
+
```
nzdev, line 1102: This does not generate a count query

src/Umbraco.Core/Persistence/Repositories/Implement/DocumentRepository.cs
```diff
@@ -884,6 +884,73 @@ public void AddOrUpdatePermissions(ContentPermissionSet permission)
...
+        {
+            Sql<ISqlContext> filterSql = null;
+
+            // if we have a filter, map its clauses to an Sql statement
+            if (filter != null)
+            {
+                // if the clause works on "name", we need to swap the field and use the variantName instead,
+                // so that querying also works on variant content (for instance when searching a listview).
+
+                // figure out how the "name" field is going to look like - so we can look for it
```
nzdev, line 933: This does not generate a count query

PR #9295, opened 2020-10-28
Netcore: Health check notifier hosted service

AndyButland:
This PR migrates the `HealthCheckNotifier` from the legacy `BackgroundTaskRunner` to a hosted service as available in .NET Core.

For testing you'll need to add the following configuration:

```
"Healthchecks": {
  "DisabledChecks": [],
  "Notification": {
    "Enabled": true,
    "FirstRunTime": "",
    "Period": "1:00:00",
    "NotificationMethods": {
      "email": {
        "Enabled": true,
        "Settings": {
          "RecipientEmail": "test@test.com"
        }
      }
    }
  }
},
```

As a child of `Global` too, you can add this to allow sending of emails (I've used [smtp4dev](https://github.com/rnwood/smtp4dev) locally to pick them up):

```
"Smtp": {
  "From": "test@test.com",
  "Host": "localhost",
  "Port": 25,
  "SecureSocketOptions": "None"
}
```


src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
```diff
@@ -293,6 +234,80 @@ public static IServiceCollection AddUmbracoCore(this IServiceCollection services
 
             coreRuntime.Configure(services);
 
+            // Add hosted services
+            services.AddHostedServices();
```
bergmania, line 238: I dont think this should be part of AddUmbracoCore. I guess it could be a separate extension on `IUmbracoBuilder`, like we have 
```
        public static IUmbracoBuilder WithCore(this IUmbracoBuilder builder)
            => builder.AddWith(nameof(WithCore), () => builder.Services.AddUmbracoCore(builder.WebHostEnvironment, builder.Config));
```

And then of cause called in`public static IUmbracoBuilder WithAllBackOfficeComponents(this IUmbracoBuilder builder)`

bergmania (commented):
Just some intital comments.. I did not test yet

src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
AndyButland: Have updated to this.

src/Umbraco.Core/Configuration/Models/Extensions/HealthCheckSettingsExtensions.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+
+namespace Umbraco.Core.Configuration.Models.Extensions
+{
+    public static class HealthCheckSettingsExtensions
+    {
+        public static int GetNotificationPeriodInMilliseconds(this HealthChecksSettings settings)
+        {
+            return settings.Notification.PeriodInHours * 60 * 60 * 1000;
```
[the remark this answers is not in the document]
AndyButland: Not sure about this one, as it'll be defined in config, and perhaps an integer number of hours is easier to enter/manage?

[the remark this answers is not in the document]
bergmania: Not sure why it should be easier to manage? I totally agree with this post https://blog.ploeh.dk/2019/11/25/timespan-configuration-values-in-net-core/

src/Umbraco.Core/Configuration/Models/SmtpSettings.cs
```diff
@@ -15,6 +15,8 @@ public class SmtpSettings : ValidatableEntryBase
 
         public int Port { get; set; }
 
+        public string SecureSocketOptions { get; set; } = "Auto";
```
JimBobSquarePants, line 18: What are the other possible configuration values here?

src/Umbraco.Core/StringExtensions.cs
```diff
@@ -1478,5 +1478,24 @@ public static string ToSafeFileName(this string text, IShortStringHelper shortSt
...
+            if (string.IsNullOrEmpty(input) || input.Length < 3 || input.Length > 4)
+            {
+                return false;
+            }
+
+            var inputWithSeparator = input.Length == 3
+                ? $"{input.Substring(0, 1)}:{input.Substring(1)}"
+                : $"{input.Substring(0, 2)}:{input.Substring(2)}";
+
+            return TimeSpan.TryParse(inputWithSeparator, out var _);
```
JimBobSquarePants, line 1498: I feel like this should simply be. Augmenting the input is too black box.

``` c#
return TimeSpan.TryParse(input, out var _);
```

src/Umbraco.Infrastructure/HealthCheck/HealthCheckResults.cs
```diff
@@ -128,6 +128,8 @@ public string ResultsAsHtml(HealthCheckNotificationVerbosity verbosity)
             return html;
         }
 
+        internal Dictionary<string, IEnumerable<HealthCheckStatus>> ResultsAsDictionary => _results;
+
         private string ApplyHtmlHighlighting(string html)
         {
             html = ApplyHtmlHighlightingForStatus(html, StatusResultType.Success, "5cb85c");
```
JimBobSquarePants, line 135: TODO: Remove random hex strings.

src/Umbraco.Infrastructure/HostedServices/RecurringHostedServiceBase.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// </remarks>
+    public abstract class RecurringHostedServiceBase : IHostedService, IDisposable
+    {
+        protected const int DefaultDelayMilliseconds = 180000; // 3 mins
+
+        private readonly int _periodMilliseconds;
+        private readonly int _delayMilliseconds;
+        private Timer _timer;
+
+        protected RecurringHostedServiceBase(int periodMilliseconds, int delayMilliseconds)
```
JimBobSquarePants, line 22: If this uses `TimeSpan` as parameters. it becomes much more flexible.

src/Umbraco.Web.Common/Builder/UmbracoBuilderExtensions.cs
```diff
@@ -30,6 +30,9 @@ public static IUmbracoBuilder WithConfiguration(this IUmbracoBuilder builder)
         public static IUmbracoBuilder WithCore(this IUmbracoBuilder builder)
             => builder.AddWith(nameof(WithCore), () => builder.Services.AddUmbracoCore(builder.WebHostEnvironment, builder.Config));
 
+        public static IUmbracoBuilder WithHostedServices(this IUmbracoBuilder builder)
+            => builder.AddWith(nameof(WithHostedServices), () => builder.Services.AddUmbracoHostedServices());
+
```
JimBobSquarePants, line 35: `AddWith` is such an awkwardly named method, the individual methods using the `With` prefix also . `Set` is far more succinct.

src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
```diff
@@ -296,6 +237,77 @@ public static IServiceCollection AddUmbracoCore(this IServiceCollection services
...
+        /// <param name="services"></param>
+        /// <returns></returns>
+        private static IServiceCollection AddUmbracoSqlCeSupport(this IServiceCollection services)
+        {
+            try
+            {
+                var binFolder = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+                if (binFolder != null)
+                {
+                    var dllPath = Path.Combine(binFolder, "Umbraco.Persistance.SqlCe.dll");
```
JimBobSquarePants, line 252: Does this work cross platform?

bergmania: it "works" by throwing an exception that is ignored in the catch statement

src/Umbraco.Core/Configuration/Models/SmtpSettings.cs
AndyButland: I meant to flag this one... it should be an enum I agree, but, the enum is defined in `MailKit.Security` which I think breaks the rules on "no external dependencies in `Umbraco.Core`.  So that's why I made it a string on the configuration model.  Thoughts?

src/Umbraco.Core/StringExtensions.cs
AndyButland: Should probably have flagged the thinking here too... the issue I found was that the currently configuration value is expected to be a time formatted like this: `Hmm` - so `330` or `1640`.  To validate though, it seems that `TimeSpan.TryParse("2500")` returns true, not false.

So I've added in the hour/minute separator, making `2500` to `25:00`, and when you try parse that, it returns false as you'd expect.

Hence in trying to retain the existing configuration format and be able to validate it as being a valid time, I've done this.

src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
AndyButland: I've just moved this, not added it.

src/Umbraco.Core/Configuration/Models/Extensions/HealthCheckSettingsExtensions.cs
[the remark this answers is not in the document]
AndyButland: OK - so currently we have `PeriodInHours` in config, e.g. a value of "24" for every 24 hours, instead we'd change the name to something like `Period` and have "24:00:00" - have I got that right?

src/Umbraco.Core/Configuration/Models/SmtpSettings.cs
AndyButland: I found I had to - otherwise smtp4dev that I was testing with locally wouldn't accept the connection and "send" the email.  I've probably got an old copy and maybe there's a bug in it, but the local problem did go away when I changed this setting.

src/Umbraco.Core/StringExtensions.cs
JimBobSquarePants: I think a breaking change is warranted then as mentioned above by @bergmania  

Incidentally I saw a tweet triggered by bad use of primitives to configure time spans in System.Web.
https://twitter.com/khellang/status/1321461708301475842

src/Umbraco.Core/Configuration/Models/Extensions/HealthCheckSettingsExtensions.cs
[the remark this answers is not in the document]
AndyButland: Have updated to use `TimeSpan` (and updated PR description with new config format).

src/Umbraco.Core/StringExtensions.cs
AndyButland: Yes, agreed, if we're minorly breaking one of the config values we just as well tidy this one up too - so it now accepts a time format of `H:mm` - e.g. `12:00`, which this validation works with as expected.

src/Umbraco.Infrastructure/HostedServices/RecurringHostedServiceBase.cs
AndyButland: Have updated to use `TimeSpan` now.

src/Umbraco.Core/Configuration/Models/SmtpSettings.cs
JimBobSquarePants: The "no dependency" rule falls down hard unless you define only a very limited set of abstractions, which means potentially implementing hundreds of projects as you break down. Personally I think "Clean Architecture" is the equivalent of "in a vacuum" and should not be considered as a practical design goal unless you want hundreds of projects in your solution.

Implementation specific properties really shouldn't exist in the basic config, if you're testing there's no reason why an `ISmptService` or whatever should actually send an email.

src/Umbraco.Core/Configuration/Models/Extensions/HealthCheckSettingsExtensions.cs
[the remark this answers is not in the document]
JimBobSquarePants: That'll give you 24 days. .

"1.00:00:00" parses to 1 day with which you would use  the `TotalHours` property to get 24 from. The joy of using Timespan everywhere is that you never even need to parse the value normally, just pass it through to the runtime methods.

src/Umbraco.Core/Configuration/Models/SmtpSettings.cs
AndyButland: I've just made the change use a locally defined enum... but will leave this one unresolved for further discussion for now.

bergmania: My alternative was to not expose it as a configuraiton value, and use "Auto" in:
```
client.Connect(_globalSettings.Smtp.Host,
                        _globalSettings.Smtp.Port,
                        MailKit.Security.SecureSocketOptions.Auto)
```

I not sure you will every change that value from auto?

AndyButland: Auto is the default, so you don't need to provide it if we do that.  As I say though, in testing locally with smtp4dev, I found "Auto" wouldn't work - gave me an error in trying to set up an SSL connection - and only when I switched it to "None" did it accept the connection.

src/Umbraco.Infrastructure/HealthCheck/HealthCheckResults.cs
AndyButland: Have created constants for these.

src/Umbraco.Web.Common/Builder/UmbracoBuilderExtensions.cs
AndyButland: Suggest this is amended or otherwise outside of this PR, and it'll affect more than just what's being included here.

src/Umbraco.Core/Configuration/Models/SmtpSettings.cs
JimBobSquarePants: You could always pass the environment as a parameter to the mailkit service and switch to none if on dev?

src/Umbraco.Web.Common/Builder/UmbracoBuilderExtensions.cs
JimBobSquarePants: Yeah, agree, was just highlighting.

src/Umbraco.Core/StringExtensions.cs
```diff
@@ -1478,5 +1478,31 @@ public static string ToSafeFileName(this string text, IShortStringHelper shortSt
...
         }
+
+        /// <summary>
+        /// Validates a string matches a time.
+        /// </summary>
+        /// <param name="input">Time representation (in H:mm format, e.g. 3:30, 22:00)</param>
+        /// <returns>True if provided input is a valid time in H:mm format, false if not.</returns>
+        public static bool IsValidTimeFormat(this string input)
+        {
+            if (string.IsNullOrEmpty(input) || !input.Contains(":") || input.Length < 4 || input.Length > 5)
```
JimBobSquarePants, line 1489: I still think this should be simply `Timespan.TryParse()`. Are we assuming invariant culture?

src/Umbraco.Tests.UnitTests/Umbraco.Core/ShortStringHelper/StringExtensionsTests.cs
```diff
@@ -293,14 +293,25 @@ public void ReplaceManyWithCharMap()
...
             Assert.AreEqual(expected, output);
         }
+
+        [TestCase("", false)]
+        [TestCase("1234", false)]
+        [TestCase("6:18", true)]
+        [TestCase("12:34", true)]
+        [TestCase("25:03", false)]
+        [TestCase("18:61", false)]
+        [TestCase("1:14:23", false)]
```
JimBobSquarePants, line 310: `TimeSpan.Parse("1:14:23")` is valid. 1 hr, 14 min, 23 sec

src/Umbraco.Core/StringExtensions.cs
AndyButland: Yes, you're right - I'll revert this.  I changed it on reflection thinking I'm trying to validate an actual time here, not a timestamp.  It's to validate a configured string is correct, which has previously been intended to be a time (i.e. hours and minutes only).  But now you've nudged me, there's no reason why this couldn't be a time more than 24 hours, so a TimeSpan is perfectly valid.

src/Umbraco.Tests.UnitTests/Umbraco.Core/ShortStringHelper/StringExtensionsTests.cs
AndyButland: Have reverted my change, and made it clear this is validating a `TimsSpan`, not just a time.

bergmania:
Tested and it works for me without changing the `SecureSocketOptions` (I'm using Papercut as test SMTP).

Good work, as always 💪

KevinJump:
Hi the implementation of `RecurringHostedServiceBase` has an internal abstract class. 

```
internal abstract Task PerformExecuteAsync(object state);
```

If i am reading that right - this means external code can't inherit this base class to implement their own recurring background tasks? _(or is there something clever i need to do?)_

Was that intentional ?  - as it stands it probably means people just repeating quite a bit of code to do this in their own services to do this.

bergmania:
I don't see why it cannot be public.

AndyButland:
Me neither.  It looks like in this PR it was public, but I changed it in [this later commit](https://github.com/umbraco/Umbraco-CMS/pull/9324/commits/21c29d6446f82974df9756904bbf77b7c08790a2) resolving some issues around unit testing these hosted services.  I can't think of a reason why, so must have been a mistake.

Issue #9286, opened 2020-10-27
Save button doesn't work in Content Type Editor (when using non-live MB mode)


skttl:
When using a non-live ModelsBuilder mode, the save button in the content type editor is changed to a split button, where you can save and generate models.
![image](https://user-images.githubusercontent.com/3726467/97313289-bce83f00-1866-11eb-8c1c-662bffaab3f8.png)

Clicking Save here doesn't do anything. If I choose save and generate models, it saves and generates models as expected. After this, the save button is reset to the original state (without the split), and then the save button works. again.

![FpYcMN2w0j](https://user-images.githubusercontent.com/3726467/97313467-ee610a80-1866-11eb-90ff-70eaa8e51620.gif)

## Umbraco version

I am seeing this issue on Umbraco version: 8.9.0

skttl:
Seems like the split button is missing a click-handler
![image](https://user-images.githubusercontent.com/3726467/97314103-a55d8600-1867-11eb-8aec-80d76d60246e.png)


This is how the regular button looks
![image](https://user-images.githubusercontent.com/3726467/97314148-b1e1de80-1867-11eb-93c9-94bf3e20f817.png)

nathanwoulfe:
@skttl that makes sense - buttons work a lot better when they have something to do...

Marking it up-for-grabs 😄

umbrabot:
Hi @skttl,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
I am seing this issue as well in Umbraco v8.9.0 after switching ModelsBuilder mode from default PureLive mode to AppData mode.
It seems the save buttons is showing the spinner sometimes is the application has been recycled (e.g. via ModelsBuilder "Generate models" button), but not on subsequent clicks.

DanDiplo:
Yep, I'm getting this with AppData mode. Clicking button does absolutely nothing.

Surely this is a critical issue and not an "up-for-grabs" issue, though, as it completely breaks being able to develop with Umbraco?

[...]

PR #9300, opened 2020-10-28
Fix save buttons not working when using non-live models builder

matthew-darby:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9286

### Description

See original issue for full details. The buttons work again when the type is set to "button", which is the default (as shown below) so the types in the PR can be safely removed.

https://github.com/umbraco/Umbraco-CMS/blob/e9e8a0354ddb36d12ebdecddf0569132dc96f4be/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbutton.directive.js#L128


OwainWilliams (approved):
Thanks for fixing this! 
I've checked the changes and all looks to be working again as expected.

PR #9302, opened 2020-10-28
Updated editor.service.js for consistency

abjerner:
- Changed type names to lowercase to reflect corresponding types in JavaScript
- File used a mix of `callback` and `function`. Updated to use `function`.
- Updated sentences to end with period.


nathanwoulfe (approved):
👍 thanks @abjerner

PR #9304, opened 2020-10-29
Bugfix: Add null check for null converted treenode icon so .startsWith does not throw

nzdev:
### Prerequisites

### Description

Adds a check so ensure the converted treenode icon is not undefined before calling converted.startswith(.) which will throw.
This issue causes  the content/settings/media trees not to work. No way to fix the issue from the cms


mikecp (approved):

mikecp:
Hi @nzdev ,

Thanks for your PR! Nice catch :-)

Issue #8964, opened 2020-09-25
Umbraco Examine indexes contain duplicate fields and wrong field definitions


ronaldbarendse:
I've noticed Umbraco by default adds duplicate fields to the Examine indexes:
- `__Icon` and `icon`
- `__NodeId` and `id`
- `__Path` and `path`

Other systems fields are not prefixed with the `LuceneIndex.SpecialFieldPrefix` (double underscore):
- `level`
- `nodeName` (and all culture variants)
- `nodeType` (the numeric ID version of `__NodeTypeAlias`)
- `parentID`
- `sortOrder`
- `templateID`
- `updateDate` (and all culture variants)
- `urlName` (and all culture variants)
- `writerID`
- `writerName`

The culture variant fields for `updateDate` are also in a different format (`09/24/2020 14:01:47`) then the invariant field (numeric). This is either because the values are `DateTime?` (nullable) and not formatted correctly or the following code does not set the same field definition type for the culture variant fields:
https://github.com/umbraco/Umbraco-CMS/blob/12e88fde423f59eae1dff76b474974382d672fe3/src/Umbraco.Examine/UmbracoFieldDefinitionCollection.cs#L70-L92

--------------

Umbraco also mixes the usage of both `__Path` and `path`. As it's used as special/system field, the following code should probably use `UmbracoExamineIndex.IndexPathFieldName`:
https://github.com/umbraco/Umbraco-CMS/blob/f29c5f0e3b0dc79456e2726f3aeace04e51a28b0/src/Umbraco.Examine/UmbracoContentIndex.cs#L71
https://github.com/umbraco/Umbraco-CMS/blob/db6714d3b8f8e04dceb607c3a25117331f6b525d/src/Umbraco.Examine/ContentValueSetValidator.cs#L19
https://github.com/umbraco/Umbraco-CMS/blob/db6714d3b8f8e04dceb607c3a25117331f6b525d/src/Umbraco.Examine/ContentValueSetValidator.cs#L123

What's even more concerning, is that the system fields are set based on the non-system fields (and in the case of `path` even overridden!):
https://github.com/umbraco/Umbraco-CMS/blob/7cbb2840a40997f3bf9106b9192fbbbe87f0dc95/src/Umbraco.Examine/UmbracoExamineIndex.cs#L175-L180
[...]

PR #9305, opened 2020-10-29
Match neutral cultures as Examine culture fields

ronaldbarendse:
### Prerequisites

### Description
As part of investigating https://github.com/umbraco/Umbraco-CMS/issues/8964, I noticed the culture variant Examine fields (with the lower-case culture name suffix) weren't returned for neutral cultures (e.g. `en` or `nl`), but only for specific cultures (e.g. `en-US` or `nl-NL`).

This causes issues with searching in a specific language (as the fields for all neutral languages are considered invariant), but also with setting the right field definitions, as can be tested with the example in https://github.com/umbraco/Umbraco-CMS/issues/8964#issuecomment-700588938.

This can be tested by adding both specific and neutral languages to Umbraco (e.g. `en`, besides the default `en-US`), adding some culture variant content and inspecting the result of the `GetCultureFields()` extension method of any Examine index.

The issue with the field definitions can be tested by inspecting the field value for the culture variant `updateDate`: without this fix, the value in `updateDate_en-us` will be formatted as a number (Lucene datetime, same as the invariant field), but `updateDate_en` will be formatted as a non-searchable string. With this PR applied, both culture variant fields will have the same, correct format.


src/Umbraco.Examine/ExamineExtensions.cs
```diff
@@ -27,7 +27,7 @@ public static class ExamineExtensions
         /// <remarks>
         /// myFieldName_en-us will match the "en-us"
         /// </remarks>
-        internal static readonly Regex CultureIsoCodeFieldNameMatchExpression = new Regex("^([_\\w]+)_([a-z]{2}-[a-z0-9]{2,4})$", RegexOptions.Compiled);
+        internal static readonly Regex CultureIsoCodeFieldNameMatchExpression = new Regex("^(?<FieldName>[_\\w]+)_(?<CultureName>[a-z]{2,3}(-[a-z0-9]{2,4})?)$", RegexOptions.Compiled | RegexOptions.ExplicitCapture);
```
ronaldbarendse, line 30: The main fix is making `-[a-z0-9]{2,4}` optional, but as that also increased the group count from 3 to 4 (and this was checked by some implementations).

To make this regex easier to use and even backwards-compatible, I've named the parts for the field name and culture name and added `RegexOptions.ExplicitCapture`. This results is the same group count, order/index and even allows getting the right values by group name 🎉

ronaldbarendse: I've also changed the neutral culture name to be 2 **or 3** characters long.

Some specific cultures might even include a third part (e.g. `az-Cyrl-AZ` or `az-Latn-AZ`), but those aren't matched yet. These cultures are probably not used very often, so to avoid making the regex more complex, slower and possibly match even more non-culture fields, I've omitted that for now.

src/Umbraco.Examine/UmbracoFieldDefinitionCollection.cs
```diff
@@ -73,13 +73,13 @@ public override bool TryGetValue(string fieldName, out FieldDefinition fieldDefi
                 return true;
 
             //before we use regex to match do some faster simple matching since this is going to execute quite a lot
-            if (!fieldName.Contains("_") || !fieldName.Contains("-"))
+            if (!fieldName.Contains("_"))
```
ronaldbarendse, line 76: Neutral culture fields don't include a dash, only the underscore.

nul800sebastiaan:
Looking good, thanks @ronaldbarendse! 👍

PR #9308, opened 2020-10-29
NetCore: MSDI refactor remove IFactory & IRegister

rustybox:
Short and sweet, IFactory gone

Progress towards  #8653


rustybox:
@bergmania 🚀 ? 

1 test failing but same as upstream 

> FullDataSetRepositoryCachePolicy(True)

bergmania:
Thanks, I look into it tonight or tomorrow

rustybox:
ahh ok i'll merge next one in here too then

rustybox:
Still all green here

image

src/Umbraco.Core/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        /// <summary>
+        /// Registers a unique service with an implementation type.
+        /// </summary>
+        public static void AddMultipleUnique<TService1, TService2, TImplementing>(this IServiceCollection services)
+            where TImplementing : class, TService1, TService2
+            where TService1 : class
+            where TService2 : class
+        {
+            services.AddUnique<TImplementing>();
```
bergmania, line 57: Do there exist a better way to do this now? 

I theory we don't want to register it as the concrete `TImplementing`, but only as the `TService1`. `TService2,...

src/Umbraco.Tests/Web/UmbracoHelperTests.cs
```diff
@@ -28,8 +28,10 @@ public void TearDown()
 
         private void SetUpDependencyContainer()
```
bergmania, line 29: Seems to be dead code :)

bergmania (commented):
Impressive work again @rustybox :muscle: :muscle: :muscle: :muscle:

I found one view, that needs to be updated because it uses Current.Factory 😭  Please take a look in `Rte.cshtml`

Do we have a subtask about getting rid of our own Lifetime or it this part of the existing tasks?

I tested and found no issues:tada:

src/Umbraco.Core/ServiceCollectionExtensions.cs
rustybox: I'd argue removing the extension entirely is better than having it, it's purpose isn't particularly obvious and it's only used 3 times.

MSDI requires you to be very explicit compared to many containers so we wont be able to avoid multiple calls to add services if we want multiple registrations.

src/Umbraco.Tests/Web/UmbracoHelperTests.cs
rustybox: Aye that entire class is dead infact, but I have removed Current.Reset and all calls in 384ae54

rustybox:
> I found one view, that needs to be updated because it uses Current.Factory 😭 Please take a look in `Rte.cshtml`
Resolved with comment on alternate solution for later.

> Do we have a subtask about getting rid of our own Lifetime or it this part of the existing tasks?
No but just as easy to resolve now as it is to add a task, so done.

rustybox:
Unable to load the service index for source https://www.myget.org/F/umbracocore/api/v3/index.json welp

bergmania:
http://status.myget.org/

I will watch the status and rebuild when it is okay again

rustybox:
Not sure how I feel about 73f5fd9, it does the job but it feels weird.

bergmania:
> Not sure how I feel about [73f5fd9](https://github.com/umbraco/Umbraco-CMS/commit/73f5fd90394fa9747df6761216c68a01e1a51d45), it does the job but it feels weird.
> 
> Edit: Perhaps that's just because all of Umbraco.Core.IO.FileSystems feels weird.

I think it just how it has to be for now (We are also plan to look into filesystems). Please don't commit more on this PR. It seems to build, and if the buildserver is happy, I will merge 💪

PR #9312, opened 2020-10-29
This PR sets the page title when editing and maintaining languages

RachBreeze:
A page title should be set to allow people using assistive technology know where they are in the site/  The bonus of this is that people who work with a lot of tabs (eg me) can also find the page to work with in their browser easily.

This PR sets the page title for language maintenance section


RachBreeze:
Hi @nul800sebastiaan  would it be possible to add the accessibility label to this, please as it's an accessibility fix :-)?

mikecp:
Hi @RachBreeze, here you go :-)

RachBreeze:
@mikecp  thank you

mikecp:
Hello @RachBreeze,

Thanks for your PR, it"s a nice improvement indeed :-)

If you don't mind, I would suggest trying to integrate the "editLanguage" variable into the "vm.labels" collection so that translation labels are grouped together.
I also noticed that the line "addLanguage = vm.labels.addLanguage;" was causing troubles and can safely be removed it seems.

If you could make those two small adjustments that would be great!

Thanks again!

RachBreeze:
Hi @mikecp thank you for reviewing my PR I have made this update now :-)

mikecp (approved):
Hi @RachBreeze , thanks a lot :-) !
I took the liberty to remove the "addLabel=XXX" on line 64 because, although it had no effect on the functionality itself, it was generating JS errors.
I hope you won't mind ;-)

PR #9316, opened 2020-10-30
Make RTE toolbar sticky

### LINKED ISSUE #9315: Umbraco V8 RTE - Make menu bar sticky
I'm pretty sure prior to Umbraco v8, the rich text editor fields were independently scrollable, however in v8 I think this has been updated so that the input resizes based on the content within it. This is cool, but now if you create a lot of rich text content the editor toolbar goes off page and can be a pain to relocate. I know there is a new "distraction free" mode, but I'm not a fan of that approach as it can be buggy to get the toolbar to show when you need it.

A nice, simple alternative could be to just make the RTE toolbar sticky such that when you scroll the main editor window, the rte toolbar remains stuck just below the main menu bar. If you were to scroll beyond the RTE area, you could then just let it disapear and only re-appear when the RTE comes back into view.

I think this would be a much nicer UX and means the RTE toolbar is never out of reach.

mattbrailsford:
### Prerequisites

Fixes #9315

### Description
CSS update to the RTE toolbar to make it sticky. Now if the RTE area is larger than the viewport, as you scroll down the RTE toolbar will remain sticky to the content area meaning you don't have to scroll back up to access it.

To test, create an RTE in "classic" mode then enter a lot of content that forces it scroll beyond the view port height. When you scroll down, the RTE toolbar should now stick to the top of the viewport.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/rte.less
+++ b/src/Umbraco.Web.UI.Client/src/less/rte.less
@@ -35,6 +35,11 @@
     box-sizing: border-box;
 }
 
+.umb-rte .mce-top-part {
+	position: sticky;
+	top: 0;
+}
+
 /* make sure the menu wraps */
 .umb-rte .mce-top-part.mce-container div {
     white-space: normal;


emmaburstow:
Hey Matt,

thanks for the PR. I love it :)

We'll take a look and let you know if we have questions.

Em

emmaburstow (approved):
Love it. Works as advertised

PR #9322, opened 2020-10-30
V8/feature/ab9079 use tree for element type pickers

nielslyngsoe:
Use tree picker for Element Type Picking in Block List Prevalue Editor.

image

**test notes:**
Make a new Block from an existing ElementType.
Add Settings Element Type for Block, from an existing ElementType.
Make a new Block from an ElementType created in the flow. (use create new element type button in tree-picker)
Add Settings from an ElementType created in the flow. (use create new element type button in tree-picker)

### DIFF
--- a/src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
+++ b/src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
@@ -1825,7 +1825,7 @@ Mange hilsner fra Umbraco robotten
   </area>
   <area alias="blockEditor">
     <key alias="headlineCreateBlock">Vælg Element-Type</key>
-    <key alias="headlineAddSettingsElementType">Tilføj en indstillings afsnit</key>
+    <key alias="headlineAddSettingsElementType">Tilføj en indstillings Element-Type</key>
     <key alias="headlineAddCustomView">Tilføj visning</key>
     <key alias="headlineAddCustomStylesheet">Tilføj stylesheet</key>
     <key alias="headlineAddThumbnail">Vælg billede</key>
--- a/src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
+++ b/src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
@@ -1824,12 +1824,12 @@ Mange hilsner fra Umbraco robotten
     <key alias="tooltipForPropertyActionsMenu">Åben egenskabshandlinger</key>
   </area>
   <area alias="blockEditor">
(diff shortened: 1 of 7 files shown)


src/Umbraco.Web.UI/Umbraco/config/lang/da.xml
```diff
@@ -1824,12 +1824,12 @@ Mange hilsner fra Umbraco robotten
...
   </area>
   <area alias="blockEditor">
-    <key alias="headlineCreateBlock">Opret ny blok</key>
+    <key alias="headlineCreateBlock">Vælg Element-Type</key>
     <key alias="headlineAddSettingsElementType">Tilføj en indstillings afsnit</key>
     <key alias="headlineAddCustomView">Tilføj visning</key>
     <key alias="headlineAddCustomStylesheet">Tilføj stylesheet</key>
     <key alias="headlineAddThumbnail">Vælg billede</key>
-    <key alias="labelcreateNewElementType">Opret ny</key>
+    <key alias="labelcreateNewElementType">Opret ny Element-Type</key>
```
bjarnef, line 1832: @nielslyngsoe I think the correct term in Danish is `elementtype` like we have `dokumenttype`. It is what we in Danish call "sammensatte ord" 😁 
https://sproget.dk/raad-og-regler/typiske-problemer/et-eller-flere-ord

It would be `Elementtype` if it is in start of the sentence.

nielslyngsoe: I saw this approach begin used earlier, so I thought that was the right approach.

But now you made me look it up, and from my research, it seems that you are right to make it `elementtype`. both because we can merge names in danish but also because its not a "person-, virksomheds-, eller stednavn" so it should be with lowercase. So I learned something new today.

I think this needs to be corrected in general, so please make a PR for that, so we can have it consistent. Thanks

bjarnef: Yeah, I think there are some general translations from English to Danish, where it can't be translated directly.

"Tilføj en indstillings elementtype" would probably be "Tilføj en indstillingselementtype" or "Tilføj en indstilling for elementtypen".

warrenbuckley:
Code reads fine off to test this out

warrenbuckley:
Works as expected & described :)
Waiting for CI builds to go green

warrenbuckley:
v8/dev builds fine locally & tests all pass with the usual exception of the explicit ones that tend to fail

warrenbuckley:
Due for 8.10.0

PR #9323, opened 2020-10-30
NetCore: Fixes registration issue

AndyButland:
Following the MSDI refactor, I found a problem where a singleton class's field was null when it had previously been set (specifically the `UmbracoRequestLifetime.RequestStart` event).

Tracked it down to the fact that this class has two interfaces, and the recent changes meant two implementations were getting registered, when I think there should be just one.

The amend in this PR resolves the issue and restores the expected behaviour.

For reference, previously (before the MSDI refactor), we had this:

```
composition.RegisterMultipleUnique<IUmbracoRequestLifetime, IUmbracoRequestLifetimeManager, UmbracoRequestLifetime>();
```

### DIFF
--- a/src/Umbraco.Core/ServiceCollectionExtensions.cs
+++ b/src/Umbraco.Core/ServiceCollectionExtensions.cs
@@ -12,6 +12,21 @@ namespace Umbraco.Core
             where TImplementing : class, TService
             => services.Replace(ServiceDescriptor.Singleton<TService, TImplementing>());
 
+        /// <summary>
+        /// Registers a unique service as a single instance implementing two interfaces.
+        /// </summary>
+        /// <remarks>
+        /// Hat-tip: https://stackoverflow.com/a/55402016/489433
+        /// </remarks>
+        public static void AddUnique<TService1, TService2, TImplementing>(this IServiceCollection services)
+            where TService1 : class
+            where TService2 : class
+            where TImplementing : class, TService1, TService2
+        {
+            services.Replace(ServiceDescriptor.Singleton<TService1, TImplementing>());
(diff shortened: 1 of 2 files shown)


bergmania:
@rustybox, FYI I have merged this due to the error found by @AndyButland

rustybox:
Ahh same instance is shared across registrations, rushed that.

There's a couple more that used RegisterMultipleUnique (all 3 invocations in AspNetCoreComposer)

```csharp
     composition.RegisterMultipleUnique<IUmbracoApplicationLifetimeManager, IUmbracoApplicationLifetime, AspNetCoreUmbracoApplicationLifetime>();
     composition.RegisterMultipleUnique<IUmbracoRequestLifetime, IUmbracoRequestLifetimeManager, UmbracoRequestLifetime>();
     composition.RegisterMultipleUnique<ISessionIdResolver, ISessionManager, AspNetCoreSessionManager>();
```

I'm personally not a fan of the extension, I find it less obvious than the following

```csharp
composition.Services.AddUnique<IUmbracoApplicationLifetimeManager, AspNetCoreUmbracoApplicationLifetime>();
composition.Services.AddUnique<IUmbracoApplicationLifetime>(x => x.GetRequiredService<IUmbracoApplicationLifetimeManager>());
```

YMMV

bergmania:
> @bergmania @AndyButland i'll gladly put in a PR to resolve the other two instances, just be good to know the consensus on AddMultiple extension vs explicit as above.

I don't have a problem with the extension method (maybe that's just because I'm used to that kind of registrations..)

I think we should keep it, but maybe a better name is required, to people using it, understand that it is the same instance that is used to multiple interfaces..

PR #9324, opened 2020-10-31
Netcore: Migrates further background tasks to hosted services

AndyButland:
In this PR I've migrated three more background tasks to hosted services.

- KeepAlive
- LogScrubber
- TempFileCleanup

For the latter have moved a couple of methods to `IOHelper` to support unit testing.

Includes [this PR](https://github.com/umbraco/Umbraco-CMS/pull/9323) to fix a dependency registration issue.


src/Umbraco.Core/IO/CleanFolderResult.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.IO;
+
+namespace Umbraco.Core.IO
+{
+    public enum CleanFolderResultStatus
+    {
+        Success,
+        FailedAsDoesNotExist,
+        FailedWithException
+    }
```
bergmania, line 12: I think this should be in its own file..

src/Umbraco.Infrastructure/HostedServices/TempFileCleanup.cs
```diff
@@ -0,0 +1,92 @@
...
+using Umbraco.Core.IO;
+
+namespace Umbraco.Infrastructure.HostedServices
+{
+    /// <summary>
+    /// Used to cleanup temporary file locations.
+    /// </summary>
+    /// <remarks>
+    /// Will run on all servers - even though file upload should only be handled on the master, this will
+    /// ensure that in the case it happes on replicas that they are cleaned up too.
```
bergmania, line 14: I know this comment is not new, but it seems like it is wrong.. I only runs if server is main dom

src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
```diff
@@ -293,10 +293,24 @@ public static IServiceCollection AddUmbracoSqlServerSupport(this IServiceCollect
         public static IServiceCollection AddUmbracoHostedServices(this IServiceCollection services)
         {
             services.AddHostedService<HealthCheckNotifier>();
+            services.AddHostedService<KeepAlive>();
+            services.AddHostedService<LogScrubber>();
+            services.AddHostedService<TempFileCleanup>();
 
             return services;
         }
```
bergmania, line 301: In theoy this could now be called from the `SchedulerComposer`. Not sure whether we should stick to the umbraco way or .net way here. Thoughts?

src/Umbraco.Core/IO/IOHelper.cs
```diff
@@ -188,5 +188,62 @@ public string GetRelativePath(string path)
...
+
+            return tempFolderPaths.Select(x => new DirectoryInfo(x)).ToArray();
+        }
+
+        /// <summary>
+        /// Cleans contents of a folder.
+        /// </summary>
+        /// <param name="folder">Folder to clean.</param>
+        /// <param name="age">Age of files within folder to delete.</param>
+        /// <returns>Result of operation</returns>
```
bergmania, line 216: Maybe it should be mentioned, that on failure, some files can still have been cleaned.

src/Umbraco.Core/IO/IOHelper.cs
```diff
@@ -188,5 +188,62 @@ public string GetRelativePath(string path)
...
+            };
+
+            foreach (var tempFolderPath in tempFolderPaths)
+            {
+                // Ensure it exists
+                Directory.CreateDirectory(tempFolderPath);
+            }
+
+            return tempFolderPaths.Select(x => new DirectoryInfo(x)).ToArray();
+        }
```
bergmania, line 209: Not sure why this returns an array, do we every have multiple temp folders?

bergmania (commented):
Overall changes looks good :) Just minor comments

src/Umbraco.Core/IO/CleanFolderResult.cs
AndyButland: Agreed, have moved.

src/Umbraco.Infrastructure/HostedServices/TempFileCleanup.cs
AndyButland: I think it may still be valid.  If you look at other scheduled tasks - e.g. `KeepAlive` - these have checks via `_serverRegistrar.GetCurrentServerRole()` to ensure they don't run on replica servers.  This particular task doesn't have that check though, so seems this comment is there to call that out.

src/Umbraco.Web.Common/Extensions/UmbracoCoreServiceCollectionExtensions.cs
AndyButland: No major preference, but arguably this work to use hosted services is already moving from a custom Umbraco way to a more "standard .NET Core" way, so arguably we should continue that to include registration.

src/Umbraco.Core/IO/IOHelper.cs
AndyButland: Not sure either, but this is just copied over from existing code (I've moved it into the `IOHelper` from the original scheduled task).

src/Umbraco.Core/IO/IOHelper.cs
AndyButland: Have updated comments to reflect this.

Issue #9152, opened 2020-10-13
C# API source links are broken


jannikanker:
## Detailed description
I'm not sure where this issue belongs (stems from), but the content is on the Our domain so I figured I'd report it here. If I go to https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Core.Attempt.html (example) and click "View Source" I am pointed towards https://github.com/umbraco/Umbraco-CMS/blob/8.8/src/Umbraco.Core/Attempt.cs/#L8 which gives me a Page Not Found.

However, https://github.com/umbraco/Umbraco-CMS/blob/v8/8.8/src/Umbraco.Core/Attempt.cs/#L8 works.

It seems to be the same on every page I've tried, so I'd venture a guess that it's a common issue for all the source links currently.

I'd be game to try and fix this myself if someone could point me towards where the magic happens? Or let me know if this issue actually belongs in another repo :-)

jannikanker:
Argh indeed :-/ I'll take a look ASAP :-)

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/9326

nul800sebastiaan:
Bad news/good news: too bad, that didn't work. 

But! Good: The way to make it work is if we build the docs from a tag, it looks like we just can not have a / in the branch name, that messes up the docs. But now it works beautifully when using the tag `release-8.13.0` to build from. So all good now and we know what to do in the future.

PR #9326, opened 2020-10-31
use v8 branch as default for API Docs

bielu:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9152
### Description
Issue with generation of DocFX is that the url is always based on branch on which it was build, it means if branch is not public or was removed the link will be incorrect, to avoid that I think the best solution is to use v8/contrib branch which in most of cases will be containing the code describe with API Docs.

### DIFF
--- a/src/ApiDocs/docfx.json
+++ b/src/ApiDocs/docfx.json
@@ -50,6 +50,7 @@
     ],
     "globalMetadata": {
       "_appTitle": "Umbraco c# Api docs",
+      "branch": "v8/contrib",
       "_enableSearch": true,
       "_disableContribution": false
     },


nul800sebastiaan:
I could have sworn I merged this one already, thanks @bielu!

nul800sebastiaan:
Ah.. too bad, it didn't work, I'll revert this. The way to make it work is if we build the docs from a tag, it looks like we just can not have a `/` in the branch name, that messes up the docs. But now it works beautifully when using the tag `release-8.13.0` - thanks for the attempt anyway @bielu! 👍

bielu:
@nul800sebastiaan that's pretty weird as it was building perfectly fine locally for me, but I guess it might be on specifics of build pipeline or something like that😅 but cool you found another workaround

PR #9327, opened 2020-10-31
Add ellipsis to the Sort action in Content and Media

SteveVaneeckhout:
### Prerequisites

### Description
The Actions menu/dropdown in Content and Media contains a Sort link. It should end with an ellipsis ... because it opens a dialog just like the other items, like Move, Copy, Permissions.

To reproduce:
- Open the Content section
- Right-click on a content item and notice that Sort ends with an ellipsis ...
- Open the content item and click on the Actions dropdown in the top-right and notice that Sort ends with an ellipsis ...
- Reproduce these same steps in the Media section.

image


nathanwoulfe (approved):
Thanks @SteveVaneeckhout, this looks good to me - consistency is king!

PR #9329, opened 2020-11-01
.NetCore: change health check notifier first run time to cron expression

AndyButland:
've been round the houses a little with this on a previous PR, but still not sure we quite have it right.  It's around the scheduling for healthcheck notifications, where we want to be able to configure:

1. How often it should run (currently a timespan, defaulted to 24 hours), provided via the field `Period`.
2. When it should run first (currently a string representing a time and validated as a TimeSpan, if not provided runs after start-up), provided via the field `FirstRunTime`.

The first I think is OK.  The second though I think needs to be more flexible, so we can account for use-case 4 (and others) below.  Also, the validation, checking for a valid time span, on reflection I don't think is right here (as we want to accept something from which we can define and find a particular time in the future, not a generic period of time).

Use cases:

```
1. Run every 24 hours = set `Period` to a timespan of 24 hours
2. Run every week = set `Period` to a timespan of 7 days
3. Run every 24 hours at 4:00 = set `Period` timespan of 24 hours and `FirstRunTime` to `4:00`.
4. Run every week on Sunday at 4:00 = set `Period` to timespan of 7 days, but set `FirstRunTime` to ?.
```

Seems like this is really looking for a _cron expression_, so have implemented that, and the last 2 use cases become: 

```
3. Run every 24 hours at 4:00 = set `Period` timespan of 24 hours and `FirstRunTime` to `0 4 * * *`.
4. Run every week on Sunday at 4:00 = set `Period` to timespan of 7 days, but set `FirstRunTime` to `0 4 * * 0`.
```

I've brought in [this dependency](https://github.com/atifaziz/NCrontab) to help with this (Apache licenced).

Includes [this PR](https://github.com/umbraco/Umbraco-CMS/pull/9323) to fix a dependency registration issue.


src/Umbraco.Core/StringExtensions.cs
```diff
@@ -1480,18 +1480,23 @@ public static string ToSafeFileName(this string text, IShortStringHelper shortSt
...
+        public static bool IsValidCronTab(this string input)
         {
             if (string.IsNullOrEmpty(input))
             {
                 return false;
             }
 
-            return TimeSpan.TryParse(input, out var _);
+            var parts = input.Split(new string[] { " " }, StringSplitOptions.RemoveEmptyEntries);
+            return parts.Length == 5 || parts.Length == 6;
```
bergmania, line 1499: Not sure if we should use a more advanced validation.
Regex or even try to parse it? Currently "a b c d e f" is value

bergmania (commented):
Code looks good, and IMO it makes good sense to use Cron syntax for this information

src/Umbraco.Core/StringExtensions.cs
AndyButland: Yes, I was debating this, as the validation I've added it very light at the moment.

Think ideally we'd parse it using the same library that we'll do so to use it in the hosted service, but to do that, we'd need to add a reference to the library to `Umbraco.Core`, or move these validator components out of here and into `Umbraco.Infrastructure`.

Or we write (or [borrow](https://stackoverflow.com/questions/14203122/create-a-regular-expression-for-cron-statement)) our own regex.  Cron expressions can more complex than I'd thought though!

What would you prefer?

bergmania: I would use clean architecture to inject our one ICrontTabParser (living in core) where needed and have the implementation using NCronTab  in infrastructure

like this:
https://github.com/umbraco/Umbraco-CMS/commit/2bee18c30eee0a11744b232f43479509f103b904

I even found an issue with the validation, because the `* * * * * 1` was invalid in NCronTab

AndyButland: Nice, yes that looks good to me.  Only thing I noticed was you've got the name `cronTap` rather than `cronTab` in a few places.

bergmania: I fixed the `cronTap` vs  `cronTab` and cherry picked to commit into this pr

PR #9330, opened 2020-11-02
NetCore: Fixes registration issue (continued)

rustybox:
Add additional "mutliple unique" registrations and rename extension for clarity.

#9323

### DIFF
--- a/src/Umbraco.Core/ServiceCollectionExtensions.cs
+++ b/src/Umbraco.Core/ServiceCollectionExtensions.cs
@@ -13,18 +13,15 @@ namespace Umbraco.Core
             => services.Replace(ServiceDescriptor.Singleton<TService, TImplementing>());
 
         /// <summary>
-        /// Registers a unique service as a single instance implementing two interfaces.
+        /// Registers a singleton instance against multiple interfaces.
         /// </summary>
-        /// <remarks>
-        /// Hat-tip: https://stackoverflow.com/a/55402016/489433
-        /// </remarks>
-        public static void AddUnique<TService1, TService2, TImplementing>(this IServiceCollection services)
+        public static void AddMultipleUnique<TService1, TService2, TImplementing>(this IServiceCollection services)
             where TService1 : class
             where TService2 : class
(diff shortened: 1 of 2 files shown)


rustybox:
@bergmania @AndyButland good 2 merge?

bergmania:
Thanks @rustybox

rustybox:
I do wonder if this is worse than original implementation of RegisterUnique

```csharp
composition.RegisterUnique<TImplementing>();
composition.RegisterUnique<TService1>(factory => factory.GetInstance<TImplementing>());
composition.RegisterUnique<TService2>(factory => factory.GetInstance<TImplementing>());
```

Imagine someone comes along now and re-registers TService1 but with a concrete that doesn't implement TService2, everything will seem fine but the cast will be invalid when resolving TService2.

On the other hand, I imagine this extension is rarely used except when both interfaces are tightly coupled together as hinted by with Andy's issue where calls to one was expected to affect the other. I think this suggests that the interface boundaries are weird, if a method call on TService1 should affect state on TService2 perhaps the two operations should be as part of the same interface.

In short I don't really know what's right and wrong here but there is a potential for confusion later if anyone ever replaces TService1 without replacing TService2

bergmania:
Good points @rustybox .

I don't like the idea about merging the interfaces, however, I think i should be pretty easy to avoid this, if we just split the implementation in two.

E.g.
I think it is simple to split `AspNetCoreSessionManager` into a new `AspNetCoreSessionManager` wich does not implement `ISessionIdResolver` and a `AspNetCoreSessionIdResolver` that implements `ISessionIdResolver`.

If required we can inject `ISessionIdResolver` into `AspNetCoreSessionManager`. But I'm not sure this is even required. The main reason this is just one implementation is that it basically needs the same dependencies, IIRC.

The same approach can be done with the other 2 types where we use `AddMultipleUnique`

rustybox:
@bergmania In the short term do you think we need to revert to how it was previously where concrete is registered or do you think it is relatively safe as is?

bergmania:
I think it is relatively safe as is.

Issue #9254, opened 2020-10-22
property editor decimal does not behave as expected on step size =< 0 | and when minimum > maximum


renaa:
### Bug summary

decimal step size less than or equal 0 makes it impossible to save content, the same is true if minimum value > maximum value

## Umbraco version

I am seeing this issue on Umbraco version: 8.8

Reproduction
------------

1. Create a new data type property editor:  Decimal 
1. set min and max values 
1. set step size: 0
1. add it to a document type
1. try to set it in content editor
1. save and watch it disappear

### Specifics

The buildt-in property editor; Decimal discards values that does not adhere to the maxvalue, step size and max value 
![image](https://user-images.githubusercontent.com/1050133/96920115-0a525e00-14ad-11eb-858f-56f0ffab3963.png)

here only 8, 8.5, 9, 9.5, and 10 would be possible to input in the content editor.
Other values are unset.

If the value is unset all decimal values between 8 and 10 are possible to input in the content editor.

When step size is zero or negative then no values are possible to input in the content editor.

I think that in this scenario it should rather be better to reset the value of **step size** in the type definition if its value is 0 or negative

The same is true when the minimum value is more than maximum. 

### ideas for solving

One could possibly reuse some of the logic that unsets the value in the content editor.
One could also check for if minimum > maximum and reset one or both of these if that is the case.

bjarnef:
I noticed this as well where I needed to save X and Y coordinates between 0 and 1. When a specified step size e.g. 0.01 I couldn't save the data. For now I have leaved the step size blank in the configuration of the Decimal datatype.

nathanwoulfe:
Hi @renaa and @bjarnef 

I've been fiddling with the editor, have a couple of questions:
 - how would a step size <= 0 be expected to work? I can't see how that would ever allow valid data
 - similar with min > max - I think that should be caught when setting config values so it never actually happens
 
I've tried with step = 0.5 and step = 0.01 on latest v8/contrib and it works fine - can save data so long as it matches the step size. 

Could we please verify this is still an issue on latest? That said, there's no UI feedback when the item is invalid - if I enter 1.5 when the max is 1, the value disappears on save - instead, the user should see feedback that the value is invalid and be able to fix it and re-save.

bjarnef:
@nathanwoulfe in my case I have configurated the Decimal datatype instance like this.

[...]

PR #9334, opened 2020-11-02
Decimal step size validation

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9254

### Description
This PR fixes some of the issues I have described in https://github.com/umbraco/Umbraco-CMS/issues/9254 when having specified a step size.

For example the following configuration.

image

The Decimal datatype is useful for e.g. latitude and longitude properties... or X an Y coordinates.

In my original use-case I am using this for X and Y-coordinates stored on an image map where the values would be between 0 and 1.

However with a step size of 0.1 the `<input type="number" />` expect a number like `0, 0.1, 0.2,.. ` etc..  but not `0.01` and `0.05` in this case the decimal datatype just save an empty or null value (at least when the field is not mandatory).

With this PR it validate the step size similar to the number validation.
In fact it is possible to accomplish this feature having a step size on 0.1 while accepting 0.01 to be saved using `ng-step`. In case `ng-step` is not specified, Angular will use `step` attribute for validation.
https://stackoverflow.com/a/45620178

The number input doesn't prevent entering e.g. alphanumeric characters or a value not matching the step size.
E.g. entering an `e` will show the number validation error on save and the same happens for an invalid step size. At the moment it seems the page is saved though and not showing an red validation error here . only briefly a validation error?

Currently this PR use `ng-step="model.config.step"` but maybe we always want to allow saving a decimal value with e.g. 3 decimals or more independent of step size?

By default in HTML5 validation it will show a validation error if the value doesn't match the specified step size:
https://www.w3schools.com/tags/att_input_step.asp


nul800sebastiaan:
That's some nice extra validation, thanks @bjarnef! 🏅

bjarnef:
@nul800sebastiaan maybe we should edit the title of this to something like "decimal range" and avoid "slider to not confuse with the Slider property editor?

PR #9335, opened 2020-11-03
Make models builder properties virtual to allow mocking

Matthew-Wise:
When trying to unit test using models builder models, we are unable to mock these without writing a lot of code - https://our.umbraco.com/forum/using-umbraco-and-getting-started/81058-unit-testing-moq-mocking-for-umbraco-models-builder-object

Making the properties virtual will mean we can mock them using something like Moq with no additional code. 

**Testing**
I have updated the unit tests to reflect the virtual property change.

Thanks
Matt


dawoe:
Awesome stuff. Will help writing unit tests a lot.

Matthew-Wise:
Checks failed look to be an issue with the build server not the code changes.

nul800sebastiaan:
Cool stuff @Matthew-Wise - thanks very much! 👍

Issue #8831, opened 2020-09-04
Add Stylecop Analysers and verbose .editorconfig to the .NET Core solution.


JimBobSquarePants:
Umbraco's codebase is extremely difficult to navigate even for experienced developers. The lack of code consistency and intellisense documentation leads to [easy to avoid mistakes](https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/). 

Porting the solution to .NET Core is a massive challenge, made much more difficult by this lack of consistency. 

I propose that this is the perfect opportunity to introduce linting tools that will help enforce code consistency and quality for a new codebase that will be the building block for a generation of development. 

Contrary to [previously expressed opinion](https://github.com/umbraco/Umbraco-CMS/issues/3752#issuecomment-441984646), introducing rules does not make it less easy to contribute. A strong set of rules aids with initial development, testing, documentation generation, as well as future development by reducing the number of bugs and increasing the readability of code.

Remember, code is read many more times than it is ever written.

I own a collection of [open source repositories ](https://github.com/SixLabors/) that deal with an incredibly complex space (2D graphics). For all that complexity there are comparatively few issues considering how many users there are. This is direct result of strong rules designed to help reduce mistakes in the codebase.

I have [rules](https://github.com/SixLabors/SharedInfrastructure) you can take that work well with separate rulesets for source and test repositories.

I implore you to take this request seriously. There are significant technical and operational cost savings that can be made which would beneficial to Umbraco as a company in addition to the open source libraries.

JimBobSquarePants:
Nothing is automatically applied.

soreng:
Installing Stylecop in the https://github.com/umbraco/Umbraco-CMS/tree/netcore/netcore version of Umbraco.Core, generates 19000+ warnings currently. I imagine at least 98% of them can be fixed by just formatting the code

JimBobSquarePants:
By formatting I'm assuming adding braces, intellisense docs, missing keywords, type declarations etc? Or do you mean whitespace?

soreng:
@JimBobSquarePants yes - the ones I looked at was of the kind you mention.

AndyButland:
I recall adding this for the first time to what I thought was one of my teams pristine code-bases and getting a nasty shock!... but once I'd mentally recovered from the offence caused - think I had to uninstall it and then put it back when I'd calmed down a bit - I've appreciated using it or similar tools from then on.

[...]

PR #9338, opened 2020-11-03
Add Stylecop and .EditorConfig Linting.

JimBobSquarePants:
### Prerequisites

No tests required.

See https://github.com/umbraco/Umbraco-CMS/issues/8831

### Description
Adds [Stylecop Analyzers](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/) and [.editorconfig](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019) linting to the <strike>Umbraco.Core project.  Integration is incremental and does not touch any additional projects. </strike>

Added to all projects now since rules have been discussed.

.editorconfig file is taken from https://github.com/RehanSaeed/EditorConfig

I've also included an example API documentation fix.


bergmania:
Hi @JimBobSquarePants ..

Good work :)
Do you know whether the increased build time is mainly due to the linting or printing of all those warnings?

I see there are lots of rules in the added `.editorconfig` that is completely different from what we used to do. I think we should at least change these to fit the current codebase. 
Also, there are some rules that I think are a waste of time to have.

Some examples are:
- [SA1309] Field '_fieldName' should not begin with an underscore
- [SA1633]The file header is missing or not located at the top of the file.
- [SA1124] Do not use regions
  - Even that I agree in general, as I see regions as only a must if you files are too big, it don't make sense in the current codebase.
- ...

JimBobSquarePants:
You will get increased build times currently but that’s an easily solvable problem (and also the wrong metric to be concerned about) I.E Fix the issues and improve the read quality of the codebase. (These are easy PR wins)

I’ve just used the default stylecop rules taken from a maintained (linked) repo. Each rules has been well discussed for years. Your existing ruleset is almost empty. 

I strongly disagree re use of underscore. That’s a bad hangover from other languages that do not posses the `this` keyword to define scope. It hurts readability since all your fields now start with the same character and I cannot tell static from non static fields. 

The copyright, take it or leave it but I strongly recommend it as it ensures the copyright is present on files leached by external sources plus nobody ever reads LICENSE.md

Regions are frowned upon for a reason. They generally indicate classes containing mixed concerns. 

Most issues can be fixed per document with few keystrokes.

dawoe:
Hi @JimBobSquarePants 

I see some conflicting rules here. In editor config it's configured to place using statements inside the namespace (https://github.com/umbraco/Umbraco-CMS/pull/9338/files#diff-aa08f7bb6975923223f3119baa0b7632663c3cb121089436e5dc4b2a659d5073R180) . 

While in the stylecop linting it says outside. (https://github.com/umbraco/Umbraco-CMS/pull/9338/files#diff-d611d21594027248550ec7f3313bf3eee7a7dcf373a42c2113896b01a8c52433R5)

Is this intended ? Or am I missing something ?

Dave

JimBobSquarePants:
Yeah, I know about this. I didn't want to change it as yet before checking with someone who uses VSCode to see what takes precedence to see whether I actually needed to make a change.

>I see some conflicting rules here

Some... or one.

bergmania:
@JimBobSquarePants FYI, I scheduled an internal meeting for the core team member from HQ to discuss this topic. Don't expect an answer before 23rd November.

JimBobSquarePants:
@dawoe I updated the config to match.

JimBobSquarePants:
I've just pushed some extra configuration to reduce some of the ruleset noise that I consider unhelpful. 

All the custom rules are based upon the years of experience working on the Six Labors projects.

bergmania:
Hi @JimBobSquarePants..
 We now discussed this in the Core team.

Overall we are happy with the PR, but there are some rules we need to be modified.

- [SA1309] Field '_fieldName' should not begin with an underscore
   - We need this changed to force users to use "\_fieldName" and not "this._fieldName". I know you hate this, but we all have preferences for "\_fieldName". 😈
- [SA1633] The file header is missing or not located at the top of the file.
  - If this rule should be included, we need the header to be added automatically when new files are created/committed. Similar functionality should exist for  https://github.com/apache/lucenenet.

JimBobSquarePants:
Table flip gif

JimBobSquarePants:
@bergmania I just had a look at Lucene.NET and it doesn't automatically create the copyright text on file creation.

Copyright can be added using <kbd>Ctrl + .</kbd> with the setup I have the same as any rule fix so I don't really see the issue here.

bergmania:
> Regarding usage of this. is it being dropped only for fields or for for methods, event, properties and methods?

Can you elaborate what you mean here?

JimBobSquarePants:
@bergmania Yeah...

Rules are separate for each fields, event, properties and methods. You guys specified fields but not the others. 

As such, I've specified not to use `this.` anywhere but kept it as a suggestion so if there is a situation where it is somehow required you can use it without generating a warning. 

I've implemented rules for both instance and static private fields.

- Instance starts with `_`.
- Static starts with `s_`

I copied the rules from Roslyn. They're the same as the runtime also.

bergmania:
Normally we just use "_" for static fields also. 

I don't think we have any private events? Public events seems to use pascal case.
I have never seen methods prefixed with "_".

JimBobSquarePants:
>Normally we just use "_" for static fields also.

I think it's very important to differentiate between static and instance fields. You can avoid a series of bugs that way (_normally covered by using `this.`_)

>I don't think we have any private events? Public events seems to use pascal case.
I have never seen methods prefixed with "_".

For events and methods I meant to differentiate between instance and static methods via `this.` The rules have no effect on naming of those types. Since you're not using `this.` for fields I think it would add complexity to use it in some places and not others. 

If only there was a specific keyword baked into the language which explicitly removed all this naming hackery eh? 🙄

JimBobSquarePants:
Ok... I managed to get this all working without any solution restructuring. Src and Tests now have separate rulesets with a much relaxed set in tests. 

I'm keen to get this in and move on.

linting/.editorconfig
```diff
@@ -135,9 +135,9 @@ dotnet_style_prefer_simplified_interpolation = true:warning
 [*.{cs,csx,cake}]
 # Implicit and explicit types
 # https://docs.microsoft.com/visualstudio/ide/editorconfig-language-conventions#implicit-and-explicit-types
-csharp_style_var_for_built_in_types = true:warning
+csharp_style_var_for_built_in_types = never
 csharp_style_var_when_type_is_apparent = true:warning
-csharp_style_var_elsewhere = true:warning
+csharp_style_var_elsewhere = false:warning
```
bergmania: Why are these rules changed?

bergmania (commented):
What happened to the thing about only applying the rules for one project at a time?

Also, why did you introduce the new rules related to `var` keyword :)?

linting/.editorconfig
JimBobSquarePants: I’ll dig out the original issue on Roslyn but essentially there’s a bug which means it allows `var` when the type is not apparent on the right hand side.

JimBobSquarePants:
Regarding individual projects I just thought “what’s the point?” If we have rules we should enforce them. Doing so piecemeal will lead to revisit the same code over and over again. I think it’s best to enforce rules on any code added anywhere to the repo.

bergmania:
Still some issues with the naming rules: 
image

JimBobSquarePants:
Odd... that shouldn’t be happening. I’ll have a look in a mo.

JimBobSquarePants:
I'm not seeing that. What extensions do you have running? I'm running a clean VS 16.8.2 with no extensions installed. 

image

JimBobSquarePants:
I can see it in Rider so I've asked on Twitter because that's broken behavior. The setup I'm using is all standard stuff. 

https://twitter.com/James_M_South/status/1332280929772056576

bergmania:
I guess the issue is because rider uses the `src/.editorconfig` and not the`linting/.editorconfig`

JimBobSquarePants:
Fixed it! Rider didn't seem to recognize the private instance field rule override.

JimBobSquarePants:
>Why do both editorconfig files claims to be root?

They don't actually. The config file is copied from the linting folder to root in a[ pre build task](https://github.com/umbraco/Umbraco-CMS/pull/9338/files#diff-50e91c82311ea26f2a73202525dfdf2b0a89c178ee666b2bf3ad4c84ac4c06dcR13-R24) so it can be read. The `EditorConfigFiles ` element seems to still be a little flaky with some rules not applying properly (namespace sorting)

Ideally all these files would simply sit in the same root directory as the `.sln` file but that can be done at a later time when we reorganize the structure.

bergmania:
I can confirm it works in Rider 🚀 

I personally don't like the rule to not use `var`. I agree improved readability when you read code outside of an IDE,  if people still name their variables good - Otherwise, I think that rule makes the codebase harder to read.

Overall I agree with the statements here: http://resharper.blogspot.com/2008/03/varification-using-implicitly-typed.html

That said, I know this is spaces vs tabs vol. 3. -I can live with that change if @Shazwazza also think that rule is nice to have.

@Shazwazza, please checkout this PR and see if you can live with the rules - Then please comment or merge. 💪

JimBobSquarePants:
I'm not sure you're interpreting the `var` rule correctly. `var` is very much allowed as long as the type is apparent in the declaration.

The rule works like this:

```c#
var index = 0; // Bad. Should this be an int or a long? There's potential for overflow if not declared correctly.

Foo myfoo = new MyFoo() // Bad. Uneccessary duplication of type.

var buffer  = new byte[256]; // Good. Type is apparent.

var result = query.ToArray(); // Good. Type is apparent.
```

I think that's a pretty good mix. You have consistency of approach with a neat terse syntax.

bergmania:
I just see some examples like this, where I like `var` more. Also in IDEs, we have help if needed to show the type.
image

But as I said, I can live with it even that I do not like it.

JimBobSquarePants:
You do know that method is called `List_Ambiguous_Events` yeah? 👅

JimBobSquarePants:
Bump

Shazwazza:
This is fantastic! I actually really like this and have never actually seen style cop and these rules all working together with VS automatically being able to fix with ctrl + .

There's a boat load of squiggly lines - but that's a good thing! I stopped using r# a long time ago and VS seemed to lack a lot of the explicit rule requirements but as it turns out that was because the solution was lacking this config.

Well done and thanks for the effort this will be a great addition to the solution. I'm also a fan of `var` but happy to give this a try, seems like we've reached some compromises here :P

Shazwazza:
One question though - multi-line comments. So does this mean that anytime we have multi-line comments they need to be with `/* */` syntax, else its chirping at us that single line comments need to have a line feed before them.

Shazwazza:
Ok I have a legitimate one 😂 

It is enforcing us to have xml comments for all method and ctor parameters. You can ctrl + . to generate a default constructor text but it doesn't fill in any params. It will take forever to fill all of these in each time and they bring no value, for example, i filled in all this to get rid of the squigglies. It generated the default message which is fine but the rest doesn't really provide any value. Can we configure it to not prompt for parameters?

```xml
        /// <summary>
        /// Initializes a new instance of the <see cref="ConfigureBackOfficeCookieOptions"/> class.
        /// </summary>
        /// <param name="serviceProvider">The <see cref="IServiceProvider"/></param>
        /// <param name="umbracoContextAccessor">The <see cref="IUmbracoContextAccessor"/></param>
        /// <param name="securitySettings">The <see cref="SecuritySettings"/> options</param>
        /// <param name="globalSettings">The <see cref="GlobalSettings"/> options</param>
        /// <param name="hostingEnvironment">The <see cref="IHostingEnvironment"/></param>
        /// <param name="runtimeState">The <see cref="IRuntimeState"/></param>
        /// <param name="dataProtection">The <see cref="IDataProtectionProvider"/></param>
        /// <param name="requestCache">The <see cref="IRequestCache"/></param>
        /// <param name="userService">The <see cref="IUserService"/></param>
        /// <param name="ipResolver">The <see cref="IIpResolver"/></param>
        /// <param name="systemClock">The <see cref="ISystemClock"/></param>
```

JimBobSquarePants:
I got you fam.

Uncheck SA611 in the ruleset and save. 

image

You can then delete the param doc xml.

image


I'd try and reduce all those dependencies though... That's a lot!

PR #9340, opened 2020-11-03
Track ng-repeat with id instead of label

### LINKED ISSUE #9291: Umbraco 8.7.0: translation causes error when opening CMS member
When I open a random member in Umbraco 8.7.0 the page woudn't load and shows an console angular error instead. Looks like it has to do with the translations that are in place.

## Console error
`Error: [ngRepeat:dupes] Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: group in content.tabs track by group.label, Duplicate key: Eigenschappen, Duplicate value: {"id":0,"active":false,"label":"Eigenschappen","alias":"Generic properties","open":true,"properties":[{"label":"Id","description":null,"view":"idwithguid","config":{},"hideLabel":false,"validation":{"mandatory":false,"mandatoryMessage":null,"pattern":null,"patternMessage":null},"readonly":false,"id":0,"dataTypeKey":"00000000000000000000000000000000","value":["1264","2344b14e-dcad-4e48-aff4-aeb42d28ab16"],"alias":"_umb_id","editor":null,"isSensitive":false,"culture":null,"segment":null},{"label":"Ledentype","description":null,"view":"readonlyvalue","config":{},"hideLabel":false,"validation":`

## Umbraco version

### Bug summary
After debugging what the problem could be i've found out it has to do with duplicate names. 
Inside the nl.xml file you will find the key: **<key alias="properties">Eigenschappen</key>**
This causes errors at my end, changing the value of the key will solve my issue; **<key alias="properties">Something else</key>**

Reproduction
------------
1) Change the language of CMS user to Dutch (don;t know if other languages will also show this bug)
2) Login as Dutch user
3) Open member inside CMS
4) View log (see screenshot)

How to fix:
1) Open \umbraco\Config\Lang\nl.xml
2) change row 447: <key alias="properties">Eigenschappen</key> to a different value for example <key alias="properties">Instellingen</key>
3) recycle app pool
4) login with same (Dutch) user
5) view member again (see screenshot)

error
changed-value

SteveVaneeckhout:
### Prerequisites

If there's an existing issue for this PR then this fixes #9291 

### Description
The error is caused by multiple groups having the same label. I changed the loop to use `group.id` for tracking. I did a search to see if there were any other components using label for tracking and found `umb-tabbed-content` using it too.

To reproduce:
1) Set your backoffice language to Dutch - nl.
2) Go to Settings (Instellingen) -> Member types (Ledentypes)
3) Add another group called `Eigenschappen` and put a text field in it.
4) Create/Edit a member

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/content/umb-tabbed-content.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/content/umb-tabbed-content.html
@@ -1,6 +1,6 @@
 ﻿<div>
     <ng-form name="tabbedContentForm">
-        <div class="umb-group-panel" retrive-dom-element="registerPropertyGroup(element[0], attributes.appAnchor)" data-app-anchor="{{group.id}}" data-element="group-{{group.alias}}" ng-repeat="group in content.tabs track by group.label">
+        <div class="umb-group-panel" retrive-dom-element="registerPropertyGroup(element[0], attributes.appAnchor)" data-app-anchor="{{group.id}}" data-element="group-{{group.alias}}" ng-repeat="group in content.tabs track by group.id">
 
             <div class="umb-group-panel__header">
                 <div id="group-{{group.id}}">{{ group.label }}</div>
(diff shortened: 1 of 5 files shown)


bjarnef:
Maybe we should change this for media as well
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/media/apps/content/content.html#L2

nathanwoulfe (approved):
Thanks @SteveVaneeckhout, this looks fine to me 👍

nul800sebastiaan:
Looks good, thanks @SteveVaneeckhout! 👍

PR #9353, opened 2020-11-07
NetCore: Migrated database server registrar and messenger tasks into hosted services

AndyButland:
In this PR I've taken two more recurring background tasks and migrated them to hosted services:

- TouchServerTask
- InstructionProcessTask

These were extracted out of `DatabaseServerRegistrarAndMessengerComponent`, which has a few things going on in it, so this needs a careful look to make sure I've not broken or overlooked anything.  As far as I can tell in local testing though, it operates as it did before.

As part of this I've also moved `DatabaseServerRegistrarOptions`, which holds various configuration settings as hard-coded values, into the usual configuration classes.  This isn't so much that we'd expect people to change them (though they could if they wanted), rather more to standardise this and make them injectable via `IOptions` into the hosted services.


src/Umbraco.Infrastructure/HostedServices/ServerRegistration/InstructionProcessTask.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        private readonly IRuntimeState _runtimeState;
+        private readonly IDatabaseServerMessenger _messenger;
+        private readonly ILogger<InstructionProcessTask> _logger;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="InstructionProcessTask"/> class.
+        /// </summary>
+        public InstructionProcessTask(IRuntimeState runtimeState, IServerMessenger messenger, ILogger<InstructionProcessTask> logger)
+            : base(TimeSpan.FromSeconds(5), TimeSpan.FromMinutes(1))
```
bergmania, line 22: This was not hardcoded before (`_registrar.Options.RecurringSeconds*1000`)

src/Umbraco.Infrastructure/Compose/DatabaseServerRegistrarAndMessengerComponent.cs
```diff
@@ -81,201 +75,48 @@ public override void Compose(Composition composition)
...
-
-            // must come last, as it references some _variables
+            // The scheduled tasks - TouchServerTask and InstructionProcessTask - run as .NET Core hosted services.
+            // The former (as well as other hosted services that run outside of an HTTP request context) depends on the application URL
+            // being available (via IRequestAccessor), which can only be retrieved within an HTTP request (unless it's explicitly configured).
+            // Hence we hook up a one-off task on an HTTP request to ensure this is retrieved, which caches the value and makes it available
+            // for the hosted services to use when the HTTP request is not available.
+            _requestAccessor.RouteAttempt += EnsureApplicationUrlOnce;
+
+            // Must come last, as it references some _variables
```
bergmania, line 98: I'm in double about whether this is required. The first request was more important in V8, because all the boot of umbraco happed on the first request. Not the server boots without a request.

bergmania (commented):
Just small findings.. But i'm in doubt about the importance of the first request, now that the server can boot without a request.

src/Umbraco.Tests.UnitTests/Umbraco.Infrastructure/HostedServices/ServerRegistration/InstructionProcessTaskTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private Mock<IDatabaseServerMessenger> _mockDatabaseServerMessenger;
+
+        [Test]
+        public async Task Does_Not_Execute_When_Runtime_State_Is_Not_Run()
+        {
+            var sut = CreateInstructionProcessTask(runtimeLevel: RuntimeLevel.Boot);
+            await sut.PerformExecuteAsync(null);
+            VerifyMessengerNotSynced();
+        }
```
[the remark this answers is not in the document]
AndyButland: Have added these additional cases (also applied to other hosted services tests verifying the same behaviour).

src/Umbraco.Infrastructure/HostedServices/ServerRegistration/InstructionProcessTask.cs
AndyButland: I think this effectively was, in that it came from `DatabaseServerMessengerOptions`, which currently is hard-coded.  But I think better anyway to move this to a configuration model, so if needed, it could be configured.  So I've done this.

For this one though I couldn't completely get rid of the original class - `DatabaseServerMessengerOptions` - as it also holds references to some callbacks.  Hence I've just renamed the class to `DatabaseServerMessengerCallbacks`, but perhaps some more refactoring there could be considered (i.e. does this need to be a separate class, or could the one property is has just be moved to `DatabaseServerMessenger`?

src/Umbraco.Infrastructure/Compose/DatabaseServerRegistrarAndMessengerComponent.cs
AndyButland: This is the bit I'm in doubt about too, so it does need a careful check.

I can say though we do seem to need this call to `EnsureApplicationUrlOnce`, which fires off some code on the first request.  In particular this calls `_requestAccessor.GetApplicationUrl()`.  Really we don't need to get it here, but by calling this method in the context of an HTTP request, it will find out and cache the application URL for other calls that are made that aren't in an HTTP request - i.e. hosted services.

`TouchServerTask` in particular won't run if it can't get the application URL, so it seems we do need to ensure this is set from the first request.

bergmania:
I tested, and I'm pretty sure this works as it did before. So I merged 👍 , Thanks again, again 💪

PR #9356, opened 2020-11-07
NetCore: MSDI refactor split CoreRuntime & Remove Current

rustybox:
Further progress for #8653

~~I tried to avoid splitting the Integration test projects but everything was a nightmare other than just splitting them.~~

CoreRuntime is now separate for Config vs Start - removed two usages of BuildServiceProvider, now we can just let container handle constructing DB factory for us.

Current (one of) is gone
Unsure of changes to TreeNode is it ever not a css class for icons?
Perhaps factory to construct from controller or handle fallback path in tree.service.js/umbraco.services.js?


rustybox:
@bergmania this covers pretty much all of my outstanding task list items.

CI job didn't run dotnet.exe test against the new lib

bergmania:
Hi @rustybox. Looks interesting. I will take a look, and test one of the first days in next week :)

rustybox:
Force push to drop unintended changes to sln file

rustybox:
Bit conflicted with this, on the one hand I think it's a good thing that integration test project is starting to get split.

I think the project would be much nicer with a tests/ directory filled with many csproj instead of monolith test projects with funky namespaces. If all netcoreapp/netstandard tests were in a tests/ directory it's easy to have CI run all test projects in that folder.

On the other hand I want to know what's going wrong with it, I had a few hours yesterday but didn't really get anywhere.

Massive stack traces of circular dependencies and hacks to work around them (usage of Lazy), call stacks that shift when a  debugger is attached, it's hard to know where to start.

Whilst tinkering I managed to spot a curious Examine.Lucene error message that could provide further clues with a bit more time, or it could just be a red herring.

rustybox:
Tests are so much slower now we guarantee components terminated, running again to check if Perform_Scheduled_Publishing failure reproducible, it passes fine locally.

rustybox:
Last commit passed but 2h44m is a bit mental to run the tests on DevOps.

We only need to terminate if CoreRuntime.Start is called so added a check and added comment for clarity.

src/Umbraco.Infrastructure/Trees/TreeNode.cs
```diff
@@ -106,20 +105,8 @@ public string IconFilePath
...
-                    return Current.IOHelper.ResolveUrl("~" + Icon.TrimStart('~'));
-
-                //legacy icon path
-
-                // TODO: replace this when we have something other than Current.Configs available
-                var backOfficePath = Current.GlobalSettings.GetUmbracoMvcArea(Current.HostingEnvironment);
-
-
-                return string.Format("{0}images/umbraco/{1}", backOfficePath.EnsureEndsWith("/"), Icon);
+                // TODO: Is this ever actually used? If not remove, if so, add setter.
```
bergmania, line 108: I think this is okay.. It is most likely not working all over backoffice, and we should use css icons anyway.

bergmania (approved):
Super work @rustybox..

I think it is okay to the way you handle `TreeNode.IconFilePath`.  I think it is legacy, and it should be possible to solve the same use cases by using CSS icons.

Nice work with the merge of the tests projects again. Regarding the slow integration tests. I'm looking into deciding the tests into multiple agents, to make the build a bit faster. On my local machine it takes 17 minutes to execute all tests, so I also think one reason for the agents to be slow, is because the LocalDB is lacking memory.

rustybox:
> Regarding the slow integration tests. 

Oh they are way faster since I switched terminate to only trigger if TestOptions.Boot or if base is UmbracoTestServerTestBase

Down from 2h44 to 1h28

41df94e  vs 756d5b

src/Umbraco.Infrastructure/Trees/TreeNode.cs
rustybox: Coolio, if it's not working I'll gladly add a factory that treenodecontroller can use that has deps on IOHelper etc

Issue #9357, opened 2020-11-07
Can't register multiple trees in same tree group


callumbwhyte:
There is supposedly 2 ways to register custom trees:

- Creating a controller that inherits from `TreeController` and decorating it with the `[Tree(...)]` attribute - this is then type scanned and auto-registered

- Creating a controller that inherits from `TreeControllerBase` and registering the tree in a composer with `composition.Trees().AddTree()`

A [comment](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Trees/TreeCollectionBuilder.cs#L24) on the latter method says "this is useful for ... a single tree controller for different tree aliases ... the tree controller cannot be decorated with the TreeAttribute".

However when using this method, with a shared tree controller, but more than 1 tree is registered within the same "tree group" in the same "section" this fails. A cast error is thrown as it is expected the tree will implement `TreeController` which it cannot in this case.

Tested on the latest 8.9.0, but looking at the code this likely applies to all versions of V8.

glombek:
Fixed in #9358

PR #9358, opened 2020-11-07
Casting tree controllers to TreeControllerBase

callumbwhyte:
As described in #9357 you get a cast error when trying to register multiple trees in the same section and tree group without using the `[Tree(...)]` attribute to declare it.

To create a tree without using the attribute you must inherit from `TreeControllerBase` and not `TreeController` - here the logic expects `TreeController` and so breaks in cases when not inheriting from it.

It transpires the cast causing the error is actually unnecessary, as the methods being accessed are part of `TreeControllerBase` and being inherited by `TreeController` anyway, so changing to cast to `TreeControllerBase` seems like a nice solution.


glombek (approved):
Thanks for the PR @callumbwhyte. This seems very sensible. Tested and working well.

Issue #9352, opened 2020-11-06
Add pagination to icon picker


bjarnef:
Currently the icon picker is one long list of icons, but when having many icons in can be slow. I have previous noticed this when installing font packages, which have a lot of icons, e.g. some of these icon packages:

https://our.umbraco.com/packages/backoffice-extensions/material-design-icon-pack/
https://our.umbraco.com/packages/backoffice-extensions/fontawesome-icon-pack/
https://our.umbraco.com/packages/backoffice-extensions/line-awesome-icon-pack/

With the new feature to include SVG icons, I wonder if we should add pagination here, e.g. 100 per page.

I found this with some free SVG icons:
https://cssauthor.com/free-svg-icons/

One of the mentioned is https://iconhub.io

Performance using angular `ngRepeat` directive can be optimized with `track by` but in my experience when having e.g. 500-1000+ repeated entries in DOM, it is slowing down depending on the browser.

bjarnef:
@MMasey @nielslyngsoe have you noticed issues with icon picker when having e.g. 1000 or 2000 icons? At the moment I think we only have a single method in C# fetching all icons?

bjarnef:
If grabbing all icons from here https://iconhub.io and placing these inside `/Umbraco/assets/icons` in Umbraco v8.9.0 (or in `/ App_Plugins/MyPlugin/Icons` folder in `v8/contrib` branch, and doing a search in icon picker, the search is being slow and sometimes I get this, because it use too much memory.

![image](https://user-images.githubusercontent.com/2919859/98438717-58da3c00-20ec-11eb-8e53-6454edafa092.png)

MMasey:
Hey @bjarnef & @nul800sebastiaan, I've been having a think about this and I think we need to solve a number of other performance problems first, before we consider pagination. I do think some categorisation would be cool though.

I think the memory issues are primarily affecting the frontend due to the following reasons.

- When using the filtering, there doesn't seem to be an deferring so each time a character changes there is a paint reflow, a really big one too
- I think the reason the rendering takes so long is because I when the ng-repeat does the filter it added each of the icons individually. This isn't much of an issue when there are only a few icons returned by the filter, but if there are loads it completely clogs up the thread.
- Each of the SVG's themselves are also lazy loaded. This adds then another reflow when they get loaded in, so I wonder if maybe for the full icon listing we don't lazy load?

[...]

PR #9359, opened 2020-11-08
Add pagination to icon picker

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9352

### Description
When having a lot of icons, either font icons of SVG icons, the icon picker can be slow and cause memory issues, because a lot of icons is rendered and filtered in the DOM. Instead we should fetch a paged list of icons, e.g. 100 per page.

### DIFF
--- a/src/Umbraco.Core/Services/IIconService.cs
+++ b/src/Umbraco.Core/Services/IIconService.cs
@@ -13,9 +13,15 @@ namespace Umbraco.Core.Services
         IconModel GetIcon(string iconName);
 
         /// <summary>
-        /// Gets a list of all svg icons found at at the global icons path.
+        /// Gets a list of all svg icons found at the global icons path.
         /// </summary>
         /// <returns></returns>
         IList<IconModel> GetAllIcons();
+
+        /// <summary>
+        /// Gets a paged list of svg icons found at the global icons path.
+        /// </summary>
+        /// <returns></returns>
+        IList<IconModel> GetPagedIcons(int pageNumber, int pageSize, string filter = "");
     }
 }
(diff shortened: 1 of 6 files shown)


bjarnef:
While working at this PR I noticed we are not consistent about how to handle the following condition. Sometimes it returns an NotSupportedException, HttpException and sometimes an empty page result.

```
if (pageSize <= 0 || pageNumber <= 0) {

}
```

https://github.com/umbraco/Umbraco-CMS/blob/e9627f2ee962a73cd8c007a42845c9129e3339ee/src/Umbraco.Web/Editors/RelationTypeController.cs#L103-L106

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Editors/EntityController.cs#L592-L595

Also here it should return the requested `pageNumber` and `pageSize` in the empty paged result instead of `0`.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Editors/EntityController.cs#L610-L614

We can look at this in another PR.
@nul800sebastiaan @Shazwazza is this something we can change and should we handle this in a standard way?

bjarnef:
A challenge with this PR to get a paged result is that we at the moment have a second request inside the `getAllIcon()` to get legacy icons using `iconHelper.getIcons()` and concatenate these.

At the moment the icon picker is however causing memory issues, when having e.g. 1000-2000+ icons (items), which is something I have previous noticed when installing font icon packages with many icons:
https://our.umbraco.com/packages/backoffice-extensions/material-design-icon-pack/

There is a suggestion here to use `limitTo` and a load more approach:
https://stackoverflow.com/a/17350875

nul800sebastiaan:
I don't think paging is a great experience, maybe categories ("core", whatever we ship with, "blabla set", a new named set, etc). 

We'll need to discuss first, we'll get back to you.

bjarnef:
It could still be served as paged results, so it doesn't render e.g. 2000 DOM elements. We could either use pagination or infinite loading / load more.

Even with categories each of these could potential have many icons.

PR #9361, opened 2020-11-09
Limit access to logviewer endpoints to people with access to the settings section

nul800sebastiaan:
The access was a bit too broad before.


clausjensen (approved):
Looks good to me 👍

PR #9367, opened 2020-11-09
Ensure that upload file(s) in the MediaPicker is returned

mcl-sz:
The onUploadComplete function returns the last files that are added to a mediafolder. While this works correct by default and in most situations, it doesn't work as expected when a diffrent sorting is used for Media-items. For example, we've added events to sort Media-items automatically by name alphabetically when they are created/uploaded to keep them better organised. 
By sorting the $scope.files array by the Id-property, it ensures that the function returns the uploaded files, instead of the last files in the folder.

### Prerequisites


nul800sebastiaan:
Ah.. never thought of that! Clever stuff, thanks for the fix @mcl-sz and apologies for the delay on this one! 😅

Issue #9351, opened 2020-11-06
Block list editor shows validation error on init


bjarnef:
In Umbraco v8.9.0 when the block list editor is configurated to require minimum 1 item, it shows the validation error on init and not just on submit as previous.

![image](https://user-images.githubusercontent.com/2919859/98345391-e77b8a00-2014-11eb-8d6b-c38cbd4ddf05.png)

OwainWilliams:
Hi @bjarnef - Thanks for raising this issue. 
When you say 'as previous' do you men v8.8.x ? 

Thanks, 
O.

OwainWilliams:
This issue has been discussed at Umbraco HQ and the message is there on purpose, however, it probably should be a bit more subtle until it is an actual error. Just now it looks like there's an error state to begin with. 

The Multinode Tree Picker does it correctly, it gives a hint before validation and turns red after validation fails.
Same with the textbox character count, there's a hint in black until you go too far.

I'll mark this as up for grabs if anyone wants to make the Block list editor work similar to the MNTP.

bjarnef:
@OwainWilliams as I recall the validation message in v8.7 and v8.8 didn't show on submit, but first when the page/form was submitted.

It is probably fine to show it as a hint, but not as an error until trying saving the page.

umbrabot:
Hi @bjarnef,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
I have checked with latest changed in `v8/contrib` branch using MTNP and Multi URL Picker directly as property editors.

![image](https://user-images.githubusercontent.com/2919859/98584362-10767600-22c6-11eb-9f33-89f9e318c842.png)

I'll have a look at how it looks inside Block List editor.

bjarnef:
It seems the block list property editor by default add the class `text-error` class - that's why the text is red on init.

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umb-block-list-property-editor.html#L42
[...]

PR #9368, opened 2020-11-09
Block list editor validation

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9351

### Description
This PR makes some adjustments to Block List editor, so validation message first is shown on submit as with Content Picker property editor (used for both content picker and MNTP) and Multi URL Picker property editor.

The Block List editor also shows the hint messages like the other property editors.

I think it would be great if the localized message wasn't splitted in mutliple keys but a single key using a token or handled in controller to deal with the singular/plural message.

Maybe we could even have a shared directive/component or some angular templates to handle this and easier to re-use in the different property editors. Perhaps an option to overwrite the default message for each of these validation messages?

For now I think this is more consistent with e.g. Content Picker, MNTP and Multi URL Picker.

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umb-block-list-property-editor.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umb-block-list-property-editor.html
@@ -26,11 +26,11 @@
             </div>
         </div>
 
-        <button ng-if="vm.loading !== true"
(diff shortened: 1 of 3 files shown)


bjarnef:
@nielslyngsoe I have submitted this PR to make validation in Block List editor a bit more consistent with other property editors.

PR #9369, opened 2020-11-09
Add rel="noopener" to anchor elements with target="_blank"

bjarnef:
### Prerequisites

### Description
For security reasons anchor elements with `target="_blank"` should have `rel="noopener"`: https://web.dev/external-anchors-use-rel-noopener/


mikecp:
Hello @bjarnef ,
Thanks for the PR, I'll have a look at it :-)

mikecp:
Hello @bjarnef ,

I have started reviewing your PR and I noticed that there is kind of a tricky part in some files: the text you have updated is in fact contained in a <localize> tag, which means that the text that you updated in the file is actually the "fallback" text in case the label is not found in the language file.

So, ideally, in those cases, the same update should be performed in the language files as well. Would it be possible for you to make those changes in there too, so that all languages also benefit from this improvement?

Thanks in advance 👍

bjarnef:
@mikecp I have fixes this in the translations files.

I have also made some changes to link to https and umbraco.com (instead of umbraco.org) to be consistent.

Furthermore I noticed this link in the translation file is broken:
https://umbraco.tv/documentation/videos/for-site-builders/foundation/document-types

We should either update this to something else or there should be added a redirect for this on umbraco.tv.
It would probably be this page now:
https://umbraco.tv/videos/umbraco-v8/implementor/fundamentals/document-types/what-is-a-document-type/

bjarnef:
I have also made some updates for the `databaseNotFound` translation key. For German it wasn't using `CDATA`. However it doesn't seem this in being used anymore in core, so we want just want to remove it sometime, but it would be a different PR to cleanup unused translation keys.

georg-jung:
I noticed the same and opened #9403 before I saw this PR, sorry. Reading through https://rules.sonarsource.com/html/type/Vulnerability/RSPEC-5148 it might be better to use `rel="noopener noreferrer"` instead of just `rel="noopener"`. What do you think?

bjarnef:
@georg-jung the `noreferrer` value in `rel` attribute is mostly for older browsers (like IE11) which doesn't support `noopener` https://caniuse.com/?search=noopener

Not sure it is worth to implement since Umbraco v8 has dropped support for IE11 (although most of the backoffice might work). All other modern browsers support `noopener`.

georg-jung:
I think it could still be interesting to follow that practice, i.e. so that static analysis runs don't complain. Do you think there are any downsides to implementing it?

abjerner:
Adding `noreferrer` also has the plus side of the browser not sending a `Referer` header in the request triggered by the link. As a result of this, IMO all links linking out of the backoffice should have `noreferrer`. Not so much because of security, but because of the anonymity it adds (eg. the target page won't be able to see your secret backoffice URL).

IIRC `noreferrer` also implies `noopener`, so either of `rel="noopener noreferrer"` and `rel="noreferrer"` should be good.

mikecp:
Hi @bjarnef , 
Thanks a lot for your changes so far 👍 
Based on the interesting conversation with @georg-jung and @abjerner, would it be possible for you to add the "noreferrer" part as well? That would be ace :-)

bjarnef:
@mikecp not sure it is neccesary regarding this and I think it would be a separate PR if we want to change that which also mean the Referer header is not sent.

mikecp:
Hi @bjarnef ,

My apologies for reacting late, I checked with HQ and they are happy with noopener, because most links are in fact internal (to our, umbraco TV etc.), except maybe on the file _src/Umbraco.Web.UI.Client/src/views/packages/views/repo.html_ which contains links to potentially unsafe/uncontrolled URL's. 

Would you mind just adding both "noopener" and "noreferrer" in that single file? Then we'll be all set for a nice merge :-)

Thanks!!

bjarnef:
@mikecp I have added both "noopener" and "noreferrer" in the packages repo view.

mikecp (approved):

mikecp:
And that's a merge :-)

Thanks @bjarnef for this great PR! 👍

PR #9370, opened 2020-11-10
Anonymous telemetry

warrenbuckley:
## Feature
For new installs and for upgrades this will add a new telemetry marker file at `/umbraco/telemetrics-id.umb` with a unique GUID to identify the site.

Then this is pinged home every 24 hours to send us a request. The use of this data is to help track the active number of installs in a more accurate way.

## Test Notes
* Verify marker file is created for you in a clean install and also an upgrade with a valid GUID as its contents
* Attach debugger & boot site
* Note with the debugger attached it will send the telemetry request to the DEV RainbowsRocks service
* Verify the request is sent to the service and get back a success status code.

## Things to consider
* Removing the marker file will stop the service from calling home
* Changing the contents of the marker file to be a non-valid GUID will stop it from calling from home
* If a user has decided to delete the marker as wanting to opt-out then the next time they upgrade the file will be re-added back


warrenbuckley:
Not sure what was up with the unit tests - but can you review this please @bergmania

src/Umbraco.Web/Telemetry/ReportSiteTask.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            try
+            {
+                // Send data to LIVE telemetry
+                _httpClient.BaseAddress = new Uri("https://telemetry.umbraco.com/");
+
+//#if DEBUG
+//                // Send data to DEBUG telemetry service
+//                _httpClient.BaseAddress = new Uri("https://telemetry.rainbowsrock.net/");
+//#endif
```
bergmania, line 88: Is this correct?  Why is this commented out?

bergmania (changes requested):
Hi @warrenbuckley.. I only found small issues.. Please take a look

src/Umbraco.Web/Telemetry/ReportSiteTask.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                // Open file & read its contents
+                // It may throw due to file permissions or file locking
+                telemetricsFileContents = File.ReadAllText(telemetricsFilePath);
+            }
+            catch (Exception ex)
+            {
+                // Silently swallow ex - but lets log it (ReadAllText throws a ton of different types of ex)
+                // Hence the use of general exception type
+                _logger.Error<ReportSiteTask>(ex, "Error in reading file contents of telemetry marker file found at '{filePath}'", telemetricsFilePath);
```
Shazwazza, line 65: There is no exit here, the logic will continue below even though the file hasn't been read. It will fail on the GUID parse but better to not rely on logic below when you know you cannot continue?

src/Umbraco.Web/Telemetry/ReportSiteTask.cs
```diff
@@ -0,0 +1,128 @@
...
+        private IProfilingLogger _logger;
+        private static HttpClient _httpClient;
+
+        public ReportSiteTask(IBackgroundTaskRunner<RecurringTaskBase> runner, int delayBeforeWeStart, int howOftenWeRepeat, IRuntimeState runtime, IProfilingLogger logger)
+            : base(runner, delayBeforeWeStart, howOftenWeRepeat)
+        {
+            _runtime = runtime;
+            _logger = logger;
+
+            if (_httpClient == null)
```
Shazwazza, line 29: It will always be null in a ctor

src/Umbraco.Web/Telemetry/ReportSiteTask.cs
```diff
@@ -0,0 +1,128 @@
...
+                // The only time it will recheck when the site is recycled
+                return false;
+            }
+
+            try
+            {
+                // Send data to LIVE telemetry
+                _httpClient.BaseAddress = new Uri("https://telemetry.umbraco.com/");
+
+//#if DEBUG
```
Shazwazza, line 86: This #ifdef is commented out, so the BaseAddress gets reset

src/Umbraco.Web/Telemetry/ReportSiteTask.cs
```diff
@@ -0,0 +1,128 @@
...
+                // Send data to DEBUG telemetry service
+                _httpClient.BaseAddress = new Uri("https://telemetry.rainbowsrock.net/");
+//#endif
+
+                _httpClient.DefaultRequestHeaders.TryAddWithoutValidation("Content-Type", "application/json");
+                var postData = new TelemetryReportData { Id = telemetrySiteIdentifier, Version = UmbracoVersion.SemanticVersion.ToSemanticString() };
+                var request = new HttpRequestMessage(HttpMethod.Post, "installs/");
```
Shazwazza, line 93: this is IDisposable

src/Umbraco.Web/Telemetry/ReportSiteTask.cs
```diff
@@ -0,0 +1,128 @@
...
+                request.Content = new StringContent(JsonConvert.SerializeObject(postData), Encoding.UTF8, "application/json"); //CONTENT-TYPE header
```
Shazwazza, line 94: this is IDisposable

Shazwazza (changes requested):
Left some comments inline

src/Umbraco.Web/Telemetry/ReportSiteTask.cs
warrenbuckley: Yep makes sense - will just new it up & remove the if null check

It was a case of me looking at/following a similar use-case from KeepAlive background task
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Scheduling/KeepAlive.cs#L20-L29

src/Umbraco.Web/Telemetry/ReportSiteTask.cs
```diff
@@ -0,0 +1,133 @@
...
+        private IProfilingLogger _logger;
+        private static HttpClient _httpClient;
+
+        public ReportSiteTask(IBackgroundTaskRunner<RecurringTaskBase> runner, int delayBeforeWeStart, int howOftenWeRepeat, IRuntimeState runtime, IProfilingLogger logger)
+            : base(runner, delayBeforeWeStart, howOftenWeRepeat)
+        {
+            _runtime = runtime;
+            _logger = logger;
+            _httpClient = new HttpClient();
+        }
```
bergmania, line 29: Now this actually overrides the static _httpClient if you created a new ReportSiteTask. This this should just be initialized outside of ctor
```suggestion
        private static HttpClient _httpClient = new HttpClient();

        public ReportSiteTask(IBackgroundTaskRunner<RecurringTaskBase> runner, int delayBeforeWeStart, int howOftenWeRepeat, IRuntimeState runtime, IProfilingLogger logger)
            : base(runner, delayBeforeWeStart, howOftenWeRepeat)
        {
            _runtime = runtime;
            _logger = logger;
        }
```

PR #9372, opened 2020-11-10
Netcore: Misc Remove BuildServiceProvider call

rustybox:
Quick win, 1 less ServiceProvider


bergmania:
Nice one @rustybox. Thanks

Issue #9342, opened 2020-11-04
Not possible to replace image on media item


bjarnef:
In Umbraco 8.9.0 it doesn't seem to be possible to remove an existing file/image on a media item and replace it with a new one.

In "old days" you could remove the image, save the media item and upload a new one. However since "Upload image" property now be default is mandatory this trigger a validation error.

It used to be possible to remove the file from the media (without save) and select a new file and after this save the media item again.

bjarnef:
This doesn't seem to work via drop-and-drop, but it does work if selecting file via the click event as a regular file upload.
Has anything changes here regarding the drop-and-drop?

nul800sebastiaan:
You can fix this by making the upload field not mandatory. This was changed for this issue: https://github.com/umbraco/Umbraco-CMS/issues/4302

What does this have to do with drag & drop?

nul800sebastiaan:
I think the functionality is super awkward right now, from a quick chat with @nielslyngsoe we come to the following conclusion:

- we should keep the upload/cropper field on images as mandatory but default (but of course you can still update the media type and set it to not mandatory)
- when it's mandatory, always replace whatever was there before (delete previous file from disk), don't give users the option to select "Delete file(s)" - ![image](https://user-images.githubusercontent.com/304656/98693290-6273d600-2370-11eb-8143-8609b1edc9ff.png)
- when the property is optional, still show the option of "Delete file(s)" and allowing a saving that

In fact, I think the "Delete file(s)" button is a bit antiquated, it should be a nice delete icon with a good confirmation message that explains that deleting the picked file will also delete it from disk. Similar to what you get on, for example, a block element:

![image](https://user-images.githubusercontent.com/304656/98693800-05c4eb00-2371-11eb-8632-655f41cf5aaa.png)

![image](https://user-images.githubusercontent.com/304656/98693801-06f61800-2371-11eb-87d3-51b0eea2bcaf.png)

nul800sebastiaan:
I've tentatively marked this as breaking, as when the behavior changes, we always have some people who were relying on the old behavior and they will get surprised otherwise.

umbrabot:
Hi @bjarnef,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

[...]

PR #9376, opened 2020-11-10
Ensure it is possible to replace media via drag and drop

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9342

### Description
This PR make some changes to this previous PR https://github.com/umbraco/Umbraco-CMS/pull/8390 where it seems the `onFilesSelected` function in `umbPropertyFileUpload` component wasn't triggered after in drop event via drag and drop, but triggered via the change event, when clicking the button.

However with latest changes in `v8/contrib` I did need to change the save button in media back to `type="submit"` to save the media, which was changed in this PR https://github.com/umbraco/Umbraco-CMS/pull/9019

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbfileupload.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbfileupload.directive.js
@@ -10,25 +10,29 @@
 function umbFileUpload() {
     return {
         restrict: "A",
-        scope: true,        //create a new scope
+        scope: true, // create a new scope
         link: function (scope, el, attrs) {
             el.on('change', function (event) {
                 var files = event.target.files;
-                //emit event upward
+                // emit event upward
                 scope.$emit("filesSelected", { files: files });
-                //clear the element value - this allows us to pick the same file again and again
+                // clear the element value - this allows us to pick the same file again and again
                 el.val('');
-            });
-
-            el.on('drag dragstart dragend dragover dragenter dragleave drop', function (e) {
-                e.preventDefault();
-                e.stopPropagation();
+            })
+            .on('drag dragstart dragend dragover dragenter dragleave drop', function (event) {
+                event.preventDefault();
+                event.stopPropagation();
             })
             .on('dragover dragenter', function () {
                 scope.$emit("isDragover", { value: true });
             })
             .on('dragleave dragend drop', function () {
                 scope.$emit("isDragover", { value: false });
+            })
+            .on('drop', function (event) {
+                var files = event.originalEvent.dataTransfer.files;
+
+                scope.$emit("filesSelected", { files: files });
             });
         }
     };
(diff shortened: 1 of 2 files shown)


bjarnef:
@nathanwoulfe maybe you want to try this out? 😁 😎

nathanwoulfe:
@bjarnef might just be me, but I can no longer save media (on v8/contrib, so not related to these changes, but means I can't test this)

bjarnef:
@nathanwoulfe yes, I had same issue after this PR was merged https://github.com/umbraco/Umbraco-CMS/pull/9019

I had to revert that change back to submit button locally.

nathanwoulfe:
@bjarnef #9019 makes the button consistent with content (enter shouldn't submit) which is a positive change, so we really need to make this work with type="button". Noticing too the hotkeys aren't working, so might be evidence of a larger issue in the media edit controller?

nathanwoulfe:
Found it, I think - when type="button", it needs an action property, which will be `action="save()"`

bjarnef:
@nathanwoulfe regarding semantic and accessbility I do think it makes sense with a submit button for the primary action and it act as a page submit.

nathanwoulfe:
@bjarnef I'm split - I like the consistency with content, and submit-on-enter was causing issues, but I agree that submit is semantically correct. It should definitely be consistent, even if that means changes in content to make it work correctly.

Would be good to get @nielslyngsoe thoughts on this, given the changes might grow outside the scope of this PR.

nathanwoulfe:
@bjarnef I think the best way forward for now is to keep the button as type=button, to avoid undoing the change from #9019, and add the action attribute. We can then circle back around and consider the save buttons on content/media/members etc and determine the best way forward for all types.

Also needs to be fixed because the current v8/contrib branch will have non-saveable media, which ain't good. Tagging @kjac here too, since this relates to his changes.

bjarnef:
@nathanwoulfe @nul800sebastiaan I would be great to have this, so it is easy to replace and image via drag and drop. Currently we have to use the click event which trigger the regular file upload.
With the latest changed it seems the button element is of `type="submit"` again :)

nul800sebastiaan:
@bjarnef Sorry, where is `button type="submit"` please?

bjarnef:
@nul800sebastiaan it the latest changes, which is now conflicting with this PR.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/media/edit.html#L63

`action="save()"` added in https://github.com/umbraco/Umbraco-CMS/pull/9376/commits/fb059bfe6a8a3e11b96faffd6769f5fbddf727bc shouldn't be necessary now with a submit button :)

bjarnef:
@nul800sebastiaan it seems @nielslyngsoe meanwhile has submitted another PR with an alternative fix for this in https://github.com/umbraco/Umbraco-CMS/pull/9933

PR #9377, opened 2020-11-10
Ability to define accept file extensions for FileUpload and use these to determine Media Type for File Uploads

nielslyngsoe:
This PR adds the ability to define accepted file extensions for FileUpload and use these to determine Media Type for File Uploads.

Prepare:
Create new Media Type (named "Video") with new File Upload Data type for umbracoFile-property.
Set the accepted file extensions for this new data type to ex.: mov and mp4

Create another set that takes mp3 name it "Sound".

Test Notes:

General testing that uploaded files are made of the Media Type that fits the file extension.
Test creating new items through create menu, see the you can only pick files that matches the file extension.
Test that Media Type "File" still can pick any file.
Test dragging files of these new types get creates as the defined media type.
Check that image files that are not define by any media type are created as media.
Check that folders and other files still works as before.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbpropertyfileupload.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbpropertyfileupload.directive.js
@@ -85,7 +85,7 @@
 
         /** Called when the component has linked all elements, this is when the form controller is available */
         function postLink() {
-            
+
         }
 
         function initialize() {
@@ -186,7 +186,7 @@
                         });
                     }
                 }
-                
+
             }
         }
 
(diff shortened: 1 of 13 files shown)


skttl:
I've been wanting this for so long 😍

nielslyngsoe:
@skttl great! Do you know about other low hanging fruits, that would be great to get solved? This one was quite simple and I think it will help out editors a lot in their work. Especially with the new media picker coming up where we can limit the ability to pick certain media types.

If you like to peek:
https://github.com/umbraco/Umbraco-CMS/pull/9461

skttl:
I'll give it a spin later - looks promising!

nielslyngsoe:
@skttl cool, still in progress :-) but let us know what you think

emma-hq:
Hey Niels! I'm going to close this one as we have the functionality as part of mediapicker v3. I'm going to check for any functionality we don't already have that is in here and raise issues accordingly.

PR #9380, opened 2020-11-11
Netcore: Use DefaultControllerActivator

rustybox:
Use DefaultControllerActivator, don't register Controllers into ServiceCollection

#8653


rustybox:
## Reasons to merge
Appears to work fine

## Reasons not to merge

**This comment that used to be in Startup.cs but has since been moved and then removed**

TODO: We will need to decide on if we want to use the ServiceBasedControllerActivator to create our controllers
or use the default IControllerActivator: DefaultControllerActivator (which doesn't directly use the container to resolve controllers)
This will affect whether we need to explicitly register controllers in the container like we do today in v8.
What we absolutely must do though is make sure we explicitly opt-in to using one or the other *always* for our controllers instead of
relying on a global configuration set by a user since if a custom IControllerActivator is used for our own controllers we may not
guarantee it will work. And then... is that even possible?

**This snippet from WebInitialComposer (oh please let Umbraco.Web.Composing.Current disappear)**

TODO: This will depend on if we use ServiceBasedControllerActivator - see notes in Startup.cs
You will likely need to set DefaultRenderMvcControllerType on Umbraco.Web.Composing.Current
which is what the extension method below did previously.
.ComposeUmbracoControllers(GetType().Assembly)
.SetDefaultRenderMvcController</*RenderMvcController*/ Controller>(); // default controller for template views

rustybox:
@Shazwazza your insight would be good here.

Additional bit of detail from @bergmania 

> We ended up registering controllers and deriving them from the container. I do not remember the golden reason. But I'm pretty sure we need to get controllers from containers in a few different situations, so even if they are not necessarily returned from the container on requests, they still need to be registered.

bergmania:
We turned this on as part of the website rendering migration. So I'm sure that it is not a requirement from the backoffice (e.g.Trees).

I tested the PR and everything in backoffice also works as expected. 
I suggest we merge this and then figure out another way to solve the potential issues. I guess the issues will be related to the IRenderMVCController stuff and routing, which is an area we soon will look into

bergmania:
I resolved conflicts and will merge when buildserver is happy

PR #9381, opened 2020-11-12
Fixes #6089 JsDiff.diffWords in Rollback freezes browser

### LINKED ISSUE #6089: JsDiff.diffWords in Rollback freezes browser
After a deploy we somehow accidentally messed up our homepage. When trying to rollback to the last working version, the browser kept freezing up.

After diving into the code, I noticed that the point where the code started to hang was when it tried to call `JsDiff.diffWords(property.value, oldProperty.value)`, obviously used to display the changes between the two version.

However, we have one very big Grid on this page, resulting in an `oldProperty.value` size of 2.3MB! `JsDiff.diffWords` cannot handle this size and will freeze up (especially considering the `property.value` at this point in time was near-empty, so the percentage of difference was insane).

Eventually, I managed to skip the diffing by replacing the function with a stub in the console: `JsDiff.diffWords=function(a, b, c) { return undefined; };` and could successfully rollback the page.

I think there needs to be some sort of fail-safe in the whole diffing process. Either by truncating the data if it goes over a certain limit, or skipping it all together with a nice message why.

EDIT: Upon investigating, the Grid itself was not so big. It's all the stuff around it that virtually inflates the property value. The Grid itself contains like 10 sections and maybe 30 grid components in total. The actual property blows up in size because it's constantly repeating all the grid components in `$allowedEditors` (and we have a lot of them) and the full editor inside each placed grid component (and our grid components have a lot of configuration).

So besides `JsDiff.diffWords` needing a fail-safe, the value-format stored by the Grid datatype is HIGHLY inefficient, so that might be an issue all on its own.

nul800sebastiaan:
As described in #6089 a large-ish diff for the Grid (and also for any other prop editor that stores JSON data) will kill a browser tab if you try to roll back.

The suggestion there to have a look at `diffJson` was a good one and very easy to implement since we already detect items as being a json blob versus stored text. I tested this with 150 paragraphs of text from https://baconipsum.com/ and saved the node a few times. Trying to do a rollback definitely killed by browser tab. After this fix, selecting a version to roll back to is really fast.

This doesn't fix all of the problems, it is still slow when you have many words in a regular textarea/RTE/etc. So that's an improvement that could be made at a later stage by capping the amount of words to do a diff on or something else. I think for now this will help people enough to be useful though.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/rollback/rollback.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/rollback/rollback.controller.js
@@ -2,7 +2,7 @@
     "use strict";
 
     function RollbackController($scope, contentResource, localizationService, assetsService, dateHelper, userService) {
-        
+
         var vm = this;
 
         vm.rollback = rollback;
@@ -56,7 +56,7 @@
                 });
 
             });
-            
+
         }
 
         function changeLanguage(language) {
@@ -103,7 +103,7 @@
                             var timestampFormatted = dateHelper.getLocalDate(version.versionDate, currentUser.locale, 'LLL');
                             version.displayValue = timestampFormatted + ' - ' + version.versionAuthorName;
                             return version;
-                        }); 
+                        });
                     });
                 });
         }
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
This has been tested by a customer and found to be effective.

enkelmedia:
Great! Works a lot better than before! =D

PR #9383, opened 2020-11-12
Get CodeQL to only scan the relevant directories

nul800sebastiaan:
Turns out this was failing on at least the `node_modules` directory, so we'll try to ignore directories that are not relevant for building a release.


warrenbuckley:
Looks good to me @nul800sebastiaan

warrenbuckley (approved):

PR #9398, opened 2020-11-16
Speed up publish content query and reduce memory use by reducing array allocations for PublishedContent properties

nzdev:
### Prerequisites

### Description
What?
Change from using a list to using an array when we know the size of the array in advance.
Why?
Speed up publish content query and reduce memory use by reducing array allocations for PublishedContent properties.

How to test?
Existing tests pass.


nathanwoulfe (approved):
Hi @nzdev, makes sense to me 👍

nul800sebastiaan:
Looks good, thanks @nzdev !

PR #9401, opened 2020-11-16
Netcore: Turn container validation back on

rustybox:
Further progress for #8653

Kills RuntimeLevel attribute


rustybox:
@bergmania this closes off all outstanding items on todo list

312532e is a bit of a mess but more so due to the existing setup.

src/Umbraco.Web.Common/AspNetCore/AspNetCoreHostingEnvironment.cs
```diff
@@ -10,12 +10,12 @@ namespace Umbraco.Web.Common.AspNetCore
...
     public class AspNetCoreHostingEnvironment : Core.Hosting.IHostingEnvironment
     {
-        private IOptionsMonitor<HostingSettings>  _hostingSettings;
+        private IOptions<HostingSettings>  _hostingSettings;
         private readonly IWebHostEnvironment _webHostEnvironment;
 
         private string _localTempPath;
 
-        public AspNetCoreHostingEnvironment(IOptionsMonitor<HostingSettings> hostingSettings, IWebHostEnvironment webHostEnvironment)
+        public AspNetCoreHostingEnvironment(IOptions<HostingSettings> hostingSettings, IWebHostEnvironment webHostEnvironment)
```
bergmania, line 18: This change avoids us from updating the debug flag without restarting the app.

rustybox: Ergh, in which case I think really CoreRuntimeBootstrapper class needs to be nuked.

Everything it does should happen after ServiceProvider is built instead of this two pass approach to booting.

Will reset to fe3b57b (for which CI build has already completed successfully) and have another crack another day.

bergmania:
All seems to work 💪

src/Umbraco.Infrastructure/Migrations/Upgrade/V_8_0_0/DataTypes/PreValueMigratorComposer.cs
```diff
@@ -2,8 +2,7 @@
 
 namespace Umbraco.Core.Migrations.Upgrade.V_8_0_0.DataTypes
 {
-[RuntimeLevel(MinLevel = RuntimeLevel.Upgrade, MaxLevel = RuntimeLevel.Upgrade)] // only on upgrades
-public class PreValueMigratorComposer : ICoreComposer
+    public class PreValueMigratorComposer : ICoreComposer
```
ronaldbarendse, line 5: What ensures this only runs on upgrades now?

src/Umbraco.Web.Common/Runtime/AspNetCoreBootFailedComposer.cs
```diff
@@ -7,7 +7,6 @@ namespace Umbraco.Web.Common.Runtime
     /// <summary>
     /// Executes if the boot fails to ensure critical services are registered
     /// </summary>
-    [RuntimeLevel(MinLevel = RuntimeLevel.BootFailed)]
     public class AspNetCoreBootFailedComposer : IComposer
```
ronaldbarendse, line 10: Same for this composer: it should only be added when the application boot failed.

rustybox: Min level starts at boot failed and gets bigger, if it were supposed to stop it would use max level.

This is no longer handled in service registration, runtime state should be checked at resolve time not at ref time (which the boot failed middleware already does)

src/Umbraco.Infrastructure/Migrations/Upgrade/V_8_0_0/DataTypes/PreValueMigratorComposer.cs
rustybox: Composers add services, that should always happen.

Components etc use those services, migration story should involve checking runtime state just not at registration time.

All the discussion in #8653 is currently on this topic.

PR #9402, opened 2020-11-16
Unversion - Configure & control which versions are cleaned up

warrenbuckley:
# Feature
This adds the ability to configure and determine how many versions of all document types to keep or to add more granular and specific rules.

This is done in a background task and is performed once a day every hour as opposed to the original Unversion package which is done via the published event of a content node.

## Note:
This is still **draft**. I need to finish the following:

- [ ] Documentation
- [ ] Ship a sensible example XML config file - perhaps with commented out rules
- [x] Add more unit tests for more configuration cases

### Thanks
A BIG thanks and kudos to @skttl to allowing us to the bulk of the Unversion source code into the CMS core codebase 👍 
Along with the original owners/writers of Unversion code @mattbrailsford & @leekelleher

### DIFF
--- a/src/Umbraco.Web/Unversion/IUnversionService.cs
+++ b/src/Umbraco.Web/Unversion/IUnversionService.cs
@@ -1,0 +1,10 @@
+﻿using Umbraco.Core.Models;
+
+namespace Umbraco.Web.Unversion
+{
+    public interface IUnversionService
+    {
+        void Unversion(IContent content);
+    }
+}
+
--- a/src/Umbraco.Web/Unversion/UnversionComponent.cs
+++ b/src/Umbraco.Web/Unversion/UnversionComponent.cs
@@ -1,35 +1,41 @@
-﻿using Umbraco.Core.Composing;
-using Umbraco.Core.Events;
+﻿using Umbraco.Core;
+using Umbraco.Core.Composing;
+using Umbraco.Core.Logging;
 using Umbraco.Core.Services;
-using Umbraco.Core.Services.Implement;
+using Umbraco.Web.Scheduling;
 
 namespace Umbraco.Web.Unversion
 {
     public class UnversionComponent : IComponent
     {
-        private readonly IUnversionService unVersionService;
+        private readonly IRuntimeState _runtime;
+        private readonly IProfilingLogger _logger;
+        private readonly IContentService _contentService;
+        private readonly IUnversionService _unversionService;
+        private readonly BackgroundTaskRunner<IBackgroundTask> _unversionRunner;
 
-        public UnversionComponent(IUnversionService _unVersionService)
+        public UnversionComponent(IRuntimeState runtime, IProfilingLogger logger, IContentService contentService, IUnversionService unversionService)
         {
-            unVersionService = _unVersionService;
+            _runtime = runtime;
+            _logger = logger;            
+            _contentService = contentService;
+            _unversionService = unversionService;
+            _unversionRunner = new BackgroundTaskRunner<IBackgroundTask>("Unversion", _logger);
         }
 
         public void Initialize()
         {
-            ContentService.Published += ContentService_Published;
-        }
+            int delayBeforeWeStart = 60000; // 60000ms = 1min
+            int howOftenWeRepeat = 6000 * 60; // 60000 * 60 = 1hr
 
-        private void ContentService_Published(IContentService sender, ContentPublishedEventArgs e)
-        {
-            foreach (var content in e.PublishedEntities)
-            {
-                unVersionService.Unversion(content);
-            }
+            var task = new UnversionTask(_unversionRunner, delayBeforeWeStart, howOftenWeRepeat, _runtime, _logger, _contentService, _unversionService);
+
+            //As soon as we add our task to the runner it will start to run (after its delay period)
+            _unversionRunner.TryAdd(task);
         }
 
         public void Terminate()
         {
-            ContentService.Published -= ContentService_Published;
         }
     }
 }
--- a/src/Umbraco.Web/Unversion/UnversionComponent.cs
+++ b/src/Umbraco.Web/Unversion/UnversionComponent.cs
@@ -1,0 +1,35 @@
+﻿using Umbraco.Core.Composing;
+using Umbraco.Core.Events;
+using Umbraco.Core.Services;
+using Umbraco.Core.Services.Implement;
+
+namespace Umbraco.Web.Unversion
+{
+    public class UnversionComponent : IComponent
+    {
+        private readonly IUnversionService unVersionService;
+
+        public UnversionComponent(IUnversionService _unVersionService)
+        {
+            unVersionService = _unVersionService;
+        }
+
+        public void Initialize()
+        {
+            ContentService.Published += ContentService_Published;
+        }
+
+        private void ContentService_Published(IContentService sender, ContentPublishedEventArgs e)
+        {
+            foreach (var content in e.PublishedEntities)
+            {
+                unVersionService.Unversion(content);
+            }
+        }
+
+        public void Terminate()
+        {
+            ContentService.Published -= ContentService_Published;
+        }
+    }
+}
--- a/src/Umbraco.Web/Unversion/UnversionComposer.cs
+++ b/src/Umbraco.Web/Unversion/UnversionComposer.cs
@@ -1,0 +1,15 @@
+﻿using Umbraco.Core;
+using Umbraco.Core.Composing;
+
+namespace Umbraco.Web.Unversion
+{
+    public class UnversionComposer : ICoreComposer
+    {
+        public void Compose(Composition composition)
+        {
+            composition.Register<IUnversionConfig, UnversionConfig>();
+            composition.Register<IUnversionService, UnversionService>();
+            composition.Components().Append<UnversionComponent>();
+        }
+    }
+}
--- a/src/Umbraco.Web/Unversion/UnversionConfig.cs
+++ b/src/Umbraco.Web/Unversion/UnversionConfig.cs
@@ -49,6 +49,7 @@ namespace Umbraco.Web.Unversion
                 {
                     var configEntry = new UnversionConfigEntry
                     {
+                        // If no doctype alias set on XML entry then its assume it's for all doctypes
                         DocTypeAlias = xmlConfigEntry.Attributes["docTypeAlias"] != null
                             ? xmlConfigEntry.Attributes["docTypeAlias"].Value
                             : AllDocumentTypesKey
--- a/src/Umbraco.Web/Unversion/UnversionConfig.cs
+++ b/src/Umbraco.Web/Unversion/UnversionConfig.cs
@@ -1,0 +1,91 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Web;
+using System.Xml;
+using Umbraco.Core.Logging;
+
+namespace Umbraco.Web.Unversion
+{
+    public class UnversionConfig : IUnversionConfig
+    {
+        public const string AllDocumentTypesKey = "$_ALL";
+        public IDictionary<string, List<UnversionConfigEntry>> ConfigEntries { get; set; }
+
+        private ILogger _logger;
+
+        public UnversionConfig(ILogger logger)
+        {
+            _logger = logger;
+
+            ConfigEntries = new Dictionary<string, List<UnversionConfigEntry>>();
+
+            try
+            {
+                var appPath = HttpRuntime.AppDomainAppPath;
+                var configFilePath = Path.Combine(appPath, @"config\unversion.config");
+                LoadXmlConfig(string.Concat(configFilePath));
+            }
+            catch (Exception e)
+            {
+                _logger.Error<UnversionConfig>(e, "Error when parsing unversion.config.");
+            }
+
+        }
+
+        private void LoadXmlConfig(string configPath)
+        {
+            if (!File.Exists(configPath))
+            {
+                _logger.Warn<UnversionConfig>("Couldn't find config file {configPath}", configPath);
+                return;
+            }
+
+            var xmlConfig = new XmlDocument();
+            xmlConfig.Load(configPath);
+
+            foreach (XmlNode xmlConfigEntry in xmlConfig.SelectNodes("/unversionConfig/add"))
+                if (xmlConfigEntry.NodeType == XmlNodeType.Element)
+                {
+                    var configEntry = new UnversionConfigEntry
+                    {
+                        DocTypeAlias = xmlConfigEntry.Attributes["docTypeAlias"] != null
+                            ? xmlConfigEntry.Attributes["docTypeAlias"].Value
+                            : AllDocumentTypesKey
+                    };
+
+                    if (xmlConfigEntry.Attributes["rootXpath"] != null)
+                        configEntry.RootXPath = xmlConfigEntry.Attributes["rootXpath"].Value;
+
+                    if (xmlConfigEntry.Attributes["maxDays"] != null)
+                        configEntry.MaxDays = Convert.ToInt32(xmlConfigEntry.Attributes["maxDays"].Value);
+
+                    if (xmlConfigEntry.Attributes["maxCount"] != null)
+                        configEntry.MaxCount = Convert.ToInt32(xmlConfigEntry.Attributes["maxCount"].Value);
+
+                    if (!ConfigEntries.ContainsKey(configEntry.DocTypeAlias))
+                        ConfigEntries.Add(configEntry.DocTypeAlias, new List<UnversionConfigEntry>());
+
+                    ConfigEntries[configEntry.DocTypeAlias].Add(configEntry);
+                }
+        }
+    }
+
+    public class UnversionConfigEntry
+    {
+        public UnversionConfigEntry()
+        {
+            MaxDays = MaxCount = int.MaxValue;
+        }
+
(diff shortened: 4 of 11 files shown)


src/Umbraco.Web/Unversion/UnversionTask.cs
```diff
@@ -0,0 +1,82 @@
...
+                return true; // We return true to try again as the server role may change!
+            }
+
+            var allContent = new List<IContent>();
+
+            // Add TraceDuration - so users can see if this is taking a long time to calculate
+            using (_logger.TraceDuration<UnversionTask>("Calculating number of content nodes in site for Unversion", "Calculated number of nodes in site for Unversion"))
+            {               
+                // Get all content nodes
+                // Worried this could be bad for performance...
```
skttl, line 43: Make sure to test this on a large site, I have the same worry :)

src/Umbraco.Web/Unversion/UnversionComponent.cs
```diff
@@ -0,0 +1,41 @@
...
+            _logger = logger;            
+            _contentService = contentService;
+            _unversionService = unversionService;
+            _unversionRunner = new BackgroundTaskRunner<IBackgroundTask>("Unversion", _logger);
+        }
+
+        public void Initialize()
+        {
+            int delayBeforeWeStart = 60000; // 60000ms = 1min
+            int howOftenWeRepeat = 6000 * 60; // 60000 * 60 = 1hr
```
skttl, line 29: In the PR description, you say that the unversioning happens once every 24 hours, but here it happens every hour.

src/Umbraco.Web/Unversion/UnversionTask.cs
warrenbuckley: Yes I will do some testing shortly

src/Umbraco.Web/Unversion/UnversionComponent.cs
warrenbuckley: Thanks @skttl yes its supposed to be every hour AFAIK. Will update the description of PR 👍

src/Umbraco.Web/Unversion/UnversionService.cs
```diff
@@ -25,27 +25,31 @@ public UnversionService(IUnversionConfig config, ILogger logger, IUmbracoContext
...
 
             foreach (var configEntry in configEntries)
             {
                 var isValid = true;
 
-                // Check the RootXPath if configured
+                // Check if RootXPath is configured
                 if (!string.IsNullOrEmpty(configEntry.RootXPath))
-                    // TODO: Fix in some otherway
+                    // TODO: Fix in some otherway (Check with Soren on what he meant by this)
```
skttl, line 53: I don't really recall, but I could imagine that GetNodeIdsFromXpath was broken on the first try to port, and I probably just forgot to remove the comment :)

warrenbuckley: OK I will remove the comment & assume this is all good then :)

marcemarc:
Wow, wonderful, but now we can't all say 'unversion' is the first package we install!!

If this is shipped in a 'turned off state' eg people should opt into it, in case they lose that important version from 4yrs ago... then I was wondering if a 'way' to surface it's existence to people could be via a Content Version HealthCheck?

I created one in the past for an early version of V7 that shows some 'versioning' stats of the site and arbitrarily determines whether a site was 'overly' versioned... and of course pointed people towards installing unversion.

contentversion-healthcheck

This was the gist: https://github.com/marcemarc/Our.Umbraco.HealthChecks/blob/b1dd280b0fc07452f5e9b1e294906b51275a58b1/src/Our.Umbraco.HealthChecks/Checks/DataIntegrity/ContentVersionsCheck.cs

But, that is just to get the idea, if people run the 'health checks' when they are new to Umbraco...

then we could have a check that says , you currently don't have any unversion rules configured, but also you currently have a 'normal' amount of versions... and a link to the docs?

and then we could flag the 'most' content versioned item or show a warning if an item has over 100 versions, or if on average there are more than 5 versions per content item - and draw the attention again to the existence of the  configuration?

nathanwoulfe:
@warrenbuckley this is awesome, a great addition to core. 

Wondering though, is it possible to only unversion saved versions and keep published versions? 

For example, we have pretty stringent data retention rules, and effectively must keep copies of everything ever published (mainly because no one can provide any other advice around retention), so ideally would be able to remove saves and keep published versions only.

warrenbuckley:
@nathanwoulfe isn't that just remove all versions then? So max could be set to 0 surely or am I misunderstanding?

nathanwoulfe:
@warrenbuckley maybe I'm misunderstanding - does this remove saved versions only? Or old published versions too?

warrenbuckley:
@marcemarc yes I think it will be disabled by default and your idea to surface it sounds good. Not sure if it will make it in the first iteration, but I'll cc @filipbech in

warrenbuckley:
@nathanwoulfe I am not at my desk right now. The joys of GitHub mobile will check in the morning. However it's the same functionality as the Unversion package if you want to experiment on a local dev site until I can reply :)

nathanwoulfe:
WHAT?! YOU DONT HAVE ALL THE ANSWERS?!

All good, I'll have a play and see what turns up.

src/Umbraco.Web/Unversion/IUnversionService.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Umbraco.Core.Models;
+
+namespace Umbraco.Web.Unversion
```
Shazwazza, line 3: Why is this in Web? It has nothing to do with Web and doesn't require only 'Web' things

src/Umbraco.Web/Unversion/IUnversionService.cs
```diff
@@ -0,0 +1,10 @@
...
+{
+    public interface IUnversionService
+    {
+        void Unversion(IContent content);
```
Shazwazza, line 7: Does this mean you can only unversion one content item at a time?

src/Umbraco.Web/Unversion/UnversionComposer.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Umbraco.Core;
+using Umbraco.Core.Composing;
+
+namespace Umbraco.Web.Unversion
+{
+    public class UnversionComposer : ICoreComposer
+    {
+        public void Compose(Composition composition)
+        {
+            composition.Register<IUnversionConfig, UnversionConfig>();
```
Shazwazza, line 10: IUnversionConfig and IUnversionService should be part of core, they don't rely on anything in web and should be able to run outside of web if required. Probably should be registered how our other services and config are registered.

src/Umbraco.Web/Unversion/UnversionConfig.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Web;
+using System.Xml;
+using Umbraco.Core.Logging;
+
+namespace Umbraco.Web.Unversion
```
Shazwazza, line 8: Why is this in Web? It has nothing to do with Web and doesn't require only 'Web' things

src/Umbraco.Web/Unversion/UnversionConfig.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        public UnversionConfig(ILogger logger)
+        {
+            _logger = logger;
+
+            ConfigEntries = new Dictionary<string, List<UnversionConfigEntry>>();
+
+            try
+            {
+                var appPath = HttpRuntime.AppDomainAppPath;
```
Shazwazza, line 25: Please do not use this. Use IOHelper.MapPath to get paths which is unit testable, this is not.

src/Umbraco.Web/Unversion/UnversionConfig.cs
```diff
@@ -0,0 +1,97 @@
...
+                LoadXmlConfig(string.Concat(configFilePath));
+            }
+            catch (Exception e)
+            {
+                _logger.Error<UnversionConfig>(e, "Error when parsing unversion.config.");
+            }
+
+        }
+
+        private void LoadXmlConfig(string configPath)
```
Shazwazza, line 36: Since this isn't  a real .net config file this means any changes to the file will not result in reloading them and the app will need to be restarted for that. Is that what we want? We've previously had xml config files that were manually parsed like this resulting in lots of parsing errors because all edge cases haven't been taken into account which is why we generally rely on real .net config files.

src/Umbraco.Web/Unversion/UnversionConfig.cs
```diff
@@ -0,0 +1,97 @@
...
+                    };
+
+                    if (xmlConfigEntry.Attributes["rootXpath"] != null)
+                        configEntry.RootXPath = xmlConfigEntry.Attributes["rootXpath"].Value;
+
+                    if (xmlConfigEntry.Attributes["maxDays"] != null)
+                        configEntry.MaxDays = Convert.ToInt32(xmlConfigEntry.Attributes["maxDays"].Value);
+
+                    if (xmlConfigEntry.Attributes["maxCount"] != null)
+                        configEntry.MaxCount = Convert.ToInt32(xmlConfigEntry.Attributes["maxCount"].Value);
```
Shazwazza, line 65: For example, this will just throw an ugly exception if it's not formatted correctly, it's not being proactive in checking for correct data formats. Again, this would be solved by real .net configuration classes like we have for everything else.

src/Umbraco.Web/Unversion/UnversionConfig.cs
```diff
@@ -0,0 +1,97 @@
...
+
+                    if (!ConfigEntries.ContainsKey(configEntry.DocTypeAlias))
+                        ConfigEntries.Add(configEntry.DocTypeAlias, new List<UnversionConfigEntry>());
+
+                    ConfigEntries[configEntry.DocTypeAlias].Add(configEntry);
+                }
+        }
+    }
+
+    public class UnversionConfigEntry
```
Shazwazza, line 78: This whole class is mutable which means you could just change these values at runtime. This should be an interface with readonly {get;} (immutable)  and this could be its implementation - but IMO we should be consistent with our config and use the same methodologies we are using throughout our app with real .net config.

Shazwazza: I also feel like the amount of work being put in to this config will be more work than just making all of this configuration based on document types fields in the DB and managed in the UI.

src/Umbraco.Web/Unversion/UnversionConfig.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        public string DocTypeAlias { get; set; }
+        public string RootXPath { get; set; }
+        public int MaxDays { get; set; }
+        public int MaxCount { get; set; }
+        public int MinCount { get; set; }
+    }
+
+    public interface IUnversionConfig
```
Shazwazza, line 93: separate files

src/Umbraco.Web/Unversion/UnversionConfig.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        IDictionary<string, List<UnversionConfigEntry>> ConfigEntries { get; }
```
Shazwazza, line 95: List is mutable, and so is IDictionary we cannot have mutable config exposed in interfaces

IReadOnlyDictionary<string, IEnumerable<IUnversionConfigEntry>> 

... But again we should use the xml config approach we use throughout our app currently.

src/Umbraco.Web/Unversion/UnversionService.cs
```diff
@@ -0,0 +1,169 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Umbraco.Core;
+using Umbraco.Core.Logging;
+using Umbraco.Core.Models;
+using Umbraco.Core.Services;
+
+namespace Umbraco.Web.Unversion
```
Shazwazza, line 9: Why Web?

src/Umbraco.Web/Unversion/UnversionService.cs
```diff
@@ -0,0 +1,169 @@
...
+                if (!isValid)
+                {
+                    _logger.Debug<UnversionService>("Invalid RootXPath configuration for '{ContentTypeAlias}' the XPath selector does not contain the parent id '{ContentParentId}' for the current content item '{ContentUdi}'.",
+                        configEntry.DocTypeAlias, content.ParentId, content.GetUdi());
+
+                    continue;
+                }
+
+                // Get every version of content node
+                var allVersions = _contentService.GetVersionsSlim(content.Id, 0, int.MaxValue).ToList();
```
Shazwazza, line 68: This needs to be paged, this could be enormous. Like on slack, we should only process smaller batches each time until eventually they are all gone.

src/Umbraco.Web/Unversion/UnversionConfig.cs
```diff
@@ -0,0 +1,97 @@
...
+    public class UnversionConfigEntry
+    {
+        public UnversionConfigEntry()
+        {
+            MaxDays = MaxCount = int.MaxValue;
+            MinCount = int.MinValue;
+        }
+
+        public string DocTypeAlias { get; set; }
+        public string RootXPath { get; set; }
```
Shazwazza, line 87: Do we really want this? How important is this? It adds quite a lot of complexity and seems overkill IMO. Versioning should just be based on document type and hopefully the configuration of each type would be part of the UI not part of config - since that would require adjusting the config (code) for each doc type created/changed instead of managing it all from the back office and being able to define defaults. The amount of work involved to make this config seems to outweigh the work for having this in the UI. But specifying this XPath means we could never have it in the UI.

src/Umbraco.Web/Unversion/UnversionService.cs
```diff
@@ -0,0 +1,169 @@
...
+                // Determine versions to delete
+                var versionIdsToDelete = GetVersionsToDelete(allVersions, configEntry, DateTime.Now);
+
+                // Delete versions
+                foreach (var vid in versionIdsToDelete)
+                {
+                    _logger.Info<UnversionService>("Deleting version {VersionId} of content '{ContentName}' {ContentUdi}", vid, content.Name, content.GetUdi());
+
+                    // Ensure we only delete the specified version and do NOT delete PRIOR versions
+                    _contentService.DeleteVersion(content.Id, vid, false);
```
Shazwazza, line 84: This needs to be done in bulk, we cannot do this one at a time, this will kill performance.

src/Umbraco.Web/Unversion/UnversionService.cs
Shazwazza: We should not be processing one single item at a time, this is going to kill performance, it all needs to happen in bulk.

src/Umbraco.Web/Unversion/UnversionService.cs
```diff
@@ -0,0 +1,169 @@
...
+                }
+            }
+
+        }
+
+        /// <summary>
+        /// Iterates a list of IContent versions and returns items to be removed based on a configEntry.
+        /// </summary>
+        public List<int> GetVersionsToDelete(List<IContent> versions, UnversionConfigEntry configEntry, DateTime currentDateTime)
```
Shazwazza, line 93: List is mutable, only return mutable objects if you plan on allowing them to be mutable which is almost no cases for returning collections from methods. IEnumerable is not mutable.

src/Umbraco.Web/Unversion/UnversionService.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+
+            return versionIdsToDelete;
+
+        }
+
+        /// <summary>
+        /// Gets a list of int NodeIds from an xPath query
+        /// </summary>
+        private List<int> GetNodeIdsFromXpath(string xpath)
```
Shazwazza, line 156: This is returning nodes from the cache? Why? What about unpublished items? We should absolutely not be relying on the front-end cache for this processing. IMO we shouldn't be doing any XPath lookups and only relying on doc types.

src/Umbraco.Web/Unversion/UnversionTask.cs
Shazwazza: All of this needs to be done in bulk, this cannot happen by selecting each content item in the tree. Lets please revisit all of this and how it should be handled.

Shazwazza (changes requested):
I have left some notes. I think we need to re-think some fundamentals:

* Why is this config based? If it is to save time, I think it will be less time to implement this correctly in the UI and have these rules based on document types. If they are not based there, then anytime a developer adds a document type, they need to add config. This doesn't really work for Umbraco Cloud if we wanted to add version limits to ALL doc types by default for some tiers (I'm not saying this will happen but it could and we should think about it). Also I'm sure many devs would like to have defaults for all document types. This would just be so much easier to manage with not having this in config and if the plan is to eventually move it to the UI then we are just shipping dead code.
* I don't know why we have a content XPath defined? This should all just be based on document types
* All of the processing needs to be done in bulk and for the most part that can be done in SQL via the SQL server, not by processing things one at a time in c#. This is going to run often and it needs to perform very well.

filipbech:
After thorough consideration we have decided to put this feature on hold, until we have the capacity and focus to finish this in a performant way. Closing this PR. 
Announcement in the Product Update from last week => https://umbraco.com/blog/product-update-november-20th-2020/

UnVersion is still the recommended way to achieve this going forward.

PR #9403, opened 2020-11-16
Add noopener to target _blank links

georg-jung:
Fixes SonarQube finding "Links with "target=_blank" should prevent phishing attacks"
See https://rules.sonarsource.com/html/type/Vulnerability/RSPEC-5148

### Prerequisites

- I don't think there is a sensible way to test this in code/don't think this will be needed

If there's an existing issue for this PR then this fixes *I didn't open an issue*

### Description

I first brought this up by mailing security@umbraco.com. Jesper asked me to rather open an issue in this case. I decided to provide this PR instead right away.

While analyzing my Umbraco 8 project using SonarQube, I got notified about this vulnerability: https://rules.sonarsource.com/html/type/Vulnerability/RSPEC-5148 `Links with "target=_blank" should prevent phishing attacks`

This occurs in src/Umbraco.Web.UI/config/splashes/noNodes.aspx. Both links link to umbraco-owned domains so this shouldn't be to much of an issue. Still I think it would be nice to add "rel="noopener noreferrer"". Motivation:
* Fix this for future users of static analysis solutions
* Umbraco site could be compromised by an attacker
* I don't see why not

### DIFF
--- a/src/Umbraco.Web.UI/config/splashes/noNodes.aspx
+++ b/src/Umbraco.Web.UI/config/splashes/noNodes.aspx
@@ -39,14 +39,14 @@
 					<h2>Easy start with Umbraco.tv</h2>
 					<p>We have created a bunch of 'how-to' videos, to get you easily started with Umbraco. Learn how to build projects in just a couple of minutes. Easiest CMS in the world.</p>
 					
-					<a href="https://umbraco.tv?ref=tvFromInstaller" target="_blank">Umbraco.tv &rarr;</a>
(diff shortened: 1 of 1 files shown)


georg-jung:
Probably #9369 pursues a quite similar goal. However it seems like doing `rel="noopener noreferrer"` instead of just `rel="noopener"` is preferrable. I'll leave a comment there and leave this open for now. If you think the other PR will be merged of course feel free to close this one.

georg-jung:
Closing after the discussion in #9403

PR #9404, opened 2020-11-17
Netcore: Migration of Model classes from Umbraco.Infrastructure to Core

elit0451:
The purpose of this PR is to isolate our dependencies -> Removing/Refactoring the usage of Newtonsoft json wherever possible.

Files migrated from:
- Models
- Models/Mapping
- Models/PublishedContent
- Models/Editors
- Models/Blocks

Files that cannot be migrated from those folders ⬆️ :
- _EntityMapDefinition_ - Examine and Umbraco.Examine using
- _PathValidationExtensions_ - dependency to NodeDto which has NPoco using

> Note: **ConfigurationEditor** class was partially migrated so that we can unblock some of the model classes. TODO is left for when we look at migrating fully that class & when this happens, changes in commit 15d60668c005992114b40530b3c05e79af5646e8 should be reverted


bergmania:
To fully migrate ConfigurationEditor I think we can use a Custom JsonConverter

PR #9408, opened 2020-11-18
Ensure that TaskScheduler.Default is used anywhere that ContinueWith is used, adds more debug logging to MainDom operations

Shazwazza:
Ensure that TaskScheduler.Default is used anywhere that ContinueWith is used and adds more debug logging to MainDom operations.

ContinueWith should always have `TaskScheduler.Default` explicitly declared. There's a few instances in the codebase where this is not the case. It hasn't caused problems in the past that we know about but it is the safe thing to do. For more reading see https://blog.stephencleary.com/2013/10/continuewith-is-dangerous-too.html

This adds a little bit more debug logging to MainDom. This is for debugging purposes only because there's one known site where MainDom shuts down without the application being shutdown and without any SqlMainDomLock logging taking place which is truly odd. 

## Testing

* Configure your site to use SqlMainDomLock
* Add this logging config to your serilog.config
   ```
   <add key="serilog:minimum-level:override:Umbraco.Core.Runtime.MainDom" value="Debug" />
   <add key="serilog:minimum-level:override:Umbraco.Core.Runtime.SqlMainDomLock" value="Debug" />
   ```
* Test restarting your running app by bumping the web.config, etc... 
* Check the logs during restart MainDom should show it's being signaled, shutdown and acquired each time like normal

### DIFF
--- a/src/Umbraco.Core/Runtime/SqlMainDomLock.cs
+++ b/src/Umbraco.Core/Runtime/SqlMainDomLock.cs
@@ -119,7 +119,12 @@ namespace Umbraco.Core.Runtime
 
             // Create a long running task (dedicated thread)
             // to poll to check if we are still the MainDom registered in the DB
-            return Task.Factory.StartNew(ListeningLoop, _cancellationTokenSource.Token, TaskCreationOptions.LongRunning, TaskScheduler.Default);
+            return Task.Factory.StartNew(
+                ListeningLoop,
+                _cancellationTokenSource.Token,
+                TaskCreationOptions.LongRunning,
+                // Must explicitly specify this, see https://blog.stephencleary.com/2013/10/continuewith-is-dangerous-too.html
+                TaskScheduler.Default); 
 
         }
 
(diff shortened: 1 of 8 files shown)


src/Umbraco.Core/Runtime/SqlMainDomLock.cs
```diff
@@ -147,20 +152,18 @@ private void ListeningLoop()
                     continue;
                 }
 
-                if (!_dbFactory.Configured)
-                {
-                    // if we aren't configured, we just keep looping since we can't query the db
-                    continue;
-                }
-
```
bergmania, line 155: Based on the description of the PR, I don't understand why this is removed. Is it intentionally?

Shazwazza: This was duplicate code, if you expand this you can see this exists above, must have been a merge that inserted the duplicate code.

bergmania (approved):

nul800sebastiaan:
Cherry picked for 8.6 and above as well.

PR #9412, opened 2020-11-18
NetCore: Creates abstraction for membership helper and migrates controllers now dependent on it.

AndyButland:
I've migrated the four controllers shipped with Umbraco for supporting membership operations into the website project, removing the dependency on `MembershipHelper` (which in turn is tied to .NET Framework membership providers).  I've added just the methods we need into the new abstraction and implemented a couple simple ones (that don't depend on identity objects like `UserManager` and `SignInManager` that still need to be created).

I've moved the models up into core and removed the associated model binder references (as it doesn't seem we need them).

Previously the register and update operations returned an object representing the member, but in the form of `MembershipUser` which we can't use.  Again, we don't specifically need this for the code that ships with Umbraco so I've, at least for now, removed this.


src/Umbraco.Web.Website/Extensions/UmbracoWebstiteServiceCollectionExtensions.cs
```diff
@@ -21,6 +23,9 @@ public static void AddUmbracoWebsite(this IServiceCollection services)
 
             //TODO figure out if we need more to work on load balanced setups
             services.AddDataProtection();
+
+            // Website security
+            services.AddSingleton<IUmbracoWebsiteSecurity, UmbracoWebsiteSecurity>();
```
bergmania, line 28: We will need an Accessor, like we have for BackofficeSecurity. Otherwise the members are shared between sessions

AndyButland: I've updated to introduce this if you could please check that it's how you intended.

bergmania: Exactly like this

marcemarc:
Hi @AndyButland 

src/Umbraco.Web.Website/Extensions/UmbracoWebstiteServiceCollectionExtensions.cs

is it a typo? Webstite?

(or am I revealing my lack of knowledge here!)

AndyButland:
Typo indeed (and not an undocumented breaking change that we're getting rid of websites) - fixed.

marcemarc:
I did google first to be sure... the way to remember it seems to be that Webstites form from above and so have to hold on 'tight/tite'

bergmania:
Good work @AndyButland.. Next step could be the method required to migrate the views in `Umbraco.Web.UI/Umbraco/PartialviewMacros`

PR #9414, opened 2020-11-18
Linux paths

benjaminc:
Add support for linux paths to the IOHelper, and update the MapPath callers to instead call the IHostingEnvironment methods which distinguish between content and web roots.


src/Umbraco.Core/Configuration/XmlConfigManipulator.cs
```diff
@@ -5,25 +5,25 @@
...
-            _ioHelper = ioHelper;
+            _hostingEnvironment = hostingEnvironment;
             _logger = logger;
         }
 
         public void RemoveConnectionString()
         {
             var key = Constants.System.UmbracoConnectionName;
-            var fileName = _ioHelper.MapPath("~/web.config");
+            var fileName = _hostingEnvironment.MapPathContentRoot("~/web.config");
```
JimBobSquarePants, line 26: Why is `web.config` still a thing?

bergmania: It's a legacy thing. I removed it.

src/Umbraco.Tests.Common/TestHelperBase.cs
```diff
@@ -86,7 +87,11 @@ public IIOHelper IOHelper
             get
             {
                 if (_ioHelper == null)
-                    _ioHelper = new IOHelper(GetHostingEnvironment());
+                {
+                    var hostingEnvironment = GetHostingEnvironment();
+                    _ioHelper = RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? new IOHelperLinux(hostingEnvironment)
+                        : (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? (IIOHelper)new IOHelperOSX(hostingEnvironment) : new IOHelperWindows(hostingEnvironment));
```
rustybox, line 93: That's a lot of ternaries, would read a lot nicer if it wern't.

JimBobSquarePants: I would personally create a dedicated class for environmental properties. `TestEnvironment`

``` c#
public static bool IsLinux => RuntimeInformation.IsOSPlatform(OSPlatform.Linux);

public  static bool IsOSX => RuntimeInformation.IsOSPlatform(OSPlatform.OSX);

public  static bool IsWindows => RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```

PR #9425, opened 2020-11-20
Introduce content app helper

mattbrailsford:
I've introduced a new content app helper service which it's primary use at the moment will just be to perform checks as to whether a content app is "content" based. These checks are currently being done within other areas of code so there is some duplication. This moves that duplication to a single helper so this logic is kept in one place.

To test, you can setup some content with a custom content app. If it works, core content apps will continue to display the publish buttons, but non-content-based apps (ie, custom content apps) will not display the publish buttons.)


nul800sebastiaan:
Thanks @mattbrailsford - great to get rid of some duplication there! 👍

PR #9440, opened 2020-11-23
Support more primitives in nucache

nzdev:
### Prerequisites

### Description
What?
Support serializing more primitives to nucache

Why?
More efficient to serialize.

### DIFF
--- a/src/Umbraco.Web/PublishedCache/NuCache/DataSource/LazyCompressedString.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/DataSource/LazyCompressedString.cs
@@ -13,6 +13,7 @@ namespace Umbraco.Web.PublishedCache.NuCache.DataSource
         private byte[] _bytes;
         private string _str;
         private readonly object _locker;
+        private bool _isDecompressed;
 
         /// <summary>
         /// Constructor
@@ -22,7 +23,8 @@ namespace Umbraco.Web.PublishedCache.NuCache.DataSource
         {
             _locker = new object();
             _bytes = bytes;
-            _str = null;            
+            _str = null;
+            _isDecompressed = false;
         }
 
         public byte[] GetBytes()
@@ -30,6 +32,10 @@ namespace Umbraco.Web.PublishedCache.NuCache.DataSource
             if (_bytes == null)
                 throw new InvalidOperationException("The bytes have already been expanded");
             return _bytes;
+        }
+        public bool IsDecompressed()
+        {
+            return _isDecompressed;
         }
 
         public override string ToString()
@@ -41,6 +47,7 @@ namespace Umbraco.Web.PublishedCache.NuCache.DataSource
                 if (_bytes == null) throw new PanicException("Bytes have already been cleared");
                 _str = Encoding.UTF8.GetString(LZ4Pickler.Unpickle(_bytes));
                 _bytes = null;
+                _isDecompressed = true;
             }
             return _str;
         }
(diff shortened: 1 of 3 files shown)


src/Umbraco.Web/PublishedCache/NuCache/DataSource/LazyCompressedString.cs
```diff
@@ -31,6 +33,10 @@ public byte[] GetBytes()
                 throw new InvalidOperationException("The bytes have already been expanded");
             return _bytes;
         }
+        public bool IsDecompressed()
```
Shazwazza, line 36: Maybe nicer as a property? Be nice with some comments explaining what this is for

src/Umbraco.Web/PublishedCache/NuCache/DataSource/SerializerBase.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Umbraco.Web.PublishedCache.NuCache.DataSource
 {
-    internal abstract class SerializerBase
+    public abstract class SerializerBase
```
Shazwazza, line 7: needs to be public?

src/Umbraco.Web/PublishedCache/NuCache/DataSource/SerializerBase.cs
```diff
@@ -154,8 +172,46 @@ protected void WriteObject(object value, Stream stream)
             }
             else if (value is LazyCompressedString lazyCompressedString)
             {
-                PrimitiveSerializer.Char.WriteTo(PrefixCompressedStringByteArray, stream);
-                PrimitiveSerializer.Bytes.WriteTo(lazyCompressedString.GetBytes(), stream);
+                if (lazyCompressedString.IsDecompressed())
```
Shazwazza, line 175: Not sure I understand this one. If ToString() is called on the LazyCompressedString it will be decompressed and because of that it will write it to the cache file as a string instead of bytes? When is the logic that determines this?

Shazwazza (changes requested):
Makes sense to add the additional primitives, future proofing is good. Have left some comments/questions though.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/SerializerBase.cs
nzdev: If the type is LazyCompressedString , PrefixCompressedStringByteArray is the prefix. If the string has been decompressed, bytes are null. So it has to be written as a string.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/LazyCompressedString.cs
nzdev: done

src/Umbraco.Web/PublishedCache/NuCache/DataSource/SerializerBase.cs
nzdev: made internal.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/SerializerBase.cs
Shazwazza: Will that ever happen though? If we do that then there's going to be a problem reading it because we are always reading as bytes.

nzdev: It happens. That's how I discovered this.

Shazwazza: Righto, will have to get some testing done to see where this is all at. Been too long in between reviews/work!

Shazwazza: The only way that GetBytes() can get called multiple times and then throw an exception is during debugging if you inspect the value of LazyCompressedString since internally that will call ToString() and the bytes will be read then resulting in an exception. 

We cannot do this logic here by checking if it's decompressed because that will end up putting things into the cache file in inconsistent ways.

I'll change how LazyCompressedString works and add a DebuggerDisplay to prevent this from occurring.

Issue #8030, opened 2020-04-28
Scheduled publish fails when multiple pages are published - Umbraco 8.6.1


KevinJump:
Scheduled publishing can fail when publishing multiple pages at the same time. 

## Umbraco version
I am seeing this issue on Umbraco version: 8.6.1

Reproduction
------------

If you set multiple pages (in testing approx. 10-20 seems do it) to have the same scheduled publish date, then when that date arrives not all items will be published. and scope errors (below) will be reported in the log. 

### Specifics
It looks like its an issue with the Scope while running in a background task, but i am at a loss to see what it should be doing 😞 

### Steps to reproduce

1. New Umbraco Install - (With standard starter key, SQLCE DB)
2. Schedule the individual pages in the starter kit to publish at the same time _(takes about 5-10 mins to do, so pick a time-date that gives you time to go through the site)._
3. Wait for publish
4. Look through logs. should see the errors below: 

### Expected result
All pages should be published. 

### Actual result
Some pages are published, but the background process stops at some point with an error: 
```
System.AggregateException: Exceptions were thrown by listed actions. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Umbraco.Core.Scoping.Scope.<>c__DisplayClass71_0.<RobustExit>b__2() in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 437
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions) in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 472
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions) in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 476
   at Umbraco.Core.Scoping.Scope.TryFinally(Int32 index, Action[] actions) in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 476
   at Umbraco.Core.Scoping.Scope.RobustExit(Boolean completed, Boolean onException) in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 420
   at Umbraco.Core.Scoping.Scope.DisposeLastScope() in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 400
   at Umbraco.Core.Scoping.Scope.Dispose() in D:\a\1\s\src\Umbraco.Core\Scoping\Scope.cs:line 363
[...]

PR #9441, opened 2020-11-23
Ensure there's explicit scopes used for all service access in the Examine indexing logic

Shazwazza:
Follow up to https://github.com/umbraco/Umbraco-CMS/issues/8030#issuecomment-732232597 and https://github.com/umbraco/Umbraco-CMS/pull/7994

There is still one area of the background task logic in Umbraco Examine that wasn't using an explicit scope and if the call to that isn't cached we may get unexpected exceptions. 

This update should solve all of this now. With the previous change we put all indexing operations explicitly on a background thread which is good but we assumed that all DB/Scope operations were also being done in explicit scopes. Turns out one area was missing. This change ensures that this area has an explicit scope, but this change also goes ahead and just wraps the entire indexing logic on the background task in one outer scope so any new changes done within there will always be run within an explicit scope.

### DIFF
--- a/src/Umbraco.Core/Services/Implement/PublicAccessService.cs
+++ b/src/Umbraco.Core/Services/Implement/PublicAccessService.cs
@@ -62,12 +62,10 @@ namespace Umbraco.Core.Services.Implement
             //start with the deepest id
             ids.Reverse();
 
-            using (var scope = ScopeProvider.CreateScope())
+            using (var scope = ScopeProvider.CreateScope(autoComplete: true))
             {
                 //This will retrieve from cache!
-                var entries = _publicAccessRepository.GetMany().ToArray();
-
-                scope.Complete();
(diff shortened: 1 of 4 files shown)


mzajkowski:
Looks like the tests are failing. Might it be related with the introduction of the new code here?

Shazwazza:
Have pushed an update, we'll see if that build passes. Have assigned as community-review tag - hopefully someone in the community or PR team can pick this one up

bergmania (approved):
Changes look good..  Btw, the changes to the tests really show why Builder pattern is valuable. If we had a ContentValueSetValidatorBuilder, you would only have to update the builder with the new `Mock.Of<IScopeProvider>()` and not all tests.

Shazwazza:
@bergmania don't forget we need to apply the version tag to these things

bergmania:
Need to talk with @nul800sebastiaan to know what versions this will be part of.

PR #9448, opened 2020-11-24
NetCore: Migrated member related partial views

AndyButland:
Along with the necessary methods from `MembershipHelper` into `IUmbracoWebsiteSecurity`.

There's some `TODOs` of course, pending work on the ASP.NET Core Identity setup for members.

Conscious of a couple of further amends we might consider here:

- Should we be creating and using tag helper for `BeginUmbracoForm`?
- Should we still need to ship referencing jquery?

But think they can come later, if at all.

### DIFF
--- a/src/Umbraco.Web.UI.NetCore/umbraco/PartialViewMacros/Templates/EditProfile.cshtml
+++ b/src/Umbraco.Web.UI.NetCore/umbraco/PartialViewMacros/Templates/EditProfile.cshtml
@@ -5,7 +5,8 @@
 @inject IUmbracoWebsiteSecurityAccessor UmbracoWebsiteSecurityAccessor
 
 @{
-    var profileModel = await UmbracoWebsiteSecurityAccessor.WebsiteSecurity.GetCurrentMemberProfileModelAsync();
+    var websiteSecurity = UmbracoWebsiteSecurityAccessor.WebsiteSecurity;
+    var profileModel = await websiteSecurity.GetCurrentMemberProfileModelAsync();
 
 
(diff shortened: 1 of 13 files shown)


src/Umbraco.Web.UI.NetCore/umbraco/PartialViewMacros/Templates/EditProfile.cshtml
```diff
@@ -1,14 +1,12 @@
...
-@inherits Umbraco.Web.Macros.PartialViewMacroPage
+﻿@using Umbraco.Core.Security
+@using Umbraco.Extensions
+@using Umbraco.Web.Website.Controllers
+@inherits Umbraco.Web.Common.Macros.PartialViewMacroPage
+@inject IUmbracoWebsiteSecurityAccessor UmbracoWebsiteSecurityAccessor
 
 @{
-    var profileModel = Current.MembershipHelper.GetCurrentMemberProfileModel();
+    var profileModel = await UmbracoWebsiteSecurityAccessor.WebsiteSecurity.GetCurrentMemberProfileModelAsync();
```
bergmania: I think we should save `UmbracoWebsiteSecurityAccessor.WebsiteSecurity` in a variable

bergmania (approved):
Nice work.. Only found small things that I will fix up

bergmania:
@AndyButland, Regarding you questions..

I don't thing this is "need" to change.. Maybe a tag helper could be a nice addition, but the current extension methods still works :)

Whether we should remove jquery is IMO out of scope

PR #9450, opened 2020-11-25
Speed up boot time by not loading draft data from nucache on replica servers

nzdev:
### Prerequisites

### Description
What?
Add an option to not load the draft data (for content, media still loaded) from cmsContentNu for replica servers. On replica servers no one should be updating content or previewing so no need to load this data. An improvement to this would be to only load unpublished data on demand so the main server does not have to load draft data at startup either.

Why?
Faster boot up times by reducing the amount of joins made and data transferred, save on memory as unused draft data is not being kept in memory.

How to test:
Add to AppSettings
`<add key="Umbraco.Web.PublishedCache.NuCache.NoUnpublishedContentData" value = "true"/>`
Published content should load / render as normal. Preview content will have null values for Property Data.

### DIFF
--- a/src/Umbraco.Web/PublishedCache/NuCache/DataSource/BTree.DictionaryOfPropertyDataSerializer.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/DataSource/BTree.DictionaryOfPropertyDataSerializer.cs
@@ -51,6 +51,11 @@ namespace Umbraco.Web.PublishedCache.NuCache.DataSource
         public void WriteTo(IDictionary<string, PropertyData[]> value, Stream stream)
         {
             // write properties count
+            if (value == null)
+            {
+                PrimitiveSerializer.Int32.WriteTo(0, stream);
+                return;
+            }
             PrimitiveSerializer.Int32.WriteTo(value.Count, stream);
 
             // write each property
--- a/src/Umbraco.Web/PublishedCache/NuCache/DataSource/DatabaseDataSource.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/DataSource/DatabaseDataSource.cs
@@ -21,8 +21,13 @@ namespace Umbraco.Web.PublishedCache.NuCache.DataSource
     internal class DatabaseDataSource : IDataSource
     {
         private const int PageSize = 500;
+        private readonly DatabaseDataSourceConfiguration _dataSourceConfiguration;
 
         // we want arrays, we want them all loaded, not an enumerable
+        public DatabaseDataSource(DatabaseDataSourceConfiguration dataSourceConfiguration)
+        {
+            _dataSourceConfiguration = dataSourceConfiguration;
+        }
 
         private Sql<ISqlContext> ContentSourcesSelect(IScope scope, Func<Sql<ISqlContext>, Sql<ISqlContext>> joins = null)
         {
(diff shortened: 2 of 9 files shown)


Shazwazza:
Hey @nzdev! I have time booked in each sprint to review a lot of your work but unfortunately time has got away from me this sprint. I will get to reviewing, etc... as soon as I can including the work with the nucache stuff. Sorry for the delay!

src/Umbraco.Web/PublishedCache/NuCache/DataSource/BTree.DictionaryOfPropertyDataSerializer.cs
```diff
@@ -51,6 +51,11 @@ internal class DictionaryOfPropertyDataSerializer : SerializerBase, ISerializer<
         public void WriteTo(IDictionary<string, PropertyData[]> value, Stream stream)
         {
             // write properties count
+            if (value == null)
```
Shazwazza, line 54: I guess this is a fix for potential bug?

src/Umbraco.Web/PublishedCache/NuCache/DataSource/DatabaseDataSource.cs
```diff
@@ -21,8 +21,13 @@ namespace Umbraco.Web.PublishedCache.NuCache.DataSource
     internal class DatabaseDataSource : IDataSource
     {
         private const int PageSize = 500;
+        private readonly DatabaseDataSourceConfiguration _dataSourceConfiguration;
 
         // we want arrays, we want them all loaded, not an enumerable
+        public DatabaseDataSource(DatabaseDataSourceConfiguration dataSourceConfiguration)
+        {
+            _dataSourceConfiguration = dataSourceConfiguration;
```
Shazwazza, line 29: null check would be good

src/Umbraco.Web/PublishedCache/NuCache/DataSource/DatabaseDataSource.cs
```diff
@@ -64,8 +69,50 @@ private Sql<ISqlContext> ContentSourcesSelect(IScope scope, Func<Sql<ISqlContext
...
+            .AndSelect<ContentVersionDto>("pcver", x => Alias(x.Id, "PublishedVersionId"), x => Alias(x.Text, "PubName"), x => Alias(x.VersionDate, "PubVersionDate"), x => Alias(x.UserId, "PubWriterId"))
+            .AndSelect<DocumentVersionDto>("pdver", x => Alias(x.TemplateId, "PubTemplateId"))
+
+            .AndSelect<ContentNuDto>("nuPub", x => Alias(x.Data, "PubData"))
+
+
+            .From<NodeDto>();
+
+
+            // TODO: I'm unsure how we can format the below into SQL templates also because right.Current and right.Published end up being parameters
```
Shazwazza, line 93: My memory is blurred on this one but I feel like we solved this with another one of your PRs a while ago? Not sure if you remember or not. We can always deal with that later though.

src/Umbraco.Web/PublishedCache/NuCache/DataSource/DatabaseDataSourceConfiguration.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Umbraco.Web.PublishedCache.NuCache.DataSource
+{
+    public class DatabaseDataSourceConfiguration
+    {
+        public DatabaseDataSourceConfiguration(bool publishedContentOnly = false)
+        {
+            PublishedContentOnly = publishedContentOnly;
+        }
+        public bool PublishedContentOnly { get; private set; }
```
Shazwazza, line 15: Need some code comments so future devs know what this is configuring and the outcome and prob best to be internal for now since it's only used internally. This will all have to be converted to public IOptions in netcore project.

src/Umbraco.Web/Routing/PublishedRouter.cs
```diff
@@ -534,13 +534,13 @@ private bool FollowInternalRedirects(PublishedRequest request)
             {
                 // bad redirect - log and display the current page (legacy behavior)
                 _logger.Debug<PublishedRouter>("FollowInternalRedirects: Failed to redirect to id={InternalRedirectId}: value is not an int nor a GuidUdi.",
-                    request.PublishedContent.GetProperty(Constants.Conventions.Content.InternalRedirectId).GetSourceValue());
+                    request.PublishedContent.GetProperty(Constants.Conventions.Content.InternalRedirectId)?.GetSourceValue());
```
Shazwazza, line 537: These null checks here kinda seem like they should be there anyways?

Shazwazza (changes requested):
Makes sense! Have left a few questions inline

nzdev:
This didn't work as written so I have withdrawn this pr. This is still an issue to address however as it will make a huge impact on performance/boot/memory for frontend instances. @Shazwazza

src/Umbraco.Web/PublishedCache/NuCache/DataSource/BTree.DictionaryOfPropertyDataSerializer.cs
nzdev: Yes. Though it does end up being null for this implementation

Shazwazza:
I agree, the theory is great and would be very good to be able to achieve this. But I know there are some internal requirements scattered around for draft data to be there that we'll need to figure out.

PR #9452, opened 2020-11-25
Netcore: Fix unit tests, so they pass on linux too

bergmania:
Fixed unit tests so the pass on both windows and linux

### DIFF
--- a/src/Umbraco.Tests.UnitTests/Umbraco.Infrastructure/HostedServices/TempFileCleanupTests.cs
+++ b/src/Umbraco.Tests.UnitTests/Umbraco.Infrastructure/HostedServices/TempFileCleanupTests.cs
@@ -15,7 +15,7 @@ namespace Umbraco.Tests.UnitTests.Umbraco.Infrastructure.HostedServices
     public class TempFileCleanupTests
     {
         private Mock<IIOHelper> _mockIOHelper;
-        private string _testPath = @"c:\test\temp\path";
+        private string _testPath = Path.GetTempPath();
 
         [Test]
         public async Task Does_Not_Execute_When_Not_Main_Dom()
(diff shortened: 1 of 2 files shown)


rustybox:
Not relevant to your PR but Is there a plan to get integration tests running at some point? 

perhaps using the ms sql docker images or something?

bergmania:
> Not relevant to your PR but Is there a plan to get integration tests running at some point?
> perhaps using the ms sql docker images or something?

Currently, we don't have a plan.  I would love it to run on Linux, but it is not part of the scope for the Unicore project.

Even using a docker image, we would need functionality to create and drop databases, i guess?

rustybox:
Is that any different to using LocalDb, LocalDbTestDatabase just calls DatabaseSchemaCreator.RebuildSchema, this should work for say sql developer or sql express also shouldn't it?

bergmania:
Sorry, the `DatabaseSchemaCreator.RebuildSchema` should work without changes. By databases, I actually meant database instances/servers like the `LocalDb` class, which is used to create LocalDB instances

rustybox:
yeah but that class just connects to localdb, calls create database in a sql command or copies around an mdf if it already exists.

bergmania:
> yeah but that class just connects to localdb, calls create database in a sql command or copies around an mdf if it already exists.

It also uses the underlying `SqlLocalDB.exe`..

rustybox:
But you wont need that because localdb wouldn't be used ?

bergmania:
That's correct, but we still need functionality to somehow get or create the instance from the docker? Or would you "hardcode" that path/connection sting to the db in the docker image?

rustybox:
master connection string I'd have in a pipeline variable + docker-compose file, set it as an environment variable for use by csharp.

rustybox:
Here you go

```yaml
version: '3.1'

services:

  db:
    image: mcr.microsoft.com/mssql/server:2019-CU8-ubuntu-16.04
    restart: always
    environment:
      ACCEPT_EULA: Y
      SA_PASSWORD: bergmanias_super_secure_password
```

bergmania:
I guess that could work. But then the requirement to run integration tests would be to have docker running (as linux containers), instead of LocalDB. This would be the same for windows, mac and linux, right?

rustybox:
Doesn't have to be, you can use LocalDb for windows if you want, bit of conditional logic.

It'd just be nice if the integration tests on the CI server tested it works on Linux,

PR #9456, opened 2020-11-25
NetCore: Apply volatile attribute

nikolajlauridsen:
I've created an UmbracoVolatile attribute and an UmbracoSuppressVolatile attribute. 
 
I've then compared the V8 and NetCore branch and applied the UmbracoVolatile attribute to all classes, interfaces, methods etc. that was internal or private in V8 but is public in NetCore.

This ties in with the VolatileAnalyzer in Umbraco.Code, which will report and error if any code marked with the UmbracoVolatile attribute is used outside the Umbraco projects, essentially acting as a pseudo internal.

To make all this work I've installed or updated Umbraco.Code in most projects.


src/Umbraco.Core/CodeAnnotations/UmbracoSuppressVolatileAttribute.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+
+namespace Umbraco.Core.CodeAnnotations
+{
+    /// <summary>
+    /// Assembly level attribute used to suppress <see cref="UmbracoVolatileAttribute"/> errors to warnings.
```
bergmania, line 6: I think we need more info in this file.. More info about it is used by Umbraco.Core package etc. bla.bla.bla :) Otherwise, some would think it could be removed 🙈

src/Umbraco.Core/CodeAnnotations/UmbracoVolatileAttribute.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Umbraco.Core.CodeAnnotations
+{
+    /// <summary>
+    /// An empty attribute which is used as a marker for Roslyn Analyzers.
+    /// </summary>
+    /// <remarks>
+    /// Any class or method marked with this attribute will throw a Volatile error which can be suppressed
+    /// with <see cref="UmbracoSuppressVolatileAttribute"/>, this is intended for classes which are public
+    /// for testing purposes, but should not be accessed otherwise
+    /// </remarks>
```
bergmania, line 12: Again, just a small note about the relation to the Umbraco.Code

src/Umbraco.Core/Models/Mapping/ContentPropertyMapDefinition.cs
```diff
@@ -11,7 +12,8 @@ namespace Umbraco.Web.Models.Mapping
     /// <summary>
     /// A mapper which declares how to map content properties. These mappings are shared among media (and probably members) which is
     /// why they are in their own mapper
-    /// </summary>
+    /// </summary>'
```
bergmania, line 15:
```suggestion
    /// </summary>
```

src/Umbraco.Core/Umbraco.Core.csproj
```diff
@@ -43,5 +44,6 @@
         <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
             <_Parameter1>DynamicProxyGenAssembly2</_Parameter1>
         </AssemblyAttribute>
+        <AssemblyAttribute Include="Umbraco.Core.CodeAnnotations.UmbracoSuppressVolatileAttribute"></AssemblyAttribute>
```
bergmania, line 47: Was this required, now our own assemblies are marked in the roslyn analyser ?

src/Umbraco.Infrastructure/Umbraco.Infrastructure.csproj
```diff
@@ -92,6 +93,7 @@
         <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
             <_Parameter1>DynamicProxyGenAssembly2</_Parameter1>
         </AssemblyAttribute>
+        <AssemblyAttribute Include="Umbraco.Core.CodeAnnotations.UmbracoSuppressVolatileAttribute"></AssemblyAttribute>
```
bergmania, line 96: Is this requeired?

src/Umbraco.ModelsBuilder.Embedded/Umbraco.ModelsBuilder.Embedded.csproj
```diff
@@ -33,5 +34,6 @@
         <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
             <_Parameter1>Umbraco.Tests.Integration</_Parameter1>
         </AssemblyAttribute>
+        <AssemblyAttribute Include="Umbraco.Core.CodeAnnotations.UmbracoSuppressVolatileAttribute"></AssemblyAttribute>
```
bergmania, line 37: Required?

src/Umbraco.Persistance.SqlCe/Umbraco.Persistance.SqlCe.csproj
```diff
@@ -29,4 +30,7 @@
       <None Remove="obj\**" />
     </ItemGroup>
 
+    <ItemGroup>
+        <AssemblyAttribute Include="Umbraco.Core.CodeAnnotations.UmbracoSuppressVolatileAttribute"></AssemblyAttribute>
```
bergmania, line 34: Required?

src/Umbraco.PublishedCache.NuCache/Umbraco.PublishedCache.NuCache.csproj
```diff
@@ -26,13 +27,14 @@
...
             <_Parameter1>Umbraco.Tests.Integration</_Parameter1>
-        </AssemblyAttribute>        
+        </AssemblyAttribute>
         <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
             <_Parameter1>Umbraco.Tests.Benchmarks</_Parameter1>
         </AssemblyAttribute>
         <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
             <_Parameter1>DynamicProxyGenAssembly2</_Parameter1>
         </AssemblyAttribute>
+        <AssemblyAttribute Include="Umbraco.Core.CodeAnnotations.UmbracoSuppressVolatileAttribute"></AssemblyAttribute>
```
bergmania, line 37: Required?

src/Umbraco.Web.BackOffice/Umbraco.Web.BackOffice.csproj
```diff
@@ -36,4 +37,8 @@
         <ProjectReference Include="..\Umbraco.Infrastructure\Umbraco.Infrastructure.csproj" />
         <ProjectReference Include="..\Umbraco.Web.Common\Umbraco.Web.Common.csproj" />
     </ItemGroup>
+
+    <ItemGroup>
+        <AssemblyAttribute Include="Umbraco.Core.CodeAnnotations.UmbracoSuppressVolatileAttribute"></AssemblyAttribute>
```
bergmania, line 42: Required?

bergmania (changes requested):
Looks good.. 
Some things we need to consider:
 - Should it be part of our nuget packages (nuspec) files? So users of Umbraco automatic has Umbraco.Code as a dependency. Otherwise I guess the attribute is just ignored
 - Show we also mark with `[EditorBrowsable(EditorBrowsableState.Never)]` ?

src/Umbraco.Core/Umbraco.Core.csproj
nikolajlauridsen: No that's very true, now that they're are marked as allowed, they shouldn't report any errors, I'll remove them again.

src/Umbraco.Core/CodeAnnotations/UmbracoSuppressVolatileAttribute.cs
nikolajlauridsen: That's very true, I added a comment to the top of the file explaining what its purpose is :)

src/Umbraco.Core/CodeAnnotations/UmbracoVolatileAttribute.cs
nikolajlauridsen: I added a small note that the analyzer resides in the Umbraco.Code package and that it must be installed for the analyzer to have any effect.

nikolajlauridsen:
> Some things we need to consider:
> 
> * Should it be part of our nuget packages (nuspec) files? So users of Umbraco automatic has Umbraco.Code as a dependency. Otherwise I guess the attribute is just ignored
> * Show we also mark with `[EditorBrowsable(EditorBrowsableState.Never)]` ?

I definitely think it should be part of the nuspec files, since the analyzer only has an effect if the Umbraco.Code package is installed. The only thing to consider is that the Umbraco.Code package also contains the MapAll analyzer, however this analyzer is only activated by using the comment `// Umbraco.Code.MapAll`, so it shouldn't be a problem. 

I also think it makes good sense to mark UmbracoVolatileAttribute with `[EditorBrowsable(EditorBrowsableState.Never)]` since it has no real purpose outside the CMS projects.

I don't know about `UmbracoSuppressVolatileAttribute` since it's intended to be used outside the CMS projects, and it might not be apparent that it exists, it is however still described in the documentation, so I don't know.

bergmania:
> I also think it makes good sense to mark UmbracoVolatileAttribute with `[EditorBrowsable(EditorBrowsableState.Never)]` since it has no real purpose outside the CMS projects.

I actually thing all methods etc marked with `UmbracoVolatileAttribute`. Just to behave more like internal methods do

nikolajlauridsen:
Ahh yes, that makes good sense, if you're not supposed to use it, why show it to you then?

JimBobSquarePants:
I think before adding this attribute to hundreds of files I think a discussion is required about what the problem is that is being solved. 

Why does a volatile analyzer exist over correctly structuring code accessibility?

bergmania:
I Closed this PR. We need to change lot of the now public classes back to internal. Some of these still require us to move files from infrastructure to core.

Issue #9460, opened 2020-11-27
Ensure GetByIds in resources support Guid and Udi


bjarnef:
In `entityResource` using EntityController the `GetByIds` support both `int[]`, `Guid[]` and `Udi[]`:
https://github.com/umbraco/Umbraco-CMS/blob/0bd9e3ca9955844389556ce512fa0aab181703fb/src/Umbraco.Web/Editors/EntityController.cs#L79

However it doesn't seem to be the case for `mediaResource` using MediaController, which only have a single method to fetch multiple images by `int[]`:
https://github.com/umbraco/Umbraco-CMS/blob/0bd9e3ca9955844389556ce512fa0aab181703fb/src/Umbraco.Web/Editors/MediaController.cs#L177-L187

The same seems to be the case for `contentResource` using ContentController:
https://github.com/umbraco/Umbraco-CMS/blob/0bd9e3ca9955844389556ce512fa0aab181703fb/src/Umbraco.Web/Editors/ContentController.cs#L94-L104

It could be relevant to other resources as well, but the `contentResource` and `mediaResource` is the most important ones.

bjarnef:
I have splitted this into two PRs.

**Content and Media**
https://github.com/umbraco/Umbraco-CMS/pull/9465

**Member Groups**
https://github.com/umbraco/Umbraco-CMS/pull/9467

umbrabot:
Closing this in relation to #9467 - make sure to read the close reason on #9467

PR #9465, opened 2020-11-28
Ensure both content and media support fetching entities by ids array of Guid and Udi

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9460

### Description
This PR makes it possible for `contentResource` and `mediaResource` to fetch entities by `int[]`, `Guid[]` and `Udi[]` like `entityResource`.

I noticed we have a `GetByIds(this IContentService contentService, IEnumerable<Udi> ids)` in `ContentServiceExtensions` to extend `ContentService` with `GetByIds(IEnumerable<Udi> ids)` without changing the interface as it would a breaking change. Similar method exists for media, but it wasn't included in the project. It is now.

I have changed `getByIds` in `contentResource` and `mediaResource` to POST request similar to `getByIds` in `entityResource`.

Here is an example fetching entities using `Guid[]` and `getByIds` in the three different services.

image

In have added the dashboard as a zip-file to test with: 
[MyDashboard.zip](https://github.com/umbraco/Umbraco-CMS/files/5610745/MyDashboard.zip)


bjarnef:
It would also be great to support this in `getByIds` in `memberGroupResource`, but it seems to require some changes in `IMemberGroupService` and `IMemberGroupRepository` as well to allow fetching member group nodes by `Guid`.

bjarnef:
@nul800sebastiaan I just tried based on an Umbraco v8.13.1 project to include `Umbraco.Core.Services`.
Unfortunately `MediaServiceExtensions.cs` hasn't been included in the project, so the extension `GetByIds(Udi[] udis)` isn't available.

image

nul800sebastiaan:
Am I missing something? This PR is still open so yeah, it won't have the changes you made. :)

bjarnef:
@nul800sebastiaan yeahit enhanced a bit around GetByIds via Udi's ... however this file should be safe to include :)
https://github.com/umbraco/Umbraco-CMS/pull/9465/files#diff-241a5f45f2f7a63af9c68058564c5c2f7b78e414ac6be46aebde98cd8e3d106eR194

From what I can find this was added a long time ago
https://github.com/umbraco/Umbraco-CMS/commit/d54658009cdeb676257747bf0992599feee24554#diff-0703ef93a614d4373794998ca03f46e6c6703b6abe0cdfaecf5642f36f0ea30e but never included in the project.

It is however very similar to the ContentService extensions here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Services/ContentServiceExtensions.cs

bjarnef:
@nul800sebastiaan I think this is one of the fra missing parts to access by Guid and Udi.. and furthermore it would allow to access more content and media.

bjarnef:
@nul800sebastiaan I appreciate many new PR's are reviewed and merged, but it seems many older PR's are a bit forgotten?

Maybe PR's should have a label based on the complexity or priority? E.g. some are a minor typo, localization or styling change, while other are regarding performance, security, new feauture which may contains hundreds or thousands of changes.

bjarnef:
Not sure which file is missing from the commit 😄
The `MediaServiceExtensions.cs` was never included in the project, but has been fixed in the PR instead: https://github.com/umbraco/Umbraco-CMS/pull/10883

It wouldn't cause breaking change if the resource was used, but probably if developer had requests directly to these controller methods or used angular interceptors is expected a `GET` request.

It should be simple enough the update this to still use `GET`, while making `GetByIds` support both `int[]`, `Guid[]`, and `Udi[]` ... not just `int[]`, although limited to maximum URL length (where `POST` would be ideal to request many entities).

@nul800sebastiaan do we still want this (without breaking changes in v8) or only v9?

nul800sebastiaan:
v9 is where we'll focus from now on. We're trying to wrap up v8.18 and had to make choices on what would still make it in there. I didn't look at this one specifically, but yes, this would definitely be great for v9!

PR #9467, opened 2020-11-29
Ensure membergroup support fetching by ids array of Guid and Udi

bjarnef:
### Prerequisites

### Description
In addition to https://github.com/umbraco/Umbraco-CMS/pull/9465 we also have `getByIds` in `memberGroupResource` which also supported array of `int`, but it is useful to fetch these by `Guid` and `Udi` as well.

Currently with have extensions methods for `GetByIds(IEnumerable<Udi> ids)` in `ContentServiceExtensions.cs` and `MediaServiceExtensions.cs` located in https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Core/Services

I have added a similar extension method for `IMemberGroupService`, but maybe these should be moved to a subfolder (not sure if the namespace should change or we should ensure it still belong to `Umbraco.Core.Services` namespace.

Maybe there another way to handle the `GetByIds(IEnumerable<Udi> ids)` extension method since I guess many services might need this now or in future. Could the service interface inherits this from another interface?


bjarnef:
To test this create a few member groups and change this guids in the `memberGroupGuids` variable in the dashboard controller in this example: 
[MyDashboard.zip](https://github.com/umbraco/Umbraco-CMS/files/5612114/MyDashboard.zip)

@nul800sebastiaan @Shazwazza I could use a bit help with this one to ensure the `GetMany()` method in `MemberGroupRepository` is correct.

Currently it inherits from `IReadRepository<Guid, IMemberGroup>` as well as some other repositories to support `Guid`.

image

But it seems I am missing something to get the use of the current `GetMany()` in `MemberGroupService` to work correct.

image

bjarnef:
It think it should be good now 👍 

image

Furthermore I have replaced some of the queries in `MemberGroupRepository` using strongly typed queries in NPoco.

bjarnef:
@nul800sebastiaan it would be great if we could access member groups by Guid as well 👍

Issue #9471, opened 2020-11-30
NetCore: Members functionality needs to be migrated to .NET Core


emmagarland:
## Members functionality needs to be migrated to .NET Core

Members and related functionality needs to be migrated to .NET Core. More detail can be fleshed out from this issue but essentially we will be:

- Using the ASP.Net Core identity implementation
- Creating `MemberManager` for the *Manager*
- Creating `MembersUserStore` for the *Store*
- Creating `MembersIdentityUser` for the _User_ class
- Removing/moving code from the `MembershipProvider`
- Removing/moving code the `MembershipHelper`
- Attempt to keep the `MemberService` functional with the same signatures
- Unit and integration tests to be added for all relevant areas

There are wider areas to complete around:

- Roles
- Public Access
- Passwords
- Back Office
- Authentication/Authorization
- Macros and views
- Enabling external logins for members (TBC re scope)

Will be easier to create these as individual GitHub issues and link to this main issue.

rustybox:
I know that this is how things are now, but otherwise is there good reason to split the usermanagers? An UmbracoMember and an UmbracoUser are both just users with different claims.

Shazwazza:
Users and members are entirely different things though. They have different storage systems, different properties, different everything. Members are based on member types which are dynamic for each instance and users are not. Combining these forces has been floated around for a while and though it might be possible in some way and shape, I'm not convinced that its the better approach. It's also hugely out of the scope for the first netcore release.

rustybox:
> It's also hugely out of the scope for the first netcore release.

Fair

Shazwazza:
We can defo revisit when all is done. Will be the first time that both members and users are based on identity, it's almost 2021 😛

rustybox:
One final protest, I won't moan any further but looking at the current setup in BackOfficeServiceCollectionExtensions which is a bit of a hacky mess, and the PR which aims to duplicate that setup, I honestly believe this is the wrong approach.

> Users and members are entirely different things though. 

Authentication, Authorization, object ACL, what's different.

> They have different storage systems, different properties, different everything.

Why different storage systems?
Different properties == different claims.

> Members are based on member types which are dynamic for each instance and users are not.

Users have fixed set of claims, members have dynamic set of claims.

> It's also hugely out of the scope for the first netcore release.

I think it is a decrease in scope to only have one UserManager, one PasswordHasher, etc etc

Shazwazza:
[...]

PR #9472, opened 2020-11-30
Initial creation of vertical slice for create member 

emmagarland:
Initial creation of vertical slice for create member identity user, store and manager. 

Based on the new ASP.NET Core identity implementation for the back office, and [UmbracoIdentity](https://github.com/Shazwazza/UmbracoIdentity).

### Prerequisites

This relates to the larger issue for the whole members migration #9471

### Description

- Initial vertical slice for the back-office **Create Member** functionality (needs fleshing out) to make sure the pattern is correct to follow from
- No database changes
- Started the initial implementation of a store, manager and user for the UmbracoMembers identity implementation
- Started to adjust the controller to use the new manager
- Started with unit tests for new code

To test:

- Run unit tests
- Run integration tests
- Login to back-office and create a member. Member should be persisted to the database as it is currently in Umbraco 8.
- Login to back-office and update a member. Member updates should be persisted to the database as they are in Umbraco 8.
- Login to back-office and delete a member. Member should be deleted from database


src/Umbraco.Infrastructure/Members/UmbracoMembersUserManager.cs
```diff
@@ -0,0 +1,244 @@
...
+            string password = _passwordGenerator.GeneratePassword();
+            return password;
+        }
+
+        /// <summary>
+        /// Helper method to validate a password based on the current password validator
+        /// </summary>
+        /// <param name="password">The password to update</param>
+        /// <returns>The validated password</returns>
+        public async Task<List<IdentityResult>> ValidatePassword(string password)
```
scottbrady91, line 232: **question:** this one already exists on the UserManager, albeit with a slightly different signature. Do we need this version?

src/Umbraco.Web.BackOffice/Extensions/UmbracoMembersUserServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+    public static class UmbracoMembersUserServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Adds the services required for using Umbraco Members Identity
+        /// </summary>
+        /// <param name="services"></param>
+        public static void AddUmbracoMembersIdentity(this IServiceCollection services)
+        {
+            services.BuildUmbracoMembersIdentity()
+                .AddDefaultTokenProviders()
```
scottbrady91, line 21: **question:** does the User version also call this? We may need a Members specific version of this, as this adds providers to IdentityOptions in the DI: https://github.com/dotnet/aspnetcore/blob/master/src/Identity/Extensions.Core/src/IdentityBuilder.cs#L126

src/Umbraco.Web.BackOffice/Extensions/UmbracoMembersUserServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+                .AddUserStore<UmbracoMembersUserStore>()
+                .AddUserManager<IUmbracoMembersUserManager, UmbracoMembersUserManager>();
```
scottbrady91, line 23: **clarification:** could you point me to the implementation of this `AddUserManager` extension method?

scottbrady91: Found it. Silly Scott 🙄

src/Umbraco.Web.BackOffice/Extensions/UmbracoMemberIdentityBuilderExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        /// <summary>
+        /// Adds a <see cref="UserManager{TUser}"/> for the <seealso cref="UserType"/>.
+        /// </summary>
+        /// <typeparam name="TUserManager">The type of the user manager to add.</typeparam>
+        /// <typeparam name="TInterface"></typeparam>
+        /// <returns>The current <see cref="IdentityBuilder"/> instance.</returns>
+        public static IdentityBuilder AddUserManager<TInterface, TUserManager>(this IdentityBuilder identityBuilder) where TUserManager : UserManager<UmbracoMembersIdentityUser>, TInterface
+        {
+            identityBuilder.AddUserManager<TUserManager>();
```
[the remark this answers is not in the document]
emmagarland: Good spot, will remove. It is also in _src\Umbraco.Web.BackOffice\Extensions\IdentityBuilderExtensions.cs_ _- should that remain the default user manager then? And I keep this extension but only with the one line?

src/Umbraco.Web.BackOffice/Extensions/UmbracoMembersUserServiceCollectionExtensions.cs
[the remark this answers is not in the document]
emmagarland: In `UmbracoBuilderExtensions.cs` I created a new method, `AddMembersIdentity`, The backoffice user version has it's own method `AddBackOfficeIdentity`. Both are added in `AddAllBackOfficeComponents` method in the same class.

src/Umbraco.Infrastructure/Members/UmbracoMembersUserManager.cs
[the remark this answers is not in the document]
emmagarland: This has since been removed

src/Umbraco.Web.Common/Security/MembersUserManager.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        public MembersUserManager(
+            IIpResolver ipResolver,
+            IUserStore<MembersIdentityUser> store,
+            IOptions<MembersIdentityOptions> optionsAccessor,
+            IPasswordHasher<MembersIdentityUser> passwordHasher,
+            IEnumerable<IUserValidator<MembersIdentityUser>> userValidators,
+            IEnumerable<IPasswordValidator<MembersIdentityUser>> passwordValidators,
+            //TODO: do we need members versions of this?
+            BackOfficeLookupNormalizer keyNormalizer,
```
Shazwazza, line 34: Nope just needs to be named more generic. We also have 2x test cases for the same exact thing: NopLookupNormalizerTests + BackOfficeLookupNormalizerTests both are just testing BackOfficeLookupNormalizer :P 

Don't really think we need to pass this in with DI and instead just pass a new instance of this in the base class.

src/Umbraco.Web.Common/Security/MembersUserManager.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        private IdentityAuditEventArgs CreateArgs(AuditEvent auditEvent, IPrincipal currentUser, string affectedUserId, string affectedUsername)
+        {
+            var currentUserId = GetCurrentUserId(currentUser);
+            var ip = IpResolver.GetCurrentRequestIpAddress();
+            return new IdentityAuditEventArgs(auditEvent, ip, currentUserId, string.Empty, affectedUserId, affectedUsername);
+        }
+
+
+        // TODO: As per backoffice, review where these are raised and see if they can be simplified and either done in the this usermanager or the signin manager,
```
Shazwazza, line 59: For members we don't really have this type of auditing so it's not required. Lets remove this for now and revisit later if we need any auditing on member info at the level of users. Also since the user events will be entirely refactored to use IEventAggregator anyways.

Shazwazza (commented):
So far so good. We'll definitely want to share code where we can - seems like a lot of places this can be done - but that can be refactored as we go along. I've posted a bunch of comments inline. Cheers!

emmagarland:
> So far so good. We'll definitely want to share code where we can - seems like a lot of places this can be done - but that can be refactored as we go along. I've posted a bunch of comments inline. Cheers!

Thanks @Shazwazza for the review! I'll make the changes/reply and get this into a draft PR ASAP so we can merge and work from here. Definitely will be room to share code with the Users identity work especially since the recent updates.

src/Umbraco.Core/Services/IMembershipRoleService.cs
```diff
@@ -9,6 +10,7 @@ public interface IMembershipRoleService<out T>
     {
         void AddRole(string roleName);
         IEnumerable<string> GetAllRoles();
+        IEnumerable<IMemberGroup> GetAllRolesTyped();
```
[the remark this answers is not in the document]
emmagarland: I was thinking of removing it, but was wondering if I'd impact something I didn't know about. I'll see if I can switch to just this typed one and what it impacts.

src/Umbraco.Infrastructure/Security/IUmbracoUserManager.cs
```diff
@@ -223,12 +224,42 @@ public interface IUmbracoUserManager<TUser> : IDisposable
...
         /// </summary>
         /// <returns>A generated password</returns>
         string GeneratePassword();
 
+        /// <summary>
+        /// Generates a hashed password for a null user based on the default password hasher
+        /// </summary>
+        /// <param name="password">The password to hash</param>
+        /// <returns>The hashed password</returns>
+        string GeneratePassword(string password);
```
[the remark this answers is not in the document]
emmagarland: Yes, good shout! Will change

src/Umbraco.Infrastructure/Security/MembersUserStore.cs
```diff
@@ -0,0 +1,703 @@
...
+            ThrowIfDisposed();
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            // create member
+            // TODO: are we keeping this method? The user service creates the member directly
+            // but this way we get the member type by alias first
+            IMember memberEntity = _memberService.CreateMember(
```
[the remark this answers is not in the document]
emmagarland: Yes I agree, I felt like I wanted to replace it but was trying to stick to 1-to-1 changes. So this would be a great future task to replace these methods.

src/Umbraco.Infrastructure/Security/MembersUserStore.cs
```diff
@@ -0,0 +1,703 @@
...
+            if (found == null)
+            {
+                return null;
+            }
+
+            return new IdentityUserLogin<string>
+            {
+                LoginProvider = found.LoginProvider,
+                ProviderKey = found.ProviderKey,
+                ProviderDisplayName = found.ProviderDisplayName, // TODO: We don't store this value so it will be null
```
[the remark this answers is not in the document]
emmagarland: This is shared from the back office implementation and  is part of Microsoft IdentityUserLogin that _"Gets or sets the friendly name used in a UI for this login."_ but it isn't apparently used. It isn't set in your [UmbracoIdentity package](https://github.com/Shazwazza/UmbracoIdentity/blob/c12d07e1d3bf1aa2a7bdaba8184b1eacec9c0bce/src/UmbracoIdentity/UmbracoMembersUserStore.cs) either, so maybe unneeded? It would be a string more things like "Google", "local" etc. 

However, some of these external login _FindUserLoginAsync_ and similar methods need reviewing; they may be unwanted or could be combined during the code combining refactor.

Adding and removing external user logins etc hasn't yet been implemented or tested. I'm only implementing this and similar methods for members user store at this stage because this class inherits from identity _UserStoreBase_ and I thought I'd make a start on adding the body of some of the methods, following the `BackOfficeUserStore` approach.
 
It might be safer to throw a not implemented exception before they're implemented? 

I'm not sure the final plan for external logins for members yet so I'm happy to amend as needed.

src/Umbraco.Web.Common/Security/MembersUserManager.cs
emmagarland: Gotcha. Will merge the duplicate tests and rename to generic - `NoOpLookupNormalizer` perhaps?

Then I'll look into the shift away from DI

emmagarland: As an aside there is a TODO on the normalizer that says "Do we need this" - I'd like to delete the TODO, what do we need to know to know if we need it?

src/Umbraco.Web.Common/Security/MembersUserManager.cs
emmagarland: OK, removing all member auditing events for now. However this still leaves the following from the `IUmbracoUserManager` interface, although there is a comment saying we will revisit these anyway. I'll leave them in for now, unimplemented, until we do that:

image

src/Umbraco.Web.Common/Security/MembersUserManager.cs
emmagarland: Another side note, `BackOfficeIdentityErrorDescriber` is empty at the moment. Do I need to create an also empty `MemberIdentityErrorDescriber` or should I leave it for now?

src/Umbraco.Infrastructure/Security/MembersUserStore.cs
```diff
@@ -0,0 +1,703 @@
...
+            _mapper = mapper;
+            _scopeProvider = scopeProvider;
+        }
+
+        /// <summary>
+        /// Not supported in Umbraco
+        /// </summary>
+        /// <inheritdoc />
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override IQueryable<MembersIdentityUser> Users => throw new NotImplementedException();
```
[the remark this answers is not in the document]
emmagarland: Interesting. I'll have a think/chat and get back to you!

The shared UserManager base class actually is already doing the second suggestion of overriding this property, for example:
 
`public override bool SupportsQueryableUsers => false; `
 
So are you saying we could override all the "Supports" properties on the shared `UserManager` to explain what our user stores support? e.g. SupportsUserTwoFactorRecoveryCodes, SupportsUserPhoneNumber etc set to false?
 
I'm not sure re no longer using the Microsoft identity UserStoreBase class - even if it means we have to specify what we don't support - will speak to the team tomorrow morning and let you know what we think.

src/Umbraco.Infrastructure/Security/IUmbracoUserManager.cs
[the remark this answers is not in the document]
emmagarland: Changed in recent push

src/Umbraco.Core/Services/IMembershipRoleService.cs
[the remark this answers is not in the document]
emmagarland: Removed in recent push - along with some formatting as per style rules.

(NB, I've also just corrected a bunch of namespaces and this will look like more files changed than I've actually changed for members functionality)

src/Umbraco.Infrastructure/Search/ExamineComponent.cs
```diff
@@ -515,6 +515,7 @@ private void ReIndexForContent(IContent sender, bool isPublished)
                 DeferedReIndexForContent.Execute(this, sender, isPublished);
         }
 
+        //TODO: this causes an exception when deleting a member
```
emmagarland, line 518: @Shazwazza I'm finding that when I break on all exceptions after deleting a member, I come to this point (line 258) in the `ExamineComponent.cs`, where the member has already been deleted but it tries to reindex it. 

image

The member service is returning a null member object even though it hasn't matched anything against the Payload (populated with the now deleted member details)

image

This means it tries to run the `ReindexForMember` method but fails because the properties are null. 

Is this something you have seen before or something new introduced by this update do you think?

src/Umbraco.Core/Models/Mapping/MemberMapDefinition.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Umbraco.Core.Models.Mapping
+{
+    /// <inheritdoc />
+    public class MemberMapDefinition : IMapDefinition
+    {
+        /// <inheritdoc />
+        public void DefineMaps(UmbracoMapper mapper) => mapper.Define<MemberSave, IMember>(Map);
+
+        //TODO: put this here instead of a new mapper definition (like user). Can move
```
emmagarland, line 12: @Shazwazza @bergmania I needed to map the MemberSave (content item member) to a member, so I added this definition, although it doesn't yet have groups.

I didn't know if this was the right place to put it, so I've flagged it with a TODO to make sure I'm mapping everything correctly.

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
```diff
@@ -286,172 +308,256 @@ public MemberDisplay GetEmpty(string contentTypeAlias = null)
...
+                throw new InvalidOperationException($"No member type found with alias {contentItem.ContentTypeAlias}");
+            }
+
+            var identityMember = MembersIdentityUser.CreateNew(
+                contentItem.Username,
+                contentItem.Email,
+                memberType.Alias,
+                contentItem.Name);
+
+            // TODO: may not need to add password like this
```
emmagarland, line 354: @Shazwazza I added this `CreateAsync` overload to include the new password because we are creating it in the back-office - unlike the User creation (which creates the password with ___UIDEMPTYPWORD___ appended). Presume that is right?

src/Umbraco.Infrastructure/Security/UmbracoUserManager.cs
```diff
@@ -111,6 +111,54 @@ public string GeneratePassword()
...
+        }
+
+        /// <summary>
+        /// Used to validate the password without an identity user
+        /// Validation code is based on the default ValidatePasswordAsync code
+        /// Should return <see cref="IdentityResult.Success"/> if validation is successful
+        /// </summary>
+        /// <param name="password">The password.</param>
+        /// <returns>A <see cref="IdentityResult"/> representing whether validation was successful.</returns>
+        public async Task<IdentityResult> ValidatePasswordAsync(string password)
```
emmagarland, line 135: @Shazwazza needs confirming - is based on the default `ValidatePasswordAsync` code including the error code. I created this so our manager can carry out the validation we need based on the passwordvalidators.

src/Umbraco.Infrastructure/Security/UmbracoUserManager.cs
```diff
@@ -111,6 +111,54 @@ public string GeneratePassword()
             return password;
         }
 
+        /// <summary>
+        /// Generates a hashed password based on the default password hasher
+        /// No existing identity user is required and this does not validate the password
+        /// </summary>
+        /// <param name="password">The password to hash</param>
+        /// <returns>The hashed password</returns>
```
emmagarland, line 119: @Shazwazza brought this out into it's own function in the manager so we can hash the password as required, happy to hear thoughts

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
```diff
@@ -286,172 +308,256 @@ public MemberDisplay GetEmpty(string contentTypeAlias = null)
...
+            }
 
-            throw new NotImplementedException("Members have not been migrated to netcore");
-            // TODO: all member password processing and creation needs to be done with a new aspnet identity MemberUserManager that hasn't been created yet.
-            // set the password
-            //contentItem.PersistedContent.RawPasswordValue = _passwordSecurity.HashPasswordForStorage(contentItem.Password.NewPassword);
-        }
+            if (contentItem.Password != null)
+            {
+                IdentityResult validatePassword = await _memberManager.ValidatePasswordAsync(contentItem.Password.NewPassword);
```
emmagarland, line 433: @Shazwazza @bergmania after a fair amount of debugging I do need a bit of assistance with this. The same issue is happening here in `UpdateMemberAsync` and `ValidateMemberDataAsync` below.

This uses the new `ValidatePasswordAsync` method on Identity member manager for our the validation of a password to determine whether the password is valid. 

- If valid, we hash the password and add it to the identity object to be updated as part of the update result.
- If the password does not validate according the the `PasswordValidators`, I thought we could create a suitable validation error response with the reasons. Something like this (although the formatting of the error message may need work):

image

In this branch, if the password doesn't validate, the error message does displays in the back-office as expected. 

However, my issue is that after this, upon clicking the field it never disappears (unlike other validation messages). I think I'm missing something I need to do in terms of what I pass to the client-side code?

This doesn't happen with users since the password isn't validated in this way, as it is generated rather than manually entered. Other places where a password is manually entered seem to return a slightly different validation object.

I also saw some work happening around validation recently so I didn't want to go too far trying to debug this if it is not the approach we're looking for anyway. 

Any help appreciated!

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
```diff
@@ -286,172 +308,256 @@ public MemberDisplay GetEmpty(string contentTypeAlias = null)
...
-                // var validPassword = await _passwordValidator.ValidateAsync(_passwordConfig, contentItem.Password.NewPassword);
-                // if (!validPassword)
-                // {
-                //     ModelState.AddPropertyError(
-                //        new ValidationResult("Invalid password: " + string.Join(", ", validPassword.Result), new[] { "value" }),
-                //        string.Format("{0}password", Constants.PropertyEditors.InternalGenericPropertiesPrefix));
-                //     return false;
-                // }
+                // TODO: this currently stops the user interacting with the client-side when invalid
+                IdentityResult validPassword = await _memberManager.ValidatePasswordAsync(contentItem.Password.NewPassword);
```
emmagarland, line 470: Same issue as `UpdateMemberAsync` - when password is invalid, client-side message doesn't currently disappear on clicking off the field

src/Umbraco.Web.BackOffice/Trees/MemberTreeController.cs
```diff
@@ -64,6 +64,7 @@ public class MemberTreeController : TreeController, ISearchableTree, ITreeNodeCo
         /// <returns></returns>
         public ActionResult<TreeNode> GetTreeNode(string id, [ModelBinder(typeof(HttpQueryStringModelBinder))]FormCollection queryStrings)
         {
+            //TODO: this is currently throwing an exception when loading a member
```
emmagarland, line 67: @Shazwazza @bergmania the id being passed in is currently null. I'm trying to debug to find out why. The member does load, but an '_Object reference not set to an instance of an object._' exception is thrown first as **id** is null. 

It seems like the `/umbraco/backoffice/umbracotrees/membertree/GetTreeNode/[id]` call isn't working as expected, yet I can see the link being generated. Will need to look into this more unless you have any ideas?

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
```diff
@@ -149,134 +187,118 @@ public MemberListDisplay GetListNodeDisplay(string listName)
...
         /// </summary>
-        /// <param name="key"></param>
-        /// <returns></returns>
+        /// <param name="key">The Guid key of the member</param>
+        /// <returns>The member for display</returns>
         [OutgoingEditorModelEvent]
         public MemberDisplay GetByKey(Guid key)
         {
-            var foundMember = _memberService.GetByKey(key);
+            // TODO: this is not finding the key currently after member creation
```
emmagarland, line 195: Issue that I need to debug, straight after member creation this was throwing an exception. Looking into it.

src/Umbraco.Infrastructure/Security/MembersUserStore.cs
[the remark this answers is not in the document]
emmagarland: @Shazwazza I had a chat with Scott and we're a bit torn, but our thoughts are:

- Having all these overridden methods in the User Manager might feel a little cumbersome. However, we're already doing this in the shared base `UmbracoUserManager` class so each user manager would override with their own requirements if they differed.
- We could move away from our member user store(s) deriving from the Microsoft identity `UserStoreBase` class,
but then we end up implementing and having to maintain our own pattern. We'd also lose out on future potential updates to the Microsoft library.

Overall, I think we'll leave it in this PR as it is now, but something to consider especially if we're sharing more code between the Users and Members implementation. I think this could qualify for a separate GitHub issue to discuss approach.

src/Umbraco.Infrastructure/Search/ExamineComponent.cs
emmagarland: image

Just checked in v8 and same behaviour happens when deleting a member. `DeferedReIndexForMember` is called and then the member properties are null, an exception happens under the hood. So this isn't a recently introduced issue. Should I create a ticket for it in v8?

src/Umbraco.Web.BackOffice/Trees/MemberTreeController.cs
emmagarland: FYI I just replicated this on the default `netcore` branch so this will need a new bug to be created, but is not as part of this PR. cc @bergmania

src/Umbraco.Infrastructure/Search/ExamineComponent.cs
emmagarland: Closing as created related issue #9684

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
bergmania: Seems like an issue introduced in this PR: https://github.com/umbraco/Umbraco-CMS/pull/7263/. I will see if I can find it

emmagarland: @bergmania possibly, although I introduced this password validation functionality so it is hard to tell whether it was already happening. Does the code I have for validation need something else in order to hook into the AngularJS? I followed what was already there. Thanks for looking!

bergmania: @emmagarland, Your code seems to be correct. I hope to find the issue today, with some help from a frontend developer 🤞

emmagarland: Thanks @bergmania for letting me know, fingers crossed!

src/Umbraco.Core/Models/Mapping/MemberMapDefinition.cs
bergmania: Seems correct to me.  Both `MemberSave` and `IMember` exists in Umbraco.Core

src/Umbraco.Infrastructure/Security/MembersUserStore.cs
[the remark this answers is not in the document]
emmagarland: cc @bergmania can we create a new task for this pls?

src/Umbraco.Web.BackOffice/Trees/MemberTreeController.cs
emmagarland: Fixed by @bergmania, added `[FromRoute]`, to stop the ID being empty, many thanks

src/Umbraco.Infrastructure/Security/MembersUserStore.cs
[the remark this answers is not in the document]
bergmania: I have one internal task for that

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
emmagarland: No longer happening

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
emmagarland: Resolving as related to a general frontend issue to be resolved separately (will add ID once pushed)

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
```diff
@@ -288,172 +308,260 @@ public ActionResult<MemberDisplay> GetEmpty(string contentTypeAlias = null)
...
+                if (validatePassword.Succeeded == false)
+                {
+                    return new ValidationErrorResult(validatePassword.Errors.ToErrorMessage());
+                }
 
-        private static void UpdateName(MemberSave memberSave)
-        {
-            //Don't update the name if it is empty
-            if (memberSave.Name.IsNullOrWhiteSpace() == false)
+                string newPassword = _memberManager.HashPassword(contentItem.Password.NewPassword);
```
emmagarland, line 439: @Shazwazza need to consider best approach password changing functionality preferences from back-office. Have done this here to hash the password and update the changed date in line 453

src/Umbraco.Tests.UnitTests/Umbraco.Web.BackOffice/Controllers/MemberControllerUnitTests.cs
```diff
@@ -0,0 +1,439 @@
...
+            out MemberDisplay memberDisplay,
+            ContentSaveAction contentAction)
+        {
+            // arrange
+            MemberType memberType = MemberTypeBuilder.CreateSimpleMemberType();
+            Member member = MemberBuilder.CreateSimpleMember(memberType, "Test Member", "test@example.com", "123", "test");
+            int memberId = 123;
+            member.Id = memberId;
+
+            //TODO: replace with builder for MemberSave and MemberDisplay
```
emmagarland, line 338: I think this can be optimised and use the new builder pattern. I can look into that when I'm further along the line.

src/Umbraco.Infrastructure/Security/MemberRolesUserStore.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Identity;
+using Umbraco.Core.Scoping;
+using Umbraco.Core.Services;
+
+namespace Umbraco.Infrastructure.Security
```
emmagarland, line 11: (Not yet implemented)

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
```diff
@@ -288,172 +310,275 @@ public ActionResult<MemberDisplay> GetEmpty(string contentTypeAlias = null)
...
+
+            // map the save info over onto the user
+            member = _umbracoMapper.Map<MemberSave, IMember>(contentItem, member);
+
+            int creatorId = _backOfficeSecurityAccessor.BackOfficeSecurity.CurrentUser.Id;
+            member.CreatorId = creatorId;
+
+            // assign the mapped property values that are not part of the identity properties
+            string[] builtInAliases = ConventionsHelper.GetStandardPropertyTypeStubs(_shortStringHelper).Select(x => x.Key).ToArray();
+            foreach (ContentPropertyBasic property in contentItem.Properties)
```
emmagarland, line 374: @Shazwazza I added this because the customer member type properties weren't being saved. I've found a lot of ways this can be done around PersistedContent etc, but I went with this - best to have another set of eyes on it.

Also I saw the key being set again elsewhere, but I'm not sure if this is necessary here or not.

src/Umbraco.Web/Security/Providers/MembersMembershipProvider.cs
```diff
@@ -7,13 +7,14 @@
...
 using Umbraco.Core.Services;
-using Umbraco.Core.Models.Membership;
 using Umbraco.Web.Composing;
 using System;
 using Umbraco.Net;
 
 namespace Umbraco.Web.Security.Providers
 {
+    //TODO: Delete: should not be used
+    [Obsolete("We are now using ASP.NET Core Identity instead of membership providers")]
```
emmagarland, line 17: This needs to be deleted ultimately, once all has been replaced in terms of macros. Have obsoleted for now

src/Umbraco.Infrastructure/Security/MembersIdentityUser.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            get => _groups;
+            set
+            {
+                _groups = value.Where(x => x.Alias != null).ToArray();
+
+                var roles = new List<IdentityUserRole<string>>();
+                foreach (IdentityUserRole<string> identityUserRole in _groups.Select(x => new IdentityUserRole<string>
+                {
+                    RoleId = x.Alias,
```
emmagarland, line 91: @Shazwazza Should RoleId actually be the Umbraco group ID? In which case, I'll want to change these all to accept an integrer for the role ID instead

src/Umbraco.Infrastructure/Security/UmbracoUserManager.cs
emmagarland: If we aren't using the old password hasher though, the old member passwords won't work, is this a problem? Do we need another task so we can consider a way to roll this out? cc @Shazwazza @bergmania

src/Umbraco.Infrastructure/Security/MembersIdentityUser.cs
Shazwazza: IIRC changing `IdentityUserRole<string>` to `IdentityUserRole<int>` is problematic and not compatible with some other generic constraints we require. `IdentityUserRole` assumes the same key type for both RoleId and UserId which isn't true if we want to use the alias which in the case of users we do (though I can't recall exactly why off the top of my head, I think identity and permission constraints generally want to use a string name for this and might also avoid an extra DB lookup). Yep for members, this would be member groups. 

If you want to try INT instead go for it, you might just have to play around with how that works or doesn't work, I just remember for users it needed to be string for one reason or another.

src/Umbraco.Infrastructure/Security/MembersUserStore.cs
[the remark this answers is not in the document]
Shazwazza: Cool, we'll revisit this when we do external logins

src/Umbraco.Infrastructure/Security/NoOpLookupNormalizer.cs
```diff
@@ -1,12 +1,12 @@
...
 
-namespace Umbraco.Core.Security
+namespace Umbraco.Infrastructure.Security
 {
 
     /// <summary>
     /// No-op lookup normalizer to maintain compatibility with ASP.NET Identity 2
     /// </summary>
-    public class BackOfficeLookupNormalizer : ILookupNormalizer
+    public class NoOpLookupNormalizer : ILookupNormalizer
```
Shazwazza, line 9: For consistency should be Noop instead of NoOp since Noop is used elsewhere

src/Umbraco.Infrastructure/Security/UmbracoUserManager.cs
Shazwazza: Yep we need to have a plan to roll existing passwords forward. @bergmania that can be a separate task  if you can make one? For users we have `BackOfficePasswordHasher` to do that but for members it will need to be different since we don't store the algorithm. 

I'm unsure about this method though since in the existing codebase the `GetDefaultPasswordHasher` is not actually used anywhere. Maybe that is an oversight and it should be removed.  The underlying `PasswordHasher` property is what should be used here I think?

src/Umbraco.Infrastructure/Security/UmbracoUserManager.cs
Shazwazza: Seems good to me

Shazwazza (commented):
I ran out of time today so just submitting the notes that I've entered so far. Will continue with the rest tomorrow.

src/Umbraco.Infrastructure/Security/UmbracoUserManager.cs
bergmania: I have created a task in AB. https://dev.azure.com/umbraco/D-Team%20Tracker/_workitems/edit/10369

src/Umbraco.Infrastructure/Security/NoOpLookupNormalizer.cs
emmagarland: Cool, have done

src/Umbraco.Infrastructure/Security/UmbracoUserManager.cs
emmagarland: Good shout @Shazwazza, I have updated to use the property instead

src/Umbraco.Infrastructure/Security/IdentityMapDefinition.cs
```diff
@@ -76,9 +90,24 @@ private void Map(IUser source, BackOfficeIdentityUser target)
...
+            target.UserName = source.Username;
+            target.LastPasswordChangeDateUtc = source.LastPasswordChangeDate.ToUniversalTime();
+            target.LastLoginDateUtc = source.LastLoginDate.ToUniversalTime();
+            //target.EmailConfirmed = source.EmailConfirmedDate.HasValue;
+            target.Name = source.Name;
+            target.AccessFailedCount = source.FailedPasswordAttempts;
+            target.PasswordHash = GetPasswordHash(source.RawPasswordValue);
+            target.PasswordConfig = source.PasswordConfiguration;
+            target.IsApproved = source.IsApproved;
+            //target.SecurityStamp = source.SecurityStamp;
```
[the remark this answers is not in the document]
emmagarland: Yes, so there was no `SecurityStamp` on the `IMember`, and no `EmailConfirmed` date property either. We said we'd like to consider adding Security Stamp functionality. I don't think members had email confirmed or security stamp functionality before though?

However, I saw those properties are on the `IUser` which inherits from `IMembershipUser` (as does `IMember`).

I've moved the 2 properties to `IMembershipUser` as a start and I'm using and mapping them for `IMember` too now... although they have no store yet so there will need to be a task to update them still (it isn't on the `MemberDto`, or in the repository layer for example).

To implement these, we'll need separate tasks to add:

- **Security stamp** functionality to members allowing sessions to be ended as the result of a security event, e.g. when the user changes their password (including database storage)
- **Email confirmed** functionality to members (including database storage), which will be null if it isn't switched on

We also discussed lockout, configurable end date, MFA and external account linking so although not for this PR, something to make sure we have stories for?

src/Umbraco.Infrastructure/Security/MembersIdentityBuilder.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using System.Reflection;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Umbraco.Infrastructure.Security
+{
+    public class MembersIdentityBuilder : IdentityBuilder
```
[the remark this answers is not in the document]
emmagarland: No list yet, but I can create a GitHub task to track this on with checkboxes maybe?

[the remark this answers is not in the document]
emmagarland: Please see #9787 where we can track and add to this

src/Umbraco.Infrastructure/Security/MembersIdentityUser.cs
emmagarland: Ok, thanks, will have a look at this in the next PR for roles

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
```diff
@@ -147,138 +187,120 @@ public MemberListDisplay GetListNodeDisplay(string listName)
...
-
-            //Depending on the action we need to first do a create or update using the membership provider
+            // Depending on the action we need to first do a create or update using the membership manager
             // this ensures that passwords are formatted correctly and also performs the validation on the provider itself.
+
             switch (contentItem.Action)
             {
                 case ContentSaveAction.Save:
-                    UpdateMemberData(contentItem);
+                    Task<ActionResult<bool>> updateSuccessful = UpdateMemberAsync(contentItem);
```
Shazwazza, line 271: We don't use these responses do we? same with createSuccessful?

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
Shazwazza: I'll have to check this out with the debugger, not really sure off the top of my head.

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
Shazwazza: This is how user passwords are changed: https://github.com/umbraco/Umbraco-CMS/blob/netcore/netcore/src/Umbraco.Web.BackOffice/Controllers/UsersController.cs#L684

In theory we should do something very similar with members instead of just directly re-saving the hash. I suppose this is also why you've exposed `HashPassword` directly? Users does this just based on aspnet identity, creates a reset token and call `ChangePasswordWithResetAsync` with the token. The PasswordChanger class encompasses some of this logic, perhaps that can be shared too or parts re-used.

Shazwazza: Also the password changing in the UI for users has been changed so it's not part of the same saving request. Members should do this too but that requires UI changes. Though, that should be pretty much the same as what users do to. Maybe that is out of scope but should be done and would simplify somethings.

src/Umbraco.Web.BackOffice/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -37,6 +38,15 @@ public static void AddUmbracoBackOfficeIdentity(this IServiceCollection services
             services.ConfigureOptions<ConfigureBackOfficeSecurityStampValidatorOptions>();
         }
 
+        /// <summary>
+        /// Adds the services required for using Members Identity
+        /// </summary>
+        public static void AddMembersIdentity(this IServiceCollection services) =>
```
Shazwazza, line 44: This can't be in the BackOffice assembly because it will need to be used on the front-end. Since it will be used in both it needs to be part of "Common"

src/Umbraco.Web.BackOffice/Extensions/IdentityBuilderExtensions.cs
```diff
@@ -24,6 +25,19 @@ public static class IdentityBuilderExtensions
             return identityBuilder;
         }
 
+        /// <summary>
+        /// Adds a <see cref="UserManager{TUser}"/> for the <seealso cref="MembersIdentityUser"/>.
+        /// </summary>
+        /// <typeparam name="TInterface">The usermanager interface</typeparam>
+        /// <typeparam name="TUserManager">The usermanager type</typeparam>
+        /// <returns>The current <see cref="IdentityBuilder"/> instance.</returns>
+        public static IdentityBuilder AddMembersUserManager<TInterface, TUserManager>(this IdentityBuilder identityBuilder)
```
Shazwazza, line 34: Similar to above, these need to be part of Common since they will be used in the front-end too. The Front end assembly does not reference the back office. They can be shipped separately.

Shazwazza (changes requested):
Have added more review notes. Still not 100% complete but just submitting what I've got now.

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
bergmania: We will need to ensure `updateSuccessful.Result` is null, otherwise, it should be returned.

src/Umbraco.Web.Common/Security/MembersUserManager.cs
```diff
@@ -0,0 +1,67 @@
...
+            IServiceProvider services,
+            IHttpContextAccessor httpContextAccessor,
+            ILogger<UserManager<MembersIdentityUser>> logger,
+            IOptions<MemberPasswordConfigurationSettings> passwordConfiguration)
+            : base(ipResolver, store, optionsAccessor, passwordHasher, userValidators, passwordValidators, errors, services, logger, passwordConfiguration)
+        {
+            _httpContextAccessor = httpContextAccessor;
+        }
+
+        private string GetCurrentUserId(IPrincipal currentUser)
```
Shazwazza, line 41: I think we should remove this stuff along with all of the events and have a separate task to see if we want or need any of it. For now lets keep it as simple as possible. We didn't have events like this before for members.

Shazwazza (changes requested):
Added one more comment. That takes me to the end of all files :) The main one we'll need to review, test and tweak is the members controller. There's some existing notes there regarding the password changing. Once that is done and other comments are taken care of, I think we can look into merging and creating new cleanup tasks.

src/Umbraco.Infrastructure/Services/Implement/MemberService.cs
```diff
@@ -924,18 +931,28 @@ public void AddRole(string roleName)
...
+        /// <returns>A list of member roles</returns>
+
+        public IEnumerable<IMemberGroup> GetAllRoles()
         {
-            using (var scope = ScopeProvider.CreateScope(autoComplete: true))
+            using (IScope scope = ScopeProvider.CreateScope(autoComplete: true))
             {
                 scope.ReadLock(Constants.Locks.MemberTree);
-                return _memberGroupRepository.GetMany().Select(x => x.Name).Distinct();
+                return _memberGroupRepository.GetMany().Select(x=>x).Distinct();
```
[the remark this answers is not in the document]
emmagarland: True, updated!

src/Umbraco.Tests.Integration/Testing/UmbracoIntegrationTest.cs
```diff
@@ -229,6 +229,8 @@ public virtual void ConfigureServices(IServiceCollection services)
             }
 
             services.AddSignalR();
+            builder.AddMembersIdentity();
```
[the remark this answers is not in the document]
emmagarland: Good spot!

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
emmagarland: Ok thanks - will take a look

src/Umbraco.Web.BackOffice/DependencyInjection/ServiceCollectionExtensions.cs
emmagarland: Good point, will move

emmagarland: Moved to common

src/Umbraco.Tests.Integration/Umbraco.Web.BackOffice/MembersServiceCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,33 @@
+using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.DependencyInjection;
+using NUnit.Framework;
+using Umbraco.Core.DependencyInjection;
+using Umbraco.Infrastructure.Security;
+using Umbraco.Tests.Integration.Testing;
+using Umbraco.Web.BackOffice.DependencyInjection;
+
+namespace Umbraco.Tests.Integration.Umbraco.Web.BackOffice
```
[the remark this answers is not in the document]
emmagarland: Have moved

src/Umbraco.Web.BackOffice/Extensions/IdentityBuilderExtensions.cs
emmagarland: Good point - have moved

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
emmagarland: @bergmania what should I return if .`Result` is not null, when the controller expects `ActionResult<MemberDisplay>? The remaining controller code already checks for model state errors and displays them as needed

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
emmagarland: @bergmania please can you create a separate ticket for the Members password changing UI changes to match the users changes?

emmagarland: @Shazwazza since we already have the method, I will make some changes to reuse the same token functionality that is in `PasswordChanger` and inject it rather than newing it up (as per the TODO comment).

bergmania: I have added an intern task here: https://dev.azure.com/umbraco/D-Team%20Tracker/_workitems/edit/10496

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
bergmania: @emmagarland , The method has `ActionResult<MemberDisplay>` as return type, so you should be able to return the result

```suggestion
                    ActionResult<bool> updateSuccessful = await UpdateMemberAsync(contentItem);
                    if(!(updateSuccessful.Result  is null){
                        return updateSuccessful.Result;
                    }
                    // else we know the `updateSuccessful.Value` should be used.
```

emmagarland: Agree - have amended

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
emmagarland: @bergmania please can you create a separate task that I can do separately for removing the hashpassword and reusing the PasswordChanger class?

bergmania: Added another intern task here: https://dev.azure.com/umbraco/D-Team%20Tracker/_backlogs/backlog/CMS%20Team/Epics/?workitem=10551

Just let me know if you start on the task :)

emmagarland: Fab, thanks @bergmania

src/Umbraco.Web.Common/Security/MembersUserManager.cs
emmagarland: Have removed all events except the ones required, and left comment.

emmagarland:
Ready for review after addressing comments @bergmania @Shazwazza.

Shazwazza:
All looks good, lets get this merged and iterate from here. Here's my TODO notes:

* Review and cleanup MembersController
  * [ ] https://github.com/umbraco/Umbraco-CMS/pull/9472/files#r570428236
  * [ ] Are roles being saved? 
    * https://github.com/umbraco/Umbraco-CMS/pull/9472/files#diff-38a7bd6b22013f8a62a9ec38441d63dd12eda5b592a60d3b8c8a88b079b8f847L253
    * https://github.com/umbraco/Umbraco-CMS/pull/9472/files#diff-38a7bd6b22013f8a62a9ec38441d63dd12eda5b592a60d3b8c8a88b079b8f847L223
    * ... seems this moved here https://github.com/umbraco/Umbraco-CMS/pull/9472/files#diff-38a7bd6b22013f8a62a9ec38441d63dd12eda5b592a60d3b8c8a88b079b8f847R543... just need to test
  * [ ] Deal with TODOs https://github.com/umbraco/Umbraco-CMS/pull/9472/files#diff-38a7bd6b22013f8a62a9ec38441d63dd12eda5b592a60d3b8c8a88b079b8f847R387
  * [ ] Password changing 
    * https://dev.azure.com/umbraco/D-Team%20Tracker/_workitems/edit/10496
    * https://dev.azure.com/umbraco/D-Team%20Tracker/_workitems/edit/10551
* [ ] GetDefaultPasswordHasher could be removed?
* [ ] Rolling passwords forward

* Review Role store
  * [ ] Shouldn't this throw since we don't support it? https://github.com/umbraco/Umbraco-CMS/pull/9472/files#diff-7d11ac1f511fddbcd0452f513abb33e39c12e7036153af6e11ab4a93f431cef5R31
  * [ ] Does this role store do anything?

* [ ] Deal with adding members services when only backoffice or only front-end is enabled or both - and ensuring that duplicate services are not added

* [ ] Code cleanup and sharing, there will be a lot https://github.com/umbraco/Umbraco-CMS/issues/9787

* MemberManager/BackOfficeUserManager:
  * [ ] remove events https://github.com/umbraco/Umbraco-CMS/pull/9472/files#diff-755a86395625394129f741f54625a820ff58f4e0d7216a5815934cd126a51a5eR54 and remove events from the interface
  * [ ] for user manager change to event aggregation

* [ ] Remove MembershipHelper https://github.com/umbraco/Umbraco-CMS/pull/9472/files#diff-c9a9c2ec5d674dd51987fffaa2009fcfec5d0e44f2e1c3192ff21e0809e6f3d4R22

* [ ] remove MembersMembershipProvider - https://github.com/umbraco/Umbraco-CMS/pull/9472#discussion_r570429233 looks as though this is still tied to macros

* MembersIdentityUser
  * [ ] Need to revisit INT vs STRNIG in IdentityUserRole: https://github.com/umbraco/Umbraco-CMS/pull/9472#discussion_r573602602

* Security Stamp/Email Confirmed
  * [ ] We need to persist and handle the Security Stamp and Email Confirmed values since these are not persisted
  * [ ] This will require additional DB columns, migrations, service and repo changes
  * [ ] We then need to use the security stamp correctly which requires using the security stamp validator in the members cookies implementation
  * [ ] We then need to use the email confirmed field - how we use this is yet to be determined

* External logins
  * [ ] This needs to be implemented

* Lockout policies + MFA
  * [ ] This would be a new feature from v8. We support account lockout, but not account lockout duration (i.e. requires to be manually unlocked in the back office)
  * [ ] MFA would be a much bigger task and would require shipping a bunch of new macros - could be post first release

src/Umbraco.Web/Security/MembershipHelper.cs
```diff
@@ -19,6 +19,15 @@
 namespace Umbraco.Web.Security
 {
     // MIGRATED TO NETCORE
+    // TODO: Analyse all - much can be moved/removed since most methods will occur on the manager via identity implementation
```
emmagarland, line 22: `MembershipHelper.cs` to be removed once all methods have been analysed, moved/removed since most methods will occur on the manager via identity implementation

src/Umbraco.Infrastructure/Security/MemberRolesUserStore.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// A custom user store that uses Umbraco member data
+    /// </summary>
+    public class MemberRolesUserStore : RoleStoreBase<IdentityRole<string>, string, IdentityUserRole<string>, IdentityRoleClaim<string>>
+    {
+        private readonly IMemberService _memberService;
+        private readonly IMemberGroupService _memberGroupService;
+        private readonly IScopeProvider _scopeProvider;
+
+        public MemberRolesUserStore(IMemberService memberService, IMemberGroupService memberGroupService, IScopeProvider scopeProvider, IdentityErrorDescriber describer)
+            : base(describer)
```
emmagarland, line 23: Implement role functionality for members in .NET Core

src/Umbraco.Infrastructure/Security/MembersIdentityUser.cs
emmagarland: FYI, I have stuck to string for roles in my PR.

PR #9477, opened 2020-11-30
Add input id so click on property label set focus in input

bjarnef:
### Prerequisites

### Description
This PR specify the id attribute on dictionary textarea for each language. The property alias is used here, which is set to language iso code.

Currently the `<umb-property>` component is using alias property value from the property object:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/html/umb-control-group.html#L4

However in some cases (e.g. in splitview or when opening overlays where you have same property alias in for example block list editor as on the document type itself, the browser (Chrome) warn about multiple ids with same value.

Maybe we should allow to specify an `input-id` attribute on the component or `inputId` property on the property object. This could use `String.CreateGuid()` to ensure a unique id like in `<umb-checkbox>` component.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js#L52


bjarnef:
@nielslyngsoe maybe something to consider regarding multiple ids with same value. It is easy to reproduce if having a property with alias "image" and a block list editor with alias "sections". If some of the used blocks have a property "image" clicking this property label in the overlay it will focus on the "image" property on the document type itself.

nathanwoulfe (approved):
All good, thanks @bjarnef

nathanwoulfe:
@bjarnef I've resolved a merge conflict, would you mind confirming this still works as expected?

bjarnef:
@nathanwoulfe it should still work as the property label use alias as value in `for` attribute, so it will focus on the element with this id.

nathanwoulfe:
@bjarnef bit slow on my end for this one, completely forgot it was still pending...

nielslyngsoe:
@bjarnef great, thanks. And big thanks to @nathanwoulfe for reviewing.
I´m happy that you two make these things happen. I´m sorry I don't have the time to read and reply on everything here on GitHub, so please feel free to move on with such initiatives. I think they are all great!

PR #9482, opened 2020-12-01
Add support for unattended installs

rasmusjp:
To enable unattended install the `Umbraco.Core.ConfigurationStatus` app setting needs to have the current version and the `umbracoDbDSN` connection string needs to be configured. Then on boot we'll check if the database is empty (has no tables) and if it is, the database schema and default data will be created.

The unattended install can be disabled by adding an app setting `Umbraco.Core.RuntimeState.InstallUnattended` with the value `false`. Umbraco will then fail to boot if the version and connection string is present but the database is empty.

Note: You will not be able to login to Umbraco since no password has been set for the default user, you either need to have an external login provider configured or setting a password in some other way.

---
_This item has been added to our backlog [AB#9632](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/9632)_

### DIFF
--- a/src/Umbraco.Core/Runtime/CoreRuntime.cs
+++ b/src/Umbraco.Core/Runtime/CoreRuntime.cs
@@ -174,6 +174,7 @@ namespace Umbraco.Core.Runtime
                 // determine our runtime level
                 DetermineRuntimeLevel(databaseFactory, ProfilingLogger);
 
+
                 // get composers, and compose
                 var composerTypes = ResolveComposerTypes(typeLoader);
 
@@ -260,16 +261,17 @@ namespace Umbraco.Core.Runtime
 
             using (var database = databaseFactory.CreateDatabase())
             {
-                var isDatabaseEmpty = databaseFactory.SqlContext.SqlSyntax.GetTablesInSchema(database).Any() == false;
+                var hasUmbracoTables = database.IsUmbracoInstalled(Logger);
 
-                // database is not empty, assume Umbraco is already installed
-                if (isDatabaseEmpty == false) return;
+                // database has umbraco tables, assume Umbraco is already installed
+                if (hasUmbracoTables) return;
 
                 // all conditions fulfilled, do the install
                 Logger.Info<CoreRuntime>("Starting unattended install.");
-                database.BeginTransaction();
+                
                 try
                 {
+                    database.BeginTransaction();
                     var creator = new DatabaseSchemaCreator(database, Logger);
                     creator.InitializeDatabaseSchema();
                     database.CompleteTransaction();
--- a/src/Umbraco.Core/Runtime/CoreRuntime.cs
+++ b/src/Umbraco.Core/Runtime/CoreRuntime.cs
@@ -336,7 +336,7 @@ namespace Umbraco.Core.Runtime
             {
                 try
                 {
-                    _state.DetermineRuntimeLevel(databaseFactory, profilingLogger);
+                    _state.DetermineRuntimeLevel(databaseFactory);
 
                     profilingLogger.Debug<CoreRuntime>("Runtime level: {RuntimeLevel} - {RuntimeLevelReason}", _state.Level, _state.Reason);
 
--- a/src/Umbraco.Core/Runtime/CoreRuntime.cs
+++ b/src/Umbraco.Core/Runtime/CoreRuntime.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
+using System.Threading;
 using System.Web;
 using System.Web.Hosting;
 using Umbraco.Core.Cache;
@@ -10,6 +12,7 @@ using Umbraco.Core.Exceptions;
 using Umbraco.Core.IO;
 using Umbraco.Core.Logging;
 using Umbraco.Core.Logging.Serilog;
+using Umbraco.Core.Migrations.Install;
 using Umbraco.Core.Persistence;
 using Umbraco.Core.Persistence.Mappers;
 using Umbraco.Core.Sync;
@@ -149,7 +152,7 @@ namespace Umbraco.Core.Runtime
                 {
                     MainDom = new MainDom(Logger, new MainDomSemaphoreLock(Logger));
                 }
-                
+
 
                 // create the composition
                 composition = new Composition(register, typeLoader, ProfilingLogger, _state, configs);
@@ -157,6 +160,9 @@ namespace Umbraco.Core.Runtime
 
                 // run handlers
                 RuntimeOptions.DoRuntimeEssentials(composition, appCaches, typeLoader, databaseFactory);
+
+                // determines if unattended install is enabled and performs it if required
+                DoUnattendedInstall(databaseFactory);
 
                 // register runtime-level services
                 // there should be none, really - this is here "just in case"
@@ -175,7 +181,7 @@ namespace Umbraco.Core.Runtime
                 using (ProfilingLogger.DebugDuration<CoreRuntime>("Scanning enable/disable composer attributes"))
                 {
                     enableDisableAttributes = typeLoader.GetAssemblyAttributes(typeof(EnableComposerAttribute), typeof(DisableComposerAttribute));
-                }   
+                }
 
                 var composers = new Composers(composition, composerTypes, enableDisableAttributes, ProfilingLogger);
                 composers.Compose();
@@ -223,6 +229,62 @@ namespace Umbraco.Core.Runtime
             }
 
             return _factory;
+        }
+
+        private void DoUnattendedInstall(IUmbracoDatabaseFactory databaseFactory)
+        {
+            // unattended install is not enabled
+            if (RuntimeOptions.InstallUnattended == false) return;
+
+            var localVersion = UmbracoVersion.LocalVersion; // the local, files, version
+            var codeVersion = _state.SemanticVersion; // the executing code version
+
+            // local version and code version is not equal, an unattended install cannot be performed
+            if (localVersion != codeVersion) return;
+
+            // no connection string set
+            if (databaseFactory.Configured == false) return;
+
+            var tries = 5;
+            var connect = false;
+            for (var i = 0;;)
+            {
+                connect = databaseFactory.CanConnect;
+                if (connect || ++i == tries) break;
+                Logger.Debug<CoreRuntime>("Could not immediately connect to database, trying again.");
+                Thread.Sleep(1000);
+            }
+
+            // could not connect to the database
+            if (connect == false) return;
+
+            using (var database = databaseFactory.CreateDatabase())
+            {
+                var isDatabaseEmpty = databaseFactory.SqlContext.SqlSyntax.GetTablesInSchema(database).Any() == false;
+
+                // database is not empty, assume Umbraco is already installed
+                if (isDatabaseEmpty == false) return;
+
+                // all conditions fulfilled, do the install
+                Logger.Info<CoreRuntime>("Starting unattended install.");
+                database.BeginTransaction();
+                try
+                {
+                    var creator = new DatabaseSchemaCreator(database, Logger);
+                    creator.InitializeDatabaseSchema();
+                    database.CompleteTransaction();
+                    Logger.Info<CoreRuntime>("Unattended install completed.");
+                }
+                catch (Exception ex)
+                {
+                    Logger.Error<CoreRuntime>(ex, "Error during unattended install.");
+                    database.AbortTransaction();
+
+                    throw new UnattendedInstallException(
+                        "The database configuration failed with the following message: " + ex.Message
+                        + "\n Please check log file for additional information (can be found in '/App_Data/Logs/')");
+                }
+            }
         }
 
         protected virtual void ConfigureUnhandledException()
--- a/src/Umbraco.Core/Runtime/CoreRuntime.cs
+++ b/src/Umbraco.Core/Runtime/CoreRuntime.cs
@@ -174,7 +174,6 @@ namespace Umbraco.Core.Runtime
                 // determine our runtime level
                 DetermineRuntimeLevel(databaseFactory, ProfilingLogger);
 
-
                 // get composers, and compose
                 var composerTypes = ResolveComposerTypes(typeLoader);
 
@@ -329,8 +328,8 @@ namespace Umbraco.Core.Runtime
             }
         }
 
-        // internal for tests
-        internal void DetermineRuntimeLevel(IUmbracoDatabaseFactory databaseFactory, IProfilingLogger profilingLogger)
+        // internal/virtual for tests (i.e. hack, do not port to netcore)
+        internal virtual void DetermineRuntimeLevel(IUmbracoDatabaseFactory databaseFactory, IProfilingLogger profilingLogger)
         {
             using (var timer = profilingLogger.DebugDuration<CoreRuntime>("Determining runtime level.", "Determined."))
             {
--- a/src/Umbraco.Core/RuntimeOptions.cs
+++ b/src/Umbraco.Core/RuntimeOptions.cs
@@ -47,14 +47,12 @@ namespace Umbraco.Core
         /// </summary>
         /// <remarks>
         /// <para>By default, when a database connection string is configured and it is possible to connect to
-        /// the database, but the database is empty, the runtime enters the BootFailed level. If this options
-        /// is set to true, it enters the Install level instead.</para>
-        /// <para>It is then up to the implementor, that is setting this value, to take over the installation
-        /// sequence.</para>
+        /// the database, but the database is empty, the runtime enters the <c>Install</c> level. If this options
+        /// is set to <c>false</c>, it enters the <c>BootFailed</c> level instead.</para>
         /// </remarks>
         public static bool InstallEmptyDatabase
         {
-            get => _installEmptyDatabase ?? BoolSetting("Umbraco.Core.RuntimeState.InstallEmptyDatabase", false);
+            get => _installEmptyDatabase ?? BoolSetting("Umbraco.Core.RuntimeState.InstallEmptyDatabase", true);
             set => _installEmptyDatabase  = value;
         }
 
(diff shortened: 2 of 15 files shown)


Shazwazza:
@rasmusjp haven't had time to look at the code but the tests are failing...

rasmusjp:
Not sure why the that test is failing, it doesn't when running them locally.

src/Umbraco.Web/Compose/UnattendedInstallComponent.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            _logger.Info<UnattendedInstallComponent>("Installing Umbraco.");
+            var result = _databaseBuilder.CreateSchemaAndData();
+            _logger.Info<UnattendedInstallComponent>("Umbraco Installed.");
+
+            if (result.Success == false)
+                throw new InstallException("An error occurred while running the unattended installation.\n" + result.Message);
+
+            // the install was successful, lets restart the application to get Umbraco to boot the correct state
+            UmbracoApplication.Restart();
```
rasmusjp, line 41: Not sure if this is the best way/place to do this, as I'm not sure what happens if we want to run another component after this one, will it not run because the application is restarting?

rasmusjp: Updated to set a marker and do the restart in the `UmbracoInjectedModule` instead. This allows for running additional components after this one.

src/Umbraco.Core/Persistence/IUmbracoDatabaseFactory.cs
```diff
@@ -42,6 +42,12 @@ public interface IUmbracoDatabaseFactory : IDisposable
         /// </summary>
         bool CanConnect { get; }
 
+        /// <summary>
+        /// Gets a value indicating whether the database is empty, it should already be determined if its possible to
+        /// connect to the database (see <see cref="CanConnect"/>) otherwise an exception might be thrown.
+        /// </summary>
+        bool IsDatabaseEmpty { get; }
```
Shazwazza, line 49: This is a breaking change. This could just be an ext method right? Since all it's doing is:

```cs
 using (var database = dbFactory.CreateDatabase())
 {
      return SqlContext.SqlSyntax.GetTablesInSchema(database).Any() == false;
 }
```

src/Umbraco.Core/RuntimeOptions.cs
```diff
@@ -46,14 +46,12 @@ public static bool InstallMissingDatabase
...
-        /// is set to true, it enters the Install level instead.</para>
-        /// <para>It is then up to the implementor, that is setting this value, to take over the installation
-        /// sequence.</para>
+        /// the database, but the database is empty, the runtime enters the Install level. If this options
+        /// is set to <c>false</c>, it enters the BootFailed level instead.</para>
         /// </remarks>
         public static bool InstallEmptyDatabase
         {
-            get => _installMissingDatabase ?? BoolSetting("Umbraco.Core.RuntimeState.InstallEmptyDatabase", false);
+            get => _installMissingDatabase ?? BoolSetting("Umbraco.Core.RuntimeState.InstallEmptyDatabase", true);
```
Shazwazza, line 54: I guess this is a 'breaking' change but is just something we need to document to say that this is true by default now. We can add the breaking change tag to this PR

src/Umbraco.Web/Compose/UnattendedInstallComponent.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using Umbraco.Core;
+using Umbraco.Core.Composing;
+using Umbraco.Core.Logging;
+using Umbraco.Core.Migrations.Install;
+using Umbraco.Web.Install;
+
+namespace Umbraco.Web.Compose
```
Shazwazza, line 8: Can we have this in Core and not dependent on any web things? We can avoid the IHttpContextAccessor dependency in some other way? Maybe an event?

Shazwazza (changes requested):
Left a few comments in code. But have some other questions:

What happens if there is an error during install? Or is that work scheduled for a different task? If an error occurs during install, we need to capture that error and have the installer render as normal, but show the last error that occurred

src/Umbraco.Core/Persistence/IUmbracoDatabaseFactory.cs
rasmusjp: Yeah it could, was just thinking it were nice to have it in a shared place.

src/Umbraco.Web/Compose/UnattendedInstallComponent.cs
rasmusjp: Yeah, I didn't like putting it in Web. I'll see if it's posible to handle it with events instead.

src/Umbraco.Core/RuntimeOptions.cs
rasmusjp: I guess it is, however you'll still need to have the version number in the config, configured the connection string and an empty database for the unattended install to kick in.

rasmusjp:
If an error occours you get an YSOD with the error message. I'm not sure if redirecting to the installer would give any benefits besides showing the error in a nicer layout. But if we do want to show the error in the installer instead we need to ensure that the status code is 500, so if it was invoked by an external service it knows the install failed.

src/Umbraco.Core/Persistence/IUmbracoDatabaseFactory.cs
rasmusjp: Moved to `RuntimeState.cs`, reason for adding the check is that `EnsureUmbracoUpgradeState` could throw of other reasons than the database is empty

src/Umbraco.Web/Compose/UnattendedInstallComponent.cs
rasmusjp: Moved to Core project and added an event that `WebRuntime` subscribes to, it adds `AppPoolRequiresRestart=true` if the installation is successful which is picked up in the `UmbracoInjectedModule` that restarts the application if the runtime is in an install or upgrade state and `AppPoolRequiresRestart=true`.

src/Umbraco.Core/RuntimeState.cs
```diff
@@ -212,10 +213,17 @@ public void DetermineRuntimeLevel(IUmbracoDatabaseFactory databaseFactory, ILogg
 
                 if (RuntimeOptions.InstallEmptyDatabase)
                 {
-                    // ok to install on an empty database
-                    Level = RuntimeLevel.Install;
-                    Reason = RuntimeLevelReason.InstallEmptyDatabase;
-                    return;
+                    var isDatabaseEmpty = false;
+                    using (var database = databaseFactory.CreateDatabase())
```
Shazwazza, line 217: should use braces in this using, our netcore linting will yet at us otherwise

src/Umbraco.Core/Compose/UnattendedInstallComponent.cs
```diff
@@ -0,0 +1,45 @@
...
+        private readonly DatabaseBuilder _databaseBuilder;
+
+        public UnattendedInstallComponent(IRuntimeState runtimeState, ILogger logger, DatabaseBuilder databaseBuilder)
+        {
+            _runtimeState = runtimeState ?? throw new ArgumentNullException(nameof(runtimeState));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _databaseBuilder = databaseBuilder ?? throw new ArgumentNullException(nameof(databaseBuilder));
+        }
+
+        public static event EventHandler<UnattendedInstallEventArgs> InstallCompleted;
```
Shazwazza, line 23: What is the reason for requiring a restart? Do we need it? If not this simplifies a lot of things and we don't need this event. We'll also have to translate all of this to netcore where we are going to remove all static events and where restarts don't really exist (only shutdowns which require manually restarting the web app).

src/Umbraco.Core/Compose/UnattendedInstallComponent.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            _logger.Info<UnattendedInstallComponent>("Installing Umbraco.");
+            var result = _databaseBuilder.CreateSchemaAndData();
+            _logger.Info<UnattendedInstallComponent>("Umbraco Installed.");
+
+            if(InstallCompleted != null)
+                InstallCompleted.Invoke(this, new UnattendedInstallEventArgs(result.Success, result.Message));
+
+            if (result.Success == false)
+                throw new UnattendedInstallException(result.Message);
```
Shazwazza, line 38: If an exception is thrown in a component this generally means that exception is caught and stored which means you have to restart the app again to retry. I think I'm ok with that but wanted to see what you think?

src/Umbraco.Core/Compose/UnattendedInstallComposer.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Umbraco.Core.Composing;
+
+namespace Umbraco.Core.Compose
+{
+    [RuntimeLevel(MinLevel = RuntimeLevel.Install, MaxLevel = RuntimeLevel.Install)]
```
Shazwazza, line 5: This is using `ComponentComposer` which is an `IComposer` and not an `ICoreComposer` which means it will run after all `ICoreComposer`s . I assume that most composers don't execute on install but perhaps it's safer to also make this implement ICoreComposer? 
This is all changing in netcore too because it's very confusing to have to deal with this kinda thing.

Shazwazza (changes requested):
Regarding redirecting to the installer - yep i'm ok with just a YSOD for now since as you've mentioned, this is still an opt-in scenario.

I've left more questions, i guess my mind is skipping ahead to having to reimplement this in netcore and trying to see if we can reduce the changes.

src/Umbraco.Core/Compose/UnattendedInstallComponent.cs
rasmusjp: Since it's in an component it's running after the runtime level is determined, so we need to restart after the install is done to get into the `Run` level. We could do the install ealier in the boot phase and never enter the install level if unattended installs is configured.

rasmusjp:
Your comment about restarting made me think if we could run the install earlier which I realised we could, so I've redone the PR.

Instead of running in an Component that runs after the runtime level is determined the unattended install is now performed in `CoreRuntime` before the runtime is determined, this way we don't need to restart the application after the install has run. It'll still throw an exception if an error occours and require you to restart the application manually, which I think is fine.

I've added a new app setting `Umbraco.Core.RuntimeState.InstallUnattended` which needs to be `true` (default) for the unattended install to run. This way we don't need to change how `Umbraco.Core.RuntimeState.InstallEmptyDatabase` works which could break if someone is already using it.

src/Umbraco.Core/Runtime/CoreRuntime.cs
```diff
@@ -225,6 +231,62 @@ protected virtual IFactory Boot(IRegister register, DisposableTimer timer)
...
+            using (var database = databaseFactory.CreateDatabase())
+            {
+                var isDatabaseEmpty = databaseFactory.SqlContext.SqlSyntax.GetTablesInSchema(database).Any() == false;
+
+                // database is not empty, assume Umbraco is already installed
+                if (isDatabaseEmpty == false) return;
+
+                // all conditions fulfilled, do the install
+                Logger.Info<CoreRuntime>("Starting unattended install.");
+                database.BeginTransaction();
```
Shazwazza, line 270: This needs to be within the try/catch, creating a trans will throw if for some reason the db connection is severed. I've fixed this up.

Shazwazza (commented):
I've made a bunch of changes since during testing I realized there is problems with SqlMainDomLock and other initialization things that would be required when running on azure. I'll push the changes and comment about them on the main thread. I've left a few more comments though.

Shazwazza:
@rasmusjp The changes I've pushed are in this commit in this branch if you want to review: https://github.com/umbraco/Umbraco-CMS/pull/9482/commits/bbe2941d255318bffdd63887c22e32c8c8e48ad2

* Changes how we check if Umbraco is actually installed by using the existing DatabaseSchemaResult - this fixes a few things, like if the database exists and already has tables but Umbraco is not installed in it, and also fixes the old rudimentary check if there was a non-default user installed (which was also causing exceptions because this wasn't actually checking if tables existed)
* Fixes to SqlMainDomLock as this was not working at all because the db tables did not exist so it would just ysod with unattended installs happening. You can test this by adding this appsetting: `<add key="Umbraco.Core.MainDom.Lock" value="SqlMainDomLock"/>` which will be a requirement on azure app service

Let me know what you think.

With those changes i can confirm it all works as expected. 

I guess another question I have is whether or  not we should care if the umbraco version is configured or not? I mean if there's a connection string and the db can be connected but we know Umbraco isn't installed, should we just install it? It would mean bumping the web.config to the new version like the normal install process. Perhaps this is a 'breaking change' though since it's changing the expected outcome than what is currently there.

Another thing that we'll need to consider is when we migrate this to netcore which will probably require some manual doing. I'm sure @bergmania will be thrilled! We don't have a configured version in netcore - so I guess what I mentioned above would just be what we do: if there's a conn string, the db connects but there is no umb tables, then we do unattended install.

Shazwazza:
@rasmusjp Ah I forgot I had one other question/comment

> Note: You will not be able to login to Umbraco since no password has been set for the default user, you either need to have an external login provider configured or setting a password in some other way.

I realize this will be ok for Cloud installs since Umbraco Id will be enabled but this will still be quite unfriendly for a normal Azure user. Having unattended installs for Azure will be a wonderful thing since it's such a pain in the arse now to have to install and juggle removing connection strings, umbraco version, etc... to get the installer to run. But now if the whole thing just installs but they can't login, that also sucks.

One idea could be:

* Once unattended install is complete we continue to leave the RuntimeLevel as Install
* In the installer we dynamically choose which installation options are required to run - this should already be done IMO. For example, we don't have the database screen show up because we know it's configured and can be connected too
* The installer then just shows the welcome screen, the user screen and the starter kit screen and installation proceeds as normal
* And in this case, to me it seems like `InstallEmptyDatabase` should default to true in case a restart happens between the unattended install and the installer (for some weird reason)

For Cloud, possibly there's another config option to bypass this default operation and once unattended install is done the runtime level proceeds normally past Install.

Maybe there's other options, what do you think?

rasmusjp:
@Shazwazza Changes looks good to me and the unattended install still works as expected.

Not sure why the SqlMainDomLock would fail though, since it's acquired after the unattended install has finished, so the tables should definitely be there.

Since we're already adding an app setting to enable/disable unattended installs `Umbraco.Core.RuntimeState.InstallUnattended`, I'm not sure if adding another one for Cloud is desired. 

If we are going to show the installer I don't think we'll need to run the unattended install first. So I'm thinking we should default `InstallUnattended` to `false` and `InstallEmptyDatabase` to `true` (as you mentioned), then we can skip the connection string step if already configured and run the installer normally.

This way we won't "accidentally" enter the install level if unattended installs are enabled, and potentially allow creating a db user.

Shazwazza:
> Not sure why the SqlMainDomLock would fail though, since it's acquired after the unattended install has finished, so the tables should definitely be there.

So if you have a connection string but not version - kaboom since unattended install doesn't run

> If we are going to show the installer I don't think we'll need to run the unattended install first. So I'm thinking we should default InstallUnattended to false and InstallEmptyDatabase to true (as you mentioned), then we can skip the connection string step if already configured and run the installer normally.

If that ends up allowing a connection string + a configured version (or not, I think that is optional) to exist but still run the installer than that is good. Except we should still detect in the database step whether to show that screen or not if there's a connection string that already exists and the db can be connected to.

rasmusjp:
> So if you have a connection string but not version - kaboom since unattended install doesn't run

Ahh.. yeah that makes sense, hadn't tested without a configured version.

> If that ends up allowing a connection string + a configured version (or not, I think that is optional) to exist but still run the installer than that is good. Except we should still detect in the database step whether to show that screen or not if there's a connection string that already exists and the db can be connected to.

Yep, that's what I was thinking. I'll give it a look

rasmusjp:
I've switched the confiigs around so `InstallUnattended` is `false` by default and `InstallEmptyDatabase` is `true`, this allows the installer to run if a connection string is configured but Umbraco is not installed. 

I was having problems running the installer (even with no version and connection string configured), it turned out that `HasSomeNonDefaultUser` is needed in the `NewInstallStep`, if we only check if Umbraco is installed, the step would be shown but it would skip creating the user. This happens because `RequiresExecution` is used to both determine if the step should be shown and if it should execute, so the second time it's called, when it checks if `ExecuteAsync` should be called, Umbraco is already installed and it would return `false` and never call `ExecuteAsync` resulting in the configured user not being created.

To do a new install with a preconfigured connection string, the version needs to be empty. I tried to see if I could get the installer run a new install with a version configured, but there are multiple checks that decides if an install or upgrade should run if the version is there and I suspect changing them would break upgrades from v7.

Shazwazza:
> To do a new install with a preconfigured connection string, the version needs to be empty. I tried to see if I could get the installer run a new install with a version configured, but there are multiple checks that decides if an install or upgrade should run if the version is there and I suspect changing them would break upgrades from v7.

Found a few issues here. The check for `RuntimeOptions.InstallEmptyDatabase` only occurs when the version is actually configured which happens in the try/catch block within `DetermineRuntimeLevel`. Regardless of this setting and if the version is empty then installer will run with the configured conn string. So the change in `IsBrandNewInstall` and other places is nice and allows the installer to run with a configured database. But if a version is configured, this setting is used but then we end up in a loop of this screen:

image

which tries to run the installer again and again and again without any feedback to the user that it's broken so that is not good.

I have fixed this all now and it works as expected:

* InstallEmptyDatabase is now obsolete and not used at all - there's never a reason we wouldn't want to install on an empty database during install. This also simplifies things of whether or not the version is configured in the web.config == it shouldn't matter.
* Fixes NewInstallStep to do different checks on what view to return vs what checks to execute for install and cleans up how these checks are done
* Removes umb version check from DatabaseBuilder, there is no reason for this to exist there 
* Cleans up the RuntimeState DetermineRuntimeLevel checks

Then there was another problem with SqlMainDom throwing an exception on restart when trying to continue the install when there is no DB installed yet because it's trying to acquire a write lock on dispose but there is no table. Have fixed.

See rev for changes: 083b5d26118b4bbbf9788e9e6b8d2632ec68445d

I've tested these combinations but you'll want to do that too in case I've missed anything:

|   	|  Configured Version 	|  With Conn String |  InstallUnattended 	|  
|---	|---	|---	|---	|
|  Default	|  false 	|  false 	|  false 	|
|  Unattended (no installer, will just install without a user password)	|  true 	|  true 	|  true 	|
|  Pre configured (will run installer without DB option)	|  false 	|  true 	|  false 	|
|  Pre configured (will run installer without DB option) 	|  true 	|  true 	|  false 	|

But we also need to test:

* Upgrades from a previous v8 version
* Upgrades from a v7 version
* This version + Umbraco Deploy on a cloned Cloud site

@rasmusjp  I'm hoping you may have time to help test these?

Shazwazza:
Ah, i can see the build is failing so will have a look tomorrow but those tests can still be done if there's time

rasmusjp:
@Shazwazza I've tested the different install configurations listed above and they all work as expected, upgrading from 7.15.6 and 8.0.0 also works as expected and a cloned Cloud site with this version also runs with no problem.

There is a failing test though, from what I can tell the `IUmbracoDatabaseFactory.CanConnect` returns `true` but then `CreateDatabase` returns `null` https://github.com/umbraco/Umbraco-CMS/blob/9e3d1836f09c0171d38b93e7b9aedac97362a80b/src/Umbraco.Core/RuntimeState.cs#L236-L255

Shazwazza:
It's because this test doesn't have all things mocked and since it is now trying to check the DB on startup it throws. I'll fix up.

Shazwazza:
@rasmusjp Have pushed a fix for the test, If the build succeeds I'll get it merged 🚀

PR #9486, opened 2020-12-02
Converted  OutgoingEditorModelEventAttribute to a type filter

Matthew-Wise:
Converted OutgoingEditorModelEventAttribute to a type filter and updated the code as needed (hopefully)

Issue: #8264

Matt

### DIFF
--- a/src/Umbraco.Web.BackOffice/Filters/OutgoingEditorModelEventAttribute.cs
+++ b/src/Umbraco.Web.BackOffice/Filters/OutgoingEditorModelEventAttribute.cs
@@ -17,7 +17,7 @@ namespace Umbraco.Web.BackOffice.Filters
         }
 
 
-        private class OutgoingEditorModelEventFilter : IResultFilter
+        private class OutgoingEditorModelEventFilter : IActionFilter
         {
 
             private readonly IUmbracoContextAccessor _umbracoContextAccessor;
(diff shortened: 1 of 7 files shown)


src/Umbraco.Web.BackOffice/Filters/OutgoingEditorModelEventAttribute.cs
```diff
@@ -4,47 +4,60 @@
...
-        public OutgoingEditorModelEventAttribute(IUmbracoContextAccessor umbracoContextAccessor, IBackOfficeSecurityAccessor backofficeSecurityAccessor)
+        public OutgoingEditorModelEventAttribute() : base(typeof(OutgoingEditorModelEventFilter))
         {
-            _umbracoContextAccessor = umbracoContextAccessor ?? throw new ArgumentNullException(nameof(umbracoContextAccessor));
-            _backofficeSecurityAccessor = backofficeSecurityAccessor ?? throw new ArgumentNullException(nameof(backofficeSecurityAccessor));
         }
 
-        public override void OnActionExecuted(ActionExecutedContext context)
+
+        private class OutgoingEditorModelEventFilter : IResultFilter
```
bergmania, line 20: If you use `IActionFilter` here, you can still override `OnActionExecuted` and then all the changes to events can be reverted

bergmania (changes requested):
Hi @Matthew-Wise, thanks for the PR..

Almost how I would have solved it 💪,
But I think we should use `IActionFilter` instead of `IResultFilter`

bergmania:
Nice work, thanks a lot @Matthew-Wise  🚀

Issue #9389, opened 2020-11-12
Make ContentService.TreeChanged event public


Malthebp:
Hi Umbraco. 

I would like to suggest making the TreeChanged event in the ContentService public, so that it is possible to extend Umbraco even more than today. 

Before even considering making a PR for it, I would like to know and hopefully have an open discussion about why I could see potential in making it public but likewise I would really like knowing, why it is internal at the moment. :-)

**My initial thought and how the hell I found myself in a situation needing it**
So I was working on a new product where I wish to do the following: 
* I would like to know what variants are being published/unpublished, this I could do in the _Publishing_ event already, no problem.
* I would like to Enqueue a new action when the scope completes, aka. when Umbraco is done doing everything related to publishing. 
* I would then like to fetch the updated IPublishedContent from the cache and use the culture from the Publishing event to get properties, URL, etc. Really no problem here either, as I could pass information from the publishing event to the action. 
* I would also though, like to know whether I need to do something for the descendants BUUUT, here is the big BUUUT, I have no clue at all, because the _Publishing_ event knows something about the cultures and what has changed and the CacheRefresher knows something about what to include (RefreshNode, RefreshBranch, etc.). And for what I understand what triggers the cache to refresh is the internal TreeChanged event and that event has all I wish to know. 

So that is my use case.

Note: If you can see a solution without making it public I am very interested in that as well. 

I hope you wish to talk about it as much as I. :-)

nul800sebastiaan:
I'll ask around, I can only find that it was introduced in v8 so typically we'd keep things internal until they've stabilized. I'll see if we're happy to have it public.

Malthebp:
Thank you @nul800sebastiaan :-) 

I will wait here, excited for the answer.

Malthebp:
Hey @nul800sebastiaan 

Have you been able to find someone who are able to comment on this request?

Shazwazza:
It's on my list to review

Shazwazza:
Yes it is fine, TreeChanged exists in both MediaService and ContentService and can be made public

Malthebp:
Thank you @Shazwazza 

I will make a PR soon.

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/9487

PR #9487, opened 2020-12-02
Change access level for TreeChanged event in Media and ContentService

Malthebp:
As agreed with @Shazwazza in issue #9389.


nul800sebastiaan:
Hi @Malthebp! Sorry I missed this PR until now, this is nice, good to go! 👍

PR #9503, opened 2020-12-05
NetCore: updated some code to match linting rules 

AndyButland:
I've restricted amends here to just code we've added in the .NET Core transition to avoid issues with merges from the V8 branch.  Have covered:

- The new hosted services
- The unit test model builders

Warning count before: 34216
After: 34031
Resolved: 185...baby steps!


bergmania:
🚀

PR #9504, opened 2020-12-05
Support dependency injection for Examine IDirectoryFactory

nzdev:
### Prerequisites

### Description
What?
Retrieve new IDirectoryFactory instances from the container instead of using Activator.

Why?
Allow registering IDirectoryFactory instances with dependencies

How to test?
Case 1
Use a ICoreComposer to register an IDirectoryFactory with dependencies.
Set the registered IDirectoryFactory in web.config for key "Umbraco.Examine.LuceneDirectoryFactory"
Boot Umbraco.

Case 2
Set a built in Examine IDirectoryFactory in web.config for key "Umbraco.Examine.LuceneDirectoryFactory"
Existing IDirectoryFactory loads without having to be registered in code with the IOC container.

### DIFF
--- a/src/Umbraco.Examine/LuceneIndexCreator.cs
+++ b/src/Umbraco.Examine/LuceneIndexCreator.cs
@@ -17,6 +17,12 @@ namespace Umbraco.Examine
     /// </summary>
     public abstract class LuceneIndexCreator : IIndexCreator
     {
+
+        private readonly IFactory _factory;
+        public LuceneIndexCreator(IFactory factory)
+        {
+            _factory = factory ?? throw new System.ArgumentNullException(nameof(factory));
+        }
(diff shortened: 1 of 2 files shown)


bielu:
@nzdev that change would be breaking change for umbraco and don't think so is good way to be introduced. Also changes in Examine library which you proposed look like you would make based on Umbraco patterns. And Examine is not really Umbraco only library. I am using Examine outside of Umbraco and your changes force me to change way of handling that.

nzdev:
@bielu True. It might be better to do this another way.

nzdev:
Reopening as I still think it's worth making examine directories support DI.

bielu:
@nzdev it is worth but in v8, you can look into netcore project and proposed that there, but sadly I am not seeing option to introduce that there :)

PR #9505, opened 2020-12-06
NetCore: updated some more code to match linting rules.

AndyButland:
Again have restricted amends here to just code we've added in the .NET Core transition to avoid issues with merges from the V8 branch. Have covered:

- The configuration model classes and validators
- The authorization policy handlers and requirements


bergmania:
🚀

PR #9508, opened 2020-12-07
Speed up deleting relationship by parent. (Speeds up publishing/ saving)

nzdev:
### Prerequisites

### Description
What?
Improve query for Delete relationship by parent.

Why?
Faster deletion of relationships = reduced time to publish.

How to test?
Existing tests pass. Existing relationships on existing content are removed when the content is saved.
Execution time should be reduced on sql server/ sql azure

### DIFF
--- a/src/Umbraco.Core/Constants-SqlTemplates.cs
+++ b/src/Umbraco.Core/Constants-SqlTemplates.cs
@@ -14,7 +14,11 @@
                 public const string GetParentNode = "Umbraco.Core.VersionableRepository.GetParentNode";
                 public const string GetReservedId = "Umbraco.Core.VersionableRepository.GetReservedId";
             }
-
+            public static class RelationRepository
+            {
+                public const string DeleteByParentAll = "Umbraco.Core.RelationRepository.DeleteByParent";
+                public const string DeleteByParentIn = "Umbraco.Core.RelationRepository.DeleteByParentIn";
+            }
         }
     }
 }
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/RelationRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/RelationRepository.cs
@@ -286,23 +286,41 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
         
         public void DeleteByParent(int parentId, params string[] relationTypeAliases)
         {
-            if (relationTypeAliases.Length > 0)
+            if (Database.DatabaseType.IsSqlCe())
             {
-                var template = SqlContext.Templates.Get(Constants.SqlTemplates.RelationRepository.DeleteByParentIn, tsql => Sql().Delete<RelationDto>()
-                  .From<RelationDto>()
-                 .InnerJoin<RelationTypeDto>().On<RelationDto, RelationTypeDto>(x => x.RelationType, x => x.Id)
-                 .Where<RelationDto>(x => x.ParentId == SqlTemplate.Arg<int>("parentId"))
-                 .WhereIn<RelationTypeDto>(x => x.Alias, SqlTemplate.ArgIn<string>("relationTypeAliases")));
-                Database.Execute(template.Sql(new { parentId, relationTypeAliases }));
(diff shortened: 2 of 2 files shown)


src/Umbraco.Core/Constants-SqlTemplates.cs
```diff
@@ -14,7 +14,11 @@ public static class VersionableRepository
                 public const string GetParentNode = "Umbraco.Core.VersionableRepository.GetParentNode";
                 public const string GetReservedId = "Umbraco.Core.VersionableRepository.GetReservedId";
             }
-
+            public static class RelationRepository
+            {
+                public const string DeleteByParentAll = "Umbraco.Core.RelationRepository.DeleteByParent";
```
Shazwazza, line 19: I can see these were in use before but that has been reverted. Any reason why we still can't use Sql templates? Even if it's just for the non-sqlce?

Shazwazza (changes requested):
Yeah SqlCe is a bugger. There's probably quite a few queries that can be easily optimized for Sql server. Some of which could be wrapped possibly on ext methods to the SqlSyntaxProvider. Do you have any stats (i.e. Sql execution plan) of running the orig queries vs. the pure delete with inner join query? There was one failing test for some reason but don't think it's related to this, i've just re-run the build to see what happens.

src/Umbraco.Core/Constants-SqlTemplates.cs
nzdev: I did set up the template, but I didn't know enough about how they worked with two sets of parameters so I reverted the change. Hopefully I can get some assistance with this. I think it went from ~1000ms to <40ms. But I'm not sure. There are too many queries I've looked at lately to remember. But the difference was huge.

nzdev: Original
DELETE [umbracoRelation] 
where id in ( 
SELECT r.id
  FROM [dbo].[umbracoRelation] r
  INNER JOIN  umbracoRelationType rt on rt.id = r.relType
  where r.parentId = 155928)

image

New:
  DELETE umbracoRelation
  FROM umbracoRelation
  INNER JOIN umbracoRelationType rt on rt.id = relType
  where umbracoRelation.parentId = parentId

image

Shazwazza:
> I did set up the template, but I didn't know enough about how they worked with two sets of parameters so I reverted the change. Hopefully I can get some assistance with this

ok cool no worries, i can have a look when merged too. I haven't looked at that for a while so not sure off the top of my head.

For the query comparisons - if you put both queries in the same query window and execute, it will show results from both queries along with side by side results of the execution plan and it will tell you the "% cost relative to the batch" so you can see exactly by how much % one query is over the other.

Shazwazza:
I get this result 
image

So shows the newer version is definitely better 🎉

Shazwazza:
I'll get this merged and have a look at the SQL templates. Great stuff :)

Shazwazza:
@nzdev here's an update which now uses Sql templates for these https://github.com/umbraco/Umbraco-CMS/commit/9269271649b9b2f6296fb09b6893b3d27c0f94bf

nzdev:
@Shazwazza have a look at https://github.com/umbraco/Umbraco-CMS/pull/9508/commits/39b1de4a98c5d009bb607370fb6dc2590698b1aa . I think the issue was argIn<string> didn't work

Shazwazza:
@nzdev strange, i've just tested on my machine and it executes just fine, but that commit you've referenced also wasn't checking for sqlce so the build may have been failing because of that

nzdev:
@Shazwazza Possibly, though I checked the query in the sql profiler at the time and it was not creating the in clause for multiple alias's properly.

Shazwazza:
This is the resulting sql string

```
DELETE [umbracoRelation]
FROM [umbracoRelation]
INNER JOIN [umbracoRelationType]
ON [umbracoRelation].[relType] = [umbracoRelationType].[id]
WHERE (([umbracoRelation].[parentId] = @0))
AND ([umbracoRelationType].[alias] IN (@1,@2))
```

with these args

image

nzdev:
looks correct to me 👍

PR #9511, opened 2020-12-08
Cleans up usage of SystemDirectories, removes App_Data usages

Shazwazza:
* Cleans up usage of SystemDirectories
* Removes App_Data usages
* Removes AspNetTemp as an option since that doesn't exist anymore
* Linting cleanup

### DIFF
--- a/src/Umbraco.Core/Constants-SystemDirectories.cs
+++ b/src/Umbraco.Core/Constants-SystemDirectories.cs
@@ -1,4 +1,4 @@
-﻿namespace Umbraco.Core
+namespace Umbraco.Core
 {
     public static partial class Constants
     {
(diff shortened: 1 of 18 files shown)


src/Umbraco.Core/Constants-SystemDirectories.cs
```diff
@@ -1,15 +1,33 @@
...
     public static partial class Constants
     {
         public static class SystemDirectories
         {
+            /// <summary>
+            /// The aspnet bin folder
+            /// </summary>
             public const string Bin = "~/bin";
 
+            // TODO: Shouldn't this exist underneath /Umbraco in the content root?
```
bergmania, line 12: IMO it could as well, yes.. But historically it exists in the root

bergmania (approved):
Changes looks good.  On thing we will most likely have to still change is some of the SystemDirectories. For optimal docker images the dynamic of these should be outside the root of the solution. So the solution is static in the image.

PR #9515, opened 2020-12-08
NetCore: Get rid of OverrideAuthorizationAttribute

elit0451:
Following the approach of having special restrictive authorization on the various actions (Action level). 

Therefore, we were now able to remove  `OverrideAuthorization` attribute and related.

### Notes:
- Moving `AllowsCultureVariation()` to **ContentTypeController**;
- Refactoring AngularJS implementation;
- Removing the attributes we no longer need;
- As part of this PR, conn string is removed from the appsetting file.


rustybox:
I don't have a content tree in backoffice.

I think umbraco.resources.js needs to be updated to match

return umbRequestHelper.resourcePromise($http.get(umbRequestHelper.getApiUrl('contentApiBaseUrl', 'AllowsCultureVariation')), 'Failed to retrieve variant content types');

still looks at /content instead of /contenttypes

rustybox:
Oh that was changed, I'm serving the old version aparently, belle compile missed I guess

bergmania:
Super.. I did not understand, as I have tested multiple times :P

Issue #6451, opened 2019-09-26
v 8.1.5 - Installing Package with Content results in Error: Variation "<null>, <null>" is not supported by the property type AND "Cannot save content with an empty name."


hfloyd:
This might be related to https://github.com/umbraco/Umbraco.Cloud.Issues/issues/108

I was transferring some content between two local Umbraco 8.1.5 installs with identical schemas. 

On Site1 I created a package which only included content. When I attempted to install this package on Site2, the error `Variation "<null>, <null>" is not supported by the property type` was shown. No Content was added.

As per this thread: https://our.umbraco.com/forum/extending-umbraco-and-using-the-api/96738-setting-value-for-content-picker-attribute 
I updated all the Doctypes properties on Site2 to have  "Vary by culture" set to FALSE. Now when I try to install the package again, I see a different error:

`Cannot save content with an empty name.`

```
Exception Details
System.InvalidOperationException: Cannot save content with an empty name.
Stacktrace
at Umbraco.Core.Persistence.Repositories.Implement.DocumentRepository.EnsureInvariantNameExists(IContent content)
   at Umbraco.Core.Persistence.Repositories.Implement.DocumentRepository.SanitizeNames(IContent content, Boolean publishing)
   at Umbraco.Core.Persistence.Repositories.Implement.DocumentRepository.PersistNewItem(IContent entity)
   at Umbraco.Core.Cache.DefaultRepositoryCachePolicy`2.Create(TEntity entity, Action`1 persistNew)
   at Umbraco.Core.Services.Implement.ContentService.Save(IEnumerable`1 contents, Int32 userId, Boolean raiseEvents)
   at Umbraco.Core.Packaging.PackageDataInstallation.ImportContent(CompiledPackageDocument packageDocument, Int32 parentId, IDictionary`2 importedDocumentTypes, Int32 userId)
   at Umbraco.Core.Packaging.PackageDataInstallation.<>c__DisplayClass11_0.<ImportContent>b__0(CompiledPackageDocument x)
   at System.Linq.Enumerable.<SelectManyIterator>d__17`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Umbraco.Core.Packaging.PackageInstallation.InstallPackageData(PackageDefinition packageDefinition, CompiledPackage compiledPackage, Int32 userId)
[...]

PR #9524, opened 2020-12-09
Fix for Issue #135 - Variation < null >, < null> is not supported by the property type

BeardInASuit:
Fix for isssue: https://github.com/umbraco/Umbraco.Cloud.Issues/issues/135 and https://github.com/umbraco/Umbraco-CMS/issues/6451

The root causes of this issue was two bugs:
- Culture variant content node names were not exported in package content serialization and 
- Language node property language were not used when importing DocumentSet properties. 

That triggers an exception "Variation < null >, < null> is not supported by the property type" when importing language variant content since no language variations were initialized by setting node names and property's lang attribute were not used when importing content.

The fix is to add language specific nodeNames in node serialization and importing them in the package import module. This would add the language variant that would be used when importing node properties with the right culture (lang attribute).


BeardInASuit:
@nikolajlauridsen Good catch - I didn't have this problem since I am adding the right languages in my package when exporting it. You fix would allow a package to be imported without forcing the addition of new languages.

nul800sebastiaan:
Thanks again @BeardInASuit and sorry for the delay! I've replaced some checks with some handy Invariant extension methods we have. This part of the code is not used in Deploy for Cloud so I've removed the references to the Cloud issue to avoid confusion. Other than that this is a good change for packages, thanks very much! 👍 

I would love to give you a contrib badge on [Our Umbraco](https://our.umbraco.com) if you have an account.

Thanks again and congrats on your first contribution the Umbraco CMS! 🎉

BeardInASuit:
@nul800sebastiaan my account on Our Umbraco is Beard in a Suit (286892) Thanks!

nul800sebastiaan:
👍 And now you have that cool contrib badge on your profile! 🥇

PR #9525, opened 2020-12-09
Unhelpful unsupported property message in Block Editor

callumbwhyte:
When adding a property like Nested Content to the block editor you get an error like this:

> Property XYZ uses editor Nested Content which is not supported in **Element Types**.

The exact reasons for this have been discussed in #8732 - I'm not going to argue here and now... ;-)

But this error message is rather unhelpful, and in many cases inaccurate. It is possible to put Nested Content properties inside Element Types as Nested Content itself uses elements!

I have updated all instances of this message to be more Block Editor specific. The message now reads:

> Property XYZ uses editor Nested Content which is not supported in **blocks**.


nul800sebastiaan:
Thanks @callumbwhyte - sorry for the delay, this is good to go now! 👍

PR #9526, opened 2020-12-09
Fix consistency in translation files for the term backoffice, Umbraco and several entity types

callumbwhyte:
I'm a stickler for detail, and over the years the Umbraco backoffice dictionaries have got a little untidy.

This PR cleans up the casing and spacing for a few commonly inconsistent words:

1. All entities (e.g. Data Type, Element Type) should be use capitals as they are names of entities unique to Umbraco
2. Backoffice should be a single word, as per the [Umbraco Docs Vale linting rules](https://github.com/umbraco/UmbracoDocs/blob/main/.github/valeStyle/Terms.yml)
3. The term "master" should be avoided where possible _(in fact the only key where it was present is unused)_
4. Believe it or not, it's possible to spell Umbraco incorrectly... 😬

It might be cool to see if we can introduce Vale linting on the backoffice dictionaries much like in the Docs, but multi-lingual files and XML will probably create some challenges...


abjerner:
Great 👍 

It's a topic that come up a number of times over the years, but I'd reckon it's **backoffice** as well.

nathanwoulfe:
Thanks @callumbwhyte - I noticed similar when merging another PR last week, there was a stack of inconsistency around capitalisation with regard to noun vs proper noun.

callumbwhyte:
@nathanwoulfe Think that would be a good one to look at too, but a little more involved.

Maybe you can have some examples so we can create an up-for-grabs issue?

Shazwazza:
Bear in mind that it is BackOffice for all c# code

nul800sebastiaan:
Thanks @callumbwhyte for that dose of consistency! 👍

PR #9528, opened 2020-12-09
Make abstract FileSystemWrapper class methods virtual

callumbwhyte:
The `FileSystemWrapper` class is abstract but frustratingly none of the methods are marked as abstract / virtual so can't be replaced...

The purpose of this class is to have the ability for custom file system providers to wrap existing / core providers.

In my case I have a package where I am shipping a custom file system provider but want to allow it to be wrapped and replaced, e.g. to use Azure Blob Storage. However I also want to manipulate the path before it goes to storage.

Perhaps my case is unique, but I personally can't see a reason for none of the methods to be overridable... 😊

### DIFF
--- a/src/Umbraco.Core/IO/FileSystemWrapper.cs
+++ b/src/Umbraco.Core/IO/FileSystemWrapper.cs
@@ -23,94 +23,94 @@ namespace Umbraco.Core.IO
 
         internal IFileSystem InnerFileSystem { get; set; }
 
-        public IEnumerable<string> GetDirectories(string path)
+        public virtual IEnumerable<string> GetDirectories(string path)
         {
             return InnerFileSystem.GetDirectories(path);
         }
 
-        public void DeleteDirectory(string path)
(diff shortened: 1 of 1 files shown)


Shazwazza:
Lesser known tip for c# -  explicitly implementing any interface properties and methods 'override' the class that implements them and they don't need to be virtual to do that.

callumbwhyte:
Yup, this is how I've hacked my implementation together currently. :-)

But if I want to make one of those "overridden" methods virtual so someone can extend / override my implementation, for example, that's not possible. Perhaps there's an argument I should be exposing my own methods and wrapping the ones I want to override, but then both will be public in the class.

Shazwazza:
I wouldn't say that's hacky 😃 having then virtual is fine just wanted to mention the explicit interface. Someone can inherit from your class and explicitly implement too and override yours without virtual.

nul800sebastiaan:
Thanks @callumbwhyte!

PR #9529, opened 2020-12-09
Block list action z index

callumbwhyte:
The actions (appearing to the right hand side of block list editors on hover) clash with other UI elements, such as the publish actions menu.

Screenshot 2020-12-09 at 22 23 01

There is an insanely high `z-index` value set here to ensure they always stay on top of all custom + Umbraco styles loaded.

I have reduced the `z-index` to a more sensible level, inline with the similar `umb-confirm-action` component to avoid the clashes. Now it looks like this:

Screenshot 2020-12-09 at 22 29 08

_(I also cleaned up some of the CSS here, maybe want to review with whitespace ignored)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/blocklistentryeditors/inlineblock/inlineblock.editor.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/blocklistentryeditors/inlineblock/inlineblock.editor.html
@@ -8,6 +8,7 @@
         <i class="icon {{block.content.icon}}" aria-hidden="true"></i>
         <span class="name">{{block.label}}</span>
     </button>
+
     <div class="blockelement-inlineblock-editor__inner" ng-class="{'--singleGroup':block.content.variants[0].tabs.length === 1}" ng-if="block.active === true">
         <umb-element-editor-content model="block.content"></umb-element-editor-content>
     </div>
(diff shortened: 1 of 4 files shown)


glombek:
Hi Callum,

Thanks for this PR!

However, this doesn't appear to be working for me.

`.umb-editor-footer` has `z-index: 30` so your `z-index: 999999` on `.umb-block-list__block-actions` still paces the buttons over the top of the publish menu.

Am I missing something?

Setting the `umb-block-list__block-actions` `z-index` to 30 or less puts it below the publish menu, but I'm not sure what else the buttons are trying to be above (custom block list views, perhaps?)

It may make sense to force `.umb-split-views` to be below `.umb-editor-footer` (<= 30) so that nothing within `.umb-split-views` can possibly escape that frame... but this may have other impacts!

Be good to hear your thoughts here.

Joe

glombek:
In fact, setting `.umb-block-list__block-content>div` to have a `z-index: 0`, and then `.umb-block-list__block-actions { z-index: 1; }` would be less obtrusive! I'm not quite sure how the custom view stuff would work here though - maybe another container would need to be set to `z-index:0` also.

nul800sebastiaan (changes requested):
@callumbwhyte could you try to make the changes as @glombek suggests?

nathanwoulfe:
Looking at this on v8/contrib, I think it can be fixed by removing the z-index from umb-block-list__actions. With no z-index, the controls still sit on top of any custom view, and always behind the footer menu. 

The umb-block-list__block element is statically positioned, then has the umb-block-list-block and actions elements as children:
image

Given the actions element is absolutely positioned, it will create a new stacking context and always be atop the preceding element (unless the block has silly custom CSS to set its own z-index, which would already cause problems with overlapping the footer menu etc).

I think the easiest way forward is to kill the z-index on the block actions, and flag it as a potentially breaking change for custom block list views where the list view has a z-index (which it shouldn't need anyway)

Issue #9397, opened 2020-11-14
Netcore: Remove static events


p-m-j:
There are 180 static events in the solution.

ContentService has 27 static events, they cause a world of pain both when building umbraco sites and when testing.
+ Saving
+ Saved
+ Publishing
+ Published
+ ...

These can be replaced with an in process [EventAggregator](https://martinfowler.com/eaaDev/EventAggregator.html)

We could have a simple one defined within Umbraco with an implementation similar to [Udi Dahan - DomainEvents example](https://udidahan.com/2009/06/14/domain-events-salvation/)

Or we could make use of [MediatR](https://github.com/jbogard/MediatR), which could be referenced in Umbraco.Infrastructure with an Umbraco abstraction wrapping it.

Lots of discussion on this topic in #8653 and probably elsewhere.

----

If this work were to take place, IComponent is no longer required, most of the components just hook static events and could disappear.

Any existing components that must exec code on application startup e.g. ManifestWatcherComponent can be replaced by a handler for an ApplicationStarting event, this would have a RuntimeLevel property such that code could be conditionally run as required.

Shazwazza:
Events are a mess 😭 There's a lot of legacy in there and a lot of things that should be killed or do strange things that shouldn't be done that way. It's gonna be quite an undertaking to get this all done but the amount of code and pain that will be removed will be very substantial. I'm not sure what the best approach to take here is going to be while minimizing the sweeping changes that will need to be done so the work can be done in smaller increments. Potentially one way would be to leave the events as-is currently and add the event aggregator code side-by-side then iteratively remove the old events with the new replaced handlers. Some things to know/keep in mind:

* The way Scopes (transaction / uow) work is to only let events raise once the trans is complete for the completed style events (suffixed with "ed" like "Saved") though these events are registered to be raised from within the services themselves. This code could be massively simplified without having to work around limitations of c# events.
* events suffixed with "ing" like "Saving" are raised during the transaction at the time they are raised in the service and in many cases are cancelable 
* Some events exist at the repository level due to legacy and/or never getting around to fixing that but they shouldn't exist there at all
* Some events probably don't make any sense and could be removed but that would be a case by case assessment
* There are other static events that are not just on services that will need to be changes
  * One that needs to be changed is the BackOfficeUserManager. This has static events but the instance is a Scoped instance so it cannot have normal events. I suppose that makes it an ideal candidate for an event aggregator since we don't have to worry about binding to an instance.
* This work will require some patience

[...]

PR #9530, opened 2020-12-10
Starts getting front-end routing/rendering working

Shazwazza:
This will be the first step in getting the front-end routing/rendering working. I've added a ton of TODOs a lot of these will need to be cleaned up on the 2nd iteration but should be looked at and commented on if necessary as part of the review.

* PublishedSnapshotService - split up
  * Repository for DB interaction
    * Changes to repo base class to make things easier, RepositoryBase (for all repos) + EntityRepositoryBase (for IEntity based repos) and no more NPocoRepositoryBase
  * Service for repository interaction - INuCacheContentService replaces IDataService
  * PublishedSnapshotServiceEventHandler for wiring events for PublishedSnapshotService
  * Remove EnterPreview/RefreshPreview/ExitPreview since this was only used for the xml cache
  * Still work in progress, there's still temporal coupling with `LoadCachesOnStartup`, how to avoid this?
* UmbracoRouteValueTransformer for handling the dynamic routes - will basically replace RenderRouteHandler and the module
* ContentModelBinder fixed up and tests for it combined into one suite (there was 2x test classes for the same thing)
* UmbracoViewPage fixed up so that the dynamic model binding works
* Fixup the startup ordering for UseUmbraco, etc... UseRouting is what installs endpoint routing so any middleware after that will not execute before dynamic routes in endpoint routing even if they are declared before UseEndpoints. However some things must be declared between UseRouting and UseEndpoints like AuthN, AuthZ. Code has also been updated to cater better for enabling front-end vs back office 

Others

* Remove IPublishedContentType2 backward compat hack
* Remove other unnecessary things
* Linting cleanup

This POC will not contain surface controller routes, post backs, forms, content finder redirections, route hijacking, custom routes, macros, etc... just the first step of rendering something

Further PRs for front-end routing will be based on this work

## testing

* Ensure tests are passing
* Create a content type with a template, let's say it's called "Home", that should create a view under ~/Views/Home.cshtml
* put some simple markup in there like (basically render a menu system and the name)
   ```
    @using Umbraco.Web.PublishedModels;
    @using Umbraco.Core;
    @using Umbraco.Web.Routing; 
    @inject IPublishedUrlProvider urlProvider;
    @inherits Umbraco.Web.Common.AspNetCore.UmbracoViewPage
    @{
	    Layout = null;
        var root = Model.Root();
    }

    <ul>
        <li><a href="@root.Url(urlProvider)">@root.Name</a></li>
        @foreach (var child in root.Children)
        {
            <li><a href="@child.Url(urlProvider)">@child.Name</a></li>
        }
    </ul>

    <h1>@Model.Name</h1>
   ```
* create a content item and publish it
* the url will be populated, click on it and your site should render
* all published pages should render


src/Umbraco.Infrastructure/Persistence/Repositories/Implement/ContentRepositoryBase.cs
```diff
@@ -767,8 +767,20 @@ protected string GetQuotedFieldName(string tableName, string fieldName)
...
         #region UnitOfWork Events
 
-        // TODO: The reason these events are in the repository is for legacy, the events should exist at the service
-        // level now since we can fire these events within the transaction... so move the events to service level
+        /*
+         * TODO: The reason these events are in the repository is for legacy, the events should exist at the service
+         * level now since we can fire these events within the transaction...
+         * The reason these events 'need' to fire in the transaction is to ensure data consistency with Nucache (currently
+         * the only thing that uses them). For example, if the transaction succeeds and NuCache listened to ContentService.Saved
+         * and then NuCache failed at persisting data after the trans completed, then NuCache would be out of sync. This way
```
JimBobSquarePants, line 775: Why would NuCache fail to persist data?

I may be reading this comment  wrong but I don't think a failure of an event subscriber should trigger a rollback of a persistence layer,  operation; it's entirely back-to-front. Caches and the content queries that use them should be designed to handle misses. 

Save should always happen.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/EntityRepositoryBase.cs
```diff
@@ -122,66 +112,101 @@ protected virtual RepositoryCachePolicyOptions DefaultOptions
             }
         }
 
+        /// <summary>
+        /// Get the entity id for the <see cref="TEntity"/>
+        /// </summary>
+        protected virtual TId GetEntityId(TEntity entity)
```
JimBobSquarePants, line 118: I'm intrigued here. IEntity as it stands has a single `Id` property of type `int`. This means you're boxing the id.

Shazwazza: I'm intrigued too! Took me a bit to find the culprit but 5 years ago with a change I made https://github.com/umbraco/Umbraco-CMS/commit/5d965346786e9055257c66cc1813b53f077dc8c0 and then started investigating why. It's because trying to cast int to `TId` doesn't work because TId can be anything. So this was a hack to avoid opening pandoras "box" ( 😂 ) since this is a broken abstraction and could be fixed with `IEntity<TId>`. There's many other issues here and many old TODOs lying around (now wishing I didn't look here!). I made a start on refactoring but it's quite a lot of changes and would like to resolve some of the old TODOs lying around. I'll keep this as a branch for now and see how it goes there. None of it's pretty. Need to move on to what I'm supposed to be working on but will come back to it.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/ContentRepositoryBase.cs
Shazwazza: This is what we were talking about on the events chat and whether events should also be raised within the trans or not. NuCache manages data in the database along with it's in-memory/disk caches. The idea here is that the DB for nucache should never be inconsistent with what is in the DB for content so a rollback will trigger if it fails because these events are raised within the trans.

I agree with what you are saying and that is what I would have done, but based on discussions with various folks a while back the 'safe' thing to do is to have any DB changes related to the current events done within the transaction so everything is guaranteed consistent. This is partly what @lars-erik was referring to in the events chat https://github.com/umbraco/Umbraco-CMS/issues/9397#issuecomment-741677912 

But the thing is, that only guarantees consistency at the DB level, there are other non-DB things that shouldn't run in the transaction like how lucene indexes get populated. 

I wrote all these code comments here because I had to re-investigate all this and how/why it was done so at least we have a clue why they exist.

JimBobSquarePants: >I agree with what you are saying and that is what I would have done, but based on discussions with various folks a while back the 'safe' thing to do is to have any DB changes related to the current events done within the transaction so everything is guaranteed consistent.

That's most unfortunate. I would definitely aim to refactor to remove transactional coupling in this manner. Repositories should certainly raise an event containing a collection of entities updated during and after a transaction but NuCache et all, should really only be awaiting notifications for once a transaction is complete. If something fails they can log. A `PublishedContentService` can then handle cache misses.

nzdev: Agreed. In a load balanced scenario nucache is going to be out of sync for some amount of time anyway.

Shazwazza: > In a load balanced scenario nucache is going to be out of sync for some amount of time anyway.

This isn't about nucache's in-memory caches though, this is about nucache's source data which is in the database. Having it within the transaction guarantees that the nucache source data is always consistent with the umbraco source data.

Would like to hear @lars-erik 's opinion here since I believe he is for having some events such as this to be raised within the trans for this exact purpose. I see he's added 🚀 to james' comment but pretty sure that is different to what was mentioned here https://github.com/umbraco/Umbraco-CMS/issues/9397#issuecomment-741677912?

rustybox: Is there a problem with eventual consistency here?

lars-erik: I believe Umbraco have always benefited from treating content data and content cache as separate bounded contexts if you will. Which means, if we wanna follow good practices, that there should only be serializable messages passed between them.

Events that mutate the _entities_ could/should be ran in the same transaction. Events that perform operations in other bounded context should be possible to run OOP and in separate transactions. IE. write NuCache source data.  
NuCache is already toppling over, going out of sync etc. making us click those red buttons from bottom to top. The existing "validity layer" dependent of in-transaction and in-proc therefore obviously does not work and hence should and could be OOP instead. 😉

Often, as I've mentioned other places, I like to slap a transaction on the entire request. This means that inproc stuff would run tin the same transaction anyway. Maybe the entities would have a nested one. (The UOW at leasst)  
The important bit is to know when to allow, if at all, mutating of entities being modified, and when to mandate that the handlers have to mind their own transactions and UOW instances. The latter will still allow for running stuff inproc.  

Hope my ramblings make sense.

`EntitySaving` -> `InProcHandlerMutatingEntityOrCanceling` -> Backoffice Domain / Content Domain / Data Layer  
`ContentPublished` -> `OutOfProcHandlerWritingNuCacheData` -> Front-End Domain / Cache Layer  

Dunno if @rustybox or @JimBobSquarePants agree it's an important distinction and have ideas on how to let the APIs communicate that?

Shazwazza: Thanks @lars-erik makes sense to me, I just wanted to know your standpoint. As mentioned above I agree with that. It's not how it is currently done with these extra events and we can refactor them out. 

> Events that mutate the entities could/should be ran in the same transaction.

Yes that is what we have today so that is fine.

> Events that perform operations in other bounded context should be possible to run OOP and in separate transactions. IE. write NuCache source data.

yep so we're all in agreement on this then

rustybox: @lars-erik what's your thoughts on out of proc events, do you see them being distributed via event aggregator or are you thinking more a seperate service bus + broker?

JimBobSquarePants: Man, I should have not bothered building the aggregator if no one is going to use it!

rustybox: But of course we will use it for in proc events e.g ing (saving/publishing) events, my question is what’s the vision for the other events perhaps that’s still eventaggregator or perhaps there are separate classes to communicate between bounded contexts.

lars-erik: The event aggregator is what I'm asking between the lines. Do we pass all events to it? Do we mark events in some way to indicate whether they're safe to mutate? Should the handlers be marked in some way? Is it relevant? Should there be one per bounded context / domain?  
As of now I think all events should go through the same aggregator and have the same abstractions. The parameter class either have references to mutable things or not. The latter might possibly also be immutable "by design" just to indicate that it's supposed to be serialized and handled out of proc.  
For entities the DbContext events should translate to something for use with the aggregator. IE. UmbracoEntitiesSaved with a copy of the change tracker's entity references. (I'm presuming we're going for EF?)  
IMO it's important to prevent any leakage of any library (MS or not) into the domain events.

JimBobSquarePants: >[Edit]: Missed question from @rustybox above. I think everything should go through the event aggregator as mentioned. If you want to do something completely out of proc, then you make a handler to pass the message on to whatever you want to distribute via. A queue, a message bus, whatever.
This is all enabled by having a core aggregator (or more...) / command handling framework.
It's basically why we do this. 😁

YES!!!

All events should be triggered by the aggregator.
`INotificationHandler` instances then do whatever with the info.
Whether an event is mutable is determined by the properties of the passed `INotification` instance. If you don't want the properties to be mutable make them get only.

lars-erik: All good. So we're saying "The parameter class either have references to mutable things or not."

JimBobSquarePants: Yep, exactly. That's the power of it. You enforce the rules you want on a per instance basis.

rustybox: Coolio

lars-erik:
```csharp
public class EntitiesSaving : INotification
{
    public IEnumerable<IEntities> ChangedEntities { get; }
}

public class ContentPublished : INotification
{
    public IEnumerable<ContentDto> Content { get; }
}
```

rustybox: No more giant inheritance chain?
```csharp
ContentSavedEventArgs : SaveEventArgs<IContent>
SaveEventArgs<TEntity> : CancellableEnumerableObjectEventArgs<TEntity>
CancellableEnumerableObjectEventArgs<TEventObject> : CancellableObjectEventArgs<IEnumerable<TEventObject>>
CancellableObjectEventArgs<TEventObject> : CancellableObjectEventArgs
CancellableObjectEventArgs : CancellableEventArgs
CancellableEventArgs : EventArgs
```

lars-erik: My favorite way of cancelling event pipelines:

`throw new Exception("This does not compute!")`

src/Umbraco.Web.Website/Routing/NoContentRoutes.cs
```diff
@@ -35,13 +39,16 @@ public void CreateRoutes(IEndpointRouteBuilder endpoints)
                 case RuntimeLevel.Upgrade:
                     break;
                 case RuntimeLevel.Run:
+
+                    // TODO: I don't really think this is working AFAIK the code has just been migrated but it's not really enabled
+                    // yet. Our route handler needs to be aware that there is no content and redirect there. Though, this could all be
+                    // managed directly in UmbracoRouteValueTransformer. Else it could actually do a 'redirect' but that would need to be
```
bergmania, line 45: We did an internal rewrite before (in the netcore/dev branch). And that worked.

bergmania (commented):
Just a few comments.. Still need to look at some of the long files

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/NPocoRepositoryBase.cs
```diff
@@ -13,9 +13,7 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
     /// <summary>
     /// Represent an abstract Repository for NPoco based repositories
     /// </summary>
-    /// <typeparam name="TId"></typeparam>
-    /// <typeparam name="TEntity"></typeparam>
-    public abstract class NPocoRepositoryBase<TId, TEntity> : RepositoryBase<TId, TEntity>
+    public abstract class NPocoRepositoryBase<TId, TEntity> : EntityRepositoryBase<TId, TEntity>
```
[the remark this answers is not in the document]
Shazwazza: No point, needs to be removed, will cleanup after this PR (have a task)

src/Umbraco.Web.Common/Controllers/RenderController.cs
```diff
@@ -1,10 +1,99 @@
...
+using Umbraco.Web.Routing;
 
 namespace Umbraco.Web.Common.Controllers
 {
-    public abstract class RenderController : Controller, IRenderController
+
+    /// <summary>
+    /// Represents the default front-end rendering controller.
+    /// </summary>
+    [TypeFilter(typeof(ModelBindingExceptionFilter))]
```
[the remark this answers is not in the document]
Shazwazza: Got all tests passing, will fix these last minor things up tomorrow and merge

rustybox:
This is awesome, I have a homepage rendered 👍 👍  👍 👍  👍 👍

PR #9531, opened 2020-12-10
Add support for unattended upgrades (and custom boot failed error page)

rasmusjp:
There are two parts to this PR.

The first part is unattended upgrades, it's an opt-in feature and can be enabled by adding an app setting `Umbraco.Core.RuntimeState.UpgradeUnattended=true`.

In addition to the above setting the `Umbraco.Core.ConfigurationStatus` app setting must match the assembly version before unattended upgrades are run.

If the conditions are met, the Runtime level will be `Run` instead of `Upgrade`, this allows us to do run the migrations and continue with the normal boot when the migrations succeed without requiring to restart the application. 

The upgrade is run after Composers but before Components. This is because the migrations requires services that are registered in Composers and Components requires that Umbraco (and the database) is ready.

To test the unattended upgrade, a database with an older Umbraco version is needed (I've tested with an 8.0.0 database). Update the connection string to point to that database and update the required app settings as mentioned above, then start Umbraco as you normally would. Now instead of the seeing the upgrade page, Umbraco runs the migrations on startup and you should see the website instead. 

The second part of this PR is a custom error page for when a `BootFailedException` is thrown. The error page is a HTML file located in `~/umbraco/views/errors/BootFailed.html` it can be overwritten by creating a file `~/config/errors/BootFailed.html` (not sure about the location).

There are multiple ways to test the error page, some of them are

* Configure an invalid connection string
* Have unattended upgrades enabled and add an migration that throws an exception
* Throw an exception in a Composer or Component
* Recursive dependencies in a Component

Note that the error page is only shown if debugging is disabled by setting `/configuration/system.web/compilation/@debug` to `false` in `web.config`. Not sure if we should also check the `customErrors/@mode`?

The error page could use some love, right now it's a copy of the login page

image

**NOTE this PR targets the `v8/feature/unattended-install` branch since it relies on changes made there**

---
_This item has been added to our backlog [AB#9755](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/9755)_

### DIFF
--- a/src/Umbraco.Core/Runtime/CoreRuntime.cs
+++ b/src/Umbraco.Core/Runtime/CoreRuntime.cs
@@ -13,6 +13,7 @@ using Umbraco.Core.IO;
 using Umbraco.Core.Logging;
 using Umbraco.Core.Logging.Serilog;
 using Umbraco.Core.Migrations.Install;
+using Umbraco.Core.Migrations.Upgrade;
 using Umbraco.Core.Persistence;
 using Umbraco.Core.Persistence.Mappers;
 using Umbraco.Core.Sync;
@@ -189,6 +190,16 @@ namespace Umbraco.Core.Runtime
                 // create the factory
                 _factory = Current.Factory = composition.CreateFactory();
 
+                // if level is Run and reason is UpgradeMigrations, that means we need to perform an unattended upgrade
+                if (_state.Reason == RuntimeLevelReason.UpgradeMigrations && _state.Level == RuntimeLevel.Run)
+                {
+                    // do the upgrade
+                    DoUnattendedUpgrade(_factory.GetInstance<DatabaseBuilder>());
+
+                    // upgrade is done, set reason to Run
+                    _state.Reason = RuntimeLevelReason.Run;
+                }
+
                 // create & initialize the components
                 _components = _factory.GetInstance<ComponentCollection>();
                 _components.Initialize();
@@ -286,6 +297,17 @@ namespace Umbraco.Core.Runtime
                         + "\n Please check log file for additional information (can be found in '/App_Data/Logs/')");
                 }
             }
+        }
+
+        private void DoUnattendedUpgrade(DatabaseBuilder databaseBuilder)
+        {
+            var plan = new UmbracoPlan();
+            Logger.Info<CoreRuntime>("Starting unattended upgrade.");
+            var result = databaseBuilder.UpgradeSchemaAndData(plan);
+            Logger.Info<CoreRuntime>("Unattended upgrade completed.");
+
+            if (result.Success == false)
+                throw new UnattendedInstallException("An error occurred while running the unattended upgrade.\n" + result.Message);
         }
 
         protected virtual void ConfigureUnhandledException()
--- a/src/Umbraco.Core/Runtime/CoreRuntime.cs
+++ b/src/Umbraco.Core/Runtime/CoreRuntime.cs
@@ -302,12 +302,12 @@ namespace Umbraco.Core.Runtime
         private void DoUnattendedUpgrade(DatabaseBuilder databaseBuilder)
         {
             var plan = new UmbracoPlan();
-            Logger.Info<CoreRuntime>("Starting unattended upgrade.");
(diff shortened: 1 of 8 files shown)


Shazwazza:
@rasmusjp have added to next sprint, won't have time this sprint to get to this one

src/Umbraco.Web/Composing/ModuleInjector.cs
```diff
@@ -35,7 +35,16 @@ public void Init(HttpApplication context)
                 catch { /* don't make it worse */ }
 
                 if (runtimeState?.BootFailedException != null)
-                    BootFailedException.Rethrow(runtimeState.BootFailedException);
+                {
+                    // if we throw the exception here the HttpApplication.Application_Error method will never be hit
```
Shazwazza, line 39: This will be an interesting one to port to netcore ;)

src/Umbraco.Web/UmbracoApplicationBase.cs
```diff
@@ -15,7 +14,14 @@ namespace Umbraco.Web
...
     {
-        private IRuntime _runtime;
+        // needs to be static, if and exception is thrown in Application_Start the current HttpApplication is discarded
+        // and a new instance is created, meaning we loose value if it's already been set.
+        private static IRuntime _runtime;
+
+        /// <summary>
+        /// Gets the current <see cref="IRuntime"/>.
+        /// </summary>
+        protected IRuntime Runtime => _runtime;
```
Shazwazza, line 24: there's already a `protected  abstract IRuntime GetRuntime()` so don't think exposing this as protected should be done?

src/Umbraco.Web/UmbracoApplicationBase.cs
```diff
@@ -15,7 +14,14 @@ namespace Umbraco.Web
     /// </summary>
     public abstract class UmbracoApplicationBase : HttpApplication
     {
-        private IRuntime _runtime;
+        // needs to be static, if and exception is thrown in Application_Start the current HttpApplication is discarded
+        // and a new instance is created, meaning we loose value if it's already been set.
+        private static IRuntime _runtime;
```
Shazwazza, line 19: This static ref is the main thing that's worrisome. Is there any way you can think of that would have this cause problems by pinning this statically? Off the top of my head I can't think of any since IRuntime is a singleton anyways, if the boot fails, it fails until the app is restarted right? Any other concerns about this you can think of?

Shazwazza (changes requested):
@rasmusjp Left a couple comments but apart from that seems reasonable. If you can also fix up the merge conflict that would be great.

src/Umbraco.Web/UmbracoApplicationBase.cs
rasmusjp: `GetRuntime()` creates a new instance when called, but given I didn't use it anyway I've removed the property

src/Umbraco.Web/UmbracoApplicationBase.cs
rasmusjp: I've reverted the change since I didn't end up needing the `IRuntime` anyways

rasmusjp:
I've rebased with the latest changes from `v8/contrib`

src/Umbraco.Core/Runtime/CoreRuntime.cs
```diff
@@ -288,6 +299,17 @@ private void DoUnattendedInstall(IUmbracoDatabaseFactory databaseFactory)
             }
         }
 
+        private void DoUnattendedUpgrade(DatabaseBuilder databaseBuilder)
+        {
+            var plan = new UmbracoPlan();
+            Logger.Info<CoreRuntime>("Starting unattended upgrade.");
```
Shazwazza: I changed this to `using (ProfilingLogger.TraceDuration<CoreRuntime>("Starting unattended upgrade.", "Unattended upgrade completed."))` to profile the upgrade since it provides a little more info.

Shazwazza (approved):
Tested with/without this setting and with/without debug for upgrades and all works

PR #9543, opened 2020-12-14
SqlMainDomLock will stop listening if Sql Server connection terminates

Shazwazza:
NOTE: This PR is targeting the 8.6 branch

Related to #8398 which ensured that if any Azure Sql transient errors occurred that SqlMainDomLock would not stop listening. This didn't take into account exception that occur when trying to create a transaction which can occur if the Sql server cannot be connected to at all - for example if the service is stopped or paused or if the sql user can no longer connect.

If this unhandled exception occurs, MainDom stops listening and shuts itself down which is not what we want to happen. This extends the work from #8398 to prevent MainDom from shutting down if Sql server can no longer be reached. This in theory shouldn't happen but it is in some installs which is impacting sites.

## Testing

Prior to this fix, do this:

* ensure you are using Sql Server for your DB
* enable sqlmaindomlock in appsettings `<add key="Umbraco.Core.MainDom.Lock" value="SqlMainDomLock"/>`
* run with the debugger attached
* put a breakpoint on MainDom.OnSignal
* open sql server management, right click on your server node and select "Pause" to pause the service
* your breakpoint will hit MainDom.OnSignal which means it will shutdown - this is unwanted behavior

With this fix applied

* repeat the steps (ensure you restart the Sql server service first!) and you will not see MainDom.OnSignal hit
* while you are still debugging, resume the Sql server service
* open the `ClientDependency.config` and add some whitespace and save the file - this will restart the site since it's a .net config file - MainDom.OnSignal will be hit - this is expected behavior

You will see a huge amount of errors in the logs, one every 2 seconds while the SQL server is unreachable. This is expected.

### DIFF
--- a/src/Umbraco.Core/Runtime/MainDom.cs
+++ b/src/Umbraco.Core/Runtime/MainDom.cs
@@ -144,8 +144,16 @@ namespace Umbraco.Core.Runtime
 
             _logger.Info<MainDom>("Acquiring.");
 
-            // Get the lock 
-            var acquired = _mainDomLock.AcquireLockAsync(LockTimeoutMilliseconds).GetAwaiter().GetResult();
+            // Get the lock
+            var acquired = false;
+            try
+            {
+                acquired = _mainDomLock.AcquireLockAsync(LockTimeoutMilliseconds).GetAwaiter().GetResult();
+            }
+            catch (Exception ex)
+            {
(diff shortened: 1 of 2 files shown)


bergmania (approved):
Code changes looks fine.. I will test

bergmania:
Behavior is as expected from the test cases 💪

PR #9546, opened 2020-12-14
Fix typo

rasmusjp:
Missing "d" in "bredde" single "d" is used as an action while two "d" is used for measurements


OwainWilliams (approved):
Thanks for updating this Rasmus.

OwainWilliams:
Thanks for the update @rasmusjp - Google Translate also agrees with you 👍 😄

Issue #9446, opened 2020-11-24
Netcore: Remove internal composers


p-m-j:
Further enhancements to simplify the service registration story continuing on from #8653 

We no longer require IComposer implementations in the Umbraco codebase, however the concept must continue to exist for plugin developers such that they can continue to support no-code installations.
 
Those with the ability to edit (or replace) Startup.cs should prefer Startup.ConfigureServices & extensions on IUmbracoBuilder to register their services over custom composer implementations.

The following tasks have been constructed based on discussion in #8653

- [x] Rename IComposer to IUmbracoComposer to boost search engine results.
- [ ] Remove CoreComposer, UserComposer, there should exist only IUmbracoComposer
- [x] All internal implementations of IComposer should be replaced with extensions on IUmbracoBuilder
- [x] Remove Composers.cs, we don't require ComposeAfter or DisableComposer attributes, we will run all IUmbracoComposer instances in the order they are discovered.

Ordering of UmbracoBuilder extensions for adding core service shouldn't be important, why would Umbraco core replace its own registrations.

However, there may be some legacy cruft here so keep an eye out for duplicate registrations of services in the existing composers.

JimBobSquarePants:
>Add collection of Actions on UmbracoBuilder so a composer has a final chance to replace services after other composers run (we do not need to worry about what happens if two composers fight over late registration, this isn't supported in 8 and wont be in 9 either).

I think this is something someone can do from startup if they require. If do what I propose re collections there will be no complexity whatsoever.

rustybox:
> I think this is something someone can do from startup if they require directly against the builder there. If we do what I propose re collections there will be no complexity whatsoever.

This is the workaround for no code late reg, should only be used by plugin devs.

@Shazwazza what was the point of this again?

JimBobSquarePants:
>This is the workaround for no code late reg, should only be used by plugin devs.

I thought we'd already established that this was an edge case not worth the effort and that if someone explicitly wanted to do that they use startup?

Plugin devs won't be doing stuff like that anyway. Most plugins are property value converters.

rustybox:
The last said on the matter was

>Most concrete examples are package running things after certain core components but if all core services are registered ourselves before any package components are run then this scenario doesn't ever occur. I don't have a concrete example of a package that relies on another package using ComposeAfter but i'm sure it exists someplace, but like I said above, that would be solved in a simpler way by defining a callback during composition like the OnContainerBuilding example.

[...]

PR #9556, opened 2020-12-16
Netcore: composer / component cleanup low hanging fruit

rustybox:
Made a start on removing some components & composers

Progress towards #9446 also somewhat related to #9397.


rustybox:
Obviously the logging fixed the sql timeouts in ThreadSafetyServiceTest tests...

bergmania (approved):
Looks good.. Just wonder why this is required to execute that early

src/Umbraco.Infrastructure/Composing/HostBuilderExtensions.cs
```diff
@@ -1,19 +1,24 @@
...
     public static class HostBuilderExtensions
     {
         /// <summary>
-        /// Assigns a custom service provider factory to use Umbraco's container
+        /// Adds CoreRuntime as HostedService
         /// </summary>
-        /// <param name="builder"></param>
-        /// <returns></returns>
+        /// <remarks>
+        /// Should be called before ConfigureWebDefaults.
```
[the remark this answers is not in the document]
rustybox: Ah, it doesn't. I should have cleaned up better, new commit incoming.

This branch started with CoreRuntime added to the service collection just like everything else in Startup.ConfigureServices. Everything was fine for the site, CoreRuntime would start before the web host.

This was a problem for the test setup however because Startup.Configure is where we configured UmbracoDatabaseFactory to use one of the test databases, this would happen after CoreRuntime started and determined run level. RunLevel would always be install during tests when components were initialized.

So i moved the hosted service registration into the host builder extension and called it after UseWebHost, that way Startup.Configure comes before CoreRuntime start, good for tests bad for website e.g. install routes which are conditionally added based on run level.

I had a while where I would call HostBuilder.UseUmbraco first in website and last in tests, this worked but is obviously weird as fuck so I knew I had to come back and fix it.

The fix is simple, just call UseTestDatabase before IHost start, that way we can use the same order in tests and when running the site.

I cleaned up to remove odd instructions around tests but really I should have nuked the whole thing.

rustybox:
Guessing you re-queued the build?

Doesn't feel very stable if it's failing 2/3 of the time :(

Any ideas what's up? I think the ITestDatabase implementations could do with a tad more logging.

rustybox:
Ohh RuntimeState isn't newed up for the initial create, that feels wrong

bergmania:
> Guessing you re-queued the build?
> 
> Doesn't feel very stable if it's failing 2/3 of the time in CI (works great here every time) :(
> 
> Any ideas what's up? I think the ITestDatabase implementations could do with a tad more logging.

Yes I did, but I don't know what is going on. Can't repro locally. We could add a `RetryAttribute` to that test, with a TODO to investigate. Im pretty sure it is not related to this PR

rustybox:
Yeah no idea, will dig into it later and get some better logging for LocalDbTests database

rustybox:
It's the lack of locks around RebuildSchema, both threads are trying to do the first time create where the commands are cached in BaseTestDatabase

rustybox:
e4e26159 should be much more stable

src/Umbraco.Infrastructure/Runtime/CoreInitialServices.cs
```diff
@@ -62,15 +62,17 @@
...
+    public static class CoreInitialServices
     {
-        public override void Compose(IUmbracoBuilder builder)
+        public static IUmbracoBuilder AddCoreInitialServices(this IUmbracoBuilder builder)
         {
-            base.Compose(builder);
+            builder.AddNotificationHandler<UmbracoApplicationStarting, EssentialDirectoryCreator>();
+
+            builder.Services.AddSingleton<ManifestWatcher>();
+            builder.AddNotificationHandler<UmbracoApplicationStarting, ManifestWatcher>(factory => factory.GetRequiredService<ManifestWatcher>());
```
JimBobSquarePants, line 74: I reaaaaaaally don't like this as it opens up even dispatching to abuse. 

All event handler should be transient and resolved directly by the service provider.

src/Umbraco.Core/Services/IRuntime.cs
```diff
@@ -1,22 +1,15 @@
-﻿using System;
+using Microsoft.Extensions.Hosting;
 
 namespace Umbraco.Core
 {
     /// <summary>
     /// Defines the Umbraco runtime.
     /// </summary>
-    public interface IRuntime
+    public interface IRuntime : IHostedService
```
Shazwazza, line 8: @rustybox This seems to make a lot of sense 👏 My only concern was knowing that IHostedService.StartAsync always fires before any endpoint routing happens and having a look at the source code, that actually happens quite early in startup so that's all good and the front-end is still rendering :)

Shazwazza:
@rustybox Once https://github.com/umbraco/Umbraco-CMS/pull/9589 is merged (I think it can be if tests pass unless you see anything obvious) we could look at removing other "low hanging fruit" composers and just making them IUmbracoBuilder extensions:

* Umbraco.Web.Common.Runtime.AspNetCoreComposer
* Umbraco.Web.BackOffice.Runtime.BackOfficeComposer
* Umbraco.Web.Common.Profiler.WebProfilerComposer
* Umbraco.Web.Common.Runtime.AspNetCoreBootFailedComposer (this service is already added in AspNetCoreComposer)
* Umbraco.Web.WebAssets.WebAssetsComposer -> logic could just be moved to `AddRuntimeMinifier` 
* Umbraco.Web.Common.RuntimeMinification.SmidgeComposer -> logic could just be moved to `AddRuntimeMinifier` 
* Umbraco.Web.Website.Runtime.WebsiteComposer -> logic could just be moved to `AddWebsite`

I may do some additional work on front-end rendering tomorrow before holidays which might have some overlap with the above so just let me if you want to tackle any of these. Cheers!

JimBobSquarePants:
Just to note... Only add extensions to `IUmbracoBuilder` you want to expose to the public via `IUmbracoComposer`. Anything else should be added to `UmbracoBuilder`.

Shazwazza:
`IUmbracoComposer` still needs to be named that, it's currently still `IComposer`

> . Only add extensions to IUmbracoBuilder you want to expose to the public via IUmbracoComposer. Anything else should be added to UmbracoBuilder.

Good to know, i'll be sure to keep that in mind with future changes.

We'll have to do a sweep of what should be ext methods vs what should just exist on UmbracoBuilder. There's probably a bunch that should change

rustybox:
> so just let me if you want to tackle any of these

I'm sure I could smash a few out, I had a branch where I had rid of most of these before, I parked it when I got to ModelsBuilderComposer which has some fun problems to solve as hinted at in #9446

src/Umbraco.Core/DependencyInjection/UmbracoBuilder.Events.cs
```diff
@@ -26,5 +27,22 @@ public static partial class UmbracoBuilderExtensions
...
+        /// <typeparam name="TNotificationHandler">The type of notificiation handler.</typeparam>
+        /// <param name="builder">The Umbraco builder.</param>
+        /// <param name="factory">Factory method</param>
+        /// <returns>The <see cref="IUmbracoBuilder"/>.</returns>
+        public static IUmbracoBuilder AddNotificationHandler<TNotification, TNotificationHandler>(this IUmbracoBuilder builder, Func<IServiceProvider, TNotificationHandler> factory)
+            where TNotificationHandler : class, INotificationHandler<TNotification>
+            where TNotification : INotification
+        {
+            // Register the handler as transient. This ensures that anything can be injected into it.
+            builder.Services.AddTransient(typeof(INotificationHandler<TNotification>), factory);
```
Shazwazza, line 44: Wondering if this should be TryAddTransient so that duplicated notifications/handler combinations are not accidentally added? I don't think it would be expected that the same notification/handler combination should exist more than once?

Shazwazza:
@rustybox I have a WIP PR here https://github.com/umbraco/Umbraco-CMS/pull/9598 and a bunch more work I haven't pushed yet. As it's xmas even I probably won't finish everything but have started on some of the above and others. I just don't want to end up having both of us overlapping work. There's a lot of cleanup in that branch and moving service registrations 'up' towards Core and UmbracoBuilder directly. I'll add notes in that PR before I head off today (next 30 mins :). Maybe if you do any work over the hols, base that work off of that branch.

Anyways, happy holidays everyone!! 🎄 🎉

Shazwazza:
Re: Models builder composer - yep can tackle that with ongoing tasks for front-end rendering.

src/Umbraco.Core/DependencyInjection/UmbracoBuilder.Events.cs
JimBobSquarePants: Nope. Because that will prevent more than one handler being registered at all since they’re registered against the interface.

JimBobSquarePants: Try it and watch the unit tests fail

Shazwazza: Roger that

rustybox: Presumably you can achieve the same with replace and an explicit service descriptor, but I’m not sure I get the point. Is it just trying to save someone from shooting them selves in the foot?

who’s going to add same handler twice?

Shazwazza: I don't think anyone on purpose will add the same handler twice but since handlers are going to be registered with our "Add..." ext methods, there's nothing preventing people from calling these ext methods twice. 

I've seen places in libraries that will call `Add` methods that the library's functionality needs without a care of whether it's been called before or not (i.e. `AddHttpClient`) and I think this will inevitably be done with some of our ext methods too.  

In the case of event handlers, it would not be good to accidentally have multiple event handlers of the same handler executing.

> Is it just trying to save someone from shooting them selves in the foot?

Indeed, people are quite good at shooting their own feet :P 

Maybe something to consider later as the code progresses with the migration of core composers to ext methods. Maybe in the end it might not be a worry, lets see how it goes.

Shazwazza: This could be solved by checking this and tests still pass

```cs
var descriptor = new ServiceDescriptor(typeof(INotificationHandler<TNotification>), typeof(TNotificationHandler), ServiceLifetime.Transient);
if (!builder.Services.Contains(descriptor))
{   
    builder.Services.Add(descriptor);
}
```

An example of why this might be needed:

* in https://github.com/umbraco/Umbraco-CMS/pull/9598 there's a new ext method `AddDistributedCache` to replace a composer for distributed cache stuff
* both the back office and front-end website require this to be called
* but we want the ability to have either the back office or front-end registered independently, or both
* currently that is done with `.AddBackOffice()` and `.AddWebsite()`
* in theory, both of these methods would also call `AddDistributedCache` since they both need it, alternatively this would need to be called as an additional method which is a bit annoying

Of course there's other ways to resolve this too based on what ext methods we want people to use to add the required services while trying to make it simple for end users. 

For now I've used the above code but happy to change based on feedback.

JimBobSquarePants: Does `AddDistributedCache` have to be an extension on `IUmbracoBuilder`? Remember that we want to expose the absolute minimum on that interface to general consumers.  

That fix looks fine to me btw. 👍

Shazwazza: > Does AddDistributedCache have to be an extension on IUmbracoBuilder? Remember that we want to expose the absolute minimum on that interface to general consumers.

That's just currently how it's done but as we progress with the front-end rendering stuff a lot will be updated. I agree and shouldn't expose things publicly that aren't necessary. It can already be internal but since both the back office and front-end rendering require this it's more an 'ensure' services type of call. There will probably be others like this too. Totally open for suggestions but I guess there could be a few solutions. One would be to have a different root ext method for running umbraco as a website (as opposed to booting for things like integration tests, or non-web apps)

```cs

// add all of umbraco, back office and front-end
.AddUmbracoWebApplication() // core + shared things like dist cache and other 
  .AddBackOffice()
  .AddWebsite();

// add only back office
.AddUmbracoWebApplication()
  .AddBackOffice();

// add only front end
.AddUmbracoWebApplication()
  .AddWebsite();
```

Then perhaps UseUmbracoBackOffice/UseUmbracoWebsite do some service validation checks [like UseMvc does](https://github.com/dotnet/aspnetcore/blob/b795ac3546eb3e2f47a01a64feb3020794ca33bb/src/Mvc/Mvc.Core/src/Builder/MvcApplicationBuilderExtensions.cs#L132)

PR #9557, opened 2020-12-16
Replaced <i> with umb-icon directive in languages overview

poornimanayar:
Hi,

I have replaced `<i>` with an `umb-icon` in the languages overview screen

image

Poornima

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/languages/overview.html
+++ b/src/Umbraco.Web.UI.Client/src/views/languages/overview.html
@@ -38,7 +38,7 @@
                 <tbody>
                     <tr ng-repeat="language in vm.languages | orderBy:'name' track by language.id" ng-click="vm.editLanguage(language)" style="cursor: pointer;">
                         <td>
-                           <umb-icon icon="icon-globe" class="language-icon" aria-hidden="true"></umb-icon>
+                           <umb-icon icon="icon-globe" class="language-icon"></umb-icon>
                             <span>{{ language.name }}</span>
                         </td>
                         <td>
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web.UI.Client/src/views/languages/overview.html
```diff
@@ -38,7 +38,7 @@
                 <tbody>
                     <tr ng-repeat="language in vm.languages | orderBy:'name' track by language.id" ng-click="vm.editLanguage(language)" style="cursor: pointer;">
                         <td>
-                            <i class="icon-globe" style="color: #BBBABF; margin-right: 5px;"></i>
+                           <umb-icon icon="icon-globe" class="language-icon" aria-hidden="true"></umb-icon>
                             <span>{{ language.name }}</span>
```
bjarnef, line 42: @poornimanayar it doesn't need `aria-hidden` here since the component itself takes care of this 🙂👍

poornimanayar: Thanks @bjarnef :) I have corrected it :)

nathanwoulfe (approved):
Looks good to me, thanks @poornimanayar

bjarnef:
@nathanwoulfe for now we still need icon name in class attribute. In this case is is probably only in core, but would potentially break if you overwrite the `icon-globe` with a custom font icon.

c9mbundy:
I guess one question is - are icons really semantic content or presentation - i.e. do they belong in the html at all ?

nathanwoulfe:
@bjarnef for all uses? I thought that was only the case where an SVG version didn't exist?

PR #9575, opened 2020-12-17
Dont show block catalogue if only one block is available. Enable ente…

### LINKED ISSUE #9277: Block List needs single item mode, like Nested Content
A feature I am missing from Nested Content, when working with Block List is the single item mode described here: https://our.umbraco.com/documentation/getting-started/backoffice/property-editors/built-in-property-editors/Nested-Content/#single-item-mode

Can we get that in Block Lists too? Not sure if it should work the same (set min and max to 1), or there should be a toggle for this?

When setting a block list to min. 1 item and max 1 item the block list is empty by default, and you need to manually add your block. Also, you can add more than 1 - you just can't save.

## Umbraco version

I am seeing this issue on Umbraco version: 8.7.0 - 8.8.0

nielslyngsoe:
This PR skips showing the Block Catalogue if only one Block Type is available.
To enable users to paste content it was required to enable users to enter the clipboard dialog directly. This has been solved by adding a button specifcly to open the clipboard.

Reasoning can be found in this issue thread:

Fixed #9277

The proposed look is:
image

Original idea, but is a bit messy as it's adding another element to the composition.
image

Test notes:

The following yet should be performed with one block type available bandwidth multiple types available.

creating blocks with the "add content" button (wording: "add –BlockName–" when only one block is available)
create block via the inline create button
paste one or more blocks
create multiple entries by holding down CTRL(win) or CMD(osx) while clicking create(when one) or clicking the type(when multiple)


nielslyngsoe:
Still considering the level of attention/information this button needs, reference screenshot:
image

nielslyngsoe:
Latest UI suggestion is:
image

madsrasmussen:
Testet and everything works as described. 🥳

Issue #9321, opened 2020-10-30
Indicate fallback language to the browser when used from Model.Value


RachBreeze:
Hello 

For accessibility purposes, it's important that the correct language is defined on the page see https://www.youtube.com/watch?v=QwOoU8T24UY&feature=youtu.be for a demo.

When working on multilingual sites it's possible to define a fallback language, so if a page doesn't have content in one language, content is retrieved in another.  https://our.umbraco.com/Documentation/Getting-Started/Design/Rendering-Content/#using-fall-back-methods documents how to use fallback methods.  

However when a fall back is used the mark up should indicate this is the case 
https://www.w3.org/International/questions/qa-html-language-declarations#:~:text=Quick%20answer,an%20element%20surrounding%20that%20content. 

I _think_ the solution is to use a style similar `@model.ValueFor` and wrap the value returned in a `span`, when a fallback language is used.  I would love to help resolve the issue but recognise there are some things that need bottoming out first!

nul800sebastiaan:
@RachBreeze  We don't want to mess with the HTML output as it is, we're in fact pretty proud that Umbraco usually only outputs exactly what you ask for. I would suggest that we can add an additional optional parameter to `Model.Value` that will surround the output with something like `<span lang="zh-Hans">`, but only if a fallback language is picked of course. 

Make sense?

umbrabot:
Hi @RachBreeze,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

RachBreeze:
PR #9576 has been created to resolve this issue

PR #9576, opened 2020-12-17
Accessibility improvements for Extensions in Razor Markup

RachBreeze:
This PR is to resolve https://github.com/umbraco/Umbraco-CMS/issues/9321

It is important for accessibility purposes that the correct natural language is defined on the page via the `<HTML lang="">` element this allows screen readers and other assistive technology use the correct pronunciation when reading the page content (as shown here) https://www.youtube.com/watch?v=QwOoU8T24UY&feature=youtu.be 

Where a property on a page hasn't been defined in the user's selected language Umbraco allows the content to fall back as detailed here https://our.umbraco.com/Documentation/Getting-Started/Design/Rendering-Content/#using-fall-back-methods this means that the language defined in the HTML element is potentially not the language of the content in that property.

W3C details more about this issue here https://www.w3.org/International/questions/qa-html-language-declarations#:~:text=Quick%20answer,an%20element%20surrounding%20that%20content. This PR aims to both support the accessibility requirement following the pattern defined in thisarticle in the section "What if there's no element to hang your attribute on?"  It also gives the developer the facility to use their discretion as to whether the code should display `<span lang="fallbackLanguage">`...`</span>` around the property.

I have added  "`Fallback.DisplayFallBackLanguage`" as an optional add an additional optional parameter to `Fallback.To` to support this.

[The need for this is based on the EU funded research project for improving the process of creating accessible content by authors](https://accessibilitycluster.com/about-the-project)


src/Umbraco.Core/Models/PublishedContent/Fallback.cs
```diff
@@ -71,5 +71,11 @@ IEnumerator IEnumerable.GetEnumerator()
         {
             return GetEnumerator();
         }
+
+        public const int DisplayFallBackLanguage = 4;
+        /// <summary>
+        /// Gets the fallback to tree ancestors policy.
+        /// </summary>
+        public static Fallback ToDisplayFallBackLanguage => new Fallback(new[] { DisplayFallBackLanguage });
```
glombek, line 79: To keep this consistent it might be useful to rename this `ToDisplayFallbackLanguage` (`Fallback` rather than `FallBack`)

src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs
```diff
@@ -35,6 +36,8 @@ public bool TryGetValue<T>(IPublishedProperty property, string culture, string s
         {
             _variationContextAccessor.ContextualizeVariation(property.PropertyType.Variations, ref culture, ref segment);
 
+            var includeFallBackLanguage = fallback.Contains(Fallback.DisplayFallBackLanguage);
```
glombek, line 39: To keep this consistent it might be useful to rename both of these to have `Fallback` rather than `FallBack`.

src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs
```diff
@@ -155,9 +166,10 @@ private NotSupportedException NotSupportedFallbackMethod(int fallback, string le
         // tries to get a value, recursing the tree
         // because we recurse, content may not even have the a property with the specified alias (but only some ancestor)
         // in case no value was found, noValueProperty contains the first property that was found (which does not have a value)
-        private bool TryGetValueWithAncestorsFallback<T>(IPublishedContent content, string alias, string culture, string segment, out T value, ref IPublishedProperty noValueProperty)
+        private bool TryGetValueWithAncestorsFallback<T>(IPublishedContent content, string alias, string culture, string segment, out T value, ref IPublishedProperty noValueProperty, bool includeFallBackLanguage)
```
glombek, line 169: To keep this consistent it might be useful to rename this to contain `Fallback` rather than `FallBack`.

src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs
```diff
@@ -191,7 +207,7 @@ private bool TryGetValueWithAncestorsFallback<T>(IPublishedContent content, stri
         }
 
         // tries to get a value, falling back onto other languages
-        private bool TryGetValueWithLanguageFallback<T>(IPublishedProperty property, string culture, string segment, out T value)
+        private bool TryGetValueWithLanguageFallback<T>(IPublishedProperty property, string culture, string segment, out T value, bool includeFallBackLanguage)
```
glombek, line 210: To keep this consistent it might be useful to rename this to contain `Fallback` rather than `FallBack`.

src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs
```diff
@@ -225,7 +245,7 @@ private bool TryGetValueWithLanguageFallback<T>(IPublishedProperty property, str
         }
 
         // tries to get a value, falling back onto other languages
-        private bool TryGetValueWithLanguageFallback<T>(IPublishedElement content, string alias, string culture, string segment, out T value)
+        private bool TryGetValueWithLanguageFallback<T>(IPublishedElement content, string alias, string culture, string segment, out T value, bool includeFallBackLanguage)
```
glombek, line 248: To keep this consistent it might be useful to rename this to contain `Fallback` rather than `FallBack`.

src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs
```diff
@@ -259,7 +283,7 @@ private bool TryGetValueWithLanguageFallback<T>(IPublishedElement content, strin
         }
 
         // tries to get a value, falling back onto other languages
-        private bool TryGetValueWithLanguageFallback<T>(IPublishedContent content, string alias, string culture, string segment, out T value)
+        private bool TryGetValueWithLanguageFallback<T>(IPublishedContent content, string alias, string culture, string segment, out T value, bool includeFallBackLanguage)
```
glombek, line 286: To keep this consistent it might be useful to rename this to contain `Fallback` rather than `FallBack`.

glombek (changes requested):
Hi @RachBreeze,

Thanks for this. It works for me (mostly!)

However, I've found a few things that don't quite work seamlessly.

If I have a rich text editor property and then use ModelsBuilder to obtain the value (which will be an `IHtmlString`), the `GetMarkUpForFallbackLanguage` method fails, as it relies on the type parameter being a string (or being able to be implicitly cast to from a string), which `IHtmlString` is not. This will also likely fail for any other property types (ie `int`).

> System.InvalidCastException: 'Invalid cast from 'System.String' to 'System.Web.IHtmlString'.'

Secondly, if the type _is_ a string. the result will need to be output wrapped in an `Html.Raw` statement to escape the markup. This isn't ideal UX-wise but also poses a security risk as the initial text is not HTML escaped. Unfortunately, it may be suboptimal to simply escape the initial text as this may not be desired by the developers.

Boy, this is a bit of a mine-field, eh?!

I wonder if it might be worth making this a completely different method rather than trying to force it into the `Value` / `ValueFor` implementation? I agree something like this is necessary (the workaround at the moment is painful!) but I'm not sure this is the optimal solution either.

Happy to talk through this some time (it may also be worth us chatting to HQ to discuss options here).

Joe

src/Umbraco.Core/Models/PublishedContent/Fallback.cs
RachBreeze: Good spot I've done this now.

src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs
RachBreeze: Good spot I've done this now.

src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs
RachBreeze: Good spot I've done this now.

src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs
RachBreeze: Good spot I've done this now.

src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs
RachBreeze: Good spot I've done this now.

src/Umbraco.Web/Models/PublishedContent/PublishedValueFallback.cs
RachBreeze: Good spot I've done this now.

RachBreeze:
Hi @glombek 
Thank you for taking time to review.  

The return of the <span> needs to be part of the fallback options, as it's only the request to use a fall back that means the language could change, and the user would need to be notified of those change. 

I would need guidance on any new implementations for this. But am happy to do any required changes.  It's worth noting that @nul800sebastiaan did advise on the original implementation https://github.com/umbraco/Umbraco-CMS/issues/9321.

I did realise that the code would need wrapping in `Html.Raw` but this is how content in the RTE has to be rendered.  As you say however, any new implementation for fallback could consider that too.

I have now added type testing to `GetMarkUpForFallbackLanguage` this allows us to determine if the span needs adding, for example just because the user has fallen back to a different value of an `int`, the pronunciation of that `int `value will not change.  The only time language needs to change is if the requested value contains text.  This allowed me to resolve the `IHtmlString` issue too.

RachBreeze:
Heya I've found another demonstration of why this change is important https://vimeo.com/210258413

glombek:
Hi @RachBreeze, I tested your changes today and found a couple of errors. I hope you don't mind but I took the liberty of fixing them for you as they were pretty minor.

I think you've nailed the functionality now for different property types and the fallback of simply returning the initial value instead of falling over is much neater.

I also fixed the breaking test and merged the latest version of `v8/contrib` in to resolve conflicts.

You'll notice in my changes that I didn't use the `is` type checking for IHtmlString as because it's an interface we'd need to do some costly reflection to return the output, so it's a little bit of a compromise but works well.

If you're happy with my changes we can progress.

Thanks!

glombek (approved):

RachBreeze:
Hi @glombek the changes look good thank you

RachBreeze:
Good morning
I hope you're well, is there anything else I can help with for progressing this PR?

glombek:
Thanks @RachBreeze - no further action needed on your part, just me you were waiting for!

That's merged now and you can expect to see it in 8.16.

RachBreeze:
Thank you

nul800sebastiaan:
Hi @RachBreeze and @glombek - unfortunately based on some feedback from the team I have reverted this change for the 8.16 release, I'll quote some of the conversation here:

> The main issue is that it would need to change the returned value from a simple string (that's automatically HTML encoded) into an HTML string, so the added `<span/<div lang=` is rendered correctly, but then 
> 
> 1. the actual value isn't HTML encoded 
> 2. this is only done when the fallback is used, so when it's not, the actual string value is returned (which gets automatically HTML encoded), but otherwise you'd want to skip HTML encoding (otherwise the span/div would end up displayed as text)
> 
> If the value is a simple string, it can conditionally return HTML (for the added span/div, if the fallback to a different language is used) and you don't know when that's the case... So you'd probably end up always rendering it as raw HTML, opening up to XSS issues.
> Meaning the HTML can only reliably be added when the property return type is already `IHtmlString` (but even then it's a bit strange it's adding a div/block level element).
> 
> A better and more generic way would be to expose the used fallback and it's 'options', e.g. the returned culture for the `ToLanguage` fallback, the used content node for `ToAncestors` or a boolean indicating if the returned value is from the `defaultValue` parameter for `ToDefaultValue`. 

I hope this will help with a re-take of this PR for a future version @RachBreeze but please do ask if you need clarification.

Sorry I had to revert this one after it was merged so long ago, but I hope you'll understand we want to do the right thing here. We totally love the idea of this feature of course and we want it, but in a different way. 😅 Thanks so much for your efforts so far and again, let us know if you need some more guidance!

PR #9584, opened 2020-12-20
NetCore: Code clean-up (Umbraco.Tests.Common)

AndyButland:
In this PR I've gone through and applied fixes to all the linting warnings in the Umbraco.Tests.Common project (one we've created as part of the .NET Core migration, and hence on already migrated code that shouldn't be affected by updates that need to come over from V8).


bergmania:
Thanks @AndyButland

PR #9586, opened 2020-12-21
Netcore: Remove NotificationHandler lifetime hack

rustybox:
Off the back of https://github.com/umbraco/Umbraco-CMS/pull/9556#discussion_r546456799

@JimBobSquarePants happier with something like this?


src/Umbraco.Infrastructure/Runtime/AppPluginsManifestWatcher.cs
```diff
@@ -20,45 +20,48 @@ public sealed class ManifestWatcher :
...
 
             var appPlugins = _hostingEnvironment.MapPathContentRoot(Constants.SystemDirectories.AppPlugins);
             if (Directory.Exists(appPlugins) == false)
             {
-                return Task.CompletedTask;
+                return;
             }
 
-            _mw = new Core.Manifest.ManifestWatcher(_loggerFactory.CreateLogger<Core.Manifest.ManifestWatcher>(), _umbracoApplicationLifetime);
+            _mw = new ManifestWatcher(_loggerFactory.CreateLogger<ManifestWatcher>(), _umbracoApplicationLifetime);
```
JimBobSquarePants, line 46: Should there be more than one `ManifestWatcher` running at a time? If not' why not inject that directly?

Incidentally, that class needs cleaning up. The `DisposableObjectSlim` class it inherits is a mess since it explicitly states it is for objects with no unmanaged objects then goes on to call the `GC.SuppressFinalize(this)` method for a non-existing finalizer.

rustybox: 🤷

rustybox: In terms of handling singleton object + transient event, better?

JimBobSquarePants: Haha.... Looking at the code, I would say no. The `ManifestWatcher` should be injected as a singleton and your `INotificationHandler` simply adds the AppPlugins folder to that watcher via `Start` (_poorly named if the methods can be called more that once_). It appears to do things like check the directory/file exists already.

src/Umbraco.Infrastructure/Runtime/AppPluginsManifestWatcherNotificationHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public AppPluginsManifestWatcherNotificationHandler(ManifestWatcher manifestWatcher, IHostingEnvironment hostingEnvironment)
+        {
+            _manifestWatcher = manifestWatcher ?? throw new ArgumentNullException(nameof(manifestWatcher));
+            _hostingEnvironment = hostingEnvironment ?? throw new ArgumentNullException(nameof(hostingEnvironment));
+        }
+
+        public Task HandleAsync(UmbracoApplicationStarting notification, CancellationToken cancellationToken)
+        {
+            if (_hostingEnvironment.IsDebugMode == false)
```
JimBobSquarePants, line 28: `If(!_hostingEnvironment.IsDebugMode)`. Your IDE should be warning you that the condition can be simplified.

src/Umbraco.Infrastructure/Runtime/AppPluginsManifestWatcherNotificationHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                return Task.CompletedTask;
+            }
+
+            var appPlugins = _hostingEnvironment.MapPathContentRoot(Constants.SystemDirectories.AppPlugins);
+            if (Directory.Exists(appPlugins) == false)
```
JimBobSquarePants, line 34: No need for the check. ManifestWatcher does this already via `IsWatchable(string folder)`

rustybox: I'm just trying to remove composers / components not fix every issue with the codebase

rustybox: resolved b6015331

src/Umbraco.Infrastructure/Runtime/AppPluginsManifestWatcherNotificationHandler.cs
rustybox: It's not, I wonder if someone disabled that check because that shit is everywhere in the codebase

JimBobSquarePants (approved):
LGTM.... Might be worth moving `ManifestWatcher` to the infrastructure project in the future if it is being registered against the service collection there. (_Or keep it where it is and register it in UmbracoBuilder directly_)

src/Umbraco.Infrastructure/Runtime/AppPluginsManifestWatcherNotificationHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+        public Task HandleAsync(UmbracoApplicationStarting notification, CancellationToken cancellationToken)
+        {
+            if (!_hostingEnvironment.IsDebugMode)
+            {
+                return Task.CompletedTask;
+            }
+
+            var appPlugins = _hostingEnvironment.MapPathContentRoot(Constants.SystemDirectories.AppPlugins);
+
+            if (!Directory.Exists(appPlugins))
```
JimBobSquarePants, line 35: I don't understand why this is still here. The ManifestWatcher already does checks. 

https://github.com/umbraco/Umbraco-CMS/pull/9586/files#diff-fc7600f9983710ac030090cf5fe72eb528e0bc84c9fde3ba8eafe64426f46805R27

rustybox: Check the test failures, Directory.GetDirectories throws if we don’t check

JimBobSquarePants: The whole method could be 

``` c#
if (!_hostingEnvironment.IsDebugMode)
{
    string appPlugins = _hostingEnvironment.MapPathContentRoot(Constants.SystemDirectories.AppPlugins);
    _manifestWatcher.WatchDirectories(Directory.GetDirectories(appPlugins));
}

return Task.CompletedTask;
```

JimBobSquarePants: The fix should be in the ManifestWatcher then.

JimBobSquarePants:
Got an unrelated issue on Mac. Something to do with TypeFinder

https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=53889&view=logs&j=2aca8870-17b7-5f44-f027-fef361f83dd7&t=a71515d3-d220-5ea5-179d-e804fd94ac96&l=25931

src/Umbraco.Infrastructure/Runtime/AppPluginsManifestWatcherNotificationHandler.cs
rustybox: Manifest watcher should what take the parent path and add watchers for child directories?

JimBobSquarePants: This would work in `ManifestWatcher` called via 

```c#
foreach (var packageFolder in ParseFolders(packageFolders))...
```

```c#
private static IEnumerable<string> ParseFolders(params string[] packageFolders)
{
    var valid = new List<string>();
    foreach (string folder in packageFolders)
    {
        if (File.Exists(Path.Combine(folder, "package.manifest")))
        {
            valid.Add(folder);
        }
        else if (Directory.Exists(folder))
        {
            valid.AddRange(Directory.GetDirectories(folder).Where(x => File.Exists(Path.Combine(x, "package.manifest"))));
        }
    }

    return valid;
}
```

JimBobSquarePants: Double check my logic (_I haven't slept nearly enough_) but that should allow checking a folder and any children.

rustybox: The exception is thrown here in the handler

```
    _manifestWatcher.WatchDirectories(Directory.GetDirectories(appPlugins));
```

rustybox:
> Got an unrelated issue on Mac. Something to do with TypeFinder

Guess fragile tests, passes now

src/Umbraco.Infrastructure/Runtime/AppPluginsManifestWatcherNotificationHandler.cs
JimBobSquarePants: You would change that line to 
```c#
_manifestWatcher.WatchDirectories(appPlugins);
```

rustybox: That would of course work.

It's a change to the interface of ManifestWatcher and outside of the scope of what I'm looking to achieve which is simply remove components and composers.

JimBobSquarePants: Changing the method name already breaks the contract. You haven’t made a logic change, at worst it would mean a duplicate directory check.

src/Umbraco.Infrastructure/Runtime/AppPluginsManifestWatcher.cs
Shazwazza: `DisposableObjectSlim` is just following MS practices. GC.SupressFinalize doesn't do anything if there's no finalizer, see https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose?redirectedfrom=MSDN#the-dispose-method ... but sure that could be removed. In any case, this should still use the correct IDisposable pattern and not just implement Dispose without checking if it's already been disposed. 

Using VS when ctrl + . and choosing 'implement interface with IDispoable pattern' will also add this by default:

```cs
protected virtual void Dispose(bool disposing)
{
    if (!_disposedValue)
    {
        if (disposing)
        {
            // TODO: dispose managed state (managed objects)
        }

        // TODO: free unmanaged resources (unmanaged objects) and override finalizer
        // TODO: set large fields to null
        _disposedValue = true;
    }
}

// // TODO: override finalizer only if 'Dispose(bool disposing)' has code to free unmanaged resources
// ~ModelsBuilderComponent()
// {
//     // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
//     Dispose(disposing: false);
// }

public void Dispose()
{
    // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
    Dispose(disposing: true);
    GC.SuppressFinalize(this);
}
```

JimBobSquarePants: `ManifestWatcher` should be sealed and contains no unmanaged types so the call (_which isn't free_) is redundant. It's included in the pattern to ensure that it's not overlooked in cases when it is required. 

IMO classes like `DisposableSlim` are dangerous. If you need to manage the lifetime of objects within your object you should do so within that object only.

JimBobSquarePants: Incidentally here's probably the most in-depth article on the subject. Peer reviewed by several heavy-hitters at Microsoft.

http://joeduffyblog.com/2005/04/08/dg-update-dispose-finalization-and-resource-management/

Shazwazza: If kept, `DisposableSlim` should be updated to use the standard conventions/practices and naming. 

> so the call (which isn't free) is redundant

I agree, can be removed

Out of interest, here's the benchmark. It's not free but I don't think anyone is going to notice since its not critical code :)

``` ini

BenchmarkDotNet=v0.11.3, OS=Windows 10.0.18363
Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
  [Host]     : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.8.4250.0
  DefaultJob : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.8.4250.0


```
|                  Method |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------------ |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
|    WithSuppressFinalize | 4.913 ns | 0.1345 ns | 0.1258 ns |      0.0038 |           - |           - |                12 B |
| WithoutSuppressFinalize | 2.304 ns | 0.0649 ns | 0.0575 ns |      0.0038 |           - |           - |                12 B |


> IMO classes like DisposableSlim are dangerous

Sure, I'm not wedded to this class and happy to have it removed and implement the pattern manually wherever necessary. Just want to mention the trade-off which is that anywhere IDisposable is needed, the pattern will typically be generated from VS/Ryder snippets which means there will be plenty of people leaving that call alone since VS's comments on that code block say `Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method`

`DisposableSlim` is only for simple non-finalizer requirements and should be fine even given all the dos/donts of that article since the complexity and challenges arise from when finalization is required.

If anyone feels strongly about the removal of the class, feel free to go for it

rustybox: By default resharper doesn't bother with Dispose(bool) unless the class has a finalizer or you click a checkbox for subclass/unmanaged resources.

Shazwazza: Same with VS, there's the option to `implement interface with IDisposable pattern` or just implement the interface. For now I was just fixing up DisposableObjectSlim for the interim but noticed this when we remove suppress finalize 

image

So we either have to disable that warning in code or add config to suppress this rule.

Shazwazza (approved):
Changes seem good to me happy to merge.

src/Umbraco.Infrastructure/Runtime/AppPluginsManifestWatcher.cs
JimBobSquarePants: Or delete `DisposableObjectSlim`

Shazwazza: What I'm saying though is that anywhere IDisposable is implemented, our current analysis rules will bark at us for not using GC.SuppressFinalize.

Issue #9396, opened 2020-11-14
Netcore: Split monolith test projects


p-m-j:
The Umbraco.Tests.UnitTests project tests 8 libraries, the tests are grouped by namespace so there is an Umbraco.Web.Common folder which has all the tests for Umbraco.Web.Common.

It would be much nicer if there were more test projects with less broad a focus.

This applies to both unit/integration tests.

Goals
1. Create more tests projects e.g. Umbraco.Web.Common.Tests.UnitTests
1. Create these test projects under tests/ instead of src/
1. Autodiscover test projects in the CI build such that adding a new test project to tests/ results in tests being run.

@JimBobSquarePants did I miss anything here?

rustybox:
Bonus points for yaml build definition so community have visibility of the CI build?

rustybox:
At present this requires the assistance from staff with permissions to edit the build pipeline on azure devops as the dotnet test step is explicit about which projects it runs.

Apparently one can now just run dotnet test against the sln file, so no need to manually discover csproj under tests/ anymore.

JimBobSquarePants:
You can run `dotnet test` against [a folder](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test#arguments), doesn't even need to be a solution file.  

Put yes, I would advocate splitting up tests. Current `InternalsVisibleTo` rules would need to be updated but that's trivial and can be configured in single `Directory.Build.props/targets` files.

Those yaml files should definitely be public.

rustybox:
It's classic builds atm as opposed to yaml as far as I can tell.

rustybox:
Went to check, you can pass a path to a folder containing a sln or a csproj, you can't just pass a path to a folder with multiple test project folders underneath there's no recursion.

bergmania:
We would like the build to be YAML, but does not have time to look into it yet. 

Some things to note:
- We would like the unit tests to be executed on both a Windows agent and a Linux agent
- Integration tests should only run on Windows (Due to the LocalDB dependency)

Regarding the test projects, I'm pretty sure you can use `*` as wildcard, so it could be something like `*.UnitTests.csproj` for unit tests and `*.IntegrationTests.csproj`

rustybox:
not on a path no

```bash
dotnet test tests\**\*.csproj
```

This isn't a thing as much as i'd like it to be.

```bash
dotnet test umbraco.sln 
```

Does everything we need though.

bergmania:
[...]

PR #9589, opened 2020-12-22
Some IPublishedSnapshotService (etc) cleanup based on TODOs

Shazwazza:
Mostly based on TODOs from #9530 

* GetStatus, StatusUrl moved
* Collect should be CollectAsync
* Remove IIOHelper dependency
* PublishedSnapshotServiceBase shouldn't exist
* Remove NPocoRepositoryBase
* Remove NuCacheComposer, replace with IUmbracoBuilder extension
* IsFrontEndRequest can be ext method
* Renames 
  * AddBackOffice -> AddBackOfficeAuthentication
  * AddAllBackOfficeComponents -> AddBackOffice
  * AddUmbracoWebsite -> AddWebsite
* ModelBindingExceptionAttribute instead of a ModelBindingExceptionFilter with a TypeFilter attribute
* DependencyInjection namespaces in Umbraco.Web.Common, Umbraco.Web.BackOffice and Umbraco.Web.Website projects for DI related ext on IUmbracoBuilder and IServiceCollection - ru-shuffles DI code into those namespaces
* Changes DisposableObjectSlim to adhere to default IDipsosable implementation patter and without GC.SuppressFinalize (This whole class could/should be removed, can be done later if anyone wants to, will require approx 13 manual IDisposable implementations)
* Moves globally configuring MVC to options configuration via UmbracoMvcConfigureOptions instead of inline in the call to add mvc + views.
* Linting cleanup


src/Umbraco.Core/HealthCheck/Checks/Permissions/FolderAndFilePermissionsCheck.cs
```diff
@@ -32,22 +32,16 @@ public FolderAndFilePermissionsCheck(ILocalizedTextService textService, IOptions
...
         /// <summary>
         /// Executes the action and returns it's status
         /// </summary>
-        /// <param name="action"></param>
-        /// <returns></returns>
-        public override HealthCheckStatus ExecuteAction(HealthCheckAction action)
-        {
-            throw new InvalidOperationException("FolderAndFilePermissionsCheck has no executable actions");
-        }
+        public override HealthCheckStatus ExecuteAction(HealthCheckAction action) => throw new InvalidOperationException("FolderAndFilePermissionsCheck has no executable actions");
```
JimBobSquarePants, line 44: `NotSupportedException`. 

Or maybe an interface redesign if there are implementations that cannot support the features.

src/Umbraco.PublishedCache.NuCache/PublishedSnapshotStatus.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            var cg = contentStore.GenCount;
+            var mg = mediaStore.GenCount;
+            var cs = contentStore.SnapCount;
+            var ms = mediaStore.SnapCount;
+            var ce = contentStore.Count;
+            var me = mediaStore.Count;
+
+            return
+                " Database cache is " + (dbCacheIsOk ? "ok" : "NOT ok (rebuild?)") + "." +
```
JimBobSquarePants, line 42: I would have defined the variables first and returned an interpolated string so this was readable by human beings.

src/Umbraco.Web.Common/Extensions/UmbracoBuilderExtensions.cs
```diff
@@ -34,15 +36,20 @@
...
 using Umbraco.Web.Common.AspNetCore;
+using Umbraco.Web.Common.Extensions;
 using Umbraco.Web.Common.Filters;
 using Umbraco.Web.Common.ModelBinders;
 using Umbraco.Web.Common.Profiler;
 using Umbraco.Web.Telemetry;
 using IHostingEnvironment = Umbraco.Core.Hosting.IHostingEnvironment;
 
-namespace Umbraco.Core.DependencyInjection
+namespace Umbraco.Web.Common.Extensions
```
JimBobSquarePants, line 46: `Umbraco.Web.DependencyInjection`?

I would also move it to a folder with the same name.

src/Umbraco.Web.Common/Extensions/UmbracoBuilderExtensions.cs
```diff
@@ -135,20 +142,20 @@ public static IUmbracoBuilder AddUmbracoCore(this IUmbracoBuilder builder)
             }
 
                 );
-            builder.Services.AddUnique<IAppPolicyCache>(factory => factory.GetRequiredService<AppCaches>().RuntimeCache);
-            builder.Services.AddUnique<IRequestCache>(factory => factory.GetRequiredService<AppCaches>().RequestCache);
+            builder.Services.AddUnique(factory => factory.GetRequiredService<AppCaches>().RuntimeCache);
```
JimBobSquarePants, line 145: `AddUnique(factory)` is on my kill list. Horrible method which uses the service provider unnecessarily.

src/Umbraco.Web.Website/Extensions/WebsiteUmbracoBuilderExtensions.cs
```diff
@@ -11,12 +12,12 @@ namespace Umbraco.Extensions
     /// <summary>
     /// <see cref="IUmbracoBuilder"/> extensions for umbraco front-end website
     /// </summary>
-    public static class UmbracoBuilderExtensions
+    public static class WebsiteUmbracoBuilderExtensions
```
JimBobSquarePants, line 15: This should live with other DI code.

src/Umbraco.Core/HealthCheck/Checks/Permissions/FolderAndFilePermissionsCheck.cs
Shazwazza: Yep the plan is to change healthchecks, there will be not 'action' to execute in netcore. @bergmania has allocated tasks for this

src/Umbraco.PublishedCache.NuCache/PublishedSnapshotStatus.cs
Shazwazza: Human being are overrated 👽 Yep I agree, this is just migrated code but I can update

src/Umbraco.Web.Website/Extensions/WebsiteUmbracoBuilderExtensions.cs
Shazwazza: Sounds good - how do you recommend we organize this best then? 

Update: ah, I see your comment here https://github.com/umbraco/Umbraco-CMS/pull/9589/files#r547360727 = `Umbraco.Web.DependencyInjection` 

With the snapshot of code in this PR we have:

* Project Umbraco.Web.Common - `Umbraco.Web.Common.Extensions.UmbracoBuilderExtensions`
* Project Umbraco.Web.Website - `Umbraco.Extensions.WebsiteUmbracoBuilderExtensions`
* Project Umbraco.Web.BackOffice- `Umbraco.Extensions.BackOfficeUmbracoBuilderExtensions`

Just looking at where other DI code exists in these projects and its still in various places while we are cleaning things up. Just noticed we still have `Umbraco.Web.Common.Runtime.AspNetCoreComposer` which can easily be removed and transitioned to a builder extension. 

So should the end result be this?

* Project Umbraco.Web.Common - `Umbraco.Web.DependencyInjection.UmbracoBuilderExtensions` - `UmbracoBuilderExtensions_Common.cs`
* Project Umbraco.Web.Website - `Umbraco.Web.DependencyInjection.UmbracoBuilderExtensions` - `UmbracoBuilderExtensions_Website.cs`
* Project Umbraco.Web.BackOffice- `Umbraco.Web.DependencyInjection.UmbracoBuilderExtensions` - `UmbracoBuilderExtensions_BackOffice.cs`

JimBobSquarePants: Looking at the file contents there is actually a lot less code than I would have anticipated. When I suggested the partial classes I thought there would be a lot more.

Normally I would say namespace to match your binary and location but the namespaces of those three projects are all over the place - and, because the name namespace is spread across the projects you cannot use a single partial. 

I won't lie.... I do not understand the hierarchy.

- `Umbraco.Web.BackOffice` references `Umbraco.Web.Common`, and, `Umbraco.Examine.Lucene`
- `Umbraco.Web.Common` references `Umbraco.Core` and `Umbraco.Infrastructure`
- `Umbraco.Web` references `Umbraco.Core` , `Umbraco.Infrastructure`, and `Umbraco.Persistance.SqlCe`

This means that there is nothing common about the `Umbraco.Web.Common` project since it is not share by both the `Umbraco.Web.xxx` projects!

Personally I would fix up the project names and namespaces (_since they're not even consistent within the projects own classes_) and do this:

`Umbraco.{NAMESPACE}.DependencyInjection.UmbracoBuilderExtensions` with the standard filename.

Shazwazza: Yep namespaces are in chaos currently. There's some decisions to be made about the organization and then a big cleanup. This has been avoided currently out of merging pains but will need to happen sooner than later for sure, driving me (and probably everyone else) mad! 

 `Umbraco.Web` - is the old net framework project, it will die

The front-end one is `Umbraco.Web.Website` which does reference common.

The reason that `Umbraco.Extensions` exists between projects as a global namespace is to make `using` requirements for end users very simple.

For now though I will use the `Umbraco.{NAMESPACE}.DependencyInjection.UmbracoBuilderExtensions` for UmbracoBuilder extensions and shuffle a bit more code to align with that namespace.

JimBobSquarePants: >The reason that Umbraco.Extensions exists between projects as a global namespace is to make using requirements for end users very simple.

With good namespace hierarchy and API docs you can avoid this. 

The namespace fixup has to come soon. It makes design and discovery so much harder currently!

Shazwazza:
NOTE: This is not quire ready for review, still tests to fixup which are unravelling issues with composers, etc... that need to be moved to ext methods instead so we know what's actually going on (no magic), then turns out we have DI code all over the place in ext method classes so cleaning that up a bit, then some of our ext methods are doing way too much and need to be split out a bit. Will be ready tomorrow.

src/Umbraco.Web.Website/Extensions/WebsiteUmbracoBuilderExtensions.cs
rustybox: Namespace pain put a stop to #9396

It was hinted that both Umbraco.Core and Umbraco.Infrastructure will both be set to have a shared namespace of Umbraco.Cms which kindof confuses me, any ideas what that's about?

Shazwazza: The plan is to make the base namespace Umbraco.Cms along with the nuget package name then streamline the other umbraco products like UmbracoDeploy to be Umbraco.Deploy namespace + package name, same goes with UmbracoForms to Umbraco.Forms. 

Not 100% sure about the Core vs Infra projects sharing the base namespace. I presume the concept is that Umbraco.Core and Umbraco.Infrastructure projects are sort of 'the same' and there shouldn't be an 'Infrastructure' namespace but these projects are separated due to dependency requirements and hopefully over time more would be moved to the Core project where dependencies are not needed (i.e. Services). We should have a chat about what the intention is early in the NY with @bergmania and the netcore team

bergmania: The original idea with shared namespaces was back from the first RFC. 
Because, as @Shazwazza says, the concept is sort of the same and we don't expect to have everything that should be in Core, located in core in the first release. This would make breaking changes easier to resolve.

rustybox: I can understand there's a lot in Infra namespace that shouldn't be there, but not so much that infra namespace should exist.

FileSystemMediaService / AzureBlobStorageMediaService / S3MediaService would all be examples I'd expect to find in a infra namespace

JimBobSquarePants: The lack of decent namespace organisation is honestly intolerable. It’s hurting discovery and development. 

This weird development-by-part approach is quickly turning into a demonstration of the sunken cost fallacy. The codebase is an utter mess! Accept that and fix it properly rather than shipping a have baked result.

rustybox: lol @JimBobSquarePants you're so often right, but there's that catch more with honey than vinegar thing

PR #9591, opened 2020-12-22
NetCore: Code clean-up (Umbraco.Tests.UnitTests)

AndyButland:
Having some quiet work days leading up to Christmas... so in this PR I've gone through and applied fixes to most of the linting warnings in the Umbraco.Tests.UnitTests project (again one we've created as part of the .NET Core migration, and hence on already migrated code that shouldn't be affected too much by merges that need to come over from V8).


Shazwazza:
Great work @AndyButland ! that's a lot of tedious work :) I've looked at most of the files and all looks good. All tests pass and I'm assuming no actual logic was changed so will merge. 🚀

PR #9595, opened 2020-12-23
Noemivdp patch 1 - Add AllowedChildrenEventManager

noemivdp:
Hi,

I would like more control over the rules of what types of documents can be created at different levels of the Umbraco content tree, for me having only off and on based on type is not enough.

I would like to be able to allow only Admins to create a certain type of page for example - my feeling is  people might have different needs for their site - so thought the ways to achieve this would be like in kind to EditorModel sending events.

In PR I've tried copy this so that there is 'AllowedChildrenEventManager' and an filter applied to GetAllowedChildren api to give choice for developers to handle and change allowed types based on code logic.

An example: 
Admin's no create BlogPosts - yes create a secret adminSettings - under Blog Section:

```
public class RegisterToAllowedChildrenEvent : IUserComposer
    {

        public void Compose(Core.Composing.Composition composition)
        {
            composition.Components().Append<RegisterToAllowedChildrenEventComponent>();
        }

        public class RegisterToAllowedChildrenEventComponent : IComponent
        {
            public void Initialize()
            {
                AllowedChildrenEventManager.SendingAllowedChildrenModel += AllowedChildrenEventManager_SendingAllowedChildrenModel;
            }

            private void AllowedChildrenEventManager_SendingAllowedChildrenModel(System.Web.Http.Filters.HttpActionExecutedContext sender, AllowedChildrenEventArgs<IEnumerable<Models.ContentEditing.ContentTypeBasic>> e)
            {
                var parentNode = e.UmbracoContext.Content.GetById(e.ContentId ?? -1);
                var currentUser = e.UmbracoContext.Security.CurrentUser;
                bool isAdmin = false;
                if (currentUser.IsAdmin())
                {
                    isAdmin = true;
                }
                if (isAdmin && parentNode.ContentType.Alias=="blog")
                {
                    List<Models.ContentEditing.ContentTypeBasic> allowedChildren = e.Model.ToList();
                    var blogPost = allowedChildren.FirstOrDefault(f => f.Alias == "blogpost");
                    allowedChildren.Remove(blogPost);
                    ContentTypeBasic adminOnlyType = new ContentTypeBasic()
                    {
                        Alias = "adminSettings",
                        Icon = "icon-settings",
                        Name = "Admin Only Settings"

                    };
                    allowedChildren.Add(adminOnlyType);
                    e.Model = allowedChildren;
                }
            }

            // terminate: runs once when Umbraco stops
            public void Terminate()
            {
                AllowedChildrenEventManager.SendingAllowedChildrenModel -= AllowedChildrenEventManager_SendingAllowedChildrenModel;
            }          
        }
    }
```
adminSettings need not be allowed under blog section.

### Prerequisites

### DIFF
--- a/src/Umbraco.Web/Editors/AllowedChildrenEventArgs.cs
+++ b/src/Umbraco.Web/Editors/AllowedChildrenEventArgs.cs
@@ -1,0 +1,55 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection.Metadata;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Umbraco.Web.Editors
+{
+    public sealed class AllowedChildrenEventArgs<T> : AllowedChildrenEventArgs
+    {
+        private readonly AllowedChildrenEventArgs _baseArgs;
+        private T _model;
+
+        public AllowedChildrenEventArgs(AllowedChildrenEventArgs baseArgs)
+            : base(baseArgs.Model, baseArgs.UmbracoContext, baseArgs.ContentId)
+        {
+            _baseArgs = baseArgs;
+            Model = (T)baseArgs.Model;
+            ContentId = baseArgs.ContentId;
+        }
+
+        public AllowedChildrenEventArgs(T model, UmbracoContext umbracoContext, int? contentId)
+            : base(model, umbracoContext,contentId)
+        {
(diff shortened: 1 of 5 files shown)


noemivdp:
Hi do I need to raise issue? or is this OK? or need more information - sorry if I'm doing it wrong

PR #9597, opened 2020-12-23
Added missing alt attributes to the images on the login screen and in avatars

RachBreeze:
This PR adds the missing alt attribute to the image on the login screen, this is decorative and as such the alt text is empty.
The PR also adds the missing alt text to the image in the avatar


mikecp:
Hello @RachBreeze ,
Thanks for your PR! 
If I may suggest a small improvement, it could be nice if the avatar alt text was moved into a translation/localized label. Would that be a possible for you to adapt? Thanks :-)

RachBreeze:
Hi @mikecp 
Thank you for reviewing and please accept my apologies for the delay, I have done this now

mikecp:
Hi @RachBreeze ,
Thanks for the update and no problem for the delay :-)

I tried to run this locally and I had an issue: the name was displayed with and ending />, and when looking at the page code, the<localize> tag was actually not translated but just displayed as is. I guess this is due to the fact that it was in an "alt" attribute, hence it was not picked by the translation service.

I started making some tests and tried with the attribute translation solution using"@general_avatar", but this did not work either because the user name needed to be appended to it. I ended up with a solution where the alt text is created in the directive, where it is possible to combine both the translation and the name...

I hope you won't mind those changes, I have pushed them back here for so you can have a look at them, and if it is OK with you then I will merge all.

PS: I also added translations for French and Dutch :-)

RachBreeze:
Hi @mikecp 
The changes look great and thank you for adding the French and Dutch

mikecp:
Hi @RachBreeze ,
Let's merge this then :-)
Thanks! 👍

PR #9598, opened 2020-12-24
More cleanup & reorganizing

Shazwazza:
* moves some services 'up' to Core proj
* moves more core DI registrations 'up' to UmbracoBuilder instead of ext methods
* moves Composing ext to DependencyInjection namespaces as UmbracoBuilder ext
* moves service registrations 'up' where possible - do as much as possible in Core, then Infra, then Web, etc... 
* moves AspNetCoreComposer registrations -> AddWebComponents
* removes AspNetCoreBootFailedComposer (not needed)
* removes SerilogComposer -> part of AddCoreInitialServices
* removes IntegrationTestComposer -> moved to ext method
* removes WebsiteComposer, BackOfficeComposer -> moved to ext methods
* Do not call `AddComposers` as part of integration tests - we don't want to add any services that we aren't explicitly testing and this will implicitly add all sorts of things that we don't want. Core composers will eventually all be removed anyways so this is part of that work
* Fixes `UmbracoIntegrationTest` to use the `TestOptions.Mapper` value
* Removes `RuntimeTests`, this was an old test at the beginning of netcore porting to show that the runtime would boot and composers would initialize but we are showing this with other tests and don't want to maintain 2x different ways to boot an integration test.
* dist cache cleanup
  * no more DistributedCacheBinderComposer
  * no more DatabaseServerRegistrarAndMessengerComposer, replaced with AddDistributedCache
  * only one IServerMessenger instead of multiple inherited interfaces
  * removes IServerRegistrar and now we have IServerRoleAccessor
  * renames methods from "Perform..." to "Queue..." because that is what it does. A very very long time ago the IServerMessenger would send msgs directly to other servers, this is no longer the case and dist cache messages are queued.
  * adds notes to DatabaseServerMessengerNotificationHandler indicating that we need a new task to investigate/cleanup the entire requirement of DatabaseServerMessengerNotificationHandler
* linting cleanup
* Adds explicit nuget feed to Nuget.config to avoid this new strange issue occuring randomly in builds https://developercommunity.visualstudio.com/content/problem/983843/dotnet-build-task-does-not-use-nugetorg-for-one-pr.html 

## Testing

* make sure all tests pass
* make sure back office and front-end (only basic stuff) work


src/Umbraco.Web.Common/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -28,6 +28,9 @@ public static class ApplicationBuilderExtensions
         /// </summary>
         public static IApplicationBuilder UseUmbraco(this IApplicationBuilder app)
         {
+            // TODO: Should we do some checks like this to verify that the corresponding "Add" methods have been called for the
+            // corresponding "Use" methods?
+            // https://github.com/dotnet/aspnetcore/blob/b795ac3546eb3e2f47a01a64feb3020794ca33bb/src/Mvc/Mvc.Core/src/Builder/MvcApplicationBuilderExtensions.cs#L132
```
bergmania, line 33: What would help developers that do not use our main extension methods..

bergmania (approved):
Looks good, and everything seems to work like before 💪

PR #9601, opened 2020-12-29
NetCore: Getting rid of HttpResponseException

elit0451:
Replacing the usage of `HttpResponseException` (legacy) with `ValidationErrorResult`.

There are 2 single cases in **HttpActionContextExtensions** that are commented out for the moment as they are not used at the moment and we need to rethink their implementation. The rest has been handled


src/Umbraco.Web.BackOffice/Controllers/ContentController.cs
```diff
@@ -696,7 +695,7 @@ private async Task<ContentItemDisplay> PostSaveInternal(ContentItemSave contentI
                     // add the model state to the outgoing object and throw a validation message
                     var forDisplay = mapToDisplay(contentItem.PersistedContent);
                     forDisplay.Errors = ModelState.ToErrorDictionary();
-                    throw HttpResponseException.CreateValidationErrorResponse(forDisplay);
+                    return new ValidationErrorResult(forDisplay);
```
bergmania, line 697: You can't just replace this, you will need to handle the return type all the way to the top of the request. As you know, the throw statement skips the rest of the execution, the return don't.

In this case it can be handled by updating the return type of `PostSave` to `Task<ActionResult<ContentItemDisplay>>`

src/Umbraco.Web.BackOffice/Controllers/ContentTypeController.cs
```diff
@@ -248,13 +248,13 @@ public IActionResult GetWhereCompositionIsUsedInContentTypes(GetAvailableComposi
...
         [Authorize(Policy = AuthorizationPolicies.TreeAccessAnyContentOrTypes)]
-        public ContentPropertyDisplay GetPropertyTypeScaffold(int id)
+        public ActionResult<ContentPropertyDisplay> GetPropertyTypeScaffold(int id)
         {
             var dataTypeDiff = _dataTypeService.GetDataType(id);
 
             if (dataTypeDiff == null)
             {
-                throw new HttpResponseException(HttpStatusCode.NotFound);
+                return new ValidationErrorResult(dataTypeDiff, StatusCodes.Status404NotFound);
```
bergmania, line 257: This shoud not be a ValidationErrorResult
```suggestion
                return NotFound();
```

src/Umbraco.Web.BackOffice/Controllers/ContentTypeControllerBase.cs
```diff
@@ -205,7 +205,7 @@ protected IEnumerable<EntityBasic> PerformGetWhereCompositionIsUsedInContentType
                 }
 
                 if (source == null)
-                    throw new HttpResponseException(HttpStatusCode.NotFound);
+                    return new ValidationErrorResult(source, StatusCodes.Status404NotFound);
```
bergmania, line 208:
```suggestion
                    return NotFound();
```

bergmania (changes requested):
Hi.. 

I only added a few examples, but there are some general things that needs to be changes.

ValidationErrorResult should only be used when it is a validation issue. (When HttpResponseException.CreateValidationErrorResponse was used).
In other cases please use the build-in types, like NotFound()

bergmania:
Good work @elit0451..

I had to make some changes, but they are almost all of them related to potential null reference exceptions. When returning an `ActionResult<T>` from the submethods, we need to have the case when we do not return a value (E.g NotFound).

PR #9604, opened 2021-01-03
Reverts button type change in media edit

### LINKED ISSUE #9600: Can not create nor modify media items in back office of Umbraco 8.10.0
When I want to modify (title or image) of media item the save button does nothing 

## Umbraco version

Umbraco 8.10.0

Reproduction
------------

Whether you:

- Go to Media section and try to modify title or image of existing image item.
- Or create a new media item, give it a title and upload an image.

In both cases clicking on the save button doesn't save the media item, without any warning or error message. Navigating away will give the message "Are you sure you want to navigate away from this page? - you have unsaved changes".

### Bug summary

One can not save media items in the back office.

### Specifics

url: /umbraco/#/media/media/edit/<ID>
Umbraco 8.10.0
Tried on Edge, Firefox and Chrome

### Expected result

After clicking on save I would expect modification being saved.

### Actual result

Nothing happens, no error message, nothing changed on screen.

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #9600 

### Description
Per the discussion on #9600, changing the button type broke the click event (since there's no handler defined for the button, and changing from submit means no more default form submit action). This change reverts the button type back to `submit`

This does mean the issue raised in #9019 is also reverted, but we can find a more appropriate way to fix.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/media/edit.html
+++ b/src/Umbraco.Web.UI.Client/src/views/media/edit.html
@@ -59,7 +59,7 @@
 
                     <!-- label-key="buttons_save" -->
                     <umb-button alias="save"
-                                type="button"
+                                type="submit"
                                 button-style="success"
                                 state="page.saveButtonState"
                                 label-key="{{page.submitButtonLabelKey}}"


nul800sebastiaan:
FYI I can't reproduce the behavior of #9019 so I'm guessing something else has changed to fix that problem already.

PR #9606, opened 2021-01-03
Replace jQuery trim() using vanilla JS

bjarnef:
### Prerequisites

### Description
I noticed we have a few references using jQuery trim method `$.trim(string)`, but there is no need to use this in modern browsers, where we can just use vanilla `trim()` method, which already is use most places.
http://youmightnotneedjquery.com/#trim

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/listview/listview.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/listview/listview.controller.js
@@ -687,7 +687,7 @@ function listViewController($scope, $interpolate, $routeParams, $injector, $time
 
             if (e.nameExp) {
                 var newValue = e.nameExp({ value });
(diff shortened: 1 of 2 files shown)


nathanwoulfe (approved):
All good, thanks @bjarnef 

That's a funky line though - aren't these two the same?

````js
// check newValue exists, then reassign after trimming
if (newValue && (newValue = newValue.trim()) {
    value = newValue; // set value to the trimmed newValue
}
// check newValue exists
if (newValue) {
    value = newValue.trim(); // set value to the trimmed newValue
}
````

c9mbundy:
@nathanwoulfe
`&& (newValue = newValue.trim())`
testing the trimmed status to avoid saving an empty string value ?

nathanwoulfe:
Ah yeah, that'll be why 👍

Actually ties back to another issue which was looking at adding any extension to check null or whitespace...

PR #9616, opened 2021-01-05
Netcore: Fixes issues with user invites

bergmania:
### Details
Fixes issues with user invites
- Issue with the generated link in the invite email. The querystring was not added
- Allow anonymous access to CurrentUserController.PostSetInvitedUserPassword, as it is used by users not logged in
- Allow anonymous access to AuthenticationController.GetPasswordConfig, as this is used to set a password for newly invited users, before they login

### Tests
- Add SMTP info in config (I use [smtp4dev](https://github.com/rnwood/smtp4dev) to create a local fake smtp server)
- Invite user from user section
- Follow the link in the mail
- Ensure the user can create password and login


elit0451:
The User invite link indeed works now, but to be fully functional I still spotted a few bugs that we need to decide if they should be part of this fix or maybe we need another one. I compared the invite flow with v8 and those issues seem to show up in netcore:
- when I follow the invite link & fill in the new password for the user, I get an option to upload a photo. I tried doing that with different images but it never succeeds - it gives a Not Found error.
- after the option to upload an image, I am able to access the user profile -> meaning that the login succeeded, but if I try to change the user's password, I get "_Password reset is not allowed_" error message. In v8 I also have a field for inputting the old password but in netcore I just have the option to provide a new password.

I replicated the same steps on v8/dev but there those 2 issues ⬆️ don't appear!

elit0451:
Great work 🎉💪 Merging!

PR #9618, opened 2021-01-06
Document onChange in umbPagination

bjarnef:
### Prerequisites

### Description
I noticed `onChange` function is used several placed, but not documented.
Furthermore inside `goToPage` it has double conditional check of `scope.onChange`.

It seems the `onChange` function wasn't originally added https://github.com/umbraco/Umbraco-CMS/commit/46826ad5900528858ff1ee7b51fc0c81fec63b19 on some places in core it use `on-change` while other places in use `on-go-to-page`.


nathanwoulfe (approved):
Nice @bjarnef. Guessing the double-if was introduced by a messy merge at some point.

PR #9619, opened 2021-01-06
Migrating more routing

Shazwazza:
Split up IPublishedRequest into a builder and an immutable object. Ports redirects, headers, not-found results, and more.

* new IPublishedRequestBuilder for building the request parameters
* this builds a IPublishedRequest which is immutable and assigned to the request/route values/umbraco context
* IPublishedRequest no longer has a ref to umb ctx and removes unnecessary/old properties. Tries to simplify it as much as possible and then there's new ext methods for some of the helper methods, no more static events
* Ports over the PublishedContentNotFoundHandler -> PublishedContentNotFoundResult
* NoContentRoutes removed since we cannot 'route' to that controller, we dynamically route to it in our value transformer (you cannot 'redirect' from a transformer) also fixes up handling 404s. [AB#9790](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/9790)
* Deal with redirects [AB#9715](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/9715) from IContentFinders in the RenderController
* Deal with custom status codes and headers from IContentFinders in PublishedRequestFilterAttribute 
* IPublishedRouter is now async and uses IEventAggregator to raise the old Preparing/Prepared events - though they are renamed and use CreatingRequestNotification and RoutingRequestNotification. This interface is also cleaned up with a single route method that accepts new routing options
* Ports over more logic that was being done in the umbraco module and more from RenderRouteHandler
  * It's now pretty easy to follow the routing pipeline: dynamic endpoint routing done with `UmbracoRouteValueTransformer` -> `IPublishedRouter` using `IPublishedRequestBuilder` sending that through the routing pipeline and building the resulting `IPublishedRequest` -> `IUmbracoRouteValuesFactory` creates the `UmbracoRouteValues` which does the post inbound routing checks for things like hijacked routes
* Sets default request culture to GlobalSettings.DefaultUILanguage
* Adds custom request culture provider to set the culture based on the published request result. The existing `UmbracoBackOfficeIdentityCultureProvider` wasn't being used at all, now it is. We now have 2x culture providers, one for the back office and one for the front end and it's working now. It was broken before due to the ordering of middleware

## Testing

* All tests passing
* Back office works
* Front-end works
* Test the culture for the back office i.e. change the culture of your user, log out, log in, see that the culture is set in the request (you may need to breakpoint to see this and watch `System.Globalization.CultureInfo.CurrentCulture` , for example on ContentController.GetById)
* Test the culture for the front-end i.e. you can render the current culture out to a view with `System.Globalization.CultureInfo.CurrentCulture` and then go change your default installed language to something new, then go to domains/cultures settings of your root node and instead of choosing 'inherit' which will just use the default configured culture, choose the language. Now your pages will be rendered in that culture.
* Test that the no content result works i.e. go to a front-end page that doesn't exist and you'll see the standard 'ugly' umbraco blank page


src/Umbraco.Core/Routing/PublishedRouter.cs
```diff
@@ -635,21 +654,18 @@ private void FindTemplate(IPublishedRequest request)
...
-                // not sure how were going to do this nicely.
-
                 // TODO: We need to limit altTemplate to only allow templates that are assigned to the current document type!
                 // if the template isn't assigned to the document type we should log a warning and return 404
-
                 var templateId = request.PublishedContent.TemplateId;
-                request.TemplateModel = GetTemplateModel(templateId);
+                ITemplate template = GetTemplate(templateId);
+                request.SetTemplate(template);
+                _logger.LogDebug("FindTemplate: Running with template id={TemplateId} alias={TemplateAlias}", template.Id, template.Alias);
```
bergmania, line 668: I see template can be null, and then this line throws null ref exception

bergmania (commented):
> Test the culture for the front-end i.e. you can render the current culture out to a view with System.Globalization.CultureInfo.CurrentCulture and then go change your default installed language to something new, then go to domains/cultures settings of your root node and instead of choosing 'inherit' which will just use the default configured culture, choose the language. Now your pages will be rendered in that culture.

I can't make this work. When I change my language to danish, and set that in the "Cultures and Hostnames", `System.Globalization.CultureInfo.CurrentCulture` is still "en-US" in the view

image

image

image

image

src/Umbraco.Tests/Routing/ContentFinderByUrlAndTemplateTests.cs
```diff
@@ -25,24 +26,31 @@ Template CreateTemplate(string alias)
...
+            var result = lookup.TryFindContent(reqBuilder);
+
+            IPublishedRequest frequest = reqBuilder.Build();
 
             Assert.IsTrue(result);
             Assert.IsNotNull(frequest.PublishedContent);
-            Assert.IsNotNull(frequest.TemplateAlias);
-            Assert.AreEqual("blah".ToUpperInvariant(), frequest.TemplateAlias.ToUpperInvariant());
+            Assert.IsNotNull(frequest.GetTemplateAlias());
+            Assert.AreEqual("blah".ToUpperInvariant(), frequest.GetTemplateAlias().ToUpperInvariant());
```
bergmania, line 53:
```suggestion
            var templateAlias = frequest.GetTemplateAlias();
            Assert.IsNotNull(templateAlias );
            Assert.AreEqual("blah".ToUpperInvariant(), templateAlias.ToUpperInvariant());
```

src/Umbraco.Web.BackOffice/Mapping/ContentMapDefinition.cs
```diff
@@ -199,13 +199,21 @@ private IEnumerable<string> GetActions(IContent source)
...
-                ? new[] { UrlInfo.Message("Cannot generate URLs without a current Umbraco Context") }
-                : source.GetContentUrls(_publishedRouter, umbracoContext, _localizationService, _localizedTextService, _contentService, _variationContextAccessor, _loggerFactory.CreateLogger<IContent>(), _uriUtility, _publishedUrlProvider).ToArray();
+            if (umbracoContext == null)
+            {
+                return new[] { UrlInfo.Message("Cannot generate URLs without a current Umbraco Context") };
+            }
+
+            // NOTE: unfortunately we're not async, we'll use .Result and hope this won't cause a deadlock anywhere for now
+            var urls = source.GetContentUrlsAsync(_publishedRouter, umbracoContext, _localizationService, _localizedTextService, _contentService, _variationContextAccessor, _loggerFactory.CreateLogger<IContent>(), _uriUtility, _publishedUrlProvider)
+                .Result
```
bergmania, line 215: Is it safe ? Should we use GetAwaiter().GetResult() instead?

src/Umbraco.Web.Common/ActionsResults/PublishedContentNotFoundResult.cs
```diff
@@ -0,0 +1,60 @@
...
+            await response.WriteAsync("<html><body><h1>Page not found</h1>");
+            await response.WriteAsync("<h2>");
+            await response.WriteAsync(string.Format(reason, WebUtility.HtmlEncode(_umbracoContext.OriginalRequestUrl.PathAndQuery)));
+            await response.WriteAsync("</h2>");
+            if (string.IsNullOrWhiteSpace(_message) == false)
+            {
+                await response.WriteAsync("<p>" + _message + "</p>");
+            }
+
+            await response.WriteAsync("<p>This page can be replaced with a custom 404. Check the documentation for \"custom 404\".</p>");
```
bergmania, line 55: I know it is not a new thing, but we should make a umbra.co/404-docs link and point to the docs

Shazwazza:
> I can't make this work. When I change my language to danish, and set that in the "Cultures and Hostnames", System.Globalization.CultureInfo.CurrentCulture is still "en-US" in the view

Ah, this is unfortunate 😢 I was testing with en-AU which worked because that culture was accepted as a supported culture. It turns out that the request localization middleware does not support dynamic cultures whatsoever. You need to explicitly tell the app what cultures you support and these cannot be changed anymore, see https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.requestlocalizationoptions.supportedcultures?view=aspnetcore-5.0 "will only set the current request culture to an entry in this list. Defaults to CurrentCulture.". Microsoft will log this out in Debug mode that the culture is not supported. 

After some research I discovered that the RequestLocalizationOptions are mutable which allows us to dynamically configure the supported cultures. This is not documented but it must be mutable for this reason. I'll write a blog post about that in the future. 

Since this is global middleware, an alternative would be to swap the IRequestCultureFeature feature for our own requests using our own middleware but I think just being able to dynamically configured the supported cultures is fine. 

Also note that creating a CultureInfo is expensive: https://github.com/dotnet/aspnetcore/blob/b795ac3546eb3e2f47a01a64feb3020794ca33bb/src/Middleware/Localization/src/RequestLocalizationMiddleware.cs#L165. We used to have a concurrent dictionary to ensure we don't keep re-creating CultureInfo but as it turns out that isn't necessary anyways because [CultureInfo.GetCultureInfo](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.getcultureinfo?redirectedfrom=MSDN&view=net-5.0#System_Globalization_CultureInfo_GetCultureInfo_System_String_) returns a cached readonly version of it (essentially doing the same thing). So I've removed this concurrent dictionary and changed instances of `new CultureInfo` to `CultureInfo.GetCultureInfo` where appropriate. 

I've gone ahead and changed the builder/request to use a string for Culture instead of CultureInfo. This avoids unnecessary allocations and simplifies apis.

This is all fixed now in rev: bd4006c577035bf907f69092d5b9d8d692582b84

src/Umbraco.Core/Routing/PublishedRequest.cs
```diff
@@ -1,481 +1,68 @@
...
+            PublishedContent = publishedContent;
+            IsInternalRedirect = isInternalRedirect;
+            Template = template;
+            Domain = domain;
+            Culture = culture;
+            RedirectUrl = redirectUrl;
+            ResponseStatusCode = responseStatusCode;
+            CacheExtensions = cacheExtensions;
+            Headers = headers;
+            SetNoCacheHeader = cacheabilityNoCache;
```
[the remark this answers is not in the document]
Shazwazza: nope, just missed refactorings, fixed now.

src/Umbraco.Core/Routing/PublishedRouter.cs
```diff
@@ -690,11 +708,13 @@ private void FindTemplate(IPublishedRequest request)
 
                     // no allowed, back to default
                     var templateId = request.PublishedContent.TemplateId;
-                    request.TemplateModel = GetTemplateModel(templateId);
+                    ITemplate template = GetTemplate(templateId);
+                    request.SetTemplate(template);
+                    _logger.LogDebug("FindTemplate: Running with template id={TemplateId} alias={TemplateAlias}", template.Id, template.Alias);
```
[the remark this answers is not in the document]
Shazwazza: Fixed, checks null and logs warning

src/Umbraco.Core/Templates/UmbracoComponentRenderer.cs
```diff
@@ -43,7 +43,7 @@ public IHtmlEncodedString RenderTemplate(int contentId, int? altTemplateId = nul
             {
                 try
                 {
-                    _templateRenderer.Render(contentId, altTemplateId, sw);
```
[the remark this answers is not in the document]
Shazwazza: good catch, have fixed that and others.

src/Umbraco.Web.BackOffice/Mapping/ContentMapDefinition.cs
Shazwazza: In some cases it won't matter and it's never 100% safe. It will be safe on request threads but it won't be safe on threadpool (if maxed out)/ui threads. In some cases you can wrap in a `Task.Run` to avoid a deadlock but that all depends on where the code is called from or jump through other hoops. For now we'll do `.ConfigureAwait(false).GetAwaiter().GetResult()` so at least any exception is unwrapped and we're not capture the sync context. 

The true solution is to either make the mapper async (probably best) or create native sync methods like GetContentUrls but to do that will also require making native sync methods of the new event aggregator.

src/Umbraco.Web.Common/ActionsResults/PublishedContentNotFoundResult.cs
Shazwazza: fixed

bergmania: I changed to a tiny link, so we can change the redirect without updating umbraco if we someday need that..

PR #9624, opened 2021-01-07
Fix link localization  in EnsureInternalLinks

werenskjold:
url provider caches resolved fragments, resulting in localized url's not returning expected. 

### Prerequisites


nul800sebastiaan:
Hi there @werenskjold and I'm very sorry for the late reply!

From your description I can't quite make out what the problem is, can you give an example of the behavior you're seeing and the behavior you would expect?

PR #9629, opened 2021-01-08
V8: Fix bug where allowing culture variation would delete content from element types #7471

### LINKED ISSUE #7471: V8.5.1: 'Allow Varying by Culture' on a composition is deleting field content
With a fresh installl of umbraco 8.5.1 + Starter kit, when setting Allow Varying by Culture on compositions and then subsequently the document type (composed with said composition) removes the content from the given amended field.

I can confirm that this has occurred in my own development site, so is not just limited to starter kit.

## Umbraco version

I am seeing this issue on Umbraco version: 8.5.1

Reproduction
------------

### Bug summary

As above, or view video here...
https://www.loom.com/share/58af6f2c4d9040bda393496d80da451e

### Specifics

Using Chrome, i have'nt checked it in other browsers as i would suggest not relevant.

### Steps to reproduce

1 - Fresh Umbraco 8.5.1 installed via nuget.  Default installation with starter kit.
  
2 - Check About Us page in Umbraco backend shows a page title field populated with "About Us" and check subnode of About this Starter Kit contains a valid Page Title. - Yes

3 - Goto to the ContentBase Composition choose Permissions > Allow varying by culture

4a - Allow varying by culture on the pageTitle field too and save

5 - Repeat step 2 to confirm that About Us and any of its subnodes still contain a populated Page Title. - Yes 

6 - Goto the "Content Page" Document Type (note - composed with ContentBase) choose Permissions > Allow varying by Culture.  Save

7 - Repeat step 2 to confirm that About Us and any of its subnodes still contain a populated Page Title. - No Page Title is empty.

8 - Check Blog(root) Page in Umbraco Backend (also composed with ContentBase) to confirm contains a populated Page Title. - Yes 

9 - Goto the "Blog" Document Type choose Permissions > Allow varying by Culture.  

10 - Repeat step 8 - to confirm that Blog still contain a populated Page Title. - No Page Title is empty.

### Expected result

I expected nothing to have changed with regards to the Content in the fields changed via the document type.

### Actual result

The content in the field was deleted.

---
_This item has been added to our backlog [AB#4637](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/4637)_

nikolajlauridsen:
This should fix the bug where if you allowed varying culture in a composition and property field, that property field would be lost in any content using the property type. 

The underlying issue was that when we were checking for a change in variation we only looked at the previous state of the property type, however, when you're using it in an element type, the element type and property type can allow variation, while the document type you're using it in as a composition does not allow it. When you later allow variation in you document type the property type will already allow variation (because it's allowed in the element type), and therefore we did not copy the property data to a new column that has a language ID associated with it, causing this bug.

We now also check the source document type for variation when changing variation for composed property types.

I've also made a unit tests that hit this issue. 

fixes #7471 

### Testing: 
1. Ensure the project is empty
2. Create an element type with a text property
3. Create a document type with root access, that uses the element type as a composition
4. Create a root content node and add some text to the property
5. Enable variants on the element type and the text property
6. Enable variants on the doc type
7. Ensure the content in the text property still exists in the root content node.

I've tried to fiddle around to find any side effects of this change, but couldn't find any, would be helpful to try and fiddle around with changing culture settings as well.

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
@@ -448,8 +448,14 @@ AND umbracoNode.id <> @id",
                     // The composed property is only considered segment variant when the base content type is also segment variant.
                     // Example: Culture variant content type with a Culture+Segment variant property type will become ContentVariation.Culture
                     var target = newContentTypeVariation & composedPropertyType.Variations;
+                    // Determine the previous variation
+                    // We have to compare with the old content type variation because the composed property might already have changed
+                    // Example: A property with variations in an element type with variations is used in a document without
+                    //          when you enable variations the property has already enabled variations from the element type,
+                    //          but it's still a change from nothing because the document did not have variations, but it does now.
+                    var from = oldContentTypeVariation & composedPropertyType.Variations;
 
-                    propertyTypeVariationChanges[composedPropertyType.Id] = (composedPropertyType.Variations, target);
+                    propertyTypeVariationChanges[composedPropertyType.Id] = (from, target);
                 }
             }
 
--- a/src/Umbraco.Tests/Persistence/Repositories/ContentTypeRepositoryTest.cs
+++ b/src/Umbraco.Tests/Persistence/Repositories/ContentTypeRepositoryTest.cs
@@ -14,6 +14,7 @@ using Umbraco.Tests.TestHelpers;
 using Umbraco.Tests.TestHelpers.Entities;
 using Umbraco.Tests.Testing;
 using Umbraco.Web.Models.ContentEditing;
+using Content = Umbraco.Core.Models.Content;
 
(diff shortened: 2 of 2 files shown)


Shazwazza:
@nikolajlauridsen the change looks simple enough and having that unit test to support the change is the most important part. You'll definitely understand the issue more than me so as long as you're confident that the test you've written replicates the issue explained in https://github.com/umbraco/Umbraco-CMS/issues/7471 then I'm all good with the fix since we've got quite a few other tests which are still passing for other scenarios.

Since @warrenbuckley is assigned to review/test this today I think it's still important that he does reviews and runs through your test steps - as well as ensuring that the issue and replications steps of https://github.com/umbraco/Umbraco-CMS/issues/7471 are resolved too. 

Assuming all the tests pass I'm happy for this to be merged 🚀 

Nice work!

warrenbuckley:
Following the test notes I can confirm it works as described and the notes from the reported issue #7471

Issue #9541, opened 2020-12-11
umb-search-filter broken in "Select editor" window


glombek:
umb-search-filter is broken in the "Select editor" window when adding a new property to a document type.

## Umbraco version

I am seeing this issue on Umbraco version: v8/contrib (to be 8.11.0)

Cannot replicate in 8.9.1 (not tested in 8.10.*)

Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

### Steps to reproduce

1. Open up a document type
2. Add a new property
3. Select editor type
4. Broken search box appears

### Expected result

Search icon is aligned with text box

Screenshot from 8.9.1:
![image](https://user-images.githubusercontent.com/3120611/101952686-f398cf80-3bf0-11eb-8aa6-59ba82efe257.png)

### Actual result

Search icon is misaligned
![image](https://user-images.githubusercontent.com/3120611/101948133-76b62780-3be9-11eb-8528-80e31f8284e8.png)

This likely relates to #9037.

bjarnef:
It seems this has been affected by the changed in this PR where `icon-search` was removed from the `<umb-icon>` in the search  filter. https://github.com/umbraco/Umbraco-CMS/pull/9255/files#diff-511f1dd6fea9b25e43b0cc0b3981515130d59c98c92cae29c8b26c90ecc338b7L4

The `icon-search` is not required for the icon itself, but it was used in the search filter styling.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/components/umb-search-filter.less#L21-L28

Re-adding this make the search filter look nice again:
![image](https://user-images.githubusercontent.com/2919859/104122736-1f6db900-5347-11eb-9e2d-3625cca85345.png)

We can of course style specific on `.umb-icon` class instead, but I wonder if we can avoid breaking the styling if packages used the old markup.

bjarnef:
The old markup looked like this:

```
<div class="form-search">
         <umb-icon icon="icon-search" class="icon-search"></umb-icon>
         <input type="text"
                     style="width: 100%"
                     ng-model="searchTerm"
                     class="umb-search-field search-query input-block-level"
                     localize="placeholder"
                     placeholder="@placeholders_filter"
                     umb-auto-focus
                     no-dirty-check />
</div>
```

or:

```
<div class="form-search">
         <i class="icon-search" aria-hidden="true"></i>
[...]

PR #9631, opened 2021-01-10
Fix styling of search icon in search filter

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9541

### Description
This PR fixes some styling issues with search icon inside search filter after the `<umb-search-filter>` component has been introduced and changed to `<umb-icon>` inside search filter component, where the `icon-search` was removed. I have kept to old class so it still looks nice if packages used the old markup.

It seems a bit unnecessary to have the `w-100` class here
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/components/umb-search-filter.less#L6-L8 since we already have the utility class here https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/utilities/_width.less#L24

but removing this + the `html` selector in the component styling, it seems the styling in forms.less is prioritized, so the padding was overwritten by forms.less styling.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/components/umb-search-filter.less#L14-L19

We can look at this in another PR. For now this should fix these styling issues.

**Icon Picker**
image

**DataType Picker**
image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-search-filter.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-search-filter.less
@@ -18,6 +18,8 @@ html .umb-search-filter {
         margin: 0;
     }
 
+    // "icon-search" class it kept for backward compatibility
+    .umb-icon, 
     .icon-search {
         color: #d8d7d9;
         position: absolute;


OwainWilliams (approved):
Looks good. Thanks @bjarnef - This does indeed fix the issue.

Issue #9466, opened 2020-11-28
Broken layout for member group id


bjarnef:
With the latest changes in `v8/contrib` branch it seeems the layout of id on member groups is broken compared to how it used to look like and not matching the layout in e.g. content and media.

![image](https://user-images.githubusercontent.com/2919859/100518534-5d958b80-3192-11eb-8ed2-9e52d9b5307b.png)

bjarnef:
@nielslyngsoe I noticed you recently made some changes in the "info" group - not sure of some of those changes have affected this issue?

bjarnef:
This is how it looks in v8.9.0

![image](https://user-images.githubusercontent.com/2919859/100552370-20f78c00-3287-11eb-970e-4a8de150b282.png)

nielslyngsoe:
@bjarnef thanks for noticing. Member groups info page was not set up correctly, but that's corrected in this PR.
https://github.com/umbraco/Umbraco-CMS/pull/9469

warrenbuckley:
Will be due in 8.10

PR #9469, opened 2020-11-30
Review Remove this form-horizontal class from member groups info

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/9466

---
_This item has been added to our backlog [AB#9568](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/9568)_

PR #9632, opened 2021-01-10
Align datatype sidebar box properties

bjarnef:
### Prerequisites

### Description
I noticed the properties wasn't aligned in "General" box under datatypes. It is a similar issue as this for member groups: https://github.com/umbraco/Umbraco-CMS/issues/9466

**Before**

image

**After**

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/forms.less
+++ b/src/Umbraco.Web.UI.Client/src/less/forms.less
@@ -803,9 +803,12 @@ legend + .control-group {
 
 // adjustments for properties tab
 .form-horizontal .block-form .control-label {
-  display: block;
-  float: none;
-  width: 100%;
+    display: block;
+    float: none;
+    width: 100%;
+}
+.form-horizontal .block-form .controls {
+    margin-left: 0;
 }
 
 //make sure buttons are always on top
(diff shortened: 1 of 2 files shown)


bjarnef:
Oh, I noticed we can't fix this in similar way as member groups as it would effect the entire layout under "Settings" tab.

If removing this class it would change from this:

image


to this:

image

bjarnef:
@nielslyngsoe while your fix in https://github.com/umbraco/Umbraco-CMS/pull/9469 worked for member groups, I think we need another way to solve this, e.g. when having multiple tabs or a edit view with properties + a siderbar info box.

bjarnef:
I noticed we have various styling for `block-form` class inside `main.less` https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/main.less#L163-L178 and `forms.less` https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/forms.less#L804-L809

I have cleared left margin on `controls` inside `block-form` class.

I now looks like this:

nul800sebastiaan:
Thanks @bjarnef - looks good now! I've cherry-picked this for patching in v8.10+ as well, hope this helps for @mattbrailsford as well.

PR #9633, opened 2021-01-10
Ensure sub-view components are not self-closed elements

bjarnef:
### Prerequisites

### Description
I noticed we have several uses of `<umb-editor-sub-view>` as self-closed (void) elements, but with latest AngularJS v1.x.x version recently upgraded to, we have noticed several issues - mostly that any following sibling elements wouldn't be rendered.

In this case it isn't an issue at the moment, because it is the only child element where used, but if we later add text or elements after the component, it would cause issues rendering these elements.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/content/umb-variant-content.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/content/umb-variant-content.html
@@ -36,7 +36,8 @@
 
             <div class="umb-editor-sub-views">
                 <div ng-repeat="app in vm.editor.variantApps track by app.alias">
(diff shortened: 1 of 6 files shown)


nathanwoulfe:
@bjarnef nice job, these absolutely need to be fixed - self-closing/void elements are a very particular sub-set of html elements, and all are known by the parser (ie, it's not possible to create our own). All AngularJs (or any framework, for that matter) element directives MUST be closed/non-void.

nathanwoulfe (approved):

nul800sebastiaan:
Thanks @bjarnef ! 👍

PR #7227, opened 2019-11-26
V8/feature/0000 remove out icon

nielslyngsoe:
General style adjustments, getting rid of some visual noise, including:

- Removing the out-icon for Links in Content-Into-tab and Media-Info-Tab.
- using umb-outline for links.
- added round corners and corrected colors for Links.
- adjust the style for XS badges.
- changes User badge to use XS.

PR #9634, opened 2021-01-10
Hide rollback button for content that hasn't been created yet

bjarnef:
### Prerequisites

### Description
When creating a new content new there is no versions yet, so I think we should also hide the rollback button in this case as there is no versions to rollback to anyway.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/content/umb-content-node-info.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/content/umb-content-node-info.html
@@ -45,7 +45,7 @@
 
             <umb-box-header title="{{historyLabel}}">
                 <umb-button
-                    ng-hide="node.trashed"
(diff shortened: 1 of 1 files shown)


bjarnef:
I noticed the history label is also missing at the moment, but we can look at this in other PR.
I have added some comments to this previous PR, where it added `historyLabelKey` in directive, but still use the `historyLabel` in view. https://github.com/umbraco/Umbraco-CMS/pull/7227

nathanwoulfe (approved):
Looks good to me, thanks @bjarnef. Ideally my changes in #9648 would be merged at the same time, else we have the header collapsing on a new node as there's no label and no button.

@mikecp / @OwainWilliams feel like reviewing #9648? 😉

mikecp:
@nathanwoulfe Will try to have a look later today/tonight if @OwainWilliams does not beat me to it 😉

PR #9640, opened 2021-01-12
Fixes a memory leak caused by deep cloning

### LINKED ISSUE #9348: High CPU usage during Examine indexing which takes 16h to complete
When examine indexes are created against a site with many Public Access protected nodes (and these protected nodes have many member roles assigned to them), the indexing process takes an extremely long time (16h for 1000 protected nodes on a P1v2:1 Azure App Service Plan).

## Umbraco version

I am seeing this issue on Umbraco version: 8.8

Reproduction
------------

I have a site with 4000 content nodes, of which around 1000 are protected by public access restrictions. The protected nodes have up to 10 different member roles assigned to them. When starting the site up with no ExamineIndexes folder, or when explicitly rebuilding the indexes, indexing lasts a very long time (16h!) and uses up a constant minimum of 50% CPU during this time, significantly slowing down the website and backoffice.

### Bug summary

Umbraco 8.8 has an issue when indexing many protected nodes that have many member roles associated with them.

I have actually investigated this problem against Umbraco 8.8 source, and using CPU profiling, narrowed it down to the `PerformDeepClone` method of `PublicAccessEntry.cs`, specifically the line:

`cloneEntity._ruleCollection.CollectionChanged -= _ruleCollection_CollectionChanged;`

When removing the public access entries from all nodes, the site indexes in matter of seconds to minutes. Likewise, when using a Custom UmbracoIndexConfig as detailed [here](https://our.umbraco.com/Documentation/Reference/Searching/Examine/indexing/#changing-ivaluesetvalidator) and setting supportProtectedContent to true in the following override, the indexing also finishes very quickly.

`IContentValueSetValidator IUmbracoIndexConfig.GetPublishedContentValueSetValidator() => new ContentValueSetValidator(true, true, PublicAccessService);`

This bypasses the offending block of code (as no checks are carried out to see if a node is protected) and is our current workaround for this issue.

### Steps to reproduce

- Create a site with many protected nodes, with many member roles set against each protected node (using the public access API)
- Index the site with Examine

### Expected result

The indexing takes a few minutes to finish

### Actual result

The indexing takes many, many hours (16h for 1000 protected nodes)

---
_This item has been added to our backlog [AB#9890](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/9890)_

Shazwazza:
Resolves #9348 

There was a memory leak with PublicAccessEntry during even unassignment which was not clearing the correct handler.
This goes a step further and adds a new ClearCollectionChangedEvents method for all observable collections used in umbraco
which allows fully clearing ALL event handlers instead of having to track specific ones. This will ensure there are no
unintended memory leaks in case end-users have assigned event handlers to the collection changed event which would
not be unassigned during deep cloning. This is much safer.

This memory leak will be present in all Umbraco 8.x versions if there is any public access assigned. The more public access entries configured (i.e. the more nodes that have explicit public access settings applied) the faster the leak.

There is still one issue however which I'll look to resolve in a different PR which is that we cache all public access entries in memory. In some rare cases there could be thousands which is both inefficient to store in memory but also inefficient to query in memory. However the alternative is more SQL queries so I need to determine if there's a happy medium to be had. This cache has existed for most of v7 so it's nothing new. 

## Results

To test this I used an [artillery](https://artillery.io/docs/guides/overview/welcome.html) script. This first calls a custom endpoint `/MemberLogin` that just does a very rudimentary member login with the specified query string parameters. Artillery tracks cookies returned for the virtual user so the resulting cookie will be used for the subsequent requests. It actually isn't necessary that the member is logged in to see this leak! ... it occurs on any check of a page's public access settings but it was worth adding this to this script. 

The test is on the Starter Kit and public access is applied to the about-us page with a few dozen different rules for different user groups. That node and rules are then copied many many times which results in about 1000 entries in the `umbracoAccess` table. This isn't a requirement for testing but it makes the leak much faster.

```yml
config:
  target: 'http://localhost:8910/'
  http:
    pool: 30
  phases:
    - duration: 20
      arrivalRate: 2
scenarios:
  - flow:
    - get:
        url: "/MemberLogin"
        qs:
          username: "testmember"
          passowrd: "123456789"
    - loop:      
      - get:
          url: "/products/"
      - get:
          url: "/about-us/"
      count: 10
```

### Before

These 2 snapshots were taking after testing a few times so it's not an indication of initial memory consumption but it clearly indicates a memory leak. When the artillery script was first run way before the first snapshot was taken the `RPS sent` was 41.38. After each execution and report output this drops significantly as memory grows. This is because the CPU is now working overtime iterating through a lot of objects.

The last report output when snapshot 2 was taken reported `RPS sent`: 7.1 😱 

image

### After

These 2 snapshots were taken right after starting the app which is why the actual memory consumption figures are much lower but that's not what we're comparing here. This clearly shows there is no increase in memory retention as the tests are run.

The first report output was showing `RPS sent`: 42.22 and the last report was `RPS sent`: 42.08 so that shows that the CPU issue and RPS output is directly related to the memory leak.

_NOTE: RPS in these reports is not an indication that umbraco can only serve 42 requests per second! Whatever might be happening on my machine and how these pages are coded, etc... all impact this figure, it's just used as a measurement for comparison with other contextual tests at the same time on the same machine._

image

### DIFF
--- a/src/Umbraco.Core/Collections/EventClearingObservableCollection.cs
+++ b/src/Umbraco.Core/Collections/EventClearingObservableCollection.cs
@@ -1,0 +1,33 @@
+﻿using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Collections.Specialized;
+
+namespace Umbraco.Core.Collections
+{
+    /// <summary>
+    /// Allows clearing all event handlers
+    /// </summary>
+    /// <typeparam name="TValue"></typeparam>
+    public class EventClearingObservableCollection<TValue> : ObservableCollection<TValue>
+    {
+        public EventClearingObservableCollection()
+        {
+        }
+
+        public EventClearingObservableCollection(List<TValue> list) : base(list)
+        {
+        }
+
+        public EventClearingObservableCollection(IEnumerable<TValue> collection) : base(collection)
+        {
+        }
+
+        // need to override to clear
+        public override event NotifyCollectionChangedEventHandler CollectionChanged;
+
+        /// <summary>
+        /// Clears all event handlers for the <see cref="CollectionChanged"/> event
+        /// </summary>
+        public void ClearCollectionChangedEvents() => CollectionChanged = null;
+    }
+}
--- a/src/Umbraco.Core/Collections/EventClearingObservableCollection.cs
+++ b/src/Umbraco.Core/Collections/EventClearingObservableCollection.cs
@@ -8,7 +8,7 @@ namespace Umbraco.Core.Collections
     /// Allows clearing all event handlers
     /// </summary>
     /// <typeparam name="TValue"></typeparam>
-    public class EventClearingObservableCollection<TValue> : ObservableCollection<TValue>
+    public class EventClearingObservableCollection<TValue> : ObservableCollection<TValue>, INotifyCollectionChanged
     {
         public EventClearingObservableCollection()
         {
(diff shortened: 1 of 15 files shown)


bergmania (commented):
Changes looks good. Just found a single xmldoc that was inserted in a wrong position.

I didn't found anything that does not work in my tests.

PR #9642, opened 2021-01-12
AB#9725 - Removed IUmbracoApplicationLifetimeManager

bergmania:
Fixes [AB#9725](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/9725)

### Details
- Removed `IUmbracoApplicationLifetimeManager`
- Removed `IUmbracoApplicationLifetime.ApplicationInit`
- Removed `AspNetCoreComposer` + `AspNetCoreComponent`
- Migrated `WebProfilerComposer` + `WebProfilerComponent`
  - Added to UmbracoBuilder + Created new `InitializeWebProfiling` notification hander executed on `UmbracoApplicationStarting`
- Fixed issues with web profiler rendering on the frontend (This can't be handled in the same way anymore)

### Test
- Run in debug mode
- Visit https://localhost:44331/?umbdebug=true
  - First request should include the boot profiling
  - Subsequent request should only include the request profiling


src/Umbraco.Web.Common/Profiler/InitializeWebProfiling.cs
```diff
@@ -0,0 +1,56 @@
...
+        private readonly bool _profile;
+        private readonly WebProfiler _profiler;
+        private readonly IUmbracoRequestLifetime _umbracoRequestLifetime;
+        private readonly List<Action> _terminate = new List<Action>();
+        public InitializeWebProfiling(IProfiler profiler, IUmbracoRequestLifetime umbracoRequestLifetime, ILogger<InitializeWebProfiling> logger)
+        {
+            _umbracoRequestLifetime = umbracoRequestLifetime;
+            _profile = true;
+
+            // although registered in WebRuntime.Compose, ensure that we have not
```
Shazwazza, line 24: Is this still relevant?

src/Umbraco.Web.Common/Profiler/InitializeWebProfiling.cs
```diff
@@ -0,0 +1,56 @@
...
+using Umbraco.Web.Common.Lifetime;
+
+namespace Umbraco.Web.Common.Profiler
+{
+    public class InitializeWebProfiling : INotificationHandler<UmbracoApplicationStarting>
+    {
+        private readonly bool _profile;
+        private readonly WebProfiler _profiler;
+        private readonly IUmbracoRequestLifetime _umbracoRequestLifetime;
+        private readonly List<Action> _terminate = new List<Action>();
```
Shazwazza, line 18: Adding stuff to _terminate doesn't seem to do anything now

src/Umbraco.Tests/Scoping/ScopedNuCacheTests.cs
```diff
@@ -106,7 +105,7 @@ protected override IPublishedSnapshotService CreatePublishedSnapshotService(Glob
                 hostingEnvironment,
                 Microsoft.Extensions.Options.Options.Create(nuCacheSettings));
 
-            lifetime.Raise(e => e.ApplicationInit += null, EventArgs.Empty);
+            //lifetime.Raise(e => e.ApplicationInit += null, EventArgs.Empty);
```
Shazwazza, line 108: can remove comment

src/Umbraco.Core/DependencyInjection/UmbracoBuilder.cs
```diff
@@ -114,9 +114,8 @@ private void AddCoreServices()
             // Adds no-op registrations as many core services require these dependencies but these
             // dependencies cannot be fulfilled in the Core project
             Services.AddUnique<IMarchal, NoopMarchal>();
-            Services.AddUnique<IProfiler, NoopProfiler>();
+            //Services.AddUnique<IProfiler, NoopProfiler>();
```
Shazwazza, line 117: can remove commented out code. do we still need NoopProfiler? it's being checked for but doesn't seem like we use/need it?

Shazwazza (changes requested):
Left a few comments. Tests and profiler tests are passing though

src/Umbraco.Core/DependencyInjection/UmbracoBuilder.cs
bergmania: We still new up `NoopProfiler` in `Umbraco.Web.Common.DependencyInjection.UmbracoBuilderExtensions.GetWebProfiler`. So it is still needed

src/Umbraco.Web.Common/Profiler/InitializeWebProfiling.cs
bergmania: I'll update the comment, but in theory, the profiler can still be replaced by another type before the `UmbracoApplicationStarting` notification is handled

src/Umbraco.Web.Common/Profiler/InitializeWebProfiling.cs
bergmania: Damn, of course not..  As har as I know, we don't really have a good way to handle this? 
1. We can use the `UmbracoApplicationStopping` event, but then we don't really has the connection between the two notification handlers?
2. We could change the IUmbracoRequestLifetime.RequestStart event to be handled by the EventAggregator.

Issue #9432, opened 2020-11-23
Backoffice UI - checkboxes on doctype compositions uncheck after selecting an item


Jay-umbr:
Adding a composition to a doctype - the checkboxes become greyed out after selecting an item for a composition, so you cannot take it out unless you re-open the side panel.

## Umbraco version

I am seeing this issue on Umbraco version: .8.9.1

Reproduction
------------
Hope a video is OK.
https://drive.google.com/file/d/1idIlLDupKkt3jGKzwNk7r59EDGLJpylX/view?usp=sharing
Create two document types in the backoffice and try to use one as a composition on the other doctype.

This does NOT happen on a version 8.6, from what I tested

### Expected result

We add the item in the composition, and while it is added, the checkbox remains checked.

### Actual result

Once the item is added the checkbox is greyed out, and the item cannot be removed as quickly.

JasonElkin:
Also seeing this after upgrading to 8.9.1.

![compositions bug](https://user-images.githubusercontent.com/7151793/100280916-da412380-2f60-11eb-8889-d0b6902ddd74.gif)

marcemarc:
![CompositionGreysOutUnexpectedly](https://user-images.githubusercontent.com/260802/101657575-f7372580-3a3b-11eb-91c2-1ab4bc6ced98.gif)
Let me add this as a reminder to myself to check the issue list first before creating the video!

nul800sebastiaan:
This has been fixed with PR https://github.com/umbraco/Umbraco-CMS/pull/9644 and I've cherry picked this for all upcoming patch releases of 8.9 upwards. Additionally, this fix will be available in the upcoming v8.11.0 (of which the RC should be out later this week).

nul800sebastiaan:
*Note:* This didn't make it into the 8.11.0-RC build but will be in the final build of 8.11.0.

PR #9644, opened 2021-01-12
V8/bug/9432: Backoffice UI - checkboxes on doctype compositions uncheck after selecting an item

madsrasmussen:
I have made a couple of updates with the PR to make sure that the following functionality works as expected:

- Disable checkboxes while loading a composition
- Disable and fade the doc types which are not allowed to be added
- Highlight the doc types that is selected

How to test:

1. Open an Umbraco with The Starter Kit installed
2. Create a new doctype
3. Add the Home doc type as a composition
4. Make sure you can't select the Product and Content Base doc types
5. Select any other doc type
6. Make sure the checkmark is still in the checkbox for both Home and the new doc type
7. Make sure you can still deselect both Home and the new doc type
8. Make sure you can't select or deselect any doc types while the overlay is loading

I have also moved the loading indicator. It is now next to the selected doc type. I think this gives a better understanding of what is going on.

https://user-images.githubusercontent.com/6078361/104335118-64cee980-54f3-11eb-8593-db6a6b7f2f1b.mp4

Fixes: #9432

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/buttons/umb-button.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/buttons/umb-button.less
@@ -63,6 +63,11 @@
    border-left-color: @white;
 }
 
+.umb-button__progress.-black {
+   border-color: rgba(255, 255, 255, 0.4);
+   border-left-color: @black;
+}
+
 .umb-button__success,
 .umb-button__error {
    position: absolute;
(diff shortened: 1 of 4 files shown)


warrenbuckley:
Thanks for the video @madsrasmussen 
Makes it nice and easy to know what I am testing and looking for 👍 

Off to read code and test it out :)

warrenbuckley:
Code looks fine off to test this out :)

PR #9646, opened 2021-01-12
Eye Dropper Color Picker

bjarnef:
### Prerequisites

### Description
Now that we have `<umb-color-picker>` component since Umbraco v8.10.0 https://github.com/umbraco/Umbraco-CMS/pull/8812 and we also use this for configuration of background and icon color in Block List editor https://github.com/umbraco/Umbraco-CMS/pull/9056 it would be nice to have a basic color picker property editor.

We already have the "Approved Color" datatype instance using "Color Picker" property editor, which work great most of the time. However for various projects we have had the need using a wider range of colors, where the current color property editor might not always fit the needs.

This PR implement a Eyedropper Color Picker (naming is hard, so we might want to rename this - however "Color Picker" and "colorpicker" alias is already reserved by the current property editor).

**Configuration**
_Limited to just the most basic configuration._

image

**Content** 
image


bjarnef:
Not sure if a better name than "Eyedropper" might fit this color picker. Maybe "Compact Color Picker" or "Tiny Color Picker"?
Feel free to add more suggestions 🤓

bjarnef:
"Color Palette Picker", "Spectrum Color Picker",..?

bjarnef:
I have changed the alias of the picker to `Umbraco.ColorPicker.EyeDropper` .. a bit similar to format of `Umbraco.Dropdown.Flexible` alias.

I also considered the name "Spectrum Color Picker" with alias `Umbraco.ColorPicker.Spectrum`.

bjarnef:
@nielslyngsoe maybe you have some feedback on this PR? 🙂

bjarnef:
**Mandatory property**

image

image

nul800sebastiaan:
We're loving it! Thanks for the work @bjarnef - looking really good and works well! 👍

vaskinhu:
Any plans on making the Palette configurable?

nielslyngsoe:
@bjarnef 
Yeah the styling with the custom message seems wrong. I'm not sure if the red text color serves any purpose as the tooltip already is quite dramatic, so that might just need to go away.

bjarnef:
> Yeah the styling with the custom message seems wrong. I'm not sure if the red text color serves any purpose as the tooltip already is quite dramatic, so that might just need to go away

@nielslyngsoe I believe the red messages is just the core UI ... I recall these where later changes to a yellow/warning style instead.

PR #9647, opened 2021-01-13
Adding tests for routing

Shazwazza:
* less allocations with AbsolutePathDecoded as part of IPublishedRequestBuilder/IPublishedRequest so that content finders don't have to continuously re-allocate new strings by calling `Uri.GetAbsolutePathDecoded()`
* less allocations in UmbracoContext to lazily allocate the cleaned up URI strings (only when needed)
* removes more aspx checks
* adds interface for IRoutableDocumentFilter
* adds tests for UmbracoRouteValueTransformer
* adds tests for HijackedRouteEvaluator
* adds tests for UmbracoRouteValuesFactoryTests

### DIFF
--- a/src/Umbraco.Core/UriExtensions.cs
+++ b/src/Umbraco.Core/UriExtensions.cs
@@ -67,6 +67,8 @@ namespace Umbraco.Core
 
             // cannot get .AbsolutePath on relative uri (InvalidOperation)
             var s = uri.OriginalString;
+
+            // TODO: Shouldn't this just use Uri.GetLeftPart?
             var posq = s.IndexOf("?", StringComparison.Ordinal);
             var posf = s.IndexOf("#", StringComparison.Ordinal);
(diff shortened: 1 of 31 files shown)


src/Umbraco.Core/UriExtensions.cs
```diff
@@ -61,10 +61,14 @@ public static Uri Rewrite(this Uri uri, string path, string query)
...
         {
             if (uri.IsAbsoluteUri)
+            {
                 return uri.AbsolutePath;
+            }
 
             // cannot get .AbsolutePath on relative uri (InvalidOperation)
             var s = uri.OriginalString;
+
+            // TODO: Shouldn't this just use Uri.GetLeftPart?
```
bergmania, line 71: `Uri.GetLeftPart` don't work on relativ uri's

bergmania (approved):

PR #9648, opened 2021-01-13
fix history box label, destructure localization response

nathanwoulfe:
### Prerequisites

### Description
Noticed the history box on the info app doesn't have a label - looks like there were potentially conflicts in a previous merge and the wrong changes were kept. Changes here restore the label and tidy up a clunky bit of JS to use destructuring assignment rather than accessing each array index.

History label has been missing since at least 8.8, can't see any existing issue reporting it 😃

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbcontentnodeinfo.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbcontentnodeinfo.directive.js
@@ -54,15 +54,15 @@
 
                 localizationService.localizeMany(keys)
                     .then(function (data) {
-                        labels.deleted = data[0];
-                        labels.unpublished = data[1]; //aka draft
-                        labels.published = data[2];
-                        labels.publishedPendingChanges = data[3];
-                        labels.notCreated = data[4];
-                        labels.unsavedChanges = data[5];
-                        labels.doctypeChangeWarning = data[6];
-                        labels.notPublished = data[7];
-                        scope.chooseLabel = data[8];
+                        [labels.deleted, 
+                        labels.unpublished, 
+                        labels.published,
+                        labels.publishedPendingChanges,
+                        labels.notCreated,
+                        labels.unsavedChanges,
+                        labels.doctypeChangeWarning,
+                        labels.notPublished,
+                        scope.chooseLabel] = data;
 
                         setNodePublishStatus();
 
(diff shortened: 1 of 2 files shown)


bjarnef:
@nathanwoulfe yeah, I noticed that as well when working on this PR https://github.com/umbraco/Umbraco-CMS/pull/9634

It seems a new scope variable previous was added in controller, but not changed in view.

I tried with just the `title` property which seems to conflict with `title` attribute, so it actually showed a tooltip instead of a text label. Do you know if there's a fix for that? In newer versions of Angular I think the syntax would be `[title={someVariable}]` where the brackets indicate its a property.

nathanwoulfe:
@bjarnef `title` is absolutely a valid attribute for the `box-header` directive, but the issue here was caused by the `title-key` attribute (ie `headerLabelKey`) not existing when the link function in the directive executed. I think the tooltip behaviour is just a coincidence from using title as the attribute name.

The way the directive works it expects the scope properties to exist when link runs, and never binds a watch to manage changes. In newer Angular, the directive would be a component and we'd have access to the ngOnChanges method, which would squash this problem.

Instead, by using the ng-if condition, the box header is only rendered once the title-key string exists, so when link runs, the scoped properties all exist, and the heading string can be localized and rendered.

Angular8 would be `[title]="someVar"` to pass the value of `someVar` as an input to the box-header component. Not sure if that's changed in 9/10/11 (I'm working on a project in 8 at the moment, so it's all front-of-mind).

bjarnef:
@nathanwoulfe yes, it seems the `title` attribute/property for `box-header` also add the `title` attribute on the `<div>` element itself. This is from v8.9.0, but I think it is same behaviour with latest changes in `v8/contrib`.

image

Not a big deal, just wonder if there is a way to control this? Is this specific for `title` attribute or is a similar issue seen for e.g. `name`, `target` and other naming which collide with HTML attributes. In that case I guess the browser might only add `name` attribute to e.g. `<input>`, `<select>`, `<textarea>` while `adding `target` to e.g. `<a>`.

But maybe it is only an issue with element directives/components. Just wondering 🤔😄

nathanwoulfe:
Ah, that's due to the `replace:true` property in the directive definition. It replaces `umb-box-heading` element with the template literal, then does some funky magic to transpose the attributes from the custom element onto the template, hence we get a rendered `title` attribute on the div, as it's the ng-attribute from `umb-box-heading`, before it was replaced.

Something like that, at least.

mikecp (approved):

mikecp:
Thanks @nathanwoulfe 👍

nathanwoulfe:
No, @mikecp, thank you. 👍

PR #9650, opened 2021-01-13
Fix for security vulnerability in log4net v2.0.8 -> Upgrade to v2.0.12

kaizeneer:
According to WhiteSource Bolt there exists a documented security vulnerability in Umbraco v7 regarding the third party library log4net.

Apache log4net versions before 2.0.10 do not disable XML external entities when parsing log4net configuration files. This allows for XXE-based attacks in applications that accept attacker-controlled log4net configuration files.

Solution: Upgrade to version log4net - 2.0.12


nul800sebastiaan (changes requested):
Thanks @kaizeneer! I notice there's a lot of files that shouldn't be there in this commit. 

Could you please revert all the changes in the `build` directory? The only changes I would expect to see are to `csproj` files, `.config` files and then finally we also need to change the dependency here https://github.com/umbraco/Umbraco-CMS/blob/v7/dev/build/NuSpecs/UmbracoCms.Core.nuspec#L19 which should be the only change in the `build` directory.

nul800sebastiaan:
Thanks again @kaizeneer - we had to get a new release out, so I've fixed this here: https://github.com/umbraco/Umbraco-CMS/commit/13af8d280919813dbb2b7b76ab03f5ea0b2c6743 

I'll close this one, thanks for the help!

Issue #9564, opened 2020-12-16
Umbraco 8.9.1 Infinite Editing, document type template list - fonts are oversized and truncated


jattwood:
In Umbraco 8.9.1 using Infinite Editing. The template picker is not formatted correctly leading to oversized fonts and truncated template names.

## Umbraco version

I am seeing this issue on Umbraco version: 8.9.1

### Specifics

![image](https://user-images.githubusercontent.com/6452616/102418514-e0f51080-3fcb-11eb-95e3-ff4ef846c514.png)

### Steps to reproduce

- Open any content Item from the content section
- go to info tab
- open the doc type,
- click on templates for the document type
- click on choose extra template

NOTE: It appears it is like this when accessing through settings as well.

Result: fonts all oversized so template names are unreadable/truncated.

### Expected result
Expect the template list formatting to appear as it would on other parts of the back office with smaller fonts.
Something like this:

![image](https://user-images.githubusercontent.com/6452616/102508985-03327100-4054-11eb-9dbd-b9fb2a820884.png)

`.btn-reset {
    padding: 0;
    margin: 0;
    border: none;
    background: 0 0;
    color: currentColor;
    font-family: Lato,Helvetica Neue,Helvetica,Arial,sans-serif;
    font-size: 11px;
    line-height: 12px;
    cursor: pointer;
}`

and

`.umb-card-grid .umb-card-grid-item>span {
    position: absolute;
    top: 10px;
    bottom: 10px;
    left: 10px;
    right: 10px;
    display: flex;
    align-items: center;
    justify-content: flex-start;
    flex-direction: column;
    background-color: transparent;
    word-break: break-word;
}`

### Actual result
Result: fonts all oversized so template names are unreadable/truncated.

nul800sebastiaan:
Thanks, I'll link this to https://github.com/umbraco/Umbraco-CMS/issues/9394

nul800sebastiaan:
This is now fixed with PR https://github.com/umbraco/Umbraco-CMS/pull/9651, it will be available in patch releases for v8.8 and above, coming next week and also for 8.11.0 which is also coming out next week.

PR #9651, opened 2021-01-13
fixes: #9564: Infinite Editing, document type template list - fonts are oversized and truncated

madsrasmussen:
I have updated the CSS for the item picker. It now supports items with long titles. With the updated styles the picker now looks like this:

Screenshot 2021-01-13 at 11 51 49

We use it in a couple of places so we need to make sure that these still work/look as expected.

Where and how to test:
- Add some doc types with long titles
- Add some data types with long titles
- Make sure adding "Allowed child nodes", "properties", and templates to a Doc Type still looks correct
- Make sure adding and copying nested content items still looks correct.
- Make sure adding questions to a form still looks correct.

fixes: #9564

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/card.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/card.less
@@ -104,19 +104,18 @@
 	margin: 0 auto;
 	list-style: none;
     width: 100%;
-
 	display: flex;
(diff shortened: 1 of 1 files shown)


abjerner:
MICHAEL SHEEN NOD

nul800sebastiaan:
Thanks Mads! 👍

Issue #9636, opened 2021-01-11
Redirect URLs broken as of 8.10.0


zakinaeem-arq:
Redirect URLs don't work on 8.10.0+

## Umbraco version

I am seeing this issue on Umbraco version: 8.10.0 as well as 8.10.1

Reproduction
------------

### Bug summary

Renaming a node via Backoffice (regardless of the Doc type) adds a redirect to old URL of the page, as expected. This however results in the old URL returning a 404. This also means that if you had any redirects as a result of node renaming, upgrading to 8.10.0 or 8.10.1 will break them.

This happens on an existing project as well as a clean Umbraco install.

### Steps to reproduce

* Create a new node (any Document type) and name it "Sample page"
* Save and publish the node
* Confirm the node is accessible on "/sample-page/"
* Rename the "Sample page" node to "Sample page renamed"
* The node should now be accessible on "/sample-page-renamed/" and a redirect should automatically be added mapping "/sample-page/" to "/sample-page-renamed/". This can be verified in the "Info" tab of the node

### Expected result
* Accessing "/sample-page/" should redirect the user's browser to "/sample-page-renamed/"

### Actual result
* "/sample-page/" returns a 404

radospanny:
same to us

zakinaeem-arq:
Some findings:

I believe this is related to the updates introduced to handle the following issue https://github.com/umbraco/Umbraco-CMS/issues/8568

In a default setting, website would have a Default language however the redirects persisted in "umbracoRedirectUrl" table don't have a "Culture" value associated to them. Based on the updates [here](https://github.com/umbraco/Umbraco-CMS/pull/9191/commits/e4706fd4176852bc5ca1bd55ecfc6250e66e79d1), this would result "PublishedRequest" culture in **ContentFinderByRedirectUrl.cs** to be that of Default language, which is passed down to the "RedirectUrlService.cs" and eventually "RedirectUrlRepository.cs" resulting in zero matches.

I can handle this temporarily with a custom implementation of IContentLastChanceFinder but that would mean yet another step for the application in a fallback scenario. Looking for suggestions if there is a cleaner way to handle this until fixed.

dawoe:
Hi @zakinaeem-arq 

I think this can be solved with updating the umbracoRedirects table by setting the culture column to the default culture. But Ideally this should be handled by the upgrade.

Dave

zakinaeem-arq:
[...]

PR #9653, opened 2021-01-13
[AMEND] Updated the redirect URL Repository to search for entries wit…

NikRimington:
This fix is to address issue #9636 . If a search based on culture fails to return any redirects it will search for any that have no culture defined against them.

I feel that an additional unit test is probably required, but I wasn't sure how to set up the data for the test so have left this off.

I did consider falling back to the original method that doesn't look at culture but I felt that using "no culture" would be a better approach.

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/RedirectUrlRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/RedirectUrlRepository.cs
@@ -214,25 +214,16 @@ JOIN umbracoNode ON umbracoRedirectUrl.contentKey=umbracoNode.uniqueID");
             if (string.IsNullOrWhiteSpace(culture)) return GetMostRecentUrl(url);
             var urlHash = url.GenerateHash<SHA1>();
             var sql = GetBaseQuery(false)
-                .Where<RedirectUrlDto>(x => x.Url == url && x.UrlHash == urlHash && x.Culture == culture.ToLower())
+                .Where<RedirectUrlDto>(x => x.Url == url && x.UrlHash == urlHash &&
+                    (x.Culture == culture.ToLower() || x.Culture == string.Empty))
                 .OrderByDescending<RedirectUrlDto>(x => x.CreateDateUtc);
             var dtos = Database.Fetch<RedirectUrlDto>(sql);
-            var dto = dtos.FirstOrDefault();
+            var dto = dtos.FirstOrDefault(f => f.Culture == culture.ToLower());
 
             if (dto == null)
-                dto = GetMostRecentUrlWithNoCulture(url);
+                dto = dtos.FirstOrDefault(f => f.Culture == string.Empty);
 
             return dto == null ? null : Map(dto);
-        }
-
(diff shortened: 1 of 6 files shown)


Shazwazza:
@NikRimington there's an existing `RedirectUrlRepositoryTests`, would you be able to add tests there?

NikRimington:
Hi @Shazwazza I've added in a new test for the fallback behaviour which appears to be passing. I also cleaned up the tests a little bit so variables had more sensible names as I think it added to my confusion on how to create the new test. Hope that's okay :-)

src/Umbraco.Core/Persistence/Repositories/Implement/RedirectUrlRepository.cs
```diff
@@ -218,7 +218,21 @@ public IRedirectUrl GetMostRecentUrl(string url, string culture)
...
+
+            if (dto == null)
+                dto = GetMostRecentUrlWithNoCulture(url);
+
             return dto == null ? null : Map(dto);
         }
+
+        private RedirectUrlDto GetMostRecentUrlWithNoCulture(string url)
+        {
+            var urlHash = url.GenerateHash<SHA1>();
```
Shazwazza, line 230: This is regenerating a hash that was already generated. The url hasn't changed so no need to allocate and add overhead hashing again.

src/Umbraco.Core/Persistence/Repositories/Implement/RedirectUrlRepository.cs
```diff
@@ -218,7 +218,21 @@ public IRedirectUrl GetMostRecentUrl(string url, string culture)
...
+            var sql = GetBaseQuery(false)
+                .Where<RedirectUrlDto>(x => x.Url == url && x.UrlHash == urlHash && x.Culture == string.Empty)
```
Shazwazza, line 232: Is there any reason why we can't 'just' have the single query above check `x.Culture == culture.ToLower() || x.Culture == string.Empty)` ? Would be better to have less queries if possible.

Shazwazza (changes requested):
Have added a couple note and questions

src/Umbraco.Core/Persistence/Repositories/Implement/RedirectUrlRepository.cs
NikRimington: This function has now been removed as isn't needed any more :-)

src/Umbraco.Core/Persistence/Repositories/Implement/RedirectUrlRepository.cs
NikRimington: I've combined it into a single DB query, but this means there is filtering on the Linq expression (FirstOrDefault) now. I hope this is okay?

nul800sebastiaan:
Thanks @bjarnef - I've cherry-picked for 8.10 and 8.11 👍

nul800sebastiaan:
Duh.. Wrong PR ☝️ 

I've tested this and it now works. I've cherry-picked for 8.10 and 8.11 👍
Thanks for the excellent help @NikRimington! 🎉

PR #9656, opened 2021-01-14
Adjust minimum width of main content

bjarnef:
### Prerequisites

### Description
When using a laptop width a viewport size on 1280 × 720 px the sidebar in e.g. content and media is moved below the main content. This is because the main content is set to a minimum width on 500px where the sidebar flow under the main content.

In packages section this in fine, because the navigation area is not shown here, but in content and media I think it would be great to adjust this a bit, so the sidebar is shown on right with the default width of navigation area on 360 px on since 1280px is a pretty standard screen width.

**Before**

image

**After**

image


nathanwoulfe (approved):
Makes sense to me, thanks @bjarnef 

Would be good to rename these elements, since they're being used outside the packages context. I know it has been discussed elsewhere, just bugs me every time I see it

PR #9657, opened 2021-01-14
Move relation type id and macro id to sidebar box

bjarnef:
### Prerequisites

### Description
In backoffice we mostly have id + guid show in an sidebar box, e.g. in content, media, member group, datatype, etc.
So I think it would be a bit more consistent to move this under relation types as well.

Note part of this PR relies on the following PRs as well:
https://github.com/umbraco/Umbraco-CMS/pull/9632
https://github.com/umbraco/Umbraco-CMS/pull/9656

**Before**

image

**After**

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/macros/views/settings.html
+++ b/src/Umbraco.Web.UI.Client/src/views/macros/views/settings.html
@@ -1,62 +1,76 @@
 ﻿<div ng-controller="Umbraco.Editors.Macros.SettingsController as vm">
-    <umb-box>
-        <umb-box-header title="Info"></umb-box-header>
-        <umb-box-content>
-            <umb-control-group label="Id">
-                {{model.macro.id}}<br/>
-                {{model.macro.key}}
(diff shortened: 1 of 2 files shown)


bjarnef:
I have adjusted this for relation types as well. Then macro view label is however hidden, but there might be a better way to handle this.

**Before**

**After**

image

bjarnef:
Off topic: Since it started with packages section, but these classes now are used throughout the backoffice and in packages, we should really rename these + keep the existing ones for backward compatibility.

`umb-package-details, `umb-package-details__main-content` and `umb-package-details__sidebar`.

Furthermore it would be great if the sidebar could both be aligned to left and right. Currently it only works well on the right.

nathanwoulfe:
@bjarnef that makes good sense - consistency is king.

Also agree with renaming the elements. A bit of flexbox magic should make the alignment work nicely

nul800sebastiaan:
Looking good, thanks @bjarnef ! 👍

Issue #9663, opened 2021-01-15
Blank block overlay after switching language


bjarnef:
I noticed when having a document type which allow to vary by culture and a property e.g. "Sections" using Blocklist editor the block overlay sometimes is blank after switching language.

It only seems to happen when you have opened an overlay on the current language, switch language and edit a block element here.

**Steps to reproduce**
- Allow a document to vary by culture
- Add a property, e.g. sections, which allow to vary by culture and use Block List editor.
- Create e.g. a "Text" element type which is added to Block List editor configuration.
- Create a two languages
- Add a Text Block to Block List editor for language 1 and 2
- On language 1 open the block item and close the overlay again.
- Switch to language 2 via language selector in header
- On language 2 open the block item
- The overlay is now blank
- If not seing this try switching back to language 1 and open the block item overlay here.

![image](https://user-images.githubusercontent.com/2919859/104732098-0137fc80-573d-11eb-8a68-16021b57aead.png)

However after refreshing page for language 2 and open the block item overlay, the properties are shown again.

Seen on Umbraco v8.9.0

---
_This item has been added to our backlog [AB#10781](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10781)_

bjarnef:
I can reproduce this as well with latest changes in `v8/contrib` branch.

bjarnef:
@nielslyngsoe @nathanwoulfe I wonder if you guys have noticed this issue and have an idea what goes wrong? I can't see any specific console error, so I guess something is missing in the data being passed in to blockeditor inifinite editor or in the inifinite editor itself.

bjarnef:
I also noticed we have a few references to `umbTabs`, but where does this exist?
I searched a bit around and found we previous had this directive:
https://github.com/umbraco/Umbraco-CMS/pull/2205/files#diff-ff6ca671d28462c72c38cec25b88de820db1207fd7f1feebf24de0037e98a749

But I can't find this anywhere with the latest changes:
https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/tabs

However we still have a few references:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbautoresize.directive.js#L4

[...]

PR #9664, opened 2021-01-15
Block editor cleanup

bjarnef:
### Prerequisites

### Description
While looking at https://github.com/umbraco/Umbraco-CMS/issues/9663 I found a few things in block list editor to cleanup.
- Semi colons after named functions
- Use `<umb-editor-container>` component instead of hard coded markup.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/blockeditor/blockeditor.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/blockeditor/blockeditor.controller.js
@@ -1,6 +1,7 @@
 angular.module("umbraco")
     .controller("Umbraco.Editors.BlockEditorController",
         function ($scope, localizationService, formHelper, overlayService) {
+
             var vm = this;
 
             vm.model = $scope.model;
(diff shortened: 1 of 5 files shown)


nul800sebastiaan (changes requested):
Thanks @bjarnef! All of this cleanup is good! However, it doesn't actually fix the issue, not sure if I am missing something? I rebuilt everything a couple of times and cleared all caches just to make sure! 😅

bjarnef:
@nul800sebastiaan you are right. I didn't find a fix for this for now, so I just did some minor adjustments.

Will investigate it further and submit a new PR if I find a way to fix this 👍

nul800sebastiaan:
Ah alright, then I'm not crazy! 😂 

Great, all of this cleanup is perfectly fine of course, I'll merge this part, thanks again!

bjarnef:
@nul800sebastiaan okay, it seems it has something to do with the `page.loading` variable.

If I remove this part it is showing the view content, but the load indicator seems to spin forever.
https://github.com/umbraco/Umbraco-CMS/blob/d966e89fd97edb9986516259456465f2fad03261/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/blockeditor/blockeditor.html#L5

PQNlF8nJpL

Issue #9665, opened 2021-01-15
Icons missing from umb-node-preview


bjarnef:
It seems icons in some cases are missing from `<umb-node-preview>` component. Tested with latest changes in `v8/contrib`.

For example in users section.

![image](https://user-images.githubusercontent.com/2919859/104768412-e3cf5680-576d-11eb-907c-f0902c3dc224.png)

bjarnef:
It seems I removed the `{{icon}}` from class here, because I thought it wasn't necessary:
https://github.com/umbraco/Umbraco-CMS/commit/a9a4b0894a626adaea407f3a2bab4a6bca77e5f6#diff-6a8c6488e5f6832a8854541a572ad9893f004e32733a52fe2591fec34a0344de

However I found out it IS necessary currently, when the icon isn't available as SVG icon:
https://github.com/umbraco/Umbraco-CMS/pull/9255

PR #9666, opened 2021-01-15
Add back icon name in class attribute to work again with font icons

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9665

### Description
I noticed in users section some icons were missing 🙈 

image

I figured out that I removed the icon name from the class here, because I thought it wasn't necessary:
https://github.com/umbraco/Umbraco-CMS/commit/a9a4b0894a626adaea407f3a2bab4a6bca77e5f6#diff-6a8c6488e5f6832a8854541a572ad9893f004e32733a52fe2591fec34a0344deL3

Now it works again!

image

Mostly you won't notice in core, because we have SVG icons for most of the old font icon classes. But currently it IS necessary currently, when the icon isn't available as SVG icon: https://github.com/umbraco/Umbraco-CMS/pull/9255

At least until we find a better way to handle this and since many packages are using `<umb-node-preview>` component and probably with font icons since that was the original approach.


bjarnef:
@nul800sebastiaan @nathanwoulfe maybe you can have a look at this? 👍

nul800sebastiaan:
Thanks @bjarnef - I've cherry-picked for 8.10 and 8.11 👍

Issue #9349, opened 2020-11-05
MNTP for node type Media, when allowing items of a certain type, files of different types can be selected


Jverschoor:
I want to make a picker that only allows videos to be picked. I created a video media type that only allows mp4 files and selected this type as only allowed type for this picker.
Now if I want to pick a video, everything works fine.
![image](https://user-images.githubusercontent.com/23473606/98229452-39aba500-1f5a-11eb-8f73-a20ffd36e72d.png)
media of type video are allowed.
![image](https://user-images.githubusercontent.com/23473606/98229514-4af4b180-1f5a-11eb-996f-0711b40eaf93.png)
Images are not allowed.

But,
if I use the search bar to search for an image I'm allowed to select it and save the newly created document
![image](https://user-images.githubusercontent.com/23473606/98229606-69f34380-1f5a-11eb-9da1-c82d129c3abb.png)
![image](https://user-images.githubusercontent.com/23473606/98229735-927b3d80-1f5a-11eb-8248-221a9f28d859.png)

## Umbraco version

I am seeing this issue on Umbraco version: 8.9.0

Reproduction
------------

### Bug summary

When I create a mutlinode tree picker with node type Media that only allows a specific media type. When I create a document and try to add media using this picker I can select any type of media by using the search bar.

### Steps to reproduce

Create a media type and give it a upload field
![image](https://user-images.githubusercontent.com/23473606/98229948-dbcb8d00-1f5a-11eb-8e4f-80f66484ff06.png)
Create a document type and give it a multinode tree picker as property editor
this picker must have Media as node type and only allow your newly created media type as allowed type as below:
![image](https://user-images.githubusercontent.com/23473606/98230162-23521900-1f5b-11eb-9a35-2348e56ebfbc.png)
Add new media of type other than the newly created media type (f.e. an image)
![image](https://user-images.githubusercontent.com/23473606/98231287-82fcf400-1f5c-11eb-918c-209f32b0b9a6.png)
Create a new document of the type you just created
![image](https://user-images.githubusercontent.com/23473606/98231322-914b1000-1f5c-11eb-8a75-0b24714faaa1.png)
Click on add to add media from your tree picker
![image](https://user-images.githubusercontent.com/23473606/98231352-9ad47800-1f5c-11eb-8fc5-5005db29e41d.png)
Enter the file name of the media type that was not allowed to be selected.
Select the file and save/publish the document.
[...]

PR #9668, opened 2021-01-16
fix for issue #9349

AndreyKarandashovUKAD:
The problem is described here: https://github.com/umbraco/Umbraco-CMS/issues/9349


nul800sebastiaan:
Hi @AndreyKarandashovUKAD! Apologies for the delay and thanks for looking into this! Unfortunately this is not going to fix the problem, we're fine with having all of the nodes that are not available in the list, we just don't want them to be selectable after a search happens that returns types that are not allowed.

As you can see in the following gif, the items are displaying but not available to pick, unless you do a search:



This is being doing in javascript by flipping the `item.selectable` variable to false. So the search here needs to know about what's allowed as well. So for this to be fixed that logic needs to be implemented for the search results as well.  

I'll close this one, thanks again!

PR #9669, opened 2021-01-17
Minor optimisation by passing int instead of int? to Umbraco.Content()

nzdev:
### Prerequisites

If there's an existing issue for this PR then this fixes NA

### Description
Minor optimisation by passing int instead of int? to Umbraco.Content. 
Pass int to Umbraco.Content instead of int? given we know _contentId has value.

### DIFF
--- a/src/Umbraco.Web/Mvc/EnsurePublishedContentRequestAttribute.cs
+++ b/src/Umbraco.Web/Mvc/EnsurePublishedContentRequestAttribute.cs
@@ -103,7 +103,7 @@ namespace Umbraco.Web.Mvc
         {
             if (_contentId.HasValue)
             {
-                var content = Umbraco.Content(_contentId);
+                var content = Umbraco.Content(_contentId.Value);
                 if (content == null)
                 {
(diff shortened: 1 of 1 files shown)


src/Umbraco.Web/Mvc/EnsurePublishedContentRequestAttribute.cs
```diff
@@ -103,7 +103,7 @@ protected virtual void ConfigurePublishedContentRequest(PublishedRequest request
         {
             if (_contentId.HasValue)
             {
-                var content = Umbraco.Content(_contentId);
+                var content = Umbraco.Content(_contentId.Value);
                 if (content == null)
                 {
                     throw new InvalidOperationException("Could not resolve content with id " + _contentId);
```
bjarnef, line 109: @nzdev @Shazwazza should it use `_contentId.Value` in `InvalidOperationException` as well?

Shazwazza: That's not going to have any perf impact

Issue #9667, opened 2021-01-15
Block List Editor (Nested) Crashes and unresponsive.


robjakedorsett:
I'm having an issue where the back office is completely crashing when I'm trying to populate a block list editor.

This is happening at around 5 levels of nesting within (though I have been able to get it to happen at 4). 
There aren't any console logs, or network errors happening. 

## Umbraco version

I am seeing this issue on Umbraco version: 8.10.1 & 8.9.1

Steps to Reproduce:
- Create a Doc type - HomePage with a picker for Block List "Tabbed Content" 
- Create a Doc type  (Element type) - Tabbed Content, with a picker for Block List "Tab"
- Create a Doc type (Element type) - Tab, with a picker for Block List "Full Width Layout"
- Create a Doc type (Element type) - Full WIdth Layout with a picker for "Component"
- Create a Doctype (Element type) - "Component" with a picker for "Component Item"
- Create a Doctype (Element type) - "Component Item" with the properties - Url Picker (Mandatory) & Media Picker (Mandatory)

The Block list configs are all default and empty, excluding the "Content Model" ofcourse

Also worth noting that all the doc types have enabled varient by culture

robjakedorsett:
![image](https://user-images.githubusercontent.com/47298040/104789303-931b2600-578c-11eb-9338-deac772aa99d.png)

nathanwoulfe:
@robjakedorsett can you provide some more detail - steps to reproduce, your block config, any console errors etc. Anything that might be useful for tracking down the cause and determining if it's a bug or an implementation issue.

robjakedorsett:
Hi @nathanwoulfe I have editted my original comment to include replication steps and block list configs ( they are all default)

robjakedorsett:
I have been playing around with this, this morning, and I have discovered that it seems to occur when properties are set to mandatory,  So I assume it's got something to do with the validation javascript somehow

nathanwoulfe:
Magic, thanks @robjakedorsett. Definitely looks a bit busted in the gif...

nul800sebastiaan:
I think this is the same problem as https://github.com/umbraco/Umbraco-CMS/issues/9360

@nielslyngsoe & @madsrasmussen can you confirm that the PR for that one also fixes this problem please?

nul800sebastiaan:
[...]

PR #9677, opened 2021-01-18
Fixes: 9360: Nested validation freezes browser

### LINKED ISSUE #9360: Nested validation freezes browser
When you have a property editor that allows nesting of some sort (e.g. NestedContent) with at least 1 mandatory property the browser tab will completely freeze when Saving/Publishing when there are validation messages to display.

This seems to be caused by [`collectAllFormErrorsRecursively`](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/services/angularhelper.service.js#L13) of the AngularHelper service. This one is executed an insane amount of times (100K+, but I killed the tab at that point so it will be much more). It seems to increase exponentially when nesting increases.

Demo where I increase nesting. All fine at first but gets slower each step. At level 4 it just hangs forever (probably depends on hardware). The UI also starts to slow down noticably at the step before that. There are no console errors. It's just calling that `collectAllFormErrorsRecursively` function way too many times -- without a result apparently.
slow

When I disable `collectAllFormErrorsRecursively` (by placing a `return` immediately at the start), it runs flawlessly however deep you nest and also does not seem to break anything. So what is this function even used for :-)
fast

This was tested on a new installation of 8.9 with an empty site and only 2 doctypes (Home + Block, where Block is the doctype used in NestedContent).

## Umbraco version

I am seeing this issue on Umbraco version: 8.9.0

### Steps to reproduce

1. Create some simple Nested Content doctype with at least 1 mandatory property and a property referencing itself (to nest it).
2. Create a datatype for this Nested Content
3. Add this data type to some doctype.
4. Add some content and make sure to nest it a couple of times (4+)
5. Keep the mandatory properties empty
6. Publish
7. Observe hanging browser

### Expected result

Validation messages show within a reasonable time :)

### Actual result

Browser tab hangs and does not respond anymore.

madsrasmussen:
This PR fixes the issue where the browser crashes when validating complex nested content and block list editors.

How to test:
- Follow guide from issue #9360 to set up nested content
- Make sure validation messages are shown and updated correctly for all nested content levels.
- Make sure you can nest "nested content" more than 3 levels.

fixes #9360 

---
_This item has been added to our backlog [AB#10061](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10061)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/angularhelper.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/angularhelper.service.js
@@ -11,6 +11,11 @@ function angularHelper($q) {
     var requiredFormProps = ["$error", "$name", "$dirty", "$pristine", "$valid", "$submitted", "$pending"];
 
     function collectAllFormErrorsRecursively(formCtrl, allErrors) {
+        // skip if the control is already added to the array
+        if (allErrors.indexOf(formCtrl) !== -1) {
+            return
+        }
+
         // loop over the error dictionary (see https://docs.angularjs.org/api/ng/type/form.FormController#$error)
         var keys = Object.keys(formCtrl.$error);
         if (keys.length === 0) {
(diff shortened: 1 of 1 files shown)


nielslyngsoe:
fixes https://github.com/umbraco/Umbraco-CMS/issues/9667

nielslyngsoe:
All good, performance is good again.

PR #5607, opened 2019-06-07
Temp 5596 grid edit alttext

nathanwoulfe:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

Refer to #9681 for a fresh take on this PR - this was 18 months old, lot has changed since then...

### Description
Current editing flow when adding an image in the grid doesn't allow the editor to easily update the alt text - it is set when adding the image, and can only be modified by removing and re-adding the image, which means losing any value set for the caption.

This change modifies the grid media editor to show only the image (no more caption text), with two new buttons - one launching the media picker dialog to select a new image, the other opening the media picker details dialog to edit the caption and alt text.

The details dialog was previously inline in the mediapicker view, which made reuse impossible. I've split it out into a new infinite editor and made it available on editorService as .mediaPickerDetails(). With that change in place I've also updated the mediapicker to use the infinite editor when placing an image in the grid.

PR #9681, opened 2021-01-19
5596 allow editing alt text in grid

### LINKED ISSUE #5596: V8 - Grid - allow editing image alt text
Quick win - when adding an image control on the grid, editors can optionally provide alt text when selecting their image. What they can't do is edit that alt text without removing the image, reselecting, and providing new text.

Can be easily remedied by adding an additional input to the media grid view, below/above the existing one for the image caption.

Will have pull request shortly.

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #5596 

This PR replaces the stale one in #5607 (which I will close)

### Description
Changes here separate the alt text and caption into distinct values - previously the text set as alt text was stored as the caption, which doesn't make sense. To make this work, I've added two buttons to the grid media view, one to change the existing image, one to edit the existing image. 

Edit-About-Us---Content---localhost

Previously clicking the existing image opened the media picker overlay, which then opened a nested overlay to edit the details. That wasn't the nicest UX as closing the details overlay then returned to the picker, which is unrelated to the action the user had taken. Separating the two actions makes it clear which action will be triggered, and by adding buttons introduces text into the UI, making it more obvious than clicking the unlabeled image.

Adding a new image still triggers the details overlay as before.

No data is automatically migrated (alt text is still stored as the caption), but editing an existing image displays the alt text in the caption field, so should be obvious that it needs to be manually updated. Is it a breaking change? Potentially, since editing and saving an existing image will shift the alt text to the caption, which may cause rendering issues.

Also adds a new function to editorService => mediaCropDetails. I will also update the media picker controller to use this method rather than an inline view reference. Tidied up some existing JS, with whitespace fixes and syntax improvements.

Still needs translations for a new key, will fire up Google Translate and see how I go...

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/media.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/media.controller.js
@@ -1,10 +1,8 @@
 angular.module("umbraco")
     .controller("Umbraco.PropertyEditors.Grid.MediaController",
-    function ($scope, $timeout, userService, editorService) {
+    function ($scope, userService, editorService, localizationService) {        
         
-        
-        $scope.thumbnailUrl = getThumbnailUrl();
-        
+        $scope.thumbnailUrl = getThumbnailUrl();        
         
         if (!$scope.model.config.startNodeId) {
             if ($scope.model.config.ignoreUserStartNodes === true) {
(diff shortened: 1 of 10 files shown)


src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/media.controller.js
```diff
@@ -1,112 +1,129 @@
...
-                    image: value.image,
-                    focalPoint: value.focalPoint,
-                    coordinates: value.coordinates
-                }
-                : null;
+
             var mediaPicker = {
                 startNodeId: startNodeId,
-                startNodeIsVirtual: startNodeIsVirtual,
+                startNodeIsVirtual: startNodeId ? $scope.model.config.startNodeIsVirtual : undefined,
```
bjarnef, line 25: @nathanwoulfe I think it should be `null` instead of `undefined` here since you are assigning a value to the property.

bjarnef:
@nathanwoulfe wouldn't it be simpler with a single edit button or just open edit mode when clicking the container? At the image preview it could show an edit button to open to media crop details overlay.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/media.controller.js
nathanwoulfe: yup, it should be. That was a bit of refactoring to get rid of an unnecessary variable, didn't check that the statement made sense. Will update.

bjarnef: Cool, I guess it is the same case with `startNodeId` variable.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/media.controller.js
```diff
@@ -1,112 +1,129 @@
...
+        $scope.editImage = function() {               
+
+            const mediaCropDetailsConfig = {
+                size: 'small',
+                target: $scope.control.value,
+                submit: model => {
+                    updateControlValue(model.target);
+                    editorService.close();
+                },
+                close: function () {
```
bjarnef, line 50: @nathanwoulfe any reason you don't use the new format `()` for close function, but for submit function? 😁

nathanwoulfe:
@bjarnef my issue with a single button is that to get to the details dialog, we have to go via the media picker, which doesn't make sense if we only want to update caption/alt/focal point.

The problem I have with that is when a user closes the details overlay, they're returned to the media picker. That doesn't make sense since they've chosen to edit the existing image, but are now presented with the UI for choosing a new image. It's also an extra click to submit the changes, in an overlay where they haven't otherwise interacted.

By separating the buttons, it's two distinct actions, one to change the image, one to update the existing. Each button triggers an appropriate overlay (changing image will show the details overlay after selecting the new image).

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/media.controller.js
nathanwoulfe: no good reason... I'll blame the lack of javascript linting...

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/media.controller.js
nathanwoulfe: I've updated to use null 👍

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/editors/media.controller.js
nathanwoulfe: Tidied these up too

bjarnef:
@nathanwoulfe okay, it just seems grid editors are handling edit very different and no consistent way. But maybe the grid editor need an overhaul itself. It could maybe have actions similar to Block List editor. Not sure if there are some quick wins or HQ plan to create a Grid v2 (there has been a `grid2` branch for a long time https://github.com/umbraco/Umbraco-CMS/tree/temp8-grid2 but not sure what the status on this is). Or if the focus now more is on Block List editor and maybe a Block Grid editor in  a future release.

mikecp:
Hi @nathanwoulfe,

Great PR, thanks :-)

Maybe two questions before merging: is there a specific reason why you removed the display of the caption text under the image in editor view (cf line 16 in modified media.html file)?

And for the second question: when clicking an existing image, a small menu appears on top with a delete button, and then there are the two buttons you added at the bottom of the image. I figured this might be confusing for the editor so I was wondering if it could be worth displaying the two action buttons on top as well, or even be merged next to the "delete" icon?  

What do you think?

PS: I also solved a merge conflict + added French translation ;-)

nathanwoulfe:
Hi @mikecp, re the caption, I think my thinking was that if there are now two editable text fields (caption and alt text), it might be confusing to only display one. It's heading more down the block-list (or DTGE) editing style, where the overlay houses all the fields. Easy enough to add it back if it's warranted.

As for the buttons, the delete at the top is part of the grid view, I don't believe there's any way to add buttons to that zone for a specific editor (although that would be a really nice option). Can certainly shift the two buttons to the top though, it does make sense to keep the controls together. Makes the buttons more obvious too since there's movement at the top of the control when the blue header bar is displayed.

mikecp:
Hey @nathanwoulfe ,
It would indeed be strange to have only one of the two fields displayed, so let's keep it as it is :-)

As for the button, I was more or less expecting it to be not obvious to put in the "gird view menu" ;-), but I think it would indeed be more obvious to put them at the top.

nathanwoulfe:
@mikecp sorry, I thought I'd pushed my update to shift the buttons - it's definitely better at the top. I have a feeling I might have trashed the changes but easy enough to redo. Will update.

mikecp:
@nathanwoulfe Great, I'll be waiting for them ;-)

nathanwoulfe:
Done - buttons are now top-right.

mikecp:
Thanks for the update @nathanwoulfe 👍 
It looks indeed better on top ;-)

mikecp (approved):

TimWensink:
@mikecp @nathanwoulfe We have just updated our site to Umbraco 8.12 hoping to see this fixed.
However it doesn't really work as intended and as it is demo'd in the screencap in the description above.

I don't see the two buttons ('Select media' and 'Edit') over existing media and when i click on the image both the overlays are opened on top of eachother. In the  'edit selected media' version the new caption field that was supposed to be added is missing as well.

Does you see the same issue occur?
Could there have been a merge conflict/issue or is something missing? Or perhaps am I doing something wrong?

nathanwoulfe:
@TimWensink have you bumped the client dependency version number and ensured the backoffice cache has cleared? If the caption field is missing entirely, that sounds like you're seeing the old view from cache...

TimWensink:
@nathanwoulfe We have cleared all the cache, even restarted the entire website on the server, tried multiple browsers, and still don't get the new view. Also the clientDependency is increased when we applied the upgrade. 

This is what we currently see: 
afbeelding

Can you verify that it works in 8.12 install the same as your pull request you have made? 
We have done the same as we usually do with every upgrade. Other features in this release are visible and there has been a change regarding editting the alttag, since now both popovers open simultaneously. 

If it works with you in 8.12 I need to do further digging into our project, but I still don't really see what settings could have been overlooked to be fair.

nathanwoulfe:
Hi @TimWensink I've just done a clean 8.12 install, this feature works as advertised. I'd check that the html views for the media picker and grid are up to date, and same with the core javascript bundles. Possible that the install skipped a file due to local modifications...

TimWensink:
@nathanwoulfe After some research it turns out that there's an issue in our release. The error only occurs 'online', on the server. Locally there's no issue at all. Sorry for the confusion.

PR #9682, opened 2021-01-19
Netcore: Add support for sync notification handlers in event aggregator

bergmania:
### Details
- Added support for sync notification handlers in the `EventAggregator`
- Updated async handlers that was not using the async parts to be sync.

### Test
- Everything should work as before. E.g Set a breakpoint in one of the sync handlers and see it breaks as expected

### DIFF
--- a/src/Umbraco.Core/DependencyInjection/UmbracoBuilder.Events.cs
+++ b/src/Umbraco.Core/DependencyInjection/UmbracoBuilder.Events.cs
@@ -2,7 +2,6 @@
 // See LICENSE for more details.
 
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.DependencyInjection.Extensions;
 using Umbraco.Core.Events;
 
 namespace Umbraco.Core.DependencyInjection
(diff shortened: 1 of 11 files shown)


src/Umbraco.Core/DependencyInjection/UmbracoBuilder.Events.cs
```diff
@@ -35,5 +34,29 @@ public static partial class UmbracoBuilderExtensions
...
+        /// <param name="builder">The Umbraco builder.</param>
+        /// <returns>The <see cref="IUmbracoBuilder"/>.</returns>
+        public static IUmbracoBuilder AddNotificationAsyncHandler<TNotification, TNotificationAsyncHandler>(this IUmbracoBuilder builder)
+            where TNotificationAsyncHandler : INotificationAsyncHandler<TNotification>
+            where TNotification : INotification
+        {
+            // Register the handler as transient. This ensures that anything can be injected into it.
+            var descriptor = new ServiceDescriptor(typeof(INotificationAsyncHandler<TNotification>), typeof(TNotificationAsyncHandler), ServiceLifetime.Transient);
+
+            // TODO: Waiting on feedback here https://github.com/umbraco/Umbraco-CMS/pull/9556/files#r548365396 about whether
```
Shazwazza, line 52: Don't need this comment now, it was approved.

Shazwazza (changes requested):
But is it possible to also have a sync API for publishing? In some cases we don't have any async context to publish from

src/Umbraco.Core/Events/EventAggregator.cs
```diff
@@ -41,6 +41,20 @@ public Task PublishAsync<TNotification>(TNotification notification, Cancellation
...
+            where TNotification : INotification
+        {
+            // TODO: Introduce codegen efficient Guard classes to reduce noise.
+            if (notification == null)
+            {
+                throw new ArgumentNullException(nameof(notification));
+            }
+
+            PublishNotification(notification);
+            Task.WaitAll(PublishNotificationAsync(notification));
```
Shazwazza, line 56: I think (hope) in our case this will be ok but Task.WaitAll can also deadlock if called from a UI thread or a threadpool thread (which will most likely be us) if the threadpool is exhausted (but in that case i think we're screwed anyways) or also if this is called from an async method in a threadpool thread (which we'd hope isn't done). 

I think we're ok to leave this but make sure we don't use this one when already in async contetx.

Shazwazza (approved):

Issue #9683, opened 2021-01-19
Umbraco radio buttons not checked by default (on load)


zspasojevic:
We're implementing some custom cms section and trying to use umbraco radio buttons there.
The default button selection is not working on page load despite the fact that is has model="true" set properly. 

## Umbraco version

Seems like all 8+ versions including the current

Reproduction
------------

### Steps to reproduce

Create some custom section in CMS that has a radio button with multiple options in it and set one of them to be selected by default on page load.

### Expected result

To have an option selected by default on page load

### Actual result

None of the radio button options is selected on page load.

We had a look at some of the source files since with umbraco checkboxes this functionality works. We noticed the differences with the localisationService not being passed in inside onLoad() function for radio buttons. Could this be what's causing it to fail?

[Radio Button](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js)

[Checkbox](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbcheckbox.directive.js)

![image](https://user-images.githubusercontent.com/43961181/105060908-33b56280-5a79-11eb-84f1-f3708f48540e.png)

bjarnef:
@zspasojevic strange anyone hasn't noticed this. Yeah, it seems this PR https://github.com/umbraco/Umbraco-CMS/pull/6797 added `localizationService` to the controller, but didn't inject the service.

bjarnef:
@zspasojevic do you use the `label-key` with `<umb-radiobutton>`? 

Also ensure the component has a end-closing tag end not is self-closing (void element).

zspasojevic:
Hi @bjarnef. No, we're not using the label-key, we don't have a need for dictionary/localization. This is a code snippet: 
![image](https://user-images.githubusercontent.com/43961181/105144510-9fd9aa00-5afd-11eb-87ff-c5e84a65d090.png)
So even if we hardcode the model to true, it won't be checked by default.

bjarnef:
Okay, the missing `localizationService` might still affect this. We need to check this with the latest changes, where the missing service is injected.

However I think the issue here is that you set `value="0"` and `value="3"`, but model is `true` and `false` so they will never match.

You can see this example from core, where it is used for relation types:
[...]

PR #6797, opened 2019-10-21
Extend the umb-radio button directive to handle localizations

BatJan:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
The umb-radiobutton directive has been extended so it's possible to pass it a labelkey like it was recently added to the umb-checkbox directive in #5635, which will come in handy when it's needed to pass a dictionary item.

PR #9685, opened 2021-01-19
Inject missing localization service

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9683

### Description
It seems `localizationService` is used in controller of umbRadioButton component, which was added in this PR https://github.com/umbraco/Umbraco-CMS/pull/6797, but wasn't injected like in umbCheckbox component.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbradiobutton.directive.js
@@ -40,7 +40,7 @@
 (function () {
     'use strict';
 
-    function UmbRadiobuttonController($timeout) {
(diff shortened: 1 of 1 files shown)


bjarnef:
@nul800sebastiaan a simple PR submitted from mobile 😁🤓😅

Somehow anyone hasn't noticed this, so would be great to fix this 🤦‍♂️🙈

nathanwoulfe (approved):
Well spotted @bjarnef - either no one is using label-key for translation in anything custom, or they are and haven't noticed the error...

Wouldn't it be nice if we had a mechanism for providing strong types and intellisense in Javascript files. Could name it Typescript. 😉

nathanwoulfe:
@nul800sebastiaan this might need to be cherrypicked for the next round of patch releases...

PR #9693, opened 2021-01-20
Fix some exceptions being logged the wrong way

rbottema:
In these statements, the exception was passed as a log message parameter instead of as the exception. This meant the exception and including stack trace was not logged and thus lost.

### Prerequisites

### Description

As above, the log messages were missing important context as the exception was not logged. By switching around the parameters the correct overload is called.

It was calling https://github.com/umbraco/Umbraco-CMS/blob/2bfef741914297c802bc6c77cc48780d7534f920/src/Umbraco.Core/Logging/LoggerExtensions.cs#L65

But it intended to call https://github.com/umbraco/Umbraco-CMS/blob/2bfef741914297c802bc6c77cc48780d7534f920/src/Umbraco.Core/Logging/LoggerExtensions.cs#L37 or https://github.com/umbraco/Umbraco-CMS/blob/2bfef741914297c802bc6c77cc48780d7534f920/src/Umbraco.Core/Logging/LoggerExtensions.cs#L26

#### Testing steps:

- Trigger error
- See log now includes exception and stack traces
- Existing tests pass

### DIFF
--- a/src/Umbraco.ModelsBuilder.Embedded/LiveModelsProvider.cs
+++ b/src/Umbraco.ModelsBuilder.Embedded/LiveModelsProvider.cs
@@ -91,7 +91,7 @@ namespace Umbraco.ModelsBuilder.Embedded
             catch (Exception e)
             {
                 _mbErrors.Report("Failed to build Live models.", e);
-                _logger.Error<LiveModelsProvider>("Failed to generate models.", e);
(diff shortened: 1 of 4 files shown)


nul800sebastiaan:
Nice one @rbottema, good catch! Thanks for the help! 👍

PR #9715, opened 2021-01-25
Stop using .length !== 0 when its not jQuery

nielslyngsoe:
As I was getting this error;
image

I stumbled upon this code which is a bit tricky as it's using some JQuery and some Native code. Most of the code is using the jQuery find(), which always returns a jQuery collection. But the last part is using native find() of an Array, which will return undefined if nothing is found.

Therefor i have changed this.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/formhelper.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/formhelper.service.js
@@ -85,10 +85,11 @@ function formHelper(angularHelper, serverValidationManager, notificationsService
          */
         focusOnFirstError: function(form) {
             var invalidNgForms = form.$$element.find(`.umb-property ng-form.ng-invalid, .umb-property-editor ng-form.ng-invalid-required`);
-            if(invalidNgForms !== undefined && invalidNgForms.length > 0) {
-                var firstInvalidNgForm = invalidNgForms.first();
-                var focusableFields = firstInvalidNgForm.find("umb-range-slider .noUi-handle,input,textarea,select,button");
-                if(focusableFields !== undefined) {
+            var firstInvalidNgForm = invalidNgForms.first();
+
+            if(firstInvalidNgForm.length !== 0) {
+                var focusableFields = [...firstInvalidNgForm.find("umb-range-slider .noUi-handle,input,textarea,select,button")];
+                if(focusableFields.length !== 0) {
                     var firstErrorEl = focusableFields.find(el => el.type !== "hidden" && el.hasAttribute("readonly") === false);
                     if(firstErrorEl !== undefined) {
                         firstErrorEl.focus();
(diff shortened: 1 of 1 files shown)


src/Umbraco.Web.UI.Client/src/common/services/formhelper.service.js
```diff
@@ -89,9 +89,9 @@ function formHelper(angularHelper, serverValidationManager, notificationsService
 
             if(firstInvalidNgForm.length !== 0) {
                 var focusableFields = [...firstInvalidNgForm.find("umb-range-slider .noUi-handle,input,textarea,select,button")];
-                if(focusableFields.length !== 0) { 
+                if(focusableFields.length !== 0) {
                     var firstErrorEl = focusableFields.find(el => el.type !== "hidden" && el.hasAttribute("readonly") === false);
-                    if(firstErrorEl.length !== 0) {
+                    if(firstErrorEl !== undefined) {
```
bjarnef, line 94: @nielslyngsoe why not just the following, which would also check for `null`:

```
if(firstErrorEl) {
    firstErrorEl.focus();
}
```

nielslyngsoe: Well, that's also a fine solution. I just wrote this cause based on what the Mozilla docs says:
"If no values satisfies the testing function, undefined is returned."
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

nathanwoulfe: Personally, I'd only check length if I need the exact value - if I only need to know that the array has items, `if (arr.length)` is fine. If I'm checking a variable exists, I'd do as @bjarnef suggests.

PerplexDaniel: Alternatively I like to use `if(x != null)` which checks for both `null` and `undefined`. 
Pretty much the only case where `!=` is fine (vs. `!==`).

nathanwoulfe: I'd avoid abstract equality (==) without type safety. Intellisense should warn me if I'm comparing different types so that I only do it knowingly and without relying on coercion. 

If I'm comparing two variables, they are different if they are not the same type. If I need to compare a string and a number, I'll parse/toString one or the other first. 

Personal preference.

c9mbundy:
Of course, null and undefined can reflect potentially difference status, and not to mention that undefined is not actually reserved.
[Interesting article on the general topic](https://dmitripavlutin.com/7-tips-to-handle-undefined-in-javascript/)

nathanwoulfe (approved):
Fix is all good, but highlights discussion required around coding style for the UI rebuild. No need to dwell on it now though.

PR #9716, opened 2021-01-25
Add Check for ´err´

nielslyngsoe:
After this fix(https://github.com/umbraco/Umbraco-CMS/pull/9690) I sometimes get JS error for `err` being undefined.

This PR add a check for that.

To reproduce: make a DocType with Mandatory Media Picker, try to save without a media picked.


nathanwoulfe (approved):
@nielslyngsoe makes sense. Would be good to start using optional chaining for these cases (it's supported in all evergreen browsers, and can be polyfilled for older browsers if needed).

Issue #9699, opened 2021-01-21
Icon sizing inconsistencies in tree


jattwood:
Umbraco 8.10.1 the icon sizes in the tree display with inconsistent sizing if the text runs outside the container width/becomes truncated.

## Umbraco version

I am seeing this issue on Umbraco version: 8.10.1

Reproduction
------------
### Steps to reproduce

1) Navigate to content tree (for example). Ensure there are multiple nodes with variable node name lengths.
2) If all node names are fully visible, drag the tree section width to the left until some of the longer node names become truncated.

### Expected result

Expected result is that the icons remain at a consistent size regardless of length of node name text.

### Actual result

Actual result is that the longer the node name text, the smaller the icon:

![image](https://user-images.githubusercontent.com/6452616/105359509-88fe8a80-5bc5-11eb-83cc-c8b160469a4d.png)

### To Fix
Change the flex basis on the umb-tree-item__label label from "auto" to 0%
![image](https://user-images.githubusercontent.com/6452616/105362748-43dc5780-5bc9-11eb-8012-e914da9006cc.png)

nul800sebastiaan:
Interesting one! I've not seen that before! We'd love a little help with that please 🙏

umbrabot:
Hi @jattwood,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

mistyn8:
.umb-tree-icon {
    **flex-shrink: 0;**
}

seems to correct?

jattwood:
Yes that fixes it as well!

jattwood:
![image](https://user-images.githubusercontent.com/6452616/105742175-2846b900-5f09-11eb-9f7c-fa2df4080344.png)

results in:

![image](https://user-images.githubusercontent.com/6452616/105742231-3ac0f280-5f09-11eb-9940-74d26bd62af0.png)

mistyn8:
![image](https://user-images.githubusercontent.com/12418214/105742764-d2d3d180-5f33-11eb-882c-3e47328312a9.png)
with 0% I see the above..

[...]

PR #9718, opened 2021-01-25
Update umb-tree-item.less

mistyn8:
Icon sizing inconsistencies in tree

this fixes https://github.com/umbraco/Umbraco-CMS/issues/9699

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree-item.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree-item.less
@@ -3,6 +3,7 @@
     min-width: 100%;
     width: auto;
     margin-top:1px;
+    flex-shrink: 0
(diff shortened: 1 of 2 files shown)


nathanwoulfe:
@mistyn8 could you please double check this - adding `flex-shrink:0` to `.umb-tree-item` won't solve this problem as it's not a flex element. I think it needs to be on `.umb-tree-icon` as that's the element we want to prevent shrinking...

mistyn8:
@nathanwoulfe you aren't wrong. Sorry that the initial commit didn't reflect the solution posted on https://github.com/umbraco/Umbraco-CMS/issues/9699

Updated the PR so `flex-shrink:0` now added to `.umb-tree-icon` instead.

nathanwoulfe (approved):
All good, thanks @mistyn8

PR #9720, opened 2021-01-26
Reset valPropertyMsg when its the only error present.

nielslyngsoe:
In regards to en issue on .NetCore, it's made clear that this line prevents valPropertyMsg to be reset if it's already present.

For review with Shannon, I've send a message on Slack with a video describing this.

---
_This item has been added to our backlog [AB#10147](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10147)_


Shazwazza (approved):
@nielslyngsoe I think it makes sense. As this code is used in a huge amount of validation scenarios testing needs to be done for all of these scenarios. Such as ensuring the validation is cleared as expected when:

* server side validation for simple editors like textboxes and the change password one that this was discovered with
* server side validation for complex editors like nested content and block editor based on server side errors with sub-editors (like textboxes) at multiple nested levels ... and making sure the property errors all clear as expected at the various levels. This requires testing trying to clear errors at the deepest to shallowest levels and vice versa and perhaps also clearing errors at middle levels and moving outward

If all your testing passes then happy to have it merged.

PR #9729, opened 2021-01-28
Add more constants for tree icons

bjarnef:
### Prerequisites

### Description
Most trees was using icons from a constants static class, so I have updated this with more icons for tree nodes. Especially a constant for the generic `icon-folder` folder icon would be useful and since more areas in backoffice allow creating folders in tree.


nathanwoulfe (approved):
Sounds good to me, thanks @bjarnef

PR #9731, opened 2021-01-29
Netcore: Various bugfixes found while building realistic website

bergmania:
I found and fixed a couple of issues while building a realistic website: (Sorry all of them are in this same PR)

- ModelsBuilder: Inject PublishedValueFallback into static Mixins
- ModelsBuilder: Throw exception if the compiler can't compile the code
- CheckIfUserTicketDataIsStaleAttribute: Scope issue
- Ambiguous Actions: Couldn't determine the action when empty arrays were passed. 
  - Fixed by using more v8 like solution. 
  - Still stupid the client does not just have different endpoints for each use case. I guess there is a big overhead executing this code. Also the implementation in netcore is not as effective as netframework, as we have to return true/false for each endpoint, instead of just selecting the method to use.
- Fixed issue with reading the body from post requests. Often we were not allowed to seek in the stream.
  - This worked earlier. Maybe due to the .net5 update or just some refactoring that has gone wrong.
- UmbracoHelper: Made available on UmbracoViewPage 
- Client entity.resource.js: Don't ask the server when getByIds has 0 ids.
- Client content.resource.js: Renamed endpoint GetEmptyBlueprint to avoid ambiguous action name


bergmania:
See https://github.com/umbraco/Umbraco-CMS/pull/9740

PR #9738, opened 2021-02-01
Fix for applying searchResultFormatter on found searchresults

### LINKED ISSUE #9737: SearchResultFormatter in search.service.js is skiped in build of umbraco.services.js
We found out that when the searchbox in the linkpicker is used, the URL of a page is not being shown anymore since Umbraco 8.7 and later. It looks like that the buildproces of umbraco.services.js add an semicolon and thereby skipping the iteration of the found items. As a result, the URL is missing from the pages found.

Source without semicolon
https://github.com/umbraco/Umbraco-CMS/blob/6e1c29232a243e966bca392cee8b1f3966b04bec/src/Umbraco.Web.UI.Client/src/common/services/search.service.js#L69

Build with semicolon
afbeelding

I am seeing this issue on Umbraco version: 8.7 and later

mcl-sz:
We found out that when the searchbox in the linkpicker is used, the URL of a page is not being shown anymore since Umbraco 8.7 and later. This is because of the searchResultFormatter wasn't applied correct.

### Prerequisites

### Description

Fixes #9737

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/search.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/search.service.js
@@ -66,11 +66,12 @@ angular.module('umbraco.services')
                     throw "args.term is required";
                 }
 
-                return entityResource.search(args.term, "Document", args.searchFrom, args.canceler, args.dataTypeKey)
-                    _.each(data, function (item) {
-                        data.forEach(item => searchResultFormatter.configureContentResult(item));
-                        return data;
+                return entityResource.search(args.term, 'Document', args.searchFrom, args.canceler, args.dataTypeKey).then(function (data) {
+                    data.forEach(function (item) {
+                        return searchResultFormatter.configureContentResult(item);
                     });
+                    return data;
+                });
             },
 
             /**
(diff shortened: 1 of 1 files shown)


src/Umbraco.Web.UI.Client/src/common/services/search.service.js
```diff
@@ -66,11 +66,12 @@ angular.module('umbraco.services')
...
-                    _.each(data, function (item) {
-                        data.forEach(item => searchResultFormatter.configureContentResult(item));
-                        return data;
+                return entityResource.search(args.term, 'Document', args.searchFrom, args.canceler, args.dataTypeKey).then(function (data) {
+                    data.forEach(function (item) {
+                        return searchResultFormatter.configureContentResult(item);
                     });
+                    return data;
+                });
             },
```
bjarnef, line 75: @mcl-sz yes it seems in was changed in this PR https://github.com/umbraco/Umbraco-CMS/pull/8534

Alternatively it could use ES6 arrow syntax as with the other functions.

```
return entityResource.search(args.term, 'Document', args.searchFrom, args.canceler, args.dataTypeKey)
       .then(data => {
               data.forEach(function (item) {
                     return searchResultFormatter.configureContentResult(item);
               });
               return data;
      });
```

mcl-sz: I agree @bjarnef , that would be more consistant.

nathanwoulfe (approved):
Thanks @mcl-sz, this works correctly now. Must have been missed when 8534 was merged.

N

PR #9739, opened 2021-02-01
Prevent overflow of searchresults with long urls

mcl-sz:
Small css change to prevent overflow of searchresults with long url's

### Prerequisites

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less
@@ -104,10 +104,9 @@ body.touch .umb-tree {
         .umb-button-ellipsis--hidden {
             opacity: 1;
         }
-        
+         
         .umb-tree-icon {
             color: @ui-option-type-hover;
-            flex-shrink: 0;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


nathanwoulfe:
Hi @mcl-sz, I wasn't sure exactly where this was applied, so have checked both the search in a picker and the backoffice search (shots below). I've tweak the styling on the picker results to make sure the containing button doesn't overflow its parent, and shifted the flex-shrink on the tree-icon (which was added in #9718) to ensure it's applied to icons in search as well.

I've pushed those changes to your branch, so you should be able to review before I merge.

Contentpicker:
image

Search:
image

The page title is treated differently in the picker search versus the site search results, not sure that it's really an issue though. What could be an issue is the search results and content tree using two different directives for rendering, despite being virtually identical. If the tree-node directive was extended to include the subtitle, these could be combined. Would make things more consistent as the search result directive renders with different padding/whitespace. Ping @nielslyngsoe and @madsrasmussen - if they are ok with the suggestion, I'll create an issue.

nathanwoulfe (approved):
Will merge this, but would be good to review the differences between tree nodes and search results, and see if they can share a common UI, which would reduce the likelihood of these issues in future. All good now though, text overflow is handled properly.

PR #9743, opened 2021-02-02
WIP - Adds support for a docker images and allows specifying local configuration

Shazwazza:
Adds support for a docker images and allows specifying local configuration in debug mode that isn't committed to the repo.

* Moves Nuget.config file to where the sln file is because that is where we run the docker command from and it can't copy above it's root. Perhaps this is optional and you can just run docker from the root of the git repo, but since Nuget.config is specific to code, I figured it didn't really belong in the root anyways
* Allows for having a appsettings.local.json file to have user specific settings in there so it doesn't pollute the main one and we accidentally keep committing unwanted changes. If the file doesn't exist it doesn't do anything. For example, my local config file is
     ```json
    {
      "ConnectionStrings": {
        "umbracoDbDSN": "server=MYSERVER;database=MYDB;user id=MYUSER;password='MYPASS'"
      },
      "Umbraco": {
        "CMS": {
          "RuntimeMinification": {
            "dataFolder": "Umbraco/TEMP/Smidge",
            "version": "637429827442056840"
          }
        }
      }
    }
     ````
* Adds a docker file to build the web app (and a docker ignore file)
* Ensure razor views are not precompiled on publish (just like they are not on build either), else it will try to precompile your umbraco front-end views which will fail due to MB models not in a DLL. (plus we don't want to ship those)
* Cleans up some of the .sln file since there was lots of duplicated lines

To run the docker build, have Docker Desktop installed (or whatever you use), at the cmd line go to the /src folder, then:

```
docker build -f Umbraco.Web.UI.NetCore/Dockerfile -t umbracocms .
```

once that is build do:

```
docker run -d -p 8080:80 --name umb1 umbracocms
```

which will host it on port 8080.

... This 'works' but not quite because we get 500 errors when it redirects to the /install path which needs investigation. In docker desktop it can show a log output but we see nothing. I'm unsure if it's possible to have umbraco log out to the docker console? 

I'm a docker N00B so just figuring this stuff out.


CarlSargunar:
Excited to try this out too myself (also a docker noob) so if I can do anything please let me know - happy to help with testing + feedback :-)

nul800sebastiaan:
I'll close this one for now, I'm sure there's new ideas for docker support 2 years on! 😅

Issue #9594, opened 2020-12-22
Recurring 'Lock request time out period exceeded' exception


m4nickroll:
Production site - load balanced using Azure. Seemingly random occurrences of 'Lock request time out period exceeded' exception. Resolved by scaling the database tier up or down. Returns within days however, preventing editing. Running 8.8.0 and following the recommended config from [](https://our.umbraco.com/documentation/getting-started/setup/server-setup/load-balancing/azure-web-apps). Screenshot of successful publish then the first 'Lock request time out period exceeded' exception a few minutes later:

![image](https://user-images.githubusercontent.com/62098385/102896532-7413c780-445e-11eb-8296-788a4bb39dd9.png)

nul800sebastiaan:
@m4nickroll Oh no! 

Generally, this COULD be an indication that your load balancing setup is not quite right, most of all, it is possible your backoffice is not isolated and accessible from multiple servers.

However, we've been fixing some of these errors, some occurred due to transient errors on SQL Azure, which might be the problem in your case. The latest fix for that is available in the recently released version 8.10.0: https://github.com/umbraco/Umbraco-CMS/pull/9543 - this fix looks directly related to your error. I would recommend you try out the latest version (8.10.1) now and see if that helps. 
We also intend to patch this for 8.8.x but I don't have a release date for a patch yet I'm afraid.

m4nickroll:
@nul800sebastiaan thanks for your response.

Our backoffice is on a dedicated master app and /umbraco is not available from either of our two front-end apps. Master/Replica server role is also set correctly.

We have a separate issue with Umbraco Form submissions taking a while before producing a 504. The last time we had the lock timeout issue we noticed a (failed) form submission just before the lock timeouts began. Would that shed any light?

That issue certainly seems relevant though - will take a look at 8.10

SimonAntony:
@nul800sebastiaan hello stranger :-)

I've been getting this more and more since 8.9 and it's progressively getting worse (even after upgrading to 8.11 today). Setup is Dedicated server running IIS/SQL Server 2017 (so no azure).

Always happens after doing a few publish events (updating content) - only way to fix it is to restart SQL Server which is far from ideal, god knows how i'd do this if it was Azure.

Log entry:

2021-01-28 11:49:01,926 [P15776/D2/T17] ERROR  Umbraco.Core.Persistence.UmbracoDatabase - Exception (3909210e).
[...]

Issue #9514, opened 2020-12-08
This document is published but its url cannot be routed 


sbosell:
Sometimes a node in umbraco will be in a state of published but not routable when this should not be allowed to occur.  Umbraco in this case throws a 404 for that route.

There are several cases of this being [reported](https://our.umbraco.com/forum/using-umbraco-and-getting-started/104334-this-document-is-published-but-its-url-cannot-be-routed) as well as other issues reported here ([duplicate](https://github.com/umbraco/Umbraco-CMS/issues/9523))

Related to issue #7575 

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.2, 8.7.x, and 8.9.1

Reproduction
------------

I have a process that syncs data in the backend every 30 minutes to a node's descendants and then the content api calls SaveAndPublish and/or SaveAndPublishBranch (tested with both in the sync).  The sync uses the content api via a recurring task, all standard features of Umbraco.  In my case we have no cultures defined and only one root site.   The most basic Umbraco setup.  Others are reporting the same issue so it may have nothing to do with the Content Api.

About 1 time per day the root node of the sync will be in this state of the url can't be routed and it is random in that it only occurs every once in a while which leads me to believe there is a locking mechanism with how the cache works.  This should never happen, there should be no code path that unpublished or deroutes a node like this in the save and publish.

Hosted on Azure in a single site (not load balanced) with all the appropriate settings set via the Umbraco [Documentation](https://our.umbraco.com/documentation/getting-started/setup/server-setup/azure-web-apps).

### Bug summary

A node becomes not routable when this should never happen.

### Specifics

I can privately provide a URL 

### Steps to reproduce

It happens randomly

### Expected result

A valid published node in Umbraco should NEVER be in a state of the URL not being routable. 

### Actual result

Umbraco backend reports the URL not being routable and the node is effectively unpublished.

---
_This item has been added to our backlog [AB#9840](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/9840)_

sbosell:
This error just occurred right now.  

![image](https://user-images.githubusercontent.com/1896962/101559516-483d1000-398f-11eb-8cad-20492b2abe89.png)

sbosell:
This happened almost daily this week and is being reported by others.  Duplicate ticket: https://github.com/umbraco/Umbraco-CMS/issues/9523

[...]

PR #9744, opened 2021-02-02
Add the ability to change the SQL Write Lock TimeOut

nul800sebastiaan:
When acquiring a SQL write lock, we currently have hardcoded this to 1.8 seconds (1800ms). This has been the case since version 8.4.0 - https://github.com/umbraco/Umbraco-CMS/pull/6688

In some cases, people seem to be getting the error `Lock request time out period exceeded.`. From what I gather from Shannon this indicates the followin:

> Has to do with distributed locks when writing/reading from the database at the same time as another thread/process. This means that while something else is holding a write lock, something else is waiting for the readlock and it cannot acquire it because it waited too long. Normally do to too many write locks trying to be used or queries executing slowly.

From what I understand from conversations at HQ, the value of 1800ms is rather arbitrary and there's no objection to increasing it to 5000. This PR is a first step towards that, we have seen various issues with locks being exceeded.

- Issue https://github.com/umbraco/Umbraco-CMS/issues/8433 was resolved by actually improving the performance of operations on members. 
- Issue https://github.com/umbraco/Umbraco-CMS/issues/9386  seems to be related to doing inefficient write operations in custom code
- Issue https://github.com/umbraco/Umbraco-CMS/issues/9514 also shows us the error `Lock request time out period exceeded.`, it might be related but there's not enough info yet.
- Issue https://github.com/umbraco/Umbraco-CMS/issues/9594 also shows this error and might point to a relation to either Forms and/or uSync

With this PR merged, it will be possible to at least experiment with the lock timeout and alleviate some pain for now. It is still a good idea to investigate what's holding locks and why the lock time is not enough for some operations.

### DIFF
--- a/src/Umbraco.Core/Configuration/GlobalSettings.cs
+++ b/src/Umbraco.Core/Configuration/GlobalSettings.cs
@@ -391,5 +391,28 @@ namespace Umbraco.Core.Configuration
                 }
             }
         }
+
+
+        /// <summary>
+        /// An int value representing the time in milliseconds to lock the database for a write operation
+        /// </summary>
+        /// <remarks>
+        /// The default value is 1800 milliseconds
+        /// </remarks>
+        /// <value>The timeout in milliseconds.</value>
+        public int SqlWriteLockTimeOut
+        {
+            get
+            {
+                try
+                {
+                    return int.Parse(ConfigurationManager.AppSettings[Constants.AppSettings.SqlWriteLockTimeOut]);
+                }
+                catch
+                {
+                    return 1800;
+                }
+            }
+        }
     }
 }
--- a/src/Umbraco.Core/Persistence/SqlSyntax/SqlCeSyntaxProvider.cs
+++ b/src/Umbraco.Core/Persistence/SqlSyntax/SqlCeSyntaxProvider.cs
@@ -1,9 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Configuration;
 using System.Data;
 using System.Data.SqlServerCe;
 using System.Linq;
 using NPoco;
+using Umbraco.Core.Composing;
 using Umbraco.Core.Persistence.DatabaseAnnotations;
 using Umbraco.Core.Persistence.DatabaseModelDefinitions;
 
@@ -163,7 +165,8 @@ where table_name=@0 and column_name=@1", tableName, columnName).FirstOrDefault()
             if (db.Transaction.IsolationLevel < IsolationLevel.RepeatableRead)
                 throw new InvalidOperationException("A transaction with minimum RepeatableRead isolation level is required.");
 
-            db.Execute(@"SET LOCK_TIMEOUT 1800;");
+            var timeOut = Current.Configs.Global().SqlWriteLockTimeOut;
+            db.Execute(@"SET LOCK_TIMEOUT " + timeOut  + ";");
             // *not* using a unique 'WHERE IN' query here because the *order* of lockIds is important to avoid deadlocks
             foreach (var lockId in lockIds)
             {
--- a/src/Umbraco.Core/Configuration/IGlobalSettings.cs
+++ b/src/Umbraco.Core/Configuration/IGlobalSettings.cs
@@ -72,5 +72,10 @@
         /// Gets the location of temporary files.
         /// </summary>
         string LocalTempPath { get; }
+
+        /// <summary>
+        /// Gets the write lock timeout.
+        /// </summary>
+        int SqlWriteLockTimeOut { get; }
     }
 }
--- a/src/Umbraco.Core/Constants-AppSettings.cs
+++ b/src/Umbraco.Core/Constants-AppSettings.cs
@@ -140,6 +140,14 @@ namespace Umbraco.Core
                 /// </summary>
                 public const string DatabaseFactoryServerVersion = "Umbraco.Core.Debug.DatabaseFactoryServerVersion";
             }
+
+            /// <summary>
+            /// An int value representing the time in milliseconds to lock the database for a write operation
+            /// </summary>
+            /// <remarks>
+            /// The default value is 1800 milliseconds
+            /// </remarks>
+            public const string SqlWriteLockTimeOut = "Umbraco.Core.SqlWriteLockTimeOut";
         }
     }
 }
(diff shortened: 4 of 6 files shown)


SimonAntony (approved):
Looks good to me and nice to be configurable. Happy to test this release when ready on a site I have timeout issues on.

src/Umbraco.Core/Persistence/SqlSyntax/SqlCeSyntaxProvider.cs
```diff
@@ -163,7 +165,8 @@ public override void WriteLock(IDatabase db, params int[] lockIds)
             if (db.Transaction.IsolationLevel < IsolationLevel.RepeatableRead)
                 throw new InvalidOperationException("A transaction with minimum RepeatableRead isolation level is required.");
 
-            db.Execute(@"SET LOCK_TIMEOUT 1800;");
+            var timeOut = Current.Configs.Global().SqlWriteLockTimeOut;
+            db.Execute(@"SET LOCK_TIMEOUT " + timeOut  + ";");
```
bjarnef, line 169: @nul800sebastiaan maybe it would be better with SQL parameters here although it should be safe enough that timeout is an `int` here and not vulnerable to SQL injection.

One of these should work as well though, but I haven't tested it :)

```
var sql = new Sql(@"SET LOCK_TIMEOUT @0;", timeOut);
db.Execute(sql);
```

```
var sql = new Sql(@"SET LOCK_TIMEOUT @timeout;", new { timeout = timeOut });
db.Execute(sql);
```

nul800sebastiaan: Thanks @bjarnef - [since we force it to be an `int`](SET LOCK_TIMEOUT), we don't have to be worried about SQL injection here.

bjarnef: Yeah, I noticed that one. It is probably just a matter of readability or preferred use of NPoco.
Alternatively it could also use string interpolation syntax.

OwainWilliams (approved):
Tested with running the default values and also adding a new key to the web.config. Values were applied and published worked without any issues. Tested with a plain text string in a document type and also with Nest Content elements. 
Value set to -1 to ensure that the lock timeout was being activated.

src/Umbraco.Core/Configuration/GlobalSettings.cs
```diff
@@ -391,5 +391,28 @@ public bool UseHttps
...
+        {
+            get
+            {
+                try
+                {
+                    return int.Parse(ConfigurationManager.AppSettings[Constants.AppSettings.SqlWriteLockTimeOut]);
+                }
+                catch
+                {
+                    return 1800;
```
Shazwazza, line 413: Default should be changed to perhaps 5000.

Shazwazza (commented):
I have left some comments on the code.

Before this is released, we need to test what happens with high timeout values compared to what the db connection and command timeout is - because we aren't sure what happens if the configured lock timeout exceeds either of those, we may very well end up with other different more vague errors which would be worse.

src/Umbraco.Core/Configuration/GlobalSettings.cs
```diff
@@ -391,5 +391,28 @@ public bool UseHttps
...
+        /// The default value is 1800 milliseconds
+        /// </remarks>
+        /// <value>The timeout in milliseconds.</value>
+        public int SqlWriteLockTimeOut
+        {
+            get
+            {
+                try
+                {
+                    return int.Parse(ConfigurationManager.AppSettings[Constants.AppSettings.SqlWriteLockTimeOut]);
```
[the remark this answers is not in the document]
nul800sebastiaan: I was just repeating the code above! 😅 
Good point in adding a backing field here, I was under the assumption these settings were cached somewhere. 😄

src/Umbraco.Core/Configuration/GlobalSettings.cs
nul800sebastiaan: I kept it the same so that it will not change any behavior for now unless specifically configured.

nul800sebastiaan:
> Before this is released, we need to test what happens with high timeout values compared to what the db connection and command timeout is - because we aren't sure what happens if the configured lock timeout exceeds either of those, we may very well end up with other different more vague errors which would be worse.

Should we limit it to a max value so people can't inflict too much pain on themselves?

afinn-tech:
Sounds like a great , we struggled with the time , cant wait for this fix - great work guys

stefankip:
We're running into this issue **a lot** with Umbraco Cloud. Error logs getting spammed with `System.Data.SqlClient.SqlException (0x80131904): Lock request time out period exceeded. The statement has been terminated.` messages.
Looking forward to 8.12

afinn-tech:
@nul800sebastiaan any chance of RC of 8.12 ? so we can test this function ?

nul800sebastiaan:
@afinn-tech The 8.12 RC is out now: https://our.umbraco.com/download/releases/8120 / https://www.nuget.org/packages/UmbracoCms/8.12.0-rc

stefanoberetta:
Is the `Lock request time out period exceeded.` error supposed to be triggered when I have two websites (one running localhost and the other one on a dev server -NOT Azure-) sharing the same database?

Shazwazza:
@stefanoberetta this lock is a distributed lock that uses the database. If you have 2 websites pointing to the same database you are Load Balancing with more than one main server which is not explicitly supported (but people do it). This error isn't specific to load balancing, this error will occur on a single server or multiple servers where a write lock is held for too long while another request or process is trying to acquire another read/write lock with the same ID but it times out.

here's a description: https://github.com/umbraco/Umbraco-CMS/issues/9594#issuecomment-771259639

timmather:
We are having a similar problem with a 8.6+ build (now on 8.12). Single server setup iis and sql on the same box. After 6/7 saves we get this error, worse with multiple editors, nothing special running at all, tried increasing timeout using writelock setting, content started to disappear across the site. It’s a big problem we have two sites waiting to go live this week with 20 editors and they can’t use the site. Never been a problem before on over 80 deployments.

stefankip:
@timmather is it using Vendr perhaps?

timmather:
Unfortunately not, it’s actually a very simple build apart from having around 2k nodes.

PR #9748, opened 2021-02-03
NetCore: Updated healthchecks 

bergmania:
### Description
The current implementation of healthchecks is not supported in .NET Core, due the the new configuration paradigme. 
The new configuration reads from multiple sources (some of them can be read-only). 

Our current healthchecks try to fix the issue programmatically by updating the config file. This worked okay in v8, because the configuration was transformed.
- For production this still wasn't perfect, as the changes was not added to git.

We therefore discussed in the unicore team what to do, and we all agreed a good solution would be just to link to some documentation that can descript how to resolve the issue.

### Details
- Removed healthchecks that do not make sense for .NET Core
  - IIS custom errors
  - Tracing
- Updated all settings and config healthchecks to show a link instead.
  - All links are collected in a constants file.
  - All links are tiny urls that we can redirected from bitly.com (HQ only)
- Fix namespace issue 
  - We used `HealthCheck` as both a class name and namespace name
  - Now the namespace is called `HealthChecks`
- Fixed issues in the certification healthcheck.
  - Now uses the .NET Core way
- Fixed TODO in FolderAndFilePermissionsCheck 
  - Now uses FilePermissionHelper
  - Updated FilePermissionHelper to use an enum for keys instead of hardcoded text.
     - Also added translations to this enum

### Test
- Test the healthchecks
  - When error a link should be shown (OBS not all links are created in bitly yet)
  - The SMTP link currently redirects to a 404 at out github documentation for netcore.
- Test install when one file or folder do not have correct permissions
  - I removed write permissions on the `/wwwroot/css` folder


src/Umbraco.Core/HealthChecks/Checks/Security/HstsCheck.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Umbraco.Core.HealthChecks.Checks.Security
+{
+    /// <summary>
+    /// Health check for the recommended production setup regarding the Strict-Transport-Security header.
+    /// </summary>
+    [HealthCheck(
+        "E2048C48-21C5-4BE1-A80B-8062162DF124",
+        "Cookie hijacking and protocol downgrade attacks Protection (Strict-Transport-Security Header (HSTS))",
+        Description = "Checks if your site, when running with HTTPS, contains the Strict-Transport-Security Header (HSTS). If not, it adds with a default of 18 weeks.",
```
nikolajlauridsen, line 15: We say that we add the HSTS header if it's not there, but as I understand, we don't do that anymore, but just link to documentation.

nikolajlauridsen (commented):
Changes look good to me, except for one very minor description string, just need to do the test steps now 👍

src/Umbraco.Core/HealthChecks/Checks/Security/HstsCheck.cs
bergmania, line 15:
```suggestion
        Description = "Checks if your site, when running with HTTPS, contains the Strict-Transport-Security Header (HSTS).",
```

bergmania:
Please not that the links are not setup in bitly yet (as the docs are not merged) but you can see the docs PR here:
https://github.com/umbraco/UmbracoCMSDocs/pull/1
https://github.com/umbraco/UmbracoCMSDocs/pull/3

PR #8921, opened 2020-09-16
Reduce memory footprint of database reads and writes by avoiding params[] allocations for ReadLock() and WriteLock()

nzdev:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
What?
Reduce params[] allocation for WriteLock and ReadLock. Avoids allocating an array by implementing a non params overload.

Why?
Less allocations, better performance.

How to test?
Existing unit tests pass.

PR #9750, opened 2021-02-03
Add the ability to change the SQL Write Lock TimeOut

nul800sebastiaan:
Updates based on feedback on #9744

- Default value is now 5000
- There's a limit to the range of the timeout values
- Introduced backing field so as not to keep looking up the same value over and over again from the config

### DIFF
--- a/src/Umbraco.Core/Configuration/GlobalSettings.cs
+++ b/src/Umbraco.Core/Configuration/GlobalSettings.cs
@@ -6,7 +6,9 @@ using System.Web;
 using System.Web.Configuration;
 using System.Web.Hosting;
 using System.Xml.Linq;
+using Umbraco.Core.Composing;
 using Umbraco.Core.IO;
+using Umbraco.Core.Logging;
 
 namespace Umbraco.Core.Configuration
 {
@@ -23,6 +25,7 @@ namespace Umbraco.Core.Configuration
         // TODO these should not be static
         private static string _reservedPaths;
         private static string _reservedUrls;
+        private static int _sqlWriteLockTimeOut;
 
         //ensure the built on (non-changeable) reserved paths are there at all times
         internal const string StaticReservedPaths = "~/app_plugins/,~/install/,~/mini-profiler-resources/,"; //must end with a comma!
@@ -392,26 +395,38 @@ namespace Umbraco.Core.Configuration
             }
         }
 
-
         /// <summary>
         /// An int value representing the time in milliseconds to lock the database for a write operation
         /// </summary>
         /// <remarks>
-        /// The default value is 1800 milliseconds
+        /// The default value is 5000 milliseconds
         /// </remarks>
         /// <value>The timeout in milliseconds.</value>
         public int SqlWriteLockTimeOut
         {
             get
             {
-                try
+                if (_sqlWriteLockTimeOut != default) return _sqlWriteLockTimeOut;
+
+                var timeOut = 5000; // 5 seconds
+                var appSettingSqlWriteLockTimeOut = ConfigurationManager.AppSettings[Constants.AppSettings.SqlWriteLockTimeOut];
+                if(int.TryParse(appSettingSqlWriteLockTimeOut, out var configuredTimeOut))
                 {
-                    return int.Parse(ConfigurationManager.AppSettings[Constants.AppSettings.SqlWriteLockTimeOut]);
+                    // Only apply this setting if it's not excessively high or low
+                    const int minimumTimeOut = 100;
+                    const int maximumTimeOut = 20000;
+                    if (configuredTimeOut >= minimumTimeOut && configuredTimeOut <= maximumTimeOut) // between 0.1 and 20 seconds
+                    {
+                        timeOut = configuredTimeOut;
+                    }
+                    else
+                    {
+                      Current.Logger.Warn<GlobalSettings>($"The `Umbraco.Core.SqlWriteLockTimeOut` setting in web.config is not between the minimum of {minimumTimeOut} ms and maximum of {maximumTimeOut} ms, defaulting back to {timeOut}");
+                    }
                 }
-                catch
-                {
-                    return 1800;
-                }
+
+                _sqlWriteLockTimeOut = timeOut;
+                return _sqlWriteLockTimeOut;
             }
         }
     }
--- a/src/Umbraco.Core/Configuration/GlobalSettings.cs
+++ b/src/Umbraco.Core/Configuration/GlobalSettings.cs
@@ -422,7 +422,7 @@ namespace Umbraco.Core.Configuration
                     }
                     else
                     {
-                      Current.Logger.Warn<GlobalSettings>($"The `Umbraco.Core.SqlWriteLockTimeOut` setting in web.config is not between the minimum of {minimumTimeOut} ms and maximum of {maximumTimeOut} ms, defaulting back to {timeOut}");
+                      Current.Logger.Warn<GlobalSettings>($"The `{Constants.AppSettings.SqlWriteLockTimeOut}` setting in web.config is not between the minimum of {minimumTimeOut} ms and maximum of {maximumTimeOut} ms, defaulting back to {timeOut}");
                     }
                 }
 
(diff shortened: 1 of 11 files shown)


nul800sebastiaan:
@Shazwazza Made some updates based on your review, chose some values for a min/max range but those might need adjusting? Let me know!

src/Umbraco.Core/Configuration/GlobalSettings.cs
```diff
@@ -397,21 +400,34 @@ public bool UseHttps
...
+                    const int maximumTimeOut = 20000;
+                    if (configuredTimeOut >= minimumTimeOut && configuredTimeOut <= maximumTimeOut) // between 0.1 and 20 seconds
+                    {
+                        timeOut = configuredTimeOut;
+                    }
+                    else
+                    {
+                      Current.Logger.Warn<GlobalSettings>($"The `Umbraco.Core.SqlWriteLockTimeOut` setting in web.config is not between the minimum of {minimumTimeOut} ms and maximum of {maximumTimeOut} ms, defaulting back to {timeOut}");
+                    }
                 }
```
bjarnef, line 427: @nul800sebastiaan should it use `Constants.AppSettings.SqlWriteLockTimeOut` here instead?

nul800sebastiaan: Nice one! Updated. 👍

Shazwazza (commented):
Looks pretty good. What we really need to do though is test. I would like to know what happens when the timeout value exceeds the sql command/connection timeout value. I'm unsure if 20 seconds exceeds this or not. In any case, we need to know what happens when it is set at the max 20 seconds and a lock is still held. Do we get the same exception or a new one (i.e. if that exceeds the command/connection timeout). 

In order to be tested though, the code has to change. To do without breaking changes, the `Scope` instance can have 2x overloads: `public void ReadLock(int timeout, params int[] lockIds)` and `public void WriteLock(int timeout, params int[] lockIds)` so that explicit timeouts can be passed in. Then new overloads need to be added to `ISqlSyntaxProvider` - theoretically we are allowed these changes with our breaking change policies: `void ReadLock(IDatabase db, int timeout, params int[] lockIds);`, `void WriteLock(IDatabase db, int timeout, params int[] lockIds);` Then the implementations of the ISqlSyntaxProvider will pass in the config value to these overloads for the timeout. 

Then this can be tested:

* Have a controller action create a writelock: 
   ```cs
        using (var scope = ScopeProvider.CreateScope())
        {
            var realScope = (Scope)scope;
            // a really long timeout
            realScope.WriteLock(Constants.Locks.ContentTree, 60000);
            Thread.Sleep(601000); // Wait longer than the timeout
            scope.Complete();
        }
   ```
* Have a controller action create a readlock:
  ```cs
        using (var scope = ScopeProvider.CreateScope())
        {
            var realScope = (Scope)scope;
            // a shorter timeout
            realScope.WriteLock(Constants.Locks.ContentTree, 20000);
            Thread.Sleep(21000); // Wait longer than the timeout
            scope.Complete();
        }
  ```
* Then make a request to the first action which will hold the write lock for a long time
* Then make a request to the 2nd action which will try to acquire a read lock but will timeout since the write lock is taking longer than the read lock timeout
* An exception will be thrown - what is it

... then test really high values.

nzdev:
If new overloads are being created for WriteLock/ ReadLock, they should not use params[] in order to avoid array allocation. See https://github.com/umbraco/Umbraco-CMS/pull/8921

nul800sebastiaan:
Right gentlemen, I've updated that with the requested changes but can't get the test to do anything, probably doing it wrong and I need some help with that. I incorporated the changes in #8921 rendering that PR obsolete.

2 things:
1. I don't know what is more performant `db.Execute("SET LOCK_TIMEOUT" + timeout.TotalMilliseconds + ";");` / `db.Execute($"SET LOCK_TIMEOUT{timeout.TotalMilliseconds}";");` or string.Format or something else - happy to change to whatever is best  
2. I used TimeSpan everywhere though I wanted to go for a straight up `int` first, but this seems like it was a breaking change earlier on: https://github.com/umbraco/Umbraco-CMS/blame/02963e7b29a60865728e9824c13474ac7d8e0e1c/src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs#L263  
Probably not good to change it all to `int` now so stuck with TimeSpan to make it consistent.

I barely know what I'm doing here, so take it easy on me! 😅

src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
```diff
@@ -254,30 +254,50 @@ public override bool DoesTableExist(IDatabase db, string tableName)
...
         }
 
-        public void WriteLock(IDatabase db, TimeSpan timeout, params int[] lockIds)
+        public override void WriteLock(IDatabase db, params int[] lockIds)
         {
             // soon as we get Database, a transaction is started
 
             if (db.Transaction.IsolationLevel < IsolationLevel.ReadCommitted)
-                throw new InvalidOperationException("A transaction with minimum ReadCommitted isolation level is required.");
+                throw new InvalidOperationException("A transaction with minimum RepeatableRead isolation level is required.");
```
Shazwazza, line 272: This text should not have changed.

src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs
```diff
@@ -286,15 +306,24 @@ public override void ReadLock(IDatabase db, params int[] lockIds)
...
-                var i = db.ExecuteScalar<int?>("SELECT value FROM umbracoLock WITH (REPEATABLEREAD) WHERE id=@id", new { id = lockId });
-                if (i == null) // ensure we are actually locking!
-                    throw new ArgumentException($"LockObject with id={lockId} does not exist.", nameof(lockIds));
+                ObtainReadLock(db, timeout, lockId);
             }
         }
 
+        private static void ObtainReadLock(IDatabase db, TimeSpan timeout, int lockId)
+        {
+            db.Execute(@"SET LOCK_TIMEOUT " + timeout.TotalMilliseconds  + ";");
```
Shazwazza, line 319: This is new, this was never done before when obtaining a readlock, see https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Persistence/SqlSyntax/SqlServerSyntaxProvider.cs#L282

src/Umbraco.Core/Persistence/SqlSyntax/SqlCeSyntaxProvider.cs
```diff
@@ -158,40 +158,72 @@ public override bool DoesTableExist(IDatabase db, string tableName)
...
-                var i = db.ExecuteScalar<int?>("SELECT value FROM umbracoLock WHERE id=@id", new { id = lockId });
-                if (i == null) // ensure we are actually locking!
-                    throw new ArgumentException($"LockObject with id={lockId} does not exist.");
+                ObtainReadLock(db, timeout, lockId);
             }
         }
 
+        private static void ObtainReadLock(IDatabase db, TimeSpan timeout, int lockId)
+        {
+            db.Execute(@"SET LOCK_TIMEOUT " + timeout.TotalMilliseconds + ";");
```
Shazwazza, line 221: This is new, this was never done before when obtaining a readlock, see https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Persistence/SqlSyntax/SqlCeSyntaxProvider.cs#L180

Shazwazza (changes requested):
Have left a few comments. I've also pushed some code. But the big dangerous thing in these changes is introducing a lock timeout request for the read lock since this was never done before. I'm going to look into reverting that.

nul800sebastiaan:
>  these need to run in parallel.

Duh of course! 🤦 

> Removed the sql command to set the lock timeout on the default read locks - this was probably unintentional 

Good catch! Yep, copied a BIT too much code there! 😬  

Thanks for the help! Looks good to me now.

PR #9752, opened 2021-02-03
Fix eventService memory leaks

nzdev:
### Prerequisites

### Description
What?
Fix memory leaks by unbinding event from $rootscope.

Why?
Memory leaks in the browser when events are not unbound from the $rootscope.

How to test?
Ensure memory freed when navigating the backoffice ui.


nzdev:
tinymce.service.js also appears to be leaking via events, leaving tinymce content in memory

src/Umbraco.Web.UI.Client/src/common/services/editor.service.js
```diff
@@ -173,13 +173,14 @@ When building a custom infinite editor view you can use the same components as a
...
         // events for backdrop
-        eventsService.on("appState.backdrop", function (name, args) {
+        var backgroundEvent = eventsService.on("appState.backdrop", function (name, args) {
             if (args.show === true) {
                 blur();
             } else {
                 focus();
             }
         });
+
```
nzdev, line 183: no access to $scope, need another approach for destruction

src/Umbraco.Web.UI.Client/src/common/services/localization.service.js
```diff
@@ -326,10 +326,11 @@ angular.module('umbraco.services')
     };
 
     //This happens after login / auth and assets loading
-    eventsService.on("app.authenticated", function () {
+    var appAuthenticated = eventsService.on("app.authenticated", function () {
         resourceFileLoadStatus = "none";
         resourceLoadingPromise = [];
     });
+
```
nzdev, line 333: no access to $scope, need another approach for destruction

src/Umbraco.Web.UI.Client/src/common/directives/components/content/umbvariantcontenteditors.directive.js
```diff
@@ -201,6 +203,7 @@
 
             $location.search({"cculture": culture, "csegment": vm.editors[0].content.segment});
             splitViewChanged();
+            //unbindSplitViewRequest();
```
nzdev, line 206: remove commented out line

nzdev:
https://www.toptal.com/angular-js/top-18-most-common-angularjs-developer-mistakes

src/Umbraco.Web.UI.Client/src/common/services/editor.service.js
nathanwoulfe: @nzdev service is a singleton, so no need to worry about tear down

nathanwoulfe (approved):
Thanks @nzdev, I'd noticed memory spikes a while back when opening multiple infinite editors, this should help alleviate that issue. I've reverted the changes in the services as it's not required due to the lifespan of those object.

PR #9756, opened 2021-02-03
Make sorting in media consistent with sorting in content

bjarnef:
### Prerequisites

### Description
I noticed the sorting in Content and Media currently isn't consistent. In Content section the sorting is restricted to y-axis and within parent container, so this PR updates this as well for media sorting.
I previous worked on that in https://github.com/umbraco/Umbraco-CMS/pull/8419 but somehow I forgot to add this the media as well.

I have also cleaned up the sort view as in the old PR.


nul800sebastiaan:
Ah great! I didn't check media sorting either.. duh! 

Thanks, works as advertised! 👍

PR #9757, opened 2021-02-03
Update icons in umb-tree-search-results component

bjarnef:
### Prerequisites

### Description
Inside view of `<umb-tree-search-results>` component I noticed some unnecessary use of `ng-if`. Instead the classes can be added using inline conditional statements.

Furthermore after the link in the result item has been changed to a `<button>` element, it no longer used the full width of the container.

This is shown in eg. Move dialog when filtering, where the `<umb-tree-search-results>` component is used the list the result items.

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/tree/umb-tree.less
@@ -155,6 +155,7 @@ body.touch .umb-tree {
 
         &-link {
             display: block;
+            width: 100%;
         }
 
         &-name {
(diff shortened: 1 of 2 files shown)


mikecp:
Hello @bjarnef ,

Thanks for your PR!

I had a look at it and I experience the following display, where the path of the item is centered instead of left align, which looks kind of odd.
Do you experience the same or would this be some local hiccup?

image

bjarnef:
Hi @mikecp 

I will have another look at this. Immediately I don't see something specific which should center this text, but be default text inside `<button>` element is centered.

Do you see this with latest changes in `v8/contrib` branch as well?

mikecp:
HI @abjerner ,

Yes it's on the latest version of V8/contrib. I think that due to the fact that you have set the width to 100% on the "&-link" styl,e, the button (which has class `umb-search-group-item-link`) is now taking 100% of the width, and therefore its text is centered through the whole width of the dialog

bjarnef:
@mikecp yes, I must have experimented with some styling in the browser at that time. By default text is aligned in center of buttons.

I have now aligned the text to the left.

mikecp (approved):

mikecp:
Thanks for the update @bjarnef , looks great now 👍

PR #9759, opened 2021-02-03
Use umb-icon component in tree item

bjarnef:
### Prerequisites

### Description
This PR replace the icon `icon-check` using `<i>` element with `<umb-icon>` in tree, where `enablecheckboxes` is used, e.g. in Move dialog.

GhM6FNaZ5c


nathanwoulfe (approved):
Nice work, thanks @bjarnef

PR #9762, opened 2021-02-04
SurfaceController routing for netcore

Shazwazza:
Gets SurfaceController routing for netcore working with form rendering, postbacks and redirects.

* Auto routes SurfaceControllers
* Auto routes front-end UmbracoApiControllers
* UmbracoRequestBegin/UmbracoRequestEnd moved to core, doesn't have an HttpContext requirement
* No more events on IRequestAccessor
* Fixes the DatabaseServerMessengerNotificationHandler mess
* Moves integration controller tests to correct namespace/folder
* Don't rely on route values with key UmbracoRouteDefinitionDataToken anymore. That was based on old ways of working with DataTokens in routes. We now use http features which are strongly typed and simplify a ton of the code.
* Controller/Action discovery now done natively with aspnetcore which has optimized perf/caches using IActionSelector (within ControllerActionSearcher)
* UmbracoRouteValues changed to simply store the result of IActionSelector as a ControllerActionDescriptor since this contains all the information we need and is actually required for re-routing which saves us another controller lookup when proxying with UmbracoPageResult
* HtmlHelper extensions updated to correctly output an umbraco form and always outputs anti-forgery input field. Re-enables being able to resolve a URL/link for a surface controller. Adds a new method to be able to directly render the hidden surface controller route field. This will be useful if people wish to not use BeginUmbracoForm and instead want to use a normal html form, or a tag helper form, this allows them to output the required field manually.
* SurfaceController updates with `[AutoValidateAntiforgeryToken]` to always validate antiforgery
* RedirectToUmbracoPageResult fixed and simplified, we retain temp data simply with aspnetcore's IKeepTempDataResult 
* UmbracoPageResult fixed and simplified. The way that we proxy controller execution is much nicer in aspnetcore. To retain ViewData we have a new ProxyViewDataFeature which saves and then is re-read out and applied on the RenderController base class.
* Deletes a TON of old code

## Testing

Create a controller - I would advise to test your own logic but will provide test code below: 

```cs
// test filter to ensure that filters applied to surface controllers and custom render controllers
// are executed even though we are proxying between controllers.
public class MyTestFilterAttribute : ActionFilterAttribute
{
    public override void OnActionExecuting(ActionExecutingContext context)
    {
        // TODO: DO something to validate its executing, or breakpoint
        base.OnActionExecuting(context);
    }
}

// Custom surface controller used for form post backs
[MyTestFilter]
public class SearchController : SurfaceController
{
    public SearchController(IUmbracoContextAccessor umbracoContextAccessor, IUmbracoDatabaseFactory databaseFactory, ServiceContext services, AppCaches appCaches, IProfilingLogger profilingLogger, IPublishedUrlProvider publishedUrlProvider)
        : base(umbracoContextAccessor, databaseFactory, services, appCaches, profilingLogger, publishedUrlProvider)
    {
    }

    [HttpPost]
    public IActionResult Search(string term)
    {
        // NOTE: Normally for this I would just have a GET form to the 
        // page accepting the query string to do the searching.
        var qry = QueryString.Create("term", term);
        return RedirectToCurrentUmbracoPage(qry);
    }

    [HttpPost]
    public IActionResult SubmitForm(FormModel model)
    {
        if (!ModelState.IsValid)
        {
            // If there are validation errors, set some ViewData
            // and render the current page (no redirecting)
            ViewBag.ErrorHeader = "YOU HAVE SOME ERRORS";
            return CurrentUmbracoPage();
        }

        // TODO: Do something with the model

        // Everything was successful, add a message to TempData since we'll redirect
        TempData["SuccessMessage"] = "YAY!";

        // Do the redirect
        return RedirectToCurrentUmbracoPage();
    }
}

// Custom render controller for route-hijacking
// For a document type named "Hello"
[MyTestFilter]
public class HelloController : RenderController
{
    public HelloController(ILogger<RenderController> logger, ICompositeViewEngine compositeViewEngine, IUmbracoContextAccessor umbracoContextAccessor)
        : base(logger, compositeViewEngine, umbracoContextAccessor)
    {
    }

    public override IActionResult Index()
    {
        // TODO: DO something to validate its executing, or breakpoint
        return base.Index();
    }
}

// The model used for the form
public class FormModel
{
    [Required]
    public string FirstName { get; set; }

    [EmailAddress]
    public string Email { get; set; }
}
```

In your root `_ViewImports.cshtml` page have some basic imports and tag helpers:

```cshml
@using Umbraco.Extensions;
@using Umbraco.Web.PublishedModels
@addTagHelper *, Microsoft.AspNetCore.Mvc.TagHelpers
```

In your view for your "Hello" page type, render a search form:

```cshtml
@using (Html.BeginUmbracoForm<SearchController>("Search"))
{
    <div class="input-group input-inline">
        <input class="form-input" type="text" placeholder="search" name="term">
        <button type="submit" class="btn btn-primary input-group-btn">Search</button>
    </div>
}
```

And include a partial view to another form like:

```cshtml
<partial name="~/Views/Partials/MyForm.cshtml" model="new MyNamespace.FormModel()" />
```

Create a partial view for this form at the ~/Views/Partials/MyForm.cshtml

```cshtml
@model Umbraco.Web.UI.NetCore.Controllers.FormModel


@using (Html.BeginUmbracoForm<SearchController>("SubmitForm"))
{
    @if(ViewData.ModelState.ErrorCount > 0)
    {
        <div>@ViewBag.ErrorHeader</div>

        <ul>
        @foreach(var m in ViewData.ModelState)
        {
            if (m.Value.Errors.Count > 0) {
                <li>
                    @m.Key
                    <ul>
                    @foreach(var e in m.Value.Errors)
                    {
                        <li>@e.ErrorMessage</li>
                    }
                    </ul>
                </li>
            }            
        }
        </ul>
    }

    @if(TempData.TryGetValue("SuccessMessage", out var msg))
    {
        <div>@msg</div>
    }    

    <div class="form-group">
        <div class="col-3 col-sm-12">
            <label class="form-label" for="input-example-1">First name</label>
        </div>
        <div class="col-9 col-sm-12">
            <input asp-for="FirstName" class="form-input" placeholder="FirstName">
        </div>
    </div>

    <div class="form-group">
        <div class="col-3 col-sm-12">
            <label class="form-label" for="input-example-1">Email</label>
        </div>
        <div class="col-9 col-sm-12">
            <input asp-for="Email" class="form-input" placeholder="Email">
        </div>
    </div>

    <button type="submit" class="btn btn-primary input-group-btn">Search</button>
}
```

Once that is all setup:

* Your custom HomeController will execute for all pages of type Home (i.e. Hijacked Routes). 
* Your custom filter will execute for all actions for both the Home and Search controllers
* The forms should work
  * Searching will perform a redirect to the current page with the query string
  * Submitting the form 
    * test that server side validation works and that the ModelState errors show what is wrong without a redirect and that the ViewData data is carried through and displayed
    * test that when everything is valid that you get a redirect and the TempData is carried through and displayed

Then... 

* [ ] Also test the other SurfaceController methods: `RedirectToUmbracoPage` (all overloads), `RedirectToUmbracoUrlResult`

Then....

* [ ] Also test that route hijacking is working at the template/action level. For example, with the above we know the controller executes on it's `Index` action, but create an action with the same name as the template. By default for a content type named Hello you will get a template named `Hello` - so create an action called `Hello` and ensure that it gets executed and not `Index`.


src/Umbraco.Web.Website/Controllers/SurfaceController.cs
```diff
@@ -37,64 +40,51 @@ protected virtual IPublishedContent CurrentPage
         {
             get
             {
-                var routeDefAttempt = TryGetRouteDefinitionFromAncestorViewContexts();
-                if (routeDefAttempt.Success == false)
+                UmbracoRouteValues umbracoRouteValues = HttpContext.Features.Get<UmbracoRouteValues>();
+                if (umbracoRouteValues == null)
```
bergmania, line 44:
```suggestion
                if (umbracoRouteValues is null)
```

src/Umbraco.Web.Website/Extensions/HtmlHelperRenderExtensions.cs
```diff
@@ -33,34 +35,20 @@ namespace Umbraco.Extensions
...
         /// </summary>
-        /// <param name="helper"></param>
-        /// <param name="partial"></param>
-        /// <param name="area"></param>
-        /// <param name="model"></param>
-        /// <param name="viewData"></param>
-        /// <returns></returns>
         public static IHtmlContent AreaPartial(this IHtmlHelper helper, string partial, string area, object model = null, ViewDataDictionary viewData = null)
         {
             var originalArea = helper.ViewContext.RouteData.DataTokens["area"];
```
bergmania, line 54: I think we should consider having `"area"` as a constant. We use it a magic string all over

bergmania (commented):
Overall it looks super good 💪 

I'm not a huge fan of the new namespace of the testserver tests. Maybe the namespace is okay, but then I would like anther postfix to the classes. 
I think there is a difference between regular integration tests of a controller, that new up the controller and use it, and then these tests that uses the testserver.

Another thing we could consider is to have the following strings as public constants: `"controller"`, `"action"`, `"area"` and `"id"`. We use these all over the codebase.

bergmania:
When I test with the following controller + view I can't post the form. What am I doing wrong:

##### BergTestController.cs
```cs
using System.ComponentModel.DataAnnotations;
using Umbraco.Core.Cache;
using Umbraco.Core.Logging;
using Umbraco.Core.Persistence;
using Umbraco.Core.Services;
using Umbraco.Web.Routing;
using Umbraco.Web.Website.Controllers;

namespace Umbraco.Web.UI.BackOffice
{
    public class BergTestController : SurfaceController
    {
        public BergTestController(IUmbracoContextAccessor umbracoContextAccessor, IUmbracoDatabaseFactory databaseFactory, ServiceContext services, AppCaches appCaches, IProfilingLogger profilingLogger, IPublishedUrlProvider publishedUrlProvider) : base(umbracoContextAccessor, databaseFactory, services, appCaches, profilingLogger, publishedUrlProvider)
        {
        }

        [HttpGet]
        public IActionResult CurrentUmbPage()
        {

            return View();
        }


        [HttpPost]
        public IActionResult CurrentUmbPage(BergModel model)
        {
            if (!ModelState.IsValid)
            {
                return CurrentUmbracoPage();
            }

            return View();
        }


    }

    public class BergModel
    {
        public string Type { get; set; }

        [Range(10,20)]
        public int Id { get; set; }
    }
}
```

##### Views/CurrentUmbPage.cshtml
```cshtml
@using Umbraco.Web.UI.BackOffice
@model Umbraco.Web.UI.BackOffice.BergModel

@{
    ViewBag.Title = "title";
}
<body>
<h2>Berg</h2>

@using (Html.BeginUmbracoForm<BergTestController>("CurrentUmbPage"))
{
    <div asp-validation-summary="ModelOnly" class="text-danger"></div>
        <div class="form-group">
            <label asp-for="Id" class="control-label"></label>
            <input asp-for="Id" class="form-control" />
            <span asp-validation-for="Id" class="text-danger"></span>
        </div>
        <div class="form-group">
            <label asp-for="Type" class="control-label"></label>
            <input asp-for="Type" class="form-control" />
            <span asp-validation-for="Type" class="text-danger"></span>
        </div>

        <div class="form-group">
            <input type="submit" value="Save" class="btn btn-primary" />
        </div>

}
</body>

```

Not sure if it is just the ordering in `UmbracoRouteValueTransformer.TransformAsync` or another deeper issue.

src/Umbraco.Web.Common/Controllers/RenderController.cs
```diff
@@ -149,6 +150,20 @@ public override async Task OnActionExecutionAsync(ActionExecutingContext context
...
+
+                    // Check if there's a ProxyViewDataFeature in the request.
+                    // If there it is means that we are proxying/executing this controller
+                    // from another controller and we need to merge it's ViewData with this one
+                    // since this one will be empty.
+                    ProxyViewDataFeature saveViewData = HttpContext.Features.Get<ProxyViewDataFeature>();
+                    if (saveViewData != null)
+                    {
+                        foreach (KeyValuePair<string, object> kv in saveViewData.ViewData)
+                        {
```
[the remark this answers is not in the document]
Shazwazza: The data will 100% always be empty because at this stage the controller context has just been created

src/Umbraco.Web.Website/Extensions/HtmlHelperRenderExtensions.cs
Shazwazza: Yep, will adjust these constants. But good find, since I'm unsure this method is even necessary or works this way anymore. I don't think 'area' in DataTokens is a thing but I'll need to test these things. This also reminds me that I need to test areas with surface controllers, etc....

Shazwazza:
> I'm not a huge fan of the new namespace of the testserver tests. Maybe the namespace is okay, but then I would like anther postfix to the classes.
> I think there is a difference between regular integration tests of a controller, that new up the controller and use it, and then these tests that uses the testserver.

I do understand that those are 2 different types of integration tests but hiding controller tests under the TestServerTest folder I don't think is clear/consistent either. 

Could we have a naming convention for files like `ContentControllerTests.Server.cs` and `ContentControllerTests.cs` ?

Shazwazza:
> When I test with the following controller + view I can't post the form. What am I doing wrong:

Not sure, that code works just fine for me. But your SurfaceController is incorrect, from a SurfaceController action you will generally never return View() since that is just going to render the view associated with the controller, you will typically either redirect with `RedirectToCurrentUmbracoPage();` or return the current page with `CurrentUmbracoPage();`

bergmania:
> How are you rendering `Views/CurrentUmbPage.cshtml` ?

I just visited the page directly: https://localhost:44331/umbraco/surface/bergtest/CurrentUmbPage

I see it works as expected if I do it correctly. I just wasn't aware the other kind of usage was not supported.

We could consider overriding the View methods on SurfaceController. 

```
public override ViewResult View() => throw new NotSupportedException("Not supported in SurfaceControllers");
```

bergmania:
I merged this one. Related stuff can be made in separate PRs.

Shazwazza:
> I just visited the page directly: https://localhost:44331/umbraco/surface/bergtest/CurrentUmbPage

Yep these routes are really just for postbacks. For the most part SurfaceControllers are only for: post backs and 'child actions' (but those don't exist anymore). I created a separate task to test view components with forms to ensure they work.

> We could consider overriding the View methods on SurfaceController.

Might be a thing but there's potential that in some edge case this is what someone would want to do. I don't know what that case would be though

PR #9770, opened 2021-02-08
Netcore: Fix missing application url

bergmania:
Fix issues with the current ApplicationUrl.
- Now it is available on IHostingEnvironment
- Now it is set in UmbracoRequestMiddelware


nikolajlauridsen:
Looks good 👍, tested it out and can confirm I no longer get any "No umbracoApplicationUrl for service (yet), skip." warnings.

PR #9774, opened 2021-02-08
Fix: Verify that grid model is JSON object and not IPublishedContent

jakob-bagterp:
The null check for Model.section fails in the event that:

* the grid editor has been added as composition to a page
* the partial `@Html.GetGridHtml(Model, "pageContentGrid")` has been added to the page template
* and the page hasn't been published with the new grid content yet

In this case the `Model` passed into `Html.GetGridHtml` will be `IPublishedContent` and not a `JObject`, and this will cause the grid render to fail.

Screenshot 2021-02-08 160455

### Prerequisites


nul800sebastiaan:
Thanks @jakob-bagterp - good catch! 👍

Issue #9732, opened 2021-01-29
Controller inherit from UmbracoAuthorizedApiController or UmbracoAuthorizedJsonController are not testable


sebafelis:
I can't create unit test of controller inherit from UmbracoAuthorizedApiController or UmbracoAuthorizedJsonController test running ends with NullReferenceException exception. 

## Umbraco version

I am seeing this issue on Umbraco version: 8.10.1

Reproduction
------------

Create any unit test of controller inherit from UmbracoAuthorizedApiController or UmbracoAuthorizedJsonController.
i.e.

```
public class SomeBackofficeController: UmbracoAuthorizedApiController
    {
          public SomeBackofficeController(IGlobalSettings globalSettings, IUmbracoContextAccessor umbracoContextAccessor,
                                ISqlContext sqlContext, ServiceContext services, AppCaches appCaches,
                                IProfilingLogger logger, IRuntimeState runtimeState, UmbracoHelper umbracoHelper) : base(globalSettings,
                                    umbracoContextAccessor, sqlContext, services, appCaches, logger, runtimeState,
                                    umbracoHelper) 
           {
           }
    }
```

```
[TestFixture]
    public class SomeBackofficeControllerTest : UmbracoBaseTest
    {
        private SomeBackofficeController controller;

        [SetUp]
        public override void SetUp()
        {
            base.SetUp();

            this.controller = new SomeBackofficeController(Mock.Of<IGlobalSettings>(), Mock.Of<IUmbracoContextAccessor>(), Mock.Of<ISqlContext>(), ServiceContext, AppCaches.NoCache, Mock.Of<IProfilingLogger>(), Mock.Of<IRuntimeState>(), UmbracoHelper);
        }

        [Test]
        public async Task BackofficeController_ShouldBeTest()
        {
            //Anything in hear
        }
    }
```

### Bug summary

After the test is running it's throw an NullReferenceException exception. 
I looked at the code and I found that `UmbracoAuthorizedApiController `and `UmbracoAuthorizedJsonController `class has no constructor with `UmbracoMapper umbracoMapper` argument like the other Umbraco controller class, so `UmbracoApiControllerBase` try to get a mapper from `Umbraco.Web.Composing.Current`.

### Specifics

Thrown exception: 

> System.NullReferenceException : Object reference not set to an instance of an object.
>   Stack Trace: 
>     FactoryExtensions.GetInstance[T](IFactory factory)
[...]

PR #9776, opened 2021-02-08
9732: Don't use obsolete method

patrickdemooij9:
Issue: https://github.com/umbraco/Umbraco-CMS/issues/9732

I wasn't able to reproduce the bug itself as I wasn't sure what the UmbracoBaseTest and SetUp did in the issue example. 
However, I was able to find all references of the obsolete ctor and move them to the actual ctor.

However, this is most likely a breaking change if people inherit/use the ctor of any of the controllers that I had to change. Should I make obsolete ctors for each controller?


sebafelis (approved):

dawoe:
Hi,

This should be marked as breaking change. By changing the constructor I know for a fact my current unit tests will break after a upgrade.

Dave

Issue #9569, opened 2020-12-17
Image Cropper buttons have no type, so are "pressed" when pressing enter in other input fields


matthewcare:
Image Cropper buttons (remove file, etc) are missing a type, which causes them to be executed when pressing enter in other from fields. As shown in this example.

![cropper](https://user-images.githubusercontent.com/1469061/102491124-a93ba680-4067-11eb-912f-f6ee7d347628.gif)

## Umbraco version
8.9.1

Reproduction
------------

Create a new document type, with an Image Cropper property.
Create a new page using this document type, and upload an image to the cropper.
Press enter in the document name field
The image is removed.

nul800sebastiaan:
Interesting find! Looks like it's been like that for at least a couple of years now! 😅 
https://github.com/umbraco/Umbraco-CMS/blame/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/imagecropper/imagecropper.html#L31

We'd love a little help getting that fixed!

umbrabot:
Hi @matthewcare,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

patrickdemooij9:
PR: https://github.com/umbraco/Umbraco-CMS/pull/9777

@nul800sebastiaan 
I did notice that the issue itself was already fixed. It wouldn't clear the image on enter anymore however, the fix wasn't done on the crops yet. I did see another issue with the ImageCropper in that it doesn't seem to be working with crops at all:

![ezgif com-gif-maker (1)](https://user-images.githubusercontent.com/11466511/107271282-4b668200-6a4c-11eb-97d0-d5064834d745.gif)

Should I make a new issue for that or will you?

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/9777

nul800sebastiaan:
@patrickdemooij9 I can't reproduce what you're seeing in that last gif, not sure if that's still an issue?

patrickdemooij9:
[...]

PR #9777, opened 2021-02-08
9569: Fix image cropper buttons

patrickdemooij9:
Issue: https://github.com/umbraco/Umbraco-CMS/issues/9569 (Has the old situation however, some parts were already fixed)

With the fix in place:
umbraco 2

Just noticed that I forgot to record with the mouse. Hopefully, the gif is still enough!


nul800sebastiaan:
Nice one @patrickdemooij9 - thanks for that! 👍

PR #9778, opened 2021-02-09
Fixes issue with broken caches used for user permissions

Shazwazza:
Calculating start nodes is expensive and this is supposed to be cached but the cache was not working.

This issue must have been around for a very long time but has gone unnoticed and unfortunately causes a gigantic amount of unnecessary and duplicate DB queries.

It's easy to see this by just having mini profiler enabled in the back office with `/umbraco?umbdebug=true`. For example, mini profiler will show a red asterisk when there are duplicate DB queries. This is before this fix:

image

And this is after - this is not comparing the ms, just noting that the duplicate SQL queries are gone... and actually that almost any SQL queries in those requests are gone.

image

The problem was that the cache for this was trying to piggy back off of the IUser object directly (which is cached) but the `AdditionalData` is not cloned in or out of the cache so anything put there was lost, except for during the current request. 

This is fixed by using a runtime cache and custom cache keys and ensuring this cache is cleared if the user is every saved.

These checks are used quite a bit in the back office as you can see by how many code changes were required when updating some method signatures. This goes to show that this has a significant impact and should potentially be back ported to some previous minors.

---
_This item has been added to our backlog [AB#10425](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10425)_

### DIFF
--- a/src/Umbraco.Core/Models/Identity/IdentityMapDefinition.cs
+++ b/src/Umbraco.Core/Models/Identity/IdentityMapDefinition.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using Umbraco.Core.Cache;
+using Umbraco.Core.Composing;
 using Umbraco.Core.Configuration;
 using Umbraco.Core.Mapping;
 using Umbraco.Core.Models.Membership;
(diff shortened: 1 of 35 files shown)


src/Umbraco.Core/Models/Identity/IdentityMapDefinition.cs
```diff
@@ -11,14 +12,24 @@ public class IdentityMapDefinition : IMapDefinition
         private readonly ILocalizedTextService _textService;
         private readonly IEntityService _entityService;
         private readonly IGlobalSettings _globalSettings;
+        private readonly AppCaches _appCaches;
 
+        [Obsolete("Use constructor specifying all dependencies")]
         public IdentityMapDefinition(ILocalizedTextService textService, IEntityService entityService, IGlobalSettings globalSettings)
```
bergmania, line 18: I guess this ctor should call the new non-obsolete ctor and pass in Current.AppCaches? Otherwise it seems like a NullReferenceException

bergmania (commented):
Looks good.. Just found a single small issue, I will fix up

bergmania (approved):

bergmania:
Confirmed it woked on the Starterkit About-Us page.. 💪

PR #9780, opened 2021-02-09
Fix mini profiler so that it actually profiles the Content APIs in the back office

Shazwazza:
Mini profiler was not profiling Umbraco Content API endpoints in the back office because mini profiler has a default ignore path of `/content/` and this path is applied with a 'Contains' check. This means that all requests to get/post content in the back office was not profiled.

This fixes it and you can test by using mini profiler with `/umbraco?umbdebug=true` and seeing that things like /umbraco/backoffice/UmbracoApi/Content/PostSave are profiled. You can also see all latest sessions profiled by going to `/mini-profiler-resources/results-index`


bergmania (approved):
Changes looks good

PR #9799, opened 2021-02-10
9769 - shift index increment before tryget to allow loop to continue if current iteration exits early

### LINKED ISSUE #9769: Infinite loop / unreachable index increment in ContentTypeCommon repository
See below - the compositionIx variable will not be reached if the condition on L174 is false (it's negated, so will break out of the current loop iteration when TryGetValue fails), which means we loop FOREVER as compositionIx is never incremented.

https://github.com/umbraco/Umbraco-CMS/blob/eb9c4e4d828bf291d5c09aa42989225b288af8eb/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeCommonRepository.cs#L168-L178

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #9769 

### Description
I ran into this issue in a v8 migration, where the migration completes but the site failed to boot - code was falling into an infinite loop where it kept checking the same composition parentId, and would continue to do so until the inevitable heat-death of the universe. I didn't have that long to wait, hence the PR.

I'm not sure though how to test this - if it was an issue affecting all installs, it would have been noticed before, so I'm thinking it was likely due to something special in my database (db was updated from inheritance to composition a while back, so may have some stale data from that change). Can confirm though with the change my site starts and everything in the backoffice appears correct...

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeCommonRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeCommonRepository.cs
@@ -171,9 +171,11 @@ namespace Umbraco.Core.Persistence.Repositories.Implement
                 while (compositionIx < compositionDtos.Count && compositionDtos[compositionIx].ChildId == contentType.Id)
                 {
                     var parentDto = compositionDtos[compositionIx];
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Yeah, I can see that potential for an infinite loop immediately. Regardless of who might (not) run into this, the logic here is now correct. Thanks @nathanwoulfe! 👍

PR #9800, opened 2021-02-11
Netcore:  WIP Update to NPoco 5

bergmania:
# Work in progress 
- Currently a test is failing.

# Details
- Updated NPoco to version 5
  - Part of this requires the namespace `System.Data.SqlClient` to be changed to `Microsoft.Data.SqlClient`.
  - Part of this is to use the NuGet package  `NPoco.SqlServer` (which has a dependency on `NPoco`)

# Why do the test fail?
My finding so far point toward this change in NPOCO:
https://github.com/schotime/NPoco/commit/8ead03fa97a3592df6b707592cd1f231dce2b6d8#diff-cd1456f38661dfc60c126f2a0d2cfae87b720d86b5e68bf1fa292dde782528a7R94-R95

In NPoco 5 the method `ConvertFromOldConvention` do not update the `posName.Name` for the two nested properties `ptAlias` and `ptName`, because the Level of these are 2, but the level in the code is 3, so this section is skipped.

This do not happed in NPoco 4, because level is only 1, when it try  `ptAlias` .
The reason this is lower in NPoco 4, is because the section updating the level is not executed for all the `NodeDto__*`

image

I Npoco 5, the new check for `MemberInfoKey` (https://github.com/schotime/NPoco/commit/8ead03fa97a3592df6b707592cd1f231dce2b6d8#diff-cd1456f38661dfc60c126f2a0d2cfae87b720d86b5e68bf1fa292dde782528a7R94-R95) means that this code is now executed, and therefore the Level becomes 3.

The order in which it is executed is not changed between NPoco 4 and 5:
image

This is how `used` looks when the foreach tries `ptAlias` in NPoco 4. In NPoco 5 `used` also contains all the `NodeDto__*`:
image

# Test
- Stuff should still work, but at least all tests should pass


bergmania:
I still think we should create a bug report for npoco. This workaround only fixes this single query, but we could easy hit similar issues in some of the queries not covered by tests.

bielu:
@bergmania if that change is going happens, it has to happen before v9 is released, NPOCO 5 contains breaking changes, so change of dependency will be breaking change by it self.

bergmania:
> @bergmania if that change is going happens, it has to happen before v9 is released, NPOCO 5 contains breaking changes, so change of dependency will be breaking change by it self.

We know, but still no feedback on https://github.com/schotime/NPoco/issues/642

bielu:
@bergmania are we going to introduce that in v9? if so I think it should go as soon as possible as it is breaking change so better to break stuff in 9.1 than 9.15 :)

bergmania:
> @bergmania are we going to introduce that in v9? if so I think it should go as soon as possible as it is breaking change so better to break stuff in 9.1 than 9.15 :)

We are not going to introduce this in v9, but we also expect v10 in Q2 next year, according to our plan

p-m-j:
Completed in v10 via #11880

PR #9801, opened 2021-02-11
Netcore: Align namespaces

nikolajlauridsen:
## Overview
This PRs is focused on aligning namespace, I have:

1. Aligned namespaces to the new schema of `Umbraco.Cms.*`
2. Ensured that namespaces match assemblies.
3. Put all public extension method in the `Umbraco.Extensions` namespace
4. Moved all extension in the root of a project into an Extensions folder

I have not: 
* Changed namespaces in Infrastructure, since it will be a separate PR since it's a bit more complex
* I haven't touched the legacy projects such as `Umbraco.Web`.
* Change the namespace of extension methods in the `Umbraco.Tests.Common` project to `Umbraco.Extensions`, my reason for this is that they're exclusively used in tests, but let me know if I should move these as well.
* I haven't change the namespace of the import resources in integration tests (`Umbraco.Tests.Services.Importing`) since this must be done in a generated file that warns you to not try and change it.

## Details
During the refactoring, I ran into some hiccups.

* I had to rename `Umbraco.Core\Extensions\CacheHelperExtensions` to `CoreCacheHelperExtensions`, because there's a second `CacheHelperExtensions` in `Umbraco.Web.Common\Extensions` that uses a constant string in the core version of `CacheHelperExtensions`, causing it to not be able to find the string because they share a namespace, so it would try to find the string on itself
* After changing the namespace in `EndpointRouteBuilderExtensions` to `Umbraco.Extensions` `NUglify.Helpers` was no longer required
* In `UmbracoBuilderExtensions.AddUmbracoSqlCeSupport` I had to change the string litterals used with `GetType` to match the new namespace (`Umbraco.Cms.Persistence.SqlCe`)
* In `DatabaseConfigureStep` I also had to change the string litereal in `IsSqlCeAvailable` from `"Umbraco.Persistence.SqlCe.SqlCeSyntaxProvider, Umbraco.Persistence.SqlCe"` to `"Umbraco.Cms.Persistence.SqlCe.SqlCeSyntaxProvider, Umbraco.Persistence.SqlCe"`
* I had to remove `IsEqual` and `IsNotEqual` from `Umbraco.Web.Website\PublishedContentExtensions` because they were identical to `Umbraco.Core\PublishedContentExtensions` and now resides in the same namespace, they were, however, both identical in signature and code. 
* I had to update the string in `Constants-ModelsBuilder` to match the new namespace
* I updated the imports in `Umbraco.Cms.ModelsBuilder.Embedded.Building.Builder` to match the new namespaces
* Updated the `ViewHelperTests`, `BuilderTests` and `UdiTests` to also reflect the new namespaces
* I also updated the strings in `ViewHelper` to reflect new namespaces

That should cover the changes I've made that might be interesting, the rest is just namespace changes, and a bit of style cop cleanup when I spotted something.

## Testing
Everything should be functionally the same, so the main goal is to test that everything still works as normal, creating content etc.


elit0451 (approved):
Good job @nikolajlauridsen 🌟 

Even github couldn't handle the changes in this massive PR 😀 : 
_"The diff you're trying to view is too large. We only load the first 3000 changed files."_

But I have done various tests of the back-office and things seem to work, so I will go ahead and merge it in.

Issue #3560, opened 2018-11-08
Update icon fonts to use svgs


MMasey:
I'm not sure if this is a feature of a bug.

Icon fonts have a fundamental accessibility problem in that if a person using a tool to change all fonts on a site to be more readable, in the case of those with server dyslexia, all the icon fonts will be replaced with a missing glyph character. This can break the UI, especially if it's dependent on icons.

For example 
![image](https://user-images.githubusercontent.com/6573613/48204771-89066d00-e362-11e8-8dc6-671d17155f33.png)

Obviously this example is a v7 site, but i think the issue will apply for v8, as such this issue only applie for v8.

Github went through a similar transition and detailed there approach at https://blog.github.com/2016-02-22-delivering-octicons-with-svg/. I think a similar approach could work for Umbraco, and it could potentially make icons easier to customise.

nul800sebastiaan:
@MMasey Thanks Mike! Can you update this issue with the problems you've encountered when you were working on this at the UK hackathon. Might not be worth the trouble to convert it over for now, but I can't remember exactly what the stumbling blocks were?

MMasey:
Sure, no problem. Most of the problems come around the fact that the use of icon fonts is so heavily embedded into the backoffice, although that's more of a time problem than a complexity one.

One on the other problems was that my approach is to use inline SVGs. There would be a folder, potential in the `umbraco` folder that could contain each of the individual SVG icon files. In order to get this inline, it would require reading the file and printing out the raw SVG data into the html, which is obviously a security issue. Now it is possible to do some validation on an SVG file to ensure that it contains SVG code and not something malicious, although i don't know how it works... yet. The following sites do some form of validation so i know it's possible

[https://validator.nu/](https://validator.nu/)
[https://jakearchibald.github.io/svgomg/](https://jakearchibald.github.io/svgomg/)

They don't prevent you inserting `script` tags into the svg for example, but from my research so far, scripts inside SVGs don't get fired.

It think we also mentioned that if they're inside the protected umbraco folder, it should be a bit easier to keep them secure. The reason for this approach is that it potentially makes it really easy for people to add their own icons. I have a feeling this problem may have already been solved for other parts of the backoffice system too. It may also only be a security issue if a hacker gets access to the server itself, which if it's that's the case, there would be far worse problems to deal with.

I'm was also thinking about having an extension method that you can use to say "look at this folder and get the SVGs for the backoffice" which then puts the onus of security on the person making use of the extension.

[...]

Issue #6495, opened 2019-10-01
Remove Microsoft.Net.Compilers dependency


ronaldbarendse:
The Umbraco 8 NuGet package includes a dependency on `Microsoft.Net.Compilers (>= 2.10.0 && < 2.999999)`: https://github.com/umbraco/Umbraco-CMS/blob/178b1ccf40aa67815be77698a259b5db31ab9013/build/NuSpecs/UmbracoCms.nuspec#L33

As commented earlier (https://github.com/umbraco/Umbraco-CMS/issues/4493#issuecomment-489080794), this is not required as `Microsoft.CodeDom.Providers.DotNetCompilerPlatform (>= 2.0.0)` already includes the compiler binaries (for dynamic compilation of views): https://github.com/aspnet/RoslynCodeDomProvider#version-200.

Installing `Microsoft.Net.Compilers` replaces the built-in compiler that's used when building the project and because of the version range, it's not possible to update this to the latest version to make use of new C# features. It's also slower than using the compiler Visual Studio ships with and version 2.10.0 is actually the same as the minimum required VS version for local development: https://our.umbraco.com/Documentation/Getting-Started/Setup/Requirements/#local-development.

Reproduction
------------

### Bug summary
Can't use the latest C# features, as the compiler gets replaced by an out-of-date version because it's included as a NuGet dependency.

### Workaround
It's possible to forcefully remove the package `PM> Uninstall-Package Microsoft.Net.Compilers -Force` (gives errors when updating Umbraco), remove the build-target from the project file (needs to be repeated after updating the package) or move code to a seperate library. This should however not be required...

nul800sebastiaan:
Sure thing!

umbrabot:
Hi @ronaldbarendse,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

ronaldbarendse:
PR is created, I'm going for that `hacktoberfest/contrib-2019` label 😉

SteveVaneeckhout:
[...]

PR #9803, opened 2021-02-11
Remove HtmlSanitizer

### LINKED ISSUE #9615: Update HtmlSanitizer
As one that's been using AngleSharp for quite a while on earlier versions of Umbraco 8 (8.6.4), i was a bit confused when trying to upgrade to 8.10.1 today.

Seeing that it has a dependency on a 2 year old version of HtmlSanitizer, and that in turn having a dependency on a 3 year old version of AngleSharp, stopping me from upgrading, unless i downgrade AngleSharp.

Any reason for these old versions suddenly appearing, or would it be possible to include a newer version of HtmlSanitizer, and by that a newer AngleSharp?

nul800sebastiaan:
Due to a security issue, HtmlSanitizer has decided to pull all v4 versions from NuGet: https://github.com/umbraco/Umbraco-CMS/issues/9615#issuecomment-773305267

We tried updating to v5 of the package but it won't build on the CI server nor locally: https://github.com/umbraco/Umbraco-CMS/issues/9615

After exploring some options we came to the conclusion that

- We only added sanitization of SVGs for the SVG icons since v8.8 (https://github.com/umbraco/Umbraco-CMS/issues/3560#issuecomment-484289060). 
- This was out of an abundance of caution, the SVGs shipped with Umbraco of course don't contain any scripting
- The sanitization was added for packages/additional icon(pack)s people might want to load 

We came to the conclusion that installing icons poses the same risk that is already present when installing any other packages: they can contain malicious javascript and/or dlls and/or other malicious content. 

Therefore, we've decided it is okay to remove the SVG sanitizer while we look for alternatives, so as to fix people's NuGet installs for now.

Furthermore: this pinned the version of Anglesharp people were using, and working around that was quite a pain, this also resolves that problem, we don't rely on AngleSharp anywhere so people can use any version they want again.

After merging, this needs to be backported to v8.8.x and above and released as a patch release for 8.8 through 8.11.

Fixes #9615


nul800sebastiaan:
Note: I don't understand why some people have problem restoring the package, the v4 version has not been deleted from NuGet, but it's been unlisted. Even after deleting my local NuGet cache I have no problems installing Umbraco nor upgrading from a version lower than 8.8.

emma-hq (approved):

lars-erik:
@nul800sebastiaan I think it might have to do with your nuget settings with regard to taking the uppermost or lowermost versions for the indirect dependencies.

I was able to upgrade 8.6 to 8.11 by installing HtmlSanitizer -v 4.0.nnn first.

ronaldbarendse:
@nul800sebastiaan Note that removing the HtmlSanitizer dependency from the Umbraco NuGet packages won't remove the entry in the `packages.config`, but does remove the version restriction (>= 4.0.217 && < 4.999999.0). This will allow it to update to version 5 (when doing an `PM> Update-Package`), which might cause similar issues as discussed in https://github.com/umbraco/Umbraco-CMS/issues/6495#issuecomment-565068508.

Uninstalling can be automatically done in the [`install.ps1`](https://github.com/umbraco/Umbraco-CMS/blob/34e80d86e8c0b754f6b7a02e307f53cb32806bbe/build/NuSpecs/tools/install.ps1) script that runs on NuGet install (including updating to newer versions), although other custom packages might still require it and throw errors (which should be ignored in a try/catch).

Or just put a warning/reminder message in the release notes (and [ReadmeUpgrade.txt](https://github.com/umbraco/Umbraco-CMS/blob/34e80d86e8c0b754f6b7a02e307f53cb32806bbe/build/NuSpecs/tools/ReadmeUpgrade.txt)) to manually uninstall the package if needed?

ronaldbarendse:
@nul800sebastiaan Besides the above comment, this should already be marked as breaking change, because the `Umbraco.Core.Constants.SvgSanitizer` class is removed, right?

nul800sebastiaan:
Correct, it is a breaking change, hopefully not something people are actively using. 

Other than that, it's not in the way if people don't uninstall the NuGet but we'll put it in the version specific upgrade docs.

PR #9804, opened 2021-02-12
Guid was not being mapped to Key

nzdev:
### Prerequisites

### Description
Fix issue where Guid was not being mapped to Key in SQL Select for content types, it was being mapped to UniqueId.


Shazwazza:
Looks like there's some conflicts to resolve, if you can fix that up i'll get it merged. Cheers!

PR #9807, opened 2021-02-12
Fixes: After deleting a member, a reindex is attempted, which fails

### LINKED ISSUE #9684: After deleting a member the index is not updated and an exception is thrown
This is a regression of #8525

This means that when a member is deleted, the index does not remove that data.

---

## Umbraco version

- This happens on the latest `netcore\netcore` branch 
- I've replicated this on the `v8\contrib` branch so could be resolved in either. 

Reproduction
------------

### Bug summary

After deleting a member, a reindex is attempted

### Specifics

After deleting a member, a reindex is attempted, which fails because the member is now null. It seems it is returned as a null item in a list of 1 member

### Steps to reproduce

- After deleting a member, a reindex is attempted and this throws an exception under the hood. 
- You need to have exceptions enabled during debugging in your IDE to see this.

### Expected result

If the member was deleted, I wouldn't expect it to attempt to be reindexed.

### Actual result

When I break on all exceptions after deleting a member, I come to this point (line 258) in `ExamineComponent.cs`, where the member has already been deleted but it tries to reindex it:

image
image

The member service returns a null member object even though it hasn't matched anything against the `Payload` (populated with the now deleted member details):

image
image

The code seems to be trying to run `ReindexForMember` method but fails because the properties are null.

I checked in `v8\contrib` and the same behaviour happens when deleting a member. `DeferedReIndexForMember` is called and then the member properties are null, an exception happens under the hood:

image

Shazwazza:
Fixes #9684 

The member index was not removing deleted members and was actually throwing exceptions when this occurred.

### DIFF
--- a/src/Umbraco.Web/Cache/DistributedCacheExtensions.cs
+++ b/src/Umbraco.Web/Cache/DistributedCacheExtensions.cs
@@ -135,7 +135,10 @@ namespace Umbraco.Web.Cache
         public static void RemoveMemberCache(this DistributedCache dc, params IMember[] members)
         {
             if (members.Length == 0) return;
-            dc.RefreshByPayload(MemberCacheRefresher.UniqueId, members.Select(x => new MemberCacheRefresher.JsonPayload(x.Id, x.Username)));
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Nice fix, thanks @Shazwazza! 👍

Issue #8704, opened 2020-08-21
umb-control-group label-for property does not populate for attribute


lordscarlet:
When using the `umb-control-group` directive, there is a `label-for` attribute. However, this attribute does not populate the `for` attribute of the `<label>` that is created as part of the directive. 

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.1

Reproduction
------------

### Specifics

Navigate to `/umbraco/#/users/users/users?create=true`

The label/input fields do not have a `for` mapping between one another

![image](https://user-images.githubusercontent.com/17951/90895762-a0281a80-e390-11ea-8e9c-7540e82a8753.png)

Looking at the source, I see a couple of problems:

1. In the `umbControlGroup` directive, there is a `labelFor` attribute. However, that is not used for the `<label>` element's `for` attribute. Instead it uses the `alias` property to set a value on the `for` attribute. https://github.com/umbraco/Umbraco-CMS/blob/55c37e0a4efacfce48c3d776b2893dd076625049/src/Umbraco.Web.UI.Client/src/views/components/html/umb-control-group.html#L4
2. Even if `alias` was used when using the `umb-control-group` directive, or (1) was not an issue, in many (all?) instances of the `umb-control-group` directive, the related `<input>` does not have an id, so the `for` attribute would not have any value regardless https://github.com/umbraco/Umbraco-CMS/blob/55c37e0a4efacfce48c3d776b2893dd076625049/src/Umbraco.Web.UI.Client/src/views/users/views/users/users.html#L318-L319
### Expected result

1. The `<label>` should have a `for` attribute with the same value as the `id` of the `<input>` field
2. The `<input>` field should have an `id` attribute that matches the `for` in the `<label>`

### Actual result

1. The `<label>` does not have a value for the `for` attribute
2. The `<input>` does not have an `id` attribute to relate to the `for` attribute

MMasey:
@nul800sebastiaan Can this be tagged with the accessibility label please. Thanks

nul800sebastiaan:
Thanks @lordscarlet ! I see you've added a few issues around accessibility and I'll label them as such and put them in `up for grabs`. I don't know if you intended to work on them or just report them? If you intend to help out with these issue then we're more than happy to look at a PR without having to create an issue for each of them, just so you know! 👍

nul800sebastiaan:
[...]

PR #9810, opened 2021-02-13
Performance and accessibility improvements on the embed screen

RachBreeze:
Added aliases to control groups to help screen readers identify elements within the embed.
Also added support for retrieving the embed item only when the size changes. This improves both performances as the preview is only repopulated if the width/ height values change. It also aids accessibility as the reload of the preview is performed when a user tabs off the width/ height textboxes and the reloading of the preview causes the focus to be reset on the preview.

There is still an accessibility bug, when the width/ height is resized as focus is still reset on the to the preview. This is due to how the preview is being reloaded, the preview is hidden, removing the textboxes for height/ width meaning they don't exist in the screen to be focused on.


nul800sebastiaan:
Good stuff, thanks very much @RachBreeze 👍

RachBreeze:
Sorry, I forgot to say this PR was the start of a series of PRs based on this issue https://github.com/umbraco/Umbraco-CMS/issues/8704

PR #9812, opened 2021-02-14
Make properties full width in small infinite editors

callumbwhyte:
When content is displayed inside a "small" editor window (e.g. in the grid or block list) the property labels take up 50% of the available space, making many properties so small they are unusable.

Before

This isn't so much of an issue in other larger sizes, e.g. medium, as the properties have enough space.

Medium

## Solution

Since the introduction of the "labels-on-top" feature in Umbraco 8.10, I propose properties in small editor windows are treated the same by default. This PR changes the labels and properties in these cases to be full width.

After the change here's how it looks:

After

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/editor/umb-editor.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/editor/umb-editor.less
@@ -35,7 +35,7 @@
         transform: none;
     }
 
-    &--small {
+    &--small .umb-property {
         .control-header {
             float: none;
             width: 100%;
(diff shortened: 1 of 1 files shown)


nathanwoulfe:
@callumbwhyte there's some specificity shenanigans at play - in a small overlay, the new `control-header` styles are overridden by the `.form-horizontal .umb-control-group .control-header` selector. It doesn't necessarily matter since removing the margin on the controls element forces it to clear the floated header, but if you feel like adding a selector to `.umb-editor-small .control-header` to bump up the specificity...

nathanwoulfe (approved):
@callumbwhyte all good - I've added `.umb-property` to the selector to ensure the appropriate specificity

Issue #9813, opened 2021-02-14
NetCore: Url Tracker not being disabled and button still showing 'Disable Url Tracker' after clicking it.


TwoMoreThings:
On Alpha 3, click on the Disable Url tracker button and confirm you want to disable the Url tracker, after this you will see a confirmation notification appear at the bottom to confirm the url tracker has been disabled. However, the button text remains as 'Disable Url tracker', even after a refresh. 

I tested and this is an issue with the Url tracker not being disabled, despite the confirmation message.

This was tested using the latest version of Chrome on Windows but not tested in any other browsers.

regards
t

bergmania:
Hi @TwoMoreThings. Thansk for reporting. I can confirm I can reproduce, and will look into it

bergmania:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/9814. The issue was due to caching of the value, even that the value was updated in the configuration file.

Again, thanks for reporting 💪

PR #9814, opened 2021-02-15
Netcore bugfix: DisableRedirectUrlTrackingValue was cached

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/9813

### Details
- The values of `WebRoutingSettings` was cache due to the injection as `IOptions<WebRoutingSettings>` and not `IOptionsMonitor<WebRoutingSettings>`
- In the method that updates the configuration value, we need to wait the configuration delay time, before we return. Otherwise, the old value will still be read. I couldn't find any way to the set real delay time, so I used the default 250ms and added TODO.


nikolajlauridsen:
Looks good, tested it out as well with no issues 👍

nikolajlauridsen (approved):

PR #9820, opened 2021-02-15
Netcore: Align infrastructure namespaces

nikolajlauridsen:
This PR is ready but since it's a continuation of #9801, Which should be merged first.

In this PR I've aligned the namespaces in Infrastructure, some of these namespaces are now Umbraco.Cms.Infrastructure, while others are Umbraco.Cms.Core, since some of this, would be best to have in Core but must be in infrastructure due to some dependencies, there are quite a few dependencies to Newtonsoft.Json as an example. 

As for what goes where I've used my best judgement, but preferring Umbraco.Cms.Core if I was very doubtful, the reasoning behind this is that it's quite easy to do an automatic refactoring to Umbraco.Cms.Infrastructure, but not Umbraco.Cms.Core because the root namespace of the assembly is Umbraco.Cms.Infrastructure. I've also attempted to do every folder as a separate commit, to make it a bit easier to have an overview of what got moved to what namespace.

I have also moved two files into the Core project because it could be done with no errors, these files are: 
* QueuingEventDispatcher
* LogHttpRequest
  * I also renamed this to LogHttpRequestExtension to follow the naming convention of the other extension methods

The only other thing I've done aside from adjusting namespaces is fixing up the builder in Umbraco.Cms.ModelsBuilder.Embedded.Building, fixing the two todos from the other PR. 

### Testing
The same as #9801, additionally take a look at the new namespaces in Infrastructure to see if you disagree with any of my choices.


bergmania:
Overall seems good.
I changed the Services* namespace. In theory, services should be pure business logic, and all technologies should have been abstracted and handled in infrastructure.

PR #9821, opened 2021-02-16
Custom controller routes within an Umbraco context (i.e. virtual nodes)

Shazwazza:
Previously in v8 there was several ways to do this which required creating conventional routes and assigning a custom handler or using a `EnsurePublishedContentRequestAttribute`. Old docs for all of this is here: 
https://our.umbraco.com/Documentation/Reference/Routing/custom-routes

In netcore we need to be able to do something similar and in netcore this is much simpler. The rule is that if there is an Http Request feature of type `UmbracoRouteValues` assigned, then the request is considered to be executing within an Umbraco context because if `UmbracoRouteValues` then all information is available to Umbraco to execute it's various components.

So with this PR, there's a few ways to have custom controller routes to execute within an Umbraco context:

1. Completely manually - have a custom route/controller (even middleware) and manually assign the `UmbracoRouteValues` as an Http Request feature however you want to do that. To create an `UmbracoRouteValues` generally requires: `IUmbracoContextAccessor` (to access the CleanedUmbracoUrl), `IPublishedRouter` (to create the `IPublishedRequestBuilder`), `IPublishedRequestBuilder` (to set the published content and to build the `IPublishedRequest`), `IPublishedRequest` to assign to the `UmbracoRouteValues`. 
2. Routing a custom controller that implements the `IVirtualPageController` interface. This interface exposes a method: `IPublishedContent FindContent()` and this method just needs to return the content instance that the request will be executing in the context of. Under the hood, we use aspnet's Application Model to apply the `UmbracoVirtualPageFilterAttribute` filter to the actions of this controller. This filter will take care of assigning the `UmbracoRouteValues` to the http request features. This also allows you to use attribute routing for this controller/actions.
3. Routing a custom controller with conventional routing using the typical call to `endpoints.MapControllerRoute` and fluently using a new `.ForUmbracoPage` call which allows specifying a callback to return an instance of `IPublishedContent` to associate with the current request. Behind the scenes, the `ForUmbracoPage` dynamically applies the `UmbracoVirtualPageFilterAttribute` to these routes to take care of assigning the `UmbracoRouteValues` to the http request features.

## Examples

To use these examples, you can either: specify these routes in your Startup.cs file in the `Configure(IApplicationBuilder app)` method **BEFORE** the call to `app.UseUmbracoWebsite();`. For example:

```cs
public void Configure(IApplicationBuilder app)
{
    if (_env.IsDevelopment())
    {
    }

    app.UseEndpoints(endpoints =>
    {
        // ... This is where to put the custom routing
    });

    app.UseUmbracoBackOffice();
    app.UseUmbracoWebsite();
}
```

OR, you can register your own `IStartupFilter` in DI with your own `IUserComposer`. This would be how you would add custom routes via a package/plugin. Example:

```cs
public class MyComposer : IUserComposer
{
    {
        builder.Services.AddTransient<IStartupFilter, MyRoutesStartupFilter>();
    }
}

public class MyRoutesStartupFilter : IStartupFilter
{
    public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next) =>
        app =>
        {
            app.UseEndpoints(endpoints =>
            {
                // ... This is where to put the custom routing
            });

            next(app);
        };
}
```

### Using the `IVirtualPageController` interface with conventional routing

```cs
app.UseEndpoints(endpoints =>
{
    // Create a normal conventional route to your custom controller
    endpoints.MapControllerRoute(
        "Test",
        "/shop/{action}/{id?}",
        new { Controller = "Shop", Action = "Index" });
});
```

This example controller is for a Shop/Product with a route parameter for the product SKU. 

```cs
// Inherits from UmbracoPageController which is an abstract controller for
// front-end rendering pages (RenderController also inherits from this).
// UmbracoPageController has a few helper methods for front-end rendering
// controllers.
// NOTE: This cannot inherit from RenderController which contains a few other
// filters that are not compatible with custom routes. RenderController is
// ONLY for native Umbraco routing (i.e. default + hijacked routes).
public class ShopController : UmbracoPageController, IVirtualPageController
{
    private readonly IUmbracoContextAccessor _umbracoContextAccessor;
    private readonly IPublishedValueFallback _publishedValueFallback;

    // constructor specifiying custom dependencies
    public ShopController(
        IUmbracoContextAccessor umbracoContextAccessor,
        IPublishedValueFallback publishedValueFallback,
        ILogger<ShopController> logger,
        ICompositeViewEngine compositeViewEngine)
        : base(logger, compositeViewEngine)
    {
        _umbracoContextAccessor = umbracoContextAccessor;
        _publishedValueFallback = publishedValueFallback;
    }

    // find the content to associate with the route (IVirtualPageController)
    public IPublishedContent FindContent(ActionExecutingContext actionExecutedContext)
    {
        var productRoot = _umbracoContextAccessor.UmbracoContext.Content
                            .GetById(1234); // The products node

        // This cast should always be true
        if (actionExecutedContext.ActionDescriptor is ControllerActionDescriptor ctrlActionDesc)
        {
            // check which action is executing
            switch (ctrlActionDesc.ActionName)
            {
                case nameof(Index):
                    return productRoot;

                case nameof(Product):
                    // Get the sku/id from the route values
                    if (actionExecutedContext.ActionArguments.TryGetValue("id", out var sku))
                    {
                        // Get the associated Umbraco node for the sku
                        return productRoot
                            .Children
                            .FirstOrDefault(c => c.Value<string>(_publishedValueFallback, "sku") == sku.ToString());
                    }
                    else
                    {
                        return productRoot;
                    }
            }
        }

        return productRoot;
    }

    // default route
    public IActionResult Index()
    {
        // CurrentPage (IPublishedContent) will be the content returned
        // from the FindContent method.

        // return the view with the IPublishedContent
        return View(CurrentPage);
    }

    public IActionResult Product(string id)
    {
        // CurrentPage (IPublishedContent) will be the content returned
        // from the FindContent method.

        // One example of using a custom route would be to include additional
        // model information based on external services. For example, if
        // we wanted to return the stores the product is available in from
        // a custom data store.

        // EXAMPLE: Lookup up content item in custom data store with sku
        var dbProduct = DbContext.Products.GetBySku(sku);

        var shopModel = new Product(CurrentPage)
        {
            Sku = id,
            AvailableStores = dbProduct.AvailableStores
        };

        // return the view with the custom model
        return View(shopModel);
    }
}
```

The custom model could look like:

```cs
// If the custom model implements IContentModel then it can be auto model
// bound to any view that requires a model of this type or IPublishedContent.
// This could also inherit from ContentModel<T> where T is the models builder model
public class Product : ContentModel
{
    public Product(IPublishedContent content)
        : base(content)
    {
    }

    public string Sku { get; set; }

    public IEnumerable<string> AvailableStores { get; set; }
}
```

The views returned can be any Umbraco view that `@inherits Umbraco.Web.Common.Views.UmbracoViewPage` or can be strongly typed to a models builder model. With the above, the Views would need to be contained within the associated Controller named folder, else you can specify an absolute path to a view.

Now go to these routes:

* `/shop` = this will execute the Index action
* `/shop/product` = this will execute the Product action but without a sku
* `/shop/product/asdf` = this will execute the Product action with a sku of `asdf`

### Using the `IVirtualPageController` interface with attribute routing

The same as above except, the custom route will just look like this to enable attribute routing:

```cs
app.UseEndpoints(endpoints =>
{
    // this maps attribute routes
    endpoints.MapControllers();
});
```

Then the controller's actions can be attributed like:

```cs
[Route("[controller]/[action]")]
[Route("[controller]")]
[HttpGet]
public IActionResult Index()
```

and 

```cs
[Route("[controller]/[action]/{id?}")]
[HttpGet]
public IActionResult Product(string id)
```

And all routes will be the same and everything will work the same.

### Using conventional routing with ForUmbracoPage

For this, use the same controller except:

* remove all attribute routing
* do not implement IVirtualPageController and remove the FindContent method

Then the routing will look like:

```cs
app.UseEndpoints(endpoints =>
{
    endpoints.MapControllerRoute(
        "Test",
        "/shop/{action}/{id?}",
        new { Controller = "Shop", Action = "Index" })
    .ForUmbracoPage(FindContent);
});
```

And `FindContent` is another method passed into `ForUmbracoPage` which is almost the same as the previous controller's FindContent except we need to either resolve services from the HttpContext directly or inject them into your `IStartupFilter` class. This example just resolves from HttpContext:

```cs
// find the content to associate with the route (IVirtualPageController)
private IPublishedContent FindContent(ActionExecutingContext actionExecutingContext)
{
    // resolve services from the container
    var umbracoContextAccessor = actionExecutingContext.HttpContext.RequestServices.GetRequiredService<IUmbracoContextAccessor>();
    var publishedValueFallback = actionExecutingContext.HttpContext.RequestServices.GetRequiredService<IPublishedValueFallback>();


    var productRoot = umbracoContextAccessor.UmbracoContext.Content
                        .GetById(1234); // The products node

    // This cast should always be true
    if (actionExecutingContext.ActionDescriptor is ControllerActionDescriptor ctrlActionDesc)
    {
        switch (ctrlActionDesc.ActionName)
        {
            case nameof(Index):
                return productRoot;

            case nameof(Product):
                // Get the sku/id from the route values
                if (actionExecutingContext.ActionArguments.TryGetValue("id", out var sku))
                {
                    // Get the associated Umbraco node for the sku
                    return productRoot
                        .Children
                        .FirstOrDefault(c => c.Value<string>(publishedValueFallback, "sku") == sku.ToString());
                }
                else
                {
                    return productRoot;
                }
        }
    }

    return productRoot;
}
```

## Notable changes

Apart from the above explanations:

* Moves the UmbracoViewPage to a better namespace: Umbraco.Web.Common.Views.UmbracoViewPage
* New abstract UmbracoPageController to be used for front-end controllers
* New UmbracoStartupFilter (IStartupFilter) which calls the the `UseUmbraco` basic requirements. This is required because - Currently the only way for a plugin/package to access IApplicationBuilder is to use IStartupFilter and we don't want to re-invent something. All IStartupFilters occur **BEFORE** the Startup.cs `Configure` method which means that none of the basic requirements will be enabled like endpoint routing. So we now call UseUmbraco as an IStartupFilter which means that it will be called before custom IStartupFilters and the basic services will be enabled. IStartupFilters are ordered by when they are registered in DI and in theory all plugin/packages services will be registered after Umbraco's. This also means less stuff required for an end user to "Use" in their Startup.cs. In theory we could also automatically put `UseUmbracoBackOffice` and `UseUmbracoWebsite` as IStartupFilters too based on what is registered in DI but maybe this is less flexible.
* Simplifies UmbracoRouteValues - it doesn't need to contain the route hijacked boolean
*


src/Umbraco.Web.Common/Controllers/UmbracoPageController.cs
```diff
@@ -0,0 +1,104 @@
...
+    public abstract class UmbracoPageController : UmbracoController
+    {
+        private UmbracoRouteValues _umbracoRouteValues;
+        private readonly ICompositeViewEngine _compositeViewEngine;
+        private readonly ILogger<UmbracoPageController> _logger;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UmbracoPageController"/> class.
+        /// </summary>
+        protected UmbracoPageController(ILoggerFactory loggerFactory, ICompositeViewEngine compositeViewEngine)
```
bergmania, line 23: Not sure why we should use an `ILoggerFactory` and not just `ILogger<UmbracoPageController>`?

Remember the definition is `ILogger<out T>`, so `ILogger<RenderController>` can be injected into the `ILogger<UmbracoPageController>` as long as `RenderController` is a `UmbracoPageController`

src/Umbraco.Web.UI.NetCore/Views/Partials/grid/editors/embed.cshtml
```diff
@@ -1,5 +1,5 @@
-﻿@using Umbraco.Core
-@inherits Umbraco.Web.Common.AspNetCore.UmbracoViewPage<dynamic>
```
bergmania, line 2: Seems strange..

src/Umbraco.Web.UI.NetCore/Views/Partials/grid/editors/macro.cshtml
```diff
@@ -1,4 +1,4 @@
-﻿@inherits Umbraco.Web.Common.AspNetCore.UmbracoViewPage<dynamic>
```
bergmania, line 1: Seems strange

bergmania (approved):
Tasted and works as expected.. In review I only found a few small things, I can fix up my self

bergmania:
Note: Due to my changes, the code examples also need to change, so injection `ILogger<T>` instead of `ILoggerFactory`

bergmania:
I updated your description with updated code examples

bergmania:
@Shazwazza, I merged this, but please take a look at my commit https://github.com/umbraco/Umbraco-CMS/pull/9821/commits/2ba3eb436cb3341b08d5f4f160deec6e237c1dbf, to ensure you agree

PR #9823, opened 2021-02-16
Reduce memory use and speed up queries by only fetching required properties from examine

nzdev:
Prerequisites
 I have added steps to test this contribution in the description below

Description
What?
Optimize examine queries to fetch the least number of fields required to keep the same result.

Why?
Improve query performance, reduce memory use.

How?
Implement SelectFields for examine 1.1.0

---
_This item has been added to our backlog [AB#12567](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/12567)_


Shazwazza:
Nice one. I had to fix up a few things since back office search was no longer working, see rev: 76d206a4badd7b267d861a36043be62cb7ab25cf . All good now 👍

PR #9827, opened 2021-02-16
Netcore: Fixed buildscripts to work with latest Umbraco.Build and Directory.Build.props

bergmania:
### Details:
Updated the build script to work with the new `Directory.Build.props`

Reads the newest version of `Umbraco.Build` from `https://www.myget.org/F/umbracoprereleases/api/v3/index.json` even that it should be backward compatible.

Basically this file is read from and updated when `Umbraco.Build` `GetUmbracoVersion` and `SetUmbracoVersion` is used. Furthermore, we update the portnumber in `launchSettings.json` instead of the old csproj of Umbraco.Web.UI.

Removed the legacy `SolutionInfo.cs`

### Test:
- In solution root call `(build/build.ps1 -get).SetUmbracoVersion("10.0.0-beta")`
  - Ensure the data in `Directory.Build.props` is updated to the new version (beta only appears in InformationalVersion).
  - Ensure the port number is updated from 9000 to 10000 in `launchSettings.json`


nikolajlauridsen (approved):
Tested it out and can confirm that the version numbers get updated to 10.0.0 and 10.0.0-beta, and that port in launchSettings.json gets updated to 10000.

Looks good 👍

PR #9830, opened 2021-02-17
Fixes logging issue with Published Router and adds logging for Collision Detection

Shazwazza:
This fixes the logging done with PublishedRouter where it was always reporting "no document was found" for every request. This was because it was resolving the HasPublishedContent flag before the finders were even executed. Now it logs correctly and also adds a much nicer verbose debug log to help with debugging.

Have also added a log output for when `DetectCollision` fails - which is used in a couple places but namely this is used when showing the URL in the back office for a content item. It will run the inbound routing for that item to see if it can be routed to. If this fails we will get the "This document is published but its url cannot be routed" error. This now adds a warning to the log for when that occurs when "IgnorePublishedContentCollisions" is false (this is always false unless a developer has chosen to set this to true in their own finder = rare and in that case it's just a Debug log).

Hopefully this will allow us to shed some more light on this ongoing issue: https://github.com/umbraco/Umbraco-CMS/issues/9514

### DIFF
--- a/src/Umbraco.Web/Routing/PublishedRouter.cs
+++ b/src/Umbraco.Web/Routing/PublishedRouter.cs
@@ -417,8 +417,8 @@ namespace Umbraco.Web.Routing
             // some finders may implement caching
 
             using (_profilingLogger.DebugDuration<PublishedRouter>(
-                $"{tracePrefix}Begin finders",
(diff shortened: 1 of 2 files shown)


bergmania (approved):
Code changes looks good

PR #9831, opened 2021-02-17
Fixes #9824 - Filtering in block picker doesn't work

### LINKED ISSUE #9824: Filtering in block picker doesn't work
When using the filter/search input in block picker it doesn't filter any of the blocks.

## Umbraco version

Umbraco v8.11.1

Reproduction
------------

### Bug summary

When using the filter in the block picker I would expect it the filter the available blocks, but it doesn't seem to be the case.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/blockpicker/blockpicker.html#L22

@madsrasmussen recently added this change to item picker in this PR https://github.com/umbraco/Umbraco-CMS/pull/9641, and it might be a similar fix the need here.

### Steps to reproduce

Open block picker in a property using Block List editor and try filtering the blocks on their names.

### Expected result

The block picker should only show filtered results.

### Actual result

The block picker still shows all available blocks.

image

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9824

### Description
Currently the block picker is showing filter, when enough blocks are available. However the filtering didn't work.
Furthermore when opening the block picker it was scrolling down to last block, because the auto focus set here was stealing focus from the filter. Event worse when start typing in the search input you suddently lost focus, probably because the filtered results now changes the indexes and the `umb-auto-focus` watched these block cards. For now I have removing the, but we can always add it again, when we fix this. For now it works better without the auto focus on the block cards. https://github.com/umbraco/Umbraco-CMS/pull/9831/commits/f6a316802c0b777c99898c63aab7c0bfbeed39a7

I have furthermore cleaner up using `<umb-editor-container>` instead of all the hardcoded classes and also replaced the search filter with `<umb-search-filter>` component.

Finally I have fixed the filtering, so it now works correctly.

ib4ESEeHm5

However I noticed it didn't seems to filter the image block correctly. I named the alias of this `blockImage` since a document type with alias `image` was already in use.

image

Strange enough it does filter the blocks in searching `block` or a rename the alias `picture` and search on `picture`.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/blockpicker/blockpicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/blockpicker/blockpicker.controller.js
@@ -1,11 +1,14 @@
 angular.module("umbraco")
 .controller("Umbraco.Editors.BlockPickerController",
     function ($scope, localizationService) {
-        var vm = this;
 
+        var vm = this;
 
         vm.navigation = [];
 
+        vm.filter = {
+            searchTerm: ''
+        };
 
         localizationService.localizeMany(["blockEditor_tabCreateEmpty", "blockEditor_tabClipboard"]).then(
             function (data) {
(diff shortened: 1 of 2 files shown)


bjarnef:
Actually for the last part I think it is because the blocks contains a property `backgroundImage` which contain the word `image`. We should probably specify specific properties to search in.

We might need to add a custom filter to better handle filtering on blocks:
https://stackoverflow.com/a/31510239

nul800sebastiaan:
Thanks @bjarnef - so far, so good! I'll put the last comment on the list to discuss at HQ, it's an interesting point!

PR #9833, opened 2021-02-17
 Netcore: Remove UmbracoBackOfficeIdentity

nikolajlauridsen:
The PR removes `UmbracoBackOfficeIdentity`, relying instead on `ClaimsIdentity`.

The primary approach is used here is making extension methods for what used to be in `UmbracoBackOfficeIdentity`, and replaces usages of `UmbracoBackOfficeIdentity` with `ClaimsIdentity`

### Details

In ClaimsPrincipalExtensions I retained GetUmbracoIdentity, but now it returns a ClaimsIdentity that is authenticated, and that has been verified to include all the required claims. I stuck with the UmbracoIdentity name since you could still say that a ClaimsIdentiy with the required Claims is considered an UmbracoBackOfficeIdentity.

---

Where we before instatiated an UmbracoBackOfficeIdentity we now instantiate a ClaimsIdentity and add all claims with the extension method AddRequiredClaims

---

UmbracoBackOfficeIdentity used to have a property for the default Issuer, this has also been moved to the extension class, this does lead to a `using ClaimsIdentityExtensions = Umbraco.Extensions.ClaimsIdentityExtensions;` which is less than nice, suggestions for what we can do instead would be appreciated.

---

I did have some issues with the RequiredBackOfficeClaimTypes enumerable (previously in UmbracoBackOfficeIdentity, now in ClaimsIdentityExtensions), because `StartContentNodeIdClaimType` and `StartMediaNodeIdClaimType` are listed as required, however these can be null if nothing is selected, for now I've just commented them out, and everything seems to work fine, but suggestions on what best to do here would also be appreciated.

---

I retained the UmbracoBackOfficeIdentity tests, but adjusted them to now be testing the extension methods instead

### Testing

Everything should be functionally the same, but some things you can test are
* Ensure that a site can still isntall proberly
* Try creating a new user, with different claims, start nodes in particular and ensure that they're enforced.


src/Umbraco.Core/Security/UmbracoBackOfficeIdentity.cs
```diff
@@ -1,233 +0,0 @@
...
-
-        public string SecurityStamp => this.FindFirstValue(Constants.Security.SecurityStampClaimType);
-
-        public string[] Roles => FindAll(x => x.Type == DefaultRoleClaimType).Select(role => role.Value).ToArray();
-
-        /// <summary>
-        /// Overridden to remove any temporary claims that shouldn't be copied
-        /// </summary>
-        /// <returns></returns>
-        public override ClaimsIdentity Clone()
```
bergmania, line 223: Seems wrong we are not handling this anymore.

@Shazwazza, do you have any suggestions regarding the `Clone` method. It is called by many of Microsoft's internal methods. IMO it seems dangerous to save `Constants.Security.TicketExpiresClaimType` as a claim now that we can't override the clone method?

Currently, if I re-introduce the clone test, it shows this is not handled anymore.

src/Umbraco.Infrastructure/Security/BackOfficeClaimsPrincipalFactory.cs
```diff
@@ -46,8 +45,7 @@ public override async Task<ClaimsPrincipal> CreateAsync(BackOfficeIdentityUser u
 
             // TODO: We want to remove UmbracoBackOfficeIdentity and only rely on ClaimsIdentity, once
             // that is done then we'll create a ClaimsIdentity with all of the requirements here instead
```
bergmania, line 47: This TODO can be removed now

src/Umbraco.Core/Extensions/ClaimsIdentityExtensions.cs
```diff
@@ -72,5 +75,203 @@ public static string FindFirstValue(this ClaimsIdentity identity, string claimTy
...
+            ClaimTypes.NameIdentifier, //id
+            ClaimTypes.Name,  //username
+            ClaimTypes.GivenName,
+            // Constants.Security.StartContentNodeIdClaimType, These seem to be able to be null...
+            // Constants.Security.StartMediaNodeIdClaimType,
+            ClaimTypes.Locality,
+            Constants.Security.SecurityStampClaimType
+        };
+
+        public const string Issuer = Constants.Security.BackOfficeAuthenticationType;
```
bergmania, line 96: I don't like the fact that we have a public constant just proxying another public constant.

I would use `Constants.Security.BackOfficeAuthenticationType` anywhere `ClaimsIdentityExtensions.Issuer` is used.

bergmania (changes requested):
I'm afraid we need to handle clone in some way or not saving temp claims.

bergmania:
> UmbracoBackOfficeIdentity used to have a property for the default Issuer, this has also been moved to the extension class, this does lead to a using ClaimsIdentityExtensions = Umbraco.Extensions.ClaimsIdentityExtensions; which is less than nice, suggestions for what we can do instead would be appreciated.

This don't seem to be required anymore. Maybe the namespace cleanup has fixed this issue?

src/Umbraco.Core/Extensions/ClaimsIdentityExtensions.cs
nikolajlauridsen: Ahh that's a good point, I've replaced usages of ClaimsIdentityExtensions.Issuer with Constants.Security.BackOfficeAuthenticationType

nikolajlauridsen:
> > UmbracoBackOfficeIdentity used to have a property for the default Issuer, this has also been moved to the extension class, this does lead to a using ClaimsIdentityExtensions = Umbraco.Extensions.ClaimsIdentityExtensions; which is less than nice, suggestions for what we can do instead would be appreciated.
> 
> This don't seem to be required anymore. Maybe the namespace cleanup has fixed this issue?

Ah yes, after the namespace changes I got multiple errors, so I moved my extension methods into the existing ones, since they both extended IIdentity and were called "ClaimsIdentityExtensions"

nikolajlauridsen:
I also just recalled that we use `Thread.CurrentPrincipal?.GetUmbracoIdentity();` in AuditEventsComponent, which we no longer can in netcore, any suggestions as to what we can do instead?

bergmania:
> I also just recalled that we use `Thread.CurrentPrincipal?.GetUmbracoIdentity();` in AuditEventsComponent, which we no longer can in netcore, any suggestions as to what we can do instead?

We only use it to get the current user. Use this instead
`_backOfficeSecurityAccessor.BackOfficeSecurity.CurrentUser` and inject `_backOfficeSecurityAccessor`

PR #9834, opened 2021-02-17
Fixes error trying to load Cosmos.CRTCompat.dll (System.BadImageForma…

nul800sebastiaan:
### Prerequisites

When trying to work with CosmosDB I added `Microsoft.Azure.Cosmos` which includes a dll `Cosmos.CRTCompat.dll`. When trying to read the assembly name from the file I get an error: 

```
Unhandled exception. System.BadImageFormatException: Could not load file or assembly 'D:\Temp\Test\tester\MyTestProject\bin\Debug\net5.0\Cosmos.CRTCompat.dll'. The module was expected to contain an assembly manifest.
File name: 'D:\Temp\Test\tester\MyTestProject\bin\Debug\net5.0\Cosmos.CRTCompat.dll'
   at System.Reflection.AssemblyName.nGetFileInformation(String s)
   at System.Reflection.AssemblyName.GetAssemblyName(String assemblyFile)
   at Umbraco.Core.Composing.ReferenceResolver.ResolveAssemblies()
   at Umbraco.Core.Composing.FindAssembliesWithReferencesTo.Find()
   at Umbraco.Core.Composing.DefaultUmbracoAssemblyProvider.get_Assemblies()
   at Umbraco.Core.Composing.TypeFinder.GetAllAssemblies()
   at Umbraco.Core.Composing.TypeFinder.GetFilteredAssemblies(IEnumerable`1 excludeFromResults, String[] exclusionFilter)
   at Umbraco.Core.Composing.TypeFinder.get_AssembliesToScan()
   at Umbraco.Core.Composing.TypeLoader.get_AssembliesToScan()
   at Umbraco.Core.Composing.TypeLoader.<>c__DisplayClass50_0`1.<GetTypes>b__0()
   at Umbraco.Core.Composing.TypeLoader.GetTypesInternalLocked(Type baseType, Type attributeType, Func`1 finder, String action, Boolean cache)
   at Umbraco.Core.Composing.TypeLoader.GetTypesInternal(Type baseType, Type attributeType, Func`1 finder, String action, Boolean cache)
   at Umbraco.Core.Composing.TypeLoader.GetTypes[T](Boolean cache, IEnumerable`1 specificAssemblies)
   at Umbraco.Core.DependencyInjection.UmbracoBuilderExtensions.AddAllCoreCollectionBuilders(IUmbracoBuilder builder)
   at Umbraco.Core.DependencyInjection.UmbracoBuilder.AddCoreServices()
   at Umbraco.Core.DependencyInjection.UmbracoBuilder..ctor(IServiceCollection services, IConfiguration config, TypeLoader typeLoader, ILoggerFactory loggerFactory)
   at Umbraco.Web.Common.DependencyInjection.UmbracoBuilderExtensions.AddUmbraco(IServiceCollection services, IWebHostEnvironment webHostEnvironment, IConfiguration config)
   at MyTestProject.Startup.ConfigureServices(IServiceCollection services) in D:\Temp\Test\tester\MyTestProject\Startup.cs:line 52
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.InvokeCore(Object instance, IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.<>c__DisplayClass9_0.<Invoke>g__Startup|0(IServiceCollection serviceCollection)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.Invoke(Object instance, IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.<>c__DisplayClass8_0.<Build>b__0(IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.GenericWebHostBuilder.UseStartup(Type startupType, HostBuilderContext context, IServiceCollection services, Object instance)
   at Microsoft.AspNetCore.Hosting.GenericWebHostBuilder.<>c__DisplayClass13_0.<UseStartup>b__0(HostBuilderContext context, IServiceCollection services)
   at Microsoft.Extensions.Hosting.HostBuilder.CreateServiceProvider()
   at Microsoft.Extensions.Hosting.HostBuilder.Build()
   at MyTestProject.Program.Main(String[] args) in D:\Temp\Test\tester\MyTestProject\Program.cs:line 11
```

For some reason this dll seems to have a weird format. seems okay to ignore this.

I tried adding logging but I had to update too many class instantiations so I am guessing this resolver might run at a point where the DI is not ready yet and I can't easily instantiate a logger? Happy to fix if someone has a pointer. I would suggest logging with the debug level so it doesn't constantly show up in logs, only if you change the level from Info to Debug.

### DIFF
--- a/src/Umbraco.Tests/Testing/UmbracoTestBase.cs
+++ b/src/Umbraco.Tests/Testing/UmbracoTestBase.cs
@@ -191,7 +191,7 @@ namespace Umbraco.Tests.Testing
             var proflogger = new ProfilingLogger(loggerFactory.CreateLogger<ProfilingLogger>(), profiler);
             IOHelper = TestHelper.IOHelper;
 
-            TypeFinder = new TypeFinder(loggerFactory.CreateLogger<TypeFinder>(), new DefaultUmbracoAssemblyProvider(GetType().Assembly), new VaryingRuntimeHash());
+            TypeFinder = new TypeFinder(loggerFactory.CreateLogger<TypeFinder>(), new DefaultUmbracoAssemblyProvider(GetType().Assembly, NullLoggerFactory.Instance), new VaryingRuntimeHash());
             var appCaches = GetAppCaches();
             var globalSettings = new GlobalSettings();
             var settings = new WebRoutingSettings();
--- a/src/Umbraco.Tests/Testing/UmbracoTestBase.cs
+++ b/src/Umbraco.Tests/Testing/UmbracoTestBase.cs
@@ -191,7 +191,7 @@ namespace Umbraco.Tests.Testing
             var proflogger = new ProfilingLogger(loggerFactory.CreateLogger<ProfilingLogger>(), profiler);
             IOHelper = TestHelper.IOHelper;
 
-            TypeFinder = new TypeFinder(loggerFactory.CreateLogger<TypeFinder>(), new DefaultUmbracoAssemblyProvider(GetType().Assembly, NullLoggerFactory.Instance), new VaryingRuntimeHash());
+            TypeFinder = new TypeFinder(loggerFactory.CreateLogger<TypeFinder>(), new DefaultUmbracoAssemblyProvider(GetType().Assembly, loggerFactory), new VaryingRuntimeHash());
             var appCaches = GetAppCaches();
             var globalSettings = new GlobalSettings();
             var settings = new WebRoutingSettings();
--- a/src/Umbraco.Core/Composing/DefaultUmbracoAssemblyProvider.cs
+++ b/src/Umbraco.Core/Composing/DefaultUmbracoAssemblyProvider.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Reflection;
+using Microsoft.Extensions.Logging;
 
 namespace Umbraco.Core.Composing
 {
@@ -14,6 +15,7 @@ namespace Umbraco.Core.Composing
     public class DefaultUmbracoAssemblyProvider : IAssemblyProvider
     {
         private readonly Assembly _entryPointAssembly;
+        private readonly ILoggerFactory _loggerFactory;
         private static readonly string[] UmbracoCoreAssemblyNames = new[]
             {
                 "Umbraco.Core",
@@ -26,9 +28,10 @@ namespace Umbraco.Core.Composing
                 "Umbraco.Web.Website",
             };
 
-        public DefaultUmbracoAssemblyProvider(Assembly entryPointAssembly)
+        public DefaultUmbracoAssemblyProvider(Assembly entryPointAssembly, ILoggerFactory loggerFactory)
         {
             _entryPointAssembly = entryPointAssembly ?? throw new ArgumentNullException(nameof(entryPointAssembly));
+            _loggerFactory = loggerFactory;
         }
 
         // TODO: It would be worth investigating a netcore3 version of this which would use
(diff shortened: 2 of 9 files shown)


bergmania:
I guess it is a good idea to inject an `ILoggerFactory`. But yes, this code is run pretty early, so the container is not available.

Should still be fairly easy to inject, and for tests, you should be able to use NullLoggerFactory.Instance, or a mock, if you do not already have a variable ready.

I guess if we are doing this, we should  maybe catch more exception types or just Exception 🙈 , as long as we are logging.

The documentation lists the following exceptions:

```cs
    /// <summary>Gets the <see cref="T:System.Reflection.AssemblyName"></see> for a given file.</summary>
    /// <param name="assemblyFile">The path for the assembly whose <see cref="T:System.Reflection.AssemblyName"></see> is to be returned.</param>
    /// <returns>An object that represents the given assembly file.</returns>
    /// <exception cref="T:System.ArgumentNullException"><paramref name="assemblyFile">assemblyFile</paramref> is null.</exception>
    /// <exception cref="T:System.ArgumentException"><paramref name="assemblyFile">assemblyFile</paramref> is invalid, such as an assembly with an invalid culture.</exception>
    /// <exception cref="T:System.IO.FileNotFoundException"><paramref name="assemblyFile">assemblyFile</paramref> is not found.</exception>
    /// <exception cref="T:System.Security.SecurityException">The caller does not have path discovery permission.</exception>
    /// <exception cref="T:System.BadImageFormatException"><paramref name="assemblyFile">assemblyFile</paramref> is not a valid assembly.</exception>
    /// <exception cref="T:System.IO.FileLoadException">An assembly or module was loaded twice with two different sets of evidence.</exception>
    public static AssemblyName GetAssemblyName(string assemblyFile);
 ```

**Comment by nul800sebastiaan** [comment]
Thanks for the feedback! I hope I did this right! 🙈 

I've added some exceptions with different levels, might be a bit overkill but in my mind:

- If the dll can't be loaded due to `BadImageFormatException` we should skip it, as we know now this can be "expected" - don't log unless logging it set to Debug level explicitly
- If its a `SecurityException` then something went wrong setting up security for that file, log as error 💥 
- The other exceptions really shouldn't happen, but if they do it would be good to see them on an informational level

Make sense? Happy to amend if anything is not quite right yet!

File: src/Umbraco.Tests/Testing/UmbracoTestBase.cs [hunk 1]
```diff
@@ -191,7 +191,7 @@ public virtual void SetUp()
             var proflogger = new ProfilingLogger(loggerFactory.CreateLogger<ProfilingLogger>(), profiler);
             IOHelper = TestHelper.IOHelper;
 
-            TypeFinder = new TypeFinder(loggerFactory.CreateLogger<TypeFinder>(), new DefaultUmbracoAssemblyProvider(GetType().Assembly), new VaryingRuntimeHash());
+            TypeFinder = new TypeFinder(loggerFactory.CreateLogger<TypeFinder>(), new DefaultUmbracoAssemblyProvider(GetType().Assembly, NullLoggerFactory.Instance), new VaryingRuntimeHash());
```
bergmania, line 194:
```suggestion
            TypeFinder = new TypeFinder(loggerFactory.CreateLogger<TypeFinder>(), new DefaultUmbracoAssemblyProvider(GetType().Assembly, loggerFactory), new VaryingRuntimeHash());
```

nul800sebastiaan: Oh duh! Good catch!

bergmania (approved):

PR #9837, opened 2021-02-17
Should allow grid editors to use hash based settings

### LINKED ISSUE #9836: Broken grid editor settings [regression: 8.10+]
#8827 seems to have removed the support for schemas used to display custom settings for grid editors.

## Umbraco version

8.10.x

Reproduction
------------

Pre 8.10 this modification of the built-in heading grid-editor would work:

```json
    {
        "name": "Headline",
        "nameTemplate": "{{ value }}",
        "alias": "headline",
        "view": "textstring",
        "icon": "icon-coin",
        "config": {
            "style": "font-size: 36px; line-height: 45px; font-weight: bold",
            "markup": "<h1>#value#</h1>",
            "settings": {
                "color": {
                    "key": "color",
                    "label": "Color",
                    "view": "textstring"
                }
            }
        }
    },
```

After 8.10, the settings property _has_ to be an array, otherwise nothing shows. So it will have to be as such:

```json
    {
        "name": "Headline",
        "nameTemplate": "{{ value }}",
        "alias": "headline",
        "view": "textstring",
        "icon": "icon-coin",
        "config": {
            "style": "font-size: 36px; line-height: 45px; font-weight: bold",
            "markup": "<h1>#value#</h1>",
            "settings": [
                {
                    "key": "color",
                    "label": "Color",
                    "view": "textstring"
                }
            ]
        }
    },
```

### Bug summary

It breaks my contentlist package as well as a s**tload of other editors we've built through the years.  
I appreciate it's not documented, but still - this is a major backwards compatibility problem imo.

---
_This item has been added to our backlog [AB#10590](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10590)_

lars-erik:
Should fix #9836

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
@@ -16,7 +16,23 @@ function ConfigController($scope) {
             $scope.model.close();
         }
     }
-    
+
+    $scope.showEmptyState = function() {
+        return $scope.model.config.length === 0
+            && Object.keys($scope.model.config).length === 0
+            && $scope.model.styles.length === 0;
+    }
+
+    $scope.showConfig = function() {
+        return $scope.model.config &&
+            ($scope.model.config.length > 0 || Object.keys($scope.model.config).length > 0);
+    }
+
+    $scope.showStyles = function() {
+        return $scope.model.styles &&
+            ($scope.model.styles.length > 0 || Object.keys($scope.model.styles).length > 0);
+    }
+
 }
 
 angular.module("umbraco").controller("Umbraco.PropertyEditors.GridPrevalueEditor.ConfigController", ConfigController);
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
@@ -21,7 +21,10 @@ function ConfigController($scope) {
     vm.showConfig = false;
     vm.showStyles = false;
 
-    $scope.$watchGroup(['model.config.length', 'model.styles.length'], function(newValues, oldValues, scope) {
+    $scope.$watchCollection('model.config', onWatch);
+    $scope.$watchCollection('model.styles', onWatch);
+
+    function onWatch() {
 
         vm.showConfig = $scope.model.config &&
             ($scope.model.config.length > 0 || Object.keys($scope.model.config).length > 0);
@@ -30,7 +33,7 @@ function ConfigController($scope) {
 
         vm.showEmptyState = vm.showConfig === false && vm.showStyles === false;
 
-    });
+    }
 
 }
 
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
@@ -17,21 +17,20 @@ function ConfigController($scope) {
         }
     }
 
-    vm.showEmptyState = function() {
-        return $scope.model.config.length === 0
-            && Object.keys($scope.model.config).length === 0
-            && $scope.model.styles.length === 0;
-    }
+    vm.showEmptyState = false;
+    vm.showConfig = false;
+    vm.showStyles = false;
 
-    vm.showConfig = function() {
-        return $scope.model.config &&
+    $scope.$watchGroup(['model.config.length', 'model.styles.length'], function(newValues, oldValues, scope) {
+
+        vm.showConfig = $scope.model.config &&
             ($scope.model.config.length > 0 || Object.keys($scope.model.config).length > 0);
-    }
-
-    vm.showStyles = function() {
-        return $scope.model.styles &&
+        vm.showStyles = $scope.model.styles &&
             ($scope.model.styles.length > 0 || Object.keys($scope.model.styles).length > 0);
-    }
+
+        vm.showEmptyState = vm.showConfig === false && vm.showStyles === false;
+
+    });
 
 }
 
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
@@ -17,18 +17,18 @@ function ConfigController($scope) {
         }
     }
 
-    $scope.showEmptyState = function() {
+    vm.showEmptyState = function() {
         return $scope.model.config.length === 0
             && Object.keys($scope.model.config).length === 0
             && $scope.model.styles.length === 0;
     }
 
-    $scope.showConfig = function() {
+    vm.showConfig = function() {
         return $scope.model.config &&
             ($scope.model.config.length > 0 || Object.keys($scope.model.config).length > 0);
     }
 
-    $scope.showStyles = function() {
+    vm.showStyles = function() {
         return $scope.model.styles &&
             ($scope.model.styles.length > 0 || Object.keys($scope.model.styles).length > 0);
     }
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.html
@@ -13,11 +13,11 @@
         <umb-editor-container>
             <ng-form name="gridItemConfigEditor" val-form-manager>
 
-                <umb-empty-state position="center" ng-if="showEmptyState()">
+                <umb-empty-state position="center" ng-if="vm.showEmptyState()">
                     <localize key="grid_noConfiguration">No further configuration available</localize>
                 </umb-empty-state>
 
-                <umb-box ng-if="showConfig()">
+                <umb-box ng-if="vm.showConfig()">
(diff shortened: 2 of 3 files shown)


src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
```diff
@@ -16,7 +16,23 @@ function ConfigController($scope) {
...
+    $scope.showConfig = function() {
+        return $scope.model.config &&
+            ($scope.model.config.length > 0 || Object.keys($scope.model.config).length > 0);
+    }
+
+    $scope.showStyles = function() {
+        return $scope.model.styles &&
+            ($scope.model.styles.length > 0 || Object.keys($scope.model.styles).length > 0);
+    }
+
```
bjarnef, line 35: @lars-erik I would probably  use the `vm` syntax here similar to close and submit.
Alternatively it should convert the settings and styles objects to array with a single object.
Not sure if that would cause any issues though.

lars-erik: I changed to vm. Also enabled the button.  
I would be super weary of changing the existing code and supported schema beyond this point.  
One broken reference and things may start behaving differently.  
I'd rather see a grid 2.0 instead of meddling too much.

Surely I'm not the only one relying on this schema.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
```diff
@@ -16,7 +16,23 @@ function ConfigController($scope) {
             $scope.model.close();
         }
     }
-    
+
+    vm.showEmptyState = function() {
+        return $scope.model.config.length === 0
+            && Object.keys($scope.model.config).length === 0
+            && $scope.model.styles.length === 0;
+    }
```
bjarnef, line 24: @lars-erik is it missing a `Object.keys($scope.model.styles).length === 0` here as well?

lars-erik: I never used styles for individual grid editor settings. Dunno if that would ever be a case.  
Could add it for good measure though?

bjarnef, line 24: @lars-erik I guess this could be defined as:

```
vm.showEmptyState = function() {
        return vm.showConfig() === false && vm.showStyles() === false;
}
```

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
```diff
@@ -16,7 +16,23 @@ function ConfigController($scope) {
...
+    vm.showConfig = function() {
+        return $scope.model.config &&
+            ($scope.model.config.length > 0 || Object.keys($scope.model.config).length > 0);
+    }
+
+    vm.showStyles = function() {
+        return $scope.model.styles &&
+            ($scope.model.styles.length > 0 || Object.keys($scope.model.styles).length > 0);
+    }
+
```
bjarnef, line 35: Maybe these should just be scope properties instead of functions:

```
vm.showConfig = $scope.model.config && ($scope.model.config.length > 0 || Object.keys($scope.model.config).length > 0);
vm.showStyles = $scope.model.styles && ($scope.model.styles.length > 0 || Object.keys($scope.model.styles).length > 0);
```

A bit more consistent with `shouldShowUrl` and `hasCrops` in mediacropdetails infinite overlay here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/overlays/mediacropdetails.html#L13

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
lars-erik: For sure! I'll see about it suddenly.

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
lars-erik: I need them to be dynamic for the contentlist package. It exploits the ability to mutate available settings.

nielslyngsoe:
Hi @lars-erik — I've taken the freedom to move on with this PR.
I've rewritten your proposal to be dynamic and still performant in AngularJS.

Hopefully making both of you happy ( @bjarnef )

If your okay with these changes then please make this PR ready for review, then we will have it handled very soon :-)

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
```diff
@@ -17,21 +17,20 @@ function ConfigController($scope) {
...
-    }
+    vm.showEmptyState = false;
+    vm.showConfig = false;
+    vm.showStyles = false;
 
-    vm.showConfig = function() {
-        return $scope.model.config &&
-            ($scope.model.config.length > 0 || Object.keys($scope.model.config).length > 0);
-    }
+    $scope.$watchGroup(['model.config.length', 'model.styles.length'], function(newValues, oldValues, scope) {
```
lars-erik, line 24: Won't this revert reacting to the cases where model.config is a hash? That's the whole point of the PR. 🤪

lars-erik: I just tested it with both configs and it seems to work. 🤯
In the "old" days, wasn't an object's length equal to it's count of keys?
If so, I don't really see why @bjarnef's changes ended up hiding everything, but oh well. :P

nielslyngsoe: @lars-erik I think you are right, this is not the right way of covering that case. So I have switched it to a watchCollection() instead. This can handle both Array or Object, and will watch for shallow changes, aka. the first level (Keys or Items)

nielslyngsoe: @lars-erik could I ask you to test this against your package — that would simplify the task of the reviewer for this. Thanks :-)

lars-erik:
I've tested this with my content list package and it works now. 🎉
Any chance for 8.12 [.1]?

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
lars-erik: I did test it with this, and I realized I mutate the settings before they arrive in this controller. After that I only mutate a nested setting thing, but that's not affected by this. However, switching to watchCollection would allow for even more dynamic settings which could be good. I do show/hide settings via a custom property editor. I don't think a deep watch of a settings object/collection will cost that much performance.

nielslyngsoe:
@lars-erik sorry I don't think so. If I got the release cycle right then, then the scope of a release is locked when the RC is released. so 8.13. :-/

lars-erik:
> @lars-erik sorry I don't think so. If I got the release cycle right then, then the scope of a release is locked when the RC is released. so 8.13. :-/

I get that @nielslyngsoe , but IMO this is a bugfix / fixing backwards compat, not a new feature. Could at least be worth a patch? (8.12.1)

lars-erik:
@nielslyngsoe Any possibility it can be available in a nightly somewhere after merge? We could likely launch with a nightly on a site I'm reluctant to downgrade for this. 😅

src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/dialogs/config.controller.js
```diff
@@ -21,7 +21,10 @@ function ConfigController($scope) {
     vm.showConfig = false;
     vm.showStyles = false;
 
-    $scope.$watchGroup(['model.config.length', 'model.styles.length'], function(newValues, oldValues, scope) {
+    $scope.$watchCollection('model.config', onWatch);
+    $scope.$watchCollection('model.styles', onWatch);
```
bjarnef, line 25: @nielslyngsoe can't this be simplified to:

```
$scope.$watchCollection(['model.config','model.styles'], onWatch);
```

madsrasmussen (approved):
I have tested before and after the fix and with the changes applied the config works in both examples. Code changes look good to me.

nul800sebastiaan:
This will be available in patch releases, if you want a nightly try here, they're still building now:

8.10.x - https://dev.azure.com/umbraco/Umbraco%20Cms/_build?definitionId=75&_a=summary&repositoryFilter=34&branchFilter=7712%2C7712 
8.11.x - https://dev.azure.com/umbraco/Umbraco%20Cms/_build?definitionId=75&_a=summary&repositoryFilter=34&branchFilter=7842%2C7842
8.12.x - https://dev.azure.com/umbraco/Umbraco%20Cms/_build?definitionId=75&_a=summary&repositoryFilter=34&branchFilter=8260%2C8260

lars-erik:
Awesome! Thanks all!

Will the patch releases end up here?  
image

nul800sebastiaan:
Patch releases go to NuGet, the nightlies don't show up anywhere except on the build output on Azure DevOps. There's nuget packages there though so you could upload them to your own myget instance if needed.

PR #9838, opened 2021-02-18
NetCore: Moving MB.Embedded into the other existing assemblies

elit0451:
ModelsBuilder.Embedded proj was needed for the front-end of Umbraco, but it had several other dependencies which meant that if people wanted to only run the front-end without the back-office, they will still be shipping with the entire Umbraco.Web.BackOffice proj.

This PR resolves this - its goal is to be able to ship the front-end without shipping the entire back-office project too.


bergmania (approved):
Only minor findings that I fixed already :)

PR #9839, opened 2021-02-18
Do not apply runtime optimization to core scripts and script libraries

kjac:
### Prerequisites

### Description

As-is we let Smidge perform runtime optimization and bundling of all client scripts - both the ones shipped with Umbraco and the ones for backoffice extensions. We even optimize some 3rd party libraries runtime, because we ship Umbraco with un-minified versions of these libraries.

This PR aims at minimizing the runtime overhead by never optimizing any of the scripts shipped with Umbraco. This is done by:

1. Ensuring we only ship minified versions of 3rd party libraries.
2. Minifying the core scripts as part of the build process.
3. Splitting the "main script bundle" into two separate bundles - a bundle for the core scripts and a bundle for the backoffice extensions scripts.
4. Only performing optimization/minification on the backoffice extensions bundle. The core bundles now only exist to handle cache busting of the core scripts.

#### Limitations

A few of the core script files aren't part of the gulp setup, namely the files in `\src\Umbraco.Web.UI.Client\lib\umbraco\`. They currently work as both source files and script libraries at the same time. This should really be changed, so they can be rolled into the same setup as the rest of the source files. For now they are excempted from the minification and will remain un-minified in this setup - even in production builds. Fortunately the files are very small, so they represent next to no overhead. 

- `\src\Umbraco.Web.UI.Client\lib\umbraco\Extensions.js`
- `\src\Umbraco.Web.UI.Client\lib\umbraco\LegacySpeechBubble.js`
- `\src\Umbraco.Web.UI.Client\lib\umbraco\NamespaceManager.js`

I'm not really sure that `LegacySpeechBubble.js` and `NamespaceManager.js` are even used by the core anymore, so perhaps they should be cleaned up for the Unicore release.

#### Test the default dev setup

First of all, test that the backoffice runs as usual:

1. Stop your dev site if it's running.
2. Delete everything under `\src\Umbraco.Web.UI.NetCore\wwwroot\umbraco` (it's all output from the client build).
3. Build the client by running `gulp dev`.
	- Notice how the generated script files in `\src\Umbraco.Web.UI.NetCore\wwwroot\umbraco\js` are actually un-minified despite their `.min.js` extension, since they have been built in dev mode.
4. Boot up the site and log into the backoffice with Chrome devtools open.
5. Verify that you can click around in the backoffice and that there are no nasty console errors.
6. Use the devtools network tab to verify that the client really did load the generated `.min.js` files.
debug-mode-js-files

#### Test the bundles

Test that the backoffice core and extensions scripts are split into separate bundles, and that the core bundle is not minified runtime:

1. Stop your dev site.
2. Disable debug mode in app settings (`Umbraco.CMS.Hosting.Debug`).
3. Increment the runtime version in app settings (`Umbraco.CMS.RuntimeMinification.version`).
runtime-version
4. Unzip this test dashboard: [MyDashboard.zip](https://github.com/umbraco/Umbraco-CMS/files/6001680/MyDashboard.zip) to `\src\Umbraco.Web.UI.NetCore\App_Plugins` _and_ `\src\Umbraco.Web.UI.NetCore\wwwroot\App_Plugins`.
	- Apparently there is a slight bug in plugin path detection for asset minification (it's not caused by this PR).
	dashboard-unzip-locations
5. Boot up the site and log into the backoffice with Chrome devtools open.
6. Find the script bundle outputs by searching for _"backoffice js"_ in the devtools network tab - you will find two bundles, one for the backoffice core and one for the backoffice extensions (i.e. the installed dashboard).
	- Notice you must look under _All_, because the generated scripts won't list under _JS_ due to the bundle file extension.
	chrome-devtools
7. Inspect each bundle output in devtools, and verify that:
	1. The backoffice core bundle is comprised by minified 3rd party libraries (e.g. jQuery, AngularJs) and un-minified Umbraco scripts (controllers, services, ...) - you'll need to scroll quite a bit down to find the Umbraco scripts.
	inspect-core-js
	2. The backoffice extensions bundle contains minified scripts.
	inspect-extensions-js

#### Test the minification

To verify that the minification of the backoffice bundle works:

1. Stop your dev site.
2. Set `minify: true` in `\src\Umbraco.Web.UI.Client\gulp\config.js` for the client dev build.
gulp-config-dev-minify
3. Build the client by running `gulp dev`.
	- Notice how the generated script files in `\src\Umbraco.Web.UI.NetCore\wwwroot\umbraco\js` are now minified.
4. Increment the runtime version in app settings (`Umbraco.CMS.RuntimeMinification.version`).
5. Boot up the site and log into the backoffice with Chrome devtools open.
6. Inspect the bundles in Chrome devtools and verify that the Umbraco scripts are now served minified, just like the 3rd party libraries (with the exception of files mentioned under Limitations in the PR description).

### Other optimizations (out of scope for this PR)

I think the CSS shipped with Umbraco could be handled in a similar approach. If this PR is accepted I'll have a look at that later on.

It would be beneficial to generate the core bundles as single file outputs in the client build, instead of having Smidge join them together at runtime. Note however that some of the individual files are referenced by other parts of Umbraco outside of their respective bundles. It would be required to create new bundles for these usages as well. On the upside that would mean we could ensure appropriate cache busting for these usages too.

### DIFF
--- a/src/Umbraco.Web.Common/RuntimeMinification/SmidgeRuntimeMinifier.cs
+++ b/src/Umbraco.Web.Common/RuntimeMinification/SmidgeRuntimeMinifier.cs
@@ -74,7 +74,7 @@ namespace Umbraco.Web.Common.RuntimeMinification
 
         public async Task<string> RenderCssHereAsync(string bundleName) => (await _smidge.SmidgeHelper.CssHereAsync(bundleName, _hostingEnvironment.IsDebugMode)).ToString();
 
-        public void CreateJsBundle(string bundleName, params string[] filePaths)
+        public void CreateJsBundle(string bundleName, bool optimizeOutput, params string[] filePaths)
         {
             if (filePaths.Any(f => !f.StartsWith("/") && !f.StartsWith("~/")))
                 throw new InvalidOperationException("All file paths must be absolute");
@@ -85,7 +85,8 @@ namespace Umbraco.Web.Common.RuntimeMinification
             // Here we could configure bundle options instead of using smidge's global defaults.
             // For example we can use our own custom cache buster for this bundle without having the global one
             // affect this or vice versa.
-            var bundle = _bundles.Create(bundleName, _jsPipeline.Value, WebFileType.Js, filePaths);
+            var pipeline = optimizeOutput ? _jsPipeline.Value : _bundles.PipelineFactory.Create();
+            var bundle = _bundles.Create(bundleName, pipeline, WebFileType.Js, filePaths);
         }
 
         public async Task<string> RenderJsHereAsync(string bundleName) => (await _smidge.SmidgeHelper.JsHereAsync(bundleName, _hostingEnvironment.IsDebugMode)).ToString();
--- a/src/Umbraco.Core/Configuration/Models/ContentSettings.cs
+++ b/src/Umbraco.Core/Configuration/Models/ContentSettings.cs
@@ -145,7 +145,7 @@ namespace Umbraco.Core.Configuration.Models
                     width:1em;
                 }}
             </style>
-            <script type=""text/javascript"" data-umbraco-path=""{0}"" src=""{0}/js/umbraco.websitepreview.js""></script>";
+            <script type=""text/javascript"" data-umbraco-path=""{0}"" src=""{0}/js/umbraco.websitepreview.min.js""></script>";
 
         /// <summary>
         /// Gets or sets a value for the content notification settings.
--- a/src/Umbraco.Core/WebAssets/IRuntimeMinifier.cs
+++ b/src/Umbraco.Core/WebAssets/IRuntimeMinifier.cs
@@ -40,6 +40,7 @@ namespace Umbraco.Core.WebAssets
         /// Creates a JS bundle
         /// </summary>
         /// <param name="bundleName"></param>
+        /// <param name="optimizeOutput"></param>
         /// <param name="filePaths"></param>
         /// <remarks>
         /// All files must be absolute paths, relative paths will throw <see cref="InvalidOperationException"/>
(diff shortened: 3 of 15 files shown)


Shazwazza (approved):
This is fantastic 🎉 Have left some questions and comments inline and below.

One issue is that we aren't copying out the source map files for the libraries. So we get warnings like this

image

I'm unsure if we can suppress those warnings? else, I'm unsure if we should ship the source maps? 

> I'm not really sure that LegacySpeechBubble.js and NamespaceManager.js are even used by the core anymore, so perhaps they should be cleaned up for the Unicore release.

Yes I think that should be another task, any cleanup for this old stuff should be done where possible

> Apparently there is a slight bug in plugin path detection for asset minification (it's not caused by this PR).

Is this why we need to unzip this in 2x locations? Do we have a task to get this fixed?

> I think the CSS shipped with Umbraco could be handled in a similar approach. If this PR is accepted I'll have a look at that later on.

sounds great!

kjac:
@Shazwazza cool! I'll have a look at those during this week 😄

kjac:
@Shazwazza comments and sourcemaps fixed in #9980 

You probably already figured out that #9840 aims to solve the issue with having to install plugins in two locations 😄 

I'll get on the CSS fixes later.

src/Umbraco.Web.Common/RuntimeMinification/SmidgeRuntimeMinifier.cs
```diff
@@ -85,7 +85,8 @@ public void CreateJsBundle(string bundleName, params string[] filePaths)
             // Here we could configure bundle options instead of using smidge's global defaults.
             // For example we can use our own custom cache buster for this bundle without having the global one
             // affect this or vice versa.
-            var bundle = _bundles.Create(bundleName, _jsPipeline.Value, WebFileType.Js, filePaths);
+            var pipeline = optimizeOutput ? _jsPipeline.Value : _bundles.PipelineFactory.Create();
```
Shazwazza, line 88: Just realized this switch statement will use the default pipeline which we don't want to do, for non-optimized output we should use our own pipeline since we don't want to rely on globals. I'll update.

Shazwazza:
@kjac I've requested your review on this one https://github.com/umbraco/Umbraco-CMS/pull/9990/ if you get some time :)

PR #9840, opened 2021-02-18
Allow serving App_Plugins files to Smidge for minification

kjac:
### Prerequisites

### Description

If you have an extension with JS in App_Plugins, Smidge will throw a `System.IO.FileNotFoundException` when you run the site without debug enabled. I suspect the same happens for CSS, but I haven't verified it.

It happens because Smidge defaults to using the `WebRootFileProvider` when loading files to minify. Since `App_Plugins` is not located within `wwwroot`, the extension files cannot be loaded and thus the exception is thrown.

This PR creates a custom file provider for Smidge. This file provider uses the `ContentRootFileProvider` to load files from `App_Plugins` and the `WebRootFileProvider` for all other files.

#### Steps to recreate

1. Disable debug mode in app settings (`Umbraco.CMS.Hosting.Debug`).
2. Increment the runtime version in app settings (`Umbraco.CMS.RuntimeMinification.version`).
3. Add this test dashboard to App_Plugins: [MyDashboard.zip](https://github.com/umbraco/Umbraco-CMS/files/6003117/MyDashboard.zip)
4. Boot up the site and log into the backoffice with Chrome devtools open.
5. The backoffice won't load. After a while, the network tab will show that the Umbraco bundle throws an error (HTTP 500)
image


src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilderExtensions.cs
```diff
@@ -395,5 +399,29 @@ private static IHostingEnvironment GetTemporaryHostingEnvironment(IWebHostEnviro
...
+                _webRooFileProvider = webRooFileProvider;
+            }
+
+            public IFileInfo GetFileInfo(string subpath) => subpath.InvariantStartsWith(Constants.SystemDirectories.AppPlugins)
+                ? _contentRootFileProvider.GetFileInfo(subpath)
+                : _webRooFileProvider.GetFileInfo(subpath);
+
+            public IDirectoryContents GetDirectoryContents(string subpath) => throw new NotSupportedException();
+
+            public IChangeToken Watch(string filter) => throw new NotSupportedException();
```
Shazwazza, line 424: This will need to be implemented else exceptions will throw if anyone (or us) decides to use Smidge's file watching feature. I think the CompositeChangeToken class could be used since this is wrapping multiple sub providers https://docs.microsoft.com/en-us/aspnet/core/fundamentals/change-tokens?view=aspnetcore-5.0#compositechangetoken-class 

But actually, this whole provider could just use the CompositeFileProvider file system: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.fileproviders.compositefileprovider?view=dotnet-plat-ext-5.0 which i think just takes care of this for us anyways.

Shazwazza (changes requested):
One change needed, see notes

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilderExtensions.cs
kjac: 👍 will have a look this week

Shazwazza: My advice here would be:

* Create a new RuntimeMinifierOptions (IOptions) class and register that in DI with the .Configure extension
* Register SmidgePhysicalFileProvider in DI with a dependency on IOptions<RuntimeMinifierOptions>
* The `RuntimeMinifierOptions` can take a dependency on `IWebHostEnvironment`
* The `SmidgePhysicalFileProvider` then has a private field of type `CompositeFileProvider` which is used to wrap all of it's methods

Then there's no need to pass around the `IWebHostEnvironment` everywhere and these extensions remain clean.

This also paves the way for additional options in the future to be used to configure Smidge with Umbraco... which is something I'm working on now.

But now realize that we'll be working on the same thing so maybe I'll merge this and implement the above :)

kjac: @Shazwazza  Hahah doh! OK.... I was fiddling around with some changes here. Note to future. It does indeed work if you swap `SmidgePhysicalFileProvider` with `CompositeFileProvider` or make `SmidgePhysicalFileProvider` inherit from `CompositeFileProvider`, so a private `CompositeFileProvider` field on `SmidgePhysicalFileProvider` would also work.

My only worry is having file provides floating around which expose the content root. That's why I made it a private class instead of registering it properly in the DI.

Shazwazza: Yep all good, i'm updating it so that it only points to app_plugins not the root and there's a new smidge 4.0.0 beta that includes additional functionality that i've updated to. When I post a PR i'll ping you for a review :)

Shazwazza: @kjac on another note - what was that bug all about that required me to unzip the MyDashboard into the www/app_plugins folder?

kjac: @Shazwazza this PR is a fix for that bug 😄 stated in the description:

> It happens because Smidge defaults to using the WebRootFileProvider when loading files to minify. Since App_Plugins is not located within wwwroot, the extension files cannot be loaded and thus the exception is thrown.

Shazwazza: lol 🤣

PR #9174, opened 2020-10-15
V8/feature/super doctype create menu

marcemarc:
### Prerequisites

- [^ ] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/6410

### Description

Hi, this is an update to the Document Type Create Menu.

Why? - there is a blog post here that contains most of the thinking: https://moriyama.co.uk/about-us/news/creating-doctypes-in-umbraco-v8/

I've been using it as a 'getting started' training support tool when working with people who are new to Umbraco to help them get used to the concepts. I released it as a package:

https://our.umbraco.com/packages/backoffice-extensions/super-doc-type-create-menu/

There has been a clamour to add it to the core, but as the issue, this is linked to states - HQ have better plans than this for the Doc Type menu.

However, have been asked if I would submit a PR anyway by thechiefunicorn:

image

and so I am doing so.

### To test this

Go to Settings - Document Types - and open the Create Menu, you should now see an updated menu with different icons, and descriptions for different types of Document Types you can create, the thinking is in the blog post above.

It's worth testing each menu item, to see if it performs as advertised, the Element Type option should create a Document Type with the 'Is Element Type' checkbox already checked.

### Wording

In terms of the wording of the options, I'm sure it could be improved, really it's the icon and the clear label of the option, and the order of the options that is the improvement (eg no Document Type Without Template option). But I'm happy if people maybe want to tweak the descriptions, etc

I'm aware this removes the Document Type Collection option.

That's it I think!

Thanks for considering this contribution to Umbraco CMS!

PR #9846, opened 2021-02-18
Fix 'Do something else' button position in create dialog

nathanwoulfe:
### Prerequisites

### Description
This fixes the issue mentioned by @bjarnef in #9174 where the Do something else button is not positioned correctly in the create document type dialog. Same issue exists in the create member type dialog, so have fixed that as well.

Nothing major - the `umb-footer-dialog` element was nested inside the `umb-dialog-body` when it should be a sibling.


nathanwoulfe:
@OwainWilliams @mikecp @nul800sebastiaan this needs to be part of the same release as #9174, since it would be a shame to release a beautiful new change with buttons sitting in the wrong place. A prompt merge would be wonderful 😃

src/Umbraco.Web.UI.Client/src/views/membertypes/create.html
```diff
@@ -1,47 +1,53 @@
...
-              val-form-manager>
-
-            <umb-control-group label="Enter a folder name" hide-label="false">
-                <input type="text" name="folderName" ng-model="model.folderName" class="umb-textstring textstring input-block-level" required />
-            </umb-control-group>
-
-            <button type="submit" class="btn btn-primary"><localize key="general_create">Create</localize></button>
-        </form>
+    <div class="umb-dialog-footer btn-toolbar umb-btn-toolbar" ng-if="!model.creatingFolder">
+        <button class="btn btn-info" ng-click="nav.hideDialog(true)">
```
bjarnef, line 49: @nathanwoulfe I think it would be great to add `type` attribute here as well like here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/datatypes/create.html#L43

But not sure it is added in all dialogs.

nathanwoulfe: @bjarnef I've updated to add type attribute in doctype, media and member dialogs 👍

mikecp (approved):

mikecp:
@nathanwoulfe Here you go ;-)
Thanks for the fix 👍

nathanwoulfe:
Thanks @mikecp. I have a feeling that had been broken for a while, but never noticed because the list of options was shorter. 

All good, thanks for the merge!

mikecp:
@nathanwoulfe  Yes probably :-)

Issue #9122, opened 2020-10-08
Attach metadata to icons


callumbwhyte:
Following a demo of the new `<umb-icon>` directive by @MMasey _(looks awesome!)_ a discussion started about how we can make the experience of working with icons nicer.

It was discussed that it would be great to attach additional metadata around icons. This could be used to displaying a label alongside the icon, or improving search within the icon picker.

I think it would be cool to define icons as JSON, rather than files. This way we could store additional metadata.

```
[
    {
        "alias": "cat",
        "path": "umbraco/icons/cat.svg",
        "group": "animals"
    }
]
```

The `alias` field is the current name (minus `-icon`). The `path` is where the image is located (this could be relative to the /umbraco folder or absolute for packages, as is the case elsewhere in the CMS).

The `group` field is a proposal for a better way of grouping icons, perhaps to improve the currently overwhelming icon picker with a grouped experience much like the data type picker.

The Umbraco would have a big JSON file defining icons in this way. Following along the lines of #8884 the `package.manifest` file could also offer a way to define / override custom icons in the same way for package developers.

It's important that any labels or user-facing text can be localised - currently the icon picker search is based on icon CSS class / file names, which are in English and don't accurately represent the icon or are misspelled. The JSON structure doesn't lend itself too well to localisation, but the existing XML dictionary files could handle this by convention:

```
<area alias="icons">
    <key alias="cat">Cat</key>
</area>
```

This is just a dump of ideas in the hope we can agree on an approach and split this into some digestible chunks!

skttl:
Just thinking out loud, how about just adding this meta data as data-attributes in the svg file itself?

```svg
<svg xmlns="http://www.w3.org/2000/svg" 
    viewBox="0 0 512 512" 
    data-umb-alias="yen-bag" 
    data-umb-category="money"
    data-umb-keywords="yen money currency china">
    <path
[...]

PR #9848, opened 2021-02-19
Fixes #9625 - Searching icons for "box" returns all icons

### LINKED ISSUE #9625: Searching icons for "box" returns all icons
I am seeing this issue on Umbraco version: 8.10.1

Reproduction
------------

When searching for icons, it seems like it's searching all the svg code. If I search for box, I get all icons back. Probably because they all have a `viewBox`.

### Steps to reproduce

Open an icon picker, ie. in the doc type editor
Search for box - notice you get a lot of irrelevant icons, also try svg, <svg etc.

### Expected result

I expected it to return only icons in which the title contains the word box.

### Actual result

Seems like I get all icons back.

image

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9625

### Description
This PR fixes issues with filtering in icon picker and searching for terms like `box`, `view` and `svg` because these are included in the `svgString` property. We only need to filter on name property. In future releases we might be able to filter on tags or similar - see https://github.com/umbraco/Umbraco-CMS/issues/9122

KnrdnWkyJc

Btw in this GIF I have installed the [Material Design Icon Pack](https://our.umbraco.com/packages/backoffice-extensions/material-design-icon-pack/), which doen't use SVG icons, but demonstrate the filtering also works with custom icons.

This is similar to other filtering where we are working with objects, e.g. in block picker where filter on "image" doesn't really filter anything as mentioned in https://github.com/umbraco/Umbraco-CMS/pull/9831
We can look at this in another PR.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/iconpicker/iconpicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/iconpicker/iconpicker.controller.js
@@ -10,6 +10,10 @@ function IconPickerController($scope, localizationService, iconHelper) {
 
     var vm = this;
 
+    vm.filter = {
+        searchTerm: ''
+    };
+
     vm.selectIcon = selectIcon;
     vm.selectColor = selectColor;
     vm.submit = submit;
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Thanks again @bjarnef - good stuff! 👍

PR #9849, opened 2021-02-19
Preview does display draft versions

kjac:
### Prerequisites

### Description

The preview function does not display draft versions: 

preview-before

Turns out the current backoffice user isn't authenticated, because the middleware to do so (`PreviewAuthenticationMiddleware`) never gets executed. I have rearranged the order of the middleware, so `PreviewAuthenticationMiddleware` is placed before `UmbracoRequestMiddleware` in the pipeline. This fixes the lack of authentication for the preview function:

preview

I also added a bit of logging to the middleware while I was at it.

I know you're working on rearranging the codebase right now. I suspect I may have done things upside down here - perhaps `UmbracoRequestMiddleware` should have been moved to `Umbraco.Cms.Web.BackOffice`, instead of moving `PreviewAuthenticationMiddleware` to `Umbraco.Cms.Web.Common`? Let me know if it's all wrong and I'll fix things up 😄


bergmania:
💪

PR #9850, opened 2021-02-19
Remove legacy member cache refresher and unused base class

kjac:
### Prerequisites

### Description

This PR removes the `LegacyMemberCacheRefresher` which has been slated for removal in Unicore. 

`LegacyMemberCacheRefresher` was the only class in the core built on top of the public base class `TypedCacheRefresherBase`, so I have removed `TypedCacheRefresherBase` as well. If need be we can always put it back in later on, but right now it feels a little like dead weight.


bergmania:
Thanks @kjac

PR #9852, opened 2021-02-19
Rename UmbracoVersion.Current to UmbracoVersion.Version

kjac:
### Prerequisites

### Description

Renaming `UmbracoVersion.Current` to `UmbracoVersion.Version` is an old TODO from the V8 code. 

I figured it would be nice to get that done, before the Unicore betas start rolling out, as runtime access to the Umbraco version is likely to be used by all sorts of 3rd party packages out there. Best to break it before the betas.

If you feel the urge to test this, create a template with the following content and observe how the current Umbraco version is very neatly printed when it's rendered 😆 

```cshtml
@using Umbraco.Cms.Core.Configuration
@inject IUmbracoVersion umbracoVersion
@inherits Umbraco.Cms.Web.Common.AspNetCore.UmbracoViewPage
@{
  Layout = null;
}
<html>
<body>
  <h1>Version is: @umbracoVersion.Version</h1>
</body>
</html>
```

### DIFF
--- a/src/Umbraco.Core/Configuration/IUmbracoVersion.cs
+++ b/src/Umbraco.Core/Configuration/IUmbracoVersion.cs
@@ -8,7 +8,7 @@ namespace Umbraco.Cms.Core.Configuration
         /// <summary>
         /// Gets the non-semantic version of the Umbraco code.
         /// </summary>
-        Version Current { get; }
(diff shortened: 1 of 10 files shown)


bergmania:
Thanks :)

PR #9853, opened 2021-02-19
Clean up the IO Helper interface

kjac:
### Prerequisites

### Description

There really is no testing this one 😄 I have performed a little housekeeping on `IIOHelper`:

- `TryResolveUrl` has been removed entirely, since it was only called from the unused method `DetectIsJavaScriptPath` on `StringExtensions` (deprecated/legacy JS handling that is no longer in use).
- `ResolveUrl` has been marked as obsolete, as it only forwards the call to `IHostingEnvironment.ToAbsolute` now.
- `IsPathFullyQualified` has been removed from the interface. Its implementation is retained in (and used by) `IOHelper`, but it's likely to be replaced by the .net standard implementation down the line. The method it is never used outside of `IOHelper`, so there is no acute need for it to be part of the interface.

### DIFF
--- a/src/Umbraco.Core/Extensions/StringExtensions.cs
+++ b/src/Umbraco.Core/Extensions/StringExtensions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Umbraco.
+// Copyright (c) Umbraco.
 // See LICENSE for more details.
 
 using System;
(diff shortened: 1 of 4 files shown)


bergmania:
Looks good @kjac . Regarding the `IsPathFullyQualified` I think we could update to .NET Standard 2.1, as soon as we remove the legacy test project

PR #9854, opened 2021-02-19
Removed unused method from migration expression interface

kjac:
### Description

The `Process()` method of the public interface `IMigrationExpression` is unused and has also been flagged for removal, so I removed it 😄 

Both implementations of `Process()` have also been removed, as all they did was to call `ToString()` eventually.


bergmania:
🚀

PR #9855, opened 2021-02-19
Remove obsolete method in UmbracoBuilderExtensions

kjac:
### Description

The extension method `CreateServiceProvider` on `UmbracoBuilderExtensions` is marked as obsolete. 

As of now the extension method is solely used for wiring up tests, so I have moved the implementation to `TestHelper` instead, so the method doesn't clutter the `UmbracoBuilderExtensions` class.


bergmania:
Makes sense :)

Issue #9841, opened 2021-02-18
Wrong focus on Nested Content with max items set to 1


jesperweber:
When having a nested content field with max items set to 1, the field will take focus when loading the node in the Umbraco backoffice.

This means if you have multiple fields on the document type and the nested content field is not at the top, the page will scroll down to the nested content field, since it takes focus, and you have to scroll to the top of the page yourself every time you load the node.

![image](https://user-images.githubusercontent.com/2575817/108368277-8d38ba80-71fa-11eb-98b4-a32beb52d78e.png)

![image](https://user-images.githubusercontent.com/2575817/108371602-3503b780-71fe-11eb-92d4-48f0683e5f5f.png)

### Umbraco version

I am seeing this issue on Umbraco version: 8.11.1

### Expected result

I would not expect the field to take focus when loading the node.

patrickdemooij9:
Created a PR for this: https://github.com/umbraco/Umbraco-CMS/pull/9863
I do wonder if we should also have an option to not auto-open the nested content item on page load.

jesperweber:
Thanks @patrickdemooij9 

Yes I have many cases where I would like the nested content item to not auto open just because I have set the max to 1, so I like the idea of an option to disable it.

patrickdemooij9:
Created a different PR for the not auto open feature: https://github.com/umbraco/Umbraco-CMS/pull/9883

PR #9863, opened 2021-02-21
9841: Don't focus on nested content node at page load

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9841

### Description
How to check (based on starter install):

- Create a new data type (Nested content with Feature as item. Min items: 0, Max items: 1).
- Add this data type to a document type
- Open a node that is of type of that document type
- Create a new nested content node
- Save and publish and reload the page
- The page shouldn't focus on the nested content data type anymore

The nested content item will still focus if you create a new one, so you are able to tab through it.

focus


OwainWilliams (approved):
HI @patrickdemooij9  thanks for the PR. 
I have reviewed it and thanks for your detailed demo of how to reproduce. 

I can confirm that this does fix the focus issue. Thanks.

Issue #9138, opened 2020-10-12
Examine Management: show multiple values for a single field


soreng:
When looking at the indexed data, you only get to see the first value that is indexed. 

In most cases this is fine, but sometimes we index values separately, to allow easier searching.

`Examine.ISearchResult` contains a property called `AllValues` that could be exposed instead of the current impl that only exposes the property called `Values`. The difference between the two, is that AllValues contains a list of strings, and values only contains the first value.

The displayed field values comes from `Umbraco.Web.Models.ContentEditing.SearchResult`. This only implements `Values`, but could easily be changed to also contain `AllValues`, and the view showing the field values, could then allow switching if needed, or just show all values as a list.

To show the values, click the "xx fields" next to a result
![image](https://user-images.githubusercontent.com/359492/95719771-9aec9b00-0c70-11eb-9916-4814b501b958.png)

You then get an editor showing the field values
![image](https://user-images.githubusercontent.com/359492/95719899-cb343980-0c70-11eb-83a9-07be7c8f8523.png)

soreng:
There has been some issues on this earlier, an example is https://github.com/umbraco/Umbraco-CMS/issues/5550

bielu:
@soreng to make that even funnier:
https://github.com/umbraco/Umbraco-CMS/pull/6014

umbrabot:
Hiya @soreng,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #9864, opened 2021-02-22
Handling multiple values per field in Examine Management

callumbwhyte:
Examine / Lucene supports having multiple values per field, but the backoffice Examine Management dashboard only renders the first.

This PR sets to fix that by rendering all values for each field.

I have modified / cleaned up the API to return the true list of fields + values straight from the Examine results. In the backoffice values are now listed in a comma separated (`, `) list using the existing `umbCmsJoinArray` filter.

Before:

Before

After:

After

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/dashboard/settings/examinemanagementresults.html
+++ b/src/Umbraco.Web.UI.Client/src/views/dashboard/settings/examinemanagementresults.html
@@ -18,9 +18,9 @@
                             </tr>
                         </thead>
                         <tbody>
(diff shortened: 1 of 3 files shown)


nathanwoulfe (approved):
Hey @callumbwhyte, a very useful change - never noticed that it only displayed the single value...

soreng:
@nul800sebastiaan seems that this resolves https://github.com/umbraco/Umbraco-CMS/issues/9138 (and the others mentioned in that issue)

Although, using a `,` (comma) as a seperator seems a but to easy, since it is used inside values

Issue #9826, opened 2021-02-16
Missing culture in mapping options error


patrickdemooij9:
Something we found when working on a multilanguage website we have.

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.4 and latest contrib branch

Reproduction
------------

Can be reproduced with the following steps based on an install with the starter kit:

- Update People and Person to be "Allow vary by culture"
- Add a new language
- Go to the log viewer and click on the "Not(@Level='Verbose') and Not(@Level='Debug')". There might be other ways of doing this, but I just managed to find this one. Basically, what you want to do is to get rid of the mculture query string.
- Go back to content and click on the People node
- You'll see the following error:

![image](https://user-images.githubusercontent.com/11466511/108048372-3fc81c00-7047-11eb-9b24-b87cfc945fae.png)

The problem most likely has to do with the mculture query string not being present. However, it should then set the new mculture query string first before attempting to get the node information or just default to the main language.

### Expected result

Shouldn't show error when trying to load node

### Actual result

Shows error like shown above. Full stack trace:
`System.InvalidOperationException: Missing culture in mapping options.
   at Umbraco.Web.Models.Mapping.ContentMapDefinition.GetName(IContent source, MapperContext context) in C:\Users\Patrick\Desktop\Projecten\Git Forks\UmbracoCms\Umbraco-CMS\src\Umbraco.Web\Models\Mapping\ContentMapDefinition.cs:line 214
   at Umbraco.Web.Models.Mapping.ContentMapDefinition.Map(IContent source, ContentItemBasic`1 target, MapperContext context) in C:\Users\Patrick\Desktop\Projecten\Git Forks\UmbracoCms\Umbraco-CMS\src\Umbraco.Web\Models\Mapping\ContentMapDefinition.cs:line 131
   at Umbraco.Core.Mapping.UmbracoMapper.<>c__DisplayClass8_0`2.<Define>b__1(Object source, Object target, MapperContext context) in C:\Users\Patrick\Desktop\Projecten\Git Forks\UmbracoCms\Umbraco-CMS\src\Umbraco.Core\Mapping\UmbracoMapper.cs:line 106
[...]

PR #9865, opened 2021-02-22
9826: Make sure to set mculture after switching sections

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9826

### Description
Bug could be reproduced like this:

- Update People and Person to be "Allow vary by culture"
- Add a new language
- Go to the log viewer and click on the "Not(@Level='Verbose') and Not(@Level='Debug')". There might be other ways of doing this, but I just managed to find this one. Basically, what you want to do is to get rid of the mculture query string.
- Go back to content and click on the People node
- You'll see an error. 

The whole stacktrace can be found in the issue. The problem here was that mculture wasn't in the url anymore and the getChildren call to Umbraco requires a culture to be given.

fixMcultureIssue


mikecp (approved):

mikecp:
Hello @patrickdemooij9 ,

I just checked your PR and it works perfectly. Thanks, this was a nice catch 👍 !

PR #9870, opened 2021-02-22
Keep custom claims that are flowed during security stamp validation

Shazwazza:
When auto linking with callbacks such as OnExternalLogin, custom claims can be added to the user which are flowed to the auth ticket/identity. However, when the security stamp validator executes, the identity is re-created manually without any knowledge of those custom claims so they are lost. 

This ensures that those custom claims flow through to the re-generated identity during the security stamp validation phase.

## Testing

This is annoying to test. The easiest way is to modify source code and breakpoint to verify that it's working.

* Change BackOfficeSignInManager and add a custom claim on sign in. This code will go in the method `PasswordSignInAsyncImpl` above the line: `// We need to verify that the user belongs to one or more groups that define content and media start nodes.` This is basically what would happen if we used auto linking with OnExternalLogin/OnAutoLinking callbacks to assign custom claims to the user.
   ```cs
                user.Claims.Add(new IdentityUserClaim<int>
                {
                    ClaimType = "TEST",
                    ClaimValue = "TESTING"
                });
   ```
* Change the 30 minute default throttling to 3 for testing in `AppBuilderExtensions.UseUmbracoBackOfficeCookieAuthentication`  in the `SecurityStampValidator` call where it is: `TimeSpan.FromMinutes(30)` -> change to `TimeSpan.FromMinutes(3)`
* Change the web.config `Umbraco.Core.TimeOutInMinutes` to 7
* Put a breakpoint in the callback function just below the change to `TimeSpan.FromMinutes(3)` on the line: `var regenerated...`
* Put a breakpoing on the line below that on: `regenerated.MergeClaimsFromBackOfficeIdentity`
* Put a breakpoint just below that on the line: `return regenerated;`

Then F5 and start the app. Login to the back office and wait a few minutes, at least 3 but less than 7. Then click on another section. In theory your break points will be hit. If they are not hit, click on another section (this is all about timing). Once your breakpoint is hit:

* Inspect the value of `identity.Claims` this will show a list of claims assigned to the current ticket and you will see a claim for "TEST"
* F5 to the next breakpoint. Inspect the value of `regenerated.Claims`, this will show a list of claims assigned to the regenerated identity, you will NOT see a "TEST" claim
* F5 to the next breakpoint. Inspect the value of `regenerated.Claims`, this will show a list of claims assigned to the regenerated identity, you WILL see a "TEST" claim because this custom claim has been flowed to the outgoing ticket

Repeat the above again (waiting at least 3 minutes). The "TEST" claim should continue to flow.

Make sure you don't commit your custom code changes :)

### DIFF
--- a/src/Umbraco.Web/Security/AppBuilderExtensions.cs
+++ b/src/Umbraco.Web/Security/AppBuilderExtensions.cs
@@ -175,7 +175,7 @@ namespace Umbraco.Web.Security
 
                     return SecurityStampValidator
                         .OnValidateIdentity<BackOfficeUserManager, BackOfficeIdentityUser, int>(
-                        TimeSpan.FromMinutes(3),
+                        TimeSpan.FromMinutes(30),
                         async (manager, user) =>
                         {
                             var regenerated = await manager.GenerateUserIdentityAsync(user);
--- a/src/Umbraco.Web/Security/AppBuilderExtensions.cs
+++ b/src/Umbraco.Web/Security/AppBuilderExtensions.cs
@@ -171,10 +171,17 @@ namespace Umbraco.Web.Security
                 // change a password or add an external login to your account.
                 OnValidateIdentity = context =>
                 {
+                    // capture the current ticket for the request
                     var identity = context.Identity;
 
                     return SecurityStampValidator
                         .OnValidateIdentity<BackOfficeUserManager, BackOfficeIdentityUser, int>(
+                        // This will re-verify the security stamp at a throttled 30 mins
+                        // (the standard/default set in aspnet identity).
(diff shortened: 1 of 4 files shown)


mikkelhm:
I've tested this on Cloud, and Claims get reapplied after the OnValidateIdentity has been run

bergmania (approved):

bergmania:
My test shows it works as expected

bergmania:
@nul800sebastiaan , what labels do we need to put on this one, and do we need to cherry pink it?

nul800sebastiaan:
Have not looked at the code, but cherry pick to versions that will benefit from this? Not sure which versions this is needed for.

PR #9881, opened 2021-02-23
Replace static events in TreeControllerBase with IEventAggregator notifications

### LINKED ISSUE #9397: Netcore: Remove static events

kjac:
### Prerequisites

### Description

This PR replaces the static events on `TreeControllerBase` with notifications via `IEventAggregator`.

You can use the following code to test the PR:

```cs
using System.Linq;
using Umbraco.Cms.Web.BackOffice.Trees;

namespace My.Testsite
{
    public class MenuRenderingTestHandler : INotificationHandler<MenuRenderingNotification>
    {
        public void Handle(MenuRenderingNotification notification) => notification.Menu.Items.First().Name = "MENU HANDLER";
    }

    public class RootNodeRenderingTestHandler : INotificationHandler<RootNodeRenderingNotification>
    {
        public void Handle(RootNodeRenderingNotification notification) => notification.Node.Name = "ROOT NODE HANDLER";
    }

    public class TreeNodesRenderingTestHandler : INotificationHandler<TreeNodesRenderingNotification>
    {
        public void Handle(TreeNodesRenderingNotification notification)
        {
            if (notification.Nodes.Any() == false)
            {
                return;
            }
            notification.Nodes.First().Name = "TREE NODE HANDLER";
        }
    }

    {
        {
            builder.AddNotificationHandler<MenuRenderingNotification, MenuRenderingTestHandler>();
            builder.AddNotificationHandler<RootNodeRenderingNotification, RootNodeRenderingTestHandler>();
            builder.AddNotificationHandler<TreeNodesRenderingNotification, TreeNodesRenderingTestHandler>();
        }
    }
}
```

It should yield something like this when booting up the backoffice:

image


kjac:
Update... all the notifications have been postfixed `Notification` to improve the discoverability of the notifications.

bergmania:
Amazing work @kjac

PR #9883, opened 2021-02-23
Add an option for closed nested content

patrickdemooij9:
This is a feature that came up in this issue: https://github.com/umbraco/Umbraco-CMS/issues/9841.
I've made an additional PR for this because I think it's something different from the actual fix.

Can be checked with:

- Create a new nested content with min 0, max 1.
- Whenever you open it on a node, it will automatically open if you have an item in it.
- If you check the new option, it'll not automatically open on load.

Personally, I don't see why it should automatically open, but it's probably best to be able to disable it with an option instead of removing the feature.

ezgif com-gif-maker (3)


OwainWilliams:
Hi @patrickdemooij9 
This is a really nice feature. I have one request though - could the help text be reworded? 

"Disables opening when there is one max item available"

Does it mean it will disable open on load when there is more than one item available?

PR #9884, opened 2021-02-23
Don't show configuration header if there is none

patrickdemooij9:
Currently, there is a small bug that shows the configuration header for a data type even if there is no configuration:
image

This PR makes it so that the configuration header is only shown if there actually is one:
image

The best way to check this is:

- Go to a document type
- Add a new property
- Click on "Select editor"
- Choose the member picker
- Click on the member picker so that the editor settings window opens

Now, you won't see the configuration header if there is none.


nathanwoulfe (approved):
Thanks @patrickdemooij9, works perfectly.

N

PR #9885, opened 2021-02-23
Netcore: Update code docs generation

bergmania:
### Details:
Updated the API docs to match new assemblies and namespaces

Static API docs are only generated on "release" builds, so we do not waste time on each continues build.

I made [this custom build](https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=59576&view=results), that "fakes" a release build. The original[ continues build is here](https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=59575&view=results), where you can see the docs are not generated.

The Zip files with docs are available from the [artifacts section of the release build](https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=59576&view=artifacts&pathAsName=false&type=publishedArtifacts).

I [released ](https://dev.azure.com/umbraco/Umbraco%20Cms/_releaseProgress?_a=release-pipeline-progress&releaseId=3188) the docs, so they are available on this new blog storage website:
C# Docs: https://apidocs.umbraco.com/v9/csharp/
JS Docs: https://apidocs.umbraco.com/v9/ui/

As part of this task I also reorganised the azure pipeline stages, so they now represents some features other than just the OS.

New stages (It is possible to only one some stages)
- Unit tests
- Integration tests
- Build artifacts
- Build docs

I had lots of troubles with Azure pipeline and docfx. For some reason the `Umbraco.Infrastructure.csproj` failed (Works localhost) . So instead I filtered such that all `*.cs` files from that project is added. So I think the end result is the same.

Note: Please `Squash` when merging, so all my test commits are combined 🙈


nikolajlauridsen:
Looks good to me 👍 

The only thing that seems a bit odd is that the Infrastructure section only contains `EmailSender` and `PublishedContentQuery` and that the web section only contains `UmbracoHelper` but I assume this is because it's the only public classes in the root namespace.

bergmania:
> but I assume this is because it's the only public classes in the root namespace.

Exactly.. Not really the docs' problem :P

build/azure-pipelines.yml
```diff
@@ -1,277 +1,316 @@
...
+                  displayName: Use .Net Core sdk 5.x
+                  inputs:
+                      version: 5.x
+                - task: DotNetCoreCLI@2
+                  displayName: dotnet test
+                  inputs:
+                      command: test
+                      projects: '**/*.Tests.UnitTests.csproj'
+    - stage: Integration_Tests
+      dependsOn: []
```
nikolajlauridsen, line 63: Oh very minor thing I noticed, but Integration_Tests has no display name, which I think is causing it to show up as Integration_Tests Windows/Linux

build/azure-pipelines.yml
```diff
@@ -1,277 +1,316 @@
...
+      jobs:
+          - job: Linux_Integration_Tests
```
bergmania, line 65:
```suggestion
          - job: Linux_Integration_Tests
            displayName: Integration Tests
```

nikolajlauridsen (approved):

build/azure-pipelines.yml
```diff
@@ -1,277 +1,317 @@
...
+                  displayName: Use .Net Core sdk 5.x
+                  inputs:
+                      version: 5.x
+                - task: DotNetCoreCLI@2
+                  displayName: dotnet test
+                  inputs:
+                      command: test
+                      projects: '**/*.Tests.UnitTests.csproj'
+    - stage: Integration_Tests
+      displayName: Unit Tests
```
nikolajlauridsen, line 63:
```suggestion
      displayName: Integration Tests
```

bergmania: DOH! 🙈

Issue #9825, opened 2021-02-16
Umbraco Grid layout can only configure allowed rows for one column


D-Inventor:
When creating and configuring a Grid datatype, I want to configure particular rows for particular sections in a layout. When attempting this, I find that I can only configure one section and then the configuration of all the other sections (including ones in a different layout) are gone / reset.

## Umbraco version

I am seeing this issue on Umbraco version: 8.9.1 & 8.10.1

Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

### Bug summary

The allowed rows configuration of a section in an umbraco grid layout datatype seems to be maintained only for a single section.

### Specifics

Using Umbraco version 8.9.1, but the same behaviour also occurs on version 8.10.1
Using Edge version 88.0.705.68 (64 bit)
Also checked with Firefox version 84.0.2 (64 bit)
No error messages

### Steps to reproduce

1. Log in to backoffice
2. Go to Settings and create a new Grid layout data type
3. Create a layout with 2+ columns
4. Create 2+ row configurations
5. Select the layout, select a column and disable "Allow all row configurations" and enable or disable a few row configurations
6. Select another column in this layout and repeat previous step with different rows enabled/disabled
7. Go back to previous column and notice that the row configurations are now the same as on the second
8. Go back to the second column and notice that the row configuration has been reset and "Allow all row configurations" is enabled again

### Expected result

I expected to configure different allowed rows for different columns in a grid layout. I expect the configurations to be maintained for each column.

### Actual result

After configuring 2 columns, the configuration for one of the columns disappears / resets.
Following gif shows the behaviour that I'm seeing:
![ezgif-5-93cdbd2d0bc9](https://user-images.githubusercontent.com/25585418/108047960-b9abd580-7046-11eb-892b-4ac878ce9cca.gif)

---
_This item has been added to our backlog [AB#10595](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10595)_

---
_This item has been added to our backlog [AB#10593](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10593)_

nul800sebastiaan:
[...]

PR #9887, opened 2021-02-23
Adjust grid editor layout configuration

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9825

### Description
This PR fixes a few issues with configuration of layout/rows.

- Fixed wrong end-closing tag of `<button>` element which was `</a>` + added missing `type` attribute on button element.
- Registred functions on `vm` to be a bit more consistent with the rest of code base. I found a function `closeArea` in row configuration, which doesn't seem to be used, but I have leaved it for now.
- Previous the rows in layout configuration was using traditional checkboxes along with `checklist-model` and `checklist-value` which auto-populated the `allowed` array with row names. However since we are using `<umb-checkbox>` we need to bind the value to a model, but the row itself was not unique per section/column. I have extended the `sections` in the layout with a rows, which basically have a copy of the rows and a selected state. This is used to bind the value on the checkbox and/or if the button element next to it is clicked. On submit this property is deleted again, so we have to original JSON.
The might be a better way to handle this in future, but for now it seems to do the trick :)

**Layout configuration**

jzjtVAmXkf

**Row configuration**

YvozltVuL3


nathanwoulfe:
Relates to #9410

madsrasmussen:
Hi Bjarne, 

Thanks for the PR 🎉 

When testing it a found a couple of smaller things that I have fixed.

* When adding a new section in the layouts dialog we didn't copy the rows so there was nothing to select.
* When cleaning up rows in the submit function it looked in the wrong place.
* I have also made sure to clean up on close

You can see my commit here: 
https://github.com/umbraco/Umbraco-CMS/pull/9887/commits/ca4d43d28a8ce537a61eae35688a3fc0cc9cc494

Does this look ok to you and can I get you to do an extra test with the new fixes?

bjarnef:
Hi Mads

Great, I have tested the PR with your changes, and it seems to work as expected. I minor thing is when checking all rows or editors and submit the overlay and re-open it again, it could probably have the "Allow all row configurations" or "Allow all editors" toggle enabled, but I think it has always worked this way.

The input fields could be closer the labels by removing the `form-horizontal` class, but it may cause issues with other overlays and something that can be looked at in another PR.

image

image

madsrasmussen:
Thanks Bjarne for taking you time to do an extra test 🥇 

Yes, I agree that the overall design of the dialogs could use some more love. I can see that there is a PR from Nathan with some style clean-up. I will have to see if that also includes some UI love :)

I think this PR does what it should and I will get it merged!

nathanwoulfe:
Most of my changes are polish and cleanup, not substantially changing anything. Still be good to merge, nothing wrong with a bit of polish ☺️

PR #9888, opened 2021-02-24
The Value() method for IPublishedContent was not working with the defaultValue parameter

Shazwazza:
The method PublishedPropertyExtension `public static T Value<T>(this IPublishedProperty property, string culture = null, string segment = null, Fallback fallback = default, T defaultValue = default)` does not pass along the `defaultValue` and instead just returns `default` which isn't correct.

The method `public static T Value<T>(this IPublishedContent content, string alias, string culture = null, string segment = null, Fallback fallback = default, T defaultValue = default)` was not passing defaultValue to the inner property.Value call

### DIFF
--- a/src/Umbraco.Web/PublishedContentExtensions.cs
+++ b/src/Umbraco.Web/PublishedContentExtensions.cs
@@ -173,7 +173,7 @@ namespace Umbraco.Web
                 return value;
 
             // else... if we have a property, at least let the converter return its own
-            // vision of 'no value' (could be an empty enumerable) - otherwise, default
+            // vision of 'no value' (could be an empty enumerable) - otherwise, defaultValue
             return property == null ? defaultValue : property.Value<T>(culture, segment, defaultValue: defaultValue);
         }
 
--- a/src/Umbraco.Web/PublishedContentExtensions.cs
+++ b/src/Umbraco.Web/PublishedContentExtensions.cs
@@ -174,7 +174,7 @@ namespace Umbraco.Web
 
             // else... if we have a property, at least let the converter return its own
             // vision of 'no value' (could be an empty enumerable) - otherwise, default
-            return property == null ? default : property.Value<T>(culture, segment);
+            return property == null ? default : property.Value<T>(culture, segment, defaultValue: defaultValue);
         }
 
         #endregion
--- a/src/Umbraco.Web/PublishedContentExtensions.cs
+++ b/src/Umbraco.Web/PublishedContentExtensions.cs
@@ -174,7 +174,7 @@ namespace Umbraco.Web
 
             // else... if we have a property, at least let the converter return its own
             // vision of 'no value' (could be an empty enumerable) - otherwise, default
-            return property == null ? default : property.Value<T>(culture, segment, defaultValue: defaultValue);
+            return property == null ? defaultValue : property.Value<T>(culture, segment, defaultValue: defaultValue);
         }
 
         #endregion
--- a/src/Umbraco.Web/PublishedPropertyExtension.cs
+++ b/src/Umbraco.Web/PublishedPropertyExtension.cs
@@ -64,7 +64,7 @@ namespace Umbraco.Web
             var noValueConverted = noValue.TryConvertTo<T>();
             if (noValueConverted) return noValueConverted.Result;
 
-            // cannot cast noValue nor convert it, nothing we can return but 'default'
+            // cannot cast noValue nor convert it, nothing we can return but 'defaultValue'
             return defaultValue;
         }
 
--- a/src/Umbraco.Web/PublishedPropertyExtension.cs
+++ b/src/Umbraco.Web/PublishedPropertyExtension.cs
@@ -43,22 +43,29 @@ namespace Umbraco.Web
 
                 // cannot cast nor convert the value, nothing we can return but 'default'
                 // note: we don't want to fallback in that case - would make little sense
-                return default;
+                return defaultValue;
             }
 
             // we don't have a value, try fallback
             if (PublishedValueFallback.TryGetValue(property, culture, segment, fallback, defaultValue, out var fallbackValue))
+            {
                 return fallbackValue;
+            }
 
             // we don't have a value - neither direct nor fallback
             // give a chance to the converter to return something (eg empty enumerable)
             var noValue = property.GetValue(culture, segment);
             if (noValue is T noValueAsT) return noValueAsT;
+            if (noValue == null)
+            {
+                return defaultValue;
+            }
+
             var noValueConverted = noValue.TryConvertTo<T>();
             if (noValueConverted) return noValueConverted.Result;
 
             // cannot cast noValue nor convert it, nothing we can return but 'default'
-            return default;
+            return defaultValue;
         }
 
         #endregion
--- a/src/Umbraco.Web/PublishedPropertyExtension.cs
+++ b/src/Umbraco.Web/PublishedPropertyExtension.cs
@@ -55,11 +55,11 @@ namespace Umbraco.Web
             // we don't have a value - neither direct nor fallback
             // give a chance to the converter to return something (eg empty enumerable)
             var noValue = property.GetValue(culture, segment);
-            if (noValue is T noValueAsT) return noValueAsT;
             if (noValue == null)
             {
                 return defaultValue;
             }
+            if (noValue is T noValueAsT) return noValueAsT;
 
             var noValueConverted = noValue.TryConvertTo<T>();
             if (noValueConverted) return noValueConverted.Result;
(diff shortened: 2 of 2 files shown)


src/Umbraco.Web/PublishedContentExtensions.cs
```diff
@@ -174,7 +174,7 @@ public static T Value<T>(this IPublishedContent content, string alias, string cu
 
             // else... if we have a property, at least let the converter return its own
             // vision of 'no value' (could be an empty enumerable) - otherwise, default
-            return property == null ? default : property.Value<T>(culture, segment);
+            return property == null ? default : property.Value<T>(culture, segment, defaultValue: defaultValue);
```
mikecp, line 177: Shouldn't we return defaultValue in case the property is null as well? Or do we want to make a difference between "property exists but has no value" and "property does not exist"?

src/Umbraco.Web/PublishedPropertyExtension.cs
```diff
@@ -43,22 +43,29 @@ public static T Value<T>(this IPublishedProperty property, string culture = null
...
             if (PublishedValueFallback.TryGetValue(property, culture, segment, fallback, defaultValue, out var fallbackValue))
+            {
                 return fallbackValue;
+            }
 
             // we don't have a value - neither direct nor fallback
             // give a chance to the converter to return something (eg empty enumerable)
             var noValue = property.GetValue(culture, segment);
             if (noValue is T noValueAsT) return noValueAsT;
+            if (noValue == null)
```
mikecp, line 59: Maybe we could put this before the test for noValueAsT?

mikecp:
Hello @Shazwazza ,

I've been through the PR and left a few comments/questions for you :-)
I have also a question about the comments in general: could it be worth replacing the `nothing we can return but 'default'` comments by `nothing we can return but 'defaultValue'`?

Cheers!

Michael.

src/Umbraco.Web/PublishedContentExtensions.cs
Shazwazza: Yes good catch

Shazwazza, line 177:
```suggestion
            return property == null ? defaultValue : property.Value<T>(culture, segment, defaultValue: defaultValue);
```

src/Umbraco.Web/PublishedPropertyExtension.cs
Shazwazza: yep sure

src/Umbraco.Web/PublishedPropertyExtension.cs
```diff
@@ -43,22 +43,29 @@ public static T Value<T>(this IPublishedProperty property, string culture = null
...
+            {
+                return defaultValue;
+            }
```
Shazwazza, line 62:
```suggestion
            if (noValue == null)
            {
                return defaultValue;
            }
            if (noValue is T noValueAsT) return noValueAsT;
```

src/Umbraco.Web/PublishedPropertyExtension.cs
```diff
@@ -43,22 +43,29 @@ public static T Value<T>(this IPublishedProperty property, string culture = null
 
                 // cannot cast nor convert the value, nothing we can return but 'default'
```
Shazwazza, line 44:
```suggestion
                // cannot cast nor convert the value, nothing we can return but 'defaultValue'
```

src/Umbraco.Web/PublishedPropertyExtension.cs
```diff
@@ -43,22 +43,29 @@ public static T Value<T>(this IPublishedProperty property, string culture = null
...
+            if (noValue == null)
+            {
+                return defaultValue;
+            }
             if (noValue is T noValueAsT) return noValueAsT;
+
             var noValueConverted = noValue.TryConvertTo<T>();
             if (noValueConverted) return noValueConverted.Result;
 
             // cannot cast noValue nor convert it, nothing we can return but 'default'
```
Shazwazza, line 67:
```suggestion
            // cannot cast noValue nor convert it, nothing we can return but 'defaultValue'
```

Shazwazza:
@mikecp I have updated the code based on your comments :)

mikecp:
Yay! Thanks @Shazwazza , let's merge it now ;-)
Cheers!

mikecp:
Hello again @Shazwazza ,
I received a mention from @nul800sebastiaan that some unit tests were failing since I merged this PR.
I investigated and found the crashing test in `Umbraco.Tests.PublishedContent.PublishedContentLanguageVariantTests`:
```

 [Test]
  public void Can_Fallback_To_Default_Value()
  {
      var content = Current.UmbracoContext.Content.GetAtRoot().First().Children.First();

      // no Dutch value is assigned, so getting null
      var value = content.Value("welcomeText", "nl");
      Assert.IsNull(value);

      // even if we 'just' provide a default value
      value = content.Value("welcomeText", "nl", defaultValue: "woop");
     Assert.IsNull(value);

      // but it works with proper fallback settings
      value = content.Value("welcomeText", "nl", fallback: Fallback.ToDefaultValue, defaultValue: "woop");
      Assert.AreEqual("woop", value);
  }

```

The second Assert is now crashing because in that case we now return the defaultValue "woop" instead of the expected "null'.

So it seems that the test actually expects the defaultValue parameter to be ignored, except if we define a fallback option Fallback.ToDefaultValue, and we have changed this. The behaviour we now have is more or less as if the default value for the fallback parameter was now FallBack.DefaultValue

I was wondering what was the best way of handling this and @nul800sebastiaan asked me to check with you: adapt the unit test to the new logic or revert to the old logic?

Thanks in advance for your feedback!

Shazwazza:
@nul800sebastiaan + @mikecp I suppose the PublishedValueFallback intended was overlooked and that this would be a breaking change. Perhaps I misunderstood the 'defaultValue' parameter and how it interacts with the PublishedValueFallback.

So perhaps the only fix here is to ensure we pass `defaultValue` to the inner Value method and the rest should be reverted?

mikecp:
@Shazwazza @nul800sebastiaan In the original inner Value method, the "defaultValue" is only used when checking for a Fallback value:

```
// we don't have a value, try fallback
if (PublishedValueFallback.TryGetValue(property, culture, segment, fallback, defaultValue, out var fallbackValue))
    return fallbackValue;

```

So I guess that we should either revert completely, either send both the fallback and the defaultValue to that inner method.

But this could then be a breaking change as well, because I think the inner method might start sending back `defaultValue` instead of `default` to that "parent" method (I can check/maybe there is a unit test missing on that?). But this, on the other hand might be seen as a (breaking) bugfix, since in the original version we were calling the inner method in a way that ignored the fallback and defaultValue parameters provided to the "parent" method.

What do you think?

Shazwazza:
I'm pretty sure we want to call the inner with the defaultValue, i think that is just something overlooked. A breaking fix is fine in most cases - really all fixes of old bugs are breaking fixes - but I think changing the null return to defaultValue is too much of a change and seems like that was the intention.

mikecp:
OK, seems fair to me 😊

Just coming back on one point on which I was too vague in previous comment, and wanting to be sure: when we call the inner method and send only the defautValue parameter, it means that the `fallback `parameter of that inner method will take the value `default`:

`public static T Value<T>(this IPublishedProperty property, string culture = null, string segment = null, Fallback fallback = default, T defaultValue = default)
`

And therefore, the fallback call I mentioned will in fact never return the defautlValue, since fallback will not be set to Fallback.ToDefaultValue, making the sending of only the "defaultValue" parameter kind of useless as far as I understand :

```
// we don't have a value, try fallback
if (PublishedValueFallback.TryGetValue(property, culture, segment, fallback, defaultValue, out var fallbackValue))
    return fallbackValue;

```

This is why I was suggesting to also send the `fallback` parameter of the parent to the inner method in that case, so that getting the fallback in the inner method has the same behaviour as getting the fallback in the "parent method".

If `Fallback.ToDefaultValue` was sent to the parent, I suppose we can/should send that to the inner method as well 😊

Shazwazza:
Yep! I agree, the fallback should be sent to

mikecp:
Cool! How do we proceed further? Do you create a revert PR or do you want me to create it?

Shazwazza:
Hey @mikecp If you have time to make a revert PR that would be fabulous! else I'll try to get one done tomorrow. Cheers!

mikecp:
Hello @Shazwazza, OK no prob, I'll probably make it tonight then (which will be "tomorrow" for you😊)
Cheers!

nul800sebastiaan:
This has mostly been reverted in https://github.com/umbraco/Umbraco-CMS/pull/9989 but while preserving passing on the fallback value.

PR #9889, opened 2021-02-24
Improve performance, readability and handling of FollowInternalRedirects

Shazwazza:
The FollowInternalRedirects didn't handle values very well and would end up logging warnings (as a debug log) that shouldn't be there. Also there was unnecessary performance overhead of this method even though it runs on **every** request.

Don't double parse the property `Constants.Conventions.Content.InternalRedirectId` value. This was trying to do a `Value<int>()` with a defaultValue and then a `Value<GuidUdi>()`. This unnecessary and causes overhead, boxing and a bunch of irrelevant conversions and doesn't allow for doing the simple null/empty check that it should have been doing. Now we just get the string, check for empty/null, then check if it parses to INT or GuidUdi

Don't log warnings if the value is empty - an empty value is perfectly valid

This simplifies the logic


nathanwoulfe (commented):
@Shazwazza definitely more readable. Logic makes sense too 👍

Shazwazza:
@nathanwoulfe For any PRs I've submitted with `community-review` on them I'm just assuming that the community will merge if when testing/code review is approved. Is that the case?

nathanwoulfe:
@Shazwazza I don't see why not... I'll give it a proper run today, but it all makes waaay more sense than the previous version.

mikecp:
We can only merge to v8/contrib though, so I guess "Approve" is as far as we can go here 😉

nathanwoulfe:
@Shazwazza any issue with retargeting this to v8/contrib?

Shazwazza:
I have retargeted and fixed up merge conflicts 👍

nathanwoulfe (approved):
Tested with:
- node with no umbracoInternalRedirectId property
- node with empty property (content picker)
- node with empty property (textbox)
- node with selected content
- node with int string
- node with non-int string

All works fine.

PR #9892, opened 2021-02-24
Replace static editor model events events with IEventAggregator notifications

### LINKED ISSUE #9397: Netcore: Remove static events

kjac:
### Prerequisites

### Description

This PR replaces the static events for editor models event handling with notifications via `IEventAggregator`.

You can use the following code to test the PR:

```cs
using System.Linq;
using Umbraco.Cms.Web.BackOffice.Filters;

namespace My.Testsite
{
    public class SendingContentTestHandler : INotificationHandler<SendingContentNotification>
    {
        public void Handle(SendingContentNotification notification) => notification.Content.Variants.First().Name = "CONTENT NOTIFICATION HANDLER";
    }

    public class SendingMediaTestHandler : INotificationHandler<SendingMediaNotification>
    {
        public void Handle(SendingMediaNotification notification) => notification.Media.Name = "MEDIA NOTIFICATION HANDLER";
    }

    public class SendingMemberTestHandler : INotificationHandler<SendingMemberNotification>
    {
        public void Handle(SendingMemberNotification notification) => notification.Member.Name = "MEMBER NOTIFICATION HANDLER";
    }

    public class SendingUserTestHandler : INotificationHandler<SendingUserNotification>
    {
        public void Handle(SendingUserNotification notification) => notification.User.Name = "USER NOTIFICATION HANDLER";
    }

    public class SendingDashboardsTestHander : INotificationHandler<SendingDashboardsNotification>
    {
        public void Handle(SendingDashboardsNotification notification)
        {
            foreach (var dashboard in notification.Dashboards)
            {
                dashboard.Label = "DASHBOARD NOTIFICATION HANDLER";
            }
        }
    }

    {
        {
            builder.AddNotificationHandler<SendingContentNotification, SendingContentTestHandler>();
            builder.AddNotificationHandler<SendingMediaNotification, SendingMediaTestHandler>();
            builder.AddNotificationHandler<SendingMemberNotification, SendingMemberTestHandler>();
            builder.AddNotificationHandler<SendingUserNotification, SendingUserTestHandler>();
            builder.AddNotificationHandler<SendingDashboardsNotification, SendingDashboardsTestHander>();
        }
    }
}
```

It should yield something like this for the respective parts of the backoffice:

dashboard handler

content handler

media handler

member handler

user handler


src/Umbraco.Web.BackOffice/Filters/OutgoingEditorModelEventAttribute.cs
```diff
@@ -25,33 +29,51 @@ private class OutgoingEditorModelEventFilter : IActionFilter
...
+                    else if (model is UserDisplay user)
+                    {
+                        _eventAggregator.Publish(new SendingUserNotification(user, umbracoContext));
+                    }
+                    else if (model is IEnumerable<Tab<IDashboardSlim>> dashboards)
                     {
-                        var args = new EditorModelEventArgs(model, umbracoContext);
-                        EditorModelEventManager.EmitEvent(context, args);
-                        objectContent.Value = args.Model;
+                        _eventAggregator.Publish(new SendingDashboardsNotification(dashboards, umbracoContext));
```
bergmania, line 76: I prefer a switch case with pattern matching in this case..

bergmania (commented):
Looks good.. What do you think about my suggestion here?

bergmania (approved):

src/Umbraco.Web.BackOffice/Filters/OutgoingEditorModelEventAttribute.cs
kjac: @bergmania right you are 😄 I have updated the PR

bergmania:
🚀

src/Umbraco.Web.BackOffice/Filters/SendingContentNotification.cs
```diff
@@ -0,0 +1,19 @@
+using Umbraco.Cms.Core.Events;
+using Umbraco.Cms.Core.Models.ContentEditing;
+using Umbraco.Cms.Core.Web;
+
+namespace Umbraco.Cms.Web.BackOffice.Filters
+{
+    public class SendingContentNotification : INotification
```
Shazwazza, line 7: I feel like it makes sense to make all `INotification` implementing classes as `sealed` since I don't really see a scenario where they would be inherited? What do you think? cc @bergmania ?

bergmania: Cool with me..

kjac: Makes sense. I've got a huge PR incoming for content and media events rewrite... I'll make all the new notifications there sealed, then I can revisit the already merged ones afterwards.

PR #9893, opened 2021-02-24
Netcore: Fix installing packages

nikolajlauridsen:
Fixed an issue where you could no longer install packages.

The issue was that the ValidateAngularAntiForgeryTokenFilter checked for a cookie path if that path didn't exist it called next but didn't return causing next to be called twice leading to some issues. I ~~removed this since it seems like legacy code, we check for `ClaimTypes.CookiePath` but never use it~~, added a return statement, see the comment below.

The second issue was with the UI, if the server hasn't restarted within 2 seconds we called install error, we now wait longer.

Made these changes with help from @bergmania so this has in effect already been reviewed.

### DIFF
--- a/src/Umbraco.Web.BackOffice/Filters/ValidateAngularAntiForgeryTokenAttribute.cs
+++ b/src/Umbraco.Web.BackOffice/Filters/ValidateAngularAntiForgeryTokenAttribute.cs
@@ -44,15 +44,6 @@ namespace Umbraco.Cms.Web.BackOffice.Filters
 
             public async Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next)
             {
-                if (context.Controller is ControllerBase controller && controller.User.Identity is ClaimsIdentity userIdentity)
-                {
-                    // if there is not CookiePath claim, then exit
-                    if (userIdentity.HasClaim(x => x.Type == ClaimTypes.CookiePath) == false)
-                    {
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web.BackOffice/Filters/ValidateAngularAntiForgeryTokenAttribute.cs
```diff
@@ -44,15 +44,6 @@ public ValidateAngularAntiForgeryTokenFilter(ILogger<ValidateAngularAntiForgeryT
 
             public async Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next)
             {
-                if (context.Controller is ControllerBase controller && controller.User.Identity is ClaimsIdentity userIdentity)
```
Shazwazza, line 47: This is a check to ensure that any endpoint that requires anti forgery done is actually performed when the authentication ticket is based on a cookie. Without this check it means that other auth types like Bearer token auth that does not rely on cookies will fail the anti forgery check.

nikolajlauridsen: Ah i see, then we of course can't remove it completely. I've reintroduced it, adding a return statement after the `next()` call, if I understand it correctly, we want to entirely skip this middleware if there is no cookie since it's dependant upon it?

PR #9900, opened 2021-02-25
Remove static methods and events from EmailSender

### LINKED ISSUE #9397: Netcore: Remove static events

kjac:
### Prerequisites

### Description

This PR replaces static methods with instance methods and removes the static event from `EmailSender`. 

As it turns out, the static event isn't used by the core. With the removal of the static methods on `EmailSender`, the whole reason for the static event also becomes non-existent, as it is now possible to replace `EmailSender` entirely with your own implementation - something like this:

```cs
using System.Threading.Tasks;
using Umbraco.Cms.Core.Mail;
using Umbraco.Cms.Core.Models;
using Umbraco.Extensions;

namespace My.Testsite
{
    public class MyEmailSender : IEmailSender
    {
        public Task SendAsync(EmailMessage message) => Task.CompletedTask;

        public bool CanSendRequiredEmail() => true;
    }

    {
        public void Compose(IUmbracoBuilder builder) => builder.Services.AddUnique<IEmailSender, MyEmailSender>();
    }
}
```

This PR also paves the way for replacing the static events in `BackOfficeUserManager` with notifications using `IEventAggregator`. I'll look at a PR for that later on, if this one is accepted.

### Testing this PR

Use the code above to test this. As `CanSendRequiredEmail()` returns `true`, the backoffice login screen will show an option to recover your password by email:

image


src/Umbraco.Infrastructure/EmailSender.cs
```diff
@@ -75,35 +31,33 @@ public void Send(EmailMessage message)
...
 
-                    await client.DisconnectAsync(true);
+                var mailMessage = ConstructEmailMessage(message);
+                if (_globalSettings.Smtp.DeliveryMethod == SmtpDeliveryMethod.Network)
+                {
+                    await client.SendAsync(mailMessage);
+                }
+                else
+                {
+                    await client.SendAsync(mailMessage);
```
bergmania, line 57: This seems odd.. I guess there was a reason it was not async in some cases before

bergmania (commented):
Just found one thing we need to investigate

src/Umbraco.Infrastructure/EmailSender.cs
bergmania: It has been like that since: https://github.com/umbraco/Umbraco-CMS/commit/4b7bd6482efd999588ccac27c174d8ab6c6de3b2, where it was copied from another file..

kjac: Hmm... one might argue that there is no reason for the if/else statement 😆 silly me. That one slipped my own review. On the other hand I guess there is no problem leaving the synchronous email send be for now, just to play it safe?

bergmania: I agree.. Play safe and we can merge

bergmania (approved):

src/Umbraco.Infrastructure/EmailSender.cs
kjac: @bergmania it's updated now 👍

bergmania: Now we have another conflicts 🙈

kjac: Ha! I'll resolve 😆

kjac: That ought to do it 😄

PR #9902, opened 2021-02-26
Update to enable members to use the same password change as users

emmagarland:
### Prerequisites

This links to PR feedback by @Shazwazza on #9472.

### Description

- Updating password changing so that the logic to decide if they are allowed to change the password is in the controller
- Made `PasswordChanger` class generic so it can be reused by the other controllers with the specific usermanager required
- Made `PasswordChanger` injectable
- Updated tests

### To test

- Test password changing functionality for the `UsersController`, the `CurrentUserController `and the `MemberController`.
- Ensure this works in all expected ways for Umbraco functionality

### Questions

- [ ]  Is it right that I can change the password via the Users interface without being shown the password policy?
- [ ] Members now returns errors when changing in Members and the password policy isn't met - but this doesn't happen for Users. Is that right?
- [ ] Are there any other password rules to follow? e.g. not the same as previous password, etc.
- [ ] Do you need to check the user permissions for current and non-current user controller? (I'm presuming not since they're already logged in)

### Known Issues

- The validation error doesn't disappear when the password validation is wrong. See [this chat](https://github.com/umbraco/Umbraco-CMS/pull/9472#discussion_r555959744)


src/Umbraco.Tests.Integration/Umbraco.Core/Packaging/PackageDataInstallationTests.cs
```diff
@@ -21,8 +21,8 @@
 using Umbraco.Cms.Infrastructure.Serialization;
 using Umbraco.Cms.Tests.Common.Testing;
 using Umbraco.Cms.Tests.Integration.Testing;
+using Umbraco.Cms.Tests.Integration.Umbraco.Infrastructure.Services.Importing;
```
emmagarland, line 24: I don't know why these designer files changed!

src/Umbraco.Tests.Integration/Umbraco.Infrastructure/Services/Importing/ImportResources.Designer.cs
```diff
@@ -8,7 +8,7 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
```
emmagarland, line 9: I didn't consciously change these - but they updated

src/Umbraco.Web.BackOffice/Controllers/MemberController.cs
```diff
@@ -453,13 +459,56 @@ private async Task<ActionResult<bool>> UpdateMemberAsync(MemberSave contentItem)
...
+                }
+
+                IMember foundMember = _memberService.GetById(intId.Result);
+                if (foundMember == null)
+                {
+                    return new ValidationErrorResult("Member was not found");
+                }
+
+                IUser currentUser = _backOfficeSecurityAccessor.BackOfficeSecurity.CurrentUser;
+                // if the current user has access to reset/manually change the password
```
emmagarland, line 475: @Shazwazza now the logic has been moved to the controller, I don't know if we need to check the section access for changing member passwords?

emmagarland:
Made the amendments to password changing as requested on #9472 

Also, just wanted to check this is right - when I try to edit a member with no access to the sensitive data, I see all these properties like this:

image

scottbrady91 (approved):
Looks good to me!

src/Umbraco.Web.BackOffice/Security/PasswordChanger.cs
```diff
@@ -1,107 +1,104 @@
...
+            if (identityUser == null)
             {
-                //this really shouldn't ever happen... but just in case
+                // this really shouldn't ever happen... but just in case
                 return Attempt.Fail(new PasswordChangedModel { ChangeError = new ValidationResult("Password could not be verified", new[] { "oldPassword" }) });
             }
 
-            //Are we just changing another user's password?
-            if (passwordModel.OldPassword.IsNullOrWhiteSpace())
+            // Are we just changing another user/member's password?
```
emmagarland, line 65: @Shazwazza @bergmania I've just merged from dev as there was a conflict, this line been added but I am relying on this logic being done in the calling controller now:

   ```
         if (!currentUser.IsAdmin() && savingUser.IsAdmin())
            {
                return Attempt.Fail(new PasswordChangedModel { ChangeError = new ValidationResult("The current user cannot change the password for the specified user", new[] { "resetPassword" }) });
```

Comment (emmagarland) [reply src/Umbraco.Web.BackOffice/Security/PasswordChanger.cs:65]: I will update this logic in the calling controller

Comment (emmagarland) [reply src/Umbraco.Web.BackOffice/Security/PasswordChanger.cs:65]: @bergmania I have updated this logic in `UsersController`. This `PasswordChanger` is also used in 'CurrentUserController and 'MembersController'. I don't think the same admin logic applies there, but please let me know if it should.

Comment (emmagarland) [reply src/Umbraco.Web.BackOffice/Security/PasswordChanger.cs:65]: (I suggest someone who made the security fix tests this branch against the same criteria)

Comment (bergmania) [reply src/Umbraco.Web.BackOffice/Security/PasswordChanger.cs:65]: Works as expected

File: src/Umbraco.Web.BackOffice/Controllers/MemberController.cs [hunk 5]
```diff
@@ -453,13 +459,57 @@ private async Task<ActionResult<bool>> UpdateMemberAsync(MemberSave contentItem)
...
+
+                if (!passwordChangeResult.Success)
+                {
+                    return new ValidationErrorResult("The password could not be changed");
+                }
+
+                // get the identity member now the password and dates have changed
+                identityMember = await _memberManager.FindByIdAsync(contentItem.Id.ToString());
+
+                //TODO: confirm this is correct
```
Shazwazza, line 507: we don't need to do this, the password changer persists the values.

src/Umbraco.Web.BackOffice/Controllers/UsersController.cs
```diff
@@ -661,21 +666,38 @@ public async Task<ActionResult<ModelWithNotifications<string>>> PostChangePasswo
...
+            IUser currentUser = _backofficeSecurityAccessor.BackOfficeSecurity.CurrentUser;
+
+            // if it's the current user, the current user cannot reset their own password
+            if (currentUser.Username == found.Username)
+            {
+                return new ValidationErrorResult("Password reset is not allowed");
+            }
+
+            // if the current user has access to reset/manually change the password
+            if (currentUser.HasSectionAccess(Constants.Applications.Users) == false)
```
Shazwazza, line 690: This is already handled with the controllers `[Authorize(Policy = AuthorizationPolicies.SectionAccessUsers)]`

Shazwazza (changes requested):
Great progress. I've left some comments inline but am also making some changes to your PR today (have never done that before so we'll see if it works!). I'll reply with the rev when I'm done. I'll deal with most of the comments in the member controller.

Shazwazza:
> Also, just wanted to check this is right - when I try to edit a member with no access to the sensitive data, I see all these properties like this:

I think that's right since all of those properties are marked as sensitive on the member type:

image

> Is it right that I can change the password via the Users interface without being shown the password policy?

Not sure if it's 'right' but I think that is what happens in v8 so for this PR we shouldn't worry about that, though it would be nice to have it. In the future too, it would be much better to change the member password changing to be like the users where it is "out of context" and done in a separate request. This would also simplify the UpdateUser logic too.

>  Members now returns errors when changing in Members and the password policy isn't met - but this doesn't happen for Users. Is that right?

This happens for users, if you enter an invalid password it will tell you. I think you are referring to the member controller using `ValidatePasswordAsync` and the user stuff not doing that? Under the hood, the call to change the password will also validate it here https://github.com/dotnet/aspnetcore/blob/main/src/Identity/Extensions.Core/src/UserManager.cs#L2408

image

image

> Are there any other password rules to follow? e.g. not the same as previous password, etc.

Nope, maybe one day but there's no additional password rules

> Do you need to check the user permissions for current and non-current user controller? (I'm presuming not since they're already logged in)

nope all good, the current user controller already validates the current user who can change their own password and it will only change the user with their ID based on their auth ticket and the user controller is authed by policies.

Shazwazza:
@emmagarland I'll merge this in 🚀  but please have a look at the additional commit/changes I made to see if I mucked something up or if you have questions about it. Cheers!

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/MemberRepository.cs
```diff
@@ -314,6 +314,7 @@ protected override void PersistNewItem(IMember entity)
             // persist the member dto
             dto.NodeId = nodeDto.NodeId;
 
+            // TODO: password parts of this file need updating
```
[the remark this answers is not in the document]
emmagarland: @Shazwazza I saw the password functionality on line 323 and just wanted to check it was still right based on our newer password work

emmagarland:
> @emmagarland I'll merge this in 🚀 but please have a look at the additional commit/changes I made to see if I mucked something up or if you have questions about it. Cheers!

Thanks @Shazwazza! From what I can see, adding the Scope and the locking, `needSync`, it makes sense to me

PR #9906, opened 2021-02-26
Add allowedChildren call to EditorModelEventManager

patrickdemooij9:
This is something I've been wanting to add to Umbraco for some time now. The need arises because we have some node types that should only be used once in a project however, there is no way to restrict that right now. So I added the GetAllowedChildren response to the EditorModelEventManager so that we are able to have some more control over it. I've added it for Content and for Media items.

Here are two use cases together with the user composer so that you can test this:
RestrictedSettings
```
public class OneSettingsComposer : IUserComposer
    {
        {
            EditorModelEventManager.SendingAllowedContentChildrenModel += EditorModelEventManager_SendingAllowedContentChildrenModel;
        }

        private void EditorModelEventManager_SendingAllowedContentChildrenModel(System.Web.Http.Filters.HttpActionExecutedContext sender, EditorModelEventArgs<IEnumerable<Models.ContentEditing.ContentTypeBasic>> e)
        {
            if (e.Model.All(it => it.Alias != "settings"))
                return;

            var currentSettings = e.UmbracoContext.Content.GetSingleByXPath("//settings");
            if (currentSettings is null)
                return;

            e.Model = e.Model.Where(it => it.Alias != "settings");
        }
    }
```

RestrictedPeople
```
public class LimitPeopleComposer : IUserComposer
    {
        {
            EditorModelEventManager.SendingAllowedContentChildrenModel += EditorModelEventManager_SendingAllowedContentChildrenModel;
        }

        private void EditorModelEventManager_SendingAllowedContentChildrenModel(System.Web.Http.Filters.HttpActionExecutedContext sender, EditorModelEventArgs<IEnumerable<Models.ContentEditing.ContentTypeBasic>> e)
        {
            var queryDictionary = HttpUtility.ParseQueryString(sender.Request.RequestUri.Query);
            if (!int.TryParse(queryDictionary["contentId"], out var contentId))
            {
                return;
            }

            var content = e.UmbracoContext.Content.GetById(contentId);
            if (content is null)
                return;
            if (content.Children.Count() >= 6)
            {
                e.Model = e.Model.Where(it => it.Alias != "person");
            }
        }
    }
```

### DIFF
--- a/src/Umbraco.Web/Editors/ContentTypeController.cs
+++ b/src/Umbraco.Web/Editors/ContentTypeController.cs
@@ -413,6 +413,7 @@ namespace Umbraco.Web.Editors
         /// </summary>
         /// <param name="contentId"></param>
         [UmbracoTreeAuthorize(Constants.Trees.DocumentTypes, Constants.Trees.Content)]
+        [OutgoingEditorModelEvent]
         public IEnumerable<ContentTypeBasic> GetAllowedChildren(int contentId)
         {
             if (contentId == Constants.System.RecycleBinContent)
--- a/src/Umbraco.Web/Editors/EditorModelEventManager.cs
+++ b/src/Umbraco.Web/Editors/EditorModelEventManager.cs
@@ -23,6 +23,8 @@ namespace Umbraco.Web.Editors
         public static event TypedEventHandler<HttpActionExecutedContext, EditorModelEventArgs<IEnumerable<Tab<IDashboard>>>> SendingDashboardModel;
         public static event TypedEventHandler<HttpActionExecutedContext, EditorModelEventArgs<IEnumerable<Tab<IDashboardSlim>>>> SendingDashboardSlimModel;
 
+        public static event TypedEventHandler<HttpActionExecutedContext, EditorModelEventArgs<IEnumerable<ContentTypeBasic>>> SendingAllowedContentChildrenModel;
+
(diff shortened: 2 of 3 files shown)


marcemarc:
Is this one related? https://github.com/umbraco/Umbraco-CMS/pull/9595

patrickdemooij9:
No, didn't know there was another pull request out there about this. I'll let HQ decide which one to accept, as I think accepting both is a bit weird

Shazwazza:
Side note - it would be possible to add a webapi global filter to your application on startup and handle it that way. You could then handle any response's object model and tweak it if you wanted. This would be more flexible in the long run instead of adding a new event for every web api response.

patrickdemooij9:
Interesting didn't think about doing it that way. I do think that something like an event gives users an easier way of connecting to it. However, I do agree that we don't need these events for every web API response.

And a question for the reviewer of the PR. Currently, both the content and media go to the same event handler. Should that be seperarated into two different eventhandlers?

patrickdemooij9:
@sebastiandammark, I think this could eventually be combined with a UI if someone sees the need to. This PR mostly allows the user to write their own logic there. My examples were about limiting the number of allowed items. However, you could also write another implementation of it. For example:

- Only users within a certain group are allowed to add certain items
- You don't want to allow editors to create new items in the back-office, only in the front-end.

You can change it any way you want, this was just an example using the code changes that I made

PragmaticIT:
Looks like a great idea to me. 
Is there a plan to merge this code with Umbraco? Seems to be a pretty simple yet elegant and powerful addition

PR #9909, opened 2021-02-28
Paging support for media picker #9766

### LINKED ISSUE #9766: Media picker slow to load with a large number of images at the root / in a folder
When you have a large number of images at any level / folder the media picker is very slow to load.
A fix for this would be to add paging to the media picker, similar to the list view.

## Umbraco version

I am seeing this issue on Umbraco version: 8.10.1

Reproduction
------------

### Bug summary

Media picker slow to load items when there are a large number of images (example 1000 at the root)

### Steps to reproduce

1. Create 1000 images at the root
2. Use the media picker on a document type.

### Expected result

Media picker loads fast, paging possible

### Actual result

Media Picker loads slowly, paging not possible

nzdev:
### Prerequisites

If there's an existing issue for this PR then this fixes #9766

### Description
What?
Support for paging in the media picker. 

Why?
Existing media picker code fetches all children. If you have a large number of nodes in a folder / at the media root this will be very slow.

How:
Switch from using getChildren to getPagedChildren. 
Lower pagesize to 20.

How to test?

Put more than 20 media nodes at any level. The paging controls should appear and page as expected when using the media picker.


src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js
```diff
@@ -500,23 +500,27 @@ angular.module("umbraco")
...
-                            data[i].image = mediaHelper.resolveFileFromEntity(data[i], false);
-                        }
-                        if (data[i].metaData.UpdateDate !== null){
-                            data[i].updateDate = data[i].metaData.UpdateDate;
-                        }
-                        data[i].filtered = allowedTypes && allowedTypes.indexOf(data[i].metaData.ContentTypeAlias) < 0;
+                    // update image data to work with image grid
+                    if (data.items) {
+                        data.items.forEach(mediaItem => setMediaMetaData(mediaItem));
+                        data.items.forEach(mediaItem => mediaItem.filtered = allowedTypes && allowedTypes.indexOf(mediaItem.metaData.ContentTypeAlias) < 0);
```
bjarnef, line 510: Why do we need to loop items twice?
Can't we do the following?

```
data.items.forEach(mediaItem => {
     setMediaMetaData(mediaItem);

    mediaItem.filtered = allowedTypes && allowedTypes.indexOf(mediaItem.metaData.ContentTypeAlias) < 0;
});
                    
```

nzdev: Yes

mikecp:
Hello @nzdev ,

Thanks a lot for this PR, it will be a great improvement!

However, I noticed some unexpected behaviors during my review and I thought it might be worth checking them prior to merging.

The first one is that, if you change the toggle of the checkbox "Search only this folder", the paging disappears after the list is refreshed, and you seem to be presented with the full list again.

Example, initial view, displayed as expected:

image

If I toggle the checkbox:

image

The second point is that, when selecting media from the Grid, the paging seems to "forget" in which folder it is looking. Example below: I want to insert an image in the grid, I selected Media\People. I navigate to the second page, and when I come back to page 1, I see the following, where it displays pictures from products, and I even see folders in there.

Page 1 OK:

image

Page 2 more or less OK (I see more than just the 21st image):

image

Back to Page1 not OK:

image

Could you maybe have a look and see if something can be done to solve those issues (or maybe it will work and then it means something is wrong with my setup)? It might also be due to some default behavior of the paging, but then my first impression is that it might be kind of confusing for the editors.

Thanks for your time!

marcemarc:
@mikecp just reading this through - made me wonder - does the paging work with the 'List View' toggle option of the Media Picker? , when you hit this toggle option in the Media Picker, you can sort the Media items by Last Edited date and by Name...

... with this PR - would this now only sort the items currently loaded by the picker on the current page of the paging?

eg, if you chose to sort by Last Edited to find the most recently uploaded Media Item in a folder, would it only sort those items on 'Page 1'?

If so we may need to implement the sorting options into the serverside paging request here.

bjarnef:
@marcemarc it probably need to have `orderBy` and `orderDirection` similar to other paged results to be able to sort the paged results, e.g. in `entityResource`:
https://github.com/umbraco/Umbraco-CMS/blob/34e80d86e8c0b754f6b7a02e307f53cb32806bbe/src/Umbraco.Web.UI.Client/src/common/resources/entity.resource.js#L434-L435

mikecp:
@marcemarc Good points ;-)
I just checked and the sorting currently occurs only within the "displayed page" indeed. Also, when changing page, the selected sorting option goes away and is reset to default.

nzdev:
It may be a bit of time before I can address the feedback. If anyone wants to continue coding from here I won't be offended. 😀

hamishrouse:
Hi All, thought I would jump in and help

As part of my pull request:
- Fixed pagination so that they will default to page 1 when switching from folder browsing to search filtering results
- Fixed the issue with checking and unchecking `Search only this folder`
- Fixed the issue the wrong results from other folders - mentioned by @mikecp 
- Changed the paging size for both Search and Browsing to 20 items

Remaining issue/s:
- Ordering still needs to addressed - any further advice will be great

hamishrouse:
Tackling the sort ordering has been harder than expected.

Issues encountered: 
When I change the layout to list view
And then choose either `Name` or `UpdateDate`

I expect the ordering to update
And show it ordered by either `Name` or `UpdateDate`
And show paged ordering of the whole list

Actual:
The ordering seems to be ordering the first page instead of getting the first page of the ordered list 
e.g. 
* Actual: get first page of results and then order the list
* Expected:  get listed order and show first page of results

Example
`Z Unicorn`* image file name are listed on page 2 of the results and should be either at the bottom of page 2 (when ascending) OR the top of page 1 (when descending)
image

Digging further into the code - the following point:
https://github.com/umbraco/Umbraco-CMS/blob/34e80d86e8c0b754f6b7a02e307f53cb32806bbe/src/Umbraco.Core/Persistence/Repositories/Implement/EntityRepository.cs#L85
may need some further review...

marcemarc:
@hamishrouse also there is a subtlety to the sorting, in that 'folders' are grouped together in the sort, separate from actual files, so you'll get A-Z of folders followed by A-Z of files to mimic a file system... so to support paging, that same logic would need to exist in the serverside, and clicking the sort filter headings for name or update date, would probably need to request the whole data again to apply the paging to the new ordering...whereas currently it's just doing it on the clientside :-(

PR #9916, opened 2021-03-01
Netcore: Fix coupling between IBackOfficeSecurity and IUmbracoContext

bergmania:
Fixes https://dev.azure.com/umbraco/D-Team%20Tracker/_workitems/edit/10710

Now we use `IHttpContextAccessor` to get information about whether the current user is `null` or not. The user is expected to be `null` in background tasks.


src/Umbraco.Web.Common/UmbracoContext/UmbracoContext.cs
```diff
@@ -33,13 +33,13 @@ public class UmbracoContext : DisposableObjectSlim, IDisposeOnRequestEnd, IUmbra
...
         internal UmbracoContext(
             IPublishedSnapshotService publishedSnapshotService,
-            IBackOfficeSecurity backofficeSecurity,
             UmbracoRequestPaths umbracoRequestPaths,
             IHostingEnvironment hostingEnvironment,
             IVariationContextAccessor variationContextAccessor,
             UriUtility uriUtility,
             ICookieManager cookieManager,
-            IRequestAccessor requestAccessor)
+            IRequestAccessor requestAccessor,
```
Shazwazza, line 41: We shouldn't need both the IRequestAccessor and IHttpContextAccessor

Shazwazza (commented):
Change looks good, and I've simplified things further in https://github.com/umbraco/Umbraco-CMS/pull/9926 . I've added a note here but I'll have a look at it and let you know.

PR #9920, opened 2021-03-01
Fixes #9918 - Update Lightinject to v6.4 due to memory leaks in LightInject.Microsoft.DependecyInjection

### LINKED ISSUE #9918: Update Lightinject to v6.4 due to memory leaks in LightInject.Microsoft.DependecyInjection
For the development of one of our project we ran into a memory leak in the package LightInject.Microsoft.DependecyInjection. This is a package we installed so we could add the HttpClientFactory and Polly to the project due to the necessity of resilience in our dependency on external API's. Both of these packages requires a service registration not possible with the Component/Composer pattern in Umbraco.

A fix for this memory leak was introduced in version 3.x of LightInject.Microsoft.DependencyInjection (https://github.com/seesharper/LightInject.Microsoft.DependencyInjection/pull/152) which has a dependecy for LightInject 6.x. Currently Umbraco has a hard dependency on LightInject 5.x and with this request I want to request an update to LightInject 6.4 (latest version) so adding the bugfix in LightInject.Microsoft.DependencyInjection to the project is possible as the older version of LightInject is no longer supported.

deMD:
This PR fixes #9918

#### Description
Updated LightInject to version 6.4.0 and made sure that:
1. All relevant Unit Tests succeed
2. Site still runs when using the updated package

---
_This item has been added to our backlog [AB#10761](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10761)_

### DIFF
--- a/build/NuSpecs/UmbracoCms.Core.nuspec
+++ b/build/NuSpecs/UmbracoCms.Core.nuspec
@@ -23,7 +23,7 @@
                   the latter would pick anything below 3.0.0 and that includes prereleases such as 3.0.0-alpha, and we do
                   not want this to happen as the alpha of the next major is, really, the next major already.
                 -->
-                <dependency id="LightInject" version="[5.4.0,5.999999)" />
+                <dependency id="LightInject" version="[6.4.0,6.999999)" />
                 <dependency id="LightInject.Annotation" version="[1.1.0,1.999999)" />
                 <dependency id="LightInject.Web" version="[2.0.0,2.999999)" />
                 <dependency id="Microsoft.AspNet.Identity.Core" version="[2.2.2,2.999999)" />
--- a/src/Umbraco.Core/Umbraco.Core.csproj
+++ b/src/Umbraco.Core/Umbraco.Core.csproj
@@ -77,7 +77,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="LightInject" Version="5.4.0" />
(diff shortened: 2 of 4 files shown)


nul800sebastiaan:
Thanks @deMD! That looks better! 👍 

I can't figure out why the LightInject version was bumped to a new major version but from looking at the changes I think there are no breaking changes.  https://github.com/seesharper/LightInject/releases/tag/v6.0.0 
Assume that it has something to do with the new "flowing" scope they're describing. I see no interesting issues created since that release: https://github.com/seesharper/LightInject/issues?q=is%3Aissue+created%3A%3E%3D2019-09-28

So I'm assuming it's fine to upgrade this, HQ will need to do a double-check before we merge.

deMD:
Major had been bumped because https://github.com/seesharper/LightInject.Microsoft.DependencyInjection/pull/152 requires LightInject 6.x (not sure which version it was released in exactly but at least a version dependent on LightInject 6.3.5) and there is a memory leak issue fixed there where the HttpClientFactory can't find existing handlers due to the scoping issue causing it to keep creating handlers that are never removed by the GC. As that bug is only fixed in the newer major of that package, to get that fix in custom Umbraco code/packages we'll need an Upgrade of LightInject.

nul800sebastiaan:
Thanks for the help @deMD - this is now merged! 👍

nul800sebastiaan:
We found breaking changes when trying this out on Cloud (we found Forms and Deploy to be broken with this version since they were relying on v5 - getting null ref errors now), which means that many more people will run into breaking changes if they're currently using version 5 in their code. 

I'm going to have to revert this change unfortunately. :-(

Issue #9861, opened 2021-02-21
Members not locked out after failed login attempts


iishnz:
### Bug summary

A front-end member can enter an unlimited amount of incorrect password attempts without being locked out. The number of failed password attempts does not get recorded in the member property in the Backoffice.
There is also an issue the the last login date in that it can revert to an older date but I think this is caused by the same bug.

## Umbraco version

This is not working in 8.11.1 - it was previously working in 8.6.4

### Reproduction

1. On a fresh install of Umbraco with built-in with the starter kit

2. Create the following partial views using the built-in snippets:
	* Login
	* Login Status

3. Edit the Home template and added the following somewhere near the top:

```
<section class="section">
@Html.Partial("~/Views/Partials/Login.cshtml")
@Html.Partial("~/Views/Partials/Login Status.cshtml")
</section>
```

4. Create a member in the back-office to test with.

5. On the homepage, login with your test member using an invalid password

6. In the back-office, go to the members section and review the "Failed Password Attempts" property for your test member, it will probably say "1"

7. On the homepage, try and log in again with your test member using an invalid password, then check the back-office again, the "Failed Password Attempts" did not change.

8. By default a member should be locked out after 5 invalid attempts. Try logging in 20 times with and invalid password, then try with the correct password - The member is logged in.

### Expected result

Failed login attempts should be record, A member should be locked after a certain amount of failed password attempts.

### Actual result

The member was not locked out and the number of failed login attempts was not recorded.

---
_This item has been added to our backlog [AB#10589](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10589)_

reneemhaas-diagram:
Adding that this is also not working in 8.9.2.

bergmania:
I can reproduce, and my finding so far is that it is a caching issue.

bergmania:
@nul800sebastiaan, The bug was introduced in [8.6.5](https://github.com/umbraco/Umbraco-CMS/compare/release-8.6.4...release-8.6.5)

nul800sebastiaan:
Fixed in PR https://github.com/umbraco/Umbraco-CMS/pull/9921 and cherry picked for patch releases from 8.6 upwards.

PR #9921, opened 2021-03-02
Bugfix: Members not locked out after failed login attempts

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/9861

# Details
Members were not locked out even if they entered the wrong password more then the maximum allowed times.  The issue was due to caching. So the database was actually correct. And it worked after a reboot.

This PR fixes the issue by raising the events when we save the info about `FailedPasswordAttempts`. Thereby the caches are refreshed as expected. 

Even that it is a heavy operation to save the member with event raised because e.g. Examine indexes are updated, we consider that okay, because successful logins are still cheap.

Also added a test that shows this error, even that it required lots of plumbing 😢 

# Test
- Follow reproduction steps on https://github.com/umbraco/Umbraco-CMS/issues/9861


elit0451:
Test + changes look good! The issue's been eliminated => merging 👍

PR #9922, opened 2021-03-02
Replace static backoffice user manager events with IEventAggregator notifications 

### LINKED ISSUE #9397: Netcore: Remove static events

kjac:
### Prerequisites

### Description

This PR replaces the static events in the backoffice user manager with notifications via `IEventAggregator`.

Special points of interest:

1. The notification methods that are necessary to expose for use outside of `BackOfficeUserManager` (e.g. `NotifyForgotPasswordRequested`) have been moved from `IUmbracoUserManager<TUser>` to `IBackOfficeUserManager`, so `MemberManager` does not have to implement those methods (all of them are currently implemented as `throw new NotImplementedException()`).
2. The option create a custom "sending user invite email" implementation (currently `RaiseSendingUserInvite`) has been removed from `BackOfficeUserManager`. Moving forward you must create your own email sender instead and implement custom email handling there (see #9900)

### Testing this PR

The easiest way to test this PR is to log in, log out and perform failed logins in the backoffice. The user auditing in the `umbracoAudit` table will contain records of the operations performed, and these records are inserted by `BackOfficeUserManagerAuditer` using the `IEventAggregator` notifications.

Grab the audit records with this:

```sql
SELECT TOP 100 * FROM umbracoAudit ORDER BY id DESC
```


src/Umbraco.Web.BackOffice/Security/BackOfficeUserManagerAuditer.cs
```diff
@@ -1,82 +1,55 @@
...
+        INotificationHandler<UserPasswordChangedNotification>,
+        INotificationHandler<UserPasswordResetNotification>
     {
         private readonly IAuditService _auditService;
         private readonly IUserService _userService;
-        private readonly GlobalSettings _globalSettings;
-        private bool _disposedValue;
 
-        public BackOfficeUserManagerAuditer(IAuditService auditService, IUserService userService, IOptions<GlobalSettings> globalSettings)
+        public BackOfficeUserManagerAuditer(IAuditService auditService, IUserService userService/*, GlobalSettings globalSettings*/)
```
bergmania, line 25:
```suggestion
        public BackOfficeUserManagerAuditer(IAuditService auditService, IUserService userService)
```

bergmania (approved):
Just a small unused code that should be removed.. I'll fix that and merge

bergmania:
Amazing work @kjac, thanks 🚀

PR #9926, opened 2021-03-03
Fixes ExecutionContext flowing to child threads (Ambient Scope issues)

Shazwazza:
This fixes quite a lot of things. Primarily this was to fix the "not ambient scope" exception we were seeing every time we saved a content type which we presumed was an issue with AsyncLocal ... and it was. Well, it's not an issue with AsyncLocal, it's an issue with understanding it and how it flows to child threads. We have this same issue in v8 but it's dealt with different because we don't use AsyncLocal there but instead use it's older cousin: CallContext. Both concepts 'flow' to child threads but in slightly different ways. 

## Changes for AsyncLocal and ExecutionContext

* Remove `SafeCallContext`. This was a work around for using the `CallContext` which we don't use anymore. This work around was required due to the craziness of serialization of whatever is in the CallContext between app domains. Stephane wrote some stuff about that here https://www.zpqrtbnk.net/posts/putting-things-in-contexts/ . AsyncLocal doesn't suffer from the same serialization problem that CallContext suffers from because there is nothing that forcefully tries to serialize things in AsyncLocal like it does with CallContext (i.e. it was built for remoting and that's what remoting does). SafeCallContext was a way to prevent the flow of the ExecutionContext to child threads but did it in a unique way. The correct way to prevent the flow is built directly into .NET with `ExecutionContext.SuppressFlow()`. This also works for CallContext but due to the serialization issues many hurdles needed to be jumped to make it work and that's how we ended up with `SafeCallContext`. In netcore - we just use `ExecutionContext.SupressFlow()`. NOTE: We still have a class called `CallContext<T>` that was ported from v8, but this wraps AsyncLocal now.
* TaskHelper has been updated to work correctly and to properly deal with fire and forget tasks which require **not** flowing the ExecutionContext. 
* Ensure `ExecutionContext.SuppressFlow()` is used where it's needed. This needs to happen when we know there will be a child thread created that is not going to be joined to the currently running thread:
  * `UmbracoExamineIndex` - `PerformDeleteFromIndex` (which previously did this in v8 with SafeCallContext) and `PerformIndexItems` (which needs to be back ported to v8!). Both of these methods call into Examine which will have fire and forget tasks that do not join. This means I need to fix Examine directly so this isn't a concern to the caller ... and then we can remove this logic in Umbraco. 
  * SqlMainDomLock - We execute a long running task here. This is just a safety check since at that state there should be nothing of concequence being flowed but its better to be safe since we know we don't want the context flowed.
* Fixes debugging when the DEBUG_SCOPES flag is set. To enable this, this flag can be set in both Umbraco.Infrastructure.csproj and Umbraco.Tests.UnitTests.csproj by adding this:
   ```xml
    <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
       <DefineConstants>$(DefineConstants);DEBUG_SCOPES</DefineConstants>
    </PropertyGroup>
   ```
* Much better detailed exception messages for Scopes when there are leak issues
* Updates lots of tests - previously in some tests, SafeCallContext was used to 'fake' suppressing the execution flow but in reality this would never happen so the tests where that occurred now uses the correct approach of Task.Run along with ExecutionContext.SuppressFlow() or by using TaskHelper (which takes care of this for us). Where SafeCallContext was used along with threading, this is just replaced with ExecutionContext.SuppressFlow(). 
* Had to update to latest NUnit because it suffered from similar issues where the ExecutionContext was flowing between tests: https://github.com/nunit/nunit/issues/3283
* ScopeProvider is changed in the way that it works with 'Ambient' objects. This is the larger change and some design decisions should be discussed **(more below)**. 

... There's a few things that will need to be backported to v8 which may be causing some underlying bizarre issues.

## ScopeProvider design

_I'm writing this out here for information/documentation/discussion purposes_

What is an 'Ambient' Scope/ScopeContext? We need to consider 2 concepts here: Child Scopes and Child Threads. 

### The general rule

In the simplest terms we don't consider Child Threads, in which case the Ambient Scope/Context is the most recently created, or 'deepest' instance of the Scope hierarchy. For example, when the first outer-most scope is created (ScopeA), it will be Ambient. If another Scope (ScopeB) is created with the ScopeProvider while ScopeA is still Ambient, then ScopeB is now Ambient, and so on. The rule is that the Ambient (deepest) Scope **must** be dispose before it's parent. 

### Threading

This gets a lot more complicated when you start including Child Threads. Child Threads will naturally occur with any async code and in 98% of the cases it will work exactly the same way listed above because in most cases async code is awaited and joined with the current thread. As an example, we'll keep the ScopeA and ScopeB hierarchy but we'll assume that ScopeA is created in ThreadA and calls and awaits an async method which creates ScopeB on ThreadB. How do we 'flow' ScopeA to ThreadB? In the case of an HTTP request, this is managed by the HttpContext (in both v8 and netcore). In other cases, we use AsyncLocal (in v8, we used CallContext). In aspnetcore, the HttpContext is actually managed by AsyncLocal so there isn't really any magic happening between an HTTP request or not, it's the same thing. In v8, HttpContext does it's own custom management of flowing contexts. But let's focus on AsyncLocal (netcore) and CallContext (v8).

### AsyncLocal

* When ScopeA on ThreadA is created, it is the AmbientContext set on AsyncLocal.
* When ScopeB on ThreadB is created, it is the AmbientContext set on AsyncLocal.
* At this snapshot in time:
  * ThreadA: AsyncLocal == ScopeA
  * ThreadB: AsyncLocal == ScopeB

This is how AsyncLocal works, it creates a copy at each child thread. Setting AsyncLocal.Value at a child thread does **not** affect the value at the parent thread. This is by design is is very effective in most cases. But this is problematic in our case because ThreadA does not actually know the true AmbientContext. 

The recommended practice in using AsyncLocal is to not use mutable objects because it can lead to a lot of accidental headaches and confusion, but in order to achieve what we want, we need to have a mutable object for AsyncLocal.Value so that the values within that object, when changed, are reflected on all threads in the hierarchy. Based on the rules listed above, that child Scopes must be disposed before parent Scopes, that rule can be achieved by using a Stack where the top of the stack must be disposed and removed first. So now the ScopeProvider does exactly that and keeps a ConcurrentStack<Scope> as the AsyncLocal value. Each new scope is added to the stack and the top of the stack is the 'Ambient' Scope and can be seen across the thread hierarchy.

### CallContext

In v8 this behavior is quite different and is due to quite a lot of things like dealing with serialization issues. V8 didn't use any Stack, instead due to the weird behavior of CallContext, it maintained a static `Dictionary<Guid, object>` to track scope instances based on their corresponding IDs stored in CallContext. In this case, the CallContext value just stored an immutable Id (GUID) which would flow to all child threads but when a child thread creates a new scope, this ID value is replaced but only for that thread because CallContext (and AsyncLocal) are copies and writes that occur downstream don't affect values upstream. This meant that the Ambient context returned from CallContext would be different for different threads that were in the same execution context hierarchy and each thread would see it's own scope as the Ambient (even though it wasn't the 'true' ambient).

This is why a Stack is much more robust in tracking the current Scope instance and enforcing the rule above instead of just a single value representing the 'top'.

### Future considerations

There's some things in the code that are tested but exist primarily for Deploy: Detatched Scopes and moving Scopes between contexts (i.e. Http -> AsyncLocal and vice versa). We may need to revisit some of this when Deploy is ported to netcore to see if this is working as it should. Also you'll see notes in the ScopeProvider for when it tries to move items between context like Http -> AsyncLocal and vice versa. Due to the way that AsyncLocal works, this won't 100% work if the 'Move' is occurring from HTTP -> AsyncLocal on a child thread because the Value of AsyncLocal will not reflect back to the parent thread. I'm unsure if this is going to matter since I think that this "Move" operation is an edge case and it's only used for Deploy.

Parallel threads is also a conundrum. There's tests in v8 that are passing but this is sort of by fluke. There is a test in both v8 and netcore that tests this: `ThreadSafetyServiceTest`. This passes here now but only because I suppressed the execution context from flowing from the main thread which means that each thread has it's own 100% isolated Scope hierarchy. In v8 this isn't the case but it still passes... why is that? ... that answer is complicated and confusing :) The reason it works is because of what is documented above under CallContext: each thread sees the Ambient context it wants to see, not the true one. So while the test passes, there's probably some strange side affects happening too.

So how do we proceed with netcore? If people try to use parallel threads it will explode without supressing execution context but that is not very clear. We don't want people to use parallel threads with ambient scopes since that really shouldn't be supported ([it's not with EFCore either ](https://stackoverflow.com/questions/44063832/what-is-the-best-practice-in-ef-core-for-using-parallel-async-calls-with-an-inje). What could be done is:

* We keep the mutable Stack in AsyncLocal like we have it. We know this allows us to track the 'true' ambient between threads.
* We add another AsyncLocal value that just tracks the Scope for the current execution context. This would allow a Scope to know which ambient context was created in its own execution context AND the 'true' ambient context that exists across threads. This could provide us with additional information for raising the exception in the Scope.Dispose. 

## IRequestCache dead locks

We had a deadlock scenario in the IRequestCache because it was working differently to v8. In v8 the HttpContext.Items is an IDictionary which has a SyncLock property which allows being able to lock on an object intended to be locked on. In netcore, no part of the HttpContext is threadsafe and the HttpContext.Items is a different type of dictionary that doesn't expose such a property. So during the transition to netcore this lock was changed to use the HttpContext.Items dictionary itself as a lock. This leads to deadlocks. The HttpContext can be accessed by more than one thread at a time depending on some async logic and if folds are using Task.Run within a request, so our IRequestCache needs to be thread safe. This shows one deadlocks scenario I ran into:

image

Several things required to deal with this:

* Removes `GenericDictionaryRequestAppCache`
* Remove HybridBackofficeSecurityAccessor.cs - this should never be hybrid. The IBackOfficeSecurity should only ever exist in a request. 
* Remove IBackOfficeSecurityFactory - we don't need a factory, BackOfficeSecurity can be managed by DI so now it is and IBackOfficeSecurityAccessor just returns it from the httpcontext (if there is one) request services. This fixes the TODOs of the dependency chain being broken.
* Remove HybridUmbracoWebsiteSecurityAccessor.cs - this should never be hybrid and should only ever exist in a request. This is now just managed with DI and exposed on an accessor that returns the instance from the httpcontext (if there is one) request services.
* Changes HttpRequestAppCache -> HttpContextRequestAppCache to be in a netcore project and just use HttpContext instead of trying to wrap it's dictionary in a different factory. This then allows us to use HttpContext request features to set an object to be used as a SyncRoot locker. We lock on the httpContext itself when setting this but the features collection is captured first so no deadlocks can occur. We use a simple lock to do a Monitor lock and not a ReaderWriterLockSlim because the ReaderWriterLockSlim must be disposed and the aspnet request pipeline that disposes http context items at the end of the request occurs too soon and we need this lock during disposal (this is also why this lock isn't managed as a Scoped DI item).

## Other

* Remove IDisposeOnRequestEnd - we don't need this, it's an old way of working with request scoped things
* ScopeReference changed to HttpScopeReference. This is/was used to ensure that any orphaned Scopes are always reset and disposed at the end of a request. For example, if a new scope was created during an http request but was never disposed, this would ensure that would happen. This now works simply by being a Scoped DI service and its lifetime is taken care of by DI instead of trying to managed this via the ScopeProvider. This is enabled now by our request middleware which opts-in for it to be cleaned up per request.
* The IDisposeOnRequestEnd was broken and nothing was being disposed at the end of the request because this used to be based directly on HttpContext items but now it is based on IRequestCache and since IRequestCache deals with Lazy<object> values, nothing was being detected as IDisposeOnRequestEnd. This has been fixed and now IRequestCache is responsible for disposing the current items since it knows how to deal with them.
* Updates patch versions of packages and had to reconfigure certain package versions in order for this to build. For whatever reason, Umbraco.Web cannot reference Umbraco.Code now due to some CSharp references. And we cannot update to `Microsoft.CodeAnalysis.CSharp -Version 3.9.0` from `Microsoft.CodeAnalysis.CSharp -Version 3.8.0` since this seems to cause weird references problems again.
* Change UmbracoRequestMiddleware to always log (trace) the request start/end, as it was the `isFrontEndRequest` would 100% be false because at that stage the front-end routing hasn't even happened yet
* Ensure things using ReaderWriterLockSlim are IDisposable and that the lock is disposed - TODO: we still need to do this  in more places and also back port that to v8

## Testing

* Ensure unit tests pass
* Start the app, make sure things are working, save and publish some content and then update/save content types - we should no longer have the 'not ambient scope' issue.


bergmania:
When creating doc types, looks like something is wrong with the translations
image

bergmania:
I didn't find anything unexpected in my test 💪

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilderExtensions.cs
```diff
@@ -86,15 +87,18 @@ public static class UmbracoBuilderExtensions
...
 
-            var loggingDir = tempHostingEnvironment.MapPathContentRoot(Cms.Core.Constants.SystemDirectories.LogFiles);
+            var loggingDir = tempHostingEnvironment.MapPathContentRoot(Constants.SystemDirectories.LogFiles);
             var loggingConfig = new LoggingConfiguration(loggingDir);
 
             services.AddLogger(tempHostingEnvironment, loggingConfig, config);
 
+            // TODO: This doesn't seem right? The HttpContextAccessor is normally added to the container
+            // with ASP.NET Core's own ext methods. Is there a chance we can end up with a different
+            // accessor registered and resolved?
```
bergmania, line 97: Seems like we are also doint this in `AddWebComponents` At least, we should not register it twice

src/Umbraco.Web.Common/Cache/HttpContextRequestAppCache.cs
```diff
@@ -0,0 +1,290 @@
...
+    /// </summary>
+    /// <remarks>
+    /// <para>The HttpContext is not thread safe and no part of it is which means we need to include our own thread
+    /// safety mechanisms. This relies on notifications: <see cref="UmbracoRequestBegin"/> and <see cref="UmbracoRequestEnd"/>
+    /// in order to facilitate the correct locking and releasing allocations.
+    /// </para>
+    /// </remarks>
+    public class HttpContextRequestAppCache : FastDictionaryAppCacheBase, IRequestCache, IDisposable
+    {
+        //private static readonly string s_contextItemsLockKey = $"{typeof(HttpContextRequestAppCache).FullName}::LockEntered";
```
bergmania, line 26:
```suggestion

```

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilderExtensions.cs
Shazwazza: Yeah, that's gonna end up causing unwanted behavior IMO. I'll have a look

Shazwazza: I guess it won't cause problems because that's all based on AsyncLocal access, see https://github.com/dotnet/aspnetcore/blob/main/src/Http/Http/src/HttpContextAccessor.cs ... but it will mean there's different HttpContextAccessor instances.  I'll just remove the 2nd one.

bergmania (approved):

PR #9930, opened 2021-03-03
Automated install user with Environment Variables & unattended.user.json 

warrenbuckley:
This is @nikolajlauridsen work on Unattended installs from his cycle work, so credit for him for this.

I have ever so slightly modified this so that we can change the default/super user aka -1 as opposed to creating a brand new user, in the same approach as it is done in `NewInstallStep.cs`

>Going to try to rework this in the same approach that the NETCore approach #9934 was done with Bjarke. That an event is raised when Unattended Install completes.

## Environment Variables
Variable Name | Example
-----------------|----------
unattendedName | Warren Buckley
unattendedEmail | warren@umbraco.com
unattendedPassword | password1234

## JSON file
The values can be read from a JSON file at the following location `/App_Data/unattended.user.json`
After the unattended install the JSON file will be deleted.

Example
```json
{
    "name": "Warren via JSON",
    "email": "warren@umbraco.com",
    "password": "password12345"
}
```


warrenbuckley:
Going to try to rework this in the same approach that the NETCore approach #9934 was done with Bjarke. That an event is raised when Unattended Install completes.

src/Umbraco.Core/Runtime/CoreRuntime.cs
```diff
@@ -465,5 +572,23 @@ protected virtual Configs GetConfigs()
...
 
         #endregion
+
+        /// <summary>
+        /// Event to be used to notify when the Unattended Install has finished
+        /// </summary>
+        public static event TypedEventHandler<IRuntime, UnattendedInstallEventArgs> UnattendedInstalled;
+
+        [DataContract]
+        public class UnattenedUserConfig
```
marcemarc, line 582:
```suggestion
        public class UnattendedUserConfig
```

src/Umbraco.Core/Runtime/CoreRuntime.cs
```diff
@@ -242,6 +256,93 @@ protected virtual IFactory Boot(IRegister register, DisposableTimer timer)
...
+                if (fileExists == false)
+                {
+                    return;
+                }
+
+                // Attempt to deserialize JSON
+                try
+                {
+                    var fileContents = File.ReadAllText(filePath);
+                    var credentials = JsonConvert.DeserializeObject<UnattenedUserConfig>(fileContents);
```
marcemarc, line 284:
```suggestion
                    var credentials = JsonConvert.DeserializeObject<UnattendedUserConfig>(fileContents);
```

warrenbuckley:
@nul800sebastiaan & @emma-hq would it be possible for you or your lovely community team to review & test this at all as not sure @bergmania & gang has time to review this sprint?

emma-hq (approved):
Lovely stuff ☺

Issue #9722, opened 2021-01-26
Scheduled publish does not run after a document is copied


sean-p-m:
If you copy a document that has been scheduled to publish at a particular time, the scheduled publish information is copied to the new document as you might expect. Scheduled publish information is shown on both the original and new documents in the Umbraco backoffice, as you might expect. However the scheduled publish runs on the new document, but not the original document.

I am seeing this issue on Umbraco version: 8.10.1

Steps to reproduce
------------

1. Create a document in the Umbraco backoffice and set a schedule for it to publish at a particular time.
2. Copy the new document.
3. Verify that the schedule now exists on both the new and old documents by clicking Save and Publish / Schedule...
4. After the scheduled publish observe that the publish has run on the new document but not the original document

Expected result
------------

When a document is copied I would expect detail of a scheduled publish to be preserved on the original document.

Actual result
------------

The schedule publish does not run on the original document.

emma-hq:
Hey @sean-p-m 

Thanks for reporting this. I've reproduced this behaviour in the latest release, 8.11.1, and is an issue at least as far back as 8.7.2 so it's defo something that'll need looking at. I'll label as up-for-grabs.

Emma

umbrabot:
Hi @sean-p-m,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

sean-p-m:
Hi @emma-hq,
I think I might have a go at fixing this one. I haven't contributed before. I've found this to read through:
https://github.com/umbraco/Umbraco-CMS/blob/0bd4dced0b3e9205660114406b7e814f817179c7/.github/CONTRIBUTING.md
Do I need to assign the bug to myself or anything like that?
Thanks,
Sean.

emma-hq:
Hey @sean-p-m - wonderful! That md is a great place to start! We're not able to assign to community members, or to let you assign it, so we usually suggest asking people just to call dibs on the issue itself, as you have done ;)

[...]

PR #9932, opened 2021-03-03
Scheduled publish does not run after a document is copied

sean-p-m:
Scheduled publish does not run after a document is copied
https://github.com/umbraco/Umbraco-CMS/issues/9722

#### What did you do?

I updated the **DeepClone** method on the **ContentSchedule** model to not include the row id in the clone. I also updated the comment on **ContentSchedule.Id** which refers to the property as the unique identifier of the scheduled document when it is actually the unique identifier of the scheduled action.

#### Why did you do it?

To prevent umbracoContentSchedule information being overwritten upon copying, when actually a new row should be inserted into umbracoContentSchedule.

#### How can the changes be tested?

1. Unpublish a live document
2. Add a schedule to publish the document
3. Copy the document

Previously information in umbracoContentSchedule was overwritten with the details of the new document. Now a new record is created in umbracoContentSchedule.

N.B. prior to the fix, when checking the original document schedule through the UI by clicking the **Schedule...** button it can appear that the schedule information is still intact when that is not the case. Check the db to verify the original schedule information is missing. May need to restart the web app before is disappears from the UI.

### DIFF
--- a/src/Umbraco.Core/Models/ContentSchedule.cs
+++ b/src/Umbraco.Core/Models/ContentSchedule.cs
@@ -33,7 +33,7 @@ namespace Umbraco.Core.Models
         }
 
         /// <summary>
-        /// Gets the unique identifier of the document targeted by the scheduled action.
+        /// Gets the unique identifier of the scheduled action.
         /// </summary>
         [DataMember]
         public Guid Id { get; internal set; }
@@ -71,7 +71,7 @@ namespace Umbraco.Core.Models
 
         public object DeepClone()
         {
-            return new ContentSchedule(Id, Culture, Date, Action);
+            return new ContentSchedule(default, Culture, Date, Action);
         }
     }
 }
--- a/src/Umbraco.Core/Services/Implement/ContentService.cs
+++ b/src/Umbraco.Core/Services/Implement/ContentService.cs
@@ -2175,6 +2175,9 @@ namespace Umbraco.Core.Services.Implement
                 copy.CreatorId = userId;
                 copy.WriterId = userId;
 
+                // ensure any content schedule is not copied
+                copy.ContentSchedule = new ContentScheduleCollection();
+
                 //get the current permissions, if there are any explicit ones they need to be copied
                 var currentPermissions = GetPermissions(content);
                 currentPermissions.RemoveWhere(p => p.IsDefaultPermissions);


sean-p-m:
Hi @emma-hq - just wondering if there is anything I need to do to progress this?
Thanks,
Sean.

emma-hq:
Hey @sean-p-m,

I'll give the CC team a nudge and see if anyone is available to take a look. It's fab when someone reports an issue and then fixes it 😁  

Emma

OwainWilliams:
Good morning @sean-p-m 
Sorry this has taken a while to check but I just wanted to give you an update. 
I'm now reviewing the PR and if I have any questions I'll give you a shout. 

Thanks for the contribution!

OwainWilliams:
A couple of question and maybe this is for @nul800sebastiaan or @emma-hq to help with. 

If you copy a node which has been scheduled, should it copy the schedule with it? I would say yes, if it has 'Relate to original' ticked. However, just now, if relate to original is unchecked, the schedule still transfers with the copy. 

Other than that, your fix works as expected and I can confirm the original behavior does overwrite the entry in the database, the fix creates a new entry.

nul800sebastiaan:
@OwainWilliams I wouldn't say in any case nodes should publish themselves, it should be something that is done purposely and manually. 
Of course people are free to look at the "Relate to orignal" relation in `ContentService` events and set the same schedule themselves.

So if this fix currently does not schedule the copied node then we're all good.

OwainWilliams:
Hi @sean-p-m
I've done some testing and I've got some suggestion for the PR which would be great if you were able to implement? 

- When a node is copied, the copied node should not inherit the original nodes schedule. The copy should have no schedule set.
- This is a bonus item but perhaps we can update the icon in the content tree to indicate that a specific node has a schedule set?

sean-p-m:
Hi @OwainWilliams - apologies for the delay in getting back to you - I have been tied up on some project work! 

I've now updated the PR with a change for the first bullet (ensuring that a copied node does not inherit the original's schedule).

For the second bonus request I would argue that should be a separate ticket as it has the potential to be much larger and much more encompassing than the original bug.

Let me know if this is okay.

Cheers,
Sean.

sean-p-m:
Hi @emma-hq @OwainWilliams @nul800sebastiaan - just following up to see if there is any news on this one?

Thanks,
Sean.

Issue #9886, opened 2021-02-23
Drag Upload File not working


nadiamac:
After upgrading from 8.6.4 to the latest version 8.11.1, the ability to drag and upload a file in the media section no longer working.

note: The file upload still works when you click and select on it. Also the drag and drop on the folder level still works

## Umbraco version

I am seeing this issue on Umbraco version: 8.11.1 

Reproduction
------------

### Steps to reproduce

1. Go to media section
2. Select Create new File
3. In file explorer, pick file to upload and drag to the upload area

### Expected result

File is loaded into box, same as when you click on upload and select the file

### Actual result
Nothing happens

---
_This item has been added to our backlog [AB#10606](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10606)_

emma-hq:
Hey @nadiamac 

Thanks for this issue report! I have done a bit of investigating and the last release the drag and drop worked on was 8.7.2 - by 8.0.0, it's not happening at all. 

Emma

bjarnef:
I wonder if the issue described is the same as this one, where it is still possible to upload via "click" though or from dropzone in folder level.
https://github.com/umbraco/Umbraco-CMS/issues/9342

emma-hq:
Hi @bjarnef - yes I think this is related. Thanks for this - it's a really helpful bit of info for someone working with this.

nul800sebastiaan:
This was fixed in PR https://github.com/umbraco/Umbraco-CMS/pull/9933 and a fix will be available in the listed versions. Thanks for reporting! 👍

PR #9933, opened 2021-03-03
#9886 drag n drop files for upload

nielslyngsoe:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/9886

No need to preventDefault or stopPropagation. As this prevents the input from retrieving the dropped-file.

It was introduced in this PR: https://github.com/umbraco/Umbraco-CMS/pull/8390
And those lines didn't serve any purpose towards the goal of that PR. As well it's not used anywhere else, so I conclude it has no purpose and can go away.


madsrasmussen:
Changes look fine. Drag and drop works as expected 🎉

PR #9934, opened 2021-03-03
Unattended Install - Automated User Creation

warrenbuckley:
# Feature
This feature builds upon the unattended install feature and allows the default/super user to be setup with name, email and password with configuration values. Allowing you to skip the installer and go straight to logging into the backoffice.

This works by raising an event/notification once the UnattendedInstall is complete and uses the same logic/approach as the `NewInstallStep.cs` class to modify the default super user as the installer UI would do.

## Test Notes
* Have an empty site
* Create a blank DB
* Update appsettings.development.json or use env variables to set similar config values below
* Verify the Database Schema gets created
* Goto site once its booted and login with credentials you specified in config values
* Also note with this approach without installer - no restart of dotnet run process needed it seems

## Example Settings
This PR adds three new settings that can be configured with the usual .NETCore config approach with appsettings.json files and appsettings.environment.json files and environment variables etc...

### Environment Variables
Variable Name | Example Value
------|------
ConnectionStrings__umbracoDbDSN | server=localhost;database=UmbracoUnicore;user id=sa;password='P@ssw0rd
Umbraco__CMS__Global__InstallUnattended | true
Umbraco__CMS__Global__UnattendedUserName | Warren Buckley Automation
Umbraco__CMS__Global__UnattendedUserEmail | warren@umbraco.com
Umbraco__CMS__Global__UnattendedUserPassword | password1234

### appsettings.development.json
```json
{
  "ConnectionStrings": {
    "umbracoDbDSN": "server=localhost;database=UmbracoUnicore;user id=sa;password='P@ssw0rd'"
  },
  "Umbraco": {
    "CMS": {
      "Global": {
        "InstallUnattended": true,
        "UnattendedUserName": "Warren Buckley Automation",
        "UnattendedUserEmail": "warren@umbraco.com",
        "UnattendedUserPassword": "password1234"        
      },
      "Hosting": {
        "Debug": true
      }
    }
  }
}
```


warrenbuckley:
@bergmania fancy doing a review of this to make this official feature or does it need some rework or.... ?

src/Umbraco.Infrastructure/Runtime/CoreRuntime.cs
```diff
@@ -60,6 +63,8 @@ public class CoreRuntime : IRuntime
         /// <inheritdoc/>
         public async Task StartAsync(CancellationToken cancellationToken)
         {
+            using IServiceScope scope = _serviceScopeFactory.CreateScope();
+
```
bergmania, line 67: These changes in this file can be rolled back, sorry that was my fault..

src/Umbraco.Web.BackOffice/DependencyInjection/UmbracoBuilderExtensions.cs
```diff
@@ -47,7 +47,8 @@ public static class UmbracoBuilderExtensions
                 .AddPreviewSupport()
                 .AddHostedServices()
                 .AddDistributedCache()
-                .AddModelsBuilderDashboard();
+                .AddModelsBuilderDashboard()
+                .AddUnattedInstallCreateUser(); // Put last to test that everything else injected/setup & happy
```
bergmania, line 51: Please remove that comment.  The ordering of composing should not be important, because all composing happen before the actual pipeline boot.

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilderExtensions.cs
```diff
@@ -291,6 +296,13 @@ public static IUmbracoBuilder AddWebComponents(this IUmbracoBuilder builder)
             return builder;
         }
 
+        public static IUmbracoBuilder AddUnattedInstallCreateUser(this IUmbracoBuilder builder)
+        {
+            builder.AddNotificationAsyncHandler<UnattendedInstallNotification, CreateUnattendedUserNotificationHandler>();
+            //builder.AddNotificationHandler<UmbracoApplicationStarting, CreateUnattendedUserNotificationHandler>();
```
bergmania, line 302: Create the uncommented code

src/Umbraco.Web.Common/Install/CreateUnattendedUserNotificationHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+                // Update name, email & login & save user
+                admin.Name = unattendedName.Trim();
+                admin.Email = unattendedEmail.Trim();
+                admin.Username = unattendedEmail.Trim();
+                _userService.Save(admin);
+
+                // Change Password for the default user we ship out of the box
+                // Uses same approach as NewInstall Step
+
+                // TODO: usermanager why you no inject?!
```
bergmania, line 76: Remove todo

bergmania (changes requested):

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilderExtensions.cs
warrenbuckley: I was going to remove it as the original event worked.

warrenbuckley:
Will fix up all the loose ends when back at computer at some point over weekend.

src/Umbraco.Web.BackOffice/DependencyInjection/UmbracoBuilderExtensions.cs
warrenbuckley, line 51:
```suggestion
                .AddUnattedInstallCreateUser();
```

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilderExtensions.cs
warrenbuckley, line 302:
```suggestion
```

src/Umbraco.Web.Common/Install/CreateUnattendedUserNotificationHandler.cs
warrenbuckley, line 76:
```suggestion
```

warrenbuckley:
Fixed up all the little odds and ends that needed to be tidied up @bergmania 
Let me know if the naming of the settings makes sense or perhaps need to be renamed at all?

src/Umbraco.Web.Common/Install/CreateUnattendedUserNotificationHandler.cs
```diff
@@ -0,0 +1,99 @@
...
+                var unattendedEmail = _globalSettings.UnattendedUserEmail;
+                var unattendedPassword = _globalSettings.UnattendedUserPassword;
+
+                // Missing configuration values (json, env variables etc)
+                if (unattendedName.IsNullOrWhiteSpace()
+                    || unattendedEmail.IsNullOrWhiteSpace()
+                    || unattendedPassword.IsNullOrWhiteSpace())
+                {
+                    return;
+                }
```
bergmania, line 53: This is okay, but I think you should consider implementing some validation logic in `GlobalSettingsValidator`..

Currently, I guess It will explode if I use a non-email as `UnattendedUserEmail`.

You could also failed earlier if some, but not all of these settings are set.

bergmania (approved):

bergmania:
A few more comments... I can fix those up..

I think the Unattended-settings needs to be in their own section, now that we have so many..

Also I think we need to validate

bergmania:
I moved the settings from 

```
{
  "Umbraco": {
    "CMS": {
      "Global": {
        "UnattendedUserEmail": "...",
        "UnattendedUserName": "...",
        "UnattendedUserPassword": "...",
        "InstallUnattended" : true
      }
    }
  }
}

```

to

```
{
  "Umbraco": {
    "CMS": {
      "Unattended": {
        "UnattendedUserEmail": "...",
        "UnattendedUserName": "...",
        "UnattendedUserPassword": "...",
        "InstallUnattended" : true
      }
    }
  }
}
```

I also added validation. So if any of the `UnattendedUserEmail`, `UnattendedUserName`, `UnattendedUserPassword` is set when `InstallUnattended == true`, all of them needs to be set.  Also the Email is validated as email, so the defaultvalues are changed from `string.Empty` to `null`

PR #9935, opened 2021-03-04
User manager related updates

bergmania:


src/Umbraco.Web/Editors/PasswordChanger.cs
```diff
@@ -84,6 +85,11 @@ public PasswordChanger(ILogger logger, IUserService userService, HttpContextBase
                     return Attempt.Fail(new PasswordChangedModel { ChangeError = new ValidationResult("The current user is not authorized", new[] { "resetPassword" }) });
                 }
 
+                if (!currentUser.IsAdmin() && savingUser.IsAdmin())
+                {
+                    return Attempt.Fail(new PasswordChangedModel { ChangeError = new ValidationResult("The current user can't change password for specific user", new[] { "resetPassword" }) });
```
Shazwazza, line 90: Wording maybe should be "The current user cannot change password for the specified user" ?

src/Umbraco.Web/Editors/Filters/UserGroupEditorAuthorizationHelper.cs
```diff
@@ -77,17 +77,17 @@ public Attempt<string> AuthorizeGroupAccess(IUser currentUser, params string[] g
...
         public Attempt<string> AuthorizeSectionChanges(IUser currentUser,
-            IEnumerable<string> currentAllowedSections,
             IEnumerable<string> proposedAllowedSections)
         {
             if (currentUser.IsAdmin())
                 return Attempt<string>.Succeed();
 
-            var sectionsAdded = currentAllowedSections.Except(proposedAllowedSections).ToArray();
+
+            var currentAllowedSections = currentUser.Groups.SelectMany(x => x.AllowedSections).Distinct();
```
Shazwazza, line 89: This doesn't need to be here since `currentUser.AllowedSections` is exactly this and already exists.
I wish I could remember why `currentAllowedSections` was different. The value for that was only the sections on the current group.

Shazwazza (changes requested):
left a couple comments, the build isn't happy either https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=58284&view=results

src/Umbraco.Web/Editors/Filters/UserGroupEditorAuthorizationHelper.cs
bergmania: `currentAllowedSections` was the value of `userGroupSave.PersistedUserGroup.AllowedSections`

but that was equal to the values posted from the client due to the mapper.

bergmania: So is the big problem that the mapper are changing the sections too early? This should only happen if all checks are successeed just before we save the persisted usergroup again?

Shazwazza (approved):
Looks good to me!

src/Umbraco.Web/Editors/UserEditorAuthorizationHelper.cs
```diff
@@ -79,6 +79,13 @@ public UserEditorAuthorizationHelper(IContentService contentService, IMediaServi
             {
                 var savingGroupAliases = userGroupAliases.ToArray();
 
+                // As we know the current user is not admin, it is only allowed to use groups that the user do have themselves.
```
AndyButland, line 82: Doesn't seem like the comment here matches up with the logic below.  I.e. it sounds like the intention is to not allow the assignment of groups that the user isn't themselves part of, but actually it's only restricting the setting of the admin group.

bergmania: You are right, the comment is not updated.. The comment match my first implementation but that was not correct. Ill fix, thanks

bergmania: Actually, the logic in the UI seems to match the original comment.. I will look into what is up and down..

nikolajlauridsen:
Gave it a thorough testing, all clear 👍

PR #9937, opened 2021-03-04
Remove static events from ContentService and MediaService

### LINKED ISSUE #9397: Netcore: Remove static events

kjac:
### Prerequisites

### Description

This PR replaces (almost) all static events from `ContentService` and `MediaService` with the use of notifications via `IEventAggregator`. 

I tried to split it into two smaller PRs, one for each service... unfortunately it didn't really work well, so instead I ended up with this rather large PR.

### Items of particular interest in this PR

~~I have chosen not to port `EventState` and `AdditionalData` from `CancellableEventArgs` in this PR. I don't really like how the "ed" and "ing" events are tightly coupled like that... it feels fragile. In my opinion, event state should be a matter for the notification subscribers to handle in their specific implementations. See an example in `RedirectTrackingHandler` that uses `IRequestCache` to handle event state.~~ Notifications can now be stateful by implementing `IStatefulNotification`

I am painfully aware that the namespacing is not optimal right now. There is a general confusion going on as-is, mixing notifications and events in the same namespaces. The same goes for composers and handlers. I have tried my best to blend in 😄 but I'm sure there needs to be a thorough clean-up down the road.

### Out of scope / things to do moving forward

This PR does not rewrite the `TreeChanged` event on either service. This event is used for publishing events to the distributed cache system, and I feel like rewriting that would cause this PR to spin out of control. However, since these events are so generic, it is unlikely that they are used for a whole lot outside of the Umbraco core, so for the time being I have made them internal and marked them as up for rewrite later on.

`RelateOnTrashNotificationHandler` contains a lot of duplicate code that has simply been moved there from `RelateOnTrashComponent`. I have added a TODO in the class, so we can clean up this duplicate code later on.

`UserNotificationsHandler` contains a lot of duplicate code from `NotificationsComponent`. The latter is still around, but is anticipated to be removed once the remaining few static events used by it are removed, so I have not done anything to refactor this duplicate code.

### Testing this PR

In the Umbraco core, static events for `ContentService` and `MediaService` are currently used for things like:

- Redirects tracking.
- Email notifications.
- Building relations on copy / on move to recycle bin.
- File property editors cleaning up after themselves on delete.
- Complex property editors handling key regeneration on copy.

These are the things that should be tested to verify this PR. Here are some ideas:

#### Content

- Copy content and verify that a relation between the original and the copy is created if so chosen - and not created if not chosen.
- Move content to the recycle bin and verify that it can be restored to its original location.
- Copy content that contains a Nested Content or Block Editor property _with_ one or more items, and verify that the item keys are regenerated on copy (e.g. by inspecting the network traffic).
- Copy content that contains a File Upload or Image Cropper property _with_ an uploaded file, and verify that the file is copied along with the content (and the file reference is updated accordingly on the content).
- Delete content that contains a File Upload or Image Cropper property _with_ an uploaded file, and verify that the file is deleted on disk.
- Move and rename published content and verify that appropriate redirects appear in the "Redirect URL Management" dashboard.
- Enter a valid mail host in appsettings, trigger a notification email (i.e. content update notification) and verify that you receive the notification email.

#### Media

- Upload an image and verify that the built-in image properties (width, height, size etc) are automatically filled out and persisted.
- Move the uploaded image to the recycle bin and verify that it can be restored to its original location.
- Delete the uploaded image and verify that the file is deleted on disk.


src/Umbraco.Infrastructure/Services/Implement/ContentService.cs
```diff
@@ -773,7 +773,7 @@ public OperationResult Save(IContent content, int userId = Cms.Core.Constants.Se
 
                 if (raiseEvents)
                 {
-                    scope.Events.Dispatch(Saved, this, saveEventArgs.ToContentSavedEventArgs(), nameof(Saved));
+                    scope.Notifications.Publish(new SavedNotification<IContent>(content, evtMsgs));
```
Shazwazza, line 776: @kjac There's one potential issue here with these changes. The reason why in many cases there's an extensions like `ToContentSavedEventArgs()` is to ensure that the `AdditionalData` flows and is the same instance. This concept is used in a few places to be able to share data between the "Saving" (before) and "Saved" (after) events. 

This is just an example but this concept is used in most event raising places. 

Perhaps you've already addressed that somewhere in this code? I haven't had a deep dive into this, just a quick scan.

kjac: @Shazwazza yeah I know. I forgot to add that to the description - will update it. I have chosen to remove both `EventState` and `AdditionalData` in this implementation. I really dislike how the "ed" and "ing" events are tightly coupled like that... it feels fragile. In my opinion, event state should be a matter for the notification subscribers to handle in their specific implementations. 

If you look at the `RedirectTrackingHandler` rewrite in this PR, I have solved it there using `IRequestCache` as a temporary storage mechanism between the "ed" and "ing" events.

With this PR, `EventState` is no longer used anywhere in the Umbraco core, and even before this PR, `AdditionalData` is only ever used once (by `ModelsBuilderNotificationHandler`). 

Thoughts?

Shazwazza: Unfortunately that isn't gonna work. Request cache is only available during an http request. That will be null when running in a non request which happens all the time with services. We need to be able to pass state between events and cannot rely on an external state machine. That is even more fragile and is exactly why this has been done, else people will hack some static state bag. Happy to make this nicer some how but the event instance passed for before and after events will need to be able to carry state

kjac: Makes sense. This is not the hill I will die on, and I'll be happy to reintroduce the event state if that's the best solution. It just feels a bit strange to me to clone state between two instances of different classes in order to pass it between events.

Another idea could be to have a root scoped cache implementation on `Scope`. This could also be used by others throughout the current scope, thus not bound to events (notifications).

I dunno... just want to be sure we're doing state handling the best possible way here, since we'll have a hard time changing it later on. Again, happy to redo the event state on the cancelable notifications if that's the best approach.

Shazwazza: Totally happy to re-think how we do it! The 'AdditionalData' thing was just added as a means to an end without breaking changes. The Scope idea is interesting however I'm unsure that will work with the 'post' events since those execute when the scope is disposed and since we are not passing Scope around in these events we'd have to rely on the Ambient Scope which if acquired in the 'post' events may not be the same Scope instance (or in the hierarchy chain) as the one used in the service.

kjac: @Shazwazza I just had a go at it. You're absolutely right, it won't fly. I'll revert to reintroducing state on the notifications. 👍

kjac: Alright. I have introduced a new interface `IStatefulNotification`, which defines a `State` dictionary, and extension methods in `NotificationExtensions` to copy state between stateful notifications. 

`RedirectTrackingHandler` now utilizes this state to track redirects.

Note that both content and media service as-is have a few events that do not carry state between "ing" and "ed". I have tried to retain this behavior for the time being.

kjac:
Merged with latest `netcore/dev`

kjac:
Merged with latest `netcore/dev`

bergmania (changes requested):
Overall looks amazing.. Only thing I think we should improve is to have explicit notifications for content instead of generic. Also find a small typo

kjac:
@bergmania done and done 😄 it resulted in a looooooot of notification classes, and once we start working on the remaining static events there will be even more. But it _does_ make things way more discoverable, which outweighs the many classes 👍

bergmania:
> @bergmania done and done 😄 it resulted in a looooooot of notification classes, and once we start working on the remaining static events there will be even more. But it _does_ make things way more discoverable, which outweighs the many classes 👍

I know, but I agree, it is still a better solution, and it makes discoverability easier.. Amazing work.. Looks good, I will test again and merge if I do not find anything

bergmania (approved):

PR #9939, opened 2021-03-04
7383: Fixes urlAliases content finder failing on multilingual websites

### LINKED ISSUE #7383: UmbracoUrlAlias is not fully working on multi language website

OwainJ:
### Prerequisites

This fixes #7383 

## Quick description of the issue
If you have a multilingual site and have one-level path domains/hostnames set on the root node (e.g. **/en** & **/cy**)
image
And you have a urlAlias on any of your content nodes:
image
image
Then navigating to **"/en/test1"** or **"/en/test-2"**, as shown in the info tab, won't work; as the _ContentFinderByUrlAlias.cs_ is incorrectly including the **"/en"** portion of the URL in its comparison. As you can see here:
image

Interestingly, the unit tests were completely missing this issue, as the test cases were actually including the **/en** in their urlAliases! Because of this, I have completely re-written the unit tests for the ContentFinderByUrlAlias to hopefully provide better test coverage.

## Description of the fix
I have added an additional check to the _ContentFinderByUrlAlias_ comparison; It now retrieves the absolute URL of the root node, joins it with the _urlAlias_ of the node being checked, and then compares it with the absolute URL of the current request.

So instead of this comparison:
FindContentByAlias_old_comparison

It now does this comparison:
FindContentByAlias_new_comparison

image

GIF of the urlAlias now working:
ContentFinderByUrlAlias_working

### Why have I kept the original comparison, instead of only using the new absolute URL comparison?
The original comparison has a slight issue, as it sees this "https://localhost:44331/test1/", instead of https://localhost:44331/en/test1/, as valid URL (you can see this behaviour in the gif on #7383) and fixing that would probably count as a breaking change, so I left that behaviour in.

### Steps to test
- Create a multilingual site
- Set the hostnames of the root node to "/en" and "/cy" (or similar)
- Add some urlAliases to your content nodes
- Attempt to visit the URLs shown in the node's info tab

I have re-written the _ContentFinderByAliasWithDomainsTests_ unit tests to provide more coverage and test a wider range of hostnames. I would also really appreciate it if someone could look over and double-check the new unit tests, just in case I've done something silly or missed an edge case. :)


emma-hq:
Hey @OwainJ 

I just realised that while I've had this PR open in my browser for days, I haven't even said a thank you for raising it 😁 

We're super grateful for the contribution - I'm particularly grateful as it's linked to a very old and popular issue. We're looking and talking and when we have any conclusions or questions we will tell you them.

Emma

src/Umbraco.Web/Routing/ContentFinderByUrlAlias.cs
```diff
@@ -90,9 +93,20 @@ bool IsMatch(IPublishedContent c, string a1, string a2)
...
-                return v.InvariantContains(a1) || v.InvariantContains(a2);
+
+                // Build up a comparison list of the current node's urlAlias to compare with the current request
+                var sb = new StringBuilder(",");
+                foreach (var urlAlias in v.Split(','))
+                {
+                    // Add the node's urlAlias to the list
+                    sb.Append(urlAlias.Replace(" ", "") + ",");
+                    // Add the root node's absolute url joined to the node's urlAlias to the list
+                    sb.Append(rootAbsoluteUrl + (urlAlias.StartsWith("/") || rootAbsoluteUrl.EndsWith("/") ? "" : "/") + urlAlias.Replace(" ", "") + ",");
```
benjaminc, line 106: If the `rootAbsoluteUrl` ends with a slash AND the `urlAlias` starts with a slash, you'll not match here because there will be a double slash.  Better to add `.TrimEnd('/') + '/'` to then end of lines 118 and 124 below (to ensure the `rootAbsoluteUrl` does end in a slash), and change this line to read as follows:
```
sb.Append(rootAbsoluteUrl + urlAlias.TrimStart('/').Replace(" ", "") + ",");
```

ronaldbarendse: There's an extension method within Umbraco for doing exactly this: https://github.com/umbraco/Umbraco-CMS/blob/7f1d5fe7cc4615846f2df555246bb58e8cd023c8/src/Umbraco.Core/StringExtensions.cs#L463-L466

OwainJ: @benjaminc Good shout! I'll change that asap! 👍 
@ronaldbarendse Ah awesome! Thanks for letting me know about that 😀

nul800sebastiaan:
Hey all, and so sorry that we haven't gotten back to this yet. We recently had a lively discussion at HQ and with some community people on what is going on here.. 
I know I encouraged a PR for this but little did I know the rabbit hole that we would find ourselves in when actually discussing this.

For a bit of background `umbracoUrlAlias` was introduced a long, long time ago in Umbraco 3 if I'm not mistaken, those were more innocent days where we hadn't even started thinking about variants yet. 

However, at the time we DID have multilingual sites and the ability to set a hostname to `/en` etc. The idea behind `umbracoUrlAlias` has always been to take it very, very literally. Whatever you type in there will be the exact new url where the content will be available. We can't fix it this way as it will change that behavior. 

From a quick skim of the updated code, I believe there's a fallback in there so hopefully the existing URLs don't actually break.. but we feel that this code is already too hard to understand and by bolting on even more if's and else's and complicating it even more we're not helping anyone.

Instead.. we would love to see a package with a property editor that can help you set up this alias with the appropriate prefixes if the editor wishes. To help the editor understand what's going to be happening as well, a preview URL could be shown right inside that prop editor.

Finally.. the correct fix for this bug should be that the URL panel on the Info tab in Umbraco should show you the actual URL that this will create (as in: exactly what has been typed in the textbox for it).

Apologies to @OwainJ and everyone involved in this thread, we try not to waste anybody's time and in this case the directions in the issues should have been clearer, I will try to pay more attention to this in the future.

OwainJ:
Ah no worries @nul800sebastiaan! I don't feel like my time was wasted, this gave me a good opportunity to poke around and learn how unit tests worked in the Umbraco source! 

I completely understand that having the `umbracoUrlAlias` be a literal URL and ignore the root node's hostname is extremely useful, but I'm torn, as I also do agree with @DanDiplo; that expecting the `umbracoUrlAlias` be appended onto the root node's hostname also makes complete logical sense, and having to manually add `/en` every time feels like a pain... But, on the other hand, does the increased flexibility of having the `umbracoUrlAlias` whatever you want, outweigh that annoyance?

Maybe having a configurable toggle to turn this behaviour on and off would be a good idea? Could default it to off to preserve the current behaviour for backwards compatibility.

Our client who originally reported this issue to us actually uses the `umbracoUrlAlias` as a quick easy way to redirect old SharePoint URLs to the new Umbraco content. So yea, perhaps having the flexibility of not being forced to have the `/en` prefixed on the `umbracoUrlAlias` would be more useful to them in their use case. (but I still think a toggle would be a good idea)

Either way, fixing the preview URLs in the info tab to correctly reflect the actual URL is **definitely** a priority.

umbracocommunity:
This pull request has been mentioned on **Umbraco community forum**. There might be relevant details there:

https://forum.umbraco.com/t/friendly-url/4323/6

PR #9942, opened 2021-03-05
Netcore: Individual nuget  packages for each assembly

nikolajlauridsen:
### Description
This PR switches the build procedure to use `dotnet pack` which creates a separate nuget package for each project in the solution. 

Dependencies are handles through project references. UmbracoCms.nuspec has been changed to have a dependency to `Umbraco.Cms.Web.Website` and `Umbraco.Cms.Web.BackOffice`, all other projects will then be included as implicit references.

### Test procedure
To test the pr: 
1. Enter the root folder in a powershell prompt
2. Run the build with `.\build\build.ps1`
   1. Once the build is done you can see the nuget packages in the `build.out` folder
2. Open a normal terminal and enter the folder where you want to create the test project. 
3. To be on the safe side I'd recommend clearing you nuget cache with `dotnet nuget locals all --clear`
4. Add the Examine-Beta pipeline with `dotnet nuget add source https://shazwazza.pkgs.visualstudio.com/Examine/_packaging/Examine-Beta/nuget/v3/index.json "ExamineAzurePipelines"`
5. Add the `build.out` folder as a local nuget feed with: `dotnet nuget add source <Path\to\build.out> -n "LocalFeed"`
6. Install the template: `dotnet new -i Umbraco.Templates::9.0.0-alpha004`
7. Create a project from the template: `dotnet new umbraco -n <projectName>`
   1. To test it with SqlCE do `dotnet new umbraco -U true -n <projectName>`

Ensure that the project builds and run and that the correct dependencies are installed, and that the metadata has been updated.


build/NuSpecs/UmbracoCms.Examine.Lucene.nuspec
```diff
@@ -0,0 +1,48 @@
...
+        <language>en-US</language>
+        <tags>umbraco</tags>
+        <repository type="git" url="https://github.com/umbraco/umbraco-cms" />
+        <dependencies>
+
+            <group targetFramework="netstandard2.0">
+                <dependency id="Umbraco.Cms.Core" version="[$version$]" />
+                <dependency id="Umbraco.Cms.Infrastructure" version="[$version$]" />
+                <dependency id="Newtonsoft.Json" version="[12.0.3,12.999999)" />
+                <dependency id="NPoco" version="[4.0.2,4.999999)" />
```
bergmania, line 24: Are we not missing `System.Configuration.ConfigurationManager` here? 

We have the following TODO in netcore csproj, that could be relavant

```
<!-- TODO: remove the reference to System.Configuration.ConfigurationManager when Examine/lucene dont need it-->
<PackageReference Include="System.Configuration.ConfigurationManager" Version="5.0.0" />

```

src/Umbraco.Examine.Lucene/Umbraco.Examine.Lucene.csproj
```diff
@@ -5,6 +5,10 @@
         <RootNamespace>Umbraco.Cms.Infrastructure.Examine</RootNamespace>
         <Product>Umbraco CMS</Product>
         <Title>Umbraco.Examine.Lucene</Title>
+        <!--We have a specific nuspec file for this, do not pack. -->
+        <IsPackable>false</IsPackable>
```
bergmania, line 9:
```suggestion
        <!--TODO: Remove this and the specific nuspec file when examine.Lucene is netstandard compatible.-->
        <!--We have a specific nuspec file for this, do not pack. -->
        <IsPackable>false</IsPackable>
```

build/NuSpecs/UmbracoCms.Examine.Lucene.nuspec
nikolajlauridsen: Hmm that's a good question, I didn't see that TODO in the netcore csproj. I just used the dependencies from the Examine.Lucene.csproj.

I don't know if we actually need to add it to the Examine.Lucene.nuspec file though, since Umbraco.Core.csproj already has `<PackageReference Include="System.Configuration.ConfigurationManager" Version="5.0.0" />` so it will still be installed as an implicitly referenced version from there.

On the other hand I don't think it would hurt either, if we want to be 100% sure that it's there regardless of Umbraco.Core

bergmania: Cool I guess that is another problem.. Core should not have that dependency. Seems to be because we just use an exception.

I can fix that part up

Issue #9940, opened 2021-03-05
BackOffice Translations tab (Dictionary), overview table has shuffled columns names


Meyhem:
## Description

We are using translation feature of umbraco and it contains bug in dictionary overview table, where it is supposed to show whether value is present for each language. Problem is that the table has shuffled columns and displays the language name in the column wrongly. 

e.g. key "account.Accounting.AvailableBalance" is translated for English (united states) and Indonesia (no other), the table columns names are showing that it's translated for English (India) and English (Malaysia) (no other). So my guess is that table has shuffled languages in the columns names (see screenshots)

Actual translated values:
![image](https://user-images.githubusercontent.com/6241657/110097839-c7c85880-7d9f-11eb-94b0-801baa144a40.png)

Table showing different language:
![image](https://user-images.githubusercontent.com/6241657/110097740-af583e00-7d9f-11eb-9178-7f142a71d6c5.png)

## Umbraco version
**8.11.1** (It was observed also on 8.8.0)

### Bug summary
When having site with multiple languages and using Translations feature, the Translations overview has shuffled columns names with language names and provide wrong information about which languages needs to be added to translation keys.

### Specifics
Translations overview table path: **/umbraco/#/translation/dictionary/list**
Version: 8.11.1
Reproducible in browsers: Firefox 86, Chrome 89 (appears to be browser independent)

### Steps to reproduce

Start umbraco backoffice. Add several languages to Umbraco (Settings -> Languages), in my case, the languages used are 
**English (India), English (Malaysia), English (United States), Hindi, Indonesia, Malay (Malaysia)**.

Then go to Translations tab in backoffice and create some key, but provide translated values only for English (united states) & Indonesia. Save.

Then go back to Translations overview, and observe the columns names. They should be shuffled.

### Expected result

The overview table should display correct languages in columns names.

### Actual result

The overview table displays shuffled languages in columns names, misguiding user to think different languages are filled/missing

patrickdemooij9:
I created a pull request for it: https://github.com/umbraco/Umbraco-CMS/pull/9944. There was no ordering on the overview list

Meyhem:
@patrickdemooij9 wow, thanks for the quick fix.

nathanwoulfe:
Fixed in #9944 👍

PR #9944, opened 2021-03-06
9940: Make sure that ordering is correct

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9940

### Description
The ordering in the translation tab was being used on the wrong ng-repeat element. Moving it to the correct one fixes the issue.

Steps to reproduce can be found in the issue.


nathanwoulfe (approved):
Thank @patrickdemooij9, works correctly now. I've also added changes to use the umb-icon directive for the icons (nothing related to your fix, just saw they still used the old approach)

N

PR #9946, opened 2021-03-07
Member group (identity role) and store

emmagarland:
Implement role functionality for members in .NET Core (identity roles and store). Uses default manager.

Based on the new ASP.NET Core identity implementation for the back office, members, and [UmbracoIdentity](https://github.com/Shazwazza/UmbracoIdentity).

This relates to the larger issue for the whole members migration #9471

### Prerequisites

[ x ] I have added steps to test this contribution in the description below

### Description

Continuing from the Create Member PR #9472 

- No database changes
- Implementation of a store and controller updates for the Members identity implementation
- Adjusted the controller to use the new manager- several questions
- Unit tests for new code

### Testing

1. Run unit tests
1. Run integration tests
1. Login to back-office and create a member with a role. Member should be persisted to the database as it is currently in Umbraco 8, with associated role.
1. Login to back-office and create a role (member group). Role should be persisted to the database as it is currently in Umbraco 8
1. Login to back-office and update a role. Role updates should be persisted to the database as they are in Umbraco 8.
1. Login to back-office and delete a role. Role should be deleted from database


src/Umbraco.Core/Models/Mapping/MemberTabsAndPropertiesMapper.cs
```diff
@@ -237,6 +237,7 @@ internal static ContentPropertyDisplay GetLoginProperty(IMemberTypeService membe
             var userRoles = username.IsNullOrWhiteSpace() ? null : _memberService.GetAllRoles(username);
 
             // create a dictionary of all roles (except internal roles) + "false"
+            //TODO: use MembersRoleStore
```
emmagarland, line 240: Do we need to use the member roles store in places like this?

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/MemberRepository.cs
```diff
@@ -314,6 +314,7 @@ protected override void PersistNewItem(IMember entity)
             // persist the member dto
             dto.NodeId = nodeDto.NodeId;
 
+            // TODO: password parts of this file need updating
```
emmagarland, line 317: I found a bit of legacy password functionality in `PersistNewItem` - does it need changing now?

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
```diff
@@ -0,0 +1,342 @@
...
+{
+    /// <summary>
+    /// A custom user store that uses Umbraco member data
+    /// </summary>
+    public class MemberRoleStore<TRole> : IRoleStore<TRole> where TRole : IdentityRole
+    {
+        private readonly IMemberGroupService _memberGroupService;
+        private bool _disposed;
+
+        //TODO: Move into custom error describer.
```
emmagarland, line 18: What is our preferred error message approach?

src/Umbraco.Web.BackOffice/Controllers/ContentController.cs
```diff
@@ -2435,6 +2435,7 @@ public IActionResult GetPublicAccess(int contentId)
                 .Select(_umbracoMapper.Map<MemberDisplay>)
                 .ToArray();
 
+            //TODO: change to role store
```
emmagarland, line 2438: Would this also need a change to the role store? Or remain with the service?

src/Umbraco.Web.BackOffice/Controllers/MemberGroupController.cs
```diff
@@ -26,45 +28,62 @@ public class MemberGroupController : UmbracoAuthorizedJsonController
...
         /// <summary>
         /// Gets the member group json for the member group guid
         /// </summary>
         /// <param name="id"></param>
         /// <returns></returns>
-        public ActionResult<MemberGroupDisplay> GetById(Guid id)
+        public async Task<ActionResult<MemberGroupDisplay>> GetById(Guid id)
         {
-            var memberGroup = _memberGroupService.GetById(id);
+            //TODO: did we envisage just identity or a combination of service and identity manager?
```
emmagarland, line 80: In these instances, not sure whether to replace with identity or remain as a service, because the display needs to map to the role anyway and we don't have all those properties

src/Umbraco.Web.BackOffice/Mapping/MemberMapDefinition.cs
```diff
@@ -100,5 +103,18 @@ private static void Map(IMember source, ContentPropertyCollectionDto target, Map
...
+
+        /// <summary>
+        /// Maps an identity role to a member group display
+        /// </summary>
+        /// <param name="arg1"></param>
+        /// <param name="arg2"></param>
+        /// <param name="arg3"></param>
+        private void Map(IdentityRole source, MemberGroupDisplay target, MapperContext context)
+        {
+            //TODO: this is all that is mapped at this time, we're losing a lot of properties
```
emmagarland, line 115: As per above comment re mapping - what do we need to return in the controller cases? This might be unwanted in the `MemberGroupController` cases.

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
```diff
@@ -0,0 +1,342 @@
...
+        public Task<IdentityResult> CreateAsync(TRole role, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                ThrowIfDisposed();
+
+                if (role == null)
+                {
+                    throw new ArgumentNullException(nameof(role));
```
scottbrady91, line 45: **nitpick:** consider letting these exceptions throw. I wouldn't return an `IdentityResult` for things like this. If the store manages to be called with a null role, then I'd say that's exceptional behavior. I'm not sure what the convention is though

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
```diff
@@ -0,0 +1,342 @@
...
+        {
+            try
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                if (role == null)
+                {
+                    throw new ArgumentNullException(nameof(role));
+                }
+
+                ThrowIfDisposed();
```
scottbrady91, line 77: **nitpick:** the other methods check for disposed before null role

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
```diff
@@ -0,0 +1,342 @@
...
+            if (role == null)
+            {
+                throw new ArgumentNullException(nameof(role));
+            }
+
+            return Task.FromResult(role.Name);
+        }
+
+        /// <inheritdoc />
+        public Task SetRoleNameAsync(TRole role, string roleName, CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
scottbrady91: https://github.com/dotnet/aspnetcore/blob/main/src/Identity/Extensions.Stores/src/RoleStoreBase.cs#L116

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
```diff
@@ -0,0 +1,342 @@
...
+            cancellationToken.ThrowIfCancellationRequested();
+            ThrowIfDisposed();
+
+            if (role == null)
+            {
+                throw new ArgumentNullException(nameof(role));
+            }
+
+            //TODO: are we utilising NormalizedRoleName?
+            return Task.FromResult(role.NormalizedName);
```
scottbrady91, line 219: **suggestion:** consider just returning role.Name

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
```diff
@@ -0,0 +1,342 @@
...
+            cancellationToken.ThrowIfCancellationRequested();
+            ThrowIfDisposed();
+
+            if (role == null)
+            {
+                throw new ArgumentNullException(nameof(role));
+            }
+
+            //TODO: are we utilising NormalizedRoleName and do we need to set it in the memberGroupService?
+            role.NormalizedName = normalizedName;
```
scottbrady91, line 234: **suggestion:** I would treat name and normalized name as one and the same for now. I think that's what users is doing

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
```diff
@@ -0,0 +1,342 @@
...
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            ThrowIfDisposed();
+
+            if (string.IsNullOrWhiteSpace(name))
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+            IMemberGroup memberGroup = _memberGroupService.GetByName(name);
+            //TODO: throw exception when not found?
```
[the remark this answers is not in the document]
scottbrady91: The RoleManager will pass null back too. The caller can decide what to do with that

src/Umbraco.Web.BackOffice/Controllers/ContentController.cs
scottbrady91: I assume the controller would use the `RoleManager`, rather than the store directly

src/Umbraco.Web.Common/Umbraco.Web.Common.csproj
```diff
@@ -21,6 +21,9 @@
     </ItemGroup>
 
     <ItemGroup>
+      <PackageReference Include="Microsoft.AspNet.Identity.Core" Version="2.2.3" />
```
scottbrady91, line 24: **issue:** This is part of the framework now (which is why there is no v3+ package). If you're using .NET 5, you'll need a `FrameworkReference` in here instead.

scottbrady91 (changes requested):
I've tried answering some of your TODOs too. Let me know what you think.

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
emmagarland: Good point @scottbrady91 - @bergmania  @Shazwazza  what is our preference here? Let the exception throw, or catch and return a failed identity result like:

`return Task.FromResult(IdentityResult.Failed(new IdentityError { Code = genericIdentityErrorCode, Description = ex.Message }));`

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
emmagarland: Good spot, amended

src/Umbraco.Web.BackOffice/Controllers/ContentController.cs
emmagarland: Changed to role manager wording

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
emmagarland: Updated

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
emmagarland: Thanks - I have changed to just call the Get/ Set Name methods

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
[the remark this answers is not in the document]
emmagarland: Thanks, removed TODO

src/Umbraco.Web.Common/Umbraco.Web.Common.csproj
emmagarland: Oh, I didn't mean for this change to happen. Will revert!

src/Umbraco.Core/Models/Mapping/MemberTabsAndPropertiesMapper.cs
emmagarland: *manager, not role store

src/Umbraco.Web.BackOffice/Controllers/ContentController.cs
emmagarland: Question still stands from Umbraco logic POV, how far are we going with replacing services with Role Manager in this PR?

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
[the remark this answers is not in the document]
emmagarland: Like: ` role.Name = roleName;`?

@scottbrady91 thanks, just to check: 

Does that mean any role name change that needs to be persisted via the `MemberGroupService` will need to be done via  `_roleManager.UpdateAsync(role);`? 

What if the user only wants to change the role name and persist it to the Umbraco store?
`_roleManager.SetRoleNameAsync(role);`

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
emmagarland: Leaving conversation open to confirm with Umbraco HQ

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
bergmania: IMO an `ArgumentNullException` should throw..

emmagarland: Thanks, have removed the try catch

emmagarland:
@bergmania I merged from dev but it's still failing, any idea why? Most of the other build checks passed though

src/Umbraco.Tests.UnitTests/Umbraco.Tests.UnitTests.csproj
```diff
@@ -37,4 +37,10 @@
     <None Remove="umbraco\Data\TEMP\TypesCache\umbraco-types.DESKTOP-EU212M2.hash" />
   </ItemGroup>
 
+    <ItemGroup>
+      <Reference Include="Umbraco.Infrastructure">
+        <HintPath>..\Umbraco.Web\bin\Debug\Umbraco.Infrastructure.dll</HintPath>
+      </Reference>
+    </ItemGroup>
```
bergmania, line 44: This is a problem

src/Umbraco.Web.Common/Umbraco.Web.Common.csproj
```diff
@@ -24,6 +24,8 @@
     </ItemGroup>
 
     <ItemGroup>
+      <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="5.0.2" />
+      <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="5.0.2" />
```
bergmania, line 28: Looks like merge issue here.. I think both should be removed here..

bergmania (commented):
I can fix the build and my two comments here

bergmania:
The reason the build artifacts failed, is shown in the error description here:
https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=59424&view=logs&j=03747cdd-653d-5fb0-85c6-8a40b0c1e8d2&t=34c4a62f-3734-5b93-ff31-64ca97696641

bergmania:
> @bergmania I merged from dev but it's still failing, any idea why? Most of the other build checks passed though

The error mentions:
```
Microsoft.AspNetCore.Mvc.NewtonsoftJson
 5.0.3 required by Umbraco.Web.UI.NetCore
 5.0.2 required by Umbraco.Web.Common
Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation
 5.0.2 required by Umbraco.Web.Common
 5.0.3 required by Umbraco.Web.Common

```

src/Umbraco.Core/Models/Mapping/MemberTabsAndPropertiesMapper.cs
```diff
@@ -237,6 +237,7 @@ internal static ContentPropertyDisplay GetLoginProperty(IMemberTypeService membe
             var userRoles = username.IsNullOrWhiteSpace() ? null : _memberService.GetAllRoles(username);
 
             // create a dictionary of all roles (except internal roles) + "false"
+            //TODO: use member role manager instead
```
Shazwazza, line 240: Nope, editing members in the back office is purely to do with our own member data. We don't want to mix this editing with the ability to extend/replace a role manager - which will only be used in limited cases.

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
Shazwazza: I think we'll need to revisit errors from our identity implementation since that isn't a thing in v8. We probably have some of these errors already translated in our language files/service. I think we'll need to create a separate task for dealing with that. (i'll create one)

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
```diff
@@ -0,0 +1,293 @@
...
+                if (MapToMemberGroup(role, memberGroup))
+                {
+                    _memberGroupService.Save(memberGroup);
+                }
+                //TODO: if nothing changed, do we need to report this?
+                return Task.FromResult(IdentityResult.Success);
+            }
+            else
+            {
+                //TODO: throw exception when not found, or return failure?
```
Shazwazza, line 87: Its a good question and seems that the Entity Framework implementation doesn't specifically deal with this  but seems to indicate that no exceptions are thrown if something fails. https://github.com/dotnet/aspnetcore/blob/main/src/Identity/EntityFrameworkCore/src/RoleStore.cs#L157

So I think we stick to what you have with the Failed approach.

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
```diff
@@ -0,0 +1,293 @@
...
+            }
+
+            IMemberGroup memberGroup = _memberGroupService.GetById(roleId);
+            if (memberGroup != null)
+            {
+                if (MapToMemberGroup(role, memberGroup))
+                {
+                    _memberGroupService.Save(memberGroup);
+                }
+                //TODO: if nothing changed, do we need to report this?
```
Shazwazza, line 82: Yep I think that is correct

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
```diff
@@ -0,0 +1,293 @@
...
+            ThrowIfDisposed();
+
+            if (role == null)
+            {
+                throw new ArgumentNullException(nameof(role));
+            }
+
+            if (!int.TryParse(role.Id, out int roleId))
+            {
+                //TODO: what identity error should we return in this case?
```
Shazwazza, line 106: I think in this case we throw an exception since it shouldn't happen

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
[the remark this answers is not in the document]
Shazwazza: Yep like @scottbrady91  mentions, these methods will just need to set properties and then they will be persisted with UpdateAsync. This means we need to do change tracking on all the properties (if we aren't already) just like we do with the MemberIdentity. And yes if using this method, you then need to call UpdateAsync.

Is there a reason we aren't inheriting from RoleStoreBase? Seems like we would get most of the functionality for free? and don't have to implement methods like this.

src/Umbraco.Infrastructure/Security/MemberUserStore.cs
```diff
@@ -308,24 +363,37 @@ protected override async Task<IdentityUserLogin<string>> FindUserLoginAsync(stri
...
-            MembersIdentityUser user = await FindUserAsync(userId, cancellationToken);
+            MemberIdentityUser user = await FindUserAsync(userId, cancellationToken);
             if (user == null)
             {
-                return null;
+                //TODO: error throw or null result?
+                return await Task.FromResult((IdentityUserLogin<string>)null);
+            }
+
+            if (string.IsNullOrWhiteSpace(loginProvider))
```
Shazwazza, line 373: These checks should be before member lookups

src/Umbraco.Web.BackOffice/Controllers/ContentController.cs
```diff
@@ -2435,6 +2435,7 @@ public IActionResult GetPublicAccess(int contentId)
                 .Select(_umbracoMapper.Map<MemberDisplay>)
                 .ToArray();
 
+            //TODO: change to role manager
```
Shazwazza, line 2438: Yes public access will need to change to use the role manager. We have a separate task for public access

src/Umbraco.Web.BackOffice/Controllers/MemberGroupController.cs
```diff
@@ -26,45 +28,62 @@ public class MemberGroupController : UmbracoAuthorizedJsonController
...
         {
             _memberGroupService = memberGroupService ?? throw new ArgumentNullException(nameof(memberGroupService));
+            _roleManager = roleManager ?? throw new ArgumentNullException(nameof(roleManager));
             _umbracoMapper = umbracoMapper ?? throw new ArgumentNullException(nameof(umbracoMapper));
-            _localizedTextService =
-                localizedTextService ?? throw new ArgumentNullException(nameof(localizedTextService));
+            _localizedTextService = localizedTextService ?? throw new ArgumentNullException(nameof(localizedTextService));
         }
 
+        //TODO: are there any repercussions elsewhere for us changing these to async?
```
Shazwazza, line 46: nope

src/Umbraco.Web.BackOffice/Controllers/MemberGroupController.cs
```diff
@@ -26,45 +28,62 @@ public class MemberGroupController : UmbracoAuthorizedJsonController
...
         /// <summary>
         /// Gets the member group json for the member group id
         /// </summary>
         /// <param name="id"></param>
         /// <returns></returns>
-        public ActionResult<MemberGroupDisplay> GetById(int id)
+        public async Task<ActionResult<MemberGroupDisplay>> GetById(int id)
         {
-            var memberGroup = _memberGroupService.GetById(id);
+            //TODO: did we envisage this - combination of service and identity manager?
```
Shazwazza, line 55: Nope, we shouldn't use the role manager for these editors, only our member service. The Role manager will almost solely be used for public access.

src/Umbraco.Web.BackOffice/Controllers/MemberGroupController.cs
Shazwazza: We should only use our member services for these editors .The Role manager will almost solely be used for public access

src/Umbraco.Web.BackOffice/Mapping/MemberMapDefinition.cs
Shazwazza: I don't think we need this map since we shouldn't be using the role manager for our member editors.

src/Umbraco.Web.BackOffice/Trees/MemberGroupTreeController.cs
```diff
@@ -49,6 +50,7 @@ protected override ActionResult<TreeNode> CreateRootNode(FormCollection queryStr
             var root = rootResult.Value;
 
             //check if there are any groups
+            //TODO: change to role store
```
Shazwazza, line 53: nope, we'll only use our member services for our member editors.

Shazwazza (changes requested):
Sorry I should have provided a bunch of additional info for the role manager requirements before beginning. The role manager will almost solely be used for public access and that's pretty much it. For all member editors including member group editors in the back office, those will only use our member services, not the identity implementations. The only thing Identity is used for in our editors is for security/password management. We won't be using the role manager at all in our member editors/trees so most of that can be rolled back.

I'll spend some time today hopefully and update this PR. Awesome work! and thanks so much for all the effort :)

src/Umbraco.Infrastructure/Security/MemberUserStore.cs
```diff
@@ -308,24 +363,37 @@ protected override async Task<IdentityUserLogin<string>> FindUserLoginAsync(stri
             cancellationToken.ThrowIfCancellationRequested();
             ThrowIfDisposed();
 
-            MembersIdentityUser user = await FindUserAsync(userId, cancellationToken);
+            MemberIdentityUser user = await FindUserAsync(userId, cancellationToken);
             if (user == null)
             {
-                return null;
+                //TODO: error throw or null result?
```
Shazwazza, line 369: Nulls are good, we'll follow what the entity framework implementation does https://github.com/dotnet/aspnetcore/blob/main/src/Identity/EntityFrameworkCore/src/UserStore.cs#L308

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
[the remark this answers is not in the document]
Shazwazza: Ah, I see RoleStoreBase is a large object that implements much more than we will support. Make sense we just implement the interface 👍  I'll fix this one up so that it's just setting the property and it will be persisted during update.

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/MemberRepository.cs
Shazwazza: This password logic here can stay. It's purely used for dealing with flagging a member as not having a set password so the value uses a special prefix. The usage of the IPasswordHasher here is irrelevant, it might as well be a SHA hash but it's fine the way it is.

Shazwazza:
Hi all,

I've pushed changes:

* Removed TODOs where I've answered above
* Created a sub class UmbracoIdentityRole : IdentityRole which adds change tracking like our other identity classes
* Removes the generic argument from MemberRoleStore since we know what object we're dealing with
* Changes some tests to use builder pattern
* Reverts changes to member controllers/trees since we don't want to use the role manager in these scenarios, this is purely for editing Umbraco member data and we won't support external role data being edited if someone chose to replace the role manager

Just waiting for tests to pass and i will merge. Please let me know if you have any question. Cheers!

src/Umbraco.Infrastructure/Security/MemberRoleStore.cs
emmagarland: Ah yes, makes sense as a separate task.

emmagarland:
> The reason the build artifacts failed, is shown in the error description here:
> https://dev.azure.com/umbraco/Umbraco%20Cms/_build/results?buildId=59424&view=logs&j=03747cdd-653d-5fb0-85c6-8a40b0c1e8d2&t=34c4a62f-3734-5b93-ff31-64ca97696641
> 
> ```
> Verify NuGet consistency
> 
> Microsoft.AspNetCore.Mvc.NewtonsoftJson
>  5.0.3 required by Umbraco.Web.UI.NetCore
>  5.0.2 required by Umbraco.Web.Common
> Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation
>  5.0.2 required by Umbraco.Web.Common
>  5.0.3 required by Umbraco.Web.Common
> ##[error]Exception calling "VerifyNuGet" with "0" argument(s): "
> ```

Thanks @bergmania for explaining

src/Umbraco.Core/Models/Mapping/MemberTabsAndPropertiesMapper.cs
emmagarland: Gotcha, we're only using it for our own limited case not for the whole application. Thanks for explaining

emmagarland:
> Hi all,
> 
> I've pushed changes:
> 
> * Removed TODOs where I've answered above
> * Created a sub class UmbracoIdentityRole : IdentityRole which adds change tracking like our other identity classes
> * Removes the generic argument from MemberRoleStore since we know what object we're dealing with
> * Changes some tests to use builder pattern
> * Reverts changes to member controllers/trees since we don't want to use the role manager in these scenarios, this is purely for editing Umbraco member data and we won't support external role data being edited if someone chose to replace the role manager
> 
> Just waiting for tests to pass and i will merge. Please let me know if you have any question. Cheers!

Amazing stuff @Shazwazza  - yes I see what you mean re the Role Manager only being needed for the back-office side of things now. Thanks for explaining and sorting :)

PR #9951, opened 2021-03-08
Netcore: create service/repo and remove SQL calls from DatabaseServerMessenger

AndyButland:
Relating to issue #7668, this PR extracts the database queries from `DatabaseServerMessenger` (and derived `BatchedDatabaseServerMessenger`) and creates a new service and repository class where appropriate business logic and database access is moved.

As far as I can see from stepping through it's still working as it did before, but will need a careful review of course.  Have added some tests to verify the new service and repository, but as there wasn't anything for the messenger classes before I can't verify from automated tests the behaviour is exactly the same.


AndyButland:
Putting this back to draft as have had some thoughts about better handling the locks around the initialization code...

AndyButland:
Putting this back as ready for review now.

nikolajlauridsen:
I notice that in `CacheInstructionRepository` you inject `IScopeAccessor` and use the `AmbientScope`, this will work, because you in `CacheInstructionService` create explicit scopes before calling the methods from the repository, which means that AmbientScope will be the explicit scope.

I am however worried of the possibility of someone calling the repository methods without first creating an explicit scope, which could leave to unexpected behaviour, one way we could handle this is by instead requesting an explicit scope in the methods, for instance:

```c#
public bool Exists(IScope scope, int id) => scope.Database.Exists<CacheInstructionDto>(id);
```

Instead of:

```c#
public bool Exists(int id) => AmbientScope.Database.Exists<CacheInstructionDto>(id);
```

This way we're 100% certain that the methods will be called in an explicit scope, thoughts?

src/Umbraco.Infrastructure/Services/Implement/CacheInstructionService.cs
```diff
@@ -0,0 +1,459 @@
...
+    public class CacheInstructionService : RepositoryService, ICacheInstructionService
+    {
+        private readonly IServerRoleAccessor _serverRoleAccessor;
+        private readonly CacheRefresherCollection _cacheRefreshers;
+        private readonly ICacheInstructionRepository _cacheInstructionRepository;
+        private readonly IProfilingLogger _profilingLogger;
+        private readonly ILogger<CacheInstructionService> _logger;
+        private readonly GlobalSettings _globalSettings;
+
+        private readonly object _locko = new object();
```
nikolajlauridsen, line 32: This lock object is never used as far as I can tell.

src/Umbraco.Infrastructure/Services/Implement/CacheInstructionService.cs
```diff
@@ -0,0 +1,459 @@
...
+                var result = NotifyRefreshers(instructionBatch, processed, released);
+                if (result)
+                {
+                    // If all instructions were processed, set the last id.
+                    lastId = instruction.Id;
+                }
+
+                return result;
+            }
+            //catch (ThreadAbortException ex)
```
nikolajlauridsen, line 313: Should this be removed?

src/Umbraco.Core/Services/ICacheInstructionService.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <returns></returns>
+        int GetMaxInstructionId();
+
+        /// <summary>
+        /// Creates a cache instruction record from a set of individual instructions and saves it.
+        /// </summary>
+        void DeliverInstructions(IEnumerable<RefreshInstruction> instructions, string localIdentity);
+
+        /// <summary>
+        /// Creates one or more cache instruction records base on the configured batch size from a set of individual instructions and saves them.
```
nikolajlauridsen, line 32:
```suggestion
        /// Creates one or more cache instruction records based on the configured batch size from a set of individual instructions and saves them.
```

Very minor typo

src/Umbraco.Core/Services/ICacheInstructionService.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        void DeliverInstructionsInBatches(IEnumerable<RefreshInstruction> instructions, string localIdentity);
+
+        /// <summary>
+        /// Processes and then prunes pending database cache instructions.
+        /// </summary>
+        /// <param name="released">Flag indicating if process is shutting now and operations should exit.</param>
+        /// <param name="localIdentity">Local local identity of the executing AppDomain.</param>
```
nikolajlauridsen, line 40:
```suggestion
        /// <param name="localIdentity">Local identity of the executing AppDomain.</param>
```

Same here

src/Umbraco.Infrastructure/Services/Implement/CacheInstructionService.cs
AndyButland: No, it's not - have removed.

src/Umbraco.Infrastructure/Services/Implement/CacheInstructionService.cs
AndyButland: Probably.  I didn't add it, but agree it doesn't seem to be useful.  Will remove.

AndyButland:
Thanks @nikolajlauridsen - I see your point.  Am unsure whether to change though, just as I've followed the pattern established here for another "messages" style repository: `NotificationsRepository`.  And also as the `CacheInstructionRepository` is internal, so we only have to worry about core developers using it "wrong", and not packages or Umbraco developers.

So think either should leave as is, or should update - but also do the the `NotificationsRepository` too in the same way.

I think these are the only two - most of the rest inherit from `EntityRepositoryBase<TId, TEntity>` which uses a different pattern, but it seemed more than was needed for this requirement (i.e. implementations for all the CRUD operations that we don't need in this case).

Will take your advice on the best way to proceed...

nikolajlauridsen:
I see what you mean with them being internal, so it's not a huge issue. I think the best way to proceed would be to change both the `CacheInstructionRepository` and the `NotificationsRepository`, especially since there are only two repositories to change, so it might not be a bad time to nip it in the bud.

But if you're pressed for time, I wouldn't mind merging this as is, and maybe changing the repos in a separate task?

AndyButland:
I think let's do it here - as you say, makes sense to get it right early.  I'll have a bit of time later so can do that.

AndyButland:
Actually... it's not so easy.  We'd need to move `IScope` into the "Core" project for it to be usable from `INotificationRepository/ICacheInstructionRepository`, and we can't very easily do that as it has other dependencies that are persistence related (`IUmbracoDatabase` that inherits from `NPoco.IDatabase`.

We could perhaps move these repository interfaces into "Core"?  Doesn't feel quite right but think it would work dependency-wise.  What do you think?

nikolajlauridsen:
Hmm I see what you mean, that's a tricky one. I think moving the interface into the "Core" namespace in the infrastructure could work, but also that it doesn't feel right. It's really a question of which is the lesser evil.

AndyButland:
I've updated the repository layer interface to match the suggestion now.  We actually already have a few repository interfaces in "Infrastructure", so think moving these two is the best option here.

src/Umbraco.Core/Models/CacheInstruction.cs
```diff
@@ -0,0 +1,51 @@
...
+            UtcStamp = utcStamp;
+            Instructions = instructions;
+            OriginIdentity = originIdentity;
+            InstructionCount = instructionCount;
+        }
+
+        /// <summary>
+        /// Cache instruction Id.
+        /// </summary>
+        public int Id { get; private set; }
```
Shazwazza, line 28: Do these need to be private set since they are only set in the ctor? Could just be { get; } ?

src/Umbraco.Core/Services/CacheInstructionServiceProcessInstructionsResult.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <summary>
+    /// Defines a result object for the <see cref="ICacheInstructionService.ProcessInstructions(bool, DateTime)"/> operation.
+    /// </summary>
+    public class CacheInstructionServiceProcessInstructionsResult
+    {
+        private CacheInstructionServiceProcessInstructionsResult()
+        {
+        }
+
+        public int NumberOfInstructionsProcessed { get; private set; }
```
Shazwazza, line 14: Same thing here, if these are readonly they should just be { get; }

Shazwazza (commented):
Hi all, I'm just not so sure we want to go changing the pattern for repositories now. I mean, I get the argument that we're relying on the scope being created in the service, but that is how 100% of things work today. It also means that methods don't have a dependency on the IScope like you've discovered and with that it means repo interfaces could probably move to core. 

At some stage, we will re-design the service/repo layer (its not great). But I think having inconsistent approaches can lead to further confusion. When we fix this all, we will fix it all the way through. 

Thoughts? //cc @bergmania

bergmania:
> Hi all, I'm just not so sure we want to go changing the pattern for repositories now. I mean, I get the argument that we're relying on the scope being created in the service, but that is how 100% of things work today. It also means that methods don't have a dependency on the IScope like you've discovered and with that it means repo interfaces could probably move to core.
> 
> At some stage, we will re-design the service/repo layer (its not great). But I think having inconsistent approaches can lead to further confusion. When we fix this all, we will fix it all the way through.
> 
> Thoughts? //cc @bergmania

I also think we should follow the pattern that is started. But when we do a re-design of the database, this is something I would like to change to be more explict.

AndyButland:
Happy to revert this last change - @nikolajlauridsen and I were debating which is the best way to go.  Just to be clear though, there's already really two patterns for repositories:

- Most inherit from `EntityRepositoryBase<int, T>`, which provides (and means implementing) the full set of CRUD operations).
- `NotificationsRepository` (and now the new `CacheInstructionRepository`) don't have a base class, just an interface defining a small set of specific methods that are needed.

For the PR I've introduced `CacheInstructionRepository` and followed the latter pattern, as it seems a better fit with the limited data operations we need.

So is the preference to:

1. Use `EntityRepositoryBase` throughout, and implement the new repository and the existing `NotificationsRepository` in this way?
2. Or just revert the last change - keeping `NotificationsRepository` as it was and use that same lighter pattenr for `CacheInstructionRepository`?

src/Umbraco.Core/Models/CacheInstruction.cs
AndyButland: They could yes, have made that change.

src/Umbraco.Core/Services/CacheInstructionServiceProcessInstructionsResult.cs
AndyButland: This one does need to be `private set` actually as it's set not in the constructor but in the static factory methods.

AndyButland:
Have gone with 2. above - this seems best option to me and the least change to what's already there.

Shazwazza:
Yep I think 2) is fine for now. The `EntityRepositoryBase` is ugly and wish we could kill it or dissect it further to not be so bulkly. It's primarily just used for anything that stores data against umbracoNode.

nikolajlauridsen:
Sounds good, I'll merge this then 👍

PR #9952, opened 2021-03-08
Netcore: Make it possible to run some code before umbraco pipeline

bergmania:
### Details
- This PR simple adds a way to execute some pipeline before umbraco. By default we only want the `DeveloperExceptionPageMiddelware` to execute before, but it has to be configurable.

To archive this, I introduce a new `UmbracoStartupFilterOptions` that contains a delegate that is executed just before we call `IApplicationBuilder.UseUmbraco()`. 

The default implementation of this delegate is 
```cs
app =>
{
    IWebHostEnvironment env = app.ApplicationServices.GetRequiredService<IWebHostEnvironment>();
    if (env.IsDevelopment())
    {
    }
};
```

But it is easy to override, due to the `IOptions` patterns.

To change the "PreUmbracoPipeline", you can do the following in `Startup.ConfigureServices`:

```cs 
{
    ...

    services.Configure<UmbracoStartupFilterOptions>(options =>
    {
        options.PreUmbracoPipeline = app => 
        {
           // Do whatever before Umbraco runs - or just clear what we do my default. 
        };
    });
}

```

### DIFF
--- a/src/Umbraco.Web.Common/DependencyInjection/UmbracoStartupFilter.cs
+++ b/src/Umbraco.Web.Common/DependencyInjection/UmbracoStartupFilter.cs
@@ -1,6 +1,7 @@
 using System;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Options;
 using Umbraco.Extensions;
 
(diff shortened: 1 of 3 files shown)


elit0451:
Nice and clean! 💪

PR #9953, opened 2021-03-09
Fixes and Updates for DB Scope and Ambient Context leaks

Shazwazza:
Ensure that ExecutionContext.SuppressFlow() is used for any child threads that are spawned where we must not allow the ambient context to flow. This is required for any background or concurrent threads.

Ensure that UmbracoExamineIndex.PerformIndexItems uses SafeCallContext just like the PerformDeleteFromIndex to ensure that there is no scope leaks. 

## Testing

Everything should work normally and all tests pass. Pretty hard to test anything else for this one.


drummerboy46:
@Shazwazza Can you check this in Umbraco 7 as well please? We are getting suspiciously similar connectionLock and timeout issues around publishing that seem to occur at random times. We do some extra calls to Examine at point of publish so it rings true that the problem is also in Umbraco 7/Examine.

Issue #9950, opened 2021-02-12
Health check showing error and warning if `/umbraco` is being redirected


jeannieSoS:
When I run a Health Check on my Live environment I am now getting a Security warning and error:

   Error pinging the URL https://ncc-microsites.s1.umbraco.io/umbraco - 'The remote server returned an error: (404) Not Found.'

It also shows a warning under the Excessive Headers section in the Security section:

   [httpsCheckInvalidUrl]

This has only appeared in the last few weeks. Previously running a check would show all ticks and no errors.

To reproduce run a Health Check and check the Security section. 

![healthcheck](https://user-images.githubusercontent.com/15905948/107768482-6b3cc500-6d2e-11eb-90ba-0c4b6cdf220a.PNG)
![healthcheck-2](https://user-images.githubusercontent.com/15905948/107768484-6bd55b80-6d2e-11eb-81a7-79a14cc003b4.PNG)

nul800sebastiaan:
FYI I've moved this issue to the CMS tracker since it is an issue that needs to be fixed in CMS!

umbrabot:
Hi @jeannieSoS,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

umbrabot:
Hiya @jeannieSoS,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #9955, opened 2021-03-09
Fix for #9950 - HttpsCheck/ExcessiveHeaders when UmbracoID is in use

Jeavon:
Fix for #9950

Updated HttpsCheck to request background login image if ApplicationUrl returns 301/302.

Note: a HEAD request to /umbraco with UmbracoID will return a 404 and throw an exception, this is why I've changed to a GET request but have disabled auto-follow.

Changed ExcessiveHeaders to make the test request to the root instead of the /umbraco path. Also fixed a long gone localisation alias. 

Tested on Umbraco Cloud enabled with UmbracoID. Note that Cloud "Public Access" restrictions will cause 401 on both checks unless all Umbraco Cloud IPs are added to the whitelist.

### DIFF
--- a/src/Umbraco.Web/HealthCheck/Checks/Security/ExcessiveHeadersCheck.cs
+++ b/src/Umbraco.Web/HealthCheck/Checks/Security/ExcessiveHeadersCheck.cs
@@ -49,7 +49,7 @@ namespace Umbraco.Web.HealthCheck.Checks.Security
         {
             var message = string.Empty;
             var success = false;
-            var url = _runtime.ApplicationUrl;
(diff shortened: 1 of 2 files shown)


mikecp (approved):

mikecp:
Hello @Jeavon ,
This looks great, thanks for your PR!

mikecp:
Hey @Jeavon @nul800sebastiaan ,

Just thinking back on it this morning, I was wondering if there was a specific reason why we could in fact not check for the logo right away (maybe with a comment explaining the 301-302 issue). 
I might be overseeing something, but if we know that the logo check works in all cases, what would be the reason to start with an "uncertain check" (which is a GET and not a HEAD anymore) and then fallback on the logo if necessary?

PR #9958, opened 2021-03-10
Allows for structured logging with the IProfilingLogger

Shazwazza:
Allows for structured logging with the IProfilingLogger.

Prior to this you couldn't use structured logging. Now you can do:

```cs
using(_profilingLogger.DebugDuration<ProfilingLogger>(
                    "Hello {World}, this is a Structured {Log}",
                    new object[] { Foo, Bar })) {

        // ...

}
```

And there are other overloads for specifying start/end/failed messages and parameters. The behavior when no args are specified will execute the way it normally does.

### DIFF
--- a/src/Umbraco.Core/Logging/DisposableTimer.cs
+++ b/src/Umbraco.Core/Logging/DisposableTimer.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Diagnostics;
 using Microsoft.Extensions.Logging;
 
@@ -118,7 +118,7 @@ namespace Umbraco.Cms.Core.Logging
             {
                 if (_failed)
                 {
-                    if (_failMessageArgs == null)
+                    if (_failMessageArgs is null)
                     {
                         _logger.LogError(_failException, "{FailMessage} ({Duration}ms) [Timing {TimingId}]", _failMessage, Stopwatch.ElapsedMilliseconds, _timingId);
                     }
@@ -144,8 +144,8 @@ namespace Umbraco.Cms.Core.Logging
                             {
                                 var args = new object[_endMessageArgs.Length + 2];
                                 _endMessageArgs.CopyTo(args, 0);
-                                args[_endMessageArgs.Length - 1] = Stopwatch.ElapsedMilliseconds;
-                                args[_endMessageArgs.Length] = _timingId;
+                                args[args.Length - 1] = Stopwatch.ElapsedMilliseconds;
+                                args[args.Length] = _timingId;
                                 _logger.LogDebug(_endMessage + " ({Duration}ms) [Timing {TimingId}]", args);
                             }
                             break;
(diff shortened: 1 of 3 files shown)


src/Umbraco.Core/Logging/DisposableTimer.cs
```diff
@@ -85,19 +118,55 @@ protected override void DisposeResources()
...
+                    if (_failMessageArgs == null)
+                    {
+                        _logger.LogError(_failException, "{FailMessage} ({Duration}ms) [Timing {TimingId}]", _failMessage, Stopwatch.ElapsedMilliseconds, _timingId);
+                    }
+                    else
+                    {
+                        var args = new object[_failMessageArgs.Length + 2];
+                        _failMessageArgs.CopyTo(args, 0);
+                        args[_failMessageArgs.Length - 1] = Stopwatch.ElapsedMilliseconds;
+                        args[_failMessageArgs.Length] = _timingId;
```
nikolajlauridsen, line 130: Seems like we're doing more or less the same thing three times, maybe wrap this in a method? Maybe just pass in the  `object[]`, and it can be used for both failed and succeeded.

src/Umbraco.Core/Logging/DisposableTimer.cs
```diff
@@ -85,19 +118,55 @@ protected override void DisposeResources()
             {
                 if (_failed)
                 {
-                    _logger.LogError(_failException, "{FailMessage} ({Duration}ms) [Timing {TimingId}]", _failMessage, Stopwatch.ElapsedMilliseconds, _timingId);
+                    if (_failMessageArgs == null)
```
nikolajlauridsen, line 121: Very minor thing but I think that `is null` is more readable that `== null`, however looking it up the Roslyn compiler has been updated so the performance is the same

nikolajlauridsen (commented):
Looks good 👍 I found two very minor potential improvements.

src/Umbraco.Core/Logging/DisposableTimer.cs
```diff
@@ -85,19 +118,55 @@ protected override void DisposeResources()
...
+                            if (_endMessageArgs == null)
+                            {
+                                _logger.LogDebug("{EndMessage} ({Duration}ms) [Timing {TimingId}]", _endMessage, Stopwatch.ElapsedMilliseconds, _timingId);
+                            }
+                            else
+                            {
+                                var args = new object[_endMessageArgs.Length + 2];
+                                _endMessageArgs.CopyTo(args, 0);
+                                args[_endMessageArgs.Length - 1] = Stopwatch.ElapsedMilliseconds;
+                                args[_endMessageArgs.Length] = _timingId;
```
nikolajlauridsen, line 148: While doing some testing I noticed that the log output was unexpected, shouldn't it be `args[args.Length - 1]` and not `args[_endMessageArgs.Length - 1]`?

For instance I provided one argument, so `Stopwatch.ElapsedMilliseconds` became the parameter in my message, and the `_timingId` became the `Duration` and `TimingId` was set to null.

nikolajlauridsen (changes requested):
I found one issue that should be resolved before we merge this, I added it inline but forgot to request changes when submitting oops

Nikkelmann:
@Shazwazza
Looking at the code, it seems like `TraceDuration` is not passing the args along to the `DisposableTimer`.

PR #9960, opened 2021-03-10
Fix dead lock with TypeLoader

Shazwazza:
This deadlock can occur in both v8 and the netcore branches, this fix is for v8 (the netcore fix is #9959)

This is why:

image

There is actually no reason at all for GetFileBasePath to use the _locko lock and I believe it was used by accident and just re-using an existing lock field. The GetFileBasePath method is actually just a lazy initializer for a value so that's what it should be and that is how this is fixed. Now we have a Lazy _fileBasePath which is initialized with GetFileBasePath and then _fileBasePath.Value is used where needed and this fixes the deadlock.

### DIFF
--- a/src/Umbraco.Core/Composing/TypeLoader.cs
+++ b/src/Umbraco.Core/Composing/TypeLoader.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


mikecp (approved):

mikecp:
Hi @Shazwazza, it's doing the trick indeed! Thanks!

mikecp:
Woops, can't merge this one 😄

mikecp:
@Shazwazza any issue with retargeting this to v8/contrib?

Shazwazza:
@mikecp I've updated to v8/contrib 👍

PR #9961, opened 2021-03-10
Allow for having appsettings.local.json

Shazwazza:
This file is not checked in and can be used to overwrite any custom settings that are in the repo without worrying about accidentally checking in developer settings as you work.

This also fixes up a couple other very minor settings.

How does it work?

Just create a file in the root of the Umbraco.Web.UI.NetCore project called: `appsettings.local.json` and it will automagicaly just be put underneath the appsettings.json.

Then any settings you apply here overwrite the defaults. This means you can have your own persistent connection string in there without having to worry about checking it in or merging when pulling, etc...

For example, mine looks like:

```json
{
  "ConnectionStrings": {
    "umbracoDbDSN": "server=MyServer;database=MyDatabase;user id=MyUser;password='MyPassword'"
  },
  "Serilog": {
    "MinimumLevel": {
      "Default": "Information",
      "Override": {
        "Microsoft": "Warning",
        "Microsoft.Hosting.Lifetime": "Information",
        "System": "Warning",
        "Umbraco.Cms.Core.Scoping.ScopeProvider": "Debug",
        "Umbraco.Cms.Core.Scoping.Scope": "Debug"
      }
    }
  },
  "Umbraco": {
    "CMS": {
      "Core": {
        "Debug": {
          "LogIncompletedScopes":  true
        }
      },
      "RuntimeMinification": {
        "dataFolder": "Umbraco/TEMP/Smidge",
        "version": "637429827442056840"
      }
    }
  }
}
```

This was also checked into the WIP docker branch https://github.com/umbraco/Umbraco-CMS/pull/9743/files ... but I've separated that out now into it's own PR. I've been running locally with this for a long time now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -201,3 +201,4 @@ src/Umbraco.Tests/TEMP/
 
 /src/Umbraco.Web.UI/config/umbracoSettings.config
 /src/Umbraco.Web.UI.NetCore/Umbraco/models/*
+/src/Umbraco.Web.UI.NetCore/appsettings.local.json
--- a/.gitignore
+++ b/.gitignore
@@ -201,4 +201,4 @@ src/Umbraco.Tests/TEMP/
 
 /src/Umbraco.Web.UI/config/umbracoSettings.config
 /src/Umbraco.Web.UI.NetCore/Umbraco/models/*
-/src/Umbraco.Web.UI.NetCore/appsettings.local.json
+/src/Umbraco.Web.UI.NetCore/appsettings.Local.json
--- a/src/Umbraco.Web.UI.NetCore/Program.cs
+++ b/src/Umbraco.Web.UI.NetCore/Program.cs
@@ -1,4 +1,5 @@
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 
@@ -7,18 +8,20 @@ namespace Umbraco.Cms.Web.UI.NetCore
     public class Program
     {
         public static void Main(string[] args)
-        {
-            CreateHostBuilder(args)
+            => CreateHostBuilder(args)
                 .Build()
                 .Run();
-        }
 
         public static IHostBuilder CreateHostBuilder(string[] args) =>
             Host.CreateDefaultBuilder(args)
-                .ConfigureLogging(x =>
-                {
-                    x.ClearProviders();
-                })
-                .ConfigureWebHostDefaults(webBuilder => { webBuilder.UseStartup<Startup>(); });
+#if DEBUG
+                .ConfigureAppConfiguration(config
+                    => config.AddJsonFile(
+                            "appsettings.Local.json",
+                            optional: true,
+                            reloadOnChange: true))
+#endif
+                .ConfigureLogging(x => x.ClearProviders())
+                .ConfigureWebHostDefaults(webBuilder => webBuilder.UseStartup<Startup>());
     }
 }
--- a/src/Umbraco.Web.UI.NetCore/appsettings.json
+++ b/src/Umbraco.Web.UI.NetCore/appsettings.json
@@ -38,7 +38,7 @@
         "ConvertUrlsToAscii": "try"
       },
       "RuntimeMinification": {
-        "dataFolder": "Umbraco/Data/TEMP/Smidge",
+        "dataFolder": "umbraco/Data/TEMP/Smidge",
         "version": "1"
       },
       "Security": {
--- a/src/Umbraco.Web.UI.NetCore/appsettings.json
+++ b/src/Umbraco.Web.UI.NetCore/appsettings.json
@@ -38,7 +38,7 @@
         "ConvertUrlsToAscii": "try"
       },
       "RuntimeMinification": {
-        "dataFolder": "Umbraco/TEMP/Smidge",
+        "dataFolder": "Umbraco/Data/TEMP/Smidge",
         "version": "1"
       },
       "Security": {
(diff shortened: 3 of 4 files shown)


src/Umbraco.Web.UI.NetCore/appsettings.json
```diff
@@ -38,8 +38,8 @@
         "ConvertUrlsToAscii": "try"
       },
       "RuntimeMinification": {
-        "dataFolder": "App_Data/TEMP/Smidge",
-        "version": "637432008251409860"
+        "dataFolder": "Umbraco/TEMP/Smidge",
```
stevetemple, line 41: Is moving this to under Umbraco intentional?

stevetemple (commented):
LGTM, one question

stevetemple:
I'd also suggest something in a readme/other doc for this?

src/Umbraco.Web.UI.NetCore/appsettings.json
Shazwazza: Yes, but actually this should be `Umbraco/Data/TEMP/Smidge`. App_Data is not a thing in netcore. Each folder outside of www is essentially a form of 'app data' :) I'll update the path.

Shazwazza, line 41:
```suggestion
        "dataFolder": "Umbraco/Data/TEMP/Smidge",
```

bergmania:
We need to figure out what to do with the `Program.cs`. Currently, we copy the one we have in `Umbraco.Web.UI.Netcore` into the dotnet template that we publish. Do we want this change to "leak" into the Umbraco dotnet template?

Also, I'm in doubt about whether I like this file to be prioritised over or under the appsetting.json.

If it is prioritised under (`appSetting.json` overrides `appSetting.local.json`), then `appSetting.local.json` will be written into when you install or do any changes in Umbraco that write to config, because we write in the JSON source with the lowest priority. I like that part.
On the other hand, for all settings I provide manually I would like `appSetting.local.json` to override `appSetting.json`. 🤯

benjaminc:
I think `appsettings.local.json` should be applied after `appsettings.json` so that your local config overrides anything that is committed.  I've done this with a few other .NET Core projects recently as well, and it works really nicely to have the option to override any setting locally without worrying about it getting checked in.

If I understand your concern though Bjarke, if we have `appsettings.local.json` applied later, then changes Umbraco makes (such as setting the connection string during an install) would be put into the main `appsettings.json` file instead of in the local file, correct?  That seems like the right behavior to me.  The default should be having a project that just works for the most basic use case (a single developer, or a small team all working with the same DB), and then provide options for more advanced scenarios.  The local file just gives a developer the option to apply local settings without checking them in, and without having to add this change to every Program.cs file they work with so that they have the option.

bergmania:
> If I understand your concern though Bjarke, if we have `appsettings.local.json` applied later, then changes Umbraco makes (such as setting the connection string during an install) would be put into the main `appsettings.json` file instead of in the local file, correct?

Yep..

> That seems like the right behavior to me. The default should be having a project that just works for the most basic use case (a single developer, or a small team all working with the same DB), and then provide options for more advanced scenarios. The local file just gives a developer the option to apply local settings without checking them in, and without having to add this change to every Program.cs file they work with so that they have the option.

Cool, so you think it is a good idea that this is "leaking" into our Umbraco dotnet templates, correct?

benjaminc:
> Cool, so you think it is a good idea that this is "leaking" into our Umbraco dotnet templates, correct?

Yes, I think having this would be a nice thing for developers who have different local DBs, or other local settings they don't want checked in.

.gitignore
```diff
@@ -201,3 +201,4 @@ src/Umbraco.Tests/TEMP/
 
 /src/Umbraco.Web.UI/config/umbracoSettings.config
 /src/Umbraco.Web.UI.NetCore/Umbraco/models/*
+/src/Umbraco.Web.UI.NetCore/appsettings.local.json
```
benjaminc, line 204: @bergmania Does the .gitignore file also get pushed into the template, or is this change something that developers would need to know to add?

src/Umbraco.Web.UI.NetCore/Program.cs
```diff
@@ -7,18 +8,20 @@ namespace Umbraco.Cms.Web.UI.NetCore
...
             Host.CreateDefaultBuilder(args)
-                .ConfigureLogging(x =>
-                {
-                    x.ClearProviders();
-                })
-                .ConfigureWebHostDefaults(webBuilder => { webBuilder.UseStartup<Startup>(); });
+#if DEBUG
+                .ConfigureAppConfiguration(config
+                    => config.AddJsonFile(
+                            "appsettings.Local.json",
```
benjaminc, line 20: @Shazwazza Any reason to use a capital L?  I think it would only matter for Linux developers, but they are also the ones who are notoriously lazy about capitalization.

bergmania: IMO capital L is correct. It follows the convention from: `appsettings.Production.json` and `appsettings.Development.json`.

.gitignore
bergmania: Currently not

src/Umbraco.Web.UI.NetCore/Program.cs
Shazwazza: Yep but i guess the .gitignore doesn't have the correct case, i'll fix

.gitignore
Shazwazza, line 204:
```suggestion
/src/Umbraco.Web.UI.NetCore/appsettings.Local.json
```

PR #9966, opened 2021-03-10
Netcore: Get rid of default files in config folder

bergmania:
### Reason
- /config folder content is overridden on build, when using the nuget package

### Details
This PR get rid of the folder by
- `*.user.xml` language files
   -  Remove them. All of them were empty, we can document how to add these files manually. All of them was empty, but if you add them manually in the same location Umbraco pick up the info.
- `grid.editor.config.js`
   - Embedded the default file. You can manually add the file in the same location, and that will be used instead of the embedded file
- `logviewer.searches.config.js`
   - This was basically a database in a file. Migrated to use a real sql table instead.
   - Added migrations, that reads from the file if it exists, otherwise insert default data (same content as file before)
   - NOTE: The service etc support add of new saved log queries, but the UI do not support that (Of of scope for this task)
-`tinyMceConfig.config`
  - Migrated to use `appSettings.json` and default values in code
- Tours
  -  Embedded the default tours. Plugins can still add more tours.

### Bonus
- Fixed issue when saving a grid with a media cell, without any media picked. Before an exception was thrown from the code that persists relations between media and content for grids

### Test
- Add a `*.user.xml` and check it is still picked up
- Use the grid
- Use the log viewer and the saved queries. Also try to remove one.
  - Try to update an existing database e.g. netcore db created from other branch or v8 db
    - If the file exists in the old location, it should add these to the db (if the db is empty)
    - f the file do not exists (e.g. from this PR, when the file is removed from git), default should be installed.
  - Install new umbraco from this branch
    - Expect the default saved queries to be installed
- Use the RTE
- Try the default tours..


src/Umbraco.Web.BackOffice/Controllers/TourController.cs
```diff
@@ -57,15 +59,22 @@ public IEnumerable<BackOfficeTourFile> GetTours()
             var nonPluginFilters = _filters.Where(x => x.PluginName == null).ToList();
 
             //add core tour files
-            var coreToursPath = Path.Combine(_hostingEnvironment.MapPathContentRoot(Constants.SystemDirectories.Config), "BackOfficeTours");
-            if (Directory.Exists(coreToursPath))
+            // var coreToursPath = Path.Combine(_hostingEnvironment.MapPathContentRoot(Constants.SystemDirectories.Config), "BackOfficeTours");
+            // if (Directory.Exists(coreToursPath))
+            // {
+            //
+            // }
```
elit0451, line 66: Do we still need these comments?

bergmania:
```suggestion

```

Nope

elit0451:
Tests didn't reveal any strange behaviour 🙌 so I will be merging!

PR #9968, opened 2021-03-11
New table and service to store auth tokens

Shazwazza:
Implements the auth token retrieval and storage for the identity implementation. This is now automatically done for providers and the back office user manager can be used to retreive and store all tokens.

Fixes locking on the config writer.

Removes the abstract NodeObjectTypeId from the base repo since it shouldn't be there.

## Testing

I have definitely tested that this passes tokens and stores them, it's a bit of a pain to set that up and test but it works :) 

Since this has some migrations, it will launch the upgrader so you'll want to make sure that all works.

Tests are passing 👍


src/Umbraco.Infrastructure/Security/UmbracoIdentityUser.cs
```diff
@@ -180,6 +182,37 @@ public ICollection<IIdentityUserLogin> Logins
...
 
+        /// <summary>
+        /// Gets the external login tokens collection
+        /// </summary>
+        public ICollection<IIdentityUserToken> LoginTokens
+        {
+            get
+            {
+                // return if it exists
+                if (_tokens != null)
```
bergmania, line 193:
```suggestion
                if (_tokens is not null)
```

src/Umbraco.Infrastructure/Security/UmbracoIdentityUser.cs
```diff
@@ -180,6 +182,37 @@ public ICollection<IIdentityUserLogin> Logins
...
+                {
+                    return _tokens;
+                }
+
+                _tokens = new ObservableCollection<IIdentityUserToken>();
+
+                // if the callback is there and hasn't been created yet then execute it and populate the logins
+                if (_getTokens != null && !_getTokens.IsValueCreated)
```
bergmania, line 201:
```suggestion
                if (_getTokens?.IsValueCreated != true)
```

src/Umbraco.Tests.Integration/Umbraco.Infrastructure/Services/ExternalLoginServiceTests.cs
```diff
@@ -202,6 +203,31 @@ public void Add_Logins()
             }
         }
 
+        [Test]
+        public void Add_Tokens()
+        {
+            var user = new User(GlobalSettings, "Test", "test@test.com", "test", "helloworldtest");
```
bergmania, line 209: I know it is just copy/pasted, but when we don't use any of this dummy parameters, we really should use our builder. This suggestion can be done for the entire test class

```suggestion
            var user = new UserBuilder().Build();
```

bergmania (approved):
Looks good.. Only found small things I will fix up for you.

bergmania:
@Shazwazza , if you are happy with my changes, feel free to merge.. My tests did not find anything unexpected.

Shazwazza:
@bergmania awesome 💪

PR #9970, opened 2021-03-11
Netcore: Reintroduce friendly bun untestable published content extensions

bergmania:
### Details
- Introduce an internal service location 🤯 
- Use that service location in extension methods so views can easily be migrated.

### Test
- Use the normal extension methods that could be used in v8.
- Use the query builder to generate queries, and ensure they work.

### DIFF
--- a/src/Umbraco.Web.Common/DependencyInjection/StaticServiceProvider.cs
+++ b/src/Umbraco.Web.Common/DependencyInjection/StaticServiceProvider.cs
@@ -11,7 +11,7 @@ namespace Umbraco.Cms.Web.Common.DependencyInjection
     /// 1) Continue to have the same extension methods on IPublishedContent and IPublishedElement as in V8. To make migration easier.
     /// 2) To have a tool to avoid breaking changes in minor versions. All methods using this should in theory be obsolete.
     ///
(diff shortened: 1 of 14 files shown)


src/Umbraco.Web.Common/DependencyInjection/StaticServiceProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    /// <summary>
+    /// INTERNAL Service locator. Should only be used if no other ways exist.
+    /// </summary>
+    /// <remarks>
+    /// It is created with only two goals in mind
+    /// 1) Continue to have the same extension methods on IPublishedContent and IPublishedElement as in V8. To make migration easier.
+    /// 2) To have a tool to avoid breaking changes in minor versions. All methods using this should in theory be obsolete.
+    ///
+    /// Keep in mind, everything this is used, the code becomes basically untestable.
```
AndyButland, line 14:
```suggestion
    /// Keep in mind, every time this is used, the code becomes basically untestable.
```

AndyButland (approved):
Looks good to me (just from reviewing the code, I haven't tested).  Just fixed a typo.

nikolajlauridsen:
Gave it a testing, and all looks good now 👍

PR #9973, opened 2021-03-12
Fix issue with SqlMainDomLock that cannot use implicit lock timeouts …

bergmania:
### Details
Fixes issue with `SqlMainDomLock ` that need to explicit pass in Lock timeouts, because `Current.Config` is not available that early in the boot process..

This PR specify the timeout explicitly by reading the config. The part of reading the config is made shareable.

### Test:
- Boot umbraco with the following appSetting in web.config
 `<add key="Umbraco.Core.MainDom.Lock" value="SqlMainDomLock" />`
   - Before this PR:
image
   - With this PR you can boot correctly

### DIFF
--- a/src/Umbraco.Core/Configuration/GlobalSettings.cs
+++ b/src/Umbraco.Core/Configuration/GlobalSettings.cs
@@ -409,14 +409,14 @@ namespace Umbraco.Core.Configuration
             {
                 if (_sqlWriteLockTimeOut != default) return _sqlWriteLockTimeOut;
 
-                var timeOut = GetSqlWriteLockTimeoutFromConfigFile();
+                var timeOut = GetSqlWriteLockTimeoutFromConfigFile(Current.Logger);
 
                 _sqlWriteLockTimeOut = timeOut;
                 return _sqlWriteLockTimeOut;
             }
         }
 
-        internal static int GetSqlWriteLockTimeoutFromConfigFile()
+        internal static int GetSqlWriteLockTimeoutFromConfigFile(ILogger logger)
         {
             var timeOut = 5000; // 5 seconds
             var appSettingSqlWriteLockTimeOut = ConfigurationManager.AppSettings[Constants.AppSettings.SqlWriteLockTimeOut];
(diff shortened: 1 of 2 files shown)


src/Umbraco.Core/Configuration/GlobalSettings.cs
```diff
@@ -409,26 +409,34 @@ public int SqlWriteLockTimeOut
...
+                const int minimumTimeOut = 100;
+                const int maximumTimeOut = 20000;
+                if (configuredTimeOut >= minimumTimeOut && configuredTimeOut <= maximumTimeOut) // between 0.1 and 20 seconds
+                {
+                    timeOut = configuredTimeOut;
+                }
+                else
+                {
+                    Current.Logger.Warn<GlobalSettings>(
+                        $"The `{Constants.AppSettings.SqlWriteLockTimeOut}` setting in web.config is not between the minimum of {minimumTimeOut} ms and maximum of {maximumTimeOut} ms, defaulting back to {timeOut}");
```
nul800sebastiaan, line 435: This is also using `Current`, luckily it just fails silently when using `SqlMainDom` but can we do this some other way so that it still logs?

bergmania: Fixed :)

bjarnef:
I had a similar error in an Umbraco v8.9.0 project hosted on Azure, where adding the `SqlMainDomLock` as recommended on Azure Web Apps. Also I don't had issues with this locally (probably because Umbraco already was installed), but when colleague forked + cloned the project locally and cleared the `Umbraco.Core.ConfigurationStatus` the booting failed.
Removing the `SqlMainDomLock` key from web.config locally fixed this and the installler was running successfully.
Not sure it is specific related to this, but worth mentioning 😁

src/Umbraco.Core/Configuration/GlobalSettings.cs
nul800sebastiaan: Yes! Now it works, thanks! 👍

nul800sebastiaan:
It shouldn't be @bjarnef since this timeout setting was introduced in 8.12.0. 

Not sure what that error was but I don't think SqlMainDom should be used locally, so would recommend to add a config transform during your release process so it only gets added on the live environments.

Shazwazza:
It is perfectly safe to use locally too 👍 but older versions have other sqlmaindom issues, so just best to upgrade to vlatest

PR #9977, opened 2021-03-15
Made recurring hosted service base class PerformExecuteAsync method public.

AndyButland:
This was raised in a comment by a community member, and seems correct.  We can make this method public allowing package developers and/or custom solutions creating hosted services to use our base class.


bergmania (approved):

bergmania:
Make sense..

PR #9980, opened 2021-03-16
Add gulp comments and sourcemap dependencies

kjac:
### Prerequisites

### Description

This PR fixes the comments made by @Shazwazza on #9839

### Testing this PR

Run the client build and boot up the backoffice with devtools open. You should no longer see warnings in the console about missing source maps - as-is you'll see something like this:

image


bergmania (approved):
Works as expected 💪

PR #9984, opened 2021-03-16
Remove unnecessary aria-hidden attribute from umb-icon

bjarnef:
### Prerequisites

### Description
This PR just removes an unnecessary `aria-hidden` on `<umb-icon>` introduced in https://github.com/umbraco/Umbraco-CMS/pull/9264 since it is already taken care of in the component itself https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/components/umb-icon.html#L1

In fact it was the only place I could find where this was added to `<umb-icon>`.


nathanwoulfe (approved):

nathanwoulfe:
Thanks @bjarnef, small wins all add up 👍

PR #9987, opened 2021-03-16
Add images in grid - fixes 9982

### LINKED ISSUE #9982: Cannot add image to Grid in 8.12
After upgrading a site from 8.11.1 to 8.12, I can no longer add an image to a Grid editor using the standard "Image" grid control.

## Umbraco version

- 8.12.0
- 8.12.1

Reproduction
------------

### Specifics

After selecting an image, another infinity editor opens "Edit selected media".
Clicking Submit causes a JS error in `umbraco.controllers.js` `function updateControlValue(selectedImage)` on line 18732 in 8.12.1:

Error: "$scope.control.value is null"

### Steps to reproduce

- Edit a content item with a Grid editor
- Add an `Image` control
- Click "Click to insert image"
- Select an image
- Another infinity editor pops up "Edit selected media"
- Click "Select"
- Nothing happens (apart from a console JS error)

### Expected result

Image added to the grid

### Actual result

Image isn't selected, I can only "Close" the dialog

nathanwoulfe:
access control values after reassigning - check for regenerating the thumbnail needs to happen after the new value has been assigned to the control

If there's an existing issue for this PR then this fixes #9982


nul800sebastiaan:
_whoops, wrong button_ 

Thanks @nathanwoulfe, seems like we're now missing something though, the image doesn't preview after adding it: 


https://user-images.githubusercontent.com/304656/111453302-b70ad180-8713-11eb-9e2b-82946c750dcd.mp4

Not sure what's not being passed back, will investigate.

nul800sebastiaan:
`$scope.control.value.image` and `selectedImage.image` are always the same, even after picking a different image, same for `$scope.control.value.focalPoint` and `selectedImage.focalPoint` even after changing the focal point. It's because they were updated in the scope just before that, so updating the thumbnail URL needs to be done first before setting the values on the `$scope.control`.

nul800sebastiaan:
I don't see a way out of this, as far as I can tell we don't have access to the values before everything changes in `updateControlValue` so we can't "only" do `getThumnailUrl()` if the value changed. I suggest always recalculating the thumbnail:

```javascript
        function updateControlValue(selectedImage) {
            // we could apply selectedImage directly to $scope.control.value,
            // but this allows excluding fields in future if needed
            $scope.control.value = {
                focalPoint: selectedImage.focalPoint,
                coordinates: selectedImage.coordinates,
                id: selectedImage.id,
                udi: selectedImage.udi,
                image: selectedImage.image,
                caption: selectedImage.caption,
                altText: selectedImage.altText
            };
            $scope.thumbnailUrl = getThumbnailUrl();
        }
```

This works.. but it might be possible to get the old values somehow, my brain just doesn't let me see how! 😅

nathanwoulfe:
Yeah, that makes sense - I don't think it's a huge deal regenerating the thumbnail each time. If it works, it works.

nul800sebastiaan:
Thanks for the feedback! Fixed that one up merging now 👍

PR #9989, opened 2021-03-16
Revert "The Value() method for IPublishedContent was not working with the defaultValue parameter"

mikecp:
Partially reverts umbraco/Umbraco-CMS#9888

PR #9888 introduced an unintended breaking change, so reverting it but with keeping one change: we send the fallback + default value parameter to the inner Value method.

### DIFF
--- a/src/Umbraco.Web/PublishedContentExtensions.cs
+++ b/src/Umbraco.Web/PublishedContentExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Linq;
@@ -174,7 +174,7 @@ namespace Umbraco.Web
 
             // else... if we have a property, at least let the converter return its own
             // vision of 'no value' (could be an empty enumerable) - otherwise, default
-            return property == null ? default : property.Value<T>(culture, segment);
+            return property == null ? default : property.Value<T>(culture, segment, fallback, defaultValue);
         }
 
         #endregion
--- a/src/Umbraco.Web/PublishedContentExtensions.cs
+++ b/src/Umbraco.Web/PublishedContentExtensions.cs
@@ -173,8 +173,8 @@ namespace Umbraco.Web
                 return value;
 
             // else... if we have a property, at least let the converter return its own
-            // vision of 'no value' (could be an empty enumerable) - otherwise, defaultValue
-            return property == null ? defaultValue : property.Value<T>(culture, segment, defaultValue: defaultValue);
+            // vision of 'no value' (could be an empty enumerable) - otherwise, default
+            return property == null ? default : property.Value<T>(culture, segment);
         }
 
         #endregion
--- a/src/Umbraco.Web/PublishedPropertyExtension.cs
+++ b/src/Umbraco.Web/PublishedPropertyExtension.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using Umbraco.Core;
 using Umbraco.Core.Composing;
 using Umbraco.Core.Models.PublishedContent;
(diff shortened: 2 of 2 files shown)


nul800sebastiaan:
Thanks @mikecp, looking good now! ⭐

src/Umbraco.Web/PublishedContentExtensions.cs
```diff
@@ -173,8 +173,8 @@ public static T Value<T>(this IPublishedContent content, string alias, string cu
                 return value;
 
             // else... if we have a property, at least let the converter return its own
-            // vision of 'no value' (could be an empty enumerable) - otherwise, defaultValue
-            return property == null ? defaultValue : property.Value<T>(culture, segment, defaultValue: defaultValue);
+            // vision of 'no value' (could be an empty enumerable) - otherwise, default
+            return property == null ? default : property.Value<T>(culture, segment, fallback, defaultValue);
```
rickbutterfield, line 177: @nul800sebastiaan @mikecp Just upgraded to 8.13-rc and was trying to debug why I was seeing the following issue trying to get a fallback value...
> Fallback PublishedValueFallback does not support fallback code '3' at property level.

I'm not convinced that on line 177 we should be passing `fallback` into the `property.Value<T>()` method as we've tried to get the fallback on line 172 and presumably it's failed?

nul800sebastiaan: @rickbutterfield Are you saying this is a breaking change for you?
I will need to look into this later but if this is breaking then please let us know what happened before that is now different.

rickbutterfield: Yes I believe it is a breaking change. I current have this code in an extension method for generated model `IPageSettings`

```csharp
public static string GetMetaColour(this IPageSettings pageSettings)
{
    if (!string.IsNullOrEmpty(pageSettings.MetaColour))
    {
        return pageSettings.MetaColour;
    }

    return pageSettings.Value<string>("metaColour", fallback: Fallback.ToAncestors);
}
```

That last return value now shows the error I've listed above on every page I try to visit. 

I've managed to change 

```csharp
return property == null ? default : property.Value<T>(culture, segment, fallback, defaultValue);
```

to

```csharp
return property == null ? default : property.Value<T>(culture, segment, defaultValue: defaultValue);
```

locally and it resolves the issue.

Happy to open a new issue if that is easier.

jaddie: I can confirm that I have just made this change locally, I have also debugged & tested the issue and can confirm this was not an issue with 8.12.2, however running the latest codebase this now breaks the site. Removing the fallback parameter from line 177 of PublishedContentExtensions.cs in Umbraco.Web project solves the problem & the site functions as intended.

nul800sebastiaan: @jaddie, just making THIS change alone on an 8.12 project is not going to work, you'd have to apply https://github.com/umbraco/Umbraco-CMS/pull/9888 first, unless I'm missing something?

Thanks @rickbutterfield - we'll have to have a look at that, very helpful to have some sample code!

nul800sebastiaan: Ah ignore me @jaddie - I see you mentioned you run the latest codebase, presumably 8.13 👍

jaddie: No worries, yeah the same version as Rick is using, grabbed the latest from the repo & compiled it & replicated the above change which indeed fixed the issue :)

nul800sebastiaan:
FYI: Due to feedback about breaking changes this PR and #9888 was reverted and will not be part of 8.13 final.

PR #9990, opened 2021-03-17
Ability to have an in-memory runtime minifier cache and configurable cache busters

Shazwazza:
* Configuration based cache busters (see enum `RuntimeMinificationCacheBuster`). I've added this config to the appsettings.Development.json file so that by default when developing we use an in-memory runtime cache (no persisted files) and the cache buster is the timestamp which means there is never browser cached versions of anything.:
   ```json
      "RuntimeMinification": {
          "useInMemoryCache": true,
          "cacheBuster": "Timestamp"
        },
   ```
* Removes old composers and moves that code to UmbracoBuilder extensions
* Ensures that bundles are created lazily and only when required: the first request to either the back office or to a bundle endpoint
* Updates latest Smidge (beta 4)
* Ensures the cache buster value returned from `SmidgeRuntimeMinifier` is the value of the actual cache buster instance, not just the configured version
* Ensures that our own bundles aren't using global settings and that file watching is enabled asset files when in debug mode which automatically refreshes their caches. Though with Timestamp cache busting this doesn't really matter, but it can't hurt in case people change the development cache buster to something else, it will mean that in debug mode any changes to these files will automatically refresh cache
* Adds the same 'optimize' parameter for CSS files and as we had for JS files
* Adds a custom ISmidgeFileProvider as a composite file provider to load assets from the normal www location as well as in the App_Plugins location but ensures that no other files can be loaded from the root, this is done with the globbing file provider
* In debug mode, we will still load optimized scripts in bundles instead of individually

The same testing should be done with regards to the custom dashboard JS mentioned here https://github.com/umbraco/Umbraco-CMS/pull/9839 to ensure that this script is loaded in the back office extensions bundle.


kjac:
@Shazwazza I love everything about this ❤️ I'll give it a spin tomorrow morning. And yes, CSS minification for assets shouldn't be done runtime... I'll update the Gulp setup to handle that once this PR is merged.

I can't quite figure out if `SmidgeRuntimeMinifier` really needs both a "MinPipeline" and an "OptimizedPipeline" (`_jsMinPipeline`/`_jsOptimizedPipeline` and `_cssMinPipeline`/`_cssOptimizedPipeline` respectively). The "MinPipelines" are used for async runtime minification... if the "OptimizedPipelines" can't be used for that then it's all good. But if the "MinPipelines" can be replaced with the "OptimizedPipelines" without any extra code, they probably should be.

kjac (approved):
@Shazwazza it works like a charm. I think I'm in love with the in-memory cache, so goddamn awesome that 🚀 

This change is definitively subject to documentation at some point. Particularly how the different cache busting modes work in and out of debug mode. Perhaps you can make a note somewhere for someone to pickup at some point?

bergmania (approved):
Looks good

PR #9996, opened 2021-03-17
Events to Notifications Migration: Packaging Service

warrenbuckley:
Does the migration of Events to Notifications for Packaging Service

## Test events/notifications
Implement NotificationHandler ensure that the events/breakpoints are hit

### Example NotificationHandler

```csharp
using Umbraco.Cms.Core.Hosting;

namespace Umbraco.Cms.Core.Packaging
{
    public class PackageNotificationHandler : INotificationHandler<UninstallPackageNotification>
    {
        public void Handle(UninstallPackageNotification notification)
        {
            // Do something with the uninstalled package

            var foo = notification.UninstallationSummary;
            var bar = 5;
        }
    }

    public class PackageNotificationInstallingHandler : INotificationHandler<ImportingPackageNotification>
    {
        public void Handle(ImportingPackageNotification notification)
        {
            // What package is being installed...
            var name = notification.PackageMetaData.Name;

            if(name == "Testing")
            {
                notification.Cancel = true;
            }
        }
    }


    public class PackageNotificationInstallHandler : INotificationHandler<ImportedPackageNotification>
    {
        private IHostingEnvironment _hostingEnvironment;

        public PackageNotificationInstallHandler(IHostingEnvironment hostingEnvironment)
        {
        }
        public void Handle(ImportedPackageNotification notification)
        {
            var foo = notification.InstallationSummary;
            var bar = notification.PackageMetaData;
        }
    }
}
```

### Registering Notification Handlers
Probably not the best way to register them in our codebase but for quickness of testing, add them to `UmbracoBuilder`

```csharp
this.AddNotificationHandler<UninstallPackageNotification, PackageNotificationHandler>();
this.AddNotificationHandler<ImportingPackageNotification, PackageNotificationInstallingHandler>();
this.AddNotificationHandler<ImportedPackageNotification, PackageNotificationInstallHandler>();
```


bergmania:
Overall it looks good, but I realized a lot of files in Infrastructure should have been in Core, so I moved these. Then we could reuse them.

I also enabled state on the `ImportingPackageNotification` => `ImportedPackageNotification` as these are like Saveing vs Saved events.

Finally, I added methods on IEvenAggregator to PublishCancelable instead of having that logic in the controller, it can now be reused.

bergmania (approved):

warrenbuckley:
Thanks @bergmania off to read your changes so I follow along

PR #9999, opened 2021-03-17
Granting access to the content dashboard for all user groups

elit0451:
## Details 
Adds the `AllowContentDashboardAccessToAllUsers` appSetting in the Web.config file, so the intro dashboard ("Getting Started") is shown to all user groups, not just admins. The change will affect all new projects. In the case that the config setting is not present the old access rules will apply, therefore not changing any existing behaviour.

Additionally, more information is reported from an existing dashboard - an admin parameter is added by the end.

## Test
- Remove the **AllowContentDashboardAccessToAllUsers** entry from the Web.config or change its value to `false` and verify that only users in the admin group have access to the Content dashboard;
- Add it back or set it to `true` and create a few new users (ex: belonging to the Editors or Writers group) and check that the Content dashboard also appears (in addition to Redirect URL Management which was already there);

-----
**Note:**
SensitiveData and Translators users have different UIs so the dashboard is not present at first glance, however, they have access to it!


src/Umbraco.Web/Dashboards/ContentDashboard.cs
```diff
@@ -9,19 +12,30 @@ public class ContentDashboard : IDashboard
...
             get
             {
-                var rules = new IAccessRule[]
+                IAccessRule[] rules;
+                var dashboardConfig = Path.Combine(IOHelper.MapPath(SystemDirectories.Config), "content.dashboard.access.config.js");
+
+                if (File.Exists(dashboardConfig))
+                {
+                    var rawJson = File.ReadAllText(dashboardConfig);
+                    rules = JsonConvert.DeserializeObject<AccessRule[]>(rawJson);
```
nikolajlauridsen, line 29: Maybe we want to create a backing field for the rules? If I attach a debugger to this get property, I can see that it's called three times every time the content page is loaded, which means that we'll be reading a file from disk three times each time the content section is loaded which seems inefficient.

The major downside to this though is that the site has to rebooted before the changes take effect, so it might not work.

nikolajlauridsen:
Look good overall, I only have one concern that I've added as a comment, not sure what would be the best course of action though. 

I see what you mean with SensitiveData and Translators not seeing the dashboard since it's only shown in the content section, but I agree that it's okay as it is regardless

src/Umbraco.Web/Dashboards/ContentDashboard.cs
```diff
@@ -9,19 +12,30 @@ public class ContentDashboard : IDashboard
...
+        public string[] Sections => new[] { "content" };
 
         public string View => "views/dashboard/default/startupdashboardintro.html";
 
         public IAccessRule[] AccessRules
         {
             get
             {
-                var rules = new IAccessRule[]
+                IAccessRule[] rules;
```
bergmania, line 23: Any reason not just to allow all user groups? If the end user has added custom user groups, these are not allowed to see the dashboard now. I think a simple appsetting in webconfig is enough, just to tell all user groups are allowed to see the dashboard

src/Umbraco.Web/Dashboards/ContentDashboard.cs
elit0451: Good find @nikolajlauridsen, will refactor!

src/Umbraco.Web/Dashboards/ContentDashboard.cs
elit0451: Good idea, will refactor!

elit0451:
Have done some refactoring 😉 maybe one of you @nikolajlauridsen or @bergmania can give it a second review 🙏

src/Umbraco.Core/Configuration/GlobalSettings.cs
```diff
@@ -395,6 +395,28 @@ public bool UseHttps
...
+        ///     <c>true</c> if the dashboard is visible for all user groups; otherwise, <c>false</c>
+        ///     and the default access rules for that dashboard will be in use.
+        /// </value>
+        public bool AllowContentDashboardAccessToAllUsers
+        {
+            get
+            {
+                try
+                {
+                    return bool.Parse(ConfigurationManager.AppSettings[Constants.AppSettings.AllowContentDashboardAccessToAllUsers]);
```
nikolajlauridsen, line 411: Maybe a bit nitpicky, but maybe a `bool.TryParse()` would be better here, if the parsing fails the `out` value will be set to false, so that way the try,catch won't be necessary.

src/Umbraco.Core/Configuration/GlobalSettings.cs
```diff
@@ -395,6 +395,28 @@ public bool UseHttps
...
         }
 
+        /// <summary>
+        /// Gets a value indicating whether the content dashboard should be available to all users.
+        /// </summary>
+        /// <value>
+        ///     <c>true</c> if the dashboard is visible for all user groups; otherwise, <c>false</c>
+        ///     and the default access rules for that dashboard will be in use.
+        /// </value>
+        public bool AllowContentDashboardAccessToAllUsers
```
bergmania, line 405: This method should just live on ContentDashboardSettings instead of GlobalSettings :)

src/Umbraco.Core/Dashboards/ContentDashboardSettings.cs
```diff
@@ -0,0 +1,43 @@
...
+                {
+                    rules.Add(new AccessRule
+                    {
+                        Type = AccessRuleType.Grant,
+                        Value = userGroup.Alias
+                    });
+                }
+            }
+
+            return rules.ToArray();
```
bergmania, line 40: Could be saved in a backing field. I guess this will be requested everytime we show the dashboard

src/Umbraco.Web.UI/web.Template.config
```diff
@@ -37,6 +37,7 @@
         <add key="Umbraco.Core.TimeOutInMinutes" value="20" />
         <add key="Umbraco.Core.DefaultUILanguage" value="en-US" />
         <add key="Umbraco.Core.UseHttps" value="false" />
+        <add key="Umbraco.Core.AllowContentDashboardAccessToAllUsers" value="true"/>
```
bergmania, line 40: Is this enough to make sure it will be added to future projects, when created?

src/Umbraco.Core/Dashboards/ContentDashboardSettings.cs
bergmania: On the other hand, I think this should just be a private method on the dashboard. Seems strange to inject a service into a settings. So this setting should expose what's currently added to the GlobalSettings

bergmania (changes requested):

src/Umbraco.Web.UI/web.Template.config
elit0451: Hmm.. 
Good question, that I don't know

bergmania: @nul800sebastiaan What is needed to ensure a new setting will be part of new projects? (but not existing projects)

nul800sebastiaan: This is a setting that will only exist in new installs at the moment.

So in existing projects this will not exist. So maybe you can see if this key exists and if it does not set the default to false?

elit0451: We are already doing that in [here](https://github.com/umbraco/Umbraco-CMS/blob/56d5704167e940c540ae67864cdb5263b9b535f3/src/Umbraco.Core/Configuration/GlobalSettings.cs#L409) 🙂

nul800sebastiaan: Ah perfect, so yes: it is always false unless this appsetting exists and is explicitly set to true. 

This web.template.config contains all the defaults that new sites get, so this is perfect. 

For NuGet installs: we only add anything new when it is defined in `build\NuSpecs\tools\Web.config.install.xdt` 

So in other words, this should work as expected: only brand new sites get this new key set to true.

elit0451: Great! Thanks @nul800sebastiaan 🙌

bergmania (approved):

nikolajlauridsen (approved):
Looks good 👍. Ran through the tests again and everything works splendidly 🎉

Issue #9843, opened 2021-02-18
Starting Alpha 3 with empty database crashes on startup


benjaminc:
When starting a new site, if you have already configured a database connection pointing to an empty database, the application crashes.

Reproduction
------------

1. Create an empty SQL database
2. Create a new project using the Alpha 3 release
3. Update the appsettings.json to put the connection string to the database in the ConnectionString.umbracoDbDSN field
4. Execute `dotnet run` to start the application
5. See that the application crashes

### Expected result

In previous versions, including alpha 2, the upgrade wizard would display to update the database to the current version.  Whether it does this, or takes you into the installer with the connection information already filled in, either way is fine, but it shouldn't crash the application.

### Actual result

Umbraco does not start, and reports the following error.
```
System.Data.SqlClient.SqlException (0x80131904): Invalid object name 'umbracoCacheInstruction'.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   at System.Data.SqlClient.SqlDataReader.get_MetaData()
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString)
   at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, SqlDataReader ds)
[...]

PR #10001, opened 2021-03-17
#9843 - Allow install with DB string pre-configured

benjaminc:
This resolves the error found in #9843 and allows for the install to proceed.  I did take out the check for whether or not the DB can connect, because this check happens in the `RuntimeState.GetUmbracoDatabaseState` method.  If it cannot connect, then either the Level is set to `RuntimeLevel.Install`, in which case we have the same behavior as before, or a boot exception is thrown in the `RuntimeState.DetermineRuntimeLevel` method.  In either case, we are handling the same scenarios without having to make a second connection to the database during server startup.

### DIFF
--- a/src/Umbraco.Infrastructure/Cache/DatabaseServerMessengerNotificationHandler.cs
+++ b/src/Umbraco.Infrastructure/Cache/DatabaseServerMessengerNotificationHandler.cs
@@ -14,20 +14,23 @@ namespace Umbraco.Cms.Core.Cache
     /// </summary>
     public sealed class DatabaseServerMessengerNotificationHandler : INotificationHandler<UmbracoApplicationStarting>, INotificationHandler<UmbracoRequestEnd>
     {
-        private readonly IServerMessenger _messenger;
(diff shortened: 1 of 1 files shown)


bergmania:
I see the reason this is happening. In v8, the `DistributedCacheBinderComposer` had minimum runtime level `RuntimeLevel.Run`. So this code was never executed doing installs.

I think the right solution is to return from the handler if runtime level is not `RuntimeLevel.Run` (Without logging). I'm also not sure it is a good idea to remove the old check for CanConnect. 

@benjaminc, what do you think?

benjaminc:
My first preference was to limit to just the Run level, but I wanted to not overreach in the code changes.  Good to hear that is how it was at v8; I'll set it back to that.

Regarding the connection check, I believe it isn't needed because the boot process checks if it can connect when it sets the run level, so checking the connection again after we see we are in the Run level seems redundant.  However I can put that back in if you like.

bergmania (approved):

bergmania:
Thanks @benjaminc

PR #10002, opened 2021-03-18
Fixes #9983 - Getting kicked, if document type has a Umbraco.UserPicker property

### LINKED ISSUE #9983: Getting kicked, if document type has a Umbraco.UserPicker property
If an Umbraco user, without access to the Users section, enters a page in backoffice, and the page has a Umbraco.UserPicker property, then the user is kicked out.

## Umbraco version

I am seeing this issue on Umbraco version: Umbraco version 8.10.2

Reproduction
------------

### Bug summary

1. Create a document type with a Umbraco.UserPicker property.
2. Access this page with a user that doesn't have access to the Users section.
3. You're getting kicked

### Specifics

It's tested in Chrome, Edge, Firefox

### Steps to reproduce

See the bug summary

### Expected result

I was expecting to see the Umbraco.UserPicker, with a list of users.
The only restriction should be the access to the actual User section

### Actual result

You're getting kicked from Umbraco back office

nul800sebastiaan:
Temporary fix for this issue. using the entityservice like before.

### Prerequisites

If there's an existing issue for this PR then this fixes #9983 

### Description
The new user picker only works properly if you're an admin user because it calls an API that is only available to people who have access to the Users section of the backoffice. If you try to use it without having access you get logged out of the backoffice.

In PR https://github.com/umbraco/Umbraco-CMS/pull/7613/ the endpoint usage changed, before that the user picker would use the entity service to get information about the available users. 

We can't make the new endpoint available to people who don't have access to the users section as it discloses too many details that should not be available to a user of type editor (for example). The new endpoint returns:

```javascript
{
	"userStates": {
		"All": 2,
		"Active": 1,
		"Disabled": 0,
		"LockedOut": 0,
		"Invited": 0,
		"Inactive": 1
	},
	"items": [{
		"username": "test@test.com",
		"emailHash": "b642b4217b34b1e8d3bd915fc65c4452",
		"lastLoginDate": "2021-03-18T08:30:12.15Z",
		"avatars": ["https://www.gravatar.com/avatar/b642b4217b34b1e8d3bd915fc65c4452?d=404&s=30", "https://www.gravatar.com/avatar/b642b4217b34b1e8d3bd915fc65c4452?d=404&s=60", "https://www.gravatar.com/avatar/b642b4217b34b1e8d3bd915fc65c4452?d=404&s=90", "https://www.gravatar.com/avatar/b642b4217b34b1e8d3bd915fc65c4452?d=404&s=150", "https://www.gravatar.com/avatar/b642b4217b34b1e8d3bd915fc65c4452?d=404&s=300"],
		"userState": 0,
		"culture": "en-US",
		"email": "test@test.com",
		"userGroups": [{
			"notifications": [],
			"sections": [{
				"name": "Content",
				"alias": "content",
				"routePath": null
			}, {
				"name": "Media",
				"alias": "media",
				"routePath": null
			}, {
				"name": "Settings",
				"alias": "settings",
				"routePath": null
			}, {
				"name": "Packages",
				"alias": "packages",
				"routePath": null
			}, {
				"name": "Users",
				"alias": "users",
				"routePath": null
			}, {
				"name": "Members",
				"alias": "member",
				"routePath": null
			}, {
				"name": "Forms",
				"alias": "forms",
				"routePath": null
			}, {
				"name": "Translation",
				"alias": "translation",
				"routePath": null
			}],
			"contentStartNode": {
				"name": "Content root",
				"id": -1,
				"udi": null,
				"icon": "icon-folder",
				"trashed": false,
				"key": "00000000-0000-0000-0000-000000000000",
				"parentId": -1,
				"alias": null,
				"path": "-1",
				"metaData": {}
			},
			"mediaStartNode": {
				"name": "Media root",
				"id": -1,
				"udi": null,
				"icon": "icon-folder",
				"trashed": false,
				"key": "00000000-0000-0000-0000-000000000000",
				"parentId": -1,
				"alias": null,
				"path": "-1",
				"metaData": {}
			},
			"userCount": 0,
			"isSystemUserGroup": true,
			"name": "Administrators",
			"id": 1,
			"udi": null,
			"icon": "icon-medal",
			"trashed": false,
			"key": "00000000-0000-0000-0000-000000000000",
			"parentId": -1,
			"alias": "admin",
			"path": "-1,1",
			"metaData": {}
		}],
		"isCurrentUser": true,
		"notifications": [],
		"name": "Sebastiaan Janssen",
		"id": 1,
		"udi": null,
		"icon": null,
		"trashed": false,
		"key": "00000001-0000-0000-0000-000000000000",
		"parentId": -1,
		"alias": null,
		"path": "-1,1",
		"metaData": {}
	}],
	"pageNumber": 1,
	"pageSize": 25,
	"totalPages": 1,
	"totalItems": 1
}
```

Whereas the entityservice only returns 

```javascript
[{
	"name": "Test",
	"id": -1,
	"udi": null,
	"icon": "icon-user",
	"trashed": false,
	"key": "ffffffff-0000-0000-0000-000000000000",
	"parentId": -1,
	"alias": "test@test.com",
	"path": "",
	"metaData": {}
}]
```

Please also note the overhead in all that extra unused information.

So with this PR, the user display will be a little less rich as the avatar is not available, so we go from:

image

To: 

image

But the functionality is preserved for now and we can still pick users even if we're not an admin. We'll need to figure out what to do about endpoints in the future, this is just fixing the regression for now.

Reproduction:
- Add a user picker on a doctype
- Log in as a user with no access to the Users section
- Note that you get logged out on the pages with the user picker
- Note that after this PR you no longer get kicked out

Additionally, do the same steps as an admin user and note that user picking works in the same way, except it won't show you user's avatars.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/userpicker/userpicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/userpicker/userpicker.controller.js
@@ -1,8 +1,8 @@
 (function () {
     "use strict";
 
-    function UserPickerController($scope, usersResource, localizationService, eventsService) {
-        
+    function UserPickerController($scope, usersResource, entityResource, localizationService, eventsService) {
+
         var vm = this;
 
         vm.users = [];
@@ -102,17 +102,9 @@
             vm.loading = true;
 
             // Get users
-            usersResource.getPagedResults(vm.usersOptions).then(function (users) {
-
-                vm.users = users.items;
-
-                vm.usersOptions.pageNumber = users.pageNumber;
-                vm.usersOptions.pageSize = users.pageSize;
-                vm.usersOptions.totalItems = users.totalItems;
-                vm.usersOptions.totalPages = users.totalPages;
-
+            entityResource.getAll("User").then(function (data) {
+                vm.users = data;
                 preSelect($scope.model.selection, vm.users);
-
                 vm.loading = false;
             });
         }
--- a/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/userpicker/userpicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/userpicker/userpicker.controller.js
@@ -1,7 +1,7 @@
 (function () {
     "use strict";
 
-    function UserPickerController($scope, usersResource, entityResource, localizationService, eventsService) {
+    function UserPickerController($scope, entityResource, localizationService, eventsService) {
 
         var vm = this;
 
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/userpicker/userpicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/userpicker/userpicker.controller.js
@@ -1,4 +1,4 @@
-function userPickerController($scope, usersResource , iconHelper, editorService, overlayService){
+function userPickerController($scope, usersResource , iconHelper, editorService, overlayService, entityResource) {
 
     function trim(str, chr) {
         var rgxtrim = (!chr) ? new RegExp('^\\s+|\\s+$', 'g') : new RegExp('^' + chr + '+|' + chr + '+$', 'g');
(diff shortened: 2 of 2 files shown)


src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/userpicker/userpicker.controller.js
```diff
@@ -1,8 +1,8 @@
 (function () {
     "use strict";
 
-    function UserPickerController($scope, usersResource, localizationService, eventsService) {
-        
+    function UserPickerController($scope, usersResource, entityResource, localizationService, eventsService) {
```
bjarnef, line 4: @nul800sebastiaan can we remove `usersResource` then?

nul800sebastiaan: Yep, done! 👍

madsrasmussen (approved):
I have tested the updated code and the user picker works for alle users now. It is not optimal that we have to get all users from the entity resource but it is better than being kicked out of the back office.

Issue #6053, opened 2019-08-05
V8: backoffice memory leak?


nathanwoulfe:
Seems to be a memory leak in the v8 backoffice - if I open and close infinite editors, the DOM node count continues to increase (100k+ nodes is a few clicks away). JS heap doesn't grow at the same rate, but does increase in line with the node count - can hit 200mb heap relatively easily.

Total DOM nodes stays flat after the increases, suggesting GC is not running/not running effectively.

Could be caused by a few things:
- ngAnimate [not likely - have removed from the Umbraco app, issue remains]
- not unregistering DOM and scope event handlers in $destroy
- potentially jQuery in directives [thinking this might be the culprit for infinite editors - these call into jquery's .remove() function, which should probably be replaced with vanilla js]

I've checked in an incognito window to rule out 3rd party stuff. Running v8.1.0.

Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

Shazwazza:
wonder if it's related to this? https://github.com/umbraco/Umbraco-CMS/pull/6051

nathanwoulfe:
@Shazwazza I don't think so - breakpoint in that function is never hit when I'm spawning infinite editors. Doesn't mean it's not contributing, but not to this exact issue...

I think it's probably related to using jQuery's .remove(), and that not affecting scope so the nodes aren't truly removed. Or something like that.

JeffreyPerplex:
I can confirm that the backoffice is sometimes really, really slow and using a lot of memory...

We we're creating a website and only creating some doctypes and datatypes and moving these datatypes into folders and every 5 minutes the browser crashed.

Using Umbraco 8.1.1, in both Firefox and Chrome

idseefeld:
One of my customers has encountered the issues as Jeffrey has described. Currently using 8.1.1. I am still investigating this...

skartknet:
This looks like is still happening on v8.8. I have no idea how to debug Angular but if I take a heap snapshot I can see that there are a lot of detached nodes.

![image](https://user-images.githubusercontent.com/1685262/96799209-5326fa00-144e-11eb-897f-bbd20dde1387.png)

![image](https://user-images.githubusercontent.com/1685262/96799227-5d48f880-144e-11eb-82c7-f7d4078b334f.png)

nathanwoulfe:
[...]

PR #10011, opened 2021-03-19
Unbind listener in variant content editor

nathanwoulfe:
Listening for splitViewRequest was only unbound if the split view editor was opened. Not cleaning up the listener caused a memory leak when changing between nodes as the spit view editor was detached but not garbage-collected.

This contributes to fixing #6053 but I'm not sure if it's solved the issue entirely - GC seems to be running more efficiently, node count doesn't continue to grow when travelling around the backoffice.


mikecp (approved):

mikecp:
Thanks @nathanwoulfe ,
It might not be the entire solution but it's an improvement for sure 👍

PR #10012, opened 2021-03-19
Netcore: Separate NuGet package for the static assets

bergmania:
### Details
This PR moved the static content and target+props file to a new separate nuget package. Now out root package `Umbraco.CMS` is only used to reference other packages, to make it easier in the general use case. 
This now allows users to only references what is needed for a backoffice Umbraco or a website Umbraco (e.g. for load-balanced environments).

### Bonus
When testing this, I found an issue if all indexes were not built. This happened because we tried to rebuild indexes, even if we were not in the `Run` state. This is now fixed, and I moved the logic from a component to the ExamineNotificationHandler in the UmbracoApplicationStarting event, instead of having a 5-sec. delay.


nikolajlauridsen (approved):
Looks good, tested it out and everything seems to work as intended 👍

PR #10013, opened 2021-03-19
Fixes some property editors ensuring ngModelController

### LINKED ISSUE #8954: Nested content mandatory field is not empty, but getting validation errors
## Umbraco version

Reproduction
------------

### Bug summary

After update from version 8.6.4 to 8.7.0, some nested element mandatory fields showing errors and prevents to save document. I noticed its only happens, when property names are general like 'name', 'image', it doesn't happen on other properties.

### Specifics

umbraco_bug

---
_This item has been added to our backlog [AB#9075](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/9075)_

Shazwazza:
Fixes #8954

> The underlying problem is that all input, value and validation logic in AngularJs is controlled by ngModel and we haven't made this a requirement for property editors. For all of this to work we need to know when a property editor value changes and the only way to know that is with ngModel.

> We cannot rely on AngularJs's FormController because those $dirty changes cascade upwards so it's impossible to know if a value changed for that form or a nested form (or nested/nested form, etc...). We cannot rely on manual $watch expressions for a model value because complex editors like nested content, block editors, etc... have a model that is ... well ... nested. Watching on that value would be horrible for performance and because it's nested we cannot tell which actual property is changed (same issue with monitoring FormController).

> If we were to rebuild property editors in AngularJs: They would be components that require ngModel (i.e. like this https://docs.angularjs.org/api/ng/type/ngModel.NgModelController#custom-control-example).

> There's all sorts of hacks in place in property editors due to the fact that many are not using ngModel. Anytime we are doing things like: angularHelper.getCurrentForm($scope).$setDirty(); to set a property editor to be marked as $dirty when user interaction occurs. This should not be required if ngModel was used correctly. We also have newer/modern components that are not using ngModel when they should be such as the umb-toggle component. In AngularJs anytime there is a value being bound that changes with user input, the component should be backed with ngModel so that it can be properly validated, and monitored.

> But we can't go back in time so the only way to fix this is with more hacks 😢

> Anywhere this is used angularHelper.getCurrentForm($scope).$setDirty(); in property editors means that this property editor doesn't use an ngModel to track it's current value and one will need to be added as a hidden input field. Because hidden input fields cannot be interacted with by the user, they need to be manually set to $dirty, but that's ok. This just means that the call to manually set the editor to dirty would be done more explicitly with this new field like: vm.modelValueForm.modelValue.$setDirty()
After doing 1) Need to go through every property editor to ensure that it uses an ngModel in some way to track it's model value

## Changes

* Expose `setDirty` API method on `umb-property` to avoid this ugly thing that was used by nested content: `$scope.$parent.$parent.propertyForm.$setDirty();`
* Updates many property editors to ensure there is an ngModel controller as part of their editor with a value that represents the current value of the editor with regards to knowing if it changes. Any editor that used `angularHelper.getCurrentForm($scope).$setDirty();` needed to be changed to call $setDirty() on it's own ngModel **not** an arbitrary parent form. This typically means adding a hidden input field (and in many cases a locally declared ng-form) with ngModel for the value representing the editor and creating a `setDirty` method for that editor to call which would set this ngModel to dirty when there was user interaction.
  * Boolean editor
  * Media picker (I also changed this one to use `vm` and simplified some things like)
  * color/multi-color picker
  * content picker
  * date picker
  * image cropper
  * markdown editor
  * member picker (dirty tracking already wasn't working correctly)
  * multi url picker
  * slider
  * user picker (this wasn't tracking dirty changes at all)
  * member group picker (this wasn't tracking dirty changes at all)
  * member groups (used in the member editor, wasn't tracking dirty changes at all)

## Testing

Make a Nested Content editor that has all of the above (and more if you can test them all!) as properties and marked as "Required". Also allow this Nested Content property to be a property of itself so you can keep nesting them.

Create some content with this editor and try to save. You will get Required errors on all properties. Go through and edit each one and ensure that the property level error goes away and that you can save it. Now do the same with a couple levels of nested items so that you get multiple levels of validation errors at once and go through and change each property at each level ensuring that the property level errors are removed as they should be.

Now do the same with the block list editor.


nielslyngsoe:
Have read the code, tested the described scenarios, and it all works as it should. I did try some other property editor and had minor trouble with the Radio Buttons, but that's not deal-breaking, and nothing introduced by this. So here is a new issue for that: https://github.com/umbraco/Umbraco-CMS/issues/10122

BTW. I read the newly added TODOs as things we should do, but we cant cause it would be a breaking change.

I approve and merge this PR.

PR #10017, opened 2021-03-19
Proof of concept members password hasher

scottbrady91:
### Prerequisites

### Description

Password hasher for members that supports the legacy hashing algorithm and the ASP.NET Identity hashing algorithm. The hashing algorithm is determined by hash format (base64 or not) and the ASP.NET Identity marker byte.

Since `LegacyPasswordSecurity` didn't seem 100% testable I wrote some integration tests using known password hashes.

Let me know if there are any other requirements that I am not aware of.

### DIFF
--- a/src/Umbraco.Infrastructure/Security/MemberPasswordHasher.cs
+++ b/src/Umbraco.Infrastructure/Security/MemberPasswordHasher.cs
@@ -1,0 +1,63 @@
+using System;
+using Microsoft.AspNetCore.Identity;
+using Umbraco.Cms.Core;
+using Umbraco.Cms.Core.Security;
+
+namespace Umbraco.Cms.Infrastructure.Security
+{
+    /// <summary>
+    /// A password hasher for members
+    /// </summary>
+    /// <remarks>
+    /// This will check for the ASP.NET Identity password hash flag before falling back to the legacy password hashing format ("HMACSHA256")
+    /// </remarks>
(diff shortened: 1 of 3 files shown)


scottbrady91:
@bergmania @Shazwazza

bergmania:
Looks good @scottbrady91.

Not sure where, but I think we should update the users' algorithm to the new format and a new hash, if we see it is the legacy? What do you think @Shazwazza.

Shazwazza:
> Not sure where, but I think we should update the users' algorithm to the new format and a new hash, if we see it is the legacy? What do you think @Shazwazza.

We already do this and is handled by `BackOfficePasswordHasher`. It is handled differently for users because we store the current hashing algorithm with them. Ideally we do this for members too but that requires DB changes, etc... for members (which we want to do for other reasons anyways)

scottbrady91:
By returning `SuccessRehashNeeded`, the UserManager will automatically update the user's password hash: https://github.com/dotnet/aspnetcore/blob/main/src/Identity/Extensions.Core/src/UserManager.cs#L710

Shazwazza:
> By returning SuccessRehashNeeded, the UserManager will automatically update the user's password hash:

ah yep, that's also how the user stuff works too. I'll get this merged and updated. Thanks!

Shazwazza:
Merged in commit 8b8bf4774af867c7f7b75c9dd12e597a8b638a61
Not sure why that isn't reflecting here. 

Thanks!!

Issue #10016, opened 2021-03-19
Edit item doesn't work in custom listview layout using listViewHelper


bjarnef:
It seems current implementation of `listViewHelper.editItem(item);` doesn't work correct after upgrade to Umbraco v8.12.1
I seems it was affected by the changes in this PR: https://github.com/umbraco/Umbraco-CMS/pull/9026

In this case I upgraded a site from v8.8.4 to v.8.12.1 the "go to" link doesn't work in my custom listview layout. The "go to" links does work in the default list and grid layouts though.

```
$scope.gotoItem = function (item) {
    listViewHelper.editItem(item);
};
```

It is throwing this console error:
![image](https://user-images.githubusercontent.com/2919859/111779211-23bdd180-88b6-11eb-8231-c03aa2464bd1.png)

Add a custom listview layout, e.g "grid-employees" and place these files in a folder in `App_Plugins`: /App_Plugins/ListviewLayouts/grid-employees/grid.html

**Controller**
```
(function () {
    "use strict";

    function ListViewGridLayoutController($scope, $routeParams, mediaHelper, mediaResource, listViewHelper) {

        var vm = this;

        vm.nodeId = $scope.contentId;

        $scope.toggleItem = function (item) {
            if (item.selected) {
                listViewHelper.deselectItem(item, $scope.selection);
            } else {
                listViewHelper.selectItem(item, $scope.selection);
            }
        };

        $scope.gotoItem = function (item) {
            listViewHelper.editItem(item);
        };

        $scope.items = $scope.items.map(function (obj) {

            if (obj.contentTypeAlias === "employee") {
                var imageProp = _.find(obj.properties, function (p) { return (p.alias === "image"); });
                if (imageProp) {
                    obj.thumbnail = "https://www.gravatar.com/avatar/?d=mm&s=100&f=y";

                    var mediaId = imageProp.value;
                    if (mediaId) {
                        mediaResource.getById(mediaId).then(function (media) {
                            obj.thumbnail = mediaHelper.resolveFile(media) + "?width=100&height=100&mode=crop&animationprocessmode=first";
                        });
                    }
                }
            }

            return obj;
        });
    }

    angular.module("umbraco").controller("ListView.GridMembersLayoutController", ListViewGridLayoutController);
})();
```

**View**
```
<div ng-controller="ListView.GridMembersLayoutController as vm">
    
    <div class="umb-content-grid employees-grid">     
        <div ng-repeat="item in items" 
[...]

PR #9026, opened 2020-10-01
Allow use of infinite editor in list views

skttl:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes #8341

### Description
This one adds a new setting to a list view definition, allowing you to enable infinite editing for the content in the list view.
image

vEcb4rPQqL

To test, set a list view to use infinite editing and verify that clicking an item in the list view opens the item in an infinite editor.
Also verify that creating new items opens in an infinite editor.

**Note:** Blueprints hasn't been implemented in the editorService, so they don't open an infinite editor when getting created.

PR #10018, opened 2021-03-19
Ensure backward compatibility in listview edit with custom listview layout

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10016

### Description
This PR fixes the changes introduced in https://github.com/umbraco/Umbraco-CMS/pull/9026 which broke  backward compatibility using custom listview layouts.

Not sure why it need to pass in the entire scope as second parameter and not just the options though, but we probably can't change that now without making another breaking change. It would probably be simpler with something like this:

```
const options = {
      id: 1234,
      useInfiniteEditor: true
}
```

The rest is available from `item`, but maybe it has something to do with it also need `scope.getContent(scope.contentId)`.

To make in easy to test I have create a custom listview for people in default starter kit. Unzip this "ListviewLayouts" folder to `App_Plugins` and configurate a custom listview for the "people" using the following path: **/App_Plugins/ListviewLayouts/grid-people/grid.html**

[ListviewLayouts.zip](https://github.com/umbraco/Umbraco-CMS/files/6172556/ListviewLayouts.zip)

image

**Before**

chrome_mLKtQRzRbU

**After**

zvdTwoh0tf

### Test notes

- Check clicking link works for both standard list and grid + custom layout
- Enabled "Edit in Infinite Editor" in listview configuration and ensure both standard list and grid open infinite editor. The custom listview would do that, but if modifying the listview controller with this following it handle this as well:

```
$scope.gotoItem = function (item) {
    listViewHelper.editItem(item, $scope);
};
```

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/listviewhelper.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/listviewhelper.service.js
@@ -572,13 +572,15 @@
         * Method for opening an item in a list view for editing.
         *
         * @param {Object} item The item to edit
+        * @param {Object} scope The scope with options
         */
         function editItem(item, scope) {
+
             if (!item.editPath) {
                 return;
             }
 
-            if (scope.options.useInfiniteEditor)
+            if (scope && scope.options && scope.options.useInfiniteEditor)
             {
                 var editorModel = {
                     id: item.id,


bjarnef:
I wonder if the "Edit in Infinite Editor" also should work with the old format:

```
$scope.gotoItem = function (item) {
    listViewHelper.editItem(item);
};
```

Currently the toggle has no effect on the listview layout unless you modify the implementation in your custom listview layout.

nathanwoulfe (approved):
@bjarnef this works nicely - I agree it would be good to not have to pass the entire scope object into the listviewhelper, but it is due to needing both the options object and the getContent method. I guess too by passing only what's required, it would be possible for implementors to pass a different method for getContent, which would then cause issues.

getContent is originally passed into the umbListViewLayout directive, I'm guessing to ensure the displayed values in the list reflect any changes made in the infinite editor. 

So yeah, while it's not ideal given it will required updates in custom listviews to use infinite editing, I think your fix here is sensible since it stops anything from breaking.

bjarnef:
@nul800sebastiaan can this one be cherry picked to next patch release of v8.12 as well?

nul800sebastiaan:
Yep! Cherry picked for 8.12.2 in https://github.com/umbraco/Umbraco-CMS/commit/72e468428ad0c76ea4076d22693cc38f8a331eae

bjarnef:
@nul800sebastiaan I guess you mean v8.12.3 😁👍🚀

nul800sebastiaan:
Yes!! Duh.. 😅

Issue #9991, opened 2021-03-17
Copy member type


bjarnef:
I just had the need to create to new member types and would copy from the default member type, but I noticed this isn't an option.
However we can copy document types and media types.

It would be great if we could have this option for member types as well making it more consistent and easier to create new member types, where you want the default properties starting with `umbraco` in alias.

~~Currently I don't think there is an easy way to add all these standard properties to a new member type except adding them manually - at least from backoffice UI.~~
The standard properties are added when saving a new member type and locked, so the group "Membership" and the standard properties in this group can't be deleted.

Alternatively it could be useful to have compositions on member types like on document types and media types: https://github.com/umbraco/Umbraco-CMS/issues/4280

nul800sebastiaan:
Hey @bjarnef, thanks for the suggestion! I think copying member types would be fairly easy to do and a good idea. 

However, compositions are complex and I think difficult to implement correctly. It doesn't seem worth the big effort since the audience for that feature would be very limited. 

So I would encourage working on a copy feature, and not to focus on compositions.

nul800sebastiaan:
Sidenote: we'll need to test with deploy if the copy feature gets implemented, I'm sure it will work but good to make sure anyway.

umbrabot:
Hi @bjarnef,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
PR submitte https://github.com/umbraco/Umbraco-CMS/pull/10020 🎉 🌮

PR #10020, opened 2021-03-19
Copy member type

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9991

### Description
This PR implement the copy feature for member type similar to how it works for content type and media type. Currently this works (almost) as expected. However unlike content/document types and media types where it use `customtreeparams="foldersonly=1"` I guess it in this context doesn't know what a folder is, so it both shows folders (or only root node) and also all member types. The member type being copied isn't selectable, but the others are, which cause an error if trying to copy a member type under another member type (which is allowed for e.g. media types).

Besides that it is a great feature and could save some time in development if working with several member types in multiple projects.

```
<umb-tree section="settings"
                  treealias="memberTypes"
                  customtreeparams="foldersonly=1"
                  hideheader="false"
                  hideoptions="true"
                  isdialog="true"
                  api="dialogTreeApi"
                  on-init="onTreeInit()"
                  enablecheckboxes="true">
</umb-tree>
```

OHEzQHul5j

Furthermore failed move and copy of media types was actually using an incorrect localization since it just use the wording `media` when it is `media type`.
So I have added new areas for specific content type, media type and member type. "Document type" is still used, but now with "Element type" as well, "content type"  area could be used for both.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/resources/contenttype.resource.js
+++ b/src/Umbraco.Web.UI.Client/src/common/resources/contenttype.resource.js
@@ -432,13 +432,15 @@ function contentTypeResource($q, $http, umbRequestHelper, umbDataFormatter, loca
                 throw "args.id cannot be null";
             }
 
+            var promise = localizationService.localize("contentType_moveFailed");
+
             return umbRequestHelper.resourcePromise(
                 $http.post(umbRequestHelper.getApiUrl("contentTypeApiBaseUrl", "PostMove"),
                     {
                         parentId: args.parentId,
                         id: args.id
                     }, { responseType: 'text' }),
-                'Failed to move content');
+                promise);
         },
 
         /**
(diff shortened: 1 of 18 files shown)


bjarnef:
Okay, it turns out that `foldersonly` was a custom querystring specified, so I have added that as well for member types (as member groups since they use same base controller, but it doesn't seem to effect member groups with the current functionality). https://github.com/umbraco/Umbraco-CMS/pull/10020/commits/16e625dcd39be8e0d647c1fcbaa312f686eb124c

mikecp:
Thanks a lot @bjarnef for this PR, it works as expected and this will add a nice new feature to the back-office! 👍 
I took the liberty to add French translations as well 😁

mikecp (approved):

bjarnef:
@nul800sebastiaan hasn't this been merged up to v9? I can't find the copy feature in v9.2.0 or latest changed in `v9/contrib` branch.

Issue #9929, opened 2021-03-03
Nested content description field - show on hover - translation


thomashdk:
We are using dictionary items a lot to translate the backoffice to multiple languages.

For that we are using #Name on doctypes etc. 
On our componets description fields it would be nice if the description text that are showed on hover uses the same feature as the naming for the nested content does :)

![Nested content](https://user-images.githubusercontent.com/6445636/109803376-40ed7180-7c21-11eb-8e0d-1c9e18c40719.gif)

nul800sebastiaan:
Ah interesting! Yeah that seems like an oversight!

umbrabot:
Hi @thomashdk,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

RyanHelmn:
Hey @thomashdk I've created a PR for this here: https://github.com/umbraco/Umbraco-CMS/pull/10025 will keep you posted how it goes.

umbrabot:
Hiya @thomashdk,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #10025, opened 2021-03-21
Added translated descriptions to nested content & block list

RyanHelmn:
### Prerequisites

- I have added steps to test this contribution in the description below

Fixes: https://github.com/umbraco/Umbraco-CMS/issues/9929

### Description
I've included ContentTypeDescription at the SimpleContentType level so that once it's translated and passed back to the nested content and block list editor controller it can use the new property.

In the ContentController (src/Umbraco.Web/Editors/ContentController.cs ~ 388) I've also included the DocumentType.Description translation as it seems that the block list editor uses that.

### DIFF
--- a/src/Umbraco.Core/Models/ISimpleContentType.cs
+++ b/src/Umbraco.Core/Models/ISimpleContentType.cs
@@ -9,6 +9,7 @@ namespace Umbraco.Core.Models
     {
         new int Id { get; }
         new string Name { get; }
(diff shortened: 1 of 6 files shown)


OwainWilliams:
Hi @RyanHelmn 
Thanks for taking the time to investigate this and submit a PR. We'll take a look at this as soon as possible. #h5yr 

O.

PR #10026, opened 2021-03-21
Fix issue with active style affected nested block list elements

bjarnef:
### Prerequisites

### Description
I noticed some styling issues with nested block lists, where the active style also affected all nested block list block elements mainly regarding the caret to visualize the active/open block element.

This happens when using inline mode for both parent and nested block lists.

**Before**

8pYziVD8QU

**After**

B9dPnCC3Wl


bjarnef:
//cc @nielslyngsoe @madsrasmussen

nathanwoulfe (approved):
Thanks @bjarnef - it's some chunky CSS, but I think that's unavoidable due to the element nesting. Fixes the issue, so that's a winner.

PR #10028, opened 2021-03-22
Do not apply runtime optimization to core CSS and CSS libraries

kjac:
### Prerequisites

### Description

_This is the CSS equivalent of #9839_

This PR ensures that we ship minified CSS assets and dependencies instead of minifying them at runtime.

As it happens we're already minifying the core CSS when building the backoffice, but we still tell Smidge to minify it (again) at runtime. I have changed the destination files for the backoffice CSS to have a `.min.css` extension to make it clearer that they are actually minified.

Note that the `bootstrap-social` NPM package does not ship with a `.min.css` stylesheet. But `bootstrap-social.css` is still minified, or at least almost minified, so I have chosen to include it as-is:

image

### Testing this PR

If the backoffice looks alright, the PR works 😄 

Also you might want to test that you can configure a color picker datatype:

image

..and use a date/time picker in content:

image

### DIFF
--- a/src/Umbraco.Infrastructure/WebAssets/BackOfficeWebAssets.cs
+++ b/src/Umbraco.Infrastructure/WebAssets/BackOfficeWebAssets.cs
@@ -48,19 +48,18 @@ namespace Umbraco.Cms.Infrastructure.WebAssets
         {
             // Create bundles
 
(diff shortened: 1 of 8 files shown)


nikolajlauridsen (approved):
Changes look good to me, ran through the tests as well with no issues 👍

PR #10033, opened 2021-03-22
UI enhancement: Green checkmark icon for selected item in various views and components

jakob-bagterp:
### Prerequisites

### Description

With the intention to improve the user interface and visibility of interaction in editors, tree pickers, and other components, let's make the checkmark icon green so it appears distinctive from the other unselected items in dark blue/black. Example from the umb-tree-item.html file.

Before:

Screenshot 2021-03-22 084155

After:

Screenshot 2021-03-22 084210

Also, this makes this change consistent with existing use of the checkmark icon:

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/tree/umb-tree-item.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/tree/umb-tree-item.html
@@ -9,7 +9,7 @@
             <span class="sr-only">Expand child items for {{node.name}}</span>
         </button>
 
-        <umb-icon icon="{{node.selected ? 'icon-check' : node.icon}}" class="icon umb-tree-icon sprTree" ng-class="node.cssClass" title="{{::node.title}}" ng-click="select(node, $event)" ng-style="::node.style" tabindex="-1"></umb-icon>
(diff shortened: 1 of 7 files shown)


nul800sebastiaan:
Hi @jakob-bagterp, thanks for trying to help here. Unfortunately, we actually recently changed the color to match the blue of the other colors, this keeps the color scheme consistent and the page more calm. 
For now, we feel like the big blue border around the node, combined with the checkmark is a good enough indicator. 

We'll treat this as a signal that you think it's not entirely clear enough though and there's good news/bad news in that regard: I understand from the frontend devs at HQ that In the future we want to change this to be even more obvious but due to some technical limitations we can't do it yet. 

For now we'll leave it as it is until the ideas that the frontend team has can be implemented!

Thanks again for the help! We do appreciate it. As advice for the future, before diving in and making a change, it's usually good to ask in an issue first if it's an idea that we'd support. I mention this as I feel guilty for the time you spent making the changes while we can't accept them. Sorry about that!

jakob-bagterp:
@nul800sebastiaan: All good and understandable. Have a great day!

PR #10036, opened 2021-03-22
Listview config icons

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
This PR adjust a few things in listview configuration:

- Update icons using `<umb-icon>` component
- Ensure sortable handle use `move` cursor style
- Simplified markup for icon button for system/custom icon
- Increase font size / icon size of custom app icon. Note the class is also used in icon in e.g. icon picker button on content, media and member types.

PR #10038, opened 2021-03-22
Update casing of listview layout name

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/pull/10036

### Description
This PR update casing of "Grid" listview layout which was changed a long time ago from "Grid" to "grid" in https://github.com/umbraco/Umbraco-CMS/commit/fb246de5fc27a911fe8381026218abc37fbf5a60

It seems to fix the issue with naming of "Grid" layout in "List View - Members".

For the custom listview configuration in still seemed to use the old "grid" name though.

image

After deleting the datatype instance an configurate a new custom listview for the document type if does use the correct "Grid" layout name.

image

The reason the listview "Grid" layout name sometimes was correct is because it is added here during install:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Migrations/Install/DatabaseDataCreator.cs#L267


nathanwoulfe (approved):
Nice, thanks @bjarnef 👍

Issue #9761, opened 2021-02-03
Email address property editor validation doesn't block post back when invalid email address is provided


EneldoDev:
When editing content node with property using Email address property editor, frontend email validation doesn't block postback when email is provided in wrong format.

## Umbraco version

I am seeing this issue on Umbraco version: 8.11.0

Reproduction
------------

### Steps to reproduce
1. Create document type with property using Email Address property editor (field NOT mandatory). 
2. Using this document type create content node. Fill email address field with improper value. 
3. Save/Save and publish content.

### Expected result

Invalid email error message is shown and postback on save is blocked. No page reload.

### Actual result

Once Save/Save and publish button is clicked validation message "Invalid email" is shown but form postback isn't blocked. Content node is saved, page reloaded. Email address field is empty though what means backend validation works correctly.

Validation is working properly only when "Field is mandatory" checkbox is checked.

emma-hq:
Hey @EneldoDev 

Thanks for reporting this. I'll try to reproduce and let you know how we'll progress once I have :)

Emma

emma-hq:
Hey @EneldoDev 

Yep - I have exactly this problem on both 8.11.0 and later releases. We'll label this up for grabs so someone can fix it 😁 Of course, you're very welcome to submit a pr if you have the inclination.

![emailvalidation](https://user-images.githubusercontent.com/78740801/108851866-d1ea9a00-75dc-11eb-9495-b786073a7ddc.gif)

Emma

umbrabot:
Hi @EneldoDev,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

emma-hq:
For the person picking up this issue - please also look at #9727 The numeric prop editor exhibits the same behaviour

Emma

bjarnef:
Not specific fixing this issue, but related to the integer/numeric datatype validation if a step size has been set: https://github.com/umbraco/Umbraco-CMS/pull/10039

bjarnef:
[...]

PR #10039, opened 2021-03-22
Fix validation of step size in integer/numeric field

bjarnef:
### Prerequisites

### Description
When a step size is configurated in integer/numeric datatype the value isn't saved if it doesn't fit with the step size.
Considering the following:

Minimum: 100
Step: 2
Maximum: 200

image

Values below 100 and above 200 are invalid, but also a number like 101 due how the step attribute works with HTML5 input fields.

With this change it add a nicer validation, which has already been fixed for the Decimal datatype in https://github.com/umbraco/Umbraco-CMS/pull/9334

image


bjarnef:
Note to test this it is probably easiest to set the property as mandatory due a bug with both email and integer/numeric input when saving a non-mandatory property, but with an invalid value https://github.com/umbraco/Umbraco-CMS/issues/9761

mikecp (approved):

mikecp:
Thanks @bjarnef  for this fix 👍

PR #10041, opened 2021-03-22
Feature/examine logging

bielu:
### Prerequisites

### Description
As debugging of examine on server was really painful specially when using Azure blob directory, I introduce some small changes in examine to allow inject logging service to examine, so as soon Shannon will release official build that PR can be merged.

### DIFF
--- a/src/Umbraco.Web/Search/ExamineComposer.cs
+++ b/src/Umbraco.Web/Search/ExamineComposer.cs
@@ -1,5 +1,10 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Configuration;
+using System.Linq;
 using Examine;
+using Examine.Logging;
+using Examine.LuceneEngine.Directories;
 using Umbraco.Core;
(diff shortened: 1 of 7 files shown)


src/Umbraco.Web/Search/ExamineComposer.cs
```diff
@@ -32,6 +37,12 @@ public override void Compose(Composition composition)
             composition.Register<IndexRebuilder>(Lifetime.Singleton);
             composition.RegisterUnique<IUmbracoIndexConfig, UmbracoIndexConfig>();
             composition.RegisterUnique<IUmbracoIndexesCreator, UmbracoIndexesCreator>();
+            composition.RegisterUnique<ILoggingService, UmbracoExamineLoggingService>();
+            foreach (Type factory in composition.TypeLoader.GetTypes<IDirectoryFactory>())
```
Shazwazza, line 41: We should not type scan here. We only ever type scan for a single type: IDiscoverable else there is a lot more overhead. Each directory factory should just be explicitly registered.

Issue #10030, opened 2021-03-22
NetCore: No authenticationScheme was specified


elit0451:
### Bug summary
The same unhandled exception occurs while executing the following requests:
- http://localhost:5000/umbraco/backoffice/umbracoapi/content/PostSaveUserGroupPermissions
- http://localhost:5000/umbraco/backoffice/umbracoapi/media/PostSort

Setting permissions for a Content node or sorting media on **alpha04** fails. 

Permissions for a Content node             |  Sorting media
:-------------------------:|:-------------------------:
![](https://user-images.githubusercontent.com/21998037/111993838-603c3800-8b17-11eb-9c4d-492e3fbedb24.png)  |  ![](https://user-images.githubusercontent.com/21998037/111996905-c5ddf380-8b1a-11eb-87a6-068a99691096.png)

### Specifics
Enabling or disabling any permission for any user group results in a "_Failed to save permissions_" error.
While sorting media results in a "_Failed to sort media_" error.

#### Permissions for a Content node
- Create a simple Document type, and in the Content section create a node from that document type;
- Right-click on the node and click **Permissions...**;
- Then click **Set permissions for...**;
- Pick a user group and enable or disable an action;
- Click Save.
- That will result in an Internal Server Error (http://localhost:5000/umbraco/backoffice/umbracoapi/content/PostSaveUserGroupPermissions);

</br>

#### Sorting Media
- Create 2 Folders as part of the Media tree;
- Right-click on the Media dialog;
- Click Sort and attempt to change the order of the 2 folders;
- Then Save;
- The error will pop up;

### Expected result
The permission(s) are saved and the Media items are sorted;

### Actual result
![image](https://user-images.githubusercontent.com/21998037/111994956-afcf3380-8b18-11eb-957c-96d42638c0d9.png)

bergmania:
Fixed in this PR: https://github.com/umbraco/Umbraco-CMS/pull/10042. Same issue as we have seen before. Because `IAuthorizationService.AuthorizeAsync` just accepts an `object` instead of some marker interface, we did not see these errors compile time.

nikolajlauridsen:
Just merged #10042 so I'm closing this issue 👍

PR #10042, opened 2021-03-23
Netcore: Fix no authentication schema was specified bugs

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/10030


nikolajlauridsen (approved):
Looks good 👍 

Tested it as well and can confirm that the issue is resolved

Issue #10046, opened 2021-03-23
V8 GetPagedChildren supports filter by content type alias


nick-hoang:
I would like to filter by content type alias in GetPagedChilderen method, for example:

```
var filter = SqlContext.Query<IContent>().Where(x => x.ContentType.Alias == "contentTypeAlias")

IEnumerable<IContent> Children = Services.ContentService.GetPagedChildren(Id, PageIndex, PageSize, out long TotalChildren, filter, Ordering.By("CreateDate", Direction.Descending, null, false));
```

nick-hoang:
#10047

emma-hq:
Hi @nick-hoang 

Thanks for raising this one and the PR too. Top marks for being proactive! We'll take a look at the approach and let you know after we've discussed the PR. 

Emma

PR #10047, opened 2021-03-23
GetPagedChilderen method supports filter by content type alias

nick-hoang:
TODO: Only do inner join to cmsContentType table if needed (depends on the where clauses)


mikecp:
Thanks for this PR @nick-hoang 👍 
I have assigned it the same lable `hq-discussion` as the linked issue, I hope @nul800sebastiaan can give us the result of the discussion soon so that we know if we can merge 😉

nul800sebastiaan:
Thanks @nick-hoang, my concern with this would be performance plus the fact that it is not immediately obvious for me what the use case is. I saw some code in the related issue #10046 but can you give me a scenario of what you can not do today?

Also, this PR description only contains a TODO, I assume it's not done yet? Before you do more work on this, we'd first like to  get an idea of what it is you're trying to achieve. 👍

nick-hoang:
Hi @nul800sebastiaan

In fact, there's many cases that we need to filter all children in a specific content type alias and we did it easily in Umbraco 7, do you agree with this? This is a basic scenario that we should support in the core I think

And the performance is one of my concern as well but it's quite complicated for me, that's why I had a TODO in the description, just in case some one can help or have another idea.

Cheers,
Nick

Issue #9282, opened 2020-10-27
Allow specific document types in listview to be rendered in tree


bjarnef:
Sometimes you might have a listview, e.g. a Employee Archive and inside this Employee nodes and therefore enabling listview as their could be hundreds or even thousands of employee nodes.

However you might also have another document type used in the listview, e.g. "Employee Category" which could be used as data in a category picker on the employee to select in which area the employee is working in, e.g. "Marketing", "Project Managers", "Developers", etc.

However these categories could only be about 10-20 nodes. It would in this scenario be great if specific document types could be rendered as tree nodes. It would be okay if they still show up in listview.

If the nodes would still be shown in listview, it could be nice to have a filter option for the different document types shown in listview similar to the filter options in logviewer or users section.

nathanwoulfe:
@bjarnef that's a pretty slick idea - I've run into the same issue where we use a list view to display 100+ nodes, but mixed in among those are a handful of a different type, which can then be difficult to find. 

Would need to run this past HQ, but personally I think it would be a usability improvement. @nielslyngsoe might have a view on this too...

umbrabot:
Hiya @bjarnef,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

Issue #10053, opened 2021-03-24
Show content created under listview


bjarnef:
Based on Umbraco v8.12.1 it is possible to create content in a listview and under the listview content items.
However when listview is enabled there is no way to see the content for these deeper level although it is possible to select a listview item an create another node under this via the actions menu in top right corner.

Related to this issue https://github.com/umbraco/Umbraco-CMS/issues/9282 but in this case you could have a "Brand Archive" with "Brand Page" nodes. Under each brand you could have text pages, etc. but when listview is enabled none of these descendants are visible.

I would expect a "Child Items" or similar here to show child nodes. I can enable this for the "Brand Page", but I would need to enable listview to all descedants to see their child nodes.

![image](https://user-images.githubusercontent.com/2919859/112313525-3fefb300-8ca8-11eb-9cf3-c05c2ce982ac.png)

Furthermore a Text Page could be used in multiple levels, where you don't want to enable "listview" everywhere.

Alternatively it is also possible to add a property using listview datatype instance instead of enable listview content app.

Is there a better way to handle this? I think when listview is enabled we need to ensure any child items in always accessible in a consistent way.

nul800sebastiaan:
You can add the listview datatype on child types, I think that's what you're trying to do? Then your child item is not configured as a list view but it will have a list view property with all the children under it.

bjarnef:
Yes, I could do that, but let's say I have a structure like the following, it would require to adding listview data type on the text page to be able to access to on lower levels, but it also mean I will get listview property at this level "Home Page > Text Page".

- Home Page
   - Text Page
   - Text Page
   - Brand Archive
     - Brand (list view)
      - Text Page
        - Text Page
        - Text Page

Sometimes I would like some node types could be excludes from listview, but currently this isn't possible.
I made this POC https://github.com/umbraco/Umbraco-CMS/pull/10057 but not sure how we could render a "normal" tree under listview, but exclude the listview items (currently just an empty collection).

nul800sebastiaan:
Yeah.. I doubt we'll want to go there, it is not what list view was meant for really, it is mostly meant for very flat lists of items where there is too many items for them to make sense in a tree. It's not exactly meant for frequently editing in a site structure.

We'll have a chat about it at HQ though.

bjarnef:
Sometimes it is however useful to have an archive of items while using another type in archive as well.
E.g. to use "category" as data in "category picker" on employee, while is it is own page (listing employees in that category).

[...]

PR #10057, opened 2021-03-24
Add config option to select content/media/member type to exclude

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9282

### Description
This PR is WIP and add a configuration option in listview to select content/media/member type to be excluded from listview and rendered in tree. We might remove this for members listview as if currently not is possible to create child nodes of members.

I have modified the `treesource` prevalue editor a bit, so it is possible to disable node picker and xpath and just use the node type as this emit the `treeSourceChanged` event used in `treesourcetypepicker` prevalue editor.

w4FMc70xn2

Currently I am not sure how to change this part, so it by default still not render child nodes in tree, but whitelist some content types to be rendered as child nodes.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Trees/ContentTreeControllerBase.cs#L398-L403

Now it "just" need to render these node types in tree, listview create action and maybe exclude these from listview layout (grid/list/custom).

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/treesource.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/treesource.controller.js
@@ -13,13 +13,21 @@ angular.module('umbraco')
 	            type: "content"
 	        };
         }
-        if (!$scope.model.config) {
-            $scope.model.config = {
-                idType: "udi"
-            };
-        }
 
-        if($scope.model.value.id && $scope.model.value.type !== "member"){
+        // setup the default config
+        var config = {
+            idType: "udi",
+            showXpath: true,
+            allowSelectNode: true
+        };
+
+        // map the user config
+        Utilities.extend(config, $scope.model.config);
+
+        // map back to the model
+        $scope.model.config = config;
+
+        if ($scope.model.value.id && $scope.model.value.type !== "member"){
             entityResource.getById($scope.model.value.id, entityType()).then(function(item){
                 populate(item);
             });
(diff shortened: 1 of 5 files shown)


bjarnef:
@Shazwazza @nielslyngsoe @nul800sebastiaan I have created this POC, since I in a few projects have had a need to have a few nodes, e.g. a "Category" node in a listview which could potential could have a lot of nodes like an archive of employees, products, etc.

In this case it would be great to render these types as tree nodes.
Before continue the work if work be great to know if the thoughts on this from HQ.

nielslyngsoe:
Could it make sense to do it in an inverted manner, so you pick which types go into the list-view? If none has been picked all goes?
Could this eventually enable multiple list views? ... just some thoughts.

Could you provide a use case where this provides value for you? :-)

bjarnef:
@nielslyngsoe do you mean picking content types (content, media, (member)) from listview configuration, on the content type itself or..?

I have described a few use cases here:
https://github.com/umbraco/Umbraco-CMS/issues/10053

bjarnef:
@nielslyngsoe when working on this I also noticed that `treesourcetypepicker` prevalue editor doesn't work on its own as if requires `currentItemType` to be set, which currently is set in config of MNTP when using `treesource` prevalue editor.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/treesourcetypepicker.controller.js#L38-L40

Unfortunately Umbraco doesn't easily ships with a membertype picker (like we have a property editor for this) `treesourcetypepicker` can't be used alone.

It would have been a nice feature to use in a custom datasource in Contentment (instead of a data source per member type). I can however add a custom prevalue editor to handle this.
// cc @leekelleher 

image

PR #10061, opened 2021-03-25
RecycleBinSmells caching

elit0451:
Adding runtime caching for the RecycleBinSmells value so it is not called too often. The cache is added at the repository level, so that most of our DB caches are at the same level.

It will be evicted by using cache refreshers so that it works in load balancing. The cache is cleared whenever content or media cache is refreshed (executed based on events), so anytime an IContent or IMedia item is updated.


nikolajlauridsen:
Looks great 👍 Tested it out at I can see that the cache gets cleared when creating/deleting content/media.

The only sort of odd thing I found is that the cache is not cleared when clearing the media recycle bin, however looking at the code I assume that this is intended since `_publishedSnapshotService.Notify(payloads, out var anythingChanged);` reports nothing changed, so I'm assuming this is intentional?

elit0451:
Thanks, Nikolaj!

Regarding your comment, maybe @Shazwazza has some more insights on what is the desired behaviour here 🙂

nikolajlauridsen:
Looks good to me 👍 With the new changes the cache is also cleared when emptying the recycle bin 🎉

PR #10063, opened 2021-03-25
Orders member type grouping of members alphabetically, matching the listing of member types.

AndyButland:
This was reported in the Deploy tracker [here](https://github.com/umbraco/Umbraco.Deploy.Issues/issues/3) but the fix is needed in core.

I've just added an alphabetical ordering to the member type listing used to group members in the members section, which matches what is used when member types are ordered within settings.

To Test:

- Create a member type.
- Create another one with a name that starts with a letter earlier in the alphabet than the first one.
- Check they render under the "Members" section in alphabetical order.

### DIFF
--- a/src/Umbraco.Web/Trees/MemberTreeController.cs
+++ b/src/Umbraco.Web/Trees/MemberTreeController.cs
@@ -129,10 +129,12 @@ namespace Umbraco.Web.Trees
 
                 if (_isUmbracoProvider)
                 {
-                    nodes.AddRange(Services.MemberTypeService.GetAll()
-                        .Select(memberType =>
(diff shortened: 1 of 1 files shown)


nathanwoulfe:
@AndyButland this works fine, but I can't merge to v8/dev, only to v8/contrib. I changed the base to v8/contrib, but that caused other commits from v8/dev to be included in the PR (probably not an issue as eventually they'll all be on contrib anyway).

Are you able to re-target your change to contrib?

AndyButland:
Sure, thanks @nathanwoulfe - new PR [here](https://github.com/umbraco/Umbraco-CMS/pull/10077).

PR #10067, opened 2021-03-26
Netcore: Migrate localization service events

nikolajlauridsen:
### Description
Migrate the events in `LocalizationService` to the new `EventAggregator` pattern.

Most of this was fairly straight forward, I did run in to a bit of trouble with the integration tests, there were two issues:
1. Since we now rely on dependency injection to register events, we have to use the registered `IServerMessenger`. The default `IServerMessenger` registered in integration tests is a `NoopServerMessenger` which meant that when the events was raised nothing really happened because the `IServerMessenger` did nothing. This was fixed by registering a `LocalServerMessenger` in the `CustomTestSetup` overridden method.
2. In the `SingleItemsOnlyRepositoryCachePolicy` test, we didn't bind the `DistributedCache` until after we had populated the cache with some test objects, from what I can tell this was done in order to test that the global cache was not changed in a scope. Since we can't unbind/bind the `DistrubutedCache` anymore this caused the test to fail. This was fixed by manually setting the cache to the same state by simply getting the two expected objects from the service, thus populating the cache after it had been cleared.

I noticed that we have an events in the `Umbraco.Infrastructure` project that could be in the `Umbraco.Core`  project, the `DeletingNotification` is in the infrastructure project, but the `DeletedNotification` is in the core project, it seems like `DeletingNotification` should be moved, but I didn't do so, because I don't know if there's a reason for it being in the infrastructure project?

### Testing
This is the testing I've performed.

Existing events: 
1. Set breakpoints in the new `Handle` methods in `DistributedCacheBinder_Handlers`
2. Set a breakpoint in the `Handle` method in `PublishedSnapshotServiceEventHandler`
3. Add/remove a Language and a Dictionary item and ensure that the associated breakpoints are hit

Binding to events:
In the `Umbraco.Web.UI.NetCore` project I created a file `TestHandler` that binds to all the new notifications:

```c#
    public class TestHandler :
        INotificationHandler<DictionaryItemDeletedNotification>,
        INotificationHandler<DictionaryItemDeletingNotification>,
        INotificationHandler<DictionaryItemSavingNotification>,
        INotificationHandler<DictionaryItemSavedNotification>,
        INotificationHandler<LanguageDeletingNotification>,
        INotificationHandler<LanguageDeletedNotification>,
        INotificationHandler<LanguageSavingNotification>,
        INotificationHandler<LanguageSavedNotification>
    {
        public void Handle(DictionaryItemDeletedNotification notification)
        {
            foreach (var entity in notification.DeletedEntities)
            {
                Console.WriteLine($"Dictionary Item deleted {entity.Key}");
            }
        }

        public void Handle(DictionaryItemDeletingNotification notification)
        {
{...}
```

And then registered them with a Composer in the same project: 

```c#
    public class NotificationsComposer : ComponentComposer<NotificationsComponent>
    {
        public override void Compose(IUmbracoBuilder builder)
        {
            base.Compose(builder);

            builder
                .AddNotificationHandler<DictionaryItemDeletedNotification, TestHandler>()
                .AddNotificationHandler<DictionaryItemDeletingNotification, TestHandler>()
                .AddNotificationHandler<DictionaryItemSavingNotification, TestHandler>()
                .AddNotificationHandler<DictionaryItemSavedNotification, TestHandler>()
                .AddNotificationHandler<LanguageDeletingNotification, TestHandler>()
                .AddNotificationHandler<LanguageDeletedNotification, TestHandler>()
                .AddNotificationHandler<LanguageSavingNotification, TestHandler>()
                .AddNotificationHandler<LanguageSavedNotification, TestHandler>();
        }
    }
```

I then set breakpoints in all the `Handle` methods in the `TestHandler` and ensured that they were hit and that I could cancel the "ing" notifications.

### DIFF
--- a/src/Umbraco.Infrastructure/Compose/NotificationsComposer.cs
+++ b/src/Umbraco.Infrastructure/Compose/NotificationsComposer.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Umbraco.
 // See LICENSE for more details.
 
+using Umbraco.Cms.Core.Cache;
 using Umbraco.Cms.Core.Composing;
 using Umbraco.Cms.Core.DependencyInjection;
 using Umbraco.Cms.Core.Events;
@@ -62,6 +63,13 @@ namespace Umbraco.Cms.Core.Compose
                 .AddNotificationHandler<ContentPublishedNotification, RedirectTrackingHandler>()
                 .AddNotificationHandler<ContentMovingNotification, RedirectTrackingHandler>()
                 .AddNotificationHandler<ContentMovedNotification, RedirectTrackingHandler>();
+
+            // Add notification handlers for DistributedCache
+            builder
+                .AddNotificationHandler<DictionaryItemDeletedNotification, DistributedCacheBinder>()
+                .AddNotificationHandler<DictionaryItemSavedNotification, DistributedCacheBinder>()
+                .AddNotificationHandler<LanguageSavedNotification, DistributedCacheBinder>()
+                .AddNotificationHandler<LanguageDeletedNotification, DistributedCacheBinder>();
         }
     }
 }
--- a/src/Umbraco.Infrastructure/Services/Implement/LocalizationService.cs
+++ b/src/Umbraco.Infrastructure/Services/Implement/LocalizationService.cs
@@ -18,7 +18,6 @@ namespace Umbraco.Cms.Core.Services.Implement
     {
         private readonly IDictionaryRepository _dictionaryRepository;
         private readonly ILanguageRepository _languageRepository;
-        private readonly IEventAggregator _eventAggregator;
         private readonly IAuditRepository _auditRepository;
 
         public LocalizationService(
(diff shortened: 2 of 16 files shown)


bergmania:
`DeletingNotification` should just be moved.. Everything that can easily be in Core without introducing new dependencies should be moved :)

src/Umbraco.Infrastructure/Compose/NotificationsComposer.cs
```diff
@@ -62,6 +63,13 @@ public override void Compose(IUmbracoBuilder builder)
                 .AddNotificationHandler<ContentPublishedNotification, RedirectTrackingHandler>()
                 .AddNotificationHandler<ContentMovingNotification, RedirectTrackingHandler>()
                 .AddNotificationHandler<ContentMovedNotification, RedirectTrackingHandler>();
+
```
bergmania, line 66: This is fine for now, but later we should get rid of NotificationsComposer and just make all of this into an extension method on IUmbracoBuilder and call it explicitly.

src/Umbraco.Infrastructure/Services/Implement/LocalizationService.cs
```diff
@@ -463,47 +472,5 @@ private void EnsureDictionaryItemLanguageCallback(IDictionaryItem d)
                 return _dictionaryRepository.GetDictionaryItemKeyMap();
             }
         }
-
```
bergmania, line 466: Now it should be possible to set the class to internal :)

bergmania (changes requested):
Looks good, only small changes required

bergmania (approved):

bergmania:
Can be merged if the buildserver is happy

PR #10069, opened 2021-03-26
Events to Notifications Migration: DataType Service

warrenbuckley:
## Test Notes
Implement/register your own event handlers similar as to PR https://github.com/umbraco/Umbraco-CMS/pull/9996

### DataType
* DataTypeDeleting
* DataTypeDeleted
* DataTypeSaving
* DataTypeSaved
* DataTypeSaving (Collection)
* DataTypeSaved (Collection)
* DataTypeMoving
* DataTypeMoved

### Entity Container aka DataType Folder
* EntityContainerDeleting
* EntityContainerDeleted
* EntityContainerSaving
* EntityContainerSaved
* EntityContainerRenamed


bergmania:
@warrenbuckley , Failing tests and conflicts 🙈

src/Umbraco.Infrastructure/Services/Implement/DataTypeService.cs
```diff
@@ -220,8 +226,7 @@ public Attempt<OperationResult> DeleteContainer(int containerId, int userId = Cm
                     _dataTypeContainerRepository.Save(container);
                     scope.Complete();
 
-                    // TODO: triggering SavedContainer with a different name?!
-                    scope.Events.Dispatch(SavedContainer, this, new SaveEventArgs<EntityContainer>(container, evtMsgs), "RenamedContainer");
+                    scope.Notifications.Publish(new EntityContainerRenamedNotification(container, evtMsgs));
```
kjac, line 229: I get what you're doing here... I just can't help but wonder if I as a 3rd party implementer would expect "saving + saved" notifications (with an option to cancel) here? I know that's not the case as-is, but since we're redoing this whole thing... worth a consideration.

src/Umbraco.Infrastructure/Services/Implement/DataTypeService.cs
```diff
@@ -341,8 +346,10 @@ public Attempt<OperationResult<MoveOperationStatusType>> Move(IDataType toMove,
             using (var scope = ScopeProvider.CreateScope())
             {
                 var moveEventInfo = new MoveEventInfo<IDataType>(toMove, toMove.Path, parentId);
-                var moveEventArgs = new MoveEventArgs<IDataType>(evtMsgs, moveEventInfo);
-                if (scope.Events.DispatchCancelable(Moving, this, moveEventArgs))
+                // var moveEventArgs = new MoveEventArgs<IDataType>(evtMsgs, moveEventInfo);
```
kjac, line 349: Leftover comment 😄

src/Umbraco.Infrastructure/Services/Implement/DataTypeService.cs
```diff
@@ -429,12 +438,14 @@ public void Save(IEnumerable<IDataType> dataTypeDefinitions, int userId = Cms.Co
         /// <param name="raiseEvents">Boolean indicating whether or not to raise events</param>
         public void Save(IEnumerable<IDataType> dataTypeDefinitions, int userId, bool raiseEvents)
         {
+            var evtMsgs = EventMessagesFactory.Get();
             var dataTypeDefinitionsA = dataTypeDefinitions.ToArray();
-            var saveEventArgs = new SaveEventArgs<IDataType>(dataTypeDefinitionsA);
+            //var saveEventArgs = new SaveEventArgs<IDataType>(dataTypeDefinitionsA);
```
kjac, line 443: Leftover comment 😄

kjac:
This looks brilliant! I've left a few comments for you to consider 😄

src/Umbraco.Infrastructure/Services/Implement/DataTypeService.cs
warrenbuckley: OK will add a Moving & Moved

src/Umbraco.Tests.Integration/Cache/DistributedCacheBinderTests.cs
```diff
@@ -47,9 +47,6 @@ public void Can_Find_All_Event_Handlers()
                 new EventDefinition<IUserService, SaveEventArgs<UserGroupWithUsers>>(null, UserService, new SaveEventArgs<UserGroupWithUsers>(Enumerable.Empty<UserGroupWithUsers>())),
                 new EventDefinition<IUserService, DeleteEventArgs<IUserGroup>>(null, UserService, new DeleteEventArgs<IUserGroup>(Enumerable.Empty<IUserGroup>())),
 
-                new EventDefinition<IDataTypeService, SaveEventArgs<IDataType>>(null, DataTypeService, new SaveEventArgs<IDataType>(Enumerable.Empty<IDataType>())),
```
warrenbuckley, line 50: @bergmania I removed this to get test passing, but as the EventArgs for this is no longer a thing. Did I need to replace this with a different approach instead?

nikolajlauridsen: Not Bjarke, but I had to do the same when I migrated Localization events, as far as I can tell we don't have a different approach for this yet.

src/Umbraco.Infrastructure/Cache/DistributedCacheBinder_Handlers.cs
```diff
@@ -66,11 +66,6 @@ public void BindEvents(bool supportUnbinding = false)
             Bind(() => UserService.UserGroupPermissionsAssigned += UserService_UserGroupPermissionsAssigned,
                 () => UserService.UserGroupPermissionsAssigned -= UserService_UserGroupPermissionsAssigned);
 
-            // bind to data type events
-            Bind(() => DataTypeService.Deleted += DataTypeService_Deleted,
-                () => DataTypeService.Deleted -= DataTypeService_Deleted);
-            Bind(() => DataTypeService.Saved += DataTypeService_Saved,
-                () => DataTypeService.Saved -= DataTypeService_Saved);
```
nikolajlauridsen, line 73: We need to hook up to these events again using the new pattern, like we did with LanguageSaved/Deleted etc.

src/Umbraco.Infrastructure/Services/Implement/DataTypeService.cs
```diff
@@ -31,13 +31,14 @@ public class DataTypeService : RepositoryService, IDataTypeService
         private readonly ILocalizationService _localizationService;
         private readonly IShortStringHelper _shortStringHelper;
         private readonly IJsonSerializer _jsonSerializer;
+        private readonly IEventAggregator _eventAggregator;
```
nikolajlauridsen, line 34: IEventAggregator is not used because we're using scope instead, I'm guessing this is something left over? (I did the exact same thing 😁

nikolajlauridsen (changes requested):
Overall looks good, I did find a few things though that I've left as inline comments.

nikolajlauridsen:
Code changes looks good now, but you forgot to register the events for the `DistributedCacheBinder` in the `NotificicationsComposer`, since I had the solution already open, I just quickly added it and pushed, so we're all good on that front 👍 

During testing though I found a bug, if you cancel an entity container creation in the notification you get a `NullReferenceException` error in the front end, but I don't think this is necessarily a problem with your implementation, and suspect that this is a problem in more places than just this. 

During a bit of investigation I found that this is cause because we in the controller assume that if `DataTypeService.CreateContainer` doesn't result in a success then it's because there were an exception, as can be seen in the `PostCreateContainer` action:

`DataTypeController.PostCreateContainer`
```c#
        public IActionResult PostCreateContainer(int parentId, string name)
        {
            var currentUser = _backOfficeSecurityAccessor.BackOfficeSecurity.CurrentUser;
            var result = _dataTypeService.CreateContainer(parentId, name, currentUser.Id);

            if (result.Success)
                return Ok(result.Result); //return the id
            else
                return ValidationErrorResult.CreateNotificationValidationErrorResult(result.Exception.Message);
        }
```

Here the exception in `result.Exception.Message` will be null, because well, no exception was thrown, we just canceled it. 


I can "fix" this by adding a condition for `OperationResultType.FailedCancelledByEvent` like so:
```c#
        public IActionResult PostCreateContainer(int parentId, string name)
        {
            var currentUser = _backOfficeSecurityAccessor.BackOfficeSecurity.CurrentUser;
            var result = _dataTypeService.CreateContainer(parentId, name, currentUser.Id);

            if (result.Success)
                return Ok(result.Result); //return the id
            if (result.Result.Result == OperationResultType.FailedCancelledByEvent)
                return ValidationErrorResult.CreateNotificationValidationErrorResult(result.Result.EventMessages.GetAll().First().Message);
            else
                return ValidationErrorResult.CreateNotificationValidationErrorResult(result.Exception.Message);
        }
```

Now this is far from great either, because it will show two notifications in the UI, one from the notification, and one from the `ValidationErrorResult` and if there's more that one, only one will be show, but it was more just a quick "test fix" to understand the problem better. 

I'm not sure how we best handle this, but as I also said, this is not necessarily limited to this specific case, I'd guess that we do the same thing in more places, so maybe fixing this should be a separate task, thoughts

nikolajlauridsen:
Tested the rest of the notifications and can see that it happens for renaming a folder as well.

Generally how we handle cancelling of operations isn't super consistent, sometimes, for instance when canceling deleting a datatype, the UI acts as if the deletion was successful, but when reloading the UI the item is still there, other times it shows an error, and the eventmessage, so this seems a bit messy right now.

nikolajlauridsen (approved):
I just tested against the netcore/dev branch with the old events and can confirm that it's a problem there as well, so I'd say we just merge this and create a separate task for fixing the other problem.

PR #10076, opened 2021-03-29
Media Types Video, Sound, Document and Vector Graphics

nielslyngsoe:
Adding new Media Types, Media Types can be used for Allowed Media Types in the upcoming Media Picker v.3.

Media Types are differentiated by file extensions, here goes the definitions:

Video: mp4, webm, ogv
Audio: mp3, weba, oga, opus
Document: pdf, docx, doc, xlsx, xls, pptx, ppt, epub, txt, rtf, csv
Vector Graphics: svg

image

Test notes:

Make sure that new types can be created in folders
make sure new types can be created at root.


src/Umbraco.Core/Constants-Conventions.cs
```diff
@@ -118,6 +118,26 @@ public static class MediaTypes
                 /// </summary>
                 public const string Image = "Image";
 
+                /// <summary>
+                /// MediaType alias for an video-.
+                /// </summary>
+                public const string Video = "Video";
```
marcemarc, line 124: One thing, I just thought reading through this draft thing (exciting, I like the icon for sound!) would people already have custom MediaTypes in their project with these Alias... ? just thinking about upgrades - I know I always use 'Video' for alias of a custom MediaType that points to an external video resource... anyway as the thought occurred, I thought I'd say it out loud, but sorry, if I've misconstrued!

src/Umbraco.Core/Constants-Icons.cs
```diff
@@ -59,6 +59,26 @@ public static class Icons
...
+
+            /// <summary>
+            /// System media document icon
+            /// </summary>
+            public const string MediaDocument = "icon-article";
+
+            /// <summary>
+            /// System media vector icon
+            /// </summary>
+            public const string MediaVector = "icon-picture";
```
bjarnef, line 80: It would be great if this could use a different icon since it is a different media type, maybe a shape icon as SVG could be any graphic, icon, etc. and not neccesary a photo/picture.

src/Umbraco.Core/Constants-Conventions.cs
nielslyngsoe: Perfect insight, thanks for sharing this. The new Media Types only gets created for new installs, so shouldn't be a concern. Existing installations will have to setup the types on their own.
If I´m not mistaken the configuration of file extensions accepted by a Upload-field is already part of the pre-values in the latest release.

src/Umbraco.Core/Constants-Icons.cs
nielslyngsoe: Could this Geometric icon be used?
image

nielslyngsoe: Otherwise this could be one:
image

bjarnef: Yeah, one of those could work. It is difficult to pick the right icon, but at least it would be great with a different icon for each media type.

I noticed Nik asked on Snack for a specific media type for archives.
Should we have one for zip, rar etc. which could use `icon-compress`.

bjarnef: Some other archive file formats are `.tar`, `.iso` or `.7z`.

bjarnef:
@nielslyngsoe is this PR still relevant? It seems we also have this implemented now?

nathanwoulfe:
@bjarnef you're right, these are out in the wild now. @nielslyngsoe I'll close this, but feel free to re-open if we've missed something 😄

PR #10080, opened 2021-03-30
Netcore: Migrate fileservice events

nikolajlauridsen:
### Description
Migrated the static events in `FileService` over to the new notifications pattern. 

There are a few interesting things this time. 

#### Stylesheet

For some reason `DistributedCacheBinder_Handler` bind to the `DeletedStylesheet` and `SavedStylesheet` event, but do nothing with the events, the methods are just empty. There is a note saying "TODO: our weird events handling wants this for now", but I don't see any reason why this should be necessary for the notifications pattern, so I've just removed the event bindings.

#### Templates

This one used what seems like an old pattern, adding "Additional data"  to the event, I added this to the Notifications classes instead.

This data is then used in the `ModelsBuilderNotificationHandler`, in the same class we also used the `UmbracoApplicationStarting` notification to bind to the `SavingTemplate` event, since we can no longer bind/unbind to events, I moved the check (`_config.ModelsMode != ModelsMode.Nothing`) into the handle method, and then removed the `UmbracoApplicationStarting` handler, since it's no longer used for anything. 

### Testing
The same as the other event PRs, take look at #10075, the approach is the same, just using different events.

### DIFF
--- a/src/Umbraco.Infrastructure/Services/Implement/FileService.cs
+++ b/src/Umbraco.Infrastructure/Services/Implement/FileService.cs
@@ -75,10 +75,11 @@ namespace Umbraco.Cms.Core.Services.Implement
         /// <inheritdoc />
         public void SaveStylesheet(IStylesheet stylesheet, int userId = Cms.Core.Constants.Security.SuperUserId)
         {
-            using (var scope = ScopeProvider.CreateScope())
+            using (IScope scope = ScopeProvider.CreateScope())
             {
-                var saveEventArgs = new SaveEventArgs<IStylesheet>(stylesheet);
-                if (scope.Events.DispatchCancelable(SavingStylesheet, this, saveEventArgs))
+                EventMessages eventMessages = EventMessagesFactory.Get();
+                var savingNotification = new StylesheetSavingNotification(stylesheet, eventMessages);
+                if (scope.Notifications.PublishCancelable(savingNotification))
                 {
                     scope.Complete();
                     return;
@@ -86,8 +87,7 @@ namespace Umbraco.Cms.Core.Services.Implement
 
 
                 _stylesheetRepository.Save(stylesheet);
-                saveEventArgs.CanCancel = false;
-                scope.Events.Dispatch(SavedStylesheet, this, saveEventArgs);
+                scope.Notifications.Publish(new StylesheetSavedNotification(stylesheet, eventMessages).WithStateFrom(savingNotification));
                 Audit(AuditType.Save, userId, -1, "Stylesheet");
 
                 scope.Complete();
@@ -200,18 +200,18 @@ namespace Umbraco.Cms.Core.Services.Implement
         /// <inheritdoc />
         public void SaveScript(IScript script, int userId = Cms.Core.Constants.Security.SuperUserId)
         {
-            using (var scope = ScopeProvider.CreateScope())
+            using (IScope scope = ScopeProvider.CreateScope())
             {
-                var saveEventArgs = new SaveEventArgs<IScript>(script);
-                if (scope.Events.DispatchCancelable(SavingScript, this, saveEventArgs))
+                EventMessages eventMessages = EventMessagesFactory.Get();
+                var savingNotification = new ScriptSavingNotification(script, eventMessages);
+                if (scope.Notifications.PublishCancelable(savingNotification))
                 {
                     scope.Complete();
                     return;
                 }
 
                 _scriptRepository.Save(script);
-                saveEventArgs.CanCancel = false;
-                scope.Events.Dispatch(SavedScript, this, saveEventArgs);
+                scope.Notifications.Publish(new ScriptSavedNotification(script, eventMessages).WithStateFrom(savingNotification));
 
                 Audit(AuditType.Save, userId, -1, "Script");
                 scope.Complete();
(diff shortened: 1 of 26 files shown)


src/Umbraco.Infrastructure/Services/Implement/FileService.cs
```diff
@@ -342,27 +344,23 @@ public long GetScriptFileSize(string filepath)
...
                 {
                     scope.Complete();
-                    return OperationResult.Attempt.Fail<OperationResultType, ITemplate>(OperationResultType.FailedCancelledByEvent, evtMsgs, template);
+                    return OperationResult.Attempt.Fail<OperationResultType, ITemplate>(OperationResultType.FailedCancelledByEvent, eventMessages, template);
                 }
 
                 _templateRepository.Save(template);
-                saveEventArgs.CanCancel = false;
-                scope.Events.Dispatch(SavedTemplate, this, saveEventArgs);
+                scope.Notifications.Publish(new TemplateSavedNotification(template, eventMessages, savingEvent.AdditionalData).WithStateFrom(savingEvent));
```
bergmania, line 357: I think we will need to do something similar to `WithStateFrom(savingEvent)` for the `AdditionalData`, otherwise that object is not shared between the `TemplateSavingNotification` and `TemplateSavedNotification` like it was before

bergmania: Nevermind, I see you are using that explict type when you create the `TemplateSavedNotification `

src/Umbraco.Infrastructure/Services/Implement/FileService.cs
```diff
@@ -318,9 +320,9 @@ public long GetScriptFileSize(string filepath)
                 // This fixes: http://issues.umbraco.org/issue/U4-7953
                 contentTypeName);
 
-            var evtMsgs = EventMessagesFactory.Get();
+            EventMessages eventMessages = EventMessagesFactory.Get();
 
-            // TODO: This isn't pretty because we we're required to maintain backwards compatibility so we could not change
+            // TODO: This isn't pretty because we're required to maintain backwards compatibility so we could not change
             // the event args here. The other option is to create a different event with different event
             // args specifically for this method... which also isn't pretty. So fix this in v8!
```
bergmania, line 327: I guess we should fix this TODO now, instead of having the Additional Data?

nikolajlauridsen: Or we could of course add a `CreateTemplateForContentType` and a `ContentTypeAlias` field to the Saving/Saved Notification, I'm not sure what's better?

bergmania: I think we both are better options then the current :) But maybe a hybrid, where we have properties that use the state as backing field, so we are sure the values are kept between the Saving...Saved notifications?

nikolajlauridsen: Oh that's a great idea :D

src/Umbraco.Infrastructure/Services/Implement/FileService.cs
```diff
@@ -346,15 +346,15 @@ public long GetScriptFileSize(string filepath)
 
             using (IScope scope = ScopeProvider.CreateScope())
             {
-                var savingEvent = new TemplateSavingNotification(template, eventMessages, additionalData);
+                var savingEvent = new TemplateSavingNotification(template, eventMessages, true, contentTypeAlias);
```
bergmania, line 349: Can't we remove the `additionalData` variable now?

src/Umbraco.Infrastructure/Services/Notifications/TemplateSavedNotification.cs
```diff
@@ -17,12 +17,24 @@ public TemplateSavedNotification(IEnumerable<ITemplate> target, EventMessages me
...
-        public TemplateSavedNotification(ITemplate target, EventMessages messages,
-            Dictionary<string, object> additionalData) : base(target, messages) => AdditionalData = additionalData;
-
-        public TemplateSavedNotification(IEnumerable<ITemplate> target, EventMessages messages,
-            Dictionary<string, object> additionalData) : base(target, messages) => AdditionalData = additionalData;
+        public bool? CreateTemplateForContentType
+        {
+            get
+            {
+                State.TryGetValue("CreateTemplateForContentType", out var result);
```
bergmania, line 24: `"CreateTemplateForContentType"` should be a private const

src/Umbraco.Infrastructure/Services/Notifications/TemplateSavedNotification.cs
```diff
@@ -17,12 +17,24 @@ public TemplateSavedNotification(IEnumerable<ITemplate> target, EventMessages me
...
+            }
+            set => State["CreateTemplateForContentType"] = value;
+        }
 
-        public Dictionary<string, object> AdditionalData;
+        public string ContentTypeAlias
+        {
+            get
+            {
+                State.TryGetValue("ContentTypeAlias", out var result);
```
bergmania, line 34: `"ContentTypeAlias"` should be a private const

bergmania (changes requested):

src/Umbraco.Infrastructure/Services/Notifications/TemplateSavedNotification.cs
```diff
@@ -17,12 +17,24 @@ public TemplateSavedNotification(IEnumerable<ITemplate> target, EventMessages me
         {
         }
 
-        public TemplateSavedNotification(ITemplate target, EventMessages messages,
-            Dictionary<string, object> additionalData) : base(target, messages) => AdditionalData = additionalData;
-
-        public TemplateSavedNotification(IEnumerable<ITemplate> target, EventMessages messages,
-            Dictionary<string, object> additionalData) : base(target, messages) => AdditionalData = additionalData;
+        public bool? CreateTemplateForContentType
```
bergmania, line 23: I guess this should just be bool, and default to false

nikolajlauridsen: Ah true, I change it so it just defaults to false

bergmania (approved):

PR #10084, opened 2021-03-30
Use umb-icon component in listview dropdown

bjarnef:
### Prerequisites

### Description
This PR update the icons in listview action dropdown to use `<umb-icon>` component, so if using custom SVG icons for content types, these will work here as well, where it currently only works with font icons.


nathanwoulfe (approved):
Looks good to me, thanks @bjarnef. Hopefully as part of the backoffice UI rebuild, the icon directive can be improved to avoid the need to pass the same value for icon and class attributes - surely the value from icon can internally be set as a classname... Would simplify the directive even further, so the class attribute would only be required for cases where the class is in addition to the icon name.

Issue #9962, opened 2021-03-10
Unable to move grid block


AaronSadlerUK:
As seen on the forum here:
https://our.umbraco.com/packages/website-utilities/umf/bug-reports/104700-this-content-is-now-allowed-here-prevents-reordering, this breaks one of the main and fundamental reasons for using grid.

The post is for an earlier version, however it still exists in 8.10.2.+

It seems it is not possible to drag and drop grid blocks if allow all editors is enabled.

The workaround for now involves commenting out a line in a core file:

```
  /*if (area.$allowedEditors) {
      allowedEditors = (allowedEditors ?? []).concat(area.$allowedEditors.map(e => e.alias));
  }*/
```

`src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js`

nul800sebastiaan:
I don't know if that's the exact fix but it would be good to investigate what that if statement does and how it should be ammended. 👍

umbrabot:
Hi @AaronSadlerUK,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

AaronSadlerUK:
@nul800sebastiaan surely this is a core bug, given that one of the fundamental features of grid is the drag and drop?

nul800sebastiaan:
Hi @AaronSadlerUK, certainly it's a core bug. The way we work with incoming bug reports is usually: if we broke it recently then we fix it. If it's been broken for a long time then it seems not many people are running into the problem, we kindly ask for some help in getting it fixed. 
While we would love to dive in an fix each and every bug on the issue tracker, that unfortunately doesn't scale, so we rely on the help of our community for some of this.

AaronSadlerUK:
I noticed 3 people last night having the same issue on Facebook, others as on the forum post are also running into this issue.
It might be that editors haven't reported it (we know that happens a lot).

Just concerned that this seems to have been around a while, and as such won't be picked up and fixed in a timely manner.
Unfortunately I don't know angular otherwise I would look into it.

[...]

PR #10086, opened 2021-03-30
9962: Use $allowedEditors instead of allowed

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9962

How to check (based on the default install):
Set all grid rows on "Allow all editors". Without this fix, you'll not be able to move between the different rows.

The issue here was that the .allow only has the editors selected by the user. So if you were to only select RTE, it would work because .allow is ["RTE"]. If you use the selectAll editors, it will set allowAll to true, but leave .allow empty.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js
@@ -143,7 +143,7 @@ angular.module("umbraco")
                     var area = event.target.getScope_HackForSortable().area;
                     var allowedEditors = area.$allowedEditors.map(e => e.alias);
 
-                    if (($.inArray(ui.item[0].getScope_HackForSortable().control.editor.alias, allowedEditors) < 0 && allowedEditors) ||
+                    if (($.inArray(ui.item[0].getScope_HackForSortable().control.editor.alias, allowedEditors) < 0) ||
                         (startingArea != area && area.maxItems != '' && area.maxItems > 0 && area.maxItems < area.controls.length + 1)) {
 
                         $scope.$apply(function () {
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js
@@ -141,7 +141,7 @@ angular.module("umbraco")
                 over: function (event, ui) {
 
                     var area = event.target.getScope_HackForSortable().area;
-                    var allowedEditors = area.allowed;
+                    var allowedEditors = area.$allowedEditors.map(e => e.alias);
 
                     if (($.inArray(ui.item[0].getScope_HackForSortable().control.editor.alias, allowedEditors) < 0 && allowedEditors) ||
                         (startingArea != area && area.maxItems != '' && area.maxItems > 0 && area.maxItems < area.controls.length + 1)) {


mikecp:
Hello @patrickdemooij9 ,

Thanks for this PR!

It took me some time to find how to reproduce this, but eventually I ended up reproducing the behavior: this happens when reordering within/moving to a column which has the property `maxItems` set to a specific value.

For some reason, in "Allow All Editors" scenario,, when maxItems is not set, the value of area.Allowed is `undefined`, while it is an empty array `[]` when maxItems is specified.

This empty array creates an unexpected behavior in the test that follows, which basically checks that you are moving an allowed editor to the target cell or that you still have some space left for an extra editor in the target cell;

```
if (($.inArray(ui.item[0].getScope_HackForSortable().control.editor.alias, allowedEditors) < 0 && allowedEditors) ||
    (startingArea != area && area.maxItems != '' && area.maxItems > 0 && area.maxItems < area.controls.length + 1))
``` 

The check on moving an allowed editor works as expected when there is no maxItems specified, because in that case allowedEditors is `undefined` and therefore the check on `&& allowedEditors` has the foreseen effect of making that part of the condition false anyway. 
But when maxItems is specified and allowedEditors is an empty array, the test `&& allowedEditors` will in fact evaluate to true and therefore the global check will take the first part of the condition into account, which returns true as well since allowedEditors is an empty array and the editor will not be found in the list... This makes the whole condition true and you enter in an incorrect "not allowed" scenario.

In addition to this, the condition itself seems written in the wrong order: why not first check to see if you are in an "allow all editors" scenario before (not) checking that you are moving an allowed editor?

All this to say that I have the impression that it might  be more interesting to keep the `area.allowed` value but rewrite that first part of the condition.

So something like the following:

```
//var allowedEditors = area.allowed; /*Only one usage anymore so can directly be used in the condition */

if ((!area.allowAll && $.inArray(ui.item[0].getScope_HackForSortable().control.editor.alias, area.allowed) < 0) ||
   (startingArea != area && area.maxItems != '' && area.maxItems > 0 && area.maxItems < area.controls.length + 1)) {
```

Enough writing for now 😀
I am wondering what is your feeling about this?

patrickdemooij9:
Hey @mikecp,

Thanks for the response to the PR! I checked where the issue of the undefined and empty array is coming from. Apparently, the PostSave of the dataType is only sending the empty array in the request when the max items are set. Is this something that should be fixed too? I assume it should always send an empty array?

> The check on moving an allowed editor works as expected when there is no maxItems specified, because in that case allowedEditors is undefined and therefore the check on && allowedEditors has the foreseen effect of making that part of the condition false anyway.
But when maxItems is specified and allowedEditors is an empty array, the test && allowedEditors will in fact evaluate to true and therefore the global check will take the first part of the condition into account, which returns true as well since allowedEditors is an empty array and the editor will not be found in the list... This makes the whole condition true and you enter in an incorrect "not allowed" scenario.

Is this true though? In the old scenario, it would indeed give a false statement (and true if the array existed) due to the allowedEditors check. But then the first part of the statement would give false anyway, right? I'll write some scenarios based on the old code: `($.inArray(ui.item[0].getScope_HackForSortable().control.editor.alias, allowedEditors) < 0 && allowedEditors`

Scenario 1:
The .allowed returns ["rte", "textarea", "something"] and the item we are moving has the alias "textarea". This will then return true on both sides of the statement.

Scenario 2:
The .allowed returns undefined and the item we are moving has the alias "textarea". The $.inArray will return -1, making the first part of the statement false.

Scenario 3:
The .allowed returns [] and the item we are moving has the alias "textarea". The $.inArray will return -1, making the first part of the statement false.

It might be that I am missing something here, but I don't see how the statement returns true when the .allowed is an empty array. But please correct me if I am wrong!

And then the reason why I am using $allowedEditors here is that it already has the logic for the allowAll applied in it. That code for that is here: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/grid/grid.controller.js#L848. And the $allowedEditors is also used in other parts of the code instead of .allowed. Yes, we can use .allowed here, but I think using $allowedEditors would be safer and more consistent here.

As for the if statement. It does look like I can tidy that up a bit better. With my change to the code, the `&allowedEditors` check is not needed anymore.

Let me know what you think!

mikecp:
Hey @patrickdemooij9 ,

Thanks for the clarification about your usage of `$allowedEditors`, it makes more sense to use it indeed! So it's all ready to merge 👍 

Just for the record, I think you may have skipped the fact that the `inArray` condition checks on the result being less than 0, not greater than 0 😉 

So in your scenario's for old behavior:
1/ inArray returns 1 which is not < 0 => evaluated to false
    allowedEditors evaluated to true 
=> result for that part of the condition is false, which is correct ("textarea" editor allowed)

2/ inArray < 0 => true
    allowedEditors evaluated to false
=> result for that part is false, which is correct (all editors allowed)

3/ inArray < 0 => true
allowedEditors evaluated to true
=> result for that part is true, which is _incorrect_ since all editors are allowed

This last scenario was indeed the situation that I mentioned that was creating the bug and preventing the move although all editors were allowed.

Thanks again for your PR! It will solve some tricky wrong behavior of the grid 👍   

Cheers!

mikecp (approved):

patrickdemooij9:
Hi @mikecp ,
You are correct about the scenarios being wrong. I totally looked over the < 0 instead of it being > 0 😅

Thanks for merging the PR. I am sure a lot of people will appreciate this fix

mikecp:
I think too 😀 👍

PR #10089, opened 2021-03-31
Netcore: Migrate relationservice events

nikolajlauridsen:
### Description

Migrated the static events in `RelationService` over to the new notifications pattern.

#### DeleteRelationsOfType (RelationDeletedNotification)
For some reason we don't raise the `RelationDeletingNotification` here, only the `DeletedRelationNotification`, I'm not sure I understand why we don't, we could just publish one cancelable notification with the list of relations?

#### DeletedNotification

DeletedNotification didn't have a constructor for IEnumerable<T>, even though it inherited EnumerableObjectNotification, I'm assuming it's because this wasn't used until now, however, it's needed for the above, so I added it.

#### RelationSavingNotification
In the `Save(IEnumerable<IRelation> relations)` method I got a multiple enumeration warning, so I turned the `IEnumerable<IRelation>` into an array.

### Testing

The same as the other event PRs, take look at #10075, the approach is the same, just using different events.

### DIFF
--- a/src/Umbraco.Core/Events/DeletedNotification.cs
+++ b/src/Umbraco.Core/Events/DeletedNotification.cs
@@ -11,6 +11,10 @@ namespace Umbraco.Cms.Core.Events
         {
         }
 
+        protected DeletedNotification(IEnumerable<T> target, EventMessages messages) : base(target, messages)
+        {
+        }
+
         public IEnumerable<T> DeletedEntities => Target;
     }
 }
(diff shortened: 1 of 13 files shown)


bergmania:
> For some reason we don't raise the RelationDeletingNotification here, only the DeletedRelationNotification, I'm not sure I understand why we don't, we could just publish one cancelable notification with the list of relations?

I agree that seems strange, but if we fix that, we should fix that for v8 too. Not as part of this PR.

PR #10091, opened 2021-03-31
Fixed copy preserving sort order

### LINKED ISSUE #8033: Copy content node with descendants is not preserving sort

ricbrady:
Fixed the order of the Descendants nodes during the copy
I've also created a Unit Test to test it

Fixes #8033

### DIFF
--- a/src/Umbraco.Core/Services/Implement/ContentService.cs
+++ b/src/Umbraco.Core/Services/Implement/ContentService.cs
@@ -1415,7 +1415,7 @@ namespace Umbraco.Core.Services.Implement
 
                         var result = CommitDocumentChangesInternal(scope, d, saveEventArgs, allLangs.Value, d.WriterId);
                         if (result.Success == false)
-                            Logger.Error<ContentService,int,PublishResultType>(null, "Failed to publish document id={DocumentId}, reason={Reason}.", d.Id, result.Result);
+                            Logger.Error<ContentService, int, PublishResultType>(null, "Failed to publish document id={DocumentId}, reason={Reason}.", d.Id, result.Result);
                         results.Add(result);
 
                     }
@@ -2201,7 +2201,7 @@ namespace Umbraco.Core.Services.Implement
                     while (page * pageSize < total)
                     {
                         var descendants = GetPagedDescendants(content.Id, page++, pageSize, out total);
-                        foreach (var descendant in descendants)
+                        foreach (var descendant in descendants.OrderBy(x => x.Level).ThenBy(y => y.SortOrder))
                         {
                             // if parent has not been copied, skip, else gets its copy id
                             if (idmap.TryGetValue(descendant.ParentId, out parentId) == false) continue;
@@ -2420,7 +2420,7 @@ namespace Umbraco.Core.Services.Implement
                 if (report.FixedIssues.Count > 0)
                 {
                     //The event args needs a content item so we'll make a fake one with enough properties to not cause a null ref
-                    var root = new Content("root", -1, new ContentType(-1)) {Id = -1, Key = Guid.Empty};
+                    var root = new Content("root", -1, new ContentType(-1)) { Id = -1, Key = Guid.Empty };
                     scope.Events.Dispatch(TreeChanged, this, new TreeChange<IContent>.EventArgs(new TreeChange<IContent>(root, TreeChangeTypes.RefreshAll)));
                 }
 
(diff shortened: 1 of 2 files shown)


ricbrady:
Hi @OwainWilliams 

thank you.
Why this test errors? they are not about my changes.

OwainWilliams:
Hi @ricbrady  - No, no need to worry about the tests failing. It's nothing to do with your code 😊

OwainWilliams (approved):
Tested and everything works as expected. Thanks for the added tests too.

OwainWilliams:
Hey @ricbrady - This has now been merged in! 
Can you let us know what your Our umbraco account is e.g. (https://our.umbraco.com/members/yourusername/) so that we can add a contrib badge on your profile. Thanks.

ricbrady:
Hey @OwainWilliams 

it is https://our.umbraco.com/members/ricbrady/ . I've just created it.
Thanks

ricbrady:
Hi @nul800sebastiaan 
why is this fix added to release 8.14 and not any more to 8.13?
It's something very important in my Environment.
Thnaks

nul800sebastiaan:
8.13 was already closed for updates at the time of merging this, all new fixes go into the next major version which is being worked on for the next 6 weeks. 

We generally only apply patches for regression bugs (something we broke in recent minors) or if it is an urgent fix for some reason or another.

In this case, it's not a regression (it always worked this way) and as a workaround, until 8.14 is out, you can hook into the `ContentService` copy events fixing the sortorder after copying.

nul800sebastiaan:
Please also note this introduced a problem with the `Umbraco.Tests.Integration.ContentEventsTests.CopyBranch` unit test which is no longer passing. We'll need to have a closer look if this PR breaks more than it fixes.

src/Umbraco.Core/Services/Implement/ContentService.cs
```diff
@@ -2201,7 +2201,7 @@ public IContent Copy(IContent content, int parentId, bool relateToOriginal, bool
                     while (page * pageSize < total)
                     {
                         var descendants = GetPagedDescendants(content.Id, page++, pageSize, out total);
-                        foreach (var descendant in descendants)
+                        foreach (var descendant in descendants.OrderBy(x => x.Level).ThenBy(y => y.SortOrder))
```
Shazwazza, line 2204: This should be sorted via SQL/Database, this will not perform well.

nul800sebastiaan: I tried passing "level" to the `GetPagedDescendants` method: 

`var descendants = GetPagedDescendants(content.Id, page++, pageSize, out total, ordering: Ordering.By("level"));`

But that doesn't work because we end up in `ContentRepositoryBase.ApplySystemOrdering` and `level` is not a supported field there. I don't know if it can "just" be added but even if we could then `Umbraco.Core.Services.Ordering` does not seem to support ordering on multiple columns so we wouldn't be able to construct the same ordering that being doing in Linq here.

nul800sebastiaan:
Due to the problems with the unit test and the performance problem that @Shazwazza noted, I will have to revert this change for now. Would be awesome if someone wanted to give this another try.

FYI: I did fix the tests by giving them the correct order: 

```csharp
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copy.Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copyC[0].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copyC[1].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copyC[2].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copyC[3].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copy2C[0].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copy3C[0].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copy4C[0].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copy5C[0].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copy2C[1].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copy3C[1].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m++:000}: ContentRepository/Refresh/{copy4C[1].Id}.u=u", _events[i++].ToString());
            Assert.AreEqual($"{m:000}: ContentRepository/Refresh/{copy5C[1].Id}.u=u", _events[i++].ToString());
```

However I am pretty concerned that this PR might be causing weird unexpected behavior, or maybe the initial test relied on the weird behavior that was wrong to begin with..

PR #10093, opened 2021-03-31
Netcore: Made internals visible to Umbraco Forms assemblies

AndyButland:
This replicates what we have in V8, where we make internal classes and methods available to Umbraco Forms.


bergmania:
@AndyButland what internal types do you need in forms? I guess we should be able to implement forms without such hacks,  Maybe some internal classes needs to be public

AndyButland:
When I've finished migrating the projects I'll revert back to depending on alpha 4, and then can get a list.

PR #10094, opened 2021-04-01
Update icon in member group picker

bjarnef:
### Prerequisites

### Description
Currently we have a Member Group Property editor here:
https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/membergrouppicker

but we also have Member Group picker here, which is used on the core member group picker on a member.
https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/membergroups

Not sure why we have two as they a pretty much doing the same thing. The newer picker was probably added as part of this infinite editor, where the older picker seems to store a different format , while I guess the new one just stores an comma delimited string id of int ids.
However the new picker used the icon property, but this wasn't mapped to any value in the mapping definition and therefore was null.

However I noticed the first mentioned doesn't show an icon like the second picker.  Furthermore unlike most other pickers it actually doesn't use the entity icon. Currently it is not possible to set a specific icon for a member group, but maybe in future.

0yD64Q8ZaS

With the changed in this PR the two member group pickers are a bit more consistent and use the member group entity icon instead, which is more consistent with other pickers like Content Picker, MNTP, etc.

j9fJXjQKFc


bjarnef:
Furthermore it seems the `$scope.add` function was previous used, but it usage was removed here, but didn`t cleanup that function:
https://github.com/umbraco/Umbraco-CMS/commit/bb53c020e90ecf6612a35b8672091d9469292946#diff-fd3363b8299d7c8b77cfa240b85491529a67a2aa91abfc32efdb5b1caa457eda

bjarnef:
There is this other open PR https://github.com/umbraco/Umbraco-CMS/pull/9467 to support fetching Member Group by `Udi` and `Guid` as well, so we could maybe later in a new version store member group picker items using udi like other pickers 😁👍🚀

bjarnef:
The references of `remove` and `clean` functions was removed in https://github.com/umbraco/Umbraco-CMS/commit/bb53c020e90ecf6612a35b8672091d9469292946#diff-fd3363b8299d7c8b77cfa240b85491529a67a2aa91abfc32efdb5b1caa457eda so I have removed these as they are no longer used.

nathanwoulfe (approved):
@bjarnef, lovely work as always 😄

nathanwoulfe:
Annnnd just saw the conflicts. @bjarnef feel like dealing with those?

bjarnef:
@nathanwoulfe ahh, need to fix the merge conflicts.. 🙈🤔

bjarnef:
@nathanwoulfe I have resolved the merge conflicts.

I have also removed a few comments from @Shazwazza in https://github.com/umbraco/Umbraco-CMS/pull/10013 as the `remove` function is used and the `clear` function wasn't used, BUT I have this another open PR, where it would use the function the clear the property https://github.com/umbraco/Umbraco-CMS/pull/10096

PR #10096, opened 2021-04-02
Member group picker actions

bjarnef:
### Prerequisites

### Description
This PR add the property action to member group picker to remove all items similar to other pickers.
Some of the adjustments from https://github.com/umbraco/Umbraco-CMS/pull/10094 is included as well, so it should be a bit easier to merge.

U2dr8cNrTK

This could also be useful in the member group picker used by default on members: https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/membergroups


nathanwoulfe (approved):
Hi @bjarnef I took this for a spin, works as advertised 👍

PR #10099, opened 2021-04-04
Improved accessibility of link picker

RachBreeze:
As a result of investigating issue #8704 I have added label fors and ids to many of the controls on the link picker.

This allows screen readers to identify many of the controls on the link picker.

PR #10100, opened 2021-04-04
Accessibility: Alerts the user how many results have been returned on a tree search

RachBreeze:
Hello,

Whilst working on #10099 I realised that the screen reader user had no information on the number of results returned from a search on the link picker so wouldn't know how many items they would have to work through until they found the link they wanted to work with. I have updated the tree search to report on the number of items found. Currently I have left this as visible, as I felt this could benefit many users. I am however happy to change this so that only screen reader users hear the number of results


mikecp (approved):

mikecp:
Hi @RachBreeze ,
I think this can indeed be useful to everybody, especially if there are lots of results, so I would leave it visible to all like you implemented it.
Thanks! 👍

PR #10101, opened 2021-04-04
Accessibility: Adding label fors and control ids for the macro picker

RachBreeze:
As a result of investigating issue #8704 I have added label fors and ids to many of the controls on the macro picker.

This allows screen readers to identify many of the controls on the link picker.


RachBreeze (commented):
Sorry the last commit message is wrong, the change allows screen reader users to have an alert on the number of matching macros returned in a search

mikecp:
Hello @RachBreeze , 
Thanks for this new accessibility improvement! 👍 

I was just wondering if there was a particular reason why you kept the display of the number of elements found as "sr-only" here and displayed to all in the link picker. Maybe it could be interesting to have the same behavior in both and display this information to everybody in the macro picker as well?

RachBreeze:
Hello @mikecp 
The reason for hiding the message was just that in my test cases I only had a handful of macros. However you are correct to be consistent it should show the same messages as the link picker, and just because I only have a handful doesn't mean others won't have more. I have updated the display now. Thank you for your review

mikecp (approved):

mikecp:
Hello @RachBreeze ,

Thanks for the update, we're all set to merge now :-)

Issue #10103, opened 2021-04-06
Support Upload in Block List


bjarnef:
Based on Umbraco v8.12.2 it seems the `Umbraco.Upload` is currently not supported in block list.
Is there a specific reason this isn't supported?

It could be useful in a list of files, where a "Block File" is used.

![image](https://user-images.githubusercontent.com/2919859/113680850-98b84600-96c1-11eb-9042-9003ea0f7b56.png)

A media picker could be used, but we don't always want to create uploads in media section.
For this specific use-case the block list items are created dynamically via a form, where it is possible to use set a title/name and select a file via file upload.

nul800sebastiaan:
I can't remember exactly why this isn't possible, but unfortunately we've tried multiple times and have not been able to make this work. Therefore you will get this error instead.

bjarnef:
Yes, it could be useful though, but not sure if some if there is something technically preventing this from working. For the specific use-case I have used media picker instead, where I however don't really need to re-use the media picker thus a file upload would have been great.

bjarnef:
I just tried comment out this line with `Umbraco.UploadField`:

https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js#L232

The UI part seems to be working:

![image](https://user-images.githubusercontent.com/2919859/114071749-c52ab880-98a1-11eb-8172-5c7f058c99db.png)

But it breaks here on save because content guid is empty:

![image](https://user-images.githubusercontent.com/2919859/114071706-b47a4280-98a1-11eb-885a-327fa2479226.png)

![image](https://user-images.githubusercontent.com/2919859/114071490-7846e200-98a1-11eb-8964-ec12288a7db0.png)

bjarnef:
I had a look at it here https://github.com/umbraco/Umbraco-CMS/pull/10115 but it seems something is missing. However I am able to render to file name 😅

nielslyngsoe:
Hi @bjarnef, If I recall correctly the problem of uploading files within Bl or NC, is that the input field might be gone when submitting the document. I think the HTML-input holding the file still needs to be represented when submitting.

There might be other details to why its not working of today, which i cant remember right now.

It would though be possible if we handled files differently. Example uploaded the file instantly and not just when the document is begin submitted. But that would require another infrastructure for file upload.

[...]

PR #10115, opened 2021-04-08
Block list editor upload field

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10103

### Description
Attempt to support Upload property editor in block list. Currently it fails save because content key and property type key are required here:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/FileUploadPropertyValueEditor.cs#L57-L60

Looking the the `ControllerBase` these are set here:
https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web/Editors/ContentControllerBase.cs#L88-L89

With these changes I can save the property with Upload property editor and it also render some values in frontend.

```
<section class="section section">
    @{
        var items = Model.Files.Select(x => x.Content).OfType<BlockFile>();
        foreach (var item in items)
        {
            <h4>@item.Title</h4>
            if (item.File.HasValue())
            {
                <a href="@item.File">@Path.GetFileName(item.File)</a>
            }
        }
    }
</section>
```

However it seems something is not right when opening the editor again. Maybe something in the `FromEditor()` or `ToEditor()` method:

image

According to the docs it should also be possible to use the `HasValue()` extention method, but in this case it throw an ysod.
https://our.umbraco.com/documentation/getting-started/backoffice/Property-Editors/Built-in-Property-Editors/File-Upload/#with-modelsbuilder

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
@@ -229,7 +229,7 @@
          */
         var notSupportedProperties = [
             "Umbraco.Tags",
-            "Umbraco.UploadField",
+            //"Umbraco.UploadField",
             "Umbraco.ImageCropper",
             "Umbraco.NestedContent"
         ];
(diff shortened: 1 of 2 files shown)


bjarnef:
image

Currently it does seem it load the last media in Upload property, when opening the block.
Furthermore seems the part of the path is missing and not sure about the `editorValue.Files` inside `BlockEditorPropertyValueEditor`.

PR #10117, opened 2021-04-09
Members working on the front-end with macro updates

Shazwazza:
This does quite a lot:

* Fixes antiforgery behavior, this became problematic once having both members and users working together. The updates simplify this quite a bit
* Fixes TempData - we were creating new TempDataDictionaries which is never how it should be done because then nothing will get cleared, instead we need to use TempDataDictionaryFactory
* Removes a ton of old code
* Removes member code that was not needed
* Implements front-end member authentication - all of this is done using the default AddIdentity .net implementation, we are only tweaking that slightly
* Updates all member macro snippets which now use bootstrap 5 markup
* Updates all member macro controllers

## Testing

* Make sure you can login and out of the back office without issue
* Add a custom property to the Member type and allow showing/editing on member profile
* Create partial view macro file and use the snippets: (this will create associated macros too)
  * Login
  * Login Status
  * Register
  * Edit Profile
* Edit each macro created and allow rendering in the RTE
* Add each of these macros to an RTE and render that content page
* You should be able to register, login, logout, and edit your profile

Instead of putting them all in an RTE, here's a full HTML page that will work that renders these macros directly with a bootstrap layout. Just be sure your macros are named accordingly.

```html
@using Umbraco.Cms.Web.Common.PublishedModels;
@using Microsoft.AspNetCore.Http.Extensions
@using Umbraco.Cms.Core;
@using Umbraco.Cms.Core.Routing;
@using Umbraco.Cms.Web.UI.NetCore.Controllers;
@inject IPublishedUrlProvider urlProvider;
@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage
@{
    Layout = null;
    var root = Model.Root();
    var isLoggedIn = Context.User?.Identity?.IsAuthenticated ?? false;
}

<html>
<head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">

    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.0-beta3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-eOJMYsd53ii+scO/bJGFsiCZc+5NDVN2yr8+0RDqr0Ql0h+rP48ckxlpbzKgwra6" crossorigin="anonymous">
    
</head>
<body class="d-flex flex-column h-100">
    <main class="flex-shrink-0">
        <div class="container">
            <header class="d-flex flex-wrap justify-content-center py-3 mb-4 border-bottom">
                <a href="@root.Url(urlProvider)" class="d-flex align-items-center mb-3 mb-md-0 me-md-auto text-dark text-decoration-none">
                    <span class="fs-4">@root.Name</span>
                </a>

                <ul class="nav nav-pills">
                    @foreach (var child in root.Children)
                    {
                        var childUrl = @child.Url(urlProvider);
                        var isActive = childUrl.InvariantEquals(Context.Request.Path);
                        var navClass = $"nav-link {(isActive ? "active" : "")}";

                        <li class="nav-item"><a href="@childUrl" class="@navClass">@child.Name</a></li>
                    }
                </ul>
            </header>
        </div>

        <div class="container">
            <div class="row">
                @if (!isLoggedIn)
                {
                    <div class="col-5">
                        @Umbraco.RenderMacro("Login")
                    </div>
                    <div class="col-2 text-center fs-3 text-muted">
                        Or
                    </div>
                    <div class="col-5">
                        @Umbraco.RenderMacro("RegisterMember")
                    </div>
                }
                else
                {
                    <div class="col-12">
                        @Umbraco.RenderMacro("LoginStatus")
                    </div>
                    <div class="col-12">
                        @Umbraco.RenderMacro("EditProfile")
                    </div>
                    
                }
            </div>

        </div>

    </main>

    <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.0.0-beta3/dist/js/bootstrap.bundle.min.js" integrity="sha384-JEW9xMcG8R+pH31jmWH6WWP0WintQrMb4s7ZOdauHnUtxwoG2vI5DkLtS3qm9Ekf" crossorigin="anonymous"></script>

</body>
</html>

```


src/Umbraco.Web.Common/Security/MemberManager.cs
```diff
@@ -28,5 +28,8 @@ public class MemberManager : UmbracoUserManager<MemberIdentityUser, MemberPasswo
                 services, logger, passwordConfiguration)
         {
         }
+
+        public bool IsMemberAuthorized(IEnumerable<string> allowTypes = null, IEnumerable<string> allowGroups = null, IEnumerable<int> allowMembers = null)
+            => true; // TODO: Implement!
```
bergmania, line 33: What is stopping us from implementing this now?

src/Umbraco.Web.Website/Controllers/UmbRegisterController.cs
```diff
@@ -36,59 +51,109 @@ public async Task<IActionResult> HandleRegisterMember([Bind(Prefix = "registerMo
...
             }
 
+            MergeRouteValuesToModel(model);
+
             // U4-10762 Server error with "Register Member" snippet (Cannot save member with empty name)
             // If name field is empty, add the email address instead.
             if (string.IsNullOrEmpty(model.Name) && string.IsNullOrEmpty(model.Email) == false)
             {
                 model.Name = model.Email;
             }
```
bergmania, line 61: This could be part of `RegisterMemberAsync` as well, in case the method will be reused in the future

src/Umbraco.Web.Website/Controllers/UmbRegisterController.cs
```diff
@@ -36,59 +51,109 @@ public async Task<IActionResult> HandleRegisterMember([Bind(Prefix = "registerMo
...
+                if (model.MemberProperties != null)
+                {
+                    foreach (MemberPropertyModel property in model.MemberProperties.Where(p => p.Value != null)
+                        .Where(property => member.Properties.Contains(property.Alias)))
                     {
-                        return Redirect(model.RedirectUrl);
+                        member.Properties[property.Alias].SetValue(property.Value);
                     }
+                }
+                _memberService.Save(member);
```
bergmania, line 147: Explict scope?

bergmania:
My test didn't find anything unexpected.. 💪

src/Umbraco.Web.Common/Security/MemberManager.cs
Shazwazza: Doing this as part of public access

Shazwazza:
@bergmania Have pushed updates for your comments, should be good to merge now 👍

Issue #9976, opened 2021-03-13
'Empty guid' exception when copying multiple nested Nested Content items


creativesuspects:
When copying multiple Nested Content items to another page using the 'Copy Items' feature the copied nested IPublishedElement items end up with empty guid's causing a 'System.ArgumentException: Empty guid.' exception on the frontend. This also results in some quirks in te backoffice: when opening one nested Nested Content item all the nested Nested Content items are expanded, this is probably due to the fact that all of them have an empty guid.

## Umbraco version

I am seeing this issue on Umbraco version: 8.12.0

Reproduction
------------

### Specifics

I'm using the following nested Nested Content setup:

Document type X has a Nested Content property that allows the following nested Nested Content structure (element types):

- nestedParent1
  - nestedChild1
  - nestedChild1
  - nestedChild1
  - [...]
- nestedParent2
  - nestedChild2
  - nestedChild2
  - nestedChild2
  - [...]

I've reproduced the issue on a clean install of Umbraco 8.12.0 with the Starter Kit which can be downloaded here:

https://www.dropbox.com/s/n0zg18o77vqqxif/UmbracoCms.8.12.0.zip?dl=0

Backoffice credentials:

username: test@creativesuspects.com
password: Test12345!

### Steps to reproduce

The issue occurs in Chrome and Firefox. I haven't tested it in other browsers. It only happens when using the 'Copy Items' feature to copy multiple Nested Content items at once and you immediately 'Save and publish' the page. It looks like the guids of the nested IPublishedElement items aren't properly being reset on paste. Strangely enough when you expand all the copied items before saving and publishing then the issue doesn't occur. I've demonstrated all of this here:

https://user-images.githubusercontent.com/161342/111034288-2f704a80-8415-11eb-8534-a26bde313b1f.mp4

### Expected result

The nested IPublishedElement guid's should not be empty after copy/pasting multiple Nested Content items at once.

### Actual result

Nested IPublishedElement guid's are left empty causing an exception on the frontend and quirks in the backoffice as described above.

---
_This item has been added to our backlog [AB#10938](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/10938)_

nul800sebastiaan:
I can confirm this works in 8.6 but broke in 8.7 presumably due to changes introduced for Block List Editor, we'll have to have a look.

patelarpi:
Hello,
[...]

PR #10121, opened 2021-04-12
use paste-resolvers

nielslyngsoe:
fixes https://github.com/umbraco/Umbraco-CMS/issues/9976

The issues described should have been resolved by C#, but since ClipboardService has evolved since this feature was brought to NC, we now have options to deliver new GUIDs on paste. So I have chosen to change the behaviour to such as this is the way we tackle the other property editors.


madsrasmussen:
The PR fixes the described issue. I have tested a couple of levels deep and everything works as expected.

Screenshot 2021-04-12 at 14 32 43

🥳 🎉

Issue #10109, opened 2021-04-07
KeepAlive Ping could be used for external health probes


Jeavon:
The ping method of KeepAliveController has OnlyLocalRequests attribute which checks the request for IsLocal, this in effect means only requests from 127.0.0.1 will get a valid response. (is this used much/ever?)

See -
https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web/Editors/KeepAliveController.cs#L11

I would like to either see this local attribute removed or at least add an option to allow non-local requests because this "ping" could be used by external health probes (Azure Web App Health Check Probe, AZFD Probe etc). The ping method should also accept HTTP HEAD

I've read the fantastic PR #7116 where this attribute was removed and then re-added but I was unable to fully understand why.

Any thoughts @Swimburger @kjac @nul800sebastiaan 

I'll happily create a PR if it's worthwhile (just feel I'm missing something in my understanding)....

Swimburger:
I share your opinion, but it was not critical for me personally.
I think security was the reason behind this, tho I don't really see a security issue with removing the attribute.

Or making it configurable as you suggest, maybe you can specify an IP range or disable it.

nul800sebastiaan:
I have no objection to just removing the attribute again, all good with me! 👍

umbrabot:
Hi @Jeavon,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

Jeavon:
Fixed in #10126

PR #10126, opened 2021-04-12
Allow KeepAlive controller Ping method to be requested by non local requests

Jeavon:
As per #10109 I have swapped the OnlyLocalRequests attribute for HttpHead so that the Ping method can be used by external health probes (Azure Web App Health Check Probe, AZFD Probe etc)

### DIFF
--- a/src/Umbraco.Web/Editors/KeepAliveController.cs
+++ b/src/Umbraco.Web/Editors/KeepAliveController.cs
@@ -1,8 +1,6 @@
 ﻿using System.Runtime.Serialization;
 using System.Web.Http;
-using Umbraco.Web.Mvc;
 using Umbraco.Web.WebApi;
-using Umbraco.Web.WebApi.Filters;
 
 namespace Umbraco.Web.Editors
 {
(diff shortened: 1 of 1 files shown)


mikecp (approved):

mikecp:
Hello @Jeavon ,
Thanks for your PR, it can indeed be useful to access that Ping from outside 👍 

I took the liberty of removing two unused references from that class as well, I hope you won't mind.

Shazwazza:
From what I can tell KeepAlive is 'broken' in a way. This tries to create a URL to ping: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Scheduling/KeepAlive.cs#L53

which will result in `/api/keepalive/ping` will resolve to the auto-routed KeepAliveController, but if you change the `KeepAlivePingUrl` configuration setting, it doesn't change the URL for the controller at all. 

I'm going to fix this all up in netcore and use middleware, but just thought I'd mention it here

Shazwazza:
Hrm, but now that I look at it, it seems you cannot change this config anyways 🤔 (why is it config then! :P )

mikecp:
Good question indeed... Maybe worth cleaning the config settings then?

PR #10140, opened 2021-04-16
Security stamp implementation for members

Shazwazza:
**https://github.com/umbraco/Umbraco-CMS/pull/10138 MUST BE MERGED FIRST**

* [x] We need to persist and handle the Security Stamp and Email Confirmed values since these are not persisted
  * [x] This will require additional DB columns, migrations, service and repo changes
  * [x] We then need to use the security stamp correctly which requires using the security stamp validator in the members cookies implementation

## Testing

Testing should confirm the following:

* This will add a couple columns to your member table
* If you have existing members, as soon as you save or login one of them, the security stamp will be populated
* If you create a member the security stamp will be populated
* If you change an existing members email, password or username, the security stamp will be changed


bergmania (approved):

bergmania:
Seems to work as expected 💪

PR #10141, opened 2021-04-16
Netcore: Migrate RepositoryBase and ContentTypeServiceBase events

nikolajlauridsen:
## Description
This PR migrates the events on the RepositoryBase and ContentTypeServiceBase events.

There were some hiccups along the way which has been resolved.

### RepositoryBase events

We can now use the "ing" events instead of specific "scoped" events, the one major hurdle here was that when deleting a tree of Content and Media only entities in the "DeletedEntities" was the root entity, in order to get around that we now get the descendants in PublishedSnapshotServiceEventHandler.

There was a similar problem with recycle bin emptying, but additionally, the notification didn't include the entities that had been deleted, it does now.

There also was one event that can't be published from the service level yet, that being OnUowRefreshedEntity, I've added a comment on why this is, but the gist of it is that NuCache requires the state to be "Publishing" which is changed in the repository Save method, which means the notification has to be published from that method.

Since the "RefreshedEntity" notification has to be published while in scope, I've injected the IEventAggregator into the RepositoryBase, we could however probably create a new method on "ScopedNotificationPublisher", maybe something like "PublishScopedNotification", and use the ambient scope to publish the notification instead if that is preferred, however, it's only really used two places.

### ContentTypeServiceBase Events

The ContentTypeServiceBase events were kind of tricky because the Save/Delete/etc.. methods all reside in the `ContentTypeServiceBase<TRepository, TItem, TService>` base service, so we need a way to create the appropriate notification classes. The way we do this now is by the ServiceBase having a series of abstract methods for getting notifications which is then overridden by the specific services I.E. `ContentTypeService`

You may also notice that `ContentTypeRefreshNotification` and `ContentTypeChangeNotification` are pretty much identical, to the point of `ContentTypeRefreshNotification` just inheriting from `ContentTypeChangeNotification`. This is necessary because the RefreshNotification must be published within the scope for NuCache, so I opted to create a separate one for this, to allow you to create a handler for in scope, or not in scope.

I think the class `ContentTypeServiceBase<TItem, TService>` can be deleted as well, it seems like the purpose of it was to house all our static events, which are no longer there, however, there is one abstract property `This` still left there, it seems unused though, I don't know if there's an obscure framework reason for this, or if there are any plans for this, so I've left it there, but it's something worth considering. 

### DistributedCacheBinder
I've removed the `DistributedCacheBinderTests` because there's not really any events left to test. Taking a look at it, it seems like the `DistributedCacheBinder` can be simplified once these last event prs are merged, since methods like `FindHandler` and `HandleEvents` are not really relevant with the new notification pattern.

## Testing
Again the testing is very much like the other events PRs, additionally, I've also set a breakpoint in the handlers in `DistributedCacheBinder` and Nucache, to ensure that these handlers are indeed being called.


bergmania:
> I think the class ContentTypeServiceBase<TItem, TService> can be deleted as well, it seems like the purpose of it was to house all our static events, which are no longer there, however, there is one abstract property This still left there, it seems unused though, I don't know if there's an obscure framework reason for this, or if there are any plans for this, so I've left it there, but it's something worth considering.

Sounds like we should just get rid of it

Shazwazza:
Re 

> the one major hurdle here was that when deleting a tree of Content and Media only entities in the "DeletedEntities" was the root entity, in order to get around that we now get the descendants in PublishedSnapshotServiceEventHandler.

I haven't looked into the code but not sure about this. We should never look up all descendants of an item to populate the args, that will be tremendous overhead. I'll look into this more on Monday to try to see what you mean.

nikolajlauridsen:
I went back to publishing the notifications for Media/Content from the repository again, this way we don't have to get the descendants again. 

I marked all the notifications that should only be used for nucache, we want to change these in the future, as obsolete and `EditorBrowsableState.Never`.

I opted to go with a single notification `ScopedEntityRemoveNotification` for both content and media, since nucache doesn't care, since it only uses IContentBase, and this notification shouldn't be used outside nucache.

For members we still  just use `MemberDeletingNotification` because it doesn't suffer from the same issue as content/media (members has not descendants)

PR #10142, opened 2021-04-16
Netcore: Migrate DatabaseSchemaCreation events to notifications

nikolajlauridsen:
### Description
Migrates the two events in `DatabaseSchemaCreator`

This was a bit of an odd one, the two events are essentially just empty cancelable events, and seemingly not used, but seeing how the methods that dispatched the events are virtual I assume it's used somewhere else for some integration purposes? 

I also noticed that the events were internal, so I've just made two new, very very basic, empty notifications, with one of them being cancelable.

I opted to inject IEventAggretor and use that to publish the events, partially because from the comments it seems like the notifications are expected to be published instantly, but also because when the method publishing the events are called from unattended install, it's done so out of a scope.

### Testing
I tested that Umbraco still installs without issue. 

I also tried testing unattended install, however unattended install has been broken by another events migration. I didn't spend a long time looking into this, but it seems like the issue occurs with `AuditNotificationsHandler`, during an unattended install we create a new user, causing the `Handle(UserSavedNotification notification)` method to be invoked, there we try to get the performing user, but `BackOfficeSecurity` is null at this point (probably because the site hasn't fully booted yet)

### DIFF
--- a/src/Umbraco.Core/Events/CancelableNotification.cs
+++ b/src/Umbraco.Core/Events/CancelableNotification.cs
@@ -1,0 +1,11 @@
+namespace Umbraco.Cms.Core.Events
+{
+    public class CancelableNotification : ICancelableNotification
+    {
+        public bool Cancel { get; set; }
+        public void CancelOperation()
+        {
+            Cancel = true;
+        }
+    }
+}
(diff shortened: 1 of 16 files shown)


bergmania (commented):
I have a feeling that DatabaseSchemaCreatingNotification and DatabaseSchemaCreatedNotification should have a shared state like all other Creating..Created notification pairs

nikolajlauridsen:
> I have a feeling that DatabaseSchemaCreatingNotification and DatabaseSchemaCreatedNotification should have a shared state like all other Creating..Created notification pairs

Good point, I've made them stateful, what do you think about EventMessages? Is it necessary?

bergmania (approved):

PR #10150, opened 2021-04-20
Ensure that any ReaderWriterLockSlim is disposed, or replaced with a more suitable lock

Shazwazza:
ReaderWriterLockSlim  is IDisposable and we unfortunately use this a lot and we never dispose it. Ouch. We certainly don't need these locks in most places that we currently use them, like PropertyGroupCollection/PropertyTypeCollection . In theory anything that uses ReaderWriterLockSlim  needs to be IDisposable or dispose it when it's done. I had a look at the source https://referencesource.microsoft.com/#system.core/system/threading/ReaderWriterLockSlim/ReaderWriterLockSlim.cs,52286bb5e7c0fbd4 and the dispose ensures that EventWaitHandles are disposed, else they are not, which could lead to potential handle leaks.

## Work done

* UpgradeableReadLock, WriteLock and ReadLock are obsolete. These are utilities that just wrap ReaderWriterLockSlim but shouldn't be used and should be deleted in v9.
* PropertyGroupCollection/PropertyTypeCollection don't need locks. I have no idea why locks were added but this was done in the initial commit 9 years ago. I can only assume that's because inadvertently these models were going into cache without any deep cloning which would mean the same instance would be shared by many threads. This is not the case today, these are just models that are not intended to be thread safe.
* AppCaches is IDisposable and the Container will dispose of it at the end of the app. This will now try to dispose all of it's caches at the end of the app (if they are also IDisposable)
* DeepCloneAppCache is IDisposable and will be disposed of by the AppCaches
* ObjectCacheAppCache is IDisposable and will be disposed of by the AppCaches and wrapping DeepCloneAppCache
* AppPolicedCacheDictionary is IDisposable and will be disposed of by AppCaches and wrapping DeepCloneAppCache
* WebCachingAppCache is IDisposable and will be disposed of by AppCaches and wrapping DeepCloneAppCache. Removes usage of UpgradeableReadLock
* Refactors ConcurrentHashSet which was originally copied from a SO thread. This can be hugely simplified by just wrapping ConcurrentDictionary and only using it's Keys (with an empty byte as a value). Then it takes care of all locks and doesn't require a ReaderWriterLockSlim, nor the IDisposable requirement
* WebProfilerProvider converts to a normal lock, it doesn't need Read/Write locking
* MemberCache and it's underlying PublishedContentTypeCache is Disposable and will be disposed of by the PublishedSnapshot
* SiteDomainHelper is IDisposable and will dispose of it's ReaderWriterLockSlim at the end of the app because it is registered in DI
* PureLiveModelFactory is IDisposable and will dispose of it's ReaderWriterLockSlim at the end of the app

## Testing

Everything should work, tests should pass.

If you wish to ensure that the IDisposable implementations work, you can put break points into their Dispose methods and while debugging, you can bump the ClientDependency.config file and start clicking around the site to force a restart, the appdomain shutdown logic will all kick in without about 10 seconds and the Dispose methods will execute. I've tested this but feel free to do so as well.

---
_This item has been added to our backlog [AB#12562](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/12562)_


nikolajlauridsen:
Code changes look good and everything still works 👍 

The CopyBranch test is failing however like in #10171, once that is fixed I'd say this is good to merge

Shazwazza:
I merged contrib into this branch and tests seem to pass now

nikolajlauridsen (approved):
Ah yeah, makes sense why the same tests fails both here and in the other PR then, but I'll go ahead and merge, everything looks good 👍

PR #10154, opened 2021-04-20
Made two further classes public to provide access for Forms and other packages.

AndyButland:
As discussed - not sure why we didn't see these before, but they are marked internal currently not accessible from forms.


bergmania:
What is needed in `ModelStateExtensions`? Just asking, because most methods are still internal.

AndyButland:
Got catch, but it's just `ToErrorDictionary` - which is public.

bergmania (approved):

PR #10158, opened 2021-04-21
Fix copy button being inactive when selecting a valid item

nzdev:
### Prerequisites

### Description
Fixes the copy button being inactive when trying to copy content from a document.


nathanwoulfe (approved):
Thanks @nzdev, looking forward to a typed backoffice preventing these sorts of bugs...

PR #10159, opened 2021-04-21
Published members cleanup

Shazwazza:
* removes old membership provider code 🎉 
* IPublishedMemberCache is simplified. It should really have never implemented all that it does especially not IXPathNavigable. All it does now is convert a IMember to IPublishedContent
* New IMemberManager.AsPublishedMember method for working with members on the front-end when you need a member as IPublishedContent. The only real reason this is needed is when working with custom properties and property value converters. Else, all member access on the front-end is just done with IMemberManager and the `MemberIdentityUser` (aspnet identity implementation)
* Removes Groups from MemberIdentityUser since that was a copy/paste mistake during migration
* Moves old PublishedMemberCache that was in core but shouldn't be, to tests since that is the only place its used


src/Umbraco.Infrastructure/Examine/ExamineExtensions.cs
```diff
@@ -75,8 +75,7 @@ public static IEnumerable<PublishedSearchResult> ToPublishedSearchResults(this I
                             content = snapshot.Media.GetById(contentId);
                             break;
                         case IndexTypes.Member:
-                            content = snapshot.Members.GetById(contentId);
-                            break;
+                            throw new NotSupportedException("Cannot convert search results to member instances");
```
bergmania, line 78: I guess we need to support this use case still? It make sense to be allowed for search for members on the website?

Shazwazza: It's not about searching for members, you can always search for members. This is about converting search results into IPublishedContent instances. I forgot this existed in v8 and is a horrible O(N^2) type of mess. If you do this it will be horrible for performance. Listing a ton of member data can be done just by using the search results but converting each search result one at a time to IPublishedContent requires resolving IMember one at at time too. There can be unlimited members, this really shouldn't be supported.

Shazwazza:
This comment is referring the old `ToPublishedSearchResults` for members which used to exist here: https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web/ExamineExtensions.cs#L80 but like the comment says this could be bad for performance because there is no actual member cache, instead in v8, there was a snapshot.Members which would just wrap the IMemberService and look a member up one at a time based on each search result and then convert to IPublishedContent. So the DB queries are N+1 and each DB member lookup is expensive.

Your approach is the same because in v8, the .Content property needed to resolve an IPublishedContent and for members this would lookup each member one at a time (seen here https://github.com/umbraco/Umbraco-CMS/blob/db5a0bad7c15779b0ce9cb526097c7c2b8c65009/src/Umbraco.Web/ExamineExtensions.cs#L38)

So while you think it's performing OK, it will not be if your search results have a ton of members and this will be very heavy on your DB especially if there is high concurrent traffic hitting the API where you are searching with a lot of results.

Your approach is just searching and converting 

In this PR description, it describes how to work with members on the front-end which is typically with IMemberManager. If you need IPublishedContent of members (as per description), use the `IMemberManager.AsPublishedMember`. The main question then is how do you search for members and then get them converted to `MemberIdentityUser` so they can be used with `IMemberManager.AsPublishedMember`? You can do your Examine search (you can also limit the search to return only the ID field for improved performance), take those Ids and lookup the members in bulk by Id and then convert to IPublishedContent.

That said, I don't remember at all if there is an API available on IMemberManager to lookup members in bulk. Based on what is listed here https://github.com/umbraco/Umbraco-CMS/blob/v9/dev/src/Umbraco.Infrastructure/Security/IMemberManager.cs#L17 it doesn't look like it but maybe @bergmania has a better memory than me. I haven't worked on this codebase for a year now.

deekoulos:
Thanks for your response. Well I thought this is the right thread because my original posted issue https://github.com/umbraco/Umbraco-CMS/issues/13001 was closed and marked as a duplicate because of this issue: https://github.com/umbraco/Umbraco-CMS/issues/12891 and here i this comment: https://github.com/umbraco/Umbraco-CMS/issues/12891#issuecomment-1239225991 saying that this is a bad idea. 

So basically I just don't know if this issue will be fixed or not since it is commented as a "bad idea".

Issue #10144, opened 2021-04-19
DataType infinite editing, prevalue alias not set


leekelleher:
Hiya Team Umbraco!

I'd noticed in the infinite editing for the DataType, e.g. when adding a new property to a ContentType, that the alias for the prevalue isn't correct, (it's auto-generated).

![doctype-datatype-infinityediting-wrong-alias](https://user-images.githubusercontent.com/209066/115256369-100cc180-a127-11eb-8a45-f1304552923f.png)

> _Screengrab above shows the auto-generated alias - viewable on hover state_

The reason I spotted this is that with an editor my Umbraco package (Contentment Data List), there's a preview field that checks the values of the other fields, but given the aliases are randomised, my previewer feature doesn't work with the DataType infinite editor. 😞 

I believe the reason for this is that the API call to get the DataType's Prevalue fields has a `key` property, but it doesn't have an 'alias' field.  From exploring the source code, I'd spotted that the regular DataType editor manually sets the `alias` field in the AngularJS controller, here: <https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/datatypes/datatype.edit.controller.js#L44>

I've tried to dig around to see where to fix the underlying issue, but to be honest, I got a little bit lost. 😕 

I did find where the random alias is generated: <https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/directives/components/property/umbpropertyeditor.directive.js#L33>
I wonder whether a simpler fix is to check at that point whether the field has a `key` property before generated the random string.  But could do with guidance on the best approach there. 😃 

## Umbraco version

I am seeing this issue on Umbraco version: 8.13.0-rc. _(But it's probably in earlier versions, I hadn't noticed previously)_

Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

### Steps to reproduce

When adding a property to a ContentType, go into the DataType's configuration editor (e.g. the infinite editing mode). Hover your mouse cursor over a property's name and the title attribute should display a randomly generated alias ID.

### Expected result

Expectation is that the alias is the intended alias for the property-editor's configuration field. e.g. not randomly generated.

### Actual result

_(Ignoring section, feel that I'd already explained this further above. ☮️ )_

patrickdemooij9:
[...]

PR #10163, opened 2021-04-21
Fixed 10144: Prevalue alias not set

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10144

Can be tested by:
When adding a property to a ContentType, go into the DataType's configuration editor (e.g. the infinite editing mode). Hover your mouse cursor over a property's name and the title attribute should display the correct alias instead of a randomly generated ID as it was doing before this.


src/Umbraco.Web.UI.Client/src/views/datatypes/datatype.edit.controller.js
```diff
@@ -71,6 +60,7 @@ function DataTypeEditController($scope, $routeParams, appState, navigationServic
                 vm.preValuesLoaded = true;
                 vm.content = data;
 
+                console.log(vm.content.preValues);
```
leekelleher, line 63: 👀 @patrickdemooij9 Spotted a `console.log` left in.

patrickdemooij9: Whoops! I've removed it

nathanwoulfe:
Before: 
image
After:
image

Yeah, I couldn't get the screen grab with the tooltip showing. All works fine though

nathanwoulfe:
I've merged latest v8 into the PR branch, hopefully that resolves the failing test (I don't think the test was anything related to this change, but good to get all green lights before I merge).

nathanwoulfe (approved):

nul800sebastiaan:
FYI @patrickdemooij9 @nathanwoulfe 

I had to update this one a little bit, turns out just editing datatypes from the datatypes tree was working fine but when you're editing the document type and opening the datatype in infinite editing it would no longer send the `key` to the server, totally preventing us from saving the datatype because we couldn't figure out what prevalues were being sent in.

When you edit a datatype normally we see this in the collection:

image

But from infinite editing we see this, it's sending `alias` instead of `key`:

image

So I've expected either one of them now in this commit: dc27a31a6fe35a885eab9079d2e002fc53da6391 which seems to work in both scenarios now.

PR #10164, opened 2021-04-21
Add "Insert Macro" as language key

patrickdemooij9:
This was something that Alexander on the slack channel figured out. The "Click to insert macro" is not a translatable key in the CMS. I've changed the code so that it is part of the translatable keys.

Can be tested by changing the key "clickToInsertMacro" and then see it change in the grid editor.

(I am not quite sure what happened to the formatting)


mikecp (approved):

mikecp:
Thanks @patrickdemooij9 for this new localization key 👍

PR #10170, opened 2021-04-22
Implement password config storage for members

Shazwazza:
For users, we've stored their current password configuration in the DB. This allows us to be able to migrate from > to any new hashing algorithm. This is now implemented for members too. 

* Shares code between members/users for password checking with password config
* Fixes the default cookie options for members to now use a custom cookie manager to ensure that we are not auth'ing any requests for the back office
* Fixes startup issue where public access stopped working due to our fluent Startup syntax change. This is now updated to be two methods with delegates: WithMiddlware + WithEndpoints
* Fixes default password hashing type for members

## Testing

This will force an upgrade and it should work. Make sure that members work, you can login and public access works. Then check your members database after logging in or updating a member - they will now have a value for passwordConfig in the db.


src/Umbraco.Infrastructure/Migrations/Upgrade/V_9_0_0/MemberTableColumns2.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Linq;
+using Umbraco.Cms.Infrastructure.Persistence.Dtos;
+
+namespace Umbraco.Cms.Infrastructure.Migrations.Upgrade.V_9_0_0
+{
+    public class MemberTableColumns2 : MigrationBase
```
bergmania, line 6:
```suggestion
    public class AddPasswordConfigToMemberTable : MigrationBase
```

src/Umbraco.Infrastructure/Persistence/Dtos/MemberDto.cs
```diff
@@ -32,6 +32,14 @@ internal class MemberDto
         [Constraint(Default = "''")]
         public string Password { get; set; }
 
+        /// <summary>
+        /// This will represent a JSON structure of how the password has been created (i.e hash algorithm, iterations)
+        /// </summary>
+        [Column("passwordConfig")]
+        [NullSetting(NullSetting = NullSettings.Null)]
+        [Length(500)]
```
bergmania, line 40: This sounds like a very long possible value?

bergmania (commented):
Ill fix the name of the migration, else it looks good.. Off to test

bergmania (approved):

bergmania:
Seems to work, nice work 💪 

Can be merged if/when build server is happy

PR #10177, opened 2021-04-22
Minor Danish translations

abjerner:
The shortcut overview for templates has some items that are not translated to Danish. The Danish language file already contains the items, but their values are in English - so now I've fixed them:

image


nathanwoulfe (approved):
Thanks @abjerner 👍

Issue #10179, opened 2021-04-23
Disabling a tree controller using UmbracoFeatures does not remove the tree from backoffice


rasmusjp:
In Heartcore we use `UmbracoFeatures` to disable different controllers we don't support. We do that both for editor controllers and tree controllers.

In Umbraco 8, when disabling a tree controller, the tree was also removed from the backoffice, this does not happend in Umbraco 9, here it's still visible but interacting with it gives a 403 (so the controller is disabled).

## Umbraco version

I am seeing this issue on Umbraco version: 9.0.0-alpha0004

Reproduction
------------

### Steps to reproduce

Add the following code to the project, compile and run the application, then navigate to the Settings section and see the Templates tree is visible. If you add the some code to an Umbraco 8 project, the tree is not visible

```
public class BackofficeComposer : ComponentComposer<BackofficeComponent>
{
}

public class BackofficeComponent : IComponent
{
    private UmbracoFeatures _umbracoFeatures;

    public BackofficeComponent(UmbracoFeatures umbracoFeatures)
    {
        _umbracoFeatures = umbracoFeatures;
    }
    public void Initialize()
    {
        _umbracoFeatures.Disabled.Controllers.Add<TemplatesTreeController>();
    }

    public void Terminate()
    {
    }
}
```

### Expected result

The tree is not shown

### Actual result

The tree is shown

PR #10180, opened 2021-04-23
Fix check authorization policies on nested tree requests

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/10179

Not sure this is the best way to resolved it, but it is the only solution I was able to find.

This code is inspired by what the [AuthorizationMiddleware](https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs) are doing.

It relies on the fact that our `FeatureAuthorizeHandler` are still supporting pre-.NET 5 behaviour, where the resource can be an Endpoint. 🙈


warrenbuckley:
Code reads fine
Tested before and after PR applied and it works as described :)

PR #10181, opened 2021-04-23
Netcore: File systems rework

nikolajlauridsen:
Reworks the filesystems to allow the underlying filesystem for Media to be replaced, and prevents you from replacing the FileSystems.

This pr:
* Removes IMediaFileSystem
* Removes MediaFileSystem
* Removes FileSystemWrapper
* Removes IFileSystems
* Simplifies FileSystems
* Adds MediaFileManager

### MediaFileManager

MediaFileManager replaces what was before MediaFileSystem, the additional methods on MediaFileSystem was essentially just helper methods using the underlying filesystem, the fact that it was also a FileSystem mudied and complicated things. Now MediaFileManager is just a concrete type that has the extra helper methods on it, and exposes the filesystem used by media through its FileSystem property. Where you would do `MediaFileSystem.AddFile` you now do `MediaFileManager.FileSystem.AddFile`, I chose to do it this way to make the differentiation between the MediaFileManager and the underlying FileSystem obvious. 

The way the `UmbracoBuilder.SetMediaFileSystem` method is used is pretty much as before, but the implementation has changed, now we use FileSystems to wrap the filesystem for the MediaFileSystem before we create it, and then pass the wrapped filesystem to the MediaFileManager. 

MediaFileManager does not have an interface, because we don't want to have this class replaced, since that is a sure fire way to cause trouble, you can only change the underlying filesystem.

### FileSystems

The FileSystems has been cleaned up somewhat, since the new implemtation of MediaFileManager means we no longer need to have the `GetFileSystem` method, which was a confusing method because it actually did three things, get, or create and register if it doesn't exist yet, we don't support that anymore. 

IFileSystems has also been removed, for much the same reason MediaFileManager no longer has an interface, this could somewhat be seen as a problem for unit testing, but a new internal constructor has been added, that allows you set the various filesystems to whatever you want, again making it testable. 

In order for FileSystems to also be testable outside Umbraco there has been added a `FilesystemsCreator.CreateTestFileSystems` static method to the Umbraco.Tests.Common project, this allows for the internal constructor to be used by test project outside Umbraco as long as they have that package. 

## Testing

Test that the built in file system works as intended, you can add media files, script files etc..

A nice test to do would also be to try and replace the underlying filesystem for the MediaFileManager with the `IUmbracoBuilder.SetMediaFileSystem` extension method, however the IFileSystem interface has not changed, so I don't see it being an issue.


rasmusjp:
Haven't looked at the code yet, but I'm wondering if it's still possible to overwrite e.g. the `StylesheetFilesystem` with a custom one? We do this in Heartcore to store RTE Stylesheet files created in the backoffice in Blob Storage (and setting the others to a `VoidFileSystem` that does nothing)

nikolajlauridsen:
> Haven't looked at the code yet, but I'm wondering if it's still possible to overwrite e.g. the `StylesheetFilesystem` with a custom one? We do this in Heartcore to store RTE Stylesheet files created in the backoffice in Blob Storage (and setting the others to a `VoidFileSystem` that does nothing)

The way `FileSystems` works currently that is not possible, when a filesystem is requested the first time all filesystems are created as `PhysicalFileSystem`. 

There is no concrete type `StyleSheetFilesystem` though, the filesystem for stylesheets is just a IFileSystem

rasmusjp:
> > Haven't looked at the code yet, but I'm wondering if it's still possible to overwrite e.g. the `StylesheetFilesystem` with a custom one? We do this in Heartcore to store RTE Stylesheet files created in the backoffice in Blob Storage (and setting the others to a `VoidFileSystem` that does nothing)
> 
> The way `FileSystems` works currently that is not possible, when a filesystem is requested the first time all filesystems are created as `PhysicalFileSystem`.
> 
> There is no concrete type `StyleSheetFilesystem` though, the filesystem for stylesheets is just an `IFileSystem`

Okay, what we do right now is we have a class that implements the `IFileSystems` interface and we replace it in a Composer

```csharp
public class FileSystems : IFileSystems
{
    public FileSystems(IFileSystem macroPartialsFileSystem, IFileSystem partialViewsFileSystem, IFileSystem stylesheetsFileSystem, IFileSystem scriptsFileSystem, IFileSystem mvcViewsFileSystem)
    {
        MacroPartialsFileSystem = macroPartialsFileSystem;
        PartialViewsFileSystem = partialViewsFileSystem;
        StylesheetsFileSystem = stylesheetsFileSystem;
        ScriptsFileSystem = scriptsFileSystem;
        MvcViewsFileSystem = mvcViewsFileSystem;
    }

    public IFileSystem MacroPartialsFileSystem { get; }
    public IFileSystem PartialViewsFileSystem { get; }
    public IFileSystem StylesheetsFileSystem { get; }
    public IFileSystem ScriptsFileSystem { get; }
    public IFileSystem MvcViewsFileSystem { get; }
}
```

and in the composer we have

```csharp
composition.RegisterUnique<IFileSystems>(_ => new FileSystems(
    new VoidFileSystem(),
    new VoidFileSystem(),
    new FileSystem("~/css"), // our azure blob file system
    new VoidFileSystem(),
    new VoidFileSystem()
));
```

Being able to do replace all the filesystems is a must for us, given we don't want to store files on the physical filesystem

Shazwazza:
Thanks @rasmusjp ,

The issue with replacing IFileSystems this way means you are no longer shadowing any file system which means in Deploy there will be no rolled back files if a deployment fails and needs to rollback.

We want to entirely overhaul file systems in v9 but it's too much work for the v9.0 release so we need to postpone the overhaul to v10 (probably) and in the meantime make what we have make a little more sense and at the same time don't let developers replace services that will end up breaking things.

Potentially we'll just need to have methods to set each file system like:

SetMediaFileSystem
SetCssFileSystem

in those methods we ensure these are wrapped with shadow wrappers to ensure that deployments can be rolled back.

I guess we need to understand more about what you mean here

> Being able to do replace all the filesystems is a must for us, given we don't want to store files on the physical filesystem

Shadow file systems use the physical file system to temporary shadow the underlying system so during a deploy, the physical file system will be used and once it's committed and successful, the files are committed to the underlying IFileSystem.

Also given that view features are disabled, is it that important to be able to set the views file systems as void? 

We basically don't want support for any IFileSystems outside of media (and it seems in your case CSS) because for now that's all we want to support people being able to replace. The only reason IFileSystem exists for others is really for this shadowing functionality. 

Maybe we should have a zoom call about all of this?

rasmusjp:
> The issue with replacing IFileSystems this way means you are no longer shadowing any file system which means in Deploy there will be no rolled back files if a deployment fails and needs to rollback.

I get that, the reason for us replacing `IFileSystems` is because it's currently not possible to replace a single filesystem implementation.

> Potentially we'll just need to have methods to set each file system like:
> 
> SetMediaFileSystem
> SetCssFileSystem
> 
> in those methods we ensure these are wrapped with shadow wrappers to ensure that deployments can be rolled back.

That would be great, as we really only need to replace the `StylesheetFileSystem` at the moment.

> 
> I guess we need to understand more about what you mean here
> 
> > Being able to do replace all the filesystems is a must for us, given we don't want to store files on the physical filesystem
> 
> Shadow file systems use the physical file system to temporary shadow the underlying system so during a deploy, the physical file system will be used and once it's committed and successful, the files are committed to the underlying IFileSystem.

Using the physical filesystem for temp files is fine, we just dont't want to accidentally store user generated files on the physical filesystem.

> Also given that view features are disabled, is it that important to be able to set the views file systems as void?

No it's not that important, IIRC the reason we did it in the current implementation was because there were no way to replace the `StylestheetFileSystem` without having a custom implementation of the `IFileSystem`.

> Maybe we should have a zoom call about all of this?

Sure, just message me on Slack and let us find a timeslot that works for both of us :)

nikolajlauridsen:
Okay, after some poking about and trial and error I've made it possible to replace the filesystem used for stylesheets, sadly I had to gleam the cube and hack the planet, but it should work.

FileSystems now take an IFileSystem in its constructor, which is wrapped in a shadow wrapper and set as `_stylesheetsFileSystem`, this is then used in the new IUmbracoBuilder extension method AddFileSystems.

AddFileSystem takes a factory method for creating an IFileSystem, it will then create that filesystem and register FileSystems with that filesystem injected into the constructor with `provider.CreateInstance<FileSystems>(stylesheetFileSystem);`.

It has to be done this way because we need to supply the factory method with an IServiceProvider to resolve depedencies the filesystem needs. In order to get an IServiceProvider you need to register something in the DI container, and if you use a factory method it will only be invoked if and when the registered service is requested. 

I also tried adding a "SetStylesheetFileSystem" to FileSystems and then resolve FileSystems with a service provider gotten with: `            builder.Services.BuildServiceProvider();`, however this will resolve in a different instance of FileSystems than is used in the final container, so this is not possible either. You could with this approach register the stylesheet filesystem in a Component's initalize method, but this is even more hacky, so I opted not to go that route. 


Another problem I found is that changing the root path or root url of the stylesheet filesystem will cause it to break. This is because we have a tight coupling to the virtual path passed by the front end (/css/), we take that path and call GetRelativePath on the underlying filesystem, the physical file system will then take that path and see if that path starts with the `_rootUrl`, and of course if you change the root url, the path won't match anymore, and cause things to break, the same goes for the GetFullPath method.  

To try and mitigate the issue of users reporting bugs after trying to replace the location for storing stylesheets I've added this check in AddFileSystems: 

```c#
try
{
    var rootPath = stylesheetFileSystem.GetFullPath("/css/");
}
catch (UnauthorizedAccessException exception)
{
    throw new InvalidOperationException("Can't register the stylesheet filesystem, " +
                                        "this is most likely caused by using a PhysicalFileSystem with an incorrect " +
                                        "rootPath/rootUrl. RootPath must be <installation folder>\\wwwroot\\css" +
                                        " and rootUrl must be /css", exception);
}
``` 

I don't know if this will cause issues for you in Heartcore @rasmusjp? If so I can remove the check, it's really just to try and give users a bit more friendly error message.

rasmusjp:
Hi @nikolajlauridsen 

While it does work, you're right it feels quite hacky so I made some changes in this commit https://github.com/umbraco/Umbraco-CMS/commit/b160df37b13f653f2e1e3641c011e8663a24cdfd adding the `SetStylesheetFileSystem` method, and removes the overloads that takes the stylesheet filesystem factory method.

It can be used the following way:

```cs
builder.AddFileSystems((factory, fileSystems) =>
{
    IIOHelper ioHelper = factory.GetRequiredService<IIOHelper>();
    IHostingEnvironment hostingEnvironment = factory.GetRequiredService<IHostingEnvironment>();
    ILogger<PhysicalFileSystem> logger = factory.GetRequiredService<ILogger<PhysicalFileSystem>>();
    GlobalSettings settings = factory.GetRequiredService<IOptions<GlobalSettings>>().Value;

    var path = settings.UmbracoCssPath;
    var rootPath = hostingEnvironment.MapPathWebRoot(path);
    var rootUrl = hostingEnvironment.ToAbsolute(path);
    var fileSystem = new WhateverStylesheetFileSystem(ioHelper, hostingEnvironment, logger, rootPath, rootUrl);

    fileSystems.SetStylesheetFilesystem(fileSystem);
});
```

What do you think of this approach?

Given it's an edge case and not something that normally should be done, we can even remove the `AddFileSystems` overload and do the `AddUnique` in Heartcore.

> I don't know if this will cause issues for you in Heartcore @rasmusjp? If so I can remove the check, it's really just to try and give users a bit more friendly error message.

This is fine as we don't change the path anyways

rasmusjp:
I also noticed a couple of other issues with the order of `AddFileSystems` and `SetMediaManager` was called. It's also fixed in the same branch. Let me know if you want me to merge my changes into this branch or you can just do it.

src/Umbraco.Core/IO/FileSystems.cs
```diff
@@ -38,92 +40,137 @@ public class FileSystems : IFileSystems
         #region Constructor
 
         // DI wants a public ctor
-        public FileSystems(IServiceProvider container, ILogger<FileSystems> logger, ILoggerFactory loggerFactory, IIOHelper ioHelper, IOptions<GlobalSettings> globalSettings, IHostingEnvironment hostingEnvironment)
+        public FileSystems(
+            ILoggerFactory loggerFactory,
+            IIOHelper ioHelper,
+            IOptions<GlobalSettings> globalSettings,
+            IHostingEnvironment hostingEnvironment,
+            IFileSystem stylesheetFileSystem)
```
Shazwazza, line 48: If we are going to allow injection of the stylesheetFileSystem, it must be `IStylesheetFileSystem` which inherits from `IFileSystem`. Each one must be a unique service because MSDI doesn't support named services.

There was previous issues when trying to inject the media file system, so just wondering why this isn't a problem here?

src/Umbraco.Core/IO/FileSystems.cs
```diff
@@ -38,92 +40,137 @@ public class FileSystems : IFileSystems
...
         {
-            _container = container;
-            _logger = logger;
+            _logger = loggerFactory.CreateLogger<FileSystems>();
             _loggerFactory = loggerFactory;
             _ioHelper = ioHelper;
             _globalSettings = globalSettings.Value;
             _hostingEnvironment = hostingEnvironment;
+
+            _stylesheetsFileSystem = (ShadowWrapper)CreateShadowWrapper(stylesheetFileSystem, "css");
```
Shazwazza, line 56: Instead of casting this each time wouldn't it be better to have a private method like CreateShadowWrapperInternal that actually just returns ShadowWrapper and CreateShadowWrapper just calls CreateShadowWrapperInternal?

src/Umbraco.Core/IO/FileSystems.cs
```diff
@@ -38,92 +40,137 @@ public class FileSystems : IFileSystems
...
+            _wkfsInitialized = true;
+
         }
 
         // for tests only, totally unsafe
-        internal static void ResetShadowId()
-        {
-            _shadowCurrentId = null;
-        }
+        internal static void ResetShadowId() => _shadowCurrentId = null;
```
Shazwazza, line 81: Do we still need this? Ideally all static things are gone, if it's needed I don't think it needs to be static?

src/Umbraco.Infrastructure/DependencyInjection/UmbracoBuilder.Uniques.cs
```diff
@@ -109,18 +109,53 @@ public static void SetShortStringHelper(this IUmbracoBuilder builder, IShortStri
...
-                return fileSystems.GetFileSystem<MediaFileSystem>(filesystemFactory(factory));
+                IFileSystem stylesheetFileSystem = stylesheetFactory(provider);
+                // Verify that _rootUrl/_rootPath is correct
+                // We have to do this because there's a tight coupling
+                // to the VirtualPath we get with CodeFileDisplay from the frontend.
+                try
+                {
+                    var rootPath = stylesheetFileSystem.GetFullPath("/css/");
+                }
+                catch (UnauthorizedAccessException exception)
```
Shazwazza, line 151: Should be `InvalidOperationException`

src/Umbraco.Infrastructure/DependencyInjection/UmbracoBuilder.Uniques.cs
```diff
@@ -109,18 +109,53 @@ public static void SetShortStringHelper(this IUmbracoBuilder builder, IShortStri
...
-        /// Using this helper will ensure that your IFileSystem implementation is wrapped by the ShadowWrapper
+        /// Be careful when using this, the root path and root url must be correct for this to work.
         /// </remarks>
-        public static void SetMediaFileSystem(this IUmbracoBuilder builder, Func<IServiceProvider, IFileSystem> filesystemFactory)
-            => builder.Services.AddUnique<IMediaFileSystem>(factory =>
+        /// <param name="builder">A builder.</param>
+        /// <param name="stylesheetFactory">Factory method to create an IFileSystem implementation used for stylesheets.</param>
+        /// <exception cref="InvalidOperationException">Throws exception if full path can't be resolved successfully.</exception>
+        public static void AddFileSystems(this IUmbracoBuilder builder,
+            Func<IServiceProvider, IFileSystem> stylesheetFactory) => builder.Services.AddUnique(
```
Shazwazza, line 140: As above, `IFileSystem` is a base interface/implementation, if we are injecting a specific `IFileSystem` it has to be it's own type: `IStylesheetFileSystem`

src/Umbraco.Infrastructure/DependencyInjection/UmbracoBuilder.Uniques.cs
```diff
@@ -109,18 +109,53 @@ public static void SetShortStringHelper(this IUmbracoBuilder builder, IShortStri
...
         /// <remarks>
-        /// Using this helper will ensure that your IFileSystem implementation is wrapped by the ShadowWrapper
+        /// Be careful when using this, the root path and root url must be correct for this to work.
         /// </remarks>
-        public static void SetMediaFileSystem(this IUmbracoBuilder builder, Func<IServiceProvider, IFileSystem> filesystemFactory)
-            => builder.Services.AddUnique<IMediaFileSystem>(factory =>
+        /// <param name="builder">A builder.</param>
+        /// <param name="stylesheetFactory">Factory method to create an IFileSystem implementation used for stylesheets.</param>
+        /// <exception cref="InvalidOperationException">Throws exception if full path can't be resolved successfully.</exception>
+        public static void AddFileSystems(this IUmbracoBuilder builder,
```
Shazwazza, line 139: Any reason this is called AddFileSystems when it's only for stylesheets? IMO the naming should be consistent like the media one: `SetStylesheetFileSystem` ?

Shazwazza (changes requested):
Have left some comments/questions inline.

src/Umbraco.Core/IO/FileSystems.cs
nikolajlauridsen: Me saying that we inject is a bit misleading, I don't actually register the filesystem in the container and inject it, I pass it to the constructor using `provider.CreateInstance<FileSystems>(stylesheetFileSystem);` that way the filesystem never actually gets registered in the container, it just gets passed to the constructor when FileSystems is created. 

The reason that the media file system was an issue was because we needed to shadow the inner media file system, and not the media file system itself, but the stylesheet filesystem is just an arbitrary IFileSystem that's only accessed through FileSystems, so that's why it's not an issue, but I think it's better to go with Rasmus suggestion, that way we don't need to pass anything to the constructor.

src/Umbraco.Core/IO/FileSystems.cs
nikolajlauridsen: Yeah that's a great idea

src/Umbraco.Infrastructure/DependencyInjection/UmbracoBuilder.Uniques.cs
nikolajlauridsen: Not sure I understand this? GetFullPath throws an UnauthorizedAccessException if the path is not set up correctly, but I suppose we should re throw an UnauthorizedAccessException then but with a different message added?

src/Umbraco.Infrastructure/DependencyInjection/UmbracoBuilder.Uniques.cs
nikolajlauridsen: I chose to call it AddFileSystems because that's what the extension method does, but I can see that it's a bit confusing that we use AddFileSystems to change the stylesheet filesystem.

rasmusjp:
Just a FYI @nikolajlauridsen, some of the conserns @Shazwazza has have, been fixed in my branch https://github.com/umbraco/Umbraco-CMS/tree/netcore/feature/stylesheet-filesystem

nikolajlauridsen:
> Just a FYI @nikolajlauridsen, some of the conserns @Shazwazza has have, been fixed in my branch https://github.com/umbraco/Umbraco-CMS/tree/netcore/feature/stylesheet-filesystem

Yeah that's true, just wanted to let him know what my thoughts initially was. 

I've merged your branch and made a few changes, I've removed TrySetMediaFileSystems because the original AddFileSystems (the one in UmbracoBuilder.FileSystems) will always run first, I've also renamed the new AddFileSystems to ConfigureFilesystems, because it gives confusion since there would be two AddFileSystems, one to add all filesystems, and one to register the Filesystems class.

So now ConfigureFilesystems just registers a new FileSystems class that you can configure like you added, but don't call the original AddFilesystems.

src/Umbraco.Core/IO/FileSystems.cs
nikolajlauridsen: I thought it was still required for a test, but I was wrong, I've removed it.

rasmusjp:
> Yeah that's true, just wanted to let him know what my thoughts initially was.

:+1:  just wanted to make sure you had seen my changes before you started looking at fixes

> I've merged your branch and made a few changes, I've removed TrySetMediaFileSystems because the original AddFileSystems (the one in UmbracoBuilder.FileSystems) will always run first, I've also renamed the new AddFileSystems to ConfigureFilesystems, because it gives confusion since there would be two AddFileSystems, one to add all filesystems, and one to register the Filesystems class.
>
> So now ConfigureFilesystems just registers a new FileSystems class that you can configure like you added, but don't call the original AddFilesystems.

Cool, that should fix the issue I had with `AddFileSystems(configure)` overwriting the media filesystem, if `SetMediaFileSystem` was called first :) and I agree, `ConfigureFileSystems` makes more sense in this context

PR #9859, opened 2021-02-21
Reduce cold boot times by loading content and media only once on cold boot

nzdev:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes #9779

### Description
What?

Lets the PublishedSnapshotService know if the servermessenger is going to cold boot so PublishedSnapshotService does not attempt to load all content and media twice.

Why?
Load content and media only once on cold boot saving significant boot time on sites with large amounts of content/media . Saves on memory as otherwise there is 2 generations of every content and media in memory at boot.

How to test?
Delete the NuCache.db and server sync state text file. The log message about loading content from the database should appear only once.

PR #10182, opened 2021-04-26
Fix startup issues and timing with Examine and Nucache

### LINKED ISSUE #8893: Examine rebuilds the indexes on startup before custom fields or transformers are registered
We have extended the External Searcher with some extra fields and have also attached an event to the TransformingIndexValues index event so that we can modify the values that are being indexed.

The problem is that Examine rebuilds the indexes on startup and wipes all the correct values in the index with lots of missing values.  It is basically rebuilding the index before our customisations were able to be applied.

From my own investigation, this is because Examine configures itself at the ICoreComposer stage and then registers the ExamineFinalComponent to run after all ICoreComposer's have finished.

See:
* https://github.com/umbraco/Umbraco-CMS/blob/db6714d3b8f8e04dceb607c3a25117331f6b525d/src/Umbraco.Web/Search/ExamineFinalComposer.cs

According to the documentation [here](https://our.umbraco.com/documentation/reference/searching/examine/indexing/), all examples register components at IUserComposer stage, which means we are too late to the party on startup.

Is this a documentation problem or a code issue?  I suppose more specifically, should Examine rebuild after all IUserComposer's and their components have complete?

I can confirm, if I change my custom composer to a ICoreComposer, then everything starts working again. 

I think this issue is worth recording because it can certainly go unnoticed unless you write a site that's heavily dependant on Examine like I am. 

## Umbraco version

I am seeing this issue on Umbraco version: 8.6.1

Reproduction
------------

Set up a IUserComposer as per documentation that registers some custom examine fields, and also attach to the index TransformingIndexValues event to manipulate some values.

### Bug summary

Examine rebuilds the indexes on startup but completely ignores the custom code because the custom code hasn't run yet.

### Specifics

See above.

### Expected result

Examine would startup and rebuild the indexes, but would wait for me to apply my customisations before going ahead.

### Actual result

Missing stuff.

Shazwazza:
This fixes #8893 and refactors #9859.

* Moves logic for cold boot into the components responsible for what needs to happen in cold boot. This used to be in callbacks in DatabaseServerMessenger.
* Removes the callbacks from DatabaseServerMessenger and obsoletes these callbacks.
* Ensure examine rebuilds don't occur until the first http request is done instead of on a timer which could be problematic with cold boots. A reason for the #8893 can be that during cold boots, the cold boot processing would take more than 5 seconds and then after those 5 seconds indexing occurs and this will happen before custom components execute that customize those indexes.
* PublishedSnapshotService now only initializes lazily when it's needed.
* Removes BatchedDatabaseServerMessenger.Startup, startup logic is just done in the component along with events.

The whole way that startup/init is done in v8 is not ideal but we are stuck with backwards compat.  Once main problems is that nucache (PublishedSnapshotService) does it's init in it's ctor which doesn't leave much flexibility with how the init process works. This will be fixed/changed in netcore (if not already, I know some of it is). 

## Testing

To force a cold boot, just delete the /TEMP/DistCache folder, then restart your app, this will cold boot.

* Ensure all indexes are rebuilt on cold boot
* Ensure indexes are not rebuilt on non-cold boot
* Ensure nucache is rebuilt (once) on a cold boot
* Ensure nucache is not rebuilt on non-cold boot
* Ensure that you can define custom Examine field types (i.e. #8893) on startup with a IUserComposer and those field types work both in a cold boot and non-cold boot

### DIFF
--- a/src/Umbraco.Web/Search/ExamineFinalComponent.cs
+++ b/src/Umbraco.Web/Search/ExamineFinalComponent.cs
@@ -3,6 +3,8 @@ using Umbraco.Core.Logging;
 using Umbraco.Examine;
 using Umbraco.Core.Composing;
 using Umbraco.Core;
+using Umbraco.Core.Sync;
+using Umbraco.Web.Routing;
 
 namespace Umbraco.Web.Search
 {
@@ -16,27 +18,65 @@ namespace Umbraco.Web.Search
         private readonly IExamineManager _examineManager;
         BackgroundIndexRebuilder _indexRebuilder;
         private readonly IMainDom _mainDom;
-        
-        public ExamineFinalComponent(IProfilingLogger logger, IExamineManager examineManager, BackgroundIndexRebuilder indexRebuilder, IMainDom mainDom)
+        private readonly ISyncBootStateAccessor _syncBootStateAccessor;
+        private readonly object _locker = new object();
+        private bool _initialized = false;
+
+        public ExamineFinalComponent(IProfilingLogger logger, IExamineManager examineManager, BackgroundIndexRebuilder indexRebuilder, IMainDom mainDom, ISyncBootStateAccessor syncBootStateAccessor)
         {
             _logger = logger;
             _examineManager = examineManager;
             _indexRebuilder = indexRebuilder;
             _mainDom = mainDom;
+            _syncBootStateAccessor = syncBootStateAccessor;
+
+            // must add the handler in the ctor because it will be too late in Initialize
+            // TODO: All of this boot synchronization for cold boot logic needs should be fixed in netcore
+            _syncBootStateAccessor.Booting += _syncBootStateAccessor_Booting;
+        }
+
+        /// <summary>
+        /// Once the boot state is known we can see if we require rebuilds
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void _syncBootStateAccessor_Booting(object sender, SyncBootState e)
+        {
+            UmbracoModule.RouteAttempt += UmbracoModule_RouteAttempt;
+        }
+
+        private void UmbracoModule_RouteAttempt(object sender, RoutableAttemptEventArgs e)
+        {
+            UmbracoModule.RouteAttempt -= UmbracoModule_RouteAttempt;
+
+            if (!_initialized)
+            {
+                lock (_locker)
+                {
+                    // double check lock, we must only do this once
+                    if (!_initialized)
+                    {
+                        if (!_mainDom.IsMainDom) return;
+
+                        var bootState = _syncBootStateAccessor.GetSyncBootState();
+
+                        _examineManager.ConfigureIndexes(_mainDom, _logger);
+
+                        // if it's a cold boot, rebuild all indexes including non-empty ones
+                        _indexRebuilder.RebuildIndexes(bootState != SyncBootState.ColdBoot, 0);
+                    }
+                }
+            }
+            
         }
 
         public void Initialize()
-        {
-            if (!_mainDom.IsMainDom) return;
-
-            _examineManager.ConfigureIndexes(_mainDom, _logger);
-
-            // TODO: Instead of waiting 5000 ms, we could add an event handler on to fulfilling the first request, then start?
-            _indexRebuilder.RebuildIndexes(true, 5000);
+        {   
         }
 
         public void Terminate()
         {
+            _syncBootStateAccessor.Booting -= _syncBootStateAccessor_Booting;
         }
     }
 }
--- a/src/Umbraco.Web/Search/ExamineFinalComponent.cs
+++ b/src/Umbraco.Web/Search/ExamineFinalComponent.cs
@@ -56,6 +56,8 @@ namespace Umbraco.Web.Search
                     // double check lock, we must only do this once
                     if (!_initialized)
                     {
+                        _initialized = true;
+
                         if (!_mainDom.IsMainDom) return;
 
                         var bootState = _syncBootStateAccessor.GetSyncBootState();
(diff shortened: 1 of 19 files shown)


Shazwazza:
Hi @nzdev, if you have anytime at all can you please cast your eye over this one? It refactors your work a little and ensures that cold boot logic is done by the components/classes that manage their data (i.e. removed all those init callbacks).

nzdev:
@Shazwazza this has some test failures.

nzdev:
Bugfix for the test @Shazwazza https://github.com/umbraco/Umbraco-CMS/pull/10227

src/Umbraco.Web/Search/ExamineFinalComponent.cs
```diff
@@ -16,27 +18,65 @@ public sealed class ExamineFinalComponent : IComponent
...
 
-        public void Initialize()
+        /// <summary>
+        /// Once the boot state is known we can see if we require rebuilds
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void _syncBootStateAccessor_Booting(object sender, SyncBootState e)
+        {
+            UmbracoModule.RouteAttempt += UmbracoModule_RouteAttempt;
```
nzdev, line 45: Will this block routing till the index rebuilds?

src/Umbraco.Web/Search/ExamineFinalComponent.cs
```diff
@@ -16,27 +18,65 @@ public sealed class ExamineFinalComponent : IComponent
...
+                        if (!_mainDom.IsMainDom) return;
 
-            // TODO: Instead of waiting 5000 ms, we could add an event handler on to fulfilling the first request, then start?
-            _indexRebuilder.RebuildIndexes(true, 5000);
+                        var bootState = _syncBootStateAccessor.GetSyncBootState();
+
+                        _examineManager.ConfigureIndexes(_mainDom, _logger);
+
+                        // if it's a cold boot, rebuild all indexes including non-empty ones
+                        _indexRebuilder.RebuildIndexes(bootState != SyncBootState.ColdBoot, 0);
```
nzdev, line 66: Perhaps some delay may still be appropriate as the first few requests may be expensive till caches are built and indexing is very cpu/memory expensive.

src/Umbraco.Web/Search/ExamineFinalComponent.cs
```diff
@@ -16,27 +18,65 @@ public sealed class ExamineFinalComponent : IComponent
         private readonly IExamineManager _examineManager;
         BackgroundIndexRebuilder _indexRebuilder;
         private readonly IMainDom _mainDom;
-        
-        public ExamineFinalComponent(IProfilingLogger logger, IExamineManager examineManager, BackgroundIndexRebuilder indexRebuilder, IMainDom mainDom)
+        private readonly ISyncBootStateAccessor _syncBootStateAccessor;
+        private readonly object _locker = new object();
+        private bool _initialized = false;
```
nzdev, line 23: Is _initialized  ever set to true?

src/Umbraco.Web/Search/ExamineFinalComponent.cs
```diff
@@ -16,27 +18,65 @@ public sealed class ExamineFinalComponent : IComponent
...
+
+        private void UmbracoModule_RouteAttempt(object sender, RoutableAttemptEventArgs e)
         {
-            if (!_mainDom.IsMainDom) return;
+            UmbracoModule.RouteAttempt -= UmbracoModule_RouteAttempt;
 
-            _examineManager.ConfigureIndexes(_mainDom, _logger);
+            if (!_initialized)
+            {
+                lock (_locker)
```
nzdev, line 54: Does this lock need to be held past checking _initailized and IsMainDom? Could set an _initializing and return from the lock earlier to reduce blocking requests.

src/Umbraco.Web/Search/ExamineFinalComponent.cs
```diff
@@ -16,27 +18,65 @@ public sealed class ExamineFinalComponent : IComponent
...
+        /// <param name="e"></param>
+        private void _syncBootStateAccessor_Booting(object sender, SyncBootState e)
+        {
+            UmbracoModule.RouteAttempt += UmbracoModule_RouteAttempt;
+        }
+
+        private void UmbracoModule_RouteAttempt(object sender, RoutableAttemptEventArgs e)
         {
-            if (!_mainDom.IsMainDom) return;
+            UmbracoModule.RouteAttempt -= UmbracoModule_RouteAttempt;
```
nzdev, line 50: Should unregistering the event be in a lock ?

src/Umbraco.Web/Search/ExamineFinalComponent.cs
Shazwazza: nope good catch

src/Umbraco.Web/Search/ExamineFinalComponent.cs
Shazwazza: nope, index rebuilding occurs on a background thread, this just queues it

src/Umbraco.Web/Search/ExamineFinalComponent.cs
Shazwazza: Doesn't really matter if this gets executed by multiple threads

src/Umbraco.Web/Search/ExamineFinalComponent.cs
Shazwazza: This will only check on one request because the event is unbound as soon as it hits. The _initialize flag is now set as soon as it enters the double check, there won't be any blocking apart from the one request which just queues an operation.

src/Umbraco.Web/Search/ExamineFinalComponent.cs
Shazwazza: yep sure, we do that in netcore now too

Shazwazza:
The logic for this has been updated. Initialization is all done lazily (like we do in v9). There is no more Booting event since it's unnecessary.

bergmania:
`NuCache.Media.db` still seems to be rebuilt on non-cold boots?

bergmania (approved):

Shazwazza:
> Update: - It seems to only happen because I did not have any media.

I think that's because it thinks its empty so it rebuilds? I'm guessing its the same thing in v8 but would have to check

bergmania:
> I think that's because it thinks its empty so it rebuilds? I'm guessing its the same thing in v8 but would have to check

This is what happens.. I don't think that's a big problem, as it is super cheap when it is empty :) 

I'm fine with merging

PR #10185, opened 2021-04-26
Get rid of ICoreComposer

bergmania:
- Removed ICoreComposer..
- Left ExamineLucene composers, as we need these to only execute for windows (so far)


Shazwazza (approved):
Looks good. 

What do you mean by

> Left ExamineLucene composers, as we need these to only execute for windows (so far)

There are no composer/components for ExamineLucene and it runs on Linux now

Shazwazza:
Oh, i see this was a super old PR :)

PR #10190, opened 2021-04-27
Fix: Load DLLs from Package ZIPs

warrenbuckley:
`Assembly.Load()` wont load any package ZIP DLLs that were not published as part of the app, it would be OK if it was a Nuget package reference to the website

## Reference
https://github.com/dotnet/runtime/issues/13511#issuecomment-537475896

## Test Notes
* Try uploading the ZIP package before PR applied to site in the backoffice
*  Note installing a package via the backoffice drag & drop causes the site to restart & NOT a fun user experience
* Site will YSOD/error with FileNotFound Exception
* Checkout PR and see if site will now boot & run

## Example ZIP
[Take_Out_The_Trash_2.0.0.zip](https://github.com/umbraco/Umbraco-CMS/files/6384976/Take_Out_The_Trash_2.0.0.zip)

@bergmania or @Shazwazza you able to take a look at this and ensure using `LoadFrom` not gonna be a pain or a security problem etc especially as this is to do with our own TypeLoader stuff


Shazwazza:
Assembly.LoadFrom loads into a different context which we don't want for any assemblies that are actually shipped with the application. Using AssemblyLoadContext.Default.LoadFromAssemblyPath loads into the default context. IMO this is a much safer option and is probably more like what we want since we never want to support having the same assembly with different versions loaded multiple times. We also don't want to `Assembly.LoadFrom` for any assemblies that are shipped as part of the application.

Assembly.Load usage doesn't matter if it's an assembly that's shipped as part of the package and I 'think' using `AssemblyLoadContext.Default.LoadFromAssemblyPath` won't matter either but I could be wrong. IMO it would be best if we can check if the assembly is already loaded before trying to load it. Possibly something like this but maybe there's a better way

```cs
private bool IsAssemblyLoaded(string fullName,out Assembly lAssembly)
        {
            Assembly[] assemblies = AppDomain.CurrentDomain.GetAssemblies();
   
            foreach(Assembly assembly in assemblies)
            {
                if (assembly.FullName == fullName)
                {
                    lAssembly = assembly;
                    return true;
                }
            }

            lAssembly = null;
            return false;
        }
```

But then we also need to be aware of performance and track assemblies we've already checked in a dictionary, or just eagerly create a dictionary of loaded assemblies before scanning (would need to benchmark).

So I think `AssemblyLoadContext.Default.LoadFromAssemblyPath` would be the better option for non-shipped assemblies but then we need to figure out if this causes file lock issues. I suppose a prob with netcore is there's always file lock issues until the IIS shadow copy feature is rolled out by .net 6. 

That's another thing that needs testing which I'm unsure if there's a real solution for: installing a new version of the same package in the back office will try to overwrite the existing DLL which I believe will cause a file lock issue and not actually work.

warrenbuckley:
@Shazwazza I don't think this is `just` a simple thing for me to fix up as I may have suspected and probably needs yourself or Bjarke on this to ensure its done right. Rather than my attempt of just making it work.

warrenbuckley:
Closing this PR as Shan & Bjarke looking at nuget approach

PR #10195, opened 2021-04-28
Update block objects property editor references

nielslyngsoe:
fixes https://github.com/umbraco/Umbraco-CMS/issues/9663

The issue is that the angularJS scope when opening an infinite editor after the language switch did reference the old $scope of the property editor.

*How did that come?* well, we gave the Block Objects references to the property editor, so it easily can perform actions like Edit, Delete etc. But as the Block Objects are bonded to their data, so they can persist across variant-editors for runtime performance on editing content in split view, then we now have Block Objects that are specific for a property editor, though it could be used by several.

In the above case, the problem was made clear as the property editor reference initially baked into the Block Object wasn't alive anymore and the scope retrieved for complex validation is now destroyed and the experience broken.

This PR ensures that the Block Objects are updated with the latest property editor scope. This is thought a faulty architecture, as we can have invariant property editors on variant documents. In other words, the last editor to initialize sets the scope and langauge for the blocks on the other variants. Meaning the language and scope of an invariant block editor could change depending on how it was navigated. Whether this is important or not I can't tell right now, but it's a fact.

For this to be fixed 100% correctly we would need each call to a Block Object to go retrieve its current Property Editor. But this PR fixes editing Blocks, meaning we will have to look into the architectural change if it becomes a problem that we have incorrect property editor references on invariant Block List Editors.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/editor/umbeditors.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/editor/umbeditors.directive.js
@@ -135,11 +135,11 @@
     }
 
     // This directive allows for us to run a custom $compile for the view within the repeater which allows
-    // us to maintain a $scope hierarchy with the rendered view based on the $scope that initiated the 
+    // us to maintain a $scope hierarchy with the rendered view based on the $scope that initiated the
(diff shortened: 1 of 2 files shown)


madsrasmussen:
Works as expected 🎉

PR #10197, opened 2021-04-28
Enable copy/paste for Grid items

Emetico:
Ive converted Søren Kottal's "Copy Paste Grid Editor" for Umbraco 7 to work in Umbraco 8 and added it into the core instead of a plugin as part of Enable copy/paste for grid items [#6638](https://github.com/umbraco/Umbraco-CMS/issues/6638)
 
To test this i created a document with a GRID and tried to copy and paste both rows with settings and editors with content.

**Added:**
1. Copy icon beside the rows delete and settings icons, that copy the entire row with settings.
2. Copy icon beside the editors delete icon, that copy the editor with all content
3. Option to paste a row in the top when you add a new row, if any allowed rows are copied.
4. Option to paste a editor in the top of the itempicker, if the editor is allowed.
5. Option to delete copied editors and rows the same place where you paste them.
6. When a row or editor is copied a notification message is showing and telling the user that its copied.

**Need to be looked at:**
1. Using localStorageService instead of the clipboardService, because i do not know enough about the clipboardService.
2. The localStorageService alias needs to be set as a const, but i dunno how Umbraco prefer it.
3. Translation for area clipboard for keys clipboard, gridEditorCopiedToClipboard and gridRowCopiedToClipboard in other languages than English and Danish is missing.
4. Maybe some styling of the icons are needed :)


nathanwoulfe:
Hey @Emetico this sounds like a great addition - I know it's something a lot of my users will appreciate. 

Ideally we'd be using the clipboard service for consistency with the other editors. Would be cool too if that meant it was possible to copy between grid and block editor etc, assuming the same underlying element types. Not sure of the practicalities of making that work...

Feel like getting to know the clipboard service a bit better? 😁😁

Emetico:
> Hey @Emetico this sounds like a great addition - I know it's something a lot of my users will appreciate.
> 
> Ideally we'd be using the clipboard service for consistency with the other editors. Would be cool too if that meant it was possible to copy between grid and block editor etc, assuming the same underlying element types. Not sure of the practicalities of making that work...
> 
> Feel like getting to know the clipboard service a bit better? 😁😁

Hi @nathanwoulfe i have been looking a little at the clipboardService and from what i can see its calling the localStorageService with a storage key called umbClipboardService where all the copied stuff lay as a JSON. 

But if i could get pointet in the right direction i would like to learn about the clipboardService.

There are some types:
    const TYPES = {};
    TYPES.ELEMENT_TYPE = "elementType";
    TYPES.BLOCK = "block";
    TYPES.RAW = "raw";

and we could add like:
    TYPES.GRID_ROW= "gridRow";
    TYPES.GRID_EDITOR = "gridEditor";

The thing where i am a bit lost is at clipboardTypeResolvers[TYPES.GRID_EDITOR ] and how to make it work with the service.copy and how we get the copied data back.

mikecp:
Hi @Emetico , 
I took the liberty to bring my 2 cents here and add the French translations, I hope you won't mind.

If I also may make a small UI remark, I think that the copy and delete icons look a little misaligned

image

compared to nested content for example

image

So it would be great to get them aligned in the Grid as well 😁

In any case, this is a great addition 👍

garpunkal:
Anything we can do to push this PR into the main Umbraco branch? I'm happy to help contribute.

PR #10199, opened 2021-04-29
Backport cache key fix and optimizations

nikolajlauridsen:
This PR backports the fixes made by Shannon in netcore. The commit containing the netcore fixes can be seen here: https://github.com/umbraco/Umbraco-CMS/commit/8326b0428c4698552d3730ddc1167014ce7d5513

Shannon mentioned that there might be some breaking changes, however I was unable to find any since we're changing internal classes, so custom code shouldn't be affected by this.

### Setup

Before you start testing you need to set up your site to allow members to login, you can follow [this video](https://umbraco.tv/videos/umbraco-v8/content-editor/administrative-content/members/role-based-protection/) if you're unsure how to do it.

### Test steps

1. Create a user that is not approved
2. Login using casing that is not equal to the Login on the member.
    * You should not be able to log in because the member is not approved
3. Go the the backoffice and approve and save the member
4. Login with the same casing as the first attempt
5. You should now be logged in

Before this would fail and only allow you to login when using the same casing as on the member Login field

### DIFF
--- a/src/Umbraco.Core/Cache/DefaultRepositoryCachePolicy.cs
+++ b/src/Umbraco.Core/Cache/DefaultRepositoryCachePolicy.cs
@@ -20,7 +20,7 @@ namespace Umbraco.Core.Cache
     internal class DefaultRepositoryCachePolicy<TEntity, TId> : RepositoryCachePolicyBase<TEntity, TId>
         where TEntity : class, IEntity
     {
-        private static readonly TEntity[] EmptyEntities = new TEntity[0]; // const
+        private static readonly TEntity[] s_emptyEntities = new TEntity[0]; // const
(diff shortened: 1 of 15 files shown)


elit0451:
After revising [our documentation for breaking changes in Umbraco]( https://our.umbraco.com/documentation/development-guidelines/breaking-changes) just to make sure we are not missing anything, I can confirm that the backport is not introducing any breaking changes. The only doubt I had is regarding `GetKey()` in **RepositoryCacheKeys.cs** but again, as you already pointed out it is located in an internal class, so should be alright.

I also followed the video you recommended and the testing steps and I can confirm that the issue has been fixed also for v8.

Therefore, I am merging!

PR #10202, opened 2021-04-29
#10193 The listview search within a dialog (eg. minilistview) shows the loading indicator forever for zero results

### LINKED ISSUE #10193: The listview search within a dialog (eg. minilistview) shows the loading indicator forever for zero results
When you search for something that does not exist inside a listview search in a dialog (eg. the minilistview on a link picker), you'll get a loading indicator forever, and no indication that the search has failed to find anything.

I am seeing this issue on Umbraco version: 8.13

Reproduction
------------

1. Start with a fresh install with the starter kit
2. Navigate to Home and edit the page
3. Remove the existing call to action link, and add a new one
4. Navigate into the People list view
5. Type in "Hello", or anything that won't match a person in the list
6. Note console log error and forever loading

listview-search

Desired Result
------------
Like the top-level content search in the dialog, the listview search should return a localized message "Sorry, we can not find what you are looking for."

Bug Analysis
------------
Console error:
```
TypeError: Cannot read property 'forEach' of null
    at DependencyHandler.axd
    at processQueue (DependencyHandler.axd)
    at DependencyHandler.axd
    at Scope.$digest (DependencyHandler.axd)
    at Scope.$apply (DependencyHandler.axd)
    at done (DependencyHandler.axd)
    at completeRequest (DependencyHandler.axd)
    at XMLHttpRequest.requestLoaded (DependencyHandler.axd) "Possibly unhandled rejection: {}"
```

This happens here:
```
miniListView.children = data.items;
miniListView.children.forEach(function(c) {
```

The forEach fails because data.items is null.

That's because the request to run the search returns null:

`http://localhost:14250/umbraco/backoffice/UmbracoApi/Entity/GetPagedChildren?id=1114&type=Document&pageNumber=1&pageSize=10&orderBy=SortOrder&orderDirection=Ascending&filter=Hello&dataTypeKey=`

`{"items":null,"pageNumber":0,"pageSize":0,"totalPages":1,"totalItems":0}`

The controller decides to return null here when it gets zero results:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Editors/EntityController.cs#L633

I'm not sure if the fix should be applied to the C# controller (return an empty array), or the angular controller (handle a null `items` collection).

Thanks.

spudstuff:
### Prerequisites

This fixes #10193 

### Testing this fix
1. Start with a fresh install with the starter kit
2. Navigate to Home and edit the page
3. Remove the existing call to action link, and add a new one
4. Navigate into the People list view
5. Type in "Hello", or anything that won't match a person in the list
6. View now correctly shows `Sorry, we can not find what you are looking for.`

listview-search-fix

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/umb-mini-list-view.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/umb-mini-list-view.html
@@ -80,7 +80,7 @@
                     <span ng-if="search !== ''"><localize key="general_searchNoResult">Sorry, we can not find what you are looking for.</localize></span>
                 </div>
 
-                <!-- Load indicator when the list doesn't have items -->
+                <!-- Load indicator while items are retrieved -->
                 <div ng-if="miniListView.loading && miniListView.children.length === 0" class="umb-table-row umb-table-row--empty">
                     <umb-load-indicator></umb-load-indicator>
                 </div>
--- a/src/Umbraco.Web.UI.Client/src/views/components/umb-mini-list-view.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/umb-mini-list-view.html
@@ -74,7 +74,7 @@
                     <div class="umb-table-cell black" ng-class="{'umb-table-cell--faded':child.published === false}">{{ child.name }}</div>
                 </div>
 
-                <!-- Load indicator when the list doesn't have items -->
+                <!-- Show a message when no items are found, depending on whether you searched or the list was empty anyway -->
                 <div ng-if="!miniListView.loading && miniListView.children.length === 0" class="umb-table-row umb-table-row--empty">
                     <span ng-if="search === ''"><localize key="general_noItemsInList">No items have been added</localize></span>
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web.UI.Client/src/views/components/umb-mini-list-view.html
```diff
@@ -75,13 +75,13 @@ <h4 class="umb-mini-list-view__title-text">{{ miniListView.node.name }}</h4>
                 </div>
 
                 <!-- Load indicator when the list doesn't have items -->
```
rbottema, line 77: This comment here is incorrect, as the following code is not about the load indicator, right? We should either remove it, or replace it with the truth.

spudstuff: Updated comment, thanks @rbottema.

nathanwoulfe (approved):
Hey @spudstuff this looks good to me 👍

Issue #10194, opened 2021-04-28
Treepicker prevalue editor doesn't initialize values correct


bjarnef:
While working on Umbraco v8.12.2 and Contentment v1.4.1 adding a custom data source, which used `treepicker` prevalue editor, I noticed multiple values are stored as comma delimited string of udi's:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/treepicker.controller.js#L107

However on init the part is wrong, where it split by comma if `model.value` in an array, which it isn't in this case.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/treepicker.controller.js#L29-L33

So in `entityResource.getByIds()` an array with a single item with multiple udi's:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/treepicker.controller.js#L35

![image](https://user-images.githubusercontent.com/2919859/116391188-239ee300-a81f-11eb-9fc1-257fec16b77e.png)

See this as reference: https://github.com/leekelleher/umbraco-contentment/issues/101#issuecomment-827855438

nul800sebastiaan:
Sounds like something that needs to be fixed, we'd appreciate a helping hand! 👍

umbrabot:
Hi @bjarnef,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

PR #10204, opened 2021-04-29
Fix issue with show selected member types on init using treepicker

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10194

### Description
When using treepicker prevalue editor, it currently fails to show the selected entities, because it for a comma delimited string just push the value to an array and pass this into `entityResource.getByIds()`.

It should instead split by comma if the value is NOT an array.

pxjJqId4iE

I have attached a test property editor here with the prevalues: 
[TestPropertyEditor.zip](https://github.com/umbraco/Umbraco-CMS/files/6400079/TestPropertyEditor.zip)

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/treepicker.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/treepicker.controller.js
@@ -49,7 +49,7 @@ angular.module('umbraco')
             });
         }
 
-        $scope.openContentPicker = function () {
+        $scope.openTreePicker = function () {
             var treePicker = config;
             treePicker.section = config.type;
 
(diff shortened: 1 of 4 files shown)


bjarnef:
I found it a bit strange that "Edit" just open the existing tree picker and not the infinite editing for e.g. media type or member type. Not sure if it has something to do with access to settings?

nathanwoulfe (approved):
Cheers @bjarnef, looks good to me - I agree re the icons, I've hit the same issue in other places, where the mapper doesn't provide the icon (or hit the code responsible for setting the icon).

nul800sebastiaan:
I think @nathanwoulfe meant to merge this a few months ago, so now it's my pleasure to do it! ;-) 

Thanks @bjarnef - that works a lot better! 👍

nathanwoulfe:
@nul800sebastiaan I believe you may be correct. Thanks for cleaning up my mess ☺️

PR #10207, opened 2021-04-30
appsettings "WebRouting:DisableRedirectUrlTracking" should be a bool

MartinAmsinck:
### Description
When toggling UrlTracker in the backoffice, WebRouting:DisableRedirectUrlTracking is set as a string 'true'/'false' in appsettings.json.

image

Changed to a bool 

image


bergmania:
Hi @MartinAmsinck ..

Your right, thanks for the PR.. Ill merge if the buildserver is happy :) 

Also note I changed the base branch to `v9/dev`

bergmania (approved):

PR #10209, opened 2021-04-30
Fix issue with hardcoded '\' as directory separator char.

comstyle:
Fixes issue with hardcoded '\\' as directory separator char. This caused all files in packages to be restored in the root folder with '\\' in the filename on Linux.

I tested the new beta version of Umbraco 9 on Linux. When I install any package, for example by clicking on install in the forms tab, then I get an install which results in the attached screenshot of the folder listing:
directory listing with backslash in filename

Of course I also tested this patch. With this patch any package will install into the right folders. The forms plugin ends up in the App_Plugins folder. Please see:
directory listing with correct installation of package into the App_plugins folder


bergmania:
Thanks, @comstyle.. We must have missed this one..  

Fix seems correct, just waiting for buildserver to be happy

Issue #10214, opened 2021-05-01
Scheduled publishing overlay uses wrong time when enter pressed in time fields


spudstuff:
When scheduling content to publish or unpublish, use of the enter key on the keyboard can scheduled content with the wrong time to publish. If the mouse is used only, the time is set correctly.

The bug is due to the fact that the overlay dialog by default submits itself when Enter is pressed. However the time fields in flatpickr do not update the main textbox in time before that submit happens. Instead, the *current local* hour or *current local* minute is substituted into the chosen time, resulting in scheduled publishing or release happening at _the wrong time_.

https://user-images.githubusercontent.com/675943/116797138-a7a9e100-ab25-11eb-9832-611992723b68.mp4

## Umbraco version

I am seeing this issue on Umbraco version: 8.13.0

Reproduction
------------

1. Go into any content.
2. Click on the Info tab (just so you can see the publish date).
3. Hit Schedule against that content.
4. Take not of your local clock hour and minute.
5. Click on either the *Publish at* or *Unpublish at* calendars (both are affected).
6. Change the minute to something that isn't the current minute (eg. 45).
7. Change the hour to something that isn't the current hour (but in the future), eg. 23 _via the keyboard_.
8. Hit enter while still in the hour field.
9. Note that your accepted time will be put into the *Publish at* field for half a second, but the overlay will begin submitting.
10. Note that the *Publish at* field back on the Info tab will now have the correct date and minutes value, but the hour will have reverted to the local clock hour.
11. This works for any hour or minute field. The field you press enter in will be reverted to local time.

Expected Result
------------
Pressing enter in a time field should close the calendar with the accepted time, but not submit the overlay.

Impact
------------
Content is being published at the wrong time. For embargoed content in our industry that cannot be released until a certain time, this is certainly problematic.

spudstuff:
Fixed by https://github.com/umbraco/Umbraco-CMS/pull/10215, thanks.

PR #10215, opened 2021-05-01
Fix textboxes inside flatpickr popup to not cause the scheduled publi…

spudstuff:
…shing overlay to submit when enter is pressed.

Overlays listen for enter keypresses and submit themselves automatically. This can be disabled by adding the attribute `overlay-submit-on-enter="false"` on any element that might accept keypresses, such as textfields inside the flatpickr calendar. This change dynamically adds this attribute to the flatpickr popup on this overlay whenever it opens (if not already present from a previous open).

### Prerequisites

This fixes https://github.com/umbraco/Umbraco-CMS/issues/10214

### Testing

1. Go into any content.
2. Click on the Info tab (just so you can see the publish date).
3. Hit Schedule against that content.
4. Take not of your local clock hour and minute.
5. Click on either the Publish at or Unpublish at calendars (both are affected).
6. Change the minute to something that isn't the current minute (eg. 45).
7. Change the hour to something that isn't the current hour (but in the future), eg. 23 via the keyboard.
8. Hit enter while still in the hour field.
9. The calendar will close and your accepted time will be put into the Publish at field correctly.
10. The scheduled publishing overlay *will not submit*.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/overlays/schedule.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/overlays/schedule.controller.js
@@ -121,11 +121,25 @@
          * @param {any} type publish or unpublish
          */
         function datePickerShow(variant, type) {
+            var activeDatePickerInstance;
             if (type === 'publish') {
                 variant.releaseDatePickerOpen = true;
+                activeDatePickerInstance = variant.releaseDatePickerInstance;
             } else if (type === 'unpublish') {
                 variant.expireDatePickerOpen = true;
+                activeDatePickerInstance = variant.expireDatePickerInstance;
             }
+
+            // Prevent enter key in time fields from submitting the overlay before the associated input gets the updated time
+            if (activeDatePickerInstance && !activeDatePickerInstance.hourElement.hasAttribute("overlay-submit-on-enter"))
+            {
+                activeDatePickerInstance.hourElement.setAttribute("overlay-submit-on-enter", "false");
+            }
+            if (activeDatePickerInstance && !activeDatePickerInstance.minuteElement.hasAttribute("overlay-submit-on-enter"))
+            {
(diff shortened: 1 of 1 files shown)


bjarnef:
I think it may be a better solution to update Flatpickr dependency to latest v4.5.x or v4.5.5 which handle enter keypress on time inputs:
https://github.com/flatpickr/flatpickr/releases/tag/v4.5.5

spudstuff:
> I think it may be a better solution to update Flatpickr dependency to latest v4.5.x or v4.5.5 which handle enter keypress on time inputs:
> https://github.com/flatpickr/flatpickr/releases/tag/v4.5.5
> 
> Currently Umbraco has a dependency on Flatpickr v4.5.2

I think that will still submit the overlay @bjarnef. It _may_ use the correct time but it won't give you a chance to put in an unpublish date or any stuff around variants (I don't run a multi-variant site so I'm not sure what the variant actions look like in this dialog).

https://github.com/flatpickr/flatpickr/commit/d85f5a781335e52681ca115bbfb4c5edb790137f

This change won't stop the keycode event from propagating.

bjarnef:
I haven't tried upgrading the flatpickr component, but it would solve some other bugs as well.

It may has events to hook into to stop event bubbling. Although this fix in for this specific case, it won't solve if we then use the datepicker in other overlays or developers use the datepicker in a custom overlay.

nul800sebastiaan:
I upgraded flatpickr to the latest (4.6.9) but that didn't help. Besides, the new behavior in this PR is much nicer, we shouldn't just schedule something when pressing enter, it's good to stay in the dialog in case we also want to schedule an unpublish.

So, good job @spudstuff - sorry for the late reply but this will be part of the next minor! 👍

nul800sebastiaan:
To be clear, the upgrade to flatpickr 4.6.9 will also be part of the next minor :)

PR #10216, opened 2021-05-03
V9: Implementing Error Describers for translated error messages

elit0451:
## Details

Implementing the following error describers:
- BackOfficeErrorDescriber
- MembersErrorDescriber

By using our own translation engine, `ILocalizedTextService`,  to return localized error message from identity correctly.
Translation keys that were missing for the specific methods of error describers are added to the default language file **en_us.xml**.

---

## Notes
- Every time the IdentityErrorDescriber methods mention "role" we refer to it as _group_;
- For **RecoveryCodeRedemptionFailed** I used `login/resetCodeExpired` key which I found similar to the original description `Recovery code redemption failed`. Let me know if you think that we should add a new key for that!;
- I saw that we use both  member role and member group when referring to members but I kept it as _group_ as I wasn't sure;
- `DuplicateUsername` and `InvalidUsername` are reused for members.

### DIFF
--- a/src/Umbraco.Infrastructure/Security/BackOfficeErrorDescriber.cs
+++ b/src/Umbraco.Infrastructure/Security/BackOfficeErrorDescriber.cs
@@ -1,4 +1,6 @@
 using Microsoft.AspNetCore.Identity;
+using Umbraco.Cms.Core.Services;
+using Umbraco.Extensions;
 
 namespace Umbraco.Cms.Core.Security
 {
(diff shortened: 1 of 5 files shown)


bergmania (commented):
It looks like `BackOfficeErrorDescriber` and `MembersErrorDescriber` has alot of the same implementations.. Maybe we should introduce a common `UmbracoErrorDescriberBase` class, so we only have the to overwrite those that are different? between users and members in each class?

elit0451:
Good suggestion, I will get right on it!

bergmania (approved):
Looks good :)

PR #10221, opened 2021-05-04
Setting the default culture when no domain rule is defined breaks "ContentFinderByUrl"

### LINKED ISSUE #10218: NetCore: Only "en-US" pages are reachable.
### Bug summary

For an unknown reason, this release routes only pages in "en-US".
If I change the default language to any other language than "en-US" the page will not be reachable anymore.
Video sample: http://www.youtube.com/watch?v=pCmTfT4Mu30

### Specifics

Umbraco: 9.0.0-beta001
Tested on Chrome, Edge & Firefox

### Steps to reproduce

* Setup a simple website
* Add an empty document type with "Allow vary by culture" checked.
* Create a page in content
* Save & Publish
* The page is available 👍
* Go in Settings>Languages
* Update the default language to any other language than "en-US"
* Return to Content
* Save & Publish
* In "Links" we have now: "This document is published but its URL cannot be routed" 😭

### Expected result

The page should be routable whatever its language.

### Actual result

Only en-US pages are reachable (even "en" are not routed).

olibos:
fixes #10218

### DIFF
--- a/src/Umbraco.Core/Routing/PublishedRouter.cs
+++ b/src/Umbraco.Core/Routing/PublishedRouter.cs
@@ -319,8 +319,6 @@ namespace Umbraco.Cms.Core.Routing
             {
                 // not matching any existing domain
                 _logger.LogDebug("{TracePrefix}Matches no domain", tracePrefix);
-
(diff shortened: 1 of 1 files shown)


olibos:
Not enough tested on my side sorry

PR #10222, opened 2021-05-04
Netcore: dotnet new template updates and change HideTopLevelNodeFromPath default value to `true`

warrenbuckley:
* Fixes the dotnet new template issue due to non escaped quotes
* Sets the default in code to HideTopLevel to true
* Explicitly add JSON schema URL to appsettings in case IDEs do no auto map it for the user


bergmania (approved):
Looks good..

Issue #10162, opened 2021-04-21
FileUploadCleanupFilterAttribute with no incoming model doesn't work


AndyButland:
`FileUploadCleanupFilterAttribute` is used to clean up uploaded files from the `~/App_Data/TEMP/FileUploads` folder once they have been moved into the correct location.

It has two modes - indicated by the boolean parameter - and in Core seems we're only using the form where `incomingModel == true`.  In Forms though, we have a use case where `incomingModel == false`, in `PreValueFileController.PostAddFile()`.

In this case what I'm find is:
- A warning of `The context.Result is null!!??`, which is triggered within the filter when the `ActionExecutingContext` is null.
- The attribute effectively short-circuiting the response, so an empty OK result is return before the main method has returned it's result - so on the client-side there's an empty response instead of the expected data.  I can see this with a breakpoint on the method.

PR #10230, opened 2021-05-05
Fixes FileUploadCleanupFilterAttribute when incomimg model is not provided.

AndyButland:
Fixes issue #10162.

As noted in the issue, we aren't using this code path in CMS but we do in Forms.  I've tested this by implementing the code first in the Forms repo and then once verified copied into CMS for this PR.


bergmania (approved):
Looks good 💪

PR #10233, opened 2021-05-06
V9: Reintroduce missing friendly get crop url extensions

bergmania:
Reintroduce missing friendly `GetCropUrl` extensions by using the internal service locator to inject the required services.

### DIFF
--- a/src/Umbraco.Web.Common/Extensions/FriendlyImageCropperTemplateExtensions.cs
+++ b/src/Umbraco.Web.Common/Extensions/FriendlyImageCropperTemplateExtensions.cs
@@ -1,0 +1,319 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Umbraco.Cms.Core.Media;
+using Umbraco.Cms.Core.Models;
+using Umbraco.Cms.Core.Models.PublishedContent;
+using Umbraco.Cms.Core.PropertyEditors.ValueConverters;
+using Umbraco.Cms.Core.Routing;
+using Umbraco.Cms.Web.Common.DependencyInjection;
+
+namespace Umbraco.Extensions
+{
+    public static class FriendlyImageCropperTemplateExtensions
+    {
+        private static IImageUrlGenerator ImageUrlGenerator { get; } =
+            StaticServiceProvider.Instance.GetRequiredService<IImageUrlGenerator>();
+
+        private static IPublishedValueFallback PublishedValueFallback { get; } =
+            StaticServiceProvider.Instance.GetRequiredService<IPublishedValueFallback>();
+
+        private static IPublishedUrlProvider PublishedUrlProvider { get; } =
+            StaticServiceProvider.Instance.GetRequiredService<IPublishedUrlProvider>();
+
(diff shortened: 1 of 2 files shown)


src/Umbraco.Web.Common/Extensions/FriendlyImageCropperTemplateExtensions.cs
```diff
@@ -0,0 +1,319 @@
...
+        /// Use focal point to generate an output image using the focal point instead of the predefined crop if there is one
+        /// </param>
+        /// <param name="useCropDimensions">
+        /// Use crop dimensions to have the output image sized according to the predefined crop sizes, this will override the width and height parameters
+        /// </param>
+        /// <param name="cacheBusterValue">
+        /// Add a serialized date of the last edit of the item to ensure client cache refresh when updated
+        /// </param>
+        /// <param name="furtherOptions">
+        /// These are any query string parameters (formatted as query strings) that ImageProcessor supports. For example:
```
AndyButland, line 270: Guess we shouldn't really be referring to "ImageProcessor" in the intellisense comments.  And not "ImageSharp" either as it could be replaced with something else in any particular solution.  So perhaps replace these references with "the underlying image processing service" or similar?

There's at least a couple references to this in the comments.

src/Umbraco.Web.Common/Extensions/FriendlyImageCropperTemplateExtensions.cs
```diff
@@ -0,0 +1,319 @@
...
+        /// </example>
+        /// </param>
+        /// <param name="ratioMode">
+        /// Use a dimension as a ratio
+        /// </param>
+        /// <param name="upScale">
+        /// If the image should be upscaled to requested dimensions
+        /// </param>
+        /// <returns>
+        /// The <see cref="string"/>.
```
AndyButland, line 200: Maybe better as "the URL of the cropped image" or similar.

AndyButland (approved):
Just a couple of comments on the comments :-)

bergmania:
Make sense.. I updated the comments, and the original source of these comments :P

Jeavon:
In v8 there was a move to get developers to use GetCropUrl extensions methods on UrlHelper and discourage the use of the "old" IPublishedContent extension methods. In v9 the UrlHelper extension methods are not present so am I right that using the new friendly extension methods is the correct usage... (full circle)?

bergmania:
> In v8 there was a move to get developers to use GetCropUrl extensions methods on UrlHelper and discourage the use of the "old" IPublishedContent extension methods. In v9 the UrlHelper extension methods are not present so am I right that using the new friendly extension methods is the correct usage... (full circle)?

I think we should just reintroduce the extensions on UrlHelper (`IUrlHelper`) if they are missing in v9 too..

Jeavon:
Ah ok, well there are all commented at https://github.com/umbraco/Umbraco-CMS/blob/v9/dev/src/Umbraco.Web/UrlHelperRenderExtensions.cs#L16

Issue #10232, opened 2021-05-05
Doctypes with parent doctypes won't compile in modelsbuilder


Shadyn:
When a doctype inherits from another doctype, modelsbuilder fails on compile (using appdata for mode) by saying the parent doctype does not have a publishedValueFallback

Reproduction
------------

### Bug summary

When a doctype inherits from another doctype, modelsbuilder fails on compile (using appdata for mode) by saying the parent doctype does not have a publishedValueFallback.  Only by adding another constructor to the genereated class files can I get it to compile.  The parent class may need a constructor that doesn't contain the publishedValueFallback parameter.  NOTE: These doctypes were not created on the beta - they came over in the upgrade

![image](https://user-images.githubusercontent.com/4174228/117160713-f99d7180-ad86-11eb-90a8-b40754c5e263.png)

### Specifics

This has happened in every version of the beta so far

### Steps to reproduce

- Upgraded from v8 to v9 beta
- Generated models
- Compile fails on every doctype that is inherited in other doctypes

### Expected result

Models would build like in V8

### Actual result

All doctypes using inheritance fail on build with "There is no argument given that corresponds to the required formal parameter 'publishedValueFallback''"

bergmania:
Hi @Shadyn 

Thanks for reporting. I can confirm I can reproduce

PR #10234, opened 2021-05-06
V9: Bugfix for modelsbuilder when having inherited doctypes

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/10232

# Details
- Fixed issue with ModelsBuilder that was failing when you created inherited doc types

# Test
- Create an inherited doc type:
image
- Create some content with that nested doc type (HomePage in the image)
- Ensure you can use ModelsBuilder properties from both the child and parent doc type (Layout and Homepage in the image)


AndyButland:
Confirmed in testing and code looks fine, so merging.

Issue #10184, opened 2021-04-26
Performance issues with nested block list editors


BrionyMcKenz:
Nested block list editors are very slow to load when editing content, particularly when there are lots of blocks.

## Umbraco version

I am seeing this issue on Umbraco version: 8.13.0

Reproduction
------------

### Bug summary

I have a block list with 8 available blocks. 3 of these blocks contain multiple block list editor properties as I'm using them to create a grid structure (based on this article https://24days.in/umbraco-cms/2020/grid-nouveau-block-list/). I am using custom views to display the nested block list editors. There is a request called GetEmptyByKey being called for every content and settings model for every available block in every block list editor (even though I'm using the same block list editor data type everywhere). For every block I add, there is an additional ~20 requests because the GetEmptyByKey is being called twice for every block list editor property.  

### Specifics

    * Happening on any back office node where I have added nested block list editors.
    * Using Chrome 90.0.4430.85

### Steps to reproduce

Create a block list editor property and add to a content node. Add a block to this that has multiple block list editor properties. Go to the content node and add a block that has multiple block list editor properties. 

### Expected result

Some level of caching or optimisation so that the block models are not being requested over and over again despite being the same for each block list editor.

### Actual result

Lots of requests for the same content resulting in very slow load/render times when editing content.

---
_This item has been added to our backlog [AB#16689](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/16689)_

patrickdemooij9:
I agree with you. Nested Content has the same issue here. I wonder if we should just merge all getEmpty requests together, but I don't know if there was a reason to keep them separate

emma-hq:
Hey folks, 

Thanks for reporting this @BrionyMcKenz and for chipping in @patrickdemooij9 - we'll have a chat internally and see what we can suggest. 

Emma

emma-hq:
Hey peeps,

After a bit of discussion, I've been pointed in the direction of these performance changes that are currently in the works. Would you like to take a look and decide if this will address your issue? It looks promising to me. If so, I'll link the issue to that PR and it will close this automatically once the PR is merged.

See here: https://github.com/umbraco/Umbraco-CMS/pull/10134

Emma

BrionyMcKenz:
[...]

PR #10235, opened 2021-05-06
Optimizations related to BlockEditorPropertyEditor.ToEditor()

elit0451:
Aims at eliminating/decreasing the Block editor performance issues.
Reference: https://github.com/umbraco/Umbraco-CMS/issues/10184

## Details
<s>- Changing the built-in cache at **BlockEditorValues** level for Content Types - before that dictionary cache was never used because it was short-lived and re-created while looping through block items. Now, it is created in **BlockEditorPropertyEditor** and passed on, so it can be reused.</s>
- Calling `GetValueEditor()` for unique data type ids - reducing the number of objects allocated
- Adding explicit scope around `GetEmptyByKey()` - so everything is handled in 1 scope
- Introducing `GetEmptyByKeys()` which collects all the block list editor items and executes a single call for block list editor instead of making numerous requests as mentioned in #10184 
- Fix tests
</br>

_Before_:

⬆️ all ❗ marks from the marked one down are all from GetEmptyByKey  

</br>
</br>
</br>

_After_:
image

## Test
- Follow the steps mentioned in the above issue and review the optimizations

---
Note:
This PR contains client-side code changes, so remember to run `npm run build` in Umbraco.Web.UI.Client


src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -367,13 +373,17 @@ public ContentItemDisplay GetEmpty(string contentTypeAlias, int parentId)
...
-            }
+                var contentType = Services.ContentTypeService.Get(contentTypeKey);
+                if (contentType == null)
+                {
+                    throw new HttpResponseException(HttpStatusCode.NotFound);
+                }
+                scope.Complete();
 
-            return GetEmpty(contentType, parentId);
+                return GetEmpty(contentType, parentId);
```
nikolajlauridsen, line 385: I could be wrong, but I think we should call `scope.Complete()` after doing `GetEmpty` as far as I can see `GetEmpty` performs write operations through `Services.ContentService.Create` so if something goes wrong I think we want to have that rolled back.

nikolajlauridsen (commented):
Overall looks good to me, found one minor thing in ContentController though

src/Umbraco.Web/Editors/ContentController.cs
elit0451: Good point!

nikolajlauridsen (approved):
Just finished testing, everything seems to work and I'm seeing a performance improvement as well, as you also showed in your description 👍

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -367,13 +373,19 @@ public ContentItemDisplay GetEmpty(string contentTypeAlias, int parentId)
...
-            }
+                var contentType = Services.ContentTypeService.Get(contentTypeKey);
+                if (contentType == null)
+                {
+                    throw new HttpResponseException(HttpStatusCode.NotFound);
+                }
 
-            return GetEmpty(contentType, parentId);
+                var contentItem = GetEmpty(contentType, parentId);
+                scope.Complete();
```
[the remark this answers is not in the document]
elit0451: Good tip, thanks, will have it in mind for the next time ;)

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -392,6 +404,37 @@ private ContentItemDisplay GetEmpty(IContentType contentType, int parentId)
...
+                var contentTypes = Services.ContentTypeService.GetAll(contentTypeKeys).ToList();
+
+                if (contentTypes.Any(contentType => contentType == null))
+                {
+                    throw new HttpResponseException(HttpStatusCode.NotFound);
+                }
+
+                foreach (var contentTypeKey in contentTypeKeys)
+                {
+                    var contentType = contentTypes.First(c => c.Key == contentTypeKey);
```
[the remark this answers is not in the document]
elit0451: You are absolutely right, a total oversight and misleading from the IDE static optimizer. I Will fix it up, thanks 🙌

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -392,6 +404,32 @@ private ContentItemDisplay GetEmpty(IContentType contentType, int parentId)
...
+        public IDictionary<Guid, ContentItemDisplay> GetEmptyByKeys([FromUri] Guid[] contentTypeKeys, [FromUri] int parentId)
+        {
+            var result = new Dictionary<Guid, ContentItemDisplay>();
+
+            using var scope = _scopeProvider.CreateScope(autoComplete: true);
+            var contentTypes = Services.ContentTypeService.GetAll(contentTypeKeys).ToList();
+
+            foreach (var contentType in contentTypes)
+            {
+                if (contentType is null)
```
Shazwazza, line 422: This whole block could just be:

```cs
var result = Services.ContentTypeService.GetAll(contentTypeKeys)
   .WhereNotNull()
   .ToDictionary(x => x.Key, x => GetEmpty(x, parentId));
```

src/Umbraco.Web/PropertyEditors/BlockEditorPropertyEditor.cs
```diff
@@ -118,7 +120,7 @@ public override object ToEditor(Property property, IDataTypeService dataTypeServ
                     return string.Empty;
                 }
 
-                if (blockEditorData == null || blockEditorData.BlockValue.ContentData.Count == 0)
+                if (blockEditorData == null)
```
Shazwazza, line 123: do you recall why this change was made?

Shazwazza (commented):
@elit0451 + @nikolajlauridsen Left a question and a minor comment. I don't think these should block merging this PR but if you can respond (or make the change) that would be great. @nikolajlauridsen I'll look at your other PR now but if you have time tomorrow, feel free to merge this one and we'll flag as 8.15.

src/Umbraco.Web/PropertyEditors/BlockEditorPropertyEditor.cs
elit0451: If you have a closer look to what exactly `blockEditorData = _blockEditorValues.DeserializeAndClean(val)` is doing (a few lines above), you will notice that we already do the check I removed ([`blockEditorData.BlockValue.ContentData.Count == 0`](https://github.com/umbraco/Umbraco-CMS/blob/d8b7e24b7f575e8b756f950e4de43a21e4da8795/src/Umbraco.Web/PropertyEditors/BlockEditorPropertyEditor.cs#L353)). And if the count is indeed 0, we will return `null` that would be caught in the single `is null` check 🙂

nikolajlauridsen: Ahh yes, and `DeserializeAndClean` might return null, hence the null check 👍

src/Umbraco.Web/Editors/ContentController.cs
nikolajlauridsen: Ah good catch, however, I'm changing this method in the other PR, so I think I'll just merge this as-is for now, just to avoid a merge conflict 😄

PR #10236, opened 2021-05-06
Nested content optimizations

elit0451:
## Details

<s>- Changing the built-in cache at **NestedContentValues** level for Content Types - before that dictionary cache was never used because it was short-lived and re-created while looping through the items. Now, it is created in **NestedContentPropertyValueEditor** and passed on, so it can be reused.</s>
- Calling `GetValueEditor()` for unique data type ids - reducing the number of objects allocated

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
+++ b/src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
@@ -35,6 +35,7 @@ namespace Umbraco.Web.PropertyEditors
         private readonly IDataTypeService _dataTypeService;
         private readonly IContentTypeService _contentTypeService;
         private readonly ILocalizedTextService _localizedTextService;
+        private readonly Lazy<Dictionary<string, IContentType>> _contentTypes;
         internal const string ContentTypeAliasPropertyKey = "ncContentTypeAlias";
 
         [Obsolete("Use the constructor specifying all parameters instead")]
@@ -48,6 +49,7 @@ namespace Umbraco.Web.PropertyEditors
             _dataTypeService = dataTypeService;
             _contentTypeService = contentTypeService;
             _localizedTextService = localizedTextService;
+            _contentTypes = new Lazy<Dictionary<string, IContentType>>(() => contentTypeService.GetAll().ToDictionary(c => c.Alias));
         }
 
         // has to be lazy else circular dep in ctor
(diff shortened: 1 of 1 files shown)


nikolajlauridsen (approved):
Looks good to me 👍 💪

mikecp (approved):

mikecp:
Thanks @elit0451 for this update!
Unfortunately I can't merge it as it is a branch on the umbraco/Umbraco-CMS repo directly...

src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
```diff
@@ -48,6 +49,7 @@ public NestedContentPropertyEditor(ILogger logger, Lazy<PropertyEditorCollection
             _dataTypeService = dataTypeService;
             _contentTypeService = contentTypeService;
             _localizedTextService = localizedTextService;
+            _contentTypes = new Lazy<Dictionary<string, IContentType>>(() => contentTypeService.GetAll().ToDictionary(c => c.Alias));
```
Shazwazza, line 52: Sorry, same issue as I mention in the block list change https://github.com/umbraco/Umbraco-CMS/pull/10235#pullrequestreview-657110265, this part will need to be reverted.

src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
```diff
@@ -184,9 +187,16 @@ public override object ToEditor(Property property, IDataTypeService dataTypeServ
                                 continue;
                             }
 
-                            var tempConfig = dataTypeService.GetDataType(prop.Value.PropertyType.DataTypeId).Configuration;
-                            var valEditor = propEditor.GetValueEditor(tempConfig);
-                            var convValue = valEditor.ToEditor(tempProp, dataTypeService);
+                            var dataTypeId = prop.Value.PropertyType.DataTypeId;
+                            if (!valEditors.ContainsKey(dataTypeId))
```
Shazwazza, line 191: valEditors.TryGetValue is a faster way to lookup and retrieve from a dictionary at the same time.

Shazwazza (changes requested):
Left some notes inline

skttl:
Is it possible that the optimization in `ToEditor()` can be applied to `FromEditor()` too?

Doc Type Grid Editor suffers from sql calls in FromEditor - https://github.com/skttl/umbraco-doc-type-grid-editor/issues/194

src/Umbraco.Web/PropertyEditors/NestedContentPropertyEditor.cs
elit0451: Addressed

nul800sebastiaan:
Cherry picked for 8.15.0 3e89e468406f4eb3ad33522bcad8a357fc7bff14

PR #10237, opened 2021-05-06
Deterministic Builds for Sourcelink 

warrenbuckley:
Get more green checkmarks for Nuget builds

## Resources
https://github.com/clairernovotny/DeterministicBuilds
https://mitchelsellers.com/blog/article/net-5-deterministic-builds-source-linking
https://lurumad.github.io/using-source-link-in-net-projects-and-how-to-configure-visual-studio-to-use-it

image

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -17,5 +17,24 @@
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>https://github.com/umbraco/umbraco-cms</RepositoryUrl>
   </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- SourceLink: Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element)-->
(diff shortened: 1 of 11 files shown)


bergmania (approved):
Looks good.. Do we need to set the `TF_BUILD` variable or is it preset in azure pipelines?

warrenbuckley:
>Deterministic builds require a property to be set to true during CI: ContinuousIntegrationBuild. These should not be enabled during local dev or the debugger won't be able to find the local source files.
>
>Therefore, you should use your CI system's variable to set them conditionally. For Azure Pipelines, it looks like this
>
>```xml
><PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
>  <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
></PropertyGroup>
>```

warrenbuckley:
Also lets wait for the CI build to finish up and check its all OK

PR #10244, opened 2021-05-07
V9: Do not save runtime level in ctors

bergmania:
# Details:

Fixed some bugs that were caused by us saving the runtime level in the ctor, before it was determined.

# Test:
1) You can use the debugger and ensure the runtime level is now not `Unknown` anymore.
2) You can change the default language and use a dictionary item in a view. Then you will now see the correct default language from the service is used instead of the one in config.


AndyButland:
Code looks fine and test cases pass - merging in.

PR #10247, opened 2021-05-07
Added target="_blank" to "Update available" link

abjerner:
Feels a bit weird that the **Update available** link opens in the same window, so I've set the `target` option to `_blank`.

image


mikecp (approved):

mikecp:
Thanks for the PR @abjerner, it seems to make more sense indeed 😀👍

PR #10250, opened 2021-05-07
V9 - Include models folder in the project

### LINKED ISSUE #10213: Models Builder generated classes cannot be used straight away.
I have configured Models Builder as LiveAppData and generated the models correctly, but the classes cannot be used in code because they are created in the project with a Build Action of 'None', they should be as 'C# Compiler'

Reproduction
------------

### Bug summary

I cannot use the generated Models Bulder classes straight away. Changing their Build Action from None to 'C# compiler' fixes the problem.

### Specifics

Umbraco V9 beta1.
Models Builder configured as LiveAppData.

### Steps to reproduce

1. Install Umbraco V9 beta1.
2. Configure Models Builder as LiveAppData.
3. Generate Models.
4. Classes are created on the /umbraco/models folder with a None Build Action.

### Expected result

Classes should be available to use in the code straight away.

### Actual result

Other classes on the project cannot discover the generated classes.

skartknet:
Fixes #10213 When the models folder is included in the project the cs files have a Build Action of 'C# compiler' so they can be used in code.

As a bonus, I have fixed a couple of comments that indicate the right models' location.

### DIFF
--- a/build/templates/UmbracoProject/UmbracoProject.csproj
+++ b/build/templates/UmbracoProject/UmbracoProject.csproj
@@ -18,16 +18,13 @@
         <Compile Remove="umbraco\Data\**" />
         <Compile Remove="umbraco\logs\**" />
         <Compile Remove="umbraco\MediaCache\**" />
-        <Compile Remove="umbraco\models\**" />
     </ItemGroup>
     <ItemGroup>
         <EmbeddedResource Remove="umbraco\Data\**" />
         <EmbeddedResource Remove="umbraco\logs\**" />
         <EmbeddedResource Remove="umbraco\MediaCache\**" />
-        <EmbeddedResource Remove="umbraco\models\**" />
     </ItemGroup>
     <ItemGroup>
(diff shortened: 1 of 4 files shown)


skartknet:
Ok, I'll listen to you 😄 I just did that change on the project file and all seems to be working fine so I thought no code changes were really needed.

skartknet:
It seems to be working fine now. I have modified the PureLive models location to '~/umbraco/data/Temp/PureLive' and added a rule on the VS project to include the generated AppData models that are generated on '~/umbraco/models/`

bergmania:
Thanks, @skartknet.
Looks good.. I cleaned up a bit more in the project file and aligned the project file used for newly generated projects (From the dotnet new template).

Tested PureLive, which still works.. 
Changed to App_Data and it worked without I had to modify the project file 💪 

The only minor concern, which I don't have any idea how to archive is that you still have to clean you `/umbraco/models/` if you change back to PureLive. - Not a big deal, so I will merge this one :)

PR #10257, opened 2021-05-10
Ability to configure a remote URL for displaying different content on the content dashboard

elit0451:
## Details 

Adds the `ContentDashboardBaseUrl` appSetting in the Web.config file, so the intro dashboard ("Getting Started") can be used to display different content on Umbraco Cloud. The change will affect all new projects. In the case that the config setting is not present the default content-endpoint will be called, therefore not changing any existing behaviour.

Additionally, more information is reported from the existing dashboard - a `hosting` parameter is added by the end to indicate if the solution is running on Umbraco Cloud.
- If that is the case **Umbraco.Deploy.Cloud.dll** will exist

## Test:
1. Test 1
   1. Before switching to this branch, check what content is being displayed on the "Getting Started" dashboard.
   2. Switch to this branch and access the dashboard again, verifying that nothing has changed.
2. Test 2
   1.  Check that since you don't have _Umbraco.Deploy.Cloud.dll_ in **...src\Umbraco.Web.UI\bin** the value of the `hosting` parameter is `onpremises` - mimicking installation that is not on Umbraco Cloud.
3. Test 3
   1. Change the value of _Umbraco.Deploy.Cloud.dll_ (inside `IOHelper.IsOnCloud`) to a dll that is present in **...src\Umbraco.Web.UI\bin** & verify that the `hosting` value is `cloud`.
   2. Add `<add key="Umbraco.Core.ContentDashboardBaseUrl" value="https://dashboard.umbraco.org/"/>` to Web.config and check the content persists. (later on, this endpoint should be changed but for the moment, we don't have a different one for testing)


elit0451:
Thanks for the feedback @AndyButland 💪 I still haven't heard from Filip about 1 but I tried to address points 2, 3 and 4

AndyButland:
Looks good to me.  I guess `IOHelper` is as good a place as any to put this check.

Would just suggest to replace `Unknown` with `OnPremises`.  Really just as if something isn't hosted on Cloud, it must be "on-premises" (at least in the sense that the product "Umbraco Deploy On-Premises" means "anywhere other than cloud").  And "unknown" suggests we don't know if it's on cloud or not.

elit0451:
I used "Unknown" as we really wouldn't know if it is on any other cloud provider or not, but sure thing, I will change it :)

AndyButland:
I've made some updates to this following a clarification of the requirements.  Now the URL is hard-coded, but the path can be amended in config.  By default, without any config, the URL looked for will be https://dashboard.umbraco.com/cms.

In config, the following app setting can be applied:

```
<add key="Umbraco.Core.ContentDashboardPath" value=""/>      
```

The value provided here will replace the path part of the URL.  The intention is that Cloud sites will be updated to have a value of "cloud" (and in future, we could have "heartcore" and "uno" too).

To Test:

- With no config, the content should be retrieved from the default URL.
- With config applied, the path should be overridden by that added to the app setting.

nul800sebastiaan:
Cherry picked for 8.15 in 861fc40572c82564a5f925bdda36323d94033f55

PR #10262, opened 2021-05-11
Creates a method allowing validation of a user's credentials without actually logging them in.

AndyButland:
This is used in Deploy, and was available in V8, so have made it available in V9.

There's a unit test to verify functionality, but you can also test by dropping in the following controller into the code base:

```
using System.Threading.Tasks;
using Umbraco.Cms.Core.Security;
using Umbraco.Cms.Web.Common.Attributes;
using Constants = Umbraco.Cms.Core.Constants;

namespace Umbraco.Cms.Web.BackOffice.Controllers
{
    [PluginController(Constants.Web.Mvc.BackOfficeApiArea)]
    public class TestController : UmbracoAuthorizedApiController
    {
        private readonly IBackOfficeUserManager _backOfficeUserManager;

        public TestController(
            IBackOfficeUserManager backOfficeUserManager) => _backOfficeUserManager = backOfficeUserManager;

        public async Task<ValidateCredentialsResult> ValidateCredentials(string username, string password) =>
            new ValidateCredentialsResult
            {
                UserName = username,
                Password = password,
                Result = await _backOfficeUserManager.ValidateCredentialsAsync(username, password),
            };

        public class ValidateCredentialsResult
        {
            public string UserName { get; set; }

            public string Password { get; set; }

            public bool Result { get; set; }
        }

    }
}
```

And requesting via:

```
https://localhost:44331/umbraco/backoffice/umbracoapi/test/ValidateCredentials?username=<username>&password=<password>
```

### DIFF
--- a/src/Umbraco.Infrastructure/Security/IUmbracoUserManager.cs
+++ b/src/Umbraco.Infrastructure/Security/IUmbracoUserManager.cs
@@ -375,5 +375,13 @@ namespace Umbraco.Cms.Core.Security
         /// A user can only support a phone number if the BackOfficeUserStore is replaced with another that implements IUserPhoneNumberStore
         /// </remarks>
         Task<string> GetPhoneNumberAsync(TUser user);
+
+        /// <summary>
+        /// Validates that a user's credentials are correct without actually logging them in.
+        /// </summary>
+        /// <param name="username">The user name.</param>
+        /// <param name="password">The password.</param>
+        /// <returns>True if the credentials are valid.</returns>
+        Task<bool> ValidateCredentialsAsync(string username, string password);
     }
 }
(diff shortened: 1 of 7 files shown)


bergmania (approved):
Looks good to me

nikolajlauridsen (approved):
Looks good to me too, tested it out as well, and I'm getting expected results 👍 

Just need the tests to pass and then we can merge

rbottema:
As this can be used to verify if a password is correct, shouldn't this have some sort of rate limit to prevent attacks? Or does some of the underlying code take care of this?

AndyButland:
Thanks for checking @rbottema - but I think this is OK.  The underlying code is a method on a service, but isn't actually exposed to be requested via a controller action or other endpoint (hence me including a "test" controller in the PR so it could be accessing for testing).

PR #10264, opened 2021-05-11
Fixed issue with data types not being cleared in cache refresh operations,

AndyButland:
The results of this issue have been reported a couple of times on the Deploy tracker: issues [49](https://github.com/umbraco/Umbraco.Deploy.Issues/issues/49) and [34](https://github.com/umbraco/Umbraco.Deploy.Issues/issues/34).

The issue manifests itself by when data types are updated via Deploy, their old values are cached and not updated until a website restart. I found I could fix this by getting a reference to the appropriate "isolated cache" and clearing by key, but for other entities we don't have to do this - instead relying on the cache refreshers in core.

Then seeing this it looked like whilst we were clearing caches for related entities - e.g. content and content types - we weren't clearing for the data type itself.

If you compare e.g. `MemberCacheRefresher` and `MacroCacheRefresher` you can see we are doing what I've added here for `DataTypeCacheRefresher` - so it looks to me that this is a safe and necessary update.

Otherwise it's a bit tricky to test.  The way I did it was by building and copying in the dlls from core with this update applied into my local Deploy test website, and observing that the issue is resolved with this update.

### DIFF
--- a/src/Umbraco.Web/Cache/DataTypeCacheRefresher.cs
+++ b/src/Umbraco.Web/Cache/DataTypeCacheRefresher.cs
@@ -3,6 +3,7 @@ using Umbraco.Core;
 using Umbraco.Core.Cache;
 using Umbraco.Core.Models;
 using Umbraco.Core.Models.PublishedContent;
+using Umbraco.Core.Persistence.Repositories.Implement;
 using Umbraco.Core.PropertyEditors.ValueConverters;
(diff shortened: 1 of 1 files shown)


AndyButland:
@nul800sebastiaan - just pinging you for information on this one, resolving a couple of issues reported in Deploy, but looks like it needs to be resolved in CMS.

PR #10266, opened 2021-05-12
Re-implemented various file service methods from V8 required in Deploy.

AndyButland:
As the title says, this PR is adding in some methods that weren't migrated from V8, and presumably not needed in CMS, so they are available for use in Deploy.

I've done a little tidy-up but mostly the code is just a copy from V8.

They can be tested in views for example with:

```
@inject Umbraco.Cms.Core.Services.IFileService FileService
<p>File size: @FileService.GetPartialViewFileSize(@"C:\Repos\Umbraco\Umbraco-CMS-V9\src\Umbraco.Web.UI.NetCore\Views\Partials\ChildDocType.cshtml")</p>
```

### DIFF
--- a/src/Umbraco.Infrastructure/Persistence/Repositories/Implement/TemplateRepository.cs
+++ b/src/Umbraco.Infrastructure/Persistence/Repositories/Implement/TemplateRepository.cs
@@ -39,39 +39,38 @@ namespace Umbraco.Cms.Infrastructure.Persistence.Repositories.Implement
             _viewHelper = new ViewHelper(_viewsFileSystem);
         }
 
-        protected override IRepositoryCachePolicy<ITemplate, int> CreateCachePolicy()
-        {
-            return new FullDataSetRepositoryCachePolicy<ITemplate, int>(GlobalIsolatedCache, ScopeAccessor, GetEntityId, /*expires:*/ false);
-        }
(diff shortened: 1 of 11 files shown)


src/Umbraco.Infrastructure/Persistence/Repositories/Implement/TemplateRepository.cs
```diff
@@ -449,9 +458,28 @@ public Stream GetFileContentStream(string filepath)
...
+        public long GetFileSize(string filename)
+        {
+            if (GetFileSystem(filename).FileExists(filename) == false)
+            {
+                return -1;
+            }
+
+            try
+            {
+                return GetFileSystem(filename).GetSize(filename);
```
nikolajlauridsen, line 472: Small thing, but is there a reason to call GetFileSystem twice? Instead of saving it in a variable once, and using that twice?

nikolajlauridsen (approved):
Looks good to me, found one small odd thing though. I also tested it out using the snippet you provided and it works 👍

src/Umbraco.Infrastructure/Persistence/Repositories/Implement/TemplateRepository.cs
AndyButland: No, this was just a copy and paste, but I'll amend, as you're right, there's no reason.

PR #10267, opened 2021-05-13
Add option to remove/cancel added crops

bjarnef:
### Prerequisites

### Description
With the new Media Picker v3 it is possible the added crops in the configuration. However if added multiple and where some of these "rows" are empty it wasn't possible the save the datatype, because of the validation.
So you would either have to fill in some dummy data temporary or refresh page, but could then potentially loose already defined crops.

These changes add a option to remove/cancel a crop row.

zibse1HLQV

@madsrasmussen @nielslyngsoe


nathanwoulfe (approved):
Nice work @bjarnef, as usual :)

This is a sensible addition - users should never need to save dummy data, should always be able to cancel/revert an action

PR #10272, opened 2021-05-16
Rename models builder modes

callumbwhyte:
I appreciate this is a rather controversial one, and I'm happy to discuss some alternative options!

This pull request streamlines the names of the various Models Builder modes available in V9.

As discussed with Bjarke, it doesn't feel right having a Models Builder mode named "AppData" in V9 where models are no longer generated into the `App_Data` folder - they are now generated in `/umbraco/models/*`.

Considering what to rename this mode to forced me to think about the wider Models Builder config options as a whole. I landed on:

- PureLive => Runtime
- AppData => Code
- LiveAppData => LiveCode

I have always felt the term "PureLive" was confusing to new users and didn't make much sense. I opted for "Runtime" because this should hopefully better resonate with developers as meaning generated on-the-fly, which I feel accurately reflects what this mode does (generating invisibly in memory on app-start and when saving content types).

I chose "Code" because the output is generated C# code files. As the location of these files can be changed via configuration I feel it's important to avoid calling it anything related to where the output is stored. The name doesn't specify how the models ultimately end up being used in someone's project or when / if they might be compiled.

I kept the "Live" convention for the incarnation of "Code" mode that generates changes on the fly.

I've separated each rename as it's own commit, so this can be easily cherrypicked if desired.

As mentioned, happy to hear some alternatives - I just feel these are clearest 😊


marcemarc:
Yes, have always thought PureLive was weird...

with these three options, the naming is difficult because there is a commonality between the type of the output on two options and a  different commonality between the trigger on two other options, with these kinds of naming dilemmas then it can be helpful to always have a two-word combination

eg

Word1 - Word2

Where 'Word1' describes the action of the option and 'Word2' describes the output.

So it's easy to see contrast the difference between the three options... and their similarities.

But of course, deciding on the words is the hard bit...

Not sure about 'Live' because it seems to imply a Live environment when really we're describing the fact that the change to the models output happens automatically? (even in development environments)

So maybe you could say the difference in the Trigger could be described as Automatic vs Manual, eg whether the change will just happen automatically or require the pressing of the button...

then the output? is it 'In Memory?' vs 'Code'? - or is it 'when' the generation occurs eg Runtime vs CompileTime ? that is more helpful?

so you'd have

Auto-InMemory
Manual-Code
Auto-Code

or

Auto-Runtime
Manual-CompileTime
Auto-CompileTime

neither of which sound fantastic! - but that was the gist of the thinking!

c9mbundy:
@marcemarc ...or perhaps the other way around?

Runtime-Auto
CompileTime-Manual
CompileTime-Auto

marcemarc:
> Runtime-Auto
> CompileTime-Manual
> CompileTime-Auto

Yes, Mike,  that is a better example!, I think,  because the output is more 'important' than the 'trigger action' in the description of and therefore understanding of how the mode operates...

Although I've now gone down a tangent of whether Compiletime is one word with a lower T or two words Compile Time and therefore upper case T, and Runtime is definitely one word, or is it :-P

callumbwhyte:
Just to add here, I do feel there's a difference between how the modes are referred to in Core vs Configuration as in theory only the latter is public.

I don't think we should get hung up on the semantics of if / where dashes come 😆

The options available are defined via the `ModelsMode` enum.

If a user configures Models Builder in **code**, then they have intelisense on the names of each option.

If a user configures Models Builder in **config**, then the `GenericEnumConverter` is used to bind a string representation of the enum value to the config class.

I haven't looked extensively but OOTB I don't think we can bind anything other than alpha-numeric PascalCase strings.

marcemarc:
> I haven't looked extensively but OOTB I don't think we can bind anything other than alpha-numeric PascalCase strings, and therefore dashes couldn't be used anyway.

RuntimeAuto
CompileTimeManual
CompileTimeAuto

Shazwazza:
My 2c is that we shouldn't use the term 'Compile' because that is not what is happening. The compiling is done by your VS solution, models builder is just 'Generating' so IMO the term "Generated" is more accurate.

nathanwoulfe:
Generated also resolves @marcemarc's dilemma around capitalised T in CompileTime 🙄

marcemarc:
> 'Generating' so IMO the term "Generated"

did think about generated... does it distinguish enough between the three modes? eg on some level doesn't it feel like in PureLive mode, that the models are still being generated? or would there be another word, the opposite of generated that would describe the pure live situation well?

All I'm really proposing is two words combined, one that describes the action, and one that describes the output, so it gives a hint on how it operates in that mode and what you get! - so every mention of the modes doesn't have to be followed with a 2 line description of what that means... but agree 'compile' and 'compiletime' are not necessarily accurate!

Is SourceCode more accurate?

InMemoryAuto
SourceCodeManual
SourceCodeAuto

filipbech:
Hi everyone
Good discussions here. We would like to get this change into a beta soon, and we are leaning towards going with @marcemarc latest suggestions: 

> InMemoryAuto
> SourceCodeManual
> SourceCodeAuto

Speak now or forever hold your peace... (final call tomorrow morning)

ronaldbarendse:
@filipbech Totally agree on the mode names, but wouldn't it make more sense to split the type of generation (`InMemory` or `SourceCode`) and when it's triggered (`Auto` or `Manual`)? This would mean you can then also manually trigger the in-memory generation/recompile, which might avoid a lot of unnecessary work/restarts/slowness when doing multiple edits to document types etc.

Code wise it would mean the trigger could be implemented the same for all modes, so a lot of code/if-statements/switches can be simplified.

bergmania:
> @filipbech Totally agree on the mode names, but wouldn't it make more sense to split the type of generation (`InMemory` or `SourceCode`) and when it's triggered (`Auto` or `Manual`)? This would mean you can then also manually trigger the in-memory generation/recompile, which might avoid a lot of unnecessary work/restarts/slowness when doing multiple edits to document types etc.
> 
> Code wise it would mean the trigger could be implemented the same for all modes, so a lot of code/if-statements/switches can be simplified.

We also have a `Nothing` mode that disables ModelsBuilder completely.  I fear a separation of the concepts will make stuff more complex when we have to support additional modes (InMemoryManual). 
The number of possible configurations is already one of the things that make testing complex..

PR #10275, opened 2021-05-17
Updated serialization defaults for ArtifactBase

AndyButland:
In this PR I've removed the `[DataContract]` attribute added on `ArtifactBase` in the migration from V8 to V9, which has the effect of making all properties on derived artifacts opt-in.

And then handled the omission of the checksum value from the serialized string without taking a Json.Net dependency.

With this included and the Deploy dependency on CMS updated, that should resolve the serialization issues that would otherwise need all properties marked to opt them into serialization, and revert to the V8 behaviour where this isn't necessary.

Have verified with a test that the previous behaviour and the new with regards serialization is the same.

### DIFF
--- a/src/Umbraco.Core/Deploy/ArtifactBase.cs
+++ b/src/Umbraco.Core/Deploy/ArtifactBase.cs
@@ -1,22 +1,18 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.Serialization;
 
 namespace Umbraco.Cms.Core.Deploy
 {
     /// <summary>
     /// Provides a base class to all artifacts.
     /// </summary>
-    [DataContract]
     public abstract class ArtifactBase<TUdi> : IArtifact
         where TUdi : Udi
     {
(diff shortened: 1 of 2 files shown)


src/Umbraco.Core/Deploy/ArtifactBase.cs
```diff
@@ -1,61 +1,50 @@
...
         public TUdi Udi { get; set; }
 
-        [IgnoreDataMember]
-        public string Checksum
-        {
-            get { return _checksum.Value; }
-        }
+        public string Checksum => _checksum.Value;
+
+        public bool ShouldSerializeChecksum() => false;
```
bergmania, line 36: We need a comment on this one pointing to some kind of docs, like https://www.newtonsoft.com/json/help/html/ConditionalProperties.htm

I think this is the only place in the entire codebase we use this functionality

AndyButland: I've added a comment referencing this.

bergmania (approved):

Issue #10274, opened 2021-05-17
Variants dialog crashes when publishing (segment + no culture)


PerplexDaniel:
## Umbraco version

I am seeing this issue on Umbraco version: 8.13.0

### Bug summary

If a page has multiple variants that are varied only by segment (**not** by culture), the save/publish/unpublish dialog will crash when the user tries to perform a Save/Publish/Unpublish action. 

The dialog will not render any variants and the console shows an error, see images below.

This was caused by the changes of PR #9797 which improves the sorting of variants in the specified dialog, specially in the [`sortVariants`](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/services/contenteditinghelper.service.js#L790) and [`getSortedVariantsAndSegments`](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/services/contenteditinghelper.service.js#L814) functions.
These functions directly access `variant.language.{culture,isDefault,isMandatory}` but do not account for the case where `variant.language` is simply `null`. That happens when culture variation is not enabled. So this will crash every time when only segment variations are enabled and some segment variants are created.

So the fix for this is to account for language being null and handle accordingly. That is, something like 
`a.language.isDefault` 
should probably just be
`a.language != null && a.language.isDefault`

![image](https://user-images.githubusercontent.com/24297947/118468436-0f0a8800-b705-11eb-9ad7-a41f17240eee.png)
![image](https://user-images.githubusercontent.com/24297947/118468524-25b0df00-b705-11eb-9699-a0baa68d9d54.png)

Language is NULL:
![image](https://user-images.githubusercontent.com/24297947/118469666-34e45c80-b706-11eb-9f00-59abac3394cd.png)

### Steps to reproduce

- Add some variants to a node that have segmentation enabled but not culture. 
  - Instructions are available in #9797 at the "How and what to test:" section. 
- Try to save/publish/unpublish one or more variants in Umbraco.

### Expected result

The overlay shows all variants and they can be saved/published/unpublished without issues.

### Actual result

The overlay is empty because the JavaScript crashes, we cannot save/publish/unpublish anything.

---
_This item has been added to our backlog [AB#11985](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/11985)_

nul800sebastiaan:
[...]

PR #10278, opened 2021-05-17
#10274 Variant sorting should take into account that there might not be any language

nielslyngsoe:
Ensures that language is available before testing on the properties of it.

test-notes etc can be found on issue:

Fixes https://github.com/umbraco/Umbraco-CMS/issues/10274

Follow the instructions in this PR to get segments set up: #7973


madsrasmussen:
I have tested the following dialogs with only segments allowed:
* Variant picker
* Save
* Save And Publish
* Schedule
* Unpublish
* Publish with descendants

The same dialogs has also been tested with both languages and segments allowed.

PR #10279, opened 2021-05-18
V9: Resolve virtual view paths from DataEditorAttribute in DataValueEditor

### LINKED ISSUE #10265: NetCore: DataValueEditor.View no longer resolves virtual path
When developing a custom property-editor using the `DataEditorAttribute` for v9.0.0-beta001, if the `View` path is virtual, (e.g. prefixed with `~/`) it is not resolved.

In my Contentment package, many of the property-editors use the `DataEditorAttribute`, an example here: <https://github.com/leekelleher/umbraco-contentment/blob/1.4.2/src/Umbraco.Community.Contentment/DataEditors/CodeEditor/CodeEditorDataEditor.cs#L16>

> Please ignore that I'm using string constants, the value would be `"~/App_Plugins/Contentment/editors/code-editor.html"`.
> For an example from the Umbraco documentation, see here: <https://our.umbraco.com/documentation/Tutorials/Creating-a-Property-Editor/#setting-up-a-property-editor-with-c>

When I used this editor on a DocumentType, I get the following error in the back-office...

> **Request error: The URL returned a 404 (not found):**
> **`~/App_Plugins/Contentment/editors/code-editor.html`**

In v8, the `View` path is resolved using the `IOHelper.ResolveVirtualUrl` helper method.

<https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/PropertyEditors/DataValueEditor.cs#L78>

For v9.0.0-beta001, that has been removed.

<https://github.com/umbraco/Umbraco-CMS/blob/release-9.0.0-beta001/src/Umbraco.Core/PropertyEditors/DataValueEditor.cs#L92>

---

I'm aware in most cases that I should be injecting the `IIOHelper` dependency into my property-editor, but I'm unsure how that would work with the `DataEditorAttribute`?

Would we considered this to be a regression bug?

Reproduction
------------

### Bug summary

_(At this level, I kinda feel that I'm teaching you to suck eggs here?)_ 🤷 

- Create a custom property-editor using the C# approach: see <https://our.umbraco.com/documentation/Tutorials/Creating-a-Property-Editor/#setting-up-a-property-editor-with-c>
- For the `View` property, use a path that starts with `"~/"`.
- Compile/build/whatever.
- Create the DataType.
- Add new property to DocumentType - **At this point, the error will occur.**
- Create/open a content node (of the above DocumentType) - **you'll get the error here too.**

### Specifics

- Umbraco version: v9.0.0-beta001
- Browser: Chrome vLatest
- I haven't checked on any other browser.
- Error message outlined above, e.g. "Request error: The URL returned a 404 (not found)".

### Steps to reproduce

_(Outlined above)_

### Expected result

I'd expect the relative/virtual `View` path to have been resolved, as it did in v8.

### Actual result

_(Outlined above)_

bergmania:
Fixes #10265

- Resolve virtual view paths from DataEditorAttribute in DataValueEditor 

- Introduced an IDataValueEditorFactory, so we don't need to inject nested dependencies every time we introduce a new dependency in DataValueEditor..


src/Umbraco.Core/Services/IPropertyValidationService.cs
```diff
@@ -25,5 +25,9 @@ public interface IPropertyValidationService
             string validationRegExp,
             string isRequiredMessage,
             string validationRegExpMessage);
+
+        public IEnumerable<ValidationResult> ValidatePropertyValue(
```
AndyButland, line 29:
```suggestion
        IEnumerable<ValidationResult> ValidatePropertyValue(
```

AndyButland: Perhaps add an XML header comment to this and the other method missing it in this interface too.

src/Umbraco.Infrastructure/Manifest/ManifestParser.cs
```diff
@@ -53,13 +54,14 @@ public class ManifestParser : IManifestParser
             ILocalizationService localizationService,
             IJsonSerializer jsonSerializer,
             ILocalizedTextService localizedTextService,
-            IShortStringHelper shortStringHelper)
+            IShortStringHelper shortStringHelper,
```
AndyButland, line 57: I don't think you need to inject `ILoggerFactory`, `IDataTypeService`, or `ILocalizationService`  in here now.

AndyButland (approved):
Looks a good update to me - much cleaner, and I suspect a few package developers will appreciate this as it makes creating property editors a bit less cumbersome with all the dependencies that otherwise need to be provided in the constructor.

Have tested locally by creating and amending datatypes, adding them to doc types and using them - and all seems fine.

Just a few minor comments inline.

PR #10283, opened 2021-05-18
Make dashboards support deep linking

madsrasmussen:
This PR makes dashboards support deep linking. It adds the alias as a query param when switching between dashboards and loads the correct dashboard on page load.

Screenshot 2021-05-18 at 14 25 25

**How to test:**
* Make sure you can change dashboards in each section.
* Make sure the query params get updated with the dashboard alias
* Make sure you can link directly to a dashboard
* Make sure the first dashboard is active if you don't have a dashboard query param or type in a random alias


nathanwoulfe:
@madsrasmussen I tried changing the base branch to v8/contrib so I can merge, but that picked up an additional commit (so I flipped back to v8/dev). Reckon you could change the base and address the additional commit?

This is a great addition, it's something I've tried to hack around in packages so will be brilliant to have in core.

src/Umbraco.Web.UI.Client/src/views/common/dashboard.controller.js
```diff
@@ -20,23 +22,38 @@ function DashboardController($scope, $routeParams, dashboardResource, localizati
...
             tab.active = false;
         });
+
         tab.active = true;
+
+        $location.search(DASHBOARD_QUERY_PARAM, tab.alias);
     };
 
+    function initActiveTab () {
+        if (!$scope.dashboard.tabs) {
```
nielslyngsoe, line 41: Before there was a check for $scope.dashboard.tabs.length > 0, I think we need to add that again then. to avoid eventual issue on line 51.

madsrasmussen: Good catch. I have updated the PR.

src/Umbraco.Web.UI.Client/src/views/common/dashboard.controller.js
```diff
@@ -34,9 +35,25 @@ function DashboardController($scope, $routeParams, dashboardResource, localizati
         $scope.dashboard.tabs.forEach(function(tab) {
             tab.active = false;
         });
```
ronaldbarendse, line 37: This should also check `$scope.dashboard.tabs` before calling `forEach` on it (as it might still be undefined):
```suggestion
        if ($scope.dashboard.tabs && $scope.dashboard.tabs.length > 0) {
            $scope.dashboard.tabs.forEach(function(tab) {
                tab.active = false;
            });
        }
```

src/Umbraco.Web.UI.Client/src/views/common/dashboard.controller.js
```diff
@@ -34,9 +35,25 @@ function DashboardController($scope, $routeParams, dashboardResource, localizati
...
         tab.active = true;
+
+        $location.search(DASHBOARD_QUERY_PARAM, tab.alias);
     };
 
+    function initActiveTab () {
+        // check the query param for a dashboard alias
+        const dashboardAlias = $location.search()[DASHBOARD_QUERY_PARAM];
+        const dashboardIndex = $scope.dashboard.tabs.findIndex(tab => tab.alias === dashboardAlias);
+        // set the first dashboard to active if there is no query param of we can't find a matching dashboard for the alias
```
ronaldbarendse, line 48:
```suggestion
        // set the first dashboard to active if there is no query parameter or we can't find a matching dashboard for the alias
```

src/Umbraco.Web.UI.Client/src/views/common/dashboard.controller.js
```diff
@@ -21,10 +23,9 @@ function DashboardController($scope, $routeParams, dashboardResource, localizati
...
-        
-        // set first tab to active
-        if($scope.dashboard.tabs && $scope.dashboard.tabs.length > 0) {
-            $scope.dashboard.tabs[0].active = true;
+
+        if ($scope.dashboard.tabs && $scope.dashboard.tabs.length > 0) {
+            initActiveTab();
         }
 
         $scope.page.loading = false;
```
ronaldbarendse, line 31: This _might_ remove the loading indicator before the localization service promise has succeeded. It's best to add/push all promises to an array and wait for all to succeed:

```js
// Add $q to the controllers constructor!

var promises = [];

promises.push(localizationService.localize("sections_" + $routeParams.section).then(function(name) {
    $scope.dashboard.name = name;
}));

// Same for dashboardResource.getDashboard

$q.all(promises).then(function () {
    $scope.page.loading = false;
}
```

If the un-localized name is not hidden while the page is loading, the indicator/overlay might need to be moved to do so (otherwise you could end up seeing the `[dashboard_name]` format). This is something I've also used/fixed in this PR: https://github.com/umbraco/Umbraco-CMS/pull/8384/files#diff-c9f6a6eb26104b4a81f7e06b9c6f2cdaf563a0854822413f1ca2083a15c6a066R73.

nielslyngsoe:
Cool, tested and it works well. Even if you are not logged in you are still getting to the right dashboard when logging in.

PR #10290, opened 2021-05-19
Adds an extension method allowing checks for whether a notification handler is registered for a given notification.

AndyButland:
We appear to have a use case for this in Forms, where we have different code paths for pre-populating a form for whether or not custom code is registered with the "pre-populate form" event.

Hence considering this as a method of checking if a particular notification has one or more handlers.


AndyButland:
Closing this as after discussion doesn't seem necessary.

PR #10299, opened 2021-05-20
V9: Allow custom notification publisher in scope

nikolajlauridsen:
Allows a custom `IScopedNotificationPublisher` to be injected into a scope using the `IScopeProvider.CreateScope` method.

The logic is very much the same as for the `IEventDispatcher`, not allowing you to set it anywhere but the outermost scope, I know we talked about allowing a custom publisher in inner scopes as well for a NoOpPublisher, to suppress events, but this requires some more changes on how we handle the notifications, and I think that should be a separate task. This change shouldn't compromise our ability to do that later though.

We also talked about removing the `IEventDispatcher`, but thinking more about it, I'm not sure that's the best idea, wouldn't it make it quite a bit harder to port over custom code where you might still use (and want to use) the old event pattern for your own events?

I added a couple of unit tests as well to test that it is in fact the injected `IScopedNotificationPublisher` that is used for notifications.

I think this is a breaking change though if you happen to have used `CreateScope` with all the arguments as positional arguments


bergmania:
Looks good 💪

PR #10302, opened 2021-05-21
Fix getBlueprintScaffold endpoint

### LINKED ISSUE #10300: (v9-beta002) Using content templates does not work

(v9-beta002)

### Bug summary

After creating a Content template in the backoffice, and trying to create a new content node based on the content template, it fails and throws a 404:

_Request error: The URL returned a 404 (not found):
/umbraco/backoffice/umbracoapi/content/GetEmpty_ 

And the node is not created and the content section keeps loading.

### Specifics

Looking in the browser console, it is possible to see an angular error:
image
****

### Steps to reproduce

1. Follow documentation here: https://our.umbraco.com/documentation/Getting-Started/Backoffice/Content-Templates/ And create a content template

2. Try and create a node based on the content template.

3.  Click save and it fails with an 404 error message.

Expected a new content node based on my content template to be created.

Instead, it ends up failing.

nathanwoulfe:
### Prerequisites

If there's an existing issue for this PR then this fixes #10300 

### Description
Method name in ContentController was updated to GetEmptyBlueprint, reference in content.resource.js was updated too, but looks to have reverted at some point since, because managing merges across two diverging branches must be lots of fun.

Change here fixes the reference in the javascript - `GetEmpty` becomes `GetEmptyBlueprint`

To test, create a content blueprint, then create content using that blueprint - should see no errors and new content is created as expected, using the template.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/resources/content.resource.js
+++ b/src/Umbraco.Web.UI.Client/src/common/resources/content.resource.js
@@ -694,7 +694,7 @@ function contentResource($q, $http, umbDataFormatter, umbRequestHelper) {
                 $http.get(
                     umbRequestHelper.getApiUrl(
                         "contentApiBaseUrl",
-                        "GetEmpty",
+                        "GetEmptyBlueprint",
                         { blueprintId: blueprintId, parentId: parentId })),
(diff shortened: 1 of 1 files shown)


bergmania:
Thanks @nathanwoulfe ..

PR #10303, opened 2021-05-21
fixes #9894 . Truncate block editor model title to 100 chars.

### LINKED ISSUE #9894: Block List editor fails submit if heading exceeds 255 characters
In Umbraco 8.11.1 I noticed an issue with Block List editor with label expression.

For example a Text Block with a property `bodyText` and a label expression `{{bodyText}}` and the text including HTML characters increase 255 characters. Or an label expression like this: `{{bodyText | ncRichText | truncate:true:256}}`.

Furthermore I noticed if an existing block had a longer text and the aforementioned label expression, you also can't removing text because the submit fails.

However if using a truncated label expression to 255 characters `{{bodyText | ncRichText | truncate:true:256}}`, or no label expression or a custom view the submit works.

skartknet:
The problem on #9894 is that the `umb-editor-header` used to display the block title has a maxlength of 255. I have truncated the value passed from the Block Editor to the directive. I have chosen 100 chars and it looks good to me.


bjarnef:
I would probably add this logic in controller instead.

skartknet:
Yep, I guess the ideal solution would be to truncate that title when it's saved or set up on the main model, but I can/t find how and where that title set up first using the configured expression. :(

nathanwoulfe:
Hi @skartknet and @bjarnef, trying to work through some older PRs...

I've updated the proposed solution to pass the max length in as a property on the umb-editor-header directive, so that it can be configured in other contexts too. I've set the length to 255 as that matches the max length permitted when the name field is unlocked (in umb-editor-header), so keeps it consistent.

Haven't merged, but commit has been added to this PR. If we're happy with the update, I'll merge.

skartknet:
Thanks Nathan, looks good to me.

nathanwoulfe (approved):
Added a null-check before attempting to access `scope.name.length`, just in case it's empty.

PR #10310, opened 2021-05-21
Add a collection to allow adding CSS and JS to the back office via C#

### LINKED ISSUE #10293: Allow Css / JS assets to be added via c# code in addition to package.manifest
### Umbraco version

v9

### Description

Currently when adding a section/dashboard etc to the back office, if you wish to add any custom CSS or JS you need to add a package.manifest file to include them.

However for Property editors there is the `PropertyEdtiorAsset` attribute which allows this.

Looking at the implementation `IRuntimeMinifier` `SmidgeRuntimeMinifier `  with the help of  @soreng we noticed that Smidges `IBundleManager` has an AddToBundle function

Exposing this via the interface would allow us to inject IRuntimeMinifier or access it from something similar to how you add sections currently.

Thanks
Matt

Matthew-Wise:
### Prerequisites

If there's an existing issue for this PR then this fixes #10293

### Description

This allows the bundle to be added to using C# from a composer or an extension on IUmbracoBuilder. An example of how this code would be used by a developer.

```
// injected or extended
// IUmbracoBuilder builder;

builder.BackOfficeAssets()
.Append<MyCustomJsFile>()
.Append<MyCustomCssFile>();


public class MyCustomJsFile : JavaScriptFile
{
    public MyCustomJsFile() : base("/abc.js")
    {

    }
}

public class MyCustomCssFile : CssFile
    {
        public MyCustomJsFile() : base("/my-styles.css")
        {
        }
    }
```

The files in the example are within wwwroot

Thanks
Matt

### DIFF
--- a/src/Umbraco.Core/DependencyInjection/UmbracoBuilder.Collections.cs
+++ b/src/Umbraco.Core/DependencyInjection/UmbracoBuilder.Collections.cs
@@ -16,6 +16,7 @@ using Umbraco.Cms.Core.Sections;
 using Umbraco.Cms.Core.Strings;
 using Umbraco.Cms.Core.Tour;
 using Umbraco.Cms.Core.Trees;
+using Umbraco.Cms.Core.WebAssets;
 using Umbraco.Extensions;
 
 namespace Umbraco.Cms.Core.DependencyInjection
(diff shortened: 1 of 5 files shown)


bergmania:
Hi @Matthew-Wise, it looks like you need to call your extension method BackOfficeAssets, so the collection is registrated

Matthew-Wise:
Missed that one :) updated.

bergmania (approved):
Looks good to me :) Thanks 💪

Matthew-Wise:
@bergmania should PropertyEditorAssetAttribute be marked as obsolete? With this as the replacement?

bergmania:
Hi @Matthew-Wise , looked like I only closed the PR not merging it 🤦 ..

Can you recreate the branch, so we can reopen the PR, and merge.. 

Regarding the `PropertyEditorAssetAttribute`, it would make sense to mark it as obsolete at point to this new way of doing stuff.

Matthew-Wise:
Restored :) Ill add the obselete attribute now a well.

bergmania:
🚀

PR #10313, opened 2021-05-23
Link picker: Make link input full width if anchor option is disabled

BatJan:
### Prerequisites

### Description
I noticed that the "link" input was only 50% wide when the "Anchor option" has been disabled in the configuration for the multi url picker datatype, which I address with this PR so it's 100% wide if the anchor option is disabled. If the anchor option is not disabled both fields are still 50% wide as usual 👍🏻 

**Before**
link-before

**After**
link-after


nathanwoulfe (approved):
Hi @BatJan hope you don't mind but I changed the implementation here - rather than adding a new class, we can set the width when the URL input is the only child. Same result, just a smaller change to make it happen.

BatJan:
Yeah no worries mate 😉

PR #10314, opened 2021-05-23
NETCore: Fix dotnet new template for safe namespaces

warrenbuckley:
Fix for invalid chars used in dotnet new templates creations to ensure namespace is valid & compiles

## Testing
* Try the dotnet new template pack before this PR with a name `Umbraco-v9-2021-05-12`
* `dotnet new umbraco --name Umbraco-v9-2021-05-12`
* This will fail to build due to invalid namespace if you do a `dotnet build`
* Install/update new dotnet template pack and retry with the same command & verify it makes the namespace safe & builds
* Retry with various other characters that could make unsafe namespace, perhaps with spaces
* `dotnet new umbraco --name "Umbraco v9 2021 05 12"`

### DIFF
--- a/build/templates/UmbracoProject/.template.config/template.json
+++ b/build/templates/UmbracoProject/.template.config/template.json
@@ -29,7 +29,7 @@
             "continueOnError": true
         }
     ],
-    "sourceName": "UmbracoProject",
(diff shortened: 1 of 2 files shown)


bergmania:
Works as expected 💪

Issue #10295, opened 2021-05-20
Uploading SVG results in a "file"


soreng:
### Which Umbraco version are you using?

v9-beta002

### Bug summary

When uploading an SVG to the media lib, it creates a media of type "File", instead of the "Vector Graphics (SVG)" that is created by the installer.

### Steps to reproduce

Upload https://upload.wikimedia.org/wikipedia/commons/4/4f/SVG_Logo.svg to the media section

### Expected result / actual result

The resulting media is of type "File", expected it to be of type "Vector Graphics (SVG)"

nul800sebastiaan:
Thanks @soreng, looks like we're a bit out of sync between the v8 and v9 branch, doing a merge should fix this and https://github.com/umbraco/Umbraco-CMS/issues/10289 as well. 🤞

bergmania:
I can repro, even after the last merge.. Ill look into it..

warrenbuckley:
Fixed in this PR for V9
https://github.com/umbraco/Umbraco-CMS/pull/10320

PR #10320, opened 2021-05-25
V9: Fix issue with determining the uploaded media type

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/10295


warrenbuckley:
Code change fine and works as expected.

PR #10325, opened 2021-05-25
added windowResizeListener

nielslyngsoe:
Added missing dependency, to fix js issue that its undefined.


sniffdk:
This seems to still be a problem on 8.14.1 when opening a crop on a selected image in the new media picker.
image

The offending code is here:
image

But it seems that there is still a problem injecting the dependency?

PR #10329, opened 2021-05-26
fixes #10297 8.14-RC - A few areas where focus lock doesn't work prop…

### LINKED ISSUE #10297: 8.14-RC - A few areas where focus lock doesn't work properly

8.14.0-rc

### Bug summary

When performing some actions in the backoffice you get stuck, you can not click anywhere and have to refresh the whole backoffice. 

### Specifics

_No response_

### Steps to reproduce

The identified features are:

- Adding an image to the grid allows you to select an image but then you can't do anything else after that
- Using the "Insert" button in Template, Views, Partial Views and Partial View Macros
- Updating granular permissions on user groups

A known workaround for now is to use keyboard shortcuts (CTRL+S) to save what you're working on but can't use the backoffice any more, then at least you don't lose your progress.

_No response_

---
_This item has been added to our backlog [AB#12067](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/12067)_

madsrasmussen:
In all cases mentioned in the issue (https://github.com/umbraco/Umbraco-CMS/issues/10297), two infinite editors are closed at the same time which results in "editors.length" being counted wrong because of the slide animation.

We need to keep another count of open editors because the length of the editors' array is first changed when the animations are done.

How to test:
* Test the scenarios from the original issue and make sure you can use the interface when the infinite editors are closed.

### DIFF
--- a/src/SolutionInfo.cs
+++ b/src/SolutionInfo.cs
@@ -18,5 +18,5 @@ using System.Resources;
 [assembly: AssemblyVersion("8.0.0")]
 
 // these are FYI and changed automatically
-[assembly: AssemblyFileVersion("8.13.0")]
-[assembly: AssemblyInformationalVersion("8.13.0-rc")]
+[assembly: AssemblyFileVersion("8.14.0")]
+[assembly: AssemblyInformationalVersion("8.14.0-rc")]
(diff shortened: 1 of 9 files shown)


nielslyngsoe:
Close in favor for https://github.com/umbraco/Umbraco-CMS/pull/10331

Issue #10308, opened 2021-05-21
8.14-RC Switching Media Picker 2 and Media Picker 3 looses data


konius:
### Which Umbraco version are you using?

8.14.0-rc

### Bug summary

Switching Media Picker 2 to Media Picker 3 looses the current image selection due to data schema differences.

![Screenshot 2021-05-21 153450](https://user-images.githubusercontent.com/3933562/119138805-f4107e80-ba4a-11eb-8af3-9ab1db564bb9.jpg)

### Steps to reproduce

1. Create property for Media Picker 2 and pick media items (can be skipped if upgrading)
2. Change property data type to Media Picker 3
3. Navigate back to node with property 
4. Selection is gone

### Expected result / actual result

I was expecting the currently stored data (comma separated UDI's) to be transformed to the new array of objects with keys.

nul800sebastiaan:
Thanks @konius, unfortunately, like with many other datatypes, we don't support switching between datatypes that need to have a different schema stored. 
Specifically, we don't want to make this a migration either since we don't know the size of the sites we'll be dealing with and how much data is stored in it. So for performance reasons we don't want to do this automatically when upgrading. For now, you would be able to write a migration yourself to find all existing nodes with stored data and convert them to the new format, after which you can switch the data type on the related doctypes.

The other thing I can think of is doing a conversion when each media picker is loaded in Umbraco, the datatype would take the invalid format, update it and show you the picked items, then you can progressively save the new data. This comes with the handicap that you will need to test for this in your razor templates, so you would have to implement two code paths in your template.

In any case, we don't plan to work on this for the initial release at the moment, so for now that would have to be a custom converter.

PR #10334, opened 2021-05-26
Shows a warning for people tempted to switch from the Obsolete to the new MediaPicker(v3)

nul800sebastiaan:
This is in relation to #10308, please read that first.

Which looks like this: 

image

### DIFF
--- a/src/Umbraco.Web/PropertyEditors/MediaPickerConfiguration.cs
+++ b/src/Umbraco.Web/PropertyEditors/MediaPickerConfiguration.cs
@@ -8,6 +8,9 @@ namespace Umbraco.Web.PropertyEditors
     /// </summary>
     public class MediaPickerConfiguration : IIgnoreUserStartNodesConfig
     {
+        [ConfigurationField("notice", "You can NOT change the property editor", "obsoletemediapickernotice")]
+        public bool Notice { get; set; }
+
         [ConfigurationField("multiPicker", "Pick multiple items", "boolean")]
         public bool Multiple { get; set; }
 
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Relates to: #10308

src/Umbraco.Web/PropertyEditors/MediaPickerConfiguration.cs
```diff
@@ -8,6 +8,9 @@ namespace Umbraco.Web.PropertyEditors
     /// </summary>
     public class MediaPickerConfiguration : IIgnoreUserStartNodesConfig
     {
+        [ConfigurationField("notice", "You can NOT change the property editor", "obsoletemediapickernotice")]
+        public bool Notice { get; set; }
+
```
bjarnef, line 13: @nul800sebastiaan I think this should be defined as `string`.

nul800sebastiaan: sure but it doesn't really matter, it's not a REAL prevalue, just a hack to show a warning for now while we think of a betters solution. 😄

skttl:
I see the good intention, but isn't this the case for most data editors already?

nathanwoulfe:
I think it's valuable since previous obsolete editors can be swapped out - ContentPicker for ContentPicker2, for example, works fine. 
Probably a good pattern to adopt for all obsolete/superseded patterns, so users can make changes knowing they're not going to lose 1500 images from their site...

nul800sebastiaan:
> I see the good intention, but isn't this the case for most data editors already?

@skttl yes of course, this is just a first step since we already KNOW that people will try it and get burned: https://github.com/umbraco/Umbraco-CMS/issues/10308

We need to expand this somehow but we don't have the bandwith before release.

PR #10342, opened 2021-05-27
adjusted the maxQueryString and maxQueryStringLength

### LINKED ISSUE #9695: Limitation of total of concatenated rolenames (Member groups)

There seems to be an issue with the maximum number of roles that can be added to a Page. This seems to be dependent of the total concatenated string of all the role names that are added to it.
No more roles can be added when the total role names exceed a length of about 1024 characters.
When there are about 15 roles of 90 characters no more roles can be added.

## Umbraco version

8.10.1 

Reproduction
------------
This can easy being tested.

1. Create a new Umbraco Demo Site (8.10.1)
2. Create 25 different Member groups. with a length of at least 100 characters
3. Add all of them via the Public Access to just one node, for example 'Home'.
4. The result is an error (see attachment)
5. When they only add 18 of them, no error will appear.

2021-01-20 13_29_27-Edit_ Home - Content - localhost

vlaellayug:
Adjust the value for the bug with the limitation of total of concatenated rolenames (Member groups)

This fixes #9695

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/resources/content.resource.js
+++ b/src/Umbraco.Web.UI.Client/src/common/resources/content.resource.js
@@ -1261,20 +1261,26 @@ function contentResource($q, $http, umbDataFormatter, umbRequestHelper) {
                 errorPageId: errorPageId
             };
             if (Utilities.isArray(groups) && groups.length) {
-                publicAccess.groups = groups;
+                return umbRequestHelper.resourcePromise(
+                    $http.post(
+                        umbRequestHelper.getApiUrl("contentApiBaseUrl", "PostPublicAccess", publicAccess),
+                        groups
+                    ),
+                    "Failed to update public access for content item with id " + contentId
+                );
             }
             else if (Utilities.isArray(usernames) && usernames.length) {
                 publicAccess.usernames = usernames;
+                return umbRequestHelper.resourcePromise(
+                    $http.post(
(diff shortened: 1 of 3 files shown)


nul800sebastiaan (changes requested):
Thanks @vlaellayug! I don't think this is the way to fix it, I don't understand why we don't do a `Post` request to this endpoint? A 'Post' can be much longer and is not limited by the querystring length. We can update this as a suggestion for users but since this the web.config and it's configurable, people will change this config and the original bug will appear again.

Can you look into how we can avoid putting these items in the querystring and make them part of a `Post` request instead please?

vlaellayug:
Hello @nul800sebastiaan,

I agree at some point it will appear again.

I've looked into it and it looks like it's being passed as queryString in the [updatePublicAccess](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/resources/content.resource.js#L1257) method.

That's why the parameter "groups" in [PostPublicAccess](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Editors/ContentController.cs#L2439) is using "[FromUri]"

Updated the [pull request](https://github.com/umbraco/Umbraco-CMS/pull/10342/commits/aacea29d9e523043b1ed79eddc73f45515c1d9d1). Also, the mentioned issue on #9695 is also replicable for "Specific members protection" too. If the query string for the "username" parameter exceeds the limit.

PSRCX0wF1e

If it's nice to have. I can update the code with this pull request.

PR #10343, opened 2021-05-27
inject clearAddUserForm() to inviteUser condition

### LINKED ISSUE #10280: Invite user in sequence does not clear the form
If you try to invite two users in a row using the "Invite another user" option, the form still shows the previously entered information.

I would expect this to be cleared instead. But not sure if this is also intended by HQ?

Seen in Umbraco 8.11 in this case.

vlaellayug:
Injected the clearAddUserForm().

This fixes #10280


nul800sebastiaan:
Oh, nice and simple, good one @vlaellayug, thanks! 👍

Issue #10340, opened 2021-05-27
Document type child nodes not showing custom icons


skttl:
### Which Umbraco version are you using?

8.13.0

### Bug summary

When using custom icons from /App_Plugins/*/backoffice/icons, they don't show up in the document type editor, when showing allowed child node types.

![image](https://user-images.githubusercontent.com/3726467/119814049-44278f00-beea-11eb-87a8-e185fa01e718.png)

### Steps to reproduce

Add some icons in App_Plugins/*/backoffice/icons, and select them on a doctype used as a child node for another document type.

Open the parent document type, and see the icon is missing on the child types.

### Expected result / actual result

Expected the icon to show, just like the default icons does.

bjarnef:
PR submitted: https://github.com/umbraco/Umbraco-CMS/pull/10344 🔥

nathanwoulfe:
Fixed in #10344

PR #10344, opened 2021-05-27
Support custom SVG icon in umb-child-selector component

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10340

### Description
When using custom SVG icons these were not show in `<umb-child-selector>` component.
Furthermore this PR also replace the search filter markup with `<umb-search-filter>` component.

image

image

The compass icon is a custom font icon from my [Material Design Icon Pack](https://our.umbraco.com/packages/backoffice-extensions/material-design-icon-pack/), the coffee icon a custom SVG icon and the rest Umbraco core icons.

image

---
_This item has been added to our backlog [AB#12146](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/12146)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/card.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/card.less
@@ -140,7 +140,7 @@
 
 .umb-card-grid .umb-card-grid-item {
     position: relative;
-	display: block;
+    display: block;
     width: 100%;
     height: 100%;
     padding: 10px 5px;
@@ -153,6 +153,28 @@
     align-items: center;
     justify-content: flex-start;
 
+    &__loading {
+        position: absolute;
+        background-color: rgba(255,255,255,0.8);
+        top: 0;
+        right: 0;
+        bottom: 0;
+        left: 0;
+    }
+
+    .umb-icon {
+        font-size: 30px;
+        line-height: 20px;
+        margin-top: 6px;
+        margin-bottom: 10px;
+        display: block;
+    }
+
+    a {
+        color: @ui-option-type;
+        text-decoration: none;
+    }
+
     > span {
         position: relative;
         display: flex;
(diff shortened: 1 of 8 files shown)


bjarnef:
Okay, I noticed a minor issue in list of datatypes / property editors, because items doesn't have the `umb-card-grid-item` class, but previous just used `.umb-card-grid i` in less.

image

Furthermore I noticed if using a custom SVG icon in a property editor, e.g. `icon-coffee-cup` this wasn't shown either.

bjarnef:
It now looks like this.

image

bjarnef:
I also noticed it is a similar issue if using property editor as macro parameter editor: https://our.umbraco.com/documentation/extending/macro-parameter-editors/

image

bjarnef:
A custom SVG icon was missing here as well - fixed in https://github.com/umbraco/Umbraco-CMS/pull/10344/commits/a54b0b1b0fb4a0d4e059eeb6185902fa94393737

image

bjarnef:
@nul800sebastiaan is it too late to fix this in 8.14.0 release?

nul800sebastiaan:
@bjarnef Awesome! But yes, about 2 hours too late for the 8.14 release, not to worry, we'll get it into a patch release instead 👍 Thanks for the help, you're a star!

bjarnef:
Oh no! 😱 ... no espresso for me then ☕
Hopefully in the next patch releases then 🩹

nathanwoulfe (approved):
Beautiful, thank you @bjarnef

It still bugs me that we pass the icon string into the umbIcon directive twice (once as the icon attribute, and again in the class). Probably not worth worrying about now, given it would be a pretty big change, but I'm sure we could be using the class attribute only for additional styling (color, size etc), then adding the icon attribute into the class string when we're rendering. Ideally the new UI will drop support for legacy icons, so the umbIcon component would be up for a rebuild anyway

PR #10347, opened 2021-05-28
Update keyboard.service.js

nathanwoulfe:
don't update default opts - missing this in merging #10023 10023


bjarnef:
@nathanwoulfe should this have been merged instead of closed?
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/services/keyboard.service.js#L266-L267

nathanwoulfe:
@bjarnef this is the same fix as I need to get into #10023 but if I create the PR I can't self-approve, so didn't want to be waiting for it to get picked up, hence my comment on #10023 asking if you could submit the PR, then I can approve and merge 😄

PR #10349, opened 2021-05-28
Update connect color in noUiSlider

bjarnef:
### Prerequisites

### Description
This PR updates the connect color in noUiSlider when this feature is used.
We may update this later, but for now I think it fits in UI better.

**Before**

image

**After**

image


nathanwoulfe (approved):
@bjarnef definitely an improvement - the teal-green color was a bit V7 😺

PR #10352, opened 2021-05-28
Enable email sender to publish an event if smtp is not configured

nikolajlauridsen:
In V8 we had an event you could listen to `EmailSender.SendEmail`, if this event had a handler registered, and `EmailSender` was constructed with enableEvents set to true, we would raise the event if the SMTP was not configured.

This PR reinstates this using notifications instead since it's used in deploy.

 Since EmailSender is now registered in DI, I created an overload for `SendAsync` instead which accepts the bool, looking through V8 I can see that this was only used for `UsersController.SendUserInviteEmailAsync`, so I've set enableNotifications to true there.

`CanSendRequiredEmail()` now also returns true if there is a notification handler for `SendEmailNotification` registered, same as V8.

### Testing 
To test this I first just ran the site normally and made sure that `CanSendRequiredEmail()` returned false, and that InviteUser was hidden.

Then I added this handler: 

```C#
public class EmailHandler : INotificationHandler<SendEmailNotification>
{
    private readonly ILogger<EmailHandler> _logger;

    public EmailHandler(ILogger<EmailHandler> logger)
    {
        _logger = logger;
    }
    
    public void Handle(SendEmailNotification notification)
    {
        _logger.LogInformation("Was requested to send email to {Reciever} body:{Body}", notification.Message.To,
            notification.Message.Body);
    }
}
```

And made sure that when I now run the site `CanSendRequiredEmail()` returns true, and I can invite a user, when I do the Handle method is invoked.

### DIFF
--- a/src/Umbraco.Core/Mail/IEmailSender.cs
+++ b/src/Umbraco.Core/Mail/IEmailSender.cs
@@ -10,6 +10,8 @@ namespace Umbraco.Cms.Core.Mail
     {
         Task SendAsync(EmailMessage message);
 
+        Task SendAsync(EmailMessage message, bool enableNotification);
+
         bool CanSendRequiredEmail();
     }
 }
--- a/src/Umbraco.Core/Mail/NotImplementedEmailSender.cs
+++ b/src/Umbraco.Core/Mail/NotImplementedEmailSender.cs
@@ -9,6 +9,10 @@ namespace Umbraco.Cms.Core.Mail
         public Task SendAsync(EmailMessage message)
             => throw new NotImplementedException("To send an Email ensure IEmailSender is implemented with a custom implementation");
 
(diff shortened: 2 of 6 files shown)


nikolajlauridsen:
One thing I also noticed however is that the event used to be internal, however since we would ideally not have internals be visible to deploy, I don't know if it's a good idea to make this notification internal? Maybe it would be better to rename it to something like `SendInviteNotification` to signal that you won't get a notification for all emails to be sent?

bergmania:
> Maybe it would be better to rename it to something like `SendInviteNotification` to convey that you won't get a notification for all emails to be sent?

IMO that sounds too specific..  I think it is okay as is..

bergmania (approved):

Issue #10354, opened 2021-05-28
Configure 404 error page per site using XPath 


kmeilander:
### Which Umbraco version are you using?

8.6

### Bug summary

When configuring my site's error404 page using XPath, I cannot use a $site to specify the page in relation to the requested site.  

### Steps to reproduce

On an instance with 2 sites with different domains, configure a 404 Error page in the umbracoSettings.config using an XPath using the $site tag.

Such as: 
 ```
   <errors>
      <error404>$site//error[@nodeName='404']</error404>
    ...
```

Make sure you have a page setup in each site at that XPath and then go to a non existent URL for each of the 2 domains.  

### Expected result / actual result

EXPECTED: For each site i see the corresponding error page.

ACTUAL: Both sites show the same error page, the first one in the content tree.

emma-hq:
Hey @kmeilander, 

Thanks for reporting this one and thanks also for the PR associated with it. We'll take a look at both and let you know our thoughts.

Emma

umbrabot:
Hi @kmeilander,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

kmeilander:
@emma-hq  thanks.  It's quite possible there's a good reason why my fix is a bad idea and shouldn't be done that way, but I couldn't think of anything.

Let me know if you need anything.

PR #10356, opened 2021-05-28
#10354 - added domain content id as node context for the content find…

kmeilander:
…er for the 404 error page so $site works

### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10354

### Description

I needed to make different 404 pages for different sites and so wanted to use $site in the xPath.  This doesn't currently work since the nodeContextId in the xpath isn't set.  

However, when we get the domain as part of the PublishedRequest, we also get the content id of the site for that domain so we can use this to pass in the site context to the xpath query and make $site work.   I think $current will also look at the site root.  

I opened a issue with more details around this.  https://github.com/umbraco/Umbraco-CMS/issues/10354 

To test this, setup multiple sites with multiple domains that have an error pages setup.  Then configure the error404 using xPath and $site.  See that each domain returns different nodes.

See the ticket for more details around this issue.

### DIFF
--- a/src/Umbraco.Web/Routing/ContentFinderByConfigured404.cs
+++ b/src/Umbraco.Web/Routing/ContentFinderByConfigured404.cs
@@ -33,11 +33,14 @@ namespace Umbraco.Web.Routing
         {
             _logger.Debug<ContentFinderByConfigured404>("Looking for a page to handle 404.");
 
+            int? domainConentId = null;
+
             // try to find a culture as best as we can
             var errorCulture = CultureInfo.CurrentUICulture;
             if (frequest.HasDomain)
             {
                 errorCulture = frequest.Domain.Culture;
+                domainConentId = frequest.Domain.ContentId;
             }
             else
             {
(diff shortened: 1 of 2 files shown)


ronaldbarendse:
It does indeed make sense to use the (automatically) resolved domain content ID as `$site` variable 👍🏻 

I think we should also make resolving the `Domain`/`DomainAndUri` more open, so this can be done outside of the current Umbraco request, like in custom HTTP modules for generating sitemaps or doing redirects. I've seen a lot of duplicate code because of this exact use-case:
- https://our.umbraco.com/packages/website-utilities/friendly-sitemap/, which uses https://github.com/callumbwhyte/umbraco-routing-extensions
- https://github.com/skybrud/Skybrud.Umbraco.Redirects/blob/v2/latest/src/Skybrud.Umbraco.Redirects/Domains/DomainUtils.cs

nul800sebastiaan:
Sorry for the long delay on this one @kmeilander! This is now merged and ready for the next release, good stuff! 👍

PR #10360, opened 2021-05-31
Avoid repeated reloads of content and media cache following Deploy operations.

AndyButland:
This PR is an attempt to mitigate some issues reported on Umbraco Cloud with deployment operations containing document type changes and large databases (either numbers of nodes, versions or both).

I've observed from logs that there are repeated calls to reload the content and media cache - twice for every document type in the deployment batch - so with 5 document types, there are 10 full reloads.

This seems unnecessary - as when a document type changes _all_ content is reloaded.  There's no logic to refresh parts of the cache for different document types, hence it seems we don't really care which document type has been updated in this context, just that one has.

So I've amended the logic for refreshing the cache to handle only single content type event, even if there are several.

I [initially tried to adapt this in Deploy](https://github.com/umbraco/Umbraco-Deploy/pull/540) - by only raising one event instead of several - but there are other components, namely Heartcore, that rely on knowing exactly which content type has changed via the distributed cache.

To Test:

- With a Cloud site, push an update containing several document type changes.
- View the logs.  You should see only one entry like the following associated with the deployment.  Before this PR you'd see 2 x number of document types.

```
{"@t":"2021-05-31T13:54:34.1191291Z","@mt":"{StartMessage} [Timing {TimingId}]","StartMessage":"Loading content from database",...
{"@t":"2021-05-31T13:55:08.4989863Z","@mt":"{EndMessage} ({Duration}ms) [Timing {TimingId}]","EndMessage":"Completed.",...
{"@t":"2021-05-31T13:55:08.5249691Z","@mt":"{StartMessage} [Timing {TimingId}]","StartMessage":"Loading media from database",...
{"@t":"2021-05-31T13:55:09.3561273Z","@mt":"{EndMessage} ({Duration}ms) [Timing {TimingId}]","EndMessage":"Completed.",...
```

### DIFF
--- a/src/Umbraco.Web/Cache/DistributedCacheBinder.cs
+++ b/src/Umbraco.Web/Cache/DistributedCacheBinder.cs
@@ -61,11 +61,15 @@ namespace Umbraco.Web.Cache
         /// <inheritdoc />
         public void HandleEvents(IEnumerable<IEventDefinition> events)
         {
-            // ensure we run with an UmbracoContext, because this may run in a background task,
-            // yet developers may be using the 'current' UmbracoContext in the event handlers
+            // Ensure we run with an UmbracoContext, because this may run in a background task,
+            // yet developers may be using the 'current' UmbracoContext in the event handlers.
             using (_umbracoContextFactory.EnsureUmbracoContext())
             {
-                foreach (var e in events)
+                // When it comes to content types types, a change to any single one will trigger a reload of the content and media caches.
+                // As far as I (AB) can tell, there's no type specific logic here, they all clear caches for all content types, and trigger a reload of all content and media.
+                // We also have events registered for Changed and Saved, which do the same thing, so really only need one of these.
+                // Hence if we have more than one document or media types, we can and should only handle one of the events for one, to avoid repeated cache reloads.
+                foreach (var e in GetReducedEventList(events))
                 {
                     var handler = FindHandler(e);
                     if (handler == null)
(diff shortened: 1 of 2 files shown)


nikolajlauridsen:
Code looks fine to me. I'm just wondering how we'd port this over to netcore, since the DistributedCacheBinder is no longer a thing and there is no equivalent to `HandleEvents` for notifications, so there's not really any place for this logic to live.

rasmusjp:
I'm wondering if there's a way (maybe in Deploy) to aggregate the events of the same type into a single one, given the properties that contain the changed data is an `IEnumerable`. That way we'll end up with at a single event for each service containing all the changes

src/Umbraco.Web/Cache/DistributedCacheBinder.cs
```diff
@@ -80,5 +84,49 @@ public void HandleEvents(IEnumerable<IEventDefinition> events)
...
+                        gotMemberType = true;
+                    }
+                }
+                else
+                {
+                    reducedEvents.Add(evt);
+                }
+            }
+
+            return reducedEvents;
```
Shazwazza, line 129: Doesn't this mean it will just return an empty list of events if none of the events are of the types being looked for in this method? Does that mean that we've essentially removed all events from occurring for non content type changes?

AndyButland: That's not my intention... if it's a non content type change it should hit the `else` clause and be added to the list.  So only the content/media/member type based events are reduced down to at most one of each.

Shazwazza: Ah, yes i see that now 👍

bergmania (approved):

PR #10367, opened 2021-05-31
Support custom SVG icon in user group filter

bjarnef:
### Prerequisites

### Description
If using a custom SVG icon for a user group this wasn't shown in user group filter.

image


nathanwoulfe (approved):
Thanks @bjarnef, will tag this for the 13/14 patch releases.

PR #10361, opened 2021-05-31
Support custom SVG icon in grid editor row configuration

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
This PR adjust grid editors to allow use custom SVG icons, e.g. via package.manifest:

```
{
    "gridEditors": [
        {
            "name": "Custom editor",
            "alias": "custom",
            "view": "textstring",
            "icon": "coffee-cup"
        }
    ]
}
```

image

I have replaced icons in grid editor placeholders, so if you wanted to replace the default grid editors with a custom icon it should potentially work.
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI/config/grid.editors.config.js

However in my test it didn't seem to use `coffee-cup` but fallback to e.g. `icon-picture`.

image

PR #10368, opened 2021-05-31
Support custom SVG icon in Nested Content

bjarnef:
### Prerequisites

### Description
If using a custom SVG icon this wasn't shown in Nested Content. With a few adjustments this can be supported although the newer Block List editor probably is preferred on new projects.

image


nathanwoulfe:
@bjarnef this is all good, but depends on my update in #10361 to update the itempicker dialog. Will need to merge that one first, to avoid introducing (transient) issues with icons not showing in the picker. Otherwise, this works fine - the NC items display the SVG icon.

nathanwoulfe (approved):

Issue #10289, opened 2021-05-19
Media Picker v3 - Amount (min and max) isn't stored and working


cheeseytoastie:
### Which Umbraco version are you using?

v9 beta002

### Bug summary

If you create a fresh site I usually create a Single Image Picker. 
Chose the v3 Media Picker. Went to set the amount from 0 - 1 - this isn't saved on save / submit. 

If you then try to use this datatype on a content node

### Steps to reproduce

Create a data type "Single Media Picker"

Select "Media Picker v3"

Set the amounts - any figures though I used 0 - 1. 
Save

Reload the datatype you'll see the amounts are not saved. 

If you add to a doctype and try to add an image you get a warning:

Minimum undefined entries, requires -1 more.  

### Expected result / actual result

Expected values to save. 

Expected ability to have 0 -1 images picked on doctype / content node.

nul800sebastiaan:
Thanks @cheeseytoastie! I can reproduce this, even in the latest nightly build. I don't know if this because the v8 branch has only been partially merged up (Media Picker v3 is brand new in the upcoming 8.14) or if there's a different problem. In v8 it works perfectly.

nul800sebastiaan:
Looks like we're definitely out of sync on the v9 branch, so that must be it, I'll like this to https://github.com/umbraco/Umbraco-CMS/issues/10295 which should have the same cause.

nul800sebastiaan:
Update: it wasn't because we were out of sync, after merging, this is still a problem.

elit0451:
Fixed in #10370

PR #10370, opened 2021-06-01
V9: Fixed serialization issue Amount (min and max) isn't stored and working with Media picker 3

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/10289


elit0451:
Looks good and the issue reported is fixed. I've added the same fix for `CropConfiguration` as its values weren't persisted either. Merging 👍

PR #10376, opened 2021-06-03
Fix Rollback feature not creating diffs when properties are moved between tabs in EditorModel events

nzdev:
The rollback feature breaks if you change which tab a property is on using the EditorModel events as it can't generate a diff.
This is because the createDiff function assumes the old version and the current version have the same tab/property indexes.
This PR fixes this by finding the old property by the current property alias.

### Prerequisites

How to test:
Reorder properties on tabs in  the EditorModelEvent.

Create two versions of a document.
Try to roll back to the previous version
Rollback diff won't appear (just shows loading)
Unable to rollback.


nathanwoulfe:
@nzdev I've added some changes to the property lookup to avoid unnecessary iteration - using `findIndex` to get the property rather than an inner `for...`, which means we can break from the outer loop once we find the correct property and not need to loop through all the properties/tabs.

Will let the tests run, then merge on green lights.

nathanwoulfe (approved):
All works as described, thanks @nzdev

PR #10378, opened 2021-06-03
Close Modal window on error

faithbr1nger:
### Prerequisites

### Description
Issue:
When unPublish event is cancelled modal window doesn't close and remains stuck even when API has returned HTTP 40x. 
    
Reproduce steps:
Intercept unpublish event and cancel it. Expected result - modal window closes. Actual result - modal remains open.

Resolution:
Added overlayService.close(); to error function. Without it modal window remains stuck even though unPublish event is cancelled.

Image to bug attached:
    
image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/content/edit.controller.js
@@ -598,6 +598,7 @@
                                 formHelper.resetForm({ scope: $scope, hasErrors: true });
                                 $scope.page.buttonGroupState = 'error';
                                 handleHttpException(err);
+                                overlayService.close();
                             });
 
 


mikecp:
Hi @faithbr1nger ,
Thanks for your first contribution to Umbraco ! 👍 

I have checked your update locally with the latest version of the CMS, and I am not sure I get the same behavior as you, so I wanted to make sure:

- Without your change, in case  of error, the window is indeed not closed but I see the error message in the top part of it (cf. picture)

image

-  With your change, the window is indeed closed immediately, but then I miss the error information.

So I would tend to say that the first situation is preferable to the second.
Could you maybe check if you have the same behavior? Or maybe I misunderstood something in the case you want to solve?

Thanks!

nul800sebastiaan:
Since there's been no reply I'll close this for now. Feel free to update @faithbr1nger when you have a chance, then we can re-evaluate.

PR #10386, opened 2021-06-03
Implements Examine/Lucene index syncing for use on Azure

Shazwazza:
This is the new implementation of index syncing for use with Azure. Previously this was with "SyncTempEnvDirectoryFactory". With v9, the setting will be similar but the underlying mechanism that does this is different. Previously this used a Lucene Directory that proxied to 2 underlying Directories (main and local) and would only read from local. This also meant that it would lock both underlying indexes.

With Lucene 4.8, the Lucene Replicator is built in natively which is a nicer way to do this. This was the very original implementation of this some years ago but it required a huge amount of code and maintenance to do the replication. Now this code is built into Lucene, it's much better.

So now on startup, when Examine is activated, the local index is replicated from the main index location (once). Then the local index is opened/locked and is used for all searching and indexing operations. When a write occurs, behind the scenes on a timer the replicator will replicate the local index back to the main storage. 

## Testing

To enable the syncing directory, you need to specify this in config:

```json
"Umbraco": {
    "CMS": {
      "Examine": {
        "LuceneDirectoryFactory": "SyncedTempFileSystemDirectoryFactory"
      }
   }
}
```

The options for "LuceneDirectoryFactory" are:

* SyncedTempFileSystemDirectoryFactory - which means the index will operate on a local index created in the processes %temp% location and will replicate back to main storage in Umbraco/Data/Temp/ExamineIndexes
* TempFileSystemDirectoryFactory - which means the index will operate only in the processes %temp% directory location
* Default - use the default location Umbraco/Data/Temp/ExamineIndexes. This is used if the config value is not supplied.

Each of these options should be tested including not having an option specified. The %temp% folder on windows when running dotnet is typically: `C:\Users\{USERNAME}\AppData\Local\Temp\ExamineIndexes\0c7653e3909dd61acb53466d2baf53b6\ExternalIndex` but this may differ on your machine.

Make sure you test with not having any indexes on startup. NOTE: Index rebuild on startup is delayed by 1 minute so it won't rebuild until a minute after the first request.

## Other changes

* Index rebuild on startup will now occur if the index exists but it has no documents
* Fixes issue where we would re-index all IUmbracoIndexes for content or members which is added overhead, we now only reindex content items for IUmbracoContentIndex and member data for IUmbracoMemberIndex
* The index rebuild background task on startup has a minute delay but this was delaying the whole background task queue, this was accidentally caused because the Task.Run was awaited instead of being a fire and forget task


bergmania (approved):
Everything seems to work as expected.. Good cache with the fire and forget.

PR #10393, opened 2021-06-04
V9: Add new email Model for notifications

nikolajlauridsen:
This PR adds a new email model to be sent with email notifications.

The reason for this is to be able to both send the email address and the email display name, since this is used in Deploy.

I opted to add a new model instead of trying to change the old one, since it doesn't make much sense for that to have the display name, since it gets converted into a MimeMessage down the line.

While I was working on this I also noticed that we added the ability to fall back to a "from email" from the configuration, however the way the code currently was, the configured "from email" was never used, so I've also amended the code so that is no longer the case.

### DIFF
--- a/src/Umbraco.Core/Models/Email/NotificationEmailAddress.cs
+++ b/src/Umbraco.Core/Models/Email/NotificationEmailAddress.cs
@@ -7,11 +7,11 @@ namespace Umbraco.Cms.Core.Models.Email
     {
         public string DisplayName { get; }
 
-        public string Adress { get; }
+        public string Address { get; }
 
         public NotificationEmailAddress(string address, string displayName)
         {
-            Adress = address;
+            Address = address;
             DisplayName = displayName;
         }
     }
--- a/src/Umbraco.Core/Models/Email/NotificationEmailAddress.cs
+++ b/src/Umbraco.Core/Models/Email/NotificationEmailAddress.cs
@@ -1,0 +1,18 @@
+namespace Umbraco.Cms.Core.Models.Email
+{
+    /// <summary>
+    /// Represents an email address used for notifications. Contains both the address and its display name.
+    /// </summary>
+    public class NotificationEmailAddress
+    {
+        public string DisplayName { get; }
+
+        public string Adress { get; }
+
+        public NotificationEmailAddress(string address, string displayName)
+        {
+            Adress = address;
+            DisplayName = displayName;
+        }
+    }
+}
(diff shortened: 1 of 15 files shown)


src/Umbraco.Core/Models/Email/NotificationEmailAddress.cs
```diff
@@ -0,0 +1,18 @@
+namespace Umbraco.Cms.Core.Models.Email
+{
+    /// <summary>
+    /// Represents an email address used for notifications. Contains both the address and its display name.
+    /// </summary>
+    public class NotificationEmailAddress
+    {
+        public string DisplayName { get; }
+
+        public string Adress { get; }
```
AndyButland, line 10: Typo here - `Address`.

src/Umbraco.Infrastructure/Extensions/EmailMessageExtensions.cs
```diff
@@ -58,6 +55,58 @@ public static MimeMessage ToMimeMessage(this EmailMessage mailMessage, string co
...
+                if (notificationAddress is not null)
+                {
+                    notificationAddresses.Add(notificationAddress);
+                }
+            }
+
+            return notificationAddresses;
+        }
+
         private static void AddAddresses(MimeMessage message, string[] addresses, Func<MimeMessage, InternetAddressList> addressListGetter, bool throwIfNoneValid = false)
```
AndyButland, line 110: Just to make the file easier to read top to bottom, maybe move this helper method up to below the public method that uses it.  Or if you prefer public methods at the top, make this the first private helper method in the class.

AndyButland (approved):

src/Umbraco.Core/Models/Email/NotificationEmailAddress.cs
nikolajlauridsen: Ahh good spot, for some reason I always misspell Address 😄

src/Umbraco.Infrastructure/Extensions/EmailMessageExtensions.cs
nikolajlauridsen: Good point, I don't think it's too big an issue, so I moved the helper method up below the public method.

AndyButland: No, no big deal - just think it's nice when you read a class from to to bottom that methods are in an order reflecting how they are called.  Sounds good.

nikolajlauridsen: Oh I agree, I was referring to the mixing of public/private methods 😄

PR #10396, opened 2021-06-04
Add missing UrlHelper GetCropUrl methods

Jeavon:
As discussed at https://github.com/umbraco/Umbraco-CMS/pull/10233 I have added 3 extensions to UrlHelper. The first 2 are friendly extension methods, they could both be dropped as all could be done with the 3rd method accepting the IPublishedContent and all the other parameters

There are a few other methods in v8 such as the string method but I think these should be removed as they were added over time for compatibility https://github.com/umbraco/Umbraco-CMS/blob/809fd819821770472dec3eb2d7241f7d39f3c918/src/Umbraco.Web/UrlHelperRenderExtensions.cs#L217

### DIFF
--- a/src/Umbraco.Web.Common/Extensions/UrlHelperExtensions.cs
+++ b/src/Umbraco.Web.Common/Extensions/UrlHelperExtensions.cs
@@ -10,6 +10,7 @@ using Umbraco.Cms.Core;
 using Umbraco.Cms.Core.Configuration;
 using Umbraco.Cms.Core.Hosting;
 using Umbraco.Cms.Core.Models;
+using Umbraco.Cms.Core.Models.PublishedContent;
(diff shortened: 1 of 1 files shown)


ronaldbarendse:
IMHO the real issue is returning `IHtmlString`, which prevents Razor to correctly encode the URL. Using `HttpUtility.HtmlEncode()` is also incorrect if the value is used in an attribute (which it most likely is): you should then use `HttpUtility.HtmlAttributeEncode()` instead...

Razor already automatically HTML/URL encodes values: https://docs.microsoft.com/en-us/aspnet/core/security/cross-site-scripting?view=aspnetcore-5.0#html-encoding-using-razor.

Changing the returned value type is a breaking change, so that might be a good thing to do before v9 is released!

Jeavon:
The reason for that is that it's mainly used within inline style blocks, such as background images

PR #10400, opened 2021-06-05
Updated umbtable.directive.js for a working example

patrickdemooij9:
I noticed that the example on https://our.umbraco.com/apidocs/v8/ui/#/api/umbraco.directives.directive:umbTable is actually not working. items should be replaced with vm.items.


nathanwoulfe (approved):
Thanks @patrickdemooij9, well spotted!

PR #10406, opened 2021-06-07
Updated signature to Umbraco pipeline filter to avoid multiple calls to UseEndpoints().

AndyButland:
I'm not 100% sure if this is necessary or advisable, but thought would PR for consideration anyway.

In my attempts to get Deploy's SignalR authentication working I read about some issues with calling `UseEndpoints()` multiple times - see [here](https://github.com/dotnet/aspnetcore/issues/17750) and [here](https://stackoverflow.com/questions/59281967/why-is-middleware-not-executed-when-there-are-multiple-calls-to-app-useendpoints) - which effectively we do when endpoints are registered via the `IUmbracoPipelineFilter`.

With this change to the signature instead of adding this code like this:

```
    applicationBuilder =>
    {
        var hubRoutes = applicationBuilder.ApplicationServices.GetRequiredService<DeployUiHubRoutes>();
        applicationBuilder.UseEndpoints(endpoints =>
        {
            hubRoutes.CreateRoutes(endpoints);
        });
    }
```

We'd do this:

```
    endpointBuilder =>
    {
        var hubRoutes = endpointBuilder.AppBuilder.ApplicationServices.GetRequiredService<DeployUiHubRoutes>();
        hubRoutes.CreateRoutes(endpointBuilder.EndpointRouteBuilder);
    }
```

Meaning there'd only be one `UseEndpoints()` call.

To repeat, this didn't solve the problem I was looking into but it may still be incorrect how we are doing things currently and perhaps this is a better approach.

### DIFF
--- a/src/Umbraco.Web.Common/ApplicationBuilder/IUmbracoApplicationBuilder.cs
+++ b/src/Umbraco.Web.Common/ApplicationBuilder/IUmbracoApplicationBuilder.cs
@@ -1,6 +1,4 @@
 using System;
-using Microsoft.AspNetCore.Builder;
-using Umbraco.Cms.Core.Services;
 
 namespace Umbraco.Cms.Web.Common.ApplicationBuilder
 {
--- a/src/Umbraco.Web.Common/ApplicationBuilder/IUmbracoPipelineFilter.cs
+++ b/src/Umbraco.Web.Common/ApplicationBuilder/IUmbracoPipelineFilter.cs
@@ -1,4 +1,6 @@
+using System;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Routing;
 
 namespace Umbraco.Cms.Web.Common.ApplicationBuilder
 {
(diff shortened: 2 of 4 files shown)


bergmania:
Looks good to me. @Shazwazza, do you see any reason not to do this

nikolajlauridsen (approved):
Looks good to me as well 👍

Shazwazza:
The reason why other middleware doesn't execute between some calls to UseEndpoints is due to the ordering. UseEndpoints is a terminating middleware. Anything registered after that will not really work. The CMS has a catch all endpoint which should in theory come last which is why we didn't have a pre/post. Calling UseEndpoints many times isn't a problem, it's trying to add some non-endpoint middleware in between these calls that is the problem.

We have PrePipeline/PostPipeline for non-endpoint middleware and then OnEndpoints for endpoints. Just not sure I see the reason for pre/post endpoints, is there a particular reason?

AndyButland:
The change here really is just to provide a signature that prevents the pipeline being configured with multiple `OnEndpoints()` calls, which I read might be a problem (links above).  If we're sure that's not an issue, then there's probably no need for it.

In implementing this rather than replacing the existing method I added a new one - so we have a hook before `OnEndpoints()` that exists currently, and now also within it - and just to maintain existing functionality.   Given we're still in beta, we could just replace the existing with the new one if there's no value in having both.

Shazwazza:
> multiple OnEndpoints() calls, which I read might be a problem (links above).

This is not an issue. The issue is when people try to use other non endpoint middleware after a call to UseEndpoints - it will not work because UseEndpoints is terminating middleware. Any middleware that is registered after UseEndpoints that is not also endpoints will never be executed if one of the endpoints matches. Which in our case will always be correct because we have a catch all endpoint. Calling UseEndpoints multiple times is totally ok and will be done in a lot of cases. 

People can (and will) call their own UseEndpoints after our own UseUmbraco call and that is fine. But any custom middleware that is registered after UseUmbraco will not execute because our last UseEndpoints call is a catch all. 

Is this making sense?

AndyButland:
Will close this then as doesn't appear it's necessary.

PR #10409, opened 2021-06-07
Moves internal logic for populating the current user's identity from the authentication cookie into a public extension method to allow for use in packages.

AndyButland:
Assuming the approach to resolving SignalR authentication in Deploy [here](https://github.com/umbraco/Umbraco-Deploy/pull/554) is approved, this amend will allow some core code to be reused and avoid duplicating between the two projects.


AndyButland:
Closing as not necessary.

PR #10410, opened 2021-06-07
NetCore: Fix UmbracoProject template to make use of the custom name value

elit0451:
## Details

**UmbracoProject** dotnet template will use the value passed as `-n/--name` parameter and create a new Umbraco project with it - *.csproj name will reflect that, as well as the namespace value.

If no name is specified, the name of the current directory is used.

## Test
- Run the **build.ps1** script
- Use `dotnet new -i pathtoyourproject\build.out\Umbraco.Templates.9.0.0-beta003.nupkg` command to install the templates
- Create a project and give it a name like `dotnet new umbraco -n MyCustomUmbracoSolution`
- Project name, folder, namespace, etc. should be with `MyCustomUmbracoSolution` name 
  - You can also try adding a new class, the namespace shouldn't be different
- Confirm that you can install Umbraco using the template

### DIFF
--- a/build/templates/UmbracoProject/.template.config/template.json
+++ b/build/templates/UmbracoProject/.template.config/template.json
@@ -15,7 +15,7 @@
     },
     "primaryOutputs": [
         {
-			"path": "UmbracoProject.csproj"
+            "path": "UmbracoProject.csproj"
         }
     ],
     "postActions": [
--- a/build/templates/UmbracoProject/.template.config/template.json
+++ b/build/templates/UmbracoProject/.template.config/template.json
@@ -15,7 +15,7 @@
     },
     "primaryOutputs": [
         {
-        "path": "UmbracoProject.csproj"
+			"path": "UmbracoProject.csproj"
         }
     ],
     "postActions": [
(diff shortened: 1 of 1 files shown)


warrenbuckley:
I haven't tested this PR @elit0451 but curious why we needed to added this Namespace Replacer back which I removed recently in another PR.

Curious as to what I missed or why it doesn't work...
https://github.com/umbraco/Umbraco-CMS/pull/10314

elit0451:
Hey @warrenbuckley 

We have at least 2 places where we want to use the name provided in the `dotnet new` command (the one coming from the `-n` argument):
- In `primaryOutputs` we want `UmbracoProject.csproj` to be renamed;
- And the namespace of _Startup.cs_ and _Program.cs_ file - `Umbraco.Cms.Web.UI.NetCore` - since the file contents are copied over from the CMS codebase, therefore the naming is persisted.

Usually, the value declared after `sourceName` is the one that will be replaced with the name the user specifies, so it will look for any occurrences of `Umbraco.Cms.Web.UI.NetCore` and not for **UmbracoProject**. Meaning that when a user creates a project, they will end up with the correct namespace (ex: **MyCustomUmbracoSolution**) in those 2 file but any other file added will have the namespace `UmbracoProject` because of the UmbracoProject.csproj.
That leads us to the fact that the csproj should have the same name as the one specified for `sourceName`:
- either have `Umbraco.Cms.Web.UI.NetCore.csproj` when `sourceName: Umbraco.Cms.Web.UI.NetCore`;
- or replace the `sourceName` value with `UmbracoProject` - but then the namespaces of _Startup.cs_ and _Program.cs_ wouldn't reflect the correct naming.

mroberts91:
I would love to see this get merged. I am always creating projects with `dotnet new` and it is always just a touch annoying when the   `-n` flag does not contain standard functionality and I have to remember to change the project name and namespaces..

bergmania:
Merged in as part of this https://github.com/umbraco/Umbraco-CMS/pull/10467 - As I tested both in one go

PR #10417, opened 2021-06-08
Refactor MigrationPlan to separate the executor from the plan itself.

Shazwazza:
No logic changes, just shifting around responsibilities. Now we can have the MigrationPlan as part of the core project and a separate IMigrationPlanExecutor. 

A bunch of code cleaned up as well like injecting services where it makes sense and removing some code that was not used.


bergmania (approved):
Looks good, and nice with more logic moved to Core 💪

PR #10418, opened 2021-06-08
Implements unattended package migrations for both explicit and implicit migrations

Shazwazza:
__#10449 Must be reviewed and merged first__

## Changes

* Moves unattended logic to notifications/handlers
* Removes unneeded code like package uninstall and package files
* Adds the actual MigrationPlan instances to the migration context so any step can reference the plan itself (we use this for the implicit migrations)
* Adds package migration expressions to do the actual package install in steps

## Testing

* try the package migrations using code mentioned below
* make one of your steps throw an exception and ensure that the boot failed exception middleware reports the problem

## To create an explicit package migration

An explicit package migration is a package migration created by a package developer that wants full control over the migration process and possibly has custom steps (not just installing the umbraco data from the package.xml file).

```cs
public class TestPackageMigrationPlan : PackageMigrationPlan
{
    public TestPackageMigrationPlan()
        // The name of your package, this will be used
        // to track the packages migration state.
        : base("Articulate") { }

    protected override void DefinePlan()
    {
        // Define migration plan.
        // Each step must have a new unique GUID.
        // Each step maps to a migration class which executes migration logic.
            
        To<TestMigration1>(new Guid("BB02C392-4007-4A6C-A550-28BA2FF7E43D"));

        // The last step listed is the "Final" step.
        // This is the value that will be recorded in the database.
        // If the database value doesn't match this value, the the steps
        // registered in this migration will execute from the last registered
        // step in the database (if any).
        // This is how incremental migrations work (i.e. for upgrades)

        To<TestMigration2>(new Guid("1E3DA142-D5E0-4E3D-A57E-29E8B5B2A9E8"));
    }

    private class TestMigration1 : PackageMigrationBase
    {
        public TestMigration1(IPackagingService packagingService, IMigrationContext context)
            : base(packagingService, context)
        {
        }

        // This is where the logic goes to process a migration step.
        // Logic in here can do all sorts of things like created/modify Database
        // tables or execute c# logic.
        // You can even run logic using the Umbraco services (you need to
        // inject them).
        public override void Migrate()
        {
            // This step will import the Umbraco data (package.xml) file
            // that was exported from the package builder section in the
            // back office.
            // In this case, this is an embedded resource that exists in
            // the same location/namespace as this migration class.
            ImportPackage.FromEmbeddedResource<TestMigration2>().Do();
        }
    }

    private class TestMigration2 : PackageMigrationBase
    {
        public TestMigration2(IPackagingService packagingService, IMigrationContext context) : base(packagingService, context)
        {
        }

        public override void Migrate()
        {
            // The first step already did this, however this is an example
            // of an 'upgrade'. Perhaps the first step previously ran
            // and now you've updated the package.xml manifest with some
            // changes and you want those applied in a new package release.
            // So now you can just execute this step again which will be
            // exactly like you were to upgrade a v8 package in the back office.
            ImportPackage.FromEmbeddedResource<TestMigration2>().Do();
        }
    }
}
```

## To create an implicit package migration

An implicit package migration is for package developers who just want the package.xml installed. To do this, only a single class needs to be created:

```cs
public class TestAutomaticPackageMigrationPlan : AutomaticPackgeMigrationPlan
{
    public TestAutomaticPackageMigrationPlan()
        // The name of your package, this will be used
        // to track the packages migration state.
        : base("MyPackage")
    {
    }
}
```

The requirements for this is that the package.xml file is an embedded resource that exists in the same folder/namespace as this class.


src/Umbraco.Tests.Integration/Umbraco.Infrastructure/Services/KeyValueServiceTests.cs
```diff
@@ -19,6 +20,27 @@ public class KeyValueServiceTests : UmbracoIntegrationTest
...
 
+        [Test]
+        public void Can_Query_For_Key_Prefix()
+        {
+            // Arrange
+            KeyValueService.SetValue("test1", "hello1");
+            KeyValueService.SetValue("test2", "hello2");
+            KeyValueService.SetValue("test3", "hello3");
+            KeyValueService.SetValue("test4", "hello4");
+
```
bergmania, line 31: Just to make it even more clear, that the service do not just return everything
```suggestion
            KeyValueService.SetValue("test4", "hello4");
            KeyValueService.SetValue("someotherprefix1", "helloagain1");
```

bergmania (approved):
Looks good.. Just small suggestions, that I think would make it easier to understand. What do you think?

src/Umbraco.Tests.Integration/Umbraco.Infrastructure/Services/KeyValueServiceTests.cs
```diff
@@ -19,6 +20,27 @@ public class KeyValueServiceTests : UmbracoIntegrationTest
...
+        public void Can_Query_For_Key_Prefix()
+        {
+            // Arrange
+            KeyValueService.SetValue("test1", "hello1");
+            KeyValueService.SetValue("test2", "hello2");
+            KeyValueService.SetValue("test3", "hello3");
+            KeyValueService.SetValue("test4", "hello4");
+            KeyValueService.SetValue("someotherprefix1", "helloagain1");
+            // Act
+            IReadOnlyDictionary<string, string> value = KeyValueService.Find("test");
```
Shazwazza, line 33:
```suggestion
            IReadOnlyDictionary<string, string> value = KeyValueService.FindByKeyPrefix("test");
```

Shazwazza:
@bergmania I've committed your proposals and fixed up a usage of it. I'll change the repository method signature too but will do that in my current active branch.

src/Umbraco.Infrastructure/Runtime/RuntimeState.cs
```diff
@@ -203,11 +201,14 @@ private UmbracoDatabaseState GetUmbracoDatabaseState(IUmbracoDatabaseFactory dat
                         return UmbracoDatabaseState.NeedsUpgrade;
                     }
 
+                    // TODO: We will need to scan for implicit migrations.
 
                     // TODO: Can we save the result of this since we'll need to re-use it?
```
bergmania, line 206: I think that is `PackageMigrationState`'s responsibility and it is in the container as a singleton, so if it just saves the result in a field and uses that if it is not null, I guess that should be fine.

src/Umbraco.Infrastructure/Packaging/AutomaticPackgeMigrationPlan.cs
```diff
@@ -0,0 +1,50 @@
...
+        protected AutomaticPackgeMigrationPlan(string name)
+            : base(name)
+        {
+        }
+
+        protected sealed override void DefinePlan()
+        {
+            // calculate the final state based on the hash value of the embedded resource
+            Type planType = GetType();
+            PackageDataManifest = PackageMigrationResource.GetEmbeddedPackageDataManifest(planType);
```
bergmania: I wonder if this should just happen in the getter of the `PackageDataManifest` and then save the result in a backing field? Seems safer to me if we in the future wanna use  `PackageDataManifest` before we call call `DefinePlan`

bergmania (approved):
Overall looks very good...

src/Umbraco.Infrastructure/Runtime/RuntimeState.cs
Shazwazza: oops, i left these TODOs in here. it's all taken care of already so i'll just remove the todos

src/Umbraco.Core/Packaging/IPackageInstallation.cs
```diff
@@ -1,42 +1,30 @@
...
-        /// <param name="packageDefinition"></param>
-        /// <param name="compiledPackage"></param>
-        /// <param name="userId"></param>
-        /// <returns></returns>
-        IEnumerable<string> InstallPackageFiles(PackageDefinition packageDefinition, CompiledPackage compiledPackage, int userId);
+        // TODO: The resulting PackageDefinition is only if we wanted to persist what was saved during package data installation.
+        // This used to be for the installedPackages.config but we don't have that anymore and don't really want it if we can help it.
+        // Possibly, we could continue to persist that file so that you could uninstall package data for an installed package in the
+        // back office (but it won't actually uninstall the package until you do that via nuget). If we want that functionality we'll have
+        // to restore a bunch of deleted code.
```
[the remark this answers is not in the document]
Shazwazza: Yep defo a bigger discussion and will add more work. I don't think this is critical for 9.0 or anything. It would also require storing this data in the DB I think.

src/Umbraco.Infrastructure/Packaging/AutomaticPackgeMigrationPlan.cs
Shazwazza: yep makes sense

PR #10425, opened 2021-06-09
V9: add missing overload is allowed template

### LINKED ISSUE #10391: v9: IsHelpers template methods not working correctly

v9

### Bug summary

When verifying https://our.umbraco.com/Documentation/Reference/Querying/IPublishedContent/IsHelpers/ against v9 I noticed these two methods:

IsAllowedTemplate(int templateId)
IsAllowedTemplate(string templateAlias)

Do not work correctly.

### Specifics

IsAllowedTemplate(int templateId) always returns true, no matter on which page it is called and whether the ID even exists or not.

IsAllowedTemplate(string templateAlias) does not seem to be implemented (cannot convert string to int). It throws an error on the frontend, as it expects an integer.

### Steps to reproduce

I tried this with a simple setup, where I tried to check whether the current page is allowed to use the home page template or not.
Something like

```
if(Model.IsAllowedTemplate(1021)){
<p>True</p>
}
else{
<p>False</p>
}
```

IsAllowedTemplate(int templateId) returns true or false depending on whether the template selected is allowed under the model/doctype.
IsAllowedTemplate(string templateAlias) should work similarly, except with an alias.

warrenbuckley:
Fixes #10391 
This adds in a missing overload for IsAllowedTemplate that allowed you to simply pass the template alias

See the notes in 10391 about testing as out of the box in V8 and V9 it will not work until you turn on `ValidateAlternativeTemplates` to true in `WebRouting` for AppSettings configuration.

## Test Notes
* Ensure you update your appsettings.config as noted above
* Create a home doctype with matching template
* Use this in the Razor view/template

```
@if(Model.IsAllowedTemplate("home"))
{
    <h2>YEP</h2>
}
else {
    <h2>NOPE</h2>
}
```

Then modify it to an invalid string that wont be found as a template

### DIFF
--- a/src/Umbraco.Web.Common/Extensions/FriendlyPublishedContentExtensions.cs
+++ b/src/Umbraco.Web.Common/Extensions/FriendlyPublishedContentExtensions.cs
@@ -109,6 +109,9 @@ namespace Umbraco.Extensions
         public static bool IsAllowedTemplate(this IPublishedContent content, int templateId)
             => content.IsAllowedTemplate(ContentTypeService, WebRoutingSettings.Value, templateId);
 
+        public static bool IsAllowedTemplate(this IPublishedContent content, string templateAlias)
+            => content.IsAllowedTemplate(WebRoutingSettings.Value.DisableAlternativeTemplates, WebRoutingSettings.Value.ValidateAlternativeTemplates, templateAlias);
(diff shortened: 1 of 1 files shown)


bergmania:
💪

Issue #10433, opened 2021-06-10
v9: View Model validation fails with any value type


JoseMarcenaro:
### Which Umbraco version are you using?

9.0.0-beta003

### Bug summary

When using `@Html.Partial( )` , an **InvalidOperationException** is thrown if the Model type is a value type instead of a reference type.

I will submit a PR to address this issue.

### Specifics

**InvalidOperationException:** The model item passed is null, but this ViewDataDictionary instance requires a non-null model item of type 'System.Object'.

```
Microsoft.AspNetCore.Mvc.ViewFeatures.ViewDataDictionary.EnsureCompatible(object value)
Microsoft.AspNetCore.Mvc.ViewFeatures.ViewDataDictionary.SetModel(object value)
Microsoft.AspNetCore.Mvc.ViewFeatures.ViewDataDictionary.set_Model(object value)
Umbraco.Cms.Web.Common.Views.UmbracoViewPage<TModel>.MapViewDataDictionary(ViewDataDictionary viewData, Type modelType)
Umbraco.Cms.Web.Common.Views.UmbracoViewPage<TModel>.BindViewData(ContentModelBinder contentModelBinder, ViewDataDictionary viewData)
Umbraco.Cms.Web.Common.Views.UmbracoViewPage<TModel>.set_ViewContext(ViewContext value)
Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderPageCoreAsync(IRazorPage page, ViewContext context)
Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderPageAsync(IRazorPage page, ViewContext context, bool invokeViewStarts)
Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderAsync(ViewContext context)
Microsoft.AspNetCore.Mvc.ViewFeatures.HtmlHelper.RenderPartialCoreAsync(string partialViewName, object model, ViewDataDictionary viewData, TextWriter writer)
Microsoft.AspNetCore.Mvc.ViewFeatures.HtmlHelper.PartialAsync(string partialViewName, object model, ViewDataDictionary viewData)
Microsoft.AspNetCore.Mvc.Rendering.HtmlHelperPartialExtensions.Partial(IHtmlHelper htmlHelper, string partialViewName, object model, ViewDataDictionary viewData)
Microsoft.AspNetCore.Mvc.Rendering.HtmlHelperPartialExtensions.Partial(IHtmlHelper htmlHelper, string partialViewName, object model)
[...]

PR #10435, opened 2021-06-10
Allow value types as View Models

JoseMarcenaro:

### Description
Fixes issue [#10433](https://github.com/umbraco/Umbraco-CMS/issues/10433) - View Model validation fails with any value type


bergmania:
Amazing.. Good finding, and thanks for fixing it 🎉 💪

Issue #10431, opened 2021-06-10
MediaPicker3 Not showing selection after removal


DanDiplo:
### Which Umbraco version are you using?

8.14

### Bug summary

In a specific set of circumstances I get a JavaScript error from the new MediaPicker3 saying: `vm.model.value.splice is not a function` that prevents the image being selected.

### Specifics

If you have a non-mandatory image field with new `MediaPicker3` and then publish, remove and publish again you can't re-select a new image without triggering a JavaScript error in the Angular model.

### Steps to reproduce

- Create a new non-mandatory property using the new MediaPicker3
- Select an image
- Save & Publish the page
- Remove the image (using the trashcan icon)
- Save & Publish the page again (so it saves with no image)
- Select an image again
- The image won't be selected because of the following JS error:

```
angular.js?cdv=494828325:15635 TypeError: vm.model.value.splice is not a function
    at umbraco.directives.js?cdv=494828325:20243
    at Array.forEach (<anonymous>)
    at Object.submit (umbraco.directives.js?cdv=494828325:20238)
    at Object.submit (umbraco.controllers.js?cdv=494828325:3303)
    at fn (eval at compile (angular.js?cdv=494828325:1), <anonymous>:4:232)
    at UmbButtonController.destination.<computed> [as action] (angular.js?cdv=494828325:11475)
    at UmbButtonController.clickButton (umbraco.directives.js?cdv=494828325:2264)
    at fn (eval at compile (angular.js?cdv=494828325:1), <anonymous>:4:245)
    at callback (angular.js?cdv=494828325:29019)
    at ChildScope.$eval (angular.js?cdv=494828325:19461)
```

![image](https://user-images.githubusercontent.com/6676168/121511726-df3b6100-c9e0-11eb-9d60-0166d5c05692.png)

### Expected result / actual result

After removing an image in these circumstances I expect to be able to select a replacement image.

---
_This item has been added to our backlog [AB#12541](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/12541)_

patrickdemooij9:
I did some investigating and I was able to figure out why this is happening. Whenever you submit an empty mediapicker v3, the value will be an empty string on the return of PostSave:
![image](https://user-images.githubusercontent.com/11466511/121589205-19375200-ca37-11eb-8ed2-febc8cbc6a5c.png)

[...]

PR #10439, opened 2021-06-10
10431: Make sure that the return value of PostSave doesn't break the …

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10431

The issue here was that the vm.model.value was being set to "" due to the return value of PostSave from the content service.
image
In the old media picker this was fine, because we were holding a separate value containing the values. However, it's not doing that for the v3 one. I added that logic for that here. 

If there is a better way of doing this, I would love to know it!

Can be tested by saving a page with an empty media picker on it. Previously this would break the picker.


nul800sebastiaan:
Thanks again @patrickdemooij9 - but closing this in favor of https://github.com/umbraco/Umbraco-CMS/pull/10633 👍

PR #10445, opened 2021-06-11
Color picker: Fix label view

BatJan:
### Prerequisites

### Description
While implementing a recent site I noticed that the color swatch label can break a little bit if the text is longer than the swatch width allows. This tiny little PR fixes the issue by substracting the padding + border width from the allowed max-width of 100%.

**Before**
before

**After**
after


nathanwoulfe (approved):
Nice pickup @BatJan!

PR #10450, opened 2021-06-11
Added notifications toggle to user groups

nikcio:
### Prerequisites

### Description

Added a notifications toggle to the default permissions of user groups.

image

It's now possible to hide the "Notifications..." menu in the content node actions menu.

image

I've added the letter to the default user groups so Umbraco functions as normal with the notifications tab activated. (See DatabaseDataCreator.cs)
But I'm not quite sure if a migration should be added to the upgrade folder so it doesn't deactivate the function on already running installations of Umbraco.


src/Umbraco.Core/Migrations/Install/DatabaseDataCreator.cs
```diff
@@ -188,11 +188,11 @@ private void CreateUserData()
...
+            _database.Insert(Constants.DatabaseSchema.Tables.UserGroup, "id", false, new UserGroupDto { Id = 3, StartMediaId = -1, StartContentId = -1, Alias = Constants.Security.EditorGroupAlias, Name = "Editors", DefaultPermissions = "CADMOSKTPUZ:5FïN", CreateDate = DateTime.Now, UpdateDate = DateTime.Now, Icon = "icon-tools" });
+            _database.Insert(Constants.DatabaseSchema.Tables.UserGroup, "id", false, new UserGroupDto { Id = 4, StartMediaId = -1, StartContentId = -1, Alias = Constants.Security.TranslatorGroupAlias, Name = "Translators", DefaultPermissions = "AFN", CreateDate = DateTime.Now, UpdateDate = DateTime.Now, Icon = "icon-globe" });
+            _database.Insert(Constants.DatabaseSchema.Tables.UserGroup, "id", false, new UserGroupDto { Id = 5, StartMediaId = -1, StartContentId = -1, Alias = Constants.Security.SensitiveDataGroupAlias, Name = "Sensitive data", DefaultPermissions = "N", CreateDate = DateTime.Now, UpdateDate = DateTime.Now, Icon = "icon-lock" });
```
ronaldbarendse, line 195: I don't think the Translators user group should have this permission by default and the Sensitive data user group is only used as a marker/indicator, so shouldn't have it either: 
```suggestion
            _database.Insert(Constants.DatabaseSchema.Tables.UserGroup, "id", false, new UserGroupDto { Id = 4, StartMediaId = -1, StartContentId = -1, Alias = Constants.Security.TranslatorGroupAlias, Name = "Translators", DefaultPermissions = "AF", CreateDate = DateTime.Now, UpdateDate = DateTime.Now, Icon = "icon-globe" });
            _database.Insert(Constants.DatabaseSchema.Tables.UserGroup, "id", false, new UserGroupDto { Id = 5, StartMediaId = -1, StartContentId = -1, Alias = Constants.Security.SensitiveDataGroupAlias, Name = "Sensitive data", DefaultPermissions = "", CreateDate = DateTime.Now, UpdateDate = DateTime.Now, Icon = "icon-lock" });
```

nikcio:
@ronaldbarendse any news on this? I've included your changes a while back.

mikecp (changes requested):
Hi @nikcio ,

Thanks for this PR, I've just tested it and it adds some nice extra granularity👍
And our apologies for picking this up so lately 😬

To answer your question about providing a migration for the DB, this will indeed be needed as we would otherwise introduce a breaking change on all existing installations where the "Notifications" menu would just disappear.

I hope you're still up for the challenge and I promise I'll check the migration right away 😁 

Thanks!

nikcio:
Hi @mikecp,

I'm not that strong in NPoco but I think I figured it out. The migration I added should result in SQL like this being run:
```sql
Update umbracoUserGroup SET userGroupDefaultPermissions = userGroupDefaultPermissions + 'N' WHERE id = '1' OR id = '2' OR id = '3'
```

mikecp (approved):

mikecp:
HI @nikcio ,

Thanks for the NPoco migration. Unfortunately it did not work because the "Update" flow expects an anonymous object containing the properties/values to adapt, not a "SQL statement string".
In our case, we need the property value itself in order to update/append the "N", so I don't think it is possible with the Update syntax.

I took the liberty of adapting it to a SQL/Execute syntax, which I tested and it worked as expected. I also updated the "where" clause to replace the Id's by aliases. The idea here is that, if somehow one of the user groups was deleted and then re-created in the backoffice, the Id would be different and therefore the migration would not have the expected outcome. Checking on alias ensures that all the groups we want will be updated, regardless of their Id.

I also adapted the UmbracoPlan file, because your migration needs to be the last in the list, but due to a previous "bug/not following guidelines", some migration was put under the "//FINAL" line. So, I moved that one back up, just on top of yours.

I hope you won't mind my changes, but since I had already made them locally to test, I figured it would be more efficient to just push them instead of asking you to "copy/paste" them at your side 😁

Cheers and thanks again for this nice addition! 👍🎉

Issue #10429, opened 2021-06-10
Searching in mediapicker doesn't filter according to allowedtypes


SofusMunch:
### Which Umbraco version are you using?

8.14

### Bug summary

When setting allowed media types for mediapicker3 the search function doesn't filter the result according to the selected allowed types. It does show a "not allowed" error in the property editor if invalid types are chosen, but in my case there are specific media types some users shouldn't even see. 

It's also inconsistent since not allowed types aren't selectable when navigating the tree in the picker, but are selectable as a search result. 

### Steps to reproduce

Not selectable;
![not selectable](https://user-images.githubusercontent.com/789813/121506602-26772100-c9e4-11eb-9cd3-30e62a8be26a.png)

Selectable after search:
![searched_selectable](https://user-images.githubusercontent.com/789813/121506614-29721180-c9e4-11eb-8fb9-5b5858c11ff3.png)

patrickdemooij9:
Created a PR to fix this issue: https://github.com/umbraco/Umbraco-CMS/pull/10451

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/10451 and cherry picked for 8.14.1 in 8a810e698c1b99affc9536abcd06f0cb3f40e581

PR #10451, opened 2021-06-11
Fix: Searching in mediapicker doesn't filter according to allowedtypes #10429

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10429

On normal load, the code will pass through the filtered property set (https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/common/infiniteeditors/mediapicker/mediapicker.controller.js#L548). However, this doesn't occur when the results come from the search. I added the same logic to the search results.

fix-10429


nul800sebastiaan:
Thanks a bunch @patrickdemooij9 - that does the trick and works well! 👍

SofusMunch:
I forgot to mention there might be a similar problem when opening the detailsdialog (cropping, focal point). There is an option to reselect image. I’m pretty the filter isn’t passed along to that either.

nul800sebastiaan:
@SofusMunch no that works now as well 👍 

image

PR #10452, opened 2021-06-12
Numberrange view: Various minor improvements

BatJan:
### Prerequisites

### Description
I noticed that the "Amount" label did not reference the first input field, so for the sake of consistency I've done some modifications for this to happen, which also improves accessibility.

1. Added an "id" to the first input mapping it to the `model.alias`
2. Added a visually hidden label for the "max input" field

In addition to the above I have also changed the displaying of the validation messages so they appear below the input fields instead of both appearing to the right of the last input field. I think it provides a nicer visual experience even though I know it would probably be rare that people will ever see these messages anyway.

I've added `aria-describedby` to the error messages so they're referencing the id of the input fields their messages belong to as well.

Finally I've mapped an existing dictionary key for the validation messages and added a new one for the "maximum amount" label.

You can see the differences in the GIF's below

**Before**
numberrange-before

**After**
numberrange-after


nul800sebastiaan:
Lovely, thanks @BatJan!

PR #10453, opened 2021-06-12
Textstring prevalue editor view: Set id attribute

BatJan:
### Prerequisites

### Description
In this tiny PR I have added the `id` attribute setting the value to the `model.alias` so click the associated label will behave as expected. The effect of the change can be seen in the GIF below.

textstringprevalue-after


nathanwoulfe (approved):
Cheers @batjan, looks good to me 😃

PR #10454, opened 2021-06-12
Change MultiUrlPicker config and property editor

BatJan:
### Prerequisites

### Description
I think the changes I've made in this PR might be breaking or perhaps we need to add a set of migrations to ensure things still work when people upgrade? Either way I trust someone to guide me about this 😅 Please also let me know if there is something I've overlooked and need to change accordingly as well - I've only tested setting up new multi url property editors inside the backoffice and they appear to be working just fine after these changes though.

The reason for the change is because I was thinking it would be nice to align the UI and overall developer experience so min/max is setup in the same fashion whether we use Media Picker v3, Block List editor or other editors where we have min/max configuration.

You can have a look at the before/after images below to see the changed UI these changes result in - I'm going to mark the PR as a draft for now until we're certain this won't mean trouble when/if this change is merged 😄 

**Before**
multi-url-picker-config-before

**After**
multi-url-picker-config-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/multiurlpicker/multiurlpicker.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/multiurlpicker/multiurlpicker.html
@@ -17,7 +17,7 @@
             </umb-node-preview>
         </div>
 
-        <button ng-show="!model.config.maxNumber || renderModel.length < model.config.maxNumber"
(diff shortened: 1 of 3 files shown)


BatJan:
@nul800sebastiaan Would it be better to make this PR on the v9dev branch? I'm not sure about the impact of the proposed change as mentioned in the initial comment so might be better suited for the major release instead?

PR #10455, opened 2021-06-13
HTML Symantic error UL > LI >LI changes to UL > LI > UL > LI

lucasmichaelsengorm:
I spottede while playing around with MultiUrlPicker that the Link overlayer had a HTML symantic error.
While in the big picture is does not matter, but my screen reader, did a little hike up on over it.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/tree/umb-tree.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/tree/umb-tree.html
@@ -24,7 +24,7 @@
 
         </div>
 
-        <ul>
+        <ul ng-if="tree.root.children.length > 0">
             <umb-tree-item class="umb-animated"
                            ng-repeat="child in tree.root.children"
                            enablelistviewexpand="{{enablelistviewexpand}}"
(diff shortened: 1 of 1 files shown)


src/Umbraco.Web.UI.Client/src/views/components/tree/umb-tree.html
```diff
@@ -13,26 +13,28 @@ <h5>
...
-                       ng-repeat="child in tree.root.children"
-                       enablelistviewexpand="{{enablelistviewexpand}}"
-                       node="child"
-                       current-node="currentNode"
-                       tree="this"
-                       is-dialog="isdialog"
-                       section="{{section}}">
-        </umb-tree-item>
+
+        <ul>
```
bjarnef, line 27: It should probably also include `ng-if="tree.root.children.length > 0` to ensure it doesn't render an empty `<ul>` without `<li>` elements.

lucasmichaelsengorm: @bjarnef I agree on this, I have made a condition on this, on both ul elements, to ensurce is does not render if empty

nathanwoulfe (approved):
@lucasmichaelsengorm this all looks good to me - will let the test run complete and will merge when it's all green (which I fully expect it to be).

Awesome to see another first-time contributor too! We'd love to decorate your profile on our.umbraco.com, if you could let me know where to find it (assuming you have one). 

@nul800sebastiaan, one contrib badge, please and thank you!

PR #10460, opened 2021-06-14
Run `npm update caniuse-lite`

warrenbuckley:
Powershell build failing due to an npm issue with `caniuse-lite` package
Only updated caniuse-lite with `Run `npm update caniuse-lite`

image

image


nul800sebastiaan:
Will be merged up from v8.

PR #10473, opened 2021-06-15
V9: Fix issue with recurring services that executes too often

bergmania:
This PR fixes two small issues:
- Fix exception in ReportSiteTask.cs, when running multiple times..
  - The `s_httpClient.BaseAddress` and `s_httpClient.Timeout` cannot be changed after it has executed a request (This is the case every day)
- Fix issue with recurring tasks that was executed too often.
  - Due to a recent fix for not executing long running tasks in parallel, another bug was introduced. So basically we executed the recurring jobs after `delay`, and then again after `delay` instead of `period`.

### DIFF
--- a/src/Umbraco.Infrastructure/HostedServices/ReportSiteTask.cs
+++ b/src/Umbraco.Infrastructure/HostedServices/ReportSiteTask.cs
@@ -58,12 +58,14 @@ namespace Umbraco.Cms.Infrastructure.HostedServices
                     // Send data to LIVE telemetry
                     s_httpClient.BaseAddress = new Uri("https://telemetry.umbraco.com/");
 
+                    // Set a low timeout - no need to use a larger default timeout for this POST request
+                    s_httpClient.Timeout = new TimeSpan(0, 0, 1);
+
 #if DEBUG
                     // Send data to DEBUG telemetry service
                     s_httpClient.BaseAddress = new Uri("https://telemetry.rainbowsrock.net/");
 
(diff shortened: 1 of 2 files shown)


src/Umbraco.Infrastructure/HostedServices/ReportSiteTask.cs
```diff
@@ -52,13 +52,22 @@ public override async Task PerformExecuteAsync(object state)
...
+                    s_httpClient.BaseAddress = new Uri("https://telemetry.umbraco.com/");
 
 #if DEBUG
-                // Send data to DEBUG telemetry service
-                s_httpClient.BaseAddress = new Uri("https://telemetry.rainbowsrock.net/");
+                    // Send data to DEBUG telemetry service
+                    s_httpClient.BaseAddress = new Uri("https://telemetry.rainbowsrock.net/");
+
+                    // Set a low timeout - no need to use a larger default timeout for this POST request
+                    s_httpClient.Timeout = new TimeSpan(0, 0, 1);
```
AndyButland, line 66: Looks like you are only setting the timeout now in "debug", whereas previously it was looking to set this for all requests.

bergmania: good spot..

warrenbuckley:
Code looks good to me, but looks like @AndyButland was on this as well...
Waiting for CI to go green then will merge this in

bergmania:
The issue with the build is fixed in `v9/dev`

PR #10476, opened 2021-06-15
Made the AppendQueryStringToUrl method public to allow use from packages.

AndyButland:
This is an internal method that we use in Deploy, so should make public.  But I've only noticed the problem whilst experimenting further with the CI/CD pipeline, and specifically renaming the `Umbraco.Deploy` project to `Umbraco.Deploy.Core`. 

I can see why - it's as we've got an `AssemblyInfo.cs` file still that's making certain classes accessible outside of the core solution.  And which I don't think we should have or need... maybe it came back after a merge?

Anyway, I have:

- Removed the `AssemblyInfo,cs` file
- Made the method we need in Deploy public

And once merged I'll update Forms and Deploy to reference the latest nightly and make sure there aren't any more that we've inadvertently made public only to these known HQ assemblies.


bergmania:
Seems fine :)

PR #10480, opened 2021-06-16
Rename server roles, master = SchedulingPublisher and replica = Subscriber

Jeavon:
Following on from discussions on #9088 I propose renaming the server roles in v9 to the following:

Master = Scheduler - this role is responsible for scheduled publishing and background tasks
Replica = Website - there is no replication happening, this role is used for serving the website (front end) 

There are other potential role names suggested in #9088 but I think these are still the best at describing what the roles actually do


ronaldbarendse:
As per comments on the linked issue, naming shouldn't be based on the usage of the server (like backoffice/front-end/website), but the role it is assigned.

Renaming `Master` to `Main` makes sense and recently a common/inclusive thing to do 😉 This also aligns with having the Main server also being the one that locks the MainDom: https://our.umbraco.com/Documentation/Fundamentals/Setup/server-setup/load-balancing/flexible-advanced#explicit-master-scheduling-server.

Jeavon:
IMO, "Main" isn't a good descripiton of the role either, this role is scheduled publishing and running background tasks. Did someone say naming was hard?

bergmania:
Naming is hard 🙈 .. IMO we should in general avoid introducing too many name changes between v8 and v9. Even when the look simple, it will be more breaking changes, and we would have to revisit more documentation.

marcemarc:
I think Master / Replica is a good description of the setup, it's just we don't want to use the word Master.

Just changing Master to Main, (is what you'd do with a branch) but having Main / Replica makes no sense as a pairing - a Replica isn't a copy of a 'Main'. 

When teaching this subject to new devs when you mention 'Main' the immediate thought process is the 'Main' one is the one that the public visit, the scalable Main servers, that all the traffic goes to - where we're trying to imply that 'Main' is actually the Backoffice non-scalable one  'not the main one'....

I like Scheduler, in that as Jeavon says it describes what the difference is... but then 'Website' doesn't give too heavier scent as to that this is the 'Scalable Instance' -

So literally you could have 'Scheduler' and 'Scalable Instance' as your Role names?

I get Shannon's point that one day the backoffice will be potentially scalable ... it nearly is now... but in that 'future scaled out' scenario - which instance does the scheduling? eg will it be clever enough to not reside as one instance's responsibility...? eg wil there always be the notion of a scheduler...

I get Ronald's concern that by naming server Roles 'after what they do' we'll end up with a Role called 'Scheduler' that doesn't do Scheduling, or it's the Scheduling server but that also does unknown task x, because it is used as the backoffice server that is nothing to do with 'scheduling'...

I've been reading around these patterns trying to find something that fits, and it's sort of a Leader, Follower pattern - and we have 'election' and in an 'election' you elect a leader... so it fits quite well - and the scalable instances follow the leader, but it doesn't really map to our domain - Leader to me doesn't give a hint of 'that's the backoffice one'.

but related to that pattern there is also

Publisher - Subscriber

Would that work?

What I quite like about that is Publisher has a scent which hints at 'backoffice' (because people publish from there) AND scheduling. Someone who is a publisher has responsibility for getting things out there, and the other instances 'subscribe' to the cache instructions...  it reinforces, one Publisher - multiple subscribers...

so it half suits naming the instances after what they do, but also half suits naming after the servers general relationship to each other...

... but I only publish the thinking, to help move the debate on, and usually, Shannon now tells me why that's stupid :-P - and that then leads us on to a better thing!

Agree with the fact we've renamed a few things v8->v9 - and renaming causes confusion when switching and extra documentation, (but also major versions are the only opportunity to rename things - so if we don't do it now then when) - so we should be careful, eg but I think there is that underlying driver here, which is to remove 'Master', but without ending up with terms like 'Main' which don't help hint at the understanding of the configuration. I think also it isn't a change that affects heavily implementation, more hosting server configuration... which I think will change for .net core app a little anyway? Think Jeavon and I would update the docs for you!

ronaldbarendse:
@marcemarc Agreed, just renaming for the sake of it (like changing `Master` to `Main`) won't make it better for anyone 😄 And like @bergmania and you mention, would mean a lot of documentation updates, additional manual steps to migrate from v8 to v9 and probably even more confusion about how to correctly configure load-balancing 😉 

We had an internal discussion where I suggested to use similar terms as used with SQL replication, which would actually make sense (and allow for a future load-balanced back-office):
- `Distributor`: the current `Master` role that's responsible for scheduling, refreshing distributed caches, etc. The distributer can be considered as the publisher and distributer at once, which is similar to running a single back-office instance (and would currently be the only supported configuration).
- `Publisher`: responsible for publishing content (but not distributing these). This might become a separate role in the future to allow load-balancing the back-office.
- `Subscriber`: the current `Replica` role that only subscribes to content/cache instructions and is responsible for rendering the front-end (and can be load-balanced).

This aligns with your patterns about having a separate scheduling instance and the publisher/subscriber roles 😇

marcemarc:
@ronaldbarendse thanks for thinking about this - much better than 'main'!!

The only thing I'd say, is I sort of discounted 'Distributor' as an option - because I didn't think anything was being delivered or distributed to the other services by the main scheduling/publishing instance? to me it implies 'pushing out', when the 'subscribers' 'pull' instructions rather than receive them? 

and the Legacy version of Umbraco Load Balancing, which did use to reach out and ping Load Balanced instances to update their cache used to use the configuration of 'DistributedCalls', so was worried about backwards configuration confusion.

But maybe that term makes more sense if in the future  the roles of the current 'master' main are split...

eg if for now it became

Publisher - the current Master role that's responsible for scheduling, refreshing distributed caches, etc. The distributer can be considered as the publisher and distributer at once, which is similar to running a single back-office instance (and would currently be the only supported configuration).
Subscriber - the current Replica role that only subscribes to content/cache instructions and is responsible for rendering the front-end (and can be load-balanced).

But I'm only mentioning this in a 'this is why I didn't mention distributor before' context.

I think the suggestions are infinitely more teachable, and understandable than what we have now, but it would be good to get @Jeavon 's thoughts on this, given it's his PR, and he has tons more insights on this, authoring the LB training course for one!

nathanwoulfe:
Late to the party, but FWIW I'd be going for Publisher and Subscriber. 

From a truly semantic, technical position these may not be entirely correct, but they make really clear sense as a pair. Pub-sub is a common pattern and reasonably familiar terminology. 

If I was including Distributor, I'd be swapping the Publisher and Distributor roles from what @ronaldbarendse suggested. 

Why? Take traditional publishing as an analogy - the publisher produced a newspaper (remember those?), which was literally distributed to subscribers.

ronaldbarendse:
@marcemarc @nathanwoulfe Maybe this overview from the Microsoft documentation explains it more clearly (it even makes the same metaphor to magazines 😇): https://docs.microsoft.com/en-us/sql/relational-databases/replication/publish/replication-publishing-model-overview.

Jeavon:
I think “Subscriber” is a great improvement on “Replica” and would be explainable in docs and training.

The current Master (Master scheduler) role is responsible for scheduled tasks and scheduled publishing (and keep alive) it is not necessarily responsible for creating cache instructions, so personally, I don’t think `Distributor` works for this.

So what role is responsible for running background tasks and scheduled publishing?  In the future when scaling of the back office is possible it might be necessary to have 2 back-office roles….

Based on these discussions my current combo would be:

"Scheduling Publisher" – replaces current “Master Scheduler”
"Publisher" (future role) – a new role for scalable back office, doesn’t run background tasks/scheduled publishing
"Subscriber" – replaces current “Replica”

Currently, any role can create cache instructions, perhaps this should change in the future so that only back-office roles can create cache instructions and even maybe restrict login to the back-office to these roles.

nul800sebastiaan:
Thanks all for the discussion! We gave this another think at HQ and tend to agree with @Jeavon's expertise here. Maybe some terms are a bit more known elsewhere but we also feel that our setup here is probably a bit more specialized. 

The current proposal looks good and pragmatic, and as mentioned we can always add a new name to the list in the eventuality that the backoffice at some point can be load-balanced. 

@marcemarc can I then be a bit cheeky and ask you to give us a hand in updating the docs for v9?

PR #10485, opened 2021-06-16
RelationType parent/child type are stored in the wrong order

inetzo:
When creating a new relationtype, the parent and child type are stored in reverse order (parent- and childtype are switched).

### Prerequisites

### Description
When creating a new relation (via the backoffice), the parent and the child-type (for example Document => Member) are stored in wrong order (Member => Document).


nul800sebastiaan:
Nice catch @inetzo - thanks! 👍

PR #10486, opened 2021-06-16
Adjust the runtime state and keep disabling unattended package migrations simple

Shazwazza:
__#10477 needs to be merged first__

Removes the runtime PackageMigrations state, the state is just run if unattended migrations are disabled. This means it's quite easy to disable unattended package migrations. Which could be done by config or dynamically (i.e. Deploy based on an environment if necessary). If unattended migrations are disabled, the state is just run. And then we'll allow for running the migrations via the package section manually if people want.


src/Umbraco.Core/Extensions/XmlExtensions.cs
```diff
@@ -234,6 +234,33 @@ public static XElement ToXElement(this XmlNode xmlElement)
...
+            {
+                throw new InvalidOperationException($"{attributeName} not found in xml");
+            }
+
+            if (xml.Attribute(attributeName) == null)
+            {
+                throw new InvalidOperationException($"{attributeName} not found in xml");
+            }
+
+            var val = xml.Attribute(attributeName).Value;
```
bergmania, line 254: Think we should do these check in one go (so we do not have `throw new InvalidOperationException($"{attributeName} not found in xml");` duplicated), and we cloud save `xml.Attribute(attributeName)` in a variable instead of executing it twice.

src/Umbraco.Infrastructure/Runtime/RuntimeState.cs
```diff
@@ -144,9 +144,20 @@ public void DetermineRuntimeLevel()
...
+                    Level = RuntimeLevel.Run;
+                    
+                    if (_unattendedSettings.Value.PackageMigrationsUnattended)
+                    {
+                        _logger.LogDebug("Package migrations need to execute.");
+                        Reason = RuntimeLevelReason.UpgradePackageMigrations;
+                    }
+                    else
+                    {
+                        _logger.LogDebug("Package migrations need to execute but unattended package migrations is disabled. They will need to be run from the back office.");
```
bergmania, line 158: I think this should be logged as info.

bergmania (approved):
Only small things, that I will change and merge..

PR #10496, opened 2021-06-18
Fixes clashing aliases when new mediatypes had been installed

nul800sebastiaan:
From #10483

> I just started a new 8.14 project and went to create a new "Article" doctype but I couldn't because there's already a media type with that name/alias.

I've updated the aliases so the chance of them clashing should be very low for new installs from 8.14.1 onwards.

I kept with what looked like the biggest "convention" in these constants and spelled out `umbraco` in full. 

Luckily they are constants so changing them here, will lead to changes everywhere else where it's needed.

So what has changed?

## If you started on 8.14.0

- If you have an existing 8.14.0 site then the aliases will continue to be `video`, `audio`, `article`, `vectorGraphics`
- If you're querying on 8.1.4.0 using `Constants.Conventions.MediaTypes.*` (so: `VideoAlias`, `ArticleAlias`, `AudioAlias` and `VectorGraphicsAlias`) then nothing changes, the alias gets returned. 
- Upgrading to 8.14.1 doesn't change anything for you, the above will give the same results as before

## If you started on 8.14.1

- If you start fresh from 8.14.1, the aliases for the mediatypes Video, Article, Audio and Vector Graphics (SVG) will become: `umbracoMediaVideo`, `umbracoMediaArticle`, `umbracoMediaAudio`, `umbracoMediaVectorGraphics` instead of `video`, `audio`, `article`, `vectorGraphics`
- If you start fresh from 8.14.1 you will query using the new constants that are suffixed with `Alias` (so: `Constants.Conventions.MediaTypes.VideoAlias`, `Constants.Conventions.MediaTypes.ArticleAlias`, `Constants.Conventions.MediaTypes.AudioAlias` and `Constants.Conventions.MediaTypes.VectorGraphicsAlias`)


AndyButland:
This will set the name of the new created content types as well (as you've realised).

Looks like you'll need to add new constants, so we have them for the name and the alias.  And then in `DatabaseDataCreator` keep using the ones you have for the alias [here](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Migrations/Install/DatabaseDataCreator.cs#L174).

But use the new ones for the names [here](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Migrations/Install/DatabaseDataCreator.cs#L130).

nul800sebastiaan:
Thanks @AndyButland I realized that but couldn't find a way to add the alias to `NodeDto`, will try to change to `ContentTypeDto` 👍

nul800sebastiaan:
This is amended now, so I've kept the old constants and they are now the alias (people might already be querying using these constants) and then I've added the `Name` aliases.

And now I've just realized it's a breaking change, since if you're already querying with them.. it will return something different than before! 😅 so that needs to be inverted.. :)

src/Umbraco.Core/Constants-Conventions.cs
```diff
@@ -118,25 +118,45 @@ public static class MediaTypes
...
+
+                /// <summary>
+                /// MediaType name for an article.
+                /// </summary>
+                public const string ArticleName = "Article";
+
+                /// <summary>
+                /// MediaType name for vector graphics.
+                /// </summary>
+                public const string VectorGraphicsName = "VectorGraphics";
```
AndyButland, line 139: This name could be "friendlier", in the sense that it's name and not an alias, so can have spaces and punctuation.  Previously it was "Vector Graphics (SVG)".

nul800sebastiaan: Ah! That's why it was hardcoded.. Fixed that up as well 👍

rasmusjp:
I guess this is going to be confusing for projects crated on 8.14 that is then upgraded 🙈 at least if you use the constants to query media type. I mean you now need to know why you are using one constant over the other even if you are running the samee Umbraco version. But I guess there aren't any good alternative to this.

nul800sebastiaan:
If you upgrade nothing changes. This is only relevant for new installs.

nul800sebastiaan:
The only confusion will be if you then start an 8.14.1 project and you need to use `VideoAlias` instead of `Video`. There's no good way around it indeed, luckily not that many people are using 8.14 yet so, it will be okay. :)

nul800sebastiaan:
Boo.. hardcoded strings in the unit tests.. updated now :)

rasmusjp:
Yep, that's what I was trying to say 😄 it could also be confusing that the comment now says `MediaType name for video` on the "old" constants, but it's still the alias for projects created on 8.14. But I think if the change it's explained in the release notes, you could rename the aliases your self and update the code to use the new `VideoAlias` constant instead, to align it with projects created on 8.14.1+

nul800sebastiaan:
Good point, updated this PR description to make it clearer.

warrenbuckley:
Code changes look fine
Will test

warrenbuckley (approved):

PR #10498, opened 2021-06-18
Package migrations back office updates

Shazwazza:
If you disable unattended package migrations, you can then run the migrations in the back office. 

```json
"Umbraco": {
    "CMS": {
      "Unattended": {
        "PackageMigrationsUnattended":  false
      }
    }
}
```  

Then if you create a new AutomaticPackageMigrationPlan (or change an existing package.manifest file) and then run the site, the plan will not execute automatically. If you go to the package section it will show the package there under installed and you will be able to run the migration there. You can only run pending migrations, else the button will not show up.

This also adds the packageView and name to the package.manifest. So if you create a folder: /App_Plugins/MyPackage and put these files there

package.manifest

```json
{
    "name": "MyPackage",
    "packageView" :  "~/App_Plugins/MyPackage/my-package.html"
}

```

my-package.html

```html
<div>
  Hello world.
</div>
```

If you load the back office, that package will show up there with a Package Options button to go to the view.

Now if you want to have your package plan and package.manifest 'linked' as one package, they just have to have the same name. 

Since it is possible to have multiple plans per package, this PR also adds the ability to link a plan with a package name. By default the ctor overloads for `AutomaticPackageMigrationPlan` and `PackageMigrationPlan` will just take a `name` parameter which should be the package name. In 99% of cases there will be just one plan per package. But for multiple plans, then you can use the other overload like:

```cs
public class BlogAutomaticPackageMigrationPlan : AutomaticPackageMigrationPlan
{
    public BlogAutomaticPackageMigrationPlan()
        : base("MyPackage", "Blog")
    {
    }
}

public class CRMAutomaticPackageMigrationPlan : AutomaticPackageMigrationPlan
{
    public CRMAutomaticPackageMigrationPlan()
        : base("MyPackage", "CRM")
    {
    }
}
```

and then both plans are linked to the same MyPackage.


src/Umbraco.Core/Manifest/PackageManifest.cs
```diff
@@ -10,6 +10,12 @@ namespace Umbraco.Cms.Core.Manifest
     [DataContract]
     public class PackageManifest
     {
+        [DataMember(Name = "name", IsRequired = true)]
+        public string PackageName { get; set; }
+
+        [DataMember(Name = "packageView", IsRequired = true)]
+        public string PackageView { get; set; }
```
bergmania, line 17: I see no reason these are required?

IMO we should aim for not have this as a breaking change in the `package.manifest`.

src/Umbraco.Infrastructure/Services/Implement/PackagingService.cs
```diff
@@ -95,10 +106,60 @@ public void DeleteCreatedPackage(int id, int userId = Cms.Core.Constants.Securit
...
+                {
+                    installedPackage = new InstalledPackage
+                    {
+                        PackageName = package.PackageName
+                    };
+                    installedPackages.Add(package.PackageName, installedPackage);
+                }
+
+                installedPackage.PackageView = package.PackageView;
+            }
```
bergmania, line 154: If the packagename is not required, we could just use the foldername inside app_plugins. 
So If I have the following :

* `~\App_Plugins\My.Package1\package.manifest`  with the following content
```
{
  "name": "My Package",
  "packageView" :  "~/App_Plugins/My.Package1/my-package.html"
}
```
* `~\App_Plugins\My.Package2\package.manifest`  with the following content
```
{
  "packageView" :  "~/App_Plugins/My.Package2/my-package.html"
}
```
* `~\App_Plugins\My.Package3\package.manifest`  with the following content
```
{
  "name": "My Package 3",
}
```

This would list:
- "My Package" with package options (Name from package.manifest)
- "My.Package2" with package options (Name from folder name)
- "My Package 3" If and only if there exist package migrations (Name from package.manifest)

```suggestion
            // Collect and merge the packages from the manifests
            foreach(PackageManifest package in _manifestParser.GetManifests())
            {
                var packageName = package.PackageName ?? Path.GetFileName(Path.GetDirectoryName(package.Source));
                if (packageName is null)
                {
                    continue;
                }

                if (!installedPackages.TryGetValue(packageName, out InstalledPackage installedPackage))
                {
                    installedPackage = new InstalledPackage
                    {
                        PackageName = packageName
                    };
                    installedPackages.Add(packageName, installedPackage);
                }

                installedPackage.PackageView = package.PackageView;
            }
```

bergmania (changes requested):
Seems to work as expected, but I'm not completely happy with the fact that this will break all existing `package.manifest` files.

Here's is my small suggestions that would also make the tests happy again as far as I can see..

src/Umbraco.Core/Manifest/PackageManifest.cs
Shazwazza: > I see no reason these are required?

No, packageView required is a mistake.

src/Umbraco.Infrastructure/Services/Implement/PackagingService.cs
Shazwazza: But IMO the package name from the directory name should be done at the package manifest parser. I'll update.

Shazwazza:
@bergmania changes are pushed

bergmania (approved):

PR #10503, opened 2021-06-21
Updates to latest smidge and changes how debug assets work

Shazwazza:
This updates to latest Smidge which is now on Nuget and updates how package assets are delivered when in debug mode. This is based off of some discussions like https://github.com/umbraco/Umbraco-CMS/discussions/10469

> Disable composite files for package files when in debug mode. This means that the files will be delivered as-is without being processed but will contain the cache buster value. This is more like how v8 worked. This just makes it simpler for all package source debugging since you'll just be working with the raw files. I realize that the way it is now is not great for various reasons: Even if source maps worked, etc... the minify engine will remove debugging statements like debugger; or possibly even logging statements which is not good for debugging. I'm sure nuglify could be tweaked for debug scenarios but IMO it's not worth the effort. Lets just let devs use their raw files to debug with cache busting enabled and then in non-debug we'll process all package files like we did in v8. This will mean that package devs will still need to test their package in non-debug mode to make sure that the minification/composite processing doesn't have any issues. (this will still be much better than v8 since nuglify is a much smarter js/css processor).

### DIFF
--- a/src/Umbraco.Examine.Lucene/Umbraco.Examine.Lucene.csproj
+++ b/src/Umbraco.Examine.Lucene/Umbraco.Examine.Lucene.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <RootNamespace>Umbraco.Cms.Infrastructure.Examine</RootNamespace>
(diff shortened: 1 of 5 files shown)


bergmania (approved):
Changes look good.

Shazwazza:
I also just pushed a bump for latest Examine on Nuget. But that's just a version change. Do you want me to just merge it?

bergmania:
> I also just pushed a bump for latest Examine on Nuget. But that's just a version change. Do you want me to just merge it?

Yeah, that's fine :)

Shazwazza:
Cool, that's all merged into beta4 branch.

PR #10504, opened 2021-06-21
Convert Media Picker v2 to v3 progressively

nul800sebastiaan:
In order to be easily able to move from Media Picker 2 (MP2) to Media Picker 3 (MP3), this PR helps convert existing values when you switch the datatype. 

In the backoffice, when you switch a datatype from MP2 to MP3 and load any node with this new datatype, we'll detect if the old values were stored (a list of UDIs) and if so, we convert those to the new `MediaWithCrops` type. This allows us to show the previously selected media items in the backoffice. This happens in `MediaPicker3PropertyEditor.ToEditor`. 

Once you save and publish the item with the new MP3 datatype, the converted data will be saved and we won't have to do that conversion again in the future. This way you will be progressively updating your nodes when an editor saves them and it is not needed to do a database migration to convert all the data on all the nodes.

In the frontend, you will need to replace your MP2 querying with MP3 querying. As we can see from the documentation the return type is change, so an example of updating a multiple media picker would be:

- [Media Picker 2](https://our.umbraco.com/documentation/Fundamentals/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker/): `var typedMultiMediaPicker = Model.Value<IEnumerable<IPublishedContent>>("medias");` / `@entry.Url`
- [Media Picker 3](https://our.umbraco.com/documentation/Fundamentals/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/): `var typedMultiMediaPicker = Model.Value<IEnumerable<MediaWithCrops>>("medias");` / `@entry.MediaItem.Url()`

We do the exact same detection and conversion that we did in the backoffice on the frontend, see `MediaPickerWithCropsValueConverter`.

Both of the conversions are ultimately being done in the `JsonConverter` named `MediaWithCropsDtoConverter`.  
A note: in Umbraco 7, we stored picked media items by integer identifiers, this was never the case in v8 and the property value converter in v8 never dealt with integer ids. Therefore I did not implement conversions from int in the `MediaWithCropsDtoConverter` either.

If you want to make it an even simpler conversion and you do not plan to use the "local crops" feature on the MP3 datatype (don't configure any crops) then you can write your own property value converter (PVC) to give you the same data as MP2 gave you (`IPublishedContent`/`IEnumerable<IPublishedContent>`). If you want to add local crops in the future you could probably find a solution with an extension method in your PVC.   
We recommend updating your templates with the new `MediaWithCrops` return type.

_Final note: going back from MP3 to MP2 is not possible if MP3 values have been stored already, this is a forward-only fix._

### DIFF
--- a/src/Umbraco.Core/Serialization/MediaWithCropsDtoConverter.cs
+++ b/src/Umbraco.Core/Serialization/MediaWithCropsDtoConverter.cs
@@ -1,0 +1,73 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Umbraco.Core.PropertyEditors.ValueConverters;
+
+namespace Umbraco.Core.Serialization
+{
+    public class MediaWithCropsDtoConverter : JsonConverter
+    {
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            var mediaWithCropsDto = new List<MediaWithCropsDto>();
+
+            var mediaIds = value.ToString().Split(',');
+            if (mediaIds.Any())
+            {
+                foreach (var mediaId in mediaIds)
+                {
+                    Guid mediaItemKey = default;
+                    if (GuidUdi.TryParse(mediaId, out var guidUdi))
+                    {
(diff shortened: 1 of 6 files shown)


nul800sebastiaan:
Closing in favor of https://github.com/umbraco/Umbraco-CMS/pull/10517

Issue #10508, opened 2021-06-22
v9 - Render Block List partial async


bjarnef:
### Which *exact* Umbraco version are you using?

9.0.0-beta004

### Bug summary

In https://github.com/umbraco/Umbraco-CMS/issues/10507#issuecomment-865886350 I noticed Visual Studio warning about `@Html.Partial()` in Block List default.chtml.

```
@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<Umbraco.Cms.Core.Models.Blocks.BlockListModel>
@{
    if (!Model.Any()) { return; }
}
<div class="umb-block-list">
    @foreach (var block in Model)
    {
        if (block?.ContentUdi == null) { continue; }
        var data = block.Content;

        @Html.Partial("BlockList/Components/" + data.ContentType.Alias, block)
    }
</div>
````

https://github.com/umbraco/Umbraco-CMS/blob/v9/9.0-beta004/src/Umbraco.Web.UI.NetCore/Views/Partials/blocklist/default.cshtml

### Steps to reproduce

I found this: https://stackoverflow.com/a/52723364/1693918

It seems we can use the following instead, which still render the block elements as expected.

```
@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<Umbraco.Cms.Core.Models.Blocks.BlockListModel>
@{
    if (!Model.Any()) { return; }
}
<div class="umb-block-list">
    @foreach (var block in Model)
    {
        if (block?.ContentUdi == null) { continue; }
        var data = block.Content;

        @await Html.PartialAsync("BlockList/Components/" + data.ContentType.Alias, block)
    }
</div>
```

![image](https://user-images.githubusercontent.com/2919859/122927931-a5b40f80-d369-11eb-8bd6-0c36740ebaf2.png)

![image](https://user-images.githubusercontent.com/2919859/122927876-992fb700-d369-11eb-8b03-6fcd48125790.png)

Furthermore it seems the following would also work in .NET Core (the first one is probably the preferred and easiest to handle if migrating a site to Umbraco v9):

```
@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<Umbraco.Cms.Core.Models.Blocks.BlockListModel>
@{
    if (!Model.Any()) { return; }
}
<div class="umb-block-list">
    @foreach (var block in Model)
    {
        if (block?.ContentUdi == null) { continue; }
        var data = block.Content;

[...]

PR #10511, opened 2021-06-22
Render Block List component partial async

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10508

### Description
Currently Visual Studio shows a warning regarding the use of `@Html.Partial()` as it may result in application deadlocks.

It seems we can render the component async in .NET Core:
https://stackoverflow.com/questions/50774702/should-i-change-html-partial-to-html-partialasync-as-visual-studio-suggest/52723364#52723364

### DIFF
--- a/src/Umbraco.Web.UI.NetCore/Views/Partials/blocklist/default.cshtml
+++ b/src/Umbraco.Web.UI.NetCore/Views/Partials/blocklist/default.cshtml
@@ -7,6 +7,7 @@
     {
         if (block?.ContentUdi == null) { continue; }
         var data = block.Content;
-        @Html.Partial("BlockList/Components/" + data.ContentType.Alias, block)
+
+        @await Html.PartialAsync("BlockList/Components/" + data.ContentType.Alias, block)
     }
 </div>


bjarnef:
I guess we can use a similar approach in Grid: https://github.com/umbraco/Umbraco-CMS/blob/v9/9.0-beta004/src/Umbraco.Web.UI.NetCore/Views/Partials/grid/bootstrap3.cshtml#L54

Instead of:

```
<text>@Html.Partial("grid/editors/base", (object)control)</text> 
```

this might work:

```
<text>@await Html.PartialAsync("grid/editors/base", (object)control)</text>
```

or

```
<partial name="grid/editors/base" model="@((object)control)" />
```

In Bootstrap3 templates it does however require function `renderRow` to change to async method.

bergmania:
Looks good and works :)

bjarnef:
@bergmania should this one be marked with `project/v9` and `release/9.0.0-beta005` labels as well?

bergmania:
@bjarnef, normally we only label the releases on issue if there is one. And not both the issue and PR. If there is only a PR, then we label the PR.

Think it is correct now. :)

PR #10512, opened 2021-06-22
Render grid editor partial async

bjarnef:
### Prerequisites

### Description
While working on https://github.com/umbraco/Umbraco-CMS/pull/10511 I also noticed some partials in grid wasn't rendered async, but we did it for macro grid editor.

https://github.com/umbraco/Umbraco-CMS/blob/v9/9.0-beta004/src/Umbraco.Web.UI.NetCore/Views/Partials/grid/editors/macro.cshtml#L12-L14

The grid editors seems to render as expected on frontend. I guess it would also improve the performance a bit in backoffice when rendering previews of many grid editors.

Pf1pVOkhrG

OFXKTPGdvb


bergmania:
Nice work @bjarnef, Thanks alot :)

PR #10516, opened 2021-06-23
Fixes the acceptance tests "Content with contentpicker"

bergmania:
Fixes the acceptance test  "Content with contentpicker" that are failing on each run.


nikolajlauridsen (approved):
Looks good and works fine 👍

PR #10517, opened 2021-06-23
Add the ability to convert Media Picker v2 to v3 progressively

nul800sebastiaan:
In order to be easily able to move from Media Picker 2 (MP2) to Media Picker 3 (MP3), this PR helps convert existing values when you switch the datatype. 

In the backoffice, when you switch a datatype from MP2 to MP3 and load any node with this new datatype, we'll detect if the old values were stored (a list of UDIs) and if so, we convert those to the new `MediaWithCrops` type. This allows us to show the previously selected media items in the backoffice. This happens in `MediaPicker3PropertyEditor.ToEditor`. 

Once you save and publish the item with the new MP3 datatype, the converted data will be saved and we won't have to do that conversion again in the future. This way you will be progressively updating your nodes when an editor saves them and it is not needed to do a database migration to convert all the data on all the nodes.

In the frontend, you will need to replace your MP2 querying with MP3 querying. As we can see from the documentation the return type is change, so an example of updating a multiple media picker would be:

- [Media Picker 2](https://our.umbraco.com/documentation/Fundamentals/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker/): `var typedMultiMediaPicker = Model.Value<IEnumerable<IPublishedContent>>("medias");` / `@entry.Url`
- [Media Picker 3](https://our.umbraco.com/documentation/Fundamentals/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/): `var typedMultiMediaPicker = Model.Value<IEnumerable<MediaWithCrops>>("medias");` / `@entry.MediaItem.Url()`

We do the exact same detection and conversion that we did in the backoffice on the frontend, see `MediaPickerWithCropsValueConverter`.

Both of the conversions are ultimately being done in `MediaPicker3PropertyEditor.MediaPicker3PropertyValueEditor.Deserialize`

A note: in Umbraco 7, we stored picked media items by integer identifiers, this was never the case in v8 and the property value converter in v8 never dealt with integer ids. Therefore we did not implement conversions here either.

If you want to make it an even simpler conversion and you do not plan to use the "local crops" feature on the MP3 datatype (don't configure any crops) then you can write your own property value converter (PVC) to give you the same data as MP2 gave you (`IPublishedContent`/`IEnumerable<IPublishedContent>`). If you want to add local crops in the future you could probably find a solution with an extension method in your PVC.   
We recommend updating your templates with the new `MediaWithCrops` return type.

_Final note: going back from MP3 to MP2 is not possible if MP3 values have been stored already, this is a forward-only fix._


src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerWithCropsValueConverter.cs
```diff
@@ -24,96 +22,58 @@ public MediaPickerWithCropsValueConverter(IPublishedSnapshotAccessor publishedSn
...
-        /// </summary>
         public override bool IsConverter(IPublishedPropertyType propertyType) => propertyType.EditorAlias.Equals(Core.Constants.PropertyEditors.Aliases.MediaPicker3);
 
-        /// <summary>
-        /// Check if the raw JSON value is not an empty array
-        /// </summary>
-        public override bool? IsValue(object value, PropertyValueLevel level) => value?.ToString() != "[]";
+        public override bool? IsValue(object value, PropertyValueLevel level)
+            => value?.ToString() is string stringValue &&
+                stringValue != null &&
```
nul800sebastiaan, line 29:
```suggestion
```

This is getting checked for `null` again on the next line! 😅 and since it `is string` it can not be `null` anwyay.

src/Umbraco.Web/PropertyEditors/MediaPicker3PropertyEditor.cs
```diff
@@ -22,43 +27,103 @@ namespace Umbraco.Web.PropertyEditors
...
                     yield break;
+                }
 
-                var mediaWithCropsDtos = JsonConvert.DeserializeObject<MediaPickerWithCropsValueConverter.MediaWithCropsDto[]>(rawJson);
-
-                foreach (var mediaWithCropsDto in mediaWithCropsDtos)
+                if (!rawJson.DetectIsJson())
+                {
+                    // Old comma seperated UDI format
+                    foreach (var udiStr in rawJson.Split(Constants.CharArrays.Comma))
```
nul800sebastiaan, line 81: I still really would like to throw this in a try-catch in case people have weird values stored, logging the stored value will make it a lot easier to debug these cases which I am sure will come up :-) 

Maybe not a try-catch but at least if the `TryParse` fails we could log what weird value is in there. What do you think @ronaldbarendse ?

nul800sebastiaan (commented):
Left two comments for you to consider @ronaldbarendse but otherwise this is all great, I tested a few different scenarios and it seems like it should all work 🎉

src/Umbraco.Web/PropertyEditors/MediaPicker3PropertyEditor.cs
ronaldbarendse: There wasn't additional logging before and the Media Picker 2 even used `Udi.Parse` (without handling formatting/parsing errors) in the PVC:

https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerValueConverter.cs#L59-L62

Because we can't do anything with incorrectly formatted UDIs, I think it's safe to simply ignore/skip them, just like the `GetReferences` implementation of MP2 does:

https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web/PropertyEditors/MediaPickerPropertyEditor.cs#L48-L52

nul800sebastiaan: @ronaldbarendse Really good point and in fact switching from MP3 back to MP2 already gave me ugly errors indeed, so it's safe to assume people wouldn't have any other data in there, it just would've failed to parse to begin with! 👍

PR #10519, opened 2021-06-23
V9: Fixes issue with "Rebuild" nucache not doing anything

### LINKED ISSUE #10394: Rebuild Database Cache button does nothing

v9.0.0-beta003

### Bug summary

Clicking the "Rebuild" button on the Caches block of the Settings / Published Status section seems to do nothing.
My cache was broken and I could not fix it with this.
As a test, I manually emptied the cmsContentNu table:
```
delete [dbo].[cmsContentNu] 
```
After doing that, clicking the button does not rebuild the table contents

### Specifics

_No response_

### Steps to reproduce

Remove all rows from the cmsContentNu table and run the command

The table is not rebuild and remains empty
The expected / v8 behavior is to rebuild it.

bergmania:
Fixes #10394.

The issue was a couple injected `null`s was handled like "no nothing" instead of "clear all" as the XML docs describe.

Steps to test in the issue #10394..


warrenbuckley:
Works as described

PR #10171, opened 2021-04-22
Don't eagerly acquire distributed (SQL) locks

Shazwazza:
This is a proof of concept so far. Although it works there's one caveat (see below).

Instead we queue the request for distributed locks and execute them only when necessary. This means that if a repository resolves the entity from cache there are no SQL calls made for acquiring locks.

This should save on a large amount of SQL calls being made for locks.

This also improves performance within the Scope object itself. In many cases it was using it's own public `Database` property accessor to access things like `SqlContext` but as you can see the `Database` property performs some logic but there's no point in performing this logic each time. `SqlContext` is exposed directly already and in other places where `Database` was used, we can just use `_database` since in those cases we know that field has been set.

So why isn't this perfect?

You can see the TODO I've added in Scope.cs. This change means that we will acquire any queued locks when Scope.Database is accessed. However, if this occurs:

```cs
var db = Scope.Database;
Scope.ReadLock(123);
db.ExecuteScalar(...)
```

Then the read lock will not execute. However, doing this will execute them:

```cs
Scope.ReadLock(123);
Scope.Database.ExecuteScalar(...)
```

The latter is how all locks are acquired within Umbraco, however it's not ideal that they will not be executed when a db command is trigged.

The better way to achieve these non-eager locks would be to modify the behavior of the UmbracoDatabase instance so that non-eager locks are executed before any DB commands (excluding lock commands). This would require checking that the IDatabaseFactory is of our known type so that we know that the resolved IDatabase supports this scenario. Then internally we would configure that database instance to do the lazy lock execution. The Scope would fallback to eager locking if the IDatabase doesn't support this.

It's certainly possible! I think this could have some very good performance improvements.

---
_This item has been added to our backlog [AB#12561](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/12561)_

PR #10521, opened 2021-06-23
Reduce sql queries when rendering blocklists in the content panel

nikolajlauridsen:
## Description
This is a continuation of #10235, here I try to reduce the number of SQL queries used when rendering blocklists when adding new content. A lot of the queries and time spent came from mapping the empty `Content` to a `ContentItemDisplay` 

## Details 
There were some low hanging fruit namely I: 

* Added a new `Create` overload to the content service which takes a content type instead of a key. Before we would get all the content types through the content type service, only to pass it to `GetEmpty` where we would then call `ContentService.Create` with the content type alias, only for the content service to then get the content type object one more time. Now we just pass the content type to `Create` which then uses it to create new content.
* Changed it so if we map something that's an element type we don't try to get allowed templates and just return an empty dictionary, we know element types doesn't have any allowed templates
* Both `GetActions` and `DetermineIsChildOfListView` uses the parent object, so instead of getting it twice, we get it once in and then pass it to both those methods.

One of the larger areas of improvement was with the `GetActions` part of mapping to `ContentItemDisplay`. Every time we get actions with ` _userService.GetPermissionsForPath(umbracoContext.Security.CurrentUser, path).GetAllPermissions();` it causes an SQL query, this means that for every possible element type in the blocklist we would execute a get query. But since these are element types on the same doctype the permissions will be the same.

What I tried to do to resolve this is added a new method `GetEmpties` which takes an `IEnumberable` of content types and creates an empty `ContentItemDisplay` for each of these. This method is not super nice, but it seems to work quite well. In the method we first try to get the parent once, this could potentially lead to an extra SQL call, but since we'll save 1 per element type it seems worth it, we'll then generate the path (same as in the mapper), and for the very first item, we'll get the permissions with the path and current user, this permission when then be saved and used for all the entries.

We pass the permissions, path and current user to the mapper in the mapper context, within the mapper we then check if the path in the mapper context matches the one generated by the mapper, and if the user is the same, if they are we can safely assume the permissions are the same and just return the permissions stored in the mapper context.

In reality, I don't know it's truly necessary to pass along the path and user and check them since is currently only used for the blocklist where they should always be the same, however it seemed a bit overly risky to just naively say "if the mapper context contains permissions, then just return those" without doing any checks. 

## Tests

I tried doing some testing before and after with the mini profiler, of course this is just running locally on my machine with a local SQL server, but this is what I found.

I have a document type with two block list editors of the same type, each blocklist editor have 5 element types to chose from.

Before this and Elitsa's PR we'd make a request for each element type available and for each blocklist property, so a total of 10 requests:

Call nr | Time in ms | Queries
-- | -- | --
1 | 30.6 | 7
2 | 30.3 | 7
3 | 26.9 | 7
4 | 34.7 | 7
5 | 26.6 | 7
6 | 25.3 | 7
7 | 25.6 | 7
8 | 33.7 | 7
9 | 35.2 | 7
10 | 33.5 | 7
Total | 302.4 | 70

After Elitsa's and this PR we only do two requests, one for each blocklist property

Block nr | Time in ms | Queries
-- | -- | --
1 | 24.5 | 2
2 | 14.5 | 2
Total | 39 | 4

One of these queries are for the SQL lock, the other is for the permissions, it's of course worth mentioning that getting the parent and so on doesn't result in an SQL query since it's cached.

### DIFF
--- a/src/Umbraco.Core/Services/Implement/ContentService.cs
+++ b/src/Umbraco.Core/Services/Implement/ContentService.cs
@@ -186,7 +186,6 @@ namespace Umbraco.Core.Services.Implement
         public IContent Create(string name, int parentId, string contentTypeAlias, int userId = Constants.Security.SuperUserId)
         {
             // TODO: what about culture?
-
             var contentType = GetContentType(contentTypeAlias);
             return Create(name, parentId, contentType, userId);
         }
@@ -200,24 +199,40 @@ namespace Umbraco.Core.Services.Implement
         /// </remarks>
         /// <param name="name">The name of the content object.</param>
         /// <param name="parentId">The identifier of the parent, or -1.</param>
-        /// <param name="contentType">The content type of the content</param>
+        /// <param name="contentType">The content type.</param>
         /// <param name="userId">The optional id of the user creating the content.</param>
         /// <returns>The content object.</returns>
-        public IContent Create(string name, int parentId, IContentType contentType,
-            int userId = Constants.Security.SuperUserId)
+        public IContent Create(string name, int parentId, IContentType contentType, int userId = Constants.Security.SuperUserId)
         {
-            if (contentType is null)
-            {
-                throw new ArgumentException("Content type must be specified", nameof(contentType));
-            }
-
             var parent = parentId > 0 ? GetById(parentId) : null;
             if (parentId > 0 && parent is null)
             {
                 throw new ArgumentException("No content with that id.", nameof(parentId));
             }
 
-            var content = new Content(name, parentId, contentType);
+            return Create(name, parent, contentType, userId);
+        }
+
+        /// <summary>
+        /// Creates an <see cref="IContent"/> object of a specified content type.
+        /// </summary>
+        /// <remarks>This method simply returns a new, non-persisted, IContent without any identity. It
+        /// is intended as a shortcut to creating new content objects that does not invoke a save
+        /// operation against the database.
+        /// </remarks>
+        /// <param name="name">The name of the content object.</param>
+        /// <param name="parent">The parent of the content</param>
+        /// <param name="contentType">The content type of the content</param>
+        /// <param name="userId">The optional id of the user creating the content.</param>
+        /// <returns>The content object.</returns>
+        public IContent Create(string name, IContent parent, IContentType contentType, int userId = Constants.Security.SuperUserId)
+        {
+            if (contentType is null)
+            {
+                throw new ArgumentException("Content type must be specified", nameof(contentType));
+            }
+
+            var content = new Content(name, parent, contentType);
             using (var scope = ScopeProvider.CreateScope())
             {
                 CreateContent(scope, content, userId, false);
--- a/src/Umbraco.Core/Services/Implement/ContentService.cs
+++ b/src/Umbraco.Core/Services/Implement/ContentService.cs
@@ -188,11 +188,34 @@ namespace Umbraco.Core.Services.Implement
             // TODO: what about culture?
 
             var contentType = GetContentType(contentTypeAlias);
-            if (contentType == null)
-                throw new ArgumentException("No content type with that alias.", nameof(contentTypeAlias));
+            return Create(name, parentId, contentType, userId);
+        }
+
+        /// <summary>
+        /// Creates an <see cref="IContent"/> object of a specified content type.
+        /// </summary>
+        /// <remarks>This method simply returns a new, non-persisted, IContent without any identity. It
+        /// is intended as a shortcut to creating new content objects that does not invoke a save
+        /// operation against the database.
+        /// </remarks>
+        /// <param name="name">The name of the content object.</param>
+        /// <param name="parentId">The identifier of the parent, or -1.</param>
+        /// <param name="contentType">The content type of the content</param>
+        /// <param name="userId">The optional id of the user creating the content.</param>
+        /// <returns>The content object.</returns>
+        public IContent Create(string name, int parentId, IContentType contentType,
+            int userId = Constants.Security.SuperUserId)
+        {
+            if (contentType is null)
+            {
+                throw new ArgumentException("Content type must be specified", nameof(contentType));
+            }
+
             var parent = parentId > 0 ? GetById(parentId) : null;
-            if (parentId > 0 && parent == null)
+            if (parentId > 0 && parent is null)
+            {
                 throw new ArgumentException("No content with that id.", nameof(parentId));
+            }
 
             var content = new Content(name, parentId, contentType);
             using (var scope = ScopeProvider.CreateScope())
@@ -1088,7 +1111,7 @@ namespace Umbraco.Core.Services.Implement
         /// <remarks>
         /// <para>
         /// Business logic cases such: as unpublishing a mandatory culture, or unpublishing the last culture, checking for pending scheduled publishing, etc... is dealt with in this method.
-        /// There is quite a lot of cases to take into account along with logic that needs to deal with scheduled saving/publishing, branch saving/publishing, etc...       
+        /// There is quite a lot of cases to take into account along with logic that needs to deal with scheduled saving/publishing, branch saving/publishing, etc...
         /// </para>
         /// </remarks>
         private PublishResult CommitDocumentChangesInternal(IScope scope, IContent content,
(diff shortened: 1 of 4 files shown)


Shazwazza:
Also note that the SQL calls for obtaining a lock can be optimized and a PR is already created here: https://github.com/umbraco/Umbraco-CMS/pull/10171

src/Umbraco.Core/Services/Implement/ContentService.cs
```diff
@@ -188,11 +188,34 @@ public IContent Create(string name, int parentId, string contentTypeAlias, int u
...
+        /// <remarks>This method simply returns a new, non-persisted, IContent without any identity. It
+        /// is intended as a shortcut to creating new content objects that does not invoke a save
+        /// operation against the database.
+        /// </remarks>
+        /// <param name="name">The name of the content object.</param>
+        /// <param name="parentId">The identifier of the parent, or -1.</param>
+        /// <param name="contentType">The content type of the content</param>
+        /// <param name="userId">The optional id of the user creating the content.</param>
+        /// <returns>The content object.</returns>
+        public IContent Create(string name, int parentId, IContentType contentType,
```
Shazwazza, line 206: I wonder if this would be better to supply the IContent as the parent (it could be null) instead of relying on this method to do the lookup. I understand that it 'could' be cached but that is not always the case. This method could be used for better performance to (for example) create a bunch of content under a single parent and not have to rely on the service to keep looking it up. Cache isn't permanent so we can't always rely on it.

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -388,48 +388,100 @@ public ContentItemDisplay GetEmptyByKey(Guid contentTypeKey, int parentId)
...
+            // Since this is called with parent ID it's safe to assume that the parent is the same for all the content types.
+            var parent = parentId > 0 ? Services.ContentService.GetById(parentId) : null;
+            EntityPermissionSet permissions = null;
 
             foreach (var contentType in contentTypes)
             {
-                if (contentType is null)
+                var emptyContent = Services.ContentService.Create("", parentId, contentType, userId);
+
+                if (emptyContent is null)
```
Shazwazza, line 432: I don't think the Create method would ever return a null value? It's a factory method so in theory should never return null

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -388,48 +388,100 @@ public ContentItemDisplay GetEmptyByKey(Guid contentTypeKey, int parentId)
...
                 {
                     throw new HttpResponseException(HttpStatusCode.NotFound);
                 }
 
-                result.Add(contentType.Key, GetEmpty(contentType, parentId));
+                // Get the path for the content, we get this for each just in case some content has Identity and a different path
+                // Since if that's the case we might not be able to re-use the permissions.
+                string path;
+                if (emptyContent.HasIdentity)
```
Shazwazza, line 440: Does the IContentService.Create method ever return a content with an identity? It's a factory method to create a brand new content item so not sure this will ever be the case?

src/Umbraco.Web/Editors/ContentController.cs
```diff
@@ -388,48 +388,100 @@ public ContentItemDisplay GetEmptyByKey(Guid contentTypeKey, int parentId)
...
+                }
+                else
+                {
+                    path = parent == null ? "-1" : parent.Path;
+                }
+
+                // Only assign permissions once, the idea is that we'll be able to re-use the same permissions most of the time.
+                if (permissions is null)
+                {
+                    permissions = Services.UserService.GetPermissionsForPath(currentUser, path);
```
Shazwazza, line 452: Would it be simpler if permissions was a `Dictionary<string, EntityPermissionCollection>` and you would just lookup permissions for any unique path?

src/Umbraco.Web/Models/Mapping/ContentMapDefinition.cs
```diff
@@ -86,7 +86,10 @@ private static void Map(IContent source, ContentPropertyCollectionDto target, Ma
         // Umbraco.Code.MapAll -AllowPreview -Errors -PersistedContent
         private void Map(IContent source, ContentItemDisplay target, MapperContext context)
         {
-            target.AllowedActions = GetActions(source);
+            // Both GetActions and DetermineIsChildOfListView use parent, so get it once here
+            var parent = _contentService.GetParent(source);
```
Shazwazza, line 90: In theory this will all ready be looked up so we can avoid re-looking it back up again if it's in the context. In the content service the parent is explicitly looked up.

src/Umbraco.Web/Models/Mapping/ContentMapDefinition.cs
```diff
@@ -169,10 +172,24 @@ private IEnumerable<string> GetActions(IContent source)
...
+            // A bit of a mess, but we need to ensure that all the required values are here AND that they're the right type.
+            if (context.Items.TryGetValue("CurrentUser", out var userObject) &&
+                context.Items.TryGetValue("Path", out var pathObject) &&
+                context.Items.TryGetValue("Permissions", out var permissionsObject) &&
+                userObject is IUser currentUser &&
+                pathObject is string contextPath &&
+                permissionsObject is EntityPermissionSet permissions)
+            {
+                // Both the path and the user is the same, it's safe to assume that the permissions are the same as well.
+                if (contextPath.Equals(path) && umbracoContext.Security.CurrentUser.Id == currentUser.Id)
```
Shazwazza, line 187: Mapping doesn't always occur on a web request and in that case umbracoContext will be null.

Shazwazza (changes requested):
Great stuff! have left a few comments and questions.

src/Umbraco.Web/Editors/ContentController.cs
nikolajlauridsen: Ahh, that's very true, the Create method even calls `CreateContent` with `withIdentity` set to false, I can't believe I missed this 🙈

src/Umbraco.Core/Services/Implement/ContentService.cs
nikolajlauridsen: Ahh that's a very good point, however, `GetEmpty` and `GetEmptyByKey` only creates a single content, so they just pass in the parent ID, so I've added an additional overload, so we now have: 

* Create content with  parent ID and content type alias
* Create content with parent ID and content type
* Create content with parent and content type

So we cover all cases 😄

src/Umbraco.Web/Models/Mapping/ContentMapDefinition.cs
nikolajlauridsen: Yeah, there's also a comment about in the method, at the very top though we check if umbracoContext is null, and if it is we just return `Enumerable.Empty<string>();` so it's handled before we get to that part.

src/Umbraco.Web/Models/Mapping/ContentMapDefinition.cs
nikolajlauridsen: Good idea, I've added the parent to the context, and look there before trying to use the service.

src/Umbraco.Web/Editors/ContentController.cs
nikolajlauridsen: Hmm I don't think it would necessarily be simpler, looking at it now, the path used to get the permissions is the parent path, and since the parent here is always the same, we actually don't need to keep creating the path, so we don't need to check if permission is null, we can just get the path and permissions once before the loop and use that for all of them.

However, I think using the `Dictionary<string, EntityPermissionCollection>` might be a good idea, mainly because it would make it easier to make optimisations in future if, for instance, we wanted to map multiple IContent where some have a different path then the dictionary approach would allow for that, whereas the current approach really only allows for a single path.

src/Umbraco.Core/Services/Implement/ContentService.cs
nikolajlauridsen: Hmm I'm sorry but I just had a closer look at this after the tests were failing, and I don't think we can do this in the end, it turns out that the constructor differs for the `Content` depending on whether you use an ID or and `IContent`, if we use `IContent` i must not be null. 

This becomes problematic when `Create` is called with an ID of a parent that does not exists, since `IContent parent` will be null. I did consider if we could just assume that if the parent was null that the parent ID would be -1 since that's root, and call the constructor with -1 as the parent ID, however when a parent is null the ID is not always -1, for instance with the blocklist we send the ID as -20. Since the `IContent parent` would be null we have no way of knowing this and would change the ID to -1, potentially messing it up for some if they expect the ID to be -20 for some reason.

nikolajlauridsen:
I've gone through the suggestions, and think this should be ready to merge now 👍

Issue #10515, opened 2021-06-23
Richtext table missing color plugin


bjarnef:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

8.13.1

### Bug summary

By default when table is enabled in richtext editor, the table properties are available to set e.g. background or border color.

![image](https://user-images.githubusercontent.com/2919859/123052679-60442080-d403-11eb-82b6-9ad1dbab53e4.png)

However it seems this doesn't use the color plugin which is shown in the demo here:

https://www.tiny.cloud/docs-4x/demo/full-featured/
![image](https://user-images.githubusercontent.com/2919859/123053150-e3fe0d00-d403-11eb-937b-48f3627d9585.png)

### Steps to reproduce

- Enable table in a richtext datatype instance
- Place cursor in table and click table properties
- Under "Anvanced" tab notice that not color picker is available

It does seem the HTML as the following next to e.g. border color text input: `<i id="mceu_356-status" class="mce-status mce-ico" style="display: none;"></i>`

Removing `display:none` does show the "button" although it doesn't do anything:

![image](https://user-images.githubusercontent.com/2919859/123053811-92a24d80-d404-11eb-9ba0-c8f053191458.png)

bjarnef:
Okay, a workaround is to add the following in `tinyMceConfig.config` to enable color picker plugin, but not necessarily enable `forecolor` and `backcolor` for the richtext editor datatype instance: https://www.tiny.cloud/docs-4x/plugins/colorpicker/

```
<plugin>colorpicker</plugin>
```

I think this should be added by default, so the color picker works on table properties.

![image](https://user-images.githubusercontent.com/2919859/123055103-d3e72d00-d405-11eb-8ed0-f22b60176dd4.png)

umbrabot:
Hi @bjarnef,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
@nul800sebastiaan I wonder if we should add this in the config file or add it here?
[...]

PR #10523, opened 2021-06-23
Use color picker plugin when plugins includes table plugin

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10515

### Description
When table plugin is used in rich text editor, it has some advanced table tools, e.g. set set background and border colors.
It was possible to specify the color code / name, but not editor friendly.

The `table` plugin uses the `colorpicker` plugin, so we can add this, when plugins includes the table plugin.

**Before**

zZFYXwuD2B

**After**

7cbJChwFQa

Not sure why the color picker dialog shows the title "color" and not "Color" like the demo here:
https://www.tiny.cloud/docs-4x/demo/full-featured/

Actually the basic demo here also doesn't show color picker under table advances properties, but I think it makes sense to include this, when table plugin is used.
https://www.tiny.cloud/docs-4x/demo/basic-example/

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
@@ -352,18 +352,22 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
                     return plugin.name;
                 });
 
-                //plugins that must always be active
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Cheers @bjarnef!

PR #10524, opened 2021-06-24
Fixes #10402 - GetCropUrl Url extension has duplicate signatures and results in error

### LINKED ISSUE #10402: GetCropUrl Url extension has duplicate signatures and results in error

8.14.0

### Bug summary

After upgrading to 8.14.0, some of my image crops were not working. After looking into it, I believe I have found the issue.

In my code I have the following: @Url.GetCropUrl(item.Property, cropAlias:"Featured", width: 1200)

And when trying to compile, I get "This call is ambiguous between the following methods or properties".

### Specifics

_No response_

### Steps to reproduce

This seems to also be the case for a new install of 8.14.0

- Add image cropper property to document, add a named crop
- This in code results in an error @Url.GetCropUrl(Model.TestImage, cropAlias: "Test", width:200)

Note: if I don't the parameter name, such as @Url.GetCropUrl(Model.TestImage, "Test"), it works. (Edit, t

I am not quite sure, but I think the offending methods are:
https://github.com/umbraco/Umbraco-CMS/blob/c2c6413eb7ba78cb64e7c470bbd406ed1156b70d/src/Umbraco.Web/UrlHelperRenderExtensions.cs#L240

and 

https://github.com/umbraco/Umbraco-CMS/blob/c2c6413eb7ba78cb64e7c470bbd406ed1156b70d/src/Umbraco.Web/UrlHelperRenderExtensions.cs#L265

Hoping someone can verify, thanks!

_No response_

nul800sebastiaan:
This is a breaking change for people using `GetCropUrl` to get local crops from the media picker v3. In order to disambiguate the extension method an make  it a bit clearer what it does, I propose the method name change, after which there is no need to have the first argument being the local crops, but it can be the whole MediaWithCrops object, since the method name already makes clear you're not trying to get the global crops.

Fixes #10402

### DIFF
--- a/src/Umbraco.Web/UrlHelperRenderExtensions.cs
+++ b/src/Umbraco.Web/UrlHelperRenderExtensions.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Web;
 using System.Web.Mvc;
 using Umbraco.Core;
+using Umbraco.Core.Models;
 using Umbraco.Core.Models.PublishedContent;
 using Umbraco.Core.PropertyEditors.ValueConverters;
 using Umbraco.Web.Composing;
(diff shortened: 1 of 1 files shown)


ronaldbarendse:
@nul800sebastiaan Changing the method parameter types would indeed be a breaking change, but this already uses a non-standard default value for `bool useCropDimensions = true` (all other overloads with this parameter have it set to `false`).

I've created PR https://github.com/umbraco/Umbraco-CMS/pull/10527 to align the overloads on `UrlHelper` with those on `IPublishedContent`, which would be a better fix IMHO 😇

nul800sebastiaan:
Closing in favor of the upcoming changes in 8.15 (https://github.com/umbraco/Umbraco-CMS/pull/10529).

PR #9461, opened 2020-11-27
Media Picker v3

nielslyngsoe:
This new media picker supports three new main features, Controlling accepted Media Types, Local Image Crops, and Copy/Paste.

This Media Picker will at some point replace the v2 Media Picker we have today. Most likely v2 will not exist in Umbraco v.9. Tomorrow it will just become an additional choice.

The features of this Media Picker is made clear by looking at the DataType configuration:

image

- Limit picking to certain Media Types.
- Better control of amounts of picked items.
- Set media start-node & ignore user start nodes(same feature as from v2)
- Local Image Croppings

**Test notes:**
- Set local focal point on images.
- Use local focus points in templates.
- Setup image croppings, change, delete, add new ones.
- Pick media and set local crops.
- Use image croppings in templates.
- Limiting accepted media types.
- Set validation limits, min & ma or just min and just max.
- Have pick multiple items turned on but limit the editor to pick maximum one item. (in this way the template still gets an IEnumerable)
- Put Media Item in trash bin and see it presented in the picker.
- Have more items picked than the validation allows
- Test copying media items from one property editor to another.
- Test that you cannot paste none-allowed media types.
- Copy media in media picker v2 to v2.
- Copy media from media picker v2 to v3.
- copy media from v3 to v2.
- copy a collection of media items(done in the property-actions-menu).
- test pasting a collection of media items that contain none-allowed types.(should not be possible)
- Test remove all media, from property-actions
- copy non-image media item from v3 to media picker v2 setup to only accept Image. (should not be possible)
- insert or paste media items at a certain spot in a multiple media picker.
- Read the documentation for ideas on more test cases

**Documentation PR:**
https://github.com/umbraco/UmbracoDocs/pull/3106

**Additional fixes:**
https://github.com/umbraco/Umbraco-CMS/issues/9948
https://github.com/umbraco/Umbraco-CMS/issues/10111

PR #10527, opened 2021-06-24
Fix GetCropUrl() extension methods on UrlHelper

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10402.

### Description
With the introduction of the new Media picker (v3 - with local crops), a new `GetCropUrl()` extension method on `UrlHelper` was added that causes a `The call is ambiguous between the following methods or properties` compiler error when used with named parameters.

This PR aligns the extension methods/parameters on `UrlHelper` with those on `IPublishedContent`. Some code might still need updating, but the impact would be a lot lower (as the new extension method has less parameters that can cause ambiguity) and the change will be a lot easier:

```c#
Url.GetCropUrl(Model.Image.LocalCrops, cropAlias: "Test") // Still ambiguous, but that's currently also the case for the IPublishedContent extension method
Url.GetCropUrl(Model.Image.LocalCrops, "Test") // Uses the crop alias and its dimensions, easy fix!
Url.GetCropUrl(Model.Image.LocalCrops, cropAlias: "Test", useCropDimensions: true) // Not ambiguous anymore, but the same as the previous call, a bit more explicit, but still an easy fix!
```

### DIFF
--- a/src/Umbraco.Core/Models/MediaWithCrops.cs
+++ b/src/Umbraco.Core/Models/MediaWithCrops.cs
@@ -1,15 +1,86 @@
+using System;
 using Umbraco.Core.Models.PublishedContent;
 using Umbraco.Core.PropertyEditors.ValueConverters;
 
 namespace Umbraco.Core.Models
 {
     /// <summary>
-    /// Model used in Razor Views for rendering
+    /// Represents a media item with local crops.
     /// </summary>
-    public class MediaWithCrops
+    /// <seealso cref="Umbraco.Core.Models.PublishedContent.PublishedContentWrapped" />
+    public class MediaWithCrops : PublishedContentWrapped
     {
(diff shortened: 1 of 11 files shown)


ronaldbarendse:
The changes in this PR show up a bit weird, but if you look at the separate commits, you'll clearly see I've removed the offending `GetCropUrl()` overload that caused the ambiguous method compiler error (introduced in https://github.com/umbraco/Umbraco-CMS/pull/9461) and added a new one to allow passing in local crops.

These commits are also part of PR https://github.com/umbraco/Umbraco-CMS/pull/10529, but that also contains new features.

ronaldbarendse:
I'll close this PR as #10529 that includes these changes is now merged!

Issue #10397, opened 2021-06-04
Media Picker 3 - It needs better backward compatibility and ease of use


DanDiplo:
### Umbraco version

v8

### Description

The new Media Picker 3 introduced in Umbraco 8.14 and also present in v9 adds some great new features and has a much nicer UI. However, it also creates many issues which don't seem to have been considered.

### The Problem

Firstly, it's not backward compatible with the previous Media Picker and there are [no plans for migrations to be included](https://github.com/umbraco/Umbraco-CMS/issues/10308). This creates a huge problem when upgrading existing sites as it requires editors to re-select *every single* piece of media and then republish the pages. For large sites, this is a massive undertaking.

Secondly, the property value converter no longer returns `IPublishedContent` but now returns a completely new type of [MediaWithCrops](https://github.com/umbraco/Umbraco-CMS/blob/809fd819821770472dec3eb2d7241f7d39f3c918/src/Umbraco.Core/Models/MediaWithCrops.cs). 

```
public class MediaWithCrops    
{        
        public IPublishedContent MediaItem { get; set; }
        public ImageCropperValue LocalCrops { get; set; }   
 }
```
In one fell swoop all code that has been written that assumes Media is `IPublishedContent` falls over and all current documentation, blog posts and articles on rendering media have been made obsolete. As far as I can tell, none of the official docs have been updated to show how you are supposed to deal with it - they still say [it returns IPublishedContent](https://our.umbraco.com/Documentation/Fundamentals/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker/) . I'm guessing that lots of packages etc. and existing code is going to require a lot of rewriting to deal with this.

If you install the default starter kit and swap the old media picker for Media Picker 3 you get exceptions galore. The site I tried to upgrade it to was strewn with exceptions where it expected `IPublishedContent` but got a different type or converters were throwing JSON parse errors all over the place - hardly a friendly experience.

This is further compounded by the fact that the old Media Picker is now hidden in Umbraco, meaning if you want to create a new media picker property you are forced down the road of using the new Media Picker 3. So now, even if you leave an upgraded site using the old picker, if you create new media properties these are now not compatible (so your rendering code now has to work with two different types, which is horrible).

[...]

PR #6950, opened 2019-10-29
Get missing crops from configuration

cleversolutions:
### Prerequisites

This fixes issue #5737 and Replaces pull request #5739 

### Description
When you add a crop alias to Image Cropper after media already exists in your Media library it doesn't pick up any of the crops. This PR updates ImageCropperTemplateExtensions to check for this condition and configure crops from the DataType.Configuration. 

### Steps to reproduce
As per issue #5737 
- V8 with SK installed
- Update Image Media Type -> Update umbracoFile Image Cropper DataType settings to include a crop type (ie test)
- Update Products.cshtml / template to use that crop ie ```@product.Pho0tos.GetCropUrl("test")``` 
- View page on frontend & confirm images render (they don't before PR)
- Goto Media Section & Choose a Product Image
- Set manual crops for an image
- View page on frontend & confirm all images render (only the updated image does before PR)

PR #10529, opened 2021-06-24
Improvements to media pickers/crop handling and URL generation

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10308, https://github.com/umbraco/Umbraco-CMS/issues/10402 and https://github.com/umbraco/Umbraco-CMS/issues/10397.

### Description
Building upon PR https://github.com/umbraco/Umbraco-CMS/pull/10517 that introduces backwards compatibility for the stored data format between `Media Picker (legacy)` (v2) and the new `Media Picker` (v3), this PR adds compatibility to the returned models on the front-end (and includes some additional improvements).

#### Model compatibility with `IPublishedContent`
This compatibility is added by making `MediaWithCrops` inherit from `PublishedContentWrapped`, making it an `IPublishedContent` instance again, so the following Razor/view code won't break by switching to the new picker:

```cshtml
@if (Model.Image is IPublishedContent image)
{
   <img src="@image.Url()" alt="@image.Name" />
}
```

One thing to note though: the concrete implementation of the `IPublishedContent` will change to `MediaWithCrops<T>`, so you can't cast directly to the Models Builder generated model (e.g. `var image = (Image)Model.Image` will still break). If you want the concrete media item model, you can either cast the 'wrapped' content to `Image` or better yet, cast to `MediaWithCrops<Image>` and access it using the new `Content` property:

```cshtml
@if (Model.Image?.Content is Image image)
{
    <img src="@image.Url()" width="@image.UmbracoWidth" height="@image.UmbracoHeight" alt="@image.Name" />
}

@if (Model.Image is MediaWithCrops<Image> image)
{
    // Now you still have access to the local crops (image.LocalCrops)
    <img src="@image.Url()" width="@image.Content.UmbracoWidth" height="@image.Content.UmbracoHeight" alt="@image.Name" />
}
```

#### New `GetCropUrl()` extension methods
Using local crops (defined on the Media Picker data type) together with media crops (defined on the Image Cropper data type) wasn't yet possible, so you'd have to ensure you're using the correct crops and aliases. This PR now merges local crops with the media ones, so the calling code can be the same for local and media crops:

```cshtml
@if (Model.Image is MediaWithCrops<Image> image)
{
    // Use the Hero crop either from the local or media crops
    <img src="@image.GetCropUrl("Hero")" width="@image.Content.UmbracoWidth" height="@image.Content.UmbracoHeight" alt="@image.Name" />

    // Use the Hero crop only from the local crops
    <img src="@image.GetCropUrl(image.LocalCrops, "Hero")" width="@image.Content.UmbracoWidth" height="@image.Content.UmbracoHeight" alt="@image.Name" />

    // Use the Hero crop only from the media crops
    <img src="@image.Content.GetCropUrl("Hero")" width="@image.Content.UmbracoWidth" height="@image.Content.UmbracoHeight" alt="@image.Name" />
}
```

In short: if the local crops doesn't contain the crop alias, it will look in the media crops. Also, if you have the focal point disabled on the Media Picker, it will fall-back to the global focal point (on the media item/Image Cropper).

#### Only return configured crops in the model
Because the stored data contains all crop information (as JSON), removing a crop from the configuration didn't remove existing crops from the returned model. A previous PR (https://github.com/umbraco/Umbraco-CMS/pull/6950) already added missing crops (not in the stored data, but later added to the configuration), but because removing a local crop should now fall-back to the media crops, this caused a problem. This PR ensures crops removed from the configuration (either local crops on the Media Picker or media crops on the Image Cropper) are not returned in the model anymore, making sure you can't use outdated crops on the front-end.

I'll make sure to add some comments to the code, as it might be quite a lot to go though 😇

### DIFF
--- a/src/Umbraco.Core/PropertyEditors/ImageCropperConfiguration.cs
+++ b/src/Umbraco.Core/PropertyEditors/ImageCropperConfiguration.cs
@@ -42,7 +42,7 @@ namespace Umbraco.Core.PropertyEditors
             {
                 foreach (var configuredCrop in configuredCrops)
                 {
-                    var crop = imageCropperValue.Crops.FirstOrDefault(x => x.Alias == configuredCrop.Alias);
+                    var crop = imageCropperValue.Crops?.FirstOrDefault(x => x.Alias == configuredCrop.Alias);
 
                     crops.Add(new ImageCropperCrop
                     {
--- a/src/Umbraco.Core/PropertyEditors/ImageCropperConfiguration.cs
+++ b/src/Umbraco.Core/PropertyEditors/ImageCropperConfiguration.cs
@@ -1,4 +1,8 @@
 ﻿using Newtonsoft.Json;
+using System.Collections.Generic;
+using System.Linq;
+using Umbraco.Core.PropertyEditors.ValueConverters;
+using static Umbraco.Core.PropertyEditors.ValueConverters.ImageCropperValue;
 
 namespace Umbraco.Core.PropertyEditors
 {
@@ -20,6 +24,37 @@ namespace Umbraco.Core.PropertyEditors
 
             [JsonProperty("height")]
             public int Height { get; set; }
+        }
+    }
+
+    internal static class ImageCropperConfigurationExtensions
+    {
+        /// <summary>
+        /// Applies the configuration to ensure only valid crops are kept and have the correct width/height.
+        /// </summary>
+        /// <param name="configuration">The configuration.</param>
+        public static void ApplyConfiguration(this ImageCropperValue imageCropperValue, ImageCropperConfiguration configuration)
+        {
+            var crops = new List<ImageCropperCrop>();
+
+            var configuredCrops = configuration?.Crops;
+            if (configuredCrops != null)
+            {
+                foreach (var configuredCrop in configuredCrops)
+                {
+                    var crop = imageCropperValue.Crops.FirstOrDefault(x => x.Alias == configuredCrop.Alias);
+
+                    crops.Add(new ImageCropperCrop
+                    {
+                        Alias = configuredCrop.Alias,
+                        Width = configuredCrop.Width,
+                        Height = configuredCrop.Height,
+                        Coordinates = crop?.Coordinates
+                    });
+                }
+            }
+
+            imageCropperValue.Crops = crops;
         }
     }
 }
--- a/src/Umbraco.Web/ImageCropperTemplateExtensions.cs
+++ b/src/Umbraco.Web/ImageCropperTemplateExtensions.cs
@@ -30,6 +30,9 @@ namespace Umbraco.Web
         /// </returns>
         public static string GetCropUrl(this IPublishedContent mediaItem, string cropAlias) => ImageCropperTemplateCoreExtensions.GetCropUrl(mediaItem, cropAlias, Current.ImageUrlGenerator);
 
+        public static string GetCropUrl(this MediaWithCrops mediaWithCrops, string cropAlias) => ImageCropperTemplateCoreExtensions.GetCropUrl(mediaWithCrops, cropAlias, Current.ImageUrlGenerator);
+
+        [Obsolete("This method does not get the crops or cache buster value from the media item.")]
         public static string GetCropUrl(this IPublishedContent mediaItem, string cropAlias, ImageCropperValue imageCropperValue) => ImageCropperTemplateCoreExtensions.GetCropUrl(mediaItem, cropAlias, Current.ImageUrlGenerator, imageCropperValue);
 
         /// <summary>
@@ -48,6 +51,8 @@ namespace Umbraco.Web
         /// The ImageProcessor.Web URL.
         /// </returns>
         public static string GetCropUrl(this IPublishedContent mediaItem, string propertyAlias, string cropAlias) => ImageCropperTemplateCoreExtensions.GetCropUrl(mediaItem, propertyAlias, cropAlias, Current.ImageUrlGenerator);
+
+        public static string GetCropUrl(this MediaWithCrops mediaWithCrops, string propertyAlias, string cropAlias) => ImageCropperTemplateCoreExtensions.GetCropUrl(mediaWithCrops, propertyAlias, cropAlias, Current.ImageUrlGenerator);
 
         /// <summary>
         /// Gets the ImageProcessor URL from the IPublishedContent item.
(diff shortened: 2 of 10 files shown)


src/Umbraco.Core/PropertyEditors/ImageCropperConfiguration.cs
```diff
@@ -22,4 +26,35 @@ public class Crop
...
         }
     }
+
+    internal static class ImageCropperConfigurationExtensions
+    {
+        /// <summary>
+        /// Applies the configuration to ensure only valid crops are kept and have the correct width/height.
+        /// </summary>
+        /// <param name="configuration">The configuration.</param>
+        public static void ApplyConfiguration(this ImageCropperValue imageCropperValue, ImageCropperConfiguration configuration)
```
ronaldbarendse, line 36: I've moved this internal method from the `ImageCropperValue` to an extension method, so it aligns with the other `ApplyConfiguration` method and lives nicely next the the configuration it applies.

src/Umbraco.Core/PropertyEditors/ImageCropperConfiguration.cs
```diff
@@ -22,4 +26,35 @@ public class Crop
...
+                foreach (var configuredCrop in configuredCrops)
+                {
+                    var crop = imageCropperValue.Crops?.FirstOrDefault(x => x.Alias == configuredCrop.Alias);
+
+                    crops.Add(new ImageCropperCrop
+                    {
+                        Alias = configuredCrop.Alias,
+                        Width = configuredCrop.Width,
+                        Height = configuredCrop.Height,
+                        Coordinates = crop?.Coordinates
```
ronaldbarendse, line 52: The coordinates are actually the only data we need to store for a crop (together with the alias). We could reduce the size of all stored crop data by removing the width and height!

src/Umbraco.Web/ImageCropperTemplateExtensions.cs
```diff
@@ -30,6 +30,9 @@ public static class ImageCropperTemplateExtensions
         /// </returns>
         public static string GetCropUrl(this IPublishedContent mediaItem, string cropAlias) => ImageCropperTemplateCoreExtensions.GetCropUrl(mediaItem, cropAlias, Current.ImageUrlGenerator);
 
+        public static string GetCropUrl(this MediaWithCrops mediaWithCrops, string cropAlias) => ImageCropperTemplateCoreExtensions.GetCropUrl(mediaWithCrops, cropAlias, Current.ImageUrlGenerator);
+
+        [Obsolete("This method does not get the crops or cache buster value from the media item.")]
```
ronaldbarendse, line 35: We might want to fix this method instead of making it obsolete, will get back to this later!

src/Umbraco.Web/PropertyEditors/MediaPicker3Configuration.cs
```diff
@@ -57,4 +61,40 @@ public class CropConfiguration
...
+                        Width = configuredCrop.Width,
+                        Height = configuredCrop.Height,
+                        Coordinates = crop?.Coordinates
+                    });
+                }
+            }
+
+            imageCropperValue.Crops = crops;
+
+            if (configuration?.EnableLocalFocalPoint == false)
```
ronaldbarendse, line 94: We need to check for this configuration setting, as the picker always stores the default focal point (0.5, 0.5). Removing this together with the crop width/height would also reduce the size of data stored!

src/Umbraco.Web/ImageCropperTemplateExtensions.cs
ronaldbarendse: This method could be useful to only get a crop from the local crops: `image.Content.GetCropUrl("Hero", image.LocalCrops)`, but lacks quite a few parameters and doesn't add the automatic cache buster value based on the media update date/time. It also doesn't align with the `UrlHelper` extension method: `Url.GetCropUrl(image.LocalCrops, "Hero")` (which has the same flaw BTW).

We should probably look into having all these extension methods aligned better and decide on how to best handle only using local crops: not something we need to block this PR for.

src/Umbraco.Web/PropertyEditors/ValueConverters/MediaPickerWithCropsValueConverter.cs
```diff
@@ -51,22 +51,27 @@ public override object ConvertIntermediateToObject(IPublishedElement owner, IPub
...
-                        }
-                    });
+                        Crops = dto.Crops,
+                        FocalPoint = dto.FocalPoint,
+                        Src = mediaItem.Url()
+                    };
+
+                    localCrops.ApplyConfiguration(configuration);
+
+                    // TODO: This should be optimized/cached, as calling Activator.CreateInstance is slow
```
ronaldbarendse, line 70: I've added this to-do, because it's also something that's done in the `BlockListPropertyValueConverter` and might have considerable performance gains by using similar techniques as in `PublishedModelFactory` (caching emitted constructor functions using `ReflectionUtilities`). Maybe @Shazwazza can give some pointers on how to best handle/implement this?

src/Umbraco.Web/ImageCropperTemplateExtensions.cs
ronaldbarendse: Fixed in 2980ff9c54fe7c4bdf4f0c19a5672f148c3cd5d2 by forwarding this to the `GetCropUrl` extension method on `IPublishedContent` and adding flag to only use the local crops that are passed in 👍🏻

ronaldbarendse:
I've created an overview of the changes in this PR, as it contains some breaking and functional changes. These should all be very minimal (or not noticeable at all if you're upgrading from pre 8.14):

### Breaking changes:
- `MediaWithCrops` model: no parameterless constructor and properties don't have a setter anymore
- Removed ambiguous method: `IHtmlString GetCropUrl(this UrlHelper urlHelper, ImageCropperValue imageCropperValue, string cropAlias, bool htmlEncode = true)`  (was also available as separate PR [#10527](https://github.com/umbraco/Umbraco-CMS/pull/10527))

### Functional changes:
- Local crops are merged with media crops when using the `GetCropUrl()` extension methods on `MediaWithCrops`
  - The `GetLocalCropUrl()` extension methods are now obsoleted because of this
- The concrete implementation of `MediaWithCrops` model is now `MediaWithCrops<T>` and implements `IPublishedContent`
  - The `MediaItem` property is obsoleted, as it's now recommended to use the `Content` property (that is of a specific type when using the generic model)
- Only configured crops are returned in the `ImageCropperValue` and `MediaWithCrops` models

We should make sure to also update the documentation for 8.15: https://our.umbraco.com/Documentation/Fundamentals/Backoffice/property-editors/built-in-property-editors/Media-Picker-3/

PR #10531, opened 2021-06-25
V9: Autogeneration of AppSettings JSON schema and Merge the official + Umbraco

bergmania:
### Description
This PRs is the result of a Cycle hack.
We have added a new project `JsonSchema` that uses [NJsonSchema](https://github.com/RicoSuter/NJsonSchema) to generate the schema from Models.

The project creates some dummy models to simulate the `AppSettings:Umbraco:CMS`, `AppSettings:Umbraco:Forms` and `AppSettings:Umbraco:Deploy` part of the configuration paths.

It copies the XML docs from forms from the nuget cache to use inside the `JsonSchema` project and use the docs as information in the generated JSON Schema.

This PR also updates the `AppSettings*.json` files to point to this newly generated schema. To avoid the issue of not showing all non-Umbraco generate IntelliSense for `AppSettings*.json`, we merge the official [AppSettings JsonSchema](https://json.schemastore.org/appsettings.json) into our generated schema.

Futhermore, this PR also contains some refactoring to the way we register configurations using an `UmbracoOptionsAttribute`.

### Next steps
Some ideas for the future
- Add dependency to Deploy, as soon as a nightly is available and generate schema for these.

### Test
- Ensure the generated artifact [`Umbraco.Cms.StaticAssets.9.0.0-preview*.nupkg`](https://dev.azure.com/umbraco/_apis/resources/Containers/18560616/nupkg?itemPath=nupkg%2FUmbraco.Cms.StaticAssets.9.0.0-preview20210625.66532.nupkg) contains the new schema file `content/umbraco/config/appsettings-schema.json`.
- Ensure the generated artifact  [`Umbraco.Templates.9.0.0-preview*.nupkg`](https://dev.azure.com/umbraco/_apis/resources/Containers/18560616/nupkg?itemPath=nupkg%2FUmbraco.Templates.9.0.0-preview20210625.66532.nupkg) has this line 
``` "$schema" : "./umbraco/config/appsettings-schema.json",``` 
in the following files:
  - `appsettings.Development.json`
  - `appsettings.json`


warrenbuckley:
## Notes
You will need to add a Nuget source to your build.out folder so that doing a dotnet new template install can install the local build artifacts

`dotnet nuget add source --name LocalNetCore C:\Code\Umbraco-CMS_netcore\build.out`

Secondly when opening the folder/project after a `dotnet new umbraco` VSCode and probably other IDEs will complain that the schema JSON file is missing.

This is because dotnet build or dotnet run needs to be done which will copy in the static assets and thus the /umbraco/config folder containing the JSON schema

warrenbuckley:
## Feedback

### Remove Calculated Properties
Not sure this makes sense that you could add/set this property `Global.IsSmtpServerConfigured` as this is a calculated property that we do at runtime to help forgotten password email invite flow etc...

image

bergmania:
> A few config values are camelCased where most start with Capitals

We can't control this, as the RuntimeMinification class is not owned by the CMS

> Some of the XML comments for these are not very helpful as it does not give me context as to what it is doing.
Agree, could be cool with better descriptions.. But is not necessarily this task.

> Hide certain keys/config
We can only hide entire groups, e.g. CoreDebug.

PR #10534, opened 2021-06-25
Streamlines response handling in controllers

Shazwazza:
These changes are due to me porting Articulate over to v9 and I found it quite difficult to know exactly what to return as responses in controllers since it's not clear what Umbraco expects. Also until these changes you cannot use the normal aspnetcore controller patterns for returning results which is normally based on base class methods. This fixes that.

This makes validation response handling more inline with aspnetcore patterns by using the base class `ValidationProblem` method instead of having to know how to handle messages, notifications, model state, etc...

Now, based on the the controller you inherit from you should just use the normal methods:

* ValidationProblem for any validation problem. This overrides some default behavior and also adds custom protected methods.
* Ok for OK actions and not have to worry about notifications, etc.. I've removed UmbracoNotificationSuccessResponse since people shouldn't need to know about this

All controllers are changed to use these base class methods instead of newing up the different validation responses. 

This will make it much easier for developers with custom controllers to just return ValidationProblem and Ok for responses and the base classes will handle notifications, etc... 

I've added notes about using aspnetcore's ProblemDetails class which would be preferred since that would follow aspnetcore's conventions but doing that would require changes to angular which we're not prepared to do at the moment.


src/Umbraco.Web.Common/ActionsResults/ValidationErrorResult.cs
```diff
@@ -1,10 +1,19 @@
...
 using Umbraco.Extensions;
 
 namespace Umbraco.Cms.Web.Common.ActionsResults
 {
+    // TODO: This should probably follow the same conventions as in aspnet core and use ProblemDetails
+    // and ProblemDetails factory. See https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/ControllerBase.cs#L1977
+    // ProblemDetails is explicitly checked for in the application model.
+    // In our base class UmbracoAuthorizedApiController the logic is there to create a ProblemDetails.
+    // However, to do this will require changing how angular deals with errors since the response will
+    // probably be different. Would just be better to follow the aspnet patterns.
```
bergmania, line 15: Agreed, but I think it is better to wait for the new Backoffice, before we make such change.

src/Umbraco.Web.Common/ActionsResults/UmbracoProblemResult.cs
```diff
@@ -1,8 +1,9 @@
-﻿using System.Net;
+using System.Net;
 using Microsoft.AspNetCore.Mvc;
 
 namespace Umbraco.Cms.Web.Common.ActionsResults
 {
+    // TODO: What is the purpose of this? Doesn't seem to add any benefit
```
bergmania, line 6: The only purpose is to ensure the result is the expected format for the frontend (the `new {Message = message}` part).

bergmania (approved):

PR #10541, opened 2021-06-27
Accessibility Amends for the Content Info Tab

RachBreeze:
As a result of investigating #8704 I have added an alias for the content info tab to ensure that the template drop down is labelled correctly.

I have also added a space on the open buttons to ensure that the copy reads better than shown below.
image

Finally I have removed the title attributes on the links in `umb-node-preview` as they held duplicate (but unlocalised) content as the link text and could result in some screen reader users hearing duplicate messages.


nul800sebastiaan:
Thanks very much @RachBreeze 👍

nul800sebastiaan:
Thanks @bjarnef - I overlooked that one! 😅  Fixed: f7dee7ec74198918693fb6ba50caed9e1af53a80

Issue #10548, opened 2021-06-29
v9-beta0004: Can't use AddUnique and consume scoped dependencies at the same time


mattbrailsford:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

v9.0.0-beta004

### Bug summary

Registering unique services are always registered as singleton and so you can't access transient or scoped services within them.

### Specifics

I'm creating a service in v9 which needs to be able to query for content and so I a have a dependency on the `IPublishedContentQuery` interface. At the same time, I need my class to be replaceable so that developers can swap the implementation and so I register is using `AddUnique`

Unfortunately this setup is not possible as `IPublishedContentQuery` is registered as a scoped dependency and `AddUnique` ALWAYS registers the given service as a singleton and so .NET Core blows up on boot with the error.

    Cannot consume scoped service 'Umbraco.Cms.Core.IPublishedContentQuery' from singleton 'Vendr.Umbraco.Extractors.IUmbracoProductNameExtractor'.

I'm personally not sure why `IPublishedContentQuery` needs to be scoped as this shouldn't be specific to the given request. Unfortunately I can't find any other way of fetching published content that would work outside of a scoped dependency.

### Steps to reproduce

Create a service that consumes the `IPublishedContentQuery` interface and register is with the DI container using `AddUnique`. On application boot the error will be displayed.

### Expected result / actual result

`IPublishedContentQuery` shouldn't be a scoped dependency and so no error should be thrown, or, there should be some other way of querying for content that doesn't need to be scoped.

bergmania:
@Shazwazza do you see any reason `PublishedContentQuery` cant inject `IPublishedSnapshotAccessor` instead of `IPublishedSnapshot` and thereby open up for `IPublishedContentQuery` and `UmbracoHelper` is registered as singletons?

Shazwazza:
Have replied with review here: https://github.com/umbraco/Umbraco-CMS/pull/10551#pullrequestreview-700291297

Issue #10540, opened 2021-06-27
when resolving UmbracoHelper in custom middleware DI throws NullReferenceException


vmachacek:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0-beta0004

### Bug summary

When I'm trying to resolve `UmbracoHelper` in custom middleware I get NullReferenceException in umbraco infrastructure code. 

registration:
```
app.UseUmbraco()
      .WithMiddleware(u =>
    {
        u.WithBackOffice();
        u.WithWebsite();
        u.AppBuilder.MapWhen(ctx => ctx.Request.Path.StartsWithSegments("/robots.txt"),
                        b => b.UseMiddleware<RobotsTxtMiddleware>());
    })
    .WithEndpoints(u =>
    {
        u.UseInstallerEndpoints();
        u.UseBackOfficeEndpoints();
        u.UseWebsiteEndpoints();
});
```

middleware:
```
public class RobotsTxtMiddleware
    {
        const string Default =
            @"User-Agent: *\nAllow: /";

        private readonly RequestDelegate next;

        public RobotsTxtMiddleware(RequestDelegate next)
        {
            this.next = next;
        }

        public async Task InvokeAsync(
            HttpContext context,
            UmbracoHelper uHelper)
        {
            ....
        }
    }
```

### Steps to reproduce

- dotnet new 
- add middleware
- register middleware
- go to /robots.txt
- see exception

### Expected result / actual result

expected: insance of UmbracoHelper to be resolved. 

actual: NullReferenceException is thrown in infrastructure code

bergmania:
Hi @vmachacek, I just looked into this, and it makes good sense. 

The `UmbracoHelper` requires `UmbracoContext` to be initialized, but it is not because "/robots.txt" is considered a client-side request.

What do you need from `UmbracoHelper`? 

There are a couple of possible workarounds. It would be very bad for performance if we had to initialize `UmbracoContext` for all client requests.

vmachacek:
My main objective is to generate sitemap.xml based on latest published content and also robots.txt based at the root of umbraco content. The root node have property which contains contents for robots.txt in a text field.

The respective middlewares are registered with `builder.AppBuilder.MapWhen` with condition on the url, so they wont kick in during processing of irrelevant requests.

I understand creating `UmbracoHelper` after calling `EnsureUmbracoContext` is expensive operation, so I will cache it in the middleware. 

This is what I ended up with 

[...]

PR #10551, opened 2021-06-29
V9: IPublishedContentQuery as singleton

bergmania:
Potential fix for 
https://github.com/umbraco/Umbraco-CMS/issues/10548 
and
https://github.com/umbraco/Umbraco-CMS/issues/10540

Not sure if this I going to introduce new issues...

### DIFF
--- a/src/Umbraco.Infrastructure/PublishedContentQuery.cs
+++ b/src/Umbraco.Infrastructure/PublishedContentQuery.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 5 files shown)


src/Umbraco.Infrastructure/PublishedContentQuery.cs
```diff
@@ -93,14 +93,14 @@ private static bool ConvertIdObjectToUdi(object id, out Udi guidId)
 
         #region Content
 
-        public IPublishedContent Content(int id) => ItemById(id, _publishedSnapshot.Content);
+        public IPublishedContent Content(int id) => ItemById(id, _publishedSnapshotAccessor.PublishedSnapshot.Content);
```
Shazwazza, line 96: There might not be a published snapshot, these need null checks and then to throw a meaningful exception explaining why. It won't work without a snapshot which requires a current umbraco context

Shazwazza (changes requested):
So sure, it's possible to make things singletons but the truth is that they still require scoped services. So null checks need to be put in place for all accessors which are missing. UmbracoHelper, PublishedContentQuery, etc... won't 'just' work as a singleton, there needs to be a current umbraco context else it will explode. IMO it's safer to make scoped services as scoped services instead of trying to shoe-horn them into singletons where you don't really ever know if they are going to work?

src/Umbraco.Infrastructure/PublishedContentQuery.cs
bergmania: Alternatively we could add `IUmbracoHelperAccessor` and `IPublishedContentQueryAccessor`? It will be a fairly used usecase to need one of these from a singleton. 

I know this will just be syntactical sugar for injecting IHttpContextAccessor and use that as ServiceProvider.

Shazwazza: I think that would make things a lot clearer as to what to expect. I just don't see how people are going to use these in singletons though. The only way that will be possible is that within the singleton they use the UmbracoContextFactory and while using the resulting context then use one of these accessors.

PR #10555, opened 2021-06-29
Update contributing.md Visual Studio min version

dchallener:
Having had an older version of visual studio running I've been unable to build the source without making changes to the solution. However, upgrading to at least VS 16.8 has meant those changes haven't been required. This is to do with how the language target "latest" was working.  
Propose upping the min version listed here to avoid build issues being raised.


nul800sebastiaan:
Thanks for helping out here @dchallener 👍

PR #10556, opened 2021-06-29
Disable cached partials if in preview mode

### LINKED ISSUE #10384: Unpublished items showing on live site when using preview and CachedPartial

8.14.0

### Bug summary

When using the CachedPartial helper it shows items in "draft" state on the live site if any editor previews the page, until an editor publishes a page and clears the cache.

### Specifics

- Once the page is previewed, it seems to add all the items to cache.
- Publishing the page invalidates the cache again and all good.
- Changing the CachedPartial for a @Html.Partial works as expected.
- Interestingly, using cacheByPage: true doesn't effect this - the incorrect behaviour still occurs.
- Testing in a new browser to make sure the preview cookie isn't set still shows the issue.

This also happens in 8.6.3 as this was the version I noticed this behaviour.

### Steps to reproduce

Install the latest Umbraco (8.14.0 at time of writing) with starter kit
Change the people.cshtml to use a cachedpartial for displaying children
Add a new person (unpublished in draft state)
Preview people
Open the live /people page in a new browser

Expected to see the live site only show published items whilst the preview site shows both draft and published items.
Actually see both live and preview links showing all items.

dchallener:
We don't want unpublished items in cached views - possible leak to live site. Show previews using normal partial view instead.

This PR fixes #10384 

### Description

This adds a check to see if running in preview mode and using the normal non-cached partials if it is. 
The cachehelper already has a check for whether debugging is enabled and have extended this approach to allow for in-preview mode.

Due to the debugging check, you do need to remember to set debug="false" to test.

### DIFF
--- a/src/Umbraco.Web/CacheHelperExtensions.cs
+++ b/src/Umbraco.Web/CacheHelperExtensions.cs
@@ -37,7 +37,8 @@ namespace Umbraco.Web
             ViewDataDictionary viewData = null)
         {
             //disable cached partials in debug mode: http://issues.umbraco.org/issue/U4-5940
-            if (htmlHelper.ViewContext.HttpContext.IsDebuggingEnabled)
+            //disable cached partials in preview mode: https://github.com/umbraco/Umbraco-CMS/issues/10384
+            if (htmlHelper.ViewContext.HttpContext.IsDebuggingEnabled || Umbraco.Web.Composing.Current.UmbracoContext.InPreviewMode)
             {
                 // just return a normal partial view instead
                 return htmlHelper.Partial(partialViewName, model, viewData);


nathanwoulfe (approved):
Hey @dchallener I've had a look at this, and have been able to reproduce the issue, and see it resolved with your changes. Nice work!

nathanwoulfe:
And given this is your first Umbraco PR (🥇), we'd love to offer a shiny contributors badge for your our.umbraco.com profile (assuming you have one?). Let us know your profile URL, and @nul800sebastiaan will make the magic happen.

nul800sebastiaan:
I think @dchallener might have contributed on other Umbraco repositories before since the badge is already there! https://our.umbraco.com/members/dchallener But thanks for your first code update to CMS David! 👍

dchallener:
Thanks Nathan, Seb - that's me. I have contributed but first time to codebase. Glad I could help!

marcemarc:
Nice @dchallener !!

Also

Fixes: https://issues.umbraco.org/issue/U4-9022 :-P

dchallener:
Thanks @marcemarc ! 
Oh yeah, well that's the same issue, from 5 years ago :D

PR #10560, opened 2021-06-30
V9: Route plugin controllers by area

nikolajlauridsen:
When you created a surface controller in V8 with the `[PluginController("SomeArea")]` attribute, the surface controller would be routed like `/umbraco/{areaname}/{controllername}/{action}/{id}`, however in V9 we don't account for area names when creating the pattern, so even if you have the attribute your plugin surface controller will be routed like a normal surface controller I.E: `/umbraco/surface/{controllername}/{action}/{id}`, which might cause issues since a plugin controller might clash with a locally declared one.

I noticed that we handle this with front end API's, so I've applied the same kind of logic to the normal surface controller routing, so now a surface controller with the plugin controller attribute will be routed the same way as in V8.

I've also added an integration test that shows this, if you revert back to `MapUmbracoRoute` instead of the new `MapUmbracoSurfaceRoute` you can see that the test fails.

It's worth noting that this PR won't affect front end API controllers, only surface controllers

### DIFF
--- a/src/Umbraco.Tests.Integration/Umbraco.Web.Website/Routing/FrontEndRouteTests.cs
+++ b/src/Umbraco.Tests.Integration/Umbraco.Web.Website/Routing/FrontEndRouteTests.cs
@@ -1,5 +1,7 @@
+using System;
 using System.Net;
 using System.Net.Http;
+using System.Reflection;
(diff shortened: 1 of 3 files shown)


bergmania:
Seems correct 💪

PR #10562, opened 2021-06-30
#10520 ensure property culture in Nested Content

### LINKED ISSUE #10520: Mandatory Nested Content forces validation on all properties in said NC

8.14.0

### Bug summary

When you have a Nested content propertype in the backoffice that forces validation to pick a element type, it will then force the same validation across all the properties inside the element type, even though it is not set as mandatory on the element type.

image

image

image

### Steps to reproduce

1. Make an Element type with the following: 
Two textfields.
Make one of them Mandatory, leave the other as non-mandatory

2. Make a Documentype with a nested content property editor, where you select the element type from earlier.
3. Enable mandatory state on the Documentype for the Nested content, and allow to be created in root.
4. Create a content node.
5. You will now see that the non-mandatory field is set to be Mandatory

I expect that the Mandatory setting is only applied for the select property and not everything inside the Nested content.

The actual result is that we get everything set as mandatory that is in the Nested content

---
_This item has been added to our backlog [AB#12571](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/12571)_

nielslyngsoe:
This fixes #10520

The description of the above issue doesn't cover exactly whats going on. But from my investigation the problem appears when the Nested Content property is set to vary by culture. If so any server-side validations will not be binded correctly with the form fields, therefor its not possible to revalidate the fields by changing the value of the field. (since the field isn't connected to the server-side validation-message).

In order to fix this we needed the property to have the right culture, this was already solved in Block List Editor, so using the same approach. Parsing the NC property culture to each property of the scaffold model of each Element Type in Nested Content makes this work.

Test notes:
All tests are with a document type with a Nested Content property which has fields for each of these configs: regex, mandatory, non-mandatory.

- Test culture-invarint DocType.
- Test culture-variant DocType, with invariant NC field.
- Test culture-variant DocType with culture-varying NC field.

---
_This item has been added to our backlog [AB#12612](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/12612)_

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
@@ -190,7 +190,7 @@
         };
 
         vm.openNodeTypePicker = function ($event) {
-            
+
             if (vm.nodes.length >= vm.maxItems) {
                 return;
             }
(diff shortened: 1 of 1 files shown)


madsrasmussen (approved):
Tested and it works as described. Code changes look good to me. 🥳

PR #10563, opened 2021-06-30
#10520 Correct mandatory markers in nested content

madsrasmussen:
Scenario: 
When two nested content property editors are nested inside each other, and the inner nested content is set to mandatory. All properties inside the inner nested content will be marked with an asterisk (*). Validation doesn't kick in though.

_Before: Text 4 is not set to mandatory_
Screenshot 2021-06-30 at 15 25 37

_After:_
The PR fixes so only mandatory properties are marked with an asterisk (*).
Screenshot 2021-06-30 at 15 24 19

#10520

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-nested-content.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-nested-content.less
@@ -15,17 +15,6 @@
     pointer-events: none;
 }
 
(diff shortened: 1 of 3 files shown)


nielslyngsoe (approved):
Cool, less code and it works.

nul800sebastiaan:
Cherry picked for 8.15 in fc108511d28153495353abd8527032a78170df8f

PR #10565, opened 2021-06-30
Copy block configurations

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/discussions/10561

### Description
Some initial work to support copying block configuration.

U4Se5Fy10R

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/prevalue/blocklist.blockconfiguration.controller.js
@@ -50,9 +50,6 @@
             deleteAllBlocksAction
         ];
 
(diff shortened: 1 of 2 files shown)


bjarnef:
@nielslyngsoe @madsrasmussen do you have some thoughts on this? Should the data be stored as `RAW`, `BlOCK`, `ELEMENT_TYPE` or do we need a new type for this?

bjarnef:
Do we need a different/new storage type? When configuration block list editor it takes a fair amount of time to configurate instances of block list editors, so it would be a time saver to copy all or individual block configurations.

bjarnef:
@nul800sebastiaan do we still want this in v8 or only v9? I would like some feedback from HQ how we could copy configuration of datatypes, e.g. alm block items or just a single block configuration. It can also be useful to copy e.g. predefined colors in color picker datatype.

PR #10571, opened 2021-07-01
Block list editor: Align "Add" button behavior with the way Nested Content works

BatJan:
### Prerequisites

### Description
While I was building a new site making use of the Block List Editor it took me by surprise that the "Add" button remained active even though I had added a maximum of allowed entries. It was only after I tried adding one more block / element that I got the message that there is a limitation of 1. Therefore I'm doing this PR to align the UX with what we have with Nested Content where the button is disabled when the maximum of allowed entries have been met saving the editors an unnecessary click only to get frustrated that they can't add more blocks even though the UI up front indicates it :-)

If this PR is accepted I'm wondering if before merging I should remove the validation from `umbBlockListPropertyEditor.component.js`, which would be these lines since they will not longer be used I think?

```
// validate limits:
if (vm.propertyForm && vm.validationLimit) {

    var isMinRequirementGood = vm.validationLimit.min === null || vm.layout.length >= vm.validationLimit.min;
    vm.propertyForm.minCount.$setValidity("minCount", isMinRequirementGood);

    var isMaxRequirementGood = vm.validationLimit.max === null || vm.layout.length <= vm.validationLimit.max;
    vm.propertyForm.maxCount.$setValidity("maxCount", isMaxRequirementGood);
}
```

**Before**
ble-add-block-before

**After**
ble-add-block-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umb-block-list-property-editor.html
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umb-block-list-property-editor.html
@@ -31,7 +31,7 @@
                     id="{{vm.model.alias}}"
                     type="button"
                     class="btn-reset umb-block-list__create-button umb-outline"
-                    ng-disabled="vm.availableBlockTypes.length === 0"
+                    ng-disabled="vm.availableBlockTypes.length === 0 || vm.layout.length >= vm.validationLimit.max"
                     ng-click="vm.requestShowCreate(vm.layout.length, $event)">
                 <localize ng-if="vm.availableBlockTypes.length !== 1" key="blockEditor_addBlock">Add content</localize>
                 <localize ng-if="vm.availableBlockTypes.length === 1" key="blockEditor_addThis" tokens="[vm.availableBlockTypes[0].elementTypeModel.name]">Add content</localize>
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umb-block-list-property-editor.less
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/blocklist/umb-block-list-property-editor.less
@@ -289,7 +289,7 @@ ng-form.ng-invalid-val-server-match-settings > .umb-block-list__block > .umb-blo
         &[disabled]:hover {
             color: @gray-7;
             border-color: @gray-7;
-            cursor: default;
+            cursor: not-allowed;
         }
 
         &.umb-block-list__create-button {


rasmusjp:
Personally I think disabling the add button is bad UX and I would leave the validation in place.

E.g. if I want to replace an existing block with another one, but theres some content in the existing that I want to copy over, I then need to copy the content to somewhere else (which is not always possible with all property editors, and you could potentially loose formatting or markup if it's RTE content you copy to a document and then back), then remove the old block, add the new one and copy back the content from whereever I copied to.

nul800sebastiaan:
I think this is on purpose, you get to do what you wanted to do, but you will need to decide which one to delete afterwards. This would help editors to copy and paste some things from an existing item before removing it.

I'm not giving the full and thoughtful explanation of it here, but I'm sure @nielslyngsoe can help fill in the details!

BatJan:
@rasmusjp @nul800sebastiaan That is some good feedback - Had not considered this scenario. Old habbits and old that 😄 But the current implementation does not feel right either. But maybe it's a matter of changing the text from "Add" to "Change" or what do I know when the maximum limit of insertion has been reached? Or maybe adding a little "x of y blocks inserted" somewhere suitable? Not only for the block list editor but for every editor where it's possible to insert a range of items - Overall I think we would all like the experience to be consistent whether we're using one or the other property editor 👼🏻

BatJan:
@nul800sebastiaan @rasmusjp @nielslyngsoe Ok, so another issue this PR does not address is that the "blue" + button keeps appearing as well - But let's see if we can figure out a better way of handling it and I'll make sure to adjust this PR accordingly 👍🏻

nielslyngsoe:
Hi @BatJan, thanks for noticing the difference between these two. Seb and Rasmus are right that it is fully on purpose that we have enabled this experience. Which when you think about it might seem odd when thinking about the system but when thinking about how it is to be working with content this makes much more sense.
In general, the limits are there to ensure Editor don't publish too much content, but when working with content it can be relevant to have more than you aim for, either to keep old-version while writing the new version. Or as Seb suggests if you want to paste some blocks and select which to continue with. So for those working with content its good that you are able to have more while working.

I aim for this to be a general way of working with limits in the CMS. Therefore you will see a similar experience when having character limits in TextFields. As the content-editor must be able to paste long texts and then manually select which words to remove from it, it is also here possible to have more than what is allowed. Again publishing is prevented by the validation.

So in general the aim is to provide space for working with content and then ensure the right amount on Publishing. I hope this helps give a better understanding of the choice, and if you spot other places where this isn't the case then please let us know.

Btw. Nested Content's way of working is kept as we don't want to change existing behavior for sites using this, but hopefully, new websites use Blocks instead and will their for get an consistent experience through out Backoffice.

I suggest you create a new PR with the other thing you mentioned, just to keep the PRs cleaner to review. (Therefor I close this)

Thanks

PR #10575, opened 2021-07-01
V9: Add UmbracoAuthorizedController

nikolajlauridsen:
This PR adds back the `UmbracoAuthorizedController`, this is just a helper to make a normal controller that gets authorized for back office, it's described in the V8 docs [here](https://our.umbraco.com/documentation/Reference/Routing/Authorized/#mvc-controllers-for-the-backoffice)

### Testing

I've tested it out and everything seems to work, to test, first add a controller: 

```C#
using Umbraco.Cms.Web.Common.Controllers;

namespace Umbraco.Cms.Web.UI.NetCore
{
    [Area("Test")]
    public class AuthController : UmbracoAuthorizedController
    {
        public IActionResult Index()
        {
            return Content("Hello from authorized controller");
        }
    }
}
```

And register it, I used a composer, since I can re-use it for V9 docs 😉 

```C#
using Microsoft.AspNetCore.Builder;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Options;
using Umbraco.Cms.Core;
using Umbraco.Cms.Core.Configuration.Models;
using Umbraco.Cms.Core.Hosting;
using Umbraco.Cms.Web.Common.ApplicationBuilder;
using Umbraco.Extensions;

namespace Umbraco.Cms.Web.UI.NetCore
{
    public class AuthComposer : IUserComposer
    {
        {
            builder.Services.Configure<UmbracoPipelineOptions>(options =>
            {
                options.AddFilter(new UmbracoPipelineFilter(
                    "CustomBackOfficeController",
                    applicationBuilder => {},
                    applicationBuilder => {},
                    applicationBuilder =>
                    {
                        applicationBuilder.UseEndpoints(endpoints =>
                        {
                            var globalSettings = applicationBuilder.ApplicationServices
                                .GetRequiredService<IOptions<GlobalSettings>>().Value;
                            var hostingEnvironment = applicationBuilder.ApplicationServices
                                .GetRequiredService<IHostingEnvironment>();
                            var backofficeArea = Constants.Web.Mvc.BackOfficePathSegment;

                            var rootSegment = $"{globalSettings.GetUmbracoMvcArea(hostingEnvironment)}/{backofficeArea}";
                            var areaName = "Test";
                            endpoints.MapUmbracoRoute<AuthController>(rootSegment, areaName, areaName);
                        });
                    }
                ));
            });
        }
    }
}

```

If you then run the site and go to `~/umbraco/BackOffice/Test/Auth/` you will be redirected to the login screen if you're not logged in, if you're logged in you will see "Hello from authorized controller"

### DIFF
--- a/src/Umbraco.Web.Common/Controllers/UmbracoAuthorizedController.cs
+++ b/src/Umbraco.Web.Common/Controllers/UmbracoAuthorizedController.cs
@@ -1,0 +1,16 @@
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Umbraco.Cms.Web.Common.Authorization;
+using Umbraco.Cms.Web.Common.Filters;
+
+namespace Umbraco.Cms.Web.Common.Controllers
+{
+    /// <summary>
+    /// Provides a base class for backoffice authorized controllers.
+    /// </summary>
+    [Authorize(Policy = AuthorizationPolicies.BackOfficeAccess)]
+    [DisableBrowserCache]
+    public abstract class UmbracoAuthorizedController : ControllerBase
+    {
+    }
+}


AndyButland:
Code looks fine, but just testing this using your example code, it seems I can access it even when not logged in.

AndyButland:
I see why - my installation was in "upgrade" status, which will allow access (see [here](https://github.com/umbraco/Umbraco-CMS/blob/v9/dev/src/Umbraco.Web.BackOffice/Authorization/BackOfficeHandler.cs#L34) in `BackOfficeHandler`.

Raises a couple of questions though, of which one at least and possibly both are our of scope of this PR:

1. Should this be authorized?  If the intention is the implementers will do something custom, it probably shouldn't be accessible in any runtime state?
2. Not sure why I didn't get redirect to the installation screen.  Seemed I had to manually go to `/install/` to trigger the upgrade.

nikolajlauridsen:
Hmm, that's quite odd indeed, honestly, I don't think it should be authorized, but I'm guessing that it happens because we might need backoffice access without being logged in when installing/upgrading? But I honestly think that this is a separate issue, since it's part of the policy, so that would mean that `UmbracoAuthorizedApiController` and `UmbracoAuthorizedJsonController` has the same issue, since all `UmbracoAuthorizedController` does is apply the policy and disables browser cache, it's really just a helper so people don't have to know what attributes to apply to their controllers.

In regards to you not getting redirected, that's pretty odd too I think, but seems like it's also a separate issue?

AndyButland:
Agree that both are separate from this PR, but think you might raise the discussion on the policy, as could be there's an existing security concern there that should be addressed.

Shazwazza:
FYI, this syntax is super ugly and doesn't need to be this way:

```cs
 options.AddFilter(new UmbracoPipelineFilter(
                    "CustomBackOfficeController",
                    applicationBuilder => {},
                    applicationBuilder => {},
                    applicationBuilder =>
                    {
                        applicationBuilder.UseEndpoints(endpoints =>
                        {
                            // TODO: Do stuff
                        });
                    }
                ));
```

can just be rewritten as

```cs
options.AddFilter(new UmbracoPipelineFilter(nameof(CustomBackOfficeController))
            {
                Endpoints = app => app.UseEndpoints(endpoints =>
                {
                    // TODO: Do stuff
                })
            });
```

nikolajlauridsen:
Ahh I see, yes that's much, much nicer, I'll update it for the documentation 👍

bergmania:
I think we should create tasks on the AB to investigate these findings. I agree there are not related to this PR

PR #10581, opened 2021-07-02
Bugfix: #10414 - Validation message doesn't disappear once the issue is fixed

### LINKED ISSUE #10414: Validation message doesn't disappear once the issue is fixed

8.14.0

### Bug summary

With a textbox set to having a maxmium of 500 characters, after correctly receiving a warning that the text it too long, the error is not removed if you set the textbox to empty.

### Specifics

The error is removed if you type new content, but if you leave the textbox blank you continue to get the error

### Steps to reproduce

1. Create a textbox with a maximum length (e.g. 500)
2. Enter 500+ characters and try to save. You should be correctly prohibited
image

3. Select all characters and delete them, and the error message does not go away.
image

The error message is only removed when typing a text in to the box

As soon as the text length falls within the accepted bounds the error message should be removed.

BatJan:
### Prerequisites

If there's an existing issue for this PR then this fixes #10414 

### Description
This PR ensures the validation message is removed in case one decides to remove all chars at once.

**Before**
textstring-validation-before

**After**
textstring-validation-after

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -136,7 +136,7 @@ You can get in touch with [the core contributors team](#the-core-contributors-te
 
 In order to build the Umbraco source code locally, first make sure you have the following installed.
 
-  * [Visual Studio 2019 v16.3+ (with .NET Core 3.0)](https://visualstudio.microsoft.com/vs/)
+  * [Visual Studio 2019 v16.8+ (with .NET Core 3.0)](https://visualstudio.microsoft.com/vs/)
   * [Node.js v10+](https://nodejs.org/en/download/)
   * npm v6.4.1+ (installed with Node.js)
   * [Git command line](https://git-scm.com/download/)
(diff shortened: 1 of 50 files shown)


OwainWilliams:
Morning @BatJan  - I hope to have a chance to review this over the next day or so. If I've any questions I'll drop a message. 
Thanks for the contrib!

OwainWilliams (approved):

OwainWilliams:
Thanks @BatJan - I've tested this and the validation now works as expected. Great spot!

nibbe02:
@OwainWilliams I'm wondering if the bug remains in version 8.16. Unfortunately, I experience a similar problem with the media selector.

https://user-images.githubusercontent.com/10845659/131214634-4912c4e8-77b6-45d3-ab4a-5de192b0d8cb.mp4

BatJan:
@nibbe02 It's not directly related - The issue fixed here is for the validation of the "textstring" property editor. The issue you showcase above is probably related to the media picker, which has it's own set of logic :-)

If time allows and no one has a look before me I'll see if I can give it a go and find a fix 👍🏻

nibbe02:
Thank you for explaining it to me. I thought these shared validation functionality, but now I know :)

PR #10589, opened 2021-07-03
Adding extension for MapDefinitions collection builder

callumbwhyte:
Registering custom mappers with Umbraco's built in object mapper is a little verbose.

In order to register custom mappers you must first get the collection builder like so:

```
composing.WithCollectionBuilder<MapDefinitionCollectionBuilder>()
```

Most of Umbraco's internal collection builders have short-hand extension methods that can be used instead – e.g. `composing.Components()` – but not for MapDefinitions for some reason... This pull request changes that!

Confusingly an extension for `.Mappers()` already exists, but this relates to the mappers used by the persistence layer. I have tried to distinguish the MapDefinitions accordingly by naming the extension method `.MapDefinitions()`


nul800sebastiaan:
Thanks @callumbwhyte for the help here! 👍

Issue #10577, opened 2021-07-01
Umbraco backoffice allows you to assign the same 'login' on two different Members


fishj123:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

8.13.1

### Bug summary

If there are two member accounts in Umbraco (Member A and Member B) it is possible to change their 'login' field via the backoffice to set them both to the same value.

### Specifics

For example;

Member A can be Saved with a login of 'jackfisher'
Member B can be Saved with a login of 'jackfisher'

### Steps to reproduce

1. Create two different Member accounts in Umbraco either via the backoffice or code.
2. Using the Umbraco backoffice, Save the first member account with a login value of 'jackfisher'
3. Using the Umbraco backoffice, Save the second member account with a login value of 'jackfisher'

### Expected result / actual result

Expected Result is that Umbraco will not allow you to save the Second Member, an error message should be shown informing you of the login which is already in use.

Actual result is that Umbraco allows you to save the Member account and you end up with two Members with the same login.

nul800sebastiaan:
Thanks @fishj123! That's interesting! We do validate if an email or login exist when first creating the user but clearly not when updating their login! 😬 

We'd love some help getting this fixed up! 👍 


https://user-images.githubusercontent.com/304656/124142695-bc094c00-da8a-11eb-97db-504cf3f662d6.mp4

umbrabot:
Hi @fishj123,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/10591

PR #10591, opened 2021-07-04
#10577 bugfix - Umbraco backoffice allows you to assign the same 'login' on two different Members 

hannakfoo:
### Prerequisites

If there's an existing issue for this PR then this fixes [#10577 ](https://github.com/umbraco/Umbraco-CMS/issues/10577)

### Description
While investigating the issue  [#10577 ](https://github.com/umbraco/Umbraco-CMS/issues/10577) I stumbled upon the UmbracoMemberShipProvider > FindUsersByName not matching the exact same username. While one would expect it on this place. So now when updating a member login with the exact same username used by another member it returns the exception that the username is already in use . I checked if there weren't any other references to the method not to break some things on other places.

**Before**

https://user-images.githubusercontent.com/3263156/124398156-8b3c4780-dd14-11eb-9e0a-362a7204ef1e.mp4

**After**

https://user-images.githubusercontent.com/3263156/124397766-28e24780-dd12-11eb-849b-33c350b0d802.mp4


nul800sebastiaan:
Interesting find there @hannakfoo! 👍 

I just want to check with @Shazwazza if this was set to `Wildcard` on purpose for some reason? It's been 8 years, so it might be hard to remember 🙈 

This is a public method on a public class, so it will most likely break something for people using this method. 

I think we might want to add a method with an param of `StringPropertyMatchType`, that defaults to `Wildcard` and use that as the one Umbraco uses internally (passing in `Exact` when we need to match the login). 
Then external consumers can keep using the current method which calls the new method and always uses `Wildcard` like before. 

But let's wait to see what @Shazwazza has to say about that!

Shazwazza:
I think this just goes back to backward compat way back in the day. Here's the orig change from the very very old code to use the member service: https://github.com/umbraco/Umbraco-CMS/commit/2dd04799b216651314e780776e88e547290c7b8f#diff-7c81ffa1c1861136697bac6d988ec7631428cdd4f7bf2e710a820317f1c45417L424

Before that, it used the old `Member.GetMemberByName` method and the orig code looks like https://github.com/umbraco/Umbraco-CMS/commit/f6d0d043b5701713cb5163e3e1766c38921f09bb#diff-d12145d93aaa15c96f5feb5d8039749f6fde81b06621359589ae8b13c96eec6aR245 which did a wildcard query on the username ... therefore that logic was ported over.

This differs from the aspnet default SQL membership provider implementation which allows the `usernameToMatch` string to have it's own wildcard characters and it internally handles security issues relating to sql injection, etc... https://github.com/microsoft/referencesource/blob/master/System.Web/Security/SQLMembershipProvider.cs#L1348.

However, this is just another reason why the old membership provider API is a broken API. There is no clear indication of what is 'supposed' to happen. The docs for this method: https://docs.microsoft.com/en-us/dotnet/api/system.web.security.membershipprovider.findusersbyname?view=netframework-4.8 purely state: `If your data source supports additional search capabilities, such as wildcard characters, you can provide more extensive search capabilities for user names.` Which means it totally up to the provider to determine how to handle the string or to do the search how it feels like doing it.

So the implementation in Umbraco isn't 'wrong', it's just how it was written, but it does differ from the implementation of the Sql membership provider.

With all that said, I have no idea what the implications of making this change are but I feel like it will break other parts of Umbraco and/or other people's implementation. 

To me, it sounds like the bug is not related to this particular method but the code that calls this method to determine if the username already exists. In theory this bug would exist for the unique email check to because that query also uses a wildcard query - so does this bug exist for email too? Else if it doesn't, then there is probably code that correctly checks the uniqueness of the email and that similar code is not being applied to login names.

hannakfoo:
@Shazwazza Thank you for your clear feedback and analyses of the history of this code. Your suggestion to take a look at the calling method (so not to break anything) seems to me the right direction. I will change my solution and research the mechanism of the unique email check and report back!

nul800sebastiaan:
Wow, nice find there @hannakfoo, clearly it was just a copy/paste that was never updated correctly. Amazing that this hasn't been found for years! 😅 

I'll let the automated build run before I merge this but should be all good!

nul800sebastiaan:
Brilliant, all good and what a wonderful fix and what an amazing oversight, this code has not changed in years 😅 

I have reverted [your change from `Wildcard` to `Exact`](https://github.com/umbraco/Umbraco-CMS/pull/10591/commits/759eb34e36283c8d3d4ced262f476872f6ba16c0) since with the fix of checking the correct property it was no longer necessary.  

Thanks again for your contribution and incorporating our feedback, well done and you've earned yourself a shiny new contributor badge on Our Umbraco https://our.umbraco.com/members/hannakfoo 🥇

nzdev:
Shouldn't the loginName column on the cmsMember table have a unique index applied?

nul800sebastiaan:
There are legitimate situations where the login could be allowed to be a duplicate, for example if it's for a different membertype. Rare, but I've seen it being used sometimes.

nzdev:
In that case a unique index could be on the loginName and the membertype?

nul800sebastiaan:
Maybe.. 🙂
Email is also often used as the login, so also email + membertype
But membertype is not in the CmsMember table and I am rusty on my db stuff, but I don't think you can make a unique index over 2 tables? :)

For users the `UserLogin` should be unique and the `UserEmail` probably as well.

PR #10597, opened 2021-07-05
Fix the introduction tour getting lost if dashboard is not shown

### LINKED ISSUE #10570: v9-beta004: Initial Backoffice tour gets lost..

v9-beta004

### Bug summary

Initial tour of the backoffice got to users (no next button on the dialog)

Then resumed and got this.
Screenshot 2021-07-01 122241

Starting the tour again from the help sidebar.. get to the user step
image

### Specifics

_No response_

### Steps to reproduce

Simply followed the v9 install here https://our.umbraco.com/documentation/UmbracoNetCoreUpdates#umbraco-9-net-core-beta
and first run started to walk through the backoffice tour.

Completed the tour without getting "lost" ;-)

nikolajlauridsen:
As the title states, the introduction tour will get lost and break if you've had a poke around and have hidden the dashboard before starting the tour, for instance if you start while creating a document type or so on.

I've added a conditional step that's only shown if the dashboard is not, which guides the user to click on the Content section if the dashboard is not shown. I realise that the dashboard could potentially by hidden from the user trying to create a content node, and then it might be a bit misleading to say "click the content button to enter the content section", however clicking the content button when creating content does navigate you back and shows the dashboard, so I think it's still an improvement to have it here.

Now I'm no writer, so I'm not sure if the contents of the step is perfect.

This was initially pointed out in an issue, so this fixes #10570

### Testing
1. Run `npm run build` from the `Umbraco.Web.UI.Client` project
2. Start the tour while the dashboard is not shown
3. Ensure that the tour does not get stuck


nathanwoulfe (approved):
Thanks @nikolajlauridsen, this works nicely - no more getting lost when staring the tour from outside the content section. I've added a change to the content for the main menu step, to remove the reference to being in the content section, since we know (given your PR) that this is not always the case.

Will wait for a green test suite then hit the merge button 👍

Issue #10593, opened 2021-07-05
Title of property info button not localized


bjarnef:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

8.14.0

### Bug summary

The `<umb-property-info-button>` was added to block list configuration in https://github.com/umbraco/Umbraco-CMS/commit/6a5bfcaa9c7eb6f04fc124bdb72aba3f5dcc9fe2#diff-25d22fe2efcdd341dfb661b2ece3b5882de2be7b446c6fbacc1e4ef8383ee7fe

However it seems the localization doesn't work correct.

![image](https://user-images.githubusercontent.com/2919859/124442602-32ed6000-dd7d-11eb-8cfa-d4b13842d962.png)

It seems it later was localized in https://github.com/umbraco/Umbraco-CMS/commit/9882edeca098bcae9f96384f5af87cf216e671e8#diff-836ee8c2a3addc755dcbadc5e674a38966e085020554a95d397325136e6de5f2

### Steps to reproduce

- Add a new datatype based on Block List editor
- Add a new element type to "Available blocks"
- In configuration of the block list element type not the tooltip for the "info" icon at custom view isn't localized.

umbrabot:
Hi @bjarnef,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
I think the issue here is that it pass in `@blockEditor_labelCustomViewInfoTitle` as string to the component before the `localize` attribute directive has localized text.

PR #10600, opened 2021-07-05
Fix localization of title attribute in umb-property-info-button

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10593

### Description
This PR fixes localization of title attribute in the `<umb-property-info-button>`.

I don't think the `localize` attribute directive will work in this case (or it actually ever has worked here) as it seems, it will just pass in the `@blockEditor_labelCustomViewInfoTitle` as string to the component before the localize attribute directive has localized the text.

The `buttonTitle` should still work for static text, while `buttonTitleKey` will work for localized title similar to `label` and `labelKey` in many other components.

xYKUokLTG0


nul800sebastiaan:
Thanks @bjarnef! Nice update!

PR #10603, opened 2021-07-06
Fix duplicate ID when creating document types with SQL CE

### LINKED ISSUE #10381: Install Package containing Document Types Fails if using SQL Server Compact

Umbraco 8.14.0

### Bug summary

This issue only occurs when using SQL Server Compact.

### Specifics

_No response_

### Steps to reproduce

If you attempt to install a package containing document types, records will be inserted into table cmsPropertyTypeGroup.

This table now contains 4 new records relating to the new Media Picker. IDs of these new Media Types are 52,53,54,55

If you install a package you will get error 'A duplicate value cannot be inserted into a unique index.' error on the id column.

For some reason which I cannot figure out SQL Server Compact does not set the identity seed for the ID column to start at ID 56 where as SQL Server does not have this issue.

If you run the following SQL command before attempting to install a package containing document types you will not get the issue and the package will install successfully:

ALTER TABLE [cmsPropertyTypeGroup] ALTER COLUMN [Id] IDENTITY (56,1)

_No response_

---
_This item has been added to our backlog [AB#12326](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/12326)_

nikolajlauridsen:
This fixes #10381. 

Turns out it was not a problem with installing packages containing document types, but instead that you simply can't create more than 39 property type groups at all before getting this error 😱.

This is because when the new media types were adding, 4 new property type groups were also added with ID 52, 53, 54, and 55, however, the identity seed was not updated to 56, but was still the old 12, meaning that once you tried to create your 40th property group the ID would be 52 and you will get a duplicate value error because the ID must be unique. This also means that this has been a problem since the new media types were added (Video, Audio, Article, Vector Graphics).

This was only a problem with SQL CE because for SQL Server we don't use the seed to determine what the next ID should be (I honestly think SQL Server takes care of that for us?), but this can also be seen in `SqlServerSyntaxProvider.GetIdentityString` where we always return 1 for the seed:

```C#
private static string GetIdentityString(ColumnDefinition column)
{
    return "IDENTITY(1,1)";
}
```

The fix for new installs is pretty simple, I just set the identity seed to 56 as to not cause collisions. I have also tried to create a migration for existing sites which sets the seed to 56, however, I've never touched migrations, so I'm not 100% I've done everything right.

### Testing

To help with testing I've created this package: [PackageWithManyPropertyGroups_1.0.0.zip](https://github.com/umbraco/Umbraco-CMS/files/6769617/PackageWithManyPropertyGroups_1.0.0.zip) which contains a document type with 36 groups, so you don't have to create very many before hitting the error.

1. Create a fresh site from v8/dev or similar
2. Install the package provided, and try creating 5 or so more groups, and you should see the error
3. Checkout this branch
4. Make sure the migration runs
5. Try creating more groups now, you should not see the issue anymore

You might also want to try and create a fresh project from this branch and ensure that the problem is gone.

### DIFF
--- a/src/Umbraco.Core/Migrations/Upgrade/V_8_15_0/UpdateCmsPropertyGroupIdSeed.cs
+++ b/src/Umbraco.Core/Migrations/Upgrade/V_8_15_0/UpdateCmsPropertyGroupIdSeed.cs
@@ -1,0 +1,21 @@
+using NPoco;
+using Umbraco.Core.Persistence;
+using Umbraco.Core.Persistence.Dtos;
+
+namespace Umbraco.Core.Migrations.Upgrade.V_8_15_0
+{
+    public class UpdateCmsPropertyGroupIdSeed : MigrationBase
+    {
+        public UpdateCmsPropertyGroupIdSeed(IMigrationContext context) : base(context)
+        {
+        }
+
+        public override void Migrate()
+        {
+            if (DatabaseType.IsSqlCe())
+            {
+                Database.Execute(Sql("ALTER TABLE [cmsPropertyTypeGroup] ALTER COLUMN [id] IDENTITY (56,1)"));
+            }
+        }
+    }
+}
--- a/src/Umbraco.Core/Migrations/Upgrade/UmbracoPlan.cs
+++ b/src/Umbraco.Core/Migrations/Upgrade/UmbracoPlan.cs
@@ -206,6 +206,7 @@ namespace Umbraco.Core.Migrations.Upgrade
             // to 8.15.0...
             To<AddCmsContentNuByteColumn>("{8DDDCD0B-D7D5-4C97-BD6A-6B38CA65752F}");
             To<UpgradedIncludeIndexes>("{4695D0C9-0729-4976-985B-048D503665D8}");
+            To<UpdateCmsPropertyGroupIdSeed>("{5C424554-A32D-4852-8ED1-A13508187901}");
 
             //FINAL
         }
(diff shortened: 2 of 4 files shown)


src/Umbraco.Core/Migrations/Upgrade/V_8_15_0/UpdateCmsPropertyGroupIdSeed.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public class UpdateCmsPropertyGroupIdSeed : MigrationBase
+    {
+        public UpdateCmsPropertyGroupIdSeed(IMigrationContext context) : base(context)
+        {
+        }
+
+        public override void Migrate()
+        {
+            if (DatabaseType.IsSqlCe())
```
warrenbuckley, line 15: Is it worth performing this SQL statement regardless of DB type ?

nikolajlauridsen: Since we always use `IDENTITY(1,1)` for SQL Server I honestly don't think it makes sense to make this change to other types than SqlCe, since these are we only two types we support. Should we add support for another database type this won't be necessary since the identity seed on `PropertyTypeGroupDto` has been update 😄

warrenbuckley:
Works as described :)

PR #10607, opened 2021-07-06
Update MemberAuthorizeAttribute.cs

mistyn8:
Authenticated but not in group/type/members created a circular login cycle (with OAuth)
override HandleUnauthorizedRequest to correctly return forbidden (403) rather than unauthorized (401)

Might also want to extend further to pass a no-access URI?
```
//var response = actionContext.Request.CreateResponse(HttpStatusCode.Redirect);
//response.Headers.Location = new Uri("https://....");
//actionContext.Response = response;
```

### DIFF
--- a/src/Umbraco.Web/WebApi/MemberAuthorizeAttribute.cs
+++ b/src/Umbraco.Web/WebApi/MemberAuthorizeAttribute.cs
@@ -1,5 +1,6 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Web.Http;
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Uh oh, the build failed @mistyn8 🙈 

`##[error]src\Umbraco.Web\WebApi\MemberAuthorizeAttribute.cs(53,59): Error CS0246: The type or namespace name 'HttpActionContext' could not be found (are you missing a using directive or an assembly reference?)`

mistyn8:
oops missed off a required using.

PR #10615, opened 2021-07-07
V9: Fix GetLocalCropUrl

nikolajlauridsen:
This fixes a bug which would cause your page to be unable to render if you used `@Model.MyImage.GetLocalCropUrl("someAlias")`. 

This was caused by `GetLocalCropUrl` calling itself recursively and causing a stack overflow. I used the same logic as in V8, and it seems to work fine, I also marked it as obsolete like in V8, letting people know they should use `GetCrop` instead


Zeegaan:
Tested this, and this solution fixes the problem!

PR #10616, opened 2021-07-07
Move UrlHelper extensions to Umbraco.Extensions

mattbrailsford:
All other extensions in v9 are in the `Umbraco.Extensions` namespace so these should be moved to the same namespace for consistency.

### DIFF
--- a/src/Umbraco.Web/UrlHelperRenderExtensions.cs
+++ b/src/Umbraco.Web/UrlHelperRenderExtensions.cs
@@ -7,7 +7,7 @@ using Umbraco.Core.Models;
 using Umbraco.Extensions;
 using Umbraco.Web.Composing;
 
-namespace Umbraco.Web
+namespace Umbraco.Extensions
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


bergmania:
Hi @mattbrailsford 

Whatever is in `Umbraco.Web` project is not shipped as part of the v9 releases. `Umbraco.Web` is the remaining parts of v8, that we still need to confirm are migrated or whether it should.

mattbrailsford:
@bergmania yea, I just noticed this when investigating further. I'm ultimately trying to get the `SurfaceAction` extensions to work, but I don't know what the v9 equivalent is 🤔

bergmania:
It seems to be a method that needs to be migrated.

mattbrailsford:
Ok, so we can't currently do surface actions? Dang, ok. Will have to see if I can work around it 🤔

bergmania:
I think this should do it: https://github.com/umbraco/Umbraco-CMS/pull/10618

PR #10619, opened 2021-07-07
Changes the RoslynCompiler to use assemblies resolved from the DependencyContext

Shazwazza:
This fixes an issue with invalid framework references when using ModelsBuilder InMemoryAuto mode. We would end up with exceptions like:

```
Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation.CompilationFailedException: One or more compilation failures occurred:
D:\Projects\Umbraco-CMS-netcore\src\Umbraco.Web.UI.NetCore\Views\Products.cshtml(28,24): error CS0012: The type 'IEnumerable<>' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'.
D:\Projects\Umbraco-CMS-netcore\src\Umbraco.Web.UI.NetCore\Views\Products.cshtml(30,51): error CS0012: The type 'IEnumerable<>' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'.
D:\Projects\Umbraco-CMS-netcore\src\Umbraco.Web.UI.NetCore\Views\Products.cshtml(35,100): error CS0012: The type 'Decimal' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'.
```

And we'd get this exact error about decimal: https://github.com/dotnet/core/issues/2082 

This changes our internal compiler to use the DependencyContext as referenced in https://github.com/dotnet/core/issues/2082 to add refs to the compiler. Before we were manually adding refs to the runtime libs - which may have almost worked but potentially we weren't adding enough or incorrect ones. Before we were adding refs using `AssemblyLoadContext.All.SelectMany(x => x.Assemblies)` but this is prob incorrect for adding refs to the roslyn compiler - also because this would probably only include assemblies already loaded into the context and some may have been missing.

This change works for me but needs thorough testing.

## Testing

This is with InMemoryAuto mode.

Prior to this, if you created a decimal property editor on a doc type, save the doc type and render that property in a template, then render the page, you'd get an exception like the above.

With this change, you will probably need to go re-save your doc type and then re-render your page so that re-compilation occurs and it should work.

### DIFF
--- a/src/Umbraco.Infrastructure/ModelsBuilder/Building/Builder.cs
+++ b/src/Umbraco.Infrastructure/ModelsBuilder/Building/Builder.cs
@@ -19,12 +19,10 @@ namespace Umbraco.Cms.Infrastructure.ModelsBuilder.Building
     /// </summary>
     public abstract class Builder
     {
-        private readonly IList<TypeModel> _typeModels;
-
         protected Dictionary<string, string> ModelsMap { get; } = new Dictionary<string, string>();
 
         // the list of assemblies that will be 'using' by default
-        protected readonly IList<string> TypesUsing = new List<string>
+        protected IList<string> TypesUsing { get; } = new List<string>
         {
             "System",
             "System.Linq.Expressions",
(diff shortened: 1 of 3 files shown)


bergmania (approved):
Tested and it works 💪

PR #10623, opened 2021-07-08
V9: Fix ImageSharp integration and add support for custom crops

nikolajlauridsen:
This fixes an issue where custom crops was not reflected if you zoomed in or moved the crop around.

This was because we had replace `ImageProcessor` with `ImageSharp.Web`, which uses different parameters and doesn't support the type of cropping we use out of box.

To fix this I created a custom Crop `IImageWebProcessor` which grabs the old crop parameter from the URL string, and recalculate it into the ImageSharp `Rectangle` format, and then crop the image, which will be sent on the resizer where it will be resized. 

I also considered creating a custom resizer that uses the TargetRectagle and the manual mode, however, I think this will be much harder to manage and maintain, than separating the concerns of cropping and resizing. The documentation is not entirely clear if it is at all possible to do it in this manner regardless.

### Testing

To this out, you must 
1. Create a document type that uses the media picker
2. Specify a crop on the picker
3. Render the media in your template
3. Create a piece of content with an image in the media picker
   1. Specify your crop on the image, to see the difference you must move the crop and/or use the slider to resize it
4. Navigate to the page and see the different pre/post PR 

(You might have to clear your browser cache) 

I made an image with a custom crop, before the changes it looked like:

before

After the changes:

after


ronaldbarendse (changes requested):

src/Umbraco.Web.Common/ImageProcessors/CropWebProcessor.cs
```diff
@@ -0,0 +1,79 @@
...
+            return new ImageCropperCropCoordinates()
+            {
+                X1 = ParseDecimal(crops[0]),
+                Y1 = ParseDecimal(crops[1]),
+                X2 = ParseDecimal(crops[2]),
+                Y2 = ParseDecimal(crops[3])
+            };
+        }
+
+        private static decimal ParseDecimal(string decimalString) => decimal.Parse(decimalString, CultureInfo.InvariantCulture);
```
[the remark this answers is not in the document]
nikolajlauridsen: Yeah great idea, what a nifty feature, I've added this but changed the float to a decimal since `ImageCropperCropCoordinates` excepts decimal and not float 😄

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilder.ImageSharp.cs
```diff
@@ -51,6 +52,8 @@ public static IServiceCollection AddUmbracoImageSharp(this IUmbracoBuilder build
                 .SetCache<PhysicalFileSystemCache>()
                 .SetCacheHash<CacheHash>()
                 .AddProvider<PhysicalFileSystemProvider>()
+                .ClearProcessors()  // ImageSharp includes the processors by default, so we have to clear and re-add to control the order
```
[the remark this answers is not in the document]
nikolajlauridsen: That's a very good point, I've changed it to only remove the `ResizeWebProcessor`, it's a bit weird though that it seems like it was added by default, so I'm wondering if we're also adding `FormatWebProcessor` and `BackgroundColorWebProcessor` twice, or maybe ImageSharp handles that for us like we have `AddUnique`

src/Umbraco.Web.Common/ImageProcessors/CropWebProcessor.cs
```diff
@@ -0,0 +1,79 @@
...
+            // Get coordinates of top left corner of the rectangle
+            var topX = decimal.ToInt32(width * coordinates.X1);
+            var topY = decimal.ToInt32(height * coordinates.Y1);
+            // Get coordinated of bottom right corner
+            var bottomX = decimal.ToInt32(width - (width * coordinates.X2));
+            var bottomY = decimal.ToInt32(height - (height * coordinates.Y2));
+
+            // Get width and height of crop
+            var cropWidth = bottomX - topX;
+            var cropHeight = bottomY - topY;
```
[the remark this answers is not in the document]
nikolajlauridsen: Yeah that might be a good idea, I initially didn't since it could only be off by 1 pixel, but I guess it's better to be more accurate, I've changed it so we now round the values before converting them to ints.

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilder.ImageSharp.cs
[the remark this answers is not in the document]
nikolajlauridsen: I've changed it back to clear processors like we talked about, to make sure the order is correct 😄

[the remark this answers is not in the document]
ronaldbarendse: Looks like [`AddProvider`](https://github.com/SixLabors/ImageSharp.Web/blob/15cb36a034d52a18e364e901e028f8991adad4d5/src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs#L135) uses [`TryAddEnumerable`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.extensions.servicecollectiondescriptorextensions.tryaddenumerable?view=dotnet-plat-ext-5.0&viewFallbackFrom=net-5.0), so it won't add the service(s) of the same type more than once 👍🏻

The ImageSharp sample isn't quite clear on this, as it does the same on the [default setup](https://github.com/SixLabors/ImageSharp.Web/blob/d0559520245dffc484cd7fc77ae35fd20f0ff3b9/samples/ImageSharp.Web.Sample/Startup.cs#L43-L62), but clears them on the [custom one](https://github.com/SixLabors/ImageSharp.Web/blob/d0559520245dffc484cd7fc77ae35fd20f0ff3b9/samples/ImageSharp.Web.Sample/Startup.cs#L107-L139).

I think it's same to clear them first though, so we are sure they are added in the correct order. We only need to make sure this is run before any custom code does (e.g. from composers, but I think that will be the case)...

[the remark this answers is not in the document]
nikolajlauridsen: Yeah I'm thinking it's best to clear them first to make sure the order is as we want it, and they should be run before any custom code since they're added when we do `.AddBackOffice`, which is before `.AddComposers()`, so you should be able to add your own processors with no issues.

Jeavon:
I just found the ImageSharp.Web query string variable names were incorrect, for example `centre `should by `rxy `and `mode `should be `rmode`.  I think there are more in ImageSharpImageUrlGenerator.cs  that need to be changed

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilder.ImageSharp.cs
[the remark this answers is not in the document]
JimBobSquarePants: Order of processors doesn't matter (and shouldn't). Processors should have unique, non-collision inducing commands.

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilder.ImageSharp.cs
```diff
@@ -50,7 +51,9 @@ public static IServiceCollection AddUmbracoImageSharp(this IUmbracoBuilder build
                 })
                 .SetCache<PhysicalFileSystemCache>()
                 .SetCacheHash<CacheHash>()
-                .AddProvider<PhysicalFileSystemProvider>()
+                .ClearProcessors()
+                .RemoveProcessor<ResizeWebProcessor>() // The Resize processor is added by default, so remove it to ensure that the crop processor runs first
```
JimBobSquarePants, line 55: Unnecessary.

src/Umbraco.Web.Common/ImageProcessors/CropWebProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+        public const string Crop = "crop";
+
+        public FormattedImage Process(
+            FormattedImage image,
+            ILogger logger,
+            IDictionary<string, string> commands,
+            CommandParser parser,
+            CultureInfo culture)
+        {
+            ImageCropperCropCoordinates cropCoordinates = GetCropCoordinates(commands, parser, culture);
```
JimBobSquarePants, line 28: Use the most common result.
```c#
ImageCropperCropCoordinates cropCoordinates = GetCropCoordinates(commands, parser, culture);

if (cropCoordinates != null)
{
    Size size = image.Image.Size();
    Rectangle crop = GetCropRectangle(size.Width, size.Height, cropCoordinates);
    image.Image.Mutate(x => x.Crop(crop));
}

return image;
```

src/Umbraco.Web.Common/ImageProcessors/CropWebProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// Gets a rectangle that is calculated from crop coordinates.
+        /// </summary>
+        /// <param name="width">The width of the image being cropped.</param>
+        /// <param name="height">The height of the image being cropped.</param>
+        /// <param name="coordinates">Coordinate set to calculate rectangle from.</param>
+        /// <returns>Rectangle with the position and sized described in coordinates.</returns>
+        private static Rectangle GetCropRectangle(int width, int height, ImageCropperCropCoordinates coordinates)
+        {
+            // Get coordinates of top left corner of the rectangle
+            var topX = RoundToInt(width * coordinates.X1);
```
JimBobSquarePants, line 54: `int topX` etc....

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilder.ImageSharp.cs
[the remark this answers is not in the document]
ronaldbarendse: Just to be sure/clear: the order of the processors added to the options/services doesn't matter, but the order in which the commands are added to the query string does, right?

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilder.ImageSharp.cs
ronaldbarendse: If the order of adding processors doesn't matter and the default ones are already added by `AddImageSharp()`, it's unnecessary to clear and explicitly add these... So we should only add our custom processor here.

src/Umbraco.Web.Common/ImageProcessors/CropWebProcessor.cs
ronaldbarendse: Agree, having a single return statement at the bottom of the function is a lot more readable 😃

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilder.ImageSharp.cs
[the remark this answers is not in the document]
JimBobSquarePants: Yes, that’s correct. The API is imperative

ronaldbarendse:
I've removed any Umbraco dependencies in `CropWebProcessor` and cleaned it up to be more or less the same as the [Crop processor from ImageProcessor](https://github.com/JimBobSquarePants/ImageProcessor/blob/4af8de4a6376d9426fadd5c4596bc1de187899c0/src/ImageProcessor/Processors/Crop.cs). This should allow it to be integrated into ImageSharp or used outside of Umbraco (e.g. to offload the image processing on Umbraco Cloud).

Like @Jeavon already mentioned in https://github.com/umbraco/Umbraco-CMS/pull/10715, we also need to look into `upscale`, `animationprocessmode`, `heightratio` and `widthratio` parameters.

JimBobSquarePants:
>Like @Jeavon already mentioned in #10715, we also need to look into upscale, animationprocessmode, heightratio and widthratio parameters.

I wouldn't recommend a like-for-like feature swap. The idea should be that you have the most limited DDOS attack surface available.

src/Umbraco.Web.Common/DependencyInjection/UmbracoBuilder.ImageSharp.cs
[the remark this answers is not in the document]
ronaldbarendse: @JimBobSquarePants The ImageSharp.Web API _does_ use the order of processors, as I get the resized image width/height when the `CropWebProcessor` is added last, but the `crop` query string is the first one in the URL. 

The following code confirms this behavior:
https://github.com/SixLabors/ImageSharp.Web/blob/d0559520245dffc484cd7fc77ae35fd20f0ff3b9/src/ImageSharp.Web/Processors/WebProcessingExtensions.cs#L38-L41.

Another thing to note is that this will run every (DI registered) processor, even when none if the commands matches!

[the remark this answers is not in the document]
JimBobSquarePants: That’s a bug. Can you please raise an issue.

ronaldbarendse:
This PR gradually converted from adding support for crop coordinates (something `ImageProcessor.Web` supported, but `ImageSharp.Web` currently doesn't) into ensuring the whole image URL integration works correctly. As @JimBobSquarePants mentions, the idea for ImageSharp is to limit the DDOS attack surface, so I've tried to keep the API as minimal as possible.

We now add a custom `CropWebProcessor` to the ImageSharp middleware. Currently, this needs to be in front of the `ResizeWebProcessor`, so images are cropped before being resized - this looks like an issue within ImageSharp, as the query string order should be used.

This processor adds back the following image URL parameters:
- `crop` - defaults to 4 comma separated pixel values for the X, Y, width and height to use as crop area, but can be switched to percentages that specify the top/left and bottom/right X and Y coordinates;
- `cropmode` - defines how to use the crop values, defaults to `pixels`, but Umbraco will always use `percentage` (as the Image Cropper/Media Picker v3 stores relative values - which don't have to be updated when the original image is resized/scaled).

------------

I've included the changes from PR https://github.com/umbraco/Umbraco-CMS/pull/10715 and cleaned up the following:
- `center` = `rxy` - defines the comma separated X and Y coordinates used as center (focal point) when resizing. Resizing is triggered when either `width` or `height` is added to the URL;
- `anchor` = `ranchor` - defines a named parameter to align the image when resizing (similar to the X,Y coordinates, the names from [`ImageCropAnchor`](https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web/Models/ImageCropAnchor.cs) match ImageSharps [`AnchorPositionMode`](https://docs.sixlabors.com/api/ImageSharp/SixLabors.ImageSharp.Processing.AnchorPositionMode.html) enum);
- `mode` = `rmode` - the resize mode (the names from [`ImageCropMode`](https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web/Models/ImageCropMode.cs) match ImageSharps [`ResizeMode`](https://docs.sixlabors.com/api/ImageSharp/SixLabors.ImageSharp.Processing.ResizeMode.html) enum);
- `upscale` - removed, as you don't want different image sizes returned depending on the original image dimension and you can otherwise use `rmode=min`;
- `animationprocessmode` - removed, as the memory issues don't exist anymore in ImageSharp;
- `widthratio` - removed, as this isn't supported by ImageSharp and we can calculate the missing dimension on the server;
- `heightratio` - removed, same as `widthratio`.

For a complete overview of the built-in `ImageSharp.Web` processing API, I'd recommend checking the documentation: https://docs.sixlabors.com/articles/imagesharp.web/processingcommands.html.

The `ImageUrlGenerationOptions` now only contain supported properties (and new ones can otherwise be added again or use the `FurtherOptions` property). I first made the unsupported properties (`WidthRatio`, `HeightRatio`, `DefaultCrop`, `UpScale` and `AnimationProcessMode`) obsolete, but as this needs to get merged before v9 gets released and we had other breaking changes, I've completely removed them...

ronaldbarendse:
When testing the changes, I noticed the focal point didn't work as expected. This is because ImageProcessors `center` parameter expects top/left coordinates, but the ImageSharp `rxy` parameter has these values swapped (x = left, y = top).

Because this change isn't backwards compatible, I also updated the `CropWebProcessor` to only allow crop coordinates using the `cc` parameter and remove the `CropMode` enum and requirement to add `&cropmode=percentage` to every generated image URL that uses a crop. As a bonus, this will make the URLs a little bit shorter too 😄

src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
```diff
@@ -1,66 +1,68 @@
...
 
         /// <summary>
-        /// The focal point position, in whatever units the registered IImageUrlGenerator uses,
-        /// typically a percentage of the total image from 0.0 to 1.0.
+        /// The focal point position, in whatever units the registered IImageUrlGenerator uses, typically a percentage of the total image from 0.0 to 1.0.
         /// </summary>
         public class FocalPointPosition
         {
-            public FocalPointPosition (decimal top, decimal left)
+            public FocalPointPosition(decimal left, decimal top)
```
ronaldbarendse, line 35: I've changed the parameter order of this constructor to align it with the way it's used by ImageSharp and the property order below (including the order in which they are set).

We might also want to change the names to `x` and `y`. And while we're doing that, change from `decimal` to `float` (as we don't need the additional precision) and convert this class into a struct, as that has additional performance improvements: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct.

src/Umbraco.Infrastructure/Media/ImageSharpImageUrlGenerator.cs
```diff
@@ -1,68 +1,88 @@
...
-using Umbraco.Extensions;
 
 namespace Umbraco.Cms.Infrastructure.Media
 {
     public class ImageSharpImageUrlGenerator : IImageUrlGenerator
     {
-        public IEnumerable<string> SupportedImageFileTypes => new[] { "jpeg", "jpg", "gif", "bmp", "png" };
+        private static readonly string[] s_supportedImageFileTypes = Configuration.Default.ImageFormats.SelectMany(f => f.FileExtensions).ToArray();
+
+        public IEnumerable<string> SupportedImageFileTypes { get; } = s_supportedImageFileTypes;
```
ronaldbarendse, line 16: This should ensure all image extensions supported by ImageSharp are handled. I'm not sure if `Configuration.Default` reflects any custom added formats though (e.g. when WebP support is added as separate package). @JimBobSquarePants can you confirm this is the right way to get all supported image formats?

src/Umbraco.Infrastructure/Media/ImageSharpImageUrlGenerator.cs
```diff
@@ -1,68 +1,88 @@
...
             {
-                imageProcessorUrl.Append("?mode=").Append((options.ImageCropMode ?? ImageCropMode.Crop).ToString().ToLower());
+                imageUrl.Append(queryStringHasStarted ? '&' : '?');
+                queryStringHasStarted = true;
 
-                if (options.ImageCropAnchor != null) imageProcessorUrl.Append("&anchor=").Append(options.ImageCropAnchor.ToString().ToLower());
+                imageUrl.Append(value);
             }
+            void AddQueryString(string key, params IConvertible[] values)
+                => AppendQueryString(key + '=' + string.Join(",", values.Select(x => x.ToString(CultureInfo.InvariantCulture))));
```
ronaldbarendse, line 36: I've added these local methods to make it easier to add query string parameters, so we don't have to deal with conditionally adding the first `?` or constructive `&` and always format the values using the invariant culture.

src/Umbraco.Infrastructure/Media/ImageSharpImageUrlGenerator.cs
```diff
@@ -1,68 +1,88 @@
...
+            }
+
+            if (options.Quality.HasValue)
+            {
+                AddQueryString("quality", options.Quality.Value);
+            }
+
+            if (!string.IsNullOrWhiteSpace(options.FurtherOptions))
+            {
+                AppendQueryString(options.FurtherOptions.TrimStart('?', '&'));
```
ronaldbarendse, line 75: This ensures the further options is always added correctly (e.g. with a leading `?` when it's the first/only query string added) and you don't have to prefix this value with `&` anymore.

ronaldbarendse:
Besides the testing steps from the initial PR comment, also make sure to check whether changing the focal point uses the correct center position when resizing/cropping:
- Add a portrait and landscape media image;
- Render these images as a square image on the front-end, e.g. ``;
- When changing the focal point, the rendered image should show that part of the image (e.g. more of the top/bottom on the portrait version and left/right on the landscape one).

src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
AndyButland: I think this is probably OK named as is.  `x` would correspond to `left`, but `y` arguably makes more sense related to the `bottom` rather than `top`, as typical X/Y coordinates are plotted.  With `left` and `top` we resolve any ambiguiity.

ronaldbarendse: It's more to align to the naming used by Microsoft and ImageSharp for [rectangles](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.rectangle.-ctor?view=net-5.0) and [points](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.point?view=net-5.0). Both have aliases for left and top (only property getters), but the constructors always reference X and Y.

Also keep in mind that these aren't coordinates as used in graphs and images use the upper-left corner as origin in basically all editing programs.

AndyButland (commented):
Here's where I've got to so far - have reviewed code, and looks good.  I fixed up a couple of tiny things and pushed up some unit tests for the `CropWebProcessor` and `ImageCropperTemplateCoreExtensions`.

Will add an inline comment on some code that seems like it's not adding anything, and we could perhaps remove?

I'm unfortunately not understanding how to use a crop on the front-end... so I need some more instruction there!  Probably something obvious that I'm missing,,, have shared on a Slack message.

src/Umbraco.Infrastructure/Media/ImageSharpImageUrlGenerator.cs
```diff
@@ -1,68 +1,88 @@
...
+            {
+                AddQueryString("height", options.Height.Value);
+            }
+
+            if (options.Quality.HasValue)
+            {
+                AddQueryString("quality", options.Quality.Value);
+            }
+
+            if (!string.IsNullOrWhiteSpace(options.FurtherOptions))
```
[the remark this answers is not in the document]
AndyButland: Have applied.

src/Umbraco.Web.Common/Extensions/ImageCropperTemplateCoreExtensions.cs
```diff
@@ -473,23 +365,33 @@ public static class ImageCropperTemplateCoreExtensions
...
-                if (width == null) width = height;
-                options.WidthRatio = (decimal)width / (decimal)height;
+                // If only height specified then assume a square
+                if (width == null)
+                {
+                    options.Width = height;
+                }
+                else
+                {
+                    options.Width = (int)MathF.Round(height.Value * ((float)width.Value / height.Value));
```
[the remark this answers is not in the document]
ronaldbarendse: I've gone through this logic again and with the removal of width/height ratios, this indeed doesn't make sense anymore. The `ImageCropRatioMode` enum is now also removed and all `GetCropUrl` overloads are cleaned up 👍🏻

AndyButland (approved):
I believe I've verified now the changes in this PR, so approving.

ronaldbarendse (approved):

src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs
ronaldbarendse: Maybe it makes more sense to just use the `PointF` and `RectangleF` structs from [`System.Drawing.Primitives`](https://www.nuget.org/packages/System.Drawing.Primitives/)? Lets do that in a new PR though 😉

src/Umbraco.Infrastructure/Media/ImageSharpImageUrlGenerator.cs
JimBobSquarePants: Yep. They’re all loaded there by default

ronaldbarendse: This all worked fine whenever the ImageSharp.Web configuration used the same `Configuration.Default` instance (which always returns the same instance, because it's lazily initialized). If `ImageSharpMiddlewareOptions.Configuration` was set to a different instance, it would return the wrong image formats/file extensions.

I've updated the implementation to use the `IOptions` pattern, so it uses the same configuration as the middleware.

We'll probably have to do the same for `ImageDimensionExtractor`, as that uses `Image.Load(stream)` (without specifying the configuration and thus using the default).

@JimBobSquarePants Does ImageSharp support MS DI for fetching the default configuration? If so, we should probably register the configuration once and use that as default for both ImageSharp and ImageSharp.Web.

ronaldbarendse: I've now registered the default ImageSharp configuration and used that throughout Umbraco, so every implementation using ImageSharp can use that same instance (currently only `ImageDimensionExtractor` and the ImageSharp.Web middleware):
https://github.com/umbraco/Umbraco-CMS/blob/0b53249d5742404e7511d129ae55503ce8673c85/src/Umbraco.Infrastructure/DependencyInjection/UmbracoBuilder.CoreServices.cs#L182

As the supported image file types/extensions are fetched from this configuration, it's possible to only have PNG support using the following composer:
```c#
public class OnlyPngSupportComposer : IUserComposer
{
    public void Compose(IUmbracoBuilder builder) => builder.Services.AddUnique(new Configuration(new PngConfigurationModule()));
}
```

ronaldbarendse:
After moving some things around, the latest commits ensure we're always using the same ImageSharp configuration (currently in the `ImageDimensionExtractor` and `ImageSharpMiddlewareOptions`) by registering it into the DI container (see [comment](https://github.com/umbraco/Umbraco-CMS/pull/10623#discussion_r689463380)).

**Note to all implementers:** whenever you're using an ImageSharp API that allows injecting a `SixLabors.ImageSharp.Configuration` instance, make sure to resolve it from the [DI container](https://docs.microsoft.com/en-us/dotnet/core/extensions/dependency-injection).

bergmania (approved):

PR #10628, opened 2021-07-08
Fix for #10627 - UmbracoDictionaryTranslate checks.

### LINKED ISSUE #10627: v9-rc001 bug in UmbracoDictionaryTranslate method causes error in blocklist editor

v9 rc001

### Bug summary

A bug in the UmbracoDictionaryTranslate method causes an object exception when viewing the block editor. 

**if**
one of the labels or descriptions of a doctype being using in the block editor begins with the `#` character but does not contain and `_`

```
System.IndexOutOfRangeException, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e: Index was outside the bounds of the array.
Stacktrace
at Umbraco.Extensions.LocalizedTextServiceExtensions.UmbracoDictionaryTranslate(ILocalizedTextService manager, ICultureDictionary cultureDictionary, String text)
...
```

This is because there is a split method used in the UmbracoDictionaryTranslate function that does not then do a bounds check.

### Specifics

_No response_

### Steps to reproduce

- create a new block list editor datatype. 
- include a doctype as a block
- save
- chanse a property label on the block doctype to start with a '#' 
- save the block doctype
- view the blocklist datatype

see the error

should not error. 
should fail silently ? and display the value as is e.g ` #textstring`

There might be a migration issue, v8 to v9 people might label a control `#custom/pageTitle`. however in v8 this needs to be `#custom_pageTitle` - ideally this would update but at the very least i suppose it should not error.

KevinJump:
### Prerequisites

If there's an existing issue for this PR then this fixes #10627 

### Description
Adds some range checking for splitting the string for localization 

if a string doesn't contain a _ then the source text is returned (localization requires an area and key separated by _)

### DIFF
--- a/src/Umbraco.Core/Services/LocalizedTextServiceExtensions.cs
+++ b/src/Umbraco.Core/Services/LocalizedTextServiceExtensions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Umbraco.
+// Copyright (c) Umbraco.
 // See LICENSE for more details.
 
 using System;
@@ -76,7 +76,7 @@ namespace Umbraco.Extensions
              if (text == null)
                  return null;
 
-             if (text.StartsWith("#") == false)
+             if (text.StartsWith("#") == false || text.IndexOf('_') == -1)
                  return text;
 
              text = text.Substring(1);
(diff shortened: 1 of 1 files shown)


nikolajlauridsen:
Changes look good, and it tests out great 👍 thanks a bunch, just need the build server to happy and then I'll go ahead and merge 😄

PR #10636, opened 2021-07-09
Change path to log files in error.html (installation)

erikjanwestendorp:
### Description
Change path of the log folder since the App_Data folder doesn't exist anymore.


nikolajlauridsen (approved):
Good catch, thanks 😄 

Looks perfectly reasonable to me 👍

Issue #10554, opened 2021-06-29
v9-beta004: TypeFinder only searches assemblies that reference Umbraco


mattbrailsford:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

v9.0.0-beta004

### Bug summary

In Umbraco v8 it was possible to use the TypeFinder to locate types across any assembly, however it appears that in v9 the default implementation only searches assemblies that hold a reference to Umbraco core dll.

### Specifics

In Vendr we have a series of custom collection builder implementations that should be dynamically populated by looking for types in assemblies that are decorated with a specific attribute. This used to work in v8 where I we could define our collection builder and then register it with the composer using Umbraco's TypeFinder to auto look up these types (these types could be created by me or 3rd parties).

In v9 however, this doesn't appear to work the same. In Vendr we have some of these types defined in our core project Vendr.Core and some in another project Vendr.Umbraco. When loading my site and checking the collection it appears that only the ones defined in Vendr.Umbraco have been found. On deeper investigation I believe this is because the TypeFinder uses an `IAssemblyProvider` to know which assemblies to search though, which appears to have a default implementation via `DefaultUmbracoAssemblyProvider` which only searches assemblies that themselves have a reference to the Umbraco core dll, which Vendr.Core does not.

I can test this to an extent by explicitly passing the type finder a list of assemblies to search, and when I do that, it does indeed find those types. Unfortunately, as types may be implemented by third parties, I can't explicitly list the assemblies to search and so ideally I need the type finder to search all assemblies in the bin folder (which I believe is what v8 used to do?)

### Steps to reproduce

Define some types in an assembly that doesn't reference Umbraco. Then use the type finder to try and locate those types and non of them will be found.

### Expected result / actual result

I was expecting that the type finder would search all DLLs and not just those that reference Umbraco

mattbrailsford:
Just confirming that in v8 there were no restrictions on the dlls searched https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Core/Composing/TypeFinder.cs#L99

mattbrailsford:
The only way I've managed to get around this currently is to pass assemblies to the type finder. Given my comments earlier though that I don't know the assemblies that reference mine, I attempted to use the `FindAssembliesWithReferencesTo` from Umbraco, but this is internal so have had to do a dirty hack right now.

````csharp
[...]

PR #10646, opened 2021-07-09
Fixes runtime hash calculation

Shazwazza:
we are calculating the runtime hash incorrectly in v9. Have a look at the code in AddTypeFinder we do:

```cs
var runtimeHashPaths = new RuntimeHashPaths().AddFolder(new DirectoryInfo(Path.Combine(webHostEnvironment.ContentRootPath, "bin")));
```

There is no 'bin' folder so this will just be the same and we'll never type scan again. We need to re-think this for netcore based on the current environment. For example, if it's a published project, in which case everything exists in the root, else if it's a development project, in which case the binaries live in the /bin/....

So this fix just uses the Assembly provider to use returned assemblies for calculating the hash.

But wait... there's more! This also fixes the issue https://github.com/umbraco/Umbraco-CMS/issues/10554 where for custom apps that need to search for types based on a different entry assembly it was almost impossible (although I thought it was at the time). So this changes a few things:

* Changes some dependency chain issues with TypeFinder/Loader and others
* Allows for configuring additional entry point assemblies for the TypeFinder
* Exposes startup services on the IUmbracoBuilder if people absolutely require those services since it's impossible to reference them otherwise 
* Fixes issue with TypeLoader to use a cache key dependant on the passed in runtime hash value so that if additional TypeLoaders are created they can co-exist
* Allows for creating a ProfilingLogger with a generic logger so that it logs under a specific type instead of just "ProfilingLogger" which is mostly unhelpful.
* Some additional code to prevent re-allocating strings and collections.

### DIFF
--- a/src/Umbraco.Core/Composing/DefaultUmbracoAssemblyProvider.cs
+++ b/src/Umbraco.Core/Composing/DefaultUmbracoAssemblyProvider.cs
@@ -17,12 +17,17 @@ namespace Umbraco.Cms.Core.Composing
     {
         private readonly Assembly _entryPointAssembly;
         private readonly ILoggerFactory _loggerFactory;
+        private readonly IEnumerable<string> _additionalTargetAssemblies;
         private List<Assembly> _discovered;
 
-        public DefaultUmbracoAssemblyProvider(Assembly entryPointAssembly, ILoggerFactory loggerFactory)
+        public DefaultUmbracoAssemblyProvider(
+            Assembly entryPointAssembly,
+            ILoggerFactory loggerFactory,
+            IEnumerable<string> additionalTargetAssemblies = null)
         {
             _entryPointAssembly = entryPointAssembly ?? throw new ArgumentNullException(nameof(entryPointAssembly));
             _loggerFactory = loggerFactory;
+            _additionalTargetAssemblies = additionalTargetAssemblies;
         }
 
         // TODO: It would be worth investigating a netcore3 version of this which would use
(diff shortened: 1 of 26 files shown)


nikolajlauridsen:
Looks good to me, I haven't done any specific test steps though, just had a poke around, let me know if there's anything specific you want me to test.

I am getting two errors, nothing that prevents build though. One error is in `TypeLoaderBenchmarks.cs` where we try to poulate the cache:

```C#
cache.Insert(
    TypeLoader.CacheKey,
    GetCache,
    TimeSpan.FromDays(1));
```

We're trying to access `TypeLoader.CacheKey` as a static but it's not.

The other error is in `TestData`, however this project is targeting .net framework, and has a reference to `Umbraco.Web.csproj` so I'm wondering if it's just something we haven't gotten around to deleting yet?

Shazwazza:
@nikolajlauridsen I've pushed a fix for the benchmark project. The TestData project should be unloaded right now. It needs to be ported over to netcore. I'm unsure if we have a task for that yet. If you have any time it would be great to get a task made for that.

For testing, I've added some debug logging to the TypeFinder. You'll want to verify that re-scanning does not occur when there are no assembly changes but that it does re-scan when there are assembly changes.

nikolajlauridsen:
I had to completely delete the `TestData` project to be able to compile if I just unload it, I get a different error, which is a bit weird to me.

However, I can't seem to get the re-scanning to occur? I think I might be doing something wrong, I've tried to add some classes with the `IDiscoverable` interface, even a controller, to the umbraco projects, but it doesn't seem to want to rescan? Am I doing something horribly wrong?

Shazwazza:
@nikolajlauridsen how are you determining if it is rescanning? Just checking the logs or setting a breakpoint in TypeLoader? If it's logs - this is a DEBUG level log so you need change your logging level to output that. 

Any code changes you make to any of the Umbraco projects will result in a new scan. This is because when you build, the time stamp of the assembly will have changed which is taken into account when calculating the hash. You don't necessarily need to make any IDiscoverable changes, any code changes will do.

when using a breakpoint in TypeLoader `if (RequiresRescanning)` it always works for me on any code changes but returns false if there are no code changes.

Else, let me know how to replicate.

nikolajlauridsen (approved):
Yep sorry about that, that was me being stupid yesterday, Examine has an override on the development json, so I just assumed that debug logging was enabled because I saw debug entries in the log 🙈 

But it all checks out, and is good to go as far as I'm aware 👍 I'll hold off on merging this since I can see that Matt is doing some testing as well, but feel free to merge 😄

PR #10651, opened 2021-07-11
8.15RC Don't create new empty arrays

nzdev:
### Prerequisites

### Description
Fixes issue of creating new Empty arrays during NuCache deserialization.


nul800sebastiaan:
Thanks @nzdev - didn't see this for 8.15, so it will end up in 8.16-RC next week but I also cherry picked for 8.15.2 67e812d93cf1d13e7aa0dec85f3adf4e2e869cd6

PR #10653, opened 2021-07-12
Additional optional sanitization of scripting in TinyMCE

nielslyngsoe:
As of https://github.com/advisories/GHSA-w7jx-j77m-wp65 and https://github.com/advisories/GHSA-5vm8-hhgr-jcjp we need to ensure to prevent inserting inline arbitrary JavaScript execution in HTML of the TinyMCE editor.

This fix eliminates the option to insert inline 'on' attributes.

This fix also eliminates using javascript in URIs of various element types. Including checking the validity of the URIs. If not a valid URI then it will be removed.

This also means that inline data in src tag is being limited to only be possible for data of type image or svg.

The additional sanitization is disabled for sites that are upgrading from an older version. Clean installs of Umbraco will have the option enabled by default. Enabling the option will not automatically remove any stored scripting, each TinyMCE instance on each content item will need to be saved manually.

Enabling the additional sanitization can be done in the `web.config` by adding the following `appSetting`: 

```xml
<add key="Umbraco.Web.SanitizeTinyMce" value="true" />
```

This works is based on the fixes done for TinyMCE v5:

- https://github.com/tinymce/tinymce/pull/6179
- https://github.com/tinymce/tinymce/pull/6563

We determined which attributes can contain possibly problematic scripting.

Enabling this configuration will do the following:

- Remove any attributes in HTML starting with `on` (`onClick`, `onMouseOver`, etc).
- Remove any scripting (javascript, vbscript, mhtml) from the following attributes on any html tag
  - `src`
  - `href`
  - `data`
  - `background`
  - `action`
  - `formaction`
  - `poster`
  - `xlink:href`
- Preserve embedded svgs in the `data:image` attribute on `img` and `video` tags, scripts in those tags never get executed by browsers

---
_This item has been added to our backlog [AB#13289](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13289)_


nikolajlauridsen:
I've updated this so the patch is only applied if the following is present in the appsettings section of web.config: 

```
<add key="Umbraco.Web.MCE.SecurityPatches" value="true" />
```

If the key is missing or set the false the patch will not be applied, making it opt-in, I've also updated the `web.template.config` to have this key, so future projects will automatically have this key, making it opt-out for future projects. 

Ideally, I would like to have placed this in either the security section of `umbracoSettings.config` or in the `tinyMce.config` however as far as I can tell these configs are not generated from a template, meaning that if I change it there it would break existing sites, which is no bueno, however, this would not be a problem in V9 since all settings are loaded from `appsettings.json` which is generated from a template.

nul800sebastiaan:
Cherry picked for 8.17.1 in 23ce69db2fe3b0677a0d8520aa48efce3b1773a7

PR #10654, opened 2021-07-12
Refactor NPoco queries to use constants and dtos

bjarnef:
### Prerequisites

### Description
I noticed some SQL querying which combined some strongly typed qyerying using NPoco and static text. While we still can construct sql as strings many places we can use the constants or dto classes to avoid making typos.

There `WhereIn<T>` method makes it very easy to construct sql `IN` queries with less code.


nul800sebastiaan:
Looking good @bjarnef! Thanks for the help! 👍

PR #10655, opened 2021-07-12
Remove raiseEvent method parameters and allow suppressing events at the scope level

Shazwazza:
For years we've wanted to remove the raiseEvent service parameters from various services. Why?

* Because it is entirely inconsistent, not all services have this as method parameters and maintaining that consistency is impossible especially if 3rd party libraries support events/notifications.
* It's hacky. There's no good way to suppress events/notifications this way at a higher (scoped) level. 
* There's also hard coded logic to ignore these parameters sometimes which makes it even more inconsistent. 
* There's events below services at the repository level that cannot be controlled by this flag.

Instead, we can suppress notifications at the scope level which makes things consistent, easy and will work for all services that use a Scope and there's no required maintenance to make sure that new service methods will also work.

How to do it? It's very easy:

* Create an explicit scope and then call scope.Notifications.Supress(). 
* The result of Suppress() is IDisposable, so until that is disposed, notifications will not be added to the queue. 

Example:

```cs
using (IScope scope = ScopeProvider.CreateScope(autoComplete: true))
using (IDisposable _ = scope.Notifications.Supress())
{
    // TODO: Calls to service methods here will not have notifications
}
```

Child scope will inherit the parent Scope's notification object which means if a parent scope has notifications supressed, then so does the child scope. You cannot call Supress() more than once for the same outer scope instance else an exception will be thrown. This is to ensure that you cannot un-suppress notifications at a child level for an outer scope and ensures that suppressing events is a very explicit thing to do.

Why would you want to suppress events? The main reason for ever doing this would be performance for bulk operations. The caller will need to be aware that suppressing events will lead to a totally inconsistent content cache state (if notifications are suppressed for content or media services). This is because notifications are used by things such as nucache to populate the cmsContentNu table and to populate the content caches. They are also used to populate the Examine indexes. 

So if you did suppress events, it will require you to rebuild the nucache and examine data manually.

Questions - 

We had hard coded checks previously for raiseEvents where we would ignore this flag. These checks are now gone. We also have notifications at the repository level for a few things to keep nucache data in sync. With this change, these notifications can be suppressed whereas before they couldn't. This leads me to believe that some things are not setup correctly and it's all based on nucache. The repo notifications (which are marked as obsolete in the codebase since they should only be used internally) are used to keep the cmsContentNu table in sync and during the same transaction. IMO I think if we 100% rely on this data (which we do) then this repository logic should be build directly into the repository (i.e. the document repository just calls into the NuCacheContentRepository?). In the meantime, I've added some hacks back into place because it's quite a pain if the cmsContentNu data is not consistent with the DB. So there's a `[CannotSuppressNotification]` attribute. Would like to know your thoughts on this.

### DIFF
--- a/src/Umbraco.Core/Events/IScopedNotificationPublisher.cs
+++ b/src/Umbraco.Core/Events/IScopedNotificationPublisher.cs
@@ -13,7 +13,7 @@ namespace Umbraco.Cms.Core.Events
         /// Supresses all notifications from being added/created until the result object is disposed.
         /// </summary>
         /// <returns></returns>
-        IDisposable Supress();
+        IDisposable Suppress();
 
         /// <summary>
         /// Publishes a cancelable notification to the notification subscribers
--- a/src/Umbraco.Core/Events/IScopedNotificationPublisher.cs
+++ b/src/Umbraco.Core/Events/IScopedNotificationPublisher.cs
@@ -10,7 +10,7 @@ namespace Umbraco.Cms.Core.Events
     public interface IScopedNotificationPublisher
     {
         /// <summary>
-        /// Supresses all notifications from being added/created until the result object is disposed.
+        /// Suppresses all notifications from being added/created until the result object is disposed.
         /// </summary>
         /// <returns></returns>
         IDisposable Suppress();
--- a/src/Umbraco.Core/Events/IScopedNotificationPublisher.cs
+++ b/src/Umbraco.Core/Events/IScopedNotificationPublisher.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Umbraco.
 // See LICENSE for more details.
 
+using System;
 using System.Threading.Tasks;
 using Umbraco.Cms.Core.Notifications;
 
@@ -8,6 +9,12 @@ namespace Umbraco.Cms.Core.Events
 {
     public interface IScopedNotificationPublisher
     {
+        /// <summary>
+        /// Supresses all notifications from being added/created until the result object is disposed.
+        /// </summary>
+        /// <returns></returns>
+        IDisposable Supress();
+
         /// <summary>
         /// Publishes a cancelable notification to the notification subscribers
         /// </summary>
--- a/src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
+++ b/src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
@@ -1030,7 +1030,7 @@ namespace Umbraco.Cms.Infrastructure.Packaging
 
             if (dataTypes.Count > 0)
             {
-                _dataTypeService.Save(dataTypes, userId, true);
+                _dataTypeService.Save(dataTypes, userId);
             }
 
             return dataTypes;
--- a/src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
+++ b/src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
@@ -1129,7 +1129,7 @@ namespace Umbraco.Cms.Core.Services.Implement
                 }
 
                 scope.Notifications.Publish(new MediaSavedNotification(itemsA, messages).WithStateFrom(savingNotification));
-                // TODO: See note about supressing events in content service
+                // TODO: See note about suppressing events in content service
                 scope.Notifications.Publish(new MediaTreeChangeNotification(saved, TreeChangeTypes.RefreshNode, messages));
                 Audit(AuditType.Sort, userId, 0);
 
--- a/src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
+++ b/src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
@@ -685,7 +685,7 @@ namespace Umbraco.Cms.Core.Services.Implement
 
                 _mediaRepository.Save(media);
                 scope.Notifications.Publish(new MediaSavedNotification(media, eventMessages).WithStateFrom(savingNotification));
-                // TODO: See note about supressing events in content service
+                // TODO: See note about suppressing events in content service
                 scope.Notifications.Publish(new MediaTreeChangeNotification(media, TreeChangeTypes.RefreshNode, eventMessages));
 
                 Audit(AuditType.Save, userId, media.Id);
--- a/src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
+++ b/src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
@@ -728,7 +728,7 @@ namespace Umbraco.Cms.Core.Services.Implement
                 }
 
                 scope.Notifications.Publish(new MediaSavedNotification(mediasA, messages).WithStateFrom(savingNotification));
-                // TODO: See note about supressing events in content service
+                // TODO: See note about suppressing events in content service
                 scope.Notifications.Publish(new MediaTreeChangeNotification(treeChanges, messages));
                 Audit(AuditType.Save, userId == -1 ? 0 : userId, Cms.Core.Constants.System.Root, "Bulk save media");
 
--- a/src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
+++ b/src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
@@ -654,14 +654,14 @@ namespace Umbraco.Cms.Core.Services.Implement
         /// <param name="media">The <see cref="IMedia"/> to save</param>
         /// <param name="userId">Id of the User saving the Media</param>
         /// <param name="raiseEvents">Optional boolean indicating whether or not to raise events.</param>
-        public Attempt<OperationResult> Save(IMedia media, int userId = Cms.Core.Constants.Security.SuperUserId, bool raiseEvents = true)
+        public Attempt<OperationResult> Save(IMedia media, int userId = Cms.Core.Constants.Security.SuperUserId)
         {
             EventMessages eventMessages = EventMessagesFactory.Get();
 
             using (IScope scope = ScopeProvider.CreateScope())
             {
                 var savingNotification = new MediaSavingNotification(media, eventMessages);
-                if (raiseEvents && scope.Notifications.PublishCancelable(savingNotification))
+                if (scope.Notifications.PublishCancelable(savingNotification))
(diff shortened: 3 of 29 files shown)


src/Umbraco.Core/Events/IScopedNotificationPublisher.cs
```diff
@@ -1,13 +1,20 @@
...
+using System;
 using System.Threading.Tasks;
 using Umbraco.Cms.Core.Notifications;
 
 namespace Umbraco.Cms.Core.Events
 {
     public interface IScopedNotificationPublisher
     {
+        /// <summary>
+        /// Supresses all notifications from being added/created until the result object is disposed.
```
AndyButland, line 13:
```suggestion
        /// Suppresses all notifications from being added/created until the result object is disposed.
```

src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
```diff
@@ -685,10 +684,8 @@ public Attempt<OperationResult> Save(IMedia media, int userId = Cms.Core.Constan
                 }
 
                 _mediaRepository.Save(media);
-                if (raiseEvents)
-                {
-                    scope.Notifications.Publish(new MediaSavedNotification(media, eventMessages).WithStateFrom(savingNotification));
-                }
+                scope.Notifications.Publish(new MediaSavedNotification(media, eventMessages).WithStateFrom(savingNotification));
+                // TODO: See note about supressing events in content service
```
AndyButland, line 688:
```suggestion
                // TODO: See note about suppressing events in content service
```

src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
```diff
@@ -731,10 +727,8 @@ public Attempt<OperationResult> Save(IEnumerable<IMedia> medias, int userId = Cm
                     _mediaRepository.Save(media);
                 }
 
-                if (raiseEvents)
-                {
-                    scope.Notifications.Publish(new MediaSavedNotification(mediasA, messages).WithStateFrom(savingNotification));
-                }
+                scope.Notifications.Publish(new MediaSavedNotification(mediasA, messages).WithStateFrom(savingNotification));
+                // TODO: See note about supressing events in content service
```
AndyButland, line 731:
```suggestion
                // TODO: See note about suppressing events in content service
```

src/Umbraco.Infrastructure/Services/Implement/MediaService.cs
```diff
@@ -1135,10 +1128,8 @@ public bool Sort(IEnumerable<IMedia> items, int userId = Cms.Core.Constants.Secu
                     _mediaRepository.Save(media);
                 }
 
-                if (raiseEvents)
-                {
-                    scope.Notifications.Publish(new MediaSavedNotification(itemsA, messages).WithStateFrom(savingNotification));
-                }
+                scope.Notifications.Publish(new MediaSavedNotification(itemsA, messages).WithStateFrom(savingNotification));
+                // TODO: See note about supressing events in content service
```
AndyButland, line 1132:
```suggestion
                // TODO: See note about suppressing events in content service
```

src/Umbraco.Tests.Integration/Umbraco.Infrastructure/Scoping/SupressNotificationsTests.cs
```diff
@@ -0,0 +1,117 @@
...
+using Umbraco.Cms.Core.Services;
+using Umbraco.Cms.Tests.Common.Builders;
+using Umbraco.Cms.Tests.Common.Testing;
+using Umbraco.Cms.Tests.Integration.Testing;
+
+namespace Umbraco.Cms.Tests.Integration.Umbraco.Infrastructure.Scoping
+{
+    [TestFixture]
+    [UmbracoTest(Database = UmbracoTestOptions.Database.NewSchemaPerTest)]
+    public class SupressNotificationsTests : UmbracoIntegrationTest
```
AndyButland, line 20:
```suggestion
    public class SuppressNotificationsTests : UmbracoIntegrationTest
```

You'll also need to update the file name.

src/Umbraco.Tests.Integration/Umbraco.Infrastructure/Scoping/SupressNotificationsTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                asserted = TestContext.CurrentContext.AssertCount;
+                Media media = MediaBuilder.CreateMediaImage(mediaType, -1);
+                MediaService.Save(media);
+            }   
+
+            Assert.AreEqual(asserted + 1, TestContext.CurrentContext.AssertCount);
+        }
+
+        [Test]
+        public void GivenSuppressedNotificationsOnParent_WhenChildSupresses_ThenExceptionIsThrown()
```
AndyButland, line 87:
```suggestion
        public void GivenSuppressedNotificationsOnParent_WhenChildSuppresses_ThenExceptionIsThrown()
```

AndyButland (commented):
This all looks good to me - just suggested fixes for a mis-spelling.

The only thing I wondered from an API perspective was whether we could also support this?
		
```
using (IScope scope = ScopeProvider.CreateScope(autoComplete: true, suppressNotifications: true))
```

I.e. if did want to suppress notifications, you may well want to do it for the whole scope, so being able to do so in one line is a little more succinct.  But still keep what you have if you want more control.

For Deploy I think we'll need to keep the existing implementation as there what we're not using these flags, and really what we're doing is not suppressing all events, but rather gathering them up and publishing only some of them at the end.  I'll ping @nikolajlauridsen for his thoughts though as he spent a fair bit of time thinking about and working on this a month or two ago.

src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
```diff
@@ -1030,7 +1030,7 @@ public IReadOnlyList<IDataType> ImportDataTypes(IReadOnlyCollection<XElement> da
 
             if (dataTypes.Count > 0)
             {
-                _dataTypeService.Save(dataTypes, userId, true);
+                _dataTypeService.Save(dataTypes, userId);
```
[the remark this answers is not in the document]
AndyButland: I think this is OK.  The `true` being passed is for the `raiseEvents` parameter, not to suppress notifications, so it's the other way around.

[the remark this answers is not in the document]
nikolajlauridsen: Ahh jeez, yes of course, that's my bad, better go make a second cup of coffee 😄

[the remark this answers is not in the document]
AndyButland: Actually a good example of why this PR is a good idea 😄.

nikolajlauridsen:
Looks very nice to me, this cleans everything up quite a bit! 

In regards to Deploy, we would indeed need to keep our existing implementation of the `DeployScopedNotificationPublisher` because as Andy mentioned we're not just suppressing all events, but just only publishing them to the `DistributedCacheBinder` but more relevant the `SignatureRefresher` which is a deploy specific thing that needs the notifications to keep the signatures up to date. 

> IMO I think if we 100% rely on this data (which we do) then this repository logic should be build directly into the repository (i.e. the document repository just calls into the NuCacheContentRepository?).

I think this would be a better solution as well, the way it's set up with notifications now is quite messy, because they depend on the notification being sent at a very specific place, as I recall, that's only really useful for Nucache, so it seems like an odd fit to me. This also leads me to a second thought, that we might need to apply the `CannotSuppressNotification` attribute to a few more attributes:

* ContentTypeRefreshedNotification
* MediaTypeRefreshedNotification
* MemberTypeRefreshedNotification

These are all also used by the `PublishedSnapshotServiceEventHandler`. there's also `LanguageSavedNotification` and `MemberDeletingNotification`, wouldn't it cause issues with the cache if these a suppressed as well?

Shazwazza:
@nikolajlauridsen 

> These are all also used by the PublishedSnapshotServiceEventHandler. there's also LanguageSavedNotification and MemberDeletingNotification, wouldn't it cause issues with the cache if these a suppressed as well?

Yes they will. These will basically cause the whole nucache db table to be rebuilt. The reason I left the other ones in there is because these commonly occur.

The primary (only?) reason you would ever suppress events is for bulk DB performance. Which means that you need to know before hand what you are about to do and that in this case you will absolutely need to rebuild the nucache DB table and then the nucache cache.

IMO I think this is the better and more consistent option with less hacks (we should remove the [CannotSuppressNotification]). When we suppress events, we do it 100%. It will mean that developers will need to manually (or programmatically) rebuild the nucache DB table and the nucache cache. 

@AndyButland 

> I.e. if did want to suppress notifications, you may well want to do it for the whole scope, so being able to do so in one line is a little more succinct. But still keep what you have if you want more control.

Although that seems like a nicer syntax, it is also less explicit. If people are going to suppress events they need to know what they are doing (see above) and we shouldn't make it very easy to do. IMO this should be a very explicit operation.


thoughts?

AndyButland:
> Although that seems like a nicer syntax, it is also less explicit. If people are going to suppress events they need to know what they are doing (see above) and we shouldn't make it very easy to do. IMO this should be a very explicit operation.

Fine with me, and agree it makes sense to not make it to seem like a standard thing to do.

nikolajlauridsen:
> The primary (only?) reason you would ever suppress events is for bulk DB performance. Which means that you need to know before hand what you are about to do and that in this case you will absolutely need to rebuild the nucache DB table and then the nucache cache.

Hmm yes, I see what you mean, with this in mind, I think it might be better to just completely remove the `CannotSuppressNotificationAttribute` then? This way it's consistent, you know you *have* to rebuild the cache if you've suppressed notifications, otherwise, there will be some actions that you can perform without having to rebuild the cache, and others that you can't, which seems kind of confusing? How are you supposed to know what you can and can't do with suppressed notifications without rebuilding the cache?

Shazwazza:
Cool yep I'll remove the `CannotSuppressNotificationAttribute` hack :) Consistency is key I think .

Shazwazza:
I've pushed that change, if you are happy with it please feel free to merge.

PR #10657, opened 2021-07-12
Changes all collections from collection builders to resolve the concrete instances lazily.

Shazwazza:
This means we don't have to inject Lazy<T> all over the place when dealing with
collection builders and circular references since this will automatically just work OOTB.
This in theory should also allocate less instances during startup.

## Testing

Ensure all of the back office and front-end works. Test upgrading from v8 to ensure that those collections all resolve correctly. It will be pretty clear when something doesn't work because it will just explode.

### DIFF
--- a/src/Umbraco.Core/Actions/ActionCollection.cs
+++ b/src/Umbraco.Core/Actions/ActionCollection.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+using System;
+using System.Collections.Generic;
 using System.Linq;
 using Umbraco.Cms.Core.Composing;
 using Umbraco.Cms.Core.Models.Membership;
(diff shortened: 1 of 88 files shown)


nikolajlauridsen:
Changes looks good, but `ScopeBuilderRespectsContainerScope` fails because the service collection has been disposed, which kind of makes sense since the second `TestCollection` doesn't get resolved till after the `IServiceScope` is disposed. If I force the second collection to resolve within the scope with `var _ = col2.Count;` the test passes with no issues.

I haven't tested it out yet, but I'll do that first thing tomorrow morning 👍

Shazwazza:
I've pushed a fix.

nikolajlauridsen (approved):
Did some various testing, and was unable to make anything grenade, so it looks all good to me 👍 

In regards to upgrading from V8, I created a V8 site with a SQL Server database, create some content, and then pointed a V9 install to that database, and the migrations ran through fine, I assume that's what you mean by upgrading from V8, if so feel free to merge this 😄

Shazwazza:
Great! thanks :)

PR #10659, opened 2021-07-12
Fixes issue that the update culture dates were not synced with the version date like they are for the published culture dates

Shazwazza:
The issue isn't so much about dirty tracking since that is done, it was more about when the calls to `entity.SetCultureEdited(editedCultures)` were done in the DocumentRepository since these were set too late and therefore no culture changes were known. The 2nd part was that the `AdjustDates` wasn't called for saving, only for publishing.


nikolajlauridsen:
Hmm changes looks fine to me, but seems like `Can_SaveRead_Variations` fails now

Shazwazza:
@nikolajlauridsen Have pushed a fix :) The problem was when we were calling SetCultureEdited and AdjustDates within the repository.

PR #10669, opened 2021-07-13
V9: Fix recursive localize when resetting password

### LINKED ISSUE #10668: Unable to Send Forgotten Password Email

9.0.0-RC

### Bug summary

I get a StackOverflow exception and application shuts down using Kestrel server, when I try to send out a forgotten password email from the login screen.

### Specifics

_No response_

### Steps to reproduce

* Add SMTP configuration settings to appsettings.json
* Use a local SMTP server app such as PaperCut or similar
* Start application
* Goto login to backoffice
* Ensure forgotten password link is visible
* Try to send yourself a forgotten password email

**Expected Result**
An email is sent out to the specified user email with a forgotten password link

**Actual Result**
A stack overflow message in the console from logs and no email is sent out with forgotten password

nikolajlauridsen:
This fixes #10668.

The problem was that during a merge the `Localize` extension on `ILocalizedTextService` which takes a string array of tokens was accidentally made to call itself, resulting in an endless recursion, which meant we'd get a stack overflow.

This PR calls `ConvertToDictionaryVars` before calling `Manager.Localize`, so no more recursion.

One thing I noticed while here is that we always use `Thread.CurrentThread.CurrentUICulture` even though we take `CultureInfo` as a parameter, wouldn't it make more sense to use that if available? And use `Thread.CurrentThread.CurrentUICulture` as a fallback?

See issue for steps to test

### DIFF
--- a/src/Umbraco.Core/Services/LocalizedTextServiceExtensions.cs
+++ b/src/Umbraco.Core/Services/LocalizedTextServiceExtensions.cs
@@ -55,7 +55,7 @@ namespace Umbraco.Extensions
         /// <param name="tokens"></param>
         /// <returns></returns>
         public static string Localize(this ILocalizedTextService manager, string area, string alias, CultureInfo culture, string[] tokens)
-            => manager.Localize(area, alias, Thread.CurrentThread.CurrentUICulture, tokens);
+            => manager.Localize(area, alias, Thread.CurrentThread.CurrentUICulture, ConvertToDictionaryVars(tokens));
 
          /// <summary>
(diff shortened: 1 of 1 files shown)


warrenbuckley:
Code change looks good to me 👀 
Off to test & verify this is all good now 😄

warrenbuckley:
Can confirm this fixes the problem with emails being sent out for Forgotten Password & Notifications.
Thanks for being SONIC fast !!

Will wait for CI to go green & will merge this in

Issue #10526, opened 2021-06-24
Examine overview


patrickdemooij9:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

8.13 and up

### Bug summary

I am not quite sure when this started, but when you are going to the examine management section and search for something, it'll return the name as an array instead of an actual name. It also doesn't have the link to the content item anymore.

Old version:
![image](https://user-images.githubusercontent.com/11466511/123269518-5b649700-d4ff-11eb-809c-b1a1b9bf5c14.png)

8.13+
![image](https://user-images.githubusercontent.com/11466511/123269575-661f2c00-d4ff-11eb-85cb-ad61a50423bf.png)

8.13 also doesn't have the link on the name anymore that opens an infinite editor of the content item

### Steps to reproduce

- Start with a clean Umbraco install
- Go to examine management and search for something
- See that all names are returned as arrays

### Expected result / actual result

Logic should be the same as before:

- Name showing as string, not as array
- Link to content item

nul800sebastiaan:
Hey @patrickdemooij9 that is weird, it must be related to the change here: https://github.com/umbraco/Umbraco-CMS/pull/9864

I can't figure out exactly what has changed the search results though. If you have a bit of time to look into it that would be awesome! 🙏

umbrabot:
Hi @patrickdemooij9,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

patrickdemooij9:
Seems like it was missing some logic on the result overview. Fixed it in this PR: https://github.com/umbraco/Umbraco-CMS/pull/10672

patrickdemooij9:
@nul800sebastiaan, now that the period of v8 pull requests is over, does this mean I'll also have to make a PR targeting the v9 branch?

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/10672

@patrickdemooij9 We still merge all of v8 up to v9, no need to also make a PR for 9!

PR #10672, opened 2021-07-13
Fix examine overview not working as expected

patrickdemooij9:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/10526

With one of the recent Examine changes, the values are now returned as an array. This new response was already working well for the overview of values but was missing from the overview of the results.

I've added the same logic as in the overview of the values and fixed the logic with the editUrl.

ExamineOverviewFi

Can be tested by going to the Examine Overview and searching.


nul800sebastiaan:
Wonderful, that looks better! Thanks @patrickdemooij9 👍

Issue #10383, opened 2021-06-03
Error when trying to update a culture variant property to invariant


jreurink:
### Which Umbraco version are you using?

8.8.2

### Bug summary

We have an Umbraco on-premise installation with a large amount (thousands) of culture variant nodes. We are trying to update some culture variant properties to invariant, but we get a server error during saving. The error is as follows: 
System.ArgumentException: An item with the same key has already been added. 

Stack trace:
[stacktrace.txt](https://github.com/umbraco/Umbraco-CMS/files/6591681/stacktrace.txt)

### Specifics

We've found the issue after some digging around in the source of Umbraco.Core, in the '[RenormalizeDocumentEditedFlags](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs#L1099)' method of the [ContentTypeRepositoryBase](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs), in this statement:

```
var docCultureVariationsToUpdate = editedLanguageVersions.InGroupsOf(2000)
.SelectMany(_ => Database.Fetch<DocumentCultureVariationDto>(
Sql().Select<DocumentCultureVariationDto>().From<DocumentCultureVariationDto>()
.WhereIn<DocumentCultureVariationDto>(x => x.LanguageId, editedLanguageVersions.Keys.Select(x => x.langId).ToList())
.WhereIn<DocumentCultureVariationDto>(x => x.NodeId, editedLanguageVersions.Keys.Select(x => x.nodeId))))
//convert to dictionary with the same key type
.ToDictionary(x => (x.NodeId, (int?)x.LanguageId), x => x);
```

The collection of editedLanguageVersions contains more than 2000 items, which causes the subsequent Database.Fetch call to duplicate its results, which causes the .ToDictionary call to fail.

### Steps to reproduce

The issue only occurs when there are more than 2000 items in the collection of editedLanguageVersions, which requires a lot of content nodes, at least a thousand for a site with two languages. An easy way to reproduce it is by altering the source of Umbraco.Core and copying the updated dll in your project. By changing the line 
[...]

PR #10677, opened 2021-07-14
Add DistinctBy to avoid duplications 

erikjanwestendorp:
Add DistinctBy to avoid duplications

### Prerequisites

See issue description: 
https://github.com/umbraco/Umbraco-CMS/issues/10383 

### Description
Add DistinctBy to avoid duplications

### DIFF
--- a/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs
@@ -1101,6 +1101,7 @@ AND umbracoNode.id <> @id",
                     Sql().Select<DocumentCultureVariationDto>().From<DocumentCultureVariationDto>()
                             .WhereIn<DocumentCultureVariationDto>(x => x.LanguageId, editedLanguageVersions.Keys.Select(x => x.langId).ToList())
                             .WhereIn<DocumentCultureVariationDto>(x => x.NodeId, editedLanguageVersions.Keys.Select(x => x.nodeId))))
+                .DistinctBy(x => (x.NodeId, (int?)x.LanguageId))
                 //convert to dictionary with the same key type
                 .ToDictionary(x => (x.NodeId, (int?)x.LanguageId), x => x);
 


ronaldbarendse:
@erikjanwestendorp Thanks for creating a PR to fix the linked issue 👍🏻 Although this does indeed remove the duplications, it doesn't fix the underlying issue when more than 2000 items are in the collection.

https://github.com/umbraco/Umbraco-CMS/blob/b8540ca128bf1622ff26eb21d331577776d5a725/src/Umbraco.Core/Persistence/Repositories/Implement/ContentTypeRepositoryBase.cs#L1099-L1105

This issue is twofold:
- Splitting the `editedLanguageVersions` into groups of 2000 items, but then not using those groups, but all items in the SQL query: this could result in multiple queries (for every group of 2000 items) doing the exact same query, causing the duplicate items
- The grouping is added to avoid hitting the maximum SQL parameter count of 2100, but for every item 2 parameters are added (for both the language and node ID): this could result in 4000 parameters.

I expect the `WhereIn` removes duplicate values, so that would explain why the maximum SQL parameter count isn't always reached... So this fix should also ensure the amount of items per group is lowered to 1000 and the actual items in that group are used (the `SelectMany(_ => ...)` currently discards the grouped items).

ronaldbarendse:
But thinking about this a little bit more, grouping the items and doing multiple SQL queries could result in missing rows. To illustrate:
- This would return a row with `languageId = NULL` and `nodeId = 3000`:
  ```sql
  WHERE languageId IN (null, 1, 2) AND WHERE nodeId IN (1000, 2000, 3000)
   ```
- But splitting this into groups of 2 won't:
  ```sql
  WHERE languageId IN (null, 1) AND WHERE nodeId IN (1000, 2000)
  WHERE languageId IN (2) AND WHERE nodeId IN (3000)
  ```

So we should ensure one of the `WHERE IN` statements is always completely added, which should be the least amount of items (typically we have less languages than nodes) and then try to add the remaining maximum amount of parameters for the other `WHERE IN` statement. So the above example would end up as:
```sql
WHERE languageId IN (null, 1, 2) AND WHERE nodeId IN (1000, 2000)
WHERE languageId IN (null, 1, 2) AND WHERE nodeId IN (3000)
```

erikjanwestendorp:
@ronaldbarendse Thanks for your review and your feedback. Just updated the PR, would you like to look at this again? 😄. 

- [x] Work with groups now
- [x] Items per group lowered to 1000

ronaldbarendse:
Hi @erikjanwestendorp, I just got around to looking at this again and unfortunately this still has the issue I described in https://github.com/umbraco/Umbraco-CMS/pull/10677#issuecomment-879853655 (we can't group both language and node IDs at the same time, as that might result in missing results).

I've created PR https://github.com/umbraco/Umbraco-CMS/pull/11369 to introduce a constant for the 'magic' number we use in quite a few places to prevent going over the maximum amount of SQL parameters. It also fixes this and a similar issue when fetching user group permissions, so I'll close this one in favor of that PR. Thanks for the effort you've put into this though! 🤗

Issue #10667, opened 2021-07-13
Changing BackOffice User/s passwords does not show validation messages


warrenbuckley:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0-RC

### Bug summary

When changing password for the user in the backoffice no validation UI is being wired upto the backoffice.

### Specifics

I tested this in Umbraco 8.14.1 so something has changed or been merged up wrong potentially in V9 for this to have stopped working.

## V8.14.1
![image](https://user-images.githubusercontent.com/1389894/125452118-c10bcc91-943a-4c60-a772-7ebad58006e1.png)

## V9.0.0RC
![image](https://user-images.githubusercontent.com/1389894/125452076-e8ee707f-0727-48c4-bb87-2b7f44edd970.png)

## Note
See the difference between the two versions network responses. You can see that the response object has changed and no longer contains ModelState which the UI.Client expects for wiring up errors to server-val-field and other error handling in the UI/backoffice

### Steps to reproduce

* Have V9.0.0RC installed
* Login to backoffice
* Click avatar top right
* Click change password
* Attempt to change password - but ensure that the OLD password is incorrect

### Expected result / actual result

**Expected Result**
Expect red box around old password input field and message to be shown to 

**Actual Result**
No validation shown to user for invalid old password

nul800sebastiaan:
Same as https://github.com/umbraco/Umbraco-CMS/issues/10667 I guess.

warrenbuckley:
## Other locations
* **Changing password in the Users section**
This has the same problem and the dialog/overlay closes instantly so you assume have successfully changed the password with no visual feedback.

nul800sebastiaan:
Same as https://github.com/umbraco/Umbraco-CMS/issues/10658 I guess

warrenbuckley:
@nul800sebastiaan feel free to close this issue then if needed if its a dupe

Shazwazza:
@warrenbuckley  + @nikolajlauridsen 

Related to this https://github.com/umbraco/Umbraco-CMS/pull/10534 where we streamline the response usages in all controllers to be consistent with out .net is supposed to work.

The problem is this (for example):

[...]

PR #10680, opened 2021-07-14
V9: Fix validation messages

### LINKED ISSUE #10658: v9: validation not showing

v9-rc001

### Bug summary

Validation errors are not being displayed. 

It seems to be missing in multiple places.

### Specifics

_No response_

### Steps to reproduce

install umbraco v9
add a user
add a user with the same email

This result more or less nothing:
image

On v8 the steps described result in :
image

The output from the server is similar to this:

)]}',
{"Email":{"ChildNodes":null,"Children":null,"Key":"Email","SubKey":{"Buffer":"Email","Offset":0,"Length":5,"Value":"Email","HasValue":true},"IsContainerNode":false,"RawValue":null,"AttemptedValue":null,"Errors":[{"Exception":null,"ErrorMessage":"A user with the email already exists"}],"ValidationState":"Invalid"}}

It seems to be missing the "modelState", that the formHelper.handleError requires.

nikolajlauridsen:
This fixes #10658 and #10667.

As Shannon explains in #10667 this was due to ambiguous methods in `UmbracoAuthorizedApiController`, to fix this I removed the default value from the `statusCode` parameter.

After removing the default I noticed that `ValidationProblem` is used in a lot of places with some sort of `INotificationModel`, so I added another method that explicitly takes in an `INotificationModel` which still have the default status code, just to make everything a bit easier to use.

### DIFF
--- a/src/Umbraco.Web.BackOffice/Controllers/UmbracoAuthorizedApiController.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/UmbracoAuthorizedApiController.cs
@@ -120,7 +120,7 @@ namespace Umbraco.Cms.Web.BackOffice.Controllers
             => new ValidationErrorResult(value, statusCode);
 
         /// <summary>
-        /// Returns an Umbraco compatible validation problem for the object result
+        /// Returns an Umbraco compatible validation problem for the given notification model
         /// </summary>
         /// <param name="model"></param>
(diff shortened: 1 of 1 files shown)


warrenbuckley:
Code change looks good to me, will go off and test this out :)

warrenbuckley:
I can confirm this fixes #10658 & #10667

Issue #10679, opened 2021-07-14
UserService.SavedUserGroup event returns wrong list of AddedUsers when removing user


inetzo:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

8.14.1 and below

### Bug summary

When a user is removed from the Usergroup and the UserService.SavedUserGroup event is fired, the property AddedUsers returns a list of all backoffice users. When a user is added to the group, it only contains the added users.

![afbeelding](https://user-images.githubusercontent.com/18665536/125621177-972aa69c-da4d-47b4-bf19-69bd00a0ec54.png)

![afbeelding](https://user-images.githubusercontent.com/18665536/125621210-8a93b7af-6c33-4ef5-87f5-1efaab8c2d67.png)

![afbeelding](https://user-images.githubusercontent.com/18665536/125621234-143d4087-2529-4833-8104-10620625c9cb.png)

### Steps to reproduce

Add an component to catch the UserService.SavedUserGroup event:

```
    public class UserGroupComponent: IComponent
    {

        public UserGroupComponent()
        {

        }

        public void Initialize()
        {
            UserService.SavedUserGroup += UserService_SavedUserGroup;
        }

        private void UserService_SavedUserGroup(Umbraco.Core.Services.IUserService sender, Umbraco.Core.Events.SaveEventArgs<Umbraco.Core.Events.UserGroupWithUsers> e)
        {

            foreach (var group in e.SavedEntities)
            {

                int addedUsersCount = group.AddedUsers.Count();

                int removedUsersCount = group.RemovedUsers.Count();

            }

        }

        public void Terminate()
        {

        }


    }


    public class UserGroupComposer : IUserComposer
    {

        public void Compose(Composition composition)
        {
            composition.Components().Append<UserGroupComponent>();
        }

    }
```

### Expected result / actual result

I woud expect that the AddedUsers list has no items when no new users are added.

inetzo:
Looks like the RepositoryBase.GetMany returns all items if the length of array with ids is 0.

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/10681

PR #10681, opened 2021-07-14
UserService.SavedUserGroup event returns wrong list of AddedUsers 

inetzo:
When no new users are  added, the _userRepository.GetMany is called with an empty array. The empty array causes that all backoffice users are returned instead of 0 users.

### Prerequisites

If there's an existing issue for this PR then this fixes  https://github.com/umbraco/Umbraco-CMS/issues/10679

### Description

- Create a component to catch the UserService.SavedUserGroup event.

```
    public class UserGroupComponent: IComponent
    {

        public UserGroupComponent()
        {

        }

        public void Initialize()
        {
            UserService.SavedUserGroup += UserService_SavedUserGroup;
        }

        private void UserService_SavedUserGroup(Umbraco.Core.Services.IUserService sender, Umbraco.Core.Events.SaveEventArgs<Umbraco.Core.Events.UserGroupWithUsers> e)
        {

            foreach (var group in e.SavedEntities)
            {

                int addedUsersCount = group.AddedUsers.Count();

                int removedUsersCount = group.RemovedUsers.Count();

            }

        }

        public void Terminate()
        {

        }


    }


    public class UserGroupComposer : IUserComposer
    {

        {
            composition.Components().Append<UserGroupComponent>();
        }

    }
```

- Create a usergroup and add a few users.
- Save the group
- Remove 1 user
- Save the group
- The length of AddedUsers should be 0

### DIFF
--- a/src/Umbraco.Core/Services/Implement/UserService.cs
+++ b/src/Umbraco.Core/Services/Implement/UserService.cs
@@ -857,8 +857,9 @@ namespace Umbraco.Core.Services.Implement
                     var groupUsers = userGroup.HasIdentity ? _userRepository.GetAllInGroup(userGroup.Id).ToArray() : empty;
                     var xGroupUsers = groupUsers.ToDictionary(x => x.Id, x => x);
                     var groupIds = groupUsers.Select(x => x.Id).ToArray();
+                    var addedUserIds = userIds.Except(groupIds);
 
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Looking good, thanks @inetzo 👍

PR #10684, opened 2021-07-14
Set isMainPage to true on TemplateRenderer.ExecuteTemplateRendering

### LINKED ISSUE #10683: V9: TemplateRenderer does not render _ViewStart.cshtml

9.0.0-rc001

### Bug summary

When the `RenderTemplateAsync` method of the `IUmbracoComponentRenderer` is called to render a template, only the actual content of the view is returned, the `_ViewStart.cshtml` content is ignored.

### Specifics

On the `ExecuteTemplateRendering` method, when calling `this._viewEngine.GetView`, `isMainPage` is set to `false` resulting in the `_ViewStart.cshtml` never being called.
This should be either `true` or configurable.

### Steps to reproduce

1. Create a `_ViewStart.cshtml` file under the Views directory with any content.
2. Add the code snippet below to a page template, replacing the `1` in line 2 with an existing node id in your umbraco instance. Make sure the node id used uses a different template to prevent infinite recursion.
```
@inject IUmbracoComponentRenderer _umbracoComponentRenderer
@await _umbracoComponentRenderer.RenderTemplateAsync(1);
```
3. Check that the rendered result does not contain the `_ViewStart.cshtml` code

_No response_

lauraneto:
### Prerequisites

If there's an existing issue for this PR then this fixes #10683

### Description
In the `TemplateRenderer`, on the `ExecuteTemplateRendering` method, when calling `this._viewEngine.GetView`, `isMainPage` was set to `false` resulting in the `_ViewStart.cshtml` never being called. This has now been changed to `true`.

### DIFF
--- a/src/Umbraco.Web.Common/Templates/TemplateRenderer.cs
+++ b/src/Umbraco.Web.Common/Templates/TemplateRenderer.cs
@@ -152,7 +152,7 @@ namespace Umbraco.Cms.Web.Common.Templates
         {
             var httpContext = _httpContextAccessor.GetRequiredHttpContext();
 
-            var viewResult = _viewEngine.GetView(null, $"~/Views/{request.GetTemplateAlias()}.cshtml", false);
+            var viewResult = _viewEngine.GetView(null, $"~/Views/{request.GetTemplateAlias()}.cshtml", true);
 
             if (viewResult.Success == false)
             {
--- a/src/Umbraco.Web.Common/Templates/TemplateRenderer.cs
+++ b/src/Umbraco.Web.Common/Templates/TemplateRenderer.cs
@@ -152,6 +152,7 @@ namespace Umbraco.Cms.Web.Common.Templates
         {
             var httpContext = _httpContextAccessor.GetRequiredHttpContext();
 
+            // isMainPage is set to true here in order for the _ViewStart.cshtml to be rendered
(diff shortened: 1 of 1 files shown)


p-m-j:
Hey Laura, thanks for the PR!

I checked back in v8 for UmbracoHelper.RenderTemplate and the previous behavior matches your change such that ViewStart content is included, as such it would be an unintended breaking change for us not to merge this into v9.

Would it be possible for you to add a comment in TemplateRenderer explaining that the argument is only used to determine whether to render ViewStarts or not as it doesn’t become clear until you drill down to RazorViewEngine.CreateCacheResult?

Additionally, for our reference could you describe your use case for RenderTemplateAsync in the PR or the issue as we don’t have a decent example of what it would be useful for vs partial views etc?

lauraneto:
Hi @p-m-j , I updated the PR.

I ended up not using this functionality in Umbraco 9, but I did use it in the past in order to render pages dynamically only by passing in the id of the page, and it worked pretty well.

In my specific case it was in order to let the editor select in the cms which page they wanted to show when an error occurred, so it wouldn't be tied to a specific type/template, id or path.
Then I would have a controller for which I would point the http errors to which would check the error code and look for the respective setting in the cms and render the selected page (using umbracoHelper.RenderTemplate).

p-m-j:
Thanks again!

Issue #10666, opened 2021-07-13
Unable to create a member (against existing/migrated v8 database)


adamhearn:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0-rc001

### Bug summary

Existing v8 site being migrated to v9. It seems that my existing logic of adding a new member is not working in the RC of v9.

The member UI works (can create new members) as expected.

I get no errors/exceptions in the view where the code is being invoked from.

### Specifics

In my view I have:
```
@inject IMemberService _memberService

_memberService.CreateMemberWithIdentity( User.Identity.Name, String.Format( "{0}@domain.com", empId ), System.Environment.UserName, "Member" );
```
**empid** is just a string name calculated prior to the call

It goes through Umbraco layers without issue and no error is raised. However, a SQL trace shows that the following SQL statements are executed:

```
exec sp_executesql N'UPDATE umbracoLock WITH (REPEATABLEREAD) SET value = (CASE WHEN (value=1) THEN -1 ELSE 1 END) WHERE id=@0',N'@0 int',@0=-335
go
exec sp_executesql N'SELECT value FROM umbracoLock WITH (REPEATABLEREAD) WHERE id=@0',N'@0 int',@0=-337
go
exec sp_executesql N'SELECT [umbracoNode].[id] AS [NodeId], [umbracoNode].[uniqueId] AS [UniqueId], [umbracoNode].[parentId] AS [ParentId], [umbracoNode].[level] AS [Level], [umbracoNode].[path] AS [Path], [umbracoNode].[sortOrder] AS [SortOrder], [umbracoNode].[trashed] AS [Trashed], [umbracoNode].[nodeUser] AS [UserId], [umbracoNode].[text] AS [Text], [umbracoNode].[nodeObjectType] AS [NodeObjectType], [umbracoNode].[createDate] AS [CreateDate]
FROM [umbracoNode]
WHERE (([umbracoNode].[id] = @0))',N'@0 int',@0=-1
go
exec sp_executesql N'SELECT MAX(sortOrder)
FROM [umbracoNode]
WHERE ((([umbracoNode].[nodeObjectType] = @0) AND ([umbracoNode].[parentId] = @1)))',N'@0 uniqueidentifier,@1 int',@0='39EB0F98-B348-42A1-8662-E7EB18487560',@1=-1
go
exec sp_executesql N'SELECT [umbracoNode].[id] AS [NodeId]
FROM [umbracoNode]
[...]

PR #10685, opened 2021-07-15
Add integration test for MemberService::CreateMemberWithIdentity

adamhearn:
### Prerequisites

This PR is in support of #10666 

### Description

Added integration test as requested by @Shazwazza to highlight the issue of MemberService::CreateMemberWithIdentity not persisting a member even though the call produced no error and return a member instance.

The test method assertion will fail until the issue is resolved.


Shazwazza:
Great stuff, i'll merge this in and see what to do about a fix.

PR #10686, opened 2021-07-15
Moved nullcheck to top to catch possible nullreference exceptions

deMD:
Moved the nullcheck in the useUmbraco method to the top as `app` was already accessed before the nullcheck initially which would've caused null reference exceptions if `app` is ever null, as per request from @Shazwazza


Shazwazza:
Thanks!

PR #10687, opened 2021-07-15
V9: Fix translations not shown everywhere

### LINKED ISSUE #10645: Umbraco 9 (Dutch) translations are not shown everywhere

9.0.0-rc001

### Bug summary

In 9.0.0-rc001 and before, the translations for example the Save and Publish buttons are not shown. Some parts are translated, but some commonly used parts, like the save buttons, are not. The translations are in the XML file: https://github.com/umbraco/Umbraco-CMS/blob/4f06c1f33a54daf4752a439778626202960db68c/src/Umbraco.Web.UI.NetCore/umbraco/config/lang/nl.xml#L173

afbeelding

### Specifics

_No response_

### Steps to reproduce

Set the user's language to Dutch. The "Save and Publish" button for example should show the translation "Opslaan en publiceren".

I would expect that translation in the XML file are used in the backoffice interface.

nikolajlauridsen:
This fixes #10645, the problem was that translations would only be shown in some places in the backoffice.

The problem was in the `BackOfficeController` `LocalizedText` endpoint that is used to get localized text for javascript. I found two problems here.

The first was that we used `_backofficeSecurityAccessor.BackOfficeSecurity.IsAuthenticated();` to try and authenticate the user, but since this is not an authorized endpoint this would always return false. I changed it to use `this.AuthenticateBackOfficeAsync()` like in the other actions. 

The other issue I found was that we used `Thread.CurrentThread.CurrentCulture` to try and get the users culture, however this won't work, because a user's backoffice language (and therefore culture), is not necessarily the same as their system, for instance, I typically have the backoffice language set to en-US, but `Thread.CurrentThread.CurrentCulture` will return `da-DK` because my system culture in Danish, leading to a very interesting looking half Danish half English backoffice.

To fix this aspect I changed it over so that we now get the culture using the Identity, we get when authenticating the backoffice user, which returns the correct culture info.

### Testing
1. Change a user's language to something other than English
2. Log out and in again
3. Verify that everything that should be translated is

As with anything front-end adjacent, you might have to clear your cache and do a hard reload

### DIFF
--- a/src/Umbraco.Web.BackOffice/Controllers/BackOfficeController.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/BackOfficeController.cs
@@ -4,7 +4,6 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Security.Claims;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authorization;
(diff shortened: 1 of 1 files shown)


Zeegaan:
Tested this fix, this fix solves the issue!

PR #10688, opened 2021-07-15
Support member names using Down-Level Logon Name format

### LINKED ISSUE #10665: Member's username with backslash not supported (domain\username format)

9.0.0-rc001

### Bug summary

I have a site that I'm updating to v9 and am seeing an issue with Members. The problem is that having a backslash in the username field causes the member to be uneditable.

### Specifics

I use IIS with Windows Authentication enabled and record (programatically) the connecting user as a member using the Name property of the authenticated Identity. This is in the format DOMAIN\username. The members show was per:

Web capture_13-7-2021_93045_localhost

Whenever I edit the member I get the following message and the Save button spins forever:

image

Prior releases of v9 (beta) didn't work in various areas but I don't recall that I was unable to edit one of the existing member entries in my DB.

### Steps to reproduce

Migrate a v8 database content with member's usernames in the form domain\username to v9

Edit the member

Username should support the form provided by User.Identity.Name

adamhearn:
### Prerequisites

This fixes #10665

### Description

Added new config option **AllowedUserNameCharacters** to the **Security** section. This value is used when configuring         IdentityOptions for a Member to override the default member name character validation.

- Default value includes the backslash character [in addition to all characters as presently defined by ASP.NET Core]
- Value can be overridden in config

Tested using a v9 test site:

- With no local application config changes, can a username of "DOMAIN\user" be successfully a) created b) edited
- Adding the setting value of "abcd1234"application config changes, are the following usernames able to be added:
  - "DOMAIN\user" - Invalid
  - "abc\123" - Invalid
  - "abc123" - Valid

### DIFF
--- a/src/Umbraco.Core/Configuration/Models/SecuritySettings.cs
+++ b/src/Umbraco.Core/Configuration/Models/SecuritySettings.cs
@@ -15,6 +15,7 @@ namespace Umbraco.Cms.Core.Configuration.Models
         internal const bool StaticHideDisabledUsersInBackOffice = false;
         internal const bool StaticAllowPasswordReset = true;
         internal const string StaticAuthCookieName = "UMB_UCONTEXT";
+        internal const string StaticAllowedUserNameCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-._@+\\";
 
         /// <summary>
         /// Gets or sets a value indicating whether to keep the user logged in.
@@ -49,6 +50,12 @@ namespace Umbraco.Cms.Core.Configuration.Models
         /// Gets or sets a value indicating whether the user's email address is to be considered as their username.
         /// </summary>
         public bool UsernameIsEmail { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets the set of allowed characters for a username
+        /// </summary>
+        [DefaultValue(StaticAllowedUserNameCharacters)]
+        public string AllowedUserNameCharacters { get; set; } = StaticAllowedUserNameCharacters;
 
         /// <summary>
         /// Gets or sets a value for the user password settings.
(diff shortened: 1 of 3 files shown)


adamhearn:
**Please note [as per the issue report] there's a UI issue when validation fails** - the Save button does not return to an active state.

The problem stems from the ValidationProblem instance that isn't constructed with the model or modelstate for errors generated in the ASP.NET layer (IdentityResult). As is, within the JS the following call to **reBindChangedProperties** will fail as it attempts to retrieve data that is not provided **this.getAllProps(savedVariant)** thus the call to **handleSaveError** is never executed.

```
                contentEditingHelper.handleSaveError({
                    err: err,
                    rebindCallback: contentEditingHelper.reBindChangedProperties($scope.content, err.data)
                }),

```

To provide a fix, I tried the following:

```
                    MemberDisplay forDisplay = _umbracoMapper.Map<MemberDisplay>( contentItem.PersistedContent );
                    created.Errors.ToList().ForEach( e => ModelState.AddModelError( e.Code, e.Description ) );
                    return ValidationProblem( forDisplay, ModelState );
```

Unfortunately the validation error neither shows against the field or the form in general. I thought the best plan would be to assign the correct fieldname as the key but the mechanism by which field names used as keys seems to require some form of manual mapping:

```
                    MemberDisplay forDisplay = _umbracoMapper.Map<MemberDisplay>( contentItem.PersistedContent );
                    Dictionary<string, string> propertyMap = new Dictionary<string, string> {
                        { nameof(UmbracoErrorDescriberBase.InvalidUserName), "_Properties._umb_login.invariant.null.value" },
                    };
                    created.Errors.ToList().ForEach( e => ModelState.AddModelError( propertyMap[e.Code] ?? String.Empty, e.Description ) );
                    return ValidationProblem( forDisplay, ModelState );
```

Whilst with that works (client validation shows an appropriate error against the field) I'm sure there's a better way!! Anyone have any hints/pointers to a here's how to build the fieldname snippet please?

Could also add a new issue/fix against that as required.

Zeegaan:
Tested this, and the solution works!

Thanks alot for your contribution!

We view the error with the UI as seperate issue, but you're on the right track!

I will fix it in another pull request and link it here!

Zeegaan:
Here is the PR with the fix:

https://github.com/umbraco/Umbraco-CMS/compare/v9/dev...v9/bugfix/member_validation_doesnt_update_properly?expand=1

PR #10701, opened 2021-07-16
add fix for issue #10700

### LINKED ISSUE #10700: ERROR The type 'ValueType' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'

8.15.0

### Bug summary

Calling extension method .FirstOrDefault on an IEnumerable<T> object throws an error with description:

_The type 'ValueType' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'_

```
var roots = Umbraco.ContentAtRoot();
var root = roots.FirstOrDefault();
```

I found a solution on stakeoverflow: https://stackoverflow.com/questions/49925484/you-must-add-a-reference-to-assembly-netstandard-version-2-0-0-0

Essentially you have to add this to the Web.config:

```
<system.web>
...
  <compilation defaultLanguage="c#" debug="true" batch="true" targetFramework="4.7.2" numRecompilesBeforeAppRestart="50">
    <assemblies>
      <add assembly="netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51"/>
    </assemblies>
  </compilation>
...
```
by replacing:

```
<system.web>
...
<compilation defaultLanguage="c#" debug="true" batch="true" targetFramework="4.7.2" numRecompilesBeforeAppRestart="50" />
...

```

### Specifics

The issue has been added between v8.15.0-rc and final v8.15.0.
I could not figure out which change introduced the issue.
Maybe the use of System.Collections.Immutable as this was added after _RC_ release.

### Steps to reproduce

In a fresh v8.15.0 install create e.g. a simple WebApi class in `\App_Code\SomeController.cs` folder:

```
public class SomeController : UmbracoApiController
{
  public IEnumerable<int> GetIds()
  {
    var roots = Umbraco.ContentAtRoot();
    var root = roots.FirstOrDefault(); //here you get the error
    return null;
  }
}
```

_No response_

idseefeld:
### Prerequisites

If there's an existing issue for this PR then this fixes #10700 

### Description
In the issue I have described how you can reproduce the issue.
Actually I have no clue how you could write an automated test for this🙄

### DIFF
--- a/src/Umbraco.Web.UI/web.Template.config
+++ b/src/Umbraco.Web.UI/web.Template.config
@@ -104,7 +104,11 @@
             <add verb="*" path="DependencyHandler.axd" type="ClientDependency.Core.CompositeFiles.CompositeDependencyHandler, ClientDependency.Core " />
         </httpHandlers>
 
-        <compilation defaultLanguage="c#" debug="false" batch="true" targetFramework="4.7.2" numRecompilesBeforeAppRestart="50" />
+        <compilation defaultLanguage="c#" debug="false" batch="true" targetFramework="4.7.2" numRecompilesBeforeAppRestart="50">
+            <assemblies>
+                <add assembly="netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51"/>
+            </assemblies>
+        </compilation>
 
         <authentication mode="Forms">
             <forms name="yourAuthCookie" loginUrl="login.aspx" protection="All" path="/" />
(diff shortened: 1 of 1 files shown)


emma-hq:
Hey Dirk,

I've just popped a comment on the issue to let you know what we're up to here but we've got a fix we are going to merge for this one. Thank you for the work though! It's very rare that we'd make a change to the web.config as there is a good chance there'll be side effects that we can't predict. 

Will let you know when we have the fix merged! 😄 

Emma

idseefeld:
Hey Emma,
makes sense to leave web.config untouched😊
Dirk

Issue #10682, opened 2021-07-14
Umbraco 9 - Can't add CORS because netcore configuration is hidden in UseUmbraco


deMD:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0-rc001

### Bug summary

I am trying to add CORS to a specific endpoint in my project. For this I need to add UseCors in the configure method of the startup.cs. Sadly this is not possible because it needs to be between UseRouting and UseAuthorization which both are wrapped in a compiled UseUmbraco method.

https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-5.0
```The call to UseCors must be placed after UseRouting, but before UseAuthorization. For more information, see Middleware order.```

### Steps to reproduce

Install umbraco 9 and try to add cors in startup

### Expected result / actual result

Is it not better to keep the NetCore middleware outside of any wrapped method so it's possible for developers to add their own middleware or middleware from others that might need to go "in between" standard NetCore middleware? If this is done for ease of use then I suggest moving the extensions class out of the assembly and place it in de root of the project so it can be edited without the need to ~~copy/paste~~ (nvm the copy/paste `UmbracoApplicationBuilder` is an internal class so that is not possible).

---
_This item has been added to our backlog [AB#12884](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/12884)_

**Comment by deMD** [comment]
Btw this also looks a bit sketch..

```csharp
public static IUmbracoApplicationBuilder UseUmbraco(this IApplicationBuilder app)
        {
            IOptions<UmbracoPipelineOptions> startupOptions = app.ApplicationServices.GetRequiredService<IOptions<UmbracoPipelineOptions>>();

           // Shouldn't the null check come before the RunPrePipeline call?
            app.RunPrePipeline(startupOptions.Value);
            if (app == null)
            {
                throw new ArgumentNullException(nameof(app));
            }
            ...
}
```

Shazwazza:
> // Shouldn't the null check come before the RunPrePipeline call?

yes :) would be fab if you could submit a PR for that

> Is it not better to keep the NetCore middleware outside of any wrapped method so it's possible for developers to add their own middleware or middleware from others that might need to go "in between" standard NetCore middleware? 

[...]

PR #10702, opened 2021-07-16
More flexible Umbraco startup code - new method WithCustomMiddleware

Shazwazza:
Based on this discussion: https://github.com/umbraco/Umbraco-CMS/issues/10682

To be able to fully customize the default middleware pipeline. 

* Renames some interfaces and methods to be more consistent. (i.e. installing middleware should be "Use" not "With".
* Changes the startup builder interface types to be more type safe. (i.e. before you could call WithMiddleware multiple times which is not what we want to allow).
* Cleans up where the startup code lives so that it's all in the `UmbracoApplicationBuilder`.
* Makes `UmbracoApplicationBuilder` public in case anyone wants to ever just create one manually.

The result is that you can now do this if you want full control over the middleware pipeline that Umbraco normally installs.
This allows you to replace it entirely, or just add middleware before or after the default middleware that Umbraco adds.
As an end user this means you don't need to use `IUmbracoPipelineFilter` (which should typically only be used for packages) to achieve that.

This is the default:

```cs
app.UseUmbraco()
    .WithMiddleware(u =>
    {
        u.UseBackOffice();
        u.UseWebsite();
    })
    .WithEndpoints(u =>
    {
        u.UseInstallerEndpoints();
        u.UseBackOfficeEndpoints();
        u.UseWebsiteEndpoints();
    });
```

This is how to take full control:

```cs
app.UseUmbraco()
    .WithCustomMiddleware(u =>
    {
        // You control the pre pipeline calls, if you do not include this
        // then packages IUmbracoPipelineFilter will not execute
        u.RunPrePipeline();

        // You can call the method to register the default middleware
        // that Umbraco installs. You could then register custom
        // middleware before or after this call.
        u.RegisterDefaultRequiredMiddleware();

        // ELSE, you control the entire pipeline, i.e.
        // u.UseUmbracoCoreMiddleware();
        // TODO: Add the rest ...
        // u.AppBuilder.UseStatusCodePages();
        // u.AppBuilder.UseImageSharp();
        // u.AppBuilder.UseStaticFiles();
        // u.AppBuilder.UseUmbracoPlugins();
        // etc ...

        // You control the post pipeline calls, if you do not include this
        // then packages IUmbracoPipelineFilter will not execute
        u.RunPostPipeline();

        // Now include the Umbraco request middleware
        u.UseBackOffice();
        u.UseWebsite();
    })
    .WithEndpoints(u =>
    {
        u.UseInstallerEndpoints();
        u.UseBackOfficeEndpoints();
        u.UseWebsiteEndpoints();
    });
```


src/Umbraco.Web.Common/ApplicationBuilder/UmbracoApplicationBuilder.cs
```diff
@@ -2,50 +2,130 @@
...
+        private Action<IUmbracoApplicationBuilderContext> _customMiddlewareRegistration;
+
+        public UmbracoApplicationBuilder(IApplicationBuilder appBuilder)
         {
-            ApplicationServices = services;
-            RuntimeState = runtimeState;
-            AppBuilder = appBuilder;
+            ApplicationServices = appBuilder.ApplicationServices;
+            RuntimeState = appBuilder.ApplicationServices.GetRequiredService<IRuntimeState>();
+            AppBuilder = appBuilder ?? throw new ArgumentNullException(nameof(appBuilder));
```
AndyButland, line 22: This null check doesn't look quite right... as if `appBuilder` were null we'd get exceptions on the two lines above.  Just need to move the `AppBuilder` property setting first in the method.

src/Umbraco.Web.Common/ApplicationBuilder/IUmbracoEndpointBuilderContext.cs
```diff
@@ -0,0 +1,13 @@
+using Microsoft.AspNetCore.Routing;
+
+namespace Umbraco.Cms.Web.Common.ApplicationBuilder
+{
+
+    /// <summary>
+    /// A builder to allow encapsulating the enabled routing features in Umbraco
```
AndyButland, line 7: Looks like this comment needs updating, as the interface is for a context now rather than a builder.

AndyButland (commented):
Was just asked for a comment on the API here, but that all looks sensible to me with the appropriate hooks.  Spotted a couple of small other things I've noted.

deMD:
Apart from Andy's comments it's looking good. I have discussed with direct colleagues of mines and they would've rather preferred to have no extension methods even though that creates more usings, but this is good as a compromise.

vsilvar:
Looks good to me, flexible enough for anything I can think of.

nikolajlauridsen:
Code looks good to me besides the null check mentioned, and everything still seems to work like intended 👍 

In regards to the API it seems quite sensible to me, fairly simple to use if you only want to add something extra since we have `RegisterDefaultRequiredMiddleware`, but still allows you to completely control the default middleware if needed, so it's a go from here 😄

Shazwazza:
Hi, @deMD , I'm curious as to what you mean by:

> they would've rather preferred to have no extension methods even though that creates more usings, but this is good as a compromise.

Do you have an example of what you mean?

Shazwazza:
side note - since the code that is in the repo in Startup.cs makes it's way into the Umbraco template, IMO it would be a good idea to have these methods documented and then have a tinyurl link to those docs in this code so it's clear what it all does and how an end user can customize the startup.

deMD:
> 
> 
> Hi, @deMD , I'm curious as to what you mean by:
> 
> > they would've rather preferred to have no extension methods even though that creates more usings, but this is good as a compromise.
> 
> Do you have an example of what you mean?

What I really mean is that I personally and my direct colleagues would rather see it somewhat like this by default:
```csharp
public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
        {
            if (env.IsDevelopment())
            {
            }

            app.UseUmbraco()
                .WithCustomDefaultMiddleware(u =>
                {
                    UseCoreUmbracoMiddleWare(u.AppBuilder);*

                    u.AppBuilder.UseStatusCodePages();

                    // Important we handle image manipulations before the static files, otherwise the querystring is just ignored.
                    u.AppBuilder.UseImageSharp();
                    u.AppBuilder.UseStaticFiles();
                    u.AppBuilder.UseUmbracoPlugins();

                    // UseRouting adds endpoint routing middleware, this means that middlewares registered after this one
                    // will execute after endpoint routing. The ordering of everything is quite important here, see
                    // https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-5.0
                    // where we need to have UseAuthentication and UseAuthorization proceeding this call but before
                    // endpoints are defined.
                    u.AppBuilder.UseRouting();
                    u.AppBuilder.UseAuthentication();
                    u.AppBuilder.UseAuthorization();

                    // This must come after auth because the culture is based on the auth'd user
                    u.AppBuilder.UseRequestLocalization();

                    // Must be called after UseRouting and before UseEndpoints
                    u.AppBuilder.UseSession();
                })
                .WithMiddleware(u =>
                {
                    u.UseBackOffice();
                    u.UseWebsite();
                })
                .WithEndpoints(u =>
                {
                    u.UseInstallerEndpoints();
                    u.UseBackOfficeEndpoints();
                    u.UseWebsiteEndpoints();
                });
        }
```
        
We don't feel like wrapping those default middleware calls in extensions methods by default, forcing developers to do this every time they need custom middleware (which in our case would be like 95% of the projects we do servicing enterprises) is beneficial in any way. Obviously it would be even nicer to just be able to use `app.UseRouting()` etc. but if UseUmbraco() really has to return an `IUmbracoApplicationBuilder` then this will work as well.

\* I created a private method to wrap the core umbraco middleware to make the method somewhat shorter
```csharp
private static void UseCoreUmbracoMiddleWare(IApplicationBuilder app)
        {
            app.UseUmbracoCore();
            app.UseUmbracoRequestLogging();
            app.UseUmbracoRouting();
        }
```

Shazwazza:
@deMD thanks a lot for the feedback. As I understand it, you simply just want explicit control of the middleware pipeline which is understandable. We need to consider both scenarios - max flexibility and max simplicity. I have made some further changes which I think make more sense now but please let me know what you think. 

The default registrations remain the same. But to use a custom middleware pipeline you would do this:

```cs
app.UseUmbraco()
    .WithCustomMiddleware(u =>
    {
        // You control the pre pipeline calls, if you do not include this
        // then packages IUmbracoPipelineFilter will not execute
        u.RunPrePipeline();

        // You can call the method to register the default middleware
        // that Umbraco installs. You could then register custom
        // middleware before or after this call.
        u.RegisterDefaultRequiredMiddleware();

        // ELSE, you control the entire pipeline, i.e.
        // u.UseUmbracoCoreMiddleware();
        // TODO: Add the rest ...
        // u.AppBuilder.UseStatusCodePages();
        // u.AppBuilder.UseImageSharp();
        // u.AppBuilder.UseStaticFiles();
        // u.AppBuilder.UseUmbracoPlugins();
        // etc ...

        // You control the post pipeline calls, if you do not include this
        // then packages IUmbracoPipelineFilter will not execute
        u.RunPostPipeline();

        // Now include the Umbraco request middleware
        u.UseBackOffice();
        u.UseWebsite();
    })
    .WithEndpoints(u =>
    {
        u.UseInstallerEndpoints();
        u.UseBackOfficeEndpoints();
        u.UseWebsiteEndpoints();
    });
```

If you use `WithCustomMiddleware` there is now no additional `WithMiddleware` method, its either you use `WithMiddleware` or `WithCustomMiddleware` and if you use `WithCustomMiddleware` then you are in 100% control of everything including running the pre/post IUmbracoPipelineFilter calls. 

I also added the `UseUmbracoCoreMiddleware()` extension like you did.

* Do you think this is better/worse? 
* Do you think it's beneficial to be in control of executing the the pre/post IUmbracoPipelineFilter calls or would you rather have these automatically called somehow?

Thanks!

Shazwazza:
@vsilvar thanks for the questions

> For example, any reason why the user shouldn't add UseStatusCodePages earlier to allow rewrites to go through the UmbracoRouting again?
> Or even, UseImageSharp and UseStaticFiles? Do they have any dependency on UmbracoRouting or any other reason why they couldn't be placed earlier in the pipeline?

I'm unsure about this off the top of my head. I don't think there's any reason why UseStatusCodePages can't occur first but Umbraco does rely on this middleware so it's a requirement. It could go before UseUmbracoCoreMiddleware AFAIK.

For the others, there is a comment in the ordering:

```cs
// Important we handle image manipulations before the static files, otherwise the querystring is just ignored.            
AppBuilder.UseImageSharp();
AppBuilder.UseStaticFiles();
AppBuilder.UseUmbracoPluginsStaticFiles();
```

I don't think there's any reason they couldn't go before `UseUmbracoCoreMiddleware` but don't think it would make any difference either.

deMD:
> 
> 
> @deMD thanks a lot for the feedback. As I understand it, you simply just want explicit control of the middleware pipeline which is understandable. We need to consider both scenarios - max flexibility and max simplicity. I have made some further changes which I think make more sense now but please let me know what you think.

I am trying to get an idea for whom that consideration is necessary? While I understand that you want to make everything as foolproof and simple as possible, I don't see any scenario where hiding these calls is beneficial for developers using Umbraco. If we don't need Middleware we don't care how the method looks, but as soon as we need Middleware there is a huge change we'll need to explicitly define the order.

>     * Do you think it's beneficial to be in control of executing the the pre/post IUmbracoPipelineFilter calls or would you rather have these automatically called somehow?

I think the way it is now it good enough, also makes it transparent when they get called exactly.

Shazwazza:
This has been accepted by the team so I've merged.

@deMD  re:

> I am trying to get an idea for whom that consideration is necessary? While I understand that you want to make everything as foolproof and simple as possible, I don't see any scenario where hiding these calls is beneficial for developers using Umbraco. If we don't need Middleware we don't care how the method looks, but as soon as we need Middleware there is a huge change we'll need to explicitly define the order.

You can add custom middleware without having to replace this entire list everytime https://github.com/umbraco/Umbraco-CMS/blob/v9/dev/src/Umbraco.Web.Common/ApplicationBuilder/UmbracoApplicationBuilder.cs#L84

Apart from CORS, it will be interesting to know which ones must exist within that list that couldn't just be added either before the `app.UseUmbraco()` call or within the `.WithMiddleware` callback. If CORS ends up being required in a lot of installs (which is probably true), then yes that will be annoying. I still think we should just add this by default so people don't have to add it within this list. 

Are there other middlewares that you know that must exist somewhere within this list that cannot exist before or after?

MartinAmsinck:
Starting from Umbraco 9.3 (as far as i can see)

There is added some code for mediaFileManager

https://github.com/umbraco/Umbraco-CMS/blob/eea02137ae0b709861b45ded11882279a990c421/src/Umbraco.Web.Common/ApplicationBuilder/UmbracoApplicationBuilder.cs#L101-L111

With an internal extensions webHostEnvironment.WebRootFileProvider.ConcatComposite:
https://github.com/umbraco/Umbraco-CMS/blob/8eb3ef65f686619d09af478bd6fd563fc90580b2/src/Umbraco.Web.Common/Extensions/FileProviderExtensions.cs#L6-L18

Is there a way around this, besides copying the exension ?

Issue #10106, opened 2021-04-06
Build better API's for testing (and more)


lars-erik:
I've just logged off an awesome Leeds meetup where @Adolfi showed off his V8 tests. We had some good chats afterwards about what's good and not about Umbraco and testing in V8. Amongst others @mattbrailsford challenging how we can claim "unit testing should be easy". Which lead me to this post.

I have a dream for testing in V9 that I couldn't get fulfilled for V8.  
Recently I read this [twitter thread from GeePaw Hill about creating a "making app"](https://twitter.com/GeePawHill/status/1376203909606309890).  
It reminded me of a POC I built once for something like usync or deploy, only code-driven. [Here's a code example, please have a quick look](https://github.com/lars-erik/Our.Umbraco.Fluent.ContentTypes/blob/master/Our.Umbraco.Fluent.ContentTypes.Tests/When_Ensuring_DocumentTypes.cs#L48).  
I'm hoping that dropping the "configuration as code" bomb into a discussion about a .NET Core port is good timing. 😇  
Maybe it doesn't even have to be fluent and C# as linked, it could be JSON. As GeePaw mentions, you could make a simpler "GUI" (fluent?) for the "making app". (Which is a super abstract term, and his other referenced thread has to be dug up)  
We also have a super "GUI" for a "making" app - the backoffice. 😁

A rant is starting, please forgive me and scroll to the feature request heading if you're not inclined to read it.

### &lt;rant&gt;

My biggest issue with Umbraco 8 is that a super-nice option we had of using *Umbraco* to test serialized content-cache as C# classes was removed. We could extract a bit of cached content from the XML-on-disk, and then provide it to a base class that fed it through the whole cache and model factory pipeline. NOT TO MENTION THAT THE NEW CACHE-MECHANISM IS NOT TESTED AT ALL!!!

This is the point in the post where I'll apologize I didn't spend more time on PR's the last couple of years. I regularly work 45+ hours a week and have a family. So I'm sad to say I have limited capacity. I'm happy to contribute rubber ducking and discussion, though. (And suddenly coding efforts)

I do have a [three line code samples that deserializes the btree on disk](https://github.com/lars-erik/umbraco-unit-testing-samples/blob/b18e400751ffd3c502b5293f79da971e895c63fd/Umbraco.UnitTesting.Adapter/ContentCacheSupport.cs#L13). So the step of extracting data from "making-app" generated content from our editors can be easily handled.

[...]

PR #10711, opened 2021-07-19
Decouple NuCache.Property from NuCache.PublishedSnapshot

lars-erik:
### Description

`NuCache.Property` is unnecessarily tight coupled to `NuCache.PublishedSnapshot`.  
This prevents unit-testing parts of NuCache.

With this change, I'm actually able to use NuCache from btree file on disk to proper published content via the NuCache implementations. A few stubs for sure, but not that bad. Content types must be set up _before_ we "boot" per test, but otherwise the perf. & api seems quite good!

image

This is just a minor fix to see if we can get going with more unit testing again.  
I think the state is pretty much "as bad" still for v9, but I'm getting somewhere.  

See https://github.com/lars-erik/umbraco-unit-testing-samples/blob/master-v8/UmbracoV8.Testing.Tests/Can_I_Haz_The_Content_Cache.cs

Cross posting a bit on #10106.

### DIFF
--- a/src/Umbraco.Web/PublishedCache/NuCache/Property.cs
+++ b/src/Umbraco.Web/PublishedCache/NuCache/Property.cs
@@ -124,7 +124,7 @@ namespace Umbraco.Web.PublishedCache.NuCache
         private CacheValues GetCacheValues(PropertyCacheLevel cacheLevel)
         {
             CacheValues cacheValues;
-            PublishedSnapshot publishedSnapshot;
+            IPublishedSnapshot publishedSnapshot;
             IAppCache cache;
             switch (cacheLevel)
             {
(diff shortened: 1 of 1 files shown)


nathanwoulfe (approved):
Hey @lars-erik, this looks fine to me, keen to see what testing magic you're able to conjure!

N

lars-erik:
Cheers, @nathanwoulfe! 🥳  
For the V8 tests to work we still need to use `IgnoresAccessChecksToGenerator` which sucks a bit, but now that everything is public in V9 maybe V8 could follow and get the same shiny public surface?

nathanwoulfe:
@lars-erik I don't see why not... Might be internal HQ reasons, but it certainly makes sense to do so. 

@nul800sebastiaan or @emma-hq might be able to ask the right people?

PR #10712, opened 2021-07-19
Fix unit tests for #10171

nikolajlauridsen:
This fixes all but one unit test for #10171.

While reviewing I noticed the tests were failing and got a bit ahead of myself, and fixed up the obvious ones, this is just using eager locks to ensure that the locks are actually requested for tests.

Feel free to close this if you don't want to use it 😄


Shazwazza:
Thanks @nikolajlauridsen ! I've manually merged some of this but discovered there was issues with the new lazy locking stuff. See this comment: https://github.com/umbraco/Umbraco-CMS/pull/10171#issuecomment-898650955

PR #10713, opened 2021-07-19
Exposed MVC configuration action in AddBackOffice - #10709

### LINKED ISSUE #10709: V9: Cannot access IMvcBuilder

9.0.0-rc001

### Bug summary

Due to the way the code is built, there is no way to configure the IMvcBuilder.
In my case, I wanted to try to enable xml in my controllers by using the `.AddXmlSerializerFormatters()` extension method of the IMvcBuilder but at the moment this is not exposed.

### Specifics

The `.AddMvcAndRazor()` method can receive an `Action<IMvcBuilder>` as a parameter but at the moment `.AddBackOffice()` (from which it is called) does not expose it.

### Steps to reproduce

Install Umbraco 9 and try to enable xml on the controllers by using `.AddXmlSerializerFormatters()`

I expected the IMvcBuilder to either be accessible, or a way to configure it to be exposed.

vsilvar:
### Prerequisites

If there's an existing issue for this PR then this fixes #10709 

### Description
Exposed an optional `IMvcBuilder` configuration action in the `AddBackOffice` method, to allow the end user to more easily configure extra MVC services functionality.

While not strictly required, I think it provides a better user experience.

### DIFF
--- a/src/Umbraco.Web.BackOffice/DependencyInjection/UmbracoBuilderExtensions.cs
+++ b/src/Umbraco.Web.BackOffice/DependencyInjection/UmbracoBuilderExtensions.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Linq;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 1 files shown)


nikolajlauridsen (approved):
Hey @vsilvar, thanks alot for the PR, looks great to me 👍 #H5YR

PR #10715, opened 2021-07-20
Fix ImageSharp.Web Commands

Jeavon:
The commands in the ImageSharpImageUrlGenerator.cs are mostly from ImageProcessor.Web and need to be updated for ImageSharp.Web

I have updated the following

- [x]  `center `= `rxy`
- [x] `anchor `= `ranchor`
- [x] `mode `= `rmode`

To Remove or create ImageWebProcessor to handle

- [ ] `upscale`
- [ ] `animationprocessmode`
- [ ] `heightratio`
- [ ] `widthratio`

If removing the various ratio modes then the ImageCropRatioMode.cs should be removed also

To create ImageWebProcessor to handle

- [ ] **`crop`** - ImageSharp.Web doesn't support cropping from coordinates by default a new ImageWebProcessor will be required and the coordinates will need to be translated to a Rectangle - this exists in https://github.com/umbraco/Umbraco-CMS/pull/10623

### DIFF
--- a/src/Umbraco.Infrastructure/Media/ImageSharpImageUrlGenerator.cs
+++ b/src/Umbraco.Infrastructure/Media/ImageSharpImageUrlGenerator.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Text;
 using Umbraco.Cms.Core.Media;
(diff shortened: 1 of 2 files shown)


Jeavon:
So, this needs some discussion! 

I started removing the commands that are not supported by ImageSharp.Web but I realised this will be a big breaking change as there are a lot of places ratio mode is currently used.

For example, looking at ImageUrlGenerationOptions.cs, we might remove HeightRatio, WidthRatio, UpScale & AnimationProcessMode as these are quite specific to ImageProcessor and unlikely to be supported elsewhere (e.g. Cloudinary SDK). This can lead onto removing ImageCropRatioMode.cs etc
https://github.com/umbraco/Umbraco-CMS/blob/v9/dev/src/Umbraco.Core/Models/ImageUrlGenerationOptions.cs 

I believe the intention is that ImageUrlGenerationOptions.cs should be generic to support multiple ImageUrlGenerator and so removing the ImageProcessor specific options would be the best option.

Alternative to removing all the unsupported commands I think it might be possible to create an ImageWebProcessor for the four commands to make them supported with ImageSharp.Web, this could potentially be a bridge to a future breaking version if necessary.

Any views or suggestions @nikolajlauridsen @Shazwazza @bergmania @filipbech

argentini:
I was going to write some code to handle this for the time being, but the ImageSharp.Web rmode=manual that is documented does not appear to be implemented in their library. There's no way to pass the location and size coordinates to ImageSharp to get an equivalent crop. If anyone has a workaround for this manual mode coordinate thing, please let me know and I will share my final code with you all.

argentini:
I wrote some code to perform the old ImageProcessor crops using ImageSharp. It honors the focal point as well as custom crop coordinates and image resizing. If anyone is interested in using the code let me know. It's not an ImageSharp pipeline. It's something I call in the "saving" event and use to generate my own crop image cache. My project is using custom models and a REST api so this works well for my use case. The code could probably be helpful in creating a proper ImageSharp pipeline. I'm surprised that in the RC launch video the Umbraco team said the release candidate was ready to be put into production when the image cropper URLs don't work at all.

ronaldbarendse:
@argentini You're right about `ImageSharp.Web` currently not supporting the crop rectangle (the 4 percentage based coordinates used by Umbraco's built-in Image Cropper/Media Picker v3): this is added in https://github.com/umbraco/Umbraco-CMS/pull/10623. This PR fixes the focal point and crop modes (by renaming the query string parameters), but we still need to figure out how to handle the other parameters @Jeavon mentions.

ronaldbarendse:
@Jeavon Thanks for starting this PR, like @argentini mentioned, we really need to have this fixed before v9 is released!

I've included your changes into PR https://github.com/umbraco/Umbraco-CMS/pull/10623, as that contains additional (and partly overlapping) fixes, and will close this one.

PR #10716, opened 2021-07-20
Testable BTree to ContentCache in V9 / Separate SqlCe from integration tests

lars-erik:
### Description

I'm looking into whether it's feasible to reuse `Umbraco.Tests.Integration` in the same way we've reused Umbraco.Tests for v7+.  
It won't install OOTB as a dependency with SqlCe being a dependency. There's only _one_ test left depending on it, so I figure it's just as well to remove the dependency from the tests as well.

This PR also includes the preliminary steps to package up `Umbraco.Tests.Integration`.  
If everything goes according to plan (read hope), I'll base another PR off this "completing" any other special needs there is to reuse more from the integration tests. Maybe some of the "tools" and "base classes" could be a "common integration tools" package like the "commont test tools" package?

I haven't updated the build pipeline to run the SqlCe test. Can do before leaving draft.


lars-erik:
There was only three leftovers of internal things that needed to be public for my V9-compatible `UmbracoSupport` POC.  
Awesome job on making everything else better™! #h5yr

I'll leave this here for discussion / review, and hopefully someone can give me a pointer to what's left in order to properly package and distribute the integration tests. I'm kind of inclined to move the stuff we want to use into its own Tests.Integration.Common assembly.

Looking forward to hearing thoughts. :)

(Dropping a short note in #10106 as well)

bergmania:
Hi @lars-erik.. 

I'll take a look at this, but first I just merge latest v9/dev into your branch :)

bergmania:
@lars-erik, Nice changes.. 

I just updated the "windows" solution file too, to include the SqlCe integration tests + Added building on azure (hopefully, as this cannot be tested without pushing).

I like the idea of separating some common stuff in a `Tests.Integration.Common` that should be the distributed package instead of the full one including all our tests too. I think that could be done in a separate PR, so we can merge this as soon as build server is happy.

lars-erik:
Awesome! I do have to live with 8 for a while too. Hope you'll have a look at that PR as well. 🙃
Hopefully able to work towards common on this.

bergmania:
Now we should have an `Umbraco.Tests.Integration` nuget package in the nightly builds (from tomorrow)

lars-erik:
🥳

PR #10717, opened 2021-07-20
Update umbeditorview.directive.js

abjerner:
The `<umb-editor-view />` directive has a "hidden" parameter as it's only used in the view (and not in the JS, which may be why it has been overlooked.

By default, `<umb-editor-view />` creates a bit of distance to the bottom so there is space for a `<umb-editor-footer/>` element. For editors that don't have a footer, the `footer` parameter of `<umb-editor-view />` can be set to `false`.

This commit targets `v8/contrib`, but the parameter is also missing in `v9/dev`.


nathanwoulfe (approved):
Thanks @abjerner, quick and easy win.

PR #10670, opened 2021-07-13
Umbraco 9 - LoginProvider Duplicate Key

cleversolutions:
### Prerequisites

- [X ] I have added steps to test this contribution in the description below

This PR fixes #10656 

### Description
This PR updates the IX_umbracoExternalLogin_LoginProvider to include both loginProvider and userId. It also shortens the loginProvider to 400 chars instead of 4000 which was throwing a warning regarding the index anyway. 

I updated the V9 migration to make the same change to the index; however, I didn't want to shorten anybody's existing loginProvider column so I didn't change the length of that column in the migration.

This change allows more than one back office user to be able to use the same external login provider.

### Testing:
 * Ran all unit tests (all passed)
 * Ran all integration tests (all passed)
 * Test on a new site:
   * Build and run Umbraco.Web.UI.Netcore
   * Provision a new site with a new database
   * Check that IX_umbracoExternalLogin_LoginProvider now indexes both loginProvider and userId
 * Link two BackOffice user accounts to the same login provider and do not observe the duplicate key exception

PR #10721, opened 2021-07-21
external login changes/enhancements

Shazwazza:
## Notes 
* Incorporates community PR: #10670
* Better support for custom DB types - so an underlying string can be used to configure the custom type instead of limiting it only to what we have defined in an enum. This is required because Umbraco Id requires custom sql server column types that was not contained in our old enum. This is now much more flexible.
* Fixes SqlSyntaxProvider classes. There were public fields that should be properties, the whole initialization logic was problematic. DateTimeOffset column types were wrong.
* Adds InviteDateUtc property to the BackOfficeIdentityUser so it can be persisted via identity.
* Fixes the CSS ordering for the social icons.
* Fixes user deletion where we weren't clearing data from UserLogin table resulting in an exception. (Not sure if there's an existing issue for this, I couldn't find one)
* Email notification changes - We previously were only sending notifications when SMTP was configured which is not what we want. We want the notification to fire all of the time and we need to be able to tell Umbraco that the notification has handled sending the email. So this has been updated.
* Changes BackOfficeExternalLoginProviderOptions to use the IOptions pattern with named options so they can be set after DI. We also resolve the external login provider DisplayName from the authentication scheme like we are supposed to instead of relying on the hard coded display name that is passed through to the BackOfficeAuthenticationBuilder which was problematic since the DisplayName can be set via IOptions<AuthenticationOptions>.


bergmania (approved):
Code changes looks good..

I tested upgrades of both SqlServer and SqlCe. Both works as expected.

I added a single commit, that changes the upgrade screen to only show the SemVer version and not also the unique build id.

PR #10722, opened 2021-07-22
V9: Fix multi url picker

### LINKED ISSUE #10694: MultiUrlPicker throws api error

v9 RC001

### Bug summary

When I try to create a page with a multiurl picker I am encountering this error upon page load. 

Request error: The URL returned a 404 (not found):
/umbraco/backoffice/umbracoapi/entity/GetUrl

### Specifics

image
Each time you click on a page with a multi url picker on it, you get this error.

### Steps to reproduce

On a clean install
1. Create a doctype of any page
2. Add page of content
2. Add a multi url picker property to doc type
3. Select a page in the picker
4. Save and publish
5. Load page. 

No error

nikolajlauridsen:
This PR fixes #10694.

Using the multi url picker would result in a 404 error when it was rendered to the front page.

The cause was that the UDI parameter was named `udi`, and with our `EntityController` we use a custom attribute `ParameterSwapControllerActionSelector` to handle ambiguous actions, however, this doesn't work if the parameters differ in name between the two actions. Since the we have a `GetUrl` action with a parameter `id` and the action is registered as such: `[ParameterSwapControllerActionSelector(nameof(GetUrl), "id", typeof(int), typeof(Udi))]`, the attribute was unable to resolve the request when the parameter was `udi`.

To fix this I renamed the `udi` parameter to `id` and reflected the change in the javascript code as well, after this the action is invoked correctly and we don't get any more 404's

### Testing
Follow the steps in #10694, after the PR this should no longer fail, remember to rebuild the client project though for the javascript changes to take effect.

### DIFF
--- a/src/Umbraco.Web.BackOffice/Controllers/EntityController.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/EntityController.cs
@@ -258,16 +258,16 @@ namespace Umbraco.Cms.Web.BackOffice.Controllers
         /// <summary>
         /// Gets the URL of an entity
         /// </summary>
-        /// <param name="udi">UDI of the entity to fetch URL for</param>
+        /// <param name="id">UDI of the entity to fetch URL for</param>
         /// <param name="culture">The culture to fetch the URL for</param>
         /// <returns>The URL or path to the item</returns>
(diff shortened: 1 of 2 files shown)


Zeegaan:
Code makes sense and looks good, i tested it, and this solution fixes the issue! 👍

PR #10731, opened 2021-07-23
V9 Updates exception message to show correct path to logs

nikcio:
### Prerequisites

### Description

This just updates the exception message shown on a failed unattended install to show the correct path to the log files.


AndyButland (approved):
Can confirm this is now referencing the correct path.

PR #10735, opened 2021-07-25
Examine dashboard search adjustments

bjarnef:
### Prerequisites

### Description
This PR add a few adjustments to Examine dashboard in Settings section:

- Replace search input with `<umb-search-filter>` component.
- Replace icons with `<umb-icon>` component.

Currently `<umb-search-filter>` only acted on change, but not on a keypress/keydown event, e.g. search on enter.
I have handled this a bit similar to have it works in `umbMiniSearch` which has an `onSearch` function.
https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web.UI.Client/src/common/directives/components/umbminisearch.component.js#L13

We could in this case call the function `onFilter`, but I thought it might be to close to the `onChange` function.

## Test notes
- Check Examine search works searching via enter in search input and when clicking Search button. With default starter kit search e.g. on "home".
- Test filtering on e.g. Icon Picker still works.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbsearchfilter.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/forms/umbsearchfilter.directive.js
@@ -40,6 +40,8 @@
 
         vm.$onInit = onInit;
         vm.change = change;
+        vm.keyDown = keyDown;
+        vm.blur = blur;
 
         function onInit() {
             vm.inputId = vm.inputId || "umb-search-filter_" + String.CreateGuid();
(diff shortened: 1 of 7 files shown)


nul800sebastiaan:
Thanks @bjarnef, I found a funny one: 

image

Didn't look too healthy.. ;-) 

But that seems to have been a copy/paste error, I fixed it here: [`75007df` (#10735)](https://github.com/umbraco/Umbraco-CMS/pull/10735/commits/75007df80c0a71acb30a35b2674c5daabc0de514)

PR #10720, opened 2021-07-21
Update LocalizedTextServiceExtensions.cs - use the culture parameter, instead of Thread.CurrentThread.CurrentUICulture

soreng:
Use the culture passed in, instead if the static Thread.CurrentThread.CurrentUICulture

### Prerequisites

- [ ] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes

PR #10741, opened 2021-07-26
Localize using the provided culture rather than the thread culture (#3)

nathanwoulfe:
Localize method accepted a CultureInfo parameter, but passed `Thread.CurrentThread.CurrentUICulture` to the internal localization function. 

This means that if I requested (in code) localization to dn-DK, I'd still get back an English string, based on my UI culture. That's an issue if I'm localizing strings for users other than the logged in user - for example when creating a batch of workflow emails to users whose culture can differ from mine.


bjarnef:
@nathanwoulfe it seems to be a duplicate of https://github.com/umbraco/Umbraco-CMS/pull/10720 😊

nathanwoulfe:
@bjarnef it seems you are correct, and very, very speedy in being correct.

PR #10742, opened 2021-07-26
Migrated Umbraco.TestData to .NET 5.

AndyButland:
@Shazwazza - you raised this on the Azure board, so if you've chance to review this week please do... otherwise I'll create a task for Bjarke to review or re-assign next week.

I've converted the project with the usual updates for namespaces, injected dependencies, configuration, services setup etc.  Tested via the instructions in the `readme.md` file and checked some of the end-points on the `LoadDataController`, which all seem fine.

I've added a comment and removed a line reporting details of the IIS site name, which I'm not sure we could retrieve.  And the `Restart()` method now just does a shut-down - again, not sure we can do more than that.


src/Umbraco.Web.UI.NetCore/Umbraco.Web.UI.NetCore.csproj
```diff
@@ -16,6 +16,7 @@
     <ProjectReference Include="../Umbraco.Web.Common/Umbraco.Web.Common.csproj" />
     <ProjectReference Include="../Umbraco.Web.Website/Umbraco.Web.Website.csproj" />
     <ProjectReference Include="../Umbraco.Persistence.SqlCe/Umbraco.Persistence.SqlCe.csproj" Condition="'$(OS)' == 'Windows_NT'" />
+    <ProjectReference Include="../Umbraco.TestData/Umbraco.TestData.csproj" Condition="'$(Configuration)' == 'Debug'" />
```
bergmania, line 19: This seems strange.. Why do we need this project reference?

AndyButland: Good question.  There was a note in the readme for the test data project, saying that: _Currently this project is referenced in the Umbraco.Web.UI project but only when it's being built in Debug mode (i.e. when testing within Visual Studio)._.

So I was attempting to replicate that setup for V9.

But now I check, there doesn't seem to be this project reference in V8, so I guess the readme note is out of date.

Hence have removed.

bergmania (approved):
Seems to work as expected.. Nice work

PR #10750, opened 2021-07-27
Removes `inert` attribute when `editorService.closeAll()` is called

mattbrailsford:
Same as #10710 but re-applied for v9 which seems to have the same issue.


nul800sebastiaan:
This will eventually be merged up to v9 😄  We're still merging all of v8 up to v9!

mattbrailsford:
Ahh, ok, I thought that had stopped. Np 👍

PR #10752, opened 2021-07-27
Abstracts out IIndexRebuilder so it can be replaced.

Shazwazza:
Currently there is no way to control externally how index rebuilds occur (without a ton of hacks). This PR creates an abstraction for IIndexRebuilder so that it can be replaced if necessary by extensions. Currently in ExamineX there is an issue because Umbraco force rebuilds indexes on a cold boot which requires a lot of hacks to deal with. https://github.com/SDKits/ExamineX/issues/46

### DIFF
--- a/src/Umbraco.Examine/IIndexRebuilder.cs
+++ b/src/Umbraco.Examine/IIndexRebuilder.cs
@@ -1,0 +1,29 @@
+﻿using Examine;
+
+namespace Umbraco.Examine
+{
+    /// <summary>
+    /// Used to rebuild Umbraco indexes.
+    /// </summary>
+    public interface IIndexRebuilder
+    {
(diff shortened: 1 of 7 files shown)


nathanwoulfe (commented):
@Shazwazza code makes sense to my Monday-morning brain. Anything in particular that should be tested here?

Shazwazza:
There's nothing too special going on here but possibly binary breaking changes might need to be reviewed to see if this would break anyone's installs.

nzdev:
Looks good to me

nzdev:
Any chance of getting this one merged?

nathanwoulfe:
Would need to be retargeted to v9/contrib, which may or may not be a messy process...

PR #10755, opened 2021-07-27
Use umb-icon component in user history

bjarnef:
### Prerequisites

### Description
I noticed in user history it wasn't showing node icons, if using custom SVG icons.
This PR fixes that by using the `<umb-icon>` component.

**Before**

image

**After**

image


bjarnef:
I am not sure the history works as expected though.
It seems to be gone when refreshing page.

Ideally it should probably store the history in e.g. local storage.

nathanwoulfe (approved):

nathanwoulfe:
Thanks @bjarnef. Can't be too many more non-umbIcon icons out in the wild any more?

bjarnef:
@nathanwoulfe at least a few more open PRs from me regarding umb-icon to use custom SVG icons 😊😁

Issue #10754, opened 2021-07-27
Backdrop is removed when closing user overlay in tour


bjarnef:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

8.15.0

### Bug summary

I noticed when taking "Getting Started" tour in content the tour has the backdrop to highlight on the specific element.
However in step 9/12 when close user overlay the backdrop is lost.

### Steps to reproduce

Run "Getting Started" tour in content section and notice in step 9/12 when close user overlay, the backdrop / backshadow is removed.

![cN70bTRVvA](https://user-images.githubusercontent.com/2919859/127204079-4f4ae4c8-48cc-4f6c-93b7-b0efd1e2be20.gif)

bjarnef:
@BatJan I have noticed you previous has working with the backdrop. Maybe something has affected this?

nul800sebastiaan:
I just tried a few versions and the backdrop started disappearing in 8.7.0.

umbrabot:
Hi @bjarnef,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

bjarnef:
Could be related to changes of some z-indexes or moving DOM elements, maybe changes in this PR https://github.com/umbraco/Umbraco-CMS/pull/7923

bjarnef:
In content section when running tour "Create content -> Creating content" tour and clicking "Content Page" menu item in step 4/9 something seems to set `backdrop.show` to `false` as the DOM element is removed here:

![image](https://user-images.githubusercontent.com/2919859/127371734-0b9a7cd8-e6ec-4ff1-9003-9c88782071a7.png)

bjarnef:
It seems it is an issue with this function as the leftcolumn actually does have the `above-backdrop` class.
https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web.UI.Client/src/common/directives/components/tree/umbtree.directive.js#L327-L330

[...]

PR #10756, opened 2021-07-27
Only make umb-button-ellipsis visible when above backdrop

bjarnef:
### Prerequisites

### Description
After `<umb-button-ellipsis>` was implemented and used in tree I have noticed when starting a tour all "options" buttons in tree become visible.

I think this should be a bit more specific when the tree get focus.

In content section run tour "Create content -> Creating content".

**Before**

image

**After**

image

image

Could be improved further, but at least it doesn't show all `<umb-button-ellipsis>` when tour is active, but only then the tree is above backdrop.

When running this tour I also noticed the issue reported here: https://github.com/umbraco/Umbraco-CMS/issues/10754


nul800sebastiaan:
Thanks @bjarnef, looks good to me! 👍

PR #10759, opened 2021-07-28
Replace angular.forEach with Utilities.forEach

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/7718

### Description
Previous this PR https://github.com/umbraco/Umbraco-CMS/pull/7928 was submitted to replace `angular.forEach`
However it seems not all references has been replaced or maybe because other PRs has been merged afterwards.

It can use vanilla JS `array.forEach` or `Utilities.forEach`.
Current both versions are used in core, but `Utilities.forEach` is a bit safer as it only iterate if array variable not is null or undefined.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbutton.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/buttons/umbbutton.directive.js
@@ -135,7 +135,7 @@ Use this directive to render an umbraco button. The directive can be used to gen
             if (vm.buttonStyle) {
 
                 // make it possible to pass in multiple styles
-                if(vm.buttonStyle.startsWith("[") && vm.buttonStyle.endsWith("]")) {
+                if (vm.buttonStyle.startsWith("[") && vm.buttonStyle.endsWith("]")) {
 
                     // when using an attr it will always be a string so we need to remove square brackets
                     // and turn it into and array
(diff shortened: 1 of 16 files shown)


nathanwoulfe (approved):
Nice work @bjarnef, as always. Wonder if it's worth considering adding JS lint rules for the angular methods?

nul800sebastiaan:
@bjarnef @nathanwoulfe - ehmmmm... not sure what's happening here?? Help!! 😁 

This started happening with this commit, so something is not quite right here..

https://user-images.githubusercontent.com/304656/128375242-bc2297c0-9086-42ca-b3a6-96d1acde5a52.mp4

bjarnef:
@nul800sebastiaan strange. It should be the same just with a null and undefined check.
https://github.com/umbraco/Umbraco-CMS/blob/7cccbb3fbf7bc8c7be6ae3f01df0face2fa53d9a/src/Umbraco.Web.UI.Client/src/utilities.js#L109-L117

It seems there's something failing in `checkfocus`  in a directive/component.

bjarnef:
@nul800sebastiaan I think it fails here because `obj` passed in is an jQuery object and not and array. Maybe it would also be more correct naming the parameter name `array/arr` instead of `obj`.

image

nul800sebastiaan:
I'm not sure, currently I don't understand what this means! 😅 😊 I'd love some help here if you have a bit of time!

bjarnef:
It seems `element.find("li :tabbable");`returns an empty array. Not sure if it previous returned something?
However the `forEach()` will iterate this:

```
const forEach = (arr, iterator) => {
    if (arr) {
        return arr.forEach(iterator);
    }
    return arr;
};
```

Also to clearify: https://stackoverflow.com/a/31096661/1693918

> Object does not have forEach, it belongs to Array prototype.

bjarnef:
I made some changes in `utilities.js` but it didn't seem these wasn't compiled using `dev` gulp task.
Also when clicking root node I get this error from PR https://github.com/umbraco/Umbraco-CMS/pull/10762

image

PR #10762, opened 2021-07-28
Tour backdrop missing

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10754

### Description
Various changes to context menu and other has caused backdrop from being hidden/removed e.g. when selecting a menu item in context menu. This is okay in the regular flow, but not in tour mode, where we want focus/highlight on the specific context.

I have also adjust some code to just use vanilla js and the exact same `closeBackdrop` function existed twice in code.

jbPDUcBYBY

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/navigation.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/navigation.service.js
@@ -136,7 +136,7 @@ function navigationService($routeParams, $location, $q, $injector, eventsService
 
     function showBackdrop() {
         var backDropOptions = {
-            'element': $('#leftcolumn')[0]
+            'element': document.getElementById('leftcolumn')
         };
         backdropService.open(backDropOptions);
     }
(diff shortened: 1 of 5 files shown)


bjarnef:
Running tour Getting Started -> Introduction

moCp279AOY

bjarnef:
@nul800sebastiaan this should fix the missing backdrop when tour is open.

nul800sebastiaan:
Great work @bjarnef and nice to not have that duplicate code in there any more! 👍

src/Umbraco.Web.UI.Client/src/common/services/navigation.service.js
```diff
@@ -118,19 +118,25 @@ function navigationService($routeParams, $location, $q, $injector, eventsService
...
-        var isLeftColumnOnTop = leftColumn.hasClass(aboveClass);
 
-        if(isLeftColumnOnTop){
+        var tourIsOpen = document.body.classList.contains("umb-tour-is-visible");
+        if (tourIsOpen) {
+            return;
+        }
+
+        var aboveClass = "above-backdrop";
+        var leftColumn = document.getElementById("leftcolumn");
```
nul800sebastiaan, line 128: Apparently this could result in a null result, so the tests were failing, fixed in https://github.com/umbraco/Umbraco-CMS/commit/d55b740b16faf8b8faaffa533ee5130371ae6169

bjarnef: In that case `if (leftColumn)` should be sufficient :)

PR #10763, opened 2021-07-28
Don't squeeze avatar when help panel is open

bjarnef:
### Prerequisites

### Description
On laptop size (1280 x 720 pixels) when opening help panel, the avatar icon is squeezed. It might doesn't have same spacing right to the avatar, but I think it looks nicer and it could optimize it further later. Maybe if "header actions" is added sometime in future: https://github.com/umbraco/Umbraco-CMS/pull/9016

**Before**

image

**After**

image


nathanwoulfe (approved):
Nice find @bjarnef!

PR #10766, opened 2021-07-29
Changed ConfigConnectionString to accept more valid connectionstrings

### LINKED ISSUE #10749: "Cannot determine provider name from connection string" exception when using a valid connection string

9.0.0-rc001

### Bug summary

When using a [valid](https://www.connectionstrings.com/sql-server/) connectionstring consisting of "Data Source" and "Initial Catalog" the [ParseProvider](https://github.com/umbraco/Umbraco-CMS/blob/f1bc76e7094db1e548ee53f708a845c5f0d48280/src/Umbraco.Core/Configuration/ConfigConnectionString.cs#L20) method throws an exception.

### Specifics

I don't now why that method is implemented the way it is, but it's clearly causing issues it shouldn't. If there is only support for SqlCE or Sql Server is a simple check for the extension enough?
```csharp
return (dataSource.EndsWith(".sdf")) 
? Umbraco.Cms.Core.Constants.DbProviderNames.SqlCe;
: Umbraco.Cms.Core.Constants.DbProviderNames.SqlServer;
````

### Steps to reproduce

Use a connectionstring is the following format:
`Data Source=[server];Initial Catalog=[database];User Id=[userid]Password=[password];`

A working database connection

deMD:
### Prerequisites

Fixes #10749

### Description
Data Source is also a valid value in a SQL Server Connectionstring. This PR changed the ConfigConnectionString to take that into account. 

I also added testcases for the connectionstring variations code has been changed for.


elit0451 (approved):

elit0451:
Works as expected! Good job 🎉 Merging this one

PR #9671, opened 2021-01-17
Reduce memory allocations and speed up text localization

nzdev:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

If there's an existing issue for this PR then this fixes N/A

### Description
What?
Greatly reduce memory allocations and speed up text localisation by optimising TextLocalizationService.

Why?
Memory profiling showed megabytes of allocations caused by TextLocalizationService.Localize() while just visiting nodes in the backoffice.

```
// * Summary *

BenchmarkDotNet=v0.11.3, OS=Windows 10.0.18362
Intel Core i5-8265U CPU 1.60GHz (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
  [Host]     : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.8.4250.0
  Job-JIATTD : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.8.4250.0

IterationCount=3  IterationTime=100.0000 ms  LaunchCount=1
WarmupCount=3

                     Method |       Mean |      Error |    StdDev | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
--------------------------- |-----------:|-----------:|----------:|------:|--------:|------------:|------------:|------------:|--------------------:|
   OriginalDictionaryGetAll |  23.015 ms | 44.5607 ms | 2.4425 ms |  0.21 |    0.01 |   1750.0000 |           - |           - |           6008954 B |
          OriginalXmlGetAll | 101.288 ms | 84.2304 ms | 4.6170 ms |  0.94 |    0.05 |  13000.0000 |           - |           - |          43467420 B |
 OriginalDictionaryLocalize |  17.674 ms | 17.5819 ms | 0.9637 ms |  0.16 |    0.00 |   1000.0000 |           - |           - |           3766319 B |
        OriginalXmlLocalize | 108.289 ms | 83.5021 ms | 4.5770 ms |  1.00 |    0.00 |  14000.0000 |           - |           - |          44065592 B |
         OptimizedXmlGetAll |  13.112 ms | 12.5939 ms | 0.6903 ms |  0.12 |    0.01 |   1750.0000 |           - |           - |           5849228 B |
[...]

PR #10767, opened 2021-07-29
Fix: Issue when upgraded from 8.14 to 8.15 when inviting user into the backoffice. #10746

### LINKED ISSUE #10746: Issue when upgraded from 8.14 to 8.15 when inviting user into the backoffice.

8.14.0 & 8.15

### Bug summary

On our instance of umbraco 8.14 i could invite users into the backoffice (stmp config was working)

When upgrading to 8.15 the invite loader keeps on spinning and when running locally i get a stackoverflow error. This is when using the exact same STMP config. 

Debug console output:
An unhandled exception of type 'System.StackOverflowException' occurred in Unknown Module.

The user is still made but no mail ever gets send to the emailaddress

### Specifics

_No response_

### Steps to reproduce

Install 8.14, check if the smtp settings work and invite user.
Upgrade using nuget to 8.15, run the website and let it install the upgrade. 
Invite user again 

Invite user email send

nul800sebastiaan:
This fixes #10746 

There was a recursive call in this method spinning into an infinite loop, which is why there was no error messages or feedback, IIS just gives up after a while.

I also removed an unreachable method. I left some comments here with details: https://github.com/umbraco/Umbraco-CMS/pull/9671#commitcomment-54134090

Make sure to squash & commit this one after review so we only have 1 commit to cherry-pick.

### DIFF
--- a/src/Umbraco.Core/Services/LocalizedTextServiceExtensions.cs
+++ b/src/Umbraco.Core/Services/LocalizedTextServiceExtensions.cs
@@ -72,30 +72,6 @@ namespace Umbraco.Core.Services
         }
 
         /// <summary>
-        /// Localize using the current thread culture
-        /// </summary>
-        /// <param name="manager"></param>
-        /// <param name="area"></param>
-        /// <param name="alias"></param>
-        /// <param name="tokens"></param>
-        /// <returns></returns>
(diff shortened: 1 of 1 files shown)


emma-hq (approved):
Yip! All tested and all good

PR #10789, opened 2021-08-04
v9: Resolving UmbracoHelper in custom middleware was null

Zeegaan:
Follow-up fix for https://github.com/umbraco/Umbraco-CMS/pull/10551

# Notes
- Added IUmbracoHelperAccessor interface
- Added IUmbracoHelperAccessor class
- Added IPublishedContentQueryAccessor interface
- Added PublishedContentQueryAccessor class
- Added IUmbracoHelperAccessor as a service
- Added IPublishedContentQueryAccessor as a service

# How to test
- Make your own custom middleware
- Try to inject IUmbracoHelperAccessor & IPublishedContentQueryAccessor


Shazwazza:
I still think my comment on the issue is relevant https://github.com/umbraco/Umbraco-CMS/pull/10551#pullrequestreview-700291297

Adding accessors just means you now have to null check all of them since they will not resolve anything without an umbraco context which is a scoped based service. So sure, you can inject these in a singleton but they will always yield null.

Zeegaan:
Updated the `IUmbracoHelperAccessor` and` IPublishedContentQueryAccessor` interfaces to have a `TryGetxxx` instead, to force people to do null checks

bergmania:
> I still think my comment on the issue is relevant [#10551 (review)](https://github.com/umbraco/Umbraco-CMS/pull/10551#pullrequestreview-700291297)
> 
> Adding accessors just means you now have to null check all of them since they will not resolve anything without an umbraco context which is a scoped based service. So sure, you can inject these in a singleton but they will always yield null.

IMO the accessors are basically syntactical sugar, so people are not forced to have all their custom classes registered as scoped or need to use the service provider in the method, like we did couple of times. e.g getting the service provider from HttpContext.

bergmania:
Main part merged together with this PR https://github.com/umbraco/Umbraco-CMS/pull/10833

Issue #10736, opened 2021-07-25
Block thumbnail picker lets you pick an image outside of wwwroot


AaronSadlerUK:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0-rc001

### Bug summary

You can select an block thumbnail image outside of wwwroot, which then returns a 404 error

However if you put the image inside of wwwroot, it will append wwwroot to the url also returning a 404 error

### Steps to reproduce

Upload an image to a folder outside of root
Set a block to use this image thumbnail
Try to add a block showing a thumbnail

### Expected result / actual result

You should either be restricted from selecting an image outside of the wwwroot

PR #10794, opened 2021-08-05
V9: Limit the BlockList thumbnail picker to only the directories serving static files 

elit0451:
Fixes: #10736 

## Details
- When selecting an image for your BlockList thumbnail, the choice will be limited to only images from the static files serving directories;
- If the image is located under the webroot folder, the thumbnail path is kept as, for example, `~/media/...` instead of `~/wwwroot/media/...`.

## Test
- Set up a doc type with a BlockList property editor;
- Upload a few images in the Media section and/or under _"Umbraco.Web.UI.NetCore\App_Plugins"_ folder;
- Add a block and under **Catalogue appearance** try selecting different images for the block's thumbnail and make sure it is all working properly (the image is rendered);
- &#43; follow the "**Steps to reproduce**" from the original issue ⬆️.

-------------------
❗ NB: remember to run `npm install` and `npm run build` as there are client-side changes.


Zeegaan:
Tested and review this, and it works!

Code makes sense aswell!

bjarnef:
It could eventually service this as a `filePicker` in editor service, so it is a bit simpler to re-use: https://github.com/umbraco/Umbraco-CMS/blob/v9/dev/src/Umbraco.Web.UI.Client/src/common/services/editor.service.js#L1056-L1095

In v8 the "filePicker" was also used in packages section, but with NuGet packages in v9, I am not sure if this is still used here.

However it is probably also useful to select a css and custom HTML view in block configuration.

PR #10797, opened 2021-08-05
Fixes: Unable to unlock locked-out member #10678

### LINKED ISSUE #10678: Unable to unlock locked-out member

8.14.0

### Bug summary

After clicking on the Unlock button in the left column of the user profile, it would appear that the user is actually unlocked based on the notifications provided. However, after navigating away from the user or reloading the page, the locked status returns on the user.

### Specifics

I recall this bug also from version 7 and seem to recall that is was related to a datestamping issue. Note that this database has been upgraded from V7, so this might be part of the issue.

In order to unlock the user, I needed to manually alter the database in dbo.umbracoUser, setting user noConsole to false (I also set lastLockoutDate back to null as well but not sure if that was required). Interestingly after all the database changes were made, the user was still locked out in the UI. I needed to touch the web.config file to trigger the changes.

I do note that the response from the server indicates no errors:

```
)]}',
{"notifications":[{"header":"xxx xxxx is now unlocked","message":"","type":3}],"message":"xxx xxx is now unlocked"}
```

### Steps to reproduce
Lock out a test user by updating userNoConsole (true) in table dbo.umbracoUser
Navigate to the locked out user in umbraco
Click the unlock button
Reload the page - user is still logged out

Expect the user to actually be unlocked

jattwood:
### Prerequisites

Fixes #10678

### Description
Basically, `UsersConroller.cs` `PostUnlockUsers` passes `DateTime.Now` to `UserManager.SetLockoutEndDateAsync` to unlock users. It's possible for that value to still evaluate to `==` inside `UserManager.SetLockoutEndDateAsync`. Regardless of the outcome of the if statement in `SetLockoutEndDateAsync`, that method always returns positive giving the false illusion that the user has been unlocked. Changing `>=` to just `>` should remedy this. Alternately we could also pass a date time in the past to `UserManager.SetLockoutEndDateAsync` as well. 

Line 775 could be changed to:

`var unlockResult = await UserManager.SetLockoutEndDateAsync(u, DateTimeOffset.Now.AddMinutes(-1));`

### DIFF
--- a/src/Umbraco.Core/Models/Identity/BackOfficeIdentityUser.cs
+++ b/src/Umbraco.Core/Models/Identity/BackOfficeIdentityUser.cs
@@ -259,7 +259,7 @@ namespace Umbraco.Core.Models.Identity
         {
             get
             {
-                var isLocked = LockoutEndDateUtc.HasValue && LockoutEndDateUtc.Value.ToLocalTime() >= DateTime.Now;
+                var isLocked = LockoutEndDateUtc.HasValue && LockoutEndDateUtc.Value.ToLocalTime() > DateTime.Now;
                 return isLocked;
             }
         }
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Thanks for that @jattwood! And congrats on your first contribution to Umbraco-CMS! 🎉 

I've added a shiny new contributors badge to your Our profile: https://our.umbraco.com/members/id:74599/ 🎖️ 

Thanks again!

jattwood:
Awesome! Thank you.

PR #10799, opened 2021-08-05
Clean-up and extend MigrationBase

ronaldbarendse:
### Prerequisites

### Description
This PR cleans up the `MigrationBase` class, as I noticed it had some unclear methods:
- `AddColumnIfNotExists<T>()` did an unnecessary check, as `AddColumn<T>()` already checks whether the column exists (by re-querying all tables/columns). I've obsoleted these and added new `AddColumn<T>()` overloads that accept a list of existing columns (so this can be queried once manually when adding multiple columns);
- The `ColumnExists()` method now also accepts this list of existing columns to prevent the above re-querying;
- The repeated `DefinitionFactory.GetTableDefinition()` calls are removed and the table name is automatically set based on the table definition when `tableName: null` is used;
- I've added XML documentation to all methods, as it wasn't clear to me what the overloads with the `out IEnumerable<string> sqls` parameter were used for (when this overload is used, the column is added as nullable, so a default value can be set and after that the outputted SQL can be executed to make the column non-nullable again).
- These methods were already split into a different file (using partial classes): I've moved the `AddColumn`, `ReplaceColumn` and `AlterColumn` overloads into separate files.
---------------

#### New overloads to simplify setting a default value for new non-nullable columns
`bool AddColumn<T>(string columnName, NPoco.Sql updateSql, string tableName = null, IEnumerable<ColumnInfo> columns = null)`
Instead of using the overload with the `out IEnumerable<string> sqls` parameter, running a query to set default values and making sure these SQL commands are always executed, I've added an overload that accepts an SQL query to set the default value. This has the added benefit of not running this query when the column already exists. The return value indicates whether the column is added and I've added `tableName` and `columns` as optional parameters, so it's clear these are not required.

`bool AddColumn<T>(string columnName, Func<T, T> updateAction, IEnumerable<ColumnInfo> columns = null)`
Similar to the above overload, this fetches all existing rows from the database and executes an action for every row, so more complex logic can be used to set a default value.

---------------

To properly test this PR, it's easiest to use the database of an Umbraco 7 installation and upgrade it: this should not return any SQL errors and end up in the same state as before this PR.


ronaldbarendse:
For testing the new `AddColumn<T>()` overloads, the following code adds a custom migration plan with a single migration that has 3 test cases (switch by commenting out the other ones):
1. This should result in a `BootException` because there's no default value
2. This should add the column with the same value for all rows
3. This sets a different value based on existing data.

Besides checking the column values, also ensure it's not nullable in the end. To re-run the migration, execute following SQL queries after a test:
```sql
DROP TABLE MigrationTests
DELETE FROM umbracoKeyValue WHERE [key] = 'Umbraco.Core.Upgrader.State+MigrationTests'
```
----------
```c#
using NPoco;
using Umbraco.Core.Composing;
using Umbraco.Core.Logging;
using Umbraco.Core.Migrations;
using Umbraco.Core.Migrations.Upgrade;
using Umbraco.Core.Persistence.DatabaseAnnotations;
using Umbraco.Core.Scoping;
using Umbraco.Core.Services;

public class MigrationTestsComposer : ComponentComposer<MigrationTestsComponent>
{ }

public class MigrationTestsComponent : IComponent
{
    private IScopeProvider _scopeProvider;
    private IMigrationBuilder _migrationBuilder;
    private IKeyValueService _keyValueService;
    private ILogger _logger;

    public MigrationTestsComponent(IScopeProvider scopeProvider, IMigrationBuilder migrationBuilder, IKeyValueService keyValueService, ILogger logger)
    {
        _scopeProvider = scopeProvider;
        _migrationBuilder = migrationBuilder;
        _keyValueService = keyValueService;
        _logger = logger;
    }

    public void Initialize()
    {
        var migrationPlan = new MigrationPlan("MigrationTests")
            .From(string.Empty)
            .To<MigrationTests>("test");

        new Upgrader(migrationPlan)
            .Execute(_scopeProvider, _migrationBuilder, _keyValueService, _logger);
    }

    public void Terminate()
    { }
}

public class MigrationTests : MigrationBase
{
    public MigrationTests(IMigrationContext context)
        : base(context)
    { }

    public override void Migrate()
    {
        Create.Table<MigrationTestsDto>().Do();
        Insert.IntoTable("MigrationTests").Row(new { name = "Test" }).Do();
        Insert.IntoTable("MigrationTests").Row(new { name = "Testing something else" }).Do();

        // 1: this fails, as it doesn't set a default value
        //AddColumn<MigrationTestsExtendedDto>("sortOrder");

        // 2: this sets the sortOrder to 0 for all rows
        //AddColumn<MigrationTestsExtendedDto>("sortOrder", Sql("UPDATE [MigrationTests] SET [sortOrder] = @0", 0));

        // 3: this conditionally sets the sortOrder based on existing values
        AddColumn<MigrationTestsExtendedDto>("sortOrder", x =>
        {
            if (x.Name == "Test")
            {
                x.SortOrder = -1;
            }
            else
            {
                x.SortOrder = 0;
            }

            return x;
        });
    }

    [TableName("MigrationTests")]
    [PrimaryKey("id")]
    [ExplicitColumns]
    private class MigrationTestsDto
    {
        [Column("id")]
        [PrimaryKeyColumn]
        public int Id { get; set; }

        [Column("name")]
        public string Name { get; set; }
    }

    [TableName("MigrationTests")]
    [PrimaryKey("id")]
    [ExplicitColumns]
    private class MigrationTestsExtendedDto
    {
        [Column("id")]
        [PrimaryKeyColumn]
        public int Id { get; set; }

        [Column("name")]
        public string Name { get; set; }

        [Column("sortOrder")]
        public int SortOrder { get; set; }
    }
}
```

PR #10800, opened 2021-08-05
Fixes various issues with javascript services

bjarnef:
### Prerequisites

### Description
This fixes some issues after recent PRs has been merged, e.g. `closeBackdrop` in `navigationService` which wasn't a public function. However instead of exposing `closeBackdrop` in `navigationService` I think it makes more sense the call `navigationService.hideMenu()` which also close backdrop. In fact I think `backdropService`should be responsible for logic regarding backdrops.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/tree/umbtree.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/tree/umbtree.directive.js
@@ -358,7 +358,7 @@ function umbTreeDirective($q, treeService, navigationService, notificationsServi
             */
             $scope.select = function (n, ev) {
 
-                navigationService.closeBackdrop();
+                navigationService.hideMenu();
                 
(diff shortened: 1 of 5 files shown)


bjarnef:
@nul800sebastiaan this fixes part of the the issue, but it seems gulp doesn't compile all changes, e.g. the `isArray(obj)` changes in this PR.
Furthermore with latest changes in `v8/contrib` branch and these changes, the seach overlay looks strange.

image

bjarnef:
Maybe it just took some time to compile the JS or something was cached (although I refreshed and cleared browser cache).
Anyway now I don't get the console error:

image

image

nul800sebastiaan:
Aha! Now I understand what you meant with the array check! I amended it to also update the null check like you suggested yesterday - 0404670ada3bd6a75dfb6218bcf052f0c8ffba64

nul800sebastiaan:
Thanks for the quick assist @bjarnef, you're a star! ⭐

PR #10802, opened 2021-08-06
Supporting multiple databases

AndyButland:
This PR is a spike coming out of this cycle's "hack time" - it may or may not go anywhere in future.

The idea behind it is that for some data, it can be useful to be able to store it in a separate database from the core Umbraco content.  Members is an example, another is the tables for the Umbraco Forms package, and it could be other package developers would find this useful.  Sometimes databases get copied around - e.g. a restore from production into a test environment - and ideally we don't want to copy PII data when we do this.  If such data could be configured to be stored in a separate database, this could be done more safely to get the content, without having to worry about deleting or anonymising PII.

It seemed a little tricky as Umbraco has for years been built on the assumption that there's one database, but I found I could add some seams in a couple of places:

- For a `Scope` - so when created, it's based on a particular database's connection string.
- For a `MigrationPlan` - so a set of migrations created within a plan can all run on a configured database.

Configuration-wise it works like this - where you have the default connection string as normal, plus multiple named connection strings, each identified with a unique alias.

```
  "ConnectionStrings": {
    "umbracoDbDSN": "Server=.\\SQLEXPRESS;Database=TestDbForCms;Integrated Security=true",
    "NamedConnectionStrings": [
      {
        "Alias": "Forms",
        "ConnectionString": "Server=.\\SQLEXPRESS;Database=TestDbForForms;Integrated Security=true"
      }
    ]
  },
```

In start-up we then register an `UmbracoDatabaseFactory` for each connection string.

And then that can be used to ensure the correct database is instantiated within a scope or a migration.

There's some test code in here that can be used for verification - e.g. `Test`, `TestService`, `TestRepository`, `TestPlan` etc. that will be removed later.

Further to consider:

- Nested scopes.  Currently one scope created within another becomes part of the parent scopes transaction, and is only committed when the parent scope completes.  I've had to change that in the case where the parent and child scopes are for different databases (which happens for example with a migration, with the parent scope managing the updates of the `umbracoKeyValue` table and the child scope for the migration potentially now being for a different database).  So in this PR, when the database's are different, the child scope's transaction completes when it completes, not when it's parent does.  Perhaps a transaction across both is possible - from what I read it is fairly easily if they are in the same server instance - but as we are only referencing the databases via a connection string, we don't know for sure that's the case.

### DIFF
--- a/src/Umbraco.Core/Configuration/Models/ConnectionStrings.cs
+++ b/src/Umbraco.Core/Configuration/Models/ConnectionStrings.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Umbraco.
 // See LICENSE for more details.
 
+using System.Collections.Generic;
+
 namespace Umbraco.Cms.Core.Configuration.Models
 {
     /// <summary>
@@ -27,5 +29,7 @@ namespace Umbraco.Cms.Core.Configuration.Models
         /// Gets or sets a value for the Umbraco database connection string..
         /// </summary>
         public ConfigConnectionString UmbracoConnectionString { get; set; }
+
+        public List<NamedConnectionString> NamedConnectionStrings { get; set; } = new List<NamedConnectionString>();
     }
 }
--- a/src/Umbraco.Core/Configuration/Models/NamedConnectionString.cs
+++ b/src/Umbraco.Core/Configuration/Models/NamedConnectionString.cs
@@ -1,0 +1,23 @@
+// Copyright (c) Umbraco.
+// See LICENSE for more details.
+
+namespace Umbraco.Cms.Core.Configuration.Models
+{
+    public class NamedConnectionString
+    {
+        /// <summary>
+        /// Gets or sets the named connection string's alias.
+        /// </summary>
+        public string Alias { get; set; }
+
+        /// <summary>
+        /// Gets or sets the named connection string value.
+        /// </summary>
+        public string ConnectionString { get; set; }
+
+        /// <summary>
+        /// Gets the connection string as a <see cref="Configuration.ConfigConnectionString"/>.
+        /// </summary>
+        public ConfigConnectionString ConfigConnectionString => new ConfigConnectionString(Alias, ConnectionString);
+    }
+}
--- a/src/Umbraco.Core/Constants-Conventions.cs
+++ b/src/Umbraco.Core/Constants-Conventions.cs
@@ -10,6 +10,7 @@ namespace Umbraco.Cms.Core
             public static class Migrations
             {
                 public const string UmbracoUpgradePlanName = "Umbraco.Core";
+                public const string TestPlanName = "Test";
                 public const string KeyValuePrefix = "Umbraco.Core.Upgrader.State+";
                 public const string UmbracoUpgradePlanKey = KeyValuePrefix + UmbracoUpgradePlanName;
             }
--- a/src/Umbraco.Core/Models/ITest.cs
+++ b/src/Umbraco.Core/Models/ITest.cs
@@ -1,0 +1,11 @@
+using System.Runtime.Serialization;
+using Umbraco.Cms.Core.Models.Entities;
+
+namespace Umbraco.Cms.Core.Models
+{
+    public interface ITest : IEntity, IRememberBeingDirty
+    {
+        [DataMember]
+        string Name { get; set; }
+    }
+}
--- a/src/Umbraco.Core/Models/Test.cs
+++ b/src/Umbraco.Core/Models/Test.cs
@@ -1,0 +1,25 @@
+using System;
+using System.Runtime.Serialization;
+using Umbraco.Cms.Core.Models.Entities;
+
+namespace Umbraco.Cms.Core.Models
+{
+    [Serializable]
+    [DataContract(IsReference = true)]
+    public class Test : EntityBase, ITest
+    {
+        private string _name;
+
+        public Test(int id, string name)
+        {
+            Id = id;
+            Name = name;
+        }
+
+        public string Name
+        {
+            get => _name;
+            set => SetPropertyValueAndDetectChanges(value, ref _name, nameof(Name));
+        }
+    }
+}
--- a/src/Umbraco.Core/Persistence/Constants-DatabaseSchema.cs
+++ b/src/Umbraco.Core/Persistence/Constants-DatabaseSchema.cs
@@ -54,6 +54,8 @@ namespace Umbraco.Cms.Core
                 public const string ExternalLogin = TableNamePrefix + "ExternalLogin";
                 public const string ExternalLoginToken = TableNamePrefix + "ExternalLoginToken";
 
+                public const string Test = TableNamePrefix + "Test";
+
                 public const string Macro = /*TableNamePrefix*/ "cms" + "Macro";
                 public const string MacroProperty = /*TableNamePrefix*/ "cms" + "MacroProperty";
 
--- a/src/Umbraco.Core/Persistence/Repositories/ITestRepository.cs
+++ b/src/Umbraco.Core/Persistence/Repositories/ITestRepository.cs
@@ -1,0 +1,8 @@
+using Umbraco.Cms.Core.Models;
+
+namespace Umbraco.Cms.Core.Persistence.Repositories
+{
+    public interface ITestRepository : IReadWriteQueryRepository<int, ITest>
+    {
+    }
+}
--- a/src/Umbraco.Core/Services/ITestService.cs
+++ b/src/Umbraco.Core/Services/ITestService.cs
@@ -1,0 +1,10 @@
+using System.Collections.Generic;
+using Umbraco.Cms.Core.Models;
+
+namespace Umbraco.Cms.Core.Services
+{
+    public interface ITestService : IService
+    {
+        IEnumerable<ITest> GetAll();
+    }
+}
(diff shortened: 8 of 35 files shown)


ronaldbarendse:
The configuration can probably be simplified, as the keys are already aliases, so the example would become:
```json
"ConnectionStrings": {
  "umbracoDbDSN": "Server=.\\SQLEXPRESS;Database=TestDbForCms;Integrated Security=true",
  "Forms": "Server=.\\SQLEXPRESS;Database=TestDbForForms;Integrated Security=true"
}
```

Getting the value could then be done using `Configuration.GetConnectionString("Forms")` (see [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationextensions.getconnectionstring?view=dotnet-plat-ext-5.0#Microsoft_Extensions_Configuration_ConfigurationExtensions_GetConnectionString_Microsoft_Extensions_Configuration_IConfiguration_System_String_)) and this structure would make it possible to add/override settings from e.g. environment variables (and merge different `NamedConnectionStrings` items).

--------------
Although the Umbraco CMS connection string name is currently hardcoded (in `Constants.System.UmbracoConnectionName`), this could use the same configuration convention, so we end up with:

```json
{
  "ConnectionStrings": {
    "umbracoDbDSN": "Server=(local);Database=Umbraco9;Integrated Security=true",
    "Forms": "Server=(local);Database=Forms9;Integrated Security=true"
  },
  "Umbraco": {
    "CMS": {
      "ConnectionStringName": "umbracoDbDSN"
    },
    "Forms": {
      "ConnectionStringName": "Forms"
    }
  }
}
```

In case `Umbraco:CMS:ConnectionStringName` is empty (not set), it defaults back to the hardcoded `umbracoDbDSN` and `Umbraco:Forms:ConnectionStringName` can default back to the CMS connection string...

AndyButland:
Good points, I'll have a play when I come back to this.  However there might be a few difficulties in what you've suggested, in that whilst we know about `umbracoDbDSN` at compile time, we don't know about the others, so not sure how we'd create the strongly typed configuration model for that.  With the array of name/values, we can.

bielu:
@AndyButland why not to implement dictionary of connection strings and modifed Umbraco to use ConnectionStrings[UmbracoDbAlias] instead of UmbracoConnectionString?

AndyButland:
@bielu - only reason not to might be backward compatibility for existing configuration (if and when this feature is added).

bielu:
@AndyButland  backward compatibility  with what? v9 is one big breaking change, it is not released so it is last moment do to hat as it should be :)

AndyButland:
This isn't going to be in the first release of V9... it was just a "cycle hack" idea, so will need picking up and properly considering later.  So if it came in 9.X, we'd need to make sure we respect existing config.  Even if in 10.0, there still might be value in not breaking the existing config... but we could decide to make a change then as per your suggestion.

bielu:
@AndyButland I really think it should be part of first release, as later we will never get rid off of that weird idea each connection string has to be as strongtype property :) and it is fairly easy to do that now, not later #breakingChanges

AndyButland:
Perhaps... but I don't think my colleagues would thank me for such a change just now!  Particularly for a feature that's just an idea, and not on any road map yet.

bielu:
@AndyButland but developers would thank you as it would resolve a lot issues with Security of storing form data on same db as Umbraco etc. it is actually one of few features which should be considered much earlier 😱

mzajkowski:
Great stuff @AndyButland and very valuable input @bielu!

I'd suggest inspiring on e.g. how Squidex has done it (https://github.com/Squidex/squidex/blob/master/backend/src/Squidex/appsettings.json). We've had today an internal discussion with @aochmann who was working with some multi-domain .NET Core projects and this was a similar way where domain-level connection strings make the most sense.

In this case, it could be:

```
{
    "umbraco": {
        "connectionString": ""
    },
    "forms": {
        "connectionString": ""
    },
    "members": {
        "connectionString": ""
    }
}
```

And so on. Until service/configuration will know from where to read it + if there would be a fallback to a standard ("old") name if needed - it shouldn't be a problem to maintain it (in theory of course).

Mixing name of the well-known property such as "connectionString" with anything else as a prefix or suffix is making it less readable in my opinion.

ronaldbarendse:
I agree with @bielu that the strongly typed `ConnectionStrings` class (that binds to the identically named configuration section) with the `UmbracoConnectionString` property (that is populated with the `umbracoDbDSN` value) makes it very hard to extend.

However, we also need to take the following into account:
1. This allows injecting `IOptions<ConnectionStrings>` to get the connection string without relying on the [`GetConnectionString()`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationextensions.getconnectionstring?view=dotnet-plat-ext-5.0&viewFallbackFrom=net-5.0) extension method that requires the global `IConfiguration` and uses 'magic strings';
2. The `UmbracoConnectionString` also contains parsing code to replace the `|DataDirectory|` token for LocalDB connection strings and to determine the provider name (to differentiate between SQL CE and SQL Server).

We could improve this by using a [named options approach](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-5.0#named-options-support-using-iconfigurenamedoptions), so we can have a simple POCO to represent the connection string and provider name:
```c#
public class ConnectionStringSettings
{
    public string ConnectionString { get; set; }
    public string ProviderName { get; set; }
}
```

We then set the default options name to `umbracoDbDSN`, so if you inject a `IOptions<ConnectionStringSettings>`, you'd get the default Umbraco one. If you want a different one, you'd use the `IOptionsSnapshot<ConnectionStringSettings>.Get()` or `IOptionsMonitor<ConnectionStringSettings>.Get()` methods (preferably with a string constant).

The `|DataDirectory|` token replacement can be done in a single `ConfigureAll()` call and the provider name can be parsed similarly. There's even a Microsoft convention that's used for environment variables, where special prefixes (e.g. `SQLAZURECONNSTR_`) adds a provider configuration entry with the provider name: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-5.0#connection-string-prefixes-1. If we can automatically map this value, we might even be able to remove the provider name parsing.

Another benefit would be to remove the tight coupling of SQL CE and LocalDB, as the configuration can be setup (including the parsing/replacement) in separate projects.

aochmann:
Idea: Support for both `ConnectionStrings` and domain.

Workflow could be:
 * if domain `ConnectionString` is null / empty try to use from `ConnectionStrings`.

I also like idea of override `ConnectionStrings` aliases: https://github.com/umbraco/Umbraco-CMS/pull/10802#issuecomment-894087275

In this case the structure could be fallowing based on domain connection strings:

```json
{
    "ConnectionStrings":{
        "umbracoDbDSN": "", //optional, used as fallback
        "Forms": "", //optional, used as fallback

    },
    "Umbraco": {
        "CMS": {
            "ConnectionStringName": "umbracoDbDSN", //override alias
            "ConnectionString": "", //optional
        },
        "Forms": {
            "ConnectionStringName": "Forms",
            "ConnectionString": "",
        }
    },
    "Mongo": {

    }
}
```

We can fail if domain connection string and `ConnectionStrings` are not set. We need to assign only one of them.

The `ConnectionStrings` is good for getting connection string by name with extension method: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationextensions.getconnectionstring?view=dotnet-plat-ext-5.0

The domain will require additional registration logic:

```csharp
var umbracoCmsSection = context.GetService<IConfiguration>()
                .GetSection("Umbraco")
                .GetSection("CMS");

var connectionString = umbracoCmsSection.GetValue<string>("ConnectionString");
```

But give more flexible structure for other configurations.

IMO we could do both ways `ConnectionStrings` and domain connection string, but for simplicity IMO it would be better to keep it simple at domain level:

```json
{
    "Umbraco": {
        "CMS": {
            "ConnectionString": "",
        },
        "Forms": {
            "ConnectionString": "",
        }
    },
    "Mongo": {

    }
}
```

ronaldbarendse:
@aochmann We should keep all connection strings in a single place in the `ConnectionStrings` section, so the `GetConnectionStrings()` extension method can always be used (e.g. by external packages, without a dependency on Umbraco). The connection string names should then be used to get the right one and these should be configurable, but use `Constants.System.UmbracoConnectionName` as fallback/default.

When combined with the environment variable/provider name convention (see my previous comment), we should end up with the following configuration:

```json
{
    "ConnectionStrings": {
        "umbracoDbDSN": "Data Source=(localdb)\\MSSQLLocalDB;AttachDbFilename=|DataDirectory|\\Umbraco.mdf;Integrated Security=True",
        "umbracoDbDSN_ProviderName": "System.Data.SqlClient",
        "Forms": "Data Source=|DataDirectory|\\Umbraco.sdf;Flush Interval=1",
        "Forms_ProviderName": "System.Data.SqlServerCe.4.0"
    },
    "Umbraco": {
        "CMS": {
            "ConnectionStringName": "umbracoDbDSN"
        },
        "Forms": {
            "ConnectionStringName": "Forms"
        }
    }
}
```

iOvergaard:
@AndyButland @ronaldbarendse Should we keep this alive or close?

AndyButland:
Not sure... it's very out of date and was only a proof of concept in the first place.  Maybe though it's worth you taking it in the team just to see if a) the idea and perhaps b) some of what I've done here, is useful?  And if there's interest in taking it forward?  I do think there's value in being able to split of say members, or form submissions, into a separate database.

PR #10803, opened 2021-08-06
V9: Add basic auth middleware

bergmania:
## Notes
Adds a middleware, that allows for prompting for basic auth on the website part if:
- The basic auth is turned on
- The IP is not allow listed
- No valid backoffice cookie exists

## Details
- Added new settings under `Umbraco:CMS:BasicAuth`
- Added service that caches the settings, and updates the cache on changes
- Added middleware that executes as the first middleware (by default)

## Tests
- Ensure you are not prompted when disabled (Default)
- Turn on basic auth in config `Umbraco:CMS:BasicAuth:Enabled = true`
- Go to the website when not logged in in the backoffice.
   - You will now be prompted using basic auth
   - Type backoffice credentials
 - Ensure you are not prompted again on future requests.
 
**Note that, browsers cache the basic auth credentials, so if you login using basic auth, and go to backoffice and logout, you will automatically be signed in on next website request.**


src/Umbraco.Core/Configuration/Models/BasicAuthSettings.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Umbraco.
+// See LICENSE for more details.
+
+using System.ComponentModel;
+using System.Net;
+
+namespace Umbraco.Cms.Core.Configuration.Models
+{
+    /// <summary>
+    /// Typed configuration options for security settings.
```
AndyButland, line 10:
```suggestion
    /// Typed configuration options for basic authentication settings.
```

src/Umbraco.Web.BackOffice/Extensions/UmbracoApplicationBuilder.BackOffice.cs
```diff
@@ -30,6 +31,7 @@ public static IUmbracoMiddlewareBuilder WithBackOffice(this IUmbracoMiddlewareBu
                 a => a.UseMiddleware<KeepAliveMiddleware>());
 
             builder.AppBuilder.UseMiddleware<BackOfficeExternalLoginProviderErrorMiddleware>();
+            builder.AppBuilder.UseMiddleware<BasicAuthAuthenticationMiddleware>();
```
AndyButland, line 34: Should this be in `WithBackOffice`?  Really it's running to protected public access to the front-end websites, so does it make more sense in `WithWebsite()`?

src/Umbraco.Web.BackOffice/Middleware/BasicAuthAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,101 @@
...
+using Umbraco.Cms.Core.Services;
+using Umbraco.Cms.Web.BackOffice.Security;
+using Umbraco.Extensions;
+
+namespace Umbraco.Cms.Web.Common.Middleware
+{
+    /// <summary>
+    ///     Ensures that preview pages (front-end routed) are authenticated with the back office identity appended to the
+    ///     principal alongside any default authentication that takes place
+    /// </summary>
```
AndyButland, line 18:
```suggestion
    /// <summary>
    ///     Provides basic authentication via back-office credentials for public website access if configured for use and the client IP is not allow listed.
    /// </summary>
```

src/Umbraco.Web.BackOffice/Middleware/BasicAuthAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// <inheritdoc />
+        public async Task InvokeAsync(HttpContext context, RequestDelegate next)
+        {
+            if (_runtimeState.Level < RuntimeLevel.Run || !_basicAuthService.IsBasicAuthEnabled())
+            {
+                await next(context);
+                return;
+            }
+
+            var clientIPAddress = context.Connection.RemoteIpAddress;
```
AndyButland, line 41: Wondering if you need to jump through more hoops here to ensure you get the client's IP when accessing through a CDN?  E.g. something like [this](https://stackoverflow.com/a/66936819/489433), considering anything else that might be custom for common CDN providers.

src/Umbraco.Web.BackOffice/Middleware/BasicAuthAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,101 @@
...
+                SetUnauthorizedHeader(context);
+            }
+        }
+
+
+
+        private static void SetUnauthorizedHeader(HttpContext context)
+        {
+            context.Response.StatusCode = 401;
+            context.Response.Headers.Add("WWW-Authenticate", "Basic realm=\"Umbraco as a Service login\"");
```
AndyButland, line 98: We aren't necessarily on Umbraco as a Service here.

src/Umbraco.Web.BackOffice/Middleware/BasicAuthAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,101 @@
...
+                await next(context);
+                return;
+            }
+
+
+            string authHeader = context.Request.Headers["Authorization"];
+            if (authHeader != null && authHeader.StartsWith("Basic"))
+            {
+                //Extract credentials
+                var encodedUsernamePassword = authHeader.Substring(6).Trim();
```
AndyButland, line 60: Could move this out into a helper method, or perhaps an extension method on `HttpRequest` called `.TryGetBasicAuthCredentials(out string username, out string password)`, and then can unit test.

src/Umbraco.Web.BackOffice/Middleware/BasicAuthAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,101 @@
...
+                var encoding = Encoding.UTF8;
+                var usernamePassword = encoding.GetString(Convert.FromBase64String(encodedUsernamePassword));
+
+                var seperatorIndex = usernamePassword.IndexOf(':');
+
+                var username = usernamePassword.Substring(0, seperatorIndex);
+                var password = usernamePassword.Substring(seperatorIndex + 1);
+
+
+                IBackOfficeSignInManager backOfficeSignInManager =
```
AndyButland, line 70: Given `BasicAuthAuthenticationMiddleware` is added scoped, could you inject this?

AndyButland (commented):
Looks good - have added a few comments inline and will now test it out.

src/Umbraco.Web.BackOffice/Middleware/BasicAuthAuthenticationMiddleware.cs
bergmania: Good spot.. This was a test, but I still could not.. Ill change the middleware to be singleton

src/Umbraco.Web.BackOffice/Middleware/BasicAuthAuthenticationMiddleware.cs
bergmania: I don't think this is necessary. if the setup is correct.. Back in the days you added another middleware that handled x-forwarded-for etc.

Now it looks even more simple: https://devblogs.microsoft.com/aspnet/forwarded-headers-middleware-updates-in-net-core-3-0-preview-6/#configuration-only-wire-up-in-preview-6

src/Umbraco.Web.BackOffice/Extensions/UmbracoApplicationBuilder.BackOffice.cs
bergmania: make sense.. but requires we move the IBackOfficeSignInManager to the common project

AndyButland: That's seems OK to me if it's just the interface.  Raises another thought though in that if there's a long-term aim to be able to run Umbraco just with `WithWebsite()` - for front-end only installations - you might also want to change this:

```
    IBackOfficeSignInManager backOfficeSignInManager =
        context.RequestServices.GetRequiredService<IBackOfficeSignInManager>();
```

To `GetService` and handle if it's not found.

AndyButland:
Just pushed a minor update to add a unit test for the helper method you've introduced.

I've unresolved one comment above, just as not sure if you saw my update and may want to do something related to it.

In testing all works as described, but I've noticed this middleware kicks-in on back-office requests too - e.g. to `/umbraco/` - in discussion with Morten yesterday think we understood that back-office requests should go through without this and allow you to login in the usual way... but I may have misremembered that.

AndyButland (commented):
All looks good now and works as expected.

AndyButland (approved):
All looks good now and works as expected.

PR #10810, opened 2021-08-06
Adding [DebuggerDisplay] attribute to core published content classes …

DanDiplo:
…to make debugging easier

### Prerequisites

### Description

This adds the `[DebuggerDisplay]` attribute to core published content classes to make debugging content easier. You can read about the [attribute here](https://docs.microsoft.com/en-us/visualstudio/debugger/using-the-debuggerdisplay-attribute?view=vs-2019).

To illustrate this, imagine you want to view the raw data for a particular property on a class that derives from `PublishedContentModel`. I want to inspect the property with the alias "siteName" to check the raw content, but how do I know which property it is? If you try and view the properties in the debugger currently you will just see this:

Properties_Before

Not very helpful. Finding out which of the dozen properties is the one you want to inspect is impossible. But with this change it will now look like this:

Properties_After

You can now easily identify the property by it's alias and the type of editor used to create the property.

Another example. Viewing a content model currently in the debugger would show you this:

Home_Before

But with this change you get more useful info at a glance:

Home_After

I've done this for a number of the core base property classes, so most of the derived classes will benefit.

### DIFF
--- a/src/Umbraco.Core/Models/PublishedContent/PublishedContentType.cs
+++ b/src/Umbraco.Core/Models/PublishedContent/PublishedContentType.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 
 namespace Umbraco.Core.Models.PublishedContent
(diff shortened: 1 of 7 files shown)


Zeegaan (approved):
Tested this, and it works great, makes things alot more clear in the debugger 👍 

Thanks alot for contributing! 💪

Issue #9289, opened 2020-10-27
Only possible to use custom listview if name match alias


bjarnef:
It seems when creating a custom listview layout from the document type it will by default create an datatype instance "List View - {contentTypeAlias}", in this case for a Employee Archive document type with alias "employeeArchive" the datatype instance created will be "List View - employeeArchive".

![image](https://user-images.githubusercontent.com/2919859/97341788-cb921e80-1885-11eb-8777-8c67f953a78f.png)

This also works when having a custom listview layout.

![image](https://user-images.githubusercontent.com/2919859/97342406-79053200-1886-11eb-9400-58fe4762c82c.png)

However in first place I couldn't understand why the custom listview layout wasn't shown up in the layout selector. After a while I figured out it was because I renamed the datatype instance to "List View - Employees" and when checked the document type again it was using the "List View - Content" datatype instance.

![image](https://user-images.githubusercontent.com/2919859/97342673-d13c3400-1886-11eb-9a83-5026e5edca6a.png)

After renaming the datatype instance again back to "List View - employeeArchive" it seems to work again.

Why has this changed? It didn't use to behave like this. I think it should always be possible to rename datatype instance without loosing the relation to the content/media/member type.

Furthermore previous you could also create a listview instance first if you prefer and afterwards select this from the document type afterwards. This doesn't seem to be possible either like in this article where the settings icon opened a dialog to select an existing listview.

https://24days.in/umbraco-cms/2016/custom-list-view-layouts/

Tested on Umbraco v8.8.0 and v8.8.1

bjarnef:
@nielslyngsoe @nathanwoulfe do you recall if this PR https://github.com/umbraco/Umbraco-CMS/pull/6028 changed some of this behavior or it is related to other changes?

I haven't used custom listview layout much in recent projects, but I think it previous was possible to rename the datatype instance afterwards. https://24days.in/umbraco-cms/2016/custom-list-view-layouts/

At the moment it seems to have a "relation" to listview using "List View - {contentTypeAlias}", but it should probably have a reference to the datatype instance using an udi/guid.

nathanwoulfe:
@bjarnef @nielslyngsoe I don't think 6028 will have changed this - the function `const checkForCustomListView = () => scope.dataType.name === "List View - " + scope.modelAlias;` exists in the previous version, and from a quick look, this is returning the bool responsible for determining the buttons to display (ie Create/Remove custom list view). 

[...]

PR #10813, opened 2021-08-08
Check invariant name of custom listview

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9289

### Description
This PR doesn't fixes https://github.com/umbraco/Umbraco-CMS/issues/9289 but it may be good enough for now.
Currently it assume a custom listview is configurated if data type name matches "List View - {alias}" where alias by default is camel case, e.g. "List View - people"

It then cause issues if the data type was renamed to "List View - People" as it would then use default "List View - Content".

With this change it ignore casing, so the datatype can be renamed to "List View - People".

q2EA75D48G

Additionally it could also try create the datatype name in pascal casing in first place:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Editors/DataTypeController.cs#L179

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/umblistviewsettings.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/umblistviewsettings.directive.js
@@ -8,9 +8,9 @@
             scope.dataType = {};
             scope.customListViewCreated = false;
 
-            const listViewNamePrefix = "List View";
+            const listViewPrefix = "List View - ";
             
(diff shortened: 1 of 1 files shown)


nathanwoulfe (approved):
Thanks @bjarnef, looks good to me. Agree too re the invariantEquals method, maybe in the new UI we'll just be able to use module imports and do the whole `import { invariantEqual } from 'Utilities'`

bjarnef:
@nathanwoulfe yes, that would be useful. To have various utilities divided in various categories / files, e.g. methods regarding DOM manipulation, accessibility etc.

PR #10814, opened 2021-08-08
Password tip not updated correct on init

bjarnef:
### Prerequisites

### Description
PR https://github.com/umbraco/Umbraco-CMS/pull/9256 introduced a `<umb-password-tip>`. However it wasn't updated correct on init. For example when creating a new member Umbraco generate a minimum password length, by default 10 characters, but the password tip says 10 characters were "Minimum 10 character(s) to go!"

First when updating password and `keyup` was triggered the password tip was updated.

Furthermore is was unnecessary to render value in curly brackets `{{vm.passwordTip}}` when using `ng-bind-html="vm.passwordTip"`.

**Before**

maYoTrCUpZ

**After**

yu1RcMmJd1


nathanwoulfe (approved):
Thanks @bjarnef, fixes the issue nicely

nathanwoulfe:
However, a related issue - when selecting a value from the browser's autofill suggestions (ie Chrome's suggested strong password), the tooltip is not triggered because the $onChanges event is never raised. It's a browser issue, not an AngularJs or Umbraco-implementation issue, so can be a bit tricky to resolve. Doesn't need to block this fix, but probably needs consideration in the future.

PR #10815, opened 2021-08-08
UI improvement: Add option to select/deselect all filters in Log Viewer drop-down

jakob-bagterp:
### Prerequisites

Feature update per discussion: #10765

### Description
Add select/deselect all feature to the Log Viewer.

Animation

Screenshot 2021-08-08 214227

Screenshot 2021-08-08 214247

And it also supports localisations of the button labels (Danish included in this pull request):

Screenshot 2021-08-08 214329

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/logviewer/search.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/logviewer/search.controller.js
@@ -298,16 +298,11 @@
         }
 
         function updateAllLogLevelFilterCheckboxes(bool) {
-            for (var i = 0; i < vm.logLevels.length; i++) {
-                vm.logLevels[i].selected = bool;
-            }
+            vm.logLevels.forEach(logLevel => logLevel.selected = bool);
         }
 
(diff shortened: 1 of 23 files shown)


bjarnef:
It should probably add a single og two `<umb-dropdown-item>` as last item(s) and buttons inside here to clear and select all.

@mattbrailsford has added something similar in Vendr where a clear button is shown at bottom on one or more selection. Maybe he has some feedback as well?

jakob-bagterp:
> It should probably add a single og two `<umb-dropdown-item>` as last item(s) and buttons inside here to clear and select all.

Good point. The buttons are hereby wrapped in the `<umb-dropdown-item>` directive. It looks and behaves as before, and I agree that the HTML is more semantically correct now.

nathanwoulfe (approved):
Thanks @jakob-bagterp, this works really nicely. Extra bonus points for going to the effort of adding translations, that's greatly appreciated.

I made a couple of minor changes to use forEach and map in place of the for... loops, functionally all the same, just reduces LOC a little.

PR #10805, opened 2021-08-06
Remove requirement to specify icon class in umb-icon directive

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
This PR is a follow-up on this PR https://github.com/umbraco/Umbraco-CMS/pull/9255

Currently `<umb-icon>` use SVG icons, but also fallback to legacy font icons for now. To support both version is was required to specify icon name in class attribute as well otherwise it didn't work in case a font icon was used.

E.g. `<umb-icon icon="icon-document" class="icon-document"></umb-icon>`

This would be much nicer: 
`<umb-icon icon="icon-document"></umb-icon>` or 
`<umb-icon icon="icon-document" class="another-class"></umb-icon>`

I tried added icon name to `class="umb-icon {{icon}}` but it has some side effects since the directive use `transclude` and `replace`, so e.g. in editor actions menu which added `-hidden` class via `ngClass` it didn't work and then hidden error and success icons was suddently all visible.

We really only need this specific to the font/legacy icon, so instead we can wrap the `<i>` element and ensure this inherit the correct font , while not adding the pseudo classes on the parent element.

In future we might remove this part, so only SVG icons are supported (maybe in backoffce rebuild ~ v10).

## Test notes

- SVG icons are shown. Try with both core SVG icons and custom SVG icon in App_Plugins folder.
Example folder with `coffee-cup.svg`: [CustomIcons.zip](https://github.com/umbraco/Umbraco-CMS/files/6948111/CustomIcons.zip)

- Font icons are shown. Try with a font icon package like [Material Design Icon Pack](https://our.umbraco.com/packages/backoffice-extensions/material-design-icon-pack/)

- Colors work on SVG icon and font icons.

- When changing icon and/or color in icon picker in document/media/member type and submitting dialog, the icon + color is updated in editor header.

- Polling icon is spinning when enabled.

- Adding an external login provider the icon is shown correct using a font awesome icon, custom font icon or SVG icon.

PR #10822, opened 2021-08-10
Update styles on editor-header icon picker

nathanwoulfe:
### Prerequisites

### Description

Updates the border color and radius on the icon picker in the editor header, to match the styling of the name input. Also removes some inline styles and unnecessary markup, and update the back button to remove the font-awesome icon in favor of the same icon from the backoffice set.

Didn't look quite right having rounded corners beside the squared input, and the slightly different border color triggered my OCD.

Before:
before
After:
after

And the back button, before:
backbtn-before
After:
backbtn-after

To test, verify the header renders correctly on a content node, a content node accessed from a list view, a media item, a content type, a user and a media/content/blocklist picker. It is used in a stack more places (70+), but these will cover the variations.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/editor.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/editor.less
@@ -79,12 +79,16 @@
 }
 
 .umb-editor-header__back {
-    background: transparent;
-    border: 0;
+	font-size:20px;
+	display:flex;
     color: @gray-6;
-    margin: 0 0 1px 0;
+    margin-right: 15px;
     padding: 0;
 	transition: color 0.1s ease-in-out;
+
+	&:hover {
+		color:@gray-7;
+	}
 }
 
 .umb-editor-header__name-wrapper {
@@ -97,6 +101,7 @@
 }
 
 .umb-editor-header__name-and-description {
+	flex:1 1 auto;
     margin-right: 20px;
     .umb-panel-header-description {
     	padding: 0 10px;
--- a/src/Umbraco.Web.UI.Client/src/less/panel.less
+++ b/src/Umbraco.Web.UI.Client/src/less/panel.less
@@ -351,15 +351,13 @@
   cursor: pointer;
   font-size: 2rem;
   margin-right: 5px;
-  margin-top: -6px;
   height: 50px;
   display: flex;
   justify-content: center;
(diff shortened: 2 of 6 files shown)


bjarnef:
I noticed the back icon in this PR as well https://github.com/umbraco/Umbraco-CMS/pull/10805 but kept the Font Awesome back icon. I recall the mini listview is using the `icon-arrow-left` as back icon as well.

We could also use any custom SVG icon here if we want to change this later. For now Font Awesome is still used for e.g. logviewer polling icon and external login providers. In backoffice rebuild we may choose the remove font awesome and maybe support of font icons in general? Currently we have many hacks to ensure correct font, use correct line height etc, so in backoffice rebuild it would be worth considering if we still need font icons or just should stick with only SVG icons.

nathanwoulfe:
I did a search for any instances of `fa-`, only other I saw was the logviewer, which I'll look at removing too. The rounded arrow annoyed me, since the other arrows and chevrons in the menu are all nice and sharp. 

I think dropping FontAwesome would be, well, awesome. All the FA icons are available as SVG, so if anyone wanted to include these, they absolutely can, but it's really just extra weight that the CMS doesn't need.

bjarnef:
Yeah, although the polling icon use `fa-spin` as well, but we could probably just have a utility class to make CSS rotation of an icon. Maybe other CSS animations could be useful too later.

Currently there are these four references:
https://github.com/umbraco/Umbraco-CMS/search?l=HTML&q=fa

And this one using a core font icon, but font awesome class for animation:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/views/logviewer/search.controller.js#L100

It would be great if you have time to look at https://github.com/umbraco/Umbraco-CMS/pull/10805 as well as it would clean up many of the classes and remove pretty much all `<i>` elements representing icons.

nathanwoulfe:
@bjarnef, I'd forgotten about the Helveticons icon font - I'm going to clean up (aka remove) this from the mini-list-view and log spinner, but will deal with #10805 first, so there's no conflicts.

Not sure how we'd determine usages of the Helveticons font, given devs could be using it in packages/customisation. I guess if it's shipped as a resource, but not used by the CMS, that's a good enough compromise.

PR #10823, opened 2021-08-10
V9: Package export and import both xml and zip

bergmania:
###  Summary
Packages can now use both XML and Zip. Zip is only returned from the package creation if media is included, otherwise, just the XML is returned.

The automatic package migrations now support both an embedded XML or and embedded Zip. If both exist only the Zip is used.

I also fixed an issue with imports, where media was not installed if the parent was not changed.

To test this, do the following:
- Add folder to media..
- Add multiple images into the folder
- Create a package with the folder and include descendants
- Remove one of the items inside the folder + empty the recycle bin
- Install the package using auto package migration
- Add ZIP or XML into the project, and in the same folder as you create the code snippet
- Embed the ZIP or XML file in SLN/Project
    
 ```
 public class TestAutomaticPackageMigrationPlan : AutomaticPackageMigrationPlan
 {
     public TestAutomaticPackageMigrationPlan()
         : base("TestZipPackage")
     {
     }
 }
```
- Verify the deleted items are reinstalled.

Other tests:
- Embed a XML
- Embed a Zip
- Download package without media (expect XML)
- Download package with media (Expect zip)

### DIFF
--- a/src/Umbraco.Core/Packaging/PackageMigrationResource.cs
+++ b/src/Umbraco.Core/Packaging/PackageMigrationResource.cs
@@ -78,7 +78,7 @@ namespace Umbraco.Cms.Core.Packaging
 
             if (xml is not null)
             {
-                return stream.GetStreamHash();
+                return xml.ToString();
             }
 
             throw new IOException("Missing embedded files for planType: " + planType);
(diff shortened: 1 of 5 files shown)


warrenbuckley:
Code changes seem fine to me, off to do some battle testing...

warrenbuckley:
All works as expected and I was able to download just an XML file or ZIP file if I had media/files to include.
Additionally I was able to import both types of packages back into Umbraco and for it to generate the necessary pieces again

PR #10826, opened 2021-08-10
V9: Add unique constraints to cmsDictionary and cmsLanguageText tables

elit0451:
## Details

- Adding a unique index on the `key` (_dictionary name_) value on the **cmsDictionary** table;
- Similarly, a unique index constraint is added for the **cmsLanguageText** table on the `languageId` and `UniqueID` ~and `value`~ columns - otherwise, duplicate values can be created for the same dictionary item;
- Silently log the failure of that database migration step, so people don’t have upgrade issues if they had duplicates (incorrect data) - which prevents the indexes from being created;
- If they want the indexes, they first need to remove the duplicates from their database.

### Code changes
- Updating `DictionaryDto` and `LanguageTextDto` to have unique indexes and unique indexes across columns;
- Creating a new V9 migration, where we drop the existing indexes;
- When trying to add the unique constraints, we will first check if there are any duplicates in the data - as the approach for catching an exception when creating the index directly leads to transaction errors: `“This SqlTransaction has completed; It is no longer usable”`;
- Fix tests;

:exclamation: Note: Couldn’t call the shorthand method `AlterColumn<DictionaryDto>(DictionaryDto.TableName, “key”)` or `Alter.Table(DictionaryDto.TableName).AlterColumn(“key”).AsString().Unique().Do()` to re-create the indexes - The column was altered but no indexes were added (even when there were no duplicates).

---

## Test
- Install the backoffice  on `v9/dev` branch;

### Inserting data
- Navigate to the Translation section;
- Create an item under Dictionary (ex: _“Test”_) and add a corresponding value (ex: _“Test item”_);
- Create another one (ex: _“Test1"_ with value _“Another test item”_) - UI prevents you from creating an item with the same name;
- Make a change in the DB - update the record with a duplicate value:
  - Make a selection to check the `pk` value of the _“Test1"_ record and execute the following query:
```sql
  UPDATE [cmsDictionary]
  SET [key] = ‘Test’
  WHERE pk = 2;
```

This will result in incorrect data for the `cmsDictionary` table - 2 entries have the same value in the `key` column.

You can also add items under one of the “Test”s and make them with the same value in the DB.

</br>

Similar thing for `[cmsLanguageText]` table:
- Add a new language in the Settings section;
- Update the values for the new language on the 2 dictionary items - (ex: “Teste”);
- Update the value of `languageId` (check the `pk` of the record you want to update):
```sql
  UPDATE [cmsLanguageText]
  SET [languageId] = 1
  WHERE pk = 3;
```

You will end up having 2 records with the same `languageId` and `UniqueId` ~and `value`~.

### Executing the migration
- Checkout to the PR’s branch;
- Check the existing indexes of **cmsDictionary** table and **cmsLanguageText** table;
  - They should look like this:

> cmsDictionary

</br>

> cmsLanguageText

</br>

- Run the backoffice and start the migration;
- Make sure the migration succeeds and you can continue navigating through the backoffice;
- Check the errors in the Log Viewer (_UPDATE: The 'value' column is no longer part of the unique constraint_):

err

- Check that the 2 indexes were not added.

</br>

- Go back and delete the duplicate records;
- Run the following script in the database, to reset the migration id back to the current latest as per _v9/dev_:
```sql
update [umbracoKeyValue] set value = '{10F7BB61-C550-426B-830B-7F954F689CDF}'
where [key] = 'Umbraco.Core.Upgrader.State+Umbraco.Core'
```
- Follow the migration steps through the backoffice again;
- This time there should be no errors, as there shouldn't be any duplicates and the indexes should be added:

> cmsDictionary
cmsDictionary_After

</br>

> cmsLanguageText


nzdev:
Could we make fixing the invalid data non optional?

AndyButland:
I realise this was requested in the ticket, but looking at it now, I'm not sure `value` should be included in the `cmsLanguageText` unique index.  This would allow you to have two dictionary values for the same key and language, which would be wrong.  So this index should be across just the `languageId` and `unqiueId` fields.

And I think that's why the automatic migration can't automatically delete the duplicates - as say somehow an installation had two values for the same key and language, we couldn't know which is the correct one to keep.  We would need to actually fail the migration by throwing an exception if duplicates were found.  Which is possible... but perhaps not everyone would be happy digging into database tables and finding duplicates, when normally the Umbraco database isn't something you need to manually modify.  And presumably if in an installation duplicates exist, it isn't currently fatal to the running of the application, even if it's not ideal.

An idea for future though - with Umbraco Forms we had a similar thing to do, and there also added a health-check to check and report on such things, in case they are missed in the logs for the migration itself.  That'll be out of scope of this PR, but it could well be a useful future addition.

elit0451:
Thanks for the input @AndyButland 😊 I've updated the PR

nikolajlauridsen:
Overall looks good to me 👍 I did find one thing, however.

when the migration fails for the `cmsDictionary` we don't re-add the index, so the `IX_cmsDictionary_key` index will be gone entirely if the migration fails. One solution could be to check if there are any duplicates before we delete the index since we know that if there are duplicates we won't be able to create the new unique index.

 Before migration:
image

After migration: 
image

Granted once you fix up the duplicates, the new indexes will be created just fine, but I think it still might be better to not end up with a missing index.

Another thought I have is, how are people supposed to re-run the migration if the index creation fails? I know you can manually set the migration key, but I don't think that's something that's really viable for people to do? Are we okay with this just being left as is if there are duplicates? If that's the case I think it's even more important that the old index remains.

bergmania:
> Another thought I have is, how are people supposed to re-run the migration if the index creation fails? I know you can manually set the migration key, but I don't think that's something that's really viable for people to do? Are we okay with this just being left as is if there are duplicates? If that's the case I think it's even more important that the old index remains.

We already talked about a nice-to-have/up-for-grabs task for adding healthchecks that could fix this and add the index if missing.

nikolajlauridsen:
> We already talked about a nice-to-have/up-for-grabs task for adding healthchecks that could fix this and add the index if missing.

Ah yes, I see that now, sorry about that, I'll go ahead and merge this then 👍

bergmania:
I still think it is a good idea to add the old index again :)

elit0451:
Good find @nikolajlauridsen 👏 Will make a new PR 😉

PR #10829, opened 2021-08-10
V9: Infinite editor file picker

bjarnef:
### Prerequisites

### Description
As suggested in https://github.com/umbraco/Umbraco-CMS/pull/10794#issuecomment-895136210 we can re-use file picker and make it simpler to use.

I noticed we have both static files picker, but also the existing file picker used for custom CSS and custom view. I guess we could use static file picker as well to pick CSS and HTML?

image

Currently it pass in the filter function, but maybe later we want a different view than `treePicker` or some additional options, e.g. a property "allowedFileExtensions" which could be an array of file extensions.


Zeegaan:
Tested and it works great!

Code looks good to me 👍 

Thank you for your contribution 💪

PR #10832, opened 2021-08-11
Fixed issue with CachedPartial that cached an unmaterialized value

bergmania:
### Summary 
Fixed issue with CachedPartial that did only work in cache misses.

The reason was that the value cached was un-materialized. Fix is to materialize as HtmlString before we cache.

Also changed timeout parameter from `int` to timespan.

### Test notes
- Create a partial view`, eg `hej.cshtml`
- Use the partial view form a template: `@await Html.CachedPartialAsync("hej", null, TimeSpan.FromSeconds(20))`
- Load the page (without debug mode in config)


Zeegaan:
Tested it, and this solution fixes the issue! 

Code makes sense to me 👍

PR #10833, opened 2021-08-11
v9: Updated IUmbracoContextAccessor Interface and HybridUmbracoAccessorClass

Zeegaan:
Follow up fix for: 
https://github.com/umbraco/Umbraco-CMS/pull/10551
and
https://github.com/umbraco/Umbraco-CMS/pull/10789

# Notes
- Updated IUmbracoContextAccessor to have:
TryGetUmbracoContext(out IUmbracoContext umbracoContext
Clear()
Set(IUmbracoContext)
- Updated HybridUmbracoContext to implement updated interface 
- Updated dependencies

# How to test
- Same as in https://github.com/umbraco/Umbraco-CMS/pull/10789
- Make your own custom middleware
- Try to inject IUmbracoContextAccessor


bergmania:
There is a lot of code duplication with the following code


```
if (!_publishedSnapshotAccessor.TryGetPublishedSnapshot(out var publishedSnapshot))
  {
      throw new InvalidOperationException("Wasn't possible to a get a valid Snapshot");
  }
```

I suggest to make an extension method to handle this. Something like 
```
public static IPublishedSnapshot GetRequiredPublishedSnapshot(this IPublishedSnapshotAccessor publishedSnapshotAccessor)
{
  if (publishedSnapshotAccessor.TryGetPublishedSnapshot(out var publishedSnapshot))
  {
      return publishedSnapshot
  }
  throw new InvalidOperationException("Wasn't possible to a get a valid Snapshot");
}
```

And then replace all the usages with 
```
var publishedSnapshot = _publishedSnapshotAccessor.GetRequiredPublishedSnapshot()
```

T cases where we can handle when it is not there, should still just use `TryGetPublishedSnapshot`

src/Umbraco.Web.Common/UmbracoHelperAccessor.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class UmbracoHelperAccessor : IUmbracoHelperAccessor
+    {
+        private readonly IHttpContextAccessor _httpContextAccessor;
+
+        public UmbracoHelperAccessor(IHttpContextAccessor httpContextAccessor) => _httpContextAccessor = httpContextAccessor;
+
+        public bool TryGetUmbracoHelper(out UmbracoHelper umbracoHelper)
+        {
+            umbracoHelper = _httpContextAccessor.HttpContext.RequestServices.GetService<UmbracoHelper>();
```
Shazwazza, line 19: There is no null check here. This will fail with a `NullReferenceException` when running in a non request thread.

Zeegaan: I changed it to check for null 👍

elit0451: I think Shannon meant to do a null check on `_httpContextAccessor`.

src/Umbraco.Web.Common/Controllers/RenderController.cs
```diff
@@ -37,7 +38,17 @@ public RenderController(ILogger<RenderController> logger, ICompositeViewEngine c
...
-        protected IUmbracoContext UmbracoContext => _umbracoContextAccessor.UmbracoContext;
+        protected IUmbracoContext UmbracoContext
+        {
+            get
+            {
+                if (!_umbracoContextAccessor.TryGetUmbracoContext(out var umbracoContext))
+                {
+                    throw new InvalidOperationException("Wasn't able to get an UmbracoContext");
+                }
+                return umbracoContext;
```
elit0451, line 49: Here you can just return `_umbracoContextAccessor.GetRequiredUmbracoContext();` after adding the `using Umbraco.Extensions;`

This is basically a similar case to the code duplication comment that @bergmania made but for the umbracoContext, as there are many places where we are doing this check

src/Umbraco.Web.Common/Controllers/PluginController.cs
```diff
@@ -27,7 +27,17 @@ public abstract class PluginController : Controller, IDiscoverable
...
-        public virtual IUmbracoContext UmbracoContext => UmbracoContextAccessor.UmbracoContext;
+        public virtual IUmbracoContext UmbracoContext
+        {
+            get
+            {
+                if (!UmbracoContextAccessor.TryGetUmbracoContext(out var umbracoContext))
+                {
+                    throw new InvalidOperationException("Wasn't able to get an UmbracoContext");
+                }
+                return umbracoContext;
```
elit0451, line 38: Check my comment in **RenderController.cs**

src/Umbraco.Web.Common/Extensions/PublishedContentExtensions.cs
```diff
@@ -48,7 +48,13 @@ public static class PublishedContentExtensions
         /// a culture to that document.</para>
         /// </remarks>
         public static string GetCultureFromDomains(this IPublishedContent content, IUmbracoContextAccessor umbracoContextAccessor, ISiteDomainMapper siteDomainHelper, Uri current = null)
-            => DomainUtilities.GetCultureFromDomains(content.Id, content.Path, current, umbracoContextAccessor.UmbracoContext, siteDomainHelper);
+        {
+                if (!umbracoContextAccessor.TryGetUmbracoContext(out var umbracoContext))
+                {
+                    throw new InvalidOperationException("Wasn't able to get an UmbracoContext");
+                }
```
elit0451, line 55: Duplication

src/Umbraco.Web.Common/Extensions/PublishedContentExtensions.cs
```diff
@@ -69,8 +75,11 @@ public static IEnumerable<PublishedSearchResult> SearchDescendants(this IPublish
                 .Field(UmbracoExamineFieldNames.IndexPathFieldName, (content.Path + ",").MultipleCharacterWildcard())
                 .And()
                 .ManagedQuery(term);
-
-            return query.Execute().ToPublishedSearchResults(umbracoContextAccessor.UmbracoContext.Content);
+            if (!umbracoContextAccessor.TryGetUmbracoContext(out var umbracoContext))
+            {
+                throw new InvalidOperationException("Wasn't able to get an UmbracoContext");
+            }
```
elit0451, line 81: Duplication

src/Umbraco.Web.Common/Extensions/PublishedContentExtensions.cs
```diff
@@ -88,8 +97,12 @@ public static IEnumerable<PublishedSearchResult> SearchChildren(this IPublishedC
                 .Field("parentID", content.Id)
                 .And()
                 .ManagedQuery(term);
+            if (!umbracoContextAccessor.TryGetUmbracoContext(out var umbracoContext))
+            {
+                throw new InvalidOperationException("Wasn't able to get an UmbracoContext");
+            }
```
elit0451, line 103: Same (_there are more cases further down_)

src/Umbraco.Web.Common/UmbracoHelperAccessor.cs
```diff
@@ -0,0 +1,27 @@
...
+            if(umbracoHelper == null)
+            {
+                return false;
+            }
```
elit0451, line 23: We are already returning `false` if `umbracoHelper` is null ⬇️

src/Umbraco.Web.Website/Routing/UmbracoRouteValueTransformer.cs
```diff
@@ -94,11 +94,11 @@ public override async ValueTask<RouteValueDictionary> TransformAsync(HttpContext
             {
                 return values;
             }
-
             // will be null for any client side requests like JS, etc...
-            if (_umbracoContextAccessor.UmbracoContext == null)
+            if (!_umbracoContextAccessor.TryGetUmbracoContext(out var umbracoContext))
             {
                 return values;
+                ;
```
elit0451, line 101:
```suggestion
```

src/Umbraco.Web.Common/UmbracoHelperAccessor.cs
Zeegaan: Yea i must have misunderstood, fixing that right now

elit0451 (changes requested):
Good work 💪  left some inline comments and suggestions for you to have a look 🙂

elit0451 (approved):
Thanks for the fixes @Zeegaan ⭐ 

I've pushed some cleanup - it mainly included places where we still had duplicate code related to the cases where we couldn't get the `umbracoContext` - now we will be throwing the `InvalidOperationException("Wasn't able to get an UmbracoContext");` through the `_umbracoContextAccessor.GetRequiredUmbracoContext();`.

I am happy to merge after we see all the tests passing 😊

PR #10834, opened 2021-08-11
V9 Remove System.Configuration

warrenbuckley:
System.Configuration is used for XML based config which we are no longer using Umbraco V9

## Test Notes
* Add project reference to TestData into Web.UI project
* Enable it in appsettings.json
* Visit /LoadTest (Not 100% sure we still use this)
* Ensure it works as it did before
* Alternative way is to visit `/umbraco/surface/umbracotestdata/CreateTree?count=100&depth=5`

```json
"Umbraco": {
    "CMS": {
        "TestData": {
            "Enabled" : true,
        }
    }
}
```


src/Umbraco.TestData/LoadTestController.cs
```diff
@@ -44,7 +44,7 @@ public class LoadTestController : Controller
 <body>
   <div class=""head"">
   <h1><a href=""/LoadTest"">LoadTest</a></h1>
-  <div class=""ver"">" + System.Configuration.ConfigurationManager.AppSettings["umbracoConfigurationStatus"] + @"</div>
+  <div class=""ver"">@_umbracoVersion.SemanticVersion</div>
```
bergmania, line 47: Not sure if it is required, but guess you would like to use the `ToSemanticStringWithoutBuild` extension method.

Otherwise, it returns the build info you added in another PR, to make github source available IIRC.

bergmania (approved):

Issue #10820, opened 2021-08-09
RedirectToCurrentUmbracoPage not working with UrlProviderMode Absolute


konius:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0-rc001

### Bug summary

In SurfaceController if using `RedirectToCurrentUmbracoPage` with  `UrlProviderMode Absolute` throws an exception (attached).

![Screenshot 2021-08-09 190038](https://user-images.githubusercontent.com/3933562/128737165-715445a9-185e-42c0-8d47-eb7d0feae34a.jpg)

### Specifics

Setting the following in the apsettings.json:

```
"WebRouting": {
  "UrlProviderMode": "Absolute"
}
```

If the provider is `Relative` - works.

### Steps to reproduce

1. Create a test SurfaceController
2. Post to surface controller using `Html.BeginUmbracoForm<TestController>`
3. Return `RedirectToCurrentUmbracoPage`

### Expected result / actual result

Expected to be redirected to the current page.

Zeegaan:
Issue should now be fixed in https://github.com/umbraco/Umbraco-CMS/pull/10838 

Thank you for reporting this 👍

PR #10838, opened 2021-08-12
V9: Fix RedirectToCurrentUmbracoPage when UrlProviderMode is set to "Absolute"

elit0451:
Fixes: #10820

## Details
- `UrlProviderMode` can be set to "Absolute", "Auto" and "Relative";
- When the mode was set to "Absolute', `RedirectToCurrentUmbracoPage()` was not working;
- Adding a check for this case to avoid the exception being thrown - `$"The value {virtualPath} for parameter {nameof(virtualPath)} must start with ~/ or /"`.

## Test
- Follow the steps in the original issue OR
</br>

- Set ⬇️ to appsettings.json
```json
"WebRouting": {
  "UrlProviderMode": "Absolute"
}
```
- Create a test _SurfaceController_ and add a PostMethod like:
```c#
[HttpPost]
public IActionResult PostMethod()
{
    if (!ModelState.IsValid)
    {
        return CurrentUmbracoPage();
    }

    return RedirectToCurrentUmbracoPage();
}
```
- Post to the added surface controller:
  - Create a new Partial View Macro file;
  - Insert the following code:
```cshtml
@inherits Umbraco.Cms.Web.Common.Macros.PartialViewMacroPage

@using Umbraco.Cms.Web.Website.Controllers

@using (Html.BeginUmbracoForm<MyNewController>("PostMethod"))
{
    <button type="submit" class="btn btn-primary">Submit</button>
}
```
- Go to the newly created Macro and enable the **Use in rich text editor and the grid**;
- Create a Document type and add a Rich Text Editor as a property;
- Edit the node's template to render the RTE using `@Model.Value("[RTE alias]")`;
- In the Content section ( in the RTE property) insert the macro you created;
- Save and Publish;
- Navigate to the published link of that content node;
- Clicking the button shouldn't trigger the reported error.

</br>

---
❗ **Note**: The path for the `WebRouting` setting in apsettings.json is `Umbraco:Cms:WebRouting`.

**Bonus**: Removing the mistakenly committed connection string from appsettings.json 😉

### DIFF
--- a/src/Umbraco.Web.Common/AspNetCore/AspNetCoreHostingEnvironment.cs
+++ b/src/Umbraco.Web.Common/AspNetCore/AspNetCoreHostingEnvironment.cs
@@ -8,6 +8,7 @@ using Microsoft.Extensions.Options;
 using Umbraco.Cms.Core.Collections;
 using Umbraco.Cms.Core.Configuration;
 using Umbraco.Cms.Core.Configuration.Models;
+using Umbraco.Cms.Core.Models.PublishedContent;
 using Umbraco.Extensions;
 using IHostingEnvironment = Umbraco.Cms.Core.Hosting.IHostingEnvironment;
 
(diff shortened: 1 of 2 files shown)


Zeegaan:
Tested it, and this solution fixes the issue 👍 

Code makes sense to me!

Issue #10841, opened 2021-08-12
Grid editor headline gets encoded twice


Nikkelmann:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0-rc001

### Bug summary

When rendering a grid layout with Html.GetGridHtml() any headlines in the grid will be encoded twice.
This was fixed in https://github.com/umbraco/Umbraco-CMS/pull/9535#issue-536004808.

### Steps to reproduce

Setup a document type with a grid layout and add a content node with this document type (remember to setup the template with `Html.GetGridHtml()`).
Add a headline to the grid layout and type in characters that would get encoded (eg. æøå).
Save and publish, then visit the url.

### Expected result / actual result

Headline shows &#230;&#248;&#229; instead of æøå.

nul800sebastiaan:
I can reproduce. 

Indeed, looks like `markup = markup.Replace("#value#", Html.ReplaceLineBreaks(HttpUtility.HtmlEncode((string)Model.value.ToString())).ToString());` needs to be `markup = markup.Replace("#value#", Html.ReplaceLineBreaks((string)Model.value.ToString()).ToString());`. I'll whip up a PR.

Nikkelmann:
@nul800sebastiaan I've added a PR just now: https://github.com/umbraco/Umbraco-CMS/pull/10842

nul800sebastiaan:
Haha! Me too! I'll merge yours though! 👍

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/10842 - thanks again @Nikkelmann

PR #10842, opened 2021-08-12
Only encode text once

Nikkelmann:
Html.ReplaceLineBreaks calls WebUtility.HtmlEncode internally, so the text gets encoded by this method.

As far as I know, WebUtility.HtmlEncode and HttpUtility.HtmlEncode produces the same result (unlike WebUtility.UrlEncode and HttpUtility.UrlEncode which produces different results; see https://edi.wang/post/2018/11/25/netcore-webutility-urlencode-httputility-urlencode), so it should be safe to omit HttpUtility.HtmlEncode here.

### Prerequisites

If there's an existing issue for this PR then this fixes: #10841

### Description
The steps to reproduce the error is mentioned in the issue above.

### DIFF
--- a/src/Umbraco.Web.UI.NetCore/Views/Partials/grid/editors/textstring.cshtml
+++ b/src/Umbraco.Web.UI.NetCore/Views/Partials/grid/editors/textstring.cshtml
@@ -4,7 +4,7 @@
 @if (Model.editor.config.markup != null)
 {
     string markup = Model.editor.config.markup.ToString();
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Wonderful, thanks very much @Nikkelmann and congrats on your first merged PR for Umbraco-CMS! Do you have an account on [Our Umbraco](https://our.umbraco.com)? I would love to give you a contributor badge!

nul800sebastiaan:
Great, gave you a shiny new badge! 👍

PR #10845, opened 2021-08-12
Fixed incorrect node url being returned when no suitable domain exist…

### LINKED ISSUE #10827: Inconsistent url generation when some of the languages do not have domains set

9.0.0-rc001

### Bug summary

If I have a multi-lingual website, and I publish a node in multiple languages but only setup the domain for some of them, I get inconsistent behavior when trying to retrieve the url (both in the backoffice and by using `.Url(culture)`).

This is also reproducible in `8.15.1`.

Example:
Languages: `en-US` (main), `nl-NL`, `pt-PT`
Domains: Only `/nl` - `nl-NL` is set

image

Root node:
- Backoffice:
`nl-NL`: `/nl`
`en-US`: `/nl` (but accessible through `/`)
`pt-PT`: `This item is not published` (but if I do `.IsPublished(culture)` I get `true`)
- .Url(culture):
`nl-NL`: `/nl`
`en-US`: `/nl` (but accessible through `/`)
`pt-PT`: `/nl`

Child node:
- Backoffice
`nl-NL`: `/nl/nl-content`
`en-US`: `This document is published but its URL cannot be routed` (but accessible through `/content`)
`pt-PT`: `This item is not published` (but if I do `.IsPublished(culture)` I get `true`)
- .Url(culture):
`nl-NL`: `/nl/nl-content`
`en-US`: `/nl/nl-content` (but accessible through `/content`)
`pt-PT`: `/nl/nl-content`

### Specifics

_No response_

### Steps to reproduce

1. Setup the languages in the CMS as follows
image
2. Create a root node which varies by culture and setup the following domain.
image
3. Publish that node in all 3 languages and check the urls for all 3 (both in the backoffice and through code by using `node.Url(culture)`. You can also try to use `node.IsPublished(culture)` and check that some that have `This item is not published` showing in the backoffice, are actually considered published).
image
4. Add a child node, which also varies by culture, and also publish it in all the 3 languages and check the urls.

I would expect that if I only have an `nl-NL` domain setup, that my main language would have url `/` and not the same url as `nl-NL`, since it's possible to access it through `/`.
Regarding the rest of the languages, I'm not sure what the behavior should be but I think that showing `This item is not published` can be misleading and I would expect to see `This document is published but its URL cannot be routed` instead.
I would also expect to have a simple way to check if the node in a certain culture is routable or not through code.

vsilvar:
### Prerequisites

If there's an existing issue for this PR then this fixes #10827 

### Description
This changes the code to have a behavior close to what is expected while trying not to change to much.
Instead of always falling back to the first registered domain, we simply don't match one and use the default route.

This will more closely follow the default routing behavior and properly show it in the CMS as well.

### DIFF
--- a/src/Umbraco.Core/Routing/DomainUtilities.cs
+++ b/src/Umbraco.Core/Routing/DomainUtilities.cs
@@ -190,7 +190,7 @@ namespace Umbraco.Cms.Core.Routing
             }
 
             // look for domains that would be the base of the uri
-            var baseDomains = SelectByBase(considerForBaseDomains, uri);
+            var baseDomains = SelectByBase(considerForBaseDomains, uri, culture);
             if (baseDomains.Count > 0) // found, return
                 return baseDomains.First();
 
@@ -208,12 +208,15 @@ namespace Umbraco.Cms.Core.Routing
         private static bool IsBaseOf(DomainAndUri domain, Uri uri)
             => domain.Uri.EndPathWithSlash().IsBaseOf(uri);
 
-        private static IReadOnlyCollection<DomainAndUri> SelectByBase(IReadOnlyCollection<DomainAndUri> domainsAndUris, Uri uri)
+        private static bool MatchesCulture(DomainAndUri domain, string culture)
+            => culture == null || domain.Culture == culture;
+
+        private static IReadOnlyCollection<DomainAndUri> SelectByBase(IReadOnlyCollection<DomainAndUri> domainsAndUris, Uri uri, string culture)
         {
             // look for domains that would be the base of the uri
             // ie current is www.example.com/foo/bar, look for domain www.example.com
             var currentWithSlash = uri.EndPathWithSlash();
-            var baseDomains = domainsAndUris.Where(d => IsBaseOf(d, currentWithSlash)).ToList();
+            var baseDomains = domainsAndUris.Where(d => IsBaseOf(d, currentWithSlash) && MatchesCulture(d, culture)).ToList();
 
             // if none matches, try again without the port
             // ie current is www.example.com:1234/foo/bar, look for domain www.example.com
--- a/src/Umbraco.Core/Routing/DomainUtilities.cs
+++ b/src/Umbraco.Core/Routing/DomainUtilities.cs
@@ -199,10 +199,6 @@ namespace Umbraco.Cms.Core.Routing
             if (filter != null)
             {
                 var domainAndUri = filter(cultureDomains ?? domainsAndUris, uri, culture, defaultCulture);
-                // if still nothing, pick the first one?
-                // no: move that constraint to the filter, but check
-                if (domainAndUri == null)
-                    throw new InvalidOperationException("The filter returned null.");
                 return domainAndUri;
             }
 
--- a/src/Umbraco.Core/Routing/SiteDomainMapper.cs
+++ b/src/Umbraco.Core/Routing/SiteDomainMapper.cs
@@ -336,8 +336,7 @@ namespace Umbraco.Cms.Core.Routing
             if (qualifiedSites == null)
             {
                 return domainAndUris.FirstOrDefault(x => x.Culture.InvariantEquals(culture))
-                    ?? domainAndUris.FirstOrDefault(x => x.Culture.InvariantEquals(defaultCulture))
-                    ?? domainAndUris.First();
+                    ?? domainAndUris.FirstOrDefault(x => x.Culture.InvariantEquals(defaultCulture));
             }
 
             // find a site that contains the current authority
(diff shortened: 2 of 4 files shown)


nikolajlauridsen:
First of all, thanks a lot for another PR, love to see it 🎉 

But having done some testing now, I'm not really sure this is what we want to do, while it does make it better in some ways we end up breaking other things, and it still doesn't seem quite right, although it is better. 

This is gonna be a bit lengthy, but it's just as much to get some notes down for my self as well 😄 

### The test setup

Like in the PR I have three languages, with English as the default:
image

I have some document types that vary by culture.

Only one domain is assigned:

image


I add this to the template

```html
<br/>
<span>EN is published: @Model.IsPublished("en-us")</span>
<br/>
<span>DK is published: @Model.IsPublished("da")</span>
<br/>
<span>NO is published: @Model.IsPublished("nb")</span>
<br/>


<hr/>
<br/>
<span>EN URL: @Model.Url("en-us")</span>
<br/>
<span>DK URL: @Model.Url("da")</span>
<br/>
<span>NO: @Model.Url("nb")</span>
<br/>
```

### The results

Now before this PR I would see this in the backoffice when I publish:

* EN - `/dk/` -  but it's accessible through `/`
* DA - ` https://localhost:44331/dk/` - Which seems as expected.
* NB - `This item is not published`  - Even though it is published, it's just not routable

and in the rendered templates all cultures return true for `.IsPublished`, and `.Url` returns `/dk/` for all, which seems weird, but makes sense from the comment:

```
// no match means that either current does not belong to a site, or the site it belongs to
// does not contain any of domainAndUris. Yet we have to return something. here, it becomes
// a bit arbitrary.
```

So as I understand it if we can't find anything we more or less just return anything we can, because we must return something.

The problem changing this is that suddenly we don't return anything, so after the changes, if you try to access the `en` variation through `/` which you still can, you receive a server error because the filter returns null, when trying to get a URL for `@Model.Url("<someCulture>")`: `InvalidOperationException: The filter returned null.`

Another thing that is still weird is that, is that it now in the backoffice say: 

* EN - `Could not get the URL` -  but the content is published and accessible through `/`, but this is still okay, because the domain is not set, which it should be.
* DA - ` /dk/` and `https://localhost:44331/dk/` - As expected, all good
* NB - `This item is not published`  - Even though it is published, it's just not routable, so it should really say "The content is published, but cannot be routed".

So it seems to still be inconsistent but in a different way. I do think this is a step on the ways and kudos for investigating it, but I think we need to take a closer looks at how we can do this in a way where it becomes consistent and that `.Url` doesn't break, and what it requires.

vsilvar:
@nikolajlauridsen Thanks for taking the time to look at the PR. 
When cleaning up the PR I forgot to include an important part of the fix, causing the erroneous behavior. Sorry about that.

With my latest commit the behavior should now be:
EN - `/`
DA - `/dk/`
NB - This document is published but its URL cannot be routed (.Url returns `/` still)

I think this improves the current behavior but it's still not perfect due to the last case, which while completely correct in the CMS returns an incorrect Url with `.Url`

This last case could also be fixed by adding the following in `GetUrlFromRoute`:
```
...
var domainUri = ...
if (domainUri == null && !umbracoContext.PublishedSnapshot.Domains.DefaultCulture.InvariantEquals(culture))
{
    return null;
}
...
```
The downside being that the CMS message will default to the generic:
`This document is published but is not in the cache`

Please let me know if the behavior makes sense now, and whether the additional fix mentioned above is preferred or not.

nikolajlauridsen:
Just did some more testing, and the behaviour after your changes is a significant improvement compared to what we currently have. In regards to the behaviour of `.Url`, it's of course not optimal, however, I think it's fine as it is now since the backofffice message is correct, and this way nothing should spectacularly explode if you don't expect the null, so it's not all bad, and at the end of the day, you really shouldn't have unmapped domains 😄. 

I've tried to do as much testing as I can, and all seem to still work out fine, but just with more correct messages 👍 , of course, I can't possibly test every single edge case, so there might be some gremlins, however, I think that if we're going to change this behaviour, now is the time to do it since it's a new major, so I'll go ahead and merge this in.

Amazing work, thanks a lot for your contribution 🎉

vsilvar:
@nikolajlauridsen
Just created #11032 as this PR was already merged and I noticed I was doing a case sensitive comparison of the cultures.
It still worked, but it's better to keep it consistent to prevent unforeseen issues.

Shazwazza:
I'm a little worried about the changes here because the functionality is now different from v8. I'm porting over some tests for v8, specifically with nested domains and these changes now make that fail so I fear this has introduced new issues.

Preferably, we need to know if this issue existed in v8 and if so how to fix it there.

@bergmania I think this may cause other underlying issues. The test specifically in v8 that I'm porting over is called `UrlsWithNestedDomains.DoNotPolluteCache`. This now no longer works in v9. Specifically this logic:

```cs
// Set a current URL
const string url = "http://domain1.com/1001-1/1001-1-1";

// ... we then create an Umbraco context with this URL

// Then get the absolute URL for a node that exists under a node that has a domain assigned that is not the current domain
string absUrl = publishedUrlProvider.GetUrl(100111, UrlMode.Absolute);

// This will NOT pass because it's no longer respecting the domain.
Assert.AreEqual("http://domain2.com/1001-1-1/", absUrl);
```

Shazwazza:
Actually, this might be an issue with my branch not being merged with the very latest. I'll report back if I still have the issue.

Issue #10774, opened 2021-08-01
Adding content in multiple threads from composer cause IOException


bielu:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

v9-rc1

### Bug summary

When adding content in multiple thread I can see IOException as example
`System.IO.IOException: The process cannot access the file 'C:\path\Temp\UmbracoData\ac7553eab8aeeb152f9336b0f51241b90ebde15a\NuCache\NuCache.Content.db' because it is being used by another process.`
from that what I can see we run NotifyPublishedSnapshotService on each thread separatly what cause dead lock on file as both threads try to access nucache file

### Steps to reproduce

There is hard to add steps to reproduce, but basically create composer which will be adding content through background queue (as example: https://michaelscodingspot.com/c-job-queues/) and add for example about 20 nodes of content and 20 of media, it should cause above described issue.

### Expected result / actual result

Content is populated without breaking boot sequence and hitting IOException.

I guess one of solution would be make sure we are using one cache refresher queue to avoid multithread access, but I guess @Shazwazza will have better ideas how to deal with that. :) 

---
_This item has been added to our backlog [AB#13390](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13390)_

Shazwazza:
Can you please supply the code you are using to re-create this issue so that we can understand what code you are using that causes this? Also please supply the full stack trace. Thanks!

bielu:
I can't provide exact code which I am using but will prepare example over the weekend 😃

p-m-j:
This works consistently in v8

```csharp
    public class BugFixComposer : IUserComposer
    {
        public void Compose(Composition composition)
        {
            composition
                .Components()
                .Append(typeof(BugFixComponent));
        }
    }

    public class BugFixComponent : IComponent
    {
        private readonly IContentService _contentService;
        private readonly IContentTypeService _typeService;
        private readonly IScopeProvider _scopeProvider;
        private readonly IUmbracoContextFactory _contextFactory;

        private const string _bugFixContentType = "bugfixContentType";

        public BugFixComponent(
            IContentService contentService,
            IContentTypeService typeService,
            IScopeProvider scopeProvider,
            IUmbracoContextFactory contextFactory)
        {
            _contentService = contentService;
            _typeService = typeService;
            _scopeProvider = scopeProvider;
            _contextFactory = contextFactory;
[...]

PR #4553, opened 2019-02-13
Refactor post-migrations

zpqrtbnk:
With this PR:

Post-migrations as we knew them, that were magically discovered, are entirely gone. Most of them were obsolete anyways and the only one that is still required in v8 is `ClearCsrfCookies`.

They are replaced with... new post-migrations. These are standard `IMigration` implementations, but they are not registered in the migration plan. Instead:

* The plan has an `AddPostMigration<TMigration>()` method that can be used to pre-register a post-migration before it is executed. That's what we do for `ClearCsrfCookies`: when executing the plan in the upgrade step, we add the post-migration.
* The `IMigrationContext` has an `AddPostMigration<TMigration>()` method, and therefore each migration that executes can add post-migrations.
* After the plan has executed, the list is de-duplicated and post-migrations are executed.

For instance, there is a now a `RebuildPublishedSnapshot` migration (which is not part of the plan). If you create a migration that requires that the cache is rebuilt, you would do:

```
public override Migrate()
{
  // do whatever is needed

  // and then, add the post-migration
  Context.AddPostMigration<RebuildPublishedSnapshot>();
}
```

And this means that, after the plan has run correctly, but as part of the same db transaction, the `RebuildPublishedSnapshot` migration will run.

#### Discussion:

Post-migrations run in the order they have been inserted. In case of duplicates, only the first occurence executes. The `MigrationPlan` has a `PreparePostMigrations` that could be used to filter and/or reorder the list. But there is no plan for a clever ordering mechanism (weights or dependencies). Later, if required.

Thoughts?

#### Test:

Ensure that Umbraco still installs. Ensure that Umbraco still upgrades. Implement a migration that requires rebuilding the cache, and test.

PR #10847, opened 2021-08-12
Backport fix for #10774 to v8

p-m-j:


Shazwazza:
Looks good. I had a look at the long comment regarding `SetUmbracoVersionStep` and the code that used to be called no longer exists. The file deletion logic was then added in rev `d16bcdc8417249faaa879be22be8ab19e4d1fb7e` which was part of this PR https://github.com/umbraco/Umbraco-CMS/pull/4553

So the code in question that can run during an upgrade is `Umbraco.Core.Migrations.PostMigrations.RebuildPublishedSnapshot` which calls into `IPublishedSnapshotRebuilder.Rebuild`. This will rebuild the nucache db table and then issue a `_distributedCache.RefreshAllPublishedSnapshot();` which will rebuild the in-memory nucache caches (and write to disk).

The whole comment in question is about property value converters not being initialized during the upgrade time and therefore the nucache could be in an incorrect state - however, property value editors are always initialized at startup and they are discovered so this comment is incorrect and most likely due to old legacy behavior. 

I've just tested a new migration added that uses the `RebuildPublishedSnapshot` and everything works fine. I've removed the comment :)

nul800sebastiaan:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/10774

Issue #10323, opened 2021-05-25
Property value converter for Eye Dropper Color Picker


konius:
### Umbraco version

v8

### Description

Currently Eye Dropper Color Picker returns `object` in the ModelsBuilder and no meaningful context what the colour is. 

![Screenshot 2021-05-25 133933](https://user-images.githubusercontent.com/3933562/119496832-8fb42e80-bd6c-11eb-8320-d097659545d5.jpg)

Would be great to return a custom Model with a range of various properties to work with:

```
public class EyeDropperColorPickerModel
    {
        [JsonProperty("rgba")]
        public long[] Rgba { get; set; }

        [JsonProperty("hsla")]
        public double[] Hsla { get; set; }

        [JsonProperty("rgbString")]
        public string RgbString { get; set; }

        [JsonProperty("rgbaString")]
        public string RgbaString { get; set; }

        [JsonProperty("hslString")]
        public string HslString { get; set; }

        [JsonProperty("hslaString")]
        public string HslaString { get; set; }

        [JsonProperty("hex")]
        public string Hex { get; set; }

        [JsonProperty("alpha")]
        public double Alpha { get; set; }
    }
````

bjarnef:
It could probably implement a custom property editor using `DefaultPropertyValueConverter` similar to `ColorPickerValueConverter`: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/PropertyEditors/ValueConverters/ColorPickerValueConverter.cs

Currently it just store an `object`, but storing it a `string` (or another type) would probably be considered as a breaking change.

Not sure we want to store all the additional data you suggest, where it may be better/simpler to store an `JSON` object.

There are however ways vai C# or JavaScript to get alpha, convert hex > rgba, rgba to hex(a), etc.
A JavaScript solution could be to use [TinyColor](https://github.com/bgrins/TinyColor)

Most developers probably don't need more than just the color code: hex or rgba.
If you do need this I would suggest creating a custom property editor using the `<umb-color-picker>` component. That way you have full control or the return type.

The events returns `tinyColor` object:
https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web.UI.Client/src/common/directives/components/umbcolorpicker.directive.js#L179-L221

[...]

PR #9988, opened 2021-03-16
Eye Dropper macro parameter editor

bjarnef:
### Prerequisites

- [x] I have added steps to test this contribution in the description below

### Description
This PR is a follow-up on https://github.com/umbraco/Umbraco-CMS/pull/9646 allowing the Eye Dropper Color Picker to be used as macro parameter editor.

image

image

PR #10853, opened 2021-08-13
V9: Eye Dropper property value converter

bjarnef:
### Prerequisites

### Description
This PR add a property value converter for Eye Dropper property editor. It related to this issue in v8 https://github.com/umbraco/Umbraco-CMS/issues/10323 but simplify the value converter to a string. Most probably only need to render the color, e.g. as inline background color.

If you want a more complex property editor and a value converter returning a complex model, you could always create your own property editor using `<umb-color-picker>` component and use `tinyColor` to store a complex JSON object with e.g. properties hex, rgba, alpha, maybe even R, G, B and A splitted in separate properties.

image

Furthermore it allow the property editor to be used as macro parameter editor as in this PR for v8:
https://github.com/umbraco/Umbraco-CMS/pull/9988

### DIFF
--- a/src/Umbraco.Core/Constants-PropertyEditors.cs
+++ b/src/Umbraco.Core/Constants-PropertyEditors.cs
@@ -1,4 +1,4 @@
-﻿using Umbraco.Cms.Core.PropertyEditors;
+using Umbraco.Cms.Core.PropertyEditors;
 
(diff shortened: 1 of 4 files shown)


Zeegaan:
Tested this and it works !

Code looks good to me 👍 

Thank you for your contribution 💪

PR #10855, opened 2021-08-13
V9: Fix packages overflow

bjarnef:
### Prerequisites

### Description

In Umbraco v8 the packages panel header was an `<button>` element, which inherited `box-sizing: border-box`.

image

In Umbraco v8 this is no longer the case as the header is just a `<div>` element, which by default has `box-sizing: content-box` and thus cause page owerflow.

image

With this change it looks much nicer.

image

image


Zeegaan:
I took a look at this, and tested it, works great 👍 

Code makes sense to me, altough i am no front-end expert 😄 

Thank you for your contribution! 💪

PR #10859, opened 2021-08-15
Process requires disposal

nzdev:
### Prerequisites

### Description
Process class needs to be disposed as it is IDisposable.


bergmania:
Thanks @nzdev, it totally makes sense 🚀

PR #10860, opened 2021-08-15
Avoid mixing using of string interpolation and string formatting in log messages

nzdev:
### Prerequisites

### Description
Avoid mixing using of string interpolation and string formatting in log messages. The messageformat parameter gets cached and should be a constant string.


bergmania (approved):

bergmania:
Thanks 💪

PR #10865, opened 2021-08-16
v9: Removed VariationContextAccessor from IUmbracoContext

Zeegaan:
# Notes

- Removed VariationContextAccessor from IUmbracoContext
- Removed VariationContextAccessor from UmbracoContextFactory

# How to test
- Set a breakpoint in `ContentFinderByConfigured404.TryFindContent()`
- Watch the local_variationContextAccessor
- Make an empty doctype and add a page of content
- Try navigating to a site that is not ur homepage, example "localhost:44331/DoesntExist"
- _variationContextAccessor should not be null


nikolajlauridsen (approved):
Looks great 👍

PR #10869, opened 2021-08-16
Raise SendingContentModel event for each element type in blocklist

### LINKED ISSUE #10792: EditorModelEventManager does not emit SendingContentModel event on adding / editing block item to the BlockList dditor

8.15.0

### Bug summary

In the previous versions (for example 8.10) every time when new (empty) block item was added to the BlockList a `SendingContentModel` event was emitetd by `EditorModelEventManager`. It was allowing to perfrom some custom actions, for example to set default values for block item.

After latest update I found out that this event is raising no more. After short investigation I realized the difference is: `GetEmptyByKeys` in v8.15 in comparision with `GetEmptyByKey` is that, it returns a list and I think following event attribute [](https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/WebApi/Filters/OutgoingEditorModelEventAttribute.cs) is skiped because it is expecting `ObjectContent objectContent`, but in fact in the current implementationa collection is passed as a content.

### Specifics

_No response_

### Steps to reproduce

Just try to subscribe to `EditorModelEventManager.SendingContentModel` (for example in custom `IComponent` component) and try to add new item to the BlockList. 

The `EditorModelEventManager.SendingContentModel `event should be raised.

---
_This item has been added to our backlog [AB#13244](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13244)_

nikolajlauridsen:
This PR fixes #10792: `SendingContentModel` event is not raised for blocklist blocks.

The problem is that after the optimizations to the BlockList editor (#10235), we request all empty models for a blocklist in one call with `GetEmptyByKeys`, this broke the `EditorModelEventManager` because we return `IDictionary<Guid, ContentItemDisplay>`, instead of a `ContentItemDisplay`.

To fix this I've updated `OutgoingEditorModelEventAttribute` to now check if we're returning an `IDictionary`, if we are, we iterate through each `DictionaryEntry`, and raise an event with the value. 

### Testing

1. Create a component and listen to the `EditorModelEventManager.SendingContentModel` event
2. Create a document type with a blocklist property with at least one available block
3. Create a piece of content using the document type.
4. Ensure that the event is raised for each allowed block in your blocklist editor. 
5. Ensure that changes made when handling the events are reflected in the frontend, for instance, if you set a default value in a property.

I used this snippet for testing (Note: this snippet only works if your element type name contains "block" and 1 or more text strings/text boxes): 

```C#
using System.Linq;
using System.Web.Http.Filters;
using Umbraco.Core;
using Umbraco.Core.Composing;
using Umbraco.Web.Editors;
using Umbraco.Web.Models.ContentEditing;

namespace Umbraco.Web.UI
{
    public class EventTester : IComponent
    {
        public void Initialize()
        {
            EditorModelEventManager.SendingContentModel += TestSendingContentModelEvent;
        }

        public void Terminate()
        {
            EditorModelEventManager.SendingContentModel -= TestSendingContentModelEvent;
        }

        public void TestSendingContentModelEvent(HttpActionExecutedContext sender, EditorModelEventArgs<ContentItemDisplay> e)
        {
            Current.Logger.Info(GetType(), $"Received SendingContentModelEvent: {e.Model.ContentTypeName}");
            var props  = e.Model.Variants.FirstOrDefault()?.Tabs.SelectMany(tab => tab.Properties).ToList();

            if (e.Model.ContentTypeName.ToLower().Contains("block"))
            {
                foreach (var prop in props)
                {
                    if (prop.Editor.Equals(Constants.PropertyEditors.Aliases.TextBox))
                    {
                        prop.Value = "This is a text box";
                    }
                }
            }
        }
    }


    public class EventComposer : ComponentComposer<EventTester>
    { }
}

```

### DIFF
--- a/src/Umbraco.Web/WebApi/Filters/OutgoingEditorModelEventAttribute.cs
+++ b/src/Umbraco.Web/WebApi/Filters/OutgoingEditorModelEventAttribute.cs
@@ -1,10 +1,8 @@
-﻿using System;
+﻿using System.Collections;
 using System.Net.Http;
 using System.Web.Http.Filters;
-using Umbraco.Core;
 using Umbraco.Web.Composing;
 using Umbraco.Web.Editors;
-using Umbraco.Web.Models.ContentEditing;
 
 namespace Umbraco.Web.WebApi.Filters
 {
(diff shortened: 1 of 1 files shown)


elit0451:
Tested and it works 🎉 Code looks good as well 💪 
Merging 👍 

_P.S. Thanks for the code snippet_

PR #10872, opened 2021-08-17
V8: fix umb-form-check icon z-index to bring icon to top

nathanwoulfe:
### Prerequisites

Relates to #10805 

### Description

The tick icon in the umb-form-check was sitting underneath the ::before pseudo element, so was not visible. Setting the z-index to a non-zero value bumps it back to the top.

@nul800sebastiaan given this was introduced in #10805, it should be included in the same release, being 8.17.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-form-check.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-form-check.less
@@ -98,6 +98,7 @@
     .icon,
     .umb-icon {
         font-size: 1.2rem;
+        z-index: 1;
     }
 
     &__state {


src/Umbraco.Web.UI.Client/src/less/components/umb-form-check.less
```diff
@@ -98,6 +98,7 @@
     .icon,
     .umb-icon {
         font-size: 1.2rem;
+        z-index: 1;
```
bjarnef, line 101: @nathanwoulfe alternatively just `z-index: 1` to `umb-form-check__icon` or `position: relative` to `ng-include` here: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/components/umb-icon.less#L41

bjarnef:
A few other things I noticed.

Grid add icon - seem to be fixed with `line-height: 1;`

image

Icons in media picker

image

bjarnef:
Picture/Image icons not ideal aligned.

image

bjarnef:
I also noticed a minor issue in `<umb-block-card>` where icon doesn't seen to inherit the color.

image

Mostly it is minor alignment or sizing issues though ☺

nathanwoulfe:
@bjarnef well spotted, I'll look at those too - some of the SVG icons may be slightly different sizes to the icon-font equivalent.

The block icon one will be due to setting color rather than fill/stroke

bjarnef:
@nathanwoulfe that is taken care of here: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/less/components/umb-icon.less#L10

A workaround, but there might be a better way to handle this 😅

image

bjarnef:
I fixed and cleaned up some stuff regarding block card icon in https://github.com/umbraco/Umbraco-CMS/pull/10894

src/Umbraco.Web.UI.Client/src/less/components/umb-form-check.less
bjarnef: @nathanwoulfe maybe we should add this to `umb-form-check__icon` which also set `opacity: 1;`
//cc @nul800sebastiaan

PR #10875, opened 2021-08-17
Minor tour update

warrenbuckley:
Creating content tour (with no SK) had a problem with CSS selector as you was unable to input the text into the textarea on step 6 of 7


nul800sebastiaan:
Just to illustrate the problem: 
image

nul800sebastiaan:
Also removed that dangling `.` you can see in the screen shot under the suggested content block

nul800sebastiaan (commented):
👍

PR #10877, opened 2021-08-17
V9: Help improve using Smidge in FrontEnd views

warrenbuckley:
This just helps users who wish to use Smidge in their frontend views a little easier by adding the TagHelper and injecting SmidgeHelper to align with Shan's examples on the Smidge repo.

## Test Notes
* From the CI build update dotnet new templates
* Create new dotnet new umbraco --name TestingSmidge
* Try some of the examples from Shan's docs
https://github.com/Shazwazza/Smidge/wiki/Declarations
https://github.com/Shazwazza/Smidge/wiki/Rendering
https://github.com/Shazwazza/Smidge/wiki/Bundling-strategies

### DIFF
--- a/src/Umbraco.Web.UI.NetCore/Startup.cs
+++ b/src/Umbraco.Web.UI.NetCore/Startup.cs
@@ -6,6 +6,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Umbraco.Cms.Core.DependencyInjection;
 using Umbraco.Extensions;
+using Smidge;
 
 namespace Umbraco.Cms.Web.UI.NetCore
 {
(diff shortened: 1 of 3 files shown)


src/Umbraco.Web.UI.NetCore/Startup.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.Extensions.Hosting;
 using Umbraco.Cms.Core.DependencyInjection;
 using Umbraco.Extensions;
+using Smidge;
```
bergmania, line 9: Why is this committed?

warrenbuckley: The thought behind this is that if you want to add app.UseSmidge() and setup some bundles. You would discover that extension method if you are poking around with `app.` with IntelliSense.

So for me its about discoverability. Happy to remove it though

bergmania: IMO you should do that in your own extension methods anyway and not directly in startup.cs. I think we should remove it.

bergmania:
```suggestion

```

PR #10878, opened 2021-08-17
V9: Keep the old indexes in case the DictionaryTablesIndexes migration fails

elit0451:
Additional refactoring based on the [comment](https://github.com/umbraco/Umbraco-CMS/pull/10826#issuecomment-900176240) of #10826.


nikolajlauridsen (approved):
Tested it out and it works as expected 👍, code looks good too 😄.

Just need the build server to be happy and I'll merge

PR #10880, opened 2021-08-17
Suppress scope notifications during migrations

Shazwazza:
* Runs all unattended package migrations in a single outer scope so they all run in a single transaction. 
* Suppress scope notifications for each migration in MigrationPlanExecutor.  This is required because if Umbraco services are used in one package's migrations (lets call it package "A") and those service notifications are published while another package (lets call it package "B") is listening on those notifications but it's migrations have not been executed yet - this can lead to exceptions since package "B" may be trying to do things when it is not ready to do them since it's migrations haven't been run.
* Adds new notifications:
  * `MigrationPlanExecuted` published whenever a single migration plan is executed. This will be published during Umbraco's own migration plans being executed and any package migration plans. It will be published during both unattended migrations and explicit migrations in the back office. Handlers can inject the IRuntimeState to determine at what phase the migration is being executed.
  * `UnattendedPackageMigrationsExecutedNotification` raised after an unattended package migration phase is carried out successfully.

I added some TODOs which could/should be looked into as well.

---
_This item has been added to our backlog [AB#13389](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13389)_


p-m-j:
I wonder if it makes more sense to fire the same event from both UI and unattended on completion (e.g. fire MigrationPlanExecuted once unattended has finished all migrations), otherwise any one that cares will likely have to subscribe to both notiifcations as you will only get one or the other (notifications supressed during unattended so only UnattendedPackageMigrationsExecutedNotification would be fired)

If there are good arguments for keeping both, we could do with sign positing (see also etc) on the events as someone may discover they need one but not know about the other use case.

Shazwazza:
@p-m-j  I've pushed an update. There's only one notification raised now and all package migrations are done with the `PackageMigrationRunner` now. 

Not sure what you mean by?

> Edit: Checked the point at which notifications are supressed and realised most of it is incorrect.

p-m-j:
I assumed you would have fired higher up the call stack, probably should have looked before commenting originally.

p-m-j:
Will hit this with some package migrations Monday am

PR #10881, opened 2021-08-17
v9: Remove Composers.cs, various composition attributes and IUserComposer

p-m-j:
Closes off some of the outstanding items from #9446

Removed Composers.cs
Removed attributes for ordering / disabling
Removed IUserComposer as it does nothing that IComposer doesn't

Skipped rename IComposer -> IUmbracoComposer as it doesn't feel so important and will probably just add effort for package folks.

Would be really nice to get this in rc2 to give people more time to handle issues without attributes.


p-m-j:
Waiting for feedback in #10900

PR #10884, opened 2021-08-18
V9: Move default ImageSharp cache folder to TEMP and map configured value to content root

ronaldbarendse:
### Prerequisites

### Description
The default ImageSharp.Web cache folder was configured as `../umbraco/mediacache`, which used a relative path (rooted to the web root), didn't follow the pascal case naming convention and wasn't placed in the `Data/TEMP` folder (where all other caches/temporary data resides). This folder also wasn't excluded in the `.gitignore`, so all caches images would show up as newly added/available files in git.

This PR changes the default cache folder to `~/umbraco/Data/TEMP/MediaCache` and maps this value to the content root (to resolve the tilde).

### Testing steps
Make sure the previous cache folder is deleted before starting the application and then simply ensure the ImageSharp middleware needs to process an image by uploading a media item and either:
- Manually add `?width=500` to the URL of the media item URL;
- Opening the media section overview in the back-office (this will create the thumbnails using ImageSharp).

The cached images should be stored in the new default cache folder.

To test changing the configured folder, add the following `Umbraco:CMS` setting: `"Imaging:Cache:CacheFolder": "~/umbraco/Data/TEMP/is-cache"` and check whether this folder is created/populated with cached images using the above testing steps.

_On a related note: the log files directory isn't located within the `umbraco/Data` folder (the current value is hard-coded as `~/umbraco/Logs`), which is strange, as all files written by the CMS should probably live in the data folder, right? That would at least make it easier to grant write permissions to this single folder..._


bergmania:
@ronaldbarendse , are you aware of the integration tests failing?

ronaldbarendse:
@bergmania No, but it looks like the tests aren't setup to correctly mock the call to `builder.BuilderHostingEnvironment.MapPathContentRoot()`. I'll make sure to fix them 👍🏻

ronaldbarendse:
The documentation also needs updating after merging this, as it contains the default values and explanation of the relative cache folder path: https://our.umbraco.com/documentation/Reference/V9-Config/ImagingSettings/.

bergmania (approved):

PR #10885, opened 2021-08-18
V8: Improve nested content optimization in the backoffice

nikolajlauridsen:
This PR is working towards giving nested content the same treatment as the blocklist editor. 

The majority of this is changing it so all the empty content types are requested in a single request, additionally, this also allows us to apply the same mapping optimizations as for blocklist by using the `GetEmpties` method on the controller.

I've also made some additional changes to the frontend side of things before we kept calling `initIfAllScaffoldsHaveLoaded` because we did not know exactly when all the scaffolds had been loaded since we loaded them asynchronously one by one. Now we just call `initNestedContent` once the response from `GetEmptyByAliases` has been processed. 

Additionally, I noticed that we would create an array of content type aliases more than once, I've changed this so we only do this once when the controller is initialized. 

#### The gains

As mentioned before we would call `GetEmpty` separately for every element type in the nested content, each of these calls could take anywhere from 20-60 ms, and produce somewhere between 2 and 11 db queries: 

image

Now we do it all in a single request, which in my testing with 5 element types took around 20-30 ms, and uses 2 db queries:

image

Of course this will take longer the more element types you have, but is regardless a significant improvement from before.

#### Testing

Remember to run `npm run build` and do a "Empty Cache and Hard Reload" (I'm only mentioning this because I keep forgetting this myself 😆)

1. Create some element types, I also tested with an element type with a nested content picker as well.
2. Create a doctype with a nested content picker.
3. Ensure that it still works as expected.

Since `EditorModelEventManager` now handles dictionaries, the `SendingContentModel` should still work, but to be on the safe side I also tested it with a snippet similar to #10869:

```C#
using System.Linq;
using System.Web.Http.Filters;
using Umbraco.Core;
using Umbraco.Core.Composing;
using Umbraco.Web.Editors;
using Umbraco.Web.Models.ContentEditing;

namespace Umbraco.Web.UI
{
    public class EventTester : IComponent
    {
        public void Initialize()
        {
            EditorModelEventManager.SendingContentModel += TestSendingContentModelEvent;
        }

        public void Terminate()
        {
            EditorModelEventManager.SendingContentModel -= TestSendingContentModelEvent;
        }

        public void TestSendingContentModelEvent(HttpActionExecutedContext sender, EditorModelEventArgs<ContentItemDisplay> e)
        {
            Current.Logger.Info(GetType(), $"Received SendingContentModelEvent: {e.Model.ContentTypeName}");
            if (e.Model.IsElement)
            {
                var props  = e.Model.Variants.FirstOrDefault()?.Tabs.SelectMany(tab => tab.Properties).ToList();
                foreach (var prop in props)
                {
                    if (prop.Editor.Equals(Constants.PropertyEditors.Aliases.TextBox))
                    {
                        prop.Value = "This is a text box";
                    }
                }
            }
        }
    }


    public class EventComposer : ComponentComposer<EventTester>
    { }
}
```

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/resources/content.resource.js
+++ b/src/Umbraco.Web.UI.Client/src/common/resources/content.resource.js
@@ -642,6 +642,24 @@ function contentResource($q, $http, umbDataFormatter, umbRequestHelper) {
                     return $q.when(umbDataFormatter.formatContentGetData(result));
                 });
         },
+
+        getScaffolds: function(parentId, aliases){
+            return umbRequestHelper.resourcePromise(
+                $http.post(
+                    umbRequestHelper.getApiUrl(
+                        "contentApiBaseUrl",
+                        "GetEmptyByAliases"),
+                        { parentId: parentId, contentTypeAliases: aliases }
+                    ),
+                    'Failed to retrieve data for empty content item aliases ' + aliases.join(", ")
+                ).then(function(result) {
+                    Object.keys(result).map(function(key){
+                        result[key] = umbDataFormatter.formatContentGetData(result[key]);
+                    });
+
+                    return $q.when(result);
+                });
+        },
         /**
          * @ngdoc method
          * @name umbraco.resources.contentResource#getScaffoldByKey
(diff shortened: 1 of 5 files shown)


elit0451 (commented):
The code looks good, found a couple of typos that I will fix. Onto testing now 👍

elit0451:
Tested and could see the optimizations 💪 The event still works as well.
Great work! Merging

PR #10886, opened 2021-08-18
V9: Rename Umbraco.Web.UI.Netcore to Umbraco.Web.UI

elit0451:
## Details
- Removing old references of the _Umbraco.Web.UI_ ASP.NET project;
- Renaming _"Umbraco.Web.UI.Netcore"_ to _"Umbraco.Web.UI"_ (project, folder, references) and respective fixes (to build script, Gulp config, dotnet new templates, etc.);
- Putting "Umbraco.Web.UI" project reference as the first one in the solution files (.sln) since it will be selected as the startup one by default.

## Test
- Backoffice should work as before;
- Creating a new project through the `dotnet new template` should still work after this PR.


bergmania:
I removed the label `release/9.0.0-rc003`. Even that it is technically correct, the result of this PR should not be relevant for users of RC3 :)

Zeegaan:
Great work @elit0451 ! 💪 

I have tested this, and it works. Changes looks good to me!

Will be happy to merge once tests are passed 👍

PR #10889, opened 2021-08-18
v9: Implemented new asserts

Zeegaan:
# Notes
- Updated CreatedPackagesRepositoryTests.Update() to use Assert.Multiple so it is easyer to debug
- Updated CreatedPackagesRepositoryTests.Export_Zip()
- Updated CreatedPackagesRepositoryTests.Export_Xml()
- Updated PackageInstallationTest.Can_Read_Compiled_Package_1()
- Updated PackageInstallationTest.Can_Read_Compiled_Package_2()
- Updated PackageInstallationTest.Can_Read_Compiled_Package_Warnings()
# How to test
- Run the unit test to see if they pass

### DIFF
--- a/src/Umbraco.Tests.Integration/Umbraco.Core/Packaging/CreatedPackagesRepositoryTests.cs
+++ b/src/Umbraco.Tests.Integration/Umbraco.Core/Packaging/CreatedPackagesRepositoryTests.cs
@@ -148,16 +148,34 @@ namespace Umbraco.Cms.Tests.Integration.Umbraco.Core.Packaging
                 Name = "test",
             };
             bool result = PackageBuilder.SavePackage(def);
-
+            //Update values and save
             def.Name = "updated";
+            def.ContentNodeId = "test";
+            def.Languages.Add("Danish");
+            def.Languages.Add("English");
+            def.Scripts.Add("TestScript1");
+            def.Scripts.Add("TestScript2");
             result = PackageBuilder.SavePackage(def);
             Assert.IsTrue(result);
-
             // re-get
             def = PackageBuilder.GetById(def.Id);
             Assert.AreEqual("updated", def.Name);
-
-            // TODO: There's a whole lot more assertions to be done
+            Assert.Multiple(() =>
+            {
+                Assert.AreEqual("updated", def.Name);
+                Assert.AreEqual("test", def.ContentNodeId);
+                Assert.AreEqual(2, def.Languages.Count());
+                Assert.AreEqual(2, def.Scripts.Count());
+                Assert.AreEqual(0, def.DataTypes.Count());
+                Assert.AreEqual(0, def.DictionaryItems.Count());
+                Assert.AreEqual(0, def.DocumentTypes.Count());
+                Assert.AreEqual(0, def.Macros.Count());
+                Assert.AreEqual(0, def.MediaTypes.Count());
+                Assert.AreEqual(0, def.MediaUdis.Count());
+                Assert.AreEqual(0, def.PartialViews.Count());
+                Assert.AreEqual(0, def.Stylesheets.Count());
+                Assert.AreEqual(0, def.Templates.Count());
+            });
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


src/Umbraco.Tests.Integration/Umbraco.Core/Packaging/CreatedPackagesRepositoryTests.cs
```diff
@@ -259,15 +277,24 @@ public void Export_Zip()
...
+                        packageXml.Element("umbPackage").Element("MediaItems").ToString(SaveOptions.DisableFormatting));
+                    Assert.AreEqual(2, zipArchive.Entries.Count());
+                    Assert.AreEqual(ZipArchiveMode.Read, zipArchive.Mode);
+                    Assert.IsNull(packageXml.DocumentType);
+                    Assert.IsNull(packageXml.NextNode);
+                    Assert.IsNull(packageXml.Parent);
+                    Assert.IsNull(packageXml.PreviousNode);
+                    Assert.AreEqual(test, zipArchive.Entries.Last().Name);
+                    Assert.AreEqual(7, zipArchive.Entries.Last().Length);
+                });
```
nikolajlauridsen, line 297: Here you first fetch the `test-file.txt` entry with `zipArchive.GetEntry("media/media/test-file.txt");`, later you fetch the same entry again when you do `Assert.AreEqual(test, zipArchive.Entries.Last().Name);`.

Instead you could just do something like:

```C#
var mediaEntry = zipArchive.GetEntry("media/media/test-file.txt")
Assert.IsNotNull(mediaEntry);
Assert.AreEqual(test, mediaEntry.Name);
Assert.AreEqual(7, mediaEntry.Length);
```

When that's said I don't think we should assert the uncompressed size of the entry, since this is bound to change on different platforms in a text file, for instance Microsoft uses CRLF for line endings where linux just uses LF, this will result in different file sizes, as evident by the Linux tests failing 😄

src/Umbraco.Tests.Integration/Umbraco.Core/Packaging/CreatedPackagesRepositoryTests.cs
```diff
@@ -297,12 +324,16 @@ public void Export_Xml()
...
-
-                Assert.AreEqual($"<Templates><Template><Name>Text page</Name><Alias>textPage</Alias><Design><![CDATA[@using Umbraco.Cms.Web.Common.PublishedModels;{Environment.NewLine}@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage{Environment.NewLine}@{{{Environment.NewLine}\tLayout = null;{Environment.NewLine}}}]]></Design></Template></Templates>", xml.Element("umbPackage").Element("Templates").ToString(SaveOptions.DisableFormatting));
-
-                // TODO: There's a whole lot more assertions to be done
-
+                Assert.Multiple(() =>
+                {
+                    Assert.AreEqual("umbPackage", xml.Root.Name.ToString());
+                    Assert.AreEqual($"<Templates><Template><Name>Text page</Name><Alias>textPage</Alias><Design><![CDATA[@using Umbraco.Cms.Web.Common.PublishedModels;{Environment.NewLine}@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage{Environment.NewLine}@{{{Environment.NewLine}\tLayout = null;{Environment.NewLine}}}]]></Design></Template></Templates>", xml.Element("umbPackage").Element("Templates").ToString(SaveOptions.DisableFormatting));
+                    Assert.AreEqual(xml.FirstNode, xml.LastNode);
```
nikolajlauridsen, line 331: I'm not entirely sure what we're testing here? Except maybe that there's only a single node in the XML?

src/Umbraco.Tests.Integration/Umbraco.Infrastructure/Packaging/PackageInstallationTest.cs
```diff
@@ -66,9 +72,17 @@ public void Can_Read_Compiled_Package_Warnings()
...
             InstallWarnings preInstallWarnings = package.Warnings;
-            Assert.IsNotNull(preInstallWarnings);
-
-            // TODO: More Asserts
+            var dataType = package.DataTypes.First();
+            Assert.Multiple(() =>
+            {
+                Assert.AreEqual("Document Type Picker", package.Name);
+                Assert.AreEqual("3593d8e7-8b35-47b9-beda-5e830ca8c93c", dataType.LastAttribute.Value);
+                Assert.AreEqual("Document Type Picker", dataType.FirstAttribute.Value);
```
nikolajlauridsen, line 80:
```suggestion
                Assert.AreEqual("3593d8e7-8b35-47b9-beda-5e830ca8c93c", dataType.LastAttribute?.Value);
                Assert.AreEqual("Document Type Picker", dataType.FirstAttribute?.Value);
```

Possible null reference exception, with this change you would see something like "expected Document Type Picker, but got null", instead of a null reference exception, not a major thing, but it's a bit easier to understand than a null reference  exception 😄

Issue #9908, opened 2021-02-27
Parenthesis around node name when using variants


uSkinned:
When setting up a Document Type to vary by culture the node name will always have parenthesis around its name if you only have one language installed.

This scenario will occur if you setup a site to use variants and want it to go live with only one language, with a plan to launch new languages at a later date.

See image attached demonstrating the issue. I have also attached a package which must be installed against 8.12.0 RC due to another issue regarding variants that has been fixed in this release ([https://github.com/umbraco/Umbraco-CMS/pull/9524])

The attached package will install a basic home node setup demonstrating this issue as shown in the screenshot.

![parenthesis](https://user-images.githubusercontent.com/22817718/109392442-f499ed00-7913-11eb-9030-2a11537f57f6.png)

[Parentheses_Issue_1.0.0.zip](https://github.com/umbraco/Umbraco-CMS/files/6054850/Parentheses_Issue_1.0.0.zip)

emma-hq:
Hiya @uSkinned,

This seems like a sensible change. We hadn't really anticipated this as a use-case -  a multi-lingual site with one language at launch and then a second added later. I can see why the default behaviour could change here.

I'll label as up-for-grabs and someone can pick this up. 

Emma

umbrabot:
Hi @uSkinned,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

mistyn8:
https://github.com/umbraco/Umbraco-CMS/blob/4aff5d38f83f7f2500ed9b75eb74abd34272f5c3/src/Umbraco.Web/Trees/ContentTreeController.cs#L283

Think this is that culture querystring [?mculture=en-GB] is only appended if more than one language, with one language culture is no resolved, so always get the invariant ()

https://github.com/umbraco/Umbraco-CMS/blob/187db95368f7c87126a83d8603465c6e65cd4c0c/src/Umbraco.Web.UI.Client/src/navigation.controller.js#L220

Andrei-:
[...]

PR #10892, opened 2021-08-18
9908: No parentheses & language picker with only 1 language

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/9908

### Description
The issue here was that because the language wasn't passed, it would set the language to check with to an empty string. It would then try to find an empty string culture on the content item (which was of course not present) and add the parentheses on the name. Adding a fallback to the default language id solves this issue.

For the language picker, I had to update a bit of JavaScript code that checks if it has a variant or not. Updating this check to always check above 1 fixes the issue here. It'll still show the language picker if there is one 1 published variant and 1 unpublished variant.

How to test:

1. Use starter kit
2. Set the homepage doc type to variant by culture
3. Observe the homepage. In a past version, it would now show parentheses around the homepage and it would show the language picker.
4. Add a second language
5. You'll see that the parentheses logic still works when switching between main languages as well as the language picker.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/editor/umbeditorcontentheader.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/editor/umbeditorcontentheader.directive.js
@@ -105,8 +105,8 @@
                     }
                 });
 
-                scope.vm.hasVariants = (scope.vm.hasCulture || scope.vm.hasSegments);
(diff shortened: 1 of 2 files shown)


nathanwoulfe (approved):
Nice work @patrickdemooij9, works as advertised. Also fixes the case where deleting a language leaves only one culture, which left the `scope.vm.hasCulture` property as true, likely because the culture from when the node was variant is still set. Something like that.

Regardless, all works nicely, thank you!

PR #10894, opened 2021-08-18
Inherit text color from parent element

bjarnef:
### Prerequisites

### Description
After PR https://github.com/umbraco/Umbraco-CMS/pull/10805 was merged I noticed block card didn't show the selected icon color.

The icon name includes color class as well, which is a bit of a mess and color classes are forcing color with `!important` as well. Furthermore the inline color style need `!important` as well to overwrite this 🙀

I could continue the trend with another `!important` rule as in https://github.com/umbraco/Umbraco-CMS/commit/bc5a3815fdd64069e1c8e7d9a5537bc00870e13d to inherit the correct color as well in `<umb-icon>` component.

image

However I think this would be just another hack.

Instead it should just include the icon and furthermore we don't need to use `!important` on the inline style to get the correct color.

image

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-icon.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-icon.less
@@ -32,6 +32,8 @@
     }
 
     &__inner {
+        color: inherit !important;
+
         // Clear pseudo classes
         &::before,
         &::after {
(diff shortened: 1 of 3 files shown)


nathanwoulfe (approved):
All good @bjarnef, removing !important is always a good thing.

I made a minor change in the component js, to use an interpolated string when setting `styleBackgroundImage`, since that removes the need to escape the quotes, which makes for an easier-to-read value:
```
'url(\'' + path + '\')' => `url('${path}')`
```

Issue #10341, opened 2021-05-27
Use the dialog from MNTP when selecting allowed child node types in a document type.


skttl:
### Umbraco version

v8

### Description

When adding allowed child node types to a document type, the available document types opens in an overlay, where they are presented in a grid.

Besides not showing custom icons, this grid can be hard to grasp, especially if you have a lot of document types, or document types with equal names.

![image](https://user-images.githubusercontent.com/3726467/119814458-b5ffd880-beea-11eb-9409-b85b092ce8dc.png)

I think it would be better if it could open a dialog, with a hierarchical view, like the MNTP does when selecting allowed document types.

![image](https://user-images.githubusercontent.com/3726467/119814592-dd56a580-beea-11eb-95d6-07797a36a271.png)

### How can you help?

I'll be happy to submit a PR changing this

nathanwoulfe:
I like it. The dialog icons are pretty big and overpowering, and the tree works more naturally for a folder-based structure. 

If it were up to me, I'd happily merge a PR 😁

enkelmedia:
Great idea! It’s not really clear from the screenshot of the dialog used the same folder structure as your doc types-folder. It I guess that this is the case?

nielslyngsoe:
@skttl YES, yes yes yes. Much better on multiple levels.

- First, there is a better visual overview of the options.
- It's a recognizable view, as the Document Types are presented in the same way as we know from the Setting-section tree.

As well I would propose to change the User Flow a little bit with this change. So the user clicks `Allow child node`, the Infinite Layer with the Tree structure appears. In this, the user can select multiple DocTypes, when clicking `Submit`(bottom right corner) the dialogs closes, and the data is set.
This will give one more click to add a single child, but it makes it much faster to allow multiple child-types.

Additionally, it enables us to make a button called `Create new Document Type`, which enables the user to create a DocType structure top-down, but this is for another PR.

Thanks for the awareness!

nul800sebastiaan:
💡 How about the other way around as well, a button "allow this under the following doctypes". Traditionally we did not have this because it would take people out of the flow of creating a doctype.. this has definitely led to some frustration. But now with infinite editing it becomes part of the flow.. could work?

umbrabot:
Hi @skttl,

[...]

PR #10896, opened 2021-08-18
10341: Use different picker for content types

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10341

### Description
Replaces the content type dialog with the content type picker for both the content and the media types.
ContentTypes


src/Umbraco.Web.UI.Client/src/views/documenttypes/views/permissions/permissions.controller.js
```diff
@@ -6,10 +6,10 @@
  * @description
  * The controller for the content type editor property dialog
  */
-(function() {
+(function () {
     'use strict';
 
-    function PermissionsController($scope, $timeout, contentTypeResource, iconHelper, contentTypeHelper, localizationService, overlayService) {
+    function PermissionsController($scope, $timeout, contentTypeResource, iconHelper, contentTypeHelper, localizationService, overlayService, editorService) {
```
patrickdemooij9, line 12: I should remove localizationService and overlayService here as they aren't used anymore

nathanwoulfe: Hey @patrickdemooij9, this is a really nice change, it's a much, much improved UX, worthy of a double thumbs up 👍 👍  

I removed those params and did a quick syntax update in both files (using es6 methods instead of underscore, single quotes, arrows).

nathanwoulfe (approved):

src/Umbraco.Web.UI.Client/src/views/documenttypes/views/permissions/permissions.controller.js
patrickdemooij9: Ah great, thank you! I totally forgot that I had left that comment here 😅

PR #10898, opened 2021-08-19
v9: removed obsolete property from User and IUser

Zeegaan:
# Notes
- Removed AdditionalData from User
- Removed ToUserCache and FromUserCache from User
- Removed ToUserCache and FromUserCache from IUser

# How to test
- Unit tests should run


nikolajlauridsen (changes requested):
Overall looks good, however, now that we've removed `internal IDictionary<string, object> AdditionalData`, we should also be able to remove the private backing variable and its lock:

```C#
private IDictionary<string, object> _additionalData;
private object _additionalDataLock = new object();
```

Had a quick look and it doesn't seem to be really used for anything other than being cloned, so it should be safe to remove.

nikolajlauridsen (approved):
Looks good 👍

Issue #10903, opened 2021-08-19
Umbraco 9.0.0 RC002 - Can't copy content at root


KevinJump:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0 rc002

### Bug summary

Attempting to copy a content item/tree at the root level fails. 

if you right click on the root node of your site and select copy, and then attempt to copy the content to the root. the screen disappears and nothing gets copies. 

_copying content beneath the root level works._ 

### Steps to reproduce

1. Create a content item in the root of the site.
2. right click - copy
3. select 'Content' as the destination

screen closes no copy happens. 

![rc002-copyfail](https://user-images.githubusercontent.com/431231/130058746-2bbb65bf-9f41-44c1-815b-31127d4a7877.gif)

### Expected result / actual result

Content should be copied to root level next to existing content.

bergmania:
Thanks for reporting @KevinJump. I can confirm I can repro on both `v9/dev` and `v8/dev`.. Seems to be introduced by this PR: https://github.com/umbraco/Umbraco-CMS/pull/10762

nul800sebastiaan:
Turns out this is an issue that originates from 8.16-RC https://github.com/umbraco/Umbraco-CMS/pull/10762

@bjarnef Would you have a chance to check this out please? I can just revert the PR for final release of 8.16 but it would be nicer to fix properly! 😁

bjarnef:
@nul800sebastiaan yeah, I know what is going on. I'll have a look - not only an issue regarding copy, but also move action.

bjarnef:
PR: https://github.com/umbraco/Umbraco-CMS/pull/10907

PR #10907, opened 2021-08-19
Hide backdrop when closing/select navigation

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10903

### Description
While optimizing some duplicate code in https://github.com/umbraco/Umbraco-CMS/pull/10762 I also changed to call `hideMenu()` method from navigationService.

However this method also do some other stuff like changing menu state.

Besides this I think it goes longer back as it originally hide backdrop in `select()` function, but may also make sense when the dialog is closed. However in copy and move action when selecting root node, the dialog isn't closed, but the backdrop is hidden.

NdCdrq07Qd

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/tree/umbtree.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/tree/umbtree.directive.js
@@ -358,7 +358,7 @@ function umbTreeDirective($q, treeService, navigationService, notificationsServi
             */
             $scope.select = function (n, ev) {
 
-                navigationService.hideMenu();
+                navigationService.hideBackdrop();
                 
                 if (n.metaData && n.metaData.noAccess === true) {
                     ev.preventDefault();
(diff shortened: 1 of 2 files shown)


bjarnef:
@nul800sebastiaan this fixes the issue, although the backdrop is closed/hidden when selecting root node.
This was originally added in https://github.com/umbraco/Umbraco-CMS/pull/8121

Not sure if we actually should close/hide backdrop here as the select method not necessarily close the dialog.

bjarnef:
Actually I don't think we need to hide backdrop, when selecting a menu item, so I have changed this in https://github.com/umbraco/Umbraco-CMS/pull/10907/commits/bd9e141c61923e13c222550836cc9ff5279e6583

Maybe @BatJan recall why we needed that?

nathanwoulfe:
@bjarnef second example looks correct - the backdrop should stay visible as long as the overlay is open. It shouldn't be possible to close one without closing the other.

bjarnef:
@nathanwoulfe I just tried on a clean Umbraco v8.15.1 and it seems to be same behaviour closing backdrop when clicking root node in move or copy dialog, which it shouldn't.

PR #10913, opened 2021-08-20
Fix PhysicalFileSystem.GetRelaitvePath for *nix

### LINKED ISSUE #10912: v9 Block List Thumbnail Selector Empty on *nix
### Version

v9/dev

### Bug summary

The catalog appearance thumbnail selector is empty when hosted on MacOS. It should render a tree view allowing selection of static files from App_Plugins and wwwroot.

### Specifics

On Windows you get a tree picker with App_Plugins and wwwroot folders; however, on MacOS, the tree picker is empty.
image

I believe this is a regression from #10794 

Specifically, `_fileSystem.GetDirectories` is returning full paths, but we are comparing only local paths so `AppPlugins` and `WebRoot` fail to be recognized, they pass the test on line 42 and the loop continues. The screenshot below shows `directory` and `AppPlugins` constant failing to compare.
image

Even more specifically, I think there is a bug in `Umbraco.Cms.Core.IO.PhysicalFileSystem.GetRelativePath(string fullPathOrUrl)`. When I call it on directory it returns an absolute path. 
```
_fileSystem.GetRelativePath(directory)
"Users/evanm/src/Umbraco-CMS/src/Umbraco.Web.UI/App_Plugins"
```

I believe`GetRelativePath` is assuming a file system url starts with c:\ (or equivalent) rather than '/' ie '/var/wwwroot/'
```
public string GetRelativePath(string fullPathOrUrl)
        {
            // test URL
            var path = fullPathOrUrl.Replace('\\', '/'); // ensure URL separator char

            // if it starts with the root URL, strip it and trim the starting slash to make it relative
            // eg "/Media/1234/img.jpg" => "1234/img.jpg"
            
            // **** I think this is wrong. File paths on Unix systems start with '/' I think this assumes file paths start with C:\
            // and so will pass onto the next test.
            if (_ioHelper.PathStartsWith(path, _rootUrl, '/'))
                return path.Substring(_rootUrl.Length).TrimStart(Constants.CharArrays.ForwardSlash);

            // if it starts with the root path, strip it and trim the starting slash to make it relative
            // eg "c:/websites/test/root/Media/1234/img.jpg" => "1234/img.jpg"
            if (_ioHelper.PathStartsWith(path, _rootPathFwd, '/'))
                return path.Substring(_rootPathFwd.Length).TrimStart(Constants.CharArrays.ForwardSlash);

            // unchanged - what else?
            return path;
        }
```

I think something as simple as reversing these two tests would fix it such as:
```
        public string GetRelativePath(string fullPathOrUrl)
        {
            // test URL
            var path = fullPathOrUrl.Replace('\\', '/'); // ensure URL separator char

            // if it starts with the root path, strip it and trim the starting slash to make it relative
            // eg "c:/websites/test/root/Media/1234/img.jpg" => "1234/img.jpg"
            // or on unix systems "/var/wwwroot/test/Meia/1234/img.jpg"
            if (_ioHelper.PathStartsWith(path, _rootPathFwd, '/'))
                return path.Substring(_rootPathFwd.Length).TrimStart(Constants.CharArrays.ForwardSlash);

            // if it starts with the root URL, strip it and trim the starting slash to make it relative
            // eg "/Media/1234/img.jpg" => "1234/img.jpg"
            if (_ioHelper.PathStartsWith(path, _rootUrl, '/'))
                return path.Substring(_rootUrl.Length).TrimStart(Constants.CharArrays.ForwardSlash);

            // unchanged - what else?
            return path;
        }
```

### Steps to reproduce

On MacOS (again this works great on Windows)
1. Create a clean v9/dev site
2. Create a BlockList data type
3. Add a Block to it
4. Click Add Thumbnail under Catalogue Appearance

The StaicFilesTreeController allows App_Plugins and Webroot to be navigable on MacOS.

cleversolutions:
### Prerequisites

Fixes #10912 

### Description
This PR fixes a problem where GetRelativePath fails on Unix systems. It resolves a fullPathOrUrl as a Path before trying to resolve it as a URL. This is done by simply moving the path handling code above the url handling code.

### Tests
The following tests were run and passed:
- From Umbraco.Tests.Integration I ran Umbraco.Cms.Tests.Integration.Umbraco.Core.IO.FileSystemsTests
- All tests in Umbraco.Tests.UnitTests project

This bug was showing up when choosing a thumbnail for a block list. Here it can be seen working as expected on MacOS
image
And I also tested on a windows system, where the results look also as expected.

### DIFF
--- a/src/Umbraco.Core/IO/PhysicalFileSystem.cs
+++ b/src/Umbraco.Core/IO/PhysicalFileSystem.cs
@@ -258,15 +258,16 @@ namespace Umbraco.Cms.Core.IO
             // test URL
             var path = fullPathOrUrl.Replace('\\', '/'); // ensure URL separator char
 
+            // if it starts with the root path, strip it and trim the starting slash to make it relative
+            // eg "c:/websites/test/root/Media/1234/img.jpg" => "1234/img.jpg"
+            // or on unix systems "/var/wwwroot/test/Meia/1234/img.jpg"
+            if (_ioHelper.PathStartsWith(path, _rootPathFwd, '/'))
+                return path.Substring(_rootPathFwd.Length).TrimStart(Constants.CharArrays.ForwardSlash);
+
             // if it starts with the root URL, strip it and trim the starting slash to make it relative
             // eg "/Media/1234/img.jpg" => "1234/img.jpg"
             if (_ioHelper.PathStartsWith(path, _rootUrl, '/'))
                 return path.Substring(_rootUrl.Length).TrimStart(Constants.CharArrays.ForwardSlash);
-
-            // if it starts with the root path, strip it and trim the starting slash to make it relative
-            // eg "c:/websites/test/root/Media/1234/img.jpg" => "1234/img.jpg"
-            if (_ioHelper.PathStartsWith(path, _rootPathFwd, '/'))
-                return path.Substring(_rootPathFwd.Length).TrimStart(Constants.CharArrays.ForwardSlash);
 
             // unchanged - what else?
             return path;


p-m-j:
LGTM, If there are any issues I can't see them.

Thanks very much!

PR #10917, opened 2021-08-22
Replace use of dynamic type with POCOs

nzdev:
### Prerequisites

### Description
Replace the use of the dynamic type with strongly typed POCO's for SQL.


src/Umbraco.Core/Persistence/Dtos/AxisDefintionDto.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Umbraco.Core.Persistence.Dtos
+{
+    public class AxisDefintionDto
+    {
+        public int nodeId { get; set; }
+        public string alias { get; set; }
+        public int parentID { get; set; }
```
bjarnef, line 13: I think these properties should be Pascal Case and eventually decorated with attributes.

src/Umbraco.Core/Persistence/Dtos/DefaultConstraintPerColumnDto.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Threading.Tasks;
+
+namespace Umbraco.Core.Persistence.Dtos
+{
+    public class DefaultConstraintPerColumnDto
+    {
+        public string TableName { get; set; }
+        public string ColumnName { get; set; }
+        public string Name { get; set; }
+        public string Definition { get; set; }
```
bjarnef, line 14: I think it would be great to have the other Dto classes using Pascal Case syntax as well.
Not sure the casing matter in the SQL query since it seems to use uppercase, but in that case it can probably be decorated with attributes as well.

ronaldbarendse:
The following DTOs can also be internal/private classes, as they aren't exposed publicly:
- DefaultConstraintPerColumnDto
- TableInSchemaDto
- ColumnInSchemaDto
- ConstraintPerTableDto
- ConstraintPerColumnDto
- DefinedIndexDto

ronaldbarendse:
And the DTOs with a single property can probably also be removed by fetching an array of that type instead 😄

src/Umbraco.Core/Persistence/Dtos/DefaultConstraintPerColumnDto.cs
ronaldbarendse: Some SQL queries already use aliases (`column AS newName` or `newName = column`), which can also be used to match the pascal casing.

nul800sebastiaan:
While reviewing and taking into account the comments I did a round of pair programming with @ronaldbarendse to improve the 2 Dto's that were just 1 property, those weren't necessary. In the process the query for UserStates also got improved, should perform a little better.

As suggested all DTO's now use Pascal cased properties, mapping the value using attributess

Also made the suggested DTO's internal. 

Thanks for getting this one going @nzdev, the fewer dynamics the better! 🎉

nul800sebastiaan:
I'll squash + merge after all the unit tests have ran again.

src/Umbraco.Core/Persistence/Dtos/AxisDefintionDto.cs
```diff
@@ -0,0 +1,16 @@
+﻿using NPoco;
+
+namespace Umbraco.Core.Persistence.Dtos
+{
+    public class AxisDefintionDto
```
ronaldbarendse, line 5: This DTO is still public, can we also change it to internal?

nul800sebastiaan: Done plus `UserNotificationDto` as well.

nul800sebastiaan:
There was one failing test, which is always a bit of a flakey one and doesn't have anything to do with these changes. It works locally when I run it, so should be fine. I'll merge this as it is now. 👍 

Once again, thanks @nzdev!

PR #10919, opened 2021-08-22
Update Serilog.Enrichers.Process to 2.0.2 as it fixes a leak

nzdev:
### Prerequisites

### Description
Update Serilog.Enrichers.Process to 2.0.2 as it fixes a leak


nul800sebastiaan:
The fix mentioned is: https://github.com/serilog/serilog-enrichers-process/pull/17

nul800sebastiaan:
Thanks for fixing our dependency and updating it! 👍

PR #10923, opened 2021-08-22
Add button to reset focal point on umb-image-gravity directive

TheBekker:
PR related to discussion #10882 

### Description

Added simple button on umb-image-gravity directive, to reset focal point of changed from center.

ezgif-3-179ea04d08e1

Missing translations for all except: 
'en'
'en-us'
'da'

### DIFF
--- a/build/NuSpecs/UmbracoCms.Core.nuspec
+++ b/build/NuSpecs/UmbracoCms.Core.nuspec
@@ -33,7 +33,7 @@
                 <dependency id="Newtonsoft.Json" version="[12.0.1,12.999999)" />
                 <dependency id="Semver" version="[2.0.4,2.999999)" />
                 <dependency id="Serilog" version="[2.10.0,2.999999)" />
-                <dependency id="Serilog.Enrichers.Process" version="[2.0.1,2.999999)" />
+                <dependency id="Serilog.Enrichers.Process" version="[2.0.2,2.999999)" />
                 <dependency id="Serilog.Enrichers.Thread" version="[3.1.0,3.999999)" />
                 <dependency id="Serilog.Filters.Expressions" version="[2.1.0,2.999999)" />
                 <dependency id="Serilog.Formatting.Compact" version="[1.1.0,1.999999)" />
--- a/src/SolutionInfo.cs
+++ b/src/SolutionInfo.cs
@@ -19,4 +19,4 @@ using System.Resources;
 
 // these are FYI and changed automatically
 [assembly: AssemblyFileVersion("8.16.0")]
-[assembly: AssemblyInformationalVersion("8.16.0-rc")]
+[assembly: AssemblyInformationalVersion("8.16.0")]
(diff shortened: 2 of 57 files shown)


nul800sebastiaan:
Thanks @TheBekker! No problem on the translations, always fine to only do the ones that you know are correct!

You've zoomed out a little far in the GIF so it's hard to see what's going on (mp4 is allowed to be uploaded to GitHub these days!). However, it looks like the one request I would have for now is to right-align the reset button and don't show it at all when the focal point is at 0.5/0.5. We also don't want to show it when there is no focal point at all of course. Can you make those changes please? 

In case you don't know, adding a new commit to your temp branch and pushing it will update this PR with the latest changes.

TheBekker:
Hi @nul800sebastiaan 

I've just pushed some changes that should fix your concerns.
Here's a new video showing that:
https://user-images.githubusercontent.com/3404996/130580146-edb1003c-0ee2-468e-86ef-723f93b3e674.mp4

Also i noticed that the "umb-image-gravity" directive is also used in: ``Umbraco.Web.UI.Client\src\views\common\infiniteeditors\mediapicker\overlays\mediacropdetails.html`` .
But i can't seem to find anywhere in umbraco where its actually rendered, so right now i have not implemented the reset in that view.

Do you have any idea how to test that view?

nul800sebastiaan:
It's a good question, I'll see if I can find it while reviewing this one!

nul800sebastiaan:
Aha, it's the media picker in the rich text editor.. not quite sure how you get a focal point to appear there though, by default it's not available..

image

nul800sebastiaan:
Ah yes, the media picker in the Grid does show you a focal point, would be great to amend that one with this reset button as well, that's where the `Umbraco.Web.UI.Client\src\views\common\infiniteeditors\mediapicker\overlays\mediacropdetails.html` file is used!

TheBekker:
There we go, reset button on the editor used in Grid as well :)

https://user-images.githubusercontent.com/3404996/131394080-37b0e3c5-69ed-4654-9517-2a0449f70418.mp4

nul800sebastiaan:
Beautiful, thanks very much @TheBekker! 

Just a note, but you couldn't have known this: no need to commit `package-lock.json` especially if you haven't updated any npm dependencies. It doesn't make much of a difference to me, I just reverted it and we're good to go again.

This is all merged now, thanks for the work on this and digging to find every last cropper! 😉

ronaldbarendse:
Looks great! The only suggestion I can give is to make sure the centered buttons in the Media Picker don't jump when the reset button gets activated (maybe by aligning them to the left/right, similar to the remove file button in the image cropper?).

TheBekker:
> Beautiful, thanks very much @TheBekker!
> 
> Just a note, but you couldn't have known this: no need to commit `package-lock.json` especially if you haven't updated any npm dependencies. It doesn't make much of a difference to me, I just reverted it and we're good to go again.
> 
> This is all merged now, thanks for the work on this and digging to find every last cropper! 😉

Ahh sorry must have added it by accident :)

hemraker:
@nul800sebastiaan Pretty sure this should not be categorized as a Modelsbuilder improvement 😅

nul800sebastiaan:
Hehe, sure thing 👍

PR #10925, opened 2021-08-23
Cannot copy page with Block List block with property that 'looks like' JSON #10879

### LINKED ISSUE #10879: Cannot copy page with Block List block with property that 'looks like' JSON

8.15.1

### Bug summary

When trying to copy a page that contains a Block List block that contains a property that 'looks like' JSON (according to the `DetectIsJson` method), an error is thrown.

### Specifics

The copy action will fail with the stack trace below. It is caused by [the logic](https://github.com/umbraco/Umbraco-CMS/blob/release-8.15.1/src/Umbraco.Web/Compose/BlockEditorComponent.cs#L95-L129) going through all properties in order to properly copy them (probably related to making sure the GUIDs are unique). In order to properly handle nested blocks, it tries to parse the value as JSON if the value looks like JSON according to the [DetectIsJson method](https://github.com/umbraco/Umbraco-CMS/blob/release-8.15.1/src/Umbraco.Core/StringExtensions.cs#L147-L153). However, if the value is not actually JSON, this will throw an exception causing the entire copy action to fail.

Because of the lenient nature of the detect method, it also sees values like '[PLACEHOLDER]' or even '[FOO] bar [BAZ]' as JSON, basically any string that starts and ends with `{}` or `[]` brackets. Looking at the rest of the logic, the `Parse` should probably be wrapped in a try-catch (due to a lack of `TryParse` method on `JToken`) and ignore any `JsonReaderException` thrown, as the functionality only needs to do something if the value was an actual JSON object. If that is the correct approach, I can put together a PR for that.

```
Newtonsoft.Json.JsonReaderException: Error parsing boolean value. Path '', line 1, position 2.
   at Newtonsoft.Json.JsonTextReader.ParseTrue() in /_/Src/Newtonsoft.Json/JsonTextReader.cs:line 2403
   at Newtonsoft.Json.JsonTextReader.ParseValue() in /_/Src/Newtonsoft.Json/JsonTextReader.cs:line 1680
   at Newtonsoft.Json.JsonTextReader.Read() in /_/Src/Newtonsoft.Json/JsonTextReader.cs:line 419
   at Newtonsoft.Json.Linq.JContainer.ReadTokenFrom(JsonReader reader, JsonLoadSettings options) in /_/Src/Newtonsoft.Json/Linq/JContainer.cs:line 729
   at Newtonsoft.Json.Linq.JArray.Load(JsonReader reader, JsonLoadSettings settings) in /_/Src/Newtonsoft.Json/Linq/JArray.cs:line 138
   at Newtonsoft.Json.Linq.JToken.ReadFrom(JsonReader reader, JsonLoadSettings settings) in /_/Src/Newtonsoft.Json/Linq/JToken.cs:line 2133
   at Newtonsoft.Json.Linq.JToken.Parse(String json, JsonLoadSettings settings) in /_/Src/Newtonsoft.Json/Linq/JToken.cs:line 2183
   at Umbraco.Web.Compose.BlockEditorComponent.RecursePropertyValues(IEnumerable`1 blockData, Func`1 createGuid) in D:\a\1\s\src\Umbraco.Web\Compose\BlockEditorComponent.cs:line 114
   at Umbraco.Web.Compose.BlockEditorComponent.UpdateBlockListRecursively(BlockEditorData blockListData, Func`1 createGuid) in D:\a\1\s\src\Umbraco.Web\Compose\BlockEditorComponent.cs:line 81
   at Umbraco.Web.Compose.BlockEditorComponent.ReplaceBlockListUdis(String rawJson, Func`1 createGuid) in D:\a\1\s\src\Umbraco.Web\Compose\BlockEditorComponent.cs:line 52
   at Umbraco.Web.Compose.BlockEditorComponent.ReplaceBlockListUdis(String rawJson, Boolean onlyMissingUdis) in D:\a\1\s\src\Umbraco.Web\Compose\BlockEditorComponent.cs:line 33
   at Umbraco.Core.PropertyEditors.ComplexPropertyEditorContentEventHandler.UpdatePropertyValues(IEnumerable`1 props, Boolean onlyMissingKeys) in D:\a\1\s\src\Umbraco.Core\PropertyEditors\ComplexPropertyEditorContentEventHandler.cs:line 62
   at Umbraco.Core.PropertyEditors.ComplexPropertyEditorContentEventHandler.ContentService_Copying(IContentService sender, CopyEventArgs`1 e) in D:\a\1\s\src\Umbraco.Core\PropertyEditors\ComplexPropertyEditorContentEventHandler.cs:line 36
   at Umbraco.Core.Events.TypedEventHandler`2.Invoke(TSender sender, TEventArgs e)
   at Umbraco.Core.Events.QueuingEventDispatcherBase.DispatchCancelable[TSender,TArgs](TypedEventHandler`2 eventHandler, TSender sender, TArgs args, String eventName) in D:\a\1\s\src\Umbraco.Core\Events\QueuingEventDispatcherBase.cs:line 54
   at Umbraco.Core.Services.Implement.ContentService.Copy(IContent content, Int32 parentId, Boolean relateToOriginal, Boolean recursive, Int32 userId) in D:\a\1\s\src\Umbraco.Core\Services\Implement\ContentService.cs:line 2189
   at Umbraco.Web.Editors.ContentController.PostCopy(MoveOrCopy copy) in D:\a\1\s\src\Umbraco.Web\Editors\ContentController.cs:line 1716
```

### Steps to reproduce

- Install fresh 8.15.1 site with StarterKit
- Create element type. Add a single Textstring property.
- Create Block List data type. Add the element type we just created as possible block.
- On the Home document type (can be any), add a new property with the data type we just created.
- On the homepage, add a single block in our property and give it the title `[test]`. (some other test cases: `[test] test [test]` and `{test}`) Save.
- Try to create a copy of the homepage. Above error occurs.

Be able to copy the page without any errors.

---
_This item has been added to our backlog [AB#13401](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13401)_

nul800sebastiaan:
If our crude JSON detection gives us a false positive, don't throw an exception if it can't be parsed as JSON

Fixes #10879

### Prerequisites

### Description

To test, follow the scenario in the related issue #10879 and then test again with this fix applied. Note the message appears in the log as well.

I left the content type as a GUID as I didn't want to add overhead in trying to resolve the content type from the GUID to find the name of the content type just so it could be logged, that seemed a bit overkill.

### DIFF
--- a/src/Umbraco.Web/Compose/BlockEditorComponent.cs
+++ b/src/Umbraco.Web/Compose/BlockEditorComponent.cs
@@ -5,6 +5,7 @@ using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Umbraco.Core;
 using Umbraco.Core.Composing;
+using Umbraco.Core.Logging;
 using Umbraco.Core.Models.Blocks;
 using Umbraco.Core.PropertyEditors;
 
@@ -116,8 +117,23 @@ namespace Umbraco.Web.Compose
                             // this gets a little ugly because there could be some other complex editor that contains another block editor
                             // and since we would have no idea how to parse that, all we can do is try JSON Path to find another block editor
                             // of our type
-                            var json = JToken.Parse(asString);
-                            if (ProcessJToken(json, createGuid, out var result))
+                            JToken json = null;
+                            try
+                            {
+                                json = JToken.Parse(asString);
+                            }
+                            catch (Exception e)
+                            {
+                                // See issue https://github.com/umbraco/Umbraco-CMS/issues/10879
+                                // We are detecting JSON data by seeing if a string is surrounded by [] or {}
+                                // If people enter text like [PLACEHOLDER] JToken  parsing fails, it's safe to ignore though
+                                // Logging this just in case in the future we find values that are not safe to ignore
+                                Current.Logger.Warn<BlockEditorComponent>(
+                                    "The property {PropertyAlias} on content type {ContentTypeKey} has a value of: {BlockItemValue} - this was recognized as JSON but could not be parsed",
+                                    data.Key, propertyAliasToBlockItemData.Key, asString);
+                            }
+
+                            if (json != null && ProcessJToken(json, createGuid, out var result))
                             {
                                 // need to re-save this back to the RawPropertyValues
                                 data.RawPropertyValues[propertyAliasToBlockItemData.Key] = result;
--- a/src/Umbraco.Web/Compose/BlockEditorComponent.cs
+++ b/src/Umbraco.Web/Compose/BlockEditorComponent.cs
@@ -18,6 +18,17 @@ namespace Umbraco.Web.Compose
     {
         private ComplexPropertyEditorContentEventHandler _handler;
         private readonly BlockListEditorDataConverter _converter = new BlockListEditorDataConverter();
+        private readonly ILogger _logger;
+
+        [Obsolete("Use the ctor injecting dependencies.")]
+        public BlockEditorComponent() : this(Current.Logger)
+        {
+        }
+
+        public BlockEditorComponent(ILogger logger)
+        {
+            _logger = logger;
+        }
 
         public void Initialize()
         {
@@ -128,7 +139,7 @@ namespace Umbraco.Web.Compose
                                 // We are detecting JSON data by seeing if a string is surrounded by [] or {}
                                 // If people enter text like [PLACEHOLDER] JToken  parsing fails, it's safe to ignore though
                                 // Logging this just in case in the future we find values that are not safe to ignore
-                                Current.Logger.Warn<BlockEditorComponent>(
+                                _logger.Warn<BlockEditorComponent>(
                                     "The property {PropertyAlias} on content type {ContentTypeKey} has a value of: {BlockItemValue} - this was recognized as JSON but could not be parsed",
                                     data.Key, propertyAliasToBlockItemData.Key, asString);
                             }


src/Umbraco.Web/Compose/BlockEditorComponent.cs
```diff
@@ -116,8 +117,23 @@ private void RecursePropertyValues(IEnumerable<BlockItemData> blockData, Func<Gu
...
+                            {
+                                json = JToken.Parse(asString);
+                            }
+                            catch (Exception e)
+                            {
+                                // See issue https://github.com/umbraco/Umbraco-CMS/issues/10879
+                                // We are detecting JSON data by seeing if a string is surrounded by [] or {}
+                                // If people enter text like [PLACEHOLDER] JToken  parsing fails, it's safe to ignore though
+                                // Logging this just in case in the future we find values that are not safe to ignore
+                                Current.Logger.Warn<BlockEditorComponent>(
```
bergmania, line 131: I think you should inject the `ILogger` instead of using `Current`

nul800sebastiaan: @bergmania  Yeah I tried that but it's a breaking change 😢 
Unit tests then need to provide an `ILogger` as well and since it's a public class, people could be using it.

bergmania: I would still add a default ctor, just injection `Current.Logger` into the other ctor.. That's the way we normally handle such things..

bergmania:
```
        private readonly ILogger _logger;

        [Obsolete("Use the ctor injecting dependencies.")]
        public BlockEditorComponent() : this(Current.Logger)
        {
        }

        public BlockEditorComponent(ILogger logger)
        {
            _logger = logger;
        }
```

And then use `_logger` in the code

nul800sebastiaan: Ah.. that makes sense! Yes, that should be fine, I'll update!

nul800sebastiaan: That's now fixed! 👍

bergmania (approved):

Issue #10920, opened 2021-08-22
Umbraco 9.0.0-rc002: No warning but error when uploading files to Media that exceed maxAllowedContentLength


inetzo:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0-rc002

### Bug summary

When the a file is uploaded to Media, and the filesize exceeds the maxAllowedContentLength, there is no waring given. 
Instead `umbraco/backoffice/umbracoapi/media/PostAddFile` returns a error:

HTTP Error 413.1 - Request Entity Too Large
The request filtering module is configured to deny a request that exceeds the request content length.

### Specifics

Occures in Firefox 91 and Chrome 92

### Steps to reproduce

Upload an file to the Media section that exceeds maxAllowedContentLength. An error wil be shown in de console.

### Expected result / actual result

In Umbraco 8 you get an warning that the file exceeds maxAllowedContentLength before uploading. It would be nice to have this back again in Umbraco 9.

---
_This item has been added to our backlog [AB#13489](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13489)_

elit0451:
Hi @inetzo 
Thanks for reporting this issue! I've tried to check the warning that you mention you get in Umbraco 8 but couldn't really trigger it - and I tried different versions of Umbraco 8 which simply resulted in the same behaviour as in Umbraco 9 RC2 - the `Request Entity Too Large` error.

Could you specify in which exact Umbraco 8 version you experienced the _expected/actual_ result? Also, are you uploading the file through the File Drop Zone or through the **Create an item under Media** menu?

nul800sebastiaan:
Tried 5 different ways to upload as well in v8, all resulting in javascript errors. I'm curious where you've seen this message! 😅 

In any case, it would be awesome to handle the problem when it happens, instead of the error being hidden in the javascript console. We'd love a little help with that if you're up for it. We still merge up from v8 to v9, so a contribution for v8 would be great.

umbrabot:
Hi @inetzo,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

[...]

PR #10930, opened 2021-08-23
Fix BackOfficeServerVariables GetMaxRequestLength

inetzo:
_runtimeSettings.MaxRequestLength returned always null so the GetMaxRequestLength method returned always an empty string. 
Now it returns the MaxRequestBodySize in kilobytes (as a string).

### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10920

### Description

The Media upload dropzone did not show a warning when the filesize of the upload exceeded the max request length.
To test upload a file larger than 30 megabyte.

afbeelding

I'm pretty new to .NetCore so please double check if this works ok.

### DIFF
--- a/src/Umbraco.Web.BackOffice/Controllers/BackOfficeServerVariables.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/BackOfficeServerVariables.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Runtime.Serialization;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Options;
 using Umbraco.Cms.Core;
@@ -529,9 +530,7 @@ namespace Umbraco.Cms.Web.BackOffice.Controllers
             return app;
         }
 
-        private string GetMaxRequestLength()
-        {
-            return _runtimeSettings.MaxRequestLength.HasValue ? _runtimeSettings.MaxRequestLength.Value.ToString() : string.Empty;
-        }
+        private string GetMaxRequestLength() => _httpContextAccessor.HttpContext.Features.Get<IHttpMaxRequestBodySizeFeature>().MaxRequestBodySize != null ? (_httpContextAccessor.HttpContext.Features.Get<IHttpMaxRequestBodySizeFeature>().MaxRequestBodySize / 1024).ToString() : string.Empty;
+    
     }
 }
--- a/src/Umbraco.Web.BackOffice/Controllers/BackOfficeServerVariables.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/BackOfficeServerVariables.cs
@@ -529,8 +529,12 @@ namespace Umbraco.Cms.Web.BackOffice.Controllers
 
             return app;
         }
-
-        private string GetMaxRequestLength() => _httpContextAccessor.HttpContext.Features.Get<IHttpMaxRequestBodySizeFeature>().MaxRequestBodySize != null ? (_httpContextAccessor.HttpContext.Features.Get<IHttpMaxRequestBodySizeFeature>().MaxRequestBodySize / 1024).ToString() : string.Empty;
+        
+        private string GetMaxRequestLength() => _runtimeSettings.MaxRequestLength.HasValue ?
+            _runtimeSettings.MaxRequestLength.Value.ToString()
+            : _httpContextAccessor.HttpContext.Features.Get<IHttpMaxRequestBodySizeFeature>().MaxRequestBodySize != null ?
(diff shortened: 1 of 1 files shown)


elit0451:
Hi again @inetzo 😊 

Thanks for digging deeper into the issue! As you saw from the code we had, we are checking if the value for **MaxRequestLength** is set: `_runtimeSettings.MaxRequestLength.HasValue`. In order to set it in v9, please have a look at [this article](https://our.umbraco.com/documentation/Reference/V9-Config/RuntimeSettings/) - you need to add those settings into appsettings.json yourself.

There we also mention that _"If nothing is configured requests and query string can be any size."_ - or in other words an empty string (we don't put any limit). 

The solution you provided is using `KestrelServerLimits.MaxRequestBodySize` ([article](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserverlimits.maxrequestbodysize?view=aspnetcore-5.0)). The default value there is 30,000,000 bytes, which is approximately 28.6MB and since you divide by 1024, you get the value shown on the screenshot (29296KB). They also mention that _"When set to null, the maximum request body size is unlimited"_ - or again the value will be an empty string. 

So the real issue is client-side related, the empty string is not handled and therefore you get no message. 🙂  

As you said the issue is also present in the latest Umbraco 8.

inetzo:
Hi @elit0451 ,

Thanks for your explanation.
Wouldn't it be a better solution that, if MaxRequestLength is not set, which apparently is not done by default like in Umbraco 8, the KestrelServerLimits.MaxRequestBodySize is used instead of empty string? 
Any request greater than KestrelServerLimits.MaxRequestBodySize will alwyas result in an error (with no feedback to the user).

elit0451:
Hi @inetzo and sorry for the late reply.

So, you are right that in v8 the value is set by default in the Web.config file https://github.com/umbraco/Umbraco-CMS/blob/7800fe83a8978440be9debc6800d1403f8fdfc0b/src/Umbraco.Web.UI/web.Template.config#L90 


and then in `GetMaxRequestLength`: https://github.com/umbraco/Umbraco-CMS/blob/51fcde5fcdde68b5bed3b9420cfd92e3b29cc70c/src/Umbraco.Web/Editors/BackOfficeServerVariables.cs#L480-L485
we do pretty much the same as we have in v9.

As we mention in our documentation, we recommend people to manually set it in the new configuration file, otherwise,  _"If nothing is configured reqests and query string can be any size."_ ([link](https://our.umbraco.com/documentation/Reference/V9-Config/RuntimeSettings/)). Setting it up to a "fall-back" value (if no value is configured) doesn't really solve the real front-end issue, it simply hides it 🤷

elit0451:
The reason why no "Max file size" error is not shown when a file with a greater size than the max specified is uploaded is because the `files` array passed into the function below is empty in this scenario
https://github.com/umbraco/Umbraco-CMS/blob/be3545ce4a18a332c243d5dfebd1691faf70b1eb/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbfiledropzone.directive.js#L245-L249

And therefore cannot be added to the `rejected` pool which I think is then used to show the validation error we want

https://github.com/umbraco/Umbraco-CMS/blob/be3545ce4a18a332c243d5dfebd1691faf70b1eb/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbfiledropzone.directive.js#L65-L77

inetzo:
Hi @elit0451,

What I still don't understand is why MaxRequestLength should be set in Runtime settings (and maxRequestLength in V8 in web.config), while the technically possible/maximum upload size is determined by KestrelServerLimits.MaxRequestBodySize. An upload can technically never be greater than this value.

Setting MaxRequestLength in Runtime settings only makes sense in my opinion if you want to limit the upload size.

I think that it might be a good idea to check if MaxRequestLength is set, and if so if it is less than or equal to KestrelServerLimits.MaxRequestBodySize (in KB). If it is not set or larger than the KestrelServerLimits.MaxRequestBodySize  (which is the technical limit), KestrelServerLimits.MaxRequestBodySize could be returned.

elit0451:
Hi again @inetzo 

You are right that in the current state of the code, our upload limit was determined by `KestrelServerLimits.MaxRequestBodySize` since if we configured a value was bigger than the default 30,000,000 bytes, we were not setting the MaxRequestBodySize to null which will enable unlimited size (_"When set to null, the maximum request length will not be restricted in ASP.NET Core"_ - [link](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserverlimits.maxrequestbodysize?view=aspnetcore-5.0)).

I've made a PR that will keep the values updated based on what we configure in Runtime settings #11052 
But there is still the UI issue I mentioned here that we need to investigate further what has changed for v9 and why is the Max file size error not appearing.

Hope it makes sense!

bergmania:
This should be fixed here https://github.com/umbraco/Umbraco-CMS/pull/11052

PR #10935, opened 2021-08-24
Avoid generating the maindom hash twice to get the Id.

nzdev:
### Prerequisites

### Description
Store and reuse the MainDomId, avoids having to recalculate the Hash.


nul800sebastiaan:
Nice one! 👍

PR #10936, opened 2021-08-24
Avoid allocating an array every call to IsClientSideRequest()

nzdev:
### Prerequisites

### Description
IsClientSideRequest() is called multiple times per request and allocates an array, then loops through all the entries in the array to do an Invariant string comparison.

This PR changes that to an invariant string dictionary lookup to avoid the array allocations.

### DIFF
--- a/src/Umbraco.Core/UriExtensions.cs
+++ b/src/Umbraco.Core/UriExtensions.cs
@@ -141,9 +141,9 @@ namespace Umbraco.Core
         /// <summary>
         /// Non Client Side request Extensions <see cref="IsClientSideRequest"/>
         /// </summary>
-        internal readonly static IDictionary<string, bool> NonClientSideRequestExtensions = new Dictionary<string, bool>(5,StringComparer.InvariantCultureIgnoreCase)
+        internal readonly static HashSet<string> NonClientSideRequestExtensions = new (5, StringComparer.InvariantCultureIgnoreCase)
         {
-            { ".aspx",true },{ ".ashx",true},{ ".asmx",true},{ ".axd",true}, {".svc",true}
+            ".aspx", ".ashx", ".asmx", ".axd", ".svc"
         };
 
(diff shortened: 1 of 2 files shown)


src/Umbraco.Core/UriExtensions.cs
```diff
@@ -137,6 +138,14 @@ internal static bool IsDefaultBackOfficeRequest(this Uri url, IGlobalSettings gl
             return false;
         }
 
+        /// <summary>
+        /// Non Client Side request Extensions <see cref="IsClientSideRequest"/>
+        /// </summary>
+        internal readonly static IDictionary<string, bool> NonClientSideRequestExtensions = new Dictionary<string, bool>(5,StringComparer.InvariantCultureIgnoreCase)
```
abjerner, line 144: @nzdev feel free to correct me if I'm wrong, but wouldn't a `HashSet<string>` be even better here? Since all values in the dictionary are `true`, it seems like we only really need the key.

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.-ctor?view=netframework-4.7.2#System_Collections_Generic_Dictionary_2__ctor_System_Int32_System_Collections_Generic_IEqualityComparer__0__

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.-ctor?view=netframework-4.7.2#System_Collections_Generic_HashSet_1__ctor_System_Int32_System_Collections_Generic_IEqualityComparer__0__

nzdev: Makes sense to me

nul800sebastiaan: Thanks for the feedback @abjerner - I've applied that change now.

nul800sebastiaan:
Good stuff @nzdev - I updated to `HashSet<string>` as suggested, merging this now. 👍

PR #10940, opened 2021-08-24
Remove restricted image size when opening media

bjarnef:
### Prerequisites

### Description
With Media Picker 3 and opening e.g. a SVG image (which doesn't have dimensions) I noticed the image preview is quite small (limited to max 200px with and 200px height).

**Before**

image

**After**

image


bjarnef:
With a folder selected it didn't look great either.

**Before**

image

**After**

nathanwoulfe (approved):
All good at my end, thanks @bjarnef

PR #10942, opened 2021-08-24
Added url mode parameter to the GetCropUrl extension methods

lauraneto:
### Prerequisites

### Description
I've added a new parameter `urlMode` to the image cropper `.GetCropUrl()` extension methods in order to be able to select which url mode I need in that specific call, like it works now for `.Url()`.
In my case, I have the default url mode set to relative, but for things like the Open Graph image, I need it to be absolute.

I tested my changes by calling some of the extension methods from a view as follows

```
@Model.Image.GetCropUrl(1200, 620)
@Model.Image.GetCropUrl(1200, 620, urlMode: UrlMode.Absolute)
@Model.Image.GetCropUrl("crop1")
@Model.Image.GetCropUrl("crop1", UrlMode.Absolute)
```


Zeegaan:
Thank you for your contribution, and congrats on your first PR 👍 

I tested this both with the legacy Media Picker and the new one, and they both work 💪 

Code looks good to me!

PR #10946, opened 2021-08-25
V9: Reuse value editors for validation

nikolajlauridsen:
This PR introduces a ValueEditorCache, this cache will be used by the `PropertyValidationService`, meaning that we no longer create a new `valueEditor` every time something needs to be validated.  This should optimize validation for all data types, but in particular for nested content and the blocklist editor, since these calls validate on their children recursively.

To ensure that the value editors are up to date I've also added a `ValueEditorCacheRefresher` which listens to `DataTypeSavedNotification` and `DataTypeDeletedNoticiation`, and clears the cache accordingly. 

### Testing

Since everything should be the same it can be kind of hard to test, but a good thing to test is to try and create a document type with some custom data types like blocklist and nested content, create a content node, and then try and update the custom data types and ensure that the validation gets updated as well. A handy tip is to place a breakpoint in `ValueEditorCache.GetValueEditor` and verify whether a new value editor gets created or not.

### DIFF
--- a/src/Umbraco.Core/Cache/IValueEditorCache.cs
+++ b/src/Umbraco.Core/Cache/IValueEditorCache.cs
@@ -1,0 +1,10 @@
+﻿using Umbraco.Cms.Core.Models;
+using Umbraco.Cms.Core.PropertyEditors;
+
+namespace Umbraco.Cms.Core.Cache
+{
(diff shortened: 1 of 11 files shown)


Zeegaan:
Tested this PR and it works! 👍 

Code looks good to me 💪

PR #10948, opened 2021-08-25
Fixed potential null reference exception in KeepAlive

bergmania:
Fixed potential null reference exception in KeepAlive


nikolajlauridsen (approved):
Looks good 👍

PR #10958, opened 2021-08-26
Do not use the guid when calling the breadcrumbs to get ancestors on …

### LINKED ISSUE #10950: Open In Media Library link on media view blade causes an exception on load

V9-RC002

### Bug summary

Content Structure:
Home[Homepage Blocks][Thumbnail Image Block][Image]-> Click image to open blade, Image doesn't load properly for certain files, but also if at this point click Open In Media Library link, attached error appears
image

Although the image does appear as even regardless of the error which seems to be trying to find ancestors, but is expecting id to be an int when a guid is being provided instead. This does not appear to be a regression from RC1->RC2 & existed before RC1 as I tried the older code to no avail.
Example link:
https://localhost:44362/umbraco/backoffice/umbracoapi/entity/GetAncestors?id=5ae6fb5e-902e-4b9f-adb2-8c293a1f24fa&type=media&culture=

### Specifics

_No response_

### Steps to reproduce

Create Doc Type
Add Property -> Image Media Picker
Create Content
Select image
Save & publish
Click title of the image displayed on the thumbnail
Click Open In Media Library on the blade displayed

Media is loaded without errors popping up

---
_This item has been added to our backlog [AB#13527](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13527)_

jaddie:
…media items for the library blade, fixes #10950

This fixes https://github.com/umbraco/Umbraco-CMS/issues/10950

1. Create content type with media picker property
2. Select media item
3. Go back to item, item title to open slide out view
4. Press open in media library -> error occurs due to breadcrumbs error for getting ancestors due to GUID being used not ID

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/media/media.edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/media/media.edit.controller.js
@@ -256,7 +256,7 @@ function mediaEditController($scope, $routeParams, $location, $http, $q, appStat
 
                 if ($scope.content.parentId && $scope.content.parentId !== -1 && $scope.content.parentId !== -21) {
                     //We fetch all ancestors of the node to generate the footer breadcrump navigation
-                    entityResource.getAncestors(nodeId, "media")
+                    entityResource.getAncestors(data.id, "media")
                         .then(function (anc) {
                             $scope.ancestors = anc;
                         });


bjarnef:
Maybe we just need to support `Guid` in `GetAncestors()` method:
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Editors/EntityController.cs#L756-L759

https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/Editors/EntityController.cs#L75-L79

There's a few other open PRs fixing some missing parts not supporting fetching entities by `Guid` and `Udi´:
https://github.com/umbraco/Umbraco-CMS/pull/9465
https://github.com/umbraco/Umbraco-CMS/pull/9467

jaddie:
Yes @bjarnef that is what I was suggesting would potentially be the preferred long term solution, however those PR's do not implement that from what I can see the GetById method already supports getting by the GUID, which in the context of this issue, when that is done the response then allows access to the data variable which is populated with the properties of the content item, one of which is id -- the numeric int ID for the item, which when used as above in the JS, fixes this bug, however I agree there should probably be an overload created that provides support for using a GUID

jaddie:
@nul800sebastiaan What is your opinion on this? Do you think perhaps merge this PR for the interim & then look to create another PR that implements a fix for this by adding a getancestors method overload that supports a GUID which will be for the long term?

bjarnef:
I tried implemented this, which almost worked, but because of the `type` parameter og failed with multiple endpoints with the signature although `id` parameter should swap. Alternatively er could name to method different by id, guid and udi.

jaddie:
Sorry I am not quite sure I follow your message, so you are saying you added an overload for GetAncestors with a GUID it broke? I am not sure I see why, it should be able to pick the difference between them based on the input data type, what errors did you have?

bjarnef:
@jaddie I have submitted this draft PR https://github.com/umbraco/Umbraco-CMS/pull/10975 but it throw an error that mulitple actions were found for the request because of the `type` parameter in ` GetAncestors()` method.

nul800sebastiaan:
@jaddie - as noted, the issue has been fixed in the mean time but we're happy to see contributions, so keep them coming in the future! 👍

PR #10959, opened 2021-08-26
V8: Fix IsPublished only working with lowercase.

### LINKED ISSUE #10695: IsPublished only works with lowercase cultures (8.15)

8.15

### Bug summary

When using .IsPublished(culture) on a multi-language website, it only works when the culture has been set to lower case.

I think this has something to do with the MsgPack change or just the 8.15 update in general, as we haven't run into this issue before. I can also reproduce this issue on the latest version of the v8/contrib branch (which is still 8.15-rc).

Model.IsPublished("en-US") will return false, while Model.IsPublished("en-us") will return true.
image

### Specifics

_No response_

### Steps to reproduce

- Start with the starter kit
- Change the homepage document type so that it is multi-language. You don't have to change any of the properties
- Update the home.cshtml to show the following:
```
Published with culture: @Model.IsPublished(Culture)
<br />
Published with en-US: @Model.IsPublished("en-US")
<br />
Published with lower case en-US: @Model.IsPublished("en-us")
```
- Go to the homepage. You should now see that all 3 are shown with true. It could be that they aren't already, but I think this has something to do with the cache.
- Publish the homepage
- The first two will return false, while the last one will return true. Even though all 3 should be returning true.

IsPublished() should return true even if the culture is not all lower-case

---
_This item has been added to our backlog [AB#13193](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13193)_

nikolajlauridsen:
This fixes #10695 

Turns out the issue was not with MessagePack, but instead with the JsonConverter used when saving the item to the cache after when updating it.

This is also why it seemingly worked the first time after booting, but once you've saved and published it, it stopped working. The fix is fairly straightforward, instead of creating a new dictionary normally, we call `Create` on the parent class which creates a dictionary with `StringComparer.OrdinalIgnoreCase`, making the dictionary case insensitive. 

## Testing 

* Create a document type and make it multi-language 
* Add the following to the template (thanks to the original issue create 😄 )

```html
Published with culture: @Model.IsPublished(Culture)
<br />
Published with en-US: @Model.IsPublished("en-US")
<br />
Published with lower case en-US: @Model.IsPublished("en-us")
```

* Make sure you save & publish the content node
* Ensure that they all output true
* Add `<add key="Umbraco.Web.PublishedCache.NuCache.Serializer" value="JSON"/>` to your `web.config` and ensure that the results are still the same (remember to save & publish)

### DIFF
--- a/src/Umbraco.Core/Serialization/AutoInterningStringKeyCaseInsensitiveDictionaryConverter.cs
+++ b/src/Umbraco.Core/Serialization/AutoInterningStringKeyCaseInsensitiveDictionaryConverter.cs
@@ -7,7 +7,7 @@ using Newtonsoft.Json.Converters;
 namespace Umbraco.Core.Serialization
 {
     /// <summary>
-    /// When applied to a dictionary with a string key, will ensure the deserialized string keys are interned 
+    /// When applied to a dictionary with a string key, will ensure the deserialized string keys are interned
     /// </summary>
     /// <typeparam name="TValue"></typeparam>
     /// <remarks>
@@ -26,7 +26,7 @@ namespace Umbraco.Core.Serialization
         {
             if (reader.TokenType == JsonToken.StartObject)
             {
-                var dictionary = new Dictionary<string, TValue>();
+                var dictionary = Create(objectType);
                 while (reader.Read())
                 {
                     switch (reader.TokenType)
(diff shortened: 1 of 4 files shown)


Zeegaan:
Tested this solution with the very helpful template code 👍  

Code looks good to me 💪

PR #10963, opened 2021-08-26
Falling back to contentTypeName when Block List label is empty

callumbwhyte:
I'm a big fan of the Block List Editor! And especially love how powerful the label templating feature is, allowing us to give our editors more context. Where a block content has a title / heading property we usually opt for this as our label.

However, sometimes properties are optional... and therefore the label can be empty...

The way to combat this is to have a condition in the label template `{{title || 'My Block Name'}}` but this means hardcoding content type names...

This PR adds a null check around the result of the label interpolator, and if there is no output falls back to the default value: the block list content type name.

There are some cases where hardcoding the content type names is necessary, such as `Block name: {{title}}`. This PR also introduces a variable to easily access the name in label templates: `$contentTypeName`

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
@@ -103,7 +103,7 @@
          */
         function getBlockLabel(blockObject) {
             if (blockObject.labelInterpolator !== undefined) {
-                var labelVars = Object.assign({"$settings": blockObject.settingsData || {}, "$layout": blockObject.layout || {}, "$index": (blockObject.index || 0)+1 }, blockObject.data);
+                var labelVars = Object.assign({"$contentTypeName": blockObject.content.contentTypeName, "$settings": blockObject.settingsData || {}, "$layout": blockObject.layout || {}, "$index": (blockObject.index || 0)+1 }, blockObject.data);
                 var label = blockObject.labelInterpolator(labelVars);
                 if (label) {
                     return label;
--- a/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/blockeditormodelobject.service.js
@@ -97,7 +97,6 @@
             }
         }
 
-
         /**
          * Generate label for Block, uses either the labelInterpolator or falls back to the contentTypeName.
          * @param {Object} blockObject BlockObject to recive data values from.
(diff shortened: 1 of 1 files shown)


FransdeJong:
@callumbwhyte 
I think I see a issue with this solution.

We name our blocks with the content block name and if the value we are looking for isn't empty we add "- block title"

For example: Text block - text block one.

The reason we do this is that the editor can always see the type of block and if the optional value is filled it is visible for them to.

I think with your solution what we will end up with is: Text block - Text block

callumbwhyte:
@FransdeJong Not sure I understand the issue...

If you have a value in the "label" field that returns something then that will be used, no changes there.

In your case, it sounds like you have a label template like this: `{{blockTitle ? 'Block name - ' + blockTitle : 'Block name'}}`?

This change won't affect your case. You could even adjust it to `{{blockTitle ? 'Block name - ' + blockTitle : null}}` if you wanted to avoid hardcoding the block name.

FransdeJong:
We had something like this: Block Name {title ? '- ' + title : ''}
But I can make it work with your example. I have to refactor the old projects when we update.

callumbwhyte:
In which case that should keep working! 😊 As your label will always have "Block Name" in it then that will always be returned.

Something I considered adding to this PR was a variable for the contentTypeName, e.g. `$contentTypeName`, which could be used in your case to avoid hardcoding the name... Thoughts?

FransdeJong:
In my case that would be brilliant. That way I can work with a default code in that field for every block.

I can look at it with you after I return from my holiday if you like

callumbwhyte:
@FransdeJong Consider it done, just added to this PR!

nathanwoulfe:
Looks good to me, have tested with the following (block has a property aliased featureName):
- label: `{{ featureName }} - {{ $contentTypeName }}` => renders featureName value and content type name
- label: `{{ foo }}` => renders content type name as no value exists for foo
- label: `{{ featureName }} - foo` => renders featureName value and foo
- label: empty => renders content type name
- label per @FransdeJong example: `Block Name {{ featureName ? '- ' + featureName : '' }}` => renders correctly when featureName has a value, and renders `Block Name` when featureName is null

Unless I'm missing any other valid test cases, this is good to go.

nathanwoulfe (approved):

callumbwhyte:
Just wondering what the process is for also getting this merged up to V9? 🙏

nul800sebastiaan:
Everything in v8 will be merged up to v9 still for the foreseeable future. :-) This didn't make it for the v9 final though, that one is in sync with 8.17. So I think it will be 9.1.0.

PR #10964, opened 2021-08-26
Use ImageSharps Image.Identify for dimension extraction

ronaldbarendse:
### Prerequisites

### Description
This replaces the `Image.Load(stream)` call in the `IImageDimensionExtractor` implementation with `Image.Identify(stream)`, which is more memory and CPU efficient: https://docs.sixlabors.com/articles/imagesharp/imageformats.html#metadata-only-decoding.

I've also cleaned up related bits, notably:
- Removing `ExifImageDimensionExtractor`, as the same is already done by ImageSharp
- Removing `ImageSize` struct, as we can use the built-in `Size` struct
- Move the default size fallback outside of the `IImageDimensionExtractor` implementation

### Testing
The changes impact the automatic population of the `umbracoWidth` and `umbracoHeight` properties (part of the built-in Image media type):
- Upload an image with a valid file extension (e.g. `.jpg`) to the media section: the properties should reflect the correct image size;
- Remove the file from the media item and upload a dummy text file with some contents and the same file extension as the image (e.g. `.jpg`): the default sizes of 200x200 should be set, as the file extension is a valid image extension, but it can't parse the dimensions from the 'image';
- Remove the file again and upload the dummy text file with a `.txt` extension: the dimensions should be cleared, as it's not a valid image extension.


src/Umbraco.Core/Media/UploadAutoFillProperties.cs
```diff
@@ -74,8 +75,11 @@ public void Populate(IContentBase content, ImagingAutoFillUploadField autoFillCo
                     {
                         using (var filestream = _mediaFileManager.FileSystem.OpenFile(filepath))
                         {
-                            var extension = (Path.GetExtension(filepath) ?? "").TrimStart(Constants.CharArrays.Period);
-                            var size = _imageUrlGenerator.IsSupportedImageFormat(extension) ? (ImageSize?)_imageDimensionExtractor.GetDimensions(filestream) : null;
+                            var extension = (Path.GetExtension(filepath) ?? string.Empty).TrimStart(Constants.CharArrays.Period);
+                            var size = _imageUrlGenerator.IsSupportedImageFormat(extension)
+                                ? _imageDimensionExtractor.GetDimensions(filestream) ?? (Size?)new Size(Constants.Conventions.Media.DefaultSize, Constants.Conventions.Media.DefaultSize)
+                                : null;
+
```
nikolajlauridsen, line 82: I think we should put this into a private method since it's used again in the other overload, DRY and all that 😉

src/Umbraco.Infrastructure/Media/ImageSharpDimensionExtractor.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <param name="stream">A stream containing the image bytes.</param>
+        /// <returns>
+        /// The dimension of the image.
+        /// </returns>
+        public Size? GetDimensions(Stream stream)
+        {
+            if (stream.CanRead && stream.CanSeek)
+            {
+                stream.Seek(0, SeekOrigin.Begin);
+            }
```
nikolajlauridsen, line 22: Not sure, but here if we can't read and seek the stream, wouldn't it make more sense to just return null, instead of still passing it to `Image.Identify`, or is image sharp doing some wild voodoo magic to still extract the data?

nikolajlauridsen (commented):
Code looks good overall, but found a few very minor things, still need to do some testing though.

nikolajlauridsen:
Tests out good as well, tried with different media types, png, jpg, gif, and all seems to work as expected 👍.

src/Umbraco.Core/Media/UploadAutoFillProperties.cs
ronaldbarendse: We should indeed keep it DRY 👍🏻

src/Umbraco.Infrastructure/Media/ImageSharpDimensionExtractor.cs
ronaldbarendse: I noticed the `Image.Identify` didn't use the configuration that's registered in the DI (required if the configuration is altered with additional image formats, etc.):

https://github.com/umbraco/Umbraco-CMS/blob/061f1b3b879f88f327a2fb215aa937811be9cbcf/src/Umbraco.Infrastructure/DependencyInjection/UmbracoBuilder.CoreServices.cs#L185

ImageSharp already contains a setting to automatically seek to the beginning if you want: [`Configuration.ReadOrigin`](https://docs.sixlabors.com/api/ImageSharp/SixLabors.ImageSharp.Configuration.html#SixLabors_ImageSharp_Configuration_ReadOrigin). The default configuration uses `ReadOrigin.Current`, but within the CMS we only use a newly created stream, so the position will already be at the beginning anyway.

nikolajlauridsen (approved):
Looks good now, and still tests out fine 👍 

There were a very simple merge conflict, I took the liberty of fixing it up, and will merge once the build servers are happy 😄

Issue #10943, opened 2021-08-24
Media entry editor doesn't show child items


bjarnef:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

8.16.0

### Bug summary

When browsing a folder via Media Picker 3 it shows child items app and list view, but nothing is loaded although the folder has several children.

If not supported it should probably either hide "Child items" content app or show a message. In that case it should probably also hide listview search + layout selector. 

### Steps to reproduce

- Browse a folder in media section or media picker and confirm the folder has children.

![image](https://user-images.githubusercontent.com/2919859/130662565-d67f10a4-d31b-4f59-b19d-90cebf67f87c.png)

![image](https://user-images.githubusercontent.com/2919859/130662628-708f4dd4-93cc-4889-a124-a1adbef41b36.png)

- Select a folder in Media Picker 3 and afterwards browse this folder via " Open in Media Library" button and notice child items content app it shown, but no children are loaded.

![image](https://user-images.githubusercontent.com/2919859/130662851-3a7985f2-2511-4fe8-8569-e8835da2eb45.png)

![image](https://user-images.githubusercontent.com/2919859/130663039-c0d3f9e5-bf9a-42dd-98cf-ee9f90520ee8.png)

umbrabot:
Closing this in relation to PR #10966 - make sure to read the close reason on #10966

PR #10966, opened 2021-08-27
Show listview children when opening media entry editor

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/10943

### Description
When opening media from media entry editor via the new media picker 3, it didn't show the children in the listview, because it used id from `$routeParams` which in this case was the current content node it and not the media node id the load it children from in listview.

Unlike other places in code we here don't have access to `infiniteMode` property directly from `$scope.model`, e.g.

```
var infiniteMode = $scope.model && $scope.model.infiniteMode;
$scope.contentId = infiniteMode ? $scope.model.id : id;
```

It seems we can get the id from `editorState.current` though.

wtgFL5ipLX


bjarnef:
@nathanwoulfe maybe you want the review this? It seems like an oversight it doesn't ret id from `editorState`.

PR #10968, opened 2021-08-27
Obsolete IUserComposer and rename Composers

p-m-j:


src/Umbraco.Core/Composing/IUserComposer.cs
```diff
@@ -1,11 +1,9 @@
-﻿namespace Umbraco.Cms.Core.Composing
+namespace Umbraco.Cms.Core.Composing
 {
     /// <summary>
     /// Represents a user <see cref="IComposer"/>.
     /// </summary>
-    /// <remarks>
-    /// <para>User composers compose after core composers, and before the final composer.</para>
-    /// </remarks>
+    [System.Obsolete("This interface is obsolete. Use IComposer instead.")]
```
ronaldbarendse, line 6: Let's add a `using System;` above and update to a more helpful message:
```suggestion
    [Obsolete("Composing is always done after core services are configured during startup, use IComposer instead.")]
```

We might want to explain that `ICoreComposer` is removed/replaced with the native DI container... The above message is already quite long, so it should be described more compactly or otherwise be added as remarks/XML documentation.

ronaldbarendse (changes requested):

src/Umbraco.Core/Composing/ComposerGraph.cs
```diff
@@ -48,8 +45,8 @@ public Composers(IUmbracoBuilder builder, IEnumerable<Type> composerTypes, IEnum
 
         private class EnableInfo
```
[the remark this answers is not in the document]
p-m-j: I think that would be a good suggestion if Enabled/Disabled were ever actually used which I'm not convinced they are.

Goals for this PR were simply 
1. Rename Composers -> ComposerGraph 
1. Mark obsolete IUserComposer

Further effort in cleaning up this class is probably better spent moving towards the pipeline approach on umbraco builder alluded to in https://github.com/umbraco/Umbraco-CMS/discussions/10900#discussioncomment-1238584

I only changed these lines because the field names were causing linting errors but I probably should have left them as they were.

ronaldbarendse (approved):

src/Umbraco.Core/Composing/ComposerGraph.cs
```diff
@@ -9,36 +9,33 @@
...
 namespace Umbraco.Cms.Core.Composing
 {
-    // note: this class is NOT thread-safe in any ways
+    // note: this class is NOT thread-safe in any way
 
     /// <summary>
     /// Handles the composers.
     /// </summary>
-    public class Composers
+    public class ComposerGraph
```
bergmania, line 17: Wonder if we should do this internal. I see no reason users of Umbraco should use this.

bergmania (approved):

src/Umbraco.Core/Composing/ComposerGraph.cs
ronaldbarendse: And renaming a public class is a breaking change anyway... I could see this get used by tests to check the enabled status and order of your own composers, but because this can be influenced by any external assembly (and its attributes), those tests will be of minimal use anyway.

PR #10970, opened 2021-08-27
Devcontainer: Update SMTP FROM address

iOvergaard:
Updated the SMTP "from" email address in docker-compose.yml from `warren-env` to a noreply email address.


iOvergaard:
@p-m-j I'm a little in over my head here. According to [this reply on SO](https://stackoverflow.com/a/38017715) `* text=auto` is not recommended, but should instead be set for particular file types instead. I see that we have no setting for .yml files, so perhaps it should be enough to add that?

nul800sebastiaan:
Ha! I love that we have one for `.php` 😉, but yeah, would be good to add `.yml`! 

> I just tried to make sure the file is definitely using CRLF and Github now still shows changes to all lines. So I think that perhaps this file was not using CRLF line breaks before. 

Aha! Makes sense! I'll have a look.

p-m-j:
IMHO `* text=auto` is a sensible default that can be customized per file type easily afterwards.

https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings

nul800sebastiaan:
I've found what most people use for yml files and applied that, seems to work nicely and now we'll know that all future yml changes won't change the leading whitespace for those files. 

image

Adding `* text=auto` doesn't seem to give immediate problems so I think that's fine to start with and then we overrule with later entries. I've added that as well.

nul800sebastiaan:
Thanks very much @iOvergaard for your (first!) contribution to Umbraco-CMS!

I've updated your Our Umbraco profile with a shiny new contributor badge: https://our.umbraco.com/members/id:277819/ 🏅 

Hope to see more contributions from your hand in the future. 👍

iOvergaard:
This is a good way of handling it. I like it, @nul800sebastiaan !

PR #10971, opened 2021-08-27
Check if umbFileDropzone has queued items before showing ChooseMediaTypeDialog

inetzo:
The ChooseMediaTypeDialog is shown even if there are no queue items, for example due file errors. 
Checking the queue lenght in _requestChooseMediaTypeDialog prevents showing the dialog unnecessary.

### Prerequisites

### Description

When a user uploads an file that exceeds the maxFileSize an error is shown but also the ChooseMediaTypeDialog.
Checking the queue length before showing prevents this.

To test: upload an file to media via the dropzone that exceeds the maxFileSize.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbfiledropzone.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbfiledropzone.directive.js
@@ -187,6 +187,11 @@ angular.module("umbraco.directives")
 
                     function _requestChooseMediaTypeDialog() {
 
+                        if (scope.queue.length === 0) {
+                            // if queue has no items so there is nothing to choose a type for
+                            return false;
+                        }
+                        
                         if (scope.acceptedMediatypes.length === 1) {
                             // if only one accepted type, then we wont ask to choose.
                             return false;


nul800sebastiaan:
@inetzo Did you mean to target `v9/dev` with this one? In v8, this already seems to work as expected, unless I'm missing something?

image

inetzo:
Hi @nul800sebastiaan , this occures in both V8 and V9,I guess since the new media types are introduced. This is an other issues than the 'Max file size' (https://github.com/umbraco/Umbraco-CMS/pull/10930) but also related to the File Dropzone. Thats why i've mentioned this fix in the same issue (https://github.com/umbraco/Umbraco-CMS/issues/10920) but created two separate PR's.

nul800sebastiaan:
Aha.. I was testing on a slightly older version I guess, I see what is happening now in 8.16. I'll test this PR again!

By the way, we still merge up from v8 to v9, so anything added to v8 will also make it to v9.

nul800sebastiaan:
Alright, I was having a really tough time testing this for some reason, but got there in the end!

The current behavior in v8.16 is: when you change `maxRequestLength` to something small like 500 kb you can upload the file but then suddenly you get the file type selector as shown in your screenshot. Even though the error is already showing behind the modal.

If we apply your PR, then do the same thing, the modal is not shown after uploading a file that's too large and you get to see the error message as expected.

image

So that's perfect! I'll merge this one, it will be part of 8.17 and should also make it to the next RC of v9! 👍 

Thanks for the help!

PR #10973, opened 2021-08-27
Fix self-closing tags which should have a end-closing tag

bjarnef:
### Prerequisites

### Description
I searched view files and found these references where mostly `<localize>` directive was self-closed, but since it is a custom HTML element it isn't a valid void-element and thus it should have a end-closing tag.


nul800sebastiaan:
Thanks for the help @bjarnef! 👍

PR #10975, opened 2021-08-29
Get ancestors by Guid and Udi

bjarnef:
### Prerequisites

### Description
This PR is related to https://github.com/umbraco/Umbraco-CMS/pull/10958 and an alternative fix for https://github.com/umbraco/Umbraco-CMS/issues/10950 which support Guid and Udi as well.

E.g find a media node preferable at level 2 or deeper, copy id and guid and use these in a controller like a dashboard controller in content section.

```
entityResource.getAncestors("1134", "media")
    .then(function (anc) {
        console.log("getAncestors (id)", anc);
    });

entityResource.getAncestors("1bc5280b-8658-4027-89d9-58e2576e469b", "media")
    .then(function (anc) {
        console.log("getAncestors (guid)", anc);
    });

entityResource.getAncestors("umb://1bc5280b-8658-4027-89d9-58e2576e469b", "media")
    .then(function (anc) {
        console.log("getAncestors (udi)", anc);
    });
```

The requests except the `int` id will fail.

chrome_o1RaN8H4di

However with the changes in this PR we get the error, because of the second parameter `type`:

image

Not sure if there a way to fix this? Alternative the other methods could by renamed, e.g. `GetAncestorsByGuid()`, `GetAncestorsByUdi()` and maybe `GetAncestorsById()` and obsolete the current `GetAncestors()` method.

We could probably also refactor the `GetResultForAncestors()` overload method a bit, so the shared logic is moved to a separate method.

### DIFF
--- a/src/Umbraco.Web/Editors/EntityController.cs
+++ b/src/Umbraco.Web/Editors/EntityController.cs
@@ -76,7 +76,9 @@ namespace Umbraco.Web.Editors
                     new ParameterSwapControllerActionSelector.ParameterSwapInfo("GetPath", "id", typeof(int), typeof(Guid), typeof(Udi)),
                     new ParameterSwapControllerActionSelector.ParameterSwapInfo("GetUrlAndAnchors", "id", typeof(int), typeof(Guid), typeof(Udi)),
                     new ParameterSwapControllerActionSelector.ParameterSwapInfo("GetById", "id", typeof(int), typeof(Guid), typeof(Udi)),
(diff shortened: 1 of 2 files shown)


bjarnef:
The udi should be this instead:

```
entityResource.getAncestors("umb://media/1bc5280b8658402789d958e2576e469b", "media")
        .then(function (anc) {
            console.log("getAncestors (udi)", anc);
        });
``` 

which result in this error instead.

image

bjarnef:
Ahh of course, it shouldn't use array, but just `int`, `Guid` and `Udi`. 🤦‍♂️

Fixed in https://github.com/umbraco/Umbraco-CMS/pull/10975/commits/ee2c3abda2935b77bb17bf956a9546845b6cc29a

It now works.

image

PR #10978, opened 2021-08-30
Grid: Add button styling fix

BatJan:
### Prerequisites

### Description
With the current release of version 8.16 I noticed the "Add button" in the grid was no longer a perfect square after the icon was converted to using the `<umb-icon>` directive. That's addressed with a few tweaks to the `.iconBox` class in this PR.

Furthermore some unused classes and uneeded CSS properties have been removed 😃 

**Before**
plus-button-before

**After**
plus-button-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/less/components/umb-grid.less
+++ b/src/Umbraco.Web.UI.Client/src/less/components/umb-grid.less
@@ -565,12 +565,11 @@
 // ICONS
 // -------------------------
 .umb-grid .iconBox {
-    padding: 4px 6px;
-    display: inline-block;
-    cursor: pointer;
+    padding: 6px;
+    display: flex;
     border-radius: 200px;
     border: 1px solid @ui-action-discreet-border;
-    margin: 2px;
+    margin: 0 auto;
 
     &:hover, &:hover * {
         background: @ui-action-discreet-type-hover !important;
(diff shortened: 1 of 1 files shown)


bjarnef:
Another thing that is redundant is setting `border-color` here and then override it right after on next line 😅
https://github.com/umbraco/Umbraco-CMS/blob/4eed7d6a56bce391c10d26847e3f6e2a88623177/src/Umbraco.Web.UI.Client/src/less/components/umb-grid.less#L601-L602

BatJan:
@bjarnef Ah yes, removed that as well now 👍🏻 All the browsers that Umbraco aim to support has rgba() support for a long time so no worries about that one 😃

nul800sebastiaan (commented):
@BatJan Could package developers be using `.umb-grid .iconBox i`? Cleaning up is great but we have to consider who else is using our styling. :-)

BatJan:
@nul800sebastiaan Good catch! Reverted now 👍🏻

nathanwoulfe (approved):
Looks good to me, thanks @BatJan

nathanwoulfe:
@nul800sebastiaan I've labelled this for 8.17 - is it possible to cherry pick for that release or will it need to wait? Not a big issue, but would be good to resolve.

nathanwoulfe:
Failed test isn't related to this change

nul800sebastiaan:
Cherry picked for 8.17 in e269d1bcd04e379fd2e50db990401d71b3dfcd7b

PR #10979, opened 2021-08-30
Grid: labels offset fix

BatJan:
### Prerequisites

### Description
In the v8.16 release I noticed that the grid labels looked a bit "cramped" - I'm not sure what is causing the issue but it needs to have `top: 15px;` in order to look correct, which is what this PR addresses.

**Before**
grid-options-before

**After**
grid-options-after


nathanwoulfe (approved):
@BatJan, hello. Bit of a strange one, must have been caused by other CSS changes leaking into this component. Your fix does the job though, thanks.

PR #10982, opened 2021-08-30
Close content delete dialog on error

### LINKED ISSUE #10980: Content delete dialog does't close if status code is 400

9.0.0-rc001

### Bug summary

The content delete overlay does not close if status is not success. 

### Specifics

I implemented a `INotificationHandler<ContentMovingToRecycleBinNotification>` and in some scenario's I want to cancel the operation. 

``` c# 
public class CannotDeleteContentNotification : INotificationHandler<ContentMovingToRecycleBinNotification>
  {


      public void Handle(ContentMovingToRecycleBinNotification notification)
      {

          foreach (var moveEventInfo in notification.MoveInfoCollection)
          {
              notification.CancelOperation(new EventMessage("cancel", "lorem ipsum"));

          }
      }
  }
```

This will return a 400 status code but in the `content.delete.controller.js` is no error handling for status code 400 . 

See: https://github.com/umbraco/Umbraco-CMS/blob/695d4688d3d65197c3158d2c6ade921757545dc4/src/Umbraco.Web.UI.Client/src/views/content/content.delete.controller.js#L67 

### Steps to reproduce

Implement a INotificationHandler<ContentMovingToRecycleBinNotification>
Register the Notification
Cancel the operation 

In this case I want to close the dialog.

erikjanwestendorp:
Belongs to https://github.com/umbraco/Umbraco-CMS/issues/10980. Not sure if this is the way to go but I give it a try 😄 

Fixes #10980

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/content/content.delete.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/content/content.delete.controller.js
@@ -68,6 +68,8 @@ function ContentDeleteController($scope, $timeout, contentResource, treeService,
                 // TODO: All YSOD handling should be done with an interceptor
                 overlayService.ysod(err);
             }
+
+            navigationService.hideDialog();
         });
 
     };


nul800sebastiaan:
Thanks @erikjanwestendorp - I haven't tried this out yet, but it looks like something that could also be a problem in v8, so might be nice to target v8 instead (we keep merging up anything new from v8 to v9).

In any case, my real question is: is it always true that we want to hide the panel if there's an error? Is the error always shown outside the panel or do we handle other errors inside the panel (therefore this would hide the actual error if you don't show the panel any more).

erikjanwestendorp:
Thank you for your review. I didn't know this is also an issue in v8. I'm closing this PR for now and will create a new one targeting the v8 branche.

I'm not sure it's always true that we want to close the dialog. But can we assume that if the response data contains a notification, it should be shown outside the dialog and the dialog can be closed? And does the same apply to status 500? Because then an ysod overlay is shown.

I've created a new PR targeting the v8 banche. #11009

Issue #10927, opened 2021-08-23
Umbraco V9 rc002: Keys are shown instead of values


jkulker:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

Umbrac 9 RC0002

### Bug summary

After updating to rc002 the dictonary values are not shown anymore in the fields when creating content. Instead the key values are shown.

![image](https://user-images.githubusercontent.com/38132206/130441625-8645388a-0fb5-4514-a0d2-b8b4d79c4fd7.png)

### Specifics

Editing content

### Steps to reproduce

Updating from release candidate 1 to 2

### Expected result / actual result

When using #title in  in field translate it to a specific value in the dictonary per language

---
_This item has been added to our backlog [AB#13539](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13539)_

jkulker:
Looks like translation are completly broken

meyntony:
**Steps to reproduce**
I have made a repo, that contains the SQL CE database so just switching between the commits can help you see the error:

nul800sebastiaan:
This might be the same problem as https://github.com/umbraco/Umbraco-CMS/issues/10977 - would be good to check while fixing this as well.

bergmania:
I can verify this issue exists, and seems to have been introduced by this PR: https://github.com/umbraco/Umbraco-CMS/pull/10628

Will need to find a solution that works on both cases..

elit0451:
Fixed in #10986

PR #10983, opened 2021-08-30
Bugfix for #10927 - Body steal focus when opening folder overlay

### LINKED ISSUE #10972: Body steal focus when opening folder overlay

8.16.0

### Bug summary

When creating a new folder under e.g. document types, media types or datatypes, the inout field have `umb-auto-focus` directive, however it seems something steal focus and set active focus on body element.

t2I7W6TQRy

### Specifics

_No response_

### Steps to reproduce

Create a new folder in e.g. document types, and notice the input doesn't get focus.

In Chrome use the following to see which element currently has focus: https://developer.chrome.com/docs/devtools/accessibility/focus/

I can reproduce this issue in v8.15.1 as well and back to v8.8.0.

_No response_

BatJan:
### Prerequisites

If there's an existing issue for this PR then this fixes #10972

### Description
When creating a folder under "Settings -> Document types" the input field for the name that appears don't get autofocused. I suspect it's due to the section containing the input element is hidden using `ng-show` instead of `ng-if` making the autofocus directive compete with all other places this directive might have been used and also settled on the element to focus on before the dialog for naming the folder appears. Changing it to use `ng-if` instead makes the directive run and set focus again.

**Before**
focus-issue-before

**After**
focus-issue-after

### DIFF
--- a/build/NuSpecs/UmbracoCms.Core.nuspec
+++ b/build/NuSpecs/UmbracoCms.Core.nuspec
@@ -33,7 +33,7 @@
                 <dependency id="Newtonsoft.Json" version="[12.0.1,12.999999)" />
                 <dependency id="Semver" version="[2.0.4,2.999999)" />
                 <dependency id="Serilog" version="[2.10.0,2.999999)" />
-                <dependency id="Serilog.Enrichers.Process" version="[2.0.1,2.999999)" />
+                <dependency id="Serilog.Enrichers.Process" version="[2.0.2,2.999999)" />
                 <dependency id="Serilog.Enrichers.Thread" version="[3.1.0,3.999999)" />
                 <dependency id="Serilog.Filters.Expressions" version="[2.1.0,2.999999)" />
                 <dependency id="Serilog.Formatting.Compact" version="[1.1.0,1.999999)" />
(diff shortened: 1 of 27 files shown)


bjarnef:
@BatJan I think it was also an issue when creating a folder under media types.
https://github.com/umbraco/Umbraco-CMS/blob/23675a33ee2383cf7ceef38cc2d4942c46e3ee6c/src/Umbraco.Web.UI.Client/src/views/mediatypes/create.html

The part with `ng-show="!model.creatingFolder"` and `ng-show="model.creatingFolder"`

BatJan:
@bjarnef Yes it indeed is - I'm going to push shortly. Just need to figure out an issue with the view for https://github.com/umbraco/Umbraco-CMS/blob/23675a33ee2383cf7ceef38cc2d4942c46e3ee6c/src/Umbraco.Web.UI.Client/src/views/datatype/create.html where I see a "Autofocus processing was blocked because a document already has a focused element" after making the change... So need to figure that one out before comitting the extra changes - But good spot 🙌🏻

BatJan:
@bjarnef Ok, found it - Turns out that for some reason the native `autofocus` attribute was being used instead of the `umb-auto-focus` directive in the view for the datatypes dialog 😁 

But should be all good now 👍🏻 Let me know if you spot anything I might have missed - Have been checking all the create.html views I could find.

bjarnef:
@BatJan yeah, I also noticed one dialog using the native `autofocus`. while others used `umb-auto-focus` and wondering why it didn't use same attribute :)

BatJan:
Oh dear - The changes I though I pushed for this PR ended up in an old branch doh! 🙈 All good now!

nul800sebastiaan:
Awesome, thanks for the help @BatJan and @bjarnef ! 👍

PR #10986, opened 2021-08-30
Fixed issue where keys are shown instead of values

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/10927
while still support the fix for https://github.com/umbraco/Umbraco-CMS/issues/10627


elit0451:
Both cases are supported 👍 Merging

PR #10987, opened 2021-08-30
V8: Validate uploaded files based on the configuration of the FileUpload property configuration

### LINKED ISSUE #10897: File Upload extensions isn't validated

8.14.1

### Bug summary

With the new file upload extensions I want to limit a specific property to accept csv-files only and not e.g. xslx-files.
However when selecting an xslx-file no validation errors are shown and the xslx-file is uploaded.

### Specifics

_No response_

### Steps to reproduce

- Create an instance of file upload and add csv to accepted extensions.

image

- Add a property to a document type using this datatype.

image

- In content upload a xslx-file and click save or save & publish. Notice no validation errors are shown on upload (via click or drag & drop) or during save.

image

When uploading file when an non accepted format it should show a validation error and don't upload the file.

The csv extension is added to the `accept` attribute here though.

image

---
_This item has been added to our backlog [AB#13506](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13506)_

nikolajlauridsen:
This fixes #10897.

Turns out that file extensions of uploaded files weren't validated at all based on the editor configuration, only the configuration in `umbracoSettings.config`. I've added extra validation so we now validate on both.

### Testing 

1. Create a document type with 2 `UploadField` properties, one that allows all extensions, and one that only accepts csv.
2. Try and create a content node, and ensure that:
   1. **Only** csv files are allowed in the limited upload field
   2. Other filetypes are allowed in the unlimited filetypes field
   3. Try adding your csv to the `disallowedUploadFiles` property of `umbracoSettings.config`, and ensure that you are no longer allowed to upload csv files, even to the upload field where you specified csv
   4. Try adding `<allowedUploadFiles>csv</allowedUploadFiles>` and ensure that only csv files are allowed, even in the upload field where everything is allowed


src/Umbraco.Web/PropertyEditors/UploadFileTypeValidator.cs
```diff
@@ -36,20 +36,45 @@ public IEnumerable<ValidationResult> Validate(object value, string valueType, ob
...
+
+        internal static bool IsAllowedInDataTypeConfiguration(string filename, object dataTypeConfiguration)
+        {
+            if (TryGetFileExtension(filename, out var extension) is false) return false;
+
+            if (dataTypeConfiguration is FileUploadConfiguration fileUploadConfiguration)
+            {
+                // If FileExtensions is empty no allowed extensions has been specified, so we allow everything.
+                // If there is any extensions specified, we need to check that the uploaded extension is one of them.
+                return fileUploadConfiguration.FileExtensions.IsCollectionEmpty() ||
```
elit0451, line 64:
```suggestion
                // If FileExtensions is empty and no allowed extensions have been specified, we allow everything.
                // If there are any extensions specified, we need to check that the uploaded extension is one of them.
                return fileUploadConfiguration.FileExtensions.IsCollectionEmpty() ||
```

nikolajlauridsen:
Hmm, I'm not if I'm misunderstanding, but I can't seem to reproduce it, if I select a file, then update the `disallowedUploadFiles`, and wait for the process to reboot, it will show the error and not upload the image, from my testing it doesn't seem to upload the file either, are you sure the site had restarted for the changes in the config to take effect? 


Gif for clarification
File upload

elit0451:
After we figured out that the changes are just saved and not published when the validation errors I spoke about appear and the current workflow makes sense, I am happy to merge.

Nice work 💪

PR #10992, opened 2021-08-31
Suggestion: Make the overlay size width in multiurlpicker configurable

BatJan:
### Prerequisites

### Description
This PR is a suggestion to add a configuration option for the size of the overlay width when using the multiurlpicker - Currently it's always displayed using the small width overlay. Recently I have found it a wee bit frustrating to work with since it can be hard to actually see the full url in the link input field and then it hit me - With the block list editor we can define whether the size should be "Small", "Medium" or "Large" and I figured why not do the same for the multiurlpicker property editor as well? So therefore I made this PR enabling us to define the desired size - What do you think about the idea?

If one does not change the configuration the size defaults to "small".

Below you can see a GIF illustrating how it all works 👍🏻 

overlaysize-mup


nul800sebastiaan:
Good stuff @BatJan! I think `Large` might be a bit too large, but I guess if you have a small screen size and a deep tree it might still make sense!

nathanwoulfe:

PR #10995, opened 2021-08-31
Automatically create emtpy SQL CE database with unattended install

jesperweber:
### Prerequisites

1. Set a connectionString to a non-existing SQL CE database
2. Enable unattended install
3. See that the  empty SQL CE database is created before unattended install will create the schema

### Description

When using unattended install an empty database must be manually created before you start up the application. 

This PR will automatically create an empty SQL CE database on the location configured in the connectionString if the provider name is "System.Data.SqlServerCe.4.0" before the unattended install tries to create the schema.

### DIFF
--- a/src/Umbraco.Core/Runtime/CoreRuntime.cs
+++ b/src/Umbraco.Core/Runtime/CoreRuntime.cs
@@ -1,9 +1,9 @@
 ﻿using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
+using System.Data.SqlServerCe;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using System.Runtime.Serialization;
 using System.Threading;
 using System.Web;
@@ -20,9 +20,7 @@ using Umbraco.Core.Migrations.Install;
 using Umbraco.Core.Migrations.Upgrade;
 using Umbraco.Core.Persistence;
 using Umbraco.Core.Persistence.Mappers;
-using Umbraco.Core.Scoping;
 using Umbraco.Core.Security;
-using Umbraco.Core.Services;
 using Umbraco.Core.Sync;
 
 namespace Umbraco.Core.Runtime
@@ -182,7 +180,7 @@ namespace Umbraco.Core.Runtime
                 AcquireMainDom(MainDom);
 
                 // determine our runtime level
-                DetermineRuntimeLevel(databaseFactory, ProfilingLogger);
+                //DetermineRuntimeLevel(databaseFactory, ProfilingLogger);
 
                 // get composers, and compose
                 var composerTypes = ResolveComposerTypes(typeLoader);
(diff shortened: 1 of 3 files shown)


src/Umbraco.Core/Runtime/CoreRuntime.cs
```diff
@@ -182,7 +180,7 @@ protected virtual IFactory Boot(IRegister register, DisposableTimer timer)
                 AcquireMainDom(MainDom);
 
                 // determine our runtime level
-                DetermineRuntimeLevel(databaseFactory, ProfilingLogger);
+                //DetermineRuntimeLevel(databaseFactory, ProfilingLogger);
```
jesperweber, line 183: Had to disable the DetermineRuntimeLevel here since the SQL CE database might not exists yet, causing an exception.

src/Umbraco.Core/Runtime/CoreRuntime.cs
```diff
@@ -363,6 +361,22 @@ private void DoUnattendedInstall(IUmbracoDatabaseFactory databaseFactory)
...
+
+                if(!File.Exists(dbFilePath))
+                {
+
+                    var engine = new SqlCeEngine(databaseFactory.ConnectionString);
+                    engine.CreateDatabase();
+                }
+            }
+
+            DetermineRuntimeLevel(databaseFactory, ProfilingLogger);
```
jesperweber, line 378: Had to add the DeterminRuntimeLevel here, after the SQL CE database is created.

jesperweber (commented):
I had some problems with the DeterminRuntimeLevel check, as it couldn't run before the database was created and must run after the database creation. Not sure what the method do and when the correct time to call the method is.

nul800sebastiaan:
Thanks @jesperweber! I'm still doing a bit of testing but it seems to be working beautifully! 🎉  

I install Umbraco SO many times in a week, this is really going to help me as well, though I do wish I could install The StarterKit at the same time, but I think that's a super rare request that only a few people in the world really need.

https://user-images.githubusercontent.com/304656/131658753-69dd2981-b37d-495e-84b5-5dec4ffe6e3b.mp4

I made a little helper because I was annoyed I had to manually update and copy files, you can see me operating that in the beginning of the video. The source is available so hack away on it! https://github.com/nul800sebastiaan/Cultiv.Umbraco.Unattended

nul800sebastiaan:
I've tested quite a few scenario's now and I'm happy with the results. Specifically I was a bit worried about cloning Cloud sites where we do some black magic to make a LocalDb instead of SQL CE - but of course that can't kick in if we don't have the web.config key(s) for unattended installs set. 

I would love for this to be properly documented by the way, if you're happy to help there as well? Currently the docs don't even talk about the `unattended.user.json` file. Also the docs say that you need to pre-configure a db, which is now no longer true if you use SQL CE (🎉).

jesperweber:
Thanks for the review and test @nul800sebastiaan. I will see if I get the time to update the documentation as well, it would of course be nice with documentation reflecting the feature.

ronaldbarendse:
I noticed there were quite a few things you need to have correctly configured before you can use the unattended install with automatic database creation:
1. Set the configuration status in your `Web.config` to the current version, e.g.:
```xml
<add key="Umbraco.Core.ConfigurationStatus" value="8.17.0" />
```
2. Enable unattended install and missing database creation:
```xml
<add key="Umbraco.Core.RuntimeState.InstallUnattended" value="true" />
<add key="Umbraco.Core.RuntimeState.InstallMissingDatabase" value="true" />
```
3. Update the connection string to use SQL CE:
```xml
<add name="umbracoDbDSN" connectionString="Data Source=|DataDirectory|\Umbraco.sdf;Flush Interval=1;" providerName="System.Data.SqlServerCe.4.0" />
```
4. Optionally: add unattended user details using environment variables or creating the `App_Data\unattended.user.json` file as described in https://github.com/umbraco/Umbraco-CMS/pull/9930

With all this done, the first request to the site will create the database and setup an empty Umbraco installation 🎉 

I'm proud to say this is greatly improved in Umbraco 9, which also supports automatically creating a LocalDB database and allows configuring all the above from the CLI. The `InstallMissingDatabase` setting also isn't required for embedded databases anymore, but only for creating a new database on a full SQL Server instance (obviously, the configuration string then requires user credentials with the `CREATE DATABASE` permission): https://github.com/umbraco/Umbraco-CMS/pull/11127.

PR #10996, opened 2021-08-31
Option to delete unattended.user.json fil after unattended install

jesperweber:
### Prerequisites

1. Enable unattended install
2. add the following appSetting in web.config `<add key="Umbraco.Core.RuntimeState.DeleteUnattendedUserFile" value="false"/>`
3. See that the unattended.user.json user file is not deleted

### Description

We use unattended install to quickly setup local environments when developers should start working on other projects. 

For that we use the new feature to automatically create a user, but in our local environment we don't want the unattended.user.json file to be deleted every we setup the solution, we only want the unattended.user.json file to be deleted on our test, preproduction and production environments.

This PR add a new option to indicate if the unattended.user.json file should be deleted or not during the unattended install. The default value is true

### DIFF
--- a/src/Umbraco.Core/Runtime/CoreRuntime.cs
+++ b/src/Umbraco.Core/Runtime/CoreRuntime.cs
@@ -343,7 +343,7 @@ namespace Umbraco.Core.Runtime
             Current.Services.UserService.Save(admin);
 
             // Delete JSON file if it existed to tidy 
-            if (fileExists)
+            if (fileExists && RuntimeOptions.DeleteUnattendedUserFile)
             {
                 File.Delete(filePath);
             }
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
I understand the desire, but having this file around is already risky security-wise. I can only see problems where this file stays around forever with clear-text admin credentials. I'll chat to the team, but I feel like this is not a good idea. 

The alternative, which is way more secure AND robust is to use environment variables on your build servers. On Azure DevOps, for example, these can be secrets as well so nobody can ever see them again after they have been saved.

nul800sebastiaan:
Thanks again @jesperweber, indeed we feel like it is a little to prone to errors to keep this file laying around for too long, if you have a look in the other PR I've left you some code that might be useful for you to easily set up local sites as well, hope that helps! https://github.com/umbraco/Umbraco-CMS/pull/10995#issuecomment-910189008

I'll close this PR, but thanks for giving it a try! 👍

jesperweber:
Fair enough.

PR #10998, opened 2021-08-31
Updates email sender to have some metadata

Shazwazza:
Without this there is no way to handle the `SendEmailNotification` when it is triggered (currently only for user invites) to know what the email is for. It is not reasonable to inspect the subject or message body to see if the notification should handle it or not because those things are dynamic based on the current user's culture.

This is just a simple way to know what type of email it is for.


warrenbuckley:
Code looks fine to me, just changed to use some constants.
Waiting for CI to go green & will merge it in

PR #10999, opened 2021-08-31
Avoid array allocations for splitting csv tags

nzdev:
### Prerequisites

### Description
Avoid array allocations for splitting csv tags by using existing array.


nathanwoulfe (approved):
:+1

PR #11003, opened 2021-09-01
Color swatch "selected" state - Icon looks weird

BatJan:
### Prerequisites

### Description
I noticed the "Selected" icon was showing a bit misplaced in the current release and had a stab at the stylesheet to ensure it's centered and showing like it used to 👼🏻 

**Before**
circle-before

**After**
circle-after


nul800sebastiaan:
Thanks @BatJan - this is also a problem with the color picker datatype:

image

And this PR fixes that one as well 🎉 

image

PR #11004, opened 2021-09-01
Suggestion: Increase the icon picker overlay window width

BatJan:
### Prerequisites

### Description
Follow the merge of #10992 @nul800sebastiaan gave a kind hint that it might be a good idea to increase the overlay width for the icon picker as well so without further ado... here it is! 🎉 However I don't know if we should add another size class to make the window a wee bit wider so the color swatches can fit on one line but if more colors are added in the future it won't make a difference long term... who knows 😃 

**Before**
icon-picker-window-before

**After**
icon-picker-window-after

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/editor.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/editor.service.js
@@ -632,6 +632,7 @@ When building a custom infinite editor view you can use the same components as a
          */
         function iconPicker(editor) {
             editor.view = "views/common/infiniteeditors/iconpicker/iconpicker.html";
+            editor.size = "medium";
             if (!editor.size) editor.size = "small";
             open(editor);
         }
--- a/src/Umbraco.Core/Exceptions/InvalidCompositionException.cs
+++ b/src/Umbraco.Core/Exceptions/InvalidCompositionException.cs
@@ -35,6 +35,14 @@ namespace Umbraco.Core.Exceptions
         public string[] PropertyTypeAliases { get; }
 
         /// <summary>
+        /// Gets the property group aliases.
+        /// </summary>
+        /// <value>
+        /// The property group aliases.
+        /// </value>
+        public string[] PropertyGroupAliases { get; }
+
+        /// <summary>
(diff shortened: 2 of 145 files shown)


BatJan:
@nul800sebastiaan Just did a few changes to ensure the colors can be displayed in one line, which looks like the image below showcases 👍🏻 

The selected state icon is adressed in #11003 and I have tested this update against it and it appears to work fine 👍🏻 

iconpicker-swatch-adjustments

nul800sebastiaan:
@BatJan Cool! Can you check against the Approved Color datatype as well, that's the other place these color swatches are being shown (see PR https://github.com/umbraco/Umbraco-CMS/pull/11003 for screenshots).

src/Umbraco.Web.UI.Client/src/common/services/editor.service.js
```diff
@@ -632,6 +632,7 @@ When building a custom infinite editor view you can use the same components as a
          */
         function iconPicker(editor) {
             editor.view = "views/common/infiniteeditors/iconpicker/iconpicker.html";
+            editor.size = "medium";
```
bjarnef, line 635: I think it need to pass in the `size` property instead in case developers are implementing their own functionality use `iconPicker`. Could you e.g. add a button in a dashboard, which trigger iconPicker with size "large"?

BatJan:
@nul800sebastiaan It will work but it will also make the swatch smaller in that scenario. We can introduce a modifier for the icon-picker, maybe call it "--small" and ensure that proportions will only be changed where the modifier is used. I'll look into it 👍🏻

src/Umbraco.Web.UI.Client/src/common/services/editor.service.js
BatJan: Yeah that's probably safer. Will give it a shot 👍🏻

BatJan:
@nul800sebastiaan @bjarnef I have decided to make the changes in a new PR since my commit history is messed up for some obscure reason I can't figure out so I've made a fresh clone to avoid this mess in the future. Please see #11015 for the changes made 👍🏻

PR #11006, opened 2021-09-01
Suggestion: Extend markdown editor settings for overlay width

BatJan:
### Prerequisites

### Description
In relation to #10992 being merged I have now extended the config settings on the property editor for the Markdown Editor datatype making it possible to adjust the size of the link picker window as well. See how it works below.

But perhaps we don't need the settings for this and just agree to always use the medium size instead? However it might be nice to be able to configure it? Let me know your thoughts!

overlaysize-markdowneditor


nul800sebastiaan:
I love it, thanks again @BatJan! 👍

nathanwoulfe:

PR #11007, opened 2021-09-01
Suggestion: Add config setting in Rich Text Editor to control the linkpicker overlay size

BatJan:
### Prerequisites

### Description
This PR is similar to #11006 and please refer to the GIF illustration in that PR when understanding what this PR does. Only difference is that this affects the Rich Text Editor 😃

### DIFF
--- a/src/Umbraco.Core/Exceptions/InvalidCompositionException.cs
+++ b/src/Umbraco.Core/Exceptions/InvalidCompositionException.cs
@@ -35,6 +35,14 @@ namespace Umbraco.Core.Exceptions
         public string[] PropertyTypeAliases { get; }
 
         /// <summary>
+        /// Gets the property group aliases.
+        /// </summary>
+        /// <value>
+        /// The property group aliases.
+        /// </value>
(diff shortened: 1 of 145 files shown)


bjarnef:
@BatJan as mentioned here https://github.com/umbraco/Umbraco-CMS/commit/e707490b471098d229cd23b44b24b408a4e12c78#r55772637 could we add a `overlaysize` global prevalue editor?

If there is a specific reason to add a `overlaysize` prevalue editor for a property editor, we can always add that later. But for now if seems to be redundant.

nul800sebastiaan:
Nice one again @BatJan 😁

bjarnef:
I have cleaned a bit up in https://github.com/umbraco/Umbraco-CMS/pull/11010 so we don't have a redundant prevalue editor.

bjarnef:
It seems this doesn't work, but fixed in https://github.com/umbraco/Umbraco-CMS/pull/11010

The `overlayWidthSize` prevalue editor is not part of the `args.model.config.editor` object, but `args.model.config`.

nathanwoulfe:

Issue #10108, opened 2021-04-07
Select icon when creating a composition


soreng:
When using the new create doctype dialog in v8.12, the the icons in the dialog should be selected in the new type being created.

Right now we have to select the same icon every time we create a composition. 

Also, it would be nice to have some colours pre-selected too. Maybe based on how @prjseal does in his awesome tutorials.

nul800sebastiaan:
I am a little lost here @soreng - would you be able to illustrate with a couple of screenshots/gif please?

soreng:
When you click the type of type (?) you want to create, there is no icon set.
![image](https://user-images.githubusercontent.com/359492/114313617-6eb6b780-9af7-11eb-8b13-1ddeee2ef090.png)

It would be nice if, instead of nothing, the icon from the type, ie the "defrag" icon for compositions, was set on the new type you are creating.

skttl:
I think it makes perfect sense for the default icon to match the ones in the dialog 👍

nielslyngsoe:
I think it's a good idea. :-)

The only one of these icons I'm concerned about is the "Document Type" (without template). 

I´m not sure what the icon symbolizes, plus I would think many use this choice as a default initial step, which ends up begin used as any of the other choices. Then the icon ends up giving a messy look. 
I myself always pick "Document Type" and then I later use it for a composition or with a template. I probably already have a template that I wanted to use for it.

As the icon of the DocumentType is used in the Editor Experience we are not going to ensure that icons represent the features/configuration of the DocumentType, in other words, we cannot make sure that we have an icon that provides the right representation. This argues against having such a feature, but I think for the choices "Element Type" and "Composition" we are fairly certain of the purpose and it will most likely not be changed, therefore I think we can set the default icon for those.

In order to provide a meaningful solution to the problem, I think the icon ofthe choice "Document Type"(No template) should be the document icon as the one for "Document Type with Template"

I can add that I think this dialog would be improved as well by using the same icon as Document Type with Template, as they are almost the same, this would visually provide a relation, making it easier to distinguish these two from the others.

That's my 50 cents :-)

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/11008 - I've moved the labels there!

PR #11008, opened 2021-09-01
Implement icon parameter for doctype editor

soreng:
This PR aims to resolve issue #10108

By enabling a parameter for the icon (+color), we can mimic the new icons introduced in the "create document type dialog".

_In this example I have also added colors to the icons_
image

The icon and color are passed in as parameters in the actions triggered by clicking the different types. The icon + color are then passed in to the model being passed to the document type editor.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/documenttypes/create.html
+++ b/src/Umbraco.Web.UI.Client/src/views/documenttypes/create.html
@@ -5,8 +5,8 @@
 
             <ul class="umb-actions umb-actions-child">
                 <li data-element="action-documentType" class="umb-action" ng-hide="model.disableTemplates">
-                    <button type="button" ng-click="createDocType()" class="umb-action-link umb-outline btn-reset" umb-auto-focus>
-                        <umb-icon icon="icon-document" class="icon large"></umb-icon>
+                    <button type="button" ng-click="createDocType('icon-document', 'color-blue')" class="umb-action-link umb-outline btn-reset" umb-auto-focus>
+                        <umb-icon icon="icon-document color-blue" class="icon large"></umb-icon>
                         <span class="menu-label">
                             <localize key="create_documentTypeWithTemplate">Document Type with Template</localize>
                             <small><localize key="create_documentTypeWithTemplateDescription">The data definition for a content page that can be created by editors in the content tree and is directly accessible via a URL.</localize></small>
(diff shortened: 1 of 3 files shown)


src/Umbraco.Web.UI.Client/src/views/documenttypes/create.html
```diff
@@ -5,35 +5,35 @@ <h5><localize key="create_createUnder">Create an item under</localize> {{current
 
             <ul class="umb-actions umb-actions-child">
                 <li data-element="action-documentType" class="umb-action" ng-hide="model.disableTemplates">
-                    <button type="button" ng-click="createDocType()" class="umb-action-link umb-outline btn-reset" umb-auto-focus>
-                        <umb-icon icon="icon-document" class="icon large"></umb-icon>
+                    <button type="button" ng-click="createDocType('icon-document', 'color-blue')" class="umb-action-link umb-outline btn-reset" umb-auto-focus>
+                        <umb-icon icon="icon-document color-blue" class="icon large"></umb-icon>
```
bjarnef, line 9: @soreng with the recent changed it should be sufficient with icon name in `icon` attribute. The color class can be added to `class` attribute and should be inherited down to the font icon or SVG icon.

soreng: @bjarnef I have move the "color-"-classes to the class attribute

soreng:
> I know there's a strong faction at HQ who doesn't think doctypes should have differentiating colors at all.

I was hoping for this discussion. The colors is also fairly easy to remove :)

nul800sebastiaan:
As I expected: the feedback is to stick with the colors being black by default, we won't be able to please everyone with a suggested color pallet anyway, so most people will probably have to change the colors anyway.

In short: if you can make them black, then we'd be happy with this PR! 😁

soreng:
@nul800sebastiaan done :)

nathanwoulfe (approved):
Works perfectly, thanks @soreng. I've added an additional commit to remove some unused parameters

nathanwoulfe:
Will merge when the tests go green

src/Umbraco.Web.UI.Client/src/views/documenttypes/edit.controller.js
```diff
@@ -415,6 +417,12 @@
             if (isElement) {
                 contentType.isElement = true;
             }
+
+            // set isElement checkbox by default
+            if (icon !== null) {
+                contentType.icon = icon;
+            }
+
```
bjarnef, line 425: This comment doesn't make much sense.

PR #11010, opened 2021-09-01
Introduce overlaysize prevalue editor

bjarnef:
... instead of redundant prevalue editor for each property editor.

So instead of three redundant prevalue editors, we just need one!
If we later need only to show e.g. "Small" and "Medium" options in prevalues for a specific property editor, we can configurate the prevalue editor further or add a specific prevalue editor for that property editor. But it most scenarios we don't need a specific prevalue for a property editor, but one that can be re-used in other property editors and/or by package developers.

I have renamed the property from "Overlay Size Width" to just "Overlay Size" to keep it simple and instead use the description property to add a more detailed explanation.

### Prerequisites

### Description
I don't like that we now have redundant prevalue editors for each property editor - we should have a global prevalue editor to use.

Relatated to these PRs:
https://github.com/umbraco/Umbraco-CMS/pull/10992
https://github.com/umbraco/Umbraco-CMS/pull/11006
https://github.com/umbraco/Umbraco-CMS/pull/11007

**Multi URL Picker**

image

**Markdown Editor**

image

**Richtext Editor**

image

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/tinymce.service.js
@@ -1543,7 +1543,7 @@ function tinyMceService($rootScope, $q, imageHelper, $locale, $http, $timeout, s
                         dataTypeKey: args.model.dataTypeKey,
                         ignoreUserStartNodes: args.model.config.ignoreUserStartNodes,
                         anchors: anchorValues,
-                        size: args.model.config.editor.overlaySize,
+                        size: args.model.config.overlaySize,
                         submit: model => {
                             self.insertLinkInEditor(args.editor, model.target, anchorElement);
                             editorService.close();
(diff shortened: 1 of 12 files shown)


bjarnef:
@nul800sebastiaan hopefully we can include this in v8.17 instead, so we only introduce a single prevalue editor for overlay size instead of three (which does the same job) 😬

bjarnef:
@madsrasmussen @nielslyngsoe could we include just an `overlaysize` prevalue editor instead the of the three identical prevalue editors please? 😁

bjarnef:
The overlay size didn't work for link picker in rich text editor.
The prevalue `overlaySize` isn't part of `args.model.config.editor` object, but `args.model.config`.

bjarnef:
@nul800sebastiaan can we fix this before release of 8.17 .. at least the configured overlay size for richtext editor doesn't work 😁

nathanwoulfe (approved):
Hey @bjarnef, this all works nicely - can set (and save) different overlay sizes for the three editors, overlays are rendered at the correct size.

nathanwoulfe:
@nul800sebastiaan do you seen any reason why this shouldn't be pulled into the 8.17 release?

PR #11014, opened 2021-09-02
Migrate SQL Server ntext columns to NVARCHAR(MAX)

nzdev:
### Prerequisites

### Description
It's not possible to set up indexes on ntext columns. This PR migrates ntext columns on SQL server to nvarchar(MAX) which is supported in indexes. SQL CE columns are left as is. There is handling in SQLCE for NVARCHAR(MAX) columns in queries already when using strongly typed queries.

How to test?
Umbraco upgrades the database without error.
Existing functionality still works.


nul800sebastiaan:
Thanks @nzdev - I would really love to get this merged, but I'm a bit afraid of doing this on a large database. Have you tested against a significant database size to see if this works well? I get the SQL timeout can be temporarily increased for those scenarios?

PR #11015, opened 2021-09-02
Suggestion: Change size of iconpickers in backoffice

BatJan:
### Prerequisites

There is a current PR where some changes were suggested in #11004, which I have chosen to do in a new PR since there has been some weirdness around the commit history when creating the PR. I have checked my git log but only see the commits made so made a fresh clone and started over to see if that would fix it - It appears to do so. But keep track of the history for this PR in #11004 which is now closed.

In this new PR I have added a new "xs" modifier so the swatch can be made a wee bit smaller than the original size "s", which ensures we don't end up breaking something.

Also as suggested I moved the setting of the size into the object being passed to the iconpicker method instead so we don't force package creators, and developer relying on the iconpicker method in their own extensions to use the new overlay with size if they don't want to - Ensuring it's fully configurable on the object passed.

Furthermore I made this change in the datatype for configuring custom list views as well so it's all aligned where ever the iconpicker overlay is used in the backoffice 👍🏻

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/editor/umbeditorheader.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/editor/umbeditorheader.directive.js
@@ -251,6 +251,7 @@ Use this directive to construct a header inside the main editor window.
                 var iconPicker = {
                     icon: scope.icon.split(' ')[0],
                     color: scope.icon.split(' ')[1],
+                    size: "medium",
                     submit: function (model) {
                         if (model.icon) {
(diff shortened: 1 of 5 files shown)


nathanwoulfe (commented):
Hey @BatJan this all looks great, but seems the PR is coming from umbraco:Umbraco-CMS, rather than BatJan:Umbraco-CMS?

BatJan:
@nathanwoulfe Oh dear! What a mess! Fixed here #11205

nathanwoulfe:
Resolved in #11205 👍

PR #11019, opened 2021-09-02
Don't cache AuthenticationController calls including GetRemainingTimeoutSeconds

### LINKED ISSUE #11011: Some user-specific requests in the backoffice should not be cached

8.16.0

### Bug summary

When a caching server is set up in front of an Umbraco website that respects the `Cache-control` header and I log in in the backoffice, I see some requests to the backoffice API that are cached by the caching server (Azure Front Door in my case) that should not be cached because it's user-specific.

### Specifics

One specific example is the request to `/umbraco/backoffice/UmbracoApi/Authentication/IsAuthenticated` that is cached that shouldn't be. When I go on the backoffice login page, the request is fired and it returns me `false`, so when I log in to the backoffice and refresh the page, I'm redirected to the backoffice login page because the `IsAuthenticated` request returns 
`false` when it should return `true`.

It's also a problem for the request `/umbraco/backoffice/UmbracoApi/Authentication/GetRemainingTimeoutSeconds`.

### Steps to reproduce

1. Setup a caching server in front of the Umbraco website that respects the `Cache-control` header
2. Login to the backoffice
3. Refresh the page

Expected: I should stay on the backoffice dashboard
Actual: I'm redirected to the login page

pguilmettenmedia:
### Prerequisites

1. Login to the backoffice
2. Look for the `Cache-control` response header on requests made to the `AuthenticationController` (`/umbraco/backoffice/UmbracoApi/Authentication/`)
3. Verify the value is always `no-store, must-revalidate, no-cache, max-age=0`

Example for the `IsAuthenticated` request:
image

Example for the `GetRemainingTimeoutSeconds` request:
image

Fixes #11011 

### Description
I added the `[DisableBrowserCache]` attribute on the `AuthenticationController` because the controller doesn't inherit from `UmbracoAuthorizedApiController ` as @Shazwazza mentionned in the linked issue.

For the `GetRemainingTimeoutSeconds` request specifically, I added the missing parts of the `Cache-control` header manually in the string array that already existed.

So now, when a caching server is setup in front of an Umbraco website, user-specific requests will never be cached.

### DIFF
--- a/src/Umbraco.Web/Editors/AuthenticationController.cs
+++ b/src/Umbraco.Web/Editors/AuthenticationController.cs
@@ -17,7 +17,6 @@ using Umbraco.Core.Models.Identity;
 using Umbraco.Core.Services;
 using Umbraco.Web.Models;
 using Umbraco.Web.Models.ContentEditing;
-using Umbraco.Web.Mvc;
 using Umbraco.Web.Security;
 using Umbraco.Web.WebApi;
 using Umbraco.Web.WebApi.Filters;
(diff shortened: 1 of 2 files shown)


mikecp (approved):

mikecp:
Thanks for spotting the issue and for this PR  @pguilmettenmedia, that was a tricky one 👏

It now behaves as expected 👍

PR #11027, opened 2021-09-04
Fix typo

TBVanderley4545:
Updated typo retrive to be retrieve throughout project.


bjarnef:
I would keep the description starting with an uppercase.
Furthermore we can't just rename the directive without a breaking change. So we should probably add the same directive fixing the typo, while keeping the existing one.

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
```diff
@@ -77,7 +77,7 @@ function clipboardService($window, notificationsService, eventsService, localSto
 
     var STORAGE_KEY = "umbClipboardService";
 
-    var retriveStorage = function() {
+    var retrieveStorage = function() {
```
nul800sebastiaan, line 80: I don't think we can just rename this one, it would be a breaking change for people already using the old function name. The way we typically deal with that is to leave the old name in place and redirect all usages to the new function name.

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
```diff
@@ -414,9 +414,9 @@ function clipboardService($window, notificationsService, eventsService, localSto
     * @description
     * Returns an array of entries matching the given type and one of the provided aliases.
     */
-    service.retriveEntriesOfType = function(type, allowedAliases) {
+    service.retrieveEntriesOfType = function(type, allowedAliases) {
```
nul800sebastiaan, line 417: Same for this one, we need to redirect calls from the old to the new function.

nul800sebastiaan (changes requested):
Thanks @TBVanderley4545 for the help here. I found a few things that should probably be updated (but correct me if I'm wrong). Could you have a look at those please?

nul800sebastiaan:
Ah I said the same thing as @bjarnef - then it seems like I'm on the right track! ;-)

src/Umbraco.Web.UI.Client/src/common/services/clipboard.service.js
nathanwoulfe: This one should be ok - it's an internal function in the service, so won't be called directly. Point still stands though that any public function should be aliased/redirected from a new correctly spelled version.

nathanwoulfe:
@TBVanderley4545 would you be able to resolve the issues here? Short of it is the service will need the existing misspelled methods to remain, but can be supplemented with new correctly spelled functions, which call the existing ones. Something like this:

```
service.retrieveEntriesOfType = (type, allowedAliases) => retriveEntriesOfType(type, allowedAliases);
```
Saves introducing breaking changes 😃

nathanwoulfe (approved):
Restored the typo'd functions in the clipboard service, these now call the correctly named version and log a console warning to encourage devs to use the correct versions 👍

PR #11029, opened 2021-09-05
Fix message template to not mix string interpolating and formatting.

nzdev:
### Prerequisites

### Description
Fix message template to not mix string interpolating and formatting. Message template parameter gets cached, using $"" will create more unique message templates, increasing memory and cpu consumption.


nul800sebastiaan:
Thanks @nzdev!

PR #11032, opened 2021-09-06
Improvement for #10845 - Changed culture comparison of domain to be case insensitive

vsilvar:
### Prerequisites

### Description
Just a small improvement to #10845 to make sure we use case insensitive comparison of cultures.


nikolajlauridsen (approved):
Ah good catch, thanks 😄

PR #11040, opened 2021-09-07
Fixes folder creation for doc, data, and media types

madsrasmussen:
PR: https://github.com/umbraco/Umbraco-CMS/pull/10983 introduces a bug where you can't create folders for doctypes and data types anymore.

I have updated the view to use `ng-show` instead of `ng-if.` Ng-if introduces a new scope that breaks the relation to the form. I have also updated the focus directive to use `focus-when` instead of `auto-focus`. It makes it possible to control when a field should have focus.

How to test:
* Make sure the input field has focus when you create a new folder for doctypes and datatypes
* Make sure you can save and the folder gets created

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/datatypes/create.html
+++ b/src/Umbraco.Web.UI.Client/src/views/datatypes/create.html
@@ -36,7 +36,7 @@
                         ng-model="model.folderName" 
                         class="umb-textstring textstring input-block-level"
                         focus-when="{{model.creatingFolder}}" 
-                        required umb-auto-focus />
+                        required />
                 </umb-control-group>
 
                 <button type="submit" class="btn btn-primary"><localize key="general_create">Create</localize></button>
--- a/src/Umbraco.Web.UI.Client/src/views/datatypes/create.html
+++ b/src/Umbraco.Web.UI.Client/src/views/datatypes/create.html
@@ -24,13 +24,19 @@
             </ul>
         </div>
 
-        <div class="umb-pane" ng-if="model.creatingFolder">
+        <div class="umb-pane" ng-show="model.creatingFolder">
             <form novalidate name="createFolderForm"
                 ng-submit="createContainer()"
                 val-form-manager>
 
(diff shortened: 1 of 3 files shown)


AndyButland (approved):
Have tested this before and after the PR, and can confirm previously that I couldn't create folders, but with the fix here - I can.  Focus appears in the text box as expected too.

(Also double-checked creating folders in forms... fortunately I used `ng-show` to start with so seems fine there).

src/Umbraco.Web.UI.Client/src/views/datatypes/create.html
```diff
@@ -24,13 +24,19 @@ <h5><localize key="create_createUnder">Create an item under</localize> {{current
...
 
                 <umb-control-group label="Enter a folder name" hide-label="false">
-                    <input type="text" name="folderName" ng-model="model.folderName" class="umb-textstring textstring input-block-level" required umb-auto-focus />
+                    <input 
+                        type="text" 
+                        name="folderName" 
+                        ng-model="model.folderName" 
+                        class="umb-textstring textstring input-block-level"
+                        focus-when="{{model.creatingFolder}}" 
+                        required umb-auto-focus />
```
bjarnef, line 39: @madsrasmussen do this need `umb-auto-focus` directive now it has `focus-when`. It is not the case for media types.

madsrasmussen: Nice catch! 💯 

It is removed 👍

bjarnef:
@AndyButland I haven't tried out Umbraco Forms v8.8 yet, but I wonder if it auto focus the input field when creating a folder? Does it also use `focus-when`?

AndyButland:
It's using this:

```
<input type="text" name="folderName" maxlength="255" ng-model="model.folderName" class="umb-textstring textstring input-block-level" umb-auto-focus required />
```

But that isn't working to auto-focus I see. 

I'll have a look at this for the next update and try it with `focus-when` as is used here.  Thanks for flagging.

PR #11051, opened 2021-09-08
Fix concurrency and certificate days to expiry issues in HttpsCheck

ronaldbarendse:
The `HttpsCheck` health check stored the amount of days the certificate is valid in a static field and didn't update it in subsequent checks. Besides concurrency issues, not updating the value would cause the check to silently ignore certificates that are about to expire.

Because the static field/value is reset on an application restart and you're alerted when the certificate is valid for less than 14 days, the real-world impact will be minimal though. But once the certificate will expire within this period, it's nice to see it correctly updated in the health check message (if the application doesn't restart daily).


bergmania:
Changes make sense, and seems to work 💪

PR #11052, opened 2021-09-08
V9: Update MaxRequestBodySize on both IIS Server and Kestrel Server

elit0451:
## Details
This PR is made as a follow-up fix based on the discussion in #10930.
It makes sure that we update the MaxRequestBodySize based on what value we have configured for `Umbraco:CMS:Runtime:MaxRequestLength` (_[docs](https://our.umbraco.com/documentation/Reference/V9-Config/RuntimeSettings/)_) - If no value is specified, the `MaxRequestBodySize` value on the server will be set to null or in other words to unlimited.

It fixes the back-end for the reported #10920 issue. The front-end should be investigated further in order to completely address the reported problem. 

## Test
- Run Umbraco with IIS Express (IIS Server) or Umbraco.Web.UI (Kestrel Server);
- Configure Runtime settings as shown on the docs example;
- Make sure that the value set in `MaxRequestLength` updates `IISServerOptions.MaxRequestBodySize` or `KestrelServerOptions.MaxRequestBodySize` depending on how are you running Umbraco - set a break point in **ConfigureIISServerOptions.cs** and **ConfigureKestrelServerOptions.cs** and check the values (you no longer should see the default 30,000,000 bytes).

-----
The problem with the front-end is that when a file is bigger than the specified limit, no "_Max file size_" error is shown on the UI.


inetzo:
Hi @elit0451, when i apply your PR, i still can't upload files larger than 30 MB. I tried to upload a 307 MB mp4 file. I did not configure the Runtime MaxRequestLength setting. 

On Kestrel i'll get an error on https://github.com/umbraco/Umbraco-CMS/blob/64e958217cfe3d9a150554dc751f70e42f7500d9/src/Umbraco.Web.BackOffice/Controllers/MediaController.cs#L704 because `file == null` . 

The MaxRequestBodySize is set to null (30000000 before) so you would expect that there is no longer any limit.
afbeelding


On IIS Express it `/umbraco/backoffice/umbracoapi/media/PostAddFile` returns "HTTP Error 413.1 - Request Entity Too Large"
MaxRequestBodySize is here also set to null 
afbeelding

inetzo:
This is really wierd: i could upload files up to 130 MB without any errors, but files larger than 130 MB fail on the MediaController. In that case all parameters of PostAddFile are null.

inetzo:
After applying MultipartBodyLengthLimit on FormsOptions the upload of large files on Kestrel works like a charm!
But unfortunately for IIS the maximum upload size is still determined by IIS maxAllowedContentLength. When Umbraco:CMS:Runtime:MaxRequestLength is equal or less than IIS maxAllowedContentLength the problem does not occur but in that case it is necessary to manually configure Umbraco:CMS:Runtime:MaxRequestLength to avoid this.

Issue #10403, opened 2021-06-07
Backoffice notifications not displaying


nathanwoulfe:
### Which Umbraco version are you using?

V9 beta003

### Bug summary

I return a `List<BackOfficeNotification>` to display UI messages in response to editor actions. These are not displaying in v9 as the notification objects have a field named `NotificationType`, while the front-end notificationsService checks for `type`.

Not sure why this isn't working, given the BackOfficeNotification type hasn't changed for V9, nor has the AngularJs service.

### Steps to reproduce

Create a controller like so:

```csharp
public class NotificationsExampleController : UmbracoAuthorizedApiController {
  public IActionResult Get() {
    var simpleNotificationsModel = new SimpleNotificationsModel();
    simpleNotificationsModel.AddNotification("Header", "message", NotificationStyle.Info);

    return Ok(new
    {
        notifications = simpleNotificationsModel.Notifications
    });
  }
}
```

Modify umbraco.services.js, adding the below at line 3360 (quicker than adding a new package, all we need is the js to execute).

```javascript
  $http.get('/umbraco/backoffice/api/notificationsexample/get').then(resp => {
    resp.data.notifications.forEach(r => notificationsService.showNotification(r));
  });
```

### Expected result / actual result

I'd expect my notifications to render, instead I get console errors:

`args.type cannot be null`

This throws at line 4981 of umbraco.services.js:

![image](https://user-images.githubusercontent.com/3248070/120974993-1a0c8180-c7b4-11eb-83fe-f317da281d7b.png)

because the returned notifications look like so:

![image](https://user-images.githubusercontent.com/3248070/120975092-33153280-c7b4-11eb-9f90-e43f6b6567bb.png)

rasmusjp:
This might be because the models are serialized using `System.Text.Json` which doesn't use the `DataContract` attributes

nathanwoulfe:
That sounds likely, will have a proper look tomorrow. For now, I've extended the BackOfficeNotification object to add a Type field, which solves the problem but is a bit ugly.

bergmania:
I have looked into this and you are right @rasmusjp.

 i wonder why all of our API controllers are not using the same serilization. Would it be a problem if all of them return the angular prefix and use the newtonsoft serialization.
 
 In that case this controller would return this:
```json
)]}',
{"notifications":[{"header":"Header","message":"message","type":"Info"}]}
```

nikolajlauridsen:
Fixed in #11054

PR #11054, opened 2021-09-08
v9: Bugfix - Implemented attribute changes in controller

Zeegaan:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/10403

# Notes
- Removed attributes from UmbracoAuthorizedJsonController.cs
- Added those attributes to UmbracoAuthorizedController.cs

# How to test
- Set up a new document type with a rich text editor and tags
- Try making some content and fill out a few tags
- Upload an image in the rich text editor
- Save content
- Navigate to the contents information 
- Go to settings and click published status and try to reload the Memory Cache, this should hit all affected classes and everything should work as intended 👍

### DIFF
--- a/src/Umbraco.Web.BackOffice/Controllers/UmbracoAuthorizedApiController.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/UmbracoAuthorizedApiController.cs
@@ -25,7 +25,6 @@ namespace Umbraco.Cms.Web.BackOffice.Controllers
     /// </remarks>
     [AngularJsonOnlyConfiguration] // TODO: This could be applied with our Application Model conventions
     [JsonExceptionFilter]
-    [AngularJsonOnlyConfiguration]
     [IsBackOffice]
     [UmbracoUserTimeoutFilter]
     [Authorize(Policy = AuthorizationPolicies.BackOfficeAccess)]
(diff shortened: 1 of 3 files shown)


src/Umbraco.Web.BackOffice/Controllers/UmbracoAuthorizedApiController.cs
```diff
@@ -23,6 +23,9 @@ namespace Umbraco.Cms.Web.BackOffice.Controllers
     /// is logged in using forms authentication which indicates the seconds remaining
     /// before their timeout expires.
     /// </remarks>
+    [AngularJsonOnlyConfiguration] // TODO: This could be applied with our Application Model conventions
+    [JsonExceptionFilter]
+    [AngularJsonOnlyConfiguration]
```
bjarnef, line 28: @Zeegan I guess we don't need `[AngularJsonOnlyConfiguration]` twice here.

Zeegaan: Nice catch 💪

Zeegaan, line 28:
```suggestion
```

nikolajlauridsen (approved):
Code looks good, going through your tests everything works fine as well, and I can confirm that the bug no longer occurs 👍.

PR #11058, opened 2021-09-08
Fixes email handler detection

Shazwazza:
This fixes an issue where we weren't properly handling the detection of existing `SendEmailNotification` handlers. 

* DI by default wasn't using the ctor with the most params
* We weren't detecting async handlers
* I've changed all usages of AddUnique to AddSingleton in the UmbracoBuilder.CoreServices.cs (AddUniqe should really be marked as obsolete).


bergmania (approved):
Code changes looks fine, and it makes sense it was a bug not checking the async handlers

PR #11059, opened 2021-09-09
V9: Fix ApplicationMainUrl not being set from appsettings.json

nikolajlauridsen:
This fixes an issue where if you specify your umbraco application URL in the appsettings, it would never actually get set as the `AspNetCoreHostingEnvironment.ApplicationMainUrl`, causing `TouchServerTask` to log `No umbracoApplicationUrl for service (yet), skip.` over and over and over again indefinitely.

This was simply just an oversight, in `EnsureApplicationMainUrl` we simply return if `_webRoutingSettings.CurrentValue.UmbracoApplicationUrl is not null` assuming it has been set, however it hadn't, to remedy this I've added a check in the constructor where we set the URL if it's specified in the settings.

### Testing 

Manually specify the UmbracoApplicationUrl in appsettings.json

```json
"WebRouting": {
  "UmbracoApplicationUrl": "https://localhost:44331/umbraco"
}
```

Ensure that `TouchServerTask` doesn't log `No umbracoApplicationUrl for service (yet), skip.` over and over again.

### DIFF
--- a/src/Umbraco.Web.Common/AspNetCore/AspNetCoreHostingEnvironment.cs
+++ b/src/Umbraco.Web.Common/AspNetCore/AspNetCoreHostingEnvironment.cs
@@ -39,6 +39,11 @@ namespace Umbraco.Cms.Web.Common.AspNetCore
 
             SiteName = webHostEnvironment.ApplicationName;
             ApplicationPhysicalPath = webHostEnvironment.ContentRootPath;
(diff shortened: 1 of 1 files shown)


Zeegaan:
Tested this(Thanks for the code snippet 💪 ) and it works! 

Code looks good to me 👓

PR #11061, opened 2021-09-09
Show UI errors for invalid files in media upload

madsrasmussen:
We don't get the invalid files from the uploader as part of the "files"-array anymore but instead as a separate array. I am not sure when this has changed.

This PR handles the new array of invalid files. It combines it with the files array, so everything works the way it used to.

After the PR is applied the errors are back in the UI.
Screenshot 2021-09-09 at 13 25 16

Relates to #10920

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbfiledropzone.directive.js
+++ b/src/Umbraco.Web.UI.Client/src/common/directives/components/upload/umbfiledropzone.directive.js
@@ -65,16 +65,15 @@ angular.module("umbraco.directives")
                     function _filesQueued(files, event) {
                         //Push into the queue
                         Utilities.forEach(files, file => {
+                            if (_filterFile(file) === true) {
 
-                                if (_filterFile(file) === true) {
-
-                                    if (file.$error) {
-                                        scope.rejected.push(file);
-                                    } else {
-                                        scope.queue.push(file);
-                                    }
+                                if (file.$error) {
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Hmm, this works when I drag & drop files onto the drop zone but not when I pick files. In this case in ` web.config `  I have `<requestLimits maxAllowedContentLength="5000" />`

https://user-images.githubusercontent.com/304656/133076699-38eb2b78-61b2-47ef-9059-681e835b2ec0.mp4

The error is ` Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: file in rejected, Duplicate key: object:598, Duplicate value: {"$error":"maxSize","$errorMessages":{"maxSize":true},"$errorParam":"500KB"}`

madsrasmussen:
In the Web.Config we have two values that influence how large files you can upload in the Backoffice: `maxRequestLength` and `maxAllowedContentLength`. 

`maxRequestLength` is exposed to the client in "Umbraco.Sys.ServerVariables.umbracoSettings.maxFileSize". If a file exceeds this size, we let the client-side handle validation, and the file never hit the server. If maxRequestLength is larger than maxAllowedContentLength client-side validation doesn't kick in, and we get a server error that isn't handled.

I have added code to handle the server error now. We should expose `maxAllowedContentLength` to the client so we can do client-side validation for that also.

nul800sebastiaan:
That looks good now, both drag & drop and "click to choose files" 👍

PR #11062, opened 2021-09-09
Fix string interpolation for health check messages

Jeavon:
Security header health check messages in 8.15+ are not retrieving their localised values, this is due to the changes made here https://github.com/umbraco/Umbraco-CMS/commit/dd111226090642cd6777630ff45e432ce8349656#diff-4e1e06ab6889463c603c0face1da054732d42dfea0e81aad113191b2ba048980R90 

image

This PR fixes the interpolation oversight


nul800sebastiaan:
Good spot there @Jeavon, thanks very much! 👍

Issue #10650, opened 2021-07-10
v9-rc001 After upgrading code the upgrade process is not launched on /umbraco


JoseMarcenaro:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

v9.0.0-rc001

### Bug summary

After upgrading the *code* from v9.0.0-beta004 to rc001 and running http://localhost:NNNN/umbraco (instead of running the root site), the DB and nuCache are not upgraded, and the backoffice throws multiple exceptions.

### Specifics

This happens even using a brand new v9.0.0-beta004 site and upgrading (see Steps to reproduce)

If immediately after migration the https://localhost:44390/umbraco URL is opened, the upgrade does not happen and navigating to Settings / Publish Settings throws this exception:

![image](https://user-images.githubusercontent.com/923289/125178272-a3e7aa00-e1b9-11eb-8eb1-d6be6c96d846.png)

The log reports "Panic, exception while loading cache data.","@l":"Fatal","@x":"System.Data.SqlClient.SqlException (0x80131904): Invalid column name 'dataRaw'"

If the app is unloaded and run again, navigating to the root URL (site) does perform the DB and cache migration and everything works fine.

### Steps to reproduce

1. Install a beta004 project from scratch and run the site (the install process is triggered)
```
dotnet new -i Umbraco.Templates::9.0.0-beta004
dotnet new umbraco -n MyWebsite
cd MyWebsite
dotnet run
```

2. Upgrade to rc001 by editing the project file and changing the appropriate line
```
<PackageReference Include="Umbraco.Cms" Version="9.0.0-rc001" />
```

3. Build and run again (note that the Nuget package is updated)
```
dotnet build
dotnet run
```

4. Open the appropriate /umbraco URL, i.e. https://localhost:44390/umbraco

You'll notice that the upgrade process is not launched, and navigating the backoffice throws the errors described above.

 

### Expected result / actual result

The upgrade process should have been launched, as it does on v8 (no matter what URL you use)

PR #11064, opened 2021-09-09
Add redirect to /umbraco if you are not in run state.

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/10650

Added a redirect to root (which redirects to installer) if you are not in runtime state = Run.

### Test
- Open backoffice
- Add a fake migration in Umbraco Plan
- Reboot
- Refresh browser in backoffice 
  -  verify you are redirected to the upgrader.


Zeegaan:
Tested this with a fake migration like suggested, works great 💪 

Code looks good to me 👓

PR #11068, opened 2021-09-09
Fixes reading connection strings after install

Shazwazza:
The issue was discovered when testing Umbraco Id. If you have an existing install and there is a pending Umbraco migration (upgrade), but you clear out the connection string to force a new install screen, fill out the existing connection string details, it all works and redirects to the back office. This will throw exceptions because the runtime state is in an upgrade state (which is fixed in #11064) but if you then logout and try to log in with an active OAuth provider, it means there is a redirect outside of Umbraco and back again and you'll end up back on the installer screen - but it will not show you the upgrade screen, instead it shows you the normal install screen. This is because we are not using IOptionsMonitor for connection string settings which means it's already read the original empty connection string setting fom the very beginning and isn't reading the current/updated value.

We need to review all IOptions usages. Most of them should be IOptionsMonitor unless its impossible to change the app behavior at runtime with a particular config option.

### Prerequisites


bergmania:
Looks good.. 
I have created a task for next sprint to revisit all `IOptions` vs `IOptionsMonitor`. I also think we could benefit from caching the `IOptionsMonitor.CurrentValue` but updating the cache using `IOptionsMonitor.OnChange`

PR #11071, opened 2021-09-09
Fixes empty recycle bin performance with indexing

Shazwazza:
Currently when the recycle bin is empty, it is going to individually delete each item from the index. This is going to cause tons of allocations in Umbraco for DeferedDeleteIndex objects for each item and then down within Examine is going to process each one individually instead of just doing it in bulk. There will be a lot of allocations made there too along with a bunch of extra and unnecessary threads.

@nul800sebastiaan I haven't looked at v8 but it's most likely doing the same thing and this logic should be back ported.


bergmania (approved):
Code changes looks fine and I tried testing emptying recycle bin with both a single item, with many and deleting one item from the recycle bin. All seemed to work still :)

PR #9231, opened 2020-10-20
Listview Template {{ value | ncNodeName }} Problem #9145

cleversolutions:
### Prerequisites

Fixes #9145 

### Description
Fixes 9145 by polling nameExp until it returns a value other than 'Loading...'
It's an ugly way to do this, but the angular infrastructure wouldn't accept a promise from a filter, and setPropertyValues isn't called by interpolation or ng-bind so simply resolving the filter in a safeApply wasn't re-evaluating the list view.

PR #11072, opened 2021-09-10
fix ncNodeName filter broken on refresh

### LINKED ISSUE #9145: Listview Template {{ value | ncNodeName }} Problem
If you follow instructions here:
https://github.com/umbraco/Umbraco-CMS/pull/6597
and create a column in a list view that uses {{ value | ncNodeName }} as the template, it works like so..

image

 until you are viewing the list view and do a hard refresh of the browser.. At which time the items appear as so

image

## Umbraco version

I am seeing this issue on Umbraco version: 8.8

Reproduction
------------

see above.. 

### Specifics
Not sure browser plays a role but I am using the latest version of safari.. and also Firefox..
I have my list view set to display 100 items so performance could be playing a role..

cleversolutions:
### Prerequisites

Fixes #9145 and replaces #9231

### Description
Fixes 9145 by polling nameExp until it returns a value other than 'Loading...'
It's an ugly way to do this, but the angular infrastructure wouldn't accept a promise from a filter, and setPropertyValues isn't called by interpolation or ng-bind so simply resolving the filter in a safeApply wasn't re-evaluating the list view.

This PR replaces #9231 which ended up with a bunch of other commits in it (I likely botched a rebase).

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/listview/listview.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/listview/listview.controller.js
@@ -687,15 +687,30 @@ function listViewController($scope, $interpolate, $routeParams, $injector, $time
             }
 
             if (e.nameExp) {
-                var newValue = e.nameExp({ value });
-                if (newValue && (newValue = newValue.trim())) {
-                    value = newValue;
-                }
+                value = resolveExpression(e, result, value, alias, 0);
             }
 
             // set what we've got on the result
             result[alias] = value;
         });
+    }
+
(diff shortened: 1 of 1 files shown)


jacklawry:
Thanks @cleversolutions - This fix works well for me in v8.16.0

mistyn8:
just wondering if this also has legs.. https://stackoverflow.com/questions/36018721/angular-filter-doesnt-update-when-service-property-changes

I'm trying to use the `ncNodeName filter` in the umbraco forms backoffice in conjunction with `settings(view="picckers.content")`
image
image

But getting the same ...loading issue on refresh, and also when the id is changed by selecting another node.. the id updates but the nodename isn't refreshed.

PR #11075, opened 2021-09-10
Allows replacing MainDom with alternate DB

Shazwazza:
There are some cases where there is a complex hosting strategy and folks want a readonly database and are hosting on Azure. In that case, it is not entirely possible to have a readonly Umbraco database because SqlMainDom is required and part of that requirement is to have read/write access to the umbraco key value table.
This PR allows for the default MainDom to be replaced and to allow for an SqlMainDomLock to use an alternate connection string so that a separate read/write database can be used.

This is specifically regarding these docs: https://our.umbraco.com/Documentation/Fundamentals/Setup/Server-Setup/Load-Balancing/flexible-advanced#front-end-servers---read-only-database-access

> SQL Server Replica databases cannot be used as they are read-only without replacing the default MainDomLock with a custom provider.

But without this PR it's impossible to replace the default MainDomLock. 

To replace it with this PR, you would:

```cs
public class MyUmbracoApplication : UmbracoApplication
{
    protected override IMainDom GetMainDom(ILogger logger)
    {
        // TODO: return custom IMainDom
    }
}
```

then in global.asax, replace with your custom `MyUmbracoApplication`:

```
<%@ Application Inherits="MyApplication.MyUmbracoApplication " Language="C#" %>
```

To continue using the SqlMainDom lock but with a diff connection string, return this in `GetMainDom`:

```cs
return new MainDom(logger, new SqlMainDomLock(logger, "MyConnectionStringName"));
```

Where "MyConnectionStringName" is an alternate connection string in your web.config.
in `GetMainDomLock` create `new SqlMainDomLock(logger, "MyConnectionStringName")`

### DIFF
--- a/src/Umbraco.Core/Runtime/MainDom.cs
+++ b/src/Umbraco.Core/Runtime/MainDom.cs
@@ -19,7 +19,7 @@ namespace Umbraco.Core.Runtime
     /// <para>When an AppDomain starts, it tries to acquire the main domain status.</para>
     /// <para>When an AppDomain stops (eg the application is restarting) it should release the main domain status.</para>
     /// </remarks>
-    internal class MainDom : IMainDom, IRegisteredObject, IDisposable
+    public class MainDom : IMainDom, IRegisteredObject, IDisposable
     {
         #region Vars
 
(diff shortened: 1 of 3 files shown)


nul800sebastiaan:
Cherry picked for 8.17 in 5c0e42e1cbf60b06214a29afb27eb35baf873b7f

PR #11076, opened 2021-09-10
Fixes #9661 boolean conversion

### LINKED ISSUE #9661: Conversion exception silently handled for `IsMemberLockedOut`
## Umbraco version

I am seeing this issue on Umbraco version: branch `netcore\netcore`

Reproduction
------------
I found this while debugging changes to membership and then realised it was already present before the new branch (part of #9472).

### Bug summary

- There is a conversion bug for `IsMemberLockedOut` when trying to save a member as _"No"_ cannot be converted to a Boolean

### Specifics

- The latest `netcore\netcore` branch has this issue
- You need to be debugging with all exceptions on to observe this
- Exceptions shown below
    
### Steps to reproduce

- Attempt to save a member (even though the functionality is not yet implemented in this branch)
- You will hit an issue when debugging which is a conversion bug for `IsMemberLockedOut`

### Expected result

- I would expect the property to convert as expected, to a Boolean

### Actual result

An exception occurs in `src\Umbraco.Core\Models\ContentEditing\MemberSave.cs` on line 45:

image

This is on the property:

`public bool IsLockedOut => GetPropertyValue<bool>(Constants.Conventions.Member.IsLockedOut);`

The issue is:

`System.InvalidCastException: 'Unable to cast object of type 'System.String' to type 'System.Boolean'.'`

The property value is _"No"_ and this doesn't get converted as expected to a Boolean.

This happens in the `ObjectExtensions` line 74 where we try to convert

`public static Attempt<T> TryConvertTo<T>(this object input)`

I think that because it throws an exception that is handled, it actually defaults to _false_ anyway, and so isn't necessarily an issue. However, I would say we should stop the exception happening in the first place.

I'm not sure how to check what happens if the value is _"Yes"_ since this is when a new member is created. 

NB: the same thing doesn't happen on v8, although a similar exception is silently handled when trying to convert an int. 

---
_This item has been added to our backlog [AB#13679](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13679)_

Shazwazza:
Fixes #9661 boolean conversion issue where the property editor is passing in "No" or "Yes" for the value but its not getting converted. The fix is easy to just allow for this conversion with our custom boolean converter.

This may need to be backported, though if this was an issue in v8 we'd surely know about it since it would mean you couldn't lock or unlock members.

### DIFF
--- a/src/Umbraco.Core/CustomBooleanTypeConverter.cs
+++ b/src/Umbraco.Core/CustomBooleanTypeConverter.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.ComponentModel;
 
 namespace Umbraco.Cms.Core
@@ -24,6 +24,8 @@ namespace Umbraco.Cms.Core
                 var str = (string)value;
                 if (str == null || str.Length == 0 || str == "0") return false;
                 if (str == "1") return true;
+                if (str.Equals("Yes", StringComparison.OrdinalIgnoreCase)) return true;
+                if (str.Equals("No", StringComparison.OrdinalIgnoreCase)) return false;
             }
 
             return base.ConvertFrom(context, culture, value);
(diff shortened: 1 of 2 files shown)


bergmania (approved):
Changes look safe to merge :)

PR #11089, opened 2021-09-13
Exclude "Server" header from excessive check if the value is "cloudflare"

Jeavon:
Cloudflare adds the header "Server" with the value "cloudflare". As this header cannot be removed it should be excluded from the healthcheck excessive headers check - https://community.cloudflare.com/t/how-to-remove-server-cloudflare-header/133417

All UmbracoCloud sites that have been migrated to the new infrastructure currently show this healthcheck warning.


nathanwoulfe (approved):
Hey @Jeavon this makes sense to me, thank you 👍

Issue #11070, opened 2021-09-09
Tabs: Delete then navigate gives no warning, tab not deleted.


crgrieve:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

v9 RC003

### Bug summary

If you delete a tab and say "delete" on the pop up, navigate away and return the tab is still there.
When making other changes, like adding a tab it was ask if you are sure you want to navigate away while there are pending changes.

### Specifics

Settings -> Document Types
Edit a doctype

### Steps to reproduce

Delete a tab, select "delete":
![image](https://user-images.githubusercontent.com/9353655/132762717-42a84fb4-e469-4c16-89be-ac850f3fb670.png)
Don't save, navigate away to somewhere else in nav, like "content".
Return to your doctype and the tab is back:
![image](https://user-images.githubusercontent.com/9353655/132762800-d4bec5a3-cf0e-4e16-a9fb-04e55cef371c.png)

### Expected result / actual result

Like with other changes (add tab for example): 
![image](https://user-images.githubusercontent.com/9353655/132762829-8b372eb6-e1b7-4571-a2f5-7dba1136650c.png)

---
_This item has been added to our backlog [AB#13897](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13897)_

madsrasmussen:
PR: https://github.com/umbraco/Umbraco-CMS/pull/11090

nul800sebastiaan:
Fixed in https://github.com/umbraco/Umbraco-CMS/pull/11090 - thanks for the report @crgrieve! 👍

PR #11090, opened 2021-09-13
Set content type form to dirty when add tab, delete tab, add group, and delete group

madsrasmussen:
Fixes: #11070

The PR sets the content type form to dirty when adding a new tab, deleting a tab, adding a group, and deleting a group.

How to test:
* Add tab (don't enter a name) - navigate to another node. Make sure the discard changes dialog is shown
* Do the same for delete tab, add group, delete group.


nul800sebastiaan:
👍 thanks!

PR #11095, opened 2021-09-14
V9 dotnet new template - allow just setting connection string on its own

warrenbuckley:
Remove condition around connection string needed to have all unattended arguments in dotnet new template set such as friendly name, email and password just to set a connection string.

This is for Umbraco cloud team as they want to set a connection string with the dotnet new template but not set the rest of the unattended arguments

## Test
`dotnet new umbraco --name mynewproject --SqlCe --connection-string "Data Source=|DataDirectory|\\Umbraco.sdf;Flush Interval=1;"`

### DIFF
--- a/build/templates/UmbracoProject/appsettings.Development.json
+++ b/build/templates/UmbracoProject/appsettings.Development.json
@@ -17,11 +17,9 @@
       }
     ]
   },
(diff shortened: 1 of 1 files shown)


sitereactor:
Seems legit to me 👍

PR #11102, opened 2021-09-15
Use IFileService to save macro partial views in package migration

ronaldbarendse:
### Prerequisites

### Description
When a package migration tried to save macro partial views from an embedded package.xml file (like in The Starter Kit), it directly used `File.WriteAllText()` instead of the `IFileService.SavePartialViewMacro()`. If the `Views/MacroPartials` directory didn't exist yet, this also caused a `DirectoryNotFoundException` to be thrown (which would cause a `BootFailedException` when the migrations are automatically run).

Testing steps:
1. Before applying this PR on a clean install, remove the `Views/MacroPartials` directory and try to install The Starter Kit (`dotnet add package Umbraco.TheStarterKit --version 9.0.0-rc002`): this will throw an exception while booting up;
2. Manually create the `Views/MacroPartials` directory and start the site: the package migration will now successfully run and the `FeaturedProducts.cshtml` and `LatestBlogposts.cshtml` files should be created in this directory;
3. Apply this PR to a clean install and repeat the first step: it should now run the package migration without issues and the macro partial view files should also exist.

### DIFF
--- a/src/Umbraco.Core/Packaging/PackagesRepository.cs
+++ b/src/Umbraco.Core/Packaging/PackagesRepository.cs
@@ -414,9 +414,10 @@ namespace Umbraco.Cms.Core.Packaging
 
             root.Add(macros);
 
-            // get the partial views for macros and package those
-            IEnumerable<string> views = packagedMacros.Select(x => x.MacroSource).Where(x => x.EndsWith(".cshtml"));
-            PackageMacroPartialViews(views, root);
+            // Get the partial views for macros and package those (exclude views outside of the default directory, e.g. App_Plugins\*\Views)
+            IEnumerable<string> views = packagedMacros.Where(x => x.MacroSource.StartsWith(Constants.SystemDirectories.MacroPartials))
+                .Select(x => x.MacroSource.Substring(Constants.SystemDirectories.MacroPartials.Length).Replace('/', '\\'));
+            PackageStaticFiles(views, root, "MacroPartialViews", "View", _fileSystems.MacroPartialsFileSystem);
         }
 
         private void PackageStylesheets(PackageDefinition definition, XContainer root)
@@ -485,33 +486,6 @@ namespace Umbraco.Cms.Core.Packaging
                 templatesXml.Add(_serializer.Serialize(template));
             }
             root.Add(templatesXml);
-        }
-
-        private void PackageMacroPartialViews(IEnumerable<string> viewPaths, XContainer root)
-        {
-            var viewsXml = new XElement("MacroPartialViews");
-            foreach (var viewPath in viewPaths)
-            {
-                // TODO: See TODO note in MacrosController about the inconsistencies of usages of partial views
-                // and how paths are saved. We have no choice currently but to assume that all views are 100% always
-                // on the content path.
-
-                var physicalPath = _hostingEnvironment.MapPathContentRoot(viewPath);
-                if (!File.Exists(physicalPath))
-                {
-                    throw new InvalidOperationException("Could not find partial view at path " + viewPath);
-                }
-
-                var fileContents = File.ReadAllText(physicalPath, Encoding.UTF8);
-
-                viewsXml.Add(
-                    new XElement(
-                        "View",
-                        new XAttribute("path", viewPath),
-                        new XCData(fileContents)));
-            }
-
-            root.Add(viewsXml);
         }
 
         private void PackageDocumentTypes(PackageDefinition definition, XContainer root)
--- a/src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
+++ b/src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
@@ -122,31 +122,21 @@ namespace Umbraco.Cms.Infrastructure.Packaging
         /// <param name="userId">Optional id of the User performing the operation. Default is zero (admin).</param>
         /// <returns>An enumerable list of generated ContentTypes</returns>
         public IReadOnlyList<IMediaType> ImportMediaTypes(IEnumerable<XElement> docTypeElements, int userId)
-        {
-            return ImportDocumentTypes(docTypeElements.ToList(), true, userId, _mediaTypeService);
-        }
+            => ImportDocumentTypes(docTypeElements.ToList(), true, userId, _mediaTypeService);
 
         #endregion
 
         #region Content
 
-        public IReadOnlyList<T> ImportContentBase<T, S>(
+        public IReadOnlyList<TContentBase> ImportContentBase<TContentBase, TContentTypeComposition>(
             IEnumerable<CompiledPackageContentBase> docs,
-            IDictionary<string, S> importedDocumentTypes,
+            IDictionary<string, TContentTypeComposition> importedDocumentTypes,
             int userId,
-            IContentTypeBaseService<S> typeService,
-            IContentServiceBase<T> service)
-        where T : class, IContentBase
-        where S : IContentTypeComposition
-        {
-            return docs.SelectMany(x => ImportContentBase(
-                x.XmlData.Elements().Where(doc => (string)doc.Attribute("isDoc") == string.Empty),
-                -1,
-                importedDocumentTypes,
-                userId,
-                typeService,
-                service)).ToList();
-        }
+            IContentTypeBaseService<TContentTypeComposition> typeService,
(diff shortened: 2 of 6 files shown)


bergmania:
When testing your PR. I get this error on the starterkit frontpage:
image

```
System.InvalidOperationException: The view 'Components//~/Views/MacroPartials/LatestBlogposts.cshtml' was not found. The following locations were searched:
~/Views/MacroPartials/LatestBlogposts.cshtml
   at Microsoft.AspNetCore.Mvc.ViewEngines.ViewEngineResult.EnsureSuccessful(IEnumerable`1 originalLocations)
   at Microsoft.AspNetCore.Mvc.ViewComponents.ViewViewComponentResult.ExecuteAsync(ViewComponentContext context)
   at Microsoft.AspNetCore.Mvc.ViewComponents.ViewViewComponentResult.Execute(ViewComponentContext context)
   at Umbraco.Cms.Web.Common.Macros.PartialViewMacroEngine.Execute(MacroModel macro, IPublishedContent content)
   at Umbraco.Cms.Web.Common.Macros.MacroRenderer.<>c__DisplayClass24_0.<ExecuteMacroOfType>b__0()
   at Umbraco.Cms.Web.Common.Macros.MacroRenderer.ExecuteProfileMacroWithErrorWrapper(MacroModel macro, String msgIn, Func`1 getMacroContent, Func`1 msgErr)
   at Umbraco.Cms.Web.Common.Macros.MacroRenderer.ExecuteMacroWithErrorWrapper(MacroModel macro, String msgIn, String msgOut, Func`1 getMacroContent, Func`1 msgErr)
   at Umbraco.Cms.Web.Common.Macros.MacroRenderer.ExecuteMacroOfType(MacroModel model, IPublishedContent content)
   at Umbraco.Cms.Web.Common.Macros.MacroRenderer.RenderAsync(MacroModel macro, IPublishedContent content)
   at Umbraco.Cms.Web.Common.Macros.MacroRenderer.RenderAsync(String macroAlias, IPublishedContent content, IDictionary`2 macroParams)
   at Umbraco.Cms.Core.Templates.UmbracoComponentRenderer.RenderMacroAsync(IPublishedContent content, String alias, IDictionary`2 parameters)
   at Umbraco.Cms.Core.Templates.UmbracoComponentRenderer.RenderMacroAsync(Int32 contentId, String alias, IDictionary`2 parameters)
   at Umbraco.Cms.Web.Common.UmbracoHelper.RenderMacroAsync(String alias, IDictionary`2 parameters)
   at AspNetCore.Views_Partials_grid_editors_macro.ExecuteAsync() in C:\Users\Bjarke\TestRonald\Views\Partials\grid\editors\macro.cshtml:line 13
   at Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderPageCoreAsync(IRazorPage page, ViewContext context)
   at Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderPageAsync(IRazorPage page, ViewContext context, Boolean invokeViewStarts)
   at Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderAsync(ViewContext context)
   at Microsoft.AspNetCore.Mvc.ViewFeatures.HtmlHelper.RenderPartialCoreAsync(String partialViewName, Object model, ViewDataDictionary viewData, TextWriter writer)
   at Microsoft.AspNetCore.Mvc.ViewFeatures.HtmlHelper.PartialAsync(String partialViewName, Object model, ViewDataDictionary viewData)
   at AspNetCore.Views_Partials_grid_editors_base.ExecuteAsync() in C:\Users\Bjarke\TestRonald\Views\Partials\grid\editors\base.cshtml:line 6
```

bergmania:
Also the MacroPartialViews are saved incorrectly

image

bergmania:
Not sure how to best solve this.. 
- IMO the packages should not be created with the full path, but only the relative path after `~/Views/MacroPartials`. But if we assume this, we will break existing packages. So we should most likely handle both cases in the import feature

ronaldbarendse:
@bergmania It's indeed inconsistent, as this todo comment points out: https://github.com/umbraco/Umbraco-CMS/blob/f68717c1c8c7dfe994f5c8931052c9c8b793e2b3/src/Umbraco.Web.BackOffice/Controllers/MacrosController.cs#L318-L322

While selecting/picking a macro partial view, it can come from the `Views\MacroPartials` directory (which is editable from the tree and is using the `MacroPartialsFileSystem`) or from `App_Plugins\*\Views` (provided by an external package). We should only include the ones from `Views\MacroPartials` in the exported XML, as the others would/should already be included in a package (as NuGet content file).

I've fixed this by only exporting macro partial views from the `Views\MacroPartials` directory and removing the prefix. For backwards compatibility, the prefix is also removed while importing the package XML and macro partial views outside of the MacroPartialsFileSystem are now skipped (a warning is logged when this happens).

src/Umbraco.Core/Packaging/PackagesRepository.cs
```diff
@@ -414,9 +414,10 @@ private void PackageMacros(PackageDefinition definition, XContainer root)
 
             root.Add(macros);
 
-            // get the partial views for macros and package those
-            IEnumerable<string> views = packagedMacros.Select(x => x.MacroSource).Where(x => x.EndsWith(".cshtml"));
-            PackageMacroPartialViews(views, root);
+            // Get the partial views for macros and package those (exclude views outside of the default directory, e.g. App_Plugins\*\Views)
+            IEnumerable<string> views = packagedMacros.Where(x => x.MacroSource.StartsWith(Constants.SystemDirectories.MacroPartials))
+                .Select(x => x.MacroSource.Substring(Constants.SystemDirectories.MacroPartials.Length).Replace('/', '\\'));
```
ronaldbarendse, line 419: This does the filtering on and removing of the prefix. It also replaces forward slashes to backward slashes, as the virtual path is constructed differently compared to the normal views, so this aligns everything a bit more (both are handled correctly though).

src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
```diff
@@ -1280,29 +1255,46 @@ public IReadOnlyList<ILanguage> ImportLanguages(IEnumerable<XElement> languageEl
...
+                if (path.StartsWith(Constants.SystemDirectories.MacroPartials))
                 {
-                    System.IO.File.WriteAllText(physicalPath, contents, Encoding.UTF8);
+                    path = path.Substring(Constants.SystemDirectories.MacroPartials.Length);
+                }
+                else if (path.StartsWith("~"))
+                {
+                    _logger.LogWarning("Importing macro partial views outside of the Views/MacroPartials directory is not supported: {Path}", path);
+                    continue;
+                }
```
ronaldbarendse, line 1282: This is only required to maintain backwards compatibility with package XMLs that still contain the virtual path.

src/Umbraco.Infrastructure/Packaging/PackageInstallation.cs
```diff
@@ -44,20 +44,29 @@ public InstallationSummary InstallPackageData(CompiledPackage compiledPackage, i
...
-            var contentInstalled = installationSummary.ContentInstalled.ToList();
-            packageDefinition.ContentNodeId = contentInstalled.Count > 0 ? contentInstalled[0].Id.ToInvariantString() : null;
+
+            foreach (var x in installationSummary.MediaTypesInstalled)
+                packageDefinition.MediaTypes.Add(x.Id.ToInvariantString());
+
+            packageDefinition.ContentNodeId = installationSummary.ContentInstalled.FirstOrDefault()?.Id.ToInvariantString();
+            
+            foreach (var x in installationSummary.MediaInstalled)
+                packageDefinition.MediaUdis.Add(x.GetUdi());
```
ronaldbarendse, line 69: I'm not sure where the updated package definition is used, but I've added media and removed the stylesheets as they aren't tracked by ID.

So the following types aren't currently tracked, because that would require serializing them to XML (that includes the path and file contents):
- `MacroPartialViewsInstalled`
- `StylesheetsInstalled`
- `ScriptsInstalled`
- `PartialViewsInstalled`

ronaldbarendse: The package definition is now also updated with the imported stylesheets, scripts and partial views paths. Macro partial views are part of the macro, so isn't defined as separate entity in the package definition, which is fine I guess.

bergmania (approved):

PR #11107, opened 2021-09-15
Bugfix: Accessibility in Tabs - Ensure the selected tab is properly exposed to aria-selected

BatJan:
…has a proper true/false value

### Prerequisites

Playing around with the new tabs I by coincidence noticed that the `aria-selected` value was rendering `{tab.active}` instead of either `true/false` due to the expression missing a set of extra curly braces so the code would read `aria-selected="{{tab.active}}"` in order to render `true/false`.


BatJan:
@nul800sebastiaan I think it would be good if this could make it into the final release of v8.17 and v9 🙏🏻

mikecp (approved):

mikecp:
Hello @BatJan ,
Thanks for your PR and H5YR for spotting this !! 👍
Let's merge it to make Umbraco more accessible 😉

PR #11113, opened 2021-09-16
Fix incorrect attribute in hidden prevalue attribute

bjarnef:
### Prerequisites

### Description
At some point the hidden prevalue editor has changed, so it set name attribute twice and not id attribute.

Strange nobody has noticed this since in was added in https://github.com/umbraco/Umbraco-CMS/commit/5b289a4971ccb0f6d431377e9e199c6d27bb2839#diff-5712c2b4c44f9e8368d7437b866a916290dfcd5814e062c74ed99362329ca19eR1

It seems the latter one would be ignored https://stackoverflow.com/a/26341866/1693918 .. but you never know if that's the case in future browsers.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/hidden.html
+++ b/src/Umbraco.Web.UI.Client/src/views/prevalueeditors/hidden.html
@@ -1,1 +1,1 @@
-<input name="hidden" type="hidden" ng-model="model.value" name="{{model.alias}}" /> 
+<input name="hidden" type="hidden" ng-model="model.value" id="{{model.alias}}" />


bjarnef:
@nul800sebastiaan could we fix this up? Maybe cherry pick in upcoming patch releases?

bjarnef:
@Shazwazza I guess it should have been `id={{model.alias}}` when you added in back in 2013 😁

bjarnef:
@bergmania it would be great to fix this in v9 as well 🤓😁👍

nathanwoulfe (approved):
Hey @bjarnef I can only find a single reference to this, in the Email Address property editor, where the prevalue is marked as obsolete and to be removed in 9 (whether that has happened I haven't checked). Fix still makes sense, even if it's got a short lifespan 😃

bjarnef:
@nathanwoulfe yeah, but prevalue editors can still be used by developers in e.g. grid editor and property editors. Not sure the use-case where you want to use this hidden prevalue editor though. From what I have read browsers ignore the latter attribute(s) in case there are duplicates.

nul800sebastiaan:
Cherry picked for 8.17 in 7548074fa390e44a706eb4e98bfccbfff27a093a

PR #11127, opened 2021-09-17
Add LocalDB database install option and implement automatic database creation

ronaldbarendse:
### Prerequisites

### Description
This PR adds a new database option for SQL Server Express (LocalDB) to the installer screen:
Configure your database

This option is only shown if supported (running Windows with [SQL Server Express with Advanced Services](https://docs.microsoft.com/en-us/sql/database-engine/configure-windows/sql-server-express-localdb?view=sql-server-ver15) installed) and is then also used by default when you select Install on the first screen. If it's unsupported, SQL CE is used on Windows and otherwise you'll be asked to enter a manual connection string.

~_Note: The `|DataDirectory|` placeholder in the connection string is currently only supported for SQL CE (because the SQL CE engine has built-in support for replacing this placeholder with the value we set during boot using `AppDomain.CurrentDomain.SetData()` - which is currently `~/umbraco/Data`). For LocalDB the absolute physical file path is written in the connection string, making it less portable._~ This is now also fixed in this PR.

----------------

Because the connection string can be read from multiple configuration sources and the installer writes it to the `appsettings.json` during install, this PR also removes the code that treated a non-existent SQL CE database file as an empty connection string. This could otherwise cause an install loop if you'd have a SQL CE connection string configured in e.g. `appsettings.Development.json` that referenced a non-existent database file.

~To automatically create the database file, you must now set `Umbraco:CMS:Global:InstallMissingDatabase=true`. This will work for SQL CE, LocalDB and also normal SQL Server (assuming you have the rights to create new databases).~

Together with the [unattended install](https://our.umbraco.com/Documentation/Fundamentals/Setup/Install/Unattended-Install-v9), just setting a connection string and correct `Umbraco:CMS:Unattended` options, you'll be able to have a clean site installed on the first startup.

------------

This contains the following breaking changes:
- A non-existent SQL CE database file isn't treated as an empty connection string anymore (but it is still automatically created)
- The `IEmbeddedDatabaseCreator` is renamed and slightly refactored to `IDatabaseCreator`

~**TODO: Instead of checking the `InstallMissingDatabase` setting in multiple places, we should use `RuntimeLevelReason.InstallMissingDatabase` instead any probably not require this setting to be enabled when the current connection string is set to SQL CE or a LocalDB database.**~


ronaldbarendse:
Both SQL CE and LocalDB databases are now automatically created, so the `InstallMissingDatabase` setting is only required if you want to create the database on a full SQL Server instance (although you obviously need the `CREATE DATABASE` permission).

The `|DataDirectory|` placeholder is now also replaced in a LocaDB connection string, so you can add the following in your `appsettings.json` before starting Umbraco for the first time and have Umbraco installed using LocalDB: `Server=(localdb)\\MSSQLLocalDB;Integrated Security=true;AttachDbFileName=|DataDirectory|\\Umbraco.mdf`.

If you combine this with the unattended install option, you can have Umbraco installed and ready to login on the first request! 🎉 You can even have The Starter Kit installed (as NuGet package) and have a site ready for testing (which is super handy for testing PRs)! 🤯

src/Umbraco.Core/Configuration/Models/ConnectionStrings.cs
```diff
@@ -26,6 +26,6 @@ private string umbracoDbDSN
         /// <summary>
         /// Gets or sets a value for the Umbraco database connection string..
         /// </summary>
-        public ConfigConnectionString UmbracoConnectionString { get; set; }
+        public ConfigConnectionString UmbracoConnectionString { get; set; } = new ConfigConnectionString(Constants.System.UmbracoConnectionName, null);
```
ronaldbarendse, line 29: If the `umbracoDbDSN` didn't exist in `appsettings.json`, this property was `null` and caused a lot of `NullReferenceException`s to be thrown. Although it would be better to do proper null-checks (as you could theoretically set it to `null` yourself), this was an easy fix for the initial issue.

src/Umbraco.Infrastructure/Persistence/SqlServerDatabaseCreator.cs
```diff
@@ -0,0 +1,63 @@
...
+                    database = "Umbraco-" + Guid.NewGuid();
+                }
+
+                using var connection = new SqlConnection(masterConnectionString);
+                connection.Open();
+
+                using var command = new SqlCommand(
+                    $"CREATE DATABASE [{database}] ON (NAME='{database}', FILENAME='{fileName}');" +
+                    $"ALTER DATABASE [{database}] SET SINGLE_USER WITH ROLLBACK IMMEDIATE;" +
+                    $"EXEC sp_detach_db @dbname='{database}';",
```
ronaldbarendse, line 42: We might need some additional checks/sanitation on these SQL queries, as you can't use parameters in DDL queries.

src/Umbraco.Infrastructure/Persistence/SqlServerDbProviderFactoryCreator.cs
```diff
@@ -3,12 +3,12 @@
 using System.Linq;
 using Microsoft.Extensions.Options;
 using NPoco;
-using Umbraco.Cms.Core.Composing;
 using Umbraco.Cms.Core.Configuration.Models;
 using Umbraco.Cms.Infrastructure.Persistence.SqlSyntax;
 
 namespace Umbraco.Cms.Infrastructure.Persistence
 {
+    [Obsolete("This is only used for integration tests and should be moved into a test project.")]
```
ronaldbarendse, line 11: Or remove this class entirely?

ronaldbarendse:
As a proof of concept, I've created a composer that automatically configures a LocalDB or SQL CE connection string (depending on which is available) and enables unattended install, so just adding the following class into your project would result in a clean Umbraco install on the first request. If you also want to configure the administrator account, you just have to configure the right values in `Umbraco:CMS:Unattended`:

```c#
using Microsoft.Extensions.DependencyInjection;
using Umbraco.Cms.Core;
using Umbraco.Cms.Core.Configuration;
using Umbraco.Cms.Core.Configuration.Models;
using Umbraco.Cms.Infrastructure.Install.InstallSteps;
using Umbraco.Cms.Infrastructure.Migrations.Install;
using Umbraco.Extensions;

namespace Umbraco.Cms.Web.UI
{
    public class AutoSetupComposer : IComposer
    {
        {
            builder.Services.PostConfigure<ConnectionStrings>(options =>
            {
                if (!options.UmbracoConnectionString.IsConnectionStringConfigured())
                {
                    if (DatabaseConfigureStep.IsLocalDbAvailable())
                    {
                        options.UmbracoConnectionString = new ConfigConnectionString(Constants.System.UmbracoConnectionName, DatabaseBuilder.LocalDbConnectionString, Constants.DbProviderNames.SqlServer);
                    }
                    else if (DatabaseConfigureStep.IsSqlCeAvailable())
                    {
                        options.UmbracoConnectionString = new ConfigConnectionString(Constants.System.UmbracoConnectionName, DatabaseBuilder.EmbeddedDatabaseConnectionString, Constants.DbProviderNames.SqlCe);
                    }
                }
            });

            builder.Services.PostConfigure<UnattendedSettings>(options => options.InstallUnattended = true);
        }
    }
}

```

bergmania:
Before we test and review, I think the unit tests should be fixed.

ronaldbarendse:
@bergmania I've updated the parsing of the connection string to only mutate it when it's replacing the `|DataDirectory|` placeholder, as `DbConnectionStringBuilder` lowercases all properties. This should fix the unit tests 👍🏻 

It would be better if we could use the provider specific connection string builders, but that would require taking dependencies on additional assemblies in `Umbraco.Core`. I've at least ensured our default connection strings match those from the parsed/built ones:

```c#
new System.Data.SqlServerCe.SqlCeConnectionStringBuilder
{
    ConnectionString = @"Data Source=|DataDirectory|\Umbraco.sdf;Flush Interval=1"
}.ToString();

new System.Data.SqlClient.SqlConnectionStringBuilder
{
    ConnectionString = @"Data Source=(localdb)\MSSQLLocalDB;AttachDbFilename=|DataDirectory|\Umbraco.mdf;Integrated Security=True"
}.ToString();
```

src/Umbraco.Web.UI.Client/src/installer/steps/database.html
```diff
@@ -10,50 +10,61 @@ <h1>Configure your database</h1>
...
 
-        <div class="controls" ng-if="installer.current.model.dbType == 'SqlCe' ">
+         <div class="controls" ng-if="installer.current.model.dbType == 'SqlLocalDb'">
             <p>Great! No need to configure anything, you can simply click the <strong>continue</strong> button below to continue to the next step</p>
         </div>
 
-        <div ng-if="installer.current.model.dbType == 'Custom' ">
+        <div class="controls" ng-if="installer.current.model.dbType == 'SqlCe'">
+            <p>Great! No need to configure anything, you can simply click the <strong>continue</strong> button below to continue to the next step</p>
+        </div>
```
bergmania, line 27: This could also be shown for `installer.current.model.dbType == 'SqlLocalDb'`

src/Umbraco.Infrastructure/Persistence/SqlServerDbProviderFactoryCreator.cs
bergmania: Obsolete is okay, I fear some packages could use it, so we keep it for v9, and move for v10

ronaldbarendse:
To show the power of the new CLI commands and `dotnet new` templates together with this feature; this is all that's needed to get a clean Umbraco 9 site (including a back-office user) - with The Started Kit and Forms installed - on the first web request:

```powershell
# Ensure we have the latest Umbraco templates
dotnet new -i Umbraco.Templates::9.0.0

# Create solution/project
dotnet new sln -n Umbraco9
dotnet new umbraco -n Umbraco9 --friendly-name "The Rabbit" --email "rabbit@umbraco.com" --password "1234567890" --connection-string "Data Source=(localdb)\MSSQLLocalDB;AttachDbFilename=|DataDirectory|\Umbraco.mdf;Integrated Security=True"
dotnet sln add Umbraco9
dotnet add Umbraco9 package Umbraco.TheStarterKit --version 9.0.0
dotnet add Umbraco9 package Umbraco.Forms --version 9.0.0

# Run
dotnet run --project Umbraco9
```

If you want to use SQL CE instead of LocalDB, you have include the `-ce`/`--SqlCe` flag and provide the correct connection string:
```powershell
dotnet new umbraco -n Umbraco9 --friendly-name "The Rabbit" --email "rabbit@umbraco.com" --password "1234567890" --connection-string "Data Source=|DataDirectory|\Umbraco.sdf;Flush Interval=1" -ce
```

PR #11130, opened 2021-09-17
UmbracoRouteValueTransformer fixes

Shazwazza:
There are some issues with our current implementation which make it almost impossible for other DynamicRouteValueTransformer to play nicely with Umbraco especially those that integrate with Umbraco using `UmbracoRouteValues` (i.e. Articulate).

* UmbracoRouteValueTransformer as singleton - this will save memory, there only needs to be one instance and it doesn't need to be scoped. It was only scoped based on some old example code in the aspnet repo.
* Return `null` when we don't match. See https://github.com/dotnet/aspnetcore/pull/36653.
* Don't execute if there is already `UmbracoRouteValues` available. In this case it would mean that another `DynamicRouteValueTransformer` has already executed and routed an Umbraco route and we shouldn't overwrite it.
* Don't modify the passed in `values`, these should not be changed and we should be returning new values, see: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.routing.dynamicroutevaluetransformer.transformasync?view=aspnetcore-5.0#parameters 
   > Implementations should not modify values.
* Return null if we are on 404 and no 404 handlers or last chance finders resolved content
* Deal with a true 404 in a custom IEndpointSelectorPolicy

The way we were using DynamicRouteValueTransformer meant that any other DynamicRouteValueTransformer  registered or ordered after ours will not have any affect. It will execute and new values can be returned but they will not be used (by default).

This lead me on some additional investigation - how to make 2 or more DynamicRouteValueTransformer that match the request execute the "last matched"? By default it seems that the first match always wins. Can this be changed? It turns out it can by implementing a `IEndpointSelectorPolicy` for which there are no docs (see: https://github.com/dotnet/AspNetCore.Docs/issues/12593). But there are some examples and mentions of how to make it work here: https://github.com/dotnet/aspnetcore/issues/10574#issuecomment-497195495. Why would we want this? Here's an example:

* Since Umbraco has a catch all dynamic route and never returns null even for 404s, any additional DynamicRouteValueTransformer registered that executes after Umbraco will just not work.
* In some cases, like Articulate, it would be preferable to check if Umbraco matches a real document, else fallback to a custom DynamicRouteValueTransformer to perform additional matching.

So more ideally would be that umbraco's DynamicRouteValueTransformer returns null when there is no content match at all so that any additional DynamicRouteValueTransformer just work as intended. So the final piece of the puzzle is then, how do we direct true 404s (not handled at all by the end user) to our custom 'page is left ugly' route = custom IEndpointSelectorPolicy.

## Testing

* Configure a 404 handler in the content settings like https://our.umbraco.com/Documentation/Tutorials/Custom-Error-Pages/index-v9#set-a-custom-404-page-in-appsettingsjson and make sure it works on a not found page.
* Remove this setting so there is no configured 404 page, then go to a page that doesn't exist and you should still get the 'this page is ugly'...

### DIFF
--- a/src/Umbraco.Tests.UnitTests/Umbraco.Web.Website/Routing/UmbracoRouteValueTransformerTests.cs
+++ b/src/Umbraco.Tests.UnitTests/Umbraco.Web.Website/Routing/UmbracoRouteValueTransformerTests.cs
@@ -14,6 +14,7 @@ using NUnit.Framework;
 using Umbraco.Cms.Core;
 using Umbraco.Cms.Core.Configuration.Models;
 using Umbraco.Cms.Core.Events;
+using Umbraco.Cms.Core.Models.PublishedContent;
 using Umbraco.Cms.Core.PublishedCache;
 using Umbraco.Cms.Core.Routing;
 using Umbraco.Cms.Core.Services;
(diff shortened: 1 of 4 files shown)


bergmania (approved):
Code changes makes sense based on all your findings.. Off to test :)

bergmania:
Seems to work as expected.. One note for potential documentation is that Chrome/Edge prints there own 404 page, if the template is empty. Like in only this

```razor
@using Umbraco.Cms.Web.Common.PublishedModels;
@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<ContentModels.ErrorPage>
@using ContentModels = Umbraco.Cms.Web.Common.PublishedModels;
@{
	Layout = null;
}
```

Shazwazza:
That 404 page isn't the one we ship with though, it should have the "this page is left ugly..." text. I'll re-test this all today again.

bergmania:
These screenshots are with a custom 404 page configured. I had the "this page is left ugly." when it was not configured

Shazwazza:
ahh, ok then :)

Issue #11083, opened 2021-09-12
Nested content missing properties on Member edit


AaronSadlerUK:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0-RC3

### Bug summary

If you add a nested content property to a member type, when you try to add an item the properties will be missing.

![image](https://user-images.githubusercontent.com/9937803/133003238-3182e6ef-cfce-4965-983c-57289e0ed10b.png)

### Steps to reproduce

Create a nested element
Add a nested content property to a member type
Try to edit a member

### Expected result / actual result

The nested content properties to display

OzoneNZ:
Can confirm Nested Content editors are broken throughout our app on `9.0.0-rc003`, as well as in a fresh installation

**Element:**

![image](https://user-images.githubusercontent.com/4223195/133030507-33db540e-e0d0-492a-b938-0ec9f22b3a82.png)

**Composition:**

![image](https://user-images.githubusercontent.com/4223195/133030614-34c866b0-2dbf-488f-8d14-fd84268dd9f2.png)

**Content:**

![image](https://user-images.githubusercontent.com/4223195/133030558-a66261ad-02f9-462a-96b1-a113eb388861.png)

May have something to do with the introduction of tabs?

nul800sebastiaan:
FYI - Nested Content on Member Types does work in v8.

OzoneNZ:
Not exactly a definitive diagnosis, but it seems this was broken by something within #11046 / 1d55d50

OzoneNZ:
Have done some more digging into this, `nestedcontent.controller.js` is falling over around line 538:

```javascript
var tab = _.find(tabs, function (tab) {
    return tab.id !== 0 && (tab.alias.toLowerCase() === contentType.ncTabAlias.toLowerCase() || contentType.ncTabAlias === "");
});
```

![image](https://user-images.githubusercontent.com/4223195/133106284-9e717bf3-a039-4092-9f8e-8efe352d6ec6.png)

I wouldn't have a clue what the fix is, but I'm assuming `contentType.ncTabAlias` doesn't have the correct value (doesn't look like an alias)?

bergmania:
Thanks for the investigation @OzoneNZ

PR #11145, opened 2021-09-21
Fixes Nested Content not rendering

### LINKED ISSUE #11114: v9: Nested Content not rendered when group name contains "System".

v9 RC3

### Bug summary

If you have nested content with a group name that contains **System**, the properties don't get rendered:
image

### Specifics

_No response_

### Steps to reproduce

- Create a composition.
- Create a group and name it **Test System**.
- Add any property (doesn't matter).
- Create content type from said composition.
- Create nested content with said content type.
- Add nested content to a document type.
- Go to content and try to add the nested content.
- Notice that nothing gets rendered apart from "Item 1".

This should work, it does in Umbraco 8.

---
_This item has been added to our backlog [AB#13949](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/13949)_

madsrasmussen:
This PR fixes the rendering of nested content after the tabs feature has been added. To prevent any breaking changes, we had to limit the allowed nested content groups to generic groups. Tabs and groups inside a tab are not supported in nested content. The Block List Editor can be used if tab support is needed.

**How to test:**
* Create an element type with a generic group, a tab, and a group inside the tab
* Set up a doctype with nested content.
* Make sure you can only select the group from "Generic" in nested content.
* Make sure nested content renders correctly in the content section. 

fixes #11114

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/propertyeditors/nestedcontent/nestedcontent.controller.js
@@ -535,7 +535,7 @@
                     // remove all tabs except the specified tab
                     var tabs = scaffold.variants[0].tabs;
                     var tab = _.find(tabs, function (tab) {
-                        return tab.id !== 0 && (tab.alias.toLowerCase() === contentType.ncTabAlias.toLowerCase() || contentType.ncTabAlias === "");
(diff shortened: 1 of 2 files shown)


Zeegaan:
Tested this both in v8 and v9 and it works great! 

Code looks good to me 🤓

Zeegaan (approved):

Zeegaan:
This PR also fixes: https://github.com/umbraco/Umbraco-CMS/issues/11083

PR #11146, opened 2021-09-21
Handle maindom was acquired doing install, so we do not have our lock…

bergmania:
Handle maindom was acquired doing install, so we do not have our lock in table.. We now insert it in the listening loop if missing


AndyButland (approved):
Looks fine - just made a couple of tidy-up suggestions.

I take it in the end you didn't see the need to move the acquire main dom step before the unattended install?

This looks a clear improvement/fix to me - but please let me know if you are happy to merge this on visual code inspection, or whether you'd like me to take a custom build and use in a Cloud project to verify it fixes the environment setup issue we found there.

bergmania:
> I take it in the end you didn't see the need to move the acquire main dom step after the unattended install?

It was not possible :)

bergmania:
> This looks a clear improvement/fix to me - but please let me know if you are happy to merge this on visual code inspection, or whether you'd like me to take a custom build and use in a Cloud project to verify it fixes the environment setup issue we found there.

We can merge as soon as the buildserver is happy

PR #11151, opened 2021-09-21
v9: fixed tabs test on every resolution

Zeegaan:
# Notes
- Updated tabs test to look for the caret element, and only click it if its active 
# How to test
- Go to Umbraco.Tests.AcceptanceTest and open cypress.json
- Change the height and width of the viewport to 1200 x 1024
- Open up the command prompt in the same folder
- Run `npm install` and `npm run tests`
- All tests should pass


bergmania:
Looks good.. I removed the release label, as these changes do not affect the released bits (just testing them :))

PR #11153, opened 2021-09-21
Fixes persisting the IsApproved for users

Shazwazza:
Setting IsApproved on the `BackOfficeIdentityUser` didn't persist when saving the member. This fixes that and adds an integration test for it.


bergmania (approved):
Changes looks good :)

PR #11155, opened 2021-09-21
Fixes 10730 - Route hijacking with public access

### LINKED ISSUE #10730: Route hijacking with restricted public access - protected page request is not handled by custom login page controller

9.0.0-rc001

### Bug summary

In v8, when a custom controller is in place for route hijacking on a member login page, a request to a protected page is handled by the custom member login page controller, and returns the login page template. 

In v9, a similar request, whilst still returning the login page template, is handled by the default render controller. In other words, the custom login page controller only works of you navigate direct to the login page - it does not work if the login template is presented due to access being restricted to the requested page.

### Specifics

My login page document type is **MemberLoginPage**
In v8, my route hijacking controller is like:
```
public class MemberLoginPageController : RenderMvcController
{
    public override ActionResult Index(ContentModel model)
    {
        // do custom stuff here
        return CurrentTemplate(model);
    }
}
```
In v9, it's
```
public class MemberLoginPageController : RenderController
{
    public MemberLoginPageController(
        ILogger<MemberLoginPageController> logger,
        ICompositeViewEngine compositeViewEngine,
        IUmbracoContextAccessor umbracoContextAccessor)
        : base(logger, compositeViewEngine, umbracoContextAccessor)
    {
    }

    public override IActionResult Index()
    {
        // do custom stuff here
        return CurrentTemplate(CurrentPage);
    }
}
```

### Steps to reproduce

Create 3 document types for HomePage, TextPage and MemberLoginPage. Create content, with Restrict Public Access on the Restricted Access TextPage node like this:
Screenshot 2021-07-23 150742
In Visual Studio, create a custom controller called MemberLoginPageController (as above in Specifics), and place a breakpoint on the return line in the Index() method. Debug the solution, and browse to /login/ and /restricted-access/ in turn.

I'd expect the breakpoint to be hit when browsing to both /login/ and /restricted-access/, but the breakpoint is only hit when browsing to /login/.

Shazwazza:
Fixes #10730 - Route hijacking with public access

The reason why it doesn't work is because DynamicRouteValueTransformer is responsible for assigning the controller/action tokens. ASP.NET then uses it's response to map to an endpoint (controller/action). Our public access middleware executes after the transformer and re-assigns our own UmbracoRouteValues so that the correct page is rendered, but this doesn't affect the results returned from our transformer - which is why the original endpoint is selected.

The endpoint selection is stored in the HttpContext as a feature: IEndpointFeature. It may be possible to replace this before endpoint is executed within the middleware but this is against best practices (undocumented) and will probably result in some caching not being done behind the scenes in aspnet.

The only real way to deal with this is going to be to move the public access code directly up into the original routing pipeline and not use middleware so that we can return the correct route values from within our route transformer so aspnet selects the correct endpoint from that data.

* Changes PublicAccessMiddleware to IPublicAccessRequestHandler and puts the logic for this directly inside of the transformer
* The IPublicAccessRequestHandler no longer sets the httpfeature for UmbracoRouteValues, it just returns new ones or the original ones
* The transformer logic rearranged a bit:
  *  We were checking for posted values even if there was no content matched, so the no content check was moved above
  * We check public access before checking posted values - since this would have also occurred with the old middleware solution

## Testing

* Follow the repro steps in #10730 and make sure it works
* Make sure that a surface controller POST form still works as expected

### DIFF
--- a/src/Umbraco.Tests.UnitTests/Umbraco.Web.Website/Routing/UmbracoRouteValueTransformerTests.cs
+++ b/src/Umbraco.Tests.UnitTests/Umbraco.Web.Website/Routing/UmbracoRouteValueTransformerTests.cs
@@ -48,6 +48,10 @@ namespace Umbraco.Cms.Tests.UnitTests.Umbraco.Web.Website.Routing
             IPublishedRouter router = null,
             IUmbracoRouteValuesFactory routeValuesFactory = null)
         {
+            var publicAccessRequestHandler = new Mock<IPublicAccessRequestHandler>();
+            publicAccessRequestHandler.Setup(x => x.RewriteForPublishedContentAccessAsync(It.IsAny<HttpContext>(), It.IsAny<UmbracoRouteValues>()))
+                .Returns((HttpContext ctx, UmbracoRouteValues routeVals) => Task.FromResult(routeVals));
+
             var transformer = new UmbracoRouteValueTransformer(
                 new NullLogger<UmbracoRouteValueTransformer>(),
                 ctx,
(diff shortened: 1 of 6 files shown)


bergmania:
I can confirm this fixes #10730 and surface controllers do still work as expected.. Code changes also makes good sense, with this description :)

Amazing work - as always 💪

PR #11157, opened 2021-09-22
Fix Multi URL Picker in Block List Editor

### LINKED ISSUE #11100: Multi-URL Picker does not save values when used in a nested block list item when using the new tabs feature

Umbraco v9 RC3

### Bug summary

Multi-URL Picker does not save values when used in a nested block list item when using the new tabs feature. The picker is empty after save. The property editors should be directly under the tab, not in a standard panel.

### Specifics

_No response_

### Steps to reproduce

Add a multi-URL picker to a block list item child of another block list item and try to save data. The block list items should both use the new tabs feature.

_No response_

madsrasmussen:
fixes #11100

Properties from tabs were rendered twice. One for direct properties for a tab and one for groups because they are part of the groups' array. This PR prevents a group to be rendered in the groups' array if it is a "tab".

Multi URL Picker listens for a form submit event and sets the value based on a render value. Because the property is rendered twice, when it is a direct property on a tab, it resets its own value. 

The reset only happens in the Block List editor because of how it syncs its values. I have added the fix to the content, media, and members section for performance reasons. We don't want more properties in the DOM than needed.

How to test:
* See the issue for testing instructions for the bug. 
* Please also do some "general" testing of tabs and groups in content, media and member sections and this changes how we render them.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/components/content/umb-tabbed-content.html
+++ b/src/Umbraco.Web.UI.Client/src/views/components/content/umb-tabbed-content.html
@@ -29,7 +29,8 @@
             data-app-anchor="{{group.id}}"
             data-element="group-{{group.alias}}"
             ng-repeat="group in content.tabs track by group.key"
-            ng-show="(group.parentAlias === activeTabAlias && group.type === 0) || tabs.length === 0">
+            ng-if="group.type === 0"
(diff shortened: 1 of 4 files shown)


Zeegaan:
Tested this and it works great! 👍 

Code looks good to my untrained frontend eyes 👓

PR #11162, opened 2021-09-22
Create BackOfficeIdentityUser with IsApproved property

sitereactor:
When signing into Umbraco Cloud with Umbraco Id the user is automatically created in the Umbraco Backoffice, but right now starts off with being disabled.

The `IsApproved` property is handled (ref.: https://github.com/umbraco/Umbraco-CMS/pull/11153) when the `BackOfficeIdentityUser` is updated, but not upon creation.


sitereactor:
This won't be needed, as we found the real issue in the Umbraco Id autolinker.

PR #11163, opened 2021-09-22
Serializes key for templates in package definition and imports if ava…

AndyButland:
Noticed in testing that whilst most entities serialized to the `package.xml` file have keys provided - so when installed in new environments they have the same unique identifier - for templates this isn't the case.

This can cause issues for Umbraco Deploy, as, if you install a package locally and then deploy, the package migration will run and templates with the same details but different unique Ids will be created in the upstream environment.

This PR adds keys to the serialized XML, and if found on import, uses them to create the new template.

To Test:

- Create and download a package XML file from the back-office with a template in it.
- Delete the template.
- Install the package, check the template has been created.
- Check in the database to see that the unique of the installed package is the same as that serialized to the XML file.

```
select nodeId,alias,uniqueId
from cmsTemplate t
inner join umbracoNode n on n.Id = t.NodeId
```

### DIFF
--- a/src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
+++ b/src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
@@ -1558,7 +1558,15 @@ namespace Umbraco.Cms.Infrastructure.Packaging
                 var masterElement = templateElement.Element("Master");
 
                 var existingTemplate = _fileService.GetTemplate(alias) as Template;
+
                 var template = existingTemplate ?? new Template(_shortStringHelper, templateName, alias);
+
+                // For new templates, use the serialized key if avaialble.
+                if (existingTemplate == null && templateElement.Element("Key") != null)
+                {
+                    template.Key = Guid.Parse(templateElement.Element("Key").Value);
+                }
+
                 template.Content = design;
 
                 if (masterElement != null && string.IsNullOrEmpty((string)masterElement) == false)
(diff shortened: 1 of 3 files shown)


src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
```diff
@@ -1558,7 +1558,15 @@ public IReadOnlyList<ITemplate> ImportTemplates(IReadOnlyCollection<XElement> te
...
 
                 var existingTemplate = _fileService.GetTemplate(alias) as Template;
+
                 var template = existingTemplate ?? new Template(_shortStringHelper, templateName, alias);
+
+                // For new templates, use the serialized key if avaialble.
+                if (existingTemplate == null && templateElement.Element("Key") != null)
+                {
+                    template.Key = Guid.Parse(templateElement.Element("Key").Value);
+                }
```
ronaldbarendse, line 1568: I've used the following in `UpdateContentTypesPropertyGroups()`:

https://github.com/umbraco/Umbraco-CMS/blob/2c530cf59bcb6b08a5a2b939831c9492a5f030ed/src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs#L813-L816

I would suggest changing this to remove the duplicate `Element("Key")` calls and gracefully ignore invalid GUIDs:
```suggestion
                if (existingTemplate == null && Guid.TryParse(templateElement.Element("Key")?.Value, out var key)
                {
                    template.Key = key;
                }
```

ronaldbarendse (approved):
Looks good, just one minor suggestion! Have you also checked all the other entity types or was template the only one without keys?

src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
AndyButland: Sounds good, making this change.

AndyButland:
Templates was the only one I had issues with using the starter kit, and looking through `EntityXmlSerializer` I think anything that can have a key, has one serialized.

PR #11167, opened 2021-09-22
Add migration for isMaster column to isSchedulingPublisher in UmbracoServer table

Jeavon:
Following on from https://github.com/umbraco/Umbraco-CMS/pull/10480 In this PR I added a migration for isMaster column to isSchedulingPublisher in UmbracoServer table

Creating this as a draft PR for now as my local v9 is having problems so I've not yet verified it works


bergmania:
Could potentially fail with SqlCE.. But needs to be tested for both SqlServer and SqlCE

Jeavon:
Fixed my build (somehow I had a hooky version of Examine 2.0 that had different method signatures in my package cache! 🤯)

Tested with CE 
Before
image
After
image

And tested with LocalDB

All good.

I did find 2 isMaster references in SQL that I've updated in https://github.com/umbraco/Umbraco-CMS/pull/11167/commits/a3ca09c82e690859c132488eb68290237dfd0643 which have updated this PR

bergmania (approved):

PR #11171, opened 2021-09-22
add MaybeNull attribute to mixin properties

### LINKED ISSUE #11118: Inconsistent nullability (MayBeNull) when using compositions with modelsbuilder in v9

9.0.0-rc003

### Bug summary

When using Nullable Reference Types in v9 the `System.Diagnostics.CodeAnalysis.MaybeNull` attribute seems to be inconsistently applied. It exists on the composition but not the content model itself i.e.

image

### Specifics

Here's the generated `IPageHeader`
image

Here's the actual property on the doctype:
image

### Steps to reproduce

1. Configure ModelsBuilder to use SourceCodeAuto or SourceCodeManual
2. Create a Composition with a nullable property value.
2. Apply composition to another content type.
3. Generate ModelsBuilder models.

Properties on composed content types would have the same `MayBeNull` attribute as the compositions they implement.

JasonElkin:
### Prerequisites

Fixes #11118 

### Description

Adds the `System.Diagnostics.CodeAnalysis.MaybeNull` attribute on composition/mixin properties so nullability can be determined when using Nullable Reference Types in a project.

Before:
image

After:
image

To test:

1. Enable SourceCode/SourceCodeAuto mode for ModelsBuilder
2. Add a composition with a nullable property type to a doctype.
3. Generate models.
4. nullable property type should have the `MayBeNull` attribute applied.

### DIFF
--- a/src/Umbraco.Infrastructure/ModelsBuilder/Building/TextBuilder.cs
+++ b/src/Umbraco.Infrastructure/ModelsBuilder/Building/TextBuilder.cs
@@ -255,6 +255,11 @@ namespace Umbraco.Cms.Infrastructure.ModelsBuilder.Building
             }
 
             WriteGeneratedCodeAttribute(sb, "\t\t");
+
+            if (!property.ModelClrType.IsValueType)
+            {
+                WriteMaybeNullAttribute(sb, "\t\t", false);
+            }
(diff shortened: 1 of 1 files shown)


bergmania:
Thanks @JasonElkin

JasonElkin:
Special thanks @bergmania for getting this in under the wire for RC4! 🎉

PR #11172, opened 2021-09-22
Add folderkeys to package xml.

### LINKED ISSUE #11170: V9: Keys not serialized into package.xml for content type folders and tabs

Latest v9/dev

### Bug summary

Similar to the issue fixed with [this PR](https://github.com/umbraco/Umbraco-CMS/pull/11163), we really need Guid keys serialized for all Umbraco schema data that might be created from a package migration.

For most we have - but for those without it presents an issue for Umbraco Deploy in that the `.uda` files generated in a local environment when the package is installed will differ from the keys used in schema created by the package migration running in an upstream environment.

What's still missing are:

- Content type folders (serialized only with the name currently on a child document type - e.g. `<DocumentType Folders="Compositions">`.

### Steps to reproduce

- Install package containing the features noted in a local environment
- Push updated .csproj file to an upstream environment using Umbraco Deploy
- Attempt to transfer content

Should be able to transfer content but currently get schema mismatch errors.

bergmania:
Fixes #11170..

Adds folder keys to datatype folders and content folders..

### DIFF
--- a/src/Umbraco.Core/Services/IContentTypeServiceBase.cs
+++ b/src/Umbraco.Core/Services/IContentTypeServiceBase.cs
@@ -78,7 +78,7 @@ namespace Umbraco.Cms.Core.Services
         /// <returns></returns>
         bool HasContainerInPath(params int[] ids);
 
-        Attempt<OperationResult<OperationResultType, EntityContainer>> CreateContainer(int parentContainerId, string name, int userId = Constants.Security.SuperUserId);
+        Attempt<OperationResult<OperationResultType, EntityContainer>> CreateContainer(int parentContainerId, Guid key, string name, int userId = Constants.Security.SuperUserId);
         Attempt<OperationResult> SaveContainer(EntityContainer container, int userId = Constants.Security.SuperUserId);
         EntityContainer GetContainer(int containerId);
         EntityContainer GetContainer(Guid containerId);
--- a/src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
+++ b/src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
@@ -599,7 +599,7 @@ namespace Umbraco.Cms.Infrastructure.Packaging
         private EntityContainer CreateContentTypeChildFolder(string folderName, Guid folderKey, IUmbracoEntity current)
         {
             var children = _entityService.GetChildren(current.Id).ToArray();
-            var found = children.Any(x => x.Name.InvariantEquals(folderName));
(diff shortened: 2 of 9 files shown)


src/Umbraco.Infrastructure/Packaging/PackageDataInstallation.cs
```diff
@@ -546,13 +546,32 @@ public IReadOnlyList<T> ImportDocumentTypes<T>(IReadOnlyCollection<XElement> uns
...
+                    if (folderKeysAttribute != null)
+                    {
+                        folderKeys = folderKeysAttribute.Value.Split(Constants.CharArrays.ForwardSlash).Select(x=>Guid.Parse(x)).ToArray();
+                    }
                     var rootFolder = WebUtility.UrlDecode(folders[0]);
-                    //level 1 = root level folders, there can only be one with the same name
-                    var current = _contentTypeService.GetContainers(rootFolder, 1).FirstOrDefault();
+
+                    EntityContainer current;
+                    if(folderKeys.Length == folders.Length)
```
AndyButland, line 561: If I'm reading it right... looks like there could be a bug here if `folders.Length == 0`.  `folderKeys.Length == folders.Length` would be `true` and then we'd be looking for an element that doesn't exist `folderKeys[0]`.

src/Umbraco.Core/Services/IContentTypeServiceBase.cs
```diff
@@ -78,7 +78,7 @@ public interface IContentTypeBaseService<TItem> : IContentTypeBaseService, IServ
         /// <returns></returns>
         bool HasContainerInPath(params int[] ids);
 
-        Attempt<OperationResult<OperationResultType, EntityContainer>> CreateContainer(int parentContainerId, string name, int userId = Constants.Security.SuperUserId);
+        Attempt<OperationResult<OperationResultType, EntityContainer>> CreateContainer(int parentContainerId, Guid key, string name, int userId = Constants.Security.SuperUserId);
```
bielu, line 81: @bergmania please flag that pr as breaking, we are doing changes in to public interfaces :)

AndyButland (approved):
This looks good to me now.  I've tested exporting and importing with the following scenarios:

- Doc type in the root.
- Doc type in a folder.
- Doc type in a folder two levels deep.
- Doc type in a folder added in the current migration.
- Data type in an existing folder
- Data type created in a folder two levels deep

bielu:
@bergmania rc04 is already live, so I think you mean label rc05 or v9 final :)

PR #11174, opened 2021-09-23
Default parameter in ToSafeAlias / ToUrlSegment

JoseMarcenaro:
### Prerequisites

This fixes #11173 

### Description
Provides a default value for the `IShortStringHelper shortStringHelper` argument in the simplest overload of `ToUrlSegment()` and `ToSafeAlias()` 

### Steps to test:

```
string urlSegment = "my document weird name & adiós".ToUrlSegment();
```

### DIFF
--- a/src/Umbraco.Core/Extensions/StringExtensions.cs
+++ b/src/Umbraco.Core/Extensions/StringExtensions.cs
@@ -25,6 +25,7 @@ namespace Umbraco.Extensions
         private const char DefaultEscapedStringEscapeChar = '\\';
         private static readonly char[] ToCSharpHexDigitLower = "0123456789abcdef".ToCharArray();
         private static readonly char[] ToCSharpEscapeChars;
+        private static readonly IShortStringHelper _defaultShortStringHelper = new DefaultShortStringHelper(new DefaultShortStringHelperConfig());
 
         static StringExtensions()
         {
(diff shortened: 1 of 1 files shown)


nathanwoulfe (commented):
@JoseMarcenaro this makes sense, just want to check with @p-m-j that allocating the static helper is preferable to instantiating it when required in the new extensions?

p-m-j:
:man_shrugging: might be nicer to service locate the registerd instance, probably a good candidate for the crossover slack channel.

nathanwoulfe:
Hey @JoseMarcenaro, checked in with HQ and the preference here is to add the new extension in Umbraco.Web.Common/Extensions, and use service location to get the registered IShortStringHelper instance rather than instantiating a new object with the default config (as the registered instance may be different to the default).

[Here's a good example](https://github.com/umbraco/Umbraco-CMS/blob/v9/contrib/src/Umbraco.Web.Common/Extensions/FriendlyUrlHelperExtensions.cs) of how this would work.

Hope you don't mind refactoring your PR 😃

JoseMarcenaro:
Hi @nathanwoulfe  - thanks for the explanation, and of course you can dynamite my PR and build it the right way 😆

nathanwoulfe:
@JoseMarcenaro no intention of dynamiting anything on your behalf - if you want to refactor the existing PR, I'm fine with merging it 😄

emmagarland:
@JoseMarcenaro also, just as an FYI I have just updated the base to `contrib`.

I've noticed that `StringExtensions.cs` now has conflicts to be resolved. 

Emma 🙂

JoseMarcenaro:
Hi @emmagarland thanks for contacting me.
This is a very old issue and I'm really not sure about the best action to take - please disregard this PR

emmagarland:
@JoseMarcenaro thanks very much for the update and letting me know. 

Emma

shearer3000:
hi - what was the end result here? can we use ToUrlSegment() without needing to pass in IShortStringHelper? thanks

JoseMarcenaro:
@shearer3000 no you cannot - As far as I know, that option is not available

Issue #11183, opened 2021-09-23
V9 RC4: Content finder causing redirect to `/account/accessdenied`


mattbrailsford:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

Umbraco v9 RC4

### Bug summary

Accessing a URL in Umbraco that gets handled by a content finder, the content finder successfully triggers and calls `request.SetRedirect(url)` and returns `true`, but the site is getting redirected to `/account/accessdenied` in some kind of infinite loop until the error `The request filtering module is configured to deny a request where the query string is too long.` is thrown.

### Steps to reproduce

* Create a content finder that redirects to a parent node
* Register the content finder before `ContentFinderByUrl`
* Call a URL on the front end that gets handled by the content finder
* Response ends up in an infinite loop

### Expected result / actual result

I'd expect to be redirected accordingly to the returned URL.

PR #11193, opened 2021-09-25
Fixes #11189 - protected content not working

### LINKED ISSUE #11189: Member groups stopped working

v9-rc004

### Bug summary

When a node is restricted to a member group and the member is in that group, it redirects to the login page.

### Specifics

_No response_

### Steps to reproduce

Create a member
Assign member to a group
Restrict public access to a node to the group
Try to access the node with the member

The member should be able to access the node once logged in

---
_This item has been added to our backlog [AB#14115](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/14115)_

Shazwazza:
Fixes #11189 + #11183

The problem was with a previous fix to route hijacking for members. Which meant the protected page would continue to redirect. That is because the auth middleware wasn't executing yet since our transformer runs first. So to fix, we force the auth for members to run when it's a protected page.
This also just ensures that when a member signs in, that the current HttpContext.User is assigned.

For 11183, the problem was a change in returning null where I presumed if there was no published content we would assume it's a 404. I didn't take into account that a finder would just redirect. So the code changes now check exactly for this.

* Makes sure public access works as expected
* Make sure you can still hijack routes for all public access pages 
* Make sure that the scenario in 11183 is resolved
* All routing works as expected

### DIFF
--- a/src/Umbraco.Web.Website/Routing/PublicAccessRequestHandler.cs
+++ b/src/Umbraco.Web.Website/Routing/PublicAccessRequestHandler.cs
@@ -77,7 +77,7 @@ namespace Umbraco.Cms.Web.Website.Routing
                         // instead of relying directly on the user assigned to the http context, and then the auth middleware
                         // will run anyways and assign the user. Perhaps that is a little cleaner, but would require more code
                         // changes right now, and really it's not any different in the end result.
-                        httpContext.User = authResult.Principal;
+                        httpContext.SetPrincipalForRequest(authResult.Principal);
                     }
 
                     publicAccessStatus = await _publicAccessChecker.HasMemberAccessToContentAsync(publishedContent.Id);
--- a/src/Umbraco.Web.Website/Routing/PublicAccessRequestHandler.cs
+++ b/src/Umbraco.Web.Website/Routing/PublicAccessRequestHandler.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Identity;
 using Microsoft.Extensions.Logging;
 using Umbraco.Cms.Core;
 using Umbraco.Cms.Core.Models;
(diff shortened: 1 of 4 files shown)


p-m-j:
Can confirm this resolves both linked issues.

Do wonder if in the future it might be better to separate endpoint selection and auth checks in the future by reintroducing a middleware to check access after UseRouting and dropping the checks in UmbracoRouteValueTransformer presumably with a redirect to member login page as suggested by @AndyButland in https://github.com/umbraco/Umbraco-CMS/issues/10730#issuecomment-892521782

Shannon and I discussed my suggested alternate fix of moving UseAuthentication before UseRouting which is strongly recomended against by all the documentation on middleware pipelines but does seem to be working ok for some folks. 
See https://stackoverflow.com/questions/58475596/why-does-useauthentication-have-to-be-placed-after-userouting-and-not-before for all lots of interesting reading in various linked github issues and articles.

src/Umbraco.Web.Website/Routing/PublicAccessRequestHandler.cs
```diff
@@ -65,6 +67,19 @@ public async Task<UmbracoRouteValues> RewriteForPublishedContentAccessAsync(Http
...
+                    AuthenticateResult authResult = await httpContext.AuthenticateAsync(IdentityConstants.ApplicationScheme);
+                    if (authResult.Succeeded)
+                    {
+                        // set the user to the auth result. we need to do this here because this occurs
+                        // before the authentication middleware.
+                        // NOTE: It would be possible to just pass the authResult to the HasMemberAccessToContentAsync method
+                        // instead of relying directly on the user assigned to the http context, and then the auth middleware
+                        // will run anyways and assign the user. Perhaps that is a little cleaner, but would require more code
+                        // changes right now, and really it's not any different in the end result.
+                        httpContext.User = authResult.Principal;
```
bergmania, line 80: Just to align with the codebase.
```suggestion
                        httpContext.SetPrincipalForRequest(authResult.Principal);
```

bergmania (approved):

PR #11198, opened 2021-09-26
Added extension method to easily register Multiple Notification Handlers.

mclausen:
### Description
Let me ask you a question. How many tears have you felt, running down your chin, every-time you’d forgot to register your freshly  created notification? How many times do you think you’d started the cms, and didn’t add your notifications? If you are like me, that is more than once, and that is too many.

What if i told you that you’d never had to forget to registers any custom notification handlers ever again, and that the solution would be helpful and friendly at the same time?

Enter this PR.

What does it do? It is an immediate pain reliever, that comes in a form of an Extension Method to `IUmbracoBuilder` named `AddNotificationsFromAssembly<T>`  where T is the type within the assembly your or the customer custom notifications is located. It locates all your implementations of `INotificationHandler<TNotification>` and registers them in the `IServiceCollector` so you dont have to manually remember and add them your self.

### How does it look in the wild?

You can add all notifications on **One** line

```
public class MyFavoriteComposer : IComposer
{
    {
        builder.AddNotificationsFromAssembly<Startup>();
    }
}
```

Whatever custom notification that you have in your project will now be registered and the best thing is, that if you create a new one, then you no longer have to forget to register it.

Cheers 🍻


nul800sebastiaan:
Cheers indeed! But it would be really nice if you could rebase on `v9/dev` since.. I'm guessing you didn't actually change 5000+ files? 😅

mclausen:
Closing, I will try to create a new PR 🤦‍♂️

mclausen:
Recreated the PR #11199

PR #11200, opened 2021-09-27
v9: Implement IOptionsMonitor or IOptionsSnapshot instead of IOptions

Zeegaan:
# Notes
-Implemented IOptionsMonitor instead of IOptions for Unique and singleston services, this should allow changes in your config while running the application, so you can update your config on the fly!
- Implemented IOptionsSnapshot instead of IOptions for scoped services, as this will cache the value per request

# How to test
- Use https://our.umbraco.com/documentation/Reference/V9-Config/ for a list of config options
- Try to breakpoint the classes where I've made the change, it helps changing the OnChange to be like:
```
config.OnChange(x =>
                {
                    _myConfig = x;
                });
```
So you can see whenever the settings are changed

### DIFF
--- a/src/Umbraco.Web.Common/Filters/AngularJsonOnlyConfigurationAttribute.cs
+++ b/src/Umbraco.Web.Common/Filters/AngularJsonOnlyConfigurationAttribute.cs
@@ -24,11 +24,10 @@ namespace Umbraco.Cms.Web.Common.Filters
             private readonly ArrayPool<char> _arrayPool;
             private MvcOptions _options;
 
-            public AngularJsonOnlyConfigurationFilter(ArrayPool<char> arrayPool, IOptionsMonitor<MvcOptions> options)
+            public AngularJsonOnlyConfigurationFilter(ArrayPool<char> arrayPool, IOptionsSnapshot<MvcOptions> options)
             {
                 _arrayPool = arrayPool;
-                _options = options.CurrentValue;
-                options.OnChange(x => _options = x);
+                _options = options.Value;
             }
 
             public void OnResultExecuted(ResultExecutedContext context)
--- a/src/Umbraco.Web.Common/Filters/AngularJsonOnlyConfigurationAttribute.cs
+++ b/src/Umbraco.Web.Common/Filters/AngularJsonOnlyConfigurationAttribute.cs
@@ -22,12 +22,13 @@ namespace Umbraco.Cms.Web.Common.Filters
         private class AngularJsonOnlyConfigurationFilter : IResultFilter
         {
             private readonly ArrayPool<char> _arrayPool;
-            private readonly IOptions<MvcOptions> _options;
+            private MvcOptions _options;
 
-            public AngularJsonOnlyConfigurationFilter(ArrayPool<char> arrayPool, IOptions<MvcOptions> options)
+            public AngularJsonOnlyConfigurationFilter(ArrayPool<char> arrayPool, IOptionsMonitor<MvcOptions> options)
             {
                 _arrayPool = arrayPool;
-                _options = options;
+                _options = options.CurrentValue;
+                options.OnChange(x => _options = x);
             }
 
             public void OnResultExecuted(ResultExecutedContext context)
@@ -45,7 +46,7 @@ namespace Umbraco.Cms.Web.Common.Filters
                     };
 
                     objectResult.Formatters.Clear();
-                    objectResult.Formatters.Add(new AngularJsonMediaTypeFormatter(serializerSettings, _arrayPool, _options.Value));
+                    objectResult.Formatters.Add(new AngularJsonMediaTypeFormatter(serializerSettings, _arrayPool, _options));
                 }
             }
         }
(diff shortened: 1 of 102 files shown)


src/Umbraco.Web.Common/Filters/AngularJsonOnlyConfigurationAttribute.cs
```diff
@@ -22,12 +22,13 @@ public AngularJsonOnlyConfigurationAttribute() : base(typeof(AngularJsonOnlyConf
         private class AngularJsonOnlyConfigurationFilter : IResultFilter
         {
             private readonly ArrayPool<char> _arrayPool;
-            private readonly IOptions<MvcOptions> _options;
+            private MvcOptions _options;
 
-            public AngularJsonOnlyConfigurationFilter(ArrayPool<char> arrayPool, IOptions<MvcOptions> options)
+            public AngularJsonOnlyConfigurationFilter(ArrayPool<char> arrayPool, IOptionsMonitor<MvcOptions> options)
```
nikolajlauridsen, line 27: I'd guess that this is scoped too since it's a result filter?

src/Umbraco.Web.Common/Localization/UmbracoRequestLocalizationOptions.cs
```diff
@@ -11,18 +11,22 @@ namespace Umbraco.Cms.Web.Common.Localization
...
     public class UmbracoRequestLocalizationOptions : IConfigureOptions<RequestLocalizationOptions>
     {
-        private readonly IOptions<GlobalSettings> _globalSettings;
+        private GlobalSettings _globalSettings;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="UmbracoRequestLocalizationOptions"/> class.
         /// </summary>
-        public UmbracoRequestLocalizationOptions(IOptions<GlobalSettings> globalSettings) => _globalSettings = globalSettings;
+        public UmbracoRequestLocalizationOptions(IOptionsMonitor<GlobalSettings> globalSettings)
```
nikolajlauridsen, line 19: I think this is transient too, if you navigate to implementation of `ConfigureOptions` in `AddWebComponents` you can see that it gets added as transient

src/Umbraco.Web.Common/ModelsBuilder/InMemoryModelFactory.cs
```diff
@@ -68,7 +68,7 @@ internal class InMemoryModelFactory : IAutoPublishedModelFactory, IRegisteredObj
             _umbracoServices = umbracoServices;
             _profilingLogger = profilingLogger;
             _logger = logger;
-            _config = config.Value;
+            _config = config.CurrentValue;
```
nikolajlauridsen, line 71: `InMemoryModelFactory` is a singleton, so the config won't update unless we hook into the event

src/Umbraco.Web.UI/Views/Root.cshtml
```diff
@@ -0,0 +1,6 @@
+﻿@using Umbraco.Cms.Web.Common.PublishedModels;
+@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<ContentModels.Root>
+@using ContentModels = Umbraco.Cms.Web.Common.PublishedModels;
+@{
+	Layout = null;
+}
```
nikolajlauridsen, line 6: Accidentally committed file?

nikolajlauridsen (changes requested):
Most of the files looks good, however there were some places where I'm not sure if this implements the correct approach regarding lifetimes.

However these suggestions are not the end all be all, I'm not entirely sure about some of them myself, one question I have is regarding transient services. Should we only use `IOptionsSnapshot` if we know that the class will be used in the context of a request? For instance all notificationhandlers are transient, but they aren't always used where there exists an HTTP context.

src/Umbraco.Web.Common/ModelsBuilder/InMemoryModelFactory.cs
Zeegaan: We should actually still do that in the ModelsGenerationError.cs

nikolajlauridsen:
Just so we don't forget, we can't merge this into the V9 branch since it contains breaking changes 😄

src/Umbraco.Web.Common/ModelsBuilder/InMemoryModelFactory.cs
nikolajlauridsen: Yes, but that `ModelsBuilderSettings` is also used within `InMemoryModelFactory` for instance on line 138:

```C#
public bool Enabled => _config.ModelsMode == ModelsMode.InMemoryAuto;
```

nikolajlauridsen (approved):
Just need builder servers to be happy and then we can create a v10/dev branch and merge 🎉

KevinJump:
Hi, Just a little note - it would be good to know what the breaking change is - makes its easier for us to see if it effects us as there 96 files in this on.

nikolajlauridsen:
It's been a while since I review this PR, but as I recall all the breaking changes in this PR is constructor changes, where `IOptions` has been replaced with `IOptionsSnapshot` or `IOptionsMonitor`, so if you inherit from or instantiate any of these files you'll have to update the constructor :)

PR #11201, opened 2021-09-27
Added missing GetCropUrl overload for MediaWithCrops

lauraneto:
The missing overload would mean that if anyone tried to use the extra options with MediaWithCrops, they would be using the IPublishedContent extension method instead.
This would cause several issues, like not loading the local crops and returning null for a lot of scenarios that should work.

### Prerequisites

### Description
Without this change the following line of code wouldn't work properly:
```
MediaWithCrops image = ...;
string localCropAlias = ...;
image.GetCropUrl(cropAlias: localCropAlias, useCropDimensions: true, imageCropMode: ImageCropMode.Crop, quality: quality, urlMode: urlMode);
```
Kept the same structure as the already existing overloads for consistency sake.

### DIFF
--- a/src/Umbraco.Web.Common/Extensions/FriendlyImageCropperTemplateExtensions.cs
+++ b/src/Umbraco.Web.Common/Extensions/FriendlyImageCropperTemplateExtensions.cs
@@ -126,6 +126,39 @@ namespace Umbraco.Extensions
                 urlMode
             );
 
+        public static string GetCropUrl(
+            this MediaWithCrops mediaWithCrops,
+            int? width = null,
+            int? height = null,
+            string propertyAlias = Cms.Core.Constants.Conventions.Media.File,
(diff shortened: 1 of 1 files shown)


mikecp:
Hi @lauraneto ,

Thanks a lot for this PR! It's a nice catch 👍 
If I can make a suggestion, I think it would be nice if you could add the comments on top of the method like it is done for the other overloads, so that our documentation/intellisense remains homogeneous 😁

Also, before I can merge it, I would need you to target the `v9/contrib` branch, and not the `v9/dev` branch which is targeted for HQ developers. So, it would be great if you could make that adaptation!

Thanks again and sorry for this little delay.

mikecp:
Thanks for the update @lauraneto! This PR is ready to be merged now  😁👍

And I'll have a look at the other one right after!

PR #11212, opened 2021-09-28
V9 Move sln files to root

p-m-j:


p-m-j:
>The WebApplicationFactory constructor infers the app content root path by searching for a WebApplicationFactoryContentRootAttribute on the assembly containing the integration tests with a key equal to the TEntryPoint assembly System.Reflection.Assembly.FullName. In case an attribute with the correct key isn't found, WebApplicationFactory <b>falls back to searching for a solution file (.sln) and appends the TEntryPoint assembly name to the solution directory.</b> The app root directory (the content root path) is used to discover views and content files.

https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-5.0&viewFallbackFrom=aspnetcore-2.2#how-the-test-infrastructure-infers-the-app-content-root-path-1

p-m-j:
WRT WebApplicationFactory our usage is still pretty weird, we should be calling Web.UI.Startup.ConfigureServices then replacing things in ConfigureTestServices, but can refactor that out another day.

Issue #11215, opened 2021-09-28
Rebuild causes build errors in Visual Studio 2020 Preview 4 and  Visual Studio 2019


JohnBergman:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0

### Bug summary

Brand new project created using the visual studio 2020 File-New method.

close, reopen visual studio, then do a rebuild all with multiple projects of multiple types (in my case I have a database project, a couple of class libraries that are .net 5, and the umbraco project.

Right-clicking on the solution and choosing rebuild leaves a bunch of build errors, as seen in the imag
<img width="1207" alt="Rebuild Errors" src="https://user-images.githubusercontent.com/5505858/135164187-4e9c9175-152f-4bef-b2f6-33ca574eeb53.png">
es

If I right click on the project itself and select build or rebuild it is fine.  

### Steps to reproduce

See the bug summary for a way to see the repo steps.

### Expected result / actual result

Regardless of how I choose the build/rebuild, the results should be consistent.

OzoneNZ:
Can confirm this happens in Visual Studio 2019 as well as 2022 Preview 4

Issue seems to be the `BeforeTargets` value specified on the `CopyUmbracoAssets` target here, which is leading to the target being skipped when running a **Rebuild** rather than just a **Build**

https://github.com/umbraco/Umbraco-CMS/blob/release-9.0.0/build/NuSpecs/build/Umbraco.Cms.StaticAssets.targets

`BeforeTargets="Rebuild;Build"` rather than just `BeforeTargets="Build"` may fix this

Replicated on a fresh `9.0.0` project

p-m-j:
Closed by #11216

PR #11216, opened 2021-09-29
Copy static assets during BeforeBuild target to fix Visual Studio "rebuild" failures

OzoneNZ:
### Description

Fixes issue #11215 when using `Rebuild` option in Visual Studio/msbuild where every second rebuild would fail due to missing static asset files

PZPpEVXscy

msbuild output prior to this change:

```
1>Target _CopyOutOfDateSourceItemsToOutputDirectoryAlways:
1>  Microsoft.Common.CurrentVersion.targets(4996,5): error MSB3030: Could not copy the file "C:\Umbraco9Test\umbraco\config\appsettings-schema.json" because it was not found.
1>  Microsoft.Common.CurrentVersion.targets(4996,5): error MSB3030: Could not copy the file "C:\Umbraco9Test\umbraco\config\lang\cy.xml" because it was not found.
1>  Microsoft.Common.CurrentVersion.targets(4996,5): error MSB3030: Could not copy the file "C:\Umbraco9Test\umbraco\config\lang\en.xml" because it was not found.
1>  Microsoft.Common.CurrentVersion.targets(4996,5): error MSB3030: Could not copy the file "C:\Umbraco9Test\umbraco\config\lang\da.xml" because it was not found.

...

1>Done building target "_CopyOutOfDateSourceItemsToOutputDirectoryAlways" in project "Umbraco9Test.csproj" -- FAILED.
1>
1>Done building project "Umbraco9Test.csproj" -- FAILED.
1>
1>Build FAILED.
```

### DIFF
--- a/build/NuSpecs/build/Umbraco.Cms.StaticAssets.targets
+++ b/build/NuSpecs/build/Umbraco.Cms.StaticAssets.targets
@@ -6,9 +6,8 @@
         <UmbracoWwwrootName Condition="'$(UmbracoWwwrootName)' == ''">umbraco</UmbracoWwwrootName>
     </PropertyGroup>
 
-    <Target Name="CopyUmbracoAssets" BeforeTargets="Build">
+    <Target Name="CopyUmbracoAssets" BeforeTargets="BeforeBuild">
         <ItemGroup>
-
             <ContentFiles Include="$(ContentFilesPath)" />
             <ContentWwwrootFiles Include="$(ContentWwwrootFilesPath)" />
         </ItemGroup>


p-m-j:
Works like a charm #h5yr!

nathanwoulfe:
Am I right in assuming package devs should be making a similar change in .targets to avoid the same issue?

p-m-j:
@nathanwoulfe I guess that depends on if the packages need to write files into the project directories for editing by their users.

We're hoping to move away from these msbuild hacks to immutable content in nuget packages and package developers should try to do so where possible also.

- If content files in the nupkg are only required to be when publishing, a nuspec with contentFiles will be great.
- If it's just adding static files to wwwroot then a razor class library is probably better (basically does the same as contentFiles but with a handy file provider for debugging).
- For immutable razor views it makes sense to embed them in the class library (or potentially use RCL's for distributing partials)
- What we're doing here is last resort and requires cleanup.

bergmania:
Cherry picked to 9.0.1

PR #11217, opened 2021-09-29
Fix for Excessive header health check when excluding Cloudflare

Jeavon:
Bug fixed the check as TryGetValues returns a collection so need FirstOrDefault


nathanwoulfe (approved):
Cheers @Jeavon, looks fine to me. Failing tests won't be related to this change.

PR #11219, opened 2021-09-29
Changed the case of BlockList to blocklist as it breaks on Linux Mint

prjseal:
Changed the case of BlockList to blocklist as it breaks on Linux Mint. Thanks to @craigs100 for spotting this.

### DIFF
--- a/src/Umbraco.Web.UI/Views/Partials/blocklist/default.cshtml
+++ b/src/Umbraco.Web.UI/Views/Partials/blocklist/default.cshtml
@@ -8,6 +8,6 @@
         if (block?.ContentUdi == null) { continue; }
         var data = block.Content;
 
-        @await Html.PartialAsync("BlockList/Components/" + data.ContentType.Alias, block)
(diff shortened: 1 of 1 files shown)


glombek:
This looks good @prjseal - thanks for contributing!

I'm just wondering why these folders are lower-case to begin with while all the rest are Pascal case. Let me check if anybody knows why this is the case.

prjseal:
I don't know why. The grid folder is lowercase too. I think it would be best if all folders were lowercase personally, to be consistent and avoid errors on Linux based operating systems.

glombek:
Thanks @prjseal, I've checked with HQ and it's staying that way. Consistency is key for avoiding issues on *nix systems. All the .NET-y bits are capitalized (as is convention) while the front-endy bits are lower case (as is convention there too!) We just need to ensure all code references to these folders follow that convention too.

So your fix looks good to go. Thanks again!

glombek (approved):

PR #11222, opened 2021-09-30
Fix JsonConfigManipulator causing upgrades to fail when appsettings.json files are not present

OzoneNZ:
### Description

Database upgrades currently fail when no JSON configuration file is present (e.g. `appsettings.json`).

This is because `JsonConfigManipulator` always expects a JSON configuration file to be present, which isn't consistent with `Host.CreateDefaultBuilder` which considers the JSON files to be optional.

All of the public methods defined in `JsonConfigManipulator` and `IConfigManipulator` are susceptible to this crash.

Exceptions are also being encountered during fresh installs, but unlike upgrades, the install can still proceed:

image

## Replication

I encountered this issue when running Umbraco `9.0.0` under a Docker container and trying to configure the application purely with environment variables.

Presumably, using any configuration method that is **not** JSON files is affected by this (command line arguments, environment variables).

JSON configuration files are not desirable in Docker containers, as we want to configure it only by providing environment variables.

It should be replicable under Visual Studio by just removing `appsettings.json` files and using the `launchSettings.json` => `environmentVariables` section.

As a workaround, creating a dummy `appsettings.json` in my Docker containers with no contents works.

**Raw stack trace:**

```
System.IO.FileStream.ValidateFileHandle(SafeFileHandle fileHandle)
System.IO.FileStream.CreateFileOpenHandle(FileMode mode, FileShare share, FileOptions options)
System.IO.FileStream..ctor(string path, FileMode mode, FileAccess access, FileShare share, int bufferSize, FileOptions options)
System.IO.StreamReader.ValidateArgsAndOpenPath(string path, Encoding encoding, int bufferSize)
System.IO.StreamReader..ctor(string path)
Umbraco.Cms.Core.Configuration.JsonConfigManipulator.GetJson(JsonConfigurationProvider provider)
Umbraco.Cms.Core.Configuration.JsonConfigManipulator.SaveConfigValue(string key, object value)
Umbraco.Cms.Web.Common.RuntimeMinification.SmidgeRuntimeMinifier.Reset()
Umbraco.Cms.Web.BackOffice.Install.InstallController.Index()
```

image

## Changes

`FileNotFoundException` exceptions are now caught when attempting to open `appsettings.json`, and a logger warning is emitted.

New logger output looks like this:

image

### DIFF
--- a/src/Umbraco.Infrastructure/Configuration/JsonConfigManipulator.cs
+++ b/src/Umbraco.Infrastructure/Configuration/JsonConfigManipulator.cs
@@ -2,10 +2,12 @@ using System;
 using System.IO;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Configuration.Json;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using Umbraco.Cms.Web.Common.DependencyInjection;
 
 namespace Umbraco.Cms.Core.Configuration
 {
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Looking good to me now! Thanks for the help and the updates after reviews @OzoneNZ! 👍

Issue #11221, opened 2021-09-29
v9: SQL error when trying to create a new property on a tab that is inherited from a Composition (SQL CE and LocalDB)


BarryFogarty:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0

### Bug summary

A SQL error is generated when trying to create a new property on a tab that is inherited from a Composition (using SQL CE).  The error is: A duplicate value cannot be inserted into a unique index.  [ Table name = cmsPropertyTypeGroup,Constraint name = IX_cmsPropertyTypeGroupUniqueID ]

I believe this is caused by a rename of the tab in the Composition, and/or on the Document Type prior to the Save attempt.

### Specifics

Received an error from the server
```
An error occurred
A duplicate value cannot be inserted into a unique index. [ Table name = cmsPropertyTypeGroup,Constraint name = IX_cmsPropertyTypeGroupUniqueID ]

Exception Details
System.Data.SqlServerCe.SqlCeException, System.Data.SqlServerCe, Version=4.0.0.1, Culture=neutral, PublicKeyToken=89845dcd8080cc91: A duplicate value cannot be inserted into a unique index. [ Table name = cmsPropertyTypeGroup,Constraint name = IX_cmsPropertyTypeGroupUniqueID ]
Stacktrace
at System.Data.SqlServerCe.SqlCeCommand.ProcessResults(Int32 hr)
   at System.Data.SqlServerCe.SqlCeCommand.ExecuteCommandText(IntPtr& pCursor, Boolean& isBaseTableCursor)
   at System.Data.SqlServerCe.SqlCeCommand.ExecuteCommand(CommandBehavior behavior, String method, ResultSetOptions options)
   at System.Data.SqlServerCe.SqlCeCommand.ExecuteNonQuery()
   at Umbraco.Cms.Infrastructure.Persistence.FaultHandling.RetryPolicy.ExecuteAction[TResult](Func`1 func)
   at NPoco.Database.ExecuteNonQueryHelper(DbCommand cmd)
   at NPoco.DatabaseTypes.SqlServerCEDatabaseType.ExecuteInsert[T](Database db, DbCommand cmd, String primaryKeyName, Boolean useOutputClause, T poco, Object[] args)
   at NPoco.Database.InsertImp[T](PocoData pocoData, String tableName, String primaryKeyName, Boolean autoIncrement, T poco)
[...]

PR #11231, opened 2021-10-01
Remove inherited property group id/key when local properties are added

ronaldbarendse:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/11221.

### Description
While testing 8.17 RC1 we noticed all property group keys were re-generated on every save, because they weren't posted back to the server (which caused issues in Umbraco Deploy). This was fixed in PR https://github.com/umbraco/Umbraco-CMS/pull/11121 that was part of 8.17 RC2 (and the final 9.0.0 release).

One minor oversight/regression was that when you added properties to an inherited tab or group, we can't use the same id/key, as it needs to store that property into a new group that's local to the content type (and not the composition). This caused an exception while trying to save the new local group with the same uniqueId/key as the group from the composition.

This PR ensures that when a local property is added and the group isn't local (doesn't have the same content type ID as the one you're saving), it will omit these values, so the server can generate new, non-conflicting/unique keys and create the local group 😄 

@nul800sebastiaan We should ensure this is included in the final 8.17.0 release and the next 9.0.1 patch release!

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/contenttypehelper.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/contenttypehelper.service.js
@@ -439,6 +439,45 @@ function contentTypeHelper(contentTypeResource, dataTypeResource, $filter, $inje
 
             array.push(placeholder);
 
+        },
+
+        rebindSavedContentType: function (contentType, savedContentType) {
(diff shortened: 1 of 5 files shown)


madsrasmussen (approved):
Changes looks good to me and they work as expected. Thank you for moving all the rebind logic to the service :D

PR #11235, opened 2021-10-01
Fix casing issues for Linux users

failureflawless:
### Prerequisites

### Description

Run on Linux and try to create a blocklist editor. It should fail before this PR and work after applying these changes.


nul800sebastiaan:
☝️  that was me demoing

PR #11240, opened 2021-10-01
v8: Backoffice Welsh language translation updates

OwainJ:
### Description

Updated the Welsh language backoffice translation file to include new keys found in the **_en_** and **_en_us_** language files.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 # The MIT License (MIT) #
 
-Copyright (c) 2013-present Umbraco
+Copyright (c) 2005-present Umbraco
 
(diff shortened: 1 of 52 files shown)


OwainJ:
Just to double-check, but which of the language files is considered the "source of truth" for translation keys? As currently, I've been basing the **_cy.xml_** file off of both the **_en.xml_** and **_en_us.xml_** files, but, I've noticed that there are lang keys present in other lang files that are not present in _**en**_, and vice versa, but I'm not sure if they are just old unused keys?

Also, there are some keys present in **_cy.xml_** that are no longer present in _**en.xml**_, but I'm a little unsure on whether or not to remove them?

nathanwoulfe (approved):
Hey @OwainJ, to be honest, I'm not sure which files would be the most correct. I'd assume the English versions, but if there are discrepancies between those I'd err on the side of caution and include rather than exclude. I know there was some work done recently to clean up language files in v9, but that doesn't really help for v8. Difficult to clean up as keys can be generated in code too...

All said, this will be fine to merge - if you're adding keys that were missing, awesome, if any of those aren't in use there's no harm.

PR #11243, opened 2021-10-01
Change template helper to use async partials

### LINKED ISSUE #11220: V9: Inserting partial via template doesn't use await

9.0.0

### Bug summary

When inserting a partial view via template editor in backoffice it just use the following like in v8:

```
@Html.Partial("Breadcrumb")
```

It should instead use the recommended approach in .NET5:

```
@(await Html.PartialAsync("Breadcrumb"))
```

### Specifics

_No response_

### Steps to reproduce

Insert a partial via backoffice template editor and notice it render the partial like in v8, but not `async`.

_No response_

louisjrdev:
### Prerequisites

If there's an existing issue for this PR then this fixes #11220

### Description
Changes the templatehelper service so that when inserting a partial view in the template editor it uses the recommended approach of async partials.

### Testing
To test this create a new document type with a template and create a content page for it. Now create a partial with some example content. Now return to the document type template and use the editor to insert a partial view and pick the one you just created. You should then save this and view your content page and see that it renders the partial content correctly.

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/common/services/templatehelper.service.js
+++ b/src/Umbraco.Web.UI.Client/src/common/services/templatehelper.service.js
@@ -16,7 +16,7 @@
                 partialViewName = parentId + "/" + partialViewName;
             }
 
-            return "@Html.Partial(\"" + partialViewName + "\")";
+            return "@await Html.PartialAsync(\"" + partialViewName + "\")";
         }
 
         function getQuerySnippet(queryExpression) {
--- a/src/Umbraco.Web.UI.Client/test/unit/common/services/template-helper.spec.js
+++ b/src/Umbraco.Web.UI.Client/test/unit/common/services/template-helper.spec.js
@@ -26,28 +26,28 @@ describe('service: templateHelper', function () {
         it('should return the snippet for inserting a partial from the root', function () {
             var parentId = "";
             var nodeName = "Footer.cshtml";
-            var snippet = '@Html.Partial("Footer")';
+            var snippet = '@await Html.PartialAsync("Footer")';
             expect(templateHelper.getInsertPartialSnippet(parentId, nodeName)).toBe(snippet);
         });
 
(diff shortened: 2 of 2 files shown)


src/Umbraco.Web.UI.Client/src/common/services/templatehelper.service.js
```diff
@@ -16,7 +16,7 @@
                 partialViewName = parentId + "/" + partialViewName;
             }
 
-            return "@Html.Partial(\"" + partialViewName + "\")";
+            return "@await Html.PartialAsync(\"" + partialViewName + "\")";
```
bjarnef, line 19: It seems you are missing parenthesis to wrap this when using `await` keyword unlike when the static `Html` class is used.

louisjrdev: The parenthesis aren't required for it to work. I tested out and in the MS docs they use it without. I can add it though but I thought it just looked abit cleaner.

nathanwoulfe: If cleanliness is the goal, maybe interpolate the string, which removes the need to escape quotes? 

Related - a JS style guide would be awesome.

louisjrdev: Possibly but then it should probably be done for the rest of the file and that's a bit out of scope of this change.

Could always raise another issue for it.

bjarnef: Yes, it seems you are right. I recall I tried in previous, but not sure why it didn't worked to me at that time.
Anyway I justed tested it in a project on v9.0.0 and it does seem to work :)

bergmania (approved):
Thanks.. This is an oversight, and should definitely be async :)

bergmania:
Cherry picked for v9.0.1

PR #11244, opened 2021-10-01
Change references of Web.config to be appsettings.json

### LINKED ISSUE #11191: V9: Many (13) references to settings in web.config in en.xml (and translated equivalents)

v9/dev

### Bug summary

> Debug mode is turned on by setting <b>debug="true"</b> on the <b>&lt;compilation /&gt;</b> element in web.config.
> Database not found! Please check that the information in the "connection string" of the "web.config" file is correct.

These need updating and presumably removing for the time being in the translation files.

### Specifics

_No response_

### Steps to reproduce

Open src/Umbraco.Web.UI/umbraco/config/lang/en.xml and search for web.config

_No response_

louisjrdev:
If there's an existing issue for this PR then this fixes #11191 

### Description
Changes the references to the Web.config to now reference the appsetting.json in the en.xml language file.

### DIFF
--- a/src/Umbraco.Web.UI/umbraco/config/lang/en.xml
+++ b/src/Umbraco.Web.UI/umbraco/config/lang/en.xml
@@ -811,7 +811,7 @@
   </area>
   <area alias="installer">
     <key alias="databaseErrorCannotConnect">The installer cannot connect to the database.</key>
-    <key alias="databaseErrorWebConfig">Could not save the web.config file. Please modify the connection string manually.</key>
+    <key alias="databaseErrorWebConfig">Could not save the appsettings.json file. Please modify the connection string manually.</key>
     <key alias="databaseFound">Your database has been found and is identified as</key>
     <key alias="databaseHeader">Database configuration</key>
         <key alias="databaseInstall">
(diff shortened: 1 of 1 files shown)


louisjrdev:
I am not quite sure how the reference to the Web.config should be handled here: 
https://github.com/umbraco/Umbraco-CMS/pull/11244/files#R910

As that config setting does not exist in V9

mikecp:
Hello @louisjrdev ,
Thanks for this PR, it is indeed better to refer to the correct files and settings 😀
Would you mind porting those updates to the en_us.xml file as well? This way I can merge the two "default" translation files at once.
Thanks for your help! 👍

nul800sebastiaan:
Hello again @louisjrdev, we had a chat about some of these items and they could be improved some more, a lot of them refer to updating the `appsettings.json` file but there's MANY place to put config values, so instead of just saying "appsettings.json", we should refer to the config options and the docs for that.

Something like:

> your configuration source, for example, appsettings.json (see <a href="https://umbra.co/configdocs" target="_blank" rel="noopener">the documentation</a>)

You can add links if they are in a CDATA block, so for example:

```xml
<key alias="scheduledPublishDocumentation"><![CDATA[<a href="https://our.umbraco.com/documentation/Getting-Started/Data/Scheduled-Publishing/#timezones" target="_blank" rel="noopener">What does this mean?</a>]]></key>
```

Would you be up for going through them one more time in `en.xml` and `en_us.xml` to updated where it makes sense?

PR #11245, opened 2021-10-02
Cypress test for textbox max length

jemayn:
### Prerequisites

### Description

Added Cypress test for testing the textbox datatype's maxlength setting.
It roughly does this:

- Creates a doctype with a new textbox datatype with max chars 10
- Creates a content node based on that and add 8 chars to the textbox - asserts no helptext occurs
- Adds another char - assert it tells the user there is only 1 char left
- Adds 2 more - asserts it now errors as we are above the limit

Also renamed the member group test group from user group to member group as that confused me a bit 😁

Let me know if anything here isn't how it should be!

PS: Runs for around 4s on my machine - could probably shave a bit off if we skipped the initial save and publish assertion, already removed the middle one as it seemed unnecessary - the final one is necessary to trigger validation errors and is also fast as it errors before the actual save & publish action runs.

### DIFF
--- a/src/Umbraco.Tests.AcceptanceTest/cypress/integration/DataTypes/dataTypes.ts
+++ b/src/Umbraco.Tests.AcceptanceTest/cypress/integration/DataTypes/dataTypes.ts
@@ -99,7 +99,7 @@ context('DataTypes', () => {
         cy.umbracoButtonByLabelKey('buttons_saveAndPublish').click();
         cy.get('.property-error').should('be.visible');
 
-        // //Clean
+        // Clean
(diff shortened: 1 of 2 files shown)


nathanwoulfe (approved):
Looks good to me, thanks @jemayn. Test suite is running really slowly on my machine (at least 20 seconds per test), not sure what's going on there, but would like to see the suite run through before I merge.

nul800sebastiaan:
Cherry picked for v8 in 445f31ed090ddc80c7d898f0746612ed677c4675

PR #11246, opened 2021-10-02
Fixes wrong reference to enterSubmitFolder method in ng-keydown in mediapicker

skttl:
### Prerequisites

### Description
When creating a folder in the mediapicker, by clicking the + button, you have to unfocus the input field to make it create the folder. Previously you could press return / enter (13) on your keyboard to make it create the folder.

I found out the ng-keydown attribute was missing `vm.` in the reference to the method.

**To test**
- Open a media picker
- Click the + to start creating a new folder
- Type something and press return /enter
- The folder should be created


nathanwoulfe (approved):
Nicely done @skttl, works as advertised.

PR #11247, opened 2021-10-02
Align template picking more towards the other pickers

patrickdemooij9:
I noticed that the two template pickers were lacking a bit behind the other ones. I was able to find two template pickers:

- The one used when selecting one or more templates for your document type
- The one used when selecting a master template

And even those two were different from each other. The master template was the cleanest by using the editorService itempicker, but the document type template was still using a modal.

I've used the tree picker editorService to create an view that is more in line with other pickers that we currently have.

Master template:
masterTemplateChoosing

Document Type:
documentTypeTemplates

Some things that might need addressing:

- I now allow you to add a custom picker for the umbGridSelector. I am not sure if this is the correct way of doing it. Open to suggestions here!
- And I think my formatting isn't setup correctly in visual studio. I had a lot of issues with formatting not working that great. I don't know if that reflects in the PR.

Let me know what you think of those changes!

### DIFF
--- a/src/Umbraco.Web.BackOffice/Controllers/EntityController.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/EntityController.cs
@@ -1008,6 +1008,15 @@ namespace Umbraco.Cms.Web.BackOffice.Controllers
 
                 case UmbracoEntityTypes.Macro:
 
+                case UmbracoEntityTypes.Template:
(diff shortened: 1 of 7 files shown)


patrickdemooij9:
I see that my formatting has indeed come through with the PR :( Any way to get the correct formatting for .js files? Then I'll be able to fix the formatting

PR #11250, opened 2021-10-02
Ability to remove trees in the backoffice

patrickdemooij9:
During one of my developments on a website, I had to replace a tree within Umbraco and replace it with my own copy due to some extensions that weren't possible.
However, I learned that you are currently not able to remove a tree from the backoffice. As a workaround, I started to hide the tree with some CSS classes.

This PR includes some logic to be able to remove trees within a composer so that you are able to replace them with your own implementation.

Code you can use to test:
```
public class TestComposer : IComposer
    {
        {
            builder.Trees().RemoveTreeController<DictionaryTreeController>();
        }
    }
```

This will remove the tree controller responsible for the dictionary items. So after running this code, you can go to the translation section to check it. The translation section should be completely empty.

### DIFF
--- a/src/Umbraco.Tests.UnitTests/Umbraco.Web.BackOffice/CollectionBuilders/TreeCollectionBuilderTests.cs
+++ b/src/Umbraco.Tests.UnitTests/Umbraco.Web.BackOffice/CollectionBuilders/TreeCollectionBuilderTests.cs
@@ -1,0 +1,36 @@
+using System.Linq;
+using NUnit.Framework;
+using Umbraco.Cms.Core.Trees;
+using Umbraco.Cms.Web.BackOffice.Trees;
+
+namespace Umbraco.Cms.Tests.UnitTests.Umbraco.Web.BackOffice.CollectionBuilders
+{
+    public class TreeCollectionBuilderTests
+    {
+        [Test]
+        public void Adding_Tree_To_Collection_Builder()
+        {
+            var collectionBuilder = new TreeCollectionBuilder();
+            var treeDefinition = new Tree(0, "test", "test", "test", "test", TreeUse.Main, typeof(LanguageTreeController), false);
+
+            collectionBuilder.AddTree(treeDefinition);
+            var collection = collectionBuilder.CreateCollection(null);
+
+            Assert.AreEqual(1, collection.Count);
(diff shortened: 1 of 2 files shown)


patrickdemooij9:
Not quite sure why there is a failing test as I haven't touched that test at all. @nul800sebastiaan, could you perhaps rerun the tests?

bjarnef:
@patrickdemooij9 could be useful in some scenarios. I guess it would be possible to remove a specific tree for a specific user group or specific user?

I wonder if it could be used to remove a specific tree group or only specific trees.
For example maybe some clients want access to settings section. It could be useful if we could limit only to "Templating" and "Settings" group. Might be possible to remove "Settings" group or all trees in "Settings" tree group. However I am not sure how it would work with localization?

image

In could be useful if some clients wants to manage HTML, CSS and JS, but we want to restrict access to document types, etc.
Just a thought :)

patrickdemooij9:
@bjarnef , with this current PR it would remove it for everyone as it is executed in a composer.

We don't really have somethimg along the lines of IAppFactory and content apps for the trees. I could write it that way, but then that would probably be a breaking change.

What I could check out, but might be out of scope for this, is to check what happens if I clear the tree in the TreeRendering event (if that still happens in v9). There an user could add custom logic for who is able to view what tree.

bielu:
@patrickdemooij9 @bjarnef about "specific tree for a specific user group or specific user", I would do  more granular permission about sections, not overrides of trees :) it would resolve most of that type of issues. Umbraco really need redesign permission set, as now it is really hard to manage parts of sections

bjarnef:
Just testing this on a v9.1.2 project, but it doesn't seem to have a `RemoveTreeController()` method - only `AddTreeController()`.
What am I missing?

image

bjarnef:
Okay, it seems the class library did contain Umbraco.Cms.Web.BackOffice but not listed under Packages in VS. Installed Umbraco.Cms.Web.BackOffice updated this from 9.0.0 to 9.1.2 and the `RemoveTreeController()` now exits 👍

bjarnef:
Shouldn't this use `TController` and ensure the type inherits from `TreeControllerBase` like the `AddController<TController>()` method?

Issue #11251, opened 2021-10-02
Cannot easily go back from section with dashboards


patrickdemooij9:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0 (But I assume some latest versions of v8 as well)

### Bug summary

I assume that this bug was introduced with the change to dashboards so that you can link to them. http://localhost:9000/umbraco#/settings?dashboard=settingsPublishedStatus for example would go to the settings section and automatically open the settingsPublishedStatus dashboard.

However, it also defaults itself. So if I were to go to /umbraco/settings, it would automatically redirect me to /umbraco/settings?dashboard=settingsWelcome. This is annoying because if I click on the back button now, it'll go back to /umbraco/settings and redirect me again.

### Steps to reproduce

- Open a Umbraco website on the content section.
- Click on the settings section
- Verify that you have the dashboard= querystring in your url
- Try to click on the back button

### Expected result / actual result

I expect the back button to actually move me to the latest page (in the example above that would be the content section).

This can either be fixed by pushing the URL change without adding it to the history. Or by not adding the dashboard query string when there is none present and just use the default dashboard available.

PR #11252, opened 2021-10-02
11251: Don't add default dashboard to url

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/11251

The fix is to not load the dashboard to the URL when no dashboard URL is found (when you navigate towards it from the navigation).

Old situation:
oldSituation

New situation:
newSituation


nul800sebastiaan:
Should also be fixed for v8! We would love it if you could try to retarget this one for v9 @patrickdemooij9 - all backoffice "frontend" changes will still be merged up to v9, and it will still count for Hacktoberfest as it will end up in v9 eventually. 👍

patrickdemooij9:
@nul800sebastiaan , I've created a v8 version here: https://github.com/umbraco/Umbraco-CMS/pull/11267. I'll close this on then.

Issue #11048, opened 2021-09-07
Existing Property Disappears


prjseal:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

v9 RC 2

### Bug summary

v9 RC 2 with default starter kit installed.

I went to add a new property to the home page, but it had the same alias as an existing property (heroHeader)

When I pressed save it said it can't save because a property already exists with that alias.

I then changed the alias of my new property and pressed save.

This sort of thing happens a lot and doesn't normally cause an issue.

I noticed that the original property has disappeared the page doesn't load on the front end any more, complaining about the missing property Hero Header

### Specifics

Please see the bug summary

### Steps to reproduce

Please see the bug summary

### Expected result / actual result

I expected to get the warning message that i can't save a property with the same alias and nothing bad would happen. I would then rename my new alias and hit save and all be set up correctly.

The original Hero Header property was lost.

Zeegaan:
Hello there! Thank you for the bug report 🐛 

I can't seem to replicate the issue sadly, below I have posted a GIF of me trying to replicate it, can you tell me if I'm doing something wrong ? 🕵️ 

![Animation](https://user-images.githubusercontent.com/70372949/132482733-ab2ef809-3406-4cf2-8b1b-c3b7851ae7a5.gif)

Thanks again for the report! 🐛

prjseal:
@Zeegaan I was able to replicate it. Try doing it in the same group as the hero header property. Please see the attached gif of me replicating it.

![RC2Bug](https://user-images.githubusercontent.com/9142936/132510063-56924a51-69b1-4596-9c6c-eb397480b8c2.gif)

prjseal:
@Zeegaan I just tested it in 8.16 and it happens there too 😢

prjseal:
@Zeegaan I have just tested it in 8.8.4 and it happens there too.

prjseal:
@Zeegaan as a final test it also happens in 8.0.3 too.

prjseal:
@Zeegaan Okay Okay I hear you asking for me to test in v7 as well. Sadly it also happens in 7.15.7

I wasn't expecting that at all.

Zeegaan:
Hi @prjseal, thank you for the thorough investigation, it helped a lot 👏 
I have been doing some digging, and I think I have found the issue! 🕵️ 

It has something to do with how the frontend determines the ID of a given property.
[...]

PR #11257, opened 2021-10-03
11048: Bugfix for groups and properties that get replaced

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/11048

Steps for reproducing the bug can be found in the issue.

The issue was that whenever the save API call would return an error (due to the duplicate aliases here), it would still rebind the properties & groups based on their alias. So if you have a duplicate alias, they would both get the same id which eventually makes it so that the last property will override the first one. 
I am not really sure why the code is rebinding on an error as there isn't anything saved at that point. However, I have not encountered any issues with not rebinding the properties & groups on an error. If there is a scenario that I haven't tested yet where this is important, then please let me know and I'll take a look at it.
ValidProperties

### DIFF
--- a/src/Umbraco.Web.UI.Client/src/views/documentTypes/edit.controller.js
+++ b/src/Umbraco.Web.UI.Client/src/views/documentTypes/edit.controller.js
@@ -328,6 +328,10 @@
                     rebindCallback: function (origContentType, savedContentType) {
                         vm.contentType.ModelState = savedContentType.ModelState;
                         vm.contentType.id = savedContentType.id;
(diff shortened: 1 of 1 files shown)


patrickdemooij9:
Thanks @nul800sebastiaan , let me know if the other fix fixes this issue as well. Otherwise I will fix the merge conflict here.

nul800sebastiaan:
Alright, I tried a lot of whacky things and this seems to help indeed! 👍

nul800sebastiaan:
Cherry picked for v8.18 here: d02440d4111b59e560cfe49eae7be2bf0366bf7b

Issue #11227, opened 2021-09-30
v9 - UmbracoHelper GetDictionaryValue not working when Restrict Public Access is setup


konius:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0

### Bug summary

When a page has Restrict Public Access and user gets redirected to "Login Page" the dictionary lookup doesn't work.

Going to the "Login Page" manually works as expected so something happens in the rewrite middleware.

Code I am using (there is no route hijacking - there was but simplified this to test) :

```
@inherits UmbracoViewPage<LoginPage>
@{
    Layout = null;
}

<h1>@Model.Name</h1>
<p>Dictionary: @Umbraco.GetDictionaryValue("UI.Account.DontHaveAnAccountYet")</p>
```

![Screenshot 2021-09-30 114343](https://user-images.githubusercontent.com/3933562/135419694-a6aa0f1f-f8ce-4334-b1c4-0aa68552f027.jpg)
![Screenshot 2021-09-30 114423](https://user-images.githubusercontent.com/3933562/135419701-9f3df7cc-7538-4303-b711-6e1b20ae814b.jpg)
![Screenshot 2021-09-30 114441](https://user-images.githubusercontent.com/3933562/135419703-2379e180-76e2-4ecb-a48f-5c8c4f41a644.jpg)
![Screenshot 2021-09-30 114512](https://user-images.githubusercontent.com/3933562/135419705-fa160ebb-150b-4598-ac66-42bf391fb5b0.jpg)

### Steps to reproduce

1. Setup a protected page
2. Set Login Page
3. Navigate to protected page
4. Dictionary value is missing

### Expected result / actual result

I'd expect the dictionary value to show.

bergmania:
Hi @konius..

Thank you for the report.. I can repro, and will provide a fix for 9.0.1

bergmania:
PR: https://github.com/umbraco/Umbraco-CMS/pull/11262

nikolajlauridsen:
Fixed in #11262

PR #11262, opened 2021-10-04
Fixed issue with culture not set when doing internal redirects due to public access

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/11227

When rewriting the published content due to public access. E.g. redirect to login page, the culture was not specified. Thereby the dictionary item culture falled back to the default from config. If is not the correct one, no translations was shown.


nikolajlauridsen (approved):
Looks good, and can confirm that this fixes the issue 💪

Issue #11258, opened 2021-10-03
Error when accessing Membes section without Settings section: membertype/GetAllTypes


LennardF1989:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0

### Bug summary

If you login with a User that has no access to the Settings section, but does have access to the Members section, you will get the following error:

```
Authorization error: Unauthorized access to URL:
/umbraco/backoffice/umbracoapi/membertype/GetAllTypes
Contact your administrator for information.
```

### Steps to reproduce

1. Create a usergroup with access to Members, but not to Settings.
2. Create a user with the created usergroup
3. Login as the user
4. Go to the Members section, the error will show immediately.

### Expected result / actual result

No error is shown and/or the API should not be authorized for the Settings section permission only.

---
_This item has been added to our backlog [AB#14271](https://dev.azure.com/umbraco/243e7927-03b2-44e2-908f-d4ac7ea5daaa/_workitems/edit/14271)_

nul800sebastiaan:
Interesting, I tried this in v8, but it seems to be only a v9 problem.

I can reproduce in v9: 


![image](https://user-images.githubusercontent.com/304656/135847413-9ebbc4a7-8835-436f-b37b-68f421a80c4e.png)

bergmania:
PR: https://github.com/umbraco/Umbraco-CMS/pull/11264

PR #11264, opened 2021-10-04
Moved endpoint to new controller to avoid issue with too hard access requirements

bergmania:
Fixes https://github.com/umbraco/Umbraco-CMS/issues/11258

The issue is the auth policies cannot be overridden.. You need all of them, and the old controller requires you to have access to member types

### DIFF
--- a/src/Umbraco.Web.BackOffice/Controllers/MemberTypeQueryController.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/MemberTypeQueryController.cs
@@ -1,0 +1,44 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Authorization;
+using Umbraco.Cms.Core.Mapping;
+using Umbraco.Cms.Core.Models;
+using Umbraco.Cms.Core.Models.ContentEditing;
+using Umbraco.Cms.Core.Services;
+using Umbraco.Cms.Web.Common.Attributes;
+using Umbraco.Cms.Web.Common.Authorization;
+using Constants = Umbraco.Cms.Core.Constants;
+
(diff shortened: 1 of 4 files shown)


src/Umbraco.Web.BackOffice/Controllers/MemberTypeQueryController.cs
```diff
@@ -0,0 +1,44 @@
...
+            _umbracoMapper = umbracoMapper ?? throw new ArgumentNullException(nameof(umbracoMapper));
+        }
+
+        /// <summary>
+        /// Returns all member types
+        /// </summary>
+        public IEnumerable<ContentTypeBasic> GetAllTypes()
+        {
+            return _memberTypeService.GetAll()
+                               .Select(_umbracoMapper.Map<IMemberType, ContentTypeBasic>);
```
Zeegaan, line 40:
```suggestion
            return _memberTypeService.GetAll()
                .Select(_umbracoMapper.Map<IMemberType, ContentTypeBasic>);
```
More readable when its not indented so far

Zeegaan:
Found 1 minor thing, else everythings looks good to me 💪 

Tested this and works as intended!

src/Umbraco.Web.BackOffice/Controllers/MemberTypeQueryController.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
```
Zeegaan, line 41:
```suggestion
        public IEnumerable<ContentTypeBasic> GetAllTypes() =>
            _memberTypeService.GetAll()
                .Select(_umbracoMapper.Map<IMemberType, ContentTypeBasic>);
```
Stylecop suggestion 👍

PR #11265, opened 2021-10-04
Ensure dependency order for Deploy artifacts

AndyButland:
This issue raised this this should resolve is [here](https://github.com/umbraco/Umbraco.Deploy.Issues/issues/72), where it seemed to matter whether or not a template had a master template defined or not as to whether you could successfully deploy content.

After a bit of digging I found that Deploy was reporting a schema mismatch between two entities that were identical, due to differences in how the dependencies were ordered within the serialized output.  This wasn't reproducible between two local environments, only between local and Cloud.

Initially I thought we weren't ensuring the order and were relying on the collection to come back in a particular way, but actually we are doing this.  However the difference in behaviour looks to be due to differences in globalization settings (as we've seen in other contexts, related to [this](https://jimmybogard.com/mind-your-strings-with-net-5-0/)).

I've been able to demonstrate that the ordering of the string representations of these two UDIs comes back one way locally, and the opposite way in Cloud:

```
  var dependencies = new List<string>
  {
      "umb://template/d4651496fad24c1290a53ea4d55d945b",
      "umb://template-file/TestPage.cshtml"
  };
  foreach(var dependency in dependencies.OrderBy(x => x, StringComparer.InvariantCultureIgnoreCase))
  {
      <div>@dependency</div>
  }
```

Using `OrdinalIgnoreCase` makes them consistent though in both environments.  Hence I've applied the same setting for the ordering we do in core.

In the end think we've just been quite unlucky in having two entities that happen to be ordered by `-` v `/`.

### DIFF
--- a/src/Umbraco.Core/Udi.cs
+++ b/src/Umbraco.Core/Udi.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.ComponentModel;
 using System.Linq;
 
@@ -44,7 +44,7 @@ namespace Umbraco.Cms.Core
 
         public int CompareTo(Udi other)
         {
-            return string.Compare(UriValue.ToString(), other.UriValue.ToString(), StringComparison.InvariantCultureIgnoreCase);
+            return string.Compare(UriValue.ToString(), other.UriValue.ToString(), StringComparison.OrdinalIgnoreCase);
         }
 
         public override string ToString()
--- a/src/Umbraco.Tests.UnitTests/Umbraco.Core/Deploy/ArtifactBaseTests.cs
+++ b/src/Umbraco.Tests.UnitTests/Umbraco.Core/Deploy/ArtifactBaseTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Newtonsoft.Json;
 using NUnit.Framework;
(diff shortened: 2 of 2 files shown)


AndyButland:
I've tested this further by releasing the build from this PR as a nightly, updating my Cloud project to reference that, and can now transfer content for documents depending on a template that both has and doesn't have a master template.

bergmania:
Thanks..   That issue is ugly :P 

I'll merge this, but I think the right thing is to ensure the environments run using same globalization. 
E.g. by using this 
```xml
  <ItemGroup Condition="'$(OS)' == 'Windows_NT'">
    <PackageReference Include="Microsoft.ICU.ICU4C.Runtime" Version="68.2.0.6" />
    <RuntimeHostConfigurationOption Include="System.Globalization.AppLocalIcu" Value="68.2" />
  </ItemGroup>
```
in the csproj

bergmania:
I also cherry picked for 9.0.1

bjarnef:
@bergmania I added your suggestion https://github.com/umbraco/Umbraco-CMS/pull/11265#issuecomment-933657039 to the UmbracoProject.csproj file and deployed to Cloud environment, but it didn't seem to make a difference.

I still get same mismatch errors, when transfer content from local to remote environment.

Maybe it also need the changes in this PR as well?

Issue #11263, opened 2021-10-04
Smtp configuration is ignoring the From appsetting


enashnash:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0

### Bug summary

Smtp configuration is ignoring the From appsetting.

### Specifics

Inviting a user does not take account of the From setting in appsettings.

I'm getting the below error with

```
"Smtp": {
"From": "username@domain.com",
"Host": "smtp.office365.com",
"Port": 587,
"Username": "username@domain.com",
"Password": "********"
}
```

Failed to invite user
5.2.252 SendAsDenied; username@domain.com not allowed to send as loggedinuser@domain.com; STOREDRV.Submission.Exception:SendAsDeniedException.MapiExceptionSendAsDenied; Failed to process message due to a permanent exception with message Cannot submit message.

Note the error is stating that the email is sending as loggedinuser@domain.com (the logged in user) instead of username@domain.com.

This works fine with Google Workspace accounts but not for Office 365 which doesn't allow this.

Also note that the equivalent configuration for v8 works as expected.

### Steps to reproduce

Set up the appsettings file to send an email through Office 365.

Invite a new user from inside Umbraco.

The invitation will fail and show an error dialog in Umbraco with the above error message.

### Expected result / actual result

The email should be sent successfully.

bergmania:
Hi @enashnash.. Thanks for reporting.. I found the reason and will revert to same behavior as in Umbraco 8.

Zeegaan:
This issue is now fixed in https://github.com/umbraco/Umbraco-CMS/pull/11266, thanks again for reporting the issue 🐛

PR #11266, opened 2021-10-04
Use email from SMTP settings instead of current users email when sending invite emails

bergmania:
https://github.com/umbraco/Umbraco-CMS/issues/11263

Fixes issue where we used the current users email when sending invite emails. Now we use the one from SMTP settings like in v8


Zeegaan:
Tested this and it works! 

Code looks good to me 💪

PR #11270, opened 2021-10-04
Add test to update user

jemayn:
### Prerequisites

### Description

Adds a test that creates a user then updates their name and asserts that the update succeeds.

Question: Should we assert on the content of the xhr call? Or do we just assume it's all good with the success notification in Umbraco? 🙂

Also may make sense to extract the create user stuff in this new test into a cypress command, but had trouble figuring out how to test it works in the commands repo 🤷‍♂️

### DIFF
--- a/src/Umbraco.Tests.AcceptanceTest/cypress/integration/Users/users.ts
+++ b/src/Umbraco.Tests.AcceptanceTest/cypress/integration/Users/users.ts
@@ -32,4 +32,57 @@ context('Users', () => {
 
   });
 
+  it('Update user', () => {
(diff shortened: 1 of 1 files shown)


nathanwoulfe:
Hey @jemayn, these are great, keep em coming! Re the question on asserting the request content, I could go either way - we're testing the action, which Umbraco tells us has completed successfully by way of the notification, but for extra depth asserting the update exists in the response wouldn't hurt either... Might be a decision from HQ to be applied across the entire test suite?

For now though, this is good to go - new test passes 👍

Issue #11269, opened 2021-10-04
TemplateId returns 0 instead of NULL


patrickdemooij9:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

v9

### Bug summary

I was playing around with some content finders for nodes without a template when I noticed that the TemplateId is always 0 when there is no template. TemplateId is a nullable int though, so should just give NULL instead (unless there is a reason to return 0).

### Steps to reproduce

- Create a page with a template attached.
- Create a page without a template.
- Create the page without template as a child of the one with template
- Publish both pages
- Add the following snippet to the page with template:
```
<h1>TemplateId: @Model.TemplateId</h1>
@foreach (var child in Model.Children)
{
    <h2>Template @child.Name @child.TemplateId</h2>
}
```

Above code will return this:
![image](https://user-images.githubusercontent.com/11466511/135906761-3c24cf5f-3453-4f4e-8b99-4e81e96a52bd.png)

While it shouldn't show 0 for the second one.

### Expected result / actual result

TemplateId should return NULL when there is not a template attached.

umbrabot:
Hiya @patrickdemooij9,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #11271, opened 2021-10-04
11269: Make sure TemplateId is set correctly from cache

patrickdemooij9:
If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/11269

Way to reproduce bug can be found in the issue.
The issue here was that the id was saved as 0 in the Nucache instead of NULL. I am not sure if the way that I did this is the best way to fix it. I tried setting the template ids to int? within ContentSourceDto, but that then gave issues with the serialization. Ideally, I would want a serializer for nullable ints, but I couldn't find a lot of documentation about the serialization there (Or perhaps I was looking at the wrong place).

This way, it will still store the template ids as 0 in the Nucache, but it'll translate that back to NULL when loading the data.

Would love to implement any suggestions on how to fix this in a more efficient way.

### DIFF
--- a/src/Umbraco.Infrastructure/Persistence/Repositories/Implement/DocumentRepository.cs
+++ b/src/Umbraco.Infrastructure/Persistence/Repositories/Implement/DocumentRepository.cs
@@ -1232,12 +1232,12 @@ namespace Umbraco.Cms.Infrastructure.Persistence.Repositories.Implement
                 {
                     // need templates
                     var templateId = dto.DocumentVersionDto.TemplateId;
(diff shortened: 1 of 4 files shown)


nul800sebastiaan:
I'm afraid this would be a breaking change, we kept this behavior the same as how it worked in v8 so people knew what to expect. Maybe we can break in v10, will need to check with the team first.

bergmania:
This PR can be pointed to `v10/dev`, then it can be merged.

PR #11276, opened 2021-10-05
Force ICU usage on windows.

bergmania:
This PR forces new projects to use ICU.

By default Linux+Mac+Windows 10 (2019 may update) did use ICU, but Windows server 2016+2019 and Windows 10 (before 2019 may update) is using NLS.


nikolajlauridsen (approved):
LGTM 💪

Issue #11094, opened 2021-09-14
MediaSaving event messages always assumed as errors


matthewcare:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

8.16+

### Bug summary

If you have a MediaSaving event / notification handler, and want to add an event message, it is always interpreted as an error when uploading new media items regardless of if you set `EventMessageType` to `Success`, or any other non-error option.
This is an issue when uploading new media, as you then have to reload in order to view what you have just uploaded.

The issue appears to stem from `umbfiledropzone.directive.js` whereby when a file is uploaded this is checked
```
.success(function(data, status, headers, config) {
    if (data.notifications && data.notifications.length > 0) {
        // set error status on file
        file.uploadStatus = "error";
        // Throw message back to user with the cause of the error
        file.serverErrorMessage = data.notifications[0].message;
        // Put the file in the rejected pool
        scope.rejected.push(file);
```

The status should only be set to error if the `EventMessageType` is `Error`.
It also appears that you cannot have multiple messages here either, as the first item in the array is always chosen for the display message

### Steps to reproduce

Create a MediaSaving event / notification handler, and add a message - Umbraco 9 example:

```
public class IssueExampleComponent : INotificationHandler<MediaSavingNotification>
{
	public void Handle(MediaSavingNotification notification)
	{
		foreach (var entity in notification.SavedEntities)
		{
			notification.Messages.Add(new EventMessage("Event message",
			"An event message to show",
			EventMessageType.Info));
		}
	}
}
```

Upload a **new media item**, and the media will appear to have errored, and will show a red error message. You will then have to refresh to see the media you've just uploaded.
Note: This is for **new media items**, as saving an existing media item will show a notification as expected.

### Expected result / actual result

Expected result - A notification added during a media saving event / notification, should not be treated as an error unless `EventMessageType.Error` is used.
Actual result - All notifications are treated as errors, even if `EventMessageType.Success` is used.

matthewcare:
Whilst writing a fix for this for Umbraco 9, I found that the `PostAddFile` endpoint isn't correctly returning response data.
Issue: https://github.com/umbraco/Umbraco-CMS/issues/11278
Which I have opened a PR to fix
https://github.com/umbraco/Umbraco-CMS/pull/11279

matthewcare:
I've made an attempt to fix this, but I think it needs a bit of discussion, as current functionality prevents notifications from working in a nice way at all.

[...]

PR #11279, opened 2021-10-05
Return tempFiles from PostAddFile result

### LINKED ISSUE #11278: Response data is missing when adding new media

9.0

### Bug summary

When adding new media there is no response data. This is due to `tempFiles` not being passed into the return result.

This missing data means that the dropzone post upload isn't working properly if there is an error in the upload.

### Specifics

_No response_

### Steps to reproduce

Add new media, and check the response from the `PostAddFile`.

_No response_

matthewcare:
Add the response data to the return. Fixes #11278

### Prerequisites

- [ ✔] I have added steps to test this contribution in the description below

Fixes https://github.com/umbraco/Umbraco-CMS/issues/11278

### Description
Response data was not being included in the reutrn of `PostAddFile`, which caused notifications to not show correctly.

### DIFF
--- a/src/Umbraco.Web.BackOffice/Controllers/MediaController.cs
+++ b/src/Umbraco.Web.BackOffice/Controllers/MediaController.cs
@@ -858,7 +858,7 @@ namespace Umbraco.Cms.Web.BackOffice.Controllers
                 }
             }
 
-            return Ok();
+            return Ok(tempFiles);
         }
 
         private IMedia FindInChildren(int mediaId, string nameToFind, string contentTypeAlias)


nul800sebastiaan:
Thanks @matthewcare! I am trying to figure out what this is actually fixing though? You mentioned notifications not working, but they seem fine? I believe you are noticing error notifications not coming through? How do you make it error so that we can observe that the behavior is now correctly fixed?

Thanks in advance!

Note: this behavior doesn't seem changed in v8, so we might need to backport this to v8 as well.

Ps. you are not required to make an issue before you create a PR. If there something to discuss before making the PR then an issue is good, but if you have code already then only a PR is fine as well. 👍

matthewcare:
Hi @nul800sebastiaan 

In Umbraco 8, you'll notice that the `MediaController` action for this has a return result of
`Request.CreateResponse(HttpStatusCode.OK, tempFiles);`

Whereas in Umbraco 9 it has been ported to be `Ok()` which misses the `tempFiles` data.

I am currently fixing issue https://github.com/umbraco/Umbraco-CMS/issues/11094, and found that the dropzone response data is an empty string as the `tempFiles` data is not returned.

So in the method
```
.success(function(data, status, headers, config) {
    if (data.notifications && data.notifications.length > 0) {
        // set error status on file
        file.uploadStatus = "error";
        // Throw message back to user with the cause of the error
        file.serverErrorMessage = data.notifications[0].message;
        // Put the file in the rejected pool
        scope.rejected.push(file);
```

all uploaded files are now accepted, even if there is an error.

Thanks for the info about not needing to create an issue. I'll bear in mind for the future

matthewcare:
Sorry, I wasn't super clear on what the PR fixes.  When you upload files via the dropzone, and have a media saving notificaiton handler which adds messages, the js in `dropzone.directives.js` is meant to (although currently incorrectly), display that message as an error.

This PR adds the data back to the response, so that these messages once again show.

image

PR #11284, opened 2021-10-05
v8: Fix childless tab persistence

### LINKED ISSUE #11069: Tab not saved when no content fields added

v9 RC003

### Bug summary

If you add a tab but don't add any fields or groups it before saving, your tab isn't saved.

The main issue is it tells you all is saved but it is gone when you return.

### Specifics

Settings -> Document types
Create new or edit doctype

### Steps to reproduce

Add tab with no fields in:
image
Save (it looks ok):
image
New tab is gone when we return:
image

I am not sure what is best experience, perhaps not saving the tab is fair enough response to having no fields, but we should atleast have better UX on letting user setting doctype up know it isn't really saved?

lewishazell:
### Prerequisites

If there's an existing issue for this PR then this fixes #11069

### Description

A check which removes empty groups was also removing empty tabs. which is not the expected behaviour as per the issue above.

I have added a condition where only groups should undergo this check; that is, tabs should not.

#### Tests
As illustrated below,
- Create a new tab with no children
- Save the content type
- Switch to another page and return to the content type
- The tab should persist with no children
Da6bEW7Wy3

### DIFF
--- a/src/Umbraco.Web/Models/Mapping/ContentTypeMapDefinition.cs
+++ b/src/Umbraco.Web/Models/Mapping/ContentTypeMapDefinition.cs
@@ -455,7 +455,8 @@ namespace Umbraco.Web.Models.Mapping
 
                 // if the group has no local properties and is not used as parent, skip it, ie sort-of garbage-collect
                 // local groups which would not have local properties anymore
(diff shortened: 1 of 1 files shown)


lewishazell:
Just realised the expected behaviour should be that empty tabs and groups should be persisted.

Closing to correct.

PR #11286, opened 2021-10-05
Persist childless groups/tabs

### LINKED ISSUE #11069: Tab not saved when no content fields added

lewishazell:
### Prerequisites

If there's an existing issue for this PR then this fixes #11069

### Description

A check in the `ContentTypeMapDefinition` was removing childless groups and tabs, given tabs are a type of group, before saving. this check has been removed so that childless groups and tabs will be persisted.

#### Tests
As illustrated below,
- Create a new group/tab with no children
- Save the content type
- Switch to another page and return to the content type
- The group/tab should persist with no children

Da6bEW7Wy3

je7xYSllvn

### DIFF
--- a/src/Umbraco.Web/Models/Mapping/ContentTypeMapDefinition.cs
+++ b/src/Umbraco.Web/Models/Mapping/ContentTypeMapDefinition.cs
@@ -453,11 +453,6 @@ namespace Umbraco.Web.Models.Mapping
                     .Select(x => MapSaveProperty(x, destOrigProperties, context))
                     .ToArray();
 
-                // if the group has no local properties and is not used as parent, skip it, ie sort-of garbage-collect
-                // local groups which would not have local properties anymore
-                if (destProperties.Length == 0 && !sourceGroupParentAliases.Contains(sourceGroup.Alias))
-                    continue;
-
                 // ensure no duplicate alias, then assign the group properties collection
                 EnsureUniqueAliases(destProperties);
                 destGroup.PropertyTypes = new PropertyTypeCollection(isPublishing, destProperties);


src/Umbraco.Web/Models/Mapping/ContentTypeMapDefinition.cs
```diff
@@ -453,11 +453,6 @@ private static void Map(MemberPropertyTypeBasic source, MemberPropertyTypeDispla
                     .Select(x => MapSaveProperty(x, destOrigProperties, context))
                     .ToArray();
 
-                // if the group has no local properties and is not used as parent, skip it, ie sort-of garbage-collect
-                // local groups which would not have local properties anymore
-                if (destProperties.Length == 0 && !sourceGroupParentAliases.Contains(sourceGroup.Alias))
```
ronaldbarendse, line 458: We should probably add `sourceGroup.Inherited == true` to this condition, so empty 'local' groups that are already inherited from a composition are still garbage-collected.

lewishazell: Ah, that makes more sense, an empty inherited group is kind of useless. I thought this seemed a little to easy 😄

nathanwoulfe:
@lewishazell I'm going to take this for a spin. 

Re your last comments about being confusing if a user is shown an empty tab, I don't see that as an issue - if the content type has an empty tab, it makes sense to not show it, but if the tab has any children (groups or props), those should be shown (empty or not). 

There are plenty of ways to make the backoffice confusing for users (eg adding unused properties to a content type), so there really still needs to be responsibility on the implementor to do the right thing. I'd see adding a tab with an empty group as not the right thing 😄

Issue #11259, opened 2021-10-04
Slider breaks if switching datatype from numeric


bjarnef:
### Which *exact* Umbraco version are you using? For example: 8.13.1 - don't just write v8

9.0.0

### Bug summary

General speaking I think there is not great support for switching property editor on an existing property. Often you'll need to remove the property and re-publish the content node or add the property as a new property with a different alias.

E.g. if having an block settings property using Numeric datatype and switching property to a Slider datatype, it break here:
https://github.com/umbraco/Umbraco-CMS/blob/v9/contrib/src/Umbraco.Web.UI.Client/src/views/propertyeditors/slider/slider.controller.js#L49

This is due the value is stored as an numeric value, which hasn't a `split()` function.

We could probably improve this a bit further using `Utilities.isString()` to check whether `$scope.model.value` is a string and only split if it is otherwise fallback to null.
https://github.com/umbraco/Umbraco-CMS/blob/v9/contrib/src/Umbraco.Web.UI.Client/src/utilities.js#L58

### Steps to reproduce

- Create a property using Numeric datatype on a document type.
- On the content node save a numeric value in the property.
- Go back to the document type and change the datatype of the property to a Slider datatype.
- On the content node notice the console error and the slider isn't shown.

bjarnef:
I checked if it was an issue in latest v8 as well, but I couldn't reproduce on a content node or in content of a block element.
First when I changed a numeric property on a block settings element type to a slider datatype, I got the same issue, which was the same scenario as when I reported this issue.

PR #11287, opened 2021-10-05
Allow switching property editor from numeric to slider

bjarnef:
### Prerequisites

If there's an existing issue for this PR then this fixes https://github.com/umbraco/Umbraco-CMS/issues/11259

### Description
This PR allow switching property editor from numeric/decimal datatype to a slider datatype without breaking editor.
I couldn't reproduce on directly on a content node or as a property on a block, but I can reproduce when it is a block settings property. I guess there a different in how the value is stored.

However it would also allow changing from other datatypes storing a number/decimal value or at least improve it a bit.


mikecp:
Hi @bjarnef ,

This must have been a tricky one to catch, well done! 👍
Your PR solves the problem as announced, so let's merge it right away!

PR #11294, opened 2021-10-05
Removes annoying wait text, which causes layout jank

skttl:
### Prerequisites

### Description
On the Published Status dashboard, when loading or working, there is an incredibly annoying (Wait) text, that causes the layout to shift.

See this gif, where I've throttled my connection to emphasize the effect.
RUTbeTmiSC

I have removed the text, and the dictionary keys, as they aren't used anywhere else. I also added a z-index value to the overlay, to make it lay above the buttons of the dashboard.

Now it looks like this:
tmwFuzZdjV

To test:
Go to the Published Status dashboard, and verify that no (Wait) text is flashing while loading or working.


nathanwoulfe:
@nul800sebastiaan do you have a preference re labelling for changes to cherrypick up to 9?

PR #11304, opened 2021-10-06
Add support for media saving messages in dropzone

### LINKED ISSUE #11094: MediaSaving event messages always assumed as errors

matthewcare:
### Prerequisites

#11094 

### Description
Reworked how dropzone handles messages returned during the upload of new files.
Previously if you had a media saving event, and added a message, then this message was always assumed to be a server error message, which would prevent the finalising behaviour.

This means you would have to manually reload the page in order to see any new media.
current-functionality-umb8

This was even more of an issue in pickers, where you would have to close the picker, and re-open it again.

functionality-in-side

In Umbraco 9, functionality of messages was completely broken (see issue #11278), which results in no messages showing at all

current-functionality-umb9

In this PR, I have attempted to address these issues, and ended up with the following functionality
final-functionality
new-functionality-in-side

To test functionality, add a notification handler

```
namespace Umbraco.Cms.Web.UI
{
    public class ImageComponent : INotificationHandler<MediaSavingNotification>
    {
        public void Handle(MediaSavingNotification notification)
        {

            foreach (var entity in notification.SavedEntities)
            {
                if (entity.Name == "A Name")
                {
                    notification.Messages.Add(new EventMessage("Error", "Error Message", EventMessageType.Error));
                }

                if (entity.Name == "Another Name")
                {
                    notification.Messages.Add(new EventMessage("Info", "Info Message", EventMessageType.Info));
                    notification.Messages.Add(new EventMessage("Default", "Default Message"));
                    notification.Messages.Add(new EventMessage("Success", "Success Message", EventMessageType.Success));
                    notification.Messages.Add(new EventMessage("Warning", "Warning Message", EventMessageType.Warning));
                }


            }
        }
    }
}
```

Updating the `entity.Name` as required, or add event messages to everything if that's better for you.

Add the notificaiton to the builder in startup

```
services.AddUmbraco(_env, _config)
    .AddBackOffice()
    .AddWebsite()
    .AddComposers()
    .AddNotificationHandler<MediaSavingNotification, ImageComponent>()
    .Build();
```


matthewcare:
Reasoning behind this functionality being required:
I have a media saving event which resizes images to a configured size, and want to alert the user that their media has been resized. Also image optimisation tasks, which can display how much the file size has reduced.
Current Umbraco 8 functionality is hindering users from uploading media, as they are forced to open and close media pickers, and when upgrading to Umbraco 9, they just didn't work at all

mikecp (approved):

mikecp:
Hi @matthewcare,,

Thanks for this PR, this is really great work! 👍 

It works as described, but I had to resolve a conflict, so I'll wait for the tests to be ran successfully again and then I'll be happy to merge this 🎉

mikecp:
Hi again @matthewcare ,

The tests have run and, unfortunately, it seems that some "cy" tests are failing. This is probably due to the fact that the PR changes the structure of the file umb_file_dropzone.html, and there is probably one of the cy tests that relies on that previous structure 😕

It would be great if you could have a look at what is going wrong and adapt the cy 😁👍

I also had one question/suggestion that I forgot to mention yesterday: could it be interesting to have an "OK to all" button that would kind of simulate the click on all visible "OK" buttons and refresh the page at once?

My reasoning was the following: imagine that the user uploads 100 pictures and somehow gets 50 warnings/feedback messages displayed, I don't think s.he will enjoy clicking on 50 OK buttons to get the page refreshed 😅
Of course it is possible to reload the page, but since one of the goals of the PR was to avoid that, I figured it could be nice to provide an alternative solution in that case as well.

What do you think?

mikecp (changes requested):
Some errors with the tests need to be taken care of.

matthewcare:
Hey @mikecp 
I've added your suggestion of an "ok to all", I'd intended to do it originally, but somehow forgot about it :(
I'm running the Cypress tests locally now, and I am seeing some of them fail, but I'm not seeing them be due to anything I've changed. I've watched all of the videos Cypress creates, and none  of them are media related.

image

mikecp (approved):

mikecp (approved):

mikecp:
Hey @matthewcare ,

Thanks for the update, it's all working great! The failing tests don't seem to be linked to he PR changes indeed, so they where probably introduced in a previous merge...

Anyway, this is a nice addition to the CMS, thanks 👍

PR #11306, opened 2021-10-06
Update ImageSharp to version 1.0.4

### LINKED ISSUE #10961: ImageSharp does not appear to handle some PNG files properly

V9-RC002

### Bug summary

I've been messing around trying to work out if there is a bug in Umbraco or ImageSharp but in the latest version I've noticed images are not rendering properly in some cases now, anything that is a PNG won't render properly. I've attached one for testing with, it renders about 30% of the way & then glitches and that's it.

Any help to figure out why this is happening would be great!
pump99logo-1

### Specifics

https://staging.pump99.co.uk/ as an example of what it looks like on the front-end, scroll to the bottom of the page, JPG files are unaffected it seems

### Steps to reproduce

Not 100% sure, seems to be PNG files but not all of them causing issues, if I had to hazard a guess I'd say it seems like ImageSharp is still too unstable to use really, I did try to use the pre-release version & rebuild Umbraco, but that caused all sorts of fun with .netstandard so I gave up & am taking a break from it now.

The image should render without issue, or if there is a problem manipulating the image, then it should catch this error & at least serve the original full size file so the site doesn't look terrible

brianpopow:
### Prerequisites

If there's an existing issue for this PR then this fixes #10961

### Description

This updates the ImageSharp reference from 1.0.3 to 1.0.4. This hotfix of ImageSharp fixes an issue with [png images with .net6.0](https://github.com/SixLabors/ImageSharp/issues/1704). 
Since .net6.0 release is around the corner, I thought it would be good to update the reference in advance.

### DIFF
--- a/src/Umbraco.Infrastructure/Umbraco.Infrastructure.csproj
+++ b/src/Umbraco.Infrastructure/Umbraco.Infrastructure.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
         <TargetFramework>netstandard2.0</TargetFramework>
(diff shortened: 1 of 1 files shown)


bergmania:
ImageSharp.Web should be updated at the same time, right

ronaldbarendse:
ImageSharp.Web 1.0.4 also fixes a bug with the order in which the commands are processed, making our workarounds obsolete. PR https://github.com/umbraco/Umbraco-CMS/pull/11349 removes these workarounds.

nul800sebastiaan:
Hi @brianpopow, you might have missed the call for Hacktoberfest rewards since I have no registration for you. Please fill in the Hacktoberfest form before the end of the week if you want swag, else we'll plant trees 👍  https://docs.google.com/forms/d/e/1FAIpQLScaxJCay8M-uvyvNmX7o4xVibuMh_yfokjk3csyqFuGwb7DgA/viewform

PR #11307, opened 2021-10-06
add create language test

jemayn:
### Prerequisites

### Description

Adds a new Cypress test that creates a language.

I have a question about adding new cypress commands like the one I added at the bottom of my test file here. Seems all commands are currently imported from Bjarkes package - however that makes it super difficult to test and add new ones.

What is the preferred way to deal with this? Am I meant to do a PR to the package with the command and then open a PR here that depends on the other one being closed and a new NPM version of the package being released?

Doesn't seem that nice of a workflow for outside collaborators 🙂


jemayn:
In later PRs (fx #11333) I've started adding the commands in the commands.js file which is the Cypress convention, I'm happy to move it wherever, but probably needs someone to make a decision about how it should be done 🙂

bergmania:
> > @jemayn.. It's totally fine to just add them in ` commands.js` directly in Umbraco.
> 
> Awesome @bergmania, what about the TS definitions then? 🙂 Or do I just add them without those? (The file for it is gitignored)

I have no idea, we need a frontend dev to answer that 🙈  @madsrasmussen can you help answer 🙏

jemayn:
Closing this and remaking, did a rebase which for some reason shows all changes instead of just the ones between my branch and the target 🤷‍♂️

PR #11308, opened 2021-10-06
New interface for custom backoffice assets in C#

patrickdemooij9:
I noticed that there was a way to register js and CSS scripts within C#, but I think the current interface for it is not the greatest.
I also didn't even know that this existed, so I'll certainly have to write some documentation for this later (depending on what will happen with this PR of course).

I decided to remove the OrderedCollectionBuilder that was being used here as adding a new type for each file is just undoable.

The interface will then change from:
```
public class CustomPackageComposer : IComposer
    {
        {
            builder.BackOfficeAssets().Append<CustomPackageScript>();
        }
    }

    public class CustomPackageScript : JavaScriptFile
    {
        public CustomPackageScript() : base("/App_Plugins/Test/testJavascript.js") { }
    }
```

To
```
public class CustomPackageComposer : IComposer
    {
        {
            builder.BackOfficeAssets().AddAssetFile(new JavaScriptFile("/App_Plugins/Test/testJavascript.js"));
        }
    }
```

This is mostly how the TreeCollectionBuilder also handles TreeDefinitions.
However, this will be a breaking change, so I am wondering how to best proceed with this!

### DIFF
--- a/src/Umbraco.Core/WebAssets/CustomBackOfficeAssetsCollectionBuilder.cs
+++ b/src/Umbraco.Core/WebAssets/CustomBackOfficeAssetsCollectionBuilder.cs
@@ -1,9 +1,22 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.DependencyInjection;
 using Umbraco.Cms.Core.Composing;
 
 namespace Umbraco.Cms.Core.WebAssets
 {
-    public class CustomBackOfficeAssetsCollectionBuilder : OrderedCollectionBuilderBase<CustomBackOfficeAssetsCollectionBuilder, CustomBackOfficeAssetsCollection, IAssetFile>
+    public class CustomBackOfficeAssetsCollectionBuilder : ICollectionBuilder<CustomBackOfficeAssetsCollection, IAssetFile>
     {
-        protected override CustomBackOfficeAssetsCollectionBuilder This => this;
+        private readonly List<IAssetFile> _files = new List<IAssetFile>();
+
+        public CustomBackOfficeAssetsCollection CreateCollection(IServiceProvider factory) => new CustomBackOfficeAssetsCollection(() => _files);
+
+        public void RegisterWith(IServiceCollection services) => services.Add(new ServiceDescriptor(typeof(CustomBackOfficeAssetsCollection), CreateCollection, ServiceLifetime.Singleton));
+
+        public void AddAssetFile(IAssetFile file)
+        {
+            if (file is null) throw new ArgumentNullException(nameof(file));
+            _files.Add(file);
+        }
     }
 }


nathanwoulfe:
@patrickdemooij9 I think this would need to wait for v10, and should probably have some input from HQ. Does make sense to me though, a new class per script registration feels a bit heavy handed. @nul800sebastiaan thoughts on this one?

nzdev:
Perhaps this could be written as an extension method instead?

patrickdemooij9:
> Perhaps this could be written as an extension method instead?

@nzdev. Not quite sure if that is possible (but feel free to suggest a way, I might just see it!). In the current version it expects types while in my version it actually expects instances. I could try to make Append an extension/method in my version that'll then generate an instance based on the type that you have given if we really want to keep the old syntax around.

But personally, I don't think we should really keep the old syntax. The only advantage of the old approach is that you can use dependency injection in your asset files, but I don't know if that really is an valid use case.

nzdev:
Best way would be to override  IEnumerable<TItem> CreateItems(IServiceProvider factory), call the base method and add the rest that you need. That way only the docs need updating and there are no code changes required.

patrickdemooij9:
That does seem like an interesting collection to have, especially if there is a requirement from Umbraco to keep the old logic in place as well. However, is it really ordered now? Doesn't it always show the typed instances in front of the manually added instances?

nzdev:
Could be modified further by overriding the createitems option.

abjerner:
I just added a discussion about introducing an [`IPackageManifest` interface](https://github.com/umbraco/Umbraco-CMS/discussions/12054). Maybe it's more on the long term, but couldn't this be a solution as well?

nzdev:
Sounds good to me @abjerner

JasonElkin:
Hey @patrickdemooij9,

Sorry that this got stuck in PR limbo. It's certainly a better way of doing it, but it's not a change that we can make now.

As it's breaking, and this method of adding JavaScript to the backoffice is going away in v13, we can't deprecate while keeping to the "next ?major + 1" promise.

In the mean time, and in light of @abjerner's comment, I can recommend taking a look at the `IManifestFilter` which is becoming a more well-documented approach.

It's discussed in [Kevin Jump's RCL package article](https://dev.to/kevinjump/umbraco-10-razor-class-library-packages-pt2-58an) and used in the [Opinionated Package Starter](https://github.com/LottePitcher/opinionated-package-starter).

abjerner:
Just to add up, I created the discussion (Feb 2022 seems so no ago) because I back then didn't know about manifest filters, but I've used them extensively since then for my packages, and I now consider them *the* way to go. They can do the same as `.BackOfficeAssets()`, and more to - and are IMO more user (developer) friendly 😃

